PR #6089, opened 2019-01-10
Add context menu item "Copy link" to Commit Info

### LINKED ISSUE #6088: Copy links to clipboard
## Feature description

Add a context menu item `Copy link` to the Commit Info which copies the link under the mouse to the Clipboard.

#### new
grafik

#### current
grafik

## Environment

- Git Extensions 3.01.00.0
- Build f7e97be0cc4190340516d86ab661f72a7e95f2b3 (Dirty)
- Git 2.20.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6088

## Proposed changes

- add a context menu item `Copy link` to the Commit Info which copies the link under the mouse to the Clipboard
- add shortcuts to other menu items which trigger actions
- remove the unused function and its overload `RichTextBoxXhtmlSupportExtension.AddLink()`

### Before

grafik

### After

grafikgrafik

## Test methodology

- manual testing
- NUnit test for `RichTextBoxXhtmlSupportExtension.GetLink()`

## Test environment

- Git Extensions 3.01.00.0
- Build 6c16ec5a9a04632859ec2d1f882f36eb9bce5959
- Git 2.20.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

### DIFF
--- a/GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
+++ b/GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
@@ -389,22 +389,6 @@ namespace GitUI.Editor.RichTextBoxExtension
             return (cf.dwEffects & CFE.LINK) == CFE.LINK;
         }
 
-        private static void AddLink(this RichTextBox rtb, string text)
-        {
-            int position = rtb.SelectionStart;
-            if (position < 0 || position > rtb.Text.Length)
-            {
-                throw new InvalidOperationException("SelectionStart has invalid value");
-            }
-
-            rtb.SelectionStart = position;
-            rtb.SelectedText = text;
-            int length = rtb.SelectionStart - position;
-            rtb.Select(position, length);
-            rtb.SetLink(true);
-            rtb.Select(position + length, 0);
-        }
-
         private static void RtbSetSelectedRtf(RichTextBox rtb, string str)
         {
             // Work around bug in DotNet.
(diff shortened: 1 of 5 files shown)


GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
```diff
@@ -1146,6 +1146,80 @@ public static string GetPlainText(this RichTextBox rtb, int from, int to)
...
+                // get something like an array of charformat and paraformat
+                //--------------------------------
+                rtb.Select(charIndex, 1);
+                if (!rtb.IsLink())
+                {
+                    return null;
+                }
+
+                int from = charIndex;
+                while (true)
```
gerhardol: How can we sure that this is not endless?

mstv: `from` is decremented in every iteration.
Though I should change it to `while (from > 0)`.
The same applies to the loop for `to`.

I have not written the NUnit test yet.

mstv:
If I remember correctly, the updated file `GitUI/Translation/English.xlf` should not be commited.

gerhardol:
> If I remember correctly, the updated file GitUI/Translation/English.xlf should not be commited.

Not required, it will be updated by running the TranslationApp.
But it has to be done to see changed strings.

(I would like to add a recommendation to run the Translation app in PRs, to avoid that translation issues are popping up just before a release.)

vbjay:
> I would like to add a recommendation to run the Translation app in PRs, to avoid that translation issues are popping up just before a release.)

I disagree.  Our code is the source of translation strings.   XLF files are an output of processing the files.  The release process contains a step to prepare translations.  Our pull requests should not contain changes not related to solving the problems.  All it takes is to run the translation app and generate xlf change when release is being prepared.

mstv:
I have rebased, added NUnit tests and removed the unused function and its overload `RichTextBoxXhtmlSupportExtension.AddLink()`.

mstv:
Added another test case.

RussKie:
You seem to have grown fond of tests 😁
👍

GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
```diff
@@ -1146,6 +1101,85 @@ public static string GetPlainText(this RichTextBox rtb, int from, int to)
...
+                    {
+                        ++to;
+                    }
+                }
+
+                return text.Substring(from, to - from);
+            }
+            catch (Exception /*ex*/)
+            {
+                ////MessageBox.Show(ex.Message);
```
RussKie: I would return `null` here instead of line 1180, as it feels more logical.
Also please remove the commented code.

gerhardol:
> You seem to have grown fond of tests

These tests makes sense too, actually testing something.

GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
mstv: Yes, copied from an other function and not tought to the end...

RussKie:
All our tests test something - some test small nuanced things, some tests big 😏

PR #6100, opened 2019-01-14
Tests for GitRemoteManager.GetDisabledRemotes, GetDisabledRemoteNames…

amaiorano:
…, GetEnabledRemoteNames, GetEnabledRemotesNameWithoutBranches

## Proposed changes

- Adding tests for new functions

## Test environment(s) 

- GIT 2.19.1.windows.1
- Windows 10

### DIFF
--- a/GitCommands/Remotes/GitRemoteManager.cs
+++ b/GitCommands/Remotes/GitRemoteManager.cs
@@ -52,6 +52,26 @@ namespace GitCommands.Remotes
         /// </summary>
         /// <param name="remoteName">The name of the remote.</param>
         void ToggleRemoteState(string remoteName, bool disabled);
+
+        /// <summary>
+        /// Retrieves disabled remotes from .git/config file
+        /// </summary>
+        IReadOnlyList<Remote> GetDisabledRemotes();
+
+        /// <summary>
+        /// Retrieves disabled remote names from the .git/config file.
+        /// </summary>
+        IReadOnlyList<string> GetDisabledRemoteNames();
+
+        /// <summary>
+        /// Retrieves enabled remote names
+        /// </summary>
+        IReadOnlyList<string> GetEnabledRemoteNames();
+
+        /// <summary>
+        /// Retrieves enabled remote names of remotes without branches (i.e. that require a fetch)
+        /// </summary>
+        IReadOnlyList<string> GetEnabledRemoteNamesWithoutBranches();
     }
 
     public class GitRemoteManager : IGitRemoteManager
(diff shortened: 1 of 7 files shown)


RussKie:
🎉 pushed in a set of UI tests for remotes in `RepoObjectTree`
Please have a look and add more tests as necessary

While adding these I discovered a bug in remotes handling...

amaiorano:
> For some reasons `RepoObjectTree_should_order_inactive_remotes_alphabetically` test hung...
> 😕

Cool tests! When I run them, they don't hang. You're saying it hangs intermittently? Where is that output you pasted from?

RussKie:
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21626573/tests
search for "RepoObjectTree_should_order_inactive_remotes_alphabetically"

amaiorano:
> https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21626573/tests
> search for "RepoObjectTree_should_order_inactive_remotes_alphabetically"

So when an entry is yellow with no time, this indicates a hung test?

RussKie:
The build went on for 37+ minutes, which is longer than our typical build
of 10-12 minutes. I cancelled the build and that was the test that was
running at the time. There are 3 other tests which haven't run at that
point.

amaiorano:
So one thing that has happened to me a couple of times when developing is that I try to execute a command in GitEx, and it opens the ConEmu dialog, and nothing happens. Basically, the ConEmu dialog is effectively empty. I debugged this quickly last night to realize that the ConEmu.exe was not being found. So effectively, my command hangs. I was able to fix this with a rebuild of the project, which I guess rebuilds ConEmu and outputs it in the right place. I didn't spend much more time on it. I have no idea why the ConEmu.exe was suddenly gone, and it has happened to me perhaps three times in the last few months, and one of those was last night. Perhaps unrelated, but since we're talking about a hang in GE, I thought it might be worth bringing up. Regardless, this error of missing ConEmu.exe should fail this way.

amaiorano:
> Yes, it is happening to me every now and then too. I could never find why either.

When a test hangs like this, do we have a way to grab GE's log?

amaiorano:
> No. We can tweak the build, if necessary though

Feels like it would be useful to see the GitEx output from integration tests. Hangs like this can be difficult to diagnose, especially when difficult to reproduce.

vbjay:
The build script does a rebuild. Appveyor calls build script. You can also right click solution and click rebuild. If you do a git clean you will need to do a rebuild.

amaiorano:
> ...and now it built successfully 😠

Which would match the idea that this is an intermittent failure. How difficult would it be to include the logs of the running instances of GitEx for integration tests?

vbjay:
Not hard.  If we output anything then a simple redirect will log the
output.

Gitext ...  > LogfileHere.txt

RussKie:
I'm not sure it's that easy. Happy to be proven wrong

amaiorano:
So what do we do? Can we merge this as is?

RussKie:
Squash and merge, if there are no more tests to write

amaiorano:
In this case, I'd rather keep the two commits separate as-is. Cool?

RussKie:
sure

PR #6107, opened 2019-01-15
Fix side panel sometimes ending up with duplicate sets of remotes whe…

### LINKED ISSUE #6106: Side panel activate remotes: duplication
## Current behaviour
After #6015 
Activate&Fetch duplicated the remotes
(a second activation did not add a third)

## Expected behaviour

No duplication

## Steps to reproduce

Activate a remote

## Screenshots
image

no, new in master

## Environment

- Git Extensions 3.01.00.0
- Build 043c96c6ee00671091d55197f6a3512bdc5d9402 (#6105 )
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

## Diagnostics

n/a

amaiorano:
…n "Activate & fetch" selected on an inactive remote

Problem is that the "fetch" operation would invoke UICommands.RepoChangedNotifier.Notify(), then this same function would get invoked yet again right after. This second call would sometimes cancel the first, which would cancel the repopulation of nodes in the TreeView.

Fixes #6106

## Proposed changes

- Don't call UICommands.RepoChangedNotifier.Notify() twice in a row from the side panel as this might cancel the tree view update

## Test methodology 

- Manual testing. This was easier to reproduce without a debugger attached (timing), and when activating a repo with many branches (e.g. gerhardol's fork)

## Test environment(s) 

- GIT 2.19.1.windows.1
- Windows 10

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
@@ -288,10 +288,13 @@ namespace GitUI.BranchTreePanel
                 _gitRemoteManager.ToggleRemoteState(Name, disabled: false);
                 if (fetch)
                 {
+                    // DoFetch invokes UICommands.RepoChangedNotifier.Notify
                     DoFetch();
                 }
-
-                UICommands.RepoChangedNotifier.Notify();
+                else
+                {
+                    UICommands.RepoChangedNotifier.Notify();
+                }
             }
 
             public void Disable()


amaiorano:
@gerhardol can you try this branch and see if it fixes the problem for you?

RussKie:
I'm adding a scaffold for ROT UI tests, we should be able simulate the
issue and write a test to confirm the fix.

gerhardol:
Will try tonight

amaiorano:
> I'm adding a scaffold for ROT UI tests, we should be able simulate the issue and write a test to confirm the fix.

Maybe, but it was actually very difficult to reproduce, as you can probably see from my comments above. I managed to reproduce it by figuring out what the problem might be. In fact, I modified the code to invoke ```UICommands.RepoChangedNotifier.Notify();``` three times in a loop, then detached my debugger and was able to reproduce more easily. In truth, this might mean we have a deeper problem related to cancellation in general for the left panel.

gerhardol:
Works fine
Testing with debug build of GE, standalone (not in debugger), Activate&Fetch followed by Deactivate of drewnoakes in my GE repo with 113 branches an6 6+25 remotes.
Before, I got duplication 5 out of 5 times and after I got none of 10 times.

> In truth, this might mean we have a deeper problem related to cancellation in general for the left panel.

Will your pubsub branch improve here?

amaiorano:
> Will your pubsub branch improve here?

Maybe. I think we just have to be careful not to issue more than one refresh from the left panel code itself, since a second refresh cancels the first.

Thanks for the review. Can I merge this immediately?

gerhardol:
> Thanks for the review. Can I merge this immediately?

I do not think that the chief will mind that, go ahead.

RussKie:
I can see you 😎

Issue #6096, opened 2019-01-12
NBug and other assemblies reports 3.00.rc2


gerhardol:
## Current behaviour

The following files have set the display information:

[assembly: AssemblyInformationalVersion("3.00.rc2")]

Externals/conemu-inside/ConEmuWinForms/Properties/AssemblyInfo.cs
Externals/Git.hub/Git.hub/Properties/AssemblyInfo.cs
Externals/ICSharpCode.TextEditor/Project/Properties/AssemblyInfo.cs
Externals/NBug/NBug/Properties/AssemblyInfo.cs
NetSpell.SpellChecker/Properties/AssemblyInfo.cs

This can be confusing in for instance NBug popups
Seen in #6093 

## Expected behaviour

Either the value should be updated (including the subrepos) or set to a generic value (preferred).

## Did this work in previous version of GitExtensions

This was changed in the "Fix assembly info" update in the beginning of December, generic before.

## Environment
- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

vbjay:
https://github.com/gitextensions/gitextensions/commit/9bc4321707b36432b644ade89a2a3001deeaf1f4

PR #6110, opened 2019-01-15
WIP Update appveyor.yml

RussKie:
Set version to 3.1.0 instead of 3.01.00

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 3.01.00.{build}
+version: 3.1.0.{build}
 
 os: Visual Studio 2017
 


gerhardol:
Should update versions in GE source too, as well as #6096

RussKie:
Do you mean `AssemblyInfo`s?
```
[assembly: AssemblyVersion("3.01.00.0")]
[assembly: AssemblyFileVersion("3.01.00.0")]
[assembly: AssemblyInformationalVersion("3.01.00.0")]
```

gerhardol:
> Like this

Yes, but it is required to always update all these versions?
For instance set AssemblyInformationalVersion to something neutral so it does not have to be changed?

vbjay:
It's standard to have a version bump commit. Run prepare release script and
look at the changes.

gerhardol:
> It's standard to have a version bump commit. Run prepare release script and look at the changes.

Yhese are not in the prepare release script. (Several of them were changed just before 3.0.0 release. Even if they are added, this means more maintenance, submodules must be added too which is inconvenient, a submodule should never contain a reference to the superproject.)

vbjay:
No. Prepare release is in main repo. So is setversion script. They are in
setup folder. If other version files need to be changed, then add them to
the python script.

RussKie:
I think we should update these as part of the build, we shouldn't be doing this manually.
I'll add/modify our scripts to do so.

vbjay:
They already are submodules. They are already forked. We have already
adjusted some things on them.

gerhardol:
> They already are submodules. They are already forked. We have already adjusted some things on them.

We do not want to have multiple branches if not needed: One for each GE branch we are working with as they contain information about the superproject. It is this scenario that users to hate submodules.

RussKie:
@vbjay may I please get help?
I've update the python script to update **all** `AssemblyInfo.cs` files, including submodules.
https://github.com/gitextensions/gitextensions/blob/release/3.0.2/Setup/set_version_to.py#L26-L28

As a result the build comes out as dirty, e.g. 
image
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/22050162#L68

How should I script it to ignore the changes to `AssemblyInfo.cs` files?

vbjay:
So the fix is to turn on the git status debugging piece of the appveyor script.  It will give you a status artifact.  Then add skipworktree for each path that is dirty.  If in a submodule do what I did for the docs submodule.  

Non release builds are always dirty.

Issue #6093, opened 2019-01-11
System.Exception opening an existing repository


fabvit:
## Current behaviour

1. Open one of my local repositories
after the repository is opened I got a System exception:
```
Unable to parse git command output.

Target System: System.Collections.Generic.IReadOnlyList`1[GitCommands.Remote] <GetRemotes>g__ParseRemotes|154_0(System.Collections.Generic.IEnumerable`1[System.String]) @ GitCommands.GitModule
```
Application: GitExtensions.exe [3.00.00.4433]

NBug: 3.00.rc2

CLR: 4.0.30319.42000

Please note that thes exception happens *only* when the left panel (the one enumerating branches and remotes) is visible. 
If I hide the the left panel, no exception arises.

## Expected behaviour

No exceptions opening a repo.

## Steps to reproduce

See Current behaviour section above

## Did this work in previous version of GitExtensions

Yes it works fine in GitExtensions-2.51.05

## Environment

- GitExtensions version: GitExtensions-3.00.00.4433
- GIT version: $ git --version
git version 2.20.1.windows.1

- OS version: Windows 7 Pro 64 bit
- .NET version: 4.0.30319.42000

vbjay:
https://github.com/gitextensions/gitextensions/releases/latest

gerhardol:
Please include the complete stacktrace from NBug
If you dismiss the dialog, can you continue
Can you share the repo where this occurred if the stacktrace is not obvious

I believe @vbjay reacted on that NBug said 3.00.rc2. That may be incorrectly set there if so, for someone to follow up.

vbjay:
Oh right. That issue has been fixed.

fabvit:
@gerhardol and @vbjay yes I am on the very last version (https://github.com/gitextensions/gitextensions/releases/latest)
@vbjay at this point I cunfused: what issue was been fixed?

        1. the wrong NBug: 3.00.rc2 text
*OR*
        2. my issue of System.Exception when opening an existing repository

Yes I can continue after dismissing the dialog  *and* hiding the left panel.

Unfortunately I cannot share the repo.

vbjay:
Yes.  the wrong NBug: 3.00.rc2 text

gerhardol:
The complete stacktrace is needed

I am not aware that this is reported before (but there are related corrections fixed in the upcoming 3.0.1
The complete stacktrace is needed

How many remotes do you have?
What is seen in the command log? (F12)

[...]

PR #6037, opened 2019-01-04
Solved column size problem in FileStatusList

NikolayXHD:
Fixes #5917

## Proposed changes

Turns out the problem was caused by setting `Dock.Fill` to `ListView` control.
After figuring that out there is no need in hack with resizing column after a small delay.

## Test methodology

Manual

## Test environment(s)

Windows 7 SP1

PR #6111, opened 2019-01-15
Fix AE: Can not add a section that already exists: -remote

RussKie:
Discovered while adding UI tests in #6100

Use case:
- a user has added a remote
- then deactivated the remote via GE
- then added a remote with the same name from a command line or via UI
- then attempted to deactivate the new remote

Application then crash with:

    System.ArgumentException : Can not add a section that already exists: -remote
       at GitCommands.Config.ConfigFile.AddConfigSection(IConfigSection configSection) in C:\Development\gitextensions\GitCommands\Config\ConfigFile.cs:line 253
       at GitCommands.Settings.ConfigFileSettingsCache.<>c__DisplayClass9_0.<AddConfigSection>b__0() in C:\Development\gitextensions\GitCommands\Settings\ConfigFileSettingsCache.cs:line 79
       at GitCommands.SettingsCache.<>c__DisplayClass3_0.<LockedAction>b__0() in C:\Development\gitextensions\GitCommands\Settings\SettingsCache.cs:line 26
       at GitCommands.SettingsCache.LockedAction[T](Func`1 action) in C:\Development\gitextensions\GitCommands\Settings\SettingsCache.cs:line 35
       at GitCommands.SettingsCache.LockedAction(Action action) in C:\Development\gitextensions\GitCommands\Settings\SettingsCache.cs:line 24
       at GitCommands.Settings.ConfigFileSettingsCache.AddConfigSection(IConfigSection configSection) in C:\Development\gitextensions\GitCommands\Settings\ConfigFileSettingsCache.cs:line 76
       at GitCommands.Settings.ConfigFileSettings.AddConfigSection(IConfigSection configSection) in C:\Development\gitextensions\GitCommands\Settings\ConfigFileSettings.cs:line 113
       at GitCommands.Remotes.GitRemoteManager.ToggleRemoteState(String remoteName, Boolean disabled) in C:\Development\gitextensions\GitCommands\Remotes\GitRemoteManager.cs:line 341
       at GitCommandsTests.Remote.GitRemoteManagerTests.IntegrationTests.ToggleRemoteState_should_not_fail_if_deactivate_repeatedly() in C:\Development\gitextensions\UnitTests\GitCommandsTests\Remote\GitRemoteManagerTests.cs:line 368

### DIFF
--- a/GitCommands/Remotes/GitRemoteManager.cs
+++ b/GitCommands/Remotes/GitRemoteManager.cs
@@ -31,6 +31,20 @@ namespace GitCommands.Remotes
         string RemoveRemote(GitRemote remote);
 
         /// <summary>
+        /// Returns true if input remote exists and is enabled.
+        /// </summary>
+        /// <param name="remoteName">Name of remote to check.</param>
+        /// <returns>True if input remote exists and is enabled.</returns>
+        bool EnabledRemoteExists(string remoteName);
+
+        /// <summary>
+        /// Returns true if input remote exists and is disabled.
+        /// </summary>
+        /// <param name="remoteName">Name of remote to check.</param>
+        /// <returns>True if input remote exists and is disabled.</returns>
+        bool DisabledRemoteExists(string remoteName);
+
+        /// <summary>
         ///   Saves the remote details by creating a new or updating an existing remote entry in .git/config file.
         /// </summary>
         /// <param name="remote">An existing remote instance or <see langword="null"/> if creating a new entry.</param>
@@ -240,6 +254,26 @@ namespace GitCommands.Remotes
         }
 
         /// <summary>
+        /// Returns true if input remote exists and is enabled.
+        /// </summary>
+        /// <param name="remoteName">Name of remote to check.</param>
+        /// <returns>True if input remote exists and is enabled.</returns>
+        public bool EnabledRemoteExists(string remoteName)
+        {
+            return GetEnabledRemoteNames().FirstOrDefault(r => r == remoteName) != null;
+        }
+
+        /// <summary>
+        /// Returns true if input remote exists and is disabled.
+        /// </summary>
+        /// <param name="remoteName">Name of remote to check.</param>
+        /// <returns>True if input remote exists and is disabled.</returns>
+        public bool DisabledRemoteExists(string remoteName)
+        {
+            return GetDisabledRemoteNames().FirstOrDefault(r => r == remoteName) != null;
+        }
+
+        /// <summary>
         ///   Saves the remote details by creating a new or updating an existing remote entry in .git/config file.
         /// </summary>
         /// <param name="remote">An existing remote instance or <see langword="null"/> if creating a new entry.</param>
@@ -275,6 +309,13 @@ namespace GitCommands.Remotes
             if (creatingNew)
             {
                 output = module.AddRemote(remoteName, remoteUrl);
+
+                // If output was returned, something went wrong
+                if (!string.IsNullOrWhiteSpace(output))
+                {
+                    return new GitRemoteSaveResult(output, false);
+                }
+
                 updateRemoteRequired = true;
             }
             else
--- a/GitCommands/Remotes/GitRemoteManager.cs
+++ b/GitCommands/Remotes/GitRemoteManager.cs
@@ -341,8 +341,22 @@ namespace GitCommands.Remotes
                 module.LocalConfigFile.RemoveConfigSection($"{sectionName}.{remoteName}");
             }
 
+            var newSectionName = (disabled ? DisabledSectionPrefix : "") + SectionRemote;
+
+            // ensure that the section with the same name doesn't already exist
+            // use case:
+            // - a user has added a remote,
+            // - then deactivated the remote via GE
+            // - then added a remote with the same name from a command line or via UI
+            // - then attempted to deactivate the new remote
+            var dupSection = sections.FirstOrDefault(s => s.SectionName == newSectionName && s.SubSection == remoteName);
+            if (dupSection != null)
+            {
+                module.LocalConfigFile.RemoveConfigSection($"{newSectionName}.{remoteName}");
+            }
+
             // rename the remote
-            section.SectionName = (disabled ? DisabledSectionPrefix : "") + SectionRemote;
+            section.SectionName = newSectionName;
 
             module.LocalConfigFile.AddConfigSection(section);
             module.LocalConfigFile.Save();
--- a/GitUI/CommandsDialogs/FormRemotes.cs
+++ b/GitUI/CommandsDialogs/FormRemotes.cs
@@ -82,6 +82,12 @@ Inactive remote is completely invisible to git.");
 
         private readonly TranslationString _lvgDisabledHeader =
             new TranslationString("Inactive");
+
+        private readonly TranslationString _enabledRemoteAlreadyExists =
+            new TranslationString("An active remote named \"{0}\" already exists.");
+
+        private readonly TranslationString _disabledRemoteAlreadyExists =
+            new TranslationString("An inactive remote named \"{0}\" already exists.");
         #endregion
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
@@ -360,6 +366,18 @@ Inactive remote is completely invisible to git.");
                     checkBoxSepPushUrl.Checked = false;
                 }
 
+                if (_remoteManager.EnabledRemoteExists(remote))
+                {
+                    MessageBox.Show(this, string.Format(_enabledRemoteAlreadyExists.Text, remote), _gitMessage.Text, MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
+                    return;
+                }
+
+                if (_remoteManager.DisabledRemoteExists(remote))
+                {
+                    MessageBox.Show(this, string.Format(_disabledRemoteAlreadyExists.Text, remote), _gitMessage.Text, MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
+                    return;
+                }
+
                 // update all other remote properties
                 var result = _remoteManager.SaveRemote(_selectedRemote,
                                                        remote,
@@ -369,7 +387,8 @@ Inactive remote is completely invisible to git.");
 
                 if (!string.IsNullOrEmpty(result.UserMessage))
                 {
-                    MessageBox.Show(this, result.UserMessage, _gitMessage.Text);
+                    MessageBox.Show(this, result.UserMessage, _gitMessage.Text, MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
+                    return;
                 }
                 else
                 {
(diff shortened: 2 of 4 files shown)


amaiorano:
Did you also try:

* a user has added a remote
* then deactivated the remote via GE
* then added a remote with the same name from a command line or via UI
* **then attempted to reactivate the old remote**

?

RussKie:
I was planning to test that use case, but then got sidetracked and forgot. Thank you for reminding, I'll add it too.

gerhardol:
> we would lose the old deactivated one, right? Is that okay?

Better than a crash, so I say yes.
Compare to #6093

gerhardol:
I end up with two inactive remotes, which probably is better

image

RussKie:
> I end up with two inactive remotes, which probably is better

wow, how did you do that?
the code is leaky as, it seems...

RussKie:
> Is that okay?

At this stage I think this suffices. We can certainly enhance it, but it would take more time than I can allocate to this.

gerhardol:
> wow, how did you do that?

As you wrote in GitRemoteManager.cs
```         
            // - a user has added a remote,
            // - then deactivated the remote via GE
            // - then added a remote with the same name from a command line or via UI
            // - then attempted to deactivate the new remote
```           
Now the gui has two inactive remotes

amaiorano:
All I was trying to say is that, like Git, GitEx doesn't do anything destructive without asking the user if they're sure. In this case, we would silently overwrite the disabled value in the config, right?

vbjay:
I think we could do better. We could read in remotes and warn user when
adding that they already have a disabled remote setup with that name. If
they didn't use git extensions to modify remotes then we will take the
active one in the config and wipe out the disabled one.

RussKie:
We can for sure.
This is an edge case, so the priority for me is relatively low.
PRs are welcome.

RussKie:
Thanks you Gerhard. I repro'ed and fixed the UI glitch.

GitCommands/Remotes/GitRemoteManager.cs
```diff
@@ -274,8 +274,26 @@ public GitRemoteSaveResult SaveRemote(GitRemote remote, string remoteName, strin
             bool remoteDisabled = false;
             if (creatingNew)
             {
-                output = module.AddRemote(remoteName, remoteUrl);
-                updateRemoteRequired = true;
+                if (GetEnabledRemoteNames().FirstOrDefault(r => r == remoteName) != null)
```
amaiorano: Technically, this check to see if an enabled remote of the same name already exists isn't necessary as the module.AddRemote below will fail and return an error message. But this message is more user friendly. Module.AddRemote returns something like "fatal: remote <name> already exists".

GitCommands/Remotes/GitRemoteManager.cs
```diff
@@ -274,8 +274,26 @@ public GitRemoteSaveResult SaveRemote(GitRemote remote, string remoteName, strin
...
+                {
+                    return new GitRemoteSaveResult($"An active remote named \"{remoteName}\" already exists.", false);
```
amaiorano: Note that we want to return an error result right away, and not fall through below, otherwise the remote will get updated.

GitCommands/Remotes/GitRemoteManager.cs
```diff
@@ -274,8 +274,26 @@ public GitRemoteSaveResult SaveRemote(GitRemote remote, string remoteName, strin
...
+                else if (GetDisabledRemoteNames().FirstOrDefault(r => r == remoteName) != null)
+                {
+                    return new GitRemoteSaveResult($"An inactive remote named \"{remoteName}\" already exists.", false);
+                }
+                else
+                {
+                    output = module.AddRemote(remoteName, remoteUrl);
+
+                    // If output was returned, something went wrong
+                    if (output.Length > 0)
```
amaiorano: If Module.AddRemote succeeds, it returns an empty string; so if it fails, we shouldn't fall through as it did before. This was already broken. You can test it by adding a new remote, but not populating the URL field and clicking Save Changes.

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -369,7 +369,8 @@ private void SaveClick(object sender, EventArgs e)
 
                 if (!string.IsNullOrEmpty(result.UserMessage))
                 {
-                    MessageBox.Show(this, result.UserMessage, _gitMessage.Text);
+                    MessageBox.Show(this, result.UserMessage, _gitMessage.Text, MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
```
amaiorano: This produces a nicer looking messagebox in my opinion, but we can revert if you like.

RussKie:
Good proposal, but how does address the use case above? A user may have an inactive remote, then adds a new remote outside GE, and then attempts to deactivate or reactivate the duplicated remote.

GitCommands/Remotes/GitRemoteManager.cs
```diff
@@ -274,8 +274,26 @@ public GitRemoteSaveResult SaveRemote(GitRemote remote, string remoteName, strin
             bool remoteDisabled = false;
             if (creatingNew)
             {
-                output = module.AddRemote(remoteName, remoteUrl);
-                updateRemoteRequired = true;
+                if (GetEnabledRemoteNames().FirstOrDefault(r => r == remoteName) != null)
+                {
+                    return new GitRemoteSaveResult($"An active remote named \"{remoteName}\" already exists.", false);
+                }
+                else if (GetDisabledRemoteNames().FirstOrDefault(r => r == remoteName) != null)
```
RussKie: These `else`s are redundant since the code above returns.

GitCommands/Remotes/GitRemoteManager.cs
RussKie: Nice!

amaiorano:
> Good proposal, but how does address the use case above? A user may have an inactive remote, then adds a new remote outside GE, and then attempts to deactivate or reactivate the duplicated remote.

Yes, you're right. We could also add similar checks when clicking the "eye" icon to activate/deactivate. Or we can just leave it as is, along with your code to overwrite the existing remote entry.

amaiorano:
I wanted to make the strings I added TranslationStrings:

```csharp
return new GitRemoteSaveResult($"An active remote named \"{remoteName}\" already exists.", false);
...
return new GitRemoteSaveResult($"An inactive remote named \"{remoteName}\" already exists.", false);
```

...but GitCommands doesn't have a reference on ResourceManager. I'm assuming we want to keep it that way? Or can we just add the reference?

amaiorano:
> We need to add unit tests covering the new set of changes

Yes, I can do for GitRemoteManager. Do we have UI integration tests for FormRemote?

RussKie:
> Do we have UI integration tests for FormRemote?

No, but it should be fairly easy to setup. 😎 
Please have a look how I added a new set of UI tests for the left panel in your PR #6100: https://github.com/gitextensions/gitextensions/pull/6100/commits/9fa367634255f6428b18540426ab5d23e4c666f7

GitCommands/Remotes/GitRemoteManager.cs
```diff
@@ -278,22 +278,21 @@ public GitRemoteSaveResult SaveRemote(GitRemote remote, string remoteName, strin
...
-                    // If output was returned, something went wrong
-                    if (output.Length > 0)
-                    {
-                        return new GitRemoteSaveResult(output, false);
-                    }
+                output = module.AddRemote(remoteName, remoteUrl);
 
-                    updateRemoteRequired = true;
+                // If output was returned, something went wrong
+                if (output.Length > 0)
```
RussKie: Wouldn't this be better expressed as `!string.IsNullOrWhitespace(output)`?

RussKie:
Another thought - we can have a method a `GitRemotesManager` that checks whether a remote with a given name exists:
```
bool CheckExists(string remoteName);
```

This could simplify `Save` logic:
a) it could simply overwrite sections as I have currently implemented; or
b) it could throw if there is a dup (a specialised exception, e.g. `DuplicateRemoveException(string remoteName, bool active)`).

I don't have a particular preference to either. Which one do you think is better?

At the UI layer we do a check before attempting to add a new remote, and all UI-related implementations stay at GitUI layer. And GitCommands remain detached from UI concerns.

amaiorano:
> Another thought - we can have a method a `GitRemotesManager` that checks whether a remote with a given name exists:
> 
> ```
> bool CheckExists(string remoteName);
> ```

Yes, I like this much better! Will add two funcs, one for checking if exists as enabled, and another as disabled. This way, I can give the same more specific error messages.

> This could simplify `Save` logic:
> a) it could simply overwrite sections as I have currently implemented; or
> b) it could throw if there is a dup (a specialised exception, e.g. `DuplicateRemoveException(string remoteName, bool active)`).
> I don't have a particular preference to either. Which one do you think is better?

I will leave it as is (option a). This way, as you explained, if someone manipulates remotes outside of GitEx, and then performs activate/deactivate operations in FormRemotes, it will overwrite silently. I think this is a good enough compromise.

> At the UI layer we do a check before attempting to add a new remote, and all UI-related implementations stay at GitUI layer. And GitCommands remain detached from UI concerns.

Yes, that's perfect. PR update forthcoming!

amaiorano:
@RussKie I've made the modifications discussed to move the messaging to the UI layer, and have added unit tests for the two new functions I added to GitRemoteManager. Perhaps we can tackle the FormRemote integration tests separately?

amaiorano:
Hmm, so my change to return a failure state from GitRemoteManager.SaveRemote when module.AddRemote returns a non-empty string is failing a couple of our tests. I believe my change is correct, so I will update the tests.

amaiorano:
Hmm, so now AppVeyor seems to be failing on unrelated tests?

vbjay:
Checked out this branch and ran tests.  Just the win32 api tests fail on a 64 bit box.

image

amaiorano:
> Checked out this branch and ran tests. Just the win32 api tests fail on a 64 bit box.
> 

I don't see how this relates to my change. I suspect you'll have failures on master if you run tests a few times. Just restarting the tests on the same commit passed the second time.

gerhardol:
The AppVeyor builds hangs (killed after an hour) when finishing or reports a number of weird errors lately.
See the logs.
The workaround is to rebuild (can be done by maintainers if logging in) or close/open the PR.
Obviously, this must be stabilized.

RussKie:
All tests build on my machine (c)
Failing tests are likely a byproduct of #6037

RussKie:
Appveyor has been somewhat unstable, the same build succeeds after hanging...

amaiorano:
@RussKie I think the message got lost in the conversation about AppVeyor. This PR should be complete. I didn't add UI tests for FormRemote, but I think we should do that separately.

vbjay:
> @RussKie I think the message got lost in the conversation about AppVeyor. This PR should be complete. I didn't add UI tests for FormRemote, but I think we should do that separately.

Then take WIP off.

vbjay:
@amaiorano 

Oh. Thought it was your pull request.

amaiorano:
@RussKie do you have time to review my latest changes?

RussKie:
I have, thank you.
I have squashed your fixups.

Issue #4737, opened 2018-03-29
Relicense request


sharwell:
I noticed the project is currently licensed under a GPL license. As a contributor, this license works well for me in the following conditions:

1. The contribution is completely my own code, so I can relicense it if I choose, e.g. #4735
2. The contribution is known throwaway code (nothing particularly new or interesting from an algorithms perspective), e.g. #4601 or #4717 

However, this license is strong barrier to progress on novel work. For example, I will be unable to work with other contributors to address some of the outstanding hard problem related to asynchronous tasks that are discussed in topics like #4501, or work with others to build upon the 

In order to proceed as a participant, one of the following needs to occur:

1. GitExtensions needs to be relicensed under a contributor-friendly open source license¹
2. A utility library needs to be created with GitExtensions which is licensed under a contributor-friendly open source license, with new and interested code related to hard problems placed primarily within that library
3. A new repository created for work related to the above, which is licensed under a contributor-friendly open source license

/cc @drewnoakes @jbialobr @gerhardol @RussKie @KindDragon 

¹ The contributor-friendly licenses I would consider contributing under with other users, in order of most desirable to least, are MIT (with or without an explicit patent grant), Apache 2.0, or BSD (2- or 3-clause).

gerhardol:
I prefer 3

GPL is contributor friendly in the sense that it is not permitted to steal the code for other projects.

RussKie:
I acknowledge the request. In principal I am not opposed to the motion.
I will need to spend some time studying different licensing options.

On 30 Mar 2018 9:57 am, "Gerhard Olsson" <notifications@github.com> wrote:

I prefer 3

GPL is contributor friendly in the sense that it is not permitted to steal
the code for other projects.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/gitextensions/gitextensions/issues/4737#issuecomment-377397434>,
or mute the thread

drewnoakes:
Thanks for bringing this up @sharwell. I would also be happy to see the project under a less viral license. GPL gives me some caution when it comes to contributing.

With so many contributors to the project to date it will probably be very difficult to change the license however.

Does option 3 mean a submodule is not an option? Would it have to be a binary reference? That would suck from a practicality standpoint.

sharwell:
[...]

PR #6124, opened 2019-01-18
Update PULL_REQUEST_TEMPLATE.md

RussKie:
Include DCO ~and release under MIT~

Relates to #4737.

Further down the track we can consider enforcing sign-off with a [Probot](https://github.com/probot/dco).

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -1,0 +1,55 @@
+Git Extensions Project Contributors Certification of Origin and Rights
+
+All contributors to Git Extensions must formally agree to abide by this
+certificate of origin by signing on the bottom with their github
+userid, full name, email address (you can obscure your e-mail, but it
+must be computable by human), and date.
+
+By signing this agreement, you are warranting and representing that
+you have the right to release code contributions or other content free
+of any obligations to third parties and are granting Henk Westhuis and
+Git Extensions project contributors, henceforth referred to as 
+The Git Extensions Project, a license to incorporate it into 
+The Git Extensions Project or related works under the MIT license. 
+You understand that The Git Extensions Project may or may not incorporate
+your contribution and you warrant and represent the following:
+
+1. I am the creator of all my contributions. I am the author of all
+   contributed work submitted and further warrant and represent that
+   such work is my original creation and I have the right to license
+   it to The Git Extensions Project for release under the MIT and/or GPL
+   license. I hereby grant The Git Extensions Project a nonexclusive,
+   irrevocable, royalty-free, worldwide license to reproduce,
+   distribute, prepare derivative works, and otherwise use my
+   contributions as part of the Git Extensions project, associated
+   documentation, books, and tools at no cost to The Git Extensions Project.
+
+2. I have the right to submit. This submission does not violate the
+   rights of any person or entity and that I have legal authority over
+   this submission and to make this certification.
+
+3. If I violate another's rights, liability lies with me. I agree to
+   defend, indemnify, and hold The Git Extensions Project and Git Extensions 
+   users harmless from any claim or demand, including reasonable attorney
+   fees, made by any third party due to or arising out of my violation
+   of these terms and conditions or my violation of the rights of
+   another person or entity.
+
+4. I understand and agree that this project and the contribution are
+   public and that a record of the contribution (including all
+   personal information I submit with it, including my sign-off) is
+   maintained indefinitely and may be redistributed consistent with
+   this project or the open source license indicated in the file.
+
+I have read this agreement and do so certify by adding my signoff to
+the end of the following contributors list.
+
+CONTRIBUTORS:
+
+YYYY/MM/DD, github id, Full name, email
+2019/01/23, spdr870, Henk Westhuis, henk.westhuis@gmail.com
+2019/01/23, RussKie, Igor Velikorossov, russkie@gmail.com
+2019/01/23, gerhardol, Gerhard Olsson, gerhard.nospam@gmail.com
+2019/01/23, drewnoakes, Drew Noakes, git@drewnoakes.com
+2019/01/23, amaiorano, Antonio Maiorano, amaiorano@gmail.com
+2019/01/23, sharwell, Sam Harwell, sam@tunnelvisionlabs.com
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -34,3 +34,8 @@ Fixes #
 - Windows <!-- Add version 7 SP1 or above -->
 
 <!-- Mention language, UI scaling, or anything else that might be relevant -->
+
+
+----
+
+:black_nib: I contribute this code under [The Developer Certificate of Origin](../../blob/master/contributors.txt).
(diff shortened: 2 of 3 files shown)


amaiorano:
OMG sorry about the review approved spam! Last time I use my mobile browser to do this 😕

amaiorano:
> Are you SURE you approve? Just making sure. 😆

Apparently I've never approved of anything more in my life! :trollface:

RussKie:
@spdr870  @drewnoakes @sharwell waiting for your approvals

sharwell:
I'm neither for nor against this change. It would be good to update Contributing.md at the same time.

RussKie:
Slight update to include MIT licence and CONTRIBUTING.md

The PR template will have the following footer:
image

CONTRIBUTING.md will have the following section added:
image

I'm planning to merge this shortly.

vbjay:
Does this require sign offs?  If so, then I suggest we add the probot first.

RussKie:
No, no sign off for now. The presence of the above text should be sufficient.
No PRs without the above text should be merged in.

vbjay:
Also, should the license be replaced and not have dual licenses?

RussKie:
I'm not a lawyer, but I don't believe we are in a position to remove GPL licence, as the code is still licensed under it until we receive releases and waivers from all contributors past.

gerhardol:
It is possible to have dual licenses, but that should not be necessary here when switched.
However, in the meanwhile, contributors must accept to contribute under both MIT and GPL. I believe it is possible to re-licence MIT -> GPL (if MIT copyright notice is kept at least), so the wording should be OK.

All forks done before license is switched will be using GPL and it should be possible to import code that allows GPL (so code changes, maybe not external libraries).

drewnoakes:
Apologies all for missing this. @RussKie pinged me offline and I'm just seeing it now.

I don't want to block this, but does it make sense to include two incompatible licenses in the project? If I didn't know the history and stumbled upon them I might think it was dual licensed, which it isn't.

I approve of moving away from GPL and MIT seems fine to me. I wonder if getting a CLA in place makes more sense. I'm not a laywer obviously, but I believe a CLA could grant the maintainers the right to relicense contributions at some later point. That would allow changes from this point forward to be relicensed at the future point where we also have legal grounds to relicense historic commits.

sharwell:
@RussKie Easiest way to start with would be to use the approach of ANTLR:
https://github.com/antlr/antlr4/blob/master/contributors.txt

That would transfer forward to the CLA if the project joins the .NET Foundation.

RussKie:
Awesome, thanks Same

RussKie:
Take 3, I have updated the proposal following the ATLR's way.
I have taken my liberty and added all of you to the contributors.txt. 

Please let me know if you have any objections.

contributors.txt
```diff
@@ -0,0 +1,57 @@
...
+certificate of origin by signing on the bottom with their github
+userid, full name, email address (you can obscure your e-mail, but it
+must be computable by human), and date.
+
+By signing this agreement, you are warranting and representing that
+you have the right to release code contributions or other content free
+of any obligations to third parties and are granting Henk Westhuis and
+GitExtensions project contributors, henceforth referred to as The GitExtensions
+Project, a license to incorporate it into The GitExtensions Project 
+or related works under the BSD license. You understand that The GitExtensions
```
sharwell:
```suggestion
or related works under the MIT license. You understand that The GitExtensions
```

contributors.txt
```diff
@@ -0,0 +1,57 @@
+GitExtensions Project Contributors Certification of Origin and Rights
```
sharwell: 💭 The project isn't especially consistent on "GitExtensions" vs. "Git Extensions". I assume this is the preferred form?

contributors.txt
```diff
@@ -0,0 +1,57 @@
...
+GitExtensions project contributors, henceforth referred to as The GitExtensions
+Project, a license to incorporate it into The GitExtensions Project 
+or related works under the BSD license. You understand that The GitExtensions
+Project may or may not incorporate your contribution and you warrant 
+and represent the following:
+
+1. I am the creator of all my contributions. I am the author of all
+   contributed work submitted and further warrant and represent that
+   such work is my original creation and I have the right to license
+   it to The GitExtensions Project for release under the 3-clause BSD
```
sharwell:
```suggestion
   it to The GitExtensions Project for release under the MIT and/or GPL
```

This will simplify to only state MIT once we complete the history review.

contributors.txt
```diff
@@ -0,0 +1,57 @@
...
+
+YYYY/MM/DD, github id, Full name, email
+2019/01/23, spdr870, Henk Westhuis, henk.westhuis@gmail.com
+2019/01/23, KindDragon, Arkady Shapkin, arkady.shapkin@gmail.com
+2019/01/23, jbialobr, Janusz Białobrzewski, jbialobr@o2.pl
+2019/01/23, RussKie, Igor Velikorossov, russkie@gmail.com
+2019/01/23, gerhardol, Gerhard Olsson, gerhard.nospam@gmail.com
+2019/01/23, drewnoakes, Drew Noakes, git@drewnoakes.com
+2019/01/23, amaiorano, Antonio Maiorano, amaiorano@gmail.com
+2019/01/23, sharwell, Sam Harwell, sam@tunnelvisionlabs.com
```
sharwell: 📝 We should probably all sign this individually

drewnoakes: At the very least everyone should re-accept this PR post this change.

vbjay:
@amaiorano  Only one approval this time?  :grinning:

amaiorano:
> @amaiorano Only one approval this time? 😀

Feeling less strongly about this one ;)

contributors.txt
RussKie: Yes, that was my plan.
I hope you're all ok with it.

contributors.txt
RussKie: Everywhere we use "Git Extensions", I'll fix this one up
Thank you

contributors.txt
```diff
@@ -0,0 +1,57 @@
...
+your contribution and you warrant and represent the following:
+
+1. I am the creator of all my contributions. I am the author of all
+   contributed work submitted and further warrant and represent that
+   such work is my original creation and I have the right to license
+   it to The Git Extensions Project for release under the MIT and/or GPL
+   license. I hereby grant The Git Extensions Project a nonexclusive,
+   irrevocable, royalty-free, worldwide license to reproduce,
+   distribute, prepare derivative works, and otherwise use this
+   contribution as part of the Git Extensions project, associated
```
sharwell: 💡 ... use ~~this contribution~~ **my contributions** as part ...

RussKie: Thank you

RussKie:
To expedite the merger, I removed @jbialobr and @KindDragon from the list as I haven't received a re-approval from them. I will work with them to get their sign off.

vbjay:
What about submodules?

drewnoakes:
> What about submodules?

I guess that's less important as they're not GPL (afaik)...

Issue #4775, opened 2018-04-04
High DPI Issues


drewnoakes:
# High DPI Issues

Issues by form/dialog:

- [ ] Remote repositories (default behaviours tab) (#5028)
- [ ] Recent repositories settings (#4774)
- [x] ~Settings: Colors~ (PR part of #5081)
- [x] ~Settings: Confirmations~ (PR #5560)
- [x] ~Settings: Shell extension~ (PR #5561)
- [x] ~Delete tag~ (#5300, #5380, PR #5381)
- [x] ~About dialog~ (#5201, PR #5248)
- [x] ~Commit template settings form~ (#5059)
- [x] ~FormCommit: Text editor markers~ (#5024, PR #5215)
- [x] ~FormCommit: TextBox watermark~ (#5023, PR #5087)
- [x] ~GravatarControl image scaling~ (PR #4711)
- [x] ~Settings: treeview navigation~ (#4949, PR #4950)
- [x] ~Settings: Change HOME~ (#4690, PR #4755, v2.51 #4780)
- [x] ~Remote repositories~ (PR #4714, v2.51 6d49acd2de4ae719b87c7e369a24c085dbe71ffd)
- [x] ~Revision graph / grid~ (PR #4772)
- [x] ~Open local repository~ (#4174, PR #4887, v2.51 #4893)
- [x] ~Clean working directory~ (#3828, PR #4890, v2.51 #4891)
- [x] ~Clone repository~ (#4776, PR #4888, v2.51 7dfe467752d02a8d030961ae03d65b3586ee767b)
- [x] ~Commit picker control~ (#4702 #4862, PR #4889, v2.51 268fa27d759241aac59e3045268b9fb05a31abe9)
- [x] ~NBug scaling issues~ (#4989, PR gitextensions/NBug#10)

Issues by control:

- [ ] Splitter positions
- [x] ~Tab control header is too small~ (#4777)
- [x] ~Columns aren't sized correctly~ (#3603, PR #4772)
- [x] ~Images in buttons are too small~ (#4778, PR #4772)
- [x] ~Tree view icons are too small~ (PR #4772)
- [x] ~Data grid columns are too sma~l~ (PR #4772)

Longer term goals:

- [ ] High DPI icons (#4825)
- [ ] Per-monitor DPI awareness (#4822)

---

# Reporting new issues

[...]

PR #6125, opened 2019-01-18
Use high DPI mouse cursors

drewnoakes:
Relates to #4775.

## Proposed changes

- Replace some 96-dpi .cur files provided by the framework with high-DPI system cursors

Full disclaimer: this uses private reflection, which is generally a no-no, but in doing so it catches any exception and continues with no negative effects that I can see.

An equivalent change is being made to dotnet/winforms, but until we move to .NET Core we are stuck with this approach.

### Before

image

image

### After

image

image

## Test methodology 

- Manual testing

## Test environment(s) 

- Windows 10
- UI Scaling: 150%


RussKie:
👍

PR #6129, opened 2019-01-18
Fix CodeCov coverage - exclude unit test files

wachulski:
Fixes Unit test coverage calculated by CodeCov. It excludes `/UnitTests/` sources recursively.

**Before**: https://codecov.io/gh/gitextensions/gitextensions (44.61% overall)

**Now**: https://codecov.io/gh/wachulski/gitextensions/tree/06a679fa84b0adc4a48ab763e5b7af2d6a4a206f (34.05% overall)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -71,13 +71,14 @@ test_script:
         -filter:"+[*]* -[FluentAssertions*]* -[SmartFormat*]* -[nunit*]*" `
         -excludebyattribute:*.ExcludeFromCodeCoverage* `
         -excludebyfile:*\*Designer.cs `
-        -output:"OpenCover.GitExtensions.xml" `
+        -excludedirs:UnitTests\ `
+        -output:"OpenCover.GitExtensions.ProdOnly.xml" `
         -target:"nunit3-console.exe" `
         -targetargs:"$testAssemblies"
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
     $codecov_version = $packageConfig.SelectSingleNode('/packages/package[@id="Codecov"]').version
     $codecov = "packages\Codecov.$codecov_version\tools\codecov.exe"
-    &$codecov -f ".\OpenCover.GitExtensions.xml"
+    &$codecov -f ".\OpenCover.GitExtensions.ProdOnly.xml" --flag production
 
 
 # here we are going to override common configuration
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -71,14 +71,14 @@ test_script:
         -filter:"+[*]* -[FluentAssertions*]* -[SmartFormat*]* -[nunit*]*" `
         -excludebyattribute:*.ExcludeFromCodeCoverage* `
         -excludebyfile:*\*Designer.cs `
-        -excludedirs:UnitTests\ `
-        -output:"OpenCover.GitExtensions.ProdOnly.xml" `
+        -output:"OpenCover.GitExtensions.xml" `
         -target:"nunit3-console.exe" `
         -targetargs:"$testAssemblies"
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
     $codecov_version = $packageConfig.SelectSingleNode('/packages/package[@id="Codecov"]').version
     $codecov = "packages\Codecov.$codecov_version\tools\codecov.exe"
-    &$codecov -f ".\OpenCover.GitExtensions.ProdOnly.xml" --flag production
+    &$codecov -f ".\OpenCover.GitExtensions.xml" --flag production
+    &$codecov -f ".\OpenCover.GitExtensions.xml" --flag tests
 
 
 # here we are going to override common configuration
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,7 +9,7 @@ coverage:
       default: false
 
 comment:
-  layout: "diff"
+  layout: "diff,flags"
 
 flags:
   production:
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,3 +10,16 @@ coverage:
 
 comment:
   layout: "diff"
+
+flags:
+  production:
+    paths:
+      - GitCommands/
+      - GitExtUtils/
+      - GitUI/
+      - NetSpell.SpellChecker/
+      - Plugins/
+      - ResourceManager/
+  tests:
+    paths:
+      - UnitTests/


sharwell:
Alternatives are available if you are trying to accomplish something specific. Please let me know if you have questions.

vbjay:
The change uses test coverage but excludes the test files in line counts. It gives a better view of the coverage.

wachulski:
> Unit test coverage is one of the primary values of code coverage automation. It should not be excluded.

@sharwell You mean auxiliary means that setup/arrange/assert for actual unit tests. I kind of agree, that some people would like to have it included in the coverage. Then you can see e.g. which lines of unit test automation are run in unit test execution. But I'm not sure if it's what you meant. Could you confirm. And what do others think?

wachulski:
IMO the most painful effect of having unit tests included is code coverage on new lines reported by CodeCov. Consider this: if I added a new feature (100 lines) and accompanying unit tests (300 lines) that actually cover 50% of new feature then 350 lines are executed of 400 added in total. IMO should be 50 of 100 total. That's particularly significant for PRs and CodeCov being a gatekeeper.

sharwell:
> The change uses test coverage but excludes the test files in line counts.

This does not appear to be correct. Consider the views:

* Before this change: https://codecov.io/gh/gitextensions/gitextensions/tree/4695d0e12c5ba1697b7901eed0e82691e18a7517
* With this change: https://codecov.io/gh/gitextensions/gitextensions/tree/06a679fa84b0adc4a48ab763e5b7af2d6a4a206f

> IMO the most painful effect of having unit tests included is code coverage on new lines reported by CodeCov. Consider this: if I added a new feature (100 lines) and accompanying unit tests (300 lines) that actually cover 50% of new feature then 350 lines are executed of 400 added in total. IMO should be 50 of 100 total.

Codecov can split these out using the flags feature so coverage numbers for production and test code are reported separately.

> You mean auxiliary means that setup/arrange/assert for actual unit tests? I kind of agree, that some people would like to have it included in the coverage. Then you can see e.g. which lines of unit test automation are run in unit test execution. But I'm not sure if it's what you meant. Could you confirm? And what do others think?

Code coverage tooling can highlight cases where a test failure occurs for a reason that wasn't expected. For example, if an exception is expected from a call but a previous line of the same test throws an exception, the test passes but not for the reason intended to be covered by the test. I've seen various forms of this caught over time. It's not an all-the-time type thing, but it's enough that it's been a help.

> ... and CodeCov being a gatekeeper.

Codecov is not a gatekeeper. That's been disabled here:

https://github.com/gitextensions/gitextensions/blob/4695d0e12c5ba1697b7901eed0e82691e18a7517/codecov.yml#L5-L9

wachulski:
@sharwell Well, you're simply right! Especially such tricky cases, when covered by automation, that's life saver. And thanks for pointing out CodeCov's flags feature https://docs.codecov.io/docs/flags So, do you also think by introducing the flag:
```yaml
tests:
    paths:
      - UnitTests/
```
the reports would be a bit more explicit? E.g.
code-cov-flags

RussKie:
I was going to suggest to report both metrics - including and excluding
tests.

wachulski:
Well, the *flags* feature needs a separate results upload (marked with a flag) per flag category. And even if I did such (excluding UnitTest/ directory in OpenCover), then the report was broken on CodeCov side. 

Since it's clear we want all the following to be covered:

* production code
* test utils (to detect which are not used)
* unit tests themselves (for the reason pointed out by @sharwell)

I don't see any quick and compelling way to restructure the coverage output. => closing this PR

The last idea is that the coverage doesn't reflect some portions of sources (e.g. TranslationApp etc.) I know, adding such would decrease the coverage numbers, which doesn't play that important role here https://github.com/gitextensions/gitextensions/pull/6129#issuecomment-455709410. On the other, maybe it should be direct and evident for transparency reasons? Frankly, don't know.

gerhardol:
I rather see only application (excludidin UnitTests, TranslationApp, Setup, Externals, GitExtensionsDoc) than all if that is the options.

sharwell:
Ask me about this again on 22 January in Gitter. There is one other option that may become available that this issue will help drive to conclusion.

RussKie:
@sharwell is there a logical way for the PR to go forward?

appveyor.yml
```diff
@@ -71,13 +71,14 @@ test_script:
         -filter:"+[*]* -[FluentAssertions*]* -[SmartFormat*]* -[nunit*]*" `
         -excludebyattribute:*.ExcludeFromCodeCoverage* `
         -excludebyfile:*\*Designer.cs `
-        -output:"OpenCover.GitExtensions.xml" `
+        -excludedirs:UnitTests\ `
```
sharwell:
```suggestion
```

appveyor.yml
```diff
@@ -71,13 +71,14 @@ test_script:
...
+        -output:"OpenCover.GitExtensions.ProdOnly.xml" `
         -target:"nunit3-console.exe" `
         -targetargs:"$testAssemblies"
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
     $codecov_version = $packageConfig.SelectSingleNode('/packages/package[@id="Codecov"]').version
     $codecov = "packages\Codecov.$codecov_version\tools\codecov.exe"
-    &$codecov -f ".\OpenCover.GitExtensions.xml"
+    &$codecov -f ".\OpenCover.GitExtensions.ProdOnly.xml" --flag production
```
sharwell:
```suggestion
    &$codecov -f ".\OpenCover.GitExtensions.xml" --flag production
    &$codecov -f ".\OpenCover.GitExtensions.xml" --flag tests
```

appveyor.yml
```diff
@@ -71,13 +71,14 @@ test_script:
         -filter:"+[*]* -[FluentAssertions*]* -[SmartFormat*]* -[nunit*]*" `
         -excludebyattribute:*.ExcludeFromCodeCoverage* `
         -excludebyfile:*\*Designer.cs `
-        -output:"OpenCover.GitExtensions.xml" `
+        -excludedirs:UnitTests\ `
+        -output:"OpenCover.GitExtensions.ProdOnly.xml" `
```
sharwell:
```suggestion
        -output:"OpenCover.GitExtensions.xml" `
```

codecov.yml
```diff
@@ -10,3 +10,16 @@ coverage:
 
 comment:
   layout: "diff"
```
sharwell:
```suggestion
  layout: "diff,flags"
```

sharwell:
> @sharwell is there a logical way for the PR to go forward?

The somewhat ugly workaround involves the changes mentioned above:

1. Revert the filtering of the coverage file itself (the uploaded XML contains all coverage data).
2. Upload the file twice using different flags. The first will be filtered to one set of paths, the second will be filtered to the other.
3. Add `flags` to the PR comment so you can see coverage changes in production and test code separately.

In addition, you can modify your code coverage badge in the README to show the numbers from just production code by using `badge.svg?flag=production`.

RussKie:
Thank you

@wachulski over to you.
Please also send your sign off to contributors.txt in a separate PR.

wachulski:
sure, getting back to it in the next couple of days, thx for help @sharwell

appveyor.yml
wachulski: done

appveyor.yml
wachulski: done

appveyor.yml
wachulski: done

codecov.yml
wachulski: done

sharwell:
:memo: The flags table will not show complete data for this pull request because the **master** branch it compares against does not contain flag data. After this is merged, pull requests will start to show data here.

The README badge can also be updated if you want.

wachulski:
I don't have any strong opinion against (non)updating the badge. Personally I'd show production code only in terms of overall stats. The test flag/set serves PRs IMO.

RussKie:
Thank you

RussKie:
Apologies @gerhardol, I completely missed that you wanted to review it.

gerhardol:
> Apologies @gerhardol, I completely missed that you wanted to review it.

Approved

Issue #2452, opened 2014-09-10
Blame opens wrong file


brandf:
--- GitExtensions2.48, Git 1.9.4.msysgit.1, Microsoft Windows NT 6.2.9200.0

I have two files in a directory:
MenuViewHost.js
MainMenuViewHost.js

When I right click on MenuViewHost.js and choose Blame it opens MainMenuViewHost.js instead of MenuViewHost.js.

The title says "FileHistory - [omitted path]MenuViewHost.js ([omitted path]MainMenuViewHost.js) - [omitted path]".  So it knows something is wonky, but it's hard to imagine this is a feature.

alistairmgreen:
I have just hit a similar problem, and I think it's caused by similarity of *content.*

* I have two files - let's call them A and B.
* The content of B is identical to one of the previous revisions of A.
* When I open up the history or blame view of B, it shows me the history and current contents of A.
* At the command line, 'git log' and 'git blame' give me the correct information.

Unfortunately, this is in my employer's private repository so I can't share it with you. The history was quite convoluted:
* A was created first and went through several revisions.
* A was renamed to B.
* In the next commit, somebody copied B back to A by mistake. At this point, both files existed and they were identical.
* All subsequent work was done on A.

vbjay:
Look at the follow renames setting.

alistairmgreen:
Yes, "follow renames" is turned on (this appears to be the default). However, it is behaving in a surprising way.

I have created an [example repository which reproduces the issue.](https://github.com/alistairmgreen/git-extensions-history-wrong-file)

It appears that the rename detection in Git Extensions is more eager than the detection performed by `git log --follow`.

At the command line, Git looks backwards in time and shows the history of the file specified (noting that `exampleb.txt` is descended from the first version of `examplea.txt`).

However, Git Extensions looks both forwards and backwards in time. When I open the history dialog for `exampleb.txt`, the first thing it shows is a "rename" of `exampleb.txt` to `examplea.txt` with some new content inserted. At first glance the history of B looks exactly like the history of A, even though the files are now different.

This means that Git Extensions is showing *the most recent content descended from the content of the file that I clicked on.* Unfortunately, that version is *newer* than the one that I clicked on and the result is a bit confusing.

loop-evgeny:
[...]

PR #6132, opened 2019-01-19
WIP Fix/rename file history

### LINKED ISSUE #5944: History no longer shown for renamed files
Bug

History is not shown for files before their rename (but blame correctly identifies commits)

Take a file with history, rename it, perform additional commits after changing it.

History is shown for the file before it got renamed (as it currently works for Blame)

- Git Extensions 3.00.00.4433
- Git 2.20.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

(In settings "Follow renames in History" is enabled, "Follow exact renames and copies only" has been tried as both disabled and enabled - the files have 100% similarity".)

I recall the history being shown correctly for the renamed files in Version 2.5x

vbjay:
Fixes #5944

## Proposed changes

- Refactor FileHistory to use ArgumentString
- Make Renamed file history work again
- ~~Maybe make folder history show relevant ui~~ 

## Test methodology 
Just starting
- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

VERY WIP  DO NOT even think of reviewing for style or why something is somewhere.

### DIFF
--- a/GitCommands/Git/GitRevision.cs
+++ b/GitCommands/Git/GitRevision.cs
@@ -98,7 +98,7 @@ namespace GitCommands
         /// Used to capture the other output from log calls.  Ex: getting commit history from a file using --follow and --name-only gives us the filename.
         /// </summary>
         [CanBeNull]
-        public string OtherOutput { get; set; }
+        public string AdditionalOutput { get; set; }
 
         public override string ToString() => $"{ObjectId.ToShortString(8)}:{Subject}";
 
--- a/GitCommands/Git/GitRevision.cs
+++ b/GitCommands/Git/GitRevision.cs
@@ -94,6 +94,12 @@ namespace GitCommands
 
         public string Name { get; set; }
 
+        /// <summary>
+        /// Used to capture the other output from log calls.  Ex: getting commit history from a file using --follow and --name-only gives us the filename.
+        /// </summary>
+        [CanBeNull]
+        public string OtherOutput { get; set; }
+
         public override string ToString() => $"{ObjectId.ToShortString(8)}:{Subject}";
 
         /// <summary>
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -28,6 +28,7 @@ namespace GitCommands
 
     public sealed class RevisionReader : IDisposable
     {
+        private const string EndOfBody = "\x2"; // STX control char.  Nobody will ever type character 2 in the body.
         private const string FullFormat =
 
               // These header entries can all be decoded from the bytes directly.
@@ -47,7 +48,7 @@ namespace GitCommands
               /* Committer name  */ "%cN%n" +
               /* Committer email */ "%cE%n" +
               /* Commit subject  */ "%s%n%n" +
-              /* Commit body     */ "%b";
+              /* Commit body     */ "%b" + EndOfBody;
 
         private readonly CancellationTokenSequence _cancellationTokenSequence = new CancellationTokenSequence();
 
@@ -415,6 +416,14 @@ namespace GitCommands
                 return false;
             }
 
+            int bodyEnd = body.IndexOf(EndOfBody);
+            string otherOutput = null;
+            if (bodyEnd > -1)
+            {
+                var otherStartIndex = bodyEnd + EndOfBody.Length;
+                otherOutput = body.Length > otherStartIndex ? body.Substring(otherStartIndex) : null;
+                body = body.Substring(0, bodyEnd);
+            }
             #endregion
 
             revision = new GitRevision(objectId)
@@ -431,7 +440,8 @@ namespace GitCommands
                 Subject = subject,
                 Body = body,
                 HasMultiLineMessage = !ReferenceEquals(subject, body),
-                HasNotes = false
+                HasNotes = false,
+                OtherOutput = otherOutput
             };
 
             return true;
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -441,7 +441,7 @@ namespace GitCommands
                 Body = body,
                 HasMultiLineMessage = !ReferenceEquals(subject, body),
                 HasNotes = false,
-                OtherOutput = otherOutput
+                AdditionalOutput = otherOutput
             };
 
             return true;
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -317,6 +317,9 @@ namespace GitUI.CommandsDialogs
                 fileName = FileName;
             }
 
+            var tree = Module.GetTree(revision.ObjectId, false, fileName).First() as GitCommands.Git.GitItem;
+            bool isDirectory = tree.ObjectType == GitCommands.Git.GitObjectType.Tree;
+
             SetTitle(fileName);
 
             if (revision.IsArtificial)
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -197,7 +197,7 @@ namespace GitUI.CommandsDialogs
 
             return;
 
-            (string revision, string path) BuildFilter()
+            (ArgumentString revision, string path) BuildFilter()
             {
                 var fileName = FileName;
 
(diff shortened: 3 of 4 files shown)


RussKie:
It may be Appveyor playing up. I had a PR that hung after I added UI tests, but on a second rebuild it build correctly....

vbjay:
@spdr870 Any thoughts on this?

image


They are not linked commits.  They are commits where x file was changed.  Is this view ok or is there something we can do to prevent the graph from stretching like it does now?
```bash
git -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%e%n%aN%n%aE%n%cN%n%cE%n%s%n%n%b" --all --boundary --max-count=100000 --name-only --parents --follow  --find-renames --find-copies --full-history -- "GitUI/BrowseForPrivateKey.cs"
```

gerhardol:
@vbjay 
Submodule history behaves the same, maybe the same solution

vbjay:
@gerhardol Do you mean

image

or
image

gerhardol:
@vbjay 
2nd, the revisiongraph. It should be possible to present this repo branches for the submodule in a better wayy. (Submodule history could possibly be removed.)

vbjay:
Seeing if I can figure out #6139  @drewnoakes any ideas would be appreciated on that.

GitCommands/RevisionReader.cs
```diff
@@ -28,6 +28,7 @@ public enum RefFilterOptions
 
     public sealed class RevisionReader : IDisposable
     {
+        private const string EndOfBody = "\x2"; // STX control char.  Nobody will ever type character 2 in the body.
```
drewnoakes: Hmm. I don't think you can make that assumption unfortunately. It seems reasonable for ASCII, but in UTF-8 or other encodings it's more likely to pop up as part of a multi-byte character.

vbjay:
Run it. It does the conversion. It most certainly does work.

GitCommands/RevisionReader.cs
drewnoakes: > Run it. It does the conversion. It most certainly does work.

I'm sure it works on the GE repo, but for users who write commit messages in non Latin languages, `\x2` is likely to pop up and cause this to fail.

It took me a while to conclude that a null byte was safe. Because `\0` is special in C-style strings, it's avoided in encodings (at least as far as I could determine). That's not the case for `x2`, so we need a different solution. Thinking...

vbjay: Oh crap. Dang it. Good catch.

drewnoakes:
What if the chunking of the byte stream into commits was delimited by more than one consecutive `\0`?

I tried a few combos and I think this would work:

```
git -c log.showSignature=false log master -z --pretty=format:"%x00%x00%H%T%P%n%at%n%ct %n%e%n%aN%n%aE%n%cN%n%cE%n%s%n%n%b%x00" --name-only -n3 --no-merges
```

I believe it's necessary to start with a few nulls, and the parser will need to handle these for the first one. The last item will also end with only a single null. But I think there's enough information there for the parser to work correctly. Let me know what you think.

If this works we may be able to use it to grab other info down too.

vbjay:
:clap:   I get it.  That will work.  We will always just handle double nulls as the delimiter.  Thanks a ton.

vbjay:
> Would `Span<T>` be of any help perf-wise?

Great idea.  I think just getting the parser to read the full data first and then I am more than willing to try that as a separate commit to test with.

drewnoakes:
> If there are nulls at the beginning we can just trim them.

That's fine so long as you don't add any extra byte copying for the trim. The concept sounds good though.

vbjay:
> > If there are nulls at the beginning we can just trim them.
> 
> That's fine so long as you don't add any extra byte copying for the trim. The concept sounds good though.

Nope.  Just a simple adjustment of the start index to the first non null byte index before yielding the chunk.

GitCommands/Git/GitRevision.cs
```diff
@@ -94,6 +94,12 @@ public BuildInfo BuildStatus
 
         public string Name { get; set; }
 
+        /// <summary>
+        /// Used to capture the other output from log calls.  Ex: getting commit history from a file using --follow and --name-only gives us the filename.
+        /// </summary>
+        [CanBeNull]
+        public string AdditionalOutput { get; set; }
```
fschmied: Apparently, the `Name` property is meant to hold the filename, so is `AdditionalOutput` really required? You can check out https://github.com/gitextensions/gitextensions/blob/release/2.51/GitCommands/RevisionGraph.cs#L403 and https://github.com/gitextensions/gitextensions/blob/release/2.51/GitCommands/RevisionGraph.cs#L321 to see how it was used before one of the refactorings that killed the `--follow` feature.

fschmied: (I guess if you leave the `AdditionalOutput`, you should probably remove the `Name` property, as it's no longer assigned.)

GitCommands/RevisionReader.cs
fschmied: You could use a Guid instead... :)

GitUI/CommandsDialogs/FormFileHistory.cs
```diff
@@ -233,64 +246,33 @@ private void LoadFileHistory()
 
                 FileName = fileName;
 
-                var res = (revision: (string)null, path: $" \"{fileName}\"");
+                var res = (revision: (ArgumentBuilder)new ArgumentBuilder(), path: fileName);
 
-                if (AppSettings.FollowRenamesInFileHistory && !Directory.Exists(fullFilePath))
+                if (AppSettings.FollowRenamesInFileHistory)
                 {
-                    // git log --follow is not working as expected (see  http://kerneltrap.org/mailarchive/git/2009/1/30/4856404/thread)
```
fschmied: You're effectively removing this workaround, right? Are you sure it's no longer required? (The linked Git mailing list post can be found [here](https://www.spinics.net/lists/git/msg92132.html) as the original URL seems to be broken. Thread topic for googling: "Git log can not show history before rename".)

GitCommands/Git/GitRevision.cs
vbjay: I named it AdditionalOutput because that is exactly what that is.  The AdditionalOutput is then processed based on what output it is.  In this case name will eventually be filled in but Other commands could do this to.  So I will be leaving name and additional output so that it leaves a spot for all commands to process the additional output after the null.

GitCommands/RevisionReader.cs
vbjay: I don't want a crazy long value the nulls are the better solution.

GitUI/CommandsDialogs/FormFileHistory.cs
vbjay: Why use a workaround when the problem the workaround solves goes away?

fschmied: The workaround was added in https://github.com/gitextensions/gitextensions/commit/e1b029d207624c26c1ed0143b772afc87f10f683 with this message:

> Directly using --follow can hide some commits relevant (even if the file was never renamed) - this approach might show more commits than needed.

It was changed later to resolve #2452. I can't really tell if these issues have gone away :)

GitCommands/Git/GitRevision.cs
fschmied: Good point. Then probably just remove the `GitRevision.Name` property as it's never assigned any more?

GitCommands/RevisionReader.cs
fschmied: Sure, but if I read the code correctly, the nulls are currently used to separate the individual revisions. This makes the parser much simpler, but the corresponding refactoring (https://github.com/gitextensions/gitextensions/commit/1f6b0cacc8bc78be076b4e706e5aa030b6d36ea0) broke (part of) rename following in the first place.

I guess you could revert back to a similar parsing logic like before and use a ["(__BEGIN_COMMIT__)>" marker](https://github.com/gitextensions/gitextensions/commit/1f6b0cacc8bc78be076b4e706e5aa030b6d36ea0#diff-746e3f7b1adff7195f50bea04896ccc2L34)... I'd find it simpler, though, to use just separate commit body and file name by a pattern extremely unlikely to occur within a commit. Guid was only an example, maybe a binary pattern such as `de ad be ef` or something instead?

(About the length: the commits are already quite long anyway, I don't think it would matter to add a separator string the length of a Guid. You could also binary-encode the Guid to make it shorter.)

RussKie: Please consider how we can test the changes

vbjay: Most definitely.

GitCommands/Git/GitRevision.cs
vbjay: I don't think you get the idea. AditionalOutput is the basically raw output.  This is low level code. You can't assume it's always filenames. I named it to be quite clear on what it was. Other commands may create additional output that is not filenames or other info. So eventually in this case name will be set up but I can't 100% say additional output is always filenames. So I will not be removing name or AditionalOutput.

GitUI/CommandsDialogs/FormFileHistory.cs
vbjay: It will. It is because of the parsing that we were only getting first file name.

GitCommands/Git/GitRevision.cs
vbjay: Look. This is wip I haven't finished. I need parsing done first.

RussKie: Maybe outline what the plan and the vision are, it could help in understanding and remove some questions

fschmied: > I don't think you get the idea

I think I do get the idea and, as I said, I think you have a good point.

However. `GitRevision.Name` used to contain the filename back in 2.50, when rename history still worked. It is now currently always `null`, due to [one of the two refactorings that broke rename file history](https://github.com/gitextensions/gitextensions/commit/1f6b0cacc8bc78be076b4e706e5aa030b6d36ea0). Your fix leaves it at `null` and introduces `AdditionalInfo` instead.

The property is still read in over 100 places, though, which might still be broken. That's why I suggested restoring the old behavior and using `Name` even though it's not always the filename. And if you don't want to do this - as I said, you're making a good argument, but one should still think about what to do about the `Name` property.

But getting the parsing done has higher priority, I see, np.

gerhardol:
What is the status for this PR?
Should #6807 be merged for now?

GintasS:

GitCommands/RevisionReader.cs
gerhardol: Change in master to only get filenames in the initial `git-log --follow` as it is virtually impossible to parse filenames in the "normal" `git-log`.

RussKie:
@vbjay is this something you're still interested in getting merged?

vbjay:
@RussKie yes. Let me review and see if I can make it work or need to rework at all.

RussKie:

PR #6137, opened 2019-01-20
Diff List sorting options

### LINKED ISSUE #5909: [3.0] Commit status list is sorted by change type
I would call it a bug/oversight. The commit list still works but its behavior is inconsistent with the other default diff lists.

In 3.0 The commit dialog sorts unstaged file changes by type (edit, new, deletion). In previous releases they were sorted alphabetically. The other diff lists in 3.0 (selected commit in Git graph and staged change list) are also sorted alphabetically.

Open the commit dialog with multiple change types and multiple files of each type.

The files listed in the unstaged list should be sorted alphabetically by default like all the other diff lists.

If alternative sorting is desired I think we should implement that separately as described in #2518. Although sorting should be extended to all diff lists based on one setting.

- Git Extensions 3.00.00.4433
- Git 2.18.0.windows.1
- .NET Framework 4.7.3130.0
- DPI 96dpi (no scaling)
- Windows 7 64bit

<= 2.51.05

glen-nicol:
Fixes #5909 

## Proposed changes

- Add context menu to diff lists to change the sorting method between alphanumeric and git change type.
- Default to previous sorting of alphanumeric.

## Screenshots 
image

image

## Test methodology 

- TODO

## Test environment(s) 

- GIT version 2.20.1.windows.1
- Windows 10 1803


glen-nicol:
I forgot to rebase so there are merge conflicts, but I will be adding tests and squashing these after quick code review.

GitCommands/Git/AlphaGitItemStatusComparer.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace GitCommands.Git
+{
+    /// <summary>
+    /// Compares <see cref="GitItemStatus"/> objects by <see cref="GitItemStatus.Name"/> and then by <see cref="GitItemStatus.OldName"/> alphabetically.
+    /// </summary>
+    public class AlphaGitItemStatusComparer : Comparer<GitItemStatus>
+    {
+        public override int Compare(GitItemStatus x, GitItemStatus y)
+        {
+            return x.CompareTo(y);
```
drewnoakes: If this only delegates to `IComparable`, can't this type be removed and `Comparer<GitItemStatus>.Default` used instead?

Alternatively, actually do the comparison here without delegating, so it's explicit. I prefer custom comparers over one-size-might-fit-all IComparable implementations.

drewnoakes:
This looks great! I will use alphabetical sorting myself once this is merged.

What do you think of changing the menu items to read as follows?

- Sort by
  - Path
  - File status

RussKie:
@glen-nicol  could you please rebase on top of latest master and add your sign off under [Git Extensions Project Contributors Certification of Origin and Rights](https://github.com/gitextensions/gitextensions/blob/master/contributors.txt) as part of this PR.
Thank you.

RussKie:
I'm on a phone, I like alphabet-column one....

drewnoakes:
> I like alphabet-column one....

That one's good too.

drewnoakes:
@glen-nicol if you were so inclined, you could add a third sorting option: file extension

gerhardol:
Path should be secondary for extension and git type

RussKie:
> Also yes I can add an Icon and didn't know about that resource for icons. What is the TLDR for using those?

The icons are provided under [CCA/3.0](http://creativecommons.org/licenses/by/3.0/), meaning we could use them freely.
In terms of how we use icons, please follow the established pattern - images are added to and referenced from `Images.resx`.

gerhardol:
@glen-nicol Can you sign contributors.txt (possibly in a separate PR) and rebase as well as the final touchups so this PR can be finalized.

glen-nicol:
Yeah sorry I have been a bit busy the last couple weekends. I'll try to get to this in the next couple days. 
But just to be clear, do you want the final code in a separate PR or just the contributors.txt change?

RussKie:
> do you want the final code in a separate PR or just the contributors.txt change

A separate PR with your name added to contributors.txt, then it will make it easier for this PR to be merged once it is ready
Thank you

GitCommands/DiffListSortService.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private DiffListSortType GetSettingValueOrDefault()
+        {
+            var persistedSetting = AppSettings.DiffListSorting;
+            if (persistedSetting == DiffListSortType.Invalid)
+            {
+                return DiffListSortType.FilePath;
+            }
+
+            return persistedSetting;
```
RussKie: I'd move this block into `AppSettings` and delegate the responsibility for providing a correct value to `DiffListSorting` implementation

GitCommands/DiffListSortService.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        protected void RaiseSortChange()
```
RussKie: ❓  Why is it `protected`?

RussKie: To align to how MS is doing it:
```suggestion
        protected void OnDiffListSortingChanged()
```

GitCommands/DiffListSortService.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            DiffListSortingChanged?.Invoke(this, new EventArgs());
```
RussKie:
```suggestion
            DiffListSortingChanged?.Invoke(this, EventArgs.Empty);
```

GitCommands/DiffListSortService.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public event EventHandler DiffListSortingChanged;
```
RussKie: Please move before the ctor

GitCommands/DiffListSortService.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace GitCommands
+{
+    public class DiffListSortService : IDiffListSortService
+    {
+        private static readonly Lazy<DiffListSortService> _lazyDiffListSorting = new Lazy<DiffListSortService>(() => new DiffListSortService());
+
+        public static DiffListSortService Instance => _lazyDiffListSorting.Value;
+
+        private DiffListSortType _diffListSorting;
```
RussKie: please consolidate all fields, properties and methods in their logical groups
this is a good reference - https://blogs.msdn.microsoft.com/brada/2005/01/26/internal-coding-guidelines/

GitCommands/Git/GitItemStatusFileExtensionComparer.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace GitCommands.Git
+{
+    /// <summary>
+    /// Compares the file extension of <see cref="GitItemStatus.Name"/> and then by <see cref="GitItemStatus.CompareTo(GitItemStatus)"/>.
+    /// </summary>
+    public class GitItemStatusFileExtensionComparer : Comparer<GitItemStatus>
+    {
+        public override int Compare(GitItemStatus x, GitItemStatus y)
+        {
+            var lhsExt = Path.GetExtension(x.Name);
```
RussKie: ❗️  Input parameters must be validated, otherwise there is a risk of NRE

GitCommands/Git/GitItemStatusFileExtensionComparer.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitCommands.Git
+{
+    /// <summary>
+    /// Compares the file extension of <see cref="GitItemStatus.Name"/> and then by <see cref="GitItemStatus.CompareTo(GitItemStatus)"/>.
+    /// </summary>
+    public class GitItemStatusFileExtensionComparer : Comparer<GitItemStatus>
```
RussKie: Tests?

GitCommands/IDiffListSortService.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitCommands
+{
+    public enum DiffListSortType
```
RussKie: We strive for one public class per file (at least for the new submissions).
Interfaces live in the same file as their implementations (unless there are multiple implementations).

GitCommands/IDiffListSortService.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public static class DiffListSortServiceExtensions
+    {
+        /// <summary>
+        /// Provides the <see cref="IDiffListSortService.DiffListSorting"/> immediately and then an element for each <see cref="IDiffListSortService.DiffListSortingChanged"/>.
+        /// </summary>
+        /// <param name="diffListSortService">The diff list service</param>
+        /// <returns>A hot stream with one immediate cold element.</returns>
+        public static IObservable<DiffListSortType> CurrentAndFutureSorting(this IDiffListSortService diffListSortService)
```
RussKie: Is there a reason this can't be a member of `IDiffListSortService`?

GitCommands/IDiffListSortService.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+
+namespace GitCommands
+{
+    public enum DiffListSortType
+    {
+        /// <summary>
+        /// Invalid default
+        /// </summary>
+        Invalid,
```
RussKie: I would remove this, and make `FilePath = 0`

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1395,6 +1469,41 @@ private void FilterComboBox_SizeChanged(object sender, EventArgs e)
...
+                return StatusComparer.Compare(x.Tag as GitItemStatus, y.Tag as GitItemStatus);
+            }
+        }
+
+        private class ImageIndexListSorter : Comparer<ListViewItem>
+        {
+            /// <summary>
+            /// Secondary sort should be by file path.
+            /// </summary>
+            private static readonly GitStatusListSorter THEN_BY = new GitStatusListSorter(Comparer<GitItemStatus>.Default);
```
```suggestion
            private static readonly GitStatusListSorter ThenBy = new GitStatusListSorter(Comparer<GitItemStatus>.Default);
```

GitUI/UserControls/SortDiffListContextMenuItem.cs
```diff
@@ -0,0 +1,84 @@
...
+            foreach (var item in AllItems())
+            {
+                item.Click += Item_Click;
+                DropDownItems.Add(item);
+            }
+
+            RequerySortingMethod();
+        }
+
+        private ToolStripMenuItem[] AllItems()
```
RussKie: Please create a `private static readonly` variable and assign it in the ctor.
There is no need to allocate the array multiple times

UnitTests/GitUITests/UserControls/SortDiffListContextMenuItemTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        private SortDiffListContextMenuItem _itemUnderTest;
+        private IDiffListSortService _testingSortService;
+
+        private string _originalTranslation;
+
+        [OneTimeSetUp]
+        public void OneTimeSetup()
+        {
+            _originalTranslation = AppSettings.CurrentTranslation;
+            AppSettings.CurrentTranslation = "en";
```
RussKie: I think you were looking for these attributes:
```
    [SetCulture("en-US")]
    [SetUICulture("en-US")]
    [TestFixture]
    public class SortDiffListContextMenuItemTests
    {
```

GitUI/UserControls/SortDiffListContextMenuItem.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public void RequerySortingMethod()
+        {
+            var currentSort = _sortService.DiffListSorting;
+            foreach (var item in AllItems())
+            {
+                item.Checked = currentSort.Equals(item.Tag);
+            }
+        }
```
RussKie: Can't we do something like (pseudo code):
```
this.OnOpening += (s, e) => {
            var currentSort = _sortService.DiffListSorting;
            foreach (var item in AllItems())
            {
                item.Checked = currentSort.Equals(item.Tag);
            }
}
```

GitCommands/DiffListSortService.cs
glen-nicol: I put this code here so that it was easier to test. Is there a way to adequately test the AppSettings class given that there is an invalid enum value?

GitCommands/DiffListSortService.cs
glen-nicol: So that [sub classes can raise the event](https://stackoverflow.com/questions/4742214/c-sharp-accessing-parent-classs-events-not-possible).  Don't want to make it public either.

glen-nicol: Will do.

GitCommands/DiffListSortService.cs
glen-nicol: I had no idea that was even there! Thanks for teaching me something very useful.

GitCommands/DiffListSortService.cs
glen-nicol: Will do.

GitCommands/Git/GitItemStatusFileExtensionComparer.cs
glen-nicol: Good catch.

GitCommands/Git/GitItemStatusFileExtensionComparer.cs
glen-nicol: I'll add tests for these.

GitCommands/IDiffListSortService.cs
glen-nicol: So to be clear you want the enum and extensions class in their own files and the interface moved to the class's file?

GitCommands/IDiffListSortService.cs
glen-nicol: Will do

GitCommands/IDiffListSortService.cs
glen-nicol: Mostly because it doesn't have to be. Should be the same for any possible IDiffListSortService.

GitUI/UserControls/FileStatusList.cs
glen-nicol: Will do.

GitUI/UserControls/SortDiffListContextMenuItem.cs
glen-nicol: Will do.

GitUI/UserControls/SortDiffListContextMenuItem.cs
glen-nicol: Seams reasonable. I'll try that out. I don't know winforms classes much so I didn't think of that.

UnitTests/GitUITests/UserControls/SortDiffListContextMenuItemTests.cs
glen-nicol: I copied that from another test fixture. I can't remember which one now. If that ensures that the text is english then that is much better. Thanks for the tip.

GitCommands/DiffListSortService.cs
glen-nicol: Are you just referring to the lines between the fields? Sorry, haven't gotten my new VS environment completely up to snuff on the new computer. Used to tools cleaning that up for me.

GitUI/UserControls/SortDiffListContextMenuItem.cs
glen-nicol: I assume you didn't mean static. Right?

GitUI/UserControls/SortDiffListContextMenuItem.cs
glen-nicol: Will Do.

illfated:
Rebase needed to get rid of the conflict and make it mergeable again.

GitCommands/IDiffListSortService.cs
RussKie: Correct

GitUI/UserControls/SortDiffListContextMenuItem.cs
RussKie: 💭  I think it can be, the items are static and we can only show one instance of the menu at a time...
But what you have is good too. 

I'd change an array for a readonly list, as returning arrays is a bad practice (items can be replaced outside your control):
```suggestion
        private IReadOnlyList<ToolStripMenuItem> AllItems()
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -712,11 +759,30 @@ public void StoreNextIndexToSelect()
...
+        {
+            FileStatusListView.ListViewItemSorter = new GitStatusListSorter(new GitItemStatusFileExtensionComparer());
+            FileStatusListView.Sort();
+        }
+
+        public void SortByFileStatus()
+        {
+            FileStatusListView.ListViewItemSorter = new ImageIndexListSorter();
+            FileStatusListView.Sort();
+        }
```
RussKie: These new methods don't need to be `public`.
We can make them public when there is a real need to use them from outside.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1395,6 +1467,54 @@ private void FilterComboBox_SizeChanged(object sender, EventArgs e)
...
+            {
+                if (ReferenceEquals(x, y))
+                {
+                    return 0;
+                }
+                else if (x == null)
+                {
+                    return -1;
+                }
+                else if (y == null)
```
RussKie: [minor] `else`s are redundant

GitUI/UserControls/SortDiffListContextMenuItem.cs
```diff
@@ -0,0 +1,88 @@
...
+            DropDownOpening += (s, e) => RequerySortingMethod();
+            RequerySortingMethod();
+        }
+
+        private ToolStripMenuItem[] AllItems()
+        {
+            return _allItems;
+        }
+
+        public void RequerySortingMethod()
```
RussKie: This method doesn't need to be `public`. 

If I understand correctly it is make public to be available in tests.
We use a `TestAccessor` pattern to expose/test non public members, e.g.
https://github.com/gitextensions/gitextensions/blob/d4e2919f54232ddc231a88465266f6a337321bdb/GitUI/CommandsDialogs/ToolStripPushButton.cs#L83-L96

GitUI/UserControls/FileStatusList.cs
glen-nicol: Will do. That was careless of me. At one point I was calling them externally but didn't clean that up when I refactored. Thanks for catching that.

GitUI/UserControls/FileStatusList.cs
glen-nicol: Done.

GitUI/UserControls/SortDiffListContextMenuItem.cs
glen-nicol: Will do.

GitUI/UserControls/FileStatusList.cs
RussKie: If you want to test these (great that you do!), please follow the `TestAccessor` pattern we have

glen-nicol:
What are the next steps for this?

RussKie:
Rename from "wip" :) and get it merged

RussKie:
Thank you

klorinczi:
I see that there is a new context menu named "Sort by..."
The menu items are: "File Path, File Extension, File Status".

How about other sort options like: "by filename, by git creation date, by git last modification"?

The Sort by Status works fine, this is what I missed the most.

glen-nicol:
I think those are good ideas. For this PR, I was primarily focused on getting the old alphabetical sort back while provided a way for people who prefer the newer accidental status sorting of 3.0. I won't have time to do more in the near future.

If you want to do the work yourself, It should be fairly easy to add new sorting methods by creating new comparer classes and enum values to match them. There is a single place for them to be applied to the list. The only gotcha may be lack of data with current code. I can't remember if those fields were available from the current git status class.

Issue #6140, opened 2019-01-20
Build status is displayed only for a few commits


mstv:
## Current behaviour

The build status is displayed only for very few commits.

![grafik](https://user-images.githubusercontent.com/36601201/51445372-613d6900-1d04-11e9-8016-1c94d09a0265.png)
![grafik](https://user-images.githubusercontent.com/36601201/51445274-91383c80-1d03-11e9-9094-c165b3b2e102.png)

## Expected behaviour

Build status is shown.

## Steps to reproduce

- switch on the build server integration
- switch on the display of the build status icon and/or text
- restart GE

## Did this work in previous version of GitExtensions

not in 3.00.00.4433

## Environment

- Git Extensions 3.01.00.0 (and 3.00.00.4433)
- Build 67643416568784aa1f59e14864b64b8a7335ec30
- Git 2.20.1.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

## Diagnostics

```
2019-01-20T23:28:13.8721318+01:00	75	24548	UI	1	git	config --get ą	
2019-01-20T23:28:13.9506418+01:00	74	21868	UI	1	git		
2019-01-20T23:28:14.2601811+01:00	69	25260	UI	0	git	--version	
2019-01-20T23:28:15.4473319+01:00	123	5696		0	git	for-each-ref --sort=-committerdate refs/heads/ --format="%(objectname) %(refname)"	F:\Build\gitextensions\
2019-01-20T23:28:15.5713476+01:00	102	20120		0	git	ls-files -z --unmerged	F:\Build\gitextensions\
2019-01-20T23:28:15.6408564+01:00	106	14264	UI	0	git	rev-parse HEAD	F:\Build\gitextensions\
2019-01-20T23:28:15.7588714+01:00	87	13380	UI	0	git	rev-parse --git-common-dir	F:\Build\gitextensions\
2019-01-20T23:28:15.8618845+01:00	136	21620	UI	0	git	rev-parse HEAD	F:\Build\gitextensions\
2019-01-20T23:28:15.9999020+01:00	175	18296	UI	0	git	show-ref --dereference	F:\Build\gitextensions\
2019-01-20T23:28:16.2409326+01:00	139	2524		0	git	for-each-ref --sort=-committerdate refs/heads/ --format="%(objectname) %(refname)"	F:\Build\gitextensions\
[...]

PR #6159, opened 2019-01-23
Fix loading of some plugins that failed

pmiossec:
due to a bad version of "Newtonsoft.Json"

PS: the binding redirect was already there
Fixes https://github.com/gitextensions/gitextensions/issues/6140#issuecomment-456101158

## Proposed changes

Other projects use v10.0.3 and a dependency of "TfsInterop.Vs2015" use
a v8 that could end up in the output directory.
So the plugins fails to load because can find "Newtonsoft.Json" in v10

### Before

When building locally (and perhaps on AppVeyor ?), sometimes, the plugins requiring "Newtonsoft.Json" as dependency failed to load

### After

The plugins could load.


pmiossec:
As I said, I don't know which version of the dll is copied in the output folder in the AppVeyor build.
But I think that if we want to be cautious, I think we should...

PR #6162, opened 2019-01-24
Make FormReflog use UICommands when displaying FormCreateBranch

amaiorano:
With this, we can remove the ShouldRefresh-related code since interested parties already subscribe to the UICommands_PostRepositoryChanged event, including FormBrowse itself for the revision graph, and the left panel.

## Proposed changes

This change basically improves the code around FormReflog, making it use UICommands like most other places. Apart from clean up, this not only updates the revision grid when we create or reset a branch from FormReflog, but it also updates the left panel,  which is originally why I made this change.

## Screenshots 

No screenshot as there is no UI difference, but there is a UX difference: prior, when creating/resetting a branch from FormReflog, the revision grid would only get updated on closing the dialog. Now, the revision grid (and left panel, and anything else that might register on the UICommands_PostRepositoryChanged event) gets updated immediately while FormReflog is still up.

## Test methodology 

- Basic manual testing

## Test environment(s) 

- GIT 2.19.1.windows.1
- Windows 10

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2955,11 +2955,9 @@ namespace GitUI.CommandsDialogs
 
         private void toolStripMenuItemReflog_Click(object sender, EventArgs e)
         {
-            var formReflog = new FormReflog(UICommands);
-            formReflog.ShowDialog();
-            if (formReflog.ShouldRefresh)
+            using (var formReflog = new FormReflog(UICommands))
             {
-                RefreshRevisions();
+                formReflog.ShowDialog();
             }
         }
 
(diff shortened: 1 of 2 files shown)


vbjay:
NICE!

vbjay:
> FormRefLog is not much used or important as reflogs can be displayed in the revision grid

Not really.  It is very useful to see the steps taken.  If you just want commits, that's fine.  Use the grid.  You want to see the movements of the git history, use the reflog.  

"What was the thing I did 3 steps back?  Which of these many reflog commits is the one I did before x action?"

gerhardol:
>  It is very useful to see the steps taken. 

I do not disagree with that, but I have seen few users using it for other than lost commits. Command line reflog and revision grid is a workaround. But this workaround is irrelevant, this will work.

vbjay:
I love this change.  It is a great idea.

PR #6165, opened 2019-01-24
fix: Pull dialog options from shell extensions

### LINKED ISSUE #6144: Shell Extension Menu Pull
Related: #6060 
## Current behaviour

The shell extension menu pull is not pulling but doing a fetch by default.  The shell extension is only passing in pull and not pull --merge or pull --rebase.

## Expected behaviour

Users are expecting it to start with pull merge/rebase option.  The shell extension should pass in --merge or --rebase and always --autostash

## Steps to reproduce

image

Other users are saying 2.51.05 works.  

## Environment

- Git Extensions 3.1.0.4945
- Build 231592f2e9e9bce58ebfb02cb124be86e4b4e563 (Dirty)
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

RussKie:
Fixes #6144

## Proposed changes

- Pull dialog options now correctly pre-selects options when "Pull" option is invoked from Windows Explorer shell extension.

I haven't bisected it, but it is plausible we regressed as far back as afc7fa0984068d617b3e95dcb3c0fa5e57392e40.

## Test methodology 

- I ran the app from VS with debugger command line args: `pull "C:\Development\gitextensions"` and observed "Pull" dialog with "Merge" option preselected

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1729,7 +1729,7 @@ namespace GitUI
             }
             else
             {
-                StartPullDialog(remoteBranch: remoteBranch);
+                StartPullDialog(remoteBranch: remoteBranch, pullAction: AppSettings.PullAction.Merge);
             }
         }
 


SetoKaiba:
I vote for no change of this if use the default pull action from #6153. Because it can meet the both demand, @PowerAccess prefer merge and @vbjay prefer fetch.

vbjay:
I agree. Let the default they configured open. The ... at the end of the menu implies the fact that a dialog will open with a chance to change.

amaiorano:
Just so it's clear, when selecting "Git Extensions -> Pull..." from explorer:

* This change from @RussKie makes it so that the default selected action is "Merge" in the FormPull dialog.

* The change from @SetoKaiba (https://github.com/gitextensions/gitextensions/pull/6153) makes it so that the default selected action is the one selected as the "Set default Pull button action" menu (which is "Merge" if it's never set).

If what I've described above is correct, then I would opt for @SetoKaiba's change, as it's more flexible and configurable.

RussKie:
No worries, I will close this PR as obsolete.

/cc: @PowerAccess

PR #6169, opened 2019-01-25
Fix sorting of commit templates.

stivx0:
Fixes #

## Proposed changes

- Jira commit templates may be ordered by JQL and this sorting have to be saved on UI. ConcurrentDictionary is not ordered class and must not be used to store ordered data.

### Before

Incorrect sorting (have to be ordered by updateDate)
image

### After

Sorting fixed

image

## Test methodology 

- Setup Jira plugin with sorting
- Open commit form and template popup
- Templates are ordered
- Close commit form
- Open commit form again and check templates

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitCommands/CommitTemplateManager.cs
+++ b/GitCommands/CommitTemplateManager.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
@@ -43,11 +42,11 @@ namespace GitCommands
 
     public sealed class CommitTemplateManager : ICommitTemplateManager
     {
-        private class RegisteredCommitTemplateItem
+        private struct RegisteredCommitTemplateItem
         {
-            public string Name { get; }
+            public readonly string Name;
 
-            public Func<string> Text { get; }
+            public readonly Func<string> Text;
 
             public RegisteredCommitTemplateItem(string name, Func<string> text)
             {
@@ -82,7 +81,7 @@ namespace GitCommands
             {
                 lock (RegisteredTemplatesStorage)
                 {
-                    return RegisteredTemplatesStorage.Select(item => new CommitTemplateItem(item.Name, item.Text()));
+                    return RegisteredTemplatesStorage.Select(item => new CommitTemplateItem(item.Name, item.Text())).AsReadOnlyList();
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


GitCommands/CommitTemplateManager.cs
```diff
@@ -63,7 +76,16 @@ public CommitTemplateManager(IGitModule module)
...
         /// Gets the collection of all currently registered commit templates provided by plugins.
         /// </summary>
-        public IEnumerable<CommitTemplateItem> RegisteredTemplates => RegisteredTemplatesDic.Select(item => new CommitTemplateItem(item.Key, item.Value()));
+        public IEnumerable<CommitTemplateItem> RegisteredTemplates
+        {
+            get
+            {
+                lock (RegisteredTemplatesStorage)
+                {
+                    return RegisteredTemplatesStorage.Select(item => new CommitTemplateItem(item.Name, item.Text()));
```
drewnoakes: This is not thread safe. The returned enumerable will be evaluated outside the lock, and can still result in a "collection modified" exception.

The easiest fix is to call `ToList()` within the loop. In such a case I'd also return `IReadOnlyList<CTI>`.

GitCommands/CommitTemplateManager.cs
```diff
@@ -43,7 +43,20 @@ public interface ICommitTemplateManager
 
     public sealed class CommitTemplateManager : ICommitTemplateManager
     {
-        private static readonly ConcurrentDictionary<string, Func<string>> RegisteredTemplatesDic = new ConcurrentDictionary<string, Func<string>>();
+        private class RegisteredCommitTemplateItem
```
drewnoakes: This could be a `readonly struct`.

GitCommands/CommitTemplateManager.cs
```diff
@@ -63,7 +76,16 @@ public CommitTemplateManager(IGitModule module)
         /// <summary>
         /// Gets the collection of all currently registered commit templates provided by plugins.
         /// </summary>
-        public IEnumerable<CommitTemplateItem> RegisteredTemplates => RegisteredTemplatesDic.Select(item => new CommitTemplateItem(item.Key, item.Value()));
```
drewnoakes: Why couldn't you have just added something like `.OrderBy(template => template.Name)` here?

stivx0: We shouldn't hardcode sorting algorithm because users can use custom JQL sorting. Users can sort Jira issues by date, key or by other attributes. Therefore I choose this approach (to save initial ordering of Jira query result) to solve this problem.

RussKie:
Please sign the [DCO](https://github.com/gitextensions/gitextensions/blob/master/contributors.txt) before we get this merged

RussKie:
Thank you

Issue #4814, opened 2018-04-10
LeftPanel Branch tree view outstanding issues


RussKie:
Migrating issues identified during [the review](https://github.com/gitextensions/gitextensions/pull/4499#issuecomment-372022550) here.
Relates to #4499

If some of them addressed, we just need to tick them off.

**Outstanding:**

- [x] The selected icon is typically used to represent "Settings" or "Configurations". The proposed behaviour offers a different metaphor.
To me, the proposed looks very similar to what VS offers:
![image](https://user-images.githubusercontent.com/4403806/37241533-b413b4e4-24ae-11e8-94a9-29f5a4ae6d95.png)
- [x] Toolbar buttons must provide tooltips (proposed in #6173)
![image](https://user-images.githubusercontent.com/4403806/37241514-85e1a23e-24ae-11e8-8e98-f6db1cc8dc78.png)
----

<details>
<summary>Complete items</summary>

- [x] High DPI review
- [x] ~Toggle tree branch button must have `CheckOnClick = true` to indicate when the panel is expanded~
- [x] ~Suggest to consider adding a button to the panel above the tree similar to the VS Solution Explorer~  (#6174)
![image](https://user-images.githubusercontent.com/4403806/37241437-1b128a5a-24ad-11e8-83d3-0d4de0048319.png)
- [ ] ~Context menu must not show "Expand all", when all expanded, and not "Collapse all", when all collapsed~ Not going to do, as it requires more code than added value
![image](https://user-images.githubusercontent.com/4403806/37241416-b33f820c-24ac-11e8-828b-f0aee3c5328c.png)
- [x] ~Expand/collapse items must be also available, if clicked on tree nodes. Otherwise it may impractical to scroll to collapse~ (#5012)
![image](https://user-images.githubusercontent.com/4403806/37241405-85ea45d0-24ac-11e8-912a-6c7f192b6a20.png)
- [x] ~The tree must automatically expand to the current checked out branch.~ (#5012)
![image](https://user-images.githubusercontent.com/4403806/37241570-6887c38e-24af-11e8-9f31-bb154517c2d2.png)
- [x] ~Context menu must contain common items, such as "Manage remotes"~
![image](https://user-images.githubusercontent.com/4403806/37241416-b33f820c-24ac-11e8-828b-f0aee3c5328c.png)
- [x] ~Missing space~
[...]

PR #5994, opened 2018-12-26
Display ahead and behind information in the push button

pmiossec:
when the current branch is tracking a remote one.

## Proposed changes

* Display the count of unpushed commits and not integrated commit in the '2↑ 5↓' format
* Update the tooltip of the "push" button to explain the numbers
* Change the color of the "push" button if commits need to be integrated (or force pushed)

This change was made to improve understanding by the user of what Git will do when it will push.
It's a special ask by a git/gitextensions newcomer that liked the Visual Studio way to present data (because he don't really used eanough git to understand the git graph displayed in GitExtensions)

Requires Git >=2.5 (there is a control).

### Before
A normal push button:

image

### After
When there is something to push:

image

When the branch is behind (not up to date):

image

When the branch is ahead and also behind commits (that required an integration --rebase or merge-- or a force push):

image

The setting to disable the text in the "push" button for thoose that don't like button that resize themself, like the commit one (but that don't disable the tooltip and the color change in the button):

image

PS: I reused the "unstage" icon but I would prefer a new color (orange?) but I'm not a gimp expert. I will have a look later... (Or if someone could help).

## Test methodology

- manual tests

## Test environment(s) 

- GIT 2.20
- Windows 10

PR #6173, opened 2019-01-26
Toggle visibility of refs in the left panel

### LINKED ISSUE #5962: Option to show tags not stored on restart
Bug

On closing Git Extensions the option to show tags in the new left hand side branch overview does not get stored and get's reset to not checked

1. open a repository which includes tags
1. check `Show tags` found in the gear menu on the left hand panel
1. close Git Extensions and open it again
1. open any repository containing tags (might be the same as in the first step or another one)

(see gif)

gitexerror

The check box should stay checked after restart

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

no, checkbox has been introduced in 3.x (as far as I know)

RussKie:
Part of #4814
Fixes #5962

## Proposed changes

* Replace cogs icon with a tree icon
* Allow toggle visibility of branches/remotes/tags
* Visibility state is persisted in user settings

### Before

image

### After

image

----


RussKie:
The current implementation will show a node as the last one in the tree, I'm somewhat struggling to come up with an elegant solution.
@amaiorano do you think you could help a little?

amaiorano:
> The current implementation will show a node as the last one in the tree, I'm somewhat struggling to come up with an elegant solution.
> @amaiorano do you think you could help a little?

Hi @RussKie, I would be happy to help! But, would you be okay with putting this change on ice until I get my pub/sub rot change (https://github.com/gitextensions/gitextensions/pull/5669) done first? Just looking over your changes, I can see that it's going to be painful for me to rework it with my pubsub work, since the pubsub change consists mainly of replacing ReloadAsync at the RepoObjectsTree level with separate calls to a new ReloadNodes function that is invoked by each "top" node (i.e. Branches, Remotes, Tags, and eventually Submodules).

Already, I have to rework a change that went in recently (https://github.com/gitextensions/gitextensions/pull/5994) because the AheadBehind stuff should have been in RepoObjectsTree.Nodes.Branches.cs, not in RepoObjectsTree.cs. I'm trying to rework this as part of #5669, but I think I might have to make a separate PR for this, and rebase my work again on top of it.

So yeah, would you be cool with putting this one on the back burner for a bit? I actually started working again on #5669 just two days ago.

RussKie:
> However the text labels get out of sync.

If a node is shown, the item is checked and the text should say "hide"; else - unticked and "show".
Is it not what you're seeing?

* [x] "Show branches" when branches are shown to me looked odd, and that prompted me to add alternate labels.

Would it help if items were not checked?

gerhardol:
> If a node is shown, the item is checked and the text should say "hide"; else - unticked and "show".
Is it not what you're seeing?

Correct, 
I would prefer to not change the label though, just have the checkbox. "Show tags" is what is used in for instance View menu, I suggest that only 'Tags' is used in this situation.

RussKie:
Ok, simplify items labels to "Branches", "Tags", etc.
Sounds sensible, will update

RussKie:
@gerhardol  I've updated the labels.

@amaiorano please let me know when it is a good time to resurrect this. 
...or feel free to take over this piece (and close it).

amaiorano:
> @amaiorano please let me know when it is a good time to resurrect this.
> ...or feel free to take over this piece (and close it).

Since my #5669 change has been merged, you can go ahead and resurrect this. I'm not sure I'll be able to take it over, though. Am presently swamped, and trying to get my left panel submodules change in, which is a series of me breaking it down into smaller PRs.

amaiorano:
Am taking a crack at this, @RussKie.

amaiorano:
@RussKie take a look at the rebased commits. I patched up the first commit so that it makes use of the new RefreshTree code I added.

The 3rd commit makes sure to disable the entire left panel while reloading nodes, which fixes issues with toggling visibility of top nodes while refreshing, which would lead to NullReferenceExceptions. Disabling the entire panel is much safer.

The 4th commit is my crack at fixing the ordering problem. It's pretty straightforward, and does the trick. I think this should also allow us to make the order configurable easily enough.

Anyway, take it for a spin.

RussKie:
My apologies, I have started the rebase and tweaks but I have run out of time to push them in.
Thank you for finishing it up

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -201,53 +211,89 @@ protected override void OnUICommandsSourceSet(IGitUICommandsSource source)
...
+                SelectedImageKey = nameof(Images.TagHorizontal),
+            };
             _tagTree = new TagTree(_tagTreeRootNode, UICommandsSource);
-            AddTree(_tagTree);
+            AddTree(_tagTree, 2);
             _searchResult = null;
         }
 
-        private void AddTree(Tree tree)
+        private Dictionary<Tree, int> _treeToPositionIndex = new Dictionary<Tree, int>();
```
RussKie: this need to move to the top with the rest of fields

RussKie:
It works. 👍 

However there is an annoying flicker when a node is added back to the tree.

RussKie:
One more annoying behaviour - when the branches node is added it is automatically expanded, even though it may have been collapsed before hiding

amaiorano:
> One more annoying behaviour - when the branches node is added it is automatically expanded, even though it may have been collapsed before hiding

I think this is totally fine. If a user hides a node, they are opting into not seeing it anymore, likely for a long time. Adding it back and having if reset to it's default state (expanded for Branches, collapses for Tags) is fine. It's only annoying to you because you're testing it's behavior.

amaiorano:
Btw, good catch on the double refresh!

amaiorano:
> My apologies, I have started the rebase and tweaks but I have run out of time to push them in.
> Thank you for finishing it up

No worries! It's fun working together :)

PR #6174, opened 2019-01-26
Add "collapse all" button to ROT

RussKie:
Part of #4818

## Proposed changes

- Add "collapse all" button to ROT

## Screenshots 

image

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
@@ -87,9 +87,11 @@ namespace GitUI.BranchTreePanel
             this.branchSearchPanel = new System.Windows.Forms.TableLayoutPanel();
             this.lblSearchBranch = new System.Windows.Forms.Label();
             this.btnSearch = new System.Windows.Forms.Button();
+            this.btnCollapseAll = new System.Windows.Forms.Button();
             this.btnSettings = new System.Windows.Forms.Button();
             this.menuSettings = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.showTagsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.toolTip = new System.Windows.Forms.ToolTip(this.components);
             this.menuMain.SuspendLayout();
             this.menuBranch.SuspendLayout();
             this.menuRemotes.SuspendLayout();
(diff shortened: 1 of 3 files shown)


RussKie:
@amaiorano any objections to this merged?

gerhardol:
> This looks fine, though shouldn't we also have an expand all button as well? And just to make sure, it's already possible to do both by right-clicking anywhere in the left panel

Expand-all has a limited usecase in real world scenarios
The button is better placed in the bar, it is easier to get to for instance to get to remotes when branches is open (even more if submodules is added)
Remove context entries?

amaiorano:
> it is easier to get to for instance to get to remotes when branches is open

I'm not sure what you mean by this, since you can right-click anywhere in the tree and select Collapse All. I'm not against adding a button, though. Just think it's inconsistent to have 2 options in the context menu, and only 1 mapped to a button. As for removing from context menu, I don't think we should. There's no reason to remove functionality.

Actually, one thing I wanted to add at some point that might be useful in terms of expand would be to "expand all children" of selected node. In Explorer, this is mapped to the "asterisk" key on the numpad, which is one of those super useful hidden features.

amaiorano:
Or... we could do like the Test Explorer in VS and replace the current Expand/Collapse All to only apply to the currently selected node:

image

And map it to Ctrl + Right/Left Arrow.

RussKie:
I'm not sure about the shortcuts, I think you can expand/collapse selected nodes left/right arrow.

I agree with Gerhard, "expand all" is probably not as useful as "collapse all" - the VS does have the latter, but not the former.
I support the removal of the "expand/collapse" items from the shared context menu.

amaiorano:
> I agree with Gerhard, "expand all" is probably not as useful as "collapse all" - the VS does have the latter, but not the former.

Fair enough.

> I'm not sure about the shortcuts, I think you can expand/collapse selected nodes with the left/right arrows.
> I support the removal of the "expand/collapse" items from the shared context menu.

Let's not remove them just yet. I think if we transform them into expand/collapse all of selected node, they will become useful, just like in VS. Don't forget that my submodules left panel change is coming, which means more deeply nested trees than what we currently tend to see in the left panel.

RussKie:
No objections from me.

gerhardol:
> I think if we transform them into expand/collapse all of selected node, they will become useful, just like in VS.

Agree. I considered to propose collapse/expand node too.

RussKie:
Would it be ok to merge this one, or is there anything you'd like changed?

gerhardol:
> Would it be ok to merge this one, or is there anything you'd like changed?

For me, of course.

amaiorano:
> Would it be ok to merge this one, or is there anything you'd like changed?

I assumed you were asking @gerhardol, but in case you were asking me as well, this change looks fine and shouldn't cause any pain for merging against #5669.

RussKie:
> but in case you were asking me as well

I was primarily asking you as you're currently working on it, and I don't want to step on your toes.
😄

amaiorano:
> > but in case you were asking me as well
> 
> I was primarily asking you as you're currently working on it, and I don't want to step on your toes.
> 😄

Thanks, @RussKie :)

PR #4124, opened 2017-11-07
Allow to choose branch order between alphabetical and by date

Gua-naiko-che:
Remark: this is work in progress, the location of the setting is under discussion, it should be moved somewhere else.

Fixes #3907.

Changes proposed in this pull request:
 - Add a configuration setting to choose the ordering criteria of the branches in the dropdown: by date (default) or alphabetically. 
 
Screenshots before and after (if PR changes UI):
imagen

How did I test this code:
Manually: 
1. Select branches in some order other than alphabetical
2. Verify that the dropdown shows the branches by selection date, from more recently to less recently selected.
3. Change the configuration from "Last access date" to "Alphabetically".
4. Verify that the dropdown shows the branches alphabetically
5. Change the configuration from "Alphabetically" to "Last access date".
6. Verify that the branches are ordered as explained in point 2.

Has been tested on (remove any that don't apply):
 - GIT 2.14
 - Windows 10

PR #6175, opened 2019-01-26
WIP: Simplify "Go to commit" form

RussKie:
Closes #4869 as obsolete

## Proposed changes

- Since branches and tags can be easily navigated to via the left panel, `FormGoToCommit` can now be simplified to only offer navigation by commit expression.

### Before

image

### After

image

## Test environment(s) 

- Windows 10, 125% scaling

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormGoToCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormGoToCommit.Designer.cs
@@ -19,22 +19,23 @@
             this.textboxCommitExpression = new System.Windows.Forms.TextBox();
             this.label1 = new System.Windows.Forms.Label();
             this.groupBox1 = new System.Windows.Forms.GroupBox();
-            this.linkGitRevParse = new System.Windows.Forms.LinkLabel();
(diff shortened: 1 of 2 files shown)


RussKie:
perhaps, don't think it would be simple or cost effective though...

vbjay:
https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.textbox.autocompletemode

RussKie:
> The top margin should be increased.

Yes, I had a second look at it and I'm not liking the current result either.
Sadly there is no consistency amongst our dialogs, and that makes it hard to decide how a dialog should look.


> Go button should be to the right...

From my observations, a majority of our dialogs have their action buttons underneath, not to the right.

> ...or bottom as normal

Underneath the help label?

RussKie:
The trouble with providing branches/tags suggestions is the same as we had in #4869 - we must decide how to sort, and that opens a whole can of worms.

* We had sorting of branches requested and implemented in #4124;
* Then we had an alignment between the branch's dropdown and the left panel done in #5020;
* Then it was removed in 9c7c76fb619480429cbe02e11ce889b00423eac2 (it took me some time to track where it got removed).

Issue #6146, opened 2019-01-21
Saving an old revision of an LFS file saves only the LFS pointer


markonius:
## Current behaviour

When viewing the history of an LFS-tracked file, if I click "Save as" in the right click menu of an old revision of the file, only the LFS pointer exists in the saved file.

If, however, I choose "Difftool selected < - > local", the entire file is shown in KDiff.

## Expected behaviour

The entire file should be saved to disk when selecting "Save as" on an old revision of an LFS-tracked file.

## Steps to reproduce

1) Have a file in git-LFS
2) Commit some changes to it several times
3) Find the file in the file tree
4) Select "Viev history"
3) Try to save an old revision of the file
4) Observe that the file is corrupted (only the LFS pointer is saved)
5) Try to diff it
6) Observe that the entire file is visible in KDiff

## Screenshots

![image](https://user-images.githubusercontent.com/19593085/51487077-b29c3580-1da2-11e9-9abe-5580c4b754c1.png)
![image](https://user-images.githubusercontent.com/19593085/51487178-ed05d280-1da2-11e9-912d-871f5373dee0.png)
![image](https://user-images.githubusercontent.com/19593085/51487244-20486180-1da3-11e9-87a3-0fbff60988c6.png)

## Did this work in previous version of GitExtensions

As far as I'm aware, no.

## Environment

- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.19.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

vbjay:
Just a lark.  Try upgrading git.

RussKie:
@markonius are you in a position to debug it?

markonius:
Yo, I can spare some time. Gonna try upgrading git first. I'm available on gitter currently.

gerhardol:
GE uses what Git supplies here
Git does not use the LFS filters for git-diff, only git-difftool
So to provide a solution, GE would have to detect that the file is LFS and wrap the git-diff call

The textual diff uses 
git -c diff.submodule=short -c diff.noprefix=false diff --no-color --unified=3 --patience -M -C 

The GUI tool uses 
git difftool --gui --no-prompt -M -C 

[...]

PR #6189, opened 2019-01-30
WIP Add handling of LFS files to GitModule.GetFileStream

markonius:
Proposed solution for #6146  
Files tracked by LFS are now displayed properly in the "View" tab and can be "Saved as".

## Proposed changes

- GitModule.GetFileStream checks if the blob is an LFS pointer and runs `git lfs smudge` to fetch the actual file

## Test methodology 

Manual testing:
- View a history of a flile tracked by LFS
- Select an old revision of the file
- Open "View" tab or "Save as"

## Test environment(s) 

- GIT 2.20.1.windows.1
- Windows 10
- git-lfs/2.6.1

# Help wanted
This is my first contribution to the project. I'm still not familiar with the codebase.  
I've noticed that `GetFileStream` gets invoked for viewing some file types and for saving all (I think) file types.  
Should the LFS check be run for text files as well on `GitModule.GetFileText`? This one gets invoke quite a bit more often though.  
All files can be saved correctly via "Save as".  
EDIT: Also, I'm still not very knowlegable about inner workings of git, so some of my assumptions might be wrong.

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3524,6 +3524,60 @@ namespace GitCommands
 
             try
             {
+                MemoryStream memoryStream = CatFile();
+                if (!IsTrackedByLFS(memoryStream))
+                {
+                    return memoryStream;
+                }
+                else
+                {
+                    return LFSSmudge(memoryStream);
+                }
+            }
+            catch (Win32Exception ex)
+            {
+                Trace.WriteLine(ex);
+                return null;
+            }
+
+            MemoryStream LFSSmudge(MemoryStream cleanStream)
+            {
+                var args = new GitArgumentBuilder("lfs") { "smudge" };
+                using (var process = RunGitCmdDetached(args, redirectInput: true, redirectOutput: true))
+                {
+                    cleanStream.CopyTo(process.StandardInput.BaseStream);
+                    process.StandardInput.Close();
+                    cleanStream.Dispose();
+                    var dirtyStream = new MemoryStream();
+                    process.StandardOutput.BaseStream.CopyTo(dirtyStream);
+                    process.WaitForExit();
+                    dirtyStream.Position = 0;
+                    return dirtyStream;
+                }
+            }
+
+            // This is a heuristic because there's no way to get this information *exactly* without checking
+            // .gitattributes from the relevant commit, and that seems painful
+            bool IsTrackedByLFS(MemoryStream memoryStream)
+            {
+                const int bufferSize = 80 + 100 + 30; // Should be more than enough for 3 lines of the pointer
+                char[] buffer = new char[bufferSize];
+
+                using (var reader = new StreamReader(memoryStream, Encoding.UTF8, false, bufferSize, true))
+                {
+                    reader.Read(buffer, 0, bufferSize);
+                    memoryStream.Position = 0;
+                }
+
+                string possiblePointer = new string(buffer);
+
+                // From git LFS spec (loosely)
+                string regexPattern = @"version .*\noid sha256:[a-fA-F0-9]*\nsize [0-9]*\n";
+                return Regex.IsMatch(possiblePointer, regexPattern);
+            }
+
+            MemoryStream CatFile()
+            {
                 var args = new GitArgumentBuilder("cat-file")
                 {
                     "blob",
@@ -3537,11 +3591,6 @@ namespace GitCommands
                     stream.Position = 0;
                     return stream;
                 }
-            }
-            catch (Win32Exception ex)
-            {
-                Trace.WriteLine(ex);
-                return null;
             }
         }
 


GitCommands/Git/GitModule.cs
```diff
@@ -3523,6 +3523,60 @@ public MemoryStream GetFileStream(string blob)
...
+                    var dirtyStream = new MemoryStream();
+                    process.StandardOutput.BaseStream.CopyTo(dirtyStream);
+                    process.WaitForExit();
+                    dirtyStream.Position = 0;
+                    return dirtyStream;
+                }
+            }
+
+            // This is a heuristic because there's no way to get this information *exactly* without checking
+            // .gitattributes from the relevant commit, and that seems painful
```
vbjay: Exactly what you should do.  LFS may change the header.  The correct way is to use .gitattributes.  https://github.com/git-lfs/git-lfs/issues/2748#issuecomment-347938047

contributors.txt
```diff
@@ -64,3 +64,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/01/24, IEVin, Ivan Vinogradov, i.e.vin@ya.ru
 2019/01/29, maraf, Marek Fišera, mara@neptuo.com
 2019/01/30, stivx0, Ivan Sterkhov, stivx0@gmail.com
+2019/01/30, markonius, Marko Bakšić, markonius@protonmail.ch
```
vbjay: To prevent contention between other pull requests, create a separate pull request that does just this..  Once they have the cla bot in place, we won't have to worry about this.

GitCommands/Git/GitModule.cs
markonius: Oh, this is useful! But still, there is an issue that at the point of `GetFileStream` I only have the blob, not the path of the file, nor the sha of the tree. Do you have any suggestions on how I could get this information here?

contributors.txt
vbjay: No.  Base the pr off of master.  Look at the other pull requests like it.

GitCommands/Git/GitModule.cs
markonius: Is it safe to return the `process.StandardOutput` stream? Is it possible it will never get consumed and the LFS process is left hanging?  
`GetFileStream` was returning a `MemoryStream` before I touched it, I didn't want to change that because I'm not very familiar with the code yet.

vbjay: Good point.  It was just something I noticed.  Don't worry about it now.

vbjay: > Is it safe to return the process.StandardOutput stream?
No.  It should be copied.  Don't worry about the memorystream stuff.  That is a different issue. 
 
There are only two references.  
image

I'm trying to think on how to get the path from the blob.  We need at least the commit sha and the blob id.  I'll get back to you once I have a solution.

contributors.txt
markonius: Done

GitCommands/Git/GitModule.cs
vbjay: We can shorten this to something real simple. 

```bash
git cat-file --filters <blob Sha>
```

https://git-scm.com/docs/git-cat-file#git-cat-file---filters

vbjay: Dang it.  **current working tree**

markonius: The more I think about it, the more it seems to me that my original issue was just a symptom of the fact that files are *in general* not filtered when accessed through the tree. Should they be? [here](https://github.com/gitextensions/gitextensions/issues/6146#issuecomment-456462655) it seems that `git difftool` thinks they should...  
Does `git difftool` filter the files correctly anyway? Does it perhaps use the current working tree filter rules?

This gets invoked when diff the .png:  
`difftool --gui --no-prompt -M -C "<sha_rev_1>" "<sha_rev_2>" -- "<path_to_file>"`

vbjay: Yes. Difftool needs the actual contents of the file so filters are run. Read up on git filters.

markonius: Yes, I get that.

What I'm wondering is: *should* there be a similar process of applying filters *every* time GitExtensions shows a file (`cat-file` or whatever).  

Note that, as @gerhardol said, filters are not applied on git-diff.

On another note, I see it would be possible to pass all the required information down to `GetFileStream` in every case except when it is invoked from `FormVerify.saveAsToolStripMenuItem_Click`.

vbjay:
We can deal with when it should be run after it works. I think there are
reasons why diff doesn't run filters. If you think about it. Diff should
just show the pointer.  You also don't want to download every single or
just about lfs file just because you ran diff. Difftool makes sense because
the yzer expects the actual contents. Don't want to be editing a pointer
file.

GitCommands/Git/GitModule.cs
vbjay: We can deal with when it should be run after it works. I think there are
reasons why diff doesn't run filters. If you think about it. Diff should
just show the pointer.  You also don't want to download every single or
just about lfs file just because you ran diff. Difftool makes sense because
the user expects the actual contents. Don't want to be editing a pointer
file

vbjay: --textconv
--no-textconv
Allow (or disallow) external text conversion filters to be run when comparing binary files. See gitattributes[5] for details. Because textconv filters are typically a one-way conversion, the resulting diff is suitable for human consumption, but cannot be applied. For this reason, textconv filters are enabled by default only for git-diff[1] and git-log[1], but not for git-format-patch[1] or diff plumbing commands.

vbjay: --textconv
--no-textconv
Allow (or disallow) external text conversion filters to be run when comparing binary files. See gitattributes[5] for details. Because textconv filters are typically a one-way conversion, the resulting diff is suitable for human consumption, but cannot be applied. For this reason, textconv filters are enabled by default only for git-diff[1] and git-log[1], but not for git-format-patch[1] or diff plumbing commands.

markonius: OK.  
So, what's my next step? Do I try to get the commit sha and file path down to `GetFileStream` or do you want me to do something else? Or do you want me to do nothing for now?

vbjay: I know the lfs team has to get their work done and it will have to be pulled into git itself. Also, we will have to do a version check on a version that supports the official --check argument. Good that we got the ball rolling on that.

I hate that we have to read the header but it looks like it that is the best bet at the moment.  We will have to maintain this and handle version changes of the header. I would look at how git lfs reads the pointer and try to match as close as possible to that. 

Note for later  

Only do this if version doesn't support --check option when it is available.

vbjay: I know the lfs team has to get their work done and it will have to be pulled into git itself. Also, we will have to do a version check on a version that supports the official --check argument. Good that we got the ball rolling on that.

I hate that we have to read the header but it looks like it that is the best bet at the moment.  We will have to maintain this and handle version changes of the header. I would look at how git lfs reads the pointer and try to match as close as possible to that. 

Note for later  

Only do this if version doesn't support --check option when it is available.

gerhardol: > Note that, as @gerhardol said, filters are not applied on git-diff.

It could be possible to force the filters to run also for git-diff, if this is a file tracked by LFS, similar to what is planned for Save As.

> git cat-file --filters ...
> Dang it. current working tree

Is this not good enough?
The main argument against is if a file has been added to LFS at a later stage so maybe no.

The program is named GitExtensions for a reason.

vbjay: No. The issue is working directory . gitattributes could not have that filter. We want lfs to say yes you have a pointer file. We have pushed that idea at the lfs team. I can see holes in the logic and results. I want to be able to say this is a pointer without any gitattributes issues.

vbjay: --filters  

Show the content as converted by the filters configured in the current working tree for the given <path> (i.e. smudge filters, end-of-line conversion, etc). In this case, <object> has to be of the form <tree-ish>:<path>, or :<path>.

vbjay: Ok. Game plan is this. 

Use a version of the command mentioned here. https://github.com/git-lfs/git-lfs/pull/3501#issuecomment-459021407

We will have to wait till lfs supports it. For now use the header like you are.

RussKie:
Please rebase and some rework may be required after #6195

markonius:
Sorry I've gone missing for a few days already. I've had some commitments sneak up on me. Gonna get back on this as soon as I can.

vbjay:
No problem. We are all volunteers. I had to do the same. I keep up with the issues but needed to focus on rw.

GintasS:

astos-marcb:
Checking attributes for the `filter` value should suffice as a consistent starting point.
Rejecting invalid LFS pointer content would be a 2nd step.

Support for getting attribute values for a specific `commit` was introduced in [`Git 2.40`](https://github.com/git/git/commit/47cfc9bd7d0add617cf6d928e96b7d207be614f1).
Do **not** to use `--cached` for this use case (will yield undesired result)!

The alternative hack up to `v2.39` was to create an index file and use it (via GIT_INDEX_FILE environment and _with_ `--cached`) during the `check-attr` invocation.

RussKie:
Closing this PR as stale.

becm:
There is also a _different_ hack(?) required when using `libgit2` (supported in versions `v1.13.3+`, I think).
While I could not find *any* documentation and there is no define to _skip index and working directory_ when collecting attribute data, there is an implicit fall-trough [in the code](https://github.com/libgit2/libgit2/blob/0da8e725f1ebacd8093aefe966e25ba71c3d995c/src/libgit2/attr.c#L520) for `flags == 3`. ☺

So the respective value to commit attributes from *only* a commit would be setting
`git_attr_options.flags = 3 + GIT_ATTR_CHECK_INCLUDE_COMMIT`
and submit the desired commit in
`git_attr_options.attr_commit_id`.

pmiossec:
> Checking attributes for the `filter` value should suffice as a consistent starting point. Rejecting invalid LFS pointer content would be a 2nd step.

👍 

Adding also a link toward a discussion if someone wants to rise this PR from the dead : https://github.com/git-lfs/git-lfs/discussions/5313

Then the right command to use in `IsTrackedByLFS` seems to be:
`git check-attr -a --source 535b17fe4c04ce0c0e6beeaf9a6f0a959c14293c CustomDictionary.xml`

and looking for `filter=lfs`

mstv:
(without having read the discussion)
The concept of git filters - including LFS - is that they are automatically applied by git.
So, there should be a suiting git command for getting the filtered contents.

becm:
On operations like `git cat-file`, application of filters (`--filter`) must be requested explicitly.
Not sure if `libgit2` has something out of the box that can do this _in-memory_.

A successive `smudge` filter run will likely **not** catch files that were [saved outside](https://github.com/gitextensions/gitextensions/issues/12041) of the repo or with a mismatching file pattern (→ better use `git cat-file --filter` for cases like _Save as_).

The (raw) use of a `diff` filter digs up a little bit of a strange case for binary LFS data.
Should a diff:
- show LFS pointer data differences (diff=lfs) or
- create a text diff for binary content (diff=_somethingelse_)

and there _may_ be related pitfalls when comparing data from commits/index with working directory (not tested!).

gerhardol:
LFS files are often binary or big, git-diff is often not useful. 
I suggest to use difftool, maybe with a pointer like tfor very large diffs. A check for this info could be done if the diff seem to be a lfs diff (i.e. not delay the general use).

mstv:
`git -c difftool.saveremote.cmd="\"C:\Program Files\Git\usr\bin\bash.exe\" -c \"cp --force \\\"$LOCAL\\\" 'DEST_FILE'\"" difftool --tool=saveremote --find-renames --find-copies --no-prompt "COMMIT_HASH" -- "REPO_FILE"`
has the same limitations as `git catfile --filters`: It uses the configuration of the current checkout.
And in addition:
- It does nothing if `COMMIT_HASH` is currently checked out and `REPO_FILE` is unmodified. (Git should finally tell that it does not start the difftool because the object identifiers are the same. I am to create a git feature request.)
- This command has the disadvantage that the file is written to the `TMP` folder and is copied to the destination file.

So, I give up to let git handle this in a generic way, and I am going to:
- run `git cat-file blob OID` as implemented
- check the output for matching `"^version https://git-lfs.github.com/spec/v[0-9.]+\n"` (LFS pointer files must always start with the version spec. We do not need to parse the file. And there is no need to check for other attributes.)
- if necessary, let `git lfs smudge` filter that output
- if `git lfs smudge` returns an error, `trace` it and return the initial output

---

GE relies on the original git executable. We do not want to depend on alternative implementations like libgit.

The issue is about saving files (and viewing image files) but not about diffs. Users can configure how diffs shall be displayed. GE does not need to add functionality for diffs of LFS files (refer to https://github.com/mstv/test_lfs/commit/f063b8fd7d053524c1af240964d55a02c356c75b).

PR #5729, opened 2018-11-08
Remove untranslatable strings

gerhardol:
Issues raised at Transifex

Changes proposed in this pull request:
- Eliminate some untranslatable strings.

TranslationApp not run, await other possible changes
 
Screenshots before and after (if PR changes UI):
- n/a

What did I do to test the code and ensure quality:
- Built

Has been tested on (remove any that don't apply):

Issue #3600, opened 2017-03-20
Concatenate diffs of all changed files in a commit


drewnoakes:
The commit diff currently shows changed files in a list, and the diff for the selected file. To review all the changes in a commit can involve a lot of clicking/typing and moving the mouse around.

Here's an example with two modified files:

![image](https://cloud.githubusercontent.com/assets/350947/24124299/f87d12d0-0dba-11e7-856a-1fbcf6e67009.png)

Contrast this with how `gitk` shows a commit's diff (or `git show`), where all changes are available in one scrollable view:

![image](https://cloud.githubusercontent.com/assets/350947/24124361/442d3ebc-0dbb-11e7-88c0-aa42767d0f6a.png)

This is much quicker and easier to read through.

In `gitk` there is an interaction between the two views, where they track one another. Scrolling down the diff view causes the selected file to update accordingly. You can also navigate using the file list.

Because reviewing diffs requires extra manual steps in Git Extensions, I launch gitk whenever I need that feature. I'd rather have everything in one tool.

jbialobr:
What do you think about adding buttons: `Go to the previous file` and  `Go to the next file` at the bottom of the Diff panel? They could be shown when the Diff view got scrolled to the end. The buttons should also have configurable hot keys. That kind of solution also works when GitExt shows diff with all parents and combined diff.

drewnoakes:
@jbialobr I understand that such buttons would be easier to implement, but having everything in one continuous scrollable area is better for the user, imho.

Your eyes are just so much faster than your fingers. If you can get all the info in one scroll panel, the mouse wheel and your eyes can skim through very, very quickly.

As an example, imagine a long Wikipedia article was broken up with next/previous buttons for each section of the document. It'd be much harder to find what you are looking for.

Gitk has an extra nice feature here where the arrow key input focus remains on the commit tree, so you can scroll with the mouse, and navigate through commits with the keyboard.

jbialobr:
Yes I know that. I am only afraid about Diff for the second parent. This Diff is often very large. Perhaps there should be a toggle button that switches from per-file view to all-files view. The per-file view could be set as a default one when the number of changed files exceeds some limit. Also when comparing two distant revisions, there is a lot of changed files shown in the Diff.

webmaster33:
"there should be a toggle button that switches from per-file view to all-files view." 
This is a good idea.

Radrik5:
[...]

PR #6195, opened 2019-01-31
Cleanup GitModule

RussKie:
House keeping, trying to add some order to `GitModule`

### DIFF
--- a/GitCommands/CommitDataManager.cs
+++ b/GitCommands/CommitDataManager.cs
@@ -206,7 +206,7 @@ namespace GitCommands
             };
 
             // Do not cache this command, since notes can be added
-            data = GetModule().RunGitCmd(arguments, GitModule.LosslessEncoding);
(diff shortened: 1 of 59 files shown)


sharwell:
Looks good through 2023be1ac3c4ccabb51f6f240801ddedb0c05db9

vbjay:
For the large pieces that have to stick with git module, we could use partial to split into sections.
```c#
partial class Git module
{
// Diffs
} 

partial class Git module
{
// Refs
} 
```
Each one in a file nested under git module.cs

All code can be accessed but it would be easier to focus on the specific area.

RussKie:
> Thanks for making this reviewable...

I'm learning too  😄 

> Sidenote: For reviewing PRs like this, #5729 (#3600) would be useful...

I'm still not entirely sold on that UX (my _personal_ preference). It very much like the GH's UX and it is shocking...

RussKie:
Merging as is, will have more clean up in follow up submissions

PR #6196, opened 2019-01-31
fix build server settings page that could be clipped

pmiossec:
## Proposed changes

- The panel could grow depending on its content

### Before

We can't see all the fields, just the first one:

image

### After

We could see all the fields :

image

## Test methodology 

- Launching GE and using my own eyes 😉 

## Test environment(s) 

- Git Extensions 3.1.0
- Build 5ec1507ab077027c3508f4429d8d7e0e2210839d (Dirty)
- Git 2.20.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 192dpi (200% scaling)

----


pmiossec:
@gerhardol I'm not sure to understand...
Your screenshot is before the fix? 
I hope. Can you confirm it?
If, yes, I think that the clipping depend of the dpi setting.

gerhardol:
> Your screenshot is before the fix?

Yes, before the fix, just to show that it may look differently

RussKie:
Thank you

Issue #6118, opened 2019-01-17
Show push destination in commit dialog


LukeOwlclaw:
## Feature description

The commit window allows to click "Commit & push". Unfortunately one does not see to where the commit is pushed. One has to go to: Main windows -> Repository menu -> Remote repositories -> tab: Default pull behavior -> find concerning branch.

I would be much nicer if the push destination branch was shown directly next to the  "Commit & push" button. Being able to change the tracked remote branch there would also be most awesome.

## Environment

- Git Extensions 3.00.00.03-RC2
- 8cbcd2b25223c73fbe9f7dad08e498cf82734bfa  (Dirty)
- Git 2.19.2.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI X:100,00 % Y:100,00 %

RussKie:
It is a good feature to have. I too sometime wonder where would the code go, if I pushed.

@LukeOwlclaw would you like to have a go at implementing the proposal?

f-i-x-7:
The related PR was merged. I think that issue should be closed, shouldn't it?

RussKie:
Aye, thank you

PR #6158, opened 2019-01-23
Show remote branch in status bar in commit dialog

LukeOwlclaw:
Fixes issue #6118. Replaces PR https://github.com/gitextensions/gitextensions/pull/6128

## Proposed changes

- Added remote branch to status bar item branchNameLabel
- Added click event to open "Default pull behavior" for current branch- 

## Screenshots

Status bar:
image
image

## Test environment(s) 

- Git Extensions 3.1.0
- Build 4a5573077e83a20c25cb7c777619b6670c24a98b
- Git 2.19.2.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

PR #6199, opened 2019-01-31
Show remote branch in Commit dialog

pmiossec:
Update with some fixes of PR #6158 ( intended to fix issue #6118) 

## Proposed changes

- Added remote branch to status bar item branchNameLabel
- Added click event to open "Default pull behavior" for current branch

### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 595c2b99ca427defe82894e3b22e286ebdfdeeb8
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
@@ -139,6 +139,9 @@ namespace GitUI.CommandsDialogs
             + Environment.NewLine + "Use this feature when a file is big and never change."
             + Environment.NewLine + "Git will never check if the file has changed that will improve status check performance.");
         private readonly TranslationString _stopTrackingFail = new TranslationString("Fail to stop tracking the file '{0}'.");
+
+        private readonly TranslationString _statusBarBranchWithoutRemote = new TranslationString("(remote not configured)");
+        private readonly TranslationString _untrackedRemote = new TranslationString("(untracked)");
         #endregion
 
         private event Action OnStageAreaLoaded;
@@ -171,6 +174,7 @@ namespace GitUI.CommandsDialogs
         private bool _selectedDiffReloaded = true;
         [CanBeNull] private IReadOnlyList<GitItemStatus> _currentSelection;
         private int _alreadyLoadedTemplatesCount = -1;
+        private EventHandler _branchNameLabelOnClick;
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
         private FormCommit()
@@ -956,10 +960,44 @@ namespace GitUI.CommandsDialogs
             await TaskScheduler.Default;
 
             var currentBranchName = Module.GetSelectedBranch();
+            if (_branchNameLabelOnClick != null)
+            {
+                remoteNameLabel.Click -= _branchNameLabelOnClick;
+            }
+
+            var currentBranch = Module.GetRefs(false, true).FirstOrDefault(r => r.LocalName == currentBranchName);
+            if (currentBranch == null)
+            {
+                branchNameLabel.Text = currentBranchName;
+                remoteNameLabel.Text = string.Empty;
+                return;
+            }
+
+            string pushTo;
+            if (string.IsNullOrEmpty(currentBranch.TrackingRemote) || string.IsNullOrEmpty(currentBranch.MergeWith))
+            {
+                string defaultRemote = Module.GetRemoteNames().FirstOrDefault(r => r == "origin") ?? Module.GetRemoteNames().OrderBy(r => r).FirstOrDefault();
+
+                pushTo = defaultRemote != null
+                    ? $"{defaultRemote}/{currentBranchName} {_untrackedRemote.Text}"
+                    : _statusBarBranchWithoutRemote.Text;
+            }
+            else
+            {
+                pushTo = $"{currentBranch.TrackingRemote}/{currentBranch.MergeWith}";
+            }
 
             await this.SwitchToMainThreadAsync();
 
-            branchNameLabel.Text = currentBranchName;
+            branchNameLabel.Text = $"{currentBranchName} {char.ConvertFromUtf32(0x2192)}";
+            remoteNameLabel.Text = pushTo;
+
+            _branchNameLabelOnClick = async (object sender, EventArgs e) =>
+            {
+                UICommands.StartRemotesDialog(this, null, currentBranchName);
+                await UpdateBranchNameDisplayAsync();
+            };
+            remoteNameLabel.Click += _branchNameLabelOnClick;
             Text = string.Format(_formTitle.Text, currentBranchName, PathUtil.GetDisplayPath(Module.WorkingDir));
         }
 
@@ -3212,6 +3250,10 @@ namespace GitUI.CommandsDialogs
             internal ToolStripDropDownButton CommitMessageToolStripMenuItem => _formCommit.commitMessageToolStripMenuItem;
 
             internal ToolStripStatusLabel CommitAuthorStatusToolStripStatusLabel => _formCommit.commitAuthorStatus;
+
+            internal ToolStripStatusLabel CurrentBranchNameLabelStatus => _formCommit.branchNameLabel;
+
+            internal ToolStripStatusLabel RemoteNameLabelStatus => _formCommit.remoteNameLabel;
 
             internal bool ExecuteCommand(Command command)
             {
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -146,6 +146,7 @@ namespace GitUI.CommandsDialogs
             this.commitAuthorStatus = new System.Windows.Forms.ToolStripStatusLabel();
             this.toolStripStatusBranchIcon = new System.Windows.Forms.ToolStripStatusLabel();
             this.branchNameLabel = new System.Windows.Forms.ToolStripStatusLabel();
+            this.remoteNameLabel = new System.Windows.Forms.ToolStripStatusLabel();
             this.commitStagedCountLabel = new System.Windows.Forms.ToolStripStatusLabel();
             this.commitStagedCount = new System.Windows.Forms.ToolStripStatusLabel();
             this.commitCursorLineLabel = new System.Windows.Forms.ToolStripStatusLabel();
@@ -1357,6 +1358,7 @@ namespace GitUI.CommandsDialogs
             this.commitAuthorStatus,
             this.toolStripStatusBranchIcon,
             this.branchNameLabel,
+            this.remoteNameLabel,
             this.commitStagedCountLabel,
             this.commitStagedCount,
             this.commitCursorLineLabel,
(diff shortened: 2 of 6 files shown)


pmiossec:
I made the remote name label a link to make it more discoverable that you could click on it:
image

gerhardol:
Should be merged when @LukeOwlclaw has signed contributors.txt

RussKie:
@LukeOwlclaw we require your sign off.
https://github.com/gitextensions/gitextensions/pull/6158#issuecomment-460003440

RussKie:
We need to handle detached heads more gracefully:
image

RussKie:
Please add tests similar to #6202

LukeOwlclaw:
@RussKie and @gerhardol I do understand the necessity to sign https://github.com/gitextensions/gitextensions/blob/master/CONTRIBUTING.md. And that file does not explain it either. When I started working on the project that was no requirement.

@pmiossec Would you just take the changes of my commits and rebase (remove) my commits? I gift them to you. Thanks!

vbjay:
We are relicensing. Just create a pull request signing the file.

pmiossec:
>Please add tests similar to #6202

@RussKie done. Even if, due to the fact that it's more integration test and there are different cases to handle, that is not as elegant as expected. 


>@pmiossec Would you just take the changes of my commits and rebase (remove) my commits? I gift them to you. Thanks!

@LukeOwlclaw I'm not sure it is possible. Any idea on that, the others? But it's quite easy to add your name to the file. I think you could do it 😉

RussKie:
> @LukeOwlclaw I'm not sure it is possible. Any idea on that, the others? But it's quite easy to add your name to the file. I think you could do it 😉

You could squash all the changes locally and force-push it, this will overwrite the authorship. This way it would comply with the DCO.
Otherwise we need a sign-off from @LukeOwlclaw before we can merge this.

RussKie:
@pmiossec overall it looks good, but it looks like I found a use-case which doesn't appear to be providing the best UX.

I have a number of remotes configured, there is a default remote configured in git, which is automatically selected when I open `Push` dialog:
image
However opening `Commit` form I get "remote not configured" message:
image


If there is no tracking remote, we should fallback to to the default remote.

Steps to repro:
* create a new branch
* open `Commit` form
* observe "remote not configured"

gerhardol:
> If there is no tracking remote, we should fallback to to the default remote.

I believe the correct information is shown, there is no default yet which I prefer to see. Basically, _remote not configured_ would only be shown when there are no remotes.

However, the information of what will be used could be shown in some specific way.

RussKie:
Whilst technically it is correct, it is a questionable  UX.
If I were to push from the dialog, it would be pushed to my default remote.

RussKie:
ux-6199-1

We need to show where the code will be pushed to, should the user click "commit and push".

gerhardol:
> We need to show where the code will be pushed to, should the user click "commit and push".

I want to see if I have not pushed to this branch yet and the remote to use (if available) should be somehow different. But it would be a good addition.

drewnoakes:
Could you replace the ASCII `->` arrow in the footer with a unicode character? It might look a bit more professional.

For example `char.ConvertFromUtf32(0x25BA)` gives ► (which looks better in the charmap.exe screenshot than on this web page for some reason).

image

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigDQwgJwkA+AAgAwAEnARgDcAWABQnAMz8ATLwDCvAN7jea/tMEA2fgBZeAWSgESACgCUy1epuCAnKYDGAC0oA6eaQBuEBDABiCEQAtgCqMGCSMqbcmDIArABCAILm5qJiNgC+4llAA===

RussKie:
> I want to see if I have not pushed to this branch yet and the remote to use (if available) should be somehow different. But it would be a good addition.

I don't disagree, but I'm short on ideas how it should look like. Do you have any ideas/proposals?
The way it is currently implemented conveys a misleading message.



> Could you replace the ASCII `->` arrow in the footer with a unicode character? It might look a bit more professional.

👍

pmiossec:
>Could you replace the ASCII `->` arrow in the footer with a unicode character? It might look a bit more professional.

Done. But it looks horrible. The characters seemed reduced in the height. I don't know why...

>> I want to see if I have not pushed to this branch yet and the remote to use (if available) should be somehow different. But it would be a good addition.

>I don't disagree, but I'm short on ideas how it should look like. Do you have any ideas/proposals?
The way it is currently implemented conveys a misleading message.

I have tried something. Adding "(default)" if the branch is not tracked. But I'm not very pleased...
I can also change the label to "(not tracked)" and move the text outside of the link (should be better). It was a first try.

We could also play on the color of the arrow but it will be less understandable for others than just the 3 of us (except if you also add a tooltip explaining).

You could see the 2 changes here:
image

drewnoakes:
http://shapecatcher.com suggested some more unicode characters to try (that website is fun if you haven't tried it before.)

0x25B6 looks promising:

image

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigDQwgJwkA+AAgAwAEnARgDcAWABQnAMz8ATLwDCvAN7jea/tMEA2fgBZeAWSgESACgCUy1epuCAnKYDGAC0oA6eaQBuEBDABiCEQAtgCqMGCSMqbcmDIArMBa5uaiYjYAvuIZQA==

gerhardol:
> I can also change the label to "(not tracked)" and move the text outside of the link (should be better). It was a first try.

OK for me

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -956,10 +959,34 @@ private async Task UpdateBranchNameDisplayAsync()
...
+            var currentBranch = Module.GetRefs(false, true).FirstOrDefault(r => r.LocalName == currentBranchName);
+            if (currentBranch == null)
+            {
+                branchNameLabel.Text = currentBranchName;
+                remoteNameLabel.Text = string.Empty;
+                return;
+            }
+
+            var pushTo = string.IsNullOrEmpty(currentBranch.TrackingRemote) || string.IsNullOrEmpty(currentBranch.MergeWith)
+                ? Module.GetRemotes().Any(r => r.Name == "origin") ? $"(default) origin/{currentBranchName}" : _statusBarBranchWithoutRemote.Text
```
```suggestion
                ? Module.GetRemotes().Any(r => r.Name == "origin") ? $"origin/{currentBranchName} (untracked)" : _statusBarBranchWithoutRemote.Text
```
thoughts?

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -956,10 +959,34 @@ private async Task UpdateBranchNameDisplayAsync()
...
+
+            var currentBranch = Module.GetRefs(false, true).FirstOrDefault(r => r.LocalName == currentBranchName);
+            if (currentBranch == null)
+            {
+                branchNameLabel.Text = currentBranchName;
+                remoteNameLabel.Text = string.Empty;
+                return;
+            }
+
+            var pushTo = string.IsNullOrEmpty(currentBranch.TrackingRemote) || string.IsNullOrEmpty(currentBranch.MergeWith)
```
RussKie: I'm finding the condition is too long to grasp it easily
Do you think we we can present it in a more traditional way - `if..else`?

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -956,10 +959,34 @@ private async Task UpdateBranchNameDisplayAsync()
...
+                ? Module.GetRemotes().Any(r => r.Name == "origin") ? $"(default) origin/{currentBranchName}" : _statusBarBranchWithoutRemote.Text
+                : $"{currentBranch.TrackingRemote}/{currentBranch.MergeWith}";
 
             await this.SwitchToMainThreadAsync();
 
-            branchNameLabel.Text = currentBranchName;
+            branchNameLabel.Text = $"{currentBranchName} {char.ConvertFromUtf32(0x25b6)}";
```
RussKie: This symbol looks kind of rough...

What do you all think of the following?
image

GitUI/CommandsDialogs/FormCommit.cs
RussKie: `...Module.GetRemotes().Any(r => r.Name == "origin")...`
I'm afraid this is an incorrect assumption. I use "remote" and "fork" instead of "origin", and in my case I always get "_remote not configured_"

pmiossec: I also think but I take the same than in the FormPush.
How the FormPush is it working for you?

GitUI/CommandsDialogs/FormCommit.cs
pmiossec: I will have a look if it looks better but the second proposition of @drewnoakes  was decent.

GitUI/CommandsDialogs/FormCommit.cs
pmiossec: Yes better. But the question was also if this text should be in the link.
Or displayed outside which is a more intrusive change and required a new text label...

RussKie: I was looking at it just now...
This is how `FormPush` does it
https://github.com/gitextensions/gitextensions/blob/14e881406bd69ed6cd279906b849ccb342443f5b/GitUI/CommandsDialogs/FormPush.cs#L222-L251

Looks like we made an assumption about "origin" in the `FormPush` as well...

. . .

image

image

RussKie: I don't particularly mind it all being a link

pmiossec: Yes. most of the times, it's ok because that's the remote created when you clone.
But indeed, that's not perfect (especially in your case where "default" remote is renamed) and there is, as I know, no way in git to have a "default" remote.

vbjay:
> ..
there is, as I know, no way in git to have a "default" remote.

Correct. origin and upstream are defacto remote names. Those are the names we should default to. Clone creates an origin remote so stick with it. 

One thing we could do is check for origin remote and if it doesn't exist use the first remote found. If no remotes then "no remotes configured "would be appropriate.

GitUI/CommandsDialogs/FormCommit.cs
drewnoakes: The thin arrow looks nice too.

gerhardol:
> One thing we could do is check for origin remote and if it doesn't exist use the first remote found. If no remotes then "no remotes configured "would be appropriate.

Agree with Jay

pmiossec:
>One thing we could do is check for origin remote and if it doesn't exist use the first remote found. If no remotes then "no remotes configured "would be appropriate.

done.

How that looks:

image

RussKie:
image

Love it!

RussKie:
Please fix the tests and it's good to go

pmiossec:
>Please fix the tests and it's good to go

done (even if the build is stuck)

RussKie:
Top job! Thank you

RussKie:
@pmiossec looks like the master is failing due to tests
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/22425784#L241

Could you please have a look?

pmiossec:
>Could you please have a look?

It ony fails on the CI build 😢  I have tried something.
but the test are stuck 😢 
I have to leave, so i will retry later...

PR #6202, opened 2019-02-01
FormCommit no longer show the committer info

### LINKED ISSUE #6201: FormCommit: Commiter are no more displayed but link is clickable
## Current behaviour

The commiter info are not displayed but the link is there clickable.
When the bug is debbuged, the link appears. [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug)?

## Expected behaviour

We could see the commiter info.

## Steps to reproduce

Open Commit Form.

## Screenshots

image

Yes:

* 2.51 for sure
* 3, I don't know

## Environment

- Git Extensions 3.1.0
- Build 5ec1507ab077027c3508f4429d8d7e0e2210839d
- Git 2.20.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 192dpi (200% scaling)

## Diagnostics

Putting `UpdateAuthorInfo()` call in `OnShown()` instead of in `OnApplicationActivated()` solve the problem.

/cc @drewnoakes I see that you worked on it in a688346a21bd0ba9f22d0294c3547703e5a2ea20
Do you reproduce and have a idea on that?

RussKie:
The regression is caused by changes in #6134

Fixes #6201

## Test methodology 

- manual
- unit tests

/cc: @pmiossec could you confirm the fix please

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -366,17 +366,22 @@ namespace GitUI.CommandsDialogs
 
         protected override void OnApplicationActivated()
         {
-            if (!_bypassActivatedEventHandler)
+            if (!_bypassActivatedEventHandler && AppSettings.RefreshCommitDialogOnFormFocus)
             {
-                if (AppSettings.RefreshCommitDialogOnFormFocus)
-                {
-                    RescanChanges();
-                }
-
-                UpdateAuthorInfo();
+                RescanChanges();
             }
 
             base.OnApplicationActivated();
+        }
+
+        protected override void OnActivated(EventArgs e)
+        {
+            if (!_bypassActivatedEventHandler)
+            {
+                UpdateAuthorInfo();
+            }
+
+            base.OnActivated(e);
         }
 
         protected override void OnFormClosing(FormClosingEventArgs e)
@@ -408,6 +413,8 @@ namespace GitUI.CommandsDialogs
             {
                 Initialize();
             }
+
+            UpdateAuthorInfo();
 
             string message;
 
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -409,6 +404,8 @@ protected override void OnShown(EventArgs e)
                 Initialize();
             }
 
+            UpdateAuthorInfo();
+
```
vbjay: Needs to be activated event. OnShown runs once. If user alters name or email it will not stay current.

pmiossec:
>/cc: @pmiossec could you confirm the fix please

Successfully fixed.

GitUI/CommandsDialogs/FormCommit.cs
pmiossec: I think he needs to close the form to update the settings and reopen it.
If he did it in an external console, I think that we could live with that...

RussKie: Thanks Jay.
I'm split both ways... I suppose we can call it from both places.

vbjay: Wrong. I developed that piece. I know what it does. Click the link to alter the user name or email. It WAS in activated event for a reason.

RussKie: Fixed

vbjay: 👍   The issue is the user will click the link, change user name and save.  They close settings dialog.  Commit form activates and reads in new settings.  The code was in the activated event to catch user changing setting even from console and then creating issues because their info is not correct when they will commit.  When he changed the code from Activated to Application Activated, he changed that functionality.

RussKie: I have updated the code, please have a look.

pmiossec: @RussKie Still OK for me.

PR #6203, opened 2019-02-01
Jira Commit Hint plugin credentials stored in plain text

### LINKED ISSUE #4447: Jira Commit Hint plugin credentials stored in plain text
bug

Credentials are stored in plain text

Choose to persist the settings into "Distributed with current repository" settings source (it doesn't matter which layer it is stored, this layer makes it easier to see)
image

Open the GitExtensions.settings file:
image

Password isn't stored at all.

The issue may not be a big deal for home users who rarely change passwords and do not share their computers with anyone. However in a corporate environments storing passwords in plain text, especially in settings, which may be committed into a public repository, is a massive security issue.
The username must not be persisted into a configuration file either. Credentials must be persisted into Windows Vault (credential store).
Few links to start with:
* https://msdn.microsoft.com/en-us/library/aa480470.aspx
* https://stackoverflow.com/questions/17741424/retrieve-credentials-from-windows-credentials-store-using-c-sharp

2.51 / master

/cc: @ierof @pmiossec

stivx0:
Fixes #4447 

## Proposed changes

- Credentials will be saved to Windows credential manager (#4447 ). Global and local (local = distributed in this case) setting levels are supported. It is need for users, who works with several Jira instances and different accounts. Distributed level is impossible because credential manager is not file storage.
- Added CredentialManager with custom logic for GitExtensions
- Refactored ISettingControlBinding (added separate files)
- Partially updated translations, but it did not work for plugins and does not work now.

## Test methodology 
- Add, edit, delete credentials for Jira plugin
- Run Jira plugin

## Test environment(s) 

- Windows 10

----


gerhardol:
Took a quick look and a quick spin without any major comments.

This PR uses AdysTech.CredentialManager to interface the credential manager. Can someone comment on that?
(Probably OK, but I have not investigated.)

.xlf does not have to be updated, can be dropped (to not intimidate reviewers...).

GitCommands/Settings/ConfigFileSettings.cs
```diff
@@ -20,7 +20,9 @@ public ConfigFileSettings(ConfigFileSettings lowerPriority, ConfigFileSettingsCa
 
         public static ConfigFileSettings CreateEffective(GitModule module)
         {
-            return CreateLocal(module, CreateGlobal(CreateSystemWide()));
+            var result = CreateLocal(module, CreateGlobal(CreateSystemWide()));
+            result.SettingLevel = SettingLevel.Effective;
+            return result;
```
RussKie: The submission exhibits two different ways of initialising settings.
One way is via an intermediately variable (like here) and another is without.
Please settle on a single style and use it consistently.

GitExtUtils/CredentialManager.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(target) || AdysTech.CredentialManager.CredentialManager.GetCredentials(GetTarget(target)) == null)
+            {
+                return false;
+            }
+
+            return AdysTech.CredentialManager.CredentialManager.RemoveCredentials(GetTarget(target));
+        }
+
+        public static bool UpdateCredentials(string target, string userName, string password)
```
RussKie: Should this be a [SecureString](https://docs.microsoft.com/en-us/dotnet/api/system.security.securestring?view=netframework-4.7.2)?
```suggestion
        public static bool UpdateCredentials(string target, string userName, SecureString password)
```

GitExtUtils/CredentialManager.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace GitExtUtils
+{
+    public static class CredentialManager
+    {
+        private const string TargetPrefix = "GitExtensions_";
+
+        private static string GetTarget(string rawTarget)
+        {
+            return $"{TargetPrefix}{(rawTarget?.StartsWith(TargetPrefix) == true ? rawTarget.Replace(TargetPrefix, string.Empty) : rawTarget)}";
```
RussKie: The inlined conditionals are hard to review, please extract it.

GitExtUtils/CredentialManager.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Net;
+
+namespace GitExtUtils
+{
+    public static class CredentialManager
+    {
+        private const string TargetPrefix = "GitExtensions_";
+
+        private static string GetTarget(string rawTarget)
```
RussKie: All public API must validate input parameters. By not doing this `rawTarget` containing `null` or whitespace falls through to here... IMO situations like that should be treated as an exception.

GitExtUtils/CredentialManager.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            return $"{TargetPrefix}{(rawTarget?.StartsWith(TargetPrefix) == true ? rawTarget.Replace(TargetPrefix, string.Empty) : rawTarget)}";
+        }
+
+        public static bool TryGetCredentials(string target, out NetworkCredential credentials)
+        {
+            credentials = AdysTech.CredentialManager.CredentialManager.GetCredentials(GetTarget(target));
```
RussKie: Please use `using` for namespace imports.

GitUI/Plugin/GitPluginSettingsContainer.cs
```diff
@@ -35,5 +36,15 @@ public override void SetValue<T>(string name, T value, Func<T, string> encode)
         {
             ExternalSettings.SetValue(_pluginName + name, value, encode);
         }
+
+        public override void SetCredential(string name, NetworkCredential value, IGitModule gitModule)
```
RussKie:
```suggestion
        public override void SetCredentials(string name, NetworkCredential value, IGitModule gitModule)
```

GitUI/Plugin/GitPluginSettingsContainer.cs
```diff
@@ -35,5 +36,15 @@ public override void SetValue<T>(string name, T value, Func<T, string> encode)
...
+        {
+            ExternalSettings.SetCredential(name, value, gitModule);
+        }
+
+        public override NetworkCredential GetCredential(string name, IGitModule gitModule, NetworkCredential defaultValue)
```
RussKie:
```suggestion
        public override NetworkCredential GetCredentials(string name, IGitModule gitModule, NetworkCredential defaultValue)
```

RussKie: Inconsistent order of parameters
```
void SetCredential(string name, NetworkCredential value, IGitModule gitModule)
NetworkCredential GetCredential(string name, IGitModule gitModule, NetworkCredential defaultValue)
```

GitExtUtils/CredentialManager.cs
stivx0: Yes, it is more proper, but it does not make sense for Jira plugin, because there is not Jira API that take SecureString. If you wish I can add overloaded methods with SecureString. Also it does not make sense when users edit settings, because it requires special controls, which work with SecureString.

GitUI/Plugin/GitPluginSettingsContainer.cs
stivx0: It is not correct, because method return one credential, but not set of credentials

GitUI/Plugin/GitPluginSettingsContainer.cs
stivx0: It is not correct, because method take one credential, but not set of credentials

stivx0:
> I'm also not happy with `IGitModule` being passed around, it will be reviewed separately.

I'm too, but I haven't found better way to get current repository path (it is need to create credential for specific repository). May be can you advice something?

Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.csproj
```diff
@@ -63,6 +64,10 @@
       <Project>{0F1F1168-A4B2-4FA2-B17B-735140D17F39}</Project>
       <Name>GitExtUtils</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\GitUI\GitUI.csproj">
```
RussKie: Plugins must not reference any projects, that introduces coupling.

stivx0: I did it because I found other plugins with this reference. Also I planned that new CredentialsControl will be used for Bitbucket plugin, because this plugin also stores credential as plain text. What assembly should I use to implement common controls?

GitExtUtils/CredentialManager.cs
RussKie: That's cool, thank you for clarifying.

RussKie:
> I'm too, but I haven't found better way to get current repository path (it is need to create credential for specific repository). May be can you advice something?

I will allocate some time and will go through the proposal to see how I would change it.

From what I can see now the only reason to pass `IGitModule` is to get `WorkingDir` property. `IGitModule` is a catch-all interface with a mixed bag of responsibilities (state and behaviours), lugging it around to get access to a single property doesn't seem right. It also requires state and thread management (that's why it is typically injected as a `Func<IGitModule>`).
A better solution to construct a key we wish to read or write under and make a call to the settings store (writing from memory, `CredentialSetting`?).

Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.csproj
RussKie: Make sense, however GitUI is only used by the main app. Plugins should not be coupled to this project. I believe plugins current have two projects that could couple to - GitUIPluginInterfaces and ResourceManager, so anything plugin-purpose should be placed in these.

Please note we are working on decoupling and moving out plugins from the solution and the main repo, and changes like this would make that work harder.

RussKie:
I have taken a liberty and reverted a number of your changes and refactored your proposal to reduce the scope of the API change and made it more self contained. 
The code still has changes that can/need to be reverted, and it can do with more tweaks to make `CredentialsManager` unit testable, but it is outside the scope of this immediate exercise.
I have done some light testing locally (without access to Jira), and it all seems to work.

I'm eager to hear your thoughts.

RussKie:
A number of changes from earlier commits do not appear to be needed (like passing `GitModule` into various methods).
Could you please clean the PR up and we maybe in a position to get it merged.

stivx0:
When I saw proposed changes more carefully I found that your changes produced new UI bugs.

RussKie:
Could you please elaborate?

stivx0:
> Could you please elaborate?

PR #6204, opened 2019-02-01
Fix tests hang

### LINKED ISSUE #6172: Appveyor build hangs
## Current behaviour

Builds intermittently hang and get killed by timeout.
However the same builds succeed on re-run.

For example [this build](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21850786/tests) has not modified any code files, however it hang on `GitUITests.CommandsDialogs.CommitDialog.FormBrowseLeftPanelTests.RepoObjectTree_should_order_should_display_inactive_after_active_nodes` test in GitUITests.dll.

It looks like tests from `GitUITests.CommandsDialogs.CommitDialog.FormBrowseLeftPanelTests` are intermittently causing builds to fail.
For example:
* [build 21850786](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21850786/tests)
* [build 21856427](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21856427/tests)
* [build 21876661](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/21876661/tests)

## Expected behaviour

All builds finish within reasonable time (~15min) before the timeout

RussKie:
Closes #6172

## Proposed changes

- Ensure cancellation token sequences get disposed when their hosts get disposed

## Test methodology 

- many manual tests, time will tell

----

### DIFF
--- a/GitUI/UserControls/AvatarControl.Designer.cs
+++ b/GitUI/UserControls/AvatarControl.Designer.cs
@@ -7,19 +7,6 @@
         /// </summary>
         private System.ComponentModel.IContainer components = null;
 
-        /// <summary> 
-        /// Clean up any resources being used.
-        /// </summary>
-        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
-        protected override void Dispose(bool disposing)
-        {
-            if (disposing && (components != null))
-            {
-                components.Dispose();
-            }
-            base.Dispose(disposing);
-        }
-
         #region Component Designer generated code
 
         /// <summary> 
--- a/GitUI/UserControls/AvatarControl.cs
+++ b/GitUI/UserControls/AvatarControl.cs
@@ -58,6 +58,21 @@ namespace GitUI
             }
         }
 
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                _cancellationTokenSequence.Dispose();
+                components.Dispose();
+            }
+
+            base.Dispose(disposing);
+        }
+
         [CanBeNull]
         [Browsable(false)]
         public string Email { get; private set; }
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -742,6 +742,7 @@ namespace GitUI.CommitInfo
 
         protected override void DisposeCustomResources()
         {
+            _asyncLoadCancellation.Dispose();
             _revisionInfoResizedSubscription.Dispose();
             _commitMessageResizedSubscription.Dispose();
             base.DisposeCustomResources();


RussKie:
I've gone through the codebase checking all uses of the `CancellationTokenSequence`.
It looks like it generally gets disposed of with the exception of these 2 instances and one instance in `Tree` class in RepoObjectsTree.Nodes.cs.

https://github.com/gitextensions/gitextensions/blob/c27b60904a7162b4f2bbdef592a5e876a63ae520/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs#L106-L111

@amaiorano could you please ensure this gets disposed of correctly. 
I also wonder if it has anything to do with hangs we observe when the left panel is visible.

GitUI/UserControls/AvatarControl.Designer.cs
```diff
@@ -15,6 +15,7 @@ protected override void Dispose(bool disposing)
         {
             if (disposing && (components != null))
             {
+                _cancellationTokenSequence.Dispose();
```
sharwell: ❗️ Isn't this a generated code file that will get generated again when the form is changed? If so, you should be able either use an event handler or create a wrapper `IComponent` that is added to `components` and delegates the `Dispose` implementation to this code.

sharwell: 📝 If this code _doesn't_ change when controls are added to AvatarControl, let me know.

amaiorano:
> I've gone through the codebase checking all uses of the `CancellationTokenSequence`.
> It looks like it generally gets disposed of with the exception of these 2 instances and one instance in `Tree` class in RepoObjectsTree.Nodes.cs.
> 
> [gitextensions/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs](https://github.com/gitextensions/gitextensions/blob/c27b60904a7162b4f2bbdef592a5e876a63ae520/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs#L106-L111)
> 
> Lines 106 to 111 in [c27b609](/gitextensions/gitextensions/commit/c27b60904a7162b4f2bbdef592a5e876a63ae520)
> 
>  private abstract class Tree 
>  { 
>      protected readonly Nodes Nodes; 
>      private readonly IGitUICommandsSource _uiCommandsSource; 
>   
>      private readonly CancellationTokenSequence _reloadCancellationTokenSequence = new CancellationTokenSequence(); 
> @amaiorano could you please ensure this gets disposed of correctly.
> I also wonder if it has anything to do with hangs we observe when the left panel is visible.

Sure, but I'm pretty sure the Tree objects get created only once when the application starts, and live until we close the application, so I don't think disposing the CancellationTokenSequence will do much. But I'll take a look.

GitUI/UserControls/AvatarControl.Designer.cs
vbjay: I'd move it to the non designer file. See if designing the form adds the method back to designer.

RussKie:
Please let me know, if there is anything else and I'll address that in a follow up PR.

GitUI/UserControls/AvatarControl.cs
```diff
@@ -58,6 +58,21 @@ void ClearCache()
...
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                _cancellationTokenSequence.Dispose();
+                components.Dispose();
+            }
```
mstv: Why dispose the token only if there are components?

gerhardol: _cancellationTokenSequence.Dispose(); 
should probably always be called

I was confused when looking at #7365 , but it is likely not the cancellation token there

I have seen some crashes in the debugger when disposing too (no details).

RussKie: I think that is a bug, good find 👍

Issue #6180, opened 2019-01-28
Left panel ('branches') - multiple selection


seshonaar:
## Feature description

Hello!

I follow  up on [this issue](https://github.com/gitextensions/gitextensions/issues/6179) with another idea:
allow multiple selection in the branch list, in order to be able to 'delete' multiple branches.

Thank you and best regards!
Andrei

## Environment

- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

amaiorano:
Yes, I want this feature myself. Currently, though, this is non-trivial due to how context actions are implemented in the left panel. I have some ideas for how to refactor that code to make it easier to manage multiple selection actions.

seshonaar:
@amaiorano I have some work in progress for this.
Please if you like have a look at my branch
https://github.com/seshonaar/gitextensions/tree/feature/local_branch_context_menu
thanks!

RussKie:
It would make the reviewing and discussions easier if you open a PR. Just
add "WIP" to the subject.

amaiorano:
Just to be clear, #6209 doesn't address this issue, right? It addresses #6179. Is this correct?

seshonaar:
@amaiorano True. I have that on my radar as well, but I'll see to the menu items generation first.

amaiorano:
> @amaiorano True. I have that on my radar as well, but I'll see to the menu items generation first.

👍 Btw, great work so far. We're all looking forward to your changes :)

dnarayan:
I would like to multiple select branches on the left, right click, and push all of them (or make those the branches the only branches selected in the multiple push dialog). That would be awesome!

hieuxlu:
@seshonaar Can you tell me what is the current status of your implementation? How does the context menu look like now with multiple branches? Do you have the time to continue implementing this?

I'm also interested in making left panel multi-selectable, to enable viewing multiple branches (like [hiding/soloing in GitKraken](https://support.gitkraken.com/working-with-repositories/hiding-and-soloing/)). Would like to continue from your effort if you don't have the time.

gerhardol:
For "solo" you can use _Show this branch only_
Delete and compare is the cases I have seen

hieuxlu:
[...]

PR #6209, opened 2019-02-02
 #left-panel add "checkout" group context menu items for local branches

seshonaar:
#6180

Add local branch context menu items for  checkout, merge, rebase, create branch and reset. I'll call this the _check out group_.
## Proposed changes

- group the creation and enumeration of the _check out group_ under the umbrella of a generic class on branch type.
- remove the designer entries for local and remote branches that refer to menu items generated by the items generator class. 
- replace the designer entries with two instances of the generic class, one for local and one for remote. 

### Before

image

### After

image

## Test methodology 

- manual testing
- unit tests

----


GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
```diff
@@ -17,6 +17,21 @@ public partial class RepoObjectsTree
...
+
+        public interface ICreateBranch
+        {
+            bool CreateBranch();
+        }
+
+        public interface IMerge
+        {
+            bool Merge();
+        }
```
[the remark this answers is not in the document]
RussKie: I wouldn't create any new namespaces, I would just have them as first class citizens (i.e. files) under GitUI/BranchTreePanel/ folder.
I have no problem with public interfaces, arguably classes can be public too (but this is besides the point).

I do believe that there are some private nested classes when we wish to hide implementation details, however it is important to remember it is a serious trade off on testability.

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -151,5 +148,90 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
+            private ToolStripMenuItem _merge;
+
+            private ToolStripMenuItem _rebase;
+
+            protected ToolStripMenuItem _createBranch;
+
+            private ToolStripMenuItem _reset;
+
+            private ToolStripSeparator _separator;
+
```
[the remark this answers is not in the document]
RussKie: We keep fields together in our codebase.
I would probably ask for an empty line if fields had xml-doc, otherwise it is a wasted space in my books.

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -151,5 +148,77 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
+            result.Text = caption;
+            result.ToolTipText = toolTip;
+            RegisterClick(result, onClick);
+            return result;
+        }
+
+        /// <summary>
+        /// Branch node context menu "Check out" group. Contains: checkout, merge, rebase, create branch, reset and a trailing separator
+        /// </summary>
+        public class CheckoutMenuItemsGroup<TBranch> : IEnumerable<ToolStripItem>
```
RussKie: Since this is public, let's move it to own file too

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -161,27 +156,14 @@ private void InitializeComponent()
...
             // menuBranch
             // 
-            this.menuBranch.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
-            this.mnuBtnCheckoutLocal,
-            this.tsmiSpacer2,
+            this.menuBranch.Items.AddRange(_localBranchCheckoutMenuItems.Union(
+            new System.Windows.Forms.ToolStripItem[] {
             this.mnubtnBranchDelete,
-            this.mnubtnFilterLocalBranchInRevisionGrid});
+            this.mnubtnFilterLocalBranchInRevisionGrid}).ToArray());
```
RussKie: Is this a designer- or a manual change?

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -626,24 +557,27 @@ private void InitializeComponent()
...
         private ToolStripSeparator tsmiSpacer3;
         private ToolTip toolTip;
+        /// <summary>
+        /// Local branch context menu [check out] actions group
+        /// </summary>
+        private CheckoutMenuItemsGroup<LocalBranchNode> _localBranchCheckoutMenuItems;
+        /// <summary>
+        /// Remote branch context menu [check out] actions group
+        /// </summary>
+        private CheckoutMenuItemsGroup<RemoteBranchNode> _remoteBranchCheckoutMenuItems;
```
RussKie: This file is auto-generated and for the most part auto-updated, manual changes should be avoided in this file and placed in the main (i.e. non designer) file.

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
seshonaar: I moved it outside 👍

amaiorano:
Sorry for not chiming in earlier, have been busy. This is a pretty interesting change. I like the idea of creating a menu items group. I do wonder, though, whether we need to create separate interfaces for ICanCheckout, ICanCreateBranch, ICanMerge, ICanRebase, and ICanReset. Can't we have just one interface like IBranchActions or something?

Also, I wanted to mention that when I was thinking about future work on context actions for the left panel, what I was really hoping for was to find a way to have the same context menu options for branches that we have in FormBrowse in the left panel too. Right now, it's weird that they are distinct, with some actions only available in one or the other. Perhaps having a common IBranchActions might help us get there.

seshonaar:
@amaiorano I guess the advantage of these single method interfaces is that they are easy to name and might be useful in other contexts as well. I scratched my head a bit and the names I came up for an aggregate interface are not worth being mentioned here 😺 . Still, it's an open option as soon as more menu groups are born and the concepts get some stabilization.

amaiorano:
> @amaiorano I guess the advantage of these single method interfaces is that they are easy to name and might be useful in other contexts as well. I scratched my head a bit and the names I came up for an aggregate interface are not worth being mentioned here 😺 . Still, it's an open option as soon as more menu groups are born and the concepts get some stabilization.

I understand, but realistically, there aren't many types that will subclass a few of these. I personally think we should group them into IBranchActions, or perhaps more generally, IGitRefActions (or just IGitRef). One thing worth mentioning is that Tags are also shown in the left panel, and we can check them out as well:

image

seshonaar:
@amaiorano I like the idea with IGitRefActions! Also, regarding Tags, I don't see why rebase or reset wouldn't work (the log view context menu allows these actions)

RussKie:
We need to add "Rename branch" option too

amaiorano:
> @amaiorano I like the idea with IGitRefActions! Also, regarding Tags, I don't see why rebase or reset wouldn't work (the log view context menu allows these actions)

You're absolutely right!

amaiorano:
@seshonaar I like your latest changes with IGitRefActions a lot more :)

I think this is a great step in the right direction. After this, it would be nice to take a look at whether we can generalize this to the context menu in FormBrowse (the commit graph).

vbjay:
You will need to add IsBranch knowledge and other Is* methods or a way to
get such info. Right click commit that is not a branch head and renane
branch should not be enabled.

amaiorano:
> You will need to add IsBranch knowledge and other Is* methods or a way to get such info. Right click commit that is not a branch head and renane branch should not be enabled.

That's a good point. Perhaps these would actually belong in an IBranchActions interface, since a branch is a specialized ref.

vbjay:
No. Ref actions belong on IrefActions while branch specific would belong on branch. 

How about a builder class? It detects the interfaces a menu source has and the builder is a factory that builds the possible menus from the source's info. 

I am just a ref so the builder only builds the ref menu options.   I am a branch which means I am a ref + branch so I build ref +branch menus. 

This would let us mark a tree node with whatever. We could also do the same with staged and unstaged and ....  Just create different builders that generate the correct menus.

vbjay:
This would allow plugins and scripts to extend the menus too.

amaiorano:
I like your idea, @vbjay, and it seems like the kind of approach we should take in the future, when trying to consolidate context menu actions across the application; that is, in FormBrowse, the left panel, and perhaps even explorer integration (if that's at all possible).

However, for this specific PR, I suggest we keep something close to what was done here. If rename branch can't be easily added now, let's tackle that in a separate change.

seshonaar:
Well, baby steps I guess 😄 . Transforming each designer menu item into a generated one, without regressions. Already in this PR I made some ahm "adjustments" to the tooltips that contained "remote branch" and now they contain "this branch" (also visible for tags...). Also,  should the texts come from a resx file?
Regarding Rename, it should draw together with Delete, but not under a INodeEditActions interface I guess. Instead as @vbjay  said, a factory that checks ICanRename, ICanDelete (yes, the `can` interfaces again 😄 ). Insert the IRefActions there as well and there we are 🛫 . The separators are problematic as long as there are still menu items that are not generated. The 'fetch group' seems to be another big chunk to move, then delete (+ add rename) and it should be covered 📔

amaiorano:
> Already in this PR I made some ahm "adjustments" to the tooltips that contained "remote branch" and now they contain "this branch" (also visible for tags...). Also, should the texts come from a resx file?

Perhaps "this ref" so that it works for all? You can use TranslationString(), not a resx. Look at how TranslationString() is used in the code.

seshonaar:
@vbjay I'm sure the pattern will emerge, [probably] with a single menu items source scanning for capabilities. My concern for now is not to get too theoretical 😄 . The practical matter for _this_ PR was that _I wanted Rebase for local branches_ , and after @RussKie asked me if I would also like to attempt the implementation (thank you for that!), here we are. I moved 5 menu items from the designer to a more controlled place. There are _more_ of these, and migrating them to the "menu items core" is not a one PR thing. Maybe 20 😸 .

amaiorano:
Yes, I suggest we move the discussion of how to approach the future of the context menus to a new issue. Let's stay focused on this PR. Since @RussKie has done most of the review, could you do another pass on this?

seshonaar:
I added Rename and migrated Delete as well. The code now evolved towards the single menu items source as @vbjay was proposing.

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -151,5 +167,26 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
+        {
+            var result = new TMenuItem();
+            result.Image = icon;
+            result.Text = new TranslationString(captionKey).Text;
+            result.ToolTipText = new TranslationString(toolTipKey).Text;
+            RegisterClick(result, onClick);
+            return result;
+        }
+
+        private void AddContextMenuItems(IEnumerable<ToolStripItem> items, ContextMenuStrip menu, ToolStripItem insertAfter = null)
```
RussKie:
```suggestion
        private void AddContextMenuItems(ContextMenuStrip menu, IEnumerable<ToolStripItem> items, ToolStripItem insertAfter = null)
```

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -151,5 +167,26 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
+        {
+            menu.SuspendLayout();
+            int index = insertAfter == null ? 0 : Math.Max(0, menu.Items.IndexOf(insertAfter) + 1);
+            items.Reverse().ForEach(item => menu.Items.Insert(index, item));
```
RussKie: It feels like a misplaced responsibility to reverse the menu items here.
Why aren't items supplied in the desired order? Shouldn't the underlying collections be sorted in the first place?

Overall this method feels like it could be turned into an extension method and made generally available for other context menus.

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -151,5 +167,26 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
         }
+
+        /// <inheritdoc />
+        public TMenuItem CreateMenuItem<TMenuItem, TNode>(Action<TNode> onClick, string captionKey, string toolTipKey, Bitmap icon = null)
+            where TMenuItem : ToolStripItem, new()
+            where TNode : class, INode
+        {
+            var result = new TMenuItem();
+            result.Image = icon;
+            result.Text = new TranslationString(captionKey).Text;
```
RussKie: I don't believe this is a correct way of using `TranslationString`, as I am pretty sure they won't be picked up by a translation app and, as a result, won't be emitted into xlf files bound for translations.
Nor that they would get reloaded with a different language.

Please follow the established patterns for the localisation.

GitUI/BranchTreePanel/MenuItemsGenerator.cs
```diff
@@ -0,0 +1,124 @@
...
+        public bool TryGetMenuItem(MenuItemKey key, out ToolStripItem item)
+        {
+            return _itemsIndex.TryGetValue(key, out item);
+        }
+
+        private void AddRange(IEnumerable<ToolStripItemWithKey> createResults)
+        {
+            var array = createResults.ToArray();
+            _items.AddRange(array);
+            array.ForEach(r => _itemsIndex.Add(r.Key, r.Item));
```
RussKie:
```suggestion
            _items.ForEach(r => _itemsIndex[r.Key] = r.Item));
```

GitUI/BranchTreePanel/MenuItemsGenerator.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        private IEnumerable<ToolStripItemWithKey> CreateGitRefItems()
+        {
+            if (!Implements<IGitRefActions>())
+            {
+                yield break;
+            }
+
+            yield return _menuItemFactory.CreateMenuItem<ToolStripMenuItem, TNode>(
+                node => (node as IGitRefActions).Checkout(), "Checkout", GetTooltip(MenuItemKey.GitRefCheckout) ?? "Checkout this branch", Properties.Images.BranchCheckout)
```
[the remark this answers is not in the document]
seshonaar: It can't be null because the interface is checked at method entry. Also, that's a lamda on INode so I need to cast it locally

GitUI/BranchTreePanel/Interfaces/IMenuItemsGenerator.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// Builds context menu actions for a <see cref="INode"/> depending on declared interfaces
+    /// </summary>
+    public interface IMenuItemsGenerator<TNode> : IEnumerable<ToolStripItemWithKey>
+        where TNode : class, INode
+    {
+        bool TryGetMenuItem(MenuItemKey key, out ToolStripItem item);
+    }
+
+    public enum MenuItemKey
```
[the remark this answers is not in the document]
seshonaar: There's no easy way to retrieve specific menu items. For example, when filtering the local branch actions (active/inactive) or in unit tests. It also documents the capabilities of the generator.

GitUI/BranchTreePanel/Interfaces/INode.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace GitUI.BranchTreePanel.Interfaces
+{
+    public interface INode
```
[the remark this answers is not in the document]
seshonaar: OK. It's only purpose is to hide the Node class in unit tests, as working with `impl` directly is difficult

GitUI/BranchTreePanel/MenuItemsGenerator.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Linq;
+using System.Windows.Forms;
+using GitUI.BranchTreePanel.Interfaces;
+
+namespace GitUI.BranchTreePanel
+{
+    internal class MenuItemsGenerator<TNode> : IMenuItemsGenerator<TNode> where TNode : class, INode
+    {
+        private readonly IMenuItemFactory _menuItemFactory;
+
```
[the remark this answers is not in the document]
seshonaar: ah, the second time this one.  Maybe it should be enforced, as my brain is locked in this mode 😄

GitUI/BranchTreePanel/MenuItemsGenerator.cs
```diff
@@ -0,0 +1,124 @@
...
+        public bool TryGetMenuItem(MenuItemKey key, out ToolStripItem item)
+        {
+            return _itemsIndex.TryGetValue(key, out item);
+        }
+
+        private void AddRange(IEnumerable<ToolStripItemWithKey> createResults)
+        {
+            var array = createResults.ToArray();
+            _items.AddRange(array);
```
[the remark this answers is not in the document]
seshonaar: The specific danger is to enumerate createResults twice. But I'll later use `_items` for the dictionary to avoid any issue

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
seshonaar: The items are supplied in the right order. It's the insert at the same index that needs reverse. I'll changed this to increment index to avoid Reverse 😄

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
seshonaar: I'm still wrapping my head on how TranslationString is used [loading...]

GitUI/BranchTreePanel/MenuItemsGenerator.cs
seshonaar: _items is already containg items, so it would re-add old items and crash here 🙈

GitUI/BranchTreePanel/LocalBranchMenuItemsBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            return this.Where(t => keysToRetrieve.Contains(t.Key));
+        }
+
+        protected override string GetTooltip(MenuItemKey key)
+        {
+            switch (key)
+            {
+                case MenuItemKey.Rename:
+                    return "Rename this branch";
```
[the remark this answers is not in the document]
seshonaar: The returned value is a translation key, not a translated string... it will be handled be CreateMenuItem and transformed in a new TranslationString. The final item will have the text assigned through a TranslationString. Isn't this correct? ❗️

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
seshonaar: I'll close this conversation, as there's another above regarding translations

GitUI/BranchTreePanel/LocalBranchMenuItemsBehavior.cs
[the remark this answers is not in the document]
seshonaar:
```
result.Text = new TranslationString(captionKey).Text;
result.ToolTipText = new TranslationString(toolTipKey).Text;
```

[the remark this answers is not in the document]
seshonaar: @RussKie @amaiorano I understand now how the translation system works. I currently have work in progress in order to use it correctly

seshonaar:
So guys, any other feeback? Maybe some comments on the enum to translation string simple system I used? Or merge 😄

gerhardol:
Can you please submit a separate PR signing contributors.txt?
The GE team is quite small and busy with real life so sometimes responses are not so quick. Thanks to RussKie, the train is rolling...

seshonaar:
@gerhardol thanks for the instructions! Just created #6333 !

PR #6211, opened 2019-02-03
Remove .vsix from Wix installer

### LINKED ISSUE #5979: Remove VSIX from installer
feature

The Visual Studio 2015-2017 vsix is included in the GE installer with an option to install.
The install option is copying the vsix to the GE installation dir. Installation can be done from the Program Files folder. This is not clear in the installer.
This option in the installer installed the old VS addin, the vsix was likely just added in a similar way

VSIX is officially distributed on VS Marketplace.

The vsix is a simple wrapper to call GE. It is visually updated in 3.0. In master VS2019 is enabled, that should be the official release.
 
 * Remove vsix from the installer.
 * Add vsix as an artifact in Appveyor builds
 * Possibly add vsix to downloads at release builds
 * Add the vsix in master to Marketplace

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

no

gerhardol:
Upload vsix as a separate artifact to Appveyor

Fixes #5979

## Proposed changes

- Remove .vsix from Wix installer. It was included in the bundle, ticking installation only unpacked the file to the GE install dir.
- Upload vsix as a separate artifact to Appveyor

The current handling is a remain from when the old VS addin ran from the GE installer.
The vsix can be installed from the marketplace, but the current version should be offered for PR/master builds and I see no reason to not offer the vsix for releases too.

### Before

image

### After

image

## Test methodology 

- Manual iinstall
 
## Test environment(s)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -111,6 +111,7 @@ for:
     # upload the generated portable archive only
     - path: 'Setup/GitExtensions-Portable-*.zip'
     - path: 'Setup/GitExtensions-pdbs-*.zip'
+    - path: 'GitExtensionsVSIX/bin/Release/GitExtensionsVSIX.vsix'
 
 # configuration for release branches ONLY
 -
@@ -154,6 +155,7 @@ for:
           Exit-AppVeyorBuild
           return
       }
+      Push-AppveyorArtifact .\GitExtensionsVSIX\bin\Release\GitExtensionsVSIX.vsix
       # continue on with signing
       & Setup\Sign-Artifacts.ps1
       if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
--- a/Setup/Product.wxs
+++ b/Setup/Product.wxs
@@ -72,7 +72,6 @@
         <Environment Id="PATH" Name="PATH" Value="[INSTALLDIR]" Permanent="no" Part="last" Action="set" System="yes" />
       </Component>
     </Directory>
-    <PropertyRef Id="VS2015DEVENV" />
     <DirectoryRef Id="INSTALLDIR">
       <Component Id="gitextensions.ico" Guid="*">
         <File Source="..\Logo\git-extensions-logo.ico" />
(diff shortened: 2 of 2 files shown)


RussKie:
Things will need to do:
* setup a new build pipeline in Appveyor (or we could try Azure DevOps)
* setup a new signing project/definition in SignPath

I guess we could even automate publishing a new release onto the VS Market place.

gerhardol:
I suggest to open a new issue for the suggestions and merge this as is. @RussKie ?

RussKie:
> merge this as is

Yes, please liaise with @stefan-wenig to get a correct signing policy for the amended artifacts, and to make sure 3.0.0 builds get signed correctly as well.

appveyor.yml
```diff
@@ -161,6 +162,7 @@ for:
           Exit-AppVeyorBuild
           return
       }
+      Push-AppveyorArtifact .\GitExtensionsVSIX\bin\Release\GitExtensionsVSIX.vsix
```
RussKie: Having a second thought, I don't think this is a correct way.
The artifact should be sent for signing (if signing is enabled and it is a release build), or it should be published unsigned otherwise.
All of that logic is in `Setup\Sign-Artifacts.ps1`.

The most up-to date version of `Setup\Sign-Artifacts.ps1` is in 3.0.2 branch, it may worth porting it over to the master (it will happen, a matter of timing).

gerhardol: This (untested as I have not made any release builds in this branch) will upload the vsix unmodified.
Signing the vsix is an enhancement as I see it, it is not done now. Should likely be done, even if VS has not complained.
Must this be done in this PR?
(I can look into this regardless, I got privileges to SignPath project.)

sharwell: ❓ Why is this line needed? It seems like a duplicate of the line above:

https://github.com/gitextensions/gitextensions/blob/da096174401e285472afd077418e9f9d584d58cb/appveyor.yml#L121

RussKie: The VSIX is signed as part of the MSI installer, you can see the rules if you log into SignPath portal.

RussKie: We have 2 distinct paths in the appveyor:
* all non-release branches: https://github.com/gitextensions/gitextensions/blob/master/appveyor.yml#L83-L90
* release branches: https://github.com/gitextensions/gitextensions/blob/master/appveyor.yml#L122-L126

And we publish different artifacts for each.
In fact we publish different artifacts for release braches depending on whether they are from an official release (i.e. signed) or not.

RussKie: Thinking out loud

v2.x - artifacts don't get signed
v3.0 - VSIX is included into MSI, it gets signed as part of it
v3.1+ - VSIX is separate, MSI signing must continue to work as previously. We may need a new signing policy for signing VSIX

gerhardol: Signing is primarily done to ensure install time integrity. The 3.0 vsix is therefore not signed at installation time. It should be OK to distribute the vsix in the 3.0 and 3.1+ way, the official distribute is in the vs marketplace anyway. But it may be required to sign the marketplace file and it should be similar for the non-marketplace vsix as it is for non signed GE installer.

I will have to read about this. But signing the vsix should not be critical.

gerhardol: It may be possible to add an "artifacts: path:" section for release too, I choose to follow the pattern for release. But it is hard to test.

gerhardol:
I suggest this is merged as is and a new issue is created to sign  the .vsix and possibly separate the .vsix to a separate repo.

RussKie:
👍

PR #6213, opened 2019-02-04
Make sure RestoreSelectedFiles always updates SelectedIndex of both staged and unstaged file list

### LINKED ISSUE #6212: The method "RestoreSelectedFiles" should always update "SelectedIndex" of both staged and unstaged file list
## Current behaviour

Unstaging a file in the commit window makes that file *appear* selected in "unstaged file list" area. But the truth is that the file is only highlighted, not *selected*. The actual selected file is the first one in the list.

## Expected behaviour

The unstaged file should both be the highlighted and the selected file.

## Steps to reproduce

1. Make changes in several files (like at least 4) in a repository.
2. Open gitextensions commit window.
3. Stage the last file in the list (Either double click on the file name or highlight the file and press <kbd>S</kbd>.
4. Unstage that file by either double clicking or pressing <kbd>U</kbd>.
5. Press <kbd>&uarr;</kbd> and see that nothing happens (because the actual selected file is the first one).
6. Press <kbd>&darr;</kbd> and see that the second file gets highlighted.

## Screenshots

This is the initial view:

image

`contributors.txt` is staged:

image

`contributors.txt` is unstaged (either by double clicking on the file name or by pressing <kbd>U</kbd> while the file is selected:

image

The result of pressing <kbd>&darr;</kbd>:

image

Version 2.51 also had this problem.

## Environment

- OS version: Microsoft Windows NT 10.0.17134.0
- .NET version: .NET Framework 4.7.3324.0

## Proposed fix

https://github.com/monoblaine/gitextensions/commit/62e5287a8a2f4e8bb332bf9711ef8b0df25dfbfc#diff-16d3bf20df9b2330e5146f334a6e7f04

After some debugging I've found that the method `RestoreSelectedFiles` (in `FormCommit.cs`) does not call `SelectStoredNextIndex` after updating `_currentFilesList.SelectedItems`. But it *should* call that local function bacause updating `SelectedItems` makes a call to `StoreNextIndexToSelect` (in `FileStatusList.cs`)

*(This is my first contribution attempt. Hope to send a PR if everything's all right.)*

monoblaine:
Fixes #6212

## Proposed changes

- Make sure `RestoreSelectedFiles` always calls the local function `SelectStoredNextIndex`

## Test methodology 

- Using the steps I've stated in the [issue description](https://github.com/gitextensions/gitextensions/issues/6212).

## Test environment(s) 

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1154,10 +1154,8 @@ namespace GitUI.CommandsDialogs
             {
                 _currentFilesList.SelectedItems = newSelection;
             }
-            else
-            {
-                SelectStoredNextIndex();
-            }
+
+            SelectStoredNextIndex();
 
             return;
 
--- a/contributors.txt
+++ b/contributors.txt
@@ -69,6 +69,7 @@ YYYY/MM/DD, github id, Full name, email
 2019/02/09, glen-nicol, Glen Nicol, glen.nicol@live.com
 2019/02/16, mdonatas, Donatas Mačiūnas, mdonatas@gmail.com
 2019/02/16, DmitryZhelnin, Dmitry Zhelnin, mitya_zh@mail.ru
+2019/02/17, monoblaine, Serhan Apaydın, zerhan@gmail.com
 2019/02/18, wachulski, Marcin Wachulski, marcin.wachulski@gmail.com
 2019/02/25, oriash93, Ori Ashual, oriash93@gmail.com
 2019/02/25, coreyasmith, Corey Smith, coreyas@gmail.com


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1116,10 +1116,8 @@ private void RestoreSelectedFiles(IReadOnlyList<GitItemStatus> unstagedFiles, IR
             {
                 _currentFilesList.SelectedItems = newSelection;
             }
-            else
-            {
-                SelectStoredNextIndex();
-            }
+
+            SelectStoredNextIndex();
```
gerhardol: This seem to change the Unstaged selection, if there is still a valid unstaged file, selection should not be changed.

monoblaine: Could you give an example? I'm afraid I'm not following you.

RussKie:
Closing as stale. If you wish to pursue it further please ping us to re-open.
Thanks

PR #6219, opened 2019-02-06
Skip grid refresh after running a navigational script

### LINKED ISSUE #5834: Unnecessary grid refresh when running a script through a hotkey
A bug.

When a _navigational_ script is run using a hotkey, the revision grid is unnecessarily refreshed (which also deletes the navigation history). This doesn't happen when running the script through a context menu command.

1. Define a user script with command=`navigateTo:git` and arguments=`show --format="%H" HEAD^1`
2. Define a hotkey for the script
3. Select an arbitrary commit in the revision grid
4. Press the hotkey defined in step 2.

The selected commit is changed without refreshing the grid.

- Git Extensions 3.00.rc2
- Build ec111f40701aabc99829f651576d484cb2be7bce
- Git 2.19.2.windows.1
- .NET Framework 4.7.3221.0
- DPI 96dpi (no scaling)

No. _Navigational_ scripts were added in #5615. The return value of `ScriptRunner.RunScript` (indicating whether the grid should be refreshed) is considered in the context menu command handlers (see `RevisionGridControl.AddOwnScripts`) but ignored in `ScriptRunned.ExecuteScriptCommand` (which is run from `RevisionGridControl.ExecuteCommand`). I think `ExecuteScriptCommand` should return two values - whether any script ran at all and whether any of them needs the refresh.

crypto-rsa:
Fixes #5834.

## Proposed changes

- methods executing a command now return a new `struct CommandStatus` instead of a simple bool. Besides specifying whether the command was actually executed, it also contains a flag indicating whether the revision grid should be refreshed.
- navigational scripts no longer cause the grid refresh

## Test methodology 

- Manual testing (running commands both from context menu and using a hotkey)

## Test environment(s) 

- git version 2.20.1.windows.1
- Windows 10

----

### DIFF
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -73,10 +73,10 @@ namespace GitUI.Script
                 return false;
             }
 
-            return RunScript(owner, module, script, revisionGrid);
+            return RunScript(owner, module, script, uiCommands, revisionGrid);
         }
 
-        private static CommandStatus RunScript(IWin32Window owner, GitModule module, ScriptInfo scriptInfo, RevisionGridControl revisionGrid)
+        private static CommandStatus RunScript(IWin32Window owner, GitModule module, ScriptInfo scriptInfo, IGitUICommands uiCommands, RevisionGridControl revisionGrid)
         {
             if (scriptInfo.AskConfirmation && MessageBox.Show(owner, $"Do you want to execute '{scriptInfo.Name}'?", "Script", MessageBoxButtons.YesNo, MessageBoxIcon.Question) == DialogResult.No)
             {
@@ -106,7 +106,7 @@ namespace GitUI.Script
                 {
                     if (plugin.Description.ToLower().Equals(command, StringComparison.CurrentCultureIgnoreCase))
                     {
-                        var eventArgs = new GitUIEventArgs(owner, revisionGrid.UICommands);
+                        var eventArgs = new GitUIEventArgs(owner, uiCommands);
                         return new CommandStatus(true, plugin.Execute(eventArgs));
                     }
                 }
@@ -116,6 +116,11 @@ namespace GitUI.Script
 
             if (command.StartsWith(NavigateToPrefix))
             {
+                if (revisionGrid == null)
+                {
+                    return false;
+                }
+
                 command = command.Replace(NavigateToPrefix, string.Empty);
                 if (!command.IsNullOrEmpty())
                 {
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -10,25 +10,26 @@ namespace GitUI.Script
     /// <summary>Runs scripts.</summary>
     public static class ScriptRunner
     {
-        /// <summary>Tries to run scripts identified by a <paramref name="command"/>
-        /// and returns true if any executed.</summary>
-        public static bool ExecuteScriptCommand(IWin32Window owner, GitModule module, int command, RevisionGridControl revisionGrid = null)
+        /// <summary>Tries to run scripts identified by a <paramref name="command"/></summary>
+        public static CommandStatus ExecuteScriptCommand(IWin32Window owner, GitModule module, int command, RevisionGridControl revisionGrid = null)
         {
             var anyScriptExecuted = false;
+            var needsGridRefresh = false;
 
             foreach (var script in ScriptManager.GetScripts())
             {
                 if (script.HotkeyCommandIdentifier == command)
                 {
-                    RunScript(owner, module, script.Name, revisionGrid);
+                    var result = RunScript(owner, module, script.Name, revisionGrid);
                     anyScriptExecuted = true;
+                    needsGridRefresh |= result.NeedsGridRefresh;
                 }
             }
 
-            return anyScriptExecuted;
+            return new CommandStatus(anyScriptExecuted, needsGridRefresh);
         }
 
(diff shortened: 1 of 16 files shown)


RussKie:
@crypto-rsa please send a separate PR with your sign off to contributors.txt

crypto-rsa:
> However it looks like the only place we may require a refresh is in `return new CommandStatus(true, !scriptInfo.RunInBackground);`... I.e. no scripts will refresh the grid.
> Am I reading it correctly?

Not quite. See `new CommandStatus(true, plugin.Execute(eventArgs))` at line 110.

GitUI/Script/ScriptRunner.cs
```diff
@@ -115,6 +116,11 @@ private static bool RunScript(IWin32Window owner, GitModule module, ScriptInfo s
 
             if (command.StartsWith(NavigateToPrefix))
             {
+                if (revisionGrid == null)
+                {
+                    return false;
+                }
```
RussKie: Should this check be significantly higher?
`revisionGrid` is used in other places in the method in a way that would result in a NRE... E.g. line:109

RussKie:
The build is failing because if tests

GitUI/Script/ScriptRunner.cs
crypto-rsa: Good point. I was able to produce a NRE for a `plugin:` script. However, in the future there might be other types of scripts which would not require the `revisionGrid` instance, so I'd rather not abort running the script if it is `null` in the general case. Instead, I'd just do the null check for the `plugin:` scripts the same way I did for the `navigateTo:` scripts. Is that OK?

crypto-rsa: Hmm, after giving it a second thought, should we really not allow a script to run without the revision grid? It might not actually need it. We should just make sure it doesn't crash when creating the `GitUIEventArgs` instance (at line 109).

vbjay:
The goal is to expand scripts to other areas.

GitUI/Script/ScriptRunner.cs
RussKie: > should we really not allow a script to run without the revision grid?

No, we should not. Scripts must be able to run without the revision grid - #6135 is a perfect example of the use case.

crypto-rsa: For the `plugin:` scripts, the revision grid is only required to get a `GitUICommands` instance (through its `UICommands` property). Is it possible to get it in a different way? Or should we just pass `null` instead into the `GitUIEventArgs` constructor?

RussKie: `NavigateTo` scripts must only be available in the revision grid context, since it is a command performed on the grid itself.
We should be able to supply the `UICommands` instance to all scripts since it has a number of useful events and properties. 
I will need to investigate when I have time. Feel free to offer your vision.

crypto-rsa:
Also I just had an idea we could show a message box when an attempt is made to run a navigational script without the revision grid (for example through a hotkey when the diff control is focused). Right now, the script is just silently aborted and the user has no idea what happened. However, I don't know what is the GE policy on using message boxes. They are sometimes considered harmful to the workflow.

RussKie:
I think from a user's perspective that would constitute to a poor UX.

Imagine you run a script that instead of a result shows a popup saying you
need to run it having focused the revision grid, because the developers did
a poor job procuring the necessary context.

RussKie:
Here's something we can do to pass `GitUICommands`:
```diff
diff --git a/GitUI/Script/ScriptRunner.cs b/GitUI/Script/ScriptRunner.cs
index e5315d92a..752529009 100644
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -12,7 +12,7 @@ public static class ScriptRunner
     {
         /// <summary>Tries to run scripts identified by a <paramref name="command"/>
         /// and returns true if any executed.</summary>
-        public static bool ExecuteScriptCommand(IWin32Window owner, GitModule module, int command, RevisionGridControl revisionGrid = null)
+        public static bool ExecuteScriptCommand(IWin32Window owner, GitModule module, int command, IGitUICommands uiCommands, RevisionGridControl revisionGrid = null)
         {
             var anyScriptExecuted = false;
 
@@ -20,7 +20,7 @@ public static bool ExecuteScriptCommand(IWin32Window owner, GitModule module, in
             {
                 if (script.HotkeyCommandIdentifier == command)
                 {
-                    RunScript(owner, module, script.Name, revisionGrid);
+                    RunScript(owner, module, script.Name, uiCommands, revisionGrid);
                     anyScriptExecuted = true;
                 }
             }
@@ -28,7 +28,7 @@ public static bool ExecuteScriptCommand(IWin32Window owner, GitModule module, in
             return anyScriptExecuted;
         }
 
-        public static bool RunScript(IWin32Window owner, GitModule module, string scriptKey, RevisionGridControl revisionGrid)
+        public static bool RunScript(IWin32Window owner, GitModule module, string scriptKey, IGitUICommands uiCommands, RevisionGridControl revisionGrid)
         {
             if (string.IsNullOrEmpty(scriptKey))
             {
```

crypto-rsa:
> Here's something we can do to pass `GitUICommands`:

Thanks @RussKie! I've used the `IGitUICommands` instance you've passed in for the `plugin:` scripts which now work even without a valid reference to `RevisionGridControl`.
I also fixed the failing test and rebased on top of `master`.

RussKie:
Thank you

PR #6230, opened 2019-02-09
Mark all files that we inject version number into as clean

RussKie:
Mark all files that we inject version number into as clean

### DIFF
--- a/Setup/Mark-RepoClean.ps1
+++ b/Setup/Mark-RepoClean.ps1
@@ -1,0 +1,29 @@
+# Mark all files that we inject version number into as clean
+
+pushd $PSScriptRoot\..
+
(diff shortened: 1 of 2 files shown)


RussKie:
@vbjay what do you think?
Is there a better way?

gerhardol:
I would prefer to commit (and push) the changes instead

RussKie:
The build number is a dynamic value, we cant determin it before hand.
Pushing it from the build server doesnt seem correct either, and it would
trigger a cascade of never ending builds...

vbjay:
That's fine but if a submodule needs more changes you will have to go into the submodule twice. One to do the assemblyInfo and then cd into the path again to add another file. Why I had the cd then skipworktree then cd.. 

Yes could have used pushd/popd.

PR #6231, opened 2019-02-09
issue #4826: Commit form should have a selection whenever possible

### LINKED ISSUE #4826: Commit form should have a selection whenever possible
Feature.

The commit form can appear in a state where:

1. There is at least one item in either the change list or index list
2. Nothing is selected, and the UI looks empty

image

In the above example it's a nuisance to the user to have to click that file. What other state actually makes sense? Why do we force the user into one anyway?

I don't yet know how to reproduce this state, but it's not that rare. We can work out a reproduction case I'm sure.

An item is selected whenever possible. Changes take precedence as the user is more likely to be about to stage some stuff than review the index.

- GitExtensions version: 2.51.1
- GIT version: 2.16.2
- OS version: Windows 10

**Did this work in previous version of GitExtensions?**

I think I've seen this behaviour for many years. It may have always been like this.

mast-eu:
Fixes #4826 

## Proposed changes

- When the user clicks not exactly on a item (i.e. the empty space of the unstaged or staged area), the first visible item in that area is automatically selected (fixes reproduction method 1).
This fix works well when you have only one modified file, because the same file will be re-selected. 
However, when you have 2 or more modified files, you might get a somehow  unxpected behaviour: when you have the second file selected and than click on the empty space, it will change the selected file to the first one in the list.

- The `SelectedDiff` is not cleared immediately when user hits `DEL`, but only if the confirmation dialog is actually confirmed (fixes reproduction method 2)

## Test methodology

- manual testing
- ran unit tests

## Test environment(s)

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 120dpi (125% scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1180,7 +1180,7 @@ namespace GitUI.CommandsDialogs
 
         private void ClearDiffViewIfNoFilesLeft()
         {
-            if ((Staged.IsEmpty && Unstaged.IsEmpty) || (!Unstaged.SelectedItems.Any() && !Staged.SelectedItems.Any()))
+            if (Staged.IsEmpty && Unstaged.IsEmpty)
             {
                 SelectedDiff.Clear();
             }
@@ -1515,12 +1515,13 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            ClearDiffViewIfNoFilesLeft();
-
             if (!Unstaged.SelectedItems.Any())
             {
+                Unstaged.SelectFirstVisibleItem();
                 return;
             }
+
+            ClearDiffViewIfNoFilesLeft();
 
             Staged.ClearSelected();
 
@@ -1775,6 +1776,12 @@ namespace GitUI.CommandsDialogs
         {
             if (_currentFilesList != Staged || _skipUpdate)
             {
+                return;
+            }
+
+            if (!Staged.SelectedItems.Any())
+            {
+                Staged.SelectFirstVisibleItem();
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


mast-eu:
Just a note: the link to the The Developer Certificate of Origin in the template is wrong.

RussKie:
Could you please split it in two different PRs.

> * The `SelectedDiff` is not cleared immediately when user hits `DEL`, but only if the confirmation dialog is actually confirmed (fixes reproduction method 2)

This can be merged straight away.

The other is not as simple and must be discussed. 
I have run your fix locally and I don't think it offers the best UX possible, and we may need to iterate over it before we can get it merged.

Few issues identified with a quick test:
* Tops items now get selected twice:
double-select
* The top item gets selected even though it may be outside the view
incorrect-select

mast-eu:
You are right. The UX regarding reproduction method 1 deserves more discussion. I'll think about it.

Reproduction method 2 has been separated in an own PR (#6234).

mast-eu:
I realized there are currently 2 different strategies for dealing with clicks on empty areas when showing diffs.

The main form unselects the file and clears the diff view:
main2

The commit dialog (without the modification from this PR) does the same like the main form:
commit

However, double-clicking a commit opens a window that behaves slightly differently. Clicking the empty area still de-selects the file, but the diff view is not cleared.
diff_popup2
Also this seems a bit odd to me. Shouldn't the file remain selected in that case?

Is there a particular reason for the different behaviors? Or is it simply a non consistency?
And in general: Is there a use case where it makes sense to unselect all items (either with or without clearing the diff view)? I cannot think of one.

gerhardol:
> Is there a particular reason for the different behaviors? Or is it simply a non consistency?

Probably happens to be that way, some problem fixed in one of the views.
For Commit I recall something about context menu problems when selections persisted. I can see now that you get an Exception if right clicking in Staged file viewer and selecting Edit file.

RussKie:
> However, double-clicking a commit opens a window that behaves slightly differently. Clicking the empty area still de-selects the file, but the diff view is not cleared.

Yes, it looks like an inconsistency, and needs to be aligned with the others.

mast-eu:
> I can see now that you get an Exception if right clicking in Staged file viewer and selecting Edit file.

I can confirm the Exception in the staged file viewer. However it is not related to this PR, but happens also with the released version and with the current master. I just filled an issue for this (#6246).

RussKie:
Altering the behaviour in the `FormCommit` is undesirable, let's align the behaviour in the history form.

mast-eu:
> Altering the behaviour in the `FormCommit` is undesirable, let's align the behaviour in the history form.

I agree, since I cannot find a satisfactory fix for the `FormCommit`.

PR #6049, opened 2019-01-07
#6029 fix broken multiline textboxes

NikolayXHD:
Fixes #6029

## Proposed changes
Handle missing `Ctrl`+`Backspace` hotkey by emulating `Ctrl`+`Shift`+`Left`, `Backspace` key sequence.
This approach works fine for both single-line and multiline textbox.

There is a  visual downside, one can see how the word becomes selected and then removed one moment after.

## Test methodology

Manually tested by editing multiline & single-line inputs in commit template window.

## Test environment(s)

Windows 7 SP1

Issue #6029, opened 2019-01-02
Commit Template Settings: Commit Template field one line only


vbjay:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
![image](https://user-images.githubusercontent.com/3629489/50607879-ca198a00-0e98-11e9-90e6-20e8899da6db.png)
Try pressing enter or shift enter in textbox to add lines.
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
- In commit window, open 
![image](https://user-images.githubusercontent.com/3629489/50607947-ffbe7300-0e98-11e9-8bd8-caf2a6201851.png)
- Try to type multiple lines

**What is the expected behavior?**
Be able to type multiple lines.
**Environment you encounter the issue:**

- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 6.3.9600.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

**Did this work in previous version of GitExtensions (which)?**
Not sure.

vbjay:
Workaround.  Type info in text editor and copy.  Paste into textbox.

gerhardol:
Same problem as #5975 (other forms too?)
I have not found the core problem.
(Only a very limited workaround)

vbjay:
You know. I wonder if we are altering with win32 API?

gerhardol:
> You know. I wonder if we are altering with win32 API?

I am not aware of such changes.

RussKie:
@NikolayXHD any ideas?

vbjay:
> I am not aware of such changes.

I bet it might be unintentionally done. Think I'll use spy++ and look at window styles of multiline textboxes.

vbjay:
Commit Text Textbox
![image](https://user-images.githubusercontent.com/3629489/50617156-be8a8b00-0eb9-11e9-9d2a-34336a6aa465.png)

![image](https://user-images.githubusercontent.com/3629489/50617172-ccd8a700-0eb9-11e9-9f12-645b8218b35d.png)

![image](https://user-images.githubusercontent.com/3629489/50617207-fa255500-0eb9-11e9-820b-7fa49365cfb5.png)


[...]

PR #6256, opened 2019-02-16
Manually handle Ctrl+Backspace on TextBox and ComboBox controls

NikolayXHD:
Fixes problems introduced by #6029

## Proposed changes

1. Removing word on `Ctrl`+`Backspace` manually instead of using `SendKeys.Send("^+{LEFT} {BACKSPACE}")`

turns out `SendKeys.Send` solution is
- Unreliable: in some environments, maybe depending on keyboard type, the second and subsequent words are not removed by 1 `Backspace` click.
[proof](https://github.com/gitextensions/gitextensions/pull/6049#issuecomment-464287771)

- Incomplete: Only `LeftCtrl` + `Backspace` was handled correctly

2. Apply the same fix to `ComboBox`es

## Test methodology
manual

## Test environment(s)

- Windows 7SP1

----

### DIFF
--- a/GitExtUtils/GitUI/ControlHotkeyExtensions.cs
+++ b/GitExtUtils/GitUI/ControlHotkeyExtensions.cs
@@ -63,10 +63,9 @@ namespace GitUI
                     // after the word is selected and before it is erased.
                     _beginUpdateMethod.Invoke(sender, parameters: null);
 
-                    SetText(text.Substring(0, from) + text.Substring(from + length));
-
                     SetSelectionStart(from);
-                    SetSelectionLength(0);
+                    SetSelectionLength(length);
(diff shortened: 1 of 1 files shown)


RussKie:
Thank you

Quppa:
I just noticed that I'm able to delete words from read-only text boxes now when I press `Ctrl`+`Backspace` (e.g. in the change log dialog) - is there any easy way to check whether the current control is read-only before removing text?

NikolayXHD:
> I just noticed that I'm able to delete words from read-only text boxes now when I press `Ctrl`+`Backspace` (e.g. in the change log dialog) - is there any easy way to check whether the current control is read-only before removing text?

Sure, I will fix that. Thanks for reporting.

PR #6261, opened 2019-02-17
Fix unit test that fails only on CI server

pmiossec:
Test failing: FormCommitTests.Should_display_branch_and_no_remote_info_in_statusbar()


pmiossec:
@RussKie in fact, it seems good...

RussKie:
Thank you

RussKie:
Bah! It is still failing...
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/22439282

I managed to get it to fail running `FormCommitTests` tests repeatedly...

RussKie:
The test failed locally only once and I wasn't debugging...
It kind of feels it is a race condition, where one test checks out a revision whilst this test is expecting to be on master...

PR #6264, opened 2019-02-18
Tests: run tests in their own repositories

pmiossec:
to try to avoid race conditions

Fixes unit test failing in AppVeyor

### DIFF
--- a/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
+++ b/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using CommonTestUtils;
 using FluentAssertions;
 using GitUI;
 using GitUI.CommandsDialogs;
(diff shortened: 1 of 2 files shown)


UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -85,42 +86,57 @@ public void Should_update_committer_info_on_form_activated()
         [Test]
         public void Should_display_branch_and_no_remote_info_in_statusbar()
         {
-            _referenceRepository.CheckoutMaster();
-            RunFormTest(form =>
+            var repository = new ReferenceRepository(new GitModuleTestHelper("branch_and_no_remote"));
```
RussKie: NB: `repository` will need to be disposed of so the underlying folder is deleted

pmiossec: Thanks. done

pmiossec: @RussKie can you cancel the current build that is stuck and also cancel the builds that  are not related to building code?

RussKie: done
you can reach me faster on gitter

pmiossec:
@RussKie I think I tackled it this time ;)

It was due to the variable `_moduleTestHelper` that was static for no good reason.

RussKie:
Nice! 👍

pmiossec:
In fact, the test is still failing and a better look at the code make me understand what is the problem exactly....

That is not a race condition problem (which is more in line with the value we get where it is not initialized instead of with the value of another test)

The problem lie here: https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormCommit.cs#L1008

`ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync());`

The code we want to test is run in a task and we don't wait for the task to complete.
So sometimes the assert happens after, sometimes before.

There is no way to guarantee the execution is completed or not :(

@sharwell An idea how to fix the test?

sharwell:
This breaks an optimization where the repository does not need to be fully recreated between tests. See d8ff048cd68cfe4ac4ab00b967b286f72c6868b6.

sharwell:
There should not be any race conditions between these tests. Is it possible that the tests are set to run in parallel? UI tests cannot run in parallel.

sharwell:
> The code we want to test is run in a task and we don't wait for the task to complete.

At a high level, you'll want to call `JoinPendingOperationsAsync`. However, you'll need to modify `UITest.RunForm` to use a `JoinableTaskFactory` instance which is not backed by a `JoinableTaskCollection`, or else the join operation will deadlock.

RussKie:
Thinking about it (that I wasn't afforded in the past few days) we had a similar problem with the left panel tests.
I think the underlying problem is caused by us not passing a cancellation token into `UpdateBranchNameDisplayAsync` and not breaking out when the form is disposed.

RussKie:
Something like this perhaps?

```diff
diff --git a/GitUI/CommandsDialogs/FormCommit.cs b/GitUI/CommandsDialogs/FormCommit.cs
index 1a9bd8131..ad967d959 100644
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -154,6 +154,7 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly ToolStripMenuItem _resetSelectedLinesToolStripMenuItem;
         private readonly AsyncLoader _unstagedLoader = new AsyncLoader();
         private readonly bool _useFormCommitMessage = AppSettings.UseFormCommitMessage;
+        private readonly CancellationTokenSequence _cancellationTokenSequence = new CancellationTokenSequence();
         private readonly CancellationTokenSequence _interactiveAddSequence = new CancellationTokenSequence();
         private readonly SplitterManager _splitterManager = new SplitterManager(new AppSettingsPath("CommitDialog"));
         private readonly Subject<string> _selectionFilterSubject = new Subject<string>();
@@ -350,6 +351,7 @@ protected override void Dispose(bool disposing)
             if (disposing)
             {
                 _unstagedLoader.Dispose();
+                _cancellationTokenSequence.Dispose();
                 _interactiveAddSequence.Dispose();
                 components?.Dispose();
             }
@@ -955,7 +957,7 @@ private void EnableStageButtons(bool enable)
             ResetUnStaged.Enabled = Unstaged.AllItems.Any();
         }
 
-        private async Task UpdateBranchNameDisplayAsync()
+        private async Task UpdateBranchNameDisplayAsync(CancellationToken cancellationToken)
         {
             await TaskScheduler.Default;
 
@@ -987,7 +989,7 @@ private async Task UpdateBranchNameDisplayAsync()
                 pushTo = $"{currentBranch.TrackingRemote}/{currentBranch.MergeWith}";
             }
 
-            await this.SwitchToMainThreadAsync();
+            await this.SwitchToMainThreadAsync(cancellationToken);
 
             branchNameLabel.Text = $"{currentBranchName} {char.ConvertFromUtf32(0x2192)}";
             remoteNameLabel.Text = pushTo;
@@ -995,7 +997,7 @@ private async Task UpdateBranchNameDisplayAsync()
             _branchNameLabelOnClick = async (object sender, EventArgs e) =>
             {
                 UICommands.StartRemotesDialog(this, null, currentBranchName);
-                await UpdateBranchNameDisplayAsync();
+                await UpdateBranchNameDisplayAsync(cancellationToken);
             };
             remoteNameLabel.Click += _branchNameLabelOnClick;
             Text = string.Format(_formTitle.Text, currentBranchName, PathUtil.GetDisplayPath(Module.WorkingDir));
@@ -1004,8 +1006,9 @@ private async Task UpdateBranchNameDisplayAsync()
         private void Initialize(bool loadUnstaged = true)
         {
             _initialized = true;
+            var cancellationToken = _cancellationTokenSequence.Next();
 
-            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync());
+            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(cancellationToken));
 
             using (WaitCursorScope.Enter())
             {
@@ -3218,7 +3221,8 @@ private void createBranchToolStripButton_Click(object sender, EventArgs e)
                 return;
             }
 
-            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync());
+            var cancellationToken = _cancellationTokenSequence.Next();
+            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(cancellationToken));
         }

```

RussKie:
I've submitted #6273, let's see if that addressed the problem

PR #6273, opened 2019-02-19
Fix flaky form commit tests

RussKie:


sharwell:
I don't think cancelling operations will fix this. The tests do not reuse form instances from previous test, and the asynchronous operations _shouldn't_ be operating on static state. If I understand the problem correctly, you need to wait for asynchronous fire-and-forget operations to complete before testing assertions. To do this, you can either add the fire-and-forget operations started by the form to a `JoinableTaskCollection` held by the form (which the test can join when necessary), or implement some other tracking strategy.

RussKie:
I think you're right.
🤔

RussKie:
Updated

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -186,6 +187,7 @@ public FormCommit([NotNull] GitUICommands commands, CommitKind commitKind = Comm
             : base(commands)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
+            _joinableTasks = new JoinableTaskCollection(ThreadHelper.JoinableTaskContext);
```
sharwell: 💭 Not sure if there's a difference between this and `ThreadHelper.JoinableTaskContext.CreateCollection()`

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -88,6 +88,8 @@ public void Should_display_branch_and_no_remote_info_in_statusbar()
             _referenceRepository.CheckoutMaster();
             RunFormTest(form =>
             {
+                form.GetTestAccessor().RunningTasks.Join();
```
sharwell: 💭 Not sure if we need to use this or `JoinTillEmptyAsync`

/cc @AArnott

GitUI/CommandsDialogs/FormCommit.cs
AArnott: There isn't -- today. But we recommend the `CreateCollection` method when you don't care about the concrete type you're creating so that our "factory" can create a collection best suited for the VS app (if it were to change in the future). So ya, using `ThreadHelper.JoinableTaskContext.CreateCollection()` is preferred.

---
In reply to: [258625981](https://github.com/gitextensions/gitextensions/pull/6273#discussion_r258625981) [](ancestors = 258625981)

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1185,11 +1187,13 @@ private void ShowChanges(GitItemStatus item, bool staged)
...
 
-            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            var task = ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
                 await SetSelectedDiffAsync(item, staged);
                 _selectedDiffReloaded = true;
-            }).FileAndForget();
+            });
+            _joinableTasks.Add(task);
+            task.FileAndForget();
```
AArnott: You can make this step implicit by using a jtf instance that is created specifically for your `_joinableTasks` collection, so that the tasks are added automatically. 
Just add this to your ctor:
```cs
_joinableTaskFactory = ThreadHelper.JoinableTaskContext.CreateFactory(_joinableTasks);
```
Then use that jtf instead of ThreadHelper.JTF here (and anywhere else you call RunAsync without awaiting the result).

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
AArnott: Oh boy, I forgot the Join method even existed on that collection type. That's an unfortunate name, since `JoinableTask.Join()` is a blocking call, one might expect *this* to be as well. But it isn't. This immediately returns. You need to call `JoinTillEmptyAsync` (and await or synchronously block with JTF.Run) in order to suspend your test till the operations are done.

---
In reply to: [258626551](https://github.com/gitextensions/gitextensions/pull/6273#discussion_r258626551) [](ancestors = 258626551)

GitUI/CommandsDialogs/FormCommit.cs
sharwell: We still want to be able to join outstanding operations started by `ThreadHelper.JoinableTaskFactory`. Is there a way to automatically add these fire-and-forget operations to both collections automatically?

AArnott: What are "both collections" that you refer to? There's only the one collection that's created in this class. `ThreadHelper.JoinableTaskFactory` has no collection backing it.

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
RussKie: > ...one might expect _this_ to be as well. But it isn't. This immediately returns.

Aha! That probably explains why the test continued to fail.

@sharwell I'm a bit at loss how to correctly await `JoinTillEmptyAsync` call in the test:
```
            RunFormTest(form =>
            {
                form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync();

                var currentBranchNameLabelStatus = form.GetTestAccessor().CurrentBranchNameLabelStatus;
                var remoteNameLabelStatus = form.GetTestAccessor().RemoteNameLabelStatus;

                Assert.AreEqual("master →", currentBranchNameLabelStatus.Text);
                Assert.AreEqual("(remote not configured)", remoteNameLabelStatus.Text);
            });
```

How would you await it?

AArnott: If you can adjust `RunFormTest` to take a `Func<Task>` instead of (or in addition to) an `Action`, then  you can make this entire delegate async and then just *await*. That would be ideal since it lets your test emulate a real user who isn't always controlling the UI thread.
But your other option is to simply call:
```cs
joinableTaskFactory.Run(() => form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync());
```

GitUI/CommandsDialogs/FormCommit.cs
sharwell: Our `ThreadHelper` provides a collection to the factory:
https://github.com/gitextensions/gitextensions/blob/a2a6defafb96de0847ff92bfe01cb8275fa55084/GitExtUtils/GitUI/ThreadHelper.cs#L45-L47

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -86,64 +85,49 @@ public void Should_update_committer_info_on_form_activated()
...
             {
-                var commands = new GitUICommands(repository.Module);
-                repository.CheckoutRevision();
-                UITest.RunForm<FormCommit>(
-                    () => { Assert.True(commands.StartCommitDialog(owner: null)); },
-                    form =>
-                    {
-                        var currentBranchNameLabelStatus = form.GetTestAccessor().CurrentBranchNameLabelStatus;
-                        var remoteNameLabelStatus = form.GetTestAccessor().RemoteNameLabelStatus;
+                form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync();
```
sharwell: ❗️ This needs to use something like this:

```suggestion
                ThreadHelper.JoinableTaskFactory.Run(() => form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync());
```

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -86,64 +85,49 @@ public void Should_update_committer_info_on_form_activated()
...
             {
-                var commands = new GitUICommands(repository.Module);
-                repository.CreateRemoteForMasterBranch();
-                UITest.RunForm<FormCommit>(
-                    () => { Assert.True(commands.StartCommitDialog(owner: null)); },
-                    form =>
-                {
-                    var currentBranchNameLabelStatus = form.GetTestAccessor().CurrentBranchNameLabelStatus;
-                    var remoteNameLabelStatus = form.GetTestAccessor().RemoteNameLabelStatus;
+                form.GetTestAccessor().RunningTasks.Join();
```
sharwell:
```suggestion
                ThreadHelper.JoinableTaskFactory.Run(() => form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync());
```

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -86,64 +85,49 @@ public void Should_update_committer_info_on_form_activated()
...
             {
-                var commands = new GitUICommands(repository.Module);
-                repository.CheckoutMaster();
-                UITest.RunForm<FormCommit>(
-                    () => { Assert.True(commands.StartCommitDialog(owner: null)); },
-                    form =>
-                    {
-                        var currentBranchNameLabelStatus = form.GetTestAccessor().CurrentBranchNameLabelStatus;
-                        var remoteNameLabelStatus = form.GetTestAccessor().RemoteNameLabelStatus;
+                form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync();
```
sharwell:
```suggestion
                ThreadHelper.JoinableTaskFactory.Run(() => form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync());
```

GitUI/CommandsDialogs/FormCommit.cs
AArnott: Oh! I have a hard time *not* reading `ThreadHelper` to mean the one in VS, which doesn't have a backing collection. 
No, there's no way at present to have one instance of JTF add a JoinableTask to *two* collections automatically.

RussKie:
I've spent few hours debugging and pulling my hair... 
The test would sometimes pass, sometimes fail with the message we've observed and sometimes hang. It is like a concurrency issue, however it feels caused by something else other than `UpdateBranchNameDisplayAsync`.

I arrived to a conclusion that the culprit is `EditNetSpell` control and its autocompletion functionality.
When the `FormCommit` is loaded, `EditNetSpell` control is loaded and fires off `ToggleAutoCompletion`:
https://github.com/gitextensions/gitextensions/blob/74956a5c6d4f67798eac88f7a2810fda158b629d/GitUI/SpellChecker/EditNetSpell.cs#L206-L227

That in turn fires off `InitializeAutoCompleteWordsTask` that initialises available `IAutoCompleteProvider`s, and that's where the problem seems to be happening... 

I arrived to the above empirically, as debugging fails on line:86 because GitExtensions.dll isn't loaded in the test appdomain:
https://github.com/gitextensions/gitextensions/blob/14e881406bd69ed6cd279906b849ccb342443f5b/GitUI/AutoCompletion/CommitAutoCompleteProvider.cs#L77-L86

Once I have set `AppSettings.ProvideAutocompletion=false`, the tests started behaving as expected (although it is virtually impossible to simulate "not failing" tests with 100% assurance).

RussKie:
Looks like tests are passing on Appveyor.
@sharwell please have a look. when you have time. Thanks

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -155,10 +155,13 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly AsyncLoader _unstagedLoader = new AsyncLoader();
         private readonly bool _useFormCommitMessage = AppSettings.UseFormCommitMessage;
         private readonly CancellationTokenSequence _interactiveAddSequence = new CancellationTokenSequence();
+        private readonly CancellationTokenSequence _cancellationTokenSequence = new CancellationTokenSequence();
         private readonly SplitterManager _splitterManager = new SplitterManager(new AppSettingsPath("CommitDialog"));
         private readonly Subject<string> _selectionFilterSubject = new Subject<string>();
         private readonly IFullPathResolver _fullPathResolver;
         private readonly List<string> _formattedLines = new List<string>();
+        private readonly JoinableTaskCollection _joinableTasks;
+        private readonly JoinableTaskFactory _joinableTaskFactory;
```
sharwell: ❗️ We need the tasks to be in the main task collection, and there's no way to create a factory that is backed by two collections. This means we need to use the original factory and manually add the resulting tasks to `_joinableTasks`.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -987,6 +996,7 @@ private async Task UpdateBranchNameDisplayAsync()
                 pushTo = $"{currentBranch.TrackingRemote}/{currentBranch.MergeWith}";
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
             await this.SwitchToMainThreadAsync();
```
sharwell: Better to have this the other way around. Also pass the cancellation token to `SwitchToMainThreadAsync`.

```suggestion
             await this.SwitchToMainThreadAsync(cancellationToken);
             cancellationToken.ThrowIfCancellationRequested();
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -155,10 +155,13 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly AsyncLoader _unstagedLoader = new AsyncLoader();
         private readonly bool _useFormCommitMessage = AppSettings.UseFormCommitMessage;
         private readonly CancellationTokenSequence _interactiveAddSequence = new CancellationTokenSequence();
+        private readonly CancellationTokenSequence _cancellationTokenSequence = new CancellationTokenSequence();
```
sharwell: 💡 Maybe a more descriptive name for this, since it's associated with a specific operation?

UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
```diff
@@ -15,6 +15,8 @@ namespace GitUITests.CommandsDialogs.CommitDialog
     [Apartment(ApartmentState.STA)]
     public class FormCommitTests
     {
+        private bool _originalProviderAutocompletion;
```
sharwell:
```suggestion
        private bool _originalProvideAutocompletion;
```

GitUI/CommandsDialogs/FormCommit.cs
AArnott: Can you file this use case on vs-threading? We recently picked up the ability to nest collections in JTF, but have not exposed this. It sounds like it would be useful to you.

RussKie: Do I need to change the code in mean time, or what I have now is passable?

GitUI/CommandsDialogs/FormCommit.cs
RussKie: I'm a bit confused and would appreciate some schooling.

After `await this.SwitchToMainThreadAsync(...)` we are on the UI thread. Suppose a cancellation was requested between these two statements, would calling `cancellationToken.ThrowIfCancellationRequested()` not result in an exception to be thrown on UI thread and thus crash the app?

RussKie:
I've applied the changes, please have a look again.

GitUI/CommandsDialogs/FormCommit.cs
sharwell: We need to change the code. Our test cleanup sequence relies on asynchronous operations getting tracked in the main collection. Failing to add operations there could lead to cross-test interference.

https://github.com/gitextensions/gitextensions/blob/96090c8d0114397fb0bcfbbb342261d1da45e904/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs#L53-L57

GitUI/CommandsDialogs/FormCommit.cs
sharwell: This one is a rather complicated subject.

* Checking cancellation before `SwitchToMainThreadAsync` is simply unnecessary. You can pass the cancellation token to `SwitchToMainThreadAsync` instead.
* `SwitchToMainThreadAsync` is *allowed* to complete successfully even if cancellation is requested if _any_ of certain conditions are met:
    1. The code is already on the UI thread
    2. The UI thread becomes available _earlier_ than a thread in the thread pool becomes available to process the cancellation request

The explicit checking of the cancellation token after `SwitchToMainThreadAsync` is often optional, but if you want to guarantee execution stops when cancellation is requested, you should add it since `SwitchToMainThreadAsync` does not make this guarantee.

💡 Note that our own extension method `SwitchToMainThreadAsync` *does* guarantee that cancellation has not been requested.

https://github.com/gitextensions/gitextensions/blob/96090c8d0114397fb0bcfbbb342261d1da45e904/GitExtUtils/GitUI/ControlThreadingExtensions.cs#L128-L131

➡️ Since you are using the extension method for UI controls here (`this.SwitchToMainThreadAsync`), you do not need to explicitly check cancellation. You should still pass the cancellation token to `SwitchToMainThreadAsync` though.

> would calling cancellationToken.ThrowIfCancellationRequested() not result in an exception to be thrown on UI thread and thus crash the app?

No, the exception would still get caught and wrapped in the `Task`.

RussKie: Thank you 👍

GitUI/CommandsDialogs/FormCommit.cs
RussKie: I think I'm going mental... Clearly I'm doing something wrong.

I've tried changing the main code back to use `ThreadHelper.JoinableTaskFactory.RunAsync().FileAndForget()` and then changed the test code to use `ThreadHelper.JoinableTaskFactory.Run(() => ThreadHelper.JoinPendingOperationsAsync())`. 
```diff
diff --git a/GitUI/CommandsDialogs/FormCommit.cs b/GitUI/CommandsDialogs/FormCommit.cs
index ac0526643..3ff1d4162 100644
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -427,5 +427,5 @@ protected override void OnShown(EventArgs e)
 
             var token = _updateBranchNameDisplaySequence.Next();
-            _joinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(token)).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(token)).FileAndForget();
             UpdateAuthorInfo();
 
@@ -1195,5 +1195,5 @@ private void ShowChanges(GitItemStatus item, bool staged)
             }
 
-            _joinableTaskFactory.RunAsync(async () =>
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
                 await SetSelectedDiffAsync(item, staged);
@@ -3229,5 +3229,5 @@ private void createBranchToolStripButton_Click(object sender, EventArgs e)
 
             var token = _updateBranchNameDisplaySequence.Next();
-            _joinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(token)).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync(token)).FileAndForget();
         }
``` 

This way the background tasks should be added to the main collection and we could await all of them. Or so I thought.

However I get a deadlock in `UpdateBranchNameDisplayAsync_should_display_branch_and_no_remote_info_in_statusbar` test.
```diff
diff --git a/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs b/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
index 121c04a74..a8b961c9f 100644
--- a/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
+++ b/UnitTests/GitUITests/CommandsDialogs/CommitDialog/FormCommitTests.cs
@@ -96,5 +96,5 @@ public void UpdateBranchNameDisplayAsync_should_display_branch_and_no_remote_inf
             RunFormTest(form =>
             {
-                ThreadHelper.JoinableTaskFactory.Run(() => form.GetTestAccessor().RunningTasks.JoinTillEmptyAsync());
+                ThreadHelper.JoinableTaskFactory.Run(() => ThreadHelper.JoinPendingOperationsAsync());
 
                 var currentBranchNameLabelStatus = form.GetTestAccessor().CurrentBranchNameLabelStatus;
```
And I can't seem to figure out what causes it...

@AArnott is there a way to see what tasks are in a collection? I've tried going through a debug window but I couldn't see anything meaningful.

AArnott: `ThreadHelper.JoinableTaskFactory.Run(() => ThreadHelper.JoinPendingOperationsAsync());` will probably deadlock all on its own if you try it. Assuming `ThreadHelper.JoinPendingOperationsAsync()` calls `JoinTillEmptyAsync()` on its associated collection, you're waiting on yourself to finish, which will never happen. You need to be careful to _not_ use a JTF instance to wait for completion of its own collection because the waiting is itself in that same collection. Try this instead:

`ThreadHelper.JoinableTaskContext.Factory.Run(() => ThreadHelper.JoinPendingOperationsAsync());`

sharwell:
@RussKie I can work on this tomorrow

RussKie:
That'd much appreciated, it feels like I'm running in circles...
And it probably something simple that I can't see

PR #6287, opened 2019-02-21
Don't load avatars, if network is not available

### LINKED ISSUE #5828: Author's avatar settings are not working correctly in Version 3.00.00 RC2
bug

Under Settings -> GitExtensions -> Appearance -> Author images 
"No Image Service" is set to "None" and the checkbox "Show author's avatar in the commit info view" ticked off.
But the avatar images get fetched (in wireshark I see GET requests to gravatar.com)
Also the clear cache button seem not properly working.
From some users like Janusz I see his image in every settings combination

Set the settings as described above and open a repo (for example the gitextensions repo).

If the checkbox "Show author's avatar ..." is not set and Image service is set to none I expect no avatar image (and no outgoing network requests) at all.
In V3 I also find the labeling of the UI problematic, gravatar is not named anymore and the drop down labeled "No Image Service" is kinda confusing.

- GitExtensions version: 3.00 RC2
- GIT version: 2.19.2.windows.1
- OS version: Win 10 Pro 64 
- .NET version: 4.7.3190

I use the portable version (GitExtensions.exe.config => IsPortable: True) with GitExtensions.settings in the app folder

2.51.05

### LINKED ISSUE #6223: Slow startup if never connected to internet
## Current behaviour
With a clean install, accepting all default options of GitExtensions-3.0.1.4933.msi but NO internet connection, after installation the program takes 30+ seconds to startup each time (i.e. double-click icon, 30+ seconds before first window appears) This is before opening any repositories, just starting the program.

If the machine is connected to the internet ONCE and it is started, program takes less than 3 seconds.  Disconnecting from internet again, program always starts up quickly (<3 seconds).

## Expected behaviour
Once installation is complete, program should not require internet connection to 'finish' some final configuration or installation.

## Steps to reproduce
Copy msi file to machine that has no internet connectivity.
Run installation accepting all default settings
After installation, starting always takes 30+ seconds to open.
Close program, connect to internet and then start program.

Sorry, don't know, haven't used previous versions.

## Environment

- Git Extensions 3.0.1.4933
- Build a269358b22a5dc1ea36928bb87f9e535de76a058
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.14393.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

RussKie:
Fixes #5828
Fixes #6223

## Proposed changes

- Check whether a client is connected to Internet before attempting to
fetch avatars.
- Add tooltip to clarify purpose of "No image" dropdown
image
- Add a setting to show author's avatar column
image

## Test methodology 

- manual
----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
@@ -52,7 +52,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
 
             chkShowCurrentBranchInVisualStudio.Checked = AppSettings.ShowCurrentBranchInVisualStudio;
             _NO_TRANSLATE_DaysToCacheImages.Value = AppSettings.AvatarImageCacheDays;
-            ShowAuthorAvatar.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
+            ShowAuthorAvatarInCommitInfo.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
             ShowAuthorAvatarInCommitGraph.Checked = AppSettings.ShowAuthorAvatarColumn;
             NoImageService.Text = AppSettings.GravatarDefaultImageType.ToString();
 
@@ -109,7 +109,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
             AppSettings.TruncatePathMethod = GetTruncatePathMethodString(truncatePathMethod.SelectedIndex);
             AppSettings.ShowCurrentBranchInVisualStudio = chkShowCurrentBranchInVisualStudio.Checked;
             AppSettings.ShowAuthorAvatarColumn = ShowAuthorAvatarInCommitGraph.Checked;
-            AppSettings.ShowAuthorAvatarInCommitInfo = ShowAuthorAvatar.Checked;
+            AppSettings.ShowAuthorAvatarInCommitInfo = ShowAuthorAvatarInCommitInfo.Checked;
             AppSettings.AvatarImageCacheDays = (int)_NO_TRANSLATE_DaysToCacheImages.Value;
 
             AppSettings.Translation = Language.Text;
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
@@ -12,6 +12,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
     {
         private readonly TranslationString _noDictFile = new TranslationString("None");
         private readonly TranslationString _noDictFilesFound = new TranslationString("No dictionary files found in: {0}");
+        private readonly TranslationString _noImageServiceTooltip = new TranslationString("A default image, if an email address has no matching Gravatar image.\r\nSee http://en.gravatar.com/site/implement/images/ for more details.");
 
         public AppearanceSettingsPage()
         {
@@ -24,6 +25,8 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         protected override void OnRuntimeLoad()
         {
             base.OnRuntimeLoad();
+
+            ToolTip.SetToolTip(NoImageService, _noImageServiceTooltip.Text);
 
             // align 1st columns across all tables
             tlpnlGeneral.AdjustWidthToSize(0, truncateLongFilenames, lblCacheDays, lblNoImageService, lblLanguage, lblSpellingDictionary);
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
@@ -53,6 +53,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
             chkShowCurrentBranchInVisualStudio.Checked = AppSettings.ShowCurrentBranchInVisualStudio;
             _NO_TRANSLATE_DaysToCacheImages.Value = AppSettings.AvatarImageCacheDays;
             ShowAuthorAvatar.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
+            ShowAuthorAvatarInCommitGraph.Checked = AppSettings.ShowAuthorAvatarColumn;
             NoImageService.Text = AppSettings.GravatarDefaultImageType.ToString();
 
             Language.Items.Clear();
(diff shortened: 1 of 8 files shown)


gerhardol:
(will review tonight, it may be correct as it is, I looked at this some time ago)

gerhardol:
(No time for test tonight.)
Gravatar service None means that no backup service exists, not that there are no icons.
GE first tries GitHub for github addresses, this change should disable that too.

This setting should be clarified, there were some discussions in a previous issue.

RussKie:
Actually you're absolutely correct. I had a mental eclipse last night.
Thank you

We need to use the switch that controls the display of the avatar column.
We also need an extra pivot point inside the code that checks a network
connectivity (this would also address start timing issue if not connected
to a network).

RussKie:
Change of plan.
Avatars got fetched from the internet, if absent locally. And if a client isn't connected to internet, these calls time out, and that causes slowness.

I've tested the fix locally and observed a significant improvement in responsiveness.

RussKie:
@gerhardol  could you please test the settings page at 100% scaling?

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -104,6 +108,7 @@ protected override void PageToSettings()
             AppSettings.EnableAutoScale = chkEnableAutoScale.Checked;
             AppSettings.TruncatePathMethod = GetTruncatePathMethodString(truncatePathMethod.SelectedIndex);
             AppSettings.ShowCurrentBranchInVisualStudio = chkShowCurrentBranchInVisualStudio.Checked;
+            AppSettings.ShowAuthorAvatarColumn = ShowAuthorAvatarInCommitGraph.Checked;
```
gerhardol: Minor: Can we move this setting together with the existing setting? (easier to review)

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -50,6 +53,7 @@ protected override void SettingsToPage()
             chkShowCurrentBranchInVisualStudio.Checked = AppSettings.ShowCurrentBranchInVisualStudio;
             _NO_TRANSLATE_DaysToCacheImages.Value = AppSettings.AvatarImageCacheDays;
             ShowAuthorAvatar.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
```
gerhardol: This setting should be renamed (not related to this PR)

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -25,6 +26,8 @@ protected override void OnRuntimeLoad()
         {
             base.OnRuntimeLoad();
 
+            ToolTip.SetToolTip(NoImageService, _noImageServiceTooltip.Text);
```
gerhardol: It would be good if the link was clickable, but no obvious good way to do that

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -12,6 +12,7 @@ public partial class AppearanceSettingsPage : SettingsPageWithHeader
     {
         private readonly TranslationString _noDictFile = new TranslationString("None");
         private readonly TranslationString _noDictFilesFound = new TranslationString("No dictionary files found in: {0}");
+        private readonly TranslationString _noImageServiceTooltip = new TranslationString("A default image, if an email address has no matching Gravatar image.\r\nSee http://en.gravatar.com/site/implement/images/ for more details.");
```
gerhardol: Not just gravatar, GitHub (may be) checked first

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -50,6 +53,7 @@ protected override void SettingsToPage()
             chkShowCurrentBranchInVisualStudio.Checked = AppSettings.ShowCurrentBranchInVisualStudio;
             _NO_TRANSLATE_DaysToCacheImages.Value = AppSettings.AvatarImageCacheDays;
             ShowAuthorAvatar.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
+            ShowAuthorAvatarInCommitGraph.Checked = AppSettings.ShowAuthorAvatarColumn;
```
gerhardol: I would have preferred that this setting was needed here, but this PR is an improvement. 
A tooltip for the CommitInfo setting is an alternative

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
RussKie: Sure, np

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
RussKie: Sure, np

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
RussKie: The default image comes from Gravatar though, if we can't get an image from GH or GR.

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
RussKie: Apologies, I don't understand what you're saying

gerhardol: Related: This setting should have been called ShowAuthorAvatarInRevisionGraph

The columns in the revision graph is normally not set from settings. The existing (now renamed) ShowAuthorAvatarInCommitInfo should have been moved out of here. A "block all avatars" would have been cleaner.

I do not request changes though

RussKie:
We can look at it as a separate piece of work

robot1marvin:
Tried the zip file you emailed containing Git Exteions 3.1.0.5347 and it is much faster.

However, each time I open it I get a popup with an exception.  I couldn't send that directly, but I screen-shot the dialog along with the stack-trace window.
gitextportexception

RussKie:
That's a different error.
Please see #6093

Garreye:
I'm still having slow loading issue on PCs that have no internet access, but are still connected to a LAN. If I physically pull out the Ethernet connection to my PC then Git Extensions opens in about 2 seconds, but if Ethernet is plugged in it takes about 32 seconds. 

Looks like the check for InternetGetConnectedState method from wininet.dll returns true if there is a LAN connection present, even though there's no internet. From googling around a bit it looks like this method is not very reliable and this behavior is expected.

Is there a setting to turn off avatar loading completely? I have the avatar provider set to 'Author initials' which I though might do it, but didn't fix the issue.

Garreye:
The slow loading issue I'm having is happening on installed version 3.4.3.  I've now checked out the release/3.4 branch and built release binaries from there and run these binaries on one of the PCs where I was seeing this issue it loads in 2 seconds. I overwrote the installed C;/Program Data (x86)/GitExtensions/GitExtensions.exe with the one I built and it works fine from there. The about window for this has the same build number as the installed version, but says versions 3.3 for some reason. 

I don't have VS configure to be able to build the setup, so I'm not able to try installing the build I did, but maybe something related to installing? I built with VS Community 2019 (16.6.4).

gerhardol:
> I overwrote the installed C;/Program Data (x86)/GitExtensions/GitExtensions.exe with the one I built and it works fine from there. 

Just run from the build folder or run in the debugger, you only need to install for registering file extensions and the shell extensions, compare to the portable build

There are other reasons to access internet like update check, telemetry and build server integration.

Garreye:
Ya I'm fine with running from the build I did. I just thought that it was weird that the build I did behaves differently even though it's built from the same code as far as I can tell (same build ID in about window).

Portable build takes the same 32 seconds to open as installed.

PR #6116, opened 2019-01-16
CI integration in revision grid adjustments

pmiossec:
## Proposed changes

- Update AppVeyor display string to display the more important information first
- Update AzureDevops display string to display the more important information first
- Increase the width of the "Build status text" column (because we can't resize it)

## Screenshots 

AppVeyor:
### Before

image

### After

image

AzureDevops:
### Before

image

### After

image

## Test methodology 

- Running the application (see screenshots)

## Test environment(s) 

- GIT 2.20
- Windows 10

PR #6292, opened 2019-02-21
Clean and unify build status label and tooltip

pmiossec:
Follow up to #6116

## Proposed changes

- Remove sometimes empty line in tooltip
- Display duration in the same format in all CI plugins
- other small tweaks

### Before

I will try to do it later... (if needed)

### After

Appveyor:
image


Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
```diff
@@ -386,13 +386,13 @@ private void UpdateDescription(AppVeyorBuildInfo buildDetails, CancellationToken
             {
                 int failedTestCount = buildDescription["failedTestsCount"].ToObject<int>();
                 int skippedTestCount = testCount - buildDescription["passedTestsCount"].ToObject<int>();
-                var testResults = " : " + testCount + " tests";
+                var testResults = testCount + " tests";
```
gerhardol: This should be a translatable string, I let this pass though...

PR #6293, opened 2019-02-21
Move prune/prune all/update in Leftpanel remotes contextual menu

pmiossec:
Fixes #6284

## Proposed changes

- Add "prune remote" in contextual menu of a remote in left panel (moved from FormRemote)
- Add "Update all remotes" in contextual menu of remotes in left panel (moved from FormRemote)
- Add "Update and prune all remotes" in contextual menu of remote in left panel (because just "prune all" like requested in #6284  is not something provided by git and I don't want to do it myself)
- Remove "Prune remote branches" and "Update all remote branch info" from FormRemotes

## Screenshots 

Left panel (enabled only) remote:
### Before
image

### After
image

Left panel remotes:
### Before

image

### After

image

### Before
image

### After
image

## Test methodology 

- Manual 

## Test environment(s) 

- Git Extensions 3.1.0
- Build debc4f89828705b57937e7a11e8f65ac3517f8c0 (Dirty)
- Git 2.20.0.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3324.0
- DPI 192dpi (200% scaling)

Issue #6200, opened 2019-02-01
QuickPull's hotkey is not working


glacjay:
## Current behaviour

I've reset all hotkeys to defaults, then press QuickPull's hotkey ctrl-shift-p, but the command been executed is `git fetch`, not `git pull`.

## Expected behaviour

`git pull` should be executed when I press ctrl-shift-p.

## Steps to reproduce

Just reset all hotkeys to default, and press ctrl-shift-p.

## Environment

- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

RussKie:
Would you like to try this build https://ci.appveyor.com/api/buildjobs/iona492quavhndbg/artifacts/GitExtensions-Portable-3.0.2.5120.zip and see if it is fixed

RussKie:
@glacjay could you please test this build
https://ci.appveyor.com/api/buildjobs/nadckhnva1hy9ljc/artifacts/GitExtensions-Portable-3.0.2.5223.zip

Issue #6194, opened 2019-01-31
Memorize last pull mode option in pull dialog


osevoso:
**Feature request**

Add programm option for memorize last options state in pull-dialog. As least for pull-modes.

In main window, starting with v3.00.00, pull-actions may to run immediatly, without possibility to customize it, but it is not covering all needs for usage in all cases. 

Setting default behavior for Pull-button (in main window) for opening pull-dialog does correcting the situation, but always impose "fetch mode" for pulling.

I mean that "checking and running" instead of "checking, correct and running" are more usability, when immediate runs of predefined pull-mode's is not suitable.

- Git Extensions 3.0.1.4933
- Build a269358b22a5dc1ea36928bb87f9e535de76a058
- Git 2.20.0.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

RussKie:
I think this is duplicate of #6144
Please have a look at #6153 and #6155.

osevoso:
Yes, this is close to a duplicate, the pull mechanics became very uncomfortable and not intuitive in 3.x compare to 2.51 Workaround, which i found for me - always to open pull-dialog. 
Therefore, the proposal - without changing what is now, add a global software option to remember the last pull-action in this dialog to restore the expected behavior close to was at 2.51.

RussKie:
Would you like to try this build https://ci.appveyor.com/api/buildjobs/iona492quavhndbg/artifacts/GitExtensions-Portable-3.0.2.5120.zip and see if it is fixed

osevoso:
This build have a other extreme - no one pull-mode option are pre-selected when pull-dialog is openining every time.

RussKie:
Please elaborate and provide more info.

I tested a brand new installation (v3.0.2.5120) on a brand new VM, invoked "Pull" from a Windows Explorer context menu:
![image](https://user-images.githubusercontent.com/4403806/52198643-9971a980-28b7-11e9-8e00-c4d7fbed8763.png)

vbjay:
No. It uses the set default in the pull menu in toolbar.

osevoso:
> No. It uses the set default in the pull menu in toolbar.

vbjay, you are right. And it too complex for use.

Why not separate opening of pull dialog by corresponding button from influence of other buttons?
The next behavior may be well: 

- Main pull-button on toolbar do action depending from user-selected default action;
[...]

PR #6301, opened 2019-02-22
Automatically load PuTTY SSH key files when needed in left panel

mast-eu:
This is a follow-up of #6293 and needs to be rebased before it can be merged.

## Proposed changes

- copy `LoadPuttyKey()` from `FormPull.cs` into `RepoObjectsTree.cs` in order to automatically load SSH keys (if available) when `Update all remotes`, `Update and prune all remotes` or `Prune remote` is clicked in the left panel (#6293 adds these to the left panel context menu)

## Test methodology

- manual tests
- ran unit tests

## Test environment(s) 

- Git Extensions 3.00.00.4433
- Git Extensions 3.1.0
- Build 5cc1e8e40e59ce000aba01ce5351ebeb01343144
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 120dpi (125% scaling)

----


RussKie:
Could you please rebase and fix the MC

mast-eu:
> Could you please rebase and fix the MC

done

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -472,17 +474,72 @@ public TestAccessor(RepoObjectsTree repoObjectsTree)
...
 
         private void mnuBtnPruneAllRemotes_Click(object sender, EventArgs e)
         {
+            LoadPuttyKey();
             FormRemoteProcess.ShowDialog((GitModuleForm)ParentForm, "remote update --prune");
         }
 
         private void mnuBtnPrune_Click(object sender, EventArgs e)
         {
+            LoadPuttyKey();
```
RussKie: `LoadPuttyKey` is run before each `FormRemoteProcess.ShowDialog` thus indicating it is a concern better placed with `FormRemoteProcess`.
It also doesn't seem correct to make the left panel responsible for loading or otherwise managing SSH keys.

`FormRemoteProcess` already has some functionality that looks like it interacts with Plink, e.g.
https://github.com/gitextensions/gitextensions/blob/027eabec3be497f8d780cc68ece268c64a43a7d5/GitUI/FormRemoteProcess.cs#L145-L153

pmiossec: I also think that we should not duplicate `LoadPuttyKey()` code and do a refactoring to prevent it....

mast-eu: Actually, there is already a good working implementation in place for "Fetch" (i.e. the option that was already implemented before #6293):



"Fetch" is connected to:
https://github.com/gitextensions/gitextensions/blob/96090c8d0114397fb0bcfbbb342261d1da45e904/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs#L349-L357

Which uses the pull dialog, including it's `LoadPuttyKey()`:
https://github.com/gitextensions/gitextensions/blob/96090c8d0114397fb0bcfbbb342261d1da45e904/GitUI/CommandsDialogs/FormPull.cs#L739

I propose to extend the same mechanism for all newly added options. I have to modify  @pmiossec's implementation, but I think it's worth it. No code duplication and all options to fetch/update/prune remotes from the left panel will work the same way.

mast-eu:
## Proposed changes

- Left panel remotes context menu: changed "Update all remotes" to "Fetch all remotes" (both text and git command); added icon (same as corresponding pull dialog shortcut)
- Left panel remotes context menu: changed "Update and prune all remotes" to "Fetch and prune all remotes" (both text and git command); added icon (same as corresponding pull dialog shortcut)
- Left panel remote context menu: changed icon for "Fetch" (same as corresponding pull dialog shortcut)
- Left panel remote context menu : changed "Prune remote" to "Fetch and prune" (both text and git command); created new icon, similar to the other ones
- all above actions now use the existing pull dialog, thus avoid code duplication and keep logic for handling SSH keys away from the left panel
- refactoring of the MessageBox "Warning! The fetch with prune will remove all the remote-tracking references which no longer exist on remotes. Do you want to proceed?", in order to show it also when pruning is triggered from the left panel (via "Fetch and prune all remotes" or "Fetch and prune")


## Screenshots

### Before
image

 image

oldmessage

### After
contextall

contextsingle

newmessageall

newmessagesingle


## Test methodology

- manual tests
- ran unit tests


## Test environment(s) 

- Git Extensions 3.1.0
- Build 34bc9550a47a3c50aacd4b5faf6352719240efcd
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 120dpi (125% scaling)


----

mast-eu:
@pmiossec  I changed quite a bit what you have implemented in #6293. Could you please have a look?

GitCommands/Settings/AppSettings.cs
```diff
@@ -691,6 +691,7 @@ public enum PullAction
             Merge,
             Rebase,
             Fetch,
+            FetchPrune,
```
RussKie: ❗️ This will have a profound effect on the default pull action dropdown. 
We can't do that.

image

GitUI/CommandsDialogs/FormPull.cs
```diff
@@ -97,6 +97,8 @@ public sealed partial class FormPull : GitModuleForm
         private readonly TranslationString _formTitleFetch = new TranslationString("Fetch ({0})");
         private readonly TranslationString _buttonPull = new TranslationString("&Pull");
         private readonly TranslationString _buttonFetch = new TranslationString("&Fetch");
+
+        private readonly TranslationString _pullFetchPruneAllConfirmation = new TranslationString("Warning! The fetch with prune will remove all the remote-tracking references which no longer exist on remotes. Do you want to proceed?");
```
RussKie: The display of the warning must be configurable
https://github.com/gitextensions/gitextensions/blob/f6929f477c8194dcf3ad3179df3572b9029bf18a/GitUI/CommandsDialogs/FormBrowse.cs#L2421-L2448

mast-eu: Display of the warning is controlled by `AppSettings.DontConfirmFetchAndPruneAll`, which can be set in `Settings` > `Advanced` > `Confirmations`. I didn't change anything in this regard.

fetchpruneall

Thinking about it, maybe the text should be changed to something like " Fetch and prune branches", because now it applies also to a single repo and no longer necessarily to all repos at once?

GitCommands/Settings/AppSettings.cs
mast-eu: I didn't find any problem (yet). What I tested so far:
1) set all available default pull actions and verify that they are saved correctly in GitExtensions.settings item `DefaultPullAction` -> ok
2) set all available default pull actions and verify that they are applied correctly after clicking `Open pull dialog ...` and the dialog is setup correctly -> ok
3) trigger all available options from the pull dialog dropdown and verify that they execute the expected git command -> ok
4) trigger all available options from the left panel and verify that they execute the expected git command -> ok
5) Extended unit tests `Should_correctly_setup_for_defined_pull_action` and `Should_use_user_DefaultPullAction_pull_action_None` wich cases for `PullAction.FetchPrune` -> ok

Do I miss something?

Would it be better to add the new "Fetch and prune" to the default pull action dropdown as well?

RussKie: These options are also used when the app is invoked via the shell extensions.
We had a number of issues associated with this area, e.g. #6144, #6194, #6200; and I wouldn't want a repeat of these...

GitUI/CommandsDialogs/FormPull.cs
RussKie: Confirm that it is working as intended, thank you

RussKie: > maybe the text should be changed to something like "Fetch and prune branches",

Sounds good, the suggestion is better than the existing text.

GitUI/CommandsDialogs/FormPull.cs
```diff
@@ -216,8 +222,35 @@ private void BindRemotesDropDown(string selectedRemoteName)
...
+            // Special case for FetchPrune and FetchPruneAll to make sure user confirms the action.
+            string messageBoxTitle = null;
+            if (pullAction == AppSettings.PullAction.FetchPrune)
+            {
+                messageBoxTitle = string.Format(_pruneFromCaption.Text, remote);
+            }
+
+            if (pullAction == AppSettings.PullAction.FetchPruneAll)
+            {
+                messageBoxTitle = string.Format(_pruneFromCaption.Text, AllRemotes);
```
RussKie: Can we express the difference under `PullAction.FetchPruneAll` but supply `AllRemotes` if `remote == null`?

mast-eu: Sorry, I'm not sure if I understand you correctly. Do you want to simply improve the readability of this method?

Or do you want to eliminate `PullAction.FetchPrune` by using `PullAction.FetchPruneAll` for both operations "Fetch and prune" and "Fetch and prune all"? Something like:
```
    if (pullAction == AppSettings.PullAction.FetchPruneAll)
    {
        if (remote.IsNullOrEmpty())
        {
            // Fetch and prune AllRemotes
        }
        else
        {
            // Fetch and prune remote
        }

```

RussKie: Apologies for being unclear.
I'd like to eliminate `FetchPrune` altogether, the implementation looks similar to `FetchPruneAll`.

mast-eu: Indeed, they are very similar. I can make it work without `FetchPrune`.

GitCommands/Settings/AppSettings.cs
mast-eu: I force pushed changes to remove `PullAction.FetchPrune`.

GitUI/CommandsDialogs/FormPull.cs
mast-eu: Done, please have a look.

RussKie:
@pmiossec are you happy with the changes?

pmiossec:
>@pmiossec are you happy with the changes?

Yes. The only thing that we could improve is that we could remove the last use of `FormRemoteProcess.ShowDialog(this, "remote update");` in 
https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormRemotes.cs#L450
to also use `UICommands.StartPullDialogAndPullImmediately()`.

mast-eu:
> The only thing that we could improve is that we could remove the last use of `FormRemoteProcess.ShowDialog(this, "remote update");` in
> https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormRemotes.cs#L450
> to also use `UICommands.StartPullDialogAndPullImmediately()`.

Good point. I will add this.

mast-eu:
I added a commit implementing @pmiossec's suggestion.

RussKie:
Thank you

PR #6316, opened 2019-02-27
Disable the display of notes Dialog after Ctrl+Shift+N for artificial…

### LINKED ISSUE #6308: Crash on Ctrl+Shift+N for artificial commits
## Current behaviour

Selecting Working Directory and hitting Ctrl+Shift+N crashes gitextensions.

## Expected behaviour

No crash and either add note or not add note dialog.

## Steps to reproduce

In Commit tree
Select working directory
Press Ctrl+Shift+N

Crash dialog comes up.

Not sure.

## Environment

- GitExtensions version: 3.0.2
- GIT version: 2.21.0
- OS version: Windows 10 1809
- .NET version: Not sure

ivangrek:
Fixes #6308

## Test methodology 

- Pressed the key combination Ctrl+Shift+N

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2156,7 +2156,7 @@ namespace GitUI.CommandsDialogs
         {
             var objectId = RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId;
 
-            if (objectId == ObjectId.IndexId || objectId == ObjectId.WorkTreeId)
+            if (objectId == null || objectId.IsArtificial)
             {
                 return;
             }
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2154,7 +2154,14 @@ namespace GitUI.CommandsDialogs
 
         private void AddNotes()
         {
-            Module.EditNotes(RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId);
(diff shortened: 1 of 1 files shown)


gerhardol:
Can you please sign first, preferable as a separate PR?
https://github.com/gitextensions/gitextensions/blob/master/contributors.txt

(wiki must be updated)

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2154,7 +2154,14 @@ internal Keys GetShortcutKeys(Command cmd)
 
         private void AddNotes()
         {
-            Module.EditNotes(RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId);
+            var objectId = RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId;
+
+            if (objectId == ObjectId.IndexId || objectId == ObjectId.WorkTreeId)
```
gerhardol: Should use objectId.IsArtificial

RussKie: :warning: `objectId` can be `null`

ivangrek:
In what cases objectId can be null?

RussKie:
> In what cases objectId can be null?

`RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId` is a nullable result.

ivangrek:
Maybe it requires an exception call?

vbjay:
Nope. Why use an exception for program flow?

https://docs.microsoft.com/en-us/dotnet/api/system.nullable-1.hasvalue

RussKie:
Thank you

PR #6321, opened 2019-02-28
Preserve original byte order mark when editing files

sharwell:


sharwell:
> Is it feasible to have some tests around it?

That's my hope! I was also hoping to use coverage to see what's already covered, but that appears to be none of it 😆

sharwell:
@RussKie Tests are added. Took me a few tries to get them just right. I didn't add complete tests for this form but I added a couple and covered both paths of the new functionality.

RussKie:
> I didn't add complete tests for this form but I added a couple and covered both paths of the new functionality.

It would be totally unreasonable to expect to have full test coverage. But it is good to have tests for the code that changed.
Thank you

sharwell:
@RussKie this code _may_ increase flakiness because it exacerbates a race condition for file save operations, but it's being addressed by #6327.

PR #5617, opened 2018-10-24
Feature: add a Submodules tree to the left panel

amaiorano:
Changes proposed in this pull request:
- Add a Submodules tree to the left panel
- This tree offers the same functionality as the Submodule menu, but also adds more:
  - Context-menu actions on specific submodules (e.g. you can update any submodule, not just the current)
  - Double-click a submodule to open it
  - Possibility for more context-based actions in the future

This is related to the discussion here: https://github.com/gitextensions/gitextensions/issues/5569
 
Screenshots before and after (if PR changes UI):

(Edit: screenshot of the entire window):
image

The main view:
image

Context menu on the top node to update all, go to super or top project:
image

Context menu on a sibling submodule node that allows us to open it (can also double-click), and to update it, even if it's not current:
image

When there are no submodules, we display the top node with the count at 0:
image

What did I do to test the code and ensure quality:

I mainly tested within the gitextensions repo, and added a submodule to one of the existing submodules to make sure nested submodules worked. However, if someone can point me to a repo with many more submodules, especially deeply nested, it would be nice to test that too.

I also tested a repo with no submodules.

Has been tested on (remove any that don't apply):
- GIT 2.9.0.windows.1
- Windows 10

PR #6325, opened 2019-03-01
Throw OperationCanceledException on cts.Next() when cts is disposed.

jbialobr:
CancellationTokenSequence is written in a thread-safe manner but it can not be accessed when it is disposed. To properly deal with this, any async call that can access cts would have to be tracked and awaited before disposing cts. 
I propose to throw OperationCanceledException instead of ObjectDisposedException
so it is possible to access cts from a Task without worrying of cts being disposed or not.

The problem occurred in https://github.com/gitextensions/gitextensions/pull/5617

### DIFF
--- a/GitExtUtils/GitUI/CancellationTokenSequence.cs
+++ b/GitExtUtils/GitUI/CancellationTokenSequence.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading;
 
 namespace GitUI
@@ -46,7 +46,7 @@ namespace GitUI
         /// <para>This method is thread-safe.</para>
         /// </remarks>
         /// <returns>A <see cref="CancellationToken"/> to be used by the commencing asynchronous operation.</returns>
-        /// <exception cref="ObjectDisposedException">This object is disposed.</exception>
+        /// <exception cref="OperationCanceledException">This object is disposed.</exception>
         public CancellationToken Next()
         {
             var next = new CancellationTokenSource();
@@ -76,8 +76,9 @@ namespace GitUI
             // Detect and handle the case where the sequence was already disposed
             if (prior is null)
             {
+                next.Cancel();
                 next.Dispose();
-                throw new ObjectDisposedException(nameof(CancellationTokenSequence));
+                nextToken.ThrowIfCancellationRequested();
             }
 
             prior.Cancel();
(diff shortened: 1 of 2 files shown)


GitExtUtils/GitUI/CancellationTokenSequence.cs
```diff
@@ -28,6 +28,17 @@ namespace GitUI
     /// </example>
     public sealed class CancellationTokenSequence : IDisposable
     {
+        private static readonly CancellationToken _preCancelledToken;
```
sharwell: 📝  This is not necessary

GitExtUtils/GitUI/CancellationTokenSequence.cs
```diff
@@ -77,7 +88,7 @@ public CancellationToken Next()
             if (prior is null)
             {
                 next.Dispose();
-                throw new ObjectDisposedException(nameof(CancellationTokenSequence));
+                _preCancelledToken.ThrowIfCancellationRequested();
```
sharwell: 💡 Either of the following also works:

```csharp
next.Cancel();
next.Dispose();
nextToken.ThrowIfCancellationRequested();
```

```csharp
next.Dispose();
new CancellationToken(true).ThrowIfCancellationRequested();
```

jbialobr: I was considering the first option, but thought it would be better to have a precancelled token to not have to call `Cancel` each time. Given that it should not be a common case where Next() is extensively called for a disposed cts it should not be a problem to go for the first option you proposed.

jbialobr:
Revert this commit in my branch and see half of the UI unit tests failing.
The reason is here
https://github.com/amaiorano/gitextensions/pull/1/files#diff-46207722e81d5ab6eb64272f0aab6ec2R210
where an async Tasks from SubmoduleStatusProvider  invokes ReloadNodesAsync and hits the cts being disposed.
The bug in the code that comes to my mind is that when we stop the GE we do not  cancel and await all the background/async jobs.

jbialobr:
Here are the failing tests:
https://ci.appveyor.com/project/jbialobr/gitextensions/build/3.1.0.575/tests

sharwell:
Thanks! It looks like there isn't a clean way to get the next token from a background thread unless this is changed, so it makes sense to me.

GitExtUtils/GitUI/CancellationTokenSequence.cs
sharwell: The performance really won't matter either way because there won't be any listeners registered for the cancellation event.

PR #6330, opened 2019-03-01
Add depth option to GitHub Fork/Clone form

### LINKED ISSUE #3517: Need -depth option on clone github dialog
Cloning a repo now has the option to restrict history (-depth) - it would be awesome if this same capability was available when cloning a github repo .

oriash93:
Fixes #3517

## Proposed changes

- Added an option to restrict history (`--depth`) when cloning from GitHub.
- The `--depth <depth>` flag will only be added to the `git clone` command if the checkbox was checked and a valid number was entered in the textbox.

### Before
before

### After

after

## Test methodology 

- Cloned a repo from GitHub using the new feature, ensured `git clone` was commanded with the correct value in the flag.
- Ensured no flag was added when the checkbox is unchecked, or if the textbox contains no valid number.

----

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.Designer.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.Designer.cs
@@ -41,7 +41,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
             this.myReposPage = new System.Windows.Forms.TabPage();
             this.tableLayoutPanel5 = new System.Windows.Forms.TableLayoutPanel();
             this.helpTextLbl = new System.Windows.Forms.Label();
-            this.myReposLV = new UserControls.NativeListView();
+            this.myReposLV = new GitUI.UserControls.NativeListView();
             this.columnHeader2 = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.searchReposPage = new System.Windows.Forms.TabPage();
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
@@ -52,13 +52,16 @@ namespace GitUI.CommandsDialogs.RepoHosting
             this.getFromUserBtn = new System.Windows.Forms.Button();
             this.forkBtn = new System.Windows.Forms.Button();
             this.tableLayoutPanel3 = new System.Windows.Forms.TableLayoutPanel();
-            this.searchResultsLV = new UserControls.NativeListView();
+            this.searchResultsLV = new GitUI.UserControls.NativeListView();
             this.columnHeader6 = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.tableLayoutPanel4 = new System.Windows.Forms.TableLayoutPanel();
             this.openGitupPageBtn = new System.Windows.Forms.Button();
             this.searchResultItemDescription = new System.Windows.Forms.TextBox();
             this.descriptionLbl = new System.Windows.Forms.Label();
             this.cloneSetupGB = new System.Windows.Forms.GroupBox();
+            this.depthLabel = new System.Windows.Forms.Label();
+            this.ProtocolDropdownList = new System.Windows.Forms.ComboBox();
+            this.ProtocolLabel = new System.Windows.Forms.Label();
             this.cloneInfoText = new System.Windows.Forms.Label();
             this.addRemoteAsTB = new System.Windows.Forms.TextBox();
             this.label3 = new System.Windows.Forms.Label();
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.Designer.cs
```diff
@@ -507,36 +561,19 @@ private void InitializeComponent()
...
-            this.ProtocolDropdownList.Location = new System.Drawing.Point(470, 32);
-            this.ProtocolDropdownList.Name = "ProtocolDropdownList";
-            this.ProtocolDropdownList.Size = new System.Drawing.Size(121, 21);
-            this.ProtocolDropdownList.TabIndex = 2;
-            this.ProtocolDropdownList.SelectedIndexChanged += new System.EventHandler(this.ProtocolSelectionChanged);
-            // 
             // ForkAndCloneForm
             // 
-            this.AutoScaleDimensions = new System.Drawing.SizeF(96F, 96F);
+            this.AutoScaleDimensions = new System.Drawing.SizeF(120F, 120F);
```
RussKie: The design must be done at 100% scale / 96dpi. The latest VS2017 supports design at 100% scaling
image

You can start it manually too with
```
"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\devenv.exe" /noScale "<path>\GitExtensions.sln"
```

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.Designer.cs
```diff
@@ -597,5 +634,7 @@ private void InitializeComponent()
         private System.Windows.Forms.TableLayoutPanel tableLayoutPanel5;
         private System.Windows.Forms.ComboBox ProtocolDropdownList;
         private System.Windows.Forms.Label ProtocolLabel;
+        private System.Windows.Forms.CheckBox depthCb;
+        private System.Windows.Forms.TextBox depthTb;
```
RussKie: This should be represented as a numeric up/down control with sensible limits.

oriash93: What would those limits be? I couldn't find any restrictions in Git documentation.

gerhardol: It could be just >=1 (or 0==disabled).
Default 1 if enabled?

A up/down is nicer than parsing text - but if you want to use this option, you probably know what you are doing anyway

RussKie: > (or 0==disabled).

👍 For some reason it didn't occur to me. This way we don't need the checkbox.
The label could be "Depth limit (0 - no limit)".

I supposed max can be set at 100, if you need deeper you may as well clone the whole thing.
Thoughts?

gerhardol: > I supposed max can be set at 100, if you need deeper you may as well clone the whole thing.

I see no reason enforce a limit

RussKie: We need to set the upper limit on the updown control. 
Obviously after some point (e.g. depth 100 or depth 1,000) specifying a depth is essentially the same as doing a full clone. We just need to establish what that "reasonable" depth is, and set the control's limit

gerhardol: Setting something like 1000 is not reasonable, but there could be scenarios where this could be desired. Linux has 650K commits and you may not be interested in the complete history.
In reality, very few will use depth at all, if you use it it will almost always be 1.

RussKie: There we go, we established an upper boundary 😄
So the updown control with the range [0, 1,000) and the default to 0 (that means 'full depth'). 
Correct?

gerhardol: For Linux repo, 100K is a possible value. But we can leave that for the command line if you want a "neat" box.

RussKie: As much as I'd like GE to be the default git client for Linux devs, I think all of them choke themselves before using a Windows UI tool 
🤣

oriash93: Done :)

RussKie: > the gap is reserved for an existing Label, which I missed at first

It does look somewhat untidy. What is that label used for? 🤔

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.Designer.cs
```diff
@@ -501,6 +517,28 @@ private void InitializeComponent()
...
+            0,
+            0,
+            0});
+            this.depthUpDown.Maximum = new decimal(new int[] {
+            999,
+            0,
+            0,
+            0});
+            this.depthUpDown.Size = new System.Drawing.Size(100, 20);
+            this.depthUpDown.TabIndex = 27;
```
RussKie: Can you please confirm the tab order is logical?

oriash93: To be honest, I just copied it. It's the same as any other usage of NumericUpDown in GE.

RussKie:
Could you please ensure the tab order on the form is logical?

oriash93:
Sorry, I don't follow. Please explain

vbjay:
Click view-> tab order. Fix tab order in form.

RussKie:
The tab order is still not logical.

You need to click "Tab Order" toolbar item (it is usually visible by default) and then go through the controls and mark them in order.

You should see something like this:
image

RussKie:
I have pushed a fix to the tab order.
Please verify everything still works as expected, squash and we'll get it in.
Thank you

oriash93:
> I have pushed a fix to the tab order.
> Please verify everything still works as expected, squash and we'll get it in.
> Thank you

Thank you! Done. 👍

RussKie:
Thank you

PR #6332, opened 2019-03-02
fix: Thread exhaustion by commit info data load

### LINKED ISSUE #6319: [Blocker] Thread exhaustion
## Current behaviour

Scrolling the revision graph (with attached debugger) is very slow and jerky. 
It looks like we are firing too many background threads and this leads to thread exhaustion
image
image

Obviously our threading flows are far from efficient or performant.
The main problem seems to be `StartAsyncDataLoad`:
https://github.com/gitextensions/gitextensions/blob/1d889787237a7c6f46bd7ca7d8b6c3a1a8cac127/GitUI/CommitInfo/CommitInfo.cs#L228-L259

## Expected behaviour

The scroll is fast and smooth. No excessive use of blocked thread.

## Steps to reproduce

* Open and run GE solution in VS
* Scroll the graph by pressing and holding the <kbd>&darr;</kbd> or rapidly scroll down for mouse-scroll

Yes, v2 didn't exhibit as bad perf.

## Environment

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version:

RussKie:
Fixes #6319

## Proposed changes

- Combine all async calls that load commit information into a single background call that reduces the number of used threads.

Overall the app feels snappier. I made a number of tests but couldn't replicate hung up or blocked threads anymore.

### Before

image

### After

image

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -274,29 +274,40 @@ namespace GitUI.CommitInfo
             {
                 var cancellationToken = _asyncLoadCancellation.Next();
 
-                ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadLinksForRevisionAsync(_revision)).FileAndForget();
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () => { await LoadLinksForRevisionAsync(_revision); }).FileAndForget();
 
-                // No branch/tag data for artificial commands
-
-                if (AppSettings.CommitInfoShowContainedInBranches)
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadBranchInfoAsync(_revision.ObjectId)).FileAndForget();
-                }
+                    if (_refsOrderDict == null)
+                    {
+                        await LoadSortedRefsAsync();
+                    }
 
-                if (AppSettings.ShowAnnotatedTagsMessages)
-                {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadAnnotatedTagInfoAsync(_revision.Refs)).FileAndForget();
-                }
+                    // No branch/tag data for artificial commands
+                    if (AppSettings.CommitInfoShowContainedInBranches)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        await LoadBranchInfoAsync(_revision.ObjectId);
+                    }
 
-                if (AppSettings.CommitInfoShowContainedInTags)
-                {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadTagInfoAsync(_revision.ObjectId)).FileAndForget();
-                }
+                    if (AppSettings.ShowAnnotatedTagsMessages)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        await LoadAnnotatedTagInfoAsync(_revision.Refs);
+                    }
 
-                if (AppSettings.CommitInfoShowTagThisCommitDerivesFrom)
-                {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadDescribeInfoAsync(_revision.ObjectId)).FileAndForget();
-                }
+                    if (AppSettings.CommitInfoShowContainedInTags)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        await LoadTagInfoAsync(_revision.ObjectId);
+                    }
+
+                    if (AppSettings.CommitInfoShowTagThisCommitDerivesFrom)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        await LoadDescribeInfoAsync(_revision.ObjectId);
+                    }
+                }).FileAndForget();
 
                 return;
 
@@ -336,9 +347,30 @@ namespace GitUI.CommitInfo
                     }
                 }
 
+                async Task LoadSortedRefsAsync()
+                {
+                    await TaskScheduler.Default;
+                    _refsOrderDict = ToDictionary(Module.GetSortedRefs());
+
+                    await this.SwitchToMainThreadAsync(cancellationToken);
+                    UpdateRevisionInfo();
+
+                    IDictionary<string, int> ToDictionary(IReadOnlyList<string> list)
+                    {
+                        var dict = new Dictionary<string, int>();
+                        for (int i = 0; i < list.Count; i++)
+                        {
+                            dict.Add(list[i], i);
+                        }
+
+                        return dict;
+                    }
+                }
+
                 async Task LoadAnnotatedTagInfoAsync(IReadOnlyList<IGitRef> refs)
                 {
-                    await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                    await TaskScheduler.Default;
+
                     var annotatedTagsMessages = GetAnnotatedTagsMessages();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
@@ -400,7 +432,8 @@ namespace GitUI.CommitInfo
 
                 async Task LoadTagInfoAsync(ObjectId objectId)
                 {
-                    await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                    await TaskScheduler.Default;
+
                     var tags = Module.GetAllTagsWhichContainGivenCommit(objectId).ToList();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
@@ -410,7 +443,7 @@ namespace GitUI.CommitInfo
 
                 async Task LoadBranchInfoAsync(ObjectId revision)
                 {
-                    await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                    await TaskScheduler.Default;
 
                     // Include local branches if explicitly requested or when needed to decide whether to show remotes
                     bool getLocal = AppSettings.CommitInfoShowContainedInBranchesLocal ||
@@ -429,6 +462,7 @@ namespace GitUI.CommitInfo
                 async Task LoadDescribeInfoAsync(ObjectId commitId)
                 {
                     await TaskScheduler.Default;
+
                     var info = GetDescribeInfoForRevision();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);


sharwell:
The original bug is caused by our process executor not being async-all-the-way-down (we block while waiting for git calls to complete). This pull request mitigates (does not fix) the issue by sequencing one set of such calls instead of running them in parallel.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -277,6 +280,8 @@ async Task LoadLinksForRevisionAsync(GitRevision revision)
                     }
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
sharwell: 📝 This is handled already by the extension method on the previous line.

```suggestion
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -318,10 +326,12 @@ async Task LoadSortedRefsAsync()
 
                 async Task LoadAnnotatedTagInfoAsync(IReadOnlyList<IGitRef> refs)
                 {
-                    await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                    await TaskScheduler.Default;
                     var annotatedTagsMessages = GetAnnotatedTagsMessages();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
sharwell:
```suggestion
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -380,17 +390,22 @@ async Task LoadAnnotatedTagInfoAsync(IReadOnlyList<IGitRef> refs)
...
                 async Task LoadTagInfoAsync(ObjectId objectId)
                 {
-                    await TaskScheduler.Default.SwitchTo(alwaysYield: true);
+                    cancellationToken.ThrowIfCancellationRequested();
+                    await TaskScheduler.Default;
+
                     var tags = Module.GetAllTagsWhichContainGivenCommit(objectId).ToList();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
sharwell:
```suggestion
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -402,16 +417,22 @@ async Task LoadBranchInfoAsync(ObjectId revision)
                     var branches = Module.GetAllBranchesWhichContainGivenCommit(revision, getLocal, getRemote).ToList();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
sharwell:
```suggestion
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -402,16 +417,22 @@ async Task LoadBranchInfoAsync(ObjectId revision)
...
 
                 async Task LoadDescribeInfoAsync(ObjectId commitId)
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
                     await TaskScheduler.Default;
+
                     var info = GetDescribeInfoForRevision();
 
                     await this.SwitchToMainThreadAsync(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
sharwell:
```suggestion
```

RussKie:
> The original bug is caused by our process executor not being async-all-the-way-down (we block while waiting for git calls to complete). This pull request mitigates (does not fix) the issue by sequencing one set of such calls instead of running them in parallel.

Yes, I considered reworking the code to pass cancellation token in each individual method, but it felt like more work for relatively low value or no value at all.
Each git command is executed in sequence, so spawning multiple threads that get queued anyway probably causes more harm than good.

GitUI/CommitInfo/CommitInfo.cs
RussKie: Thank you

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -229,34 +229,36 @@ void StartAsyncDataLoad()
             {
                 var cancellationToken = _asyncLoadCancellation.Next();
 
-                ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadLinksForRevisionAsync(_revision)).FileAndForget();
```
jbialobr: It seems that LoadLinksForRevisionAsync does not block, so it could remain async.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -229,34 +229,36 @@ void StartAsyncDataLoad()
...
-
-                if (_refsOrderDict == null)
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadSortedRefsAsync()).FileAndForget();
-                }
+                    await LoadLinksForRevisionAsync(_revision);
 
-                // No branch/tag data for artificial commands
+                    if (_refsOrderDict == null)
```
jbialobr: `_refsOrderDict ` should be wrapped in AsyncLazy and not called here at all.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -267,6 +269,7 @@ async Task LoadLinksForRevisionAsync(GitRevision revision)
                         return;
                     }
 
+                    cancellationToken.ThrowIfCancellationRequested();
```
jbialobr: I think that all these `cancellationToken.ThrowIfCancellationRequested()`
should be directly in `StartAsyncDataLoad`. When we get to the state that all the execution is async-all-the-way-down then it will be clear that `cancellationToken.ThrowIfCancellationRequested()` are to be removed from `StartAsyncDataLoad`.
Putting `cancellationToken.ThrowIfCancellationRequested()` here brings the risk that they will stay here.

RussKie: I am afraid I don't quite understand what you're trying to say, sorry.

The way I understand it is as follows.
At the top of `StartAsyncDataLoad` we provision a new cancellation token. A previously issued token (if any) is cancelled, and aborts any pending operations. To achieve this we need `cancellationToken.ThrowIfCancellationRequested()` calls and pass the token to `this.SwitchToMainThreadAsync()` calls.

GitUI/CommitInfo/CommitInfo.cs
RussKie: Yes and no. It isn't executing a git operation, you're right; but he parser may be reading definitions from the filesystem.
I'll update it.

GitUI/CommitInfo/CommitInfo.cs
RussKie: Perhaps you're right, however it seems like a bigger change than I can afford allocate time for.
If it is left as-is for now we could query refs few extra times, anything else?

jbialobr: >anything else?

No, only that.

GitUI/CommitInfo/CommitInfo.cs
jbialobr: What I am trying to say is that after your change `LoadAnnotatedTagInfoAsync` knows that it will be called in sequence.
Which I think it should not. Only the caller should know that, and it is its job to call `cancellationToken.ThrowIfCancellationRequested()`. If `LoadAnnotatedTagInfoAsync` was called in parallel then it would not be much sense to call `cancellationToken.ThrowIfCancellationRequested()` at the beginning, since we know that no cancellation could occur.
In addition there is a big chance that someone who makes it parallel again (when all the execution is async-all-the-way-down) will not remove this `cancellationToken.ThrowIfCancellationRequested()` at the beginning of `LoadAnnotatedTagInfoAsync`. The same applies to the other methods.

GitUI/CommitInfo/CommitInfo.cs
jbialobr: I created a PR for this: https://github.com/gitextensions/gitextensions/pull/6335
It turned out that once loaded the sortedRefs were never refreshed on repository change.

jbialobr:
>Yes, I considered reworking the code to pass cancellation token in each individual method, but it felt like more work for relatively low value or no value at all.

It will not help for cancelling the token will not stop the running git command. The only we can do is as @sharwell said make it async-all-the-way-down to not block background threads, but even then we will be spawning multiple git processes.

>Each git command is executed in sequence, so spawning multiple threads that get queued anyway probably causes more harm than good.

Why do you think so? Is it done by git or is GE sequencing them?

RussKie:
> > Each git command is executed in sequence, so spawning multiple threads that get queued anyway probably causes more harm than good.
> 
> Why do you think so? Is it done by git or is GE sequencing them?

For some reason I was under impression we had them sequenced... 
I guess I was wrong. 😕

RussKie:
> It will not help for cancelling the token will not stop the running git command.

It won't, no. But it would prevent new commands from being issued, wouldn't it?

GitUI/CommitInfo/CommitInfo.cs
RussKie: I disagree, each method doesn't have a notion of the order it is called; the composition is still the function of `StartAsyncDataLoad`.

I am working on an update to the commit info control (though it is a glacial progress given my load) and planning to refactor/rework how data is populated and displayed to the user.

RussKie:
@jbialobr I understand you'd like to see a better change how I currently don't have capacity to rewrite the code in a "proper all-the-way-down-async". 
We need to address the immediate problem however and avoid exhausting available threads. This issue must be resolved for the 3.1.0 release, which we should ship pretty soon.

Would you like to take over and provide a better fix?

GitUI/CommitInfo/CommitInfo.cs
jbialobr: I am afraid you still do not know what I am trying to say.

jbialobr:
>@jbialobr I understand you'd like to see a better change how I currently don't have capacity to rewrite the code in a "proper all-the-way-down-async".
We need to address the immediate problem however and avoid exhausting available threads. This issue must be resolved for the 3.1.0 release, which we should ship pretty soon.
Would you like to take over and provide a better fix?

You took me wrong. I just wanted you to be aware where the problem is. I read your answer to @sharwell 's comment and had impression that you are talking about sth totally different than @sharwell said. Thus I wanted to clarify it. I never wanted (and even thought) you to make it all-the-way-down-async.

GitUI/CommitInfo/CommitInfo.cs
jbialobr: I am not talking here about the order. I am talking about parallel vs sequential invocation. Putting cancellationToken.ThrowIfCancellationRequested() at the beginning of LoadAnnotatedTagInfoAsync makes only sense when `LoadAnnotatedTagInfoAsync ` knows it be invoked in a sequential way.

vbjay:
A custom task scheduler would allow full async and prevent thread exhaustion.

GitUI/CommitInfo/CommitInfo.cs
jbialobr: Saying this in other words:
Lets assume I did the all-the-way-down-async and now I am changing StartAsyncDataLoad that it calls all the Load* functions in parallel. Does it make sense to leave `cancellationToken.ThrowIfCancellationRequested();` at the beginning of Load* functions? Or should them be removed?

RussKie: Updated

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -274,29 +274,36 @@ void StartAsyncDataLoad()
...
+                    if (AppSettings.CommitInfoShowContainedInBranches)
+                    {
+                        await LoadBranchInfoAsync(_revision.ObjectId);
+                    }
 
-                if (AppSettings.CommitInfoShowContainedInTags)
-                {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadTagInfoAsync(_revision.ObjectId)).FileAndForget();
-                }
+                    if (AppSettings.ShowAnnotatedTagsMessages)
```
jbialobr: cancellationToken.ThrowIfCancellationRequested();

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -274,29 +274,36 @@ void StartAsyncDataLoad()
...
+                    {
+                        await LoadAnnotatedTagInfoAsync(_revision.Refs);
+                    }
 
-                if (AppSettings.CommitInfoShowTagThisCommitDerivesFrom)
-                {
-                    ThreadHelper.JoinableTaskFactory.RunAsync(() => LoadDescribeInfoAsync(_revision.ObjectId)).FileAndForget();
-                }
+                    if (AppSettings.CommitInfoShowContainedInTags)
```
jbialobr: cancellationToken.ThrowIfCancellationRequested();

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -274,29 +274,36 @@ void StartAsyncDataLoad()
...
+                    {
+                        await LoadTagInfoAsync(_revision.ObjectId);
+                    }
+
+                    if (AppSettings.CommitInfoShowTagThisCommitDerivesFrom)
```
jbialobr: cancellationToken.ThrowIfCancellationRequested();

GitUI/CommitInfo/CommitInfo.cs
RussKie: But then `cancellationToken.ThrowIfCancellationRequested();` should be as the first line in `LoadAnnotatedTagInfoAsync`.

jbialobr: We have awaited `LoadBranchInfoAsync` and now we want to make sure no cancellation was requested. `LoadBranchInfoAsync` does not know there was an async call before, hence we should do the check here.

RussKie: updated

RussKie:
Thank you

PR #6339, opened 2019-03-03
Add ability to Copy to Clipboard, similar to commit context menu from left panel

### LINKED ISSUE #6304: Branches pane context menu should support Copy to Clipboard, similar to commit context menu
## Feature description
The branches context menu is missing a number of useful commands only available in the commit graph window.  

image

One class of such commands are "Copy to clipboard" options, similar to what is available when right-clicking in the commit graph.

These are the Copy commands I'd like to see:

1. Branch name
2. Commit hash (for current branch HEAD)

## Environment

- Git Extensions 3.0.2.5232
- Git 2.19.0.gvfs.1.34.gc7fb556
- Microsoft Windows NT 10.0.18343.0
- .NET Framework 4.8.3745.0
- DPI 96dpi (no scaling)

ivangrek:
Fixes #6304

## Proposed changes

- Create and add to context menu on the left panel CopyContextMenuItem

### Before

15825c7a-f60b-4df4-ac95-607621232487 png

### After

475142ba-b6f2-4d2a-bee5-d81db495e385 png

## Test methodology 

Manually 

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.cs
@@ -12,6 +12,7 @@ using GitExtUtils.GitUI;
 using GitExtUtils.GitUI.Theming;
 using GitUI.CommandsDialogs;
 using GitUI.Properties;
+using GitUI.Script;
 using GitUI.UserControls;
 using GitUI.UserControls.RevisionGrid;
 using GitUIPluginInterfaces;
@@ -41,6 +42,7 @@ namespace GitUI.BranchTreePanel
         private IAheadBehindDataProvider _aheadBehindDataProvider;
         private bool _searchCriteriaChanged;
         private ICheckRefs _refsSource;
+        private IScriptHostControl _scriptHost;
 
         public RepoObjectsTree()
         {
@@ -227,11 +229,12 @@ namespace GitUI.BranchTreePanel
             }
         }
 
-        public void Initialize([CanBeNull] IAheadBehindDataProvider aheadBehindDataProvider, FilterBranchHelper filterBranchHelper, ICheckRefs refsSource)
+        public void Initialize([CanBeNull] IAheadBehindDataProvider aheadBehindDataProvider, FilterBranchHelper filterBranchHelper, ICheckRefs refsSource, IScriptHostControl scriptHost)
         {
             _aheadBehindDataProvider = aheadBehindDataProvider;
             _filterBranchHelper = filterBranchHelper;
             _refsSource = refsSource;
+            _scriptHost = scriptHost;
 
             // This lazily sets the command source, invoking OnUICommandsSourceSet, which is required for setting up
             // notifications for each Tree.
--- a/GitUI/UserControls/RevisionGrid/CopyContextMenuItem.cs
+++ b/GitUI/UserControls/RevisionGrid/CopyContextMenuItem.cs
@@ -63,7 +63,7 @@ namespace GitUI.UserControls.RevisionGrid
 
             var item = new ToolStripMenuItem
             {
-                Text = displayText,
+                Text = displayText.TrimEnd('\r', '\n'),
                 ShowShortcutKeys = true,
                 Image = image
             };
--- a/GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
@@ -8,6 +8,8 @@ using System.Windows.Forms;
 using GitCommands;
 using GitUI.BranchTreePanel.ContextMenu;
 using GitUI.BranchTreePanel.Interfaces;
+using GitUI.UserControls.RevisionGrid;
+using GitUIPluginInterfaces;
 using ResourceManager;
 
 namespace GitUI.BranchTreePanel
@@ -17,6 +19,8 @@ namespace GitUI.BranchTreePanel
         private GitRefsSortOrderContextMenuItem _sortOrderContextMenuItem;
         private GitRefsSortByContextMenuItem _sortByContextMenuItem;
         private ToolStripSeparator _tsmiSortMenuSpacer = new ToolStripSeparator { Name = "tsmiSortMenuSpacer" };
+        private ToolStripItem[] _menuBranchCopyContextMenuItems = Array.Empty<ToolStripItem>();
+        private ToolStripItem[] _menuRemoteCopyContextMenuItems = Array.Empty<ToolStripItem>();
 
         /// <summary>
         /// Local branch context menu [git ref / rename / delete] actions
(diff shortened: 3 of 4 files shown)


GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -89,7 +89,6 @@ private void InitializeComponent()
             this.tsmiShowBranches = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiShowTags = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiShowRemotes = new System.Windows.Forms.ToolStripMenuItem();
-            this.toolTip = new System.Windows.Forms.ToolTip(this.components);
```
gerhardol: Should not be removed

ivangrek: There are two lines the same

RussKie:
I've checked out your revision. There are some visual artifacts:
image
image

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -173,6 +176,11 @@ void CollectFromNodes(TreeNodeCollection nodes)
             }
         }
 
+        public void AddDependencies(ICanGetSelectedRevisions canGetSelectedRevisions)
+        {
+            _canGetSelectedRevisions = canGetSelectedRevisions;
+        }
```
RussKie: This introduces a new pattern, which isn't a problem in itself, however it increases an entropy.
All necessary dependencies to `RepoObjectsTree` are supplied via `Initialize`
https://github.com/gitextensions/gitextensions/blob/521e9e8793799a9b6969868fdfc60f2f5f12a32d/GitUI/BranchTreePanel/RepoObjectsTree.cs#L210-L218

GitUI/Browsing/ICanGetSelectedRevisions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.DirectoryServices;
+using GitCommands;
+
+namespace GitUI.Browsing
```
RussKie: Could you please explain the choice of the namespace?

GitUI/UserControls/RevisionGrid/CopyContextMenuItem.cs
```diff
@@ -62,7 +62,7 @@ private void AddItem([NotNull] string displayText, [NotNull] string textToCopy,
 
             var item = new ToolStripMenuItem
             {
-                Text = displayText,
+                Text = displayText.TrimEnd('\r', '\n'),
```
RussKie: Why is it not needed when invoked from the revision graph?

GitUI/BranchTreePanel/RepoObjectsTree.cs
ivangrek: Sorry. 😄

GitUI/Browsing/ICanGetSelectedRevisions.cs
ivangrek: By analogy with FormBrowse. I wanted to isolate code that is difficult or impossible to place in controls in a separate abstract namespace. Browsing is the main area, Settings is some other area and etc.

GitUI/UserControls/RevisionGrid/CopyContextMenuItem.cs
ivangrek: This is a general situation for the main grid and for the left panel. When multiple lines increase the height of the menu item this break the layout. If it is in the middle, it does not look weird than if it is at the end of the text.

ivangrek:
For discussion in #8414 

We can get git revision on  user request.
At the moment of opening the menu. This will be similar to running a command in the console.

`public GitRevision GetRevision([CanBeNull] ObjectId objectId = null, bool shortFormat = false, bool loadRefs = false)`

Or we can add "GetRevisions" method with commands.

`git show branch1 branch2 [list others]  --format=%H%n%T%n%P%n%aN%n%aE%n%at%n%cN%n%cE%n%ct%n%e%n --no-patch`

image

And add revision information on tree build

```
private async Task<Nodes> LoadNodesAsync(CancellationToken token)
{
    await TaskScheduler.Default;

    var branchNames = Module.GetRefs(tags: false, branches: true, noLocks: true).Select(b => b.Name);
    return FillBranchTree(branchNames, token);
}
```
After

`var branchNames = Module.GetRefs(tags: false, branches: true, noLocks: true).Select(b => b.Name);`

Most likely there is a way to combine both commands.

---

In the current solution, the situation is similar with hidden branches - it shows the last selected one.

ivangrek:
Also for discussion #8414

`Text = displayText.TrimEnd('\r', '\n')`

For fix

image

RussKie:
> Also for discussion #8414
> 
> `Text = displayText.TrimEnd('\r', '\n')`
> 
> For fix

Yes, I think we'll have to do this 👍 

> For discussion in #8414
> 
> We can get git revision on user request.
> At the moment of opening the menu. This will be similar to running a command in the console.
> 
> `public GitRevision GetRevision([CanBeNull] ObjectId objectId = null, bool shortFormat = false, bool loadRefs = false)`
> 
> Or we can add "GetRevisions" method with commands.
> 
> `git show branch1 branch2 [list others] --format=%H%n%T%n%P%n%aN%n%aE%n%at%n%cN%n%cE%n%ct%n%e%n --no-patch`

If possible I'd rather reduce a number of commands we run (we run a lot of them as is, some commands are run multiple times during the same operation).
When a user clicks on a node in the left panel, we select the corresponding commit in the revision grid, and from it we have access to the same information if the user were to right click on the grid entry itself.

Pertaining to the discussion in #8414 - if a branch is filtered out of the view, I can't see a good use case where we need to bend over backwards to provide extra information for it. Similar to an inactive remote - there's no information until it is activated and fetched.

RussKie (commented):
👍 
Currently blocked by https://github.com/gitextensions/gitextensions/pull/8434 and https://github.com/gitextensions/gitextensions/pull/8414 due to pending infra changes.

ivangrek:
> Pertaining to the discussion in #8414 - if a branch is filtered out of the view, I can't see a good use case where we need to bend over backwards to provide extra information for it. Similar to an inactive remote - there's no information until it is activated and fetched.

I understand correctly that there is no separation between model and view. Everything is stored in a grid, and after deletion you need to access the repository?

ivangrek:
As I imagine, we have a hash of the commit, and we can get all the necessary information from the model by the hash. (ObjectId)
But that's not how it works now.

RussKie:
> I understand correctly that there is no separation between model and view. Everything is stored in a grid, and after deletion you need to access the repository?

Correct, unfortunately there's no separation (and it does bother me). And this makes a lot of thing significantly harder than they should be. E.g. we do a lot of refreshes, which cause flicker, and result in multiple git calls.
Ideally the grid should just bind to an observable list of commits.

ivangrek:
Removed the display for the hidden node.

image

image

RussKie:
I ran it locally, and something isn't working 100% but I couldn't quite see from the diff.

When I right click on the latest master (43557461def26bbcfe7ef982c04cf6a7559c38a1) in the revision grid and select copy commit:
image
I get the following:
```
Merge pull request #8527 from BrianFreemanAtlanta/RemoveFilesUpdateNameTooLong

updated removeFiles to use GetBatchOutput and added unit tests
```

When I do the same from the left panel I only get the first line:
```
Merge pull request #8527 from BrianFreemanAtlanta/RemoveFilesUpdateNameTooLong
```

ivangrek:
Changed revisions provider.

`copyContextMenuItem.SetRevisionFunc(() => _scriptHost.GetSelectedRevisions());`

RussKie (approved):

RussKie:
Thank you 🎉 

This certainly has been a long journey, thank you for sticking around.

PR #6349, opened 2019-03-07
Left pane reorder trees

amaiorano:
## Proposed changes

- Add the ability to reorder root Tree nodes in left panel: Branches, Remotes, Tags, and Submodules.
- This is done by right-clicking the root node and selecting from "Move Up" or "Move Down" context actions.
- The "Show" menu items in the left panel settings menu are also reordered to match.

## Screenshots 

Context menu action, note that because it's the top-most node, "Move Up" is disabled:
image

Similarly, if bottom-most node is selected, "Move Down" is disabled. And there's only one node (because the other three are hidden), then both are disabled:

image

The settings menu where you can show/hide Trees matches the order:

image

If Trees are hidden, reordering still works as expected. For instance, say Tags and Branches is hidden like so:

image

If we right-click Submodules and move it down:

image

It will swap it with Remotes, but Branches and Tags keep their positions:

image

## Test methodology 

- Manual testing.

## Test environment(s) 

- GIT version 2.19.1.windows.1
- Windows 10

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Windows.Forms;
 using GitCommands;
@@ -9,6 +8,26 @@ namespace GitUI.BranchTreePanel
 {
     public partial class RepoObjectsTree
     {
+        /// <summary>
+        /// We assume tree to position indices are 0-based and sequential. In case this
+        /// is no longer true, because for e.g. user has reverted to an earlier version,
+        /// this function will fix the indices, attempting to maintain the existing order.
+        /// </summary>
+        private void FixInvalidTreeToPositionIndices()
+        {
+            // Sort by index, then force assign 0-based sequential indices
+            var treeToIndex = GetTreeToPositionIndex();
+
+            int i = 0;
+            foreach (var kvp in treeToIndex.OrderBy(kvp => kvp.Value))
+            {
+                treeToIndex[kvp.Key] = i;
+                ++i;
+            }
+
+            SaveTreeToPositionIndex(treeToIndex);
+        }
+
         private Dictionary<Tree, int> GetTreeToPositionIndex()
         {
             return new Dictionary<Tree, int>
--- a/GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
@@ -1,30 +1,120 @@
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Windows.Forms;
 using GitCommands;
 
 namespace GitUI.BranchTreePanel
 {
     public partial class RepoObjectsTree
     {
+        private Dictionary<Tree, int> GetTreeToPositionIndex()
+        {
+            return new Dictionary<Tree, int>
+            {
+                [_branchesTree] = AppSettings.RepoObjectsTreeBranchesIndex,
+                [_remotesTree] = AppSettings.RepoObjectsTreeRemotesIndex,
+                [_tagTree] = AppSettings.RepoObjectsTreeTagsIndex,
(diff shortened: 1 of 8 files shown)


amaiorano:
@gerhardol This one's for you ;)

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -279,21 +282,19 @@ private void CreateSubmodules()
...
-
             // Add Tree's node in position index order. Because TreeNodeCollections cannot be sorted,
             // we create a list from it, sort it, then clear and re-add the nodes back to the collection.
             treeMain.BeginUpdate();
             List<TreeNode> nodeList = treeMain.Nodes.OfType<TreeNode>().ToList();
             nodeList.Add(tree.TreeViewNode);
             treeMain.Nodes.Clear();
-            treeMain.Nodes.AddRange(nodeList.OrderBy(treeNode => _treeToPositionIndex[treeNode.Tag as Tree]).ToArray());
+            var treeToPositionIndex = GetTreeToPositionIndex();
+            treeMain.Nodes.AddRange(nodeList.OrderBy(treeNode => treeToPositionIndex[treeNode.Tag as Tree]).ToArray());
```
gerhardol: This should handle "corrupted" indexes like gaps and not unique

GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
```diff
@@ -1,30 +1,119 @@
...
+            var treeToShowMenuItem = new Dictionary<Tree, ToolStripMenuItem>
+            {
+                [_branchesTree] = tsmiShowBranches,
+                [_remotesTree] = tsmiShowRemotes,
+                [_tagTree] = tsmiShowTags,
+                [_submoduleTree] = tsmiShowSubmodules
+            };
+            for (int i = 0; i < treeToShowMenuItem.Count(); ++i)
+            {
+                var tsmi = treeToShowMenuItem[indexToTree[i]];
```
gerhardol: This will not chandle corrupted indexes (for instance if adding a new tree and downgrading)

amaiorano:
> One comment about corrupted indexes

I'll take a look and see how I can handle corrupted indices.

> I do not like that the trees expand when moving, but when Submodules is on top, I will not change this again

Yes, I also don't like this; however, as you note, reordering nodes isn't something that should be done often. I considered passing an extra argument to AddTree/RemoveTree to tell it whether to run the Attached/Detached code, but it feels... dirty. Another option is to make it so that Branches/Remote/Tag trees don't update unless they have to; this is already what SubmoduleTree via it's _currentSubmoduleInfo variable (it's only from the SubmoduleStatusProvider callback, and cleared once the tree has been built). This should be fairly easy to do, and will get rid of extraneous updates. I'll add this as another task to do for the left panel.

amaiorano:
@gerhardol Pushed a new commit that solves the potential invalid/corrupt indices problem. For a while, I thought about modifying the code you pointed out to handle corrupt indices, but it made things more difficult to understand. So I came up with what I think is a better solution: at startup, we fix up invalid indices, attempting to maintain the current ordering. You can try to corrupt the values, like making one out of range, or making multiple the same, etc. and it will work.

Let me know what you think.

gerhardol:
time to remove WIP?

amaiorano:
Merge?

amaiorano:
@RussKie @gerhardol approved this a few days ago. I wasn't sure if I should merge yet, or wait for another review. Since I haven't heard from @gerhardol in a couple days, I decided to ask you to take a look.

Cheers.

gerhardol:
I am ok with merging no more comments. But I have a different view on tests for instance so I normally give @russkie a few days to comment before merging

RussKie:
I'm currently busy with RL commitments, please don't wait for me.

amaiorano:
I started writing tests for this. Going back to WIP, and will update once the test is ready.

amaiorano:
Updated with tests. If no comments, we can merge this in a few hours.

pmiossec:
>I'm currently busy with RL commitments, please don't wait for me.

From french writer "Alphonse de Lamartine" : "Sometimes, only one person is missing, and the whole word seems depopulated" 😄

RussKie:
> From french writer "Alphonse de Lamartine" : "Sometimes, only one person is missing, and the whole word seems depopulated" 😄

That's so sweet and sad at the same time.
Hopefully I'll be back soon.

PR #6356, opened 2019-03-09
WIP: Format patch: Select more than two: path is outside repository

### LINKED ISSUE #5923: Format patch: Select more than two: path is outside repository
bug

If more than two commits are selected in Format patch form, the patches are not created.
image

* Open Apply Patch
* Select three or more patches. 
* Create patch

Possible to create more than two patches at the same time

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

2.51

NOYB:
Format patch: Select more than two: path is outside repository
Fixes #5923

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1411,7 +1411,7 @@ namespace GitCommands
                 new GitArgumentBuilder("format-patch")
                 {
                     "-M -C -B",
-                    { start != null, $"-- start-number {start}" },
+                    { start != null, $"--start-number {start}" },
                     $"{from.Quote()}..{to.Quote()}",
                     $"-o {output.ToPosixPath().Quote()}"
                 });


vbjay:
See
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/22954065.
Test with the portable version. If it fixes the problem then say so as a
reply in this pull request.

NOYB:
Yes this fixed the problem.

During testing discovered another Format patch... problem that existed at least as far back as Git Extensions 2.51.04
https://github.com/gitextensions/gitextensions/issues/6358

RussKie:
Please send your [sign off to the contributors.txt](https://github.com/gitextensions/gitextensions/wiki/Guidelines%3A-contributor#contributor-agreement) in a separate PR, thanks

gerhardol:
> Please send your sign off to the contributors.txt in a separate PR, thanks

Reminder

RussKie will ask for tests too.

pmiossec:
>Please add tests

I find it harsh **for a first time contributor** to ask him to write tests for such a quick fix on a clearly bad command line parameter. I think this PR should have been merged from a long time ago (that would have prevented me to loose half an hour trying to understand how to use this feature just to end up with the same fix :cry: )

RussKie:
> I find it harsh **for a first time contributor** to ask him to write tests

Respectfully disagree. 
I don't suppose at work you merge an arbitrary change without any tests from a developer who just started _because that is his first day_.

In the past we've merged changes without tests that eventually would result in regressions. But by then the original develop disappeared and we had to own and fix the change.... 

Unit tests isn't something was invented in this repository, it is public knowledge, and I expect any developer worth his salt to know how to write tests.

RussKie:
> ...that would have prevented me to loose half an hour trying to understand how to use this feature just to end up with the same fix

Would you like to send a fix with tests? 😏

pmiossec:
>Respectfully disagree.

I know, I already knew (and understood) your point of view (which is, IMHO, the right one in a perfect world but that can't be applied ever times) , that's why I gave mine (and with great respect also 😉) which is to be pragmatic... 

>I don't suppose at work you merge an arbitrary change without any tests from a developer who just started because that is his first day.

IMHO, that's an error for such a case to compare opensource and enterprise development here.
Because I agree with this behavior/expectations at work (because you can), but here, you have to be more pragmatic because asking someone to do a 100 times harder/more consuming task will end up that he will run away (which should already be the case here).

I personally think (for a lot of reasons that will be difficult to explain here) that we can't ask to (obliged?) occasional contributors to solve an already existing problem much more important/difficult than what he initially try to fix. 

So, by waiting for something that will never come, once again, for such a trivial fix, you will never merge the PR and kept this bug (and I could assure you that the behavior is very bad at the moment, with for exemple not understandable behavior and empty pop-up appearing). 

So, being pragmatic, what is worse for the software/team ? Having no tests or still having bugs? Hard question... 


Ps: it's a shame that you didn't quote because "for such a quick fix" was an important part of what motivated my comment from a pragmatic point of view.

>Would you like to send a fix with tests? 😏

Nice try to ask but to be frank, I never used this feature and so don't care about it. I was just obligated to use it to see if I didn't introduce a regression when fixing another bad behavior in GE and was very intrigued by how bad it was working only to understand that it was just due to this typo! 

Anyway, I could tell you that I will add it to my already very long todo list but unfortunately it will be near the end (because I don't have return on investment, I'm not motivated by it and it requires me to invest a lot of time because I really don't know what should be tested).

So, the reality will surely be that I will never do it...

RussKie:
Thank you for sharing your view. My view that in FOSS tests are especially more important than in a commercial environment. 

A bug in a commercial software may result in "all hands on deck" situation and a quick patch and a quick release to meet SLAs in place. 
The FOSS is (for the most part) run on a good will and free time of maintainers, and thus can rarely afford "emergency" situations. Look how often OSS projects release. I have several pending contributions (including bug fixes) to several repos (some of which are backed by software powerhouses), and yet it has been months with no interactions...
Look at how may bugs we have in our backlog. How may of them do you think could have been avoided if we had tests in place?

Our release cycle is 2 releases per year at best. If we accept and ship a change that results in a regression, our users may have to live with it for a long time. 
Speculatively estimating we have 200-300K users, are you ready to break their experience?
Granted it is broken now, however judging by a number of reports it isn't likely used by many. By accepting the fix without tests I have little confidence we are not going to degrade the experience further. Nor I have confidence the same bug won't reappear due to a sloppy refactor in a future.

It is also important to note, the lack of tests was mentioned by @gerhardol a week before my review. 
If a contributor doesn't know how to write tests - we can help, just ask. If a contributor doesn't care about tests - I'd rather not accept the contribution.

Another important note - tests are only expected for a given change. 
It is both unreasonable and unfair to expect tests for the code outside the scope of change. (Tests are welcome, and we're certainly not going to reject it, if more tests provided).

mstv:
I put it on my list because it had failed for me, too, I had found the existing issues and read that @NOYB bugfixed it in GH directly without a dev environment.

gerhardol:
It is RussKie as current leader to decide the rules and set the way forward.
(My personal opinion on argument level testing should be quite easy to find but RussKie deserves support.)

> Look at how may bugs we have in our backlog. How may of them do you think could have been avoided if we had tests in place?

A quick check of the 25 latest issues, about 16 could be considered to be bugs.
I am guessing that maybe two could have been detected with API level tests (that we have few of in GE).

Of course, the argument level tests in GE could prevent bugs from being entered in the first place. It also slows down new features, which will limit new bugs too.

gerhardol:
Closed with #6469

Issue #6357, opened 2019-03-09
UI freezes when updating submodule status


gerhardol:
## Current behaviour

When the submodule status are updated for the drop down (or side panel in master), the UI can freeze and stop to respond.

Note: There are several issues reported that are related to this already. However, most are quite generic and have been partly fixed (or a workaround applied). This issue is a fresh start to describe a (the core?) problem.

## Expected behaviour

The UI responding.

## Steps to reproduce

This is described for master, can be reproduced in a similar way for 3.0, but master (3.1) (~~even better with #6349, now merged to master~~) has better visual clues (status in left panel). In 3.0 the submodule structure and status is calculated immediately, in 3.1 the structure is calculated immediately but the status is evaluated when commit-count reports that submodules are dirty or changed.

The more submodules there is in a repo, the easier it is to reproduce. I find it easier to reproduce if I add GE as a submodule to itself 4 times. It can be reproduced with fewer and more reliably occurs with more clones, but the freeze is annoying...
The following add a clone to an existing local directory, repeat four times:
```
$ git submodule add ../gitextensions subtest/x001
```
To create many:
```
for i in `seq 1 100`; do git submodule add ../gitextensions subtest/x$i; done
```

 * Enable submodule status (require commit status) (GE 3.0 do not require commit status, submodule status queried instantly).

 * Change revision or make a submodule dirty (to initiate the submodule status when commit status sees changes)
$ touch GitExtensionsDoc/t.t

 * Show the Git Command log

View the Diff tab (to get GE to create Git commands).
View artificial commits
Reload (F5)

Watch the command log to emit updates every second (instead of many per second). If no freeze when the status markers appears (~~use #6349, now merged~~), refresh again until the problem appear.

To see that the "slow log update" is actually a problem:
When Commit button changes from gray to yellow switch between Worktree and Index artificial commits until the UI freezes. 
Note: It is easier to recreate by selecting commits, but the command cache requires that you scroll to get new commits at each try.

## Screenshots

One recording, first try experiences the problem, the second not
![ge_subm](https://user-images.githubusercontent.com/6248932/54077839-462bb600-42be-11e9-962c-4e23bf272bfb.gif)
[ge_subm.txt](https://github.com/gitextensions/gitextensions/files/2948880/ge_subm.txt)

## Did this work in previous version of GitExtensions

Working fine in 2.51

To recreate the same scenario in GE2.51:
Wait 15 s (the menu structure/status is evaluated at Browse activation, throttled)
[...]

PR #6359, opened 2019-03-10
GitModule Encoding lazy encoding

gerhardol:
Part of #6357 

## Proposed changes
- Lazy evaluation of GitModule.FileEncoding 
~~- Optimize copy GitModule copy~~ #6373

This fixes the performance problem I see in my "real" repos even if the core problem should be fixed too.

## Test methodology

See #6357 for a test setup
This is verified by git-rev-parse calls for only the modules with changes

## Test environment(s)
See #6357 , a submodule with changes required

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2309,7 +2309,7 @@ namespace GitCommands
                 cache: cache,
                 outputEncoding: LosslessEncoding);
 
-            var patches = PatchProcessor.CreatePatchesFromString(patch, encoding).ToList();
+            var patches = PatchProcessor.CreatePatchesFromString(patch, new Lazy<Encoding>(() => encoding)).ToList();
 
             return GetPatch(patches, fileName, oldFileName);
         }
@@ -2593,7 +2593,7 @@ namespace GitCommands
         }
 
         internal GitArgumentBuilder GetCurrentChangesCmd(string fileName, [CanBeNull] string oldFileName, bool staged,
-            string extraDiffArguments, Encoding encoding, bool noLocks)
+            string extraDiffArguments, bool noLocks)
         {
             return new GitArgumentBuilder("diff", gitOptions:
                     noLocks && GitVersion.Current.SupportNoOptionalLocks
@@ -2611,12 +2611,12 @@ namespace GitCommands
         }
 
         [CanBeNull]
-        public Patch GetCurrentChanges(string fileName, [CanBeNull] string oldFileName, bool staged, string extraDiffArguments, Encoding encoding, bool noLocks = false)
+        public Patch GetCurrentChanges(string fileName, [CanBeNull] string oldFileName, bool staged, string extraDiffArguments, Encoding encoding = null, bool noLocks = false)
         {
-            var output = _gitExecutable.GetOutput(GetCurrentChangesCmd(fileName, oldFileName, staged, extraDiffArguments, encoding, noLocks),
+            var output = _gitExecutable.GetOutput(GetCurrentChangesCmd(fileName, oldFileName, staged, extraDiffArguments, noLocks),
                 outputEncoding: LosslessEncoding);
 
-            var patches = PatchProcessor.CreatePatchesFromString(output, encoding).ToList();
+            IReadOnlyList<Patch> patches = PatchProcessor.CreatePatchesFromString(output, new Lazy<Encoding>(() => encoding ?? FilesEncoding)).ToList();
 
             return GetPatch(patches, fileName, oldFileName);
         }
@@ -3949,7 +3949,7 @@ namespace GitCommands
                 return "";
             }
 
-            var patches = PatchProcessor.CreatePatchesFromString(patch, encoding).ToList();
+            var patches = PatchProcessor.CreatePatchesFromString(patch, new Lazy<Encoding>(() => encoding)).ToList();
 
             return GetPatch(patches, filePath, filePath)?.Text;
         }
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -496,7 +496,7 @@ namespace GitCommands
         [CanBeNull]
         public static GitSubmoduleStatus GetCurrentSubmoduleChanges(GitModule module, string fileName, string oldFileName, bool staged, bool noLocks = false)
         {
-            Patch patch = module.GetCurrentChanges(fileName, oldFileName, staged, "", module.FilesEncoding, noLocks: noLocks);
+            Patch patch = module.GetCurrentChanges(fileName, oldFileName, staged, "", noLocks: noLocks);
             string text = patch != null ? patch.Text : "";
             return ParseSubmoduleStatus(text, module, fileName);
         }
--- a/GitCommands/Patches/PatchProcessor.cs
+++ b/GitCommands/Patches/PatchProcessor.cs
@@ -32,7 +32,7 @@ namespace GitCommands.Patches
         /// File paths can be quoted (see <c>core.quotepath</c>). They are unquoted by <see cref="GitModule.ReEncodeFileNameFromLossless"/>.
         /// </remarks>
         [NotNull, ItemNotNull, Pure]
-        public static IEnumerable<Patch> CreatePatchesFromString([NotNull] string patchText, [NotNull] Encoding filesContentEncoding)
+        public static IEnumerable<Patch> CreatePatchesFromString([NotNull] string patchText, [NotNull] Lazy<Encoding> filesContentEncoding)
         {
             // TODO encoding for each file in patch should be obtained separately from .gitattributes
 
@@ -59,7 +59,7 @@ namespace GitCommands.Patches
         }
 
         [CanBeNull]
-        private static Patch CreatePatchFromString([ItemNotNull, NotNull] string[] lines, [NotNull] Encoding filesContentEncoding, ref int lineIndex)
+        private static Patch CreatePatchFromString([ItemNotNull, NotNull] string[] lines, [NotNull] Lazy<Encoding> filesContentEncoding, ref int lineIndex)
         {
             if (lineIndex >= lines.Length)
             {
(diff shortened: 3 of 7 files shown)


gerhardol:
CodeFactor reports complexity on methods where the signature was changed only, no logic changed.
I propose those are handled separately.

GitCommands/Git/GitModule.cs
```diff
@@ -39,6 +39,16 @@ public sealed class GitModule : IGitModule
         private readonly IRevisionDiffProvider _revisionDiffProvider = new RevisionDiffProvider();
         private readonly IGitCommandRunner _gitCommandRunner;
         private readonly IExecutable _gitExecutable;
+        private readonly Lazy<Encoding> _filesEncoding;
+
+        public GitModule(GitModule module) : this(module.WorkingDir)
```
RussKie:
```suggestion
        public GitModule(IGitModule module) : this(module.WorkingDir)
```

GitCommands/Git/GitModule.cs
```diff
@@ -39,6 +39,16 @@ public sealed class GitModule : IGitModule
...
+        {
+            _gitCommonDirectory = module._gitCommonDirectory;
+            if (module._filesEncoding.IsValueCreated)
+            {
+                _filesEncoding = new Lazy<Encoding>(() => module._filesEncoding.Value);
```
RussKie: Any reason we can't write `_filesEncoding = module._filesEncoding;`?

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -100,7 +100,7 @@ public void UpdateSubmodulesStatus(bool updateStatus, string workingDirectory, s
                 // First task: Gather list of submodules on a background thread.
 
                 // Don't access Module directly because it's not thread-safe.  Use a thread-local version:
-                var threadModule = new GitModule(workingDirectory);
+                var threadModule = new GitModule(gitModule);
```
RussKie: ❗️ I don't like this very much, this may lead to more threading problems.

An instance of `GitModule` is not thread safe and it is a mixed bag of state and behaviours. That's why an instance of `GitModule` is either constructed for a given working directory or it is passed as `Func<IGitModule>`.

gerhardol: This constructor copies static fields from the existing GitModule that do not need to be queried again, no threading issue

GitCommands/Git/GitModule.cs
gerhardol: Because you cannot assign Lazy fields directly

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: Note: The second commit is not important, but it resolves one of the issues I investigated.

GitCommands/Git/GitModule.cs
gerhardol: Func<T> could be used instead

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: I changed this, to make a simple copy, as used in other similar scenarios.

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -100,7 +100,7 @@ public void UpdateSubmodulesStatus(bool updateStatus, string workingDirectory, s
                 // First task: Gather list of submodules on a background thread.
 
                 // Don't access Module directly because it's not thread-safe.  Use a thread-local version:
-                var threadModule = new GitModule(workingDirectory);
+                var threadModule = gitModule;
```
jbialobr: I think that with this change the whole Lazy changes are unnecessary. Aren't they?

I don't understand how the Lazy encoding would help, for it is always evaluated when passed through and it is already lazily created.

gerhardol: This commit (including fixup) avoids recreating the latest GitModule only, so only one git call.

The encoding is only evaluated when there are changes to submodules. The normal scenario is that there are changes to few submodules only (if there are no submodule changes, the submodule status is not evaluated at all (just structure)).
The encoding is only needed when there is a change for the submodule.

jbialobr: Thanks, I see it now. Perhaps FilesEncoding don't have to be declared as Lazy, it is enough to pass it as Lazy or Func<Encoding> where needed.

gerhardol: Func would work too, I declared it as Lazy as I found it simpler.  (Lazy is hardly usedin GE right now so it is basically a new "pattern").
Do request to change it?

jbialobr: I mean:

```csharp
public Encoding FilesEncoding => EffectiveConfigFile.FilesEncoding ?? new UTF8Encoding(false);

public static IEnumerable<Patch> 
CreatePatchesFromString([NotNull] string patchText, [NotNull] Func<Encoding> filesContentEncoding)
        

var patches = PatchProcessor.CreatePatchesFromString(output, () => encoding ?? FilesEncoding).ToList();
```
or
```csharp
public Encoding FilesEncoding => EffectiveConfigFile.FilesEncoding ?? new UTF8Encoding(false);

public static IEnumerable<Patch> 
CreatePatchesFromString([NotNull] string patchText, [NotNull] Lazy<Encoding> filesContentEncoding)
        

var patches = PatchProcessor.CreatePatchesFromString(output, new Lazy(() => encoding ?? FilesEncoding)).ToList();


```

RussKie: I don't like this change. I don't like passing an instance of `GitModule` around because it is not thread safe.
Line 102 also states "Don't access Module directly because it's not thread-safe.  Use a thread-local version".

gerhardol: This is how the thread local version is created everywhere else, like in GitStatusMonitor. Before the fix up commit an optimized copy was used instead, we can apply that everywhere modules are used in threads instead.

gerhardol: I prefer to use Lazy. Will change to your proposal, it further emphasis why the change is done. If most "expensive" fields were Lazy, it would be a different setup.

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: Submitted #6373 for this

GitCommands/Git/GitModule.cs
gerhardol: This specific situation changed previously, also changed Lazy handling

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: Changed - fixup

gerhardol: Moved this to #6373

gerhardol:
@RussKie @amaiorano 
Any comments? I would like to merge this and base the changes @jbialobr is doing on top of it
(this change basically solves the problem for me with a little under 20 submodules in total (the problem itself can easily be reproduced with more submodules, but this helps).

amaiorano:
> @RussKie @amaiorano
> Any comments? I would like to merge this and base the changes @jbialobr is doing on top of it
> (this change basically solves the problem for me with a little under 20 submodules in total (the problem itself can easily be reproduced with more submodules, but this helps).

I've looked over the diffs here in the browser, and I don't really understand this change. For instance, I see that ```GitModule.GetSingleDiff``` accepts an Encoding, and from that point on, it passes down a Lazy<Encoding> to ```PatchProcessor.CreatePatchesFromString```, which eventually gets down to ```PatchProcessor.CreatePatchFromString```, which might invoke filesContentEncoding.Value. What I don't understand is how this helps performance; isn't Encoding just a reference type? Passing it down the call stack shouldn't have any extra cost, right? I'm sure I'm just not seeing it...

gerhardol:
> What I don't understand is how this helps performance

When Encoding is passed as a parameter, it is evaluated directly when calling, which requires a git-rev-parse call. However, Encoding is only needed when there are actual differences, so most Encoding calculations are not needed.

If Lazy is used, it is evaluated when it is first used.

amaiorano:
> When Encoding is passed as a parameter, it is evaluated directly when calling, which requires a git-rev-parse call. However, Encoding is only needed when there are actual differences, so most Encoding calculations are not needed.
> 
> If Lazy is used, it is evaluated when it is first used.

You mean that for every function call where Encoding was being passed, it gets evaluated? Like when it first gets passed to ```GitModule.GetSingleDiff```, is it evaluated then? And when it then gets passed to ```PatchProcessor.CreatePatchesFromString``` (before your change), it gets evaluated again? And what does evaluation mean? What gets called on Encoding? I don't understand because Encoding is a class, therefore a reference type, so passing an instance just means passing a reference around, right?

I feel like I'm missing something super obvious... sorry.

vbjay:
https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1

amaiorano:
> https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1

Yes, I know what Lazy<T> does, but in the specific use case I outline above, we create a Lazy around an instance that already exists and has been passed in:

```csharp
            var patches = PatchProcessor.CreatePatchesFromString(patch, new Lazy<Encoding>(() => encoding)).ToList();
```

Normally, Lazy<T> is provided with a factory function to create an instance lazily. That's what I don't understand here. If the instance has already been created, then what are we saving on? Passing a reference type as argument doesn't cost anything.

gerhardol:
@amaiorano 

```
public Encoding FilesEncoding => EffectiveConfigFile.FilesEncoding ?? new UTF8Encoding(false);
```
The EffectiveConfigFile getter will require git-rev-parse to get the current config.

As the original code passed FilesEncoding reference, EffectiveConfigFile had to be evaluated.

https://github.com/gitextensions/gitextensions/pull/6359/files#diff-05956b8e9b35344894a0ffe609cf32b9L2619

If the value is needed, not much other than to calculate it. So if all submodules are changed, this PR will not improve anything. However, the value is only required when there are changes (and the normal scenario with submodules is few changes, otherwise submodules are a pain):

https://github.com/gitextensions/gitextensions/pull/6359/files#diff-d832ec87de5903d3b8ffd40176d06050L273

For other than submodules, GE normally knows when there are changes. For submodules we cannot know (without calling other commands first).

So if a

gerhardol:
Unless there are protests, I will merge this tomorrow

RussKie:
:+1: from me

PR #6362, opened 2019-03-10
Support Git pre 2.11 for status

### LINKED ISSUE #6322: Commits do not show after updating Git Extensions
## Current behaviour

I decided to upgrade my Git Extensions to version 3.0.2.5232 from version 2.5.1. After doing this, I was prompted to upgrade Git as well to have a more compatible version (upgrade higher than 2.9.0

image

When trying to upgrade, I am blocked by my organization. I have coworkers who do not have this issue though, even though they did not upgrade Git itself. Only Git Extensions. When I ignore this prompt and continue using Git Extensions, I am able to push and pull code just fine but I am not able to see my most recent changes underneath my Commits. Instead I see this:

image

How can I get my commits and diffs to display here?

## Expected behaviour

Commits and diffs display after upgrading to Git Extensions 3.0.2.5232 without upgrading git from 2.9.0

## Steps to reproduce

Follow prompts to upgrade to the latest version of Git Extensions. Do not upgrade Git.

## Screenshots

See screen shots above

Not Sure.

## Environment

- Git Extensions 3.0.2.5232
- Git 2.9.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

gerhardol:
Resolves #6322

## Proposed changes
Fix the options for Git prior to 2.11

Note that Git prior to 2.11 is not officially supported, this patch will not change that.
However, the error messages will be less confusing and the code can be used as an example.
This do not need to be included in Release Notes.

## Test methodology 

Primarily: Test that Git 2.21.0 works
Also debugging in the code, viewing the command line argument

## Test environment(s)

- GIT 2.21.0

### DIFF
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -486,7 +486,7 @@ namespace GitCommands
                     ? (ArgumentString)"--no-optional-locks"
                     : default)
             {
-                $"--porcelain={(GitVersion.Current.SupportStatusPorcelainV2 ? 2 : 1)} -z",
+                $"--porcelain{(GitVersion.Current.SupportStatusPorcelainV2 ? "=2" : "")} -z",
                 untrackedFiles,
                 ignoreSubmodules,
                 { !excludeIgnoredFiles, "--ignored" }


RussKie:
LGTM, though I'd like to get a little more information.

The commit message is confusing me - the fix is for 2.11- versions, yet you tested it with 2.21.

> the error messages will be less confusing and the code can be used as an example.

Could you please elaborate on what error message a user would get before and after the fix.

Thank you

gerhardol:
> Could you please elaborate on what error message a user would get before and after the fix.

See #6322

gerhardol:
Will merge this minor change tomorrow unless there are objections

PR #5181, opened 2018-07-20
Provide capability to generate git -c calls in git commands

vbjay:
Fixes #5179

Changes proposed in this pull request:
- Add --no-show-signature option to log calls
 
Screenshots before and after (if PR changes UI):

What did I do to test the code and ensure quality:
- Signed a commit and made sure it showed in revision grid.  
- Verified log calls were not broken

Has been tested on (remove any that don't apply):
- GIT 2.18 and above

PR #6363, opened 2019-03-10
Remove SSH options from Installer

### LINKED ISSUE #6286: Setup can't keep "Other ssh client" setting
In the setup dialog I'm asked whether I want to choose OpenSSH or PuTTY. But I want to use the new built-in OpenSSH for Windows (`C:\Windows\System32\OpenSSH\ssh.exe`), which I have set up as "Other ssh client".

Whenever I update to a newer version, the setup asks me again, and whichever of the two options I select, my current setting is not kept.

There should either be an option to just keep the current setting, or (even better) that particular dialog page should just not be shown for upgrade installations, because it worked fine before the upgrade and there's no reason to change that setting during an upgrade.

gerhardol:
Fixes #6286 
Part of #5181

## Proposed changes

The VSIX installer currently sets ssh option to PuTTY both at initial install and at upgrades, regardless of what the user has set. This removes the SSH page, not changing the current setting.
OpenSSH is already the default in the portable version.

Note: #5181 proposes to remove the binary too. That is a second step that will require migration.

This will require changes to the documentation too.

### Before

The following appears:
https://github.com/gitextensions/GitExtensionsDoc/blob/master/source/images/install/ssh.png

### After

no SSH page

## Test methodology

- Update existing installation - no SSH page and no change of defaults
- New installation - default set to OpenSSH

## Test environment(s) 
n/a

----


RussKie:
I've always been using OpenSSH and never used Putty...

To confirm:
_As a user installing GE from afresh, would Putty be my only choice?_

vbjay:
No. It has always been a choice. Frankly fur windows it provides a better
UI and when you have protected keys it wutks better. Openssh is a pain with
protected keys.

RussKie:
> Frankly for windows it provides a better UI and when you have protected keys it works better.

Perhaps, though I never needed those.

I don't particularly want to derail into a discussion about pros and cons of each SSH tool, I'd like to better understand the UX after this change is merged.

gerhardol:
> I don't particularly want to derail into a discussion about pros and cons of each SSH tool, I'd like to better understand the UX after this change is merged.

This removes the SSH page from the installer

This removes setting to PuTTY in the installer.
 * At upgrades, it changes to PuTTY right now, so you have to change back to OpenSSH in the menu or go to GE Settings with Other
 * At installation, it uses OpenSSH by default, which is more reasonable as default (no requirement to run PageANT in the background).

vbjay:
Fine. However. My point is that we autostart the agent.  Point 2 is dead wrong. That is an opinion. The agent runs in the background. I do a pull. The password input for the key automatically opens. I can kill the agent any time. I frankly don't care to the my key password each time. I challenge you to use putty for a week with a remote set to use the putty key and see if it is such a problem. Make sure you secure it with a password.

vbjay:
The fix shouldn't be removal but instead read current setting on upgrade.  Leave putty default on new install.

vbjay:
Let me just how you the putty workflow.  

No agent running.  

- Setup remote one time
image
- Cause agent to be needed.  Push, pull, fetch a remote that is configured with ssh key.  
image
image

vbjay:
Before this is approved, I ask that both of you use putty for a week.  See how it works.  Actually use the feature before acting like it is a thorn.

gerhardol:
> Before this is approved, I ask that both of you use putty for a week. 

I have tried to use Pageant but have not seen any benefit.
No one where I work use PuTTY, it mostly is seen as a nuisance (primarily due to 1. that we agree on though, this is still a minor point for setup.)

@vbjay I see your point but still believe that having OpenSSH as the default is the way forward.
An alternative is to confirm the choice of SSH after first setup.

vbjay:
The benefit is when you are doing a set of work and need to do multiple fetch pull push. I use secured keys. I unlock at first remote action do the work and close agent when I'm done.

RussKie:
In 5 years I haven't seen a single dev in my surrounds using pageant. 
That's not to say it is not useful (clearly it is for some), but the question of how many users indeed use it.

We need telemetry data... 🤔

gerhardol:
I still want to merge this for 3.2.
I could modify this to keep the option during initial install.

RussKie:
Sorry Gerhard, I am against removing it. I don't use Putty and I haven't worked with anyone who did.
I'm happy to remove it after we collect some telemetry data and find that it is used by, say, only 30% of consumers (that we collect).

If your wish to remove the setting stems from a fact the installer doesn't read user's configured settings, then perhaps we should fix that, rather than remove the setting completely.

Setup/UI/WixUI.wxs
```diff
@@ -50,20 +50,10 @@
       <Publish Dialog="GitEx_InstallDirDlg" Control="Next" Event="SpawnDialog" Value="InvalidDirDlg" Order="3"><![CDATA[NOT WIXUI_DONTVALIDATEPATH AND WIXUI_INSTALLDIR_VALID<>"1"]]></Publish>
       <Publish Dialog="GitEx_InstallDirDlg" Control="Next" Event="NewDialog" Value="GitEx_CustomizeDlg" Order="4">WIXUI_DONTVALIDATEPATH OR WIXUI_INSTALLDIR_VALID="1"</Publish>
 
-      <Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_MaintenanceTypeDlg" Order="1">Installed</Publish>
-      <Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_InstallDirDlg" Order="2">NOT Installed</Publish>
-      <Publish Dialog="GitEx_CustomizeDlg" Control="Next" Event="NewDialog" Value="GitEx_VerifyReadyDlg" Order="1">Installed</Publish>
```
gerhardol: @Johno-ACSLive Do you know how this should be handled, currently "NOT Installed" is always selected (this PR is postponed, just changing the behavior at upgrades)

Johno-ACSLive:
> Sorry Gerhard, I am against removing it. I don't use Putty and I haven't worked with anyone who did.
> I'm happy to remove it after we collect some telemetry data and find that it is used by, say, only 30% of consumers (that we collect).
> 
> If your wish to remove the setting stems from a fact the installer doesn't read user's configured settings, then perhaps we should fix that, rather than remove the setting completely.

I agree, i use Putty simply because i have it installed and use it to remotely connect to Linux servers. I'd rather not have multiple SSH tools. During a fresh install i told GE to use Putty.

During upgrade, all settings should be preserved and the installer should not prompt unless there is a major change requiring some input. If a user has chosen a default settings and we replace a default tool then this is not a scenario to prompt. If the user has changed the default then this may be a reason to ask for input (depends on the config item in question).

I believe there is an issue for this which i was going to look into.

Setup/UI/WixUI.wxs
Johno-ACSLive: Let me look into this, there are a few rules that impact behaviour

Johno-ACSLive: I just cloned from master branch and built the installer, it seems to ask all the questions it would on fresh install.

The installer is detecting the current version, if i leave the version attributes to 3.1.0 it complains but if i set to 3.2.0 i can see it checks and then moves on. This suggests that the installer detection result is not being passed onto later stages of the process (or for some reason they are ignoring the result).

Let me dig a little further, I'm not a fan of WiX 😜 

Creating installers should be a breeze (well it usually is with other tools 😃)

I have looked at porting to Advanced Installer, the WiX project fails to convert though. It seems to allow me to continue but I don't really trust the results. I have reported this to the Advanced Installer team, they have been able to reproduce the issue and hopefully will resolve it.

Anyway that's a little off topic.

RussKie:
👍

gerhardol:
This pr is about removing putty questions from the installer, setting openssh as the default. This is a step to unbundle putty binaries from ge without too complex dependencies. 

The bug is that ssh options page is shown also at upgrades nuking the user settings.

Johno-ACSLive:
Hmm the more I dig into the MSI project the more I find bad things. This installer needs a lot of work / clean-up.

However, back on topic, this is taking time for me to locate everything on how WiX was configured to run through it's rules. I'll provide an update once I know more.

@RussKie is saying not to remove Putty just yet. I agree that unless you have telemetry around it's usage, there is no point removing it as you won't know what kind of impact that will have on the user base.

If the bug is that the installer is changing the users settings on upgrade (which is the assumption i was working with), then let's address this first. The bug doesn't seem to be specific to SSH tools. The upgrader prompts for per user/per machine config, location etc.

I think we need a new PR for an installer overhaul because the upgrader is prompting like a brand new install (not to mention alot of other things i have found that are bad with the installer). We should ideally have a 32-bit MSI and 64-bit MSI (rather than mixing everything into a 32-bit MSI).

I have started working on a fresh Advanced Installer project to replicate the current WiX project to clean things up somewhat and I'll try seperate that out as described above (unless anyone has background on why we shouldn't have a 32-bit and 64-bit installer).

Johno-ACSLive:
Ok so part one of the change will be to add an additional condition on these two components:

`<Component Id="gitssh_openssh.reg" Guid="5f730698-ac38-4e63-ad5d-505b21d1fa22">
        <Condition>SSHCLIENT="OpenSSH"</Condition>
        <RegistryValue Name="gitssh" Value="" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
  </Component>`

`<Component Id="gitssh_putty.reg" Guid="2cb64776-8035-413e-9266-4ad9e5ecbad3">
        <Condition>SSHCLIENT="PuTTY"</Condition>
        <RegistryValue Name="gitssh" Value="[INSTALLDIR]PuTTY\plink.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
      </Component>`

If you don't show the dialog, the installer will still replace the config from what i can see (the SSHClient is set to PuTTy by default). The condition for each component should check that SSHClient property is the appropriate value AND the regkey does not exist. This approach is hacky but i guess doesn't really matter for upgrade, user can change their SSH preference in GE.

I'll post a follow up for disabling the UI.

RussKie:
> Hmm the more I dig into the MSI project the more I find bad things. This installer needs a lot of work / clean-up.

Please open a new issue and let's track it there

Johno-ACSLive:
> > Hmm the more I dig into the MSI project the more I find bad things. This installer needs a lot of work / clean-up.
> 
> Please open a new issue and let's track it there

Of course

Johno-ACSLive:
OK part two, I have not tested on a machine where GE is not installed. However for upgrade to skip SSH options UI, open WixUI.wxs and ensure lines 53 to 62 have the same configuration as the below lines.

`<Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_MaintenanceTypeDlg" Order="1">Installed</Publish>`
 
`<Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_InstallDirDlg" Order="2">NOT Installed</Publish>`

`<Publish Dialog="GitEx_CustomizeDlg" Control="Next" Event="NewDialog" Value="GitEx_VerifyReadyDlg" Order="1">NOT Installed</Publish>`
 
`<Publish Dialog="GitEx_CustomizeDlg" Control="Next" Event="NewDialog" Value="GitEx_SSHClientDlg" Order="2">Installed</Publish>`

`<Publish Dialog="GitEx_SSHClientDlg" Control="Back" Event="NewDialog" Value="GitEx_CustomizeDlg">1</Publish>`
 
`<Publish Dialog="GitEx_SSHClientDlg" Control="Next" Event="NewDialog" Value="GitEx_VerifyReadyDlg">1</Publish>`

`<Publish Dialog="GitEx_VerifyReadyDlg" Control="Back" Event="NewDialog" Value="GitEx_SSHClientDlg" Order="1">Installed</Publish>`

`<Publish Dialog="GitEx_VerifyReadyDlg" Control="Back" Event="NewDialog" Value="GitEx_CustomizeDlg" Order="2">NOT Installed</Publish>`

gerhardol:
@Johno-ACSLive 
Thanks, but I do not get that working.
Part1 is already in WixUI.wxs. If I change part2 (basically switching the conditions on line 55 and 56) the question does not appear neither for upgrades or new installations. The setting is always changed to PuTTY too.

I have no intention to remove PuTTY support. This PR just removes the page in the installer.

...and as you can see there are not good knowledge about the installer in the team.

Johno-ACSLive:
> @Johno-ACSLive
> Thanks, but I do not get that working.
> Part1 is already in WixUI.wxs. If I change part2 (basically switching the conditions on line 55 and 56) the question does not appear neither for upgrades or new installations. The setting is always changed to PuTTY too.
> 
> I have no intention to remove PuTTY support. This PR just removes the page in the installer.
> 
> ...and as you can see there are not good knowledge about the installer in the team.

For Part 1, I see you have changed the condition but it does not match the condition i mentioned. You have also removed the Putty registration. Both need to be restored to how they were and only the condition on both need to be modified as i described.

Based on your current changes, this will not allow Putty to be registered correctly on a default new install (and will default to OpenSSH even though the SSHClient property is set to Putty).

As for the page not displaying on install, this is odd as WiX's documentation doesn't go into much detail. The problem is that WiX's documentation is missing alot of detail and the tool has a steep learning curve.

Hence why I'm looking to see if we can change tools (I will raise a new issue / PR for this). I think its best that minimal time is spent on installers, knowing MSI and MSI rules aside, we shouldn't need to spend more time trying to fight with an installer tool then getting enhancements / fixes added to GE.

Let me see if i can find any info on what WiX thinks it's trying to do and I'll get back to you on an update for part two!

Johno-ACSLive:
OK it seems that NOT Installed is true on major upgrade. When a new installer is generated, a new product code is generated indicating it is a major upgrade (minor upgrades i believe are only applicable when MSP's or patches are generated).

I'll find out what other properties we can use for this check.

Johno-ACSLive:
OK, if you change lines 53 to 64 with what i have below, it will work as expected for upgrade and new install. While the conditions have changed, I have fixed up some formatting issues and changed a couple parameters as they didn't look right.

      <Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_MaintenanceTypeDlg" Order="1">Installed OR WIX_UPGRADE_DETECTED</Publish>
      <Publish Dialog="GitEx_CustomizeDlg" Control="Back" Event="NewDialog" Value="GitEx_InstallDirDlg" Order="2">NOT Installed AND NOT WIX_UPGRADE_DETECTED</Publish>
      <Publish Dialog="GitEx_CustomizeDlg" Control="Next" Event="NewDialog" Value="GitEx_VerifyReadyDlg" Order="1">Installed OR WIX_UPGRADE_DETECTED</Publish>
      <Publish Dialog="GitEx_CustomizeDlg" Control="Next" Event="NewDialog" Value="GitEx_SSHClientDlg" Order="2">NOT Installed AND NOT WIX_UPGRADE_DETECTED</Publish>

      <Publish Dialog="GitEx_SSHClientDlg" Control="Back" Event="NewDialog" Value="GitEx_CustomizeDlg">1</Publish>
      <Publish Dialog="GitEx_SSHClientDlg" Control="Next" Event="NewDialog" Value="GitEx_VerifyReadyDlg">1</Publish>

      <Publish Dialog="GitEx_VerifyReadyDlg" Control="Back" Event="NewDialog" Value="GitEx_SSHClientDlg" Order="1">NOT Installed AND NOT WIX_UPGRADE_DETECTED</Publish>
      <Publish Dialog="GitEx_VerifyReadyDlg" Control="Back" Event="NewDialog" Value="GitEx_CustomizeDlg" Order="2">Installed OR WIX_UPGRADE_DETECTED</Publish>

      <Publish Dialog="GitEx_MaintenanceWelcomeDlg" Control="Next" Event="NewDialog" Value="GitEx_MaintenanceTypeDlg">1</Publish>

gerhardol:
> For Part 1, I see you have changed the condition but it does not match the condition i mentioned.

I tested against current master, not to this branch. 
If you believe replacing Wix is better, that would be the way forward. To avoid changing the SSH option would be a bugfix though.
I will look into this later.

RussKie:
Closing until we get more telemetry that suggests SSH is under-used.

gerhardol:
Will reopen when telemetry shows that putty usage is too low to set as the default (and force at upgrades) and to have in the installer at all.

RussKie:
Aggregated telemetry data for the past 90 days, it will surprise you @gerhardol :)



ssh | perc | sshCount
-- | --: | --: 
| OpenSSH | 65.43% | 13,492 |  
| PuTTY | 34.23% | 7,058 |  
| Other | 0.34% | 71

gerhardol:
>  it will surprise you

Yes, I am surprised. Is the low telemetry participation skewing the data somehow?

However, it is still 2 to 1 for OpenSSH and there is no reason to overwrite the default setting.

RussKie:
> Is the low telemetry participation skewing the data somehow?

The low amount of telemetry is certainly a factor. 
Whether the results are skewed or not - we won't know unless we have 100% participation rate and all users perform exactly the same number of requests. We don't have no way to identify users, and a handful of super-busy dudes and dudettes using OpenSSH can outweigh hundreds of not-so-busy Putty users.

So all we can say now is "2/3rds of of all requests we captured were performed via Git Extensions configured to use OpenSSH".

So dropping OpenSSH from the installer would likely be a very unpopular move. Dropping PuTTY may not be as bad, but would likely affect a significant portion of our users too.

> no reason to overwrite the default setting.

Totally agree, we need to fix the installer to select a currently configured setting, if possible.
Perhaps it can be done as part of #7368

Issue #6225, opened 2019-02-08
Frozen once staging 1000 files with CRLF line break and core.autocrlf = input


Denitz:
## Current behaviour

Updated repo, there are 1000 updated files, trying to stage all, program is frozen and can be terminated only.

## Expected behaviour

Just stage all 1000 files.

## Steps to reproduce

Try to stage 1000 files with CRLF line break while you have core.autocrlf = input

## Did this work in previous version of GitExtensions

Yes, before 3.0

## Environment

- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

steeintw:
I have the same problem while staging about 400+ files using "Stage All" in "Commit" dialog.
My core.autocrlf =true.
Staging one file at a time works fine.
Use "Git Extensions -> Add Files" from Windows Explorer context menu also works fine.

RussKie:
Do you have a repro repo to share?

mast-eu:
I observed this issue as well. But I believe the key is not only the sheer number of files.
At least in my case, I rather think it's the fact that I copied a bunch of files from a Linux system (LF line endings) to a Windows system (CRLF line endings) and made GitExtensions substitute the line endings during staging.

I think I can make this reproducible and provide an example repo.

mast-eu:
### Reproduction repository
https://github.com/mast-eu/GitExtensions.issue6225.git

### Reproduction steps
1. Make sure you have set "Checkout as-is, commit as-is" for line endings.
![lineendings](https://user-images.githubusercontent.com/46861028/53016044-d8256900-344c-11e9-9bcb-fc1507c85c90.png)
2. Clone the repo given above and open it in GitExtensions.
It contains 3 folders `files_with_CR`, `files_with_CRLF` or `files_with_LF`, each one containing 1024 files with the respective line ending.
3. In Windows Explorer, copy one of the folders, in order to make them appear as 1024 new files for git. For example:
![lineendings2](https://user-images.githubusercontent.com/46861028/53016045-d8256900-344c-11e9-9ad0-bcb57f1408b0.png)
4. Open the commit dialog and click on `Stage all`. It works, if you did no yet change the setting for line endings.
[...]

PR #6366, opened 2019-03-11
Avoid freeze when staging many files at once (works only when AppSettings.ShowErrorsWhenStagingFiles==false)

mast-eu:
Fixes part of #6225

## Proposed changes

- Changed how `AppSettings.ShowErrorsWhenStagingFiles=false` works (flag for skipping "warning: LF will be replaced by CRLF" messages). Instead of simply not showing the warning messages after the git command has been executed, pass an additional option to git to not even generate the warning messages.
Advantages:
  - A cleaner implementation.
  - It provides at least a workaround for #6225. Setting `AppSettings.ShowErrorsWhenStagingFiles` to false will allow to stage any number of files, regardless of their line ending and the user's `core.autocrlf` setting. The freeze remains though with `AppSettings.ShowErrorsWhenStagingFiles` true (the default value).

## Test methodology

- manual tests, in particular with the reproduction repo and steps described in https://github.com/gitextensions/gitextensions/issues/6225#issuecomment-465784080

## Test environment(s)
- Git Extensions 3.1.0
- Build c68cf2382d5f5f1d33855619ec37f14d4bfc654b
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 120dpi (125% scaling)

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1747,11 +1747,7 @@ namespace GitCommands
             if (nonDeletedFiles.Count != 0)
             {
                 var execution = _gitExecutable.Execute(
-                    new GitArgumentBuilder("update-index")
-                    {
-                        "--add",
-                        "--stdin"
-                    },
+                    UpdateIndexCmd(AppSettings.ShowErrorsWhenStagingFiles),
                     inputWriter =>
                     {
                         foreach (var file in nonDeletedFiles)
@@ -1889,6 +1885,18 @@ namespace GitCommands
                 $"\"{filename.ToPosixPath()}\"{inputWriter.NewLine}");
 
             inputWriter.BaseStream.Write(bytes, 0, bytes.Length);
+        }
+
+        private GitArgumentBuilder UpdateIndexCmd(bool showErrorsWhenStagingFiles)
+        {
+            return new GitArgumentBuilder("update-index", gitOptions:
+                            showErrorsWhenStagingFiles
+                                ? default
+                                : (ArgumentString)"-c core.safecrlf=false")
+                    {
+                        "--add",
+                        "--stdin"
+                    };
         }
 
         #endregion
@@ -4057,6 +4065,20 @@ namespace GitCommands
                     ? $"{param}=\"{date:yyyy-MM-dd hh:mm:ss}\""
                     : "";
             }
+        }
+
+        internal TestAccessor GetTestAccessor() => new TestAccessor(this);
+
+        internal readonly struct TestAccessor
+        {
+            private readonly GitModule _gitModule;
+
+            public TestAccessor(GitModule gitModule)
+            {
+                _gitModule = gitModule;
+            }
+
+            public GitArgumentBuilder UpdateIndexCmd(bool showErrorsWhenStagingFiles) => _gitModule.UpdateIndexCmd(showErrorsWhenStagingFiles);
         }
     }
 
--- a/UnitTests/GitCommandsTests/GitModuleTest.cs
+++ b/UnitTests/GitCommandsTests/GitModuleTest.cs
@@ -579,5 +579,15 @@ namespace GitCommandsTests
         {
             Assert.AreEqual(expected, _gitModule.GetRefsCmd(tags, branches, noLocks).ToString());
         }
+
+        [TestCase(false, @"-c core.safecrlf=false update-index --add --stdin")]
+        [TestCase(true, @"update-index --add --stdin")]
+        public void UpdateIndexCmd_should_add_core_safecrlf(bool showErrorsWhenStagingFiles, string expected)
+        {
+            var accessor = _gitModule.GetTestAccessor();
+
+            var actual = accessor.UpdateIndexCmd(showErrorsWhenStagingFiles).ToString();
+            Assert.AreEqual(expected, actual);
+        }
     }
 }


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1883,26 +1883,19 @@ private void Stage(IReadOnlyList<GitItemStatus> items)
                     }
 
                     bool wereErrors = false;
-                    if (AppSettings.ShowErrorsWhenStagingFiles)
```
gerhardol: I am not convinced this check should be removed, other errors should be possible

mast-eu: I just checked the [documentation](https://git-extensions-documentation.readthedocs.io/en/latest/settings.html#git-extensions-detailed-commit-dialog):
> Show errors when staging files
If an error occurs when files are staged (in the Commit dialog), then the process dialog showing the results of the git command is shown if this setting is checked.

Since it speaks generally of "an error" I guess the real intention of this flag is "any kind of error", not only those specific to line endings. So you're right that this check should not be removed.

mast-eu: Done.

gerhardol:
@RussKie I see no obvious test here. OK to merge?

RussKie:
> I see no obvious test here.

With a little refactor we can test this :)

Extract the command building into own method:
```
private GitArgumentBuilder BuildUpdateIndexCommand(bool showErrorsWhenStagingFiles)
{
    return new GitArgumentBuilder("update-index", gitOptions:
                    showErrorsWhenStagingFiles
                        ? default
                        : (ArgumentString)"-c core.safecrlf=false")
                    {
                        "--add",
                        "--stdin"
                    });
}
```
Then update the callsite:
```
                var execution = _gitExecutable.Execute(
                    BuildUpdateIndexCommand(AppSettings.ShowErrorsWhenStagingFiles),
                    inputWriter =>
                    {
                        foreach (var file in nonDeletedFiles)
                        {
                            UpdateIndex(inputWriter, file.Name);
                        }
                    },
                    SystemEncoding);
```
And then we can test `BuildUpdateIndexCommand` via the `TestAccessor` pattern.

@mast-eu do you think you could do that?

mast-eu:
> do you think you could do that?

Yes. Thank you for the suggestion. I will add it.

mast-eu:
@RussKie : Done, please have a look.
I named the method `UpdateIndexCmd` instead of `BuildUpdateIndexCommand`, because I saw that GitModule.cs already contains many methods with this nomenclature. Apart from that, I followed your suggestion.

GitCommands/Git/GitModule.cs
```diff
@@ -1890,6 +1886,18 @@ private static void UpdateIndex(StreamWriter inputWriter, string filename)
             inputWriter.BaseStream.Write(bytes, 0, bytes.Length);
         }
 
+        public GitArgumentBuilder UpdateIndexCmd(bool showErrorsWhenStagingFiles)
```
RussKie: Please make it `private`, we don't want to increase the public surface of this class.
You can test the method via a [`TestAccessor`](https://github.com/gitextensions/gitextensions/search?q=testaccessor&unscoped_q=testaccessor)

UnitTests/GitCommandsTests/GitModuleTest.cs
```diff
@@ -572,5 +572,13 @@ public void GetRefsCmd(string expected, bool tags, bool branches, bool noLocks)
         {
             Assert.AreEqual(expected, _gitModule.GetRefsCmd(tags, branches, noLocks).ToString());
         }
+
+        [TestCase(false, @"-c core.safecrlf=false update-index --add --stdin")]
+        [TestCase(true, @"update-index --add --stdin")]
+        public void UpdateIndexCmd_should_handle_core_safecrlf(bool showErrorsWhenStagingFiles, string expected)
+        {
+            var actual = _gitModule.UpdateIndexCmd(showErrorsWhenStagingFiles).ToString();
+            StringAssert.AreEqualIgnoringCase(expected, actual);
```
RussKie: Git commands are case sensitive, so the assertion should be too

mast-eu: You're right.

GitCommands/Git/GitModule.cs
mast-eu: Sorry, the "`TestAccessor` pattern" was the missing piece that I didn't fully understand. Thus my "workaround" of making the method public.
I think I got it right this time (or at least I'm getting closer :smirk: ). Please check again.

GitCommands/Git/GitModule.cs
```diff
@@ -4058,6 +4066,23 @@ string GetDateParameter(string param, DateTime? date)
...
+
+            public TestAccessor(GitModule gitModule)
+            {
+                _gitModule = gitModule;
+            }
+
+            public GitArgumentBuilder UpdateIndexCmd(bool showErrorsWhenStagingFiles)
+            {
+                return _gitModule.UpdateIndexCmd(showErrorsWhenStagingFiles);
+            }
```
RussKie: [nit] this could be simplified to 
```
public GitArgumentBuilder UpdateIndexCmd(bool showErrorsWhenStagingFiles) => _gitModule.UpdateIndexCmd(showErrorsWhenStagingFiles);
```

mast-eu: Than the same simplification could be used here as well, or am I wrong?
https://github.com/gitextensions/gitextensions/blob/e920a4916efaf0779491715fc07ee3825676e7ed/GitUI/Editor/FileViewer.cs#L1532-L1545
I recently added these in #6397, and re-used the same approach again.

To make things right I would send a dedicated PR to apply the same simplification to all 4 methods, since I don't want to mix changes relative to a different PR in here. Are you ok with that?

RussKie: Yes 👍

mast-eu: I force pushed the change in `GitModule.cs` to this PR and opened #6433 for `FileViewer.cs`.

Issue #5769, opened 2018-11-13
Improve Submodule Status performance


gerhardol:
**Do you want to request a *feature* or report a *bug*?**
"feature"

**What is the current behavior?**
Obtaining the submodule status may require a lot of resources, as in #5733. 
It is possible to deactivate the status (#5756), a few very minor improvements too (#5747 and #5742).
The performance can be improved, some discussions in the issues/PRs. This is a proposal on the way forward.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**

Some possible improvements, listed in some kind of order:
- [x] #5777 Only update submodule status when there are changes (when commit-count indicate submodule status)
- [x] #5983 Run the submodule diff with --no-optional-locks (set a 'background' attribute to GitModule?)
- [x] Trigger submodule status update from commit-count only. (Limited effect before splitting the structure/status.) This can be done on commit-count only, removing the delay in #5747.
- [x] #6445 Separate obtaining structure (that should be done soon after switching repo and refresh only, maybe at submodule form too but ignore external changes). This will give a quick setup of the structure used in the menu (and later in #5617) without requiring the status to be obtained directly.
- [x] #6445 Only calculate the status for submodules that are dirty/changed, known from the commit-count status (this really requires Git 2.15 or something). This should prune the number of commands considerably - the normal use case with submodules are to update few of the submodules.
~~- [ ] #6373 Optimize creation of GitModule, much data can be copied when creating GitModule for super/submodules. (At least git ref-parse --git-common-dir is required, the about 50 ms increases to over 300 ms with simultaneous invocations).~~

See also #6372, improved the overall responsiveness

**Environment you encounter the issue:**

- GitExtensions version: master

**Did this work in previous version of GitExtensions (which)?**
2.x with status worked slightly better than 3.x with status, 3.x much snappier without status.

RussKie:
Can any of the items be ticked off?

gerhardol:
#6445 is still open (after the structure change the missing tests were easily spotted, however they are failing now).

gerhardol:
Closed by #6445

PR #6372, opened 2019-03-16
Serialize submodules info loading

### LINKED ISSUE #6357: UI freezes when updating submodule status

jbialobr:
Fixes #6357

## Proposed changes

- Load submodules info with async api. It was not enough, the lag was smaller but still significant.
- Load submodules info sequentially - this helped.

### Before

image

### After

image

## Test methodology 

- Manual tests

----

### DIFF
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
@@ -183,8 +184,6 @@ namespace GitCommands
             Encoding outputEncoding = null,
             bool stripAnsiEscapeCodes = true)
         {
-            // TODO make this method async, maybe via IAsyncEnumerable<...>?
-
             if (outputEncoding == null)
             {
                 outputEncoding = _defaultOutputEncoding.Value;
@@ -224,6 +223,27 @@ namespace GitCommands
 
                 process.WaitForExit();
             }
+        }
+
+        /// <summary>
+        /// Launches a process for the executable and returns output lines as they become available.
+        /// </summary>
+        /// <param name="executable">The executable from which to launch a process.</param>
+        /// <param name="arguments">The arguments to pass to the executable</param>
+        /// <param name="writeInput">A callback that writes bytes to the process's standard input stream, or <c>null</c> if no input is required.</param>
+        /// <param name="outputEncoding">The text encoding to use when decoding bytes read from the process's standard output and standard error streams, or <c>null</c> if the default encoding is to be used.</param>
+        /// <param name="stripAnsiEscapeCodes">A flag indicating whether ANSI escape codes should be removed from output strings.</param>
+        /// <returns>An enumerable sequence of lines that yields lines as they become available. Lines from standard output are returned first, followed by lines from standard error.</returns>
+        [MustUseReturnValue("If output lines are not required, use " + nameof(RunCommand) + " instead")]
+        public static async Task<IEnumerable<string>> GetOutputLinesAsync(
+            this IExecutable executable,
+            ArgumentString arguments = default,
+            Action<StreamWriter> writeInput = null,
+            Encoding outputEncoding = null,
+            bool stripAnsiEscapeCodes = true)
+        {
+            var result = await executable.ExecuteAsync(arguments, writeInput, outputEncoding, stripAnsiEscapeCodes);
+            return result.StandardOutput.SplitLines().Concat(result.StandardError.SplitLines());
         }
 
         /// <summary>
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2494,9 +2494,8 @@ namespace GitCommands
                     var localItem = item;
                     localItem.SetSubmoduleStatus(ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                     {
-                        await TaskScheduler.Default.SwitchTo(alwaysYield: true);
-
-                        var submoduleStatus = GitCommandHelpers.GetCurrentSubmoduleChanges(this, localItem.Name, localItem.OldName, localItem.Staged == StagedStatus.Index);
+                        var submoduleStatus = await GitCommandHelpers.GetCurrentSubmoduleChangesAsync(this, localItem.Name, localItem.OldName, localItem.Staged == StagedStatus.Index)
+                        .ConfigureAwait(false);
                         if (submoduleStatus != null && submoduleStatus.Commit != submoduleStatus.OldCommit)
                         {
                             var submodule = submoduleStatus.GetSubmodule(this);
@@ -2611,10 +2610,10 @@ namespace GitCommands
         }
 
         [CanBeNull]
-        public Patch GetCurrentChanges(string fileName, [CanBeNull] string oldFileName, bool staged, string extraDiffArguments, Encoding encoding = null, bool noLocks = false)
+        public async Task<Patch> GetCurrentChangesAsync(string fileName, [CanBeNull] string oldFileName, bool staged, string extraDiffArguments, Encoding encoding = null, bool noLocks = false)
         {
-            var output = _gitExecutable.GetOutput(GetCurrentChangesCmd(fileName, oldFileName, staged, extraDiffArguments, noLocks),
-                outputEncoding: LosslessEncoding);
+            var output = await _gitExecutable.GetOutputAsync(GetCurrentChangesCmd(fileName, oldFileName, staged, extraDiffArguments, noLocks),
+                outputEncoding: LosslessEncoding).ConfigureAwait(false);
 
             IReadOnlyList<Patch> patches = PatchProcessor.CreatePatchesFromString(output, new Lazy<Encoding>(() => encoding ?? FilesEncoding)).ToList();
 
@@ -2622,10 +2621,10 @@ namespace GitCommands
         }
 
         [CanBeNull]
-        private string GetFileContents(string path)
+        private async Task<string> GetFileContentsAsync(string path)
         {
             var args = new GitArgumentBuilder("show") { $"HEAD:{path.ToPosixPath().Quote()}" };
-            var result = _gitExecutable.Execute(args);
+            var result = await _gitExecutable.ExecuteAsync(args).ConfigureAwaitRunInline();
 
             return result.ExitCode == 0
                 ? result.StandardOutput
@@ -2633,11 +2632,11 @@ namespace GitCommands
         }
 
         [CanBeNull]
-        public string GetFileContents(GitItemStatus file)
+        public async Task<string> GetFileContentsAsync(GitItemStatus file)
         {
             var contents = new StringBuilder();
 
-            string currentContents = GetFileContents(file.Name);
+            string currentContents = await GetFileContentsAsync(file.Name).ConfigureAwaitRunInline();
             if (currentContents != null)
             {
                 contents.Append(currentContents);
@@ -2645,7 +2644,7 @@ namespace GitCommands
 
             if (file.OldName != null)
             {
-                string oldContents = GetFileContents(file.OldName);
+                string oldContents = await GetFileContentsAsync(file.OldName).ConfigureAwaitRunInline();
                 if (oldContents != null)
                 {
                     contents.Append(oldContents);
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -696,19 +696,19 @@ namespace GitCommands
             }
         }
 
-        public ConflictData GetConflict(string filename)
+        public async Task<ConflictData> GetConflictAsync(string filename)
         {
-            return GetConflicts(filename).SingleOrDefault();
+            return (await GetConflictsAsync(filename)).SingleOrDefault();
         }
 
-        public List<ConflictData> GetConflicts(string filename = "")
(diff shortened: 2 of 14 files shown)


GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -226,6 +225,27 @@ string ComposeOutput()
...
+        /// <returns>An enumerable sequence of lines that yields lines as they become available. Lines from standard output are returned first, followed by lines from standard error.</returns>
+        [MustUseReturnValue("If output lines are not required, use " + nameof(RunCommand) + " instead")]
+        public static async Task<IEnumerable<string>> GetOutputLinesAsync(
+            this IExecutable executable,
+            ArgumentString arguments = default,
+            Action<StreamWriter> writeInput = null,
+            Encoding outputEncoding = null,
+            bool stripAnsiEscapeCodes = true)
+        {
+            var result = await executable.ExecuteAsync(arguments, writeInput, outputEncoding, stripAnsiEscapeCodes);
```
jbialobr: It is not an async stream, but at least is non blocking.

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -170,12 +168,15 @@ private async Task GetSubmoduleStatusAsync(bool updateStatus, SubmoduleInfo info
                 return;
             }
 
+            // sequential prefetching FilesEncoding in a parent thread
+            var foo = supermodule.FilesEncoding;
```
jbialobr: After this PR it is at least responsive, but there is no GitModules caching. On every refersh GitModules are created for each submodule and have to load the same data (like git dir) again and again. I think we should observe submodules for modifications and refresh only these submodules which have changed. The same way as we use file system watchers. Otherwise if there is a lots of submodules the refresh action queries for changes for each submodule causing overall system overloading and cpu/hdd wasting.

gerhardol: File status reloading: #6359 (will not handle the case when all submodules are changed)
Diff loading: See #5769 Git-status already provides information about which submodules that are changed. Has been awaiting the other submodule changes, sidepanel etc

gerhardol:
Thanks!
Works OK for me. There is still a little lag selecting commits while loading, but no "hanging" seen. 2.51 is still snappier loading the data, but 2.51 starts show the revision graph later, so there are other cons.
Would like to see this with #6359 that fixed my real life problem (but not the extreme setup)

GitCommands/Submodules/SubmoduleStatusProvider.cs
jbialobr: Thanks for referencing the issues, my knowledge about the project is little outdated.

gerhardol: It is great to have you around, of course I have tried to do this myself too but it takes me quite some time to get into this.

jbialobr:
>For c62d4f9, I prefer #6359 instead
👍 

>There are quite some changes otherwise, mostly seem fine. I would like to see the outcome of the discussion in #6357 (comment) before merging.

Even if we go for a dedicated task scheduler for background tasks, running tens of background tasks simultaneously will degrade the overall system performance.

GitCommands/Git/GitModule.cs
```diff
@@ -2611,41 +2611,41 @@ public bool IsDirtyDir()
...
             return GetPatch(patches, fileName, oldFileName);
         }
 
         [CanBeNull]
-        private string GetFileContents(string path)
+        private async Task<string> GetFileContentsAsync(string path)
         {
             var args = new GitArgumentBuilder("show") { $"HEAD:{path.ToPosixPath().Quote()}" };
-            var result = _gitExecutable.Execute(args);
+            var result = await _gitExecutable.ExecuteAsync(args).ConfigureAwaitRunInline();
```
RussKie: Could you please explain how you chose when to use `ConfigureAwaitRunInline()` and when `ConfigureAwait(false)` calls?

I've found https://github.com/Microsoft/vs-threading/pull/292 (via https://github.com/dotnet/project-system/pull/3662), however it is not clear to me why we aren't using one or the other.

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -170,12 +168,15 @@ private async Task GetSubmoduleStatusAsync(bool updateStatus, SubmoduleInfo info
...
+
             info.Detailed = new AsyncLazy<DetailedSubmoduleInfo>(async () =>
             {
-                await TaskScheduler.Default;
                 cancelToken.ThrowIfCancellationRequested();
 
-                var submoduleStatus = GitCommandHelpers.GetCurrentSubmoduleChanges(supermodule, submoduleName, noLocks: true);
+                var submoduleStatus = await GitCommandHelpers.GetCurrentSubmoduleChangesAsync(supermodule, submoduleName, noLocks: true)
+                .ConfigureAwait(false);
```
RussKie: [nit] seems mis-indented, maybe worth keeping on the line above

GitCommands/Git/GitModule.cs
jbialobr: I use `ConfigureAwait(false)` when I don't want the continuation to be executed on the main thread.
I use `ConfigureAwaitRunInline` when I know it is safe to continue on the thread that completed the awaited task (regarding both: accessing variables and continuation's duration time).

RussKie:
Thank you

gerhardol:
How to continue this?
In everyday use #6359 is solving most issues for me. However, with just a few more submodules than what I use (or something else changed), the problem will occur.

This PR goes after the core problem even if there may be some issues left. From what I can tell, the changes are improvements anyway. Should this be merged anyway?
(At least first commit seem necessary.)

@jbialobr @RussKie 
How to proceed?

I rebased and removed one change in c62d4f940a7349aef375099df19a4ba472f2f061 (rest of the commit is probably OK):
https://github.com/gerhardol/gitextensions/tree/jbialobr/_jb/fix%236357
Builds here:
https://ci.appveyor.com/project/gerhardol/gitextensions/builds/23473927/artifacts

There are other ways to limit job (#6373 to minor extent, some suggestions in #5369) too, but that will not handle situations with many submodule changes.

RussKie:
I don't currently work with any repos that have submodules (other than GE), so I have no skin in the game.
The changes look sensible to me, hence my approval.

If you're happy with the changes, let's merge them.

gerhardol:
> If you're happy with the changes, let's merge them.

There are roughly three parts to this PR (partly from the PR description):

1. Load submodules info with async api. It was not enough, the lag was smaller but still significant.
First half, except 2nd commit

Good from what I know about async handling, but I am not the expert here, the existing GE code is my reference. If I were the maintainer, I would have squash merged these commits.

2. Pre-load module encoding. 
2nd commit

Not needed after #6359. At least one line should be removed, maybe the other are an improvement.

3. Load submodules info sequentially
Second half of the commits

Probably OK change, no need to make massive starts of background requests.
However, the change should not be needed according to the discussions in #6357, but if it is an improvement, it should be merged.

@jbialobr @sharwell @drewnoakes @spdr870 @RussKie 
How to proceed?

jbialobr:
>I would have squash merged these commits.

I see no value in squashing commits. It was already discussed.

>Pre-load module encoding.
2nd commit

It is no longer a part of this PR.

>However, the change should not be needed according to the discussions in #6357, but if it is an improvement, it should be merged.

In my opinion this change is not needed only if we implement refreshing those submodules which actually  have changed. Otherwise querying for all the submodules at once puts an unnecessary pressure on the system resources.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -263,9 +263,28 @@ private async Task<Nodes> LoadNodesAsync(SubmoduleInfoResult info, CancellationT
...
             private async Task LoadNodeDetailsAsync(CancellationToken token, Nodes loadedNodes)
             {
+                await TaskScheduler.Default;
+                foreach (var node in loadedNodes.DepthEnumerator<SubmoduleNode>())
+                {
+                    token.ThrowIfCancellationRequested();
+                    await node.LoadDetailsAsync(token).ConfigureAwaitRunInline();
+                }
+
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
sharwell:
```suggestion
                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
```

UnitTests/GitCommandsTests/GitModuleTest.cs
```diff
@@ -534,9 +538,12 @@ public void GetSuperprojectCurrentCheckout()
                 moduleTestHelperSuper.Module.GitExecutable.GetOutput(@"commit -am ""Update submodule ref""");
 
                 // Assert
-                (char code, ObjectId commitId) = moduleSub.GetSuperprojectCurrentCheckout();
-                Assert.AreEqual(32, code);
-                Assert.AreEqual(commitRef, commitId.ToString());
+                ThreadHelper.JoinableTaskFactory.Run(async () =>
```
sharwell: 💭 Why not just make the containing test method asynchronous?

GitCommands/Git/ExecutableExtensions.cs
sharwell: What kind of stream is it? What would it take to make it properly asynchronous?

jbialobr: I believe it is https://docs.microsoft.com/pl-pl/dotnet/api/system.collections.generic.iasyncenumerable-1?view=netcore-3.0

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
jbialobr: In the line below already is `token.ThrowIfCancellationRequested();` and it can not be removed. We can pass the token here as well to emphasize that it is not enough to pass it here and we have to recheck if cancellation was requested.

UnitTests/GitCommandsTests/GitModuleTest.cs
jbialobr: I don't know. Is doing so compatible with JTF?

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -335,11 +335,9 @@ private void Update()
                     {
                         try
                         {
-                            await TaskScheduler.Default;
-
                             var cmd = GitCommandHelpers.GetAllChangedFilesCmd(true, UntrackedFilesMode.Default,
                                 noLocks: true);
-                            var output = module.GitExecutable.GetOutput(cmd);
+                            var output = await module.GitExecutable.GetOutputAsync(cmd).ConfigureAwait(false);
```
gerhardol: This change causes git-status to run on the UI thread. Is this really necessary?

jbialobr: Do you observe any issues related to this? This is an async call and does not block the UI.

gerhardol: No problem seem, just confused by it.

PR #6373, opened 2019-03-17
WIP Optimize create copy of GitModule()

gerhardol:
Part of #6357 

Separating the second part of #6359 to not mixup discussions
https://github.com/gitextensions/gitextensions/pull/6359/files/07b797b3460c0ceef870f73cab169e9a7cafacdf#diff-a77f67d7f76305a8eeaae638ef09120a

## Proposed changes

Make a copy of gitModule instead of creating a new GitModule.
This is the pattern used in similar situations, for instance GitStatusMonitor.

This avoids a call to ```git rev-parse --common-dir``` when querying submodules structure/status.
This is not critical in itself, but it is a distraction for #6359.

Note: A previous version (in #6357) used an optimized constructor to create a new instance. This is maybe the pattern that should be used in all similar situations.

## Test methodology

* Open refresh a repo with changes in several submodule. For example GE repo with extra files in the submodules
* Verify that there is not a separate call to git-rev-parse for each of the modified repos

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -120,6 +120,18 @@ namespace GitCommands
         }
 
         /// <summary>
+        /// Create a thread-safe copy of this module, copying "expensive" properties
+        /// </summary>
+        /// <returns>Copy of the current module</returns>
+        public GitModule Copy()
+        {
+            GitModule module = new GitModule(WorkingDir);
+            module._gitCommonDirectory = _gitCommonDirectory;
+
+            return module;
+        }
+
+        /// <summary>
         /// Gets the directory which contains the git repository.
         /// </summary>
         [NotNull]
--- a/GitCommands/Submodules/SubmoduleInfoResult.cs
+++ b/GitCommands/Submodules/SubmoduleInfoResult.cs
@@ -10,7 +10,7 @@ namespace GitCommands.Submodules
     public class SubmoduleInfoResult
     {
         // Module that was used to gather submodule info
-        public IGitModule Module { get; internal set; }
+        public GitModule Module { get; internal set; }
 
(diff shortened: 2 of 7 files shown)


RussKie:
> Make a copy of gitModule instead of creating a new GitModule.
> This is the pattern used in similar situations, for instance GitStatusMonitor.

That's actually the only instance of such use I could find through the codebase.
We always construct a new `GitModule` instance.

gerhardol:
> That's actually the only instance of such use I could find through the codebase.

I found this too in a quick search:

  F:\dev\gc\gitextensions_4\GitUI\UserControls\RevisionGrid\RevisionGridControl.cs(749):                GitModule capturedModule = Module;

There are at least 20 occurrences in constructors, that is about the same though.

I find about 20 occurrences of new GitModule(path), none to create a thread local module.

For the three (at least) occurrences copying the variable as well as the possibly other occurrences, the creation could be optimized (similar to what I used in the original version):

```
        public GitModule(GitModule module) : this(module.WorkingDir)
        {
            _gitCommonDirectory = module._gitCommonDirectory;
        }
```

However, when I test again, I do not see the problem. Will modify the solution some.

gerhardol:
Pushed an update of the Copy API, as well as avoid creating GitModules for all changed submodules.
The majority of the improvements is from #6389, this primarily removes a distraction.
The second part could be separated (and the API improved).
Set to WIP for now, leave it open for the discussions for #6372.
If not done earlier,  this is relevant for remaining parts of #5769

gerhardol:
Rebased on #6445 (in WIP)
Just two changes remains.
This is not making much difference though (but should maybe be applied in other situations).

gerhardol:
Will not push this change and likely not other potential changes
(plan to close #5769 without this)

PR #6376, opened 2019-03-17
Update ExListView on expand / collapse of a group

### LINKED ISSUE #6375: ExListView not redrawn on expand / collapse
## Current behaviour

only the icons are redrawn when ExListView groups are expanded / collapsed, e.g. in the FileStatusList

## Expected behaviour

texts are redrawn, too

## Steps to reproduce

- select a merge commit
- collapse a group
grafik

## Screenshots

grafik

yes, in 3.01.00.4867 Build 843573e9f3f7a8bdf5d0efd3bcde05ce93b0aa67 and 3.0.0.4433
but *not* in 3.1.0.4985 (of 2019-01-23)

## Environment

- Git Extensions 3.1.0
- Build d35b0c8c0ed61444b54d96f897e35042c78fd1b3
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6375.

## Proposed changes

- workaround: invalidate ExListView retarded on changed `ColumnHeader` width

### Before

before

### After

after double click

## Test methodology 

- manual testing
  - [x] double click
  - [x] click the collapse / expand button

## Test environment(s) 

- Git Extensions 3.1.0
- Build c4ea6ea42ecd0e9d9b4f1988218145799f88557d
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -1021,9 +1021,21 @@ namespace GitUI
             if (!_updatingColumnWidth)
             {
                 _updatingColumnWidth = true;
-                columnHeader.Width = GetWidth();
+
+                int width = GetWidth();
+                if (columnHeader.Width != width)
+                {
+                    columnHeader.Width = width;
+
+                    // invalidate the FileStatusListView explicitely and asynchronously
+                    // Else texts will not be redrawn when scrollbars are shown or hidden.
+                    FileStatusListView.BeginInvoke(new Action(() => FileStatusListView.Invalidate()));
+                }
+
                 _updatingColumnWidth = false;
             }
+
+            return;
 
             int GetWidth()
             {


GitUI/UserControls/FileStatusList.cs
```diff
@@ -1021,10 +1021,19 @@ private void UpdateColumnWidth()
             if (!_updatingColumnWidth)
             {
                 _updatingColumnWidth = true;
-                columnHeader.Width = GetWidth();
+
+                int width = GetWidth();
+                if (columnHeader.Width != width)
+                {
+                    columnHeader.Width = width;
+                    FileStatusListView.ForceRepaint();
```
RussKie: Could we use something like (pseudocode)
```
FileStatusListView.BeginInvoke(() => FileStatusListView.Invalidate());
```
This should invalidate the control asynchronously.

mstv: Works indeed. Though the compiler wants a `new Action()` around the lambda.

RussKie:
it sounds about right, I don't remember the syntax by heart.
:+1:

vbjay:
This can be accomplished  with a bash script in the meantime. Get the repo
path and use clip to store in clipboard.

NikolayXHD:
The problem is caused by following line in `ExListView.cs`
```csharp
case NativeMethods.WM_REFLECT_NOTIFY when IsCustomDraw(m) && !_isInWmPaintMsg:
```

it suppresses `ListView`'s custom draw after collapsing / expanding group so we see the default look of list which is just icons from `ImageList`.

I removed the line (in current master branch commit 80df3ec608f12b919eacb8ea3ce58fb69f80768b and the bug is fixed.

NikolayXHD:
submitted alternative PR #6453

RussKie:
Replaced by #6453 

Thank you

PR #6383, opened 2019-03-18
FormAbout: fix "Git Extensions is open source. Get involved!" text wa…

DmitryZhelnin:
…s cropped when scaling is greater than 100%

### Before

about125_old

### After

about125
about150

----


RussKie:
Thank you

PR #6385, opened 2019-03-20
#6147 inverted diffs after refresh

### LINKED ISSUE #6147: inverted diffs after refresh
## Current behaviour

Under certain circumstances, the diffs shown in the Diff tab can be inverted just by refreshing the GUI.
I.e. when showing the differences from _<some_old_commit>_ to _<some_new_commit>_ and refreshing the GUI, the Diff tab shows the differences from _<some_new_commit>_ to _<some_old_commit>_ instead.

## Expected behaviour

Diffs shall not be shown as inverted, no matter whether the GUI is refreshed or not.

## Steps to reproduce

1. Load any repository in GitExtensions and click on the Diff tab.

2. Select any commit, except the latest one in your revision graph.
bugdiffrefresh1

3. Hold the CTRL key and select another commit, which is **after** the first one (i.e. more top-wards in the revision graph). It doesn't have to be on the same branch and it doesn't have to be a direct child commit. It may also be one of the artificial commits (if enabled).
It is important, however, that the second commit is shown above the first one in the revision graph.
bugdiffrefresh2

4. Press F5 or click on the Refresh icon.
bugdiffrefresh3

Yes, GitExtensions 2.51.05 is not inverting the diffs.

## Environment

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

ignacioerrico:
Fixes #6147 

## Proposed changes

- The `HashSet ToBeSelectedObjectIds` should be a `List`, because order matters.
- There is no need for `ToBeSelectedRowIndexes` to be a `ConcurrentQueue`, since we want to wait till we have all the rows to be selected and, when we have them, we select them. The list of indices can thus be stored in a `List` that includes order information.

### Before

The grid attempts to select a commit as soon as it is loaded. Consequently, rows are being selected in the order in which revisions are read, not in the order in which they had originally been selected.

### After

By waiting till all rows to be selected have been added to the grid before attempting to select them, we can select them in the correct order. With the current implementation, that's the best that can be done if we don't want to wait till all commits have been loaded before attempting to select the rows.

## Test methodology 

- Manual testing. I followed the same steps as indicated in the ticket, and verified that the behavior is as expected.
  - I considered special cases, like selection of many commits, and selection of commits at both ends of the commit history. (By the way, without these changes, a selection scattered throughout the commit history, would produce a new random selection when refreshing.)

## Test environment(s) 

- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -355,13 +355,6 @@ namespace GitUI.UserControls.RevisionGrid
             return _revisionGraph.GetNodeForRow(rowIndex)?.GitRevision;
         }
 
-        public void Prune()
-        {
-            _revisionGraph.Clear();
-
-            SetRowCount(_revisionGraph.Count);
-        }
-
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Reliability", "CA2002:DoNotLockOnObjectsWithWeakIdentity", Justification = "It looks like such lock was made intentionally but it is better to rewrite this")]
         private void SetRowCount(int count)
         {
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


mast-eu:
I made some functional tests and can confirm that it closes the issue. No undesired side effects. Seems to be fine.

I did not have a detailed look at the implementation though.

mast-eu:
I noticed only now that you based your work on branch **release/3.0.2**. I think you should rebase on **master** and also use **gitextensions:master** for the PR. For more details see the [contributor guidelines](https://github.com/gitextensions/gitextensions/wiki/Guidelines%3A-contributor).

ignacioerrico:
No problem to do that, but I'd actually read the [contributor guidelines](https://github.com/gitextensions/gitextensions/wiki/Guidelines%3A-contributor) and, according to item `2.iii`, my work should be based on `release/3.0.2`. Or did I misunderstand that?

> 2.iii. Locate the newest release branch (highest version number) which exhibits the bug, but with the version listed in the GitExtensions.releases file. This is the base branch for your work. [...]

If you can confirm, I'll rebase my work on the `master` branch when I have access to my computer later today. :-)

gerhardol:
> If you can confirm, I'll rebase my work on the master

Yes, this should not (in my opinon, but RussKie will agree) only to master, to be 3.1
I tried to clarify the wiki

RussKie:
Thank you for highlighting it, we need to tweak the guide.
Given then next release will be 3.1, the fix needs to go to the master branch.
Thank you

ignacioerrico:
Thanks for explaining. I recreated my branch, so I created a [new pull request](https://github.com/gitextensions/gitextensions/pull/6394) based on `master`.

PR #6390, opened 2019-03-21
Fix exception thrown GitModule.GetRemotes on an invalid git directory

### LINKED ISSUE #6093: System.Exception opening an existing repository

amaiorano:
Fixes #6093 

## Proposed changes

- GitModule.GetRemotes no longer throws an exception if the module opened isn't a valid Git module; rather, it returns an empty list.
- Opening and closing Settings dialog doesn't throw an exception if a valid module isn't open.

## Test methodology 

- Manual testing

## Test environment(s) 

- GIT version 2.19.1.windows.1
- Windows 10

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2183,6 +2183,12 @@ namespace GitCommands
                     {
                         var fetchLine = enumerator.Current;
 
+                        // An invalid module is not an error; we simply return an empty list of remotes
+                        if (fetchLine.Contains("not a git repository"))
+                        {
+                            return remotes;
+                        }
+
                         if (!enumerator.MoveNext())
                         {
                             throw new Exception("Remote URLs should appear in pairs.");
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1447,7 +1447,7 @@ namespace GitUI.CommandsDialogs
 
             _dashboard?.RefreshContent();
 
-            _gitStatusMonitor.Active = NeedsGitStatusMonitor();
+            _gitStatusMonitor.Active = NeedsGitStatusMonitor() && Module.IsValidGitWorkingDir();
         }
 
         private void TagToolStripMenuItemClick(object sender, EventArgs e)
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -310,7 +310,10 @@ namespace GitUI
             }
             finally
             {
-                RepoChangedNotifier.UnLock(changesRepo && actionDone);
+                // The action may not have required a valid working directory to run, but if there isn't one,
+                // we shouldn't send a "repo changed" notify.
+                bool requestNotify = actionDone && changesRepo && Module.IsValidGitWorkingDir();
+                RepoChangedNotifier.UnLock(requestNotify);
             }
 
             return actionDone;


amaiorano:
Looks like I broke some tests. Will take a look tomorrow.

amaiorano:
I added review comments to the diffs directly. They don't show up here on the Conversation page, though...

amaiorano:
@RussKie For clarity, there are 3 commits:

* Fix exception thrown from GitModule.GetRemotes on an invalid git directory
* Fix exception thrown when opening then closing Settings dialog on an invalid git repo
* Possibly better fix for GitModule.GetRemotes

The last commit provides an alternative to the first one. It might be easier to look at each commit separately, and let me know which one you prefer.

RussKie:
I would opt for the initial approach you took.
Whilst the approach in eaf14dc may look simpler, it does create a coupling between this method and a `GitModule` implementation.

I would like to go the opposite way and break down the monolith of `GitModule` into smaller (preferably SOLID) classes that we can reason about.

This method is also only used in the ROT, and it is causing a bit of noise (e.g. #6093). All other access to remotes is done via `IGitRemoteManager`.
Looking through the use, we only need the information to display tooltips and the icon. I'm pretty certain we can get all the same information from .gitconfig.

I think we need to sit down and do a bit of API clean/make up. 
E.g. `IGitRemoteManager` is only interacts with configured remotes in .gitconfig it should probably be renamed to `IUserConfiguredRemotes` etc.
I'll open a ticket about it.

Right now I'm learning more towards that we need to completely remove `IGtiModule.GetRemotes` method.

amaiorano:
> This method is also only used in the ROT, and it is causing a bit of noise (e.g. #6093)

Yes, this PR addresses that issue.

> I would like to go the opposite way and break down the monolith of GitModule into smaller (preferably SOLID) classes that we can reason about.

I agree with you, @RussKie. Do you think we should go with the first commit, get this in, and work towards getting rid of the remote functions on IGitModule afterwards?

gerhardol:
Seem OK to me.

RussKie:
Yes, lets get a minimal fix first and do a clean up as a follow up
Thank you

amaiorano:
Removed the last commit.

amaiorano:
Will merge in a few hours if no objections...

amaiorano:
> Will merge in a few hours if no objections...

Oh, sorry @RussKie, I thought you had already approved. I didn't mean to presume. Thanks for approving!

PR #6392, opened 2019-03-21
Fix vertical ruler position not correctly read from settings

### LINKED ISSUE #6391: Vertical ruler position not correctly read from settings
## Current behaviour
When opening a new diff window (like a file history), the vertical ruler position is not updated according to the settings - it has the default value.

## Expected behaviour
The ruler should have correct position right from the start.

## Steps to reproduce
1. Set a non-default value for the vertical ruler position (in Settings - Detailed - Diff Viewer).
2. Open the history of any file.
3. The ruler is at the default position.

The ruler only moves to the correct position after clicking the Settings icon in the diff view and clicking OK in the settings window (no editing needed), which suggests the settings are not loaded correctly the first time the window opens.

Did not check.

## Environment

- Git Extensions 3.1.0
- Build 7a5680d2fe317c08a4cb99f66a3f5e5e1c1c6c81
- Git 2.21.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

DmitryZhelnin:
Fixes #6391 

## Proposed changes
-  Apply settings right after command source has changed in `FileViewer`.
-  Remove default value of 80 that is set to the `FileViewerInternal.VRulerPosition` property in `FileViewer`'s `InitializeComponent` method. The `FileViewerInternal` gets actual value for vertical ruler from app settings in its constructor.

## Test methodology 

- Manual testing

## Test environment(s) 

- Git Extensions 3.1.0
- Build 7a5680d2fe317c08a4cb99f66a3f5e5e1c1c6c81 (Dirty)
- Git 2.20.0.windows.1
----

### DIFF
--- a/GitUI/Editor/FileViewer.Designer.cs
+++ b/GitUI/Editor/FileViewer.Designer.cs
@@ -379,7 +379,6 @@ namespace GitUI.Editor
             this.internalFileViewer.ShowTabs = false;
             this.internalFileViewer.Size = new System.Drawing.Size(757, 518);
             this.internalFileViewer.TabIndex = 1;
-            this.internalFileViewer.VRulerPosition = 80;
             // 
             // FileViewer
             // 
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -257,6 +257,7 @@ namespace GitUI.Editor
             if (commandSource?.UICommands != null)
             {
                 commandSource.UICommands.PostSettings += UICommands_PostSettings;
+                UICommands_PostSettings(commandSource.UICommands, null);
             }
 
             Encoding = null;
@@ -264,7 +265,11 @@ namespace GitUI.Editor
 
         private void UICommands_PostSettings(object sender, GitUIPostActionEventArgs e)
         {
-            internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await internalFileViewer.SwitchToMainThreadAsync();
+                internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            }).FileAndForget();
         }
 
         protected override void OnRuntimeLoad()


crypto-rsa:
Unfortunately I get a an `InvalidOperationException: “Cross-thread operation not valid: Control accessed from a thread other than the thread it was created on.”`  when opening a file history window. In `UICommands_PostSettings` the `VRulerPosition` is being set but I think this needs to run on the UI thread.

DmitryZhelnin:
> 
> 
> Unfortunately I get a an `InvalidOperationException: “Cross-thread operation not valid: Control accessed from a thread other than the thread it was created on.”` when opening a file history window. In `UICommands_PostSettings` the `VRulerPosition` is being set but I think this needs to run on the UI thread.

Thanks for your reply.
Can someone help with an example how to invoke an action on UI thread?
I've found this method
`public static ControlMainThreadAwaitable SwitchToMainThreadAsync(this Control control, CancellationToken cancellationToken = default)`
but not sure if it is the right solution.

RussKie:
You could do something like this:

```diff
diff --git a/GitUI/Editor/FileViewer.cs b/GitUI/Editor/FileViewer.cs
index 2e4cf78d7..99bb1489f 100644
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -255,7 +255,10 @@ private void OnUICommandsChanged(object sender, [CanBeNull] GitUICommandsChanged
 
         private void UICommands_PostSettings(object sender, GitUIPostActionEventArgs e)
         {
-            internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            internalFileViewer.BeginInvoke(new Action(() =>
+            {
+                internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            }));
         }
 
         protected override void OnRuntimeLoad()
```

The following also works, though I think it is more resource intensive:
```diff
diff --git a/GitUI/Editor/FileViewer.cs b/GitUI/Editor/FileViewer.cs
index 2e4cf78d7..38a80ed8f 100644
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -255,7 +255,11 @@ private void OnUICommandsChanged(object sender, [CanBeNull] GitUICommandsChanged
 
         private void UICommands_PostSettings(object sender, GitUIPostActionEventArgs e)
         {
-            internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await internalFileViewer.SwitchToMainThreadAsync();
+                internalFileViewer.VRulerPosition = AppSettings.DiffVerticalRulerPosition;
+            }).FileAndForget();
         }
 
         protected override void OnRuntimeLoad()
```

@jbialobr any opinions?

jbialobr:
>@jbialobr any opinions?

I would go for the second proposal because we are sure it is JTF compatible.

DmitryZhelnin:
Fixed and rebased

crypto-rsa:
I just tested the branch and it seems to work properly.

RussKie:
Thank you all

PR #5682, opened 2018-11-01
WIP 5678: Really removed horizontal scrollbar, instead of making it invisble

spdr870:
Fixes #5678

Changes proposed in this pull request:
- The horizontalscrollbar in the blame view was there and active. Only made invisible by a hack.
- Applied another hack, by making the committer panel wider, so no horizontal scrollbar is there
- The code is not really good. But this hack is better then the old one. This doesn't rely on the internals of the ICSharpCode control.
- Before, scrolling horizontal was possible. Either by using mouse buttons, or by selecting the text. This broke the control. Now scrolling horizontal is not possible. Scrolling is impossible because the control is really wide now. This is not visible, it is clipped by its parent.
 
Screenshots before and after (if PR changes UI):
- No UI changes, only a bug fix

What did I do to test the code and ensure quality:
- Manual testing
- Testing on 200% scaling

Has been tested on (remove any that don't apply):
- GIT 2.19
- Windows 10

PR #6393, opened 2019-03-21
Some improvments on blame form

### LINKED ISSUE #6389: Date formats used are inconsistent and confusing
## Current behaviour

Based on issue #4692, I gather that the date format used (in most places) is based on the Windows regional settings. That seems perfectly logical to me.

However, the `Blame` view does not comply with this and shows the date in the US date format, which is very confusing when you expect a different format.

## Expected behaviour

That dates in the `Blame` view respect the user preference based on the Windows regional settings.

## Steps to reproduce

1. Go to the Windows regional settings and set the date format to anything, but US format.
    Example: 
    GE-regional settings
2. Confirm that the date format is respected in the main view.
    GE-dates-in-main-view
3. Open the `Blame` view for a file and see the date format not being respected.
    GE-dates-in-blame-view

I never noticed this in GE 2.x, so I'm pretty sure it worked OK then.

## Environment

- OS version: Windows 7
- Git Extensions 3.0.2.5232
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 96dpi (no scaling)

Please let me know if you need more info.

pmiossec:
Fixes #6389

## Proposed changes

-  Fix date formatting in blame form (fix #6389)
-  Blame: Remove clutter by displaying file path only when not corresponding to the file blamed

Before:

image

After:

image

-  Better not costly approximation to have all the lines of the same size & lines of 120 characters is long enough

Before:

image

After:  (all the lines are of the same length and shorter)

image

-  Better display by inverting date and author

Before:

image

After: (all is better aligned and the dates are much more readable without compromising on author readability)

image

-  Align content of tooltip info with commit info content 

Before:

image

After: (labels updated to the same than the commit info and commit hash added)

image

Note: My only regrets here is that we can't use the translated strings here because the project `ResourceManager` that contains the translated strings has the `GitCommands` (where this code lies) as a dependency 😢 
I also wanted to display relative dates here which would have been of great added value but that's not possible for the same reason. 

After: (committer and commit date not displayed if the same than author and author date --same behavior than in commit info-- )

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 4aebb322d358882cdd5fcc174c9faa1583918fad
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

### DIFF
--- a/GitCommands/Git/GitBlame.cs
+++ b/GitCommands/Git/GitBlame.cs
@@ -69,9 +69,14 @@ namespace GitCommands
             var s = new StringBuilder();
 
             s.Append("Author: ").AppendLine(Author);
-            s.Append("AuthorTime: ").AppendLine(AuthorTime.ToString(CultureInfo.CurrentCulture));
-            s.Append("Committer: ").AppendLine(Committer);
-            s.Append("CommitterTime: ").AppendLine(CommitterTime.ToString(CultureInfo.CurrentCulture));
+            s.Append("Author date: ").AppendLine(AuthorTime.ToString(CultureInfo.CurrentCulture));
+            if (Author != Committer || AuthorTime != CommitterTime)
+            {
+                s.Append("Committer: ").AppendLine(Committer);
+                s.Append("Commit date: ").AppendLine(CommitterTime.ToString(CultureInfo.CurrentCulture));
+            }
+
+            s.Append("Commit hash: ").AppendLine(ObjectId.ToShortString());
             s.Append("Summary: ").AppendLine(Summary);
             s.AppendLine();
             s.Append("FileName: ").Append(FileName);
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -274,7 +274,7 @@ namespace GitUI.Blame
                 {
                     gutter.Append(line.Commit.Author);
                     gutter.Append(" - ");
-                    gutter.Append(line.Commit.AuthorTime.ToString(CultureInfo.CurrentUICulture));
+                    gutter.Append(line.Commit.AuthorTime.ToString(CultureInfo.CurrentCulture));
                     gutter.Append(" - ");
                     gutter.Append(line.Commit.FileName);
                     gutter.Append(' ', 100).AppendLine();
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -268,7 +268,7 @@ namespace GitUI.Blame
             {
                 if (line.Commit == lastCommit)
                 {
-                    gutter.Append(' ', 200).AppendLine();
+                    gutter.Append(' ', 120).AppendLine();
                 }
                 else
                 {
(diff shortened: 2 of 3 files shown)


pmiossec:
Perhaps @jrfnl , you could test and report if your problem is fixed?

jrfnl:
@pmiossec I'd be happy to if someone could get me a compiled version to install... Sorry, not a C# dev myself.

vbjay:
Look at the appveyor link at the bottom of the pull request in the checks.

pmiossec:
If the build pass, there will be a built version here:
https://ci.appveyor.com/project/spdr870/gitextensions-f0pvf/builds/23267449/artifacts

jrfnl:
@vbjay @pmiossec Excellent news! Will test in a minute.

jrfnl:
Tested and found working re #6389. Thanks for the quick response to the issue!

I'm not sure if I'm a fan of the date vs committer name reversal, but that may just be me - I more often look at `blame` to see **who** created the issue, than to see _when_ it was created. And for the _when_, the date is normally enough for me, I rarely need the precise time.

On another note: same as @gerhardol, while I tested this I noticed that there is no horizontal scrollbar in that pane. More than that: if you "click-hold-move-mouse-to-the-right" to move to the right in the pane to see the rest of the text, there is no way to ever move back.
"click-hold-move-mouse-to-the-left" doesn't seem to work.

This never was really problematic previously, but with the date/name reversal this now becomes a more prominent issue as the space taken up by the time means the name is often (partially) hidden, while that is the info I need to see.
(depends of course on screen size, but still)

RussKie:
> Note: My only regrets here is that we can't use the translated strings here because the project `ResourceManager` that contains the translated strings has the `GitCommands` (where this code lies) as a dependency 😢
> I also wanted to display relative dates here which would have been of great added value but that's not possible for the same reason.

Could you please raise a new issue, I think it is well worth consideration.
It may also be required for us to move with the plugin extraction and separation story.

pmiossec:
>On another note: same as @gerhardol, while I tested this I noticed that there is no horizontal scrollbar in that pane. 

I tried a little to fix it. From what I've seen, that's not possible to show only the horizontal scroll bar.
So we have to decide if we want to display the 2 scrollbars or none.

gerhardol:
> > On another note: same as @gerhardol, while I tested this I noticed that there is no horizontal scrollbar in that pane.
> 
> I tried a little to fix it. From what I've seen, that's not possible to show only the horizontal scroll bar.
> So we have to decide if we want to display the 2 scrollbars or none.

See also #5682, previous attempt

gerhardol:
a3ce9b6
I believe I want to see the time too. Maybe I get used to using the tooltip instead, not suggesting changes now

pmiossec:
I have removed the 2 changes where there is no consensus ( invert date and author and remove of time ).
I will try to find something satisfying for everyone (but in another PR)

I also have improved the contextual menu.

So now, it looks like that:
image

And before:

image

I hope we could merge it quickly so that I could build on it to try to improve further in another PR...

pmiossec:
>if you "click-hold-move-mouse-to-the-right" to move to the right in the pane to see the rest of the text, there is no way to ever move back.

I have found a way, by disabling the selection, to prevent that.  It needs to be correctly implemented but that could do another PR...

DmitryZhelnin:
> I believe I want to see the time too.
What about moving this to the application settings?

pmiossec:
@DmitryZhelnin That's the idea but that requires more time to do it properly. So, that will wait another PR.

gerhardol:
> I have removed the 2 changes where there is no consensus

I did not mean to block on my comment.
But for me, this can be merged now.

RussKie:
> > if you "click-hold-move-mouse-to-the-right" to move to the right in the pane to see the rest of the text, there is no way to ever move back.
> 
> I have found a way, by disabling the selection, to prevent that. It needs to be correctly implemented but that could do another PR...

I have a fix. I'll send it as a separate PR for you to test.

RussKie:
A nice set of changes! 👍 

May I ask you to separate changes and present them in separate PRs though? One PR for the dates and another for the context menu.

This way it is easier to reason about each and discuss the change.

Thank you.

pmiossec:
I have removed the changes in the contextual menu (and opened the PR #6400 to handle that).

RussKie:
Thank you

PR #6384, opened 2019-03-20
[#6147] Sign developer certificate of origin

ignacioerrico:
Signing the _developer certificate of origin_ as specified in [`CONTRIBUTING.md`](https://github.com/gitextensions/gitextensions/blob/master/CONTRIBUTING.md), since I'm making my first contribution to the project with code that attempts to fix #6147.

PR #6394, opened 2019-03-21
#6147 inverted diffs after refresh

### LINKED ISSUE #6147: inverted diffs after refresh

ignacioerrico:
Fixes #6147 

## Proposed changes

- The `HashSet ToBeSelectedObjectIds` should be a `List`, because order matters.
- There is no need for `ToBeSelectedRowIndexes` to be a `ConcurrentQueue`, since we want to wait till we have all the rows to be selected and, when we have them, we select them. The list of indices can thus be stored in a `List` that includes order information.

### Before

The grid attempts to select a commit as soon as it is loaded. Consequently, rows are being selected in the order in which revisions are read, not in the order in which they had originally been selected.

### After

By waiting till all rows to be selected have been added to the grid before attempting to select them, we can select them in the correct order. With the current implementation, that's the best that can be done if we don't want to wait till all commits have been loaded before attempting to select the rows.

## Test methodology 

- Manual testing. I followed the same steps as indicated in the ticket, and verified that the behavior is as expected.
  - I considered special cases, like selection of many commits, and selection of commits at both ends of the commit history. (By the way, without these changes, a selection scattered throughout the commit history, would produce a new random selection when refreshing.)

## Test environment(s) 

- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt), which I signed in [this pull request](https://github.com/gitextensions/gitextensions/pull/6384).

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -178,11 +177,21 @@ namespace GitUI.UserControls.RevisionGrid
 
         internal AuthorRevisionHighlighting AuthorHighlighting { get; set; }
 
-        // Contains the object Id's that will be selected as soon as they are loaded.
-        public HashSet<ObjectId> ToBeSelectedObjectIds { get; set; } = new HashSet<ObjectId>();
+        // Contains the object Id's that will be selected as soon as all of them have been loaded.
+        // The object Id's are in the order in which they were originally selected.
+        public IReadOnlyList<ObjectId> ToBeSelectedObjectIds { get; set; } = new List<ObjectId>();
 
-        // The ToBeSelectedObjectId's should be converted to indexes. This queue is then used to select the correct index. This is used cross-threads.
-        private ConcurrentQueue<int> ToBeSelectedRowIndexes { get; set; } = new ConcurrentQueue<int>();
+        // The ToBeSelectedObjectIds's should be converted to indexes.
+        // Since revisions are read in order, this list stores those indexes along with information on the order in which they were selected.
+        private List<RowIndexToBeSelectedWithOrderInfo> ToBeSelectedRowIndexes { get; set; } = new List<RowIndexToBeSelectedWithOrderInfo>();
+
+        private class RowIndexToBeSelectedWithOrderInfo
+        {
+            public int RowIndex { get; set; }
+
+            // Order in which the row should be selected.
+            public int SelectionOrder { get; set; }
+        }
 
         public bool HasSelection()
         {
@@ -290,9 +299,15 @@ namespace GitUI.UserControls.RevisionGrid
         {
             _revisionGraph.Add(revision, types);
 
-            if (ToBeSelectedObjectIds.Remove(revision.ObjectId))
+            if (ToBeSelectedObjectIds.Contains(revision.ObjectId))
             {
-                ToBeSelectedRowIndexes.Enqueue(_revisionGraph.Count - 1);
+                var rowIndexToBeSelectedWithOrderInfo = new RowIndexToBeSelectedWithOrderInfo
+                {
+                    RowIndex = _revisionGraph.Count - 1,
+                    SelectionOrder = ToBeSelectedObjectIds.IndexOf(o => o == revision.ObjectId)
+                };
+
+                ToBeSelectedRowIndexes.Add(rowIndexToBeSelectedWithOrderInfo);
             }
 
             UpdateVisibleRowRange();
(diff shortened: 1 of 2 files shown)


ignacioerrico:
It looks much better now! Thanks for explaining the correct workflow, @mast-eu!

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -178,11 +177,21 @@ internal Font NormalFont
 
         internal AuthorRevisionHighlighting AuthorHighlighting { get; set; }
 
-        // Contains the object Id's that will be selected as soon as they are loaded.
-        public HashSet<ObjectId> ToBeSelectedObjectIds { get; set; } = new HashSet<ObjectId>();
+        // Contains the object Id's that will be selected as soon as all of them have been loaded.
+        // The object Id's are in the order in which they were originally selected.
+        public List<ObjectId> ToBeSelectedObjectIds { get; set; } = new List<ObjectId>();
```
RussKie:
```suggestion
        public IReadOnlyList<ObjectId> ToBeSelectedObjectIds { get; set; } = new List<ObjectId>();
```

RussKie:
@spdr870 could you please review

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: Fails to compile with this change

gerhardol:
Looks OK and works for me.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -290,9 +299,15 @@ public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlag
...
             _revisionGraph.Add(revision, types);
 
-            if (ToBeSelectedObjectIds.Remove(revision.ObjectId))
+            if (ToBeSelectedObjectIds.Contains(revision.ObjectId))
             {
-                ToBeSelectedRowIndexes.Enqueue(_revisionGraph.Count - 1);
+                var rowIndexToBeSelectedWithOrderInfo = new RowIndexToBeSelectedWithOrderInfo
+                {
+                    RowIndex = _revisionGraph.Count - 1,
+                    SelectionOrder = ToBeSelectedObjectIds.IndexOf(revision.ObjectId)
```
illfated:
```csharp
UserControls\RevisionGrid\RevisionDataGridView.cs(307,68):
 error CS1503: Argument 2:
 cannot convert from 'GitUIPluginInterfaces.ObjectId' to 'System.Func<GitUIPluginInterfaces.ObjectId, bool>' 
```

gerhardol: See my review comment about the compile error introduced by the latest commit.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
ignacioerrico: That was my fault.

RussKie suggested turning `ToBeSelectedObjectIds` from a `List` to an `IReadOnlyList`. That makes sense, since `ToBeSelectedObjectIds` contains a fixed list of commits that are to be selected, so I accepted the change.

However, I overlooked the fact that `IndexOf(T)` is a method of `List<T>`. For an `IReadOnlyList`, the extension method `IndexOf` on `IEnumerable<T>` takes a lambda instead of an object, as the error message indicates.

ignacioerrico:
The commits in my branch have been squashed.

By the way, and I hope I don't sound pretentious by asking this, but should I add my name to the resource file `GitUI\Properties\Resources.resx`, which includes the list of contributors? How is that file updated?

illfated:
I imagine some kind of "magic" script would pick the coder names from the list of commits or the blame list from the current released source. Then again, I could be wrong and the list could be a manually written one.

vbjay:
No. Our release generation automatically adds authors.

gerhardol:
See Setup/Prepare-Release.ps1 for contributors update

RussKie:
Thank you.

@ignacioerrico you will see your name added in the next release.

ignacioerrico:
Thank you guys for explaining that. I'm really glad to be a contributor to this terrific project! :-)

PR #6399, opened 2019-03-24
Fix blame control click hold move mouse to the right

RussKie:
## Proposed changes

As noted in https://github.com/gitextensions/gitextensions/pull/6393#issuecomment-475447254 if you "click-hold-move-mouse-to-the-right" to move to the right in
the pane to see the rest of the text, there is no way to ever move back.

The fix handles horizontal scroll for the committer and blocks the hscroll.

## Test methodology 

- manual tests to ensure there is no hscroll in the committer control.

----

### DIFF
--- a/GitUI/Editor/FileViewer.Designer.cs
+++ b/GitUI/Editor/FileViewer.Designer.cs
@@ -372,7 +372,7 @@ namespace GitUI.Editor
             this.internalFileViewer.Location = new System.Drawing.Point(0, 40);
             this.internalFileViewer.Margin = new System.Windows.Forms.Padding(0);
             this.internalFileViewer.Name = "internalFileViewer";
-            this.internalFileViewer.ScrollPos = 0;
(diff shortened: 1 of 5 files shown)


pmiossec:
That fixes the bad behavior. Well done!

jrfnl:
Tested, not sure if this was the intended behaviour, but basically "click-hold-mouse button-move right" now has no effect at all any more.

This, quite effectively, solves the problem of not being able to _move back_.

As for not being able to see what's "hidden" from view in the left pane: as the screen divider can be moved, that information can be accessed in that manner.

So, I'm OK with this change.

RussKie:
> As for not being able to see what's "hidden" from view in the left pane: as the screen divider can be moved, that information can be accessed in that manner.

That is the intended behaviour, if a user requires to see more or less information for the committer - they can use the divider.

RussKie:
Thank you for testing.

PR #6401, opened 2019-03-24
Fix fileviewer ShowLineNumbers setting that was not honored

pmiossec:
and thus line numbers were displayed in blame even if explicitly set to be hidden

That was the case since 456ea65d7143e1555455c44d8bd385169f2feaf4

See https://github.com/gitextensions/gitextensions/commit/456ea65d7143e1555455c44d8bd385169f2feaf4#diff-555178f3fd75ec9d4f3aebcef1f48882R128

The strategy adopted is to not set the value by default so the setting is determined by the content (actual behavior)
but if explicitly set, the value is used.

Fixes regression (surely introduced in 456ea65d7143e1555455c44d8bd385169f2feaf4 /cc @RussKie ) on blame commiter fileviewer that show the line numbers even if explicitly set to not show them.

### Before

The line numbers are displayed even if set to be hidden ( https://github.com/gitextensions/gitextensions/blob/16aea8ca2ad8b3d4bf84f0f26c1c6f71ff0684d0/GitUI/UserControls/BlameControl.cs#L48 )

image

### After

The line numbers is hidden:
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 6a6522413da8dc6c740124fde3c614f7eb3ea38f
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

### DIFF
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -3,6 +3,7 @@ using System.Drawing;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
+using GitExtUtils.GitUI;
 using GitUI.Editor.Diff;
 using ICSharpCode.TextEditor;
 using ICSharpCode.TextEditor.Document;
@@ -104,11 +105,7 @@ namespace GitUI.Editor
             return TextEditor.Text;
         }
 
-        public bool ShowLineNumbers
-        {
-            get => TextEditor.ShowLineNumbers;
-            set => TextEditor.ShowLineNumbers = value;
-        }
+        public bool? ShowLineNumbers { get; set; }
 
         public void SetText(string text, Action openWithDifftool, bool isDiff = false)
         {
@@ -137,10 +134,27 @@ namespace GitUI.Editor
 
             // important to set after the text was changed
             // otherwise the may be rendering artifacts as noted in #5568
-            TextEditor.ShowLineNumbers = !isDiff;
+            TextEditor.ShowLineNumbers = ShowLineNumbers ?? !isDiff;
+            if (ShowLineNumbers.HasValue && !ShowLineNumbers.Value)
+            {
+                Padding = new Padding(DpiUtil.Scale(5), Padding.Top, Padding.Right, Padding.Bottom);
+            }
+
             TextEditor.Refresh();
 
             _currentViewPositionCache.Restore(isDiff);
+        }
+
+        protected override void OnPaintBackground(PaintEventArgs e)
+        {
+            if (ShowLineNumbers.HasValue && !ShowLineNumbers.Value)
+            {
+                e.Graphics.FillRectangle(SystemBrushes.Window, e.ClipRectangle);
+            }
+            else
+            {
+                base.OnPaintBackground(e);
+            }
         }
 
         public void SetHighlighting(string syntax)
--- a/GitUI/Editor/FileViewer.Designer.cs
+++ b/GitUI/Editor/FileViewer.Designer.cs
@@ -394,7 +394,6 @@ namespace GitUI.Editor
             this.internalFileViewer.Name = "internalFileViewer";
             this.internalFileViewer.VScrollPosition = 0;
             this.internalFileViewer.ShowEOLMarkers = false;
-            this.internalFileViewer.ShowLineNumbers = true;
             this.internalFileViewer.ShowSpaces = false;
             this.internalFileViewer.ShowTabs = false;
             this.internalFileViewer.Size = new System.Drawing.Size(757, 518);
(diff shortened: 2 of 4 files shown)


RussKie:
Thank you. I liked the numbers, but have no dramas to let them turned off.

However the lack of a gutter on the LHS is an issue, we need to have a margin to look good.

image
(left: old, right: new)

pmiossec:
>However the lack of a gutter on the LHS is an issue, we need to have a margin to look good.

&

>Agree with RussKie, margins should be adjusted slightly

I agree. I worked on it but that's a harder thing to fix than we could thought.
We can't add a margin due to the way the control is placed (fill to parent size).
We can add padding but that is display in grey and that look very horrible.

>I liked the numbers, but have no dramas to let them turned off.

I worked also on having the possibility to change some settings when displaying blame result. And so be able to show or hide the line numbers. That's why I don't want to do big bang changes just to fix this margin problem. I postpone the fix of this problem because it could be impacted by other things I'm working on also...

The 2 ways I found to fix it easily without too much effort at the moment are:

* creating and adding a new type of white gutter with a specific width just to play the role of a margin (but it don't play well with another feature I want to add to the blame results)

* just adding a space at the beginning of every lines when line numbers is hidden. It perfectly fix the problem for this case but can't be generalized everywhere line numbers are hidden (in the future because at the moment that's the only place where this is the case. So I find this solution acceptable).

So, to sum up. I didn't find a perfect solution and I'm working on some others features and will try to fix that later.

RussKie:
> We can add padding but that is display in grey and that look very horrible.

Can we change the color?

pmiossec:
>Can we change the color?

I didn't find a property for that but a search on internet gave me a solution.
So, it's ok now...

pmiossec:
>I would have asked for a wider padding than 3 if not that is what is normally used

I'm not against that but did you tried it and found it not enough? Because I carefully choose this value after some tests because I found it visually more compelling. Even if I'm not sure it would like well on all dpi configuration...

gerhardol:
> I'm not against that but did you tried it and found it not enough? Because I carefully choose this value after some tests because I found it visually more compelling. Even if I'm not sure it would like well on all dpi configuration...

3 dp is maybe too little for text to the border. I first did not see the margin. I use 100%.
But I am OK with the margin, this was more a sidenote for another reviewer.

pmiossec:
@gerhardol In fact, you were right and on my other computer without scaling, 3px is too small.
So I have updated to 5.

RussKie:
Are you scaling It?
E.g. make the margin 4px wide and then multiply if by the scale factor (use
DpiUtil class)

pmiossec:
>Are you scaling It?

Yes. That's why I don't really understand why I found 3 satisfying on my work computer (hdpi) and not on my personal one (no hdpi). That's perhaps because I displayed it only on one of the 2 screens I have that have not the same resolution ( even that seems strange....).

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -137,12 +134,30 @@ public void SetText(string text, Action openWithDifftool, bool isDiff = false)
...
+            {
+                Padding = new Padding(DpiUtil.Scale(5), Padding.Top, Padding.Right, Padding.Bottom);
+            }
+
             TextEditor.Refresh();
 
             _currentViewPositionCache.Restore(isDiff);
         }
 
+        private static readonly SolidBrush PaddingBrush = new SolidBrush(Color.White);
```
RussKie: ❗️ We can't do this. This will come and bite us when we tackle the theme.
The colour must be derived from the control.

pmiossec: It was hard to find but I get it!
What do you prefer `new SolidBrush(TextEditor.ActiveTextAreaControl.TextArea.Document.HighlightingStrategy.GetColorFor("Default").BackgroundColor)` ( https://github.com/gitextensions/ICSharpCode.TextEditor/blob/9f443b2435241e4a528558af1b02d0691c285efc/Project/Src/Gui/TextView.cs#L238 ) which returns `SystemColor.Window` or directly `SystemColor.Window`?

RussKie:
Let's use SystemColor.Window.
Thank you

pmiossec:
>Let's use SystemColor.Window.

Ok. Done.

RussKie:
I don't mean to be annoying, but Im fairly sure there is a SystemBrushes
object instead of creating a brush from a SystemColor.

Could you please check.


On Apr 1, 2019 10:00 PM, "Philippe Miossec" <notifications@github.com>
wrote:

Let's use SystemColor.Window.

Ok. Done.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/gitextensions/gitextensions/pull/6401#issuecomment-478533801>,
or mute the thread

pmiossec:
>I don't mean to be annoying, but Im fairly sure there is a SystemBrushes

You're right. I planned to use it but that went out of my head.
Thanks! Done.

RussKie:
Thank you

PR #6406, opened 2019-03-26
refactor: Rename IGitRemoteManager to IConfigFileRemoteSettingsManager

RussKie:
`IGitRemoteManager` only interacts with configured remotes in .gitconfig so it is renamed to `IConfigFileRemoteSettingsManager` to better reflect its nature.

----

### DIFF
--- a/GitCommands/ExternalLinks/ExternalLinkRevisionParser.cs
+++ b/GitCommands/ExternalLinks/ExternalLinkRevisionParser.cs
@@ -13,11 +13,11 @@ namespace GitCommands.ExternalLinks
 
     public sealed class ExternalLinkRevisionParser : IExternalLinkRevisionParser
     {
-        private readonly IGitRemoteManager _gitRemoteManager;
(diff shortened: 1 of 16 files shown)


amaiorano:
> `IGitRemoteManager` only interacts with configured remotes in .gitconfig so it is renamed to `IUserConfiguredRemotes` to better reflect its nature.

This looks alright, but I'm not sure about the "UserConfigured" part of the rename. When it comes to remotes, and pretty much anything in .git/config, it's all user-configured.

So we've got this `IConfigFileSettings` interface, which abstracts the .git/config file at the highest level. `IGitRemoteManager` uses a `IGitModule` and it's `LocalConfigFile` (of type `IConfigFileSettings`) to manipulate the remotes sections of the config file. So maybe we should rename IGitRemoteManager to `IConfigFileRemoteSettings`? Or `IRemotesConfigFileSettings`?

What do you think?

RussKie:
Good point, thank you. Let's go with `IConfigFileRemoteSettings`

RussKie:
Updated, please check

PR #6408, opened 2019-03-26
Fix issue #6296: Incorrect context menu items in commit diff tab

### LINKED ISSUE #6296: Incorrect context menu items in commit diff tab
## Current behaviour

Worktree files cannot be deleted using the "Delete file" context menu item in the commit diff tab.

## Expected behaviour

I should be able to delete the file.

## Steps to reproduce

- Choose a commit from the history
- Open the diff tab
- Right click on a file
- Select 'Delete file' from the context menu

Alternatively, hit the DELETE key.

## Screenshots

image

Not at least a few versions.

## Environment

- GitExtensions version: 2.51.02
- GIT version: 2.16.1.windows.4
- OS version: Windows 10

realkompot:
Fixes #6296

## Proposed changes

Small changes to RevisionDiffControl as suggested in #6296:
* "Delete file" context menu item is no longer shown for 'historical' commits, and only visible for 'artificial' i.e. WorkTree or Index, and it makes sense to leave the menu item in that case.
* "Edit file" is replaced with: "Edit working directory file", "Edit working directory file with...", "Open this revision (temp file)", "Open this revision with... (temp file)" similar to the items in commit file tree tab. 
* Menu items "Edit working directory file" and "Edit working directory file with..." only appear if a file actually exists.
* Menu items "Open this revision (temp file)", "Open this revision with... (temp file)" are hidden in case if commit is "artificial" because there's no revision.
* All new menu items are hidden in case of submodule change.

### Discussion points 
* Decided to keep "Delete file" because there's code that handles deletion for Index/WorkTree (deletes in case for WorkTree and unstages/deletes for Index).
* GitItemStatus.TreeGuid is always null in RevisionDiffControl, this is because of the way it grabs diff from git executable, couldn't find a good way to make it grab filenames along with hashes. That's why i had to resort to Module.GetFileBlobHash to get blob hash.
* "Open this revision" menu item will be present for deleted files too, let me know if I should hide it. "Show in File Tree" also doesn't check if a file is "Deleted" so I thought I don't know enough to make such check.
* All menu items are hidden for submodule changes.

## Screenshots 

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs
@@ -1,4 +1,4 @@
-﻿using System.Windows.Forms;
+using System.Windows.Forms;
 
 namespace GitUI.CommandsDialogs
 {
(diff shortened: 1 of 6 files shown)


gerhardol:
Please sign contributors.txt in a separate PR
https://github.com/gitextensions/gitextensions/wiki/Guidelines%3A-contributor#contributor-agreement

RussKie:
Rather than "delete file" should it be "reset file"?

realkompot:
You mean naming? I don't thing so, the ultimate goal for it is to delete a file, and if it appears to be staged it resets it first.

This is how it's implemented, I didn't change anything in that code:


I'll remove this menu item if you think it's useless

RussKie:
Sorry, I didn't check the original issue.
The description for the PR is a bit confusing and is missing the statement of work. 
Could you please add a paragraph what the PR is addressing to the commit message.

realkompot:
Updated PR comment, commit message contains actual issue number.

realkompot:
Yes, i thought about duplication. Provided that "Open this revision (temp file)", "Open this revision with... (temp file)" event handlers have slightly different implementation between RevisionDiff and RevissionFileTree (due to different data structures used there), i thought extracting 3 lines of code to a shared place would only make it more difficult to read.

PR #6417, opened 2019-03-29
Generate avatars with author initials

pmiossec:
Personal opinion: I like to use gravatar but I find that there fallback generated avatars are useless. We are not able to remember which avatar correspond to who.

This PR try to improve that by generating more useful avatars.

## Proposed changes

- Better label for the different choices
- Add 2 more options:
   * Always use generated avatars with author initials
   * Use generated avatars with author initials as fallback of gravatar

### Before
image

### After

image

### Before

image

### After: when avatar generated for all the users

image

### After: when avatar generated only as fallback

image

### Before

image

### After

image

## Test methodology 

- Unit tests
- Manual

## Test environment(s) 
- Git Extensions 3.1.0
- Build 277d25f085a36bc6dba282a4d70eaaf986cb8027 (Dirty)
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)


gerhardol:
The existing handling with first asking GitHub (if GH repo) then Gravatar with fallback is confusing.
A recent PR added a checkbox in the configuration (wrong location but hard to find otherwise.

The objection I have to this PR is that the handling is even more confusing with Initials in "No image service" but having a higher prio than GH and Gravatar.
Can we make this a prio list instead:
 1. Initials
 2. GitHub
 3. Gravatar with backup (can be None)
 4. Initials

If 1. Initials is set, the selections below are grayed out.
If Gravatar have a backup other than None, 4. Initials is grayed out.
Maybe add 0. None too.
The selection in CommitIfo would just link to Settings.

No comments on the code now.
Primarily new file InitialsAvatarGenerator.cs that need to be reviewed.

pmiossec:
@gerhardol Your comment, that I didn't understood at first, made me think and I realize that, due to a very poor ergonomic, and after years using GE and looking at the code to do this PR, I didn't understood the UI 😭 

And indeed, this PR increased the problem 😢 

In CommitInfo panel, "Default Image" is a real bad label and don't convey any meaning.
In the settings, "No image service" is really not clear. I don't know if someone ever understood it. It took me a lot of time after reading your comment to understand it... The tooltip is clearer by I never saw it before today! (And a link into it is not the best clever idea ever found).

There is a lot to improve! I will try to  

>A recent PR added a checkbox in the configuration (wrong location but hard to find otherwise.

I didn't found it. Could you give the PR id or give more info about that?

>The existing handling with first asking GitHub (if GH repo)

From what I saw, that's just to retrieve avatars of emails looking like `@users.noreply.github.com`, so that can't be a distinct choice. And GitHub use Gravatar.

So, for me that should be:

1. Initials (fully local)
2. Gravatar (and Github). With the sub options :
 * Fallback None ( Which perhaps make no sense anymore now that we have initials, but i won't take the choice to remove it myself)
 * Fallback MonsterId
 * Fallback ....
 * Fallback Initials

gerhardol:
> In the settings, "No image service" is really not clear.

It is a mess. (I spend most available time on admin finding what the intention was for functionality like this so do not code much myself.)
"Fallback" is a better word than "No image" service.

> There is a lot to improve! I will try to

Yes, maybe a separate PR.
The part having Initials as a backup to Gravatar is fitting the current (dis)order, maybe splitting the PR

> I didn't found it. Could you give the PR id or give more info about that?

#6287 

> From what I saw, that's just to retrieve avatars of emails looking like @users.noreply.github.com, so that can't be a distinct choice. And GitHub use Gravatar.

I still consider that a separate level, could be extended to other providers: BitBucket etc that has their own avatar service.

pmiossec:
I tried to solve the not enough good avatar UI.

##Settings
### After

Notice the checkbox that allow to use only generated author initials avatars whithout using gravatar.
The label has been updated too.

image

There is also an improvement with the help icon that display the tooltip and is also clickable to open to right place in the gravatar help page.

## Avatar control

##After

The option to use the "Generate avatar with author initials" is now on its own menu item.
The menu label has been updated too.

image

PS: Writing the message, I see that perhaps I should have used the label "Use generated avatar with author initials" (instead of "Generate avatar with author initials"). What do you think about it?

pmiossec:
PS2:  Perhaps I should have also used the label "Use generated avatar with author initials **(instead of gravatar)**" in the settings? What do you think about it, again?

vbjay:
Frankly, if we have a hierarchy, we need a hierarchical view and
manipulation.  Possible options and add  to a list and let the user move up
and down the order.

gerhardol:
> Perhaps I should have also used the label "Use generated avatar with author initials (instead of gravatar)" in the settings?

I would like to have an "always" in there as well. However, if the other choices are grayed out, it should be sufficient. Radiobox; Initials Gravatar, None (None is new), with _Gravatar fallback_ if Gravatar is selected?
Will think a little .

RussKie:
@pmiossec for the initiative and for the work 👍 It looks though the work is non trivial and has a high visibility impact.

I suggest we do an in-depth discussion of what we want to achieve in the target state and, once we all agree on it, we can start with the implementation. I'd hate to see you doing a lot of work which may not go anyway (unless you like experiments), get frustrated and give up.

I agree the current implementation is not exactly straight forward (but better than it used to be few years ago) and could do with an upgrade.

One more thing to consider, not everyone uses GitHub, people use GitLabs, BitBucket, Perforce and on-prem installations of those. There have been requests to support other image registries other than Gravatar, so it would be good that after the refactor we were in a position to add extra registries as well.

RussKie:
> I spend most available time on admin finding what the intention was for functionality like this so do not code much myself.

Welcome to the club 😞 
Though it is exciting that there is a constant stream of contributions.

gerhardol:
Suggestion: Add Initials as fallback now, maybe reword to _Gravatar fallback style_
The "Only use Gravatar" can be discussed later. The GUI is tricky here.

> One more thing to consider, not everyone uses GitHub, people use GitLabs, BitBucket, Perforce and on-prem installations of those. There have been requests to support other image registries other than Gravatar, so it would be good that after the refactor we were in a position to add extra registries as well.

One reason to why I want a second level between "Always Initials" and Gravatar

gerhardol:
The latest updates did not update the settings screenshot.
I believe this is better than what we have now and I believe the concept is OK. Settings UI need to be updated at further changes.

image

pmiossec:
>I believe this is better than what we have now and I believe the concept is OK.

That's also what I think but i understand that not everyone agree. So, I have cut this PR and done a first PR containing only the gravatar fallback, so without UI changes in #6499.
I hope this PR could be merged quicker and let us the time to discuss and implement the UI improvements...

>There have been requests to support other image registries other than Gravatar, so it would be good that after the refactor we were in a position to add extra registries as well.

In fact, as I technically did not add a remote avatar provider, I thought this refactoring would have to be done later, when a new one was added. It seems I was wrong...

gerhardol:
> So, I have cut this PR and done a first PR containing only the gravatar fallback, so without UI changes in #6499.

Good
Setting this as WIP as it needs to be rebased on #6499 when that is merged.

> In fact, as I technically did not add a remote avatar provider, I thought this refactoring would have to be done later, when a new one was added. It seems I was wrong...

Adding a new local provider to functionality that is confusing to maintainers is slowing down enough even if the outcome was an improvement.

pmiossec:
I have reworked the settings to make it multi providers frendly:

image

image

image

and also in the contextual menu of the commit info panel:

When gravatar provider selected:
image

image

When author initial provider selected (gravatar options are hidden)
image

gerhardol:
No comments from me on the implementation or how this is presented for the current functionality.
This is a great improvement to the current settings.
(I may even use initials when RussKie has changed his Avatar.)

However, there is still a problem to explain for the user for github and I would like to discuss future providers slightly too. This should probably go in a separate feature request eventually, maybe documentation too.

The current implementation checks with GitHub for email address ends with users.noreply.github.com instead of Gravatar. GitHub use IdentIcon by default, can be configured to use Gravatar but has no fallback service (not GE settings). So in this case, the fallback service is not used and "user80" image is seen

Example of commit with different addresses:
valid Gravatar address
invalid user at users.noreply.github.com
user not gravatar.com (not users.noreply.github.com)
image at GitHub

image

The GitHub setup is not reflected in the settings. We could add other services too, most git servers have their own setup. That should be based on the server rather than the email address (users.noreply.github.com is a convenience hack). So for github.com reposask github first, gitlab.com ask gitlab etc. GitLab works similar to GitHub with no auth and Gravatar fallback. However BitBucket Server requires authentication and has no fallback.

The UI for these extra services will be tricky. 
 * We probably want to keep the users.noreply.github.com check before any services, 
 * After that server specific services with added fallback (github etc)
 * Gravatar
 * fallback services - not sure if those can be combined with other services than Gravatar

GitUI/UserControls/AvatarControl.cs
```diff
@@ -24,14 +25,35 @@ public AvatarControl()
             InitializeComplete();
 
             clearImagecacheToolStripMenuItem.Click += delegate { ClearCache(); };
+            UpdateGravatarOptionDisplayState();
+
+            foreach (AvatarProvider type in Enum.GetValues(typeof(AvatarProvider)))
```
RussKie: Can't we use `EnumHelper.GetValues<T>` method?

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -180,5 +207,29 @@ private void downloadDictionary_LinkClicked(object sender, LinkLabelLinkClickedE
...
+        private void ManageGravatarOptionsDisplay()
+        {
+            var showAvatarOptions = (AvatarProvider)AvatarProvider.SelectedValue == GitCommands.AvatarProvider.Gravatar;
+            lblNoImageService.Visible = showAvatarOptions;
+            _NO_TRANSLATE_NoImageService.Visible = showAvatarOptions;
+            pictureAvatarHelp.Visible = showAvatarOptions;
+        }
+    }
+
+    public class ComboBoxItem<T>
```
RussKie: I'm pretty sure this can be turned into private
```suggestion
    private class ComboBoxItem<T>
```

RussKie:
> I may even use initials when RussKie has changed his Avatar.

😢 you don't like my new avatar...


@gerhardol  I like your analysis 👍

pmiossec:
Done.

I have also made some renaming from `DefaultImageType` to `GravatarFallbackAvatarType` because the naming was too unclear. I imagine it was taken from the [gravatar context](http://en.gravatar.com/site/implement/images#default-image) but once used in our GitExtension context, it makes no sense...

pmiossec:
Should I do something to move this PR forward?

gerhardol:
> Should I do something to move this PR forward?

@RussKie assigned me and I approved so unless he protests by tomorrow, I suggest you merge

pmiossec:
> I suggest you merge

I'm always reluctant to merge my own PRs, but if there are approved, I'm OK with that...

RussKie:
if you have approvals - you can merge it.
If someone wants to take a closer look, they will assign themselves as a
reviewer (I do that)

PR #6428, opened 2019-04-01
Updated links to git-scm.com

mast-eu:
## Proposed changes

- [git-scm.com](https://www.git-scm.com) recently changed the structure of the HTML anchors in the git reference manual. These are used to link directly to specific options of git commands.
GE has a few of those links in code comments, which still linked to the correct page (i.e. git command), but no longer to the correct section (i.e. option). For example:
  - old: https://git-scm.com/docs/git-rev-parse#git-rev-parse---git-common-dir
  - new: https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---git-common-dir

## Test methodology 

- N/A. This PR changes only some URLs contained in comments. Nothing more. The code is not altered in any way.

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -323,7 +323,7 @@ namespace GitCommands
         /// <summary>
         /// Asks git to resolve the given relativePath
         /// git special folders are located in different directories depending on the kind of repo: submodule, worktree, main
-        /// See https://git-scm.com/docs/git-rev-parse#git-rev-parse---git-pathltpathgt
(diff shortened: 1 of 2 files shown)


mast-eu:
> I would have expected more links

Honestly, me too. I probably wouldn't even have started if I had known in advance that there are only 3 impacted links.
But then I found myself with the work done, so I decided to submit it anyhow...

gerhardol:
There are links in the documentation too

mast-eu:
I searched "git-scm.com/docs"in the documentation and found only 2 relevant links on page `Settings`

https://git-extensions-documentation.readthedocs.io/en/latest/search.html?q=%22git-scm.com%2Fdocs%22

Am I missing something?

PR #6430, opened 2019-04-02
Add settings to configure blame display

pmiossec:
## Proposed changes

- Add settings to customize blame display (see screenshot for list of settings)
- Also improve a little more memory usage when building the commiter text

### Before

image

### After

image

Result with some settings enabled:

image

## Test methodology 

- Manual 

## Test environment(s) 

- Git Extensions 3.1.0
- Build 84b65c3f6542f6adb817c308906522c66c611009
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -296,7 +296,7 @@ namespace GitUI.Blame
                 }
                 else
                 {
-                    BuildAuthorLine(line);
+                    BuildAuthorLine(line, lineBuilder, dateTimeFormat, filename);
 
                     gutter.Append(lineBuilder);
                     gutter.Append(' ', lineLength - lineBuilder.Length).AppendLine();
@@ -328,38 +328,38 @@ namespace GitUI.Blame
             CommitInfo.SetRevisionWithChildren(revision, children);
 
             controlToMask?.UnMask();
+        }
 
-            void BuildAuthorLine(GitBlameLine line)
+        private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, string dateTimeFormat, string filename)
+        {
+            if (!AppSettings.BlameHideAuthor && AppSettings.BlameDisplayAuthorFirst)
             {
-                if (!AppSettings.BlameHideAuthor && AppSettings.BlameDisplayAuthorFirst)
-                {
-                    lineBuilder.Append(line.Commit.Author);
-                    if (!AppSettings.BlameHideAuthorDate)
-                    {
-                        lineBuilder.Append(" - ");
-                    }
-                }
-
+                lineBuilder.Append(line.Commit.Author);
                 if (!AppSettings.BlameHideAuthorDate)
                 {
-                    lineBuilder.Append(line.Commit.AuthorTime.ToString(dateTimeFormat));
+                    lineBuilder.Append(" - ");
                 }
+            }
 
-                if (!AppSettings.BlameHideAuthor && !AppSettings.BlameDisplayAuthorFirst)
-                {
-                    if (!AppSettings.BlameHideAuthorDate)
-                    {
-                        lineBuilder.Append(" - ");
-                    }
+            if (!AppSettings.BlameHideAuthorDate)
+            {
+                lineBuilder.Append(line.Commit.AuthorTime.ToString(dateTimeFormat));
+            }
 
-                    lineBuilder.Append(line.Commit.Author);
-                }
-
-                if (filename != line.Commit.FileName)
+            if (!AppSettings.BlameHideAuthor && !AppSettings.BlameDisplayAuthorFirst)
+            {
+                if (!AppSettings.BlameHideAuthorDate)
                 {
                     lineBuilder.Append(" - ");
-                    lineBuilder.Append(line.Commit.FileName);
                 }
+
+                lineBuilder.Append(line.Commit.Author);
+            }
+
+            if (filename != line.Commit.FileName)
+            {
+                lineBuilder.Append(" - ");
+                lineBuilder.Append(line.Commit.FileName);
             }
         }
 
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Windows.Forms;
 using GitCommands;
 using GitExtUtils;
+using GitUI.BranchTreePanel;
 using GitUI.CommitInfo;
 using GitUI.Editor;
 using GitUI.HelperDialogs;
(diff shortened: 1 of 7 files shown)


gerhardol:
Is it reasonable to address the CodeFactor warning?

pmiossec:
Yes. When I will be in front of a computer 😊

RussKie:
❓ What is displayed when all new settings are off?

🤔 Just looking at the screenshot, we need to do few tweaks.

 * All "Hide xxx" must be turned into "Show xxx" to make it consistent with other menus, and make them sound positive (presence) as oppose to negative (absence).
* All appsettings need to be tweaked accordingly.

I'd like to see tests for `ProcessBlame` and `BuildAuthorLine`

Thank you

pmiossec:
>❓ What is displayed when all new settings are off?

The UI doesn't allow that. When you uncheck date, author is check (and inverse is true also).

>🤔 Just looking at the screenshot, we need to do few tweaks.

done.

>I'd like to see tests for ProcessBlame and BuildAuthorLine

done.

pmiossec:
I don't know if there is a problem with GitHub or AppVeyor but the build is no more triggered 😢

pmiossec:
@RussKie We will retry in some hours 😉

RussKie:
I'm guessing Appveyor is having troubles
Here's one of the ongoing reports - https://help.appveyor.com/discussions/problems/23201-request-to-github-web-hook-has-been-made-from-prohibited-ip-address

GitUI/UserControls/BlameControl.cs
```diff
@@ -278,19 +278,19 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
                 }
                 else
                 {
-                    gutter.Append(line.Commit.AuthorTime.ToString(CultureInfo.CurrentCulture));
+                    gutter.Append(line.Commit.AuthorTime.ToString(CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern));
```
RussKie: ⚠️ This needs to be discussed. We do show dates and times everywhere else.
For now please remove it from this PR.

Besides we have a setting to show relative dates (_Settings > Appearance > General_), perhaps we should honour this settings here.

GitUI/UserControls/BlameControl.cs
```diff
@@ -265,6 +270,10 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
 
             GitBlameCommit lastCommit = null;
 
+            var dateTimeFormat = AppSettings.BlameHideAuthorTime
+                ? CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern
+                : CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern + " " + CultureInfo.CurrentCulture.DateTimeFormat.ShortTimePattern;
```
RussKie: We should also honour user's date presentation settings (_Settings > Appearance > General_)
image

GitUI/UserControls/BlameControl.cs
```diff
@@ -279,49 +279,50 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
             // If it could be done with a solid rectangle around the text,
             // the extra spaces added here could be omitted.
 
+            var lineLength = Math.Max(80, 25 + _blame.Lines.Max(l => l.Commit.Author?.Length ?? 0) + _blame.Lines.Where(l => filename != l.Commit.FileName).Max(l => l.Commit.FileName.Length));
```
RussKie: Could you please wrap the line in a way that it becomes readable.
Right now it is near impossible to see what's going on.

GitUI/UserControls/BlameControl.cs
```diff
@@ -323,38 +323,38 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
             CommitInfo.SetRevisionWithChildren(revision, children);
 
             controlToMask?.UnMask();
+        }
 
-            void BuildAuthorLine(GitBlameLine line)
+        private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, string dateTimeFormat, string filename)
```
RussKie: To make this more testable (e.g. less dependent on the shared state - `AppSettings`) I recommend passing in all `AppSettings.*` flags.
This way it is easy to add test for all sorts of permutations.

GitUI/UserControls/BlameControl.cs
```diff
@@ -64,6 +66,37 @@ public BlameControl()
...
+            public (string gutter, string body) BuildBlameContents(string filename)
+            {
+                return _control.BuildBlameContents(filename);
+            }
+        }
+
+        public void UpdateShowLineNumbers()
+        {
+            BlameAuthor.ShowLineNumbers = AppSettings.BlameShowLineNumbers;
+        }
```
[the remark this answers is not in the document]
pmiossec: Because it's used from `FormFileHistory` also.

GitUI/UserControls/BlameControl.cs
pmiossec: I'm not against it's a new feature (as we didn't do that at the moment and this PR was just about adding some configuration on existing feature). Even if I'm pleased to do it, I will let it for another PR ;)

And, also, I see it as a distinct blame settings (why not initially set with this settings value) because we could like to display as relative in the application and the opposite in the blame (or the other way). Does it make sense?

GitUI/UserControls/BlameControl.cs
pmiossec: I will not be very pleased if I have to remove it even if this not consistent with what is done everywhere else in the application because I think that miss the point of why this setting is added here. 

The goal of this PR was to give more flexibility to gain precious horizontal space (in case we need to display long lines on small screens). So the constraints are a little different that elsewhere in the application and I think it add value to behave a little differently.

I did it on purpose once I discovered that most of the times, displaying time in the blame is useless where date is enough.

The other reason is that I'm not very motivated to take some time to remove part of the code already written that add value 😢 It's just a setting that most will ignore if not interested...
If you **really** want to remove it, I will do it to make the other settings merged but it's not really what I want to end of doing...

pmiossec:
Done most except 2 things!
Let's talk!!! 😄

GitUI/UserControls/BlameControl.cs
RussKie: Yes, no worries.
Just flagging that we have an inconsistent UX.

GitUI/UserControls/BlameControl.cs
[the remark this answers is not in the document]
RussKie: Thank you, I didn't see the usage.

RussKie:
Looks like the master is failing after this PR got merged
Could you please have a look https://ci.appveyor.com/project/gitextensions/gitextensions/builds/24310002 ?

pmiossec:
I saw that even if I didn't had a careful look at it.
I don't think it is caused by this PR. It seems not linked. 
I made 2 PRs based on master that built successfully.

I don't know if it's possible to trigger another build before investigating...

vbjay:
Close and reopen it will launch another build.

RussKie:
Kicked off a new build, let's see

pmiossec:
@vbjay we can't close and open master 😉

vbjay:
Lol. Hmm. Woops. Well appveyor should allow manual build starts from users
that have access.

pmiossec:
Yes. I thought I didn't have access but after sign in, I could have started a rebuild.

Good to know for the future.

Anyway, the build succeeded. 👍
Time to sleep 😁

PR #6444, opened 2019-04-06
Detailed submodule status in dropdown menu was not updated

gerhardol:
Regression from #6372

## Proposed changes
The submenu status in the dropdown menu was not updated after rework in #6372 
As a not released regression, this is not added to the 3.1.0 milestone

### Before

image

### After

image

## Test methodology 
Manual test: Make a test to a submodule

## Test environment(s) 
----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2720,12 +2720,13 @@ namespace GitUI.CommandsDialogs
             // then refresh all items at once with a single switch to the main thread
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
-                var loadDetalils = newItems.Select(e => e.loadDetails).Where(e => e != null);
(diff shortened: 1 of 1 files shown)


RussKie:
🤔 A seditious thought, having the left panel can we remove the toolbar item completely?

pmiossec:
> 🤔 A seditious thought, having the left panel can we remove the toolbar
> item completely?
>

I don't use submodules but I think we can't do it (at least at the moment)
because that obliged people to use the left panel which could be
problematic in terms of screen space and also maybe in terms of performance
(we have seen that the left panel slow down repository refresh in some
cases and I don't know if it was completely solved).

>

gerhardol:
> A seditious thought, having the left panel can we remove the toolbar item completely?

Not in 3.1, I expect that we should transition users first.
It is not much maintenance for the menu

RussKie:
👍

gerhardol:
Plan to merge this trivial regression tonight

PR #5777, opened 2018-11-14
only update submodule status at changes

gerhardol:
This is a follow up to #5747 after discussions with @amaiorano 
This is part 1 of #5769 - should maybe be considered for 3.0

This is based on #5747 minus last commit, submitted for @amaiorano to give feedback

Changes proposed in this pull request:

SubmoduleStatus: Update only when commit-count reports changes to submodules

Update submodule status at changes directly (async)
when changes detected, do not wait for dropdown or form reactivation
Submodule status setting is now dependent on the status updates

There are two special cases:
 * If submodule status changes to None, the status must be cleared
 * Changes are only triggered on submodule changes, supermodule changes are not detected, If a module has supermodules, the status is always calculated when opening the module (but not triggered if there are changes in the supermodules.
 
Screenshots before and after (if PR changes UI):
- n/a

What did I do to test the code and ensure quality:
- Manual tests

Has been tested on (remove any that don't apply):
- Windows 7/10

PR #6445, opened 2019-04-06
Separate submodule structure and status

gerhardol:
Part of #5769
(closes the issue with #6373 that may be dropped)

## Proposed changes
Submodule structure is separated from detailed status
Previously, the structure was always recalculated when status was updated
Also, the status is only calculated for the submodule trees where
git-status has detected changes.

This has simplified the code as well decreased the time before the
detailed status is displayed. The status was also flashing for several
seconds while updating (occurred regularly if any submodule had a change).

The normal use case with submodules is that most changes are done in the top project and that there are no or few changes in submodules.
#5777 (3.1.0) changed to only get the detailed status if there were any changes in submodules so the no-change scenario is handled already in master.
However, a change to any submodule would still require the status for all submodules.
Before #6372 this handling could hang the UI.
In addition to quicker update, the cpu load decreases with this PR. 

For a tree like below, previously was 5+8x6 git-diff commands required, with the update 1+2x6 (this is still an example where there are many dirty submodule trees).
This could of course be further improved by git-status on "dirty" levels, but the cost will likely be larger than the benefit (and complexity be increased).

image

Open:
 * Maybe update the test for submodule status handling (no such test existed before)
 * ~~AsyncLazy is not really needed, the status is serialized anyway~~

## Screenshots 
No change to the UI

## Test methodology
Test by making changes to submodules on several levels.
See #6357 for hints how to set this up
Check the Git Command log for expected commands

## Test environment(s)
See #6357 

----

### DIFF
--- a/GitCommands/Submodules/SubmoduleStatusProvider.cs
+++ b/GitCommands/Submodules/SubmoduleStatusProvider.cs
@@ -18,16 +18,30 @@ namespace GitCommands.Submodules
         event EventHandler StatusUpdating;
         event EventHandler<SubmoduleStatusEventArgs> StatusUpdated;
         void Init();
-        bool HasChangedToNone([CanBeNull] IReadOnlyList<GitItemStatus> allChangedFiles);
-        bool HasStatusChanges([CanBeNull] IReadOnlyList<GitItemStatus> allChangedFiles);
-        void UpdateSubmodulesStatus(bool updateStatus, string workingDirectory, string noBranchText);
+
+        /// <summary>
+        /// Update the submodule structure; find superprojects and submodules
+        /// </summary>
+        /// <param name="workingDirectory">Current module working directory</param>
+        /// <param name="noBranchText">The text where no branch is checked out for the submodule</param>
+        /// <param name="updateStatus">Update the detailed submodule status (set when current module is not top project)</param>
+        void UpdateSubmodulesStructure(string workingDirectory, string noBranchText, bool updateStatus);
+
+        /// <summary>
+        /// Update the submodule status
+        /// </summary>
+        /// <param name="workingDirectory">Current module working directory</param>
+        /// <param name="gitStatus">The Git status for the changes (also other than submodules)</param>
+        void UpdateSubmodulesStatus(string workingDirectory, [CanBeNull] IReadOnlyList<GitItemStatus> gitStatus);
     }
 
     public sealed class SubmoduleStatusProvider : ISubmoduleStatusProvider
     {
         private readonly CancellationTokenSequence _submodulesStatusSequence = new CancellationTokenSequence();
         private DateTime _previousSubmoduleUpdateTime;
-        private bool _previousSubmoduleHadChanges;
+        private SubmoduleInfoResult _submoduleInfoResult;
+        private readonly Dictionary<string, SubmoduleInfo> _submoduleInfos = new Dictionary<string, SubmoduleInfo>();
+        private IReadOnlyList<GitItemStatus> _gitStatusWhileUpdatingStructure;
 
         // Singleton accessor
         public static SubmoduleStatusProvider Default { get; } = new SubmoduleStatusProvider();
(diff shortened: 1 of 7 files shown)


GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -46,76 +60,98 @@ public void Dispose()
...
-
-            // If any submodules are changed, trigger an update
-            // TBD: This should check the status against last updated list and only update the required modules
-            // (maybe even ignore count)
-            return allChangedFiles.Any(i => i.IsSubmodule && (i.IsChanged || !i.IsTracked));
-        }
-
-        public void UpdateSubmodulesStatus(bool updateStatus, string workingDirectory, string noBranchText)
+        /// <inheritdoc />
+        public void UpdateSubmodulesStructure(GitModule gitModule, string noBranchText, bool updateStatus)
```
RussKie: ⚠️ `GitModule` should not be passed as an input parameter for a number of reasons.
1. It is a combination of state and behaviours, behavioural components should not be passed around.
2. Passing a concrete implementation creates a strong coupling, this reduces cohesion and negatively affects testability
3. Passing an object with a massive public surface in order to get a single property is also a bad practice.

Please rework the API in a way that does not require passing neither `IGitModule` nor `GitModule`.

GitCommands/Submodules/SubmoduleInfoResult.cs
```diff
@@ -10,7 +10,7 @@ namespace GitCommands.Submodules
     public class SubmoduleInfoResult
     {
         // Module that was used to gather submodule info
-        public IGitModule Module { get; internal set; }
+        public GitModule Module { get; internal set; }
```
RussKie: Why?

gerhardol: GetCurrentSubmoduleLocalPath() and SuperprojectModule() is not in IGitModule

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: This will require some calls to Git for data that already exists.
OK to use an optimized copy from #6373?

gerhardol: API changed (without  #6373 for now)

GitCommands/Submodules/SubmoduleInfoResult.cs
gerhardol: Passing currentModule around in addition to result (similar to how it was before)

RussKie: Please add them to the interface

gerhardol: GetCurrentSubmoduleLocalPath() can be added, but SuperprojectModule() is not trivial (you have discussed this with @amaiorano related to the sidepanel) as the return type need to be changed which will cause an avalanche of other additions to IGitModule. So that should be a separate refactoring.

gerhardol: A proposal in https://github.com/gerhardol/gitextensions/tree/feature/i6359-submodule-selective-update-interfaces
However, requires the use of casts...

RussKie:
I'll pull it down and have a look in few days.

gerhardol:
Added a test for status handling. (No such test existed before, I had removed a test for some other functionality that no longer were needed.)
To do this, I removed AsyncLazy which no longer really was stricly needed after structure/status changes to be able to make some type of test.

gerhardol:
@jbialobr 
Do you have a possibility to review, especially the changes to Async?

UnitTests/GitCommandsTests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -98,49 +98,31 @@ public void UpdateSubmoduleStatus_valid_result_for_second_nested_submodule()
...
+                result.OurSubmodules.All(i => i.Detailed == null).Should().BeTrue();
 
                 // Make a change in repo2
                 _repo1.CreateFile(_repo2Module.WorkingDir, "test.txt", "test");
-                changedFiles = GetStatusChangedFiles(_repo1Module);
+                changedFiles = GetStatusChangedFiles(currentModule);
                 changedFiles.Should().HaveCount(1);
-                _provider.HasChangedToNone(changedFiles).Should().BeFalse();
+                SubmoduleTestHelpers.UpdateSubmoduleStatusAndWaitForResult(_provider, currentModule, changedFiles);
+                ////result.OurSubmodules[0].Detailed.IsDirty.Should().BeTrue();
```
gerhardol: AsyncTestHelper.WaitForPendingOperations(); is not awaiting the status here, so the test fails (unless the debugger is connected).
I am guessing that is the test helper that is the problem.
Note: This was not tested before, not a new problem.

RussKie: Your message looks cryptic...
It sounds like you're having a problem. Could you please explain further?

gerhardol: > It sounds like you're having a problem.

The test is not working. (The reason this is still WIP)

The testhelper is not awaiting the result.
See below for a commit (in  https://github.com/gerhardol/gitextensions/pull/new/feature/i6359-submodule-selective-update-1) where the test works, but definitely worse API.

RussKie: I had it enabled and it failed.
Then I force rebuilt and it worked...

gerhardol: > I had it enabled and it failed.

Was this with this PR or the alternative branch?
I have not seen a solution for the testhelper or how to rewrite the tests

RussKie:
This branch.

gerhardol:
Added another test, also partly failing to document the expected behavior.
Removed WIP, the test (maybe after changes to the interface) can be in a follow up.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -623,7 +616,7 @@ public void GoToRef(string refName, bool showNoRevisionMsg)
             }
         }
 
-        #endregion
+#endregion
```
RussKie: Please undo these changes

UnitTests/GitCommandsTests/Submodules/SubmoduleStatusProviderTests.cs
RussKie: Maybe @jbialobr or @sharwell can help...

gerhardol:
Alternative tests in d3c699f using TestAccessor

gerhardol:
I want to squash...

RussKie:
Commendable desire! Don't deny yourself 😁

gerhardol:
Rebased on top of new PR #6488 (to be able to enable one failed test)
Included the change to remove AsyncLazy, no longer separate.

PR #6448, opened 2019-04-07
Mention regex in label of FileStatusList's filter

### LINKED ISSUE #6447: FileStatusList filter: mention Regular Expression
## Feature description

Mention that a Regular Expression (not a shell pattern) must be given to filter the files in the FileStatusList.

before / after:
before   after

## Environment

- Git Extensions 3.1.0
- Build 28fcbf1e864ecc2238ef9001586c537957c6c73b
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6447

## Proposed changes

- extend the label text

### Before

before

### After

after

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 385cf8b60c17f3ab2fbac3f10aa2858a59a51833
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.Designer.cs
+++ b/GitUI/UserControls/FileStatusList.Designer.cs
@@ -111,7 +111,7 @@
             this.FilterWatermarkLabel.Name = "FilterWatermarkLabel";
             this.FilterWatermarkLabel.Size = new System.Drawing.Size(65, 13);
             this.FilterWatermarkLabel.TabIndex = 0;
-            this.FilterWatermarkLabel.Text = "Filter files...";
+            this.FilterWatermarkLabel.Text = "Filter files using a regular expression...";
             this.FilterWatermarkLabel.Click += new System.EventHandler(this.FilterWatermarkLabel_Click);
             // 
             // FilterToolTip
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -1150,7 +1150,7 @@ Please make sure git (Git for Windows or cygwin) is installed or set the correct
         <target />
       </trans-unit>
       <trans-unit id="FilterWatermarkLabel.Text">
-        <source>Filter files...</source>
+        <source>Filter files using a regular expression...</source>
         <target />
       </trans-unit>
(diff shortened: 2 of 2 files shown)


GitUI/UserControls/FileStatusList.Designer.cs
```diff
@@ -111,7 +111,7 @@ private void InitializeComponent()
             this.FilterWatermarkLabel.Name = "FilterWatermarkLabel";
             this.FilterWatermarkLabel.Size = new System.Drawing.Size(65, 13);
             this.FilterWatermarkLabel.TabIndex = 0;
-            this.FilterWatermarkLabel.Text = "Filter files...";
+            this.FilterWatermarkLabel.Text = "Filter files using a Regular Expression...";
```
gerhardol: I would prefer regular expression here, GE (mostly) uses lowercase

mstv: Changed. (First I thought to use upper case for the proper name "Regular Expression".)

mstv:
> Can we not filter using a plain text?

Of course, we could or use the suggesting themselves shell patterns -- as a new feature.  (I'd prefer to avoid platform dependent shell patters.)
Though it was implemented using regex only long ago. Some users might be using it really.

I just want to emphasize that the dot will have a special meaning.

vbjay:
I've used regex several times there. It was quite helpful.

RussKie:
I had to look up the code, it looks like we don't allow plain text searches, only regex
https://github.com/gitextensions/gitextensions/blob/186a2d00ad745153937bd758d2900668135b75bd/GitUI/UserControls/FileStatusList.cs#L1340-L1347

This is subtle, and could cause confusion to users.
The label would help alleviate the confusion somewhat, however it would be great to allow plain text searched.
E.g. regexes typically marked with `/` (e.g. `/\w+\d{1.3}/`), and we could apply some smarts around this.
I don't expect the changes in this PR obviously, just thinking out loud.

PR #6450, opened 2019-04-07
Clarify quick search by adding a prefix

### LINKED ISSUE #6419: Text entering over the top of commits tree
## Current behaviour
Text displays over the top of commits tree

## Expected behaviour
User shouldn't have possibility to enter text or text should not be displayed.

## Steps to reproduce

1. Click on higher commit on commits tree

2. Start enter text

3. Then you can see text over the top of commits tree

I don't now

## Environment

- Git Extensions 3.0.2.5232
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

oriash93:
Fixes #6419

## Proposed changes

- Added a string "Searching for: " before the query when performing a quick search, to clarify it's purpose.
- I only changed the label string, so it does not affect the search query nor the logic behind it.

### Before
Untitled3

### After

Untitled2

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -9169,6 +9169,10 @@ Details: Is {1} expected {2}</source>
         <source>Comments</source>
         <target />
       </trans-unit>
+      <trans-unit id="_searchingFor.Text">
+        <source>Searching for: </source>
+        <target />
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/GitUI/UserControls/RevisionGrid/QuickSearchProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/QuickSearchProvider.cs
@@ -4,6 +4,7 @@ using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
 using GitUI.UserControls.RevisionGrid;
+using ResourceManager;
 
 namespace GitUI
 {
@@ -118,7 +119,7 @@ namespace GitUI
         {
             _label.Visible = true;
             _label.BringToFront();
-            _label.Text = _quickSearchString;
+            _label.Text = Strings.SearchingFor + _quickSearchString;
             _label.AutoSize = true;
         }
 
--- a/ResourceManager/Strings.cs
+++ b/ResourceManager/Strings.cs
@@ -47,6 +47,7 @@ namespace ResourceManager
         public static string Submodules => _instance.Value._submodulesText.Text;
 
         public static string BodyNotLoaded => _instance.Value._bodyNotLoaded.Text;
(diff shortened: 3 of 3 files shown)


pmiossec:
Great! But can you please use a 'TranslatedString' so that this string could be translated?

And also, I don't know what others think about it but I think we could improve the change by changing the displayed string when the string is not found (with something like "Searching for (not found):" )

oriash93:
> And also, I don't know what others think about it but I think we could improve the change by changing the displayed string when the string is not found (with something like "Searching for (not found):" )

The text in the label turns red when it's not found, I believe that's enough.

pmiossec:
>The text in the label turns red when it's not found, I believe that's enough.

I made my remark knowing that but because I found it not noticeable enough (a dark red seems to be used). I found it better to have an explicit message for first time users (for me I don't care 😉 ).
But I will let the others decide...

RussKie:
Screenshots?

pmiossec:
>Screenshots?

gitextensions_quick_search

GitUI/UserControls/RevisionGrid/QuickSearchProvider.cs
```diff
@@ -17,6 +17,7 @@ internal sealed class QuickSearchProvider
         private readonly RevisionDataGridView _gridView;
         private readonly Timer _quickSearchTimer;
         private readonly IGitRevisionTester _gitRevisionTester;
+        private readonly string _initialString = "Searching for: ";
```
RussKie: This string must be made localisable. 
Please have a look at the `Strings` class, and add this string there.

RussKie:
> but because I found it not noticeable enough (a dark red seems to be used).

How would you make it more noticeable?

pmiossec:
> How would you make it more noticeable?

😄 https://github.com/gitextensions/gitextensions/pull/6450#issuecomment-480616311

Choose your text:  `Searching for (not found):`  or `Text not found:`  (or other)

GitUI/UserControls/RevisionGrid/QuickSearchProvider.cs
oriash93: Done

GitUI/Translation/English.xlf
```diff
@@ -9169,6 +9169,10 @@ Details: Is {1} expected {2}</source>
         <source>Comments</source>
         <target />
       </trans-unit>
+      <trans-unit id="_searchingFor.Text">
+        <source>Searching For: </source>
```
RussKie:
```suggestion
        <source>Searching for: </source>
```

RussKie:
Thank you

PR #6466, opened 2019-04-11
Highlight the new name in the FileStatusList

### LINKED ISSUE #6465: FileStatusList: highlight the new name
## Current behaviour

The path is displayed gray and the filename black.
For renamed files the new name is gray, too. Only the old is black which usually must be scrolled into view first.

## Expected behaviour

The new filename or both should be displayed in black.

## Steps to reproduce

select a commit with a moved file

### Screenshots

grafik

If yes, then long time ago in 3.x. N/A for 2.5x.

## Environment

- Git Extensions 3.1.0
- Build 827c746e82af33ba959512fba4bd69251e37a782
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6465

## Proposed changes

- detect renamed/copied files
- draw only the new filename black

### Before

before moved
before renamed without path

### After

after moved
after renamed without path

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 4ba5be25f33398170adc8651825b3c0107092ae2
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -430,18 +430,19 @@ namespace GitUI
             }
         }
 
-        private static (Image Image, string Text, int TextStart, int TextWidth, int TextMaxWidth) FormatListViewItem(ListViewItem item, PathFormatter formatter, int itemWidth)
+        private static (Image Image, string Prefix, string Text, string Suffix, int PrefixTextStartX, int TextWidth, int TextMaxWidth)
+            FormatListViewItem(ListViewItem item, PathFormatter formatter, int itemWidth)
         {
             var gitItemStatus = item.Tag<GitItemStatus>();
             var image = item.Image();
             int itemLeft = item.Position.X;
 
-            var textStart = itemLeft + (image?.Width ?? 0);
-            var textMaxWidth = itemWidth - textStart;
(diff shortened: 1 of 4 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -1117,25 +1117,43 @@ private void FileStatusListView_DrawSubItem(object sender, DrawListViewSubItemEv
...
+                string prefix = null;
                 {
-                    formatter.DrawString(text, textRect, SystemColors.ControlText);
-                    return;
+                    int slashIndex = text.LastIndexOf('/');
+                    if (slashIndex >= 0 && slashIndex < text.Length)
+                    {
+                        prefix = text.Substring(0, slashIndex + 1);
+                        text = text.Substring(slashIndex + 1);
+                    }
```
RussKie: Let's extract prefix/suffix calculations into own private method and test these extensively (via `TestAccessor`)

mstv: How to extract the old name safely? Let PathFormatter return it separately! Rewritten including test for new methods.

PR #6467, opened 2019-04-11
feat: JiraCommitHintPlugin credentials secure store

### LINKED ISSUE #4447: Jira Commit Hint plugin credentials stored in plain text

RussKie:
Fixes #4447
Obsoletes #6203

Co-authored-by: Ivan Sterkhov <stivx0@users.noreply.github.com>

## Proposed changes

* Credentials for JIRA plugin are saved to Windows credential manager.

* Global and local (local = distributed in this case) setting levels are supported. It is need for users, who works with several JIRA instances and different accounts.
Distributed level is not supported because credential manager is not a file storage.

## Screenshots 

image

## Test methodology 

- I have done some light testing locally (without access to Jira), and it all seems to work. 

## Test environment(s) 

- Git Extensions 3.1.0
- Build 620367ee5c9f5a999715f362260b6d272b49028e (Dirty)
- Git 2.19.0.windows.1
- .NET Framework 4.7.3324.0
- DPI 144dpi (150% scaling)

----


RussKie:
@stivx0 @gerhardol @pmiossec @PeterKneale I know you use JIRA on a regular basis, could you take it for a spin?

PeterKneale:
Got this prompt on opening settings

image

pmiossec:
>@pmiossec I know you use JIRA on a regular basis

no more... (I won't promise but perhaps one day you will see a PR for a similar plugin for AzureDevops once this one is merged)

GitExtUtils/GitExtUtils.csproj
```diff
@@ -80,6 +80,9 @@
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="AdysTech.CredentialManager">
+      <Version>1.7.0</Version>
```
pmiossec: Small detail  but I prefer to have it on one line (like the others).

gerhardol:
Will test on Monday

Plugins/GitUIPluginInterfaces/CredentialsManager.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                return;
+            }
+
+            Credentials.Clear();
+
+            foreach (var networkCredentials in credentials)
+            {
+                AdysTechCredentialManagerWrapper.UpdateCredentials(networkCredentials.Key,
+                    networkCredentials.Value.UserName,
```
stivx0: Bug what I have mentioned in my pull request is here. Value can be equal null. It is need to use Value?.UserName and Value?.Password.
Test scenario:
1. Remove all settings from Credential manager
2. Open settings
3. Select Local settings
4. Apply settings

p.s. sorry for my long absence. Unfortunately I can't check plugin because our company has enabled two-factor authentication and login/password authentication does not work now.

stivx0: Unfortunately other settings in plugin work incorrect when user switches between different setting levels. I have fixed only new credential settings and therefore Value can be empty.

gerhardol:
Works for me - no review yet

Plugins/GitUIPluginInterfaces/CredentialsManager.cs
RussKie: > Bug what I have mentioned in my pull request is here

Thank you for the repro steps. Fixed.

> Unfortunately other settings in plugin work incorrect when user switches between different setting levels.

That's an existing bug. I'll fix it while I am at it.

RussKie:
Added a fix as part of the PR (happy to separate it, but it depends on the base commit).

## Fix:

The plugin only renders the screen correctly upon load. Switching settings (e.g. effective > local > effective) erases values for the default fields (such as "Jira fields").

The fix addresses the problem in the following way:

* `StringSetting`'s input control is presented by a `TextBox`. Its value is expected to come from settings.
However if the textbox is marked as readonly the value isn't expected to be modified by a user (such as "Jira fields").
Caveat: a readonly control must have its `Text` property set.

* If the textbox must be be configurable by the user, but must retain defaults (or not allow empty input because it makes no sense, e.g. "JQL Query" field) then the setting can be marked with an additional flag `UseDefaultValueIfBlank`.

Also disable "Preview" button whilst a preview is being generated to prevent multiple concurrent calls.

## Screenshots:
* Before

image

* After:
image

PR #6470, opened 2019-04-14
FileStatusList: Indicate active file filter and regex error

### LINKED ISSUE #6449: FileStatusList: indicate active filter
## Feature description

An entered file filter for the FileStatusList can be overlooked too easily.
The ComboBox's window color (or text color for a future dark theme) should be changed if the filter string is not empty.
Maybe we could even indicate the validity of the regex, e.g. green and red.

In preparation of switching color schemes, a new Control class should be created.

Thoughts?

### Before

old

### After

regex valid
ok

regex error
error

## Environment

- Git Extensions 3.1.0
- Build 28fcbf1e864ecc2238ef9001586c537957c6c73b
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6449

## Proposed changes

- indicate active filter of FileStatusList
- indicate regex error in filter of FileStatusList

### Before

old

### After

regex valid
ok

regex error
error

unfiltered
unfiltered focused
unfiltered unfocused

## Test methodology 

- manual tests
- NUnit test for the changes to `FileStatusList.FilterFiles()` factored out as a new method

## Test environment(s) 

- Git Extensions 3.1.0
- Build 0e96050c7ea1bdddf7c98d50ec80529f25e27368
- Git 2.21.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
@@ -59,6 +58,7 @@ namespace GitUI
             _sortByContextMenu = CreateSortByContextMenuItem();
             SetupUnifiedDiffListSorting();
             lblSplitter.Height = DpiUtil.Scale(1);
+            FilterComboBox.Font = new Font(FilterComboBox.Font, FontStyle.Bold);
             InitializeComplete();
             FilterVisible = false;
 
@@ -1338,9 +1338,8 @@ namespace GitUI
 
         private int FilterFiles(string value)
         {
-            _filter = string.IsNullOrEmpty(value)
-                ? null
-                : new Regex(value, RegexOptions.Compiled | RegexOptions.IgnoreCase);
+            StoreFilter(value);
+
             UpdateFileStatusListView(updateCausedByFilter: true);
             return FileStatusListView.Items.Count;
         }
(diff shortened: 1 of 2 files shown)


mstv:
Rebased, added tests and minor improvements.

RussKie:
Thank you for the effort. 
I'm not particular happy with theming implementation, I am not sure how it will scale to the whole app and how it affect testability. 
I would like to have a separate discussion about how we can tackle theming in the context of the whole app with the view of possible offering a dark theme.

In an interim, let's just reference/assign colours directly (e.g. `Color.FromArgb(0xFF, 0xC8, 0xC8)`).

Sorry about the confusing directions.

mstv:
I have removed the class `Theme` but kept the branch with the commit c2fb1a50fa1cbf6c2a9b7db815cf1291f383b4b7 as an example for the discussion what should be considered for the dark theme.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1518,6 +1538,12 @@ public override int Compare(ListViewItem x, ListViewItem y)
 
         #endregion
 
+        #region private Color constants
+        //// Do not declare the colors "static" because Color.FromArgb() will not work at their initialization.
```
gerhardol: nitpick: //// is intended for commenting out code (this "feature" is not used now though)

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1518,6 +1538,12 @@ public override int Compare(ListViewItem x, ListViewItem y)
...
+        private readonly Color _activeInputColor = Color.FromArgb(0xC8, 0xFF, 0xC8);
```
gerhardol: the active color is too visible for my preference but let someone better at UI clean this up...

GitUI/UserControls/FileStatusList.cs
mstv: Yes. Though with only //, the StyleCop scolds "Single-line comment should be preceded by blank line". And I didn't want a blank line after #region here. (The GE code uses both: #region with and without a following blank line.)

GitUI/UserControls/FileStatusList.cs
mstv: It is the same color as used for new lines in the diff view as agreed upon in the feature request.

PR #6479, opened 2019-04-16
Clear diff in history form when no file selected

### LINKED ISSUE #6271: Clear diff in history form when clicked on empty area, just like main form and commit form do
## Feature description

Currently there are 2 different strategies for dealing with clicks on empty areas when showing diffs.

The main form unselects the file and clears the diff view:
main2

The commit form does the same like the main form:
commit

However, the history form behaves slightly differently. Clicking the empty area still de-selects the file, but the diff view is not cleared.
diff_popup2
The behaviour should be aligned to the other two forms. This has been discussed in 
#6231 .

## Environment

GitExtensions 3.

mast-eu:
Fixes #6271

## Proposed changes

- Clear FileViewer in history form when empty area is clicked (i.e. below the changed files). This is the same behavior as already implemented in the main form and commit form. 

### Before

old

### After

new

## Test methodology 

- manual tests

## Test environment(s) 

- Git Extensions 3.1.0
- Build 5d2c8c94ba6b7f3e788e82707dc1ca2c8391f8d0
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 120dpi (125% scaling)

----

### DIFF
--- a/GitUI/UserControls/CommitDiff.cs
+++ b/GitUI/UserControls/CommitDiff.cs
@@ -82,6 +82,7 @@ namespace GitUI.UserControls
         {
             if (DiffFiles.SelectedItem == null || DiffFiles.Revision == null)
             {
+                DiffText.Clear();
                 return;
             }
 


pmiossec:
I understand that it is the way it is handled in other forms but I don't see a clear added value at handling this click like that.

I would have preferred to disable the deselection (or quickly reselect the last file selected) because seing a blank diff won't add value and the user will obligatory select another file just after.

What do you think?

mast-eu:
> I understand that it is the way it is handled in other forms but I don't see a clear added value at handling this click like that.

It's just a really minor aspect, that emerged from another PR (#6231). It has been agreed with @RussKie (https://github.com/gitextensions/gitextensions/pull/6231#issuecomment-462710973) and @gerhardol (https://github.com/gitextensions/gitextensions/issues/6271#issuecomment-465337725).

> I would have preferred to disable the deselection

I don't think this is easily achievable, because you can also click in the empty area (deselecting all files), hold down the mouse button and drag upwards (selecting one or more files).

> or quickly reselect the last file selected

I tried exactly that in #6231, but the UX was not satisfying (see https://github.com/gitextensions/gitextensions/pull/6231#issuecomment-462127295).

> because seing a blank diff won't add value and the user will obligatory select another file just after.

Unfortunately having no file selected is possible, and cannot be easily avoided, as discussed above. So you can end up with having _no file selected_ on the left, but still _seeing some diff_ on the right. This makes no sense and is handled more logically in other parts of GE (_no file selected -> no diff shown_).
Thus this PR to use the same logic everywhere. It is maybe not the best solution, but it is the "least bad" one.

pmiossec:
Thanks for the clarification!

RussKie:
> I would have preferred to disable the deselection (or quickly reselect the last file selected) because seing a blank diff won't add value and the user will obligatory select another file just after.

We need to make it consistent across the app one way or another. The proposed way seems easier to achieve the consistency.
I see some value in not allowing to select nothing, however it requires handling a number of edge cases to ensure the UX is intuitive and correct.
If you feel strongly about it - please send a PR, but I think we can all agree there are other issues in our backlog that can deliver higher value to our users.

RussKie:
Thank you

PR #6490, opened 2019-04-19
Remove executable arg for GitModule

### LINKED ISSUE #6487: Test: GitModule executable should be set with test accessor
## Current behaviour

GitModule has an "executable" argument, that only is used in tests for mocking.
The argument does not make sense for normal modules (see #6373 for a rejected PR with some other optimisations). However, the argument is used incorrectly in test for the submodules provider.

https://github.com/gitextensions/gitextensions/blob/master/UnitTests/CommonTestUtils/GitModuleTestHelper.cs#L138

The git executable has the working directory of the superproject, so it executes in wrong directory.
(Current tests does not use this but updated tests in #6445 fails due to this).

## Expected behaviour
The executable in the submodule tests should not be overridden. (Separate PR will come.)

However, expose the argument makes no sense for the normal usage, this should be set in a test accessor instead.

## Steps to reproduce
code

executable mocking added in 3.0, submodule provider in 3.1

## Environment
- Git Extensions 3.1.0
- Build 5e7bb3c2f5210d859308fd713465489a3b43f484
- Git 2.21.0.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #6487 

## Proposed changes

Only affects test code

This is only used for mocking in tests and should be avoided in the public API.
A TestAccessor is not straightforward (this is a readonly field),
reflection is used instead.

## Test methodology 
Tests still working

## Test environment(s)

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -40,13 +40,13 @@ namespace GitCommands
         private readonly IGitCommandRunner _gitCommandRunner;
         private readonly IExecutable _gitExecutable;
 
-        public GitModule([CanBeNull] string workingDir, [CanBeNull] IExecutable executable = null)
+        public GitModule([CanBeNull] string workingDir)
         {
             WorkingDir = (workingDir ?? "").NormalizePath().EnsureTrailingPathSeparator();
             WorkingDirGitDir = GitDirectoryResolverInstance.Resolve(WorkingDir);
             _indexLockManager = new IndexLockManager(this);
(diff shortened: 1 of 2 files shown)


RussKie:
Instead of reflection could we set these properties via TestAccessor?

gerhardol:
@RussKie 
That was the plan, but the fields are readonly why another level backing variables will be needed. 

An alternative is to make the current constructor private and add a static "test accessor" method to access it. 

Or just name the argument mockExecutable

RussKie:
:+1:

PR #6491, opened 2019-04-19
Execute scripts also if RevisionGrid is not focused

### LINKED ISSUE #6285: NRE on script hotkey unless the revision grid has the focus
## Current behaviour

A NullReferenceException is thrown when the hotkey of a script is hit while the Revision Grid is not focused.

## Expected behaviour

No exception, no hard exit.

## Steps to reproduce

- assign a hotkey to a script
- focus something else than the Revision Grid, e.g. the Diff files list or the Commit Info
- hit the assigned hotkey

## Screenshots

script hotkeys

Yes, partially: Both 2.51.05 and 3.0.0.4433 show an explanation:
grafik

*Not* caused by the appreciated PR # 6219 crypto-rsa/5834-skip-grid-refresh-after-navigational-script.

## Environment

- Git Extensions 3.1.0
- Build 5e825a186d8ea793be4c3bff9893ab56b158700b
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

## Diagnostics

```
   at GitUI.Script.ScriptOptionsParser.ParseScriptArguments(String argument, String option, IWin32Window owner, RevisionGridControl revisionGrid, IGitModule module, IReadOnlyList`1 allSelectedRevisions, IList`1& selectedTags, IList`1& selectedBranches, IList`1& selectedLocalBranches, IList`1& selectedRemoteBranches, IList`1& selectedRemotes, GitRevision selectedRevision, IList`1& currentTags, IList`1& currentBranches, IList`1& currentLocalBranches, IList`1& currentRemoteBranches, GitRevision currentRevision, String currentRemote) in F:\Build\gitextensions\GitUI\Script\ScriptOptionsParser.cs:line 351
   at GitUI.Script.ScriptOptionsParser.Parse(String argument, IGitModule module, IWin32Window owner, RevisionGridControl revisionGrid) in F:\Build\gitextensions\GitUI\Script\ScriptOptionsParser.cs:line 118
   at GitUI.Script.ScriptRunner.RunScript(IWin32Window owner, GitModule module, ScriptInfo scriptInfo, IGitUICommands uiCommands, RevisionGridControl revisionGrid) in F:\Build\gitextensions\GitUI\Script\ScriptRunner.cs:line 83
   at GitUI.Script.ScriptRunner.ExecuteScriptCommand(IWin32Window owner, GitModule module, Int32 command, IGitUICommands uiCommands, RevisionGridControl revisionGrid) in F:\Build\gitextensions\GitUI\Script\ScriptRunner.cs:line 19
   at GitUI.GitModuleForm.ExecuteCommand(Int32 command) in F:\Build\gitextensions\GitUI\GitModuleForm.cs:line 69
   at GitUI.CommandsDialogs.FormBrowse.ExecuteCommand(Int32 cmd) in F:\Build\gitextensions\GitUI\CommandsDialogs\FormBrowse.cs:line 2205
   at ResourceManager.GitExtensionsFormBase.ProcessCmdKey(Message& msg, Keys keyData) in F:\Build\gitextensions\ResourceManager\GitExtensionsFormBase.cs:line 52
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at ResourceManager.GitExtensionsControl.ProcessCmdKey(Message& msg, Keys keyData) in F:\Build\gitextensions\ResourceManager\GitExtensionsControl.cs:line 90
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.Control.ProcessCmdKey(Message& msg, Keys keyData)
```

mstv:
Fixes #6285

## Proposed changes

1. commit: refactoring
  - ScriptOptionsParser.GetCurrentRevision: remove argument that is always null
  - ScriptRunner.RunScript: positive logic in conditions for message
2. commit: resolve regressions introduced by 1b8bb93c7121cd291a17303edfdd313c8edb9f88
  The regression was a NRE instead of the message (see screenshot).
  A second minor regression was that the options were searched without the surrounding '{' and '}'.
  - ScriptOptionsParser to do all parsing (correctly) with factored out methods
  - script.Arguments always in plural
3. commit: feature implementation
  - GitModuleForm and GitModuleControl to provide the RevisionGridControl instance to ScriptRunner
4. commit: tests with a litte refactoring
5. commit: bugfix
  - The script options {cHash}, {cMessage}, {cAuthor}, {cCommitter}, {cAuthorDate}, {cCommitDate} did not work due to currentRevision being null when the script was not run from RevisionGridControl.

### Before

before 1b8bb93c7121cd291a17303edfdd313c8edb9f88
grafik

since 1b8bb93c7121cd291a17303edfdd313c8edb9f88
NRE 

### After

script is executed without neither exception nor message

## Test methodology 

- manual
- added NUnit tests

## Test environment(s) 

- Git Extensions 3.1.0
- Build 7fb79b9aab842d6635e5908c3cdea27a1437dd38
- Git 2.21.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/GitModuleControl.cs
+++ b/GitUI/GitModuleControl.cs
@@ -153,7 +153,13 @@ namespace GitUI
 
             CommandStatus ExecuteScriptCommand()
             {
-                return Script.ScriptRunner.ExecuteScriptCommand(this, Module, command, UICommands, this as RevisionGridControl);
+                var revisionGridControl = this as RevisionGridControl;
+                if (revisionGridControl == null)
+                {
+                    revisionGridControl = (FindForm() as GitModuleForm)?.RevisionGridControl;
+                }
+
+                return Script.ScriptRunner.ExecuteScriptCommand(this, Module, command, UICommands, revisionGridControl);
             }
         }
 
--- a/GitUI/Script/ScriptOptionsParser.cs
+++ b/GitUI/Script/ScriptOptionsParser.cs
@@ -98,7 +98,7 @@ namespace GitUI.Script
 
                 if (option.StartsWith("c") && currentRevision == null)
                 {
-                    currentRevision = GetCurrentRevision(module, revisionGrid, currentTags, currentLocalBranches, currentRemoteBranches, currentBranches, currentRevision);
+                    currentRevision = GetCurrentRevision(module, revisionGrid, currentTags, currentLocalBranches, currentRemoteBranches, currentBranches);
 
                     if (currentLocalBranches.Count == 1)
                     {
@@ -187,41 +187,38 @@ namespace GitUI.Script
         [CanBeNull]
         private static GitRevision GetCurrentRevision(
             IGitModule module, [CanBeNull] RevisionGridControl revisionGrid, List<IGitRef> currentTags, List<IGitRef> currentLocalBranches,
-            List<IGitRef> currentRemoteBranches, List<IGitRef> currentBranches, [CanBeNull] GitRevision currentRevision)
+            List<IGitRef> currentRemoteBranches, List<IGitRef> currentBranches)
         {
-            if (currentRevision == null)
+            GitRevision currentRevision;
+            IEnumerable<IGitRef> refs;
+            if (revisionGrid == null)
             {
-                IEnumerable<IGitRef> refs;
+                var currentRevisionGuid = module.GetCurrentCheckout();
+                currentRevision = new GitRevision(currentRevisionGuid);
+                refs = module.GetRefs(true, true).Where(gitRef => gitRef.ObjectId == currentRevisionGuid).ToList();
+            }
+            else
+            {
+                currentRevision = revisionGrid.GetCurrentRevision();
+                refs = currentRevision.Refs;
+            }
 
-                if (revisionGrid == null)
+            foreach (var gitRef in refs)
+            {
+                if (gitRef.IsTag)
                 {
-                    var currentRevisionGuid = module.GetCurrentCheckout();
-                    currentRevision = new GitRevision(currentRevisionGuid);
-                    refs = module.GetRefs(true, true).Where(gitRef => gitRef.ObjectId == currentRevisionGuid).ToList();
+                    currentTags.Add(gitRef);
                 }
-                else
+                else if (gitRef.IsHead || gitRef.IsRemote)
                 {
-                    currentRevision = revisionGrid.GetCurrentRevision();
-                    refs = currentRevision.Refs;
-                }
-
-                foreach (var gitRef in refs)
-                {
-                    if (gitRef.IsTag)
+                    currentBranches.Add(gitRef);
+                    if (gitRef.IsRemote)
                     {
-                        currentTags.Add(gitRef);
+                        currentRemoteBranches.Add(gitRef);
                     }
-                    else if (gitRef.IsHead || gitRef.IsRemote)
+                    else
                     {
-                        currentBranches.Add(gitRef);
-                        if (gitRef.IsRemote)
-                        {
-                            currentRemoteBranches.Add(gitRef);
-                        }
-                        else
-                        {
-                            currentLocalBranches.Add(gitRef);
-                        }
+                        currentLocalBranches.Add(gitRef);
                     }
                 }
             }
--- a/GitUI/Script/ScriptOptionsParser.cs
+++ b/GitUI/Script/ScriptOptionsParser.cs
@@ -52,7 +52,8 @@ namespace GitUI.Script
             "WorkingDir"
         };
 
-        private static string CreateOption(string option, bool quoted)
+        [NotNull]
+        private static string CreateOption([NotNull] string option, bool quoted)
         {
             var result = "{" + option + "}";
 
@@ -64,7 +65,17 @@ namespace GitUI.Script
             return result;
         }
 
-        public static (string argument, bool abort) Parse(string argument, IGitModule module, IWin32Window owner, RevisionGridControl revisionGrid)
+        public static bool Contains([NotNull] string arguments, [NotNull] string option)
+        {
+            return arguments.Contains(CreateOption(option, quoted: false));
+        }
+
+        public static bool DependsOnSelectedRevision([NotNull] string option)
+        {
+            return option.StartsWith("s");
+        }
+
+        public static (string arguments, bool abort) Parse(string arguments, IGitModule module, IWin32Window owner, RevisionGridControl revisionGrid)
         {
             GitRevision selectedRevision = null;
             GitRevision currentRevision = null;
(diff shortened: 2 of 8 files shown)


mstv:
I see both a Codecov report and a green AppVeyor build for spdr870/gitextensions-f0pvf:

grafik

and a red one for gitextensions/gitextensions:

grafik

Though there is poor coverage - adding some tests.

How to test `MessageBox.Show`?

gerhardol:
OK after forced rebuild - more flaky tests? (I do not believe that these changes introduce instability.)

mstv:
Thank you, Gerhard.

I have written amongst others a new test `RunScript_with_arguments_with_s_option_with_RevisionGrid_shall_succeed` that opens FormBrowse.
It succeeds sometimes and it often fails with a NRE in `ControlThreadingExtensions.SwitchToMainThreadAsync()` due to `ThreadHelper.JoinableTaskFactory` being null.

What can I do to make it run reliably?

mstv:
This time, Codecov was unreachable for both builds twice each.

GitUI/Script/ScriptOptionsParser.cs
```diff
@@ -109,20 +117,20 @@ public static (string argument, bool abort) Parse(string argument, IGitModule mo
                         }
                     }
                 }
-                else if (option.StartsWith("s") && selectedRevision == null && revisionGrid != null)
+                else if (DependsOnSelectedRevision(option) && selectedRevision == null && revisionGrid != null)
```
RussKie: Please perform `null` checks first

GitUI/GitModuleControl.cs
```diff
@@ -153,7 +153,13 @@ protected override CommandStatus ExecuteCommand(int command)
 
             CommandStatus ExecuteScriptCommand()
             {
-                return Script.ScriptRunner.ExecuteScriptCommand(this, Module, command, UICommands, this as RevisionGridControl);
+                var revisionGridControl = this as RevisionGridControl;
+                if (revisionGridControl == null)
+                {
+                    revisionGridControl = (FindForm() as GitModuleForm)?.RevisionGridControl;
```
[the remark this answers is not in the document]
mstv: No, it won't. As before, the result of the `as` operator can be null. In this case, the messagebox "Option {optionDependingOnSelectedRevision} is only supported when started from revision grid." will still be displayed.

GitUI/Script/ScriptOptionsParser.cs
mstv: I can change this. I have just factored out the existing quite simple an IMO cheap condition `.StartsWith`.

Do you suggest this for performance reasons? In general, for understandability of the code, I prefer a logical sequence of conditions over a performance-optimized. Else (not in this case) comments could be needed, which tend to be not maintained.

Then I would perform the `null` check here first, too:
https://github.com/gitextensions/gitextensions/blob/7fb79b9aab842d6635e5908c3cdea27a1437dd38/GitUI/Script/ScriptOptionsParser.cs#L102

RussKie: In cases like this I favour perf - it doesn't really impede readability but guards against possible perf implications. 
null checks are cheap and a function call may be quite expensive (load it into the stack, and execute). It is not impossible that a method small and benign today may turn into a multiline monstrosity tomorrow.

mstv:
I guess #6348 will not be ready for 3.1.0.
There's no hurry to bring in this PR's feature to run scripts if the RevisionGrid is not focused.
Though I'd like to resolve the regression #6285 introduced by 1b8bb93c7121cd291a17303edfdd313c8edb9f88.

Unfortunately, one of the testcases does not work reliably due to race conditions(?) starting FormBrowse. And the completely new ScriptRunnerTests.cs depend on some refactorings which will be done by #6348 in a more generic way.

That's why I would
- put this PR on hold until #6348 will have been merged and
- create another PR which resolves only the regressions without the refactorings and without NUnit tests.

GitUI/Script/ScriptOptionsParser.cs
mstv: done

mstv:
The test case `RunScript_with_arguments_with_s_option_with_RevisionGrid_shall_succeed` often fails because `CheckAndRepairInitialRevision` tries to select row 2 although `_gridView.Rows[]` is empty yet.

Callstack:
```
 	GitUI.dll!GitUI.RevisionGridControl.SetSelectedIndex(int index) Line 534	C#
 	GitUI.dll!GitUI.RevisionGridControl.CheckAndRepairInitialRevision() Line 1121	C#
	GitUI.dll!GitUI.RevisionGridControl.ForceRefreshRevisions.AnonymousMethod__11() Line 1012	C#
 	[External Code]	
 	GitUI.dll!GitUI.GitUICommands.StartBrowseDialog(System.Windows.Forms.IWin32Window owner, string filter, GitUIPluginInterfaces.ObjectId selectedCommit) Line 1152	C#
 	GitUITests.dll!GitUITests.Script.ScriptRunnerTests.RunFormTest.AnonymousMethod__17_0() Line 170	C#
 	GitUITests.dll!GitUITests.UITest.RunForm<GitUI.CommandsDialogs.FormBrowse>(System.Action showDialog, System.Func<GitUI.CommandsDialogs.FormBrowse, System.Threading.Tasks.Task> runAsync) Line 52	C#
 	GitUITests.dll!GitUITests.Script.ScriptRunnerTests.RunFormTest(System.Func<GitUI.CommandsDialogs.FormBrowse, System.Threading.Tasks.Task> testDriverAsync) Line 170	C#
 	GitUITests.dll!GitUITests.Script.ScriptRunnerTests.RunFormTest(System.Action<GitUI.CommandsDialogs.FormBrowse> testDriver) Line 161	C#
 	GitUITests.dll!GitUITests.Script.ScriptRunnerTests.RunScript_with_arguments_with_s_option_with_RevisionGrid_shall_succeed() Line 150	C#
```

PR #6492, opened 2019-04-20
Fix freezing when staging many files with autocrlf

### LINKED ISSUE #6225: Frozen once staging 1000 files with CRLF line break and core.autocrlf = input

Hemaolle:
Fix by setting up reading from output & error buffers before starting to
write to input. When staging files one by one git process will write to
error buffer at the same time as we write commands to the input buffer.
If we don't read from the error buffer before all the input has be
written, the output buffer can fill up, resulting in Git blocking to
wait until we read something.

Fixes #6225

## Test methodology 

- Tested that the issue is fixed on a repository with 1000 files with LF line endings and core.autorcrlf = true.
- Also a case where the files have CRLF line endings and core.autocrlf=input was tested.
- Ran unit tests.

## Test environment(s) 

- GIT git version 2.20.1.windows.1
- Windows 10 version 1803

----

### DIFF
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -295,6 +295,11 @@ namespace GitCommands
 
             using (var process = executable.Start(arguments, createWindow: false, redirectInput: writeInput != null, redirectOutput: true, outputEncoding))
             {
+                var outputBuffer = new MemoryStream();
+                var errorBuffer = new MemoryStream();
+                var outputTask = process.StandardOutput.BaseStream.CopyToAsync(outputBuffer);
+                var errorTask = process.StandardError.BaseStream.CopyToAsync(errorBuffer);
+
                 if (writeInput != null)
                 {
                     // TODO do we want to make this async?
@@ -302,10 +307,6 @@ namespace GitCommands
                     process.StandardInput.Close();
                 }
 
-                var outputBuffer = new MemoryStream();
-                var errorBuffer = new MemoryStream();
-                var outputTask = process.StandardOutput.BaseStream.CopyToAsync(outputBuffer);
-                var errorTask = process.StandardError.BaseStream.CopyToAsync(errorBuffer);
                 var exitTask = process.WaitForExitAsync();
 
                 await Task.WhenAll(outputTask, errorTask, exitTask);
--- a/contributors.txt
+++ b/contributors.txt
@@ -81,3 +81,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/03/24, fschmied, Fabian Schmied, fabian.schmied(at)gmail.com
 2019/03/29, realkompot, Andrei Matsei, therealkompot@gmail.com
 2019/04/09, vbjay, Jay Asbury, vbjay.net@gmail.com
+2019/04/20, Hemaolle, Oskari Leppäaho, oskari.leppaaho@gmail.com


gerhardol:
@mast-eu 
Any comments?

mast-eu:
I would love to test this PR, but I'm on holidays and won't have a PC at hand for the entire next week.

However, I remember that I got to the same conclusion, that the problem is somehow caused by how the StandardError buffer is managed. So @Hemaolle explaination makes perfectly sense to me.

mast-eu:
> Tested that the issue is fixed on a repository with 1000 files with CRLF line endings and core.autorcrlf = true.

@Hemaolle : did you also explicitly test with 1000 files with CRLF line endings and core.autocrlf=input? And 1000 files with LF line endings and core.autocrlf=true? These are the 2 combinations I found to cause problems in #6225.

Hemaolle:
@mast-eu, right, indeed I tested with 1000 files with **LF** endings and core.autocrlf= true. Now tested also the case with CRLF line endings and core.autocrlf=input and it seems to work fine also. I'll update the PR description.

gerhardol:
@Hemaolle 
Tested _with_ showing errors in console?

image

Hemaolle:
@gerhardol, looks like that flag was off for me. Tried now toggling it on. Still no freeze. But I'm not seeing any difference in behavior either, I'm getting a popup with `warning: LF will be replaced by CRLF in 1.txt.
The file will have its original line endings in your working directory` whether the setting you pointed out is on or not.

mast-eu:
> I'm getting a popup with `warning: LF will be replaced by CRLF in 1.txt. The file will have its original line endings in your working directory` 

The presence of this popup is controlled by `Settings` > `Git Extensions` >`Detailed` > `Commit Dialog` > `Show errors when staging files`. It's important you have this option active for all your tests, because when you have it deactivated, there's already another fix in place for the same issue (#6366, works **only** when the option is deactivated).

gerhardol:
@Hemaolle @mast-eu 
Wrong screenshot from me but if you get the popup it is fine. 

I suggest we merge this now and set the 3.1 milestone a little later.

RussKie:
Works for me.
Though I'd like to get 3.1 out of the door asap.

RussKie:
Any thoughts on whether we can design an integration/acceptance test for the fix?

Hemaolle:
You mean a test that would have caught the bug?

I suppose we are not running Git in the tests?

How about creating a dummy executable that would for example print back every line it gets in standard input back to standard error. You could pass that to `ExecuteAsync` and with sufficiently large input it should have hanged without the fix.

RussKie:
> You mean a test that would have caught the bug?

Yes. Primarily to prevent us regressing in the future.


> I suppose we are not running Git in the tests?

Yes, we are :)
We have a number of integration tests that use libgit2sharp, we create git repo and execute git commands.
Please have a look at `ReferenceRepository` and how it is used, I think it should give you some ideas.

PR #6499, opened 2019-04-22
Generate avatars with author initials as fallback of gravatar

pmiossec:
A smaller step than #6417 as it requires no ("not everyone agrees") UI changes but bringing most of the added value...

## Proposed changes

- Add a new new gravatar fallback possibility with avatars generated from author initials

### Before
image

image

### After

image

image

## Test methodology 

* Unit tests
* Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build c937777048244969bee4b89f19e641543dd9a6e7
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3362.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #5189, opened 2018-07-22
Research compile-time XLF tools


RussKie:
As suggested in https://github.com/gitextensions/gitextensions/pull/5178#discussion_r204242511

> It seems like alternatives are now available that would remove the need for TranslateApp altogether. I haven't seen something like this needed in the past; why not just use the compile-time XLF tools?

RussKie:
https://github.com/dotnet/xliff-tasks

mast-eu:
From their readme:

> Many teams using `XliffTasks` default `UpdateXlfOnBuild` to true for local developer builds, but leave it off for CI builds. This way the .xlf files are automatically updated as the developer works, and the CI build will fail if the developer forgets to include the changes to the .xlf files as part of their PR. This way the .xlf files are always in sync with the source files, and can be handed off to a localization team at any time.

Sounds exactly like what we need.

RussKie:
I have read the very same thing the other day.

We'll need enhance our build pipeline to run the translator app as part of
the build and upload the updated xlfs to Transifex.

It is something I'm planning to do once the telemetry is sorted out.
Unless, of course, this is already done before then 😉

RussKie:
I've started making inroads into automating updates to translations.

There are several pieces to the puzzle, as I see it, but there may be other things I'm currently oblivious to.

We would like to:
1. have the English translations to be automatically updated with each PR (if there are any additions/changes);
2. have Appveyor fail a build if the English translations haven't been updated by an author;
3. automate task of downloading/updating other translations for the master/release branches;
I see this task as manual and run on demand, but for the most part it should be a "click-and-submit" operation. 
The reason for being manual - if there are new languages, tx/.config file must be updated with mappings and new language/flag gifs must be added (see #6576);
4. (tied to the point 3) the updating script must update `Setup/MakePortableArchive.cmd` and `Setup/Product.wxs` so we don't have to deal with these manually; this would significantly simply our life and reduce mistakes during release times.

Thoughts?

mast-eu:
> Thoughts? 

As I see it:

- 1 & 2 are mandatory to ensure that the source code and `English.xlf` always remain aligned.
These will fully replace the current TranslationApp, correct?

- There must be a (currently manual?) step of uploading `English.xlf` to Transifex during release preparation.
**Questions**: Can/should this be automated, too?
If yes, can/should this be done continuously or only during release preparation?

- 3 & 4, apart from language mapping, should be possible to automate.
[...]

PR #6505, opened 2019-04-23
Bugfix/settings appearance

gerhardol:
## Proposed changes

Translation related
- Settings-Appearance got lost earlier in 3.1 (#6287 c28c023b5b9bdc50af97962a306818ea0e158cb2),(visible after TranslationApp ran). This is now being updated in the translations, the English.xlf source should be pushed again.
- NoImageService items were translated, with anonymous references. The newly added AuthorInitials in #6499 screwed up this list as it was added first (this is the correct place for this service, but adding must not reorder translations). 

### Before
image

### After

image

## Test methodology 
Manual

To avoid issues like this, we must enforce running the TranslationApp at updates, and fail builds if there are updates.

## Test environment(s)

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
@@ -49,7 +49,7 @@
             this.ShowAuthorAvatarInCommitGraph = new System.Windows.Forms.CheckBox();
             this.ShowAuthorAvatarInCommitInfo = new System.Windows.Forms.CheckBox();
             this.ClearImageCache = new System.Windows.Forms.Button();
-            this.NoImageService = new System.Windows.Forms.ComboBox();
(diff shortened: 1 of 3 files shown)


RussKie:
Thank you

mast-eu:
> To avoid issues like this, we must enforce running the TranslationApp at updates, and fail builds if there are updates.

I second this, because I remember I had some trouble the first time I had to deal with TranslationStrings:
- first I did it manually in XLF files, than by change I read there's a TranslationApp
- the TranslationApp didn't work (now fixed in #6501). I made it work locally but was not confident enough to file a PR.
- once it ran, it updated more strings than I actually changed. This was because the strings were changed in earlier commits but without running the TranslationApp.

RussKie:
Related #5189

gerhardol:
The hard check is to run at all AppVeyor builds and fail if .xlf files updates

Hopefully the PR submitter calls out if he/she cannot figure out the error themselves...

RussKie:
Yes.
Let's continue xlf specific discussions in the dedicated thread

PR #5287, opened 2018-08-04
Use ArgumentBuilder and move SmartFormat package reference

drewnoakes:
SmartFormat is used by the ResourceManager project, and with this
change it's no longer needed in GitCommands. We can move the
PackageReference accordingly.

Manual testing.

PR #6508, opened 2019-04-24
Checkout branch: Fix inopportune line return in message box message

### LINKED ISSUE #6507: Checkout branch: Badly formatted message
## Current behaviour

Checkout branch: Badly formatted message containing a not desired line return.

## Expected behaviour

no line return

## Steps to reproduce

* Checkout a remote branch and choose "reset local branch"
* The badly formatted message appears

## Screenshots

image

don't know

## Environment

- Git Extensions 3.1.0
- Build e738f740a1af363f03d4ea2b89049be5b5cf9b94
- Git 2.21.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 192dpi (200% scaling)

pmiossec:
Fixes #6507

## Proposed changes

- Checkout branch: Remove inopportune line return in message box message

### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 290db968d32089960a018a67e30da65c65182b40 (Dirty)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormCheckoutBranch.cs
+++ b/GitUI/CommandsDialogs/FormCheckoutBranch.cs
@@ -33,8 +33,7 @@ namespace GitUI.CommandsDialogs
             new TranslationString("Don't show me this message again.");
 
         private readonly TranslationString _resetNonFastForwardBranch =
-            new TranslationString("You are going to reset the “{0}” branch to a new location\n" +
-                "discarding ALL the commited changes since the {1} revision.\n\nAre you sure?");
+            new TranslationString("You are going to reset the “{0}” branch to a new location discarding ALL the commited changes since the {1} revision.\n\nAre you sure?");
         private readonly TranslationString _resetCaption = new TranslationString("Reset branch");
         #endregion
 
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -2632,8 +2632,7 @@ Enter valid branch name or select predefined value.</source>
         <target />
       </trans-unit>
       <trans-unit id="_resetNonFastForwardBranch.Text">
-        <source>You are going to reset the “{0}” branch to a new location
-discarding ALL the commited changes since the {1} revision.
+        <source>You are going to reset the “{0}” branch to a new location discarding ALL the commited changes since the {1} revision.
 
 Are you sure?</source>
         <target />


pmiossec:
>Updating the non-english .XLF files is not necessary. They are updated directly from Transifex as part of the release preparation. More details on this can be found in the wiki.

I have removed them. The strangest is that you need to update the english file to be able to test your change but must not commit it. This is very counter-intuitive if you imagine that someone will want to test it before merging...

mast-eu:
> The strangest is that you need to update the english file to be able to test your change but must not commit it. This is very counter-intuitive if you imagine that someone will want to test it before merging...

Sorry if my previous comment was not clear enough. Changes to the english XLF file _may be part of the PR_, but is currently not a strict retirement. It is a _recommendation_ and makes testing and maintaining easier, as you correctly noticed.
Only the _other languages (Czech, Dutch, etc)_ should not be included in the PR since they are updated from Transifex.

I approve the PR anyhow, since it fulfills the current mandatory requirements and solves the underlying issue.

pmiossec:
>Sorry if my previous comment was not clear enough

That was not your message that was not clear but the wiki and also a comment talking about failing if a xlf file was added ( https://github.com/gitextensions/gitextensions/pull/6505#issuecomment-486311318 ).

So, the `english.xlf` file is a little different and I'm not sure how we should handle it.

I just removed all the files because that was the easier for me because that was the original commit I pushed and I have just to take it back from the reflog 😉

mast-eu:
> That was not your message that was not clear but the wiki and also a comment talking about failing if a xlf file was added ( #6505 (comment) ).

The comment is only a proposal on how we might enforce checks in the future. It is not yet in place and thus AppVeyor will not fail. The wiki should be clarified, but this will depend also on the outcome of #5189.

> So, the english.xlf file is a little different and I'm not sure how we should handle it.

If you want, you can add also the modified `english.xlf`, containing only the updated string relative to this PR. This will make @RussKie's life a little easier.

vbjay:
Here is why I think ALL xlf should not ever be included except for
releases. The authority of the strings in the xlf is the actual strings in
the code. Our tooling auto generates the English xlf from this source. You
can run the tooling yourself to update xlf file. You manually alter the xlf
and you get the possiblity of divergence from the code. It's not hard d to
run the tooling so you aren't saving Russkie anything and it's part of the
release cycle.

gerhardol:
It is not an error to include non English .xlf but meaningless as they will be overwritten. (We could actually push these changes to Transifex too, but that will be a little tricky for the maintainer.)

There are reportedly non compile time methods for updating the English .xlf (#5287)..
If this fails, we should either _enforce_ to update .xlf and the translation app by adding it to appVeyor or someone very regulary will have submit PRs with updates, so we do not see issues in the translation handling just before the release #6505.

RussKie:
> Here is why I think ALL xlf should not ever be included except for releases. The authority of the strings in the xlf is the actual strings in the code. Our tooling auto generates the English xlf from this source. You can run the tooling yourself to update xlf file. You manually alter the xlf and you get the possiblity of divergence from the code. It's not hard d to run the tooling so you aren't saving Russkie anything and it's part of the release cycle.

Even if our tooling auto generates the English xlf from the source, the updated xlf must be committed to the repo. So IMO it is sensible to ask a contributor to commit the changes originated from the submitted change.

Please continue XLF-related discussions in  #5189.

PR #6328, opened 2019-03-01
GitHub: easily add upstream remote

pmiossec:
## Proposed changes

~~Note: better reviewable and merged after #6514 (a small part of the refactoring done in this PR was extracted in #6514)~~ done.

- Add a menu item to easily add the "upstream" remote to the repository
- refactoring:  also introduce a refactoring on `GitHubHostedRemote` where the underneath repository should be initialized with the `GitProtocol` already used ( make use of `GitHubHostedRemote` easier)

## Screenshots 

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 9c567e63a5bfb30af412bce86e356e169735d5af
- Git 2.20.0.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3324.0
- DPI 192dpi (200% scaling)

PR #6514, opened 2019-04-26
Refactor: Create an enum GitProtocol.cs

pmiossec:
Refactoring, so fixes nothing...

## Proposed changes

- Introduce a class (edit: after some refactorings, that end up as an enum) `GitProtocol` to handle supported git protocol (a refactoring required to be used in 2 PR #6328 and #6515)
- Easier to review and merge before the 2 PRs...

## Test methodology 

- Added some unit tests

## Test environment(s) 

- Git Extensions 3.1.0
- Build 9602479c6bb5464dee4add337527ffcf47fb35fc (Dirty)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 192dpi (200% scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
@@ -438,11 +438,11 @@ namespace GitUI.CommandsDialogs.RepoHosting
 
             if (repo != null)
             {
-                bool multipleProtocols = repo.SupportedCloneProtocols.Length > 0;
+                bool multipleProtocols = repo.SupportedCloneProtocols.Any();
 
                 if (multipleProtocols && updateProtocols)
                 {
-                    var currentSelection = (string)ProtocolDropdownList.SelectedItem;
+                    var currentSelection = (GitProtocol)(ProtocolDropdownList.SelectedItem ?? repo.SupportedCloneProtocols.First());
                     ProtocolDropdownList.DataSource = CurrentySelectedGitRepo.SupportedCloneProtocols;
                     if (CurrentySelectedGitRepo.SupportedCloneProtocols.Contains(currentSelection))
                     {
@@ -535,7 +535,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
 
         private void ProtocolSelectionChanged(object sender, EventArgs e)
         {
-            CurrentySelectedGitRepo.CloneProtocol = (string)ProtocolDropdownList.SelectedItem;
+            CurrentySelectedGitRepo.CloneProtocol = (GitProtocol)ProtocolDropdownList.SelectedItem;
             SetCloneInfoText(CurrentySelectedGitRepo);
         }
     }
--- a/Plugins/GitHub3/GitHubRepo.cs
+++ b/Plugins/GitHub3/GitHubRepo.cs
@@ -70,23 +70,7 @@ namespace GitHub3
             }
         }
 
-        public string CloneReadWriteUrl
-        {
-            get
-            {
-                string url;
-                if (CloneProtocol == "SSH")
-                {
-                    url = _repo.SshUrl;
-                }
-                else
-                {
-                    url = _repo.CloneUrl;
-                }
-
-                return url;
-            }
-        }
+        public string CloneReadWriteUrl => CloneProtocol == GitProtocol.Ssh ? _repo.SshUrl : _repo.CloneUrl;
 
         public string CloneReadOnlyUrl => _repo.GitUrl;
 
@@ -134,9 +118,9 @@ namespace GitHub3
             return pullRequest.Number;
         }
 
-        public string CloneProtocol { get; set; } = "SSH";
+        public GitProtocol CloneProtocol { get; set; } = GitProtocol.Ssh;
 
-        public string[] SupportedCloneProtocols { get; set; } = new string[] { "SSH", "HTTPS" };
+        public IReadOnlyList<GitProtocol> SupportedCloneProtocols { get; set; } = new[] { GitProtocol.Ssh, GitProtocol.Https };
 
         public override string ToString() => $"{Owner}/{Name}";
     }
--- a/Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
+++ b/Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
@@ -1,0 +1,9 @@
+﻿namespace GitUIPluginInterfaces.RepositoryHosts
+{
+    // https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols
+    public enum GitProtocol
+    {
+        Https = 0,
+        Ssh
+    }
+}
--- a/Plugins/GitUIPluginInterfaces/RepositoryHosts/IHostedRepository.cs
+++ b/Plugins/GitUIPluginInterfaces/RepositoryHosts/IHostedRepository.cs
@@ -47,9 +47,9 @@ namespace GitUIPluginInterfaces.RepositoryHosts
         /// <returns>Pull request number</returns>
         int CreatePullRequest(string myBranch, string remoteBranch, string title, string body);
 
-        string CloneProtocol { get; set; }
+        GitProtocol CloneProtocol { get; set; }
 
-        string[] SupportedCloneProtocols { get; set; }
+        IReadOnlyList<GitProtocol> SupportedCloneProtocols { get; set; }
     }
 
     public interface IHostedBranch
--- a/UnitTests/Plugins/GitUIPluginInterfacesTests/GitProtocolTests.cs
+++ b/UnitTests/Plugins/GitUIPluginInterfacesTests/GitProtocolTests.cs
@@ -1,0 +1,23 @@
+﻿using FluentAssertions;
+using GitUIPluginInterfaces.RepositoryHosts;
+using NUnit.Framework;
+
+namespace GitUIPluginInterfacesTests
+{
+    [TestFixture]
+    public class GitProtocolTests
+    {
+        [TestCase("https://github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("htTps://github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("HTTPS://github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("http://github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("https_github.com/gitextensions/gitextensions.git", false)]
+        [TestCase("ssh://git@github.com/gitextensions/gitextensions.git", false)]
+        [TestCase("git@github.com/gitextensions/gitextensions.git", false)]
+        [TestCase("github.com/gitextensions/gitextensions.git", false)]
+        public void IsUrlUsingHttp(string url, bool expected)
+        {
+            url.IsUrlUsingHttp().Should().Be(expected);
+        }
+    }
+}
--- a/Plugins/GitUIPluginInterfaces/GitUIPluginInterfaces.csproj
+++ b/Plugins/GitUIPluginInterfaces/GitUIPluginInterfaces.csproj
@@ -92,6 +92,8 @@
     <Compile Include="IProcess.cs" />
     <Compile Include="ISettingControlBinding.cs" />
     <Compile Include="PluginsPathScanner.cs" />
+    <Compile Include="RepositoryHosts\GitProtocol.cs" />
+    <Compile Include="RepositoryHosts\GitProtocolExtensions.cs" />
     <Compile Include="SettingLevel.cs" />
     <Compile Include="ISettingsValueGetter.cs" />
     <Compile Include="ObjectId.cs" />
--- a/Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocolExtensions.cs
+++ b/Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocolExtensions.cs
@@ -1,0 +1,13 @@
+﻿using System;
+
+namespace GitUIPluginInterfaces.RepositoryHosts
+{
+    public static class GitProtocolExtensions
+    {
+        public static bool IsUrlUsingHttp(this string url)
+        {
+            return url.StartsWith($"https://", StringComparison.CurrentCultureIgnoreCase) ||
+                url.StartsWith($"http://", StringComparison.CurrentCultureIgnoreCase);
+        }
+    }
+}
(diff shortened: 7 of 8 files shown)


Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const string Https = "HTTPS";
+
+        public static bool IsUrlUsingProtocol(this string url, string protocol)
+        {
+            if (string.IsNullOrWhiteSpace(protocol) || string.IsNullOrWhiteSpace(url))
+            {
+                return false;
+            }
+
+            return url.StartsWith($"{protocol}://", StringComparison.CurrentCultureIgnoreCase);
```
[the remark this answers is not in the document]
pmiossec: >ssh protocol do not require ssh: prefix

I know. I wanted to keep it simple for the moment but I was wrong. Fixed.

>Also, local paths shouls be considered for reference repos

I thought it could be added afterward when required (because I don't know exactly the need)

>However, I do not see that this PR limits the functionality compared to now, it is possible to enhance later.

That was the goal initialy but I derived a little adding this ssh method...

pmiossec:
>I have not run it but see no problem.

There should have one because that was an easy refactoring...

>Can this and relatewd PRs wait until 3.1 is released?

This 3 PRs are not critical, so they could wait as long as needed.
But to make my life easier, I would have wished to see at least this one (#6514) merged.

And #6515 is fixing something that is not working (and could be considered a bug), so maybe we could wished to see merged. In the same time, that never worked, so it could wait longer. 

#6328 could definitely wait.

Plugins/GitUIPluginInterfaces/RepositoryHosts/IHostedRepository.cs
```diff
@@ -49,7 +49,7 @@ public interface IHostedRepository
 
         string CloneProtocol { get; set; }
 
-        string[] SupportedCloneProtocols { get; set; }
+        IReadOnlyList<string> SupportedCloneProtocols { get; set; }
```
RussKie:
```suggestion
        IReadOnlyList<string> SupportedCloneProtocols { get; }
```

Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
```diff
@@ -24,7 +27,12 @@ public static bool IsUrlUsingHttps(this string url)
 
         public static bool IsUrlUsingSsh(this string url)
         {
-            return url.IsUrlUsingProtocol(Ssh);
+            return url.IsUrlUsingProtocol(Ssh) || SshRegex.IsMatch(url);
+        }
```
mstv: Then this function has no benefit over IsGitUrl, hasn't it?

Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return url.IsUrlUsingProtocol(Ssh) || SshRegex.IsMatch(url);
+        }
+
+        public static bool IsGitUrl(this string url)
+        {
+            return GitUrlRegex.IsMatch(url);
+        }
+    }
+}
```
RussKie: I have a bit of discomfort with the proposed API.

1. 🤔  `IsUrlUsingProtocol` appears to be only used in this class, and feels like an implementation detail that shouldn't be part of the public API.

4. ⚠️  All input parameters for public API must be validated. The current implementation is prone to [ANE](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.ismatch?view=netframework-4.6.1#System_Text_RegularExpressions_Regex_IsMatch_System_String_).

6. 💭  Given this class contains a set of extensions, perhaps the following name is more fitting: `public static class GitProtocolExtensions`. 
This however makes protocol definitions look awkward, e.g. `GitProtocolExtensions.Ssh` and `GitProtocolExtensions.Https`. 
Do you think we could define an enum instead? This way we could significantly reduce our public API surface:
    ```
    public enum GitProtocol
    {
        None = 0,
        Http,
        Ssh
    }

    public static class GitProtocolExtensions
    {
        public static bool IsUrlUsingProtocol(this string url, GitProtocol protocol)
        {
            if (string.IsNullOrWhiteSpace(url))
            {
                throw new ArgumentException(nameof(url));
            }

            switch (protocol)
            {
                case GitProtocol.Https:
                {
                    return url.IsUrlUsingProtocol(Https);
                }

                case GitProtocol.Ssh:
                {
                    return url.IsUrlUsingProtocol(Ssh) || SshRegex.IsMatch(url);
                }

                case GitProtocol.Git:
                {
                    return GitUrlRegex.IsMatch(url);
                }

                default:
                {
                    throw new NotSupportedException();
                }
            }
        }
    }
    ```

7. 💭  Is there any reason why we couldn't achieve the desired with [`Uri` class](https://docs.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=netframework-4.6.1#System_Uri__ctor_System_String_System_UriKind_)?
    ```
    public static class GitProtocolExtensions
    {
        public static bool IsUrlUsingProtocol(this string url, GitProtocol protocol)
        {
            if (string.IsNullOrWhiteSpace(url))
            {
                throw new ArgumentException(nameof(url));
            }

            try 
            {
                var uri = new Uri(url, UriKind.Absolute);
                return string.Equals(uri.Scheme, protocol, StringComparer.InvariantCultureIgnoreCase)
            }
            catch (UriFormatException)
            {
                return false;
            }
        }
    }
    ```

Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Text.RegularExpressions;
+
+namespace GitUIPluginInterfaces.RepositoryHosts
+{
+    public static class GitProtocol
```
RussKie: 🤔 
1. The problem with `static` classes is that they impede/complicate testability of implementations that depend on these static classes. 
However the implementation is relatively benign and can remain as is.
2. The public surface of the API is too broad, `IsGitUrl` nor `IsUrlUsingSsh` have any consumers (I've checked #6328 and #6515).
I understand they are added for completeness however we need to be careful adding functionality for sake of adding it.

UnitTests/Plugins/GitUIPluginInterfacesTests/GitProtocolTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [TestCase("HTTPS://github.com/gitextensions/gitextensions.git")]
+        public void IsUrlUsingHttps_Should_be_true(string url)
+        {
+            url.IsUrlUsingHttps().Should().BeTrue();
+        }
+
+        [TestCase("http://github.com/gitextensions/gitextensions.git")]
+        [TestCase("https_github.com/gitextensions/gitextensions.git")]
+        [TestCase("ssh://git@github.com/gitextensions/gitextensions.git")]
+        public void IsUrlUsingHttps_Should_be_false(string url)
```
RussKie: We can merge tests for the same method like so:
```
[TestCase("https://github.com/gitextensions/gitextensions.git", true)]
[TestCase("htTps://github.com/gitextensions/gitextensions.git", true)]
[TestCase("HTTPS://github.com/gitextensions/gitextensions.git", true)]
[[TestCase("http://github.com/gitextensions/gitextensions.git", false)]
public void IsUrlUsingHttps_should_return_expected(string url, bool expected)
{
    url.IsUrlUsingHttps().Should().Be(expected);
}```

File: Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs (hunk 5 above)
Comment (gerhardol) [reply Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs]: I thought about including these too when looking at the follow up PRs. The current usage is really 'if http else ssh' which is not too bad. I had comments about the completeness but also said I was OK with it (I want to raise such concerns for other reviewers).

**Comment by gerhardol** [comment]
> This 3 PRs are not critical, so they could wait as long as needed.

The PRs should be merged soon, just a question if they have to be included in the 3.1.0 release branch.  (and #6515 is a bug, maybe not critical now but https usage should increase over time).

File: Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs (hunk 5 above)
Comment (RussKie) [reply Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs]: These aren't blockers, more raising for the general awareness

File: Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs (hunk 4 above)
Comment (RussKie) [reply Plugins/GitUIPluginInterfaces/RepositoryHosts/GitProtocol.cs]: I've been thinking overnight, are these really "git protocols" or something else?
```public enum GitProtocol
{
    None = 0,
    Http,
    Ssh
}
```

gerhardol: It is correct to say that these are Git protocols. 
There is a special git protocol too as well as local
https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

RussKie:
I pushed a cleanup - I propose to remove all methods that we have no immediate use for.
Looking at #6328 and #6515 - `IsUrlUsingHttps` is the only method we require. If we need other methods - we'll add them then

pmiossec:
>I pushed a cleanup - I propose to remove all methods that we have no immediate use for.

I'm ok with that. That's what I planned to do but lost myself a little 😉 

I fixed what was broken following the cleanup.

Seems good for me.
I will squash if needed once ready for merge...

Plugins/GitHub3/GitHubRepo.cs
```diff
@@ -134,9 +118,9 @@ public int CreatePullRequest(string myBranch, string remoteBranch, string title,
             return pullRequest.Number;
         }
 
-        public string CloneProtocol { get; set; } = "SSH";
+        public GitProtocol CloneProtocol { get; set; } = GitProtocol.Ssh;
 
-        public string[] SupportedCloneProtocols { get; set; } = new string[] { "SSH", "HTTPS" };
+        public IReadOnlyList<GitProtocol> SupportedCloneProtocols { get; set; } = new List<GitProtocol> { GitProtocol.Ssh, GitProtocol.Https };
```
RussKie: Do we actually need to create a list here?
Will the following work?
```suggestion
        public IReadOnlyList<GitProtocol> SupportedCloneProtocols { get; set; } = new[] { GitProtocol.Ssh, GitProtocol.Https };
```

GitUI/CommandsDialogs/SettingsDialog/Pages/SshSettingsPage.cs
```diff
@@ -15,7 +16,7 @@ public partial class SshSettingsPage : SettingsPageWithHeader
         public SshSettingsPage()
         {
             InitializeComponent();
-            Text = "SSH";
+            Text = GitProtocol.Ssh.ToString("G");
```
RussKie: What does this do?

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
```diff
@@ -438,11 +438,11 @@ private void UpdateCloneInfo(bool updateCreateDirTB = true, bool updateProtocols
 
             if (repo != null)
             {
-                bool multipleProtocols = repo.SupportedCloneProtocols.Length > 0;
+                bool multipleProtocols = repo.SupportedCloneProtocols.Any();
 
                 if (multipleProtocols && updateProtocols)
                 {
-                    var currentSelection = (string)ProtocolDropdownList.SelectedItem;
+                    var currentSelection = (GitProtocol)(ProtocolDropdownList.SelectedItem ?? repo.SupportedCloneProtocols.First());
```
RussKie: This doesn't look safe?
Are we certain this won't fail?

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
```diff
@@ -438,11 +438,11 @@ private void UpdateCloneInfo(bool updateCreateDirTB = true, bool updateProtocols
 
             if (repo != null)
             {
-                bool multipleProtocols = repo.SupportedCloneProtocols.Length > 0;
+                bool multipleProtocols = repo.SupportedCloneProtocols.Any();
```
RussKie: Any particular reason?
AFAIK accessing a property is faster than doing a method call, because inside it would do the same property check.

Plugins/GitHub3/GitHubRepo.cs
pmiossec: Surprisingly (because I though I had tested it days ago without success!), yes.

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
pmiossec: It seems to me. We checked that `repo.SupportedCloneProtocols` is not empty 2 lines above.

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
pmiossec: I find it more expressive (and since especially here, we don't care about performance)

GitUI/CommandsDialogs/SettingsDialog/Pages/SshSettingsPage.cs
pmiossec: I searched for the "ssh" key when I refactored and I hesitated to replace it.
Thinking about it again, it was a bad idea. I will revert it.

RussKie:
> GitProtocol.Ssh.ToString("G");

I was curious what ToString call did

GitUI/CommandsDialogs/SettingsDialog/Pages/SshSettingsPage.cs
pmiossec: For an enum, to string return the int value as string and `.ToString("G")` return the name of the key, so here "Ssh".

RussKie:
Ah, could do `(string)GitProtocol.Ssh`

pmiossec:
>Ah, could do `(string)GitProtocol.Ssh`

I don't know if one day I knew it and forgot it but conceptually it seems strange to cast as something that is not your type. I would expect an exception.

But ok for a language trick...

RussKie:
Please ignore that, it was a brain fart, it's almost midnight here, time to
sleep

pmiossec:
>Please ignore that, it was a brain fart, it's almost midnight here, time to sleep

Maybe but `nameof(GitProtocol.Ssh)` is surely the best solution 😉

drewnoakes:
Looks good, but a minor point is that SSH / HTTPS are really just the transport, not the protocol itself. The Git Protocol is typically one of either "dumb" (obsolete) or "smart" (the standard, which can run over SSH or HTTPS).

https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols

PR #6515, opened 2019-04-26
Fix GitHub PR protocol

pmiossec:
Note: better reviewable and merged after #6514

Fixes error when using the feature "GitHub" > "View pull request" :
* "Fetch to pr/ branch" 
* "Add remote and fetch"

when you are not using ssh.

The git command failed because it can't fetch

## Proposed changes

- Auto detect the protocol that should be used

### Before

When using "Fetch to pr/ branch" 
image

When using "Add remote and fetch"
image

### After

When using "Fetch to pr/ branch" 
image

When using "Add remote and fetch"
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build e1485dd7b8252277c01234a9c32df31a3c3b0e7d
- Git 2.21.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 192dpi (200% scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
@@ -27,8 +27,13 @@ namespace GitUI.CommandsDialogs.RepoHosting
         private readonly TranslationString _strCouldNotAddRemote = new TranslationString("Could not add remote with name {0} and URL {1}");
         #endregion
 
+        private GitProtocol _cloneGitProtocol;
+        private IPullRequestInformation _currentPullRequestInfo;
+        private Dictionary<string, string> _diffCache;
         private readonly IRepositoryHostPlugin _gitHoster;
+        private IReadOnlyList<IHostedRemote> _hostedRemotes;
         private bool _isFirstLoad;
+        private IReadOnlyList<IPullRequestInformation> _pullRequestsInfo;
         private readonly AsyncLoader _loader = new AsyncLoader();
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
@@ -55,10 +60,6 @@ namespace GitUI.CommandsDialogs.RepoHosting
         {
             _gitHoster = gitHoster;
         }
-
-        private IReadOnlyList<IHostedRemote> _hostedRemotes;
-        private IReadOnlyList<IPullRequestInformation> _pullRequestsInfo;
-        private IPullRequestInformation _currentPullRequestInfo;
 
         private void ViewPullRequestsForm_Load(object sender, EventArgs e)
         {
@@ -154,6 +155,8 @@ namespace GitUI.CommandsDialogs.RepoHosting
         private void SelectHostedRepositoryForCurrentRemote()
         {
             var currentRemote = Module.GetCurrentRemote();
+            _cloneGitProtocol = ThreadHelper.JoinableTaskFactory.Run(async () => (await Module.GetRemotesAsync())
+                .First(r => r.Name == currentRemote).FetchUrl.IsUrlUsingHttp() ? GitProtocol.Https : GitProtocol.Ssh);
             var hostedRemote = _selectHostedRepoCB.Items.
                 Cast<IHostedRemote>().
                 FirstOrDefault(remote => remote.Name.Equals(currentRemote, StringComparison.OrdinalIgnoreCase));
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -395,16 +401,19 @@ private void _addAsRemoteAndFetch_Click(object sender, EventArgs e)
...
+                var remoteUrl = _cloneWithHttps ? _currentPullRequestInfo.HeadRepo.CloneReadWriteUrl : _currentPullRequestInfo.HeadRepo.CloneReadOnlyUrl;
                 var remoteRef = _currentPullRequestInfo.HeadRef;
 
                 var existingRepo = _hostedRemotes.FirstOrDefault(el => el.Name == remoteName);
                 if (existingRepo != null)
                 {
-                    if (existingRepo.GetHostedRepository().CloneReadOnlyUrl != remoteUrl)
+                    var hostedRepository = existingRepo.GetHostedRepository();
+                    hostedRepository.CloneProtocol = _cloneWithHttps ? GitProtocol.Https : GitProtocol.Ssh;
+                    var hostedRepositoryCloneUrl = _cloneWithHttps ? hostedRepository.CloneReadWriteUrl : hostedRepository.CloneReadOnlyUrl;
```
mstv: I suggest to factor out setting `CloneProtocol` and `*CloneUrl` and
to comment why the `CloneReadWriteUrl` is the right choice for HTTPS.

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -330,6 +333,8 @@ private void LoadDiffPatch()
         }
 
         private Dictionary<string, string> _diffCache;
+        private bool _cloneWithHttps;
```
mstv: Nitpick: Define the private fields in one place and sort them alphabetically, please; also applies to the definitions at line 59-61.

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
pmiossec: I don't like the solution I adopted to fix the problem but that was due to the existing code https://github.com/gitextensions/gitextensions/blob/553db71b4487d38778b40e6d9d3fe43ed2a0a246/Plugins/GitHub3/GitHubRepo.cs#L73 that do a bad abstraction (in this case) of the underlying code (that would be better in my case):
and https://github.com/gitextensions/Git.hub/blob/master/Git.hub/Repository.cs#L39

We have to set the the protocol https to get the good ReadWrite clone url...

I don't understand what you saying, so perhaps you could better explain (for a not native english speaker) or write some (pseudo) code?

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
mstv: `_diffCache`, too, please.

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
mstv: My point is "DRY!". You do two things twice the same way.

`private bool _cloneWithHttps` should be replaced with `private GitProtocol _cloneProtocol`.

The selection of the `cloneUrl` should be factored out, e.g. as a property, and to be explained:
```C#
        /// <summary>
        /// explain why HTTPS must be handled differently
        /// </summary>
        private string CloneUrl { get => _cloneProtocol == GitProtocol.Https ? _currentPullRequestInfo.HeadRepo.CloneReadWriteUrl : _currentPullRequestInfo.HeadRepo.CloneReadOnlyUrl; }
```

pmiossec:
It should be ready. Asking for a review...

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -27,8 +27,13 @@ public partial class ViewPullRequestsForm : GitModuleForm
         private readonly TranslationString _strCouldNotAddRemote = new TranslationString("Could not add remote with name {0} and URL {1}");
         #endregion
 
+        private bool _cloneWithHttps;
```
mstv:
```suggestion
        private GitProtocol _cloneProtocol;
```

vbjay:
Then take off wip in the title.

mstv:
> Then take off wip in the title.

Already done. Unfortunately, GH does not send e-mails when the title or the description are changed.

vbjay:
Got it.

mstv:
I'm going to merge this tomorrow.

PR #6519, opened 2019-05-01
Remove duplicated implementation and call of function LoadSortedRefsAsync

### LINKED ISSUE #6516: Unhandled RefsWarningException "warning: ignoring broken ref"
## Current behaviour

Occasionally, some of my GE repo's refs are overwritten with zeroes. (It resides on an external, self-encrypting SSD from a leading manufacturer, whose standby / hibernation handling seems to be unstable.)
Then `git for-each-ref --sort=-committerdate --sort=-taggerdate --format="%(refname)" refs/` prints "warning: ignoring broken ref ..." (among correct refs) - as in #6442 for remotes, though for a broken local ref this time.

When GE opens this repo, it does not select any commit - without any error message (see screenshot) although `git rev-parse HEAD` returns an error.

When a commit is selected, the `RefsWarningException`, which was added in the PR #6451, is thrown but not handled in a sub-function of `ReloadCommitInfo()`, which contains an almost one-to-one copy of the member function `LoadSortedRefsAsync()`. The latter does handle the exception, the copy does not. When I created #6451, I did not realize that `LoadSortedRefsAsync()` exists twice. Edit: I couldn't because it was duplicated in a parallel branch.

When the exception is ignored by closing the window, the Git warning is displayed in a MessageBox.

## Expected behaviour

Git command errors are not silently ignored but handled correctly.
Git warnings and errors are displayed clearly, not only as an unhandled exception.
(see https://github.com/gitextensions/gitextensions/pull/6513#issuecomment-487369960, too)

Refactoring of CommitInfo.cs to be finished (DRY).

## Steps to reproduce

- overwrite the file `.git\refs\heads\master` with zeroes (41 bytes)
- re-open the repo -> no commit selected
- select a commit -> RefsWarningException
- close the exception dialog with `Alt+F4` -> MessageBox

## Screenshots

no commit selected
RefsWarningException
broken ref warning

at least no unhandled exception in 3.00.00.4433

## Environment

- Git Extensions 3.1.0
- Build 2125c1beef18fd6eac66f368b026e921243067dd
- Git 2.21.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

## Diagnostics

stack trace
```
   at GitCommands.GitModule.GetSortedRefs() in F:\Build\gitextensions\GitCommands\Git\GitModule.cs:line 767
   at GitUI.CommitInfo.CommitInfo.<>c__DisplayClass56_0.<<ReloadCommitInfo>g__LoadSortedRefsAsync|5>d.MoveNext() in F:\Build\gitextensions\GitUI\CommitInfo\CommitInfo.cs:line 365
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at GitUI.CommitInfo.CommitInfo.<>c__DisplayClass56_0.<<ReloadCommitInfo>b__3>d.MoveNext() in F:\Build\gitextensions\GitUI\CommitInfo\CommitInfo.cs:line 295
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at GitUI.ThreadHelper.<>c__DisplayClass13_0.<<FileAndForget>b__0>d.MoveNext() in F:\Build\gitextensions\GitExtUtils\GitUI\ThreadHelper.cs:line 101
```

mstv:
Fixes #6516

## Proposed changes

- remove the unnecessary call of the function `LoadSortedRefsAsync` from `StartAsyncDataLoad`,
  which results in `GetSortedRefs` being called three times on startup
- remove the duplicated implementation which does not handle the possible `RefsWarningException`

## Test methodology 

- manual tests
  finding: On startup, `StartAsyncDataLoad` is called twice. I.e. the startup actions should be optimized. -> #6520

## Test environment(s) 

- Git Extensions 3.1.0
- Build e0cfed7d0d82a5ebb5c2d20f103b25a56bbec874
- Git 2.21.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Net;
 using System.Reactive.Linq;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
@@ -203,25 +202,17 @@ namespace GitUI.CommitInfo
             ReloadCommitInfo();
         }
 
-        private async Task LoadSortedRefsAsync(CancellationToken cancellationToken = default)
+        private async Task LoadSortedRefsAsync()
         {
-            if (cancellationToken == default)
-            {
-                ThreadHelper.AssertOnUIThread();
-                _refsOrderDict = null;
+            ThreadHelper.AssertOnUIThread();
+            _refsOrderDict = null;
 
-                await TaskScheduler.Default.SwitchTo();
-            }
-            else
-            {
-                await TaskScheduler.Default;
-            }
-
+            await TaskScheduler.Default.SwitchTo();
             try
             {
                 var refsOrderDict = ToDictionary(Module.GetSortedRefs());
 
-                await this.SwitchToMainThreadAsync(cancellationToken);
+                await this.SwitchToMainThreadAsync();
                 _refsOrderDict = refsOrderDict;
                 UpdateRevisionInfo();
             }
@@ -299,11 +290,6 @@ namespace GitUI.CommitInfo
 
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    if (_refsOrderDict == null)
-                    {
-                        await LoadSortedRefsAsync(cancellationToken);
-                    }
-
                     // No branch/tag data for artificial commands
                     if (AppSettings.CommitInfoShowContainedInBranches)
                     {
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net;
 using System.Reactive.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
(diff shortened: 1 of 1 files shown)


GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -292,7 +301,7 @@ void StartAsyncDataLoad()
                 {
                     if (_refsOrderDict == null)
                     {
-                        await LoadSortedRefsAsync();
+                        await LoadSortedRefsAsync(cancellationToken);
```
jbialobr: My thinking is that this invocation of `LoadSortedRefsAsync` is not needed and should be just removed. 
The other implementation of `LoadSortedRefsAsync` should be the only one and should remain untouched (passing there a cancellation token from here is wrong). To be super safe, the `LoadSortedRefsAsync` should be protected by an instance of the CancellationTokenSequence.
I believe that this extra `LoadSortedRefsAsync` results from an improper conflict solving while merging https://github.com/gitextensions/gitextensions/pull/6332

mstv: With this call of `await LoadSortedRefsAsync(cancellationToken)` removed at all, `_refsOrderDict` *is* `null` here on startup. It seems to still work, no NRE occurs.

`_refsOrderDict` is `null` even **twice** on startup until it is assigned by the other implementation at line 216. Perhaps this is a starting point for #6520.

Next finding: GetSortedRefs() is not called at all when refreshing FormBrowse (`F5`).
https://github.com/gitextensions/gitextensions/blob/3b015a56b7802c77a3157b080cc0e2cedf0f1851/GitUI/CommitInfo/CommitInfo.cs#L205-L237

jbialobr: >Next finding: GetSortedRefs() is not called at all when refreshing FormBrowse (F5).

This is how it has been working from the beginning. We could mark it as "needs to be refreshed" when there is a not found item during the tags/branches sorting phase.

Issue #6021, opened 2018-12-31
Discussion: telemetry - what and how?


RussKie:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

We have literally no telemetry data besides that reported to us via issue reports on GH. We have no tangible data on OS distribution, most used or least used forms and controls.

**What is the expected behavior?**

I'd like to have an opt-in telemetry that would capture **anonymous** GE-specific data such as but not limited to:
* OS version/flavour/bitness
* number of app-launches since last report
* number of form/control launches since last report
* git version
* number of recent/favourite apps (just numbers no paths/names/etc)

This can be facilitated via cloud offerings which I think can be done in a very cost-effective manner.
The way I envisage it (with relatively limited knowledge of current tech offerings) could be: 
* setup an AWS or Azure account
* have a small database (rdb or nosql)
* have a lambda function that receives telemetry data and inserts into the database
* have a lambda function that would help dealing with possible abuse (e.g. dos attacks)
* setup a CloudFlare account that would limit requests by IP, similar to [what Troy Hunt did for his HIBP site](https://www.troyhunt.com/serverless-to-the-max-doing-big-things-for-small-dollars-with-cloudflare-workers-and-azure-functions/)

The same architecture can be used for bug reports as well - we could have a lambda function that could post a new issue on GH with stack traces and other relevant info.

Thoughts?

vbjay:
![image](https://user-images.githubusercontent.com/3629489/50563431-c21be780-0cea-11e9-9b66-c12335be2b0d.png)
![image](https://user-images.githubusercontent.com/3629489/50563444-dcee5c00-0cea-11e9-81bd-1cffe4f9e68a.png)

RussKie:
Whilst I don't have experience with it, from what I read I don't think it offers what I'm after.
Looking at the [Azure Monitor pricing](https://azure.microsoft.com/en-au/pricing/details/monitor/) the data is only retained for 30 days, where as I'd like to have a longer retention.

PR #6522, opened 2019-05-02
feature: Add ApplicationInsights telemetry

RussKie:
Relates to #6021

The implementation is shamelessly borrowed from https://github.com/NuGetPackageExplorer/NuGetPackageExplorer, credits to @onovotny

## Proposed changes

* Add functionality to capture application telemetry information via AppInsights
* Add privacy statement
* Modify setup to capture user's consent to capture telemetry information
image
* Add functionality to toggle telemetry information capture in a form of a status strip button
image
* Allow configure telemetry via _Settings_
image
* Prompt user to allow telemetry upon start up, if it isn't configured
image
The text is localisable:
image

The following information is captured:
* Application-level:
    * App version number (e.g. 2.0.x.x)
    * Build type (whether the application is an official release build or not)
    * Selected layout settings (such as visibility of the left panel, commit info position etc)
    * Change of selected layout settings
    * Git version (e.g. 2.19.0)
    * SSH client (e.g. OpenSSH or PuTTY)
* Operating System-Level:
    * OS version (e.g. Windows 10.0.17763.0)
    * .NET CLR version (e.g. 4.0.30319.42000)
    * .NET SDK version (e.g. dotnet:2.10.0-24102)
    * Current culture
    * Current UI culture
    * Number of monitors
    * Resolution of all monitors
    * Primary monitor DPI / scale factor

## Screenshots 

* `Startup.Environment`
image

* `Startup.FormBrowse`
image

----


RussKie:
I have initiates a 30 day trial of AppInsights on Azure and would like to merge the current PR into master for us to collect info and see whether we wish to collect more metrics/events, whether we should name events differently, build dashboard/charts etc...

@gitextensions/git-extensions-source you can login into the Azure portal using the credentials shared in the private channel and play with the dashboards.

GitExtensions/Program.cs
```diff
@@ -48,6 +49,13 @@ private static void Main()
                     return UserEnvironmentInformation.GetInformation();
                 };
 
+                DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
+
+                //
+                Application.ApplicationExit += (s, e) => DiagnosticsClient.OnExit();
+                AppDomain.CurrentDomain.UnhandledException += (s, e) => DiagnosticsClient.Notify((Exception)e.ExceptionObject);
+                Application.ThreadException += (s, e) => DiagnosticsClient.Notify(e.Exception);
```
mstv: Nitpick: What comment did you have in mind for line 54?

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
```diff
@@ -0,0 +1,110 @@
...
+            _client = new TelemetryClient();
+
+            var properties = new Dictionary<string, string>
+            {
+                // environment
+                { "DotNetVersion", RuntimeInformation.FrameworkDescription },
+
+                // languages
+                { "CurrentCulture", CultureInfo.CurrentCulture.Name },
+                { "CurrentUICulture", CultureInfo.CurrentUICulture.Name }
```
mstv: We should trace the `AppSettings.CurrentTranslation`, too. (I don't like localized development tools at all.)

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            if (!_initialized)
+            {
+                return;
+            }
+
+            _client.Flush();
+
+            // Allow time for flushing:
+            System.Threading.Thread.Sleep(1000);
```
mstv: This is OK for me since the application still disappears at once. It is only visible in the Process Explorer for this additional second.

mstv:
The portable build is missing a file.
My local debug build seems to work well.
It has created the following additional DLLs:
DLLs

  Problemereignisname:	CLR20r3
  Problemsignatur 01:	GitExtensions.exe
  Problemsignatur 02:	3.2.0.443
  Problemsignatur 03:	5ccae7de
  Problemsignatur 04:	GitExtensions
  Problemsignatur 05:	3.2.0.443
  Problemsignatur 06:	5ccae7de
  Problemsignatur 07:	2
  Problemsignatur 08:	72
  Problemsignatur 09:	System.IO.FileNotFoundException
  Betriebsystemversion:	6.1.7601.2.1.0.256.48
  Gebietsschema-ID:	1031
  Zusatzinformation 1:	e6e6
  Zusatzinformation 2:	e6e6c208b90db2aa028f8f653cabbf23
  Zusatzinformation 3:	dab7
  Zusatzinformation 4:	dab7e87280b076b693a79044fcd2eaee

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
mstv: We should trace `AppSettings.CurrentTranslation`, too. (I don't like localized development tools at all.)

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
mstv: This is OK for me since the application disappears at once. It is only visible in the Process Explorer for this additional second.

GitExtensions/Program.cs
RussKie: Mmm... it was late in the evening... 😊 
Thanks for spotting

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
RussKie: 👍

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
RussKie: I should have mentioned it that I shamelessly borrows implementations from https://github.com/NuGetPackageExplorer/NuGetPackageExplorer.

RussKie:
> We should have an opt-in dialog, especially for European users.

Ah yes, thank you for reminding me about your GDPR rules.
👍 

> The portable build is missing a file.

Thank you, another things I've overlooked.

mstv:
The portable builds are complete now. Though the second AppVeyor build - the `Required` one - references a temporary(?) merge commit, which is only known to GH and cannot be seen in GE:
- Git Extensions 3.2.0.5782 Build 71f996350b2ebd55cbd05d7023c5f4963c77ec4e (Dirty)
- Git Extensions 3.2.0.5781 Build 458c9522a281fddc4de4482646c5f65de8cc3af1 (Dirty)

RussKie:
> The portable builds are complete now.

Thank you for confirming it.

I'm going to add an "opt-out" switch too.

I've been thinking to add telemetry to the RC too. Thoughts?

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
```diff
@@ -0,0 +1,113 @@
...
+    public static class DiagnosticsClient
+    {
+        private static bool _initialized;
+
+        private static TelemetryClient _client;
+
+        public static void Initialize(bool isDirty)
+        {
+            TelemetryConfiguration.Active.InstrumentationKey = "2ef275e3-8850-4305-9d7c-825a60c3d296";
+            TelemetryConfiguration.Active.TelemetryChannel.DeveloperMode = Debugger.IsAttached;
```
onovotny: You no longer need this if you use the latest AppInsights.WindowsDesktop library

GitUI/GitUI.csproj
```diff
@@ -65,6 +66,9 @@
     <PackageReference Include="Microsoft-WindowsAPICodePack-Shell">
       <Version>1.1.3.3</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel">
```
onovotny: I would recommend AppInsights.WindowsDesktop instead as the persistence channel is more robust. It writes to disk first and resends on next launch if needed.

RussKie: Thank you for taking time to review 👍

RussKie:
> We should have an opt-in dialog, especially for European users.

Added a button to toggle the telemetry
image

I think _Settings > General_ is the logic place to add the setting, however the tab is already full and requires some rethinking/rework before we can add the setting.

onovotny:
Couple more things you may want to consider:

The ApplicationInsights.config file has additional configuration:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/ApplicationInsights.config

You can do it in code if you want as well, but you'd need to add the same things, like the PersistenaceChannel.

You can also use the Session initializer to get a session id correlation and there's also modules/initializers for default device and unhandled exceptions.

gerhardol:
> I'm going to add an "opt-out" switch too.

It seems like it is an "opt-in", which is OK

> I've been thinking to add telemetry to the RC too. Thoughts?

I would prefer to have some documentation (and info popup) for inclusion in official builds. Maybe OK for a RC2, but this should maybe be removed in the final build.

 > The following information is captured:

Missing at least:
* Git version
* SSH config (primary OpenSSH or PuTTY)
* Current repo is ssh or http
* diff/mergetool

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -506,8 +507,37 @@ protected override void OnLoad(EventArgs e)
...
+                    new Dictionary<string, string>
+                    {
+                        // layout
+                        { "ShowLeftPanel", MainSplitContainer.Panel1Collapsed.ToString() },
+                        { nameof(AppSettings.ShowSplitViewLayout), AppSettings.ShowSplitViewLayout.ToString() },
+                        { nameof(AppSettings.CommitInfoPosition), AppSettings.CommitInfoPosition.ToString() },
+
+                        // revision grid
+                        { nameof(AppSettings.ShowAuthorAvatarColumn), AppSettings.ShowAuthorAvatarColumn.ToString() },
+                        { nameof(AppSettings.ShowAuthorNameColumn), AppSettings.ShowAheadBehindData.ToString() },
```
mast-eu:
```suggestion
                        { nameof(AppSettings.ShowAuthorNameColumn), AppSettings.ShowAuthorNameColumn.ToString() },
```

RussKie:
Thanks @onovotny 👍 



> > I'm going to add an "opt-out" switch too.
> 
> It seems like it is an "opt-in", which is OK

It is "opt-out" as it enabled by default.



> I would prefer to have some documentation (and info popup) for inclusion in official builds. Maybe OK for a RC2, but this should maybe be removed in the final build.

I added a privacy policy (copied from NuGetPackageExplorer, kudos to Oren).
Not sure what the best way to make the user aware of the telemetry. Ideas are welcome.




> Missing at least:
> 
> * Git version
> * SSH config (primary OpenSSH or PuTTY)
> * Current repo is ssh or http
> * diff/mergetool

Added a collection for git/ssh.
The other two are a bit tricky, can be added later (as well as many other things).

GitUI/Infrastructure/Telemetry/FormBrowseDiagnosticsReporter.cs
```diff
@@ -0,0 +1,45 @@
...
+                new Dictionary<string, string>
+                {
+                    // layout
+                    { "ShowLeftPanel", _owner.MainSplitContainer.Panel1Collapsed.ToString() },
+                    { nameof(AppSettings.ShowSplitViewLayout), AppSettings.ShowSplitViewLayout.ToString() },
+                    { nameof(AppSettings.CommitInfoPosition), AppSettings.CommitInfoPosition.ToString() },
+
+                    // revision grid
+                    { nameof(AppSettings.ShowAuthorAvatarColumn), AppSettings.ShowAuthorAvatarColumn.ToString() },
+                    { nameof(AppSettings.ShowAuthorNameColumn), AppSettings.ShowAheadBehindData.ToString() },
```
mstv: as already mentioned by @mast-eu
```suggestion
                    { nameof(AppSettings.ShowAuthorNameColumn), AppSettings.ShowAuthorNameColumn.ToString() },
```

PrivacyPolicy.md
```diff
@@ -0,0 +1,50 @@
...
+First, a reminder: Git Extensions is provided "as is", without warranty of any kind, express or implied, including but not limited to the warranties of merchantability, fitness for a particular purpose and noninfringement.
+
+With that out of the way, here's a breakdown of all the information we may collect via Application Insights.
+
+### Application-Level
+Includes:
+* Exception information
+  * Could, in rare cases, contain paths packages on your computer
+* Machine name
+* Host name
```
mstv: I cannot see how names of machine and host could help tracking down anything (data minimization!).

PrivacyPolicy.md
```diff
@@ -0,0 +1,50 @@
...
+* Selected layout settings (such as visibility of the left panel, commit info position etc)
+* Change of selected layout settings
+
+### Operating System-Level
+Includes:
+* Architecture (e.g. 32-bit)
+* Version (e.g. Windows 10.0.17763.0)
+* Build (e.g. 17134.1.amd64fre.rs4_release.180410-1804)
+* Available processors/cores (e.g. 8 cores)
+* Machine Name (e.g. MyFastPC)
```
mstv: ditto

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            string sshClient;
+            var sshPath = _sshPathLocator.Find(AppSettings.GitBinDir);
+            if (string.IsNullOrEmpty(sshPath))
+            {
+                sshClient = "OpenSSH";
+            }
+            else if (GitCommandHelpers.Plink())
+            {
+                sshClient = "PutTy";
```
mstv:
```suggestion
                sshClient = "PuTTY";
```

GitCommands/Settings/AppSettings.cs
```diff
@@ -90,6 +90,12 @@ static AppSettings()
             }
         }
 
+        public static bool TelemetryEnabled
+        {
+            get => GetBool("TelemetryEnabled", true);
```
mstv: We could add another Setting like `TelemetryApproved` in oder to ask the user on the first start of GE.
```suggestion
            get => GetBool("TelemetryEnabled", false);
```

GitExtensions/Program.cs
```diff
@@ -48,6 +49,17 @@ private static void Main()
...
+                DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
+                Application.ApplicationExit += (s, e) =>
+                {
+                    DiagnosticsClient.TrackEvent("AppExit");
+                    DiagnosticsClient.OnExit();
+                };
+
+                // Report diagnostics irrespective of whether a debugger attached or not
+                AppDomain.CurrentDomain.UnhandledException += (s, e) => DiagnosticsClient.Notify((Exception)e.ExceptionObject);
+                Application.ThreadException += (s, e) => DiagnosticsClient.Notify(e.Exception);
```
RussKie: @onovotny do I need these having the following in the ApplicationInsights.config:
```
  <TelemetryModules>
    <Add Type="Microsoft.ApplicationInsights.WindowsDesktop.DeveloperModeWithDebuggerAttachedTelemetryModule, AppInsights.WindowsDesktop"/>
    <Add Type="Microsoft.ApplicationInsights.WindowsDesktop.UnhandledExceptionTelemetryModule, AppInsights.WindowsDesktop"/>
    <Add Type="Microsoft.ApplicationInsights.WindowsDesktop.UnobservedExceptionTelemetryModule, AppInsights.WindowsDesktop" />
    <!--<Add Type="Microsoft.ApplicationInsights.WindowsDesktop.FirstChanceExceptionStatisticsTelemetryModule, AppInsights.WindowsDesktop" />-->
  </TelemetryModules>
```

I peeked inside, and they looked like they are handling the same events.

PrivacyPolicy.md
RussKie: It isn't something we collect explicitly, I'd think these are collected by one of packages referenced in ApplicationInsights.config.
I'm guessing, it would be `DeviceTelemetryInitializer`

 @onovotny are you able to shed some light?
Is there any way to exclude these extra props we don't require?

GitExtensions/Program.cs
onovotny: Those are handling many of the same events, it's often easier to let the initializers do it though.

PrivacyPolicy.md
onovotny: Those are set in the AppInsights TelemetryClient:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/80025b5d79cc52485510d422cfa5a0a8159dac83/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L544

Looks like you could manually set RoleInstance to something else using a TelemetryInitalizer and it'd avoid sending it.

GitCommands/Settings/AppSettings.cs
RussKie: I'm planning something better
image

Need to work out how to execute a custom action at the end of the install.

mast-eu: I like this idea. The installer is the right place to inform about telemetry and giving the option to opt-out. In this case I also agree with having the checkbox selected by default (i.e. _opt-out_, not _opt-in_), because it is well visible.

mstv: Yes, that's less intrusive than a nag screen. But the portable version will still need something else.

RussKie: > But the portable version will still need something else.

Bah! Yes, you're right 😒 
Any thoughts?

vbjay:
Instead of registry use a per user setting. Let the registry setting act as
the default. Then portable will be saved just like all settings saved to
user settings.

RussKie:
> Instead of registry use a per user setting. Let the registry setting act as the default. Then portable will be saved just like all settings saved to user settings.

The telemetry setting is store in `C:\Users\<user>\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.setting`.
However I think you somewhat miss the point of the conversation.

A user's consent to capture telemetry information is captured during a setup (please refer to the screenshots in the description). The consent is then injected into the user's settings file.

The question how we can make a user aware of the telemetry in the portable version _in a nice and simple way_. 
I would like it to be "opt out" (i.e. enabled by default).

vbjay:
We have the in a rebase and merge conflict notifications on the bottom of
the browser window. Provide another that shows on or off with colors and
text. When clicked it let's user alter the setting.

PrivacyPolicy.md
RussKie: I've looked through the logs and I don't see them logged

GitUI/Infrastructure/Telemetry/DiagnosticsClient.cs
```diff
@@ -0,0 +1,103 @@
...
+            _client = new TelemetryClient();
+
+            _client.TrackEvent("Monitors", AttachMonitorInformation());
+        }
+
+        public static void OnExit()
+        {
+            if (!Enabled)
+            {
+                return;
```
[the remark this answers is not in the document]
RussKie: Thanks, the call in `Initialize` is meant to go via `TrackEvent()` that has the safe guard.
Fixed

RussKie:
> We have the in a rebase and merge conflict notifications on the bottom of the browser window. Provide another that shows on or off with colors and text. When clicked it let's user alter the setting.

That's a good idea 👍

RussKie:
All cleaned and made up, ready for another round of reviews.
I have updated the PR's description and added screenshots.

The only outstanding item is a better notification for a portable version. @vbjay's idea sounds good.

PrivacyPolicy.md
mstv: Good. Then the regarding three lines could be removed.

mstv: You have already removed these two of the three lines but kept a copy of the original section "Application-Level".

PrivacyPolicy.md
```diff
@@ -0,0 +1,42 @@
...
+    * .NET CLR version (e.g. 4.0.30319.42000)
+    * .NET SDK version (e.g. dotnet:2.10.0-24102)
+    * Current culture
+    * Current UI culture
+    * Number of monitors
+    * Resolution of all monitors
+    * Primary monitor DPI / scale factor
+
+
+### Application-Level
```
mstv: This section is an outdated copy.

PrivacyPolicy.md
```diff
@@ -0,0 +1,42 @@
...
+We would like your permission to collect a little information about how you use Git Extensions. This will help us to make the application better.
+
+First, a reminder: Git Extensions is provided "as is", without warranty of any kind, express or implied, including but not limited to the warranties of merchantability, fitness for a particular purpose and noninfringement.
+
+With that out of the way, here's a breakdown of all the information we may collect via Application Insights.
+
+### Application-Level
+Includes:
+* Exception information (could, in rare cases, contain paths on your computer)
+  * Version number (e.g. 2.0.x.x)
```
mstv: indent could be removed

PrivacyPolicy.md
```diff
@@ -0,0 +1,42 @@
...
+* Selected layout settings (such as visibility of the left panel, commit info position etc)
+* Change of selected layout settings
+* Git version (e.g. 2.19.0)
+* SSH client (e.g. OpenSSH or PuTTY)
+
+
+### Operating System-Level
+Includes:
+    * Version (e.g. Windows 10.0.17763.0)
+    * Machine Name (e.g. MyFastPC)
```
mstv: Can this be removed, too?

onovotny: Unless you're explicitly setting the "RoleInstance" property in an initializer, AppInsights core library will set it to the hostname. It's your choice, but if you do that, I'd include it in the disclosures.

I find OS version quite useful to record so you know what API's you can safely use.

RussKie: @mstv could you elaborate why you are concerned about the hostname?

For me it seems more effort in removing it since it is captured by AppInsights libraries. 
(To a degree) I trust these libraries to make reasonable decisions about collecting a minimal subset of data.

PrivacyPolicy.md
RussKie: Thank you, a bad rebase at midnight...

PrivacyPolicy.md
mstv: I meant only the machine name. It might contain *personal or identifiable information*, e.g. the owner's name.

RussKie: Using Oren's tip, it appeared relatively painless to override the hostname capture.

GitCommands/Settings/AppSettings.cs
RussKie: I made the button more prominent, perhaps it is sufficient.
It is not visible in the dashboard though...

image

RussKie:
Two options, need to decide which one. I don't want both of them.

1. A toolbar button (a1072fb4a687281f7e7a5b01f0e24cfbc94b2fe2)
    + more prominent
    + has look and feel of a button
    - not visible in the dashboard

image

2. A status bar label (7e7df51e2289524ca9ce61a5bc392c61093a2a57)
    + always visible (including the dashboard)
    - doesn't have a look and feel of a button

image


Thoughts? Opinions?

mast-eu:
I vote for the statusbar. It is common these days to have things in the statusbar that are actually clickable and not only labels. Visual Studio, VS Code, my webbrowser, FTP and email clients all have it. Even the trusty MS Office.
Thus, I think the common user can figure out that this is clickable, too.

gerhardol:
Vote for statusbar (assuming it is shown until selected once and accepted/declined).

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            string sshClient;
+            var sshPath = _sshPathLocator.Find(AppSettings.GitBinDir);
+            if (string.IsNullOrEmpty(sshPath))
+            {
+                sshClient = "OpenSSH";
+            }
+            else if (GitCommandHelpers.Plink())
+            {
+                sshClient = "PuTTY";
```
gerhardol: This should be logging the path to find what PuTTY (bundled/external) or what external client that is used.

RussKie:
Alright, status bar it is then. Thank you.

> This should be logging the path to find what PuTTY (bundled/external) or
what external client that is used.

Are you able to help?

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
gerhardol: > Are you able to help? 

Will look at this tonight. Maybe just internal/external putty is sufficient.

GitCommands/Settings/AppSettings.cs
gerhardol: The default should still be off, use opt-in

GitUI/CommandsDialogs/FormBrowse.Designer.cs
```diff
@@ -221,26 +222,32 @@ private void InitializeComponent()
             // toolStripMenuItem2
             // 
             toolStripMenuItem2.Name = "toolStripMenuItem2";
-            toolStripMenuItem2.Size = new System.Drawing.Size(122, 22);
+            toolStripMenuItem2.Size = new System.Drawing.Size(184, 34);
```
gerhardol: The commit that moves to the status bar also increases the size for icons in the button bar, looks ugly. Designer...

GitUI/CommandsDialogs/SettingsDialog/Pages/GeneralSettingsPage.cs
```diff
@@ -156,5 +138,10 @@ private void ShowGitStatus_CheckedChanged(object sender, System.EventArgs e)
         {
             SetSubmoduleStatus();
         }
+
+        private void LlblTelemetryPrivacyLink_LinkClicked(object sender, System.Windows.Forms.LinkLabelLinkClickedEventArgs e)
+        {
+            Process.Start("https://github.com/gitextensions/gitextensions/blob/master/PrivacyPolicy.md");
```
gerhardol: This links to master. Should we link to the release branch instead? Or the wiki?

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                sshClient = "Other";
+            }
+
+            DiagnosticsClient.TrackEvent("AppStart",
+                new Dictionary<string, string>
+                {
+                    { "Portable", AppSettings.IsPortable().ToString() },
+                    { "Git", GitVersion.Current.ToString() },
+                    { "SSH", sshClient },
```
gerhardol: Setting to sshPath will be better I believe, then we can find internal/ external PuTTY as well as use of specific Other (something to improve for Other?)

PrivacyPolicy.md
```diff
@@ -0,0 +1,31 @@
...
+### Application-Level
+Includes:
+* Exception information (could, in rare cases, contain paths on your computer)
+* App version number (e.g. 2.0.x.x)
+* Is portable version
+* Build type (whether the application is an official release build or not)
+* Selected layout settings (such as visibility of the left panel, commit info position etc)
+* Change of selected layout settings
+* Git version (e.g. 2.19.0)
+* SSH client (e.g. OpenSSH or PuTTY)
```
gerhardol: If SSH logging is changed:
SSH client path (e.g. OpenSSH or PuTTY)

GitUI/CommandsDialogs/SettingsDialog/Pages/GeneralSettingsPage.cs
RussKie: Good point. I think linking to the branch's doc is better because it outlines what each version captures.
In set_version_to.py we update the link to the docs, in the same manner we can update the link to the branch's document:
https://github.com/gitextensions/gitextensions/blob/5e825a186d8ea793be4c3bff9893ab56b158700b/Setup/set_version_to.py#L38-L45

GitUI/CommandsDialogs/FormBrowse.Designer.cs
RussKie: I'll double check but I am unable to see any artifacts locally.

GitCommands/Settings/AppSettings.cs
RussKie: Respectfully disagree, if we have it disabled by default no one will turn it on. It is just a fact of life.

However if we have it on by default, those who actively dislike tracking (I'm one of those) will go out of their way and turn if off.

GitUI/CommandsDialogs/FormBrowse.Designer.cs
gerhardol: Should be
image

Is at 100%
image

RussKie:
I didn't like the statusbar placement, so I moved the option under [Help] menu:
image

vbjay:
Bad idea. That means it will be hidden. The whole point of putting it on the statusbar is visibility.

RussKie:
Yes, but also clutter.

I've been thinking over the past week or so - a lot of apps employ telemetry - Visual Studio, VSCode, Resharper to name a few. None of them hide the fact, but neither display permanent notifications (I haven't seen them at least).

We don't intend to hide the fact that we will be collecting telemetry data - it will be visible on the readme, in the setup and in the app's settings; however I don't want clutter the UI with unnecessary information.

vbjay:
Then make sure it defaults to off and then asks the user on first run.

gerhardol:
> Then make sure it defaults to off and then asks the user on first run.

Or in the installer. Default should be off. (The combination of this is that the portable builds will not have telemetry though.)
To have the checkbox in Help or settings is not much different for me.

GitUI/CommandsDialogs/SettingsDialog/Pages/GeneralSettingsPage.cs
RussKie: Mulling it over, the wiki points to [the master version for COC](https://github.com/gitextensions/gitextensions/blob/master/CODE_OF_CONDUCT.md). 
So it is probably ok to point to the privacy policy in the same fashion.

We can create a separate wiki page, if you think it is worth it, but I don't see much value in it.

I have also played with the py script, but it seems complex and brittle (with my limited knowledge of python).

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
RussKie: I'm ok with it.
However would we run a risk of inadvertently exposing sensitive paths?

RussKie:
I've rolled back the alt implementation and restored the statusbar placement to make it prominently visible.

I retain my position that the telemetry should be "on" by default, otherwise we won't get any of it.
A user will have a chance to opt-out during the installation.

I still need to investigate how we can a user for a portable version (I have an idea to test).

RussKie:
@onovotny I wonder if you could give me some guidance.

I'd like to track general startup settings, such as _git version_, _isPortable_, _current translation_ etc.
I'm doing it before showing the first form by building a dictionary of properties and then sending it: 
```
            DiagnosticsClient.TrackEvent("AppStart",
                new Dictionary<string, string>
                {
                    { "Portable", AppSettings.IsPortable().ToString() },
                    { "Git", GitVersion.Current.ToString() },
                    { "SSH", sshClient },
                    { nameof(AppSettings.CurrentTranslation), AppSettings.CurrentTranslation },
                    { nameof(AppSettings.StartWithRecentWorkingDir), AppSettings.StartWithRecentWorkingDir.ToString() },
                });
```

Then in the portal I need to search for a custom event `AppStart` and then expand custom properties:
image

Is this a correct/recommended way of collecting this kind of data?
Or should I fire a series of events - one for each property?

In the end I want to have a understanding of what git version our users use, whether they prefer to install or run portable versions, use OpenSSH vs Putty etc.

RussKie:
I added a check when a user starts an app and telemetry isn't configured, the user will be prompted:
image


The text is localisable:
image

mstv:
@RussKie, do you have a clue why a FormCommit test failed?

```
1) Error : GitUITests.CommandsDialogs.CommitDialog.FormCommitTests.Should_show_committer_info_on_open
System.InvalidOperationException : Value Dispose() cannot be called while doing CreateHandle().
   at System.Windows.Forms.Control.Dispose(Boolean disposing)
   at System.Windows.Forms.ToolStrip.Dispose(Boolean disposing)
   at System.Windows.Forms.StatusStrip.Dispose(Boolean disposing)
   at System.ComponentModel.Component.Dispose()
   at System.Windows.Forms.Control.Dispose(Boolean disposing)
   at System.Windows.Forms.ContainerControl.Dispose(Boolean disposing)
   at System.Windows.Forms.Form.Dispose(Boolean disposing)
   at GitUI.CommandsDialogs.FormCommit.Dispose(Boolean disposing) in C:\projects\gitextensions-f0pvf\GitUI\CommandsDialogs\FormCommit.cs:line 364
   at System.ComponentModel.Component.Dispose()
   at GitUI.GitUICommands.<>c__DisplayClass84_0.<StartCommitDialog>g__Action|0() in C:\projects\gitextensions-f0pvf\GitUI\GitUICommands.cs:line 505
   at GitUI.GitUICommands.DoActionOnRepo(IWin32Window owner, Boolean requiresValidWorkingDir, Boolean changesRepo, EventHandler`1 preEvent, EventHandler`1 postEvent, Func`1 action) in C:\projects\gitextensions-f0pvf\GitUI\GitUICommands.cs:line 303
   at GitUI.GitUICommands.StartCommitDialog(IWin32Window owner, Boolean showOnlyWhenChanges) in C:\projects\gitextensions-f0pvf\GitUI\GitUICommands.cs:line 518
   at GitUITests.CommandsDialogs.CommitDialog.FormCommitTests.<>c__DisplayClass20_0.<RunFormTest>b__0() in C:\projects\gitextensions-f0pvf\UnitTests\GitUITests\CommandsDialogs\FormCommitTests.cs:line 303
   at GitUITests.UITest.RunForm[T](Action showDialog, Func`2 runAsync) in C:\projects\gitextensions-f0pvf\UnitTests\GitUITests\UITest.cs:line 52
   at GitUITests.CommandsDialogs.CommitDialog.FormCommitTests.RunFormTest(Func`2 testDriverAsync, CommitKind commitKind) in C:\projects\gitextensions-f0pvf\UnitTests\GitUITests\CommandsDialogs\FormCommitTests.cs:line 297
   at GitUITests.CommandsDialogs.CommitDialog.FormCommitTests.RunFormTest(Action`1 testDriver, CommitKind commitKind) in C:\projects\gitextensions-f0pvf\UnitTests\GitUITests\CommandsDialogs\FormCommitTests.cs:line 286
   at GitUITests.CommandsDialogs.CommitDialog.FormCommitTests.Should_show_committer_info_on_open() in C:\projects\gitextensions-f0pvf\UnitTests\GitUITests\CommandsDialogs\FormCommitTests.cs:line 53
```

GitUI/CommandsDialogs/FormBrowse.Designer.cs
```diff
@@ -82,7 +82,8 @@ private void InitializeComponent()
             this.revisionGpgInfo1 = new GitUI.CommandsDialogs.RevisionGpgInfoControl();
             this.FilterToolTip = new System.Windows.Forms.ToolTip(this.components);
             this.statusStrip = new System.Windows.Forms.StatusStrip();
-            this.closeStatusBarItem = new System.Windows.Forms.ToolStripStatusLabel();
+            this.tsslblTelemetryEnabled = new System.Windows.Forms.ToolStripStatusLabel();
```
gerhardol: I do not want to occupy space to just show that telemetry is enabled/disabled

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
gerhardol: Still prefer the ssh path

RussKie: Ok, I don't have a strong preference

GitUI/CommandsDialogs/FormBrowse.Designer.cs
RussKie: I'm not 100% happy with the status bar myself. And with the latest change that prompts a user to opt-in/out, I've been thinking of moving it to menus - https://github.com/gitextensions/gitextensions/pull/6522#issuecomment-494367466

RussKie:
> do you have a clue why a FormCommit test failed?

No, we'll have a look later

GitUI/CommandsDialogs/FormBrowse.Designer.cs
gerhardol: As this is set during installation (or first start) it should be enough, no need to constantly inform the user in the toolbar. (about form?)

RussKie: The About form is kind of overload already, but can add to the info being copied/reported by NBug

gerhardol: About would be instead of status bar no need in the report

RussKie:
Version 53464567.

I have moved the telemetry settings from the status bar to the menubar under _Help_ menu - refer to https://github.com/gitextensions/gitextensions/pull/6522#issuecomment-494367466.

I have also somewhat reworked how information is recorded (no changes to _what_ is captured, only _how_ is changed). 
This is to group the info for ease of discovery:
image

PrivacyPolicy.md
RussKie: Done
image

GitUI/Infrastructure/Telemetry/AppStartDiagnosticsReporter.cs
RussKie: Done
image

RussKie:
> do you have a clue why a FormCommit test failed?

Must have been a glitch... Tests pass now

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3232,5 +3248,15 @@ private void FormBrowse_DragEnter(object sender, DragEventArgs e)
                 e.Effect = DragDropEffects.Move;
             }
         }
+
+        private void TsmiTelemetryEnabled_Click(object sender, EventArgs e)
+        {
+            UICommands.StartGeneralSettingsDialog(this);
```
mstv: I'd prefer what I guess other user would expect, too. The only little drawback of this is that the link to the PrivacyPolicy.md is not directly available.
```suggestion
            AppSettings.TelemetryEnabled = !tsmiTelemetryEnabled.Checked;
```
BTW, the time it takes to open the settings dialog is really annoying.

ResourceManager/Strings.cs
```diff
@@ -53,6 +62,9 @@ public static void Reinitialize()
         public static string InstallGitInstructions => _instance.Value._installGitInstructions.Text;
         public static string GitExecutableNotFound => _instance.Value._gitExecutableNotFoundText.Text;
 
+        public static string TelemetryPersmissionCaption => _instance.Value._telemetryPersmissionCaption.Text;
+        public static string TelemetryPersmissionMessage => _instance.Value._telemetryPersmissionMessage.Text;
```
mstv: Replace all `elemetryPersmission` with `elemetryPermission`.

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: I specifically don't want to do that - I want to make users aware of why we collect telemetry and we can't show the link in the menus.

Agree, the settings is a little slow (it is slower in a Debug build), but I don't expect users constantly go back and forth... So delay is probably not an end of the world here. Besides it gives user a second chance to think 😉

ResourceManager/Strings.cs
RussKie: Thank you!

RussKie:
> I cannot build and test the installer due to MSVS 2019.

Please make sure to install the latest Wix toolset - https://wixtoolset.org/

RussKie:
I've had a quick chat with @onovotny today and then had a read of GDPR, and I'm starting to get convinced that we don't need to ask permissions to collect telemetry _because_ we are not collecting any personal information.
All data we collect is about the app, and it cannot be used to identify individual users.

E.g., here's what we collect:


## Page View Properties

Property | Value
-- | -- 
Country or region | Russia 
State or province | Moscow Oblast |  
City | Odintsovo |  
Page view duration | 0 |  
Event time | 12/5/2019, 3:23:18 AM (Local time) |  
Telemetry type | pageView |  
Session Id | a8b87cad-1bf8-4b1f-9497-11021907683a |  
User Id | uDI//FwN5ESVKUIhU6g7oPaOPc7gzH1F8aLNTRVJ5Q0= |  
Application version | 3.2.0 |  
Device type | PC |  
Device model | System Product Name |  
Operating system | Windows 10.0.18362 |  
Client IP address | 0.0.0.0 |  
Cloud role name | Git Extensions |  
Cloud role instance | Git Extensions |  
SDK version | dotnet:2.10.0-24102 |  
Sample rate | 1 |  
View page name | GitUI.CommandsDialogs.FormBrowse |  
Performance | <250ms |  


## Custom Properties
Property | Value
-- | -- 
GE.StartWithRecentWorkingDir | False |  
Monitor primary resolution | 1920x1200 |  
GE.CurrentTranslation | English |  
Monitor secondary1 resolution | 1920x1200 |  
Monitor primary DPI | 96 |  
Thread culture | en-US |  
Monitor count | 3 |  
GE.IsPortable | False |  
Monitor secondary2 resolution | 1920x1200 |  
GE.IsRelease | False |  
Network type | Ethernet |  
CLR version | 4.8.4018.0 |  
UI culture | en-US |  
Time zone | Eastern Standard Time |  
SSH | PuTTY |  
Git | 2.22.0 |  
SSH.Path | C:\Program Files (x86)\GitExtensions\PuTTY\plink.exe

RussKie:
I'm planning to enable the telemetry by default and remove the page from the installer.

sharwell:
@RussKie We should eliminate unnecessary data linked to specific users before enabling telemetry by default. This includes:

* User Id
* Client IP address
* SSH.Path

onovotny:
UserId is already a pseuononymous hash and helps track usage data, ID that an issue?

mstv:
Yes, it is an issue regarding the GDPR. A pseudonymized id makes it still "identifiable data" because a hash function is nearly collision-free.

sharwell:
Put another way, it is not sufficient to only store hashes of information. If the stored telemetry _in combination with another data source_ can reassociate the telemetry with a particular user, then the telemetry would be considered personal data. In this case, a user could derive the User ID separately and accidentally post it via a different mechanism (e.g. a heap dump of a process), and the telemetry immediately becomes personally identifiable.

onovotny:
What is the proper way to correlate sessions for a given user then? Knowing the number of users using x feature is really useful. I don’t need to know the user details.

sharwell:
@onovotny The easiest way is to approximate it using the number of sessions using the feature. Most business decisions you would make by number of users will arrive at the same decision using number of sessions as an input.

RussKie:
I don't know what and how User Id is generated.

I have set User Id field to a custom string:
```cs
_client.Context.User.Id = "Git Extensions User";
```

...yet it isn't sent. Here's the payload we send:
```
{
    "name": "Microsoft.ApplicationInsights.Dev.3d09f61296404682829c3f6ac1ed3e86.PageView",
    "time": "2019-12-10T11:46:38.2079884Z",
    "iKey": "3d09f612-9640-4682-829c-3f6ac1ed3e86",
    "tags": {
        "ai.application.ver": "3.3.0",
        "ai.device.type": "PC",
        "ai.device.id": "5c18ca51-a6b6-4e6d-ab9a-cd2df8f476fd",
        "ai.device.osVersion": "Windows 10.0.18362",
        "ai.device.oemName": "Parallels Software International Inc.",
        "ai.device.model": "Parallels Virtual Platform",
        "ai.cloud.role": "Git Extensions",
        "ai.cloud.roleInstance": "Git Extensions",
        "ai.session.id": "24fb6206-6bc3-4462-b3f3-5261cf215f20",
        "ai.user.id": "W2h0uafNQgL39kLgfhzSxQeAiCMqib+KvShiiu1P37g=",        <--- there!
        "ai.internal.sdkVersion": "dotnet:2.10.0-24102",
        "ai.internal.nodeName": "MacW10"
    },
    "data": {
        "baseType": "PageViewData",
        "baseData": {
            "ver": 2,
            "name": "GitUI.CommandsDialogs.FormRemotes",
            "duration": "00:00:00",
            "properties": {
                "Monitor count": "1",
                "DeveloperMode": "true",
                "GE.IsRelease": "True",
                "CLR version": "4.8.4018.0",
                "Time zone": "AUS Eastern Standard Time",
                "UI culture": "en-GB",
                "Network type": "Ethernet",
                "Thread culture": "en-GB",
                "GE.StartWithRecentWorkingDir": "False",
                "Monitor primary DPI": "144",
                "Monitor primary resolution": "2868x1650",
                "GE.IsPortable": "False",
                "GE.Environment": "development",
                "SSH": "OpenSSH",
                "GE.CurrentTranslation": "English",
                "Git": "2.16.1.windows.4"
            }
        }
    }
}
```

@onovotny any thoughts?

onovotny:
Not sure without digging into the App Insights core code

RussKie:
> Most business decisions you would make by number of users will arrive at the same decision using number of sessions as an input.

I don't think I completely agree with this statement.
Right now our telemetry has records for 430K sessions from some 2,800 opted-in users for the last 90 days. Here are OS stats:

OS | Count | Percentage
--|---|--
Windows 10.0.18363 | 163425 | 37.93%
Windows 10.0.19041 | 78891 | 18.31%
Windows 10.0.19042 | 65888 | 15.29%
Windows 10.0.17763 | 52378 | 12.16%
Windows 10.0.16299 | 14021 | 3.25%
Windows 10.0.18362 | 12057 | 2.80%
Windows 10.0.17134 | 11223 | 2.60%
Windows 6.2.9200 | 9225 | 2.14%
Windows 6.1.7601 | 8959 | 2.08%
Windows 10.0.14393 | 7314 | 1.70%
Windows 6.3.9600 | 4834 | 1.12%
Windows 10.0.(other) | 2655 | 0.61%

Based on number of sessions alone one could make very different assumptions about our user base, e.g. that majority of our users are on W10.
However given that is only ~1.5% of our users (2.8K out of ~200K, based on GitHub download counter) it is a totally wild, and likely incorrect, guess.

PR #6525, opened 2019-05-03
 Format patch: fix patch creation when selection order is invalid 

pmiossec:
Fixes :

When selecting 2 revisions in the wrong order (even if it seems the more logical one: from selected first and to selected last), the patchs are not generated and a blank message box is displayed

When selecting more than 2 revisions, patch are generated in the wrong order.

## Proposed changes

- Order of selection of revisions should not have effect on what is generated (patches are now generated in chronological order). Patches are generated successfully in any selection order.
- prevent an empty message box when no file generated (replace by an error message)
- Don't close form when generation failed

### Before

gitextensions_format_patch
And patches are not generated.

### After

* Patches are generated successfully in any selection order.

* New error message (even if I can't display it anymore due to the fix on selection order):

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 9602479c6bb5464dee4add337527ffcf47fb35fc (Dirty)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormFormatPatch.cs
+++ b/GitUI/CommandsDialogs/FormFormatPatch.cs
@@ -34,6 +34,8 @@ namespace GitUI.CommandsDialogs
             new TranslationString("Patch result");
         private readonly TranslationString _noGitMailConfigured =
             new TranslationString("There is no email address configured in the settings dialog.");
+        private readonly TranslationString _failCreatePatch =
+            new TranslationString("Unable to create patch file(s)");
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
         private FormFormatPatch()
@@ -186,8 +188,15 @@ namespace GitUI.CommandsDialogs
                 }
             }
 
-            MessageBox.Show(this, result, _patchResultCaption.Text);
-            Close();
+            if (string.IsNullOrEmpty(result))
+            {
+                MessageBox.Show(this, _failCreatePatch.Text, _revisionsNeededCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+            }
+            else
+            {
+                MessageBox.Show(this, result, _patchResultCaption.Text);
+                Close();
+            }
         }
 
         private bool SendMail(string dir)
--- a/GitUI/CommandsDialogs/FormFormatPatch.cs
+++ b/GitUI/CommandsDialogs/FormFormatPatch.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.DirectoryServices;
 using System.IO;
 using System.Net;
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormFormatPatch.cs
```diff
@@ -33,6 +34,8 @@ public partial class FormFormatPatch : GitModuleForm
             new TranslationString("Patch result");
         private readonly TranslationString _noGitMailConfigured =
             new TranslationString("There is no email address configured in the settings dialog.");
+        private readonly TranslationString _failCreatePatch =
+            new TranslationString("Fail to create the patch file(s)");
```
RussKie:
```suggestion
            new TranslationString("Unable to create patch file(s)");
```

pmiossec: Thanks! Fixed.

pmiossec:
FYI, I have made the PR on v3.1
It seems to be the more logical for me.

gerhardol:
> FYI, I have made the PR on v3.1
It seems to be the more logical for me.

The release/3.1 branch already exists, the standard is to merge to master first, then cherry pick to release if required. I would like to get 3.1 released and merge PRs to master again.

RussKie:
> > FYI, I have made the PR on v3.1
> > It seems to be the more logical for me.
> 
> The release/3.1 branch already exists, the standard is to merge to master first, then cherry pick to release if required. I would like to get 3.1 released and merge PRs to master again.

I prefer the same please.

RussKie:
❗️This fix touches translations, we shouldn't be pulling anything that changes strings in to an upcoming release.
I'm willing to make an exception - there is still a week left before a release, and this error message isn't something a lot of users see.

> > FYI, I have made the PR on v3.1
> > It seems to be the more logical for me.
> 
> The release/3.1 branch already exists, the standard is to merge to master first, then cherry pick to release if required. I would like to get 3.1 released and merge PRs to master again.

I prefer the same.

Please action it quickly.

pmiossec:
️>This fix touches translations, we shouldn't be pulling anything that changes strings in to an upcoming release.
I'm willing to make an exception - there is still a week left before a release, and this error message isn't something a lot of users see.

I know that but it was still made in purpose. With this fix, I don't know if there is a way to see the message. This is just in case because that's better to see a not translated message than an empty message box.

>I prefer the same.

I was not sure of the GE workflow, so I made my branch from the last ancestor commit 😉. 
That way, I was able to easily do my PR from my smartphone!

So, done!

PR #6530, opened 2019-05-05
Always keep a manually entered commit message

### LINKED ISSUE #6244: Commit message not stored after reverting a commit
## Current behaviour

A manually adapted commit message is not stored during an unfinished merge.

I know that there are reasons and test cases which ensure this behavior. Though there are valid use cases where this does not make very much sense.

## Steps to reproduce

- `Revert this commit...` with `Automatically create a commit` unchecked
- optional: modify the restored code
- enter an adapted commit message in the Commit dialog
- close the (modal!) Commit dialog (e.g. to lookup something in another commit)
- reopen the Commit dialog
- --> frustration

Even worse:
- try to revert a commit of a different branch
- no changes
- but the commit message 'Revert "..."' is sticky

## Expected behaviour

A manually entered commit message must not be discarded.

No.

## Environment

- Git Extensions 3.1.0
- Build 4ebb916d67f24b09e4d8cc2213bae9be8b2cecfd
- Git 2.20.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6244

## Proposed changes

- store the commit message in the correct(!) file (either MERGE_MSG or COMMITMESSAGE)
  - refactor `CommitHelper` into `CommitMessageManager`
- disable changing the commit message and amending a commit unless `CommitKind.Normal`
- replace the commit message in an undo-able way when recent messages or commit templates are applied etc.

### After

for *fixup* and *squash* commits
fixup

## Test methodology

- manual tests
- NUnit tests for `CommitMessageManager`

## Test environment(s)

- Git Extensions 3.1.0
- Build 678b69e632a2bff81c5e093770f605e0fe1151c8
- Git 2.21.0.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

----


GitCommands/CommitHelper.cs
```diff
@@ -4,74 +4,83 @@ namespace GitCommands
...
-
-            if (AppSettings.RememberAmendCommitState && File.Exists(GetAmendPath(module)))
+            string mergeMessagePathName = GetMergeMessagePath(module);
+            if (File.Exists(mergeMessagePathName))
             {
-                var amendSaveStateFilePath = GetAmendPath(module);
-                bool.TryParse(File.ReadAllText(amendSaveStateFilePath), out amendState);
-                try
-                {
-                    File.Delete(amendSaveStateFilePath);
```
mstv: This was unexpected, unnecessary and simply wrong in the case of a crash in FormCommit.

GitCommands/Git/GitModule.cs
```diff
@@ -868,16 +868,6 @@ public string GetCommitCountString(string from, string to)
                 (added > 0 ? ("+" + added) : "");
         }
 
-        public string GetMergeMessage()
```
mstv: moved to CommitHelper.cs where it belongs to

GitCommands/CommitHelper.cs
```diff
@@ -4,74 +4,83 @@ namespace GitCommands
 {
     public static class CommitHelper
     {
-        public static void SetCommitMessage(GitModule module, string commitMessageText, bool amendCommit)
+        public static string GetCommitMessagePath(GitModule module)
```
RussKie: Any API that pass an instance of `IGitModule` around is bad.
It looks like we only need `WorkingDirGitDir` and `CommitEncoding`, let's pass them instead

GitCommands/CommitHelper.cs
```diff
@@ -4,74 +4,83 @@ namespace GitCommands
...
+            return GetFilePath(module, "GitExtensions.amend");
         }
 
-        public static string GetCommitMessagePath(GitModule module)
+        private static bool GetAmendState(GitModule module)
         {
-            return GetFilePath(module, "COMMITMESSAGE");
+            bool amendState = false;
+
+            if (AppSettings.RememberAmendCommitState)
```
RussKie: Please invert to reduce nesting

mstv:
I have created the `CommitMessageManager `. More tests will follow.

mstv:
> I suggest we take this as an opportunity to clean and refactor the API.

Thank you for giving the direction. I like it much more this way, too.
I have completed the NUnit tests.

RussKie:
If this is the same as #6600 - please merge away

Issue #6402, opened 2019-03-24
Unable to use translated strings and date formatting in project GitCommands


pmiossec:
(Issue asked to be opened in https://github.com/gitextensions/gitextensions/pull/6393#issuecomment-475467630 ) 

## Current behaviour

We can't use the translated strings `Strings.cs` (or `LocalizationHelpers.cs`)  in `GitBlame.cs` ( tGithttps://github.com/gitextensions/gitextensions/blob/16aea8ca2ad8b3d4bf84f0f26c1c6f71ff0684d0/GitUI/UserControls/BlameControl.cs#L48 ) because the project ResourceManager that contains the translated strings has the GitCommands (where this code lies) as a dependency 😢

## Expected behaviour

Being able to use translated strings  `Strings.cs` (or `LocalizationHelpers.cs`)

## Steps to reproduce

!?! Code 😉

pmiossec:
I had a look and that's not an easy task because relative date formatting and translations depends on `AppSettings` to get the current translation that require GitCommands to read settings in the repository.

PR #6531, opened 2019-05-05
Fix rebase and apply patches displayed patches status

pmiossec:
Fixes a wrong status displayed 

## Proposed changes

- Form Rebase & Form Apply Patches: The status on each row display the good patch state
- The transiant "skipped" state is handled
- The "skip"button label as been changed to better explain what it does (skip the currently applying patch and not skipping the selected commit/patch)
- Little ergonomical improvement: The "Applying" patch row is selected when opening the form

### Before

The patches applied are wrongly displayed as "skipped"
image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build 1acb4c9940f8bf0d7c828b3a7a3a1ff77c246cc0
- Git 2.20.1.windows.1
- .NET Framework 4.7.3394.0
- DPI 96dpi (no scaling)

### DIFF
--- a/GitCommands/Patches/PatchFile.cs
+++ b/GitCommands/Patches/PatchFile.cs
@@ -19,19 +19,25 @@ namespace GitCommands.Patches
         public bool IsNext { get; set; }
 
         public bool IsSkipped { get; set; }
+        public bool IsApplied { get; set; }
 
         public string Status
         {
             get
             {
-                if (IsNext)
-                {
-                    return "Next to apply";
-                }
-
                 if (IsSkipped)
                 {
                     return "Skipped";
+                }
+
+                if (IsApplied)
+                {
+                    return "Applied";
+                }
+
+                if (IsNext)
+                {
+                    return "Applying...";
                 }
 
                 if (!string.IsNullOrEmpty(FullName) && !File.Exists(FullName))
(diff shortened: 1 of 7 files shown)


GitCommands/Patches/PatchFile.cs
```diff
@@ -19,19 +19,25 @@ public class PatchFile
...
-                if (IsSkipped)
+                if (IsApplied)
                 {
-                    return "Skipped";
+                    return "Applied";
+                }
+
+                if (IsNext)
+                {
+                    return "Applying...";
```
RussKie: These should be translated 🤔

pmiossec: Oh! Thanks a lot for the remainder.

I planned to do it but I didn't remember if I already looked at it and it was impossible due to #6402 or if I purely forgot to do it....

I will have a look!

pmiossec: Unfortunately, that's due to #6402. We have to fix it before being able to translate it 😢 
I didn't remember but I also wrote it in the commit message: https://github.com/gitextensions/gitextensions/pull/6531/commits/ec4180ea5a40bc8b90374ac454dde369def995dd

RussKie: Thank you for checking.

PR #6539, opened 2019-05-08
Fix new lines being eaten by commit form

### LINKED ISSUE #5908: Commit message line endings modified by commit dialog
I am experiencing a bug/regression with 3.0

When I create a commit using the commit dialog, line endings appear to be getting removed, or converted from Windows (CRLF) to Unix (LF). I haven't been able to confirm which of these is the actual cause, but the result is that all the commit message lines appear to be squashed into one line in the commit message after the commit has completed.

This occasionally does not occur on my initial commit made from a commit template. It always seems to occur if I amend a commit by bringing up the previous commit message via the drop down menu.

**Environment**
- GIT version: 2.12.2.windows.2
- OS version: Microsoft Windows NT 10.0.16299.0
- .NET version: 4.7.3221.0

This was not happening with version 2.51 prior to upgrading to 3.0.

bsruth:
Fixes #5908 

## Proposed changes

- Allow empty lines to be processed by the commit form
- 
- 

## Test methodology 

- Tested with a failing repo for over 2 weeks and did not have a single commit generate the error.
- 
- 

## Test environment(s) 

- GIT 2.12.2
- Windows 7 SP1

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -2199,12 +2199,12 @@ namespace GitUI.CommandsDialogs
                 {
                     // When a committemplate is used, skip comments
                     // otherwise: "#" is probably not used for comment but for issue number
-                    if ((!string.IsNullOrEmpty(line) && !line.StartsWith("#")) ||
+                    if ((line != null && !line.StartsWith("#")) ||
                         string.IsNullOrEmpty(_commitTemplate))
                     {
                         if (addNewlineToCommitMessageWhenMissing)
                         {
-                            if (lineNumber == 1)
+                            if (lineNumber == 1 && !string.IsNullOrEmpty(line))
                             {
                                 textWriter.WriteLine();
                             }
--- a/contributors.txt
+++ b/contributors.txt
@@ -82,3 +82,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/03/29, realkompot, Andrei Matsei, therealkompot@gmail.com
 2019/04/09, vbjay, Jay Asbury, vbjay.net@gmail.com
 2019/04/20, Hemaolle, Oskari Leppäaho, oskari.leppaaho@gmail.com
+2019/05/07, bsruth, Brian Ruth, bsruth@gmail.com


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2199,12 +2199,12 @@ private void SetCommitMessageFromTextBox(string commitMessageText)
                 {
                     // When a committemplate is used, skip comments
                     // otherwise: "#" is probably not used for comment but for issue number
-                    if ((!string.IsNullOrEmpty(line) && !line.StartsWith("#")) ||
+                    if ((line != null && !line.StartsWith("#")) ||
                         string.IsNullOrEmpty(_commitTemplate))
```
mstv: If `string.IsNullOrEmpty(_commitTemplate)` is true, this lets the (very unlikely) `line == null` into the block to the final `WriteLine(line)`.
The logic was and is quite hard to read. And `AppSettings.AddNewlineToCommitMessageWhenMissing` is not self-explaining: In the settings dialog, it is labeled `"Ensure the second line of commit message is empty"`.
As @RussKie requested, we should really have unit tests for this function.
Before I try to describe everything (positive logic, combine nested `if`, etc.), I take the liberty to suggest:
```C#
            using (var textWriter = new StreamWriter(path, false, encoding))
            {
                bool addNewlineToCommitMessageWhenMissing = AppSettings.AddNewlineToCommitMessageWhenMissing;
                bool usingCommitTemplate = !string.IsNullOrEmpty(_commitTemplate);

                var lineNumber = 1;
                foreach (var line in commitMessageText.Split('\n'))
                {
                    string nonNullLine = line ?? string.Empty;

                    // When a committemplate is used, skip comments and do not count them as line.
                    // otherwise: "#" is probably not used for comment but for issue number
                    if (usingCommitTemplate && nonNullLine.StartsWith("#"))
                    {
                        continue;
                    }

                    if (addNewlineToCommitMessageWhenMissing && lineNumber == 2 && !string.IsNullOrEmpty(nonNullLine))
                    {
                        textWriter.WriteLine();
                    }

                    textWriter.WriteLine(nonNullLine);

                    lineNumber++;
                }
            }
```

RussKie: > And `AppSettings.AddNewlineToCommitMessageWhenMissing` is not self-explaining: In the settings dialog, it is labeled `"Ensure the second line of commit message is empty"`.

I've noticed that too.
We should rename it, but it can be done separately. Perhaps call it `EnsureCommitSecondLineEmpty`?

mstv:
@bsruth, are you going to add unit tests for this function?
Should we explain how to do this in more detail?

illfated:
Also, it is useful to create a separate PR for `contributors.txt`.

bsruth:
@mstv Sorry, been pretty busy lately and haven't had a chance to get to adding the tests. I looked at it briefly a couple weeks ago, and couldn't find a good way to access the commit template string without having to restructure a bunch of code. I should have time to play around with it next week.

mstv:
No need to apologize, Brian, real life takes precedence.

Both the private function to be tested, `SetCommitMessageFromTextBox`, and the `_commitTemplate` can be made accessible using the `TestAccessor` class at the end of the file. The `StreamWriter` might be mocked up by means of a new function argument of the type `IFileSystem`.

mstv:
@bsruth, I would like to include the bugfix in the next release. Do you mind me completing this PR?

mstv:
Thank you for the analysis of the bug, @bsruth.

Superseded by #6877 with added NUnit tests.

bsruth:
@mstv Wish I would have been able to complete it for you. I was looking forward to completing my first PR on a real open source project. Might take a stab again when things slow down.

RussKie:
> I was looking forward to completing my first PR on a real open source project.

Please don't despair, we have plenty of open issues to tackle.
And we always welcome new contributors 😄

PR #6545, opened 2019-05-09
fix: Mark portable version as such

RussKie:
It appears we aren't marking portable builds as such:
image

/cc: @vbjay


vbjay:
Yes. I do t know how it got changed but the call without the flag is to
reset it after the make portable call.

RussKie:
A published "portable.zip" was no marked as portable

Issue #6415, opened 2019-03-28
Plugin installation path


maraf:
## Feature description

In PluginManager scenario, when GE is in the write-protected directory, UAC must be confirmed everytime PM tries to do changes.

In maraf/GitExtensions.PluginManager#59 and gitextensions/gitextensions#5775 we have briefly discussed that all plugins could be placed somewhere else, eg `CommonApplicationData`.

I would like to discuss the solution and then would be happy to do a PR.

RussKie:
User's settings are stored in `C:\Users\<user>\AppData\Roaming\GitExtensions\GitExtensions`, would this be a good place to store plugins too?

mast-eu:
Quick check on software with downloadable plugins found on my PC:
- Visual Studio uses `C:\Users\<user>\AppData\Local\Microsoft\VisualStudio\15.0_b13c254b\Extensions`
- Visual Studio Code uses `C:\Users\<user>\.vscode\extensions`
- GIMP uses `C:\Users\martin.steinisch\AppData\Roaming\GIMP\2.10\plug-ins`

And a few more I looked up on the internet:
- [Firefox](https://developer.mozilla.org/en-US/docs/Plugins/Guide/Plug-in_Basics#How_Gecko_Finds_Plug-ins) uses `C:\Users\<user>\AppData\Local\Mozilla\plugins`
- [Chrome](https://stackoverflow.com/questions/14543896/where-does-chrome-store-extensions) uses `C:\Users\<user>\AppData\Local\Google\Chrome\User Data\Default\Extensions `
- [Notepad++](http://docs.notepad-plus-plus.org/index.php/Plugin_Central) uses `C:\Users\<user>\AppData\Roaming\Notepad++\plugins`

There seems to be a common agreement to use `C:\Users\<user>\AppData\<Local|Roaming>`, with VS Code being the only notable exception I found. But even VS Code uses `C:\Users\<user>`, so different users on the same machine are isolated against each other.

What's the difference between `C:\Users\<user>\AppData\Local` and `C:\Users\<user>\AppData\Roaming`? [MSDN explains](https://msdn.microsoft.com/en-us/library/windows/desktop/bb776892(v=vs.85).aspx) that Roaming User Profiles are automatically synced when the user logs on any PC on the same (company) network. On the other hand, Local User Profiles remain on a single PC.
IMHO it's better to not sync plugins between PCs. Maybe they have different GE versions, .NET versions or other troublemakers.

[...]

PR #6551, opened 2019-05-12
Loading user plugins from app data folder.

maraf:
Solves #6415.

In addition to standard plugin location, this change adds:

In portable mode: GE loads plugins from &lt;GE&gt;\UserPlugins.
In installed mode: GE loads plugins from &lt;GEAppData&gt;\UserPlugins.

Questions: 
- The implementation places user plugins in `Roaming` profile, where most of GE stuff is placed, but I have found that `Local` is also used by GE - window positions are saved there. Should user plugins be `Roaming` or `Local`?

- Method `ManagedExtensibility.SetUserPluginsPath` should be internal, but needs to be called from `PluginRegistry`. Can I add `AssemblyInternalsVisibleToAttribute`?


Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
```diff
@@ -13,6 +13,17 @@ namespace GitUIPluginInterfaces
 {
     public static class ManagedExtensibility
     {
+        /// <summary>
+        /// Gets a root path where user installed plugins are located.
+        /// </summary>
+        public static string UserPluginsPath { get; private set; }
```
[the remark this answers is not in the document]
maraf: I'm going move to the PluginManager there in the upcoming PR, so the folder will always exist.

mstv:
> Please make `Local` the default for user plugins.

The intention of `Roaming` is that when I log on to another PC, my settings including my selected plugins do roam to this PC. So why not `Roaming` as the default?
(Window positions depend on the _local_ monitor.)

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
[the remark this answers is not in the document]
mast-eu: Good point.

mast-eu:
> So why not `Roaming` as the default?

Suppose you have 2 PCs with some GE version (say 3.2) and some PluginA.
Now you update 1 PC to GE version 4.0, which brings incompatible changes in the plugin API. 
The PluginManager finds a new version of PluginA, already compatible with the new API, and installs it in `Roaming`.
Now you log into PC 2, which did not yet get GE update to version 4.0 (I think it is checking every 7 days for new updates). But the new version of PluginA has already been synced from PC 1 and you have a problem.

API changes will be frequent, at least while we're getting everything up and running.
We can save a lot of trouble by avoiding to sync plugins through roaming. The downside of having to install plugins on multiple PCs shouldn't be a big problem for the majority of our users.

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
```diff
@@ -79,6 +86,21 @@ private static ExportProvider CreateExportProvider(string applicationDataFolder)
...
 
+        private static IEnumerable<FileInfo> GetPluginFiles(params string[] pluginsPaths)
+        {
+            var result = Enumerable.Empty<FileInfo>();
+
+            foreach (string pluginsPath in pluginsPaths)
+            {
+                if (!string.IsNullOrEmpty(pluginsPath) && Directory.Exists(pluginsPath))
+                {
+                    result = Enumerable.Concat(result, new DirectoryInfo(pluginsPath).GetFiles("*.dll"));
```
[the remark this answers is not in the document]
maraf: Ok, here I just used code that was there..

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -31,6 +32,9 @@ public static void Initialize()
 
                 try
                 {
+                    string userPluginsPath = Path.Combine(AppSettings.ApplicationDataPath.Value, "UserPlugins");
```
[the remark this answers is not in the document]
maraf: Ok

maraf:
Applied changes based on codereview.

- Refactored Local profile path from `WindowPositionList` to `AppSettings`.
- UserPlugins are loaded from `Local` profile.
- Plugins are loaded recursively under plugins folders.

GitCommands/Settings/AppSettings.cs
```diff
@@ -82,6 +85,24 @@ static AppSettings()
                 }
             });
 
+            LocalApplicationDataPath = new Lazy<string>(() =>
+            {
+                if (IsPortable())
+                {
+                    return GetGitExtensionsDirectory();
+                }
+                else
```
RussKie: this `else` is redundant and should be removed to reduce nesting

GitUI/WindowPositionList.cs
```diff
@@ -44,17 +45,9 @@ public class WindowPositionList
         private static readonly string AppDataDir = Path.Combine(
             Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "GitExtensions");
 
-        private static readonly string ConfigFilePath = Path.Combine(AppDataDir, "WindowPositions.xml");
+        private static readonly string ConfigFilePath = Path.Combine(AppSettings.LocalApplicationDataPath.Value, "WindowPositions.xml");
```
RussKie: This will cause a reset of all windows after an upgrade... 🤔 
May not seem like a big deal but could upset some users.

Can we/how hard is it to load from local settings, if it exists, and fallback to load from roaming?
All changes should be written to the local folder though.

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
```diff
@@ -79,6 +86,21 @@ private static ExportProvider CreateExportProvider(string applicationDataFolder)
...
+            foreach (string pluginsPath in pluginsPaths)
+            {
+                if (!string.IsNullOrEmpty(pluginsPath) && Directory.Exists(pluginsPath))
+                {
+                    result = Enumerable.Concat(result, new DirectoryInfo(pluginsPath).GetFiles("*.dll", SearchOption.AllDirectories));
+                }
+            }
+
+            return result;
+        }
```
RussKie: 1. Tests?
2. Do we need to load _all_ \*.dlls or only those matching a specific patten (e.g. GitExtensions.*.dll)? 
3. Is there a perf risk of searching for a looooooong time? 🤔

GitUI/WindowPositionList.cs
mast-eu: The resulting path is the same as before, it is just constructed differently.

So there should be no reset of window positions.

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
maraf: 1. Yeah, tests. I'm sorry, I'm going to add them.
2. I had a discussion with @mast-eu that GE org could reserve prefix GitExtensions.* on nuget.org, so that official/core plugins could be in this form and community plugins will have any other pattern.
3. The idea is that there will only be 1 level of folder nesting.

maraf:
@RussKie What do you think about this?

> Method `ManagedExtensibility.SetUserPluginsPath` should be internal, but needs to be called from `PluginRegistry`. Can I add `AssemblyInternalsVisibleToAttribute`?

GitUI/WindowPositionList.cs
```diff
@@ -44,17 +45,9 @@ public class WindowPositionList
         private static readonly string AppDataDir = Path.Combine(
```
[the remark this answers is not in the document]
maraf: Yeah, I forget it there, I definitely wanted to remove it.

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
mast-eu: 2. The current plugins have no common naming scheme. So, at least for the moment, we need to load all DLLs from `defaultPluginPath`. A decision about enforcing a naming scheme or not can be taken later.
3. No, I don't expect a performance impact. I didn't perform any tests yet though.

mast-eu: 2. The current plugins have no common naming scheme. So, at least for the moment, we need to load all DLLs from `defaultPluginPath`. A decision about enforcing a naming scheme or not can be taken later.
3. No, I don't expect a performance impact. I didn't perform any tests yet though.

RussKie: > GE org could reserve prefix GitExtensions.* on nuget.org

👍  Any hints how to do that?


> 3\. The idea is that there will only be 1 level of folder nesting.

I'm afraid we can't assume that. We need to be strict in our implementation and don't go deeper than one level.

GitUI/WindowPositionList.cs
RussKie: Good, thank you for the confirmation.

RussKie:
> @RussKie What do you think about this?
> 
> > Method `ManagedExtensibility.SetUserPluginsPath` should be internal, but needs to be called from `PluginRegistry`. Can I add `AssemblyInternalsVisibleToAttribute`?

I haven't had a chance to look deep in the code.
It is on my todo list (...and it is getting longer by the day). I'll try to review it over the weekend.

RussKie:
> @RussKie What do you think about this?
> 
> > Method `ManagedExtensibility.SetUserPluginsPath` should be internal, but needs to be called from `PluginRegistry`. Can I add `AssemblyInternalsVisibleToAttribute`?

Do you see any problems with `ManagedExtensibility.SetUserPluginsPath` remaining `public`?

maraf:
> > @RussKie What do you think about this?
> > > Method `ManagedExtensibility.SetUserPluginsPath` should be internal, but needs to be called from `PluginRegistry`. Can I add `AssemblyInternalsVisibleToAttribute`?
> 
> Do you see any problems with `ManagedExtensibility.SetUserPluginsPath` remaining `public`?

Only that any plugin can call it, but if it's ok for you, I'm ok with it.

GitCommands/Settings/AppSettings.cs
maraf: Ok

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
maraf: > > GE org could reserve prefix GitExtensions.* on nuget.org
> 
> 👍 Any hints how to do that?

Send email to account@nuget.org, documentation https://docs.microsoft.com/en-us/nuget/reference/id-prefix-reservation#id-prefix-reservation-application-process.

> 
> > 3. The idea is that there will only be 1 level of folder nesting.
> 
> I'm afraid we can't assume that. We need to be strict in our implementation and don't go deeper than one level.

Ok than.

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
```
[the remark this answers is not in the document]
maraf: Ok

GitUI/WindowPositionList.cs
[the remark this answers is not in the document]
maraf: Ok

maraf:
Resolved notes from code review vol.2.
Not sure about the unit test, is it ok this way?
Should I squash the commits?

RussKie:
> Only that any plugin can call it..

Can we make so that once it is initialized it can't be changed?

maraf:
> Can we make so that once it is initialized it can't be changed?

Sure.

mast-eu:
Just browsed through the unit tests. They look ok.

mast-eu:
I think commits should be squashed before merging.

RussKie:
I'm currently away, so if you happy with it - merge it.

UnitTests/Plugins/GitUIPluginInterfacesTests/PluginsPathScannerTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using FluentAssertions;
+using GitUIPluginInterfaces;
+using NUnit.Framework;
+
+namespace GitUIPluginInterfacesTests
+{
+    [TestFixture]
+    public class PluginsPathScannerTests
+    {
+        [TestCase(@"..\..\PathScanningData", "PluginInRootDir.dll", "PluginInOwnDir.dll")]
```
RussKie: Could you clarify it for me (I'm a bit slow, sorry).

Plugins are stored in `c:\Users\<user>\Local\GitExtensions\GitExtensions\UserPlugins` folder:
```
UserPlugins
  \Plugin1
      \Plugin1.dll
      \Plugin1.config
  \Plugin2
      \Plugin2.dll
      \Plugin2.config
```

In this scenario this looks like:
```
PathScanningData
  \PluginInRootDir
      \PluginInRootDir.dll
  \NotFoundPlugin
      \SubDir
          \NotFoundPlugin.dll1.dll
  \PluginInRootDir.dll
```

Why are we expecting to load `PluginInRootDir.dll`?

maraf: Backward compatibility. I'm using the method for both core and user plugins. Is it ok? Also PM now extracts plugins in a single directory.

maraf:
I'm this weekend away, so I'm going to squash them on Monday.

UnitTests/Plugins/GitUIPluginInterfacesTests/PluginsPathScannerTests.cs
RussKie: Ah, makes sense, thank you.
all good

Plugins/GitUIPluginInterfaces/PluginsPathScanner.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var result = Enumerable.Empty<FileInfo>();
+
+            foreach (string pluginsPath in pluginsPaths)
+            {
+                if (!string.IsNullOrEmpty(pluginsPath) && Directory.Exists(pluginsPath))
+                {
+                    DirectoryInfo directory = new DirectoryInfo(pluginsPath);
+
+                    result = Enumerable.Concat(result, directory.GetFiles("*.dll"));
```
RussKie: I would add a paragraph explaining why we are loading two levels of assemblies.
E.g. load plugins located in the folder to provide back compat, and load new style plugin...
Maybe even add a structure to help visualise
```
UserPlugins
  \OldPlugin1.dll
  \OldPlugin2.dll
  \OldPlugin2.config
  \NewPlugin1
      \Plugin1.dll
      \Plugin1.config
  \NewPlugin2
      \Plugin2.dll
      \Plugin2.config
```

mast-eu:
Will merge after squash.

Plugins/GitUIPluginInterfaces/PluginsPathScanner.cs
maraf: Ok

maraf:
Squashed commits and added detailed comment to the `PluginsPathScanner.cs`.

maraf:
Any hint what is wrong with the build? 
All GitUITests failed..

mast-eu:
> Any hint on what is wrong with the build?

Like with most software, sometimes AppVeyor builds just need to be restarted ... :wink:

mast-eu:
Thank you.

PR #6576, opened 2019-05-16
Update translations

RussKie:
Update translations

* Update how translations are downloaded
* Allow translations with 75% completion in the dev cycle
* Remove translations which are below 75% mark
* Exclude translations from GitUI project to reduce maintenance aspect, instead copy available translations via MSBuild targets; targets in GitExtensions projects are required for the setup packaging
* Add new languages / flags

Relates to #5189

----


RussKie:
Sorry it is WIP, it's not exactly doing what I want it yet.

RussKie:
> The translations may still be out of sync with the source translation, but should be aligned with English.xlf if that is pushed.

The English translations were updated in 0b872d9e0183d0287c4db5637ec777ffc1b4e568, so the translations should be reasonably up to date.
Obviously there is more automation to be done.

RussKie:
Please feel free to take over

PR #6600, opened 2019-05-18
Store the commit message in the correct file

### LINKED ISSUE #6244: Commit message not stored after reverting a commit

mstv:
Fixes #6244
I'd like to have this bug fixed in 3.1.1.

## Proposed changes

- store the commit message in the correct(!) file (either MERGE_MSG or COMMITMESSAGE)
  - refactor `CommitHelper` into `CommitMessageManager`
- disable changing the commit message and amending a commit unless `CommitKind.Normal`
- replace the commit message in an undo-able way when recent messages or commit templates are applied etc.

### After

Commit message, message menus and `Amend` button disabled for *fixup* and *squash* commits
fixup

## Test methodology

- manual tests
- NUnit tests for `CommitMessageManager`

## Test environment(s)

- Git Extensions 3.1.0
- Build 82f125d435dc966b877c8bf63eb1ecdd882fc627
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 96dpi (no scaling)

----


mstv:
I'd like to have this bug fixed in 3.1.1, @RussKie.

RussKie:
There are few things I'd like to change, I'm not particularly happy with
the statefulness and concerned it may lead to obscure bugs.
We can push it out in 3.1.2

mstv:
> I'm not particularly happy with the statefulness and concerned it may lead to obscure bugs.

I have not added any state. I just made it work correctly. Loading  and storing of the message is symmetrical now.
Maybe the whole commit message handling should be reworked. Then we should open an additional issue for it.

RussKie:
I'll find sometime to share my thoughts with you

GitCommands/CommitMessageManager.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+        }
+
+        private CommitMessageManager(string workingDirGitDir, Encoding commitEncoding, IFileSystem fileSystem)
+        {
+            _fileSystem = fileSystem;
+            _commitEncoding = commitEncoding;
+            _amendSaveStatePath = GetFilePath(workingDirGitDir, "GitExtensions.amend");
+            _commitMessagePath = GetFilePath(workingDirGitDir, "COMMITMESSAGE");
+            _mergeMessagePath = GetFilePath(workingDirGitDir, "MERGE_MSG");
```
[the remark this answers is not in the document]
RussKie: 🤔  Perhaps it is worth to use an instance of `IFullPathResolver` here constructed as
```
var resolver = new FullPathResolver(() => workingDirGitDir)
```

GitCommands/CommitMessageManager.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            set
+            {
+                if (AppSettings.RememberAmendCommitState && value)
+                {
+                    _fileSystem.File.WriteAllText(_amendSaveStatePath, true.ToString());
+                }
+                else
+                {
+                    _fileSystem.File.Delete(_amendSaveStatePath);
```
RussKie: 🤔 Would it no fail if the file doesn't exist or otherwise inaccessible?

GitCommands/CommitMessageManager.cs
```diff
@@ -0,0 +1,124 @@
...
+            set
+            {
+                _fileSystem.File.WriteAllText(GetMergeOrCommitMessagePath().FilePath, value ?? string.Empty, _commitEncoding);
+            }
+        }
+
+        public void ResetCommitMessage()
+        {
+            _fileSystem.File.Delete(_commitMessagePath);
+            _fileSystem.File.Delete(_amendSaveStatePath);
```
RussKie: ditto

GitCommands/CommitMessageManager.cs
mstv: It does not fail if the file does not exist (idempotent function).
If the file is not accessible, it should throw an exception because this should not happen with a GE-private file and I am opposed to silently ignore real errors regardless of their severity.

GitCommands/CommitMessageManager.cs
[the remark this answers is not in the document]
mstv: I understand your point.
Though that's where you guided me to: https://github.com/gitextensions/gitextensions/pull/6530#pullrequestreview-233799973. A lambda `Func<string> getWorkingDirGitDir` would have the same problem and is less visible IMO.
And I think, a change of the repo needs consideration everywhere in GE at all. The lifetime of the instance is bound to FormCommit and it's not very likely to be used in a different manner.

If you insist, I would always create new instances of the `CommitMessageManager` and hide this behind a static interface (facade).

> Perhaps it is worth to use an instance of IFullPathResolver.

(There are two known bugs, which shouldn't have a negative effect on accessing files in the `.git` folder.)
My main concern is that the lambda will be hidden even more.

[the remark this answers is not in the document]
RussKie: This isn't a blocking comment, more like a discussion point. I'm always looking at how we can reduce a duplication of the code, especially when we have existing implementations already tested.
Perhaps I worry too much :) and you're right, the _current_ context is scoped to `FormCommit`, so your implementation is fine.

My original suggestion was about not passing an instance of `IGitModule`, as it may lead to operating on a different repo to what a user may be currently in.

GitCommands/CommitMessageManager.cs
RussKie: Fair point, however we need to be mindful of a UX - we shouldn't crash the app because of that.

PR #6604, opened 2019-05-19
Create branch or tag: Not for artificial commits

### LINKED ISSUE #6597: Create branch on "working directory" creates revision 11111 instead of last known commit
## Current behaviour
I wanted to create a branch on the current branch selected "master" for example to do a merge request but didn't look where I stood in the current tree of commits so just hit Ctrl+B

Standing on "Working Directory" showed this dialog:

image

There is no way to add this and then create commits to it. To fix this I had to go out from that dialog and then select the last known commit in the tree and create a branch.

image

Then it works.

## Expected behaviour
Either add branch from last known commit in the tree or make selecting the last known commit in the dialog.

Or just skip and do not allow creating a branch on "Working directory"

## Steps to reproduce

1. Go to any repo
2. Select the "Working directory" commit in the tree
3. Ctrl+B to create a new branch

Its happened on all the repos I have tested so far including GitExtensionDocs repo

Not sure

## Environment

- Git Extensions 3.1.0.5877
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #6597

Low priority

## Proposed changes
For artificial commits, use HEAD instead
This affected commands invoked by the hardcoded Ctrl-B and Ctrl-T shortcuts only (the shortcut in FormBrowse is not invoved).

The menu items are disabled for artificial commands already, the shortcut is not blocked though. The commands could be blocked in RevisionGridControl (as well as FormBrowse, even if that code will not be executed as the menu items are checked). However, the Form should check this anyway and it seem OK that the HEAD is used here.

## Screenshots
See issue

## Test methodology

Manual test:
- Select an an artificial command
- Ctrl-B/Ctrl-T to invoke the dialogs
- Verify that HEAD is set as base version

## Test environment(s)
----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1955,7 +1955,7 @@ namespace GitUI.CommandsDialogs
 
         private void CreateBranchToolStripMenuItemClick(object sender, EventArgs e)
         {
-            UICommands.StartCreateBranchDialog(this, RevisionGrid.GetSelectedRevisions().FirstOrDefault()?.ObjectId);
+            UICommands.StartCreateBranchDialog(this, RevisionGrid.LatestSelectedRevision?.ObjectId);
         }
 
         private void GitBashClick(object sender, EventArgs e)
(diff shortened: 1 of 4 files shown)


RussKie:
Thank you

RussKie:
Please apply the same to v3.1.1

PR #6634, opened 2019-05-22
Improve GitHub clone/fork form

### LINKED ISSUE #6404: GitHub Plugin: Allow sorting on the column headers
## Feature description

When retrieving the "My repositories"/"Search for repositories" list via `GitHub` -> `Fork/Clone repositories`, the order in which the repos are listed has always been unfathomable to me.

See the red box in the screenshot for an example.

So, I'd like to suggest the following:

#### Basic solution

Sort the list of repositories received from the GH API by name (case-insensitive) before displaying them.

#### More advanced solution

Allow the end-user to sort the list based on the columns being display by clicking on any of the column headers (blue box in the screenshot).

Similar to other typical sorting interfaces in Windows programs, I would imagine this would work like so:
* One click on a column header: sort ascending on that column.
* Second click on that same column header: reverse sort order (sort descending)
* Click on another column header: sort ascending on that column with any previous sorting as secondary criteria.

GE-GH-sorting

## Environment

- Git Extensions 3.0.2.5232
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 96dpi (no scaling)

pmiossec:
Fixes #6404

## Proposed changes

* Order results in the lists (My repositories & Search for repositories). Fixes #6404
* Improve column sizing and resizing
* Add option to easily call an upstream remote "upstream" (and update the label to be clearer)

### Before

image

### After

See:
* results sorted
* column width is better
* Editable dropdown to be able to select "upstream" + update on label
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 8cc5933320a245b19c4fc7ff5a2285d319ac0d9e
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
@@ -252,8 +252,8 @@ namespace GitUI.CommandsDialogs.RepoHosting
                     SubItems =
                     {
                         repo.Owner,
-                        repo.Forks.ToString(),
-                        repo.IsAFork ? _strYes.Text : _strNo.Text
+                        repo.IsAFork ? _strYes.Text : _strNo.Text,
+                        repo.Forks.ToString()
                     }
                 });
             }
--- a/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
@@ -8,6 +8,7 @@ using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
 using GitCommands.UserRepositoryHistory;
+using GitExtUtils.GitUI;
 using GitUI.UserControls;
 using GitUIPluginInterfaces.RepositoryHosts;
 using JetBrains.Annotations;
@@ -40,6 +41,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
 
         #endregion
 
+        private const string UpstreamRemoteName = "upstream";
         private readonly IRepositoryHostPlugin _gitHoster;
         private readonly EventHandler<GitModuleEventArgs> _gitModuleChanged;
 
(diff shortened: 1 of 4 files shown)


RussKie:
Would it be a big deal to center-align the 3 columns?

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
```diff
@@ -49,6 +51,16 @@ public ForkAndCloneForm(IRepositoryHostPlugin gitHoster, EventHandler<GitModuleE
             _gitHoster = gitHoster;
             InitializeComponent();
             InitializeComplete();
+
+            foreach (ColumnHeader column in myReposLV.Columns)
+            {
+                column.Width = (int)(column.Width * DpiUtil.ScaleX);
```
RussKie:
```suggestion
                column.Width = DpiUtil.Scale(column.Width);
```

pmiossec:
>Would it be a big deal to center-align the 3 columns?

No, that should not be a big deal. But don't you think that right aligned is better? Because center alignment for numbers is strange...

RussKie:
True, though for "Yes"/"No" center is probably better
Can we reorder "Private" "Is fork" "# Forks" (does it make sense to re-order)?

pmiossec:
>True, though for "Yes"/"No" center is probably better

Ok

> Can we reorder "Private" "Is fork" "# Forks" (does it make sense to re-order)?

No, they are not re-orderable. And, No. There is no added value to do it for something that is barely read...

RussKie:
> No, they are not re-orderable.

I meant in the designer, not at runtime.


> no added value to do it for something that is barely read...

👍 works for me

pmiossec:
I have made some more column adjustments:

* change column alignments (centered for yes/no and right for numbers)
* rename column headers
* change column order to have the same in the 2 lists
* change column header text to have the same in the 2 lists

Now, it looks like:
image


image

GitUI/Translation/English.xlf
```diff
@@ -1557,7 +1557,7 @@ You will have push access. {2}</source>
         <target />
       </trans-unit>
       <trans-unit id="label3.Text">
-        <source>Add remote as:</source>
```
gerhardol: Is this really the only translation related change?
I expected that the renamed controls and at least "Forks" -> "# Forks" to show here.

pmiossec: Thanks for this remarks that make me see that some columns were not translated.
I fixed it.

pmiossec:
Should I do something to move this PR forward?

gerhardol:
> Should I do something to move this PR forward?

Same as for #6417, you should be able to merge yourself: Approved (including RussKie) and you have now announced the intention that would allow someone late to the party to review. Merge tomorrow.

jrfnl:
@pmiossec Thanks for this. Changes look great! I look forward to seeing them in the next release!

Issue #4033, opened 2017-10-01
Enable "Show current directory changes in revision graph" by default


gerhardol:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The setting is disabled by default

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
This is a great feature, can be hard to find and should be enabled by default.
https://git-extensions-documentation.readthedocs.io/en/latest/settings.html#git-extensions-performance-working-dir-changes
Some discussions in #4031 
Also related to #564

The settings indicates that this is slow. I do not see this (limited repos so far), do not see any fundamental reasons to why this feature should slow down more than e.g. the file counter on the commit button.

**Environment you encounter the issue:**

 *  GitExtensions version: 2.50.02
 *  GIT version: 2.14.1
 *  OS version: Win10

**Did this work in previous version of GitExtensions (which)?**
Not to my knowledge

srudin:
I feel that was a very bad change. Took me quite a while to get rid of this confusing lines. I don't appreciate this being turned on by default. Just a feedback.

RussKie:
Thank you for your feedback.

Could you please describe what was confusing about the artificial commits?

osmanmeer:
I didn't like it being on by default either. For me it was just wasted space where I could see a couple commits. And I didn't know it was called 'artificial commit' and hence didn't know I could just turn it off.

mstv:
I use the artificial commits to get deltas with previous commits quite often.

It is already described in the label of the regarding checkbox:
![grafik](https://user-images.githubusercontent.com/36601201/57567516-b3e98080-73da-11e9-918c-f52931287f09.png)

pmiossec:
Perhaps we could add the setting in the "view" menu to enhance option visibility?

gerhardol:
> Perhaps we could add the setting in the "view" menu to enhance option visibility?

What I was thinking too. 
I have had good feedback for this change, it simplifies comparisions and show what is not committed.

PR #6638, opened 2019-05-22
Move ArtificialCommits from Settings to grid View menu

gerhardol:
https://github.com/gitextensions/gitextensions/issues/4033#issuecomment-491493535

## Proposed changes
Move the toggle to view artificial commits (working directory and Index commits) in the revision grid from Settings to the View menu.
This is consistent with other grid 'decorations'.
This setting was maybe affecting the performance when first added but has no direct effect right now. The General tab is also too crowded right now.

Note: The old setting was called "Show current working directory changes as an artificial commit".
I choose to call the new "Show artificial commits". As the setting now is closer to the change, it should be easier to understand. (This will have to be described in the documentation.)

### Before
image

image

### After
image

image

## Test methodology 
Manual test

## Test environment(s) 

----


gerhardol:
Most English.xlf changes due to changed Blame settings
(ignored to update the version number for plugins, no text change there)

PR #6645, opened 2019-05-23
refactor: GitModule.GetSortedRefs

### LINKED ISSUE #6616: Loading repository results in Exception: System.ArgumentException: An item with the same key has already been added.
## Current behaviour

When opening a repository "GitExtensions.exe Error" window is shown: "The application has crashed...."  Exception: System.ArgumentException: An item with the same key has already been added.

## Expected behaviour

Application doesn't crash.

## Steps to reproduce

It seems to be related to our product repository. Opening other opensource repositories don't result in the error. Hitting refresh doesn't trigger the error. The branch doesn't matter. The error occurs when loading the repository. Also when switching between them.

## Error Details
```
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at GitUI.CommitInfo.CommitInfo.<LoadSortedRefsAsync>g__ToDictionary|55_0(IReadOnlyList`1 list)
   at GitUI.CommitInfo.CommitInfo.<LoadSortedRefsAsync>d__55.MoveNext()
   at GitUI.CommitInfo.CommitInfo.<<RefreshSortedRefs>b__49_0>d.MoveNext()
```

## Environment
- Git Extensions 3.1.0.5877
- Git 2.18.0.windows.1
- .NET Framework 4.7.3062.0
- DPI 96dpi (no scaling)

RussKie:
## Proposed changes

`GitModule.GetSortedRefs` method has been moved into `CommitInfo` control because that's was its only call site. As such the method added little value and a significant overhead to the public API surface.

To get the tests working,
* `MockExecutable` was moved to `CommonTestUtils` projects to it can now be accessed across all tests; and
* `LoadSortedRefsAsync()` method was refactored and `ToDictionary` method was inlined into `GetSortedRefs()` to facilitate testing.

Also while building tests I discovered the following:
* `CommitInfo` would call `RefreshSortedRefs()` when the control was passed an instance of `UICommandsSource`, which was pretty much at the instantiation time, way before a display of a commit information was requested.
* A second redundant call to `RefreshSortedRefs()` was made when a repository was closed and a user navigated to the dashboard.
* `UpdateCommitMessage()` method could get called with a `null` revision, and that would cause a NRE.

Fixes #6616

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -98,7 +98,6 @@ namespace GitUI.CommitInfo
                 this.InvokeAsync(() =>
                 {
                     UICommandsSource.UICommandsChanged += UICommandsSource_UICommandsChanged;
-                    RefreshSortedRefs();
                     ReloadCommitInfo();
                 }).FileAndForget();
             };
@@ -139,6 +138,11 @@ namespace GitUI.CommitInfo
 
         private void UICommandsSource_UICommandsChanged(object sender, GitUICommandsChangedEventArgs e)
         {
+            if (!Module.IsValidGitWorkingDir())
+            {
+                return;
+            }
+
             RefreshSortedRefs();
         }
 
@@ -202,6 +206,34 @@ namespace GitUI.CommitInfo
             ReloadCommitInfo();
         }
 
+        private IDictionary<string, int> GetSortedRefs()
+        {
+            var args = new GitArgumentBuilder("for-each-ref")
+            {
+                "--sort=-committerdate",
+                "--sort=-taggerdate",
+                "--format=\"%(refname)\"",
+                "refs/"
+            };
+
+            string tree = Module.GitExecutable.GetOutput(args);
+            int warningPos = tree.IndexOf("warning:");
+            if (warningPos >= 0)
+            {
+                throw new RefsWarningException(tree.Substring(warningPos).SplitLines()[0]);
+            }
+
+            int i = 0;
+            var dict = new Dictionary<string, int>();
+            foreach (var entry in tree.Split('\n'))
+            {
+                dict.Add(entry, i);
+                i++;
+            }
+
+            return dict;
+        }
+
         private async Task LoadSortedRefsAsync()
         {
             ThreadHelper.AssertOnUIThread();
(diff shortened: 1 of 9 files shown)


GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -98,7 +98,6 @@ public CommitInfo()
                 this.InvokeAsync(() =>
                 {
                     UICommandsSource.UICommandsChanged += UICommandsSource_UICommandsChanged;
-                    RefreshSortedRefs();
```
mstv: Is necessary for "Contained in branches/tags". Restored in #6727.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -266,20 +287,23 @@ private void ReloadCommitInfo()
...
                     _revision.Body = data.Body;
                     _revision.HasNotes = true;
                 }
 
                 var commitMessage = _commitDataBodyRenderer.Render(data, showRevisionsAsLinks: CommandClickedEvent != null);
-
-                // workaround the problem that with some DPI RichTextBox size height
-                // passed to ContentResized event is ~1 line less than needed
-                rtbxCommitMessage.SetXHTMLText(commitMessage + Environment.NewLine);
+                rtbxCommitMessage.SetXHTMLText(commitMessage);
```
mstv: Are you sure, this will not cause a regression?

RussKie: I added this in 3.0 for 150% scaling.
This shouldn't cause any regressions

GitUI/CommitInfo/CommitInfo.cs
RussKie: Thank you, I couldn't quite work out what this was for...

GitUI/CommitInfo/CommitInfo.cs
mstv: Do I understand correctly that this workaround is not needed anymore?

RussKie: I don't think so. I don't believe we should have it.

I added it for 150% scaling because the textbox control felt squashed at the bottom. At 100% and 200% scaling the textbox has the correct margin.
However having run it for the past 4-5 months, I don't think it was a good addition.

Issue #6648, opened 2019-05-23
Jenkins: Unhandled malformed response


gerhardol:
## Current behaviour
If Jenkins has a malformed response, GE will not show any build info at all due to unhandled exceptions.

## Expected behaviour
GE shows the available build info.

## Steps to reproduce
No simple way to provoke the problem.

This seem to occur primarily where the Jenkins server responds with a lot of builds, several MB or so. A follow up should try to limit the data sent or at least limit the updates (hardcoded to every 10th second right now). 

## Screenshots
Just no build data.

## Did this work in previous version of GitExtensions
No

## Environment
- Git Extensions 3.1.0
- Build f11fad3e20bfc3bbd33377caa22f7ef44da0093f (Dirty)
- Git 2.21.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

gerhardol:
https://github.com/gitextensions/gitextensions/pull/6650#issuecomment-495404845

Tests should be added too.

PR #6650, opened 2019-05-23
Jenkins: Improved response checking for latest update

gerhardol:
Related to #6648, #6649 for release/3.1
----


RussKie:
Is it possible to add some sort of a unit test?

gerhardol:
Tests will require some restructuring first.
The problem is that the format is not defined, so there is no reference to test.
The return on investment is very low though.
Something else probable needs to be done to limit load on the Jenkins server, then this part may be removed.

Issue #6275, opened 2019-02-19
Application Unusable with High Contrast Theme


HK6:
## Current behaviour

Branch view does not use the system-specified text colour. This leads to Git Extensions being unusable when using a high contrast theme with a black background colour.

## Expected behaviour

Branch view should use the system-specified text colour.

## Steps to reproduce

1) Apply a high contrast theme in Windows with a black background.
2) Open Git Extensions.
3) Apply 'Colors' settings from screenshot below if required.

## Screenshots

![image](https://user-images.githubusercontent.com/547894/53018831-5a983380-3421-11e9-9d48-82e272e875df.png)

'Colors' Settings:
![image](https://user-images.githubusercontent.com/547894/53018856-697ee600-3421-11e9-88ab-fd85124c3c5a.png)

High Contrast Settings:
![image](https://user-images.githubusercontent.com/547894/53018877-77cd0200-3421-11e9-94c8-4502d259fde6.png)

## Did this work in previous version of GitExtensions

It "worked" in the first release of 3.0 but that was only because the background colour did not use the system-specified background colour.

## Environment

- Git Extensions 3.0.2.5232
- Build 23b6f51905006ccdde8701591df706284b4155dc
- Git 2.18.0.windows.1
- Microsoft Windows NT 10.0.17134.0 (Windows 10 Enterprise version 1803 build 17134.523)
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

RussKie:
Relates to #5489

gerhardol:
Closed by #6651

Issue #1947, opened 2013-07-02
Unreadable texts in dark theme


matthew11:
![gitext](https://f.cloud.github.com/assets/4918748/735228/fed6dfb8-e2ed-11e2-87f1-adcd54149506.png)
--- GitExtensions2.44, 1.8.1.2, Linux 3.8.0.25 with XFCE4 and mono 2.10.8.1

See the attachment.

aaronbond:
i also get this issue, any chance it will eventually be fixed?

mica100:
still a problem in the history log background. Is there a chance to set the background color?

Text is white but background too, even if set to something dark in Windows settings.

Menus, also right mouse menu, lower part of browse main window (with commit details) and commit dialog take the windows colors. There is no problem with this anymore.

Only the background in the history log window.

aaronbond:
this project has 636 issues :( not much chance this one would be a priority, i've just had to move to a light theme because i really do use gitextensions a lot, it's a nice ui to have when i need one on top of git.

jbialobr:
Only partially fixed. (Closed automagically by github).

safield:
I am having this issue on Linux Mint 18.2, using Mint-y dark theme.

The proper solution is to have git ext adopt the dark theme color for its windows, or a less desirable fix would be to NOT adopt the dark themes light text.

The problem arises because the font is changing to the dark themes light font, but the background is not changing to toe dark themes dark themes dark background. This results on light text on light background.

RussKie:
Is the latest version still the same bad as on the screenshot above?

There are a number of issues with dark theme support, even on Windows.
Sadly for you and other Mono users Linux support isn't high on our todo list - no one from the core team uses Linux/Mono full time.
PRs are always welcome though.

safield:
In 2.50.02 version the issue is just as bad as the first screenshot provided.

gerhardol:
Closed by #6651

PR #6651, opened 2019-05-23
Dark theme fixes

cryham:
Related issues:
#5489 Dark Theme, #6275 High Contrast, #1947

## Fixes

Here are some (most) fixes I did to make GitExt usable in my black-violet theme.
GitExt is not black / white theme aware, as it uses some fixed colors set in code. Even worse, some are just for background without matching text set too.
Either way I think I fixed most of the stuff I saw (in Browse and Commit dialogs).

## Proposed changes

There is still a remaining issue with text / code editor line numbers background.
I just hid it in my [fork here](https://github.com/cryham/gitextensions/commit/f29130419f11fcb025e5cc8d2c4667def6205c0f), for now (in internal viewer only).

Second thing left, is that changing themes black / white doesn't change GitExt colors in setting (Appearance). I think it'd  be great to have 2 columns with colors (twice as many): one for white, other for dark themes.

### Before

Before

### After

After

## Test environment(s) 

- GIT version 2.21
- Windows 7 SP1

----


gerhardol:
Can you please open a separate PR to sign contributors.txt?

RussKie:
Sexy!
We'll need to test in the light mode, but it is an awesome start!

gerhardol:
The dark mode definitely look better now...

Stupid question: How do I activate the dark mode?
Win10 1809 with Dark theme by default.

The light theme is changed, see the screenshot. (Not considered what I prefer.)

image

cryham:
Right.
So the dark theme is detected and set automatically in code when system theme color luminosity is 0.5 or below.
Not sure about Win 10, I don't have access to it yet.
In Win 7 I test by switching my theme 
I signed contributors.txt in #6659

gerhardol:
Win10 dark mode is not activating, will test with other themes

It would be good if the mode could be overridden too, i.e. default/light/dark.
(if one column is kept, this could control what is updated too).

I would be surprised if this PR is the complete solution for a dark theme, but it for sure is a great improvement.

gerhardol:
First, I am not a 'themer' myself. My wife does not like bright screens though...
This is something others have asked for, so it is a nice addition to GE.

Some background regarding Win10 dark theme:
Dark mode in Win10 only affect UWP apps, no simple way to find if this setting in other apps.
You can use HighContrast-Dark, that is very ugly...
In Win7 it was easier to apply custom themes. In Win10, this is harder, something like WindowsBlinds can be used:
https://www.howtogeek.com/222614/how-to-enable-windows-10s-hidden-dark-theme/

Using system colors consistyenly changes the normal theme too. Thisreverts some of the careful changes that @drewnoakes applied but I am mostly fine with that. Tweaks will have to be done.

This is what GE looks with the DarkMode skin.
Some items are not applied for me.

image

RussKie:
Thank you Gerhard, that doesn't look too bad - looks like few more tweaks are required

cryham:
Fixed line numbers on Diff tab. They go dark with theme.
I also changed graph lane colors somewhat. There was one too dark color. I generally like them more now.

File Tree tab is a different story. For dark theme only some file types are dark (e.g. txt, md, gitignore). All language files have files here \Externals\ICSharpCode.TextEditor\Project\Resources\C#.xshd. Lots of them but the top section <Environment> seems to look almost identical across them. And yeah <Default color="Black" bgcolor="#FFFFFF"/> is white background (bad for eyes in dark theme.
Even worse, all those files go into ICSharpCode.TextEditor.dll as resources.

How would you approach it? There are themes for Console, I guess it'd be great to have them for File Tree viewer too. But this seems like a big task to me.

gerhardol:
> Should there be 2 sets, i.e. a 2nd column for all colors for dark theme?

With the approach in this PR, with SystemColors and special handling depending on light/dark only, I would first like to have "hardcoded" translation. If this is futile, I prefer a second set of colors to avoid a very manual setup.

The discussions in #5489 was about creating theme_s_ by a proxy. 
This PR uses the light variant by choosing to use SystemColors where possible and choosing a light/dark variant where needed. There are quite many other forms that will need adjustments still with hardcoded colors. With certain dark themes, GE may even look nice. Use of SystemColors only also avoids the issue where some controls uses SystemColors internally.

What do the prominent WinForm developers think?
@NikolayXHD @persn @digitalcoyote 

> Even worse, all those files go into ICSharpCode.TextEditor.dll as resources.

GE uses the text editor in the submodule, it can be changed.

Some issues I noted when using Win10 dark theme. All issues in all windows themes will not be solved now....

Browse
Icon not seen
image

image

Commit
image

image

About
Maybe link color in theme?
image

GitHub
image
image

Other
image

image

RussKie:
we will need to have a second set of icons for the dark theme, let's leave
it out for now

cryham:
Icons that are black (or almost) and aren't visible are (inside GitUI\Resources\Icons):

Delete.png
Develop.png
FileStatusRenamed.png
github.png
GitHub32.png
Link.png
PullRequest.png
RecoverLostObjects.png
ShowWhitespace.png
Translate.png
WhitespaceIgnore.png
WhitespaceIgnoreAll.png
WhitespaceIgnoreEol.png

I could probably make a light gray version of them and if we can switch in code at init between them that would work right?

cryham:
OK I think the icons switch are done now.
I didn't know how to check few (e.g. pull request, delete is red?). I left github as it is.

gerhardol:
GE looks mostly OK in Win7 with your theme.
Not much worse than the diff margin (as described above), the default diff, hlight colors and the following from Apply Patch (random check though):
image

However in Win10 with the dark mode I test with, there are some artifacts (mostly black text or links on dark/black background). Most of that are due to hardcoded colors, like Color.Black.
Do you want a PR for that (to your branch), me to push to your branch directly or fix it yourself?

When it is OK for me I will try to make a better review (I am away this weekend) (do not expect many comments blocking merge for me, but more about improvements). 
More users must test/review before merge though.

cryham:
So I think a PR to my branch then (if not too much trouble).
I think I searched for Black colors set in code but likely still didn't catch them all.
I don't have Win 10 yet and don't know GUI that well to find / show them.

Apropos this previous screen with disabled buttons on Apply Patch etc. Is there another way? I'm used to disabled things looking like that. I.e. barely visible or invisible in menus of other programs.

Just for info here, there are like 2 more details I see (that I don't know how or if possible to fix):
Here's a screen with them:
screen
1. Checkboxes (in menus) have black tick, which makes them invisible for me, only the rectangle around is slightly visible but not much (I think it's system selected color).
2. Some icons have still few pixels almost white on edges (around icon). E.g. Commit, GitExt in settings tree, Fonts etc. But I think this would be a lot of work to fix those pixels in icons, maybe easier to make new. Probably not worth it.

gerhardol:
For most of the black text on black I see: That seem to be the theme, for inactive buttons.
image

There are some hardcoded use of Color.Black, but many of those are in ICSharpCode.TextEditor, that is a separate repo.

The links are the most difficult to read now. (other than disabled)

For the background for line number in diff: Should that really be black and not the same as the file viewer?

image

Otherwise most artifacts has been fixed. No PR from me today at least, probably few changes if any at all tomorrow.

todo?
AsIs, Special handling (hardcoded) or double configuration for special colors like diff?
Fix TextEditor now too?
Replace SystemColors with "proxy" to prepare for theming within GE?

RussKie:
Is it possible to make text lighter? Or is it a standard font color in the theme?

> AsIs, Special handling (hardcoded) or double configuration for special colors like diff?
> Fix TextEditor now too?
> Replace SystemColors with "proxy" to prepare for theming within GE?

We can open separate issue for these.
I don't expect _all_ things to be addressed in one go, I fully appreciate amount of effort required.

gerhardol:
> Is it possible to make text lighter? Or is it a standard font color in the theme?

Set in the Windows theme, so a bad theme will give a bad experience.
This a reason to have a "proxy"/theme engine to override certain colors, not just the colors that clash with the theme (like the diff colors in my example). That could be done separately). Reason to do it here is to reduce noise, to avoid 

This PR makes GE usable in dark theme and high contrast, more changes to make the app great indark themes.

gerhardol:
For links: The colors are fixed in WinForms (W3C standard, derived from InternetExplorer).
Most forms use the default which is hard to read with dark themes.
A few forms (like Settings/Appearance) use SystemColors.HotTrack instead.

This PR makes a special change in About Form. That improves, but hard to do for all links.
Should maybe be reverted.

A solution is to set colors for all LinkLabel, another to use a custom control.

gerhardol:
For the line numbers background, the following looks better for me.

GE3.1

image


This PR, Win10 default and the theme I test with

image
image

Win10 patch
image
image

cryham:
OK. I think I'll do the 2nd column for Diff colors next. Will be convenient when switching system theme, if they change automatically.
Only thing with line numbers background I'm worried about is finding good colors, for text too. So they don't get too much or too little contrast on various black/gray/white themes.

RussKie:
May be look at how our competitors do it.
GitKraken and Fork both have dark colored themes.
I can post few screenshots from Fork, if you want.

gerhardol:
> May be look at how our competitors do it.

This PR makes GE from being horrible to play nicely with the Windows theme, using SystemColors where appropriate and a adjusting a few situations to dark/light where GE mselects colors. It does not handle:
 * GE default colors (like diff colors that do not match the theme) 
  -> could have separate light/dark colors here, but await theming first
 * Theme colors clashing with the theme (like links dark blue)
  -> also a general question that will require override of the bad controls
 * Theme bad setup (like black text on dark background like disabled controls)
  -> Not much to do about
 * Some corner cases where bad hardcoded colors or SystemColors use give a bad layout. The line number background with the theme I test with is one example.
  ->Depends on the theme, will have to fixed where found

Most of the above should not be handled in this PR.

To have built-in themes will require a "theme-engine". That is a lot more work as it will require to proxy SystemColors and override controls that use SystemColors. This PR is a step to that as it identifies where colors are set.

GitUI/CommandsDialogs/FormAbout.cs
```diff
@@ -21,6 +22,11 @@ public FormAbout()
 
             environmentInfo.SetCopyButtonTooltip(_copyTooltip.Text);
 
+            Color clrLink = ColorHelper.IsLightTheme() ? Color.Blue : Color.SkyBlue;
```
gerhardol: While the links are hard to read in dark themes, this is a general problem and this increases inconsistency.
Some links use SystemColors.HotTrack instead (see for instance CommitInfo. Looks a little worse.
Even if I commented on this earlier, I propose to revert this change, at least for now.

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -60,7 +60,10 @@ public FileViewerInternal()
                         TextEditor.ActiveTextAreaControl.TextArea.TextView.GetLogicalLine(e.Y)));
             };
 
-            HighlightingManager.Manager.DefaultHighlighting.SetColorFor("LineNumbers", new HighlightColor(Color.FromArgb(80, 0, 0, 0), Color.White, false, false));
```
gerhardol: The background looks like a visual artifact in the theme I test with and I suggest changing this (maybe as in https://github.com/gitextensions/gitextensions/pull/6651#issuecomment-497133999) also if GE normal theme changes.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphLaneColor.cs
```diff
@@ -8,15 +8,15 @@ namespace GitUI.UserControls.RevisionGrid.Graph
     public static class RevisionGraphLaneColor
     {
         internal static readonly IReadOnlyList<Color> PresetGraphColors = new[]
-{
```
gerhardol: Changed lane colors, OK for me

GitUI/CommandsDialogs/FormAbout.cs
cryham: Why not use HotTrack in About too then? It is skyblue on my theme and looks good.
I changed to blue it in next commits (they aren't that visible now).

GitUI/Editor/FileViewerInternal.cs
cryham: Right. I changed lines to just use System Control and Text colors. Looks good on my black/gray/white themes.

GitUI/CommandsDialogs/FormAbout.cs
gerhardol: We could change all to HotTrack, but there are quite some LinkLabel uses.
It is maybe better to use a custom control to set also visited links.
The color should probably be unset, to use the defaults.

RussKie:
I'm too busy at the moment with the new job, leaving it with you.
I'm hoping Crystal isn't going to disappear immediately after the merge.

RussKie:
Totally agree. I don't expect this PR will solve all our theming problems.
It is a huge effort as such, and I massively appreciate it.

I was pointing that we can look up the colour selection our
more-theme-advanced competitors have made and take an inspiration from that.

gerhardol:
I plan to merge this in the next days unless there are protests. If there are comments on the (minor) visual changes in normal themes, we handle them later.

@cryham 
Can you please rebase on gitextensions/master?

cryham:
I think I did just now, but that was my first time, so I'm not sure if this went bad or not.

pmiossec:
No, that's not good :disappointed:
I'm not in front of a computer so I can't really diagnose what you did but at least at the you did a merge.

What @gerhardol was expecting is that you rebase all your commits onto 'upstream/master'.

So, if you did it well, you should have seen **only** your commits in the 'commits' tabs of this PR in github.

cryham:
OK now it is good?

pmiossec:
I don't know if it contains all your changes in this commit, but in terms of git history, yes, it seems OK.

I think @gerhardol will do a last quick check but seems :+1:

gerhardol:
@cryham 
Thanks for the contribution

Your rebase had the same result as mine, so it should be correct...
We normally prefer rebase to avoid merge conflicts (better done by the PR submitter) and to get a cleaner history. I hope you learnt something!

If you are interested in improving the dark theme further, I suggest you open a new feature.

cryham:
Great. Yes, I did.

RussKie:
Thank you :+1:

hinstw:
I saw this issue in change-log for version 3.2.0 but could not find a way to change UI theme. Can someone please tell me how to change theme, or is it not possible yet? (i assumed that it should be somehow possible because the issue title says "fixes" so it implies that the dark theme exists)

NikolayXHD:
The fixes here make GitExtension look better with  Windows dark themes.

gerhardol:
@hinstw 
Some hints to use a System UI dark theme in Win10
https://github.com/gitextensions/gitextensions/pull/6651#issuecomment-495798173

RussKie:
We will porting the app to .NET Core, which is also open sourced.
So theming related fixes can be sent that way 😉

Issue #6542, opened 2019-05-08
Roadmap: move plugins


mast-eu:
## Scope

Move plugins from Git Extensions main repo into their own repos (https://github.com/gitextensions/gitextensions/issues/5574).

## Step 1) Preparation

Mandatory things that need to be clarified or done before we can actually start relocating the plugins.

- [ ] Create NuGet package `GitExtensions.Plugins` containing GE's plugin API (https://github.com/maraf/GitExtensions.PluginManager/issues/67, https://github.com/maraf/GitExtensions.PluginManager/issues/50).
❗️ All plugins must include this as dependency and link against it.
- [ ] Integrate PluginManager (https://github.com/maraf/GitExtensions.PluginManager)
  - [x] merge code (https://github.com/gitextensions/gitextensions/pull/5775 / https://github.com/gitextensions/gitextensions/issues/4885)
  - [ ] Plugin installation path (https://github.com/gitextensions/gitextensions/issues/6415)
  - [ ] ship PluginManager as part of a release
- [ ] Test deployment, installation, execution, update and removal of a simple plugin through PluginManager.
 💡  I'm currently working on https://github.com/mast-eu/GitExtensions.SVN, which will probably be the first (public) NuGet-based plugin and should be a good candidate for these tests.
- [ ] ❓ Who will be the owner of the plugins' repos and the publisher of the NuGet packages?
- [ ] ❓ Has the plugin relocation any impact on the GE relicensing process (https://github.com/gitextensions/gitextensions/issues/4737)?

## Step 2.a) Move plugins out

Start with the "easier" plugins, i.e. those who do not have specific settings (less contact points with GE).
- [ ] Create local tracking branches
- [ ] Gerrit Code Review
- [ ] GitFlow
- [ ] Impact Graph
- [ ] Release Notes Generator

Move the remaining plugins in a second round (maybe tackling https://github.com/gitextensions/gitextensions/issues/5797 as well).
- [ ] Auto compile SubModules
- [ ] Bitbucket Server
- [ ] Delete obsolete branches
- [ ] Find large files
- [ ] GitHub
- [ ] Gource
- [ ] Jira Commit Hint
- [ ] Periodic background fetch
- [ ] Proxy switcher
- [ ] Statistics

#### For each plugin do:
1. Extract current plugin code "as is" from GE, wrap it as NuGet package and push to it's own repo. Do not remove it from the main repo yet.
[...]

PR #6664, opened 2019-05-24
Reintegration of PluginManager.

maraf:
This PR reintegrates PluginManager to the GitExtensions installer and portable zip.

To make PM self updateable I'm moving it to user plugins folder. 

Here I'm facing a problem with WIX. 

```
error LGHT0204: ICE64: The directory LocalAppDataOrganizationDir is in the user profile but is not listed in the RemoveFile table.
```

All folders in user profile must be uninstalled, so I have defined their removal. But on of the folders is generated by `<HeatDirectory />` build task and has generated id (Setup/Product.wxs#L75). Which may change unexpectedly..

The reason why PM WIX component is generated is that its file structure may change in the future, so I tried to make the installer universal. 

Does anyone of you guys know how to make `<HeatDirectory />` generate uninstall rules?
The only solution that came to my mind is replacement of HeatDirectory with powershell script.


Setup/Product.wxs
```diff
@@ -64,6 +64,20 @@
...
       </Directory>
+      <Directory Id="LocalAppDataFolder">
+        <Directory Id="LocalAppDataOrganizationDir" Name="GitExtensions">
+          <Directory Id="LocalAppDataApplicationDir" Name="GitExtensions">
+            <Directory Id="UserPluginsDir" Name="UserPlugins">
+              <Component Id="UserPlugins" Guid="{81556E1B-270C-448D-A68D-FD30678D4827}">
+                <RemoveFolder Id="RemoveLocalAppDataOrganizationDir" Directory="LocalAppDataOrganizationDir" On="uninstall" />
+                <RemoveFolder Id="RemoveLocalAppDataApplicationDir" Directory="LocalAppDataApplicationDir" On="uninstall" />
+                <RemoveFolder Id="RemoveUserPluginsDir" Directory="UserPluginsDir" On="uninstall" />
+                <RemoveFolder Id="Removedir7351E76F113B364403061A39D2447F8D" Directory="dir7351E76F113B364403061A39D2447F8D" On="uninstall" />
```
maraf: This directory is generated by `<HeatDirectory />` but is required to be listed for removal.

Setup/Prepare-Release.ps1
```diff
@@ -147,10 +147,10 @@ try {
...
     Generate-Changelog -milestones $milestoneNumbers
 
-    #Write-Host ----------------------------------------------------------------------
-    #Write-Host Download PluginManager
-    #Write-Host ----------------------------------------------------------------------
-    #.\Download-PluginManager.ps1 -ExtractRootPath '..\Plugins\GitExtensions.PluginManager'
+    Write-Host ----------------------------------------------------------------------
+    Write-Host Download PluginManager
+    Write-Host ----------------------------------------------------------------------
+    .\Download-PluginManager.ps1 -ExtractRootPath '..\Plugins\GitExtensions.PluginManager'
```
mast-eu: https://github.com/gitextensions/gitextensions/blob/775d297e583625ea1098eed6e7b58f0f78d2e403/Setup/Download-PluginManager.ps1#L10
`Download-PluginManager.ps1` still is configured to download from `maraf/GitExtensions.PluginManager`. This is no longer in line with the new approach for the PluginManager and we would skip a few important steps.

I think we should move in order:
- first set up `gitextensions/GitExtensions.PluginManager`
- review the code (at least I for myself never really looked into it)
- build a release from the reviewed code under `gitextensions/GitExtensions.PluginManager`
- include it in the installer

RussKie: We'll need to have it signed too

RussKie:
@Johno-ACSLive any thoughts?

Johno-ACSLive:
> @Johno-ACSLive any thoughts?

I've had a quick look at this PR, roadmap item #6542 etc. If I understand this correctly, you wish to use a plugin manager for GE and the plugin manager will pull plugins using NuGet.

This PR is to remove plugins from the installer / portable and only include the plugin manager. There's also changes between system install for all users vs per user installation of plugins.

I saw this piece:

> I think we should move in order:
> 
> - first set up gitextensions/GitExtensions.PluginManager
> - review the code (at least I for myself never really looked into it)
> - build a release from the reviewed code under gitextensions/GitExtensions.PluginManager
> - include it in the installer

I agree in principle with the above approach.

From a distribution mechanism, the installer / portable just needs to know where the plugins go and if they are per user or per machine based once we know the details of how the plugin managger will work etc.

What is the decision around plugins, will they be installed for all users, per user or configurable for the installer?

For portable, it doesn't make sense to have these options as it should be on a per instance (folder copy) basis. Portables are not expected to be copied to sensitive locations and so should have full access to it's local structure.

For the installer this could be different based on the per machine / per user support, but it sounds like we simply need to create the appropriate directory based on a per machine or per user install. Uninstall just needs to know about the root Plugins folder for clean-up, we don't care what is below as we can tell the installer to recursively delete the Plugins directory.

Do we want similar capability to VS where users can "Roam" their plugins? I know this question is out of scope of this PR and should be covered under the roadmap. If the Plugin Manager does some smart things, then the installer and portable will need to know how to handle these things.

I may have gone into too much detail here 😄 not sure if this helps at all?

mast-eu:
>  If I understand this correctly, you wish to use a plugin manager for GE and the plugin manager will pull plugins using NuGet.

Yes, you captured it nicely.

> This PR is to remove plugins from the installer / portable and only include the plugin manager.

Not exactly. This PR will _add_ the PluginManager, but not yet _remove_ the plugins. They will be removed later, after they have been moved - one by one - into own repos and made available as NuGet packages.

> What is the decision around plugins, will they be installed for all users, per user or configurable for the installer?

> Do we want similar capability to VS where users can "Roam" their plugins?

There is a related discussion in https://github.com/gitextensions/gitextensions/issues/6415 / https://github.com/gitextensions/gitextensions/pull/6551. The conclusion was to go for plugins on a per user basis and without roaming.

> Uninstall just needs to know about the root Plugins folder for clean-up, we don't care what is below as we can tell the installer to recursively delete the Plugins directory.

This is mostly correct. Deleting the plugins directory is sufficient to remove the plugin itself, but not it's settings.
The caveat is that all current plugins are saving their settings to GE's default settings file (`C:\Users\<UserName>\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.settings` for the installed version and `<PortableRootDir>\GitExtensions.settings` for the portable version). See the related discussion in https://github.com/gitextensions/gitextensions.extensibility/issues/4.
Therefore, if we simply delete the plugin directory, these plugin-specific settings would remain in `GitExtensions.settings`. This would not create any immediate trouble (they will be simply "dead weight"), but it would clutter the settings file over time if we don't take care of removing them.

gerhardol:
> The caveat is that all current plugins are saving their settings to GE's default settings file

As well as possibly in settings per repo

Johno-ACSLive:
OK understand now, I agree that the plugins themselves shouldn't roam (i.e. the binaries should be stored in %LOCALAPPDATA%. I think that roaming which plugin a user has enabled and potentially it's settings could be stored in %APPDATA%.

What this means is that the plugin manager can recognise that a plugin has been downloaded and installed on device A and plugin manager can then see that has occured and download the associated plugin on device B (irrespective of GE / Plugin version, it just cares about the plugin being installed). This assumes the version of GE supports the plugin being migrated to PM otherwise it will have no effect.

For individual plugins, it would depend on how breaking changes implement their configuration handling, perhaps initially %LOCALAPPDATA% is used and as the plugins mature their configuration management they can support %APPDATA% for roaming.

From an installer perspective, cleaning up configuration from main GE settings and having a config file per plugin would depend on the migration strategy to the PM. If we move a plugin from Legacy to PM, something needs to migrate the data.

So lets say we upgrade GE to a version where plugin A is removed from GE and is now available from PM. The installer will have the payload removed (the binaries will no longer be included and during upgrade, the installer will remove the previous version which will remove the exisiting binaries) but something in GE / PM will need to detect if the user was using the Plugin, install the package via PM and migrate the settings from GE main config to plugin config.

Is this expected to be the installer or GE / PM / Plugin? I believe that GE / PM / Plugin should be responsible to perform this action as I am assuming that a plugin via PM will be self contained (it will contain base config of the plugin). I have looked at the TODO's and there are still some decisions to be made.

However, what we can do in the installer is launch an executable (or script) with a flag to indicate a migration needs to be performed. The installer UI can indicate it is migrating plugins while the custom action is occuring. The executable or script (especially a PS script) will also need to be signed.

Alternatively, the first load of GE can run through a Plugin migration wizard and we could detect what plugins we think a user is using and have them select which ones to preserve.

Again, depends on the migration strategy for the plugins.

In the case where GE is being uninstalled and not upgraded, we would remove all directories associated with GE. So if we create two user profile locations e.g. %LOCALAPPDATA%\gitextensions and %APPDATA%\gitextensions then uninstall just deletes those to ensure a clean removal. For upgrades these directories will of course be preserved (the installer just needs to ensure this is the case).

Is the PM ready to have its binaries compiled / signed with GE ready for inclusion in the installer? At the moment I see a PS script to download the binaries to sub folder of the Plugins directory during installation. Ideally we would bundle the PM binaries in the installer.

mast-eu:
Actually I am thinking of a much simpler approach: leaving it entirely to users what plugins they want to install. Detecting which plugins a user might have used in the past would be very tricky and error-prone. For example, I often _look_ at the results from AppVeyor (mostly the green and red dots in the revision graph) and occasionally _open_ the Statistics. Both are plugins and for both I didn't even change the default settings. I cannot think of a way to reliably detect this kind of usage.

Also if you use multiple devices, I would require you install the plugins multiple times. At least for the first release, since I think the majority of our users does not require roaming of plugins (or of the metadata of installed plugins). It would be nice to have, but it is not a mandatory feature.

> Is the PM ready to have its binaries compiled / signed with GE ready for inclusion in the installer?

Yes, almost. The PM code can be found in [gitextensions/gitextensions.pluginmanager](https://github.com/gitextensions/gitextensions.pluginmanager) and the artifact can be [downloaded from AppVeyor](https://ci.appveyor.com/api/buildjobs/bh9cxeidgtsefj65/artifacts/GitExtensions.PluginManager.0.9.0-build0004%2Bbe4f782927.zip). It is not yet signed (see https://github.com/gitextensions/gitextensions.pluginmanager/issues/17).

> At the moment I see a PS script to download the binaries to sub folder of the Plugins directory during installation. Ideally we would bundle the PM binaries in the installer.

:thinking: The script is downloading the PM while assembling the installer, so that it is already bundled with the MSI. There shouldn't be a separate download during installation on the target PC.

RussKie:
The 3.2 release is imminent, so if you want to include the PM in the distribution - it is a good time to bring it back now.

maraf:
In my option only `GitExtensions.Extensibility` nuget package is a must have from our todo list.
All other things are "nice to have".

RussKie:
Let's aim for the following release then

Setup/Product.wxs
mast-eu: Are there updates regarding this point, or is it still pending?

maraf: I still don't know how to solve it.

Johno-ACSLive: > All folders in user profile must be uninstalled, so I have defined their removal. But on of the folders is generated by <HeatDirectory /> build task and has generated id (Setup/Product.wxs#L75). Which may change unexpectedly..

> The reason why PM WIX component is generated is that its file structure may change in the future, so I tried to make the installer universal.

Top level directory created / deleted by installer should never change, PM should only operate under the top level. The installer can then recursively delete anything underneath during uninstall. Does the PM currently work this way?

maraf: Yes, PM is placed under top level directory, but when I don't explicitly add `<RemoveFolder` for all PM stuff, compiler throws errors.

```
error LGHT0204: ICE64: The directory LocalAppDataOrganizationDir is in the user profile but is not listed in the RemoveFile table.
```

Setup/PluginManager.wxs
```diff
@@ -1,29 +1,21 @@
...
-                    <File Id="fil04BF9952F70979845606C493533B8068" KeyPath="yes" Source="$(var.PluginManagerSourceDir)\PackageManager\PackageManager.UI.exe" />
+            <Directory Id="dir7351E76F113B364403061A39D2447F8D" Name="PackageManager">
+                <Component Id="cmp143681A5E573D00A5CF51D86FF5F085C" Guid="*">
+                    <File Id="fil8EFAEC01EDFFA6AA045ADA67A54446CB" KeyPath="yes" Source="$(var.PluginManagerSourceDir)\PackageManager\PackageManager.UI.exe" />
                 </Component>
             </Directory>
         </DirectoryRef>
     </Fragment>
     <Fragment>
         <ComponentGroup Id="PluginMananger">
```
mast-eu:
```suggestion
        <ComponentGroup Id="PluginManager">
```

Setup/Product.wxs
```diff
@@ -1676,9 +1690,8 @@
           <ComponentRef Id="NString.dll" />
           <ComponentRef Id="System.Collections.Immutable.dll" />
         </Feature>
-        <!--
+        <ComponentRef Id="UserPlugins" />
         <ComponentGroupRef Id="PluginMananger" />
```
mast-eu:
```suggestion
        <ComponentGroupRef Id="PluginManager" />
```

Setup/Setup.wixproj
```diff
@@ -99,12 +94,10 @@
 	call BuildGitExtNative.cmd $(Configuration) build">
     </Exec>
   </Target>
-  <!--
   <Target Name="HeatPluginMananger" BeforeTargets="BeforeBuild">
```
mast-eu:
```suggestion
  <Target Name="HeatPluginManager" BeforeTargets="BeforeBuild">
```

Setup/Setup.wixproj
```diff
@@ -99,12 +94,10 @@
...
-    <HeatDirectory Directory="..\Plugins\GitExtensions.PluginManager\Output" OutputFile="PluginManager.wxs" ComponentGroupName="PluginMananger" DirectoryRefId="PluginsDir" PreprocessorVariable="var.PluginManagerSourceDir" AutogenerateGuids="true" ToolPath="$(WixToolPath)" SuppressFragments="true" SuppressRegistry="true" SuppressRootDirectory="true" />
+    <HeatDirectory Directory="..\Plugins\GitExtensions.PluginManager\Output" OutputFile="PluginManager.wxs" ComponentGroupName="PluginMananger" DirectoryRefId="UserPluginsDir" PreprocessorVariable="var.PluginManagerSourceDir" AutogenerateGuids="true" ToolPath="$(WixToolPath)" SuppressFragments="true" SuppressRegistry="true" SuppressRootDirectory="true" />
```
mast-eu:
```suggestion
    <HeatDirectory Directory="..\Plugins\GitExtensions.PluginManager\Output" OutputFile="PluginManager.wxs" ComponentGroupName="PluginManager" DirectoryRefId="UserPluginsDir" PreprocessorVariable="var.PluginManagerSourceDir" AutogenerateGuids="true" ToolPath="$(WixToolPath)" SuppressFragments="true" SuppressRegistry="true" SuppressRootDirectory="true" />
```

Setup/Setup.wixproj
```diff
@@ -99,12 +94,10 @@
...
   </Target>
   <Target Name="DownloadPluginMananger" BeforeTargets="HeatPluginMananger">
```
mast-eu:
```suggestion
  <Target Name="DownloadPluginManager" BeforeTargets="HeatPluginManager">
```

mast-eu:
> All folders in user profile must be uninstalled, so I have defined their removal. But on of the folders is generated by <HeatDirectory /> build task and has generated id (Setup/Product.wxs#L75). Which may change unexpectedly..

I deleted the file `Setup/PluginManager.wxs` locally on my system and recreated it from scratch by running `Setup\BuildInstallers.cmd`.
It is regenerated identically to your commited version, with the same ID in https://github.com/gitextensions/gitextensions/blob/7a994d7ef01235fcd5f6a0595c9eb0adf9b177ae/Setup/PluginManager.wxs#L8

Thus I don't think this ID can change _unexpectedly_, as long as the path and filenames do not change.

> The reason why PM WIX component is generated is that its file structure may change in the future, so I tried to make the installer universal.

Sure, the file structure of the PM artifact might change in the future, but in this case I think it would be acceptable that the installer needs to be updated, too. Furthermore, I don't think that changes will be frequent.
In other words, I doubt that we need an universal installer. It would be _nice to have_, but it is _not mandatory_. Let's remove the HeatDirectory task if it is creating trouble.
IMHO, it is acceptable for the installer to work only with the precise locations and filenames `$(var.PluginManagerSourceDir)\GitExtensions.PluginManager.dll` and `$(var.PluginManagerSourceDir)\PackageManager\PackageManager.UI.exe`.

maraf:
That makes things a lost simpler, the heat stuff was very painful. I'm going to update the PR.

maraf:
The download URL still points to the original repository, as the official repository doesn't any release yet.

mast-eu:
> The PM download URL still points to the original repository, as the official repository doesn't any release yet.

I just created pre-release `v1.0.0-RC1` in the official repo.
Please update the URL to `https://github.com/gitextensions/gitextensions.pluginmanager/releases`.

maraf:
I have updated the URL to the gitextensions/gitextensions.pluginmanager repository.

Setup/Product.wxs
mast-eu: Has become obsolete.

mast-eu:
Any idea why the CI build is failing? I didn't look into details yet...

Could you please rebase on current `master` branch and fix the [remaining typo](https://github.com/gitextensions/gitextensions/pull/6664#pullrequestreview-271624851)?

maraf:
@mast-eu Oh, sorry. I completely missed that. Do it now..

RussKie:
Release 3.2 is technically ready, I'm happy to wait a little longer, if you think this PR is almost ready.

maraf:
@RussKie There is some problem with PM download on AppVeyor. Locally it works, but on AppVeyor it doesn't report error with download, still the files are not at destination. 

Later today I'm going to investigate...

maraf:
@RussKie @mast-eu I have missed that portable build nolonger uses `Setup\Prepare-Release.ps1`, but instead `Build\Prepare-Artifacts.ps1`. PR now builds and final zip contains working PM.

mast-eu:
Did you also check the installer build?

maraf:
@mast-eu I'm not sure why, but it's not on AppVeyor, only portable zip.

mast-eu:
As far as I know (@RussKie may correct me) the installer is usually not build for the `master` branch. It is build only  on `release\*` branches, those have a different appveyor.yml file.

RussKie:
It is built but we do not publish it.
We publish msi for releases only.

To test it, please rebase on top of release/3.2 - it is wip now, but it is
working

RussKie:
> I have missed that portable build nolonger uses `Setup\Prepare-Release.ps1`, but instead `Build\Prepare-Artifacts.ps1`. PR now builds and final zip contains working PM.

Sorry, I have refactored it few times.
I have refactored it again in 3.2 to streamline the build pipeline.

maraf:
@RussKie I have rebased the branch on `release/3.2`, but it fails to build. It is related to my changes?

mast-eu:
I changed the base branch for this PR from `master` to `release/3.2` and triggered a rebuild on AppVeyor:

image


The changeset of this PR is now back to normal (before it was 44 files with thousands of changed lines), but the build still fails.
Looking at the output on AppVeyor, it seems there's a problem in getting the XLF-files from Transifex.

Build/Prepare-Artifacts.ps1
```diff
@@ -18,6 +18,11 @@ if ($env:ARTIFACT_DEBUG_ENABLED -eq $true) {
     Push-AppveyorArtifact .\tree.txt
 }
 
+# ----------------------------------------------------------------------
+# download PluginManager
+# ----------------------------------------------------------------------
+& Setup\Download-PluginManager.ps1 -ExtractRootPath '..\Plugins\GitExtensions.PluginManager'
```
RussKie: Please move the script from `Setup` into `Build` folder

Setup/Prepare-Release.ps1
```diff
@@ -143,10 +143,10 @@ try {
...
```
RussKie: Setup/Prepare-Release.ps1  script is not used during the build pipeline, it is (still) only used for some manual tasks but by far and large it is obsolete

RussKie:
> Looking at the output on AppVeyor, it seems there's a problem in getting the XLF-files from Transifex.

You've found a bug 🎉 

Here's a fix:
```diff
diff --git a/Build/transifex/UpdateLocalEnglishTranslations.ps1 b/Build/transifex/UpdateLocalEnglishTranslations.ps1
index e2c8abbae..01e7eb4af 100644
--- a/Build/transifex/UpdateLocalEnglishTranslations.ps1
+++ b/Build/transifex/UpdateLocalEnglishTranslations.ps1
@@ -3,8 +3,8 @@ param (
    [string] $Configuration = 'Release'
 )
 
-pushd $PSScriptRoot
-
+Push-Location $PSScriptRoot
+try {
     Write-Host "Copying the latest English translation before the update..."
     $translationsFolder = Resolve-Path .\
     $releaseTranslationsFolder = Resolve-Path ..\..\GitExtensions\bin\$Configuration\Translation
@@ -13,14 +13,21 @@ xcopy "$translationsFolder\English*.xlf" "$releaseTranslationsFolder" /Y
     
     $src = Resolve-Path ..\..\GitExtensions\bin\$Configuration
     pushd "$src"
+    try {
         Write-Host "Updating the English translation..."
         Start-Process -FilePath "$src\TranslationApp.exe" -ArgumentList "update" -Wait
         if ($LASTEXITCODE -ne 0) {
-    popd
             exit -1
         }
-popd
+    }
+    finally {
+        Pop-Location
+    }
 
     Write-Host "Copying the updated English translation to commit..."
-Write-Debug " > $releaseTranslationsFolder`r`n > $translationsFolder"
-xcopy "$releaseTranslationsFolder\English*.xlf" "$translationsFolder"  /Y
+    Write-Host " > $releaseTranslationsFolder`r`n > $translationsFolder"
+    Move-Item -Path "$translationsFolder\English*.xlf" -Destination "$releaseTranslationsFolder" -Force
+}
+finally {
+    Pop-Location
+}
```

mast-eu:
Last night I tested the portable build from this PR when it was based on master branch: https://ci.appveyor.com/api/buildjobs/d4jebf45vply353j/artifacts/Setup%2FGitExtensions-Portable-3.2.0.6454.zip

I can confirm the PM was integrated and worked correctly. One issue emerged (https://github.com/gitextensions/gitextensions.pluginmanager/issues/27), but this is an issue in PM itself, not related to it's _integration_ into GE. Thus it is beyond the scope of this PR.

Will approve this PR as soon as the correct integration into GE installer is confirmed.

RussKie:
I fixed the build

mast-eu:
Unfortunately there are still hiccups (see AppVeyor log) and no published
artifact.

Il ven 23 ago 2019, 10:11 Igor Velikorossov <notifications@github.com> ha
scritto:

RussKie:
indeed. thank you, I'll have a look

RussKie:
The build is almost fixed.

The first error can be ignored, it needs a check that file exists before attempting to delete it
```
Removing GitExtensions-3.2.0.6480.msi
Remove-Item : Cannot find path 'C:\projects\gitextensions\GitExtensions-3.2.0.6480.msi' because it does not exist.
At C:\projects\gitextensions\Build\Prepare-Artifacts.ps1:40 char:5
```

The other one - the missing vsix - needs investigation.
I'm not sure I'll have time on the weekend, feel free to tinker.

mast-eu:
I fixed the build (by simply restarting it 😁 )

Seems lick the hiccup was on AppVeyors servers, not in our code: https://ci.appveyor.com/project/gitextensions/gitextensions/builds/26923655#L286

maraf:
@mast-eu I still don't see the installer.

mast-eu:
MSI artifacts are not available on GitExtensions' AppVeyor account, because installers are deliberately not build on pull requests. Therefore, I built it on my personal account.

https://ci.appveyor.com/project/mast-eu/gitextensions/builds/26958591/artifacts

RussKie:
You don't see an msi because this is a PR build, add a commit on top of
release/3.2 then you'll see it (you may not have permissions though, Martin
does).

mast-eu:
The last build on AppVeyor was successful, but apparently still has some issue in the log:
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/27022215#L240

mast-eu:
I noticed only after my force-push that this PR had merge conflicts with it's base branch `release/3.2`. 

I resolved the conflicts in my fork, but did not yet force-push them, because `release/3.2` doesn't currently build on AppVeyor (didn't check why exactly).

@RussKie @maraf I can rebase and update this branch, once the build on `release/3.2` branch is up and running again. Or should we go back on `master` for this PR?

RussKie:
I spent the past evening trying to build 3.2, it is failing and I was
unable to figure why.
I need to spend more time on this.

I suppose in mean time you can try to base this PR on the master, but I'm
not sure it will be conflict free.

RussKie:
The 3.2 branch appears to be in a good shape.
Please finalise the PR and we'll do a release.

mast-eu:
The change I pushed corrects the path where PluginManager is copied to. It was `C:\<UserDir>\AppData\Local\GitExtensions\GitExtensions\UserPlugins\*`, but should be `C:\<UserDir>\AppData\Local\GitExtensions\UserPlugins\GitExtensions.PluginManager\*`. See https://github.com/gitextensions/gitextensions.pluginmanager/issues/29.
The portable build has also been corrected.

@maraf: I changed your original implementation a bit. Most importantly, I removed `/Setup/PluginManager.wxs` and put everything in `/Setup/Product.wxs`. Please have a look.

RussKie:
Great news!
Not showing / not allowing to deselect is fully acceptable

RussKie:
if it builds, merge at will

PR #6668, opened 2019-05-25
Fix Ctrl+Backspace affecting ReadOnly TextBox

### LINKED ISSUE #6667: Ctrl+Backspace affects ReadOnly TextBox
## Current behaviour

`Ctrl`+`Bacspace` removes a word before the input caret even when `TextBox` is `ReadOnly`

Reported by @Quppa [here](https://github.com/gitextensions/gitextensions/pull/6256#issuecomment-493826018)

## Expected behaviour

`Ctrl`+`Bacspace` does not affect `ReadOnly` `TextBox`

## Steps to reproduce

image

## Diagnostics

Introduced by #6256

NikolayXHD:
Fixes #6667 introduced by #6256

## Test methodology

Manually verified the problem from screenshot below was fixed.

image

## Test environment(s)

- Windows 7 SP1

----

### DIFF
--- a/GitExtUtils/GitUI/ControlHotkeyExtensions.cs
+++ b/GitExtUtils/GitUI/ControlHotkeyExtensions.cs
@@ -35,7 +35,7 @@ namespace GitUI
 
         private static void HandleKeyDown(object sender, KeyEventArgs e)
         {
-            if (e.KeyData == (Keys.Control | Keys.Back))
+            if (e.KeyData == (Keys.Control | Keys.Back) && !IsReadOnly(sender))
             {
                 e.Handled = true;
                 e.SuppressKeyPress = true;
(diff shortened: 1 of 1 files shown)


RussKie:
Thank you.

Btw have you seen how the similar behaviour was implemented in https://github.com/dotnet/winforms/pull/629?

NikolayXHD:
> Thank you.
> 
> Btw have you seen how the similar behaviour was implemented in [dotnet/winforms#629](https://github.com/dotnet/winforms/pull/629)?

Thanks for pointing me to it. Looking at it I found a similar case I missed here: `ComboBox` when `DropDownStyle != ComboBoxStyle.DropDownList` should not be affected by hotkey as well.

PR #6671, opened 2019-05-26
Checkbox for line numbers hide/show

cryham:
## Proposed changes
I added a checkbox in Settings on Detailed - Diff viewer (DiffViewerSettingsPage) for line numbers hide/show.
Works right for Diff page, but on File tree page it works on next run..

## Screenshot
screen

## Test methodology
Manually.

## Test environment(s)
- GIT 2.21
- Windows 7 SP1

----

### DIFF
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -112,6 +112,13 @@ namespace GitUI.Editor
             _currentViewPositionCache.Capture();
 
             OpenWithDifftool = openWithDifftool;
+            if (!AppSettings.ShowLineNumbers)
+            {
+                isDiff = false;
+            }
+
+            TextEditor.ShowLineNumbers = false;
+
             _lineNumbersControl.Clear(isDiff);
             _lineNumbersControl.SetVisibility(isDiff);
 
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1364,6 +1364,12 @@ namespace GitCommands
             set => SetBool("RememberNumberOfContextLines", value);
         }
 
+        public static bool ShowLineNumbers
+        {
+            get => GetBool("ShowLineNumbers", true);
+            set => SetBool("ShowLineNumbers", value);
+        }
+
         public static string GetDictionaryDir()
         {
             return Path.Combine(GetResourceDir(), "Dictionaries");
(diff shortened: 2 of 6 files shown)


GitUI/Editor/FileViewerInternal.cs
```diff
@@ -112,6 +112,13 @@ public void SetText(string text, Action openWithDifftool, bool isDiff = false)
             _currentViewPositionCache.Capture();
 
             OpenWithDifftool = openWithDifftool;
+            if (!AppSettings.ShowLineNumbers)
+            {
+                isDiff = false;
+            }
+
+            TextEditor.ShowLineNumbers = false;
```
gerhardol: As intended?

cryham: Yes, this hides an extra 3rd row (after the 2 own with diff) with numbers going 1,2,3..

RussKie:
Could you please describe a use case where a user would wish to disable line numbers in the diff control?
I can understand a desire to hide line number for authors in the Blame control (#6430), but don't see a compelling value here...
Thank you

drewnoakes:
@RussKie one way of looking at this is that it matches the output of `git diff` on the command line. The line numbers are mentioned in the header.

On the other hand it creates yet another option in the already busy options pages.

RussKie:
> one way of looking at this is that it matches the output of `git diff` on the command line. The line numbers are mentioned in the header.

But why?

cryham:
Well, I personally don't need to see line numbers at all, anywhere, mostly. They take up some valuable space (width). Without them I can see more code, e.g. if lines are long, and with file paths on left. But even if all fits, the view is clearer for me without those numbers.
I do occasionally jump to a line by number, but I don't feel like seeing them all the time. I can easily show external diff for better orientation if needed.

mstv:
I suggest to add this setting to the FileDiff's toolbar:
grafik

cryham:
Right, I also think it's a good idea.
I there an icon for line numbers?
Still to do here is making checkbox toggle update the FileViewer (not only Internal one). Currently works after program restart.

mstv:
We prefer to use the [Fugue Icons](https://p.yusukekamiyamane.com/) if applicable.

mstv:
1. Switching line numbers off and on using the toolbar icon does not affect all necessary flags.
2. There's an additional margin when the Commit dialog is opened with line numbers off.
3. The toolbar icon should swap places with the `Show entire file` button.
4. The hash icon has a different meaning in the domain of Git. I would use one showing an enumerated list.

RussKie:
Closing as stale. None of the comments have been addressed.
If you wish to pursue it further please ping us to re-open. 
Thanks

PR #6727, opened 2019-05-31
CommitInfo: Bugfix "Contained in branches/tags"

### LINKED ISSUE #6725: CommitInfo: Display of "Contained in branches/tags" broken
## Current behaviour

CommitInfo omits the info "Contained in branches/tags: ...".

## Expected behaviour

"Contained in branches/tags: ..." is displayed.

## Screenshots

3.2.0.5986:
grafik

3.2.0.5985:
grafik

Yes, 3.2.0.5985.
Broken by #6645.

## Environment

- Git Extensions 3.2.0.5986
- Build 0622e28a3f6c5fd9c6b5e19685af363db53b66ec (Dirty)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

mstv:
Fixes #6725

## Proposed changes

- restore call of `RefreshSortedRefs` in the `UICommandsSourceSet` event handler
- remove unnecessary call of `ReloadCommitInfo`
- adapt test setup

### Before

grafik

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build a62ac4df274a67c0811011181011ad8e33ba82dd
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -97,8 +97,10 @@ namespace GitUI.CommitInfo
             {
                 this.InvokeAsync(() =>
                 {
-                    UICommandsSource.UICommandsChanged += UICommandsSource_UICommandsChanged;
-                    ReloadCommitInfo();
+                    UICommandsSource.UICommandsChanged += delegate { RefreshSortedRefs(); };
+
+                    // call this event handler also now (necessary for "Contained in branches/tags")
+                    RefreshSortedRefs();
                 }).FileAndForget();
             };
 
@@ -136,18 +138,13 @@ namespace GitUI.CommitInfo
             commitInfoHeader.SetContextMenuStrip(commitInfoContextMenuStrip);
         }
 
-        private void UICommandsSource_UICommandsChanged(object sender, GitUICommandsChangedEventArgs e)
+        private void RefreshSortedRefs()
         {
             if (!Module.IsValidGitWorkingDir())
             {
                 return;
             }
 
-            RefreshSortedRefs();
-        }
-
-        private void RefreshSortedRefs()
-        {
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
                 await LoadSortedRefsAsync();
@@ -284,6 +281,8 @@ namespace GitUI.CommitInfo
                 UpdateRevisionInfo();
                 StartAsyncDataLoad();
             }
+
+            return;
 
             void UpdateCommitMessage()
             {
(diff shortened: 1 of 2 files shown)


mstv:
The AppVeyor build failed twice because CodeCov seems to be down.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -98,6 +98,7 @@ public CommitInfo()
                 this.InvokeAsync(() =>
                 {
                     UICommandsSource.UICommandsChanged += UICommandsSource_UICommandsChanged;
+                    RefreshSortedRefs();
```
RussKie:
```suggestion

                    // necessary for "Contained in branches/tags"
                    RefreshSortedRefs();
```

mstv:
Sorry CodeCov! The build script uploads coverage data although the tests have failed. This is OK but it eluded me.

The mockup of several tests needs to be adpated because of the restored early call of `GetSortedRefs`:
> Message: System.Exception : Unexpected arguments: for-each-ref --sort=-committerdate --sort=-taggerdate --format="%(refname)" refs/

RussKie:
It started to come back to me... I really struggled with testing the implementation because of the background calls kicked off from the ctor... I think that was one of the reasons for me to remove the line in question (obviously without realising it was needed).
😒

mstv:
The test setup of `CommitInfoTests` does already execute a Git command, which needs a mockup, too.
I guess this made you think `GetSortedRefs` was called redundantly.

But the other call there - `ReloadCommitInfo` - actually is not necessary since `GetSortedRefs` calls `UpdateCommitInfo` in the end.

There must be something missing in the test teardown because failing tests cause other tests to fail, which do pass if run separately.

mstv:
Review and merge, please.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -98,7 +98,9 @@ public CommitInfo()
                 this.InvokeAsync(() =>
                 {
                     UICommandsSource.UICommandsChanged += UICommandsSource_UICommandsChanged;
-                    ReloadCommitInfo();
+
+                    // call this event handler also now (necessary for "Contained in branches/tags")
+                    UICommandsSource_UICommandsChanged(null, null);
```
gerhardol: Normally, we do not call the event handler directly but use a shared method.

mstv: Yes, I have eliminated the separate handler.

PR #6728, opened 2019-06-01
Do not limit "Contained in branches/tags"

### LINKED ISSUE #5966: Overflow of superproject references
bug

scrambled display and unclear `...` reference(s) after some time
3.00.00.4433:
grafik

current pre 3.01.00 master 7211073:
grafik

- open the GE repo
- go to the subproject "ICSharpCode.TextEditor"
- move the mouse around for about 30 seconds

The revision grid is invalidated/refreshed correctly when the background process has finished.
The omitted info behind grafik should be displayed somewhere (if there is really something behind -- see the difference between the head commits in the screenshots).

- Git Extensions 3.00.00.4433
- Git 2.10.1.windows.1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

new feature in 3.x

mstv:
Fixes #5966 (part 2 of 2).

## Proposed changes

- remove the limit of at most 20 branches and 20 tags displayed which contain the current commit

### Before

grafik

### After

grafik

Don't bother regarding the fonts. It's just settings.

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 631db5be269073e6e90faf20503c3a3d07f18505
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -56,7 +56,6 @@ namespace GitUI.CommitInfo
         private static readonly TranslationString _plusCommits = new TranslationString("commits");
         private static readonly TranslationString _repoFailure = new TranslationString("Repository failure");
 
-        private const int MaximumDisplayedRefs = 20;
         private readonly ILinkFactory _linkFactory = new LinkFactory();
         private readonly ICommitDataManager _commitDataManager;
         private readonly ICommitDataBodyRenderer _commitDataBodyRenderer;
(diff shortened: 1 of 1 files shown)


mstv:
It's not the first time that I get in conflict with the maintenance window(?) of CodeCov at this time of night.

mstv:
> OK for me after #6727 is merged

Just rebased on merge of #6727.

> I may be annoyed by the information this introduces, but also if info is not available.

So do I. I have even copied this info to a text editor in order to be able to search.

pmiossec:
>So do I. I have even copied this info to a text editor in order to be able to search.

I wonder what is the possibility to load it on demand (by a click on a link), because that's an information we rarely use.

And to display them one ref by line to make it a lot more readable.

gerhardol:
> I wonder what is the possibility to load it on demand (by a click on a link), because that's an information we rarely use.
And to display them one ref by line to make it a lot more readable.

Good idea. I support change in this PR or merge now and add a separate issue.

RussKie:
I'm not going to block this, but this may be a pain for those who marks every successful build with a tag (in all my previous engagements we did). 
When you have a sizable repo with thousands of commits - the UX with this change won't be pretty.

Perhaps we may want to show only the last 10 tags and have `[Show all]` linklabel to show all.

RussKie:
Closing as stale.
The proposed functionality in its current form may have an unexpected and unpleasant experience for some users.

PR #6766, opened 2019-06-07
Fixed Gerrit plugin using deprecated features

### LINKED ISSUE #6127: Gerrit plugin using deprecated features
## Current behaviour

The Gerrit plugin is using deprecated features
 - draft change mentioned on the publish form, Gerrit now supports 'work in progress' and 'private' changes
 - commits are pushed to 'refs/publish' instead of 'ref/for'

## Expected behaviour

I presume we don't want to define a minimum version of Gerrit for the plugin, and there isn't an easy way of using a older version of the plugins, so I suspect we'll need to detect the version of Gerrit that the project is using and adapt the form/paths appropriately.

veremenko-y:
Fixes #6127
Changed deprecated options for publish type and topic in Gerrit to support newer version

## Proposed changes

Changed push reference to use /refs/for with %wip and %private for instead of deprecated /refs/publish and /refs/draft
Changed the way topic is passed into gerrit to use new `-o topic=TOPIC` format.

## Version compatibility

Draft changes (that are currently used in Git Extensions) have been marked obsolete in Gerrit 2.15 in 2017 (
https://github.com/GerritCodeReview/gerrit/commit/4b44bdc7022653cec0cf0322b1161a210d46b8fb).
`refs/publish` have been removed last year (https://github.com/GerritCodeReview/gerrit/commit/605941e57f1fefd029e8c6b6af63a04dfbcaf08d)
`refs/for` syntax used in this PR to publish changes, have been introduced around 2011-2013, and were synonym of `refs/publish`. This change will break pushing "draft" changes, but it should not break basic publish for older versions of Gerrit.

As of Gerrit 3.0.0 `refs/draft` and `refs/publish` are deprecated and removed from the system. Instead, private and work-in-progress should be used.

Current changes in this PR are based on current 3.0.0 documentation https://gerrit-documentation.storage.googleapis.com/Documentation/3.0.0/user-upload.html#push_create

### Before

image

### After

image

## Test methodology

Tested on Gerrit 3.0.0 with combinations of Publish Type and topic

## Test environment(s)

- GIT 2.21.0.windows.1
- Windows 10
- Gerrit 3.0.0

----

### DIFF
--- a/Plugins/Gerrit/FormGerritPublish.Designer.cs
+++ b/Plugins/Gerrit/FormGerritPublish.Designer.cs
@@ -33,13 +33,14 @@
             this.AddRemote = new System.Windows.Forms.Button();
             this._NO_TRANSLATE_Remotes = new System.Windows.Forms.ComboBox();
             this.labelRemote = new System.Windows.Forms.Label();
-            this.PublishDraft = new System.Windows.Forms.CheckBox();
             this.labelBranch = new System.Windows.Forms.Label();
             this.labelTopic = new System.Windows.Forms.Label();
             this._NO_TRANSLATE_Branch = new System.Windows.Forms.TextBox();
             this._NO_TRANSLATE_Topic = new System.Windows.Forms.TextBox();
             this._NO_TRANSLATE_Reviewers = new System.Windows.Forms.TextBox();
             this.labelReviewers = new System.Windows.Forms.Label();
+            this.PublishType = new System.Windows.Forms.ComboBox();
+            this.labelPublishType = new System.Windows.Forms.Label();
             this.SuspendLayout();
             // 
             // Publish
@@ -86,19 +87,9 @@
             this.labelRemote.Size = new System.Drawing.Size(48, 13);
             this.labelRemote.TabIndex = 0;
             this.labelRemote.Text = "Remote:";
-            // 
-            // PublishDraft
-            // 
-            this.PublishDraft.AutoSize = true;
-            this.PublishDraft.Location = new System.Drawing.Point(101, 144);
-            this.PublishDraft.Name = "PublishDraft";
-            this.PublishDraft.Size = new System.Drawing.Size(134, 17);
-            this.PublishDraft.TabIndex = 9;
-            this.PublishDraft.Text = "Submit review as draft";
-            this.PublishDraft.UseVisualStyleBackColor = true;
-            // 
+            //
             // labelBranch
-            // 
+            //
             this.labelBranch.AutoSize = true;
             this.labelBranch.Location = new System.Drawing.Point(13, 52);
             this.labelBranch.Name = "labelBranch";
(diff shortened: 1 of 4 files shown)


gerhardol:
Any gerrit users that have an opinion?

Plugins/Gerrit/FormGerritPublish.cs
```diff
@@ -95,6 +101,13 @@ private bool PublishChange(IWin32Window owner)
             pushCommand.CommandText = PushCmd(
                 _NO_TRANSLATE_Remotes.Text,
                 targetBranch);
+
+            string topic = _NO_TRANSLATE_Topic.Text.Trim();
+            if (!string.IsNullOrEmpty(topic))
+            {
+                pushCommand.CommandText += " -o topic=" + topic;
```
gerhardol: Note: This will break if using Git version < 2.10.2. Not supported, but users occasionally tries using older versions.

RussKie: Since we do not force the minimum git version (though I think we should), it is better to wrap this in a git version check.

veremenko-y: Hi @gerhardol, I have added the reference to the documentation and reasoning for the changes, please let me know if that is sufficient. As of using `-o` flag, I will fix it with git check, as there is still another way to post the same data to Gerrit.

Do you have any documentation or example on how should I proceed with translations? I checked the wiki on GitHub, but failed to find anything.

gerhardol: I will approve also with the minor problem with using unsupported Git - if you use latest Gerrit, it is more likely that you use recent Git too. (Also GitArgumentBuilder could be used.)
To merge, I also want at least another Gerrit user to review - I have no feeling for when this should be merged. Also, some other member need to approve.

gerhardol:
(translation changes looks as expected)

Plugins/Gerrit/FormGerritPublish.Designer.cs
```diff
@@ -73,102 +74,123 @@ private void InitializeComponent()
...
+            this.labelPublishType.Name = "labelPublishType";
+            this.labelPublishType.Size = new System.Drawing.Size(90, 17);
+            this.labelPublishType.TabIndex = 12;
+            this.labelPublishType.Text = "Publish Type";
+            // 
             // FormGerritPublish
             // 
             this.AcceptButton = this.Publish;
-            this.AutoScaleDimensions = new System.Drawing.SizeF(96F, 96F);
+            this.AutoScaleDimensions = new System.Drawing.SizeF(120F, 120F);
```
RussKie: ⚠️  This is not good. Please update the designer at 100% scaling.
The easiest is to do with `devenv /noscaling` (writing from memory)

https://docs.microsoft.com/en-us/dotnet/framework/winforms/disable-dpi-awareness-visual-studio#restart-visual-studio-as-a-dpi-unaware-process
image

RussKie:
I don't use Gerrit so I am quite indifferent. The code change looks sensible.
We just need to fix up the designer.

Plugins/Gerrit/FormGerritPublish.Designer.cs
veremenko-y: Thank you. Will do.

RussKie: Correction the correct switch is `/noscale`
image

veremenko-y: I have fixed the designer with `/noscale`

nick-bulleid:
I've used this PR with Gerrit 3.0.0. It works as expected.

RussKie:
Please squash commits and it is ready to go.
Thanks

Issue #5949, opened 2018-12-18
commit message hilight


chris-findlay:
**Question**

**What is the current behavior?**
The commit message gets a yellow hilight after 50 chars or 1 linebreak.  Where do I get to configure the colours/column limit?

I tried searching in the User Manual, but it's borken at the moment...

**Environment you encounter the issue:**
- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

vbjay:
Start in this code.
https://github.com/vbjay/gitextensions/blob/9a490bfb3d7d85cf316a9daf256eb2c2703983b1/GitUI/CommandsDialogs/FormCommit.cs#L2623

RussKie:
I believe the highlighting is done here:
https://github.com/gitextensions/gitextensions/blob/f575439c38ed8ab433bbf20f9f87659eff138db6/GitUI/SpellChecker/EditNetSpell.cs#L314-L343

The actual setting is toggled via a context menu on the `FormCommit`:
![image](https://user-images.githubusercontent.com/4403806/50203178-bd833500-03b4-11e9-9a21-285dfce572ef.png)

And it appears the limit is hardcoded and can't be configured.

chris-findlay:
Thanx.  Now I understand the pattern it makes more sense, but in combination with the option that ensures the 2nd line is blank (i.e. a 2 line message entered becomes 3 by the insertion of the blank line) the hilight for the 2nd line is wrong.
Either way, being able to configure this (including what it looks like) would be great.

RussKie:
I agree. The original code probably dates a number of years back, and may
do with a refactor/cleanup.

In the grand scheme of things this request is unlikely to have a high
priority stuck behind of a number of other burning issues.

A PR is certainly welcome, if you wish to have a go at it.

drewnoakes:
> in combination with the option that ensures the 2nd line is blank (i.e. a 2 line message entered becomes 3 by the insertion of the blank line) the hilight for the 2nd line is wrong.

I am not following this, sorry. Can you paste a screenshot showing what you mean? Are you talking about a line so long it wraps in the editor? That shouldn't make a difference.

[...]

PR #6774, opened 2019-06-09
#5949 - Update EditNetSpell to respect comment line limits

lukevp:
- Update EditNetSpell to respect the line limits configured in the commit settings rather than highlighting based on hardcoded values.

Fixes #

## Proposed changes

- Update the spell check code to respect line length settings in appsettings rather than being hardcoded

### Before

image
image

### After

No highlighting with settings disabled
image

With settings enabled, highlights as before.
image
image

## Test methodology 

- Tested with settings enabled and disabled and ensured proper functionality.

## Test environment(s) 

- GIT 2.22.0.windows.1
- Windows 10 Pro 1903 Build 18362.53

----

### DIFF
--- a/GitUI/SpellChecker/EditNetSpell.cs
+++ b/GitUI/SpellChecker/EditNetSpell.cs
@@ -432,20 +432,37 @@ namespace GitUI.SpellChecker
                 return;
             }
 
+            int limit1 = AppSettings.CommitValidationMaxCntCharsFirstLine;
+            int limitX = AppSettings.CommitValidationMaxCntCharsPerLine;
+            bool empty2 = AppSettings.CommitValidationSecondLineMustBeEmpty;
+
             var numLines = TextBox.Lines.Length;
             var chars = 0;
             for (var curLine = 0; curLine < numLines; ++curLine)
             {
                 var curLength = TextBox.Lines[curLine].Length;
-                var curMaxLength = 72;
+                var curMaxLength = limitX;
                 if (curLine == 0)
                 {
-                    curMaxLength = 50;
+                    curMaxLength = limit1;
+                    if (limit1 == 0)
+                    {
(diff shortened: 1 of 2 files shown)


Bejoscha:
I like this change and am all for it.

pmiossec:
I don't like all the consequences of this change (if I understand well what it is doing)...

You are linking what is a hard commit validation feature (that must be respected) to the one of ill commit highlight (which is just a help).

So, with your changes, I can't have yellow highlight anymore (that I want to enforce good practices), without the hard limit (because I want sometimes to write longer lines and so exceed the limits).

But I like the idea to be able to change the values. But I tend to think that it should be 2 other fields that should be added.

And in the same time, perhaps also add the checkbox to enable the 'mark ill commit message' here.

vbjay:
Read up on 50/72 rule.

GitUI/SpellChecker/EditNetSpell.cs
```diff
@@ -432,21 +432,38 @@ private void MarkLines()
                 return;
             }
 
+            int limit1 = AppSettings.CommitValidationMaxCntCharsFirstLine;
+            int limitX = AppSettings.CommitValidationMaxCntCharsPerLine;
+            bool empty2 = AppSettings.CommitValidationSecondLineMustBeEmpty;
```
RussKie: Please rename variable to better communicate their purpose, e.g.
```
limitMaxCharsFirstLine
limitMaxCharsOtherLines
secondLineMustBeEmpty
```

RussKie:
I would like to have tests for the change.
Do you think you can add some? Happy to help you with the setup

RussKie:
Also please update the commit message with "Fixes #5949", this way it stays in the history.

GintasS:

Issue #6757, opened 2019-06-06
[NBug] Object reference not set to an instance of an object.


SergiiGalai:
## Current behaviour
No pull request window or redirect to browser happened. See an exception window instead

## Expected behaviour
No exception window, create pull request dialog is opened

## Steps to reproduce
1. Do Not authenticate on Github inside Git Extensions
2. Press GitHub=>Create Pull Request ... menu item for existing branch
3. Exception occurred with an option to send bug

## Error Details
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Git.hub.Repository.GetBranches()
   at GitHub3.GitHubRepo.GetBranches()
   at GitUI.CommandsDialogs.RepoHosting.CreatePullRequestForm.<>c__DisplayClass22_0.<<PopulateBranchesComboAndEnableCreateButton>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.ThreadHelper.<>c__DisplayClass13_0.<<FileAndForget>b__0>d.MoveNext()
```

## Environment
- Git Extensions 3.1.1.6049
- Build 2f872105b4eb42421d9fdafd68e90a9e27d20d96
- Git 2.20.0.windows.1
- Microsoft Windows NT 10.0.16299.0
- .NET Framework 4.7.3394.0
- DPI 96dpi (no scaling)

SergiiGalai:
I have reviewed all potential duplicates listed above - all of them are different issues

gerhardol:
> I have reviewed all potential duplicates listed above - all of them are different issues

#6625 is a duplicate of this.

RussKie:
Let's continue on this one though as @SergiiGalai has provided information.

I couldn't repro the issue though...

pmiossec:
I have investigated a little and I have an idea (though, not 100% clear) of what happens...

It seems to be a problem of authentication when querying the github apis.
By luck on my personal computer, I don't have the OAuth token filled (probably lost due to  #6696 ).
Anyway....

[...]

PR #6791, opened 2019-06-12
CreatePullRequestForm: prevent multiple same call to github api

pmiossec:
Part of a solution to #6757

## Proposed changes

- Move a call to a github api outside of a for loop to do only one call

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.0
- Build c501b6ac6f64d845dc35af19b009fea50570f14c
- Git 2.20.1.windows.1
- .NET Framework 4.7.3394.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
@@ -141,9 +141,10 @@ namespace GitUI.CommandsDialogs.RepoHosting
                         comboBox.Items.Clear();
 
                         var selectItem = 0;
+                        var defaultBranch = hostedRepository.GetDefaultBranch();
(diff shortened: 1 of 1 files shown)


pmiossec:
>The change is fine. If this is just a part of the problem, is worth merging?

Yes, it is worth merging. It solves the performance problem that makes the use of a real pain to use (when you have some branches). So it will benefit to all the users of the feature.

>Part of a solution to #6757

I thought it was true, but thinking twice, I'm not still sure. It doesn't solve the problem and maybe it prevent it to happen some times, but I can't be sure. It depends how github manage the quota to access the api when not authenticated.

And I still wait that the solution to fill a github token to be tested by a user to see if it solves the problem.
And also, I still don't know if I will take some time to fix the problem soon (I have some other priorities).

So in the mean time, that PR could be merged...

Issue #6607, opened 2019-05-19
Request more information via NBug submission form


RussKie:
## Feature description

The uplift of NBug submission mechanism seems to be quick successful - we are now getting more reports about issues.
However most users fail to provide all required information that causes a) a lot of duplicates and b) a lot of requests from the Team for more information.

The NBug sumission form must be enhanced to request more information before a submission can be made.

The following information is required:
- Current behaviour
- Expected behaviour
- Steps to reproduce

Also advise users if their git version is below recommended.

pmiossec:
It will be much more complicated but I was wondering if it's possible to search for duplicate before asking for the informations and opening the existing issue instead of creating one...

RussKie:
#6608 :wink:

pmiossec:
yes, I just saw it 😉

gerhardol:
For 3.1.1, should we just add a comment in the default text asking the user to provide information or the issue will be closed?

RussKie:
For 3.1.1 may be easier to spin up a duplicate check bot as discussed in #6608?

gerhardol:
> For 3.1.1 may be easier to spin up a duplicate check bot as discussed in #6608?

Duplicates and improve information are separate.
We may choose to ignore this in 3.1.1 though.
I just suggest we do something like the following in 3.1.1. Could be new text boxes too in NBug, but this should be clear enough.

## Current behaviour

Replace this text with the current behavior.
Issues without information will be closed.

## Expected behaviour

## Steps to reproduce

Take some time to try and reproduce the issue, then explain how to do so here.

RussKie:
How about the following dialog which appears when a user clicks on "Send and Quit" button?

![image](https://user-images.githubusercontent.com/4403806/58365342-5723b500-7f06-11e9-8f22-2c6ea94e7883.png)

or

![image](https://user-images.githubusercontent.com/4403806/58365502-d2866600-7f08-11e9-9358-550ae60610c6.png)

gerhardol:
A popup is good. There is still some value in empty issues to get statistics so I would like to change the wording. 
A default text in the issue can be used by a bot and explain for other users why issues at just closed. 

On mobile now but can write a proposal tonight if no-one else does that.

gerhardol:
Second popup:
Without this information your issue will be closed (but the information can be together with other issues).

Default text - or something the bot can use?
See above.

RussKie:
Apologies Gerhard, but your last message eludes me.
[...]

PR #6805, opened 2019-06-15
feat: Migrate NBug bug report from NBug to Git Extensions project

RussKie:
Relates to #6607

## Proposed changes

Migrate NBug bug report form into the main project from subproject.

The NBug project was introduced into the project to automate capturing exception information. 
However the NBug project is now stale and it missed a number of features required by the Git Extensions project (e.g. it wasn't integrated with GitHub). So we have made diverging changes to the NBug project to accommodate requirements of the Git Extensions project (e.g. #6446, #6455).

However further changes were necessary (i.e. to address #6607), which further would take us further from the original NBug project (e.g. request repro steps before a submission). 
Beside the NBug project contained too much of unnecessary functionality not used in our project, and we didn't have access to the NBug's localisation pipelines.

Bringing only necessary functionality into Git Extensions addresses the above issues.

## Screenshots:

* Before
image

* After
image

## Test methodology 

- manually
- run a packaging script

----


RussKie:
> However what has been done in the new NBug files?

For the most part it is a copy-paste change.
There are few minor changes: "NBug version" is replaced with "Git version" and the environment info is passed as a parameter instead of a function call.

The real changes will be in a separate PR.

RussKie:
Added screenshots and fixed the wording

PR #6807, opened 2019-06-16
WIP - Fix #5944 by fixing --follow and revision reading

### LINKED ISSUE #5944: History no longer shown for renamed files

fschmied:
Quickfix for #5944 based on release/3.1, no tests, no big refactoring.

Intended as a placeholder until #6132 is done.

## Proposed changes
- a306eb6 by mistake removed the "--follow" option when getting the git log for history. This is reverted by this PR.
- 1f6b0ca by mistake removed parsing of the file name data into GitRevision.Name. This is also reverted by this PR.
- Together, these should fix rename tracking in the file history.

"WIP" label because #6132 was first and I don't intend to steal this issue.

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -28,6 +28,7 @@ namespace GitCommands
 
     public sealed class RevisionReader : IDisposable
     {
+        private const string EndOfBody = "1DEA7CC4-FB39-450A-8DDF-762FCEA28B05";
         private const string FullFormat =
 
               // These header entries can all be decoded from the bytes directly.
@@ -47,7 +48,7 @@ namespace GitCommands
               /* Committer name  */ "%cN%n" +
               /* Committer email */ "%cE%n" +
               /* Commit subject  */ "%s%n%n" +
-              /* Commit body     */ "%b";
+              /* Commit body     */ "%b" + EndOfBody;
 
         private readonly CancellationTokenSequence _cancellationTokenSequence = new CancellationTokenSequence();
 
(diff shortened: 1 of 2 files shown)


gerhardol:
This should be submitted to master, next planned release is 3.2
For me, it looks OK until #6132 is done

gerhardol:
@fschmied 
I used your commit without asking first (was in a hurry and should have been clearer in the previous comment).
Is it OK to continue on your commit even if I need to make changes and rebase/squash you may longer be the author?

Also, it is not likely that this is included in a 3.1 release, so this PR may be closed.

fschmied:
> Is it OK to continue on your commit even if I need to make changes and rebase/squash you may longer be the author?
@gerhardol Yes, no problem. Just note that this PR is just a temporary quickfix (dealing with the two changes that broke the behavior) while @vbjay is working on a better solution in #6132 .

jrfnl:
I've just tested this fix and I can confirm that when opening the history of a renamed file _after_ the rename,, all commits now show again correctly - including the commits from before the rename.

However, when the history is opened from a commit _before_ the rename, the history still hangs.

Issue #6556, opened 2019-05-14
Add Submodule: Fill combobox with branches regardless of the repo URL's type


mstv:
## Current behaviour

I have tried to add a (first) submodule to a repo.
The exception is thrown when I try to open the yet empty `Branch` combobox.

This combobox is not filled for remote repos at all, regardless whether specified with HTTPS URL or with SSH URL.

## Expected behaviour

list of available branches filled correctly
no unhandled exception
no exception for a valid URL

## Steps to reproduce

- create a standalone clone of the repo which shall be used a submodule
- copy its `URL` from the `Remote repositories` dialog ("git@host:user/path.git")
- open GE for the destination repo
- `Manage submodules`, `Add submodule`
- paste the URL
- drop down the `Branch` combobox, which was empty yet

## Error details 3.2.0
- combobox does not contain the branches of the sub-repo unless an URI of type `//file:` is given

(no unhandled exception)

## Error Details 3.1.0
```
System.UriFormatException: Ungültiger URI: Das URI-Schema ist ungültig.
   bei System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   bei GitCommands.PathUtil.NormalizePath(String path)
   bei GitCommands.GitModule..ctor(String workingDir)
   bei GitUI.CommandsDialogs.SubmodulesDialog.FormAddSubmodule.BranchDropDown(Object sender, EventArgs e)
   bei System.Windows.Forms.ComboBox.OnDropDown(EventArgs e)
   bei System.Windows.Forms.ComboBox.WmReflectCommand(Message& m)
   bei System.Windows.Forms.ComboBox.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

## Environment
- Git Extensions 3.1.0
- Build 678b69e632a2bff81c5e093770f605e0fe1151c8
- Git 2.21.0.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

RussKie:
Do you see this as a blocker for the release?

mstv:
This is a RL use case and should be fixed in the release, not necessarily in the RC2.

mstv:
The fix #6560 improves the behavior a little by ignoring the exception.
[...]

PR #6811, opened 2019-06-16
fix: Invalid URI when open invalid path

gerhardol:
fix: Invalid URI when open invalid path

Related issue: #6556

Cherry pick of #6560 for release/3.1 as this will fix some reported exceptions (mostly [NBug]).
This is not suggesting that there should be a 3.1.2 release, just a preparation.

Note: Translations are affected, could be reverted for 3.1

## Proposed changes
See #6560 
----


RussKie:
> This is not suggesting that there should be a 3.1.2 release, just a preparation.

We could publish a 3.1.2, if necessary

RussKie:
Thank you

gerhardol:
> We could publish a 3.1.2, if necessary

I probably prefer an early 3.2 instead, but it is easier to create PRs for important changes directly so they are ready if a 3.1.2 is necessary.

nnfhhvbb (approved):

PR #6814, opened 2019-06-17
Requires OAuthToken set to use GitHub3Plugin features

### LINKED ISSUE #6757: [NBug] Object reference not set to an instance of an object.

pmiossec:
That will open the web browser window to generate the token if needed.

+ fix a duplicate string

Fixes #6757 now that github api will return the good data instead of a 403

## Test methodology 

- manual

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -14,6 +14,7 @@ using GitUI.HelperDialogs;
 using GitUIPluginInterfaces;
 using GitUIPluginInterfaces.RepositoryHosts;
 using JetBrains.Annotations;
+using ResourceManager;
 
 namespace GitUI
 {
@@ -23,6 +24,9 @@ namespace GitUI
         private readonly ICommitTemplateManager _commitTemplateManager;
         private readonly IFullPathResolver _fullPathResolver;
         private readonly IFindFilePredicateProvider _findFilePredicateProvider;
+        private readonly TranslationString _viewPullRequest = new TranslationString("View pull requests");
+        private readonly TranslationString _createPullRequest = new TranslationString("Create pull request");
+        private readonly TranslationString _forkCloneRepo = new TranslationString("Fork or clone a repository");
 
         [NotNull]
         public GitModule Module { get; private set; }
@@ -1326,7 +1330,7 @@ namespace GitUI
 
         public void StartCloneForkFromHoster(IWin32Window owner, IRepositoryHostPlugin gitHoster, EventHandler<GitModuleEventArgs> gitModuleChanged)
         {
-            WrapRepoHostingCall("View pull requests", gitHoster, gh =>
+            WrapRepoHostingCall(_forkCloneRepo.Text, gitHoster, gh =>
             {
                 using (var frm = new ForkAndCloneForm(gitHoster, gitModuleChanged))
                 {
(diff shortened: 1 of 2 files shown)


GitUI/GitUICommands.cs
```diff
@@ -1326,7 +1326,7 @@ private void WrapRepoHostingCall(string name, IRepositoryHostPlugin gitHoster, A
 
         public void StartCloneForkFromHoster(IWin32Window owner, IRepositoryHostPlugin gitHoster, EventHandler<GitModuleEventArgs> gitModuleChanged)
         {
-            WrapRepoHostingCall("View pull requests", gitHoster, gh =>
+            WrapRepoHostingCall("Fork or clone a repository", gitHoster, gh =>
```
RussKie: We need to translate this

pmiossec: done.

RussKie:
Could you please also update translations.

I just pushed into master an updated script (e7688f6fd82614979d964cf69233b951ad521970).
You could either rebase on top of the latest master or just cherry-pick it, and run it:

```powershell
PS> pushd .\Build\transifex
PS> .\UpdateLocalEnglishTranslations.ps1 -Configuration Debug
```

pmiossec:
>Could you please also update translations.

I don't know why but the strings doesn't end up in the xlf files. I have other strings updated but not these ones 😢 
I don't know what is the rule to find `TranslationString`...

pmiossec:
>
> Dont really see why it works now and there is the history bug.
>

The main reason of the exception is that the github api returned a 403
error because the api is queried without token. So some calls failed,
throwing exceptions that were swallowed.

 All that end up with the 'Owner' property not set, so the value is null,
which is used for a last call and causing a NRE.

The change made here is that for all the feature provided by the github
plug-in, before opening the form, we check if the token is filled. If, not,
the web page to generate it is open.

All the code was already there, just missing the good implementation of
'IsConfigurationOK'...

This fix prevents the NRE to happen in the nominal scenario but looking at
the code, all api errors are not well handled and would require a lot of
work.

>

RussKie:
> I don't know why but the strings doesn't end up in the xlf files

I'll have a look

RussKie:
> I don't know what is the rule to find `TranslationString`...

Translatable strings must be defined in a class that implements `ITranslate`. Have a look at how `ResourceManager.Strings` is implemented.

PR #6838, opened 2019-06-22
Show changed images in all diff tabs

### LINKED ISSUE #1391: Browse --> Diff Tab: Added images are not shown visualized
When viewing changes before committing, GE shows images files as acutal pictures in the diff window.

After committing one can got to the Diff Tab but instead of an image only text is displayed:

```
diff --git a/GitUI/Resources/Icons/originals/orig_73.png b/GitUI/Resources/Icons/originals/orig_73.png
new file mode 100644
index 0000000..da0d890
```

(Additional nice to have feature: If there is a changed image, one could display the new version next to the old version of the image, like GitHub does it for example here https://github.com/gitextensions/gitextensions/commit/3a81cdfda136e73508a9fc524680a68a7e676dcf)

gerhardol:
Resolves #1391

Previously, images were displayed as binary diffs, except some in
FormCommit:  .png and new worktree images

Other:
* FormCommit always show the worktree commit for index .png
* Show all images in FormCommit
(i.e. show index images and changed other than .png)
* Regression from a94a30af90ddcb8bad08220e171af20fd1f31b19 (ObjectId)
where blob guid were ignored if existing "cache miss" and attempted to be used if null
(not sure where this is seen, blob id is not always fetched though)
* Combined guid should never be used to get file blob guid - assert added
* openWithDiffTool action were not called from stash and RevisionDiff

### Before
image

image

### After

image

image

## Test methodology 

Manual, change and stage image files

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3347,11 +3347,9 @@ namespace GitCommands
         [CanBeNull]
         public ObjectId GetFileBlobHash(string fileName, ObjectId objectId)
         {
-            if (objectId == ObjectId.WorkTreeId)
+            if (objectId == ObjectId.WorkTreeId || objectId == ObjectId.CombinedDiffId)
             {
-                // working directory changes
-                Debug.Assert(false, "Tried to get blob for worktree file");
-                return null;
+                throw new ArgumentException($"Tried to get blob for unsupported revision: {objectId} and file: {fileName}");
             }
 
             // TODO use regex for parsing
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1204,17 +1204,14 @@ namespace GitUI.CommandsDialogs
         private async Task SetSelectedDiffAsync(GitItemStatus item, bool staged)
         {
             Action openWithDiffTool = () => (staged ? stagedOpenDifftoolToolStripMenuItem9 : openWithDifftoolToolStripMenuItem).PerformClick();
-            if (item.Name.EndsWith(".png"))
+            if (!item.IsTracked || FileHelper.IsImage(item.Name))
             {
-                await SelectedDiff.ViewFileAsync(item.Name, openWithDiffTool);
-            }
-            else if (item.IsTracked)
-            {
-                SelectedDiff.ViewCurrentChanges(item, staged, openWithDiffTool);
+                var guid = staged ? ObjectId.IndexId : ObjectId.WorkTreeId;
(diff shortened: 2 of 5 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -3347,10 +3347,11 @@ public string GetFileText(ObjectId id, Encoding encoding)
         [CanBeNull]
         public ObjectId GetFileBlobHash(string fileName, ObjectId objectId)
         {
-            if (objectId == ObjectId.WorkTreeId)
+            if (objectId == ObjectId.WorkTreeId || objectId == ObjectId.CombinedDiffId)
             {
                 // working directory changes
-                Debug.Assert(false, "Tried to get blob for worktree file");
+                Debug.Assert(false, "Tried to get blob for unsupported revision: " +
```
RussKie: [nit] Interpolated it would probably read nicer

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1204,17 +1204,14 @@ private void ShowChanges(GitItemStatus item, bool staged)
         private async Task SetSelectedDiffAsync(GitItemStatus item, bool staged)
         {
             Action openWithDiffTool = () => (staged ? stagedOpenDifftoolToolStripMenuItem9 : openWithDifftoolToolStripMenuItem).PerformClick();
-            if (item.Name.EndsWith(".png"))
+            if (FileHelper.IsImage(item.Name) || !item.IsTracked)
```
```suggestion
            if (!item.IsTracked || FileHelper.IsImage(item.Name))
```

GitCommands/Git/GitModule.cs
gerhardol: Done. Also made into ArgumentException, reviewed use again and ignored Combined in one call path (where it never was used though).

mstv:
👍 Thank you! LGTM, too.

gerhardol:
Rebased and squashed - leaving one or two days for further comments before merging

pmiossec:
Very good!
A futur improvement could be to display the old and new image side by side in the diff view...

RussKie:
> A futur improvement could be to display the old and new image side by side in the diff view...

Indeed.
I think at some point I looked at that... I don't think it was a quick fix 😞

gerhardol:
> A futur improvement could be to display the old and new image side by side in the diff view...

I believe it is possible. A System.Drawing.Image must be created from two file streams, quite basic image handling. A special version of ViewFileAsync() with two filename in parameters.
The call chains will be slightly more complicated though.

Another improvement would be to scale up small images like icons.

pmiossec:
>
> I believe it is possible. A System.Drawing.Image must be created from two
> file streams, quite basic image handling


It's also what I have in mind.

PR #6842, opened 2019-06-22
Enforce translations

RussKie:
## Proposed changes

- Update/cleanup appveyor.yml
- Enforce update to translations as part of a PR

I've tweaked the script to do the following:

* stage *.xlf files - this way we ignore EOL issues,
* check if there are any xlf files are modified, and if so:
    - capture and publish the diff for the xlf files as git-diff-xlf.txt
    - break the build
image


pmiossec:
Could we fix the ` .\Build\transifex\UpdateLocalEnglishTranslations.ps1` script in this PR?

I think there is a `popd` command in the script.
When I run the script, the current directory is changed to be `.\Build\transifex\` and that is boring.
I would like the script to not change the current directory (to be able to launch the script multiple times without being obliged to return to the git repo root folder).

RussKie:
> I think there is a `popd` command in the script.
> When I run the script, the current directory is changed to be `.\Build\transifex\` and that is boring.

Sorry, you've lost me here...
The script changes its working directory and, when it is done, changes it back to the original.
Why is this a problem?

pmiossec:
>The script changes its working directory and, when it is done, changes it back to the original.
>Why is this a problem?

The problem is that it is also what I want but it is not working for me.
Adding a 'popd' at the end of the script fixes the bad behavior.

RussKie:
Sorry, I have just realised, the script pushd's twice, and popd's once...
Thanks for pointing that out!

RussKie:
> Howcome there is a Appveyor branch build too?

I think this is because I (accidentally) pushed to the main trunk and not into my fork.

pmiossec:
Could the 'Git Status.txt' file contains the 'git diff' of the English.*.xlf files? To better see what is missing?

mstv:
Perhaps a litte bit off topic: Though `GitUI/Translation/English.Plugins.xlf` is out of date and seems to have inconsistent line ends.

mstv:
`Build\transifex\UpdateLocalEnglishTranslations.ps1` produces the inconsistent line ends.

RussKie:
I've tweaked the script to do the following:
- stage *.xlf files - this way we ignore EOL issues, 
- check if there are any xlf files are modified, and if so:
  - capture and publish the diff for the xlf files as `git-diff-xlf.txt`
  - break the build

Issue #6183, opened 2019-01-29
UI: dialogs layout alignment


RussKie:
## Current behaviour

We do not currently have a consistent UI guidelines, that led to a disparate UI layouts of our forms and dialogs.
Overall we are in a pretty sad state...

## Expected behaviour

A consistent UI layouts and design guidelines.

### Current state

Below is the list of forms and dialogs I could reach via the top level menus.
NB: all plugins forms are excluded.

Discussion about margins and paddings: https://github.com/gitextensions/gitextensions/pull/8409

#### Updated:

<details>
<summary>FormInit</summary>

Old:
![image](https://user-images.githubusercontent.com/4403806/51907490-136ae400-241b-11e9-9230-71f6b365bb05.png)

New:

![image](https://user-images.githubusercontent.com/4403806/86530763-e6b7d780-befe-11ea-8847-535677aa0970.png)
![image](https://user-images.githubusercontent.com/4403806/86530759-e0c1f680-befe-11ea-93cb-19e0b6617aa7.png)

</details>

<details>
<summary>FormClone</summary>

Old:
![image](https://user-images.githubusercontent.com/4403806/51907665-9855fd80-241b-11e9-80db-e07c38b184b8.png)

New:
![image](https://user-images.githubusercontent.com/4403806/86530769-f46d5d00-befe-11ea-8afe-73c8b4350ab9.png)
![image](https://user-images.githubusercontent.com/4403806/86530765-f0413f80-befe-11ea-9a4f-7e32d3ba2117.png)

</details>

<details>
<summary>FormSatus / FormProcess</summary>

Old:
???

New:
![ge-status-dark](https://user-images.githubusercontent.com/4403806/87222608-6854ad00-c3b8-11ea-93c4-9dcd792466d4.gif)
![ge-status-light](https://user-images.githubusercontent.com/4403806/87222609-6a1e7080-c3b8-11ea-948c-a1c619a016ee.gif)

</details>

<details>
<summary>FormCheckoutBranch</summary>

Old:

![image](https://user-images.githubusercontent.com/4403806/51907985-93457e00-241c-11e9-80c7-0b5dc66652d9.png)
![image](https://user-images.githubusercontent.com/4403806/51908000-a0fb0380-241c-11e9-8a84-9dc4e6283551.png)

New:
![image](https://user-images.githubusercontent.com/4403806/88481892-62a6bc00-cfa1-11ea-9f25-28016b55356b.png)
[...]

PR #6860, opened 2019-06-25
FormCreateBranch: Improve UX of CheckBox "Checkout after create"

mstv:
## Proposed changes

make all checkboxes of the form work the same standard way
by setting CheckBox.Text instead of a separate Label
in order to indicate the focus and to make the hotkey work directly 

(I know this won't win a design price. Though I think it's not worth breaking the table layout and trying to align a left-labeled CheckBox under DPI scaling.)

### Before

`Checkout after create` _is_ focused:

grafik

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 7b273006d0221522e6fb0a5e6097afbcf777965b
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
@@ -41,7 +41,6 @@
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
             this.groupBox1 = new System.Windows.Forms.GroupBox();
             this.flowLayoutPanel1 = new System.Windows.Forms.FlowLayoutPanel();
-            this.CheckoutAfterCreate = new System.Windows.Forms.Label();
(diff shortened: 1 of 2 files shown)


mstv:
Ah, snap! But I did not change any text. Yes, moving counts, too...

RussKie:
There are few other similar dialogs. Are they suffering from the same
issue? If not, what is the problem with this dialog?

If we are making this change we must review all other dialogs and align
them too.
Please search for an open issue where I posted screenshots for all dialogs
we have.

RussKie:
Here's the issue I mentioned earlier - https://github.com/gitextensions/gitextensions/issues/6183

It looks like all checkbox controls have their labels to the right, so this dialog was the odd one.

Issue #1707, opened 2013-03-10
Settings dialog reordering


feinstaub:
Coming from #1692 here a first suggestions to reorder the settings dialog pages to be more consistent, starting with the settings pages tree structure:

Suggestion 1:
- Git Extensions
  - Git Integration
    - SSH
  - Global Git Settings
  - Local Git Settings
  - Performance
  - UI Options [currently: "Git Extensions"]
    - Appearance
    - Colors
    - Start Page
    - Advanced
    - Confirmations
  - Hotkeys
  - Scripts
  - Shell extension
- Plugins
  - ...

GintasS:
I think the settings window are pretty good now in 2021.

RussKie:
Personally I feel there are further improvements can be made. I'd leave it
open.

PR #6881, opened 2019-06-30
refactor: Move email settings from General to Detailed

RussKie:
## Proposed changes

- Move email settings from `General` to `Detailed` 

### Before

* image

### After

* image
* image

----


gerhardol:
With thais change, the size of the settings window could have been decreased. Maybe split performance and behavior too.

RussKie:
> Maybe split performance and behavior too.

We'll have nothing left on the page then 😄 

There was a good discussion (way before my time) https://github.com/gitextensions/gitextensions/issues/1707 that discussed Settings.
The discussion eventually derailed into a different direction, but has a number of interesting points.

It is worth having a discussion again.

gerhardol:
The settings have been cleaned up since that discussion but it could be discussed again. 
The outcome of that discussion I see that is directly applicable is to a plug General one move Start page to the dashboard.

Issue #6892, opened 2019-07-05
File history issue in a repository with an enabled sparse checkout


montoner0:
## Current behaviour

File history doesn't load (infinite spinner) for files that weren't checked out due to enabled sparse checkout.

## Expected behaviour

File history loaded successfully for such files.

## Steps to reproduce

Create a repository with enabled sparse checkout. E. g.:
```
git init
git remote add origin https://github.com/gitextensions/gitextensions.git
git config core.sparsecheckout true
echo "Setup/*" >> .git/info/sparse-checkout
git pull origin master
```
Then open it with Git Extensions, select any commit, click `File tree` and try to view the file history of any file outside of `Setup` folder.

## Screenshots

Nothing special, actually, `File history` window with a spinner.
![image](https://user-images.githubusercontent.com/13542804/60710699-f7182980-9f1b-11e9-817c-81ed9862b8cc.png)

## Did this work in previous version of GitExtensions

Not tested.

## Environment

- Git Extensions 3.1.1.6049
- Build 2f872105b4eb42421d9fdafd68e90a9e27d20d96
- Git 2.21.0.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

## Diagnostics

Command log doesn't show anything remarkable:
![image](https://user-images.githubusercontent.com/13542804/60710521-90930b80-9f1b-11e9-9155-a11e1067a926.png)

pmiossec:
That is another side effect of #6458.
The current code is allowing to view file history or blame only files that are in the working directory. See https://github.com/gitextensions/gitextensions/issues/6458#issuecomment-481644007

The problem is in `FormFileHistoryController.TryGetExactPath() `

gerhardol:
Closed by #6909

Issue #6815, opened 2019-06-17
Unable to blame a file


pmiossec:
## Current behaviour

Blaming the file `GitHub3Plugin.cs` in the GitExtensions sources open the File History/Blame form displaying the message "This repository does not yet contain any commits". The blame depends of the case of the folder in the working directory.

![image](https://user-images.githubusercontent.com/460196/59609747-42f36200-9118-11e9-9786-9798d4fde175.png)

It is due to a problem with case sensitivity in the `GitHub3` path because there is 2 folders in the filetree (and the blame in the folder `Github3` works but don't work for all the files of  `GitHub3`):
![image](https://user-images.githubusercontent.com/460196/59609829-6dddb600-9118-11e9-9b0e-da0f54d9dab1.png)

Setting `core.ignorecase`  is set at `true` in the repository.

edit: after a better test, the works if the folder name match the case of the folder in the working directory. If the case of the folder is changed to match `GitHub3`, then the blame work for the `GitHub3Plugin.cs` file and fail for files of the other folder.

## Expected behaviour

Blame is working for this file.

## Did this work in previous version of GitExtensions

i don't know...

## Environment

- Git Extensions 3.2.0
- Build 6767731f758d2b45bbdd8125fada5b564d5f46d2
- Git 2.21.0.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

pmiossec:
It seems that `FormFileHistoryController.TryGetExactPath()` is not doing the job.
Especially `NativeMethods.GetLongPathName()`...

crypto-rsa:
Could it be related to #6458?

pmiossec:
>Could it be related to #6458?

Not exactly even if the problem lie in the same `TryGetExactPath()` method.

The problem is that, maybe it is returning a good cased filename but it don't fix the case of the entire path, so the blame fails.

pmiossec:
After a short investigation, `FormFileHistoryController.TryGetExactPath()` is doing the opposite of what it is expecting to do: the path in the input is well cased and the output is not well cased 😭

gerhardol:
Closed by #6909

PR #6893, opened 2019-07-05
Fix FileHistory/Blame for files not existing in the working directory

### LINKED ISSUE #6458: History not shown for a new file in a different branch
## Current behaviour
When trying to show the history of a file which has been added in a different branch than the current one, nothing is shown.

## Expected behaviour
The history is shown regardless of whether the file actually exists in the working directory.

## Steps to reproduce
1. Create a new branch
2. Create a file `new.txt`
3. Change the contents of the file in a few commits
4. Switch to a different branch
5. Open the history form for `new.txt`

Yes, before commit 08737aa147bcc6e5917f370910ca64a4a47bb6a7. There, the full path is retrieved through `FormFileHistoryController.TryGetExactPath`. However, it returns `null` if the given file does not actually exist - which obviously fails for new files from a different branch.

## Environment

- Git Extensions 3.1.0
- Build 9ff970ca36007056992f90db1dcc0f46edcc5fd4 (the last one where it works)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 96dpi (no scaling)

pmiossec:
Fixes #6458 and #6892

## Proposed changes

The bad behavior comes from the fact that `TryGetExactPath()`
overwrite the out variable `fileName` with null
when return is false.
We should use another temporary variable.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 8fee24f359f4422a9b038775fce2e13d4f748d0d (Dirty)
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -210,9 +210,9 @@ namespace GitUI.CommandsDialogs
                 // we will need this later to look up proper casing for the file
                 var fullFilePath = _fullPathResolver.Resolve(fileName);
 
-                if (_controller.TryGetExactPath(fullFilePath, out fileName))
+                if (_controller.TryGetExactPath(fullFilePath, out var exactFileName))
                 {
-                    fileName = fileName.Substring(Module.WorkingDir.Length);
+                    fileName = exactFileName.Substring(Module.WorkingDir.Length);
                 }
 
                 // Replace windows path separator to Linux path separator.
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -1,10 +1,8 @@
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Windows.Forms;
 using GitCommands;
@@ -26,7 +24,7 @@ namespace GitUI.CommandsDialogs
         private readonly FilterBranchHelper _filterBranchHelper;
         private readonly FormBrowseMenus _formBrowseMenus;
         private readonly IFullPathResolver _fullPathResolver;
-        private readonly FormFileHistoryController _controller = new FormFileHistoryController();
+        private readonly FormFileHistoryController _controller;
 
         private BuildReportTabPageExtension _buildReportTabPageExtension;
 
@@ -55,6 +53,7 @@ namespace GitUI.CommandsDialogs
 
             _commitDataManager = new CommitDataManager(() => Module);
             _fullPathResolver = new FullPathResolver(() => Module.WorkingDir);
+            _controller = new FormFileHistoryController(() => Module, _fullPathResolver);
 
             CommitDiff.EscapePressed += Close;
             View.EscapePressed += Close;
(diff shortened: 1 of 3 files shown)


pmiossec:
Even if I don't really know if 'TryGetExactPath()' is doing it's office...

I discovered with https://github.com/gitextensions/gitextensions/issues/6815#issuecomment-508713050, that the method is not working all the time well.

In this special case, where the casing is good because we got it from the file tree, the method that have for purpose to fix the casing, is to the contrary, returning a bad cased path....

My conclusion is that it should be either removed or  called only when the input path is provided from places where we are not sure of the case accuracy, for example, the command line.

RussKie:
Let's change the `BuildFilter` method from local to private and expose it via a `TestAccessor` (and `FileName` too). 
Then add a simple test for the following codepath (we can add tests for other code paths later):
```
            var fileName = FileName;

            // we will need this later to look up proper casing for the file
            var fullFilePath = _fullPathResolver.Resolve(fileName);

            if (_controller.TryGetExactPath(fullFilePath, out fileName))
            {
                fileName = fileName.Substring(Module.WorkingDir.Length);
            }

            // Replace windows path separator to Linux path separator.
            // This is needed to keep the file history working when started from file tree in
            // browse dialog.
            FileName = fileName.ToPosixPath();
```

RussKie:
@pmiossec I pushed a refactoring idea, to make the code testable.
Could you please have a look and confirm whether it is functionally equal to the original code.

gerhardol:
> I discovered with #6815 (comment), that the method is not working all the time well.

For 3.2, I propose this is merged.

For next release, I propose the call to TryGetExactPath() in FormFileHistory is removed and added to
GitUICommands.cs RunFileHistoryCommand(), the only "external" source for filenames.
Files in GitHub3 from the command line will still fail though...
This change could be done already for 3.2 if we are sure this is sufficient.

pmiossec:
@RussKie It seems ok. I have added tests (and also refactored the one on `TryGetExactPathName()`).

>This change could be done already for 3.2 if we are sure this is sufficient.

@gerhardol I have the same point of view on how to solve the problem.
And I'm not confident enough about not introducing bad side effects so I let the fix for a next release...

gerhardol:
I approved the first commit. I am not sure after the rewrite, not sure if I see this as an improvement. Also most of the tests and changes should be removed with the proper fix: Remove TryGetExactPath from FormFileHistory*.cs

My first vote is to merge only the first commit here for 3.2.
Second alternative for 3.2 is to remove TryGetExactPath - this alternative feels better and better the more I look at it.
Third is to merge as is.
But I will not withdraw approval.

Post 3.2 is another discussion

RussKie:
Feel free to take the first into 3.2 and then merge the rest after the 3.2 cut

gerhardol:
> I have the same point of view on how to solve the problem.
And I'm not confident enough about not introducing bad side effects so I let the fix for a next release...

I can only see that the only call path where a path is not a GitItem is for GitUICommands.cs RunFileHistoryCommand()
I feel confident enough to suggest that this should be the solution in 3.2
The refactoring and added tests can be added to 3.3 (some refactoring needed).
It may be OK to skip updating RunFileHistoryCommand() in 3.2, the change still fixes the case where the path do not exist.

@pmiossec 
Since you done the investigation here, do you want to submit a PR for 3.2?

pmiossec:
I would try to do it but most probably not before 2 or 3 days...

gerhardol:
> I would try to do it but most probably not before 2 or 3 days...

You can copy the commit I did, or I submit the pr

gerhardol:
This should be WIP and rebased on master with slightly updated testcases after the 3.2 release branch is created.
(This is probably an improvement even if I was negative to merge close to a release.)

RussKie:
Please resolve the merge conflict. We can probably squash it too 🤔

RussKie:
@pmiossec could you please resolve conflicts and squash, and we can take it in the 3.2 release.

pmiossec:
For me, this PR is no more needed since the merge of #6909

gerhardol:
@pmiossec 
I believe you did some refactoring here. It is up to you if that is something that should be included in master

pmiossec:
No, I won't work on that. Maybe later.
I have some more priority changes to include first...

RussKie:
Closing. @pmiossec please re-open or open a new PR when you're ready.

PR #6907, opened 2019-07-09
Stash: Add cherry pick of file's changes

### LINKED ISSUE #6902: Apply stash to some files
## Feature description

From the stash window, I want to apply only one of the diffs.
For example I might have the files `foo.x` and `bar.x` in the stash. I expected the right-click menu of `foo.x` to have an option to apply the diff to `foo.x` without touching `bar.x`.

In the command line this is pretty easy to do, in the example it would be `git checkout stash@{0} -- foo.x`

## Environment

- Git Extensions 3.1.1.6049
- Git 2.21.0.windows.1
- .NET Framework 4.8.3801.0
- DPI 120dpi (125% scaling)

pmiossec:
Fixes #6902

## Proposed changes

- In the file list of a stash, add a contextual menu that allows the cherry pick of a file's changes.

### Before

nothing

### After

image

PS: I don't know why but that also add a 'Sort by' menu item... but that options works, so I didn't investigated on that subject.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 12ee015dbb06e3fbff2752d3bc12fc2ceea5f87b (Dirty)
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormStash.Designer.cs
+++ b/GitUI/CommandsDialogs/FormStash.Designer.cs
@@ -48,6 +48,8 @@ namespace GitUI.CommandsDialogs
             this.panel1 = new System.Windows.Forms.Panel();
             this.Loading = new LoadingControl();
             this.Stashed = new GitUI.FileStatusList();
+            this.contextMenuStripStashedFiles = new System.Windows.Forms.ContextMenuStrip(this.components);
+            this.cherryPickFileChangesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStrip1 = new GitUI.ToolStripEx();
             this.showToolStripLabel = new System.Windows.Forms.ToolStripLabel();
             this.Stashes = new System.Windows.Forms.ToolStripComboBox();
@@ -65,6 +67,7 @@ namespace GitUI.CommandsDialogs
             this.tableLayoutPanel1.SuspendLayout();
             this.panel1.SuspendLayout();
             this.toolStrip1.SuspendLayout();
+            this.contextMenuStripStashedFiles.SuspendLayout();
             this.SuspendLayout();
             // 
             // gitStashBindingSource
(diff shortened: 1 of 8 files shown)


pmiossec:
>If you select several files, the changes for the (last) selected is applied only which could be confusing.
>Similar if no file is selected.

Thanks for the heads up. I should have thought about it but didn't because I was focus on something else.

I think it makes no sense to multi-select so I think I will disable this feature. And manage also when nothing is selected.

GitUI/CommandsDialogs/FormStash.Designer.cs
```diff
@@ -269,6 +273,21 @@ private void InitializeComponent()
...
+            this.cherryPickFilesChangesToolStripMenuItem});
+            this.contextMenuStripStashedFiles.Name = "contextMenuStripStashedFiles";
+            this.contextMenuStripStashedFiles.Size = new System.Drawing.Size(209, 76);
+            // 
+            // cherryPickFilesChangesToolStripMenuItem
+            // 
+            this.cherryPickFilesChangesToolStripMenuItem.Image = global::GitUI.Properties.Resources.CherryPick;
+            this.cherryPickFilesChangesToolStripMenuItem.Name = "cherryPickFilesChangesToolStripMenuItem";
+            this.cherryPickFilesChangesToolStripMenuItem.Size = new System.Drawing.Size(208, 22);
+            this.cherryPickFilesChangesToolStripMenuItem.Text = "Cherry pick file\'s changes";
```
crypto-rsa: Just a nitpick, but I'd prefer **not** including the _'s_ in _file's_. It seems to me that the possessive marker is generally used for "live" objects (like _dog's_, _girl's_, _John's_ etc.). However, I am not a native speaker so I might be wrong.

RussKie:
You're absolutely right - https://en.m.wikipedia.org/wiki/Possessive (I had
to do some reading)
Thank you

GitUI/CommandsDialogs/FormStash.Designer.cs
pmiossec: >Just a nitpick, but I'd prefer not including the 's in file's.

I used the same text as the one already used elsewhere.
Is everyone ok to fix **all** the strings be part of this PR?

RussKie:
Would you mind doing it in a separate PR please?

pmiossec:
@gerhardol done

RussKie:
Thank you

PR #6909, opened 2019-07-10
Fix different problems around FileHistory/Blame

### LINKED ISSUE #6458: History not shown for a new file in a different branch

pmiossec:
Fixes #6458, #6892 and #6815

Alternative for #6893 (like discussed in https://github.com/gitextensions/gitextensions/pull/6893#issuecomment-509390885 )

## Proposed changes

- Fix FileHistory/Blame for files not existing in the working directory

The bad behavior comes from the fact that `TryGetExactPath()`
changes the path to the file system path (if not existing, the path is cleared)
- Improve readability of unit tests on `TryGetExactPath()`

/cc @gerhardol  @RussKie

### DIFF
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -207,20 +207,13 @@ namespace GitUI.CommandsDialogs
             {
                 var fileName = FileName;
 
-                // we will need this later to look up proper casing for the file
-                var fullFilePath = _fullPathResolver.Resolve(fileName);
-
-                if (_controller.TryGetExactPath(fullFilePath, out fileName))
-                {
-                    fileName = fileName.Substring(Module.WorkingDir.Length);
-                }
-
                 // Replace windows path separator to Linux path separator.
                 // This is needed to keep the file history working when started from file tree in
                 // browse dialog.
                 FileName = fileName.ToPosixPath();
 
                 var res = (revision: (string)null, path: $" \"{fileName}\"");
+                var fullFilePath = _fullPathResolver.Resolve(fileName);
 
                 if (AppSettings.FollowRenamesInFileHistory && !Directory.Exists(fullFilePath))
                 {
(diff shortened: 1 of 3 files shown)


pmiossec:
Thanks  @gerhardol for the commit "ready made"! 😉 

I have included the commit where I have improved the readability of tests on TryGetExactPath() because I think it can't hurt 😉

gerhardol:
> Alternative for #6893

The refactoring and new tests in #6893 are probably OK, even if I am a little hesitant to approve the changes close to a release (my vote is not the most important, but I still raise my opinion).
That PR should be rebased on master after this is merged, for 3.3.

pmiossec:
This PR is better than #6893 because it fixes 3 issues where #6893 only fixes 2...

RussKie:
Merge at will

gerhardol:
One note regarding the change: It does not fix #6815 when starting from the commandline if the path is not case correct, for instance for Plugins\Github3\GitHubRepo.cs (Should be Plugins\GitHub3\GitHubRepo.cs)
That is a limitation I believe we can live with, you can view the history from the application.

pmiossec:
>One note regarding the change: It does not fix #6815 when starting from the commandline if the path is not case correct

I don't know if it could be interesting to mitigate the issue by looking for the good case path in the filetree.

PR #6913, opened 2019-07-11
fix: Commit message incorrect height at 150% scale factor

### LINKED ISSUE #6898: Commit message  incorrect height at 150% scale factor
## Current behaviour

A regression (probably caused by me) in calculation of the commit message box in a scaled environment:
❗️ **This is only happening at 150% scale factor!**
image

## Expected behaviour

Commit message must look the same at 150% scale factor.

## Steps to reproduce

Set Windows to 150% and run the app.

## Screenshots

I have tested all other scale factors available to me:

* 100%
image

* 125%
image

* 175%
image

* 200%
image

* 225%
image

* 250%
image

Yes, because the commit message had a newline appended that caused an additional margin at the bottom.

RussKie:
Closes #6898

## Proposed changes

-  Absent of an obvious way to fix it, hardcode the fix - add an extra space at the bottom

### Before

image

### After

image

## Test methodology 

- manual testing at various scale factors

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -14,6 +14,7 @@ using GitCommands.ExternalLinks;
 using GitCommands.Git;
 using GitCommands.Remotes;
 using GitExtUtils;
+using GitExtUtils.GitUI;
 using GitUI.CommandsDialogs;
 using GitUI.Editor.RichTextBoxExtension;
 using GitUI.Hotkey;
@@ -751,6 +752,17 @@ namespace GitUI.CommitInfo
         private void CommitMessage_ContentsResized(ContentsResizedEventArgs e)
         {
             _commitMessageHeight = e.NewRectangle.Height;
+
+            // at scale factor of 150% there is rendering artifact - the last line is almost lost
+            // at all other scaling factors from 100% to 350% everything is rendered as expected
+            // see https://github.com/gitextensions/gitextensions/issues/6898 for more details
+            //
+            // absent of an obvious way to fix it, hardcode the fix - add an extra space at the bottom
+            if (DpiUtil.ScaleX - 1.5f < float.Epsilon)
+            {
+                _commitMessageHeight += 16;
+            }
+
             PerformLayout();
         }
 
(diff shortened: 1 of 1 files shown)


RussKie:
@NikolayXHD

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -751,6 +752,17 @@ private void RevisionInfo_ContentsResized(ContentsResizedEventArgs e)
...
             _commitMessageHeight = e.NewRectangle.Height;
+
+            // at scale factor of 150% there is rendering artifact - the last line is almost lost
+            // at all other scaling factors from 100% to 350% everything is rendered as expected
+            // see https://github.com/gitextensions/gitextensions/issues/6898 for more details
+            //
+            // absent of an obvious way to fix it, hardcode the fix - add an extra space at the bottom
+            if (DpiUtil.ScaleX - 1.5f < float.Epsilon)
+            {
+                _commitMessageHeight += 16;
```
NikolayXHD: Maybe measure actual string height instead?
```
using (var g = Graphics.FromHwnd(rtbxCommitMessage.Handle))
  rowHeight = (int)g.MeasureString("A", rtbxCommitMessage.Font).Height
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -751,6 +752,17 @@ private void RevisionInfo_ContentsResized(ContentsResizedEventArgs e)
         private void CommitMessage_ContentsResized(ContentsResizedEventArgs e)
         {
             _commitMessageHeight = e.NewRectangle.Height;
+
+            // at scale factor of 150% there is rendering artifact - the last line is almost lost
+            // at all other scaling factors from 100% to 350% everything is rendered as expected
+            // see https://github.com/gitextensions/gitextensions/issues/6898 for more details
+            //
+            // absent of an obvious way to fix it, hardcode the fix - add an extra space at the bottom
+            if (DpiUtil.ScaleX - 1.5f < float.Epsilon)
```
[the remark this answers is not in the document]
RussKie: Thank you, you're absolutely correct

GitUI/CommitInfo/CommitInfo.cs
RussKie: I've tried increasing the height of the box by a height of a measured string, as you suggested.
However the effect wasn't particularly eye pleasing - a measured string was 23 pixels.

Another thing, we would consume CPU cycles and memory to measure a string every time we display a commit info.

PR #6769, opened 2019-06-09
Set recommended Git version to 2.22.0

gerhardol:
## Proposed changes
Set recommended version to 2.22

## Test methodology 

- Review of Release Notes.  
No major changes in this release, similar to the stable 2.21.0. 
git-diff --no-index --dir-diff is no longer allowed, not used together in ST.

- Use of Git - to be in use for week or so before removing WIP

## Test environment(s) 

n/a

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #6920, opened 2019-07-13
Add support for "--rebase-merges" for newest version of git

pmiossec:
now that we recommend v2.22 (#6769) that deprecate "--preserve-merges"

See:
* for option deprection: https://github.com/git/git/commit/fa1b86e45743fd5895c33adcd3769782e608bb40#diff-c7361e406139e8cd3a300b80b8f8cc8dR1220
* for some doc: https://stackoverflow.com/questions/15915430/what-exactly-does-gits-rebase-preserve-merges-do-and-why/50555740#50555740

## Test methodology 

- Manual (not heavily)
- Updated unit tests

## Test environment(s) 

- Git Extensions 3.2.0
- Build 8ee3620d3fa4b960b24240cb5ccf1822c01a33f7
- Git 2.20.1.windows.1 (recommended: 2.22.0 or later)
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

### DIFF
--- a/UnitTests/GitCommandsTests/Git/GitCommandHelpersTest.cs
+++ b/UnitTests/GitCommandsTests/Git/GitCommandHelpersTest.cs
@@ -605,7 +605,7 @@ namespace GitCommandsTests.Git
                 "-c rebase.autoSquash=false rebase -i --no-autosquash \"branch\"",
                 GitCommandHelpers.RebaseCmd("branch", interactive: true, preserveMerges: false, autosquash: false, autoStash: false).Arguments);
             Assert.AreEqual(
-                "-c rebase.autoSquash=false rebase --preserve-merges \"branch\"",
+                GitVersion.Current.SupportRebaseMerges ? "-c rebase.autoSquash=false rebase --rebase-merges \"branch\"" : "-c rebase.autoSquash=false rebase --preserve-merges \"branch\"",
                 GitCommandHelpers.RebaseCmd("branch", interactive: false, preserveMerges: true, autosquash: false, autoStash: false).Arguments);
             Assert.AreEqual(
                 "-c rebase.autoSquash=false rebase \"branch\"",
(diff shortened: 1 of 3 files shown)


UnitTests/GitCommandsTests/Git/GitCommandHelpersTest.cs
```diff
@@ -605,7 +605,7 @@ public void RebaseCmd()
                 "-c rebase.autoSquash=false rebase -i --no-autosquash \"branch\"",
                 GitCommandHelpers.RebaseCmd("branch", interactive: true, preserveMerges: false, autosquash: false, autoStash: false).Arguments);
             Assert.AreEqual(
-                "-c rebase.autoSquash=false rebase --preserve-merges \"branch\"",
+                "-c rebase.autoSquash=false rebase --rebase-merges \"branch\"",
```
RussKie: Tests should pivot on git version.
So if someone runs a codebase against an older version of git (like me while testing) they shouldn't get a test failure.

UnitTests/GitCommandsTests/Git/GitCommandHelpersTest.cs
```diff
@@ -617,7 +617,7 @@ public void RebaseCmd()
                 "-c rebase.autoSquash=false rebase -i --autosquash \"branch\"",
                 GitCommandHelpers.RebaseCmd("branch", interactive: true, preserveMerges: false, autosquash: true, autoStash: false).Arguments);
             Assert.AreEqual(
-                "-c rebase.autoSquash=false rebase -i --autosquash --preserve-merges --autostash \"branch\"",
+                "-c rebase.autoSquash=false rebase -i --autosquash --rebase-merges --autostash \"branch\"",
```
RussKie: ditto

appveyor.yml
```diff
@@ -1,6 +1,6 @@
 version: 3.2.0.{build}
```
RussKie: Let's submit this change separately, as it is orthogonal to the main change

pmiossec: that was a last test (before going to bed ;) ) to gather information, not planned to be merged...

pmiossec:
done.

UnitTests/GitCommandsTests/Git/GitCommandHelpersTest.cs
gerhardol: Note: We have a few tests for features not in the oldest supported version, but those tests do not use Git binaries (only argument tests).

gerhardol:
We could include this in 3.2

RussKie:
Thank you

PR #6949, opened 2019-07-27
ReleaseNotesGenerator plugin To defaults to "HEAD"

### LINKED ISSUE #4444: Release Note Generator - Validation Bug
I think this is a bug.

In Git Extensions V2.50.02 there was no validation on the release note generator "From" and "To" fields. This meant you could put something in the "From" fields but leave the "To" field blank and it would show commit history up to head.
In V2.51 validation has been added so that both fields have to be filled. in. This means that to replicate the same report you have to enter "head" in the "To" field.

Can we remove validation on the "To" field please.

Whilst on the surface this seems a minor irritation, in my particular use case I manually go through dozens of git repositories running this to get release notes and it is a major slow down for me.

manuelspezzani:
Fixes #4444

## Proposed changes

The ReleaseNotesGenerator is initialized with a default "HEAD" text for the To field.
Another option could have been to completely remove the To field validation (blank is an "alias" for HEAD) but IMHO having an explicit default value provides a more clear/usable experience.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/Plugins/ReleaseNotesGenerator/ReleaseNotesGeneratorForm.Designer.cs
+++ b/Plugins/ReleaseNotesGenerator/ReleaseNotesGeneratorForm.Designer.cs
@@ -30,7 +30,7 @@
         {
             this.label1 = new System.Windows.Forms.Label();
             this.label2 = new System.Windows.Forms.Label();
-            this.textBoxRevTo = new System.Windows.Forms.TextBox();
+            this._NO_TRANSLATE_textBoxRevTo = new System.Windows.Forms.TextBox();
             this.textBoxRevFrom = new System.Windows.Forms.TextBox();
             this.label3 = new System.Windows.Forms.Label();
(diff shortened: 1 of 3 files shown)


gerhardol:
The field must be renamed with a NOTRANSLATE prefix to not be translated.

gerhardol:
Need to be rebased, or even better to submit the contributors.txt change in a separate PR, then rebase

gerhardol:
(to simplify merge at a later stage it be easier if you rebase on gitextensions/master so GitHub do not see any merge conflicts (as it is based on commits merged to master, GH may be OK, but it is easier to fix now)).

RussKie:
Thank you

PR #6954, opened 2019-07-27
Fix/search unmerged branches

### LINKED ISSUE #6813: Plugin "Delete obsolete branches" include unmerged branches give error "fatal: 'HEAD' is not a valid branch name."
When trying to find unmerged branches in the plugin, get an error 
"fatal: 'HEAD' is not a valid branch name.".

I have noticed this in 3.1.1.6049. Not sure of other 3.x versions. This feature used to work in 2.51.05.

Steps to reproduce:
1) Open the plugin "Delete obsolete branches"
2) Click on "Include unmerged branches"
3) Click on on the caution dialog
4) Click on "Search branches"

Expected result:
Should be able to get unmerged branches based on settings and filters.

Actual result:
Get an error "fatal: 'HEAD' is not a valid branch name." in the Name column.

manuelspezzani:
Fixes #6813

## Proposed changes

- added a "--list" option to the issued "git branch" command to prevent unexpected behaviors (e.g. creating a new branch, see comments on #6813)
- if the command fails a message box is displayed 

## Screenshots 
This is what happens in case of error:

error message

## Test methodology 
- Manual test

## Test environment(s) 
- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
+++ b/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
@@ -34,6 +34,7 @@ namespace DeleteUnusedBranches
         private readonly IGitModule _gitCommands;
         private readonly IGitUICommands _gitUiCommands;
         private readonly IGitPlugin _gitPlugin;
+        private readonly GitBranchOutputCommandParser _commandOutputParser;
         private CancellationTokenSource _refreshCancellation;
 
         public DeleteUnusedBranchesForm(DeleteUnusedBranchesFormSettings settings, IGitModule gitCommands, IGitUICommands gitUiCommands, IGitPlugin gitPlugin)
@@ -42,6 +43,7 @@ namespace DeleteUnusedBranches
             _gitCommands = gitCommands;
             _gitUiCommands = gitUiCommands;
             _gitPlugin = gitPlugin;
+            _commandOutputParser = new GitBranchOutputCommandParser();
 
             InitializeComponent();
 
@@ -87,7 +89,7 @@ namespace DeleteUnusedBranches
             BranchesGrid.DataSource = _branches;
         }
 
-        private static IEnumerable<Branch> GetObsoleteBranches(RefreshContext context, string curBranch)
+        private IEnumerable<Branch> GetObsoleteBranches(RefreshContext context, string curBranch)
         {
             foreach (string branchName in GetObsoleteBranchNames(context, curBranch))
             {
@@ -108,7 +110,7 @@ namespace DeleteUnusedBranches
             }
         }
 
-        private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context, string curBranch)
+        private IEnumerable<string> GetObsoleteBranchNames(RefreshContext context, string curBranch)
         {
             RegexOptions options;
             if (context.RegexIgnoreCase)
@@ -125,21 +127,24 @@ namespace DeleteUnusedBranches
 
             var args = new GitArgumentBuilder("branch")
             {
+                 "--list",
                  { context.IncludeRemotes, "-r" },
                  { !context.IncludeUnmerged, "--merged" },
                  context.ReferenceBranch
             };
 
-            string[] branches = context.Commands.GitExecutable.GetOutput(args)
-                .Split('\n')
-                .Where(branchName => !string.IsNullOrEmpty(branchName))
-                .Select(branchName => branchName.Trim('*', ' ', '\n', '\r'))
-                .Where(branchName => branchName != "HEAD" && branchName != curBranch &&
-                                   branchName != context.ReferenceBranch)
-                .ToArray();
+            var result = context.Commands.GitExecutable.Execute(args);
 
-            return branches.Where(branchName => (!context.IncludeRemotes || branchName.StartsWith(context.RemoteRepositoryName + "/")) &&
-                                   (regex == null || regex.IsMatch(branchName) == regexMustMatch)).ToArray();
+            if (!result.ExitedSuccessfully)
+            {
+                MessageBox.Show(this, result.AllOutput, $"git {args}", MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return Array.Empty<string>();
+            }
+
+            return _commandOutputParser.GetBranchNames(result.AllOutput)
+                                        .Where(branchName => branchName != curBranch && branchName != context.ReferenceBranch)
+                                        .Where(branchName => (!context.IncludeRemotes || branchName.StartsWith(context.RemoteRepositoryName + "/"))
+                                                            && (regex == null || regex.IsMatch(branchName) == regexMustMatch));
         }
 
         private void Delete_Click(object sender, EventArgs e)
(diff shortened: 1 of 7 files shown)


RussKie:
Briefly look on my mobile. I have comments to write.

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -87,7 +89,7 @@ protected override void OnLoad(EventArgs e)
             BranchesGrid.DataSource = _branches;
         }
 
-        private static IEnumerable<Branch> GetObsoleteBranches(RefreshContext context, string curBranch)
+        private IEnumerable<Branch> GetObsoleteBranches(RefreshContext context, string curBranch)
```
RussKie: Why has this changed to an instance method?

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -108,7 +110,7 @@ private static IEnumerable<Branch> GetObsoleteBranches(RefreshContext context, s
             }
         }
 
-        private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context, string curBranch)
+        private IEnumerable<string> GetObsoleteBranchNames(RefreshContext context, string curBranch)
```
RussKie: ditto

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -125,21 +127,26 @@ private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context
...
-                .ToArray();
+            var result = context.Commands.GitExecutable.Execute(args);
 
-            return branches.Where(branchName => (!context.IncludeRemotes || branchName.StartsWith(context.RemoteRepositoryName + "/")) &&
-                                   (regex == null || regex.IsMatch(branchName) == regexMustMatch)).ToArray();
+            if (result.ExitedSuccessfully)
+            {
+                return _commandOutputParser.GetBranchNames(result.AllOutput)
+                                           .Where(branchName => branchName != curBranch && branchName != context.ReferenceBranch)
+                                           .Where(branchName => (!context.IncludeRemotes || branchName.StartsWith(context.RemoteRepositoryName + "/")) && (regex == null || regex.IsMatch(branchName) == regexMustMatch))
```
RussKie: Please wrap to make it readable

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -125,21 +127,26 @@ private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context
...
+                                           .ToArray();
```
RussKie: The method returns `IEnumerable<T>`, and the caller will materialise the sequence.
Why do we need to enumerate and allocate here?

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -125,21 +127,26 @@ private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context
...
+            }
+            else
```
RussKie: `else` is redundant.

Purely as an observation, I prefer to structure the code to have the "happy path" exit as the very last statement. This typically makes the code easier to follow.

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -125,21 +127,26 @@ private static IEnumerable<string> GetObsoleteBranchNames(RefreshContext context
...
+            {
+                MessageBox.Show(this, result.AllOutput, $"git {args}", MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return new string[0];
```
RussKie:
```suggestion
                return Array.Empty<string>();
```

Plugins/DeleteUnusedBranches/GitBranchOutputCommandParser.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections.Generic;
+
+namespace DeleteUnusedBranches
+{
+    public class GitBranchOutputCommandParser
+    {
+        public IEnumerable<string> GetBranchNames(string commandOutput)
+        {
+            if (commandOutput != null)
```
RussKie: Please invert to reduce nesting

Plugins/DeleteUnusedBranches/GitBranchOutputCommandParser.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                foreach (var line in commandOutput.Split('\n'))
+                {
+                    if (!string.IsNullOrWhiteSpace(line))
```
RussKie: Please amalgamate into the `foreach`.

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
manuelspezzani: Because the actual parsing is now performed by the new GitBranchOutputCommandParser: it's a class field so I remove the static clause on both methods.

I could have instantiated the parser as a local variable, but I prefer to instantiate all the collaborators in the constructor. Anyway, having a consistent code base it's more important so just let me know if you prefer to change it and I'll do my best. ;)

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
manuelspezzani: same as the other method

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
manuelspezzani: Honestly? It was materializing before and it's materializing now :)
I had a look at the caller method and you are right: that ToArray is useless. 
I'll fix it.

Plugins/DeleteUnusedBranches/GitBranchOutputCommandParser.cs
manuelspezzani: Can you please elaborate on this?

Do you mean something like:
```
foreach (var line in commandOutput.Split('\n').Where(x => !string.IsNullOrWhiteSpace(x)))
{
 ...
}
```

I don't think it's actually better than the original in terms of readability...

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
RussKie: I missed that `GetObsoleteBranchNames` references an instance member.
Thank you for the clarification.

RussKie:
Please squash and we'll get it merged.
Thank you

RussKie:
Thank you

PR #6957, opened 2019-07-28
improved validation on create new repository dialog

### LINKED ISSUE #6955: Crash on "Create new repository"
## Current behaviour

Entering a random string on the directory field (Create new repository dialog) crashes the application 

2019-07-28 01_39_56-Create new repository

## Expected behaviour

An error is displayed: the directory field doesn't contains a valid uri/path

## Steps to reproduce

- from dashboard, "Create new repository"
- put some random text on the directory field (not an actual folder/path/uri, e.g. "foobar")
- click on Create

Not tried

## Environment

- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

manuelspezzani:
Fixes #6955

## Proposed changes

- on the create new repository dialog I've improved the directory textbox validation in order to prevent the crash described in #6955 

## Test methodology 

- Manual tests
- Added some unit tests for the path validator

## Test environment(s) 

- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormInit.cs
+++ b/GitUI/CommandsDialogs/FormInit.cs
@@ -4,6 +4,7 @@ using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
 using GitCommands.UserRepositoryHistory;
+using GitExtUtils;
 using ResourceManager;
 
 namespace GitUI.CommandsDialogs
@@ -48,19 +49,21 @@ namespace GitUI.CommandsDialogs
 
         private void InitClick(object sender, EventArgs e)
         {
-            if (string.IsNullOrEmpty(Directory.Text))
+            var directoryPath = Directory.Text;
+
+            if (!IsRootedDirectoryPath(directoryPath))
             {
                 MessageBox.Show(this, _chooseDirectory.Text, _chooseDirectoryCaption.Text);
                 return;
             }
 
-            if (File.Exists(Directory.Text))
+            if (File.Exists(directoryPath))
             {
                 MessageBox.Show(this, _chooseDirectoryNotFile.Text, _chooseDirectoryNotFileCaption.Text);
                 return;
             }
 
-            var module = new GitModule(Directory.Text);
+            var module = new GitModule(directoryPath);
 
             if (!System.IO.Directory.Exists(module.WorkingDir))
             {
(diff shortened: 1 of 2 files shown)


GitExtUtils/FileSystemUtils.cs
```diff
@@ -0,0 +1,28 @@
...
+                {
+                    return false;
+                }
+
+                // this is going to throw if it's an invalid path (e.g. contains special chars)
+                var info = new DirectoryInfo(path);
+
+                return Path.IsPathRooted(path.Trim());
+            }
+            catch (Exception)
```
gerhardol: Should this catch on all Exceptions or just the specific I/O exception?

GitUI/CommandsDialogs/FormInit.cs
```diff
@@ -48,19 +49,21 @@ public FormInit(string dir, EventHandler<GitModuleEventArgs> gitModuleChanged)
 
         private void InitClick(object sender, EventArgs e)
         {
-            if (string.IsNullOrEmpty(Directory.Text))
+            var directoryPath = Directory.Text;
+
+            if (FileSystemUtils.IsRootedDirectoryPath(directoryPath) == false)
```
gerhardol: Prefer ! instead of == false

GitExtUtils/FileSystemUtils.cs
manuelspezzani: I got your point and in general it makes a lot of sense.

In this specific case, I'd prefer to use a more generic exception handler: I guess it's because it communicates the fact that, no matter what, this is not expected to throw any exception. And that if anything goes bad, "false" is the result. 

That said, I recognize it's probably more about having a more consistent code base (in terms of coding styles) so just let me know and I'll fix it! ;)

manuelspezzani:
> This requires that a full path is always entered. 

Yes, for two reasons:
- the released version actually don't allow relative paths (it crashes) so I assumed it wasn't an actual use-case
- a relative path can be quite confusing for the end user: I mean, relative to what? 

Anyway, let's discuss about it: maybe I'm just biased by my own way of using GE!! :)

GitExtUtils/FileSystemUtils.cs
gerhardol: The code style varies wildly...
A comment would be good.

gerhardol:
> a relative path can be quite confusing for the end user: I mean, relative to what?

It would have to be prefilled if it exists (second alternative). But this was more a comment to prevent further discussions.

RussKie:
I had a quick look on my mobile, and i want to review it in more detail.
Few things didn't quite look correct to me.

GitExtUtils/FileSystemUtils.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.IO;
+
+namespace GitExtUtils
+{
+    public static class FileSystemUtils
+    {
+        public static bool IsRootedDirectoryPath(string path)
```
RussKie: I don't see any need for a new public API. Not only it serves a very specific use case, it also requires extensive testing (significantly more than targeting the specific case).
Please move this method into `FormInit` as a private method and test if via the `TestAccessor` pattern we use.

GitUI/CommandsDialogs/FormInit.cs
```diff
@@ -98,6 +124,11 @@ public TestAccessor(FormInit form)
             }
 
             public ComboBox DirectoryCombo => _form.Directory;
+
+            public bool IsRootedDirectoryPath(string path)
```
RussKie: [nit] 
```suggestion
            public bool IsRootedDirectoryPath(string path) => FormInit.IsRootedDirectoryPath(path);
```

RussKie:
Thank you

PR #6997, opened 2019-08-08
Replace git-diff patience with histogram

### LINKED ISSUE #6991: Support diff algorithm "histogram"
## Feature description

Git supports four different algorithms in git-diff: 
 * myers (default)
 * minimal (improved myers)
 * patience (try to give contextual diff)
 * histogram (kind of enhanced patience, normally faster too)
https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-algorithmpatienceminimalhistogrammyers
All supported in at least since Git 1.7.7

A comparision of myers vs histogram here:
https://www.researchgate.net/publication/330955835_How_Different_Are_Different_diff_Algorithms_in_Git_Use_--histogram_for_Code_Changes

Example myers vs patience.
https://gist.github.com/roryokane/6f9061d3a60c1ba41237

GE supports --patience argument (added just before histogram was added).
Of course, a custom algorithm can be configured if patience is not ticked:
```$ git config --global --add diff.algorithm histogram```

I have normally used patience as this can be configured in GE.
Cannot see much difference between patience vs histogram (not much tests).

Some ways to support histogram:
 * Just replace patience with histogram. That will give the 'best' experience for most users with a minimal effort (translations mostly)
 * Add a dropdown to select the algorithm. Should probably have a recommendation too.
 * Tooltip etc to explain how to configure this?

I propose the first alternative, to keep it simple.

## Environment
- Git Extensions 3.2.0
- Build gpg: Signature made Sat Aug  3 21:50:59 2019;gpg:                using RSA key ;f9bc9e76f2bee9687f57088d1262d371dd2dae46 (Dirty)
- Git 2.22.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

### PR DESCRIPTION
**Pull request by gerhardol**
Fixes #6991 

## Proposed changes

Use --histogram instead of --patience for the setting.
See issue for motivation.

TBD Is a setting tooltip required?
If so, a proposal
```
Use --histogram for git-diff, replacing the default algorithm as well as the git config diff.algorithm option.
```

### Before

image

### After

image

## Test methodology 
Check the Git Command log that --histogram is used instead of --patience.

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2266,7 +2266,7 @@ namespace GitCommands
             {
                 "--no-color",
                 extraDiffArguments,
-                { AppSettings.UsePatienceDiffAlgorithm, "--patience" },
+                { AppSettings.UseHistogramDiffAlgorithm, "--histogram" },
                 "-M -C",
                 diffOptions
             };
@@ -2577,7 +2577,7 @@ namespace GitCommands
                     "--no-color",
                     { staged, "-M -C --cached" },
                     extraDiffArguments,
-                    { AppSettings.UsePatienceDiffAlgorithm, "--patience" },
+                    { AppSettings.UseHistogramDiffAlgorithm, "--histogram" },
                     "--",
                     fileName.ToPosixPath().Quote(),
                     { staged, oldFileName?.ToPosixPath().Quote() }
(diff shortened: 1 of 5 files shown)


RussKie:
I've been thinking of it - perhaps we should replace it with the dropdown,
and allow users to select; defaulting to the histogram.

gerhardol:
> I've been thinking of it - perhaps we should replace it with the dropdown,
and allow users to select; defaulting to the histogram.

The default should be same as in Git, so default.
That was I was thinking that a dropdown should be used at first too. However, there is no major need to give the user all options here. If you want to have full control, you can use diff.algorithm.
Keep it simple.

As this affects translations, it is questionable to include in 3.1.

RussKie:
Let's avoid putting anything in 3.2 that we aren't quite sure about :)

PR #7004, opened 2019-08-10
TortoiseGitMerge.exe old default name

gerhardol:
## Proposed changes

TortoiseGit changed the name of the Git/Merge executable in v1.8 (2013).
TortoiseMerge.exe -> TortoiseGitMerge.exe

When selecting tmerge as difftool (or TortoiseMerge as mergetool) and clicking Browse, GE searced for the old executable name, changed in 2013.
Using Suggest was OK.

Note: There is still compatibility when selecting Suggest to select tgit <1.8 or TortoiseSVN if no TortoiseGit >=1.8. This could be removed. I added a comment, this could be removed.

### Before

image

### After

image

## Test methodology 
Manual

----


gerhardol:
Tests fail in FormBrowseLeftPanelReorderNodesTest, unrelated

RussKie:
these tests seem to have become unstable of late.
a build needs to be kicked off again until green

PR #7007, opened 2019-08-11
Updated changed github id

KarstenRa:
Update due to github id change


mast-eu:
I re-launched the build on AppVeyor twice. First and second run failed with a problem related to codecov.io (see log for details). Anyhow the problem is not connected to this PR.
The third run on AppVeyor was successful.

mast-eu:
Thank you.

Issue #5919, opened 2018-12-15
mergetool should take the --gui option


gerhardol:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
Git 2.20.0 introduces the --gui option for mergetool, similar to difftools.
https://github.com/git/git/blob/master/Documentation/RelNotes/2.20.0.txt

GE should use this option if Git uses 2.20 or later.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
n/a

**What is the expected behavior?**
Same behavior for difftool and mergetool

**Environment you encounter the issue:**
- Git Extensions 3.00.00.4433
- Build fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

**Did this work in previous version of GitExtensions (which)?**
no

PR #7009, opened 2019-08-11
Remove KDiff3 hardcoding

gerhardol:
Refactoring
Somehow related to #5919

GE has a lot of hardcoded support for KDiff3. That made sense when kdiff was bundled with GE but has really no effect now and is just confusing and makes other changes harder.
There is also some duplication of code.

At startup and in the settings form, GE tries to set kdiff3 if no diif or mergetool is set.
(The situation that kdiff3 is a configured tool and that no tool is set is not a use case to optimize for.)

To let Git handle the tools supported (the list in GE is not same as in Git, also diff like tmerge -> tortoisemerge) and to cleanup is another MR.
(as well as better handle custom tools)

This is WIP as it should not be merged before 3.2
Other changes may be added to the PR

## Proposed changes

If no diff or mergetool, send users to the Git config settings form and let the user select manually.

### Before

image

### After

Directly to:

image

## Test methodology 

Remove diff.tool, diff.guitool and merge.tool in .gitconfig.

## Test environment(s) 

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
@@ -72,13 +72,6 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         private readonly TranslationString _opensshUsed =
             new TranslationString("Default SSH client, OpenSSH, will be used. (commandline window will appear on pull, push and clone operations)");
 
-        private readonly TranslationString _noMergeToolConfigured =
-            new TranslationString("There is no mergetool configured. Do you want to configure kdiff3 as your mergetool?" +
-                Environment.NewLine + "Select no if you want to configure a different mergetool yourself.");
-
-        private readonly TranslationString _noMergeToolConfiguredCaption =
-            new TranslationString("Mergetool");
-
         private readonly TranslationString _languageConfigured =
             new TranslationString("The configured language is {0}.");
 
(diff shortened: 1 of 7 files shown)


GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -587,59 +502,55 @@ private bool CheckGitCmdValid()
...
                 RenderSettingUnset(MergeTool, MergeTool_Fix, _configureMergeTool.Text);
                 return false;
             }
 
-            string mergetool = CommonLogic.GetGlobalMergeTool().ToLowerInvariant();
+            // Hardcode parameters to some mergetools
             if (EnvUtils.RunningOnWindows())
             {
-                if (CommonLogic.IsMergeTool("kdiff3"))
+                if (mergetool.ToLowerInvariant() == "kdiff3")
```
RussKie: I suggest we completely remove all explicit dependencies on specfic tools

gerhardol: Yes, this PR is changing the behavior, not auto suggesting KDiff3 (as the exe is not bundled). For KDiff3 GE has knowledge about parameters etc, why it is specially handled here.

RussKie:
We can remove the rest in the follow up work

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
gerhardol: > We can remove the rest in the follow up work

I have been staring at these lines of code for a long time now and do not fully realize what to replace it with.
The current code is handling the previous but should be removed. A follow up is needed anyway.

gerhardol:
I may look at the hardcoded checks in ChecklistSettingsPage.cs, maybe they are not needed at all.
Or this is merged now and fixed later, the remaining is clear for me.

RussKie:
@msftbot retry merge

RussKie:
@msftbot auto merge

PR #7012, opened 2019-08-12
should fix: [NBug] Sequence contains no elements #7011

### LINKED ISSUE #7011: [NBug] Sequence contains no elements
## Current behaviour

Git Extension process crash and show a popup window of bug reporting instead of compare to menu

## Expected behaviour

No process crash and ability to compare a commit with a branch

## Steps to reproduce
Click to select a commit row
Ctrl+Click to unselect the commit
Right click on that commit row
Go to Compare sub menu, then click on "Compare To branch..."

## Error Details
```
System.InvalidOperationException: Sequence contains no elements
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at GitUI.RevisionGridControl.CompareToBranchToolStripMenuItem_Click(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
```

## Additional information

Looking at the code the exception is on that line and cause because there is no selected commit
https://github.com/gitextensions/gitextensions/blob/ddce80631f19c41662fef4632b6d0aa22a8cddde/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs#L2365

## Environment
- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

manuc66:
Fixes #7011 

## Proposed changes

- When right clicking on a row that has been removed from selection force the row to be selected

## Test methodology 

- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
+++ b/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
@@ -52,7 +52,7 @@ See the changes in the commit form.");
         private readonly IRevisionFileTreeController _revisionFileTreeController;
         private readonly IFullPathResolver _fullPathResolver;
         private readonly IFindFilePredicateProvider _findFilePredicateProvider;
-        private GitRevision _revision;
+        [CanBeNull] private GitRevision _revision;
 
         public RevisionFileTreeControl()
         {
@@ -138,7 +138,7 @@ See the changes in the commit form.");
             findToolStripMenuItem_Click(null, null);
         }
 
-        public void LoadRevision(GitRevision revision)
+        public void LoadRevision([CanBeNull] GitRevision revision)
         {
             _revision = revision;
             _revisionFileTreeController.ResetCache();
@@ -287,7 +287,7 @@ See the changes in the commit form.");
 
         private IEnumerable<string> FindFileMatches(string name)
         {
-            var candidates = Module.GetFullTree(_revision.TreeGuid.ToString());
+            var candidates = _revision == null ? Enumerable.Empty<string>() : Module.GetFullTree(_revision.TreeGuid.ToString());
             var predicate = _findFilePredicateProvider.Get(name, Module.WorkingDir);
 
             return candidates.Where(predicate);
@@ -652,7 +652,7 @@ See the changes in the commit form.");
 
         private void resetToThisRevisionToolStripMenuItem_Click(object sender, EventArgs e)
         {
-            if (tvGitTree.SelectedNode?.Tag is GitItem gitItem)
+            if (tvGitTree.SelectedNode?.Tag is GitItem gitItem && _revision != null)
             {
                 if (MessageBox.Show(_resetFileText.Text, _resetFileCaption.Text, MessageBoxButtons.OKCancel) == DialogResult.OK)
                 {
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -1344,7 +1344,7 @@ namespace GitUI
             var pt = _gridView.PointToClient(Cursor.Position);
             var hti = _gridView.HitTest(pt.X, pt.Y);
 
-            if (_latestSelectedRowIndex == hti.RowIndex)
+            if (_latestSelectedRowIndex == hti.RowIndex && _gridView.Rows[_latestSelectedRowIndex].Selected)
             {
                 return;
             }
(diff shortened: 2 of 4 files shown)


manuc66:
@RussKie The real problem is that the line was not selected, the fix should be better now as it should fix the root cause

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1344,7 +1344,7 @@ private void OnGridViewCellMouseDown(object sender, DataGridViewCellMouseEventAr
             var pt = _gridView.PointToClient(Cursor.Position);
             var hti = _gridView.HitTest(pt.X, pt.Y);
 
-            if (_latestSelectedRowIndex == hti.RowIndex)
+            if (_latestSelectedRowIndex == hti.RowIndex && _gridView.Rows[_latestSelectedRowIndex].Selected)
```
gerhardol: This is basically "handle rightclick as selecting the revision". 
That the Selected property deviates from _latestSelectedRowIndexis an error that should be handled. Maybe this is sufficient.

This should be fine, but I cannot see that _gridView.Rows[_latestSelectedRowIndex] may not give null exceptions, so that should be checked.

However, it is not a guarantee that _latestSelectedRowIndex is still correct. There may have to be a check in CompareToBranchToolStripMenuItem_Click() and/or in OnGridViewSelectionChanged() or ContextMenuOpening().

manuc66: `_gridView.Rows[_latestSelectedRowIndex]` is de-referenced on line 1357 with a guard `IsValidRevisionIndex(_latestSelectedRowIndex)` but I wouldn't imagine that  `_latestSelectedRowIndex == hti.RowIndex` beeing true and not `_latestSelectedRowIndex >= 0 && _latestSelectedRowIndex < _gridView.RowCount` since `_latestSelectedRowIndex ` is equal to ` hti.RowIndex` which basicaly is an existing row (it has been clicked). Do you think I really should check for null or `IsValidRevisionIndex` ?

manuc66: With this fix I was not able to open the conext menu without having a row selected. I think that context menu scenarios are fixed.

But action exposed by the context menu can also be triggered whit hot keys when there is no selection in the grid. If I hit `CTRL+D` (Which is Campare to working directory) without a section the app keep crashing ! => In this case what should be the expected behavior ?

gerhardol: No exception with hot keys - a null check is needed.

RussKie:
> I would have preferred to separate git-fetch and git-pull actions, but I can live with this.

I don't think you've meant it for this PR :)

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -287,7 +287,7 @@ protected override void OnRuntimeLoad()
 
         private IEnumerable<string> FindFileMatches(string name)
         {
-            var candidates = Module.GetFullTree(_revision.TreeGuid.ToString());
+            var candidates = _revision == null ? Enumerable.Empty<string>() : Module.GetFullTree(_revision.TreeGuid.ToString());
```
RussKie:
```suggestion
            var candidates = _revision == null ? Array.Empty<string>() : Module.GetFullTree(_revision.TreeGuid.ToString());
```
`Array.Empty<T>` is a singleton, whilst `Enumerable.Empty<T>` is not.
In .NET Core it was fixed, but I don't believe it was fixed in .NET Fx.

I think the type inference should work, if not the following should work:
```suggestion
            IEnumerable<string> candidates = _revision == null ? Array.Empty<string>() : Module.GetFullTree(_revision.TreeGuid.ToString());
```

manuc66: It seems fixed in framework 4.8, see : https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,1284 or in 4.6.1: https://github.com/microsoft/referencesource/blob/4.6.2/System.Core/System/Linq/Enumerable.cs#L1161

RussKie: The link is pointing to 4.6.2, not 4.6.1

manuc66:
@RussKie squash done

gerhardol:
> I don't think you've meant it for this PR :)

At least, I meant to approve this PR

RussKie:
Thank you

PR #7032, opened 2019-08-21
Fix potential bad end of line replacement

pmiossec:
If the value contains a `\r\n`, the result will contain `\r\r\n`
which duplicate blank lines (in multi-line controls) at the loading and each time a settings is changed and saved...

## Test methodology 

- Manual

## Test environment(s) 
- Git Extensions 3.2.0
- Build 885dbb88d9ed3e02e77b1291df54a634ac159e0e
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/GitUIPluginInterfaces/Settings/StringSetting.cs
+++ b/Plugins/GitUIPluginInterfaces/Settings/StringSetting.cs
@@ -61,8 +61,9 @@ namespace GitUIPluginInterfaces
                     settingVal = Setting.ValueOrDefault(settings);
                 }
 
+                // for multiline control, transform "\n" in "\r\n" but prevent "\r\n" to be transformed in "\r\r\n"
                 control.Text = control.Multiline
-                    ? settingVal?.Replace("\n", Environment.NewLine)
+                    ? settingVal?.Replace(Environment.NewLine, "\n").Replace("\n", Environment.NewLine)
                     : settingVal;
             }
 


pmiossec:
I would like to see this fix in the 3.2.
I have developed the 'Azure Devops commit hint' plugin (like the jira one), ((for internal use for the moment until it is finished or the plugin manager is merged ;) )) and this bug is very annoying...

Plugins/GitUIPluginInterfaces/Settings/StringSetting.cs
```diff
@@ -62,7 +62,7 @@ public override void LoadSetting(ISettingsSource settings, TextBox control)
                 }
 
                 control.Text = control.Multiline
-                    ? settingVal?.Replace("\n", Environment.NewLine)
+                    ? settingVal?.Replace(Environment.NewLine, "\n").Replace("\n", Environment.NewLine)
```
RussKie: Why are we doing this at all?
What is the use-case we are trying to address here?

pmiossec: I didn't investigate too much but it bite more than 2 times. Enough to push me to make this fix.

If the multi-lines value of a setting is stored with a "\r\n", the actual code introduce a bad behavior. It transforms this every "\r\n" in "\r\r\n" which is displayed everywhere as 2 end of lines. And every times the setting is saved, this bad behavior continues adding new end of lines.

I don't remember the goal of the original `.Replace()` (7fa0eccc9f007cfe8fe277b210fc44af0ad1e6ba). I imagine it's because we can't guarantee the end of line character especially when the setting is saved in the repository but the strategy I made in the past was not perfect.

This fix could be summed up as "convert only the unix end of line and don't mess the windows ones".

mstv:
```suggestion
                    ? settingVal?.Replace(Environment.NewLine, "\n")?.Replace("\n", Environment.NewLine)
```

pmiossec: @mstv I was not 100% sure about that so I tested it yesterday. And resharper seems to tell me that it is not needed.

I imagine that if the string is null, it returns the null value immediately. Otherwise, if that's not null, the string could not become null.

mstv: Perfect.

RussKie: > if the string is null, it returns the null value immediately. Otherwise, if that's not null, the string could not become null.

That's correct

RussKie: > If the multi-lines value of a setting is stored with a "\r\n", the actual code introduce a bad behavior. It transforms this every "\r\n" in "\r\r\n" which is displayed everywhere as 2 end of lines. And every times the setting is saved, this bad behavior continues adding new end of lines.



> This fix could be summed up as "convert only the unix end of line and don't mess the windows ones".

Could you please add this as a comment to the code

RussKie: Looks like you've already done it 👍

PR #7033, opened 2019-08-21
Open jira settings when not configured

pmiossec:
## Proposed changes

- Fix a bad check if settings exists
- open the settings if not configured

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 885dbb88d9ed3e02e77b1291df54a634ac159e0e
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.cs
+++ b/Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.cs
@@ -62,7 +62,7 @@ namespace JiraCommitHintPlugin
 
         public override bool Execute(GitUIEventArgs args)
         {
-            if (_enabledSettings.ValueOrDefault(Settings))
+            if (!_enabledSettings.ValueOrDefault(Settings))
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.cs
```diff
@@ -62,8 +62,9 @@ public JiraCommitHintPlugin()
 
         public override bool Execute(GitUIEventArgs args)
         {
-            if (_enabledSettings.ValueOrDefault(Settings))
+            if (!_enabledSettings.ValueOrDefault(Settings))
```
gerhardol: Does this work at all now?

RussKie:
Could you please explain the use case, it is not clear to me what was wrong that we're fixing here.

pmiossec:
>Could you please explain the use case, it is not clear to me what was wrong that we're fixing here.

It's just a little fix and improvement.

Before this PR, we had a `Jira Commit Hint` menu item in the 'Plugin' Menu that does nothing.

I discovered, when developing the same plugin for Azure DevOps that a feature was written to show a message box with an example of a commit message by querying Jira with all the parameters (not a great value but why not).

The first commit fix the condition to enable this feature.

And while I was there, I did the same thing that we do on some other plugin , that is, to display the settings window when the plugin is not configured. This is the changed introduced in the 2nd commit. Even if you don't have Jira (I don't, I just backport from the AzureDevops plugin), you could test this part.

>Does this work at all now?

@gerhardol So, yes, it works now 😉

RussKie:
Merge if @gerhardol is happy

pmiossec:
>Merge if @gerhardol is happy

So, I let @gerhardol merge it 😉

PR #7037, opened 2019-08-21
Display icons in commit templates menu items

pmiossec:
## Proposed changes

- Display icons (especially the one of the **Jira** commit hint plugin) in commit templates menu items

PS: I have created because I'm mostly interested by the api change that it introduce (most of the code is common, there is just one line to update in the commit hint plugin).

### Before

image

### After

With provider:
image

Without provider (don't block on the Azure DevOps icon, this PR add Jira icon but I only can test it with the AzureDevOps plugin I'm developping...):

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 4bb4263d32fd2787bb14c191fe27f69166e88393
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitCommands/CommitTemplateItem.cs
+++ b/GitCommands/CommitTemplateItem.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Drawing;
 using System.IO;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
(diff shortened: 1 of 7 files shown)


GitUI/Properties/Resources.Designer.cs
```diff
@@ -254,6 +254,16 @@ internal class Resources {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized resource of type System.Drawing.Bitmap.
+        /// </summary>
+        internal static System.Drawing.Bitmap Settings {
```
RussKie: We already have this icon, though it is probably under a different name:
image

Please re-use it instead of adding a new resource

pmiossec: Thanks. I didn't know of the `Images` resources. I was miss leaded by the fact that there are images in the `Resources` resources.

RussKie:
Squash and merge at will

PR #7039, opened 2019-08-23
Make plugin settings development easier

pmiossec:
Fixes some paper cuts felt when developing a plugin (so I hope that you won't ask we to do one PR for each... 😉 but review is easier commit by commit).

### Settings: Ensure that `CustomControl` is set

if we let the Setting(s) classes create the controls.
Otherwise, it could will trow a NRE if we access it to get the current value

### Add a `PseudoSetting` to display controls that are not settings

i.e. that don't save a value in the settings files.
It is used to easily display a control that is not a setting (linklabel, text,...)

PS: Also, solve problem linklabels truncated horizontally of vertically if the good size is not set
(because the size no more have to be set)

 => and apply it to the "Jira Commit Hint" plugin

PS: I'm not fond of the name of the class. If someone have a better idea (`DisplayOnlySetting` ?)

### Being able to change CredentialsControl labels

to specify other things than the currently "User name" / "Password".

For example: "email" / "Token"

PS: not used at the moment. Except in my, still in development, AzureDevops plugin. But could help if someone want to develop a new plugin

### Before

Link label truncated vertically (because the height must be set by the dev and has been done badly)
image

### After

Link label not truncated vertically (because no more needed to set the height) and the linklabel and the Jira Fields are still displayed (that prove that the `PseudoClass` is working)

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 1c94e2894362a1576a85aa19fc224b250d619d2d
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/GitUIPluginInterfaces/Settings/PseudoSetting.cs
+++ b/Plugins/GitUIPluginInterfaces/Settings/PseudoSetting.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Windows.Forms;
+
+namespace GitUIPluginInterfaces
+{
+    /// <summary>
+    /// Not a real setting (as it save no setting value). It is used to display a control that is not a setting (linklabel, text,...)
+    /// </summary>
+    public class PseudoSetting : ISetting
+    {
+        private readonly Func<TextBox> _textBoxCreator;
+
+        public PseudoSetting(Control control, string caption = "")
+        {
+            Caption = caption;
+            CustomControl = control;
+        }
+
+        public PseudoSetting(string text, string caption = "    ", int? height = null,  Action<TextBox> textboxSettings = null)
+        {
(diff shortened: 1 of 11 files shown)


Plugins/GitUIPluginInterfaces/Settings/PseudoSetting.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+
+                textboxSettings?.Invoke(textbox);
+                return textbox;
+            };
+
+            CustomControl = _textBoxCreator();
+        }
+
+        private Func<TextBox> _textBoxCreator;
```
RussKie: [nit] please move fields before ctors.
Also this can be made `readonly`

PR #7040, opened 2019-08-23
Jira commit hint plugin: fix some string not translated

pmiossec:
To be merged after #7039 will be (just the last commit is specific to this PR).

## Proposed changes

JiraPlugin: Fix way to define TranslationString that prevent translation

+ remove TranslationString redundant with settings caption
(which is translated when defined in a private field!)

## Test methodology 

- Some more strings appears in `English.Plugins.xlf`

### DIFF
--- a/GitUI/Translation/English.Plugins.xlf
+++ b/GitUI/Translation/English.Plugins.xlf
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="3.2.0">
+<xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="3.3.0">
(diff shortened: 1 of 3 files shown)


RussKie:
We can't take that in 3.2, it will go in after the release.

pmiossec:
>We can't take that in 3.2, it will go in after the release.

That was not my intention to merge it before 3.2. I just did the PR once I discovered the problem.

But, just for information, I don't understand why it can't be...

I don't see the problem because the strings are not translated at the moment, so, merging the PR will change nothing, no!?!

gerhardol:
> I don't see the problem because the strings are not translated at the moment, so, merging the PR will change nothing, no!?!

No formal problem before - no problem.
This will maybe confuse translators though

RussKie:
> I don't see the problem because the strings are not translated at the moment, so, merging the PR will change nothing, no!?!

We have announced a release on Transifex and translators have translated supplied strings.
Adding more strings may render languages which currently meet 95% translation bar ineligible for inclusion into the release, thus affecting the users.

pmiossec:
OK. Good to know even if I would have imagined that the inclusion was
calculated only on Gitextensions and not the plugins...

RussKie:
> To be merged after #7039 will be (just the last commit is specific to this PR).

Please rebase

pmiossec:
done.

PR #7044, opened 2019-08-23
Add and refactor diffmerge tools

RussKie:
## Proposed changes

- Refactor how diff/merge tools are defined and configured, each tool is moved into own type/class that defines all tool's
properties, such as name, executable, diff/merge commald line args, etc.
This will simlify how new tools are added, and allow easy deprecation of old tools. New versions of tools can easily be added too as new types.

- Streamline diff/merge tool related logic in how tools are configured by users in Configuration dialogs. All special handling for selected tools are removed (e.g. KDiff3 and TMerge). By default Git Extensions will have no awareness of diff/merge tools.
🎉  As a bonus - a user will be able to see what parameters each tool takes, and be able to configure custom scripts/executables to be executed as part of diff/merge invocation.

- Remove support for TortoiseGit prior 1.8 (TortoiseMerge.exe)

- Remove support for determining KDiff3 location from the registry.

- Remove "Keep backup (.orig) after merge" setting

## Back-compat

- Any pre-configured tools should just work. E.g. "semanticdiff" tool has been renamed to "semanticmerge" for consistency. However if it was configured in .gitconfig, it will continue to be correctly loaded in Git Extensions:
    ```
    [diff]
        tool = semanticdiff
        guitool = semanticdiff
    [difftool "semanticdiff"]
        path = C:/Users/russkie/AppData/Local/semanticmerge/semanticmergetool.exe
        cmd = \"C:/Users/russkie/AppData/Local/semanticmerge/semanticmergetool.exe\" -s \"$LOCAL\" -d \"$REMOTE\"
    [difftool]
        prompt = false
    [merge]
        tool = semanticmerge
    [mergetool "semanticmerge"]
        path = C:/Users/russkie/AppData/Local/semanticmerge/semanticmergetool.exe
        cmd = \"C:/Users/russkie/AppData/Local/semanticmerge/semanticmergetool.exe\" -s \"$REMOTE\" -d \"$LOCAL\" -b \"$BASE\" -r \"$MERGED\"
        trustExitCode = true
    [mergetool]
        prompt = false
        keepBackup = false
    ```

  image

## Screenshots:

- Open Git Extensions with no tool configured:
image

- Diff/merge tool configuration screen:
image

- Attempt to resolve a merge conflict without a diff/merge tool configured:
image

## Test methodology 

- manual
- unit tests

----


RussKie:
This is something I've started back in mid 2017, and it was lying around annoying me.
May as well finish it up....

@gerhardol this may help/align with #7009

gerhardol:
I would prefer #7009 first, as that aligns the behavior with the GE implementation, then rebase on master. I have considered ripping out som of the GE specifics and rely on Git handling only though, then most of this is a little overkill.

gerhardol:
I would like to go in the direction and remove most settings for merge/diff tools, to rely on Git instead. 
difftool.path may still be needed, but difftool.cmd will not

See for instance output of 
```git difftool --gui --tool-help```
A few need to be added like vscode (and TortoiseDiff?)
Tools the user have added should be seen too

The tricky is how to verify that a tool is setup (a test repo is one way).

This PR is not in conflict with my view, but parts of it will be reverted.

**Comment by RussKie** [comment]
Thank you for pointing that out.
I have tried resetting .gitconfig and configuring git with `git difftool
--tool=<tool>` and it didn't appear to make any difference... Like at all.

I don't think our UI needs to change, and the logic is now significantly
simpler and cleaner than it used to be.

One thing I'm really puzzled about why we are using `diff.guitool` instead
of `diff.tool`, or why we are not using `merge.guitool`.

**Comment by gerhardol** [comment]
merge.guitool is not used because it was added in Git 2.20 option and it was hard to add before the hardcoding for tools was removed.

git.difftool is a frontend to git.diff, the --gui is supposed to differ command line from terminal compatible from what I see, but in most cases there is no specific with guitool. I have used this personally to have a separate difftool from the commandline.

As I see the different "classes" of tools:
 * Git supported and found (blacklist vimdiff*) and "Git user-defined". path, cmd is not required, no editing (but save as new "tool")
 * GE built-in (not already listed by Git). Path required (but suggested by GE)
 * Git built in, not in path. Require only path
 * Custom tools, supply a default cmd - how to verify that it is working, a repo is needed? 

Checks for cmd, path has to be adapted

**Comment by RussKie** [comment]
I don't see a problem (am I missing something?):

- GE knows about several tools, e.g. their exe file names, default installation location and commands. 
Just because git knows that too doesn't mean we need to remove our configs. 
We may supply more information that git may need, but is this a problem?
- Once a tool selected, GE tries to find an exe for the selected tool. If GE can't find it, likely git won't find it either.
The user will need to supply the path, or select a different tool.
- If the user configured a different tool that GE doesn't know about, GE will just show it. (this needs to be verified)

**Comment by gerhardol** [comment]
> I don't see a problem (am I missing something?):

The problem is that difftool.<tool>.cmd and .path may not be needed and it is a little difficult to test the setup outside a repo.
The checks for .path, .cmd will have to be modified.

The rest (custom paths etc) is just details.

**Comment by RussKie** [comment]
Should be good to go.

File: UnitTests/GitUITests/DiffMergeTools/DiffMergeToolsProviderTests.cs [hunk 1]
```diff
@@ -0,0 +1,164 @@
+﻿////using System;
```
mstv: What's your plan for this file?

RussKie: Ah thanks! It was very late, and I had this file marked as "viewed", so I missed it.
It needs to go, I wrote it as an initial test harness.

gerhardol:
Will make formal review, I have looked at it and considered how to change to support "Git tools". 
The changes are mostly not in the changed code

GitCommands/DiffMergeTools/WinMerge.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class WinMerge : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string DiffCommand => "-e -u \"$LOCAL\" \"$REMOTE\"";
+
+        /// <inheritdoc />
+        public override string ExeFileName => "winmergeu.exe";
+
+        /// <inheritdoc />
+        public override string MergeCommand => "-e -u -dl \"Original\" -dr \"Modified\" \"$MERGED\" \"$REMOTE\"";
```
gerhardol: case "winmerge":
                    return "\"" + exeFile + "\" -e -u  -wl -wr -fm -dl \"Mine: $LOCAL\" -dm \"Merged: $BASE\" -dr \"Theirs: $REMOTE\" \"$LOCAL\" \"$BASE\" \"$REMOTE\" -o \"$MERGED\"";

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
```diff
@@ -0,0 +1,23 @@
+using System.Collections.Generic;
+
+namespace GitCommands.DiffMergeTools
+{
+    // deprecate pre-1.8 versions (ec9e3fd3438b38b65cdfc60dc419709ca8c03aff)
+    internal class TortoiseGitMerge : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string ExeFileName => "TortoiseGitMerge.exe";
```
gerhardol: This removes support for tgit < 1.8 as well as using tsvn binary (unless that is changed too)
(just noting, probably OK)

GitCommands/DiffMergeTools/SemanticMerge.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <inheritdoc />
+        public override IEnumerable<string> SearchPaths => Folders;
+
+        private static string[] GetFolders()
+        {
+            string folder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            return new[]
+            {
+                Path.Combine(folder, @"semanticmerge"),
```
gerhardol: Adding, OK

GitCommands/DiffMergeTools/SemanticMerge.cs
```diff
@@ -0,0 +1,36 @@
...
+        public override string DiffCommand => "-s \"$LOCAL\" -d \"$REMOTE\"";
+
+        /// <inheritdoc />
+        public override string ExeFileName => "semanticmergetool.exe";
+
+        /// <inheritdoc />
+        public override string MergeCommand => "-s \"$REMOTE\" -d \"$LOCAL\" -b \"$BASE\" -r \"$MERGED\"";
+
+        /// <inheritdoc />
+        public override string Name => "semanticmerge";
```
gerhardol: semanticdiff for the difftool currently

GitCommands/DiffMergeTools/Kdiff3.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <inheritdoc />
+        public override string ExeFileName => "kdiff3.exe";
+
+        /// <inheritdoc />
+        public override string Name => "kdiff3";
+
+        /// <inheritdoc />
+        public override IEnumerable<string> SearchPaths => new[]
+        {
+            // regkdiff3path
```
gerhardol: Registry does not seem to be queried any more - maybe OK

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -502,81 +482,42 @@ private bool CheckGitCmdValid()
...
-                    string p = GetGlobalSetting($"mergetool.{mergetool}.cmd");
-                    if (string.IsNullOrEmpty(p))
-                    {
-                        RenderSettingUnset(MergeTool, MergeTool_Fix, string.Format(_mergeToolXConfiguredNeedsCmd.Text, mergetool));
-                        return false;
-                    }
-
-                    RenderSettingSet(MergeTool, MergeTool_Fix, string.Format(_customMergeToolXConfigured.Text, mergetool));
-                    return true;
-                }
```
gerhardol: This special handling is now removed. Probably OK

GitCommands/DiffMergeTools/DiffMerge.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace GitCommands.DiffMergeTools
+{
+    internal class DiffMerge : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string ExeFileName => "DiffMerge.exe";
+
+        /// <inheritdoc />
+        public override string MergeCommand => "/m /r =\"$MERGED\" \"$LOCAL\" \"$BASE\" \"$REMOTE\"";
```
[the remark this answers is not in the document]
RussKie: Good spot, I'll investigate

GitCommands/DiffMergeTools/WinMerge.cs
RussKie: For a different version we need to add another `DiffMergeTool` definition

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
RussKie: If different versions of diff/merge tool have different exe names/different parameter sets they must be presented as different `DiffMergeTool` definitions.
This will allow tool-specific configurations and deprecation of out-dated tools in a clean way

GitCommands/DiffMergeTools/SemanticMerge.cs
RussKie: In the latest version there is a single executable, naming them differently is not a good idea.

GitCommands/DiffMergeTools/Kdiff3.cs
RussKie: Ah yes, I missed this...
Do we really want t bring it back?

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
RussKie: I don't want any more special handling, it makes the code incoherent mess.

gerhardol: Neither do I, but the behavior changes, this at least must be documented in the commit message. It is very hard to investigate changes otherwise.

GitCommands/DiffMergeTools/Kdiff3.cs
gerhardol: It is a functionality change, can probably be documented. KDiff3 is probably not the tool most install nowadays (it was a good addition when it was added for sure).

GitCommands/DiffMergeTools/SemanticMerge.cs
gerhardol: Then this should be documented in commit message and reslease notes

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
gerhardol: Should be documented in commit message

GitCommands/DiffMergeTools/WinMerge.cs
gerhardol: This was a change in the defaults

RussKie:
Good points, I'll action them.
Thank you

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
RussKie: Done

gerhardol: pushed?

GitCommands/DiffMergeTools/WinMerge.cs
RussKie: Thank you

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
RussKie: Added to the commit message

GitCommands/DiffMergeTools/SemanticMerge.cs
RussKie: This will continue to work as is - configured tools will be recognised even though they may not be in our official list:
image

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <inheritdoc />
+        public override string MergeCommand => "-base:\"$BASE\" -mine:\"$LOCAL\" -theirs:\"$REMOTE\" -merged:\"$MERGED\"";
+
+        /// <inheritdoc />
+        public override string Name => "tortoisemerge";
```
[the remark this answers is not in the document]
RussKie: fixed, thank you

GitCommands/DiffMergeTools/Kdiff3.cs
RussKie: Noted the removal in the commit message

RussKie:
I believe I addressed the review points. I have updated the commit messages and the PR description. Added few more screenshots.

Ready for another round.
Thank you

GitCommands/DiffMergeTools/SemanticMerge.cs
gerhardol: This should be documented in the commit:
The diff tool names _semanticdiff_ and _tortoisemerge_ is now named _semanticmerge_ and _tmerge_ as the mergetools already was called.

GitCommands/DiffMergeTools/DiffMerge.cs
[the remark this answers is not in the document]
gerhardol: I would prefer -merge and -result; the long names for readability and '-' to keep multi platform compatibilty in general (DiffMerge runs at least on Mac).

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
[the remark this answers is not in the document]
gerhardol: Note: "tortoisemerge" used in FormResolveConflicts, both should be added (see also suggestion below)

gerhardol:
Next steps:
 * Remove hardcoded use of .cmd and .path in FormResolveConflicts, the parameters are not required. A few manipulations for instance for Use2WayMerge() will have to be dropped. Or only provide those options when .cmd is available.
 * Make .cmd editable in settings
 * (possibly) add custom entries
 * (possibly) add Git built-in tools (automatically)

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -133,118 +137,87 @@ protected override void PageToSettings()
             }
         }
 
-        private void GlobalMergeTool_TextChanged(object sender, EventArgs e)
+        private string BrowseDiffMergeTool(string toolName, string path)
         {
-            if (IsLoadingSettings)
+            var diffMergeToolConfig = _diffMergeToolConfigurationManager.LoadDiffMergeToolConfig(toolName, null);
```
gerhardol: If the path does not exist (like if the tool is not configured in GE), exception occurs

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -133,118 +137,87 @@ protected override void PageToSettings()
...
-            if (!string.IsNullOrEmpty(diffToolPath))
-            {
-                exeFile = diffToolPath;
-                exeName = Path.GetFileName(exeFile);
-            }
-            else
+            if (sender == DifftoolPath)
             {
-                exeFile = MergeToolsHelper.FindDiffToolFullPath(ConfigFileSettingsSet, _NO_TRANSLATE_GlobalDiffTool.Text, out exeName);
+                DifftoolPath.Text = DifftoolPath.Text.ToPosixPath();
```
gerhardol: If you type a \, the cursor is moved back to first position

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -133,118 +137,87 @@ protected override void PageToSettings()
...
-                return;
+                initialDirectory = Path.GetDirectoryName(diffMergeToolConfig.Path);
             }
 
-            var mergeTool = _NO_TRANSLATE_GlobalMergeTool.Text.Trim();
-            MergetoolPath.Text = CurrentSettings.GetValue($"mergetool.{mergeTool}.path");
-            MergeToolCmd.Text = CurrentSettings.GetValue($"mergetool.{mergeTool}.cmd");
+            var filter = !string.IsNullOrWhiteSpace(diffMergeToolConfig.ExeFileName)
+                ? string.Format("{0}|{1}", toolName, diffMergeToolConfig.ExeFileName)
+                : string.Format("{0} (*.exe;*.cmd;*.bat)|*.exe;*.cmd;*.bat", toolName);
```
gerhardol: As the toolname is set to 'toolname.exe' for unknown tools, it can be hard to find the exe
Should probably be changed in 
        public DiffMergeToolConfiguration LoadDiffMergeToolConfig(string toolName, string userSuppliedPath)

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: Could you please provide me with steps to reproduce?

image

RussKie: Found the issue, thank you!

With a fix will open the `Program Files` folder, if any issues:
image

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: Could you please clarify?

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: Sorry, I am unable to repro, perhaps I'm not doing it like you are
Please provide more info

GitCommands/DiffMergeTools/DiffMerge.cs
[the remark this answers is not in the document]
RussKie: Done

RussKie:
> * Remove hardcoded use of .cmd and .path in FormResolveConflicts, the parameters are not required. A few manipulations for instance for Use2WayMerge() will have to be dropped. Or only provide those options when .cmd is available.

Thank you, I totally missed these.

I pushed https://github.com/gitextensions/gitextensions/pull/7044/commits/17f6eec67db54a3874becbc93e2a824e4cbb400b as a prototype.

GitCommands/DiffMergeTools/Araxis.cs
```diff
@@ -1,24 +0,0 @@
-using System.Collections.Generic;
```
RussKie: Chose to remove Araxis as it isn't fully configured as diff/merge tool. I've looked at the tool's configuration and it didn't look trivial.
We can add the tool back, if users request it.

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
gerhardol: Type a path, when entering a '\' the cursor moves to first position
Why not use LostFocus instead on OnChange?

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
gerhardol: If I add my tool specificTool, the filter is set to "specificTool.exe" which may not be what I expect

GitCommands/DiffMergeTools/Araxis.cs
gerhardol: araxis is supported by Git natively, so .cmd is not needed

gerhardol: https://www.araxis.com/merge/documentation-windows/integrating-with-other-applications.en#GitExtensions

The builtin support in master is not up to date
Remove 6.5 references, released 2003-2006

gerhardol:
> Thank you, I totally missed these.
I pushed 17f6eec as a prototype.

I wrote that down primary as an indication of a followup PR, not necessarily here.
Some parts that I planned to write already was implemented though (like editing the arguments), so the list was short.

However, _if_ 2way merge is kept as a feature, manipulation must still be done on mergetool.*.cmd arguments, the user should be allowed to change them (that can be done already, from the command line). The feature will then only be available if .cmd exists.

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: Gotcha.

RussKie:
> I wrote that down primary as an indication of a followup PR, not necessarily here.
> Some parts that I planned to write already was implemented though (like editing the arguments), so the list was short.
> 
> However, _if_ 2way merge is kept as a feature, manipulation must still be done on mergetool.*.cmd arguments, the user should be allowed to change them (that can be done already, from the command line). The feature will then only be available if .cmd exists.

Then I'd rather leave it as is for now, if we're ok with that. I will add a comment.

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -254,23 +242,53 @@ private void GlobalDiffTool_TextChanged(object sender, EventArgs e)
                 return;
             }
 
-            string diffTool = _NO_TRANSLATE_GlobalDiffTool.Text.Trim();
-            DifftoolPath.Text = CurrentSettings.GetValue($"difftool.{diffTool}.path");
-            DifftoolCmd.Text = CurrentSettings.GetValue($"difftool.{diffTool}.cmd");
+            var toolName = _NO_TRANSLATE_GlobalDiffTool.Text;
+
+            DifftoolPath.Enabled =
+                BrowseDiffTool.Enabled =
```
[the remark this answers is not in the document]
RussKie: What would it make less confusing?
Personally I prefer a Hungarian notation for UI control names, e.g. `txtDiffToolPath`, `btnBrowseDiffTool` etc, but it isn't prevalent in our codebase.

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
```diff
@@ -0,0 +1,23 @@
...
+        public override string MergeCommand => "-base:\"$BASE\" -mine:\"$LOCAL\" -theirs:\"$REMOTE\" -merged:\"$MERGED\"";
+
+        /// <inheritdoc />
+        public override string Name => "tmerge";
+
+        /// <inheritdoc />
+        public override IEnumerable<string> SearchPaths => new[]
+        {
+            @"TortoiseGit\bin\",
+            @"TortoiseSVN\bin\"
```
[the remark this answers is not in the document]
RussKie: done

GitCommands/DiffMergeTools/TortoiseGitMerge.cs
```diff
@@ -0,0 +1,23 @@
...
+    internal class TortoiseGitMerge : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string ExeFileName => "TortoiseGitMerge.exe";
+
+        /// <inheritdoc />
+        public override string MergeCommand => "-base:\"$BASE\" -mine:\"$LOCAL\" -theirs:\"$REMOTE\" -merged:\"$MERGED\"";
+
+        /// <inheritdoc />
+        public override string Name => "tmerge";
```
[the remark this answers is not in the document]
RussKie: done

GitCommands/DiffMergeTools/BeyondCompare4.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GitCommands.DiffMergeTools
+{
+    internal class BeyondCompare4 : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string ExeFileName => "bcomp.exe";
+
+        /// <inheritdoc />
+        public override string Name => "beyondcompare4";
```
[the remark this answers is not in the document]
RussKie: 1. We have beyondcompare3 and beyondcompare4
2. This is what users see in the dropdowns. I don't think "bc" is very descriptive

GitCommands/DiffMergeTools/BeyondCompare3.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GitCommands.DiffMergeTools
+{
+    internal class BeyondCompare3 : DiffMergeTool
+    {
+        /// <inheritdoc />
+        public override string ExeFileName => "bcomp.exe";
+
+        /// <inheritdoc />
+        public override string Name => "beyondcompare3";
```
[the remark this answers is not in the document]
RussKie: Ok

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -254,23 +242,53 @@ private void GlobalDiffTool_TextChanged(object sender, EventArgs e)
...
+                    DifftoolCmd.Enabled =
+                        DiffToolCmdSuggest.Enabled = !string.IsNullOrEmpty(toolName);
+
+            if (string.IsNullOrWhiteSpace(toolName))
+            {
+                DifftoolPath.Text =
+                    DifftoolCmd.Text = string.Empty;
+                return;
```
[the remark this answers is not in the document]
RussKie: Thank you, though the missing flow is in `PageToSettings` - if either `difftool` or `mergetool` aren't set we need to zero out respective configurations.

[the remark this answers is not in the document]
RussKie: done

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: fixed in 68ce7dcfbf88c762ef0ff47d63be42be59241bd0

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
RussKie: Fixed, thank you

RussKie:
I believe I addressed all comments. Ready for more :)

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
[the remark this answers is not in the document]
gerhardol: In this case it is three properties that start with Difftool* and one that start with BrowseDiifftool, so evey time I read this, my eyes get stuck and I look for a problem.
Not requesting a a change.

GitCommands/DiffMergeTools/Araxis.cs
gerhardol: I still believe Araxis should be kept, see the configuration in the link.
Cannot be used much though...

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -133,118 +144,124 @@ protected override void PageToSettings()
...
+        }
+
+        private string GetInitialDirectory(string path, string toolPreferredPath)
+        {
+            if (!string.IsNullOrWhiteSpace(path))
             {
-                exeFile = MergeToolsHelper.FindMergeToolFullPath(ConfigFileSettingsSet, _NO_TRANSLATE_GlobalMergeTool.Text, out exeName);
+                try
+                {
+                    string initialDirectory = Path.GetDirectoryName(path);
```
gerhardol: If path is a directory without a slash, the directory will be cleaned. Suggest the following:

                    if (Directory.Exists(path))
                    {
                        return PathUtil.ToNativePath(PathUtil.EnsureTrailingPathSeparator(path));
                    }

RussKie:
> I still believe Araxis should be kept, use the configuration in the link

I don't want to restore it because it was not fully configured.
Any user who has it configured should see no changes, however if we want to add Araxis as a first class citizen it needs to be configured. This is a separate piece of work.

GitCommands/DiffMergeTools/VsDiffMerge.cs
```diff
@@ -32,11 +32,13 @@ private static string GetVsDiffMergePath()
                 return ExeName;
             }
 
+            // For 2017 (15.0) and later, Common7 is not installed by default but often included
```
gerhardol: For 2017 (15.0) and later, VsDiffMerge is not installed by default but often included

GitCommands/DiffMergeTools/DiffMergeToolConfigurationManager.cs
```diff
@@ -0,0 +1,237 @@
...
+                return new DiffMergeToolConfiguration(exeName, fullPath ?? string.Empty, null, null);
+            }
+
+            if (!string.IsNullOrWhiteSpace(userSuppliedPath))
+            {
+                fullPath = userSuppliedPath;
+            }
+            else
+            {
+                var pathsToSearch = new[] { UnquoteString(GetToolSetting(diffTool.Name, DiffMergeToolType.Merge, "path")) }.Union(diffTool.SearchPaths);
```
gerhardol: This will get the path to the mergetool with the suggested name, which may be different from the difftool
Normally this will work (as mergetool is searched also for difftool)

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.cs
```diff
@@ -133,118 +144,125 @@ protected override void PageToSettings()
...
-                return;
+                diffMergeToolConfig = _diffMergeToolConfigurationManager.LoadDiffMergeToolConfig(toolName, null);
             }
 
-            var mergeTool = _NO_TRANSLATE_GlobalMergeTool.Text.Trim();
-            MergetoolPath.Text = CurrentSettings.GetValue($"mergetool.{mergeTool}.path");
-            MergeToolCmd.Text = CurrentSettings.GetValue($"mergetool.{mergeTool}.cmd");
-
-            MergeToolCmd.Enabled = !string.IsNullOrEmpty(MergeToolCmd.Text);
+            string initialDirectory = GetInitialDirectory(path.EnsureTrailingPathSeparator(),
```
gerhardol: Both path and diffMergeToolConfig.Path may be the full path to the executable, this method does not check that the path is a directory

RussKie: I'm due to write tests for this functionality

RussKie: Done. The code is updated to cater for both directory and file

GitCommands/DiffMergeTools/DiffMergeToolConfigurationManager.cs
RussKie: What are the changes of difftool executable be different from a mergetool executable?
From my observations all tools I've seen use the same executable.

GitCommands/DiffMergeTools/VsDiffMerge.cs
RussKie: fixed

RussKie:
> * A path like c:\ is not accepted as a directory

Opps, midnight coding sucks... I'll fix that.

> * If .path is not set, cmd should not be set either. Can be seen as a user error though...

Thank you! I'll fix it

gerhardol:
> Opps, midnight coding sucks... I'll fix that.

Midnight reviews are even worse - this PR has plenty of examples

RussKie:
> * A path like c:\ is not accepted as a directory
> * If .path is not set, cmd should not be set either. Can be seen as a user error though...

Updated

gerhardol:
Still some issues:
Paths ending with / will not work in the file dialog, as dialog only accepts Windows file separators.
CalculateInitialDirectory() should transform to Windows native paths with ToNativePath() (maybe an issue in OpenFileDialog....)
(OpenFileDialog uses a recent path that seem almost random to me, so this appears as weird behaiour for the user.)

If you press Browse, the Cancel it takes some seconds to get back control. The delay is in OpenFileDialog, not sure if something can/should be done about it.

A partial path is added to the cmd, no evaluation if the path is a file, this is fine.
However, an empty path will be be propagated with Suggest.
I could accept that as a feature (to see the arguments).

RussKie:
> However, an empty path will be be propagated with Suggest.
> I could accept that as a feature (to see the arguments).

This is intentional, precisely for the reason - to see the arguments.

> CalculateInitialDirectory() should transform to Windows native paths with ToNativePath() (maybe an issue in OpenFileDialog....)

It works for posix-formatted paths. I tried:
* C:/Program Files/Perforce
image
* C:/Program Files/Perforce/
image
* C:/Program Files/Perforce/p4merge.exe
image

Is there some other use-case I'm missing?


> OpenFileDialog uses a recent path that seem almost random to me, so this appears as weird behaiour for the user.

The initial directory is calculated as follows:
* user-provided path to a folder, if it exists, or
* the selected tool preferred path (the first existing, if there are alternatives), or
* ProgramFiles folder

gerhardol:
With dir set to c:/, I get "C:/Program Files/Perforce" (mergetool path, the last selected path).
The debugger did show initial path as "c:/" as expected, same for "c:\" also as expected. "C:" is converted to "c:\" and is OK.

So this is a specific case that can be disregarded, this is an issue with OpenFileDialog.

PR #7060, opened 2019-08-27
Azure DevOps commit message plugin

pmiossec:
## Proposed changes

- New plugin "AzureDevOpsCommitHint" that generate commit messages based on Azure DevOps user workitems retrieved with a Work items Query (WIQL)

PS:
- It is working great.
- It is here if it's useful for someone that want to use it
- ~~It is not intended to be merged until #7039 is merged  and we have good discussion around the commit `Settings: Prevent the multiple creation of controls` (because there are inconsistencies on how controls are created and that could be a problem when we develop a plugin with logic in the settings -- like this plugin and jira one for the moment--)~~ edit: PR #7039 merged.

### Before

Nothing.

### After

Settings:

image

Commit form:

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.2.0
- Build 5ca2b39b78329487963b84021605baf11171b9e3
- .NET Framework 4.7.3416.0
- DPI 192dpi (200% scaling)


RussKie:
@pmiossec can we please build is a separate plugin installable via the Plugin Manager?
I created https://github.com/gitextensions/GitExtensions.AzureDevOpsCommitHintPlugin

I'm sure @mast-eu and @maraf would be able to help you, if you get stuck.

maraf:
At your service, sir.

pmiossec:
No more needed. Converted to a plugin: 

https://www.nuget.org/packages/GitExtensions.AzureDevOpsCommitMessage

I imagine that soon, someone will be bitten by https://github.com/gitextensions/gitextensions/pull/7741#issuecomment-583797771
It works because the dependencies are loaded from the GitExtensions folder.

mast-eu:
> Converted to a plugin:
> 
> https://www.nuget.org/packages/GitExtensions.AzureDevOpsCommitMessage

Great 😄 🎉 

> I imagine that soon, someone will be bitten by [#7741 (comment)](https://github.com/gitextensions/gitextensions/pull/7741#issuecomment-583797771)
> It works because the dependencies are loaded from the GitExtensions folder.

I'd expect the same: https://github.com/gitextensions/GitExtensions.GerritPlugin/pull/2#discussion_r377614422
Let's figure out a common solution.

RussKie:
Yep. We need some working scenarios we can tinker with.

As an option to entertain, we can install the app under %AppData% folder,
just like VSCode does. This way we have no issues with read-only folders.

Another option, *I think*, is to use distributed config files. I know it
was possible to break a single web.config into several files. I wonder if
it is possible with app.config, and (if it is) whether we can have a part
stored under %AppData%, that we can modify.....

Issue #3489, opened 2017-01-10
Exception "Illegal characters in path" when invoking "browse" via comman line parameter


gimmi:
Hi, with version `2.49` the following command:

```
GitExtensions.exe browse "C:\path\to\dir\"
```

fail with the error

```
System.ArgumentException: Illegal characters in path.
```

work as expected if i remove the trailing `\` or add trailing `.`

```
GitExtensions.exe browse "C:\path\to\dir"
GitExtensions.exe browse "C:\path\to\dir\."
```

Not a big issue, but little annoying and possibly an easy fix.

b3rnhard:
This makes it difficult to invoke from Visual Studio when using "External Tools" with $(SolutionDir), which always adds a trailing backslash.

RussKie:
Is this still a problem?

b3rnhard:
Yes, it is.

RussKie:
In which version?

I've just tried it in 2.50.01. Running `>gitextensions browse C:\Development\test1` or `>gitextensions browse C:\Development\test1\` results in the same thing - the repo is opened and with a trailing slash appended if it is missing
Running `>>gitextensions browse C:\Development\test1\.` opens the repo, though the path is `>gitextensions browse C:\Development\test1\.\`

b3rnhard:
Calling 
```
"C:\Program Files (x86)\GitExtensions\GitExtensions.exe" browse "C:\Users\jahn\workspace\_Product\Prd-MIP\"
```
immediately crashes with Target Site 
```
Void CheckInvalidPathChars(System.String, Boolean) @ System.IO.Path
```
If I leave out the trailing backslash, everything works. If I remove the quotations like this but keep the trailing backslahs, it also works:
```
"C:\Program Files (x86)\GitExtensions\GitExtensions.exe" browse C:\Users\jahn\workspace\_Product\Prd-MIP\
```

RussKie:
Could you please post the full stack trace

RussKie:
I think in the first case it treats the trailing `\"` not as quoted `\` but as an escaped `"` - this leads to the path being `C:\Users\jahn\workspace\_Product\Prd-MIP"`...

b3rnhard:
Stacktrace for the version with quotation marks:
```
   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.InternalGetDirectoryName(String path)
[...]

PR #7109, opened 2019-09-06
PathUtil.GetDirectoryName did not accept root dirs

### LINKED ISSUE #7086: GE fails to identify .git folder at root of drive from cmd line
## Current behaviour

When using GITExtension's explorer context menu on a repository that is located at the root of a drive on a file or folder that is NOT at the root, it fails to find the .git folder and issue the following message instead:

`The current directory is not a valid git repository.`

## Expected behaviour

The repository should be found even if at the root of the drive.

## Steps to reproduce

Use any local repository that contains sub-folders. 
Map it to a new drive using either subst or net use. 
Navigate to the newly mapped drive letter, go inside a folder (any folder).
Right-click on any file in the repo and select "File history..." (or any other GE option, really).

## Screenshots

image
image
image

No

## Environment
- Git Extensions 3.1.1.6049
- Git 2.23.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #7086

## Proposed changes
 * Use .NET Path.GetDirectoryName instead of PathUtil to also detect root directories

The existing function could have been adapted (by removing a few lines), but the standard function should be used. 

## Test methodology 
Manual

## Test environment(s) 

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -477,7 +477,7 @@ namespace GitCommands
                     return dir.EnsureTrailingPathSeparator();
                 }
 
-                dir = PathUtil.GetDirectoryName(dir);
+                dir = Path.GetDirectoryName(dir);
             }
 
             return null;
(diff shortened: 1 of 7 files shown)


gerhardol:
Suggesting merge in 3.3

Kryptos-FR:
Unlikely to be correct since `Path.GetDirectoryName` can return `null` but `ToPosixPath` expects a non-`null` `string`.

gerhardol:
> Unlikely to be correct since Path.GetDirectoryName can return null but ToPosixPath expects a non-null string.

Path.GetDirectoryName returns null when called with null or root path. This is not the case in these situations.

However, it possible to get an exception with a call like 'GitExtensions.exe filehistory "n:/"' 
Pushing an update (so just ignoring "files" that consists of only whitespace).
This code is unrelated to the original change though.

gerhardol:
To fix exceptions for 'GitExtensions.exe filehistory "x:/"' where filehistory can be most any command and x:/ do not contain a .git dir, many changes are needed:
 * Handle trailing " (see https://github.com/gitextensions/gitextensions/issues/3489)
 * Handle that the filename is not in a Git repo and that it may be empty

Not really worth it

PR #7049, opened 2019-08-26
Use latest Windows SDK

NN---:
VS2019 has a simple way to compile with latest SDK.
https://developercommunity.visualstudio.com/content/problem/140294/windowstargetplatformversion-makes-it-impossible-t.html#reply-548052

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7112, opened 2019-09-07
Remove test assuming local file paths

gerhardol:
Test case TryGetExactPathName_Should_output_path_with_exact_casing() uses the local file system for comparisons.
The test for "c:\Program Files (x86)\Microsoft.NET\Primary Interop Assemblies" occasionally fails for me for unknown reason (when debugging it is OK), similar in #7049). (An alternative to this PR would be to remove this specific test only.
It is possible to install to other paths, the test assumes that the PC uses standard paths.
 Similar tests are already done in TryGetExactPathName_should_check_if_path_matches_case()

Note that TryGetExactPathName_Should_handle_network_path() could fail too, but there are no tests replacing that, so it is kept in this PR.

----

### DIFF
--- a/UnitTests/GitUITests/CommandsDialogs/FormFileHistoryControllerTests.cs
+++ b/UnitTests/GitUITests/CommandsDialogs/FormFileHistoryControllerTests.cs
@@ -29,21 +29,6 @@ namespace GitUITests.CommandsDialogs
             Assert.IsNull(exactPath);
         }
 
-        [TestCase(@"c:\Users\Public\desktop.ini")]
-        [TestCase(@"c:\pagefile.sys")]
-        [TestCase(@"c:\Windows\System32\cmd.exe")]
-        [TestCase(@"c:\Users\Default\NTUSER.DAT")]
-        [TestCase(@"c:\Program Files (x86)\Microsoft.NET\Primary Interop Assemblies")]
-        [TestCase(@"c:\Program Files (x86)")]
-        public void TryGetExactPathName_Should_output_path_with_exact_casing(string path)
-        {
-            var lowercasePath = path.ToLower();
-            var isExistingOnFileSystem = _controller.TryGetExactPath(lowercasePath, out string exactPath);
-
-            Assert.IsTrue(isExistingOnFileSystem);
-            Assert.AreEqual(path, exactPath);
-        }
-
         [Test]
         public void TryGetExactPathName_Should_handle_network_path()
         {


RussKie:
These tests are somewhat important because Windows is case insensitive, but git is. Incorrectly rendered paths have caused issues in the past.
I'd like to investigate why these tests fail on a AppVeyor VS2019 image too.
PS. Locally I never had any troubles with these tests.

pmiossec:
The test that fails is partly due to the refactoring I made with 1246b50aae33bf852545d49bbc102a0af163252f where I raise a little the bar with the expectations.

Indeed, before that, when the files existing, the unit test was verifying... **nothing!**

I made a change to verify that the paths cases was the ones of the real paths. I was thinking that the windows paths would have always the same cases for system files. It seems not to be the case 😢 as the AppVeyor VS2019 image has a different one...

What I can propose is to find a in-between between testing nothing (like before the refactoring) and having the expected casing (after the refactoring).

I.e:
- We start with the paths without the good casing (lower case)
- We check that the result is still equivalent (by lowering the result)
- We check that the path has been changed (expecting that it is the good case)

Something like that:
```
        [TestCase(@"c:\Users\Public\desktop.ini")]
        [TestCase(@"c:\Windows\System32\cmd.exe")]
        [TestCase(@"c:\Users\Default\NTUSER.DAT")]
        [TestCase(@"c:\Program Files (x86)\Microsoft.NET\Primary Interop Assemblies")]
        [TestCase(@"c:\Program Files (x86)")]
        public void TryGetExactPathName_Should_output_path_with_exact_casing(string path)
        {
            var pathLowered = path.ToLower();
            var isExistingOnFileSystem = _controller.TryGetExactPath(pathLowered, out string exactPath);

            Assert.IsTrue(isExistingOnFileSystem);
            Assert.AreEqual(pathLowered, exactPath.ToLower(), "The path should be still equivalent");
            Assert.AreNotEqual(pathLowered, exactPath, "The path case has been changed");
        }
```


PS: The test on @"C:\pagefile.sys" has been removed because I discovered that the call to the native api are not working with files at the root of the drive 👿  😭

pmiossec:
I have pushed my changes here: https://github.com/pmiossec/gitextensions/tree/fix_casing_test

gerhardol:
> I have pushed my changes here: https://github.com/pmiossec/gitextensions/tree/fix_casing_test

The modified test works for me - but it only failed sporadically before.
I still believe that the tests should not test local file system paths. There is a test for TryGetExactPath() already. You can control where you install also system files. At least the check for "Primary Interop Assemblies" should be dropped.

If the modified test is the majority vote (RussKie has the majority of votes regardless...) I suggest @pmiossec submits a PR. I can ofcourse pick 307f86a922bd67e16a076f763ae829c8d3f623ad too.

gerhardol:
@RussKie @pmiossec 
Can we decide on merging this or #7159 ?
I am not convinced that a test driven approach is viable when I have to comment out incorrect tests.

PR #7116, opened 2019-09-08
correctly call hMSBuild batch file

mdonatas:
Fixes a problem where a call to hMSBuild would be the last thing executed in a batch file

## Proposed changes / problem description

- From what I gather from reading about executing nested batch files is if you simply call a batch from withing another batch file then it's executed inline and if there is an `exit` in the nested file the root batch exits as it's as if there was just one big batch file. To properly executed a nested batch files one needs to:
- Use `call` to execute `hMSBuild`

more on this: https://stackoverflow.com/questions/1103994/how-to-run-multiple-bat-files-within-a-bat-file

This raises another question... are these batch files used at all? Without this fix `MakeInstallers.cmd` would not produce an installer for me.

### DIFF
--- a/Setup/Build.cmd
+++ b/Setup/Build.cmd
@@ -15,5 +15,5 @@ set msbuildparams=/p:Configuration=%Configuration% /t:Rebuild /nologo /v:m
 call BuildGitExtNative.cmd %Configuration% Rebuild
 IF ERRORLEVEL 1 EXIT /B 1
 
-%msbuild32% %solution% /p:Platform="Any CPU" %msbuildparams%
+call %msbuild32% %solution% /p:Platform="Any CPU" %msbuildparams%
 IF ERRORLEVEL 1 EXIT /B 1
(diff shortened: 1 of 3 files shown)


gerhardol:
> This raises another question... are these batch files used at all? Without this fix MakeInstallers.cmd would not produce an installer for me.

The installer are built on AppVeyor with other scripts /release has branched build scripts right now).

mdonatas:
Are forked repos built there too? How do others install custom builds?

gerhardol:
If you want to run a custom build, just run it from the build folder.

The only reason to use the installer is to test changes to the installer or to copy binaries to Program Files (and you do not want to use the explorer) to use the build also for shell/VS integration.

mdonatas:
Heh, interesting :) I've been running custom builds with extra features and small tweaks for years so I need an installer for fresh machines or when I rebase on master and have a fresh build.

PR #7123, opened 2019-09-11
fix: Unable to intstall new version

### LINKED ISSUE #7120: Latest Version of Git Extensions Installer Always Reverts For Global user Install
## Current behavior

Installer should always work, prompt for install happened today, and had a new user install it. 

## Expected behavior

Installer should work.

## Screenshots

Install latest MSI for all users. 

3.2.0.6586.msi, I'm using 3.1, do not know which MSI

## Environment

- GitExtensions version:
- GIT version:  unrelated. 
- OS version:  Windows 10
- .NET version:  both standard user machine, and dev machine with many versions installed. 

## Diagnostics
Telemetry was not enabled during the install .

RussKie:
Resolves #7120

## Proposed changes

* There are few issues with the install script related to configuring the telemetry settings.

    1. On W7 there is only PowerShell 2.0 (that is what ships with .NET 2.0). On the later versions of OS there are different versions of PowerShell and thus requiring a specific version would lead to failures.

    2. A problem related to the logic responsible for self-removal

       * First `$MyINvocation.InvocationName` returned different and unexpected         values when a script is run not from a command line but invoked as a     script-block. In the latter case, `$MyINvocation.InvocationName` is     set to `&`.

       * Secondly the installer somehow would fail to remove the script     (permissions?)

  The self-removal logic is removed.

* Bump the version to 3.2.1 to ensure the new fix flows through to all distribution channels correctly.
 

## Test methodology 

- Manual execution of the script
- Verification reports from users

----

### DIFF
--- a/Setup/UI/TelemetryDlg.wxs
+++ b/Setup/UI/TelemetryDlg.wxs
@@ -14,11 +14,10 @@
       <![CDATA[Installed OR POWERSHELLEXE]]>
     </Condition>
 
-    <!-- Windows 7SP1 supports only PowerShell 2.0 -->
     <SetProperty Id="ConfigureTelemetry"
         Before ="ConfigureTelemetry"
         Sequence="execute"
-        Value="&quot;[POWERSHELLEXE]&quot; -Version 2.0 -NoProfile -NonInteractive -InputFormat None -ExecutionPolicy Bypass -Command &quot;&amp; '[INSTALLDIR]set-telemetry.ps1' [TELEMETRY_ENABLED] ; exit $$($Error.Count)&quot;" />
+        Value="&quot;[POWERSHELLEXE]&quot; -NoProfile -NonInteractive -InputFormat None -ExecutionPolicy Bypass -Command &quot;&amp; '[INSTALLDIR]set-telemetry.ps1' [TELEMETRY_ENABLED]; exit $$($Error.Count)&quot;" />
 
     <CustomAction Id="ConfigureTelemetry" BinaryKey="WixCA" DllEntry="WixQuietExec" Execute="deferred" Return="check" Impersonate="yes" />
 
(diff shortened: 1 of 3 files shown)


Setup/UI/TelemetryDlg.wxs
```diff
@@ -14,11 +14,10 @@
       <![CDATA[Installed OR POWERSHELLEXE]]>
     </Condition>
 
-    <!-- Windows 7SP1 supports only PowerShell 2.0 -->
```
gerhardol: Why remove this comment, even if not forcing v2.0?

mast-eu: It was kind of misleading. Win7 SP1 supports _at maximum_ PowerShell 2.0, but there's no guarantee that it is actually available on all Win7 systems. It depends on the installed .NET Framework version.

RussKie:
Thank you.
Can you see from logs what's causing the failure?

mast-eu:
It still seems to be the "usual" error code 1603.
```
MSI (s) (F0:D8) [22:31:01:731]: Executing op: CustomActionSchedule(Action=ConfigureTelemetry,ActionType=1025,Source=BinaryData,Target=WixQuietExec,CustomActionData="C:\Windows\SysWOW64\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -NonInteractive -InputFormat None -ExecutionPolicy Bypass -Command "& 'C:\Program Files (x86)\GitExtensions\set-telemetry.ps1' 'C:\Program Files (x86)\GitExtensions\ ' 1; exit $($Error.Count)")
MSI (s) (F0:D4) [22:31:01:733]: Invoking remote custom action. DLL: C:\Windows\Installer\MSIBEBA.tmp, Entrypoint: WixQuietExec
WixQuietExec:  Error 0x80070001: Command line returned an error.
WixQuietExec:  Error 0x80070001: QuietExec Failed
WixQuietExec:  Error 0x80070001: Failed in ExecCommon method
CustomAction ConfigureTelemetry returned actual error code 1603 (note this may not be 100% accurate if translation happened inside sandbox)
Action ended 22:31:02: InstallFinalize. Return value 3.
```

Issue #3745, opened 2017-06-05
Tooltips are not translated


pmiossec:
2 questions about the translation in GitExtensions...

1. Should we add (or replace ? ) the link to translate GitExtensions to open the transifex website instead of open the tranlation application now that the translation is done through the website?

2. I see that none of the tooltip texts are translated in the application nor proposed to the translation in the 'xlf' files. Is someone know if something has already been done to support tooltips?

jbialobr:
1. Yes, we should direct uers to the Transifex website..
2. Can you list them? Perhaps it is a bug. I can see:

`<trans-unit id="DecreaseNumberOfLines.ToolTipText">`

![gitext](https://cloud.githubusercontent.com/assets/1003909/26815692/f21e92e6-4a8d-11e7-9e9a-b00a2c2c9684.png)

pmiossec:
1. I will do a PR to fix that
2.  Indeed, that works for control that have a `ToolTipText` but the one that rely on a `System.Windows.Forms.ToolTip` control and call `SetToolTip()` seems to be not translated (like in the `FormPull`). Do you confirm?

jbialobr:
>Do you confirm?

Yes, that is true.

RussKie:
Can this be closed?

pmiossec:
The first point has been solved by a PR.

The second is still not fixed.
So, I don't know. Perhaps opening a clearer issue.

Anyway, do what you want...

RussKie:
I can see tooltips being translated.
As per @jbialobr's recommendation, if you see any tooltips not translated please raise a new issue.

PR #7128, opened 2019-09-12
Unblock special commit message

### LINKED ISSUE #6993: Commit Message Text field stays deactivated after fixup commit
## Current behaviour

* Start making a fixup commit
  * Commit message text field is deactivated
* Do the commit
  * Commit window stays open (as configured)
  * Commit message text field stays deactivated
  * Also other controls like checkbox for amend commit stays deactivated
* Close and reopen commit window
  * Commit message text field is activated again

## Expected behaviour

After making a fixup commit all controls, in special the commit message text field, should be activated again

## Workaround

Reopen commit window

## Environment

- GitExtensions version: 3.1.1.6049 (Build 2f872105b4eb42421d9fdafd68e90a9e27d20d96)
- GIT version: git version 2.22.0.windows.1
- OS version: Windows 10 1607

### LINKED ISSUE #7104: Don't block commit message textbox for fixup! squash! commits
## Don't block commit message textbox for fixup! or squash! commits

I feel that changes
> disable changing the commit message and amending a commit unless CommitKind.Normal

introduced in https://github.com/gitextensions/gitextensions/pull/6530 limit the user in some situations which are valid:

- Open commit dialog in fixup mode and change your mind - previously we could've just deleted the text
- Turn current latest commit into a fixup commit.. not possible without typing fixup! manually as amend is disabled
- Open commit dialog in fixup mode, commit part of the files and the make another regular commit (covered by #6993)

I find myself hitting the "readonly wall" quite often in at least two of the above cases.

The proposal would be to revert this behavior to enable more flexible usage scenarios.

## Environment

- Git Extensions 3.2.0
- Build 3010403a6229cafa8a364f60ec2a2ee4517fcb26 (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 192dpi (200% scaling)

mstv:
Fixes #7104
Fixes #6993

## Proposed changes

- add button to enable editing the special commit message of `fixup!` and `squash!` commits
- add/adapt hotkeys for buttons
- allow `Amend commit'` for any `CommitKind`
- reset `CommitKind` after commiting without closing `FormCommit`
- slightly off topic: add more test cases for empty lines in commit messages (addendum to #6877)

### Before

grafik

### After

grafik
(different message font due to different font settings vs. portable build)
grafik

### (Intermediate with ToolStripItemButton)
grafik

### Before

grafik

### After

grafik

## Test methodology 

- manual tests

## Test environment(s) 

- Git Extensions 3.3.0
- Build 79dee30803b7eb1ae0ff5999ff3bb962a9cacff7
- Git 2.23.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -118,6 +118,7 @@ namespace GitUI.CommandsDialogs
             this.SelectedDiff = new GitUI.Editor.FileViewer();
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
             this.Message = new GitUI.SpellChecker.EditNetSpell();
+            this.modifyCommitMessageButton = new System.Windows.Forms.Button();
             this.flowCommitButtons = new System.Windows.Forms.FlowLayoutPanel();
             this.Commit = new System.Windows.Forms.Button();
             this.CommitAndPush = new System.Windows.Forms.Button();
@@ -318,7 +319,7 @@ namespace GitUI.CommandsDialogs
             this.toolStripSeparator5.Name = "toolStripSeparator5";
             this.toolStripSeparator5.Size = new System.Drawing.Size(229, 6);
             //
-            // addFileTogitignoreToolStripMenuItem
+            // addFileToGitIgnoreToolStripMenuItem
             //
             this.addFileToGitIgnoreToolStripMenuItem.Image = global::GitUI.Properties.Images.AddToGitIgnore;
             this.addFileToGitIgnoreToolStripMenuItem.Name = "addFileToGitIgnoreToolStripMenuItem";
@@ -326,7 +327,7 @@ namespace GitUI.CommandsDialogs
             this.addFileToGitIgnoreToolStripMenuItem.Text = "Add file to .gitignore";
             this.addFileToGitIgnoreToolStripMenuItem.Click += new System.EventHandler(this.AddFileToGitIgnoreToolStripMenuItemClick);
             //
-            // addFileTogitinfoexcludeExcludeLocallyToolStripMenuItem
+            // addFileToGitInfoExcludeLocallyToolStripMenuItem
             //
             this.addFileToGitInfoExcludeLocallyToolStripMenuItem.Image = global::GitUI.Properties.Images.AddToGitIgnore;
             this.addFileToGitInfoExcludeLocallyToolStripMenuItem.Name = "addFileToGitInfoExcludeLocallyToolStripMenuItem";
@@ -369,7 +370,7 @@ namespace GitUI.CommandsDialogs
             this.toolStripSeparator4.Name = "toolStripSeparator4";
             this.toolStripSeparator4.Size = new System.Drawing.Size(229, 6);
             //
-            // interactiveAddtoolStripMenuItem
+            // interactiveAddToolStripMenuItem
             //
             this.interactiveAddToolStripMenuItem.Name = "interactiveAddToolStripMenuItem";
             this.interactiveAddToolStripMenuItem.Size = new System.Drawing.Size(232, 22);
@@ -822,7 +823,7 @@ namespace GitUI.CommandsDialogs
             this.resetUnstagedChangesToolStripMenuItem.Text = "Reset unstaged changes";
             this.resetUnstagedChangesToolStripMenuItem.Click += new System.EventHandler(this.resetUnstagedChangesToolStripMenuItem_Click);
             //
-            // resetAlltrackedChangesToolStripMenuItem
+            // resetAllTrackedChangesToolStripMenuItem
             //
             this.resetAllTrackedChangesToolStripMenuItem.Image = global::GitUI.Properties.Images.ResetWorkingDirChanges;
             this.resetAllTrackedChangesToolStripMenuItem.Name = "resetAllTrackedChangesToolStripMenuItem";
@@ -1033,11 +1034,26 @@ namespace GitUI.CommandsDialogs
             //
             // splitRight.Panel2
             //
+            this.splitRight.Panel2.Controls.Add(this.modifyCommitMessageButton);
             this.splitRight.Panel2.Controls.Add(this.tableLayoutPanel1);
             this.splitRight.Size = new System.Drawing.Size(517, 622);
             this.splitRight.SplitterDistance = 426;
             this.splitRight.TabIndex = 0;
             this.splitRight.TabStop = false;
+            //
+            // modifyCommitMessageButton
+            //
+            this.modifyCommitMessageButton.AutoSize = true;
+            this.modifyCommitMessageButton.BackColor = System.Drawing.SystemColors.ButtonFace;
+            this.modifyCommitMessageButton.FlatStyle = System.Windows.Forms.FlatStyle.Flat;
+            this.modifyCommitMessageButton.ForeColor = System.Drawing.SystemColors.HotTrack;
+            this.modifyCommitMessageButton.Location = new System.Drawing.Point(185, 55);
+            this.modifyCommitMessageButton.Name = "modifyCommitMessageButton";
+            this.modifyCommitMessageButton.Size = new System.Drawing.Size(149, 25);
+            this.modifyCommitMessageButton.TabIndex = 9;
+            this.modifyCommitMessageButton.Text = "Modify the commit m&essage";
+            this.modifyCommitMessageButton.UseVisualStyleBackColor = false;
+            this.modifyCommitMessageButton.Click += new System.EventHandler(this.modifyCommitMessageButton_Click);
             //
             // SolveMergeconflicts
             //
@@ -1469,7 +1485,7 @@ namespace GitUI.CommandsDialogs
             this.stopTrackingThisFileToolStripMenuItem.Size = new System.Drawing.Size(428, 38);
             this.stopTrackingThisFileToolStripMenuItem.Text = "Stop tracking this file";
             this.stopTrackingThisFileToolStripMenuItem.Click += new System.EventHandler(this.stopTrackingThisFileToolStripMenuItem_Click);
-            // 
+            //
             // FormCommit
             //
             this.AutoScaleDimensions = new System.Drawing.SizeF(96F, 96F);
@@ -1661,5 +1677,6 @@ namespace GitUI.CommandsDialogs
         private ToolStripTextBox toolStripGpgKeyTextBox;
         private ToolStripComboBox gpgSignCommitToolStripComboBox;
         private ToolStripMenuItem stopTrackingThisFileToolStripMenuItem;
+        private Button modifyCommitMessageButton;
     }
 }
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -337,8 +337,6 @@ namespace GitUI.CommandsDialogs
 
             void ConfigureMessageBox()
             {
-                Amend.Enabled = _commitKind == CommitKind.Normal;
-
                 bool messageCanBeChanged = _useFormCommitMessage && _commitKind == CommitKind.Normal;
 
                 Message.Enabled = messageCanBeChanged;
@@ -1086,7 +1084,7 @@ namespace GitUI.CommandsDialogs
             LoadingStaged.Visible = false;
             Commit.Enabled = true;
             CommitAndPush.Enabled = true;
-            Amend.Enabled = _commitKind == CommitKind.Normal;
+            Amend.Enabled = true;
             Reset.Enabled = doChangesExist;
 
             EnableStageButtons(true);
@@ -1981,7 +1979,7 @@ namespace GitUI.CommandsDialogs
                 EnableStageButtons(true);
 
                 Commit.Enabled = true;
-                Amend.Enabled = _commitKind == CommitKind.Normal;
+                Amend.Enabled = true;
             }
 
             if (AppSettings.RevisionGraphShowWorkingDirChanges)
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -129,6 +129,10 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _commitCommitterInfo = new TranslationString("Committer");
         private readonly TranslationString _commitCommitterToolTip = new TranslationString("Click to change committer information.");
 
+        private readonly TranslationString _modifyCommitMessageButtonToolTip
+            = new TranslationString("If you change the first line of the commit message, git will treat this commit as an ordinary commit," + Environment.NewLine
+                                    + "i.e. it may no longer be a fixup or an autosquash commit.");
+
         private readonly TranslationString _templateNotFoundCaption = new TranslationString("Template Error");
         private readonly TranslationString _templateNotFound = new TranslationString($"Template not found: {{0}}.{Environment.NewLine}{Environment.NewLine}You can set your template:{Environment.NewLine}\t$ git config commit.template ./.git_commit_msg.txt{Environment.NewLine}You can unset the template:{Environment.NewLine}\t$ git config --unset commit.template");
         private readonly TranslationString _templateLoadErrorCaption = new TranslationString("Template could not be loaded");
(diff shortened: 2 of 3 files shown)


mstv:
Added hotkeys also for the `Reset` buttons, @mrexodia.

mrexodia:
Cool, thanks!

RussKie:
> * add/adapt hotkeys for buttons

I propose this is submitted as its own PR, as it will get merged quickly.

> * slightly off topic: add more test cases for empty lines in commit messages (addendum to #6877)

So as this.

> * allow `Amend commit'` for any `CommitKind`

Could you please explain why this is a good change?

> * add button to enable editing the special commit message of `fixup!` and `squash!` commits
> grafik

I do not disagree with the change, but I am not fond of the proposed implementation.
There is no immediate connection (visual or otherwise) between the textbox and the button.

I think we provide a better user experience by rendering a link label on top of the textbox (z-order wise), e.g. something like the following:
image
The text on the link is up for a discussion (it could be "cancel fixup"?).

mstv:
> allow Amend commit' for any CommitKind

https://github.com/gitextensions/gitextensions/issues/7104#issuecomment-528749535: "It's a totally valid scenario to turn the latest commit into a fixup! commit."

> I think we provide a better user experience by rendering a link label on top of the textbox (z-order wise)

Understood. Can it be activated using a shortcut key?

> The text on the link is up for a discussion (it could be "cancel fixup"?).

I'd prefer "Modify this special commit message", perhaps with an addition like "(The warrenty ends here.)". (I know we don't give warrenties and should never do.)
It is *not* cancelling the fixup: The message will not be changed by clicking. Just `CommitKind` is switched to `Normal` in order to store the manually entered commit message on (temporary) close of FormCommit.

RussKie:
> Understood. Can it be activated using a shortcut key?

I would think it can.
https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.label.usemnemonic?view=netframework-4.8#System_Windows_Forms_Label_UseMnemonic

> I'd prefer "Modify this special commit message", perhaps with an addition like "(The warrenty ends here.)".

We can have a tooltip that warns users (e.g. "do it at your own risk" kind of thing)

mstv:
Fiddled around with `[Link]Label` without success. As I explained above, we need a button-like control, not a label which controls only the focus - the documentation (which I read too late) states:
"Pressing ALT + the mnemonic character sets the focus to the control that follows the Label in the tab order."
BTW: The VS 2019 Designer does not understand `FormCommit` completely. And it did not allow to place `[Link]Label` in the area of the message editor.
grafik

RussKie:
I'll try to find some time and fiddle with it

mstv:
I turned the `ToolStripItemButton` into a `Label`, into a `LinkLabel` and finally into a `Button` in dedicated commits.
I think this is a good compromise:
grafik

TODO:
* make ToolTip translatable:
The `TranslationApp` seems to ignore ToolTips. Must I really create a `TranslationString` by hand?
* squash

gerhardol:
> Does it look OK with a dark theme?

It is simple to test with high contrast dark theme (nothing looks OK, But the button should be visible. 
To install a real dark system the only simple way is to use WindowsBlinds. Paid app, but have a trial.
https://github.com/gitextensions/gitextensions/pull/6651#issuecomment-495798173

mdonatas:
Food for thought. From UX perspective it seems weird to make a text-area read only and then place a button to make it editable... for what reason exactly? There is no explanation and adding an explanation would only dig this hole deeper.
Git users know that there is nothing special with `fixup!` and `squash!` they are just guidelines for interactive rebase to automate reordering rows.
If someone is going to do interactive rebase they are supposedly git savvy enough to know what they are doing and then we really just need a "heads up" UX signalling that this is a commit message with some special meaning. So something like this seems to hit the right balance:
image

#KeepItSimple :)

RussKie:
I like a different perspective, thanks @mdonatas.

Will a commit be treated as a fixup if the text changed though?

gerhardol:
> Will a commit be treated as a fixup if the text changed though?

The subject after fixup must match. 
If you change it is something else, which is what the issue is about. 

As the button is implemented, I prefer the guidence it provides but I do not have a strong preference.

mdonatas:
The subject must match. But that's the same if you are using `git` from the shell. If the user didn't know that subjects must match, this would be a good learning experience when the lines wouldn't get reordered in an interactive rebase. After all no harm is done at all.

_offtopic_
IMO there are way too many guard rails on by default already. For instance git embraces commit graph modifications with rebase and amending, there is even a saying `rebase makes it look as if you did everything perfect the first time` but GE scares you with dialogs about history rewrite which is one of the pillar features of git. And sadly very few go to change these settings.
I recommend GE to everyone because it exposes `git` features in GUI form without interpretations or restrictions as closely as possible meaning GE embraces `git` as opposed to crippling it. I am voting with both hands for continuing to embrace `git` as one of the GE pillar features. 🕊

pmiossec:
>GE scares you with dialogs about history rewrite which is one of the pillar features of git.

The perfect thing would have been to display this message **only** when the commit has been already pushed.

At least we could update the warning message to explain that.

RussKie:
> _offtopic_
> IMO there are way too many guard rails on by default already. For instance git embraces commit graph modifications with rebase and amending, there is even a saying `rebase makes it look as if you did everything perfect the first time` but GE scares you with dialogs about history rewrite which is one of the pillar features of git. And sadly very few go to change these settings.

Would you like to start a discussion about these?
I am getting tired of certain pop-ups and I would like to dial them down. Maybe even discuss creating "notice" vs "adept" modes which may have different levels of safeguards.

RussKie:
> The perfect thing would have been to display this message **only** when the commit has been already pushed.

How do you see it implemented?

RussKie:
> The subject must match.

If subject must match to the dot, it is not quite unreasonable to make the textbox readonly to indicate it to the user that it isn't something they should be changing.
However to allow the use-case where a user may change their mind and do another type of commit (i.e. not a fixup), we need to make the textbox editable.

If we leave the textbox editable for the special types of commits (i.e. fixup) our users may be left confused, change text and then wonder why they didn't get a fixup commit they wanted.

I'm speaking from first-hand experience where I had/have to teach engineers (including smart senior level devs and architects) about these git commands, and a lot of time I need(ed) to hold their hands with git. The same way others held my hand when I was learning git.

mrexodia:
With regards to detecting if something is pushed, I guess you can check
where origin/branch is and only warn if the rebase is in the past of
origin/branch.

The way me (and my team) are using git is that we rebase and force push
feature branches until everything is clean so the message is useless anyway
(similar to the force push warning). For that the concept of “protected”
branches could be explored, but generally the server is enforcing it so
it’s probably not worth it.

gerhardol:
>> offtopic

> Would you like to start a discussion about these?
I am getting tired of certain pop-ups and I would like to dial them down.

We should maybe open a separate issue to discuss this?
I do not mind a few OT posts but we should not go to far away from this change.
(As said before, I prefer guidance here, the old way to 'break-out' was not too bad for me.)

Traditional version control systems do not allow history rewriting so this is a change of mindset that does not come instantly. As Git is becoming the dominant VCS, I feel the trend is going to from banning rewrite via allowing private changes to enforcing rewrites. At work we require --ff-only, the history gets much more readable.

Maybe popup preset based on workflow is a way forward. Some warnings like amend should be reworded too.
Rules to follow the server settings like not rewrite protected branches like master should maybe be added too (that should be added to GitExtensions.settings for repo to be effective.)

RussKie:
> We should maybe open a separate issue to discuss this?

That was my suggestion :) Sorry, I should have been a bit clearer.

GitUI/Translation/English.xlf
```diff
@@ -3568,6 +3568,10 @@ You can unset the template:
         <source>Interactive Add</source>
         <target />
       </trans-unit>
+      <trans-unit id="modifyCommitMessageButton.Text">
+        <source>Modify this special commit m&amp;essage</source>
+        <target />
+      </trans-unit>
```
mstv: The TranslationApp seems to ignore ToolTips. Must I really create a TranslationString by hand?

RussKie:
Hmm... I vaguely remember we had issues with translations for tooltips, but
I don't remember whether it was resolved.
I speculate but I don't think we set tooltips in designer files.

pmiossec:
I also don't remember but perhaps #3745 could help. Especially, https://github.com/gitextensions/gitextensions/issues/3745#issuecomment-306492000

mstv:
Added tooltip translation and squashed.

GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -1140,7 +1156,7 @@ private void InitializeComponent()
             this.CommitAndPush.Size = new System.Drawing.Size(171, 26);
             this.CommitAndPush.TabIndex = 9;
             this.CommitAndPush.TabStop = false;
-            this.CommitAndPush.Text = "C&ommit && push";
+            this.CommitAndPush.Text = "Commit && &push";
```
RussKie: Changes like this look very sneaky. The PR's description briefly mentions this change, but it doesn't go into details as to what changed.
It won't be an easy/obvious bisect to find when and what changed, if users complain.

Please extract the changes to mnemonics in a separate commit and clearly describe the change.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -803,6 +789,21 @@ IReadOnlyList<GitItemStatus> GetAllChangedFilesWithSubmodulesStatus()
             }
         }
 
+        private void ConfigureMessageBox()
```
```suggestion
        private void ConfigureMessageBox(CommitKind commitKind, bool useFormCommitMessage)
```

This way we can test it. Also please add tests.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -3297,25 +3330,6 @@ internal TestAccessor(FormCommit formCommit)
                 string commitMessageText, bool usingCommitTemplate, bool ensureCommitMessageSecondLineEmpty)
                 => FormCommit.FormatCommitMessageFromTextBox(commitMessageText, usingCommitTemplate, ensureCommitMessageSecondLineEmpty);
```
```suggestion
                    => FormCommit.FormatCommitMessageFromTextBox(commitMessageText, usingCommitTemplate, ensureCommitMessageSecondLineEmpty);
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -129,6 +129,9 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly TranslationString _commitCommitterInfo = new TranslationString("Committer");
         private readonly TranslationString _commitCommitterToolTip = new TranslationString("Click to change committer information.");
 
+        private readonly TranslationString _modifyCommitMessageButtonToolTip
+            = new TranslationString("Enable the modification of this special commit message (Do it at your own risk!)");
```
```suggestion
            = new TranslationString("If you change the commit message git will treat this commit as a ordinary commit (i.e. it may no longer be a fixup or an autosquash commit)");
```

I'm not sure if newline is rendered in tooltips, if it is - perhaps, add a newline before the left paren.

GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -1033,12 +1034,27 @@ private void InitializeComponent()
...
+            //
+            this.modifyCommitMessageButton.AutoSize = true;
+            this.modifyCommitMessageButton.BackColor = System.Drawing.SystemColors.ButtonFace;
+            this.modifyCommitMessageButton.FlatStyle = System.Windows.Forms.FlatStyle.Flat;
+            this.modifyCommitMessageButton.ForeColor = System.Drawing.SystemColors.HotTrack;
+            this.modifyCommitMessageButton.Location = new System.Drawing.Point(185, 55);
+            this.modifyCommitMessageButton.Name = "modifyCommitMessageButton";
+            this.modifyCommitMessageButton.Size = new System.Drawing.Size(186, 25);
+            this.modifyCommitMessageButton.TabIndex = 9;
+            this.modifyCommitMessageButton.Text = "Modify this special commit m&essage";
```
RussKie: It feels we're creating a new terminology here. I couldn't find any references to "special commit" (or "normal commits" for that matter).
Will it be less clear if we word as follows?
```suggestion
            this.modifyCommitMessageButton.Text = "Modify the commit m&essage";
```

GitUI/CommandsDialogs/FormCommit.cs
mstv: Pro: It emphasizes the depencies.
The drawback would be that all three callers must specify these options without being able to declare `CommitInfo` `static`. It could still evaluate other members.

I don't see any substantial benefit to add a test for a method like this.
Test cases would be only a guard against the modification of this very simple logic (in addition to the mandatory PR review).
```
       private void ConfigureMessageBox(CommitKind commitKind, bool useFormCommitMessage)
        {
            modifyCommitMessageButton.Visible = useFormCommitMessage && commitKind != CommitKind.Normal;
            bool messageCanBeChanged = useFormCommitMessage && commitKind == CommitKind.Normal;

            Message.Enabled = messageCanBeChanged;

            commitMessageToolStripMenuItem.Enabled = messageCanBeChanged;
            commitTemplatesToolStripMenuItem.Enabled = messageCanBeChanged;

            Message.WatermarkText = useFormCommitMessage
                ? _enterCommitMessageHint.Text
                : _commitMessageDisabled.Text;
        }
```

GitUI/CommandsDialogs/FormCommit.Designer.cs
mstv: It is a "special (commit) _message_". I understand your objection against introducing new terminology. The new tooltip describes it clear enough.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -3263,6 +3265,37 @@ private void Message_Enter(object sender, EventArgs e)
             }
         }
 
+        private void EnableEditSpecialCommitMessage()
+        {
+            _commitKind = CommitKind.Normal;
+            ConfigureMessageBox();
+        }
```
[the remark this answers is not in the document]
mstv: Since `_useFormCommitMessage` is `readonly`, the set method of a new property `CommitKind` is the most correct replacement.

GitUI/CommandsDialogs/FormCommit.Designer.cs
mstv: Moved the submarines to dedicated PRs. In the future, I'll remember that mixing things - although related - makes it more difficult to review and to bisect.
Unfortunately, there's little support from Git developing things in parallel in the same working copy.

GitUI/CommandsDialogs/FormCommit.cs
mstv: I guess this has become obsolete with the new property `CommitKind`.
Your bad feeling regarding this method was justified, once more. 👍

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -129,6 +129,10 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly TranslationString _commitCommitterInfo = new TranslationString("Committer");
         private readonly TranslationString _commitCommitterToolTip = new TranslationString("Click to change committer information.");
 
+        private readonly TranslationString _modifyCommitMessageButtonToolTip
+            = new TranslationString("If you change the first line of the commit message, Git will treat this commit as an ordinary commit." + Environment.NewLine
```
```suggestion
            = new TranslationString("If you change the first line of the commit message, git will treat this commit as an ordinary commit," + Environment.NewLine
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -129,6 +129,10 @@ public sealed partial class FormCommit : GitModuleForm
...
+                                    + "I.e. it may no longer be a fixup or an autosquash commit.");
```
```suggestion
                                    + "i.e. it may no longer be a fixup or an autosquash commit.");
```

mstv:
Sorry for merging too early, I expected that the msftbot was configured to wait for at least one approval.
@RussKie, can the msftbot's default behavior be configured? Or must I always type something like
`merge 1 day after approval by @RussKie or after two approvals`

GitUI/CommandsDialogs/FormCommit.cs
mdonatas: How about this

> If commit message following squash! or fixup! will not match any commits in an interactive rebase this commit will not be marked for squashing or fixup.

The grammar or exact working can be changed but the idea is to not "invoke magic" by mentioning "regular" commits as if this is anything but. Fact is it's just a regular commit i.e. there is nothing in its metadata to say otherwise and only if you will proceed to do an interactive rebase the message _might_ have some effect.

RussKie:
Yes, you need to write something like "@msftbot merge if @desired_user_name approves"
Have a look at [bot's guide](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Advanced-auto-merge)

mstv:
> Have a look at bot's guide

I had; and I had the hope the instantly handled "status: auto merge" could be tamed - by the repo admin at least.
When you used it, it wrote something like you had reset its configuration. That's why I thought the default behavior could be configured.
Perhaps it can be added to the PR template.

PR #7148, opened 2019-09-15
No more "..." indicator for revisions with single-line commit message

### LINKED ISSUE #7126: "..." indicator [not] for all revisions
## Current behaviour

The more info indicator `...` is shown for every commit in the revision grid.

grafik

## Expected behaviour

No `...` for single-line commit messages, no tooltip.

grafik

## Steps to reproduce

open a repo with single-line commits

yes, regression after 3.2.0.6286 7cbb45bb5148e16a217e9804c596070d41ce78f7

## Environment

- Git Extensions 3.2.0.6326
- Build f51e398e71888771311356e8e834bbf99e6e24cf (Dirty)
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

and also

- Git Extensions 3.3.0
- Build 114b8ed39007ecb060f39264daecc217d54a100b (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

mstv:
Fixes #7126

## Proposed changes

- consider the end-of-body marker everywhere in the detection of single-line commit messages

### Before

grafik

### After

grafik

## Test methodology 

- added NUnit tests

## Test environment(s) 

- Git Extensions 3.3.0
- Build fac68fef6c8b41f779e68b965936d9baca94198c
- Git 2.23.0.windows.1
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -382,7 +382,7 @@ namespace GitCommands
 
             #endregion
 
-            #region Encoded string values (names, emails, subject, body, name)
+            #region Encoded string values (names, emails, subject, body, [file]name)
 
             // Finally, decode the names, email, subject and body strings using the required text encoding
             var s = encoding.GetString(array, offset, lastOffset - offset);
@@ -407,8 +407,7 @@ namespace GitCommands
             // NOTE the convention is that the Subject string is duplicated at the start of the Body string
             // Therefore we read the subject twice.
             // If there are not enough characters remaining for a body, then just assign the subject string directly.
-            var body = reader.Remaining - subject.Length == 2 ? subject : reader.ReadToEnd();
-
+            var (body, additionalData) = ParseCommitBody(reader, subject);
             if (body == null)
             {
                 // TODO log this parse error
@@ -416,17 +415,6 @@ namespace GitCommands
                 revision = default;
                 return false;
             }
-
-            var indexOfEndOfBody = body.LastIndexOf(EndOfBody, StringComparison.InvariantCulture);
-
-            string additionalData = null;
-            var bodyContainsAdditionalData = body.Length > indexOfEndOfBody + EndOfBody.Length;
-            if (bodyContainsAdditionalData)
-            {
-                additionalData = body.Substring(indexOfEndOfBody + EndOfBody.Length).TrimStart();
-            }
-
-            body = body.Substring(0, indexOfEndOfBody);
 
             #endregion
 
@@ -451,6 +439,35 @@ namespace GitCommands
             return true;
         }
 
+        [CanBeNull]
+        private static (string body, string additionalData) ParseCommitBody([NotNull] StringLineReader reader, [NotNull] string subject)
+        {
+            int lengthOfSubjectRepeatedInBody = subject.Length + 2/*newlines*/;
+            if (reader.Remaining == lengthOfSubjectRepeatedInBody + EndOfBody.Length)
+            {
+                return (body: subject, additionalData: null);
+            }
+
+            string tail = reader.ReadToEnd() ?? "";
+            int indexOfEndOfBody = tail.LastIndexOf(EndOfBody, StringComparison.InvariantCulture);
+            if (indexOfEndOfBody < 0)
+            {
+                // TODO log this parse error
+                Debug.Fail("Missing end-of-body marker in the log -- this should not happen");
+                return (body: null, additionalData: null);
+            }
+
+            string additionalData = null;
+            if (tail.Length > indexOfEndOfBody + EndOfBody.Length)
+            {
+                additionalData = tail.Substring(indexOfEndOfBody + EndOfBody.Length).TrimStart();
+            }
+
+            string body = indexOfEndOfBody == lengthOfSubjectRepeatedInBody
+                          ? subject : tail.Substring(0, indexOfEndOfBody).TrimEnd();
+            return (body, additionalData);
+        }
+
         public void Dispose()
         {
             _cancellationTokenSequence.Dispose();
@@ -461,7 +478,7 @@ namespace GitCommands
         /// <summary>
         /// Simple type to walk along a string, line by line, without redundant allocations.
         /// </summary>
-        private struct StringLineReader
+        internal struct StringLineReader
         {
             private readonly string _s;
             private int _index;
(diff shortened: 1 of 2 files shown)


GitCommands/RevisionReader.cs
```diff
@@ -407,7 +407,32 @@ DateTime ParseUnixDateTime()
...
+                }
+
+                var containsAdditionalData = tail.Length > indexOfEndOfBody + EndOfBody.Length;
+                if (containsAdditionalData)
+                {
+                    additionalData = tail.Substring(indexOfEndOfBody + EndOfBody.Length).TrimStart();
+                }
+
+                body = indexOfEndOfBody == subject.Length + 2/*newlines*/ ? subject : tail.Substring(0, indexOfEndOfBody).TrimEnd();
+            }
```
RussKie: I propose we extract the functionality as a method that we can test.

I think the method could have the following signature:
```cs
private (string body, string additionalData) ParseCommitBody(...)
```

The check `if (indexOfEndOfBody < 0)` will return `null` and we won't need to duplicate the code for `if (body == null)` check.

mstv: Agreed.
What do you think about the optimization using `subject.Length + 2/*newlines*/` (which must be done even twice)?

RussKie: I suppose we could use a temp var, but I don't have particularly strong feelings about it

mstv: My point was: Only the length is checked not the contents of body.

mstv: Done, factored out the function and added tests.

GitCommands/RevisionReader.cs
```diff
@@ -451,17 +439,54 @@ DateTime ParseUnixDateTime()
...
+
+            string additionalData = null;
+            if (tail.Length > indexOfEndOfBody + EndOfBody.Length)
+            {
+                additionalData = tail.Substring(indexOfEndOfBody + EndOfBody.Length).TrimStart();
+            }
+
+            string body = indexOfEndOfBody == lengthOfSubjectRepeatedInBody ? subject : tail.Substring(0, indexOfEndOfBody).TrimEnd();
+            return (body, additionalData);
```
gerhardol: Will this give another allocation of the string? If so GC will run more often during the startup, something that @drewnoakes looked into. I do not see any difference in the Linux repo when testing with a stopwatch.

GitCommands/RevisionReader.cs
```diff
@@ -451,17 +439,54 @@ DateTime ParseUnixDateTime()
...
+                return (body: null, additionalData: null);
+            }
+
+            string additionalData = null;
+            if (tail.Length > indexOfEndOfBody + EndOfBody.Length)
+            {
+                additionalData = tail.Substring(indexOfEndOfBody + EndOfBody.Length).TrimStart();
+            }
+
+            string body = indexOfEndOfBody == lengthOfSubjectRepeatedInBody ? subject : tail.Substring(0, indexOfEndOfBody).TrimEnd();
```
gerhardol: Suggest wrapping the line

UnitTests/GitCommandsTests/RevisionReaderTests.cs
```diff
@@ -85,5 +86,69 @@ public void BuildArguments_check_parameters(RefFilterOptions refFilterOptions, s
...
+        }
+
+        [TestCase("subject", "subject\n\n1DEA7CC4-FB39-450A-8DDF-762FCEA28B05", "filename")]
+        [TestCase("subject", "SUBJECT\n\n1DEA7CC4-FB39-450A-8DDF-762FCEA28B05", "filename")]
+        [TestCase("subject", "subject\n\n1DEA7CC4-FB39-450A-8DDF-762FCEA28B05")]
+        [TestCase("subject", "subject\0\01DEA7CC4-FB39-450A-8DDF-762FCEA28B05")]
+        [TestCase("subject", "SUBJECT\n\n1DEA7CC4-FB39-450A-8DDF-762FCEA28B05")]
+        [TestCase("subject", "SUBJECT\n\n____________________________________")]
+        [TestCase("subject", "_____________________________________________")]
+        [TestCase("subject", "sub\n1DEA7CC4-FB39-450A-8DDF-762FCEA28B05\nject")]
```
gerhardol: Is this test really correct? I would expect the last three to fail.

GitCommands/RevisionReader.cs
mstv: No, the intention of this code is to avoid another allocation of a string for single-line commit messages - as before the regression.

UnitTests/GitCommandsTests/RevisionReaderTests.cs
mstv: Yes, it is correct with the speed optimization (mentioned above) that only checks whether the length is as it would be for a single-line commit message.

GitCommands/RevisionReader.cs
```diff
@@ -451,17 +439,55 @@ DateTime ParseUnixDateTime()
...
+            string ReadLine([CanBeNull] StringPool pool = null, bool advance = true);
+
+            string ReadToEnd(bool advance = true);
+        }
+
         /// <summary>
         /// Simple type to walk along a string, line by line, without redundant allocations.
         /// </summary>
-        private struct StringLineReader
+        private struct StringLineReader : IStringLineReader
```
drewnoakes: Please do not make this `struct` an interface. Doing so causes it to be boxed, which allocates on the heap. If you need external access, just make the type public.

gerhardol:
> Haven't looked too deeply, however I still feel that it shouldn't be necessary to have a GUID as a separator and that the double-null idea discussed elsewhere would suffice. I may have missed some discussion where this was ruled out however.

There were no deep discussion about the separator:
https://github.com/gitextensions/gitextensions/pull/6132/files#r250354949
It was @fschmied that submitted the original PR (that I adapted), so it stayed.

fschmied:
The double null requires more effort, it's still in the works in #6132 by @vbjay, AFAIK. The placeholder I introduced was just intended as a temporary solution to make history work again until #6132 is done.

GitCommands/RevisionReader.cs
mstv: done

mstv:
I am going to merge this tomorrow if nobody chimes in.
I guess this should go into 3.2.1, too, shouldn't it?

gerhardol:
> I am going to merge this tomorrow if nobody chimes in.
I guess this should go into 3.2.1, too, shouldn't it?

This should not be a major issue from a user perspective but I am OK in including it in 3.2 (I have some candidates too).
A milestone is needed

RussKie:
I'd rather release a 3.3 than create a 3.2.1

PR #7159, opened 2019-09-17
Fix unit tests on `TryGetExactPathName()` (for AppVeyor VS2019 image)

pmiossec:
A little less punitive try to fix a failing test than #7112.
A per-requisite at the use of the VS2019 AppVeyor image.

The test that fails is partly due to a refactoring made in 1246b50 where I raise a little the bar with the expectations.

Indeed, before that, when the files existing, the unit test was verifying... nothing!

I made a change to verify that the paths cases was the ones of the real paths. I was thinking that the windows paths would have always the same cases for system files. It seems not to be the case :( as the AppVeyor VS2019 image has a different one...

I can propose is to find a in-between between testing nothing (like before the refactoring) and having the expected casing (after the refactoring).

I.e:

    We start with the paths without the good casing (lower case)
    We check that the result is still equivalent (by lowering the result)
    We check that the path has been changed (expecting that it is the good case)

PS: The test on @"C:\pagefile.sys" has been removed because I discovered that the call to the native api are not working with files at the root of the drive 👿 😭

## Test methodology 

- Unit test ;)

### DIFF
--- a/UnitTests/GitUITests/CommandsDialogs/FormFileHistoryControllerTests.cs
+++ b/UnitTests/GitUITests/CommandsDialogs/FormFileHistoryControllerTests.cs
@@ -30,18 +30,18 @@ namespace GitUITests.CommandsDialogs
         }
 
         [TestCase(@"c:\Users\Public\desktop.ini")]
-        [TestCase(@"c:\pagefile.sys")]
         [TestCase(@"c:\Windows\System32\cmd.exe")]
         [TestCase(@"c:\Users\Default\NTUSER.DAT")]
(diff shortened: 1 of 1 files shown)


gerhardol:
This still assumes that the user have installed in the standard paths, just relaxed for case.
The change fixes my problem, but is this really what we should test?

pmiossec:
For me, that is enough for the moment (to use 2019 Appveyor image)
We will re-evaluate when we will have more information...

RussKie:
I presume we no longer need this, right?

gerhardol:
> I presume we no longer need this, right?

The local file system tests were removed, so this can be closed

Issue #656, opened 2011-10-10
Error viewing Blame Tab on binary file


michaellwest:
Steps to reproduce:
1. From home screen, click "File tree" tab
2. Double click "cow-head.ico", "File History" window should appear
3. Click "Blame" tab
4. You should receive an error for object null reference when loading blame.

PR #7162, opened 2019-09-17
WIP: Convert to .NET Core 3.1

RussKie:
## Proposed changes

- Convert the project to .NET Standard 2.0 (libraries) and .NET Core 3.0 (UI)

## Test methodology 

- manual
- unit tests
- integration tests

----


.gitmodules
```diff
@@ -10,3 +10,10 @@
 [submodule "Externals/ICSharpCode.TextEditor"]
 	path = Externals/ICSharpCode.TextEditor
 	url = ../../gitextensions/ICSharpCode.TextEditor.git
+[submodule "Externals/AdysTech.CredentialManager"]
+	path = Externals/AdysTech.CredentialManager
+	url = ../../gitextensions/CredentialManager.git
+[submodule "Externals/PSTaskDialog"]
+	path = Externals/PSTaskDialog
+	url = https://github.com/gitextensions/PSTaskDialog.git
```
gerhardol: suggest a relative path

RussKie: Thank you

RussKie:
OMG this is actually working! 🚀 

The commit info rendering is broken, but everything else so far appears to be working.
image

GitExtUtils/GitUI/TaskExtensions.cs
```diff
@@ -1,200 +1,194 @@
-﻿using System;
-using System.Runtime.CompilerServices;
-using System.Threading;
-using System.Threading.Tasks;
-using Microsoft;
-
-namespace GitUI
+﻿namespace GitUI
 {
-    public static class TaskExtensions
```
RussKie: @sharwell can you please confirm this class is obsolete with Microsoft.VisualStudio.Threading v16.3.52?

RussKie:
I had to add few more submodules:
- AdysTech.CredentialManager - the referenced package is targeting .NET 4.5, I have asked the maintainers about an upgrade, and by the looks of it, it may happen https://github.com/AdysTech/CredentialManager/issues/31
- PSTaskDialog - the referenced package is referencing .NET 2.0, and the project looks abandoned for the most part. The good news that the functionality the package brings will be superseded once https://github.com/dotnet/winforms/pull/1133 is merged (expect in .NET 5.0)

I have commented out all functionality from Microsoft.WindowsAPICodePack. I found the repo, and forked it into our org, however I am unconvinced at this point that it is wise to bring another library and own its API. We can cherry-pick necessary API going forward.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -320,7 +320,9 @@ public void Clear()
             _backgroundScrollTo = 0;
 
             // Force the background thread to be killed, we need to be sure no background processes are running. Not the best practice, but safe.
-            _backgroundThread.Abort();
+            // TODO: not supported in .NET Core
+            //_backgroundThread.Abort();
+            _backgroundThread = null;
```
RussKie: @spdr870 could you please have a look.

[`Thread.Abort` is no longer supported](https://github.com/dotnet/coreclr/issues/20705), so we may need to re-write the handling.
This may be a possible solution https://stackoverflow.com/q/50957086/2338036

spdr870: This is an issue, because we need to make sure that no background thread is modifying the graph when we clear it. But I think I there is a simple solution. We need to move the (messy) code that creates RevisionGraph and starts the thread to a separate class. Also move the variable keeprunning to that class. Then in the gridview, in the clear function, set keeprunning to false in this new class and then create a new instance af that new class. Instead of killing the thread and resetting the state, it will let the thread stop in a safe way. The new instance will have it's own state, we do not have to worry about the state being modified by another thread. It's a better solution I think, because we do not need the threadabortexception.

RussKie:
Awesome, thank you!

grleachman:
@RussKie will pull down this branch, but it is quite old.
Do you have any issues open for the problem encountered in converting to .net-core perhaps labelled as net-core?

RussKie:
> will pull down this branch, but it is quite old.
> Do you have any issues open for the problem encountered in converting to .net-core perhaps labelled as net-core?

Awesome! We don't have any .NET Core specific issues yet. This was a pure POC before .NET Core 3.0 was released.

Since then we have converted to SDK-style projects, moved to use [Arcade SDK](https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md, though I have customised and tweaked it for our purposes). Now it is possible to build the solution from cli (and VS Code). With that we cenralised reference management.
This work should simplify porting to .NET Core 3.1.

Also AdysTech.CredentialManager is not .NET Core compatible (https://github.com/AdysTech/CredentialManager/issues/31).

So if you want to get your hand dirty, perhaps start afresh and repeat the steps I've done:
image

RussKie:
> will pull down this branch, but it is quite old.
> Do you have any issues open for the problem encountered in converting to .net-core perhaps labelled as net-core?

Awesome! We don't have any .NET Core specific issues yet. This was a pure POC before .NET Core 3.0 was released.

Since then we have converted to SDK-style projects, moved to use [Arcade SDK](https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md, though I have customised and tweaked it for our purposes). Now it is possible to build the solution from cli (and VS Code). With that we cenralised reference management.
This work should simplify porting to .NET Core 3.1.

Also AdysTech.CredentialManager is not .NET Core compatible (https://github.com/AdysTech/CredentialManager/issues/31).

So if you want to get your hand dirty, perhaps start afresh and repeat the steps I've done:
image

Modulo don't create unnecessary submodules (e.g. consume AdysTech.CredentialManager as a NuGet package). PSTaskDialog will be used until we move to .NET 5, then we'll replace it with the Windows Forms one.

pmiossec:
>Also AdysTech.CredentialManager is now .NET Core compatible 

It seems that we have to switch to:
https://www.nuget.org/packages/AdysTech.CredentialManager.Core/

drewnoakes:
I just filed a PR against them to consolidate their stuff in a single package that supports both .NET Framework and .NET Standard:

https://github.com/AdysTech/CredentialManager/pull/41

RussKie:
Rebased the branch, totally untested, not even sure whether it compiles.

pmiossec:
Perhaps better to rebase on #8156 once it is merged to be able to remove work that has been done around AdysTech.CredentialManager as submodule...

RussKie:
Aye, I recalled there was a package available but it was late and I was
exhausted to search for it.
Thank you for the heads up 👍

RussKie:
The app can now be run 🎉

RussKie:
@NikolayXHD I had to pull the theming out because EasyHook's implementation depends on remoting that has been deprecated and removed in .NET Core. 
There's an open issue for EasyHook to migrate (https://github.com/EasyHook/EasyHook/issues/278), but it may be some time before that happens...

GitCommands/EnvironmentPathsProvider.cs
```diff
@@ -72,22 +72,22 @@ public IEnumerable<string> GetEnvironmentValidPaths()
         // TODO: optimise?
         internal static bool IsValidPath(string path)
         {
-            FileInfo fi = null;
             try
             {
-                fi = new FileInfo(path);
+                var _ = new FileInfo(path).Attributes;
```
drewnoakes, line 77:
```suggestion
                _ = new FileInfo(path).Attributes;
```

GitCommands/StringPool.cs
```diff
@@ -178,44 +178,46 @@ internal static unsafe bool EqualsAtIndex(string source, int index, string compa
             }
 
             fixed (char* pc = comparand)
-            fixed (char* ps = source)
```
drewnoakes, line 181: Is this just an indentation change? Perhaps revert, or add braces.

GitExtensions/GitExtensions.csproj
```diff
@@ -17,37 +23,29 @@
...
+    <ProjectReference Include="..\GitUI\GitUI.csproj" />
     <ProjectReference Include="..\Plugins\GitUIPluginInterfaces\GitUIPluginInterfaces.csproj" />
-    <ProjectReference Include="..\GitCommands\GitCommands.csproj" />
     <ProjectReference Include="..\ResourceManager\ResourceManager.csproj" />
   </ItemGroup>
 
   <ItemGroup>
     <None Include="app.config" />
     <None Include="app.manifest" />
+    <None Remove="AutoCompleteRegexes.txt" />
```
drewnoakes, line 47: This line can be deleted.

GitUI/Interops/UxTheme/SetWindowTheme.cs
```diff
@@ -10,9 +10,11 @@ internal static partial class NativeMethods
         public static unsafe int SetWindowTheme(IntPtr hWnd, string subAppName, string subIdList)
         {
             fixed (char* pszSubAppName = subAppName)
-            fixed (char* pszSubIdList = subIdList)
             {
-                return SetWindowTheme(hWnd, pszSubAppName, pszSubIdList);
+                fixed (char* pszSubIdList = subIdList)
```
drewnoakes, line 14: Ditto: indentation

Plugins/Statistics/GitStatistics/LineCounter.cs
```diff
@@ -20,7 +20,7 @@ public sealed class LineCounter
 
         public void FindAndAnalyzeCodeFiles(string filePattern, string directoriesToIgnore, IEnumerable<string> filesToCheck)
         {
-            var extensions = filePattern.Replace("*", "").Split(';').ToHashSet(StringComparer.InvariantCultureIgnoreCase);
+            var extensions = LinqExtensions.ToHashSet(filePattern.Replace("*", "").Split(';'), StringComparer.InvariantCultureIgnoreCase);
```
drewnoakes, line 23: Why this change?

ResourceManager/ResourceManager.csproj
```diff
@@ -18,10 +18,4 @@
     <ProjectReference Include="..\Plugins\GitUIPluginInterfaces\GitUIPluginInterfaces.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <!-- Resources -->
-    <EmbeddedResource Update="Properties\Resources.resx" Generator="ResXFileCodeGenerator" LastGenOutput="Resources.Designer.cs" />
-    <Compile Update="Properties\Resources.Designer.cs" AutoGen="True" DependentUpon="Resources.resx" DesignTime="True" />
-  </ItemGroup>
```
drewnoakes, line 25: Is this intentional?

drewnoakes:
What's `GitUI/fullproject.xml` and should it be added? It's quite a large XML file.

RussKie:
> What's `GitUI/fullproject.xml` and should it be added? It's quite a large XML file.

Not needed, added by an oversight

Plugins/Statistics/GitStatistics/LineCounter.cs
RussKie: A name conflict, .NET Runtime added a new method `ToHashSet` with the same signature as ours. However it wasn't a straight forward replace/deletion in our codebase, so I just left it as is for now.

GitCommands/EnvironmentPathsProvider.cs
RussKie: The change here will probably cause us grief going forward, the IO API have been changed in .NET Core 2.1...

ResourceManager/ResourceManager.csproj
RussKie: No, likely a bad rebase. Thank you

GitCommands/StringPool.cs
RussKie: Braces are there, I'm guessing you've missed them.
Same for the other one.

RussKie:
Everything seems to be working. 

Though found 2 significant issues (apart from https://github.com/gitextensions/gitextensions/pull/7162/commits/75f04b2fae03628405d400883112d693a2f40183 and https://github.com/gitextensions/gitextensions/pull/7162/commits/f6cee0339bfa419e16347917f5c698f1a8bed378), and so far unable to resolve either:

- links aren't getting converted to such in richtextboxes, e.g.
    image
- incorrect code rendering:
    image
    I suspect this may be related to https://github.com/dotnet/winforms/pull/656, ¯\\_(ツ)\_/¯


I have also tried building against .NET 5.0 RC1 - literally had to do one change - to resolve ambiguous namespaces for the `TaskDialog` control 🎉

UnitTests/GitUI.Tests/ControlThreadingExtensionsTests.cs
```diff
@@ -171,6 +171,7 @@ public async Task TokenCancelledBeforeSwitchOnBackgroundThread()
         }
 
         [Test]
+        [Ignore("Hangs")]
         public async Task ControlDisposedAfterSwitchOnBackgroundThread()
```
RussKie, line 175: @sharwell this test hangs run on .NET Core 3.1 in: `await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();`.
Do you think you can have a look?

UnitTests/GitUI.Tests/ThreadHelperTests.cs
```diff
@@ -27,6 +27,7 @@ private static async Task ThrowExceptionAsync(Exception ex)
         }
 
         [Test]
+        [Ignore("Hangs")]
         public async Task FileAndForgetReportsThreadException()
         {
             using (var helper = new ThreadExceptionHelper())
```
RussKie, line 33: @sharwell this and below tests hang in `await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);`.

RussKie:
I tested net5.0-rc1, and neither blocking issues are being addressed at the Windows Forms level. 

> Though found 2 significant issues (apart from [75f04b2](https://github.com/gitextensions/gitextensions/commit/75f04b2fae03628405d400883112d693a2f40183) and [f6cee03](https://github.com/gitextensions/gitextensions/commit/f6cee0339bfa419e16347917f5c698f1a8bed378)), and so far unable to resolve either:
> 
> * links aren't getting converted to such in richtextboxes, e.g.
>   image

The RTB link issue is caused by switch from RichEdit 2.0 to RichEdit 4.1 in net471, and the newer version does not respond to the messages the older version responded. This is significant issue, and the only way I see to resolve it is to deprecate `RichTextBoxXhtmlSupportExtension` (since it is only in 2 places - `CommitInfoHeader` and `CommitInfo` to facilitate links) and provide a different implementation that injects hyperlinks in RTF.
E.g. the hyperlink [MSW](http://msw) is represented by `\field{\*\fldinst{HYPERLINK "http://msw" }}{\fldrslt{\ul\cf1\cf1\ul MSW}}`.


> * incorrect code rendering:
>   image
>   I suspect this may be related to [dotnet/winforms#656](https://github.com/dotnet/winforms/pull/656), ¯\_(ツ)_/¯

I've checked, and it isn't related to #656. But it is a clear regression from .NET Core 3.1 to .NET 5.0:
The spacing goes wild in .NET 5.0:
image
...whilst in .NET Core 3.1 it works as expected:
image

NB: at this stage it is unclear whether it is a regression in Windows Forms or .NET Runtime (that owns some of graphics libraries).

gerhardol:
@RussKie Is this replaced by #8522 ?

RussKie:
> Is this replaced by #8522 ?

Yep.

PR #7173, opened 2019-09-18
Align location of System.Collections.Immutable.dll between MSI and Portable builds

### LINKED ISSUE #7155: System.Collections.Immutable.dll missing from GitExtensions-Portable-3.2.1.6628.zip
In the portable distribution the dll mentioned above is missing (it is contained in the msi installer).

## Current behaviour
Crashes with an error message that the dll System.Collections.Immutable.dll is missing.

## Expected behaviour
No crash during startup

## Steps to reproduce
Unpack the portable zip and start it

Yes

## Environment
- GitExtensions version: 3.2.1.6628
- GIT version: irrelevant
- OS version: irrelevant
- .NET version: irrelevant

mast-eu:
Fixes #7155

## Proposed changes

- move `System.Collections.Immutable.dll` from /plugins dir to root dir for portable builds, since this is the location of the DLL in the installed version

## Test methodology 

- made portable build and verified correct location of DLL
- manual tests 

----

### DIFF
--- a/Setup/MakePortableArchive.cmd
+++ b/Setup/MakePortableArchive.cmd
@@ -25,6 +25,8 @@ xcopy /y /i ..\GitExtensions\bin\%Configuration%\System.Collections.Concurrent.d
 IF ERRORLEVEL 1 EXIT /B 1 
 xcopy /y /i ..\GitExtensions\bin\%Configuration%\System.Collections.dll GitExtensions\
 IF ERRORLEVEL 1 EXIT /B 1 
+xcopy /y /i ..\GitExtensions\bin\%Configuration%\System.Collections.Immutable.dll GitExtensions\
+IF ERRORLEVEL 1 EXIT /B 1 
 xcopy /y /i ..\GitExtensions\bin\%Configuration%\System.Collections.NonGeneric.dll GitExtensions\
 IF ERRORLEVEL 1 EXIT /B 1 
 xcopy /y /i ..\GitExtensions\bin\%Configuration%\System.Collections.Specialized.dll GitExtensions\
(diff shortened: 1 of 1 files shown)


gerhardol:
Should be added to 3.2.x too

mast-eu:
It's true that also I cannot reproduce the reported issue. I just noticed a difference between installed and portable which appears to be "by accident" rather than for a specific reason.
Thinking about it, I shouldn't write "Fixes #7155", because I do not know whether this PR actually _fixes_ something. It for sure _alignes_ something, though.

Slightly OT: running a full comparison, there are further differences between portable and installed versions.  Below is a screenshot showing the files that are different between `C:\Program Files (x86)\GitExtensions\` (left side) and [GitExtensions-Portable-3.2.1.6628.zip](https://github.com/gitextensions/gitextensions/releases/download/v3.2.1/GitExtensions-Portable-3.2.1.6628.zip) (right side).

image

Note: The screenshot shows only the differences. Files that are common in both are hidden.

RussKie:
@mast-eu thank you for the screenshot. Are these two clean installations?

The portable bundle mustn't have GitExtensionsShellEx32/64.dll as these are Explorer extensions and must be registered.

A question would be to ask why we need S.C.Immutable.dll in the root folder shipped in the msi.
Another - why we are not bundling TFS 2013 plugin in the msi - or - why we are shipping it in the portable version. I want to move these plugins out of the codebase ASAPP.

mast-eu:
> Are these two clean installations?

Yes, clean installation of 3.2.1 on Win10 with all default installer options. And the portable was just downloaded. Both have been taken directly from the official releases on GH.

One more difference that not easily seen: the portable contains 2 identical copies of GitUIPluginInterfaces.dll, one in root and in /plugins.

mast-eu:
Closing my PR because the original issue it tried to resolve turned out to be not reproducible. 

The other differences found between installed and portable https://github.com/gitextensions/gitextensions/pull/7173#issuecomment-533986737 are still strange. However I think it's not worth the trouble to sort them out right now, because most of them, if not all, will be auto-resolved by moving the plugins out of the main repo.

PR #7184, opened 2019-09-21
GitStatusMonitor: Improve logic for timer 25 days wrap

### LINKED ISSUE #7183: GitStatusMonitor can miss filesystem update when git-status runs
## Current behaviour

* It is possible that two git-status commands are starting when opening a repo
* It is possible that a git-status command is started while the first is running also if the second is not requested
* If a file system update occurs while git-status is running, the request could be ignored and the status update be done first after 5 minutes.

The problem should normally have a low probability and occur if the changes occurs on either side of the timer ticks in GitStatusMonitor and due to insufficient lock handling. The problems normally occurs only when the git-status running time is quite long.

I have seen the problems sporadically after an IT change causing the Symantec virus program (yes it is a virus, not antivirus as it causes more problems than it solves) increasing the git-status time.
The second problem I have only seen at a colleague (not been able to reproduce but confirmed the PR fixed the problem). The workaround there was to open FormCommit once. GitStatusMonitor runs git-status in the "background" with --no-optional-locks to not interfere with interactive commands. The drawback is that the caches are never updated... Subsequent GitStatusMonitor invocations are faster.

The core problem is that the implementation to handle the timer wrapping (after 25 days) causes updates to be missed. The locking could also be a part of the problem

## Expected behaviour

Handle the situations properly

## Steps to reproduce

This can be hard to reproduce, adding a delay to simulate a slow git-status can simplify:

```diff
$ git diff
diff --git a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
index 5c4e63ced..096c7cf2f 100644
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -364,6 +364,7 @@ private void Update()
                                 noLocks: true);
                             var output = await module.GitExecutable.GetOutputAsync(cmd).ConfigureAwait(false);
                             var changedFiles = GitCommandHelpers.GetStatusChangedFilesFromString(module, output);
+                            await new Executable("cmd.exe", module.WorkingDir).GetOutputAsync("/c timeout 10").ConfigureAwait(false);

                             await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
$
```

Only the third issue can be provoked:
* Start git-satus by opening a repo or refreshing
* Modify a file while git-status runs
* Verify that a new git-status starts earlier than next 5 minutes (with the pach above it will be about 3*10s)

No

## Environment
- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.23.0.windows.1
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #7183 

## Proposed changes

 * Always compare timer diff (to handle timer wraparound after 25 days)
 to avoid missing updates while command is running
 * Increase min update time 3s -> 5s at changes in the file system
 * Use locks consistently
 * Start interactive update only once to avoid race where two commands started

## Test methodology

See issue for how to provoke the problem

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -1,11 +1,9 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Threading.Tasks;
+using System.Threading;
 using System.Windows.Forms;
 using GitCommands;
 using GitUIPluginInterfaces;
-using Microsoft.VisualStudio.Threading;
 
 namespace GitUI.CommandsDialogs.BrowseDialog
 {
@@ -26,7 +24,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
         /// <summary>
         /// Minimum interval between subsequent updates
         /// </summary>
-        private const int MinUpdateInterval = 3000;
+        private const int MinUpdateInterval = 5000;
 
         /// <summary>
         /// Update every 5min, just to make sure something didn't slip through the cracks.
@@ -35,16 +33,19 @@ namespace GitUI.CommandsDialogs.BrowseDialog
 
         private readonly FileSystemWatcher _workTreeWatcher = new FileSystemWatcher();
         private readonly FileSystemWatcher _gitDirWatcher = new FileSystemWatcher();
-        private readonly Timer _timerRefresh;
+        private readonly System.Windows.Forms.Timer _timerRefresh;
         private bool _commandIsRunning;
         private string _gitPath;
         private string _submodulesPath;
         private readonly CancellationTokenSequence _statusSequence = new CancellationTokenSequence();
 
         // Timestamps to schedule status updates, limit the update interval dynamically
+        // Note that TickCount wraps after 25 days uptime, always compare diff
         private int _nextUpdateTime;
         private int _nextEarliestTime;
-        private bool _nextIsInteractive;
+
+        // Track changes detected (used while command is running)
+        private bool _pendingUpdate;
         private GitStatusMonitorState _currentStatus;
 
         public bool Active
@@ -65,7 +66,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
 
         public GitStatusMonitor(IGitUICommandsSource commandsSource)
         {
-            _timerRefresh = new Timer
+            _timerRefresh = new System.Windows.Forms.Timer
             {
                 Enabled = true,
                 Interval = InteractiveUpdateDelay / 2
@@ -125,7 +126,8 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                     return;
                 }
 
-                ScheduleNextInteractiveTime();
+                _gitDirWatcher.EnableRaisingEvents = false;
+                ScheduleNextUpdateTime(FileChangedUpdateDelay);
             }
 
             void WorkTreeChanged(object sender, FileSystemEventArgs e)
@@ -148,8 +150,8 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                     return;
                 }
 
-                ScheduleNextUpdateTime(FileChangedUpdateDelay);
                 _workTreeWatcher.EnableRaisingEvents = false;
+                ScheduleNextUpdateTime(FileChangedUpdateDelay);
             }
         }
 
@@ -169,6 +171,12 @@ namespace GitUI.CommandsDialogs.BrowseDialog
             _gitDirWatcher.Dispose();
             _timerRefresh.Dispose();
             _statusSequence.Dispose();
+        }
+
+        private bool GitDirWatcherEnableRaisingEvents()
+        {
+            return Directory.Exists(_gitDirWatcher.Path)
+                    && !_gitDirWatcher.Path.StartsWith(_workTreeWatcher.Path);
         }
 
         private GitStatusMonitorState CurrentStatus
@@ -199,18 +207,21 @@ namespace GitUI.CommandsDialogs.BrowseDialog
 
                     case GitStatusMonitorState.Running:
                         {
+                            _workTreeWatcher.EnableRaisingEvents = Directory.Exists(_workTreeWatcher.Path);
+                            _gitDirWatcher.EnableRaisingEvents = GitDirWatcherEnableRaisingEvents();
+
+                            lock (_statusSequence)
+                            {
+                                // An interactive update will be requested separately
+                                _nextUpdateTime
+                                    = _nextEarliestTime
+                                    = Environment.TickCount + FileChangedUpdateDelay;
+                                _pendingUpdate = true;
+                                _commandIsRunning = false;
+                                _statusSequence.CancelCurrent();
+                            }
+
                             _timerRefresh.Start();
-                            if (Directory.Exists(_workTreeWatcher.Path))
-                            {
-                                _workTreeWatcher.EnableRaisingEvents = true;
-                            }
-
-                            if (Directory.Exists(_gitDirWatcher.Path))
-                            {
-                                _gitDirWatcher.EnableRaisingEvents = !_gitDirWatcher.Path.StartsWith(_workTreeWatcher.Path);
-                            }
-
-                            ScheduleNextInteractiveTime();
                         }
 
                         break;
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -347,6 +348,8 @@ private void Update()
 
             var commandStartTime = Environment.TickCount;
             _workTreeWatcher.EnableRaisingEvents = true;
+            _gitDirWatcher.EnableRaisingEvents = Directory.Exists(_gitDirWatcher.Path) &&
+                    !_gitDirWatcher.Path.StartsWith(_workTreeWatcher.Path);
```
mstv: Nitpick: This code is an identical copy.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -386,16 +407,6 @@ private void Update()
...
-                        finally
-                        {
-                            lock (_statusSequence)
-                            {
-                                if (!ModuleHasChanged() && !cancelToken.IsCancellationRequested)
-                                {
-                                    _commandIsRunning = false;
-                                }
-                            }
-                        }
```
mstv: Why can this be omitted?
I guess because `CurrentStatus = GitStatusMonitorState.Stopped;` will be set and `_commandIsRunning` is reset when entering the state `GitStatusMonitorState.Running`.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -307,37 +313,32 @@ private void Update()
...
+                _pendingUpdate = false;
 
-            // Schedule update every 5 min, even if we don't know that anything changed
-            ScheduleNextUpdateTime(PeriodicUpdateInterval);
+                // Schedule update every 5 min, even if we don't know that anything changed
+                _nextUpdateTime = Environment.TickCount + PeriodicUpdateInterval;
+            }
 
             if (!Directory.Exists(_workTreeWatcher.Path))
             {
```
mstv: `_commandIsRunning = false` should be set (under `lock (_statusSequence)`) before exiting this function here, shouldn't it?
This should also be done if the `RunAsync` fails or is not reached due to an exception.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -199,18 +200,23 @@ private GitStatusMonitorState CurrentStatus
...
+                            _gitDirWatcher.EnableRaisingEvents = Directory.Exists(_gitDirWatcher.Path) &&
+                                    !_gitDirWatcher.Path.StartsWith(_workTreeWatcher.Path);
+
+                            lock (_statusSequence)
                             {
-                                _gitDirWatcher.EnableRaisingEvents = !_gitDirWatcher.Path.StartsWith(_workTreeWatcher.Path);
+                                // An interactive update will be requested separately
+                                _nextUpdateTime =
+                                    _nextEarliestTime =
+                                    Environment.TickCount + FileChangedUpdateDelay;
```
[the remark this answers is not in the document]
RussKie: @mstv I like your version even less than @gerhard's.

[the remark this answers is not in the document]
mstv: I agree regarding multiple assignments. Though it will need very good arguments to convince me that operators shouldn't be put in front of the continuation line.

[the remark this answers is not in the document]
gerhardol: I can duplicate the assignments too, but changing to having the operator first right now.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -36,15 +33,18 @@ public sealed class GitStatusMonitor : IDisposable
         private readonly FileSystemWatcher _workTreeWatcher = new FileSystemWatcher();
         private readonly FileSystemWatcher _gitDirWatcher = new FileSystemWatcher();
         private readonly Timer _timerRefresh;
-        private bool _commandIsRunning;
```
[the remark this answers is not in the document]
gerhardol: The members were not in order before, now they are grouped by usage. 
Is this a request to resort the existing or just move back _commandIsRunning?

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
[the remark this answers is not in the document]
gerhardol: I will change this according to mstv's proposal rather than duplicating the assignment

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -307,37 +313,32 @@ private void Update()
                 return;
             }
 
-            if (Environment.TickCount < _nextUpdateTime)
+            if (_nextUpdateTime - Environment.TickCount > 0)
             {
                 return;
             }
```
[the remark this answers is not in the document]
gerhardol: "25 days" is handled with a few glitches:
 * Periodic update (every 5th min if on demand is "lost") will fail.
 * The on-change update could fail with  very low probability (a manual refresh will be handled)
The problem is in ScheduleNextUpdateTime() where the 25 day handling (checking < 0) could cause on-change requests to be skipped.
This is quite hard to realize with the current structure.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: If there is a problem, the handling is disabled, now and earlier.
You will have to restart by open a new repo.
Previously the handling was deactivated but status variables were reset.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: A separate method will be (even if that hides the handling)

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: See above

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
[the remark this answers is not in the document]
mstv: I tended to grouping by usage or logic relation, too. Though it's too often not straighforward, these groups can overlap, ... That's why the simple and clear rule.
I don't think we should force every contributor to sort existing files.
When I add something, I try to insert it in alphabetical order and do not touch others.
@gitextensions/git-extensions-source, what do you think?

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
mstv: I've seen too many race conditions.
This one might be synthetic and is perhaps resolved at another place (where?) - at least by reloading the repo. (Though I consider the latter to be a only workaround.)

```C#
            lock (_statusSequence)
            {
                if (_commandIsRunning)
                {
                    return;
                }

                _commandIsRunning = true;
                _pendingUpdate = false;

                // Schedule update every 5 min, even if we don't know that anything changed
                _nextUpdateTime = Environment.TickCount + PeriodicUpdateInterval;
            }

            if (!Directory.Exists(_workTreeWatcher.Path))
            {
                // The directory no longer exists, watcher cannot be enabled
                return;
            }

            // snip

            ThreadHelper.JoinableTaskFactory.RunAsync(
                    async () =>
                    {
                         // snip

                         _commandIsRunning = false;
                    }).FileAndForget();
```
If the directory is unavailable temporarily, `_commandIsRunning` will never be reset to `false` unless the state `GitStatusMonitorState.Running` is re-entered, because this function will return before it will start the command. In the next calls of this function, it will exit early regardless whether the directory is available again.
Exceptions from e.g. `Directory.Exists()`, `_statusSequence.Next()` or `JoinableTaskFactory.RunAsync` could have the same effect.

gerhardol: !Directory.Exists(_workTreeWatcher.Path) is related to 
_workTreeWatcher.EnableRaisingEvents = true;
This should be moved into lock() to handle this situation which is relevant if you have the repo on a network drive or USB stick

The exceptions need to open a new repo.
I would like to handle that in a forced refresh too, but this is currently a part of a normal flow, so it will require a rewrite, like:
https://github.com/gerhardol/gitextensions/blob/bugfix/i7183-gitstatus-alternative/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs#L376
I prefer the current

mstv: Both variants do not address the case where the async task is not started at all because of an exception before `RunAsync`.

gerhardol: > Both variants do not address the case where the async task is not started at all because of an exception before RunAsync.

That would require handling _commandIsRunning just before Async
Do not like that either, pushed an example to the alt branch

To be clear: This is a very relevant discussion, even if I not want to change.

mstv: The alt branch looks much safer. But it must be checked that `CancellationTokenSequence.Next()` cannot cause a deadlock if called under `lock`.
> even if I not want to change

What are your concerns?

gerhardol: > What are your concerns?

My concern is that this adds complexity for something that will not improve the functionality anyway.
I could ask for another opinion, but that will likely be "do the complete implementation", so I will clean the alt branch up.

gerhardol: Pushed changes
Will squash when approved

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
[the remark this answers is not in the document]
gerhardol: Changed back the moved member

gerhardol:
The problem here is that this is about responsiveness and load. There are async, multi-thread and overflow complications. I cannot see any relevant test with a reasonable effort.

Issue #7181, opened 2019-09-21
Unexpected sort order of branches in which the current commit is contained in


mstv:
## Current behaviour

Branches are sorted alphabetically in reverse order.

## Expected behaviour

From the following Git command I'd expect the branches sorted descending by date committed:
https://github.com/gitextensions/gitextensions/blob/276431e3b125c237828ee013f939c31f417938a4/GitUI/CommitInfo/CommitInfo.cs#L216-L224

## Steps to reproduce

- select an older commit
- check "Contained in branches: ..." in Commit Info 

## Screenshots

![grafik](https://user-images.githubusercontent.com/36601201/65365805-e8890680-dc1c-11e9-881c-2fe47463fbea.png)

## Did this work in previous version of GitExtensions

don't know

## Environment

- Git Extensions 3.3.0.6647
- Build 32210cc84388272506c5a78240b43ea66db85a1b (Dirty)
- Git 2.21.0.windows.1 (recommended: 2.23.0 or later)
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

mstv:
`git for-each-ref --sort=-committerdate --sort=-taggerdate --format="%(refname)" refs/`
returns the tags in the expected order, though the branches in reverse alphabetical order.

E.g. `git for-each-ref --sort=-committerdate --format="%(refname)" refs/remotes refs/heads` works as expected.

Closing this issue because since #7186, this git command is not used anymore.

PR #7186, opened 2019-09-21
Limit removable for "Contained in branches/tags"

### LINKED ISSUE #5966: Overflow of superproject references

mstv:
Fixes #5966 (part 2 of 2).

## Proposed changes

- [x] Show up to 12 lines for "Contained in branches/tags", one branch/tag per line
- [x] Add link to show all "Contained in branches/tags"
- [x] Show relevant branches (current & masters) first
- [x] As long as GetSortedRefs() doesn't return the branches sorted by date (#7181), sort other branches alphabetically.
- [x] Add `TranslationString`s from helper classes to translation
- [x] Turn GetSortedRefs into GetSortedTags (no more need to load sorted branches)

### Before

grafik

### After

grafikgrafik
grafik

## Test methodology 

- added NUnit tests

## Test environment(s)- Git Extensions 3.3.0
- Build 7ff0c27138154120d4e886b8169119250af7fd5b
- Git 2.23.0.windows.1
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -47,10 +47,6 @@ namespace GitUI.CommitInfo
 
         private static readonly TranslationString _brokenRefs = new TranslationString("The repository refs seem to be broken:");
         private static readonly TranslationString _copyLink = new TranslationString("Copy &link ({0})");
-        private static readonly TranslationString _containedInBranches = new TranslationString("Contained in branches:");
-        private static readonly TranslationString _containedInNoBranch = new TranslationString("Contained in no branch");
-        private static readonly TranslationString _containedInTags = new TranslationString("Contained in tags:");
-        private static readonly TranslationString _containedInNoTag = new TranslationString("Contained in no tag");
         private static readonly TranslationString _trsLinksRelatedToRevision = new TranslationString("Related links:");
         private static readonly TranslationString _derivesFromTag = new TranslationString("Derives from tag:");
         private static readonly TranslationString _derivesFromNoTag = new TranslationString("Derives from no tag");
@@ -68,6 +64,7 @@ namespace GitUI.CommitInfo
         private readonly IConfigFileRemoteSettingsManager _remotesManager;
         private readonly GitDescribeProvider _gitDescribeProvider;
         private readonly CancellationTokenSequence _asyncLoadCancellation = new CancellationTokenSequence();
+        private readonly RefsFormatter _refsFormatter;
 
         private readonly IDisposable _revisionInfoResizedSubscription;
         private readonly IDisposable _commitMessageResizedSubscription;
@@ -114,6 +111,7 @@ namespace GitUI.CommitInfo
             _externalLinkRevisionParser = new ExternalLinkRevisionParser(_remotesManager);
             _gitRevisionExternalLinksParser = new GitRevisionExternalLinksParser(_effectiveLinkDefinitionsProvider, _externalLinkRevisionParser);
             _gitDescribeProvider = new GitDescribeProvider(() => Module);
+            _refsFormatter = new RefsFormatter(_linkFactory);
 
             var color = SystemColors.Window.MakeColorDarker(0.04);
             pnlCommitMessage.BackColor = color;
@@ -546,7 +544,7 @@ namespace GitUI.CommitInfo
                         _tags.RemoveRange(MaximumDisplayedRefs, _tags.Count - MaximumDisplayedRefs);
                     }
 
-                    _tagInfo = GetTagsWhichContainsThisCommit(_tags, ShowBranchesAsLinks);
+                    _tagInfo = _refsFormatter.FormatTags(_tags, ShowBranchesAsLinks);
                 }
 
                 if (_branches != null && string.IsNullOrEmpty(_branchInfo))
@@ -559,7 +557,7 @@ namespace GitUI.CommitInfo
                         _branches.RemoveRange(MaximumDisplayedRefs, _branches.Count - MaximumDisplayedRefs);
                     }
 
-                    _branchInfo = GetBranchesWhichContainsThisCommit(_branches, ShowBranchesAsLinks);
+                    _branchInfo = _refsFormatter.FormatBranches(_branches, ShowBranchesAsLinks);
                 }
             }
 
@@ -585,78 +583,6 @@ namespace GitUI.CommitInfo
                 }
 
                 return result.ToString();
-            }
-
-            string GetBranchesWhichContainsThisCommit(IEnumerable<string> branches, bool showBranchesAsLinks)
-            {
-                const string remotesPrefix = "remotes/";
-
-                // Include local branches if explicitly requested or when needed to decide whether to show remotes
-                bool getLocal = AppSettings.CommitInfoShowContainedInBranchesLocal ||
-                                AppSettings.CommitInfoShowContainedInBranchesRemoteIfNoLocal;
-
-                // Include remote branches if requested
-                bool getRemote = AppSettings.CommitInfoShowContainedInBranchesRemote ||
-                                 AppSettings.CommitInfoShowContainedInBranchesRemoteIfNoLocal;
-                var links = new List<string>();
-                bool allowLocal = AppSettings.CommitInfoShowContainedInBranchesLocal;
-                bool allowRemote = getRemote;
-
-                foreach (var branch in branches)
-                {
-                    string noPrefixBranch = branch;
-                    bool branchIsLocal;
-                    if (getLocal && getRemote)
-                    {
-                        // "git branch -a" prefixes remote branches with "remotes/"
-                        // It is possible to create a local branch named "remotes/origin/something"
-                        // so this check is not 100% reliable.
-                        // This shouldn't be a big problem if we're only displaying information.
-                        branchIsLocal = !branch.StartsWith(remotesPrefix);
-                        if (!branchIsLocal)
-                        {
-                            noPrefixBranch = branch.Substring(remotesPrefix.Length);
-                        }
-                    }
-                    else
-                    {
-                        branchIsLocal = !getRemote;
-                    }
-
-                    if ((branchIsLocal && allowLocal) || (!branchIsLocal && allowRemote))
-                    {
-                        var branchText = showBranchesAsLinks
-                            ? _linkFactory.CreateBranchLink(noPrefixBranch)
-                            : WebUtility.HtmlEncode(noPrefixBranch);
-
-                        links.Add(branchText);
-                    }
-
-                    if (branchIsLocal && AppSettings.CommitInfoShowContainedInBranchesRemoteIfNoLocal)
-                    {
-                        allowRemote = false;
-                    }
-                }
-
-                if (links.Any())
-                {
-                    return WebUtility.HtmlEncode(_containedInBranches.Text) + " " + links.Join(", ");
-                }
-
-                return WebUtility.HtmlEncode(_containedInNoBranch.Text);
-            }
-
-            string GetTagsWhichContainsThisCommit(IEnumerable<string> tags, bool showBranchesAsLinks)
-            {
-                var tagString = tags
-                    .Select(s => showBranchesAsLinks ? _linkFactory.CreateTagLink(s) : WebUtility.HtmlEncode(s)).Join(", ");
-
-                if (!string.IsNullOrEmpty(tagString))
-                {
-                    return WebUtility.HtmlEncode(_containedInTags.Text) + " " + tagString;
-                }
-
-                return WebUtility.HtmlEncode(_containedInNoTag.Text);
             }
         }
 
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -79,7 +79,7 @@ namespace GitUI.CommitInfo
         private List<string> _branches;
         private string _branchInfo;
         private string _gitDescribeInfo;
-        [CanBeNull] private IDictionary<string, int> _refsOrderDict;
+        [CanBeNull] private IDictionary<string, int> _tagsOrderDict;
         private int _revisionInfoHeight;
         private int _commitMessageHeight;
         private bool _showAllBranches;
@@ -97,10 +97,10 @@ namespace GitUI.CommitInfo
             {
                 this.InvokeAsync(() =>
                 {
-                    UICommandsSource.UICommandsChanged += delegate { RefreshSortedRefs(); };
+                    UICommandsSource.UICommandsChanged += delegate { RefreshSortedTags(); };
 
                     // call this event handler also now (necessary for "Contained in branches/tags")
-                    RefreshSortedRefs();
+                    RefreshSortedTags();
                 }).FileAndForget();
             };
 
@@ -139,7 +139,7 @@ namespace GitUI.CommitInfo
             commitInfoHeader.SetContextMenuStrip(commitInfoContextMenuStrip);
         }
 
-        private void RefreshSortedRefs()
+        private void RefreshSortedTags()
         {
             if (!Module.IsValidGitWorkingDir())
             {
@@ -148,7 +148,7 @@ namespace GitUI.CommitInfo
 
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
-                await LoadSortedRefsAsync();
+                await LoadSortedTagsAsync();
             }).FileAndForget();
         }
 
@@ -224,14 +224,13 @@ namespace GitUI.CommitInfo
             UpdateRevisionInfo();
         }
 
-        private IDictionary<string, int> GetSortedRefs()
+        private IDictionary<string, int> GetSortedTags()
         {
             var args = new GitArgumentBuilder("for-each-ref")
             {
-                "--sort=-committerdate",
                 "--sort=-taggerdate",
                 "--format=\"%(refname)\"",
-                "refs/"
+                "refs/tags/"
             };
 
             string tree = Module.GitExecutable.GetOutput(args);
@@ -257,18 +256,18 @@ namespace GitUI.CommitInfo
             return dict;
         }
 
-        private async Task LoadSortedRefsAsync()
+        private async Task LoadSortedTagsAsync()
         {
             ThreadHelper.AssertOnUIThread();
-            _refsOrderDict = null;
+            _tagsOrderDict = null;
 
             await TaskScheduler.Default.SwitchTo();
             try
             {
-                var refsOrderDict = GetSortedRefs();
+                var tagsOrderDict = GetSortedTags();
 
                 await this.SwitchToMainThreadAsync();
-                _refsOrderDict = refsOrderDict;
+                _tagsOrderDict = tagsOrderDict;
                 UpdateRevisionInfo();
             }
             catch (RefsWarningException ex)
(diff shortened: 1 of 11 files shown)


ResourceManager/LinkFactory.cs
```diff
@@ -12,12 +12,14 @@ public interface ILinkFactory
...
         string CreateBranchLink(string noPrefixBranch);
         string CreateCommitLink(ObjectId objectId, string linkText = null, bool preserveGuidInLinkText = false);
+        string CreateShowAllLink(string what);
         string ParseLink(string linkText);
     }
 
     public sealed class LinkFactory : ILinkFactory
     {
         private readonly ConcurrentDictionary<string, string> _linksMap = new ConcurrentDictionary<string, string>();
+        private readonly TranslationString _showAll = new TranslationString("[ Show all ]");
```
mstv: This `TranslationString` and those of the static class RefsFormatter are not detected.
How should I add them? `ITranslate` seems to be not prepared for `static` fields at first glance.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -264,9 +279,11 @@ private void ReloadCommitInfo()
             showMessagesOfAnnotatedTagsToolStripMenuItem.Checked = AppSettings.ShowAnnotatedTagsMessages;
             showTagThisCommitDerivesFromMenuItem.Checked = AppSettings.CommitInfoShowTagThisCommitDerivesFrom;
 
+            _showAllBranches = false;
+            _showAllTags = false;
             _branches = null;
-            _annotatedTagsMessages = null;
```
gerhardol: why change order
(just kidding, does not matter to me but you asked the same recently...)

mstv:
> One comment about missing comment, ...

Could you give an example comment, please, @gerhardol?
This small function is too obvious to me - perhaps because I wrote it myself.

> I also have seen that the branch list is empty, but that could be caused by the Synantec virus that slows down git-status 7 times. Will check again.

I don't think I have tampered with this. It always loaded all refs before. 

> I would prefer to not see the + after prio branches

Are you sure that you don't have an (additional) branch named "+ bugfix/i7086-directory-detect".
(I've seen that you had "bugfix/i7086-directory-detect" for PR 7109.)
I'd say the `+` cannot have been added by the `RefsFormatter` because it wouldn't be the first "after-prio branch" due to the alphabetical ordering.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -859,6 +828,9 @@ int IndexOf(string s)
 
                     return -1;
                 }
+
+                bool IsLocalMaster(string branch) => branch == "master";
```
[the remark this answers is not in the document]
gerhardol: // Note: This assumes that "master" is the important branch, this is not configurable
(In some projects I work with any branch without "/" is a master branch. Some have multiple like master, master_special that are related, other projects happened to use a name like main)
But right now, configuring is not worth it.

gerhardol:
> I don't think I have tampered with this. It always loaded all refs before.

Was just mentioning, I will just not approve myself until this is clear to me (it is likely slowness but I cannot troubleshoot on that PC)

> Are you sure that you don't have an (additional) branch named "+ bugfix/i7086-directory-detect".

The branch do not have + in the name.
The format of git-branch look strange though (just for this branch).
I find nothing about this in the man page

image

pmiossec:
>
> I find nothing about this in the man page
>

Is there a possibility that it's due to a worktree?

>

gerhardol:
> Is there a possibility that it's due to a worktree?

Bingo, it is the branch in the master worktree.

gerhardol:
+ for branches checked out in other worktrees also occurs in current master, it is just that this is not so visible. This is another indication that this PR is an improvement

No brances at all is due to that I did not display remote branches where I saw the behavior

gerhardol:
Separate PR for '+', could have been added here too
#7209

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                string noPrefixBranch = branch;
+                bool branchIsLocal;
+                if (getLocal && getRemote)
+                {
+                    // "git branch -a" prefixes remote branches with "remotes/"
+                    // It is possible to create a local branch named "remotes/origin/something"
+                    // so this check is not 100% reliable.
+                    // This shouldn't be a big problem if we're only displaying information.
+                    branchIsLocal = !branch.StartsWith(remotesPrefix);
```
gerhardol: improve on comment
// This could be solved by listing local and remote branches separetly

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -204,6 +201,24 @@ public void SetRevisionWithChildren(GitRevision revision, IReadOnlyList<ObjectId
             ReloadCommitInfo();
         }
 
+        public void ShowAll(string what)
+        {
+            switch (what)
+            {
+                case "branches": _showAllBranches = true; _branchInfo = null; break;
+                case "tags": _showAllTags = true; _tagInfo = null; break;
+                default: Debug.Fail("unsupported type in ShowAll(\"" + what + "\")"); return;
```
RussKie: I'm not finding this `switch` readable. It is easy to miss the second statements.
Please reformat with one statement per line.

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,131 @@
...
+        ///
+        /// If limited, the line "[Show all]" and an empty line are added.
+        /// Hence the list needs to be limited only if it exceeds MaximumDisplayedRefsIfLimited + 2.
+        /// </summary>
+        private const int MaximumDisplayedRefsIfLimited = 10;
+
+        private static readonly TranslationString _containedInBranches = new TranslationString("Contained in branches:");
+        private static readonly TranslationString _containedInNoBranch = new TranslationString("Contained in no branch");
+        private static readonly TranslationString _containedInTags = new TranslationString("Contained in tags:");
+        private static readonly TranslationString _containedInNoTag = new TranslationString("Contained in no tag");
```
RussKie: Can we not leverage the same mechanism for conditional strings as we have in `ResourceManager.Strings`?

mstv: `RefsFormatter` would be a `static` class if it did not need to implement `ITranslate`.
Only temporary instance is created in order to satisfy the translation interface.

Is there a way in C# to give the metaobject to the `Translator`?
As an alternative, these few `TranslationString`s could be moved to `ResourceManager.Strings`.

mstv: Is there a way in C# to give the metaobject to the `Translator`?

Should I move these few `TranslationString`s to `ResourceManager.Strings`, @RussKie?

mstv: I have moved the `TranslationString`s to `Strings`.

ResourceManager/LinkFactory.cs
```diff
@@ -84,6 +85,11 @@ public string CreateCommitLink(ObjectId objectId, string linkText = null, bool p
             return AddLink(linkText, "gitext://gotocommit/" + objectId);
         }
 
+        public string CreateShowAllLink(string what)
+        {
+            return AddLink("[ " + Strings.ShowAll + " ]", "gitext://showall/" + what);
```
RussKie:
```suggestion
            return AddLink($"[ {Strings.ShowAll} ]", $"gitext://showall/{what}");
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using GitCommands;
+using ResourceManager;
+
+namespace GitUI.CommitInfo
+{
+    public sealed class RefsFormatter
```
RussKie: If a class contains only `static` members, the class should be marked as `static` (a small perf optimisation).
However can we not make members static? It creates coupling and makes testing harder

mstv: Yes, it had been a `static` class until I implemented `ITranslate`. Forgot to turn it back to `static`.

This class contains only static methods and an instance of `LinkFactory`. But it does not use stateful methods of `LinkFactory`.

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            IEnumerable<string> FormatTags(IEnumerable<string> tags_)
+            {
+                return tags_.Select(s => showAsLinks ? LinkFactory.CreateTagLink(s) : WebUtility.HtmlEncode(s));
+            }
+        }
+
+        private static string ToString(IEnumerable<string> links, string prefix, string textIfEmpty, string refsType, bool truncated)
+        {
+            if (links.Any())
```
RussKie: Invert if to reduce nesting

RussKie: ❗️ Multiple enumerations

```cs
        private static string ToString(IEnumerable<string> links, string prefix, string textIfEmpty, string refsType, bool truncated)
        {
            var links = links.Join(Environment.NewLine);
            if (links.Length < 1)
                return string.Empty;
            // else format as necessary
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                var sb = new StringBuilder().AppendLine(WebUtility.HtmlEncode(prefix)).Append(links.Join(Environment.NewLine));
```
RussKie: [readability]
```suggestion
                var sb = new StringBuilder()
                  .AppendLine(WebUtility.HtmlEncode(prefix))
                  .Append(links.Join(Environment.NewLine));
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+                if (truncated)
+                {
+                    sb.AppendLine().AppendLine(LinkFactory.CreateShowAllLink(refsType));
```
RussKie:
```suggestion
                    sb.AppendLine()
                      .AppendLine(LinkFactory.CreateShowAllLink(refsType));
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+            return ToString(links, Strings.ContainedInBranches, Strings.ContainedInNoBranch, "branches", truncated);
+        }
+
+        public static string FormatTags(IReadOnlyList<string> tags, bool showAsLinks, bool limit)
+        {
+            bool truncate = limit && tags.Count > MaximumDisplayedRefsIfLimited + 2;
+            var links = FormatTags(truncate ? tags.Take(MaximumDisplayedRefsIfLimited) : tags);
+            return ToString(links, Strings.ContainedInTags, Strings.ContainedInNoTag, "tags", truncate);
+
+            IEnumerable<string> FormatTags(IEnumerable<string> tags_)
```
RussKie:
```suggestion
            IEnumerable<string> FormatTags(IEnumerable<string> selectedTags)
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+
+            return ToString(links, Strings.ContainedInBranches, Strings.ContainedInNoBranch, "branches", truncated);
+        }
+
+        public static string FormatTags(IReadOnlyList<string> tags, bool showAsLinks, bool limit)
+        {
+            bool truncate = limit && tags.Count > MaximumDisplayedRefsIfLimited + 2;
+            var links = FormatTags(truncate ? tags.Take(MaximumDisplayedRefsIfLimited) : tags);
```
RussKie:
```suggestion
            var formattedTags = FormatTags(truncate ? tags.Take(MaximumDisplayedRefsIfLimited) : tags);
```

GitUI/CommitInfo/RefsFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+                    allowRemote = false;
+                }
+            }
+
+            return ToString(links, Strings.ContainedInBranches, Strings.ContainedInNoBranch, "branches", truncated);
+        }
+
+        public static string FormatTags(IReadOnlyList<string> tags, bool showAsLinks, bool limit)
+        {
+            bool truncate = limit && tags.Count > MaximumDisplayedRefsIfLimited + 2;
```
RussKie: Please don't use magic numbers, use consts instead

GitUI/CommitInfo/RefsFormatter.cs
mstv: > Invert if to reduce nesting

I agree: negative cases first.

mstv: (Resharper works for me - thank you for the license!) I saw this warning, too. Though in this case, Re# is not perfect IMO.  I think it is cheaper to start the enumeration twice than to allocate an array (or a string in your quite efficient proposal) - because the first enumeration is stopped at once for `Any`.
Your proposal is mixing the positive with the negative case at the expense of readability.
So: How expensive is the second enumeration for `Any` really?

RussKie: No, it is not so much the issue with the `Any` call, but the fact that the underlying collection may be enumerated multiple times when `links.Any()` is called followed by `links.Join(...)`

If the `links` underlying collection is materialised (e.g. a concrete `List<string>` or `Array`) then there will be no further allocations.
However imaging you have a complex query that fetches data from a database, a file or a network stream, and the result hasn't materialised:
```cs
var links = LoadInfoFromDatabase(ids)
				.Union(FetchLinksFromHttp)
				.Where(id => id < 100);

var formattedLinks = ToString(links, ...);
```

When the code reaches `Any` it hasn't loaded anything from a db or hit a remote endpoint. Executing `Any` will perform that so we have a result - "have results or not?".
Then `Join` operation will **again** execute the db request and hit the remote endpoint.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -859,6 +832,10 @@ int IndexOf(string s)
 
                     return -1;
                 }
+
+                // Note: This assumes that "master" is the important branch, this is not configurable.
+                bool IsLocalMaster(string branch) => branch == "master";
```
gerhardol: New suggestion: Can this be set to all branches starting with "master"? 
That would improve my use case and I cannot see that this will be too cumbersome for other users.
(IsRemoteMaster will require a regexp)

mstv: Done. Would have been no big deal if identifiers like `aIsRemoteMaster` would not have been denied as Hungarian Notation. So dont blame me for `isRemoteMaster_a`!

grafik

mstv:
@RussKie 
> Public API must validate input parameters, please update tests as well.

I agree in principle. Here we are talking about a _formatter_ only.
I've declared the regarding arguments `[NotNull]` and added a test that checks for the regarding exceptions.

Or do you really want to have something like:
`Contained in branches: <null list>` and
`Contained in branches: current, master, <null>, upstream/master`?

In this area, I consider a `null` as a logical error and want to get NBug reports (if ever).
Though the exception should be considered non-fatal, i.e. shall not kill GE.
Should we introduce a base class "NonFatalException"? Or should we mark such exceptions with e.g. "[non-fatal]" in the exception message?

GitUI/CommitInfo/CommitInfo.cs
RussKie: > That would improve my use case and I cannot see that this will be too cumbersome for other users.

omg, Gerhard-specific feature 😁

RussKie:
More like:
```cs
        public string FormatBranches([NotNull] IEnumerable<string> branches, bool showAsLinks, bool limit)
        {
			if (branches == null)
			{
				return string.Empty;
			}

			// ...
```

An attribute decoration won't stop dereferencing a `null` at run time, it is only a hint for the compiler to help a developer.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -833,17 +769,69 @@ protected override void DisposeCustomResources()
...
+                {
+                    return -1;
+                }
+                else if (isRemoteMaster_b)
+                {
+                    return 1;
+                }
+
+#if true        // As long as GetSortedRefs() doesn't return the branches sorted by date, sort other branches alphabetically.
+                return defaultCompareResult;
```
gerhardol: Discussion: I wish to see local branches before remote and (especially on GH where I have many remotes) origin-upstream-other remotes after that. But this PR is an improvement as-is.

mstv: If you prefix the names of the other remotes with "z", you'll get most of this sorting - also in the left panel. Have a look at the screenshots.
Further tuning of the sorting should be addressed in conjunction with the git(?) bug #7181.

gerhardol: > If you prefix the names of the other remotes

Kind of what I use now.
With more information, more requests to format it will come. But I got the master* change, so I am satisfied

mstv:
> Except #if true...#else..#end

That's why I have opened #7181.
Though sorting first by remotes, then in alphabetical order seems to be preferred.
So we could throw the `#else` code away and close #7181. Opinions, @gitextensions/git-extensions-source?

gerhardol:
Ok for me, we can always revisit this

mstv:
Restored tags sorted by date; created a dedicated `BranchComparer`; avoided loading of sorted branches which was not used anymore and did not work (#7181).
I hope this was the final iteration.

UnitTests/GitUITests/CommitInfo/CommitInfoTests.cs
```diff
@@ -74,18 +75,94 @@ public void OneTimeTearDown()
         }
 
         [Test]
-        public void GetSortedRefs_should_throw_on_git_warning()
+        public void BranchComparer([Values(null, "current")] string currentBranch)
```
RussKie: Is this a fancy replacement for `TestCase`?

mstv: Yes, especially if you have a few arguments. It tests all combinations.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -204,14 +204,33 @@ public void SetRevisionWithChildren(GitRevision revision, IReadOnlyList<ObjectId
...
+                case "branches":
+                    _showAllBranches = true;
+                    _branchInfo = null; // forces update
+                    break;
+                case "tags":
+                    _showAllTags = true;
+                    _tagInfo = null; // forces update
+                    break;
+                default:
+                    Debug.Fail("unsupported type in ShowAll(\"" + what + "\")");
```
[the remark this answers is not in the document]
mstv: There's no reason to kill GE if this fails. This is only a notice to developers for the case this type of link will be extended in the future.
We can change it when we have an exception type that triggers a NBug report and does _not_ close the repo.

mstv:
@msftbot merge in 4 hours

PR #7198, opened 2019-09-25
Gerrit Plugin fixes/improvements

lekhmanrus:
## Fixes
Should apply only one % and then comma separated parameters for publish. Example:
%wip,topic=test

## Proposed changes
Added CC and Hashtag for Gerrit Publish dialog.
See details here: https://gerrit-review.googlesource.com/Documentation/user-upload.html#_git_push

### DIFF
--- a/Plugins/Gerrit/FormGerritPublish.Designer.cs
+++ b/Plugins/Gerrit/FormGerritPublish.Designer.cs
@@ -28,7 +28,6 @@
         /// </summary>
         private void InitializeComponent()
         {
-            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(FormGerritPublish));
             this.Publish = new System.Windows.Forms.Button();
             this.AddRemote = new System.Windows.Forms.Button();
             this._NO_TRANSLATE_Remotes = new System.Windows.Forms.ComboBox();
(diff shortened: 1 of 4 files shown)


Plugins/Gerrit/FormGerritPublish.Designer.cs
```diff
@@ -205,7 +247,11 @@ private void InitializeComponent()
         private System.Windows.Forms.TextBox _NO_TRANSLATE_Topic;
         private System.Windows.Forms.TextBox _NO_TRANSLATE_Reviewers;
         private System.Windows.Forms.Label labelReviewers;
-        private System.Windows.Forms.ComboBox PublishType;
+        private System.Windows.Forms.ComboBox _NO_TRANSLATE_PublishType;
```
RussKie: Could you please clarify why you have renamed these controls?

RussKie:
Please provide before and after screenshots for the UI change

Plugins/Gerrit/FormGerritPublish.Designer.cs
lekhmanrus: Actually, I'm not sure if it needs to be renamed. As I understand, it is _NO_TRANSLATE_ because it is also the ComboBox Name. I renamed it to keep consistency with all another controls, like: Topic, Branch, Remotes, Reviewers. They all shouldn't have _NO_TRANSLATE_ prefix, am I right?

lekhmanrus:
> Please provide before and after screenshots for the UI change

I'll provide screenshots tomorrow. And also regarding AppVeyor error - I'll add translations.

Plugins/Gerrit/FormGerritPublish.Designer.cs
RussKie: We name certain controls with `_NO_TRANSLATE` prefix to exclude those from translations, e.g. those that hold value that should not be translated, e.g. part of git commands, urls, etc.

Looking at the original LOC value of `PublishType`, it should be translated, hence no prefix should be required.
Unless there is a new requirement, which isn't totally obvious.

lekhmanrus: In this case Topic, Branch, Remotes and Reviewers also can be translated.

RussKie: Values in these textboxes are user-provided, so they are not translated

lekhmanrus:
**Before:**
old

**After:**
new

RussKie:
Thank you.

Few questions looking at the screenshots:
* Do you know why the width of the first column reduced?
* Does "hashtag" need to reside on the same line as "publish type"? If not, I prefer it to be places underneath.

Also I would prefer for the button to be placed on its own line, underneath all controls. I know it is kind of an existing arrangement, but it is a good time to adjust it.
This will get a slightly better appearance.

As a stretch goal, if you want to make it look and scale better 😉, the layout should change to a `TableLayoutPanel` with autosizing (in a separate commit).

lekhmanrus:
**Before:**
old

**After:**
new

RussKie:
Awesome! Thank you.
Please squash and it is good to go.

lekhmanrus:
Done, or any another feedback here?

RussKie:
Thank you

PR #7209, opened 2019-09-27
Branch list could contain '+'

gerhardol:
Seen in CommitInfo

See https://github.com/gitextensions/gitextensions/pull/7186#issuecomment-535709962

## Proposed changes
Remove + in front of branches that are checked out in other worktrees

## Screenshots
See #7186 

## Test methodology
Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2961,7 +2961,7 @@ namespace GitCommands
                 return Array.Empty<string>();
             }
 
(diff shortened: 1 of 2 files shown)


RussKie:
It would be great to have a test.
Could you please set up at least a stub for a test and the input data so the body of the test can be added later

gerhardol:
Added simple tests including both the Git arguments and outputs.
The parsing could be changed to skip first two spaces and set local/remote, but very low prio.

RussKie:
> Added simple tests including both the Git arguments and outputs.

Champion!

PR #4766, opened 2018-04-03
Refactor repositories persistence

RussKie:
Refactor recent and remote repositories history handling, replacing singleton classes and complex hierarchy with simplified object model.

* The proposed API retains the top level singleton class `RepositoryHistoryManager` which contains only two properties for repository managers for local git repositories (`Locals`) and remote git repositories (`Remotes`). Both the repository managers inherit from the same interface and are very similar in nature.
* The existing namespace has also been updated to resolve namespace conflict with the `Repository` DTO.
* The existing types `RepositoryCategory` and `RepositoryHistory` are deprecated and removed, the object hierarchy has been flattened from three levels to one.

This is a precursor for #3693, which will leverage the new repository handling mechanics for display and management of both recent and classified repositories.

This PR also renders the dashboard completely inoperable.

PR #7213, opened 2019-09-29
Dark theme

NikolayXHD:
Implements #5489

gitext_color_scheme


RussKie:
Clever 😉

GitUI/ColorSchemeController.cs
```diff
@@ -0,0 +1,247 @@
...
+                return true;
+            }
+
+            return false;
+
+            bool TryDeserialize(string input, out IReadOnlyDictionary<int, int> result)
+            {
+                result = null;
+                var lines = input.Split(
+                    new[] { Environment.NewLine },
```
gerhardol: Suggest to split on "\n" too (required for me).

gerhardol:
Nice start, many parts are usable. (Some controls are not affected, others like marked text are hard to read.)
It would be nice to have a configurable for the setting, then this branch can be used as a daily driver.

GitUI/ColorSchemeController.cs
NikolayXHD: ok

gerhardol:
Not just a dark theme but a theme engine!

This will be hard to describe. Maybe it is OK to keep the current load/save (maybe add Reset to Dark too) on the colors page and have the colors on a popup as is. Special usages can be documented in the wiki or GE Doc. I am sure users will contribute more schemes.

For the remaining controls to change (like panel backgrounds), have you encountered any need to override controls?

NikolayXHD:
> Are read/write case insensitive?

No, but I've put ToLowerInvariant(...) call to keep current color keys like `"othertagcolor"`.
And I initially forgot to add `"color"` to the end :)

NikolayXHD:
> For the remaining controls to change (like panel backgrounds), have you encountered any need to override controls?

scrollbars, `ComboBox`, `TabControl`, `GroupBox`, `ToolStripMenuButton` (the black triangle which opens menu).

RussKie:
> > For the remaining controls to change (like panel backgrounds), have you encountered any need to override controls?
> 
> * scrollbars, `ComboBox`, `TabControl`, `GroupBox`, `ToolStripMenuButton` (the black triangle which opens menu), also `ListBox` group header, `TreeView`,
> * `ICSharpCode.TextEditor` when showing file content uses bright colors.

We can consider making controls themeable in .NET Core 😉 Just saying

gerhardol:
> We can consider making controls themeable in .NET Core

Until then, should we release as is or override the most important controls (in this or follow up PRs)?
TreeView is the most recognizable.

If the default look is OK, the theme-setting just need to be decent to merge without overrides.

RussKie:
> Until then, should we release as is or override the most important controls (in this or follow up PRs)?

I didn't mean we should abandon this work. What @NikolayXHD is doing is pure awesomeness!
It just if we run in any nasty issues it may worth considering sending a fix to https://github.com/dotnet/winforms/ repo

RussKie:
A thought just occurred to me - we do have https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/DashboardTheme.cs.
I think with this work, it may be prudent to remove `DashboardTheme` and update the dashboard to use the new theming engine.

gerhardol:
> I didn't mean we should abandon this work. What @NikolayXHD is doing is pure awesomeness!

Of course we should use this. 
The question was what controls that is acceptable to not be themed in a release to users with a certain level of promotion to users.
It could be that we merge exactly like it is now to nicer configuration to some overridden controls (in this or follow up PRs).

NikolayXHD:
I was able to fix ComboBox, GroupBox, TextBox, Buttons.
TextBox and ComboBox fixes however required overriding system colors deeper.

Some controls call Win32API `GetSysColorBrush` directly and save the brush forever, therefore overriding `SystemBrushes` for them is not enough.

I overcame this problem by directly overriding Win32API calls to retrieve system colors and brushes using hooking mechanism.

gitext_color_scheme_v2

It however does not support `AnyCPU` without `Win32 preferred`  in executable .csproj. It fails with message 'Hooking near juncture point is not supported'.

So if we go this route we have to deliver x64 and x86 separately or turn on `Win32 preferred`

NikolayXHD:
`EasyHook` was added as a source code, instead of NuGet package, because the version from package fails with `BadImageFormat` exception when installing a hook with any build settings I tried.

gerhardol:
@cryham any opinions?

As a start, this is great!
Of course, there will be a lot of visual style issues. I will list some issues I see (as an example), primarily with the dark theme (on Win10 set to use default visuals with Light theme).
Looked very briefly at the code, no major objections.

For the x64/x86 issues, @ Johno-ACSLive has looked into using advanced installer rather than vsix, that step should maybe be done first?

My take on merging this is the following:
 * The interface configuration changes are OK (I believe so)
 * No visual breakage for the default setup (seem OK for me)
 * Dark theme has no severe visual glitches (some candidates below)
 * As much as reason able possible dark overrides are applied (probably OK)
 * Code review 

--

image

Cannot save scheme (this should probably be saved by default outside of Program Files, installed in user settings).
image

image

Plugin issue
image

I have set GE to use the path to the editor built in this branch, except for sidepanel this is OK for e.g. gitignore

image

Visual style is not applied when editing file though:
image

image

RussKie:
Mate you're a legend!

I wouldn't concern ourselves with any plugins at this stage.

On a phone, will review once I get to a computer

gerhardol:
> I wouldn't concern ourselves with any plugins at this stage.

They do not run at all, which is a blocker for merge.

NikolayXHD:
Tree view is ok except for collapse / expand buttons

image

Tab page `Background` is ok now.

Check boxes below look weird because they lie on a panel with `BackColor == Color.Transparent`.

Disabled controls do not look good, example is `TexBox` for Url.

image

Fixed merge conflict colors

image

gerhardol:
Magic!
Very few controls that are unusable now.
The plugins are back again for me.

There can be improvements of course. The most in-your-face is probably the scrollbars and tabs (both you said were difficult).
Should this PR be tweaked some more or be reviewed and merged now?
(I do not have much formal comments on the code.)

RussKie:
I do, I need to find time and write them.
Please do not merge until I respond.

Johno-ACSLive:
I'll be working on the x64/x86 installer in about 2 weeks. This will require the CPU architecture to be specified for the executables instead of using AnyCPU.

I share the same concern around EasyHook requiring elevation.

I love the idea of moving to WPF and .NET Core!

ColorScheme/EasyHook/Config.cs
```diff
@@ -0,0 +1,375 @@
...
+        public static String GetDependantSvcExecutableName()
+        {
+            return DependencyPath + GetSvcExecutableName();
+        }
+
+        /// <summary>
+        /// REQUIRES ADMIN PRIVILEGES. Installs EasyHook and all given user NET assemblies into the GAC and
+        /// ensures that all references are cleaned up if the installing application
+        /// is shutdown. Cleanup does not depend on the calling application...
+        /// </summary>
```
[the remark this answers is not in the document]
NikolayXHD: No, this method is not used. To avoid confusion I will remove it.

NikolayXHD:
> I propose to move the added projects to submodules

How do I do this? I suppose I create a repo in my account then some maintainer forks it into GitExtensions account, right?

> Do you know if this approach will work on W7, W8 and W10?

I am running it on W10 x64,, will also test on a virtual machine with W7 x64.

> Could you please clarify whether the new libraries will have any impact on the user's experience?

I think it won't, because the feature we are using is a process-local system calls hooking which is not supposed to be alarming for antiviruses or require extra privileges.

> ColorScheme/EasyHook/Config.cs states it requires admin privileges.

The method `Register` which requires elevation is not actually used by us. Its for EasyLoad and EasyHookSys libraries which reference EasyHook.dll so EasyHook.csproj has some code to address their needs. I did not take those, because there is no need, but did not alter EasyHook to leave only required code either.

NikolayXHD:
> I love the idea of moving to WPF and .NET Core!

Even I do, of cource WPF is better than this, what I call it, poor man's color scheme for WinForms :)

RussKie:
> How do I do this? I suppose I create a repo in my account then some maintainer forks it into GitExtensions account, right?
> 
> > Do you know if this approach will work on W7, W8 and W10?

I can create a repo for you and give you admins to that.
Should we fork the original repo or do you want a brand new empty repo?

RussKie:
> I love the idea of moving to WPF and .NET Core!

I am afraid I can't see a migration to WPF happening in observable future.

Johno-ACSLive:
> I am afraid I can't see a migration to WPF happening in observable future.

Is it an issue with dependent components or time issue (a lot of time to convert the UI)?

RussKie:
it is primarily a time issue (and resource of course) - to decouple UI from
all business logic.
in essence that would be a complete rewrite.

2ndly, WinForms were x-plat and ran on Linux under Mono, WPF was never
x-plat.
I'm hoping (more like dreaming) Mono WinForms will make a come back

gerhardol:
>> How do I do this? I suppose I create a repo in my account then some maintainer forks it into GitExtensions account, right?

> I can create a repo for you and give you admins to that.

If @NikolayXHD create the submodule, GE core can fork the repo (similar to GitExtensionsDoc). It is a matter of taste, if you want to see the origin of the repo or not. (I have no problem voting for admin access for any repo for you.)

NikolayXHD:
To preserve EasyHook history, I will fork EasyHook, remove all unnecessary projects / files, and commit, then GE core forks it from my fork.

RussKie:
I'll do it for you in GE org and make you an admin there

RussKie:
Done, please check your email for an invite to the repo.
Please reach out directly if you have any issues.

Johno-ACSLive:
Happy to work on porting to WPF.

Maybe we can convince the WPF team to get the UI running on Linux / Mac 😉

Unless someone in the community has started that effort.

NikolayXHD:
diff
image
view
image
blame
image
editor
image

RussKie:
> Maybe we can convince the WPF team to get the UI running on Linux / Mac

I *personally* think that would be a wasted effort, WPF isn't going to be x-plat'ed. We probably have more chance of seeing Mono WinForms coming back. Though I doubt it too... 😒

RussKie:
Is it possible not to commit dlls?

Johno-ACSLive:
That's a shame RE WPF having no plans for being x-plat!

I guess QT, some other framework or using native UI per platform is the only route.

I thought part of the .NET 5 roadmap was meant to converge all frameworks including mono. Maybe I misinterpreted the information?

NikolayXHD:
> Is it possible not to commit dlls?

Then I would have to figure out how to build them, which given my limited experience with C++ end even more limited with C++ in Windows I would rather not to.

RussKie:
> Then I would have to figure out how to build them, which given my limited experience with C++ end even more limited with C++ in Windows I would rather not to.

Would `msbuild xxx.vcproj` work?

RussKie:
Similar to how we do it here: https://github.com/gitextensions/gitextensions/blob/master/Setup/BuildGitExtNative.cmd

NikolayXHD:
> Would `msbuild xxx.vcproj` work?

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(379,5): error MSB8020: The build tools for Visual Studio 2013 (Platform Toolset = 'v120') cannot be found.
To build using the v120 build tools, please install Visual Studio 2013 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then sele
cting "Retarget solution". [f:\repo\git\gitextensions\Externals\EasyHook\EasyHookDll\EasyHookDll.vcxproj]
```

http://paste.org.ru/?2yk10k

RussKie:
That's cool no panic :) I'm no C++ expert (...or notice) either 

Do you build the full solution at all? E.g. we have few C++ projects in the solution - GitExtensionsShellEx and GitExtSshAskPass
These two use the default toolset:
image

You can update the toolset to the same

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "Externals/ICSharpCode.TextEditor"]
 	path = Externals/ICSharpCode.TextEditor
 	url = ../../gitextensions/ICSharpCode.TextEditor.git
+[submodule "Externals/EasyHook"]
+	path = Externals/EasyHook
+	url = git@github.com:gitextensions/EasyHook.git
```
RussKie:
```suggestion
	url = ../../gitextensions/EasyHook.git
```

and you need to commit the hash of the submodule... otherwise there is nothing to clone

RussKie:
image

all of these must now be gone as they are coming from the submodule.

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections.Generic;
+using System.Windows.Forms;
+
+namespace GitUI
+{
+    internal static class ControlUtils
```
RussKie: This class contains extensions, and it should be named accordingly
```suggestion
    internal static class ControlTagExtensions`
```

The class also feels misplaced in the "Theme" folder.
A number of extensions live in GitExtUtils/GitUI, perhaps it is more appropriately placed there

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public static bool HasTag<TValue>(this Control control)
+        {
+            if (control.Tag == null)
+            {
+                return false;
+            }
+
+            var dict = (Dictionary<string, object>)control.Tag;
```
RussKie: This has a high risk of throwing.
Please wrap it in `try...catch` to avoid crashing the app

Please write tests to ensure all edge cases are handled correctly

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+            GetTag<TValue>(control, typeof(TValue).FullName);
+
+        public static TValue GetTag<TValue>(this Control control, string key)
+        {
+            if (control.Tag == null)
+            {
+                control.Tag = new Dictionary<string, object>();
+            }
+
+            var dict = (Dictionary<string, object>)control.Tag;
```
RussKie: ditto

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            if (!dict.TryGetValue(key, out var result))
+            {
+                return default;
+            }
+
+            return (TValue)result;
```
RussKie: ditto

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+            SetTag(control, typeof(TValue).FullName, value);
+
+        public static void SetTag<TValue>(this Control control, string key, TValue value)
+        {
+            if (control.Tag == null)
+            {
+                control.Tag = new Dictionary<string, object>();
+            }
+
+            var dict = (Dictionary<string, object>)control.Tag;
```
RussKie: ditto

GitUI/Theme/Win32ColorHooks.cs
```diff
@@ -0,0 +1,99 @@
...
+    public static class Win32ColorHooks
+    {
+        private static Theme _theme;
+        private static ColorDelegate _colorBypass;
+        private static BrushDelegate _brushBypass;
+        private static LocalHook _colorHook;
+        private static LocalHook _brushHook;
+
+        [DllImport("gdi32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+        public static extern IntPtr CreateSolidBrush(int nIndex);
```
RussKie: Why is it public?

GitUI/Theme/Win32ColorHooks.cs
```diff
@@ -0,0 +1,99 @@
...
+            if ((nIndex & 0xffffff00) == 0)
+            {
+                nIndex |= -0x80000000;
+                return ColorTranslator.FromOle(nIndex).ToKnownColor();
+            }
+
+            return ColorTranslator.FromWin32(nIndex).ToKnownColor();
+        }
+
+        public static void Uninstall()
```
RussKie: Please keep public API definitions above privates

GitUI/Theme/dark.thm
```diff
@@ -0,0 +1,41 @@
+ActiveBorder: 5e6060
```
RussKie: Can we have `.theme` extension?

GitUI/Theme/Win32ColorHooks.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System;
+using System.Drawing;
+using System.Runtime.InteropServices;
+using EasyHook;
+
+namespace GitUI
+{
+    public static class Win32ColorHooks
```
RussKie: Can we make it `internal`?

GitUI/Theme/ThemePersistence.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Drawing;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Windows.Forms;
+
+namespace GitUI
+{
+    public static class ThemePersistence
```
RussKie: Does it have to be `static`?
Statics increase coupling and make testing story very difficult.

This Q goes to all other new API

gerhardol:
@NikolayXHD 
Can you please fork the ICSharpCode.TextEditor repo and eventually make a PR there too?
I can currently not build the project, the commit is in your private clone
(I have no problem changing the module url pointing to your GH account until it is merged - for me you could get a branch working there too)

I have been using the build (on top of other changes) for the last week. Opinion stands, there are many parts that can be improved but this is usable.
I am more irritated by the ugly white background square for the spinner when loading the revision grid, than the text that is hard to read.

GitUI/CommandsDialogs/SettingsDialog/Pages/ColorsSettingsPage.cs
```diff
@@ -44,46 +48,62 @@ protected override void SettingsToPage()
             DrawNonRelativesGray.Checked = AppSettings.RevisionGraphDrawNonRelativesGray;
             DrawNonRelativesTextGray.Checked = AppSettings.RevisionGraphDrawNonRelativesTextGray;
             chkHighlightAuthored.Checked = AppSettings.HighlightAuthoredRevisions;
+            chkUseSystemVisualStyle.Checked = AppSettings.UseSystemVisualStyle;
+
+            UpdateAppColors();
+
+            _syncingTheme = true;
+            UpdateComboBoxColorScheme(AppSettings.UITheme);
+            _syncingTheme = false;
```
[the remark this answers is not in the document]
NikolayXHD: done

GitUI/Theme/ControlUtils.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections.Generic;
+using System.Windows.Forms;
+
+namespace GitUI
```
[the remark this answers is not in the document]
NikolayXHD: ok, Theming, namespace fixed.

GitUI/Theme/ControlUtils.cs
NikolayXHD: ok

NikolayXHD:
> Can you please fork the ICSharpCode.TextEditor repo and eventually make a PR there too?
https://github.com/gitextensions/ICSharpCode.TextEditor/pull/13

GitUI/Theme/ControlUtils.cs
NikolayXHD: Rewritten the code to avoid invalid cast exceptions.

SetTag raises `InvalidOperation` if `Tag` property is occupied by something else,
because the pair GetTag / SetTag do not erase exisitng values, therefore not throwing would be silently breaking the contract.

Added tests.

GitUI/Theme/Win32ColorHooks.cs
NikolayXHD: fixed

GitUI/Theme/Win32ColorHooks.cs
NikolayXHD: ок

GitUI/Theme/dark.thm
NikolayXHD: .theme is used by windows theme

GitUI/Theme/Win32ColorHooks.cs
NikolayXHD: ok

GitUI/Theme/ThemePersistence.cs
NikolayXHD: ok

GitUI/Theme/dark.thm
NikolayXHD: changed to .colors

RussKie: > .theme is used by windows theme

It doesn't stop us from using the same 😉 



> changed to .colors

Works for me

NikolayXHD:
> 2ndly, WinForms were x-plat and ran on Linux under Mono, WPF was never x-plat.

I am using GitExtensions 2 at work with Linux, can't live without it, I am soooo grateful to GitExtensions team for dealing with all those Mono issues to let me use It.

NikolayXHD:
Fixed `CheckBox`es in `FormPush`

before:
image

after:
image

Fixed `CheckBox`es in `DataGridView`

before:
image

after:
image

I know, top and left checkbox borders are looking strange, but this is how they are supposed to be with `FlatStyle.Flat`: top and left borders are drawn with `ForeColor` which is same as the color of check glyph.

gerhardol:
Progressing...
How should we handle remaining issues?
Like titlebar, tabs and scrollbars that likely not will be handled but also minor issues like the commit with white dots:
image

NikolayXHD:
That was icon issue, now its ok

image

`ListView`s and `TreeView`s icons though are not rendered awesomely at the moment :)

gerhardol:
> That was icon issue, now its ok

Of course, the icon was just an example.
There could be many similar issues and many items that can br tuned too.
Would a wiki page be a good way to track this, listing what is considered but cannot be changed now as well as minor issues and possible improvements?

NikolayXHD:
> > That was icon issue, now its ok
> 
> Of course, the icon was just an example.
> There could be many similar issues and many items that can br tuned too.
> Would a wiki page be a good way to track this, listing what is considered but cannot be changed now as well as minor issues and possible improvements?

True, people will notice and report, so it's good to have a kind of current situation report to make this organized.

NikolayXHD:
image

image

image

image

NikolayXHD:
image

NikolayXHD:
Fixed tab controls for good. Painted completely by hand.

image

gerhardol:
I have added a page in the wiki with status, where anyone can edit
https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

Some minor issues added in addition to the posts above (some issues in the screenshots above has been fixed already, I did not update the annotated screenshots).

Adding images here too (partly due to that it is not possible to paste images directly to the wiki).
I have not looked in to these issues at all myself.
As said before, it may be OK to merge with visuals like this, this is to document the status.


image

image

image

image

image

image

image

image

image

image

image

gerhardol:
Plugins do not appear, again. Only _Find large files_ and occasionally _Gitflow_ is available.

NikolayXHD:
> I have added a page in the wiki with status, where anyone can edit
https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

I can't edit yet:
image

NikolayXHD:
> I have added a page in the wiki with status, where anyone can edit

Fixed some easy-to-fix issues

Auto lightness correction for help images

image

Branches sidebar transparent icon borders, added white background to GH icon

image

File history tab icon borders:

image

gerhardol:
>> I      have added a page in the wiki with status, where anyone can edit
    https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

> I can't edit yet:

Permissions were changed after some spamming, so only members can edit, sorry.
I would be happy to see you as a member.
Without public way to change, it will mean that a member maintains such a page (could be as a issue/PR post). Many posts will be opinions anyway. The way this PR has progressed, there may not be much issues at merge...

gerhardol:
> Plugins do not appear, again. Only Find large files and occasionally Gitflow is available.

This is likely not this PR
The problem occurs when running the Debug build tries to load the PloginManager in 
%LOCALAPPDATA%\GitExtensions\UserPlugins\GitExtensions.PluginManager

That fails:
```
"An exception was thrown while initializing part \"GitExtensions.PluginManager.Plugin\"."

InnerException = {"Attempt by method 'GitExtensions.PluginManager.Plugin..ctor()' to access method 'ResourceManager.GitPluginBase..ctor()' failed."}

"   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.RuntimePartLifecycleTracker.CreateValue()
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.Create()
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState nextState)\
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.GetValueReadyToExpose()
   at Microsoft.VisualStudio.Composition.ExportProvider.<>c__DisplayClass53_0.<CreateExport>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.Composition.ExportProvider.<>c__DisplayClass70_1`2.<GetExports>b__1()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at GitUI.PluginRegistry.<>c.<Initialize>b__10_0(Lazy`1 lazy) in F:\\dev\\gc\\gitextensions_4\\GitUI\\Plugin\\PluginRegistry.cs:line 43
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at GitUI.PluginRegistry.Initialize() in F:\\dev\\gc\\gitextensions_4\\GitUI\\Plugin\\PluginRegistry.cs:line 43"
```

PortableBuilds are not affected. Debug builds could just ignore the AppData folder too, but this will be a problem when using multiple versions of GE

gerhardol:
From what I have seen, only System.Windows.Forms.MessageBox is important to fix. But there are several hundreds of them...

RussKie:
Please reabse, the issue was fixed in #7296

gerhardol:
> Please reabse, the issue was fixed in #7296

I have rebased on master, the PR has #7296 too

NikolayXHD:
missing plugins problem should be fixed

gerhardol:
> missing plugins problem should be fixed

No, the PluginManager in AppData is for the 3.2 release and GE makes an exception when loading it.
The problem is not related to this plugin though

NikolayXHD:
> No, the PluginManager in AppData is for the 3.2 release and GE makes an exception when loading it.
The problem is not related to this plugin though

It fixed the problem for me though:
image

NikolayXHD:
> No, the PluginManager in AppData is for the 3.2 release and GE makes an exception when loading it.

That's right, still the fix enables loading the PluginManager from AppData, then all plugins load, which is convenient.

image

gerhardol:
> That's right, still the fix enables loading the PluginManager from AppData, then all plugins load, which is convenient.

Hm, it works for me now too...
Unless @NikolayXHD submits a PR the next hours I will (credit is of course @NikolayXHD )

NikolayXHD:
When correcting luminosity with `AdaptForeColor` I was always getting too dark blue color for links.

I checked the formulas, the calculations worked as expected. Then I checked with Paint whether I perceive blue, red and green colors of same luminosity as being equally bright.

Turned out Green seems way brighter and blue way darker than the value produced by formula.

I ruled out screen profiles by trying different, I ruled out screen model by trying different screen,
I ruled out my physiological vision by asking my wife to do the same comparisons.

In the picture below the background varies from 0 to 1 luminocity.
The colored bars are all 0.5.

The marks in the picture tell where do we see the colored bar luminosity match the background.

check it yourself

Now I find link colors satisfying.

link

RussKie:
I don't entirely understand the above explanation, but I am still finding the blue still somewhat too dark to be easily readable.

Can we, perhaps, leverage colours Windows uses in dark themes?
Another example, https://docs.microsoft.com/en-us/dotnet/framework/winforms/

* Light theme:
image
* Dark theme
image
* HC theme
image

NikolayXHD:
> I don't entirely understand the above explanation

I am toking about the mechanism for automatically adjusting lightness of some colors to match current theme. It is applied to CodeEditor syntax highlighting, some icons and images. This relieves us from the need to provide separate icons / images / syntax highlighting schemes for each theme.

> but I am still finding the blue still somewhat too dark to be easily readable.

This is great, now I am even more sure, it's not my vision issue!

I can adjust this:
image

> Can we, perhaps, leverage colours Windows uses in dark themes?

We could use it with links, but not with a-priori unknown colors, like text / back color in code editor or an icon.
image

Observe how poorly visible the black character is on blue background, yet the lightness was transformed the same way it was for red and violet circles.

gerhardol:
> This is great, now I am even more sure, it's not my vision issue!

I find dark themes being more sensitive to displays (and settings). When looks great at home can be hard to see on my crappy monitor at work for instance.

Other than that, I try to keep the wiki page updated (will replace som images later).
https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

As the installer is not sorted out, it will be hard to include this in 3.3
But we should get the build running so users can test portable builds
(translations need to be fixed, maybe Appveyor config tweaked too.)

mast-eu:
I pushed in the updated English.xlf to obtain the build artifacts on AppVeyor.

@NikolayXHD Please feel free to squash my commit into one of yours if you want to.

NikolayXHD:
Themes subdirectory did not make it to portable .zip, looking into it.

gerhardol:
> Themes subdirectory did not make it to portable .zip, looking into it.

All files in the portable must be explicitly added in the MakePortableArchive.cmd
A portable build is at least a way to provide the functionality to users
(Sadly, I removed this from the 3.3 milestone for now at least.)

gerhardol:
Color label in Settings is lost
https://github.com/gitextensions/gitextensions/issues/5489#issuecomment-546587438

A very minor issue with an image:

image
I do not mind if this is not fixed, but I can as well mention it.
(this level of remarks indicates how good this PR has tweaked the appearance, a great job.)

gerhardol:
For me, this PR is ready for a larger public, there are few annoyances
I have updated the wiki page with the status:
https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

How to review and merge this PR?

ICSharpCode can be merged as is. (submodule changes are merged, superproject to be updated)
Project changes like "netstandard" separately too?

EasyHook is more or less a separate step. This is an existing project, adapted (limited) to work with GE. How to review? Changes to the base project is obvious, but also the usage.

Most "Form" changes could be added separately as well, using AppSetting or System colors, but is it worth it?
The core problem that EasyHook 32/64 bit dlls have to be adapted in the installer.

Johno-ACSLive:
> The core problem that EasyHook 32/64 bit dlls have to be adapted in the installer.

I resumed working on this last night, I've been able to convert the WiX installer to Advanced Installer project but I am planning to use this as a reference. I have created both a 64-bit only installer and 32-bit only installer.

We will need to prepare the current version of GE to handle installer selection based on OS Architecture - this will be tricky for older versions skipping to future versions.

First things first - generate installers for each architecture and then see what we can do for the updater. I will generate an issue and PR for this work.

GitExtensions/GitExtensions.csproj
```diff
@@ -132,4 +132,12 @@
     </ItemGroup>
     <Copy SourceFiles="@(Translations)" DestinationFolder="$(TargetDir)Translation" ContinueOnError="false" />
   </Target>
+  <Target Name="BuildEasyHookDll" AfterTargets="AfterBuild">
```
hieuxlu: @NikolayXHD Minor build issue: I can't build the solution on my machine. As a C# dev, I hope that I don't have to install MS VC++ (some 4.5GB) to build this project. Is that possible?
 ``` 
EasyHookDLL.vcxproj:L20:  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
```

Comment (NikolayXHD) [reply GitExtensions/GitExtensions.csproj]: Made it conditional, tell me if it is was enough to fix the build in your machine.

Comment (hieuxlu) [reply GitExtensions/GitExtensions.csproj]: It did, I can build successfully now. Thank you very much.

File: GitUI/Theming/Win32ThemingHooks.cs [hunk 17]
```diff
@@ -0,0 +1,152 @@
...
+
+            (_messageBoxWHook, _messageBoxWBypass) = InstallHook<MessageBoxWDelegate>(
+                "user32.dll",
+                "MessageBoxW",
+                MessageBoxWHook);
+        }
+
+        public static void Uninstall()
+        {
+            _colorHook.Dispose();
```
hieuxlu: On my development machine, EasyHookDLL is not compiled, so all the hook objects here are null when GE exits. Do you think we need null checking here? Some thing like:
```csharp
            _colorHook?.Dispose();
            _brushHook?.Dispose();
            _messageBoxAHook?.Dispose();
            _messageBoxWHook?.Dispose();
```

gerhardol: Is it really feasible to have EasyHook as conditional?
I would prefer to have one setup, which means that C++ will be mandatory.

NikolayXHD: that's right

hieuxlu: OK, I understand. Do we need to update [build instructions on Wiki](https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions) to add EasyHook as one of the reasons why we need VS C++, apart from building the installer?

gerhardol: Yes when merging this PR.
This may require changes to the installer though, so maybe not as soon as the state of this PR indicates.

GitExtensions/GitExtensions.csproj
RussKie: We have several C++ projects in the solution, so to build an installer one needs to have C++ installed.
However I don't feel it is necessary to force people to build and use EasyHook locally, if we can avoid it and developers don't desire to do so.

GitUI/Theming/Win32ThemingHooks.cs
RussKie: I don't feel it is necessary to force people to build and use EasyHook locally, if we can avoid it and developers don't desire to do so.

GitExtensions/GitExtensions.csproj
NikolayXHD: I was doing this PR with graceful degradation in mind. There are some changes yet to be made with respect to it, but it's not hard and I was going to do them anyway.

hieuxlu:
@NikolayXHD There's 1 minor bug I found: If you click Apply -> OK, the theme is not applied. Clicking OK without Apply is fine. This is a greate feature, looking forward to customizing GE after it is merged 👍
 
image

NikolayXHD:
image

RussKie:
👍

gerhardol:
> Boy I am so proud of this!

Yet another detail to be proud of, there are so many here.
So good.

--
A minor comment:
With two scrollbars, the lower right corner is bright.
Discussion: The scrollbars are a little too dark for me. There is a nice looking animation that lightens up the bar when scrolling, but I rather keep the bars a little lighter by default (not intuitive what is the scrollbar part now). Compare to VS.

image

wiki updated:
https://github.com/gitextensions/gitextensions/wiki/Dark-Theme-status

NikolayXHD:
> There's 1 minor bug I found: If you click Apply -> OK, the theme is not applied. 

Should be fixed.

> The scrollbars are a little too dark for me

The scrollbar uses variations of `Control` system color `ControlLight`, ` ControlDark`, `DarkDark`
Adjusted these to make scrollbar more prominent

NikolayXHD:
fixed Scrollbar corner and ListView group header text color

image

NikolayXHD:
Fixed GridView column header

image

NikolayXHD:
fixed background of disabled `TextBox` and `ComboBox`es

image

gerhardol:
This PR feel complete now, appearance-wise. Much farther than I could imagine.
The installer issue and code review is really what is needed.

That said, I took screenshots of what I consider could be improved. I find these minor and did not add them to the wiki page, at least now.

--

Picker could be themed

image

Some dropdowns are not themed initially. When opening settings, it appeared like below. Selecting another page, it is OK. "problem" reappearing when opening settings again.

image

List view checkboxes still bright

image

Some headers are still bright

image

image

Form is bright white when opening, then just a bright checkbox (listview?)

image

image

Form push background is a little unclear (similar push tags)

image

NikolayXHD:
fixed ListView column header, NumericUpDown

spin

NikolayXHD:
The tecnique that helped get themed scrollbar, was applied to rest of controls, instead of using `FlatStyle.Flat` or `FlatStyle.Popup`.

Controls look less different from normal and better when disabled (except editable ComboBox).
Besides the fix is applied more consistently, it does not differentiate between CheckBox control, checkbox inside GridView or ListView and etc.

native_check_radio_treeview-expand

better_looking_disabled

NikolayXHD:
themed ToolStripMenuItem check glyph

image

GitUI/Editor/FileViewer.cs
```diff
@@ -102,29 +103,29 @@ public FileViewer()
 
             IgnoreWhitespace = AppSettings.IgnoreWhitespaceKind;
             OnIgnoreWhitespaceChanged();
-            bool light = ColorHelper.IsLightTheme();
 
-            ignoreWhitespaceAtEol.Image = light ? Images.WhitespaceIgnoreEol : Images.WhitespaceIgnoreEol_inv;
```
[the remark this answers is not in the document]
gerhardol: Yes, they are adapted dynamically
(the inv images were added this summer with the first dark theme adaptions).

gerhardol:
Note: First part of this is merged, the changes in ICSharpCode.TextEditor is now in master.

gerhardol:
When changing commits quickly, I get the following exception occasionally. No simple way to recreate
Maybe add an argument check?

```
System.ArgumentOutOfRangeException: InvalidArgument=Value of '1' is not valid for 'index'.
Parameter name: index
   at System.Windows.Forms.TabControl.GetTabRect(Int32 index)
   at GitExtUtils.GitUI.Theming.TabControlRenderer.RenderSelectedPageBackground(TabControl tabs, PaintEventArgs e) in F:\dev\gc\gitextensions_3\GitExtUtils\GitUI\Theming\TabControlRenderer.cs:line 202
   at GitExtUtils.GitUI.Theming.TabControlRenderer.HandlePaint(Object s, PaintEventArgs e) in F:\dev\gc\gitextensions_3\GitExtUtils\GitUI\Theming\TabControlRenderer.cs:line 37
   at System.Windows.Forms.Control.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.TabControl.WndProc(Message& m)
   at GitUI.CommandsDialogs.FullBleedTabControl.WndProc(Message& m) in F:\dev\gc\gitextensions_3\GitUI\CommandsDialogs\FullBleedTabControl.cs:line 24
```


## Environment
- Git Extensions 3.3.0
- Build a8733aedae6e0266a18c5282fb8918911e765b2d
- Git 2.24.0.windows.2
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

GitUI/CommandsDialogs/FormAbout.cs
```diff
@@ -22,7 +22,7 @@ public FormAbout()
 
             environmentInfo.SetCopyButtonTooltip(_copyTooltip.Text);
 
-            Color clrLink = Color.Blue; // ColorHelper.IsLightTheme() ? Color.Blue : Color.SkyBlue;
+            Color clrLink = SystemColors.Highlight; // ColorHelper.IsLightTheme() ? Color.Blue : Color.SkyBlue;
```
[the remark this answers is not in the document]
NikolayXHD: ок

GitUI/HelperDialogs/FormResetCurrentBranch.cs
```diff
@@ -33,6 +34,9 @@ public FormResetCurrentBranch(GitUICommands commands, GitRevision revision, Rese
             Revision = revision;
 
             InitializeComponent();
+            Soft.SetForeColorForBackColor();
+            Hard.SetForeColorForBackColor();
+            Mixed.SetForeColorForBackColor();
```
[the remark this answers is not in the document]
NikolayXHD: thanks, fixed

GitExtUtils/Linq/LinqExtensions.cs
```diff
@@ -129,6 +129,14 @@ public static void ForEach<T>([NotNull] this IEnumerable<T> enumerable, [NotNull
             }
         }
 
+        public static void ForEach<T, TResult>([NotNull] this IEnumerable<T> enumerable, [NotNull, InstantHandle] Func<T, TResult> func)
+        {
+            foreach (T t in enumerable)
+            {
+                var unused = func(t);
+            }
+        }
```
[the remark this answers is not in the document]
NikolayXHD: `\GitUI\Theming\Renderers\ThemeRenderer.cs:61`

```csharp
public void Dispose()
{
    _themeDataHandles.ForEach(NativeMethods.CloseThemeData);
}
```

because `CloseThemeData` returns `int`

GitUI/Editor/FileViewer.cs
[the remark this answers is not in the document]
NikolayXHD: > Yes, they are adapted dynamically

That's right

GitUI/Native.cs
```diff
@@ -84,16 +119,70 @@ internal struct COLORREF
...
+            DT_PATH_ELLIPSIS = 0x00004000,
+            DT_PREFIXONLY = 0x00200000,
+            DT_RIGHT = 0x00000002,
+            DT_RTLREADING = 0x00020000,
+            DT_SINGLELINE = 0x00000020,
+            DT_TABSTOP = 0x00000080,
+            DT_TOP = 0x00000000,
+            DT_VCENTER = 0x00000004,
+            DT_WORDBREAK = 0x00000010,
+            DT_WORD_ELLIPSIS = 0x00040000,
```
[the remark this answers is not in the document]
NikolayXHD: ok

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -238,16 +241,16 @@ internal void AddColumn(ColumnProvider columnProvider)
             Columns.Add(columnProvider.Column);
         }
 
-        private Color GetForeground(DataGridViewElementStates state, int rowIndex)
+        private Color GetForeground(DataGridViewElementStates state, int rowIndex, GitRevision revision)
```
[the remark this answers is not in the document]
NikolayXHD: incompletely reverted change, fixed

ResourceManager/GitExtensionsFormBase.cs
```diff
@@ -3,6 +3,8 @@
 using System.Linq;
 using System.Windows.Forms;
 using GitCommands;
+using GitExtUtils.GitUI;
+using GitExtUtils.GitUI.Theming;
```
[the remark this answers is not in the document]
NikolayXHD: Incompletely reverted change, fixed

GitUI/Theming/PathExtension.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.IO;
+
+namespace GitUI.Theming
+{
+    internal static class PathExtension
+    {
+        public static string InDirectory(this string file, string targetDirectory) =>
```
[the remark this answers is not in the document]
NikolayXHD: ok, moved

NikolayXHD:
> When changing commits quickly, I get the following exception occasionally. No simple way to recreate
> Maybe add an argument check?

ok

gerhardol:
Another minor:
Assert windows should not be themed
Only affect debug versions.
The assert itself is not an issue here

image

GitExtUtils/Linq/LinqExtensions.cs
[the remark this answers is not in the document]
RussKie: Is that the only use for that helper method?
If so, please inline it at the callsite.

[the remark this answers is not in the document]
NikolayXHD: ok

gerhardol:
Minor issue in Apply Patch form
Load a patch file (Browse)
Hover Apply Patch, then column headers. Appearance changes under the cursor
Switch away from GE, then back.
When hovering the changed controls, they appear correctly again

image

gerhardol:
Button themed

image

NikolayXHD:
> Minor issue in Apply Patch form
> Button themed

fixed

gerhardol:
Exception when fetching and switching commits (not easily recreated).
(Not looked at this myself)

This PR is still great, feels very complete
```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at GitExtUtils.GitUI.Theming.TabControlPaintContext.RenderSelectedPageBackground() in F:\dev\gc\gitextensions_3\GitExtUtils\GitUI\Theming\TabControlPaintContext.cs:line 244
   at GitExtUtils.GitUI.Theming.TabControlPaintContext.Paint() in F:\dev\gc\gitextensions_3\GitExtUtils\GitUI\Theming\TabControlPaintContext.cs:line 80
   at GitExtUtils.GitUI.Theming.TabControlRenderer.HandlePaint(Object s, PaintEventArgs e) in F:\dev\gc\gitextensions_3\GitExtUtils\GitUI\Theming\TabControlRenderer.cs:line 24
   at System.Windows.Forms.Control.OnPaint(PaintEventArgs e)
   at System.Windows.Forms.Control.PaintWithErrorHandling(PaintEventArgs e, Int16 layer)
   at System.Windows.Forms.Control.WmPaint(Message& m)
   at System.Windows.Forms.TabControl.WndProc(Message& m)
   at GitUI.CommandsDialogs.FullBleedTabControl.WndProc(Message& m) in F:\dev\gc\gitextensions_3\GitUI\CommandsDialogs\FullBleedTabControl.cs:line 24
```


## Environment
- Git Extensions 3.3.0
- Build 4fc4a210ba386da171d546500b54995c722d2e7a
- Git 2.24.0.windows.2
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

pmiossec:
First quick test.

* First start, I got a message box with "theme file not found" (or something like that). It's strange.
* I have selected the "dark" theme and restarted and got:

Git Extensions_dark

See also the github icons in the left panel which are not well sized.

I don't know if I missed something and I will be pleased if someone tell what I did wrong or if I could help debugging something...

gerhardol:
@pmiossec 
Did you build with the EasyHook submodule, including C++?
The colors are changed but thehhoks do not have effect
The icons are autosized, I assume that the hooks are used.

@NikolayXHD (aka the magician) has prepared to use GE without the hook, but some parts will not work well (now at least).

pmiossec:
@gerhardol 

>Did you build with the EasyHook submodule

yes, the project was loaded and it builds.

> including C++?

That was that. I will retry later. Thanks.

GitExtUtils/GitUI/Theming/ThemeFix.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Windows.Forms;
+using GitUI;
+
+namespace GitExtUtils.GitUI.Theming
+{
+    public static class ThemeFix
+    {
+        private static readonly ConditionalWeakTable<IWin32Window, IWin32Window> Ready =
```
RussKie: `Ready` is a really vague name.
Perhaps `ThemedControls` or `ProcessedControls`?

GitUI/Theming/Renderers/RectangleExtension.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Drawing;
+
+namespace GitUI.Theming
+{
+    internal static class RectangleExtension
+    {
+        public static Rectangle Inclusive(this Rectangle rect) =>
```
RussKie: How many uses does this one have?

GitUI/Theming/Renderers/ThemeProperty.cs
```diff
@@ -0,0 +1,255 @@
+﻿namespace GitUI.Theming
+{
+    internal enum ThemeProperty
```
RussKie: What are these, and where did they come from?

GitUI/Theming/ThemeController.cs
```diff
@@ -0,0 +1,142 @@
...
+        public void ResetAllColors()
+        {
+            SysColors.ForEach(Reset);
+            AppColors.ForEach(ResetInternal);
+
+            ResetGdiCaches();
+            Changed?.Invoke();
+        }
+
+        public void Reset(KnownColor name)
```
RussKie:
```suggestion
        public void ResetColor(KnownColor name)
```

GitUI/Theming/ThemeController.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            _sysColors.Remove(name);
+
+            ResetGdiCaches();
+            Changed?.Invoke();
+        }
+
+        public void Reset(AppColor name)
```
RussKie:
```suggestion
        public void ResetColor(AppColor name)
```

GitUI/Theming/ThemeController.cs
```diff
@@ -0,0 +1,142 @@
...
+                .GetField("SystemPensKey", BindingFlags.Static | BindingFlags.NonPublic)
+                .GetValue(null);
+        }
+
+        public event Action Changed;
+
+        private IDictionary ThreadData =>
+            (IDictionary)_threadDataProperty.GetValue(null, null);
+
+        public void Load(
```
RussKie: I am confused, what is the purpose of this method?

GitUI/Theming/ThemeController.cs
```diff
@@ -0,0 +1,142 @@
...
+            foreach (var (name, value) in appColors)
+            {
+                AppSettings.SetColor(name, value);
+            }
+
+            ResetGdiCaches();
+            Changed?.Invoke();
+        }
+
+        public void Save(
```
RussKie: ditto

GitUI/Theming/ThemeController.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Drawing;
+using System.Linq;
+using System.Reflection;
+using System.Windows.Forms;
+using GitCommands;
+using GitExtUtils.GitUI.Theming;
+
+namespace GitUI.Theming
+{
+    public class ThemeController : Theme
```
RussKie:
```suggestion
    public class ThemeManager : Theme
```

RussKie: Why do we need to inherit from `Theme`?

GitUI/Theming/ThemeEditor.cs
```diff
@@ -0,0 +1,214 @@
...
+using System.Drawing;
+using System.IO;
+using System.Linq;
+using System.Windows.Forms;
+using GitCommands;
+using GitExtUtils.GitUI.Theming;
+
+namespace GitUI.Theming
+{
+    public class ThemeEditor
```
RussKie: I'm struggling grasping the purpose of this class. I'm finding the name confusing, and it is overloaded with different concerns - reading/writing files, showing dialogs to the user, setting/reading colours etc. This makes the functionality quite incoherent.
We had similar problems dealing with repository histories (look at https://github.com/gitextensions/gitextensions/pull/4766) 
Can we break this apart into SOLID classes?

GitCommands/PathUtil.cs
```diff
@@ -264,5 +264,8 @@ public static IEnumerable<string> FindAncestors([NotNull] string path)
                 yield return path.EnsureTrailingPathSeparator();
             }
         }
+
+        public static string SetDirectory(this string file, string targetDirectory) =>
```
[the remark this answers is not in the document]
NikolayXHD: 3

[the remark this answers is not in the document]
NikolayXHD: Moved to `ThemeDeployment.cs`, made private, renamed to `GetPathInDirectory(file. targetDirectoru)`

GitExtUtils/GitUI/Theming/ThemeFix.cs
NikolayXHD: Indeed it is vague. Renamed to `AlreadyFixedControls`

GitUI/Theming/Renderers/RectangleExtension.cs
NikolayXHD: 11

GitUI/Theming/Renderers/ThemeProperty.cs
NikolayXHD: These are WinApi theming property identifiers, they are used in GetThemeColor.
Searchable by `public override int GetThemeColor`. Whenever possible I used a subset available in .NET which is `System.Windows.Forms.VisualStyles.ColorProperty`. Value `204 = Color` was missing though.

GitUI/Theming/ThemeController.cs
NikolayXHD: ok

GitUI/Theming/ThemeController.cs
NikolayXHD: renamed to `SetColors`

GitUI/Theming/ThemeController.cs
NikolayXHD: renamed to `GetColors`

GitUI/Theming/ThemeController.cs
NikolayXHD: ok

NikolayXHD: because it provides the same interface to get Application and System colors, to be used in `ColorHelper`.

GitUI/Theming/Renderers/ThemeProperty.cs
RussKie: Ah, found it - https://docs.microsoft.com/en-us/windows/win32/controls/property-typedefs
Though looking at the docs, these aren't all for colours...

GitUI/Theming/ThemeController.cs
RussKie: Could you please add an xml-doc explaining what the API does, it would help with comprehension.
Thank you

GitUI/Theming/ThemeController.cs
RussKie: I'm finding it confusing and prone to a number of problems including maintainability and concurrency... 😕 


We are mixing state and behaviours and this, in my experience, often leads to tighter coupling, lower cohesion, inability to inject/substitute behaviours, serialise/persist state, and threading issues.
When I think of a theme, I think of a set of properties (e.g. colours, images, etc.) that are somehow persisted. When I think of a theme manager/controller I can imagine a behavioural system that loads/saves themes and otherwise allows interacting with a theme.

By separating state and behaviour we will be able to alter theme definitions and support multiple versions, and/or substitute how themes are loaded or persisted (e.g. in a flat or a json file, read Widnows theme definitions etc) with minimal overhead.

Granted in this specific case we are dealing with themes in two points - at the app load and the settings dialog, so the risk of concurrency is likely low, we are increasing our tech debt surface and may be giving wrong ideas to other contributors.

GitUI/Theming/ThemeController.cs
NikolayXHD: ok

GitUI/Theming/ThemeController.cs
NikolayXHD: documented `ThemeManager` and it's base class `Theme`

GitUI/Theming/ThemeController.cs
NikolayXHD: `ThemeManager` stores in-memory .Net `SystemColors` colors overrides to be accessed by `Win32ThemeHooks` hook overrides. Notice that `Win32ThemeHooks` only sees base class `Theme`'s interface, which is just a storage for color values definitions.

Instead of inheriting from `Theme`, `ThemeManager` could be refactored to include a read-write-able instance of `Theme`, which would be a sort of decoupling, though in my view purely cosmetic and not useful. Still if you disagree or have different perspective, I am 100% ok to change.

In contrast with `SystemColors`, GitExtensions app-specific color overrides are stored and persisted by already present `AppSettings` mechanism, from which the rest of application reads `AppColors` values. Changing it looks pretty expensive.

Currently `ThemeManager` provides a uniform interface to read and modify both `SystemColors` and `AppColors` which is helpful in `ThemeEditor`.

GitUI/Theming/ThemeEditor.cs
NikolayXHD: There is `FormThemeEditor` which presents UI to save / load color scheme or modify individual colors. `ThemeEditor` is the part which does not depend on the `Form` itself.

It is used to handle `FormThemeEditor` button clicks, and besides that, it can be used to access `FormThemeEditor`'s functionality without creating a form. For instance, when loading current theme at application startup.

Maybe we should consider removing direct access from `FormThemeEditor` to `ThemeManager`
by proxying it through `ThemeEditor` to emphasice the role of `ThemeEditor` as handler of all 'business logic' of `FormThemeEditor`

RussKie: > It is used to handle `FormThemeEditor` button clicks, and besides that, it can be used to access `FormThemeEditor`'s functionality without creating a form

We are calling such classes "controllers", e.g. `FormBrowseController`, `FormFileHistoryController`, etc.

> Maybe we should consider removing direct access from `FormThemeEditor` to `ThemeManager`
> by proxying it through `ThemeEditor` to emphasize the role of `ThemeEditor` as handler of all 'business logic' of `FormThemeEditor`

That's a good idea!

GitUI/Theming/ThemeController.cs
RussKie: Ok let's keep it as is.

RussKie:
As part of the overall "package", could you please capture and describe the logic and the flow in a wiki. And provide guidelines on how to build a themeable UI - e.g.
* what colours are to use, 
* how to pick sensible colours, if non system colours are used,
* how to deal with images, etc.

RussKie:
@gerhardol are you happy with the current state?

RussKie:
@NikolayXHD I think we're getting close to merging this PR in, and now it is the time to think about the presentation - do you expect the PR to be squashed, do you want to separate and keep certain commits and squash the rest...

NikolayXHD:
Squashed except the commit of adding .netstandard references

gerhardol:
> @gerhardol are you happy with the current state?

Yes, the dark theme is not just usable, it is very well done, also compared to commercial programs.
The wiki page I put up with visual issues is very empty...
I have not used the default theme much though.
I see the primary goal of this PR as the dark theme. Everything else is a bonus.

Especially selections are hard with dark themes. One minor comment is that selections are not very visible if the control is not in focus.

image

image

GitUI/Theming/ThemeEditor.cs
NikolayXHD: done

NikolayXHD:
> Yes, the dark theme is not just usable, it is very well done, also compared to commercial programs.

🌞🌞🌞 Awww, sweetest music for my ears, thank you @gerhardol

> Especially selections are hard with dark themes. One minor comment is that selections are not very visible if the control is not in focus.

fixed

RussKie:
@gerhardol approve?

GitUI/Theming/FormThemeEditor.cs
```diff
@@ -0,0 +1,246 @@
...
+using System.Windows.Forms;
+using GitExtUtils.GitUI;
+using GitExtUtils.GitUI.Theming;
+
+namespace GitUI.Theming
+{
+    public class FormThemeEditor : GitExtensionsForm
+    {
+        private const string Title = "Theme editor";
+        private const string HintOnResettingColor = "middle-click to reset";
```
RussKie: If these are shown to a user, they need to be translated

GitUI/Theming/FormThemeEditor.cs
```diff
@@ -0,0 +1,246 @@
...
+                switch (te.Button)
+                {
+                    case MouseButtons.Left:
+                        var dialog = new ColorDialog();
+                        dialog.Color = ctrl.BackColor;
+                        var result = dialog.ShowDialog();
+                        if (result != DialogResult.OK)
+                        {
+                            break;
+                        }
```
RussKie: The `dialog` must be disposed of

GitUI/Theming/FormThemeEditorController.cs
```diff
@@ -0,0 +1,233 @@
...
+using System.Windows.Forms;
+using GitCommands;
+using GitExtUtils.GitUI.Theming;
+
+namespace GitUI.Theming
+{
+    public class FormThemeEditorController
+    {
+        private const string Subdirectory = "Themes";
+        public const string Extension = ".colors";
```
RussKie: Does it have to be `public`? If yes, can it be `static readonly`?

GitUI/Theming/FormThemeEditorController.cs
```diff
@@ -0,0 +1,233 @@
...
+            if (!_persistence.SaveToFile(appColors, sysColors, selectedFile, quiet: false))
+            {
+                return;
+            }
+
+            OnThemeChanged(false, selectedFile);
+
+            bool SaveDialog(out string filename)
+            {
+                var dlg = new SaveFileDialog
```
RussKie: Must be disposed of

GitUI/Theming/FormThemeEditorController.cs
```diff
@@ -0,0 +1,233 @@
...
+            if (TrySelectFile(out string selectedFile) && ApplyThemeFile(selectedFile))
+            {
+                OnThemeChanged(true, selectedFile);
+            }
+
+            bool TrySelectFile(out string result)
+            {
+                result = null;
+
+                var dlg = new OpenFileDialog
```
RussKie: ditto

RussKie:
@NikolayXHD I think it is safe to say your awesome PR is ready (subject to @gerhardol's approval).
You can make up the PR (i.e. squash/reword/update commit messages) etc.

gerhardol:
What is missing to merge is somehow outside this PR: installer 32/64 bit aware
However, this PR is done in such way that 32-bit installations should work anyway, so it should be fine.
I will try this tonight or tomorrow

> fixed


Again , a great tweak. So much care of details in this PR, these types of selections are very delicate to get right with dark themes.

image

RussKie:
We will tackle the installer separately.

pmiossec:
>Installer need changes, separate from this PR

Probably worth to create the issue now to be able to add it to the next milestone.

RussKie:
@mstv if we merge, will this create a significant issue for #7394?

mstv:
I guess, #7394 is quite orthogonal to this one. I will resolve eventual conflicts.

I had a first glance on a build of this PR about a week ago, especially with the default theme. I like it, @NikolayXHD! 👍

RussKie:
Woohoo!

RussKie:
This looks a bit odd:
image
image

I'm speculating but we may need to apply scaling properly

NikolayXHD:
> I'm speculating but we may need to apply scaling properly

Can't see what's odd in the screenshot

gerhardol:
> ah, the warning icon, ok.

RussKie thought the larger text was a scaling problem. However, that is how it was designed to look like also on normal scaling.
I believe this was changed when someone failed to see the existing warning. Before restarting, the app looks really strange....

I suggest to keep a normal size font.

NikolayXHD:
> I believe this was changed when someone failed to see the existing warning. Before restarting, the app looks really strange....
> 
> I suggest to keep a normal size font.

Yes, and also do not apply the theme before restart, to prevent consistent appearance.
To help create new themes the modifications could be visible e.g. only while theme customization window is open.

NikolayXHD:
> I checked out the latest master and try to set the dark theme, restarted the app and... it didn't work. I have restarted the app twice, reapplied the theme twice to no avail.
No idea what isn't working, need to look further into it.

Doesn't reproduce to me.

I guess something went wrong with hooking (e.g. dll not loaded silently) in which case there should be `'Failed to install Win32 theming hooks {ex}'` message in trace output. If nont, I would start Debugging with breaks on any exception, that should pinpoint the problem easily.

gerhardol:
> No idea what isn't working, need to look further into it.

Did you build with C++?

RussKie:
I believe I did :)
Anyway I'll test it again when I get around to it

mstv:
Git did the job of merging this into #7394. Only `gitextensions.sln` had minor conflicts.

Though do we really need to reference a specific Windows SDK version again?

`Error MSB8036: The Windows SDK version 10.0.18362.0 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution". (1, 1)`

RussKie:
> Though do we really need to reference a specific Windows SDK version again?

No, you need to set it to the same as version as other C++ projects: `<WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>`

I plan to convert to `<WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>` and to use VS2019 on the AppVeyor, once the conversion to SDK-style is done.

mstv:
The different SDK version was introduced by https://github.com/gitextensions/EasyHook/commit/0292388a8f7d51e8468b70feeee0fbd958d62bb1.

RussKie:
We need to set it to the same version as other C++ projects

RussKie:
I fixed the EasyHook targeting in 2d9d428e7582b54738efdfb4f0b29f42db48f5b8

RussKie:
I'm publishing a portable build now again - https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30847034/artifacts

Attempting to start an app I get the following error:
image

The error is not fatal, it complains about absent `Theme\current.colors` file, which gets created as part of the launch sequence.

Any thoughts what we need to improve the UX?

hani77979:
ت

RussKie:
Observed another artifact - disabled combos are rendered incorrectly:
image

RussKie:
image

RussKie:
For the context, some of this incorrect rendering started after I changed Windows scale factor (e.g. 100% -> 150%)

NikolayXHD:
> disabled combos are rendered incorrectly

This one is a [known problem](https://github.com/gitextensions/gitextensions/wiki/Dark-Theme#bright-disabled-combobox)

> image

> For the context, some of this incorrect rendering started after I changed Windows scale factor (e.g. 100% -> 150%)

This is great! The example with changing scale at the very least rules out incorrectly applying theme settings after installing / upgrading the application. I observed bright buttons / combo-boxes a couple of times on other developers machines. We found that change theme + restart fixes the issue, but were not able to reproduce it.

RussKie:
Heh this is my life now dealing with scaling - changing dpi 30 times an
hour... 😔

NikolayXHD:
@RussKie 

> For the context, some of this incorrect rendering started after I changed Windows scale factor (e.g. 100% -> 150%)

I can reproduce visual artifacts in dark theme after changing DPI, however if I restart Git Extensions the artifacts are gone.

I find having glitches between DPI change and app restart rather normal. I would also expect glitches after changing Windows theme. For example `Win32ThemeHooks` class relies on theme data handles to remain constant which should not be true if DPI or Windows theme changes.

Issue #7182, opened 2019-09-21
[NBug] Unable to parse git command output.


vfofanov:
## Current behaviour

Exception raises only on specific repo when I open it or make pull operation

## Expected behaviour

No exception

## Steps to reproduce

I think problem with remote branches' names on Russian
See attachment picture

## Error Details
```
System.Exception: Unable to parse git command output.
   at GitCommands.GitModule.<GetRemotesAsync>g__ParseRemotes|153_0(IEnumerable`1 lines)
   at GitCommands.GitModule.<GetRemotesAsync>d__153.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.BranchTreePanel.RepoObjectsTree.RemoteBranchTree.<LoadNodesAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.BranchTreePanel.RepoObjectsTree.Tree.<ReloadNodesAsync>d__27.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.BranchTreePanel.RepoObjectsTree.Tree.<<UICommands_PostRepositoryChanged>b__6_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.ThreadHelper.<>c__DisplayClass13_0.<<FileAndForget>b__0>d.MoveNext()
```

## Additional information
Just open repo anf got this error. I suppose the problem is with comments in Russian
I got this exception few last releases and workaround just close Error Report window

## Environment
- Git Extensions 3.2.1.6628
[...]

PR #7214, opened 2019-09-29
Allow multiple push url for git-remote

### LINKED ISSUE #6922: [NBug] Remote URLs should appear in pairs.
## Current behaviour
Application crashes when it tries to open a repository with two "url" fields in [remote "origin"] section. This is NOT the same as issue #6562 and family.

## Expected behaviour

Application should handle two "url" fields

## Steps to reproduce

Open a repository's .git/config file and add two "url" fields in [remote "origin"] section. Then open that repository using git extensions.

## Error Details
```
System.Exception: Remote URLs should appear in pairs.
   at GitCommands.GitModule.<GetRemotesAsync>g__ParseRemotes|154_0(IEnumerable`1 lines)
   at GitCommands.GitModule.<GetRemotesAsync>d__154.MoveNext()
   at GitUI.BranchTreePanel.RepoObjectsTree.RemoteBranchTree.<LoadNodesAsync>d__4.MoveNext()
   at GitUI.BranchTreePanel.RepoObjectsTree.Tree.<ReloadNodesAsync>d__27.MoveNext()
   at GitUI.BranchTreePanel.RepoObjectsTree.<>c__DisplayClass69_0.<<AddTree>b__1>d.MoveNext()
```

## Additional information
This is NOT the same as issue #6562. To reproduce, open a repository's .git/config file and add two "url" fields in [remote "origin"] section

## Environment
- Git Extensions 3.1.1.6049
- Git 2.22.0.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #6922

## Proposed changes
Allow multiple push url in git-remote
The pushurl are not really used (can be used to display the GitHub/BitBucket icon in sidepanel, updated that logic to only use fetchurl).

## Test methodology
Manual (test is updated)
See issue with how to reproduce (like add a repo on gitlab).

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2136,53 +2136,61 @@ namespace GitCommands
             {
                 var remotes = new List<Remote>();
 
-                /*
-                 * $ git remote -v
-                 * origin  git@github.com:drewnoakes/gitextensions.git (fetch)
-                 * origin  git@github.com:drewnoakes/gitextensions.git (push)
-                 * upstream        git@github.com:gitextensions/gitextensions.git (fetch)
-                 * upstream        git@github.com:gitextensions/gitextensions.git (push)
-                 *
-                 * That's a tab character between the first two records. There's a space before (fetch/push).
-                 */
+                // See tests for explanation of the format
 
                 using (var enumerator = lines.GetEnumerator())
                 {
                     while (enumerator.MoveNext())
                     {
-                        var fetchLine = enumerator.Current;
-
-                        // An invalid module is not an error; we simply return an empty list of remotes
-                        if (fetchLine.IndexOf("not a git repository", StringComparison.OrdinalIgnoreCase) >= 0)
+                        var remoteLine = enumerator.Current;
+                        if (remoteLine.IndexOf("not a git repository", StringComparison.OrdinalIgnoreCase) >= 0)
                         {
+                            // An invalid module is not an error; we simply return an empty list of remotes
                             return remotes;
+                        }
+
+                        var remoteMatch = _remoteVerboseLineRegex.Match(remoteLine);
+                        if (!remoteMatch.Success
+                            || (remoteMatch.Groups["direction"].Value != "fetch"
+                               && remoteMatch.Groups["direction"].Value != "push"))
+                        {
+                            // Ignore malformed and unknown entries
+                            continue;
+                        }
+
+                        var name = remoteMatch.Groups["name"].Value;
+                        var remoteUrl = remoteMatch.Groups["url"].Value;
+                        if (remoteMatch.Groups["direction"].Value == "push")
+                        {
+                            if (remotes.Count <= 0 || name != remotes[remotes.Count - 1].Name)
+                            {
+                                throw new Exception("Unable to update remote pushurl for command output: " + remoteLine);
+                            }
+
+                            remotes[remotes.Count - 1].PushUrls.Add(remoteUrl);
+                            continue;
                         }
 
                         if (!enumerator.MoveNext())
                         {
-                            throw new Exception("Remote URLs should appear in pairs.");
+                            throw new Exception("Remote URLs should appear in pairs, no pushurl for fetch: " + remoteLine);
                         }
 
                         var pushLine = enumerator.Current;
-
-                        var fetchMatch = _remoteVerboseLineRegex.Match(fetchLine);
                         var pushMatch = _remoteVerboseLineRegex.Match(pushLine);
-
-                        if (!fetchMatch.Success || !pushMatch.Success || fetchMatch.Groups["direction"].Value != "fetch" || pushMatch.Groups["direction"].Value != "push")
+                        if (!pushMatch.Success || pushMatch.Groups["direction"].Value != "push")
                         {
-                            throw new Exception("Unable to parse git command output.");
+                            throw new Exception("Unable to parse git remote push URL line: " + pushLine);
                         }
 
-                        var name = fetchMatch.Groups["name"].Value;
-                        var fetchUrl = fetchMatch.Groups["url"].Value;
                         var pushUrl = pushMatch.Groups["url"].Value;
-
                         if (name != pushMatch.Groups["name"].Value)
                         {
-                            throw new Exception("Fetch and push remote names must match.");
+                            throw new Exception("Fetch and push remote names must match: " +
+                                remoteLine + ", " + pushLine);
                         }
 
-                        remotes.Add(new Remote(name, pushUrl, fetchUrl));
+                        remotes.Add(new Remote(name, remoteUrl, pushUrl));
                     }
                 }
 
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
@@ -313,20 +313,20 @@ namespace GitUI.BranchTreePanel
             {
                 base.ApplyStyle();
 
-                TreeViewNode.ToolTipText = _remote.FetchUrl != _remote.PushUrl
-                    ? $"Push: {_remote.PushUrl}\nFetch: {_remote.FetchUrl}"
+                TreeViewNode.ToolTipText = _remote.FetchUrl != _remote.PushUrls[0]
+                    ? $"Push: {_remote.PushUrls[0]}\nFetch: {_remote.FetchUrl}"
                     : _remote.FetchUrl;
 
                 string imageKey;
-                if (_remote.PushUrl.Contains("github.com") || _remote.FetchUrl.Contains("github.com"))
+                if (_remote.FetchUrl.Contains("github.com"))
                 {
                     imageKey = nameof(Images.GitHub);
                 }
-                else if (_remote.PushUrl.Contains("bitbucket.") || _remote.FetchUrl.Contains("bitbucket."))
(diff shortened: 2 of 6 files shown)


RussKie:
@drewnoakes that's your baby, could you please review.

gerhardol:
After the following comment: https://github.com/gitextensions/gitextensions/issues/7182#issuecomment-538058914

I changed so illegal information is just ignored, no longer creates exceptions. We should not cause exceptions if the user has garbage data if it does not cause other functional issues.
Exceptions remains for "protocol" issues, that should not happen also with a bad configuration. (I consider changing this to avoid further exceptions for remotes since it has no negative effect.)

GitCommands/Git/GitModule.cs
```diff
@@ -2136,50 +2136,60 @@ IReadOnlyList<Remote> ParseRemotes(IEnumerable<string> lines)
...
+                        var name = fetchMatch.Groups["name"].Value;
+                        var fetchUrl = fetchMatch.Groups["url"].Value;
+                        if (fetchMatch.Groups["direction"].Value == "push")
+                        {
+                            if (remotes.Count <= 0 || name != remotes[remotes.Count - 1].Name)
+                            {
+                                throw new Exception("Unable to update remote pushurl for command output: " + fetchLine);
+                            }
+
+                            var updatedPushUrl = fetchUrl;
```
pmiossec: I don't understand what we do here.

Do we replace the previously found push url because we support only one?

If, yes, I tend to think that it is the first one the more important.
So we could keep it and fill a collection with additional push url. That we don't handle in the rest of the application for the moment but will let us display text like "+ 1 other remote" somewhere (push form?)

That would looks like `remotes[remotes.Count - 1].AddOtherPushUrl(updatedPushUrl)`...

Please add at least a comment.

GitCommands/Git/GitModule.cs
```diff
@@ -2136,50 +2136,60 @@ IReadOnlyList<Remote> ParseRemotes(IEnumerable<string> lines)
...
-                 *
-                 * That's a tab character between the first two records. There's a space before (fetch/push).
-                 */
+                // See tests for explanation of the format
 
                 using (var enumerator = lines.GetEnumerator())
                 {
                     while (enumerator.MoveNext())
                     {
                         var fetchLine = enumerator.Current;
```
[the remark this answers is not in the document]
gerhardol: ok - leftover from existing implementation (use remoteLine)

GitCommands/Git/GitModule.cs
gerhardol: I did not bother with making PushUrl a List (see description)
The last url is what you get when requesting with git-config (for instance in FormRemote), this aligns with what is displayed in FormRemote
Will add a comment, unless other changes are requested

RussKie: Are we talking about this?
image

If so, it can be quite useful.

I think we need to make it a list and read everything a user may have configured.
We'll need to update the UI to match as well (can be a separate PR)

gerhardol: > Are we talking about this?

Yes, but this remote parsing is not used (but config parsing)
Adding multiple push url in the GUI is definitely an enhancement in a separate PR (nothing I am interested in), but I could make PushUrl a list if requested here.

pmiossec: That's a code that is not easy to understand, so I think that we will not change it often.

That's why I think we could do it now as we are in it. And I even think that it will made the code clearer.

But that is just my opinion and maybe a comment is enough. I don't what to force someone to do something...

gerhardol: > But that is just my opinion and maybe a comment is enough. I don't what to force someone to do something...

When implementing, you have to draw the line somewhere, everything cannot be polished so it shines. In this case I really stopped when detecting a second pushurl, but then I found that the config extraction used the last so I changed the implementation. 
If you as a reviewer was not happy about it, of course you can request changes, this is reasonable. 

Also changed order so fetchurl is before pushurl(s)
I will squash when approved

RussKie: Let's enumerate all configured remotes (fetch and as many pushes) as user may have configured.
The UI is still limited but we can deal with this later.

gerhardol: Brief check of usage of pushurl:
Most usage of remotes is using GitCommands.Remotes.ConfigFileRemote that reads/write config SettingKeyString.RemotePushUrl, a single pushurl (last in the list)

If the multiple pushurl is a desired feature, Remote cannot replace ConfigFileRemote  directly as ConfigFileRemote  handles disabled remotes and PuttyKeys too. So disabling a remote will not display the extra push urls (but the urls will be displayed when enabling again).

Usages of ConfigFileRemote :
ExternalLinkRevisionParser.cs
GitUI\BranchTreePanel\RepoObjectsTree.Nodes.Remotes.cs
FormPull.cs
FormPush.cs (OK)
FormRemotes.cs
CommitInfo.cs (for ExternalLinkRevisionParser)


 Most is filled in by reading from config

Explicit use of pushurl from config also in FormClone: Should be OK

GitUI\CommandsDialogs\SettingsDialog\Pages\BuildServerIntegrationSettingsPage.cs
prioritizes a (single) push url over the fetchurl, which should be incorrect

Some usages like GitHub3Plugin.cs lists remote names, then the remote url

So a second update is required before multiple push url can be displayed in FormRemote

I have pushed some comments

RussKie: Thank you for the analysis.

I backtrack on my requests, because it is clearly not going to be beneficial - other parts of the app won't be able to support or utilise a list of push remotes.
So I leave it up to you to make the best decision here.

Given that you've done the analysis and (hopefully) understand the breadth of the problem, do you think you could raise an issue (or more) describing what you think needs to be done to support multiple push remotes and provide necessary UI/UX for this?

gerhardol: > could raise an issue

That is the plan, I just put down the analysis here so I do not forget about it.
I do not find the feature to be important in an GUI, but the inconsistencies in the data structures is itching.

Unless someone has better ideas, I will review the comments in ConfigFileRemote  when adding the issue, before merging.

UnitTests/GitCommandsTests/GitModuleTests.cs
```diff
@@ -467,34 +477,53 @@ public void GetRemotes_should_parse_correctly_configured_remotes()
                     "asymmetrical\thttps://github.com/gitextensions/fetch.git (fetch)",
                     "asymmetrical\thttps://github.com/gitextensions/push.git (push)",
                     "with-space\tc:\\Bare Repo (fetch)",
-                    "with-space\tc:\\Bare Repo (push)"
+                    "with-space\tc:\\Bare Repo (push)",
+                    "multi\tgit@github.com:drewnoakes/gitextensions.git (fetch)",
```
drewnoakes:
```suggestion

                    // A remote may have multiple push URLs, but only a single fetch URL
                    "multi\tgit@github.com:drewnoakes/gitextensions.git (fetch)",
```

Plugins/GitUIPluginInterfaces/Remote.cs
```diff
@@ -1,18 +1,21 @@
...
+using System.Collections.Generic;
 
 namespace GitUIPluginInterfaces
 {
     public readonly struct Remote
     {
         public string Name { get; }
-        public string PushUrl { get; }
         public string FetchUrl { get; }
+        public List<string> PushUrl { get; }
```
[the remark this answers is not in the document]
gerhardol: yes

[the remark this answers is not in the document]
drewnoakes: Sorry, I don't think I was clear. I meant the property name should be pluralised to `PushUrls`.

GitCommands/Git/GitModule.cs
gerhardol: See related issue and PR.
RepoObjectsTree.Nodes.Remotes.cs uses both Remote.cs and ConfigFileRemote

gerhardol: @pmiossec 
Can you please review again?

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -313,20 +313,20 @@ protected override void ApplyStyle()
             {
                 base.ApplyStyle();
 
-                TreeViewNode.ToolTipText = _remote.FetchUrl != _remote.PushUrl
-                    ? $"Push: {_remote.PushUrl}\nFetch: {_remote.FetchUrl}"
+                TreeViewNode.ToolTipText = _remote.FetchUrl != _remote.PushUrls[0]
```
pmiossec: Do you think it brings more meaning to introduce a property 'DefaultPushUrl' that do this job to select the first element (and also maybe a control that it exists)?

gerhardol: This is the only situation where push url is currently used. If we are to change something it is probably to show all push url instead.

gerhardol: I will propose changing to show all PushUrl in #7256 
Will just rebase before merge

gerhardol:
@msftbot merge in 24h

gerhardol:
@msftbot merge after 24h

gerhardol:
@msftbot delay merging this pull request for 24 hours

Issue #3280, opened 2016-07-26
Unstaging many files from Commit dialog freezes UI for a long time


shift-evgeny:
If I open the Commit dialog with hundreds of files staged and try to unstaging most of them (but not all of them) the UI appears to be frozen. I can see in Process Hacker that Git Extensions is running "git.exe reset HEAD -- some/file" over and over for each file. This is a poor user experience. With enough files I would have no choice but to kill the Git Extensions process.

If there is no way to reset multiple files quickly Git Extensions should not freeze while this runs and should allow the process to be cancelled at any time.

vbjay:
We could use something like the process dialog.  If the count is greater
than X then show the dialog with lines saying <file path> unstaged.

Radrik5:
Usually it's much faster to run single git process with a set of files then to start separate git processes for each file. But there is a command line length limit, so in general we cannot put all the selected files into a single command.

vbjay:
Why I suggested the process dialog.  After each unstage we add a line
saying x was unstaged.

jbialobr:
@vbjay the problem is not a lack of progress, but poor performance caused by executing a git command for each file.

@Radrik5 We could split the file list into smaller hunks and execute fewer git processes. Other option is to use Libgit2Sharp library for unstaging files.

vbjay:
@jbialobr <https://github.com/jbialobr> I was thinking if it will take a
while run the commands in a background thread and show the status if the
file count is above a minimum value.  This would eliminate the freezing and
let the user know it is processing.

shift-evgeny:
>the problem is not a lack of progress, but poor performance caused by executing a git command for each file

I think both of these are problems. Obviously if Git Extensions can do this much faster - it should. But there could be thousands of files instead of hundreds. Unless the command can *always* execute within a second or so the UI will still appear to freeze.

jbialobr:
There is a progres bar on the top of the commit dialog.

shift-evgeny:
That's true. So perhaps if it freezes for up to 5-10 seconds or so it's not so bad. But any more than that and there should still be an option to cancel rather than having to kill GitExtensions.exe.

nsartor:
In the case of pressing the "unstage all" button, the app still calls git on each file 1 at a time. Could it just call "git reset" in this case instead?

PR #7218, opened 2019-09-30
Fix #6593: GitExtension Exception resetting 1500 files in Diff tab.

### LINKED ISSUE #6593: GitExtension Exception resetting 1500 files in Diff tab
## Current behaviour
GitExtensions crashes

## Expected behaviour
No crash

## Steps to reproduce
Have a big change in working directory (1500 files)
Go to Diff tab
Select all files
Press R (Reset to index)

## Error Details
```
System.ComponentModel.Win32Exception (0x80004005): The filename or extension is too long
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at GitCommands.Executable.ProcessWrapper..ctor(String fileName, String arguments, String workDir, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding)
   at GitCommands.Executable.Start(ArgumentString arguments, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding)
   at GitCommands.ExecutableExtensions.<RunCommandAsync>d__5.MoveNext()
   at Microsoft.VisualStudio.Threading.JoinableTask`1.CompleteOnCurrentThread()
   at GitCommands.ExecutableExtensions.RunCommand(IExecutable executable, ArgumentString arguments, Byte[] input, Boolean createWindow)
   at GitCommands.GitModule.CheckoutFiles(IReadOnlyList`1 files, ObjectId revision, Boolean force)
   at GitUI.CommandsDialogs.RevisionDiffControl.ResetSelectedItemsTo(Boolean actsAsChild)
   at System.Windows.Forms.ToolStripItem.ProcessDialogKey(Keys keyData)
   at System.Windows.Forms.ToolStrip.ProcessDialogKey(Keys keyData)
```

## Environment
- Git Extensions 3.1.0.5877
- Git 2.21.0.windows.1
- .NET Framework 4.8.3801.0
- DPI 96dpi (no scaling)

hieuxlu:
Work around for Windows command line length 32767 characters limitation

Fix #6593: GitExtension Exception resetting 1500 files in Diff tab.

## Proposed changes

- Work around for Windows command line length 32767 characters limitation
- Add GitArgumentBuilderExtensions.BuildBatchArguments to wrap long commands in to multiple child commands
- Add RunBatchCommand to run commands in batch 

### Before
image

### After

image

## Test methodology 

- Add GitArgumentBuilderTest fixture
- Manual test

## Test environment(s) 

- GIT: 3.2
- Windows 10 64bit

----

### DIFF
--- a/GitCommands/ArgumentBuilderExtensions.cs
+++ b/GitCommands/ArgumentBuilderExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 using GitCommands.Git;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
@@ -299,6 +300,69 @@ namespace GitCommands
             {
                 builder.Add(objectId);
             }
+        }
+
+        /// <summary>
+        /// Split arguments exceeding max length into multiple batches.
+        /// Windows by default limit arguments length less than 32767 <see cref="short.MaxValue"/>.
+        /// Implementation by <see cref="System.Diagnostics.Process"/> will have file path included in command line arguments,
+        /// as well as added quotation and space characters, so we need base length to account for all these added characters
+        /// <see href="https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,1944"/>
+        /// </summary>
+        /// <param name="builder">Argument builder instance.</param>
+        /// <param name="arguments">Arguments.</param>
+        /// <param name="baseLength">Base executable file and command line length.</param>
+        /// <param name="maxLength">Command line max length. Default is 32767 - 1 on Windows.</param>
+        /// <returns>Array of batch arguments split by max length.</returns>
+        public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder, IEnumerable<string> arguments, int baseLength, int maxLength = short.MaxValue)
+        {
+            var baseArgument = builder.ToString();
+            if (baseLength + baseArgument.Length >= maxLength)
+            {
+                throw new ArgumentException($"Git base command \"{baseArgument}\" always reached max length of {maxLength} characters.", nameof(baseArgument));
+            }
+
+            // Clone command as argument builder
+            var batches = new List<ArgumentString>();
+            var currentArgumentLength = baseArgument.Length;
+            var lastBatchBuilder = arguments.Aggregate(builder, (currentBatchBuilder, argument) =>
+            {
+                // 1: ' ' space character length will be added
+                // When enumeration is finished, no need to add ' ' to length calculcation
+                if (baseLength + currentArgumentLength + 1 + argument.Length >= maxLength)
+                {
+                    // Handle abnormal case when base command and a single argument exceed max length
+                    if (currentArgumentLength == baseArgument.Length)
+                    {
+                        throw new ArgumentException($"Git command \"{currentBatchBuilder}\" always exceeded max length of {maxLength} characters.", nameof(arguments));
+                    }
+
+                    // Finish current command line
+                    batches.Add(currentBatchBuilder);
+
+                    // Return new argument builder
+                    currentArgumentLength = baseArgument.Length + 1 + argument.Length;
+                    return new ArgumentBuilder() { baseArgument, argument };
+                }
+
+                currentBatchBuilder.Add(argument);
+                currentArgumentLength += argument.Length + 1;
+                return currentBatchBuilder;
+            });
+
+            // Handle rare case when last argument length exceed max length
+            if (baseLength + currentArgumentLength >= maxLength)
+            {
+                throw new ArgumentException($"Git command \"{lastBatchBuilder}\" always exceeded max length of {maxLength} characters.", nameof(arguments));
+            }
+
+            // Add last commandline batch
+            if (!lastBatchBuilder.IsEmpty)
+            {
+                batches.Add(lastBatchBuilder.ToString());
+            }
+
+            return batches.ToArray();
         }
     }
 }
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -142,6 +142,30 @@ namespace GitCommands
         }
 
         /// <summary>
+        /// Launches a process for the executable per batch item, and returns <c>true</c> if all process exit codes were zero.
+        /// </summary>
+        /// <remarks>
+        /// This method uses <see cref="RunCommand"/> to execute multiple commands in batch, used in accordance with
+        /// <see cref="ArgumentBuilderExtensions.BuildBatchArguments(ArgumentBuilder, IEnumerable{string}, int, int)"/>
+        /// to work around Windows command line length 32767 character limitation
+        /// <see href="https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessa"/>.
+        /// </remarks>
+        /// <param name="executable">The executable from which to launch processes.</param>
+        /// <param name="batchArguments">The array of batch arguments to pass to the executable</param>
+        /// <param name="input">Bytes to be written to each process's standard input stream, or <c>null</c> if no input is required.</param>
+        /// <param name="createWindow">A flag indicating whether a console window should be created and bound to each process.</param>
+        /// <returns><c>true</c> if all process exit codes were zero, otherwise <c>false</c>.</returns>
+        [MustUseReturnValue("Callers should verify that " + nameof(RunBatchCommand) + " returned true")]
+        public static bool RunBatchCommand(
+            this IExecutable executable,
+            ArgumentString[] batchArguments,
+            byte[] input = null,
+            bool createWindow = false)
+        {
+            return batchArguments.Aggregate(true, (result, arguments) => executable.RunCommand(arguments, input, createWindow) && result);
+        }
+
+        /// <summary>
         /// Launches a process for the executable and returns <c>true</c> if its exit code is zero.
         /// </summary>
         /// <param name="executable">The executable from which to launch a process.</param>
--- a/UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
+++ b/UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
@@ -1,6 +1,14 @@
 ﻿using System;
+using System.ComponentModel;
+using System.IO;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 using CommonTestUtils;
 using GitCommands;
+using GitCommands.Settings;
+using GitUIPluginInterfaces;
 using NUnit.Framework;
 
 namespace GitCommandsTests.Git
@@ -9,11 +17,27 @@ namespace GitCommandsTests.Git
     public sealed class ExecutableExtensionsTests
     {
         private MockExecutable _executable;
+        private Executable _gitExecutable;
+        private string _appPath;
 
         [SetUp]
         public void SetUp()
         {
             _executable = new MockExecutable();
+
+            // Work around: When running unittest, Application.UserAppDataPath always points to
+            // %APPDATA%Roaming\Microsoft Corporation\Microsoft.TestHost.x86
+            // We need to correct it to %APPDATA%\GitExtensions\GitExtensions for v3 at least
+            var userAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var settingPath = Path.Combine(userAppDataPath, "GitExtensions\\GitExtensions\\GitExtensions.settings");
+            var settingContainer = new RepoDistSettings(null, GitExtSettingsCache.FromCache(settingPath), SettingLevel.Unknown);
+            _appPath = Path.Combine(settingContainer.GetString("gitbindir", ""), "git.exe");
+
+            // Execute process in GitExtension working directory, so that git will return success exit-code
+            // git always return non-zero exit code when run git reset outside of git repository
+            // NUnit working directory always default to MS test host
+            var workingDir = Path.GetDirectoryName(Assembly.GetAssembly(typeof(ExecutableExtensionsTests)).Location);
+            _gitExecutable = new Executable(_appPath, workingDir);
         }
 
         [TearDown]
@@ -63,6 +87,55 @@ namespace GitCommandsTests.Git
             Assert.IsTrue(cache.TryGet(arguments, out var outputBytes, out var errorBytes));
             Assert.AreEqual(GitModule.SystemEncoding.GetBytes(commandOutput), outputBytes);
             Assert.IsEmpty(errorBytes);
+        }
+
+        // Process argument upper bound is actually (short.MaxValue - 1)
+        [TestCase(32766, 32766, 32767, 2, true)]
+        [TestCase(32764, 1, 32767, 1, true)]
+        public void RunBatchCommand_can_handle_max_length_arguments(int arg1Len, int arg2Len,
+            int maxLength, int argCount, bool expectedResult)
+        {
+            // 3: double quotes + ' '
+            // 9: 'reset -- '
+            // 1: ' ' added after second Add in ArgumentBuilder
+            var appLength = _appPath.Length + 3;
+            var builder = new ArgumentBuilder() { "reset --" };
+            var len = builder.ToString().Length;
+            var args = builder.BuildBatchArguments(new string[]
+            {
+                GenerateStringByLength(Math.Max(1, arg1Len - appLength - len - 1)),
+                GenerateStringByLength(Math.Max(1, arg2Len - appLength - len - 1))
+            }, appLength, maxLength);
+
+            Assert.AreEqual(argCount, args.Length);
+            Assert.AreEqual(expectedResult, _gitExecutable.RunBatchCommand(args));
+        }
+
+        [TestCase(32766 - 8, 32766 - 8, int.MaxValue)]
+        [TestCase(32766 - 9, 1, int.MaxValue)]
+        public void RunBatchCommand_throw_when_cmd_exceed_max_length(int arg1Len, int arg2Len,
+            int maxLength)
+        {
+            var args = new ArgumentBuilder() { "reset --" }
+                .BuildBatchArguments(new string[]
+                {
+                    GenerateStringByLength(Math.Max(1, arg1Len - _appPath.Length - 4)),
+                    GenerateStringByLength(Math.Max(1, arg2Len - _appPath.Length - 4))
+                }, _appPath.Length + 3, maxLength);
+
+            Assert.Throws(typeof(Win32Exception), () => _gitExecutable.RunBatchCommand(args));
+        }
+
+        private string GenerateStringByLength(int length)
+        {
+            var sb = new StringBuilder(length);
+
+            for (int i = 0; i < length; i++)
+            {
+                sb.Append((char)((i % 26) + 97));
+            }
+
+            return sb.ToString();
         }
     }
 }
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1392,14 +1392,16 @@ namespace GitCommands
                 revision = null;
             }
 
-            _gitExecutable.RunCommand(
-                new GitArgumentBuilder("checkout")
+            // Run batch arguments to work around max command line length on Windows. Fix #6593
+            // 3: double quotes + ' '
+            // See https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,1952
+            _gitExecutable.RunBatchCommand(new GitArgumentBuilder("checkout")
                 {
                     { force, "--force" },
                     revision,
-                    "--",
-                    files.Select(f => f.ToPosixPath().Quote())
-                });
+                    "--"
+                }
+                .BuildBatchArguments(files.Select(f => f.ToPosixPath().Quote()), AppSettings.GitCommand.Length + 3));
         }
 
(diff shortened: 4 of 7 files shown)


gerhardol:
Seem OK to me at first look.
You probably want to submitt the contributors.txt change separately, to avoid conflicts.

GitExtUtils/GitArgumentBuilderExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitCommands
+{
+    public static class GitArgumentBuilderExtensions
+    {
+        private const byte BASE_LENGTH = 4; // 'git '
```
mstv: I haven't digged into how git is called exactly.
Is this really necessary? I.e., doesn't "git" be part of the executable to be started?
Isn't the full path to `git.exe` prepended if specified in the settings? Does the limit apply to arguments string or to the complete command line? It depends on how `CreateProcess` is called by .NET.

UnitTests/GitExtUtils/GitArgumentBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace GitExtUtilsTests
+{
+    [TestFixture]
+    public class GitArgumentBuilderExtensionsTest
+    {
+        // 17: 'git checkout name1'
+        [TestCase("checkout", new string[] { "name1", "name2", "name3" }, 21, new string[] { "checkout name1", "checkout name2", "checkout name3" })]
+        [TestCase("checkout", new string[] { "name1", "name2", "name3" }, 22, new string[] { "checkout name1", "checkout name2", "checkout name3" })]
+        [TestCase("checkout", new string[] { "name1", "name2", "name3" }, 27, new string[] { "checkout name1 name2", "checkout name3" })]
```
mstv: I would prefer to test the exact thresholds.

GitExtUtils/GitArgumentBuilderExtensions.cs
hieuxlu: This says if we specify application name, then commandline can exclude executable module part. I'll check on how the process is called.

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw#parameters

hieuxlu: C# implementation of `System.Diagnostic.Process` always combine application name in command line argument, and always add quotation + ' ' characters, so I modified the code to handle all that.
Reference source code: https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,1952

UnitTests/GitExtUtils/GitArgumentBuilderExtensionsTest.cs
hieuxlu: Test case for exact thresholds are added in `ExecutableExtensionsTests` fixture, which will call the actual `Process`. This only performs unit testing for the builder extensions.

hieuxlu:
@gerhardol I resolved the conflicts, do I still need a separate PR for `contributor.txt`?
@mstv Please review again.

RussKie:
> do I still need a separate PR for `contributor.txt`?

It is better to do a separate PR, as it may time to fully review and test this change, if another contributor provides a sign off in the interim - it will set this PR in a conflict state again.

RussKie:
I had a brief look at the change - 👍 

To test - do I generate 1,5K files with a reasonably long name and then try to stage all at once?

hieuxlu:
Yes, clone a really long files name, 1000 files, each > 40 characters. A for loop with echo would suffice. After that, reset the file to selected commit in Diff view to test. Staging does not cause crashing, just slow. The bug is in Diff view (and probably many other places).

gerhardol:
> To test - do I generate 1,5K files with a reasonably long name and then try to stage all at once?

Just move all contents in a repo to a new folder

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
```diff
@@ -64,5 +67,38 @@ public void GetOutput_with_cache_miss()
...
+            // Execute process
+            var exe = new Executable("cmd.exe");
+            var args = new ArgumentBuilder() { "/C echo" }
+                .BuildBatchArguments(new string[]
+                {
+                    GenerateStringByLength(arg1Len),
+                    GenerateStringByLength(arg2Len)
+                }, baseLength);
+
+            exe.RunBatchCommand(args);
```
mstv: Please check the return value.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,31 @@ string ComposeOutput()
...
+        /// <param name="createWindow">A flag indicating whether a console window should be created and bound to the process.</param>
+        /// <returns><c>true</c> if the process's exit code was zero, otherwise <c>false</c>.</returns>
+        [MustUseReturnValue("Callers should verify that " + nameof(RunBatchCommand) + " returned true")]
+        public static bool RunBatchCommand(
+            this IExecutable executable,
+            ArgumentString[] batchArguments,
+            byte[] input = null,
+            bool createWindow = false)
+        {
+            // TODO: Process somehow always return non-zero exit code.
```
mstv: I'm afraid, the limit is lower. The Win32 API doc says: If the executable is left empty and added to the arguements, the length limit was `MAXPATH`.
`[TestCase(8193 - 1 - 19, 1, 10)]` is the maximum that succeeds.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,31 @@ string ComposeOutput()
...
+            return batchArguments.Aggregate(true, (result, arguments) => result && executable.RunCommand(arguments, input, createWindow));
```
mstv: `result && executable.RunCommand(...)`: In C/C++, the function would never be called due to the lazy evaluation of logical expressions and because `result` is initialized with `true`. Though C# seems to work differently: In the debugger, I can see that `RunCommand` _is_ called. @gitextensions/git-extensions-source, can we rely on it?

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: `MAXPATH` length limit in Win32 API doc is for full path of the `git.exe` executable module, whereas `lpCommandLine` max length is always `32,768 characters, including the Unicode terminating null character`. Real test leaves us at `short.MaxValue - 1 // 32766` though.
>
> `lpCommandLine`
> The maximum length of this string is 32,768 characters, including the Unicode terminating null character. If `lpApplicationName` is NULL, the module name portion of `lpCommandLine` is limited to MAX_PATH characters.
>

The TODO was left here because in the test case `cmd.exe /C echo` always return non-zero exit code. In debug, `git.exe` returns correctly, so this is for the test case only. I will investigate the test case and resolve this.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,67 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+        public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder, IEnumerable<string> arguments, int baseLength, int maxLength = short.MaxValue)
+        {
+            var baseCommand = builder.ToString();
+            if (baseCommand.Length + baseLength >= maxLength)
+            {
+                throw new ArgumentException($"Git base command \"{baseCommand}\" always reached max length of {maxLength} characters.", nameof(baseCommand));
+            }
+
+            // Clone command as argument builder
+            var tmpBuilder = new ArgumentBuilder() { baseCommand };
```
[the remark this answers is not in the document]
hieuxlu: Originally, I want to reuse the cloned builder. On second though, this is not needed though, as a new builder is always created when command length reaches limit. So we can use the input builder.

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: I'm gonna change this to a for loop, both for clarity and for batch progress callback needed in #7237.

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
hieuxlu: Yeah the TODO above should have been left here. Will do.

GitCommands/Git/ExecutableExtensions.cs
mstv: Yes, you're right. It seems to be a limitation of the length of `<command>` in `cmd /c <command>`.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,67 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+                    // Return new argument builder
+                    return new ArgumentBuilder() { baseCommand, value };
+                }
+
+                curr.Add(value);
+                return curr;
+            });
+
+            // Handle rare case when last argument length exceed max length
+            if (argBuilder.Length >= maxLength)
```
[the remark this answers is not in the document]
mstv: Extend the NUnit tests, please.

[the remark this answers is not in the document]
hieuxlu: Sorry, had to gone to work, GMT+7. 8 more test cases with baselength added.

GitCommands/ArgumentBuilderExtensions.cs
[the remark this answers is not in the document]
hieuxlu: It turned out this was correct. I want to use ArgumentBuilder.Length, to get the full command length without calling ToString(). GitArgumentBuilder.Length is not reliable. So we had to clone a new base type instance here.

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: You're right, I need to find another executable to test, `cmd.exe` command line length limitation is `8191` characters.
https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
hieuxlu: Done checking the return value, both `true` and `false`

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: I pass max length `8191` into the test case and everything works find now. About order execution of and branch, I've switch function execution before boolean var, so in any language (C/C++), it would look logical.

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: I've switched function execution before boolean var, so in any language (C/C++), it would look logical now. The for loop change would be in another PR.

hieuxlu:
@mstv Clarified all your questions. They're all great questions, thanks.

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
```diff
@@ -71,9 +71,15 @@ public void GetOutput_with_cache_miss()
...
-        public void RunBatchCommand_can_handle_max_length_arguments(int arg1Len, int arg2Len, int baseLength)
+        // 1: ' ' character
+        // cmd.exe max length is 8191, including NULL Unicode -> 8189
+        // https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation
+        [TestCase(8189 - 1 - 18, 8189 - 1 - 18, 10, 8191, true)]
+        [TestCase(8189 - 1 - 19, 1, 10, 8191, true)]
+        [TestCase(8189 - 1 - 18, 8189 - 1 - 18, 10, 32767, true)]
+        [TestCase(8189 - 1 - 19, 1, 10, 32767, true)]
+        public void RunBatchCommand_can_handle_max_length_arguments(int arg1Len, int arg2Len, int appLength,
+            int maxLength, bool expectedResult)
```
mstv: I liked the attempt to test the maximum length of a command line.
You could switch to `sh.exe -c "echo abc..."`. This executable should exist in `AppSettings.GitBinDir`.

hieuxlu: When I run NUnit on my local machine, it resolves GitExtensions.Setting to `\AppData\Roaming\Microsoft Corporation\Microsoft.TestHost.x86\` instead of  `\AppData\Roaming\GitExtensions\GitExtensions`. How can I fix this?

mstv: That's a pity. There must be other NUnit tests which access the Git executable. These must access the GE settings, too. I haven't digged into this area yet.

hieuxlu: I've searched all the codes, they've only access `AppSettings` for setting bits that are not dependent on `Application.UserAppDataPath`, such as `AppSettings.AvatarImageCachePath`.

I've tried to manually create `SettingContainers` as below, but somehow it can't load settings in the correct folder.
```CSharp
var settingPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "GitExtensions\\GitExtensions");
var settingContainer = new RepoDistSettings(null, GitExtSettingsCache.Create(settingPath), SettingLevel.Unknown);
// gitbindir is always null
var appPath = Path.Combine(settingContainer.GetString("gitbindir", ""), "sh.exe");
```

hieuxlu: My bad, I've got it working. Correct `"GitExtensions\\GitExtensions"` to `"GitExtensions\\GitExtensions\\GitExtensions.settings"`. Works fine now.

hieuxlu: @mstv `sh.exe -c echo "abcd..."` will show below error at `32752` characters, `15` char short vs docs.
> 
> error launching git: The filename or extension is too long.
> 
Does that mean internally, sh.exe is launching `git`?

On the other hand, `git.exe reset -- "abcd..."` will correctly work until `35765` character (tested by create a batch script and execute it) with `%ERRORLEVEL% = 0`. However, when run in unit test, this command always returns non-zero exit code regardless of command length.

So I'm stuck at unit testing the exact threshold at the moment. `sh.exe` won't reach it, `git.exe` always return non-zero exit code.

hieuxlu: I was able to resolve the above issue, by setting `workingDir` of `git.exe` executable to local git repository using current Assembly location. Commited changes and passed test case with exact threshold of `32766` characters

pmiossec:
A lot of work. Well done!

hieuxlu:
@mstv Done. (using GitExtensions)
@pmiossec Thought I was doing my fulltime job!

mstv:
@msftbot merge in 10 hours if approved twice

mstv:
@msftbot merge in 10 hours, if two approvals

mstv:
@msftbot merge in 10 hours, if at least 2 approvals

GitCommands/ArgumentBuilderExtensions.cs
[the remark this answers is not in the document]
RussKie: > GitArgumentBuilder.Length is not reliable. 

Could you please elaborate on why it isn't reliable?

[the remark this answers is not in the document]
hieuxlu: GitBuilderArgument.Length is not the whole command length. Its length can only be determined after a ToString() call.

hieuxlu:
@RussKie I answered your question about using a cloned builder. Is there anything else you concern about?

GitCommands/ArgumentBuilderExtensions.cs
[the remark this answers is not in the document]
RussKie: Can you please elaborate as I don't believe this claim is correct.
I have added a set of tests in https://github.com/gitextensions/gitextensions/pull/7251 and everything works as expected

RussKie:
@msftbot please ignore

RussKie:

GitCommands/ArgumentBuilderExtensions.cs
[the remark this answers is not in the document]
hieuxlu: @RussKie  The test was fore `ArgumentBuilder`. I'm saying `GitArgumentBuilder`, which is a derived class. You can try the fixture `GitArgumentBuilderTests`. `GitArgumentBuilder` can be initialized with a git command like `reset` or `config`. In that case, its `Lengths` is 0, while actual length is not.

hieuxlu:
@RussKie I removed dependency on `ArgumentBuilder.Length`, thus remove extra allocation. There is no reference to this property in the entire source code now, we can safely remove this property now.

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
```diff
@@ -9,11 +17,27 @@ namespace GitCommandsTests.Git
...
+            var userAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var settingPath = Path.Combine(userAppDataPath, "GitExtensions\\GitExtensions\\GitExtensions.settings");
+            var settingContainer = new RepoDistSettings(null, GitExtSettingsCache.FromCache(settingPath), SettingLevel.Unknown);
+            _appPath = Path.Combine(settingContainer.GetString("gitbindir", ""), "git.exe");
+
+            // Execute process in GitExtension working directory, so that git will return success exit-code
+            // git always return non-zero exit code when run git reset outside of git repository
+            // NUnit working directory always default to MS test host
+            var workingDir = Path.GetDirectoryName(Assembly.GetAssembly(typeof(ExecutableExtensionsTests)).Location);
+            _gitExecutable = new Executable(_appPath, workingDir);
```
drewnoakes: Why do these tests need the actual `git.exe` path? I think this should be mocked out, making all the env var stuff redundant.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,68 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+        /// Implementation by <see cref="System.Diagnostics.Process"/> will have file path included in command line arguments,
+        /// as well as added quotation and space characters, so we need base length to account for all these added characters
+        /// <see href="https://referencesource.microsoft.com/#system/services/monitoring/system/diagnosticts/Process.cs,1944"/>
+        /// </summary>
+        /// <param name="builder">Argument builder instance.</param>
+        /// <param name="arguments">Arguments.</param>
+        /// <param name="baseLength">Base executable file and command line length.</param>
+        /// <param name="maxLength">Command line max length. Default is 32767 - 1 on Windows.</param>
+        /// <returns>Array of batch arguments split by max length.</returns>
+        public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder, IEnumerable<string> arguments, int baseLength, int maxLength = short.MaxValue)
```
drewnoakes: Doesn't this duplicate a bunch of stuff proposed in #7237? Which are you recommending?

drewnoakes:
This change will cause more instances of bug #3280.

drewnoakes:
Can we close this and focus on #7237 instead please? If there's a change here that's not in #7237 too, can you add it there too? The duplication makes reviewing difficult.

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
hieuxlu: After discussion with @mstv, I think testing with actual `git.exe` is valid, because we want to see the `Win32Exception` when command line reaches actual threshold. If we don't test this, we would not be able to see differences of max command line length for each executable. For example, `cmd.exe` has a limit of `8191` characters, while `sh.exe` is limited at `32710` instead of `32767` characters. So testing againt `git.exe` would ensure we have correct understanding of `git` command line max length on Windows.

GitCommands/ArgumentBuilderExtensions.cs
hieuxlu: For #6593 this is the smallest viable amount of code. For #7237, we need to add BatchProgress callback, so if each PR would do only its purpose. I think we should focus on this before #7237.

hieuxlu:
> Can we close this and focus on #7237 instead please? If there's a change here that's not in #7237 too, can you add it there too? The duplication makes reviewing difficult.

I think we should finish this before #7237. That's why #7237 is marked as WIP. #7237 is extending this, not overwriting.

drewnoakes:
> I think we should finish this before #7237. That's why #7237 is marked as WIP. #7237 is extending this, not overwriting.

Fair enough. @RussKie tagged me on #7237 so I saw and commented there first. Please consider my review there as applying here.

hieuxlu:
> Fair enough. @RussKie tagged me on #7237 so I saw and commented there first. Please consider my review there as applying here.

I did cherrypick all the suggested changes from #7237 and vice versa.

gerhardol:
@RussKie @drewnoakes 
Any more comments here, should it be merged?
@hieuxlu Good work. Please rebase on mater and squash before it is merged

hieuxlu:
> @hieuxlu Good work. Please rebase on mater and squash before it is merged

Done. Glad I could contribute back. Been using this software for 5 years and counting.

RussKie:
@msftbot merge in 24 hours

PR #7223, opened 2019-10-01
Exclude "fixup!" and "squash!" prefixes from commit message RegEx validation

pmiossec:
## Proposed changes

That prevent a boring question to be asked to the user
because the commit message doesn't follow the regex when
"fixup!" or "squash!" prefixes are used.

## Test methodology 

- Manual

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -36,6 +36,9 @@ namespace GitUI.CommandsDialogs
 {
     public sealed partial class FormCommit : GitModuleForm
     {
+        private const string fixupPrefix = "fixup!";
+        private const string squashPrefix = "squash!";
+
         #region Translation
 
         private readonly TranslationString _error = new TranslationString("Error");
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1483,6 +1486,8 @@ bool IsCommitMessageValid()
                         try
                         {
                             if (!Regex.IsMatch(Message.Text, AppSettings.CommitValidationRegEx) &&
+                                Message.Text.StartsWith(fixupPrefix) &&
+                                Message.Text.StartsWith(squashPrefix) &&
```
drewnoakes: `Message.Text` will never begin with both of these prefixes. I think there's a logic bug here.

pmiossec: Yes. Fixed. Thanks.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1483,6 +1486,8 @@ bool IsCommitMessageValid()
                         try
                         {
                             if (!Regex.IsMatch(Message.Text, AppSettings.CommitValidationRegEx) &&
+                                !Message.Text.StartsWith(fixupPrefix) &&
+                                !Message.Text.StartsWith(squashPrefix) &&
```
RussKie: [nit] I'd put `!Message.Text.StartsWith(...)` before `!Regex.IsMatch(..)`, as I think the latter would be a more costly operation.
Just a gut feeling, haven't run any benchmarks

pmiossec: Your message made me think again more carefully about it and indeed, you're right, so I have changed it.

PR #7237, opened 2019-10-03
Fix #3280: Optimize unstage/reset performance by git reset/unstage batch files

### LINKED ISSUE #3280: Unstaging many files from Commit dialog freezes UI for a long time

hieuxlu:
Fix #3280: Optimize unstage/reset performance by git reset/unstage batch files (continue from batch processing in #6593)

## Proposed changes
Use RunBatchCommand & BuildBatchArguments from #7218 to fix performance issue with git reset/unstage individual files consequentially.

## Test methodology 

- Manual test
- Unit test cases

## Test environment(s) 

- GIT: 2.23.0.windows.1
- Windows: 10 x64

----

### DIFF
--- a/GitCommands/ArgumentBuilderExtensions.cs
+++ b/GitCommands/ArgumentBuilderExtensions.cs
@@ -315,8 +315,15 @@ namespace GitCommands
         /// <param name="baseLength">Base executable file and command line length.</param>
         /// <param name="maxLength">Command line max length. Default is 32767 - 1 on Windows.</param>
         /// <returns>Array of batch arguments split by max length.</returns>
-        public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder, IEnumerable<string> arguments, int baseLength, int maxLength = short.MaxValue)
+        public static List<BatchArgumentItem> BuildBatchArguments(this ArgumentBuilder builder, IEnumerable<string> arguments, int? baseLength = null, int maxLength = short.MaxValue)
         {
+            // 3: double quotes + ' '
+            // '"git.exe" ' is always included in final command line arguments
+            if (!baseLength.HasValue)
+            {
+                baseLength = AppSettings.GitCommand.Length + 3;
+            }
+
             var baseArgument = builder.ToString();
             if (baseLength + baseArgument.Length >= maxLength)
             {
@@ -324,7 +331,8 @@ namespace GitCommands
             }
 
             // Clone command as argument builder
-            var batches = new List<ArgumentString>();
+            var batches = new List<BatchArgumentItem>();
+            var currentBatchItemCount = 0;
             var currentArgumentLength = baseArgument.Length;
             var lastBatchBuilder = arguments.Aggregate(builder, (currentBatchBuilder, argument) =>
             {
@@ -333,21 +341,23 @@ namespace GitCommands
                 if (baseLength + currentArgumentLength + 1 + argument.Length >= maxLength)
                 {
                     // Handle abnormal case when base command and a single argument exceed max length
-                    if (currentArgumentLength == baseArgument.Length)
+                    if (currentBatchItemCount == 0)
                     {
                         throw new ArgumentException($"Git command \"{currentBatchBuilder}\" always exceeded max length of {maxLength} characters.", nameof(arguments));
                     }
 
                     // Finish current command line
-                    batches.Add(currentBatchBuilder);
+                    batches.Add(new BatchArgumentItem(currentBatchBuilder, currentBatchItemCount));
 
                     // Return new argument builder
+                    currentBatchItemCount = 1;
                     currentArgumentLength = baseArgument.Length + 1 + argument.Length;
                     return new ArgumentBuilder() { baseArgument, argument };
                 }
 
                 currentBatchBuilder.Add(argument);
                 currentArgumentLength += argument.Length + 1;
+                currentBatchItemCount++;
                 return currentBatchBuilder;
             });
 
@@ -360,10 +370,10 @@ namespace GitCommands
             // Add last commandline batch
             if (!lastBatchBuilder.IsEmpty)
             {
-                batches.Add(lastBatchBuilder.ToString());
+                batches.Add(new BatchArgumentItem(lastBatchBuilder, currentBatchItemCount));
             }
 
-            return batches.ToArray();
+            return batches;
         }
     }
 }
--- a/GitCommands/BatchArgumentItem.cs
+++ b/GitCommands/BatchArgumentItem.cs
@@ -1,0 +1,24 @@
+﻿namespace GitCommands
+{
+    /// <summary>
+    /// Result model for batch processing arguments and count of items for batch progress
+    /// </summary>
+    public sealed class BatchArgumentItem
+    {
+        public BatchArgumentItem(ArgumentString argument, int count)
+        {
+            Argument = argument;
+            BatchItemsCount = count;
+        }
+
+        /// <summary>
+        /// Batch command line argument
+        /// </summary>
+        public ArgumentString Argument { get; }
+
+        /// <summary>
+        /// Count of items in batch, used for batch progress update
+        /// </summary>
+        public int BatchItemsCount { get; }
+    }
+}
--- a/GitCommands/Git/BatchProgressEventArgs.cs
+++ b/GitCommands/Git/BatchProgressEventArgs.cs
@@ -1,0 +1,26 @@
+﻿using System;
+
+namespace GitCommands
+{
+    /// <summary>
+    /// Event arguments for batch progress updating
+    /// </summary>
+    public sealed class BatchProgressEventArgs : EventArgs
+    {
+        public BatchProgressEventArgs(int batchItemsProcessed, bool executionResult)
+        {
+            ProcessedCount = batchItemsProcessed;
+            ExecutionResult = executionResult;
+        }
+
+        /// <summary>
+        /// Number of items processed in this batch event
+        /// </summary>
+        public int ProcessedCount { get; }
+
+        /// <summary>
+        /// Batch execution result
+        /// </summary>
+        public bool ExecutionResult { get; }
+    }
+}
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -50,6 +50,38 @@ namespace GitCommands
         }
 
         /// <summary>
+        /// Launches a process for the executable per batch item and returns its output.
+        /// </summary>
+        /// <remarks>
+        /// This method uses <see cref="GetOutput"/> to get concatenated outputs of multiple commands in batch.
+        /// </remarks>
+        /// <param name="executable">The executable from which to launch processes.</param>
+        /// <param name="batchArguments">The array of batch arguments to pass to the executable</param>
+        /// <param name="input">Bytes to be written to each process's standard input stream, or <c>null</c> if no input is required.</param>
+        /// <param name="outputEncoding">The text encoding to use when decoding bytes read from each process's standard output and standard error streams, or <c>null</c> if the default encoding is to be used.</param>
+        /// <param name="cache">A <see cref="CommandCache"/> to use if command results may be cached, otherwise <c>null</c>.</param>
+        /// <param name="stripAnsiEscapeCodes">A flag indicating whether ANSI escape codes should be removed from output strings.</param>
+        /// <returns>The concatenation of standard output and standard error. To receive these outputs separately, use <see cref="Execute"/> instead.</returns>
+        [NotNull]
+        [MustUseReturnValue("If output text is not required, use " + nameof(RunCommand) + " instead")]
+        public static string GetBatchOutput(
+            this IExecutable executable,
+            ICollection<BatchArgumentItem> batchArguments = default,
+            byte[] input = null,
+            Encoding outputEncoding = null,
+            CommandCache cache = null,
+            bool stripAnsiEscapeCodes = true)
+        {
+            var sb = new StringBuilder();
+            foreach (var batch in batchArguments)
+            {
+                sb.Append(executable.GetOutput(batch.Argument, input, outputEncoding, cache, stripAnsiEscapeCodes));
+            }
+
+            return sb.ToString();
+        }
+
+        /// <summary>
         /// Launches a process for the executable and returns its output.
         /// </summary>
         /// <param name="executable">The executable from which to launch a process.</param>
@@ -147,7 +179,7 @@ namespace GitCommands
         /// </summary>
         /// <remarks>
         /// This method uses <see cref="RunCommand"/> to execute multiple commands in batch, used in accordance with
-        /// <see cref="ArgumentBuilderExtensions.BuildBatchArguments(ArgumentBuilder, IEnumerable{string}, int, int)"/>
+        /// <see cref="ArgumentBuilderExtensions.BuildBatchArguments(ArgumentBuilder, IEnumerable{string}, int?, int)"/>
         /// to work around Windows command line length 32767 character limitation
         /// <see href="https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessa"/>.
         /// </remarks>
@@ -159,11 +191,23 @@ namespace GitCommands
         [MustUseReturnValue("Callers should verify that " + nameof(RunBatchCommand) + " returned true")]
         public static bool RunBatchCommand(
             this IExecutable executable,
-            ArgumentString[] batchArguments,
+            ICollection<BatchArgumentItem> batchArguments,
+            Action<BatchProgressEventArgs> action = null,
             byte[] input = null,
             bool createWindow = false)
         {
-            return batchArguments.Aggregate(true, (result, arguments) => executable.RunCommand(arguments, input, createWindow) && result);
+            int total = batchArguments.Sum(item => item.BatchItemsCount);
+            var result = true;
+
+            foreach (var item in batchArguments)
+            {
+                result &= executable.RunCommand(item.Argument, input, createWindow);
+
+                // Invoke batch progress callback
+                action?.Invoke(new BatchProgressEventArgs(item.BatchItemsCount, result));
+            }
+
+            return result;
         }
 
         /// <summary>
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1355,6 +1355,22 @@ namespace GitCommands
             });
         }
 
+        public string ResetFiles(IReadOnlyList<string> files)
+        {
+            if (files == null || files.Count == 0)
+            {
+                return string.Empty;
+            }
+
+            return _gitExecutable.GetBatchOutput(new GitArgumentBuilder("checkout-index")
+                {
+                    "--index",
+                    "--force",
+                    "--"
+                }
+                .BuildBatchArguments(files.Select(item => item.ToPosixPath().Quote())));
+        }
+
         /// <summary>
         /// Delete index.lock in the current working folder.
         /// </summary>
@@ -1402,7 +1418,7 @@ namespace GitCommands
                     revision,
                     "--"
                 }
-                .BuildBatchArguments(files.Select(f => f.ToPosixPath().Quote()), AppSettings.GitCommand.Length + 3));
+                .BuildBatchArguments(files.Select(f => f.ToPosixPath().Quote())));
         }
 
         public string RemoveFiles(IReadOnlyList<string> files, bool force)
@@ -1811,6 +1827,52 @@ namespace GitCommands
             }
 
             return output.ToString();
+        }
+
+        /// <summary>
+        /// Batch unstage files using <see cref="ExecutableExtensions.RunBatchCommand(IExecutable, ICollection{BatchArgumentItem}, Action{BatchProgressEventArgs}, byte[], bool)"/>
+        /// </summary>
+        /// <param name="selectedItems">Selected file items</param>
+        /// <param name="action">Progress update callback</param>
+        /// <returns><see langword="true" /> if changes should be rescanned; otherwise <see langword="false" /></returns>.
+        public bool BatchUnstageFiles(IEnumerable<GitItemStatus> selectedItems, Action<BatchProgressEventArgs> action = null)
+        {
+            var files = new List<GitItemStatus>();
+            var filesToRemove = new List<string>();
+            var shouldRescanChanges = false;
+            foreach (var item in selectedItems)
+            {
+                if (!item.IsNew)
+                {
+                    filesToRemove.Add(item.Name);
+
+                    if (item.IsRenamed)
+                    {
+                        filesToRemove.Add(item.OldName);
+                    }
+
+                    if (item.IsDeleted)
+                    {
+                        shouldRescanChanges = true;
+                    }
+                }
+                else
+                {
+                    files.Add(item);
+                }
+            }
+
+            if (filesToRemove.Count > 0)
+            {
+                var args = GitCommandHelpers.ResetCmd(ResetMode.ResetIndex, "HEAD");
+                _gitExecutable.RunBatchCommand(new ArgumentBuilder() { args }
+                    .BuildBatchArguments(filesToRemove),
+                    action);
+            }
+
+            UnstageFiles(files);
+
+            return shouldRescanChanges;
         }
 
         public async Task<bool> AddInteractiveAsync(GitItemStatus file)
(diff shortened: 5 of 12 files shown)


mstv:
I've changed the title of this PR to indicate that it is work in progress.
It looks like some changes done here, should be done in PR #7218. If yes, it is WIP, too.

gerhardol:
This is based on #7218 and is currently an example how #7218 can improve performance.
(It is possible to review this instead, the PR touches up ExecutableExtensions
Very quick look, without significant comments

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,92 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+            if (!argBuilder.IsEmpty)
+            {
+                result.Add(new BatchArgumentItem
+                {
+                    Argument = argBuilder.ToString(),
+                    Count = itemCount
+                });
+            }
+
+            return result.ToArray();
```
RussKie: `result` is declared as a `List<>`, why are we reallocating again?

hieuxlu: Sorry, I have bad allocation habit. Will correct.

hieuxlu: Removed unnecessary array allocation.

drewnoakes:
Can you find a solution that only batches when necessary to avoid the exception?

drewnoakes:
This change will cause more instances of bug #3280.

hieuxlu:
> Can you find a solution that only batches when necessary to avoid the exception?

Which exception are you referring to? Before this change, I find GitExtensions will optimize batches when user selects all item (then just reset/unstage all instead of batching). Except for this case, I find batching is the most optimized solution.

> I feel this regresses perf too much. Calling `git.exe` N times is wasteful if we aren't going to hit the cmd line length limit.
Can you elaborate on how this regresses perf? Too much allocation? 

> This also seems to duplicate a lot of #7218.

Yes I did comment, this PR is based on #7218, because it reuses `BuildBatchArgument` method. Since #7218 is not merged yet, this PR is marked as _WIP_. You won't see the duplicates anymore once #7218 is merged.


> This change will cause more instances of bug #3280.

After this fix, manual test of #3280 take seconds instead of minutes. Why would it cause more instances of bug #3280?

drewnoakes:
> Can you elaborate on how this regresses perf?

@hieuxlu I'm sorry, I read this too quickly and assumed that:

```text
git command -- file1 file2 file3
```

became:

```text
git command -- file1
git command -- file2
git command -- file3
```

But now I see the behaviour is optimal. Good work. I'll take another look through now.

> After this fix, manual test of #3280 take seconds instead of minutes

I'm really glad to hear this! The lack of batching there has hit me many times.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,92 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+            }
+
+            var baseCommand = builder.ToString();
+            if (baseLength + baseCommand.Length >= maxLength)
+            {
+                throw new ArgumentException($"Git base command \"{baseCommand}\" always reached max length of {maxLength} characters.", nameof(baseCommand));
+            }
+
+            // Clone command as argument builder
+            var tmpBuilder = new ArgumentBuilder() { baseCommand };
```
drewnoakes: Consider inlining `tmpBuilder`. For my taste it makes the `Aggregate` call easier to reason about.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,92 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+
+            return result;
+        }
+    }
+
+    public sealed class BatchArgumentItem
+    {
+        public ArgumentString Argument { get; set; }
+
+        public int Count { get; set; }
```
drewnoakes: Consider making these get-only and adding a constructor.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -333,4 +369,17 @@ private static string CleanString(bool stripAnsiEscapeCodes, [NotNull] string s)
                 : s;
         }
     }
+
+    public sealed class BatchProgressEventArgs : EventArgs
+    {
+        public int Total { get; set; }
+
+        public int Count { get; set; }
```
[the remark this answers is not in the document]
hieuxlu: Done, thanks for pointing out.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,42 @@ string ComposeOutput()
...
+        /// <summary>
+        /// Launches a process for the executable per batch item, and returns <c>true</c> if all process exit codes were zero.
+        /// </summary>
+        /// <remarks>
+        /// This method uses <see cref="RunCommand"/> to execute multiple commands in batch, used in accordance with
+        /// <see cref="ArgumentBuilderExtensions.BuildBatchArguments(ArgumentBuilder, IEnumerable{string}, int?, int)"/>
+        /// to work around Windows command line length 32767 character limitation
+        /// <see href="https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessa"/>.
+        /// </remarks>
+        /// <param name="executable">The executable from which to launch a process.</param>
```
drewnoakes:
```suggestion
        /// <param name="executable">The executable from which to launch processes.</param>
```

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,42 @@ string ComposeOutput()
...
+        /// <param name="batchArguments">The array of batch arguments to pass to the executable</param>
+        /// <param name="input">Bytes to be written to the process's standard input stream, or <c>null</c> if no input is required.</param>
```
drewnoakes:
```suggestion
        /// <param name="input">Bytes to be written to each process's standard input stream, or <c>null</c> if no input is required.</param>
```

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,42 @@ string ComposeOutput()
...
+        /// <param name="createWindow">A flag indicating whether a console window should be created and bound to the process.</param>
```
drewnoakes:
```suggestion
        /// <param name="createWindow">A flag indicating whether a console window should be created and bound to each process.</param>
```

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -141,6 +141,42 @@ string ComposeOutput()
...
+        /// <returns><c>true</c> if all process exit codes were zero, otherwise <c>false</c>.</returns>
+        [MustUseReturnValue("Callers should verify that " + nameof(RunBatchCommand) + " returned true")]
+        public static bool RunBatchCommand(
+            this IExecutable executable,
+            IEnumerable<BatchArgumentItem> batchArguments,
```
drewnoakes:
```suggestion
            ICollection<BatchArgumentItem> batchArguments,
```

Given `batchArguments` is enumerated more than once, require a materialised collection here to avoid duplicated lazy evaluation of an enumerable.

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -300,5 +301,92 @@ public static void Add(this ArgumentBuilder builder, [CanBeNull, ItemCanBeNull]
...
+
+            var baseCommand = builder.ToString();
+            if (baseLength + baseCommand.Length >= maxLength)
+            {
+                throw new ArgumentException($"Git base command \"{baseCommand}\" always reached max length of {maxLength} characters.", nameof(baseCommand));
+            }
+
+            // Clone command as argument builder
+            var tmpBuilder = new ArgumentBuilder() { baseCommand };
+            var result = new List<BatchArgumentItem>();
```
drewnoakes: I would find this easier to read if some locals were renamed:

- `result` ⇒ `batches`
- `argBuilder` ⇒ `lastBatchBuilder`
- `curr` ⇒ `currentBatchBuilder`
- `value` ⇒ `argument`
- `itemCount` ⇒ `currentBatchItemCount`

This is just my opinion though.

GitCommands/Git/GitModule.cs
```diff
@@ -1354,6 +1354,30 @@ public string ResetFile(string file)
...
 
+        public string ResetFiles(IReadOnlyList<string> files)
+        {
+            if (files.Count == 0)
+            {
+                return string.Empty;
+            }
+
+            var sb = new StringBuilder();
+            var arguments = new GitArgumentBuilder("checkout-index")
```
drewnoakes:
```suggestion
            var batches = new GitArgumentBuilder("checkout-index")
```

GitCommands/Git/GitModule.cs
```diff
@@ -1354,6 +1354,30 @@ public string ResetFile(string file)
...
+                {
+                    "--index",
+                    "--force",
+                    "--"
+                }
+                .BuildBatchArguments(files.Select(item => item.ToPosixPath().Quote()));
+
+            foreach (var arg in arguments)
```
drewnoakes:
```suggestion
            foreach (var batch in batches)
```

GitCommands/Git/GitModule.cs
```diff
@@ -1354,6 +1354,30 @@ public string ResetFile(string file)
...
+            {
+                sb.Append(_gitExecutable.GetOutput(arg.Argument));
+            }
+
+            return sb.ToString();
```
drewnoakes: Could the above loop be made an extension method on `_gitExecutable` similar to `RunBatchCommand`, something like:

```
            return _gitExecutable.GetBatchedOutput(batches);

File: UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs [hunk 13]
```diff
@@ -9,11 +17,27 @@ namespace GitCommandsTests.Git
...
+            var userAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            var settingPath = Path.Combine(userAppDataPath, "GitExtensions\\GitExtensions\\GitExtensions.settings");
+            var settingContainer = new RepoDistSettings(null, GitExtSettingsCache.FromCache(settingPath), SettingLevel.Unknown);
+            _appPath = Path.Combine(settingContainer.GetString("gitbindir", ""), "git.exe");
+
+            // Execute process in GitExtension working directory, so that git will return success exit-code
+            // git always return non-zero exit code when run git reset outside of git repository
+            // NUnit working directory always default to MS test host
+            var workingDir = Path.GetDirectoryName(Assembly.GetAssembly(typeof(ExecutableExtensionsTests)).Location);
+            _gitExecutable = new Executable(_appPath, workingDir);
```
drewnoakes: I'm not a fan of calling `git.exe` in unit tests. Is there a reason you can't just mock the executable?

GitCommands/ArgumentBuilderExtensions.cs
hieuxlu: I'll fix this in #7218 and do a merge.

hieuxlu: `tmpBuilder` have been removed in #7218 to avoid extra allocation. I've merged the changes into this as well.

GitCommands/ArgumentBuilderExtensions.cs
hieuxlu: Renamed in both PRs

UnitTests/GitCommandsTests/Git/ExecutableExtensionsTests.cs
hieuxlu: Explained in #7218

GitCommands/ArgumentBuilderExtensions.cs
hieuxlu: Done

GitCommands/Git/GitModule.cs
hieuxlu: Added `GetBatchOutput` extension method. Is `ed` in `Batch` required? I'm not very good with grammar (Should have Grammarly Extension for Visual Studio).

hieuxlu:
@drewnoakes We can now start review this again, #7218 got merged

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -359,10 +369,23 @@ public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder,
...
-                batches.Add(lastBatchBuilder.ToString());
+                batches.Add(new BatchArgumentItem(lastBatchBuilder, currentBatchItemCount));
             }
 
-            return batches.ToArray();
+            return batches;
+        }
+    }
+
+    public sealed class BatchArgumentItem
```
RussKie: Please put it in own file

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -158,11 +190,23 @@ string ComposeOutput()
...
             byte[] input = null,
             bool createWindow = false)
         {
-            return batchArguments.Aggregate(true, (result, arguments) => executable.RunCommand(arguments, input, createWindow) && result);
+            int total = batchArguments.Sum(item => item.Count);
+            var result = true;
+
+            foreach (var item in batchArguments)
+            {
+                result = executable.RunCommand(item.Argument, input, createWindow) && result;
```
RussKie: This looks odd. Why are we using conditional AND?
```suggestion
                result &= executable.RunCommand(item.Argument, input, createWindow);
```

GitCommands/ArgumentBuilderExtensions.cs
```diff
@@ -359,10 +369,23 @@ public static ArgumentString[] BuildBatchArguments(this ArgumentBuilder builder,
...
+    {
+        public BatchArgumentItem(ArgumentString argument, int count)
+        {
+            Argument = argument;
+            Count = count;
         }
+
+        public ArgumentString Argument { get; }
+
+        public int Count { get; }
```
RussKie: Please annotate with xml-docs these properties. I'm struggling to understand what each represents.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -357,4 +401,17 @@ private static string CleanString(bool stripAnsiEscapeCodes, [NotNull] string s)
                 : s;
         }
     }
-}
\ No newline at end of file
+
+    public sealed class BatchProgressEventArgs : EventArgs
```
RussKie: Please put it in own file

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -158,11 +190,23 @@ string ComposeOutput()
...
-            return batchArguments.Aggregate(true, (result, arguments) => executable.RunCommand(arguments, input, createWindow) && result);
+            int total = batchArguments.Sum(item => item.Count);
+            var result = true;
+
+            foreach (var item in batchArguments)
+            {
+                result = executable.RunCommand(item.Argument, input, createWindow) && result;
+
+                // Invoke batch progress callback
+                action?.Invoke(new BatchProgressEventArgs(item.Count, total));
```
RussKie: Should we pass the batch execution result as well?

GitCommands/Git/GitModule.cs
```diff
@@ -1354,6 +1354,22 @@ public string ResetFile(string file)
             });
         }
 
+        public string ResetFiles(IReadOnlyList<string> files)
+        {
+            if (files.Count == 0)
```
RussKie: This could NRE.
```suggestion
            if (files == null || files.Count == 0)
```

RussKie: All public API must perform input args validation, applies everywhere

GitCommands/Git/GitModule.cs
```diff
@@ -1812,6 +1828,60 @@ public string UnstageFiles(IReadOnlyList<GitItemStatus> files)
...
+            BatchUnstageFiles(selectedItems, out var _, action);
+        }
+
+        /// <summary>
+        /// Batch unstage files using <see cref="ExecutableExtensions.RunBatchCommand(IExecutable, ICollection{BatchArgumentItem}, Action{BatchProgressEventArgs}, byte[], bool)"/>
+        /// </summary>
+        /// <param name="selectedItems">Selected file items</param>
+        /// <param name="shouldRescanChanges">Rescan change flag</param>
+        /// <param name="action">Progress update callback</param>
+        public void BatchUnstageFiles(IEnumerable<GitItemStatus> selectedItems, out bool shouldRescanChanges, Action<BatchProgressEventArgs> action = null)
```
RussKie: This API doesn't look good. I don't see why we need an `out` variable:
```suggestion
        /// <returns><see langword="true" /> if changes should be rescanned; otherwise <see langword="false" />.
        public bool BatchUnstageFiles(IEnumerable<GitItemStatus> selectedItems, Action<BatchProgressEventArgs> action = null)
```

And the overload is now redundant.

GitCommands/Git/ExecutableExtensions.cs
hieuxlu: Added. It'll make a better API, though result is not used anywhere ATM.

hieuxlu:
@RussKie I resolved all your review comments and changed accordingly. Thanks.

GitCommands/BatchArgumentItem.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        /// <summary>
+        /// Batch command line argument
+        /// </summary>
+        public ArgumentString Argument { get; }
+
+        /// <summary>
+        /// Count of items in batch, used for batch progress update
+        /// </summary>
+        public int Count { get; }
```
RussKie: A `Count` property is self explanatory for a collection type, however here it isn't as self descriptive.
I propose to rename to make it clearer to consumer what it is:
```suggestion
        public int BatchItemsCount { get; }
```

GitCommands/Git/BatchProgressEventArgs.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+
+namespace GitCommands
+{
+    /// <summary>
+    /// Event arguments for batch progress updating
+    /// </summary>
+    public sealed class BatchProgressEventArgs : EventArgs
+    {
+        public BatchProgressEventArgs(int value, int maxValue, bool result)
```
RussKie:
```suggestion
        public BatchProgressEventArgs(int batchItemsProcessed, int batchItemsCount , bool executionResult)
```

GitCommands/Git/BatchProgressEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            Total = maxValue;
+            Count = value;
+            Result = result;
+        }
+
+        /// <summary>
+        /// Total number of items in all batches
+        /// </summary>
+        public int Total { get; }
```
RussKie:
```suggestion
        public int BatchItemsCount { get; }
```

GitCommands/Git/BatchProgressEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Number of items processed in this batch event
+        /// </summary>
+        public int Count { get; }
```
RussKie:
```suggestion
        public int ProcessedCount { get; }
```

GitCommands/Git/BatchProgressEventArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Batch execution result
+        /// </summary>
+        public bool Result { get; }
```
RussKie:
```suggestion
        public bool ExecutionResult { get; }
```

GitCommands/Git/GitModule.cs
```diff
@@ -1812,6 +1828,52 @@ public string UnstageFiles(IReadOnlyList<GitItemStatus> files)
...
+                var args = GitCommandHelpers.ResetCmd(ResetMode.ResetIndex, "HEAD");
+                _gitExecutable.RunBatchCommand(new ArgumentBuilder() { args }
+                    .BuildBatchArguments(filesToRemove),
+                    action);
+            }
+
+            UnstageFiles(files);
+
+            return shouldRescanChanges;
+        }
```
RussKie: Can we add tests for this?

GitCommands/Git/GitModule.cs
```diff
@@ -1354,6 +1354,22 @@ public string ResetFile(string file)
...
+            }
+
+            return _gitExecutable.GetBatchOutput(new GitArgumentBuilder("checkout-index")
+                {
+                    "--index",
+                    "--force",
+                    "--"
+                }
+                .BuildBatchArguments(files.Select(item => item.ToPosixPath().Quote())));
+        }
```
RussKie: ditto

RussKie:
@drewnoakes any thoughts on API names?

GitCommands/Git/GitModule.cs
hieuxlu: What is ditto?

hieuxlu: Oh, add the test here as well right? OK

RussKie: > "ditto" - The same thing again (used in lists and accounts and often indicated by a ditto mark under the word or figure to be repeated).
https://www.lexico.com/en/definition/ditto

Basically, the same issue as pointed above

RussKie: Yes, I'd like to see tests for the new functionality

hieuxlu: Thought you were into Pokemon at first (same here). Test cases added.

GitCommands/Git/GitModule.cs
hieuxlu: This one is tricky because it calls UnstageFiles internally. Had to fix NRE for MockProcess.Verify. Test cases added.

GitCommands/Git/GitModule.cs
RussKie: > Thought you were into Pokemon at first

LOL, I had to look up what Pokemons were :)

RussKie:
@drewnoakes any comments?

hieuxlu:
@RussKie Sorry to push, but can we got this merged? I want to fix #7263, so need this merged to avoid code conflicts

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1677,42 +1677,24 @@ private void Unstage(bool canUseUnstageAll = true)
...
-                        allFiles.Add(item);
-                    }
+                        return item.IsRenamed ? count + 2 : count + 1;
+                    });
 
-                    Module.UnstageFiles(files);
+                    Staged.StoreNextIndexToSelect();
+                    var shouldRescanChanges = Module.BatchUnstageFiles(allFiles, (eventArgs) =>
+                    {
+                        toolStripProgressBar1.Value = Math.Min(toolStripProgressBar1.Maximum - 1, toolStripProgressBar1.Value + eventArgs.BatchItemsCount);
```
hieuxlu: @RussKie  Sorry I think I got this mixed up after the rename. This should be ProcessedCount.

pmiossec:
>Sorry to push, but can we got this merged? I want to fix #7263, so need this merged to avoid code conflicts

Workaround: If you are confident enough, you could create your branch based on `feature/batch-performance` and you just will have to rebase on `master` once this one will be merged...

hieuxlu:
Will do

hieuxlu:
I found a bug, so I need to change code to this PR anyway.

hieuxlu:
@RussKie @gerhardol I found a bug with calculation of progress bar value and fixed it + add test cases, so please review again. Sorry for bother you guys again. Thanks. 
```diff
- toolStripProgressBar1.Value = Math.Min(toolStripProgressBar1.Maximum - 1, toolStripProgressBar1.Value + eventArgs.BatchItemsCount);
+ toolStripProgressBar1.Value = Math.Min(toolStripProgressBar1.Maximum - 1, toolStripProgressBar1.Value + eventArgs.ProcessedCount);
```

PR #7247, opened 2019-10-05
Commit dialog, FileStatusList: Fix up/down selects unexpected file

### LINKED ISSUE #6212: The method "RestoreSelectedFiles" should always update "SelectedIndex" of both staged and unstaged file list

### LINKED ISSUE #7246: Commit dialog, FileStatusList: Up/down selects unexpected file
## Current behaviour

- pushing `Up`/`Down` keys does not have an effect or selects an unexpected file, e.g. after refresh

## Expected behaviour

- the file selection can correctly be moved using the keyboard 

## Steps to reproduce

- open the Commit dialog
- have at least two unstaged files
- select the last
- refresh (`F5`)
- hit `Up` => nothing happens
- hit `Down` => the first file is selected

Have not checked

## Environment

- Git Extensions 3.3.0
- Build 00d5e2db95a84a5c3efa153fa38fb0a4a9397fbb
- Git 2.23.0.windows.1
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

mstv:
Fixes #7246 
Fixes #6212

## Proposed changes

- Set FileStatusListView.FocusedItem, too

## Test methodology 

- manual test (see STR in #7246)
- added NUnit test

## Test environment(s)

- Git Extensions 3.3.0
- Build 420bcdfe3fab4baef6a69b62698151b3170c2b35
- Git 2.23.0.windows.1
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -298,16 +298,7 @@ namespace GitUI
 
                 return -1;
             }
-            set
-            {
-                ClearSelected();
-                if (value >= 0)
-                {
-                    FileStatusListView.Items[value].Selected = true;
-                    FileStatusListView.Items[value].Focused = true;
-                    FileStatusListView.Items[value].EnsureVisible();
-                }
-            }
+            set => SelectItems(item => item.Index == value);
         }
 
         [CanBeNull]
@@ -315,42 +306,39 @@ namespace GitUI
         [Browsable(false)]
         public GitItemStatus SelectedItem
         {
-            get
-            {
-                return FileStatusListView.LastSelectedItem()?.Tag<GitItemStatus>();
-            }
-
+            get => FileStatusListView.LastSelectedItem()?.Tag<GitItemStatus>();
             set
             {
-                ClearSelected();
-                if (value == null)
-                {
-                    return;
-                }
+                var itemToBeSelected = GetItemByStatus(value);
+                SelectItems(item => item == itemToBeSelected);
+                return;
 
-                ListViewItem newSelected = null;
-                foreach (ListViewItem item in FileStatusListView.Items)
+                ListViewItem GetItemByStatus(GitItemStatus status)
                 {
-                    var gitItemStatus = item.Tag<GitItemStatus>();
-
-                    if (value.CompareTo(gitItemStatus) == 0)
+                    if (status == null)
                     {
-                        if (newSelected == null)
+                        return null;
+                    }
+
+                    ListViewItem newSelected = null;
+                    foreach (ListViewItem item in FileStatusListView.Items)
+                    {
+                        var gitItemStatus = item.Tag<GitItemStatus>();
+
+                        if (status.CompareTo(gitItemStatus) == 0)
                         {
-                            newSelected = item;
-                        }
-                        else if (gitItemStatus == value)
-                        {
-                            newSelected = item;
-                            break;
+                            if (newSelected == null)
+                            {
+                                newSelected = item;
+                            }
+                            else if (gitItemStatus == status)
+                            {
+                                return item;
+                            }
                         }
                     }
-                }
 
-                if (newSelected != null)
-                {
-                    newSelected.Selected = true;
-                    newSelected.EnsureVisible();
+                    return newSelected;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


mstv:
@mstfbot, merge in 1 day

mstv:
@msftbot, merge in 1 day

mstv:
@msftbot, merge in 16 hours

RussKie:
I'm on a phone, but it looks like we have an inconsistent behaviour for `SelectedItem` vs `SelectedItems` as far as the focus concerned.
Could you please review both and ensure they are aligned in behaviour.

A set of tests would be good too.

mstv:
You're so right, applies to `SelectedIndex`, too: Redundant interface implemented using Copy & Paste! 😠

RussKie:
We are following CPD - copy-paste development :D

GitUI/UserControls/FileStatusList.cs
```diff
@@ -592,25 +558,7 @@ int GetLastIndex()
...
-            {
-                SuspendLayout();
-
-                var itemCount = AllItems.Count();
-
-                for (var i = 0; i < itemCount; i++)
-                {
-                    FileStatusListView.Items[i].Selected = true;
-                    i++;
-                }
```
mstv: another bug fixed

GitUI/UserControls/FileStatusList.cs
```diff
@@ -826,6 +755,38 @@ private async Task OpenSubmoduleAsync()
             process.Start();
         }
 
+        private void SelectItems(Func<ListViewItem, bool> isToBeSelectedFunc)
+        {
+            try
+            {
+                SuspendLayout();
```
mstv: I guess, this is pointless, isn't it? (factored out from some of the replaced functions)

mstv:
WIP only because of missing NUnit tests

GitUI/UserControls/FileStatusList.cs
```diff
@@ -727,28 +675,9 @@ public void SetSelectedIndex(int idx, bool notify)
...
-                    return FileStatusListView.SelectedIndices.Count;
-                }
-                finally
-                {
-                    ResumeLayout(true);
-                }
-            }
+            var regex = RegexForSelecting(selectionFilter);
+            SelectItems(item => regex.IsMatch(item.Name));
+            return FileStatusListView.SelectedIndices.Count;
```
mstv: Almost hidden feature: Have you ever hit `Ctrl+F` in the Unstage Files List?

RussKie: 👀 
No... That feature looks very strange.
We have a filter dropdown that is much more useful. The same filter should be in the Staged list too (and the Diff form too).

I would recommend that feature for removal.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -298,59 +298,47 @@ public int SelectedIndex
...
-                            newSelected = item;
-                            break;
+                            if (newSelected == null)
+                            {
+                                newSelected = item;
+                            }
+                            else if (gitItemStatus == status)
+                            {
+                                return item;
+                            }
```
RussKie: Can we please invert it to be:
```cs
if (status.CompareTo(gitItemStatus) == 0)
{
    if (gitItemStatus == status) {
        return item;
    }

    if (newSelected == null) { ... }
}
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -826,6 +755,38 @@ private async Task OpenSubmoduleAsync()
             process.Start();
         }
 
+        private void SelectItems(Func<ListViewItem, bool> isToBeSelectedFunc)
```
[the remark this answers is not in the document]
mstv: > Linq has relatively simple arg names, like func, predicate, source etc.

Yes, I started with `predicate` and thought the readability of the imlementation could be better.
Switched back.

GitUI/UserControls/FileStatusList.cs
mstv: This "inversion" will not have the same behavior. Note the `else` with respect to the `foreach` loop.

I have only moved this - not very self-explaining - logic. I don't want to change it.

RussKie: You are right, I was tired and didn't think it through.

PR #7251, opened 2019-10-06
Add tests for `ArgumentBuilder`

RussKie:


hieuxlu:
@RussKie I think you misunderstand my point. I was talking about `GitArgumentBuilder`. Try 
```
new GitArgumentBuilder("foo") \\ Length = 0 while actual length is 3
```

RussKie:
> I was talking about `GitArgumentBuilder`.

Thank you, I'll look into this.

PR #7252, opened 2019-10-06
Remove `ArgumentBuilder.Length` property

RussKie:
## Proposed changes

- #7218 has removed all dependencies on `Length` property, rendering it obsolete.

## Test methodology 

- manual
- unit tests

----

### DIFF
--- a/GitExtUtils/ArgumentBuilder.cs
+++ b/GitExtUtils/ArgumentBuilder.cs
@@ -5,7 +5,7 @@ using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
 
-namespace GitCommands
+namespace GitExtUtils
 {
     /// <summary>
     /// Builds a command line argument string from zero or more arguments.
--- a/GitExtUtils/ArgumentBuilder.cs
+++ b/GitExtUtils/ArgumentBuilder.cs
@@ -39,7 +39,6 @@ namespace GitCommands
         private readonly StringBuilder _arguments = new StringBuilder(capacity: 16);
 
         public bool IsEmpty => _arguments.Length == 0;
-        public int Length => _arguments.Length;
 
         /// <summary>
         /// Adds <paramref name="s"/> to the argument list.
@@ -94,6 +93,21 @@ namespace GitCommands
         IEnumerator IEnumerable.GetEnumerator()
         {
             throw new InvalidOperationException($"{nameof(IEnumerable)} only implemented to support collection initialiser syntax.");
+        }
+
+        internal TestAccessor GetTestAccessor()
+            => new TestAccessor(this);
+
+        internal readonly struct TestAccessor
+        {
+            private readonly ArgumentBuilder _builder;
+
+            public TestAccessor(ArgumentBuilder builder)
+            {
+                _builder = builder;
+            }
+
+            public StringBuilder Arguments => _builder._arguments;
         }
     }
 }
--- a/GitExtUtils/GitArgumentBuilder.cs
+++ b/GitExtUtils/GitArgumentBuilder.cs
@@ -4,9 +4,10 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using GitCommands;
 using JetBrains.Annotations;
 
-namespace GitCommands
+namespace GitExtUtils
 {
     /// <summary>
     /// Builds a git command line string from config items, a command name and that command's arguments.
--- a/GitExtUtils/GitArgumentBuilder.cs
+++ b/GitExtUtils/GitArgumentBuilder.cs
@@ -38,7 +38,7 @@ namespace GitCommands
     /// </example>
     public sealed class GitArgumentBuilder : ArgumentBuilder
     {
-        private static readonly Regex _commandRegex = new Regex("^[a-z0-9_.-]+$", RegexOptions.Compiled);
+        private static readonly Regex CommandRegex = new Regex("^[a-z0-9_.-]+$", RegexOptions.Compiled);
 
         private readonly List<GitConfigItem> _configItems;
         private readonly ArgumentString _gitArgs;
@@ -59,7 +59,7 @@ namespace GitCommands
                 throw new ArgumentNullException(nameof(command));
             }
 
-            if (!_commandRegex.IsMatch(command))
+            if (!CommandRegex.IsMatch(command))
             {
                 throw new ArgumentException($"Git command \"{command}\" contains invalid characters.", nameof(command));
             }
--- a/GitExtUtils/GitArgumentBuilder.cs
+++ b/GitExtUtils/GitArgumentBuilder.cs
@@ -1,86 +1,13 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using JetBrains.Annotations;
+
 namespace GitCommands
 {
-    /// <summary>
-    /// A configuration key/value pair for use in git command lines.
-    /// </summary>
-    public readonly struct GitConfigItem
-    {
-        public string Key { get; }
-        public string Value { get; }
-
-        public GitConfigItem(string key, string value)
-        {
-            Key = key;
-            Value = value;
-        }
-
-        public void Deconstruct(out string key, out string value)
-        {
-            key = Key;
-            value = Value;
-        }
-    }
-
-    public sealed class GitCommandConfiguration
-    {
-        private readonly ConcurrentDictionary<string, GitConfigItem[]> _configByCommand
-            = new ConcurrentDictionary<string, GitConfigItem[]>(StringComparer.Ordinal);
-
-        /// <summary>
-        /// Gets the default configuration for git commands used by Git Extensions.
-        /// </summary>
-        public static GitCommandConfiguration Default { get; }
-
-        static GitCommandConfiguration()
-        {
-            // The set of default configuration items for Git Extensions
-            Default = new GitCommandConfiguration();
-
-            Default.Add(new GitConfigItem("rebase.autoSquash", "false"), "rebase");
-
-            Default.Add(new GitConfigItem("log.showSignature", "false"), "log", "show", "whatchanged");
-
-            Default.Add(new GitConfigItem("diff.submodule", "short"), "diff");
-            Default.Add(new GitConfigItem("diff.noprefix", "false"), "diff");
-        }
-
-        /// <summary>
-        /// Registers <paramref name="configItem"/> against one or more command names.
-        /// </summary>
-        /// <param name="configItem">The config item to register.</param>
-        /// <param name="commands">One or more command names to register this config item against.</param>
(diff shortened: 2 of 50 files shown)


RussKie:
@hieuxlu how about this?

GitExtUtils/ArgumentBuilder.cs
```diff
@@ -77,6 +76,12 @@ public void AddRange(IEnumerable<string> args)
             }
         }
 
+        /// <summary>
+        /// Gets the length of the expected command.
+        /// </summary>
+        /// <returns>THe length of the expected command.</returns>
```
drewnoakes:
```suggestion
        /// <returns>The length of the expected command.</returns>
```

GitExtUtils/ArgumentBuilder.cs
```diff
@@ -77,6 +76,12 @@ public void AddRange(IEnumerable<string> args)
...
+        public virtual int GetLength() => _arguments.Length;
```
drewnoakes: This doesn't need to be converted to a method to be made virtual. Personally I prefer it as a property.

GitExtUtils/GitArgumentBuilder.cs
```diff
@@ -171,22 +98,40 @@ public void Add(GitConfigItem configItem)
             }
         }
 
-        /// <inheritdoc />
-        public override string ToString()
+        /// <summary>
+        /// Gets the length of the expected command.
+        /// </summary>
+        /// <returns>THe length of the expected command.</returns>
```
drewnoakes: I'd use `/// <inheritdoc />` here to avoid duplicating documentation that could get out of date.

GitExtUtils/ArgumentBuilder.cs
RussKie: Yes, correct, it doesn't.
The reason I changed it to a method because an override may performs calculations and, thus, a result would be non-trivial/non-deterministic.
A property is typically safe to reuse across multiple call sites, however you won't invoke a method over and over again.

GitExtUtils/GitArgumentBuilder.cs
```diff
@@ -171,22 +98,40 @@ public void Add(GitConfigItem configItem)
...
+            var capacity = GetLength();
             var str = new StringBuilder(capacity);
 
+            var arguments = base.ToString();
+            var gitArgsLength = _gitArgs.Length;
+            if (gitArgsLength == 0)
+            {
+                gitArgsLength = -1; // prevent extra capacity when length is 0
+            }
+
```
mstv: not needed anymore

UnitTests/GitExtUtils/GitArgumentBuilderTests.cs
```diff
@@ -168,5 +169,29 @@ public void Throws_for_invalid_command_strings()
...
+            builder.GetLength().Should().Be(expectedLength + /* 'foo test ' */9);
+
+            builder = new GitArgumentBuilder("for-each-ref", gitOptions: "--no-optional-locks")
+            {
+                "--sort=-committerdate",
+                "--format=\"%(objectname) %(refname)\"",
+                "refs/"
+            };
+            var rendered = builder.ToString();
+            builder.GetLength().Should().Be(rendered.Length);
```
mstv: Nit: I would call `GetLength` before `ToString` because the state of `builder` could change in theory. (C# should have `const` methods.)

GitExtUtils/ArgumentBuilder.cs
gerhardol: nitpick: commit message protected instead of public

RussKie: Thank you, forgot to amend the message

GitExtUtils/GitArgumentBuilder.cs
RussKie: Why?

mstv: Because it is used only once: in the next line `if (>0)`.

GitExtUtils/ArgumentBuilder.cs
drewnoakes: > an override may performs calculations and, thus, a result would be non-trivial/non-deterministic

Does the existing override do that? There are lots of virtual properties in the BCL for which overrides may do whatever they like, but the API reads better with them as a property. It does not feel like querying the length of something should be particularly expensive, and certainly should be idempotent.

I think years of Java gave me a distaste for 'get' methods :)

hieuxlu:
@RussKie Personally I dont like the idea of recalculating Length in every for loop, that would be worse than cloning ArgumentBuilder just once. So I would still prefer cloning a new builder after this refactoring.

GitExtUtils/GitArgumentBuilder.cs
```diff
@@ -171,22 +98,40 @@ public void Add(GitConfigItem configItem)
...
             var gitArgsLength = _gitArgs.Length;
             if (gitArgsLength == 0)
             {
                 gitArgsLength = -1; // prevent extra capacity when length is 0
             }
 
             // 7 = "-c " + "=" + " " + 2 (for possible quotes of value)
-            var capacity = _configItems.Sum(i => i.Key.Length + i.Value.Length + 7) + _command.Length + 1 + arguments.Length;
-            capacity += gitArgsLength + 1;
+            var length = _configItems.Sum(i => i.Key.Length + i.Value.Length + 7) +
```
drewnoakes: `Sum` here will allocate. `foreach` over `List<>` will not.

GitExtUtils/GitArgumentBuilder.cs
```diff
@@ -171,22 +98,40 @@ public void Add(GitConfigItem configItem)
...
         {
-            var arguments = base.ToString();
             var gitArgsLength = _gitArgs.Length;
             if (gitArgsLength == 0)
             {
                 gitArgsLength = -1; // prevent extra capacity when length is 0
             }
 
             // 7 = "-c " + "=" + " " + 2 (for possible quotes of value)
-            var capacity = _configItems.Sum(i => i.Key.Length + i.Value.Length + 7) + _command.Length + 1 + arguments.Length;
```
hieuxlu: @RussKie  @mstv  I don't think adding `2` is correct. If you think of `GetLength` as the exact length of the command, we should not `+2` in case config item does not `NeedEscaping`. So `Length` is operating like `Capacity`, and we can easily get confused when using `GetLength`, and just received `Capacity` instead.

hieuxlu:
@RussKie I have this idea. Instead of inheriting from `ArgumentBuilder` fall into this `override GetLength` situation, we could just decouple `GitArgumentBuilder` from `ArgumentBuidler`. In fact `GitArgumentBuilder.Length` is not used in the entire source code. We just need to `explitly cast` it to `ArgumentString` and we'll be fine. 
What do you think? Mind if I submit this in another PR?

GitExtUtils/GitArgumentBuilder.cs
RussKie: Ah yes, I keep overlooking Linq often allocates
Thank you 👍

GitExtUtils/GitArgumentBuilder.cs
RussKie: > So `Length` is operating like `Capacity`, and we can easily get confused when using `GetLength`, and just received `Capacity` instead.

This is a fair remark. However in the context of the change you're proposing for breaking up a long command - does it really matter? A capacity is the top boundary (i.e. a worse case scenario). 
The reason we precompute the capacity (i.e. the max length of the command) is to ensure we allocate only once.

RussKie:
> We just need to `explicit cast` it to `ArgumentString` and we'll be fine.

I don't think this assumption is correct. `GitArgumentBuilder` has an extra logic that accounts for `GitCommandConfiguration`s, so checking for `ArgumentBuidler`'s length would be incorrect.
I actually run through these assumptions while writing tests and making this change.

RussKie:
> So I would still prefer cloning a new builder after this refactoring.

GE is all about running git commands under the hood, and we run a lot of them. This code has been designed in a way to reduce allocations and make sure we do it as efficiently as we can.
This is why I am concerned with extra allocations in this specific area.

GitExtUtils/GitArgumentBuilder.cs
RussKie: It is actually needed.
We add 1 to `gitArgsLength` in the following line, so to not add if `gitArgsLength` is empty we take 1 here.

GitExtUtils/ArgumentBuilder.cs
RussKie: Fair enough, I changed it back to a property.
I have a dislike for Java syntax too :)

GitExtUtils/GitArgumentBuilder.cs
hieuxlu: It'll work fine for long command case, but we must also change the boundary test cases to adapt this logic. That and the logic itself would be confusing. 
My point is we could add a extra calculation, so `Length` would be exact length. For example:
```csharp
bool isQuoted = configItem.Value.StartsWith("\"", StringComparison.Ordinal) && configItem.Value.EndsWith("\"", StringComparison.Ordinal);
length += configItem.Key.Length + configItem.Value.Length + (isQuoted ? 7 : 5);
```

GitExtUtils/GitArgumentBuilder.cs
mstv: Github has hidden too many lines. Look at the remaining function `ToString` (not at the factored out `GetLength`):
```
        public override string ToString()
        {
            var capacity = Length;
            var str = new StringBuilder(capacity);

            var arguments = base.ToString();
            var gitArgsLength = _gitArgs.Length;
            if (gitArgsLength == 0)
            {
                gitArgsLength = -1; // prevent extra capacity when length is 0
            }

            if (gitArgsLength > 0)
```
There's no `add 1`. The last line is the only usage. It can be replaced with `if (_gitArgs.Length) > 0`.

GitExtUtils/GitArgumentBuilder.cs
kbilsted: @drewnoakes  Pardon my ignorance, what do you mean `Sum` allocates? Doesn't a a foreach also allocate an iterator.

drewnoakes: @kbilsted for `IEnumerable<>` yes there is an allocation, but `List<>` has a struct enumerator (which is not boxed) so avoids the allocation.

hieuxlu:
@RussKie I remove #7218 dependency on `ArgumentBuilder.Length`, so we don't really need to fix this now. IMHO, we should remove `ArgumentBuilder.Length`. It is not used in the project, and it is risky to make it `virtual`. We should not rely on inheritance to assume that `Length` property is always recalculated to be exact command length. That would impact largely on responsibility of future builders extending from `ArgumentBuilder`, easily violate SOLID and so on.

RussKie:
Thank you, I will have a look in coming days.

> That would impact largely on responsibility of future builders extending from `ArgumentBuilder`, easily violate SOLID and so on.

💯

RussKie:
Updated the PR to remove the `Length` property completely.

Issue #7255, opened 2019-10-06
Multiple push URL support


gerhardol:
## Feature description

Git supports multiple push urls for remotes. This allows pushing to several repos at once (like both GitHub and GitLab) at once (instead of using several remotes). 
GE allowed this use without configuration pre 3.0 or 3.1, restored support in #7214.

A full support in the GE GUI could consist of the following:
[ ] Viewing the push url in the left panel tool tip
[ ] Viewing the push url in FormPush
[ ] Viewing the push url in FormRemote
[ ] Modifying multiple pushurl in form remote

To add pushurl manually edit .git/config or:
git remote set-url --add [--push] <name> <newurl>

To support this the internal way to retrieve pushurl must be aligned first. Some info in:
https://github.com/gitextensions/gitextensions/pull/7214#discussion_r331758566

Personally, I believe just viewing the pushurl (i.e. in the left panel tooltip) is enough for most, multiple remotes are a better solution.

## Environment

- Git Extensions 3.3.0
- Build 5725ab3207b0b06a252cd8fad3c4266159b9d1b7
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4010.0
- DPI 96dpi (no scaling)

drewnoakes:
IIRC this feature is only available in Git 1.8 or later. That's older than the last supported version, but it might still be worth considering whether we put a flag in `GitVersion` for this.

gerhardol:
#7256 was dropped as tests need to be changed.

PR #7256, opened 2019-10-06
wip multiple pushurl implementation align

gerhardol:
Preparation for #7255 , point 1

Followup from #7214

## Proposed changes
Unify the retrieval of remote push/fetchurl
This is required in a later stage, if multiple pushurls are to be supported

The existing retrieval of urls is still required for disabled remotes
There is still some access of remotenames->SettingKeyString.RemoteUrl from config in Plugins

This PR could also expose the use of multiple PushUrls in the application (but they only use one anyway, which should be unified so should provide both PushUrl and PushUrls).

Added then removed point 2 from #7255 

## Test methodology 
Tests already exists

----

### DIFF
--- a/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
+++ b/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using GitCommands.Config;
+using GitUI;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
 
@@ -190,7 +191,9 @@ namespace GitCommands.Remotes
         /// </summary>
         public IReadOnlyList<string> GetEnabledRemoteNames()
         {
-            return GetModule().GetRemoteNames();
+            return ThreadHelper.JoinableTaskFactory.Run(async () => (await GetModule().GetRemotesAsync()))
+                .Select(r => r.Name)
+                .ToList();
         }
 
         /// <summary>
@@ -214,17 +217,16 @@ namespace GitCommands.Remotes
         // TODO: candidate for Async implementations
         public IEnumerable<ConfigFileRemote> LoadRemotes(bool loadDisabled)
         {
-            var remotes = new List<ConfigFileRemote>();
             var module = _getModule();
             if (module == null)
             {
-                return remotes;
+                return Array.Empty<ConfigFileRemote>();
             }
 
-            PopulateRemotes(remotes, true);
+            List<ConfigFileRemote> remotes = PopulateEnabledRemotes();
             if (loadDisabled)
             {
-                PopulateRemotes(remotes, false);
+                PopulateDisabledRemotes(remotes);
             }
 
             return remotes.OrderBy(x => x.Name);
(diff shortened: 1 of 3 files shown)


GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
```diff
@@ -214,17 +215,16 @@ public IReadOnlyList<string> GetEnabledRemoteNamesWithoutBranches()
         // TODO: candidate for Async implementations
         public IEnumerable<ConfigFileRemote> LoadRemotes(bool loadDisabled)
         {
-            var remotes = new List<ConfigFileRemote>();
             var module = _getModule();
             if (module == null)
             {
-                return remotes;
+                return new List<ConfigFileRemote>();
```
RussKie:
```suggestion
                return Array.Empty<ConfigFileRemote>();
```

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
```diff
@@ -403,22 +403,38 @@ public void ToggleRemoteState(string remoteName, bool disabled)
...
+            List<ConfigFileRemote> allRemotes = new List<ConfigFileRemote>();
+            const bool enabled = true;
 
-            Func<IReadOnlyList<string>> func;
-            if (enabled)
-            {
-                func = () => module.GetRemoteNames();
-            }
-            else
+            var remotes = ThreadHelper.JoinableTaskFactory.Run(async () => await module.GetRemotesAsync()).ToList();
```
RussKie: Do we need to `ToList` here?

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
gerhardol: fixing

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
gerhardol: no, removing
(still looking at the tests)

gerhardol:
I pushed a change that unified more functionality broke even more tests...

ConfigFileRemoteSettingsManagerTests.cs is written to assume that certain internal GitModule methods are called. The Git remote parsing is tested separately, this handling could be adapted to test Disabled remotes instead (that is handled similar to how enabled remotes were handled before).

I looked a little at the tests but see no obvious way to go forward here.
Therefore, I expect to close this PR (even if the actual change makes sense). I do not have time (last week has been crazy), the change only unifies functionality.

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
```diff
@@ -190,7 +191,9 @@ public IReadOnlyList<string> GetDisabledRemoteNames()
         /// </summary>
         public IReadOnlyList<string> GetEnabledRemoteNames()
         {
-            return GetModule().GetRemoteNames();
+            return ThreadHelper.JoinableTaskFactory.Run(async () => (await GetModule().GetRemotesAsync()))
+            .Select(r => r.Name)
+            .ToList();
```
RussKie: Please fix the indentation

gerhardol:
Failing tests disabled. Just to be clear: I do not plan to merge like this, more likely closing the issue

gerhardol:
Closing, no good ideas how to replace the tests - at least for now

PR #7267, opened 2019-10-08
prevent flickering when selecting files in staged/unstaged FileStatusList

### LINKED ISSUE #7266: Flickering in FormCommit when selecting not the first item in Staged/Unstaged
## Current behavior
When selecting not the first item by mouse in Staged/Unstaged lists you first see diff of the first item for a short while before seeing the diff of the item clicked on.

## Expected behavior
See selected items diff immediately

## Screenshots

This happens because `FormCommit` in `Unstaged_Enter` / `Staged_Enter` resets the `FileStatusList.SelectedIndex` to `0` regardless of what's been clicked on.

Most likely introduced in https://github.com/gitextensions/gitextensions/pull/3920

mdonatas:
Fixes #7266

## Proposed changes

- It appears that first item in staged/unstaged file status list should be selected only if `Entering` the control by keyboard as when using the mouse you either click on specific item to select it or on an empty space to deselect everything
- Add overridden `Enter` event to `FileStatusList` which would provide extra details as to how control was entered
- Only select first item if control wasn't entered by mouse

## Test methodology

- Manual testing

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -36,6 +36,7 @@ namespace GitUI
         private int _nextIndexToSelect = -1;
         private bool _groupByRevision;
         private bool _enableSelectedIndexChangeEvent = true;
+        private bool _mouseEntered;
         private ToolStripItem _openSubmoduleMenuItem;
         private Rectangle _dragBoxFromMouseDown;
         private IReadOnlyList<(GitRevision revision, IReadOnlyList<GitItemStatus> statuses)> _itemsWithParent = Array.Empty<(GitRevision, IReadOnlyList<GitItemStatus>)>();
@@ -44,11 +45,14 @@ namespace GitUI
 
         private bool _updatingColumnWidth;
 
+        public delegate void EnterEventHandler(object sender, EnterEventArgs e);
+
         public event EventHandler SelectedIndexChanged;
         public event EventHandler DataSourceChanged;
 
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
+        public new event EnterEventHandler Enter;
 
         public FileStatusList()
         {
@@ -76,6 +80,8 @@ namespace GitUI
 
             _fullPathResolver = new FullPathResolver(() => Module.WorkingDir);
             _revisionTester = new GitRevisionTester(_fullPathResolver);
+
+            base.Enter += FileStatusList_Enter;
 
             return;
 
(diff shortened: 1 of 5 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -1200,6 +1205,26 @@ private void FileStatusListView_KeyDown(object sender, KeyEventArgs e)
             }
         }
 
+        private bool _mouseEntered;
```
mdonatas: whoops, this should probably be at the top

RussKie: yes, please

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1200,6 +1205,26 @@ private void FileStatusListView_KeyDown(object sender, KeyEventArgs e)
...
+
+        private void FileStatusList_Enter(object sender, EventArgs e)
+        {
+            Enter?.Invoke(this, new EnterEventArgs(_mouseEntered));
+            _mouseEntered = false;
+        }
+
+        private const int WM_MOUSEACTIVATE = 0x0021;
```
RussKie: Please use a definition from `GitUI.NativeConstants`

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1200,6 +1205,26 @@ private void FileStatusListView_KeyDown(object sender, KeyEventArgs e)
...
+
+        protected override void WndProc(ref Message m)
```
RussKie: Please move above all private methods to after `DisposeCustomResources` method.

RussKie:
To confirm - if use tab to cycle between controls, we don't get flickering, do we?

GitUI/UserControls/FileStatusList.cs
```diff
@@ -780,6 +786,16 @@ protected override void DisposeCustomResources()
             _diffListSortSubscription?.Dispose();
         }
 
+        protected override void WndProc(ref Message m)
+        {
+            if (m.Msg == NativeConstants.WM_MOUSEACTIVATE)
+            {
+                _mouseEntered = !Focused;
```
mdonatas: Changed this from simply `true` to `!Focused` as if you'd click on a list a second time this would get set to `true` and would remain so as `Enter` would not fire and this would not get reset to `False`.
Found this after some more manual testing.

GitUI/UserControls/FileStatusList.cs
mdonatas: > 
> 
> To confirm - if use tab to cycle between controls, we don't get flickering, do we?

We don't with this fix and we didn't as when you'd tab around the first item gets selected so even if it used to flicker _technically_ it wasn't visible as it'd `flicker` between the same item.

mdonatas:
> To confirm - if use tab to cycle between controls, we don't get flickering, do we?

We don't with this fix and we didn't as when you'd tab around the first item gets selected so even if it used to flicker technically it wasn't visible as it'd flicker between the same item.

PR #7300, opened 2019-10-14
Incorrect placement of  update submodules on checkout  setting #7290

### LINKED ISSUE #7290: Incorrect placement of "Update submodules on checkout" setting
## Current behaviour

It is possible to configure whether submodules are updated on checkout (e.g. when a revision is checkout).
As discovered as part of analysis for #7274, the functionality is guarded by the following checkbox:
grafik
However it is totally unexpected and confusing to a user, as the page is about configuring confirmations, and not underlying settings.

## Expected behaviour

* The existing checkbox is moved to another settings page. Proposed placement "General" page under "Performance" group, after "Show submodules status in browser menu" item.
* Add a new `DontConfirmUpdateSubmodulesOnCheckout` setting that will guard against showing confirmations
* Update all code paths to correctly use these settings

manishkungwani:
Fixes #7290
- Moved the "Update submodules on checkout" setting from Confirmation page to General page (Behaviour group).

## Proposed changes

- [x] Added the control to General page
- [x] Added the logic to General page
- [x] Removed the control to Confirmations page
- [x] Removed the logic from Confirmations page
- [x] Ensured that TabOrder is correct

## Screenshots

image

image

## Test methodology 

- Verified the UI
- Tested the settings are not adversely affected by the change in UI.

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ConfirmationsSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ConfirmationsSettingsPage.Designer.cs
@@ -31,13 +31,6 @@
             this.tlpnlMain = new System.Windows.Forms.TableLayoutPanel();
             this.gbConfirmations = new System.Windows.Forms.GroupBox();
             this.tlpnlConfirmations = new System.Windows.Forms.TableLayoutPanel();
-            this.chkFetchAndPruneAllConfirmation = new System.Windows.Forms.CheckBox();
-            this.chkSwitchWorktree = new System.Windows.Forms.CheckBox();
-            this.chkUndoLastCommitConfirmation = new System.Windows.Forms.CheckBox();
-            this.chkRebaseOnTopOfSelectedCommit = new System.Windows.Forms.CheckBox();
-            this.chkSecondAbortConfirmation = new System.Windows.Forms.CheckBox();
-            this.chkCommitAfterConflictsResolved = new System.Windows.Forms.CheckBox();
-            this.chkResolveConflicts = new System.Windows.Forms.CheckBox();
             this.chkAmend = new System.Windows.Forms.CheckBox();
             this.chkCommitIfNoBranch = new System.Windows.Forms.CheckBox();
             this.chkAutoPopStashAfterPull = new System.Windows.Forms.CheckBox();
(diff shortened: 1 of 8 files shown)


gerhardol:
As seen on the build page:
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/28101391
Please update English translations and re-submit the pull-request.
For more information please refer to https://github.com/gitextensions/gitextensions/wiki/Translations

Is not the confirmation for the dialog still wanted?
(the settings will be split now, hard to explain that the Behaviour and Confirmation settings are related).

manishkungwani:
@gerhardol I did update the translations, but it remained the same. This happened, since the control IDs did not change and the translations remained the same as well.

Here is the output of updating translations:

image

RussKie:
Thank you, though few things need further work
 
* We need to add a new `DontConfirmUpdateSubmodulesOnCheckout` setting that will guard against showing confirmations.
The setting will reside in the original placement of "Update submodules on checkout". So in essence, we need to rollback the change to ConfirmationsSettingsPage.cs, rename the checkbox and use the new settings.
* This new setting needs to be respected when showing confirmation box.

Please let me know if you have questions.

manishkungwani:
@RussKie the current setting also does a similar function, so pl confirm if this is correct:
- **UpdateSubmodulesOnCheckout** - Its state will define whether the Submodules are updated on Checkout
- If _True_, Submodules will be updated on Checkout
- If _False_, upon Checkout, we shall check the new setting **DontConfirmUpdateSubmodulesOnCheckout** and show a confirmation, if **DontConfirmUpdateSubmodulesOnCheckout** is _True_
- If **DontConfirmUpdateSubmodulesOnCheckout** is _False_, the Checkout process won't be interrupted

RussKie:
Yes, that is correct - 2 settings: one to toggle the functionality, and another to toggle confirmations.
Thanks

gerhardol:
>  I hope this clarifies the confirmation comment from you as well.

It should be OK now, I have not looked at the implementation

manishkungwani:
@RussKie pl review.

GitUI/CommandsDialogs/SettingsDialog/Pages/ConfirmationsSettingsPage.Designer.cs
```diff
@@ -31,13 +31,6 @@ private void InitializeComponent()
             this.tlpnlMain = new System.Windows.Forms.TableLayoutPanel();
             this.gbConfirmations = new System.Windows.Forms.GroupBox();
             this.tlpnlConfirmations = new System.Windows.Forms.TableLayoutPanel();
-            this.chkFetchAndPruneAllConfirmation = new System.Windows.Forms.CheckBox();
```
RussKie: Could you please clarify what changed here?

GitUI/MessageBoxes.cs
```diff
@@ -89,6 +89,7 @@ public static bool ConfirmUpdateSubmodules(IWin32Window win)
 
             if (PSTaskDialog.cTaskDialog.VerificationChecked)
             {
+                AppSettings.DontConfirmUpdateSubmodulesOnCheckout = result;
```
RussKie: Why are these two updated to the same value?

manishkungwani:
@RussKie do you need anything else from me on this?

manishkungwani:
@RussKie can you pl confirm if this needs any more work.

Thank you

RussKie:
Please resolve the MC and squash.

RussKie:
@msftbot merge in 12 hours

PR #7329, opened 2019-10-18
Improve revisiongrid tooltips

### LINKED ISSUE #7262: Long commit text freeze the GUI for seconds when mouse over a commit
In my company the merge into the master happens from weekly collection branches therefore this commit messages contains several ticket/feature number, name, author name, test validation information, therefore unfortunately occasionally it means more thousand character commit messages generated by scripts (without line breaks). When I use the gitextension GUI and the commit text contains too long or multi line information it pops up the whole text.
The problem starts when I move the mouse over one of this long one line text commits. In this case the GUI starts to use one CPU core on 100% and takes a few seconds till it can draw out the pop up box and the commit text, and the GUI freeze till this pop up displayed successfully. It's really annoying when I had to find an older commit and during scrolling down several times the GUI freezes for seconds.

## Current behaviour

In case of long commit messages without line breaks (few thousand character) the GUI freeze for a few seconds when starts to present the popup box with the commit message.
May be the problem should be when the code tries to break the text into new lines as when I tried the same text with line breaks the GUI didn't freezed.

## Expected behaviour

The GUI handle those long texts without freeze or as an workaround an option would be great to disable this pop up commit message display on mouse over event.

## Steps to reproduce

-Create a commit with a few thousand character in the commit message.
-Move the cursor on and over the commit in the GUI.

## Screenshots
git_freeze1
git_freeze2

If I know this feature were introduced after the 3.x version (but I'm not sure) and with the 2.51.x I don't remember similar problems.

## Environment

- GitExtensions version: 3.1.1, 3.2.1
- GIT version: 2.22.0
- OS version: Win 7 (windows NT 6.1.7601 SP1)
- .NET version: 4.8.3928.0

pmiossec:
Fixes #7262

## Proposed changes

- Better handling huge commit messages (Fixes #7262) by doing an ellipsis after 150 characters of a line
- Add translation for untranslated strings
- Display same tooltip on avatar column than Author column
- Some perf enhancements:
    * Only display ToolTip if hovering a different cell
    * Filter refs before sorting them
    * better manage StringBuilder

### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build c4683fe7926ee9f74699b56c200575f8a1d4cc83 (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4018.0
- DPI 192dpi (200% scaling)


GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +82,77 @@ public BuildInfo BuildStatus
...
+                }
+
+                return _bodySummary ?? (_bodySummary = BuildSummary(Body.TrimEnd()));
+            }
+        }
+
+        private const int MaxLineLength = 150;
+        private static string BuildSummary(string body)
+        {
+            var s = new StringBuilder(100);
```
RussKie: Do I understand correctly we expect to render max ~4750-4800 chars?
```cs
int worstCaseLength = (/*max line length*/150 + /* [...]*/6 + /*CR*/1) * 30 + /*last [...]CR*/7 + /*CR*/1;
```

Should we then allocate like this?
```suggestion
            var s = new StringBuilder(Math.Min(body.Length, worstCaseLength) + /*branch name*/???);
```

GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +82,77 @@ public BuildInfo BuildStatus
...
+                        AppendLine(MaxLineLength, withEllipsis: true);
+                    }
+                    else
+                    {
+                        AppendLine(pos - lineStartPos);
+                    }
+
+                    lineStartPos = pos + 1;
+
+                    if (++lineCount == 30)
```
RussKie: Please define a const, so we don't have random magic numbers throughout the code
Applies to other instances as well

UnitTests/GitCommandsTests/GitRevisionTests.cs
```diff
@@ -19,5 +20,34 @@ public void Should_validate_full_sha1_correctly()
...
+            Assert.AreEqual(bodyContent, gitRevision.BodySummary);
+        }
+
+        [Test]
+        public void Should_do_ellipsis_When_too_many_lines()
+        {
+            var gitRevision = new GitRevision(ObjectId.Random());
+            gitRevision.Body = "line1\nline2\nline3\nline4\nline5\nline6\nline7\nline8\nline9\nline10\nline11\nline12\nline13\nline14\nline15\nline16\nline17\nline18\nline19\nline20\nline21\nline22\nline23\nline24\nline25\nline26\nline27\nline28\nline29\nline30\nline31";
+            gitRevision.HasMultiLineMessage = true;
+            Assert.AreEqual("line1\nline2\nline3\nline4\nline5\nline6\nline7\nline8\nline9\nline10\nline11\nline12\nline13\nline14\nline15\nline16\nline17\nline18\nline19\nline20\nline21\nline22\nline23\nline24\nline25\nline26\nline27\nline28\nline29\nline30\n[...]", gitRevision.BodySummary);
```
RussKie: Let's convert these test (that deal with blobs of text) into using ApprovalTests.
It would make it easier to generate blobs of texts and visually assert results.

Have a look at https://github.com/gitextensions/gitextensions/blob/94621eaac6ed91aae3cba072361710c3326823ed/UnitTests/GitUITests/Editor/FileViewerTextTests.cs#L47-L66

GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
```diff
@@ -289,38 +290,41 @@ public override void OnCellFormatting(DataGridViewCellFormattingEventArgs e, Git
...
                     {
                         if (gitRef.IsBisectGood)
                         {
-                            s.AppendLine("Marked as good in bisect");
+                            _toolTipBuiler.AppendLine("Marked as good in bisect");
                         }
                         else if (gitRef.IsBisectBad)
                         {
-                            s.AppendLine("Marked as bad in bisect");
+                            _toolTipBuiler.AppendLine("Marked as bad in bisect");
```
RussKie: ^^ Can we LOC these strings?

GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
```diff
@@ -289,38 +290,41 @@ public override void OnCellFormatting(DataGridViewCellFormattingEventArgs e, Git
 
         public override bool TryGetToolTip(DataGridViewCellMouseEventArgs e, GitRevision revision, out string toolTip)
         {
+            _toolTipBuiler.Clear();
```
RussKie: It looks like we have similar implementations in multiple places (`GitRevision` and `MessageColumnProvider`).
Can we extract the functionality into a `ToolTipBuilder` class that deals with all nuances of building a tooltip?

GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
```diff
@@ -14,6 +14,7 @@ namespace GitUI.UserControls.RevisionGrid.Columns
     internal sealed class MessageColumnProvider : ColumnProvider
     {
         public const int MaxSuperprojectRefs = 4;
+        private StringBuilder _toolTipBuiler = new StringBuilder(200);
```
mstv:
```suggestion
        private StringBuilder _toolTipBuilder = new StringBuilder(200);
```

GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +84,76 @@ public BuildInfo BuildStatus
...
+        public string BodySummary
+        {
+            get
+            {
+                if (string.IsNullOrWhiteSpace(Body) || !HasMultiLineMessage)
+                {
+                    return Body;
+                }
+
+                return _bodySummary ?? (_bodySummary = BuildSummary(Body.TrimEnd()));
```
drewnoakes: Is there a good reason to cache this field once computed? My intuition is that this shouldn't be cached as it's unlikely to be reused, and requires a reference on every single git revision in memory. Put another way, the increase in size to `GitRevision` for repos with many commits seems 'worse' overall than the cost of allocating and cleaning up a few temporary strings here and there, especially when you consider that in most cases the body summary is just the body unchanged (if I understand correctly).

GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +84,76 @@ public BuildInfo BuildStatus
...
+            }
+        }
+
+        private static string BuildSummary(string body)
```
drewnoakes: In this method, if the `body` does not need to be changed (which I assume is the majority case) can you avoid the extra work and allocation by detecting this and just returning body?

GitCommands/Git/GitRevision.cs
pmiossec: I imagine that if you needed once to display the tool tip, there is a high probability that you need it another time.

GitCommands/Git/GitRevision.cs
pmiossec: Avoiding the extra cost will be difficult (and I don't worth the effort because most of the time cpu is spent during display of the tooltip).

But avoiding the allocation could be done easily. And it's a good idea.

But I think we shouldn't be afraid to keep the value in memory because they are garbage collected when the revision grid is refreshed, I think.

GitCommands/Git/GitRevision.cs
drewnoakes: I doubt that, but don't have any data to back it up. Again, the overhead of recomputing the summary is trivial, but the cost of adding this field to all revisions is non-trivial.

GitCommands/Git/GitRevision.cs
drewnoakes: If you have 1,000 commits, you will on average only populate a handful of these summary properties. That's a lot of null values, taking up 8 bytes each on a 64-bit machine. Reducing the compactness of objects has knock on effects to cache efficiency and the like, when iterating through revisions. These objects should be as small as possible.

gerhardol: Should there be a separate tooltip cache?

drewnoakes: I don't think there's any need to cache the tooltips. There are very few of them, they are quick to create, are created slowly (in response to user actions) and when you consider that they're just not large enough to worry about keeping around. The rest of the PR is good -- I just wouldn't worry about caching.

GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +83,76 @@ public BuildInfo BuildStatus
...
+                    else
+                    {
+                        AppendLine(pos - lineStartPos);
+                    }
+
+                    lineStartPos = pos + 1;
+
+                    if (++lineCount == CommitSummaryMaxNumberOfLines)
+                    {
+                        return s.Append("[...]").ToString();
```
gerhardol: This should be ellipsis …
(similar in other parts)
Inconsistent use in GE though

pmiossec:
>Ready to merge?

The GE code is indeed ready.
But @RussKie ask me to convert some tests to use Approval.Net instead. So, I need to do it but I have not found the time to do it these last days 😢

GitCommands/Git/GitRevision.cs
pmiossec: That's what I have done first but I didn't found it enough clear. Even when developing the feature and didn't understood immediately that it was added by GE.
Maybe because the user could also use  '..;` in it's commit message.

I could change it because that's not a big deal for me but before that I would like the point of view of others... /cc @RussKie @mstv @drewnoakes @Balucsek

RussKie: > This should be ellipsis …

Could you please point to examples?



> I would like the point of view of others...

If we do so in other parts (especially in high traffic dialogs), we should be consistent

GitCommands/Git/GitRevision.cs
RussKie: This is still unanswered

UnitTests/GitCommandsTests/GitRevisionTests.cs
```diff
@@ -19,5 +26,31 @@ public void Should_validate_full_sha1_correctly()
...
+            Assert.AreEqual(bodyContent, gitRevision.BodySummary);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Test]
+        [TestCase("Too_many_lines")]
+        [TestCase("Too_long_lines")]
+        public void Should_do_ellipsis(string testName)
+        {
+            var content = EmbeddedResourceLoader.Load(Assembly.GetExecutingAssembly(), $"{GetType().Namespace}.MockData.{testName}.bin");
```
RussKie: [nit] this isn't a binary file, but a text file, so .txt is probably more appropriate

pmiossec: that's what I made at first but we use `.bin` elsewhere to avoid git manage eol (cf our `.gitattributes` config).

So I was not sure what was the best choice. Even if the tests should works with every line endings...

GitCommands/Git/GitRevision.cs
pmiossec: If we want to optimize that, I'm also tempted to create only static instance of `StringBuilder` and surround the usage with a `lock() {...}`.

What do you think?

pmiossec: Because thinking about it again, allocating the worst case length is **very** pessimistic even when the body is more long than this limit.

GitCommands/Git/GitRevision.cs
mstv: I agree to use `…` for consistency. Though I am afraid that this as a single glyph would be hard to notice.
"[…]", although actually "[...]" is more similar to the multi-line body indicator grafik in the message column of the revision grid.

gerhardol: … was used in CommitInfo until recently (there are remains checking for … in LinkFactory)
It is used in the repository list still

Other shortening is using three dots

There was a discussion about this a year ago or so

UnitTests/GitCommandsTests/GitRevisionTests.cs
RussKie: ApprovalTests no longer care about EOL in *.approved.* files.

GitCommands/Git/GitRevision.cs
RussKie: In that case, it probably doesn't matter.
I don't have a strong preference, so let's go with whatever we have implemented. We can unify to one format later.

GitCommands/Git/GitRevision.cs
RussKie: > I'm also tempted to create only static instance of `StringBuilder` and surround the usage with a `lock() {...}`.

`GitRevision` isn't a singleton, we have multiple of those floating around. Introducing a singleton here will lead to race conditions and a may become a bottleneck.

pmiossec: done.

GitCommands/Git/GitRevision.cs
mstv: Since nothing `GitRevision`-specific is done here anymore, this function should be moved to a new class.
We should not bloat this class with displaying stuff.
Then this function can be used to truncate all tooltips in `RevisionGridToolTipProvider` again.

GitCommands/Git/GitRevision.cs
```diff
@@ -26,6 +27,9 @@ public sealed class GitRevision : IGitItem, INotifyPropertyChanged
         public static readonly Regex Sha1HashShortRegex = new Regex(@"\b[a-f\d]{7,40}\b", RegexOptions.Compiled);
 
         private BuildInfo _buildStatus;
+        private const int CommitSummaryMaxLineLength = 150;
+        private const int CommitSummaryMaxNumberOfLines = 30;
+        private const int CommitSummaryWorstCaseLength = (CommitSummaryMaxNumberOfLines * (CommitSummaryMaxLineLength + 7)) + 5;
```
RussKie: Please add a comment to explain the magic numbers.

GitCommands/Git/GitRevision.cs
RussKie: @mstv 👍

GitCommands/Git/GitRevision.cs
```diff
@@ -86,6 +84,76 @@ public BuildInfo BuildStatus
...
         public string Body { get; set; }
+
+        [CanBeNull]
+        public string BodySummary
+        {
+            get
+            {
+                if (string.IsNullOrWhiteSpace(Body) || !HasMultiLineMessage)
+                {
+                    return Body;
```
RussKie: What would happen if `Body` is a single line but a very long line? Won't we need to trim it?

GitCommands/Git/GitRevision.cs
pmiossec: done

GitCommands/Git/GitRevision.cs
RussKie: I pushed a refactor in which I moved `GitRevision.BuildSummary(string)` method out to its own class.
I updated callsites, but tests need to be updated to reflect the new code structure.

pmiossec:
@RussKie I finally took the time to fix the failing tests due to the last refactoring...

pmiossec:
Updated to be mergeable (following csproj files conversion).
We should not forget to merge it as it seems to be awaited (See the not very friendly remainder: https://github.com/gitextensions/gitextensions/issues/7262#issuecomment-573357783 )

pmiossec:
>Can we squash or do you want to keep it as-is?

And would like to keep most of the commits as they are not all linked and fix different problems.
Some could have been squashed but I'm not sure it possible easily and I'm not sure it worth the effort.

That's why I will merge it like that....

PR #7330, opened 2019-10-18
Consistent display of binary files

### LINKED ISSUE #7322: Binary file data view regression
## Current behaviour

a3420a28fae2afc1ef0ae10962adcebd01613875 seems to have regressed the display of binary data in the file viewer.

## Expected behaviour

Binary hex dump still displays.

This problem was discovered while browsing code.

gerhardol:
Fixes #7322
Well, does not fix what #7322 was reported about, but the scenario why #7322 was reported.

## Proposed changes

If a file is detected as binary from gitattributes or filename, the contents should be displayed as a hexdump similar to files detected as binary from file contents. This affects the FileTree, edit file etc.
(editing binary files will still be an issue, it will save the data as displayed...)

### Before

Ignore the dark theme, just see that the file contents are not displayed.

image

### After

image

## Test methodology 

I setup the following repo for testing, the GE binary files did not have the "correct" attributes set
https://github.com/gerhardol/tmp_test.git

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -562,12 +562,16 @@ namespace GitUI.Editor
                 patchText => ViewPatch(patchText.text, patchText.openWithDifftool));
         }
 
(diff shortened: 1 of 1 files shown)


gerhardol:
@msftbot merge after 24 hours

Issue #6434, opened 2019-04-04
Hex View on Blame Tab Always Shows Working Copy instead of Selected Revision


BlackStarX42:
## Current behaviour
For a binary file, the hex view on the blame tab does not show the version which is selected in the revision graph and instead always shows the version in the working copy.
## Expected behaviour
The hex view should show the revision selected in the revision graph.
## Steps to reproduce
- Create a file which is considered a binary file by git (e.g. a UTF-16 file with BOM), it must be at least 14 bytes to show up as a hex view on the blame tab (which seems to be another bug to me),
- commit the file to the repository,
- add something to the file,
- open the history of the file and go to the blame tab, you will see that the hex view shows the recent change you made to the file, though it is not even committed yet.
## Did this work in previous version of GitExtensions
Don't know.
## Environment
- Git Extensions 3.0.2.5232
- Build 23b6f51905006ccdde8701591df706284b4155dc
- Git 2.20.1.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

gerhardol:
Same in View tab

mstv:
> it must be at least 14 bytes to show up as a hex view on the blame tab (which seems to be another bug to me)

GE uses an heuristic: A file must contain at least 6 zero bytes to be displayed as binary. See https://github.com/gitextensions/gitextensions/blob/a4c5d3c30aac7c73441994ad0a85d5e59b08f1f4/GitCommands/FileHelper.cs#L165-L185

BlackStarX42:
> > it must be at least 14 bytes to show up as a hex view on the blame tab (which seems to be another bug to me)
> 
> GE uses an heuristic: A file must contain at least 6 zero bytes to be displayed as binary. See 

I wasn't aware of that. I would have expected that GE asks git for the file type. Isn't that possible and wouldn't that make more sense?

pmiossec:
Git use an heuristic, too. But not exactly the same: https://stackoverflow.com/a/54182883/717372  (And I just understood that's the reason why Git always detect UTF16 files as binary)

> I would have expected that GE asks git for the file type. Isn't that possible and wouldn't that make more sense?

Probably (but surely will be a lot less performant)

A solution (but seems not working with git-lfs where binary files are seen as text from what I have tested): 
https://www.closedinterval.com/determine-if-a-file-is-binary-using-git/

[...]

PR #7333, opened 2019-10-19
Truncate large hexdumps

gerhardol:
Part of #7322

## Proposed changes
Displaying hexdumps when viewing binary files (as well as when editing them...) is a nice touch to give an indication to the contents of a file in the repo that is not in text format. (In general though git-lfs should be used for binary files, do not store them in Git.)
However, the complete file is not interesting.
If you select a larger file like .nuget/NuGet.exe (5MB), it takes over 40s to display the file for me, with the spinner running. It seems like GE freezes.

This PR limits the display to the first 16384 bytes (1024 lines), which should be sufficient to display relevant information.
No configuration, save the file and use a dedicated hexeditor to see the complete file.

This cannot be an important feature as hexdump display has not been working properly (ever?), #6434, #7330 
This will conflict with #7330, will rebase after #7330 is merged and merge 24 after approval.

### After
image

## Test methodology 

View for instance .nuget/Nuget

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -619,9 +619,13 @@ namespace GitUI.Editor
                 return "";
             }
 
+            // Do not freeze GE when selecting large binary files
+            // Show only the header of the binary file to indicate contents and files incorrectly handled
+            // Use a dedicated editor to view the complete file
+            var limit = Math.Min(bytes.Length, columnWidth * columnCount * 256);
             var i = 0;
 
-            while (i < bytes.Length)
+            while (i < limit)
             {
                 var baseIndex = i;
 
@@ -683,6 +687,12 @@ namespace GitUI.Editor
                         i++;
                     }
                 }
+           }
+
+            if (bytes.Length > limit)
+            {
+                str.AppendLine();
+                str.Append("[Truncated]");
             }
 
             return str.ToString();


GitUI/Editor/FileViewer.cs
```diff
@@ -608,7 +608,7 @@ private FileInfo GetFileInfo(string fileName)
             return new FileInfo(resolvedPath);
         }
 
-        public static string ToHexDump(byte[] bytes, StringBuilder str, int columnWidth = 8, int columnCount = 2)
+        public static string ToHexDump(byte[] bytes, StringBuilder str, int columnWidth = 8, int columnCount = 2, int maxBytes = 8 * 2 * 1024)
```
RussKie: How did you decide on 1024 lines?
Is there a reason to show, say, more than 50 or 100 lines?

GitUI/Editor/FileViewer.cs
```diff
@@ -617,7 +617,7 @@ public static string ToHexDump(byte[] bytes, StringBuilder str, int columnWidth
 
             var i = 0;
 
-            while (i < bytes.Length)
+            while (i < bytes.Length && i < maxBytes)
```
RussKie:
```suggestion
            var limit = Math.Min(bytes.Length, maxBytes);
            while (i < limit)
```

GitUI/Editor/FileViewer.cs
gerhardol: 1024 bytes has no noticeable delay and is scroll able. 
As this is a hex dump it is reasonable with a binary number.
256 lines (0x1000 bytes) is another candidate.

RussKie:
I can't think how 1,000 of lines of hex would be of significant use to
anyone.
I suggest we do less

GitUI/Editor/FileViewer.cs
gerhardol: Will change
(I did implement like this at first, but reverted to min number of lines. )

GitUI/Editor/FileViewer.cs
gerhardol: > I can't think how 1,000 of lines of hex would be of significant use to
anyone.
I suggest we do less

It could also be that GE detects a binary file incorrectly (a file ending used for other intentions than anticipated?), this would be a reserver 16KB files are just  some hundreds of lines of code.
This is not important so I changing this to 4KB, commenting better and removing the unused parameter.

PR #7334, opened 2019-10-19
Add icon to all MessageBox that don't have one

### LINKED ISSUE #7312: Add icon in all MessageBox
## Current behaviour

There is lot's of ``MessageBox.Show`` that don't have icon. The message box looks raw and not user friendly.

## Expected behaviour

All MessageBox shall have an icon.

## Steps to reproduce

Start | Create new repository

Create a Personal reposiroty

## Screenshots

2019_10_16_08_45_16_Create_new_repository

No

## Environment

- GitExtensions version: 3.2.1.6628
- GIT version: 2.21.0.windows.1
- OS version: NT 10.0 18362.0
- .NET version: 4.8.4018.0

## Personnal comment

I'm fine to do the work if the team approves the change.

bansan85:
Fixes #7312

## Proposed changes

- Add icon to all MessageBox that miss icon
- When MessageBox has button Yes/No/Cancel, it's always a question even if the message appears in case of error
- To add icon, I need to add a title to MessageBox. This new title is not translated and is "Error" or "Information"
- There's some MessageBox that just show return string from external application. The message could be normal message or error message. Because I don't know if it's an error or a normal message, I put a Information icon.
- I really tried MessageBox only on two cases where I doubt if it should be Information or Error. For the other 99% MessageBox, I decided Icon message only by reading source code.

### DIFF
--- a/GitExtensionsVSIX/Git/GitCommands.cs
+++ b/GitExtensionsVSIX/Git/GitCommands.cs
@@ -25,7 +25,7 @@ namespace GitExtensionsVSIX.Git
             }
             catch (UnauthorizedAccessException)
             {
-                MessageBox.Show("Git Extensions has insufficient permissions to check the registry.");
+                MessageBox.Show("Git Extensions has insufficient permissions to check the registry.", "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
             }
 
             return "";
@@ -92,7 +92,7 @@ namespace GitExtensionsVSIX.Git
             {
                 if (!File.Exists(Path.Combine(path, "GitExtensions.exe")))
                 {
-                    MessageBox.Show("This plugin requires Git Extensions to be installed. This application can be downloaded from http://gitextensions.github.io/");
+                    MessageBox.Show("This plugin requires Git Extensions to be installed. This application can be downloaded from http://gitextensions.github.io/", "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
                 }
 
                 return null;
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -750,7 +750,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
             }
             catch (Exception ex)
             {
-                MessageBox.Show(this, ex.Message);
+                MessageBox.Show(this, ex.Message, "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
             }
         }
 
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -899,7 +899,7 @@ namespace GitUI.CommandsDialogs
         {
             if (!string.IsNullOrEmpty(output))
             {
-                MessageBox.Show(this, output + "\n\n" + SelectedDiff.Encoding.GetString(patch));
+                MessageBox.Show(this, output + "\n\n" + SelectedDiff.Encoding.GetString(patch), "Information", MessageBoxButtons.OK, MessageBoxIcon.Information);
             }
 
             if (_currentItemStaged)
@@ -1454,7 +1454,7 @@ namespace GitUI.CommandsDialogs
                 }
                 catch (Exception e)
                 {
-                    MessageBox.Show(this, $"Exception: {e.Message}");
+                    MessageBox.Show(this, $"Exception: {e.Message}", "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
                 }
 
                 return;
@@ -2090,12 +2090,12 @@ namespace GitUI.CommandsDialogs
 
                 if (filesInUse.Count > 0)
                 {
-                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use");
+                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use", MessageBoxButtons.OK, MessageBoxIcon.Information);
                 }
 
                 if (!string.IsNullOrEmpty(output.ToString()))
                 {
-                    MessageBox.Show(this, output.ToString(), _resetChangesCaption.Text);
+                    MessageBox.Show(this, output.ToString(), _resetChangesCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Information);
                 }
             }
             finally
@@ -2111,7 +2111,7 @@ namespace GitUI.CommandsDialogs
             try
             {
                 if (Unstaged.SelectedItem == null ||
-                    MessageBox.Show(this, _deleteSelectedFiles.Text, _deleteSelectedFilesCaption.Text, MessageBoxButtons.YesNo) !=
+                    MessageBox.Show(this, _deleteSelectedFiles.Text, _deleteSelectedFilesCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) !=
                     DialogResult.Yes)
                 {
                     return;
@@ -2129,7 +2129,7 @@ namespace GitUI.CommandsDialogs
             }
             catch (Exception ex)
             {
-                MessageBox.Show(this, _deleteFailed.Text + Environment.NewLine + ex.Message);
+                MessageBox.Show(this, _deleteFailed.Text + Environment.NewLine + ex.Message, "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
             }
         }
 
@@ -2143,7 +2143,7 @@ namespace GitUI.CommandsDialogs
 
         private void DeleteSelectedFilesToolStripMenuItemClick(object sender, EventArgs e)
         {
-            if (MessageBox.Show(this, _deleteSelectedFiles.Text, _deleteSelectedFilesCaption.Text, MessageBoxButtons.YesNo) !=
+            if (MessageBox.Show(this, _deleteSelectedFiles.Text, _deleteSelectedFilesCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) !=
                 DialogResult.Yes)
             {
                 return;
@@ -2158,7 +2158,7 @@ namespace GitUI.CommandsDialogs
             }
             catch (Exception ex)
             {
-                MessageBox.Show(this, _deleteFailed.Text + Environment.NewLine + ex);
+                MessageBox.Show(this, _deleteFailed.Text + Environment.NewLine + ex, "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
             }
 
             Initialize();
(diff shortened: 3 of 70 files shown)


GitExtensionsVSIX/Git/GitCommands.cs
```diff
@@ -25,7 +25,7 @@ private static string GetRegistryValue(RegistryKey root, string subkey, string k
             }
             catch (UnauthorizedAccessException)
             {
-                MessageBox.Show("Git Extensions has insufficient permissions to check the registry.");
+                MessageBox.Show("Git Extensions has insufficient permissions to check the registry.", "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
```
gerhardol: There is a "Error" string in MessageBoxes.cs
You should probably add the string to Resources instead

bansan85: The main message is not translated so I don't think it's necessary. But what should I do for translated message ? Had a "Error" translated ? Or add a more specific message like "Insufficient permissions" for this example ?

bansan85:
> Can you please add examples?

What kind of examples ? Screenshot ?

gerhardol:
> What kind of examples ? Screenshot ?

Yes

GitExtensionsVSIX/Git/GitCommands.cs
RussKie: @gerhardol this is a VSIX project, we don't have translations for it nor mechanism to switch languages

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -750,7 +750,7 @@ private void tsmiOpenFolder_Click(object sender, EventArgs e)
             }
             catch (Exception ex)
             {
-                MessageBox.Show(this, ex.Message);
+                MessageBox.Show(this, ex.Message, "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
```
RussKie: 📝  Here and other places the heading must be localised
Please put "Error" into https://github.com/gitextensions/gitextensions/blob/master/GitUI/Strings.cs

GitUI/CommandsDialogs/FormMailMap.cs
```diff
@@ -138,7 +138,7 @@ private void FormMailMapLoad(object sender, EventArgs e)
                 return;
             }
 
-            MessageBox.Show(this, _mailmapOnlyInWorkingDirSupported.Text, _mailmapOnlyInWorkingDirSupportedCaption.Text);
+            MessageBox.Show(this, _mailmapOnlyInWorkingDirSupported.Text, _mailmapOnlyInWorkingDirSupportedCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: I think this should be consistent with GitUI/CommandsDialogs/FormGitIgnore.cs line:197 - which is an error.
Or at least a warning, not an info

GitUI/CommandsDialogs/FormGitAttributes.cs
```diff
@@ -136,7 +136,7 @@ private void FormGitAttributesLoad(object sender, EventArgs e)
                 return;
             }
 
-            MessageBox.Show(this, _noWorkingDir.Text, _noWorkingDirCaption.Text);
+            MessageBox.Show(this, _noWorkingDir.Text, _noWorkingDirCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: ditto

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2560,7 +2561,7 @@ private void OpenFilesWithDiffTool(IEnumerable<GitItemStatus> items, string firs
                 string output = Module.OpenWithDifftool(item.Name, null, firstRevision, secondRevision, "", item.IsTracked);
                 if (!string.IsNullOrEmpty(output))
                 {
-                    MessageBox.Show(this, output);
+                    MessageBox.Show(this, output, "Information", MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: I think this should be an error:
https://github.com/gitextensions/gitextensions/blob/master/GitCommands/Git/GitModule.cs#L3561-L3563

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2062,12 +2062,12 @@ private void ResetSoftClick(object sender, EventArgs e)
...
                 if (filesInUse.Count > 0)
                 {
-                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use");
+                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use", MessageBoxButtons.OK, MessageBoxIcon.Information);
                 }
 
                 if (!string.IsNullOrEmpty(output.ToString()))
                 {
-                    MessageBox.Show(this, output.ToString(), _resetChangesCaption.Text);
+                    MessageBox.Show(this, output.ToString(), _resetChangesCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: if `output` is not empty, it means there has been warnings/errors during the execution of the git command, so it arguably this must be an error.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2062,12 +2062,12 @@ private void ResetSoftClick(object sender, EventArgs e)
 
                 if (filesInUse.Count > 0)
                 {
-                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use");
+                    MessageBox.Show(this, "The following files are currently in use and will not be reset:" + Environment.NewLine + "\u2022 " + string.Join(Environment.NewLine + "\u2022 ", filesInUse), "Files In Use", MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: This must be a warning

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -885,7 +885,7 @@ private void ProcessApplyOutput(string output, byte[] patch)
         {
             if (!string.IsNullOrEmpty(output))
             {
-                MessageBox.Show(this, output + "\n\n" + SelectedDiff.Encoding.GetString(patch));
+                MessageBox.Show(this, output + "\n\n" + SelectedDiff.Encoding.GetString(patch), "Information", MessageBoxButtons.OK, MessageBoxIcon.Information);
```
RussKie: if `output` is not empty, it means there has been warnings/errors during the execution of the git command, so it arguably this must be an error.

GitExtensionsVSIX/Git/GitCommands.cs
gerhardol: This was a general comment for all - I missed that this specific was vsix. But RussKie also commented on that.

GitUI/CommandsDialogs/FormCommit.cs
bansan85: Right. I should have open Module.OpenWithDifftool.

GitUI/CommandsDialogs/FormMailMap.cs
bansan85: I think I will use error.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
bansan85: Fine. I noticed that never "Error" is used but always a very short custom message like "Host Fingerprint not registered" for a full message "The fingerprint of this host is not registered by PuTTY.". Should I keep this rule or is it fine to use "Error" which is redundant with the icon ?

The title is a real pain. It always redundant with the message. I'm happy the title has been remove in GTK in Linux.

bansan85:
@RussKie : thanks for you review :)

bansan85:
I added an icon in all MessageBox.Show :) I hope I didn't made to much mistakes. But there's still works to do.

I have some questions to continue the  job :

Some MessageBox are question. I noticed some have warning icon. What kind of rule should I use ? Always use question icon ? Use warning when action can't be undo and question either ? Example of warning question : "Cannot find Gource in the configured path: {0}.\n\nDo you want to reset the configured path?"

For question MessageBox : what's the default button ? No/Cancel if Warning and Yes either ? Or No/Cancel always ?

In lots of MessageBox, the caption is not "Error" / "Information" but a very short message (ie. "Rebase failed"). Should I do the same ? It will need lots of work for translators (and also for me to understand the context of all message).

Should I also translate all hardcoded text in existing MessageBox ?

Any other idea ?

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
RussKie: I don't think it is redundant, and quite likely it is used for accessibility reasons.
As a minimum I suggest we translate all outstanding existing headings. Later on we can look at providing better wording.

GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
```diff
@@ -710,7 +710,7 @@ public static string GetXHTMLText(this RichTextBox rtb, bool bParaFormat)
             }
             catch (Exception /*ex*/)
             {
-                ////MessageBox.Show(ex.Message);
+                ////MessageBox.Show(ex.Message, "Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
```
RussKie: This can be removed

GitUI/Editor/RichTextBoxXhtmlSupportExtension.cs
```diff
@@ -1086,7 +1086,7 @@ public static string GetPlainText(this RichTextBox rtb, int from, int to)
...
```
RussKie: ditto

GitUI/FormRemoteProcess.cs
```diff
@@ -15,7 +15,11 @@ public partial class FormRemoteProcess : FormProcess
     {
         #region Translation
         private readonly TranslationString _fingerprintNotRegistredText =
-            new TranslationString("The fingerprint of this host is not registered by PuTTY." + Environment.NewLine + "This causes this process to hang, and that why it is automatically stopped." + Environment.NewLine + Environment.NewLine + "When the connection is opened detached from Git and Git Extensions, the host's fingerprint can be registered." + Environment.NewLine + "You could also manually add the host's fingerprint or run Test Connection from the remotes dialog." + Environment.NewLine + Environment.NewLine + "Do you want to register the host's fingerprint and restart the process?");
+            new TranslationString("The fingerprint of this host is not registered by PuTTY." + Environment.NewLine +
```
RussKie: Make it one verbatim line
```suggestion
           new TranslationString(@"The fingerprint of this host is not registered by PuTTY.
This causes this process to hang, and that why it is automatically stopped.

When the connection is opened detached from Git and Git Extensions, the host's fingerprint can be registered.
You could also manually add the host's fingerprint or run Test Connection from the remotes dialog.

Do you want to register the host's fingerprint and restart the process?");
 ```

RussKie:
IIUIC there are few little outstanding comments, other than that the PR is ready to be merged.
Please rebase as well.

bansan85:
I rebased and modified based on the 3 remaining comments

RussKie:
Thank you

PR #7339, opened 2019-10-20
Fix 7097: Invalid escape character in .gitmodule leads to crash

### LINKED ISSUE #7097: [NBug] Invalid escape character in .gitmodule leads to crash
## Current behaviour

GitExtensions crashes with an error message when an invalid escape character is found in .gitmodule

## Expected behaviour

GitExtensions should warn the user, and make the user correct the file, before attempting to reload it again.

## Steps to reproduce

When you manually edit the path entry of a module in .gitmodules, the path may contain invalid escape characters. Upon saving the file, GitExtensions reloads the file, and crasehs.

## Error Details
```
System.Exception: Invalid escape character: f
   at GitCommands.Config.ConfigFile.ConfigFileParser.ReadValue(Char c)
   at GitCommands.Config.ConfigFile.ConfigFileParser.Parse(String fileContent)
   at GitCommands.Config.ConfigFile..ctor(String fileName, Boolean local)
   at GitCommands.GitModule.GetSubmoduleConfigFile()
   at GitCommands.GitModule.<GetSubmodulesLocalPaths>g__GetSubmodulePaths|72_1(GitModule module)
   at GitCommands.GitModule.<GetSubmodulesLocalPaths>g__DoGetSubmodulesLocalPaths|72_0(GitModule module, String parentPath, List`1 paths, Boolean recurse)
   at GitCommands.GitModule.GetSubmodulesLocalPaths(Boolean recursive)
   at GitCommands.Submodules.SubmoduleStatusProvider.<UpdateSubmodulesStatusAsync>d__27.MoveNext()
   at GitCommands.Submodules.SubmoduleStatusProvider.<>c__DisplayClass17_0.<<UpdateSubmodulesStatus>b__0>d.MoveNext()
```

## Environment
- Git Extensions 3.1.1.6049
- Git 2.18.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

lhiginbotham:
- catch exceptions from the ConfigFileParser and display it using ExceptionUtils to prevent crashes

Fixes #7097

## Proposed changes

- catch exceptions from ConfigFileParser and use ExceptionUtils to display it
- return null ConfigFile reference to client code that relies on GetSubmoduleConfigFile

### Before

image

### After

image

## Test methodology 

- built in Release
- interactive testing to ensure the error message would display to the user and not result in a crash
- reproduced the error pop-up in several worflows, such as the 'commit/staging' window, hitting the refresh button on the main branch-view/commit-log browsing window, and by hitting the diff tab on a commit that updated a submodule

## Test environment(s) 

- GIT version 2.23.0.windows.1
- Windows 10

Sadly, the exceptions displayed do not have any translation capabilities at the moment.

----

### DIFF
--- a/GitCommands/Config/ConfigFile.cs
+++ b/GitCommands/Config/ConfigFile.cs
@@ -33,8 +33,7 @@ namespace GitCommands.Config
             }
             catch (Exception ex)
             {
-                ex.Data.Add(GetType().Name + ".Load", "Could not load config file: " + FileName);
-                throw;
+                throw new GitCommands.Config.GitConfigurationException(fileName, ex);
             }
         }
 
@@ -415,7 +414,6 @@ namespace GitCommands.Config
             {
                 if (_escapedValue)
                 {
-                    // TODO: case 'b' is valid, according to https://git-scm.com/docs/git-config#_syntax
                     switch (c)
                     {
                         case '\\':
--- a/GitCommands/Config/ConfigFile.cs
+++ b/GitCommands/Config/ConfigFile.cs
@@ -415,6 +415,7 @@ namespace GitCommands.Config
             {
                 if (_escapedValue)
                 {
+                    // TODO: case 'b' is valid, according to https://git-scm.com/docs/git-config#_syntax
                     switch (c)
                     {
                         case '\\':
--- a/GitCommands/Config/ConfigFile.cs
+++ b/GitCommands/Config/ConfigFile.cs
@@ -33,7 +33,7 @@ namespace GitCommands.Config
             }
             catch (Exception ex)
             {
-                throw new GitCommands.Config.GitConfigurationException(fileName, ex);
+                throw new GitConfigurationException(fileName, ex);
             }
         }
 
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -431,7 +431,15 @@ namespace GitCommands
         public IReadOnlyList<string> GetSubmodulesLocalPaths(bool recursive = true)
         {
             var localPaths = new List<string>();
-            DoGetSubmodulesLocalPaths(this, "", localPaths, recursive);
+            try
+            {
+                DoGetSubmodulesLocalPaths(this, "", localPaths, recursive);
+            }
+            catch (GitConfigurationException)
+            {
+                // swallow any exceptions here, any config exceptions would have been shown to the user already
+            }
+
             return localPaths;
 
             void DoGetSubmodulesLocalPaths(GitModule module, string parentPath, List<string> paths, bool recurse)
@@ -1208,9 +1216,7 @@ namespace GitCommands
         }
 
         public ConfigFile GetSubmoduleConfigFile()
-        {
-            return new ConfigFile(WorkingDir + ".gitmodules", true);
-        }
+            => new ConfigFile(WorkingDir + ".gitmodules", true);
 
         [CanBeNull]
         public string GetCurrentSubmoduleLocalPath()
@@ -1250,16 +1256,26 @@ namespace GitCommands
 
         public IEnumerable<IGitSubmoduleInfo> GetSubmodulesInfo()
         {
+            ConfigFile configFile;
+            try
+            {
+                configFile = GetSubmoduleConfigFile();
+            }
+            catch (GitConfigurationException)
+            {
+                // swallow any exceptions here, any config exceptions would have been shown to the user already
+                configFile = null;
+            }
+
+            if (configFile == null)
+            {
+                yield return default;
+            }
+
             var args = new GitArgumentBuilder("submodule") { "status" };
             var lines = _gitExecutable.GetOutputLines(args);
 
             string lastLine = null;
-
-            var configFile = GetSubmoduleConfigFile();
-            if (configFile == null)
-            {
-                yield break;
-            }
 
             foreach (var line in lines)
             {
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -454,6 +454,10 @@ namespace GitCommands
             List<string> GetSubmodulePaths(GitModule module)
             {
                 var configFile = module.GetSubmoduleConfigFile();
+                if (configFile == null)
+                {
+                    return new List<string>();
+                }
 
                 return configFile.ConfigSections
                     .Select(section => section.GetValue("path").Trim())
@@ -1208,9 +1212,19 @@ namespace GitCommands
             }
         }
 
+        [CanBeNull]
         public ConfigFile GetSubmoduleConfigFile()
         {
-            return new ConfigFile(WorkingDir + ".gitmodules", true);
+            try
+            {
+                return new ConfigFile(WorkingDir + ".gitmodules", true);
+            }
+            catch (Exception ex)
+            {
+                ExceptionUtils.ShowException(ex, false);
+            }
+
+            return null;
         }
 
         [CanBeNull]
@@ -1257,6 +1271,10 @@ namespace GitCommands
             string lastLine = null;
 
             var configFile = GetSubmoduleConfigFile();
+            if (configFile == null)
+            {
+                yield break;
+            }
 
             foreach (var line in lines)
             {
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -451,17 +451,12 @@ namespace GitCommands
                 }
             }
 
-            List<string> GetSubmodulePaths(GitModule module)
+            IEnumerable<string> GetSubmodulePaths(GitModule module)
             {
-                var configFile = module.GetSubmoduleConfigFile();
-                if (configFile == null)
-                {
-                    return new List<string>();
-                }
+                ConfigFile configFile = module.GetSubmoduleConfigFile();
 
                 return configFile.ConfigSections
-                    .Select(section => section.GetValue("path").Trim())
-                    .ToList();
+                    .Select(section => section.GetValue("path").Trim());
             }
         }
 
@@ -1212,19 +1207,9 @@ namespace GitCommands
             }
         }
 
-        [CanBeNull]
         public ConfigFile GetSubmoduleConfigFile()
         {
-            try
-            {
-                return new ConfigFile(WorkingDir + ".gitmodules", true);
-            }
-            catch (Exception ex)
-            {
-                ExceptionUtils.ShowException(ex, false);
-            }
-
-            return null;
+            return new ConfigFile(WorkingDir + ".gitmodules", true);
         }
 
         [CanBeNull]
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -37,11 +37,8 @@ namespace GitExtensions
             {
                 DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
 
-                if (!Debugger.IsAttached)
-                {
-                    AppDomain.CurrentDomain.UnhandledException += (s, e) => ReportBug((Exception)e.ExceptionObject);
-                    Application.ThreadException += (s, e) => ReportBug(e.Exception);
-                }
+                AppDomain.CurrentDomain.UnhandledException += (s, e) => UncaughtExceptionHandler((Exception)e.ExceptionObject);
+                Application.ThreadException += (s, e) => UncaughtExceptionHandler(e.Exception);
             }
             catch (TypeInitializationException tie)
             {
@@ -323,6 +320,22 @@ namespace GitExtensions
                     {
                         return false;
                     }
+            }
+        }
+
+        private static void UncaughtExceptionHandler(Exception ex)
+        {
+            if (ex is GitCommands.Config.GitConfigurationException)
+            {
+                var configEx = (GitCommands.Config.GitConfigurationException)ex;
+                string message = string.Format(Strings.GeneralGitConfigExceptionMessage, configEx.ConfigPath, Environment.NewLine, configEx.InnerException.Message);
+                MessageBox.Show(message, Strings.GeneralGitConfigExceptionCaption, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                Environment.Exit(-1);
+            }
+
+            if (!Debugger.IsAttached)
+            {
+                ReportBug(ex);
             }
         }
 
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -37,8 +37,11 @@ namespace GitExtensions
             {
                 DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
 
-                AppDomain.CurrentDomain.UnhandledException += (s, e) => UncaughtExceptionHandler((Exception)e.ExceptionObject);
-                Application.ThreadException += (s, e) => UncaughtExceptionHandler(e.Exception);
+                if (!Debugger.IsAttached)
+                {
+                    AppDomain.CurrentDomain.UnhandledException += (s, e) => ReportBug((Exception)e.ExceptionObject);
+                    Application.ThreadException += (s, e) => ReportBug(e.Exception);
+                }
             }
             catch (TypeInitializationException tie)
             {
@@ -320,22 +323,6 @@ namespace GitExtensions
                     {
                         return false;
                     }
-            }
-        }
-
-        private static void UncaughtExceptionHandler(Exception ex)
-        {
-            if (ex is GitCommands.Config.GitConfigurationException)
-            {
-                var configEx = (GitCommands.Config.GitConfigurationException)ex;
-                string message = string.Format(Strings.GeneralGitConfigExceptionMessage, configEx.ConfigPath, Environment.NewLine, configEx.InnerException.Message);
-                MessageBox.Show(message, Strings.GeneralGitConfigExceptionCaption, MessageBoxButtons.OK, MessageBoxIcon.Error);
-                Environment.Exit(-1);
-            }
-
-            if (!Debugger.IsAttached)
-            {
-                ReportBug(ex);
             }
         }
 
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -355,7 +355,8 @@ namespace GitUI.CommandsDialogs
                     }
                 };
 
-                gitStatusMonitor.GitWorkingDirectoryStatusChanged += (s, e) =>
+#pragma warning disable VSTHRD101 // Avoid unsupported async delegates
+                gitStatusMonitor.GitWorkingDirectoryStatusChanged += async (s, e) =>
                 {
                     var status = e.ItemStatuses?.ToList();
 
@@ -396,10 +397,18 @@ namespace GitUI.CommandsDialogs
 
                         if (AppSettings.ShowSubmoduleStatus)
                         {
-                            _submoduleStatusProvider.UpdateSubmodulesStatus(Module.WorkingDir, status);
+                            try
+                            {
+                                await _submoduleStatusProvider.UpdateSubmodulesStatusAsync(Module.WorkingDir, status);
+                            }
+                            catch (GitCommands.Config.GitConfigurationException configEx)
+                            {
+                                DoFatalExit(configEx);
+                            }
                         }
                     }
                 };
+#pragma warning restore VSTHRD101 // Avoid unsupported async delegates
             }
 
             Brush UpdateCommitButtonAndGetBrush(IReadOnlyList<GitItemStatus> status, bool showCount)
@@ -536,7 +545,7 @@ namespace GitUI.CommandsDialogs
                         })
                     .FileAndForget();
             };
-            UpdateSubmodulesStructure();
+            _ = UpdateSubmodulesStructureAsync();
             UpdateStashCount();
 
             toolStripButtonPush.Initialize(_aheadBehindDataProvider);
@@ -619,7 +628,7 @@ namespace GitUI.CommandsDialogs
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
             this.InvokeAsync(RefreshRevisions).FileAndForget();
-            UpdateSubmodulesStructure();
+            _ = UpdateSubmodulesStructureAsync();
             UpdateStashCount();
         }
 
@@ -1636,7 +1645,7 @@ namespace GitUI.CommandsDialogs
         private void ManageSubmodulesToolStripMenuItemClick(object sender, EventArgs e)
         {
             UICommands.StartSubmodulesDialog(this);
-            UpdateSubmodulesStructure();
+            _ = UpdateSubmodulesStructureAsync();
         }
 
         private void UpdateSubmoduleToolStripMenuItemClick(object sender, EventArgs e)
@@ -1653,13 +1662,13 @@ namespace GitUI.CommandsDialogs
         private void UpdateAllSubmodulesToolStripMenuItemClick(object sender, EventArgs e)
         {
             UICommands.StartUpdateSubmodulesDialog(this);
-            UpdateSubmodulesStructure();
+            _ = UpdateSubmodulesStructureAsync();
         }
 
         private void SynchronizeAllSubmodulesToolStripMenuItemClick(object sender, EventArgs e)
         {
             UICommands.StartSyncSubmodulesDialog(this);
-            UpdateSubmodulesStructure();
+            _ = UpdateSubmodulesStructureAsync();
         }
 
         private void ToolStripSplitStashButtonClick(object sender, EventArgs e)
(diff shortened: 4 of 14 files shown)


GitCommands/Config/ConfigFile.cs
```diff
@@ -454,6 +454,7 @@ private ParsePart ReadValue(char c)
             {
                 if (_escapedValue)
                 {
+                    // TODO: case 'b' is valid, according to https://git-scm.com/docs/git-config#_syntax
```
mstv: Should it be handled in this PR? Else a new issue should be opened.

gerhardol:
Seem OK, discussed in issue.

GitCommands/Config/ConfigFile.cs
lhiginbotham: @mstv 
Seems like an unrelated change to me, maybe an issue should be raised.

RussKie: Agree, please remove from this PR

GitCommands/Git/GitModule.cs
```diff
@@ -454,6 +454,10 @@ void DoGetSubmodulesLocalPaths(GitModule module, string parentPath, List<string>
             List<string> GetSubmodulePaths(GitModule module)
```
RussKie: This method needs to be fixed, as it performs unnecessary allocations.
```cs
            IEnumerable<string> GetSubmodulePaths(GitModule module)
            {
                var configFile = module.GetSubmoduleConfigFile();
                if (configFile == null)
                {
                    return Array.Empty<string>();
                }

                return configFile.ConfigSections
                    .Select(section => section.GetValue("path").Trim());
            }
        }
```

RussKie: Also `.ToList` call ont line:441 must be removed.

GitCommands/Git/GitModule.cs
```diff
@@ -1208,9 +1212,19 @@ bool IsTwoSha1Hashes(string parents)
...
         public ConfigFile GetSubmoduleConfigFile()
         {
-            return new ConfigFile(WorkingDir + ".gitmodules", true);
+            try
+            {
+                return new ConfigFile(WorkingDir + ".gitmodules", true);
+            }
+            catch (Exception ex)
+            {
+                ExceptionUtils.ShowException(ex, false);
```
RussKie: Multiple issues here.

1. This is a very poor UX, a user won't know what to do with this.
Nor the team would be in a position to provide a meaningful help.

    image
    
    The error must be concise and clear. Something like:
    ```
    Failed to parse .gitmodules config
    -------------------------------------
    
    An error occurred while parsing <path>\.gitmodules
    
    Please fix the config and reopen the repository.
    ```

2. UI rendering must only be done at the UI layer (i.e. Form or a Control).
The backend must raise exceptions in a way that is useful to callers, e.g. throw a specific exception.

    We can add a new exception (e.g. `GitConfigurationException`) to contain necessary information:
    ```cs
    public class GitConfigurationException : Exception
    {
        public GitConfigurationException(string configPath, Exception inner) : base(null, inner)
        {
            if (string.IsNullOrWhiteSpace(configPath))
            {
                throw new ArgumentException("ConfigPath is required", nameof(configPath));
            }

            ConfigPath = configPath;
        }

        public string ConfigPath { get; }
    }
    ```

    and call sites should handle the exception as necessary.

lhiginbotham: @RussKie 
I am having a little bit of trouble getting this new exception class to propagate to the UI layer.  The UI layer seems to be making calls to the SubmoduleStatusProvider, which makes an async call to GitCommands.Submodules.SubmoduleStatusProvider.UpdateSubmodulesStatusAsync.  I am unable to catch exceptions from async code on the UI thread, and am not sure how to proceed.

RussKie: @aarnott / @sharwell I don't suppose there is a way to intercept/route an exception from a `FileAndForget` call to the main thread other than the following?

```cs
void MyMethod(Action<Exception> exceptionHandler)
{
    {
        try 
        {
			await TaskScheduler.Default;
			RunCommand();
		}
		catch (Exception)
		{
			await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancelToken);
			exceptionHandler(ex);
		}
    }).FileAndForget();
}
```

AArnott: I'll first respond to the @russkie's suggestion of the simpler message. While I support a simpler message box to present to the user, the simpler message @russkie proposed *lacks* the detail that would help the user to fix the problem. Specifically the message about the invalid escape character would be *very* useful in helping the user fix the issue. What I'll usually do to craft useful error messages is for the top-level catch block that's responsible to turn the exception into a message box to concatenate the exception messages from the outer exception into the deepest inner exception, from outer to inner, as a complete sentence each. This usually produces a very user-friendly error message like "Error initializing git submodules. Error while parsing the .gitmodule file. Failure to parse a \d escape sequence."

AArnott: Now to the question about how to handle exceptions on the main thread. Yes for a 'fire and forget' method (which is usually something I try to avoid anyway, but is sometimes useful) it can catch its own exception and switch to the main thread before popping a message box just as you've done it. At least in your sample, it isn't clear from the method that the `Action<Exception>` delegate has thread affinity or which thread it would like to run on. So switching to the main thread before invoking it seems a bit odd. If you have xml doc comments on the method explaining this behavior and callers rely on that, then that's fine. Otherwise I'd consider making it a `Func<Exception, Task>` delegate so that the handler can conveniently switch to the thread it needs (if at all).

RussKie: Thank you. 
1. Totally agree, if an exception's hierarchy contains useful to the end user information - it should be shown.
2. The last snippet was pretty much written on a napkin. Indeed switching to the UI thread should occur in the exception handler.

RussKie: @lhiginbotham so I suppose we have the following plan

1. Create a specific exception, that is thrown if unable to parse the .gitmodules files
2. In sync execution flows - callsites should handle the exception at UI layer (i.e. at `GiUI` project and not in `GitCommands`)
```cs
try 
{
    // execute functionality that parses .gitmodules file
}
catch (GitConfigurationException ex)
{
    string message = ... // build a good error message
    MessageBox.Show(....);
}
```
3. In async flows, like the above, call sites must pass a delegate that would execute in the UI layer. The delegate will need to switch to the UI thread 
```cs
// callsite
Func<Exception, Task> showErrorAsync = async (exception) =>
{
    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancelToken);

    string message = ... // build a good error message
    MessageBox.Show(....);
};
_submoduleStatusProvider.UpdateSubmodulesStatus(Module.WorkingDir, status, showErrorAsync);

// worker
        public void UpdateSubmodulesStatus(string workingDirectory, [CanBeNull] IReadOnlyList<GitItemStatus> gitStatus, Func<Exception, Task> showErrorAsync)
        {
            //...

            var cancelToken = _submodulesStatusSequence.Next();
            _gitStatusWhileUpdatingStructure = null;
            {
                await TaskScheduler.Default;

                var currentModule = new GitModule(workingDirectory);

                try
                {
                    await UpdateSubmodulesStatusAsync(currentModule, gitStatus, cancelToken);
                }
                catch (Exception ex)
                {
                    await showErrorAsync(ex);
                }

                OnStatusUpdated(_submoduleInfoResult, cancelToken);
            }).FileAndForget();
        }
```

What do you think?

NB: writing in a browser, some things may be off, hopefully the general idea stands.

AArnott: That looks better. But why define a fire-and-forget method that takes a callback? That seems like you're writing old-style (pre-async) code. Why not just return `Task` from your method and let the caller await it with a try/catch around it?

lhiginbotham: @RussKie 
I'll give your latest code suggestion a try - one thing that I have noticed so far is that I need to update all of the unit tests that use this newly modified UpdateSubmodulesStatus interface, which brings two concerns on my part:

1. Should the unit tests provide a meaningful showErrorAsync callback to display any legit errors, or should it just 'do nothing'?

2. How can I switch to the main thread in the showErrorAsync's provided by the unit tests?  I first thought that I needed to create a cancellation token (was going to try `CancellationTokenSource cts = new CancellationTokenSource();` and use `cts` to get a cancellation token) but it turns out that in the following pattern used in the main GitExtensions code, `ThreadHelper` is a class defined in the `GitUI` namespace:

    `await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);`

    I can try to simply add `using GitUI;` to the top of the unit test file (SubmoduleStatusProviderTests.cs), but I've noticed that this test seems to strictly be related to GitCommands - it felt wrong getting GitUI involved.  I guess all of this stems from my poor familiarity with async code/practices and really boils down to whether it is necessary for the showErrorAsync to switch to the main thread in the unit test's case if we are just going to swallow the exception.

RussKie: > Why not just return `Task` from your method and let the caller await it with a try/catch around it?

I suppose we can do that, thank you.
@gerhardol do you see any issues?

RussKie: > 2\. How can I switch to the main thread in the showErrorAsync's provided by the unit tests?

You don't, in unit tests you don't show any UI, so there is no need to be a main thread.
Unit test just need to assert a handler is called with expected data.

If we change from a `FileAndForget` method, there won't be any handler to pass, the method will just throw.
Let's continue the discussion.

mstv: I would not change all the calls to `FileAndForget`: The caller wants to trigger an action whose result/success is not of great interest (as say function name says). I'd rather handle the exceptions inside `FileAndForget`: There could be a generic, common pipe to the main thread in order to display exceptions.

gerhardol: I see no issue in changing this. The current tests are not testing any useful anyway.
But a popup from low level should be OK here, this is misformatted data anyway.

lhiginbotham: @RussKie 

While attempting to take care of all synchronous calls that depend on the .gitsubmodules file parser, I've come to the realization that the config file parser is nearly at the entire root of GitExtensions.  There are so many dependencies on this config file parser that this entire PR might just turn into "Fix Issue #WXYZ: represent _all_ exceptions with a message box from the UI layer".  Do you think I should attempt to put try-catchs all throughout the code-base, or should I swallow some of these errors to cut down on some dependency-trees whenever possible? (For example, in methods that produce `IEnumerable`s based off of the result of parsing the git submodules config file, return an empty list instead of throwing an exception)

RussKie: What do you have in mind?

We do have handlers for unhandled exceptions, those are hooked to NBug form. So we would generally know when things go pear shape. I don't think it would be wise to put plumbing for handling all possible exceptions, because most of them probably won't happen in real life, or users won't be able to fix them themselves.

In this instance we want to handle a specific exception which users observed, and that _can_ happen in real life. And we want to give users guidance on how they can fix the error themselves.

IIUIC, there are few sync flows, where it would be a matter of handling the exception and showing a good message. In one instance, we need to change a `FileAndForget` method into an `async` and handle the exception as well.
Am I missing anything?

mstv: >> There could be a generic, common pipe to the main thread in order to display exceptions.

> We do have handlers for unhandled exceptions, those are hooked to NBug form.

Yes, of course (was quite late yesterday).

Exceptions like `GitConfigurationException` should be shown to the user with a meaningful error message. I think a NBug report is not necessary in this case, especially GE should not be terminated if the error is not fatal.

If there were many code paths which could throw such "user error message" exceptions without a need for special handling, there could be inserted a common handler, which displays these exceptions, before the NBug handler.

> In one instance, we need to change a `FileAndForget` method into an `async` and handle the exception as well.

Yes, that's better than let `FileAndForget` return `Task`.

lhiginbotham:
@RussKie 

Here is the new dialog box:

image

UnitTests/CommonTestUtils/SubmoduleTestHelpers.cs
```diff
@@ -33,16 +34,14 @@ public static void UpdateSubmoduleStatusAndWaitForResult(ISubmoduleStatusProvide
...
-            provider.UpdateSubmodulesStatus(
+            provider.UpdateSubmodulesStatusAsync(
                 workingDirectory: module.WorkingDir,
                 gitStatus: gitStatus);
 
             AsyncTestHelper.WaitForPendingOperations();
 
             provider.StatusUpdated -= Provider_StatusUpdated;
 
-            return;
```
gerhardol: Keep, not required but separate the internal methods

lhiginbotham: @gerhardol 
Are you requesting that I put the `return;` line back in so that it visually separates the implementation from the local method (`void Provider_StatusUpdated(object sender, SubmoduleStatusEventArgs e)`)?

gerhardol: Yes
This should be documented

gerhardol: Yes
This should be documented

gerhardol:
SubmoduleStatusProviderTests currently fails
(Maybe they can be fixed now and test something that actually would matter, see line 122 for instance)

GitExtensions/Program.cs
```diff
@@ -33,11 +33,8 @@ private static void Main()
             {
                 DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
 
-                if (!Debugger.IsAttached)
-                {
-                    AppDomain.CurrentDomain.UnhandledException += (s, e) => ReportBug((Exception)e.ExceptionObject);
-                    Application.ThreadException += (s, e) => ReportBug(e.Exception);
-                }
+                AppDomain.CurrentDomain.UnhandledException += (s, e) => UncaughtExceptionHandler((Exception)e.ExceptionObject);
+                Application.ThreadException += (s, e) => UncaughtExceptionHandler(e.Exception);
```
RussKie: Could you please explain?
When running under the debugger we want to debugger to handle unhandled exceptions, not have them routed to NBug

GitExtensions/Program.cs
```diff
@@ -320,6 +317,22 @@ private static bool LocateMissingGit()
...
         }
 
+        private static void UncaughtExceptionHandler(Exception ex)
+        {
+            if (ex is GitCommands.Config.GitConfigurationException)
+            {
+                var configEx = (GitCommands.Config.GitConfigurationException)ex;
+                string message = string.Format(Strings.GeneralGitConfigExceptionMessage, configEx.ConfigPath, Environment.NewLine, configEx.InnerException.Message);
+                MessageBox.Show(message, Strings.GeneralGitConfigExceptionCaption, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                Environment.Exit(-1);
```
RussKie: I don't think we should crash the app if a submodules config is incorrect.
We could close the repo, but closing the app is a bad UX.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -382,10 +383,18 @@ void InitCountArtificial(out GitStatusMonitor gitStatusMonitor)
 
                         if (AppSettings.ShowSubmoduleStatus)
                         {
-                            _submoduleStatusProvider.UpdateSubmodulesStatus(Module.WorkingDir, status);
+                            try
```
RussKie: Instead of converting to `async void` we should wrap the call in `ThreadHelper.JoinableTaskFactory.Run` instead:
```cs
ThreadHelper.JoinableTaskFactory.Run(() => 
{
    try
    {
        await TaskScheduler.Default;
        await _submoduleStatusProvider.UpdateSubmodulesStatusAsync(Module.WorkingDir, status);
    }
    catch (GitConfigurationException ex)
    {
        MessageBox.Show(....);
    }
});
```

ResourceManager/Strings.cs
```diff
@@ -42,6 +42,9 @@ public sealed class Strings : Translate
         private readonly TranslationString _parentsText = new TranslationString("{0:Parent|Parents}");
         private readonly TranslationString _childrenText = new TranslationString("{0:Child|Children}");
 
+        private readonly TranslationString _generalGitConfigExceptionMessage = new TranslationString("GitExtensions failed to read the Git configuration file \"{0}\" due to the following error:{1}{1}{2}{1}{1}Due to the nature of this problem, the behavior of the application cannot be guaranteed and must be closed.  Please correct this issue and re-open GitExtensions.");
+        private readonly TranslationString _generalGitConfigExceptionCaption = new TranslationString("Error in Git configuration file");
```
RussKie:
```suggestion
        private readonly TranslationString _generalGitConfigExceptionCaption = new TranslationString("Configuration Error");
```

ResourceManager/Strings.cs
```diff
@@ -42,6 +42,9 @@ public sealed class Strings : Translate
         private readonly TranslationString _parentsText = new TranslationString("{0:Parent|Parents}");
         private readonly TranslationString _childrenText = new TranslationString("{0:Child|Children}");
 
+        private readonly TranslationString _generalGitConfigExceptionMessage = new TranslationString("GitExtensions failed to read the Git configuration file \"{0}\" due to the following error:{1}{1}{2}{1}{1}Due to the nature of this problem, the behavior of the application cannot be guaranteed and must be closed.  Please correct this issue and re-open GitExtensions.");
```
RussKie:
```suggestion
        private readonly TranslationString _generalGitConfigExceptionMessage = new TranslationString("Failed to read \"{0}\" due to the following error:{1}{1}{2}{1}{1}Due to the nature of this problem, the behavior of the application cannot be guaranteed and must be closed.  Please correct this issue and re-open GitExtensions.");
```

GitExtensions/Program.cs
lhiginbotham: @RussKie 

With so much of GE's functionality relying on the config parser, I think it is a good idea to close the repo (at this layer, I do not think there is a ).  I was not aware that we could simply route back to the "home page/repos page" instead of quitting, however.  Is there a simple command to get us back there, sort of wiping away the working memory of the current GE session as if it was a clean re-launch of the app?

RussKie: Under [Repository] menu, there is a menu item "Close (go to Dashboard)", I think we should be able to do this.
https://github.com/gitextensions/gitextensions/blob/1219252182ff0ff895cbc99c565fd5189fafd3d8/GitUI/CommandsDialogs/FormBrowse.cs#L1769-L1772

gerhardol: Is going to the dashboard really better than exiting completely?

We could add a new icon for this state
(Similar to missing submodule, just a "pen" same as modified right now.)
Maybe a separate PR

RussKie: In my opinion it is, if a repo is corrupt, why do we need to kill the app? Close the repo.
In fact, for submodules config I don't think it is necessary to close the repo, tell the user everytime the config is read that it is corrupt and let the user to fix it. I don't think we should block the user from browsing commits.

gerhardol:
As is, this PR improves the error message but reduces the functionality.
Previously you could just dismiss the error message and continue, now the app exits.

Before and after, you cannot open a repo with a config that cannot be parsed (OK)
(You could dismiss NBug and open another repo before, not much difference)

If the error is in a submodule repo, config file, the behavior is similar too.
You get an exception in the diffwindow if the selection is for a revision.
For an artificial commit, the current implementation raises NBug on right clicking too.

The original problem is for .gitmodules parsing when getting the structure. for sidepanel/menu.
Currently, you get the popup but can work with the current module (for Diff, you get another NBug).
With this PR, GE just exits which I do not believe is correct.

I looked a little at this, skipping DoFatalExit() in FormBrowse. The problem with this is that the popup/exit() still occurs in Program.cs

RussKie:
I pushed a change with how I believe it should be. I have change only the prod code, I haven't changed any tests. Message box headings will need further work too (LOC).
Please let me know what you think of it.

lhiginbotham:
@RussKie 
Thanks, I will take a look at it the first chance I get.  This PR has been dragging along due to the amount of conversation on the changes and due to how busy I have been lately, which I apologize for.  I’ll give it a shot and apply your fix to the tests as well.

RussKie:
Thanks.
And don't stress about it, in FOSS everyone contributes as time permits

UnitTests/CommonTestUtils/SubmoduleTestHelpers.cs
```diff
@@ -11,10 +13,20 @@ public static SubmoduleInfoResult UpdateSubmoduleStructureAndWaitForResult(ISubm
...
+            {
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    await provider.UpdateSubmodulesStructureAsync(
+                        workingDirectory: module.WorkingDir,
+                        noBranchText: string.Empty,
+                        updateStatus: updateStatus);
+                });
+            }
+            catch (Exception)
```
mstv: Do you really expect exceptions here -- especially those that should be ignored silently?
I'd let NUnit catch and log them.
`provider.StatusUpdated -= Provider_StatusUpdated;` should rather be done in a `finally` block.

UnitTests/CommonTestUtils/SubmoduleTestHelpers.cs
```diff
@@ -33,9 +45,19 @@ public static void UpdateSubmoduleStatusAndWaitForResult(ISubmoduleStatusProvide
...
+            try
+            {
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    await provider.UpdateSubmodulesStatusAsync(
+                        workingDirectory: module.WorkingDir,
+                        gitStatus: gitStatus);
+                });
+            }
+            catch (Exception)
```
mstv: ditto -- wiith the little difference that `Provider_StatusUpdated` is pointless here.

UnitTests/CommonTestUtils/SubmoduleTestHelpers.cs
RussKie: We don't use JTF in tests. We just await.

mstv: > We don't use JTF in tests. We just await.

The function `UpdateSubmodulesStructure` is not `async`, i.e. no need to `await`. Or have I missed something?
Why do you think it should be called using JTF, @lhiginbotham?

Regarding exception-safety refer to https://github.com/gitextensions/gitextensions/commit/5c803acb86fee03cf3b360b9b7440559577e05ab.

lhiginbotham: I was using JTF to avoiding having to propagate async all the way up to the test level, but I have since learned that it is better to do this async propagation whenever converting sync code to async.

ResourceManager/Strings.cs
RussKie: Outstanding

ResourceManager/Strings.cs
RussKie: Outstanding

UnitTests/CommonTestUtils/SubmoduleTestHelpers.cs
```diff
@@ -28,14 +31,14 @@ void Provider_StatusUpdated(object sender, SubmoduleStatusEventArgs e)
             }
         }
 
-        public static void UpdateSubmoduleStatusAndWaitForResult(ISubmoduleStatusProvider provider, GitModule module, IReadOnlyList<GitItemStatus> gitStatus)
+        public static async void UpdateSubmoduleStatusAndWaitForResultAsync(ISubmoduleStatusProvider provider, GitModule module, IReadOnlyList<GitItemStatus> gitStatus)
```
RussKie: `async void` is evil.
```suggestion
        public static async Task UpdateSubmoduleStatusAndWaitForResultAsync(ISubmoduleStatusProvider provider, GitModule module, IReadOnlyList<GitItemStatus> gitStatus)
```

When you have some spare time, have a watch https://www.youtube.com/watch?v=yhmiWhB9Qns

GitCommands/Git/GitModule.cs
```diff
@@ -431,7 +431,15 @@ public bool HasSubmodules()
         public IReadOnlyList<string> GetSubmodulesLocalPaths(bool recursive = true)
         {
             var localPaths = new List<string>();
-            DoGetSubmodulesLocalPaths(this, "", localPaths, recursive);
+            try
+            {
+                DoGetSubmodulesLocalPaths(this, "", localPaths, recursive);
+            }
+            catch (GitConfigurationException)
```
lhiginbotham: @RussKie 

After making what seems to be the finishing touches on this PR (other than squashing), I've decided to interactively test this.  I've found that upon opening a git repo with a malformed config file, the exceptions thrown by the parser are lost here in this catch block.  The comment claims that any config exceptions would have already been shown to the user at this point, but I have not seen any such error messages.

RussKie: 👍 
We've iterated few time over the code, so it is likely this is no longer correct.
What happens you remove both instances of `try..catch` from this class?

lhiginbotham: @RussKie 

When I remove both of the `try..catch` blocks from the class, I get this non-fatal pop-up upon opening GE and upon hitting refresh:

on-start-RELEASE

When I go to the "Diff" tab on a commit, I get this fatal bug report pop-up:

on-diff-tab-RELEASE


The same behavior occurs if I ONLY remove the `try..catch` block in `GitCommands.GitModule.GetSubmodulesLocalPaths`

If I remove only the `try..catch` from `GitCommands.GitModule.GetSubmodulesInfo`, I do not get these pop-up windows.  However, if I go to commit 853f1cc2e0dec946128e7f36fb91e1f2c95987e4 (a commit that has changes in a submodule) and go to the "Diff" tab, I see this little message in the changed-files list:
image

Doing a search of "bad config" in the entire GitExtensions.sln solution, I cannot find this string at all and suspect this changed-files list is a third-party component that GE utilizes.

RussKie: I'll have a look

RussKie:
Looks like this completely dropped off my radar, sorry.
I'm struggling to regain the context here, were you waiting for me?

RussKie:
Closing as stale. Replaced by https://github.com/gitextensions/gitextensions/pull/8049

PR #7343, opened 2019-10-21
Show syntax highlighting also when displaying a diff

### LINKED ISSUE #7323: Detect syntax highlighting from from file contents
## Feature description

The GE fileviewer has a basic syntax highlighting support. Currently it is available when viewing full files (for instance in the file tree), with #7291 also in the diff window.
Detection of the syntax is based on file extension only.
The support has to be implemented in the GE build, not possible to add/change for file extensions or syntax formats.

As discussed in https://github.com/gitextensions/ICSharpCode.TextEditor/pull/15, support could be detected from pre-amble.
For diff view, this will require that the diff contents is retrieved too (currently just the diff is retrieved from Git).

meaney:
with an option in the diff toolbar to enable/disable it

Fixes #7323

## Proposed changes

- Honor highlighting strategy based on file extension when viewing patches

### Before
Default Highlighting Strategy Applied
image

### After

C# Highlighting strategy applied
image

image

image

## Test methodology 

- Unit tests
- Manual testing to verify settings can be toggled and are remembered on restart

## Test environment(s) 
git version 2.20.1.windows.1
Windows
Windows 10 Home Build 18362

I contribute this code under [The Developer Certificate of Origin]


GitUI/CommandsDialogs/SettingsDialog/Pages/DiffViewerSettingsPage.Designer.cs
```diff
@@ -50,19 +51,22 @@ private void InitializeComponent()
...
+            this.tableLayoutPanelForDiffViewer.Controls.Add(this.chkOmitUninterestingDiff, 0, 5);
+            this.tableLayoutPanelForDiffViewer.Controls.Add(this.chkOpenSubmoduleDiffInSeparateWindow, 0, 6);
+            this.tableLayoutPanelForDiffViewer.Controls.Add(this.chkShowDiffForAllParents, 0, 7);
+            this.tableLayoutPanelForDiffViewer.Controls.Add(this.label1, 0, 9);
+            this.tableLayoutPanelForDiffViewer.Controls.Add(this.VerticalRulerPosition, 1, 9);
             this.tableLayoutPanelForDiffViewer.Dock = System.Windows.Forms.DockStyle.Fill;
             this.tableLayoutPanelForDiffViewer.Location = new System.Drawing.Point(0, 0);
             this.tableLayoutPanelForDiffViewer.Name = "tableLayoutPanelForDiffViewer";
-            this.tableLayoutPanelForDiffViewer.RowCount = 9;
+            this.tableLayoutPanelForDiffViewer.RowCount = 11;
```
gerhardol: 11? why not 10?

UnitTests/GitUITests/Editor/FileViewerTextTests.cs
```diff
@@ -116,5 +116,65 @@ public void FileViewer_highlight_selected_text_when_enabled()
...
+            {
+                var uiCommands = new GitUICommands(testHelper.Module);
+                _uiCommandsSource.UICommands.Returns(x => uiCommands);
+                _fileViewer.UICommandsSource = _uiCommandsSource;
+
+                // act
+                _fileViewer.ViewPatch(sampleBinaryPatch, null, "binaryfile.bin");
+
+                // assert
+                _fileViewer.GetTestAccessor().FileViewerInternal.GetTestAccessor().TextEditor.Document.HighlightingStrategy.Should().Be(HighlightingManager.Manager.DefaultHighlighting);
```
gerhardol: What about a test that does not have default highlighting?

GitUI/Editor/FileViewer.cs
```diff
@@ -68,6 +68,10 @@ public partial class FileViewer : GitModuleControl
         [Browsable(false)]
         public byte[] FilePreamble { get; private set; }
 
+        [Description("Show syntax highlighting in diffs.")]
+        [DefaultValue(true)]
+        public bool ShowSyntaxHighlightingInDiff { get; set; }
```
gerhardol: Do we want this status variable?
I first thought AppSettings.ShowSyntaxHighlightingInDiff could be used, but that may not be stored.
showSyntaxHighlighting.Checked is still local

UnitTests/GitUITests/Editor/FileViewerTextTests.cs
meaney: Added

GitUI/Editor/FileViewer.cs
meaney: It is there to be consistent with the pattern used for each of the settings in the FileViewer toolbar.

GitUI/CommandsDialogs/SettingsDialog/Pages/DiffViewerSettingsPage.Designer.cs
meaney: Changed to 10.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1400,6 +1400,28 @@ public static bool RememberNumberOfContextLines
...
+        public static bool ShowSyntaxHighlightingInDiff
+        {
+            get
+            {
+                const bool defaultValue = true;
+                return RememberShowSyntaxHighlightingInDiff ? GetBool("ShowSyntaxHighlightingInDiff", defaultValue) : defaultValue;
+            }
+            set
+            {
+                if (RememberShowSyntaxHighlightingInDiff)
```
RussKie: Why is this conditional?

GitCommands/Settings/AppSettings.cs
```diff
@@ -1400,6 +1400,28 @@ public static bool RememberNumberOfContextLines
             set => SetBool("RememberNumberOfContextLines", value);
         }
 
+        public static bool ShowSyntaxHighlightingInDiff
+        {
+            get
+            {
+                const bool defaultValue = true;
```
RussKie: I don't believe we are achieving any gains here, but inhibiting the readability.
Please remove and convert to an expression

GitUI/Editor/FileViewer.cs
RussKie: I haven't reviewed all other properties, but this is very leaky. The property could be set from code and it won't be updating user settings; which can create problems.

This property doesn't appear to be used outside the class, so there doesn't appear necessary to expose it.
Please convert it to a private field.
For the tests it can be exposed via the `TestAccessor`.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1400,6 +1400,24 @@ public static bool RememberNumberOfContextLines
             set => SetBool("RememberNumberOfContextLines", value);
         }
 
+        public static bool ShowSyntaxHighlightingInDiff
+        {
+            get => RememberShowSyntaxHighlightingInDiff && GetBool("ShowSyntaxHighlightingInDiff", true);
```
RussKie: I disagree with this, these are settings - i.e. what a user has chosen. We should not have conditions here.
How settings combined affect the UI should be done at the UI layer.

gerhardol: The implementation follows the existing options, so I  believe this is fine.
However, I suggest the default for RememberShowSyntaxHighlightingInDiff  is changed to true, which would be about the same.

RussKie:
Ok then

gerhardol:
@msftbot merge in 1 day

PR #7360, opened 2019-10-24
Don't assume e.Node is nullable

jvitkauskas:
## Proposed changes

-  ~Assume that `e.Node` may be null in first `if` like it's done in the second `if`~
- Don't assume e.Node is nullable

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
+++ b/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
@@ -390,7 +390,7 @@ See the changes in the commit form.");
                 return;
             }
 
-            if (e.Node?.Tag is GitItem gitItem)
+            if (e.Node.Tag is GitItem gitItem)
             {
                 e.Node.Nodes.Clear();
                 _revisionFileTreeController.LoadChildren(gitItem, e.Node.Nodes, tvGitTree.ImageList.Images);


jvitkauskas:
Nope, haven't seen any, it's just strange it is assumed to may be null after using as a non-null reference.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -385,7 +385,7 @@ Task ViewGitItemAsync(GitItem gitItem)
 
         private void tvGitTree_BeforeExpand(object sender, TreeViewCancelEventArgs e)
         {
-            if (e.Node.IsExpanded)
+            if (e.Node?.IsExpanded ?? false)
```
RussKie: Whilst technically `e.Node` can be `null`, I don't believe it is ever `null`.
https://github.com/dotnet/winforms/search?q=TreeViewCancelEventArgs

If anything line:393 should be fixed to remove `?`.

jvitkauskas: Ok

RussKie:
Please squash and it is good to go

jvitkauskas:
Done

jvitkauskas:
I am already in `contributors.txt`

mstv:
No, I forgot that I had checked out an older branch.

PR #7365, opened 2019-10-27
Fix 7364: "Fetch and Prune All" confirmation modal has weird focus behavior

### LINKED ISSUE #7364: "Fetch and Prune All" confirmation modal has weird focus behavior
## Current behaviour

When clicking in the "browse form" the "Fetch and Prune All" option in the pull drop-down button, the following dialog box pops up to confirm this action:
image

If this window is up, and I click on the main GitExtensions window when it is behind other windows, the main window will not come to the front and neither will the modal.  If I have the main window and the confirmation dialog partially covered by other windows and I click the confirmation dialog, the dialog will come to the front but the main application will not. 

## Expected behaviour

Clicking the main window should bring both the main window and the dialog to the front.  Clicking the dialog should have the same effect.

## Steps to reproduce
1. Open a repository with GE.
2. Go to the "pull" drop-down button and click "Fetch and prune all"
image
3. Partially cover the GE main window and the confirmation dialog box with a few windows
4. Play around with trying to bring the GE main window to the front, either with Alt + Tab, Windows Key + Tab, or simply by clicking the window
5. Play around with the behavior of clicking the dialog box

## Screenshots

See above.

Not sure, I had this problem for a while.

## Environment

 
image

lhiginbotham:
- use the proper window owner when showing the "fetch and prune all" confirmation dialog (to fix focus issues)

Fixes #7364 

## Proposed changes

- use the correct owner window to fix focus issues 

## Screenshots 

**N/A**

## Test methodology 

- built and interactively tested the fix

## Test environment(s) 

- Windows 10

----

### DIFF
--- a/GitUI/CommandsDialogs/FormPull.cs
+++ b/GitUI/CommandsDialogs/FormPull.cs
@@ -245,7 +245,7 @@ namespace GitUI.CommandsDialogs
 
                 bool isActionConfirmed = AppSettings.DontConfirmFetchAndPruneAll
                                          || MessageBox.Show(
-                                             this,
+                                             owner,
                                              _pullFetchPruneAllConfirmation.Text,
                                              messageBoxTitle,
                                              MessageBoxButtons.YesNo) == DialogResult.Yes;


RussKie:
> Though I am afraid that this fix must be applied to the very many other occurrences of `MessageBox.Show(this, ...)` in GE.

I am not so sure, we need to inspect all of them and go from there.
In this particular case the containing method has `owner` passed in.

lhiginbotham:
@mstv @RussKie 

I think in principle this can go in since it fixes a specific issue that has high visibility.  I hit this issue so frequently and am itching to kill it 😄 We can open a second issue/research task to investigate _all_ popup dialogs for such focus issues.  There is no guarantee that these other dialogs would have similar issues, however, as I think this is more of an edge case.

mstv:
@msftbot merge in 8 hours

PR #7369, opened 2019-10-28
Commit messages menu: Fix (sometimes) broken display

### LINKED ISSUE #5115: Commit dialog inconsistently renders its dropdowns
a bug

The "commit message" dropdown icons have significantly different dimensions than items in the other dropdowns on the form.

issue

* Scale Windows to 200%. 
For example, my scaling configuration:
image

All elements across all dropdowns have consistent dimensions

- GitExtensions version: master @ 9bee0c34d72ec38fe1eb514e09c8bb00a7bc38f0
- OS version: W10 scaled

pmiossec:
Fixes #5115

## Proposed changes

When building the commit message history to display in the dropdown menu,
and the commit message is shorter than the maximum length allowed
but is a multi line commit message,
then the display is broken for all the menus items.

=> add a check that the commit message is multi line and must be shorten.

+ increase the length of text in the menu items because 50 char is too short

### Before

See the first menu item which is display onto 2 lines and so increase the height of all menu items:
 
image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build bffb060100bbe3ab838b00d2a37810cb34e1638a (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4042.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -2330,26 +2330,19 @@ namespace GitUI.CommandsDialogs
 
             void AddCommitMessageToMenu(string commitMessage)
             {
-                const int maxLabelLength = 50;
+                const int maxLabelLength = 72;
 
-                string label;
+                string label = commitMessage;
+                var newlineIndex = label.IndexOf('\n');
 
-                if (commitMessage.Length <= maxLabelLength)
+                if (newlineIndex != -1)
                 {
-                    label = commitMessage;
+                    label = label.Substring(0, newlineIndex);
                 }
-                else
-                {
-                    var newlineIndex = commitMessage.IndexOf('\n');
 
-                    if (newlineIndex != -1 && newlineIndex <= maxLabelLength)
-                    {
-                        label = commitMessage.Substring(0, newlineIndex);
-                    }
-                    else
-                    {
-                        label = commitMessage.ShortenTo(maxLabelLength);
-                    }
+                if (label.Length > maxLabelLength)
+                {
+                    label = label.ShortenTo(maxLabelLength);
                 }
 
                 commitMessageToolStripMenuItem.DropDownItems.Add(new ToolStripMenuItem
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2330,11 +2330,11 @@ private void CommitMessageToolStripMenuItemDropDownOpening(object sender, EventA
 
             void AddCommitMessageToMenu(string commitMessage)
             {
-                const int maxLabelLength = 50;
+                const int maxLabelLength = 72;
 
                 string label;
 
-                if (commitMessage.Length <= maxLabelLength)
+                if (commitMessage.Length <= maxLabelLength && commitMessage.IndexOf('\n') == -1)
```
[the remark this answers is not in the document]
RussKie: No modern systems use CR on their own anymore AFAIK, however it is probably worth guarding against it

RussKie:
I think this will fix https://github.com/gitextensions/gitextensions/issues/5115

RussKie:
I'd like to see some tests as well please

pmiossec:
done.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2330,26 +2330,19 @@ private void CommitMessageToolStripMenuItemDropDownOpening(object sender, EventA
...
-                    {
-                        label = commitMessage.Substring(0, newlineIndex);
-                    }
-                    else
-                    {
-                        label = commitMessage.ShortenTo(maxLabelLength);
-                    }
+                if (label.Length > maxLabelLength)
+                {
+                    label = label.ShortenTo(maxLabelLength);
```
gerhardol: This could make a second allocation. Should that be allowed?

mstv: It's not the RevisionGrid but only on user interaction.

pmiossec: We preferred readability over performance because there is no performance constraints here.

RussKie: I don't think we have `Span` available in 4.6.1, so leave it as is

PR #7375, opened 2019-10-31
Refactor FormClone.ToTextUpdate()

jvitkauskas:
## Proposed changes

- Refactor convoluted FormClone.ToTextUpdate() method
- Fix: text is no longer colored in red when cloning to a path containing "[" or "]" symbols

### Before

image

### After

image

## Test methodology

-  Manual testing

## Test environment(s)

- GIT 2.23
- Windows 10

----

### DIFF
--- a/GitUI/CommandsDialogs/FormClone.cs
+++ b/GitUI/CommandsDialogs/FormClone.cs
@@ -368,53 +368,32 @@ namespace GitUI.CommandsDialogs
 
         private void ToTextUpdate(object sender, EventArgs e)
         {
-            string destinationPath = string.Empty;
+            bool destinationUnfilled = string.IsNullOrEmpty(_NO_TRANSLATE_To.Text);
+            bool subDirectoryUnfilled = string.IsNullOrEmpty(_NO_TRANSLATE_NewDirectory.Text);
 
-            if (string.IsNullOrEmpty(_NO_TRANSLATE_To.Text))
-            {
-                destinationPath += "[" + label2.Text + "]";
-            }
-            else
-            {
-                destinationPath += _NO_TRANSLATE_To.Text.TrimEnd('\\', '/');
-            }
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormClone.cs
```diff
@@ -368,53 +368,32 @@ private void FromTextUpdate(object sender, EventArgs e)
...
-                else
-                {
-                    Info.ForeColor = SystemColors.ControlText;
-                }
-            }
-            else
-            {
-                Info.Text += " " + _infoDirectoryNew.Text;
-                Info.ForeColor = SystemColors.ControlText;
+                Info.Text = $@"{newRepositoryLocationInfo} {_infoDirectoryExists.Text}";
```
mstv: Is "$@" really necessary here? If yes, I guess it should also be used above.

jvitkauskas: I am not sure if it should be translated (there's a warning about localizable string if it's not a verbatim string).

mstv: IIUIC, `$@"..."` avoids the special handling of backslashes `\` in literal(?) strings. I guess this has no effect on replaced variables. Hence `$@"..."` is pointless in this function, i.e. `$"..."` should suffice. @C# experts, am I right?

RussKie: `@` marks the string as "verbatim", e.g. don't parse escaping, `$` is an interpolation.
so together they make a "verbatim interpolated string"

mstv: I've tested it: `$@{s}` and `${s}` yield the same string also if `s` contains a backslash.
So, since we don't use backslashes, i.e. no potential escaping, in this file, "verbatim" has no effect - neither positive nor negative - and _could_ be omitted.

jvitkauskas:
Done

PR #7377, opened 2019-11-03
fix incorect use of TestAccessor struct

RussKie:
Also add `NoAssertContext` class to prevent tests deadlocking or failing on `Debug.Fail` statements in production code.
The implementation is copied from https://github.com/dotnet/winforms/pull/2191

### DIFF
--- a/UnitTests/CommonTestUtils/NoAssertContext.cs
+++ b/UnitTests/CommonTestUtils/NoAssertContext.cs
@@ -1,0 +1,130 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Threading;
+
+namespace System
+{
+    /// <summary>
+    ///  Use (within a using) to eat asserts.
+    /// </summary>
+    public sealed class NoAssertContext : IDisposable
+    {
+        // For any given thread we don't need to lock to decide how to route messages, as any messages for that
+        // given thread will not happen while we're in the constructor or dispose method on that thread. That
+        // means we can safely check to see if we've hooked our thread without locking (outside of using a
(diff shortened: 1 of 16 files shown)


UnitTests/CommonTestUtils/NoAssertContext.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Threading;
+
+namespace System
+{
+    /// <summary>
+    ///  Use (within a using) to eat asserts.
+    /// </summary>
+    public sealed class NoAssertContext : IDisposable
```
mstv: This is not self-explaining to me.

Why should `ParseCommitBody_should_return_null_if_no_EndOfBody` eat which asserts?

UnitTests/CommonTestUtils/NoAssertContext.cs
```diff
@@ -0,0 +1,130 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Threading;
+
+namespace System
+{
+    /// <summary>
+    ///  Use (within a using) to eat asserts.
```
mstv:
```suggestion
    ///  Use (within a using) to eat asserts, e.g. Debug.Fail.
```

UnitTests/CommonTestUtils/NoAssertContext.cs
RussKie: We need to "eat" this assert:
https://github.com/gitextensions/gitextensions/blob/9564bb2ab333735141f767bdc73ee62501c11a05/GitCommands/RevisionReader.cs#L457
which otherwise could break tests or show an interactive dialog expecting a user input

UnitTests/CommonTestUtils/NoAssertContext.cs
RussKie: You're welcome to send a PR to https://github.com/dotnet/winforms/pull/2191 😉

PR #7398, opened 2019-11-10
fix: Crash on Missing Settings Directory

### LINKED ISSUE #7376: [NBug] Could not find a part of the path GitExtensions\GitExtensions.settings
## Current behaviour

git-extensions can't save any configuration. The setup wizard that runs on first start throws this exception once I try to make any configuration changes.

## Expected behaviour
I can finish setting up git-extensions.

## Steps to reproduce
1. Install git-extensions using an admin user account
1. launch it under a regular user account.

## Error Details
```
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\cscott\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.settings'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite, Boolean checkHost)
   at GitCommands.Settings.FileSettingsCache.SaveImpl()
   at GitCommands.SettingsCache.<>c__DisplayClass3_0.<LockedAction>b__0()
   at GitCommands.SettingsCache.LockedAction[T](Func`1 action)
   at GitCommands.Settings.SettingsContainer`2.Save()
   at GitUI.CommandsDialogs.FormSettings.Save()
   at GitUI.CommandsDialogs.SettingsDialog.Pages.ChecklistSettingsPage.SaveAndRescan_Click(Object sender, EventArgs e)
   at GitUI.CommandsDialogs.SettingsDialog.Pages.ChecklistSettingsPage.GitFound_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
```

## Workaround
1. Create the configuration folder listed in the exception message
1. Launch git-extensions

## Environment
- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.20.1.windows.1 (recommended: 2.22.0 or later)
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

### LINKED ISSUE #7395: Crash on Missing Settings Directory
## Current behaviour

When using Git Extensions from another Windows account (not the one the installer was run from) the directory `C:\Users\David\AppData\Roaming\GitExtensions\GitExtensions\` is not created. As a result settings either fail to save silently, or cause a DirectoryNotFoundException.

Manually creating this directory resolves the issue.

## Expected behaviour

If the `C:\Users\David\AppData\Roaming\GitExtensions\GitExtensions\` directory does not already exist then Git Extensions should create it.

## Steps to reproduce

Log in with a new Windows user account, or simply delete/rename the `C:\Users\David\AppData\Roaming\GitExtensions` directory. Git Extensions will start up like the first time but will be unable to save any settings.

## Environment

- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.24.0.windows.2
- .NET Framework 4.8.4042.0
- DPI 96dpi (no scaling)

RussKie:
Fixes #7376
Fixes #7395

## Proposed changes

- A settings file could not be created and the app would crash, if a user's settings directory is missing.
Ensure the directory is created, before writing the file in it.

## Test methodology 

- manual
- unit test

----

### DIFF
--- a/GitCommands/Settings/FileSettingsCache.cs
+++ b/GitCommands/Settings/FileSettingsCache.cs
@@ -141,12 +141,6 @@ namespace GitCommands.Settings
                     return;
                 }
 
-                int currentProcessId;
-                using (var currentProcess = Process.GetCurrentProcess())
-                {
-                    currentProcessId = currentProcess.Id;
-                }
-
                 var tmpFile = Path.GetTempFileName();
                 WriteSettings(tmpFile);
 
@@ -161,6 +155,13 @@ namespace GitCommands.Settings
                     {
                         // Ignore errors for the backup file
                     }
+                }
+
+                // ensure the directory structure exists
+                var parentFolder = Path.GetDirectoryName(SettingsFilePath);
+                if (!Directory.Exists(parentFolder))
+                {
+                    Directory.CreateDirectory(parentFolder);
                 }
 
                 File.Copy(tmpFile, SettingsFilePath, true);
--- a/UnitTests/GitCommandsTests/Settings/FileSettingsCacheTests.cs
+++ b/UnitTests/GitCommandsTests/Settings/FileSettingsCacheTests.cs
@@ -60,6 +60,39 @@ namespace GitCommandsTests.Settings
             ((Action)(() => cache.SaveImpl())).Should().Throw<Exception>();
         }
 
+        [Test]
+        public void SaveImpl_should_create_folder_if_absent()
+        {
+            var tempPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            var settingsFilePath = Path.Combine(tempPath, "GitExtensions.settings");
+
+            try
+            {
+                var cache = new MockFileSettingsCache(settingsFilePath, false).GetTestAccessor();
+                cache.SetLastModificationDate(DateTime.Now);
+
+                Directory.Exists(tempPath).Should().BeFalse();
+                File.Exists(settingsFilePath).Should().BeFalse();
+
+                cache.SaveImpl();
+
+                Directory.Exists(tempPath).Should().BeTrue();
+                File.Exists(settingsFilePath).Should().BeTrue();
+            }
+            finally
+            {
+                // clean up
+                try
+                {
+                    Directory.Delete(tempPath, true);
+                }
+                catch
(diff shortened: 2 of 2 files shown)


RussKie:
@msftbot merge in 24 hours

GitCommands/Settings/FileSettingsCache.cs
```diff
@@ -163,6 +157,13 @@ protected override void SaveImpl()
                     }
                 }
 
+                // ensure the directory structure exists
```
gerhardol: This creation (as well as the Copy below) should be wrapped in a try-catch too, to handle when the settings path cannot be written to

RussKie: What would we do if we can't write to the path apart from a crash?

RussKie:
@msftbot merge if @gerhardol approves

GitCommands/Settings/FileSettingsCache.cs
gerhardol: We could just ignore or show an popup (that should be limited to one per GE startup).
There are other issues related to this.
But that could be changed separately
Will approve later today, unless someone has any comments as this will be merged at my approval

RussKie: > We could just ignore or show an popup (that should be limited to one per GE startup).

NBug will intercept this and presumably show the error to a user and, likely, to us. I think this is an acceptable solution, we have bigger fish to fry atm.

gerhardol: > NBug will intercept this and presumably show the error to a user and, likely, to us. I think this is an acceptable solution, we have bigger fish to fry atm.

Neither of these are big issues. But I consider the problem to write settings is worse than this. But we take that separately.

PR #7405, opened 2019-11-11
FormCommit: Add a menu item to easily convert eol of a file

pmiossec:
## Proposed changes

- Add an option in the commit form to convert the end of lines of a file. Because that's really boring to be obliged to open an editor to convert the file when a repository don't provide a `.gitattributes`file...

### Before

None

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build ed46c2ee58d0b62f968028302b822b5440b2f72a
- Git 2.20.1.windows.1 (recommended: 2.23.0 or later)
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -157,6 +157,9 @@ namespace GitUI.CommandsDialogs
             this.commitCursorColumn = new System.Windows.Forms.ToolStripStatusLabel();
             this.commitEndPadding = new System.Windows.Forms.ToolStripStatusLabel();
             this.stopTrackingThisFileToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.convertEndOfLinesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.toWindowsCRLFToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.toUnixLFToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.UnstagedFileContext.SuspendLayout();
             this.StagedFileContext.SuspendLayout();
             this.UnstagedSubmoduleContext.SuspendLayout();
@@ -201,6 +204,7 @@ namespace GitUI.CommandsDialogs
             this.toolStripSeparator8,
             this.viewFileHistoryToolStripItem,
             this.toolStripSeparator7,
+            this.convertEndOfLinesToolStripMenuItem,
             this.editFileToolStripMenuItem,
             this.deleteFileToolStripMenuItem,
             this.toolStripSeparator5,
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -1497,7 +1501,30 @@ private void InitializeComponent()
...
-            //
+            // 
+            // convertEndOfLinesToolStripMenuItem
+            // 
+            this.convertEndOfLinesToolStripMenuItem.DropDownItems.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.toWindowsCRLFToolStripMenuItem,
+            this.toUnixLFToolStripMenuItem});
+            this.convertEndOfLinesToolStripMenuItem.Name = "convertEndOfLinesToolStripMenuItem";
+            this.convertEndOfLinesToolStripMenuItem.Size = new System.Drawing.Size(232, 22);
+            this.convertEndOfLinesToolStripMenuItem.Text = "Convert End of lines";
```
mstv: nit:
```suggestion
            this.convertEndOfLinesToolStripMenuItem.Text = "Convert end of lines";
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -3330,6 +3330,71 @@ internal TestAccessor(FormCommit formCommit)
                 string commitMessageText, bool usingCommitTemplate, bool ensureCommitMessageSecondLineEmpty)
                     => FormCommit.FormatCommitMessageFromTextBox(commitMessageText, usingCommitTemplate, ensureCommitMessageSecondLineEmpty);
         }
+
+        private void toWindowsCRLFToolStripMenuItem_Click(object sender, EventArgs e)
```
mstv: Nit: Please move all methods before the `TestAccessor`. I'd sort the private methods alphabetically.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -3330,6 +3330,71 @@ internal TestAccessor(FormCommit formCommit)
...
+        {
+            ConvertEndOfLines(Environment.NewLine);
```
mstv: I'd be explicit here. Perhaps GE will return to Linux.
```suggestion
            ConvertEndOfLines("\r\n");
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -3330,6 +3330,71 @@ internal TestAccessor(FormCommit formCommit)
...
+        }
+
+        private void ConvertEndOfLines(string eol)
```
mstv: This non-trivial function should be tested. (Another candidate for `ApprovalTests.net`?)

RussKie:
I fundamentally disagree with the addition of this feature to the commit dialog.
Its purpose is to facilitate commit, not editing files.

> * to be obliged to open an editor to convert the file when a repository don't provide a `.gitattributes`file...

We would be incorrectly shifting responsibility from a repo to the tool. A correct resolution would be to add a `.gitattributes` file to the repo.

pmiossec:
@RussKie 

>A correct resolution would be to add a .gitattributes file to the repo.

I would have loved to live in a world where every opensource project have a `.gitatttributes`.

Until that moment, that PR is for a pragmatic and convenient feature because it often happens when you contribute in new open source projects to do a bug fix.

And I could assure you that it is **very** boring to see all your files with wrong eol.

And even if I do a PR to add a `.gitattributes`, you still have to fix your files and this feature is very useful for that...

>Its purpose is to facilitate commit

I can't be less agree than that, that's the goal of that feature 🤣 
I have to change the EoL to be able to commit, so, this feature help me to do it...

> not editing files.

That's where we slightly does not all agree... I thought we already add an "Edit file" (see screenshot) which is sometimes convenient and that's was not put there by me 😉

RussKie:
You can write a before-commit script to perform EOL conversions for you. 😉
FormCommit is not a place for this logic, it is already overloaded with
auxiliary functions as is.

pmiossec:
>You can write a before-commit script to perform EOL conversions for you. 😉

No, I don't think it will possible (nor convenient) because I don't know the expected EoL of these repositories before contributing to them and seeing the problem. And I won't see me putting in place the burden to add a hook just for a one time PR.

The other reason is that I don't think I am the only one to sometimes have this kind of problem and I imagined that it could have been helpful ( like the upvote of @mstv probably attest it).

I really want this feature to land inside GE.
Is there a possibility to see this feature merged if I add it to the GE editor (instead of the commit form)?

gerhardol:
I think this is OK (will not block but FormCommit is overloaded), but I would also like to see the current encoding why the editor is more natural.
I use the difftool for this, or open in NPP

RussKie:
> I don't know the expected EoL of these repositories before contributing to them and seeing the problem.
> I really want this feature to land inside GE.

I don't seem to understand the full problem.
Could you please explain it in more details.

Say, 
* a repo doesn't have `.gitattributes` file,
* a user may have one of three possible settings in `core.autocrlf`,
* the user commits a change

What is the problem?

RussKie:
Closing until we have a discussion about the feature and explore other ways the desired can be achieved.

PR #7406, opened 2019-11-12
Add view blame in GitHub

pmiossec:
## Proposed changes

- Being able to blame a file (because it's great and quicker) and open the result on GitHub to be able to share the link in an issue, a PR or a review...

### Before

None.

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 50884ddac2894d9d5b6efaddf08e805f03632013 (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4042.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/UserControls/BlameControl.Designer.cs
+++ b/GitUI/UserControls/BlameControl.Designer.cs
@@ -31,7 +31,6 @@
             this.allCommitInfoToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.BlameFile = new GitUI.Editor.FileViewer();
             this.blameTooltip = new System.Windows.Forms.ToolTip(this.components);
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             ((System.ComponentModel.ISupportInitialize)(this.splitContainer1)).BeginInit();
             this.splitContainer1.Panel1.SuspendLayout();
             this.splitContainer1.Panel2.SuspendLayout();
@@ -110,8 +109,7 @@
             this.blamePreviousRevisionToolStripMenuItem,
             this.showChangesToolStripMenuItem,
             this.toolStripSeparator1,
-            this.copyToClipboardToolStripMenuItem,
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem});
+            this.copyToClipboardToolStripMenuItem});
             this.contextMenu.Name = "ContextMenu";
             this.contextMenu.Size = new System.Drawing.Size(239, 76);
             this.contextMenu.Opened += new System.EventHandler(this.contextMenu_Opened);
@@ -187,13 +185,6 @@
             this.BlameFile.Size = new System.Drawing.Size(666, 576);
             this.BlameFile.TabIndex = 0;
             // 
-            // viewInGitHubToolStripMenuItem
-            // 
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem.Name = "_NO_TRANSLATE_ViewInWebSiteToolStripMenuItem";
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem.Size = new System.Drawing.Size(199, 22);
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem.Text = "View in WebSite";
-            this._NO_TRANSLATE_ViewInWebSiteToolStripMenuItem.Click += new System.EventHandler(this.ViewInWebSiteToolStripMenuItem_Click);
-            // 
             // BlameControl
             // 
             this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Inherit;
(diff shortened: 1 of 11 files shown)


Plugins/GitHub3/GitHubHostedRemote.cs
```diff
@@ -52,5 +52,10 @@ public IHostedRepository GetHostedRepository()
         public string Data => Owner + "/" + RemoteRepositoryName;
         public string DisplayData => Data;
         public bool IsOwnedByMe => GitHubLoginInfo.Username == Owner;
+
+        public string GetBlameUrl(string commitHash, string fileName, int lineIndex)
+        {
+            return $"https://github.com/{Data}/blame/{commitHash}/{fileName}#L{lineIndex}";
```
gerhardol: What about gh enterprise?

GitUI/UserControls/BlameControl.cs
```diff
@@ -427,6 +440,38 @@ private void contextMenu_Opened(object sender, EventArgs e)
             blamePreviousRevisionToolStripMenuItem.Enabled = selectedRevision.HasParent;
         }
 
+        private void ConfigureViewInGitHubMenu()
```
gerhardol: This could be extended to other sites, suggest changing GitHub to Website

GitUI/UserControls/BlameControl.cs
```diff
@@ -66,6 +73,11 @@ public BlameControl()
             CommitInfo.CommandClicked += commitInfo_CommandClicked;
         }
 
+        public void SetRepositoryHostPlugin(IRepositoryHostPlugin gitHoster)
+        {
+            _gitHoster = gitHoster;
```
gerhardol: If I interpret this correctly: This will not handle if a repo has remotes in multiple hosts, like GitHub and GitLab
Maybe not needed now, but add a comment that "last host is used".

Plugins/GitHub3/GitHubHostedRemote.cs
pmiossec: Done something (even if there is no github enterprise support in GE at the moment)

GitUI/UserControls/BlameControl.cs
pmiossec: done.

GitUI/UserControls/BlameControl.Designer.cs
```diff
@@ -185,6 +187,13 @@ private void InitializeComponent()
             this.BlameFile.Size = new System.Drawing.Size(666, 576);
             this.BlameFile.TabIndex = 0;
             // 
+            // viewInGitHubToolStripMenuItem
+            // 
+            this.viewInWebSiteToolStripMenuItem.Name = "viewInWebSiteToolStripMenuItem";
+            this.viewInWebSiteToolStripMenuItem.Size = new System.Drawing.Size(199, 22);
+            this.viewInWebSiteToolStripMenuItem.Text = "View in WebSite";
```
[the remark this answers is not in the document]
pmiossec: done.

GitUI/UserControls/BlameControl.cs
pmiossec: What I did is open to support other hosting services but at the moment we only have 1 repository host plugin.

So before having support for use of 2 host plugins at the same time, we will have to have 2 plugins 🤣 and we will have to change other things before: https://github.com/gitextensions/gitextensions/blob/master/GitUI/Plugin/PluginRegistry.cs#L56 (The current strategy is to use the first supported, and not the last)

pmiossec:
>I often wish to get the branch url instead.

I don't see how to retrieve the branch name and here, for the blame, that doesn't make a lot of sense

>The same info in diff would be even nicer...

I thought about it and wanted to do it in the same PR but it's a little bit more complicated so I let it for another PR (if this one would be merged). And also I didn't found a good place to put it. Right click in the revision grid?

GitUI/UserControls/BlameControl.cs
RussKie: We will need to implement some kind of selector. But, like @pmiossec said, we'll cross the bridge once we have more than one host implementation.

RussKie:
I want to see if we can IoC here, something is bothering in the proposed implementation. I'm surprised I've approved it...

gerhardol:
> I want to see if we can IoC here, something is bothering in the proposed implementation. I'm surprised I've approved it...

Everything cannot be done in the optimal way

RussKie:
> Everything cannot be done in the optimal way

Disagree :)

RussKie:
I pushed an IoC proposal. This way you would be able to wire up and show multiple hosters simulataneously, e.g. GitHub and GitLab, with minimal change to the scaffold (i.e. pass a collection inteads of `First`)

pmiossec:
@RussKie I fixed the build and a regression. Now, it seems fine again...

RussKie:
MC

pmiossec:
solved.

PR #7408, opened 2019-11-13
fix: FormResolveConflicts incorrect row selection

RussKie:
## Proposed changes

 If a user selected a different row from the top, upon rescan/refresh of the dialog there would be 2 rows selected - the top and the one that the user wanted.
This would lead to a situation where an incorrect file would be shown, or worse, no file would be shown due to multiple selection.

Fix the selection by resetting the initial selection after databind and selecting the user's choice.
Also if the user selected the last row, and a conflict was resolved, retain the selection and select the new last row upon refresh.

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormResolveConflicts.cs
+++ b/GitUI/CommandsDialogs/FormResolveConflicts.cs
@@ -148,16 +148,40 @@ namespace GitUI.CommandsDialogs
             using (WaitCursorScope.Enter())
             {
                 int oldSelectedRow = 0;
+                bool isLastRow = false;
                 if (ConflictedFiles.SelectedRows.Count > 0)
                 {
                     oldSelectedRow = ConflictedFiles.SelectedRows[0].Index;
+                    isLastRow = ConflictedFiles.Rows.Count - 1 == oldSelectedRow;
                 }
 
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -148,16 +148,40 @@ private void Initialize()
...
                 {
                     oldSelectedRow = ConflictedFiles.SelectedRows[0].Index;
+                    isLastRow = ConflictedFiles.Rows.Count - 1 == oldSelectedRow;
                 }
 
                 ConflictedFiles.DataSource = ThreadHelper.JoinableTaskFactory.Run(() => Module.GetConflictsAsync());
                 ConflictedFiles.Columns[0].DataPropertyName = nameof(ConflictData.Filename);
+
+                // if the last row was previously selected, select the last row again
+                if (isLastRow && ConflictedFiles.Rows.Count >= oldSelectedRow)
```
mstv:
```suggestion
                if (isLastRow && oldSelectedRow >= ConflictedFiles.Rows.Count)
```

RussKie: 👍

PR #7419, opened 2019-11-14
Updated putty to version 0.73.0.0

AikoAppeldorn:
Hi,
I have updated my ssh key to ed25519 and everytime I want to add the key via GitExtension following error occurred:
1055-534-max
I found out that it was an issue with putty. In the latest version of putty the problem was solved.
So I updated all putty binaries to the latest 32bit version.

## Proposed changes

- Updated putty to version 0.73.0.0  


RussKie:
Thank you, I updated the binaries.

PR #7394, opened 2019-11-09
Increase stability of test execution; dispose members

mstv:
Fixes #7382
Fixes #7473

## Proposed changes

- Prepare for repeating the test runs
- Always call `AvatarControl._cancellationTokenSequence.Dispose`
- Dispose `FormStash._asyncLoader` only if disposing
- Check `CancellationToken` in `ROT.FillBranchTree` more often before executing git commands
- Regard `RevisionDataGridView._shouldRun` if `WaitOne` returns `false`
- Join `RevisionDataGridView._backgroundThread` on form close and when restarting the thread
- Remove not existing namespace `CommitDialog` from tests
- Make `UpdateSubmoduleStructureAndWaitForResult` exception-safe
  and cleanup `UpdateSubmoduleStatusAndWaitForResult`
- Make test `Submodule_status_changes_for_top_module` work
- Deny to run tests with joinable tasks in parallel because of singleton
- `ThreadHelper.JoinableTaskContext` must not be `null` and
  ignore follow-up exceptions from `ConfigureJoinableTaskFactoryAttribute.AfterTest`
- Replace `Sleep` with `WaitForPendingOperations` (where applicable)
- Delay tests until form has opened and loaded centrally in `UITests.RunForm`
  but wait for `RevisionDataGridView._backgroundThread` in regarding tests
- Call `Application.DoEvents` while waiting for tasks
- Dispose objects created by `TranslationTest`
- Avoid pointless instance of `FormOpenDirectory`
- More complete teardown of dialog tests
- Detect undisposed forms
- Dispose `FindAndReplaceForm` and `FormRevisionFilter` (which are nested in controls)

## Test methodology 

- run the tests

## Test environment(s) 

- AppVeyor

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7448, opened 2019-11-25
Apply timeouts to JoinTillEmptyAsync

sharwell:
Applies a 60-second timeout to `JoinTillEmptyAsync`. When a debugger is attached, the timeout is extended to an hour.

### DIFF
--- a/UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
@@ -1,7 +1,6 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
+using System.Diagnostics;
+using System.Threading;
 using System.Threading.Tasks;
 using GitUI;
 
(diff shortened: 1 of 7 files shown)


UnitTests/CommonTestUtils/AsyncTestHelper.cs
```diff
@@ -1,39 +1,48 @@
...
         }
 
-        public static void WaitForPendingOperations()
+        public static void WaitForPendingOperations(TimeSpan timeout)
         {
-            try
-            {
-                ThreadHelper.JoinableTaskContext?.Factory.Run(() => ThreadHelper.JoinPendingOperationsAsync());
-            }
-            catch (OperationCanceledException)
```
sharwell: 💭 I'm not sure why this `catch` handler was originally added, but we can't catch this exception or it will hide potential bugs.

sharwell:
@mstv does this fix the test hangs you saw?

mstv:
A first local test run with this and your other PR merged into #7394 has passed, @sharwell.
All the wait workarounds could be removed. 👍 
BTW: I must close MSVS, `git clean -xdf` and (re)build in order to get the method from the updated lib.

RussKie:
@mstv do we merge this as is, and you rebase your work on top?

mstv:
It is a good step forward. Instead of hangs we get correct NUnit timeouts in different tests in some of the test runs now.
I hope `git rebase` will do a good job. I have solved several conflicts when merging this. I don't want to do this again.
The other PR which replaces the background thread is required to be merged, too, before I can rebase. Though it must be reviewed yet.

PR #7449, opened 2019-11-25
Define .editorconfig for C#

sharwell:
This change is intended only to explicitly specify the behavior of C# coding features inside Visual Studio (i.e. the values should match what we were already using).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -35,3 +35,192 @@ trim_trailing_whitespace = false
 indent_style = space
 indent_size = 2
 trim_trailing_whitespace = false
+
+# C# files
+[*.cs]
+
+#### Core EditorConfig Options ####
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -35,3 +35,192 @@ trim_trailing_whitespace = false
...
+#### Core EditorConfig Options ####
+
+# New line preferences
+insert_final_newline = true
+
+#### .NET Coding Conventions ####
+
+# Organize usings
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = true
```
sharwell: 📝 This line is the one that triggered this pull request. My IDE settings have this set to `false`, so I was unable to use the Remove and Sort Usings feature without incorrectly reordering them.

PR #7478, opened 2019-12-02
Adapt GE version in translation

mstv:
## Proposed changes

Adapt GE version in English translation files to 4.0.0 as it is expected after the release of 3.3.0

----


gerhardol:
This should be part of the update version script (that I do not know if it is being updated anymore...)

mstv:
> This should be part of the update version script (that I do not know if it is being updated anymore...)

It is required on the _master_ branch. Should the release script really commit to the master branch?
Though an automation is always welcome.

RussKie:
These values are updated by the translator app.

PR #7489, opened 2019-12-06
Install with AppGet (https://appget.net/)

pmiossec:
## Proposed changes

- Update README.md to tell how to install GitExtensions with https://appget.net/


RussKie:
Nice! Do they have download count shields?

PR #7490, opened 2019-12-08
OpenRemoteUrlInBrowser: Issue (#7439)

### LINKED ISSUE #7439: Open remote in browser
## Feature description

Most git web GUIs allow opening the remote URL (including .git) in the browser, you will be redirected to the actual repo page. I think there should be a context menu entry on a remote to open the URL in the default system browser.

image

## Environment

- Git Extensions 3.1.1.6049
- Git 2.23.0.windows.1
- .NET Framework 4.8.4042.0
- DPI 96dpi (no scaling)

wynnfrei:
Fixes #7439

## Proposed changes
-  Adding a button, that directs you to actual repository's page

### Before

image

### After

image

## Test methodology 

 Manual testing

## Test environment(s) 

- Git Extensions 3.2.1.6628
----


GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -287,6 +287,11 @@ public bool Prune()
                 return DoPrune();
             }
 
+            public void OpenRemoteUrlInBrowser()
+            {
+                Process.Start(_remote.FetchUrl);
```
RussKie: I think this only makes sense for http, but won't make sense for other types of urls, e.g. ssh or git.

wynnfrei: @RussKie whats your suggestion? I'd like to hear and will implement/change it

gerhardol: First, in the current version, the menu item should only be available for Http remotes.
However, there is a framework for handling ssh remotes too, currently only GitHub I believe. Se ExternalLink, implemented by @pmiossec that may have some suggestions.

wynnfrei: @pmiossec any suggestions? 😄

pmiossec: Maybe you could use `GitProtocolExtensions.IsUrlUsingHttp()` and indeed disable it if that's not `true`.

wynnfrei: @pmiossec by disabling you mean to hide/deactivate the button?  if not, please specify. Thanks!

pmiossec: Yes. Hide or deactivate. It not always easy to choose between the 2 and in your case, I'm not sure to know which one make more sense.

I tend to prefer disabling the menuitem (so `Enabled = false`) because that give a kind of feedback to the user.

But I would like the others to give there point of view... (@gerhardol was just asking for my advice on the way to determine the kind of url, not especially on my UI preferences 😉 )

gerhardol: @pmiossec I was more referring to your other external link changes actually
In this case I prefer hide, most users will just see a disabled item assuming ssh is more common still (http credential store error messages are not useful).

RussKie: Yes, only show the menu for HTTP remotes. 
Please use `GitProtocolExtensions.IsUrlUsingHttp()` to detect those.

pmiossec:
I have fixed the PR.

@wynnfrei Could you please signed the CLA? i.e. add your name to the `contributors.txt`.

wynnfrei:
@pmiossec 
Hello. sorry for very late response. Was busy at work. What do you mean exactly. could you please clarify? thank you!

pmiossec:
If you want your pull request to be merged, you must update this file ( https://github.com/gitextensions/gitextensions/blob/master/contributors.txt ) with the following information: github id, Full name, email

wynnfrei:
@pmiossec sorry for late response again.. Will update it on weekends

wynnfrei:
@pmiossec Hey! just updated contributors.txt

gerhardol:
Conflict, please rebase on master
Also squash

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -287,6 +288,13 @@ public bool Prune()
...
```
RussKie: We still need to check here whether we _can_ open the URL.
The fact that the menu is not visible is orthogonal, and there is no obvious connection between these two. It is not difficult to imagine a situation where this method gets invoked programmatically without any safe guards in place.

```cs
            public void OpenRemoteUrlInBrowser()
            {
				if (!IsRemoteUrlUsingHttp)
				{
					return;
				}

                Process.Start(_remote.FetchUrl);
```

wynnfrei: Alrighty. will update it on weekends.

pmiossec:
I have addressed the last remaining comment, squashed and rebased.
So it should be good to go.

GitUI/Translation/English.xlf
```diff
@@ -7912,6 +7912,10 @@ To show all branches, right click the revision grid, select 'view' and then the
         <source>&amp;Manage...</source>
         <target />
       </trans-unit>
+      <trans-unit id="mnuBtnOpenRemoteUrlInBrowser.Text">
+        <source>Open remote url</source>
```
RussKie, line 7916: I'm on the phone. Can you please confirm we use "url" in other places too? Or "URL", etc.

pmiossec: Looking at the all file, that's mostly 'Url' even if there are some 'URL' (2 or 3) .

Indeed, that's the only 'url'. 

I will update that but should we standardize around 'Url'?

RussKie:
Hold off the merge please, I'm going through the list to see what we have outstanding. I have already announced on the Transifex and the translators've been busy for the past few days.

PR #7496, opened 2019-12-10
Filter staged files

mstv:
## Proposed changes

- enable the file filter for staged files and all other `FileStatusList`s
- add button which deletes the file filter
- workaround for match getting selected
  (as it can be seen in the slightly outdated "after" screenshot when type "y")

### Before

grafik

### Before workaround

c912f512-acdb-4109-b047-e9e92148af52

### After

grafik

## Test methodology 

- extended NUnit test

## Test environment(s) 

- Git Extensions 3.3.0
- Build bb42bec9ad42054055ac692d4b884849d4b22c17
- Git 2.23.0.windows.1
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

----


gerhardol:
I prefer the delete button to the right

mstv:
I don't like it at the right side because then it is too far away from the entered text:
grafik
We should add a setting -- just kidding 😁.

gerhardol:
All other clearable searches has the clear to the right. I want it left to the dropdown though, floating in the field

> We should add a setting

Let us put it in Git settings and add another setting that does something complete different but is named similar too.

mstv:
> All other clearable searches has the clear to the right.
> I want it left to the dropdown though, floating in the field

It will be difficult to determine the width of the drop down, especially with scaling.

---

What is such special with AppVeyor?!? There, the TranslationApp seems to not like the `FileStatusList.FilterVisible = true` set in the Designer code.
Local runs do succeed. But the TranslastionApp still generates `GitExVersion="3.3.0"`.

grafik

GitUI/UserControls/FileStatusList.cs
```diff
@@ -802,6 +804,13 @@ private void SelectItems(Func<ListViewItem, bool> predicate)
             StoreNextIndexToSelect();
         }
 
+        private void SetDeleteFilterButtonVisibilityAndFilterComboBoxSize()
```
RussKie: How about?
```suggestion
        private void ToggleFilterControl()
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -236,6 +236,7 @@ public FormCommit([NotNull] GitUICommands commands, CommitKind commitKind = Comm
             Unstaged.SetNoFilesText(_noUnstagedChanges.Text);
             Unstaged.FilterVisible = true;
             Staged.SetNoFilesText(_noStagedChanges.Text);
+            Staged.FilterVisible = true;
```
RussKie: Can we make this the default so we don't need to set it explicitly?

RussKie:
> > We should add a setting
> 
> Let us put it in Git settings and add another setting that does something complete different but is named similar too.

NOOOO! 😱

GitUI/CommandsDialogs/FormCommit.cs
mstv: This fails on AppVeyor and perhaps on other strange machines. See the screenshot above.

RussKie:
I'll check in a few days

GitUI/UserControls/FileStatusList.cs
mstv: "Toggle" means "change the state", which is not correct here.
Showing the `DeleteFilterButton` floating as Gerhard requested (for consistency) was much easier than thought. So the function name has simplified.

GitUI/CommandsDialogs/FormCommit.cs
mstv: Now (that only `Visible` is changed), it works.

GitUI/UserControls/FileStatusList.cs
RussKie: > Showing the `DeleteFilterButton` floating as Gerhard requested

Do screenshots need to be updated?

GitUI/CommandsDialogs/FormCommit.cs
RussKie: Awesome!

GitUI/UserControls/FileStatusList.Designer.cs
```diff
@@ -57,18 +58,18 @@ private void InitializeComponent()
             this.FileStatusListView.ShowItemToolTips = true;
             this.FileStatusListView.Size = new System.Drawing.Size(682, 464);
             this.FileStatusListView.Sorting = System.Windows.Forms.SortOrder.Ascending;
-            this.FileStatusListView.TabIndex = 2;
+            this.FileStatusListView.TabIndex = 4;
```
RussKie: I see the tab order has changed.
I just want a confirmation the tab order remains logical

GitUI/UserControls/FileStatusList.cs
```diff
@@ -202,6 +202,7 @@ public override ContextMenuStrip ContextMenuStrip
 
         public bool FilterFocused => FilterComboBox.Focused;
 
+        public bool ShouldSerializeFilterVisible => FilterVisible != true;
         public bool FilterVisible
```
RussKie: [nit] missing blank line

UnitTests/GitUITests/CommandsDialogs/FileStatusListTests.cs
```diff
@@ -265,9 +268,14 @@ public void Test_StoreFilter_valid(string regex, bool active)
...
             string expectedRegex = string.IsNullOrEmpty(regex) ? null : regex;
 
-            accessor.StoreFilter(regex);
+            _fileStatusList.SetFilter(regex);
 
             CheckStoreFilter(expectedColor, expectedRegex, accessor);
+
+            // FilterVisibleInternal has been reset because no items were added, toggle it
+            _fileStatusList.FilterVisible = false;
+            _fileStatusList.FilterVisible = true;
```
RussKie: Shouldn't these two lines be before the "act" phase (line:271)?
After we set a filter - the button should be visible

mstv: Then the file list must be filled which is little too much for these test cases.
I tried to explain it with the comment `// FilterVisibleInternal has been reset because no items were added, toggle it`.

GitUI/UserControls/FileStatusList.cs
mstv: These two belong tightly together. It would be better if there was an attribute like `[DefaultValue(false)]`, which does not have the intended effect "no value set by the designer" here.

GitUI/UserControls/FileStatusList.Designer.cs
mstv: This is an intended change. `(Shift+)Tab` works correctly.

GitUI/UserControls/FileStatusList.cs
mstv: I updated the screenshots yesterday.

GitUI/UserControls/FileStatusList.Designer.cs
RussKie: Thank you for confirming

GitUI/UserControls/FileStatusList.cs
RussKie: I know what `ShouldSerializeXxx` does, but it is customary to have a blank line between methods.

mstv: I have added the line and rebased.
@msftbot merge in 1 day

mstv:
@msftbot merge in 1 day

GitUI/UserControls/FileStatusList.cs
RussKie: bot only takes orders if the orders aren't mixed with other text

PR #7452, opened 2019-11-26
BuildServerIntegration: Adjust polling interval for running builds

pmiossec:
## Proposed changes

depending if some builds are running (refresh interval is 10s)
or not (interval is 120s)

... to reduce calls to the CI server when not needed.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 1d62d64ab33e9ce45b898542facff3f9a50caf6e (Dirty)
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.8.4042.0
- DPI 192dpi (200% scaling)

PR #7497, opened 2019-12-10
CI: Only look for newly finished builds if previously running build found

pmiossec:
## Proposed changes

Makes GitExtensions more CI server friendly by doing less api calls:
=> prevent 1 (most of the time) useless api call / minutes
These calls are most of the times useless because finished builds that will be found will correspond to commits not in the revision grid.
The only case where it is useful is when there was running builds the last time the revision grid was refreshed and so, to display the state of these builds once finished. 

That's why this PR makes that the calls to find newly finished builds to be made only when running builds were previously found.

Contribute to fix too intensive call of CI api. Follow up of #7477 (For issue https://github.com/gitextensions/gitextensions/pull/7452#issuecomment-559084273 )

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 63aaf21d9ea76f72cf80640421d869f6fb444463 (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4042.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/BuildServerIntegration/BuildServerWatcher.cs
+++ b/GitUI/BuildServerIntegration/BuildServerWatcher.cs
@@ -79,11 +79,20 @@ namespace GitUI.BuildServerIntegration
 
             var fullDayObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler, DateTime.Today - TimeSpan.FromDays(3));
             var fullObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler);
-            var fromNowObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler, DateTime.Now);
 
-            bool anyBuilds = false;
+            bool anyRunningBuilds = false;
             var delayObservable = Observable.Defer(() => Observable.Empty<BuildInfo>()
-                                                                   .DelaySubscription(anyBuilds ? ShortPollInterval : LongPollInterval));
+                                                                   .DelaySubscription(anyRunningBuilds ? ShortPollInterval : LongPollInterval));
+
+            var shouldLookForNewlyFinishedBuilds = false;
+            DateTime nowFrozen = DateTime.Now;
+
+            // All finished builds have already been retrieved,
+            // so looking for new finished builds make sense only if running builds have been found previously
+            var fromNowObservable = Observable.If(() => shouldLookForNewlyFinishedBuilds,
+                buildServerAdapter.GetFinishedBuildsSince(scheduler, nowFrozen)
+                            .Finally(() => shouldLookForNewlyFinishedBuilds = false));
+
             var cancellationToken = new CompositeDisposable
                     {
                         fullDayObservable.OnErrorResumeNext(fullObservable)
@@ -95,10 +104,14 @@ namespace GitUI.BuildServerIntegration
                                          .ObserveOn(MainThreadScheduler.Instance)
                                          .Subscribe(OnBuildInfoUpdate),
 
-                        runningBuildsObservable.Do(buildInfo => anyBuilds = true)
+                        runningBuildsObservable.Do(buildInfo =>
+                                                    {
+                                                        anyRunningBuilds = true;
+                                                        shouldLookForNewlyFinishedBuilds = true;
+                                                    })
                                                .Retry()
                                                .Concat(delayObservable)
-                                               .Finally(() => anyBuilds = false)
+                                               .Finally(() => anyRunningBuilds = false)
                                                .Repeat()
                                                .ObserveOn(MainThreadScheduler.Instance)
                                                .Subscribe(OnBuildInfoUpdate)


pmiossec:
Thanks again @NikolayXHD for the head up with `Observable.Defer()`

GitUI/BuildServerIntegration/BuildServerWatcher.cs
```diff
@@ -79,11 +79,23 @@ public async Task LaunchBuildServerInfoFetchOperationAsync()
...
+
+            var shouldLookForFinishedBuilds = false;
+            DateTime nowFrozen = DateTime.Now;
+            var fromNowObservable = Observable.Defer(() =>
+            {
+                // All finished builds have already been retrieved,
+                // so looking for new finished builds make sense only if running builds have been found previously
+                return !shouldLookForFinishedBuilds
+                    ? Observable.Empty<BuildInfo>()
+                    : buildServerAdapter.GetFinishedBuildsSince(scheduler, nowFrozen)
```
NikolayXHD: consider using `Observable.If()`, based on [defer source code](https://github.com/dotnet/reactive/blob/f650dded4d6d6ced6178f864f2eaecdbb57d5b43/Rx.NET/Source/src/System.Reactive/Linq/Observable/Defer.cs#L21) and [if source code](https://github.com/dotnet/reactive/blob/3f2c1f02018c07c3f34f9d54c844b358ad671721/Rx.NET/Source/src/System.Reactive/Linq/Observable/If.cs#L21) `if(condition, observable)` should be equivalent to  `defer(condition ? observable : empty)`

GitUI/BuildServerIntegration/BuildServerWatcher.cs
```diff
@@ -79,11 +79,23 @@ public async Task LaunchBuildServerInfoFetchOperationAsync()
 
             var fullDayObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler, DateTime.Today - TimeSpan.FromDays(3));
             var fullObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler);
-            var fromNowObservable = buildServerAdapter.GetFinishedBuildsSince(scheduler, DateTime.Now);
 
             bool anyBuilds = false;
```
NikolayXHD: rename `anyBuilds` -> `anyRunningBuilds` for better precision

GitUI/BuildServerIntegration/BuildServerWatcher.cs
```diff
@@ -79,11 +79,23 @@ public async Task LaunchBuildServerInfoFetchOperationAsync()
...
             var delayObservable = Observable.Defer(() => Observable.Empty<BuildInfo>()
                                                                    .DelaySubscription(anyBuilds ? ShortPollInterval : LongPollInterval));
+
+            var shouldLookForFinishedBuilds = false;
+            DateTime nowFrozen = DateTime.Now;
+            var fromNowObservable = Observable.Defer(() =>
+            {
+                // All finished builds have already been retrieved,
+                // so looking for new finished builds make sense only if running builds have been found previously
```
NikolayXHD: Then we should rather append `fromNowObservable` to `runningBuildsObservable`, right?

If so, we could use single `anyBuilds` control variable, because `shouldLookForFinishedBuilds` would always have same value as `anyBuilds`

pmiossec: >Then we should rather append fromNowObservable to runningBuildsObservable, right?

That's what I have imagined first but I decided against it because in this case it will run a lot more often (i.e. every 10s after running build check instead every minutes) and I will be obliged to add a lot more complicated logic (if possible) to keep this interval of 1 call/minutes which is enough for me...

GitUI/BuildServerIntegration/BuildServerWatcher.cs
pmiossec: Thanks, I will have a look. I don't know if that's compatible with the `.Finally()` but I will try.

GitUI/BuildServerIntegration/BuildServerWatcher.cs
NikolayXHD: True, did not realize it at first glance.

gerhardol:
Sidenote: AppVeyor and Jenkins has no filter for 'since' so GetFinishedBuildsSince() always returns empty. But this makes no difference.

pmiossec:
@NikolayXHD Is there still something to do with this PR? I think it's ok...

NikolayXHD:
> @NikolayXHD Is there still something to do with this PR? I think it's ok...

Hi @pmiossec, only this:

> consider using `Observable.If()`

pmiossec:
@NikolayXHD Indeed! Done.

RussKie:
Merge at will

PR #7537, opened 2019-12-19
CI: Improve Azure Devops builds

pmiossec:
## Proposed changes

- Solve case where a running build stay displayed as running forever even if it is finished since a long time (condition on how to filter on finished build was not good)
by filtering on the FinishTime instead of the StartTime
& solve time management as AzureDevOps use UTC time format

- AzureDevOps: Ignore first api call for finished builds
because as it is implemented now, the 2 first calls (to retrieve today finished builds and all finished builds) are doing the same api call. So we could skip the first call to only keep the one retrieving all the finished builds and so prevent a useless api call (and be more server friendly)

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 63aaf21d9ea76f72cf80640421d869f6fb444463 (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4067.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
+++ b/Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
@@ -118,10 +118,15 @@ namespace AzureDevOpsIntegration
 
             var builds = (await HttpGetAsync<ListWrapper<Build>>($"build/builds?api-version=2.0&definitions={buildDefinitionsToQuery}")).Value;
 
+            if (!running.HasValue || running.Value)
+            {
+                return builds
+                    .Where(b => !running.HasValue || running.Value == b.IsInProgress)
+                    .Where(b => !sinceDate.HasValue || b.StartTime >= sinceDate.Value.ToUniversalTime());
+            }
+
             return builds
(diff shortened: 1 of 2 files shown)


Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
```diff
@@ -118,9 +118,16 @@ public async Task<IEnumerable<Build>> QueryBuildsAsync(string buildDefinitionsTo
 
             var builds = (await HttpGetAsync<ListWrapper<Build>>($"build/builds?api-version=2.0&definitions={buildDefinitionsToQuery}")).Value;
 
+            if (!running.HasValue || running.Value)
+            {
+                return builds
+                    .Where(b => !running.HasValue || running.Value == b.IsInProgress)
+                    .Where(b => !sinceDate.HasValue || b.StartTime >= sinceDate.Value.ToUniversalTime())
+                    .ToList();
```
RussKie: Do we need to materialise the collection here? The caller will be re-materialising them again, won't they?

pmiossec: FIxed.

PR #7547, opened 2019-12-22
gitex: If no argument start browse in current workdir

### LINKED ISSUE #7542: gitex.cmd no longer opens the current repo
## Current behaviour

When I run gitex.cmd from the command line in a Git repo directory, it opens the GitExtensions dashboard.

## Expected behaviour

Directly open the current repo in GitExtensions. This worked before (I was using 3.0.x), and stopped working when I upgraded to 3.3.0.7719.

## Steps to reproduce

Execute gitex.cmd from the command line in a Git repo directory.

## Screenshots

N/A

Yes, 3.0.x

## Environment

- Git Extensions 3.3.0.7719
- Git 2.24.1.windows.2
- .NET Framework 4.8.4075.0
- DPI 144dpi (150% scaling)

## Diagnostics

N/A

---

I don't know if it's a bug or if it's intentional, but I don't like the new behavior, because it requires me to manually find my repo in the dashboard (I have about 30 of them in recent repos). Is there a way to restore the previous behavior?

gerhardol:
Fixes #7542

## Proposed changes
If gitex.cmd has no arguments, open browse with the current directory.
gitextensions.exe obeys the arguments since #7147 and defaults by opening the dashboard 

## Test methodology 
Manual

----

### DIFF
--- a/Bin/gitex.cmd
+++ b/Bin/gitex.cmd
@@ -1,4 +1,7 @@
 @setlocal
 @for /F "delims=" %%I in ("%~dp0") do @set gitex_folder=%%~fI
 @set PATH=%gitex_folder%;%PATH%
-@start /B GitExtensions.exe %*
+@REM If no arguments, try open current working directory
+@set arg=%*
+@if "%arg%"=="" SET arg=browse .
+@start /B GitExtensions.exe %arg%


gerhardol:
There are likely more elegant ways to solve this, but this works for me.
Will let this soak for some days. If we make a 3.3.1 we can include it too.

RussKie:
Yes, looks like there will be a 3.3.1, so please port it there

PR #7552, opened 2019-12-22
WIP simplify mergetool handling

### LINKED ISSUE #7550: Remove 2-way merge resolve
## Feature description

Removing 2-way merge is the functionality change, needed to clean up some messy code and to support #5919.

GE sets the path and cmd for mergetool and difftool in settings, similar to what will be done with git-config without any manipulation.

GE always call git-difftool without manipulation of arguments.

GE has two ways to invoke mergetool form FormResolveConflicts:
image

 a. "Open in <tool>", context menues etc will get git config arguments and manipulate them
 b. "Start mergetool" will call git-mergetool

a. Used to have a lot of special handling, especially involving kdiff3 that was distributed with GE in 2.x. Most of the special handling was removed in #7009 (3.3), a few special cases that could not be understood was removed in #7044. The only that remains that I see is special handling for 2-way merge manipulation, hardcoded for kdiff3 and diffmerge (as well as for beyondcompare3, araxis, tortoisemerge, toolnames no longer used in GE 4). There are a few GitHub issues when applying the 2-way fixes, at least beyond compare seem to work better without this fix #1844.

The 2-way merge could be fixed by hardcoding for the tools where this is relevant (and possibly make this configurable but I would like to avoid that). However, I propose the explicit 2-way merge is removed together with all explicit handling of .cmd, .path.

As arguments can be changed in GE GUI (they could always be changed from the commandline), it is riskier to allow this command now, manipulations may fail.

## Environment

- GitExtensions version: 4.0
- GIT version: all
- OS version: n/a
- .NET version: n/a

gerhardol:
Fixes #7550 

## Proposed changes

An example for the changes proposed in #7550 
No plan to get this merged in its current state (not even updated translations)

## Screenshots

The explanation what uses the GE filters and the plain mergetool must be clarified
image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -888,9 +888,14 @@ namespace GitCommands
             }
         }
 
-        public void RunMergeTool()
+        public void RunMergeTool([CanBeNull] string file = "")
         {
-            using (var process = _gitExecutable.Start("mergetool", createWindow: true))
+            var args = new GitArgumentBuilder("mergetool")
+            {
+                { file.IsNotNullOrWhitespace(), "--" },
+                file.ToPosixPath().QuoteNE()
+            };
+            using (var process = _gitExecutable.Start(args, createWindow: true))
             {
                 process.WaitForExit();
             }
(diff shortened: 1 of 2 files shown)


gerhardol:
No comments on removal, so I refer to the fix instead

RussKie:
I like when code is getting deleted...

gerhardol:
> I like when code is getting deleted...

So do I, but it is hard to agree on removing details like this.
#7655 at least eliminates the issue.

PR #7559, opened 2019-12-25
Gerrit: fix server version parsing

### LINKED ISSUE #7521: [NBug] Version string portion was too short or too long.

## Error Details
```
System.ArgumentException: Version string portion was too short or too long.
   at System.Version.VersionResult.SetFailure(ParseFailureKind failure, String argument)
   at System.Version.TryParseVersion(String version, VersionResult& result)
   at System.Version.Parse(String input)
   at Gerrit.FormGerritBase.OnLoad(EventArgs e)
   at Gerrit.FormGerritPublish.OnLoad(EventArgs e)
```

## Additional information
Trying to publish to Gerrit

## Environment
- Git Extensions 3.3.0.7719
- Git 2.24.1.windows.2
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

dfev77:
Do not parse version if request to retrieve it was unsuccessful

Fixes #7521

This is just a quick fix, see notes below.

### After

Currently GitExt makes the following incorrect assessments about the Gerrit server and further changes are required to have a proper Gerrit integration (separate topic):
 - server is available under https if protocol not specified
 - port is the default protocol configuration

In our case ".gerrit" configuration looks like 
`[gerrit]
host=repo.example.lan
port=1234
defaultremote=origin
project=reponame`
while the gerrit is available at "http://repo.example.lan:8080"

## Test methodology

- Manually tested

## Test environment(s)

- GIT 2.23.0
- Windows 10
- Gerrit 2.15.3-1-g047eac59e7-dirty

### DIFF
--- a/Plugins/Gerrit/FormGerritBase.cs
+++ b/Plugins/Gerrit/FormGerritBase.cs
@@ -1,15 +1,12 @@
 ﻿using System;
 using GitUI;
 using GitUIPluginInterfaces;
-using RestSharp;
 
 namespace Gerrit
 {
     public class FormGerritBase : GitExtensionsForm
     {
         protected GerritSettings Settings { get; private set; }
-        protected Version Version { get; private set; }
-        protected readonly RestClient client = new RestClient();
         protected readonly IGitUICommands UICommands;
         protected IGitModule Module => UICommands.GitModule;
 
@@ -39,34 +36,7 @@ namespace Gerrit
                 return;
             }
 
-            SetRestClientUrl();
-            Version = GetGerritVersion();
-
             base.OnLoad(e);
-        }
-
-        private void SetRestClientUrl()
-        {
-            string host = Settings.Host;
-
-            if (!host.StartsWith("http://") || !host.StartsWith("https://"))
-            {
-                host = "https://" + host;
-            }
-
-            if (!host.EndsWith("/"))
-            {
-                host += "/";
-            }
-
-            client.BaseUrl = new Uri(host);
-        }
-
-        private Version GetGerritVersion()
-        {
-            RestRequest request = new RestRequest("/config/server/version");
-            IRestResponse response = client.Execute(request);
-            return Version.Parse(response.Content.Replace(")]}'", "").Replace("\n", "").Replace("\"", "").Split('-')[0]);
         }
     }
 }
--- a/Plugins/Gerrit/Server/GerritCapabilities.cs
+++ b/Plugins/Gerrit/Server/GerritCapabilities.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using ResourceManager;
+
+namespace Gerrit.Server
+{
+    public class GerritCapabilities : Translate
+    {
+        #region Translation
+        private static readonly TranslationString _publishTypeReview = new TranslationString("For Review");
+        private static readonly TranslationString _publishTypeWip = new TranslationString("Work-in-Progress");
+        private static readonly TranslationString _publishTypePrivate = new TranslationString("Private");
+        private static readonly TranslationString _publishTypeDraft = new TranslationString("Draft");
+        #endregion
+
+        private readonly Func<CommandBuilder> _builderFactory;
+
+        private GerritCapabilities(KeyValuePair<string, string>[] publishTypes,
+            Func<CommandBuilder> builderFactory)
+        {
+            PublishTypes = publishTypes;
+            _builderFactory = builderFactory;
+        }
+
+        public IReadOnlyList<KeyValuePair<string, string>> PublishTypes { get; }
+
+        public CommandBuilder NewBuilder()
+            => _builderFactory();
+
+        public static GerritCapabilities Version2_15 { get; }
+            = new GerritCapabilities(
+                new[]
+                {
+                    new KeyValuePair<string, string>(_publishTypeReview.Text, ""),
+                    new KeyValuePair<string, string>(_publishTypeWip.Text, "wip"),
+                    new KeyValuePair<string, string>(_publishTypePrivate.Text, "private"),
+                },
+                () => new CommandBuilderWithPrivateSupport());
+
+        public static GerritCapabilities OldestVersion { get; }
+            = new GerritCapabilities(
+                new[]
+                {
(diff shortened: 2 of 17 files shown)


lekhmanrus:
LGTM

Plugins/Gerrit/FormGerritBase.cs
```diff
@@ -66,7 +66,15 @@ private Version GetGerritVersion()
         {
             RestRequest request = new RestRequest("/config/server/version");
             IRestResponse response = client.Execute(request);
-            return Version.Parse(response.Content.Replace(")]}'", "").Replace("\n", "").Replace("\"", "").Split('-')[0]);
+            if (response.IsSuccessful)
+            {
+                string versionText = response.Content.Replace(")]}'", "").Replace("\n", "").Replace("\"", "").Split('-')[0];
+                return Version.Parse(versionText);
```
RussKie: Wouldn't it be easier to replace this with a regex?
Something like \d{1,3}\.\d{1,3}(\.\d{1,3}){0,2}

RussKie: GitHub's eaten escaping, dot must be escaped too

Plugins/Gerrit/FormGerritBase.cs
```diff
@@ -66,7 +66,15 @@ private Version GetGerritVersion()
         {
             RestRequest request = new RestRequest("/config/server/version");
             IRestResponse response = client.Execute(request);
-            return Version.Parse(response.Content.Replace(")]}'", "").Replace("\n", "").Replace("\"", "").Split('-')[0]);
+            if (response.IsSuccessful)
```
RussKie: Please invert `if` to reduce nesting

Plugins/Gerrit/FormGerritBase.cs
RussKie: Unless `versionText` conforms to a string parseable to `Version` this will fail again.

This is where a regex may help, because you can enforce the format (though not perfectly): https://docs.microsoft.com/en-us/dotnet/api/system.version.-ctor?view=netframework-4.8

```^(\d*\.\d*(\.\d*){0,2})$```
image
Here I am obviously testing for a string from a beginning and to the end of the line, given I have no idea `/config/server/version` return payload is... 

However `Version.Parse(string)` call must still be wrapped in `try..catch` https://docs.microsoft.com/en-us/dotnet/api/system.version.parse?view=netframework-4.8


So the method should look something like the following:

```cs
            RestRequest request = new RestRequest("/config/server/version");
            IRestResponse response = client.Execute(request);
            if (!response.IsSuccessful)
			{
				return GerritUtil.GerritWithPrivateReviewSupport;
			}

			try
			{
				var m = Regex.Match(..., ...);
				if (m.Success)
				{
					return Version.Parse(m....);
				}
			}
			catch
			{
				// no-op
			}

			return GerritUtil.GerritWithPrivateReviewSupport;
```

I would also encourage you to write tests to test edge-cases.

Plugins/Gerrit/FormGerritBase.cs
dfev77: replaced it with ternary operator

Plugins/Gerrit/FormGerritBase.cs
dfev77: changed to regex matching

Plugins/Gerrit/FormGerritBase.cs
```diff
@@ -66,7 +66,17 @@ private Version GetGerritVersion()
         {
             RestRequest request = new RestRequest("/config/server/version");
             IRestResponse response = client.Execute(request);
-            return Version.Parse(response.Content.Replace(")]}'", "").Replace("\n", "").Replace("\"", "").Split('-')[0]);
+
+            return (response.IsSuccessful
+                && Version.TryParse(ExtractVersion(response.Content), out var version))
+                ? version : GerritUtil.GerritWithPrivateReviewSupport;
```
RussKie: We still need to wrap it in `try..catch`, otherwise we risk crashing the app again, if something is wrong.

dfev77: Looking at possible error sources:
 - rest request handles exceptions - see https://github.com/restsharp/RestSharp/blob/master/RestSharp/RestClient.Sync.cs
 - Version.TryParse does not throw
 - Regex.Match might throw if input is null or regex is invalid.

 However in this last case, as long as we intend to handle it, should we just log it or just show a message box? (need some guidance on the error handling mechanism on GitEx)

Plugins/Gerrit/FormGerritBase.cs
dfev77: @RussKie on testing, the only feasible way would be to extract the version retrieval part to a separate class and mock the RestClient. (
`class GerritServer { 
public GerritServer(IRestClient client, GerritSettings settings)
{...}
public Version GetVersion();
{...}
}`
Then the GerritSettings class would also need some changes  as it not test friendly (cannot be stubbed and requires file parsing under a IGitModule context)

So, unit tests would be indeed nice, but right now seem a little to hard to write. So it will take some time....

RussKie:
My apologies, I missed that it was TryParse. You're correct, a chance of an
exception should be low.
All good then!

Plugins/Gerrit/Server/GerritCapabilities.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Gerrit.Server
+{
+    public class GerritCapabilities : Translate
+    {
+        #region Translation
+        private static readonly TranslationString _publishTypeReview = new TranslationString("For Review");
+        private static readonly TranslationString _publishTypeWip = new TranslationString("Work-in-Progress");
+        private static readonly TranslationString _publishTypePrivate = new TranslationString("Private");
+        private static readonly TranslationString _publishTypeDraft = new TranslationString("Draft");
+        #endregion
```
dfev77: I have moved these from FormGerritPublish.cs but UpdateLocalEnglishTranslations script did not add them to the xlf file.
How to handle this?

RussKie:
Thank you.

If you able to send your original simple fix to release/3.3 branch in the next few days, I will include it in 3.3.1.

PR #7560, opened 2019-12-25
RevisionDiff: Limit parents

gerhardol:
Fixes #7469

## Proposed changes

If more than three parents are selected in the revision grid, only show the diff to the first.
This will avoid that the GUI freezes if Ctrl-A is selected or many are selected with shift-click.
It still allows multi selection in reasonable cases.

### Before

image

### After

image

## Test methodology 
Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #4154, opened 2017-11-14
Show more info in files list of two commit diff


mdonatas:
Think about two situations:
1. You want to merge `master` into your `feature` branch (or vice versa) but you would like to know if others have been working on same files as you did. Maybe to know about possible conflicts in advance or to see if you will need to adapt your changes even if the merge won't conflict.
You select two commits your `feature` `HEAD` and `master` and there are 108 files different and almost all of them from changes introduced in `master`! It's a PITA to find files of interest (the ones you changed) in that list to view their diff.
2. You rebase (but still have old commits accessible (see screenshot for an example)) and experience conflicts. Now you want to know if you didn't screw up. You again select two commits and again - many files have changed, how to easily see the diff you care about (diff of files you changed and not other devs in master).

Tadaaaam (illustrating 2nd case):
![image](https://user-images.githubusercontent.com/483659/32756252-1b76cee4-c8d1-11e7-91b1-e38a35ff08ad.png)

So what this feature does is
1. it finds the `common base` for the two commits
2. picks only files which were changed in both `common base -> 1st commit` and `common base -> 2nd commit` and adds 3 new filtered lists. Basically you can see diffs for those files for what _you_ changed; for what _they_ changed and finally a simple diff between them which is the same as diff in the full list but filtered to common files. I use this third view very often after rebasing or before merge'ing.

If you have questions or want clarifications don't hesitate to ask.
I've been using this feature in my custom builds for ~9 months now and I'm loving it. If there'll be interest I would try to get the code to _production quality_ :)

gerhardol:
Sure interesting. I compare to some base (not necessarily the common ancestor) manually.
How is no base handled?
Is "just" first base or can you select the base? (for example select first-base-second).

(I am doing some cleanups to diff, some are not accepted yet, some are not submitted yet)
Production quality takes time...

mdonatas:
To drive a point home. I had a PR for showing commit info in file history form
![image](https://user-images.githubusercontent.com/483659/37044208-8f0bbc28-216b-11e8-9802-ba9c0a146f73.png)
in the meantime a whole bunch of files got updated during cleanup
![image](https://user-images.githubusercontent.com/483659/37044248-ac16e324-216b-11e8-8227-c00d6658d8d1.png)
(and many more commits such as these)

By selecting my PR commit and latest master I get this:
[...]

PR #7561, opened 2019-12-25
RevisionDiff: Show BASE diff

gerhardol:
Part of #4154

## Proposed changes

If two revisions are selected, show the difference from the first to the BASE revision for the first-selected

This could be achieved by selecting the revisions in the grid, Navigate->Go to common ancestor (remember it), reselect revisions and finding and selecting the BASE commit again. (The last part is the most inconvenient).

Functionality is what I want, but set to WIP:
Based on #7560. 
The setting should be clarified, currently reusing AppSettings.ShowDiffForAllParents. This what I prefer, just updating the documentation.
Another reason to WIP is to get comments from @mdonatas that has implemented a solution that also includes A-BASE and B-BASE, affects a lot more code though.

image

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -623,10 +623,25 @@ namespace GitUI
                     // Get the parents for the selected revision
                     parentRevs = Revision.ParentIds?.Select(item => new GitRevision(item)).ToArray();
                 }
-                else if (revisions.Count > 4)
+                else if (revisions.Count == 2 || revisions.Count > 4)
                 {
-                    // Assume only first to selected is interesting
-                    parentRevs = new GitRevision[1] { revisions.Last() };
+                    // only first -> selected is interesting
+                    parentRevs = new[] { revisions.Last() };
+                    if (AppSettings.ShowDiffForAllParents)
+                    {
+                        // Get base commit, add as parent if unique
+                        Lazy<ObjectId> head = new Lazy<ObjectId>(() => Module.RevParse("HEAD"));
+                        var revA = parentRevs[0].ObjectId;
+                        var revB = Revision.ObjectId;
+                        ObjectId baseRevGuid = Module.GetMergeBase(GetRevisionOrHead(revA, head), GetRevisionOrHead(revB, head));
+                        if (baseRevGuid != null
+                            && baseRevGuid != revA
+                            && baseRevGuid != revB)
+                        {
+                            Array.Resize(ref parentRevs, 2);
+                            parentRevs[1] = new GitRevision(baseRevGuid);
+                        }
+                    }
                 }
                 else
                 {
@@ -669,6 +684,13 @@ namespace GitUI
             }
 
             GitItemStatusesWithParents = tuples;
+
+            return;
+
+            ObjectId GetRevisionOrHead(ObjectId rev, Lazy<ObjectId> head)
+            {
+                return rev.IsArtificial ? head.Value : rev;
+            }
         }
 
         public void SetDiffs(GitRevision selectedRev = null, GitRevision parentRev = null, IReadOnlyList<GitItemStatus> items = null)


mdonatas:
My solution is not prod-ready but I've been using it for several years now and rely on it every few days if not daily.

To make this more useful I've also made it so that diffs for `A-base` and `B-base` would be filtered only to files common between them. It's common that `feature` branch has a few files changed while `master` has many changed from many devs and so a diff has 40+ files and it's a hassle to either filter or scroll to find those few as you want to switch between diffs efficiently.
Continuing on the same thought I've also added another group which is same as the first (diff between `A` and `B`) but also filtered just to the files changed on both sides. I've also considered adding a diff-toolbar button to toggle "common files" but am fine with additional group for my needs.

You can try this out by building this branch https://github.com/mdonatas/gitextensions/tree/custom-build-temp it's relatively up to date to current `master`.

gerhardol:
> p.s. I don't even want to begin to imagine what it will mean for me to rebase my changes once this gets merged

This is small change (look at the diff) that should not affect your merge, if you ever submit a PR.

I looked at your changes, but do not feel that I want to merge that. (Maybe a quick first<->selected switch is sufficient?).
But if you want to review, it will be good.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -620,11 +620,34 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions)
...
+                        // Get base commit
+                        Lazy<ObjectId> head = new Lazy<ObjectId>(() => Module.RevParse("HEAD"));
+                        var anyArtificial = parentRevs[0].IsArtificial || Revision.IsArtificial;
+                        var revA = parentRevs[0].IsArtificial ? head.Value : parentRevs[0].ObjectId;
+                        var revB = Revision.IsArtificial ? head.Value : Revision.ObjectId;
+                        ObjectId baseRevGuid = Module.GetMergeBase(revA, revB);
+                        if (baseRevGuid != null
+                            && baseRevGuid != revA
+                            && baseRevGuid != revB
+                            && (!anyArtificial || baseRevGuid != head.Value))
```
mstv: I think, the last condition can be removed because it is redundant.

mstv:
I'd like to have both additional diffs `A - base` and `B - base`. Filtering for common files is a good idea, too. Could it be done as a `combined diff`?

gerhardol:
> I'd like to have both additional diffs A - base and B - base. Filtering for common files is a good idea, too. Could it be done as a combined diff?

This is what @mdonatas proposes in #4154 and has implemented in 55200b3bb2ccca802e8d57ff5710ab2f28d1d3d3 mdonatas/custom-build-temp

The problem is that we have one commit selected and possibly many parents. To also have BASE->A will need to either change the structure (what is done in the commit) or to add another artificial diff that really means that "RevisionDiff should go looking somewhere else for A and B that was stored when adding the artificial commit". Also a problem with the presentation, how to show that the selected (B) commit is different.
I just want to start with the easy part, just insert another parent (BASE) for now.

The control for diff groups should be enhanced to not expand by default and to not calculate unless expanded if many parents.

GitUI/UserControls/FileStatusList.cs
gerhardol: This was to handle selection of work tree/index and did not work correctly (should show HEAD as BASE). Rewrote.

RussKie:
It feels we need to have a non-UI "differ" implementation that can take a set of revisions and returns a diff.
UI classes are crumbling under the load...

gerhardol:
> It feels we need to have a non-UI "differ" implementation that can take a set of revisions and returns a diff.

The differ is not a problem, it is to keep track of the "selected" commit in the UI.
That is kind of what is implemented in 55200b3bb2ccca802e8d57ff5710ab2f28d1d3d3 
But that is a big change, see if I implement a proposal with another artificial commit -  that will be a hack though.

But that is a later story. I plan to merge this first step in a couple of days (or earlier, if RussKie approves).

gerhardol:
@msftbot merge in 1 day

PR #7569, opened 2019-12-29
Debug: add build branch to title

gerhardol:
## Proposed changes

Add the branch the PR is built from to the window title.
The commit sha is visible in About, but especially if the branch is rebased there are several extra steps to find the branch (and the intended functionality included.

An alternative is to add the branch name in About for 'debug builds.

### Before
image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/AppTitleGenerator.cs
+++ b/GitCommands/AppTitleGenerator.cs
@@ -1,4 +1,5 @@
 using GitCommands.UserRepositoryHistory;
+using GitUIPluginInterfaces;
 
 namespace GitCommands
 {
@@ -21,6 +22,10 @@ namespace GitCommands
     /// </summary>
     public sealed class AppTitleGenerator : IAppTitleGenerator
     {
+#if DEBUG
+        private static string _extraInfo;
+#endif
+
         private readonly IRepositoryDescriptionProvider _description;
 
         public AppTitleGenerator(IRepositoryDescriptionProvider description)
@@ -46,9 +51,27 @@ namespace GitCommands
             var description = _description.Get(workingDir);
 
 #if DEBUG
-            return $"{description} ({branchName}) - Git Extensions [DEBUG]";
+            return $"{description} ({branchName}) - Git Extensions{_extraInfo}";
 #else
             return $"{description} ({branchName}) - Git Extensions";
+#endif
+        }
+
+        public static void Initialise(string sha, string buildBranch)
+        {
+#if DEBUG
+            if (ObjectId.TryParse(sha, out var objectId))
+            {
+                _extraInfo = $" @{objectId.ToShortString()}";
+                if (!string.IsNullOrWhiteSpace(buildBranch))
+                {
+                    _extraInfo += $" ({buildBranch})";
+                }
+            }
+            else
+            {
+                _extraInfo = " [DEBUG]";
+            }
 #endif
         }
     }
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -64,6 +64,7 @@ namespace GitExtensions
 
             // This is done here so these values can be used in the GitGui project but this project is the authority of the values.
             UserEnvironmentInformation.Initialise(ThisAssembly.Git.Sha, ThisAssembly.Git.IsDirty);
+            AppTitleGenerator.Initialise(ThisAssembly.Git.Sha, ThisAssembly.Git.Branch);
 
             // NOTE we perform the rest of the application's startup in another method to defer
             // the JIT processing more types than required to configure NBug.


GitCommands/AppTitleGenerator.cs
```diff
@@ -46,10 +50,17 @@ public string Generate(string workingDir = null, bool isValidWorkingDir = false,
             var description = _description.Get(workingDir);
 
 #if DEBUG
-            return $"{description} ({branchName}) - Git Extensions [DEBUG]";
+            return $"{description} ({branchName}) - Git Extensions [{_buildBranch}]";
```
mstv:
```suggestion
            _buildBranch = _buildBranch ?? "DEBUG";
            return $"{description} ({branchName}) - Git Extensions [{_buildBranch}]";
```
would be bulletproof.

GitCommands/AppTitleGenerator.cs
```diff
@@ -46,10 +50,17 @@ public string Generate(string workingDir = null, bool isValidWorkingDir = false,
...
 #else
             return $"{description} ({branchName}) - Git Extensions";
+#endif
+        }
+
+        public static void Initialise(string buildBranch)
+        {
+#if DEBUG
+            _buildBranch = buildBranch;
```
mstv: Turning this into a property would improve readability.

gerhardol: done. I followed the pattern in UserEnvironmentInformation. (no static field either)

GitCommands/AppTitleGenerator.cs
```diff
@@ -46,10 +50,17 @@ public string Generate(string workingDir = null, bool isValidWorkingDir = false,
             var description = _description.Get(workingDir);
 
 #if DEBUG
-            return $"{description} ({branchName}) - Git Extensions [DEBUG]";
+            return $"{description} ({branchName}) - Git Extensions [{BuildBranch ?? "DEBUG"}]";
```
mstv: I am afraid this will not compile.

GitCommands/AppTitleGenerator.cs
```diff
@@ -46,10 +50,17 @@ public string Generate(string workingDir = null, bool isValidWorkingDir = false,
...
 #else
             return $"{description} ({branchName}) - Git Extensions";
+#endif
+        }
+
+        public static void Initialise(string buildBranch)
```
mstv: I proposed the property in order to replace this method.

GitCommands/AppTitleGenerator.cs
gerhardol: This works - if you request changes is another thing

GitCommands/AppTitleGenerator.cs
gerhardol: If this is changed, the #if DEBUG need to be handled in Program.cs too - is that requested?

GitCommands/AppTitleGenerator.cs
mstv: Perfect if it works. I had problems with similar constructs before.

GitCommands/AppTitleGenerator.cs
mstv: I'd make the property independent from `DEBUG`. Then it will just not be used. @RussKie?

RussKie: I'd like to ponder on this a little. I suspect we can do it slightly differently, but I need some time to investigate.

gerhardol: Fine, I just seen some needs to track my private builds

vbjay:
Instead of title could be in about dialog.

gerhardol:
> Instead of title could be in about dialog.

This alternative is in the Description.
I just want something, if About can be agreed on instead, I will change.

mstv:
It affects only the debug builds. It should be easy for developers to distinguish the running instances.
IMO it's a very good idea to replace the constant string "DEBUG" with something more informative.

pmiossec:
I also agree. The added value is to see it immediately without efforts. So, title is a good place.

gerhardol:
Any reviews, to merge or dismiss this PR?
We have too many PRs open right now...

GitCommands/AppTitleGenerator.cs
mstv: > Any reviews, to merge or dismiss this PR?

The private property `BuildBranch` is pointless IMO. (I don't know which is a better English word if this was too harsh.)
Either a non-`DEBUG` public propery `BuildBranch` or `Initialise`. I prefer the first for readability.

RussKie:
I'll look at it tonight

RussKie:
I'm proposing to show the following:
image

If there is no local branch but there's a remote, it will show as:
image

If there is no branches at all, it will show some git magic:
image


A release build remains unaffected.

GitCommands/AppTitleGenerator.cs
mstv: @RussKie, what do you think about non-`DEBUG` public propery `BuildBranch` vs. `Initialise`?

RussKie: I personally don't like `public static` properties. I don't like `public static` method either, but in this case it feels like a lesser evil. And we already have another object with the same behaviour.

I also entertained an idea of implementing a debug-only version of `IAppTitleGenerator` and inject it, but it felt like more work for little gain. Besides we haven't fleshed out the DI story (MEF could be the way to go, but I haven't figured how it fits with the testing story).

gerhardol:
> I'm proposing to show the following:

This is fine for me, I just want something to differ between the installations I use.

If there are approvals, I will squash and merge in some days to await further comments.

gerhardol:
@msftbot merge in 3 days

RussKie:
I just realised I have broken a test that only runs in Debug... 😒

PR #7585, opened 2019-12-31
Make message textbox bounds clear

### LINKED ISSUE #7581: Message in tag creation dialog not easy to find
## Current behaviour

Selecting a tag type that should allow messages ( annotated and signed tags) show no border on the message textbox and when it is disabled, a disabled background does not show.

## Expected behaviour
Either
- Show a border on textbox when enabled.
- Always show a border and show disabled background of textbox when messages are not allowed.

## Screenshots

image

Yes.  2.51.05

## Environment

- Git Extensions 3.3.0.7719
- Git 2.24.0.windows.2
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

vbjay:
Fixes #7581 

## Proposed changes

- Show border on textbox when user can edit message

### Before
As seen in the issue, the textbox does not show a border

### After
image

## Test methodology 

- Showed  Create Tag dialog
- Selected different tag types

## Test environment(s) 

----


RussKie:
How would it look if we make the border permanent?

vbjay:
Like above. The problem is it looks enabled. I like the fact that if you select a non message tag type, it is clear message is not included.

Issue #7587, opened 2020-01-01
Unit test "`Default_values_are_specified_in_invariant_theme` fails, if path to repository contains a hash sign "#"


c3er:
## Current behaviour

The unit test `Default_values_are_specified_in_invariant_theme` fails as only test, if the path to the local GitExtensions repository contains a hash sign.

My case:

```
D:\eigene dateien\devel\bsp\c#\gitextensions
```

## Expected behaviour

All unit tests, including the mentioned one shall be agnostic about all allowed characters in the path to the local repository.

## Steps to reproduce

1) Clone the GitExtensions repository to any path that contains a hash sign "#"
2) Execute unit test `GitUITests.Theming` -> `AppColorDefaultsTests` -> `Default_values_are_specified_in_invariant_theme`
3) Test runs red

## Screenshots

![grafik](https://user-images.githubusercontent.com/24460532/71636880-40310800-2c37-11ea-8029-1892040ac8c3.png)

## Did this work in previous version of GitExtensions

Unknown

## Environment

- GitExtensions version: Developer build, commit 86d233a1a9aaf3a0511bad315781c58569e5d2e1
- GIT version: 2.24.1.windows.2
- OS version: Windows 10, version 1909
- .NET version: Framework 4.6.1

gerhardol:
https://github.com/gitextensions/gitextensions/blob/master/GitCommands/Settings/AppSettings.cs#L1656

https://docs.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=netframework-4.8

Uri.LocalPath truncates the path
The same method is used in several other situations, expect other issues with # in the path

c3er:
@gerhardol Indeed, I got an error message "Failed to read theme" at starting the debug instance.

c3er:
@gerhardol Couldn't this issue be closed? My fixing PR #7591 was merged into master long ago.

gerhardol:
The PR did not have the "Fixes #7591" line that would have closed this automatically, closing manually

c3er:
I understand

PR #7591, opened 2020-01-02
Improve determination of the GitExt directory (fix of issue #7587)

c3er:
# Fixes

Issue #7587.

## Proposed changes

- Change determination of GitExt loacation from `new Uri(assembly.CodeBase).LocalPath` to `assembly.Location`
- Added some test cases. Though they run also green without the proposed change.

### Before

If the path to the GitExt repository contains a hash sign "#":

- Unit test `Default_values_are_specified_in_invariant_theme` fails.
- Debugging from Visual Studio starts with error message "Failed to read theme"

### After
Mentioned errors are fixed.

## Test methodology 

- Made sure related errors disappeared
- Added some test cases
- Ran all existing unit tests to make sure they are still green
- Searched for other occurrences of `new Uri...` bit did not found anything that causes similar bugs

## Test environment(s) 

- GIT 2.24.1.windows.2
- Windows 10 1909

My actual path to the local Gitext repository is

```
D:\eigene dateien\devel\bsp\c#\gitextensions
```

----


RussKie:
@c3er  I completely missed it, but we need your sign-off in contributors.txt

gerhardol:
Also note 
https://github.com/gitextensions/gitextensions/issues/7587#issuecomment-570042910

URI.Local need to be changed in several other situations. Rules for URI is not the same as for the local file system

c3er:
@gerhardol I searched the code for `new Uri` and other occurrences that involve local path seem not to be affected by truncation. I observed that if the path has the format

```
file:///C:/some#/dir/file.txt
```

the path will e truncated like in the line `var path = new Uri(assembly.CodeBase).LocalPath;`. But if the path has not the file protocol prefix, the path will not be truncated by the Uri constructor. It could be more clean to replace all Uri constructor calls to local paths with more appropriate seeming methods but I hesitated because I don't know how to find and test all related side effects.

@RussKie Shall I make a new pull request for this or how should it be handled?

gerhardol:
> I hesitated because I don't know how to find and test all related side effects.

OK

> Shall I make a new pull request for this or how should it be handled?

Yes

Issue #7523, opened 2019-12-18
Allow custom shell in Console tab


asherber:
## Feature description

In the Settings dialog, I currently have a choice of bash, cmd, or powershell as the shell to run in the Console tab of the Browse dialog.

It would be nice if there were a way for us to specify the path to whatever shell we wanted, including command line arguments.

(I also think that PowerShell Core should be added as one of the presets.)

RussKie:
> In the Settings dialog, I currently have a choice of bash, cmd, or powershell as the shell to run in the Console tab of the Browse dialog.

Could you please explain what you mean? Screenshots would work too.
Because I'm not aware of such functionality in GE. We have a feature request #6460, but it is still open...

pmiossec:
>Because I'm not aware of such functionality in GE. 
See #3403 and more especially for screenshot: https://github.com/gitextensions/gitextensions/pull/3403#issuecomment-265326058

>We have a feature request #6460, but it is still open...

This request is to be able to use another shell when opening a console window from the toolbar

asherber:
Here's the current settings page. This controls which shell gets used in the Console tab of the Browse dialog.

![image](https://user-images.githubusercontent.com/5248041/71089258-70c06e80-216e-11ea-8a98-877a14738db9.png)

As @pmiossec says, #6460 is about the toolbar button, not the Console tab. (Though I suspect it would be nice for the toolbar button to have more options too, and possibly to use the same shell specified for the Console tab. But that's for another time!)

pmiossec:
>(I also think that PowerShell Core should be added as one of the presets.)

After a really quick look, the `pwsh.exe` file is not in the `PATH` on my computer.
So it will be difficult to support it by default.

A custom shell is a better solution.

asherber:
Silly question, but did you install PowerShell Core? It's on my path after installation.

For vanilla installations, the default install dir is also `c:\Program Files\PowerShell\6\pwsh.exe`, where `6` is the current major version number. GE could probe to determine latest version installed.

A custom shell lets the user do whatever he wants, which is good; additional presets simplify the user experience for the most common cases.

pmiossec:
>Silly question, but did you install PowerShell Core? It's on my path after installation.

Yes, I don't remember when (and so I don't how I did it). I've got it in `C:\Program Files\PowerShell\6.0.2` but this folder is not in may PATH and I don't know if I am the exception or if it's the normal case. And that's the problem if we want to add it as an option.

[...]

PR #7600, opened 2020-01-06
WIP: perf: Improve speed of directory change in the console

### LINKED ISSUE #7595: Slow to switch repos after console is initialized
## Current behaviour

When switching repos using the MRU dropdown in the repo browser, it takes much longer for the operation to complete if the console tab has been initialized

## Expected behaviour

Time should be about the same

## Steps to reproduce

- Open GE in a repo
- Use dropdown to switch repos; takes 3-4 seconds (on my system) for repo browser to refresh
- Click on Console tab to init console
- Use dropdown to switch repos; takes ~12 seconds for repo browser to refresh

For most of the time, the browse window just sits there, no change. After ~6 sec, the cursor changes to a wait cursor. A couple of seconds later, the console gets the cd command. After that, the repo browser gets the spinning wait indicator and then the browser refreshes. So it almost seems like there's something very time-consuming going on calculating the cd command.

This happens once the console tab has been initialized, even if it's not the active tab when I switch repos. 

I almost always have the console tab open, so I just thought that GE was slow to change repos. Today I happened to open GE and switch to a couple of different repos before opening the console tab, and I noticed how much faster it was.

Not sure, but probably not. 

## Environment

- Build bc22d9bd138ac0166652d64ac8ece34a05cbb7ad
- Git 2.24.1.windows.2
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

pmiossec:
Fix #7595

## Proposed changes

Sending a lot of backspace in a console is slow
(especially with "cmd" but also with "bash" or "powershell").

Now, use a better strategy to send a char to go to the begin of the console line,
comment the line then evaluate it before launching the "cd" command.

Which is quicker...

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 2ae5475a5d956feeebf64c8d51d39c4c47aad3fa (Dirty)
- Git 2.23.0.windows.1
- .NET Framework 4.8.4075.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3019,7 +3019,7 @@ namespace GitUI.CommandsDialogs
             {
                 if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
                 {
-                    ClearTerminalCommandLineWithCommentAndRunCommand("cd " + posixPath, "#");
+                    ClearTerminalCommandLineWithBackspacesAndRunCommand("cd " + posixPath, "#");
                 }
             }
             else
@@ -3035,16 +3035,15 @@ namespace GitUI.CommandsDialogs
             }
         }
 
-        private void ClearTerminalCommandLineWithCommentAndRunCommand(string command, string comment)
+        private void ClearTerminalCommandLineWithBackspacesAndRunCommand(string command, string comment)
         {
             if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
             {
                 return;
             }
 
-            _terminal.RunningSession.WriteInputTextAsync($"\x01 {comment} {Environment.NewLine}");
-
-            _terminal.RunningSession.WriteInputTextAsync(command + Environment.NewLine);
+            // Clear terminal line by sending 'backspace' characters
+            _terminal.RunningSession.WriteInputTextAsync(new string('\b', 500) + command + Environment.NewLine);
         }
 
         private void ClearTerminalCommandLineWithEscapeAndRunCommand(string command)
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3019,31 +3019,27 @@ namespace GitUI.CommandsDialogs
             {
                 if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
                 {
-                    ClearTerminalCommandLineAndRunCommand("cd " + posixPath);
+                    ClearTerminalCommandLineWithCommentAndRunCommand("cd " + posixPath, "#");
                 }
             }
             else if (AppSettings.ConEmuTerminal.ValueOrDefault == "powershell")
             {
-                ClearTerminalCommandLineAndRunCommand("cd \"" + path + "\"");
+                ClearTerminalCommandLineWithCommentAndRunCommand("cd \"" + path + "\"", "#");
             }
             else
             {
-                ClearTerminalCommandLineAndRunCommand("cd /D \"" + path + "\"");
+                ClearTerminalCommandLineWithCommentAndRunCommand("cd /D \"" + path + "\"", "REM");
             }
         }
 
-        private void ClearTerminalCommandLineAndRunCommand(string command)
+        private void ClearTerminalCommandLineWithCommentAndRunCommand(string command, string comment)
         {
             if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
             {
                 return;
             }
 
-            // Clear terminal line by sending 'backspace' characters
-            for (int i = 0; i < 10000; i++)
-            {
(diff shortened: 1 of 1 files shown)


pmiossec:
@asherber Could you test the artifact produced (once the build finished)?

RussKie:
@asherber can you test the portable build and confirm the fix?

asherber:
I'll take a look later this morning. Looking at the proposed fix, I wonder if it wouldn't be quicker just to clear the line directly. In `cmd` and `powershell`, hitting `Esc` clears the line; in `bash` there are a few different approaches: https://askubuntu.com/a/471023

Of course, this both approach and the proposed fix will cause problems for #7523 if that gets implemented, since GE won't know the appropriate method to clear the line for a custom shell. (Interestingly, `Ctrl+C` seems to work on all three shells.)

vbjay:
Another option is to close the console and start another one.

asherber:
@vbjay There are costs to that as well. For example, if your Powershell profile loads several modules, it can take a few seconds to start the shell. I don't mind incurring that cost once when I initialize the Console tab, but I'd hate to do it every time I switch repos.

pmiossec:
> In cmd and powershell, hitting Esc clears the line; in bash there are a few different approaches:

The api provided by the console control don't permit to send key shortcuts. So that won't work.

The first thing I tried was to send the `ESC` char but that doesn't work for `bash` (even it works for `cmd` and `powershell`).

That's the better solution I found to not kill the console process and start it again (with loosing command line history).

asherber:
Definitely speeds things up in all 3 shells.

In cmd I get the following before the cd (the λ is my custom prompt):

```
λ ^A REM
'☺' is not recognized as an internal or external command,
operable program or batch file.
```
I also think it's odd to see that comment line in the console. Does the ConEmu API let you query things like current command line, or current cursor position? I suspect in most cases the current command line will be empty, and if we can check that first, we can avoid needing to clear it. Or if we can find the current position, we can issue just that number of backspaces, rather than 10K of them.

pmiossec:
>Does the ConEmu API let you query things like current command line, or current cursor position?

Already thougt about that but no.

>In cmd I get the following before the cd

That's because that doesn't work well for you 😢 

Could you try again with this build?

asherber:
Latest build works for all shells, and is faster on cmd and powershell because it doesn't have the extra comment command. 

For bash, though, this is better than existing but still not optimal. What about just improving the current method? The existing code sends 10K backspaces, which is excessive. I would guess that in 90% of cases, there won't be anything on the command line anyway. Wouldn't something like 250 backspaces be sufficient for just about everything else? Is it realistic that users will type more than that and leave it sitting on the command line? The delay for 250 backspaces is probably negligible. 

Another thought: Does it improve things at all to use the current method to send one long string full of backspaces, rather than hundreds or thousands of separate backspaces?

pmiossec:
@asherber another try with all that: 500 backspaces seems acceptable

>Another thought: Does it improve things at all to use the current method to send one long string full of backspaces, rather than hundreds or thousands of separate backspaces?

Seems not change things a lot but better for code cleanness... so :+1:

asherber:
Yes, that's a much improved experience on bash! It's a hair slower than cmd/powershell, as one would expect, but significantly better than the current code. Thanks!

pmiossec:
@asherber I have pushed another update: an improvement for bash to improve the strategy by commenting the current line.

I think that's definitely the better solution. No slow down and no empty line run.
Have a look!

asherber:
@pmiossec I'll take a look later this morning. 

I was also poking around https://github.com/Maximus5/conemu-inside/blob/master/ConEmuWinForms/ConEmuSession.cs and I noticed the `ExecuteGuiMacroTextSync()` method, which in combination with the [Keys](https://conemu.github.io/en/GuiMacro.html#Keys) macro would let you send `Ctrl+` combinations. You could use this with `Ctrl+A, Ctrl+K` to quickly clear the `bash` command line. So something like:

```csharp
_terminal.RunningSession.ExecuteGuiMacroTextSync("Keys(""^A"",""^K"")");
// or _terminal.RunningSession.BeginGuiMacro("Keys").WithParam("^A").WithParam("^K").ExecuteAsync()
_terminal.RunningSession.WriteInputTextAsync(command + Environment.NewLine);
```
That should be as fast as what you're doing with `cmd`, and it doesn't need to worry about comments.

asherber:
Sorry it took me me so long to get to this. Yes, what you've got in the PR works as designed and is as fast as cmd/powershell. But I do think that the macro solution in my previous comment would be a nicer looking approach. I confirmed using the ConEmuInside sample project that it works as expected.

pmiossec:
>I confirmed using the ConEmuInside sample project that it works as expected

I won't be available for few days...that let you the time to do a pull request. Please, do it, so that reviewer could choose the best solution.

RussKie:
You will have to fight it, the last one standing gets his implementation in!



j/k 😆

pmiossec:
Superseded by #7612

PR #7562, opened 2019-12-25
Color scheme followup #2

NikolayXHD:
Rename some methods, refactor tests

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7609, opened 2020-01-07
Include theming related files into installer

### LINKED ISSUE #7571: Default theme files generated too late
## Current behaviour

GE is missing the theme file on first start (with old settings).

## Expected behaviour

Default theme files are generated early enough.

## Steps to reproduce

- delete folder `%APPDATA%\Roaming\GitExtensions\GitExtensions\Themes`
- start GE
```
---------------------------
Failed to load theme
---------------------------
File not found
---------------------------
OK   
---------------------------
```

## Screenshots

grafik

no

## Environment

- Git Extensions 3.3.0
- Build 67f7d3f67c33ff87b62bfeb0edd11a1b577dc440
- Git 2.24.1.windows.2
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

NikolayXHD:
Fixes #7571

## Proposed changes

add dark theme -related files to installer

## Tested manually

in powershell
```
> cd Setup
> Set-ExecutionPolicy Unrestricted -Scope Process
> .\Build.cmd
> .\BuildInstallers.cmd
```
run \Setup\bin\Release\GitExtensions.msi

check that `EasyHook.dll`, `EasyHook32.dll` , `EasyHook64.dll`, `win10default.colors`, `dark.colors` are deployed to installation directory

start installed GitExtensions, popup "theme file not found" does not appear anymore

choose dark theme, restart, dark theme is applied.

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/Setup/Product.wxs
+++ b/Setup/Product.wxs
@@ -9,7 +9,7 @@
     <Condition Message=".NET Framework 4.6.1 must be installed prior to installation of Git Extensions.">
       Installed OR WIX_IS_NETFRAMEWORK_461_OR_LATER_INSTALLED
     </Condition>
-    <!-- 
+    <!--
       Read the currently configured value.
      -->
     <Property Id="CURRENT_SSHCLIENT" Secure="yes">
@@ -74,6 +74,7 @@
           <Directory Id="zh_Hans" Name="zh-Hans" />
           <Directory Id="zh_Hant" Name="zh-Hant" />
           <Directory Id="ru_RU" Name="ru-RU" />
+          <Directory Id="ThemesDir" Name="Themes" />
         </Directory>
       </Directory>
       <Directory Id="LocalAppDataFolder">
@@ -204,6 +205,15 @@
       </Component>
       <Component Id="System.Runtime.InteropServices.RuntimeInformation.dll" Guid="*">
         <File Source="..\GitExtensions\bin\$(var.Configuration)\System.Runtime.InteropServices.RuntimeInformation.dll" />
+      </Component>
+      <Component Id="EasyHook.dll" Guid="*">
+        <File Source="..\GitExtensions\bin\$(var.Configuration)\EasyHook.dll" />
+      </Component>
+      <Component Id="EasyHook32.dll" Guid="*">
+        <File Source="..\GitExtensions\bin\$(var.Configuration)\EasyHook32.dll" />
+      </Component>
+      <Component Id="EasyHook64.dll" Guid="*">
+        <File Source="..\GitExtensions\bin\$(var.Configuration)\EasyHook64.dll" />
       </Component>
       <Component Id="Microsoft.Win32.Primitives.dll" Guid="*">
         <File Source="..\GitExtensions\bin\$(var.Configuration)\Microsoft.Win32.Primitives.dll" />
@@ -576,6 +586,14 @@
         <RegistryKey Root="HKCR" Key="github-mac\shell\open\command">
           <RegistryValue Value="&quot;[INSTALLDIR]GitExtensions.exe&quot; %1" Type="string" />
         </RegistryKey>
+      </Component>
+    </DirectoryRef>
+    <DirectoryRef Id="ThemesDir">
+      <Component Id="win10default.colors" Guid="*">
+        <File Source="..\GitExtensions\bin\$(var.Configuration)\Themes\win10default.colors" />
+      </Component>
+      <Component Id="dark.colors" Guid="*">
+        <File Source="..\GitExtensions\bin\$(var.Configuration)\Themes\dark.colors" />
       </Component>
     </DirectoryRef>
     <DirectoryRef Id="PluginsDir">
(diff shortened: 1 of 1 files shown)


Setup/Product.wxs
```diff
@@ -931,6 +941,14 @@
         <File Source="..\GitExtensions\Bin\$(var.Configuration)\ConEmu\GitShowBranch.cmd" />
       </Component>
     </DirectoryRef>
+    <DirectoryRef Id="DirThemes">
+      <Component Id="win10default.colors" Guid="*">
+        <File Source="..\GitExtensions\Bin\$(var.Configuration)\Themes\win10default.colors" />
```
RussKie:
```suggestion
        <File Source="..\GitExtensions\bin\$(var.Configuration)\Themes\win10default.colors" />
```

Setup/Product.wxs
```diff
@@ -931,6 +941,14 @@
...
+      </Component>
+      <Component Id="dark.colors" Guid="*">
+        <File Source="..\GitExtensions\Bin\$(var.Configuration)\Themes\dark.colors" />
```
RussKie:
```suggestion
        <File Source="..\GitExtensions\bin\$(var.Configuration)\Themes\dark.colors" />
```

pmiossec: Maybe we could create a variable containing "..\GitExtensions\bin\$(var.Configuration)" and use it everywhere...

It seems it could be done using:

    <?define BuildOutputPath= "..\GitExtensions\bin\$(var.Configuration)" ?>

RussKie: We are getting better at modifying the installer :)

RussKie:
Just to confirm - you generated an installer locally and tested it?

Setup/Product.wxs
NikolayXHD: It's not straightforward, because translation-related files are added automatically

NikolayXHD:
> Just to confirm - you generated an installer locally and tested it?

yes

Setup/Product.wxs
```diff
@@ -1581,6 +1599,11 @@
       <ComponentRef Id="Microsoft.VisualStudio.Threading.dll" />
       <ComponentRef Id="Microsoft.VisualStudio.Validation.dll" />
       <ComponentRef Id="System.Runtime.InteropServices.RuntimeInformation.dll" />
+      <ComponentRef Id="EasyHook.dll" />
+      <ComponentRef Id="EasyHook32.dll" />
+      <ComponentRef Id="EasyHook64.dll" />
+      <ComponentRef Id="win10default.colors" />
+      <ComponentRef Id="dark.colors" />
```
RussKie: Are these getting deployed to the app's installation folder (i.e. under Program Files)? If so, users won't be able to modify them because the folder is write-protected.
Themes should be deployed to the `LocalAppDataOrganizationDir` folder.

NikolayXHD: My motivation was to have non-editable "technical" theme *win10default.colors* in application directory, which is only supposed to be used in color transformation calculations.

Moving to user data directory is done at runtime by `ThemeDeployment`, which should copy themes other than *win10default.colors* to user data directory.

NikolayXHD: Now that you asked, I doubt it's a good idea to do at runtime, because it's indeed installation-specific detail, so should be implemented in installer.

NikolayXHD: In https://github.com/gitextensions/gitextensions/pull/7562 I am removing possibility to edit theme from GitExtensions, until it's reimplemented as plugin.

Maybe it's ok to have all themes non-editable until then?

RussKie: If we put a theme under `C:\Users\<user>\AppData\Roaming\GitExtensions\GitExtensions` will GE load it?
If so, I think we should update the installer to put it in the right place. Otherwise we will be generating unnecessary work for us in the future.

RussKie: The installer is already configured to install files under `LocalAppDataOrganizationDir` (e.g. the PluginManager`, I think it should be relatively easy to send the themes to that location (or `LocalAppDataOrganizationDir\Themes`).

pmiossec: >Now that you asked, I doubt it was a good idea to do at runtime

Except if you have in mind the support of themes in the portable version...

RussKie: The installer has nothing to do with the portable version though. The portable version is created by a different script.
But your note is important to keep in mind - theme files could be in the app's folder (for a portable version) or under the AppData folder (an installed version).

NikolayXHD: This user data path distinction between app / portable is supported in GitExtensions codebase,
https://github.com/gitextensions/gitextensions/blob/ca9124627be4fea07a380a6a01725763ea585fa5/GitCommands/Settings/AppSettings.cs#L80

RussKie: 👍 thank you.
It is late, can't think straight now.

So let's update the installer to deploy to the AppData\Roaming\GitExtensions\GitExtensions and we're good to go

gerhardol: I prefer to have the default themes in the installation, otherwise may custom changes be overwritten at updates.

NikolayXHD: What I did initially to preserve user changes:
- copy from installation directory to AppData\Roaming
- backup any existing theme files
- backup in such a way that previous backups are not lost

I agree with @RussKie it turned out overly complicated.

I suggest to take following approach:

All theme files are installed to installation directory, they are not supposed to be changed by user.

When user edits the theme and saves the changes, they are saved in writeable AppData directory.

When we present themes list in dropdown, we somehow indicate the difference between built-in themes and custom user themes.

This way
- user's work is safe, customized theme files are never silently overwritten
- no need to spam hard drive with backups
- user is unlikely to accidentally overwrite built-in themes because they lie in Installation directory, while
save / load file dialog will point to AppData by default.

gerhardol: > This way

Optimal solution in my opinion.
What we have is maybe good enough though.

RussKie: Thank you all, good points, it makes sense.

Issue #7415, opened 2019-11-13
Bad things happen when you switch repositories (error in PATH)


LinoBarreca:
## Current behaviour
The folder stays the same (see screenshot).
I think it might mess with the repository if it launches git commands thinking he's on the right path (didn't try since it's dangerous)

## Expected behaviour
Should work.

## Steps to reproduce
Open program.
Open repository.
Do something.
Go to console.
Do something.
Click on Start - Recent repositories - choose one.
Look at the terminal.
GitExtensions issues a "CD" command but it's a wrong path because it includes the drive so the working directory stays the same.

## Screenshots
![image](https://user-images.githubusercontent.com/897723/68764428-f9516980-061a-11ea-9230-fc6e5c450ea7.png)

## Did this work in previous version of GitExtensions
Don't know.

## Environment
Same as #7413 

## Diagnostics
not relevant

pmiossec:
The problem is more than the command to change the directory must be dependent of the shell run inside the console tab.

For the moment only the one for the `bash` shell has been implemented.

pmiossec:
Looking at the code (https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormBrowse.cs#L3004), that's already implemented.

Did you change the shell directly from inside the console tab? In this case, that impossible to detect.

LinoBarreca:
I didn't change anything in settings.
Default installation of gitextensions.
Opened gitextensions.
Opened a repository.
Did a rename.
I used the console to issue the command that you read above (I didn't know if I could do that through the gui and how) and then I changed repository through menu.

that "cd" wasn't from me gitextensions tried to "CD" to the new repository but used the wrong path.

LinoBarreca:
what I did, however, with gitextension closed, is changing the visual studio code default shell to bash.
Maybe this is saved in the repository somewhere?
Maybe when I open the console tab you start the git bash instead of cmd?

pmiossec:
> it's a wrong path because it includes the drive

The problem is not due to the drive because that the format of the posix paths used by bash.
The problem is elsewhere and we must figure it where.
Perhaps the space that is not escaped by a `\`?

You could try different things by trying yourself the command until you understand what is the **real** problem...

pmiossec:
But surrounding the path by double quotes must fix the problem... (and we should probably do that...)

LinoBarreca:
> 
> 
> > it's a wrong path because it includes the drive
> 
[...]

PR #7612, opened 2020-01-08
Improve speed of directory change in console tab

### LINKED ISSUE #7595: Slow to switch repos after console is initialized

asherber:
Fixes #7595 
Alternative to #7600

## Proposed changes

Existing code sends 10K backspaces to clear the current console line, which is slow. For `cmd` and `powershell`, we can send `Esc` to clear the line; for `bash` we can use a ConEmu macro sequence to send Ctrl+A, Ctrl+K.
## Test methodology 

- Manual

## Test environment(s) 

- Git 2.24.1.windows.2
- Windows 10 v 1903 (OS Build 18362.535)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3015,27 +3015,27 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            if (AppSettings.ConEmuTerminal.ValueOrDefault == "bash")
+            switch (AppSettings.ConEmuTerminal.ValueOrDefault)
             {
-                if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
-                {
-                    ClearTerminalCommandLineWithMacroAndRunCommand("cd " + posixPath, "#");
-                }
-            }
-            else
-            {
-                if (AppSettings.ConEmuTerminal.ValueOrDefault == "cmd")
-                {
-                    ClearTerminalCommandLineWithEscapeAndRunCommand("cd /D \"" + path + "\"");
-                }
-                else
-                {
-                    ClearTerminalCommandLineWithEscapeAndRunCommand("cd \"" + path + "\"");
-                }
+                case "bash":
+                    if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
+                    {
+                        ClearTerminalCommandLineWithMacroAndRunCommand($"cd {posixPath}");
+                    }
+
+                    break;
+                case "cmd":
+                    ClearTerminalCommandLineWithEscapeAndRunCommand($"cd /D \"{path}\"");
+                    break;
+                case "powershell":
+                    ClearTerminalCommandLineWithEscapeAndRunCommand($"cd \"{path}\"");
+                    break;
+                default:
+                    break;
             }
         }
 
-        private void ClearTerminalCommandLineWithMacroAndRunCommand(string command, string shellCommentString)
+        private void ClearTerminalCommandLineWithMacroAndRunCommand(string command)
         {
             if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
             {
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3019,7 +3019,7 @@ namespace GitUI.CommandsDialogs
             {
                 if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
                 {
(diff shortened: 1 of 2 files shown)


asherber:
I'm not really sure why there's a test failure -- I don't think I changed anything that would affect that.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3019,33 +3019,42 @@ public void ChangeTerminalActiveFolder(string path)
...
                 }
             }
-            else if (AppSettings.ConEmuTerminal.ValueOrDefault == "powershell")
-            {
-                ClearTerminalCommandLineAndRunCommand("cd \"" + path + "\"");
-            }
             else
             {
-                ClearTerminalCommandLineAndRunCommand("cd /D \"" + path + "\"");
+                if (AppSettings.ConEmuTerminal.ValueOrDefault == "cmd")
```
RussKie: Why not `else if`?

In fact why not change it to a `switch` statement?
```cs
                switch (AppSettings.ConEmuTerminal.ValueOrDefault)
                {
                    case "cmd":
						ClearTerminalCommandLineWithEscapeAndRunCommand("cd /D \"" + path + "\"");
                        return;

                    case "powershell":
						ClearTerminalCommandLineWithEscapeAndRunCommand("cd \"" + path + "\"");
                        return;

                    default:
                        ClearTerminalCommandLineWithMacroAndRunCommand("cd " + posixPath, "#");
                        return;
                }
```

asherber: Thanks, done. I made the default a no-op, in case we get some unexpected value.

asherber:
Note: I also added quotes around the posix path, to help with #7415.

RussKie:
> Note: I also added quotes around the posix path, to help with #7415.

A slight scope creep, but I propose we refactor the code to use `QuoteNE()` extension method.

asherber:
Okay, done. (Although I note that the method already has a null check.)

RussKie:
@asherber please remove WIP when you think it is complete.

@pmiossec I leave it with you.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3015,37 +3015,46 @@ public void ChangeTerminalActiveFolder(string path)
...
+                    if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
+                    {
+                        ClearTerminalCommandLineWithMacroAndRunCommand($"cd {posixPath.QuoteNE()}");
+                    }
+
+                    break;
+                case "cmd":
+                    ClearTerminalCommandLineWithEscapeAndRunCommand($"cd /D {path.QuoteNE()}");
+                    break;
+                case "powershell":
```
pmiossec: I think I would like to be the default behavior. That way, there is a probability that work out of the box if we add another shell.

What do you think about it?

asherber: I don't have a strong opinion here. I did open an issue to request custom shells (or at least add `pwsh`), but I suspect that will require other changes in this code that can't really be planned for at this point. I can see your case for providing a default, but also think there's a case for no-op, since right now we are only expecting three values. So I could go either way.

RussKie: I'm happy with the current proposal.

pmiossec:
Thanks!

pmiossec:
@asherber I think that you are in the best position to continue by fixing #7523 and add `pwsh` in the list...

asherber:
Thanks, @pmiossec , I will look into it.

PR #7620, opened 2020-01-09
More complete Dispose

mstv:
## Proposed changes

- dispose dangling `_findAndReplaceForm`
- dispose `_cancellationTokenSequence` also if `components` are `null`
- add `_loadingControlAsync` and `_loadingControlSync` only once to `Controls`
- consider every field in `RevisionGridControl.Dispose`
- dispose dangling `_revisionFilter`

----

### DIFF
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -39,6 +39,14 @@ namespace GitUI.Editor
             InitializeComponent();
             InitializeComplete();
 
+            Disposed += (sender, e) =>
+            {
+                //// _diffHighlightService not disposable
+                //// _lineNumbersControl not disposable
+                //// _currentViewPositionCache not disposable
+                _findAndReplaceForm.Dispose();
+            };
+
             _currentViewPositionCache = new CurrentViewPositionCache(this);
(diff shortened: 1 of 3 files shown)


GitUI/Editor/FileViewerInternal.cs
```diff
@@ -36,6 +36,14 @@ public partial class FileViewerInternal : GitModuleControl, IFileViewer
 
         public FileViewerInternal()
         {
+            Disposed += (sender, e) =>
+            {
+                //// _diffHighlightService not disposable
+                //// _lineNumbersControl not disposable
+                //// _currentViewPositionCache not disposable
+                _findAndReplaceForm.Dispose();
+            };
```
sharwell: 💡 This should be registered after the `Initialize` methods.

mstv: Why? In order to be safe, the cleanup should be registered before something can go wrong.

sharwell: If something goes wrong in the constructor (i.e. an exception is thrown), this event will not be called.

mstv: If you like. Moved after the `Initialize` methods and rebased.

PR #7621, opened 2020-01-09
Dispose objects created by TranslationTest

mstv:
## Proposed changes

- Dispose objects created by TranslationTest

----

### DIFF
--- a/UnitTests/GitUITests/TranslationTest.cs
+++ b/UnitTests/GitUITests/TranslationTest.cs
@@ -42,10 +42,11 @@ namespace GitUITests
                 {
                     try
                     {
-                        var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
-
-                        obj.AddTranslationItems(translation);
-                        obj.TranslateItems(translation);
+                        using (var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type))
+                        {
+                            obj.AddTranslationItems(translation);
+                            obj.TranslateItems(translation);
+                        }
                     }
                     catch (Exception ex)
                     {
--- a/GitUI/CommandsDialogs/BrowseDialog/MenuCommandsBase.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/MenuCommandsBase.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using GitCommands;
 using ResourceManager;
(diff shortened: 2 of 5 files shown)


mstv:
> @sharwell: I'd like to see a follow-up issue filed to consider making ITranslate implement IDisposable (https://github.com/gitextensions/gitextensions/pull/7394#issuecomment-571154150)

`public interface ITranslate : IDisposable { ... }` does not suffice:
- `GitPluginBase` and `Translate` do not implement `Dispose`
- the following does not compile:
```
var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
obj?.Dispose();
```
error message:
`'ITranslate' does not contain a definition for 'Dispose' and no accessible extension method 'Dispose' accepting a first argument of type 'ITranslate' could be found (are you missing a using directive or an assembly reference?)`

sharwell:
> public interface ITranslate : IDisposable { ... } does not suffice:
>
> GitPluginBase and Translate do not implement Dispose

The change would mean updating those two types to implement `Dispose` (which could be an empty method).

mstv:
What about the not compiling
```
var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
obj.AddTranslationItems(translation);
...
obj?.Dispose();
```
?

mstv:
```
var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
obj.AddTranslationItems(translation);
...
obj?.Dispose();
```
compiled after I made the other assemblies compile by implementing `IDisposabe`.

Though is it worth to add several empty `Dispose`?

UnitTests/GitUITests/TranslationTest.cs
```diff
@@ -40,9 +40,10 @@ public void CreateInstanceOfClass()
 
                 foreach (var type in types)
                 {
+                    ITranslate obj = null;
                     try
                     {
-                        var obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
+                        obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
```
sharwell: 💡  If `ITranslate` implements `IDisposable`, then this can use a `using` statement instead of `try`/`finally`. Otherwise there's not much benefit to making `ITranslate` disposable. 😄

RussKie: I'm pretty sure we can write this as follows and remove the `finally`
```suggestion
                        using ITranslate obj = (ITranslate)TranslationUtil.CreateInstanceOfClass(type);
```

mstv: Already in progress...
I had the hope I could improve the <censored /> ScriptRunnerTests before.

mstv: I hope, next time I will remember myself that C# has a (partial) analogon for C++ smart pointers (although `using` does not help with ctor/Dispose unfortunately). 😃

Issue #7382, opened 2019-11-04
Appveyor tests hang/fail; tests with forms fail in debugger


mstv:
## Current behaviour

Similar to #6172, builds intermittently hang and get killed by the AppVeyor timeout after 1 hour.
However the same builds succeed on re-run.

Additionally, some tests fail intermittently.

## Expected behaviour

All builds finish within reasonable time (~15min) before the timeout.

## Environment

- Git Extensions 3.3.0
- Build 1efcf3fc75f583242f71dbe12f326aef00b6a42a (both Debug & Release)
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

## Screenshots

![grafik](https://user-images.githubusercontent.com/36601201/68159254-b411a580-ff51-11e9-8429-d5bb866d80f8.png)
![grafik](https://user-images.githubusercontent.com/36601201/68159770-b88a8e00-ff52-11e9-8ac4-0dd0458f5eab.png)

## Diagnostics

I started the tests in the debugger in order to check the callstacks in case of `Debug` assertions.
But one test did not close FormBrowse. Then I could not open the `About` dialog. `Repository | Close` did not open the Dashboard, just removed the regarding main menu items. `Start | Exit` let the testcase finish.
The following tests failed with `System.Runtime.InteropServices.InvalidComObjectException: COM object disconnected`:
```
RepoObjectTree_moving_first_up_and_last_down_does_nothing
RepoObjectTree_moving_node_across_hidden_trees_skips_them
RepoObjectTree_moving_node_legally_moves_it
RepoObjectTree_should_show_all_submodules
```
Many other tests which open forms failed the same way.
I guess the forms are loaded too slowly when the tests are executed in the debugger.

Callstacks:
```
 	GitUI.dll!GitUI.GitUICommands.StartBrowseDialog(System.Windows.Forms.IWin32Window owner, string filter, GitUIPluginInterfaces.ObjectId selectedCommit) Line 1153	C#
 	GitUITests.dll!GitUITests.CommandsDialogs.CommitDialog.FormBrowseLeftPanelReorderNodesTest.RunFormTest.AnonymousMethod__13_0() Line 192	C#
[...]

PR #7623, opened 2020-01-10
WIP Resolve race condition in UITest.RunForm

mstv:
Solves parts of #7382

## Proposed changes

- resolve the following race condition in `UITest.RunForm`:
  WaitForIdleAsync could have finished before attempting to open the form.
- wait for potential pending asynchronous tasks triggered by the form
  before running the test and after closing the form
- check for open forms, close/dispose form on failed tests
- output exceptions; verbose output in debug build

- workaround for tests hanging in conjunction with canceled operations:
  Process the message loop before and after the wait.

## Test environment(s)

- AppVeyor
- Windows 10

----

### DIFF
--- a/IntegrationTests/UITests/UITest.cs
+++ b/IntegrationTests/UITests/UITest.cs
@@ -1,58 +1,177 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using CommonTestUtils;
 using GitUI;
+using NUnit.Framework;
 
 namespace GitExtensions.UITests
 {
     public static class UITest
     {
-        public static async Task WaitForIdleAsync()
-        {
-            var idleCompletionSource = new TaskCompletionSource<VoidResult>();
-            Application.Idle += HandleApplicationIdle;
-
-            // Queue an event to make sure we don't stall if the application was already idle
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-            await Task.Yield();
-
-            await idleCompletionSource.Task;
-            Application.Idle -= HandleApplicationIdle;
-
-            void HandleApplicationIdle(object sender, EventArgs e)
-            {
-                idleCompletionSource.TrySetResult(default);
-            }
-        }
-
         public static void RunForm<T>(
-            Action showDialog,
-            Func<T, Task> runAsync)
+            Action showForm,
+            Func<T, Task> runTestAsync)
             where T : Form
         {
+            LoggingService.Log($"{MethodBase.GetCurrentMethod().Name} entry");
+            var idleCompletionSource = new TaskCompletionSource<VoidResult>();
+
+            // Start runAsync before calling showForm.
+            // The latter might block until the form is closed, especially if using Application.Run(form).
+
             // Avoid using ThreadHelper.JoinableTaskFactory for the outermost operation because we don't want the task
             // tracked by its collection. Otherwise, test code would not be able to wait for pending operations to
             // complete.
             var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
             {
-                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-                await WaitForIdleAsync();
-                var dialog = Application.OpenForms.OfType<T>().Single();
+                LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test entry");
+
+                // Wait for the form to be opened by the test thread.
+                await idleCompletionSource.Task;
+                LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test idle");
+                var form = Application.OpenForms.OfType<T>().Single();
+                if (!form.Visible)
+                {
+                    var errorMessage = $"{form.GetType().FullName} {form.Name} should be visible";
+                    form.Dispose();
+                    Assert.Fail(errorMessage);
+                }
+
                 try
                 {
-                    await runAsync(dialog);
+                    // Wait for potential pending asynchronous tasks triggered by the form.
+                    WaitForPendingOperations();
+                    LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test form loaded");
+
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test on main thread");
+
+                    try
+                    {
+                        await runTestAsync(form);
+                        LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test passed");
+                    }
+                    catch (Exception ex)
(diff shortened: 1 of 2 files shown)


IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,177 @@
...
+                    }
+                    catch (Exception inner)
+                    {
+                        // ignore
+                        LoggingService.Log($"{MethodBase.GetCurrentMethod().Name} exception", inner);
+                    }
 
-            // Queue an event to make sure we don't stall if the application was already idle
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-            await Task.Yield();
```
sharwell: 📝 This line guarantees that `WaitForIdleAsync` will not complete before `showDialog` is called.

IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,181 @@
...
+                {
+                    var errorMessage = $"{form.GetType().FullName} {form.Name} should be visible";
+                    form.Dispose();
+                    Assert.Fail(errorMessage);
+                }
+
+                try
+                {
+                    // Wait for potential pending asynchronous tasks triggered by the form.
+                    WaitForPendingOperations();
```
sharwell: ❗️ This should not be part of the test initialization. UI tests wait for the UI operation to complete, but *do not* wait for background operations to complete. If a test depends on background operations completing, the test needs to include that wait as part of the test itself.

IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,181 @@
...
+                    if (form.Visible)
+                    {
+                        actionDescription += "close";
+                        form.Close(); // also calls form.Dispose()
+                    }
+                    else
+                    {
+                        actionDescription += "dispose";
+                        form.Dispose();
+                    }
```
sharwell: ❔ Why not just call `Close`?

IntegrationTests/UITests/UITest.cs
mstv: This may apply for the generic theory of testing.
For GE, which uses asynchronous tasks extensively, we should wait for the background operations by default in order to have stable tests. This little additional runtime will not hurt.
We can add an option to switch back to `await WaitForIdleAsync();` for tests that explicitely want to start before the background tasks have finished.

sharwell: > For GE, which uses asynchronous tasks extensively, we should wait for the background operations by default in order to have stable tests.

We don't need to wait for the operations by default to have stable tests. Most of the tests I wrote had no dependency on operations except for ones tied directly to the UI thread, so waiting was simply unnecessary overhead. For the others, adding a wait as part of the test improves the clarity of the test itself (showing that a specific test covers UI state that might differ from what a user sees when the form first opens).

IntegrationTests/UITests/UITest.cs
mstv: This can come too late because this async task does not count as an "UI operation". Hence `HandleApplicationIdle` can be called before `await Task.Yield()` is reached. Then `await idleCompletionSource.Task;` returns at once.
That's why do register for `Application.Idle` in the non-UI test thread as late as possibe, i.e. right before opening the form, which makes the UI busy.

IntegrationTests/UITests/UITest.cs
mstv: I agree, that it would improve the clarity of the test cases. But I will not review every single UI test. This little overhead will not hurt.

sharwell: You only need to review the tests that have been known to fail.

IntegrationTests/UITests/UITest.cs
sharwell: > Then await idleCompletionSource.Task; returns at once.

Yes, that is allowed. It means there are no pending UI operations.

IntegrationTests/UITests/UITest.cs
mstv: Because I do not and should not know the implementation of `Close`. The documentation just states "Closes the form." That's why the obvious assumption is "nothing will happen if the form has already been closed or never been shown".

sharwell: On review, would should do both.

```
form.Close();
form.Dispose();
```

sharwell: The easiest way to implement this is change the original `var dialog` to be a using statement:

```patch
-var dialog = Application.OpenForms.OfType<T>().Single();
+using var dialog = Application.OpenForms.OfType<T>().Single();
```

IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,181 @@
...
+                    LoggingService.Log($"{description} form disposed");
                 }
-            });
-
-            showDialog();
+            }
+            finally
+            {
+                // Wait for potential pending asynchronous tasks triggered by the form.
+                WaitForPendingOperations();
```
sharwell: 📝 This is already going to happen during test cleanup in `ConfigureJoinableTaskFactoryAttribute.AfterTest`. There is no need to duplicate it here.

IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,181 @@
...
+            // complete.
+            var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
+            {
+                LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test entry");
+
+                // Wait for the form to be opened by the test thread.
+                await idleCompletionSource.Task;
+                LoggingService.Log($"{MethodBase.GetCurrentMethod().Name}.test idle");
+                var form = Application.OpenForms.OfType<T>().Single();
+                if (!form.Visible)
```
sharwell: 📝 Switching to the main thread needs to occur before these APIs are used.

IntegrationTests/UITests/UITest.cs
sharwell: 💭 As far as I can tell, this entire pull request can be simplified to this one line change.

IntegrationTests/UITests/UITest.cs
mstv: I will change this.

This pointed me to another -- maybe the actual -- race condition:
The main thread which runs the form might even not exist yet when the async task is started,
because `Application.Run(form)` will first create it.
Then `await SwitchToMainThreadAsync` should have thrown "no main thread" but it seems to block.

IntegrationTests/UITests/UITest.cs
mstv: `Dispose` does not call neither `Close` nor `OnFormClose`, etc.

sharwell: The test already calls `Close`. You were correct that `Dispose` needs to be called as well. Placing the `var dialog` in a using declaration statement ensures that the call to `Dispose` will occur after the call to `Close`.

GintasS:

PR #7625, opened 2020-01-11
WIP Improve ScriptRunnerTests

mstv:
Improve the most problematic `ScriptRunnerTests` and `FormCommitTests`

- use `AsyncTestHelper.WaitForPendingOperations` for the testcases
- add workaround for `AsyncTestHelper.WaitForPendingOperations` returning before the revisions have been loaded
- assert that no error is displayed by the tested function
- use `formBrowse` as owner for potential subdialogs
- call `Application.DoEvents` before and after `WaitForPendingOperations` as workaround for tests hanging in conjunction with canceled operations

## Test environment(s) 

- AppVeyor
- Windows 10

----


mstv:
Seems to not suffice and to require additional changes from #7394.

IntegrationTests/UITests/UITests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="FluentAssertions" Version="5.2.0" />
     <PackageReference Include="NSubstitute" Version="3.1.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
+    <Reference Include="System.DirectoryServices" />
```
RussKie: What is this needed for?

mstv: Yes, weird. Because of 83f007423dd7b5e48ff270fb98c12517a155bf26.
Maybe since calling new functions:
- `AsyncTestHelper.WaitForPendingOperations`
- `formBrowse.RevisionGridControl.GetSelectedRevisions().Count`
- something else?!?

RussKie: I can see no imports for this assembly, so it looks sus.
Does it work if you remove this reference?

mstv: No, it does not: refer to https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30049049.

RussKie: Ahhh, this looks like the culprit
https://github.com/gitextensions/gitextensions/blob/7273784935f964f74be8fe729e782c0122385892/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs#L636

I think we should create our own enum and ditch the assembly reference. 
```cs
    public enum SortDirection
    {
        /// <summary>
        ///  Sort from smallest to largest. For example, A to Z.
        /// </summary>
        Ascending,

        /// <summary>
        //   Sort from largest to smallest. For example, Z to A.
        /// </summary>
        Descending
    }
```

We will reap multiple benefits with such change (better perf, less memory, and smaller deployment footprint in .NET Core).

IntegrationTests/UITests/Script/ScriptRunnerTests.cs
```diff
@@ -187,13 +181,17 @@ public void RunScript_with_arguments_with_s_option_with_RevisionGrid_with_select
             RunFormTest(formBrowse =>
             {
                 // wait until the revisions are loaded
-                while (formBrowse.RevisionGridControl.LatestSelectedRevision == null)
+                while (formBrowse.RevisionGridControl.GetSelectedRevisions().Count == 0)
                 {
                     Application.DoEvents();
```
sharwell: This is not the correct way to wait for background operations to complete. It should be using `JoinPendingOperationsAsync`.

IntegrationTests/UITests/UITest.cs
```diff
@@ -1,58 +1,127 @@
...
+using CommonTestUtils;
 using GitUI;
+using NUnit.Framework;
 
 namespace GitExtensions.UITests
 {
     public static class UITest
     {
-        public static async Task WaitForIdleAsync()
+        public static void RunForm<T>(
```
sharwell: 📝 The changes in this file are blocked on #7623

UnitTests/CommonTestUtils/AsyncTestHelper.cs
```diff
@@ -34,10 +35,14 @@ public static T RunAndWaitForPendingOperations<T>(Func<Task<T>> asyncMethod, Can
...
         public static void WaitForPendingOperations(TimeSpan timeout)
         {
+            // Workaround for tests hanging in conjunction with canceled operations:
+            // Process the message loop before and after the wait.
+            Application.DoEvents();
             using (var cts = new CancellationTokenSource(timeout))
             {
                 WaitForPendingOperations(cts.Token);
             }
+            Application.DoEvents();
```
mstv: Without `Application.DoEvents`, the tests occasionally hang (until AppVeyor cancels them after 1 hour).

IntegrationTests/UITests/Script/ScriptRunnerTests.cs
```diff
@@ -181,13 +180,21 @@ public void RunScript_with_arguments_with_s_option_with_RevisionGrid_with_select
             RunFormTest(formBrowse =>
             {
                 // wait until the revisions are loaded
-                while (formBrowse.RevisionGridControl.LatestSelectedRevision == null)
+                AsyncTestHelper.WaitForPendingOperations(AsyncTestHelper.UnexpectedTimeout);
```
mstv: This `WaitForPendingOperations` times out very often.
What is different with tests which open FormBrowse?
- `RevisionDataGridView._backgroundProcessingTask` using `ThreadHelper.JoinableTaskContext.Factory`
- FormBrowse is opened using `Application.Run` because `Application.MessageLoop` is `false` when running tests
- something else?

IntegrationTests/UITests/UITest.cs
mstv: I have removed this temporary -- "WIP" in PR title -- experiment.
Still "WIP" due to unclear cause for timeout of `WaitForPendingOperations` (see below).

RussKie:
@mstv please push  37e437be5511331851b3392f77cad998db712b39  directly into the master. This change is a fix, and orthogonal to this RP.

mstv:
With the current state of this PR, the tests timeout quite reprodcibly.
Now, the guessing / trial and error starts over again.

pmiossec:
Off topic: @mstv I'm really impressed by all the great and hard work you are doing 👍

mstv:
`Application.Run` is essential for `GitExtensions.UITests.CommandsDialogs.FormCommitTests.PreserveCommitMessageOnReopen` to pass -- also when run locally.
Else `FormClosed` is called too late.
If `UITest.RunDialog` is forced to use just `mainForm.Show` instead of `Application.Run(mainForm)` the tests block with `mainForm` open.

IntegrationTests/UITests/UITest.cs
```diff
@@ -39,15 +39,17 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
                 try
                 {
+                    Assert.True(form.Visible, $"{form.GetType().FullName} {form.Name} should be visible");
                     await runTestAsync(form);
                 }
                 finally
                 {
                     form.Close();
+
+                    // Do not call form.Dispose() explicitely, because the FormClosed event would be called on the disposed form.
```
gerhardol: explicitly

mstv:
The tests run much better with #7691 and #7692, especially the `ScriptRunnerTests`. 🎉
Thank you very much, @sharwell!

But ROT tests were killed by the test runner in [test](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30391291) run 9 of 10 and in [test](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30392984) run 8 of 10 (not reported as an hang by `JoinableTaskContextNode`).
I have added a `JoinPendingOperationsAsync` before the ROT tests. HTH!

```
1) Failed : GitExtensions.UITests.CommandsDialogs.FormBrowseLeftPanelReorderNodesTest.RepoObjectTree_moving_node_across_hidden_trees_skips_them
Test exceeded Timeout value of 90000ms

1) Failed : GitExtensions.UITests.CommandsDialogs.FormBrowseLeftPanelReorderNodesTest.RepoObjectTree_moving_node_legally_moves_it
Test exceeded Timeout value of 90000ms
```

sharwell:
@mstv I did see there are a few more hangs not addressed by my changes. I'm hoping to review them soon and hopefully one of the proposals you have already targets the underlying cause. 😄

IntegrationTests/UITests/UITest.cs
```diff
@@ -38,18 +39,21 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
-                    await runAsync(dialog);
+                    Assert.True(form.Visible, $"{form.GetType().FullName} {form.Name} should be visible");
+                    await runTestAsync(form);
                 }
                 finally
                 {
-                    dialog.Close();
+                    form.Close();
+
+                    // Do not call form.Dispose() explicitly, because the FormClosed event would be called on the disposed form.
```
sharwell: 💭 I thought we established earlier that we do want to call Dispose. Is that no longer the case?

Build/Run-Tests.ps1
```diff
@@ -5,18 +5,26 @@ $packageConfig = [xml](Get-Content .nuget\packages.config)
...
-    -hideskipped:All `
-    -filter:"+[*]* -[FluentAssertions*]* -[SmartFormat*]* -[nunit*]*" `
-    -excludebyattribute:*.ExcludeFromCodeCoverage* `
-    -excludebyfile:*\*Designer.cs `
-    -output:"OpenCover.GitExtensions.xml" `
-    -target:"nunit3-console.exe" `
-    -targetargs:"$testAssemblies --agents=1 --workers=1 --timeout=90000"
-$testExitCode = $LastExitCode
-Push-AppveyorArtifact "TestResult.xml"
+$testRunCount = 10
```
sharwell: ❗️ This should be 1 by default

```suggestion
$testRunCount = 1
```

UnitTests/GitCommandsTests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -125,41 +125,41 @@ public void Submodule_status_changes_for_top_module()
...
                 result.OurSubmodules.All(i => i.Detailed == null).Should().BeTrue();
             }
 
             [Test]
-            public void Submodule_status_changes_for_first_nested_module()
+            public async Task Submodule_status_changes_for_first_nested_module()
             {
                 var currentModule = _repo2Module;
-                var result = SubmoduleTestHelpers.UpdateSubmoduleStructureAndWaitForResult(_provider, currentModule);
+                var result = await SubmoduleTestHelpers.UpdateSubmoduleStructureAndWaitForResultAsync(_provider, currentModule).ConfigureAwait(false);
```
sharwell: 📝 This should have been `ConfigureAwait(true)`. I'll submit a PR to fix it.

IntegrationTests/UITests/UITest.cs
mstv: Your PRs -- I guess NUnit -- have changed the sequence of Windows events.
I tried it even with `Application.DoEvents` right after `form.Close`. Both with and without this call, the `OnFormClosed` was called after `Dispose`, which would be an unnatural behavior.

Build/Run-Tests.ps1
mstv: This commit "Temporarily ..." will be reverted. This PR is marked "WIP" and labeled "experimental".

IntegrationTests/UITests/UITest.cs
sharwell: Note that the call to Dispose would not appear _here_. It would appear indirectly by turning line 42 into a using declaration statement.

sharwell: If it's still closing the window in a strange order, it might mean the implementation of `showForm` is blocking on the completion of `runTestAsync` (e.g. by pushing a message loop) and somehow calling `Close` itself.

sharwell: If we can't do it in a more convenient location, we could call `Dispose` after we call `test.Join()` below.

mstv: > Note that the call to Dispose would not appear here. It would appear indirectly by turning line 42 into a using declaration statement.

I have learned this meanwhile. That's why I have added a comment to line 42 that `Dispose` is called by `Close` (or by `Form.ShowDialog`).

> If it's still closing the window in a strange order, it might mean the implementation of `showForm` is blocking on the completion of `runTestAsync` (e.g. by pushing a message loop) and somehow calling `Close` itself.

Yes, this is how it felt debugging several variants.

> If we can't do it in a more convenient location, we could call `Dispose` after we call `test.Join()` below.

Will add this as a safety net.

IntegrationTests/UITests/UITest.cs
```diff
@@ -27,32 +30,47 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
 
-            // Join the asynchronous test operation so any exceptions are rethrown on this thread
-            test.Join();
+                // Join the asynchronous test operation so any exceptions are rethrown on this thread.
+                using var cts = new CancellationTokenSource(AsyncTestHelper.UnexpectedTimeout);
+                test.Join(cts.Token);
+            }
+            finally
+            {
+                form?.Dispose();
```
sharwell: 💭 Can we eliminate the `try`/`finally` above, and just call `Close` here?

```suggestion
                form?.Close();
                form?.Dispose();
```

IntegrationTests/UITests/UITest.cs
```diff
@@ -27,32 +30,47 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
                 {
-                    dialog.Close();
-                }
-            });
+                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                    await WaitForIdleAsync();
+                    form = Application.OpenForms.OfType<T>().Single();
+                    try
+                    {
+                        Assert.True(form.Visible, $"{form.GetType().FullName} {form.Name} should be visible");
```
sharwell: 💭 This seems like an unnecessary assertion. Is there a particular problem if the form is not visible?

IntegrationTests/UITests/UITest.cs
mstv: You've lost me. You may find `try finally` ugly. But why should we not dispose the form on exception? If the test passes, `form` will have already been disposed here.

Inserted `form?.Close`.

IntegrationTests/UITests/UITest.cs
mstv: Yes, this might be not necessary assertion. It can be removed when the tests run stable.
Though it is fulfilled for all GE tests.
I am still afraid of a race condition in `RunForm` / `WaitForIdleAsync`, e.g. in case `showForm` does not directly open a Form but runs (asynchronous) commands before.

mstv:
`FormBrowseLeftPanelReorderNodesTest` seems to hang before `runTestAsync` is called.
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30421345

vbjay:
Screenshot_20200128-203128_Email

Lol.

IntegrationTests/UITests/UITest.cs
sharwell: `showForm` is required to open the form synchronously.

IntegrationTests/UITests/UITest.cs
sharwell: > Inserted form?.Close.

The other one can be removed now.

> If the test passes, form will have already been disposed here.

Where is the other call to dispose?

sharwell: > You've lost me. You may find try finally ugly.

It's not that I find try/finally ugly, it's that I prefer to do things in one location where possible.

> But why should we not dispose the form on exception?

If `Close()` throws an exception, something already went wrong and the test is going to fail. Skipping the call to `Dispose()` in this case doesn't seem problematic.

vbjay: Should just be a using block. 

```c#
using (var x as new WhateverForm())
{
    x......;

}

```

Using is a try finally call.

mstv:
> continuous-integration/appveyor/pr — AppVeyor was unable to build non-mergable PR

That's a pity.

Build/Run-Tests.ps1
mstv: This file has become obsolete with the Arcade build.
Any objections to remove it?

IntegrationTests/UITests/UITest.cs
mstv: @vbjay, since `showForm` can block, e.g. due to `Application.RunForm(new FormBrowse())`, the Form cannot be created first (also see the comment I have added to `RunForm`).

`using` makes a variable read-only. The following is not allowed:
```
{
    using Form f = null;
    ...
    f = new Form();
    ...
} // f.Dispose is called
```

mstv: 1. `Form.Close` (or `Form.ShowDialog`) seems to call `Dispose`.
  I agree that we should not rely on this implementation detail and call `Dispose` explicitly.

2.
>> But why should we not dispose the form on exception?

> If `Close()` throws an exception, something already went wrong and the test is going to fail. Skipping the call to `Dispose()` in this case doesn't seem problematic.

Maybe yes, if one test fails, we won't accept it. 
But if something goes wrong, we should try to cleanup as much as possible in order to not affect subsequent tests. I don't like to get lots of false negatives.

3. In theory, `showForm` could throw after it has successfully created the Form. That's why `try finally`.

IntegrationTests/UITests/UITest.cs
mstv: > `showForm` is required to open the form synchronously.

That's clear. But before it opens the Form, it could do some preparation before, e.g. using `RunAsync`.
Is it really ensured that the event `Application.Idle` is not emitted before the main thread starts a message loop by opening a Form?

mstv:
Rebased and found a reproducible hang.
The [build](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30548342) of the previous commit passed.
With just some `Console.WriteLine` added in `UITest.RunForm` (b127bda986aac3df8ab63d659eb01a32e95afa50), it hangs after `Console.WriteLine($"RunForm<{nameof(T)}> entry");` with the tested Form shown.
After several timed out test cases, I have closed the open forms manually in [build](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30548552).

mstv:
> found a reproducible hang.

This hang was caused by a NRE (in the string creation for printf) early in the async task of `UITest.RunForm<T>`.
Then `Application.Run(form)` in the lambda `showForm` hangs because the form does not get closed.
This should not happen in real tests if `SwitchToMainThreadAsync` and `WaitForIdleAsync` do not throw.
I have added a mitigation in [the other trial & error PR](https://github.com/gitextensions/gitextensions/commit/99f1473d7baaff85364703fa8d0a18567da91121).

mstv:
Created dedicated PRs for the relevant parts.

PR #7626, opened 2020-01-11
Add A/B to RevisionDiff revision descriptions

gerhardol:
Discussion in #4154

## Proposed changes
Add _(A)_ and _(B)_ to some revision descriptions in RevisionDiff

### Before

image

image

### After

image

image

## Test methodology 
manual

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -28,8 +28,8 @@ namespace GitUI.CommandsDialogs
             new TranslationString("Are you sure you want to delete the selected file(s)?");
         private readonly TranslationString _deleteFailed = new TranslationString("Delete file failed");
         private readonly TranslationString _multipleDescription = new TranslationString("<multiple>");
-        private readonly TranslationString _selectedRevision = new TranslationString("Selected");
-        private readonly TranslationString _firstRevision = new TranslationString("First");
+        private readonly TranslationString _selectedRevision = new TranslationString("Selected: (B)");
+        private readonly TranslationString _firstRevision = new TranslationString("First: (A)");
 
         private RevisionGridControl _revisionGrid;
         private RevisionFileTreeControl _revisionFileTree;
@@ -686,7 +686,7 @@ namespace GitUI.CommandsDialogs
                 selectedDiffCaptionMenuItem.Visible = true;
                 MenuUtil.SetAsCaptionMenuItem(selectedDiffCaptionMenuItem, DiffContextMenu);
 
-                firstDiffCaptionMenuItem.Text = _firstRevision + ":";
+                firstDiffCaptionMenuItem.Text = _firstRevision.Text;
                 var parentDesc = DescribeSelectedParentRevision(DiffFiles.SelectedItemParents);
                 if (parentDesc.IsNotNullOrWhitespace())
                 {
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -210,18 +210,14 @@ namespace GitUI.CommandsDialogs
         /// Provide a description for the first selected or parent to the "primary" selected last
         /// </summary>
         /// <returns>A description of the selected parent</returns>
-        private string DescribeSelectedParentRevision()
+        private string DescribeSelectedParentRevision(IEnumerable<GitRevision> parents)
         {
-            var parents = DiffFiles.SelectedItemParents
-                .Distinct()
-                .Count();
-
-            if (parents == 1)
+            if (parents.Count() == 1)
             {
-                return DescribeRevision(DiffFiles.SelectedItemParent?.ObjectId, 50);
+                return DescribeRevision(parents.FirstOrDefault()?.ObjectId, 50);
             }
 
-            if (parents > 1)
+            if (parents.Count() > 1)
             {
                 return _multipleDescription.Text;
             }
@@ -686,12 +682,12 @@ namespace GitUI.CommandsDialogs
 
             if (DiffFiles.SelectedItemsWithParent.Any())
             {
-                selectedDiffCaptionMenuItem.Text = _selectedRevision + ": (" + _revisionGrid.DescribeRevision(DiffFiles.Revision, 50) + ")";
+                selectedDiffCaptionMenuItem.Text = _selectedRevision + " (" + DescribeRevision(DiffFiles.Revision?.ObjectId, 50) + ")";
                 selectedDiffCaptionMenuItem.Visible = true;
                 MenuUtil.SetAsCaptionMenuItem(selectedDiffCaptionMenuItem, DiffContextMenu);
 
                 firstDiffCaptionMenuItem.Text = _firstRevision + ":";
-                var parentDesc = DescribeSelectedParentRevision();
+                var parentDesc = DescribeSelectedParentRevision(DiffFiles.SelectedItemParents);
                 if (parentDesc.IsNotNullOrWhitespace())
                 {
                     firstDiffCaptionMenuItem.Text += " (" + parentDesc + ")";
@@ -749,10 +745,10 @@ namespace GitUI.CommandsDialogs
             {
                 resetFileToSelectedToolStripMenuItem.Visible = true;
                 resetFileToSelectedToolStripMenuItem.Text =
-                    _selectedRevision + " (" + _revisionGrid.DescribeRevision(DiffFiles.Revision, 50) + ")";
+                    _selectedRevision + " (" + DescribeRevision(DiffFiles.Revision?.ObjectId, 50) + ")";
             }
 
-            var parents = DiffFiles.SelectedItemParents.Distinct();
+            var parents = DiffFiles.SelectedItemParents;
             if (parents.Count() != 1 || !CanResetToRevision(parents.FirstOrDefault()))
             {
                 resetFileToParentToolStripMenuItem.Visible = false;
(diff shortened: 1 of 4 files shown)


mstv:
Will be helpful!

I'd omit "First"/"Selected" in the diff view because they do not have the same length.
Alternatively, `"First   "` could be used to align these lines.

As already mentioned, I'd like to clearly see which revision is the "selected" one (especially in case of a multi-selection).

gerhardol:
> because they do not have the same length.

To pad will not work well with translations, and this is not a table so the UI should not rely on lengths.
I want to keep First/Selected texts.

> As already mentioned, I'd like to clearly see which revision is the "selected" one (especially in case of a multi-selection).

To put this in the group name in general will be hard to see. You can see the selected in reset/difftool menus too.

I could live with adding to the group name, even if the selected revision will be hard to see, the expand arrow will be lost.
image

Is a better tooltip (on the group name) an acceptable solution?

Other than that, I see no other solution than to add a label to the FileStatusList, using screen estate

RussKie:
I still think we should be consistent in the rendering of hashes, i.e. use the same shortening strategy across the app.

gerhardol:
@mstv Opinion on changing to sha with 8 characters?
RussKie wants that, I find 4 chars unique enough, but if RussKie ask a question this is not clear for normal users.

pmiossec:
>@mstv Opinion on changing to sha with 8 characters?

You will have mine ;)

I always found this abbreviated hash strange. It took me a very long time to figure it out that a commit hash was displayed here for 2 reasons:
* I'm **really** not used to see a hash of only 4 characters (especially when the 4 first characters is just a number because it doesn't contains the letters of an hexa number).
* It is displayed after a long commit message string already difficult to read and often hidden because the column is not large enough by default.

So I think that it will be happier if:
* the abbreviated hash is indeed a little longer
* the hash is displayed before the commit message first line text

So, I prefer something like:

`Diff with @3003f0b7: Include theming related files into installer`

instead of what we now have:

`Diff with: Include theming related files into installer @3003`

gerhardol:
> You will have mine ;)

@pmiossec  See this as I did not want to disturb you!

Fair reasons, will rebase and use that structure

```Diff with (A) @3003f0b7: Include theming related files into installer```

Group name in #4154 - (if that is ever submitted, can of course be changed)
The 'first' (A) commit will often be hidden (the separate 'selected' (B) commit is the important).

```Diff (B) @12345678: Another description <- (A) @3003f0b7: Include theming related files into installer```

**Comment by mstv** [comment]
> Is a better tooltip (on the group name) an acceptable solution?

Yes, the missing tooltip on the (truncated / partly hidden) group name has already been on my list of issues to be created.

> I still think we should be consistent in the rendering of hashes, i.e. use the same shortening strategy across the app.

Agree. I like our (implicit) default of 7 in the RevisionGrid which is the same as used by GH.
In GE sometimes, a magic 8 is specified, sometimes the default is used:
`public unsafe string ToShortString(int length = 10)`.

> ```Diff (B) @12345678: Another description <- (A) @3003f0b7: Include theming related files into installer```

That seems to be the best compromise IMO.

**Comment by gerhardol** [comment]
Found a regression in 88fbf097096c7c0f6735c04ac1b3f306e1d0663c
Updated screenshots

**Comment by gerhardol** [comment]
Example of diff to many, for discussion

image

**Comment by RussKie** [comment]
I like that.
great suggestions 👍

**Comment by pmiossec** [comment]
I find it great also.

I thought about something that could be a little improvement but I don't really know if it could be a good idea, so I propose it as you could judge by yourself.

When displaying a diff:
image

the 'a' and 'b' are displayed in lowercase and followed by the path.

Would it be more understandable from where this 'A' and 'B' come from if we follow this pattern and so display something like:

`Diff (b/@12345678): Another description <- (a/@3003f0b7): Include theming related files into installer`

or `Diff (b/12345678): Another description <- (a/3003f0b7): Include theming related files into installer`

And in the menus:

`First (a/@12345678): Another description` and `Selected (b/3003f0b7): Include theming related files into installer`

**Comment by gerhardol** [comment]
> Would it be more understandable from where this 'A' and 'B' come from if we follow this pattern and so display something like:

Two fewer characters, but not clearer for me. But @ can be dropped anyway. (So another update for the PR.)
Another weak argument is that the "@12345678:my/branch" string is created as one string so like three lines restructuring is needed to change the output.

File: GitUI/UserControls/RevisionGrid/RevisionGridControl.cs [hunk 1]
```diff
@@ -632,21 +632,20 @@ private void HighlightBranch(ObjectId id)
 
         public string DescribeRevision(GitRevision revision, int maxLength = 0)
         {
+            var description = revision.IsArtificial
+                ? string.Empty
+                : "@" + revision.ObjectId.ToShortString(8) + ": ";
```
mstv: I'd avoid the magic number
```suggestion
                : "@" + revision.ObjectId.ToShortString() + ": ";
```
and adapt the default argument of `ToShortString` instead.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -212,16 +212,14 @@ private string DescribeRevision([CanBeNull] ObjectId objectId, int maxLength = 0
         /// <returns>A description of the selected parent</returns>
         private string DescribeSelectedParentRevision()
         {
-            var parents = DiffFiles.SelectedItemParents
-                .Distinct()
```
mstv: You added the `.Distinct()` in 1dc1f818932aea85cb9156be8bf5e16e944d69b7 -- for a reason?

gerhardol: .Distinct() added to SelectedItemParents in 8923cac336dd14bebbe3bc3da288646ea96b8a6a so this no longer needed here
Two no optimal changes, third time's a charm?

gerhardol:
Pushed an update (also two line change for the correction).
Will automerge (one day this time, I have set more when I feel that more comments can come).
If someone wants to change the format further, I may comply, but you have to admit that you are even more undecided than me.

RussKie:
> `Diff (b/@12345678): Another description <- (a/@3003f0b7): Include theming related files into installer`

1. We can definitely drop `@`, it has no added value.
2. As discussed, use `ObjectId.ToShortString()` for the hash
3. `Another description` <-- is this the name of a branch, or the subject line?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -210,18 +210,14 @@ private string DescribeRevision([CanBeNull] ObjectId objectId, int maxLength = 0
...
         /// <returns>A description of the selected parent</returns>
-        private string DescribeSelectedParentRevision()
+        private string DescribeSelectedParentRevision(IEnumerable<GitRevision> parents)
         {
-            var parents = DiffFiles.SelectedItemParents
-                .Distinct()
-                .Count();
-
-            if (parents == 1)
+            if (parents.Count() == 1)
```
RussKie: We are enumerating `parents` multiple times, better to materialise the collection first.

RussKie:
4. Support lower case `a` and `b`, as it aligns with git

mstv:
The default argument of `ObjectId.ToShortString()` needs to be adapted (perhaps in a follow-up PR which eliminates the macic number 8 spread all over the project).

> `Another description <-` is this the name of a branch, or the subject line?

`Another description` is the subject line truncated to 50 characters (another magic number).
`<-` is the separator between `(first) selected` and `base` IIUIC.

vbjay:
Why I pushed to let git do it for us.

Git already makes sure of shortest unique when you call

git rev-parse --short

Or

git log --pretty=format%h

Or

git log --abbrev-commit

Also read up on

git config core.abbrev

gerhardol:
Length of sha: The sha is used here to identify the commit from the selected (and BASE) commits. Two characters is sufficient in 99,9% (or more here), also easy to see and remember moving the eyes from the RevisionDiff tab to the grid. If there are ambiguities, you can deal with this manually (like reselecting).

However, even '@' + 4 chars was too short for the GE CEO, so it need to be clarified, increasing size is one way. I argue that 6 chars should be the longest need, rest just cuts chars from the branch name (if this is a HEAD) or commit subject. 10 chars uses too much

name@sha is more logical than sh: name, but putting the sha first with a fixed length makes this easier to read.

> Support lower case a and b, as it aligns with git

I find it easier to understand that _(A)_ is a name than _(a)_, which seems like parenthesis around text (similar with _[a]_). I vote against the CEO.

> ```(a/@3003f0b7): Include theming related files into installer```

Except for a -> A OK, if someone else agrees. But I have to refactor.

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: There are many situations where this can be optimized just in this file...

gerhardol:
@msftbot merge in 3 days

gerhardol:
No other votes on @RussKies comments, so suggest merging as is.
Auto merge with some days for others to react.
I also intend to rebase/squash before merge.

RussKie:
> too short for the GE CEO, so it need to be clarified

Wow I have been promoted? 🤣 I'm flattered

mstv:
I vote for Igor's points:
1: We can definitely drop `@`, it has no added value.
4: Support lower case `a` and `b`, as it aligns with git.

gerhardol:
@mstv 
@ is dropped already, see the screenshots
Do you persist in changing the format, to what?
I also want compliance to the following: https://github.com/gitextensions/gitextensions/pull/7626#issuecomment-574353852

> The default argument of ObjectId.ToShortString() needs to be adapted (perhaps in a follow-up PR which eliminates the macic number 8 spread all over the project).

I will do that in a separate PR (changing default to 8, removing overrides).

mstv:
> @ is dropped already, see the screenshots

Missed it.

> Do you persist in changing the format, to what?

"Diff with (a) abd1234 ..."

> I also want compliance to the following: https://github.com/gitextensions/gitextensions/pull/7626#issuecomment-574353852

Could reformulate it as a question, please?

gerhardol:

gerhardol:
@russkie
Can you please reset master to get rid of the fixup commit? 

I did not expect the bot to merge when I told it to forget the instructions...

RussKie:
I stepped on this mine before too, the label must be taken off first before telling bot to forget. Generally taking the label off stops the bot from merging.



> Can you please reset master to get rid of the fixup commit?

Done.

PR #7628, opened 2020-01-12
Improve ConfigureJoinableTaskFactoryAttribute

mstv:
Solves parts of #7382

## Proposed changes

- The singleton JTF instance must not be null.
  - Throw InvalidOperationException if `ThreadHelper.JoinableTaskFactory/JoinableTaskContext == null`.
  - Assert in `BeforeTest` that there is no dangling singleton instance,
    which should have been destroyed by `AfterTest`.
- Aggregate follow-up exceptions.
- Reset _threadException to null in `AfterTest`.
- Report detected hangs.
  - Launch debugger if `GE_TEST_LAUNCH_DEBUGGER_ON_HANG` is `1`.
  - Delay execution by `GE_TEST_SLEEP_SECONDS_ON_HANG` on `WaitForPendingOperations` timeout.

## Test environment(s) 

- AppVeyor
- Windows 10

----

### DIFF
--- a/UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
@@ -46,7 +46,19 @@ namespace CommonTestUtils
         {
             using (var cts = new CancellationTokenSource(timeout))
             {
-                WaitForPendingOperations(cts.Token);
+                try
+                {
+                    WaitForPendingOperations(cts.Token);
+                }
+                catch (OperationCanceledException) when (cts.IsCancellationRequested)
+                {
+                    if (int.TryParse(Environment.GetEnvironmentVariable("GE_TEST_SLEEP_SECONDS_ON_HANG"), out var sleepSeconds) && sleepSeconds > 0)
+                    {
+                        Thread.Sleep(TimeSpan.FromSeconds(sleepSeconds));
+                    }
+
+                    throw;
+                }
             }
         }
 
--- a/UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
@@ -18,6 +18,11 @@ namespace CommonTestUtils
 
         public static void RunAndWaitForPendingOperations(Func<Task> asyncMethod, CancellationToken cancellationToken)
         {
+            if (ThreadHelper.JoinableTaskFactory == null)
+            {
+                throw new InvalidOperationException($"Run & wait is pointless without {nameof(ThreadHelper.JoinableTaskFactory)}.");
+            }
+
             ThreadHelper.JoinableTaskFactory.Run(asyncMethod);
 
             WaitForPendingOperations(cancellationToken);
@@ -25,6 +30,11 @@ namespace CommonTestUtils
 
         public static T RunAndWaitForPendingOperations<T>(Func<Task<T>> asyncMethod, CancellationToken cancellationToken)
         {
+            if (ThreadHelper.JoinableTaskFactory == null)
+            {
+                throw new InvalidOperationException($"Run & wait is pointless without {nameof(ThreadHelper.JoinableTaskFactory)}.");
+            }
+
             var result = ThreadHelper.JoinableTaskFactory.Run(asyncMethod);
 
             WaitForPendingOperations(cancellationToken);
@@ -42,7 +52,15 @@ namespace CommonTestUtils
 
         public static void WaitForPendingOperations(CancellationToken cancellationToken)
         {
-            ThreadHelper.JoinableTaskContext?.Factory.Run(() => ThreadHelper.JoinPendingOperationsAsync(cancellationToken));
+            if (ThreadHelper.JoinableTaskContext == null)
+            {
+                throw new InvalidOperationException($"Wait is pointless without {nameof(ThreadHelper.JoinableTaskContext)}.");
+            }
+
+            // Note that ThreadHelper.JoinableTaskContext.Factory must be used to bypass the default behavior of
+            // ThreadHelper.JoinableTaskFactory since the latter adds new tasks to the collection and would therefore
+            // never complete.
+            ThreadHelper.JoinableTaskContext.Factory.Run(() => ThreadHelper.JoinPendingOperationsAsync(cancellationToken));
         }
     }
 }
--- a/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
+++ b/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
@@ -44,7 +44,7 @@ namespace CommonTestUtils
 
             Assert.AreEqual(ApartmentState.STA, Thread.CurrentThread.GetApartmentState());
 
-            // This form created for obtain UI synchronization context only
+            // This form is created to obtain a UI synchronization context only.
             using (new Form())
             {
                 // Store the shared JoinableTaskContext
@@ -72,6 +72,10 @@ namespace CommonTestUtils
 
                 _denyExecutionSynchronizationContext?.ThrowIfSwitchOccurred();
             }
+            catch (Exception ex) when (_threadException != null)
+            {
+                StoreThreadException(ex);
+            }
             finally
             {
                 Application.ThreadException -= HandleApplicationThreadException;
@@ -80,11 +84,16 @@ namespace CommonTestUtils
         }
 
         private void HandleApplicationThreadException(object sender, ThreadExceptionEventArgs e)
+            => StoreThreadException(e.Exception);
+
+        private void StoreThreadException(Exception ex)
         {
-            if (_threadException == null)
+            if (_threadException != null)
             {
-                _threadException = ExceptionDispatchInfo.Capture(e.Exception);
+                ex = new AggregateException(new Exception[] { _threadException.SourceException, ex });
             }
+
+            _threadException = ExceptionDispatchInfo.Capture(ex);
         }
 
         private class DenyExecutionSynchronizationContext : SynchronizationContext
--- a/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
+++ b/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Threading;
@@ -15,6 +16,7 @@ namespace CommonTestUtils
     public sealed class ConfigureJoinableTaskFactoryAttribute : Attribute, ITestAction
     {
         private DenyExecutionSynchronizationContext _denyExecutionSynchronizationContext;
+        private HangReporter _hangReporter;
         private ExceptionDispatchInfo _threadException;
 
         public ActionTargets Targets => ActionTargets.Test;
@@ -49,6 +51,7 @@ namespace CommonTestUtils
             {
                 // Store the shared JoinableTaskContext
                 ThreadHelper.JoinableTaskContext = new JoinableTaskContext();
+                _hangReporter = new HangReporter(ThreadHelper.JoinableTaskContext);
             }
         }
 
@@ -178,6 +181,37 @@ namespace CommonTestUtils
             private static void ThrowFailedTransferExceptionForCapture()
             {
                 throw new InvalidOperationException("Tests cannot use SwitchToMainThreadAsync unless they are marked with ApartmentState.STA.");
+            }
+        }
+
+        private sealed class HangReporter : JoinableTaskContextNode
+        {
+            public HangReporter(JoinableTaskContext context)
+                : base(context)
+            {
+                RegisterOnHangDetected();
+            }
+
+            protected override void OnHangDetected(TimeSpan hangDuration, int notificationCount, Guid hangId)
+            {
+                if (notificationCount > 1)
+                {
+                    return;
+                }
+
+                Console.ForegroundColor = ConsoleColor.Red;
+                Console.WriteLine($"{Environment.NewLine}HANG DETECTED: guid {hangId}{Environment.NewLine}");
+                Console.ResetColor();
+
+                if (Environment.GetEnvironmentVariable("GE_TEST_LAUNCH_DEBUGGER_ON_HANG") != "1")
+                {
+                    return;
+                }
+
+                Console.WriteLine("launching debugger...");
+
+                Debugger.Launch();
+                Debugger.Break();
             }
         }
     }
--- a/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
+++ b/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
@@ -21,6 +21,8 @@ namespace CommonTestUtils
 
         public void BeforeTest(ITest test)
         {
+            Assert.IsNull(ThreadHelper.JoinableTaskContext, "Tests with joinable tasks must not be run in parallel!");
+
             Application.ThreadException += HandleApplicationThreadException;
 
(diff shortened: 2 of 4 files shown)


UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -14,13 +16,25 @@ namespace CommonTestUtils
...
+
         private DenyExecutionSynchronizationContext _denyExecutionSynchronizationContext;
         private ExceptionDispatchInfo _threadException;
 
         public ActionTargets Targets => ActionTargets.Test;
 
+        static ConfigureJoinableTaskFactoryAttribute()
+        {
+            // route unhandled application exceptions to HandleApplicationThreadException
+            Application.SetUnhandledExceptionMode(UnhandledExceptionMode.CatchException);
```
sharwell: ❔ This is the default value. If this line is necessary, do you know where it's being overridden with some other value?

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -14,13 +16,25 @@ namespace CommonTestUtils
     [AttributeUsage(AttributeTargets.Assembly)]
     public sealed class ConfigureJoinableTaskFactoryAttribute : Attribute, ITestAction
     {
+        private readonly bool _checkForUpdatesOriginalValue = AppSettings.CheckForUpdates;
```
sharwell: 💡 I wouldn't bother with trying to restore the value.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -42,7 +56,7 @@ public void BeforeTest(ITest test)
 
             Assert.AreEqual(ApartmentState.STA, Thread.CurrentThread.GetApartmentState());
 
-            // This form created for obtain UI synchronization context only
+            // This form is created to obtain an UI synchronization context only.
```
sharwell:
```suggestion
            // This form is created to obtain a UI synchronization context only.
```

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -56,10 +70,9 @@ public void AfterTest(ITest test)
             {
                 try
                 {
-                    using (var cts = new CancellationTokenSource(AsyncTestHelper.UnexpectedTimeout))
-                    {
-                        ThreadHelper.JoinableTaskContext?.Factory.Run(() => ThreadHelper.JoinPendingOperationsAsync(cts.Token));
-                    }
+                    // Wait for eventual pending operations triggered by the test.
+                    // On timeout, only log this but do not let fail the test.
+                    AsyncTestHelper.WaitForPendingOperations(AsyncTestHelper.UnexpectedTimeout, throwOnTimeout: false);
```
sharwell: ❗️ Failing the test on timeout is absolutely essential. The timeout exists _only_ to avoid test hangs. It is not used to make an incorrectly behaved test pass.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -72,16 +85,29 @@ public void AfterTest(ITest test)
...
+                    LoggingService.Log($"{nameof(AfterTest)} rethrow", _threadException.SourceException);
+                    _threadException.Throw();
+                }
             }
         }
 
         private void HandleApplicationThreadException(object sender, ThreadExceptionEventArgs e)
         {
-            if (_threadException == null)
+            bool ignore = e.Exception.GetType() == typeof(ThreadAbortException);
```
sharwell: ❗️ `ThreadAbortException` is a bug by definition. If one gets thrown, the test must fail.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -72,16 +85,29 @@ public void AfterTest(ITest test)
 
                 _denyExecutionSynchronizationContext?.ThrowIfSwitchOccurred();
             }
+            catch (Exception ex) when (_threadException != null)
+            {
+                // ignore the follow-up exception
+                LoggingService.Log($"{nameof(AfterTest)} ignoring", ex);
+            }
```
sharwell: 💡 It's fine to keep track of both `ex` and `_threadException` when both occur. However, the logging is unnecessary complexity. The exception handler can instead look like this:

```csharp
catch (Exception ex) when (_threadException != null)
{
  // Track both exceptions
  _threadException = new AggregateException(_threadException, ex);
}
```

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -72,16 +85,29 @@ public void AfterTest(ITest test)
...
             finally
             {
+                AppSettings.CheckForUpdates = _checkForUpdatesOriginalValue;
                 Application.ThreadException -= HandleApplicationThreadException;
-                _threadException?.Throw();
+                if (_threadException != null)
+                {
+                    LoggingService.Log($"{nameof(AfterTest)} rethrow", _threadException.SourceException);
+                    _threadException.Throw();
+                }
```
sharwell: 💡 The original code was fine here. The test log will already report the exceptions.

UnitTests/GitUITests/ThreadHelperTests.cs
```diff
@@ -238,14 +238,7 @@ public async Task AllowAwaitForAsynchronousMTATest()
 
         private static void JoinPendingOperations()
```
sharwell: 💡 At this point you can just inline this method.

UnitTests/CommonTestUtils/AsyncTestHelper.cs
```diff
@@ -18,31 +18,56 @@ public static TimeSpan UnexpectedTimeout
...
+
             var result = ThreadHelper.JoinableTaskFactory.Run(asyncMethod);
 
             WaitForPendingOperations(cancellationToken);
 
             return result;
         }
 
-        public static void WaitForPendingOperations(TimeSpan timeout)
+        public static void WaitForPendingOperations(TimeSpan timeout, bool throwOnTimeout = true)
```
sharwell: 💡 The changes to this method are unnecessary. There are no valid places where `throwOnTimeout` can be false.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Yes, `AggregateException` is a very good hint.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Ah, that's a remainder of the background thread in FormBrowse, which you have replaced in the meantime. This thread was aborted intentionally.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: This is a remainder of the debug instrumentation. I am undecided whether we should keep all the debug instrumentation or not.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: > "a UI"

Really not "an UI"?

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: The tests shall not bollix my `AppSettings`. I am afraid they will be stored to `%APPDATA%`.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Since I have set `SetUnhandledExceptionMode`, I have not seen any tests hanging due to a popup "Unhandled **Application** Exception" anymore.
I don't know where it is set. Though I think this is not the default value, especially not for the test runner.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Then we must wrap `OperationCanceledException` at least, because it cancels all tests that folllow.

How can we find out why the pending operations of `ScriptRunnerTests` do not finish on AppVeyor occasionally? `Tasks` cannot be analyzed if the debugger is attached afterwards.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
sharwell: Do you know the name of the test runner process? I can figure out from there where it's being set.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
sharwell: It's not safe to assume the cleanup code will execute. If there is a concern with storage in APPDATA, it needs to be addressed separately. I'll look into this.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
sharwell: Correct. The use of 'a' or 'an' is determined by the pronunciation, not the spelling. In this case, 'UI' is read out loud as 'you eye', with 'you' being a consonant sound that goes with 'a'.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
sharwell:
```patch
-_threadException?.Throw();
+// reset _threadException to null, and throw if it was set during the current test
+Interlocked.Exchange(ref _threadException, null)?.Throw();
```

sharwell: > Tasks cannot be analyzed if the debugger is attached afterwards.

I'd start with this: https://github.com/microsoft/vs-threading/blob/master/doc/async_hang.md

We should also be able to use [`JoinableTaskContextNode`](https://github.com/microsoft/vs-threading/blob/b680de515d6210c4028598b988187c3aed33aabc/src/Microsoft.VisualStudio.Threading/JoinableTaskContextNode.cs) to override the `OnHangDetected` method to obtain further information about delays.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: It's `nunit3-testrunner` or the like.

sharwell: It's set here:
https://github.com/nunit/nunit-console/blob/dfd07ef27cac9082c9728f0e55b84d1237830605/src/NUnitConsole/nunit3-console/app.config#L16-L17

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
sharwell: A new attribute should be added that implements `ITestAction` and applied to the assembly. This attribute should set the value of `AppSettings.SettingsContainer` to a transient container which ignores the setting in APPDATA and also does not persist any changes there. Before each test, the container should be reset to its default values.

mstv: Thank you.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Obviously, it does not work as you expect:

grafik

Perhaps the test assemblies would need their own `app.config`.
I'll restore it.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: `OnHangDetected` works but doesn't give much information. `HangDetails` is hidden by `JoinableTaskContextNode`.
`Debug.Break` has no effect on AppVeyor.
How can I force the creation of a crash dump in order to analyze it offline?

RussKie:
try `Debugger.Launch` followed by `Debugger.Break`

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Removed `Application.SetUnhandledExceptionMode` because it does not help avoiding "unhandled *application* exception" popups.
So what can we do to catch them?

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: Moved to #7654.

vbjay:
image

vbjay:
Still going. Looks like you have a deadlock.

mstv:
I let run the tests ten times in order to catch the case where `WaitForPendingOperation` times out.
At least, the regarding deadlock is detected and the debugger is launched now. Finally, I could store a dump in order to analyze the asynchronous tasks, (which is impossible with the debugger attached at runtime). Afterwards, it takes almost 8 minutes to download < 400 KB from AppVeyor.

UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
mstv: WinDbg does not display very much:
```
(86c.768): Break instruction exception - code 80000003 (first chance)
ntdll!DbgBreakPoint:
00007ffd`e18d3150 cc              int     3
0:019> .load c:\Users\appveyor\Downloads\SosThreadingTools_x64.dll
0:019> !dumpasync
e0434352 Exception in c:\Users\appveyor\Downloads\SosThreadingTools_x64.dll.dumpasync debugger extension.
      PC: 00007ffd`de779159  VA: 00000000`00000000  R/W: 80070002  Parameter: 00000000`00000000
```

mstv:
The analysis of hangs is not the subject of this PR.
I have reverted the temporary changes.
Shall I rebase before the review?

mstv:
Shall I squash the `fixup!` commits before the review, @sharwell?
Do you prefer the author or the reviewer to mark conversations as "resolved"?

RussKie:
I mostly follow these rules:

If I think a comment is resolved and no further action is required (e.g.
there was a typo and I have fixed it) - mark it resolved. It can be marked
"unresolved" anytime.

If it is an ongoing conversation - leave it open.

Also if I think a conversation may be useful in the future (e.g. contains
technical insights or guidelines) I tend to leave it open for ease of
search.

mstv:
This PR is ready. (I have moved out the open issue "unhandled exceptions" and I have created another PR for the out-of-scope functionality regarding `AppSettings`.)

mstv:
Rebased in order to enable FF merge. I want to keep these 4 commits -- don't squash, please.

RussKie:
Merge it :)

PR #7648, opened 2020-01-17
Update to Visual Studio 2019 and C# 8

sharwell:


NetSpell.SpellChecker/Dictionary/WordDictionary.cs
```diff
@@ -14,8 +14,8 @@
 
 namespace NetSpell.SpellChecker.Dictionary
 {
-    using AffixRuleCollection = Dictionary<string, AffixRule>;
-    using PhoneticRuleCollection = List<PhoneticRule>;
+    using AffixRuleCollection = System.Collections.Generic.Dictionary<string, NetSpell.SpellChecker.Dictionary.Affix.AffixRule>;
+    using PhoneticRuleCollection = System.Collections.Generic.List<NetSpell.SpellChecker.Dictionary.Phonetic.PhoneticRule>;
```
RussKie: Why have these changed?

Build/stylecop.json
```diff
@@ -4,7 +4,8 @@
         "namingRules": {
             "allowedHungarianPrefixes": [
                 "ui", "x", "y", "my", "ls", "am", "id", "ie", "vs", "up"
-            ]
+            ],
+            "tupleElementNameCasing": "camelCase"
```
RussKie: I am not overly fond of these change, callsites look worse.
Besides it seems like a noisy change

sharwell: Both styles were in use before, so this was just a consolidation.

NetSpell.SpellChecker/Dictionary/WordDictionary.cs
sharwell: A probable bug in the analyzer, but seemed to have low impact overall.

RussKie: Please revert. I can live with the rest.

Build/stylecop.json
RussKie: Pascal case at callsites feels more C#

NetSpell.SpellChecker/Dictionary/WordDictionary.cs
sharwell: Done

Build/stylecop.json
sharwell: I went through again and confirmed that camelCase is the dominant style in the code base (by far). It follows the view that tuples are more like a bag of parameters than a type with named fields.

PR #7649, opened 2020-01-18
Use Microsoft Arcade to improve build pipelines

RussKie:
## Proposed changes

"Imitation is the sincerest form of flattery". Copy implementations from the Microsoft.DotNet.Arcade.Sdk, that is used by .NET projects.

This SDK offers a number features and provides a number of tasks such as clean, restore, build, rebuild, test, etc.
Which both work from Visual Studio and command line.

I decided against using the actual package for a number of reasons:

1. The package isn't published to nuget.org.
2. The package is being actively developed, and may be causing unnecessary overhead, if the package maintainers modify behaviours.
3. The package contains significantly more functionality, catering for the Microsoft .NET development process.
4. The package would require a significant number of adjustments to the project and its structure, to make use of it.

Taking the limited subset of the SDK provides a greater control over the executed targets and their configurations.

### Notable changes include:
* Builds can now be initiated from commands line via `.\build.cmd`. 
* All artifacts are now redirected to .\artifacts folder, that includes binaries, obj, logs and test results.
* All plugins binaries are places under .\artifacts\bin\GitExtensions\[configuration]\Plugins. This will allow for a more streamlined plugin development.

Supported CLI commands:

```ps
# build with restore and binary log enabled (default set)
# same as supplying '-restore -build -bl'
> .\build.cmd

# default build and build C++ projects
> .\build.cmd -buildNative

# default build and run unit tests
> .\build.cmd -test

# default build and run integration tests
> .\build.cmd -integrationtest

# cleans all artifacts, default bin and obj folders
> .\build.cmd -clean
```

----


RussKie:
Ready for reviews. This is a foundational change, and there is more work ahead.

I'll let it simmer for a few days, and will merge it in, unless objected.

Please pull it down and ride it hard. Review commit by commit.

RussKie:
I've looked at the CodeFactor issues and all of them are related to PowerShell. I'm not planning to address them.

RussKie:
@msftbot merge in 3 days

RussKie:
@mstv would this cause you any significant headache if I merge this?

pmiossec:
@RussKie I have tested and it seems to work as expected.
I have pushed a fixup commit with some nitpickings. Feel free to remove if you don't want it.

But I see that this PR has a lot of merge conflicts. Either I really don't understand what you expect to do or they should be fixed.

mstv:
> Would this cause you any significant headache if I merge this?

I have no concerns, @RussKie. Your branch built on my machine but I forgot to comment.

gerhardol:
I have not tested, not really reviewed
No objections to merging, we will make this work if there are problems

RussKie:
Thank you fellows. I'm going to merge it, hopefully no major interruptions to anyone's process.
I plan on getting the VSIX and the translations working next. 

I think the installer is worth looking at in conjunction with #7656.

/cc: @Johno-ACSLive

mstv:
`<NUnitVersion Condition="'$(NUnitVersion)' == ''">3.12.0</NUnitVersion>`

It seems `NUnitVersion` was set to 3.10.0 before. But we need at least 3.12.0, @RussKie.

mstv:
Fixed by adding `<NUnitVersion Condition="'$(NUnitVersion)' &lt; '3.12.0'">3.12.0</NUnitVersion>`.

mstv:
`<NUnitVersion Condition="'$(NUnitVersion)' &lt; '3.12.0'">3.12.0</NUnitVersion>` does not suffice.
`NUnit.ConsoleRunner` must be handled, too.

RussKie:
bad merge I guess. I'll look into it today

RussKie:
If you struggle with MSBuild and its ways of working, you may find a presentation from @davidwengier useful: https://youtu.be/5HEbsyU5E1g

/cc: @mstv - I remember you were frustrated by it, but I don't remember which thread it was in...

PR #7654, opened 2020-01-19
Add TempAppSettingsAttribute for all tests

mstv:
## Proposed changes

- Add `TempAppSettingsAttribute` for all tests
  in order to set `AppSettings.CheckForUpdates = false`

There is no need to tinker the `AppSettings.SettingsFilePath` because the user's GE settings file is not used but one which is specific for the test runner, e.g. 
`SettingsFilePath: C:\Users\appveyor\AppData\Roaming\NUnit Software\NUnit Engine\GitExtensions.settings`

## Test environment(s) 

- AppVeyor
- Windows 10

----

### DIFF
--- a/IntegrationTests/UITests/Properties/AssemblyInfo.cs
+++ b/IntegrationTests/UITests/Properties/AssemblyInfo.cs
@@ -37,4 +37,5 @@ using CommonTestUtils;
 [assembly: AssemblyFileVersion("1.0.0.0")]
 
 [assembly: ConfigureJoinableTaskFactory]
+[assembly: TestAppSettings]
 [assembly: UseReporter(typeof(NUnitReporter), typeof(AppVeyorReporter), typeof(DiffReporter))]
--- a/UnitTests/CommonTestUtils/Properties/AssemblyInfo.cs
+++ b/UnitTests/CommonTestUtils/Properties/AssemblyInfo.cs
@@ -1,5 +1,6 @@
 ﻿using System.Reflection;
 using System.Runtime.InteropServices;
+using CommonTestUtils;
 
 // General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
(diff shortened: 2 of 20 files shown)


mstv:
This PR is ready.

BTW: After a force-push, it does not suffice to reload the PR in the browser once: Each of the tabs, especially "Conversation" and "Files changed", must be reloaded individually.

mstv:
The test runner uses `C:\Users\appveyor\AppData\Roaming\NUnit Software\NUnit Engine\GitExtensions.settings` on AppVeyor.

RussKie:
What about when tests run locally?
I just want to confirm we won't corrupt the installed app's config.

mstv:
It is similar when run in MSVS2019:
`AppSettings.SettingsFilePath = "C:\\Users\\<user>\\AppData\\Roaming\\Microsoft Corporation\\Microsoft.TestHost.x86\\GitExtensions.settings"`.

RussKie:
All good then, thank you

UnitTests/CommonTestUtils/TempAppSettingsAttribute.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using GitCommands;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+namespace CommonTestUtils
+{
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class TempAppSettingsAttribute : Attribute, ITestAction
```
RussKie: Maybe?
```suggestion
    public sealed class TestAppSettingsAttribute : Attribute, ITestAction
```

mstv: Yes, will rename. The name `TempAppSettingsAttribute` was chosen with the expectation that the `AppSettings.SettingsFilePath` needed to be tinkered.

mstv:
> Can this be used to run tests in parallel too?

No, `AppSettings` is still a singleton.
This one suppresses only the update popup.

mstv:
@msftbot merge in 1 day

Issue #7368, opened 2019-10-27
Overhaul Installer - migrate to Advanced Installer and split into 32/64 bit MSI's


Johno-ACSLive:
## Feature description

WiX is quite demanding to up-keep and therefore simpler MSI creation and management is required. Propose to move to Advanced Installer as it has an intuitive UI/UX and can be driven via command line. Currently, professional edition is required to create an installer with equivalent capability.

During this transition, it is recommended to split the installer and have a separate installer per OS Architecture (1x MSI for 32-bit and 1x MSI for 64-bit).

gerhardol:
A paid license will be a problem
Open source licenses are offered:
https://www.advancedinstaller.com/free-license.html

@RussKie Should we approach Advanced Installer and ask for terms before we go in this direction?
If the installer may run in AppVeyor and run for PRs, it may be OK.

Johno-ACSLive:
It wouldn't hurt to have a chat to them, my experience has been pretty positive with them.

I'm happy to continue with the work in parallel with your discussions. If it turns out the project cannot go with Advanced Installer then that's fine.

Advanced Installer has added quite a few enhancements to command line building over the last couple of versions which is in my list to test. AppVeyor should be able to build the installer just fine.

RussKie:
I dropped them a line.

My biggest question is whether a licence would be tied to a build server, or whether any contributor would be able to build an installer.

RussKie:
FYI, I got a licence, just clarifying the terms of use and how it can be incorporated into our CI/CD.
Will keep you posted.

pmiossec:
>During this transition, it is recommended to split the installer and have a separate installer per OS Architecture (1x MSI for 32-bit and 1x MSI for 64-bit).

I don't know your point of view on that but my feeling is that we could drop the 32-bit support (especially if add some more work).

RussKie:
That is a fresh idea! :D
:+1: from me

Johno-ACSLive:
Do we have telemetry on % of 32-bit Windows installs vs 64-bit? I would think that most, if not all users are on 64-bit Windows installs but I think would be nice to have confirmation.

Agree it would be easier to maintain a 64-bit installer only.

It does also remove the problem of trying to detect OS architecture to pick the correct MSI etc. but creates another in that 32-bit Windows installs will fail MSI installation via the updater. The user will still have the previous version installed but if they try update it will just fail to update.

[...]

PR #7656, opened 2020-01-19
WIP: Overhaul Installer - migrate to Advanced Installer and Create 64-bit MSI

Johno-ACSLive:
## Proposed changes

This PR is based on #7368 and implements and Advanced Installer project to build a 64-bit only MSI. Provides all the capability of the current MSI while fixing some issues (e.g. per user installation, preserving all configuration etc.).

Some ShellExtensions detection Logic has been fixed to properly support scenario where 32-bit binaries have been removed and GE has been re-configured to not prefer 32-bit.

New scripts to create file and structure for MSI project and to build MSI via commandline.

## Test methodology 

- Run current installer and then run upgrade. Ensure all configuration is preserved and GE runs
- New install and ensure all parameters are configured correctly by the MSI
- Generate another upgrade to ensure future Advanced Installer MSI's will upgrade correctly

## Further discussions topics for this PR

- Do we still want per user installs since we have portable builds

- If we want per devices installs, we need to re-think per user settings configured using the MSI. For enterprise environments that may use shared devices or where the installer is launched under a service account, any per user changes will not be made. Usually per user settings are configured by the application but the installer may set per device defaults or install templates to be used by the application.

- I haven't looked into how the new themes are meant to work but the Theme directory is included in Program Files however launching GE will create a copy in users AppData plus create a current file. GE does complain the per user files are missing initially and then generates the files. It should really only generate the current file. As per previous point, the dark and light theme can be used as a template and the current file just needs to include the current or default theme plus any colour changes different to selected theme. This means any changes to defaults will automatically come through from any upgrades as GE will read the template then overwrite any changes from current file. This should be a similar approach for other based user configs that may have a master/template config.

- Plugin manager - should this really be installer per user profile? This will break on devices as per point 2. Plugin manager should probably be installed in Program Files. If GE supports new versions of Plugin manager outside of new GE releases then the deltas can be installed per user (this will double up on binaries). If the Plugin manager version is tied to the GE release then I think we should definitely move the Plugin manager there. Plugins that are installed/managed by the Plugin manager will then of course go to the user profile (which will again mean double up on files).

- I'm working on a tool that will allow common configuration and generation of the output files required for both Portable and MSI, this would mean we only have 2 locations to modify rather than 3 when adding/modifying/removing files. Although the tool could be modified to integrate with Advanced Installer and therefore only require modification in 1 location it's probably best to keep it simple. Open to suggestions on this or if everyone is just happy to modify 3 locations.

----


gerhardol:
> Do we still want per user installs since we have portable builds

I vote to remove user install

> I haven't looked into how the new themes are meant to work

Maybe changed in master already, can you rebase (also for conflicts)
https://github.com/gitextensions/gitextensions/pull/7609#discussion_r365599680

> Open to suggestions on this or if everyone is just happy to modify 3 locations.

No one is happy about that, it was just too complicated with current installer

GitExtensions.sln
```diff
@@ -211,198 +206,297 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DeleteUnusedBranchesTests",
...
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		MainBuild|Any CPU = MainBuild|Any CPU
 		Release|Any CPU = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{F15A69AF-7EBD-4F69-A026-5071FA3EC61B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{F15A69AF-7EBD-4F69-A026-5071FA3EC61B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{F15A69AF-7EBD-4F69-A026-5071FA3EC61B}.MainBuild|Any CPU.ActiveCfg = Release|Any CPU
```
gerhardol: Why add MainBuild configuration?

Johno-ACSLive:
> I vote to remove user install

Great, I'm happy to do this!

> Maybe changed in master already, can you rebase (also for conflicts)
> [#7609 (comment)](https://github.com/gitextensions/gitextensions/pull/7609#discussion_r365599680)

Yes I believe my master branch was updated to the latest and the branch I created for this PR has been rebased onto the latest master. The discussion item correlates with other items I have discussed.

> No one is happy about that, it was just too complicated with current installer

OK initially we will modify 2 locations and in a future update we'll try get it down to 1.

GitExtensions.sln
Johno-ACSLive: I believe adding the Advanced Installer project did this. Given the initial idea was to use the GE project's output for the installers input files but cannot be used, a Visual Studio project for Advanced Installer is no longer necessary.

This is because the file structure is vastly different plus external dependencies are not mapped (and in different locations).

Building the installer will not be performed by the VS project nor Advanced Installer GUI but via command line. So in this case I'll remove this which should revert the Solution back.

Johno-ACSLive:
Four items that need reviewing/refactoring are set-telemetry (this is no longer included as a file but is embedded in the installer project), the SSH options, Theme default and Plugin manager location. By default the installer sets Putty on new install, however, if the installer does not set any default, GE will by default use OpenSSH.

Perhaps we create a template with all default configs (e.g. default state of telemetry, default SSH client and default theme). The installer can modify this template based on initial install and for portable it will be the default. All users will then use this as a default and any overrides will be written to relevant config files in AppData (or local folder for portable).

GE on launch will read template and override config based on user changes to allow for simpler code, easier upgrades and consistency.

Proposed defaults:
- SSH Client: Putty
- Telemetry: Enabled
- Theme: Light

Confirming everyone is happy to move the Plugin manager to the Program Files directory? Plugins downloaded via the manager will go to AppData for installer and local folder for portable. Let me know which direction everyone would like to go.

mast-eu:
> Confirming everyone is happy to move the Plugin manager to the Program Files directory?

Wait a sec. If I remember correctly, we had issues with the plugin manager in program files. PM needs to be able to write into its own root folder for self updates and to a configuration file. The later might possible be moved, but the self update would still fail without admin rights.
Let's agree on these points before taking any action.

@maraf: do I remember correctly?

Johno-ACSLive:
> Let's agree on these points before taking any action.

Of course, that's the point of this conversation. Please see point no 4 under heading further  discussion topics for this PR.

We need to change things, let's discuss and agree which way everyone would like to go.

RussKie:
We also need to accommodate this functionality:
https://github.com/gitextensions/gitextensions/blob/dca061a6c65c89343697f94b60b086edc7e44dcc/Build/transifex/DownloadTransifexTranslations.ps1#L53-L179
Whenever we download translations we update the installer scripts.

> * SSH Client: Putty

Our telemetry shows OpenSSH is used by 70% of users (that enabled the telemetry). So I'm against this.

> PM needs to be able to write into its own root folder for self updates and to a configuration file.

Correct, so we can't put in the protected folder.

gerhardol:
>> Confirming everyone is happy to move the Plugin manager to the Program Files directory? Plugins downloaded via the manager will go to AppData for installer and local folder for portable. Let me know which direction everyone would like to go.

> Wait a sec. If I remember correctly, we had issues with the plugin manager in program files.

The issue was primarily for saving the plugins themselves to Program Files, but this is a similar problem.
The plugin manager in Program Files can only be updated with the installer, but can updates be stored/run from AppData/local?

Johno-ACSLive:
> Whenever we download translations we update the installer scripts.

The new script I wrote (which will be replaced by the tool I'm working on) currently pulls all files into one location. So long as the file names are the same the installer will include them. If they are new files or need to be removed then of course that needs to occur within the installer project - at the moment I am looking to include that in the new tool but not on first release.

Do we know how often translation files are added or removed or are they generally the same languages with file updates (and with the same file names)?

Johno-ACSLive:
> Our telemetry shows OpenSSH is used by 70% of users (that enabled the telemetry). So I'm against this.

Great, I was pointing out before that GE by default sets OpenSSH but the old installer defaults to Putty and that both should be aligned. Happy to align it to OpenSSH!

Johno-ACSLive:
> Correct, so we can't put in the protected folder.

I disagree, updates can be downloaded to AppData and the config can go there as well.

>The plugin manager in Program Files can only be updated with the installer, but can updates be stored/run from AppData/local?

Exactly!

My question is - does the plugin manager need to be self updating or do we want to tie each plugin manager release with a GE release? Based on the replies so far it seems to be the former.

In either case, we can provide a base copy (current version) of the plugin manager per GE release and plugin manager can download updates to itself into AppData, launch itself from there and also store plugins/config there.

This helps eliminate the issue I was talking about where an install will not copy plugin manager files for all users of a device (on a multi-user system). If the GE installer runs under SYSTEM context e.g. delivered by SCCM or some desktop tool then it will not install/update the plugin manager for the primary user. There are ways to overcome this in the MSI but I don't think that is the way to go as it's much more effort than what is required here.

We simply need to store files in their correct location and change the behaviour of the tools. Any files user related should be managed by GE or any sub component run by GE (e.g. Plugin manager) and not by the installer.

gerhardol:
> Do we know how often translation files are added or removed or are they generally the same languages with file updates (and with the same file names)?

Some translations toggle being in and out of GE, the limit is 95%

> My question is - does the plugin manager need to be self updating or do we want to tie each plugin manager release with a GE release? Based on the replies so far it seems to be the former.

Yes, PM should be update-able

mast-eu:
> In either case, we can provide a base copy (current version) of the plugin manager per GE release and plugin manager can download updates to itself into AppData

I don't like the idea to keep multiple versions of PluginManager in different locations. This seems error prone to me.

> This helps eliminate the issue I was talking about where an install will not copy plugin manager files for all users of a device (on a multi-user system).

I admit I didn't consider this scenario when we set up the PluginManager like it is right now.

In case we go for system-wide installs (i.e. not per single user), I propose another way:

GE goes to `C:\Program Files` as usual.
PM goes to `C:\users\AllUsers` and has also its config and user plugins in the same location.

This way:
* All users share the same binary of PM
* PM can be updated independently
* Any update of GE (which might bring breaking changes to the plugin API) can update the PM as well, hence not leaving incompatible versions scattered in other users' AppData
* All users would share the installed plugins

vbjay:
Instead keep plugin manager as you describe. Store settings for it per
user. Let user 1 have different plugins than user 2. Store plugins in per
user location.

Johno-ACSLive:
> I don't like the idea to keep multiple versions of PluginManager in different locations. This seems error prone to me.

Agree, this was something I mentioned earlier.


> PM goes to `C:\users\AllUsers` and has also its config and user plugins in the same location.

This location is redirected to `C:\ProgramData` which uses %AllUsersProfile% environment variable.

If a user creates something in this location it is not accessible by other users so I don't think this is the best place either.

A more suitable place might be somewhere under %Public% (`C:\Users\Public`) but I'm not a fan of that location.

> * Any update of GE (which might bring breaking changes to the plugin API) can update the PM as well, hence not leaving incompatible versions scattered in other users' AppData

This could also go the other way (an update to PM could break an older version of GE).

How does Visual Studio manage plugins? The Extension manager I would think is the equivalent and is bundled per VS release. Any plugins are then installed/tracked on a per user basis from what I can see.

Having a look at NuGet, it stores packages under `C:\Users\<username>\.nuget` - while I don't like this location as it should probably go under %LOCALAPPDATA% - it uses a similar model when looking at VS2019.

In any case perhaps leveraging the VS2019 Extension manager model is the best way to go. If what I'm seeing is correct that would mean PluginManager goes to GE install directory (`C:\Program Files\Git Extensions`) and plugins/config go to the user profile. This would couple a version of PluginManager per GE release.

At the moment the PluginManager is a Plugin for GE, I would think that PluginManager should be a core component of GE just like other components such as Translations or Dictionaries etc. PM can then manage Plugins if enabled via Install (or automatically included for Portable).

This configuration would still allow plugins to be installed/updated/uninstalled independently. The only thing lost is the independent updates of PM, is this a huge loss?

Which way does everyone think we should go?

vbjay:
No plug-in manager updates itself. It is a plug-in itself. ProgramData is
correct. We can alter permissions to allow read /write for everyone. Alter
permissions of our folder in ProgramData.

mast-eu:
> This could also go the other way (an update to PM could break an older version of GE).

GE is already protected against this, since it simply doesn't load incompatible plugins. Thus, in the worst case (old GE with new and incompatible PM), the PM wouldn't be usable, but the main app would not be affected.
The user would have to either update GE to the latest version or to repair the older GE installation, hence restoring a compatible PM version.

In the future we might also extend PM to avoid that an incompatible plugin is installed in the first place: https://github.com/gitextensions/gitextensions.pluginmanager/issues/12

> We can alter permissions to allow read /write for everyone. Alter permissions of our folder in ProgramData.

I agree with @vbjay.

Johno-ACSLive:
> ProgramData is correct. We can alter permissions to allow read /write for everyone. Alter permissions of our folder in ProgramData.

Modifying security permissions under Program Files or ProgramData is bad practice, is it the correct location - that depends on what you store there and how you intend something to work from that location. There are articles out there that say "Change the ACL" but most are referring to application compatibility - especially where enterprises package/re-package apps to conform with that organisations packaging and SOE standards.

What if an organisation is using a multi-user system (RDS farm or similar). The MSI process will handle this for a GE update/upgrade but PM would need to handle that possible condition of updating while another user has an instance of PM running. Not sure if we have stats on this - although that doesn't mean someone in future won't set this kind of environment up and have issues.

The likelihood of that kind of collision would be minimal but it is still a possibility. 

> The user would have to either update GE to the latest version or to repair the older GE installation, hence restoring a compatible PM version.

Yes updating to the latest version would be the only option. I don't think we should allow the installer repair process to downgrade the PM. If a newer version of PM changes config files etc. then downgrading the PM may mean the PM will no longer function correctly.

My take is this - anything that requires to be a template or base such as config (SSH, Telemetry and Theme defaults based on installer inputs) -> store in ProgramData including base version of PM.

> I don't like the idea to keep multiple versions of PluginManager in different locations. This seems error prone to me.

Given we are storing Plugins per user already (and this would cause duplication) and the PM is also a Plugin itself. I don't think duplication is a reason not to store/update PM from the profile. The logic from GE's point of view would be simple so I don't think it would make the solution more complex or prone to errors (based on the options I propose below).

Based on the conversation so far, here are the options with regard to PM (GE config changes seem to be OK with everyone so far):

- GE copies base version from ProgramData to LOCALAPPDATA if it doesn't exist and then always launches PM from that location if it exists. PM will self update per user. User can self downgrade by deleting LOCALAPPDATA files.

- GE launches PM from ProgramData and PM decides to copy itself to LOCALAPPDATA for launching if it doesn't exist, if an update is required then it can update itself in LOCALAPPDATA and then launch PM from there or simply launch PM from LOCALAPPDATA if all conditions are OK.

- A PM bootstrap is developed and the PM bootstrap program goes to ProgramData and is launched by GE. Bootstrap decides what to do, a base version can be included like the previous point.

- Implement what @vbjay suggested, put the PM in ProgramData and modify the ACL to allow self-updates with config/plugins being stored in user profile.

Options above are more or less listed in order of preference from my perspective.

I'll implement what the majority agree to 😃 apologies for the large comment!

vbjay:
So no. You store the plugins in programdata too. We can then store if user
is using plug-in and its settings per user.

Johno-ACSLive:
> You store the plugins in programdata too

This isn't the kind of purpose the ProgramData location is meant for and different to what you suggested in a previous comment.

vbjay:
No it isn't.

Look. We create a folder git extensions in programdata folder. We create a
plugins folder that is read write by everyone.

Plug-in manager and plugins reside here. Plug-in serttings are per user and
per repo. Look at how git config settings work. If user 1 upgrades plug-in
then everyone is updated. The settings are updated if marked with old
version.

This way we don't throw different versions of plugins all over the place.
They are stored in one spot. If a user doesn't want a plug-in they can
disable it for their user account. When user uninstalls we cleanup that
folder.

Johno-ACSLive:
> Let user 1 have different plugins than user 2. Store plugins in per user location.

This is what you suggested earlier which is now different - in any case it doesn't matter. More on that 👇 

> - Implement what @vbjay suggested, put the PM in ProgramData and modify the ACL to allow self-updates with config/plugins being stored in user profile.

Implementing this option would enable PM to implement either of your proposed solutions as the PM would have full access to its own directory by any user.

Again this is not considered best practice so I don't agree this is the approach that should be taken.

> Look at how git config settings work. If user 1 upgrades plug-in
then everyone is updated.

I've only seen things related to system config which does go to ProgramData for Git. This to me is similar to setting the default telemetry value during initial MSI install but each user can override the default by changing this value on a multi-user device. I've not seen plugins for git so can't comment on this. However, if we look at NuGet and Visual Studio 2019 extensions they behave a little differently.

Moving forward though, if PM will download plugin binaries to be shared by multiple users it would then need to manage the following:

- Track how many users are using a plugin before removing binaries
- Handle multiple users running an instance of PM and take care of any collisions that may occur

On the MSI side, we would need to ensure that we only replace PM binaries if the MSI contained a newer version. We would also need to ensure that the PM directory in ProgramData is not removed during upgrade, only during uninstall.

If a user decided to install GE again in future, the PM will need to restore the plugin binaries since the user config would indicate it is enabled but the binaries will be missing.

@RussKie @gerhardol and @mast-eu what is your opinion on approach from the 4 options I listed?

vbjay:
Why do we need the plugin manager in msi?  Why not have GE download it if it doesn't exist in correct location.  Obviously if they are downloading plugins, they have internet. Once it exists in location we let the plugin manager deal with keeping itself up to date when it runs.

RussKie:
Let's step back and revisit the requirements.

1. A user must be able to install a new version of the app
2. If the user installs for the first time:
    * the user should get Plugin Manager installed
    * the user should be asked whether to use OpenSSH or PuTTy
    * the user should opt-in to telemetry
    * (new) we should ask the user to pick the language as well rather than showing another dialog at the start
    * all settings should created
3. If the user updates the app - the installer should only update the app. I.e. the user shouldn't be asked the following:
    *  to choose SSH client
    * to choose language
    * to opt-in to telemetry
    * get non-bundled plugins updated.

We have discussed sharing plugins between users on the same box, and concluded it was not a good idea for a number of reasons.

I also don't think it is a good idea to deploy anything besides the app into system/restricted folders. All user content, such as plugins and settings, should reside in user folders (i.e. C:\Users\[user]\AppData\Roaming\GitExtensions).

RussKie:
@Johno-ACSLive please rebase on top of the latest master to leverage the latest build pipeline based on Arcade SDK.
The flow we'll be leveraging `.\build.cmd -pack` command to create an msi (we can rename `pack` to something else). There is a very rough plumbing, but we will need to write and wire up all the targets. I'm happy to help you with that, if needed.

As part of msi creation we will also need to dynamically reference available languages.

Johno-ACSLive:
Not a problem, will do!

RussKie:
I think we at the point where we can now look at the installer update.
The build pipeline has been for the most part overhauled and converted to MSBuild-based scripts instead of batch files. The former allows leveraging MSBuild plumbing and resolutions instead of manually listing all necessary files to include (although we will have to manually include plugins dependencies for the time being).

I suppose the first thing to try is to create an installer for the main app. I'm thinking it should be similar to how we create a portable build:
https://github.com/gitextensions/gitextensions/blob/7bf8ac984c376f1b5db949224ab26eceae14ddc8/GitExtensions/Project.Publish.targets#L55-L86

The build script has a `publish` switch (`PS> .\Build\build.ps1 -publish`) that should invoke a new target (e.g `CreateInstaller`) with the installer scripts, before or after `CreatePortable` target is invoked.

GintasS:

PR #7666, opened 2020-01-21
Add A/B to RevisionDiff revision descriptions #7626

gerhardol:
Discussion in #4154
Replacing #7626, already reviewed and approved, just changing A->a, B->b
This gives yet another chance to suggest changes

## Proposed changes
Add _(a)_ and _(b)_ to some revision descriptions in RevisionDiff

### Before

image

image

### After

image

image

## Test methodology 
manual

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -28,8 +28,8 @@ namespace GitUI.CommandsDialogs
             new TranslationString("Are you sure you want to delete the selected file(s)?");
         private readonly TranslationString _deleteFailed = new TranslationString("Delete file failed");
         private readonly TranslationString _multipleDescription = new TranslationString("<multiple>");
-        private readonly TranslationString _selectedRevision = new TranslationString("Selected");
-        private readonly TranslationString _firstRevision = new TranslationString("First");
+        private readonly TranslationString _selectedRevision = new TranslationString("Selected: b/");
+        private readonly TranslationString _firstRevision = new TranslationString("First: a/");
 
         private RevisionGridControl _revisionGrid;
         private RevisionFileTreeControl _revisionFileTree;
@@ -210,14 +210,14 @@ namespace GitUI.CommandsDialogs
         /// Provide a description for the first selected or parent to the "primary" selected last
         /// </summary>
         /// <returns>A description of the selected parent</returns>
-        private string DescribeSelectedParentRevision(IEnumerable<GitRevision> parents)
+        private string DescribeRevision(List<GitRevision> parents)
         {
-            if (parents.Count() == 1)
+            if (parents.Count == 1)
             {
                 return DescribeRevision(parents.FirstOrDefault()?.ObjectId, 50);
             }
 
-            if (parents.Count() > 1)
+            if (parents.Count > 1)
             {
                 return _multipleDescription.Text;
             }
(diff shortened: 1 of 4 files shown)


GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -210,18 +210,14 @@ private string DescribeRevision([CanBeNull] ObjectId objectId, int maxLength = 0
         /// Provide a description for the first selected or parent to the "primary" selected last
         /// </summary>
         /// <returns>A description of the selected parent</returns>
-        private string DescribeSelectedParentRevision()
+        private string DescribeRevision(IEnumerable<GitRevision> parents)
```
RussKie:
```suggestion
        private string DescribeRevision(List<GitRevision> parents)
```

and replace `parents.Count()` with `parents.Count`

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -749,18 +745,19 @@ private void resetFileToToolStripMenuItem_DropDownOpening(object sender, EventAr
             {
                 resetFileToSelectedToolStripMenuItem.Visible = true;
                 resetFileToSelectedToolStripMenuItem.Text =
-                    _selectedRevision + " (" + _revisionGrid.DescribeRevision(DiffFiles.Revision, 50) + ")";
+                    _selectedRevision + " (" + DescribeRevision(DiffFiles.Revision?.ObjectId, 50) + ")";
             }
 
-            var parents = DiffFiles.SelectedItemParents.Distinct();
+            var parents = DiffFiles.SelectedItemParents.ToList();
             if (parents.Count() != 1 || !CanResetToRevision(parents.FirstOrDefault()))
```
RussKie:
```suggestion
            if (parents.Count != 1 || !CanResetToRevision(parents.FirstOrDefault()))
```

gerhardol:
Applied RussKies suggestions
I am not happy at all about lower case (a)...

image

RussKie:
> I am not happy at all about lower case (a)...
> 
> image

Hmm that doesn't look right to me either.
I suppose I wasn't clear in communicating my thoughts. I was thinking of something like this:

```
-- Diff with a/6a34728274: feature/bla
-- Diff with Commit Index
-- Diff with Working directory
```

mstv:
I'd like to put everything in (although it might be not completely visible):
```
-- Diff a/6a347282: feature/bla with b/12345678: master commit
-- Diff a/Commit Index with b/abcdef01: bla
-- Diff a/Working directory with b/87654321: blub
```

grafik

I'd omit the parens in the context menu items:
`First: b/59d118a9: WaitForPendingOperations in FormCommitTes...`

In addition, the shortened commit hash could be added to the DiffView.
before:
grafik

after:
```
diff --git a/UnitTests/CommonTestUtils/AsyncTestHelper.cs b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
index 328374f10..278d18c10 100644
--- 851b482c: a/UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ 59d118a9: b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
```

gerhardol:
Updated to what I hoped is the final version...

I did not get the admittance I requested here:
https://github.com/gitextensions/gitextensions/pull/7626#issuecomment-574353852
You were saved by my merge mistake!

> I suppose I wasn't clear in communicating my thoughts. 

https://github.com/gitextensions/gitextensions/pull/7626#issuecomment-576122671
Changed to @mstv latest suggestion above, no refactoring neither too

> I'd like to put everything in (although it might be not completely visible):

Some thoughts for later, this is still just tuning current presentation.

> -- Diff a/6a347282: feature/bla with b/12345678: master commit

For now, selected/b revision is fixed and not included. (except for maybe Combined diff)

If #4154 I prefer to have something that makes it easier to distinguish the odd commit with a separate selected/b.

```Diff b/6a347282: feature/bla from a/12345678: master commit```

> In addition, the shortened commit hash could be added to the DiffView

Later if changed, a/b and revisions are explained in the header already.

```diff --git a/UnitTests/CommonTestUtils/AsyncTestHelper.cs b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
index 328374f10..278d18c10 100644
```
Color code the header?

gerhardol:
@msftbot merge in 1 day

mstv:
> index 328374f10..278d18c10 100644

This information is something git-internal -- I'd say _private_. It looks like commit hashs but that's not the case.
So we should rather remove it.

pmiossec:
It should be the blob hash. We could verify it easily doing `git cat-file -p`.

But I didn't know we could remove it in the diff.
we really could?

vbjay:
Yes. Its object hashes.   I think it's --no-index

vbjay:
Nope. I was wrong. --no-index does something else.

gerhardol:
> So we should rather remove it.

> But I didn't know we could remove it in the diff.

This discussion is OT for merging now, still relevant here

We create the patch, parsing Git diff output, sure we could remove or reformat the output 

Proposal

```--- a/851b482c: UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ b/59d118a9: UnitTests/CommonTestUtils/AsyncTestHelper.cs```

PR #7667, opened 2020-01-21
ObjectId ShortString: set default to 8 chars

gerhardol:
Discussed here: https://github.com/gitextensions/gitextensions/pull/7626#issuecomment-574760763

## Proposed changes
The short version of commit hash was previously 10 characters but in 9 of 19 uses the size was modified (to 8 chars, except in revision diff where it was 4, changing to the default in #7666). 

The short format that is unique in a repo can be retrieved from Git. Many repos are OK with 7 bytes (this was the initial for the Linux repo), bigger repos need longer sha, Linux at least 12 bytes to _only_ use the short sha.

This short format is only used for display, in many places the link can be copied.
The short form should be used to present the sha in the GUI to correlate info in for instance CommitInfo to the revision grid. The short form should not be required to be unique.
However, the length should be consistent in the GUI.

In the UI 10 bytes is too long to be easily remembered. The first 2 bytes are normally enough to compare sha to the revision grid or similar, but also the 4 byte sha in revision diff has confused users (see #7626), so something longer is needed. I find 6 to be more than enough, however changing to 8 that is used in many situations already is a compromise.

## Screenshots 

One example only

### Before

image

### After

image

----

### DIFF
--- a/Plugins/GitUIPluginInterfaces/ObjectId.cs
+++ b/Plugins/GitUIPluginInterfaces/ObjectId.cs
@@ -480,7 +480,7 @@ namespace GitUIPluginInterfaces
         /// <exception cref="ArgumentOutOfRangeException"><paramref name="length"/> is less than zero, or more than 40.</exception>
         [Pure]
         [NotNull]
-        public unsafe string ToShortString(int length = 10)
+        public unsafe string ToShortString(int length = 8)
         {
             if (length < 0)
             {
--- a/GitCommands/Git/GitRevision.cs
+++ b/GitCommands/Git/GitRevision.cs
@@ -89,7 +89,7 @@ namespace GitCommands
 
         public string Name { get; set; }
 
-        public override string ToString() => $"{ObjectId.ToShortString(8)}:{Subject}";
+        public override string ToString() => $"{ObjectId.ToShortString()}:{Subject}";
 
         /// <summary>
         /// Indicates whether the commit is an artificial commit.
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1578,12 +1578,12 @@ namespace GitUI.CommandsDialogs
 
                 if (revisions[0].ObjectId == currentCheckout)
                 {
-                    from = revisions[1].ObjectId.ToShortString(8);
+                    from = revisions[1].ObjectId.ToShortString();
                     to = currentBranch;
                 }
                 else if (revisions[1].ObjectId == currentCheckout)
                 {
-                    from = revisions[0].ObjectId.ToShortString(8);
+                    from = revisions[0].ObjectId.ToShortString();
                     to = currentBranch;
                 }
 
--- a/GitUI/CommandsDialogs/FormRebase.cs
+++ b/GitUI/CommandsDialogs/FormRebase.cs
@@ -308,7 +308,7 @@ namespace GitUI.CommandsDialogs
             {
                 if (chooseForm.ShowDialog(this) == DialogResult.OK && chooseForm.SelectedRevision != null)
                 {
-                    txtFrom.Text = chooseForm.SelectedRevision.ObjectId.ToShortString(8);
+                    txtFrom.Text = chooseForm.SelectedRevision.ObjectId.ToShortString();
                 }
             }
         }
--- a/GitUI/CommandsDialogs/FormResolveConflicts.cs
+++ b/GitUI/CommandsDialogs/FormResolveConflicts.cs
@@ -680,7 +680,7 @@ namespace GitUI.CommandsDialogs
                 return "@" + _deleted.Text;
             }
 
-            return '@' + item.ObjectId.ToShortString(8);
+            return '@' + item.ObjectId.ToShortString();
         }
 
         private void ConflictedFiles_SelectionChanged(object sender, EventArgs e)
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -200,7 +200,7 @@ namespace GitUI.CommandsDialogs
 
             if (revision == null)
             {
-                return objectId.ToShortString(length: 8);
+                return objectId.ToShortString();
             }
 
             return _revisionGrid.DescribeRevision(revision, maxLength);
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -1037,7 +1037,7 @@ namespace GitUI
                     return DescribeRevision(objectId);
                 }
 
-                return objectId.ToShortString(length: 8);
+                return objectId.ToShortString();
             }
         }
 
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -644,7 +644,7 @@ namespace GitUI
 
             if (!revision.IsArtificial)
             {
-                description += " @" + revision.ObjectId.ToShortString(4);
+                description += " @" + revision.ObjectId.ToShortString();
             }
 
             if (maxLength > 0)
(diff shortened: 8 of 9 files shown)


Plugins/GitUIPluginInterfaces/ObjectId.cs
```diff
@@ -480,7 +480,7 @@ public override string ToString()
         /// <exception cref="ArgumentOutOfRangeException"><paramref name="length"/> is less than zero, or more than 40.</exception>
         [Pure]
         [NotNull]
-        public unsafe string ToShortString(int length = 10)
+        public unsafe string ToShortString(int length = 8)
```
drewnoakes: Are there any usages left that specify a length? If we want to make this consistent across the board, perhaps remove the parameter altogether.

gerhardol: There are no (ObjectId external) usages of the parameter now.
I do not want to remove it as it is part of the plugin interface.
Also, I expect that a unique length may be needed in some situations like reports. That may use Git though for the abbr length, but this opens for the possibility to add safer lengths (longer than 10 chars).

mstv: > Are there any usages left that specify a length?

Yes, `CommitIdColumnProvider`.

mstv:
My approval was too rash. A test must be adapted:
`ResourceManagerTests.CommitDataRenders.CommitDataHeaderRendererTests.Render_should_render_minimal_info_for_artificial_commits`
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30293789

gerhardol:
> A test must be adapted

Looking forward to when the tests are reliable, now I often just try to rebuild. Thought I had checked this.
Something is weird in the test explorer too (in addition to the tests that fail locally.)
image

Fixed the test and rebased.
Will wait for comments a day or two more before merging.

RussKie:
As an idea, should we query a repo upon open what minimum hash size it requires, the same way we do for encoding?
E.g. add here `git rev-parse --short HEAD` (or something like that)?
https://github.com/gitextensions/gitextensions/blob/56b6a09ea7d245809a119f51554a091c4dd324f1/GitCommands/Git/GitModule.cs#L44-L57

Though I can't think how we can pass the size to the `ObjectId.ToShortString`... 🤔

pmiossec:
>As an idea, should we query a repo upon open what minimum hash size it requires, the same way we do for encoding?

I already expressed my point of view in the past, that git is better placed than us to determine the correct short hash length and so that it is, IMHO, the right way to do. So, 👍

gerhardol:
It is possible to query Git for the abbreviated sha at a certain point for the short sha
That will require async in a lot of places where not used now.
Also, the length will be varying, inconsistent UI - which was one reason for the PR.

It could be possible to find the longest sha:
https://blog.cuviper.com/2013/11/10/how-short-can-git-abbreviate/
That requires tens of seconds also for medium sized repos.

So I strongly oppose any dynamic approach.

The short sha is just for display, the full length should be available, so there is no strong need for a unique sha.

pmiossec:
@gerhardol I think there is a misunderstanding....

I just want to precise that there is no strong added value to let git find the right length so this is mostly for discussion.

>The short sha is just for display, the full length should be available, so there is no strong need for a unique sha.

Indeed, we are talking to use the full length everywhere except where it's more convenient (for display space or human readind) to use short sha.

>Also, the length will be varying, inconsistent UI - which was one reason for the PR.

&

>That requires tens of seconds also for medium sized repos.

I think that you are missing something.

Since Git 2.11, git introduced a default length for a repository.

See https://stackoverflow.com/a/21015031/717372 :

> Introduce a mechanism, where we estimate the number of objects in the repository upon the first request to abbreviate an object name with the default setting and come up with a sane default for the repository.

Or from doc:

>If unspecified or set to "auto", an appropriate value is computed based on the approximate number of packed objects in your repository, which hopefully is enough for abbreviated object names to stay unique for some time. 

So all the short sha of a same repository will have the same length and we don't have to detemine the length ouselve by a process that takes more than 10 seconds.

>That will require async in a lot of places where not used now.

Knowing all that, I don't see why.
Like @RussKie said, we will just have to ask git for the length at the repo opening and use this length everywhere we display a short sha.

gerhardol:
> Since Git 2.11 (so after this blog post was written), git introduced a default length for a repository.

```git rev-parse --short <commitish>``` 
still calculates the length for one sha, maybe not the same for all. I do not see any way (even if mentioned in the commit comment) that this is for the complete repo.
Also, for the short length to be really safe, it should be calculated at every commit.
So I resist a completely dynamic solution.

A "semi-dynamic" way could be to use the configured length. 
```git config --get core.abbrev```
(if done this way, the setting should be promoted in settings; _Git->Advanced_)
This way I can get a sha with four bytes too...

**Comment by RussKie** [comment]
A chance that a user opens a repo, makes changes, and while doing so a calculated length of hash becomes shorter than necessary, is close to 0%. And in that case the user will need to close and open the repo again... 

Anyway the conversation is diverging from the intent in this PR. I propose we start a new topic for discussing how we can achieve the desired.

**Comment by pmiossec** [comment]
>Anyway the conversation is diverging from the intent in this PR.

👌 

@gerhardol I'm sorry to tell you but I think that what you say is not how it works 😢 

>`git rev-parse --short <commitish>` still calculates the length for one sha, maybe not the same for all.

No. Since git 2.11, using an heuristic based on the number of objects in its database, git use the same length to display all the short hash everywhere. (For example, for the `GitExtensions` repository, this length is currently of 9 and for `gitextensions.plugintemplate` it is 7, and for `vscode` it is 10, ...).
So you will have **always** this length when asking for a short hash (except in some very exceptional cases that we will surely never encounter where the length is not enough because git is playing with probabilities with its heuristic)
. 
That's even the length used when you display a diff:

image

The only case where you have different lengths for short hash is when you set `core.abbrev` with a shorter value. In this case, git will not always respect the length asked when there is a risk of collision.

Like in this example for `git config --local core.abbrev 4` :

image

You could see that the first hash respect the value but that second don't respect it, surely because that's not enough and there is a 2 objects in the git database starting with this 4 characters.

So, I really think that git handle it so much better than us and letting the user set `core.abbrev` is not a good solution if we want to display short hash with the same length for beautiful UI purpose...

**Comment by gerhardol** [comment]
> Anyway the conversation is diverging from the intent in this PR

Definitely, I felt the discussion was blocking this change though
The discussion is a basis for another PR

**Comment by gerhardol** [comment]
The short length for a sha in the GUI should be consistent both from uniform presentation and for performance (to calculate the short sha when opening menus will add delays).

The sha length could be shorter in some situations (revisiondiff is a good example I believe, but I was voted down).
If a shorter sha is used in reports or similar on the other hand, some extra length should be added.

> I'm sorry to tell you but I think that what you say is not how it works

Correct, the documentation is not clear:
https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---shortlength

``` --short[=length]

    Same as --verify but shortens the object name to a unique prefix with at least length characters. The minimum length is 4, the default is the effective value of the core.abbrev configuration variable (see git-config[1]).
```

The short form is not the shortest possible, but unique. In GE repo 9 chars even if almost all can be represented with 6 chars. But if you set core.abbrev, the length 

```$ git rev-list --all --abbrev=0 --abbrev-commit |
>   awk '{ a[length] += 1 } END { for (len in a) print len, a[len] }'
4 1906
5 18686
6 3282
7 213
8 12
$ git rev-list --all --abbrev=0 --abbrev-commit |   perl -ne '{if(length($_) < 6){print $_;exit}}'
59d0
$ git rev-parse --short 59d0
59d0e30be
$ git config --local core.abbrev 4
$ git rev-parse --short 59d0      
59d0
$ git rev-list --all --abbrev=0 --abbrev-commit |   perl -ne '{if(length >8){print $_;exit}}'
153c8303
$ git rev-parse --short 153c8303
153c8303                              
```

But if core.abbrev is set, I see no simple way to get a consistent repo value:

```
$ git -c core.abbrev= rev-parse --short 59d0
fatal: bad numeric config value '' for 'core.abbrev': invalid unit
$ git -c core.abbrev=40 rev-parse --short 59d0
59d0e30bee0eaa344a4e8fa63b997cf036d845ba
$ git rev-parse --short 59d0e30bee0eaa344a4e8fa63b997cf036d845ba
59d0
$ git -c core.abbrev=10 rev-parse --short 59d0
59d0e30bee
```

So ```git rev-parse --short``` will give different length depending on the selected commit.

So I still prefer the following:
 * core.abbrev if set
 * rev-parse --short for any commitish (HEAD will work) if available
 * 8

Still for display the sha must not be unique, just long enough to be identified as a sha.

PR #7678, opened 2020-01-23
Add pwsh as a choice of shell in Console tab

asherber:
Addresses part of #7523 

## Proposed changes

- Add `pwsh` in the settings dialog as an additional option for the shell to run in the Console tab

## Screenshots 

image

### Before

Dialog offers choice of bash, cmd, and powershell

### After

Dialog offers pwsh as an additional choice

## Test methodology 
-  Manual testing

## Test environment(s) 

- GIT 2.25.0.windows.1
- Windows Win10 1903 (build 18362.592)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2854,44 +2854,7 @@ namespace GitUI.CommandsDialogs
                 }
 
                 startInfo.BaseConfiguration = startInfoBaseConfiguration;
-
-                string[] exeList;
-                switch (AppSettings.ConEmuTerminal.ValueOrDefault)
-                {
-                    case "cmd":
-                        exeList = new[] { "cmd.exe" };
-                        break;
-                    case "powershell":
-                        exeList = new[] { "powershell.exe" };
-                        break;
-                    default:
-                        // Choose the console: bash from git with fallback to cmd
-                        const string justBash = "bash.exe"; // Generic bash, should generally be in the git dir, less configured than the specific git-bash
-                        const string justSh = "sh.exe"; // Fallback to SH
-                        exeList = new[] { justBash, justSh };
-                        break;
-                }
-
-                string cmdPath = exeList.
-                      Select(shell => PathUtil.TryFindShellPath(shell, out var shellPath) ? shellPath : null).
-                      FirstOrDefault(shellPath => shellPath != null);
-
-                if (cmdPath == null)
-                {
-                    startInfo.ConsoleProcessCommandLine = ConEmuConstants.DefaultConsoleCommandLine;
-                }
-                else
-                {
-                    cmdPath = cmdPath.Quote();
-                    if (AppSettings.ConEmuTerminal.ValueOrDefault == "bash")
-                    {
-                        startInfo.ConsoleProcessCommandLine = cmdPath + " --login -i";
-                    }
-                    else
-                    {
-                        startInfo.ConsoleProcessCommandLine = cmdPath;
-                    }
-                }
+                startInfo.ConsoleProcessCommandLine = ShellHelper.GetCommandLineForCurrentShell();
 
                 if (AppSettings.ConEmuStyle.ValueOrDefault != "Default")
                 {
@@ -2925,51 +2888,7 @@ namespace GitUI.CommandsDialogs
 
         public void ChangeTerminalActiveFolder(string path)
         {
-            if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(path))
-            {
-                return;
-            }
-
-            switch (AppSettings.ConEmuTerminal.ValueOrDefault)
-            {
-                case "bash":
-                    if (PathUtil.TryConvertWindowsPathToPosix(path, out var posixPath))
-                    {
-                        ClearTerminalCommandLineWithMacroAndRunCommand($"cd {posixPath.QuoteNE()}");
-                    }
-
-                    break;
-                case "cmd":
-                    ClearTerminalCommandLineWithEscapeAndRunCommand($"cd /D {path.QuoteNE()}");
-                    break;
-                case "powershell":
-                    ClearTerminalCommandLineWithEscapeAndRunCommand($"cd {path.QuoteNE()}");
-                    break;
-                default:
-                    break;
-            }
-        }
-
-        private void ClearTerminalCommandLineWithMacroAndRunCommand(string command)
-        {
-            if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
-            {
-                return;
-            }
-
-            // Use a ConEmu macro to send the sequence for clearing the bash command line
-            _terminal.RunningSession.BeginGuiMacro("Keys").WithParam("^A").WithParam("^K").ExecuteSync();
-            _terminal.RunningSession.WriteInputTextAsync(command + Environment.NewLine);
-        }
-
-        private void ClearTerminalCommandLineWithEscapeAndRunCommand(string command)
-        {
-            if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
-            {
-                return;
-            }
-
-            _terminal.RunningSession.WriteInputTextAsync("\x1B" + command + Environment.NewLine);
+            _terminal?.ChangeFolder(path);
         }
 
         private void menuitemSparseWorkingCopy_Click(object sender, EventArgs e)
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
@@ -4,6 +4,8 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
 {
     public partial class FormBrowseRepoSettingsPage : SettingsPageWithHeader
     {
+        private int _cboTerminalPreviousIndex = -1;
+
         public FormBrowseRepoSettingsPage()
         {
             InitializeComponent();
@@ -31,6 +33,22 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         public static SettingsPageReference GetPageReference()
         {
             return new SettingsPageReferenceByType(typeof(FormBrowseRepoSettingsPage));
+        }
+
+        private void cboTerminal_SelectionChangeCommitted(object sender, System.EventArgs e)
+        {
+            if (ShellHelper.ShellIsOnPath((string)cboTerminal.SelectedItem))
+            {
+                return;
+            }
+
+            MessageBoxes.ShellNotFound(this);
+            cboTerminal.SelectedIndex = _cboTerminalPreviousIndex;
+        }
+
+        private void cboTerminal_Enter(object sender, System.EventArgs e)
+        {
+            _cboTerminalPreviousIndex = cboTerminal.SelectedIndex;
         }
     }
 }
--- a/GitUI/ShellHelper.cs
+++ b/GitUI/ShellHelper.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ConEmu.WinForms;
+using GitCommands;
+
+namespace GitUI
+{
+    public static class ShellHelper
+    {
+        internal static string GetCommandLineForCurrentShell()
+        {
+            string shell = AppSettings.ConEmuTerminal.ValueOrDefault.ToLower();
+            string result = GetShellPath(shell);
+
+            if (result == null)
+            {
+                return ConEmuConstants.DefaultConsoleCommandLine;
+            }
+
+            result = result.Quote();
+            if (shell == "bash")
+            {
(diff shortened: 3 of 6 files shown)


asherber:
I know that #7523 discussed a range of options, including a preference for specifying a custom shell. Taking my cue from some comments in #7612 , I took the narrow of option of just adding `pwsh` as an option. Among the difficulties in allowing for custom shells, as the other PR makes clear, is providing a mechanism to specify how the current console line should be cleared and the `cd` issued; for this reason, I think it may be wiser just to stick to known shells in use.

I do make the assumption that `pwsh` is somewhere on the user's path. @pmiossec  says it's not, for him, but I believe that's the exception. I will check again, but I believe that recent installers put it on the path by default, and it will also be on the path if it's installed as a dotnet global tool.

asherber:
I confirmed on a clean EC2 instance that the pwsh installer has an option to add the install dir to the path which is checked by default, and that installing pwsh as a dotnet global tool always puts it on the path.

pmiossec:
> I confirmed on a clean EC2 instance that the pwsh installer has an option to add the install dir to the path which is checked by default, and that installing pwsh as a dotnet global tool always puts it on the path.

All what have been made are good but I think that we must handle when a shell is not there either because it is not in the path or because it is not installed.

I'm pretty sure their will be users that will choose it without having it installed.

And I don't even know what is the current behavior.

I think it's the good moment to improve the handling and perhaps display a message box if the shell is not found.

asherber:
I was thinking about that. I can add code to check that the specified shell exists when the user makes a selection in the settings dialog.

pmiossec:
It would be great if you could achieve this. 👍

RussKie:
Just to confirm, if a configure shell isn't present and available at the runtime, we won't crash the app, will we?

asherber:
Current behavior, which I have left in place, is that if the selected shell can't be found at the time of Console tab activation, or if there is no shell selected in the settings dialog, ConEmu is invoked with its own `{cmd}` task. This appears to run `cmd` but makes it look and act a little more like a bash prompt. (See the two `*.cmd` files in the `ConEmu` folder inside the application folder.

What I did was add a guard clause in the settings dialog so that when the user chooses a shell, we check to make sure the shell can be found and display an error if it cannot. (User can't select a shell that we can't find.)

If for some reason a selected shell is removed between the time it is selected in the settings dialog and the time the Console tab is activated, the app will fall back to the current behavior, above. 

I could also add a guard clause here as well, to prevent Console tab activation if the selected shell can't be found. This would lead to two different behaviors where currently there is only one:

1. No shell selected in setting? Current: Launch `{cmd}` task. New: Launch `{cmd}` task.
1. Selected shell can't be found? Current: Launch `{cmd}` task. New: Show error message and prevent activation of Console tab.

I'll push what I currently have soon (no change to Console tab activation behavior). If you want me to add the second guard clause, I can do that too.

pmiossec:
@RussKie if you want to review it again since the shell existence handling addition...

GitUI/ShellHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ConEmu.WinForms;
+using GitCommands;
+
+namespace GitUI
+{
+    public class ShellHelper
```
RussKie: Why do we need to add this functionality to `GitUI` project and not to `ConEmuWinForms`? 
I do think this functionality is related to `ConEmuControl` then anything else. If we were to add or replace ConEmu with a different console, we would likely need to provide implementations specific to the new console.

With that callsites would be simpler, e.g. instead of
```cs
ShellHelper.ChangeFolderInTerminal(_terminal, path);
```
will be:
```cs
_terminal.ChangeWorkingDirectory(path);
```

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
```diff
@@ -32,5 +34,19 @@ public static SettingsPageReference GetPageReference()
         {
             return new SettingsPageReferenceByType(typeof(FormBrowseRepoSettingsPage));
         }
+
+        private void cboTerminal_SelectionChangeCommitted(object sender, System.EventArgs e)
+        {
+            if (!ShellHelper.ShellIsOnPath((string)cboTerminal.SelectedItem))
```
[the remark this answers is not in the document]
asherber: I'm not sure what you mean here.

GitUI/ShellHelper.cs
asherber: Isn't `ConEmuWinForms` an external submodule, part of the ConEmu app? 

For the methods that take a `ConEmuControl` param, I can certainly turn them into extension methods to simplify the call, if you like.

gerhardol: > Isn't ConEmuWinForms an external submodule, part of the ConEmu app?

ConEmuWinForms is an external lib, conemu-inside can be adopted though.
But extension is probably preferred.

asherber: Okay, done.

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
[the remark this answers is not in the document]
mstv: Turn
```
void foo()
{
    if (cond1)
    {
        // do 1
        if (cond2)
        {
             // do 2
        }
    }
}
```
into
```
void foo()
{
    if (!cond1)
    {
         return;
    }

    // do 1

    if (!cond2)
    {
        return;
    }

    // do 2
}
```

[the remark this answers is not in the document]
asherber: There's only one condition here -- if the shell isn't on the path, then reject the selection. IMO, this:
```csharp
void Foo()
{
  if (!okay)
  {
    reject;
  }
}
```
is clearer than:
```csharp
void Foo()
{
  if (okay)
  {
    return;
  }

  reject;
}
```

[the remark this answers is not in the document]
mstv: It is our coding style and recommended by Resharper. (I am not completely happy with it but it is consistent and reduces complexity a little.)
In my example you see the benefit, which also exists in the simple case when the "do"/`reject` code is to be extended.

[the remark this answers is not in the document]
asherber: Personally I disagree, but I'll make the change.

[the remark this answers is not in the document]
RussKie: I don't see the examples comparable to the code in question.
If the code was like in your examples, it would've been fine. However there was more than one instructions, i.e.
```cs
void Foo()
{
  if (!okay)
  {
    DoSomething();
    Reject;
  }
}
```
which is now:
```cs
void Foo()
{
  if (okay)
  {
    return;
  }

  DoSomething();
  Reject;
}
```

It is a styling preference, sure. However 'fast returns' make a big difference to readability and ability to reason about the code. It helps to reduce code rot as well - quite often one condition today turns into several nested conditions tomorrow. By reducing nesting it becomes easier to avoid nesting in the future.
And last but not least - fast returns also help in writing tests.

GitUI/ShellHelper.cs
RussKie: ConEmu is a submodule, correct. However we have long diverged from its upstream and no longer sync with it.
The new functionality is closely related to ConEmu, and it should reside in the submodule, and be exposed via an (ideally simple to use) API.
Should we decide to remove ConEmu, we should be able to remove the submodule.

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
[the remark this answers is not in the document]
asherber: I disagree about the readability of the two approaches for something like this, but it's your project and I'm happy to follow your preferred style.

GitUI/ShellHelper.cs
asherber: Ah, okay, got it about the status of ConEmu.

As for ShellHelper, half of the functionality in there is not directly related to ConEmu; it's more generic having to do with paths to shells. The methods that actually change the active folder are coupled to ConEmu functionality, but they're also conceptually related to the other shell methods in the class. And FormBrowse (and AppSettings) was already tightly coupled to ConEmu.

I can see the case for having some kind of facade class to use in FormBrowse instead of accessing ConEmu directly, but that's not how the existing code was written, and it's much bigger and more complicated than what this PR is trying to accomplish. My suggestion would be to deal with this PR on its own terms and add a task to the todo list to abstract away the specific console.

GitUI/ShellHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+        internal static string GetCommandLineForCurrentShell()
+        {
+            string shell = AppSettings.ConEmuTerminal.ValueOrDefault.ToLower();
+            string result = GetShellPath(shell);
+
+            if (result == null)
+            {
+                return ConEmuConstants.DefaultConsoleCommandLine;
+            }
+            else
```
mstv: Nit: this `else` can be removed -- reduces nesting.

GitUI/ShellHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+            string[] exeList;
+            switch (shell?.ToLower())
+            {
+                case "cmd":
+                case "powershell":
+                case "pwsh":
+                    exeList = new[] { $"{shell.ToLower()}.exe" };
+                    break;
+                case "bash":
+                    // Choose the console: bash from git with fallback to cmd
```
mstv: This comment is outdated and in the wrong place. I think it can be removed.

GitUI/ShellHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+                    const string justBash = "bash.exe"; // Generic bash, should generally be in the git dir, less configured than the specific git-bash
```
mstv: ", less configured than the specific git-bash" can be removed.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3015,46 +2978,7 @@ public void ChangeTerminalActiveFolder(string path)
                 return;
             }
```
mstv: These lines have been moved to `ShellHelper`. Remove them, please.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3015,46 +2978,7 @@ public void ChangeTerminalActiveFolder(string path)
...
-
-        private void ClearTerminalCommandLineWithEscapeAndRunCommand(string command)
-        {
-            if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(command))
-            {
-                return;
-            }
-
-            _terminal.RunningSession.WriteInputTextAsync("\x1B" + command + Environment.NewLine);
+            _terminal.ChangeFolder(path);
```
mstv: Then
```suggestion
            _terminal?.ChangeFolder(path);
```

GitUI/ShellHelper.cs
mstv:
```suggestion
                    const string justBash = "bash.exe"; // Generic bash, should generally be in the git dir
```

GitUI/CommandsDialogs/FormBrowse.cs
mstv: Remove the following lines, please. These checks are done by `_terminal.ChangeFolder`. No need to do them twice.
```
    if (_terminal?.RunningSession == null || string.IsNullOrWhiteSpace(path))
    {
        return;
    }
```

asherber: Sorry, I must have missed these comments earlier. Fixed now.

mstv:
Sorry, our toolchain regarding `English.xlf` is under reconstruction yet: #7873.

mstv:
#7873 has been fixed. So please rebase, run `Build.cmd -loc` and commit `English.xlf`.

asherber:
@mstv Sorry, I don't understand. I ran `build -loc`, but there are no changed files. Do you just want me to add the two new `TranslationString`s to `English.xlf` by hand?

RussKie:
> I ran `build -loc`, but there are no changed files.

That's strange.
Can you please rebase on top of the latest master and try again?

asherber:
Same result.

asherber:
I think I see the problem.
```
 GitExtensions -> C:\Users\Aaron\Documents\Projects\Visual Studio\gitextensions\artifacts\bin\GitExtensions\Debug\net461\GitEx
  tensions.exe
  'C:\Users\Aaron\Documents\Projects\Visual' is not recognized as an internal or external command,
  operable program or batch file.

  Verifying English translations are up-to date
  -------------------------------------------------------
  English translations are up-to date
```
Whatever is calling the translations app is missing quotes around the command line -- my projects dir has a space in it.

When I move the folder somewhere with no space in the path, `build -loc` gives me a changed `English.xlf`.

PR #7691, opened 2020-01-26
Update NUnit

sharwell:
Update to NUnit 3.12.0. This release improves support for asynchronous tests and synchronization contexts, both of which we rely on.

PR #7692, opened 2020-01-26
Join pending operations asynchronously

sharwell:
Builds on #7691

### DIFF
--- a/UnitTests/CommonTestUtils/AsyncTestHelper.cs
+++ b/UnitTests/CommonTestUtils/AsyncTestHelper.cs
@@ -1,5 +1,12 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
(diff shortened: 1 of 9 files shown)


UnitTests/CommonTestUtils/AsyncTestHelper.cs
```diff
@@ -15,64 +12,5 @@ public static TimeSpan UnexpectedTimeout
                 return Debugger.IsAttached ? TimeSpan.FromHours(1) : TimeSpan.FromMinutes(1);
             }
         }
```
mstv: I'd move this single remaining element to `GitExtUtils/GitUI/ThreadHelper.cs` where `JoinPendingOperationsAsync(TimeSpan timeout)` was added.

sharwell: I ended up moving the other way, since `ThreadHelper` is production code.

sharwell:
Merging with just one review since this only touches test code *and* the reviewer is the other developer actively working on test stability.

PR #7694, opened 2020-01-27
Delete multiple files at a same time with popup menu

### LINKED ISSUE #7637: Can't delete more than a file at a time
## Current behaviour

Since #6296, we can't delete more than a file at a time in RevisionDiffControl. I wonder why ?

## Expected behaviour

Be able to delete multiple file at a time.

## Steps to reproduce

In a repository, modify two files
Open Git Extensions
Click on "Working directory"
Go on Diff tab
Select the two files
Click right
Delete file is not visible.

No idea.

## Environment

- GIT version: 2.24.1.windows.2
- OS version: Microsoft Windows NT 10.0.18363.0
- .NET version: 4.8.4075.0

bansan85:
Fixes #7637

## Proposed changes

- Add possibility to delete multiple files
- Add possibility to delete file even if a submodule is selected
- 

### Before

"Detele File" not shown is Diff window if multiple file was selected;

### After

It's possible to delete multiple files like SUPPR keyboard allow it.

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -115,9 +115,11 @@ namespace GitUI.CommandsDialogs
                 return false;
             }
 
+            UpdateStatusOfMenuItems();
+
             switch ((Command)cmd)
             {
-                case Command.DeleteSelectedFiles: return DeleteSelectedFiles();
+                case Command.DeleteSelectedFiles: diffDeleteFileToolStripMenuItem.PerformClick(); break;
                 case Command.ShowHistory: fileHistoryDiffToolstripMenuItem.PerformClick(); break;
                 case Command.Blame: blameToolStripMenuItem.PerformClick(); break;
                 case Command.OpenWithDifftool: firstToSelectedToolStripMenuItem.PerformClick(); break;
@@ -407,7 +409,7 @@ namespace GitUI.CommandsDialogs
             _revisionFileTree.ExpandToFile(DiffFiles.SelectedItems.First().Name);
         }
 
-        private void DiffContextMenu_Opening(object sender, CancelEventArgs e)
+        private void UpdateStatusOfMenuItems()
         {
             var selectionInfo = GetSelectionInfo();
 
@@ -430,7 +432,8 @@ namespace GitUI.CommandsDialogs
             resetFileToToolStripMenuItem.Enabled = _revisionDiffController.ShouldShowResetFileMenus(selectionInfo);
 
             diffDeleteFileToolStripMenuItem.Text = ResourceManager.Strings.GetDeleteFile(selectionInfo.SelectedGitItemCount);
-            diffDeleteFileToolStripMenuItem.Visible = _revisionDiffController.ShouldShowMenuDeleteFile(selectionInfo);
+            diffDeleteFileToolStripMenuItem.Enabled = _revisionDiffController.ShouldShowMenuDeleteFile(selectionInfo);
+            diffDeleteFileToolStripMenuItem.Visible = diffDeleteFileToolStripMenuItem.Enabled;
             diffEditWorkingDirectoryFileToolStripMenuItem.Visible = _revisionDiffController.ShouldShowMenuEditWorkingDirectoryFile(selectionInfo);
             diffOpenWorkingDirectoryFileWithToolStripMenuItem.Visible = _revisionDiffController.ShouldShowMenuEditWorkingDirectoryFile(selectionInfo);
             diffOpenRevisionFileToolStripMenuItem.Visible = _revisionDiffController.ShouldShowMenuOpenRevision(selectionInfo);
@@ -446,20 +449,22 @@ namespace GitUI.CommandsDialogs
                                                _revisionDiffController.ShouldShowMenuEditWorkingDirectoryFile(selectionInfo) ||
                                                _revisionDiffController.ShouldShowMenuOpenRevision(selectionInfo);
 
-            // openContainingFolderToolStripMenuItem.Enabled or not
-            {
-                openContainingFolderToolStripMenuItem.Enabled = false;
+            openContainingFolderToolStripMenuItem.Enabled = false;
 
-                foreach (var item in DiffFiles.SelectedItems)
+            foreach (var item in DiffFiles.SelectedItems)
+            {
+                string filePath = _fullPathResolver.Resolve(item.Name);
+                if (FormBrowseUtil.FileOrParentDirectoryExists(filePath))
                 {
-                    string filePath = _fullPathResolver.Resolve(item.Name);
-                    if (FormBrowseUtil.FileOrParentDirectoryExists(filePath))
-                    {
-                        openContainingFolderToolStripMenuItem.Enabled = true;
-                        break;
-                    }
+                    openContainingFolderToolStripMenuItem.Enabled = true;
+                    break;
                 }
             }
+        }
+
+        private void DiffContextMenu_Opening(object sender, CancelEventArgs e)
+        {
+            UpdateStatusOfMenuItems();
         }
 
         private void blameToolStripMenuItem_Click(object sender, EventArgs e)
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -278,8 +278,7 @@ namespace GitUI.CommandsDialogs
 
             // Combined diff is a display only diff, no manipulations
             bool isAnyCombinedDiff = DiffFiles.SelectedItemParents.Any(item => item.ObjectId == ObjectId.CombinedDiffId);
-            bool isExactlyOneItemSelected = DiffFiles.SelectedItems.Count() == 1;
-            bool isAnyItemSelected = DiffFiles.SelectedItems.Any();
+            int selectedGitItemCount = DiffFiles.SelectedItems.Count();
 
             // No changes to files in bare repos
             bool isBareRepository = Module.IsBareRepository();
@@ -287,17 +286,16 @@ namespace GitUI.CommandsDialogs
             bool isAnyIndex = DiffFiles.SelectedItems.Any(item => item.Staged == StagedStatus.Index);
             bool isAnyWorkTree = DiffFiles.SelectedItems.Any(item => item.Staged == StagedStatus.WorkTree);
             bool isAnySubmodule = DiffFiles.SelectedItems.Any(item => item.IsSubmodule);
-            bool singleFileExists = isExactlyOneItemSelected && File.Exists(_fullPathResolver.Resolve(DiffFiles.SelectedItem.Name));
+            bool allFilesExist = DiffFiles.SelectedItems.All(item => File.Exists(_fullPathResolver.Resolve(item.Name)));
 
             var selectionInfo = new ContextMenuSelectionInfo(DiffFiles.Revision,
                 firstIsParent: firstIsParent,
                 isAnyCombinedDiff: isAnyCombinedDiff,
-                isSingleGitItemSelected: isExactlyOneItemSelected,
-                isAnyItemSelected: isAnyItemSelected,
+                selectedGitItemCount: selectedGitItemCount,
                 isAnyItemIndex: isAnyIndex,
                 isAnyItemWorkTree: isAnyWorkTree,
                 isBareRepository: isBareRepository,
-                singleFileExists: singleFileExists,
+                allFilesExist: allFilesExist,
                 isAnyTracked: isAnyTracked,
                 isAnySubmodule: isAnySubmodule);
             return selectionInfo;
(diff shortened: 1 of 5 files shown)


bansan85:
Mmmm… appveyor fails but I can't figure if it's my fault or not… Does someone has an idea ?

mstv:
It's not your fault. The tests fail randomly. I think it will not take long anymore until this will improve significantly.
I have restarted the build.

bansan85:
> It's not your fault. The tests fail randomly. I think it will not take long anymore until this will improve significantly.
> I have restarted the build.

Thanks, I was afraid it was my change in text "Delete file" => "Delete file(s)" that cause the translation tool to fail.

GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs
```diff
@@ -305,7 +305,7 @@ private void InitializeComponent()
             this.diffDeleteFileToolStripMenuItem.Image = global::GitUI.Properties.Images.DeleteFile;
             this.diffDeleteFileToolStripMenuItem.Name = "diffDeleteFileToolStripMenuItem";
             this.diffDeleteFileToolStripMenuItem.Size = new System.Drawing.Size(296, 22);
-            this.diffDeleteFileToolStripMenuItem.Text = "Delete file";
+            this.diffDeleteFileToolStripMenuItem.Text = "Delete file(s)";
```
gerhardol: I would like to remove (s) in all situations as it is used inconsistently and is hard to translate, so I prefer to not add another location
..and this will affect the translation, not sure why it is not detected...

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -274,6 +274,7 @@ private ContextMenuSelectionInfo GetSelectionInfo()
             bool isAnyWorkTree = DiffFiles.SelectedItems.Any(item => item.Staged == StagedStatus.WorkTree);
             bool isAnySubmodule = DiffFiles.SelectedItems.Any(item => item.IsSubmodule);
             bool singleFileExists = isExactlyOneItemSelected && File.Exists(_fullPathResolver.Resolve(DiffFiles.SelectedItem.Name));
+            bool isAnyFileExist = DiffFiles.SelectedItems.Any(item => File.Exists(_fullPathResolver.Resolve(item.Name)));
```
gerhardol: isAnyExistingFile
Even my English have a problem with that
This adds I/O when opening menus, is this OK?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -820,11 +822,11 @@ private bool DeleteSelectedFiles()
                 {
                     var path = _fullPathResolver.Resolve(item.Name);
                     bool isDir = (File.GetAttributes(path) & FileAttributes.Directory) == FileAttributes.Directory;
-                    if (isDir)
```
gerhardol: Hard to understand (if this needed, see below)

                    if (isDir)
                        if (deleteFolders)
                        {
                            Directory.Delete(path, true);
                        }
                    }
                    else if (deleteFiles)
                    {
                        File.Delete(path);
                    }

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -843,7 +845,7 @@ private bool DeleteSelectedFiles()
 
         private void diffDeleteFileToolStripMenuItem_Click(object sender, EventArgs e)
         {
-            DeleteSelectedFiles();
+            DeleteSelectedFiles(deleteFiles: true, deleteFolders: false);
```
gerhardol: The deletion should be what you delete, i.e. the menu item is only visible if all selected can be deleted.

GitUI/CommandsDialogs/RevisionDiffController.cs
```diff
@@ -121,8 +124,7 @@ public bool ShouldShowMenuEditWorkingDirectoryFile(ContextMenuSelectionInfo sele
 
         public bool ShouldShowMenuDeleteFile(ContextMenuSelectionInfo selectionInfo)
         {
-            return selectionInfo.IsSingleGitItemSelected && !selectionInfo.IsAnySubmodule &&
-                   selectionInfo.SingleFileExists && selectionInfo.SelectedRevision.IsArtificial;
+            return selectionInfo.IsAnyFileExist && selectionInfo.SelectedRevision.IsArtificial;
```
gerhardol: IsArtificial should be working tree only (no new issue - Index is handled but not entirely logical)

The option should not be active when there are items that cannot be deleted. Maybe OK to check if there are items that cannot be deleted?

            return !selectionInfo.IsAnySubmodule && selectionInfo.SelectedRevision.IsAnyItemWorkTree ;

bansan85:
@gerhardol 
I wanted to enable the "delete file" feature if at least one file can be deleted, no matter if a submodule is selected. This is why I added an option to delete file only.

I based my decision on the "Reset file(s) to" -> "First" behavior. It is enable even if a submodule is selected. Does this behavior should be corrected ?

GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs
RussKie: Have a look at `ResourceManager.Strings` - we have examples of single/multiple formatted strings. Please use the same mechanism.

RussKie:
@gerhardol bump

gerhardol:
> I based my decision on the "Reset file(s) to" -> "First" behavior. It is enable even if a submodule is selected. Does this behavior should be corrected ?

Submodules can be reset in index. The version actually checked out is never changed by the normal reset.
So reset worktree to index will never have any effect for submodules and should be disabled.

As this is just confusing, it is not critical.

bansan85:
Sorry for the delay.
I changed the behavior to allow deletion only when all selected files exist. The "Delete file" is not shown if a folder is selected.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -807,7 +807,7 @@ private bool DeleteSelectedFiles()
                     {
                         Directory.Delete(path, true);
                     }
-                    else
+                    else if (!isDir)
```
mstv, line 810: This condition will always be fullfilled, doesn't it?

GitUI/CommandsDialogs/RevisionDiffController.cs
```diff
@@ -34,7 +34,7 @@ public sealed class ContextMenuSelectionInfo
             bool isAnyItemIndex = false,
             bool isAnyItemWorkTree = false,
             bool isBareRepository = false,
-            bool singleFileExists = true,
+            bool isAllFilesExists = true,
```
mstv, line 37:
```suggestion
            bool allFilesExist = true,
```

GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs
mstv: still open

GitUI/CommandsDialogs/RevisionDiffController.cs
```diff
@@ -117,13 +117,12 @@ public bool ShouldShowSubmoduleMenus(ContextMenuSelectionInfo selectionInfo)
 
         public bool ShouldShowMenuEditWorkingDirectoryFile(ContextMenuSelectionInfo selectionInfo)
         {
-            return selectionInfo.IsSingleGitItemSelected && !selectionInfo.IsAnySubmodule && selectionInfo.SingleFileExists;
+            return selectionInfo.IsSingleGitItemSelected && !selectionInfo.IsAnySubmodule && selectionInfo.IsAllFilesExists;
```
mstv, line 120: Does `Edit` support opening multiple files?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -287,7 +287,7 @@ private ContextMenuSelectionInfo GetSelectionInfo()
             bool isAnyIndex = DiffFiles.SelectedItems.Any(item => item.Staged == StagedStatus.Index);
             bool isAnyWorkTree = DiffFiles.SelectedItems.Any(item => item.Staged == StagedStatus.WorkTree);
             bool isAnySubmodule = DiffFiles.SelectedItems.Any(item => item.IsSubmodule);
-            bool singleFileExists = isExactlyOneItemSelected && File.Exists(_fullPathResolver.Resolve(DiffFiles.SelectedItem.Name));
+            bool isAllFilesExists = DiffFiles.SelectedItems.All(item => File.Exists(_fullPathResolver.Resolve(item.Name)));
```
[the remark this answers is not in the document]
bansan85: @mstv are you sure ? I just wanted that all bool start with "is"

GitUI/CommandsDialogs/RevisionDiffControl.cs
bansan85: Sure

GitUI/CommandsDialogs/RevisionDiffController.cs
bansan85: I don't think so. Plus, there was already the bool "IsSingleGitItemSelected"

mstv: You are right, `IsSingleGitItemSelected` does the trick.

GitUI/CommandsDialogs/RevisionDiffControl.cs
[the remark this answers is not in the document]
mstv: The only alternative would be `doAllFilesExist`.

GitUI/CommandsDialogs/RevisionDiffController.cs
gerhardol, line 120: This should also check that only one file is selected, add option isExactlyOneItemSelected
But remove ```selectionInfo.IsSingleGitItemSelected && !selectionInfo.IsAnySubmodule && ``` should not have been there before
```return selectionInfo.isExactlyOneItemSelected && selectionInfo.IsAllFilesExists;```

File: GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs (hunk 1 above)
Comment (gerhardol) [reply GitUI/CommandsDialogs/RevisionDiffControl.Designer.cs:309]: This could be shared with FormCommit (maybe FormResolveConflicts), no requirement to do so for me

File: GitUI/CommandsDialogs/RevisionDiffControl.cs (hunk 9 above)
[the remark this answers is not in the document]
Comment (bansan85) [reply GitUI/CommandsDialogs/RevisionDiffControl.cs]: I used with allFilesExist

**Comment by bansan85** [comment]
About the ``Text`` of ``diffDeleteFileToolStripMenuItem``, I used the same trick the one in ResourceManager/Strings.cs

File: GitUI/CommandsDialogs/RevisionDiffController.cs (hunk 8 above)
Comment (bansan85) [reply GitUI/CommandsDialogs/RevisionDiffController.cs]: @gerhardol I remove the ``IsAnySubmodule`.

``isExactlyOneItemSelected`` is the same than ``IsSingleGitItemSelected``.

**Comment by bansan85** [comment]
Appveyor fails but I don't think it's my fault.

**Comment by pmiossec** [comment]
@bansan85 I have triggered a rebuild of the PR. We will see.

Tips: you could trigger a new build by closing and reopening a PR. Could be useful to know it in such cases...

**Comment by pmiossec** [comment]
>Appveyor fails but I don't think it's my fault.

Fails again. Are you sure that's not your fault as it fails on a class you changed...

**Comment by gerhardol** [comment]
Tests fail

> 1) Failed : GitUITests.CommandsDialogs.RevisionDiffControllerTests.BrowseDiff_EditOpen_IsAnySubmodule(True)
  Expected _controller to be False, but found True.

File: GitUI/CommandsDialogs/RevisionDiffControl.cs [hunk 10]
```diff
@@ -431,6 +431,7 @@ private void DiffContextMenu_Opening(object sender, CancelEventArgs e)
             blameToolStripMenuItem.Enabled = _revisionDiffController.ShouldShowMenuBlame(selectionInfo);
             resetFileToToolStripMenuItem.Enabled = _revisionDiffController.ShouldShowResetFileMenus(selectionInfo);
 
+            diffDeleteFileToolStripMenuItem.Text = ResourceManager.Strings.GetDeleteFile(selectionInfo.IsSingleGitItemSelected ? 1 : 2);
```
mstv, line 434: There are languages with different declension for different numbers, e.g. Russian.
So better turn `IsSingleGitItemSelected` into `SelectedGitItemCount`.
Or stay with "Delete File(s)".
```suggestion
            diffDeleteFileToolStripMenuItem.Text = ResourceManager.Strings.GetDeleteFile(selectionInfo.SelectedGitItemCount);
```

bansan85:
I think I solved all problems. I also fixed all tests.

gerhardol:
> Using the hotkey Del, GE attempts to delete anyway.

Also note directories can be deleted with DEL, not with the menu (you can delete removed submodule directories).

RussKie:
> Using the hotkey `Del`, GE attempts to delete anyway.

This should be fixed as part of this PR.

> Also note directories can be deleted with DEL, not with the menu (you can delete removed submodule directories).

Is there something needs changing?

bansan85:
I add a commit so the hotkey Del now has the same behavior than the menu item Delete file

gerhardol:
> This should be fixed as part of this PR.
Is there something needs changing?

> I add a commit so the hotkey Del now has the same behavior than the menu item Delete file

This unification removes one of my use cases: Remove the directory of removed submodules
I created a PR with the suggested changes

For me, the difference was intentional between the menu and DEL, but this is better functionality (but worse implementation, more I/O when opening the menu)

PR #7696, opened 2020-01-28
Remove unexpected ConfigureAwait(false)

sharwell:


sharwell:
@RussKie Please do not rebase or squash my commits on merge

sharwell:
it's weird that the snipping tool doesn't allow typing...

RussKie:
I picked a habit from the MS repos - squash-merge. Though unlike the majority of the MS repos, I generally do it for single commit PRs, and a bit more judgemental when it comes to multi-commit PRs.
"Delete obsolete branches" normally removes merged branches for me, merged normally or via SM.

Besides you haven't marked the commit as "no squash" as you used to 😉

RussKie:
...and you have write access too 😉

gerhardol:
Merge commits are evil, makes the history very hard to read.
Bisects are a pain.
The only benefit in the GH workflow is that you always see the local branches merged as sharwell wants.

I would prefer to use --ff-only merges, but GH has no support for that. Also GH rebase will rewrite the commits (so local branches diverge). Also, ff-only can be a competition to merge. msftbot can be used I assume. GitLab has a built in check (that can be set with the rest interface) to merge when checks are fulfilled.

I propose that the default merge is changed to rebase or squash.
Open a separate issue for this discussion?

sharwell:
> Bisects are a pain.

Bisect tends to perform _better_ if merge commits are preserved.

mstv:
> I would prefer to use --ff-only merges

👍 

> I propose that the default merge is changed to rebase or squash.

I cannot agree neither nor.
`Rebase` loses any connection to the PR.
`Squash` makes the history very hard to read.

RussKie:
> `Rebase` loses any connection to the PR.

How so? If it is a manual rebase, why would it be disconnected? 

> `quash` makes the history very hard to read.

Depends on a PR. If a PR contains a single commit then it isn't any different (I generally squash such PRs).
If a PR contains multiple commits, and they are good to be preserved - I merge such commits as-is.

In a busy repo with multiple merges a day squash-merge makes life easier.

gerhardol:
>> Rebase loses any connection to the PR.

> How so? If it is a manual rebase, why would it be disconnected?

GH rebase merge always rewrites, also if the branch to merge is based on the branch to merge to (often master).
Not sure of the actual Git commands, but maybe rebase --no-ff; merge -ff-only

I still prefer squash/merge and prefer to rebase the commits to keep in the history myself instead of squashing all. 
But merge commits are evil.

mstv:
> > Rebase loses any connection to the PR.

> How so? If it is a manual rebase, why would it be disconnected?

If you do it, the PR number is added to the commit message -- but not using the Web GUI (see below).

> GH rebase merge always rewrites, also if the branch to merge is based on the branch to merge to

Both are GH issues that should be improved.

grafik

result:
grafik

Issue #7436, opened 2019-11-21
[NBug] Приложение не найдено


ruslan528:
:warning: The sections below must be filled in and this text must be removed or the issue will be closed.

## Error Details
```
System.ComponentModel.Win32Exception (0x80004005): Приложение не найдено
   в System.Diagnostics.Process.StartWithShellExecuteEx(ProcessStartInfo startInfo)
   в GitUI.CommitInfo.CommitInfoHeader.rtbRevisionHeader_LinkClicked(Object sender, LinkClickedEventArgs e)
   в System.Windows.Forms.RichTextBox.OnLinkClicked(LinkClickedEventArgs e)
   в System.Windows.Forms.RichTextBox.EnLinkMsgHandler(Message& m)
   в System.Windows.Forms.RichTextBox.WmReflectNotify(Message& m)
   в System.Windows.Forms.RichTextBox.WndProc(Message& m)
   в System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

## Additional information
when click in Commit tab on Author link

## Environment
- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.21.0.windows.1 (recommended: 2.22.0 or later)
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.3752.0
- DPI 96dpi (no scaling)

ruslan528:
Nice work)

RussKie:
Clicking on a link should launch your system default email client. Do you
have one configured?

ruslan528:
You are right, there is no one. But it better to show simple message "Check your default email app" instead of crush 😉

pmiossec:
❤️ ❤️ ❤️ ❤️ **crush** ❤️ ❤️ ❤️ ❤️ is always preferable 🤣 
But indeed, a better message is always better than **crash**

f-i-x-7:
The related PR was merged. I think that issue should be closed, shouldn't it?

mstv:
Aye, thank you

PR #7714, opened 2020-02-01
GE should not crash when no email applications are available and user…

manuelspezzani:
… clicks on a "mailto:" link (issue #7436)

Fixes issue #7436 + minor refactoring

## Proposed changes

- Added a try..catch to avoid crashing the application in case Process.Start fails (e.g. no registered applications found for the clicked link)
- Replaced try..new Uri(link)..catch with Uri.TryParse

## Test methodology 

Manual tests

## Test environment(s) 

- Git Extensions 3.3.1.7897
- Git 2.22.0.windows.1 (recommended: 2.23.0 or later)
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfoHeader.cs
+++ b/GitUI/CommitInfo/CommitInfoHeader.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq;
 using System.Reactive.Linq;
@@ -123,27 +124,29 @@ namespace GitUI.CommitInfo
         {
             var link = _linkFactory.ParseLink(e.LinkText);
 
-            try
+            if (Uri.TryCreate(link, UriKind.Absolute, out var uri))
             {
-                var result = new Uri(link);
-                if (result.Scheme == "gitext")
+                if (uri.Scheme == "gitext")
                 {
-                    CommandClicked?.Invoke(sender, new CommandEventArgs(result.Host, result.AbsolutePath.TrimStart('/')));
+                    CommandClicked?.Invoke(sender, new CommandEventArgs(uri.Host, uri.AbsolutePath.TrimStart('/')));
                 }
                 else
                 {
-                    using (var process = new Process
+                    using var process = new Process
                     {
                         EnableRaisingEvents = false,
-                        StartInfo = { FileName = result.AbsoluteUri }
-                    })
+                        StartInfo = { FileName = uri.AbsoluteUri }
+                    };
+
+                    try
                     {
                         process.Start();
                     }
+                    catch (Exception ex)
+                    {
+                        MessageBox.Show(this, ex.Message, new TranslationString("Error").Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                    }
                 }
-            }
-            catch (UriFormatException)
-            {
             }
         }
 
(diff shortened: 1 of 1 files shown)


GitUI/CommitInfo/CommitInfoHeader.cs
```diff
@@ -123,28 +124,30 @@ private void rtbRevisionHeader_LinkClicked(object sender, LinkClickedEventArgs e
...
+                        StartInfo = { FileName = uri.AbsoluteUri }
+                    };
+
+                    try
                     {
                         process.Start();
                     }
+                    catch (Exception ex)
+                    {
+                        MessageBox.Show(this, ex.Message, new TranslationString("Error").Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
```
RussKie: We don't inline `TranslationString`s.
Please move it up the top as done everywhere else.

RussKie: Please provide a screenshot of the messagebox.
I suspect we may need to add some instructions to it

GitUI/CommitInfo/CommitInfoHeader.cs
```diff
@@ -123,28 +124,30 @@ private void rtbRevisionHeader_LinkClicked(object sender, LinkClickedEventArgs e
         {
             var link = _linkFactory.ParseLink(e.LinkText);
 
-            try
+            if (Uri.TryCreate(link, UriKind.Absolute, out var uri))
```
RussKie: Please invert `if` to reduce nesting

GitUI/CommitInfo/CommitInfoHeader.cs
pmiossec: I could add that if the TranslationString is inlined, the string is not detected by the Translation app and the string is not translated.

GitUI/CommitInfo/CommitInfoHeader.cs
```diff
@@ -124,28 +125,30 @@ private void rtbRevisionHeader_LinkClicked(object sender, LinkClickedEventArgs e
...
-            if (Uri.TryCreate(link, UriKind.Absolute, out var uri))
+            if (!Uri.TryCreate(link, UriKind.Absolute, out var uri))
             {
-                if (uri.Scheme == "gitext")
+                return;
+            }
+
+            if (uri.Scheme == "gitext")
+            {
+                CommandClicked?.Invoke(sender, new CommandEventArgs(uri.Host, uri.AbsolutePath.TrimStart('/')));
```
RussKie: [nit] If you add `return` here, we can remove the following `else` and remove one level of nesting.

GitUI/CommitInfo/CommitInfoHeader.cs
```diff
@@ -124,28 +125,30 @@ private void rtbRevisionHeader_LinkClicked(object sender, LinkClickedEventArgs e
...
+                {
+                    EnableRaisingEvents = false,
+                    StartInfo = { FileName = uri.AbsoluteUri }
+                };
+
+                try
                 {
-                    CommandClicked?.Invoke(sender, new CommandEventArgs(uri.Host, uri.AbsolutePath.TrimStart('/')));
+                    process.Start();
                 }
```
RussKie: I suggest placing the code together:
```cs
try
{
                using var process = new Process
                {
                    EnableRaisingEvents = false,
                    StartInfo = { FileName = uri.AbsoluteUri }
                };
                process.Start();
}
```

mstv:
@RussKie, I think your requested changes have been addressed.

RussKie:
I'll try to review as soon as possible

RussKie:
Thank you

PR #7718, opened 2020-02-02
Cleanup interops

RussKie:
## Proposed changes

* Consolidate interop definitions under Interop folder
* Break defitnions one per file
* Fix DllImport defintions
* Fix struct defintions (from classes) to ensure blitting
* Fix naming (e.g. use explicit Unicode variants)
* Optimise several calls


gerhardol:
The change in general seem OK
No objections - changes are hard to see

RussKie:
> No objections - changes are hard to see

I've reworked the PR and separated changes to make them easy to follow, please have a look

PR #7719, opened 2020-02-02
Refactor: Compare with ObjectId rather than Guid

gerhardol:
Originated as a part #4154, but broken out

## Proposed changes

Comparisions should not use GitRevision.Guid but GitRevision.ObjectId, as Guid => ObjectId.ToString()

Probably not making any noticeable difference, but sets the example and is a step to remove Guid

I would like to change e.g.  GitRevision.WorkTreeGuid -> ObjectId.WorkTreeId.ToString() when comparing to raw strings, but this will be slightly less efficient.
(The Guid version is used to compare strings, but this makes it more difficult to find the Artificial commits. There are tests that the two definitions are identical.)
 
## Test methodology 

Tests updated

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1044,7 +1044,7 @@ namespace GitCommands
             if (loadRefs)
             {
                 revision.Refs = GetRefs()
-                    .Where(r => r.Guid == revision.Guid)
+                    .Where(r => r.ObjectId == revision.ObjectId)
                     .ToList();
             }
 
@@ -3009,7 +3009,7 @@ namespace GitCommands
             foreach (var gitRef in gitRefs)
             {
                 if (headByRemote.TryGetValue(gitRef.Remote, out var defaultHead) &&
-                    gitRef.Guid == defaultHead.Guid)
+                    gitRef.ObjectId == defaultHead.ObjectId)
                 {
                     headByRemote.Remove(gitRef.Remote);
                 }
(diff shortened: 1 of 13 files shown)


gerhardol:
@msftbot merge in 42 hours

gerhardol:
Added a few more occasions, rebased

RussKie:
> msftbot merge in 42 hours

LOL how do you decided on the time range Gerhard?

gerhardol:
OT
> LOL how do you decide on the time range Gerhard?

I believe it is fair to give 2-3 days for others to comment on them - or you to merge. The auto merge are out of the load of PRs to review, at least.

This time I wanted to create another PR on this commit and wanted a not too late time to submit the PR. Since 42 is the ultimate answer, I took that instead of 45...

RussKie:
> Since 42 is the ultimate answer, I took that instead of 45...

...My math professor taught us that 28 was the ultimate answer.
And you do know that 2 days equals to 48 hours, right? 😄

gerhardol:
Douglas Adams said the answer is 42.

Setting merge at midnight, wanting to add something two days later but a little earlier -> 45h, but 42 is more fun

mstv:
`GitRevisionInfoProviderTests` seem to be broken, @gerhardol:
https://ci.appveyor.com/api/buildjobs/frtj51aw6amssle4/artifacts/artifacts%2FTestResults%2FRelease%2FGitCommands.Tests_net461_x64.log
Some of them fail locally, too.
Since the switch to Arcade build, the test results (and build errors) are ignored.

gerhardol:
> GitRevisionInfoProviderTests seem to be broken

@mstv 
#7731

PR #7730, opened 2020-02-05
WIP TRIAL Improve ScriptRunnerTests & RepoObjectsTreeTest 

mstv:
This is a copy of #7625 which I don't want to touch because it reproduces a bug -- I guess -- in one of the libraries we use: .net Threading, NUnit and/or AppVeyor's extended test runner.

## Proposed changes

Improve the most problematic `ScriptRunnerTests` and `FormCommitTests`

- use `AsyncTestHelper.WaitForPendingOperations` for the testcases
- assert that no error is displayed by `ScriptRunner.RunScript`
- use `formBrowse` as owner for potential subdialogs of `ScriptRunner.RunScript`
- close and dispose Form in `UITests.RunForm`
- join the test with timeout in `UITests.RunForm`
- "printf debugging" with avoided `Console.WriteLine` during `UITests.RunForm`

## Test environment(s) 

- AppVeyor
- Windows 10

----


UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs
```diff
@@ -21,13 +21,16 @@ public sealed class ConfigureJoinableTaskFactoryAttribute : Attribute, ITestActi
...
+        public static readonly LoggingService LoggingService = new LoggingService() { AutoFlush = true };
+
         public ConfigureJoinableTaskFactoryAttribute()
         {
             Application.ThreadException += HandleApplicationThreadException;
         }
 
         public void BeforeTest(ITest test)
         {
+            LoggingService?.Log($"{nameof(BeforeTest)} entry");
```
mstv: I cannot find this output in the logs!
So I must think that our test attributes are not included correctly.
This also might be the cause why `Application.ThreadException += HandleApplicationThreadException` seems to have no effect.

mstv: Locally, it works. Might be the same issue with the bin folder structure on AppVeyor as reported by @pmiossec.

mstv: No, this is even worse: There is output generated by the `ConfigureJoinableTaskFactoryAttribute.LoggingService` instance. Hence the assembly _is_ found and loaded.
But `BeforeTest` and `AfterTest` are not called on AppVeyor.
The most calming cause would be that the `cibuild.cmd` does not include the `AssemblyInfo.cs` files, @RussKie.

mstv: It is not `-ci` but `-c Release`.

mstv: Is there a tool for the configuration of MS Arcade? XML is not intended to be written by humans.

RussKie: The Arcade SDK is a set of scripts and MSBuild props and targets (https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md). 
I've been working with them in a VSCode, and inspecting binlogs. I'm not aware of any editors for the Arcade scripts, or any other "saner" way of working with it.

> The most calming cause would be that the `cibuild.cmd` does not include the `AssemblyInfo.cs` files

`cibuild` is a convenience helper, that essentially just calls `.\Build\build.ps1` with given parameters.
`.\Build\build.ps1` runs few different types of builds, depending on the input parameters - e.g. build C++ projects, build C# projects, etc.

I can't imagine `AssemblyInfo.cs` file aren't included, if they are part of the project.
You can see pretty much all aspects of a build inspecting a binlog, something like `CscCore` target that lists the compilation command and all files that were passed to the compiler.

RussKie: One way to confirm whether `AssemblyInfo.cs` is included in a build is to write some garbage in it and observe a failing build because of this error.

RussKie: image

mstv: `AssemblyInfo.cs` is compiled but the contained `Attribute`s are ignored.
Just build the Release, set a breakpoint at `BeforeTest` and debug some integration test.

Have you seen the green builds of the Debug variant?

RussKie: > Just build the Release, set a breakpoint at `BeforeTest` and debug some integration test.

Release build have reduce debug symbols, so not all information may be available.

Having said that, I can hit a breakpoint in `BeforeTest`:
image

RussKie: I wouldn't be surprised if in Release builds these instructions get inlined or optimised in a way where it is no longer possible to hit a breakpoint.

mstv: > I cannot find this output in the logs

It was too late in the evening. I forgot `Log(..., debugOnly = false)`.
The tests seem to pass more stably with the Debug variant. Switched back to Release.

GintasS:

RussKie:
Is this still relevant?

PR #7732, opened 2020-02-05
AzureDevOps CI: Fix and improvements

pmiossec:
## Proposed changes

One commit for each (that could be reviewed separatly)
1. Make lighter and quicker api calls
    * by providing filters values in the query parameters (less build are returned instead of filtering them after)
    * by specifying the `properties` requested (less data returned for each build)
2.  Fix "in progress" display sometimes not refreshed for the case 
where the first display was done when the build was not yet started
i.e. `StartDate` was filled with a bad default value
that prevent later refresh once the build is really started
3. Better choose the running build to display when there is more than one build to display for a specific commit.

    * Before: Multiple build was returned and due to the condition in `BuildServerWatcher.cs`/`OnBuildInfoUpdate()`, the last started build is displayed.

    * Now: Return only one build by commit by better selecting the one displayed. The first started build is now return and displayed

## Screenshots 

Nothing really changed...

## Test methodology 

- Manual & debugger

## Test environment(s) 

- Git Extensions 3.3.0
- Build ddf5a30bd261aff2e1675ddbfe76fe2e1d61a5f4
- Git 2.20.1.windows.1
- .NET Framework 4.8.3815.0
- DPI 96dpi (no scaling)

### DIFF
--- a/Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
+++ b/Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
@@ -17,6 +17,7 @@ namespace AzureDevOpsIntegration
     public class ApiClient : IDisposable
     {
         private const string BuildDefinitionsUrl = "build/definitions?api-version=2.0";
+        private const string Properties = "properties=sourceVersion,status,buildNumber,result,definition,_links,startTime,finishTime";
         private readonly HttpClient _httpClient;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Plugins/BuildServerIntegration/AzureDevOpsIntegration/ApiClient.cs
```diff
@@ -109,26 +110,45 @@ public async Task<string> GetBuildDefinitionNameFromIdAsync(int buildId)
...
+            runningBuilds = new List<Build>();
 
-            if (!running.HasValue || running.Value)
+            // Select best build to display (the first started or if none, one that will start)
+            foreach (var commitBuilds in byCommitBuilds)
             {
-                return builds
-                    .Where(b => !running.HasValue || running.Value == b.IsInProgress)
-                    .Where(b => !sinceDate.HasValue || b.StartTime >= sinceDate.Value.ToUniversalTime());
+                var builds = commitBuilds.OrderBy(b => b.StartTime).ToArray();
```
RussKie: Why do we need to materialise the query to take a top element form it? Feels wasteful, especially for a plugin that could be constantly running for a life time of the app (which can be quite long) and with multiple instances of the app open at the same time - we could be consuming more memory than necessary...

To me `.OrderBy(b => b.StartTime)` followed by `FirstOrDefault(b => b.StartTime.HasValue)` looks sus.
Could you please elaborate on what we  are doing here?

Can you please write some tests with some real response data? We need to restructure the code a little, e.g.
```cs
        public async Task<IEnumerable<Build>> QueryRunningBuildsAsync(string buildDefinitionsToQuery)
        {
            string queryUrl = QueryForBuildStatus(buildDefinitionsToQuery, "cancelling,inProgress,none,notStarted,postponed") + "&api-version=2.0";
            var runningBuilds = (await HttpGetAsync<ListWrapper<Build>>(queryUrl)).Value;
            return ParseResponse(runningBuilds);
        }
```

...and test the `ParseResponse` method.

Hint, to assert blobs of data ApprovalTests can be quite useful.

pmiossec: >Could you please elaborate on what we are doing here?

You're right, that was not optimal. It was to reuse the ordered values after the `??` if no values was found but thinking about it again that was not necessary as we could pick the first value in the unsorted collection.

So fixed.

I will have a look at the other concern...

pmiossec:
I have added tests on the code added...

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -205,5 +240,22 @@ public void Dispose()
             _apiClient?.Dispose();
             GC.SuppressFinalize(this);
         }
+
+        #region TestAccessor
+        internal TestAccessor GetTestAccessor() => new TestAccessor(this);
+
+        internal readonly struct TestAccessor
+        {
+            public AzureDevOpsAdapter Sut { get; }
```
RussKie: Make it readable
```suggestion
            public AzureDevOpsAdapter AzureDevOpsAdapter { get; }
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -11,6 +14,8 @@
 using GitUIPluginInterfaces.BuildServerIntegration;
 using Microsoft.VisualStudio.Threading;
 
+[assembly:InternalsVisibleTo("AzureDevOpsIntegration.Tests")]
```
RussKie: Attributes like this are better kept in dedicated files like AssemblyInfo.cs or InternalsVisibleTo.cs.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -205,5 +240,22 @@ public void Dispose()
...
+        internal TestAccessor GetTestAccessor() => new TestAccessor(this);
+
+        internal readonly struct TestAccessor
+        {
+            public AzureDevOpsAdapter Sut { get; }
+
+            public TestAccessor(AzureDevOpsAdapter sut)
+            {
+                Sut = sut;
+            }
```
RussKie:
```suggestion
            public TestAccessor(AzureDevOpsAdapter azureDevOpsAdapter)
            {
                AzureDevOpsAdapter = azureDevOpsAdapter;
            }
```

UnitTests/Plugins/BuildServerIntegration/AzureDevOpsIntegration.Tests/AzureDevOpsAdapterTests.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace AzureDevOpsIntegrationTests
+{
+    [TestFixture]
+    public class AzureDevOpsAdapterTests
+    {
+        [Test]
+        public void Should_not_filter_running_builds_When_only_one_returned()
+        {
+            var adapter = new AzureDevOpsAdapter();
+            var testAccessor = adapter.GetTestAccessor();
```
RussKie: [fyi] These could be moved into `Setup` method, so all tests don't have to do this.

pmiossec:
done.

pmiossec:
@msftbot merge in 1 day

PR #7739, opened 2020-02-07
FormCommit: Unaccessible COMMITMESSAGE

### LINKED ISSUE #6908: [NBug] Access to the path 'C:\Users\michal.kulinski\source\r-us\.g...

## Error Details
```
System.UnauthorizedAccessException: Access to the path 'C:\Users\michal.kulinski\source\r-us\.git\COMMITMESSAGE' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
   at System.IO.File.InternalWriteAllText(String path, String contents, Encoding encoding, Boolean checkHost)
   at System.IO.Abstractions.FileWrapper.WriteAllText(String path, String contents, Encoding encoding)
   at GitCommands.CommitMessageManager.set_MergeOrCommitMessage(String value)
   at GitUI.CommandsDialogs.FormCommit.OnFormClosing(FormClosingEventArgs e)
   at System.Windows.Forms.Form.CheckCloseDialog(Boolean closingOnly)
```

## Environment
- Git Extensions 3.1.1.6049
- Git 2.21.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #6908

## Proposed changes

Popup instead of NBug if the COMMITMESSAGE cannot be accessed

### Before

NBug

### After

image

image

## Test methodology 

Test by patching code only, not reproduced the problem.

----

### DIFF
--- a/GitCommands/CommitMessageManager.cs
+++ b/GitCommands/CommitMessageManager.cs
@@ -1,6 +1,8 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using System.IO.Abstractions;
 using System.Text;
+using System.Windows.Forms;
 using JetBrains.Annotations;
 
 namespace GitCommands
@@ -35,6 +37,10 @@ namespace GitCommands
 
     public sealed class CommitMessageManager : ICommitMessageManager
     {
+        private string CannotReadCommitMessage => "Cannot read commit message";
+        private string CannotSaveCommitMessage => "Cannot save commit message";
+        private string CannotAccessFile => "Exception: \"{0}\" when accessing {1}";
+
         private readonly string _amendSaveStatePath;
         private readonly string _commitMessagePath;
         private readonly string _mergeMessagePath;
@@ -100,7 +106,20 @@ namespace GitCommands
                 }
 
                 var (file, exists) = GetMergeOrCommitMessagePath();
-                return exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                string result;
+                try
+                {
+                    result = exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                }
+                catch (Exception ex)
+                {
+                    MessageBox.Show(null, string.Format(CannotAccessFile, ex.Message, file),
+                        CannotReadCommitMessage, MessageBoxButtons.OK,
+                        MessageBoxIcon.Exclamation);
+                    result = string.Empty;
+                }
+
+                return result;
             }
             set
             {
@@ -116,7 +135,16 @@ namespace GitCommands
                         return;
                     }
 
-                    _fileSystem.File.WriteAllText(path, content, _commitEncoding);
+                    try
+                    {
+                        _fileSystem.File.WriteAllText(path, content, _commitEncoding);
+                    }
+                    catch (Exception ex)
+                    {
+                        MessageBox.Show(null, string.Format(CannotAccessFile, ex.Message, path),
+                            CannotSaveCommitMessage, MessageBoxButtons.OK,
+                            MessageBoxIcon.Exclamation);
+                    }
                 }
             }
         }


GitCommands/CommitMessageManager.cs
```diff
@@ -42,6 +44,10 @@ public sealed class CommitMessageManager : ICommitMessageManager
         private Encoding _commitEncoding;
         private IFileSystem _fileSystem;
 
+        private const string _cannotReadCommitMessage = "Cannot read commit message";
```
gerhardol: TranslationString cannot be used in GitCommands

GitCommands/CommitMessageManager.cs
```diff
@@ -100,7 +106,20 @@ public string MergeOrCommitMessage
                 }
 
                 var (file, exists) = GetMergeOrCommitMessagePath();
-                return exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                string result;
+                try
+                {
+                    result = exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                }
+                catch (UnauthorizedAccessException)
```
gerhardol: Could be any Exception - not relevant to crash due to this

GitCommands/CommitMessageManager.cs
mstv: They can be moved to `Strings.cs`.

GitCommands/CommitMessageManager.cs
mstv: Yes, I'd prefer to catch all exceptions.

gerhardol: updated

GitCommands/CommitMessageManager.cs
```diff
@@ -116,7 +135,16 @@ public string MergeOrCommitMessage
...
+                    try
+                    {
+                        _fileSystem.File.WriteAllText(path, content, _commitEncoding);
+                    }
+                    catch (UnauthorizedAccessException)
+                    {
+                        MessageBox.Show(null, string.Format(_cannotAccessFile, path),
+                            _cannotSaveCommitMessage, MessageBoxButtons.OK,
+                            MessageBoxIcon.Exclamation, MessageBoxDefaultButton.Button1);
+                    }
```
RussKie: This look a duplicated code. I propose to move it out into an own method, so we provide a consistent experience across the module. 
...And, of course, we could test it (later as `MessageBox` isn't quite test-friendly) :)

GitCommands/CommitMessageManager.cs
gerhardol: Which file?

GitCommands/CommitMessageManager.cs
```diff
@@ -100,7 +106,20 @@ public string MergeOrCommitMessage
...
+                string result;
+                try
+                {
+                    result = exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                }
+                catch (UnauthorizedAccessException)
+                {
+                    MessageBox.Show(null, string.Format(_cannotAccessFile, file),
+                        _cannotReadCommitMessage, MessageBoxButtons.OK,
+                        MessageBoxIcon.Exclamation, MessageBoxDefaultButton.Button1);
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/CommitMessageManager.cs
gerhardol: The message differs slightly

I could do with ignoring the message box completely...

GitCommands/CommitMessageManager.cs
mstv: Open the command window (menu `View` | `Other Windows` | `Command Window`) and type "open str" ...

gerhardol: Of course I found a file called Strings.cs, but that is still in GitUI

mstv: You're right. But we have another: `ResourceManager\Strings.cs`.

RussKie:
if we missing a dedicated `Strings.cs` in GitCommands, then we need to
create one there.
ResourceManager is for globally shared strings.

gerhardol:
Updated

However, moving strings does not make sense, they are not to be shared and cannot be made into TranslationsStrings.
It makes more sense to remove the popup completely. The user has no option anyway, this just avoids the exception. A dismiss button in NBug would be useful (upper right corner seem to be unknown). (In this case a try-catch is needed anyway.)

I do not really want a discussion here, it is wasted effort that are better spent elsewhere and that just demotivates me. This is something that very few users will see (I cannot get rid of my file, I am living with it...)

Possible solutions:
 * Hardcoded strings as in the first commit
 * Only try-catch, no popup
 * Replace property with TrySave, TryRestore, returning also the temp COMMITMESSAGE file so the popup can include that too.
 * Move CommitMessageManager.cs to GitUI

mstv:
> Hardcoded strings as in the first commit

Let's return to this one. Those who tinker the `.git` folder in such a strange way, must live with an English message.
But we should also display `Exception.Message` (which is localized usually).

GitCommands/CommitMessageManager.cs
```diff
@@ -100,7 +106,20 @@ public string MergeOrCommitMessage
                 }
 
                 var (file, exists) = GetMergeOrCommitMessagePath();
-                return exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                string result;
+                try
+                {
+                    result = exists ? _fileSystem.File.ReadAllText(file, _commitEncoding) : string.Empty;
+                }
+                catch (Exception e)
```
RussKie: It is a convention to call this parameter `ex`. `e` is typically used for `EventArgs`
```suggestion
                catch (Exception ex)
```

GitCommands/CommitMessageManager.cs
```diff
@@ -116,7 +135,16 @@ public string MergeOrCommitMessage
                         return;
                     }
 
-                    _fileSystem.File.WriteAllText(path, content, _commitEncoding);
+                    try
+                    {
+                        _fileSystem.File.WriteAllText(path, content, _commitEncoding);
+                    }
+                    catch (Exception e)
```
RussKie:
```suggestion
                    catch (Exception ex)
```

GitCommands/CommitMessageManager.cs
gerhardol: fixed

gerhardol:
@msftbot merge in 1 day

gerhardol:
Rebased and squashed, no changes

PR #7741, opened 2020-02-08
fix: Plugins private binaries not resovled

### LINKED ISSUE #7735: Some plugins are broken because they fails to load there dependencies
## Current behaviour

* Plugins with assemblies dependencies which are not in the main folder of GitExtensions are broken.
* A lot of `FileNotFoundException`
* Consequently all the plugins are slow to load and some plugins are broken (AppVeyor CI, Azure DevOps CI, GitHub, ....)

It seems that the assemblies are not loaded from the good folder.

## Expected behaviour

Plugins loading without problems

## Steps to reproduce

1. Delete the `artifacts` folder
2. Compile and Run
3. Open GitExtensions repository
4. CI integration with AppVeyor is not visible on commits (silent failed)
5. Copy `NewtonSoft` dll from a plugin subfolder and paste it next to `GitExtensions.exe`
6. Restart the application and see CI integration with AppVeyor displayed.

## Screenshots

GitHub plugin not working (it seems to be due to restsharp that failed to load):
image

It seems to be due to changes done with Arcade build

## Environment

- GitExtensions version: `master` branch

## Diagnostics

* Copying the`NewtonSoft` dll in Gitextensions folder fix some plugins (that only have this missing dependency). So it seems that the dependency is loaded from the wrong folder.
* **Recreating the old folder architecture (before Arcade build) by copying the content of each plugins directly in the `Plugin` folder fix all the plugins.** but that doesn't help for the debugging which is still broken...

RussKie:
Resolves #7735

Fixes #

## Proposed changes

The recent change saw each plugin placed in own folder under the Plugins folder. This caused a number of plugins that had additional references (e.g. to Newtonsoft.Json) fail at runtime because these the app did not know where to probe for these assemblies.

Previously all plugins were placed in the same folder, Plugins, and the app was configured to probe in this folder (refer to app.config `configuration/runtime/ns:assemblyBinding/ns:probing`).

Now upon the start-up the app will ensure that all plugins under Plugin folder are added to the probing path. And if any of the plugins are missing, the probing path is updated and the app is restarted.

## Test methodology 

- TBD

----

### DIFF
--- a/GitExtensions/GitExtensions.csproj
+++ b/GitExtensions/GitExtensions.csproj
@@ -48,6 +48,31 @@
     <Copy SourceFiles="@(Translations)" DestinationFolder="$(TargetDir)Translation" ContinueOnError="false" />
   </Target>
 
+  <Target Name="EnumeratePlugins" AfterTargets="AfterBuild">
+    <PropertyGroup>
+      <PluginsFolder>$(TargetDir)Plugins</PluginsFolder>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Folders Include="$([System.IO.Directory]::GetDirectories(&quot;$(PluginsFolder)&quot;))" />
+      <OnlyDirs Include="@(Folders->Replace($(TargetDir), ''))"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <ProbingPrivatePath>Plugins;@(OnlyDirs)</ProbingPrivatePath>
+      <TargetAppConfig>@(AppConfigFileDestination)</TargetAppConfig>
+
+      <Namespace>
+        <Namespace Prefix="ns" Uri="urn:schemas-microsoft-com:asm.v1" />
+      </Namespace>
+    </PropertyGroup>
+
+    <XmlPoke XmlInputPath="$(TargetAppConfig.ToString())"
+            Namespaces="$(Namespace)"
+            Query="configuration/runtime/ns:assemblyBinding/ns:probing/@privatePath" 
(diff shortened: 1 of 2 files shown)


RussKie:
Unit tests are pending. I've done a bit of manual testing and it worked for me.
Please test, and let me know if any issues found.

RussKie:
I just realised this won't work for an installed version, so this proposed fix is only working for a dev version.
A release version (installed or portable) will have to have all paths already pre-populated. So it should be resolved as part of the build sequence.

Never the less, please check this one, see if it solves the overall issue.
I'll work more on the "how we get there" part.

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -82,5 +86,95 @@ public static void Unregister(IGitUICommands gitUiCommands)
...
+                return true;
+            }
+
+            if (!File.Exists(appConfig))
+            {
+                // no application config, presume no plugins is needed
+                // the app is broken, not going to try to fix it
+                return true;
+            }
+
```
pmiossec: There is a missing control to add here that makes that the `App.config` is always updated, the message box '"Plugins probing paths have been updated' is always displayed (but not restarted, I don't know why).

We need to update the list of paths in the `App.config` only if there is a missing path/

RussKie:
the first commit will reverted, it was late last night to undo it

RussKie:
I have cleaned the code, the probing path will be updated on build.

However it raised further questions how we would handle external plugins (e.g. those located under the user's profile).
@maraf  @mast-eu this is something we need to think through.

Issue #7742, opened 2020-02-08
JoinPendingOperationsAsync hangs in ConfigureJoinableTaskFactoryAttribute.AfterTest


mstv:
## Current behaviour

Occasionally, some tests hang in the `JoinPendingOperationsAsync`  in `ConfigureJoinableTaskFactoryAttribute.AfterTest` after the actual test has passed.
The `CancellationTokenSource(UnexpectedTimeout)` has no effect.
NUnit cancels the wait by a `ThreadAbortException`.
This seems to happen more often with the Release variant than with Debug.

Affected test cases (perhaps more):
- `GitUITests.CommitInfo.CommitInfoTests.GetSortedTags_should_remove_duplicate_refs`
- `GitUITests.CommitInfo.CommitInfoTests.GetSortedTags_should_load_ref_with_extra_spaces`
- `GitUITests.CommitInfo.CommitInfoTests.GetSortedTags_should_split_output_if_no_warning` 
- `UITests.CommandsDialogs.FormBrowseLeftPanelReorderNodesTest.RepoObjectTree_moving_node_legally_moves_it`
- `UITests.CommandsDialogs.FormBrowseLeftPanelReorderNodesTest.RepoObjectTree_moving_node_across_hidden_trees_skips_them`

example: https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30627268

## Expected behaviour

no timeout at all, especially no hang in `JoinPendingOperationsAsync`

## Steps to reproduce

run the test on AppVeyor several times

## Environment

- GE d5a44082
- AppVeyor
- Windows 10

mstv:
@sharwell, could you have a look at this, please?

sharwell:
https://github.com/gitextensions/gitextensions/pull/7747 may improve this

sharwell:
The hang may be caused by the thread exception UI getting shown. This UI needs to be disabled during tests, and instead log an exception that gets thrown during the next `AfterTest` execution.

sharwell:
The main issue here is background operations do not prevent the owning dialog from closing. As a broader work item, we need to modify each form to join its own background operations before closing, so we know after it is closed there is no other work left to join.

sharwell:
[...]

PR #7750, opened 2020-02-09
Avoid attempting to update non-existent UI

sharwell:
Fixes an exception observed during debugging #7742

### DIFF
--- a/GitCommands/Utils/EnvUtils.cs
+++ b/GitCommands/Utils/EnvUtils.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using Microsoft.Win32;
 
 namespace GitCommands.Utils
@@ -17,6 +18,22 @@ namespace GitCommands.Utils
                 default:
                     return false;
             }
+        }
+
+        public static bool RunningOnWindowsWithMainWindow()
+        {
+            if (!RunningOnWindows())
+            {
+                return false;
+            }
+
+            var currentProcess = Process.GetCurrentProcess();
+            if (currentProcess is null)
+            {
+                return false;
+            }
+
+            return currentProcess.MainWindowHandle != IntPtr.Zero;
         }
 
         public static bool IsWindowsVistaOrGreater()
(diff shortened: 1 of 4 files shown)


GitUI/TaskbarProgress.cs
```diff
@@ -7,7 +7,7 @@ public static class TaskbarProgress
     {
         private static void Try(Action<TaskbarManager> action)
         {
-            if (GitCommands.Utils.EnvUtils.RunningOnWindows() && TaskbarManager.IsPlatformSupported)
+            if (GitCommands.Utils.EnvUtils.RunningOnWindowsWithMainWindow() && TaskbarManager.IsPlatformSupported)
```
mstv: Nit: Not yours but noticed.
```suggestion
            if (EnvUtils.RunningOnWindowsWithMainWindow() && TaskbarManager.IsPlatformSupported)
```

GitCommands/Utils/EnvUtils.cs
```diff
@@ -19,6 +20,21 @@ public static bool RunningOnWindows()
...
+            {
+                return false;
+            }
+
+            if (!(Process.GetCurrentProcess() is { } currentProcess))
+            {
+                return false;
+            }
+
+            return currentProcess.MainWindowHandle != IntPtr.Zero;
```
mstv: (Strange syntax for C# amateurs.)

sharwell: It's a bit strange, but from the meetings I've seen to date it looks like this is going to stay as the "read to local and check for non-null" form.

sharwell: I can change this to the following if you prefer:

```csharp
var currentProcess = Process.GetCurrentProcess();
if (currentProcess is null)
{
  return false;
}
```

RussKie: I haven't seen this syntax until now, but I figured it would mean null-check. 
C# more and more becomes likes a JS... 😝

RussKie:
@msftbot merge when @mstv approves

GitUI/TaskbarProgress.cs
sharwell: Fixed

GitCommands/Utils/EnvUtils.cs
mstv: Yes, please. This is much more readable.

sharwell: Done

PR #7752, opened 2020-02-09
Avoid a 200ms delay when RunBackgroundAsync is cancelled

sharwell:


sharwell:
> (I can't resist - nit: old using syntax.)

It's not possible to implement this using the new syntax without producing a larger diff than necessary to implement this feature.

RussKie:
Could you please explain what will it to now?
Right now under a debugger I get a constant flood of OCE, will this change make it better/worse/no change?

sharwell:
> Right now under a debugger I get a constant flood of OCE, will this change make it better/worse/no change?

It shouldn't have any impact on that scenario. I do have some tricks for dealing with that though.

sharwell:
📝 Waiting until #7753 is merged before merging this one. The reduced delay increases the chances of hitting the failure that #7753 fixes.

RussKie:
> I do have some tricks for dealing with that though.

That would be good :)

PR #7753, opened 2020-02-09
Avoid deleting test directories while operations may be in progress

sharwell:
Fixes one source of hangs observed in #7742.

### DIFF
--- a/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
+++ b/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
@@ -59,8 +59,11 @@ namespace GitExtensions.UITests.CommandsDialogs
         }
 
         [TearDown]
-        public void TearDown()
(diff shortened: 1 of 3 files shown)


IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
```diff
@@ -59,8 +59,11 @@ public void SetUp()
         }
 
         [TearDown]
-        public void TearDown()
+        public async Task TearDown()
```
RussKie: Should we call this 
```suggestion
        public async Task TearDownAsync()
```

sharwell: I'll test to verify that the name doesn't impact whether the method gets called.

sharwell: Fixed these!

PR #7765, opened 2020-02-12
Improve exception-safety of UITests.RunForm

mstv:
Related to #7382

## Proposed changes

- `Close` and `Dispose` `Form` in `UITest.RunForm`
- try to close all Forms on exception in async task of `UITest.RunForm`
  (happens for many `FormCommitTests`)

## Test methodology 

- run the NUnit tests

## Test environment(s) 

- AppVeyor
- Windows 10

----

### DIFF
--- a/IntegrationTests/UI.IntegrationTests/UITest.cs
+++ b/IntegrationTests/UI.IntegrationTests/UITest.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitUI;
+using NUnit.Framework;
 
 namespace GitExtensions.UITests
 {
@@ -27,32 +28,69 @@ namespace GitExtensions.UITests
         }
 
         public static void RunForm<T>(
-            Action showDialog,
-            Func<T, Task> runAsync)
+            Action showForm,
+            Func<T, Task> runTestAsync)
             where T : Form
         {
-            // Avoid using ThreadHelper.JoinableTaskFactory for the outermost operation because we don't want the task
-            // tracked by its collection. Otherwise, test code would not be able to wait for pending operations to
-            // complete.
-            var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
+            // If form.Dipose was called by the async test task, closing the window would be done in a strange order.
+            T form = null;
+            try
             {
-                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-                await WaitForIdleAsync();
-                var dialog = Application.OpenForms.OfType<T>().Single();
-                try
-                {
-                    await runAsync(dialog);
-                }
-                finally
-                {
-                    dialog.Close();
-                }
-            });
+                // Start runTestAsync before calling showForm.
+                // The latter might block until the form is closed, especially if using Application.Run(form).
 
-            showDialog();
+                // Avoid using ThreadHelper.JoinableTaskFactory for the outermost operation because we don't want the task
+                // tracked by its collection. Otherwise, test code would not be able to wait for pending operations to
+                // complete.
+                var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
+                {
+                    try
+                    {
+                        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                        await WaitForIdleAsync();
+                        form = Application.OpenForms.OfType<T>().Single();
+                        await runTestAsync(form);
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            form?.Close();
+                        }
+                        finally
+                        {
+                            // Try to close all open Forms in order to let return the blocking Application.Run(form)
+                            // which might have been called by showForm() below.
+                            for (int trial = 0; trial < 3 && Application.OpenForms.Count > 0; ++trial)
+                            {
+                                Application.DoEvents();
+                                Application.OpenForms.OfType<Form>().ForEach(f =>
+                                {
+                                    try
+                                    {
+                                        f.Close();
+                                    }
+                                    catch (Exception)
+                                    {
+                                        // ignore
+                                    }
+                                });
+                            }
+                        }
+                    }
+                });
 
-            // Join the asynchronous test operation so any exceptions are rethrown on this thread
-            test.Join();
+                showForm();
+
+                // Join the asynchronous test operation so any exceptions are rethrown on this thread.
+                test.Join();
+            }
+            finally
+            {
+                form?.Close();
+                form?.Dispose();
+                Assert.IsTrue(Application.OpenForms.Count == 0, $"{Application.OpenForms.Count} open form(s) after test");
+            }
         }
 
         private readonly struct VoidResult
--- a/IntegrationTests/UI.IntegrationTests/UITest.cs
+++ b/IntegrationTests/UI.IntegrationTests/UITest.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Windows.Forms;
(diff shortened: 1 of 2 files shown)


IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +28,69 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
+                var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
                 {
-                    dialog.Close();
-                }
-            });
+                    try
+                    {
+                        await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                        await WaitForIdleAsync();
+                        form = Application.OpenForms.OfType<T>().Single();
```
sharwell: 📝 These lines do not need to be in the `try` block

IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +28,69 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
+                        await runTestAsync(form);
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            form?.Close();
```
sharwell: 📝 `form` cannot be null here once the lines above are moved back outside the `try` block

IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +28,69 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
+                                    try
+                                    {
+                                        f.Close();
+                                    }
+                                    catch (Exception)
+                                    {
+                                        // ignore
+                                    }
+                                });
+                            }
```
sharwell: I don't understand why we need this block. I would prefer to remove it.

IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +28,69 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
+                showForm();
 
-            // Join the asynchronous test operation so any exceptions are rethrown on this thread
-            test.Join();
+                // Join the asynchronous test operation so any exceptions are rethrown on this thread.
+                test.Join();
+            }
+            finally
+            {
+                form?.Close();
```
sharwell: 📝 This call to `Close()` is unnecessary. It was already closed during the test above.

IntegrationTests/UI.IntegrationTests/UITest.cs
sharwell: If there is a _specific_ type of dialog we need to account for, I would prefer to have a more specific cleanup in place to handle it.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: > These lines do not need to be in the try block.

They really should. They or added logging code could throw an exception, which results in a hang because no one closes the tested form. Hence `Application.Run(form)` and `showForm` never return.

sharwell: > They or added logging code could throw an exception

Exceptions can only be thrown here if:

* There is no main thread
* There is no form of the specific type open
* There is more than one form of the specified type open

For the first two cases, there is no cleanup necessary. For the third case, I would prefer to only take action if it's been a demonstrated problem, and even then I would prefer to assert at the very start of the test that there are no forms of the specified type open.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: Yes, this happens often for FormCommit. But yes, it has no effect on the test behavior. Then the loop always has all three iterations.

sharwell: For the FormCommit case, what form is open that needs to be closed?

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: In https://github.com/gitextensions/gitextensions/pull/7625#discussion_r372054764, you told me to add it.

mstv: I have no concern to removed it.

sharwell: Ah, yes that is better. Main point is we only need one call to `Close()`.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: You're right regarding "the earlier try does not help".
But such an exception is not reported. The test just hangs with an open form.
That's why I tend to add another `try catch(ex) { Console.WriteLine(ex)`.
(When I added debug logging, the inserted catch did not catch a NRE from an early attempt to log something copied & pasted.)

> assert at the very start of the test that there are no forms of the specified type open

👍

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: It is FormCommit itself that does not close.

mstv: Though it has closed when `showForm` returns, i.e. weird but not causing the test to fail.

IntegrationTests/UI.IntegrationTests/UITest.cs
sharwell: Is it possible to reproduce this locally?

mstv: Nothing easier than this 😉
```
           var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
            {
throw new Exception("let it hang");
```

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: done

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: removed

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: removed

IntegrationTests/UI.IntegrationTests/UITest.cs
sharwell: That undermines my claim that those three lines cannot throw a problematic exception.

I found two sources of potential issues during testing (~800 iterations so far). Will submit a small PR shortly.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: Now I remember: I added it in order to avoid the hang caused by an exception before `form = OpenForms...`. Then I noticed that FormCommit trapped here, and I added the `for (int trial...)` loop, which did not change the behavior.
I have moved this up to the new `try catch` for the early exceptions.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: Are you still analyzing, @sharwell? Are there further review comments?
Else this PR could be squash-merged.

sharwell: I'm looking now

sharwell: I'll send a PR to your fork so you can see my full proposed changes.

sharwell: https://github.com/mstv/gitextensions/pull/1

mstv: Merged although I disagree with the error handling strategy: https://github.com/mstv/gitextensions/pull/1#discussion_r380965921, https://github.com/mstv/gitextensions/pull/1#discussion_r380967048.

mstv:
This PR shall be squash-merged. @RussKie?

RussKie:
Whichever way provides a clear diff but hides the "internals" for the lack
of a better word (like 'fixed review', 'address feedback').

PR #7766, opened 2020-02-12
JoinPendingOperationsAsync right after closing the tested Form

mstv:
Related to #7382
Based on #7765

## Proposed changes

- `await JoinPendingOperationsAsync` right after closing the tested `Form`
  in the async task of `UITest.RunForm`, i.e. inside the message loop
  because with #7753 tests hang in `Application.Run(form)` called by `showForm`
- revert `await` in `TearDownAsync` of ROT tests
- standardize opening of FormBrowse and FormCommit in tests

## Test methodology 

- run the NUnit tests

## Test environment(s) 

- AppVeyor
- Windows 10

----

### DIFF
--- a/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
+++ b/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
@@ -59,11 +59,8 @@ namespace GitExtensions.UITests.CommandsDialogs
         }
 
         [TearDown]
-        public async Task TearDownAsync()
+        public void TearDown()
         {
-            // Wait for pending operations so the repository is not deleted while operations run in the background
-            await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);
-
             _referenceRepository.Dispose();
         }
 
@@ -209,10 +206,7 @@ namespace GitExtensions.UITests.CommandsDialogs
         private void RunFormTest(Func<FormBrowse, Task> testDriverAsync)
         {
             UITest.RunForm(
-                () =>
-                {
-                    Assert.True(_commands.StartBrowseDialog(owner: null));
-                },
+                showForm: () => _commands.StartBrowseDialog(owner: null).Should().BeTrue(),
                 testDriverAsync);
         }
     }
--- a/IntegrationTests/UI.IntegrationTests/UITest.cs
+++ b/IntegrationTests/UI.IntegrationTests/UITest.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using CommonTestUtils;
 using GitUI;
 using NUnit.Framework;
 
@@ -57,6 +58,10 @@ namespace GitExtensions.UITests
                         // Close the form after the test completes. This will unblock the 'showForm()' call if it's
                         // waiting for the form to close.
                         form.Close();
+
+                        // This should be changed to assert no pending operations once background operations are tied
+                        // to the life of the owning dialog - issue #7792.
+                        await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);
                     }
                 });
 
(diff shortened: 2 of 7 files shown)


IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +29,80 @@ void HandleApplicationIdle(object sender, EventArgs e)
         }
 
         public static void RunForm<T>(
-            Action showDialog,
-            Func<T, Task> runAsync)
+            Action showForm,
+            Func<T, Task> runTestAsync,
+            bool joinPendingOperationsAfterwards = false)
```
sharwell: I think it's safe to remove this parameter, and just use this:

```csharp
// This should be changed to assert no pending operations once background operations are tied
// to the life of the owning dialog.
await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);
```

IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -27,32 +29,80 @@ void HandleApplicationIdle(object sender, EventArgs e)
...
+                // Start runTestAsync before calling showForm.
+                // The latter might block until the form is closed, especially if using Application.Run(form).
+
+                // Avoid using ThreadHelper.JoinableTaskFactory for the outermost operation because we don't want the task
+                // tracked by its collection. Otherwise, test code would not be able to wait for pending operations to
+                // complete.
+                var test = ThreadHelper.JoinableTaskContext.Factory.RunAsync(async () =>
                 {
-                    dialog.Close();
-                }
```
sharwell: 📝  The `await` operation should occur here (after the closing `}` of the `finally` section).

❗️ This pull request should add one line of code, 1-2 lines of comments, and a blank line to this file. All other changes to this method should be reverted from this pull request and moved to a different pull request.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.LeftPanel.Remotes.cs
```diff
@@ -193,11 +190,10 @@ private void RunRepoObjectsTreeTest(Action<TreeNode> testDriver)
...
             UITest.RunForm(
-                () =>
-                {
-                    Assert.True(_commands.StartBrowseDialog(owner: null));
-                },
-                testDriverAsync);
+                showForm: () => _commands.StartBrowseDialog(owner: null).Should().BeTrue(),
+                testDriverAsync,
+                joinPendingOperationsAfterwards: true // so the repository is not deleted while operations run in the background
+            );
```
sharwell: 📝 All changes other than the change to `TearDown` methods should be moved to a different pull request.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: > Based on #7765

Please start with #7765.

IntegrationTests/UI.IntegrationTests/UITest.cs
mstv: (Last time, you told me to review and to adapt all tests which fail.)

Or else how should FormBrowse and FormCommit join their pending operations?
`Dispose` and `OnFormClose` are not `async` and should not use `AsyncTestHelper.JoinPendingOperationsAsync`.

IntegrationTests/UI.IntegrationTests/UITest.cs
```diff
@@ -56,7 +57,16 @@ void HandleApplicationIdle(object sender, EventArgs e)
                     {
                         // Close the form after the test completes. This will unblock the 'showForm()' call if it's
                         // waiting for the form to close.
-                        form.Close();
+                        try
```
sharwell: ❕ The try/finally block here is unnecessary. The `JoinPendingOperationsAsync` only exists to ensure successful cleanup after successful close. If the close throws an exception, the test has already failed and we're still going to join the operations here:
https://github.com/gitextensions/gitextensions/blob/147eb46ff0f0caa5d80ef46c772b0bf3f7ba5605/UnitTests/CommonTestUtils/ConfigureJoinableTaskFactoryAttribute.cs#L74

mstv: You have convinced me. If the close fails, the test will hang most likely anyway. Then it does not matter whether the async operations are tried to join during `Application.Run(form)` or afterwards.

Issue #7743, opened 2020-02-08
Flaky test RepoObjectTree_should_order_active_remotes_alphabetically


mstv:
## Current behaviour

The test `GitExtensions.UITests.CommandsDialogs.FormBrowseLeftPanelRemotesTests.RepoObjectTree_should_order_active_remotes_alphabetically` fails occasionally:
"Expected names to be a collection with 5 item(s), but found an empty collection."

Also `RepoObjectTree_should_load_active_remotes`:
"Expected remotesNode.Nodes.Count to be 5, but found 0."

Several attempts to wait and to allow missing processing did not help:
```
                    List<string> names;
                    int delayCount = 0;
                    while ((names = remotesNode.Nodes.OfType<TreeNode>().Select(x => x.Text).ToList()).Count != RemoteNames.Length && delayCount < 100)
                    {
                        ++delayCount;
                        Thread.Sleep(10);
                        Application.DoEvents();
                        await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);
                        await Task.Yield();
                        await UITest.WaitForIdleAsync();
                        Application.DoEvents();
                    }

                    if (delayCount > 0)
                    {
                        UITest.Log($"{nameof(RepoObjectTree_should_order_active_remotes_alphabetically)} delayed {delayCount} times");
                    }

                    // assert
                    names.Should().BeEquivalentTo(RemoteNames);
                    names.Should().BeInAscendingOrder();
```

## Expected behaviour

test passes stably

## Steps to reproduce

run the tests on AppVeyor several times, e.g. https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30662119

## Environment

- GE d5a4408
- AppVeyor
- Windows 10

RussKie:
Is this fixed? Or does it actually need fixing?

gerhardol:
> Is this fixed? Or does it actually need fixing?

Only seem to be a problem in the test. The test is still relevant.
#8119 fixes a problem in this test that was added after this issue was opened (Appveyor environment).
I got some way on investigating this, no conclusion

PR #7775, opened 2020-02-15
Fix incomplete reset in ReferenceRepository

sharwell:
`ReferenceRepository.Reset` is supposed to efficiently reset the repository to a fresh state. This pull request fixes two cases where state was not reset.


mstv:
> check whether this solves #7743.

It does not. The `FormBrowseLeftPanelRemotesTests` fail yet.

PR #7779, opened 2020-02-16
SidePanel Submodules: Status for top module

gerhardol:
## Proposed changes 

Show status for top module in the side panel.
If there are changes in any submodules (commit or dirty) or changes in the topmodule itself, this status should be seen in the top most icon.
(The top module cannot be ahead/behind only dirty but will be so for any changes)

Minor refactorings and removal of unused code too, to be squashed.

### Limitations
There are some limitations to the implementation, due to performance issues (git-status can require some resources).

git-status reuses the call for the commit button and artificial commits in the grid, so it is executed for the current module only. If there are changes, the side panel status is currently refreshed only for the current submodule (and below).
At the initial opening, the status is calculated for all modules though, except files in the current top module.
There is also a limiter to not refresh the status too often (ignore updates within 15s).

If the top module is current, the git-status updates are sufficient to set the status for the top module sufficient. (When rewriting this section about the limitations, I updated the implementation too.)

If another submodule is current, git-status cannot be used to refresh all status. (Previously no update at all was done.) If a submodule has changes, also the top module is dirty. However, it is not possible to determine if the status for the top module (and all modules in between) can be cleared as the modules and files in the top module is not calculated. The status is therefore is therefore additive, i.e. 'dirty' is set but not cleared.

Summary limitations if top module is not current:
 * If only files in top module is changed, top is not set as dirty (not even initially)
 * Status for top module is changed to dirty, not cleared if there are changes
 * Changes in modules outside the tree is not monitored.
 * If the current submodule is level 3 or lower, the upper modules are not changed to dirty if current module is changed (this could be changed though).

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Submodules/SubmoduleStatusProvider.cs
+++ b/GitCommands/Submodules/SubmoduleStatusProvider.cs
@@ -108,6 +108,11 @@ namespace GitCommands.Submodules
                     _submoduleInfos[info.Path] = info;
                 }
 
+                if (!_submoduleInfos.ContainsKey(result.TopProject.Path))
+                {
+                    _submoduleInfos.Add(result.TopProject.Path, result.TopProject);
+                }
+
                 // Start update status for the submodules
                 if (updateStatus)
                 {
@@ -119,7 +124,7 @@ namespace GitCommands.Submodules
 
                     if (_gitStatusWhileUpdatingStructure != null)
                     {
-                        // Current module must be updated separetly (not in _submoduleInfos)
+                        // Current module must be updated separately (not in _submoduleInfos)
                         await UpdateSubmodulesStatusAsync(currentModule, _gitStatusWhileUpdatingStructure, cancelToken);
                     }
 
@@ -230,7 +235,7 @@ namespace GitCommands.Submodules
             {
                 var localPath = superprojectModule.WorkingDir.Substring(topProject.WorkingDir.Length);
                 name = Path.GetDirectoryName(localPath).ToPosixPath();
-             }
+            }
 
             string path = superprojectModule.WorkingDir;
             name += GetBranchNameSuffix(path, noBranchText);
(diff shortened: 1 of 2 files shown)


GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -326,6 +338,24 @@ private async Task UpdateSubmodulesStatusAsync(GitModule module, [CanBeNull] IRe
             }
         }
 
+        /// <summary>
+        /// Set the top module as dirty
+        /// If any module is changed
+        /// </summary>
+        /// <param name="module">current Git module</param>
+        private void SetTopModuleAsDirty(GitModule module)
```
RussKie:
```suggestion
        private void SetTopModuleAsDirty(string submoduleWorkingDir)
```

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -381,6 +411,11 @@ private async Task GetSubmoduleDetailedStatusAsync(GitModule superModule, string
                     AddedAndRemovedText = submoduleStatus.AddedAndRemovedString()
                 };
 
+            if (submoduleStatus != null)
+            {
+                SetTopModuleAsDirty(superModule);
```
RussKie:
```suggestion
                SetTopModuleAsDirty(superModule.GetTopModule().WorkingDir);
```

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -312,6 +317,13 @@ private async Task UpdateSubmodulesStatusAsync(GitModule module, [CanBeNull] IRe
             _previousSubmoduleUpdateTime = DateTime.Now;
             await TaskScheduler.Default;
 
+            // TopModule is dirty if there are any changes in any module
+            if (gitStatus != null
+                && gitStatus.Count > 0)
+            {
+                SetTopModuleAsDirty(module);
```
RussKie:
```suggestion
                SetTopModuleAsDirty(module.GetTopModule().WorkingDir);
```

GitCommands/Submodules/SubmoduleStatusProvider.cs
gerhardol: changing

gerhardol:
Force pushed RussKies suggestion, updating description

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -326,6 +338,24 @@ private async Task UpdateSubmodulesStatusAsync(GitModule module, [CanBeNull] IRe
...
         }
 
+        /// <summary>
+        /// Set the top module as dirty
+        /// If any module is changed
+        /// </summary>
+        /// <param name="topModuleWorkingDir">path to top module</param>
+        private void SetTopModuleAsDirty(string topModuleWorkingDir)
+        {
+            if (_submoduleInfos[topModuleWorkingDir].Detailed == null)
```
RussKie: I missed this on the first review. Are you sure `_submoduleInfos[topModuleWorkingDir]` is safe?

gerhardol: Yes, added on line 111

gerhardol:
When updating the description about limitations, I realized that one of them could be eliminated so I updated slightly.

gerhardol:
@msftbot merge in 2 days

gerhardol:
@msftbot merge in 20 hours

PR #7785, opened 2020-02-17
Publish portable build

RussKie:
## Proposed changes

- Effectively revert a606467bb23721313b211b004b3456ce30572969, and publish a portable build

NB: the plugins are not included in the portable build right now

----

### DIFF
--- a/GitExtensions/Project.Publish.targets
+++ b/GitExtensions/Project.Publish.targets
@@ -1,0 +1,88 @@
+<Project>
+
+  <!--
+    ============================================================
+                       _PublishTranslations
+
+    Copies all available translations to the publish directory.
+    ============================================================
+    -->
+  <Target Name="_PublishTranslations">
+    <Copy SourceFiles="@(Translations)" DestinationFolder="$(PublishDir)Translation" ContinueOnError="false" />
+  </Target>
+
+  <!--
+    ============================================================
+                       _PublishExtraDependecies
+
+    Copies additional files (e.g. EasyHook dlls) to the publish directory.
+    ============================================================
+    -->
+  <Target Name="_PublishExtraDependecies">
+    <Copy SourceFiles="$(SolutionDir)\Externals\EasyHook\EasyHookDll\Build\Release\x64\EasyHook64.dll" DestinationFolder="$(PublishDir)" />
+    <Copy SourceFiles="$(SolutionDir)\Externals\EasyHook\EasyHookDll\Build\Release\x86\EasyHook32.dll" DestinationFolder="$(PublishDir)" />
+  </Target>
+
+  <!--
+    ============================================================
+                       _RetrieveCurrentBuildVersion
+
+    Retrieves the current build version.
+    ============================================================
+    -->
+  <Target Name="_RetrieveCurrentBuildVersion">
+    <PropertyGroup>
+      <ExeAssembly>$([MSBuild]::NormalizePath('$(PublishDir)', $(_DeploymentApplicationManifestIdentity)))</ExeAssembly>
+    </PropertyGroup>
+
+    <GetAssemblyIdentity AssemblyFiles="$(ExeAssembly)">
+      <Output TaskParameter="Assemblies" ItemName="ExeAssemblyIdentities"/>
+    </GetAssemblyIdentity>
+
+    <PropertyGroup>
+      <CurrentBuildVersion>%(ExeAssemblyIdentities.Version)</CurrentBuildVersion>
+    </PropertyGroup>    
+  </Target>
+
+  <!--
+    ============================================================
+                       CreatePortable
+
+    Creates a portable archive.
+    ============================================================
+    -->
+
+  <Target Name="CreatePortable" AfterTargets="Publish" DependsOnTargets="_RetrieveCurrentBuildVersion;_PublishTranslations;_PublishExtraDependecies">
+    <PropertyGroup>
+      <!-- Resolve app.config, so we can set/unset "portable" flag -->
+      <_PublishAppConfig>$([System.IO.Path]::GetFileName('$(TargetAppConfig)'))</_PublishAppConfig>
+      <_PublishAppConfigPath>$([System.IO.Path]::Combine('$(PublishDir)', '$(_PublishAppConfig)'))</_PublishAppConfigPath>
+
+      <!-- Resolve the output file -->
+      <_PublishPortableVersionSuffix>-$(CurrentBuildVersion.ToString())</_PublishPortableVersionSuffix>
+      <_PublishPortableCommitHashSuffix Condition="'$(GitCommit)' != ''">-$(GitCommit)</_PublishPortableCommitHashSuffix>
+      <_PublishPortableCommitHashSuffix Condition="'$(env:APPVEYOR_REPO_COMMIT)' != ''">-$(env:APPVEYOR_REPO_COMMIT)</_PublishPortableCommitHashSuffix>
+      <_PublishPortableFileName>GitExtensions-portable$(_PublishPortableVersionSuffix)$(_PublishPortableCommitHashSuffix).zip</_PublishPortableFileName>
+      <_PublishPortablePath>$([MSBuild]::NormalizePath('$(ArtifactsPublishDir)', '$(_PublishPortableFileName)'))</_PublishPortablePath>
+
+      <!-- We want to archive the whole publish folder, so get one level up -->
+      <_PublishedPath>$([MSBuild]::NormalizeDirectory('$(PublishDir)', '..'))</_PublishedPath>
+    </PropertyGroup>
+
+    <!-- Mark the package as "portable" -->
+    <XmlPoke XmlInputPath="$(_PublishAppConfigPath)"
+            Query="configuration/applicationSettings/GitCommands.Properties.Settings/setting[@name='IsPortable']/value" 
+            Value="True" />
+
+    <ZipDirectory
+            SourceDirectory="$(_PublishedPath)"
+            DestinationFile="$(_PublishPortablePath)"
+            Overwrite="true" />
+
+    <!-- Reset the "portable" flag -->
+    <XmlPoke XmlInputPath="$(_PublishAppConfigPath)"
+            Query="configuration/applicationSettings/GitCommands.Properties.Settings/setting[@name='IsPortable']/value" 
+            Value="False" />
+  </Target>
+
+</Project>
--- a/GitUI/Theming/ThemeRepository.cs
+++ b/GitUI/Theming/ThemeRepository.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using GitCommands;
 using GitExtUtils.GitUI.Theming;
+using JetBrains.Annotations;
 
 namespace GitUI.Theming
 {
@@ -19,12 +20,21 @@ namespace GitUI.Theming
             _persistence = persistence ?? throw new ArgumentNullException(nameof(persistence));
             string appDirectory = AppSettings.GetGitExtensionsDirectory() ??
                 throw new InvalidOperationException("Missing application directory");
-            AppDirectory = Path.Combine(appDirectory, Subdirectory);
-            DataDirectory = Path.Combine(AppSettings.ApplicationDataPath.Value, Subdirectory);
+            AppThemesDirectory = Path.Combine(appDirectory, Subdirectory);
+
+            string userDirectory = AppSettings.ApplicationDataPath.Value;
+
+            // in portable version appDirectory and userDirectory are same,
+            // hence we don't have a separate directory for user themes
+            UserThemesDirectory = string.Equals(appDirectory, userDirectory, StringComparison.OrdinalIgnoreCase)
+                ? null
+                : Path.Combine(userDirectory, Subdirectory);
         }
 
-        private string AppDirectory { get; }
-        private string DataDirectory { get; }
+        private string AppThemesDirectory { get; }
+
+        [CanBeNull]
+        private string UserThemesDirectory { get; }
         public string InvariantThemeName { get; } = "win10default";
 
         public Theme GetTheme(ThemeId id)
@@ -43,7 +53,7 @@ namespace GitUI.Theming
             GetBuiltinThemeIds().Concat(GetUserCustomizedThemeIds());
 
         private IEnumerable<ThemeId> GetBuiltinThemeIds() =>
-            new DirectoryInfo(AppDirectory)
+            new DirectoryInfo(AppThemesDirectory)
                 .EnumerateFiles("*" + Extension, SearchOption.TopDirectoryOnly)
                 .Select(_ => Path.GetFileNameWithoutExtension(_.Name))
                 .Where(name => !name.Equals(InvariantThemeName, StringComparison.OrdinalIgnoreCase))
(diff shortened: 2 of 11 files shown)


RussKie:
A portable build now again - https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30847034/artifacts

Attempting to start an app I get the following error:
image

The error is not fatal, it complains about absent `Theme\current.colors` file, which gets created as part of the launch sequence.

@NikolayXHD any thoughts what needs to be done to improve the UX?

NikolayXHD:
> @NikolayXHD any thoughts what needs to be done to improve the UX?

#7562 gets rid of `current.colors` and the popups

NikolayXHD:
Found 2 issues

1. I broke dark theme for portable version. Never tested color settings migration with portable build and **!!!surprise!!!** I broke dark theme completely. At least the missing file popup is gone, as I promised.

2. `EasyHook32.dll`, `EasyHook64.dll` are missing in portable .zip
`MakePortableArchive.cmd` still points to correct files, so currently I have no explanation for this.

Will be looking at it closely the coming weekend.

RussKie:
Thank you for testing, more work to do.

RussKie:
> 2\. I have no explanation for this.

I'm running a "publish" target against GitExtensions.csproj, that only copies files that are marked as "referenced" by the project. EasyHook dlls aren't technically referenced, and hence they don't get copied. That's my hypothesis.

RussKie:
I tweaked the publish target to include EasyHook dlls. I further tweaked it to include a build version in the name of the zip. And, while at it, I rebased on top of the latest master.
Please let me know if anything else I missed or is broken.

> The old portable builds had an additional directory inside

fixed



> 2\. `EasyHook32.dll`, `EasyHook64.dll` are missing in portable .zip

fixed

RussKie:
🎉 [https://ci.appveyor.com/api/buildjobs/gd3jsouxlbhsfxq1/artifacts/artifacts%2Fpublish%2FPortable-4.0.0.8598.zip](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/30907552/artifacts)

GitExtensions/Project.Publish.targets
```diff
@@ -0,0 +1,84 @@
...
+    -->
+
+  <Target Name="CreatePortable" AfterTargets="Publish" DependsOnTargets="_RetrieveCurrentBuildVersion;_PublishTranslations;_PublishExtraDependecies">
+    <PropertyGroup>
+      <!-- Resolve app.config, so we can set/unset "portable" flag -->
+      <_PublishAppConfig>$([System.IO.Path]::GetFileName('$(TargetAppConfig)'))</_PublishAppConfig>
+      <_PublishAppConfigPath>$([System.IO.Path]::Combine('$(PublishDir)', '$(_PublishAppConfig)'))</_PublishAppConfigPath>
+
+      <!-- Resolve the output file -->
+      <_PublishPortableFileName>Portable-$(CurrentBuildVersion.ToString()).zip</_PublishPortableFileName>
```
pmiossec: Is it possible to add 'GitExt' or 'GitExtensions' at the beginning of the zip file.
I think that will make it more identifiable in a full download folder ;)

And also **if it's really straight forward** to do, add the hash at the end by following the SemVer syntax (i.e  `GitExtensions-Portable-4.0.0.8598+f0d22742.zip`)

gerhardol:
Build is dirty

Still no dark theme...
No theme to select
I tried to copy my current (unmodified but not master) dark.colors, then I get an exception opening Settings

```
System.InvalidOperationException: Themes were not added to ComboBox
   at GitUI.CommandsDialogs.SettingsDialog.Pages.ColorsSettingsPage.set_SelectedThemeId(ThemeId value)
   at GitUI.CommandsDialogs.SettingsDialog.Pages.ColorsSettingsPage.SettingsToPage()
   at GitUI.CommandsDialogs.SettingsDialog.SettingsPageBase.LoadSettings()
   at GitUI.CommandsDialogs.FormSettings.LoadSettings()
   at GitUI.CommandsDialogs.FormSettings.FormSettings_Shown(Object sender, EventArgs e)
   at System.Windows.Forms.Form.OnShown(EventArgs e)
   at System.Windows.Forms.Form.CallShownEvent()
   at System.Windows.Forms.Control.InvokeMarshaledCallbackDo(ThreadMethodEntry tme)
   at System.Windows.Forms.Control.InvokeMarshaledCallbackHelper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Forms.Control.InvokeMarshaledCallback(ThreadMethodEntry tme)
   at System.Windows.Forms.Control.InvokeMarshaledCallbacks()
```


## Environment
- Git Extensions 4.0.0.8598
- Build 8896d5910e3933caf13b8c5eb6db8ac972f52641 (Dirty)
- Git 2.24.0.windows.2
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

GitExtensions/Project.Publish.targets
RussKie: Yes, it is easy to add a prefix:
```xml
<_PublishPortableFileName>GitExtensions-Portable-$(CurrentBuildVersion.ToString()).zip</_PublishPortableFileName>
```

A hash may be a (little?) more involved, but doable.

vbjay:
You can use something like this

https://github.com/gitextensions/gitextensions/blob/master/Setup/Get-Hashes.ps1

RussKie:
I think @pmiossec was referring to a commit hash.
Which we have, its retrieval just needs to be plumb in. I think we should
also bake the hash in to assembly version (one its attributes)

RussKie:
> Build is dirty

Correct, we're not building a release (if that's what you mean)

> Still no dark theme...
> No theme to select

@NikolayXHD would you be able to check this? Likely I have missed something... 😕

gerhardol:
>> Build is dirty

> Correct, we're not building a release (if that's what you mean)

No server build should be git dirty

NikolayXHD:
> @NikolayXHD would you be able to check this? Likely I have missed something... 😕

Yes, I will. It's a mistake somewhere in my last PR. Will figure it out more or less by the weekend.

gerhardol:
Works for me
I would have preferred to be able to edit and save themes also in portable, even if it means that you can modify the built in themes. Certainly not a blocker.

NikolayXHD:
> I would have preferred to be able to edit and save themes also in portable, even if it means that you can modify the built in themes. Certainly not a blocker.

Theme editing is gone at the moment, it's not portable version specific. Had to sacrifice it to keep code quality decent at the moment of release.

Theme files are just plain text with hex color codes, so in principle it can be edited in text editor with coupled with some online palette tool, such as color.adobe.com

GitUI/Theming/ThemeRepository.cs
```diff
@@ -71,9 +86,19 @@ private IEnumerable<ThemeId> GetUserCustomizedThemeIds()
...
+        private string GetPath(ThemeId id)
+        {
+            if (id.IsBuiltin)
+            {
+                return Path.Combine(AppThemesDirectory, id.Name + Extension);
+            }
+
+            if (UserThemesDirectory == null)
+            {
+                throw new NotSupportedException("There is no directory for custom user themes in portable mode");
```
RussKie:
```suggestion
                throw new InvalidOperationException("There is no directory for custom user themes in portable mode");
```
https://docs.microsoft.com/dotnet/api/system.notsupportedexception#related-exception-types

GitUI/Theming/ThemeRepository.cs
```diff
@@ -71,9 +86,19 @@ private IEnumerable<ThemeId> GetUserCustomizedThemeIds()
                 : Enumerable.Empty<ThemeId>();
         }
 
-        private string GetPath(ThemeId id) =>
-            Path.Combine(id.IsBuiltin
-                ? AppDirectory
-                : DataDirectory, id.Name + Extension);
+        private string GetPath(ThemeId id)
```
RussKie: Is it technically possible to have a theme in a portable build that isn't a 'built-in'?

RussKie:
> I would have preferred to be able to edit and save themes also in portable

The editor added a lot of code that likely won't be used by the majority of our users. With that it brought a lot of maintenance overhead for us.
We can probably release it as a plugin.

GitExtensions/Project.Publish.targets
RussKie: Done. 
For non-release builds I propose we just have a commit hash instead of a build number, e.g. **GitExtensions-portable-f511a73a7.zip**. For a release we'll have a build number, e.g. **GitExtensions-portable-3.3.0.2342.zip**.
What do you think?

RussKie:
> No server build should be git dirty

Respectfully disagree. We only made clean builds for release.
[Here's a PR build](https://ci.appveyor.com/project/gitextensions/gitextensions/builds/29905385/artifacts) from few months ago:
```
- Git Extensions 4.0.0.7889
- Build 364baef68efb2ed8f67320d16544056e4dcc6c0f (Dirty)
- Git 2.16.1.windows.4 (recommended: 2.23.0 or later)
- .NET Framework 4.8.4075.0
- DPI 144dpi (150% scaling)
```

I am yet to look at the "merge" or "release" pipeline. Let's deal with the "pr" pipeline first.

GitExtensions/Project.Publish.targets
pmiossec: I think that's good for now.

It will be perfect for 'master' builds and releases. No so much for PR (I thought about it afterwards...) 

I will try to look **later** at how to  retrieve the hash of the head of the pull request and use it as the hash here.

Because Appveyor build a commit which is a merge of the PR within 'master' so the hash will be of no use in this case 😕

pmiossec: https://www.appveyor.com/docs/environment-variables/

* APPVEYOR_PULL_REQUEST_HEAD_COMMIT - Pull (Merge) Request source commit ID (SHA)

Or more likely :

* APPVEYOR_REPO_COMMIT - commit ID (SHA)

I think that we should add something like that to override the value on Appveyor (in pseudo msbuild on smartphone 😁) :

`<_PublishPortableFileNameSuffix Condition="'$(GitCommit)' != '' And '$(env:APPVEYOR_REPO_COMMIT)' !='' ">-$(env:APPVEYOR_REPO_COMMIT)</_PublishPortableFileNameSuffix>`

GitUI/Theming/ThemeRepository.cs
gerhardol: Yes, you can still copy an existing theme and adapt that.
Name is set from the filename.

gerhardol:
>> No server build should be git dirty

> Respectfully disagree. We only made clean builds for release.
I am yet to look at the "release" pipeline. Let's deal with the "pr" pipeline first.

Future problem, agree not in this PR

GitExtensions/Project.Publish.targets
mstv: > For non-release builds I propose we just have a commit hash instead of a build number, e.g. GitExtensions-portable-f511a73a7.zip. For a release we'll have a build number, e.g. GitExtensions-portable-3.3.0.2342.zip.

Build numbers have the advantage that they can be used for a kind of bisecting without building the project. I'd prefer to have it with build number and with hash appended.

> dirty

I agree only if AppVeyor builds a temporary merge with master.
But I don't like this for two reasons:
1. AppVeyor denies to build if there are conflicts with master. So you cannot finish the feature without being forced to merge the master branch.
2. The resulting build is a moving target depending on the master branch.
I'd prefer that AppVeyor builds exactly the _clean_ commit for PRs.

RussKie: > Build numbers have the advantage that they can be used for a kind of bisecting without building the project. I'd prefer to have it with build number and with hash appended.

Apologies, I don't follow the logic here. How is a build number useful for bisecting, e.g. 4.0.0.1234? I can see a hash being useful though.



> I'd prefer that AppVeyor builds exactly the _clean_ commit for PRs.

The current function of "dirty" flag is to indicate whether a build is an official release or not. It is useful when users send us bug reports.
If we mark all builds as "clean" we won't know whether a specific build is official or not _without going to the GitHub release tab and checking there_.

> 1. AppVeyor denies to build if there are conflicts with master

If there is a conflict, it has to be resolved one way or another. I don't follow how it relates to the "dirty" flag discussion, sorry.

GitUI/Theming/ThemeRepository.cs
RussKie: In that case, won't we be falling through to `UserThemesDirectory == null` code path?

GitExtensions/Project.Publish.targets
mstv: > bisecting

Usually, I have many downloaded portable builds. They are well-ordered by build numbers.
Checking different builds for an issue with commits hashs only, would be a pain.

> dirty

My understanding of "dirty" is "There are uncommitted changes."
If I really forgot the build number of a release, I will see it in my folder with the portable builds (last one).

> If there is a conflict, it has to be resolved one way or another.

Yes, but we should allow to finish the implementation and being able to check it by AppVeyor before it is merged with master.

> I don't follow how it relates to the "dirty" flag discussion, sorry.

If AppVeyor builds a temporary merge with master, this _is_ a "dirty" build because it is no official commit.

RussKie: > this _is_ a "dirty" build because it is no official commit.

You mis-understand, 'clean' builds only come from release branches. All other commits are marked dirty. This is how we had it since we introduced GitInfo into our builds.

A 'dirty' build means there are uncommitted changes in the working directory. We inject build numbers into AssemblyInfo and C++ header files, and that makes builds 'dirty'. For release builds (non-pr commits on 'release' branches) we ignore these changes making builds 'clean'.


> Usually, I have many downloaded portable builds. They are well-ordered by build numbers.
> Checking different builds for an issue with commits hashs only, would be a pain.

Make sense.

GitUI/Theming/ThemeRepository.cs
gerhardol: You would copy the theme to the app data directory, user data directory does not exist (really the same as app data).

GitExtensions/Project.Publish.targets
mstv: I don't mis-understand this IMO. I just had not in mind, that a non-release build is modified in this way and hence is really a "dirty" build from git point of view. Though IMO this is just a technical issue how we inject build numbers. Whether this makes a commit build dirty or not, is a non-productive philosophical question. I can live with "dirty" non-release builds.

pmiossec: > I agree only if AppVeyor builds a temporary merge with master.
>[...] 
>I'd prefer that AppVeyor builds exactly the clean commit for PRs.

I think that is GitHub that trigger Appveyor builds only if there are no conflicts.
And also GitHub trigger it 2 times for each PR. One is to build only the PR. The other is for the PRs merged with 'master' (to do real Continuous Integration).

That makes me think of something important. What is the build result that we consider to be the 'dev' version and that we should propose to download and test.

I think that it's better to use the 'merged' one.

Perhaps also we should do something to differentiate the artifacts because they will have the same name 🤔

RussKie: > And also GitHub trigger it 2 times for each PR. One is to build only the PR. The other is for the PRs merged with 'master' (to do real Continuous Integration).

This only happens for commits on the main trunk. It doesn't happen for PRs coming from forks.

RussKie:
@msftbot merge in 24 hours

GitExtensions/Project.Publish.targets
pmiossec: Thanks for the clarification.

PR #7796, opened 2020-02-22
ESCAPE_COMMIT_MSG Escaping Double Quotes In Merge Commit Message If It Exists

### LINKED ISSUE #3954: Commit message field in merge dialog does not escape commit message properly
Bug

Merge with message `Test "messsage text" test` will fail with 

```
"C:\Program Files\Git\bin\git.exe" merge --no-ff -m "test "message text" test" origin/mh-service-label
merge: text test - not something we can merge
```

Fill in a merge commit message that has two words in quotes surrounded by other words. 

Commit message `test "message text" test`

- GitExtensions version: 2.50.01
- GIT version: 2.14.1.windows.1, 
- OS version: Win 10

No

hrai:
Fixes #3954

### Before
Merge commit message with quotes was failing the merge `test "message text" test`
image

### After
The merge succeeds.
image
image

## Test methodology 

- Added NUnit test

## Test environment(s) 
- GIT version 2.22.0.windows.1
- Windows 10

----


GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -1024,6 +1026,11 @@ public static ArgumentString MergeBranchCmd(string branch, bool allowFastForward
             };
         }
 
+        private static string EscapeDoubleQuotes(string message)
```
mstv: Should rather be named `EscapeQuotes`.
Code block can be omitted, i.e. turn into `=> message?.Replace("\"", "\\\"");`

hrai:
@RussKie `GetGitDirectory()` needs RepositoryPath.

Is there a method or a property that I can use to get this value?

RussKie:
If you have an instance of a `GitModule`, you can get its repo path from
`Working Dir`.

hrai:
@RussKie now the merge happens through MERGE_MSG file. I had to shuffle some methods around. Also updated the tests.

GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -1081,5 +1075,40 @@ public static void TerminateTree(this Process process)
                 process.Kill();
             }
         }
+
+        public static string FormatCommitMessageFromTextBox(
+            string commitMessageText, bool usingCommitTemplate, bool ensureCommitMessageSecondLineEmpty)
+        {
+            if (commitMessageText == null)
```
RussKie: What about `commitMessageText == ""` or `commitMessageText == "    "`?

GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -1081,5 +1075,40 @@ public static void TerminateTree(this Process process)
...
+            {
+                return string.Empty;
+            }
+
+            var formattedCommitMessage = new StringBuilder();
+
+            var lineNumber = 1;
+            foreach (var line in commitMessageText.Split('\n'))
+            {
+                string nonNullLine = line ?? string.Empty;
```
RussKie: How can `line == null`?

GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -1007,30 +1008,23 @@ private static GitItemStatus GitItemStatusFromStatusCharacter(StagedStatus stage
             };
         }
 
-        public static ArgumentString MergeBranchCmd(string branch, bool allowFastForward, bool squash, bool noCommit, string strategy, bool allowUnrelatedHistories, string message, int? log)
+        public static ArgumentString MergeBranchCmd(string branch, bool allowFastForward, bool squash, bool noCommit, string strategy, bool allowUnrelatedHistories, string folderPath, int? log)
```
RussKie: Please add tests for this method that at least cover the `-F` flag

GitUI/CommandsDialogs/FormMergeBranch.cs
```diff
@@ -102,6 +109,21 @@ private void OkClick(object sender, EventArgs e)
             }
         }
 
+        private void WriteMessageToMergeFile(string mergeMessage)
```
RussKie: Please expose it via a `TestAccessor` and add unit tests using `ReferenceRepository`

GitCommands/Git/GitCommandHelpers.cs
hrai: This is a pre-existing code. If replacing the above with `IsNullOrWhitespace()` is not going to affect the existing logic, I can push the change.

GitCommands/Git/GitCommandHelpers.cs
hrai: This is a preexisting method that I pulled from `GitUI/CommandsDialogs/FormCommit.cs`

GitCommands/Git/GitCommandHelpers.cs
hrai: All the test cases now accept a path instead of a message.

The response has `-F` flag.

https://github.com/gitextensions/gitextensions/blob/5935979af57db41db5a4e128e2d058422fb9bb57/UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs#L747

GitUI/CommandsDialogs/FormMergeBranch.cs
hrai: This is the method in FormCommit.cs that I used to write this method.

image

This method doesn't seem to have any tests. Neither is any of the invoking code covered by tests via the `TestAccessor`.

I'm afraid I don't have much idea about this test that you're proposing.

PR #7797, opened 2020-02-22
Convert theme file to css palette

NikolayXHD:
## Proposed changes

convert theme files to css palette, so they can be edited with ease directly in IDE

## Screenshots 

to display css colors inline in VS2019, [inline color picker plugin](https://marketplace.visualstudio.com/items?itemName=NikolaMSFT.InlineColorPicker) can be used

theme_as_css_palette

In JetBrains Rider css colors are visible out-of-the-box

theme_as_css_palette_rider

----


pmiossec:
@NikolayXHD Correct me if I'm wrong but I would have expected to see changes on the code that read that theme file.

Did you miss to commit a file?

GitUI/GitUI.csproj
```diff
@@ -75,7 +75,7 @@
     <None Include="gituikey.snk.pfx" />
     <None Include="Resources\ChangeLog.md" />
 
-    <Content Include="Themes\dark.colors">
```
gerhardol: win10default.colors is not converted, as well as .css are not read
But as a principle OK, a standard format to edit the files instead of the a custom, especially since the editor was excluded.

RussKie:
This is a very cool idea 👍

NikolayXHD:
> @NikolayXHD Correct me if I'm wrong but I would have expected to see changes on the code that read that theme file.
> 
> Did you miss to commit a file?

> win10default.colors is not converted, as well as .css are not read
> But as a principle OK, a standard format to edit the files instead of the a custom, especially since the > editor was excluded.

I intended to get principal approval before doing the rest, that's why I marked PR with label `in progress` :)

Now I am going to proceed.

gerhardol:
Is ThemeMigration needed at all?
Those with customized  themes before 4.0 will have to redo the changes

NikolayXHD:
> Those with customized themes before 4.0 will have to redo the changes

It's for people who changed GitExtensions colors in `ColorSettingsPage` before theming came into existence.

Issue #7338, opened 2019-10-20
Stage/unstage select lines in Browse


gerhardol:
## Feature description

GE supports staging/unstaging files from the Browse window. The commit modal dialog is still required to stage/unstage lines (as well as the actual commit, that is a separate feature). It is possible to Cherry-pick/revert lines, but that often fails (git-apply has different options).

FormCommit has additions to FileViewer compared to RevisionDiffControl.  These options should be moved to FileViewer.cs so they can be shared with RevisionDiffControl.cs for the artificial commits.
A difference is that the "stage mode" is static in FormCommit but will depend on the selectedrevisions in RevisionDiffControl

## Environment
- Git Extensions 3.3.0
- Build 395cf7af41c8bdbfe3d4ef421608623e23acf1e7
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

PR #7798, opened 2020-02-22
RevisionDiff: Try retain the file selection when switching commits

gerhardol:
Part of #7338 

## Proposed changes

RevisionDiff had some functionality to retain the selected file that did not really work.

This change try to retain the file selected, also when changing the selected commit. This makes it easier to follow a file from commit to commit. A more important use case is for handling artificial commits. If refreshing after staging lines, the same file should be kept (not select the first file).

A second change is to retain the index when staging/unstaging files (or staging all lines) for Artificial commits, rather than the first is selected again.

In some situations, no file was selected in RevisionDiff. This should make sure that a file is always selected. Not reproduced though.

## Test methodology

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -33,8 +33,6 @@ namespace GitUI.CommandsDialogs
 
         private RevisionGridControl _revisionGrid;
         private RevisionFileTreeControl _revisionFileTree;
-        private string _oldRevision;
-        private GitItemStatus _oldDiffItem;
         private IRevisionDiffController _revisionDiffController;
         private readonly IFileStatusListContextMenuController _revisionDiffContextMenuController;
         private readonly IFullPathResolver _fullPathResolver;
@@ -53,33 +51,24 @@ namespace GitUI.CommandsDialogs
             _revisionDiffContextMenuController = new FileStatusListContextMenuController();
         }
 
-        public void ForceRefreshRevisions()
-        {
-            var revisions = _revisionGrid.GetSelectedRevisions();
-            if (revisions.Count != 0)
-            {
-                _oldRevision = revisions[0].Guid;
-                _oldDiffItem = DiffFiles.SelectedItem;
-            }
-            else
-            {
-                _oldRevision = null;
-                _oldDiffItem = null;
-            }
-
-            RefreshArtificial();
-        }
-
         public void RefreshArtificial()
         {
-            if (Visible)
+            if (!Visible)
             {
-                var revisions = _revisionGrid.GetSelectedRevisions();
+                return;
+            }
 
-                if (revisions.Count > 0 && revisions[0].IsArtificial)
-                {
-                    DiffFiles.SetDiffs(revisions, _revisionGrid.GetRevision);
-                }
+            var revisions = _revisionGrid.GetSelectedRevisions();
+            if (!revisions.Any(r => r.IsArtificial))
+            {
+                return;
+            }
+
+            DiffFiles.StoreNextIndexToSelect();
+            SetDiffs(revisions);
+            if (DiffFiles.SelectedItem == null)
+            {
+                DiffFiles.SelectStoredNextIndex();
             }
         }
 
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -55,18 +53,6 @@ public RevisionDiffControl()
 
         public void ForceRefreshRevisions()
```
RussKie: This method has only one callsite, so we can update the callsite to call `RefreshArtificial` and remove the method completely.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -76,9 +62,14 @@ public void RefreshArtificial()
             {
                 var revisions = _revisionGrid.GetSelectedRevisions();
 
-                if (revisions.Count > 0 && revisions[0].IsArtificial)
+                if (revisions.Any(r => r.IsArtificial))
```
RussKie: Would be good to invert all `if`s here.

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: It is not pointless to have two public interfaces here, but there will likely be some other reasoning at next change so I removed it

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -53,33 +51,24 @@ public RevisionDiffControl()
...
+            if (DiffFiles.SelectedItem == null)
             {
-                var revisions = _revisionGrid.GetSelectedRevisions();
-
-                if (revisions.Count > 0 && revisions[0].IsArtificial)
-                {
-                    DiffFiles.SetDiffs(revisions, _revisionGrid.GetRevision);
-                }
+                DiffFiles.SelectStoredNextIndex();
             }
```
mstv: Call the identical `DisplayDiffTab` instead, please.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -156,12 +145,23 @@ private string GetShortcutKeyDisplayString(Command cmd)
...
             var revisions = _revisionGrid.GetSelectedRevisions();
+            DiffFiles.StoreNextIndexToSelect();
+            SetDiffs(revisions);
+            if (DiffFiles.SelectedItem == null)
+            {
+                DiffFiles.SelectStoredNextIndex();
+            }
+        }
+
+        private void SetDiffs(IReadOnlyList<GitRevision> revisions)
```
mstv: Can be inlined again.

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: Actually, in DisplayDiffTab(), SelectFirstVisibleItem() should be called instead, to select the first in the list if the currently selected cannot be "followed". Otherwise e.g. file with index 3 will be selected if that was selected before which can seem "random".
(Even if it generally is more useful to get to the same position in the list when switching commits).

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -156,12 +145,22 @@ private string GetShortcutKeyDisplayString(Command cmd)
...
+            SetDiffs(revisions);
+            if (DiffFiles.SelectedItem == null)
+            {
+                DiffFiles.SelectFirstVisibleItem();
+            }
+        }
+
+        private void SetDiffs(IReadOnlyList<GitRevision> revisions)
+        {
+            GitItemStatus oldDiffItem = DiffFiles.SelectedItem;
```
gerhardol: I would have preferred that oldDiffItem was set only at explicit (clicked) file settings, but this is not so easy to achieve what I find

GitUI/CommandsDialogs/RevisionDiffControl.cs
mstv: Yes, this will feel "random".
In addition, it happens that the file is found in the "Diff with BASE" only. This has confused me, especially if the heading is scrolled out of view.

mstv: In general, it is a very useful feature when jumping between two commits.
There is a flicker yet: The FileDiff is cleared before the other diff of the same file is displayed.

gerhardol: @mstv Are you OK with the limitations, can I squash and request a merge?

mstv: This "random" selection happened a few times to me. I am afraid, it willl confuse users.

gerhardol: You mean if the file is selected in BASE?
If so, the first file could be selected except for Artificial refresh.
(But I like to "follow" the file selected, even if not perfect.)

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: As explained in the change above, no inlining.

GitUI/CommandsDialogs/RevisionDiffControl.cs
mstv: > Otherwise e.g. file with index 3 will be selected if that was selected before which can seem "random".

Also this happened a few times today. Let me observe this another day.

gerhardol:
@msftbot merge in 24 h

gerhardol:
@msftbot merge in 24 hours

PR #7720, opened 2020-02-03
improve file status list presentation

gerhardol:
Part of #4154 

## Proposed changes

A few related changes, preferably reviewed commit by commit
- Avoid a call to git-rev-parse to get common ancestor (BASE) when comparing
- Show the diff for common ancestor (BASE) in the status list
- ToolTip for multiple parents setting

English.xlf is not updated, UpdateLocalEnglishTranslations.ps1 require changes

### After

image

(removed colon after with:)
image

## Test methodology 

Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7800, opened 2020-02-23
Stash: Form fail to load

gerhardol:
Fixes a regression in #7720

## Proposed changes

Stash Form was not loaded due to the unused revision was set to null

## Test methodology 

Test already exists - failed in master

----


gerhardol:
@msftbot merge when @RussKie or @mstv approves

Issue #7723, opened 2020-02-04
DiffViewer displays a trailing newline which does not exist in the file


mstv:
## Current behaviour

A trailing newline is appended to every (text) diff.

## Expected behaviour

Display the end of file exactly as it is in the file.
Display the line number for the last line also if it is empty.

## Steps to reproduce

- add/remove a newline to/from the end of a file
- open FormCommit
- tick `Show nonprinting characters`

## Screenshots

![grafik](https://user-images.githubusercontent.com/36601201/73725079-f56e4880-472c-11ea-9475-976343ae9ab3.png)

Hex Diff:
![grafik](https://user-images.githubusercontent.com/36601201/73725280-56961c00-472d-11ea-92db-f8c1d04a6060.png)

## Did this work in previous version of GitExtensions

not checked

## Environment

- Git Extensions 3.3.0
- Build 7034c98856ecb70f1d716d12e92652e6998c98ff (Dirty)
- Git 2.21.0.windows.1 (recommended: 2.23.0 or later)
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.8.4069.0
- DPI 96dpi (no scaling)

gerhardol:
This is how a diff is formatted, so I do not consider this a bug.
But it is confusing, especially as lines with a newline last as commented in https://github.com/gitextensions/gitextensions/pull/7824#issuecomment-592791409

mstv:
Fixed by #9620

Issue #6723, opened 2019-05-31
Filenames starting with a space are not handled correctly


mstv:
## Current behaviour

The contents of a new file whose name starts with a space is not displayed in the FormCommit.
Such a file cannot be staged using GE.

## Expected behaviour

If a file cannot be read, an error message is displayed instead of an empty file!
GE can handle every file existing in the filesystem.

## Steps to reproduce

(I needed an example for a new file.)
- select `.editorconfig` in your favorite file manager
- hit `Ctrl+C`, `Ctrl+V`
- open FormCommit
- select the file " - Kopie.editorconfig" which is shown without the leading space
- stage it

## Screenshots

![grafik](https://user-images.githubusercontent.com/36601201/58735358-75a01980-83fa-11e9-92cf-58f36768cdfa.png)

## Did this work in previous version of GitExtensions

don't know, at least not in 2.51.05

## Environment

- Git Extensions 3.2.0
- Build ef3cd4d8f0e4a952dee6560a64f5ee4be4725a37
- Git 2.21.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

## Diagnostics

git check-attr -z diff text crlf eol -- "- Kopie.editorconfig"

gerhardol:
@mstv Any comments?

RussKie:
Is it because we are missing `"`?

gerhardol:
> Is it because we are missing "?

Trim and Uri(), see PR

Issue #5329, opened 2018-08-19
Commit dialog: inconsistent display of new binary file


mstv:
**Do you want to request a *feature* or report a *bug*?**
inconsistency

**What is the current behavior?**
A new binary file is displayed with file contents if unstaged and without if staged.
![grafik](https://user-images.githubusercontent.com/36601201/44314342-4d6d9e00-a418-11e8-9296-52b5c62c0717.png)
![grafik](https://user-images.githubusercontent.com/36601201/44314351-6e35f380-a418-11e8-91ca-83ea5e7f9edc.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Add a new binary file, open the commit dialog and select the file. Then stage the file.

**What is the expected behavior?**
The staged **new** file should not be compared with /dev/null.
It's contents should be displayed as it is done if it is unstaged.

**Environment you encounter the issue:**

- GitExtensions version: master
- GIT version: 2.17.1.w.2-x64
- OS version: Windows 7
- .NET version: 4.7

**Did this work in previous version of GitExtensions (which)?**
have not checked yet

vbjay:
It's a new file not committed. Dev null is correct.

mstv:
Yes, /dev/null is technically correct. But that's not the issue.
From user point of view there is no reason why the display of the file should depend on whether it's staged or not.

gerhardol:
The current implementation tries to present the diff in the best way possible. Changing this will limit the functionality, I prefer to keep the "inconsistency".

In worktree (unstaged) you can only compare to files that is known by Git.
The contents are displayed for new files.

For Index (staged) files, git-diff can be used. If the file is new in Index, it will be presented to /dev/null. This allows for instance partial unstage of lines in the file.

mstv:
This doesn't work well for a new **binary** file when it has been staged:
![grafik](https://user-images.githubusercontent.com/36601201/44314351-6e35f380-a418-11e8-91ca-83ea5e7f9edc.png)

As long as it is unstaged its contents *is* displayed:
![grafik](https://user-images.githubusercontent.com/36601201/44314342-4d6d9e00-a418-11e8-9296-52b5c62c0717.png)

gerhardol:
[...]

PR #7824, opened 2020-02-28
revdiff: Show new file in worktree as file, not diff

gerhardol:
Part of #7338
(likely the last before the feature itself)

## Proposed changes

Present new files in WorkTree as plain files instead of a diff.
This aligns RevDiff with FormCommit again.
This was changed in  2.49 or so by myself when implementing diff to artificial commits, I did not find a good solution at time.

However, the actual reason I change this is to simplify #7338 to stage changes in new files. (FilePreamble is not set for diffs, which should be fixable too.)

Alternatives:
 * Change FormCommit (and more?) as well as FilePreamble
 * Show all New (and removed) as files

This also makes a minor refactoring for ViewChangesAsync, tobe squashed

### Before

image

### After

image

## Test methodology

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormDiff.cs
+++ b/GitUI/CommandsDialogs/FormDiff.cs
@@ -118,16 +118,9 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            var baseCommit = ckCompareToMergeBase.Checked ? _mergeBase : _baseRevision;
+            var baseCommit = (ckCompareToMergeBase.Checked ? _mergeBase : _baseRevision) ?? DiffFiles.SelectedItemParent;
 
-            var items = new List<GitRevision> { _headRevision, baseCommit };
-            if (baseCommit == null)
-            {
-                // This should not happen
-                items = new List<GitRevision> { _headRevision, DiffFiles.SelectedItemParent };
-            }
-
-            DiffText.ViewChangesAsync(items, DiffFiles.SelectedItem, string.Empty);
+            DiffText.ViewChangesAsync(baseCommit?.ObjectId, _headRevision, DiffFiles.SelectedItem, string.Empty);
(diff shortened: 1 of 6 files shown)


mstv:
The "after" screenshot looks suspicious - even weird than #7723. Does the file have 3 lines or 4? 😕

image

gerhardol:
> The "after" screenshot looks suspicious - even more weird than #7723. Does the file have 3 lines or 4? 😕

line 3 has a newline so line 4 is empty
This is how the file is presented in all GE fileviews as well as in Notepad++

mstv:
You are right. The diff display is confusing/inconsistent because it omits the line number of the last line if it is empty. I've added this to #7723.

mstv:
Two limitations:
- by intention in the code: Display as file is only applied for `Working directory`. Why?
  We already have an issue #5329 regarding the same inconsistency in FormCommit.
- generic issue #6723: Files with names starting with a space are not handled correctly.
  The displayed exception
  grafik
  is better than the silent fail in the current master
  grafik

gerhardol:
> by intention in the code: Display as file is only applied for Working directory. Why?

Because it happened to be this way, I guess. I can change it though, rather as a followup PR.
I will initiate a merge and open a separate PR - you can protest to include the correction in this PR too.

> generic issue #6723: Files with names starting with a space are not handled correctly.

A completely different issue
GitItemStatusFromStatusCharacter.cs:996 uses Trim()
Unsure of the consequences if removed, but TrimEnd() should be OK.
With that the GitItemStatus is correct and the file can for instance be staged

But still exception when viewing. Path.GetFullPath() in FileViewer:ViewFileAsync() also removes the leading space.

gerhardol:
@msftbot merge in 24 hours

PR #7825, opened 2020-02-28
RevisionDiff: Stage/Unstage selected lines

gerhardol:
Fixes #7338 

Based on #8708 

## Proposed changes

Move handling of stage/unstage/reset line patching from FormCommit to FileViewer.
Instead of FormCommit adding menu items to FileViewer, handle
the actions directly in FileViewer, including the hotkey handling.
Also hotkey for general stage/reset lines (previously cherry-pick and revet line patches).
This also aligns the stage/reset menu items to other forms.

Apply the line patching also to RevDiff, supporting line patches for
worktree/index properly, including hotkey handling.
When manipulating a diff involving an artificial commit, refresh the diff

Some code cleanup, like using ViewChangesAsync consistently for diffs/patches

This is a step to avoid using FormCommit to prepare commits

## Screenshots

For WorkTree, Similar for Index
Normal commits gets hotkeys and changed Copy grouping too

### Before

image

image

### After

image

image

## Test methodology 

Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7831, opened 2020-03-01
Show new files as file, not diff 

### LINKED ISSUE #5329: Commit dialog: inconsistent display of new binary file

gerhardol:
Fixes #5329 
Follow up to #7824

## Proposed changes

Always show new files as files, not as a patch

Note: A sideeffect for this is that it will not be possible to cherry-pick/revert lines for new files (except FormCommit, as is now).
This is planned for revDiff artificial commits in #7825 , not for other commits or forms right now.

### Before

image

### After

image

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -530,21 +530,7 @@ namespace GitUI.Editor
             ThreadHelper.JoinableTaskFactory.RunAsync(
                 async () =>
                 {
-                    if (!item.IsTracked)
-                    {
-                        await ViewFileAsync(item.Name, item.IsSubmodule, openWithDifftool);
-                        SetVisibilityDiffContextMenuStaging();
-                    }
-                    else if (!item.IsSubmodule)
-                    {
-                        var patch = await Module.GetCurrentChangesAsync(
-                            item.Name, item.OldName, isStaged, GetExtraDiffArguments(), Encoding);
-                        await ViewPatchAsync(item.Name, patch?.Text ?? "", openWithDifftool, isText: false);
-
-                        // For staged/unstaged, separate stage/reset menus are shown
-                        SetVisibilityDiffContextMenuStaging();
-                    }
-                    else
+                    if (item.IsSubmodule)
                     {
                         var getStatusTask = item.GetSubmoduleStatusAsync();
                         if (getStatusTask != null)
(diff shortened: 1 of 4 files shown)


gerhardol:
If this is merged, FileViewer:ViewCurrentChanges() could be replaced with GitUIExtensions.cs: ViewChangesAsync() to unify the handling further.
#7825 is needed (should be done in that PR.)

mstv:
Sorry, I am too busy for the code review this week.
But a question and a wish arose:
- How are deleted files displayed?
- A header with the file name would be very useful for me, i.e.
  "+++ b/path/file.ext" or "--- a/path/file.ext", respectively.

gerhardol:
> How are deleted files displayed?

As a patch.
Changing this a little more tricky, especially if line patching is to be supported.

> A header with the file name would be very useful for me, i.e.
    "+++ b/path/file.ext" or "--- a/path/file.ext", respectively.

That will interfere with line patching, not so easy at it seems

To change the file presentation like this for normal diffs is something discussed before and should be considered.

mstv:
> > How are deleted files displayed?

> As a patch.

Good, so a delete file looks different than an added one.

> header ... will interfere with line patching

OK, no issue for me. Line patching has high precedence.

Though a quite late question, asked for the last time: What is the benefit of displaying new files not as diff? It seems to require extra handling, e.g. for line patching.

gerhardol:
> Though a quite late question, asked for the last time: What is the benefit of displaying new files not as diff? It seems to require extra handling, e.g. for line patching.

The benefit from the previous was to align FormCommit and RevisionDiff, as well as setting FilePreamble in FileViewer, required for patching (there are other ways to fix this though).
#7824 was therefore required for #7825, this is just a follow up.

GitUI/Editor/FileViewer.cs
```diff
@@ -530,19 +530,19 @@ public void ViewCurrentChanges(GitItemStatus item, bool isStaged, [CanBeNull] Ac
             ThreadHelper.JoinableTaskFactory.RunAsync(
                 async () =>
                 {
-                    if (!item.IsTracked)
+                    if (!item.IsSubmodule)
                     {
-                        await ViewFileAsync(item.Name, item.IsSubmodule, openWithDifftool);
-                        SetVisibilityDiffContextMenuStaging();
```
mstv: Can the call of `SetVisibilityDiffContextMenuStaging` really be removed?
If yes, the method itself should be removed. Or will it be reused in a follow-up PR?

Slightly off-topic: FileStatusList context menu:
I tried to `cherry pick changes` from a new file in an unmerged branch (like "bugfix/i6723-filename-space-2") without effect.
I guess this menu item is not applicable for new files. The same applies to `Reset file(s) to | Selected`.

gerhardol:
@mstv any comments?
I have several changes depending on this, so I will initiate a merge

For cherry picking patches, I plan to clean this up but likely not add patching unless Artificial commits are selected. Such patches fail often anyway.

GitUI/Editor/FileViewer.cs
gerhardol: > Can the call of SetVisibilityDiffContextMenuStaging really be removed?
If yes, the method itself should be removed. Or will it be reused in a follow-up PR?

SetVisibilityDiffContextMenuStaging is not removed, just moved around.
However #7825 removes SetVisibilityDiffContextMenuStaging  and a follow up to that will remove ViewCurrentChanges() and let all "patchviewers" use ViewChangesAsync()
The unification is why I prefer to look over this after #7825

mstv: Sounds good.

GitUI/Editor/FileViewer.cs
```diff
@@ -530,19 +530,19 @@ public void ViewCurrentChanges(GitItemStatus item, bool isStaged, [CanBeNull] Ac
...
+                            await ViewGitItemRevisionAsync(item, id, openWithDifftool);
+                        }
+                        else
+                        {
+                            var patch = await Module.GetCurrentChangesAsync(
+                                item.Name, item.OldName, isStaged, GetExtraDiffArguments(), Encoding);
+                            await ViewPatchAsync(item.Name, patch?.Text ?? "", openWithDifftool, isText: false);
+                        }
                     }
                     else
```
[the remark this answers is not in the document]
gerhardol: Pushed a variant of this - plan to remove anyway

gerhardol:
> Slightly off-topic: Removed image files are displayed as empty (also before this PR).

A new and removed image would either be shown the same way or a text is shown. (Overcourse: a link to show the removed image.)

gerhardol:
Added a commit to handle unstage/revert for Index in FormCommit (this is actually the same as after other changes).

RevDiff for WorkTree/Index will be added to  #7825.
For other combinations I do not know yet, that depends on other reasons too, like if a file is staged or not. That is not changed from the current behavior, just that the menu item will not be visible.
Will probably leave that out (at least for some time), not important anyway.

gerhardol:
Rebased and squashed, added a variable for readability

gerhardol:
@msftbot merge in 24 hours

PR #7832, opened 2020-03-01
Files starting with space incorrectly handled

### LINKED ISSUE #6723: Filenames starting with a space are not handled correctly

gerhardol:
Part of #6723 (second part dropped)
See also https://github.com/gitextensions/gitextensions/pull/7824#issuecomment-593005586

## Proposed changes

- Do not trim spaces when parsing git-status and git-diff

Dropped from this PR
- Uri() removes spaces in pathResolver

Follow up to #7587 
Similar change could be required in PathUtil:NormalizePath() and FormClone.cs (there are reports there) 

## Screenshots 

See issue for the error message

## Test methodology
Fixed incorrect tests, added new tests

----


UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
```diff
@@ -157,7 +157,7 @@ public void TestGetDiffChangedFilesFromString()
             {
                 // git diff -M -C -z --cached --name-status
                 // Ignore unmerged (in conflict) if revision is work tree
-                string statusString = "M  testfile.txt\0U  testfile.txt\0";
+                string statusString = "M\0testfile.txt\0U\testfile.txt\0";
```
RussKie: Is one missing `0`?	
```suggestion
                string statusString = "M\0testfile.txt\0U\0testfile.txt\0";
```

GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -993,7 +993,7 @@ private static GitItemStatus GitItemStatusFromStatusCharacter(StagedStatus stage
 
             return new GitItemStatus
             {
-                Name = fileName.Trim(),
+                Name = fileName,
```
RussKie: Any thoughts as to why we had `Trim()` here?

GitCommands/FullPathResolver.cs
```diff
@@ -56,8 +56,8 @@ public string Resolve(string path)
                 basePath += Path.DirectorySeparatorChar;
             }
 
-            var uri = new Uri(new Uri(basePath), path);
-            return uri.LocalPath;
+            var uri = Path.Combine(basePath, path);
+            return uri;
```
RussKie: We must be absolutely certain we aren't regressing (e.g. #6702).

GitCommands/FullPathResolver.cs
```diff
@@ -56,8 +56,8 @@ public string Resolve(string path)
                 basePath += Path.DirectorySeparatorChar;
             }
 
-            var uri = new Uri(new Uri(basePath), path);
-            return uri.LocalPath;
+            var uri = Path.Combine(basePath, path);
```
RussKie: ❗️  This will won't work as you expect it - `path` can be anything (e.g. `..\..\`):
image

We must call `Path.GetFullPath()` to ensure the result is correct:
image


💭 We should probably enhance the docs to state that while this method resolves paths it doesn't check or ensure the path exists.

UnitTests/GitCommands.Tests/FullPathResolverTests.cs
```diff
@@ -33,22 +33,18 @@ public void Resolve_should_return_original_path_if_rooted(string path)
...
...\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\folder\fo
```
RussKie: We must now test for AE: https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netframework-4.8#System_IO_Path_Combine_System_String_System_String_

RussKie: Why? We can get paths with both types of splashes

RussKie: This was meant for this change:
image

UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
gerhardol: Yes, will change

GitCommands/Git/GitCommandHelpers.cs
gerhardol: It has been like this for a long time, so I guess it was just an assumption how things work. Maybe adjusted to tests?

This may not be complete, to handle copied files the rename in GitItemStatusFromCopyRename() must be removed too (affects git-diff and git-status porecelain v1).
This is a small part though.

GitCommands/FullPathResolver.cs
gerhardol: #6702 is a part of the problem as it uses Uri() to get the path and Uri().LocalPath has different rules than what the file system allows.

Uri() is used also in Clone, where there has been a number of illegal char handling recently, why this is maybe should be changed there too.

GitCommands/FullPathResolver.cs
gerhardol: The results are what I expect
But documenting is OK I guess

UnitTests/GitCommands.Tests/FullPathResolverTests.cs
gerhardol: The test normalizes the expected output, so we can change the test instead
Path.Combine do not normalize the path but Uri().LocalPath does

gerhardol:
I investigated to see that the problem was not something that should be changed in what I was changing anyway. 
When I found the problem, I could as well submit a PR. But the change is not important in my view.

The part regarding Git parsing should be merged regardless (maybe enhanced).
The change for Path.Combined could be dropped if it is controversial.
If both are to be merged, it is nice to have them in the same PR though, as both are needed to handle files with spaces.

GitCommands/Git/GitCommandHelpers.cs
gerhardol: Removed Trim() for rename, copy, added test case

UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
```diff
@@ -205,6 +205,26 @@ public void TestGetDiffChangedFilesFromString()
...
+                Assert.IsTrue(status[0].IsRenamed == true);
+                Assert.IsTrue(status[1].OldName == "apa.md");
+                Assert.IsTrue(status[1].Name == " apa.md");
+                Assert.IsTrue(status[1].RenameCopyPercentage == "70");
+                Assert.IsTrue(status[1].IsCopied == true);
+                Assert.IsTrue(status[2].Name == " co decov.yml");
+                Assert.IsTrue(status[2].IsNew == true);
+                Assert.IsTrue(status[3].Name == "CODE_OF_CONDUCT.md");
+                Assert.IsTrue(status[3].IsDeleted == true);
+            }
```
RussKie: I suggest we use ApprovelTests to assert the blob here.
The code would be something like the following, and that should just generate an \*.approved.\* file that you commit:
```suggestion
                string statusString = "R100\0CONTRIBUTING.md\0 CONTRIBUTI NG.md\0C70\0apa.md\0 apa.md\0A\0 co decov.yml\0D\0CODE_OF_CONDUCT.md\0";
                var status = GitCommandHelpers.GetDiffChangedFilesFromString(module, statusString, "123", "456", null);
                Approvals.Verify(status, "Names");
            }
```

gerhardol: I have spent the entire evening on this and may got it working with VerifyAll (Verify is derecated method but there has not been new ApprovalTest updates in 6 years so new versions may not be a problem). Will probably take some days until I am able to finish this.

RussKie: No rush, hola if you get stuck and need a hand

RussKie:
I have added a bunch of tests with possible use cases. Most of them don't pass. I suspect we are regressing with the change, and further tweak are required. I'll have a look in coming days.

gerhardol:
Squashed the comment (kept the exception removal separate), removing ApprovalTests and the resolver update
Previous commits in https://github.com/gerhardol/gitextensions/tree/bugfix/i6723-filename-space-2, to be handled later

gerhardol:
Now the resolver changes was reversed too

UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
RussKie: Could you please explain the issue?

gerhardol: See https://github.com/gerhardol/gitextensions/blob/bugfix/i6723-filename-space-2/UnitTests/GitCommands.Tests/Git/GitCommandsHelperTest.TestGetDiffChangedFilesFromString.ForScenario.Check_that%20spaces%20are%20not%20trimmed%20in%20file%20names.approved.txt

Only the string output, not the contents

gerhardol: @RussKie I will initiate the merge, abort that or open a new issue if you have a better solution

gerhardol:
@msftbot merge in 24 hours

GitUI/Editor/FileViewer.cs
```diff
@@ -455,7 +455,7 @@ string GetFileText()
 
                 if (!File.Exists(path))
                 {
-                    return null;
+                    return $"File {path} does not exist";
```
mstv: I get this text in combination with #7831 for an unstaged new file:
"File F:\Build\gitextensions3\- Kopie.editorconfig does not exist"

gerhardol: The part of the PR that was removed fixed that, but there may have been sideeffects so it was dropped

> Un/staging works well. Edit, Open with..., Delete are missing in the context menu.

Also related to the file resolver

UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
RussKie: You've almost nailed it, a little tweak is required.

RussKie: I can't push into your fork, so here's the diff:
```diff
diff --git a/UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs b/UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
index 6770b2f22..5f678ab0a 100644
--- a/UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
+++ b/UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
@@ -1,11 +1,14 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using ApprovalTests;
 using ApprovalTests.Namers;
 using GitCommands;
 using GitCommands.Git;
+using GitCommands.Utils;
 using GitUIPluginInterfaces;
+using Newtonsoft.Json;
 using NUnit.Framework;
 using ResourceManager;
 
@@ -122,43 +125,31 @@ public void TestFetchArguments()

         public void TestGetDiffChangedFilesFromString(string testName, string firstRevision, string secondRevision, string parentToSecond, string statusString)
         {
             // TODO produce a valid working directory
             var module = new GitModule(Path.GetTempPath());
-            using (ApprovalResults.ForScenario(testName))
+            using (ApprovalResults.ForScenario(testName.Replace(' ', '_')))
             {
                 // git diff -M -C -z --name-status
-                Approvals.VerifyAll<GitItemStatus>(GitCommandHelpers.GetDiffChangedFilesFromString(module, statusString, firstRevision,
-                    secondRevision, parentToSecond), "status");
+                var statuses = GitCommandHelpers.GetDiffChangedFilesFromString(module, statusString, firstRevision, secondRevision, parentToSecond);
+                Approvals.VerifyJson(JsonConvert.SerializeObject(statuses));
             }
         }

```

```diff
diff --git a/UnitTests/GitCommands.Tests/GitCommands.Tests.csproj b/UnitTests/GitCommands.Tests/GitCommands.Tests.csproj
index 459104224..7431b2e38 100644
--- a/UnitTests/GitCommands.Tests/GitCommands.Tests.csproj
+++ b/UnitTests/GitCommands.Tests/GitCommands.Tests.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2018.2.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
   </ItemGroup>
 
   <ItemGroup>
@@ -31,13 +32,18 @@
     </None>
     <EmbeddedResource Include="UserRepositoryHistory\Legacy\MockData\CategorisedRepositories01.xml" />
     <EmbeddedResource Include="UserRepositoryHistory\Legacy\MockData\CategorisedRepositories02.xml" />
-    <None Include="UserRepositoryHistory\RepositoryXmlSerialiserTests.Deserialize_remote_repositories_without_ns.approved.txt" />
-    <None Include="UserRepositoryHistory\RepositoryXmlSerialiserTests.Deserialize_remote_repositories_with_ns.approved.txt" />
-    <None Include="UserRepositoryHistory\RepositoryXmlSerialiserTests.Serialize_recent_repositories.approved.xml" />
-    <None Include="UserRepositoryHistory\Legacy\RepositoryCategorySerialiserTests.Verify_backwards_compatibility_of_object_graph.approved.xml" />
+    <None Include="**\*.approved.*" />
   </ItemGroup>
+  
   <ItemGroup>
-    <None Include="GitRevisionSummaryBuilderTests.Should_do_ellipsis.ForScenario.Too_long_lines.approved.txt" />
-    <None Include="GitRevisionSummaryBuilderTests.Should_do_ellipsis.ForScenario.Too_many_lines.approved.txt" />
+    <None Update="**\*.approved.*">
+      <DependentUpon>$(
+  [System.String]::Copy('%(FileName)')
+    .SubString(
+      0,
+      $([System.String]::Copy('%(FileName)').IndexOf('.'))
+    )
+)$(ProjectExt.Replace('proj', ''))</DependentUpon>
+    </None>
   </ItemGroup>
 </Project>
\ No newline at end of file
```

gerhardol: Updated

UnitTests/GitCommands.Tests/Git/GitCommandHelpersTest.cs
```diff
@@ -244,6 +198,15 @@ public void TestGetStatusChangedFilesFromString()
...
             }
+
+            {
+                // git status --porcelain=2 --untracked-files=no -z
+                string statusString = "#Header\03 unknown info\01 .M S..U 160000 160000 160000 cbca134e29be13b35f21ca4553ba04f796324b1c cbca134e29be13b35f21ca4553ba04f796324b1c  no trim space0 \01 .M SCM. 160000 160000 160000 6bd3b036fc5718a51a0d27cde134c7019798c3ce 6bd3b036fc5718a51a0d27cde134c7019798c3ce  no trim space1 \0\r\nwarning: LF will be replaced by CRLF in adfs.h.\nThe file will have its original line endings in your working directory.\nwarning: LF will be replaced by CRLF in dir.c.\nThe file will have its original line endings in your working directory.";
+                var status = GitCommandHelpers.GetStatusChangedFilesFromString(module, statusString);
+                Assert.IsTrue(status.Count == 2);
+                Assert.IsTrue(status[0].Name == " no trim space0 ");
+                Assert.IsTrue(status[1].Name == " no trim space1 ");
+            }
```
RussKie: We should convert these tests as well.

PR #7833, opened 2020-03-01
Fail for dirty translations

RussKie:
## Proposed changes

* Allow update the English translations from the source
* Enforce the English translations are upto date on the CI


gerhardol:
No comments on the changes - this is needed (and we can fix later)

Issue #7702, opened 2020-01-29
Improve Arcade build


mstv:
### Tests

- [x] Do not run tests in parallel -- as it seems to be:
  ```
  Running tests: C:\projects\gitextensions\artifacts\bin\tests\UnitTests\GitCommands.Tests\net461\GitCommands.Tests.dll [net461|x64]
  Running tests: C:\projects\gitextensions\artifacts\bin\tests\UnitTests\GitUI.Tests\net461\GitUI.Tests.dll [net461|x64]
  Tests succeeded: C:\projects\gitextensions\artifacts\bin\tests\UnitTests\GitCommands.Tests\net461\GitCommands.Tests.dll [net461|x64]
  ```
- [x] On build fail, the AppVeyor build shall fail (turn red)
- [x] On test fail, the AppVeyor build shall fail (turn red)
  - [x] after running all tests
  - [x] after storing the artifacts
  - [x] including each `TestResult.xml`
- [x] Show failed test cases on the AppVeyor console (e.g. just `type FooTest.log`)
- [ ] All test passed in about 3 minutes before. (I don't know the reason for this speedup a few weeks ago.)
  With the Arcade build, the test run takes about 10 minutes although tests are run in parallel.
- [ ] Double check all files matching `tests/.*\.txt` for correct line endings.
- [ ] There should be a default active project, which is started on Run/Debug (`F5`).
      STR: `git clean -xdf`; open sln; hit `F5` => does not build and run
- [x] Check `English.xlf` (only in CI build).
- [x] `Build/Run-Tests.ps1` has become obsolete - remove
- [x] Are there more obsolete scripts?

RussKie:
> * `Build/Run-Tests.ps1` has become obsolete - remove
> * Are there more obsolete scripts?

Yes, quite likely. There is more work to be done to complete the conversion.

RussKie:
> * Do not run tests in parallel

Tests _are_ run in sequence.
https://github.com/gitextensions/gitextensions/blob/0b00208914c1e126de69fdbf162bc2252113b344/Build/tools/NUnit/NUnit.targets#L44
You can verify it by watching at Process Explorer while executing tests - there's only one running at a time. You can also check the binlog (look under .\gitextensions\artifacts\log\[Debug|Release], use [MSBuildStructuredLog](https://github.com/KirillOsenkov/MSBuildStructuredLog)) and see what command line is invoked:
[...]

Issue #7745, opened 2020-02-08
Flaky test GetSortedTags_should_load_ref_different_in_case


mstv:
## Current behaviour

The test `GitUITests.CommitInfo.CommitInfoTests.GetSortedTags_should_load_ref_different_in_case
` fails occasionally:
```
System.Exception : Unexpected arguments: for-each-ref --sort=-taggerdate --format="%(refname)" refs/tags/
   at CommonTestUtils.MockExecutable.GetOutput(ArgumentString arguments) in C:\projects\gitextensions\UnitTests\CommonTestUtils\MockExecutable.cs:line 118
   at GitUI.CommitInfo.CommitInfo.GetSortedTags() in C:\projects\gitextensions\GitUI\CommitInfo\CommitInfo.cs:line 238
   at GitUI.CommitInfo.CommitInfo.TestAccessor.GetSortedTags() in C:\projects\gitextensions\GitUI\CommitInfo\CommitInfo.cs:line 860
   at GitUITests.CommitInfo.CommitInfoTests.GetSortedTags_should_load_ref_different_in_case() in C:\projects\gitextensions\UnitTests\GitUI.Tests\CommitInfo\CommitInfoTests.cs:line 190
```

## Expected behaviour

test passes stably

## Steps to reproduce

run the tests on AppVeyor many times

## Environment

- GE d5a4408
- AppVeyor
- Windows 10

PR #7837, opened 2020-03-03
Enhance tests output

RussKie:
Contributes to #7702

### DIFF
--- a/appveyor.experimental.yml
+++ b/appveyor.experimental.yml
@@ -1,16 +1,30 @@
-version: 4.0.0.{build}
+version: 66.66.66.{build}
 
-branches:
-  except:
-    - configdata
-    - gh-pages
-    - /experimental\/*/
+# Do not build on tags (GitHub, Bitbucket, GitLab, Gitea)
+skip_tags: true
+
+# set clone depth
+clone_depth: 1                     # clone entire repository history if not defined
 
 matrix:
-  fast_finish: true
+  fast_finish: false
 
 # https://www.appveyor.com/docs/build-environment/#build-worker-images
 image: Visual Studio 2019
+
+environment:
+  # The following entries define the default values of the environment variables.
+  # They can be overridden in AppVeyor | Settings | Environment.
+  GE_TEST_SLEEP_SECONDS_ON_HANG: 0
+  GE_TEST_LAUNCH_DEBUGGER_ON_HANG: 0
+  # How many times to execute tests
+  TEST_RUN_COUNT: 1
+  # Execute unit tests
+  RUN_UNIT_TESTS: TRUE
+  # Execute integation tests
+  RUN_INTEGRATION_TESTS: TRUE
+  # Fail the build if any of the tests fail
+  FAIL_IF_TEST_FAIL: TRUE
 
 init:
   - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
@@ -45,17 +59,26 @@ build_script:
     }
 
     # build
-    & .\cibuild.cmd -restore -build -buildNative -loc
+    & .\cibuild.cmd -restore -build -buildNative
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
 
 
 # to run your custom scripts instead of automatic tests
 test_script:
 - ps: |
-    & .\cibuild.cmd -test
-    if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode);  }
-    & .\cibuild.cmd -integrationTest
-    if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
+    $testRunCount = $env:TEST_RUN_COUNT
+    for ($testRun = 1; $testRun -le $testRunCount; $testRun++)
+    {
+      if (!$env:RUN_UNIT_TESTS) {
+        & .\cibuild.cmd -test
+        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
+      }
+
+      if (!$env:RUN_INTEGRATION_TESTS) {
+        & .\cibuild.cmd -integrationTest
+        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
+      }
+    }
 
 
 # scripts to run after tests
@@ -69,10 +92,6 @@ after_test:
 artifacts:
   - path: artifacts\log\**\*.binlog
   - path: artifacts\publish\*.zip
-#  # upload the generated portable archive only
-#  - path: 'Setup/GitExtensions-Portable-*.zip'
-#  - path: 'Setup/GitExtensions-pdbs-*.zip'
-#  - path: 'GitExtensionsVSIX/bin/Release/GitExtensionsVSIX.vsix'
 
 #on_finish:
 #  - ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
@@ -93,11 +112,13 @@ on_failure:
             $failed = $result[5].Value
             $notPassed = $total - $passed
             $testErrors = $failed
-            $testName = $resultFilePath.Split('\\')[-3]
+            $testProjectName = $resultFilePath.Split('\\')[-3]
 
             if ($testErrors -ne 0) {
-              $logFile = Get-ChildItem -recurse artifacts\$testName*.log
-              Push-AppveyorArtifact $logFile
+              $artifactName = "${testProjectName}.err${testErrors}"
+              $logFile = Get-ChildItem -recurse artifacts\$testProjectName*.log
+              Push-AppveyorArtifact $logFile -FileName "${artifactName}.log"
+              Push-AppveyorArtifact $resultFilePath -FileName "${artifactName}.xml"
             }
         }
     }
--- a/appveyor.experimental.yml
+++ b/appveyor.experimental.yml
@@ -73,14 +73,14 @@ test_script:
     $testRunCount = $env:TEST_RUN_COUNT
     for ($testRun = 1; $testRun -le $testRunCount; $testRun++)
     {
-      if (!$env:RUN_UNIT_TESTS) {
+      if ($env:RUN_UNIT_TESTS) {
         & .\cibuild.cmd -test
-        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
+        if ($LastExitCode -ne 0  -and $env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
       }
 
-      if (!$env:RUN_INTEGRATION_TESTS) {
+      if ($env:RUN_INTEGRATION_TESTS) {
         & .\cibuild.cmd -integrationTest
-        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
+        if ($LastExitCode -ne 0  -and $env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
       }
     }
 
--- a/appveyor.experimental.yml
+++ b/appveyor.experimental.yml
@@ -6,6 +6,10 @@ skip_tags: true
 # set clone depth
 clone_depth: 1                     # clone entire repository history if not defined
 
+branches:
+  only:
+  - /experimental\/*/
+
 matrix:
   fast_finish: false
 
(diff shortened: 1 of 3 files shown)


RussKie:
Please ignore the first three commits, they are used for some build failures.

appveyor.yml
```diff
@@ -102,7 +100,7 @@ test_script:
             }
         }
     }
-    if ($testErrorsTotal -ne 0  -and  $env:IGNORE_TEST_RESULT -ne "true") {
```
mstv: Do you want to remove this variable or just for testing -- WIP?

appveyor.yml
```diff
@@ -80,15 +80,15 @@ test_script:
...
-                #$result[0].Value
+
                 $total  = $result[1].Value
                 $passed = $result[4].Value
                 $failed = $result[5].Value
                 $notPassed = $total - $passed
                 $testErrors = $failed
                 $testName = $resultFilePath.Split('\\')[-3]
                 $artifactName = "TestResult#${testRun}.${testErrors}err.${testName}"
-                Push-AppveyorArtifact $resultFilePath -FileName "${artifactName}.xml"
```
mstv: I agree, the artifacts have become quite noisy.
At least on error, this file should be kept because it contains the most complete information.
(The best solution IMO would be one `.xml` with all the test results.)
An alternative could be to zip them into one archive -- with the drawback that it cannot be opened directly.

mstv:
> NUnit : error : Tests failed: C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\UI.IntegrationTests.csproj [net461|x64] [C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\UI.IntegrationTests.csproj]
Build FAILED.

The CI build should not fail before all tests have been run.

appveyor.yml
```diff
@@ -80,25 +79,31 @@ test_script:
...
+
+                Write-Host "Pass#${testRun}: passed: $passed, failed: $failed, not passed: $notPassed, total: $total"
```
mstv:
```suggestion
                Write-Host "Pass #${testRun}: passed: $passed, failed: $failed, not passed: $notPassed, total: $total"
```

appveyor.yml
```diff
@@ -80,25 +79,31 @@ test_script:
...
+
                 if ($notPassed -ne 0) {
                     $logFile = Get-ChildItem -recurse artifacts\$testname*.log
                     Get-Content $logFile
                     if ($testErrors -ne 0) { Push-AppveyorArtifact $logFile -FileName "${artifactName}.log" }
                 }
                 $testErrorsTotal += $testErrors
+                Write-Host "Pass#${testRun}: testErrorsTotal: $testErrorsTotal"
```
mstv:
```suggestion
                Write-Host "Pass #${testRun}: testErrorsTotal: $testErrorsTotal"
```

appveyor.yml
```diff
@@ -80,25 +79,31 @@ test_script:
...
                 $testErrors = $failed
                 $testName = $resultFilePath.Split('\\')[-3]
                 $artifactName = "TestResult#${testRun}.${testErrors}err.${testName}"
-                Push-AppveyorArtifact $resultFilePath -FileName "${artifactName}.xml"
+
+                Write-Host "Pass#${testRun}: passed: $passed, failed: $failed, not passed: $notPassed, total: $total"
+
                 if ($notPassed -ne 0) {
                     $logFile = Get-ChildItem -recurse artifacts\$testname*.log
                     Get-Content $logFile
```
mstv: This file is printed by the Arcade script now.
```suggestion
```

appveyor.yml
RussKie: I've been mulling over it over night.

I want to remove all of this plumbing for the following reasons:
- Both master and release/*  branches are release branches, and shouldn't contain any "for testing purposes" infra. The latter belongs to pull-requests/feature branches.
- We must fail a build if either unit or integration tests fail. If unit tests fail - we have regressed, it must be fixed, no point running integration tests.

If you need this plumbing for testing purposes, please push these changes to branches on the main trunk instead of your fork. This way changes to appveyor.yml will be executed.

appveyor.yml
RussKie: I think I may have asked before, but what's the benefit of these xml for user? 
The main reason we have them for opencover/appveyor to report the stats. As a user a I look at the *.log files.
Merging xmls together is a time consuming operation which will result in a tens of thousands of lines of xml....

appveyor.yml
RussKie: These prints are for debugging purposes only, I'll remove them in the final version.
It took me awhile to realise test failures were ignore at the AppVeyor level... 🤦‍♂

appveyor.yml
RussKie: Ah! Thank you, will clean up

RussKie:
@mstv I have cleaned the code. I think it is ready.
I will remove commits other than 92641e8c3b8b96a9ebbd0a2929fdd47b01d1c6fd before the merge.

pmiossec:
@RussKie 

>I will remove commits other than 92641e8 before the merge.

I think that you could remove it because #7745 succeed in making the build fail

appveyor.yml
```diff
@@ -1,23 +1,15 @@
...
-  # The following entries define the default values of the environment variables.
-  # They can be overridden in AppVeyor | Settings | Environment.
-  matrix:
-  - IdeVersion: VS2019
-  SKIP_PAUSE: TRUE
-  ARCHIVE_WITH_PDB: TRUE
-  GE_TEST_SLEEP_SECONDS_ON_HANG: 0
-  GE_TEST_LAUNCH_DEBUGGER_ON_HANG: 0
-  TEST_RUN_COUNT: 1
-  IGNORE_TEST_RESULT: FALSE
```
mstv: Why did you remove all AppVeyor variables?
These four still make sense to me.

appveyor.yml
```diff
@@ -1,23 +1,15 @@
...
-os: Visual Studio 2019
-
 branches:
   except:
     - configdata
     - gh-pages
 
-environment:
-  # The following entries define the default values of the environment variables.
-  # They can be overridden in AppVeyor | Settings | Environment.
```
mstv: This comment is important IMO if we keep this section.

appveyor.yml
RussKie: They are useful for debug builds, but we shouldn't have them in the main build.
The master and the release builds aren't meant to be used as test grounds. For tests we can create branches off the main trunk and modify appveyor.yml as required to complete tests.

appveyor.yml
RussKie: This whole section is gone...

appveyor.yml
mstv: Git does not support this in a user-friendly way. This will end up in lots of merges, rebases and merge conflicts.
I don't need it as AppVeyor/Environment variables -- I was told to use them.
But the "extended test infrastructure" should be available and maintained in the main trunk.

BTW: The build script seems to *not* set the exit code.

appveyor.yml
RussKie: We can't have it without something defined
image

appveyor.yml
RussKie: > I don't need it as AppVeyor/Environment variables -- I was told to use them.

I suggested to use AppVeyor env variables as a way to dynamically change behaviours, if that's what you're referring to. It is certainly a better way than hardcoding magic numbers in the common infrastructure.

> But the "extended test infrastructure" should be available and maintained in the main trunk.

Why?
The master and release branches must be stable. If we have issues with tests (like now), we should create a feature branch in the main trunk and tweak its pipeline to suit (e.g. run tests multiple times, ignore errors, etc). Once the work is complete, the fix is submitted to the master, and all temporary tweaks are discarded.

RussKie: > BTW: The build script seems to _not_ set the exit code.

Could you please elaborate?

In a typical flow, `Build` function is invoked and returns the code:
https://github.com/gitextensions/gitextensions/blob/c110ab60f1ab3075b3e8e8213a1990da698fd006/Build/Build.ps1#L75-L91

If there is an exception, it returns here:
https://github.com/gitextensions/gitextensions/blob/c110ab60f1ab3075b3e8e8213a1990da698fd006/Build/Build.ps1#L113-L116

What am I missing?

mstv: > Once the work is complete, the fix is submitted to the master, and all temporary tweaks are discarded.

"discard" is the point: I do not want to add the loop around the tests over and over again.
I just want to temporarily set `$testRunCount` from `1` to a suiting value in `AppVeyor.yml` or in a build script of such a feature branch.
AppVeyor variables do not work well because they are global.

> BTW: The build script seems to not set the exit code.

I guess it has not worked before the one week old c110ab60f1ab3075b3e8e8213a1990da698fd006, which I did not take notice of.
And `$host.SetShouldExit` seems to not exit at once, which confused me.

RussKie: > "discard" is the point: I do not want to add the loop around the tests over and over again.
> I just want to temporarily set `$testRunCount` from `1` to a suiting value in `AppVeyor.yml` or in a build script of such a feature branch.
> AppVeyor variables do not work well because they are global.

This comes down to the point I'm trying to make - master or release/* branches are not the right place to test things.
We can have an `experimental` or `test` branch off the main trunk where appveyor.yml can be modified to whatever state maybe necessary, and have env variables that dynamically alter its behaviours.

RussKie: We can create a separate build project to build a specific branch (e.g. `experimental`) that would use the custom yml file.

appveyor.yml
```diff
@@ -119,3 +75,28 @@ artifacts:
...
+            $total  = $result[1].Value
+            $passed = $result[4].Value
+            $failed = $result[5].Value
+            $notPassed = $total - $passed
+            $testErrors = $failed
+            $testName = $resultFilePath.Split('\\')[-3]
+
+            if ($notPassed -ne 0) {
+                $logFile = Get-ChildItem -recurse artifacts\$testName*.log
+                if ($testErrors -ne 0) { Push-AppveyorArtifact $logFile }
```
mstv: The XML file contains the most complete information in one place. The `.log` file is much more readable but scattered, especially console output. The AppVeyor tab "Tests" has issues (I don't trust it).

```suggestion
            if ($notPassed -ne 0) {
                $artifactName = "TestResult.${testErrors}err.${testName}"
                $logFile = Get-ChildItem -recurse artifacts\$testName*.log
                Push-AppveyorArtifact $logFile -FileName "${artifactName}.log"
                Push-AppveyorArtifact $resultFilePath -FileName "${artifactName}.xml"
```

appveyor.yml
mstv: I am not convinced. It _will_ get out of sync. But go ahead. If we are lucky, we won't need it anymore.

appveyor.yml
```diff
@@ -1,23 +1,15 @@
...
-  ARCHIVE_WITH_PDB: TRUE
-  GE_TEST_SLEEP_SECONDS_ON_HANG: 0
-  GE_TEST_LAUNCH_DEBUGGER_ON_HANG: 0
-  TEST_RUN_COUNT: 1
-  IGNORE_TEST_RESULT: FALSE
+matrix:
+  fast_finish: true
+
+# https://www.appveyor.com/docs/build-environment/#build-worker-images
+image: Visual Studio 2019
```
mstv: The following two variables are evaluated by the current code, i.e. should be kept.

```suggestion
image: Visual Studio 2019

environment:
  # The following entries define the default values of the environment variables.
  # They can be overridden in AppVeyor | Settings | Environment.
  GE_TEST_SLEEP_SECONDS_ON_HANG: 0
  GE_TEST_LAUNCH_DEBUGGER_ON_HANG: 0
```

RussKie:
I create a new build project: https://ci.appveyor.com/project/gitextensions/gitextensions-experimental that builds `experimental/*` branches using `appveyor.experimental.yml`. You're welcome to modify the yml as you see necessary for testing purposes. The main build project will ignore both the yml and the experimental branches.
This should make moving changes between branches seamless.

RussKie:
I have made an update, please check

appveyor.experimental.yml
```diff
@@ -0,0 +1,128 @@
...
+    for ($testRun = 1; $testRun -le $testRunCount; $testRun++)
+    {
+      if (!$env:RUN_UNIT_TESTS) {
+        & .\cibuild.cmd -test
+        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
+      }
+
+      if (!$env:RUN_INTEGRATION_TESTS) {
+        & .\cibuild.cmd -integrationTest
+        if ($LastExitCode -ne 0  -and !$env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
```
mstv: I think it should be:
```suggestion
      if ($env:RUN_UNIT_TESTS) {
        & .\cibuild.cmd -test
        if ($LastExitCode -ne 0  -and $env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
      }

      if ($env:RUN_INTEGRATION_TESTS) {
        & .\cibuild.cmd -integrationTest
        if ($LastExitCode -ne 0  -and $env:FAIL_IF_TEST_FAIL) { $host.SetShouldExit($LastExitCode);  }
```

mstv:
Thank you.

PR #7850, opened 2020-03-06
Fix slow plugins loading

### LINKED ISSUE #7842: Plugins are extremely slow to load
## Current behaviour

Plugins takes more than 1 minutes to load in current `master`

## Expected behaviour

- Plugins are loaded quickly
- GitExtensions stop to try to load files that are not plugins dlls (but dll dependencies)

## Steps to reproduce

* Build `master` branch
* Debug the application
* See that the `Plugins` menu is empty during a long time

Yes. 3.3.1.

## Environment

- Git Extensions 3.3.0
- Build 513b9990cdf2f8262f74837780c4c39e5b409986 (Dirty)
- Git 2.25.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 192dpi (200% scaling)

## Diagnostics

It's a bad side effect of the Arcade build where now the plugins are in there own directories with all their dependencies (even if it's a good decision).

The code that try to find the dlls of plugins to load `var pluginFiles = PluginsPathScanner.GetFiles(defaultPluginsPath, userPluginsPath);` [Source here](https://github.com/gitextensions/gitextensions/blob/master/Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs#L57) in retrieving more than 2500 files (plugins dlls and all there dependencies) and so GitExtensions take a lot of time to try to load them

A very quick and not satisfying fix like that is solving the problem (251 files trying to be loaded instead of more than 2500):

```csharp
            var pluginFiles = PluginsPathScanner.GetFiles(defaultPluginsPath, userPluginsPath)
                .Where(f => !f.Name.StartsWith("System") && !f.Name.StartsWith("Microsoft"))
                .Distinct(new FileInfoEqualityComparer());

    internal class FileInfoEqualityComparer : IEqualityComparer<FileInfo>
    {
        public bool Equals(FileInfo x, FileInfo y)
        {
            return x.Name == y.Name;
        }

        public int GetHashCode(FileInfo obj)
        {
            return obj.Name.GetHashCode();
        }
    }
```

## Proposed solution

Obliged plugins dll names to start with `GitExtensions.` (the ones coming from the PluginManager) or end with `Plugin` (the ones provided with GitExtensions`) and try to load only these files.

pmiossec:
Fixes #7842

## Proposed changes

- Try to load only plugin dlls following pattern `GitExtensions.*.dll`
- Rename all the plugin projects to follow this new convention

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 63496b38ed36086e42bb74966ea3e8554377d376
- Git 2.25.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/BuildServerIntegration/Directory.Build.props
+++ b/Plugins/BuildServerIntegration/Directory.Build.props
@@ -16,9 +16,16 @@
       Direct plugins artifacts to be placed under GitExtensions/Plugins folder
     -->
   <PropertyGroup>
-    <BaseOutputPath>$([System.IO.Path]::GetFullPath('$(GitExtensionsOutputPath)$(TargetFramework)\Plugins\$(OutDirName)'))</BaseOutputPath>
+    <_BaseOutputPath>$([MSBuild]::NormalizePath('$(GitExtensionsOutputPath)', '$(TargetFramework)', 'Plugins'))</_BaseOutputPath>
+    <AssemblyName>GitExtensions.$(MSBuildProjectName)</AssemblyName>
+    <BaseOutputPath>$([MSBuild]::NormalizePath('$(_BaseOutputPath)', '$(OutDirName)'))</BaseOutputPath>
     <OutputPath>$(BaseOutputPath)</OutputPath>
     <OutDir>$(BaseOutputPath)</OutDir>
+
+    <_IsTfsInterop>$(MSBuildProjectName.StartsWith('TfsInterop.'))</_IsTfsInterop>
+    <AssemblyName Condition="'$(_IsTfsInterop)' == true">$(MSBuildProjectName)</AssemblyName>
+    <OutDir Condition="'$(_IsTfsInterop)' == true">$([MSBuild]::NormalizePath('$(_BaseOutputPath)', 'TfsIntegration'))</OutDir>
+
   </PropertyGroup>
 
 </Project>
--- a/Plugins/Directory.Build.props
+++ b/Plugins/Directory.Build.props
@@ -10,6 +10,7 @@
       Direct plugins artifacts to be placed under GitExtensions/Plugins folder
     -->
   <PropertyGroup Condition=" '$(MSBuildProjectName)' != 'GitUIPluginInterfaces'">
+    <AssemblyName>GitExtensions.$(MSBuildProjectName)</AssemblyName>
     <BaseOutputPath>$([System.IO.Path]::GetFullPath('$(GitExtensionsOutputPath)$(TargetFramework)\Plugins\$(OutDirName)'))</BaseOutputPath>
     <OutputPath>$(BaseOutputPath)</OutputPath>
     <OutDir>$(BaseOutputPath)</OutDir>
--- a/Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
+++ b/Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
@@ -54,7 +54,8 @@ namespace GitUIPluginInterfaces
             string defaultPluginsPath = Path.Combine(new FileInfo(Application.ExecutablePath).Directory.FullName, "Plugins");
             string userPluginsPath = UserPluginsPath;
 
-            var pluginFiles = PluginsPathScanner.GetFiles(defaultPluginsPath, userPluginsPath);
+            var pluginFiles = PluginsPathScanner.GetFiles(defaultPluginsPath, userPluginsPath)
+                .Where(f => f.Name.StartsWith("GitExtensions."));
 
             var cacheFile = Path.Combine(applicationDataFolder ?? "ignored", "Plugins", "composition.cache");
             IExportProviderFactory exportProviderFactory;
--- a/Plugins/Statistics/Directory.Build.props
+++ b/Plugins/Statistics/Directory.Build.props
@@ -15,6 +15,7 @@
       Direct plugins artifacts to be placed under GitExtensions/Plugins folder
     -->
   <PropertyGroup>
+    <AssemblyName>GitExtensions.$(MSBuildProjectName)</AssemblyName>
     <BaseOutputPath>$([System.IO.Path]::GetFullPath('$(GitExtensionsOutputPath)$(TargetFramework)\Plugins\$(OutDirName)'))</BaseOutputPath>
     <OutputPath>$(BaseOutputPath)</OutputPath>
     <OutDir>$(BaseOutputPath)</OutDir>


pmiossec:
FYI, there is a test that fails but it is not linked to this PR.
It always fails since a lot of builds but strangely run without problem locally. See #7745

Plugins/Bitbucket/BitbucketPullRequestForm.cs
```diff
@@ -6,15 +6,17 @@
 using System.Text;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using GitCommands;
 using GitCommands.Git;
+using GitExtensions.Bitbucket;
```
gerhardol: remove

Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -5,15 +5,17 @@
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
-using DeleteUnusedBranches.Properties;
 using GitCommands;
+using GitCommands.Git;
+using GitExtensions.DeleteUnusedBranches;
```
gerhardol: remove

Plugins/FindLargeFiles/FindLargeFilesForm.cs
```diff
@@ -6,13 +6,15 @@
 using System.Threading;
 using System.Windows.Forms;
 using GitCommands;
+using GitCommands.Settings;
+using GitExtensions.FindLargeFiles;
```
gerhardol: remove

Plugins/Gerrit/FormGerritDownload.cs
```diff
@@ -4,12 +4,15 @@
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
+using GitCommands.Git;
+using GitExtensions.Gerrit;
+using GitExtUtils.GitUI;
```
gerhardol: remove

Plugins/Gerrit/FormGerritPublish.cs
```diff
@@ -3,16 +3,18 @@
 using System.Diagnostics;
 using System.Linq;
 using System.Windows.Forms;
-using Gerrit.Server;
 using GitCommands;
+using GitCommands.Git;
+using GitExtensions.Gerrit;
```
gerhardol: remove

Plugins/ProxySwitcher/ProxySwitcherForm.cs
```diff
@@ -3,11 +3,12 @@
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using GitCommands;
+using GitExtensions.ProxySwitcher;
```
gerhardol: remove

Plugins/ReleaseNotesGenerator/ReleaseNotesGeneratorForm.cs
```diff
@@ -6,11 +6,12 @@
 using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Utils;
+using GitExtensions.ReleaseNotesGenerator;
```
gerhardol: remove

Plugins/Statistics/GitStatistics/FormGitStatistics.cs
```diff
@@ -7,14 +7,16 @@
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
+using GitCommands.Git;
+using GitExtensions.GitStatistics;
```
gerhardol: remove

pmiossec:
>Some unneeded imports

Fixed. 

>I wish #3600 was implemented when reviewing this.

I made a quick PoC 2 years ago to handle that. I could rebase it and we'll see if it works well...

RussKie:
We can do a significantly simpler change that doesn't involve physically renaming hundreds of files.
We just need to set `<AssemblyName>` property, I pushed a commit to your branch. Please have a look

RussKie:
you can simplify the last commit:

```diff
diff --git a/Plugins/BuildServerIntegration/Directory.Build.props b/Plugins/BuildServerIntegration/Directory.Build.props
index afcda7ffb..b1d207961 100644
--- a/Plugins/BuildServerIntegration/Directory.Build.props
+++ b/Plugins/BuildServerIntegration/Directory.Build.props
@@ -16,10 +16,15 @@
       Direct plugins artifacts to be placed under GitExtensions/Plugins folder
     -->
   <PropertyGroup>
+    <_BaseOutputPath>$([MSBuild]::NormalizePath('$(GitExtensionsOutputPath)', '$(TargetFramework)', 'Plugins'))</_BaseOutputPath>
     <AssemblyName>GitExtensions.$(MSBuildProjectName)</AssemblyName>
-    <BaseOutputPath>$([System.IO.Path]::GetFullPath('$(GitExtensionsOutputPath)$(TargetFramework)\Plugins\$(OutDirName)'))</BaseOutputPath>
+    <BaseOutputPath>$([MSBuild]::NormalizePath('$(_BaseOutputPath)', '$(OutDirName)'))</BaseOutputPath>
     <OutputPath>$(BaseOutputPath)</OutputPath>
     <OutDir>$(BaseOutputPath)</OutDir>
+
+    <_IsTfsInterop>$(MSBuildProjectName.StartsWith('TfsInterop.'))</_IsTfsInterop>
+    <AssemblyName Condition="'$(_IsTfsInterop)' == true">$(MSBuildProjectName)</AssemblyName>
+    <OutDir Condition="'$(_IsTfsInterop)' == true">$([MSBuild]::NormalizePath('$(_BaseOutputPath)', 'TfsIntegration'))</OutDir>
   </PropertyGroup>
 
 </Project>
```

pmiossec:
> you can simplify the last commit: 

@RussKie I made something like that at first but it seems to me better that the logic was in the `.csproj`. Because if another project is eventually in this case, another exception would have to be managed in the `Directory.Build.props`.

But I could do that if you prefer...

RussKie:
Personally I prefer to centralize the logic rather than spread it across
project files. It tends to cause a greater cognitive overload.
My 2c

pmiossec:
@RussKie done.

pmiossec:
done.

RussKie:
I'm working on the installer, and I just realised that this change is a PIA for the installer. We need to mark all original plugin files for removal, and add all new files... 😱 

How about we make this change only for DEBUG builds? That is, a RELEASE build will retain all original plugins file names?

pmiossec:
>How about we make this change only for DEBUG builds? That is, a RELEASE build will retain all original plugins file names?

No, that's not possible because the problem exists also on RELEASE.
If we don't do that, the plugins took perhaps 1 min to load (because there is an attempt to load the file and an exception thrown for every file in the plugins folder that fails to load).

I don't really understand the problem because I know really little at how installers work (except that installers are indeed PIA) but does renaming the projects to follow the naming pattern, as I did first, could help?

RussKie:
Perhaps we aren't on the same page with regards to the build configurations.

A debug build is what I expect our contributors to run with while developing. 
A release configuration is targeted towards the CI, and for generating public artifacts (e.g. a portable build and an installer). The release build, when packaged, must look like the same way as it always have (e.g. what's in your C:\Program Files (x86)\GitExtensions folder).
(I believe) The problem will disappear when we move to .NET Core/.NET because we'll no longer have an explicit dependency on netstandard.dll, that brings "the lot" with it and clutters folders.

It is a significant overhead to cater to a handful of developers who build a release version the app from code. In that case, I trust, they can alter the configuration as required.

The installer is a complicated beast, and it must handle a number of different scenarios, such as a clean install, an update or a removal. It must know what files need to installed, and where; what files must be removed when updating, and what files must be kept or updated.

PR #7853, opened 2020-03-09
Handle gracefully a NRE on body message

pmiossec:
Fixes a not visible NRE (but boring when debugging)

## Proposed changes

- test if body is null

### DIFF
--- a/GitCommands/Git/GitRevisionSummaryBuilder.cs
+++ b/GitCommands/Git/GitRevisionSummaryBuilder.cs
@@ -21,6 +21,11 @@ namespace GitCommands
 
         public string BuildSummary(string body)
         {
+            if (body == null)
+            {
+                return null;
+            }
+
             var s = new StringBuilder(Math.Min(body.Length, CommitSummaryWorstCaseLength));
 
             int lineCount = 0;
--- a/GitCommands/Git/GitRevisionSummaryBuilder.cs
+++ b/GitCommands/Git/GitRevisionSummaryBuilder.cs
@@ -21,7 +21,7 @@ namespace GitCommands
 
         public string BuildSummary(string body)
         {
-            if (body == null)
+            if (string.IsNullOrWhiteSpace(body))
             {
                 return null;
             }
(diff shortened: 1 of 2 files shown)


GitCommands/Git/GitRevisionSummaryBuilder.cs
```diff
@@ -21,6 +21,11 @@ public class GitRevisionSummaryBuilder : IGitRevisionSummaryBuilder
 
         public string BuildSummary(string body)
         {
+            if (body == null)
```
RussKie: What about empty string or whitespace?

RussKie:
Tests?

GitCommands/Git/GitRevisionSummaryBuilder.cs
pmiossec: I imagine because the empty or whitespace is well handled by the code and I just wanted to avoid the NRE.

pmiossec:
>Tests?

done.

GitCommands/Git/GitRevisionSummaryBuilder.cs
RussKie: If `body` is empty or whitespace we will be making allocations and burning cycles. Hence my question - is there a need to do that?

mstv: This is used only for tooltips (currently). Hence I do not expect performance issues.
In general, I prefer to display exactly what's in git.

pmiossec: In fact, I think that's a false debate and that there will never be an empty or whitespace git commit message.

I only wished to prevent a NRE because, for a reason, a tooltip is generated when the git message has the null value (which is counter-intuitive).  

So, I imagine that in fact, we could add `string.IsNullOrWhiteSpace()`, it will change nothing...

pmiossec: @RussKie I have pushed the string.IsNullOrWhiteSpace().

Feel free to squash on merge.

PR #7868, opened 2020-03-15
Display text if an image cannot be displayed

gerhardol:
See https://github.com/gitextensions/gitextensions/pull/7831#issuecomment-598459433

## Proposed changes

Display a text rather just an empty pane if the image has been removed or cannot be displayed for other reasons.

### Before

image

### After

image

## Test methodology 
Manual

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -33,6 +33,7 @@ namespace GitUI.Editor
 
         private readonly TranslationString _error = new TranslationString("Error");
         private readonly TranslationString _largeFileSizeWarning = new TranslationString("This file is {0:N1} MB. Showing large files can be slow. Click to show anyway.");
+        private readonly TranslationString _cannotViewImage = new TranslationString("Cannot view image {0}");
 
         public event EventHandler<SelectedLineEventArgs> SelectedLineChanged;
         public event EventHandler HScrollPositionChanged;
@@ -813,21 +814,25 @@ namespace GitUI.Editor
                 return _async.LoadAsync(getImage,
                             image =>
                             {
-                                ResetForImage(fileName);
-                                if (image != null)
+                                if (image == null)
                                 {
-                                    var size = DpiUtil.Scale(image.Size);
-                                    if (size.Height > PictureBox.Size.Height || size.Width > PictureBox.Size.Width)
-                                    {
-                                        PictureBox.SizeMode = PictureBoxSizeMode.Zoom;
-                                    }
-                                    else
-                                    {
-                                        PictureBox.SizeMode = PictureBoxSizeMode.CenterImage;
-                                    }
+                                    ResetForText(null);
+                                    internalFileViewer.SetText(string.Format(_cannotViewImage.Text, fileName), openWithDifftool);
+                                    return;
                                 }
 
-                                PictureBox.Image = image == null ? null : DpiUtil.Scale(image);
+                                ResetForImage(fileName);
+                                var size = DpiUtil.Scale(image.Size);
+                                if (size.Height > PictureBox.Size.Height || size.Width > PictureBox.Size.Width)
+                                {
+                                    PictureBox.SizeMode = PictureBoxSizeMode.Zoom;
+                                }
+                                else
+                                {
+                                    PictureBox.SizeMode = PictureBoxSizeMode.CenterImage;
+                                }
+
+                                PictureBox.Image = DpiUtil.Scale(image);
                                 internalFileViewer.SetText("", openWithDifftool);
                             });
             }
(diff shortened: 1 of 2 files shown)


vbjay:
Why can't we get the file from git and display that?

gerhardol:
> Why can't we get the file from git and display that?

This specific change is "low level", it could be that the file could not be displayed as an image (like a text file renamed to .png).

If there is a change, the logic to have special handling for removed images could be changed.
If so, I would like to do this in a follow up:
 * I would like to get #7825 and a followup that unifies the presentation of "changes" first
 * How should removed images be displayed, so the user understands that an image is removed?
For this reason, removed text files are displayed as a diff.
 * Generally, we spend too much time on discussing these edge situations, instead of just adding something that is good enough - what is that here?

vbjay:
So the example was a removed image. My point is if that failed to show an
image, why can't we show the original file?  Yes. If we get an imageformat
exception or whatnot then by all means show a text error message. If it is
removed, we should write to temp the file and load that. If you are going
to touch that logic then do it right. Why touch it twice.

gerhardol:
> Why touch it twice.

It is another part of the logic, and I have several PRs to clean up that first before making such a change.

GitUI/Editor/FileViewer.cs
```diff
@@ -811,21 +811,25 @@ private Task ViewItemAsync(string fileName, bool isSubmodule, Func<Image> getIma
                 return _async.LoadAsync(getImage,
                             image =>
                             {
+                                if (image == null)
+                                {
+                                    ResetForText(null);
+                                    internalFileViewer.SetText($"Cannot view image {fileName}", openWithDifftool);
```
RussKie: Localise please

gerhardol: Done
Generally I believe translation is overkill for something like this, but I allow it as it is (at least currently) shown for removed images.
Unless there are further comments in the next two days, I will squash and schedule a merge

GitUI/Editor/FileViewer.cs
```diff
@@ -33,6 +34,7 @@ public partial class FileViewer : GitModuleControl
 
         private readonly TranslationString _error = new TranslationString("Error");
         private readonly TranslationString _largeFileSizeWarning = new TranslationString("This file is {0:N1} MB. Showing large files can be slow. Click to show anyway.");
+        private readonly TranslationString _cannot_view_image = new TranslationString("Cannot view image {0}");
```
RussKie: `_cannotViewImage`

gerhardol:
I updated the translation on top of #7877, still fails in master

GitUI/Translation/English.xlf
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="4.0.0">
+<xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="3.3.0">
```
mstv: For the final #7877, it needs to be
```suggestion
<xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="">
```

RussKie:
I think it will fail if merged before...

mstv:
> I think it will fail if merged before...

So fix up the final nit and merge #7877, please, @RussKie. 😀

RussKie:
It won't until late in the day. Feel free amend my PR, and squash-merge

GitUI/Editor/FileViewer.cs
```diff
@@ -20,6 +20,7 @@
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
 using ResourceManager;
+using SmartFormat;
```
RussKie: What's this used for?

gerhardol:
```
private readonly TranslationString _cannotViewImage = new TranslationString("Cannot view image {0}");
<snip>
internalFileViewer.SetText(_cannotViewImage.Text.FormatSmart(fileName), openWithDifftool);
```

gerhardol:
@msftbot merge in 24 hours

GitUI/Editor/FileViewer.cs
RussKie: Thank you

> _cannotViewImage.Text.FormatSmart(fileName)

What does this do?
Why can't we use the standard BCL function `string.Format`?

gerhardol: > Why can't we use the standard BCL function string.Format

Because I preferred the compact format. I could change to:
```
internalFileViewer.SetText(string.Format(_cannotViewImage.Text, fileName), openWithDifftool);
                                    
```

RussKie:
Please do. There is no reason to depend on a 3rd party code here.

PR #7872, opened 2020-03-17
Fixup translation (last master merge)

mstv:
Fixes #7714

## Proposed changes

- update `English.xlf` in order to make the CI build succeed

----


RussKie:
The root cause is that the PR has been built before we started enforcing the translations again. We may have few more PRs like that, and should kick them off to rebuild again before merging.

PR #7877, opened 2020-03-18
fix: Translation update issues

### LINKED ISSUE #7873: Translation update issues
## Current behaviour

The CI build checks for up-to-date `English.xlf`.
But its update is difficult:
- [ ] `TranslationApp` fails because it tries to write the file to the non-existing directory
  `artifacts\bin\TranslationApp\Debug\net461\Translation`
- [ ] The file is generated with a `GitExVersion` which is not accepted by the CI build.
  manual patch required:
``` patch
- <xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="3.3.0">
+ <xliff xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.0" GitExVersion="4.0.0">
```
- [ ] Does `.tools\transifex\UpdateLocalEnglishTranslations.ps1` work?
- [ ] Is `.tools\transifex\` the intended directory for this script?

## Expected behaviour

It is easy for developers to update the translation file.

RussKie:
Resolves #7873

## Proposed changes

* TranslationApp fails because it tries to write the file to the non-existing directory artifacts\bin\TranslationApp\Debug\net461\Translation due to an incorrect working directory

* Everytime trnaslations are updated the current application version is written in xlf files. This creates additional maintenance, yet this functionality is not utilised.
The version is set to empty string.

* Remove now defunct UpdateLocalEnglishTranslations.ps1 script

----


GitExtensions/Project.Loc.targets
```diff
@@ -42,7 +42,7 @@
     <!-- 2. Run TranslationApp.exe to update English*.xlf -->
     <Exec Command="$(_TranslationCommand)"
           LogStandardErrorAsError="false"
-          WorkingDirectory="$(GitExtensionsOutputPath)"
+          WorkingDirectory="$(_GitExtensionsOutputDir)"
```
mstv: This seems to make no difference when the `TranslationApp` is started from Visual Studio.
I do _not_ see a need to make this work because:

I found that `build.cmd -loc` does everything necessary well.
So you could mention this command in `CONTRIBUTING.md`.

RussKie: > This seems to make no difference when the `TranslationApp` is started from Visual Studio.

Ah! `TranslationApp` must be run from the `GitExtensions` folder where `Translation` folder it.
To run it from VS try setting the working directory and pass "update" parameter.

These targets aren't invoked from VS, so this change won't be visible. It is visible from the cli though.

mstv: > To run it from VS try setting the working directory and pass "update" parameter.

First: There is an easy solution, which just needs to be documented. It is not necessary to run the `TranslationApp` in VS.

Second: Neither nor. `TranslationApp` does not use command line arguments. The working directory does not matter at all:
```
        public static string GetTranslationDir()
        {
            return Path.Combine(Path.GetDirectoryName(typeof(Translator).Assembly.Location), "Translation");
        }
```

RussKie: > First: There is an easy solution, which just needs to be documented. It is not necessary to run the `TranslationApp` in VS.

Correct.
`TranslationApp` is a really outdated artifact, which we are stuck with (until we can refactor it out, it is very complicated though). 

> Second: Neither nor. `TranslationApp` does not use command line arguments.

Huh, indeed! Those were removed in https://github.com/gitextensions/gitextensions/pull/6840 by some RussKie fellow...

> The working directory does not matter at all:

You're likely correct, but it just looks incorrect to start the app from the wrong folder.

RussKie:
I have pushed few variant of an update to the README, I'm happy with the last one. Let me know what you think.

README.md
```diff
@@ -61,12 +57,23 @@ For more details please refer to [versions comparison](https://github.com/gitext
   </tr>
 </table>
 
+### Build instructions
+The solution can be build from Visual Studio or from a command line. You will still need Visual Studio for the UI work, but for everything else you can use your favourite IDE (such as Visual Studio Code).
+* From Visual Studio simply set "GitExtensions" as a start project and <kbd>F5</kbd>.
+* From command line run one of the following commands. For more switches refer to the [build script](https://github.com/gitextensions/gitextensions/blob/master/scripts/build.ps1).
+  *  `build.cmd` to build the solution
+  *  `build.cmd -test -integrationTest` to build the solution and run tests
+  *  `build.cmd -loc` to update English translations, if you have changed those
```
mstv:
```suggestion
  *  `build.cmd -loc` to update English translations if you have changed those
```

RussKie:
> In addition, I'd like to have a short comment in the `CONTRIBUTING.md` at line 61, e.g.

I have updated CONTRIBUTING.md and https://github.com/gitextensions/gitextensions/wiki/Translations

RussKie:
...also updated https://github.com/gitextensions/gitextensions/wiki/How-To:-build-instructions

vbjay:
... can be built...

README.md
```diff
@@ -11,14 +11,10 @@ Have a question? Come and talk to us: [![Gitter](https://badges.gitter.im/Join%2
...
-The work has been [split into two streams](https://github.com/gitextensions/gitextensions/wiki/Compare-versions):
- * version 3.x - **[Download it now](https://github.com/gitextensions/gitextensions/releases/latest)** or install it with [Chocolatey](https://chocolatey.org/packages/gitextensions) or [AppGet](https://appget.net/packages/i/gitextensions).
- * version 2.5x
 
-### Version 3.x
 
-This stream contains the code for the new major version with a number of significant changes&mdash;both to the app and the codebase.
-For more details please refer to [versions comparison](https://github.com/gitextensions/gitextensions/wiki/Compare-versions).
+
+## Version 3.x
```
mstv:
```suggestion
### Version 3.x
```

mstv:
Thank you.

PR #7878, opened 2020-03-18
Centralise reference version management

RussKie:


Plugins/Bitbucket/Bitbucket.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
     <PackageReference Include="RestSharp" Version="106.2.1" />
```
RussKie: Unable to use `$(RestSharpVersion)` due to API breaking changes. This will need to be investigated separately.

gerhardol: I can look at that later

RussKie: Thank you!

UnitTests/GitUI.Tests/GitUI.Tests.csproj
```diff
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
-    <PackageReference Include="LibGit2Sharp" Version="0.25.0" />
```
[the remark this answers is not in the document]
RussKie: It is added indirectly via the common test project

gerhardol:
For BitBucket, the following seem to work:
-            if (RequestBody != null)
+            if (RequestBody != null && RequestBody is string && ((string)RequestBody) != "")
             {
-                if (RequestBody is string)
-                {
-                    request.AddParameter("application/json", RequestBody, ParameterType.RequestBody);
-                }
-                else
-                {
-                    request.AddBody(RequestBody);
-                }
+                request.AddJsonBody(RequestBody);
             }

RequestBody is a json for CreatePullRequest and "" for Approve/Merge, null otherwise
The "" return should be changed to null too (and the quick patch be simplified).

PR #7881, opened 2020-03-19
BitBucket plugin update of RestSharp

### LINKED ISSUE #7855: Fail to load plugin assembly dependencies (here `RestSharp`)
## Current behaviour

The version of the `RestSharp` required by the `Git.hub` project end up in the main folder of GitExtensions (because the `Git.hub` is a dependency of the `GitUI` project).
And because of that it prevent the loading of the `RestSharp` dependency of the `BitBucket` and `JiraCommitHintPlugin` plugins because:

* the versions are different
* it tries to find the assemblies in the main folder before the plugin folders

I imagine that these plugins are broken for that reason even if I can't verify because I don't use these them

I think **it shows that we have a problem with loading plugins dependencies when the same dependency with a different version is used by the main application** (or another plugin with the same dependency that appears before in the `Probing` property of the `.config` file)

## Expected behaviour

The `restsharp` assemblies of the plugins loads without problem

## Steps to reproduce

- Enable all `Common language Runtime Exceptions` in Exceptions Settings in VisualStudio (except `TaskCanceledException` because that makes the debugging impossible)
- Run the application in debug.

I don't know.

## Environment

- GitExtensions version: `master`

## Diagnostics

???
Align versions? Use binding redirects?

gerhardol:
See https://github.com/gitextensions/gitextensions/pull/7878#issuecomment-601216437
Fixes #7855

## Proposed changes

Update plugin for newer RestSharp. Add .json body with supported methods.

## Test methodology 
Test to a private BitBucket Server (not working with BitBucket Cloud).

----

### DIFF
--- a/Plugins/Bitbucket/ApprovePullRequest.cs
+++ b/Plugins/Bitbucket/ApprovePullRequest.cs
@@ -13,7 +13,7 @@ namespace Bitbucket
             _info = info;
         }
 
-        protected override object RequestBody => "";
+        protected override object RequestBody => null;
 
         protected override Method RequestMethod => Method.POST;
 
--- a/Plugins/Bitbucket/Bitbucket.csproj
+++ b/Plugins/Bitbucket/Bitbucket.csproj
@@ -2,7 +2,7 @@
 
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
-    <PackageReference Include="RestSharp" Version="106.2.1" />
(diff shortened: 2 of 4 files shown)


pmiossec:
Should also (partly?) fix #7855

gerhardol:
> Should also (partly?) fix #7855

I see #7878 as the main resolver, this just fixes the final piece. But I will update the description

RussKie:
This will obsolete #7855.

gerhardol:
@msftbot merge in 24 hours

Issue #7882, opened 2020-03-21
Show branch name in recent repos drop down


joergpichler:
I am working on a big project and have the same repo multiple times on my developer machine. I am working on different branches in parallel. In order to make it easier to track in which repo which branch is current checked out it would make sense to (optionally) show the branch name in the recent repos drop down menu.

If you want to check it out I have a branch ready in my fork of gitextensions.

gerhardol:
The current workflow is to Close Browse and go to Dashboard, so same number of clicks
But I would find this useful, some subrepos I work with are cloned in about 50 clones on my work computer and it is hard to remember which superrepo I worked on.
There may be objections if the menu is too cluttered though and GE already have too many options.

I suggest that you open a PR for this, to possibly get comments. A preliminary version not ready for a review to your own repo with a link here and more or less final to gitextensions with WIP in the title.

vbjay:
Read up on worktrees.

gerhardol:
> Read up on worktrees.

Still you want to know what worktree 'clone' that uses a specific branch
For the submodule in worktree scenario, the submodule is unfortunately always a separate repo

joergpichler:
I created a PR here https://github.com/gitextensions/gitextensions/pull/7883

gerhardol:
Closed with #7883, even if there are renewed discussions after the merge

RussKie:
I just realised that we shouldn't have used "[ ]" because it is inconsistent with the existing formatting of branches. We based it off the submodules, but it is submodules that should have been fixed in the first place (because they came later).

We use "( )" in the following places:
* the title
  ![image](https://user-images.githubusercontent.com/4403806/78621626-63b6bf00-78c6-11ea-888d-5eeb5bd040a1.png)
* format "no branch" everywhere
  ![image](https://user-images.githubusercontent.com/4403806/78621724-9e205c00-78c6-11ea-8185-15f9dc895c53.png)
  ![image](https://user-images.githubusercontent.com/4403806/78621742-aa0c1e00-78c6-11ea-80f8-81cd5639dc45.png)
  ![image](https://user-images.githubusercontent.com/4403806/78621756-b3958600-78c6-11ea-82c8-c4140b7df5aa.png)


Secondly, we should also render "(no branch)" in the dropdown as well. Otherwise it looks very weird that some repos have branch information, and some don't.

gerhardol:
I rather change the title, [no branch]

[...]

PR #7883, opened 2020-03-21
Add setting to show branch name in recent repos drop down

joergpichler:
Issue: https://github.com/gitextensions/gitextensions/issues/7882

## Proposed changes

-  Show branch name in recent repos drop down (optional)

### Before

grafik

grafik

grafik

### After

grafik

grafik

grafik

## Test methodology 
- Manual

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1409,7 +1409,7 @@ namespace GitCommands
             }
         }
 
-        public static bool DashboardShowCurrentBranch
+        public static bool ShowRepoCurrentBranch
         {
             get => GetBool("dashboardshowcurrentbranch", true);
             set => SetBool("dashboardshowcurrentbranch", value);
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1802,6 +1802,16 @@ namespace GitUI.CommandsDialogs
 
             void AddRecentRepositories(Repository repo, string caption)
             {
+                if (AppSettings.ShowRepoCurrentBranch)
+                {
+                    var branchName = GitModule.GetSelectedBranchFast(repo.Path);
+
+                    if (!string.IsNullOrEmpty(branchName) && branchName != DetachedHeadParser.DetachedBranch)
+                    {
+                        caption = $"{caption} [{branchName}]";
+                    }
+                }
+
                 if (!string.IsNullOrEmpty(repo.Category))
                 {
                     caption = $"{caption} ({repo.Category})";
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
@@ -36,6 +36,7 @@
             this.truncatePathMethod = new System.Windows.Forms.ComboBox();
             this.truncateLongFilenames = new System.Windows.Forms.Label();
             this.chkEnableAutoScale = new System.Windows.Forms.CheckBox();
+            this.chkShowRepoCurrentBranch = new System.Windows.Forms.CheckBox();
             this.chkShowCurrentBranchInVisualStudio = new System.Windows.Forms.CheckBox();
             this.gbLanguages = new System.Windows.Forms.GroupBox();
             this.tlpnlLanguage = new System.Windows.Forms.TableLayoutPanel();
@@ -114,15 +115,17 @@
             this.tlpnlGeneral.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
             this.tlpnlGeneral.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Percent, 100F));
             this.tlpnlGeneral.Controls.Add(this.chkShowRelativeDate, 0, 0);
-            this.tlpnlGeneral.Controls.Add(this.chkShowCurrentBranchInVisualStudio, 0, 1);
-            this.tlpnlGeneral.Controls.Add(this.chkEnableAutoScale, 0, 2);
-            this.tlpnlGeneral.Controls.Add(this.chkSortByAuthorDate, 0, 3);
-            this.tlpnlGeneral.Controls.Add(this.truncateLongFilenames, 0, 4);
-            this.tlpnlGeneral.Controls.Add(this.truncatePathMethod, 1, 4);
+            this.tlpnlGeneral.Controls.Add(this.chkShowRepoCurrentBranch, 0, 1);
+            this.tlpnlGeneral.Controls.Add(this.chkShowCurrentBranchInVisualStudio, 0, 2);
+            this.tlpnlGeneral.Controls.Add(this.chkEnableAutoScale, 0, 3);
+            this.tlpnlGeneral.Controls.Add(this.chkSortByAuthorDate, 0, 4);
+            this.tlpnlGeneral.Controls.Add(this.truncateLongFilenames, 0, 5);
+            this.tlpnlGeneral.Controls.Add(this.truncatePathMethod, 1, 5);
             this.tlpnlGeneral.Dock = System.Windows.Forms.DockStyle.Fill;
             this.tlpnlGeneral.Location = new System.Drawing.Point(8, 21);
             this.tlpnlGeneral.Name = "tlpnlGeneral";
-            this.tlpnlGeneral.RowCount = 5;
+            this.tlpnlGeneral.RowCount = 6;
+            this.tlpnlGeneral.RowStyles.Add(new System.Windows.Forms.RowStyle());
             this.tlpnlGeneral.RowStyles.Add(new System.Windows.Forms.RowStyle());
             this.tlpnlGeneral.RowStyles.Add(new System.Windows.Forms.RowStyle());
             this.tlpnlGeneral.RowStyles.Add(new System.Windows.Forms.RowStyle());
@@ -142,6 +145,18 @@
             this.chkShowRelativeDate.TabIndex = 0;
             this.chkShowRelativeDate.Text = "Show relative date instead of full date";
             this.chkShowRelativeDate.UseVisualStyleBackColor = true;
+            // 
+            // chkShowRepoCurrentBranch
+            // 
+            this.chkShowRepoCurrentBranch.AutoSize = true;
+            this.tlpnlGeneral.SetColumnSpan(this.chkShowRepoCurrentBranch, 2);
+            this.chkShowRepoCurrentBranch.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.chkShowRepoCurrentBranch.Location = new System.Drawing.Point(3, 26);
+            this.chkShowRepoCurrentBranch.Name = "chkShowRepoCurrentBranch";
+            this.chkShowRepoCurrentBranch.Size = new System.Drawing.Size(314, 17);
+            this.chkShowRepoCurrentBranch.TabIndex = 5;
+            this.chkShowRepoCurrentBranch.Text = "Show current branch names in the dashboard and the recent repositories dropdown menu";
+            this.chkShowRepoCurrentBranch.UseVisualStyleBackColor = true;
             // 
             // chkShowCurrentBranchInVisualStudio
             // 
@@ -527,6 +542,7 @@
         private System.Windows.Forms.LinkLabel helpTranslate;
         private System.Windows.Forms.GroupBox gbGeneral;
         private System.Windows.Forms.CheckBox chkShowRelativeDate;
+        private System.Windows.Forms.CheckBox chkShowRepoCurrentBranch;
         private System.Windows.Forms.CheckBox chkShowCurrentBranchInVisualStudio;
         private System.Windows.Forms.CheckBox chkEnableAutoScale;
         private System.Windows.Forms.Label truncateLongFilenames;
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
@@ -67,6 +67,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         {
             chkEnableAutoScale.Checked = AppSettings.EnableAutoScale;
 
+            chkShowRepoCurrentBranch.Checked = AppSettings.ShowRepoCurrentBranch;
             chkShowCurrentBranchInVisualStudio.Checked = AppSettings.ShowCurrentBranchInVisualStudio;
             _NO_TRANSLATE_DaysToCacheImages.Value = AppSettings.AvatarImageCacheDays;
             ShowAuthorAvatarInCommitInfo.Checked = AppSettings.ShowAuthorAvatarInCommitInfo;
@@ -127,6 +128,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         {
             AppSettings.EnableAutoScale = chkEnableAutoScale.Checked;
             AppSettings.TruncatePathMethod = GetTruncatePathMethodString(truncatePathMethod.SelectedIndex);
+            AppSettings.ShowRepoCurrentBranch = chkShowRepoCurrentBranch.Checked;
             AppSettings.ShowCurrentBranchInVisualStudio = chkShowCurrentBranchInVisualStudio.Checked;
             AppSettings.ShowAuthorAvatarColumn = ShowAuthorAvatarInCommitGraph.Checked;
             AppSettings.ShowAuthorAvatarInCommitInfo = ShowAuthorAvatarInCommitInfo.Checked;
--- a/GitCommands/Submodules/SubmoduleStatusProvider.cs
+++ b/GitCommands/Submodules/SubmoduleStatusProvider.cs
@@ -182,7 +182,7 @@ namespace GitCommands.Submodules
             {
                 var name = submodule;
                 string path = result.Module.GetSubmoduleFullPath(submodule);
-                if (AppSettings.DashboardShowCurrentBranch && !GitModule.IsBareRepository(path))
+                if (AppSettings.ShowRepoCurrentBranch && !GitModule.IsBareRepository(path))
                 {
                     name = name + " " + GetModuleBranch(path, noBranchText);
                 }
@@ -290,7 +290,7 @@ namespace GitCommands.Submodules
 
         private string GetBranchNameSuffix(string repositoryPath, string noBranchText)
         {
-            if (AppSettings.DashboardShowCurrentBranch && !GitModule.IsBareRepository(repositoryPath))
+            if (AppSettings.ShowRepoCurrentBranch && !GitModule.IsBareRepository(repositoryPath))
             {
                 return " " + GetModuleBranch(repositoryPath, noBranchText);
             }
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/Dashboard.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/Dashboard.cs
@@ -23,7 +23,6 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
         private readonly TranslationString _issues = new TranslationString("Issues");
         private readonly TranslationString _openRepository = new TranslationString("Open repository");
         private readonly TranslationString _translate = new TranslationString("Translate");
-        private readonly TranslationString _showCurrentBranch = new TranslationString("Show current branch");
 
         private DashboardTheme _selectedTheme;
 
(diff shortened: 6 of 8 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1803,6 +1803,11 @@ private void PopulateRecentRepositoriesMenu(ToolStripDropDownItem container)
 
             void AddRecentRepositories(Repository repo, string caption)
             {
+                if (AppSettings.RecentReposShowBranchName)
+                {
+                    caption = $"[{GitModule.GetSelectedBranchFast(repo.Path)}] {caption} ";
```
gerhardol: I suggest the name is put after the path

gerhardol:
Translations will need to be updated for the build to be successful, #7877 is fixing the setup and description

GitUI/CommandsDialogs/BrowseDialog/FormRecentReposSettings.Designer.cs
```diff
@@ -131,18 +132,20 @@ private void InitializeComponent()
...
-            tableLayoutPanel1.RowCount = 7;
+            tableLayoutPanel1.RowCount = 8;
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
             tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle(System.Windows.Forms.SizeType.Absolute, 20F));
+            tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle());
```
mstv: These two lines might have to be swapped.
Could you add screenshots of this form, please?

GitCommands/Settings/AppSettings.cs
```diff
@@ -1451,6 +1451,12 @@ public static int RecentReposComboMinWidth
             set => SetInt("RecentReposComboMinWidth", value);
         }
 
+        public static bool RecentReposShowBranchName
+        {
+            get => GetBool("RecentReposShowBranchName", false);
```
mstv: I would set this setting default to `true`. Thoughts?
How expensive is `GitModule.GetSelectedBranchFast`?

GitUI/CommandsDialogs/FormBrowse.cs
mstv: Basically, this information must be updated -- both for the active repo and for those of other GE instances. Is it?
How expensive is `GitModule.GetSelectedBranchFast`?

GitUI/CommandsDialogs/BrowseDialog/FormRecentReposSettings.Designer.cs
joergpichler: Sure here you go
grafik

GitCommands/Settings/AppSettings.cs
joergpichler: Setting it to true was my first intention too but then I was not sure about forcing new things on the users

GitUI/CommandsDialogs/FormBrowse.cs
joergpichler: Will do

joergpichler:
> 
> 
> Translations will need to be updated for the build to be successful, #7877 is fixing the setup and description

Does that just mean that this PR will just have to wait until the one you linked is merged?

GitUI/CommandsDialogs/FormBrowse.cs
joergpichler: > 
> 
> Basically, this information must be updated -- both for the active repo and for those of other GE instances. Is it?
> How expensive is `GitModule.GetSelectedBranchFast`?

I implemented a very simple loop and called it 1000 times on my Ryzen 2500U with an SSD. Profiled it with DotTrace and seems like it takes ~200 microseconds for one call. I am no expert in profiling though...

mstv:
> Translations will need to be updated for the build to be successful, #7877 is fixing the setup and description

It means, you should rebase after it has been merged and to run `build.cmd -loc` in order to update `English.xlf`. 

In addition, you need to subscribe in `contributors.txt`. Do it in a separate PR in order to avoid merge conflicts.

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: The call parses the file system to find the head. So expect it to require hundreds of ms in the worst case. Probably acceptable.

gerhardol:
#7877 should be merged soon

mstv:
#7877 has been merged. So please rebase, run `Build.cmd -loc` and commit `English.xlf`.

pmiossec:
>#7877 has been merged. So please rebase, run Build.cmd -loc and commit English.xlf.

And also squash your commits while you're at it... 

And also don't forget the PR with your update on 'contributors.txt' file.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1803,6 +1803,11 @@ private void PopulateRecentRepositoriesMenu(ToolStripDropDownItem container)
 
             void AddRecentRepositories(Repository repo, string caption)
             {
+                if (AppSettings.RecentReposShowBranchName)
+                {
+                    caption = $"{caption} [{GitModule.GetSelectedBranchFast(repo.Path)}]";
+                }
```
RussKie: ⚠️  
1. `GitModule.GetSelectedBranchFast` may return `string.Empty`, showing "caption []" won't look good.
2. The caption is modified below as well, it won't look good at all

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1803,6 +1803,11 @@ private void PopulateRecentRepositoriesMenu(ToolStripDropDownItem container)
...
+
                 if (!string.IsNullOrEmpty(repo.Category))
                 {
                     caption = $"{caption} ({repo.Category})";
```
RussKie: As a side note

😕 I don't believe this piece of code is working at all.
The collection of recent repositories doesn't have "category" set, only the collection of favourites does, but it isn't bound here...

RussKie:
This whole dialog must be moved into Settings

GitUI/CommandsDialogs/FormBrowse.cs
joergpichler: Will fix that it wont show just brackets without content

GitUI/CommandsDialogs/FormBrowse.cs
joergpichler: You are right, category is always null here

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: GetSelectedBranchFast returns "(no branch)"/DetachedHeadParser.DetachedBranch if not set, string.Empty is not used (just open a GE subrepo where a branch is not checked out, the dafault to see this).
This should also be excluded.

joergpichler: Done

GitUI/CommandsDialogs/BrowseDialog/FormRecentReposSettings.Designer.cs
```diff
@@ -153,7 +156,7 @@ private void InitializeComponent()
             this.comboMinWidthNote.Location = new System.Drawing.Point(11, 246);
             this.comboMinWidthNote.Name = "comboMinWidthNote";
             this.comboMinWidthNote.Padding = new System.Windows.Forms.Padding(12, 0, 0, 0);
-            this.comboMinWidthNote.Size = new System.Drawing.Size(300, 73);
+            this.comboMinWidthNote.Size = new System.Drawing.Size(300, 30);
```
RussKie: Why?

RussKie:
I almost forgot we had this form. And it brings horrible memories... I want to completely nuke it, remove "most recent"/"less recent" anchoring, and move the remaining functionality into Settings.

GitUI/CommandsDialogs/BrowseDialog/FormRecentReposSettings.Designer.cs
joergpichler: This magically fixed the UI, the combobox is never rendered with this height anyways.

With height 73:
grafik
Height 30:
grafik

RussKie:
Great, thank you

mstv:
Works well with my local repos. Have not tested with repos on network shares (I know about the risks).
Seeing the dropdown near the title calls for consistency:

grafik

So use round parantheses, please.
Perhaps, the branch names could be aligned using `"\t"` (tab).
```C#
caption = $"{caption}\t({branchName})";
```

mstv:
Any objections to turn this setting on by default?

gerhardol:
"(no branch)" is hidden now, good

> So use round parantheses, please.

Square is used in the side panel, should be consistent

> Any objections to turn this setting on by default?

No, activate

mstv:
> Square is used in the side panel, should be consistent.

👍 Yes, I missed the display of submodules in the left panel and the tooltips in the RevisionGrid.
So title bar is the exception.

RussKie:
We should make it consistent.

Btw we also show branches in the dashboard. And there is a setting that
controls that.
(It just occurred to me).

Should we use *that* setting across the lot?

gerhardol:
> Should we use *that* setting across the lot?

Probably
I have one argument against that though: Branches are retrieved every time the menu is opened while the dashboard is opened more in the background. If getting the branch is slow in some setup, then it is good to separate them. But add that later?

RussKie:
Aye

It would be good to measure... not quite sure what would be a good test
though

gerhardol:
> It would be good to measure... not quite sure what would be a good test though

A Git repo on a network drive slow enough to be just not unusable - is this change making a major difference?

Reuse the current setting

joergpichler:
> Reuse the current setting

Should the user still be able to toggle that setting from the Recent Repo Settings UI or only through dashboard?

>  Perhaps, the branch names could be aligned using "\t" (tab).

Tried it. Tabs are never rendered in MenuItems

RussKie:
The setting should reside in Settings. Not sure which tab without looking.

RE: tabs - even if they rendered, paths can be arbitrary lengths, won't
look good.

joergpichler:
> The setting should reside in Settings. Not sure which tab without looking. 

I could not find it there.

I am talking about AppSettings.DashboardShowCurrentBranch. 

Looking at the code it is only set from the Dashboard Menu.

RussKie:
Correct, it is not there. We need to add it there 🙄

The name is too specific though, and we should at least rename the property
to something like `AppSettings.ShowCurrentBranch`.
We can leave the underlying setting name, though I personally don't see it
as a major drama to rename it too.

gerhardol:
> Should the user still be able to toggle that setting from the Recent Repo Settings UI or only through dashboard?

> The setting should reside in Settings. Not sure which tab without looking.

The Dashboard setting is hard to find, as well as recent menu.
The common settings is searchable, but there is no natural tab now
Suggest Appearance, where a similar VS setting is now

image

joergpichler:
> The Dashboard setting is hard to find, as well as recent menu.
> The common settings is searchable, but there is no natural tab now
> Suggest Appearance, where a similar VS setting is now
Keep the Dashboard Settings anyways or shall it be removed?

gerhardol:
> Keep the Dashboard Settings anyways or shall it be removed?

I suggest remove, @RussKie to decide

A lot of comments for a minor change, but that is not unusual for UI changes...

RussKie:
The UI defines the UX, so it can be contentious...

Remove the setting, please.
I wouldn't consider it a major loss. May cause a little frustration, but I'm ok with that.

joergpichler:
grafik

joergpichler:
> Remove the setting, please.
> I wouldn't consider it a major loss. May cause a little frustration, but I'm ok with that.

Tbh until today I wasn't even aware of that setting at all

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
```diff
@@ -143,6 +146,18 @@ private void InitializeComponent()
...
+            // chkShowRepoCurrentBranch
+            // 
+            this.chkShowRepoCurrentBranch.AutoSize = true;
+            this.tlpnlGeneral.SetColumnSpan(this.chkShowRepoCurrentBranch, 2);
+            this.chkShowRepoCurrentBranch.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.chkShowRepoCurrentBranch.Location = new System.Drawing.Point(3, 26);
+            this.chkShowRepoCurrentBranch.Name = "chkShowRepoCurrentBranch";
+            this.chkShowRepoCurrentBranch.Size = new System.Drawing.Size(314, 17);
+            this.chkShowRepoCurrentBranch.TabIndex = 5;
+            this.chkShowRepoCurrentBranch.Text = "Show current branch of repository";
```
gerhardol: Too generic, impossible to understand what this changes. Not happy with the following either:
Show current branch in dashboard and dropdown menu

joergpichler: I am not happy with it either. Is there a reason for anyone not not show the branch at all? I mean maybe remove this toggle for the user completly?

gerhardol: Some want it minimalistic, maybe performance (easier to add an option than to make good performance tests)

mstv:
> Tried it. Tabs are never rendered in MenuItems.

Thank you. Tabs are used by the Win32 API and some older GUI libs on top of it in order to display shortcut keys (really aligned). We could misuse `ToolStripMenuItem.ShortcutKeyDisplayString` for displaying the branch names.

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
joergpichler: Would adding a tooltip make sense?

gerhardol: Yes
After this, I hope there are no more discussions here...

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
```diff
@@ -143,6 +148,20 @@ private void InitializeComponent()
...
+            this.chkShowRepoCurrentBranch.AutoSize = true;
+            this.tlpnlGeneral.SetColumnSpan(this.chkShowRepoCurrentBranch, 2);
+            this.chkShowRepoCurrentBranch.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.chkShowRepoCurrentBranch.Location = new System.Drawing.Point(3, 26);
+            this.chkShowRepoCurrentBranch.Name = "chkShowRepoCurrentBranch";
+            this.chkShowRepoCurrentBranch.Size = new System.Drawing.Size(314, 17);
+            this.chkShowRepoCurrentBranch.TabIndex = 5;
+            this.chkShowRepoCurrentBranch.Text = "Show current branch of repositories";
+            this.toolTip.SetToolTip(this.chkShowRepoCurrentBranch, "Shows the branch name of the repositories in the dashboard and the recent reposit" +
+        "ories drop down menu");
```
mstv: I wonder why this tooltip does not appear in the `English.xlf`.

gerhardol: Not a translation string, should be changed. TranslationApp do not find such strings

pmiossec: I already warned about that: https://github.com/gitextensions/gitextensions/issues/3745#issuecomment-306492000

joergpichler: Fixed

RussKie:
Please update the screenshots

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
```diff
@@ -17,6 +17,7 @@ public partial class AppearanceSettingsPage : SettingsPageWithHeader
         private readonly TranslationString _noDictFilesFound = new TranslationString("No dictionary files found in: {0}");
         private readonly TranslationString _noImageServiceTooltip = new TranslationString("A default image, if an email address has no matching Gravatar image.\r\nSee http://en.gravatar.com/site/implement/images#default-image for more details.");
         private readonly TranslationString _authorDateSortWarningTooltip = new TranslationString("Sorting by author date may delay rendering of the revision graph.");
+        private readonly TranslationString _showRepoCurrentBranchTooltip = new TranslationString("Shows the branch name of the repositories in the dashboard and the recent repositories drop down menu");
```
RussKie:
```suggestion
        private readonly TranslationString _showRepoCurrentBranchTooltip = new TranslationString("Shows current branch names for repositories in the dashboard and the recent repositories dropdown menu");
```

GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.Designer.cs
```diff
@@ -143,6 +146,18 @@ private void InitializeComponent()
...
+            // chkShowRepoCurrentBranch
+            // 
+            this.chkShowRepoCurrentBranch.AutoSize = true;
+            this.tlpnlGeneral.SetColumnSpan(this.chkShowRepoCurrentBranch, 2);
+            this.chkShowRepoCurrentBranch.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.chkShowRepoCurrentBranch.Location = new System.Drawing.Point(3, 26);
+            this.chkShowRepoCurrentBranch.Name = "chkShowRepoCurrentBranch";
+            this.chkShowRepoCurrentBranch.Size = new System.Drawing.Size(314, 17);
+            this.chkShowRepoCurrentBranch.TabIndex = 5;
+            this.chkShowRepoCurrentBranch.Text = "Show current branch of repositories";
```
RussKie:
```suggestion
            this.chkShowRepoCurrentBranch.Text = "Shows current branch names in the dashboard and the recent repositories dropdown menu";
```

joergpichler: Tooltip is obsolote then I guess since it says the same thing?

Also I would suggest just using 'Show' instead of 'Shows' here to make it consistent with 'Show current branch in Visual Studio'

joergpichler:
Since label of checkbox and tooltip were the same after your proposed changes I removed the tooltip since it does not provide additional information

joergpichler:
> 
> 
> Why have these changed?
> image

Good question...I can only guess they got modified locally because I had some issue with build.cmd -loc this morning with the code open & running from VS...
I committed them by accident. Fixed now

RussKie:
Please rebase cleanly on top of the latest master,  and revert all
unnecessary changes.

RussKie:
Thank you

mstv:
What do you think about (right-)aligned branch names?

grafik

pmiossec:
>What do you think about (right-)aligned branch names?

Probably better, indeed...

RussKie:
I can see a problem with that - if a user has a very long path, then the gap becomes rather difficult to navigate visually:
image

Make that list a 100 items long, and it becomes a real accessibility issue.

RussKie:
I've been running it for a day, and I can't say I'm finding this feature useful. I find it more an annoyance because it makes it harder for me to see repo names:
image

In the dashboard the branches are displayed in a different font and color, that makes it easy to visually tell text apart:
image


And I'm starting to wonder whether @mstv suggestion would indeed make it more usable. Or whether this feature is necessary... 🤔

pmiossec:
>And I'm starting to wonder whether @mstv suggestion would indeed make it more usable.

I also of think it will make things much more readable by reducing a lot the clutter.

> I can see a problem with that - if a user has a very long path, then the gap becomes rather difficult to navigate visually:
>Make that list a 100 items long, and it becomes a real accessibility issue.

If someone is in this case, he surely better enable the "shorten strategy" to use "The most signifiant directory" to make the use of "recent dropdown" and "Dashboard" usable....

image

Am I the only one to set this setting just after a GitExtensions install? 

That looks like: 
image
and with the branches:

image

mstv:
> Am I the only one to set this setting just after a GitExtensions install?

I have not noticed this dialog before this PR.

jrfnl:
Hi all,

I really like the _idea_ of this feature, but in practical terms I find it completely unusable, unfortunately.

* I have the "shortening strategy" set to "middle parts with dots" as "significant directory" shows too little/confusing information. Think: several repos with the same name in different directories - only the name is not usable in that case.
    If "significant directory" would show "parent/dir", things would already be improved.
* At time, branches may have long names with "notes to self". Think: "WAIT/NEXT-DEP-X-RELEASE/feature/real-branch-name" to indicate to myself that the branch is not yet pulled because it's waiting for a tagged dependency release.

With this setup - the drop-down becomes so wide, that it automatically moves to another screen, making it hard to reach.

Unfortunately that means, that I had to turn off the branch name setting, which means they are also no longer available in the dashboard, decreasing usability of the dashboard yet again.

Such a shame as, as I said before, I very much like the _idea_ of this feature.

gerhardol:
This PR retrieves the branch name in the UI thread. This adds a delay of 3.5s for me to open the menu the first time. Due to file caching, subsequent openings are faster.
It took some time to investigate this, it has annoyed me for two years, without realizing it was this PR...

PR #7887, opened 2020-03-22
WIP DiffTool for large files

### LINKED ISSUE #7841: Commit dialog F3 shortcut for "open with difftool" is ignored when changed file size is "large"
## Current behaviour

1. Select a changed file in the commit dialog that is over the file size threshold that triggers the "This file is x MB. Showing large files can be slow. Click to show anyway" prompt.
1. Hit the "Open with difftool" shortcut key (F3).
1. Nothing happens. However, when I right click on the file and click "Open with difftool" from the context menu the difftool is launched as expected.

## Expected behaviour
The difftool is launched for large files when I hit F3.

## Steps to reproduce

See **Current behaviour**. This behaviour can be reproduced consistently.

## Screenshots

image

I don't know whether this has ever worked as expected but I have experienced the behaviour in previous versions (sorry I can't be more specific).

## Environment

- Git Extensions 3.3.1.7897
- Git 2.24.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #7841 
See also alternative/complement #7891 that also resolves the original issue

## Proposed changes
A few forms uses the FileViewer OpenWithDifftool Action to start the difftool.
The result if the form internal handler or the FileViewer implementation is used
should be the same, starting difftool from GitModule

This is the case for the HotKey (F3) handling in FormCommit and both
HotKey and context menu for RevisionFileTreeControl

The difftool must be set also when displaying the "empty" message.

An alternative correction is maybe to remove the difftool handling completely in FileViewer, only support in the forms.
(I will not do that now, will conflict with other changes)

## Test methodology

Only tested in FormCommit, should be the same in RevisionFileTreeControl
It is difficult to get a 5MB diff (no such file in GE repo?),
I patched the limit to a lower value in FileViewer.cs:501

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -423,7 +423,8 @@ namespace GitUI.Editor
                     getImage: GetImage,
                     getFileText: GetFileText,
                     getSubmoduleText: () => LocalizationHelpers.GetSubmoduleText(Module, fileName.TrimEnd('/'), ""),
-                    openWithDifftool));
+                    openWithDifftool),
+                openWithDifftool);
 
             Image GetImage()
             {
@@ -470,9 +471,9 @@ namespace GitUI.Editor
             }
         }
 
-        private Task ShowOrDeferAsync(string fileName, Func<Task> showFunc)
+        private Task ShowOrDeferAsync(string fileName, Func<Task> showFunc, Action openWithDifftool)
         {
-            return ShowOrDeferAsync(GetFileLength(), showFunc);
+            return ShowOrDeferAsync(GetFileLength(), showFunc, openWithDifftool);
 
             long GetFileLength()
             {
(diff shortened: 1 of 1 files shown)


RussKie:
> But it will need an improved hotkey handling because of the double use of `F3`: `find next` and `open with difftool`.

I prefer to keep <kbd>F3</kbd> for "Find next"

RussKie:
tbh, I'm quite lost. I think being a light user compared to you I don't experience all these use cases.
Could you describe these use cases and steps to repro, so I could get a better appreciation of the pain points and the proposals to address them?

RussKie:
Ignore me...

gerhardol:
>> But it will need an improved hotkey handling because of the double use of F3: find next and open with difftool.

> I prefer to keep F3 for "Find next"

Not about hotkeys, it is a common implementation of the OpenDiffTool action in FileViewer that other forms could use. This would make some kind of sense for viewing changes, the arguments are the same and revisions etc are provided anyway when viewing changes. The FileViewer implementation was hardly used though, so the FileViewer bug was only seen in FormCommit. RevisionFileTreeControl does not view changes, so that usage was incorrect anyway (which is hard to see from the interfaces).

I will merge #7891 first as it unifies the diff tool usage in FormCommit and add a feature to RevisionFileTreeControl.cs, and think some more about this change (or rather to remove difftool in FileViewer). With #7891, this functionality is likely not used and may potentially be removed - or use this and remove all custom OpenDiffTool implementations.

gerhardol:
Closing for now, will maybe reopen later

PR #7891, opened 2020-03-23
FormCommit: F3 to diff large files

### LINKED ISSUE #7841: Commit dialog F3 shortcut for "open with difftool" is ignored when changed file size is "large"

gerhardol:
Fixes #7841 
Alternative or complement to #7887

## Proposed changes

Use 'internal' difftool command instead of FileViewer to view hot key diffs

RevisionFileTree tried to use FileViewer difftool, but the action was
never enabled (so the menu option was always disabled). Instead invoke Module.DiffTool,
compare displayed item to worktree, something I have occasionally missed.
(Note that if there is no diff, difftool will not open, not changed, so to test this select an older commit).

#7887 fixes the problem in FileViewer for FormDiff. 
That PR need to set difftool in ViewGitItemAsync() (by default, use explicit or use ViewChangesAsync() to set a difftool in FileViewer)

With this change #7887 is maybe not needed and FileViewer.DiffTool could possibly be removed.
There is still some use of difftool with F3, to investigate later.

## Test methodology 

See #7887 

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -796,7 +796,7 @@ namespace GitUI.CommandsDialogs
                 case Command.ShowHistory: return StartFileHistoryDialog();
                 case Command.ToggleSelectionFilter: return ToggleSelectionFilter();
                 case Command.StageAll: return StageAllFiles();
-                case Command.OpenWithDifftool: SelectedDiff.OpenWithDifftool?.Invoke(); return true;
+                case Command.OpenWithDifftool: OpenWithDiffTool(); return true;
                 case Command.OpenFile: openToolStripMenuItem.PerformClick(); return true;
                 case Command.OpenFileWith: openWithToolStripMenuItem.PerformClick(); return true;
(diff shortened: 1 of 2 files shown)


gerhardol:
@msftbot merge in 1 day

gerhardol:
Rebased for merge

PR #7895, opened 2020-03-25
Remove Gerrit plugin

RussKie:
The plugin has been moved out to https://github.com/gitextensions/GitExtensions.GerritPlugin

----


RussKie:
Thank you for noticing.
I'm tackling the installer separately, that work prompted it.

PR #7900, opened 2020-03-25
Simulate continuous scroll to display revision diff

### LINKED ISSUE #3600: Concatenate diffs of all changed files in a commit

gerhardol:
Fixes #3600 

All credits to @pmiossec , I have just done some adjustments

## Proposed changes

The next file is automatically selected when top or bottom of the file is reached.

The continuous scrolling is enabled by default. It is possible to disable this functionality from a context menu item, see screenshot. It is also possible to scroll continuously by pressing ALT when scrolling. While there is enough settings in GE already, the menu item is a way to document this and discover the functionality. (a tool tip could be added but that is annoying.)

Note: Prevent file change too fast by limiting events to maximum every 600ms (this can be changed by editing the settings file, but no UI).

Icon from https://p.yusukekamiyamane.com/icons/search/fugue/#keyword=scroll

### After

(wrap around removed, scroll up will start at bottom of file)

contscroll

image

image

## Test methodology
Manual

----


pmiossec:
@gerhardol I like what you've done with the last commit. The feature is really satisfying like it is now...

👍

RussKie:
Suggest adding the same icon to the toolbar in the diff window

gerhardol:
> Suggest adding the same icon to the toolbar in the diff window

I rather not, for several reasons:
* The toolbar is crowded already
* The setting will at least need a tooltip
* The setting is relevant for new files too and the toolbar is only available for diffs (show non-printing characters and syntax highlighting would be relevant for new files though)
* You do normally not change this so often, it is a static preference

There is no natural form in Settings (If implemented for all forms with filelists, it could be added to DiffViewer). Maybe context menu makes it easier to discover the feature too.

RussKie:
> It is possible to disable this functionality from a context menu item, see screenshot. It is also possible to scroll continuously by pressing ALT when scrolling.

I propose to keep just one setting - scroll or not scroll, configurable from Settings and from the context menu. Many different ways to achieve the same outcome create a cognitive load.
Your point about the floating menu is valid, I accept that.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -49,6 +49,25 @@ public RevisionDiffControl()
...
+            DiffText.TopScrollReached += DiffText_TopScrollReached;
+            DiffText.BottomScrollReached += DiffText_BottomScrollReached;
+        }
+
+        private void DiffText_TopScrollReached(object sender, EventArgs e)
+        {
+            if (Control.ModifierKeys == Keys.Alt || AppSettings.ContinuousScrollToNextFile)
+            {
+                DiffFiles.SelectPreviousVisibleItem();
+                DiffText.ScrollToBottom();
```
RussKie: Does this flicker?

GitUI/Editor/FileViewer.cs
```diff
@@ -321,6 +329,22 @@ public void ReloadHotkeys()
             Hotkeys = HotkeySettingsManager.LoadHotkeys(HotkeySettingsName);
         }
 
+        private void _internalFileViewer_BottomScrollReached(object sender, EventArgs e)
+        {
+            if (BottomScrollReached != null)
+            {
+                BottomScrollReached(sender, e);
+            }
+        }
```
RussKie:
```suggestion
        private void _internalFileViewer_BottomScrollReached(object sender, EventArgs e)
            => BottomScrollReached?.Invoke(sender, e);
```

GitUI/Editor/FileViewer.cs
```diff
@@ -321,6 +329,22 @@ public void ReloadHotkeys()
...
+
+        private void _internalFileViewer_TopScrollReached(object sender, EventArgs e)
+        {
+            if (TopScrollReached != null)
+            {
+                TopScrollReached(sender, e);
+            }
+        }
```
RussKie:
```suggestion
        private void _internalFileViewer_TopScrollReached(object sender, EventArgs e)
            => TopScrollReached?.Invoke(sender, e);
```

GitUI/Editor/FileViewer.cs
```diff
@@ -1618,5 +1648,10 @@ public bool ShowSyntaxHighlightingInDiff
             public ToolStripButton IgnoreAllWhitespacesButton => _fileViewer.ignoreAllWhitespaces;
             public ToolStripMenuItem IgnoreAllWhitespacesMenuItem => _fileViewer.ignoreAllWhitespaceChangesToolStripMenuItem;
         }
+
+        public void ScrollToBottom()
+        {
+            internalFileViewer.ScrollToBottom();
+        }
```
RussKie:
```suggestion
        public void ScrollToBottom()
            => internalFileViewer.ScrollToBottom();
```

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -201,8 +205,60 @@ public async Task FindNextAsync(bool searchForwardOrOpenWithDifftool)
...
+            var isScrollingTowardBottom = e.Delta < 0;
+
+            if (isScrollingTowardTop && (TextEditor.ActiveTextAreaControl.VScrollBar.Value == 0))
+            {
+                CallTopScrollReachedEvent(sender, e);
+            }
+
+            if (isScrollingTowardBottom && (!TextEditor.ActiveTextAreaControl.VScrollBar.Visible
+                                            || TextEditor.ActiveTextAreaControl.VScrollBar.Value + TextEditor.ActiveTextAreaControl.VScrollBar.Height >
+                                            TextEditor.ActiveTextAreaControl.VScrollBar.Maximum))
```
RussKie: Keep on the same line, it is the same expression
```suggestion
                                            || TextEditor.ActiveTextAreaControl.VScrollBar.Value + TextEditor.ActiveTextAreaControl.VScrollBar.Height > TextEditor.ActiveTextAreaControl.VScrollBar.Maximum))
```

If you find it too long, create a variable on line:216 and reuse it across the method:
```cs
var vScrollBar = TextEditor.ActiveTextAreaControl.VScrollBar;
```

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -246,6 +302,18 @@ public void SetText(string text, Action openWithDifftool, bool isDiff = false)
...
 
             _currentViewPositionCache.Restore(isDiff);
+
+            if (_shouldScrollToBottom)
+            {
+                if (TextEditor.ActiveTextAreaControl.VScrollBar.Visible)
+                {
+                    TextEditor.ActiveTextAreaControl.VScrollBar.Value =
+                        Math.Max(0, TextEditor.ActiveTextAreaControl.VScrollBar.Maximum -
+                        TextEditor.ActiveTextAreaControl.VScrollBar.Height - _bottomBlankHeight);
```
RussKie: Ditto, keep on the same line
```suggestion
                        Math.Max(0, TextEditor.ActiveTextAreaControl.VScrollBar.Maximum - TextEditor.ActiveTextAreaControl.VScrollBar.Height - _bottomBlankHeight);
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1084,6 +1084,82 @@ private void HandleVisibility_NoFilesLabel_FilterComboBox(bool filesPresent)
...
+                item.Selected = true;
+                item.EnsureVisible();
+            }
+
+            ListViewItem selectedItemFound = null;
+            for (int i = FileStatusListView.Groups.Count - 1; i >= 0; i--)
+            {
+                var group = FileStatusListView.Groups[i];
+                var groupItems = FileStatusListView.Items.Cast<ListViewItem>().Where(item => item.Group == group)
+                    .Reverse();
```
```suggestion
                var groupItems = FileStatusListView.Items
                								   .Cast<ListViewItem>()
                								   .Where(item => item.Group == group)
                    							   .Reverse();
```

GitUI/Editor/FileViewer.cs
gerhardol: ok

GitUI/Editor/FileViewer.cs
gerhardol: ok

GitUI/Editor/FileViewerInternal.cs
gerhardol: have to call it scrollBar though to get Hungarian notation error...

GitUI/Editor/FileViewerInternal.cs
gerhardol: ok

GitUI/UserControls/FileStatusList.cs
gerhardol: ok, same on 1144

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: Yes, very briefly a little extra.
As the file is already updated which appears as a kind of flicker itself, it is not really noticed in itself.

You can get the entire PR to look ugly if (like it is a screen2gif) if you speed view the changes in a commit.
If you actually read one line in every file, it is a nice functionality..

pmiossec:
>I propose to keep just one setting - scroll or not scroll, configurable from Settings and from the context menu. Many different ways to achieve the same outcome create a cognitive load.
Your point about the floating menu is valid, I accept that.

@RussKie Let me disagree with you... 

* From a setting point of view, the presence of the 'alt' feature change nothing. There is just 1 on/off setting. 
* From the point of view of a user that can't have this feature enabled all the time (and I'm sure that I won't be the only one), I can assure that history of 'cognitive load' don't stand more than 5 seconds.
I find it **much much much** more boring (or adding a cognitive load for the user) to have to enable the feature, then use it, and finally, disable it just after (that lead to not use it even if I want to). 
The 'alt' key just add a 2nd way more flexible to use this feature. It is just advertised by the label... 

Another thing... I find it very frustrating to have spent time on this feature to be able to use it and finally won't be able to do so because of a final choice like that. Please, spend more time to understand my point of view and reconsider your decision.

gerhardol:
Updated comments and rebased. Will squash and initiate merge at approval

> I propose to keep just one setting - scroll or not scroll, configurable from Settings and from the context menu

Only context menu now. There is no natural place to put this in Settings.
It could be put in DiffViewer, but that is for all diff forms. This could of course be implemented in FormCommit, CommitInfo, FormCompare (where else?) that uses the FileStatusList.
  
> Many different ways to achieve the same outcome create a cognitive load.

It is only one setting, to activate the scroll with ALT or not
I could drop the ALT, but @pmiossec did not like that, he wanted to control this, spin the mousewheel occasionally, see separate post.
He would on the other hand be happy with only using ALT, then no setting is needed at all. But I would have to use one hand on the mouse, the other on the keyboard.
The setting is a way to advertise the functionality, can be hard to find.

The label/setting should maybe be inverted

"Continuous scroll to next file without ALT button"
->
"Continuous scroll to next file only with ALT button"

> I see a lot of new functionality, but no tests of any kind.
This worries me. This is a very visible part of the app, and if we miss something, it will cause a lot of pain to our users.

This is one UI detail, not more or less than a button. What do you propose is tested, that scrolling to next file works is still working?
I do not see that as time effective to make UI tests for this functionality.

mstv:
I vote for keeping the setting in the context menu because it makes this feature discoverable.
I will use it with `ALT` only I guess.

gerhardol:
Squashed and rebased

> I propose to keep just one setting - scroll or not scroll, configurable from Settings and from the context menu

Is Settings requested and is this functionality if so requested in all forms using FileStatusList?

pmiossec:
>is this functionality if so requested in all forms using FileStatusList?

For consistency, I would say yes...

Another thing I spotted: The scroll doesn't work on image file. I have added the support in my branch. Feel free to cherrypick.

RussKie:
> Another thing I spotted: The scroll doesn't work on image file.

If scroll from a text file to an binary (e.g. an image) file  - show it, then on further scroll jump to the next file?

gerhardol:
>> is this functionality if so requested in all forms using FileStatusList?

> For consistency, I would say yes...

I will look at this and add the setting

> Another thing I spotted: The scroll doesn't work on image file. I have added the support in my branch. Feel free to cherrypick.

Thanks, applied (rebased too)

> If scroll from a text file to an binary (e.g. an image) file - show it, then on further scroll jump to the next file?

As GE scales images to fit the current size of the viewer, there is no scrollbar and scrolling is similar to a small file (the builtin delay of 600 ms)

gerhardol:
Updated, will squash

The context menu is added for all viewers, also those without a FileStatusList

gerhardol:
The common string for Settings and context menu is put in Strings.cs
Howerver, the translation app also adds the usages to the .xlf

```
      <trans-unit id="chkContScrollToNextFileOnlyWithAlt.Text">
        <source>Continuous scroll to next file only with ALT button</source>
        <target />
      </trans-unit>

      <trans-unit id="chkContScrollToNextFileOnlyWithAlt.Text">
        <source>Continuous scroll to next file only with ALT button</source>
        <target />
      </trans-unit>
```

gerhardol:
Squashed on master

GitUI/Editor/ContinuousScrollEventManager.cs
```diff
@@ -0,0 +1,39 @@
...
+        public EventHandler BottomScrollReached;
+        public EventHandler TopScrollReached;
+        public DateTime LastScrollEventFiredDate { get; set; } = DateTime.MinValue;
+
+        private bool IsScrollTooFast()
+        {
+            return DateTime.Now - LastScrollEventFiredDate < TimeSpan.FromMilliseconds(AppSettings.AutoViewScrollDelay);
+        }
+
+        public void CallBottomScrollReachedEvent(object sender, EventArgs e)
```
RussKie, line 17:
```suggestion
        public void RaiseBottomScrollReached()
```
This aligns with Windows Forms vocabulary

GitUI/Editor/ContinuousScrollEventManager.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (IsScrollTooFast())
+            {
+                return;
+            }
+
+            LastScrollEventFiredDate = DateTime.Now;
+            BottomScrollReached?.Invoke(sender, e);
+        }
+
+        public void CallTopScrollReachedEvent(object sender, EventArgs e)
```
RussKie, line 28:
```suggestion
        public void RaiseTopScrollReached()
```

GitUI/Editor/ContinuousScrollEventManager.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public void CallBottomScrollReachedEvent(object sender, EventArgs e)
+        {
+            if (IsScrollTooFast())
+            {
+                return;
+            }
+
+            LastScrollEventFiredDate = DateTime.Now;
+            BottomScrollReached?.Invoke(sender, e);
```
RussKie, line 25:
```suggestion
            BottomScrollReached?.Invoke(this, EventArgs.Empty);
```

GitUI/Editor/ContinuousScrollEventManager.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public void CallTopScrollReachedEvent(object sender, EventArgs e)
+        {
+            if (IsScrollTooFast())
+            {
+                return;
+            }
+
+            LastScrollEventFiredDate = DateTime.Now;
+            TopScrollReached?.Invoke(sender, e);
```
RussKie, line 36:
```suggestion
            TopScrollReached?.Invoke(this, EventArgs.Empty);
```

GitUI/Editor/ContinuousScrollEventManager.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using GitCommands;
+
+namespace GitUI.Editor
+{
+    public class ContinuousScrollEventManager
```
RussKie, line 6:
```suggestion
    internal sealed class ContinuousScrollEventManager
```

GitUI/Editor/FileViewer.cs
```diff
@@ -1618,5 +1660,10 @@ public bool ShowSyntaxHighlightingInDiff
             public ToolStripButton IgnoreAllWhitespacesButton => _fileViewer.ignoreAllWhitespaces;
             public ToolStripMenuItem IgnoreAllWhitespacesMenuItem => _fileViewer.ignoreAllWhitespaceChangesToolStripMenuItem;
         }
+
+        public void ScrollToBottom()
```
RussKie, line 1664: [nit] move public before privates and handlers

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -203,6 +212,28 @@ public async Task FindNextAsync(bool searchForwardOrOpenWithDifftool)
         public event EventHandler VScrollPositionChanged;
         public new event EventHandler TextChanged;
 
+        private void TextArea_MouseWheel(object sender, MouseEventArgs e)
```
RussKie, line 215: [nit] move the handler after privates

GitUI/Strings.cs
```diff
@@ -32,6 +32,7 @@ internal sealed class Strings : Translate
         private readonly TranslationString _local = new TranslationString("Local");
         private readonly TranslationString _tag = new TranslationString("Tag");
         private readonly TranslationString _remote = new TranslationString("Remote");
+        private readonly TranslationString _contScrollToNextFileOnlyWithAlt = new TranslationString("Continuous scroll to next file only with ALT button");
```
RussKie, line 35: Maybe?
```suggestion
        private readonly TranslationString _contScrollToNextFileOnlyWithAlt = new TranslationString("Enable continuous scroll by holding ALT button");
```

GitUI/Editor/ContinuousScrollEventManager.cs
gerhardol: ok

GitUI/Editor/ContinuousScrollEventManager.cs
gerhardol: ok

GitUI/Editor/ContinuousScrollEventManager.cs
gerhardol: ok

GitUI/Editor/ContinuousScrollEventManager.cs
gerhardol: ok

GitUI/Editor/FileViewerInternal.cs
gerhardol: ok

GitUI/Strings.cs
gerhardol: The functionality do not require ALT by default (but you could of course use the key), no key in addition to the mousewheel.
This option is to only use the scrolling with ALT and I believe the suggestion is misleading (but current text is not refined

Important to use something reasonable, functionality can be adapted to what is easiest to define here...

GitUI/Editor/ContinuousScrollEventManager.cs
gerhardol: Must be public, FileViewerInternal.cs SetContinuousScrollManager() uses the class (that method could be set internal too)

RussKie: No worries

GitUI/Strings.cs
RussKie: In this case I propose we reword to
```suggestion
        private readonly TranslationString _contScrollToNextFileOnlyWithAlt = new TranslationString("Enable continuous scroll without ALT button");
```
and invert the setting. I wouldn't like the automatic continuous scroll enabled by default, it is confusing.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1688,6 +1688,18 @@ public static bool BlameShowOriginalFilePath
             set => SetBool("Blame.ShowOriginalFilePath", value);
         }
 
+        public static bool ContinuousScrollToNextFileOnlyWithAlt
+        {
+            get => GetBool("continuousscrolltonextfileonlywithalt", false);
+            set => SetBool("continuousscrolltonextfileonlywithalt", value);
+        }
```
RussKie, line 1695:
```suggestion
        public static bool AutomaticContinuousScroll
        {
            get => GetBool("DiffViewer.AutomaticContinuousScroll", false);
            set => SetBool("DiffViewer.AutomaticContinuousScroll", value);
        }
```

GitCommands/Settings/AppSettings.cs
```diff
@@ -1688,6 +1688,18 @@ public static bool BlameShowOriginalFilePath
...
+
+        public static int AutoViewScrollDelay
+        {
+            get => GetInt("autoviewscrolldelay", 600);
+            set => SetInt("autoviewscrolldelay", value);
+        }
```
RussKie, line 1701:
```suggestion
        public static int AutomaticContinuousScrollDelay
        {
            get => GetInt("DiffViewer.AutomaticContinuousScrollDelay", 600);
            set => SetInt("DiffViewer.AutomaticContinuousScrollDelay", value);
        }
```

This setting is not exposed anywhere, why are we making it a setting?

gerhardol:
Adapting suggestion, rebased (no squash).

> I scroll slowly yet persistently. On the scroll down, the middle file "received" the scrollbars, and it was present on the screen for some observable time. On the way up I scrolled with little pauses - the middle file was rendered without scrollbars and, as a result, disappears almost immediately replaced by the following file...

This PR do not manipulate the scrollbars, except when scrolling upwards it sets the bottom position, but nothing to do with hiding or showing the scrollbar.
There is also no "intercept" of scrollbar events, they are always handled in the ICSharpCode (I assume).
When reaching the end (or beginning) this change checks if at least 600 ms has passed since last switching file. If you do not stop scrolling immediately, I assume that scroll events may be passed to the editor early and a toolbar is added (temporarily).

> On the scroll down, the middle file "received" the scrollbars, and it was present on the screen for some observable time. 

This has not bothered me when viewing files.
I added a commit to "jump to start" similar to when scrolling up. (There is a quick flicker.)
Not sure I want to have this myself, separate commit.

> On the way up I scrolled with little pauses - the middle file was rendered without scrollbars and, as a result, disappears almost immediately replaced by the following file...

The 600 ms delay can be increased if you consider this to be too short. I probably want to decrease this time myself. GE do not need more options, but this is why made this tunable in the settings file.

If the default is off, I see nothing blocking from merging this, also if you are not fully satisfied with the functionality.

GitCommands/Settings/AppSettings.cs
RussKie: I suppose I can see the value in this.

gerhardol:
Squashed changes, rebased. Plan to merge in about a day (will let other changes merge first, probably rebase on master).

gerhardol:
@msftbot merge in 14 h

gerhardol:
@msftbot merge in 14 hours

PR #7904, opened 2020-03-27
Add root description to Git node in settings tree view

### LINKED ISSUE #5946: settings, arrow navigation in the tree
**minor bug**

If I use the up/down arrows in the settings dialog when the tree has focus, I get stuck going up at the Git node, which forces the selection to its first child when it gets selected (which makes sense, until it got focus from an up key *from* its first child).

Open settings; note that navigating down the tree with arrow keys works; note that navigating back up gets stuck.

Up from first child of Git would select the tree-previous or previous sibling of Git, rather than re-selecting the first child.

- Git Extensions 3.00.00.4433
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

Dunno - never tried it before.

cado1982:
Fixes #5946 

## Proposed changes

- Add root node to Git tree view node on Settings page.
- I simply copied the root page from the Plugins section, renamed it and changed the label.

### Before

Clicking in the Git node automatically selects the first child node
Annotation 2020-03-27 142130

### After

Annotation 2020-03-27 142129

## Test methodology

- Using the up/down keys on the tree view to make sure navigation still works
- Can still save Git settings

## Test environment(s)

- GIT 2.19.1
- Windows 10 Pro

----


RussKie:
Please update English translations and re-submit the pull-request. Refer to https://github.com/gitextensions/gitextensions/wiki/Translations

mstv:
@cado1982, I guess you must add your signature in `contributors.txt` yet.

cado1982:
@RussKie Updated translations.
@mstv I added my signature to contributors.txt in this PR. Do I need to do something else?

RussKie:
Thank you

PR #7910, opened 2020-03-28
Collect and publish plugins (portable)

RussKie:
Addendum to #7649, solves parts of #7702.

## Proposed changes

* Partially alter #7850 to not rename any plugins for CI builds
* Enumerate all plugin satellite files
* Publish plugins in the portable artifact

----

### DIFF
--- a/GitExtensions/Project.Publish.targets
+++ b/GitExtensions/Project.Publish.targets
@@ -71,7 +71,7 @@
       <_PublishedPath>$([MSBuild]::NormalizeDirectory('$(PublishDir)', '..'))</_PublishedPath>
     </PropertyGroup>
 
-    <!-- Determine plugins assemblies and thier necessary references -->
+    <!-- Determine plugins assemblies and their required references -->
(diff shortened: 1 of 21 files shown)


RussKie:
I would too, but in #7850 we addressed a perf regression, that led to
renaming plugins.
Whilst it's working well locally, it is a PITA to change the installer.

Speaking of which, I'm currently working on the installer.
I want to release asap, so we can start migrating to .NET Core, where I
expect a lot will change, including bazillion of netstandard dependencies
gone.

gerhardol:
Can we generate the files for the installer?

Go ahead if it is needed.

RussKie:
> Can we generate the files for the installer?

If you're curious, very much WIP: https://github.com/RussKie/gitextensions/tree/Fix_msi

RussKie:
Looks like it broke the build, and the failure hasn't been reported... 😒 

https://ci.appveyor.com/project/gitextensions/gitextensions/builds/31786768

PR #7911, opened 2020-03-28
Fix msi

RussKie:
Addendum to #7649, solves parts of #7702.

## Proposed changes

- reinstate the ability to generate an MSI

----


gerhardol:
No objections - if a .msi is created, it works...

RussKie:
🎉 https://ci.appveyor.com/api/buildjobs/bpfrn2nx91gysyyo/artifacts/artifacts%2Fpublish%2FGitExtensions-4.0.0.8915-543b66c05.msi

RussKie:
Updated: https://ci.appveyor.com/api/buildjobs/kkwqoetpnom0giyx/artifacts/artifacts%2Fpublish%2FGitExtensions-4.0.0.8927-0c0dc6cdf.msi

One wicked issue I have with the msi when install - Git.hub.dll is disappearing. On repair, it comes back...
And with that some of the plugins don't show up in the Plugins menu.

I have no idea...

@Johno-ACSLive would you have any ideas?

Johno-ACSLive:
@RussKie I had a quick look and there doesn't appear to be anything obvious. In terms of Git.Hub.dll. Cracked open the MSI and had a look at the tables and looks fine.

Let me generate some log files with GitExtensions-4.0.0.8927-0c0dc6cdf.msi and see what I find. I assume the issue occurs for fresh install as well as upgrade?

In terms of plugins not loading, the only thing I can think of is:

- Are there dependencies from anything that was cleaned up?

- There are multiple source locations for the same DLL but they don't appear to have changed in the installer source location. Not sure which files are being copied to the output location during build so potentially there is a version conflict.

Johno-ACSLive:
I've just performed some testing, on fresh install of GitExtensions-4.0.0.8927-0c0dc6cdf.msi the files appear to be installed correctly (I didn't check all - just Git.hub.dll).

After uninstalling that version and installing version 3.3 I then performed an upgrade and the issue is now observed. I checked both the Fresh install log and upgrade and everything is good except during the upgrade, it never actually performs the file install but the log indicates it should. Other files in the log do other things in the upgrade (e.g. removal of old files etc.) but yeah I can't see anything out of the ordinary.

Did you use a newer version of WiX to build the MSI? Perhaps there is a bug in the version used if that has changed.

RussKie:
Thank you for the tests, at least I'm not alone in this...

I've used Wix 3.11.0, which we used before. I think there's a newer
version, I'll try it....

Johno-ACSLive:
I have 3.11.1 installed but yeah 3.11.2 is the latest.

Lets see if that works, if not I'll need to speed up the Advanced Installer work.

RussKie:
Update Wix to 3.11.2, same effect... https://ci.appveyor.com/api/buildjobs/01o3wtt3sucqhap0/artifacts/artifacts%2Fpublish%2FGitExtensions-4.0.0.8987-2a488ecc7.msi 😒 

I tried seeing in Process Monitor, but nothing suspicious:
image
Except the last entry where the file is not found.

There is "ACCESS DENIED" entry, but every other assembly has the same entry, yet only this assembly is missing...
image

I have also checked the antivirus logs (ESET Internet Security) - nothing suspicious either.

I'm at loss. From the point of the installer, we haven't touched entries for this assembly.

Johno-ACSLive:
> I have also checked the antivirus logs (ESET Internet Security) - nothing suspicious either.
> 
> I'm at loss. From the point of the installer, we haven't touched entries for this assembly.

Agree, a fresh install does not have the problem so I don't believe AV to be at fault.

I just checked the file version of Git.hub.dll in both MSI's.

- GitExtensions-3.3.0.7719.msi the file version for Git.hub.dll is 3.3.0.7719
- GitExtensions-4.0.0.8927-0c0dc6cdf.msi the file version for Git.hub.dll is 1.0.0.0

If the file version is the same, MSI will usually skip and upgrade if its newer. If it's older, then depending on the config it will either skip or downgrade. In this case the file is removed but not downgraded.

This is likely the bug!

Compile Git.hub.dll and ensure the file version is higher. Generate a new MSI and this should fix the problem.

RussKie:
> I just checked the file version of Git.hub.dll in both MSI's.
> 
> * GitExtensions-3.3.0.7719.msi the file version for Git.hub.dll is 3.3.0.7719
> * GitExtensions-4.0.0.8927-0c0dc6cdf.msi the file version for Git.hub.dll is 1.0.0.0
> 
> If the file version is the same, MSI will usually skip and upgrade if its newer. If it's older, then depending on the config it will either skip or downgrade. In this case the file is removed but not downgraded.
> 
> This is likely the bug!

Mate, did I tell you're a legend! I should have thought of looking at the dll property.
Indeed the version wasn't set correctly, and now when it is fixed - it worked!
https://ci.appveyor.com/api/buildjobs/1nrvyla7v7d5xtcj/artifacts/artifacts%2Fpublish%2FGitExtensions-4.0.0.9005-6de00f1d7.msi

Johno-ACSLive:
Great to hear it worked! 😀

PR #7912, opened 2020-03-29
RevDiff: follow selection only in first list group

### LINKED ISSUE #7907: Improvement of retained file selection vs. two commit diff / merge diff
## Current behaviour

The combination of #7798 and of the #4154 implementation selects files in these additional diff groups. This can be unexpected and misleading, especially if the diff group headings are not visible due to many files.

The selection jumps to the second diff group of a merge diff when the previous commit was the left parent.

I guess the match is based only on the file name and ignores the path.

## Expected behaviour

Only matches in the first diff group are reselected.
The file path must match.

## Steps to reproduce

- select a merge commit, e.g. 37f4f9326da1ce72dafb468ab42a86e4d1eda947
- Ctrl+P, Ctrl+N

## Environment

- Git Extensions 4.0.0.8859
- Build 3bb59459d39c3fe086da40cf59994eb775562014 (Dirty)
- Git 2.24.1.windows.2
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #7907

## Proposed changes

Do not try to retain the selected item, if selection was done in
other than the first group or if the item is only appearing in other than the first group

See the issue for a description and discussion. I personally prefer the current behavior, but @mstv has requested this change. (This also conflicts with several other open PRs.)
The retain/follow for RevDiff was introduced in #7798 

## Screenshots 

Showing the behavior after only

i7907-retain-file

## Test methodology 

Code review, manual

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -400,6 +400,24 @@ namespace GitUI
                 .Where(i => i.Group?.Tag is GitRevision)
                 .Select(i => new GitItemStatusWithParent(i.Group.Tag<GitRevision>(), i.Tag<GitItemStatus>()));
 
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [Browsable(false)]
+        public IEnumerable<GitItemStatus> FirstGroupItems
+        {
+            get
+            {
+                if (FileStatusListView.Groups.Count == 0)
+                {
+                    yield break;
+                }
+
+                foreach (ListViewItem item in FileStatusListView.Groups[0].Items)
+                {
+                    yield return item.Tag<GitItemStatus>();
+                }
+            }
+        }
+
         [DefaultValue(true)]
         public bool SelectFirstItemOnSetItems { get; set; }
 
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -154,11 +154,12 @@ namespace GitUI.CommandsDialogs
 
         private void SetDiffs(IReadOnlyList<GitRevision> revisions)
         {
-            GitItemStatus oldDiffItem = DiffFiles.SelectedItem;
+            var item = DiffFiles.SelectedItem;
+            var oldDiffItem = DiffFiles.FirstGroupItems.Contains(item) ? item : null;
             DiffFiles.SetDiffs(revisions, _revisionGrid.GetRevision);
 
             // Try to restore previous item
-            if (oldDiffItem != null && DiffFiles.GitItemStatuses.Any(i => i.Name.Equals(oldDiffItem.Name)))
+            if (oldDiffItem != null && DiffFiles.FirstGroupItems.Any(i => i.Name.Equals(oldDiffItem.Name)))
             {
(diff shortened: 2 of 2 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -400,6 +400,19 @@ public IEnumerable<GitItemStatusWithParent> SelectedItemsWithParent
                 .Where(i => i.Group?.Tag is GitRevision)
                 .Select(i => new GitItemStatusWithParent(i.Group.Tag<GitRevision>(), i.Tag<GitItemStatus>()));
 
+        public IEnumerable<GitItemStatus> FirstGroupItems
+        {
+            get
+            {
+                if (FileStatusListView.Groups.Count == 0)
+                {
+                    return new List<GitItemStatus>();
```
```suggestion
                    return Enumerable.Empty<GitItemStatus>();
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -400,6 +400,19 @@ public IEnumerable<GitItemStatusWithParent> SelectedItemsWithParent
...
+                }
+
+                return from ListViewItem item in FileStatusListView.Groups[0].Items select item.Tag<GitItemStatus>();
```
RussKie: Can we write it in a functional way?
```cs
return FileStatusListView.Groups[0].Items.Select(item => item.Tag<GitItemStatus>())
```

Do we need to check for nulls?

gerhardol: ListViewItemCollection does not support Select()
This is what Re# suggested
I can write as a loop instead

GitUI/UserControls/FileStatusList.cs
gerhardol: Not supported
image

RussKie: Try `Array.Empty<GitItemStatus>()`

GitUI/UserControls/FileStatusList.cs
RussKie: Do it as a `foreach` loop, it is faster and less allocations. I think it would look like:
```cs
foreach (ListViewItem item in FileStatusListView.Groups[0].Items)
{
	yield return item.Tag<GitItemStatus>();
}
```

GitUI/UserControls/FileStatusList.cs
gerhardol: Using yield below required use here too: yield brake

GitUI/UserControls/FileStatusList.cs
gerhardol: ok, rebased and pushed

GitUI/UserControls/FileStatusList.cs
```diff
@@ -400,6 +400,22 @@ public IEnumerable<GitItemStatusWithParent> SelectedItemsWithParent
                 .Where(i => i.Group?.Tag is GitRevision)
                 .Select(i => new GitItemStatusWithParent(i.Group.Tag<GitRevision>(), i.Tag<GitItemStatus>()));
 
+        public IEnumerable<GitItemStatus> FirstGroupItems
```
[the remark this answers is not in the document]
gerhardol: done
Squashed

gerhardol:
@msftbot merge in 1 day

PR #7913, opened 2020-03-29
git-status:  use config to set ignoreSubmodules

### LINKED ISSUE #7906: .gitmodules "ignore" setting
## Current behaviour

Adding "ignore = dirty" or "ignore = all" to .gitmodules does not have the desired effect. Submodule is not ignored and is visible in the commit dialog.  However, typing "git status" to console works as expected - submodule is ignored.

## Expected behaviour

Ignore submodule and do not show it in the commit dialog.

## Steps to reproduce

1. Add submodule to your project
2.  Add ignore = dirty or ignore = all to .gitmodules
3. Change something in the submodule
4. You should be able to see changes in the superproject despite step 2.

## Screenshots
image
image

I do not know, but I guess it hadn't worked.

## Environment
- Git Extensions 3.3.1.7897
- Git 2.26.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 144dpi (150% scaling)

gerhardol:
Fixes #7906

## Proposed changes

User configuration of git config diff.ignoreSubmodules need to be overridden,
to not rely on the user having proper configuration. This was done with a
git-status option --ignoreSubmodule=none
However, git-diff was not overridden, so the RevisionDiff tab could be inconsistent with the artificial status and FormCommit

Submodule config can be overridden per submodule, not just globally.
The PR #7906 argued that those configurations should be followed

Allow those overrides by using the config parameter
'-c diff.ignoreSubmodules=none'  instead of the command line option so
user specific overrides can have effect.

Similar for git-diff, if diff.ignoreSubmodules was configured, git-diff would
use the config

## Screenshots 

Showing behavior after, editing .gitmodules

i7906-git-status

## Test methodology 

The existing meaningless tests are updated so similar testing as before.
There are no tests to test that the Git repos are parsed correctly. Such tests are quite difficult to setup and requires a lot of maintenance. 

----

### DIFF
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Runtime.Remoting.Messaging;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using GitCommands.Git;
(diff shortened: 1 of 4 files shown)


gerhardol:
@msftbot merge in 15 hours

s-sixteen:
@gerhardol 
It took me some time to get to this and build GE, sorry. I can confirm your PR fixes this bug

gerhardol:
> It took me some time to get to this and build GE, sorry. I can confirm your PR fixes this bug

Thanks
Portable builds should be available soon again

Issue #7902, opened 2020-03-26
Bug report must correctly handle `System.AggregateException`


RussKie:
## Current behaviour

![image](https://user-images.githubusercontent.com/4403806/77609841-2fadd680-6f75-11ea-9420-fb592ae11a3a.png)

## Expected behaviour

The bug report must correctly process `System.AggregateException` and report all inner exceptions

## Steps to reproduce

Modify the code to throw an `System.AggregateException`.
In my case, my repo got corrupted with HEAD and index files containing `NUL`s. Having repaired the super-repo, I was going through the submodules when this happened.

## Environment

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version: 

## Diagnostics

- Git Extensions 3.3.1.7897
- Build 5a97671645532bcedc443bac7b727f40db47cb5c
- Git 2.21.0.windows.1 (recommended: 2.23.0 or later)
- Microsoft Windows NT 10.0.18363.0
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

May be related to #7570

RussKie:
A way to simulate: replace a good config (make a backup!) of a submodule with [this one](https://github.com/gitextensions/gitextensions/files/4385030/config.txt) and open the submodule

gerhardol:
Duplicate of #6762 ?
I cannot reproduce with the patch file. 
The error message is incorrect (presented as a file), the parsing of git-diff fails. 
Not sure how to present this actually, may consider this after #7899.

![image](https://user-images.githubusercontent.com/6248932/77669955-d39f7e00-6f85-11ea-8829-9d20b2df41d3.png)

```
$ git -c diff.submodule=short -c diff.noprefix=false diff --no-color -M -C -z --name-status
fatal: bad config line 1 in file F:/dev/gc/gitextensions/.git/modules/GitExtensionsDoc/config
fatal: 'git status --porcelain=2' failed in submodule GitExtensionsDoc
```

RussKie:
Try open the submodule.

In my case all git configs and HEADs were nulled out.... nfi why

gerhardol:
> Try open the submodule.

That is reported as a separate exception: Could not load config file: F:\dev\gc\gitextensions\.git\modules\GitExtensionsDoc\config
Handled as an unresolved conflict for some reason

[...]

PR #7899, opened 2020-03-25
RevDiff: Show common files for BASE->A

gerhardol:
Fixes #4154
WIP, based on #7825 
Some cleanups are planned as a follow up to this PR

Thanks to @mdonatas for the issue and an initial implementation. I have not used any of that code, but view the initial implementation.

## Proposed changes

See #4154 for a description, this PR is to improve comparing changes between branches.
#7720 adds the diff from the common BASE to the selected revision. This also shows what files that are unique to the first/selected commits, as well the files that were changed in the same way.

FileStatusList has until now basically handled the files, selected revision and the (possibly multiple) first or parents as separate data, occasionally bundled/unbundled.

To handle the possibly multiple a->b diff pairs, the datastructures are changed to internally use datastructure including the selected revision in GitItemStatusWithParent to allow multiple selected revisions and simplify handling in other forms

To simplify the code review, this has been split in several commits. The last three could be separated to a separate PR.
 * Refactoring to use ObjectId instead of strings in one more situation.
 * The changes to show unique and common diff from BASE to the selected(b) commit mostly requires changes in FileStatusList.
 * Refactor: Prepare to support multiple selected versions. To also show BASE to first, other forms need to be aware of the changes too (really just RevDiff).
 * RevDiff: Show Unique files BASE->first(A). This is separated (one line), as I am not convinced that this information is understood.
 * Refactor: SelectedItems -> SelectedItemsWithParent. Where possible, use the structures with selected/parent revision, instead of implicit information in the FileStatusList. This will be an enabler for other changes.

Some improvement areas (discussed in the issue):
 * Show groups also if no files
 * tooltips to show the contents better
 * Collapse other than first->selected by default
 * (async loading, show results earlier)

### Before

image

### After

The BASE -> first unique files diff is marked below

image

## Test methodology 
Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #7638, opened 2020-01-15
[NBug] Collection was modified; enumeration operation may not exec...


TWJH:
:warning: The sections below must be filled in and this text must be removed or the issue will be closed.

## Error Details
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at GitUI.UserControls.RevisionGrid.RevisionDataGridView.SelectRowsIfReady(Int32 rowCount)
   at GitUI.UserControls.RevisionGrid.RevisionDataGridView.<BackgroundThreadEntry>g__UpdateRowCount|48_2(Int32 row)
   at GitUI.GitUIExtensions.<InvokeAsync>d__12`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitUI.ThreadHelper.<>c__DisplayClass13_0.<<FileAndForget>b__0>d.MoveNext()
```

## Additional information
Deleting local branches.

## Environment
- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.21.0.windows.1 (recommended: 2.22.0 or later)
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.7.3468.0
- DPI 96dpi (no scaling)

gerhardol:
Duplicate of #6924

Latest discussion in #7527

PR #7914, opened 2020-03-29
Separate staged status from GetDiff

gerhardol:
This was done as a part of #7902 but separated

## Proposed changes

Separate the evaluation of getting "staged status" to a separate function and add tests.
The Git documentation mostly uses stage/unstage as the action to move between the indexes, this text uses staged/unstaged to some extent too.

git-status reports if a file is changed in the worktree or index. This is used in FormCommit.
RevisionDiff allows comparing any commit to worktree/index and uses git-diff.
RevDiff allows the user to stage/unstage files if conditions allow.

The problem here is that git-diff reports the differences for the file but not if a file is changed in any of worktree, index or a intermediate commit. Therefore the staged status is calculated separately from commit. So if a file is changed from index to worktree, it must be changed in the index and it can be staged. If a commit other than index is compared to worktree, the staged status is Unknown. If two normal commits are compared, the status is None.

The 'Staged' calculation was bundled with the parsing of git-diff, it was a little confusing why "parent to second" had to be provided in the API, staged status is therefore evaluated explicitly (or provided manually.)

This PR also changes GetSingleDiff() from using string to ObjectId to be able to use ObjectId in the broken out method.
This requires that FormStash will have to do git-rev-parse. So while slightly more efficient in most forms, 40*2 milliseconds extra to open FormStash for me.  

The refactoring also corrected that status was set to None in some situations it should have been set to Unknown, and that the default should be Unset so that was added. The problem does not make a difference as only explicit WorkTree and Index is used, but the assert in GitItemStatus had to be removed.

It could be possible to set staged status only when parsing git-status, and let git-diff dynamically calculate the staged status from the compared commits. This requires a GitItemStatusWithParents as added in #7899 
(I do not propose that this is changed).

If you have read all the way here: This is not an important change, but some of the lengthy explanation will not be needed after the change.

## Test methodology 
Tests added

----

### DIFF
--- a/GitUI/CommandsDialogs/FormArchive.cs
+++ b/GitUI/CommandsDialogs/FormArchive.cs
@@ -160,7 +160,7 @@ namespace GitUI.CommandsDialogs
             else if (checkboxRevisionFilter.Checked)
             {
                 // 1. get all changed (and not deleted files) from selected to current revision
-                var files = UICommands.Module.GetDiffFiles(DiffSelectedRevision.Guid, SelectedRevision.Guid, SelectedRevision.ParentIds.FirstOrDefault()?.ToString()).Where(f => !f.IsDeleted);
+                var files = UICommands.Module.GetDiffFiles(DiffSelectedRevision?.Guid, SelectedRevision?.Guid, StagedStatus.None).Where(f => !f.IsDeleted);
 
                 // 2. wrap file names with ""
                 // 3. join together with space as separator
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -625,26 +625,34 @@ namespace GitCommands
         /// </summary>
         /// <param name="module">The Git module</param>
         /// <param name="statusString">output from the git command</param>
-        /// <param name="firstRevision">from revision string</param>
-        /// <param name="secondRevision">to revision</param>
-        /// <param name="parentToSecond">The parent for the second revision</param>
+        /// <param name="staged">required to determine if <see cref="StagedStatus"/> allows stage/unstage.</param>
         /// <returns>list with the parsed GitItemStatus</returns>
         /// <seealso href="https://git-scm.com/docs/git-diff"/>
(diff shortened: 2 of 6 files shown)


RussKie:
Thank you for the detailed write up 👍.
I think it is worth while to consolidate the code and use a consistent
approach to solving the problems.
Git allows hundreds different ways to achieve the same thing and shoot
yourself in a foot while doing that. It'd be good to pick one and stick
with it.

GitUI/CommandsDialogs/FormArchive.cs
```diff
@@ -160,7 +160,8 @@ private string GetPathArgumentFromGui()
             else if (checkboxRevisionFilter.Checked)
             {
                 // 1. get all changed (and not deleted files) from selected to current revision
-                var files = UICommands.Module.GetDiffFiles(DiffSelectedRevision.Guid, SelectedRevision.Guid, SelectedRevision.ParentIds.FirstOrDefault()?.ToString()).Where(f => !f.IsDeleted);
+                var stagedStatus = GitCommandHelpers.GetStagedStatus(DiffSelectedRevision?.ObjectId, SelectedRevision?.ObjectId, DiffSelectedRevision?.ObjectId);
```
mstv, line 163: In a pure refactoring, I had expected:
```suggestion
                var stagedStatus = GitCommandHelpers.GetStagedStatus(DiffSelectedRevision?.ObjectId, SelectedRevision?.ObjectId, SelectedRevision?.ParentIds?.FirstOrDefault());
```

gerhardol: I believe I had doubt if parents were set for the revision, but DiffSelectedRevision is not always the parent so this could be incorrect
However, comparisions cannot be done to artificial, so StagedStatus is always None

Rebasing and squashing this

RussKie:
I don't have a good understanding of the functionality, so can't comment on it.
From the code stand point - the change looks ok. 

I have pushed a refactor to the test.

gerhardol:
Squashed, refreshed some comments and added a testcase, rebased on #7938 (conflicts)
To be rebased again and merged within a day when #7638 is merged

PR #7922, opened 2020-03-30
FileStatus icon: submodule before status updated

gerhardol:
Seen with #7902

## Proposed changes

- No flags set should have an FileStatusUnknown icon
If a file has no flag set at all, the status should be unknown too.
If an empty ststus is created like with https://github.com/gitextensions/gitextensions/commit/75a70a9921011e76ddc86028a076cd86aec70a4c
the icon was set to FileModified instead.
The flags is not accessible but either IsTracked or IsNew must be set, so it sufficient to check those

- Separate icon for Conflict

- FileStatusList submodule temporary icon
When fixing this, also addressing something that has always bothered me:
The Submodule icon is set to FileModified until the statatus is updated (run async). Normally it is quick so you hardly see it,
but at work the virus known as Symantec slows down Git 10 times in the normal use case and much more occasionally.
It can take 10 sec to update the status, then you get confused by the file icon.
Using SubmoduleManage to show the temporary status without the up/down/dirty decorations.

I looked at doing the same for the side panel and dropdown, but that is optimized and prunes status updates why it is not worth it to show a temporary status.

### Before
i7902-submodule-icon

### After

image

image

i7902-submodule-icon2

## Test methodology

The unknown icon need to be patched to reproduced

For the submodule status, make a change to a submodule (as artificial diffs are not cached) and switch to an artificial commit.
----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -56,6 +56,7 @@ namespace GitUI
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
         public new event EnterEventHandler Enter;
+        private Dictionary<string, int> _stateImageIndexDict;
 
         public FileStatusList()
         {
@@ -71,7 +72,8 @@ namespace GitUI
             SelectFirstItemOnSetItems = true;
 
             FileStatusListView.SmallImageList = CreateImageList();
-            FileStatusListView.LargeImageList = CreateImageList();
+            FileStatusListView.LargeImageList = FileStatusListView.SmallImageList;
+
             FileStatusListView.AllowCollapseGroups = true;
             FileStatusListView.Scroll += FileStatusListView_Scroll;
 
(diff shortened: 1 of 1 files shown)


RussKie:
I want review this

gerhardol:
> Something like:

I added a variant of this, without the enum and using nameof() as key. One benefit of this is that it is easier to find where the image is really used, the logic I changed was hard to find the first time I looked at it. A similar setup is used in the sidepanel, but TreeNode can use a dictionary directly as a key without a translator. (The sidepanel has slightly different logic but similar to FormBrowse, that handles Images directly, so no direct sharing.)

> The function GetItemImageIndex ain't win a beauty price never ever.

Still no price, but much cleanup now.
Not the best return on investment for this change though...

GitUI/UserControls/FileStatusList.cs
```diff
@@ -56,6 +56,7 @@ public sealed partial class FileStatusList : GitModuleControl
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
         public new event EnterEventHandler Enter;
+        private IDictionary<string, int> _stateImages;
```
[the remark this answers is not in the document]
gerhardol: ok

gerhardol:
Updated with a separate icon for conflicts

GitUI/UserControls/FileStatusList.cs
```diff
@@ -92,31 +94,45 @@ ImageList CreateImageList()
...
+                    (nameof(Images.SubmoduleRevisionDown), Images.SubmoduleRevisionDown),
+                    (nameof(Images.SubmoduleRevisionDownDirty), Images.SubmoduleRevisionDownDirty),
+                    (nameof(Images.SubmoduleRevisionSemiUp), Images.SubmoduleRevisionSemiUp),
+                    (nameof(Images.SubmoduleRevisionSemiUpDirty), Images.SubmoduleRevisionSemiUpDirty),
+                    (nameof(Images.SubmoduleRevisionSemiDown), Images.SubmoduleRevisionSemiDown),
+                    (nameof(Images.SubmoduleRevisionSemiDownDirty), Images.SubmoduleRevisionSemiDownDirty)
+                };
+
+                var i = 0;
+                foreach ((string imageKey, Bitmap icon) in images)
```
RussKie, line 127: Suggest converting to a for loop instead of foreach. This way you're guaranteed to have the right index.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -998,69 +1014,86 @@ void EnsureSelectedIndexChangeSubscription()
 
             int GetItemImageIndex(GitItemStatus gitItemStatus)
             {
+                // If imageKey does not exist, trow an exception, implementation error
```
RussKie, line 1017: Using a mobile app, and couldn't see if we throw the exception. 
Btw,  typo "throw".

Instead throwing an exception it is better to enforce via a test.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -56,6 +56,7 @@ public sealed partial class FileStatusList : GitModuleControl
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
         public new event EnterEventHandler Enter;
+        private IDictionary<string, int> _stateImageIndexDict;
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/FileStatusList.cs
```diff
@@ -92,31 +94,45 @@ ImageList CreateImageList()
...
-                        ScaleHeight(Images.SubmoduleRevisionSemiDownDirty), // 13
-                        ScaleHeight(Images.FileStatusUnknown) // 14
-                    }
                 };
 
-                Bitmap ScaleHeight(Bitmap input)
+                _stateImageIndexDict = new Dictionary<string, int>();
+                var images = new (string imageKey, Bitmap icon)[]
+                {
+                    (nameof(Images.FileStatusUnknown), Images.FileStatusUnknown),
```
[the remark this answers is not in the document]
gerhardol: OK

GitUI/UserControls/FileStatusList.cs
gerhardol: ok

GitUI/UserControls/FileStatusList.cs
gerhardol: Check with ContainsKey

gerhardol:
@msftbot merge in 1 day

gerhardol:
> incorrect indentation

Now I know that you first have to remove the label (which I could do from mobile without too much effort), then stop msftbot to stop a merge.

RussKie:
Haha. If the label is gone, the bot stops I believe

PR #7925, opened 2020-03-31
Fix portable artifact generation.

RussKie:
Follow up for #7910

## Proposed changes

* Fix how translations are copied
* Fix how plugins are copied

----


mstv:
Thank you.

PR #7928, opened 2020-03-31
Improve blame gutter display

pmiossec:
Please, give your point of view and probable improvements...

## Proposed changes

- Display avatars in the gutter with a thin vertical marker for recency

### Before

image

### After

Note: the same information could be displayed in the gutter, that's just that it is configured differently

With avatars (notice the thin color line next to avatars that highlight recency of changes):
image

last update on heatmap colors:

* for light theme:

image

* for dark theme:
image

## Test methodology 

- Manual (for more than 6 months now...)

## Test environment(s) 

- Git Extensions 3.3.0
- Build ad8be3dafc0537dafa2a96e1cb5daf0f09b302ef
- Git 2.25.0.windows.1
- .NET Framework 4.8.4150.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/Editor/BlameAuthorMargin.cs
+++ b/GitUI/Editor/BlameAuthorMargin.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Windows.Forms;
+using GitExtUtils.GitUI;
+using ICSharpCode.TextEditor;
+
+namespace GitUI.Editor
+{
+    /// <summary>
+    /// This class display avatars in the gutter in a blame control.
+    /// </summary>
+    public class BlameAuthorMargin : AbstractMargin
+    {
+        private static readonly int AgeBucketMarkerWidth = Convert.ToInt32(4 * DpiUtil.ScaleX);
+        private List<Image> _avatars;
+        private readonly int _lineHeight;
+        private readonly Color _backgroundColor;
+        private List<GitBlameEntry> _blameLines;
+        private readonly Dictionary<int, SolidBrush> _brushs = new Dictionary<int, SolidBrush>();
+        private bool _isVisible = true;
+
+        public BlameAuthorMargin(TextArea textArea) : base(textArea)
+        {
+            _lineHeight = GetFontHeight(textArea.Font);
+            _backgroundColor = SystemColors.Window;
+            Width = _lineHeight + AgeBucketMarkerWidth + DpiUtil.Scale(2);
+        }
+
+        public override int Width { get; }
+        public override bool IsVisible => _isVisible;
+
+        private static int GetFontHeight(Font font)
+        {
+            var max = Math.Max(
+                TextRenderer.MeasureText("_", font).Height,
+                (int)Math.Ceiling(font.GetHeight()));
+
+            return max + 1;
+        }
+
+        public void Initialize(IEnumerable<GitBlameEntry> blameLines)
+        {
+            _blameLines = blameLines.ToList();
+            _avatars = _blameLines.Select(a => a.Avatar).ToList();
+
+            // Update the resolution otherwise the image is not drawn at the good size :(
+            foreach (var avatar in _avatars)
+            {
+                if (avatar is Bitmap bitmapAvatar)
+                {
+                    bitmapAvatar.SetResolution(DpiUtil.DpiX, DpiUtil.DpiY);
+                }
+            }
+
+            // Build brushes
+            foreach (var blameLine in _blameLines)
+            {
+                if (!_brushs.ContainsKey(blameLine.AgeBucketIndex))
+                {
+                    _brushs.Add(blameLine.AgeBucketIndex, new SolidBrush(blameLine.AgeBucketColor));
+                }
+            }
+        }
+
+        public void SetVisiblity(bool isVisible)
+        {
+            _isVisible = isVisible;
+        }
+
+        public override void Paint(Graphics g, Rectangle rect)
+        {
+            if (rect.Width <= 0 || rect.Height <= 0 || _blameLines == null || _blameLines.Count == 0)
+            {
+                return;
+            }
+
+            g.Clear(_backgroundColor);
+
+            if (_avatars == null || _avatars.Count == 0)
+            {
+                return;
+            }
+
+            var verticalOffset = textArea.VirtualTop.Y;
+            var lineStart = verticalOffset / _lineHeight;
+            var negativeOffset = (lineStart * _lineHeight) - verticalOffset;
+            var lineCount = (int)Math.Ceiling((double)(rect.Height - negativeOffset) / _lineHeight);
+
+            for (int i = 0; i < lineCount; i++)
+            {
+                if (lineStart + i >= _avatars.Count)
+                {
+                    break;
+                }
+
+                int y = negativeOffset + (i * _lineHeight);
+                g.FillRectangle(_brushs[_blameLines[lineStart + i].AgeBucketIndex], 0, y, AgeBucketMarkerWidth, _lineHeight);
+
+                if (_avatars[lineStart + i] != null)
+                {
+                    g.DrawImage(_avatars[lineStart + i], new Point(AgeBucketMarkerWidth, y));
+                }
+            }
+
+            base.Paint(g, rect);
+        }
+    }
+}
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Drawing;
 using System.IO;
@@ -1666,6 +1667,16 @@ namespace GitUI.Editor
 
             public ToolStripButton IgnoreAllWhitespacesButton => _fileViewer.ignoreAllWhitespaces;
             public ToolStripMenuItem IgnoreAllWhitespacesMenuItem => _fileViewer.ignoreAllWhitespaceChangesToolStripMenuItem;
+        }
+
+        public void SetGitBlameGutter(IEnumerable<GitBlameEntry> gitBlameEntries)
+        {
+            internalFileViewer.ShowGutterAvatars = AppSettings.BlameShowAuthorAvatar;
+
+            if (AppSettings.BlameShowAuthorAvatar)
+            {
+                internalFileViewer.SetGitBlameGutter(gitBlameEntries);
+            }
         }
     }
 }
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -37,6 +37,8 @@ namespace GitUI.Editor
         private bool _shouldScrollToBottom = false;
         private readonly int _bottomBlankHeight = DpiUtil.Scale(300);
         private ContinuousScrollEventManager _continuousScrollEventManager;
+        private BlameAuthorMargin _authorsAvatarMargin;
+        private bool _showGutterAvatars;
 
         public FileViewerInternal()
         {
@@ -535,6 +537,39 @@ namespace GitUI.Editor
         }
 
         #endregion
+
+        public void SetGitBlameGutter(IEnumerable<GitBlameEntry> gitBlameEntries)
+        {
+            if (_showGutterAvatars)
+            {
+                _authorsAvatarMargin.Initialize(gitBlameEntries);
+            }
+        }
+
+        public bool ShowGutterAvatars
+        {
+            get => _showGutterAvatars;
+            set
+            {
+                _showGutterAvatars = value;
+                if (!_showGutterAvatars)
+                {
+                    _authorsAvatarMargin?.SetVisiblity(false);
+
+                    return;
+                }
+
+                if (_authorsAvatarMargin == null)
+                {
+                    _authorsAvatarMargin = new BlameAuthorMargin(TextEditor.ActiveTextAreaControl.TextArea);
+                    TextEditor.ActiveTextAreaControl.TextArea.InsertLeftMargin(0, _authorsAvatarMargin);
+                }
+                else
+                {
+                    _authorsAvatarMargin.SetVisiblity(true);
+                }
+            }
+        }
 
         internal sealed class CurrentViewPositionCache
         {
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -2,19 +2,26 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.Globalization;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
 using GitExtUtils;
+using GitExtUtils.GitUI.Theming;
+using GitUI.Avatars;
 using GitUI.BranchTreePanel;
 using GitUI.CommitInfo;
 using GitUI.Editor;
 using GitUI.HelperDialogs;
+using GitUI.Properties;
 using GitUIPluginInterfaces;
 using GitUIPluginInterfaces.RepositoryHosts;
+using ICSharpCode.TextEditor;
 using JetBrains.Annotations;
+using Microsoft.VisualStudio.Threading;
 using ResourceManager;
 
 namespace GitUI.Blame
@@ -44,6 +51,7 @@ namespace GitUI.Blame
         private GitBlameCommit _tooltipCommit;
         private bool _changingScrollPosition;
         private IRepositoryHostPlugin _gitHoster;
+        private static readonly IList<Color> AgeBucketGradientColors = GetAgeBucketGradientColors();
 
         public BlameControl()
         {
@@ -283,7 +291,10 @@ namespace GitUI.Blame
 
         private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<ObjectId> children, Control controlToMask, int lineNumber, int scrollpos)
         {
-            var (gutter, body) = BuildBlameContents(filename);
+            var avatarSize = BlameAuthor.Font.Height + 1;
+            var (gutter, body, avatars) = BuildBlameContents(filename, avatarSize);
+
+            BlameAuthor.SetGitBlameGutter(avatars);
 
             ThreadHelper.JoinableTaskFactory.RunAsync(
                 () => BlameAuthor.ViewTextAsync("committer.txt", gutter));
@@ -307,11 +318,14 @@ namespace GitUI.Blame
             controlToMask?.UnMask();
         }
 
-        private (string gutter, string body) BuildBlameContents(string filename)
+        private (string gutter, string body, List<GitBlameEntry> gitBlameDisplays) BuildBlameContents(string filename, int avatarSize)
         {
             var body = new StringBuilder(capacity: 4096);
 
             GitBlameCommit lastCommit = null;
+
+            bool showAuthorAvatar = AppSettings.BlameShowAuthorAvatar;
+            var gitBlameDisplays = showAuthorAvatar ? CalculateBlameGutterData(_blame.Lines) : new List<GitBlameEntry>(0);
 
             var dateTimeFormat = AppSettings.BlameShowAuthorTime
                 ? CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern + " " +
@@ -324,22 +338,49 @@ namespace GitUI.Blame
             // the extra spaces added here could be omitted.
 
             var filePathLengthEstimate = _blame.Lines.Where(l => filename != l.Commit.FileName)
-                .Select(l => l.Commit.FileName.Length)
-                .DefaultIfEmpty(0)
-                .Max();
+                                                     .Select(l => l.Commit.FileName.Length)
+                                                     .DefaultIfEmpty(0)
+                                                     .Max();
             var lineLengthEstimate = 25 + _blame.Lines.Max(l => l.Commit.Author?.Length ?? 0) + filePathLengthEstimate;
             var lineLength = Math.Max(80, lineLengthEstimate);
             var lineBuilder = new StringBuilder(lineLength + 2);
             var gutter = new StringBuilder(capacity: lineBuilder.Capacity * _blame.Lines.Count);
             var emptyLine = new string(' ', lineLength);
-            foreach (var line in _blame.Lines)
+            var cacheAvatars = new Dictionary<string, Image>();
+            var noAuthorImage = (Image)new Bitmap(Images.User80, avatarSize, avatarSize);
+            for (var index = 0; index < _blame.Lines.Count; index++)
             {
+                var line = _blame.Lines[index];
                 if (line.Commit == lastCommit)
                 {
                     gutter.AppendLine(emptyLine);
                 }
                 else
                 {
+                    var authorEmail = line.Commit.AuthorMail?.Trim('<', '>');
+                    if (showAuthorAvatar)
+                    {
+                        if (authorEmail != null)
+                        {
+                            if (cacheAvatars.ContainsKey(authorEmail))
+                            {
+                                gitBlameDisplays[index].Avatar = cacheAvatars[authorEmail];
+                            }
+                            else
+                            {
+                                var avatar = ThreadHelper.JoinableTaskFactory.Run(() =>
+                                    AvatarService.Default.GetAvatarAsync(authorEmail, line.Commit.Author,
+                                        avatarSize));
+                                cacheAvatars.Add(authorEmail, avatar);
+                                gitBlameDisplays[index].Avatar = avatar;
+                            }
+                        }
+                        else
+                        {
+                            gitBlameDisplays[index].Avatar = noAuthorImage;
+                        }
+                    }
+
                     BuildAuthorLine(line, lineBuilder, dateTimeFormat, filename, AppSettings.BlameShowAuthor, AppSettings.BlameShowAuthorDate, AppSettings.BlameShowOriginalFilePath, AppSettings.BlameDisplayAuthorFirst);
 
                     gutter.Append(lineBuilder);
@@ -352,10 +393,11 @@ namespace GitUI.Blame
                 lastCommit = line.Commit;
             }
 
-            return (gutter.ToString(), body.ToString());
+            return (gutter.ToString(), body.ToString(), gitBlameDisplays);
         }
 
-        private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, string dateTimeFormat, string filename, bool showAuthor, bool showAuthorDate, bool showOriginalFilePath, bool displayAuthorFirst)
+        private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, string dateTimeFormat,
+            string filename, bool showAuthor, bool showAuthorDate, bool showOriginalFilePath, bool displayAuthorFirst)
         {
             if (showAuthor && displayAuthorFirst)
             {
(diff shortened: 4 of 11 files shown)


gerhardol:
No code review (no major comment when browsing though).

From the screenshots, I like the avatars, compact and easier to find matching
Recency colors (I had to look up that word that it was proper English but can understand what it means and have no better suggestion): I like the narrow band, but the full colors are too much in my face, they should be much more subtle, but even that would be catching the eye too much. Default could be to not show full colors.

The settings should be changed from Disable to Show, to match other options
The current value is not set when opening the form

(Not related to this PR). The settings for Blame is very hard to find. I had to browse the code to find it.
Maybe changing the icon would be sufficient?

Colors should probably be themeable

Not adapted to dark theme...
Fixed by the following, but adapted colors do not look good here (not very nice for normal theme either)

image

Before:
image

Changed to
image

RussKie:
Just going off the screenshots. 

I don't like too much, sorry. Less text is nice, but I feel like I've lost information. In the Git Extensions repo more or less recognise avatars of all active contributors. In other repos I would have constantly hover over avatars (I presume there are tooltips to show names). I suppose this can be a toggle switch to show one way or another...
There is also aspect of accessibility, which we have been neglecting and ignoring. But at some point we will have to seriously look into making the app more accessible and usable with tools like Windows Narrator.

I like the idea with colours, but I do not like when colours span full lines, it is just too distracting. I don't mind the flame graph near the avatars, though I would probably put it to the left of the avatars.

pmiossec:
>I don't like too much, sorry. Less text is nice, but I feel like I've lost information. 

You could get the author name like before (that's even the default display).
That's why I said "Note: the same information could be displayed in the gutter, that's just that it is configured differently" ;)

You could have the exact same display than today. Should I provide a screenshot to prove it? 🤣 

> In the Git Extensions repo more or less recognise avatars of all active contributors. In other repos I would have constantly hover over avatars (I presume there are tooltips to show names). I suppose this can be a toggle switch to show one way or another...

I don't know if that's the same for everyone but I realized that when I do a blame, I don't care to have the name of the authors to know who contribute each line.
We only search the commit that introduced some changes. And that's once you found it that you (sometimes) want to know the author.

Displaying avatars is a way:
* to have a lighter (with less text) display
* to recognize more quickly authors between lines (because it's quicker to barely recognize a picture than to read a text) when you don't know the names of authors (even if you don't care about the name in big repositories)
* to recognize more quickly authors when you know them.

>There is also aspect of accessibility, which we have been neglecting and ignoring. 

Out of context as you could get the same display as before ;)

>I like the idea with colours, but I do not like when colours span full lines, it is just too distracting.

 &

>I like the narrow band, but the full colors are too much in my face, they should be much more subtle, but even that would be catching the eye too much. 

That was the first quick development I did to more quickly grasp the recency of changes.
That's only after that I found the idea of the thin vertical line.

Even if it is more visible, like you, I was not fully satisfy by this feature. I just would like to know if someone could find that useful to know if I should save it.

That's not the case, so I removed it (I will update the screenshots)  

> I don't mind the flame graph near the avatars, though I would probably put it to the left of the avatars.

Ok, will do it 👍 

>Colors should probably be themeable

&

>Not adapted to dark theme...

I'm not fully satisfied by the colors for the moment. As a first shot, to not invest too much time, I generated them.

I was also constrained by keeping the same color in the thin vertical line and in the lines highlights which make things difficult.

But my goal idea is more to select better ones. I'm not sure I will go to something fully themeable because not one will really use it but I think I will create one set for dark themes and one for light ones.

What do you think?

gerhardol:
Looking good right now.

For theming, basically hardcoded colors are adapted to something reasonable with magic by @NikolayXHD 
Just add the command I did show to get better presentation
Themeable colors are set in the .json files, not sure if this should be added now when the band is so narrow

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +434,48 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+            var stepR = (end.R - start.R) / (steps - 1);
+            var stepG = (end.G - start.G) / (steps - 1);
+            var stepB = (end.B - start.B) / (steps - 1);
+
+            for (int i = 0; i < steps; i++)
+            {
+                yield return Color.FromArgb(start.A + (stepA * i),
+                    start.R + (stepR * i),
+                    start.G + (stepG * i),
+                    start.B + (stepB * i));
```
gerhardol: .AdaptBackColor()

pmiossec:
I have updated the heatmap colors to follow these ones: https://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=9

and also put the recency marker on the left of the avatars.

GitUI/UserControls/BlameControl.cs
pmiossec: Sorry, I have pushed a change at this code nearly at the same time of your comment.
Is this change (that I don't fully understand) really needed now that we no more write text on it?

gerhardol: The colors still look a little strange even if just a small band, so I believe so.

pmiossec: The 2 seems strange :(

Real heatmap colors we are used to (without `.AdaptBackColor()`):
image

Adapted heatmap colors that no more looks like a heatmap (with `.AdaptBackColor()`):
image

Surely that I must use other colors but I really don't know which...

pmiossec: Perhaps better:
image

RussKie: overall liking the direction, though I'd go for the green color set
Screenshot_20200403-172152

RussKie: dark green being the most recent, and whitish the oldest

RussKie: for the dark version I think we'd need to invert, light green - the youngest, dark grey the oldest

gerhardol: > for the dark version I think we'd need to invert, light green - the youngest, dark grey the oldest

That is basically what the .AdaptBackColor() does

It is easier to handle theming with a minimal of specific changes
Green with Adapt looks OK for both normal/dark theme

green without adapt

image

image

Adapt

image

RussKie:
Thank you Gerhard, that looks good

GitUI/UserControls/BlameControl.cs
pmiossec: Question: How to easily get if the current theme is mostly dark or light to adapt the colors?

RussKie: `ColorHelper.IsLightTheme()`

NikolayXHD: Use ``ColorHelper.AdaptColor(c, isForeground);`` whenever possible so that the color is chosen automatically. The advantage is, the result changes if the theme changes.

pmiossec:
I have updated the screenshot with the last adjustments

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            if (_avatars == null || _avatars.Count == 0)
+            {
+                return;
+            }
+
+            var verticalOffset = textArea.VirtualTop.Y;
+            var lineStart = verticalOffset / _lineHeight;
+            var negativeOffset = (lineStart * _lineHeight) - verticalOffset;
+            var lineCount = (int)rect.Height / _lineHeight;
```
NikolayXHD, line 98: Correct me if I am wrong, I believe `lineCount` should be rather
```csharp
var lineCount = (int)Math.Ceiling((rect.Height - negativeOffset) / _lineHeight);
```
`- negativeOffset` because the first line may start above 0 due to `negativeOffset`
`(int)Math.Ceiling()` because we should also paint the bottom partially visible row

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -646,5 +647,30 @@ public TestAccessor(FileViewerInternal control)
 
             public TextEditorControl TextEditor => _control.TextEditor;
         }
+
+        public void SetGutterAvatars(List<GitBlameDisplay> avatars)
```
NikolayXHD, line 651: Applying `AppSettings.BlameShowAuthorAvatar` looks to me like a second responsibility for the current method.
Wouldn't it be better to separate it out into a property `bool ShowGutterAvatars`?

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +435,48 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+                Color.FromArgb(247, 252, 245).AdaptBackColor(),
+                Color.FromArgb(229, 245, 224).AdaptBackColor(),
+                Color.FromArgb(199, 233, 192).AdaptBackColor(),
+                Color.FromArgb(161, 217, 155).AdaptBackColor(),
+                Color.FromArgb(116, 196, 118).AdaptBackColor(),
+                Color.FromArgb(65, 171, 93).AdaptBackColor(),
+                Color.FromArgb(35, 139, 69).AdaptBackColor(),
+                Color.FromArgb(0, 109, 44).AdaptBackColor(),
+                Color.FromArgb(0, 68, 27).AdaptBackColor(),
+            };
```
NikolayXHD, line 452: A purely stylistic suggestion: we could avoid repeating AdaptBackColor() by using `.Select()`

```csharp
new [] 
{
    Color.FromArgb(247, 252, 245),
    // ...
}.Select(ColorHelper.AdaptBackColor).ToList()
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +435,48 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+            {
+                return null;
+            }
+
+            var gitBlameDisplays = new List<GitBlameDisplay>(_blame.Lines.Count);
+            var blameLines = _blame.Lines;
+
+            var colors = GetGradients();
+            var mostRecentDate = DateTime.Now.Ticks;
+            var lessRecentDate = Math.Min(DateTime.Now.AddYears(3).Ticks, blameLines.Select(l => l.Commit.AuthorTime).Where(d => d != DateTime.MinValue).Min().Ticks);
```
NikolayXHD, line 467:
```csharp
var lessRecentDate = blameLines
    .Select(l => l.Commit.AuthorTime).Where(d => d != DateTime.MinValue)
    .DefaultIfEmpty(DateTime.Now.AddYears(3)) // otherwise Min will throw on empty enumerable
    .Min().Ticks
```

NikolayXHD, line 467:
```csharp
DateTime.Now.AddYears(3).Ticks
```
Is it `3` or `-3`? At the first glance I would expect `lessRecentDate` to be in the past rather than future.

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +435,48 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+            var intervalSize = (mostRecentDate - lessRecentDate) / colors.Count;
```
NikolayXHD, line 468:
```csharp
var intervalSize = (mostRecentDate - lessRecentDate + 1) / colors.Count;
```
because consider inclusive interval [0 -> 1] with colors.Count == 2
then size would be
```
size = 1 - 0 + 1 / 2; // == 1
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +435,48 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+            foreach (var blame in _blame.Lines)
+            {
+                var relativeTicks = Math.Max(0, blame.Commit.AuthorTime.Ticks - lessRecentDate - 1);
```
NikolayXHD, line 471: `-1` here seems redundant. I can imagine it was solving the problem due to a missing `+1` in `intervalSize` expression above

RussKie:
Looks like we need tests...

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Windows.Forms;
+using GitExtUtils.GitUI;
+using ICSharpCode.TextEditor;
+using Microsoft.VisualStudio.Threading;
+
+namespace GitUI.Editor
+{
+    public class GitBlameDisplay
+    {
+        public JoinableTask<Image> Avatar { get; set; }
```
RussKie, line 14: ❗️ No no no, we must not return tasks, especially from properties. It is semantically incorrect to "await a property".
If you want to defer you could return a `Lazy<Image>`, but even then I'm not sure it would be correct

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Drawing;
+using System.Linq;
+using System.Windows.Forms;
+using GitExtUtils.GitUI;
+using ICSharpCode.TextEditor;
+using Microsoft.VisualStudio.Threading;
+
+namespace GitUI.Editor
+{
+    public class GitBlameDisplay
```
RussKie, line 12: 1 class per file

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        public JoinableTask<Image> Avatar { get; set; }
+        public int RecencyIndex { get; set; }
```
RussKie, line 15: I don't believe there is an English word "recency".
May be call it "AgeBucket" - "AgeBucketIndex", "AgeBucketColor", etc.?

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public override int Width { get; }
+
+        public void SetVisiblity(bool isVisible)
+        {
+            _isVisible = isVisible;
+        }
+
+        public override bool IsVisible => _isVisible;
```
RussKie, line 79: [readability] keep properties before methods

GitUI/Editor/BlameAuthorMargin.cs
RussKie, line 12:
```suggestion
    public class GitBlameEntry
```

GitUI/Editor/FileViewer.cs
```diff
@@ -1618,5 +1619,10 @@ public bool ShowSyntaxHighlightingInDiff
             public ToolStripButton IgnoreAllWhitespacesButton => _fileViewer.ignoreAllWhitespaces;
             public ToolStripMenuItem IgnoreAllWhitespacesMenuItem => _fileViewer.ignoreAllWhitespaceChangesToolStripMenuItem;
         }
+
+        public void SetGutterAvatars(List<GitBlameDisplay> gitBlameDisplays)
```
RussKie, line 1623: The method is called "set avatars", yet its parameter isn't a collection of images. So the name is misleading.

Maybe something like this?
```suggestion
        public void SetGitBlameGutter(List<GitBlameEntry> blameEntries)
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,47 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
+                        {
+                            if (cacheAvatars.ContainsKey(authorEmail))
+                            {
+                                gitBlameDisplays[index].Avatar = cacheAvatars[authorEmail];
+                            }
+                            else
+                            {
+                                var avatar = ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                                    await AvatarService.Default.GetAvatarAsync(authorEmail, line.Commit.Author,
+                                        avatarSize));
```
RussKie, line 378: ❓  `GetAvatarAsync` returns an image (when awaited). Why are we attempting to cache a task?

I think the following should work
```suggestion
                                Image avatar = ThreadHelper.JoinableTaskFactory.RunAsync(() => AvatarService.Default.GetAvatarAsync(authorEmail, line.Commit.Author, avatarSize));
```

pmiossec:
Done.

GitUI/UserControls/BlameControl.cs
```diff
@@ -575,7 +658,10 @@ public GitBlame Blame
             public void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, string dateTimeFormat, string filename, bool showAuthor, bool showAuthorDate, bool showOriginalFilePath, bool displayAuthorFirst)
                 => _control.BuildAuthorLine(line, lineBuilder, dateTimeFormat, filename, showAuthor, showAuthorDate, showOriginalFilePath, displayAuthorFirst);
 
-            public (string gutter, string body) BuildBlameContents(string filename) => _control.BuildBlameContents(filename);
+            public (string gutter, string body, List<GitBlameEntry> avatars) BuildBlameContents(string filename) => _control.BuildBlameContents(filename, 10);
```
NikolayXHD, line 661:
```csharp
_control.BuildBlameContents(filename, avatarSize: 10);
```
or introduce `const int AvatarSize = 10`, because the meaning of number cannot be immediately seen from usage.

UnitTests/GitUI.Tests/UserControls/BlameControlTests.cs
```diff
@@ -149,5 +150,105 @@ public void BuildBlameContents_WithDateButNotTime()
...
+                        "Commiter", "@Committer", lineDate, string.Empty, "Summary1", "file"),
+                    index + 1, index + 1, "text");
+            }
+        }
+
+        [Test]
+        public void CalculateBlameGutterData_When_date_is_older_than_artificial_old_boundary_Then_it_define_first_age_bucket()
+        {
+            // Given
+            IReadOnlyList<GitBlameLine> blameLines = CreateBlameLine(new DateTime(2008, 9, 1)).ToList();
```
NikolayXHD, line 170: `ArtificialOldBoundary` is relative to current clock, while `new DateTime(2008, 9, 1)` is absolute.

one way to resolve it would be: 
```csharp
var olderDate = _blameControl.GetTestAccessor().ArtificialOldBoundary.AddYears(-1)
```

UnitTests/GitUI.Tests/UserControls/BlameControlTests.cs
```diff
@@ -149,5 +150,105 @@ public void BuildBlameContents_WithDateButNotTime()
...
+            blameEntries[0].AgeBucketIndex.Should().Be(0);
+            blameEntries[1].AgeBucketIndex.Should().Be(0);
+        }
+
+        [Test]
+        public void CalculateBlameGutterData_When_dates_just_after_age_bucket_limit_Then_One_date_in_each_age_bucket()
+        {
+            // Given
+            var now = DateTime.Now;
+            var marginError = TimeSpan.FromMilliseconds(10); // Due to the DateTime.Now value which is slightly different
```
NikolayXHD, line 228: Can it be hours instead of milliseconds for better reliability?

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,46 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
+                        {
+                            if (cacheAvatars.ContainsKey(authorEmail))
+                            {
+                                gitBlameDisplays[index].Avatar = cacheAvatars[authorEmail];
+                            }
+                            else
+                            {
+                                var avatar = ThreadHelper.JoinableTaskFactory.RunAsync(() =>
+                                    AvatarService.Default.GetAvatarAsync(authorEmail, line.Commit.Author,
+                                        avatarSize)).Join();
```
RussKie, line 377: Why not this?
```suggestion
                                var avatar = ThreadHelper.JoinableTaskFactory.Run(() =>
                                    AvatarService.Default.GetAvatarAsync(authorEmail, line.Commit.Author, avatarSize));
```

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,110 @@
...
+        private static int GetFontHeight(Font font)
+        {
+            var max = Math.Max(
+                TextRenderer.MeasureText("_", font).Height,
+                (int)Math.Ceiling(font.GetHeight()));
+
+            return max + 1;
+        }
+
+        public void Initialize(List<GitBlameEntry> blameLines)
```
RussKie, line 43: ❗️ Passing mutable collections can lead to hard-to-track bugs.
Since we're re-enumerating and re-materialising the collection few lines down, pass `IEnumerable<>`. 
The same applies to upstream calls.

GitUI/Editor/BlameAuthorMargin.cs
```diff
@@ -0,0 +1,110 @@
...
+using ICSharpCode.TextEditor;
+
+namespace GitUI.Editor
+{
+    /// <summary>
+    /// This class display avatars in the gutter in a blame control.
+    /// </summary>
+    public class BlameAuthorMargin : AbstractMargin
+    {
+        private static readonly int RecencyIndexWidth = Convert.ToInt32(4 * DpiUtil.ScaleX);
```
RussKie, line 16: Looks like this was missed during the rename

GitUI/UserControls/BlameControl.cs
```diff
@@ -307,12 +317,15 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
 
-        private (string gutter, string body) BuildBlameContents(string filename)
+        private (string gutter, string body, List<GitBlameEntry> gitBlameDisplays) BuildBlameContents(string filename,
+            int avatarSize)
         {
             var body = new StringBuilder(capacity: 4096);
 
             GitBlameCommit lastCommit = null;
 
+            var gitBlameDisplays = AppSettings.BlameShowAuthorAvatar ? CalculateBlameGutterData(_blame.Lines) : null;
```
RussKie, line 327: It is a bad practice to return `null` for enumerable types. This can easily lead to NREs, as it is very unexpected.

GitUI/UserControls/BlameControl.cs
```diff
@@ -352,10 +397,11 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
                 lastCommit = line.Commit;
             }
 
-            return (gutter.ToString(), body.ToString());
+            return (gutter.ToString(), body.ToString(), gitBlameDisplays);
```
RussKie, line 400: With the above comment, we have to do this:
```suggestion
            return (gutter.ToString(), body.ToString(), gitBlameDisplays ?? Array.Empty<GitBlameEntry>());
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,46 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
+            var cacheAvatars = new Dictionary<string, Image>();
+            var noAuthorTask = (Image)new Bitmap(Images.User80, avatarSize, avatarSize);
+            for (var index = 0; index < _blame.Lines.Count; index++)
             {
+                var line = _blame.Lines[index];
                 if (line.Commit == lastCommit)
                 {
+                    if (gitBlameDisplays != null)
+                    {
+                        gitBlameDisplays[index].Avatar = null;
```
RussKie, line 357: Are we certain there will be no entries in the collection with `.Avatar == null`?
I feel we must re-iterate the collection to ensure this before it gets return.

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,46 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
             var emptyLine = new string(' ', lineLength);
-            foreach (var line in _blame.Lines)
+            var cacheAvatars = new Dictionary<string, Image>();
+            var noAuthorTask = (Image)new Bitmap(Images.User80, avatarSize, avatarSize);
+            for (var index = 0; index < _blame.Lines.Count; index++)
             {
+                var line = _blame.Lines[index];
                 if (line.Commit == lastCommit)
                 {
+                    if (gitBlameDisplays != null)
```
RussKie, line 355: Invert to reduce nesting, and `continue`

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,46 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
+                    {
+                        gitBlameDisplays[index].Avatar = null;
+                    }
+
                     gutter.AppendLine(emptyLine);
                 }
                 else
                 {
+                    var authorEmail = line.Commit.AuthorMail?.Trim('<', '>');
+                    if (gitBlameDisplays != null)
```
RussKie, line 365: Invert to reduce nesting, and `continue`

GitUI/UserControls/BlameControl.cs
```diff
@@ -332,14 +345,46 @@ private void ProcessBlame(string filename, GitRevision revision, IReadOnlyList<O
...
+                    {
+                        if (authorEmail != null)
```
RussKie, line 367: Invert to reduce nesting, and `continue`

GitUI/UserControls/BlameControl.cs
```diff
@@ -388,6 +434,44 @@ private void BuildAuthorLine(GitBlameLine line, StringBuilder lineBuilder, strin
...
+        }
+
+        public DateTime ArtificialOldBoundary => DateTime.Now.AddYears(-3);
+
+        private List<GitBlameEntry> CalculateBlameGutterData(IReadOnlyList<GitBlameLine> blameLines)
+        {
+            var mostRecentDate = DateTime.Now.Ticks;
+            var artificialOldBoundary = ArtificialOldBoundary;
+            var gitBlameDisplays = new List<GitBlameEntry>(blameLines.Count);
+            var colors = GetGradients();
```
RussKie, line 459: Why is this called multiple times? This array must be static and initialised in a static ctor (could be lazy).

GitUI/UserControls/BlameControl.cs
pmiossec: That's the goal to have the avatar property to be null to not display an avatar on the corresponding line (when we are on the same commit):

image

But the initial value is already `null`, so this code is not needed. I removed it.

GitUI/UserControls/BlameControl.cs
pmiossec: That's not possible because there is common code after the `if()` block.

GitUI/UserControls/BlameControl.cs
pmiossec: same here.

GitUI/UserControls/BlameControl.cs
pmiossec: same here.

GitUI/UserControls/BlameControl.cs
pmiossec: not needed because that won't be null here.

pmiossec:
I think all the requested changes have been fixed.

NikolayXHD:
> I think all the requested changes have been fixed.

From me, 1 little point remains:
https://github.com/gitextensions/gitextensions/pull/7928/files#r404467979

UnitTests/GitUI.Tests/UserControls/BlameControlTests.cs
pmiossec: Sorry, I forgot to fix this one because I didn't know immediately what to think about your comment...

Using milliseconds better express the idea that the calculation is exact (except for a small time corresponding to the time between this `DateTime.Now` and the one in the tested method, so just the time needed to generate the values). I imagine that 10ms should be enough...

But indeed, you're right, I also want the test to be reliable.
So, what do you think if I add 1 second? I don't like the idea to add **hours**...

NikolayXHD: can we have a deal at 10 minutes? :)

pmiossec:
@NikolayXHD sold! Even if I think it's way too long ;)

RussKie:
I pushed a small code tweak in

pmiossec:
Rebased and squashed.

PR #7941, opened 2020-04-03
Windows TaskBar JumpList was not initiated

### LINKED ISSUE #7683: Windows taskbar jumplist entries don't work
## Current behaviour

1. The Windows taskbar jumplist entries of GE do not work:
"The selected element is not available. It has been moved, renamed or removed. ...",
although the `%APPDATA%\Roaming\GitExtensions\GitExtensions\Recent\*.gitext` files contain the correct paths (with a trailing backslash).
2. The recent repos list does not match the jumplist.

## Expected behaviour

1. Jumplist entries work
2. Any repo opened in GE is added to the jumplist.

## Steps to reproduce

1.
  - open a repo using the Explorer context menu
  - pin GE to the Windows taskbar
  - close GE
  - right click the GE icon in the taskbar and select the recent repo
2.
  - select a repo from the dropdown in the GE taskbar or from the GE dashboard
  - (optionally) close GE
  - right click the GE icon in the taskbar

## Screenshots

grafik

yes, but don't know / haven't checked which
issue 1. works with the same or only slightly older build on Win 7

## Environment

- Git Extensions 3.3.0
- Build 7034c98856ecb70f1d716d12e92652e6998c98ff (Dirty)
- Git 2.24.1.windows.2
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #7683

Regression after d072142baaec034a10bedd9fa15854eed6269222 from #7750 
The button handling has been incorrect for years

Still not fully working

## Proposed changes
Windows TaskBar JumpList was not initiated correctly

ApplicationId init too late
Windows TaskBar JumpList could be initiated without the window handle was available
JumpList buttons were never enabled when starting in Dashboard

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/WindowsJumpListManager.cs
+++ b/GitUI/WindowsJumpListManager.cs
@@ -23,12 +23,16 @@ namespace GitUI
         private ThumbnailToolBarButton _commitButton;
         private ThumbnailToolBarButton _pushButton;
         private ThumbnailToolBarButton _pullButton;
-        private bool _toolbarButtonsCreated;
+        private bool ToolbarButtonsCreated => _commitButton != null;
         private readonly IRepositoryDescriptionProvider _repositoryDescriptionProvider;
 
         public WindowsJumpListManager(IRepositoryDescriptionProvider repositoryDescriptionProvider)
         {
             _repositoryDescriptionProvider = repositoryDescriptionProvider;
+            if (TaskbarManager.IsPlatformSupported)
+            {
+                TaskbarManager.Instance.ApplicationId = AppSettings.ApplicationId;
+            }
         }
 
         ~WindowsJumpListManager()
@@ -52,7 +56,8 @@ namespace GitUI
             }
         }
 
-        private static bool IsSupported => EnvUtils.RunningOnWindowsWithMainWindow() && TaskbarManager.IsPlatformSupported;
+        private static bool IsSupported => EnvUtils.RunningOnWindows() && TaskbarManager.IsPlatformSupported;
+        private static bool IsSupportedAndVisible => EnvUtils.RunningOnWindowsWithMainWindow() && TaskbarManager.IsPlatformSupported;
 
         /// <summary>
         /// Adds the given working directory to the list of Recent for future quick access.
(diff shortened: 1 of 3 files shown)


GitUI/WindowsJumpListManager.cs
```diff
@@ -29,6 +29,10 @@ public sealed class WindowsJumpListManager : IDisposable
         public WindowsJumpListManager(IRepositoryDescriptionProvider repositoryDescriptionProvider)
         {
             _repositoryDescriptionProvider = repositoryDescriptionProvider;
+            if (TaskbarManager.IsPlatformSupported)
+            {
+                TaskbarManager.Instance.ApplicationId = "GitExtensions";
```
RussKie, line 34: Do we have a constant for this?

gerhardol: No, can add one. Where?

RussKie:
I vaguely remember we have a const for "GitExtensions".
Not a big deal

GitUI/WindowsJumpListManager.cs
gerhardol: > I vaguely remember we have a const for "GitExtensions".
Not a big deal

This must match the id that associates to .gitext in the installer and probably should match to the installer ProgId, various AssemblyProduct and hardcoded paths in AppSettings.
But AppSettings is the most sane, I will add it there for now.

GitUI/WindowsJumpListManager.cs
```diff
@@ -23,12 +23,16 @@ public sealed class WindowsJumpListManager : IDisposable
         private ThumbnailToolBarButton _commitButton;
         private ThumbnailToolBarButton _pushButton;
         private ThumbnailToolBarButton _pullButton;
-        private bool _toolbarButtonsCreated;
+        private bool ToolbarButtonsCreated => _commitButton != null;
```
RussKie, line 26: [nit] since it has become a property we need to move it down to where the rest of properties are defined

gerhardol: updated, squashed, rebased

PR #7955, opened 2020-04-07
Fix error during Publish build.

dimfishr:
Fix "An equivalent project (a project with the same global properties and tools version is already present in the project collection)" error during Publish build.


RussKie:
Thank you, I did observe this issue on few occasions but couldn't find out a cure.

PR #7923, opened 2020-03-30
RevDiff: Show common/unique files for BASE->selected(B)

gerhardol:
Part of #4154 
Separated from #7899 , the first part of the compare
This is partly to simplify the review, partly to simplify my merges, partly to drive the ideas forward: What more do we want to see?

## Proposed changes

#7720 added the possibility to see differences to BASE-> selected, if the two commits first/selected shared common BASE.

This PR separate the groups for the files that are unique BASE->selected(B) and common for
BASE->selected(B) and BASE->first(A) to give a better diff of branches

#7899 contains the changes to also show the unique files BASE->first(A) and to simplify other structures in GE

## Screenshots

image

### Before

image

### After

image

See #7899 for BASE->first

## Test methodology 

Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #7909, opened 2020-03-27
Diff of patches


mstv:
## Feature description

Another extension idea came up finally while reviewing the PR #7899.
We rebase and/or squash often. It is not easy to check whether the resulting commit is the same patch.

So for a two-commit selection we could have another diff group comparing the patches of these commits. Though this diff group is not very meaningful for unrelated commits. Perhaps it should appear only if the commit subjects are identical.

More general for a suiting four-commit selection, the pair of patches could be diffed.
A special case of this is a three-commit selection of `first`, `selected` and `squashed`.

This does not necessarily need to be displayed a as an additional diff group.
A context menu item / hotkey could suffice. It could open a diff of the two patch files in a FileView or the difftool or even a directory diff as in #7687,

## Environment

#7899

gerhardol:
I have considered the same but in a separate form.
But an explicit 4 commit selection could be used, adding other groups. (This would eliminate the 4 commit multi selection, but this what I use the 4 commit selection for anyway). (The 'squashed' scenario is handled either with #7899 if BASE is the same or 4 way selection, I want that 3-way multi select)

gerhardol:
Duplicate of #8448 
#8448 is actually a specific implementation of the problem, but it is Git standard functionality, which is preferred to something homegrown

PR #7961, opened 2020-04-08
Align context menu order for FileStatusList

gerhardol:
## Proposed changes 

See the commits for detailed explanations, this is a summary of changes.

The context menus are not aligned much, neither contents nor order.
This PR tries to align the options in a x more consistent way, grouping similar functionality and aligning RevisionDiff, RevisionFile and FormCommit.
The order can be discussed, but I doubt I will update the screen shots.

The convention in menus is that the first item should be the default 'double-clicked' action and that it should be default.
This PR makes sure that items are bot bold if they are not the default and sets submodule "Open with Git Extensions" as bold.
To change the default file action to open the history in RevDiff/RevFile, some deeper changes are required. If so, the History/Blame should probably be handled by FileStatusList as OpenWithGE is now. Just setting bold should be eenough, setting order will be a little complex -  I want DiffTool as default in RevDiff anyway

One functional change (not counting the setting to bold), Edit File was not blocked for FormCommit Edit file.

Some code removed too.

A squash is planned before merge, probably just one commit.

Note: This should be merged before 4.0, all screenshots in the documentation need to be updated anyway...

### Before
(Shortcut for Stage/Unstage from #7825, not in master)
pre-revdiff-file-worktree
pre-revdiff-subm-worktree

pre-revdiff-file-index
pre-revdiff-subm-index

pre-revdiff-file-commit
pre-revdiff-subm-commit

pre-revfile-subm
pre-revfile-file

pre-commit-file-worktree
pre-commit-subm-worktree

pre-commit-file-index
pre-commit-subm-index

Removed
pre-commit-subm-summary
pre-revdiff-subm-summary

### After

post-revdiff-file-worktree
post-revdiff-subm-worktree

post-revdiff-file-index
post-revdiff-subm-index

post-revdiff-file-commit
post-revdiff-subm-commit

post-revfile-file
post-revfile-subm

post-commit-file-worktree
post-commit-subm-worktree

post-commit-file-index
post-commit-subm-index

## Test methodology 
Manual

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -56,6 +56,11 @@ namespace GitUI
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
         public new event EnterEventHandler Enter;
+
+        [Description("Disable showing open submodule menu items as bold")]
+        [DefaultValue(false)]
+        public bool DisableSubmoduleMenuItemBold { get; set; }
+
         private Dictionary<string, int> _stateImageIndexDict;
 
         public FileStatusList()
@@ -868,7 +873,7 @@ namespace GitUI
             {
                 Name = "openSubmoduleMenuItem",
                 Tag = "1",
-                Text = "Open with Git Extensions",
+                Text = Strings.OpenWithGitExtensions,
                 Image = Images.GitExtensionsLogo16
             };
             _openSubmoduleMenuItem.Click += (s, ea) => { ThreadHelper.JoinableTaskFactory.RunAsync(() => OpenSubmoduleAsync()); };
@@ -1187,20 +1192,23 @@ namespace GitUI
         {
             var cm = (ContextMenuStrip)sender;
 
+            // TODO The handling of _openSubmoduleMenuItem need to be revised
+            // This code handles the 'bold' in the menu for submodules. Other default actions are not set to bold.
+            // The actual implementation of the default handling with doubleclick is in each form,
+            // separate from this menu item
+
             if (!cm.Items.Find(_openSubmoduleMenuItem.Name, true).Any())
             {
-                cm.Items.Insert(1, _openSubmoduleMenuItem);
+                cm.Items.Insert(0, _openSubmoduleMenuItem);
             }
 
             bool isSubmoduleSelected = SelectedItem != null && SelectedItem.IsSubmodule;
-
             _openSubmoduleMenuItem.Visible = isSubmoduleSelected;
-
-            if (isSubmoduleSelected)
+            if (isSubmoduleSelected && !DisableSubmoduleMenuItemBold)
             {
-                _openSubmoduleMenuItem.Font = AppSettings.OpenSubmoduleDiffInSeparateWindow ?
-                    new Font(_openSubmoduleMenuItem.Font, FontStyle.Bold) :
-                    new Font(_openSubmoduleMenuItem.Font, FontStyle.Regular);
+                _openSubmoduleMenuItem.Font = AppSettings.OpenSubmoduleDiffInSeparateWindow
+                    ? new Font(_openSubmoduleMenuItem.Font, FontStyle.Bold)
+                    : new Font(_openSubmoduleMenuItem.Font, FontStyle.Regular);
             }
 
             if (!cm.Items.Find(_sortByContextMenu.Name, true).Any())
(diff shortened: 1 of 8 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -1146,18 +1151,18 @@ private void FileStatusListView_ContextMenu_Opening(object sender, CancelEventAr
...
 
-            if (isSubmoduleSelected)
+            if (isSubmoduleSelected && !DisableSubmoduleMenuItemBold)
             {
-                _openSubmoduleMenuItem.Font = AppSettings.OpenSubmoduleDiffInSeparateWindow ?
-                    new Font(_openSubmoduleMenuItem.Font, FontStyle.Bold) :
-                    new Font(_openSubmoduleMenuItem.Font, FontStyle.Regular);
+                _openSubmoduleMenuItem.Font = AppSettings.OpenSubmoduleDiffInSeparateWindow
+                    ? new Font(_openSubmoduleMenuItem.Font, FontStyle.Bold)
+                    : new Font(_openSubmoduleMenuItem.Font, FontStyle.Regular);
```
RussKie, line 1165: I'm not sure this is correct.

This needs to happen if a submodule is selected, irrespective of "bold or not".
The second part - i.e. set font weight - depends on two conditions now - `AppSettings.OpenSubmoduleDiffInSeparateWindow` and `DisableSubmoduleMenuItemBold`.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -56,6 +56,11 @@ public sealed partial class FileStatusList : GitModuleControl
         public new event EventHandler DoubleClick;
         public new event KeyEventHandler KeyDown;
         public new event EnterEventHandler Enter;
+
+        [Description("Disable showing open submodule menu items as bold")]
+        [DefaultValue(false)]
+        public bool DisableSubmoduleMenuItemBold { get; set; }
```
RussKie, line 62: A "bold" menu item implies "default" item that gets invoked on double click.
This property controls the visual aspect, i.e. item is bold or not, but I can't see where double-click action is enabled/disabled respectively.

RussKie, line 62: Maybe?
```suggestion
        public bool OpenSubmoduleAsDefaultAction { get; set; }
```

GitUI/UserControls/FileStatusList.cs
gerhardol: This was not correct before. See the screenshot for FormCommit Index 
image

This worked for FormCommit worktree as FormCommit created a separate menu item with the same name was created.

See below as well.

GitUI/UserControls/FileStatusList.cs
gerhardol: Implementation is in each form using FileStatusList, not related to the menu item (that item is unknown in the forms).

gerhardol: This only controls the 'bold' part, not default action.
I also avoided setting this to true by default (I expect false to be default for the property).
(Note that the 'bold' is not set for the standard default action, File History.)

You are commenting on a core problem in the current implementation, which is valid but not something I tried to address here.
I primarily want to clean up the order for next release and fix that items were bold when the should not be.
Can we settle on the order and fixes to the current implementation in this PR, it has enough changes already?

Going forward, I would like to remove setting for OpenSubmoduleDiffInSeparateWindow (always enable, the name is misleading now) and use DiffTool as the default doubleclick. With this PR, the first item is also the default and more often what I believe is expected (except for RevFile, better to change for consistency?).
The _openSubmoduleMenuItem could be duplicated to all forms too as well as the bold handling, but that will duplicate a number of lines of code.

RussKie: Sounds sensible. Please add a "TODO" in the code.

gerhardol: Added a TODO, something should be added to the final commit message too (squash should probably be done, describing the shanges)

RussKie:
Merge away

gerhardol:
Any comments on the unified order of items? 
@mstv usually have something to say.

RussKie:
Looks good to me :+1:
Awesome job btw

gerhardol:
Squashed (switch order for SubmoduleUpdate and Commit), rebased
Long message...
This can be improved

gerhardol:
@msftbot merge in 12 hours

mstv:
Indended change? Also in `RevisionDiffControl.Designer.cs`?
grafik

gerhardol:
>> Squashed (switch order for SubmoduleUpdate and Commit), rebased

> Indended change? Also in RevisionDiffControl.Designer.cs?

Yes, Update is likely more likely with the most common use of submodules and Commit is the least, as I see it.

But good catch if you missed my note, something better to compare rebase is needed #7909. In this case these files could be compared with the help of hopefully soon merged #7923 (even better with #7899)

PR #7968, opened 2020-04-10
Fix branch rendering

RussKie:
Addendum to #7882

## Proposed changes

- Unify the use of "Git Extensions"
- Unify branch rendering in recent repo dropdown
- Unify repository binding

### Before

image

### After

image

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/RepositoryCurrentBranchNameProvider.cs
+++ b/GitUI/RepositoryCurrentBranchNameProvider.cs
@@ -1,0 +1,29 @@
+﻿using GitCommands;
+using GitCommands.Git;
+
+namespace GitUI
+{
+    public interface IRepositoryCurrentBranchNameProvider
+    {
+        string GetCurrentBranchName(string repositoryPath);
+    }
+
+    public sealed class RepositoryCurrentBranchNameProvider : IRepositoryCurrentBranchNameProvider
+    {
+        public string GetCurrentBranchName(string repositoryPath)
+        {
+            if (!AppSettings.ShowRepoCurrentBranch)
+            {
+                return string.Empty;
+            }
+
+            string branchName = GitModule.GetSelectedBranchFast(repositoryPath);
+            if (string.IsNullOrWhiteSpace(branchName) || branchName == DetachedHeadParser.DetachedBranch)
+            {
+                branchName = $"({Strings.NoBranch})";
+            }
+
+            return branchName;
+        }
+    }
+}
--- a/GitUI/Strings.cs
+++ b/GitUI/Strings.cs
@@ -42,6 +42,12 @@ namespace GitUI
         private readonly TranslationString _errorCaptionFailedDeleteFile = new TranslationString("Failed to delete file");
         private readonly TranslationString _errorCaptionFailedDeleteFolder = new TranslationString("Failed to delete directory");
 
+        private readonly TranslationString _noBranch = new TranslationString("no branch");
+        private readonly TranslationString _removeSelectedInvalidRepository = new TranslationString("Remove the selected invalid repository");
+        private readonly TranslationString _removeAllInvalidRepositories = new TranslationString("Remove all {0} invalid repositories");
+        private readonly TranslationString _open = new TranslationString("Open");
+        private readonly TranslationString _directoryIsNotAValidRepository = new TranslationString("The selected item is not a valid git repository.");
+
         // public only because of FormTranslate
         public Strings()
         {
@@ -99,5 +105,11 @@ namespace GitUI
 
         public static string ErrorCaptionFailedDeleteFile => _instance.Value._errorCaptionFailedDeleteFile.Text;
         public static string ErrorCaptionFailedDeleteFolder => _instance.Value._errorCaptionFailedDeleteFolder.Text;
+
+        public static string NoBranch => _instance.Value._noBranch.Text;
+        public static string RemoveSelectedInvalidRepository => _instance.Value._removeSelectedInvalidRepository.Text;
+        public static string RemoveAllInvalidRepositories => _instance.Value._removeAllInvalidRepositories.Text;
+        public static string Open => _instance.Value._open.Text;
+        public static string DirectoryInvalidRepository => _instance.Value._directoryIsNotAValidRepository.Text;
     }
 }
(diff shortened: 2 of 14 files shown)


GitUI/Strings.cs
```diff
@@ -42,6 +42,12 @@ internal sealed class Strings : Translate
         private readonly TranslationString _errorCaptionFailedDeleteFile = new TranslationString("Failed to delete file");
         private readonly TranslationString _errorCaptionFailedDeleteFolder = new TranslationString("Failed to delete directory");
 
+        private readonly TranslationString _noBranchTitle = new TranslationString("no branch");
+        private readonly TranslationString _directoryIsNotAValidRepositoryRemoveSelectedRepoCommand = new TranslationString("Remove the selected invalid repository");
```
gerhardol, line 46: Strings in Strings.cs should have the intention to be shared, the names are far to specific here to allow that

RussKie:
Ready. Should be squash-merged.

GitUI/Strings.cs
RussKie: Suggestions?

In WinForms we place all resources in a single file, e.g. https://github.com/dotnet/winforms/blob/master/src/System.Windows.Forms/src/Resources/SR.resx

RussKie:
> Sidepanel is not changed though.

Ah, yes! I thought I missed something. Thank you

GitUI/Strings.cs
RussKie: Actually a good point, we need to update the dashboard because it has the same strings.

GitUI/RepositoryCurrentBranchNameProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if (!AppSettings.ShowRepoCurrentBranch)
+            {
+                return string.Empty;
+            }
+
+            string branchName = GitModule.GetSelectedBranchFast(repositoryPath);
+            if (string.IsNullOrWhiteSpace(branchName) || branchName == DetachedHeadParser.DetachedBranch)
+            {
+                branchName = $"({Strings.NoBranch})";
```
mstv, line 23: Either wrap all branches in brackets or none.
But we can emphasize that "no branch" is not a branch by using angle brackets.
```suggestion
                branchName = $"<{Strings.NoBranch}>";
```

RussKie: Everywhere else we have it as "(no branch)".

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -600,8 +595,8 @@ private void listView1_MouseClick(object sender, MouseEventArgs e)
 
                 if (!_controller.IsValidGitWorkingDir(selected.Path))
                 {
-                    var dialogResult = MessageBox.Show(this, _directoryIsNotAValidRepository.Text,
-                                                             _directoryIsNotAValidRepositoryCaption.Text, MessageBoxButtons.YesNo,
+                    var dialogResult = MessageBox.Show(this, Strings.DirectoryInvalidRepository + "\r\n" + Strings.RemoveSelectedInvalidRepository,
```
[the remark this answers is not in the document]
RussKie: The original was: `"The selected item is not a valid git repository.\n\nDo you want to remove it from the recent repositories list?"`, so one line break is consistent with the original.
I'll replace it with `Environment.NewLine`.

GitUI/RepositoryCurrentBranchNameProvider.cs
mstv: > Everywhere else we have it as "(no branch)".

This is not the full truth. For the submodules in the ROT, all branches are appended in round brackets, i.e. there it is "no branch", else it would have been displayed as "((no branch))".

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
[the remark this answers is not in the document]
mstv: It was
```
        private readonly TranslationString _directoryIsNotAValidRepository = new TranslationString(
            "The selected item is not a valid git repository.\n\nDo you want to remove it from the recent repositories list?");
```

GitUI/Strings.cs
```diff
@@ -42,6 +42,12 @@ internal sealed class Strings : Translate
         private readonly TranslationString _errorCaptionFailedDeleteFile = new TranslationString("Failed to delete file");
         private readonly TranslationString _errorCaptionFailedDeleteFolder = new TranslationString("Failed to delete directory");
 
+        private readonly TranslationString _noBranch = new TranslationString("no branch");
+        private readonly TranslationString _removeSelectedInvalidRepository = new TranslationString("Remove the selected invalid repository");
+        private readonly TranslationString _removeAllInvalidRepositories = new TranslationString("Remove all {0} invalid repositories");
```
mstv, line 47:
```suggestion
        private readonly TranslationString _removeSelectedInvalidRepository = new TranslationString("Remove the selected invalid repository?");
        private readonly TranslationString _removeAllInvalidRepositories = new TranslationString("Remove all {0} invalid repositories?");
```

GitUI/RepositoryCurrentBranchNameProvider.cs
RussKie: The `Dashboard`, `FormCommit` or ROT do not currently use the same functionality, this is outside the scope of this PR, but it is certainly something to do.

mstv: The easiest and correct solution is to change `Strings.NoBranch`.
This does not need to be done in this PR.

gerhardol: This is a reason why I preferred to not do this change. It seems like "no branch" is the name of the branch. If you want to display the string (I would prefer to not see something that do not exist, second opinion), it must be more unique. Maybe [no branch]

RussKie:
Centralised invalid repo removal prompt:

* Before
image

* After
image
image

GitUI/RepositoryCurrentBranchNameProvider.cs
RussKie: Once we have unified "no branch" handling it will be easy to change, and we can change it to whatever we agree.
Right now I want to have a consistency.

RussKie:
I have broken the PR up, and removed unrelated changes. Those will be submitted separately.

PR #7972, opened 2020-04-11
Unify the use of "Git Extensions"

RussKie:
## Proposed changes

- Unify the use of "Git Extensions"

### DIFF
--- a/GitCommands/AppTitleGenerator.cs
+++ b/GitCommands/AppTitleGenerator.cs
@@ -43,7 +43,7 @@ namespace GitCommands
         {
             if (string.IsNullOrWhiteSpace(workingDir) || !isValidWorkingDir)
             {
-                return "Git Extensions";
+                return AppSettings.ApplicationName;
             }
 
             branchName = branchName?.Trim('(', ')') ?? "no branch";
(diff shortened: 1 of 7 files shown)


cc11112:
the root cause is when call Application.UserAppDataPath, dotnet library auto create sub folder "Git Extensions\Git Extensions"under AppData\Roaming.

{code}
 /// <summary>
        ///  Gets the path for the application data specific to the roaming user.
        /// </summary>
        /// <remarks>
        ///  Don't obsolete these. GetDataPath isn't on SystemInformation, and it provides
        ///  the Windows logo required adornments to the directory (Company\Product\Version)
        /// </remarks>
        public static string UserAppDataPath
            => GetDataPath(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData));

      
        /// <summary>
        ///  Returns a string that is the combination of the basePath + CompanyName + ProductName + ProductVersion. This
        ///  will also create the directory if it doesn't exist.
        /// </summary>
        private static string GetDataPath(string basePath)
        {
            string path = Path.Join(basePath, CompanyName, ProductName, ProductVersion);

            lock (s_internalSyncObject)
            {
                if (!Directory.Exists(path))
                {
                    Directory.CreateDirectory(path);
                }
            }

            return path;
        }
{code}

Issue #6521, opened 2019-05-01
git-status runs also when GE is minimized


gerhardol:
## Current behaviour
The commit count uses git-status and is triggered periodically:
* At changes, at least 3s in between (adjusted from the time needed to run the status)
* Every 5th minute if no changes

The command runs with --no-optional-locks and should not affect other commands directly. 
In 2.x and 3.0 the ignore directories are parsed. However, this check delayed updates and required quite some resources (both runtime and memory) and has been removed in 3.x.

However, if you are building and GE is minimized, git-status still runs periodically as well as periodic runs (it seems like the Symantec virus is triggered by git-status and constantly runs with 25% CPU for me).
I often have at least 4 GE instances running at the same time (used to be more, but the left panel submodule in 3.1 decreases a couple of instances).

This is not a major problem, but can be approved.

## Expected behaviour

If GE is minimized, it is not really necessary to update the commit count. It is nice if it is updated quickly after restoring the application, but it should not affect performance.

It is to my knowledge not simple to find if an application is minimized or hidden. It is simpl

A simple solution is to deactivate the commit count when FormBrowse is inactivated. (GitStatusMonitor has an "Active" state, probably a "Delayed" state is needed in this situation to limit git-status runs when switching active apps).

## Steps to reproduce

Start a few GE instances and check the command log.

## Screenshots

n/a

## Did this work in previous version of GitExtensions

no

## Environment

- Git Extensions 3.1.0
- Build 947078bec2cfb27fa7a6a3f04897928713d89a29 (Dirty)
- Git 2.21.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

## Diagnostics

n/a

mstv:
On the other hand, git-status should not be suppressed anymore during active FormCommit.
Currently, the status overlay in GE's taskbar icon is not updated (which has confused me several times).
(Note: The status overlay is not displayed if the taskbar is configured to show small icons.)

I agree that GE should be *nice* while not the active application. I think "application minimized" would be not be enough - at least for me: I nearly never minimize GE, I just switch to other applications.

I do not expect that GE updates its status while in the background. Though it might be a valid use case for other users.

> It is to my knowledge not simple to find if an application is minimized or hidden.

Is this really a problem with WinForms?

gerhardol:
> On the other hand, git-status should not be suppressed anymore during active FormCommit.
[...]

PR #7976, opened 2020-04-11
GitStatusMonitor: Avoid background updates if GUI is not visible

gerhardol:
Part of #6521

## Proposed changes

Avoid running git-status when GE Browse is minimized

## Test methodology
Check Git Command Log

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using System.Windows.Forms;
 using GitCommands;
+using GitCommands.Utils;
 using GitUIPluginInterfaces;
 
 namespace GitUI.CommandsDialogs.BrowseDialog
@@ -309,16 +311,29 @@ namespace GitUI.CommandsDialogs.BrowseDialog
             }
         }
 
+        private static bool IsMinimized()
+        {
+            if (!EnvUtils.RunningOnWindows())
+            {
+                return false;
+            }
+
+            var currentProcess = Process.GetCurrentProcess();
+            if (currentProcess is null)
+            {
+                return false;
+            }
+
+            return NativeMethods.IsIconic(currentProcess.MainWindowHandle).IsTrue();
+        }
+
         private void Update()
         {
             ThreadHelper.AssertOnUIThread();
 
-            if (CurrentStatus != GitStatusMonitorState.Running)
-            {
-                return;
-            }
-
-            if (_nextUpdateTime - Environment.TickCount > 0)
+            if (CurrentStatus != GitStatusMonitorState.Running
+                || _nextUpdateTime - Environment.TickCount > 0
+                || IsMinimized())
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


RussKie:
Should we also disable it if the app is not active? And resume when the app becomes active?

GitCommands/Utils/EnvUtils.cs
```diff
@@ -1,11 +1,15 @@
...
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using Microsoft.Win32;
 
 namespace GitCommands.Utils
 {
     public static class EnvUtils
     {
+        [DllImport("user32.dll")]
+        private static extern bool IsIconic(IntPtr handle);
```
RussKie, line 11: We define interops in the UI layer, have a look at GitUI/Interop.

gerhardol:
> Should we also disable it if the app is not active? And resume when the app becomes active?

No. FileSystemWatcher is paused when there is a change and the time still fires every 100 ms, so basically only the IsIconic() call would be not be needed.
(I would like to change the start/stop of GitStatusMonitor in a similar way to avoid the initial git-status when not really needed, but that is a later change.)

mstv:
> Should we also disable it if the app is not active? And resume when the app becomes active?

If you use the FileSystemWatcher, then GE is expected to update if it is visible.

gerhardol:
> If you use the FileSystemWatcher, then GE is expected to update if it is visible.

This change suspends the git-status call, when no longer minimized, git-status resumes (and FileSystemWatcher events are enabled again)

GitCommands/Utils/EnvUtils.cs
gerhardol: I added Interops, should probably be added to GitCommands instead (Interops only in GitUI now)

RussKie:
If we suppose that GitUI is Windows only and GitCommands could be portable,
then interops belong to the UI layer.

gerhardol:
> If we suppose that GitUI is Windows only and GitCommands could be portable,
then interops belong to the UI layer.

Add a UIEnv then? Or keep in GitStatusMonitor for now (TODO)?

GitCommands/Utils/EnvUtils.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
```
RussKie, line 3: We can revert this change

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -309,16 +311,29 @@ private void StartWatchingChanges(string workTreePath, string gitDirPath)
...
-            if (CurrentStatus != GitStatusMonitorState.Running)
+            var currentProcess = Process.GetCurrentProcess();
+            if (currentProcess is null)
             {
-                return;
+                return false;
             }
 
-            if (_nextUpdateTime - Environment.TickCount > 0)
+            return NativeMethods.IsIconic(currentProcess.MainWindowHandle) == Interop.BOOL.TRUE;
```
RussKie, line 327: We also have `.IsTrue()` extension
```suggestion
            return NativeMethods.IsIconic(currentProcess.MainWindowHandle).IsTrue();
```

RussKie:
> Add a UIEnv then? Or keep in GitStatusMonitor for now (TODO)?

I think what you have currently is good.
Unless you're asking about something else...

gerhardol:
> I think what you have currently is good.
Unless you're asking about something else...

Keeping then (the question was if to separate this environment specific to a separate module).
Rebased too.

gerhardol:
@msftbot merge in 24 hours

PR #7950, opened 2020-04-05
Uplift script handling

RussKie:
## Proposed changes

- Decouple script parsing from the revision grid. This will allow easier and simpler integration of other controls with the script manager (e.g. the left panel or the file tree)
- Extract user script binding to a context menu to an extension method. This will allow re-use of the functionality in other parts of the app (e.g. the left panel or the file tree). This is small step in the same direction as #6348 proposed by @ivangrek. Sadly we stalled on this work.
- Review the code and add multiple validations and add tests.

This work is foundational to allow re-use of user scripts in other parts of the app (e.g. the left panel or the file tree). Specifically I want to allow users to run custom scripts for files in the working tree, e.g. to achieve what was proposed in #6135, or open a file and/or folder in VS Code, or launch a project in a VS.

image

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7981, opened 2020-04-12
Redesign script config

RussKie:
## Proposed changes

- Redesign script configuration settings
- One of a key change is that script modifications are no longer "destructive" _until_ the settings are accepted via [OK] or [Apply] buttons. 
The race condition where a user has multiple instances of the app open and modifies scripts in several instances at once still remains, and it completely outside the scope of this PR. 
The last instance of the app to accept changes to scripts will overwrite all other changes may have been made in other instances. 

### Before

image

### After

image
image

## Test methodology 

- manual
- I may look at adding tests, but may do that at a later stage.

----

### DIFF
--- a/GitUI/Design/ExecutableFileNameEditor.cs
+++ b/GitUI/Design/ExecutableFileNameEditor.cs
@@ -1,0 +1,14 @@
+﻿using System.Windows.Forms;
+using System.Windows.Forms.Design;
+
+namespace GitUI.Design
+{
+    internal class ExecutableFileNameEditor : FileNameEditor
+    {
+        protected override void InitializeDialog(OpenFileDialog openFileDialog)
+        {
+            base.InitializeDialog(openFileDialog);
+            openFileDialog.Filter = "Executable file (*.exe;*.cmd;*.bat)|*.exe;*.cmd;*.bat|All files (*.*)|*.*";
+        }
+    }
+}
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ScriptsSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ScriptsSettingsPage.Designer.cs
@@ -37,7 +37,6 @@
             this.chdrNames = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.chdrEvents = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.chdrCommand = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
-            this.chdrArguments = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.panelButtons = new System.Windows.Forms.FlowLayoutPanel();
             this.btnAdd = new System.Windows.Forms.Button();
             this.btnDelete = new System.Windows.Forms.Button();
@@ -107,8 +106,7 @@
             this.lvScripts.Columns.AddRange(new System.Windows.Forms.ColumnHeader[] {
             this.chdrNames,
             this.chdrEvents,
-            this.chdrCommand,
-            this.chdrArguments});
+            this.chdrCommand});
             this.lvScripts.Dock = System.Windows.Forms.DockStyle.Fill;
             this.lvScripts.FullRowSelect = true;
             this.lvScripts.HeaderStyle = System.Windows.Forms.ColumnHeaderStyle.Nonclickable;
@@ -132,17 +130,12 @@
             // chdrEvents
             // 
             this.chdrEvents.Text = "Event";
-            this.chdrEvents.Width = 100;
+            this.chdrEvents.Width = 130;
             // 
             // chdrCommand
             // 
             this.chdrCommand.Text = "Command";
-            this.chdrCommand.Width = 100;
-            // 
-            // chdrArguments
-            // 
-            this.chdrArguments.Text = "Arguments";
-            this.chdrArguments.Width = 200;
+            this.chdrCommand.Width = 330;
             // 
             // panelButtons
             // 
@@ -228,6 +221,5 @@
         private System.Windows.Forms.TableLayoutPanel tableLayoutPanel1;
         private System.Windows.Forms.ColumnHeader chdrEvents;
         private System.Windows.Forms.ColumnHeader chdrCommand;
-        private System.Windows.Forms.ColumnHeader chdrArguments;
     }
 }
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ScriptsSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ScriptsSettingsPage.Designer.cs
@@ -84,6 +84,7 @@
             // 
             this.propertyGrid1.Dock = System.Windows.Forms.DockStyle.Fill;
             this.propertyGrid1.Location = new System.Drawing.Point(3, 276);
+            this.propertyGrid1.Margin = new System.Windows.Forms.Padding(3, 3, 40, 3);
             this.propertyGrid1.Name = "propertyGrid1";
             this.propertyGrid1.Size = new System.Drawing.Size(394, 259);
             this.propertyGrid1.TabIndex = 2;
@@ -96,7 +97,7 @@
             this.panel2.Dock = System.Windows.Forms.DockStyle.Fill;
             this.panel2.Location = new System.Drawing.Point(3, 3);
             this.panel2.Name = "panel2";
-            this.panel2.Size = new System.Drawing.Size(394, 259);
+            this.panel2.Size = new System.Drawing.Size(431, 259);
             this.panel2.TabIndex = 3;
             // 
             // lvScripts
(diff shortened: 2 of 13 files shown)


RussKie:
Current limitations:
* No localisation support. The propertygrid categories and descriptors are compile-time constants. There's probably a way to change them at run-time, but I will need to investigate it.
* No image preview in the propertygrid. I'm sure it is possible, but further investigation is required.

The first limitation is probably a bigger one, and I'll try to sort it out for the v4 release. The second one I wouldn't consider a high priority or blocking.

GitUI/Design/ExecutableFileNameEditor.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Windows.Forms.Design;
+
+namespace GitUI.Design
+{
+    internal class ExecutableFileNameEditor : FileNameEditor
+    {
+        protected override void InitializeDialog(OpenFileDialog openFileDialog)
+        {
+            base.InitializeDialog(openFileDialog);
+            openFileDialog.Filter = "Executable file (*.exe;*.cmd;*.bat)|*.exe;*.cmd;*.bat|All files (*.*)|*.*";
```
mstv, line 11: + "*.ps1"?

gerhardol:
> I'd like to see at least also Enabled, Event and perhaps Contexts in the list of scripts. => Vertical layout again?

Enabled could be seen in the list but more than that is harder (groups can handle one setting). 
This looks nicer (not like most GE forms...) but is less functional. This is no strong preference though
No comment on the code yet

RussKie:
> Surprisingly, the script execution does not work anymore:

This was reported in gitter as well. Likely regressed in #7950, I'll have a look



> I'd like to see at least also `Enabled`, `Event` and perhaps `Contexts` in the list of scripts.

`Enabled` is expressed as color, enabled => black, disabled => grey. The same as in `FormRemotes`.
Showing `Event` is a good idea. 
`Context` - what benefit in that? Especially if there may be multiple contexts, e.g. revision grid, file tree, left panel, etc.?



> Vertical layout again?
> (If we had more lists than this and the Revision Links, we would also need a generalized list editor.)

Apologies, I'm struggling to parse this.



> (groups can handle one setting).

I had initially added groups, the same as we have in `FormRemotes`. However that made ordering scripts (i.e. up/down) significantly harder and mentally taxing. 
I think the order in scripts can be very important, e.g. a user may have several scripts that react to the same event, and they need to be executed in a set order.


> This looks nicer (not like most GE forms...)

Thank you, we need to start somewhere. I hope that eventually we can make the rest of the UI look at least as good.

> ...but is less functional.

What is in your opinion become less functional? And why is the old design more functional?

GitUI/Design/ExecutableFileNameEditor.cs
RussKie: I considered that, but PowerShell scripts are executed by PowerShell interpreter, that must be invoked explicitly. E.g. `> powershell.exe .\somescript.ps1`

RussKie:
One more thing, we don't have any data<sup>1</sup>, but somehow I feel that scripts are very much underused. So the UI is built under an assumption that a majority of our users won't have more than a dozen of scripts.
 

--

1. except for what I observed myself and none of the users I observed was aware of scripts, let alone used them.

vbjay:
I use them I have a couple I use all the time like stage non whitespace and
get current that does the heavy lifting of getting all submodules and the
main repo pulled down and then pushed to my forks.

RussKie:
An alternative idea - scripts can be grouped by "Event", e.g.

```
--- AfterCommit ---
(icon) Fetch changes after commit

--- AfterPull ---
(icon) Update submodules after pull

--- ShowInUserMenuBar ---
(icon) Example

--- None ---
(icon) Open on GitHub
(icon) Fetch All Submodules
(icon) test
```

Obviously up/down ordering will be limited to the scope of a group, which probably makes sense.
Thoughts?

RussKie:
image

gerhardol:
I use one script all the time: submoudule update (sometimes I want to keep my checked out versions)
Other than that, I have found the triggers and available info to be a little too limited to use scripts a lot, so I use bash scripts.
There were some time since I looked in to this though.

The current view is functional, it shows a good overview in the grid. But it is not looking good, which is not important for first time users. 

With my usage, this change is not important, I can live with stepping through my scripts to see details, if Enabled is seen.
So @RussKie , I can give feedback but will probably not drive this change forward.

> An alternative idea - scripts can be grouped by "Event", e.g.

Other than Enable (Gray), this is the only dimension available. The trigger is probably not important enough to use that spot, it is more or less derived from the name anyway. Something like Show in Grid is dynamic and more important to me, but should not occupy a group.
Leave this unused for now.

RussKie:
So you would prefer something like this? (more columns can be added)
image

mstv:
> So would you prefer something like this? (more columns can be added)

Yep, this was what I meant with vertical layout.
But the current variant with `Enabled` -> Font color and column `Event` might suffice -- especially regarding the little usage.

Though I expect slightly increasing usage because we are going to add scripts to the context menus of files, folders and in the ROT.
Then I will not need to misuse the `Start shell` button in order to start VS Code.
And a column `Contexts` (plural) might help to get an overview and to find a specific script.

RussKie:
I can make it vertical, no dramas.



> Though I expect slightly increasing usage because we are going to add scripts to the context menus of files, folders and in the ROT.
> Then I will not need to misuse the `Start shell` button in order to start VS Code.

That's the plan :)
I'm still trying to work out how to expose "selected file/folder" to a script.

RussKie:
Btw, this is how you can configure VS Code:
image

mstv:
> Btw, this is how you can configure VS Code:

Thank you! Kind of an easter egg, i.e. a hidden feature. This should be turned from `Event` to a `Script context`.

vbjay:
image

RussKie:
> Kind of an easter egg, i.e. a hidden feature. This should be turned from `Event` to a `Script context`.

Don't get me started on this... 😭 
I'm not going to touch it until after we release v4.

pmiossec:
>Btw, this is how you can configure VS Code:

I just run the `code.cmd` (even if it require to specify the full path toward the file) and it works well.

RussKie:
Updated the PR and the screenshot

pmiossec:
Because space is limited in the grid, perhaps it could be interesting to merge the 'Command' and 'Arguments' in one column (and need to find a good name somewhat like 'full command').

mstv:
Whishlist:
- [ ] Icon selector
- [ ] Splitter of the PropertyGrid should have a sensible position, not centered 
- [ ] column `Event` should fit to the longest event
- [ ] focus often jumps from PropertyGrid to the table above, e.g. try typing the name of an icon
- [ ] suppress bold font in PropertyGrid (I know that it indicates non-default values. But I see no benefit for our use case.)

gerhardol:
This design is fine for me. The list has sufficient info (more columns could be added) and looks better.
I mostly agree with pmiossec and mstv's comments but that is a wishlist, so I could approve

Not adapted to dark theme: Icons need to be adapted and background set
image

RussKie:
Thank you all for the suggestions



> perhaps it could be interesting to merge the 'Command' and 'Arguments' in one column

I suppose... Not finding it convincing, but I can do that.

> * Icon selector

This may have to wait (if we go with this design). My colleagues, who understand this functionality well, are very busy, and I can't seem to figure this one out in the time I have.


> * Splitter of the PropertyGrid should have a sensible position, not centered

I am not aware of how nor think `PropertyGrid` allows that changed. 
It is an [insanely complex control](https://github.com/dotnet/winforms/blob/master/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs), but it is short on customisation options. That was one of the reasons for the initial vertical design.



> * column `Event` should fit to the longest event

Agree, I will certainly fix this.

> * focus often jumps from PropertyGrid to the table above, e.g. try typing the name of an icon

I have missed that. I will fix this.

> * suppress bold font in PropertyGrid (I know that it indicates non-default values. But I see no benefit for our use case.)

I actually added "defaults" to un-bold. I find it easier to see which properties I have configured.
...and to make all the value normal weight it would require lots of extra code: https://stackoverflow.com/a/7442412/2338036




> Not adapted to dark theme: Icons need to be adapted and background set

Sure :) I just need to find how to do it, haha.

RussKie:
> [x]  column `Event` should fit to the longest event
> [x] focus often jumps from PropertyGrid to the table above, e.g. try typing the name of an icon
> perhaps it could be interesting to merge the 'Command' and 'Arguments' in one column
> Not adapted to dark theme: Icons need to be adapted and background set

Addressed

gerhardol:
@NikolayXHD Regarding dark mode in ScriptSettingsPage.cs OnPageShown()
 * Images are read from resources and Adapt is generally applied. Will the adaption have to be conditional to certain images or is it the images that need to be fixed?
 * The background for the control (System.Windows.Forms.ListView) is incorrect (AdaptBackColor() is not called, but no change if done). Any hints what need to be done?

Not critical for the PR itself

RussKie:
Will hold off the merge until hear from @NikolayXHD

NikolayXHD:
Fixed icon borders, property grid category header color.

Did not manage to fix checkboxes in ListView for now.

image

RussKie:
Thank you all

RussKie:
Added a wiki page: https://github.com/gitextensions/gitextensions/wiki/Useful-scripts

Issue #8000, opened 2020-04-17
Review FileViewerTextTests.DoAutoCRLF tests


RussKie:
The input files have their EOL stripped in https://github.com/gitextensions/gitextensions/commit/0b00208914c1e126de69fdbf162bc2252113b344.

Review and correct the tests.

mstv:
I guess `eol` must be specified in `.gitattributes` for these files.

Issue #7417, opened 2019-11-13
Test DoAutoCRLF_should_not_unnecessarily_duplicate_line_ending opens external diff and fails


mstv:
## Current behaviour

The test `DoAutoCRLF_should_not_unnecessarily_duplicate_line_ending` fails occasionally for different arguments. Instead it opens my external diff viewer.

```
DoAutoCRLF_should_not_unnecessarily_duplicate_line_ending(input,"WindowsLines")
   Source: FileViewerTextTests.cs line 43
   Duration: 914 ms

  Message: 
      Expected string length 26 but was 0. Strings differ at index 0.
      Expected: "Line 1\r\nLine 2\r\n\r\nLine 3\r\n"
      But was:  <string.Empty>
      -----------^
```

```
DoAutoCRLF_should_not_unnecessarily_duplicate_line_ending(true,"WindowsLines")
   Source: FileViewerTextTests.cs line 43
   Duration: 423 ms

  Message: 
      Expected string length 26 but was 0. Strings differ at index 0.
      Expected: "Line 1\r\nLine 2\r\n\r\nLine 3\r\n"
      But was:  <string.Empty>
      -----------^
```

## Screenshot

![grafik](https://user-images.githubusercontent.com/36601201/68810604-e10b3a00-066e-11ea-9c27-1f5932ff3af4.png)

## Expected behaviour

test passes and does not open external diff

## Steps to reproduce

run tests of Debug build locally

## Did this work in previous version of GitExtensions

not sure

## Environment

- Git Extensions 3.3.0
- **Debug** Build ff78f708dad5bf857f880c1544dcb59194fe6e2a
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4018.0
- DPI 96dpi (no scaling)

## Diagnostics

`core.autocrlf = true` for the GE repo
`core.autocrlf` not set "globally for all repos"

gerhardol:
This problem has two parts, first that the test fails and secondly that a diff tool is started to view the result.

For the second part, this is ApprovalTests trying to be nice and search my system for every difftool it can find.
It is not the Git diff tool configuration, seem to be internal to Approval or VS test runner.
Unfortunately for me it finds TortoiseGitMerge.exe first but it cannot start the tool so I get a number of popups.
This may be locally configurable (I did not find it), but the behavior can be changed with an attribute.
[UseReporter(typeof(ClipboardReporter))]

[...]

PR #8013, opened 2020-04-19
Disable tests in DEBUG for DoAutoCRLF_should_not_unnecessarily_duplic…

gerhardol:
…ate_line_ending

Workaround for #8000

## Proposed changes

Disable test for DoAutoCRLF_should_not_unnecessarily_duplicate_line_ending in DEBUG
It often fails with empty output for unknown reason.
As it is an ApprovalTest, it tries to be friendly and opens a diff tool. That fails and is not easy to configure.
See discussion in #7417 

The test still runs in the build server, test kept in Release
(could be limited to the build server, but this change eliminates the annoyance for the developer in most situations as well as documents the problem.)

## Test methodology 

Disable flaky test

----

### DIFF
--- a/UnitTests/GitUI.Tests/Editor/FileViewerTextTests.cs
+++ b/UnitTests/GitUI.Tests/Editor/FileViewerTextTests.cs
@@ -36,6 +36,9 @@ namespace GitUITests.Editor
             _fileViewer.Dispose();
         }
 
+        // Temporarily disabled, see issue #8000
+        // These tests often fail on developer PCs, then ApprovalTest opens a diff viewer (where the open can fail too)
+#if !DEBUG
         [Test]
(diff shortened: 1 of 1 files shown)


UnitTests/GitUI.Tests/Editor/FileViewerTextTests.cs
```diff
@@ -36,6 +36,9 @@ public void TearDown()
             _fileViewer.Dispose();
         }
 
+        // Temporarily disabled, see issue #8000
+        // These tests often fails on developer PCs, then ApprovalTest opens a diff viewer (where the open can fail too)
```
mstv, line 40:
```suggestion
        // These tests often fail on developer PCs, then ApprovalTest opens a diff viewer (where the open can fail too).
```

gerhardol:
Obsoleted by #8041

PR #8015, opened 2020-04-19
Tweak commit hash of portable build

### LINKED ISSUE #7979: Portable build of PR has misleading commit hash
## Current behaviour

IIUIC, the PR is merged to the master in a temporary commit, whose hash is used in GE and the zip name then. This is quite pointless and misleading.
If we get bug reports, we must resolve the build number.

## Expected behaviour

Let GH point to the build of the unmerged commit.

mstv:
Fixes #7979

## Proposed changes

- Tweak the commit hash of portable builds both
  - in the name of zip file
  - in `ThisAssembly.Git.Sha` (by soft resetting git to the head commit of the PR)

## Screenshot

grafik

## Test methodology 

- run AppVeyor

----

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,10 @@ build:
 
 install:
 - cmd: git submodule update --init --recursive
+- cmd: echo %APPVEYOR_REPO_COMMIT%
+- cmd: echo %APPVEYOR_PULL_REQUEST_HEAD_COMMIT%
+  # if building a temporary merge with master, reset the git revision to the HEAD of the PR, keeping the files of the merge of course
+- cmd: if not "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" == "" if not "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" == "%APPVEYOR_REPO_COMMIT%" git reset --soft "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" --
 - cmd: echo /logger:"%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll">> Directory.Build.rsp
 - cmd: |-
     cd scripts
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -25,6 +25,9 @@ build:
 
 install:
 - cmd: git submodule update --init --recursive
+- cmd: echo %APPVEYOR_REPO_COMMIT%
+- cmd: echo %APPVEYOR_PULL_REQUEST_HEAD_COMMIT%
+- cmd: if not "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" == "" if not "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" == "%APPVEYOR_REPO_COMMIT%" git reset --soft "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%" --
```
RussKie, line 30: Could you please explain this line?
May be add a comment

mstv: Yes, done. I should have followed my own objections against having the documentation in the PR description.
Also FormAbout and the NBug reports referred to the temporary merge commit.

gerhardol: Why git-reset and not git-rev-parse "%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%^1"

pmiossec: Otherwise the Sha1 in the 'ThisAssembly' class populated by 'GitInfo' and so that is displayed in the application will still be at the wrong value.

mstv: Exactly. `APPVEYOR_PULL_REQUEST_HEAD_COMMIT` already contains the commit hash of interest.
The command is the _soft_ reset which does not affect the files in the working dir at all.

appveyor.yml
```diff
@@ -25,6 +25,10 @@ build:
 
 install:
 - cmd: git submodule update --init --recursive
+- cmd: echo %APPVEYOR_REPO_COMMIT%
+- cmd: echo %APPVEYOR_PULL_REQUEST_HEAD_COMMIT%
+  # if building a temorary merge with master, reset the git revision to the HEAD of the PR, keeping the files of the merge of course
```
RussKie, line 30: typo

mstv:
I guess the `git reset --soft` makes the translation check fail for PR builds.
As a workaround everything changed could be committed again before `English.xlf` is generated.

RussKie:
> I guess the `git reset --soft` makes the translation check fail for PR builds.

I should've thought of that... 🤦‍♂️ 

How about we do this?
Being on a merge commit, get the hash for the commit being merged in:
```
MINGW64 /c/Development/gitextensions (master)
$ git rev-list HEAD --parents --max-count=1                                                                                                                                                       6f2f0ba7adfff0363aceb19350221ab271df1c81 cd04b628bab1fc58cc8bf10283001a8be2e71725 c95fd78d446d016e3c5b506184fabc42d1896209

MINGW64 /c/Development/gitextensions (master)
$ git rev-list HEAD --parents --max-count=1 | cut -d' ' -f 3                                                                                                                                      c95fd78d446d016e3c5b506184fabc42d1896209
```

Then set it into a env:var, which is later used in our scripts (i.e. don't rely on AppVeyor's env vars, as we don't know how it uses them)?

PR #8026, opened 2020-04-21
Reset any branch to selected commit

### LINKED ISSUE #2876: 'Reset any branch to here' feature
Hey, I've been wondering if anyone would be interested in such a feature I use it for over a year now in custom builds and still like it :)

Most common scenario is as shown in the screenshot. You rebase your Feature branch on top of lets say master and want to push to master just the first two commits. I then reset local master to this selected commit and push master. There are other usecases but this is the most common for me.

reset-branch-to-here

pmiossec:
Fixes #2876
Implementation initially based on @mdonatas

## Proposed changes

- In the revision gird, being able to reset any branch (not only the one checked out) to the selected commit, using "git reset --force"

### Before

image

### After

image

The initial view when opening
image

The new reset form for "other branches"
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 0a69f31ca246ebe13a212db80476180c4ac662d8
- .NET Framework 4.8.4150.0
- DPI 192dpi (200% scaling)

----

### DIFF
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -228,6 +228,25 @@ namespace GitCommands
         }
 
         /// <summary>
+        /// Push a local reference to a new commit
+        /// This is similar to "git branch --force "branch" "commit", except that you get a warning if commits are lost
+        /// </summary>
+        /// <param name="repoPath">Full path to the repo</param>
+        /// <param name="gitRef">The branch to move</param>
+        /// <param name="targetId">The commit to move to</param>
+        /// <param name="force">Push the reference also if commits are lost</param>
+        /// <returns>The Git command to execute</returns>
+        public static ArgumentString PushLocalCmd(string repoPath, string gitRef, ObjectId targetId, bool force = false)
+        {
+            return new GitArgumentBuilder("push")
+            {
+                $"file://{repoPath}",
+                $"{targetId}:{gitRef}",
+                { force, "--force" }
+            };
+        }
+
+        /// <summary>
         /// Git Clone.
         /// </summary>
         /// <param name="central">Makes a bare repo.</param>
--- a/GitUI/HelperDialogs/FormResetAnotherBranch.cs
+++ b/GitUI/HelperDialogs/FormResetAnotherBranch.cs
@@ -1,7 +1,7 @@
 using System;
 using System.IO;
 using System.Linq;
-using System.Windows;
+using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
 using GitExtUtils.GitUI.Theming;
@@ -68,7 +68,7 @@ namespace GitUI.HelperDialogs
             var gitRefToReset = _localGitRefs.FirstOrDefault(b => b.Name == Branches.Text);
             if (gitRefToReset == null)
             {
-                MessageBox.Show(string.Format(_localRefInvalid.Text, Branches.Text), Strings.Error, MessageBoxButton.OK, MessageBoxImage.Error);
+                MessageBox.Show(string.Format(_localRefInvalid.Text, Branches.Text), Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
                 return;
             }
 
--- a/GitUI/HelperDialogs/FormResetAnotherBranch.cs
+++ b/GitUI/HelperDialogs/FormResetAnotherBranch.cs
@@ -1,0 +1,91 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Windows;
+using GitCommands;
+using GitCommands.Git;
+using GitExtUtils.GitUI.Theming;
+using GitUIPluginInterfaces;
+using ResourceManager;
+
+namespace GitUI.HelperDialogs
+{
+    public partial class FormResetAnotherBranch : GitModuleForm
+    {
+        private IGitRef[] _localGitRefs;
+        private readonly GitRevision _revision;
+        private readonly TranslationString _localRefInvalid = new TranslationString("The entered value '{0}' is not the name of an existing local branch.");
+
+        public static FormResetAnotherBranch Create(GitUICommands gitUiCommands, GitRevision revision)
+            => new FormResetAnotherBranch(gitUiCommands, revision ?? throw new NotSupportedException(Strings.NoRevision));
+
+        [Obsolete("For VS designer and translation test only. Do not remove.")]
+        private FormResetAnotherBranch()
+        {
+            InitializeComponent();
+        }
+
+        private FormResetAnotherBranch(GitUICommands gitUiCommands, GitRevision revision)
+            : base(gitUiCommands)
+        {
+            _revision = revision;
+
+            InitializeComponent();
+            Cancel.Text = Strings.Cancel;
+            Ok.Text = Strings.OK;
+            InitializeComplete();
+
+            labelResetBranchWarning.SetForeColorForBackColor();
+        }
+
+        private IGitRef[] GetLocalBranchesWithoutCurrent()
+        {
+            var currentBranch = Module.GetSelectedBranch();
+            var isDetachedHead = currentBranch == DetachedHeadParser.DetachedBranch;
+
+            return Module.GetRefs(false)
+                .Where(r => r.IsHead && (isDetachedHead || r.LocalName != currentBranch))
+                .ToArray();
+        }
+
+        private void FormResetAnotherBranch_Load(object sender, EventArgs e)
+        {
+            _localGitRefs = GetLocalBranchesWithoutCurrent();
+
+            Branches.DisplayMember = nameof(IGitRef.Name);
+            Branches.Items.AddRange(_localGitRefs);
+
+            commitSummaryUserControl.Revision = _revision;
+
+            if (Branches.Text.Length == 0)
+            {
+                Branches.DroppedDown = true;
+            }
+        }
+
+        private void Ok_Click(object sender, EventArgs e)
+        {
+            var gitRefToReset = _localGitRefs.FirstOrDefault(b => b.Name == Branches.Text);
+            if (gitRefToReset == null)
+            {
+                MessageBox.Show(string.Format(_localRefInvalid.Text, Branches.Text), Strings.Error, MessageBoxButton.OK, MessageBoxImage.Error);
+                return;
+            }
+
+            var repoPath = Path.GetFullPath(Module.WorkingDir);
+            var args = GitCommandHelpers.PushLocalCmd(repoPath, gitRefToReset.CompleteName, _revision.ObjectId, ForceReset.Checked);
+            var isSuccess = FormProcess.ShowDialog(this, args);
+
+            if (isSuccess)
+            {
+                UICommands.RepoChangedNotifier.Notify();
+                Close();
+            }
+        }
+
+        private void Cancel_Click(object sender, EventArgs e)
+        {
+            Close();
+        }
+    }
+}
--- a/GitUI/HelperDialogs/FormResetAnotherBranch.cs
+++ b/GitUI/HelperDialogs/FormResetAnotherBranch.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
+using GitExtUtils.GitUI;
 using GitExtUtils.GitUI.Theming;
 using GitUIPluginInterfaces;
 using ResourceManager;
@@ -31,8 +32,14 @@ namespace GitUI.HelperDialogs
             _revision = revision;
 
             InitializeComponent();
-            Cancel.Text = Strings.Cancel;
-            Ok.Text = Strings.OK;
+
+            pictureBox1.Image = DpiUtil.Scale(pictureBox1.Image);
+            labelResetBranchWarning.AutoSize = true;
+            labelResetBranchWarning.Dock = DockStyle.Fill;
+
+            Height = tableLayoutPanel1.Height + tableLayoutPanel1.Top;
+            tableLayoutPanel1.Dock = DockStyle.Fill;
+
             InitializeComplete();
 
             labelResetBranchWarning.SetForeColorForBackColor();
--- a/GitUI/HelperDialogs/FormResetAnotherBranch.cs
+++ b/GitUI/HelperDialogs/FormResetAnotherBranch.cs
@@ -38,6 +38,16 @@ namespace GitUI.HelperDialogs
             labelResetBranchWarning.SetForeColorForBackColor();
         }
 
+        private void Application_Idle(object sender, EventArgs e)
+        {
+            Application.Idle -= Application_Idle;
+
+            if (Branches.Text.Length == 0)
+            {
+                Branches.DroppedDown = true;
+            }
+        }
+
         private IGitRef[] GetLocalBranchesWithoutCurrent()
         {
             var currentBranch = Module.GetSelectedBranch();
@@ -57,10 +67,7 @@ namespace GitUI.HelperDialogs
 
             commitSummaryUserControl.Revision = _revision;
 
-            if (Branches.Text.Length == 0)
-            {
-                Branches.DroppedDown = true;
-            }
+            Application.Idle += Application_Idle;
         }
 
         private void Ok_Click(object sender, EventArgs e)
--- a/GitUI/HelperDialogs/FormResetCurrentBranch.cs
+++ b/GitUI/HelperDialogs/FormResetCurrentBranch.cs
@@ -22,13 +22,16 @@ namespace GitUI.HelperDialogs
             Hard
         }
 
+        public static FormResetCurrentBranch Create(GitUICommands commands, GitRevision revision, ResetType resetType = ResetType.Mixed)
+            => new FormResetCurrentBranch(commands, revision ?? throw new NotSupportedException(Strings.NoRevision), resetType);
+
         [Obsolete("For VS designer and translation test only. Do not remove.")]
         private FormResetCurrentBranch()
         {
             InitializeComponent();
         }
 
-        public FormResetCurrentBranch(GitUICommands commands, GitRevision revision, ResetType resetType = ResetType.Mixed)
+        private FormResetCurrentBranch(GitUICommands commands, GitRevision revision, ResetType resetType)
             : base(commands)
         {
             Revision = revision;
@@ -65,11 +68,6 @@ namespace GitUI.HelperDialogs
 
         private void FormResetCurrentBranch_Load(object sender, EventArgs e)
         {
-            if (Revision == null)
-            {
-                throw new Exception("No revision");
-            }
-
             _NO_TRANSLATE_BranchInfo.Text = string.Format(_branchInfo.Text, Module.GetSelectedBranch());
             commitSummaryUserControl1.Revision = Revision;
         }
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
@@ -33,6 +33,7 @@ namespace GitUI
             this.checkoutBranchToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.mergeBranchToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.resetCurrentBranchToHereToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.resetAnotherBranchToHereToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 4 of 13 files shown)


gerhardol:
I rather would like to see a new menu item (probably in Advanced): Reset a branch to here...
(The label is misleading: Reset Current branch to here..., and resetting the current branch is still more common - at least for me).

mstv:
Yes, I want to have two menu items: "Reset current branch to here..." as before and "Reset a branch to here..." right below, not hidden in `Advanced`. One level of submenu is 

`build.cmd -loc` ignores `FormResetBranch`, @RussKie. At least the `GitUITests.TranslationTest` fails.

gerhardol:
This PR eliminates the possibility to reset current branch, useful to remove garbage in Index/worktree
Sure, you can still use Commands-Reset, Commit-Reset or RevDiff-select-in-diff-Reset-to-first

Text incorrect in DarkTheme
Add the following line in FormResetBranch() constructor ( I considered pushing to your branch).

```            labelResetBranchWarning.SetForeColorForBackColor();
```
(the overlapping text is fixed in master for this PR)
before:
image

pmiossec:
@gerhardol 

>This PR eliminates the possibility to reset current branch

No. If you select the first line which is in bold "Current branch: {name of the current branch}", you've got the exact same behavior as before...

> ( I considered pushing to your branch).

You could. As @mstv already did it. This PR is intended for contributions as I can't have the time to fix all that should be.

@mstv & @gerhardol 

>, I want to have two menu items

I don't have a strong feeling about that. The initial development was not like that.
I think it's also slightly better for a good user understanding of what he is doing.  So let's go with that.

I like the quick branch selection in a submenu (that's why I kept it reducing the list to the 15 or 20 more recently updated). If that's still a performance problem, probably we could remove it. Not sure if I will like it but it should not be a big deal as it is not a feature intended to be used very very frequently...

gerhardol:
> > This PR eliminates the possibility to reset current branch

> No. If you select the first line which is in bold "Current branch: {name of the current branch}", you've got the exact same behavior as before...

Pushed an update to change
image


> > , I want to have two menu items

> I like the quick branch selection in a submenu (that's why I kept it reducing the list to the 15 or 20 more recently updated). If that's still a performance problem, probably we could remove it. Not sure if I will like it but it should not be a big deal as it is not a feature intended to be used very very frequently...

Yes, this is good, this is not a performance problem

Maybe a selection with there items is better, where Recent branch are listed

But the current implementation may be OK, will digest

mstv:
Now with two separate menu items and much less code.
But with the automatically dropped down branch list, it should allow equally quick operation.

grafik
grafik

GitUI/HelperDialogs/FormResetBranch.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (Revision == null)
+            {
+                throw new Exception("No revision");
+            }
+
+            commitSummaryUserControl.Revision = Revision;
+
+            if (Branches.Text.Length == 0)
+            {
+                Branches.DroppedDown = true;
```
gerhardol, line 47: I prefer to not pull down the branches by default, it seems like unintentional to get the list in your face.
Resetting any branch is not done very often, it should not be a problem to click once more

pmiossec:
> I can push the changes, but let us agree first

I don't have a strong point of view on that (except that's a useful feature that I would like to see in GE because I needed it in the past 😉 ).

So I'm Ok and  will let @mstv share its point of view.

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
```diff
@@ -228,6 +230,14 @@ private void InitializeComponent()
             this.resetCurrentBranchToHereToolStripMenuItem.Text = "Reset current branch to here...";
             this.resetCurrentBranchToHereToolStripMenuItem.Click += new System.EventHandler(this.ResetCurrentBranchToHereToolStripMenuItemClick);
             // 
+            // resetAnotherBranchToHereToolStripMenuItem
+            // 
+            this.resetAnotherBranchToHereToolStripMenuItem.Image = global::GitUI.Properties.Images.ResetCurrentBranchToHere;
+            this.resetAnotherBranchToHereToolStripMenuItem.Name = "resetAnotherBranchToHereToolStripMenuItem";
+            this.resetAnotherBranchToHereToolStripMenuItem.Size = new System.Drawing.Size(223, 22);
+            this.resetAnotherBranchToHereToolStripMenuItem.Text = "Reset another branch to here...";
```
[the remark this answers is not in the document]
mstv: > I prefer "Reset any branch to here..."

no objections

GitUI/HelperDialogs/FormResetBranch.cs
mstv: > I prefer to not pull down the branches by default.

A branch must be selected. What do the GUI designers say?

pmiossec: >What do the GUI designers say?

I'm absolutely not a GUI designers but 😃 ...

* the display when the dropdown is opening is a little strange. The list is displayed before the rest of the form...
* all the form could be totally hidden by the branch list so, on the first use, the user could have difficulties to understand how the form works
* but once it is used one time, I find it mostly good.

So, even if it is not a full win, that's interesting.
But once again, not a strong point of view for either of the solution chosen...

gerhardol: I would keep the box empty and let the user press the arrow to see the branches.
No change now, I keep my comment

gerhardol:
Rebased and squashed, with the name change I wanted but without the change to dropdown
I feel this is ready to merge (with the 'dropdown' the discussion maybe open) and would like to include this if we update the translations for 3.4

pmiossec:
I think we must remove the currently checked out branch from the list because that's not allowed and will always fail (and must be done with the normal reset feature)

mstv:
@RussKie, what do you think about the automatical drop down?

RussKie:
> Now with two separate menu items and much less code.
> But with the automatically dropped down branch list, it should allow equally quick operation.
> 
> grafik
> grafik

This looks interesting.

pmiossec:
>I think we must remove the currently checked out branch from the list because that's not allowed and will always fail (and must be done with the normal reset feature)

I did it. Have a look.

gerhardol:
This PR uses ```git push ./```, ```git branch --force``` could have been used too.
The advantage with "push" should be that you get a warning if commits are lost (you have to use --force then), otherwise the behavior.
I expected that it would be possible to move the current branch, but it is not poaaible why the latest blocking of current branch is OK. Maybe my change from Any->Another should be reverted.

However, at work I got the error message "unknown remote ./", will investigate later.

mstv:
> However, at work I got the error message "unknown remote ./", will investigate later.

Exactly the same for me: https://github.com/gitextensions/gitextensions/issues/2876#issuecomment-616775518 Join the club! :(
So we could switch to the alternate git command you proposed.
I think it will not be too dangerous: One must select a branch and click OK. ... and there is the reflog.

pmiossec:
Another thing to do: don't close the "reset form" if reset failed. It's boring to have to reopen and select the branch just because you forgot to check "Force reset".

vbjay:
So frankly I think pushing from this type of functionality is foolish. I
think all resets should be done locally and then a push can be done with
force with lease if needed later. I'd much rather see the changed branch
info and then push second instead of all at once.

mstv:
This is not about remote branches, Jay. `Force` has a different meaning here. See the explanation label in the screenshot of the dialog above.

Can you exchange a branch of a GH PR? I guess this is impossible.

vbjay:
> This PR uses ```git push ./```, ```git branch --force``` could have been used too.
> The advantage with "push" should be that you get a warning if commits are lost (you have to use --force then), otherwise the behavior.
> I expected that it would be possible to move the current branch, but it is not poaaible why the latest blocking of current branch is OK. Maybe my change from Any->Another should be reverted.
> 
> However, at work I got the error message "unknown remote ./", will investigate later.
@mstv   it was this that made me want to make sure we don't do remote branches.

pmiossec:
>Another thing to do: don't close the "reset form" if reset failed.

done.

gerhardol:
> Exactly the same for me: #2876 (comment) Join the club! :(

using ./ is not mentioned in the documentation and I find no relevant search hits for this specific situation.
https://git-scm.com/docs/git-push#_git_urls
./ can be prefixing files for sure, but for local repos file:// is mentioned

The following fails for me (from mstv example) - same Git (2.26.2)
```git.exe push ./ "303f3f4023cdbf72f0c3290f603871e320cd0ed8":"refs/heads/x" -f```

The following is OK:
```git.exe push  file:///path/to/repo "303f3f4023cdbf72f0c3290f603871e320cd0ed8":"refs/heads/x"```
(```--force``` is optional)
path can be both file:///c/dir/repo or file://c:/dir/repo
or
```git.exe branch --force x "303f3f4023cdbf72f0c3290f603871e320cd0ed8"```

Unless someone beats me, i will update with the first variant and add some comments in the code (push is slightly safer)

**Comment by gerhardol** [comment]
Updated, also ArgumentBuilder and revert to Reset Another Branch, rebased

When reviewed, squash and merge

File: GitCommands/Git/GitCommandHelpers.cs [hunk 3]
```diff
@@ -227,6 +227,25 @@ public static ArgumentString ResetCmd(ResetMode mode, string commit = null, stri
...
+        /// <param name="repoPath">Full path to the repo</param>
+        /// <param name="gitRef">The branch to move</param>
+        /// <param name="targetId">The commit to move to</param>
+        /// <param name="force">Push the reference also if commits are lost</param>
+        /// <returns>The Git command to execute</returns>
+        public static ArgumentString PushLocalCmd(string repoPath, string gitRef, ObjectId targetId, bool? force = false)
+        {
+            return new GitArgumentBuilder("push")
+            {
+                string.Format("file://{0}", repoPath),
```
pmiossec, line 243: in this case, string interpolation is more readable...

gerhardol: this and next line can be combined?

mstv: I think Philippe mainly means:
```suggestion
                $"file://{repoPath}",
                $"{targetId}:{gitRef}",
```

GitCommands/Git/GitCommandHelpers.cs
```diff
@@ -227,6 +227,25 @@ public static ArgumentString ResetCmd(ResetMode mode, string commit = null, stri
             };
         }
 
+        /// <summary>
+        /// Push a local reference to a new commit
+        /// This is similar to "git branch --force "branch" "commit", except that uo get a warning if commits are lost
```
mstv, line 232: final nit:
```suggestion
        /// This is similar to "git branch --force "branch" "commit", except that you get a warning if commits are lost
```

pmiossec:
@gerhardol 

>However, at work I got the error message "unknown remote ./", will investigate later.

I got it also (I suspect that a `.` in the name of the branch causes that).
Your change fixes it, so 👍

gerhardol:
@msftbot merge in 14 days

gerhardol:
Squashed
I believe this is ready and should be merged - for me it could be included in 3.4, but it affects translations
I setup a merge in two weeks to give some time for @RussKie to adjust

pmiossec:
I have made some more work on this PR...
Please, do a last review. Especially the error message to see if it's good english.

I have:
* done missing translation
* renamed form (sorry for the review)
* cleaned and simplified code in `FormResetAnotherBranch`
* Move ref filtering that was done partly in 2 places (before form opening and when form opening) in only one place.

mstv:
The AppVeyor build is broken:
```
Downloading vswhere
task.logissue type=error(InitializeToolset) The remote server returned an error: (403) Forbidden.
 ```
All `.binlog`s but `localise.binlog` are missing.

File: GitUI/HelperDialogs/FormResetAnotherBranch.cs [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+                .Where(r => r.IsHead && (isDetachedHead || r.LocalName != currentBranch))
+                .ToArray();
+        }
+
+        private void FormResetAnotherBranch_Load(object sender, EventArgs e)
+        {
+            if (_revision == null)
+            {
+                throw new Exception(Strings.NoRevision);
+            }
```
mstv, line 45: I'd rather do this check in the ctor and use the slightly more understandable `OperationNotSupportedException`.

GitUI/HelperDialogs/FormResetCurrentBranch.cs
```diff
@@ -67,7 +67,7 @@ private void FormResetCurrentBranch_Load(object sender, EventArgs e)
         {
             if (Revision == null)
             {
-                throw new Exception("No revision");
+                throw new Exception(Strings.NoRevision);
```
mstv, line 70: ditto

pmiossec:
Done. Squashed. Rebased.

mstv:
Run `build.cmd -loc` again, please.

pmiossec:
>Run build.cmd -loc again, please.

Yes, I saw and I did it but the result surprised me. I'm trying to understand...

gerhardol:
> I have made some more work on this PR...

Then I made a change, to move OK, Cancel to Strings

> Yes, I saw and I did it but the result surprised me. I'm trying to understand...

Something is strange with the translation app...
When trying this a couple of days ago, the moved strings were not removed from English.xlf, when trying earlier today all in ResetAnother was removed but now the result was as expected.

pmiossec:
>Something is strange with the translation app...

For my case, the behavior was normal.
That's the change asked by @mstv  to check the `revison` parameter and throw that prevent the translation of the form and so removed all the string from the translation file.

The problem didn't occurred for `FormResetCurrentBranch` so I was finding that strange. I have adopted the same strategy that is to add a private parameterless constructor fot the translation App and the Designer.

GitUI/HelperDialogs/FormResetCurrentBranch.cs
```diff
@@ -31,7 +31,7 @@ private FormResetCurrentBranch()
         public FormResetCurrentBranch(GitUICommands commands, GitRevision revision, ResetType resetType = ResetType.Mixed)
             : base(commands)
         {
-            Revision = revision;
+            Revision = revision ?? throw new NotSupportedException(Strings.NoRevision);
```
RussKie, line 34: Throwing in the ctor prevents the object from being disposed

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1493,6 +1493,17 @@ private void ResetCurrentBranchToHereToolStripMenuItemClick(object sender, Event
...
 
+        private void ResetAnotherBranchToHereToolStripMenuItemClick(object sender, EventArgs e)
+        {
+            if (LatestSelectedRevision == null)
+            {
+                return;
+            }
+
+            var frm = new FormResetAnotherBranch(UICommands, LatestSelectedRevision);
+            frm.ShowDialog(this);
```
RussKie, line 1504:
```suggestion
            using (var frm = new FormResetAnotherBranch(UICommands, LatestSelectedRevision))
            {
                frm.ShowDialog(this);
            }
```

gerhardol: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations
```
            using var frm = new FormResetAnotherBranch(UICommands, LatestSelectedRevision);
            frm.ShowDialog(this);
```

GitUI/HelperDialogs/FormResetCurrentBranch.cs
gerhardol: We could change back to throwing in the load again, but you get the NBug anyway?

mstv: The class must not be used without a revision.

pmiossec: Throwing in the load method seems a good compromise. Is everyone OK?

gerhardol: > Throwing in the load method seems a good compromise. Is everyone OK?

Yes, change back to that, it is slightly better as a template and if you dismiss the NBug
Whoever doing it, squash and rebase as well so it is ready for 3.4

mstv: I won't block this. Though you cannot convince me that this was a clear approach. It is not worth having a further discussion on the philosophy of C# regarding this.
But the `null` check could be done by a factory method, which then calls the private ctor.

mstv: done -- not squashed and rebased yet for easier review

RussKie:
The warning label overlaps and partially covers buttons (DPI 144dpi (150% scaling))
image


I think it would be more appropriate to place the label above all other elements. Being the last element it feels like something is "_btw fyi_" rather than "_IMPORTANT!_". The warning label wrapping looks awkward, but don't have a better suggestion at this stage.
The icon should be "warning" rather than "info", and it needs to be scaled to be of a greater size.

GitUI/HelperDialogs/FormResetCurrentBranch.cs
```diff
@@ -22,13 +22,10 @@ public enum ResetType
             Hard
         }
 
-        [Obsolete("For VS designer and translation test only. Do not remove.")]
-        private FormResetCurrentBranch()
-        {
-            InitializeComponent();
-        }
```
RussKie, line 29: Without a default ctor it is impossible to open the form in the designer

mstv: Sorry, this was for those who can read. Restored.

mstv:
`git reset && git checkout` modifies the `English.xlf` if it was changed both in the master and in the PR branch. But it must not be changed by the tweaking of the commit hash.
That's why I restored the approach using `git commit`, which seems to work.

Remains Igor's request to move the warning message. Form layout is none of my strength, especially not using the VS designer. Volunteers?

pmiossec:
An attempt to improve UI:
image

1. Update form design to minimize risk of label overwrite

* Increase form width (and so label and commit description box width)
* Put warning label before checkbox
* Remove label autosize (no more overlapping possible)
* increase form width to prevent wrapping in most of the case
* indent checkbox to better express dependency with warning text above

2. More explicit checkbox text "Force reset for a non-fast-forward reset"

3. Non luck in trying to resize picturebox :(

Point of view?

RussKie:
> * Increase form width (and so label and commit description box width)
> * Remove label autosize (no more overlapping possible)

This may not work well with translations, so the label needs to be fluid, auto scale/layout.
I'll get some time and send you my proposal.

> \2. More explicit checkbox text "Force reset for a non-fast-forward reset"

Nice!

RussKie:
> Form layout is none of my strength, especially not using the VS designer. Volunteers?

I'll give a hand.

pmiossec:
>This may not work well with translations, so the label needs to be fluid, auto scale/layout.

@RussKie another try before you put an hand on that.
Warning label set again at auto resize.
**But put also the form as auto resize** to keep the label on 2 lines (much more readable) and prevent vertical overlapping.

Example of the worst case:
image

If that's not good, I let you handle that ;)

pmiossec:
>Squash and rebase on master for another try?

I have squashed all commits except one (by @mstv ) related to the localization in the AppVeyor build.
I don't really get the goal of this commit and I don't know if we can/must remove it from this PR.
Or if this should end up in another PR...

mstv:
> The appveyor commit should be removed

Moved to another PR!

mstv:
This PR did not pass the translation check after changes to the `English.xlf` on `master`.

RussKie:
Just to let you know I'm reviewing the changes.

pmiossec:
>This PR did not pass the translation check after changes to the English.xlf on master.

It seems that the build passed. Is there something that I don't understand?

RussKie:
image

mstv:
Since the rebase of this branch, of course there are no concurrent changes to `English.xlf` anymore. Hence the build succeeds now without #8132.

RussKie:
Please try my changes. Few commits for the ease of reviewing, squash if/when happy.

RussKie:
Aye, good point.
The less elements the better.

pmiossec:
Closing in favor of an upcoming PR to merge in `3.4` branch instead...

gerhardol:
You can just edit this PR...

pmiossec:
> You can just edit this PR...

I think we can't change the branch to merge to...

gerhardol:
For open PRs you can

pmiossec:
Ok. I never found how. I will try to find for the next time (as I can't reopen this PR because it has been force pushed. GitHub don't let me.)

See you in #8140

mstv:
I have not noticed this was possible. But the `Edit` button beside the title really does.

PR #8037, opened 2020-04-25
Apply dark theme for check boxes in NativeListView

NikolayXHD:
## Proposed changes

The solution used to detect when check boxes are painted is dirty.

The check boxes are painted by windows once, when ListView control is created.

Theming hook method, the one that desides whether rendering shoud be overriden, receives `IntPtr hTheme` - a handle of theme data. It's always a pain to find out which type of control is being painted, because AFAIK there is no API to infer it from `hTheme`.

One workaround that worked in all previous cases, was to exploit the fact that `OpenThemeData(control.Handle, partName)` returns the same `hTheme`  until `CloseThemeData` is called with same `control.Handle` and `partName` values.

`hTheme` depends only on 2 parameters - the window class of the control, and `partName`.

So I would normally `OpenThemeData(listView.Handle, "ListView")`, remember which `hTheme` was returned and then any ListView rendereng in application lifetime would happen with this known `hTheme` value.

In this PR however I was out of luck finding which window class + part name are used by Windows when rendering the checkBoxes. A brute force search failed. 

So eventually I used a workaround: I know that that CheckBoxes are painted when NativeListView is initializing. So if the hooked `DrawThemeBackgroundEx` is called while a NativeListView with checkboxes is being initialized, the theming override is applied.

### Before

image

### After

image

## Test methodology

manually

## Test environment(s)

Windows 10, default dpi (96)

----

### DIFF
--- a/GitUI/Interops/DTBGOPTS.cs
+++ b/GitUI/Interops/DTBGOPTS.cs
@@ -7,8 +7,8 @@ namespace System
         [StructLayout(LayoutKind.Sequential)]
         public struct DTBGOPTS
         {
-            public int dwSize;
-            public int dwFlags;
+            public uint dwSize;
+            public uint dwFlags;
             public RECT rcClip;
         }
     }
--- a/GitUI/Interops/DTBGOPTS.cs
+++ b/GitUI/Interops/DTBGOPTS.cs
@@ -1,0 +1,15 @@
+﻿using System.Runtime.InteropServices;
+
+namespace System
+{
+    internal static partial class NativeMethods
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        public struct DTBGOPTS
+        {
+            public int dwSize;
+            public int dwFlags;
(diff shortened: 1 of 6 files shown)


GitUI/Interops/DTBGOPTS.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    internal static partial class NativeMethods
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        public struct DTBGOPTS
+        {
+            public int dwSize;
+            public int dwFlags;
+            public RECT rcClip;
+        }
```
RussKie, line 13: https://docs.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#common-windows-data-types

```suggestion
        [StructLayout(LayoutKind.Sequential)]
        public struct DTBGOPTS
        {
            public uint dwSize;
            public uint dwFlags;
            public RECT rcClip;
        }
```

PR #8038, opened 2020-04-25
Fix for #8030 URL in "Edit Remote Details"is case insensitive

### LINKED ISSUE #8030: URL in "Edit Remote Details" is case insensitive when checking for modifications
## Current behaviour

When changing a remote's url (eg migrating a repo from one server to another) it checks if the url has changed before saving. This check is not case sensitive. So if the url is case sensitive and changes only the casing, it will not save the change.

## Expected behaviour

Check the URL in a case sensitive manner.

## Steps to reproduce

Change Url Casing and click "Save Changes".

## Screenshots
image

N/A

## Environment

- Git Extensions 3.3.1.7897
- Git 2.26.2.windows.1
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

## Diagnostics

N/A

## PR Incoming
I should have time after work today to put together a PR for this. I have not looked at the code yet, but I can't imagine this would be a major change.  Is there any reason this should not always be case-insensitive?

digitalcoyote:
Fixes #8030

## Proposed changes

- Change SaveRemote to consider case when deciding if the url text has changed
- Override Autocomplete behavior in ComboBox to prevent it from resetting case-only changes

## Test methodology

- Manually tested by changing the url by 1,2 and 3 letters.
- Manually verified that autocomplete still works

## Test environment(s)

- GIT  2.26.2
- Windows 10 latest update

----

### DIFF
--- a/GitUI/UserControls/CaseSensitiveComboBox.cs
+++ b/GitUI/UserControls/CaseSensitiveComboBox.cs
@@ -1,0 +1,85 @@
+﻿using System;
+using System.ComponentModel;
+using System.Windows.Forms;
+
+namespace GitUI.UserControls
+{
+    public class CaseSensitiveComboBox : ComboBox
+    {
+        private const AutoCompleteMode Mode = AutoCompleteMode.SuggestAppend;
+        private string _lastTextChangedValue;
+
+        private bool SystemAutoCompleteEnabled
+        {
+            get => Mode != AutoCompleteMode.None && DropDownStyle != ComboBoxStyle.DropDownList;
+        }
+
+        protected override void OnValidating(CancelEventArgs e)
+        {
+            if (SystemAutoCompleteEnabled)
+            {
+                NotifyAutoComplete();
+            }
+        }
+
+        protected override void OnTextChanged(EventArgs e)
+        {
+            if (SystemAutoCompleteEnabled)
+            {
+                string text = Text;
+                if (text == _lastTextChangedValue)
+                {
+                    return;
+                }
+
+                _lastTextChangedValue = text;
+            }
+
+            base.OnTextChanged(e);
+        }
+
+        private int FindStringExactCase(string s)
+        {
+            for (int i = 0; i < Items.Count; i++)
+            {
+                if (GetItemText(Items[i]).Equals(s, StringComparison.Ordinal))
+                {
+                    return i;
+                }
+            }
+
+            return -1;
+        }
+
+        private void NotifyAutoComplete()
+        {
+            string text = Text;
+            bool textChanged = text != _lastTextChangedValue;
+            bool selectedIndexSet = false;
+            int index;
+            if (textChanged && text.Equals(_lastTextChangedValue, StringComparison.OrdinalIgnoreCase))
+            {
+                index = -1;
+            }
+            else
+            {
+                index = FindStringExactCase(text);
+            }
+
+            if (index != -1 && index != SelectedIndex)
+            {
+                SelectedIndex = index;
+                SelectionStart = 0;
+                SelectionLength = text.Length;
+                selectedIndexSet = true;
+            }
+
+            if (textChanged && !selectedIndexSet)
+            {
+                OnTextChanged(EventArgs.Empty);
+            }
+
+            _lastTextChangedValue = text;
+        }
+    }
+}
--- a/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
+++ b/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
@@ -329,13 +329,13 @@ namespace GitCommands.Remotes
                 }
 
                 remoteDisabled = remote.Disabled;
-                if (!string.Equals(remote.Name, remoteName, StringComparison.OrdinalIgnoreCase))
+                if (!string.Equals(remote.Name, remoteName, StringComparison.Ordinal))
                 {
                     // the name of the remote changed - perform rename
                     output = module.RenameRemote(remote.Name, remoteName);
                 }
 
-                if (!string.Equals(remote.Url, remoteUrl, StringComparison.OrdinalIgnoreCase))
+                if (!string.Equals(remote.Url, remoteUrl, StringComparison.Ordinal))
                 {
                     // the remote url changed - we may need to update remote
                     updateRemoteRequired = true;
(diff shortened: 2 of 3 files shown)


digitalcoyote:
Can someone review that build result? It seems like it should not be related my change.

> Command executed with exception: Cannot find path 'C:\projects\gitextensions\artifacts\tests\' because it does not exist.

pmiossec:
Why did you create your commits from a very old commits?

Try rebase on the updated upstream `master` branch to see if that solve the problem (because a lot has changed since, especially on how translation is handled).

PS: FYI, on GitHub, creating a PR from a `master` branch is a bad practice. I advise you to not do it again...

digitalcoyote:
Will do, I thought I pulled before starting. Thank You!

digitalcoyote:
I have not retested since I did a pull and rebase. I'll try to get that done ASAP

digitalcoyote:
Seems to still work as expected.

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -703,5 +705,84 @@ private void ShowSeparatePushUrl(bool visible)
                 ? _labelUrlAsFetch.Text
                 : _labelUrlAsFetchPush.Text;
         }
+
+        public class ComboBox : System.Windows.Forms.ComboBox
```
mstv, line 709: One class per file, please, i.e. move to `GitUI\UserControls`.

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -703,5 +705,84 @@ private void ShowSeparatePushUrl(bool visible)
...
+        {
+            private AutoCompleteMode _autoCompleteMode = AutoCompleteMode.SuggestAppend;
```
mstv, line 711: `const` or `readonly`?

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -703,5 +705,84 @@ private void ShowSeparatePushUrl(bool visible)
...
+                }
+
+                return -1;
+            }
+
+            private void NotifyAutoComplete()
+            {
+                var text = Text;
+                var flag1 = text != _lastTextChangedValue;
+                var flag2 = false;
```
mstv, line 760: Give them speaking names, please.

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -703,5 +705,84 @@ private void ShowSeparatePushUrl(bool visible)
...
+                int stringIgnoreCase;
```
mstv, line 761: Better name?

GitUI/CommandsDialogs/FormRemotes.cs
digitalcoyote: Renamed to reflect the equivalent naming in ComboBox reference source

GitUI/CommandsDialogs/FormRemotes.cs
digitalcoyote: Renamed to reflect the equivalent naming in ComboBox reference source

GitUI/CommandsDialogs/FormRemotes.cs
digitalcoyote: Made const.

GitUI/CommandsDialogs/FormRemotes.cs
digitalcoyote: Moved and renamed. It's no longer a subclass of FormRemotes. Not sure why I didn't do that to begin with since it's a general use Combobox replacement with case sensitive autocomplete.

digitalcoyote:
Rebased on the latest commit while I was at it. Thank you for the detailed review.

digitalcoyote:
> Since the sequence of the following commands seems to matter, add NUnit tests, please:
> 
> ```
>                     base.OnTextChanged(e);
>                     _lastTextChangedValue = text;
> ```

I don't believe the order should matter for these. `base.OnTextChanged(e);` alters `ComboBox._lastTextChangedValue` but this is a different variable than `CaseSensitiveComboBox._lastTextChangedValue` both are `private` fields

GitUI/UserControls/CaseSensitiveComboBox.cs
```diff
@@ -0,0 +1,89 @@
...
+                    return;
+                }
+
+                base.OnTextChanged(e);
+                _lastTextChangedValue = text;
+            }
+            else
+            {
+                base.OnTextChanged(e);
+            }
```
mstv, line 42: If the sequence does not matter, this can be simplified:
```suggestion
                _lastTextChangedValue = text;
            }

            base.OnTextChanged(e);
```

GitUI/UserControls/CaseSensitiveComboBox.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        private int FindStringExactCase(string s)
+        {
+            for (int i = 0; i < Items.Count; i++)
+            {
+                if (((Repository)Items[i]).Path.Equals(s, StringComparison.Ordinal))
```
mstv, line 49: This should be generalized. Perhaps it suffices to compare the item text.

GitUI/CommandsDialogs/FormRemotes.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
+using System.Security.Principal;
```
mstv, line 5: This can be removed, too.

RussKie:
I plan to review this as soon as I can get my head above water.
I'm surprised the ComboBox control needs to be subclassed to achieve the desired.

mstv:
> I'm surprised the ComboBox control needs to be subclassed to achieve the desired.

Perhaps it could be done using a decorator like `NativeTreeViewExplorerNavigationDecorator` or `NativeTreeViewDoubleClickDecorator`.

digitalcoyote:
> Perhaps it could be done using a decorator like NativeTreeViewExplorerNavigationDecorator or NativeTreeViewDoubleClickDecorator.

Unless I'm mistaken about how that works, I don't believe that is an option. I'm specifically avoiding some of ComboBox's private methods that cause problems. `NotifyAutoComplete()` ends up calling `FindStringIgnoreCase()` and then sets the selected index (which causes the case change to be reverted).

RussKie:
Thank you

PR #8041, opened 2020-04-26
Restore broken AutoCRLF tests

### LINKED ISSUE #8000: Review FileViewerTextTests.DoAutoCRLF tests

RussKie:
Resolves #8000

## Proposed changes

- Restore broken AutoCRLF tests

    The tests got broken for few reasons:

    1. Earlier versions of ApprovalTests considered differences in EOL, and these tests deal with precisely this.
    However recently we have updated to the newer version of ApprovalTests that no longer considered EOL differences as failures.

    2. During migration to Arcade SDK (#7649) few approved files had their EOL accidentally converted to CRLF, that was unnoticed due to the reasons explained above.

    CRLF conversion is moved out of `FileViewer` because it has no real connection to it. The tests refactored and verified for correctness.

- Move `FileViewer` members to give some logical order

- Remove unused  `FileViewer` members

Best review commit-by-commit.

❗ Do not squash!


RussKie:
OT: @gerhardol I noticed a peculiar behaviour that is quite confusing. I _think_ you worked in this area not that long ago.

Before committing files I have two different representations of files:
* added binary file (staged/unstaged):
image

* renamed/moved/deleted binary file:
image


When files are committed and I look at the diff all "binary" files are shown similar to this:
image


I can see how this can be confusing to users. Is this something we track somewhere, or I should raise an new issue for?

GitUI/Editor/FileViewer.cs
```diff
@@ -234,8 +182,6 @@ public new Font Font
             set => internalFileViewer.Font = value;
         }
 
-        public Action OpenWithDifftool => internalFileViewer.OpenWithDifftool;
```
gerhardol, line 237: Not opposing removal, just an observation: I kept intentionally after #7891 as the handling in FileViewerInternal should maybe be adapted instead, or removed too. With the follow up to #7899 and #7825 it makes sense to keep this in FileViewer instead of spreading in other forms.

RussKie: Can we add it when it is required? At the moment it just add noise.

RussKie:
> I would propose to _not_ change this behavior for diffs, it is an extra Git command that with viruses like Symantec could add 100 ms.
> Sure we could add an issue, but I do not think that will help anyone.

Certainly not changing here. Just an observation at this stage.

GitUI/Editor/FileViewer.cs
gerhardol: Of course, just noting the motivation to why not removing earlier and adding another note about the intenal opendifftool

PR #8043, opened 2020-04-27
Ignore GitHub API rate limit induced errors when search for app updates

### LINKED ISSUE #4469: Check for updates => NullReferenceException
bug

Click 'Help' -> 'Check for updates' => Error

image

- GitExtensions version: 2.51
- GIT version: 2.14.1.windows.1
- OS version: Windows 10 Enterprise, version 1703, OS Build 15063.877

### LINKED ISSUE #5928: NRE in Check for updates
**Bug**

Help > Check for updates
Does the check but also pops a NRE dialog that looks just like the one in #4469.

See above.

No crash dialog.  It's not fatal, but it greatly lessens my confidence in the no-updates-available response.

- Git Extensions 3.00.00.4433
- Git 2.20.0.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

2.51

### LINKED ISSUE #6070: Null reference exceptions
## Current behaviour

When I try and create a pull request I get the following exception.

## Expected behaviour

It creates a PR on github

## Steps to reproduce

Install 3.00.00.4433 on Windows 10 and try and create a PR

## Screenshots

image

image

Yes, all versions prior to v3

## Environment

- GitExtensions version:  version 3.00.00.4433
- GIT version: 2.20.0
- OS version: Windows 10
- .NET version: ?

### LINKED ISSUE #7413: NullReferenceException during update check
## Current behaviour
image

## Expected behaviour
:smiley: :tada: 

## Steps to reproduce
Help - Check for updates

## Screenshots
Look above.

Sorry. I installed it a few days ago. Nice app!!

## Environment
image

I'm behind a corporate proxy (maybe could be that?)

## Diagnostics

### LINKED ISSUE #8023: Unable to update through the network, release 3.3.1
## Current behaviour

The attempt to update through the Internet doesn't work for some reasons. Perhaps due to the network configuration, it's not clear. The internet connection itself is work. Hitting to the "Check for update" sub-menu the exception occurs, refer to the screenshot depicted below.

## Screenshots

image

## Environment

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version:

RussKie:
Resolves #4469
Resolves #5928
Resolves #6070
Resolves #7413
Resolves #8023

## Proposed changes

We had a number of NRE reports. Most likely scenario is that GitHub is API rate limiting unauthenticated requests  that lead to failures in Git.hub library.
Nothing we can do apart from ignoring these types of errors.

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormUpdates.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormUpdates.cs
@@ -96,6 +96,12 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                 // if a UI component gets disposed or the UI thread EXITs while a 'check for updates' thread
                 // is in the middle of its run... Ignore it, likely the user has closed the app
             }
+            catch (NullReferenceException)
+            {
+                // We had a number of NRE reports.
+                // Most likely scenario is that GitHub is API rate limiting unauthenticated requests that lead to failures in Git.hub library.
+                // Nothing we can do here, ignore it.
+            }
             catch (Exception ex)
             {
                 this.InvokeSync(() =>


GitUI/CommandsDialogs/BrowseDialog/FormUpdates.cs
```diff
@@ -96,6 +96,12 @@ private void SearchForUpdates()
                 // if a UI component gets disposed or the UI thread EXITs while a 'check for updates' thread
                 // is in the middle of its run... Ignore it, likely the user has closed the app
             }
+            catch (NullReferenceException)
+            {
+                // We had a number of NRE reports.
+                // Most likely scenario is that GitHub is API rate limiting unauthenticated requests  that lead to failures in Git.hub library.
```
gerhardol, line 102: nit: double spaces between requests and that

RussKie: Fixed, thank you

pmiossec:
I think we should update the setting storing the check date.

`AppSettings.LastUpdateCheck = DateTime.Now;`

That way we prevent the check to be done and fail at every startup of the application.
And a side effect is that we increase the probability to succeed the next time. Because I suspect that that fails in big companies where multiple employees are behind the same IP and if they all stop to call the api after a failure for some time, the probability that succeed the next time is higher....


>We had a number of NRE reports. Most likely scenario is that GitHub is API rate limiting unauthenticated requests that lead to failures in Git.hub library.
>Nothing we can do apart from ignoring these types of errors.

I was wondering...
What do you think about using the GitHub token if it is set?
That would permit some users (mostly the one that use GitHub in their company) to still have this feature working.
What do you think?

pmiossec:
>I think we should update the setting storing the check date.

Sorry. Forget this first remark. That's already set before opening the update form.

gerhardol:
> What do you think about using the GitHub token if it is set?

Yes

Issue #3923, opened 2017-08-22
Split up the pull/fetch/rebase/merge button into separate buttons


mikechamberlain:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

All fetching and merging from the toolbar has to be done through a fiddly split button that requires an unnecessarily high degree of dexterity for succeess.

![git-extensions-pull](https://user-images.githubusercontent.com/1414303/29545246-648796d2-8717-11e7-936e-9f1a5edbb8e1.png)

**What is the expected behavior?**

Remove the split button, together with the unintuitive and highly idiosyncratic "Set the next selection as the default" option. Introduce separate buttons for Merge, Rebase, Fetch, Fetch All and Open Pull Dialog.

**Motivation**

The vast majority of developers have high resolution monitors. We have plenty of empty space on the toolbar for more buttons. Pulling/merging/fetching is one of the most common operations. But we are currently forced to target a ~10 pixel-wide dropdown arrow to access the "non-default" actions.

This is made even more frustrating by the context menu's dismissal on mouse-out. It's too easy to slip off the edge and have to fiddle with the button again.

Lastly, the idea of setting the default action of a button from a menu on the button itself seems to be a completely random interaction, the likes of which I've only ever seen in GitExtensions.

All these issues are solved by breaking this weird and frustrating split button into regular top-level buttons.

NikolayXHD:
I suggest this feature to be implemented as optional.

When the dropdown is replaced by 4 (or 5) separate buttons, then whenever user fetches
he has to carefully pick one button from a group of 5 with similarly looking blue arrow icons.

I find the current look and behavior convenient, because I don't use fetch options others than the one selected as default. I also think this is true for many people, there is even a feature request #4159 which depends on current behavior.

RussKie:
I agree that the metaphor for selecting the default option is somewhat odd. This would be much better expressed with a Ribbon, if we had one.

However I also agree with @NikolayXHD that having several similarly looking buttons in the toolbar would have a high  degree of probability of creating confusion.
In my day-to-day job I see a lot with engineers who work on laptops, that do not provide have hi-res screens. Other engineers that have laptops with hi-res screens, often scale their Windows (150%, 200% etc).

Whilst I appreciate that the current UX is somewhat imperfect and it may cause some discomfort to some of our users, I am not entirely convinced the proposal would makes it universally better for everyone.

jbialobr:
[...]

Issue #5704, opened 2018-11-05
Hide submodules toolbar icon if no submodules exist


drewnoakes:
**Do you want to request a *feature* or report a *bug*?**

UX issue / bug

**What is the current behavior?**

For a repo with no submodules, the toolbar shows this:

![image](https://user-images.githubusercontent.com/350947/47991702-dd9bb500-e0e2-11e8-8478-1258a7bae64b.png)

Neither of these submenu items do anything useful. Both are confusing.

**What is the expected behavior?**

Currently this toolbar item has no value other than to cause confusion and clutter.

It should either be made useful, or removed. Submodules are a relatively "advanced" and rarely used git feature, and this button has prime position for something that's ultimately useless if you're not using submodules.

I favour removal&mdash;anything we can do to clean up the UI is a win.

If keeping it, the button should not have child items (no submodules) and should just open the "Manage Submodules" UI directly.

**Environment you encounter the issue:**

- GitExtensions version: 3.0 rc 1

RussKie:
Agree with the "remove" proposal.

gerhardol:
Suggest keeping, no big preference though

pmiossec:
I had the same idea some weeks ago, so I vote for removal, too...

PR #8047, opened 2020-04-28
Split toolbar pull button in 3 buttons

pmiossec:
* One for fetch operations,
* One for rebase operations,
* One for the more generic "pull dialog"

The first two with there own default operation setting.

Improvement over #3923

I use fetch and pull and I find it a shame to have only one default option and often be obliged to open the split menu....

**Note: I don't really know where to place the "pull dialog" button: before, in between or after the fetch and pull split buttons? Your point of view?**

### Before

image

### After

image

image

image

## Test methodology 

- Manual (used since some days)

## Test environment(s) 

- Git Extensions 3.3.0
- Build 6a07d061909403c233e87a633f0ccf3a1387e386 (Dirty)
- .NET Framework 4.8.4150.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -747,6 +747,15 @@ namespace GitCommands
         }
 
         /// <summary>
+        /// Gets or sets the default fetch action that is performed by the toolbar icon when it is clicked on.
+        /// </summary>
+        public static PullAction DefaultFetchAction
+        {
+            get => GetEnum("DefaultFetchAction", PullAction.Fetch);
+            set => SetEnum("DefaultFetchAction", value);
+        }
+
+        /// <summary>
         /// Gets or sets the default pull action that is performed by the toolbar icon when it is clicked on.
         /// </summary>
         public static PullAction DefaultPullAction
--- a/GitUI/CommandsDialogs/FormBrowse.Designer.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.Designer.cs
@@ -44,13 +44,15 @@ namespace GitUI.CommandsDialogs
             this.manageStashesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.createAStashToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStripButtonCommit = new System.Windows.Forms.ToolStripButton();
+            this.toolStripButtonFetch = new System.Windows.Forms.ToolStripSplitButton();
+            this.fetchToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.fetchAllToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.fetchPruneAllToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.setDefaultFetchButtonActionToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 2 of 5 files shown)


gerhardol:
I would prefer to not have more buttons in the toolbar. I do not see the need myself as I only use fetch. For me, this button is about all downloading. But not blocking.

Other:
 * Pull merge/rebase iconsdo not look too good, when zooming in (mostly that the semitransparent down button overlays the omega/m.
 * GE image viewer should probably scale up such images or implement zoom (original/full?).

RussKie:
I am not a fan of this proposal, sorry. Our toolbar is already crowded.
If we had a ribbon, then yes - we could have it like this.

pmiossec:
>I am not a fan of this proposal, sorry. Our toolbar is already crowded.

I don't buy this idea...
I use the "fetch" and "pull" button more often than nearly all the other buttons in the toolbar (perhaps except "commit" button).

We all have a slightly different use of GitExtensions but if I imagine that my use is not far than a lot of people. So here is my usage:

* refresh => **never** (F5 is quicker)
* toggle left panels => sometimes
* toggle split => **never**
* commit info => **never again** after first configuration
* submodules => very rarely (but depend of projects, so OK)
* repo => rarely (but OK to display repo path)
* branch => sometimes (but OK to display the branch)
* fetch => often
* rebase => often
* pull dialog => **never**
* push => often
* commit => often
* stash => sometimes
* open dir => sometimes
* bash => rarely
* settings => **very rarely**
* branches => **very rarely** (but OK because feature only there)
* filter => sometimes
* show first parents => **never**

Should I remove the useless (?) buttons to make some space to add these more often used buttons? ;)
I just think there is room for 1 or 2 more icons...

But the idea behind this PR is also to have one button for "fetch" and one for "pull" like nearly every GUI is doing it.

It seems to me to be 2 different concepts (and also 2 different git commands), and it makes more sense (and a better UX) to not aggregate them behind the same button/feature, no!?! I imagine to not be easy to understand for beginners...

mstv:
> But the idea behind this PR is also to have one button for "fetch" and one for "pull" like nearly every GUI is doing it.
It seems to me to be 2 different concepts (and also 2 different git commands)

So I could agree on two buttons, i.e. `Open pull dialog...` should be moved (back) to the drop down of the `Pull` SplitButton.

pmiossec:
>So I could agree on two buttons, i.e. Open pull dialog... should be moved (back) to the drop down of the Pull SplitButton.

👍 
That was initially what I wanted to do because it fits my use (I never use the `Open pull dialog...`).
But as the form can do pull and fetch (even if it is called "pull dialog"), I thought this configuration would have gather more consensus.

But put it in "Pull" split button makes sense also... I will do.

RussKie:
> We all have a slightly different use of GitExtensions but if I imagine that my use is not far than a lot of people. So here is my usage:

I am not happy to take changes that are user-specific or very tactical without general discussion and agreement on those. We have at least one more issue (https://github.com/gitextensions/gitextensions/issues/5704) that wants to alter the toolbar, yet it hasn't been actioned.

This said, I am interested to have a discussion about the toolbar in general. My personal use is somewhat different to yours, but there are some buttons I don't have any use for too, especially since we added the left panel. 
Other members of the team use the toolbar different as well, and our users may their own preferences too.

Please open a new issue where we can discuss our options. 
https://github.com/gitextensions/gitextensions/issues/1707 was an issue that entailed highly visible concerns (just like this one), and I like how it evolved. Perhaps we could have a discussion in a similar format.

mstv:
@pmiossec, you can add toolbar buttons using `Scripts` which are "_run_" on the event `ShowInUserMenuBar`.

Issue #8054, opened 2020-04-30
 Commit count do not match files in RevDiff WorkTree


gerhardol:
## Current behaviour

The commit count and commit dialogue (uses git-status) can report that other files are changed than in RevDiff WorkTree commit (uses git-diff), if file mode or whitespace differs.

## Expected behaviour

That commit count and FormCommit is the same as RevDiff WorkTree.

git-status can ignore file mode or whitespaces, as well as git-diff (--ignore-* options), but I do not see any arguments to get git-diff to use whitespace diff 
To ignore file mode for git-status would help, but it must be possible to commit these changes...

Note: 
GitExtensions.settings should be forced to use line-feed only in .gitattributes, as this is how GE stores the file. (Not depend on core.autocrlf). But this is local for GE repo, there should be a solution for all.

## Steps to reproduce

Toggle a change that affects GE in the current repo or change a file like
```$ unix2dos appveyor.yml
```

## Screenshots

![image](https://user-images.githubusercontent.com/6248932/80650700-c558ec00-8a74-11ea-8a5d-d25135612330.png)

![image](https://user-images.githubusercontent.com/6248932/80651058-72cbff80-8a75-11ea-94d2-0c2a90e1efea.png)

## Did this work in previous version of GitExtensions

No

## Environment

- Git Extensions 3.3.0
- Build 21d9669784aab05e491f7d3d2ee701c5f64e1734
- Git 2.26.2.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

## Diagnostics

```
$ git diff -z --name-only

$ git status --porcelain=2
1 .M N... 100644 100644 100644 02cf144834ab67a32152e8dd09b812937f246f9d 02cf144834ab67a32152e8dd09b812937f246f9d GitExtensions.settings
1 .M N... 100644 100644 100644 11ce49264d8631d5ea826514348007052e578ad0 11ce49264d8631d5ea826514348007052e578ad0 appveyor.yml
$

core.autocrlf=input
core.filemode=false as well as unset
```

PR #8059, opened 2020-05-02
Correct filemode for source files

gerhardol:
## Proposed changes

When playing around with activating core.filemode for #8054 
These files have the executable flag set for no reason which causes them to be marked as changed.

## Test methodology 

No change

----


gerhardol:
@msftbot merge in 1 day

PR #8067, opened 2020-05-05
Fix 8055 add sign-off to FormApplyPatch 

### LINKED ISSUE #8055: Apply Patch - add possibility to execute without --signoff option
## Feature description

It should be useful to apply patches without appending 'Signed-off by %USERNAME%' message to commits. Please see example of use case below.

Changes are implemented and committed in local repository that is located on Machine A. But Machine A does not have network access to remote repository, so changes are transferred somehow to Machine B (which has such network access) as patch files and then pushed to remote repository. In such a case several commits are applied simultaneously, and it is very annoying to remove 'Signed-off by %USERNAME%' line from each commit before push.

My proposal is:

- add new checkbox named 'Sign-off' to 'Apply patch' form, directly under 'Ignore wh.spc.' checkbox;
- new checkbox should be checked by default (this matches current behavior);
- if patch is applied successfully, then user choice (sign-off patch or not) should be saved and used to initialize checkbox state when 'Apply patch' form is opened next time.

I am new to GitHub and GitExtensions repo in particular. But, in my opinion, it should be not very difficult to implement proposed feature, so I would like to try it by myself and open a PR once it is done. Of course, if there are no objections from active part of community.

## Environment

- GIT version: 2.25.0.windows.1
- OS version: Microsoft Windows NT 10.0.18363.0
- .NET version: 4.8.4150.0

f-i-x-7:
FormApplyPatch - added new checkbox SignOff that controls whether 'git am' should be executed with '--signoff' option or not.

Fixes #8055 

## Proposed changes
- add new checkbox named 'Sign-off' to 'Apply patch' form, directly under 'Ignore wh.spc.' checkbox;
- new checkbox should be checked by default (this matches current behavior);
- user choice (sign-off patch or not) should be saved and used to initialize checkbox state when 'Apply patch' form is opened next time.

### Before

current

### After

proposed

## Test methodology 

- Updated unit tests for GitCommandHelpers class, checked that they are passing.
- Checked that new checkbox is checked by default (corresponds current behavior).
- Checked that state of new checkbox is remembered by GitExtensions.
- Checked that state of new checkbox defines whether '--signoff' option is added to executing 'git am' command or not.
- Checked tab indices on form.
- Checked that new checkbox is disabled when conflicts occur during applying of patch, and is enabled after patch skip/abort.
- Checked that other UI controls on form preserced their state.

## Test environment(s) 

- GIT 2.25.0.windows.1
- Windows NT 10.0.18363.0
- 144 dpi, 150% scaling

----


pmiossec:
You have to update the translation file with: https://github.com/gitextensions/gitextensions/wiki/Translations#v34-and-newer

f-i-x-7:
Some comments from my side:
1) There can be not only mailbox patches, but diff patches too. For them new checkbox has no sense ('git apply' does not have '--signoff' option). In my proposed implementation state of this checkbox is ignored for diff patches. Is it OK? Technically it is possible to check patch file contents after each filepath change, then detect whether it is diff patch or mailbox patch, and finally enable/disable checkbox. But I think it will not be very useful, so it will lead just to code complication.
2) In current version of GitExtensions, 'Apply' button has greater tab index than 'Ignore wh.spc.' checkbox, despise it is located above checkbox. I preserved such behavior ('Apply' button should be the last on its panel), so tab order in PR is 'Ignore wh.spc.' checkbox -> 'Sign-off' checkbox -> 'Apply' button.
3) Some code in form .Designer.cs file was updated automatically by VS designer (e.g., radiobuttons size, textboxes size, etc.) despite I opened Visual Studio with 100% scaling. It seems just that previous values in .Designer.cs file were incorrect (may be they were written manually), so VS Designer just updated then to actual values that were always used in runtime. Form looks good, so I believe these changes should not be reverted.

f-i-x-7:
> You have to update the translation file with: https://github.com/gitextensions/gitextensions/wiki/Translations#v34-and-newer

Thanks, I will take a look at it.

RussKie:
> * user choice (sign-off patch or not) should be saved and used to initialize checkbox state when 'Apply patch' form is opened next time.

Why? Is it a common use case to to load patches with the same set of settings?

RussKie:
> 1. There can be not only mailbox patches, but diff patches too... In my proposed implementation state of this checkbox is ignored for diff patches. Is it OK

I think it is ok




> 2\. In current version of GitExtensions, 'Apply' button has greater tab index than 'Ignore wh.spc.' checkbox, despise it is located above checkbox. I preserved such behavior ('Apply' button should be the last on its panel), so tab order in PR is 'Ignore wh.spc.' checkbox -> 'Sign-off' checkbox -> 'Apply' button.

It is ok (for now). At some point the form's layout will need to be looked at.

> 3\. Some code in form .Designer.cs file was updated automatically by VS designer

That's fine. The diff is still reasonably clear.

Thank you for raising these points 👍

f-i-x-7:
> Why? Is it a common use case to to load patches with the same set of settings?

Well, for my workflow it is true. Moreover, I would prefer to apply patches by default without signoff. I think it is something similar to commits behavior, they are performed without signoff by default, and it is OK for all.
In addition, state of 'Ignore wh.spc.' checkbox currently is saved to appsettings. So I had almost no doubts and implemented saving of new checkbox state too :)
Do you think such behavior will degrade user experience for most users?

f-i-x-7:
By the way. I deem that I should update documentation. I will take a look into existing documentation and open an additional PR in docs repo if update is required. I will perform that check after current PR will be agreed because in theory implementation can be changed after discussions here.

RussKie:
> I deem that I should update documentation. I will take a look into existing documentation and open an additional PR in docs repo if update is required.

The docs are here: https://github.com/gitextensions/GitExtensionsDoc
Thank you

RussKie:
Thank you

PR #8076, opened 2020-05-06
Run FormFileHistory in a separate GE instance

### LINKED ISSUE #7805: Commit dialog doesn't reload changes upon receiving focus
## Current behaviour

Split from #7622. While the _Refresh dialog on form focus_ works correctly when switching to the Commit dialog from a different application, it doesn't work when switching between GE windows. Since it is now possible (and useful) to create commits (especially fixups) from the File History dialog, the Commit dialog might, due to this bug, show an out-of-date state until manually refreshed.

## Expected behaviour

The Commit dialog should reload changes even when just switching to it from a different GE window, like File History.

## Steps to reproduce

(0. Make sure the _Refresh dialog on form focus_ option is on)
1. Make some changes in a repository
2. Open Commit dialog
3. On the first changed file, click View File History
4. A File History window opens
5. Pick an existing commit and press Ctrl+X
6. A new Commit dialog opens with the fixup message filled in
7. Add all the changes and commit the fixup
8. Close both the second Commit dialog and the File History dialog

The first Commit dialog is now focused again but it still shows the changed file even though it has already been committed.

I think this is due to the fact that the dialog contents is only reloaded in `OnApplicationActivated` which is in turn only called when the `WM_ACTIVATEAPP` message arrives - and that only occurs when switching from a _different_ application window.

I don't think so.

## Environment

- Git Extensions 3.3.0
- Build 0c3fe90bad6505b2d5c9f52d297364ffad3b14e4
- Git 2.25.1.windows.1
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

mstv:
Fixes #7805

Reviewed best commit by commit. Do not squash.

## Proposed changes

- Run `FormFileHistory` in a separate GE instance in order:
  - to be able to use the history after opening the modal `FormCommit`,
  - to refresh `FormCommit` when switching back from `FormFileHistory`,
    which could have been used to modify the repo.
- Resolve TODOs regarding non-zero `ExitCode` on error parsing command line arguments.
- Refactor in order to let `FormFileHistory` activate the tab "Blame" itself.

## Screenshots 

unchanged

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 09d0a839579126488423ce6a53c3236287a00615 (Dirty)
- Git 2.26.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1162,25 +1162,7 @@ namespace GitUI
         }
 
         public void StartFileHistoryDialog(IWin32Window owner, string fileName, GitRevision revision = null, bool filterByRevision = false, bool showBlame = false)
-        {
-            Form ProvideForm()
-            {
-                var form = new FormFileHistory(this, fileName, revision, filterByRevision);
-
-                if (showBlame)
-                {
-                    form.SelectBlameTab();
-                }
-                else
-                {
-                    form.SelectDiffTab();
-                }
-
-                return form;
-            }
-
-            ShowModelessForm(owner, true, null, null, ProvideForm);
-        }
+            => ShowModelessForm(owner, requiresValidWorkingDir: true, preEvent: null, postEvent: null, () => new FormFileHistory(this, fileName, revision, filterByRevision, showBlame));
 
         public void OpenWithDifftool(IWin32Window owner, IReadOnlyList<GitRevision> revisions, string fileName, string oldFileName, RevisionDiffKind diffKind, bool isTracked)
         {
@@ -1704,23 +1686,33 @@ namespace GitUI
             }
         }
 
-        private void RunFileHistoryCommand(IReadOnlyList<string> args)
+        /// <summary>
+        /// Remove working directory from filename and convert to POSIX path.
+        /// This is to prevent filenames that are too long while there is room left when the workingdir was not in the path.
+        /// </summary>
+        private string NormalizeFileName(string fileName)
         {
-            // Remove working directory from filename. This is to prevent filenames that are too
-            // long while there is room left when the workingdir was not in the path.
-            string fileHistoryFileName = string.IsNullOrEmpty(Module.WorkingDir) ? args[2] :
-                args[2].Replace(Module.WorkingDir, "").ToPosixPath();
-            var fullFilePath = _fullPathResolver.Resolve(fileHistoryFileName);
-            if (new FormFileHistoryController().TryGetExactPath(fullFilePath, out var exactFileName) &&
-                exactFileName.Length > Module.WorkingDir.Length)
+            fileName = fileName.ToPosixPath();
+            return string.IsNullOrEmpty(Module.WorkingDir) ? fileName : fileName.Replace(Module.WorkingDir.ToPosixPath(), "");
+        }
+
+        /// <returns>false on error</returns>
+        private bool RunFileHistoryCommand(IReadOnlyList<string> args)
+        {
+            string fileHistoryFileName = args[2];
+            if (new FormFileHistoryController().TryGetExactPath(_fullPathResolver.Resolve(fileHistoryFileName), out var exactFileName))
             {
-                fileHistoryFileName = exactFileName.Substring(Module.WorkingDir.Length);
+                fileHistoryFileName = NormalizeFileName(exactFileName);
             }
 
-            if (!string.IsNullOrWhiteSpace(fileHistoryFileName))
+            if (string.IsNullOrWhiteSpace(fileHistoryFileName))
             {
-                StartFileHistoryDialog(null, fileHistoryFileName);
+                return false;
             }
+
+            StartFileHistoryDialog(owner: null, fileHistoryFileName);
+
+            return true;
         }
 
         private void RunCloneCommand(IReadOnlyList<string> args)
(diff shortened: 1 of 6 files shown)


gerhardol:
Seem OK
Any suggestions for the manual testing?
Not often I ask for this, but we should maybe have testcases for the command line arguments?

mstv:
This was boring. Is there anybody who uses this multitude of command line arguments? :)

gerhardol:
> This was boring. Is there anybody who uses this multitude of command line arguments? :)
(I know that the shell extension needs several of them.)

I sympatize with you, but these are tests even I find useful. Looks good to me.

There are occasionally error reports on command line options, so there are users for some of the options. But most probably have very minor usage

RussKie:
> Is there anybody who uses this multitude of command line arguments? :)

I've been itching to refactor the command line parsing routines, it is a bowl of spaghetti... 
Until we look at this piece holistically it is hard to say. Being cli-friendly is a bonus

pmiossec:
>I've been itching to refactor the command line parsing routines, it is a bowl of spaghetti...

I like the [CommandLineParser lib](https://www.nuget.org/packages/CommandLineParser/) to work with verbs in command lines: https://github.com/commandlineparser/commandline/wiki/Verbs

I think that could help...

mstv:
Now, even AppVeyor lets pass the tests. This PR is ready (for 4.0).

mstv:
This one does not add more spaghetti. It solves minor issues and TODOs.
A rework of the cli parsing is a quite independent task and should be done in a follow-up PR. Though I think it will have low priority.

RussKie:
I suggest waiting until we move to .NET Core before attempting to rework
the cli handling.
Like Michael said, it is a quite separate task.

GitUI/GitUICommands.cs
```diff
@@ -1391,206 +1387,168 @@ public void StartCreatePullRequest(IWin32Window owner, IRepositoryHostPlugin git
...
-                case "difftool":      // filename
-                    Module.OpenWithDifftool(args[2]);
-                    return;
-                case "filehistory": // filename
+                    return Commit(arguments);
+                case "difftool":    // filename
+                    return Module.OpenWithDifftool(args[2]) == "";
+                case BlameHistoryCommand:
+                case FileHistoryCommand:
+                                    // filename [revision [--filter-by-revision]]
```
RussKie, line 1476: Bad indentation

GitUI/GitUICommands.cs
```diff
@@ -1675,25 +1637,27 @@ private void RunOpenRepoCommand(IReadOnlyList<string> args)
...
-        private void RunSynchronizeCommand(IReadOnlyDictionary<string, string> arguments)
+        private bool RunSynchronizeCommand(IReadOnlyDictionary<string, string> arguments)
         {
-            Commit(arguments);
-            Pull(arguments);
-            Push(arguments);
+            bool successful = true;
+            successful = Commit(arguments) && successful;
+            successful = Pull(arguments) && successful;
+            successful = Push(arguments) && successful;
```
RussKie, line 1648: I'm not sure this is correct. Suppose `Commit` will fail, so the `successful` on line:1646 will be `false`.
We will still execute `Pull`, which would be incorrect, wouldn't it?

Should it be this?
```suggestion
            bool successful = Commit(arguments) && Pull(arguments) && Push(arguments);
```

mstv: Stopping this sequence might make more sense.
Though I do not want to change the behavior in this commit, but only return the result.
Before, all three commands were run unconditionally one after the other:
```
            Commit(arguments);
            Pull(arguments);
            Push(arguments);
```

GitUI/GitUICommands.cs
mstv: It's not perfect, but it is aligned with the comments showing the argument lists of the other commands.

GitUI/GitUICommands.cs
RussKie: Should we fix it now?

gerhardol: Do it in a separate issue/PR
Changing this may break some specific usage and it is simple to separate it

RussKie:
👍

mstv:
@msftbot merge in 1 day

Plugins/GitUIPluginInterfaces/IGitUICommands.cs
```diff
@@ -30,9 +30,8 @@ public interface IGitUICommands
         bool StartCommandLineProcessDialog(IWin32Window owner, IGitCommand command);
         void StartBatchFileProcessDialog(string batchFile);
 
-        bool StartRemotesDialog();
```
thimmy687, line 33: Since StartRemotesDialog was removed, is it possible to add
`
 bool RunCommand(IReadOnlyList<string> args)
`
from GitUICommands or is there an other way to use this now?

My request comes from the gerrit plugin perspective

mstv: Sorry, @thimmy687, I was not aware that the interface `IGitUICommands.StartRemotesDialog()` was / is in use.
The `FormRemotes` cannot be opened by means of the CLI. So I think `RunCommand(IReadOnlyList<string> args)` would not work for your usecase.
Would it suffice to restore `IGitUICommands.StartRemotesDialog()`?

thimmy687: I have a workaround for it in this PR https://github.com/gitextensions/GitExtensions.GerritPlugin/pull/16 

So we could keep it how it is

PR #8084, opened 2020-05-08
Improve "Limit number of commits" option

### LINKED ISSUE #8083: Improve option: "Limit number of commits that will be loaded at startup"
## Current behaviour

GE needs to be restarted to apply the new value.
There is an hidden feature. If value is 0, it disables the limitation.

## Expected behaviour

It shouldn't be necessary to restart GE to apply the new value.
Disable limitation should not be an undocumented feature.

## Steps to reproduce

Open Settings.
Change "Limit number of commits that will be loaded at startup"
Close the Window
The revision grid is not updated.

Close the application
The revision grid is updated with the new limitation.

## Screenshots

image

Suggested new option

image

## Environment

- Git Extensions 3.3.1.7897
- Git 2.26.2.windows.1
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

bansan85:
Fixes #8083 

## Proposed changes

- It shouldn't be necessary to restart GE to apply the new value.
- Disable limitation should not be an undocumented feature.

### Before

image

### After

image

## Test methodology 

None

## Test environment(s) 

- Git Extensions 3.3.0
- Build d47d27eb1c7f0e566e48cce5694e6344d38dfbe4
- Git 2.26.2.windows.1
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/GeneralSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/GeneralSettingsPage.Designer.cs
@@ -42,7 +42,7 @@
             this.chkShowStashCountInBrowseWindow = new System.Windows.Forms.CheckBox();
             this.chkShowSubmoduleStatusInBrowse = new System.Windows.Forms.CheckBox();
             this.chkUseFastChecks = new System.Windows.Forms.CheckBox();
-            this.lblCommitsLimit = new System.Windows.Forms.Label();
+            this.lblCommitsLimit = new System.Windows.Forms.CheckBox();
             this._NO_TRANSLATE_MaxCommits = new System.Windows.Forms.NumericUpDown();
             this.groupBoxBehaviour = new System.Windows.Forms.GroupBox();
             this.tlpnlBehaviour = new System.Windows.Forms.TableLayoutPanel();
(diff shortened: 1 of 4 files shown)


mstv:
Is the limit applied now?

The label is too short now IMO, i.e. add " to be loaded" or the like.

I guess disabling will result in 🤔:

grafik

bansan85:
> Is the limit applied now?

It's the same behavior than before (except it's applied when the window is closed) : 

  - if 0 then no limitation,
  - if not 0 then limitation is applied,

It's just that if you uncheck the option, the limitation is forced to disable.

> 
> The label is too short now IMO, i.e. add " to be loaded" or the like.
> 

Are you sure ? I like this short text. If it not bothing you, I would like a second opinion.

> I guess disabling will result in 🤔:
> 
> grafik

No : unchecking the checkbox disables the NumeripUpDown but it does not force the value in NumericUpDown to 0. It's in case someone unchecks by mistake, it's easy reversable. But if you uncheck the box and close the window, the value is definitly 0.

pmiossec:
> > The label is too short now IMO, i.e. add " to be loaded" or the like.
> 
> Are you sure ? I like this short text. If it not bothing you, I would like a second opinion.

When this PR was discussed, I was pretty sure we were talking of the option placed in the `View` => `Set advanced filter` which is really similar.
In that case, either of the label was good for me.

But I just realized that it is another option that was discussed about here. And I must admit that I agree with @mstv and because we don't have a context here, the label must be very explicit for the user and the one chosen here is not clear enough.

Sorry to be late in the discussion...

bansan85:
Sorry, I forget about this PR :/

I changed the text of the option to "Limit number of commits to be loaded".

PR #8086, opened 2020-05-09
RevDiff: FirstRevision is null for the initial commit

### LINKED ISSUE #8085: Right clicking an item in RevisionDiffControl results in null exception for the initial commit
## Current behaviour
Null exception trying to access FirstRevision when right clicking a file in RevisionDiffControl in the initial commit of a repo. 

## Expected behaviour
No null exception when opening the right click menu.

## Steps to reproduce
Select the initial commit of a repo and right click a file in the RevisionDiffControl.

yes

## Environment
- Git Extensions 3.4.0.9381
- Git 2.26.0.rc2.windows.1
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8085 

## Proposed changes

- Refactor: Align FirstParentId name
Guid suffix is quite consistently used for the string representation of the hash, while Id suffix is used for ObjectId, this was therefore switched
Some parentGuids changed too, not changing treeGuid

- Use "--root" for first revision if null
Handle GitRevision == null as the root commit, before the initial commit, i.e. compare to /dev/null
SemanticMerge and KDiff3 shows such a diff, Tortoise does not for some reason
(The alternative would be to not add null to the revisions to diff which will result in a argument popup which is OK to me.)

- RevDiff: FirstRevision is null for the initial commit
Handle that FirstRevision is not set for the initial commit
All situations changed are not causing problems, but it is easier to handle FirstRevision consistently this way.

Should be merged with these three merges, separate PR for 3.4 when approved

## Test methodology

Review if FirstRevision usage and Manual
The (meaningless) IRevisionDiffController do not cover arguments in RevisionDiffControl

----

### DIFF
--- a/GitUI/CommandsDialogs/FormDiff.cs
+++ b/GitUI/CommandsDialogs/FormDiff.cs
@@ -8,6 +8,7 @@ using GitCommands.Git;
 using GitExtUtils.GitUI.Theming;
 using GitUI.HelperDialogs;
 using GitUI.Theming;
+using GitUI.UserControls;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
 using ResourceManager;
@@ -285,11 +286,7 @@ namespace GitUI.CommandsDialogs
 
         private ContextMenuDiffToolInfo GetContextMenuDiffToolInfo()
         {
-            var parentIds = DiffFiles.SelectedItems
-                .Where(i => i.FirstRevision != null)
-                .Select(i => i.FirstRevision.ObjectId)
-                .Distinct()
-                .ToList();
+            var parentIds = DiffFiles.SelectedItems.FirstIds().ToList();
             bool firstIsParent = _revisionTester.AllFirstAreParentsToSelected(parentIds, _headRevision);
             bool localExists = _revisionTester.AnyLocalFileExists(DiffFiles.SelectedItems.Select(i => i.Item));
 
--- a/GitUI/CommandsDialogs/FormDiff.cs
+++ b/GitUI/CommandsDialogs/FormDiff.cs
@@ -152,6 +152,11 @@ namespace GitUI.CommandsDialogs
 
             foreach (var item in DiffFiles.SelectedItems)
             {
+                if (item.FirstRevision?.ObjectId == ObjectId.CombinedDiffId)
+                {
+                    continue;
+                }
+
                 var revs = new[] { item.SecondRevision, item.FirstRevision };
                 UICommands.OpenWithDifftool(this, revs, item.Item.Name, item.Item.OldName, diffKind, item.Item.IsTracked);
             }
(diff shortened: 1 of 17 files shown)


fsfod:
I guess what happens for "Open with Diff tool" menu entries if the selected file's FirstRevision is null are going to be weird or broken. The only one enabled is selected -> first but p4 merge doesn't like "nul" as the second file name idk if thats bug or not. Trying "cherry picked changes" for the same file does nothing idk if it should or not.

gerhardol:
> I guess what happens for "Open with Diff tool" menu entries if the selected file's FirstRevision is null are going to be weird or broken. 

This PR handles firstRev==null as "--root", so in supported difftools is works OK
Diff should be handled as for new files

> The only one enabled is selected -> first but p4 merge doesn't like "nul" as the second file name idk if thats bug or not. Trying "cherry picked changes" for the same file does nothing idk if it should or not.

A specific problem in p4merge, similar to tortoisemerge but you get a popup
semanticmerge and kdiff3 displays the "diff". Blocking diff for new files would remove this functionality for these tools, just to not show the popup for p4merge

RussKie:
@fsfod please confirm the fix https://ci.appveyor.com/api/buildjobs/opo96182ksn1sqr2/artifacts/artifacts%2Fpublish%2FGitExtensions-Portable-3.4.0.9417-9d7acd0f9.zip

fsfod:
Yes the original issue is fixed for me.

GitUI/CommandsDialogs/FormDiff.cs
```diff
@@ -280,7 +285,11 @@ private void diffContextToolStripMenuItem_Opening(object sender, System.Componen
 
         private ContextMenuDiffToolInfo GetContextMenuDiffToolInfo()
         {
-            var parentIds = DiffFiles.SelectedItems.Select(i => i.FirstRevision.ObjectId).Distinct().ToList();
+            var parentIds = DiffFiles.SelectedItems
+                .Where(i => i.FirstRevision != null)
+                .Select(i => i.FirstRevision.ObjectId)
+                .Distinct()
```
RussKie, line 291: We're executing this selector multiple times. 
Have you considered creating a property on `FileStatusList` (similar to `SelectedGitItems`) to simply the call sites?

E.g. something like:
```cs
    public sealed partial class FileStatusList : GitModuleControl
    {
        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
        [Browsable(false)]
        public IEnumerable<FileStatusItem> SelectedGitObjects
        {
            get => SelectedItems
                .Where(i => i.FirstRevision != null)
                .Select(i => i.FirstRevision.ObjectId)
                .Distinct();
        }
```

gerhardol: It was a little like this before, with many access functions. These properties applies to Enumerate<FileItemStatus> for the selected items. Similar could apply to all items etc,
But I do not know how to do this in the best way, I assumed that a static helper on FileStatusItem would not be accepted, so I let that be.
Similar for First, Second, GitStatus.
(I was not ready thinking about this very late with #7899 and did not want to do any more changes...)

SelectedGitItems should be removed too, but it appeared in a little too many places.

gerhardol:
> Tests for RevisionDiffInfoProvider changes?

Added tests as well as reordered arguments

GitUI/CommandsDialogs/FormDiff.cs
gerhardol: I added extension methods, OK now?

RussKie:
Taking into 3.4

Issue #7978, opened 2020-04-11
Present Git errors in FormCommit


gerhardol:
## Current behaviour

Follow up to #7915, that was started from #7902 

Git errors are presented as files with #7915. The solution is a hack that improves the previous functionality but it has some limitations:
 * All code paths may possibly not be covered, the error message may not be presented if not 'presenting changes'. I believe this is handled now (git-status and git-diff is not really used for other presentations). This type of error presentation could be used in other situations though, then this may be an issue
 * FormCommit is not handled.
  - The fake file is added without stagedStatus, so the errors are filtered out. This requires either that StagedStatus is set for the files (faked statuses should be avoided though) or special handling in FormCommit (3 situations files are retrieved)
  - Presentation in FileViewer.cs ViewCurrentChanges() is not presenting the error. This is handled in #7825 (would be a simple change if the first item is fixed)

## Expected behaviour

Errors are presented regardless of form

## Steps to reproduce

See #7902 for an example that creates a parse error

## Did this work in previous version of GitExtensions

No

## Environment
- Git Extensions 3.3.0
- Build 0112ae532b96a8231167b3b3f62af958f671c43d
- Git 2.26.0.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

mstv:
Could the git error message be shown, e.g. appended to the dummy file name?

I hope this will also avoid the display of a `Git Error` on renamed files.
It occurs with a combination of `bugfix/i7902-diff-fatal-parse` with `feature/i4154-revdiff-base-a` (https://github.com/mstv/gitextensions/commit/e09a9d6040b6c18b7d04905cd699df17b14d3923).

#### STR:
- rename a file
- stage all
-> `Git Error`
- hit `F5`
-> correct

[F12.txt](https://github.com/gitextensions/gitextensions/files/4466293/F12.txt)

gerhardol:
> Could the git error message be shown, e.g. appended to the dummy file name?

The error item is added either as a normal GitItemStatus when parsing or as the Git error message.
In an extreme  object oriented approach, this Git Error object would be derived from a base class only having Name and ErrorMessage fields.

Appending to the name was the first approach, a little hard to see.

> I hope this will also avoid the display of a Git Error on renamed files.

That is an error corrected in the PR, see e5bc512efa3a566af13d58ea4bfe5a0f32b40993

Issue #1326, opened 2012-10-17
Diff: optionally show file list as tree


feinstaub:
For large file lists in a diff view it could be helpful to be able to show the file list as hierarchical tree (as in the File Tree Tab)

RussKie:
> Anyhow, I am making progress :)

@mdonatas anything to share yet?

vbjay:
I think grouping by folder path is better. I'd much rather keep the
listview.

SignumPL:
Any progress on this?

gerhardol:
If someone starts working on this feature, they should add a note about it. It is up for grabs.

RussKie:
For the general benefit, for the most part agreed proposal is [here](https://github.com/gitextensions/gitextensions/issues/7867#issuecomment-602094623):

![image](https://user-images.githubusercontent.com/4403806/77305732-adda6500-6d4a-11ea-906c-17e4a5e36e0b.png)

PR #8088, opened 2020-05-10
Fix 8077

### LINKED ISSUE #8077: v3.4 - Diff window no longer compares the head of 2 selected branches
Up to version v3.3.1, when I selected 2 commits (branches), I got an actual diff of those 2 commits (cumulative), but in 3.4, this no longer works.

3.3.1:

image

3.4:

image

RussKie:
Fixes #8077

## Proposed changes

- Add "no items" to empty groups
- Collapse empty diff groups

## Screenshots 

image

New screenshots:
* image
* image

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1114,13 +1114,11 @@ namespace GitUI.CommandsDialogs
 
             foreach (var fileStatus in allChangedFiles)
             {
-                if (!string.IsNullOrWhiteSpace(fileStatus.ErrorMessage))
+                if (fileStatus.IsNonFile)
                 {
                     unstagedFiles.Add(fileStatus);
-                    continue;
                 }
-
-                if (fileStatus.Staged == StagedStatus.Index)
+                else if (fileStatus.Staged == StagedStatus.Index)
                 {
                     stagedFiles.Add(fileStatus);
                 }
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1114,17 +1114,14 @@ namespace GitUI.CommandsDialogs
 
             foreach (var fileStatus in allChangedFiles)
             {
-                if (fileStatus.IsStatusOnly)
+                if (fileStatus.Staged == StagedStatus.WorkTree || fileStatus.IsStatusOnly)
                 {
+                    // Present status only errors in unstaged
                     unstagedFiles.Add(fileStatus);
                 }
                 else if (fileStatus.Staged == StagedStatus.Index)
                 {
                     stagedFiles.Add(fileStatus);
-                }
-                else if (fileStatus.Staged == StagedStatus.WorkTree)
-                {
-                    unstagedFiles.Add(fileStatus);
                 }
             }
 
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1114,7 +1114,7 @@ namespace GitUI.CommandsDialogs
 
             foreach (var fileStatus in allChangedFiles)
             {
-                if (fileStatus.IsNonFile)
+                if (fileStatus.IsStatusOnly)
                 {
                     unstagedFiles.Add(fileStatus);
                 }
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -48,7 +48,7 @@ namespace GitUI
             [NotNull] string defaultText = "",
             [CanBeNull] Action openWithDiffTool = null)
         {
-            if (!string.IsNullOrWhiteSpace(item?.Item?.ErrorMessage))
+            if (item?.Item?.IsNonFile ?? false)
             {
                 // Present error (e.g. parsing Git)
                 return fileViewer.ViewTextAsync(item.Item.Name, item.Item.ErrorMessage);
@@ -58,7 +58,7 @@ namespace GitUI
             {
                 if (!string.IsNullOrWhiteSpace(defaultText))
                 {
-                    return fileViewer.ViewTextAsync(item?.Item?.Name, defaultText);
+                    return fileViewer.ViewTextAsync(item.Item.Name, defaultText);
                 }
 
                 fileViewer.Clear();
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -47,7 +47,7 @@ namespace GitUI
             [NotNull] string defaultText = "",
             [CanBeNull] Action openWithDiffTool = null)
         {
-            if (item?.Item?.IsNonFile ?? false)
+            if (item?.Item?.IsStatusOnly ?? false)
             {
                 // Present error (e.g. parsing Git)
                 return fileViewer.ViewTextAsync(item.Item.Name, item.Item.ErrorMessage);
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Drawing;
-using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading;
@@ -58,7 +57,7 @@ namespace GitUI
             {
                 if (!string.IsNullOrWhiteSpace(defaultText))
                 {
-                    return fileViewer.ViewTextAsync(item.Item.Name, defaultText);
+                    return fileViewer.ViewTextAsync(item?.Item?.Name, defaultText);
                 }
 
                 fileViewer.Clear();
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -28,6 +28,7 @@ namespace GitUI
 
     public sealed partial class FileStatusList : GitModuleControl
     {
+        private const string ShowAllDiferencesItemName = "ShowDiffForAllParentsText";
         private static readonly TimeSpan SelectedIndexChangeThrottleDuration = TimeSpan.FromMilliseconds(50);
         private readonly TranslationString _diffWithParent = new TranslationString("Diff with a/");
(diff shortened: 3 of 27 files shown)


RussKie:
The code is "Sunday midnight" quality (i.e. far from perfect). Right now more to gather feedback and form the direction.

RussKie:
/cc: @laurin1 your feedback is welcome too

laurin1:
> /cc: @laurin1 your feedback is welcome too

Not sure if you wanted the reply here or in #8077. I prefer just turn all of that off (now that I know how) and as long as disabling is still an option, I don't have a strong opinion about how that info is handled with the secondary diff's enabled. 

However, I do want to thank you for your quick and thorough response and work being done on this issue - I was a little frustrated when I wrote my initial comments and your professionalism is much appreciated.

RussKie:
> Suggest "No items" is removed, otherwise what me and @mstv discussed - I could not implement the no-collapsing.

Apologies, I don't follow.
It is impossible to show empty groups, so a group has to have at least one entry. Hence we need "no items" (or whatever other wording we choose).

> To add such an item affects a lot of parts in the UI (context menus etc). #7978 added such a hack for error messages, but this is a more common case.

Again, apologies, I can't see a direct connection of this issue to #7978. 
If you refer to a question of adding a dummy item to show otherwise hidden from use information - then yes, I think we should do so (where logical).

gerhardol:
> It is impossible to show empty groups, so a group has to have at least one entry. Hence we need "no items" (or whatever other wording we choose).

OK, then there is no other way to show the group than adding a dummy

> Again, apologies, I can't see a direct connection of this issue to #7978.
If you refer to a question of adding a dummy item to show otherwise hidden from use information - then yes, I think we should do so (where logical).

#7978 added the "error" item to the view, this is also a dummy GitStatusItem

But then this item need special handling.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -970,7 +973,29 @@ private void UpdateFileStatusListView(bool updateCausedByFilter = false)
...
+                if (isFirstGroup && i.statuses.Count < 1)
+                {
+                    itemStatuses = noItemStatuses;
+                }
+                else
+                {
+                    itemStatuses = i.statuses;
+                }
+
+                if (GitItemStatusesWithDescription.Count > 1)
```
mstv, line 985: Why not `> 0`?
```suggestion
                if (GitItemStatusesWithDescription.Count > 0)
```

gerhardol: (Not confirmed this, but the ListGroup is Collapsible by default so this is not needed. But the code would be simpler if the compare and set state is removed.)

RussKie:
> Why are we trying to tweak the `ListView`? Why not switch to a `TreeView` [as discussed](https://github.com/gitextensions/gitextensions/issues/1326#issuecomment-602500732)? After 3.4?

This issue is for v3.4. I too find the current behaviour highly confusing. 

> For 3.4, collapsing without the dummy item "no items" might suffice

A group isn't shown _unless_ it has items, so it is imperative to have a dummy item.

GitUI/UserControls/FileStatusList.cs
RussKie: Yes, can be 0. Thank you

gerhardol:
>> For 3.4, collapsing without the dummy item "no items" might suffice

> A group isn't shown unless it has items, so it is imperative to have a dummy item.

Yes, but you still see no expanded groups, so there is a visual cue: No diff shown

https://github.com/gitextensions/gitextensions/issues/8077#issuecomment-626961077
> (Collapsing the secondary diffs by default is a good compromise IMO and will give a clear indication. I must use it a while, until I can say whether I personally like the collapsed secondary diffs or whether it annoys me to need to double click them.)

To expand secondary groups by default is a little annoying to me, the secondary groups are as important as the first when comparing branches.
Saved somehow by that empty groups are not shown

RussKie:
Ready for reviews

gerhardol:
This solution is mostly OK in normal use, also for myself
A limitation is that the guid is hardcoded and contents of a file "No Changes" would be shown
I propose a different approach, see https://github.com/gerhardol/gitextensions/pull/new/bugfix/_fix_8077

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1282,6 +1326,25 @@ private void FileStatusListView_ContextMenu_Opening(object sender, CancelEventAr
                 cm.Items.Add(new ToolStripSeparator());
                 cm.Items.Add(_sortByContextMenu);
             }
+
+            if (!cm.Items.Find(ShowAllDiferencesItemName, true).Any())
```
gerhardol, line 1330: This menu appears also in other forms, can be confusing there
The tooltip is good here, but I am leaning on removing the option anyway

RussKie: A little more information please

RussKie:
> propose a different approach, see https://github.com/gerhardol/gitextensions/pull/new/bugfix/_fix_8077

I like that. feel free to push in this branch

gerhardol:
> feel free to push in this branch

I rebased your changes on #8114 (#8101 for 3.4, we need that too, definitely for testing this) and squashed with a new commit message.
Added two fixup commits:
 * Use GitItemStatus IsNonFile to handle the special files. Somethings I am not too happy about:
   - IsNonFile should maybe be renamed. IsStatusOnly or IsNonGit is maybe better but seem strange in GitItemStatus...
   - ErrorMessage could be used as before, handling non null as non GitStatusItem. No Changes would then be an empty message, most changes in this commit reverted 
* Only show the toggle for changing the settings if relevant. I probably want to remove this still

Of course feel free to reverse.

Will probably approve this after a little thinking. This is better than the current functionality.

GitUI/GitUIExtensions.cs
```diff
@@ -58,7 +58,7 @@ public static class GitUIExtensions
             {
                 if (!string.IsNullOrWhiteSpace(defaultText))
                 {
-                    return fileViewer.ViewTextAsync(item?.Item?.Name, defaultText);
+                    return fileViewer.ViewTextAsync(item.Item.Name, defaultText);
```
RussKie, line 61: @gerhardol this may cause NRE ❗

RussKie:
> * IsNonFile should maybe be renamed. IsStatusOnly or IsNonGit is maybe better but seem strange in GitItemStatus...

I'm happy with `IsStatusOnly`. `IsNonGitStatus`?

GitUI/GitUIExtensions.cs
gerhardol: Checked on line 57, not needed

gerhardol:
> I'm happy with IsStatusOnly

Unless someone objects, I will change to IsStatusOnly

GitUI/GitUIExtensions.cs
RussKie: Here's the logic lines:57-61:

```
If `item == null OR item.Item == null OR item.SecondRevision == null OR item.SecondRevision.ObjectId == null` Then
  If `!string.IsNullOrWhiteSpace(defaultText)` Then
    return fileViewer.ViewTextAsync(item.Item.Name, defaultText);
```

So `item.Item.Name` can be `item == null OR item.Item == null` and thus result in NRE.

RussKie:
> Unless someone objects, I will change to IsStatusOnly

👍 
Btw I have rebased, so fetch the latest.

RussKie:
I did a little tweak:
image

mstv:
The tooltip comes and goes at the bottom of the screen (I am not requesting a change):

de9b1a46-843d-4ff1-945f-cc457b8af438

RussKie:
> The tooltip comes and goes at the bottom of the screen (I am not requesting a change):

That's likely a Windows Forms / Win32 behaviour

RussKie:
Tests need updating

gerhardol:
Tests are fixed in #8119 (as well as a minor problem in addition to the core feature).
(that also rebases this PR on master)

I prefer to not backport the fix and suggest we either include #8119 in 3.4 or wait with the correction.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -976,17 +999,29 @@ private void UpdateFileStatusListView(bool updateCausedByFilter = false)
                     FileStatusListView.Groups.Add(group);
                 }
 
-                foreach (var item in i.statuses)
+                if (hasChanges && i.statuses.Count < 1)
```
mstv, line 1002: Move the variable to the inner block and improve the readability of the condition:
```suggestion
                IReadOnlyList<GitItemStatus> itemStatuses;
                if (hasChanges && i.statuses.Count == 0)
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -91,6 +99,15 @@ public FileStatusList()
 
             _fullPathResolver = new FullPathResolver(() => Module.WorkingDir);
             _revisionTester = new GitRevisionTester(_fullPathResolver);
+            _noItemStatuses = new[]
+            {
+                new GitItemStatus
+                {
+                    Name = $"     - {NoFiles.Text} -",
+                    IsStatusOnly = true,
+                    ErrorMessage = ""
```
mstv, line 108: `ErrorMessage` should be assigned either `null`, i.e. not assigned, or `string.Empty`, shouldn't it?

gerhardol: Empty string has a special meaning: No icon in the status list.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1114,13 +1114,11 @@ private void LoadUnstagedOutput(IReadOnlyList<GitItemStatus> allChangedFiles)
 
             foreach (var fileStatus in allChangedFiles)
             {
-                if (!string.IsNullOrWhiteSpace(fileStatus.ErrorMessage))
+                if (fileStatus.IsStatusOnly)
```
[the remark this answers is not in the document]
gerhardol: A motivation to have these separate is that this is really two different pieces of information, but I am adding a comment too

PR #8111, opened 2020-05-15
Fix/script without icon

mstv:
## Proposed changes

- It is allowed to not assign an icon to a user script.
  So avoid regarding exceptions.
- Display the shortcuts assigned to the user scripts.
- Fixup the condition for showing the user scripts arguments cheat sheet again after closing it.

### Before

NBug exceptions unless icons are defined

grafik

### After

grafik
grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build 3486ba04ff582343f82a8d2e947328ed09ba5434
- Git 2.26.0.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8118, opened 2020-05-17
Fix up user scripts without icon

mstv:
Clone of #8111 for the `release/3.4` branch

----


RussKie:
Let's squash it.

PR #8116, opened 2020-05-16
Bugfix/i8115 submodule gitstatus

gerhardol:
Fixes #8115

## Proposed changes

First commit:
- SubmoduleStatus: Check that a path exist before updating
- Set status as non dirty if git-status has no status also for other than top module
- Improve comments about the handling

The first commit prepares for two improvements to limit the commands run by the git status provider.
No major improvement, they could be separated to a separate commit too
- SubmoduleStatus: Use all info about submodules info from git-status
- SubmoduleStatus: Prune recursive calls when module is not dirty

## Test methodology 

Manual
There are some tests for SubmoduleStatusProvider.cs that are revised in #8049 
They do not have complex enough test cases to test this

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8049, opened 2020-04-29
Fix 7097 crash invalid escape character in .gitmodule

RussKie:
Fixes #7097

## Proposed changes

Catch exceptions related to invalid configs and present them to user in a better way, giving the user an opportunity to correct it rather than crashing the app.

### Before

image

### After

image

## Test methodology 

- manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #7861, opened 2020-03-10
Commit message cleared out when fixing up commit to another commit which starts from "#"


Mikilll94:
## Current behaviour

The commit message is cleared out when doing "fixup" for a commit which is getting merged into another commit whose commit message starts with "#".

This bug is shown on the GIF below:

![git_extensions_rebase_fixup_hash](https://user-images.githubusercontent.com/17653648/76362626-9f8b5280-6321-11ea-86c1-2774b2e2efa0.gif)

## Expected behaviour
As you can see I want to fixup "Commit three" into "Commit two" but this operation results in the empty commit message for "Commit two".

This is because "Commit two" message starts with a "#".

## Steps to reproduce
1. Start the interactive rebase.
2. Fixup a commit to another commit which starts from "#".
3. Finish rebase.
4. As you can see, now there is an empty commit message.

## Environment

- Git Extensions 3.3.1.7897
- Build 5a97671645532bcedc443bac7b727f40db47cb5c
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18363.0
- .NET Framework 4.8.4121.0
- DPI 120dpi (125% scaling)

gerhardol:
Duplicate of #7859 

fixup is using interactive rebase
This is handled by Git, not GE

pmiossec:
That's how git works and not a Gitextensions problem. Git use the # as a line comment.

Your workaround : https://www.kevinkuszyk.com/2016/03/09/git-tips-3-changing-the-git-rebase-comment-character/

RussKie:
@Mikilll94 what result do you get if you do it via command line?

gerhardol:
> what result do you get if you do it via command line?

Must be the same, as this is how Git is working
The change in GE would be to implement the rebase with own code

Closing as this is a duplicate

PR #8119, opened 2020-05-17
Submodule status: Throttle instead of drop frequent updates

gerhardol:
I believe this is ready, but did not plan to include it in 3.4
Submitting now for discussion of #8049 

## Proposed changes

SubmoduleStatusProvider will not update the status more often than every 15th second to limit load.
This was added before the status update was fully async and before updates were pruned (latest change in #8116).
The status should not be noticable (at least comparable to git-status, that can take a long time in big repos with many submodules).

The update limitation remains with this PR, but instead of dropping frequent updates, the handling is throttled to update no more often than every 15th second.
The handling has also been simplified.

A change to specific mention is that previously, if not the top module was selected, the complete structure including the current submodule was updated. The current submodule was then updated a second time. (There was a comment that the update not triggered by git-status would skip this, but that was incorrect.)
This will maybe be noticeable on a deep tree where a module below submodule is updated. (but very minor effect)

## Test methodology 

Tests exists, a test is disabled for the same reason as before.
Can maybe be improved.

----

### DIFF
--- a/GitCommands/Submodules/SubmoduleStatusProvider.cs
+++ b/GitCommands/Submodules/SubmoduleStatusProvider.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using GitCommands.Git;
@@ -38,11 +37,14 @@ namespace GitCommands.Submodules
 
     public sealed class SubmoduleStatusProvider : ISubmoduleStatusProvider
     {
+        // Throttle updates triggered from status updates
+        private const int MinRefreshInterval = 15;
+
+        private readonly CancellationTokenSequence _submodulesStructureSequence = new CancellationTokenSequence();
         private readonly CancellationTokenSequence _submodulesStatusSequence = new CancellationTokenSequence();
         private DateTime _previousSubmoduleUpdateTime;
         private SubmoduleInfoResult _submoduleInfoResult;
         private readonly Dictionary<string, SubmoduleInfo> _submoduleInfos = new Dictionary<string, SubmoduleInfo>();
-        private IReadOnlyList<GitItemStatus> _gitStatusWhileUpdatingStructure;
 
         // Singleton accessor
         public static SubmoduleStatusProvider Default { get; } = new SubmoduleStatusProvider();
@@ -55,12 +57,14 @@ namespace GitCommands.Submodules
 
         public void Dispose()
         {
+            _submodulesStructureSequence.Dispose();
             _submodulesStatusSequence.Dispose();
         }
 
         public void Init()
         {
             // Cancel any previous async activities:
+            _submodulesStructureSequence.Next();
             _submodulesStatusSequence.Next();
         }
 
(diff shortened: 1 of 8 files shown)


gerhardol:
Squashed and rebased https://github.com/RussKie/gitextensions/tree/fix_8077 on master, rebased this on that commit and fixed the commented out tests, added a few lines more
This was a problem with the datastructures

The bug in the application is not too severe. The status for submodules if not the top project was checked out was incorrect. So the problem requires at least three levels of submodules.

The drop down looked like this:
image

With the correction, it looks like the following (sidepanel is correct)
image

gerhardol:
The current implementation can be further simplified, no need to separate topmodule and submodule structure update

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -136,24 +117,29 @@ public async Task UpdateSubmodulesStructureAsync(string workingDirectory, string
...
+                await Task.Delay(TimeSpan.FromMilliseconds(300), cancelToken);
+                cancelToken.ThrowIfCancellationRequested();
             }
 
-            var cancelToken = _submodulesStatusSequence.Next();
-            _gitStatusWhileUpdatingStructure = null;
-
-            await TaskScheduler.Default;
+            // Throttle updates triggered from status updates
+            TimeSpan remaining = _previousSubmoduleUpdateTime - DateTime.Now.AddSeconds(-15);
```
RussKie, line 137: Suggest defining a const and use it here. Makes it easier to track.

gerhardol:
I submitted this to support #8049 in a safe way, but it has maybe grown too big for 3.4
I would like to add some more test cases for changing revisions as well (could be done later too).
I may submit an alternative with a few line changes for 3.4 instead.

gerhardol:
Testcases added, also checking AllSubmodules/OurSubmodules (what failed in the old testcases)
I added 300 lines of testcases, now many more but not all usages are covered. So it is still not a safenet. (Also before reviews, I am blind for issues myself now.)
Added some tests for the actual working, even if it is incomplete. For instance, if not a top module is checked out, then git-status run on the submodule and can determine that the top module is dirty. However, clearing the dirty file cannot determine that the top module is no longer dirty.

mstv:
> ChangedFiles does not report the expected number of changes in Appveyor

Without having looked at the code, it might help to add
`await AsyncTestHelper.JoinPendingOperationsAsync(AsyncTestHelper.UnexpectedTimeout);`
and `forceUpdate: true`.

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -136,24 +117,30 @@ public async Task UpdateSubmodulesStructureAsync(string workingDirectory, string
...
+                await Task.Delay(TimeSpan.FromMilliseconds(300), cancelToken);
+                cancelToken.ThrowIfCancellationRequested();
             }
 
-            var cancelToken = _submodulesStatusSequence.Next();
-            _gitStatusWhileUpdatingStructure = null;
-
-            await TaskScheduler.Default;
+            // Throttle updates triggered from status updates
+            const int minRefreshInterval = 15;
```
mstv, line 137: I guess, Igor meant to create a class constant, which is visible at the top of the file.

gerhardol:
Split tests and added a few more. Six tests fails on AppVeyor though. It is consistent behavior though, no evident race condition.
I check what Git options Appveyor used, nothing strange (and GE should override user settings anyway if needed.)
Will squash this (it is test cases that has changed since the review).

@mstv any more hints to why the tests fail?
Anything specific to be done so the testcases are not cleaned up and access with RDP?

mstv:
> Six tests fails on AppVeyor though.

I have not found a regarding failed AppVeyor run.

> Will squash.

OK for me.

mstv:
Printf debugging still is not the worst idea.
You could attach the debugger to the test runner.
(IIRC, the AppVeyor VM has only one cpu core with hyperthreading.)

gerhardol:
Eventually fixed the problem, same problem as #7861: AppVeyor does not set an email address which causes commit to fail.
The testcases sets email addresses in the local config but when adding a module as a submodule, the local config is not copied.
I intend to squash the two first commits.

gerhardol:
One more corner case, with another test case

gerhardol:
@msftbot merge in 24 hours

PR #8133, opened 2020-05-22
fix: New remotes not saved, if exceed limit

RussKie:
## Proposed changes

If a user adds a new remote which index, when saved, exceeds the number of maximum recent remotes - the remote would be added to the end of the list of remotes, and trimmed off when the list of recent remotes is saved.

Fix the issue by adding the new remote to the top of the list, so it gets saved.

## Test methodology 

- manual
- unit tests

----

### DIFF
--- a/GitUI/CommandsDialogs/FormRemotes.cs
+++ b/GitUI/CommandsDialogs/FormRemotes.cs
@@ -17,6 +17,7 @@ namespace GitUI.CommandsDialogs
 {
     public partial class FormRemotes : GitModuleForm
     {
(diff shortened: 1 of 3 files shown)


RussKie:
> What is the limit for remotes, where is it?

`RemoteRepositoryManager` uses `AppSettings.RecentRepositoriesHistorySize`, which is here:
image

It is certainly debatable whether this setting should be reused for remotes, but it is what it is.

I checked local repo persistence works correctly.

RussKie:
This was causing me issues at work, I have heaps of remotes to deal with.

PR #7926, opened 2020-03-31
Use ViewChangesAsync consistently for changes

gerhardol:
Part of #7338 
Related to #7825 
Fixes #7866 

## Proposed changes

There are many codepaths that "presents changes" to files (binary, images) and submodules. The presentation differs if the file is new, tracked etc.
The logic for that is split between the separate forms and GitUIExtensions.cs, ViewChangesAsync(). Some of that has been cleaned up in previous PRs.
(FormCommit.cs also needs to have some logic in FileViewer.cs ViewCurrentChanges(), that requires #7825 first.)
The presentation therefore differs from form to form.

This change unifies the handling to be consistent, all using ViewChangesAsync()

This PR shares (conflicts) a change in #7914, rev-parse is required to get guid, see the explanation there.

This only changes how FormStash presents changes, the only actual change of this PR is that images that are changed should now be displayed.
(and FormCommit should not see any changes when this is applied in or after #78825, depending what goes in first).

## Test methodology
Manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8137, opened 2020-05-23
Fix `FormCommit` context menu issues

### LINKED ISSUE #8106: FormCommit context menu order changed
## Current behaviour

The new behaviour (v3.4):
image

image

image

RussKie:
## Proposed changes

- Move "Open with diff tool" context menu out from the top position in `FormCommit`.

   In the File Tree and Diff tabs a user would typically deal with existing commits (i.e. snapshots), and "compare" function is likely to be one of the most used (i.e. compare to the working directory).
In the Commit form the user is dealing with the working directory changes, and thus staging and reseting is likely be the primary function, then compare (which is rendered in the diff panel of the form).

   Resolves #8106

- Correct display for 'Show file differences for all parents' menu.
The context menu was shown for the first time because of the incorrect check - it assumed the item was found, whilst it wasn't.
Set visibility of the added context menu at the time of creation.

    Addendum to #8077.

### Before

image
image

### After

image
image

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -185,11 +185,11 @@ namespace GitUI.CommandsDialogs
             // UnstagedFileContext
             //
             this.UnstagedFileContext.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
-            this.openWithDifftoolToolStripMenuItem,
             this.resetChanges,
             this.resetPartOfFileToolStripMenuItem,
             this.interactiveAddToolStripMenuItem,
             this.toolStripSeparator12,
+            this.openWithDifftoolToolStripMenuItem,
             this.editFileToolStripMenuItem,
             this.openToolStripMenuItem,
             this.openWithToolStripMenuItem,
@@ -378,9 +378,9 @@ namespace GitUI.CommandsDialogs
             // StagedFileContext
             //
             this.StagedFileContext.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
-            this.stagedOpenDifftoolToolStripMenuItem9,
             this.stagedResetChanges,
             this.stagedToolStripSeparator14,
+            this.stagedOpenDifftoolToolStripMenuItem9,
             this.stagedEditFileToolStripMenuItem11,
             this.stagedOpenToolStripMenuItem7,
             this.stagedOpenWithToolStripMenuItem8,
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -228,7 +228,18 @@ namespace GitUI
             get { return FileStatusListView.ContextMenuStrip; }
             set
             {
+                if (FileStatusListView.ContextMenuStrip == value)
+                {
+                    return;
+                }
+
+                if (FileStatusListView.ContextMenuStrip != null)
+                {
+                    FileStatusListView.ContextMenuStrip.Opening -= FileStatusListView_ContextMenu_Opening;
+                }
+
                 FileStatusListView.ContextMenuStrip = value;
+
                 if (FileStatusListView.ContextMenuStrip != null)
                 {
                     FileStatusListView.ContextMenuStrip.Opening += FileStatusListView_ContextMenu_Opening;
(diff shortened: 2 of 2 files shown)


RussKie:
1. Not having "Reset" as a top item is a deal breaker for me - it is how we had it for at least 7 years (I checked back to v2.47), and it since it moved I'm consistently struggling with it.

2. The default items in `FormCommit` should be `Stage` and `Unstage` for the list of unstages/staged items respectively. This is what is happening when a user double clicks on an item.
I've never really paid any attention, but we don't have these entries in the context menu. I believe they should be added, but since it is not a regression I wouldn't consider this is a blocker for the v3.4 release.


> FormCommit - I still vote no, but will not block. I consider moving difftool to the second group consistently a less bad change

This PR is for FormCommit.
I'm happy to work through other context menus as a separate item.

gerhardol:
> 1. Not having "Reset" as a top item is a deal breaker for me - it is how we had it for at least 7 years (I checked back to v2.47), and since it moved I'm consistently struggling to invoke the correct operation.

> This PR is for FormCommit.
I'm happy to work through other context menus as a separate item.

If this is a problem, we should never change any order in any form.

Not having consistency is a deal breaker for me. I had to search the menu for most actions before as order differed from one form to another. I see no special reason to change the order for only FormCommit. 
I could live with a consistent change, even if I would expect more users to be annoyed if difftool is moved. Any change will annoy some existing users, consistency is something that is not pushing new users away.

If changed, move Open with difftool to top of next group, move down Edit/save below Open (still same group). But do it for all. Also "context menu order" methods should change, affects separators too.

> The default items in FormCommit should be Stage and Unstage for the list of unstages/staged items respectively. This is what is happening when a user double clicks on an item.
I've never really paid any attention, but we don't have these entries in the context menu. I believe they should be added, but since it is not a regression I wouldn't consider this is a blocker for the v3.4 release.

The absence has crossed my mind too. But some other issues are before in my mental queue, like https://github.com/gitextensions/gitextensions/pull/7961#pullrequestreview-391277863 and #7926 (that also makes the FileViewer context menu consistent).

RussKie:
> If this is a problem, we should never change any order in any form.
> 
> Not having consistency is a deal breaker for me. I had to search the menu for most actions before as order differed from one form to another. I see no special reason to change the order for only FormCommit.

I like how you use my mantra and spell of "consistency" against me 😆 👍 

I too like consistency, and it can be at several levels - one of which is being consistent in serving habitual preferences of our users. We need to be cognizant of the fact that users may have mechanical memory of certain actions (e.g. me in resetting items in FormCommit, or resetting the current branch in the revision graph context menu - another pain btw).

Whilst I totally support making the app look and behave consistent, one thing I disagree is a desire to make _all_ context menus consistent across the whole app. Different forms/dialogs have different primary functions, and context menus can look different to accommodate these functions. Consistent presentation (i.e. text and images) are significantly more important.

FormCommit's context menus must have "Stage"/"Unstage" entries at the very top respectively, and I will have to unlearn my current ways, but I'd see it as logical placement. Having "Open with difftool" as the first item in the FormCommit is illogical to me because the form provides this functionality natively. It may not accommodate all use cases, but what difftool does?
I don't want to add new context menus because it would require new translations, and I don't want to have another round of those.

gerhardol:
> I like how you use my mantra and spell of "consistency" against me

Dont worry, regardless if I want it or not, you will get a chance to get a revenge. For this setup my preference is consistency though.

> Whilst I totally support making the app look and behave consistent, one thing I disagree is a desire to make all context menus consistent across the whole app.

The menus are quite different (some changes for unmotivated changes needed for names etc needed). It is just that I find this change unmotivated.
I could accept change in other forms too (but I do want to set DiffTool as default in RevDiff, so there it fits on top).

>FormCommit's context menus must have "Stage"/"Unstage" entries at the very top respectively, and I will have to unlearn my current ways, but I'd see it as logical placement. Having "Open with difftool" as the first item in the FormCommit is illogical to me because the form provides this functionality natively. It may not accommodate all use cases, but what difftool does?
I don't want to add new context menus because it would require new translations, and I don't want to have another round of those.

The translation strings already exists (a hack to use the text may be needed to trick TranslationApp).
@RussKie Will adding such items be a way forward?

RussKie:
> The translation strings already exists (a hack to use the text may be needed to trick TranslationApp).

I'm not 100% positive, but I think Transifex auto-translates strings that it already has in its (project's) database. So if we have strings already (would be strange if we didn't) - I think it should work.



> Will adding such items be a way forward?

I would think so. Why?

> I do want to set DiffTool as default in RevDiff, so there it fits on top).

In the main form at the TabControl level - viewing a diff main could be viewed as one of the main operations, so placement the menu higher feels logical to me. In FormCommit - it doesn't fit this pattern.

gerhardol:
Suggestion on way forward:
 * drop 4d5a5b60f62d5aa39fbfe4f9397012c77aa1bfbc from this PR and merge it
 * merge #8170 
 * I (unless someone is faster, if I get back from work as usual it will likely take another 12h) create a new PR that orders all menus as https://github.com/gitextensions/gitextensions/issues/8106#issuecomment-633160937  - should be easy to review

gerhardol:
Submitted #8175 for my suggested changes, also bold default and a few minor tweaks

mstv:
> Whilst I totally support making the app look and behave consistent, one thing I disagree is a desire to make all context menus consistent across the whole app. Different forms/dialogs have different primary functions, and context menus can look different to accommodate these functions. Consistent presentation (i.e. text and images) are significantly more important.

Agree.

> Having "Open with difftool" as the first item in the FormCommit is illogical to me because the form provides this functionality natively. It may not accommodate all use cases, but what difftool does?

I do not see the difference between FormCommit and RevDiff. Both show a patch / unified diff (with more or less lucky matching of lines). I often find it easier to compare files side by side. In addition, my favourite diff tools allow to re-align moved lines as well as to quickly edit in-place. That's why I open the difftool from both windows.

RussKie:
Closing in favour of https://github.com/gitextensions/gitextensions/pull/8175

PR #8035, opened 2020-04-23
Indicate merged branches in left panel

mstv:
Fixes #6494

## Proposed changes

- [x] icons for branch nodes
- [x] git command
- [x] set icons in dependency on `RevisionGridControl.LastSelected`

### Before

grafik

### After

grafik

## Test methodology 

- add NUnit test
- manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build d6e12450f823906b6ae4ed527dda296cc45721f9
- Git 2.26.0.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7959, opened 2020-04-07
Fix rename file icon size

pmiossec:
by removing the DPI metadata embedded in the `FileStatusRenamed.png` file

& and do it for all the other files while at it...

using tool: https://github.com/jsummers/tweakpng

Same problem than #7942

### Before

image

### After

image

## Test methodology 

- Manual 

## Test environment(s) 

- Git Extensions 3.3.0
- Build 2636852dde06ac9a76d09244206126019044b2ce (Dirty)
- Git 2.25.0.windows.1 (recommended: 2.25.1 or later)
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4150.0
- DPI 192dpi (200% scaling)

PR #7942, opened 2020-04-03
LeftPanel: Fix size of the GitHub remote icon

pmiossec:
by removing the PPI metadata embedded in the `github.png` file

& add the little tool allowed that (in case ?):
	http://entropymine.com/jason/tweakpng/
	https://github.com/jsummers/tweakpng

Fixes #7940

### Before

image

### After

image

## Test methodology 

- With my hawk eyes 🤣 

## Test environment(s) 

- Git Extensions 3.3.0
- Build 3813d3b3ce70d1961e9f5099e0a17880a500d240
- Git 2.25.0.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4150.0
- DPI 192dpi (200% scaling)

PR #8139, opened 2020-05-23
Indicate merged branches in left panel

### LINKED ISSUE #6494: Left panel: indicate whether branch has been merged
## Feature description

left panel

E.g. a different branch icon could be displayed if the branch has been merged.
It could be flipped vertical and should have a different color.
GH uses the icon merged. Its style does not match GE's and I do not like the color very much.

Thoughts?

## Environment

- GitExtensions version: 3.x

mstv:
Clone of #8035 for release/3.4: Fixes #6494

## Proposed changes

- Set branch icons in dependency on `RevisionGridControl.LastSelected`

### Before

grafik

### After

grafik

## Test methodology 

- add NUnit test
- manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build e36f71482703ef8b48f91b69ca3df0226cbaefeb
- Git 2.26.0.windows.1
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

----


mstv:
CodeFactor seems to be down today.

pmiossec:
Bug:

image

Dpi information must be removed from the image file. See #7942 and #7959 for way to fix it... 

@gerhardol I'm pretty sure I should have not removed the exe like you told me to do it because it is useful for others (and really small)

pmiossec:
Another thing. I find disappointing that the icon of the selected branch change (even if technically it make sense because that's "contained" in the selected commit).

I would have made a tooltip to explain the meaning of the icon.
I knew the existence of this feature so I though about the real meaning of this icon but I imagine other users.......
.

mstv:
> Dpi information must be removed from the image file.

Will do.

> I knew the existence of this feature so I though about the real meaning of this icon but I imagine other users.

I agree, it is not fully self explainig. A tooltip could explain it. I would add it for "merged" only.

> I find disappointing that the icon of the selected branch change.

Please elaborate "disappointing".
We can suppress the "merged" status for the selected commit if others vote for it, too.

gerhardol:
> I'm pretty sure I should have not removed the exe like you told me to do it because it is useful for others (and really small)

If we had used git-lfs, I would have been OK with including it. Without, Git should not handle binary files, it adds quickly to the repo size (even if the file was small).
For binaries where the files are used in the build (including all images), there is not much to do.

But maybe we should start using git-lfs, it has been included in Git for windows for several years.

I like the icon and prefer to keep it, but see the point for a tooltip.

(@pmiossec Sorry for voting no to this and several others of your ideas lately, it is likely a coincidence as most of your features are very useful. I saw your GH branch with downvoted features too...)

mstv:
The current branch is contained in its head commit of course. Hence the "merged" icon is displayed for the current branch if the head commit is selected.
IIUIC, @pmiossec wants the normal icon to be shown in this case anyway.
Should we suppress the "merged" state for the branch of the **currently selected** commit? Please vote!

grafik

gerhardol:
Low prio for me, but I vote suppress.
The merged status is usable to see branches that can be deleted, so therefore keep the odd markers on them.

Opposite reasoning would be that you are primary looking at non-merged branches, then 'merged' should include current

pmiossec:
>Please elaborate "disappointing".

Sorry to not have been enough clear but you well understood my concern. :+1: The feature was "branch contained in current selected commit" and not "branch included in the selected branch" so this limit case was difficult to understand when, as @gerhardol state it well, "The merged status is usable to see branches that can be deleted"... 


@gerhardol 

>@pmiossec Sorry for voting no to this and several others of your ideas lately, 

Oh, you did that! I didn't noticed ;) I have to find some interesting in retaliation now that I know that 😈 

>I saw your GH branch with downvoted features too.

See no evil in this branch name. I tried to find an easy to memorize branch name to indeed keep some features that was rejected and be able to create a custom version. But also to test some very small features for some times and so gather feedback if it works well...

PR #8124, opened 2020-05-19
Restore splitter positions

mstv:
Fixes #8072

## Proposed changes

Workaround for not reliably restored splitter positions:
Restore the layout of `FormBrowse` a second time after `base.OnLoad(e)`.

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 3.3.0
- Build e972d9a0a1c35a31bf0495a1536547ababc872b5
- Git 2.26.0.windows.1
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8155, opened 2020-05-26
Fix up splitter positions

### LINKED ISSUE #8153: Tab settings are not restored at startup
## Current behaviour

When starting up to dashboard, Commit info is always a tab, not restored to the upper right position where I prefer to have it. An empty panel is there instead.

Also, the diff slider may not be restored correctly, but it may be a consequence of panel not restored.
Window size is restored though.

## Expected behaviour

Panel positions and sliders are retained when starting up.

## Steps to reproduce

Start to 
Portable flag or not is the same for me.
I set the commit info to the upper right, seem to be the same also when keeping as a tab: The upper right panel is empty.

## Screenshots

image

New in latest weeks
Maybe #8124?

## Environment

- Git Extensions 3.4.0.9558
- Build b613ba26577cd6262e108ae991a6ec2f9ae24fab (Dirty)
- Git 2.26.2.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

mstv:
Fixes #8153

## Proposed changes

- Restore the call sequence of before #8124
- But keep a second call of `SetSplitterPositions` after `base.OnLoad`

## Test methodology 

- manual (with an without start to the Dashboard)

## Test environment(s) 

- Git Extensions 3.3.0
- Build bec3e4e0e35853b95d307432a975ba860d82bb2d
- Git 2.26.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -532,7 +532,10 @@ namespace GitUI.CommandsDialogs
                     new WindowsThumbnailToolbarButton(toolStripButtonCommit.Text, toolStripButtonCommit.Image, CommitToolStripMenuItemClick),
                     new WindowsThumbnailToolbarButton(toolStripButtonPush.Text, toolStripButtonPush.Image, PushToolStripMenuItemClick),
                     new WindowsThumbnailToolbarButton(toolStripButtonPull.Text, toolStripButtonPull.Image, PullToolStripMenuItemClick)));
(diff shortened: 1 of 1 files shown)


mstv:
If starting to the Dashboard, the panels are restored but not the splitter positions (as before #8124).
If someone has a solution for this, add it please. I will not search for it.

gerhardol:
Seem to work for me

PR #8123, opened 2020-05-19
Fix 8122 Update Atlassian.SDK package to latest version

haldyr:
Fixes #8122 
Fixes #8089

## Proposed changes

- Updated Atlassian.SDK package to latest version

### Before
- some Jira issues could not be deserialized

### After
- updated Atlassian.SDK dependency fixes problem

## Test methodology 

- tested manualy in commit window (before change it didn't work, after change it is OK)

## Screenshots

Before
image

After
image

## Test environment(s) 
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.18363.0

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #7747, opened 2020-02-09
Update vs-threading analyzers and fix new diagnostics

sharwell:

PR #8156, opened 2020-05-27
Update packages.

NN---:
Update packages to newer versions which has .NET Standard build, so it can be consumed by .NET Core.
Fix using obsolete methods.

**I grant this project or any other developer the permissions to use unconditionally the provided code without any limitation and to use it for any purpose.**

### DIFF
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -110,7 +110,9 @@ namespace GitUI
                 if (file.IsSubmodule && file.GetSubmoduleStatusAsync() != null)
                 {
                     // Patch already evaluated
+#pragma warning disable VSTHRD103 // Call async methods when in an async method (this is intentional for the test)
                     var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync);
+#pragma warning restore VSTHRD103 // Call async methods when in an async method
                     return status != null
                         ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                         : $"Failed to get status for submodule \"{file.Name}\"";
--- a/Solution.Versions.props
+++ b/Solution.Versions.props
@@ -1,31 +1,31 @@
 <Project>
   <PropertyGroup>
     <!-- Solution dependencies -->
-    <AdysTechCredentialManagerVersion>1.7.0</AdysTechCredentialManagerVersion>
-    <AppInsightsWindowsDesktopVersion>2.10.42-preview</AppInsightsWindowsDesktopVersion>
-    <BenDemystifierVersion>0.1.4</BenDemystifierVersion>
-    <JetBrainsAnnotationsVersion>2018.2.1</JetBrainsAnnotationsVersion>
-    <LibGit2SharpVersion>0.25.0</LibGit2SharpVersion>
-    <MicrosoftVisualStudioCompositionVersion>15.6.36</MicrosoftVisualStudioCompositionVersion>
-    <MicrosoftVisualStudioThreadingVersion>16.5.132</MicrosoftVisualStudioThreadingVersion>
-    <NewtonsoftJsonVersion>10.0.3</NewtonsoftJsonVersion>
+    <AdysTechCredentialManagerVersion>2.1.1</AdysTechCredentialManagerVersion>
+    <AppInsightsWindowsDesktopVersion>2.13.1</AppInsightsWindowsDesktopVersion>
+    <BenDemystifierVersion>0.1.6</BenDemystifierVersion>
+    <JetBrainsAnnotationsVersion>2020.1.0</JetBrainsAnnotationsVersion>
+    <LibGit2SharpVersion>0.26.2</LibGit2SharpVersion>
+    <MicrosoftVisualStudioCompositionVersion>16.4.11</MicrosoftVisualStudioCompositionVersion>
+    <MicrosoftVisualStudioThreadingVersion>16.6.13</MicrosoftVisualStudioThreadingVersion>
+    <NewtonsoftJsonVersion>12.0.3</NewtonsoftJsonVersion>
     <RestSharpVersion>106.11.4</RestSharpVersion>
-    <SmartFormatNETVersion>2.0.0</SmartFormatNETVersion>
-    <StyleCopAnalyzersVersion>1.2.0-beta.113</StyleCopAnalyzersVersion>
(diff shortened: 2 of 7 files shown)


gerhardol:
For next release, I assume.
Will have to be rebased on master, for instance #8123 updates RestSharp including some code changes.
The submodules need to be updated too. (Versions must match at least, I assume that there are dependencies too.)

NN---:
Yeah, submodules needs to be updated as well.

NN---:
Btw is there a reason we use submodules instead of putting in this repo ?
It makes PRs much easier to maintain and to propose.

gerhardol:
> Btw is there a reason we use submodules instead of putting in this repo ?

The submodules are cloned from other origins, by keeping the structure it is easier to take updates and push changes upstream. Not much used for the GE submodules though.
Doc is of course special, EasyHook is a submodule so we can drop that code when no longer needed (after some .NET and Winforms updates in the future).
The editor is a separate .exe too. NBug was actually moved into GE when most functionality was removed.

Subtrees could be used instead, no major change.

Submodules is the standard way in Git to share codem this will not likely change.

RussKie:
Thank you. I'll get to it once we push v3.4 out of the door.

RussKie:
@NN--- can you please rebase and resolve the MC

NN---:
Ok

NN---:
This change required Externals/Git.hub fix

RussKie:
> This change required Externals/Git.hub fix

Please send a PR to it

NN---:
Added PR

pmiossec:
@NN--- I fixed the unit tests not passing. Rebase and squash when you review my commit.

NN---:
@pmiossec GitHub has squash merge which does essentially the same, no ?

pmiossec:
@NN--- Yes but that let the responsibility to the maintainer to squash and choose the best commit message where the contributor is sometimes the best placed to do it.

That's why if the PR could be in a finished state, where the maintainer just has to merge or rebase and merge, it's easier for everyone...

GitExtensions.ruleset
```diff
@@ -48,9 +48,11 @@
     <Rule Id="SA1403" Action="None" />
     <Rule Id="SA1404" Action="None" />
     <Rule Id="SA1413" Action="None" />
+    <Rule Id="SA1503" Action="None" />
     <Rule Id="SA1504" Action="None" />
     <Rule Id="SA1512" Action="None" />
     <Rule Id="SA1516" Action="None" />
+    <Rule Id="SA1519" Action="None" />
```
RussKie, line 55: Why are disabling these rules?

GitUI/GitUIExtensions.cs
```diff
@@ -110,7 +110,9 @@ void OpenWithDiffTool()
                 if (file.IsSubmodule && file.GetSubmoduleStatusAsync() != null)
                 {
                     // Patch already evaluated
+#pragma warning disable VSTHRD103
                     var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync);
```
RussKie, line 114: Can you please provide more details on this supression?

NN---: Same way as suppression here:
https://github.com/gitextensions/gitextensions/pull/7747/files#diff-fa0461ff6916a00b8ca6742fc94d5bd5R182

GitExtensions.ruleset
NN---: Seems like not needed.
Removing

GitExtensions.ruleset
```diff
@@ -48,6 +48,7 @@
     <Rule Id="SA1403" Action="None" />
     <Rule Id="SA1404" Action="None" />
     <Rule Id="SA1413" Action="None" />
+    <Rule Id="SA1503" Action="None" />
```
RussKie, line 51: What about this one?

GitExtensions.ruleset
NN---: Fixed by adding braces as expected by the rule: https://github.com/gitextensions/gitextensions/pull/8156/files#diff-0c448da7cf150a550546e895757d16c2R182

GitUI/GitUIExtensions.cs
RussKie: Please add comments explaining the supressions, like in the other commit.

NN---: Done

GitExtensions.ruleset
NN---: Sorry, forgot to remove.

RussKie:
Can you please provide sign-off in contributors.txt? Without it things get complicated as far as licenses concerned

Solution.Versions.props
```diff
@@ -1,31 +1,31 @@
...
-    <BenDemystifierVersion>0.1.4</BenDemystifierVersion>
-    <JetBrainsAnnotationsVersion>2018.2.1</JetBrainsAnnotationsVersion>
-    <LibGit2SharpVersion>0.25.0</LibGit2SharpVersion>
-    <MicrosoftVisualStudioCompositionVersion>15.6.36</MicrosoftVisualStudioCompositionVersion>
-    <MicrosoftVisualStudioThreadingVersion>16.5.132</MicrosoftVisualStudioThreadingVersion>
-    <NewtonsoftJsonVersion>10.0.3</NewtonsoftJsonVersion>
+    <AdysTechCredentialManagerVersion>2.1.1</AdysTechCredentialManagerVersion>
+    <AppInsightsWindowsDesktopVersion>2.13.1</AppInsightsWindowsDesktopVersion>
+    <BenDemystifierVersion>0.1.6</BenDemystifierVersion>
+    <JetBrainsAnnotationsVersion>2020.1.0 </JetBrainsAnnotationsVersion>
```
RussKie, line 7: Just noticed stray spaces in few places throughout the file
```suggestion
    <JetBrainsAnnotationsVersion>2020.1.0</JetBrainsAnnotationsVersion>
```

NN---: 👍

mstv:
@NN--- has granted the free use of these changes. So one who signed the CLA, could adopt the changes and create an identical PR.
We won't do this often but it seems to be worth in this case. Shall we proceed this way?

RussKie:
I've taken @NN--- changes in the .NET Core branch.

gerhardol:
> I've taken @NN--- changes in the .NET Core branch.

@RussKie 
Will .NET 5 be merged soon, should we make a separate PR for this in master or keep this open?

RussKie:
This was incorporated into #8522. Thank you.

PR #5087, opened 2018-06-17
UI tweaks for v3

drewnoakes:
Fixes #3072. Fixes #3978. Fixes #4195. Fixes #4892. Fixes #4902. Fixes #5026. Fixes #5032. Fixes #5036. Fixes #5074. Fixes #5149.

@RussKie has said he'd like to get an RC together for version 3.0 towards the end of the month.

It's been a long time since 2.0, so 3.0 would understandably come with some expectation of greatness!

The dashboard has seen significant improvement. GE has a new side panel. The application's innards have been tidied and sped up a fair bit.

I wanted the UI to have a more general refresh, in a bunch of subtle ways, so that it felt like 3.0 was really a new version. This PR attempts to modernise the UI in advance of 3.0 RC.

Apologies for the size of the PR. A lot of the changes are refactorings to separate out parts of the revision grid and revision graph, which both saw a lot of visual changes.

In addition to visual changes, I profiled a bunch of the rendering code for both memory and CPU usage, and made improvements. I won't discuss them here, but they should be easy to find in the commit list.

---

# FormBrowse

Before:

image

After:

image

image

## RevisionGrid

image

Each column is now backed by its own `ColumnProvider` subclass. This gives fine grained control over things such as rendering, tooltips, and so forth.

Refs (branches, tags, ...) have a flatter appearance, and more balanced internal padding.

image
image
image

Graph lines no longer have thin black borders, and the line between nodes is slightly darker than the nodes. The colours are a bit desaturated and therefore less like the EGA graphics palette.

image
image
image

A new column shows the author's avatar in the grid directly.

The message column shows the 'multiline message indicator' as a graphic rather than text `[...]`.

image
image
image

Thanks to @gerhardol, artificial commits now visually break down changes changes to the right of the tile (the before image here is from 2.51.02, not current master, which does show "Working directory" and "Commit index" tiles, but on the right of the text, not the left).

image
image
image

All columns now use the same font.

The Object ID (SHA-1) column shows enough characters to fit without trimming. It won't show ellipsis any more. Fixes #5032.

User can show/hide individual columns. Toggling these no longer reloads all revisions, where possible:

image

The ability to cycle between multiple grid layouts has been removed. Fixes #5026.

UI for empty and bare repos is no longer constructed unless needed, improving startup time and reducing flickering.

Dark borders have been removed from panels, and padding/margin tightened so everything fits more cleanly.

Author highlighting no longer sets the row background colour. Instead the author's name is highlighted in bold.

image

image

[...]

PR #8163, opened 2020-05-30
Shell extension icons

gerhardol:
## Proposed changes

Shell extensions icons are not updated to the changes in 3.x by @drewnoakes in #5087 (some are partly adapted to 3.0). Some icons are refreshed by @NikolayXHD for the dark theme and do look better also in light themes, does not seem be be any of these though.

- Renamed icon files so they match the GE icons 
- Converted the icons with https://hnet.com/png-to-ico/
- Removed a few unused png files
- Changed ResetToFile icon from blue to red in Browse, to be consistent with Commit and ShellExt

Not high priority for myself really, but several co workers use the shell extensions a lot, I then saw the GE2.x icons...

I looked at the VSIX icons too, they seem to be in sync. (As the code to adapt icons in GE is not available in vsix, some icons could be improved.)

To be squashed

### Before

shext-file
shext-dir
shext-nogit

### After

shext-dir-2
shext-file-2
shext-nogit-2

## Test methodology

Dark mode in Win10 100%dpi

----


tompollok:
I think the old difftool icon was more intuitive. Also reset in red is better to not confuse it with checkout.

gerhardol:
> I think the old difftool icon was more intuitive. Also reset in red is better to not confuse it with checkout.

Thanks for the comments but can you open a separate issue?
The master UI will not be changed now and the shell extensions should follow the master UI.

NikolayXHD:
> > I think the old difftool icon was more intuitive. Also reset in red is better to not confuse it with checkout.
> 
> Thanks for the comments but can you open a separate issue?
> The master UI will not be changed now and the shell extensions should follow the master UI.

We do have red reset icon in 3.3.1 at least

image

It is `GitUI/Resources/Icons/ResetWorkingDirChanges.png`

gerhardol:
> We do have red reset icon in 3.3.1 at least

In Browse Revert commit is purple, Reset is red for commits, green for branches, blue for files and red for lines
In FormCommit Reset file and line are both red.
So consistency is needed

NikolayXHD:
> > We do have red reset icon in 3.3.1 at least
> 
> In Browse Revert commit is purple, Reset is red for commits, green for branches, blue for files and red for lines
> In FormCommit Reset file and line are both red.
> So consistency is needed

I believe red is used in FormCommit because it's a dangerous operation, one can lose uncommited changes, which is also true for `Reset file changes...` menu item in Shell extension context menu, making red icon the consistent choise here.

gerhardol:
(edited)
> In Browse Revert commit is purple, Reset is red for commits, green for branches, blue for files, Revert lines is blue
In FormCommit Reset file and line are both red.

Pushing a change to change to red reset for "maybe destructive operations", slightly more consistent.
For Browse red for files.
Keeping green for branches, even if Reset hard will reset work tree changes (but that is red with a description).
Revert line in file viewer kept as blue as the operation will normally fail right now anyway. (This is changed in #7825 that has Reset line if artificial commits are selected.)

Screenshots are not updated.

gerhardol:
The new icon files were much bigger from 16x16 and 3KB to 256x256 to 100 KB
I converted them again with another service to 16x16, with size down to 2KB

Updated description and screenshots

RussKie:
Thank you

PR #8170, opened 2020-06-01
FormCommit: Add context menu items Stage/Unstage

### LINKED ISSUE #8106: FormCommit context menu order changed

gerhardol:
Fixes #8106 

## Proposed changes

Put bold stage/unstage menu items in FileStatusList context menus

To avoid adding a translation string, the Text for the buttons are reused

### Before

Before has #8137 applied - this PR is not necessarily an alternative to the second commit, both could be merged (even if my preference is that the second commit do not go in)

image

image

### After

image

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -268,6 +268,9 @@ namespace GitUI.CommandsDialogs
             commitAuthorStatus.ToolTipText = _commitCommitterToolTip.Text;
             skipWorktreeToolStripMenuItem.ToolTipText = _skipWorktreeToolTip.Text;
             assumeUnchangedToolStripMenuItem.ToolTipText = _assumeUnchangedToolTip.Text;
+            stageToolStripMenuItem.Text = toolStageItem.Text;
+            stagedUnstageToolStripMenuItem.Text = toolUnstageItem.Text;
+
             toolAuthor.Control.PreviewKeyDown += (_, e) =>
             {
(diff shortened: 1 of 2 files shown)


RussKie:
Thank you, but "Open with difftoll" is really out of place.
It should be:

```
**Stage**
Reset file or directory changes
Reset chunk
Interactive Add
----
Open with difftool
...
```

```
**Unstage**
Reset file or directory changes
---
Open with difftool
...
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -268,6 +268,9 @@ public FormCommit([NotNull] GitUICommands commands, CommitKind commitKind = Comm
             commitAuthorStatus.ToolTipText = _commitCommitterToolTip.Text;
             skipWorktreeToolStripMenuItem.ToolTipText = _skipWorktreeToolTip.Text;
             assumeUnchangedToolStripMenuItem.ToolTipText = _assumeUnchangedToolTip.Text;
+            stageToolStripMenuItem.Text = toolStageItem.Text;
```
gerhardol, line 271: This is temporary for release/3.4, translation string to be used in later releases (translation will be affected which we do not want now)

gerhardol:
> Thank you, but "Open with difftoll" is really out of place.

This is before #8137 , the order is a separate discussion (that could be affected by this PR)

PR #8175, opened 2020-06-02
FileStatusList context menu order

### LINKED ISSUE #8106: FormCommit context menu order changed

gerhardol:
Fixes #8106
Follow up to #7961 etc

## Proposed changes

FormCommit, RevDiff and RevFileHistory
 * Move OpenDiffTool down one group
 * Move Save, Edit down below other Open
 * RevDiff: Move Stage/Unstage (in WorkTree/Index)  to above Reset (consistent with FormCommit)
 * RevDiff, RevFile: show filehistory/openSubmodule as bold
 * Reordered 'enable' options in non Designer to match presentation order

### Before

See #7961 or something

### After

image
image

image
image

image
image

image
image

image
image

image
image
image

image

## Test methodology 

Manual - see screenshots for order

----


pmiossec:
Shortcuts for stage/unstage are missing could you add to the `FormCommit.cs`:

```
            stageToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.StageSelectedFile);
            stagedUnstageToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.UnStageSelectedFile);
```

RussKie:
A quick glance from a phone at the screenshots- looks good.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -457,39 +458,38 @@ private void UpdateStatusOfMenuItems()
             // Many options have no meaning for artificial commits or submodules
             // Hide the obviously no action options when single selected, handle them in actions if multi select
 
-            openWithDifftoolToolStripMenuItem.Enabled = _revisionDiffController.ShouldShowDifftoolMenus(selectionInfo);
-            saveAsToolStripMenuItem1.Visible = _revisionDiffController.ShouldShowMenuSaveAs(selectionInfo);
-            copyFilenameToClipboardToolStripMenuItem1.Enabled = _revisionDiffController.ShouldShowMenuCopyFileName(selectionInfo);
+            // open submodule is added in FileStatusList
+            fileHistoryDiffToolstripMenuItem.Font = selectionInfo.IsAnySubmodule && AppSettings.OpenSubmoduleDiffInSeparateWindow
```
gerhardol, line 462: This check should use selectedItem instead, that is what FileStatusList uses for the submodule item so two menu items could be bold.
Force pushing

RussKie:
I'll port my changes 2d0ebe4adcc1da52525e1e7e370ddacd0a277997 here, if that's ok with you

RussKie:
I'd like to hijack this but because there is another change I'd like to do as part of this update.

I'm finding it super challenging to distinguish between "Reset the current branch to here..." and "Reset another branch to here...", and as a result I have invoked the incorrect action multiple times. Again this is a lot to do with a muscle memory, but these two menu items look almost indistinguishable, and thus create an accessibility issue:
image

I'd like the second menu item "Reset another branch to here..." moved further down and away from the current section (that deals with the _current branch_). 
I don't have a clear idea, but may be it can be moved to _"any" branch_ section after "Delete branch...", e.g.:
```
Reset the current branch to here...
----------
Create new branch here...
Delete branch...
Reset another branch to here...
----------
```

pmiossec:
>I'm finding it super challenging to distinguish between "Reset current branch to here..." and "Reset another branch to here...", [...]

I like the idea. Maybe we could also change the icon for the same but with another color.
I don't know if there is a meaning for the different colors (otherwise I don't get it) but a red one could worth it...

gerhardol:
> I'll port my changes 2d0ebe4 here, if that's ok with you

Of course, but #8137 could have been merged with just that commit
We are all users with some wishes of design, not designers so some changes will require some extra turns

> I'd like to hijack this but because there is another change I'd like to do as part of this update.
I'm finding it super challenging to distinguish between "Reset current branch to here..." and "Reset another branch to here..."

Personally, I find it easier if similar items are close together, then it is easier to find the other if I hit the wrong item first.

I propose to change one of them to use the blue reset icon (likely the current branch) or to remove the icon for Another branch to separate it (or "faded" or a new icon, but I rather not take that discussion now).

Bug: Stage menu alternative is missing for FormCommit: WorkTree: Submodule
(not this PR, but will address it here).

RussKie:
image

I have it locally, can push it in.
Can't do the icon, sorry, have no setup for it

gerhardol:
> Can't do the icon, sorry, have no setup for it

The blue version is in the repo already (maybe not used at all)
Removing may be clearer

I would like to keep all branch related items together

RussKie:
As I said, this is an accessibility issue for me, these two items are hard to distinguish. Changing an icon won't help.

There are several distinct section, two of which deal with branches - orange ("current branch") and green ("other"):
image

The separation between the orange and the green is certainly debatable, and can be argued it is to an extent is illogical (e.g. "Checkout branch" fits with "Create new branch", "Rename branch" and "Delete branch"). However placing "Reset another branch" in the same section where "current branch" options are is illogical, and it should be moved.

gerhardol:
> The separation between the orange and the green is certainly debatable, and can be argued it is to an extent is illogical (e.g. "Checkout branch" fits with "Create new branch", "Rename branch" and "Delete branch"). However placing "Reset another branch" in the same section where "current branch" options are is illogical, and it should be moved.

The separation is likely due to the number of items, the order is not completely logical (probably I that did the cleanup from where branches/tags/commits were mixed, trying to keep current branch and the most important first).
In this case will separating the items make it harder to read, but let us stop discussing and release.

If no one else is changing this, I will put this last in the second group.

gerhardol:
Move Reset Another branch from Current to Other group
Makes a little more sense to me than second last

FormCommit: WorkTree submodule context menu missed staged 
Cleanup from #8170 (that menu is separate)

Will squash the FormCommit commit within hours

I restrained from fixing more menu items...

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -457,39 +458,38 @@ private void UpdateStatusOfMenuItems()
...
+            // open submodule is added in FileStatusList
+            fileHistoryDiffToolstripMenuItem.Font = (DiffFiles.SelectedItem?.Item.IsSubmodule ?? false) && AppSettings.OpenSubmoduleDiffInSeparateWindow
+                ? new Font(fileHistoryDiffToolstripMenuItem.Font, FontStyle.Regular)
+                : new Font(fileHistoryDiffToolstripMenuItem.Font, FontStyle.Bold);
+
+            diffUpdateSubmoduleMenuItem.Visible =
+                diffResetSubmoduleChanges.Visible =
+                    diffStashSubmoduleChangesToolStripMenuItem.Visible =
+                        diffCommitSubmoduleChanges.Visible =
+                            submoduleStripSeparator.Visible = _revisionDiffController.ShouldShowSubmoduleMenus(selectionInfo);
```
mstv, line 470: Nit: indentation.

gerhardol: Not changing now, same as before
This is how the code is reformatted automatically (e.g. with semicolon at the end)

gerhardol:
> Stage does not appear in FormCommit:

Did a fixup to the fixup for Stage for submodules

RussKie:
Merge at will

gerhardol:
> Merge at will

Since I did a fixup, I cannot do that until tonight...

gerhardol:
@msftbot merge in 6 hours

gerhardol:
> Why does this - adding stageSubmoduleToolStripMenuItem - fix my issue with a normal file? Because adding an item to a menu removes it from where it was added before!

I added the menu item to stage recently to the normal work tree context menu. I thought I could add the same item to the separate submodule item. I do not know what occurred when I tested that this was possible, but apparently that is not possible which you found. So I added a new menu item to the separate submodule menu, which restored the item in the normal worktree menu.

RussKie:
You're awesome :)

PR #8182, opened 2020-06-04
Make scripts execution safer

RussKie:
## Proposed changes

* Scripts may not have arguments, pass "" in this case instead.

* Scripts may fail during execution, show an error message instead of crashing the process.

NB: there is a new string, but I'm fine with not translating it, as the use case it likely niche.

### DIFF
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -152,6 +152,9 @@ namespace GitUI.Script
                 }
                 else
                 {
+                    // It is totally valid to have a command without an argument, e.g.:
+                    //    Command  : myscript.cmd
+                    //    Arguments: <blank>
                     new Executable(command, module.WorkingDir).Start(argument ?? string.Empty);
                 }
             }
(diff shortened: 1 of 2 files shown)


GitUI/Script/ScriptRunner.cs
```diff
@@ -152,7 +152,7 @@ string optionDependingOnSelectedRevision
                 }
                 else
                 {
-                    new Executable(command, module.WorkingDir).Start(argument);
+                    new Executable(command, module.WorkingDir).Start(argument ?? string.Empty);
```
gerhardol, line 155: Is this the right thing to do?
ArgumentString seem to require an argument (also if not provided), so throwing is less worse, but maybe a comment?

RussKie: I found it by creating the following scripts, which is totally valid:
image

I can add a comment

gerhardol: Then all scripts have defined an argument, even if it is null. It is not possible to run the scripts without the null argument exists.
If the script parses the empty argument, then it must be adjusted - there are no such scripts now.

RussKie:
Yes, I think you're right. TextBox.Text is never null, whereas the proxy
object contains nulls.
I didn't think of it....

mstv:
👍 You found and fixed it before the pre-release.

PR #8212, opened 2020-06-11
Recover lost object: Add a file preview

pmiossec:
## Proposed changes

- Add a file/commit diff preview in the "Recover lost object" form to browse blob more quickly and easily
- Try to guess the file type to enable syntax highlighting

## Screenshots 

For Commits:
### Before

image

### After

image

For Files:
### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 178ef2ae3b6126c29662930505e14f17589fba8b
- .NET Framework 4.8.4180.0
- DPI 168dpi (175% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormVerify.cs
+++ b/GitUI/CommandsDialogs/FormVerify.cs
@@ -29,6 +29,18 @@ namespace GitUI.CommandsDialogs
 
         private LostObject _previewedItem;
 
+        private static readonly Dictionary<string, string> LanguagesStartOfFile = new Dictionary<string, string>
+        {
+            { "{", "recovery.json" },
+            { "#include", "recovery.cpp" },
+            { "import", "recovery.java" },
+            { "from", "recovery.py" },
+            { "package", "recovery.go" },
+            { "#!", "recovery.sh" },
+            { "[", "recovery.ini" },
+            { "using", "recovery.cs" },
+        };
+
         [Obsolete("For VS designer and translation test only. Do not remove.")]
         private FormVerify()
         {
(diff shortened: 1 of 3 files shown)


RussKie:
Good start! But I'd like to challenge you to take it further 😛 


> image

1. Can "Type" be anything different? If not, is there a point in having this column?
2. In the revision grid the hash is shortened, should we show shortened hashes here as well?
3. Remove "Cancel" button, we don't have it anywhere else. To cancel - close the dialog
4. Apply the same settings/styling to the gridview as `RemoteBranches` in `FormRemotes`.

> image

ditto

GitUI/CommandsDialogs/FormVerify.cs
```diff
@@ -216,6 +218,68 @@ private void Warnings_CellMouseDoubleClick(object sender, DataGridViewCellMouseE
...
+
+            _previewedItem = CurrentItem;
+
+            var content = Module.ShowObject(_previewedItem.ObjectId);
+            if (_previewedItem.ObjectType == LostObjectType.Commit)
+            {
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    await fileViewer.ViewPatchAsync("commit.patch", content, null);
+                }).FileAndForget();
```
RussKie, line 236: I think it could be this
```suggestion
                ThreadHelper.JoinableTaskFactory.RunAsync(() => fileViewer.ViewPatchAsync("commit.patch", content, null))
                    .FileAndForget();
```

GitUI/CommandsDialogs/FormVerify.cs
```diff
@@ -216,6 +218,68 @@ private void Warnings_CellMouseDoubleClick(object sender, DataGridViewCellMouseE
...
+            {
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                {
+                    var filename = GuessFileTypeWithContent(content);
+                    await fileViewer.ViewTextAsync(filename, content ?? string.Empty, null);
+                }).FileAndForget();
+            }
+        }
+
+        private static readonly Dictionary<string, string> LanguagesStartOfFile = new Dictionary<string, string>
```
RussKie, line 248: Please move it to the top

pmiossec:
>Good start! But I'd like to challenge you to take it further 😛

@RussKie 

Ok, you ask me to redesign this form.
I will do it.

Another thing I always wished to change are these 2 checkbox on the first line that should be radio buttons form me. What do you think? Either you're looking for commits or for blob but never both at the same time...

But let's agree to not do it in this PR and open a new one once the work is done....

RussKie:
> Another thing I always wished to change are these 2 checkbox on the first line that should be radio buttons form me. What do you think? Either you're looking for commits or for blob but never both at the same time...

I have never used this form, so I'm not sure what these switches are...
I'd follow Windows UI design guidelines:
* https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-check-boxes
* https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-radio-buttons
* https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-drop, etc.

> But let's agree to not do it in this PR and open a new one once the work is done....

Sure

pmiossec:
>I have never used this form, so I'm not sure what these switches are...

That's the checkbox "Show commits and tag" and "Show other objects" (that show in fact `blob` and very rarely `tree` objects).

For me it makes no sense to display all that types in the same time (even more because perfs are even worst when displayed together). Because you want to either retrieve commits or file contents. Never both at the same times. So you want radio buttons...

RussKie:
Looks good as an interation.

PR #8213, opened 2020-06-11
FormBrowse: Add all supported shells to the toolbar

### LINKED ISSUE #6460: Being able to open a powershell or cmd console
## Feature description
Discussed in https://github.com/gitextensions/gitextensions/issues/6452#issuecomment-481248777

Transform the `Git bash` button in  a split button to be able to easily open a cmd, powershell console (and a custom one? #1393).

We could also uplift the Console tab to launch different sessions, similar to VSCode ( cf https://github.com/gitextensions/gitextensions/issues/6452#issuecomment-481626505 )

## Environment

- Git Extensions 3.1.0
- Build 84b65c3f6542f6adb817c308906522c66c611009
- Git 2.21.0.windows.1
- .NET Framework 4.7.3362.0
- DPI 192dpi (200% scaling)

pmiossec:
Fixes #6460

## Proposed changes

- Being able to launch console of other shells (powershell, cmd, pwsh)

### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 9bf0b97406a3da73fdc9410defa88c7645eac230
- .NET Framework 4.8.4180.0
- DPI 168dpi (175% scaling)


RussKie:
Nice!
Do you think we could get different icons for each?

GitUI/ShellHelper.cs
```diff
@@ -10,6 +10,7 @@ namespace GitUI
 {
     public static class ShellHelper
     {
+        public static string[] SupportedShells => new[] { "bash", "cmd", "powershell", "pwsh" };
```
RussKie, line 13: ❗  Public mutable API
Please use something like `ReadOnlyCollection<T>`

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -495,6 +499,17 @@ void SetDefaultPullActionMenuItemClick(object sender, EventArgs eventArgs)
...
 
+        private void FillBashButtonWithShells()
+        {
+            GitBash.DropDownItems.AddRange(ShellHelper.SupportedShells.Select(s =>
+            {
+                var toolStripMenuItem = new ToolStripMenuItem(s);
+                toolStripMenuItem.Image = Images.Console;
+                toolStripMenuItem.Click += RunShell;
+                return toolStripMenuItem;
+            }).ToArray());
```
RussKie, line 510: Multiple allocations, please refactor using `foreach`

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -76,6 +76,8 @@ public sealed partial class FormBrowse : GitModuleForm, IBrowseRepo
         private readonly TranslationString _undoLastCommitText = new TranslationString("You will still be able to find all the commit's changes in the staging area\n\nDo you want to continue?");
         private readonly TranslationString _undoLastCommitCaption = new TranslationString("Undo last commit");
 
+        private readonly TranslationString _runShellFailed = new TranslationString("Fail to run the shell. Reason: ");
```
mstv, line 79: "Failed to ..." or "Failure to ..."?
```suggestion
        private readonly TranslationString _runShellFailed = new TranslationString("Failed to run the shell. Reason:");
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1311,7 +1326,23 @@ private void ApplyPatchToolStripMenuItemClick(object sender, EventArgs e)
...
+        private void RunShell(object sender, EventArgs e)
+        {
+            try
+            {
+                string shellPath = ShellHelper.GetShellPath(((ToolStripMenuItem)sender).Text);
+                new Executable(shellPath, Module.WorkingDir).Start(createWindow: true);
+            }
+            catch (Exception exception)
+            {
+                MessageBox.Show(_runShellFailed.Text + "\n" + exception.Message,  Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
```
mstv, line 1344:
```suggestion
                MessageBox.Show($"{_runShellFailed.Text}{Environment.NewLine}{exception.Message}", Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
```
And should this `MessageBox` be moved to `GitUI\MessageBoxes.cs`?

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: Suggest telling which shell failed too:
```suggestion
        private readonly TranslationString _runShellFailed = new TranslationString("Failed to run '{0}' shell.{1}Reason: {2}");
```

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: > And should this `MessageBox` be moved to `GitUI\MessageBoxes.cs`?

👍

pmiossec:
Still need to address messagebox comments but here are custom shell icons... (with updated screenshot)

RussKie:
Looks awesome!

pmiossec:
Message boxes stuff done (with some not translated more strings...)

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -75,7 +75,6 @@ public sealed partial class FormBrowse : GitModuleForm, IBrowseRepo
 
         private readonly TranslationString _undoLastCommitText = new TranslationString("You will still be able to find all the commit's changes in the staging area\n\nDo you want to continue?");
         private readonly TranslationString _undoLastCommitCaption = new TranslationString("Undo last commit");
-
```
mstv, line 78: nit: removed too much?

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1311,7 +1321,24 @@ private void ApplyPatchToolStripMenuItemClick(object sender, EventArgs e)
 
         private void GitBashToolStripMenuItemClick1(object sender, EventArgs e)
         {
-            Module.RunBash();
+            if (!GitBash.DropDownButtonPressed)
+            {
+                Module.RunBash();
```
mstv, line 1326: This is different from `RunShell` if you don't have `bash.exe` on your path.

GitUI/MessageBoxes.cs
```diff
@@ -35,6 +35,12 @@ public class MessageBoxes : Translate
         private readonly TranslationString _shellNotFoundCaption = new TranslationString("Shell not found");
         private readonly TranslationString _shellNotFound = new TranslationString("The selected shell is not installed, or is not on your path.");
 
+        private readonly TranslationString _scriptExecutionFailed = new TranslationString("Failed to execute '{0}' script. Reason:\n\n{1}");
+        private readonly TranslationString _runShellFailed = new TranslationString("Failed to run '{0}' shell. Reason:\n\n{1}");
```
mstv, line 39: I think the popup reads better with "Failed to run shell 'pwsh'. Reas...":
```suggestion
        private readonly TranslationString _runShellFailed = new TranslationString("Failed to run shell '{0}'. Reason:\n\n{1}");
```

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: I've decided to remove the duplicate one in the menu which has the advantage to solve this issue...

RussKie:
The existing code that deals with shells is starting to rot quickly, turning into a mess that we had with difftools resolution.
I took a liberty and refactored the code to do the following:
* Break out each shell into own type that defines shell-specific definitions and settings
* Refactor and rework FormBrowse and settings screen to use the new shell definitions
* Rework how ConEmu terminal changes folders
* Remove `GitModule.RunBash()`

I also made a change that caused the toolbar button to match the shell selection for the ConEmu. I think it fits into the mental model, but the setting location is now in the "wtf?" territory. So I'm somewhat conflicted on this change. We can revert it, if you don't think it is worthwhile change.

Tests are TBD, happy to help (I had only so much time).

RussKie:
Something I didn't get to, but would be good to have - make `AppSettings.ConEmuTerminal.ValueOrDefault` of type `ShellType` (represented as `int`) instead of `string`. This would allow removal string comparison routines, and remove few allocations.

RussKie:
@pmiossec I'd like to take this into 3.4.2. I have rebased the branch (+1 commit) on to release/3.4 but I was reluctant to push it into your fork. You can see it here - https://github.com/RussKie/gitextensions/tree/launch_shell_consoles

pmiossec:
@RussKie 

>I'd like to take this into 3.4.2.

Tested. It seems to work well.

Not reviewed because there are a lot of changes and I have to take the time to understand it.
But it seems that the behavior when bash is not found has been changed. On purpose?
Before: a message in the "cmd" console open to tell that bash is not found.
After: (I'm not sure) nothing happens.

> I have rebased the branch (+1 commit) on to release/3.4 but I was reluctant to push it into your fork

Go if that's the way forward ;)

RussKie:
> Not reviewed because there are a lot of changes and I have to take the time to understand it.

Fair point. 
It is easier to review commit-by-commit. The general idea is that each shell is implemented as own type (btw 37b2301, 88ea6d2 and df15749 should be squashed for ease of review).

I have also initially changed how the toolbar item was bound to the same shell selected by a user as a ConEmu default shell. But in the last commit e3c0ed4 I've undone that, as it didn't feel completely coherent with the rest of the change.


> But it seems that the behavior when bash is not found has been changed. On purpose?
> Before: a message in the "cmd" console open to tell that bash is not found.

I tried to account for a fact that a user may not have all shells available, and only bind shells that user had.
Entirely possible that I may missed some edge case here...

GitUI/Shells/ShellProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+                    return PwshShellInstance ?? (PwshShellInstance = new PwshShell());
+
+                case "powershell":
+                    return PowerShellShellInstance ?? (PowerShellShellInstance = new PowerShellShell());
+
+                case "cmd":
+                    return CmdShellInstance ?? (CmdShellInstance = new CmdShell());
+
+                case "bash":
+                default:
```
pmiossec, line 26: A `shellType` not found is transformed in a `git-bash`. Strange...

GitUI/Shells/ShellDescriptor.cs
```diff
@@ -0,0 +1,33 @@
...
+        public string? ExecutableCommandLine { get; protected set; } = null!;
+
+        /// <inheritdoc/>
+        public string ExecutableName { get; protected set; } = null!;
+
+        /// <inheritdoc/>
+        public string? ExecutablePath { get; protected set; } = null!;
+
+        /// <inheritdoc/>
+        public bool HasExecutable => ExecutablePath is object;
```
pmiossec, line 19:
```suggestion
        public bool HasExecutable => ExecutablePath != null;
```

GitUI/Shells/ShellProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+    public class ShellProvider
+    {
+        private static PwshShell PwshShellInstance;
+        private static PowerShellShell PowerShellShellInstance;
+        private static CmdShell CmdShellInstance;
+        private static BashShell BashShellInstance;
+
+        public IShellDescriptor GetShell(string shellType)
+        {
+            switch (shellType.ToLowerInvariant())
```
pmiossec, line 14: I'm not 100% sold by the introduction of `ShellType' that:
* seems redundant with names
* makes (a little) more difficult to add a new shell because you should update 2 places (ShellType and here) and there is no help for the developer if there is a mismatch

The only advantage is to be able to so `ShellType.Bash` but for this special case, a `const` could be used.

I will push a commit with a draft in another branch https://github.com/pmiossec/gitextensions/tree/refactor_shells
Please, have a look....

GitUI/GitUICommands.cs
```diff
@@ -1505,9 +1505,6 @@ private void RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                 case "formatpatch":
                     StartFormatPatchDialog();
                     return;
-                case "gitbash":
```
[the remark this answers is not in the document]
RussKie: Yes, I did remove it, as I removed `GitModule.RunBash()` in 14690b0 as it duplicates the functionality of running bash that we have in `FormBrowse` in a way that is excessive.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -495,6 +499,48 @@ void SetDefaultPullActionMenuItemClick(object sender, EventArgs eventArgs)
...
+            bool userShellAccessible = false;
+            foreach (var shellType in Enum.GetNames(typeof(ShellType)))
+            {
+                IShellDescriptor shell = _shellProvider.GetShell(shellType);
+                if (!shell.HasExecutable)
+                {
+                    continue;
+                }
+
+                if (string.Equals(shellType, defaultShell, StringComparison.InvariantCultureIgnoreCase))
```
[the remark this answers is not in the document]
RussKie: Yes, it would. That was my intent, though I figured if you have git, you should have bash.

GitUI/GitUICommands.cs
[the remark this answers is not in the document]
pmiossec: Ok. So I pushed the fixup...

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
pmiossec: >That was my intent, though I figured if you have git, you should have bash.

Not necessarily in the PATH...
So, I think we need to update the button after the foreach with the first shell found if Tag is null...

GitUI/GitUICommands.cs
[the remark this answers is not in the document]
mstv: I think I got why you removed `GitModule.RunBash`, @RussKie. But dropping a CLI in this PR feels like a submarine.
It breaks the usecase: `bash` and e.g. `grep` *not* in the `PATH`, but `gitex.cmd` in the `PATH`.
So `gitex gitbash` should open a bash.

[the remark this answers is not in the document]
RussKie: I mulled over it, and you're probably right - it's a "dot" release, and this could be a breaking behaviour... we can drop  14690b0 from this PR, but I intent do it in the master.
 

> It breaks the usecase: `bash` and e.g. `grep` _not_ in the `PATH`, but `gitex.cmd` in the `PATH`.
> So `gitex gitbash` should open a bash.

If bash is not in the path, then we can't open it. Trying to open bash via the app feels like a wrong.

GitUI/Shells/ShellDescriptor.cs
RussKie: It would be good to write `not is null`, but this is C# 9.0 feature, which hasn't been released yet.
So `is object` or `is {}` is a replacement for `!= null`.

GitUI/Shells/ShellProvider.cs
RussKie: The idea with the enum was to convert the setting to the enum. But your version offers probably equally good alternative. 
I left a comment there.

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
RussKie: 👍

GitUI/Shells/ShellDescriptor.cs
mstv: So let's wait for `not is null`. `is {}` is write-only, `is object` is not much clearer and smells like dynamic cast, although the compiler will be able to optimize it.

RussKie:
Given there are few unresolved and outstanding issues, I'm not going to include this into 3.4.2. On the surface it looked to me as something relatively simple, but few changes may be breaking, thus not fit for a dot release.

GitUI/Shells/ShellDescriptor.cs
RussKie: No worries at all. 
A correction, once VS 16.7 (not C# 9) is shipped we can use `not is null`

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
pmiossec: done.

GitUI/GitUICommands.cs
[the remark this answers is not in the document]
mstv: This removal of the CLI command "gitbash" must be mentioned in the change log.

GitUI/Shells/BashShell.cs
```diff
@@ -0,0 +1,54 @@
...
+                    string posixPath = "/" + directoryInfo.FullName.ToPosixPath().Remove(1, 1);
+                    return $"cd {posixPath.QuoteNE()}";
+                }
+            }
+            catch
+            {
+                // no-op
+            }
+
+            return string.Empty;
```
mstv, line 51:
```suggestion
            catch
            {
                return string.Empty;
            }
```

GitUI/Shells/ConEmuControlExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static class ConEmuControlExtensions
+    {
+        public static void ChangeFolder(this ConEmuControl terminal, IShellDescriptor shell, string path)
+        {
+            if (terminal?.RunningSession == null || shell is null || string.IsNullOrWhiteSpace(path))
+            {
+                return;
+            }
+
+            string command = shell.GetTerminalStartCommand(path);
```
mstv, line 17: This name `GetTerminalStartCommand` is misleading. `GetChangeDirCommand`?

GitUI/MessageBoxes.cs
```diff
@@ -45,9 +51,14 @@ internal MessageBoxes()
 
         private static MessageBoxes Instance => instance ?? (instance = new MessageBoxes());
 
+        public static void ShowErrorMessageBox([CanBeNull] IWin32Window owner, string text, string caption = null)
+        {
+            MessageBox.Show(owner, text, caption ?? Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+        }
```
mstv, line 57:
```suggestion
        public static void ShowErrorMessageBox([CanBeNull] IWin32Window owner, string text, string caption = null)
            => MessageBox.Show(owner, text, caption ?? Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
```
This applies to the changed and new functons below, too.

GitUI/Shells/ShellProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                case "pwsh":
+                    return PwshShellInstance ?? (PwshShellInstance = new PwshShell());
```
mstv, line 17: The shell classes should provide the shell name as a static member.
```suggestion
                case PwshShell.Name:
                    return PwshShellInstance ?? (PwshShellInstance = new PwshShell());
```

GitUI/Shells/ShellType.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace GitUI.Shells
+{
+    public enum ShellType
```
mstv, line 3: Hasn't this become obsolete, has it?

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1440,7 +1511,7 @@ private void ArchiveToolStripMenuItemClick(object sender, EventArgs e)
             var revisions = RevisionGrid.GetSelectedRevisions();
             if (revisions.Count < 1 || revisions.Count > 2)
             {
-                MessageBox.Show(this, @"Select only one or two revisions. Abort.", @"Archive revision", MessageBoxButtons.OK, MessageBoxIcon.Error);
+                MessageBoxes.SelectOnlyOneOrTwoRevision(this);
```
mstv, line 1514:
```suggestion
                MessageBoxes.SelectOnlyOneOrTwoRevisions(this);
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1833,7 +1904,7 @@ private void FileExplorerToolStripMenuItemClick(object sender, EventArgs e)
             }
             catch (Exception ex)
             {
-                MessageBox.Show(this, ex.Message, Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                MessageBoxes.ShowErrorMessageBox(this, ex.Message);
```
mstv, line 1907: It  is not necessary to repeat "MessageBox" in the function name:
```suggestion
                MessageBoxes.ShowError(this, ex.Message);
```

pmiossec:
> Should I prepare a commit with my requested changes to `MessageBoxes`, @pmiossec?

Yes, it would be great!

GitUI/Shells/IShellDescriptor.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        public Image Icon { get; }
+
+        /// <summary>
+        /// Gets the user visible shell name.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// Gets the command to execute when a new terminal starts.
```
RussKie, line 40: The comment needs rewording
```suggestion
        /// Gets the command to execute when a new terminal starts, or a working directory is switched.
```

GitUI/Shells/ConEmuControlExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+using ConEmu.WinForms;
+
+namespace GitUI.Shells
+{
+    public static class ConEmuControlExtensions
+    {
+        public static void ChangeFolder(this ConEmuControl terminal, IShellDescriptor shell, string path)
+        {
+            if (terminal?.RunningSession == null || shell is null || string.IsNullOrWhiteSpace(path))
```
RussKie, line 12:
```suggestion
            if (terminal?.RunningSession is null || shell is null || string.IsNullOrWhiteSpace(path))
```

GitUI/Shells/IShellDescriptor.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        /// <param name="path">The working directory path.</param>
+        /// <returns>The start command.</returns>
+        public string GetChangeDirCommand(string path);
```
RussKie, line 44: Maybe rename to? 
```suggestion
        public string GetConEmuChangeDirCommand(string path);
```
Since this is needed for ConEmu and likely won't work with other consoles 🤔

GitUI/Shells/IShellDescriptor.cs
mstv: Yep, even remove the first part. It's a different kettle of fish that the command is executed on terminal start, too.

GitUI/Shells/IShellDescriptor.cs
mstv: The command is specific to a shell. There is nothing special for ConEmu here.

RussKie:
Squash and merge at will

mstv:
@msftbot merge in 1 day

PR #8237, opened 2020-06-16
ResetAnotherBranch: Quote path arguments

### LINKED ISSUE #8235: "Reset another branch to here" fails with paths containing spaces
## Current behaviour

Using the new "Reset another branch to here..." context menu option fails if the path to the repo contains spaces.

## Expected behaviour

Choosing a branch to reset should result in the chosen branch being reset to the chosen commit.

## Steps to reproduce

1. Create a repo in a directory that contains a space in its name.
2. Create an initial commit on master
3. Create a second branch on the initial commit, without checking it out
4. Create a second commit on master
5. Right-click the second commit and choose "Reset another branch to here..."
6. Choose the branch created on step 3.

## Screenshots

GitExtensions_pGYPgFUBgN

N/A, this is a new feature

## Environment

- Git Extensions 3.4.1.9675
- Git 2.27.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8235

## Proposed changes

Quote path (and gitref, not strictly needed).

## Test methodology

Manual (check Git command log)

----

### DIFF
--- a/GitCommands/Git/GitCommandHelpers.cs
+++ b/GitCommands/Git/GitCommandHelpers.cs
@@ -240,8 +240,8 @@ namespace GitCommands
         {
             return new GitArgumentBuilder("push")
             {
-                $"file://{repoPath}",
-                $"{targetId}:{gitRef}",
+                $"file://{repoPath}".RemoveTrailingPathSeparator().QuoteNE(),
+                $"{targetId}:{gitRef}".QuoteNE(),
                 { force, "--force" }
             };
         }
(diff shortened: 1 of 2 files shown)


gerhardol:
Rebased and added a fish-need-bicycle test

RussKie:
> added a fish-need-bicycle test

Strongly disagree here. If we had tests we likely wouldn't have a bug. Since we have a bug and provided a fix it is prudent to add a test confirming the fix and guard against regressions.

gerhardol:
> The use of the 'Worktree' pseudo objectid is a little disturbing because it should never be called with this value but that has no impact on the tests behavior,

I agree, but the alternative would be to insert a random id in the expected test, which really is not better

> Strongly disagree here. If we had tests we likely wouldn't have a bug. Since we have a bug and provided a fix it is prudent to add a test confirming the fix and guard against regressions.

This test does not test anything, it is just a parallel implementation. But if you are happy, that is the important for the project (even if I spend less time).

PR #8238, opened 2020-06-16
Display 'Plugin Manager" next to "Plugin settings" in the menu

pmiossec:
to make it more visible.

Improvement on https://github.com/gitextensions/GitExtensions.GerritPlugin/issues/13#issuecomment-644720345

### Before

image

### After

image

## Test methodology 

- Manual 

## Test environment(s) 

- Git Extensions 33.33.33
- Build f2d6c3a8bc1899389911fc6b1a8e8440945ba428
- .NET Framework 4.8.4180.0
- DPI 168dpi (175% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -760,10 +760,10 @@ namespace GitUI.CommandsDialogs
                 var pluginEntries = PluginRegistry.Plugins
                     .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
-                // the menu already contains 2 items:
-                //    [0] Separator
-                //    [1] Plugin Settings
-                // insert plugins above the separator (the list is in reverse order)
+                // pluginsToolStripMenuItem.DropDownItems menu already contains 2 items:
+                //    [1] Separator
+                //    [0] Plugin Settings
+                // insert all plugins except 'Plugin Manager' above the separator
                 foreach (var plugin in pluginEntries)
                 {
                     // don't add the plugin to the Plugins menu, if already added
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -760,7 +760,7 @@ namespace GitUI.CommandsDialogs
                 var pluginEntries = PluginRegistry.Plugins
                     .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
-                // pluginsToolStripMenuItem.DropDownItems menu already contains 2 items:
+                // pluginsToolStripMenuItem.DropDownItems menu already contains at least 2 items:
                 //    [1] Separator
                 //    [0] Plugin Settings
                 // insert all plugins except 'Plugin Manager' above the separator
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -752,34 +752,48 @@ namespace GitUI.CommandsDialogs
 
         private void RegisterPlugins()
         {
+            const string PluginManagerName = "Plugin Manager";
             var existingPluginMenus = pluginsToolStripMenuItem.DropDownItems.OfType<ToolStripMenuItem>().ToLookup(c => c.Tag);
 
             lock (PluginRegistry.Plugins)
             {
                 var pluginEntries = PluginRegistry.Plugins
-                    .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
+                    .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // the menu already contains 2 items:
+                //    [0] Separator
+                //    [1] Plugin Settings
+                // insert plugins above the separator (the list is in reverse order)
                 foreach (var plugin in pluginEntries)
                 {
-                    // Add the plugin to the Plugins menu, if not already added
-                    if (!existingPluginMenus.Contains(plugin))
+                    // don't add the plugin to the Plugins menu, if already added
+                    if (existingPluginMenus.Contains(plugin))
                     {
-                        var item = new ToolStripMenuItem
-                        {
-                            Text = plugin.Description,
-                            Image = plugin.Icon,
-                            Tag = plugin
-                        };
-                        item.Click += delegate
-                        {
-                            if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
-                            {
-                                RefreshRevisions();
-                            }
-                        };
+                        continue;
+                    }
 
-                        pluginsToolStripMenuItem.DropDownItems.Insert(pluginsToolStripMenuItem.DropDownItems.Count - 2,
-                            item);
+                    var item = new ToolStripMenuItem
+                    {
+                        Text = plugin.Description,
+                        Image = plugin.Icon,
+                        Tag = plugin
+                    };
+                    item.Click += delegate
+                    {
+                        if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
+                        {
+                            RefreshRevisions();
+                        }
+                    };
+
+                    if (plugin.Name == PluginManagerName)
+                    {
+                        // insert Plugin Manager below the separator
+                        pluginsToolStripMenuItem.DropDownItems.Insert(pluginsToolStripMenuItem.DropDownItems.Count - 1, item);
+                    }
+                    else
+                    {
+                        pluginsToolStripMenuItem.DropDownItems.Insert(0, item);
                     }
                 }
             }


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -755,31 +755,42 @@ private void RegisterPlugins()
 
             lock (PluginRegistry.Plugins)
             {
-                var pluginEntries = PluginRegistry.Plugins
+                var pluginManager = PluginRegistry.Plugins.FirstOrDefault(p => p.Name == "Plugin Manager");
```
RussKie: This isn't a localisable property, is it?

RussKie:
Please rebase on 3.4 for 3.4.2. Thank you

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -758,27 +758,46 @@ private void RegisterPlugins()
                 var pluginEntries = PluginRegistry.Plugins
                     .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // the menu already contains 2 items:
+                //    [0] Plugin Settings
+                //    [1] separator
+                // insert plugins above the separator (the list is in reverse order)
+                int insertOffset = pluginsToolStripMenuItem.DropDownItems.Count - 2;
```
mstv, line 765: This will be incorrect when `RegisterPlugins` is called a second time and the Plugin Manager was added in the previous call.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -758,27 +758,46 @@ private void RegisterPlugins()
...
-                            if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
-                            {
-                                RefreshRevisions();
-                            }
-                        };
+                            RefreshRevisions();
+                        }
+                    };
+
+                    if (plugin.Name == "Plugin Manager")
```
mstv, line 789: DRY, especially no magic constants.
```suggestion
                    bool isPluginManager = plugin.Name == "Plugin Manager";
                    if (isPluginManager )
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -758,27 +758,46 @@ private void RegisterPlugins()
...
+                    {
+                        // insert Plugin Manager below the separator
+                        insertOffset = pluginsToolStripMenuItem.DropDownItems.Count - 1;
+                    }
 
-                        pluginsToolStripMenuItem.DropDownItems.Insert(pluginsToolStripMenuItem.DropDownItems.Count - 2,
-                            item);
+                    pluginsToolStripMenuItem.DropDownItems.Insert(insertOffset, item);
+
+                    if (plugin.Name == "Plugin Manager")
```
mstv, line 797:
```suggestion
                    if (isPluginManager)
```

pmiossec:
@RussKie I have made some changes that I think make the intent clearer...

What do you think?

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -759,27 +759,43 @@ private void RegisterPlugins()
                 var pluginEntries = PluginRegistry.Plugins
                     .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // the menu already contains 2 items:
+                //    [0] Separator
+                //    [1] Plugin Settings
+                // insert plugins above the separator (the list is in reverse order)
+                int insertOffsetFromBelow = 2;
```
mstv, line 766: Yes, this is clearer but still needs:
```suggestion
                const string PluginManager = "Plugin Manager";
                int insertOffsetFromBelow = existingPluginMenus.Contains(PluginManager) ? 3 : 2;
```

pmiossec: Thinking of it, if we invert the order of the list and so add each plugin at the top (index '0'), the changes are even lighter/clearer.

```csharp
            lock (PluginRegistry.Plugins)
            {
                var pluginEntries = PluginRegistry.Plugins
                    .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);

                // the menu already contains 2 items:
                //    [0] Separator
                //    [1] Plugin Settings
                // insert plugins above the separator (the list is in reverse order)

                foreach (var plugin in pluginEntries)
                {
                    // don't add the plugin to the Plugins menu, if already added
                    if (existingPluginMenus.Contains(plugin))
                    {
                        continue;
                    }

                    var item = new ToolStripMenuItem
                    {
                        Text = plugin.Description,
                        Image = plugin.Icon,
                        Tag = plugin
                    };
                    item.Click += delegate
                    {
                        if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
                        {
                            RefreshRevisions();
                        }
                    };

                    if (plugin.Name == PluginManagerName)
                    {
                        // insert Plugin Manager below the separator
                        pluginsToolStripMenuItem.DropDownItems.Insert(pluginsToolStripMenuItem.DropDownItems.Count - 1, item);
                    }
                    else
                    {
                        pluginsToolStripMenuItem.DropDownItems.Insert(0, item);
                    }
                }
            }
```

Do I miss something? Do you see a bad side effect?

RussKie:
Thank you @mstv for the reviews. 
@pmiossec  I've tested it locally - works well.

RussKie:
@mstv is there something outstanding? I tried it switching between multiple repos and didn't observe any unwanted side effects.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -752,34 +752,51 @@ private void UpdatePluginMenu(bool validWorkingDir)
...
 
             lock (PluginRegistry.Plugins)
             {
                 var pluginEntries = PluginRegistry.Plugins
                     .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // the menu already contains 2 items:
+                //    [0] Separator
+                //    [1] Plugin Settings
+                // insert plugins above the separator (the list is in reverse order)
```
pmiossec, line 766: @RussKie There is this comment that I don't understand.

False or not clear?
Which list is in reverse order? Because the menu items and the plugins **are not** in reversed order...

I proposed a change (https://github.com/gitextensions/gitextensions/pull/8238#discussion_r441880013 ) that reverse the plugins to simplify the code. But I don't know if it handles well edge cases (like duplicate of `PluginManager` in the Application folder and also in the user folder. I don't even know how it is handled now....).

pmiossec: To be clearer, is it possible to have 2 plugins with the same name?

GitUI/CommandsDialogs/FormBrowse.cs
mstv: Yes, this should work and is clearer. And that _comment_ has become true again.
Just remove the blank line between the _comment_ and the `foreach`.

GitUI/CommandsDialogs/FormBrowse.cs
mstv: > This isn't a localisable property, is it?

Valid question.

pmiossec: I looked at that before doing the development and from what I see, no, that's not localisable.

`<joke>But nothing is localized in plugins 🤣 </joke>`

In fact, `Name` is used as the key to store the settings, so I hope it won't be localisable anytime soon....

pmiossec:
Ok. It seems ready to merge...

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: It is certainly not impossible, I don't believe we guard anywhere against this.

> Which list is in reverse order? Because the menu items and the plugins **are not** in reversed order...

`pluginsToolStripMenuItem.DropDownItems` is a stack, so if we use `Add` we would add to the top. But here we use `Insert` to add from the other side of it. That's was "reverse" when I initially wrote that comment. But since then the code has changed few times, and this comment is probably no longer relevant, and should be removed.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -752,34 +752,48 @@ private void UpdatePluginMenu(bool validWorkingDir)
...
             lock (PluginRegistry.Plugins)
             {
                 var pluginEntries = PluginRegistry.Plugins
-                    .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
+                    .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // the menu already contains 2 items:
+                //    [0] Separator
+                //    [1] Plugin Settings
+                // insert plugins above the separator (the list is in reverse order)
```
RussKie, line 766:
```suggestion
                // pluginsToolStripMenuItem.DropDownItems menu already contains 2 items:
                //    [1] Separator
                //    [0] Plugin Settings
                // insert all plugins except 'Plugin Manager' above the separator
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -752,34 +752,48 @@ private void UpdatePluginMenu(bool validWorkingDir)
...
+            const string PluginManagerName = "Plugin Manager";
             var existingPluginMenus = pluginsToolStripMenuItem.DropDownItems.OfType<ToolStripMenuItem>().ToLookup(c => c.Tag);
 
             lock (PluginRegistry.Plugins)
             {
                 var pluginEntries = PluginRegistry.Plugins
-                    .OrderBy(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
+                    .OrderByDescending(entry => entry.Name, StringComparer.CurrentCultureIgnoreCase);
 
+                // pluginsToolStripMenuItem.DropDownItems menu already contains 2 items:
```
RussKie, line 763:
```suggestion
                // pluginsToolStripMenuItem.DropDownItems menu already contains at least 2 items:
```

PR #8239, opened 2020-06-17
RevDiff: Add hotkeys to stage/unstage/reset

gerhardol:
Separating hotkey handling in RevDiff file status list from #7825 

No longer based on #8193 (some conflicts)

## Proposed changes

Add hotkeys to stage/unstage/reset in RevDiff
Checks that the status list is in focus (if the viewer is in focus in #7825, lines will be changed instead).

### After

image

image

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -33,6 +33,9 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _selectedRevision = new TranslationString("Second: b/");
         private readonly TranslationString _firstRevision = new TranslationString("First: a/");
 
+        private readonly TranslationString _resetSelectedChangesText =
+            new TranslationString("Are you sure you want to reset all selected files to {0}?");
+
         private RevisionGridControl _revisionGrid;
         private RevisionFileTreeControl _revisionFileTree;
         private IRevisionDiffController _revisionDiffController;
@@ -109,7 +112,10 @@ namespace GitUI.CommandsDialogs
             OpenAsTempFile = 5,
             OpenAsTempFileWith = 6,
             OpenWithDifftoolFirstToLocal = 7,
-            OpenWithDifftoolSelectedToLocal = 8
+            OpenWithDifftoolSelectedToLocal = 8,
+            ResetSelectedFiles = 9,
+            StageSelectedFile = 10,
+            UnStageSelectedFile = 11,
         }
 
         public CommandStatus ExecuteCommand(Command cmd)
@@ -137,6 +143,9 @@ namespace GitUI.CommandsDialogs
                 case Command.EditFile: diffEditWorkingDirectoryFileToolStripMenuItem.PerformClick(); break;
                 case Command.OpenAsTempFile: diffOpenRevisionFileToolStripMenuItem.PerformClick(); break;
                 case Command.OpenAsTempFileWith: diffOpenRevisionFileWithToolStripMenuItem.PerformClick(); break;
+                case Command.ResetSelectedFiles: return ResetSelectedFilesWithConfirmation();
+                case Command.StageSelectedFile: return StageSelectedFiles();
+                case Command.UnStageSelectedFile: return UnstageSelectedFiles();
 
                 default: return base.ExecuteCommand(cmd);
             }
@@ -156,6 +165,9 @@ namespace GitUI.CommandsDialogs
             diffEditWorkingDirectoryFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.EditFile);
             diffOpenRevisionFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenAsTempFile);
             diffOpenRevisionFileWithToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenAsTempFileWith);
+            resetFileToParentToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.ResetSelectedFiles);
+            stageFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.StageSelectedFile);
+            unstageFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.UnStageSelectedFile);
 
             DiffText.ReloadHotkeys();
         }
@@ -533,6 +545,11 @@ namespace GitUI.CommandsDialogs
 
         private void StageFileToolStripMenuItemClick(object sender, EventArgs e)
         {
+            StageFiles();
+        }
+
+        private void StageFiles()
+        {
             var files = DiffFiles.SelectedItems.Where(item => item.Item.Staged == StagedStatus.WorkTree).Select(i => i.Item).ToList();
 
             Module.StageFiles(files, out _);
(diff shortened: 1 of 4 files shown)


mstv:
> Note that there is no "are you sure?" for the reset

You're right. This is a dangerous function in combination with the hotkey `R`, which is prone to be hit by mistake. (We even had an issue complaining about the single-character hotkeys in the FileStatusList. I strongly vote for keeping them anyways.)
So we should add the confirmation popup for the hotkey here, too.
For the context menu item, it _could_ be omitted because the user must navigate to the submenu.

RussKie:
We have to be really careful here because we will be breaking an establish
user behavior.

If we introduce a warning message it has to be permanently dismissable (use
TaskDialog with Verification).

gerhardol:
Rebased on master, to speedup reviews

Added a message for the hotkey invocation only, similar to other "Reset" invocations. The menu indication here is the only situation I see that do not open a dialog.

The message here is a little special as I want to include the commit to revert to (this would not be needed for invocation from the menu).

>We have to be really careful here because we will be breaking an establish
user behavior.
If we introduce a warning message it has to be permanently dismissable (use
TaskDialog with Verification).

Change defaults is dangerous.
A dismissible dialog is a separate feature, should be global (we do not want to add 10 new options).

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -948,5 +971,62 @@ public void SwitchFocus(bool alreadyContainedFocus)
...
+        /// Hotkey handler
+        /// </summary>
+        /// <returns>true if executed</returns>
+        private bool StageSelectedFiles()
+        {
+            var selectedIds = DiffFiles.SelectedItems.SecondIds().ToList();
+            if (!DiffFiles.Focused || selectedIds.Count != 1 || selectedIds.FirstOrDefault() != ObjectId.WorkTreeId)
+            {
+                return false;
+            }
```
mstv, line 985: The meaning of the return value is rather "hotkey handled" than "command executed".
IIUIC, the return value is used to find another control which handles the same hotkey.
I would not support the reuse of these hotkeys.
So this check should be omitted and the function always returns true. Then we do not need a dedicated function.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -948,5 +971,62 @@ public void SwitchFocus(bool alreadyContainedFocus)
...
+        /// Hotkey handler
+        /// </summary>
+        /// <returns>true if executed</returns>
+        private bool UnstageSelectedFiles()
+        {
+            var selectedIds = DiffFiles.SelectedItems.SecondIds().ToList();
+            if (!DiffFiles.Focused || selectedIds.Count != 1 || selectedIds.FirstOrDefault() != ObjectId.IndexId)
+            {
+                return false;
+            }
```
mstv, line 1001: ditto

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -948,5 +971,62 @@ public void SwitchFocus(bool alreadyContainedFocus)
...
+        /// Hotkey handler
+        /// </summary>
+        /// <returns>true if executed</returns>
+        private bool ResetSelectedFiles()
+        {
+            var parentIds = DiffFiles.SelectedItems.FirstIds().ToList();
+            if (!DiffFiles.Focused || parentIds.Count != 1 || !CanResetToRevision(parentIds.FirstOrDefault()))
+            {
+                return false;
+            }
```
mstv, line 1017: ditto

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -948,5 +971,62 @@ public void SwitchFocus(bool alreadyContainedFocus)
...
+
+            UnstageFiles();
+            return true;
+        }
+
+        /// <summary>
+        /// Hotkey handler
+        /// </summary>
+        /// <returns>true if executed</returns>
+        private bool ResetSelectedFiles()
```
mstv, line 1011:
```suggestion
        private void ResetSelectedFilesWithConfirmation()
```

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -948,5 +971,62 @@ public void SwitchFocus(bool alreadyContainedFocus)
...
+            var parentIds = DiffFiles.SelectedItems.FirstIds().ToList();
+            if (!DiffFiles.Focused || parentIds.Count != 1 || !CanResetToRevision(parentIds.FirstOrDefault()))
+            {
+                return false;
+            }
+
+            var rev = _firstRevision.Text + (DescribeRevision(DiffFiles.SelectedItems.FirstRevs().ToList()) ?? string.Empty);
+            var text = string.Format(_resetSelectedChangesText.Text, rev);
+            if (MessageBox.Show(this, text, _resetChangesCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) !=
+                DialogResult.Yes)
```
mstv, line 1022: Move to `MessageBoxes`, please.

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: The check still need to be there, but return true (since early return is preferred rather than check if allowed).

GitUI/CommandsDialogs/RevisionDiffControl.cs

GitUI/CommandsDialogs/RevisionDiffControl.cs
mstv: Ah OK. Then I would return nothing, i.e. void, and return true directly in `ExecuteCommand`.

Why must `DiffFiles` be focused?
The hotkeys will be reused for operations on selected lines in the `DiffText`.
I think it suffices to check whether `DiffText` has the focus. Then "selected lines" else "selected files".

gerhardol: The logic here (with #7825 included) is the same as in FormCommit, so I propose to keep it and check for FileStatusList focus. 
In #7825 the method return false if it could not be executed, not sure what should be done here...
But I will change these functions to void as suggested.

gerhardol: Changed so return false if FileStatusList is not focused only and consider the key consumed otherwise
Other keyhandling may cause both hotkey and script execution, should probably be changed.

GitUI/MessageBoxes.cs
```diff
@@ -118,5 +119,11 @@ public static void ShellNotFound([CanBeNull] IWin32Window owner)
         {
             MessageBox.Show(owner, Instance._shellNotFound.Text, Instance._shellNotFoundCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
         }
+
+        public static bool ResetSelectedFiles(IWin32Window owner, string text)
+        {
+            return MessageBox.Show(owner, text, Instance._resetChangesCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) ==
+                   DialogResult.Yes;
+        }
```
mstv, line 127:
```suggestion
        public static bool ResetSelectedFiles(IWin32Window owner, string text)
            => MessageBox.Show(owner, text, Instance._resetChangesCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) == DialogResult.Yes;
```

gerhardol: will squash at approval

mstv:
@msftbot merge in 3 days

PR #8250, opened 2020-06-19
Fix "selection only" text search on Diff view (#7784)

### LINKED ISSUE #7784: [NBug] 0 <= offset < 9631. Parameter name: offset.
## Current behaviour

I simply navigate and search between files in a previous commit (File tree). A new search "Find next" give me this error.

## Expected behaviour
Search the text (highlight text and jump to the location)

## Steps to reproduce

I don't know how to reproduce now. Previous searches work fine.

## Error Details
```
System.ArgumentOutOfRangeException: 0 <= offset < 9631
Parameter name: offset
Actual value was 75480.
   at ICSharpCode.TextEditor.Document.GapTextBufferStrategy.GetCharAt(Int32 offset)
   at GitUI.TextEditorSearcher.FindNextIn(Int32 offset1, Int32 offset2, Boolean searchBackward)
   at GitUI.TextEditorSearcher.FindNext(Int32 beginAtOffset, Boolean searchBackward, Boolean& loopedAround)
   at GitUI.FindAndReplaceForm.<FindNextAsync>d__27.MoveNext()
   at GitUI.FindAndReplaceForm.<btnFindNext_Click>d__26.MoveNext()
```

## Additional information
I simply navigate and search between files in a previous commit (File tree). A new search "Find next" give me this error.

## Environment
- Git Extensions 3.3.1.7897
- Git 2.19.1.windows.1 (recommended: 2.23.0 or later)
- .NET Framework 4.8.4075.0
- DPI 96dpi (no scaling)

realkompot:
Fixes #7784

## Proposed changes

- Turn off "selection only" mode in Find dialog if a new text is assigned to the file viewer (see comments in the issue).
- Prevent Diff view from iterating over the files if "selection only" search is being conducted (see isMultiFileSearch in the PR).

## Test methodology 

- Unit tests
- Manual testing of Find dialog on Diff, File tree, File history views.

## Test environment(s) 

- GIT 2.18
- Windows 10 Pro 

----

### DIFF
--- a/GitUI/Editor/FindAndReplaceForm.cs
+++ b/GitUI/Editor/FindAndReplaceForm.cs
@@ -49,18 +49,9 @@ namespace GitUI
             InitializeComponent();
             InitializeComplete();
             _search = new TextEditorSearcher();
+            _search.ScanRegionChanged += ScanRegionChanged;
 
             ShowInTaskbar = false;
-        }
-
-        private TextEditorControl Editor
-        {
-            set
-            {
-                _editor = value;
-                _search.Document = _editor.Document;
-                UpdateTitleBar();
-            }
         }
 
         public bool ReplaceMode
@@ -97,7 +88,7 @@ namespace GitUI
 
         public void ShowFor(TextEditorControl editor, bool replaceMode)
         {
-            Editor = editor;
+            SetEditor(editor);
 
             _search.ClearScanRegion();
             SelectionManager sm = editor.ActiveTextAreaControl.SelectionManager;
@@ -174,7 +165,6 @@ namespace GitUI
                 {
                     // user moved outside of the originally selected region
                     _search.ClearScanRegion();
-                    UpdateTitleBar();
                 }
 
                 int startFrom = caret.Offset - (searchBackward ? 1 : 0);
@@ -183,12 +173,14 @@ namespace GitUI
                     startFrom = _search.EndOffset;
                 }
 
+                var isMultiFileSearch = _fileLoader != null && !_search.HasScanRegion;
+
                 range = _search.FindNext(startFrom, searchBackward, out _lastSearchLoopedAround);
-                if (range != null && (!_lastSearchLoopedAround || _fileLoader == null))
+                if (range != null && (!_lastSearchLoopedAround || !isMultiFileSearch))
                 {
                     SelectResult(range);
                 }
-                else if (_fileLoader != null)
+                else if (isMultiFileSearch)
                 {
                     range = null;
                     if (currentIdx != -1 && startIdx == -1)
@@ -233,6 +225,11 @@ namespace GitUI
             // Also move the caret to the end of the selection, because when the user
             // presses F3, the caret is where we start searching next time.
             _editor.ActiveTextAreaControl.Caret.Position = p2;
+        }
+
+        private void ScanRegionChanged(object sender, EventArgs e)
+        {
+            UpdateTitleBar();
         }
 
         private void btnHighlightAll_Click(object sender, EventArgs e)
@@ -382,6 +379,13 @@ namespace GitUI
             }
         }
 
+        private void SetEditor(TextEditorControl editor)
+        {
+            _editor = editor;
+            _search.Document = _editor.Document;
+            UpdateTitleBar();
+        }
+
         internal void SetFileLoader(GetNextFileFnc fileLoader)
         {
             _fileLoader = fileLoader;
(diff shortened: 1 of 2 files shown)


GitUI/Editor/FindAndReplaceForm.cs
```diff
@@ -401,11 +408,58 @@ protected override void Dispose(bool disposing)
...
+
+                return this;
+            }
+
+            public TestAccessor SetFileLoader(GetNextFileFnc fileLoader)
+            {
+                _control.SetFileLoader(fileLoader);
+
+                return this;
+            }
```
RussKie, line 451: This doesn't look needed, `SetFileLoader` is marked as `internal` so it should be accessible in tests directly.

GitUI/Editor/FindAndReplaceForm.cs
```diff
@@ -401,11 +408,58 @@ protected override void Dispose(bool disposing)
...
+        internal readonly struct TestAccessor
+        {
+            private readonly FindAndReplaceForm _control;
+
+            public TestAccessor(FindAndReplaceForm control)
+            {
+                _control = control;
+            }
+
+            public TestAccessor SetCriteria(string lookFor, bool matchCase = false, bool matchWholeWordOnly = false)
```
RussKie, line 423: The idea behind `TestAccessor` was provide access to private type members without increasing their access level, and not really act as a wrapper type. This is so we can easily replace and remove this implementation with another (which is coming) without (significant) changes to tests.

Let's change this to be something like:
```cs
public TextEditorSearcher _search => _control._search;
public TextBox txtLookFor => _control.txtLookFor;
public CheckBox chkMatchCase => _control. chkMatchCase;
public CheckBox chkMatchWholeWord => _control.chkMatchWholeWord;
```

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Removed.

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Got it. Done.

GitUI/Editor/FindAndReplaceForm.cs
```diff
@@ -503,14 +535,18 @@ public void SetScanRegion(int offset, int length)
...
         }
 
         public void ClearScanRegion()
         {
             if (_region != null)
             {
                 _document.MarkerStrategy.RemoveMarker(_region);
+                _document.TextContentChanged -= DocumentOnTextContentChanged;
                 _region = null;
+                ScanRegionChanged?.Invoke(this, null);
```
RussKie, line 549: Never send `null` for event args, this may lead to NRE
```suggestion
                ScanRegionChanged?.Invoke(this, EventArgs.Empty);
```

GitUI/Editor/FindAndReplaceForm.cs
```diff
@@ -503,14 +535,18 @@ public void SetScanRegion(int offset, int length)
             _region = new TextMarker(offset, length, TextMarkerType.SolidBlock,
                                      Globals.HalfMix(bkgColor, Color.FromArgb(160, 160, 160)));
             _document.MarkerStrategy.AddMarker(_region);
+            _document.TextContentChanged += DocumentOnTextContentChanged;
+            ScanRegionChanged?.Invoke(this, null);
```
RussKie, line 539: ditto

GitUI/Editor/FindAndReplaceForm.cs
```diff
@@ -401,11 +408,36 @@ protected override void Dispose(bool disposing)
...
+            public CheckBox ChkMatchCase => _control.chkMatchCase;
+            public CheckBox ChkMatchWholeWord => _control.chkMatchWholeWord;
+
+            public TextEditorControl Editor
+            {
+                set
+                {
+                    _control.Editor = value;
+                }
+            }
```
RussKie, line 434: Set-only properties are a code smell, in this case please use the same approach as for the other members:
```suggestion
            public TextEditorControl Editor => _control.Editor;
```

UnitTests/GitUI.Tests/Editor/FindAndReplaceFormTests.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            _findAndReplaceForm.Dispose();
+            _textEditorControl.Dispose();
+        }
+
+        [Test, TestCaseSource(typeof(FindAndReplaceTestData), nameof(FindAndReplaceTestData.MatchCaseCases))]
```
RussKie, line 37: Can you please restructure the tests slightly to have the test data defined above each test, similar to how we have it for `TestGetStagedStatus` and `DeleteWithExtremePrejudice_should_return_true_for_empty_input_or_absent_folders`.
This way it is easier to grasp what we're testing and keep the context in your head.

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Suggested code introduces a get-only property which is not needed. The reason this property is set-only is that the underlying _control.Editor is a set-only property (test accessor is supposed to be dumb as you explained and should just elevate access level). I agree that it smells :) my first thought was not to change the original code. I'll convert that underlying FindAndReplaceForm.Editor to be a method named SetEditor if you don't mind and it'll become a method here as well.

RussKie:
Sounds good

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Done

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Done

UnitTests/GitUI.Tests/Editor/FindAndReplaceFormTests.cs
realkompot: Restructured. I also tried to pick better test data so that tests are easier to understand. Please see if it's any better.

GitUI/Editor/FindAndReplaceForm.cs
realkompot: Done

RussKie:
Awesome work!
Please squash and it is good to go.

realkompot:
Thanks. Squashed.

RussKie:
Thank you

PR #8254, opened 2020-06-20
RevDiff: Remove parent diff

### LINKED ISSUE #8253: RevDiff: diff second/first parent to working directory
## Current behaviour

GE has code to compare the parent to the first/second revision to the working directory (see also 2.x below) in RevDiff.

Similar, for FormDiff "Parent to Compare -> Working directory"
(For most of the discussions, this is ignored as FormDiff is now mostly replaces by functionality in RevDiff, I have not often seen anyone use the form.)

Note: There are tests that executes as expected. This problem was introduced around the time that the code was refactored to enable tests (for a very specific part of the functionality, not meaningful at all IMO, but there are other opinions about that).

## Expected behaviour

Remove this functionality, slightly confusing and apparently very little usage. You probably want to use the File History anyway to see commit info (#8193 provides a better interface to make special comparisons too.)

## Steps to reproduce

See screenshots

Rightclick file with parent for both first/second revision where parent-to-second is not the same as first revision

Similar for FormDiff, select a commit in a branch unrelated to current and Compare to current branch.

## Screenshots

image

image

image

2.x

image

## Environment

- Git Extensions 3.4.1.9675
- Git 2.27.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8253 

## Proposed changes

Remove diff to parent of first/second revision

## Test methodology

Removed code, including tests

----

### DIFF
--- a/GitUI/CommandsDialogs/FileStatusListContextMenuController.cs
+++ b/GitUI/CommandsDialogs/FileStatusListContextMenuController.cs
@@ -11,10 +11,6 @@ namespace GitUI.CommandsDialogs
         bool ShouldShowMenuFirstToSelected(ContextMenuDiffToolInfo selectionInfo);
         bool ShouldShowMenuFirstToLocal(ContextMenuDiffToolInfo selectionInfo);
         bool ShouldShowMenuSelectedToLocal(ContextMenuDiffToolInfo selectionInfo);
-        bool ShouldShowMenuFirstParentToLocal(ContextMenuDiffToolInfo selectionInfo);
-        bool ShouldShowMenuSelectedParentToLocal(ContextMenuDiffToolInfo selectionInfo);
-        bool ShouldDisplayMenuFirstParentToLocal(ContextMenuDiffToolInfo selectionInfo);
-        bool ShouldDisplayMenuSelectedParentToLocal(ContextMenuDiffToolInfo selectionInfo);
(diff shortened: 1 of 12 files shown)


mstv:
A few translations must be removed.

gerhardol:
> Thank you. (#8193 should be merged first in order to avoid another review. I think it is better to resolve the simple merge conflict here.)

The issue need to be commented on too, need to give it some days too.

I would like to get #8193, #8254 and #8239 merged, so I can get started with #7825 again...

gerhardol:
> Could you please add screenshots to help visualise the change. I looked at the original issue but wasn't obvious to me what we're retiring.

See 2.x screenshot, not visible in in 3.x (except the red marked item for FormDiff).
No one has reported that this was missing in 3.x.

RussKie:
Squash and :shipit:

gerhardol:
Note: This functionality is mostly working after all. There should be some problem with it (and the logic is complicated).
#8193 and #8275 adds other ways to make special diffs, proposing that is used instead.

Still proposing this is merged (after the other).

gerhardol:
Will rebase and merge after #8275 
Low risk that someone wants to get back this, but do not have to make it harder than necessary to restore

PR #8266, opened 2020-06-24
Fix ssh not found

lanfeust69:
When `ssh.exe` is not in the path, the attempt to find it in git
directory will fail, resulting in a nasty crash of the whole
GitExtensions process if it is needed (an example is when
downloading a patchset in Gerrit plugin).

----

### DIFF
--- a/GitCommands/SshPathLocator.cs
+++ b/GitCommands/SshPathLocator.cs
@@ -54,7 +54,10 @@ namespace GitCommands
 
             try
             {
-                var gitDirInfo = _fileSystem.Directory.GetParent(gitBinDirectory);
+                // gitBinDirectory will normally end with a directory separator
+                // (at least this is what AppSettings.GitBinDir ensures),
+                // but then GetParent() returns the same directory, only without the trailing separator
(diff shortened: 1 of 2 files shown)


pmiossec:
Could you update the `contributors.txt` so that PR could be merged? If you could do it in another PR, that's easier for us.

And also, I'm not sure in which way we could add some (integration) tests for the case handled by this change but it would be a good thing if possible....

lanfeust69:
> Could you update the `contributors.txt` so that PR could be merged? If you could do it in another PR, that's easier for us.

https://github.com/gitextensions/gitextensions/pull/8267

> And also, I'm not sure in which way we could add some (integration) tests for the case handled by this change but it would be a good thing if possible....

I see there is [SshPathLocatorTest](/gitextensions/gitextensions/blob/master/UnitTests/GitCommands.Tests/SshPathLocatorTest.cs), where something could be added, will try to do something...

RussKie:
A test would be great to add.
Thanks!

RussKie:
Thank you

PR #8269, opened 2020-06-24
Split VSIX into gitextensions/gitextensions.vsix

### LINKED ISSUE #6300: VSIX signing and separation from GE repo
As discussed in #6211 (.vsix separated from GE installer)

## Feature description

VSIX has (probably) no relation to the normal app and could be moved to a separate repo
The vsix installer should be signed, similar to the main .exe

## Environment

- GitExtensions version: 3.1.0 dev

RussKie:
Closes #6300 (signing etc followup in the new repo)

The VSIX project moved out to https://github.com/gitextensions/gitextensions.vsix


gerhardol:
Readme should point to the new repo too
(I recall that more files needed to be duplicated, maybe that was not necessary?)

PR #8273, opened 2020-06-26
Add a hint regarding updates of portable builds

mstv:
## Proposed changes

- add hint to `README.md`

Feel free to tinker / move the text how and where you like it better!

### Before

grafik

### After

grafik

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,11 @@ For more details please refer to [versions comparison](https://github.com/gitext
 
 **[Download it now](https://github.com/gitextensions/gitextensions/releases/latest)** or install it with [Chocolatey](https://chocolatey.org/packages/gitextensions) or [AppGet](https://appget.net/packages/i/gitextensions).
 
+If you want to **update a portable version**, you should delete all the files and the subfolders from the existing folder except:
+* _GitExtensions.settings_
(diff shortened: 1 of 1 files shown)


vbjay:
Add backup setting and recent repo extensions if desired.

vbjay:
What about plugins?  We may have plugins in plugin manager. I frankly
haven't looked into it but is there a registered plugins file too?  That
way the plugins get re- downloaded.

RussKie:
Thank you

Would it be more appropriate to add to the actual release page, where a user would download a portable version from?
https://github.com/gitextensions/gitextensions/releases/tag/v3.4.2

gerhardol:
I suggest a link from the release page to the readme

mstv:
> What about plugins?

My knowledge about plugins it very limited. Any info?

> Would it be more appropriate to add to the actual release page, where a user would download a portable version from?

We should have it in both places: It applies to dev builds, too.
Is the release page copied automatically when creating a new releases? Or must it be added manually for every release?

> I suggest a link from the release page to the readme.

Yes, it avoids copies, which tend to get out of sync.

RussKie:
Thank you

PR #8275, opened 2020-06-27
RevFile: Diff any two files

gerhardol:
Based on #8193, for RevisionFileHistory

## Proposed changes

Add _Remember file for diff_ for RevisionFileHistory similar to RevDiff in #8193 The context is shared between the tabs 

### Before

image

### After

image

## Test methodology 

Tests added for RevisionDiff

----

### DIFF
--- a/GitUI/CommandsDialogs/RememberFileContextMenuController.cs
+++ b/GitUI/CommandsDialogs/RememberFileContextMenuController.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿#define CONTRACTS_FULL
+
+using System;
+using GitCommands;
 using GitUI.UserControls;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
@@ -7,6 +10,9 @@ namespace GitUI.CommandsDialogs
 {
     public class RememberFileContextMenuController
     {
+        // Singleton accessor
+        public static RememberFileContextMenuController Default { get; } = new RememberFileContextMenuController();
+
         /// <summary>
         /// The remembered file status item, to diff with other files and commits
         /// </summary>
@@ -15,12 +21,27 @@ namespace GitUI.CommandsDialogs
         // Note that the methods in this class are without side effects (static)
 
         /// <summary>
+        /// Get a FileStatusItem that can be used to compare with
+        /// </summary>
+        /// <param name="name">The file name</param>
+        /// <param name="rev">The Git revision</param>
+        /// <returns>The FileStatusItem</returns>
+        [Pure]
+        public FileStatusItem CreateFileStatusItem(string name, GitRevision rev)
+        {
+            var gis = new GitItemStatus { IsNew = true, Name = name };
+            var fsi = new FileStatusItem(null, rev, gis);
+            return fsi;
+        }
+
+        /// <summary>
         /// Check if this file and commit can be used as the first item in a diff to another item
         /// It must be possible to describe the item as a Git commitish
         /// </summary>
         /// <param name="item">The file status item</param>
         /// <param name="isSecondRevision">true if second revision can be used as the first item</param>
         /// <returns>If the item can be used</returns>
+        [Pure]
         public bool ShouldEnableFirstItemDiff(FileStatusItem item, bool isSecondRevision)
         {
             // First item must be a git reference existing in the revision, i.e. other than work tree
@@ -28,6 +49,7 @@ namespace GitUI.CommandsDialogs
                    && (isSecondRevision ? item.SecondRevision : item.FirstRevision)?.ObjectId != ObjectId.WorkTreeId;
         }
 
+        [Pure]
         public bool ShouldEnableFirstItemDiff(FileStatusItem item)
             => ShouldEnableFirstItemDiff(item, isSecondRevision: false)
                || ShouldEnableFirstItemDiff(item, isSecondRevision: true);
@@ -38,6 +60,7 @@ namespace GitUI.CommandsDialogs
         /// <param name="item">The file status item</param>
         /// <param name="isSecondRevision">true if second revision can be used as the second item</param>
         /// <returns>If the item can be used</returns>
+        [Pure]
         public bool ShouldEnableSecondItemDiff(FileStatusItem item, bool isSecondRevision)
         {
             // Git reference in this revision or work tree file existing on the file system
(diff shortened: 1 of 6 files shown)


GitUI/CommandsDialogs/RememberFileContextMenuController.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        // Note that the methods in this class are without side effects (static)
+
+        /// <summary>
+        /// Get a FileStatusItem that can be used to compare with
+        /// </summary>
+        /// <param name="name">The file name</param>
+        /// <param name="rev">The Git revision</param>
+        /// <returns>The FileStatusItem</returns>
+        public FileStatusItem getFileStatusItem(string name, GitRevision rev)
```
mstv, line 27:
```suggestion
        public static FileStatusItem CreateFileStatusItem(string name, GitRevision rev)
```
Though I am not 100% sure regarding `Create...` vs. `Get...`.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -63,6 +65,7 @@ public RevisionFileTreeControl()
             _revisionFileTreeController = new RevisionFileTreeController(() => Module.WorkingDir,
                                                                          new GitRevisionInfoProvider(() => Module),
                                                                          new FileAssociatedIconProvider());
+            _rememberFileContextMenuController = RememberFileContextMenuController.Default;
```
mstv, line 68: Can it be assigned in the definition above?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -51,6 +51,7 @@ public RevisionDiffControl()
             _findFilePredicateProvider = new FindFilePredicateProvider();
             _gitRevisionTester = new GitRevisionTester(_fullPathResolver);
             _revisionDiffContextMenuController = new FileStatusListContextMenuController();
+            _rememberFileContextMenuController = RememberFileContextMenuController.Default;
```
mstv, line 54: Can it be assigned in the definition above?

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -581,6 +584,18 @@ private void FileTreeContextMenu_Opening(object sender, System.ComponentModel.Ca
             openWithDifftoolToolStripMenuItem.Visible = isFile;
             openWithToolStripMenuItem.Visible = isFile;
             openWithToolStripMenuItem.Enabled = isExistingFileOrDirectory;
+            var fsi = _rememberFileContextMenuController.getFileStatusItem(gitItem.Name, _revision);
```
mstv, line 587: Else does not work for files from subfolders. More places?
```suggestion
            var fsi = _rememberFileContextMenuController.getFileStatusItem(gitItem.FileName, _revision);
```

GitUI/CommandsDialogs/RememberFileContextMenuController.cs
gerhardol: All methods in this class is static, but @RussKie prefers them to be instanced.
Will change the name though (As you wrote Create first, probably be that.)

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: ok

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: Name and FileName is assigned the same name when creating the GitItem, then the FileName is adjusted, so I missed that...

DiffTool has the same problem (another snippet I looked at). Will make a separate commit for that

RussKie:
I'm getting concerned with the size of our context menus.

GitUI/CommandsDialogs/RememberFileContextMenuController.cs
mstv: So label it with the `[Pure]` [attribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.contracts.pureattribute?view=netcore-3.1), please.

gerhardol:
> I'm getting concerned with the size of our context menus.

RevFileTree is not that long, should not be a concern.

mstv:
Another edge case only if the `Working directory` is selected:

grafik

I tried to get a diff with meaningful file names where, in addition, the current files can be edited.

gerhardol:
> Another edge case only if the Working directory is selected:

Intentional. If you select a commit, then work tree, Git can compare.
If you select worktree, then commit then no. Disabling the entry for all two selections where diff is not possible.
(Actually, worktree->worktree is possible but a separate case so just ignored.)

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
mstv: Does not work yet because path names from the FileTree contain backslashes, which git does not understand. Nothing happens (but the exit code 128) e.g. with
`git difftool --gui --no-prompt -M -C "9fbdaff9940063109e0824380dec3279973656b6:GitUI\CommandsDialogs\RevisionFileTreeControl.cs" "1f135d7c04783b1446ce0c52b29138015650ac2b:GitUI/MessageBoxes.cs"`

gerhardol:
Handling posix paths. When @mstv has approved I will squash (probably keep difftool in folder separate) and rebase, then merge

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: updated

GitUI/CommandsDialogs/RememberFileContextMenuController.cs
```diff
@@ -90,7 +90,10 @@ public string GetGitCommit([CanBeNull] Func<string, ObjectId, ObjectId> getFileB
                 return null;
             }
 
-            var name = !isSecondRevision && !string.IsNullOrWhiteSpace(item.Item.OldName) ? item.Item.OldName : item.Item.Name;
+            var name = (!isSecondRevision && !string.IsNullOrWhiteSpace(item.Item.OldName)
+                    ? item.Item.OldName
+                    : item.Item.Name)
+                .ToPosixPath();
```
mstv, line 96: A test has failed:
```suggestion
                ?.ToPosixPath();
```

gerhardol: thanks

gerhardol:
@msftbot merge in 24 hours

PR #8277, opened 2020-06-28
Track theme colors via Telemetry

NikolayXHD:
## Proposed changes

Track theme color and other settings

## Test methodology

Manual. What did I actually check:
- Theme properties were written to `telemetry.Context.GlobalProperties`.
- DiagnosticsClient calls OnExit, which calls _client.Flush() without exceptions.

## Test environment(s)

- Windows10

----

### DIFF
--- a/GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
+++ b/GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Drawing;
+using GitExtUtils.GitUI.Theming;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace GitUI.Infrastructure.Telemetry
+{
+    internal class ThemingTelemetryInitializer : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            var properties = telemetry.Context.GlobalProperties;
+            ThemeId themeId = Theming.ThemeModule.Settings.Theme.Id;
+            properties["Theme name"] = themeId.Name;
+            properties["Theme is builtin"] = themeId.IsBuiltin.ToString();
+            properties["Theme use system visual style"] = Theming.ThemeModule.Settings.UseSystemVisualStyle.ToString();
+
+            if (Theming.ThemeModule.Settings.Theme != Theme.Default)
+            {
+                foreach ((AppColor appColor, Color color) in Theming.ThemeModule.Settings.Theme.AppColorValues)
+                {
+                    properties[$"Theme application color {appColor}"] = Format(color);
+                }
+
+                foreach ((KnownColor sysColor, Color color) in Theming.ThemeModule.Settings.Theme.SysColorValues)
+                {
+                    properties[$"Theme system color {sysColor}"] = Format(color);
+                }
+            }
+
+            return;
+
+            static string Format(Color c) =>
+                c.ToArgb().ToString("x8");
+        }
+    }
+}
--- a/GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
+++ b/GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
@@ -1,6 +1,4 @@
-﻿using System;
-using System.Drawing;
-using GitExtUtils.GitUI.Theming;
+﻿using GitExtUtils.GitUI.Theming;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.Extensibility;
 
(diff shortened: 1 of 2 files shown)


GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            var properties = telemetry.Context.GlobalProperties;
+            ThemeId themeId = Theming.ThemeModule.Settings.Theme.Id;
+            properties["Theme name"] = themeId.Name;
+            properties["Theme is builtin"] = themeId.IsBuiltin.ToString();
+            properties["Theme use system visual style"] = Theming.ThemeModule.Settings.UseSystemVisualStyle.ToString();
+
+            if (Theming.ThemeModule.Settings.Theme != Theme.Default)
+            {
+                foreach ((AppColor appColor, Color color) in Theming.ThemeModule.Settings.Theme.AppColorValues)
```
gerhardol, line 21: Do all colors need to be tracked?
This could be identifying users?
Is the dark theme included?

RussKie: Agree with Gerhard, I don't think it is necessary to capture such detailed information.
The biggest issue with this approach I see is telemetry payload bloat.

GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    internal class ThemingTelemetryInitializer : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            var properties = telemetry.Context.GlobalProperties;
+            ThemeId themeId = Theming.ThemeModule.Settings.Theme.Id;
+            properties["Theme name"] = themeId.Name;
+            properties["Theme is builtin"] = themeId.IsBuiltin.ToString();
+            properties["Theme use system visual style"] = Theming.ThemeModule.Settings.UseSystemVisualStyle.ToString();
```
RussKie, line 17: I think for now we should just capture:
- light or dark
- built-in or custom

```suggestion
            properties["Theme"] = ...; //  bit: 0 (light) / 1 (dark)
            properties["Theme builtin"] = themeId.IsBuiltin ? 1 : 0;
```

What does this `Theming.ThemeModule.Settings.UseSystemVisualStyle` convey?

NikolayXHD: > What does this Theming.ThemeModule.Settings.UseSystemVisualStyle convey?

Before I discovered EasyHook-related approach, the initial approach was to override `SystemColors` and `SystemBrushes` via reflection and get all resiting controls obey modified colors using ad-hoc hacks: https://github.com/gitextensions/gitextensions/blob/master/GitExtUtils/GitUI/Theming/ThemeFix.cs

Since `ThemeFix` is only needed for custom colors, I introduced the flag UseSystemVisualStyle.

Most logic from `ThemeFix` was moved to https://github.com/gitextensions/gitextensions/tree/master/GitUI/Theming/Renderers, something remains though, therefore the flag remains as well.

GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
NikolayXHD: ok, removed tracking individual colors

GitUI/Infrastructure/Telemetry/ThemingTelemetryInitializer.cs
RussKie: > efore I discovered EasyHook-related approach, the initial approach was to override `SystemColors` and `SystemBrushes` via reflection and get all resisting controls obey modified colors using ad-hoc hacks: [/GitExtUtils/GitUI/Theming/ThemeFix.cs@`master`](https://github.com/gitextensions/gitextensions/blob/master/GitExtUtils/GitUI/Theming/ThemeFix.cs?rgh-link-date=2020-06-29T14%3A07%3A11Z)

We may have to fall back to this in .NET Core, since it doesn't support remoting required by EasyHook 🤔

RussKie: I updated the suggestion

NikolayXHD: > light or dark

- It is rather system-defined or dark
- Why limit ourselves to darkness bit instead of theme name? What if someone comes up with variations such as Dark(blue), should that person be also concerned with updating telemetry-related code?

> We may have to fall back to this in .NET Core, since it doesn't support remoting required by EasyHook 🤔

That would mean a significant drop in quality. I don't think GitExtension deserves a sloppy-looking UI, maybe better do not support dark theme at all then.

gerhardol: Stop supporting dark theme will more or less block me from contributing as my wife get irritated by light screens. So I hope the EasyHook changes come soon...

NikolayXHD: I should research for a workaround to continue using EasyHook, without waiting for EasyHook author. From what he said, it is doable, so there is hope.

RussKie: Awesome, the .NET Core conversion branch is generally usable, so have a look there.

RussKie: > * Why limit ourselves to darkness bit instead of theme name? What if someone comes up with variations such as Dark(blue), should that person be also concerned with updating telemetry-related code?

Looks like I missed this comment.
We don't want to accidentally leak user's personal data (e.g. "Nikolay's theme" may be considered as such). GDPR already makes collecting telemetry is major headache. If we don't collect user personal data, we are not subject to GDPR.
Right now I'm generally interested in knowing what proportion of our users uses light or dark theme, ours or custom defined.
On top of that, I am concerned with the payload size the telemetry uses too. We don't want to sent larger strings for something can be represented as bits or digits.

Hope this makes sense.

NikolayXHD: Thanks for explanation, got it, updated code.

RussKie:
Thank you

RussKie:
Bah should have squashed... 😒

Issue #7574, opened 2019-12-29
[NBug] The system cannot find the file specified


en3on:
:warning: The sections below must be filled in and this text must be removed or the issue will be closed.

## Current behaviour
Upon opening PuTTY Authentication Agent screen, I receive an error.
Additionally, I was receiving errors about PuTTY not being correctly configured

## Expected behaviour
Open PuTTY Authentication Agent screen

## Steps to reproduce
After updating Git Extensions, I navigated to Tools > PuTTY > Start Authentication Agent

## Error Details
```
System.ComponentModel.Win32Exception (0x80004005): The system cannot find the file specified
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at GitCommands.Executable.ProcessWrapper..ctor(String fileName, String arguments, String workDir, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding)
   at GitCommands.Executable.Start(ArgumentString arguments, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding)
   at GitUI.CommandsDialogs.FormBrowse.StartAuthenticationAgentToolStripMenuItemClick(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

## Additional information
Updated to 3.3.0.7719 this morning

## Environment
- Git Extensions 3.3.0.7719
[...]

Issue #6527, opened 2019-05-05
Generic handling of Git exit codes, errors and warnings


mstv:
## Current behaviour

The exit code of Git as well as errors and warnings printed to the console by Git are not checked by every GE function.

## Expected behaviour

Git command errors are not silently ignored but handled correctly.
Git warnings and errors are displayed clearly, not only as an unhandled exception.

`GitCommand` shall check the Git exit code and throw a `GitFailException` unless the calling function indicates that it will handle the exit code itself. The `Exception.Message` shall contain all Git error messages ("error: ...", "fatal: ...") and all Git warning messages ("warning: ...").

A `GitErrorException` shall be thrown with all Git error and warning messages as the `Exception.Message`. These exceptions shall be caught by the calling function or form if applicable and as a fallback by `GitModuleForm`/`GitModuleControl`, which shall display the error message in a MessageBox.

`GitCommand` shall check for Git warning messages and display them in a MessageBox unless the calling function indicates that it will handle them itself.
Though I am not sure whether a MessageBox is really feasible in async and background tasks. If not, a `GitWarningException` should be thrown with all Git warning messages as the `Exception.Message`. The drawback of this exception would be that all functions which don't handle the warnings themselves would be canceled.

Git errors and warnings shall be logged in the `GitCommandLog`.

## Environment

- Git Extensions 3.1.0
- Build 3d5da22221203dbe14bdb53ff779799c1d8c5f09
- Git 2.21.0.windows.1
- Microsoft Windows NT 6.1.7601 Service Pack 1
- .NET Framework 4.7.2117.0
- DPI 96dpi (no scaling)

RussKie:
Any thoughts on how this would look like?

mstv:
With the display of git errors, #7607 wouldn't have failed silently.
After a crash of my PC, `Reset all changes` failed silently; but at least `Reset file or directory changes` displayed the good hint below.
I'd prefer a MessageBox with the buttons `Ignore` and `Cancel`. On `Cancel`, the `GitException` shall be (re)thrown.
![grafik](https://user-images.githubusercontent.com/36601201/72060505-665c4500-32d4-11ea-9377-3db9c94ed1ec.png)

gerhardol:
A little more info in #8735

PR #8278, opened 2020-06-28
Wrap and popup exceptions from Executable

### LINKED ISSUE #7795: Generic handling of failure to start a process
## Current behaviour

#7714 wraps `Process.Start` in `try { } catch (Exception ex) { MessageBox.Show() }`.

## Expected behaviour

We should replace all instances of `Process` with a generic wrapper -- at least all calls to `Process.Start`.
Related to #6527.
Abstraction for `MessageBox.Show` should be used: https://github.com/gitextensions/gitextensions/issues/7751#issuecomment-589867209.

mstv:
Fixes #7795

## Proposed changes

- wrap any exception from starting an external process into the new `ExternalOperationException`,
- show any `ExternalOperationException` to the user when it is created in order to avoid silent errors eaten by too generic `catch`
- let the user decide whether the `ExternalOperationException` shall be reported as a bug anyways if unhandled
- adapt the process creation and the exception handling in the `ScriptRunner`
- refactor and extend `MessageBoxes`

### Before

grafik

### After

grafik

grafik

## Test methodology 

- add NUnit tests
- manual with User Scripts

## Test environment(s) 
- Git Extensions 33.33.33
- Build f33a5b45015f282bbc024605917b72a37972b015
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -15,23 +15,35 @@ namespace GitCommands
     /// <inheritdoc />
     public sealed class Executable : IExecutable
     {
-        private readonly string _workingDir;
+        private readonly string _context;
+        private readonly ExternalOperationExceptionFactory.Handling _exceptionHandling;
         private readonly Func<string> _fileNameProvider;
+        private readonly string _workingDir;
 
         public Executable([NotNull] string fileName, [NotNull] string workingDir = "")
             : this(() => fileName, workingDir)
         {
         }
 
-        public Executable([NotNull] Func<string> fileNameProvider, [NotNull] string workingDir = "")
+        public Executable([NotNull] Func<string> fileNameProvider,
+            [NotNull] string workingDir = "",
+            ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
+            [CanBeNull] string context = null)
         {
+            if (string.IsNullOrWhiteSpace(workingDir))
+            {
+                workingDir = Directory.GetCurrentDirectory();
+            }
+
             _workingDir = workingDir;
             _fileNameProvider = fileNameProvider;
+            _exceptionHandling = exceptionHandling;
+            _context = context;
         }
 
         /// <inheritdoc />
         [PermissionSet(SecurityAction.Demand, Name = "FullTrust")]
-        public IProcess Start(ArgumentString arguments = default, bool createWindow = false, bool redirectInput = false, bool redirectOutput = false, Encoding outputEncoding = null)
+        public IProcess Start(ArgumentString arguments = default, bool createWindow = false, bool redirectInput = false, bool redirectOutput = false, Encoding outputEncoding = null, bool useShellExecute = false)
         {
             // TODO should we set these on the child process only?
             EnvironmentConfiguration.SetEnvironmentVariables();
@@ -40,12 +52,19 @@ namespace GitCommands
 
             var fileName = _fileNameProvider();
 
-            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding);
+            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute, _exceptionHandling, _context);
         }
 
         public string GetOutput(ArgumentString arguments)
         {
-            return this.GetOutput(arguments, null);
+            try
+            {
+                return this.GetOutput(arguments, input: null);
+            }
+            catch (Exception ex) when (!(ex is ExternalOperationException))
+            {
+                throw ExternalOperationExceptionFactory.Default.Create(_context, command: $"{_fileNameProvider()} {arguments}", _workingDir, inner: ex, _exceptionHandling);
+            }
         }
 
         #region ProcessWrapper
@@ -69,38 +88,62 @@ namespace GitCommands
             private bool _disposed;
 
             [SuppressMessage("ReSharper", "AssignNullToNotNullAttribute")]
-            public ProcessWrapper(string fileName, string arguments, string workDir, bool createWindow, bool redirectInput, bool redirectOutput, [CanBeNull] Encoding outputEncoding)
+            public ProcessWrapper(string fileName,
+                                  string arguments,
+                                  string workDir,
+                                  bool createWindow,
+                                  bool redirectInput,
+                                  bool redirectOutput,
+                                  [CanBeNull] Encoding outputEncoding,
+                                  bool useShellExecute,
+                                  ExternalOperationExceptionFactory.Handling exceptionHandling,
+                                  [CanBeNull] string context)
             {
-                Debug.Assert(redirectOutput == (outputEncoding != null), "redirectOutput == (outputEncoding != null)");
-                _redirectInput = redirectInput;
-                _redirectOutput = redirectOutput;
-
-                _process = new Process
+                try
                 {
-                    EnableRaisingEvents = true,
-                    StartInfo =
+                    Debug.Assert(redirectOutput == (outputEncoding != null), "redirectOutput == (outputEncoding != null)");
+                    _redirectInput = redirectInput;
+                    _redirectOutput = redirectOutput;
+
+                    _process = new Process
                     {
-                        UseShellExecute = false,
-                        ErrorDialog = false,
-                        CreateNoWindow = !createWindow,
-                        RedirectStandardInput = redirectInput,
-                        RedirectStandardOutput = redirectOutput,
-                        RedirectStandardError = redirectOutput,
-                        StandardOutputEncoding = outputEncoding,
-                        StandardErrorEncoding = outputEncoding,
-                        FileName = fileName,
-                        Arguments = arguments,
-                        WorkingDirectory = workDir
-                    }
-                };
+                        EnableRaisingEvents = true,
+                        StartInfo =
+                        {
+                            UseShellExecute = useShellExecute,
+                            ErrorDialog = false,
+                            CreateNoWindow = !createWindow,
+                            RedirectStandardInput = redirectInput,
+                            RedirectStandardOutput = redirectOutput,
+                            RedirectStandardError = redirectOutput,
+                            StandardOutputEncoding = outputEncoding,
+                            StandardErrorEncoding = outputEncoding,
+                            FileName = fileName,
+                            Arguments = arguments,
+                            WorkingDirectory = workDir
+                        }
+                    };
 
-                _logOperation = CommandLog.LogProcessStart(fileName, arguments, workDir);
+                    _logOperation = CommandLog.LogProcessStart(fileName, arguments, workDir);
 
-                _process.Exited += OnProcessExit;
+                    _process.Exited += OnProcessExit;
 
-                _process.Start();
+                    _process.Start();
 
-                _logOperation.SetProcessId(_process.Id);
+                    _logOperation.SetProcessId(_process.Id);
+                }
+                catch (Exception ex)
+                {
+                    _logOperation.LogProcessEnd(ex);
+
+                    Dispose();
+
+                    throw ExternalOperationExceptionFactory.Default.Create(context,
+                        command: $"{_process.StartInfo.FileName} {_process.StartInfo.Arguments}",
+                        _process.StartInfo.WorkingDirectory,
+                        inner: ex,
+                        exceptionHandling);
+                }
             }
 
             private void OnProcessExit(object sender, EventArgs eventArgs)
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -20,11 +20,6 @@ namespace GitCommands
         private readonly Func<string> _fileNameProvider;
         private readonly string _workingDir;
 
-        public Executable([NotNull] string fileName, [NotNull] string workingDir = "")
-            : this(() => fileName, workingDir)
-        {
-        }
-
         public Executable([NotNull] Func<string> fileNameProvider,
             [NotNull] string workingDir = "",
             ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
--- a/GitCommands/Git/ExecutableFactory.cs
+++ b/GitCommands/Git/ExecutableFactory.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using GitUIPluginInterfaces;
+using JetBrains.Annotations;
+
+namespace GitCommands
+{
+    public class ExecutableFactory
+    {
+        /// <summary>
+        /// Singleton accessor.
+        /// </summary>
+        public static ExecutableFactory Default { get; } = new ExecutableFactory();
+
+        public IExecutable Create([NotNull] Func<string> fileNameProvider,
+            [NotNull] string workingDir = "",
+            ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
+            [CanBeNull] string context = null)
+            => new Executable(fileNameProvider, workingDir, exceptionHandling, context);
+
+        public IExecutable Create([NotNull] string fileName,
+            [NotNull] string workingDir = "",
+            ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
+            [CanBeNull] string context = null)
+            => Create(() => fileName, workingDir, exceptionHandling, context);
+    }
+}
--- a/GitCommands/Git/ExecutableFactory.cs
+++ b/GitCommands/Git/ExecutableFactory.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Windows.Forms;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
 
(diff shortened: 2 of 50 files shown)


mstv:
This PR is marked as WIP/draft mainly because all uses of `Executable` and `Process.Start` must be changed to use the new `ExecutableFactory` yet.

So: Is the generic approach OK -- disregarding `ScriptRunner` (see below)?

The screenshots show an edge case: In case of the `ScriptRunner`, the `ExternalOperationException` rather should not popup early (with the misleading info / question regarding the bug report). Instead, it should be translated in order to show it with the context information, i.e. a single popup with:

```
Failed to execute script "test".
Reason: Failed to execute "cmdx".
Working directory: "F:\Build\gitextensions3_dev\"
Reason: The system cannot find the file specified
```

How to translate `ExternalOperationException`s without referencing UI code?

RussKie:
👀 but looks like a lot of code... 🤔

RussKie:
> * refactor and extend `MessageBoxes`

Could you please explain the reasoning for this change? 
Also it seems unrelated to the executable changes.

GitCommands/Git/ExecutableException.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class ExecutableException : ExternalOperationException
+    {
+        public ExecutableException(ProcessStartInfo processStartInfo, [NotNull] Exception inner)
+            : base(inner)
+        {
+            ProcessStartInfo = processStartInfo;
+            if (string.IsNullOrEmpty(ProcessStartInfo.WorkingDirectory))
+            {
+                ProcessStartInfo.WorkingDirectory = Directory.GetCurrentDirectory();
+            }
```
RussKie, line 17: Why does the exception do this? The exception is merely a carrier of error information, it mustn't modify information it carries.

GitCommands/Git/Executable.cs
```diff
@@ -98,7 +128,15 @@ public ProcessWrapper(string fileName, string arguments, string workDir, bool cr
...
                 _process.Exited += OnProcessExit;
 
-                _process.Start();
+                try
+                {
+                    _process.Start();
+                }
+                catch (Exception ex)
+                {
+                    _logOperation.LogProcessEnd(null);
```
RussKie, line 137: What do you think about logging the exception here, so it can be viewed in the command log?
E.g.
```suggestion
                    _logOperation.LogProcessEnd(null, ex);
```

GitCommands/Git/Executable.cs
```diff
@@ -31,21 +32,50 @@ public Executable([NotNull] Func<string> fileNameProvider, [NotNull] string work
...
 
-            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding);
+                    return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute);
+                },
+                () => new ProcessStartInfo
+                {
+                    FileName = fileName,
+                    Arguments = args,
+                    WorkingDirectory = _workingDir
+                });
```
RussKie, line 57: Why do we need all this plumbing? The exception is caught and then handled in `ProcessWrapper`?

mstv: > Why do we need all this plumbing? The exception is caught and then handled in `ProcessWrapper`?

The exception is not consumed. It is wrapped into `ExecutableException`.

One could consider to replace the arguments `fileName, args, _workingDir` with a prefilled `ProcessStartInfo`.
Though the extension method `GetOutput` below would have to be adapted, too.

GitCommands/Git/Executable.cs
mstv: > What do you think about logging the exception here, so it can be viewed in the command log?

👍 I will extend the classes `ProcessOperation` and CommandLogEntry.

GitCommands/Git/ExecutableException.cs
mstv: > Why does the exception do this? The exception is merely a carrier of error information, it mustn't modify information it carries.

I agree in general. But in this case, it adds missing information.
Alternatively, `Executable`/`ProcessWrapper` could specify the `WorkingDirectory` always.

mstv:
I have completed this -- including proper handling for `ScriptRunner`.

RussKie:
I'll need some time for the review. There is a lot of new plumbing which makes me concerned and I need to some time to think through it.

RussKie:
👀

RussKie:
Thank you @mstv, I've reviewed the code but I can't stop thinking this adds a lot of ("enterprise level") complexity, and that we can achieve the desired with significantly less plumbing.
I've drafted a simpler alternative that I think should achieve the same but with less code. We will need to review an number of call sites that one way or another invoke external operations (e.g. git.exe, explorer.exe, etc.), but we'd have to review them with this change as well.

I'm happy to setup a call with you to talk through, if you like.

mstv:
I have removed `ExecutableException` and added `Dispose` to the `ProcessWrapper` ctor.

I agree that the TaskDialog will improve the MessageBox though with little more pressure against NBug reports as in #8358.

mstv:
Best reviewed commit by commit.

RussKie:
Thank you @mstv, but the static factories make me very uncomfortable. We are not using any other instances other than "Default", and this makes the factories look superficial.
That is, I can't see what we're gaining with such change:
```diff
-            _gitExecutable = new Executable(() => AppSettings.GitCommand, WorkingDir);
+            _gitExecutable = ExecutableFactory.Default.Create(() => AppSettings.GitCommand, WorkingDir);
```

The changes in MessageBoxes.cs feel like we're overcomplicating the implementation yet we're probably not delivering a good UX - i.e. if I read the code correctly there may be isntances where we're just showing an error with a "Ignore and continue" button. The value of such dialog is fairly low - if the user will ignore and continue anyway, why bother showing the dialog?

. . .

The changes in https://github.com/gitextensions/gitextensions/pull/8278/commits/2720f2fc4baa57e7160ad55d616705cb33d04040 modulo changes to OsShellUtil.cs are worth extracting and submitting as a separate PR, they are an improvement, and will make this work more targeted.

mstv:
Let's focus on *what* is to be changed first and _then_ think about the implementation and _how_ the SRP can be met.

> **Current behaviour**
#7714 wraps `Process.Start` in `try { } catch (Exception ex) { MessageBox.Show() }`.
**Expected behaviour**
We should replace all instances of `Process` with a generic wrapper -- at least all calls to `Process.Start`, e.g. with `GitCommands.Executable.ProcessWrapper`, which also needs generic error handling: https://github.com/gitextensions/gitextensions/issues/7574#issuecomment-588985889. 
Related to #6527.

> - wrap **any exception** from starting an **external** process into the new `ExternalOperationException`,
> - show any `ExternalOperationException` to the user __when it is created__ in order to __avoid silent errors__ eaten by too generic `catch`

Any external operation can fail. Usually, this is not caused by GE itself, i.e. it is no bug.
That's why I want to display the initial exception with context information to the user (unless the calling function is using exceptions for its normal operation, e.g. detection of the git executable).
This shall be done by one central wrapper, which is not overruled by too generic `try { } catch { // ignore }`. Silent errors are evil.

Running user scripts or shells is a special case (which I should not have chosen as the first example for the screenshots).
If the external processes fail this aint a bug of GE. That's why the error is shown to the user without the option to report a bug.

> we're just showing an error with a "Ignore and continue" button. The value of such dialog is fairly low - if the user will ignore and continue anyway, why bother showing the dialog?

Of course, we do not continue in the sense of just swallowing the exception but in "continuing to run GE".
So it is rather an "abort the current operation". You have chosen this wording instead of my "Do you want to cancel the operation and ignore the error?". I overtook it without rethinking.

> The changes in MessageBoxes.cs feel like we're overcomplicating the implementation

What do you think is unnecessary? Exceptions can be thrown in the background. So we must switch to the UI thread in order to display the TD. (In addition, `Form.ActiveForm` is `null` if the application is not the active application, e.g. if halt in a break point. I'd say this does not make much sense to me.)

> static factories make me very uncomfortable. We are not using any other instances other than "Default".

Last time, you complained about static classes. How to deal with `ExternalOperationExceptionFactory.OnException`? Just tell me. I think we need this indirection, else it is getting spaghetti.

GitCommands/Git/ExternalOperationExceptionFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+            Show,
+            None
+        }
+
+        /// <summary>
+        /// Singleton accessor.
+        /// </summary>
+        public static ExternalOperationExceptionFactory Default { get; } = new ExternalOperationExceptionFactory();
+
+        public event Action<ExternalOperationException, Handling> OnException;
```
ivangrek, line 19: Why do we need this event?

GitCommands/Git/ExternalOperationExceptionFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public ExternalOperationException Create(string context, string command, string workingDirectory, Exception inner, Handling handling = Handling.OptionalBugReport)
+        {
+            ExternalOperationException ex = new ExternalOperationException(context, command, workingDirectory, inner);
+            OnException?.Invoke(ex, handling);
+            return ex;
```
ivangrek, line 25: Could we put this event in an exception class?
I see that creating an exception does not guarantee it will be thrown. At first glance, we might not be using a factory.

GitCommands/Git/ExecutableFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+            [NotNull] string workingDir = "",
+            ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
+            [CanBeNull] string context = null)
+            => new Executable(fileNameProvider, workingDir, exceptionHandling, context);
+
+        public IExecutable Create([NotNull] string fileName,
+            [NotNull] string workingDir = "",
+            ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
+            [CanBeNull] string context = null)
+            => Create(() => fileName, workingDir, exceptionHandling, context);
```
ivangrek, line 25: I only see one place where this is used. 
Could you explain why we need these two methods and not one?

`Func<string> fileNameProvider` 

This is dependency, but it's not clear why.

GitCommands/Git/ExternalOperationException.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace GitCommands
+{
+    /// <summary>
+    /// Exception class for failures of external operations,
+    /// e.g. file access or start of other executables, which always can fail due to removed or locked files.
+    /// </summary>
+    public class ExternalOperationException : Exception
```
ivangrek, line 9: Could we put this class in GitCommands.Exceptions or some other non-primary namespace?

GitUI/MessageBoxes.cs
```diff
@@ -124,13 +134,116 @@ public static bool ConfirmUpdateSubmodules(IWin32Window win)
         public static void ShellNotFound([CanBeNull] IWin32Window owner)
             => ShowError(owner, Instance._shellNotFound.Text, Instance._shellNotFoundCaption.Text);
 
+        public static void Show([CanBeNull] IWin32Window owner, ExternalOperationException ex, ExternalOperationExceptionFactory.Handling exceptionHandling)
+        {
+            if (exceptionHandling == ExternalOperationExceptionFactory.Handling.None)
+            {
+                return;
+            }
+
```
ivangrek, line 137: Why do we need to use "IWin32Window owner".
Most likely this question applies to all methods in this class, but this one in particular. 
Are there any cases where the current window will not own the message box?

RussKie: IIRC there are cases where owner is a different window.

GitUI/OsShellUtil.cs
```diff
@@ -1,29 +1,57 @@
...
+                // ignore because already shown to the user
+            }
         }
 
         public static void OpenWithFileExplorer(string filePath)
         {
-            Process.Start("explorer.exe", filePath);
+            try
+            {
+                ExecutableFactory.Default.Create("explorer.exe", exceptionHandling: ExternalOperationExceptionFactory.Handling.Show).Start(filePath);
```
ivangrek, line 49: Could we use a constant for the "explorer.exe" line?

GitUI/OsShellUtil.cs
```diff
@@ -33,7 +61,14 @@ public static void OpenUrlInDefaultBrowser(string url)
         {
             if (!string.IsNullOrWhiteSpace(url))
             {
-                Process.Start(url);
```
ivangrek, line 36: Could we use a Systerm.Uri for url parameter?

GitCommands/Git/ExecutableFactory.cs
```diff
@@ -22,5 +23,8 @@ public class ExecutableFactory
             ExternalOperationExceptionFactory.Handling exceptionHandling = ExternalOperationExceptionFactory.Handling.OptionalBugReport,
             [CanBeNull] string context = null)
             => Create(() => fileName, workingDir, exceptionHandling, context);
+
+        public IProcess Spawn(string arguments, [NotNull] string workingDir = "")
+            => Create(Application.ExecutablePath, workingDir, ExternalOperationExceptionFactory.Handling.OptionalBugReport).Start(arguments);
```
ivangrek, line 28: Why ExecutableFactory creates IProcess if IExecutable does?

GitCommands/Git/GitModule.cs
```diff
@@ -51,7 +51,7 @@ public GitModule([CanBeNull] string workingDir)
             WorkingDirGitDir = GitDirectoryResolverInstance.Resolve(WorkingDir);
             _indexLockManager = new IndexLockManager(this);
             _commitDataManager = new CommitDataManager(() => this);
-            _gitExecutable = new Executable(() => AppSettings.GitCommand, WorkingDir);
+            _gitExecutable = ExecutableFactory.Default.Create(() => AppSettings.GitCommand, WorkingDir);
```
ivangrek, line 54: Why do we need to use () => AppSettings.GitCommand not AppSettings.GitCommand?

GitUI/MessageBoxes.cs
ivangrek: Do you know specific cases?

RussKie:
> Let's focus on _what_ is to be changed first and _then_ think about the implementation and _how_ the SRP can be met.
> 
> Any external operation can fail. Usually, this is not caused by GE itself, i.e. it is no bug.
> That's why I want to display the initial exception with context information to the user (unless the calling function is using exceptions for its normal operation, e.g. detection of the git executable).
> This shall be done by one central wrapper, which is not overruled by too generic `try { } catch { // ignore }`. Silent errors are evil.

👍  


> Running user scripts or shells is a special case (which I should not have chosen as the first example for the screenshots).
> If the external processes fail this aint a bug of GE. That's why the error is shown to the user without the option to report a bug.

👍 


> > we're just showing an error with a "Ignore and continue" button. The value of such dialog is fairly low - if the user will ignore and continue anyway, why bother showing the dialog?
> 
> Of course, we do not continue in the sense of just swallowing the exception but in "continuing to run GE".
> So it is rather an "abort the current operation". You have chosen this wording instead of my "Do you want to cancel the operation and ignore the error?". I overtook it without rethinking.

This makes sense, now when you put it like this 👍 

I'm going to ignore the other two paragraphs and questions for the moment, and try to distill the requirements.

_In event when there is an unhandled exception (including `try... catch { /* do nothing */ }` unless documented why the exception is swallowed)_
_We want to display a meaningful user error including ability to report the issue to GitHub_
_Unless the error is caused by executing a user configurable functionality (such as scripts or shells) - in this case no GitHub reporting is provided_.

Did I get it correctly?

Currently all unhandled exceptions are trapped via `ReportBug(Exception)` method, that spawn the NBug dialog, if necessary.
Could this be the place to implement the logic that checks and show the error? E.g. something like this:
```cs
private static void ReportBug(Exception ex)
{
    if (ex is ExternalOperationException externalException)
    {
        // Something happened either in the app - can submit the bug to GitHub
        ReportAppException(externalException);
        return;
    }

    // Likely a user error
    ReportUserException(ex);
}
```

This way the implementation be simpler, in places where we know we can expect a failure we can `catch (Exception ex) { throw new ExternalOperationException(ex); }`, and it will fall through to `ReportBug` above.
With this I don't think we'd need factories or changes to messageboxes, as we'd implement a messagebox/taskdialog in `ReportAppException`.

Thoughts?

RussKie:
I have updated #8358 with the above proposal. I tried few different scenarios, like git.exe is missing, user script failed, a random failure... Seemed to achieve the desired.

image
image

mstv:
> In event when there is an unhandled exception (including try... catch { /* do nothing */ } unless documented why the exception is swallowed)
We want to display a meaningful user error including ability to report the issue to GitHub
Unless the error is caused by executing a user configurable functionality (such as scripts or shells) - in this case no GitHub reporting is provided.

> Did I get it correctly?

Yes. But in contrast to your approach, I do comply to the rule "Never trust your clients!", i.e. in this case, I do not trust the GE implementation, where I have seen too many swallowed errors.

RussKie:
We shouldn't trust input data, but we should trust implementations we own ;)
We can/need to clean unnecessary catches.

RussKie:
Closing this for now, since this has been largely affected by https://github.com/gitextensions/gitextensions/pull/8358.

PR #8295, opened 2020-07-05
Refactor MessageBoxes

mstv:
## Proposed changes

- Factor out a common function `Show` to call `MessageBox.Show` in a single place in `MessageBoxes`
  in order to prepare for easy replacement / extension, e.g. for
  - mock-up in NUnit tests
  - switching to the UI thread in order to show errors of external operations when they occur
- Factor out a common function `Confirm`
- Remove unreferenced `_confirmDeleteRemoteBranch` and `ConfirmDeleteRemoteBranch`

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build b6a6047
- Git 2.27.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/MessageBoxes.cs
+++ b/GitUI/MessageBoxes.cs
@@ -37,7 +37,6 @@ namespace GitUI
         private readonly TranslationString _theRepositorySubmodules = new TranslationString("Update submodules on checkout?");
         private readonly TranslationString _updateSubmodulesToo = new TranslationString("Since this repository has submodules, it's necessary to update them on every checkout.\r\n\r\nThis will just checkout on the submodule the commit determined by the superproject.");
         private readonly TranslationString _rememberChoice = new TranslationString("Remember choice");
-        private readonly TranslationString _confirmDeleteRemoteBranch = new TranslationString("Do you want to delete the branch {0} from {1}?");
 
         private readonly TranslationString _reason = new TranslationString("Reason");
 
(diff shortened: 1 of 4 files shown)


GitUI/MessageBoxes.cs
```diff
@@ -56,6 +55,9 @@ internal MessageBoxes()
 
         private static MessageBoxes Instance => instance ?? (instance = new MessageBoxes());
 
+        public static bool Confirm([CanBeNull] IWin32Window owner, string text, string caption)
+            => Show(owner, text, caption, MessageBoxButtons.YesNo, MessageBoxIcon.Question) == DialogResult.Yes;
+
```
RussKie, line 60: I'd very much prefer to remove this, update calling methods signatures and and update all call sites to verify against `DialogResult`. We already have too many convenience methods with insignificant value-add.
https://github.com/gitextensions/gitextensions/pull/8303 was inspired by all the same reasons.

I'd want a close look at each of these methods here, most of these methods have only a single call site, meaning we generate additional layers of code and consume cpu cycles for questionable reasons.

GitUI/MessageBoxes.cs
```diff
@@ -68,27 +70,21 @@ public static void NotValidGitDirectory([CanBeNull] IWin32Window owner)
...
-                MessageBoxButtons.OK, MessageBoxIcon.Warning);
-        }
+            => Show(owner,
+                    string.Format(ResourceManager.Strings.GeneralGitConfigExceptionMessage,
+                                  exception.ConfigPath, Environment.NewLine, (exception.InnerException ?? exception).Message),
+                    ResourceManager.Strings.GeneralGitConfigExceptionCaption,
+                    MessageBoxButtons.OK,
+                    MessageBoxIcon.Warning);
 
         public static bool UnresolvedMergeConflicts(IWin32Window owner)
```
RussKie, line 80: ...linked to the above:
```suggestion
        public static DialogResult ConfirmResolveMergeConflicts(IWin32Window owner)
            => Show(owner, text, caption, MessageBoxButtons.YesNo, MessageBoxIcon.Question)
```

GitUI/MessageBoxes.cs
mstv: > I'd very much prefer to remove this, update calling methods signatures and and update all call sites to verify against DialogResult. We already have too many convenience methods with insignificant value-add.
#8303 was inspired by all the same reasons.

It is not only about the comparison with `DialogResult`. All share the same common arguments. This is where the DRY principle *begins* to become applicable for me.
This is different from #8303, (which IMO bloats all callsites with arguments which are required for one or two special cases. I agree that there were too many overloaded ctors.)

Though we could declare `Confirm` `private` since there are no other users ATM.
Or remove it if you insist.

> I'd want a close look at each of these methods here, most of these methods have only a single call site, meaning we generate additional layers of code and consume cpu cycles for questionable reasons.

Don't concern yourself about this:
1. The compiler and linker can recognize and optimize this.
2. We are showing a MessageBox.

GitUI/MessageBoxes.cs
mstv: This will make the interface of `MessageBoxes` depend more on that of `MessageBox`.
We might want to replace `MessageBox` with `TaskDialog` in the future.

GitUI/MessageBoxes.cs
RussKie: > which IMO bloats all callsites with arguments which are required for one or two special cases.

Yes, it is a concern I share. But we have to "bloat" first to be able to see what and how API are used before we can architect a better solution.

GitUI/MessageBoxes.cs
RussKie: > We might want to replace `MessageBox` with `TaskDialog` in the future.

❤️ 
That's on my ever growing backlog. The .NET 5.0 TaskDialog is test-enabled, i.e. a dialog can be closed programmatically despite the static nature of invocation - the dialog can be access via a bound page.

PR #8296, opened 2020-07-06
Disable Finish button while loading branches in GitFlow form.

bstecisa:
## Proposed changes
Disable Finish button in GitFlow form while branches are loading. Before this fix you can press Finish button while branches ComboBox displays 'Loading...', so you can finish 'Loading...' branch which doen't exists.


RussKie:
Thank you

PR #8311, opened 2020-07-12
Clean up of status reporting of failed operations

RussKie:
## Proposed changes

Undo plumbing and infra introduced in 1572cbd7da7af53678f05b8cd4dcfcdf9e8e46d3. With that we have one less dependency on WPF stack.

There are several scenarios that involve showing `FormStatus` or its descendants.
1. Show a progress as a modal dialog (this is standard way).
2. Do not show the progress dialog, and only show it if an operation that was running failed (e.g. staging files in `FormCommit`).

The approach to showing results of failed operations appears to be unnecessary complex, that IIUC led to issues while attempting to show a dialog that was already considered opened but invisible.
And the WPF's DispatcherFrameModalController was used to deal with this. 
This also necessitated defining delegates, and required more plumbing to pass those around, validation, etc.

However it is much simpler *not* to show the dialog *until after* the operation has failed. The callsites are also much easier to read and understand.

Also some general code clean up.

⚠️ NB: best to review commit by commit. Don't squash.

## Test methodology 

Lots of manual tests 
- for normal operations (e.g. running `git fetch` and `git fetch -all`)
- for no-dialog operation (e.g. file staging) that were set to fail (e.g. convert EOL from `crlf` to `lf` in one of the existing files in a repo)

### DIFF
--- a/GitUI/HelperDialogs/FormStatus.cs
+++ b/GitUI/HelperDialogs/FormStatus.cs
@@ -55,9 +55,9 @@ namespace GitUI.HelperDialogs
             InitializeComplete();
         }
 
-        protected readonly ConsoleOutputControl ConsoleOutput;
-        public Action<FormStatus> ProcessCallback;
-        public Action<FormStatus> AbortCallback;
+        private protected ConsoleOutputControl ConsoleOutput { get; }
+        private protected Action<FormStatus> ProcessCallback;
+        private protected Action<FormStatus> AbortCallback;
         private bool _errorOccurred;
 
         /// <summary>
@@ -65,7 +65,7 @@ namespace GitUI.HelperDialogs
         /// For instance, progress messages might be skipped; other messages might be added manually.
         /// </summary>
         [NotNull]
-        public readonly FormStatusOutputLog OutputLog = new FormStatusOutputLog();
+        private protected FormStatusOutputLog OutputLog { get; } = new FormStatusOutputLog();
 
         protected override CreateParams CreateParams
         {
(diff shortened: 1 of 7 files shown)


RussKie:
@msftbot merge in 2 days

GitUI/HelperDialogs/FormStatus.cs
```diff
@@ -178,48 +182,27 @@ public void Reset()
...
-            Start();
         }
 
-        private void FormStatus_FormClosed(object sender, FormClosedEventArgs e)
-        {
-            AfterClosed();
-        }
-
-        internal void Start()
+        private void Start()
```
mstv, line 205: `private` methods are not sorted.

GitUI/HelperDialogs/FormStatus.cs
```diff
@@ -55,101 +59,110 @@ public FormStatus(GitUICommands commands, [CanBeNull] ConsoleOutputControl conso
...
+                return mdiCp;
+            }
         }
 
-        protected readonly ConsoleOutputControl ConsoleOutput;
-        public Action<FormStatus> ProcessCallback;
-        public Action<FormStatus> AbortCallback;
-        private bool _errorOccurred;
-        private bool _showOnError;
+        private protected ConsoleOutputControl ConsoleOutput { get; }
```
mstv, line 72: Properties between methods?

RussKie:
Thank you. There's more clean up to do, I'll address the nits in the follow up PRs.

PR #8323, opened 2020-07-14
Dark silver theme

bstecisa:
## Proposed changes

- Create a new 'darksilver' theme, based on existing dark theme, a bit lighter

## Screenshots 
imagen


GitUI/GitUI.csproj
```diff
@@ -72,6 +72,10 @@
     <None Remove="Resources\Icons\artwork\**" />
     <None Remove="Resources\Icons\originals\**" />
   </ItemGroup>
+
```
pmiossec, line 75: Is it needed? I don't see it for other themes css files...

GitUI/GitUI.csproj
```diff
@@ -86,6 +90,10 @@
     <Content Include="Themes\dark.css">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
pmiossec, line 91: Can we instead replace all the `<Content>`s by only one `<Content Include="Themes\*.css">`?

/cc @RussKie @NikolayXHD

RussKie: Yes, we can

bstecisa:
Ok, I updated the colors for the scrollbar and now they are a tad brighter.

Normal view:
imagen

On mouse hover view:
imagen

bstecisa:
Sorry, i forgot to include the proposed changes for the csproj file.

pmiossec:
@bstecisa FYI, when using GitHub, it's a bad practice to create a pull request from `master` branch. So in the future, please create a feature branch.

Also, avoid syncing with `master` by doing merges. Just add commits to your branch. If you really need to sync, prefer to do a `rebase`. A last one can be done (with a squash if requested) just before the pull request merge to clean the history.

In your case, the history is now a little bit messy so, as @gerhardol requested it, could you squash and rebase now (to only have one commit on top of upstream `master`)?

RussKie:
Thank you @bstecisa.

Gerhard/Philippe merge at will, if you're happy.

gerhardol:
@msftbot merge in 24 hours

pmiossec:
Looks mostly good for me also.

I have just one thing that I prefer in the "dark" theme. If we could discuss about that @bstecisa and @gerhardol ?

dark silver (left) / dark (right)
image

I find that the color of the text in the commit info makes it more readable in the dark theme. White text makes it hard to read.
@bstecisa is that **really** intended? Can it be improved?

gerhardol:
@bstecisa I see that you have pushed updates to the theme already. Is that what you want to merge instead?

NikolayXHD:
> I find that the color of the text in the commit info makes it more readable in the dark theme. White text makes it hard to read.
> @bstecisa is that **really** intended? Can it be improved?

If you are referring to the blue / white color for links, the switch from blue to white happens outside .NET WinForms in RichTextBox implementation, depending on the background color, i believe it is `SystemColors.Window`.

pmiossec:
@NikolayXHD No, I was talking of the text that was "too white" for me.

@bstecisa I liked the change you did in the feature branch `feature/darksilver_theme_update`. Unfortunately, a branch name can't be changed once a PR is done 😢 So you will need to update your `master` branch a last time before we could merge this PR...

pmiossec:
@NikolayXHD The new style:

image

RussKie:
The links are non legible

RussKie:
Blocking the merge.

I'd like the blue link text to be made readable, right now it isn't, which is an accessibility issue. 
It doesn't have to be light (white/silver), but must have a better contrast ratio that it is. The general guide the contrast ratio has to be 3:1.

RussKie:
Once this is addressed I'm happy to take it into 3.4.3

bstecisa:
Hi, I didn't make a new pr because i'm trying to improve the blue color. It seems it changes by code when Window color reaches certain black hue.

NikolayXHD:
> Hi, I didn't make a new pr because i'm trying to improve the blue color. It seems it changes by code when Window color reaches certain black hue.

@bstecisa, @RussKie that's right, `RichTextBox` implementation does that, it's outside of our control. Choosing `SystemColors.Window` is the only known way we can affect it.

RussKie:
Just to confirm I understand it correctly - we can't change the blue link color in the dark mode, right?

bstecisa:
We can't change the blue link color in any mode as the OS api is who set that style. If you set the background color at some hue of dark or some hue of blue, the link color changes, but you can't change these limits neither. With the dark theme you can see the links in other color because the background color is really dark, but it doesn't work if you set the background just a bit more graish. I tried a few workarounds with no luck. You can detect the links in a RichBoxText using Regex and set the color you want manually, but you lose the open link behaviour.

RussKie:
Thank you

bstecisa:
Oh wait, there are some changes in the theme proposed by @gerhardol, I have these changes in a branch, do i make a new pr?

RussKie:
Yes, please do. @gerhardol approved this PR, so I took it as a signed it was done.

gerhardol:
I requested changes for rebase/squash and the scollbar, the rest is tweaking andfor me to have comments, I will likely need to use it some time in different setups. 
I just asked if there were already a new branch, if those changes should have been merged instead (but there were some open points).

RussKie:
Btw, the setup must be modified as well for the new theme to be packaged.

bstecisa:
@gerhardol my mistake, actually @pmiossec proposed that changes. I'm going to make a new pr.

PR #8333, opened 2020-07-17
Darksilver theme update.

bstecisa:
Minor changes in darksilver theme [(#8323)](https://github.com/gitextensions/gitextensions/pull/8323)

### DIFF
--- a/GitUI/Themes/darksilver.css
+++ b/GitUI/Themes/darksilver.css
@@ -22,8 +22,8 @@
 .MenuText { color: #cecece }
 .ScrollBar { color: #606163 }
 .Window { color: #292929 }
-.WindowFrame { color: #a0b9e7 }
-.WindowText { color: #e9e9e9 }
+.WindowFrame { color: #cdd1da }
+.WindowText { color: #bdbdbd }
 .GradientActiveCaption { color: #4e5255 }
 .GradientInactiveCaption { color: #292929 }
 .MenuBar { color: #2d3133 }
--- a/Setup/Product.wxs
+++ b/Setup/Product.wxs
@@ -592,6 +592,9 @@
       <Component Id="dark.css" Guid="*">
         <File Source="$(var.ArtifactsPublishPath)\Themes\dark.css" />
       </Component>
+      <Component Id="darksilver.css" Guid="*">
+        <File Source="$(var.ArtifactsPublishPath)\Themes\darksilver.css" />
+      </Component>
     </DirectoryRef>
     <DirectoryRef Id="PluginsDir">
       <Component Id="DeleteUnusedBranches.dll" Guid="*">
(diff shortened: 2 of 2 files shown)


pmiossec:
I have rebased the branch so that it is in good shape....

gerhardol:
Setup/Product.wxs must be updated for the installer
Commit comment should be updated

pmiossec:
Done. (Setup not tested)

RussKie:
.\build -publish

RussKie:
@bstecisa please build and verify the setup bundle. Once it is confirmed we'll take it.

bstecisa:
Ok, i'm following the instructions detailed here:

https://github.com/gitextensions/gitextensions/wiki/How-To:-build-instructions

But when I execute the 'build.cmd -pack' command I get the following exception:

MSBUILD : error MSB1001: Unknown switch.
Switch: -pack

I'm doing something wrong?.

RussKie:
It is `-publish`:
https://github.com/gitextensions/gitextensions/blob/00e5dea96261f697101888004262d1e3c470af1e/scripts/build.ps1#L13

I corrected the wiki, thank you

bstecisa:
The 'build.cmd -publish' command gives me 22 errors. I attach the output file.

[publish_log.txt](https://github.com/gitextensions/gitextensions/files/4946609/publish_log.txt)

pmiossec:
That is due to #7850 where plugins are renamed due prevent plugin loading slowness but the `Product.wsx` file has not been updated accordingly :(

@RussKie

The commit was done months ago and is contained in the v3.4 branch to fix one problem raised by the use of Arcade build...
So that does raise a lot of questions......

1. Why do we have a different behavior between the local way to create the setup and the one used for the real created setup?

I imagine that it's because we don't use the Arcade build to release....

2. Why the does the build process that put the plugins in there own subdirectory and the fix in #7850 are not done? That have as consequences that the Setup and the portable version don't have the expected format.... and that explain why the `Product.wsx` is still working.

I have made the fix for `Product.wsx` locally but I don't know what I must do with that............

pmiossec:
I have pushed the fix for the local publish but we have to understand the real problem before merging it....

RussKie:
Ah, it came back to me, thank you @pmiossec for re-jigging my memory. 788d29c is unfortunately is incorrect, and must be rolled back.

Since https://github.com/gitextensions/gitextensions/pull/7850 local builds name plugins with "GitExtensions." prefix to speed up the local dev. Production builds don't do that, hence the setup definitions look for the prefix-less files.
The build pipeline also runs a different script file (cibuild.cmd vs build.cmd) that passes `-ci` flag to the build.

Once https://github.com/gitextensions/gitextensions/pull/7162 is comeplete https://github.com/gitextensions/gitextensions/pull/7850 can be undone and building locally will again be closer to building on a build server.

So to build a setup bundle locally you need to run the same commands as we run on AppVeyor, i.e.
```
.\cibuild.cmd -restore -build -buildNative -publish
```

pmiossec:
Ok. commit removed. The ci command seems to work well.

@bstecisa Could you test with this command and see if the setup is working well?

bstecisa:
Tests passed and cibuild command works well.

PR #8334, opened 2020-07-17
SubmoduleTests: Setup repos once for all tests

### LINKED ISSUE #8279: `SubmoduleStatusProviderTests` extremely slow
## Current behaviour

Running tests under VS I've observed `SubmoduleStatusProviderTests` tests to take in excess of 7 minutes to complete.
image

## Expected behaviour

Tests must complete faster.

gerhardol:
Fixes #8279

## Proposed changes

Setup the Git repos used in the tests only once, to speed up tests.
For me, test time goes down from 1,6 mon to 11s, about twice the time for one test. (Longer time for other users.) 
- Setup attributes for Parallelizable to disallow the individual tests in parallel

More comments in the code

I consider this to be a major hack.

## Test methodology

Tests only

## Test environment(s) 

- Git Extensions 33.33.33
- Build da7659f46eed24c0d730898d8c6d0cf33d46abb2
- Git 2.27.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
+++ b/UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
@@ -10,12 +10,15 @@ using GitCommands.Git;
 using GitCommands.Submodules;
 using GitUIPluginInterfaces;
 using NUnit.Framework;
+using NUnit.Framework.Constraints;
 
 namespace GitCommandsTests.Submodules
 {
     [SetCulture("en-US")]
     [SetUICulture("en-US")]
     [TestFixture]
+
+    [Parallelizable]
     internal class SubmoduleStatusProviderTests
     {
         private GitModuleTestHelper _repo1;
@@ -30,10 +33,23 @@ namespace GitCommandsTests.Submodules
         private GitModule _repo3Module;
 
         private ISubmoduleStatusProvider _provider;
+        private bool _isInit = false;
 
+        /// <summary>
+        /// The repo creation should only be done once as it slows tests down considerably
+        /// Each test must therefore revert changes after the test
+        /// OneTimeSetUp cannot be used, as the setup must run after BeforeTest in CommonTestUtils\ConfigureJoinableTaskFactoryAttribute.cs
+        /// OneTimeTearDown can be used though
+        /// </summary>
         [SetUp]
         public void SetUp()
         {
+            if (_isInit)
+            {
+                return;
+            }
+
+            _isInit = true;
             _repo1 = new GitModuleTestHelper("repo1");
             _repo2 = new GitModuleTestHelper("repo2");
             _repo3 = new GitModuleTestHelper("repo3");
@@ -49,8 +65,8 @@ namespace GitCommandsTests.Submodules
             _provider = new SubmoduleStatusProvider();
         }
 
-        [TearDown]
-        public void TearDown()
+        [OneTimeTearDown]
+        public void OneTimeTearDown()
         {
             _provider.Dispose();
             _repo1.Dispose();
@@ -58,6 +74,7 @@ namespace GitCommandsTests.Submodules
             _repo3.Dispose();
         }
 
+        [NonParallelizable]
         [Test]
         public async Task UpdateSubmoduleStructure_valid_result_for_top_module()
         {
@@ -72,6 +89,7 @@ namespace GitCommandsTests.Submodules
             result.OurSubmodules.Should().BeEquivalentTo(result.AllSubmodules);
         }
 
+        [NonParallelizable]
         [Test]
         public async Task UpdateSubmoduleStructure_valid_result_for_first_nested_submodule()
         {
@@ -87,6 +105,7 @@ namespace GitCommandsTests.Submodules
             result.OurSubmodules.Select(info => info.Path).Should().ContainSingle(_repo3Module.WorkingDir);
         }
 
+        [NonParallelizable]
         [Test]
         public async Task UpdateSubmoduleStructure_valid_result_for_second_nested_submodule()
         {
@@ -101,6 +120,7 @@ namespace GitCommandsTests.Submodules
             result.OurSubmodules.Select(info => info.Path).Should().BeEmpty();
         }
 
+        [NonParallelizable]
         [Test]
         public async Task Submodule_status_changes_for_top_module_with_first_nested_module_change()
         {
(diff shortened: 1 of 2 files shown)


mstv:
> `[OneTimeSetUp]` cannot be used, as the setup must run after `BeforeTest` in `CommonTestUtils\ConfigureJoinableTaskFactoryAttribute.cs`

We should keep this in mind for all tests. CC: @sharwell

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -792,14 +744,24 @@ public async Task Submodule_status_changes_for_second_nested_module_with_first_n
...
+
+        /// <summary>
+        /// Check that the repo is reverted after the test, prepared for next
+        /// An explicit Git clean and reset will require additional time
+        /// </summary>
+        /// <param name="result">The existing structure, reused from the test</param>
+        /// <returns>a Task</returns>
+        private async Task CheckRevertedStatus(SubmoduleInfoResult result)
+        {
+            var changedFiles = GetStatusChangedFiles(_repo1Module);
```
mstv, line 760: Is `_repo1Module` really correct for all testcases, e.g. `Submodule_status_changes_for_first_nested_module_with_second_nested_module_changes`?

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -130,14 +150,10 @@ public async Task Submodule_status_changes_for_top_module_with_first_nested_modu
 
             // Revert the change
             File.Delete(Path.Combine(_repo2Module.WorkingDir, "test.txt"));
-            changedFiles = GetStatusChangedFiles(currentModule);
-            changedFiles.Should().HaveCount(0);
-            await SubmoduleTestHelpers.UpdateSubmoduleStatusAndWaitForResultAsync(_provider, currentModule, changedFiles);
-            result.AllSubmodules.All(i => i.Detailed == null).Should().BeTrue();
-            result.OurSubmodules.Should().BeEquivalentTo(result.AllSubmodules);
```
mstv, line 137: Why has this been removed nearly all testcases?

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -483,17 +467,10 @@ public async Task Submodule_status_changes_for_first_nested_module_with_second_n
...
             // Revert the change
             File.Delete(Path.Combine(_repo3Module.WorkingDir, "test.txt"));
-            changedFiles = GetStatusChangedFiles(currentModule);
-            changedFiles.Should().HaveCount(0);
-            await SubmoduleTestHelpers.UpdateSubmoduleStatusAndWaitForResultAsync(_provider, currentModule, changedFiles);
-            result.AllSubmodules.All(i => i.Detailed == null).Should().BeTrue();
-            result.OurSubmodules[0].Should().BeEquivalentTo(result.AllSubmodules[1]);
-
-            // Revert manually - not changed back automatically
-            result.TopProject.Detailed.IsDirty.Should().BeTrue();
```
mstv, line 493: This has assertion has been removed without replacement. Applies to further testcases below.

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
gerhardol: Yes, here we want to check from the top module that nothing is changed

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
gerhardol: Because CheckRevertedStatus() is called
There is no check that the cleanup was partially done as it was before

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
gerhardol: result.OurSubmodules.Should().BeEquivalentTo(result.AllSubmodules);
andsimilar is testedwhen result iscreated, no need to test it again

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
mstv: I mean just the line `result.OurSubmodules.Should().BeEquivalentTo(result.AllSubmodules);` which is not part of `CheckRevertedStatus`.

gerhardol: See below, tested separately

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -30,10 +33,23 @@ internal class SubmoduleStatusProviderTests
...
         [SetUp]
         public void SetUp()
         {
+            if (_isInit)
+            {
+                return;
+            }
+
+            _isInit = true;
             _repo1 = new GitModuleTestHelper("repo1");
```
RussKie, line 53: Tests shouldn't concern themselves with resetting the state after execution. I was thinking of a change along the lines of `ReferenceRepoistory`:
```cs
            if (_repo1 is null)
            {
                _repo1 = new GitModuleTestHelper("repo1");
            }
            else
            {
                _repo1.Reset();
            }
```

Do you think we can re-write these tests to utilise `ReferenceRepository` instead?

gerhardol: The repos would need not just to reset, but also undo commits. That would require that the initial commit have to be remembered, which will make a more complex setup. Also, a reset will increase the test time slightly.
The test cases cleaning up is much easier.

RussKie:
Much better now:
image

RussKie:
However on my work machine tests fail to stage and I need 649552ab5c8a86bd90ff33ec64301e41231840cc.

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
gerhardol: A clean is needed too, in addition to reset and checkout.

gerhardol:
> Much better now:

The tests requiring a second basically runs UpdateSubmoduleStructureAndWaitForResultAsync() for top module (a series of commands parsing .gitmodules creating GitModules recursively) then a couple of git-status from top module and one or two other commands. 

Stepping in one test case, one time: Submodule_status_changes_for_top_module_with_first_nested_module_commit_second_nested_module_change
1.1s without setup in debug, 550 ms normally
Repo setup: 6.6 s
UpdateSubmoduleStructureAndWaitForResultAsync: 30 ms
// No changes in repo
First status 114ms
UpdateSubmoduleStatusAndWaitForResultAsync 8 ms
// Make a change in repo3
Commit 8ms
Status 175 ms
UpdateSubmoduleStatusAndWaitForResultAsync  75 ms
// Revert the change for repo3
Status 331ms
UpdateSubmoduleStatusAndWaitForResultAsync 78 ms
// Revert the change
checkout 64 ms
CheckRevertedStatus 76 ms (git-status)

Git status is cached, the time cannot be taken out of context 

However, the check that a repo is clean before and after adds 200 ms for this test. Slightly less for some, more for others.
Those could be excluded by default. Just excluding CheckRevertedStatus reduces execution time from 12,3 s to 10,9 s for me.
Harder to find why a test fails though.

Unless there are requests to optimize, I plan to merge in a day or so.

RussKie:
Thank you. This is an improvement, let's merge it.

PR #8341, opened 2020-07-20
Constrain child windows to monitor bounds

RussKie:
If a parent window is at a corner of a desktop, and it is smaller than a child window, it is possible to end up in a situation where the child window is positioned outside of visible and reachable area.
Coupled with the fact that locations and dimensions of most windows get persisted this may lead to a situation where the window may become permanently inaccessible until the settings are reset.

## Proposed changes

- Constrain the window to be within the desktop bounds

### Before

Steps to replicate:

1. Open the app at the top left corner of the desktop
2. Resize it to be smaller than a typical "Setting" dialog windows size
3. Open "Settings" dialog
4. Observe the dialog opened at negative coordinates, something like:
    image

## Test methodology 

- Manual

❗ **NB:** I have a single monitor and unable to test multi-monitor setup, where a secondary monitor is placed to the left or to the bottom of the main monitor. It is possible that the code need to be tweaked further to account for these cases.

----

### DIFF
--- a/GitUI/GitExtensionsForm.cs
+++ b/GitUI/GitExtensionsForm.cs
@@ -109,7 +109,7 @@ namespace GitUI
 
             _needsPositionRestore = false;
 
-            var workingArea = _getScreensWorkingArea();
+            IReadOnlyList<Rectangle> workingArea = _getScreensWorkingArea();
             if (!workingArea.Any(screen => screen.IntersectsWith(position.Rect)))
             {
                 if (position.State == FormWindowState.Maximized)
@@ -133,21 +133,17 @@ namespace GitUI
 
             if (Owner is null || !windowCentred)
             {
-                var location = DpiUtil.Scale(position.Rect.Location, originalDpi: position.DeviceDpi);
+                Point calculatedLocation = DpiUtil.Scale(position.Rect.Location, originalDpi: position.DeviceDpi);
 
-                if (WindowPositionManager.FindWindowScreen(location, workingArea) is Rectangle rect)
-                {
-                    location.Y = rect.Y;
-                }
-
-                DesktopLocation = location;
+                DesktopLocation = WindowPositionManager.FitWindowOnScreen(new Rectangle(calculatedLocation, Size), workingArea);
             }
             else
             {
                 // Calculate location for modal form with parent
-                Location = new Point(
+                Point calculatedLocation = new Point(
                     Owner.Left + (Owner.Width / 2) - (Width / 2),
                     Owner.Top + (Owner.Height / 2) - (Height / 2));
+                Location = WindowPositionManager.FitWindowOnScreen(new Rectangle(calculatedLocation, Size), workingArea);
             }
 
             if (WindowState != position.State)
--- a/GitUI/WindowPositionManager.cs
+++ b/GitUI/WindowPositionManager.cs
@@ -1,10 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Drawing;
-using System.Linq;
 using System.Windows.Forms;
 using GitExtUtils.GitUI;
-using JetBrains.Annotations;
 
 namespace GitUI
 {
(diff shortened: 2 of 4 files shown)


GitUI/GitExtensionsForm.cs
```diff
@@ -132,14 +132,16 @@ protected virtual void RestorePosition()
...
                 DesktopLocation = location;
             }
             else
             {
                 // Calculate location for modal form with parent
                 Location = new Point(
-                    Owner.Left + (Owner.Width / 2) - (Width / 2),
-                    Owner.Top + (Owner.Height / 2) - (Height / 2));
+                    Math.Max(0, Owner.Left + (Owner.Width / 2) - (Width / 2)),
+                    Math.Max(0, Owner.Top + (Owner.Height / 2) - (Height / 2)));
```
mstv, line 144: Unfortunately, this will not work. Windows screen coordinates _can_ be negative, e.g. if you arrange a second display left of the primary display.
I guess we must iterate all displays and check against their display areas (excluding the taskbar).

RussKie: Thank you. Yes, I realised that, I just don't have a 2nd monitor to test. I should get one soon, and I'd be able to resume this work.

mstv: There is quite a lot of code checking `Screen.WorkingArea`s in `GitExtensionsForm` and `WindowPositionManager` already.
It looks sensible. Though it is done before the DPI-scaling.
Often, the union of the `WorkingAreas` will be no rectangle.
A pragmatic approach could be to constrain the Form to the `WorkingArea` which contains the center or the maximum part of the Form.

RussKie:
I'm sure there's an API in Windows Forms or Win32 to know whether a window
is within the visible desktop area.
Once I have a 2nd monitor I'll be able to explore this.

RussKie:
Incidentally I received my 2nd monitor today :D

RussKie:
I've spent an evening playing in a multi monitor setup, and here's what I propose.

1. Iterate over user's monitors, and test three points of a given form: (L, T), ((L - R)/2, T), (R, T) to see whether any of the user's monitors contain any these points. This way we can find out which monitors a given form is present.
2. If a point is displayed on a monitor test whether the following rectangle is fully displayed on this monitor: { p.X, p.Y, monitor.Width * 10%, monitor.Height * 10% } (inverted for the (R, T) point, and test each way for the middle point). 
This way we know that if the form isn't fully visible, it is visible sufficiently enough for the user to notice it, and be able to move, if necessary.
NB: 10% is an arbitrary choice, can be more.
3. If none of the three points and their offsets pass the check, then we move the form to main display at (0, 0).


<details>
<summary>LinqPad test script</summary>

For the test setup I have two 4K monitors: 
* primary 32" @ 100% below  {X=0,Y=0,Width=3840,Height=2120}
* secondary 27" @ 150% above {X=0,Y=-2160,Width=2467,Height=1440}

NB: the script doesn't currently take dpi into account

```cs
void Main()
{
	var screens = Screen.AllScreens.Select(screen => screen.WorkingArea).ToArray();

	int count = 0;
	foreach (Rectangle screen in screens)
	{
		Console.WriteLine($"Screen {++count}: {screen.ToString()}");
	}
	Console.WriteLine();

	Rectangle[] rects = new System.Drawing.Rectangle[]
	{
		// top left primary monitor
		new Rectangle(0, 0, 500, 500),
		// incorrect location
		new Rectangle(-100, -100, 500, 500),
		// creeps into the 2ndary screen
		new Rectangle(0, -730, 3840, 750),
		// shown on the 2ndary screen 
		new Rectangle(-100, -1730, 750, 750),
	};

	foreach (Rectangle rect in rects)
	{
		Console.WriteLine($"Testing rectangle: {rect.ToString()}\r\n---------------");
		var form = new Form();
		form.StartPosition = FormStartPosition.Manual;
		form.DesktopLocation = rect.Location;
		form.Size = rect.Size;
		form.Text = rect.ToString();
		form.Show();

		foreach (Rectangle screen in screens)
		{
			bool res = IsDisplayedOn10Percent(screen, rect, form);
			if (res)
			{
				Console.WriteLine($"Screen: {screen.ToString()}");
				form.BackColor = Color.Green;

				// no point testing other screens				
				Console.WriteLine();
				break;
			}
			else
			{
			}
		}

		if (form.BackColor != Color.Green)
		{
			form.BackColor = Color.Orange;
			form.DesktopLocation = Point.Empty;
		}

		Console.WriteLine();
	}
}


private bool IsDisplayedOn10Percent(Rectangle screen, Rectangle window, Form debugForm)
{
	if (screen.IsEmpty || window.IsEmpty) return false;

	int requiredHeight = (int)(screen.Height * 0.1);
	int requireWidth = (int)(screen.Width * 0.1);

	var label = new Label
	{
		AutoSize = false,
		Size = new Size(requireWidth, requiredHeight),
		BackColor = Color.Blue
	};

	Point p;
	if (screen.Contains(window.Location))
	{
		p = new Point(window.Left + requireWidth, window.Top + requiredHeight);
		bool leftTop = screen.Contains(p);
		if (leftTop)
		{
			Console.WriteLine($"{screen.ToString()} contains {p} (L, T)");
			label.Location = new Point(0, 0);
			debugForm.Controls.Add(label);
			return true;
		}
	}

	if (screen.Contains(new Point(window.Width / 2, window.Top)))
	{
		p = new Point(window.Width / 2 - requireWidth, window.Top + requiredHeight);
		bool middleTop = screen.Contains(p);
		if (middleTop)
		{
			Console.WriteLine($"{screen.ToString()} contains {p} (W/2-, T)");
			label.Location = new Point(p.X, 0);
			debugForm.Controls.Add(label);
			return true;
		}

		p = new Point(window.Width / 2 + requireWidth, window.Top + requiredHeight);
		middleTop = screen.Contains(p);
		if (middleTop)
		{
			Console.WriteLine($"{screen.ToString()} contains {p} (W/2+, T)");
			label.Location = new Point(window.Width / 2, 0);
			debugForm.Controls.Add(label);
			return true;
		}
	}

	if (screen.Contains(new Point(window.Right, window.Top)))
	{
		p = new Point(window.Right - requireWidth, window.Top + requiredHeight);
		bool rightTop = screen.Contains(p);
		if (rightTop)
		{
			Console.WriteLine($"{screen.ToString()} contains {p} (R, T)");
			label.Location = new Point(window.Right - requireWidth, 0);
			debugForm.Controls.Add(label);
			return true;
		}
	}

	return false;
}
```

</details>

mstv:
👍 Sounds like a very good plan. I agree the title bar is most important, and left/right of the middle but not the exact middle covers the case where a window filled two screens and one of them was disconnected.
I think the 10% should be enough. Though they must be clipped to the - for ((L - R)/2, T) remaining - width and height of the window.

RussKie:
The primary screen 4K at 100% (bottom), the secondary screen 4K at 150% (top)

* ✔️  `new Rectangle(0, 0, 500, 500)` - valid, the window is fully shown on a single monitor
image
* :x: `new Rectangle(-100, -100, 500, 500)` - invalid as vertically is visible less than 10% on the screen that has the upper border of the windows caption bar
image
* ❌ `new Rectangle(0, -730, 3840, 750)` - invalid   as vertically is visible less than 10% on the screen that has the upper border of the windows caption bar
image
* ✔️  `new Rectangle(-100, -1730, 750, 750)` - valid, visible 10% worth of the screen size in each direction starting from the middle point
image

RussKie:
* `new Rectangle(-350, -1730, 750, 750)` - valid, enough space to the right of the middle
image

* `new Rectangle(-500, -1730, 750, 750)` - valid, enough visible space from the right side
image

PR #8357, opened 2020-07-26
fix test: SaveImpl_should_throw_if_invalid_path

RussKie:
Fixes #

## Proposed changes

- Don't show a message box when unable to save settings to a file.
Instead raise a `SaveSettingsException` that gets handled at the UI layer, and displayed in a task dialog.
- Remove FormSettings.Save check for git
    - If git is missing the process will crash, so the check is not adding any value, but consumes cycles.
    - If git is present, running `Module.GitExecutable.GetOutput("")` is for the most part pointless.

### After

* image

## Test methodology 

- manual
- enabled unit tests

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitCommands/Settings/FileSettingsCache.cs
+++ b/GitCommands/Settings/FileSettingsCache.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
-using System.Windows.Forms;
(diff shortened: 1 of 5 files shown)


RussKie:
@msftbot merge in 36 hours

Issue #6220, opened 2019-02-06
Git Extension window pops on top, but without focus, on completion of task


rmc47:
## Current behaviour

1. Begin some long-running task in GitExtensions (e.g. pull from remote)
2. Switch focus to some other application which covers the GitExtensions window
3. On completion of the task, the GitExtensions window will pop to the foreground, but focus remains on the other application.

## Expected behaviour

GitExtensions should not pop to the foreground on completion of a task unless user input is required (for example, if it needs to prompt for a rebase / merge).

## Steps to reproduce

See current behaviour and screenshot below.

## Screenshots

![gitextfocus](https://user-images.githubusercontent.com/809741/52361664-a7016d80-2a36-11e9-968a-1834fc2131d2.gif)

## Did this work in previous version of GitExtensions

Pretty sure this is new behaviour since GitExtensions v3.

## Environment

- GitExtensions version: v3.0.1.4933
- GIT version:  2.16.2.windows.1
- OS version: Windows 10, build 1809 (OS build 17763.253)
- .NET version:  4.7.3260.0

vbjay:
No. It is not new. Read up on conemu issues.

rmc47:
Maybe I was previously on such an old version it was still in classic mode before the upgrade to v3!

My workaround for anyone finding this bug in the future: Settings -> Advanced -> uncheck "Use Console Emulator for console output in command dialogs". Focus then remains where it should.

@vbjay I'm not sure if you consider this a dupe of e.g. #3570 or whether you'd prefer it to continue open in its own right?

vbjay:
Don't autoclose the dialog and you will see what the issue really is.

https://github.com/gitextensions/gitextensions/search?q=is%3Aissue+conemu&type=Issues

RussKie:
@rmc47 yes, it is an annoying side effect of the Console Emulator (conemu) we use.
I think one way to mitigate the issue is to disable it (_Settings > Advanced > Use ConEmu_).

mstv:
Still after the update of ConEmu: When a rebase finishes, the GE window is moved behind other applications.

gerhardol:
> Still after the update of ConEmu: When a rebase finishes, the GE window is moved behind other applications.

I do not see any changes (neither good or bad) myself after updating ConEmu. But the ConEmu author usually requests updates when raising issues.

I see some issues with pop over/under also without ConEmu, but likely not exactly the same.

RussKie:
[...]

PR #8358, opened 2020-07-26
Better exception presentation

### LINKED ISSUE #7795: Generic handling of failure to start a process

RussKie:
Contributes to #7795

## Proposed changes

Trap and route as many exceptions as possible to the central handler.

For exceptions coming from the main thread it is possible to ignore those and continue the app's execution, but for unhandled exceptions coming from background threads the app will terminate.
The main focus of this work is to intercept exceptions originated from `Executable` (i.e. executing git commands, etc.), and user scripts. However it is not limited in this scope.

- An alternative simplified proposal for https://github.com/gitextensions/gitextensions/pull/8278.

### Before

grafik

### After

* Settings
image

* Script
image

* An arbitrary exception (reported in https://github.com/gitextensions/gitextensions/issues/8367)
image

* An arbitrary exception (reported in https://github.com/gitextensions/gitextensions/issues/8647)
image
Hitting "ignore" will close the app anyway, since the repo is gone. I'm not sure what the expect UX would be in this situation...

### DIFF
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -98,9 +98,18 @@ namespace GitCommands
 
                 _process.Exited += OnProcessExit;
 
-                _process.Start();
+                try
+                {
+                    _process.Start();
+                    _logOperation.SetProcessId(_process.Id);
+                }
+                catch (Exception ex)
+                {
+                    Dispose();
 
-                _logOperation.SetProcessId(_process.Id);
+                    _logOperation.LogProcessEnd(ex);
+                    throw new ExternalOperationException(fileName, arguments, workDir, ex);
+                }
             }
 
             private void OnProcessExit(object sender, EventArgs eventArgs)
--- a/GitCommands/Git/ExternalOperationException.cs
+++ b/GitCommands/Git/ExternalOperationException.cs
@@ -1,0 +1,44 @@
+﻿using System;
+
+#nullable enable
+
+namespace GitCommands
+{
+    /// <summary>
+    /// Represents errors that occur during execution of an external operation,
+    /// e.g. running a git operation or launching an external process.
+    /// </summary>
+    public class ExternalOperationException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExternalOperationException"/> class with a specified parameters
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="command">The command that led to the exception.</param>
+        /// <param name="arguments">The command arguments.</param>
+        /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public ExternalOperationException(string command, string arguments, string workingDirectory, Exception innerException)
+            : base(innerException?.Message, innerException)
+        {
+            Command = command;
+            Arguments = arguments;
+            WorkingDirectory = workingDirectory;
+        }
+
+        /// <summary>
+        /// The command that led to the exception.
+        /// </summary>
+        public string Command { get; }
+
+        /// <summary>
+        /// The command arguments.
+        /// </summary>
+        public string Arguments { get; }
+
+        /// <summary>
+        /// The working directory.
+        /// </summary>
+        public string WorkingDirectory { get; }
+    }
+}
--- a/GitCommands/Git/UserExternalOperationException.cs
+++ b/GitCommands/Git/UserExternalOperationException.cs
@@ -1,6 +1,4 @@
-﻿using System;
-
-#nullable enable
+﻿#nullable enable
 
 namespace GitCommands
 {
@@ -13,13 +11,14 @@ namespace GitCommands
         /// Initializes a new instance of the <see cref="UserExternalOperationException"/> class with a specified parameters
         /// and a reference to the inner exception that is the cause of this exception.
         /// </summary>
-        /// <param name="command">The command that led to the exception.</param>
-        /// <param name="arguments">The command arguments.</param>
-        /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="context">The command that led to the exception.</param>
         /// <param name="innerException">The exception that is the cause of the current exception.</param>
-        public UserExternalOperationException(string command, string arguments, string workingDirectory, Exception innerException)
-            : base(command, arguments, workingDirectory, innerException)
+        public UserExternalOperationException(string context, ExternalOperationException innerException)
+            : base(innerException.Command, innerException.Arguments, innerException.WorkingDirectory, innerException.InnerException)
         {
+            Context = context;
         }
+
+        public string Context { get; }
     }
 }
--- a/GitCommands/Git/UserExternalOperationException.cs
+++ b/GitCommands/Git/UserExternalOperationException.cs
@@ -1,0 +1,25 @@
+﻿using System;
+
+#nullable enable
+
+namespace GitCommands
+{
+    /// <summary>
+    /// Represents errors that occur during execution of user-configured operation, e.g. a script.
+    /// </summary>
+    public class UserExternalOperationException : ExternalOperationException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UserExternalOperationException"/> class with a specified parameters
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="command">The command that led to the exception.</param>
+        /// <param name="arguments">The command arguments.</param>
+        /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public UserExternalOperationException(string command, string arguments, string workingDirectory, Exception innerException)
+            : base(command, arguments, workingDirectory, innerException)
+        {
+        }
+    }
+}
--- a/GitUI/NBugReports/BugReporter.cs
+++ b/GitUI/NBugReports/BugReporter.cs
@@ -22,7 +22,10 @@ namespace GitExtensions
             StringBuilder sb = new StringBuilder();
 
             // Command: <command>
-            sb.AppendLine($"{Strings.Command}: {ex.Command}");
+            if (!string.IsNullOrWhiteSpace(ex.Command))
+            {
+                sb.AppendLine($"{Strings.Command}: {ex.Command}");
+            }
 
             // Arguments: <args>
             if (!string.IsNullOrWhiteSpace(ex.Arguments))
@@ -113,14 +116,11 @@ namespace GitExtensions
 
         private static void ReportUserException(UserExternalOperationException ex, bool isTerminating)
         {
-            // UserExternalOperationException wraps an actual exception, but be cautious just in case
-            string instructionText = ex.InnerException?.Message ?? Strings.InstructionOperationFailed;
-
             using var dialog = new TaskDialog
             {
                 OwnerWindowHandle = OwnerFormHandle,
                 Text = FormatText(ex, canRaiseBug: false),
-                InstructionText = instructionText,
+                InstructionText = ex.Context,
                 Caption = Strings.CaptionFailedExecute,
                 Icon = TaskDialogStandardIcon.Error,
                 Cancelable = true,
--- a/GitUI/NBugReports/BugReporter.cs
+++ b/GitUI/NBugReports/BugReporter.cs
@@ -1,0 +1,152 @@
+﻿#nullable enable
+
+using System;
+using System.Text;
+using System.Windows.Forms;
+using GitCommands;
+using GitUI;
+using Microsoft.WindowsAPICodePack.Dialogs;
+
+namespace GitExtensions
+{
+    public static class BugReporter
+    {
+        private static Form? OwnerForm
+            => Form.ActiveForm ?? (Application.OpenForms.Count > 0 ? Application.OpenForms[0] : null);
+
+        private static IntPtr OwnerFormHandle
+            => OwnerForm?.Handle ?? IntPtr.Zero;
+
+        private static string FormatText(ExternalOperationException ex, bool canRaiseBug)
+        {
+            StringBuilder sb = new StringBuilder();
+
+            // Command: <command>
+            sb.AppendLine($"{Strings.Command}: {ex.Command}");
+
+            // Arguments: <args>
+            if (!string.IsNullOrWhiteSpace(ex.Arguments))
+            {
+                sb.AppendLine($"{Strings.Arguments}: {ex.Arguments}");
+            }
+
+            // Working directory: <working dir>
+            sb.AppendLine($"{Strings.WorkingDirectory}: {ex.WorkingDirectory}");
+
+            if (canRaiseBug)
+            {
+                // Directions to raise a bug
+                sb.AppendLine();
+                sb.AppendLine(Strings.ReportBug);
+            }
+
+            return sb.ToString();
+        }
+
+        public static void Report(Exception ex, bool isTerminating)
+        {
+            if (ex is UserExternalOperationException userExternalException)
+            {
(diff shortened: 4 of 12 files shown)


RussKie:
@mstv pumping up on your radar, interested in your thoughts (when you have time).

mstv:
I have not forgot this but been caught in recent discussions yet.

mstv:
I haven't reviewed in detail because basic questions / comments arose quickly:

Which commits of my PR have you taken into account?

Which of my proposed changes do you disagree on?

I already gave the reasonable for starting an exception hierarchy:
> `ExecutableException` is derived from the new `ExternalOperationException` (planned to be used for file operations etc. in a follow-up PR)

Just give a short answer and let me rethink, please.

RussKie:
> Which commits of my PR have you taken into account?
> 

I looked at your proposed changes, and certainly drawn an inspiration from those.

> Which of my proposed changes do you disagree on?

This changeset is significantly smaller, and IMO achieves the same outcome - i.e. traps failures and presents them to a user in a usable (or more useable than now) manner.
Certainly this change wasn't as exhaustive in terms of updating all `Executable` callsites, but even with those I expect the scope to still be smaller.

I don't like adding extra infra plumbing such as factories, and additional exception layers. At this stage I don't see any immediate need for those.

> `ExecutableException` is derived from the new `ExternalOperationException` (planned to be used for file operations etc. in a follow-up PR)

Let's do one step at a time please. In my experience quite often after making the first step things (like design or requirements) change, but the "future-proofed code" never gets removed.

mstv:
> > Which commits of my PR have you taken into account?

> I looked at your proposed changes, and certainly drawn an inspiration from those.

That wasn't my question. See next point.

> > Which of my proposed changes do you disagree on?

> This changeset is significantly smaller, and IMO achieves the same outcome - i.e. traps failures and presents them to a user in a usable (or more useable than now) manner.

#7795: "We should replace **all** instances of `Process` with a generic wrapper".
Maybe this was not clear enough. This means "instead of bloating all the callsites with `try { }  catch { ShowError }`.

#8278: "show **every** `ExternalOperationException` to the user **when it is created** in order to avoid silent errors eaten by too generic `catch`"

(Special cases require exceptions / extensions of this rule: finding the git executable, script execution.)

> Certainly this change wasn't as exhaustive in terms of updating all `Executable` callsites, but even with those I expect the scope to still be smaller.

I want a generic solution. Every external operation can fail (process execution or just filesystem access). Usually this is not a bug of GE. So let's ensure the user gets the exact error message and avoid NBug reports.

> I don't like adding extra infra plumbing such as factories, and additional exception layers. At this stage I don't see any immediate need for those.

See above. In addition, `ExternalOperationException.Handled` is a different aspect than the process-specific context information (SRP). Though `Command` and `WorkingDirectory` could be re-used for filesystem errors. Or else we can create an `IContext` abstraction instead of an exception hierarchy -- of course only if necessary and in a follow-up PR.

> in a usable (or more useable than now) manner.

Regarding `TaskDialog`: We agreed on waiting for .Net Core 5 before replacing `MessageBox` with TD because of API changes (remember the discussion on magic numbers, which seem to be not needed anymore?).

GitCommands/Git/Executable.cs
```diff
@@ -98,9 +98,16 @@ public ProcessWrapper(string fileName, string arguments, string workDir, bool cr
...
-
-                _logOperation.SetProcessId(_process.Id);
+                try
+                {
+                    _process.Start();
+                    _logOperation.SetProcessId(_process.Id);
+                }
+                catch (Exception ex)
+                {
+                    _logOperation.LogProcessEnd(ex);
```
wischi-chr, line 108: If an exception is thrown here the object is never initialized and thus Dispose won't happen. Disposing before rethrowing should fix that.

ivangrek:
How about using status bar notifications instead of a dialog box?

image

As far as I understand, we provide an opportunity, not a necessity. The window causes rejection, the requirement to perform an action.

RussKie:
That's an interesting idea.
It kind of crossed my mind few times.
Though I think it may create a somewhat disjoint UX. I wouldn't do it as
part of this PR, but certainly something to consider.

ivangrek:
Pay attention to this method.
Should it catch all exceptions?

```
private bool InvokeEvent([CanBeNull] IWin32Window ownerForm, [CanBeNull] EventHandler<GitUIEventArgs> gitUIEventHandler)
{
    try
    {
        var e = new GitUIEventArgs(ownerForm, this);
        gitUIEventHandler?.Invoke(this, e);
        return !e.Cancel;
    }
    catch (Exception ex)
    {
        MessageBox.Show(ex.Message, "Exception", MessageBoxButtons.OK, MessageBoxIcon.Error);
    }

    return true;
}
```

RussKie:
@mstv I've just discovered that the proposed "ignore and continue" won't work as one may expect - the app will be terminated if an exception has come from a non-UI thread, which is routed through [`AppDomain.CurrentDomain.UnhandledException`](https://docs.microsoft.com/dotnet/api/system.appdomain.unhandledexception#remarks):
image

This is easily reproducible in both your work and mine by going to Settings > Git > Paths and modifying the git path:
image
Choosing to "ignore" will kill the app.

There is no easy way out of this, we'll need to catch all exceptions occurring on non-UI thread, and marshal them to the main thread.

There's also another annoyance as well - exceptions thrown in the `AsyncLoader` aren't trappable nor they get routed to `AppDomain.CurrentDomain.UnhandledException`. I had a brief investigation but I couldn't see anything obvious.
An easy way to repro this is to modify "status" to "status 1" in `GetAllChangedFilesCmd` method, set the breakpoint as follows and open the `FormCommit`:
image
This exception will not surface in anyway, and the form will have an endless spinner...

I'm not expecting or proposing these issues to be addressed as part of this work, just wanted to raise them up.

ivangrek:
Most likely we will have to forward async to the top or wrap the method call in ThreadHelper.JoinableTaskFactory.Run.

ivangrek:
Maybe we should move ThreadHelper.JoinableTaskFactory.Run to the UI only.

RussKie:
For now in unhandled exceptions situations (those that are coming from non-UI thread) I propose we do this:
image

And we'll gradually need to clean up the exceptions and their routing.

mstv (commented):
This PR improves the behavior significantly and catches some important special cases, e.g. `IsTerminating`. 👍
But it does not solve #7795 completely. So please remove "fixes" from the PR description.

Furthermore, I do not share your opinion to trust the GE code: It has grown too much -- with the wrong attitude regarding error handling / swallowing.

I am still quite busy in RL. Though I could implement the changes requested by me if you like (and agree to them mainly 😉).

gerhardol:
This PR has one big (small?) problem: The new popup is smaller than a terminal window.
Together with ConEmu pop-over (#6220) and something like #8658 it seems like GE is hanging when it it is just the modal popup waiting for input.
Maybe change the position for the popup?

Sidenote: I happen to believe that Git error handling was included in one of the error presentation PRs.
That is separate though. The Git handling do currently not check the output status, it also merges the stdout and stderr streams.

RussKie:
I'm super busy at work, but I will read the feedback and respond as soon as
I can.

Thank you

GitCommands/Git/ExternalOperationException.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExternalOperationException"/> class with a specified parameters
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="command">The command that led to the exception.</param>
+        /// <param name="arguments">The command arguments.</param>
+        /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public ExternalOperationException(string command, string arguments, string workingDirectory, Exception innerException)
+            : base(null, innerException)
```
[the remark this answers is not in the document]
RussKie: Good catch

GitUI/NBugReports/BugReporter.cs
```diff
@@ -0,0 +1,141 @@
...
+                // Something happened either in the app - can submit the bug to GitHub
+                ReportAppException(externalException, isTerminating);
+                return;
+            }
+
+            // Report any other exceptions
+            ReportGenericException(ex, isTerminating);
+        }
+
+        private static void ReportAppException(ExternalOperationException ex, bool isTerminating)
```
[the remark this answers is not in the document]
RussKie: Used NRT annotations instead.

GitCommands/Git/UserExternalOperationException.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UserExternalOperationException"/> class with a specified parameters
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="command">The command that led to the exception.</param>
+        /// <param name="arguments">The command arguments.</param>
+        /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public UserExternalOperationException(string command, string arguments, string workingDirectory, Exception innerException)
```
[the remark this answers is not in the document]
RussKie: I don't think we can do this without breaking the hierarchy: `public class UserExternalOperationException : ExternalOperationException`.
In order to initialise the exception we need to pass args to the base class, and we only use `context` we won't have information to pass.

Likewise in event where we use this exception, we no longer interested in the original `ExternalOperationException`, but interested in the underlying exception that triggered the sequence:
```cs
catch (ExternalOperationException ex)
{
    throw new UserExternalOperationException(ex.Command, ex.Arguments, ex.WorkingDirectory, ex.InnerException ?? ex);
}
```

GitUI/NBugReports/BugReporter.cs
```diff
@@ -0,0 +1,141 @@
...
+                Cancelable = true,
+            };
+            var btnIgnoreOrClose = new TaskDialogCommandLink("IgnoreOrClose", isTerminating ? Strings.ButtonCloseApp : Strings.ButtonIgnore);
+            btnIgnoreOrClose.Click += (s, e) =>
+            {
+                dialog.Close();
+            };
+            dialog.Controls.Add(btnIgnoreOrClose);
+
+            dialog.Show();
```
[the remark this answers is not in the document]
RussKie: Refactored

RussKie:
> Together with ConEmu pop-over (#6220) and something like #8658 it seems like GE is hanging when it it is just the modal popup waiting for input.
> Maybe change the position for the popup?

In the final version I have reworked owner form handling, so it should have a better chance of setting a popup as the top most window, and thus it shouldn't get lost.

gerhardol:
> In the final version I have reworked owner form handling, so it should have a better chance of setting a popup as the top most window, and thus it shouldn't get lost.

It was harder to recreate #8658 now, maybe after #8679

gerhardol (commented):
Changes work for me
I prefer to take this and move on, maybe adjust later
But I would like @mstv to approve

GitCommands/Git/UserExternalOperationException.cs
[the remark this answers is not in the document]
mstv: The /  my goal is to avoid all those redundant lengthy `catch` blocks and to provide information about the top-level operation.

```
public UserExternalOperationException(string context, ExternalOperationException ex)
    : base(ex.Command, ex.Arguments, ex.WorkingDirectory, ex.InnerException)
{
    Context = context;
}

foo()
{
    try
    {
         ...
    }
    catch (ExternatlOperationException ex)
    {
        throw new UserExternalOperationException("context information", ex);
    }
}
```

mstv:
This PR does not solve #7795 completely. So please remove "fixes" from the PR description.

The discussion about `UserExternalOperationException` and context information should be finished before the merge.

The other minor comments on nested exceptions, refactorings and the application to `MessageBoxes.FailedToRunShell` can be done in follow-up PRs.

gerhardol:
> > Together with ConEmu pop-over (#6220) and something like #8658 it seems like GE is hanging when it it is just the modal popup waiting for input.
> > Maybe change the position for the popup?
> 
> In the final version I have reworked owner form handling, so it should have a better chance of setting a popup as the top most window, and thus it shouldn't get lost.

I could recreate it once, the popup was over the console in my scenario.

GitUI/Script/ScriptRunner.cs
```diff
@@ -34,29 +34,29 @@ public static CommandStatus RunScript(IWin32Window owner, IGitModule module, str
...
 
-            var scriptInfo = ScriptManager.GetScript(scriptKey);
+            ScriptInfo scriptInfo = ScriptManager.GetScript(scriptKey);
             if (scriptInfo is null)
             {
-                showError("Cannot find script: " + scriptKey);
-                return false;
+                ThreadHelper.AssertOnUIThread();
+                throw new UserExternalOperationException($"{Strings.ScriptErrorCantFind}: '{scriptKey}'",
+                    new ExternalOperationException(command: null, arguments: null, module.WorkingDir, innerException: null));
```
RussKie, line 59: ge-script-notfound

GitUI/Script/ScriptRunner.cs
```diff
@@ -72,12 +72,15 @@ string optionDependingOnSelectedRevision
                                                                         && ScriptOptionsParser.Contains(arguments, option));
                 if (optionDependingOnSelectedRevision is object)
                 {
-                    showError($"Option {optionDependingOnSelectedRevision} is only supported when started with revision grid available.");
-                    return false;
+                    ThreadHelper.AssertOnUIThread();
+                    throw new UserExternalOperationException($"{Strings.ScriptText}: '{scriptKey}'{Environment.NewLine}'{optionDependingOnSelectedRevision}' {Strings.ScriptErrorOptionWithoutRevisionGridText}",
+                        new ExternalOperationException(scriptInfo.Command, arguments, module.WorkingDir, innerException: null));
```
RussKie, line 77: ge-script-norg

GitUI/Script/ScriptRunner.cs
```diff
@@ -86,8 +89,9 @@ string optionDependingOnSelectedRevision
             (string argument, bool abort) = ScriptOptionsParser.Parse(scriptInfo.Arguments, module, owner, revisionGrid);
             if (abort)
             {
-                showError($"There must be a revision in order to substitute the argument option(s) for the script to run.");
-                return false;
+                ThreadHelper.AssertOnUIThread();
+                throw new UserExternalOperationException($"{Strings.ScriptText}: '{scriptKey}'{Environment.NewLine}{Strings.ScriptErrorOptionWithoutRevisionText}",
+                    new ExternalOperationException(scriptInfo.Command, arguments, module.WorkingDir, innerException: null));
```
RussKie, line 94: ge-script-norev

Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <Import Project=".\scripts\sdk\Sdk.props" />
 
   <PropertyGroup>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9</LangVersion>
```
gerhardol, line 5: So support for C# 9.0 added
(Requires Visual Studio 16.8 which should be OK.)
Do you have a list what features that are supported with .NET 4 framework?

mstv (approved):
Thank you.

Directory.Build.props
RussKie: This actually not required, I was tinkering with something else.
Nullable annotations came with C# 8.0.

PR #8371, opened 2020-08-02
Added support for showing multi-line commit messages in the revision graph

FodderMK:
## Proposed changes

- Show multi-line commit messages in the revision graph.

## Reasoning
Our office has a commit-formatting policy that says the ticket is the first line of the commit followed by lines detailing the commits.  This doesn't cause a problem when I look at commits and pull requests on Bitbucket because it shows the additional lines, but in GitExtensions I only see the first line which isn't very useful.  This change allows the revision graph behave more like Bitbucket by showing the entire commit message instead of truncating after the first line.

### Before
image

### After
image

## Test methodology
- Manually tested

## Test environment(s) 

- GIT 2.27.0.windows.1
- Windows 10.0.19041 N/A Build 19041

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using GitExtUtils;
 using GitUI;
@@ -416,11 +415,6 @@ namespace GitCommands
                 Debug.Fail("Unable to read body from the log -- this should not happen");
                 revision = default;
                 return false;
-            }
-
-            if (AppSettings.ShowMultiLineCommitMessages)
-            {
-                subject = Regex.Replace(body, "[\r\n]", " ");
             }
 
             #endregion
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using GitExtUtils;
 using GitUI;
@@ -415,6 +416,11 @@ namespace GitCommands
                 Debug.Fail("Unable to read body from the log -- this should not happen");
                 revision = default;
                 return false;
+            }
+
+            if (AppSettings.ShowMultiLineCommitMessages)
+            {
+                subject = Regex.Replace(body, "[\r\n]", " ");
             }
 
             #endregion
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1719,6 +1719,12 @@ namespace GitCommands
             return Properties.Settings.Default.IsPortable;
         }
 
+        public static bool ShowMultiLineCommitMessages
+        {
+            get => GetBool("showmultilinecommitmessages", false);
+            set => SetBool("showmultilinecommitmessages", value);
+        }
+
         private static IEnumerable<(string name, string value)> GetSettingsFromRegistry()
         {
             RegistryKey oldSettings = VersionIndependentRegKey.OpenSubKey("GitExtensions");
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1128,6 +1128,12 @@ namespace GitCommands
             set => SetInt("revisiongridquicksearchtimeout", value);
         }
 
+        public static bool ShowMultiLineCommitMessages
+        {
+            get => GetBool("showmultilinecommitmessages", false);
+            set => SetBool("showmultilinecommitmessages", value);
+        }
+
         /// <summary>Gets or sets the path to the git application executable.</summary>
         public static string GitBinDir
         {
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1719,12 +1719,6 @@ namespace GitCommands
             return Properties.Settings.Default.IsPortable;
         }
 
-        public static bool ShowMultiLineCommitMessages
-        {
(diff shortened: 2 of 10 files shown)


gerhardol:
I like the concept, use the space available.
Viewed very briefly, no comments. While reordering is good for understanding, it slows down reviews.

Maybe this should be the default?
GE has too many settings already.
(The body is not stored for old messages to reduce memory usage (retrieved also for tooltip when selecting a commit?). I have not investigated how this PR deals with retrieving them, but the grid could be updated asynch or require selection as is now.)

FodderMK:
The reordering you're referring to is with the AppearanceSettingsPage.Designer.cs file?  Visual Studio did all of that when I added the new line in so I didn't want to mess around with it but I could try to localize the changes.  If we were to make this behavior the default then I could revert the file entirely as we wouldn't need the setting.

gerhardol:
>The reordering you're referring to is with the AppearanceSettingsPage.Designer.cs file?

Yes. Often it is better to manually cherry-pick what is needed.

Not requesting changes (yet), someone else should give their opinion.

gerhardol:
Adding body to the subject will require memory for all commits. A big repo like linux will need a lot of extra memory.
Also, for the commits where body is retained (less than 6 months), the body is stored twice
Also, the tooltips look a little strange, see below before and after the commit is selected.

My proposal, @gitextensions/git-extensions-source ?
I suggest the body is added to the revision grid if it is available instead. This means that the body is not displayed if it is older than 6 months and has not been selected.
Keep the ellipsis as is.
No configuration (maybe for time retaining the body if this is really needed for older commits too

image

image

FodderMK:
Thanks for the feedback!

I have removed the setting and instead use the body to populate the message column (if available).  I have to refresh the grid when an old commit is clicked and the body is updated.  I'm not a fan of this but it seemed like it would take a bigger refactor to get the body populated before the grid refreshed with the current flow.  I'm not very familiar with GUI systems though. maybe there's a better way?

RussKie:
I'd go with no changes to the settings dialogs, but with a context menu:
image


> I suggest the body is added to the revision grid if it is available instead. This means that the body is not displayed if it is older than 6 months and has not been selected.

We need to see how big of an impact this has. 
To me personally this would add (a lot of) visual noise, if I need to see the body of a commit, I select it.

RussKie:
If memory serves me well, @mstv  and @pmiossec spent a lot of time in this area.

FodderMK:
Added the context menu item to toggle the option.

RussKie:
Please update the description message and screenshots, if applicable. Looking at the code I see completely different implementation from what's been described.
Thanks.

FodderMK:
Done.

RussKie:
I'm not entirely sold on this change, but happy to give it a go.

RussKie:
Thank you

crypto-rsa:
I just recently fetched recent changes in GE and immediately noticed this change. I totally agree with @RussKie in _To me personally this would add (a lot of) visual noise, if I need to see the body of a commit, I select it._ After all, that's what the multiline indicator is for.
A single on/off option would solve this. But recently I saw a lot of arguments like _GE already has a lot of settings_ which is perhaps what prevented the addition of one in this issue as well. Well, GE is a complex software and I think one of its strengths is its flexibility which largely comes from its extensive settings. What's the problem with adding a new option? Power users (who I suppose GE is built for) can surely find options which are relevant to them.

gerhardol:
I like this change, it uses otherwise unused space. Maybe the subject and body should be separated a little more.

GE has many options, even power users and maintainers get lost in them. I generally like options myself, but GE should be cleaned up. It is hard to remove options that have been added.
Options should only be added if there is a strong need.

So I still vote against adding an option here.

pmiossec:
I'm also a little disappointed by this feature...

That's sometimes useful because you can read what is on the 2nd line of the commit message without having to select it. But the tooltip helps also on that...

**But**, in my company, we generally write a lot of text in the commit message.
The result is that we have a nearly all the lines full.
And because we can't see where the "subject"/first line of the commit ends, it makes **very** difficult to search for a specific commit in the graph.

If we want to not add a setting for that, I think this feature need to be reworked so that we see the difference between the first line of the commit and the others.

We must see the clear difference between both!

I think the other lines must be displayed lighter/in gray (on light theme).

@FodderMK talked about BitBucket to say what it is an interesting feature, but that's what they are doing:
image

For the comparison, Github choose to note display it by default to make things more readable and being able to display it with a different font when clicking on the ellipsis. That's similar to what we have with the tooltips:
image

Gitlab do the same than GitHub.

RussKie:
Do we have a desire to fix it, or roll back?

NikolayXHD:
> Do we have a desire to fix it, or roll back?

I am willing to attempt graying out

pmiossec:
>I am willing to attempt graying out

Thanks a lot!

gerhardol:
If graying out is not possible, would a marker or some more spaces be sufficient? 
Otherwise a setting could be added (I only have one vote so I could not stop it if I wanted to... ) but please do not add settings for every possibility. 

But if the wizard looks at it, there will be a better solution.

jrfnl:
I agree with with previous commenters have said: this makes the screen incredibly noisy and decreases usability because of it, especially in a repo where commits are wordy. Very big vote from me for a setting to turn this noise off.

jwfx:
Is there really no way to disable this change?

Especially in connection with `git svn` which adds quite long git-svn-ids in the commit message this is really killing my eyes...

DCFerg:
In my mind this fundamentally breaks with Git tradition that the commit title line should be short with additional information only included in the body. It makes the screen very noisy. I have used GE for 7 years and this is the first time there has been a change with so much visual impact. It was immediately obvious and bad. I searched in vain for an associated option. SURELY there would not be such a visible change without an option!

This is a classic case in favor of adding an option, IMO.  When an "improvement" makes a significant change in the product behavior that has arguably negative aspects, there should be an option to disable it; or the "improvement" should just be dropped.

I really dislike this. Please add an option.

pmiossec:
> I really dislike this. Please add an option.

I  wonder the goal and the tact of this message when the option has been added some days ago in #9301

Issue #8339, opened 2020-07-19
Settings dialog takes long to load


RussKie:
## Current behaviour

It takes observable amount of time for the settings dialog to start.

The following was measured in a debug build with a debugger attached across several runs - the numbers fluctuated slightly but remain within 10% corridor. Obviously it would be faster in a release build, but not significantly different, as the settings dialog is still very slow in the released app.

| Section | Time (ms) |
|---|----:|
| CommonLogic | 2 |
| CheckSettingsLogic | 0 |
| ChecklistSettingsPage | 686 |
| GitExtensionsSettingsGroup | 2 |
| GeneralSettingsPage | 268 |
| AppearanceSettingsPage | 332 |
| RevisionLinksSettingsPage | 616 |
| BuildServerIntegrationSettingsPage | 8 |
| ScriptsSettingsPage | 188 |
| HotkeysSettingsPage | 123 |
| ShellExtensionSettingsPage | 35 |
| AdvancedSettingsPage | 71 |
| DetailedSettingsPage | 117 |
| SshSettingsPage | 142 |
| GitSettingsGroup | 989 |
| PluginsSettingsGroup | 15 |

The total is whooping **4+ seconds** is taken by `FormSettings..ctor` **_before_** the dialog is shown.
When the dialog is show settings for each page are loaded:

| Section | Time (ms) |
|---|----:|
| GitExtensionsSettingsGroup | 0 |
| ChecklistSettingsPage | 0 |
| GeneralSettingsPage | 32 |
| AppearanceSettingsPage | 58 |
| ColorsSettingsPage | 27 |
| AppearanceFontsSettingsPage | 14 |
| RevisionLinksSettingsPage | 23 |
| BuildServerIntegrationSettingsPage | 5 |
| ScriptsSettingsPage | 2 |
| HotkeysSettingsPage | 2 |
| ShellExtensionSettingsPage | 3 |
| AdvancedSettingsPage | 2 |
| ConfirmationsSettingsPage | 1 |
| DetailedSettingsPage | 3 |
| FormBrowseRepoSettingsPage | 18 |
| CommitDialogSettingsPage | 1 |
| DiffViewerSettingsPage | 6 |
| SshSettingsPage | 38 |
| GitSettingsGroup | 0 |
| GitSettingsPage | 70 |
| GitConfigSettingsPage | 72 |
| GitConfigAdvancedSettingsPage | 0 |
| GitRootIntroductionPage | 0 |
[...]

PR #8372, opened 2020-08-02
Optimise `FormSettings` load

RussKie:
Contributes to #8339

## Proposed changes

Settings dialog takes on average 4+ seconds to load and show to a user. With that the initial load is tearing the UI, which is visually unpleasant.

Few major culprits:

* HasClippedControl() - ~2.5 seconds, which is 60%+ of the observed 4 seconds.
This appears to be trying to detect whether the settings dialig fully fits on a user's screen, and, if not, reduces the current font.

* settingsTreeView.GotoPage(_initialPage) - 500-600ms on average, which deals with raising notification events, removing the currently shown control, and binding the requested page. Depending on a complexity of UI of the selected page this can take anywhere from 10ms to 1 second (or longer).

* RevisionLinksSettingsPage - on average it takes 850-900ms to create this page, which breaks down to:
    - InitializeComponent: 50-55%
    - InitializeComplete: 40%
    - everything else: 5-10%.

To speed up the load the following has been done:

* HasClippedControl() is removed. It is up to a user to set an appropriate font size. Also the current telemetry shows that there is only a very small
percent of our users (out of those who opted in to the telemetry) have monitors with resolution of less than 800px high.

* The dialog load sequence has been rearranged to:
    - build the tree and load pages upon loading the dialog (OnRuntimeLoad),
    - navigate to a selected page upon showing the diialog.

The above makes the dialog to appear slightly faster by 50% (slightly over ~2 seconds vs 4+ seconds), and slightly more responsive.

Some of the slowness is likely caused by the Windows Forms layout engine, and could be improved by migrating the .NET Core/.NET versions.
But by far and large the handling of settings and the dialog's load sequence must be rewritten to utilise lazy, on-demand and async patterns.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormSettings.Designer.cs
+++ b/GitUI/CommandsDialogs/FormSettings.Designer.cs
@@ -221,7 +221,6 @@ namespace GitUI.CommandsDialogs
             this.SizeGripStyle = System.Windows.Forms.SizeGripStyle.Show;
             this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
             this.Text = "Settings";
-            this.Load += new System.EventHandler(this.FormSettings_Load);
             this.Shown += new System.EventHandler(this.FormSettings_Shown);
             ((System.ComponentModel.ISupportInitialize)(this.pictureBox2)).EndInit();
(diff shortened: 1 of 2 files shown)


gerhardol:
Seem OK, brief review only

PR #8384, opened 2020-08-06
Fixes #8348. Allow to create patch for the root commit.

### LINKED ISSUE #8348: Unable to "Format patch" of first commit
## Current behaviour

Patch error dialog pops up with the message "Unable to to create patch file(s)".

## Expected behaviour

Patch file created without error.

## Steps to reproduce

1) Select "Format patch..." from the "Commands" drop down menu.
2) Select the first commit.  The one at bottom of the list (the initial commit).
3) Click the "Create patch(es)" button

Doubt it previously worked.  But I don't actually know.

## Environment

- Git Extensions 3.4.2.9737
- Build 082cfd97d09c47346638b1873f5c8cf6ed139382
- Git 2.27.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version:

veremenko-y:
Fixes #8348

## Proposed changes

If base commit is not defined, use `--root` argument to create patch for root or orphaned comit.

## Test methodology 

- Tested on empty repository with one commit
- Tested with non root commits. Merges and single ones.
- Added additional test case for when `from` is empty.

## Test environment(s)

- GIT  2.27.0.windows.1
- Windows 10 1909

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1364,7 +1364,7 @@ namespace GitCommands
                 {
                     "-M -C -B",
                     { start != null, $"--start-number {start}" },
-                    $"{from.Quote()}..{to.Quote()}",
+                    { !string.IsNullOrEmpty(from), $"{from.Quote()}..{to.Quote()}", $"--root {to.Quote()}" },
                     $"-o {output.ToPosixPath().Quote()}"
                 });
         }
--- a/UnitTests/GitCommands.Tests/GitModuleTests.cs
+++ b/UnitTests/GitCommands.Tests/GitModuleTests.cs
@@ -790,6 +790,26 @@ namespace GitCommandsTests
             _gitModule.FormatPatch(from, to, outputFile, start).Should().Be(dummyCommandOutput);
         }
 
+        [TestCase(new object[] { "", "567", "output.file", null })]
+        [TestCase(new object[] { "", "567", "output.file", 1 })]
+        [TestCase(new object[] { null, "567", "output.file", 2 })]
+        public void Test_FormatPatchInRoot(string from, string to, string outputFile, int? start)
+        {
+            var arguments = new StringBuilder();
+            arguments.Append("format-patch -M -C -B");
+            if (start != null)
+            {
+                arguments.AppendFormat(" --start-number {0}", start);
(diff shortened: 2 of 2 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -1364,7 +1364,8 @@ public string FormatPatch(string from, string to, string output, int? start = nu
                 {
                     "-M -C -B",
                     { start != null, $"--start-number {start}" },
-                    $"{from.Quote()}..{to.Quote()}",
+                    { !string.IsNullOrEmpty(from), $"{from.Quote()}..{to.Quote()}" },
+                    { string.IsNullOrEmpty(from), $"--root {to.Quote()}" },
```
gerhardol, line 1368:
```suggestion
                    { !string.IsNullOrEmpty(from), $"{from.Quote()}..{to.Quote()}", $"--root {to.Quote()}" },
```

I prefer the normal outcome first here, others may prefer a non negated test

veremenko-y: I can just make an if outside the initializer, to make it clearer. Thoughts?

veremenko-y: Fixed as requested for now.

vbjay:
I just added a script.

git rebase --root -i

veremenko-y:
> I just added a script. git rebase --root -i

I'm not following.

vbjay:
Go to settings. Look at the scripts settings.

veremenko-y:
I don't see how it helps with "Format patch" function failing to create one.

gerhardol:
>> I just added a script.

> I don't see how it helps with "Format patch" function failing to create one.

I interpret the comment as "use scripts instead of changing the code".
This change does have a merit though, a user do not understand why a patch cannot be created. Scripts could replace a lot of GE core functionality if going to the extreme, but the user interface would be horrible (better to use Git directly). 
I do not believe that the importance is high for the PR, but it is well motivated.

gerhardol:
@msftbot merge in 24 hours

vbjay:
Can a rebase --root patch apply to a clone of the current repo?

PR #8408, opened 2020-08-21
Theme variations for colorblindness

### LINKED ISSUE #8350: Add ability to change color for remote and local branches for the color blind
## Feature description

Currently, local branch is red, and remote branch is green.
A person with Red-Green color blindness does not see any difference!

Please consider adding the ability to manually select colors for remote and local branches.

I've also tested to see if dark mode does any difference, but it's not :-(

## Environment

- Git Extensions 3.4.2.9737
- Build 082cfd97d09c47346638b1873f5c8cf6ed139382
- Git 2.27.0.windows.1
- .NET Framework 4.7.3620.0
- DPI 96dpi (no scaling)

NikolayXHD:
Fixes #8350

## Proposed changes

Replace red with blue. 
image

Observe the table from wikipedia.

image

- Deuteranomaly, Protanomaly - slightly decreased distinction between red / green, 
  should have no need for colorblind theme

- Protanopia, Deuteranopia - very weak distinction between red / green, 
  which is fixed in variations proposed in this PR

- Tritanopia, Tritanomaly - distinguishes red / green normally, weak distinction between blue / green and  yellow / red
  from my understanding these people should be fine with default themes, because we do not frequently use yellow / orange colors

## Screenshots

bright-colorblind
image

dark-colorblind
image

darksilver-colorblind
image

----


pmiossec:
When I see that, can we imagine to have theme inheritance?

I can imagine it working like:
The first line of the css file could have a comment line `//inherit-from:dark
And if encounter this line, we load all the colors of this theme and override only the color specified in the theme file.

What do you think?

neil5280:
This is amazing work @NikolayXHD .  Thank you for picking up the banner and running with it.  I'm excited to see this in the next release.

NikolayXHD:
> When I see that, can we imagine to have theme inheritance?
> 
> I can imagine it working like:
> The first line of the css file could have a comment line `//inherit-from:dark
> And if encounter this line, we load all the colors of this theme and override only the color specified in the theme file.
> 
> What do you think?

I could't agree more

RussKie (approved):
🥇

NikolayXHD:
I will later submit another PR with inheritance

PR #8359, opened 2020-07-26
Dialogs layout alignment `FormCreateBranch`

RussKie:
Contributes to to #6183

## Proposed changes

- Update `FormCreateBranch` layout - this is a truly nasty dialog to lay out 🤯 
- Aligned buttons sizes in other dialogs
- Lighten up the controls panel slightly

Verified at 100% and 150% scale factor, both light and dark versions.

### Before

image
image

### After

image
image

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8409, opened 2020-08-21
Ensure consistent margins

RussKie:
Resolves https://github.com/gitextensions/gitextensions/pull/8359#issuecomment-667762221

## Proposed changes

- Ensure consistent margins

### After

image

image

image


ivangrek:
I would recommend setting margins to 0.
For contained controls.

image

For 24px around.

ivangrek:
Also 
https://docs.microsoft.com/en-us/windows/uwp/design/layout/alignment-margin-padding#general-recommendations

RussKie:
Right now it 27px for the main content, and less for the buttons (they should be nudged to the left to align)
image
image

The custom "Help" button must be removed and the form's [?] (Help) button should be used instead. This is a separate piece of work altogether.

PR #8414, opened 2020-08-22
Filter out hidden branches

RussKie:
## Proposed changes

Indicate branches and tags in the left panel that are currently filtered out and not visible in the revision grid.

* The hidden refs are greyed out and their icon is changed to "closed eye".
* Each node in the tree now has a new property `IsVisible` that can be used as a conditional flag when implementing necessary logic.

Detection of whether a given ref is visible or not is done by checking the revision grid, which is currently the source of truth. 
Whenever the revision grid is refreshed, it raises an event that causes the left panel to update its state, and show or hide nodes.

### After

* Show all
image

* Show only current branch
  image
  image
  image

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
@@ -224,15 +224,15 @@ namespace GitUI.BranchTreePanel
         {
             _sortOrderContextMenuItem = new GitRefsSortOrderContextMenuItem(() =>
             {
-                _branchesTree.RefreshRefs();
-                _remotesTree.RefreshRefs();
-                _tagTree.RefreshRefs();
+                _branchesTree.Refresh();
+                _remotesTree.Refresh();
+                _tagTree.Refresh();
             });
             _sortByContextMenuItem = new GitRefsSortByContextMenuItem(() =>
             {
-                _branchesTree.RefreshRefs();
-                _remotesTree.RefreshRefs();
-                _tagTree.RefreshRefs();
+                _branchesTree.Refresh();
+                _remotesTree.Refresh();
+                _tagTree.Refresh();
             });
 
             _localBranchMenuItems = new LocalBranchMenuItems<LocalBranchNode>(this);
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
@@ -1,13 +1,16 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Drawing;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands.Git;
+using GitExtUtils.GitUI.Theming;
 using GitUI.BranchTreePanel.Interfaces;
 using GitUI.Properties;
+using GitUI.UserControls.RevisionGrid;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
 using Microsoft.VisualStudio.Threading;
@@ -18,30 +21,12 @@ namespace GitUI.BranchTreePanel
     {
         #region private classes
 
+        [DebuggerDisplay("(Node) FullPath = {FullPath}")]
         private abstract class BaseBranchNode : Node
         {
             protected const char PathSeparator = '/';
 
-            /// <summary>Short name of the branch/branch path. <example>"issue1344"</example></summary>
-            protected string Name { get; set; }
-
-            private string ParentPath { get; }
-            protected string AheadBehind { get; set; }
-
-            /// <summary>Full path of the branch. <example>"issues/issue1344"</example></summary>
-            public string FullPath => ParentPath.Combine(PathSeparator.ToString(), Name);
-
-            public override int GetHashCode()
-            {
-                return FullPath.GetHashCode();
-            }
-
-            public override bool Equals(object obj)
-            {
-                return obj is BaseBranchNode other && (ReferenceEquals(other, this) || string.Equals(FullPath, other.FullPath));
-            }
-
-            protected BaseBranchNode(Tree tree, string fullPath)
+            protected BaseBranchNode(Tree tree, string fullPath, bool visible)
                 : base(tree)
             {
                 fullPath = fullPath.Trim();
@@ -53,6 +38,44 @@ namespace GitUI.BranchTreePanel
                 var dirs = fullPath.Split(PathSeparator);
                 Name = dirs[dirs.Length - 1];
                 ParentPath = dirs.Take(dirs.Length - 1).Join(PathSeparator.ToString());
+                Visible = visible;
+            }
+
+            protected string AheadBehind { get; set; }
+
+            /// <summary>
+            /// Short name of the branch/branch path. <example>"issue1344"</example>.
+            /// </summary>
+            public string Name { get; protected set; }
+
+            protected string ParentPath { get; }
+
+            /// <summary>
+            /// Full path of the branch. <example>"issues/issue1344"</example>.
+            /// </summary>
+            public string FullPath => ParentPath.Combine(PathSeparator.ToString(), Name);
+
+            /// <summary>
+            /// Gets whether the commit that the node represents is currently visible in the revision grid.
+            /// </summary>
+            public bool Visible { get; }
+
+            protected override void ApplyStyle()
+            {
+                base.ApplyStyle();
+
+                TreeViewNode.ForeColor = Visible ? TreeViewNode.TreeView.ForeColor : Color.Silver.AdaptTextColor();
+                TreeViewNode.ImageKey =
+                    TreeViewNode.SelectedImageKey = Visible ? null : nameof(Images.EyeClosed);
+            }
+
+            public override int GetHashCode() => FullPath.GetHashCode() ^ Visible.GetHashCode();
+
+            public override bool Equals(object obj)
+            {
+                return obj is BaseBranchNode other
+                    && (ReferenceEquals(other, this) || string.Equals(FullPath, other.FullPath))
+                    && Visible == other.Visible;
             }
 
             public void UpdateAheadBehind(string aheadBehindData)
@@ -119,8 +142,8 @@ namespace GitUI.BranchTreePanel
 
             private bool _isMerged = false;
 
-            public BaseBranchLeafNode(Tree tree, in ObjectId objectId, string fullPath, string imageKeyUnmerged, string imageKeyMerged)
-                : base(tree, fullPath)
+            public BaseBranchLeafNode(Tree tree, in ObjectId objectId, string fullPath, bool visible, string imageKeyUnmerged, string imageKeyMerged)
+                : base(tree, fullPath, visible)
             {
                 ObjectId = objectId;
                 _imageKeyUnmerged = imageKeyUnmerged;
@@ -138,8 +161,13 @@ namespace GitUI.BranchTreePanel
                     }
 
                     _isMerged = value;
+
                     ApplyStyle();
-                    TreeViewNode.ToolTipText = IsMerged ? string.Format(Strings.ContainedInCurrentCommit, Name) : string.Empty;
+
+                    if (_isMerged && Visible)
+                    {
+                        TreeViewNode.ToolTipText = string.Format(Strings.ContainedInCurrentCommit, Name);
+                    }
                 }
             }
 
@@ -149,14 +177,23 @@ namespace GitUI.BranchTreePanel
             protected override void ApplyStyle()
             {
                 base.ApplyStyle();
-                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey = IsMerged ? _imageKeyMerged : _imageKeyUnmerged;
+
+                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey =
+                    Visible
+                        ? IsMerged ? _imageKeyMerged : _imageKeyUnmerged
+                        : nameof(Images.EyeClosed);
+                if (!Visible)
+                {
+                    TreeViewNode.ToolTipText = string.Format(Strings.InvisibleCommit, FullPath);
+                }
             }
         }
 
+        [DebuggerDisplay("(Local) FullPath = {FullPath}, Hash = {ObjectId}, Visible: {IsVisible}")]
         private sealed class LocalBranchNode : BaseBranchLeafNode, IGitRefActions, ICanRename, ICanDelete
         {
-            public LocalBranchNode(Tree tree, in ObjectId objectId, string fullPath, bool isCurrent)
-                : base(tree, objectId, fullPath, nameof(Images.BranchLocal), nameof(Images.BranchLocalMerged))
+            public LocalBranchNode(Tree tree, in ObjectId objectId, string fullPath, bool isCurrent, bool visible)
+                : base(tree, objectId, fullPath, visible, nameof(Images.BranchLocal), nameof(Images.BranchLocalMerged))
             {
                 IsActive = isCurrent;
             }
@@ -166,18 +203,17 @@ namespace GitUI.BranchTreePanel
             protected override void ApplyStyle()
             {
                 base.ApplyStyle();
+
                 SetNodeFont(IsActive ? FontStyle.Bold : FontStyle.Regular);
             }
 
             public override bool Equals(object obj)
-            {
-                return base.Equals(obj) && obj is LocalBranchNode localBranchNode && IsActive == localBranchNode.IsActive;
-            }
+                => base.Equals(obj)
+                    && obj is LocalBranchNode localBranchNode
+                    && IsActive == localBranchNode.IsActive;
 
             public override int GetHashCode()
-            {
-                return base.GetHashCode();
-            }
+                => base.GetHashCode() ^ IsActive.GetHashCode();
 
             internal override void OnDoubleClick()
             {
@@ -223,17 +259,20 @@ namespace GitUI.BranchTreePanel
 
         private class BasePathNode : BaseBranchNode
         {
-            public BasePathNode(Tree tree, string fullPath) : base(tree, fullPath)
+            public BasePathNode(Tree tree, string fullPath) : base(tree, fullPath, visible: true)
             {
             }
 
             protected override void ApplyStyle()
             {
                 base.ApplyStyle();
-                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey = nameof(Images.BranchFolder);
+
+                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey =
+                    FullPath == Strings.Inactive ? nameof(Images.EyeClosed) : nameof(Images.BranchFolder);
             }
         }
 
+        [DebuggerDisplay("(Branch path) FullPath = {FullPath}")]
         private class BranchPathNode : BasePathNode
         {
             public BranchPathNode(Tree tree, string fullPath)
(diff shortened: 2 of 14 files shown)


RussKie:
You can switch to another branch. Not showing hidden branches makes
switching unnecessary hard.

ivangrek:
I am trying to understand the general logic behind hiding branches and how this should affect the left panel.

If I want to "Show only the current branch". And there is no left panel. Of course, I only want to work with what I chose in the setup.
What will change in this process when you enable the left panel?

If we want to make it possible to change the current branch, then there is no need to hide the branches in the left panel, as this is an alternative access to branches.
And then launching scripts should be similar.

Now I am inclined to not hide the branches, but to have a flag so that I can understand that the script cannot be executed. But it will confuse the user.

I may be missing something else on the functionality of the left pane.

I see a branch and want to execute a script on it.
What could prevent this, besides the current implementation?

ivangrek:
So as not to get confused in concepts selected/not selected.
May be config.

`ScriptRunner.RunScript(this, Module, name, UICommands, this, config => { config.SelectedHash = "[some value]"})`

And then

```
case "sHash":
    if(config.SelectedHash != null)
    {
        newString = config.SelectedHash;
    }
    else
    {
        newString = selectedRevision.Guid;
    }
    break;
```

Configuration will overload default implementation.

RussKie:
> If I want to "Show only the current branch". And there is no left panel. Of course, I only want to work with what I chose in the setup.
> What will change in this process when you enable the left panel?

No changes to how you work. No left panel - no issues, you can't click on a branch that is no visible to execute a script.
(Btw thank you for reminding me about this scenario, no event should be raised in this case.)




> Now I am inclined to not hide the branches, but to have a flag so that I can understand that the script cannot be executed. But it will confuse the user.

The flag is `LocalBranchNode.IsVisible`.
The branches are not getting hidden, they are moved to a different node called "Hidden".


> May be config.
> 
> `ScriptRunner.RunScript(this, Module, name, UICommands, this, config => { config.SelectedHash = "[some value]"})`

There are dozens of variable, passing them all would be a maintenance headache.

> I'm sure you know the answer 😁
> 
> [#8414 (comment)](https://github.com/gitextensions/gitextensions/pull/8414#issuecomment-678737702)

Haha! Thanks
Btw with this we should mark hidden branches in the toolstip dropdown too... 🤔 
image

ivangrek:
>No changes to how you work. No left panel - no issues, you can't click on a branch that is no visible to execute a script.

In case the left panel is an alternative access to branches, adding a folder for hidden ones will add informative value, but does not remove the question of the impossibility of executing the script. Since switching a branch is the same script and it is possible.

An abstract example.
There are no products in the showcase, but we can buy them from the catalog.
Marking in the catalog that an item is not in the storefront does not change the purchase process.

Is the left pane an alternate view or second grid view?

> There are dozens of variable, passing them all would be a maintenance headache.

I see it as a parameter provider. Separate interface and implementation for the grid and left pane. And a separate interface for manipulating the grid (refresh).

RussKie:
The left panel is complimentary to the grid.

Scripts must not be executable for hidden branches, since information
necessary for execution is coming from the grid.
In fact some information may not be available at all until a revision is
selected (e.g. the full commit message).

ivangrek:
For # 8405, the flag is enough.

Regarding these changes, I am assuming that in the left pane we can execute any command, regardless of the current state of the grid.

Could you please show an example of a command that cannot be executed for the left pane?
For the test.

>Scripts must not be executable for hidden branches, since information
necessary for execution is coming from the grid.
In fact some information may not be available at all until a revision is
selected (e.g. the full commit message).

Is this a requirement or an insurmountable circumstance?

RussKie:
Think of the left panel as a navigation helper, like in Windows Explorer or
MS Outlook.

A script runner must provide enough information to substitute any supported
variable the app defines (please see the list in the code).
The left panel does not have all information (short of running and parsing
extra commands), only the grid has.

ivangrek:
`var revision = ((GitModule)module).GetRevision(ObjectId.Parse("7948b455f5c100ac9c382068d70889b647cb3f62"));`

RussKie:
Yes, but this means we need to execute a git command, and this has perf implications.

And it raises a question - why would I need to execute a command against a branch that I don't currently see?

RussKie:
> I'm sure you know the answer 😁
> 
> [#8414 (comment)](https://github.com/gitextensions/gitextensions/pull/8414#issuecomment-678737702)

image

Here's how it looks... I'm sort of sitting on a fence here. What do you all think?

gerhardol:
> Here's how it looks... I'm sort of sitting on a fence here. What do you all think?

I prefer to not move the hidden to a separate folder. The icon is fine.
But to remove the branches completely is OK too, if you see the sidepanel as an extension to the grid.

RussKie:
I find it useful having them accessible - it is easy to switch (i.e. checkout) or update filter (i.e. "show only this branch").

ivangrek:
> Yes, but this means we need to execute a git command, and this has perf implications.

Performance is another matter. A matter of implementation, not user use.
The state seems to be cached.

```
public GitRevision GetRevision([CanBeNull] ObjectId objectId = null, bool shortFormat = false, bool loadRefs = false)
{
    const string formatString =
        /* Hash           */ "%H%n" +
        /* Tree           */ "%T%n" +
        /* Parents        */ "%P%n" +
        /* Author Name    */ "%aN%n" +
        /* Author EMail   */ "%aE%n" +
        /* Author Date    */ "%at%n" +
        /* Committer Name */ "%cN%n" +
        /* Committer EMail*/ "%cE%n" +
        /* Committer Date */ "%ct%n" +
        /* Encoding       */ "%e%n";

    var format = formatString + (shortFormat ? "%s" : "%B%nNotes:%n%-N");

    {
        "-n1",
        $"--pretty=format:{format}",
        objectId
    };

    var revInfo = _gitExecutable.GetOutput(args, cache: GitCommandCache, outputEncoding: LosslessEncoding);
```

> And it raises a question - why would I need to execute a command against a branch that I don't currently see?

This is the key question. The answer depends on whether the branch is really not visible or we want to. The user sees it and asks to add an opportunity.

Perhaps the solution is to refuse implementation with a link to a violation of the concept?

Either undo the implementation, hide everything, or implement an option or an intermediate option with hidden branches. 😄

gerhardol:
>I find it useful having them accessible - it is easy to switch (i.e. checkout) or update filter (i.e. "show only this branch").

I could live without seeing them, but this is a fair view.
My vote is on seeing the hidden branches with a different icon and limited context menu.

A solution just refusing execution (i.e. after the user tried to invoke it) is OK for me - you generally know if you hide branches

ivangrek:
For the current revision in ScriptOptionsParser

```
[CanBeNull]
private static GitRevision GetCurrentRevision(
    [NotNull] IGitModule module, [CanBeNull] IScriptHostControl scriptHostControl, List<IGitRef> currentTags, List<IGitRef> currentLocalBranches,
    List<IGitRef> currentRemoteBranches, List<IGitRef> currentBranches)
{
    GitRevision currentRevision;
    IEnumerable<IGitRef> refs;
    if (scriptHostControl == null)
    {
        var currentRevisionGuid = module.GetCurrentCheckout();
        currentRevision = currentRevisionGuid == null ? null : new GitRevision(currentRevisionGuid);
        refs = module.GetRefs(true, true).Where(gitRef => gitRef.ObjectId == currentRevisionGuid);
    }
    else
    {
        currentRevision = scriptHostControl.GetCurrentRevision();
        refs = currentRevision?.Refs ?? Array.Empty<IGitRef>();
    }
```

`currentRevision = scriptHostControl.GetCurrentRevision();`

And in RevisionGridControl

```
public GitRevision GetCurrentRevision()
{
    return Module.GetRevision(CurrentCheckout, shortFormat: true, loadRefs: true);
}
```

`internal ObjectId CurrentCheckout { get; private set; }`

We use git command also.

We have given the user a tool to execute any command. The argument that the branch is not visible in the grid is not enough. Perhaps all added commands will be background and not associated with the grid.

In any case, the last word is yours. 😄 👍 

---

From basic set "Open on GitHub"

RussKie:
Sits on top of https://github.com/gitextensions/gitextensions/pull/8427, which in turn sits on top of https://github.com/gitextensions/gitextensions/pull/8434. Both introduce necessary infra and plumbing.
Both local branches and tags are filtered and correctly presented to users, remote branches are TBD.

RussKie:
> Is this the correct state?
> 
> image
> 
> image


Thank you @ivangrek, I updated the label to be `[ Hidden ]` which is an invalid git ref :)

RussKie:
https://github.com/gitextensions/gitextensions/pull/8414/commits/6392092f447ee98728977683ea4e18867b41e82d is ready for review (still based on top of https://github.com/gitextensions/gitextensions/pull/8427).
Screenshots are updated in the original post.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -158,11 +177,89 @@ internal bool FetchPruneAll()
...
+                {
+                    parentPath = ParentPath;
+                }
+                else if (string.IsNullOrEmpty(ParentPath))
+                {
+                    parentPath = Strings.Hidden;
+                }
+                else
+                {
+                    if (ParentPath == _remote.Name)
```
gerhardol, line 232: remove invert?

gerhardol (commented):
When you select a branch/tag below Hidden, nothing happens when you click them

I am not thrilled with separating Hidden, but this is not important
Good to change the icon for inactive remotes - the icon is not well designed for dark mode, separate matter

gerhardol (commented):
Briefly reviewed and used, looks good.
OK for me to merge - not enough review to formally approve.

RussKie:
> I am not thrilled with separating Hidden, but this is not important

I don't agree on "not important" bit.
What would you consider a better user experience that would make the difference more obvious for an average user? 
I tried icons, but on their own they didn't strike as "obvious", and there was still visual clutter that made it difficult to see only branches that available to me in the grid.
I also tried the strikeout styling - but it too felt as conveying wrong message - strikeout generally use to mark deleted items...
I suppose this leave a font option still on the table - i.e. grey out refs that are currently unavailable 🤔 

What other options do we have?

Also I think the branches toolstrip dropdown needs to participate in this UX too, i.e. have a "Hidden" menu which contains all unavailable items (if we do with the current proposal).
(I don't know if it is possible to change a colour of a toolstrip menu text. The docs appear to state it is possible.)

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -209,6 +209,22 @@ public FormBrowse([NotNull] GitUICommands commands, string filter, ObjectId sele
                 _selectedRevisionUpdatedTargets = UpdateTargets.None;
                 RefreshSelection();
             };
+            RevisionGrid.RevisionGraphLoaded += (sender, e) =>
+            {
+                if (/* The event is not originated from the revision grid */sender is null
```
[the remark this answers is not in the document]
RussKie: Because of this:
image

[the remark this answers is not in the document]
gerhardol: While I do not like the rule in this situation, I even more dislike the workaround, this is hard to read.

I suggest to put the comment on a separate line instead

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
gerhardol: I meant making this an else if (ParentPath)

gerhardol:
> > I am not thrilled with separating Hidden, but this is not important
> 
> I don't agree on "not important" bit.

We had that discussion earlier in the thread:
https://github.com/gitextensions/gitextensions/pull/8414#issuecomment-683331907

I find that moving around the hidden items is confusing.
No icon or a different icon fits me better, but I can live with [Hidden].

We can change this later too.

RussKie:
> We can change this later too.

There's a lot of new plumbing to facilitate new "Hidden" node, which I wouldn't want to add, if we're not totally happy with the UX. Personally I would it easier to contextualise a filter, but I'm happy to explore the design further.

RussKie:
As an alternative option:

image


@NikolayXHD I think we may have a bug - in the dark mode the colour settings are stomped over:
image

Can you give me a hand here?

<details>
<summary>Diff</summary>

```diff
diff --git a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
index 777b05c13..ec688edab 100644
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
@@ -7,6 +7,7 @@
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands.Git;
+using GitExtUtils.GitUI.Theming;
 using GitUI.BranchTreePanel.Interfaces;
 using GitUI.Properties;
 using GitUI.UserControls.RevisionGrid;
@@ -59,6 +60,16 @@ protected BaseBranchNode(Tree tree, string fullPath, bool visible)
             /// </summary>
             public bool Visible { get; }
 
+            protected override void ApplyStyle()
+            {
+                base.ApplyStyle();
+
+                if (!Visible)
+                {
+                    TreeViewNode.ForeColor = Color.Silver.AdaptTextColor();
+                }
+            }
+
             public override int GetHashCode() => FullPath.GetHashCode() ^ Visible.GetHashCode();
 
             public override bool Equals(object obj)
@@ -181,6 +192,11 @@ protected override void ApplyStyle()
             {
                 base.ApplyStyle();
                 SetNodeFont(IsActive ? FontStyle.Bold : FontStyle.Regular);
+
+                if (!Visible)
+                {
+                    TreeViewNode.ForeColor = Color.Silver.AdaptTextColor();
+                }
             }
```

</details>

NikolayXHD:
> Can you give me a hand here?

Looking at it

NikolayXHD:
> > Can you give me a hand here?
> 
> Looking at it

Done, https://github.com/RussKie/gitextensions/pull/16 @RussKie

RussKie:
As an alternative option:

image
image

Thoughts?

gerhardol:
I prefer this, looks good.

gerhardol:
Time to take this out of WIP?

RussKie:
Once I can get head above water with work...

gerhardol:
> Once I can get head above water with work...

No rush, I just do not think that there will be much more responses

ivangrek:
> No rush, I just do not think that there will be much more responses

#6339 :eye: 
And some others. 😄

RussKie:
Yes, Ivan, I'm acutely aware of the backlog I'm causing...

ivangrek:
> Yes, Ivan, I'm acutely aware of the backlog I'm causing...

It's ok

RussKie:
Ready for thrashing 😉

RussKie:
@ivangrek your reviews are equally welcome

ivangrek:
- Filter by branch
- Clear filter
- Select "Show filtered branches"

image

- Filter by commit message
- Select "Show filtered branches"
- Select "Show all branches"

image

RussKie:
This looks expected, if branches are not currently visible in the revision
grid.
That is, show all branches won't show all if there's an active filter.

...or is there something else that I'm missing?

ivangrek:
> ...or is there something else that I'm missing?

1 != Select "Show filtered branches"
2 != Filter by commit message

After completing the steps, the state before is not equal to the state after.

RussKie:
Thank you

> * Filter by branch
> * Clear filter
> * Select "Show filtered branches"

This looks like a bug - when the filter is reset in the UI, it isn't getting truly reset, and the new functionality exposes the bug:
https://github.com/gitextensions/gitextensions/blob/4b8024547f58f96802475cf5b448bccb887fb14c/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs#L513-L523



My local tests suggest the following fixes it:
```diff
diff --git a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
index 27172d635..480a8add4 100644
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -522,9 +522,10 @@ public bool SetAndApplyBranchFilter(string filter)
             {
                 AppSettings.BranchFilterEnabled = true;
                 AppSettings.ShowCurrentBranchOnly = false;
-                _revisionFilter.SetBranchFilter(filter);
             }
 
+            _revisionFilter.SetBranchFilter(filter);
+
             SetShowBranches();
             return true;
         }
```

ivangrek:
Like in https://github.com/gitextensions/gitextensions/pull/8485#issuecomment-699640469

Show filtered branches

Not equals

Show this branch only
Show all branches
Clear branches filter
Show filtered branches

image

No branches in the grid but shows all branches in the left panel.

GitUI/Strings.cs
```diff
@@ -22,6 +22,7 @@ internal sealed class Strings : Translate
         private readonly TranslationString _containedInNoBranchText = new TranslationString("Contained in no branch");
         private readonly TranslationString _containedInTagsText = new TranslationString("Contained in tags:");
         private readonly TranslationString _containedInNoTagText = new TranslationString("Contained in no tag");
+        private readonly TranslationString _invisibleCommitText = new TranslationString("'{0}' is not currently visible");
```
mstv, line 25:
```suggestion
        private readonly TranslationString _invisibleCommitText = new TranslationString("'{0}' is currently not visible");
```

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
```diff
@@ -53,6 +38,44 @@ protected BaseBranchNode(Tree tree, string fullPath)
...
+            /// Full path of the branch. <example>"issues/issue1344"</example>.
+            /// </summary>
+            public string FullPath => ParentPath.Combine(PathSeparator.ToString(), Name);
+
+            /// <summary>
+            /// Gets whether the commit that the node represents is currently visible in the revision grid.
+            /// </summary>
+            public bool Visible { get; }
+
+            protected override void ApplyStyle()
```
mstv, line 63: Just curious: Do we sort members alphabetically regardless of their visibility now?

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -115,11 +124,12 @@ private abstract class Tree : IDisposable
...
-            {
-                Detached();
-                _reloadCancellationTokenSequence.Dispose();
-            }
+            // A flag to indicate whether the data is currently being filtered or not.
+            // This helps to reduce unnecessary treeview rebinds.
+            private bool _isCurrentlyFiltering;
+
+            // A flag to indicate whether the data is being filtered (e.g. Show Current Branch Only).
+            private protected static AsyncLocal<bool> IsApplyFiltering = new AsyncLocal<bool>();
```
mstv, line 132: The identifier `IsApplyFiltering` is stranger English than mine.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -9,50 +9,61 @@
...
                 await TaskScheduler.Default;
                 token.ThrowIfCancellationRequested();
-                var nodes = new Nodes(this);
-                var pathToNodes = new Dictionary<string, BaseBranchNode>();
 
-                IEnumerable<IGitRef> branches = Module.GetRefs(tags: true, branches: true)
-                    .Where(branch => branch.IsRemote && !branch.IsTag);
+                if (!IsApplyFiltering.Value || _loadedBranches is null)
+                {
+                    _loadedBranches = Module.GetRefs(tags: true, branches: true).Where(branch => branch.IsRemote && !branch.IsTag).ToList();
```
mstv, line 57:
```suggestion
                    _loadedBranches = Module.GetRefs(tags: true, branches: true).Where(branch => branch.IsRemote && !branch.IsTag).ToList();
                    token.ThrowIfCancellationRequested();
```

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Remotes.cs
```diff
@@ -257,15 +260,28 @@ public bool FetchAndRebase()
...
 
+        [DebuggerDisplay("Remote = {_remote.Name}, FullPath = {FullPath}")]
         private sealed class RemoteRepoNode : BaseBranchNode
         {
             private readonly Remote _remote;
             private readonly IConfigFileRemoteSettingsManager _remotesManager;
 
             public RemoteRepoNode(Tree tree, string fullPath, IConfigFileRemoteSettingsManager remotesManager, Remote remote, bool isEnabled)
-                : base(tree, fullPath)
+                : base(tree, fullPath, true)
```
mstv, line 284:
```suggestion
                : base(tree, fullPath, visible: true)
```

GitUI/BranchTreePanel/RepoObjectsTree.ContextActions.cs
```diff
@@ -320,11 +327,24 @@ private void contextMenu_Opening(object sender, CancelEventArgs e)
...
+
             menu.SuspendLayout();
-            int index = insertAfter is null ? 0 : Math.Max(0, menu.Items.IndexOf(insertAfter) + 1);
-            items.ForEach(item => menu.Items.Insert(index++, item));
+
+            int index;
+            if (insertBefore != null)
+            {
+                index = Math.Max(0, menu.Items.IndexOf(insertBefore) - 1);
+                items.ForEach(item => menu.Items.Insert(index + 1, item));
```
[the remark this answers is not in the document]
RussKie: Correct, items are added in reverse - they are added above a specific item. I'll see if I can get it to add in direct order

GitUI/Strings.cs
RussKie: I'm not a native speaker, but to me this sounds odd. This kind of explains it: https://ell.stackexchange.com/a/228812

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
RussKie: "Now" as in - in the official public release - or as in - in the new implementation?

Sorting was introduced in https://github.com/gitextensions/gitextensions/pull/8427, and this is available in the Dev stream. The official version has only one sorting mode - A-Z.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
RussKie: Aye, not sure what I was thinking there...

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
mstv: > "Now" as in - in the official public release - or as in - in the new implementation?

Oh, this was confusing in two ways!
I meant "newly", i.e. the latter.
But the question was about the coding style whether to sort class members ignoring `private`, `protected`, ...

RussKie: The general rule I follow (almost OCD about this) is A-Z order within groups as per the original [MS Design Guidelines](https://docs.microsoft.com/en-us/archive/blogs/brada/internal-coding-guidelines#28-file-organization), i.e.: 
```cs
public class MyClass : IFoo
{
      // fields
      int foo;

      // constructors
      public MyClass() { … }

      // properties
      public int Bar { get { … } set { … } }
      public int Foo { get { … } set { … } }
      public int Rex { get { … } set { … } }

      // events
      public event EventHandler FooChanged { add { … } remove { … } }

      // methods
      public void DoSomething() { … }
      public void FindSomethind() { … }

      protected void FindSomethindProtected() { … }

      private void DoSomethingPrivate() { … }

      //private interface implementations
      void IFoo.DoSomething() { DoSomething(); }

      // nested types
      class NestedType { … }
}
```

But our codebase has very diverse set of ordering practices, and even our dotnet/* repos no longer follow a single ordering style (which drives me crazy, hehe).

RussKie: At the same time, when making changes to the code I have to leave members in their place to provide a meaningful diff.
I will clean up the code once this work is complete.

RussKie:
Ready for another review.
I believe this work is complete (modulo https://github.com/gitextensions/gitextensions/pull/8414#discussion_r495617242, this will be done separately later).

mstv (approved):
LGTM, have not run it

PR #8424, opened 2020-08-29
.css inheritance in themes

NikolayXHD:
## Proposed changes

- parse css using [ExCSS](https://github.com/TylerBrinks/ExCSS) v2 package (v3 is unusable, e.g. see [this issue](https://github.com/TylerBrinks/ExCSS/issues/94)). The motivation is to introduce freedom in editing .css
  - support comments, line breaks, indents
  - ~~support alternative color definitions, like e.g. `hsl(60,100%,25%)` which I find more suitable for color transformations than hexadecimal rgb~~. ExCss v2 does not support css3 which introduces hsl colors.
- use .css inheritance to introduce scheme variations
- add "Colorblind" checkbox to ColorSettings page to enable / disable Colorblind variation

## Screenshots
image

## Test methodology

manually

## Test environment(s)

Win10

----


gerhardol:
@itsho @neil5280 Comments?

GitCommands/Settings/AppSettings.cs
```diff
@@ -1247,6 +1247,18 @@ public static ThemeId ThemeId
             }
         }
 
+        public static string[] ThemeVariations
```
gerhardol, line 1250: ??? Can the variation be handled as a new theme instead?
It can be confusing if a theme has no colrorblindness variation, this will have no effect. 
It will be easier to add variants of exiting themes too, like dark vs darksilver (most fields differe there though).
I do not want to overwork this functionality, but "automatic variations" is slightly better in my opinion.

GitUI/Theming/ThemePersistence.cs
```diff
@@ -57,89 +53,124 @@ public void Save(Theme theme, string fileName)
...
+            var fileInfo = new FileInfo(fileName);
+
+            if (!fileInfo.Exists)
+            {
+                PrintTraceWarning(fileName, _fileNotFound.Text);
+                return false;
+            }
+
+            // > 1MB
+            if (fileInfo.Length > (1 << 20))
```
gerhardol, line 68: nit While I have no problem with this myself, regular C# programmers may have a problem understanding...

GitUI/Themes/bright.css
```diff
@@ -40,3 +40,8 @@
 .DiffRemovedExtra { color: #ffa5a5 }
```
gerhardol, line 40: Why keep both bright.css and invariant.css? I do not see bright.css used?

gerhardol (commented):
Looks good - some discussion points (no testdrive, not checked the ExCSS package)

GitUI/Theming/ThemePersistence.cs
NikolayXHD: true :)

GitUI/Themes/bright.css
NikolayXHD: "invariant.css" is not the theme which gets applied if the user chooses "default" theme, but rather a technical theme, user cannot choose it. The purpose for invariant theme is to provide a fixed set of colors, to compare against other themes.

When user choses "default" theme, GitExtensions will not apply any hacks to tweak colors, so we are guaranteed to get what we see in WinForms designer. As a result the "default" theme cannot have a colorblind variation.

Therefore to provide a colorblind variation of non-dark theme, we need to introduce a non-default theme, which user can choose and then have the option to check "Colorblind" flag.

NikolayXHD: > I do not see bright.css used?

bright.css can be chosen in Theme menu

GitCommands/Settings/AppSettings.cs
NikolayXHD: The point is to avoid having colorblindness as another theme, so that we do not end up with 6 themes instead of 3 (bright, dark, darksilver).

> It can be confusing if a theme has no colrorblindness variation, this will have no effect.

My idea is, at the moment all themes do have the colorblind variation, so I did not bother to hide or show the checkbox depending on what classes the actual .css provides, just wait until this is actually required.

GitUI/Themes/bright.css
```diff
@@ -40,3 +40,8 @@
...
 .DiffAdded { color: #c8ffc8 }
 .DiffAddedExtra { color: #87ff87 }
+
+.Graph.colorblind { color: #0500a8 }
```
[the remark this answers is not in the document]
NikolayXHD: ok

GitCommands/Settings/AppSettings.cs
NikolayXHD: > "automatic variations"  is slightly better in my opinion

I tried automatic variations to replace red with blue in editor syntax highlighting schemes, and I find the result somewhat unsatisfying. This is why I also kept manually defined red->blue replacements instead of having transformation formulae in the code. I just created a PR #8426 to demonstrate what I mean.

So the issue that bothers me with automatic colorblind transformation, is that the changes nessessarily need to affect a relatively wide range of colors and I have no confidence, that the modification will be for the better in all cases.

NikolayXHD: maybe it's not that bad (automatic colorblind adjustment) , at least the build status colors is modified as expected

before
image

after
image

GitUI/Themes/bright.css
gerhardol: The file is not included with the installer (should it be somewhere else too?)
otherwise OK

itsho:
Brilliant. Thank you!

GitCommands/Settings/AppSettings.cs
gerhardol: I meant more that if you have a file mytheme.css with a Branch.colorblind and Branch.myvariant, the themes offered are mytheme, mytheme-colorblind, mytheme-myvariant

#8426 is maybe contradicting this view, if there are special handling in other situations too - adding to the changes in the theme itself. This could be handled with a convention (like variant name is colorblind), but a setting is more natural. 
The importance for #8426 (as an extra step) is a separate question. I would expect that the change is welcome. It does not look good to me, but if it makes the UI usable for 1% of users, it could be worth it.

GitUI/Themes/bright.css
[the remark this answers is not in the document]
gerhardol: Now DiffAdded and DiffRemoved are separated...

GitUI/Themes/bright.css
NikolayXHD: You are right, I forgot it, thanks for the vigor!

neil5280:
> @itsho @neil5280 Comments?

I have been on vacation.  I appreciate being called out for comments.  All this work is fantastic; I appreciate that it was not scoped and took resources away from other improvements.  Approximately 8% of men and .5% of women users stand to benefit from these adjustments, so thank you all for making time for accessibility improvements.

I am looking forward to seeing this in the next release, and I will endeavor to find a way to contribute in the next year, rather than being just a noisy bystander.

Many thanks, once again.

pmiossec:
What I am more interested is supporting import to update just some values (to support for example #8432) and make theme creation or maintenance easier.

I did a quick 15min PoC using `@import url("parent.css");` syntax supported by ExCss and it is working. See https://github.com/pmiossec/gitextensions/tree/css_import

@NikolayXHD Side question: Are themes only loaded from the application theme folder? or also from a user folder? If yes, which one?

NikolayXHD:
> @NikolayXHD Side question: Are themes only loaded from the application theme folder? or also from a user folder? If yes, which one?

%AppData%\GitExtensions\GitExtensions\Themes

GitUI/Themes/bright.css
```diff
@@ -1,4 +1,5 @@
-﻿.ActiveBorder { color: #b4b4b4 }
+﻿/* System colors */
+.ActiveBorder { color: #b4b4b4 }
```
NikolayXHD, line 2: - [ ] include the file in installer

GitUI/Themes/bright.css
NikolayXHD: > What I am more interested is supporting import to update just some values (to support for example #8432) and make theme creation or maintenance easier.

Good point, feel free to add your commit to this PR so our changes get to be reviewed and delivered as 1 package

pmiossec: >Good point, feel free to add your commit to this PR so our changes get to be reviewed and delivered as 1 package

done.
I have also added some comments  in the css files to explain a little more how to create or modify a theme.
Feel free all to correct my bad English sentences...

NikolayXHD: Thank you

NikolayXHD: > Feel free all to correct my bad English sentences...

Says one non-english speakers to another. We are blessed people, for native english the code is like:
```
si (fenêtre.estUverte()) {
  retourner vérité;
} sinon {
  retourner faux;
}
```

pmiossec (approved):
It seems to work as expected.

NikolayXHD:
I still have a couple of things to do here, probably tomorrow I get them done and merge.

GitUI/Themes/bright.css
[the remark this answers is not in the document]
NikolayXHD: ok, let them be together

GitUI/Theming/ThemeCssLoader.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            if (!fileInfo.Exists)
+            {
+                PrintTraceWarning(fileName, _fileNotFound.Text);
+                return false;
+            }
+
+            if (fileInfo.Length > 1024 * 1024)
+            {
+                PrintTraceWarning(fileName, _fileTooLarge.Text);
```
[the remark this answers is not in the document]
RussKie: Overall what was the reason to use this approach instead of throwing?
E.g.
```cs
        private string ReadThemeFile(string themeFileName)
        {
            if (fileInfo.Length > 1024 * 1024)
            {
                throw new FileLoadException(_fileTooLarge.Text, themeFileName);
            }

            return File.ReadAllText(themeFileName);
        }
```

GitUI/Theming/ThemeCssLoader.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                throw new InvalidOperationException($"{nameof(ThemeCssLoader)} only supports 1 call to {nameof(TryLoadCss)}");
+            }
+
+            _parseCalled = true;
+
+            return TryLoadCssImpl(filePath, cssImportChain: Array.Empty<string>());
+        }
+
+        private bool TryLoadCssImpl(string filePath, string[] cssImportChain)
```
RussKie, line 57: Since all failures appear to be terminal, I think there is little value in "TryXxx" methods, and we may as well throw all the way through the stack. This way the code would be terse, and we'd be able to bubble errors up to the UI layer, and show error messages to the user to action.

GitUI/Theming/ThemeRepository.cs
```diff
@@ -8,16 +8,19 @@
 
 namespace GitUI.Theming
 {
-    public class ThemeRepository
+    public class ThemeRepository : ICssUrlResolver
     {
         private const string Subdirectory = "Themes";
         private const string Extension = ".css";
+        private const string CssVariableUserThemesDirectory = "$user-defined/";
```
RussKie, line 15: For scripts we use `{variable}` notation, is it possible to use the same, e.g. `{UserAppData}/`?

GitUI/Theming/ThemeRepository.cs
```diff
@@ -86,19 +89,58 @@ private IEnumerable<ThemeId> GetUserCustomizedThemeIds()
...
             if (id.IsBuiltin)
             {
-                return Path.Combine(AppThemesDirectory, id.Name + Extension);
+                path = Path.Combine(AppThemesDirectory, id.Name + Extension);
             }
+            else
+            {
+                if (UserThemesDirectory == null)
+                {
+                    throw new ThemeNotFoundException("There is no directory for custom user themes in portable mode");
```
RussKie, line 124: I think IOE is more appropriate here:
```suggestion
                    throw new InvalidOperationException("Portable mode only supports local themes");
```
Alternatively, may be `FileLoadException`?

GitUI/Theming/ThemeRepository.cs
```diff
@@ -86,19 +89,58 @@ private IEnumerable<ThemeId> GetUserCustomizedThemeIds()
...
             {
-                throw new InvalidOperationException("There is no directory for custom user themes in portable mode");
+                throw new ThemeNotFoundException($"Theme file not found: {path}");
             }
 
-            return Path.Combine(UserThemesDirectory, id.Name + Extension);
+            return path;
+        }
+
+        private class ThemeNotFoundException : Exception
```
RussKie, line 138:
```suggestion
        private class ThemeNotFoundException : FileNotFoundException
```
https://docs.microsoft.com/dotnet/api/system.io.filenotfoundexception.-ctor#System_IO_FileNotFoundException__ctor_System_String_System_String_

Setup/Product.wxs
```diff
@@ -586,15 +586,21 @@
       </Component>
     </DirectoryRef>
```
RussKie, line 587: Include ExCSS.dll?

GitUI/CommandsDialogs/SettingsDialog/Pages/ColorsSettingsPage.cs
```diff
@@ -120,6 +133,12 @@ private void ChkUseSystemVisualStyle_CheckedChanged(object sender, EventArgs e)
             EndUpdateThemeSettings();
         }
 
+        private void ChkColorblind_CheckedChanged(object sender, EventArgs e)
+        {
+            BeginUpdateThemeSettings();
+            EndUpdateThemeSettings();
```
RussKie, line 139: This looks strange. Any reason we can't have a single method?

GitUI/Themes/README.md
```diff
@@ -0,0 +1,23 @@
+﻿GitExtensions stores themes in *.css files.
```
RussKie, line 1: The readme is a great idea, but it needs some work.

It would be great to add sections, may be something like the following:

```md
# Overview

# Themes locations

### Installed app

* Official themes

...

* Custom themes

...

### Portable app

# How to create custom theme
```

It needs to discuss that there are two distinct modes - installed and portable, and each mode handles themes differently. I.e. the portable mode supports only local themes (those placed in .\Themes) folder, whereas the installed app supports official themes and user-defined themes. The official themes are located in the app installation folder (typically `%ProgramFiles(x86)%\GitExtensions\Themes`), and these mustn't be changed.

RussKie (changes requested):
👍 with some feedback.

Also I think we should layout as follows:
image

Thank you

RussKie:
Btw, cool avatar 👍

pmiossec:
Do you think it could be interesting to add a split button next to the dropdownlist to be able to open the "Application theme" and the "user theme" folders?

RussKie:
I think this may be a good idea 👍

GitUI/Themes/bright.css
```diff
@@ -0,0 +1,13 @@
+﻿/* see Themes\README.md (this directory) to know how to create or modify a theme */
```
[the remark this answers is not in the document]
NikolayXHD: ok

GitUI/Themes/invariant.css
```diff
@@ -1,4 +1,9 @@
-﻿.ActiveBorder { color: #b4b4b4 }
+﻿/* !!! If you want to customize GitExtensions appearance, do not edit this file !!! */
+
+/* see Themes\README.md (this directory) to know how to create or modify a theme */
```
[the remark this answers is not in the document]
NikolayXHD: ok

GitUI/Theming/ThemeCssLoader.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                _sysColors[sysColorName] = color.Value;
+                return true;
+            }
+
+            PrintTraceWarning(filePath, string.Format(_invalidRule.Text, rule.Value));
+            return false;
+        }
+
+        private string[] TryGetClassNames(StyleRule rule)
```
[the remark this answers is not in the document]
NikolayXHD: ok

GitUI/Theming/ThemeCssLoader.cs
```diff
@@ -0,0 +1,216 @@
...
+            result = null;
+            var fileInfo = new FileInfo(fileName);
+
+            if (!fileInfo.Exists)
+            {
+                PrintTraceWarning(fileName, _fileNotFound.Text);
+                return false;
+            }
+
+            if (fileInfo.Length > 1024 * 1024)
```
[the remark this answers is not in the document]
NikolayXHD: ok

Setup/Product.wxs
NikolayXHD: ok

GitUI/CommandsDialogs/SettingsDialog/Pages/ColorsSettingsPage.cs
NikolayXHD: ok

GitUI/Theming/ThemeCssLoader.cs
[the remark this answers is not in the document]
NikolayXHD: Because the theme is loaded when starting up the application, there is no color settings page to display the error yet.

GitUI/Theming/ThemeCssLoader.cs
NikolayXHD: ok

GitUI/Theming/ThemeRepository.cs
NikolayXHD: ok

GitUI/Theming/ThemeRepository.cs
NikolayXHD: ok, InvalidOperationException

GitUI/Theming/ThemeRepository.cs
NikolayXHD: kept 2 specialized exceptions

- `CssUrlResolverException`
- `ThemeRepositoryException` any "expected" exception when loading theme, wrapping any io, security and or `CssUrlResolverException`

GitUI/Theming/ThemeCssLoader.cs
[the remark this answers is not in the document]
NikolayXHD: - [ ] load theme when user selects it so that if exception happens, it can be shown before restarting the app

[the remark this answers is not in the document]
RussKie: I'd rather have the app fail with a popup message telling the user there is a problem with a theme (and how to fix it preferably) than silently swallow errors and have the user come here complain about why changes to a theme didn't work.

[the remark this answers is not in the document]
RussKie: These errors would be user-induced, so the user would be in a position to fix them.

GitUI/Themes/README.md
NikolayXHD: ok

GitUI/Theming/ThemeCssLoader.cs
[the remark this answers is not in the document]
NikolayXHD: ok

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
+﻿# Overview
+
+GitExtensions supports color customization, in particular user can choose between
+default (operating-system-defined), bright or dark theme.
```
RussKie, line 4:
```suggestion
Git Extensions supports color customization, specifically a user can choose between
a default (operating system defined), a bright or a dark themes.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+There is a [Wiki page on Dark Theme](https://github.com/gitextensions/gitextensions/wiki/Dark-Theme)
+with some background information and a list of known issues.
+
+# Themes location
+
+Themes are stored in *.css files. Where GitExtensions looks for theme files depends on whether it
+is Installed or Portable version. Both come with a predefined set of themes in
```
RussKie, line 12:
```suggestion
Themes are stored in *.css files. Where Git Extensions looks for theme files depends on whether it
is an 'Installed' or a 'Portable' version. Both come with a predefined set of themes in
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+`(App install)\Themes` directory.
+
+In addition the Installed version looks for user-defined themes in
+`%AppData%\Roaming\GitExtensions\GitExtensions\Themes`
```
RussKie, line 16:
```suggestion
In addition the Installed version looks for user-defined themes in
`%AppData%\Roaming\GitExtensions\GitExtensions\Themes` directory.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+For each found theme file ColorSettings page will show an entry in Theme drop down list.
```
RussKie, line 18:
```suggestion
For each found theme file `ColorSettings` page will show an entry in the Theme dropdown list.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Installed GitExtension application
+
+### Official themes
+
+Installed version of GitExtensions stores themes in `(App install)\Themes` directory.
```
RussKie, line 24:
```suggestion
The Installed version of Git Extensions stores themes in `(App install)\Themes` directory.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Custom themes
+
+Any modifications for personal use should be saved to
+`%AppData%\Roaming\GitExtensions\GitExtensions\Themes`, because any changes to
+`(App install)\Themes` directory will be lost as soon as user updates or reinstalls GitExtensions.
```
RussKie, line 30:
```suggestion
Any modifications for personal use should be saved to
`%AppData%\Roaming\GitExtensions\GitExtensions\Themes` directory, because any changes to
`(App install)\Themes` directory will be lost as soon as user updates or re-installs the app.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Portable application
+
+The portable version of GitExtensions does not have a separate directory for user-defined themes.
+It only uses preinstalled themes from `(App install)\Themes` directory.
```
RussKie, line 35:
```suggestion
The Portable version of Git Extensions does not have a separate directory for user-defined themes.
It only uses pre-installed themes from `(App install)\Themes` directory.
```

GitUI/Themes/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+# How to create a custom theme
+
+Copy `Themes\dark.css` or `Theme\bright.css` and change color values. Use a text editor, preferably
+with .css syntax highlighting and inline color display.
+
+There is a plugin for VisualStudio to show .css colors inline.
```
RussKie, line 42:
```suggestion
There is a plugin for Visual Studio to show .css colors inline.
```

UnitTests/GitUI.Tests/Theming/AppColorDefaultsTests.cs
```diff
@@ -39,17 +39,17 @@ public void Default_values_are_defined_in_AppColorDefaults()
...
+        ////{
+        ////    var repository = new ThemeRepository(new ThemePersistence(), new ThemePathProvider());
+        ////    var invariantTheme = repository.GetInvariantTheme();
+        ////    invariantTheme.Should().NotBeNull();
+        ////    foreach (AppColor name in Enum.GetValues(typeof(AppColor)))
+        ////    {
+        ////        Color value = invariantTheme.GetColor(name);
+        ////        value.Should().NotBe(Color.Empty);
+        ////    }
+        ////}
```
RussKie, line 53: ?

RussKie (commented):
👍 
Would be good to have some test coverage to confirm we haven't broken something.

RussKie (approved):

RussKie:
Please tidy up the history, and merge at will.

NikolayXHD:
> Please tidy up the history, and merge at will.

I am not done with writing tests yet. Let me label PR as `In progress` to avoid confusion.

RussKie:
Make it a draft then

mstv:
> Make it a draft then

IIRC, AppVeyor is not triggered for Draft PRs. That's why I still use "WIP".

RussKie:
> > Make it a draft then
> 
> IIRC, AppVeyor is not triggered for Draft PRs. That's why I still use "WIP".

Just tested it, an update triggered a new build.

GitHub attempted to change this behaviour, i.e. stop triggering builds for draft PRs, but it has broken a number of high profile MSFT repos (that I'm aware of) and GitHub backed the change out.

mstv:
> Just tested it, an update triggered a new build.
> GitHub attempted to change this behaviour, i.e. stop triggering builds for draft PRs, but it has broken a number of high profile MSFT repos (that I'm aware of) and GitHub backed the change out.

Perfect, thank you for the clarification.

RussKie (approved):
There are a few places where we don't validate input parameters, and thus open for NREs. 
Otherwise looks good 👍

RussKie:
MC

GitUI/GitUI.csproj
```diff
@@ -24,6 +24,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="AppInsights.WindowsDesktop" Version="$(AppInsightsWindowsDesktopVersion)" />
+    <PackageReference Include="ExCSS" Version="2.0.6" />
```
RussKie, line 27: @NikolayXHD what was the reason not to use the [latest version 3.0.0](https://www.nuget.org/packages/ExCSS/3.0.0) that supports .NET Standard?
I've tried to update the app the .NET 5.0 and this library caused quite a bit of failures...

gerhardol: This PR description says v3 is unusable, https://github.com/TylerBrinks/ExCSS/issues/94

RussKie:
Thank you. There's a link to another fork that may work...

GitUI/GitUI.csproj
NikolayXHD: ExCSS 3 is unfinished work in many aspects. ExCSS_Core does complete some things, but not all. I gave up on it when it became clear, that color values are not parsed, so while 
v2 produces ready-to-use Color structure
v3 produces a plain text 
```
"#ffffff", "hsl(1.0, ...)", ...
```

NikolayXHD: Had another look at ExCSS-Core now v4, error handling is also incomplete. Parsing errors are silently ignored, there is an event raised when error happens, but no way to subscribe to it using public API

https://github.com/HockeyJustin/ExCSS_Core/blob/9a1114494d87220f38e7a157e3307ae6d4e4efb3/src/ExCSS/Parser/Lexer.cs#L8

image

NikolayXHD: If ExCSS v2 is problematic to use, I believe I can still rollback to manual parsing, since we actually use a very small subset of CSS.

NikolayXHD: Now it feels quite nice to have theme parsing decently covered by tests 😃

RussKie: Oh bugger!

I spent a number of very frustrating hours yesterday trying to update my port to .NET 5.0, and ExCss turned out to be a major hurdle. I commented it out, but since we moved all colour handling to it, I couldn't get any colours working... 😞 

https://github.com/gitextensions/gitextensions/pull/8522

RussKie: > I believe I can still rollback to manual parsing, since we actually use a very small subset of CSS.

When you have time, may be you could work directly off my branch?
Much appreciate it, thank you

RussKie: Hmm, may be we could even fork ExCss 2.x and compile as .NET Standard?

NikolayXHD: If we are OK to fork, then I'd fork ExCSS-Core and fix the things we miss. Later submit a PR, get it merged and replace back the fork with nuget package.

NikolayXHD: :smirk: (imagening hsl-defined colors in theme .css)

RussKie: There - https://github.com/gitextensions/ExCSS, go for your life 😉

NikolayXHD: We should probably fork from another fork https://github.com/HockeyJustin/ExCSS_Core
because the original repo seems abandoned, they won't accept any PRs

https://github.com/TylerBrinks/ExCSS/issues/99
https://github.com/TylerBrinks/ExCSS/issues/102

RussKie: Apologies, I should read more thoroughly...
Here: https://github.com/gitextensions/ExCSS_Core

RussKie: Hey, I just found the parser used by Visual Studion and ASP.NET - https://github.com/dotnet/cssparser. Maybe worth looking at it too?

NikolayXHD: absolutely will have a look at it

NikolayXHD: It's only going to happen the weekend though, struggling at work.

RussKie: Absolutely no pressure, there's a major issue (https://github.com/gitextensions/gitextensions/pull/7162#issuecomment-674834068) in the editor that blocks the migration anyway.

pmiossec: I like the text in the Readme.md:

>This parser is an internal implementation detail of the .NET Core SDK, and is not supported for any other usage. We do not recommend taking a dependency on this code because there is no guarantee of long-term maintenance.

RussKie: Haha, yes I was told there's no official ownership of this repo.
But if it's good for big guys, it is _probably_ good for us. And we can fork it, if necessary.

gerhardol: ExCSS core is now resurrected, v4 has been released.

NikolayXHD: I've looked at it some time ago, It's still pretty much unusable.

> Had another look at ExCSS-Core now v4, Concrete rule types which provide access to specific parsed values are all internal
https://github.com/HockeyJustin/ExCSS_Core/blob/master/src/ExCSS/Rules/ImportRule.cs

gerhardol:
@NikolayXHD This PR changed the color for OtherTag in dark theme (the only I checked) from #ff8040 to #a1a1a1.
This is only used for artificial commits (gray in master) and stash/bisect (not shown at all in master, #8879)

Was the color change intentional?

image

image

pmiossec:
> @NikolayXHD This PR changed the color for OtherTag in dark theme (the only I checked) from #ff8040 to #a1a1a1.
> This is only used for artificial commits (gray in master) and stash/bisect (not shown at all in master, #8879)
> 
> Was the color change intentional?
> 

If the color should be restored, I imagine that we will give it a greenish color as the local and remote colors have been switched (and that the local branches are green)...

RussKie:
The original was orange (or orangish), I never perceived it to be tied to the remote colour in anyway.
On the screenshot the colours look washed out, where as on my monitor they look quite apart. Obviously YMMV.

My personal preference is to restore the orange colour. But I'm not too opposed to consider a different colour as long it isn't tied to local-remote (red-green) colour scheme.

NikolayXHD:
> Was the color change intentional?

Yes. A ghost-like gray color was supposed to reflect the ephemeric nature of artificial commits.

If I remember correctly at the time it looked to me as restoring consistensy between the light and the dark theme.

1. Initial implementation of dark theme, dark colors initially appear based on light colors
2. Someone changes artificial commit for light theme to gray.
3. I notice this while working on this PR and do the same for dark.

RussKie:
Personally I'd like a more "positive" colour. What does everyone think?

gerhardol:
While the reasoning for ghost-like is fine, I want current HEAD to stand out. I can live with grey too. ivangrek wanted to remove bold in the revgrid for current, this was voted down.
For stash, grey is probably preferred.

However, for bisect, is grey visible enough? (I do not think we want to handle more colors though.)

NikolayXHD:
> I want current HEAD to stand out

True, but artificial commits are *near*, not *at* the current head in the graph. If anything making them prominent would distract from the current head.

> However, for bisect, is grey visible enough? (I do not think we want to handle more colors though.)

Have to agree, ghost-likness argument does not apply here.

It's not that I have a strong preference either, I am fine with reverting to orange, in all themes for consistency sake.

RussKie:
> It's not that I have a strong preference either, I am fine with reverting to orange, in all themes for consistency sake.

Revert to "bisect" and "stash" only?

NikolayXHD:
> Revert to "bisect" and "stash" only?

That would be ideal

PR #8309, opened 2020-07-12
WIP Parse dereferenced tags

mstv:
Fixes #8301

## Proposed changes

- get tags using `git show-ref --tags --dereference`
- parse dereferenced tags

## Test methodology 

- select `refs/tags/v2.48` in `GoToCommitForm`
- TODO: add NUnit test

## Test environment(s) 

- Git Extensions 33.33.33
- Build 850153e311703ab052bc9ee076fab650e5b97542
- Git 2.27.0.windows.1
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8427, opened 2020-08-29
Fix 6310 Provide ability to sort/order branches and tags

### LINKED ISSUE #6310: Sort branches alphabethical/lastAccess was removed
## Current behaviour

In 2.51, it was possible to order branches in Browse by either alphabethical or last access order.
This possibility was removed in 9c7c76fb619480429cbe02e11ce889b00423eac2 (likely #5087)

## Expected behaviour

Change of sorting was useful, even little hard to find.
Sort could be restored or we decide that it should not be possible.

## Steps to reproduce

       public static BranchOrdering BranchOrderingCriteria
        {
            get { return GetEnum("BranchOrderingCriteria", BranchOrdering.ByLastAccessDate); }
            set { SetEnum("BranchOrderingCriteria", value); }
        }

## Screenshots

image

2.51

## Environment

- Git Extensions 3.1.0
- Build d753f4a7e0c175529d7c6ecffa7f297ea239faf3
- Git 2.20.1.windows.1
- .NET Framework 4.7.3324.0
- DPI 96dpi (no scaling)

### LINKED ISSUE #7472: Set minimum git version to 2.19  or later
## Feature description

With the move to .NET Core I propose we set the minimum git version to 2.19 or newer, and remove all older versions and their quirks.

Git 2.19 was release Sep 2018, .NET Core was released Sep/Dec 2019, so a user will likely need to have .NET Core runtime installed to run a new version of the app (we may or may not publish a self contained bundle).

The move will help providing simple solutions to #4869 and #7456.

### LINKED ISSUE #8301: "Go to tag" is broken
## Current behaviour

"Navigate -> Go to commit -> Go to tag" does nothing.
The list of tags is shown, but GitExtensions does not navigate to the tag I select.

## Expected behaviour
The selected tag is navigated to.

## Steps to reproduce

See above.

This has been broken since at least v3.3.x.

## Environment

- GitExtensions version: 3.4.2.9737
- GIT version: 2.23.0
- OS version: Win10 build 18363 (release 1909)
- .NET version: 4.8.4180.0

RussKie:
Fixes #6310
Fixes #8301
Resolves #7472
Closes #4869
Closes #8309 as obsolete

## Proposed changes

- Provide ability to sort/order branches

Restore the functionality (albeit improved) removed in 9c7c76fb619480429cbe02e11ce889b00423eac2.

Allow to either defer sorting to git (i.e. sort by Default) which is either implicit, or configured via `git config branch.sort`.

Alternatively expose the following sort by options:
- authordate
- committerdate
- creatordate
- taggerdate
- refname
- objectsize
- upstream

For user-specified sort, allow to select ascending or descending order (default: desc).

Git v2.19+ is required to facilitate the above, hence bump in the min version.

The sort configuration applies to both the left panel and the branches dropdown in the toolstrip.

### After

* Settings:
image
image

* The left panel:
image

## Test methodology 

- manual 
- added few unit tests

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using GitCommands.Git;
 using GitExtUtils;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
@@ -81,39 +80,59 @@ namespace GitCommands.Git.Commands
                 };
         }
 
-        public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks)
+        public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
         {
-            GitArgumentBuilder cmd;
-
-            if (tags)
+            string format;
+            if (!tags)
             {
-                cmd = new GitArgumentBuilder("show-ref", gitOptions:
-                    noLocks && GitVersion.Current.SupportNoOptionalLocks
-                        ? (ArgumentString)"--no-optional-locks"
-                        : default)
-                {
-                    { branches, "--dereference", "--tags" },
-                };
-            }
-            else if (branches)
-            {
-                // branches only
-                cmd = new GitArgumentBuilder("for-each-ref", gitOptions:
-                    noLocks && GitVersion.Current.SupportNoOptionalLocks
-                        ? (ArgumentString)"--no-optional-locks"
-                        : default)
-                {
-                    "--sort=-committerdate",
-                    @"refs/heads/",
-                    @"--format=""%(objectname) %(refname)"""
-                };
+                // If we don't need tags, it is easy.
+                format = @"--format=""%(objectname) %(refname)""";
             }
             else
             {
-                throw new ArgumentException("GetRefs: Neither branches nor tags requested");
+                // ...however, if we're interested in tags, tags may be simple (in which case they are point to commits directly),
+                // or "dereferences" (i.e. commits that contian metadata and point to other commits, "^{}").
+                // Derefence commits do not contain date information, so we need to find information from the referenced commits (those with '*').
+                // So the following format is as follows:
+                //      If (there is a 'authordate' information, then this is a simple tag/direct commit)
+                //      Then
+                //          format = %(objectname) %(refname)
+                //      Else
+                //          format = %(*objectname) %(*refname) // i.e. info from a referenced commit
+                //      End
+                format = @"--format=""%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)""";
             }
 
+            GitArgumentBuilder cmd = new GitArgumentBuilder("for-each-ref",
+                gitOptions: noLocks && GitVersion.Current.SupportNoOptionalLocks
+                    ? (ArgumentString)"--no-optional-locks"
+                    : default)
+            {
+                { sortBy != GitRefsSortBy.Default, GetSortCriteria(tags, sortBy, sortOrder), string.Empty },
+                format,
+                { branches, "refs/heads/", string.Empty },
+                { branches && tags, "refs/remotes/", string.Empty },
+                { tags, "refs/tags/", string.Empty },
+            };
+
             return cmd;
+
+            static string GetSortCriteria(bool needTags, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
+            {
+                if (!GitVersion.Current.SupportRefSort)
+                {
+                    return string.Empty;
+                }
+
+                string order = sortOrder == GitRefsSortOrder.Ascending ? string.Empty : "-";
+                if (!needTags)
+                {
+                    return $"--sort={order}{sortBy}";
+                }
+
+                // Sort by dereferenced data
+                return $"--sort={order}{sortBy} --sort={order}*{sortBy}";
+            }
         }
 
         public static ArgumentString RevertCmd(ObjectId commitId, bool autoCommit, int parentIndex)
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2895,51 +2895,16 @@ namespace GitCommands
 
         public IReadOnlyList<IGitRef> GetRefs(bool tags = true, bool branches = true)
         {
-            return GetRefs(tags, branches, false);
-        }
+            // We do not want to lock the repo for background operations.
+            // The primary use of 'noLocks' is to run git-status the commit count as a background operation,
+            // but to run the same in a foreground for FormCommit.
+            //
+            // Assume that all GetRefs() are done in the background, which may not be correct in the future.
+            const bool noLocks = true;
 
-        public IReadOnlyList<IGitRef> GetRefs(bool tags, bool branches, bool noLocks)
-        {
-            var refList = _gitExecutable.GetOutput(GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks: noLocks));
-
+            string cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            var refList = _gitExecutable.GetOutput(cmd);
             return ParseRefs(refList);
-        }
-
-        /// <param name="option">Order by date is slower.</param>
-        public IReadOnlyList<IGitRef> GetTagRefs(GetTagRefsSortOrder option)
-        {
-            var list = GetRefs(true, false);
-
-            switch (option)
-            {
-                case GetTagRefsSortOrder.ByCommitDateAscending:
-                    return list.OrderBy(GetDate).ToList();
-                case GetTagRefsSortOrder.ByCommitDateDescending:
-                    return list.OrderByDescending(GetDate).ToList();
-                default:
-                    return list;
-            }
-
-            // BUG this sorting logic has no effect as CommitDate is not set by the GitRevision constructor
-            DateTime GetDate(IGitRef head) => new GitRevision(head.ObjectId).CommitDate;
-        }
-
-        public enum GetTagRefsSortOrder
-        {
-            /// <summary>
-            /// default
-            /// </summary>
-            ByName,
-
-            /// <summary>
-            /// slower than ByName
-            /// </summary>
-            ByCommitDateAscending,
-
-            /// <summary>
-            /// slower than ByName
-            /// </summary>
-            ByCommitDateDescending
         }
 
         public async Task<string[]> GetMergedBranchesAsync(bool includeRemote = false, bool fullRefname = false, string commit = null)
--- a/GitCommands/Git/GitVersion.cs
+++ b/GitCommands/Git/GitVersion.cs
@@ -17,13 +17,14 @@ namespace GitCommands
         private static readonly GitVersion v2_7_0 = new GitVersion("2.7.0");
         private static readonly GitVersion v2_9_0 = new GitVersion("2.9.0");
         private static readonly GitVersion v2_11_0 = new GitVersion("2.11.0");
+        private static readonly GitVersion v2_14_6 = new GitVersion("2.14.6");
         private static readonly GitVersion v2_15_0 = new GitVersion("2.15.0");
         private static readonly GitVersion v2_15_2 = new GitVersion("2.15.2");
         private static readonly GitVersion v2_19_0 = new GitVersion("2.19.0");
         private static readonly GitVersion v2_20_0 = new GitVersion("2.20.0");
 
-        public static readonly GitVersion LastSupportedVersion = v2_11_0;
-        public static readonly GitVersion LastRecommendedVersion = new GitVersion("2.25.1");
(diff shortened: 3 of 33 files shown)


RussKie:
@msftbot merge in 48 hours

GitCommands/Git/GitModule.cs
```diff
@@ -2861,41 +2861,42 @@ public RemoteActionResult<IReadOnlyList<IGitRef>> GetRemoteServerRefs(string rem
...
-                    noLocks && GitVersion.Current.SupportNoOptionalLocks
-                        ? (ArgumentString)"--no-optional-locks"
-                        : default)
+                cmd = new GitArgumentBuilder("for-each-ref",
+                        gitOptions: noLocks && GitVersion.Current.SupportNoOptionalLocks
+                            ? (ArgumentString)"--no-optional-locks"
+                            : default)
                 {
-                    "--sort=-committerdate",
+                    { sortBy != GitRefsSortBy.Default, GetSortCriteria(sortBy, sortOrder), string.Empty },
```
gerhardol, line 2899: --sort was introduced in Git 2.14.6 (not 2.19, not opposing the raise)
For all other Git options, GE checks arguments to avoid crashes. Should that be done here?

GitCommands/Git/GitVersion.cs
```diff
@@ -22,7 +22,7 @@ public class GitVersion : IComparable<GitVersion>
         private static readonly GitVersion v2_19_0 = new GitVersion("2.19.0");
         private static readonly GitVersion v2_20_0 = new GitVersion("2.20.0");
 
-        public static readonly GitVersion LastSupportedVersion = v2_11_0;
+        public static readonly GitVersion LastSupportedVersion = v2_19_0;
         public static readonly GitVersion LastRecommendedVersion = new GitVersion("2.25.1");
```
gerhardol, line 26: Should be bumped to 2.27 - what we should use when "testing"

GitUI/BranchTreePanel/ContextMenu/GitRefsSortOrderContextMenuItem.cs
```diff
@@ -0,0 +1,77 @@
...
+using GitUI.Properties;
+using GitUIPluginInterfaces;
+using ResourceManager;
+
+namespace GitUI.BranchTreePanel.ContextMenu
+{
+    internal class GitRefsSortOrderContextMenuItem : ToolStripMenuItem
+    {
+        internal const string MenuItemName = "GitRefsSortOrderContextMenuItem";
+        private readonly TranslationString _sortByText = new TranslationString("&Sort order...");
```
gerhardol, line 15: translate?

gerhardol (commented):
Brief, review, brief testing. Some comments, considered blocking auto merge

gerhardol (changes requested):
git show-ref do not have --sort option
I cannot trigger the use of --sort to GetRefsCmd() with sort though

gerhardol:
@msftbot forget what I just told you

GitCommands/Git/GitModule.cs
```diff
@@ -2861,41 +2861,42 @@ public RemoteActionResult<IReadOnlyList<IGitRef>> GetRemoteServerRefs(string rem
...
-                    noLocks && GitVersion.Current.SupportNoOptionalLocks
-                        ? (ArgumentString)"--no-optional-locks"
-                        : default)
+                cmd = new GitArgumentBuilder("show-ref",
+                        gitOptions: noLocks && GitVersion.Current.SupportNoOptionalLocks
+                            ? (ArgumentString)"--no-optional-locks"
+                            : default)
                 {
                     { branches, "--dereference", "--tags" },
+                    { sortBy != GitRefsSortBy.Default, GetSortCriteria(sortBy, sortOrder), string.Empty }
```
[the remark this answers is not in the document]
RussKie: My mistake, I was too eager here, and haven't tested whether tags can be sorted.
Thank you

GitCommands/Git/GitModule.cs
RussKie: Yes, but `git branch.sort` option was only introduced in 2.19
https://github.blog/2018-09-10-highlights-from-git-2-19/#user-content-sorting-branches

> For all other Git options, GE checks arguments to avoid crashes. Should that be done here?

Please elaborate, I don't think I understand.

GitCommands/Git/GitVersion.cs
RussKie: Sure

GitUI/BranchTreePanel/ContextMenu/GitRefsSortOrderContextMenuItem.cs
RussKie: It appears these aren't translatable in the current form. 
I hoped no-one would notice :D

GitCommands/Git/GitModule.cs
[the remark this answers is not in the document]
RussKie: It looks like we don't need `git show-ref`, and instead we can replace it with `git for-each-ref` supplying necessary refs, i.e. `refs/heads/`, `refs/remotes/`, and `refs/tags/`. With this we can leverage the benefits of sorting.

Do you see any issues with this change?

GitCommands/Git/GitModule.cs
gerhardol: > Yes, but git branch.sort option was only introduced in 2.19

Not what is used here, so min version is 2.14.6

> Please elaborate, I don't think I understand.

GE has checks for Git options in versions, also for unsupported versions to avoid crashes.

GitCommands/Git/GitModule.cs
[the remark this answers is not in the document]
gerhardol: > Do you see any issues with this change?

No

RussKie:
Outstanding: resolution of the context menu placement for the Tags node. Judging by incoming PRs looks like @ivangrek has significantly better understanding of this part 😁.

GitCommands/Git/GitModule.cs
RussKie: I'm split. We're moving the bar up, so adding checks for something we no longer support feels redundant.
I'll add it if you feel it is truly necessary.

ivangrek:
Is this the correct state?

image

image

Plugins/GitUIPluginInterfaces/GitRefsOrder.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace GitUIPluginInterfaces
+{
+    // NB: The values are fed directly into git commands, casing is important!
+    public enum GitRefsSortBy
+    {
+        [Description("Git default")]
+        Default = 0,
+
+        [Description("Author date")]
+        authordate,
```
ivangrek, line 12: Enum Naming

RussKie: I left a comment on line:5 - these go directly into `git` commands

RussKie:
> Is this the correct state?

Champion! You're clearly a better tester than I am 🙇‍♂️ No, this is definitely something I didn't anticipate, and now I need to think of it more. 
Though this should go into https://github.com/gitextensions/gitextensions/pull/8414, this PR is about sorting.

RussKie:
@gerhardol I'd appreciate your re-review, I believe your points are addressed.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -81,39 +80,54 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
...
+            static string GetSortCriteria(bool needTags, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
+            {
+                string order = sortOrder == GitRefsSortOrder.Ascending ? string.Empty : "-";
+                if (!needTags)
+                {
+                    return $"--sort={order}{sortBy}";
+                }
+
+                // Sort by dereferenced data
+                return $"--sort={order}{sortBy} --sort={order}*{sortBy}";
```
gerhardol, line 129: Add a comment that Git 2.14.6 is required (maybe remove checks and add comments elsewhere?)
There is a workaround, to not set any sorting too
https://github.com/gitextensions/gitextensions/pull/8427#discussion_r483852667

gerhardol (commented):
I mostly looked at GitModule changes, a few comments only.
Briefly looked at the rest, briefly tested.
I suggest it is merged, but not deep enough review to formally approve.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -81,39 +80,54 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
...
 
+            GitArgumentBuilder cmd = new GitArgumentBuilder("for-each-ref",
+                gitOptions: noLocks && GitVersion.Current.SupportNoOptionalLocks
+                    ? (ArgumentString)"--no-optional-locks"
+                    : default)
+            {
+                { sortBy != GitRefsSortBy.Default, GetSortCriteria(tags, sortBy, sortOrder), string.Empty },
+                format,
+                { branches, "refs/heads/", string.Empty },
+                { branches && tags, "refs/remotes/", string.Empty },
```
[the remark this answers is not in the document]
RussKie: Thank you for pointing me to this. This API can do with quite a bit of improvement, but I didn't want to change to much in the context of this PR. I will do a follow up improvement later.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -81,39 +80,54 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
...
+                // or "dereferences" (i.e. commits that contian metadata and point to other commits, "^{}").
+                // Derefence commits do not contain date information, so we need to find information from the referenced commits (those with '*').
+                // So the following format is as follows:
+                //      If (there is a 'authordate' information, then this is a simple tag/direct commit)
+                //      Then
+                //          format = %(objectname) %(refname)
+                //      Else
+                //          format = %(*objectname) %(*refname) // i.e. info from a referenced commit
+                //      End
+                format = @"--format=""%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)""";
```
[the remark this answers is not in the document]
RussKie: Nice, we can certainly look at this as a follow up 👍

GitCommands/Git/GitModule.cs
```diff
@@ -2895,53 +2895,11 @@ public RemoteActionResult<IReadOnlyList<IGitRef>> GetRemoteServerRefs(string rem
...
         public IReadOnlyList<IGitRef> GetRefs(bool tags = true, bool branches = true)
         {
-            return GetRefs(tags, branches, false);
-        }
-
-        public IReadOnlyList<IGitRef> GetRefs(bool tags, bool branches, bool noLocks)
-        {
-            var refList = _gitExecutable.GetOutput(GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks: noLocks));
-
+            string cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks: true, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
```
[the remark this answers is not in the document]
RussKie: Yeah, I was somewhat confused and had to search through docs. My impression that we want to use "nolocks" for these commands.

GitCommands/Git/Commands/GitCommandHelpers.cs
RussKie: Added `GitVersion.Current.SupportRefSort`

RussKie:
> I suggest it is merged, but not deep enough review to formally approve.

Taking this as an approval. Feel free to cancel the auto-merge, if it isn't so.

RussKie:
@msftbot merge in 24 hours or if @gerhardol approves

RussKie:
@msftbot merge in 24 hours

GitCommands/Git/GitModule.cs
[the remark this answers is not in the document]
gerhardol: We do not want to lock the repo for background operations. For git-status the commit count is in the background but FormCommit is in the foreground. This is the primary use of noLocks (before this, FormCommit occasionally failed).

This change assumes that all GetRefs() is done in the background, which is the case right now but may not be correct in the future.

gerhardol (approved):
> > I suggest it is merged, but not deep enough review to formally approve.
> 
> Taking this as an approval. Feel free to cancel the auto-merge, if it isn't so.

Approval to merge, but incomplete review, focusing on GetRefs() that I found to be important. 
More of browsing the remaining, I could understand what I looked at and no obvious problems.
There are a few discussion points still, can be discussed separately.

Approving with the limitations in the review stated above.

GitCommands/Git/GitModule.cs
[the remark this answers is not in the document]
RussKie: I added your explanation as a comment to the code.

jrfnl:
Thank you!!!! ❤️  Been missing this a lot!

Only question I have is: is there a way to sort alphanumerically, but case insensitively ? Might be possible already, but I haven't figured out how to do it.

RussKie:
I don't know.
Git is a Linux tool, so it is generally case sensitive. And as you can see
from the PR diff or command log (F12) the sort command can be quite
intricate.

That said, I haven't considered or investigated case sensitivity because
for me it was never an issue...

RussKie:
Related: https://github.com/gitextensions/gitextensions/issues/8508

PR #8436, opened 2020-09-07
Prevent a crash when opening the settings

pmiossec:
## Proposed changes

Prevent a crash when opening the settings when they are not the same numbers of elements in the 2 collections.

From what I understand, it is not supposed to happen with the released version.
But it happened to me... surely when testing a PR.
So, I think it couldn't be bad to harden the code here....

## Test methodology 

- Manual

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
@@ -18,7 +18,8 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
 
         protected override void SettingsToPage()
         {
-            for (int i = 0; i < AppSettings.CascadeShellMenuItems.Length; i++)
+            var itemsCount = Math.Min(AppSettings.CascadeShellMenuItems.Length, chlMenuEntries.Items.Count);
+            for (int i = 0; i < itemsCount; i++)
             {
                 chlMenuEntries.SetItemChecked(i, AppSettings.CascadeShellMenuItems[i] == '1');
             }


RussKie:
What's the error and how to reproduce?

pmiossec:
> What's the error

The error was that the settings `AppSettings.CascadeShellMenuItems` contained one more element than entries in the control `chlMenuEntries.Items.Count`. I don't know how I end up with that.
I imagine by testing a PR that add a new entry but not sure because I don't remember one doing that...

> and how to reproduce?

No idea. Perhaps editing the settings file through an editor. But I don't think it worth the effort.

I decided to push this fix because that was the only way to solve the crash that happened every time I tried to open the settings.
And also because it could prevent another dev to loose some times on the same problem.

RussKie:
I'm not against a fix, but I'd like to understand the implications.
If these two arrays differ, then the resulting selection is likely to be incorrect. Should we then reset instead of proceeding on with the selection?

GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
```diff
@@ -18,7 +18,8 @@ public ShellExtensionSettingsPage()
 
         protected override void SettingsToPage()
         {
-            for (int i = 0; i < AppSettings.CascadeShellMenuItems.Length; i++)
+            var itemsCount = Math.Min(AppSettings.CascadeShellMenuItems.Length, chlMenuEntries.Items.Count);
```
gerhardol, line 21: The reason for the fix should be commented

gerhardol (approved):
Sometimes fixes just for dev is needed...

GintasS:

PR #8440, opened 2020-09-08
Change creating branch on HEAD by default

### LINKED ISSUE #8297: create branch from here
## Feature description

This is my first suggestion so bear with me!

This proposal is regarding the wording of the _Create branch_ feature in _Open Git Repository_ => _Commands_ menu.

When using this menu item to create a new branch the default commit to branch off of is the currently selected commit in the the git history visual browser. This is an easy mistake to make and leads to awkward merges into HEAD; ie: fixing the mistake.

### Proposal

Change the wording of the menu item from "Create branch..." to "Create branch at selection".

I think it will help users to properly branch if there is a call out that the branch will be created from the selected commit. There is already a similar call out when you right click a commit with the "Create new branch here".

## Environment

- Git Extensions 3.3.1.7897
- Git 2.25.1.windows.1
- .NET Framework 4.8.4180.0
- DPI 192dpi (200% scaling)

ivangrek:
Fixes #8297

## Proposed changes

- Change creating branch on HEAD by default

### Before

image

### After

image

commitInfo

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
@@ -42,6 +42,7 @@
             this.groupBox1 = new System.Windows.Forms.GroupBox();
             this.flowLayoutPanel1 = new System.Windows.Forms.FlowLayoutPanel();
             this.gotoUserManualControl1 = new GitUI.UserControls.GotoUserManualControl();
+            this.commitSummary = new GitUI.UserControls.CommitSummaryUserControl();
             this.tableLayoutPanel1.SuspendLayout();
             this.groupBox1.SuspendLayout();
             this.flowLayoutPanel1.SuspendLayout();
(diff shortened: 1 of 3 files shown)


timjen3:
This is a great idea. Nice thinking. 👌

Just to be clear, this will not necessarily change the default to HEAD as the title states. It will change the default to branch at the currently _checked out_ commit (`RevisionGrid.CurrentCheckout`). For example, if the user has checked out a commit `-3` then the branch would be created at `-3`, right?

ivangrek:
> For example, if the user has checked out a commit -3 then the branch would be created at -3, right?

Yes. After checkout it will be HEAD.

pmiossec:
I have to test to see the impact of this PR, but from what I understand, I don't think that I will like it.

What is the behavior if I right click on a specific commit and choose "Create new branch here..."

image

If that no more create the branch here because the HEAD is not there, we have a problem.

And also because that was the same keyboard shortcut and it was consistent with the global menu:
image

I have to test ...........

ivangrek:
If the global context, then create on HEAD.
If the context of the grid, then on the selected one.

pmiossec:
And if you do `ctrl+B`? because that's the same shortcut...
I hope it will be the same that context grid 🤞  because I use it quite often...

ivangrek:
If the grid in focus, it works like the context menu, but, for examle, left panel work in global context.

RussKie:
I am deadset opposed to this proposal, for me this would be a massive useability issue. I want to be able to create a branch at the selection. If I want to create a branch at the HEAD then I select the HEAD first.

ivangrek:
In this case, is it worth agreeing the names of the global menu items and the context menu?

This part does not look user friendly, at least for a beginner.
image

Or there is always a third option. 😄

RussKie:
It is a fair observation, I think we can iterate on this point.
I have few ideas, I'll share when get to a computer.

gerhardol:
The handling should be consistent. A number of actions must use HEAD (like reset) other use HEAD by default (like Format patch), some are aware of the selection (create branch/tag, archive).

I believe always use HEAD from command menu is the right way forward, use the grid (or change in the dialog).

RussKie:
Copying from https://github.com/gitextensions/gitextensions/issues/8297#issuecomment-690852634

> I think it is rare to branch not on HEAD. I am not the most experienced but for me 99% of the time a new branch is created off of HEAD.

I rarely create branches off the HEAD, so I can't agree with you here. In fact I mostly create branches off a tip of a remote branch, which is rarely my local HEAD.
And this is why creating branches off a HEAD for me would be a useability issue.



> Create branch
> -> from branch HEAD
> -> from selection

This just adds more cognitive overload, our context menus already busy to the point where it is easy to get lost. 
Personally I'm finding more and more often myself clicking incorrect menus because they look similar.



> When the option is invoked a user is presented with a dialog:
> image

How about showing the textbox with a commit hash, which is mostly useless on its own, we show a commit info, similar to what we do in the Reset dialog?
Something like this:
image

RussKie:
I have a working prototype in https://github.com/gitextensions/gitextensions/pull/8491
image

ivangrek:
> I have a working prototype in #8491

Can this be closed?

RussKie:
I suppose we can.

PR #8450, opened 2020-09-11
AppSettings Updated event

ivangrek:
## Proposed changes

- Add the ability to receive an event about a change in the setting value.
- Also, in the future, we can remove all classes inherited from Setting<T> and use one generic.

### After

image

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Settings/Setting.cs
+++ b/GitCommands/Settings/Setting.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
 
 namespace GitCommands.Settings
 {
@@ -8,6 +10,8 @@ namespace GitCommands.Settings
         public readonly T DefaultValue;
         public readonly string Name;
 
+        public event EventHandler Updated;
+
         protected Setting(string name, SettingsPath settingsSource, T defaultValue)
         {
             Name = name;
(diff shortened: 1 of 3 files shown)


GitCommands/Settings/Setting.cs
```diff
@@ -39,5 +56,39 @@ public virtual bool ValueIsEmpty(T value)
...
+                {
+                    case TypeCode.String:
+                        var stringValue = (string)(object)value;
+
+                        return string.IsNullOrEmpty(stringValue)
+                            ? null
+                            : stringValue;
+                }
+
+                return JsonConvert
```
RussKie, line 89: ditto, meld into the swtich

GitCommands/Settings/Setting.cs
```diff
@@ -8,14 +10,29 @@ public abstract class Setting<T>
         public readonly T DefaultValue;
         public readonly string Name;
 
+        public event Action Updated;
```
RussKie, line 13: It is customary to type events as `EventHandler` (or derivative)
```suggestion
        public event EventHandler SettingUpdated;
```

GitCommands/Settings/Setting.cs
```diff
@@ -8,14 +10,29 @@ public abstract class Setting<T>
...
+            get => GetVal(Name, DefaultValue);
+            set
+            {
+                if (Value.Equals(value))
+                {
+                    return;
+                }
+
+                SetVal(Name, value);
+                Updated?.Invoke();
```
RussKie, line 33: Can we use [`INotifyPropertyChanged `](https://docs.microsoft.com/dotnet/api/system.componentmodel.inotifypropertychanged?view=netframework-4.6.1) instead?

RussKie (changes requested):

GitCommands/Settings/Setting.cs
ivangrek: We can. But I think this is a little for another case and we have one property.

RussKie (approved):

PR #8451, opened 2020-09-12
Settings updated event [cleanup]

ivangrek:
Continuation #8450

## Proposed changes

- Add the ability to receive an event about a change in the setting value.
- Also, in the future, we can remove all classes inherited from Setting<T> and use one generic.

### After

image

We can replace

image

And use

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

----


ivangrek:
@RussKie could you please do a review.

RussKie:
So if we get more properties that we can wire up events too we should consider INotifyPropertyChanged?

GitCommands/Settings/Setting.cs
```diff
@@ -4,91 +4,131 @@
...
+
+        T DefaultValue { get; }
+
+        T ValueOrDefault { get; }
+
+        T Value { get; set; }
+
+        string FullPath { get; }
+
+        bool ValueIsEmpty(T value);
```
RussKie, line 23: How about this?
```suggestion
        bool IsUnset(T value);
```

GitCommands/Settings/Setting.cs
```diff
@@ -4,91 +4,131 @@
...
                 }
-            });
+            }
+
+            public string FullPath => SettingsSource.PathFor(Name);
+
+            public bool ValueIsEmpty(T value)
+            {
+                return EqualityComparer<T>.Default.Equals(value, default);
+            }
```
RussKie, line 99: This makes my head spin.

Are we checking that:
* the value is unset, or 
* the value is default, or
* the value is empty?

GitCommands/Settings/AppSettings.cs
```diff
@@ -404,11 +403,11 @@ public static bool ShowResetAllChanges
...
-        public static readonly StringSetting ConEmuStyle = new StringSetting("ConEmuStyle", DetailedSettingsPath, "<Solarized Light>");
-        public static readonly StringSetting ConEmuTerminal = new StringSetting("ConEmuTerminal", DetailedSettingsPath, "bash");
-        public static readonly StringSetting ConEmuFontSize = new StringSetting("ConEmuFontSize", DetailedSettingsPath, "12");
-        public static readonly BoolNullableSetting ShowGpgInformation = new BoolNullableSetting("ShowGpgInformation", DetailedSettingsPath, true);
+        public static readonly ISetting<bool> ShowConEmuTab = Setting.Create(DetailedSettingsPath, "ShowConEmuTab", true);
+        public static readonly ISetting<string> ConEmuStyle = Setting.Create(DetailedSettingsPath, "ConEmuStyle", "<Solarized Light>");
+        public static readonly ISetting<string> ConEmuTerminal = Setting.Create(DetailedSettingsPath, "ConEmuTerminal", "bash");
+        public static readonly ISetting<string> ConEmuFontSize = Setting.Create(DetailedSettingsPath, "ConEmuFontSize", "12");
+        public static readonly ISetting<bool> ShowGpgInformation = Setting.Create(DetailedSettingsPath, "ShowGpgInformation", true);
```
RussKie, line 410: Can we use `nameof()` like you did in other places?

GitCommands/Settings/Setting.cs
```diff
@@ -4,91 +4,131 @@
...
+        T Value { get; set; }
+
+        string FullPath { get; }
+
+        bool ValueIsEmpty(T value);
+    }
+
+    public static class Setting
+    {
+        public static ISetting<string> Create(SettingsPath settingsSource, string name, string defaultValue)
```
RussKie, line 28: Would `Bind` be more descriptive of a nature of the operation here? That is, we _create a binding to an existing setting_ vs we _create a new setting_.

RussKie (commented):
👍

ivangrek:
> So if we get more properties that we can wire up events too we should consider INotifyPropertyChanged?

We still have one property "Value"
The rest are constant or depend on "Value"

GitCommands/Settings/Setting.cs
ivangrek: In this case, we are creating. We bind to the UI elsewhere. Just a factory method.

GitCommands/Settings/Setting.cs
ivangrek: That the value is equal to the default value for the given type.
In the existing implementation, for data types, some value, for nullable is null.
For the given context, for value is empty or not set.

GitCommands/Settings/Setting.cs
ivangrek: Changed to "private IsDefault"
Checking for value types, in case got null

GitCommands/Settings/Setting.cs
ivangrek: Changed public to private

GitCommands/Settings/Setting.cs
```diff
@@ -4,91 +4,129 @@
...
-        }
+            public SettingsPath SettingsSource { get; }
 
-        public string FullPath => SettingsSource.PathFor(Name);
+            public string Name { get; }
 
-        private T GetValue(string name, T defaultValue = default)
-        {
-            return SettingsSource.GetValue(name, defaultValue, value =>
+            public T DefaultValue { get; }
```
RussKie, line 58: This name a bugs me a little.
There is: `Value` and `ValueOrDefault`, so it feels the 3rd options should be just called `Default`. Especially since we have `IsDefault` too.

RussKie (approved):

RussKie:
> > So if we get more properties that we can wire up events too we should consider INotifyPropertyChanged?
> 
> We still have one property "Value"
> The rest are constant or depend on "Value"

I'm thinking instead of

image

have something that works for every settings, e.g.:

```cs
AppSetitngs.SettingChanged += (s, e) => { ... }
```

ivangrek:
We have different settings, this may require more changes.
We also need to know that this particular setting has been updated.
Better if it's in one place. At least for now.

ivangrek:
image

RussKie:
Ugh... we'll have to deal with it one day :)

RussKie:
I had a go at refactoring it 3 years ago (ugh!) https://github.com/RussKie/gitextensions/tree/refactor_AppSettings

ivangrek:
Are you planning to apply these changes?

RussKie:
> > I had a go at refactoring it 3 years ago (ugh!) [https://github.com/RussKie/gitextensions/tree/refactor_AppSettings](https://github.com/RussKie/gitextensions/tree/refactor_AppSettings?rgh-link-date=2020-09-13T11%3A37%3A15Z)
> 
> Are you planning to apply these changes?

Maybe... One day.... When I have completed other things :) There are so many things I'd like to do, and there's so much time I have for this project.
Feel free to use it.

GitCommands/Settings/Setting.cs
ivangrek: Controversial purpose of this property.

For non nullable Value == DefaultValue
For nullable Value == DefaultValue, cause null is value too.

RussKie: Agree, that's why typically nullable and non-nullable API have similar yet different implementations.
In this case I wouldn't spend too much time on it though - apply the comment, rename and move on.

ivangrek:
Renamed DefaultValue to Default

PR #8458, opened 2020-09-15
Don't use "AddSettingBinding" in "FormBrowseRepoSettingsPage"

### LINKED ISSUE #8446: Settings: Checkbox not well reflecting the setting state
## Current behaviour

When the only settings source available is "Global for all repositories" and so that there is no "Effective" source, the value could be not well set to the checkbox

See checkbox in is in intermediate state even if the setting value is 'true'/checked:
image

## Expected behaviour

The checkbox is checked.

## Steps to reproduce

I don't know at the moment :(

Not sure but I don't think so.

## Environment

- Git Extensions 33.33.33
- Build 09a8426c36a826a76f6a3abd65b6914ae94d047a (Dirty)
- Git 2.28.0.windows.1
- .NET Framework 4.8.4210.0
- DPI 168dpi (175% scaling)

## Diagnostic

Hint: that's perhaps here: https://github.com/gitextensions/gitextensions/blob/master/Plugins/GitUIPluginInterfaces/Settings/BoolSetting.cs#L57

ivangrek:
Fixes #8446

## Proposed changes

- The use "AddSettingBinding" in "FormBrowseRepoSettingsPage" doesn't make sense because the page implements only IGlobalSettingsPage

### Before

image

### After

image

## Test methodology 

- Manual

### How to test

image

#### Before

image

#### After 

image

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
@@ -20,20 +20,22 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         protected override void Init(ISettingsPageHost pageHost)
         {
             base.Init(pageHost);
-
-            // Bind settings with controls
-            AddSettingBinding(AppSettings.ShowConEmuTab, chkChowConsoleTab);
-            AddSettingBinding(AppSettings.ShowGpgInformation, chkShowGpgInformation);
         }
 
         protected override void PageToSettings()
         {
+            AppSettings.ShowConEmuTab.Value = chkChowConsoleTab.Checked;
+            AppSettings.ShowGpgInformation.Value = chkShowGpgInformation.Checked;
+
             AppSettings.ConEmuTerminal.Value = ((IShellDescriptor)cboTerminal.SelectedItem).Name.ToLowerInvariant();
             base.PageToSettings();
         }
 
         protected override void SettingsToPage()
         {
+            chkChowConsoleTab.Checked = AppSettings.ShowConEmuTab.Value;
+            chkShowGpgInformation.Checked = AppSettings.ShowGpgInformation.Value;
+
             foreach (IShellDescriptor shell in _shellProvider.GetShells())
             {
                 cboTerminal.Items.Add(shell);


pmiossec:
It fixes the loading problem with the checkbox.
Was it the only page doing that? Because there are other pages with only "Global for all repositories"...

Don't you think there is a way to fix `AddSettingBinding()` to handle pages with only "Global for all repositories" to not be obliged to do something like that on all the pages like that?

ivangrek:
>Was it the only page doing that?

Only used here.

> Don't you think there is a way to fix AddSettingBinding() to handle pages with only "Global for all repositories" to not be obliged to do something like that on all the pages like that?

Something tells me that this is not implemented.

Perhaps you can clarify the purpose.

> The purpose of the "Effective" setting is not clear. I would be glad if someone could clarify the purpose.

ivangrek:
It may be necessary to download "Effective" instead of "Global". Looking deeper, there is a lot of code to touch.

Better to put it out as a separate task with clear goals.

We want to refactor

```
void PageToSettings()
void SettingsToPage()
```

to

```
void Init(ISettingsPageHost pageHost)
{
   AddSettingBinding(AppSettings.ShowConEmuTab, chkChowConsoleTab);
}
```

for all pages?

Or just for simple settings.

RussKie:
> Perhaps you can clarify the purpose.
> 
> > The purpose of the "Effective" setting is not clear. I would be glad if someone could clarify the purpose.

Some settings can be distributed:
- global settings - %UserAppData%\GitExtensions.settings
- team settings - settings that are distributed with the repo stored in .\GitExtensions.settings
- user settings - user-only repo-specific settings that reside in .\.git folder

Each layer overrides the above, so user > team > global. "Effective" is the lowest-level settings available.

Many years ago there was a discussion that you may want to read: https://github.com/gitextensions/gitextensions/issues/1707

RussKie:
MC

ivangrek:
Alternatively, we can add distributed flag.

```
public BoolSetting(string name, bool defaultValue, bool distributed)
            : this(name, name, defaultValue)
{
}
```
Or

```
public ISettingControlBinding CreateControlBinding(bool distributed)
{
    return new CheckBoxBinding(this, CustomControl, distributed);
}
```

```
public override void LoadSetting(ISettingsSource settings, CheckBox control)
{
    var applyDefaultValue = (_distributed && settings.SettingLevel == SettingLevel.Effective)
        || !_distributed;

    bool? settingVal = applyDefaultValue
        ? Setting.ValueOrDefault(settings)
        : Setting[settings];

    control.SetNullableChecked(settingVal);
}
```

But this will require changing the interfaces, which will affect the work of plugins.

If we go down this path, I would first separate Setting and Binding.
What do you think?

RussKie:
> But this will require changing the interfaces, which will affect the work of plugins.

I would avoid this for the moment.

RussKie (approved):

RussKie:
@pmiossec any objections?

pmiossec (approved):

pmiossec:
>@pmiossec any objections?

no. LGTM

PR #8433, opened 2020-09-04
Add checkbox to prune tags

ivangrek:
Fixes #8430 

### Before

image

### After

image

image

## Test methodology 

Manual

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Build d4b0f48bbf3e39a71f5d7ff5231be5678d4aa0f1
- Git 2.28.0.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4200.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8463, opened 2020-09-17
Revert "Add checkbox to prune tags"

RussKie:
Reverts gitextensions/gitextensions#8433


RussKie:
@ivangrek do you have any immediate plans to address the regression? If not, I'm going to revert the change.

ivangrek:
Removed for toolbar.

image

PR #8471, opened 2020-09-21
support for range-diff

### LINKED ISSUE #8448: Introduce UI for range-diff command
## Feature description
Git 2.19 added a new feature [range-diff](https://git-scm.com/docs/git-range-diff) which is very useful for checking for example conflict resolutions in a rebased branch. It would be awesome to have some UI for it. It could be launched in a similar manner like the regular diff window, i.e. selecting first commit using _Compare - Select as BASE to compare (Ctrl + L)_ and then selecting the other commit.
The UI itself could show the output of the command `git range-diff {first}...{second}` as a list of commits which are either unique in one of the revisions or same in both or differing. There should be options to hide each "category" of commits. For differing commits, there should be an option to show the original commit from either branch or their "combined" diff as reported by the `range-diff` command.

## Environment

- Git Extensions 33.33.33
- Build 4992b04e6e70ed491cf333365668be1fb4d51125
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

gerhardol:
Part of #8448 support for range diff

First 4 commits are refactoring, the two following are likely to be squashed (the second changes data structures, separated for review)

#8516 is a separate PR for presentation

## Proposed changes

Present range-diff in diff tab

Limit RevDiff context menu entries for combined diff and Git Errors, to limit for range-diff too (will be kept as a separate commit)

### Before

range-diff is new

combined diff
currentCombined diff

### After

See #8516 for range-diff

Combined diff

image

## Test methodology 

manual

----

### DIFF
--- a/GitCommands/Git/GitItemStatus.cs
+++ b/GitCommands/Git/GitItemStatus.cs
@@ -38,7 +38,10 @@ namespace GitCommands
             IsSkipWorktree = 1 << 10,
             IsSubmodule = 1 << 11,
             IsDirty = 1 << 12,
-            IsStatusOnly = 1 << 13
+
+            // Other flags are parsed from Git status, set fake flags for special uses
+            IsStatusOnly = 1 << 13,
+            IsRangeDiff = 1 << 14
         }
 
         private JoinableTask<GitSubmoduleStatus> _submoduleStatus;
@@ -136,14 +139,24 @@ namespace GitCommands
             set => SetFlag(value, Flags.IsDirty);
         }
 
-        /// <summary>
+        /// <remarks>
         /// This item is not a Git item, just status information
         /// If ErrorMessage is set, this is an error from Git, otherwise just a marker that nothing is changed
-        /// </summary>
+        /// </remarks>
         public bool IsStatusOnly
         {
             get => _flags.HasFlag(Flags.IsStatusOnly);
             set => SetFlag(value, Flags.IsStatusOnly);
+        }
+
+        /// <remarks>
+        /// This item is not a native git item, but a status information
+        /// calculated with git range-diff command.
+        /// </remarks>
+        public bool IsRangeDiff
+        {
+            get => _flags.HasFlag(Flags.IsRangeDiff);
+            set => SetFlag(value, Flags.IsRangeDiff);
         }
 
         private void SetFlag(bool isSet, Flags flag)
--- a/GitUI/UserControls/FileStatusItem.cs
+++ b/GitUI/UserControls/FileStatusItem.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.CompilerServices;
 using GitCommands;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
@@ -7,11 +8,13 @@ namespace GitUI.UserControls
 {
     public sealed class FileStatusItem
     {
-        public FileStatusItem([CanBeNull] GitRevision firstRev, [NotNull] GitRevision secondRev, [NotNull] GitItemStatus item)
+        public FileStatusItem([CanBeNull] GitRevision firstRev, [NotNull] GitRevision secondRev, [NotNull] GitItemStatus item, [CanBeNull] ObjectId baseA = null, [CanBeNull] ObjectId baseB = null)
         {
             FirstRevision = firstRev;
             SecondRevision = secondRev ?? throw new ArgumentNullException(nameof(secondRev));
             Item = item ?? throw new ArgumentNullException(nameof(item));
+            BaseA = baseA;
+            BaseB = baseB;
         }
 
         /// <summary>
@@ -27,6 +30,18 @@ namespace GitUI.UserControls
         /// </summary>
         [NotNull]
         public GitRevision SecondRevision { get; }
+
+        /// <summary>
+        /// If ranges are selected, the first commit (base) for <see cref="FirstRevision"/> (head)
+        /// </summary>
+        [CanBeNull]
+        public ObjectId BaseA { get; }
+
+        /// <summary>
+        /// If ranges are selected, the first commit (base) for <see cref="SecondRevision"/> (head)
+        /// </summary>
+        [CanBeNull]
+        public ObjectId BaseB { get; }
 
         /// <summary>
         /// The status item in the list
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -937,7 +937,7 @@ namespace GitUI
                         listItem.Selected = true;
                     }
 
-                    listItem.Tag = new FileStatusItem(i.FirstRev, i.SecondRev, item);
+                    listItem.Tag = new FileStatusItem(i.FirstRev, i.SecondRev, item, i.BaseA, i.BaseB);
                     list.Add(listItem);
                 }
             }
@@ -982,15 +982,20 @@ namespace GitUI
 
             static string GetItemImageKey(GitItemStatus gitItemStatus)
             {
-                if (!gitItemStatus.IsNew && !gitItemStatus.IsDeleted && !gitItemStatus.IsTracked)
+                if (!gitItemStatus.IsNew && !gitItemStatus.IsDeleted && !gitItemStatus.IsTracked && !gitItemStatus.IsRangeDiff)
                 {
-                    // Illegal combinations, no flags set?
+                    // Illegal flag combinations or no flags set?
                     return nameof(Images.FileStatusUnknown);
                 }
 
                 if (gitItemStatus.IsDeleted)
                 {
                     return nameof(Images.FileStatusRemoved);
+                }
+
+                if (gitItemStatus.IsRangeDiff)
+                {
+                    return nameof(Images.Diff);
                 }
 
                 if (gitItemStatus.IsNew || (!gitItemStatus.IsTracked && !gitItemStatus.IsSubmodule))
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -20,31 +20,23 @@ using GitUI.Properties;
 using GitUI.UserControls;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
-using ResourceManager;
 
 namespace GitUI
 {
-    public delegate string DescribeRevisionDelegate(ObjectId objectId);
-
     public sealed partial class FileStatusList : GitModuleControl
     {
         private static readonly TimeSpan SelectedIndexChangeThrottleDuration = TimeSpan.FromMilliseconds(50);
-        private readonly TranslationString _diffWithParent = new TranslationString("Diff with a/");
-        private readonly TranslationString _diffBaseToB = new TranslationString("Unique diff BASE with b/");
-        private readonly TranslationString _diffCommonBase = new TranslationString("Common diff with BASE a/");
-        private readonly TranslationString _combinedDiff = new TranslationString("Combined diff");
         private readonly IFullPathResolver _fullPathResolver;
+        private readonly FileStatusDiffCalculator _diffCalculator;
         private readonly SortDiffListContextMenuItem _sortByContextMenu;
         private readonly IReadOnlyList<GitItemStatus> _noItemStatuses;
 
         private int _nextIndexToSelect = -1;
-        private bool _groupByRevision;
         private bool _enableSelectedIndexChangeEvent = true;
         private bool _mouseEntered;
         private ToolStripItem _openSubmoduleMenuItem;
         private Rectangle _dragBoxFromMouseDown;
-        private IReadOnlyList<(GitRevision firstRev, GitRevision secondRev, string summary, IReadOnlyList<GitItemStatus> statuses)> _itemsWithDescription
-            = Array.Empty<(GitRevision, GitRevision, string, IReadOnlyList<GitItemStatus>)>();
+        private IReadOnlyList<FileStatusWithDescription> _itemsWithDescription = new List<FileStatusWithDescription>();
         [CanBeNull] private IDisposable _selectedIndexChangeSubscription;
         [CanBeNull] private IDisposable _diffListSortSubscription;
 
@@ -100,6 +92,7 @@ namespace GitUI
             FilterWatermarkLabel.Font = new Font(FilterWatermarkLabel.Font, FontStyle.Italic);
             FilterComboBox.Font = new Font(FilterComboBox.Font, FontStyle.Bold);
 
+            _diffCalculator = new FileStatusDiffCalculator(() => Module);
             _fullPathResolver = new FullPathResolver(() => Module.WorkingDir);
             _noItemStatuses = new[]
             {
@@ -145,7 +138,8 @@ namespace GitUI
                     (nameof(Images.SubmoduleRevisionSemiUp), ScaleHeight(Images.SubmoduleRevisionSemiUp)),
                     (nameof(Images.SubmoduleRevisionSemiUpDirty), ScaleHeight(Images.SubmoduleRevisionSemiUpDirty)),
                     (nameof(Images.SubmoduleRevisionSemiDown), ScaleHeight(Images.SubmoduleRevisionSemiDown)),
-                    (nameof(Images.SubmoduleRevisionSemiDownDirty), ScaleHeight(Images.SubmoduleRevisionSemiDownDirty))
+                    (nameof(Images.SubmoduleRevisionSemiDownDirty), ScaleHeight(Images.SubmoduleRevisionSemiDownDirty)),
+                    (nameof(Images.Diff), ScaleHeight(Images.Diff))
                 };
 
                 for (var i = 0; i < images.Length; i++)
@@ -243,11 +237,9 @@ namespace GitUI
 
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         [Browsable(false)]
-        public DescribeRevisionDelegate DescribeRevision { get; set; }
+        public Func<ObjectId, string> DescribeRevision { get; set; }
 
         public bool FilterFocused => FilterComboBox.Focused;
-
-        public bool ShouldSerializeFilterVisible => FilterVisible != true;
 
         public bool FilterVisible
         {
@@ -295,14 +287,14 @@ namespace GitUI
         {
             get
             {
-                return GitItemStatusesWithDescription?.SelectMany(tuple => tuple.statuses).AsReadOnlyList()
+                return GitItemStatusesWithDescription?.SelectMany(tuple => tuple.Statuses).AsReadOnlyList()
                        ?? Array.Empty<GitItemStatus>();
             }
         }
 
-        private IReadOnlyList<(GitRevision firstRev, GitRevision secondRev, string summary, IReadOnlyList<GitItemStatus> statuses)> GitItemStatusesWithDescription
+        private IReadOnlyList<FileStatusWithDescription> GitItemStatusesWithDescription
         {
-            get { return _itemsWithDescription; }
+            get => _itemsWithDescription;
             set
             {
                 _itemsWithDescription = value ?? throw new ArgumentNullException(nameof(value));
@@ -310,18 +302,7 @@ namespace GitUI
             }
         }
 
-        public bool GroupByRevision
-        {
-            get => _groupByRevision;
-            set => _groupByRevision = value;
-        }
-
-        [DefaultValue(true)]
-        public bool MultiSelect
-        {
-            get => FileStatusListView.MultiSelect;
-            set => FileStatusListView.MultiSelect = value;
-        }
+        public bool GroupByRevision { get; set; }
 
         [Browsable(false)]
         [DefaultValue(true)]
@@ -440,7 +421,7 @@ namespace GitUI
         [DefaultValue(true)]
         public bool SelectFirstItemOnSetItems { get; set; }
 
-        public int UnfilteredItemsCount => GitItemStatusesWithDescription?.Sum(tuple => tuple.statuses.Count) ?? 0;
+        public int UnfilteredItemsCount => GitItemStatusesWithDescription?.Sum(tuple => tuple.Statuses.Count) ?? 0;
 
         // Public methods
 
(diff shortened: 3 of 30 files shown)


crypto-rsa:
Unfortunately the code is not yet robust enough - I am getting `ArgumentOutOfRangeException` for parameter `count` when trying to open a diff. The stack trace is
```
   at System.String.Remove(Int32 startIndex, Int32 count)
   at GitUI.FileStatusList.SetDiffs(IReadOnlyList`1 revisions, Func`2 getRevision)
   at GitUI.CommandsDialogs.FormDiff.PopulateDiffFiles()
   at GitUI.CommandsDialogs.FormDiff.<.ctor>b__15_3(Object <p0>, EventArgs <p1>)
   at System.EventHandler.Invoke(Object sender, EventArgs e)
   at System.Windows.Forms.Form.OnLoad(EventArgs e)
   at GitUI.GitExtensionsForm.OnLoad(EventArgs e)
   at System.Windows.Forms.Form.WndProc(Message& m)
```

(sorry, I can't share the repo)

gerhardol:
> Unfortunately the code is not yet robust enough - I am getting `ArgumentOutOfRangeException` for parameter `count` when trying to open a diff. The stack trace is

For sure, the parsing is not stable enough, it is not even reviewable...
The other commits are more or less ready for review (but may be changed due to feedback about parsing).

If similar issues occurs after next update of the parsing, an obfuscated output of the raw range-diff is helpful.

gerhardol:
Adapted parsing to Git internals (quite some hardcoding there), should work better now
(icons are not setup now)

crypto-rsa:
Would it be possible to colour the lines starting with `@@` (delimiting individual file chunks), just like in a normal diff?

gerhardol:
> Would it be possible to colour the lines starting with `@@` (delimiting individual file chunks), just like in a normal diff?

For the raw output, the git colors could be adopted to the theme and displayed. But I did not see a simple way to display the terminal colors in the file viewer. 

My primary target is still to parse the output.

gerhardol:
Some cleanup (a range check for four selections simplified the code).

GitUI/UserControls/FileStatusList.cs
```diff
@@ -697,64 +703,146 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions, Func<ObjectId, GitRev
...
+                firstRev = revBase,
+                secondRev = selectedRev,
+                summary = _diffCommonBase.Text + GetDescriptionForRevision(baseRevGuid),
+                statuses = commonBaseToAandB
+            });
+
+            // Git range-diff is slow and memory consuming, so just skip if diff is large
+            // to avoid that GE seem to hang when selecting the range diff
+            const int maxRangeDiffCommits = 100;
+            int? count = Module.GetCommitCount(firstRevHead.ToString(), selectedRevHead.ToString());
```
mstv, line 827: This seems to work. I am confused how it does - at least without digging into git commands:
For `baseA..headA baseB..headB`, this will try to find the number of commits between the _unrelated_ `headA` and `headB`.
Would one or both distances between `baseA` and `headA` as well as between `baseB` and `headB` be a more comprehensible criterion?
(I have not yet thought about which one must be chosen for a selection of 3 commits.)

GitUI/UserControls/FileStatusList.cs
```diff
@@ -780,32 +868,36 @@ static ObjectId GetRevisionOrHead(GitRevision rev, Lazy<ObjectId> head)
...
 
         private string GetDescriptionForRevision(ObjectId objectId)
         {
-            if (DescribeRevision != null)
-            {
-                return DescribeRevision(objectId);
-            }
-
-            return objectId?.ToShortString();
+            return DescribeRevision != null ? DescribeRevision(objectId) : objectId?.ToShortString();
```
mstv, line 900: Then also?
```suggestion
            => DescribeRevision != null ? DescribeRevision(objectId) : objectId?.ToShortString();
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1103,6 +1195,11 @@ static string GetItemImageKey(GitItemStatus gitItemStatus)
                     return nameof(Images.FileStatusRemoved);
                 }
 
+                if (gitItemStatus.RangeDiff)
+                {
+                    return nameof(Images.Diff);
+                }
+
```
mstv, line 1202: Could be moved to the begin of the function. Would save the check at line 1187.

GitCommands/Git/GitItemStatus.cs
```diff
@@ -53,6 +53,7 @@ private enum Flags
         public string RenameCopyPercentage { get; set; }
 
         public StagedStatus Staged { get; set; }
+        public bool RangeDiff { get; set; }
```
mstv, line 56: Not sure, should this be another flag as those below, especially like `IsStatusOnly`?

mstv (commented):
The split commits have facilitated the review.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -300,7 +298,17 @@ public IReadOnlyList<GitItemStatus> GitItemStatuses
...
-        private IReadOnlyList<(GitRevision firstRev, GitRevision secondRev, string summary, IReadOnlyList<GitItemStatus> statuses)> GitItemStatusesWithDescription
+        private class FileStatusWithDescription
+        {
+            public GitRevision firstRev;
+            public GitRevision secondRev;
+            public ObjectId baseA;
+            public ObjectId baseB;
+            public string summary;
+            public IReadOnlyList<GitItemStatus> statuses;
+        }
```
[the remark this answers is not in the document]
gerhardol: This is a private class, this should not be exposed to other modules.
I rather have a class than tuples

I spent some time restructuring to make this reviewable, will wait to change names.
(Even if this change itself simplifies comparing...)

GitUI/UserControls/FileStatusList.cs
```diff
@@ -697,64 +703,146 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions, Func<ObjectId, GitRev
...
+                ? new Lazy<ObjectId>(() => getRevision(ObjectId.IndexId).FirstParentId)
+                : new Lazy<ObjectId>(() => Module.RevParse("HEAD"));
+            var firstRevHead = GetRevisionOrHead(firstRev, head);
+            var selectedRevHead = GetRevisionOrHead(selectedRev, head);
+            var baseRevGuid = Module.GetMergeBase(firstRevHead, selectedRevHead);
 
+            // Four selected, to check if two ranges are selected
+            var baseA = (revisions.Count != 4 || baseRevGuid is null)
+                ? null
+                : Module.GetMergeBase(GetRevisionOrHead(revisions[3], head), firstRevHead);
```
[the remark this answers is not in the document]
gerhardol: Nothing compared to calculating the diff between unrelated commits.
(this is not changed from current)

git-merge-base can be eliminated by using getRevision() in the normal case when the common parent is within something like 10 commits from the commits to compare. But is optimizing the simple case.

Longer term, the diffs could be calculated async, maybe only calculate other than first diff on demand (when expanding the group).

This PR adds a git-rev-list call, which adds little to the load.

GitUI/UserControls/FileStatusList.cs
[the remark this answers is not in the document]
mstv: >> Nit: One class per file?

> This is a private class, this should not be exposed to other modules.

I was not sure whether this rule applies here. @RussKie?

> I rather have a class than tuples

👍

GitUI/UserControls/FileStatusList.cs
gerhardol: baseA/baseB is only used with four selections, there is a check that baseA really is the base to firstSelected, otherwise multi diff is done. So this is already checked
For three commits, the merge-base must be selected (so a 'base' that is older is not identified and multidiff is displayed instead).

GitUI/UserControls/FileStatusList.cs
gerhardol: ok

GitUI/UserControls/FileStatusList.cs
gerhardol: Yes, this is a leftover from supporting other info in RangeDiff

GitCommands/Git/GitItemStatus.cs
gerhardol: Yes it could be a flag. I choose not to as the other flags are git properties. IsStatusOnly is a special flag to show that this is not derived from Git (this could be separate too).

But I do not think there are many GitItemStatus at the same time, so a few bytes per file should not matter.
There are limited number of files in diff tab or similar, the RevTree only uses GitItemStatus when viewing files.

I do not mind changing this, we can add a comment separating properties from Git with the extra data

GitUI/UserControls/FileStatusList.cs
mstv: Yes, `headA` and `headB` have a common ancestor here. But still, what does "commit count" between these two mean? They are unrelated -- none is the base of the other -- at least in the interesting use case.

GitCommands/Git/GitItemStatus.cs
mstv: I agree, this won't turn the scales.
(Though for merges of vendor branches, almost all files are in the diff. Then diff loading is really slow already without range-diff.)
Not requesting a change.

GitUI/UserControls/FileStatusList.cs
gerhardol: The commit count are all commits traversing from first/A to selected/B. Compare to a local branch ahead 2 and behind 3 the tracked branch would have a commit count of 5.
It seems like git-range-diff is o(2) or so; while 100 seem fine for me, 1000 or so was slow.

GitUI/UserControls/FileStatusList.cs
gerhardol: Actually do not want to change this, the first check is for illegal combinations
If the "validility check" is considered unnecessary, FileStatusUnknown should be using the fallback at the very end.
(No other code to be changed; FileStatusAdded is a bit special but still OK.)

GitCommands/Git/GitItemStatus.cs
gerhardol: Changing this, adding comments anyway
But this should probably be an item on the FileStatusItem instead. The problem with that is though that GitStatusItems are often isolated. The only special for RangeDiff would be for GetItemImageIndex() and that can be handled anyway, but there could be other usages.
(and that would require a separate property anyway, so no byte saved)

GitUI/UserControls/FileStatusList.cs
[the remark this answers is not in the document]
RussKie: > This is a private class, this should not be exposed to other modules.

A private class can happily live in its own file ;)
E.g. https://github.com/dotnet/winforms/blob/master/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.SystemIAccessibleWrapper.cs#L13-L21

GitCommands/Git/GitItemStatus.cs
```diff
@@ -146,6 +149,16 @@ public bool IsStatusOnly
             set => SetFlag(value, Flags.IsStatusOnly);
         }
 
+        /// <summary>
+        /// This item is not a native Git item, just status information
+        /// The item is to be calculated with Git range-diff
+        /// </summary>
```
RussKie, line 155:
```suggestion
        /// <remarks>
        /// This item is not a native git item, but a status information
        /// calculated with git range-diff command.
        /// </remarks>
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -295,12 +293,22 @@ public IReadOnlyList<GitItemStatus> GitItemStatuses
...
             get
             {
-                return GitItemStatusesWithDescription?.SelectMany(tuple => tuple.statuses).AsReadOnlyList()
+                return GitItemStatusesWithDescription?.SelectMany(tuple => tuple.Statuses).AsReadOnlyList()
                        ?? Array.Empty<GitItemStatus>();
             }
         }
 
-        private IReadOnlyList<(GitRevision firstRev, GitRevision secondRev, string summary, IReadOnlyList<GitItemStatus> statuses)> GitItemStatusesWithDescription
+        private class FileStatusWithDescription
```
RussKie, line 301: Please move to own file. 
Randomly sprinkling the code with additional types increases maintenance overhead.

GitUI/UserControls/FileStatusItem.cs
```diff
@@ -28,6 +31,18 @@ public FileStatusItem([CanBeNull] GitRevision firstRev, [NotNull] GitRevision se
         [NotNull]
         public GitRevision SecondRevision { get; }
 
+        /// <summary>
+        /// If ranges are selected, the first commit (base) for FirstRevision (head)
```
RussKie, line 35:
```suggestion
        /// If ranges are selected, the first commit (base) for <see cref="FirstRevision"/> (head).
```

GitUI/UserControls/FileStatusItem.cs
```diff
@@ -28,6 +31,18 @@ public FileStatusItem([CanBeNull] GitRevision firstRev, [NotNull] GitRevision se
...
+        /// </summary>
+        [CanBeNull]
+        public ObjectId BaseA { get; }
+
+        /// <summary>
+        /// If ranges are selected, the first commit (base) for SecondRevision (head)
```
RussKie, line 41:
```suggestion
        /// If ranges are selected, the first commit (base) for <see cref="SecondRevision"/> (head).
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -697,64 +703,146 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions, Func<ObjectId, GitRev
                     if (conflicts.Count != 0)
                     {
                         // Create an artificial commit
-                        var desc = _combinedDiff.Text;
-                        tuples.Add((new GitRevision(ObjectId.CombinedDiffId), selectedRev, desc, conflicts));
+                        tuples.Add(new FileStatusWithDescription
+                        {
+                            FirstRev = new GitRevision(ObjectId.CombinedDiffId), SecondRev = selectedRev, Summary = _combinedDiff.Text, Statuses = conflicts
```
RussKie, line 708:
```suggestion
                            FirstRev = new GitRevision(ObjectId.CombinedDiffId),
                            SecondRev = selectedRev,
                            Summary = _combinedDiff.Text,
                            Statuses = conflicts
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -660,7 +657,7 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions, Func<ObjectId, GitRev
             _revisions = revisions;
             _getRevision = getRevision;
 
-            var tuples = new List<(GitRevision firstRev, GitRevision secondRev, string summary, IReadOnlyList<GitItemStatus> statuses)>();
+            var tuples = new List<FileStatusWithDescription>();
```
RussKie, line 660: Need a better name for `tuples`.

RussKie (commented):
I don't have a good grasp on what's going on here, so reviewing purely from API perspective. 

`SetDiffs` method scares me due to many code paths, and overall complexity. I had a go at breaking it out for the purpose of seeing how tightly it is coupled to `FileStatisList`, and it turns out there is no strong connection between them.

I broke it out to own class (need better names, etc) in https://github.com/gitextensions/gitextensions/commit/9212d438f7e0eeff63386f39f10960bf5f29460c (I don't have access to your fork).
I feel there is more can be done to make it easier to understand, but even this is a good step in making both `SetDiff` more understandable, as well as improving `FileStatusList` overall.

GitUI/UserControls/FileStatusList.cs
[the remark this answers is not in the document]
gerhardol: That was an annoying way of doing it, just separating to follow a rule...

GitUI/UserControls/FileStatusList.cs
gerhardol: Applying this in SetStashDiffs() too, handling is uniform

GitUI/UserControls/FileStatusList.cs
gerhardol: fileStatusDescs

GitCommands/Git/GitItemStatus.cs
gerhardol: Changed to <remarks> for IsStatusOnly too

gerhardol:
> I broke it out to own class (need better names, etc) in 9212d43 (I don't have access to your fork).

You should have access to my fork (allow edit dform maintainers clicked).
I took your commit on top of other changes and adapted that though
I can live with that, some consequences:

* FileStatusWithDescription, FileStatusDiffCalculator some scripts have to be public
* Module cannot be IGitModule as GetTreeFiles() and a few more are only in GitModule
* History for changes will get lost (I had not planned to squash all commits, but this is getting too complex...)

gerhardol:
CodeFactor reports the refactored FileStatusDiffCalculator.cs:SetDiffs() as a new issue (code moved from FileStatusList
(It has been changed to simplify handling in this PR, but complexity is still relative high).

GitUI/UserControls/FileStatusList.cs
mstv: The git command used is "git rev-list headA ^headB --count". It seems to not take much time. And it yields smaller counts than you expect.

Since the range-diff is not run anymore, unless it is selected: Do we still need to suppress it if there are too many commits inbetween?

If yes, I think my initial remarks should be taken into account:

> For `baseA..headA baseB..headB`, this will try to find the number of commits between the _unrelated_ `headA` and `headB`.
Would one or both distances between `baseA` and `headA` as well as between `baseB` and `headB` be a more comprehensible criterion?

gerhardol: The command to find differences should be 
```git rev-list  --count headA...headB```
will update

There is still possible to have larger differences for commands with diverging forks when range-diff can take some time

At work with the Symantec virus running, range-diff requires 20-40s for a couple of range-diff with a count of about 110 (time is quicker if many commits are the same)
Not responsive, but you select the range-diff explicitly and it does not 'hang' forever
These are scenarios where I am interested to see the result (with this many, only the overview i)

For two forks with a count of 500 I waited 20 min before aborting
The range-diff algorithm has ```cubic runtime complexity```

I find 100 reasonable
I could be interested with more, if  implement some async calculations that can be aborted

This leads into another presentation change: I would like to make it possible to only see the summary.
(Parsing the output yield that, but adds a delay.) Will probably add to the presentation part of the PR for that.

Comment (gerhardol) [reply GitUI/UserControls/FileStatusList.cs]: One way to set the limit quite low is to include the Git command as text, then it is relative easy for the user to get the output without risking "hanging".

Comment (gerhardol) [reply GitUI/UserControls/FileStatusList.cs]: Changed commit count and added the following when exceeding

image

**Comment by gerhardol** [comment]
I had to planned to more or less keep the original commit structure, but this is hard after the refactoring to FileStatusDiffCalculator
Is it OK with a squash?
It will be hard to see why particular changes have been done though
(this is why refactoring should not be requested for functionality changes....)

File: GitUI/UserControls/FileStatusWithDescription.cs [hunk 13]
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class FileStatusWithDescription
+    {
+        public GitRevision FirstRev;
+        public GitRevision SecondRev;
+        public ObjectId BaseA;
+        public ObjectId BaseB;
+        public string Summary;
+        public IReadOnlyList<GitItemStatus> Statuses;
+    }
```
RussKie, line 15: The only thing I would prefer is to turn this into an immutable class, i.e.
```suggestion
    public class FileStatusWithDescription
    {
        public FileStatusWithDescription(GitRevision firstRev, GitRevision secondRev, ObjectId baseA, ObjectId baseB, string summary, IReadOnlyList<GitItemStatus> statuses)
        {
            FirstRev = firstRev ?? throw new ArgumentNullException(nameof(firstRev));
            SecondRev = secondRev ?? throw new ArgumentNullException(nameof(secondRev));
            BaseA = baseA ?? throw new ArgumentNullException(nameof(baseA));
            BaseB = baseB ?? throw new ArgumentNullException(nameof(baseB));
            Summary = summary ?? throw new ArgumentNullException(nameof(summary));
            Statuses = new List<GitItemStatus>();
            if (statuses != null)
            {
                Statuses.AddRange(statuses);
            }
        }
        
        public GitRevision FirstRev { get; }
        public GitRevision SecondRev { get; }
        public ObjectId BaseA { get; }
        public ObjectId BaseB { get; }
        public string Summary { get; }
        public IReadOnlyList<GitItemStatus> Statuses { get; }
    }
```

RussKie (commented):
👍

GitUI/UserControls/FileStatusList.cs
gerhardol: Will add one more tweak: Before running the range-diff command, show the command that runs
image

If the command is very slow, you at least see what is going on

mstv:
I think we do not lose too much detail by squashing this PR.
I have no more comments despite immutable `FileStatusWithDescription` with `[CanBeNull] baseA/B` will be good.

GitUI/UserControls/FileStatusWithDescription.cs
gerhardol: Pushed an update, some changes had to be done

> I have no more comments despite immutable FileStatusWithDescription with [CanBeNull] baseA/B will be good.

As well as firstRev for initial (root) commit.

OT: I found later that the initial commit uses the sha e83c5163316f89bfbde7d9ab23ca2e25604af290 within the Git code base, so [CanBeNull] could be changed...
But not here.

mstv (approved):
Neat!

gerhardol:
Squashed changes (and rebased), then separated to refactoring and range-diff specific, so a compromise
(only briefly started the "refactoring" commit but hopefully enough to troubleshoot)

The previous commits are avilable in https://github.com/gerhardol/gitextensions/tree/feature/i8448-range-diff-1

mstv:
I wonder why no range-diff appears anymore (select the commits from bottom to top):

grafik

RussKie:
If this becomes too complicated, we can take in an intermediate state,
assuming it doesn't degrade any other aspects.

Maybe we should consider adding preview flags (like VS does)?

mstv:
> I wonder why no range-diff appears anymore (select the commits from bottom to top):

Not sure what exactly I had tried before. There is no common base. Not critical, especially not blocking.

> If this becomes too complicated, we can take in an intermediate state,

👍 

> assuming it doesn't degrade any other aspects.

very unlikely

> Maybe we should consider adding preview flags (like VS does)?

Not necessary for this. I am afraid this would complicate things even more.

gerhardol:
> If this becomes too complicated, we can take in an intermediate state,
assuming it doesn't degrade any other aspects.

range-diff itself is quite simple, but the handling to prune it for not appearing is a little complex (the command can be slow)
This especially applies to 4 selections where you need to select 2 ranges: https://git-scm.com/docs/git-range-diff
A note will be added to the documentation about this

> I wonder why no range-diff appears anymore (select the commits from bottom to top):

First, the commits need a common ancestor. I did not expect 251, 1c1 to have that but apparently they have, the check is too restrictive with 4 selections
Second, with 4 selections, ranges must be selected
In this scenario you should have been able to select be7, 251, 851, 1c1

If the common merge base check is changed, you still get (you see that Git finds a common ancestor 251, 1c1)
```
# The symmetric difference from 251f7f04 to 1c1b35d9 is 14112 which is higher than the limit 100
# Git range-diff may take a long time and Git Extensions seem to hang during execution, why the command is not executed
# You can still run the command in a Git terminal
# Remove '--no-patch' to see changes to files too
git range-diff be73215b520937a1afdda28d5c5523718192356c..251f7f04dc442e3ee7b46ef6eea7f096bbc462f3 851a92de0e18196ad148505050995e1b8d6c9cd6..1c1b35d97db9d36072aed2b63a2b9c6d99c12489 --no-patch
```

The check for ranges is a little to restrictive with four, if changing that I get

```
1:  1c1b = 1:  1c1b Second delivery
```

The presentation still makes me confused, but merges are done and merges are evil

Pushing an update

mstv:
> Pushing an update

👍 Thank you.

> but merges are done and merges are evil

This is a very simple example of a "vendor branch" for tracking the integration of update deliveries from a vendor.
Believe me, git merge is a great help for this usecase, especially if the deliveries must be adapted / patched.

The second check whether the delivery has been correctly integrated (other pair of ranges in the parallelogram), is handled well by GE but a challenge for `git range-diff`:

grafik

```
$ git range-diff 74782f90baeb1cd2b86d3814186e93b57b5854e7..be73215b520937a1afdda28d5c5523718192356c 1c1b35d97db9d36072aed2b63a2b9c6d99c12489..251f7f04dc442e3ee7b46ef6eea7f096bbc462f3 --no-patch
    -:  --------- >     1:  a6a4eab40 test
    -:  --------- >     2:  bfcbc832f added files
...
    -:  --------- > 10988:  b1ba97457 Push multiple: Set default push to false
    -:  --------- > 10989:  eb120cd82 Added myself as a contributor.
    1:  d8d226860 = 10990:  d8d226860 FormPush: Add context menu to easily manage multiple push selection
    2:  fc459d4c4 = 10991:  fc459d4c4 Apply action to all selected files in solve merge conflict
    3:  2664f26dd = 10992:  2664f26dd Use SelectedNode property to track target for context menu
```

gerhardol:
> The second check whether the delivery has been correctly integrated (other pair of ranges in the parallelogram), is handled well by GE but a challenge for git range-diff

range-diff can certainly be improved - the patch output itself is not unreliable (hard to interpret too - this is complex). The speed will be a problem, has cubic runtime complexity will be slow with many commits.

GE could also be improved with async diff (not really needed to await the diff and showing it) or to even calculate when expanding group. But that is a separate issue PR, not just for range-diff.

Back to this PR: I feel that the current implementation is using range-diff as it is now with some user hints. Any more suggestions for this or follow up PRs?
(The presentation part can be tweaked still, that is in #8516.)
The range selection comes to mind (obviously GitDoc should be updated).

mstv (approved):
Thank you

gerhardol:
Rebased and squashed. Separated the context menu commit to a separate, as it applied as much to CombinedDiff as to RangeDiff (really not a new problem)
Leaving a few days for @RussKie to comment and block merge, will probably rebase before the merge

gerhardol:
@msftbot merge in 3 days

tomasfejfar:
This feature is awesome! Thanks so much. So far I've had my internal alias implementation in CLI, but having it in GitExtensions is great! 🎉 🏅

Issue #8441, opened 2020-09-09
Add shortcut for "Show only this branch"


adamshakhabov:
## Feature description
I use "Show only this branch" command a lot of and currently, to achieve it the below steps should be executed:
1) Hover on the branch name;
2) Right-click;
3) Hover on "Show only this branch" label;
4) Left-Click.

![image](https://user-images.githubusercontent.com/5969304/92585730-04f2d880-f29e-11ea-9032-69b4e686327f.png)

Would be good 
- to have some shortcut, for example, **Shift+Left-Click on the branch name** which displays commit history only of selected branch.
-  or to have a new checkbox setting like **Show only this branch by left mouse click** whose checked value by default force to use "Show only this branch" when a user clicks on a branch name.

## Environment

- Git Extensions 3.4.3.9999
- Build d4b0f48bbf3e39a71f5d7ff5231be5678d4aa0f1
- Git 2.28.0.windows.1
- Microsoft Windows NT 10.0.14393.0
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

RussKie:
> * or to have a new checkbox setting like **Show this branch only by left mouse click** whose checked value by default force to use "Show this branch only" when a user clicks on a branch name.

I don't think this is a good idea, users will turn it on and forget, or turn it on my mistake, and will find the behavior frustrating. We had similar issues with "show first parents" option in the past.

> * to have some shortcut, for example, **Shift+Left-Click on the branch name** which displays commit history only of selected branch.

This option sounds better, though I'm unsure about <kbd>SHIFT</kbd>, it is used as a conjunctive key, i.e. "add to a selection".
May be <kbd>ALT</kbd> + <kbd>Left MB</kbd>? 🤔

adamshakhabov:
Yep, `ALT` + `Left MB` is also good.

adamshakhabov:
The issue can be assigned to me.

gerhardol:
Note that #8452 allocates <kbd>ALT</kbd>+<kbd>Left MB</kbd> to _grid_ events for _Highlight selected branch_ (<kbd>CTRL</kbd>+<kbd>Shift</kbd>+<kbd>B</kbd>), this was about _Show current branch only_ (<kbd>CTRL</kbd>+<kbd>Shift</kbd>+<kbd>U</kbd>).

The handling from left panel should be consistent.
[...]

PR #8479, opened 2020-09-25
#8441: Add shortcut for "Show this branch only"

adamshakhabov:
Adding <kbd>CTRL+Left MB</kbd>  shortcut for "**Show this branch only**" command.
___

The feature is not completely finished. I am a little be stuck on the selecting top branch's revision after revision grid has been filtered. On the below gif you can see that it does not always select last revision after filtered:

GE-8441

It seems I made changes in the not right place or some additional changes should be added.
___

Could someone just point me?


adamshakhabov:
Sorry, I forget rename branch😬, it should be `8441-add-shortcut-show-this-branch-only`.

GintasS:
This is an old PR and should be closed. @RussKie

RussKie:
Thank you, but the team kind of stalled on this. There are changes coming to the filters toolbar, that should make it a little easier.
I'm hoping that we'll also reach a decision on the best short cut as well at some point (/cc: @mstv).

Closing for now.

PR #8482, opened 2020-09-26
Show empty diffs as patch, not text

### LINKED ISSUE #5547: Bug: Diff Pane blank when Ignore Whitespace hides whitespace-only diffs
Problem:
When "Ignore...whitespace" options are enabled, and a file ONLY has whitespace changes that would be ignored by the rule, then the diff pane appears completely blank. Although hovering on the diff pane displays the buttons to disable the Ignore feature, it is not clear this is what is happening, and it may appear broken to the user.

Sidenote:
Oddly, neither of the Ignore Whitespace options seemed not to have ANY effect when I added TWO newlines to the end of a file that previously ended without a newline..

Suggested Resolution:
Add a message in the Diff Preview pane whenever the "Ignore...whitespace" options cause the pane to appear blank, indicating that whitespace-only changes are currently being ignored.

Environment:
Windows 10 v1709 (build 16299.665)
GitExtensions 2.51.05
GIT v2.18.0.windows.1

---

gitextensions-bug-screenshot-20181010

gerhardol:
Fixes #5547

## Proposed changes
* A null patch was displayed as a text file instead of a patch, so the toolbar was missing

### Before

image

### After

image

## Test methodology 

Manual
----

### DIFF
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -74,10 +74,11 @@ namespace GitUI
                 return fileViewer.ViewGitItemRevisionAsync(item.Item, item.SecondRevision.ObjectId, openWithDiffTool);
             }
 
-            string selectedPatch = GetSelectedPatch(fileViewer, firstId, item.SecondRevision.ObjectId, item.Item);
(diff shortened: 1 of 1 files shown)


pmiossec (approved):

gerhardol:
@msftbot merge in 24 hours

RussKie (approved):

PR #8483, opened 2020-09-26
FileViewer: Show toolbar for all files

### LINKED ISSUE #8248: Floating panel in source view
## Feature description

Diff view has a floating panel which allows user to configure the diff and display settings (including charset selection). Bug diff view not used for showing changes of new files and for View tab in file history.
изображение

Of course, some buttons in floating panel not applicable to the source view, but buttons "Show nonprinting characters", "Show syntax hilighting"  and (most important) codepage selection combobox are applicable in source view. Right now there are no any way to switch codepage for new files in commit.
So, source view need a floating panel to configure the source view settings

## Environment

- Git Extensions 3.4.1.9675
- Git 2.23.0.windows.1 (recommended: 2.25.1 or later)
- .NET Framework 4.8.4069.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8248 
~~Related to #5547~~
~~Based on #8496~~

## Proposed changes
* The toolbar should be available also for normal textfiles, so file encoding can be set. (There are context menu actions for other actions.)

Applies for all file viewers, not just diff tab

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -141,12 +141,9 @@ namespace GitUI.Editor
 
             internalFileViewer.MouseMove += (_, e) =>
             {
-                if (IsDiffView(_viewMode) && !fileviewerToolbar.Visible)
-                {
-                    fileviewerToolbar.Visible = true;
-                    fileviewerToolbar.Location = new Point(Width - fileviewerToolbar.Width - 40, 0);
-                    fileviewerToolbar.BringToFront();
-                }
+                fileviewerToolbar.Visible = true;
+                fileviewerToolbar.Location = new Point(Width - fileviewerToolbar.Width - 40, 0);
+                fileviewerToolbar.BringToFront();
             };
             internalFileViewer.MouseLeave += (_, e) =>
             {


RussKie (approved):

RussKie:
> image

This looks weird

RussKie:
> This is also preparing for a change in #8841

We haven't gotten that far yet :) Assume typo, but can't figure out which one you meant to reference.

gerhardol:
> This looks weird

The workaround to set file encoding is to go to another file. This could maybe be addressed by keeping the value persisted.
Setting to draft.

> > This is also preparing for a change in #8841
> We haven't gotten that far yet :) Assume typo, but can't figure out which one you meant to reference.

corrected...
GitHub pull requests is another example where the toolbar need to be customized (lines of context and whitespace has no effect). But that requires more changes, separately.

gerhardol:
Closing this, will therefore close #8248 too
You will have to jump to another file and set the encoding

mstv:
> Closing this

Why? I have not reviewed it because it is in state "Draft" yet. I thought you were thinking about a solution for the resulting empty space.

mstv:
It should be shown even if there is no file in order to be able to turn off "ignore whitespace".

gerhardol:
> Why? I have not reviewed it because it is in state "Draft" yet. I thought you were thinking about a solution for the resulting empty space.

RussKie did not like this, I sometimes want this but sometimes get annoyed myself by the toolbar (in my private builds).

Another solution could be to select encoding in the context menu instead, but that would be a different PR. (I do not plan to make that soon and there are too many PRs open).

> It should be shown even if there is no file in order to be able to turn off "ignore whitespace".

#8482 you mean?
This is for text files

mstv:
>> Why? I have not reviewed it because it is in state "Draft" yet. I thought you were thinking about a solution for the resulting empty space.

> RussKie did not like this, I sometimes want this but sometimes get annoyed myself by the toolbar (in my private builds).

It disappears (and reappears) quickly with MouseOver -- not with the pesky delays as e.g. in the Windows taskbar.

>> It should be shown even if there is no file in order to be able to turn off "ignore whitespace".

> #8482 you mean?

Perfect, thank you.

gerhardol:
Merge or close.
If a context menu is not preferred, close #8248 too

RussKie:
What if.... we permanently showed the toolbar above the viewer? Like a proper toolbar, not floating one.... We could also show more items on it that are present in the context menu too... 🤔

RussKie:
Coupled with FR https://github.com/gitextensions/gitextensions/issues/8683 I think we should show the toolbar permanently

gerhardol:
> Coupled with FR #8683 I think we should show the toolbar permanently

#8683 is still about diffs, this is about text viewers so from the toolbar perspective they are unrelated.
I do not believe I want to always show the toolbar, especially not for text files. Context menu should be enough here.

But both issues are not convenient to handle from the context menu, the solution could be similar.

This should probably be replaced with a possibility to set the encoding from the context menu though.
@mstv?

mstv:
I like the compromise with the floating toolbar, especially with regard to the limited size of the Diff tab in FormBrowse.
I do not see why it would need to be permanently showed if further functions, e.g. #8683, are added.
Though for this PR, i.e. for normal files, the toolbar size should be shrinked.

gerhardol:
> Though for this PR, i.e. for normal files, the toolbar size should be shrinked.

The toolbar has only two items for normal files.
The unneeded separators should be removed in this PR (I did not notice them before).

GintasS:

gerhardol:
Closing this, at least for now

PR #8485, opened 2020-09-27
fix: Branch filter isn't truly reset

### LINKED ISSUE #8484: Branch filter isn't truly reset
Originally reported by @ivangrek in https://github.com/gitextensions/gitextensions/pull/8414#issuecomment-699523919.

## Current behaviour

Whenever a branch filter is applied via the `FormBrowse` toolstrip, and then the filter textbox is cleared, the filter isn't reset.

The bug appears to be here:
https://github.com/gitextensions/gitextensions/blob/4b8024547f58f96802475cf5b448bccb887fb14c/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs#L513-L523

## Steps to reproduce

- Filter by branch
- Clear filter
- Select "Show filtered branches"

image

- Filter by commit message
- Select "Show filtered branches"
- Select "Show all branches"

image

Appears to be a bug from day 1

RussKie:
Fixes #8484

## Proposed changes

Whenever the filter value is being cleared in the UI, the filter isn't getting reset in the underlying plumbing.
The fix resets the filter.

This unblocks #8414 

## Test methodology 

- manual
- integration tests

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -12,20 +12,24 @@ using Microsoft.VisualStudio.Threading;
 
 namespace GitCommands
 {
+#pragma warning disable SA1025 // Code should not contain multiple whitespace in a row
     [Flags]
     public enum RefFilterOptions
     {
-        Branches = 1,              // --branches
-        Remotes = 2,               // --remotes
-        Tags = 4,                  // --tags
-        Stashes = 8,               //
-        All = 15,                  // --all
-        Boundary = 16,             // --boundary
-        ShowGitNotes = 32,         // --not --glob=notes --not
-        NoMerges = 64,             // --no-merges
-        FirstParent = 128,         // --first-parent
-        SimplifyByDecoration = 256 // --simplify-by-decoration
+        None                    = 0x000,
+        Branches                = 0x001,    // --branches
+        Remotes                 = 0x002,    // --remotes
+        Tags                    = 0x004,    // --tags
+        Stashes                 = 0x008,    //
+        All                     = 0x00F,    // --all
+        Boundary                = 0x010,    // --boundary
+        ShowGitNotes            = 0x020,    // --not --glob=notes --not
+        NoMerges                = 0x040,    // --no-merges
+        FirstParent             = 0x080,    // --first-parent
+        SimplifyByDecoration    = 0x100,    // --simplify-by-decoration
+        Reflogs                 = 0x200,    // --reflog
     }
+#pragma warning restore SA1025 // Code should not contain multiple whitespace in a row
 
     public sealed class RevisionReader : IDisposable
     {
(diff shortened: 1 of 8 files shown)


RussKie:
@ivangrek please have a look

RussKie:
@msftbot merge in 36 hours

ivangrek:
Select "Show filtered branches"

Not equals

Change repository
Show this branch only
Change repository
Select "Show filtered branches"

I guess we need last changes from #8414 also.

image

BranchFilter.Text is not cleared now.

ivangrek:
Select "Show filtered branches"
With no filter selection.

At first time shows all branches. It is right?

RussKie:
> BranchFilter.Text is not cleared now.

Does https://github.com/gitextensions/gitextensions/pull/8489 solve it?

RussKie:
> Select "Show filtered branches"
> With no filter selection.
> 
> At first time shows all branches. It is right?

I presume so. I don't use filters, but I think I will, once filtering is made more usable.
If there is no filter what else would you want to show?

ivangrek:
> Does #8489 solve it?

Passed.

RussKie:
> > Does #8489 solve it?
> 
> Passed.

Awesome! Is there anything else outstanding or missing?

ivangrek:
Show filtered branches

Not equals

Show this branch only
Show all branches
Clear branches filter
Show filtered branches

ivangrek:
Is it right?

image

RussKie:
You're good! :D How did you end up in this state?

ivangrek:
> How did you end up in this state?

Show this branch only
Show all branches
Clear branches filter
Show filtered branches

After last step сheckbox not checked on "Show filtered branches".

Show filtered branches (Again)

RussKie:
> Is it right?
> 
> image
>
> ...
> Clear branches filter
> ...

If you do this why does the filter still has text?

ivangrek:
> If you do this why does the filter still has text?

Show this branch only

Work too.

RussKie:
Are you available for a chat? If so, drop me a line on my email, and I'll set up a call.

RussKie:
One thing I can tell for sure filtering is utterly broken - the menu options available via the toolstrip are executing directly on the revision grid, which doesn't have access to the toolstrip filter textbox. Actions invoked from the left panel executed via filter proxy that does have access to the toolstrip filter textbox....
So in some cases when the filter is reset - the filter textbox isn't getting cleared. I can't see an easy way of fixing it. 

Ultimately I want to remove the toolstrip filter textbox completely, and allow branch filtering via the left panel. Related discussion in #8441.

RussKie:
I pushed some crutches in, but boy is it ugly...

I think it should work now, but would be great to cross check. I'll add some more tests, once we settle on the behaviour.
Thanks for being my tester.

PR #8487, opened 2020-09-27
fix: Resolve script "c"-variables

### LINKED ISSUE #8486: Can't get correct commit message using {cMessage}
## Current behaviour
I want to execute a python script after pushing a commit to Gerrit, so I set a script.
But I get empty value by {cMessage} after push commit, and I can get only first line of commit message while right click and
choose Run Script. 

## Expected behaviour
I can get full commit message after push commit or just run script
## Steps to reproduce
Following are the script I use, just in test stage.
```python
import sys

def main():
    print sys.argv[1];

main()
```

## Screenshots
Settings
image
Empty after push
image
First line using "Run Script"
image

full commit message are:

NEWGS-12330 China - gcc.controller.impl.LoginControllerImpl.log

Issue: NEWGS-12330
Type: Internal Bug
Change-Id: I4ba4d7438affe48c9340125b8853136d4888e3e4

## Environment

- Git Extensions 3.4.3.9999
- Git 2.27.0.windows.1
- .NET Framework 4.8.4240.0
- DPI 96dpi (no scaling)

RussKie:
Fixes part of #8486

## Proposed changes

Whenever a script, run outside the revision grid, needed to resolve
a "c"-variable, it would  and Original implementation would retrieve
the current checkout information, that didn't have enough information
available to process a requested variable.

Replace the call with `GetRevision` call, that is made by the revision grid,
and which has all information about a revision (except for the full body).
With that consolidate the revision retrieval logic.

Expose `GetRevision` method at the `IGitModule` level, and move
`GitRevision`type to GitUIPluginInterfaces, so it is universally available.

## Test methodology 

- manual
- unit tests

### DIFF
--- a/GitCommands/ExternalLinks/ExternalLinkFormat.cs
+++ b/GitCommands/ExternalLinks/ExternalLinkFormat.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Xml.Serialization;
+using GitUIPluginInterfaces;
 
 namespace GitCommands.ExternalLinks
 {
--- a/GitCommands/ExternalLinks/ExternalLinkRevisionParser.cs
+++ b/GitCommands/ExternalLinks/ExternalLinkRevisionParser.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 using GitCommands.Remotes;
+using GitUIPluginInterfaces;
 
 namespace GitCommands.ExternalLinks
 {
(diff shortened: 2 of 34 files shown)


mstv (approved):
The code & tests LGTM. 
But this PR does not fix the whole issue. I have adapted the "Fixes" clause in the PR description.
The IntegrationTests hang.

RussKie:
> But this PR does not fix the whole issue. I have adapted the "Fixes" clause in the PR description.

Thank you.



> The IntegrationTests hang.

Hmm... It did hang intermittently for me locally as well.

mstv:
I have fixed the broken test which popped up a MessageBox.

RussKie:
Awesome, thank you so much.
Not sure how I missed it.

PR #8488, opened 2020-09-27
Add script options {sSubject} and {cSubject}

### LINKED ISSUE #8486: Can't get correct commit message using {cMessage}

mstv:
Fixes #8486
together with #8487, on which this PR is based.

## Proposed changes

- add script options `{cSubject}` and `{sSubject}`
- change script options `{cMessage}` and `{sMessage}`to yield the full message
  The linefeed characters are replaced with `\n`.

## Test methodology 

- adapted and added NUnit tests
- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 8bc7355730d48b869ea0389bed8ffd59f3d77695
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Script/ScriptOptionsParser.cs
+++ b/GitUI/Script/ScriptOptionsParser.cs
@@ -15,6 +15,11 @@ namespace GitUI.Script
     public sealed class ScriptOptionsParser
     {
         /// <summary>
+        /// Name of the option which requires the full commit message.
+        /// </summary>
+        private const string currentMessage = "cMessage";
+
+        /// <summary>
         /// Gets the list of available script options.
         /// </summary>
         public static readonly IReadOnlyList<string> Options = new[]
@@ -30,6 +35,7 @@ namespace GitUI.Script
             "sRemotePathFromUrl",
             "sHash",
             "sMessage",
+            "sSubject",
             "sAuthor",
             "sCommitter",
             "sAuthorDate",
@@ -40,7 +46,8 @@ namespace GitUI.Script
             "cRemoteBranch",
             "cRemoteBranchName",
             "cHash",
-            "cMessage",
+            currentMessage,
+            "cSubject",
             "cAuthor",
             "cCommitter",
             "cAuthorDate",
@@ -113,7 +120,8 @@ namespace GitUI.Script
 
                 if (currentRevision == null && option.StartsWith("c"))
                 {
-                    currentRevision = GetCurrentRevision(module, currentTags, currentLocalBranches, currentRemoteBranches, currentBranches);
+                    currentRevision = GetCurrentRevision(module, currentTags, currentLocalBranches, currentRemoteBranches, currentBranches,
+                        loadBody: Contains(arguments, currentMessage));
                     if (currentRevision == null)
                     {
                         return (arguments: null, abort: true);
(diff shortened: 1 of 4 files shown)


GitUI/Script/ScriptOptionsParser.cs
```diff
@@ -236,6 +238,10 @@ private static string CreateOption([NotNull] string option, bool quoted)
             return result;
         }
 
+        [CanBeNull]
+        private static string EscapeLinefeeds([CanBeNull] string multiLine)
```
RussKie, line 242:
```suggestion
        private static string EscapeLineFeeds([CanBeNull] string multiLine)
```

Maybe turn into a local static function?

GitUI/Script/ScriptOptionsParser.cs
```diff
@@ -390,6 +400,10 @@ private static string GetRemotePath(string url)
                     break;
 
                 case "cMessage":
+                    newString = EscapeLinefeeds(currentRevision.Body) ?? currentRevision.Subject;
```
RussKie, line 403: FWIW, I don't think `module.GetRevision()` loads the body of a commit because it is expensive.

RussKie (commented):
Please rebase

GitUI/Script/ScriptOptionsParser.cs
mstv: done

GitUI/Script/ScriptOptionsParser.cs
mstv: Yes, I tried to test it because you had mentioned this. But I trapped into the pit with multiple GE instances overwriting the settings of each other. I will trigger another or modified command if `{cMessage}` is used.

UnitTests/GitUI.Tests/Script/ScriptOptionsParserTests.cs
```diff
@@ -152,15 +152,38 @@ public void Parse_should_return_unmodified_arguments_if_no_options_matched()
...
+                Body = $"{Subject}\n\nline3"
             };
-            _module.GetRevision(shortFormat: true, loadRefs: true).Returns(x => revision);
+            _module.GetRevision(shortFormat: false, loadRefs: true).Returns(x => revision);
+
+            string expectedMessage = $"{Subject}\\n\\nline3";
+
+            var result = ScriptOptionsParser.Parse("echo {{cSubject}} {{cMessage}}", module: _module, owner: null, scriptHostControl: null);
+
+            result.arguments.Should().Be($"echo \"{revision.Subject}\" \"{expectedMessage}\"");
```
RussKie, line 167:
```suggestion
            result.arguments.Should().Be($"echo \"{revision.Subject}\" \"{revision.Body}\"");
```
...and remove `expectedMessage`.

RussKie (approved):

UnitTests/GitUI.Tests/Script/ScriptOptionsParserTests.cs
mstv: This cannot work with `revision.Body` because it contains linefeeds but in `expectedMessage` they are escaped as two characters `\n`.

PR #8489, opened 2020-09-27
Reset filter when switch repos

RussKie:
## Proposed changes

- Reset filter when switch repos

### Before

ge-filter-reset

### After

ge-filter-reset2

## Test methodology 

- manual

----


jrfnl:
Thanks for fixing this! Ran into this plenty of times and barely use the filter now due to this issue. Looking forward to being able to use it again.

PR #8490, opened 2020-09-27
Dialogs layout alignment - `FormCreateBranch`

RussKie:
Relates to #6183

## Screenshots 

* Before:
image

* After
image

❗ Please note "Help" button will be removed in the follow up PRs to reduce the wasted space at the bottom.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCreateBranch.Designer.cs
@@ -42,10 +42,17 @@
             this.groupBox1 = new System.Windows.Forms.GroupBox();
             this.flowLayoutPanel1 = new System.Windows.Forms.FlowLayoutPanel();
             this.gotoUserManualControl1 = new GitUI.UserControls.GotoUserManualControl();
+            this.MainPanel.SuspendLayout();
             this.tableLayoutPanel1.SuspendLayout();
             this.groupBox1.SuspendLayout();
             this.flowLayoutPanel1.SuspendLayout();
             this.SuspendLayout();
+            // 
+            // MainPanel
+            // 
+            this.MainPanel.Controls.Add(this.Ok);
+            this.MainPanel.Controls.Add(this.tableLayoutPanel1);
+            this.MainPanel.Size = new System.Drawing.Size(480, 203);
             // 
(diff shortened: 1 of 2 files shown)


RussKie:
@NikolayXHD need your help here - the colour of "Clear working directory and index" checkbox is too dark with the new colours. Is this something we're calculating dynamically?

NikolayXHD:
> @NikolayXHD need your help here - the colour of "Clear working directory and index" checkbox is too dark with the new colours. Is this something we're calculating dynamically?

From what I can see by debugging, the color for disabled text is calculated dynamically by WinForms

https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs,675

image

NikolayXHD:
A proper workaround would be something like

```csharp
// GitExtensionsDialog.cs
if (ColorHelper.IsDarkTheme())
  this.MainPanel.BackgroundColor = SystemColors.Control;
```

mstv:
I don't like the large borders in the newly aligned dialogs. The dialogs cover too much of the main window without reason.

RussKie:
> From what I can see by debugging, the color for disabled text is calculated dynamically by WinForms
> 
> [source.dot.net/#System.Windows.Forms/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs,675](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs,675)
> 
> image
> 
> Like the code above shows, if the background does not have magic value `SystemColors.Control`, the disabled text color will be calculated by making background color darker, regardless of whether the background is bright or dark, which produces near-black text color on dark gray background

@NikolayXHD omg 😱 wow

@merriemcgaw @jeremykuhne we have work to do 😉

RussKie:
> I don't like the large borders in the newly aligned dialogs. The dialogs cover too much of the main window without reason.

There's a dedicated thread for this discussion, let's move it there: https://github.com/gitextensions/gitextensions/pull/8409

RussKie:
Scratch that, the discussion is in https://github.com/gitextensions/gitextensions/issues/6183.

PR #8499, opened 2020-09-30
Fix JumpList being created too soon

### LINKED ISSUE #8421: JumpList still not working
At least on my machine (Win10 1709 16299.2045), the fix in #7941 doesn't work.

While we may expect the main window handle to exist when there is a control handle, this is not the case : in `FormBrowse.OnLoad()`, the `Handle` property is non-zero, but 
`Process.GetCurrentProcess().MainWindowHandle` *is* zero.

I tried moving the call to `_windowsJumpListManager.CreateJumpList()` in `FormBrowse.OnActivated()` instead, and this works. However, I'm not sure this is a correct fix, as this has at least one serious drawback : the `AddToRecent()` call doesn't work if directly launching a repo, especially from the Shell extension in Explorer, since the JumpList hasn't been created yet at that point.

lanfeust69:
Fixes #8421

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -579,12 +579,6 @@ namespace GitUI.CommandsDialogs
 
         protected override void OnLoad(EventArgs e)
         {
-            _windowsJumpListManager.CreateJumpList(
-                Handle,
-                new WindowsThumbnailToolbarButtons(
-                    new WindowsThumbnailToolbarButton(toolStripButtonCommit.Text, toolStripButtonCommit.Image, CommitToolStripMenuItemClick),
-                    new WindowsThumbnailToolbarButton(toolStripButtonPush.Text, toolStripButtonPush.Image, PushToolStripMenuItemClick),
-                    new WindowsThumbnailToolbarButton(toolStripButtonPull.Text, toolStripButtonPull.Image, PullToolStripMenuItemClick)));
             SetSplitterPositions();
             HideVariableMainMenuItems();
             RefreshSplitViewLayout();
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -630,6 +624,17 @@ protected override void OnLoad(EventArgs e)
 
         protected override void OnActivated(EventArgs e)
         {
+            // wait for windows to really be displayed, which isn't necessarily the case in OnLoad()
+            if (_windowsJumpListManager.NeedsJumpListCreation)
```
mstv, line 628: This check does not need to be done here, too, because it is the first thing what `CreateJumpList` does.
So it can be avoided to expose `NeedsJumpListCreation`.

GitUI/WindowsJumpListManager.cs
```diff
@@ -146,6 +158,13 @@ public void CreateJumpList(IntPtr windowHandle, WindowsThumbnailToolbarButtons b
                 CreateTaskbarButtons(windowHandle, buttons);
             }, nameof(CreateJumpList));
 
+            if (ToolbarButtonsCreated && _deferredAddToRecent != null)
+            {
+                var toAdd = _deferredAddToRecent;
+                _deferredAddToRecent = null;
+                AddToRecent(toAdd);
```
mstv, line 165: I think we do not need thread-safety here (if, `Interlocked.Exchange` should be used).
So the temporary variable `toAdd` can be omitted.

mstv (commented):
Should work, have not run it.

GitUI/CommandsDialogs/FormBrowse.cs
lanfeust69: Indeed, this wasn't in my first version. However, the call creates three `WindowsThumbnailToolbarButton`s, and I wasn't sure it wouldn't need a handle or other somewhat "expensive" resource (the allocation itself is surely not an issue).

GitUI/WindowsJumpListManager.cs
lanfeust69: Yes, this is an "in between" approach. Not quite thread-safe, as it's no big deal if by extraordinary `AddToRecent` is called twice, but still much less likely if the time-window doesn't include an IO-bound call. Not convinced it's worth the hassle, though...

GitUI/WindowsJumpListManager.cs
```diff
@@ -146,6 +158,13 @@ public void CreateJumpList(IntPtr windowHandle, WindowsThumbnailToolbarButtons b
                 CreateTaskbarButtons(windowHandle, buttons);
             }, nameof(CreateJumpList));
 
+            if (ToolbarButtonsCreated && _deferredAddToRecent != null)
+            {
+                var toAdd = _deferredAddToRecent;
```
RussKie, line 163: Make more explicit
```suggestion
                string recentRepoAddToRecent = _deferredAddToRecent;
```

RussKie (commented):
👍 works for me

mstv (approved):
Please fill in the PR description. We have a template. At least the very helpful "Fixes #8421" and the DCO clause should be there.

lanfeust69:
> Please fill in the PR description. We have a template. At least the very helpful "Fixes #8421" and the DCO clause should be there.

Indeed, I created the PR through GitExtensions itself, and it didn't use the template 🙂 

Note that the "Fixes" part was already in the commit message, though.

RussKie (approved):

RussKie:
Thank you

PR #8503, opened 2020-10-02
Correct Sort context menu placement

RussKie:
Relates to #8427

## Proposed changes

- Ensure correct placement of sort-related menus - occasionally the menu would be placed at a wrong index.

### Before

image

### After

image

## Test methodology 

- manual
- tests


RussKie:
Merging to unblock all other left panel work.
Will follow up on reviews, if any.

mstv (commented):
👍

PR #8504, opened 2020-10-02
Parallelized Repo Validity Checks during Startup

### LINKED ISSUE #8337: GE start hangs and non-responding after long time
## Current behaviour

GE start hangs and non-responding after long time

## Expected behaviour
GE should start immediately.

## Steps to reproduce

Suspect this may have to do with previous repositories (using network drive) if they're no longer available.

## Screenshots

image

## Environment

image

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version:

digitalcoyote:
This should prevent "hanging" during startup

Fixes #8337

## Proposed changes

- Parallelize IsValidGitWorkingDir checks.
  - Using PLINQ
  - Switched out the foreach on repos to a for loop so it could access the validity results by index

## Test methodology

- Tested Locally using Stopwatch to measure the effect on successfully validated repos
- Debugged scenarios involving Invalid Repo(s)
- I was unable to setup any shares to check 

## Test environment(s) 

- GIT 2.28.0.windows.1
- Windows 10 (Lastest as of 10/2/2020)

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -300,23 +300,26 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
 
             void BindRepositories(IReadOnlyList<RecentRepoInfo> repos, bool isFavourite)
             {
-                foreach (var repository in repos)
-                {
-                    var isInvalidRepo = !_controller.IsValidGitWorkingDir(repository.Repo.Path);
-                    _hasInvalidRepos |= isInvalidRepo;
+                var repoValidityArray = repos.AsParallel().Select(r => !_controller.IsValidGitWorkingDir(r.Repo.Path)).ToArray();
 
-                    listView1.Items.Add(new ListViewItem(repository.Caption)
+                _hasInvalidRepos = repoValidityArray.Any();
+
+                for (var index = 0; index < repos.Count; index++)
+                {
+                    listView1.Items.Add(new ListViewItem(repos[index].Caption)
                     {
                         ForeColor = ForeColor,
                         Font = AppSettings.Font,
-                        Group = isFavourite ? GetTileGroup(repository.Repo) : _lvgRecentRepositories,
-                        ImageIndex = isInvalidRepo ? 1 : 0,
+                        Group = isFavourite ? GetTileGroup(repos[index].Repo) : _lvgRecentRepositories,
+                        ImageIndex = repoValidityArray[index] ? 1 : 0,
                         UseItemStyleForSubItems = false,
-                        Tag = repository.Repo,
-                        ToolTipText = repository.Repo.Path,
+                        Tag = repos[index].Repo,
+                        ToolTipText = repos[index].Repo.Path,
                         SubItems =
                         {
-                            { _controller.GetCurrentBranchName(repository.Repo.Path), BranchNameColor, BackColor, _secondaryFont },
+                            {
+                                _controller.GetCurrentBranchName(repos[index].Repo.Path), BranchNameColor, BackColor, _secondaryFont
+                            },
                             //// NB: we can add a 3rd row as well: { repository.Repo.Category, SystemColors.GrayText, BackColor, _secondaryFont }
                         }
                     });


RussKie (commented):
I'd like to see some benchmarking here, as I'm concerned about perf implications of the proposed change.

In the proposal we get additional allocations (e.g. `ToArray()`), and multiple traversals of the underlying collection (e.g. `Select.ToArray()`, `Any()` and the for loop).
Whilst it may not have any noticeable effect in cases where the recent collection isn't large, it may become observable for users with large list of recents and favourites.

Perhaps the `foreach` loop can be turned into a traversal over an async enumerable collection (with associated changes to the underlying collection)?

digitalcoyote:
> I'd like to see some benchmarking here, as I'm concerned about perf implications of the proposed change.

I did some testing and it takes ~ 2-3 time longer to run (from ~8ms to 30ms on some runs). I was a little concerned myself. 

> Perhaps the foreach loop can be turned into a traversal over an async enumerable collection (with associated changes to the underlying collection)?
I'll see what I can do

digitalcoyote:
It appears that AsyncEnums are a .NetStandard 2.1 feature. This project would need to move to .Net Core or .Net 5 to be able to make use of that feature. 

As an Alternative, I looked deferring the validity check instead. And having it invoke the UI to mark any repos as invalid after it checks them in parallel. This should prevent a delay at startup, but does delay informing the ui about invalid repos. It decreases the impact on users who don't have many invalid repos (but costs some CPU usafe/memory). compared to the original PR, I'm seeing it max out at about 20ms with my repos. 

Depending on your thoughts, I'm going to setup a few valid/invalid repos. I'm considering testing with 20-30 valid repos and ~20 invalid repos. If you have suggestion for higher  numbers, let me know.

gerhardol:
> It appears that AsyncEnums are a .NetStandard 2.1 feature. This project would need to move to .Net Core or .Net 5 to be able to make use of that feature.

Update is planned
https://github.com/gitextensions/gitextensions/pull/7162

> Depending on your thoughts, I'm going to setup a few valid/invalid repos. I'm considering testing with 20-30 valid repos and ~20 invalid repos. If you have suggestion for higher numbers, let me know.

Seem OK to me
I use about 50 super repos (each with 0 - 15 submodules) at work (like 25 unique, 25 are work tree but I doubt that matters here)
Maybe it is enough to have like 5 invalid repos
There are for sure users with more repos, but I assume this is sufficient to see any severe problems.

RussKie:
> It appears that AsyncEnums are a .NetStandard 2.1 feature. This project would need to move to .Net Core or .Net 5 to be able to make use of that feature.

Ah... I'm been using .NET 5 features for too long so I'm taking a lot of them for granted now.
As an alternative you can consider trying on https://github.com/gitextensions/gitextensions/pull/8522 (a rough WIP on .NET 5.0, a few things broken, but working overall)

digitalcoyote:
I went ahead and checked out the .Net 5 branch and wrote up an async enumerable version of this change. was much slower due to that requiring BindRepositories to be async. The extra plumbing required to make sure the listview items were on the UI Thread had a bit of a negative impact. I'm going to start setting up repos for the test I mentioned above. Will probably be tomorrow or the next day after work before I can sit down and take some measurements.

RussKie:
The whole thing should probably be burned down and re-written from a scratch.... I rewritten the dashboard, but I reused a lot of existing code.

RussKie:
> Original Code is about 9ms for my own repos, and about 20ms for 30 (current max on recent repos). The code I have above is about 30ms with my repos, and about 40ms with 30 repos.

So the PR is making it slower? I'm confused, sorry.

digitalcoyote:
> So the PR is making it slower? I'm confused, sorry.

That was done in a Windows VM with only 2 logical processors (1 core) on a Linux Dev Machine. I booted over to my Windows install, updated for several hours, and tested again. In addition, I did tests with repos on Shares that are no longer connected. The additional plumbing that its doing behind the scenes to add in those extra threads makes it slower by ~10-20ms when everything is valid. But, it prevents hanging up (or at least shortens the hang) when repos on network shares are missing.

### 30 Invalid Network Shares
Old Code on a single core: ~ 30 seconds
PR Code on a single core: ~15 seconds

Old Code on 8-core: ~30 seconds
PR Code on 8-core: ~2.5 seconds

 It and it appears that it takes ~ 2-2.5 seconds for it to time out when looking for a network share. 

### 30 Repos not on network shares  (valid or not seems not to matter much)
Old Code on single core: 20-30ms
PR Code on single core: 30-40ms

Old Code on 8-core: 20-30ms
PR Code on 8-core: 30-40ms

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -300,23 +300,27 @@ public void ShowRecentRepositories()
...
-                        Tag = repository.Repo,
-                        ToolTipText = repository.Repo.Path,
+                        Tag = repos[index].Repo,
+                        ToolTipText = repos[index].Repo.Path,
                         SubItems =
                         {
-                            { _controller.GetCurrentBranchName(repository.Repo.Path), BranchNameColor, BackColor, _secondaryFont },
+                            {
+                                _controller.GetCurrentBranchName(repos[index].Repo.Path), BranchNameColor, BackColor,
+                                _secondaryFont
```
RussKie, line 322: Keep on the same line

RussKie (commented):
👍

RussKie (approved):

RussKie:
Thank you.

Btw it is an anti-pattern to work off the master branch, and branches you haven't created in general.
Doing so you're inviting troubles into your life when remote branch moves and you try to pull changes down.

PR #8510, opened 2020-10-03
Update layout

### LINKED ISSUE #8521: Checkout windows in GitExtensions 4.0 CI build doesn't repond to 'Enter' key
Checkout any revision.
`Enter` key doesn't do anything.
Previously it used to press on Checkout which is the default action.

- Git Extensions 4.0.0.10290
- Build e1b3b0109dd62a12e0c3dbc15a30750d487d4d8f (Dirty)
- Git 2.28.0.windows.1
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

RussKie:
Relates to #6183

Resolves #8521

## Proposed changes

- Reduce margins and paddings by at least 12px in each dimension
    * Update `FormClone`
    * Update `FormInit`
    * Update `FormStatus`
- Fix Accept and Cancel buttons that get unset by adding buttons to the controls panel.

## Screenshots 

| Before | After |
|---|---|
| image | image |
| image | image |
| image | image |
| .. | image |
| .. | image |

----

### DIFF
--- a/GitUI/CommandsDialogs/FormInit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormInit.Designer.cs
@@ -46,12 +46,12 @@
             this.MainPanel.Controls.Add(this.Browse);
             this.MainPanel.Controls.Add(this._NO_TRANSLATE_Directory);
             this.MainPanel.Controls.Add(this.label1);
-            this.MainPanel.Size = new System.Drawing.Size(572, 146);
+            this.MainPanel.Size = new System.Drawing.Size(542, 141);
             // 
             // label1
             // 
             this.label1.AutoSize = true;
-            this.label1.Location = new System.Drawing.Point(27, 24);
+            this.label1.Location = new System.Drawing.Point(12, 17);
             this.label1.Name = "label1";
             this.label1.Size = new System.Drawing.Size(49, 13);
             this.label1.TabIndex = 0;
@@ -62,14 +62,14 @@
             this._NO_TRANSLATE_Directory.AutoCompleteMode = System.Windows.Forms.AutoCompleteMode.SuggestAppend;
             this._NO_TRANSLATE_Directory.AutoCompleteSource = System.Windows.Forms.AutoCompleteSource.FileSystemDirectories;
             this._NO_TRANSLATE_Directory.FormattingEnabled = true;
-            this._NO_TRANSLATE_Directory.Location = new System.Drawing.Point(110, 21);
+            this._NO_TRANSLATE_Directory.Location = new System.Drawing.Point(95, 14);
             this._NO_TRANSLATE_Directory.Name = "_NO_TRANSLATE_Directory";
             this._NO_TRANSLATE_Directory.Size = new System.Drawing.Size(323, 21);
             this._NO_TRANSLATE_Directory.TabIndex = 1;
             // 
             // Browse
             // 
-            this.Browse.Location = new System.Drawing.Point(439, 19);
+            this.Browse.Location = new System.Drawing.Point(424, 12);
             this.Browse.Name = "Browse";
             this.Browse.PathShowingControl = this._NO_TRANSLATE_Directory;
             this.Browse.Size = new System.Drawing.Size(106, 25);
@@ -80,7 +80,7 @@
             // 
             this.groupBox1.Controls.Add(this.Central);
             this.groupBox1.Controls.Add(this.Personal);
-            this.groupBox1.Location = new System.Drawing.Point(27, 48);
+            this.groupBox1.Location = new System.Drawing.Point(12, 41);
             this.groupBox1.Name = "groupBox1";
             this.groupBox1.Size = new System.Drawing.Size(518, 78);
             this.groupBox1.TabIndex = 3;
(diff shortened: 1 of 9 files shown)


GitUI/GitExtensionsDialog.Designer.cs
```diff
@@ -41,21 +41,23 @@ private void InitializeComponent()
...
             this.ControlsPanel.FlowDirection = System.Windows.Forms.FlowDirection.RightToLeft;
-            this.ControlsPanel.Location = new System.Drawing.Point(0, 292);
-            this.ControlsPanel.MinimumSize = new System.Drawing.Size(0, 45);
+            this.ControlsPanel.Location = new System.Drawing.Point(0, 299);
+            this.ControlsPanel.Margin = new System.Windows.Forms.Padding(0);
+            this.ControlsPanel.MinimumSize = new System.Drawing.Size(0, 38);
             this.ControlsPanel.Name = "ControlsPanel";
-            this.ControlsPanel.Padding = new System.Windows.Forms.Padding(8);
-            this.ControlsPanel.Size = new System.Drawing.Size(553, 45);
+            this.ControlsPanel.Padding = new System.Windows.Forms.Padding(7);
```
gerhardol, line 48: Why is padding set to 7, should it not be a multiple of 4?

GitUI/CommandsDialogs/FormInit.Designer.cs
```diff
@@ -46,12 +46,12 @@ private void InitializeComponent()
             this.MainPanel.Controls.Add(this.Browse);
             this.MainPanel.Controls.Add(this._NO_TRANSLATE_Directory);
             this.MainPanel.Controls.Add(this.label1);
-            this.MainPanel.Size = new System.Drawing.Size(572, 146);
+            this.MainPanel.Size = new System.Drawing.Size(560, 140);
```
gerhardol, line 49: Should margin/padding be set for the MainPanel?

gerhardol (commented):
Looks better - I want smaller margins (discussed in the issue) but I will not make it better myself

GitUI/GitExtensionsDialog.Designer.cs
RussKie: Not necessarily. There must be something else in play in the layout... I have no time or interest in debugging this right now.
This aligns the margins with the Task Dialog bottom panel.

GitUI/CommandsDialogs/FormInit.Designer.cs
RussKie: They are set at `GitExtensionsDialog` level, and inherited here.
However in some situations a descendant may override it, e.g. in `FormStatus` - it has a custom layout and overrides `MainPanel.Padding`.

mstv:
I would right-align the bottom buttons with the boxes in the client area.

ivangrek:
image

RussKie:
Yes, 18 is the new reduced padding. The middle between 12 and 24.

Changing layouts of the forms is outside the scope of this PR.

ivangrek:
> Yes, 18 is the new reduced padding. The middle between 12 and 24.

Red color shows indents that do not correspond to the declared 18px.
Other colors indicate areas for possible improvement.

ivangrek:
Change paddings for base ControlsPanel.
https://github.com/ivangrek/gitextensions/commit/1989c246dab223e979525aa5d7d3f6a9913d9853

ivangrek:
For such indents, the button looks better than 32px. For comparison.

image

RussKie:
> Change paddings for base ControlsPanel.
> [ivangrek@1989c24](https://github.com/ivangrek/gitextensions/commit/1989c246dab223e979525aa5d7d3f6a9913d9853)
> 
> image

Thank you, but you have probably missed my reply in https://github.com/gitextensions/gitextensions/issues/6183#issuecomment-703089323.
The spacing in the controls panel is (almost) aligned with the Task Dialog. Increasing the vertical paddings to 18px makes it look odd, and won't go well with either Gerhard or Michael.

ivangrek:
I would then recommend using harmonic values ​​for derived margins: 1:1, 1:2, 1:3, 1:4.
This is also an argument for 12px.

RussKie:
Windows itself is inconsistent (haha!), but it generally doesn't have the same value for v- and h-padding for buttons - the TD is a perfect example of this.

There's another aspect to this - buttons placed at the bottom panel invoke actions that apply to the whole dialog, whilst there may be other action elements that don't. To make this fact more obvious, and for a consistent UX this panel is visually separated from the main content. However the buttons aren't meant to dominate visually and steal user's focus from the main content. 
There's a whole science around UX, useability and "comfortable" spacing (which I'm not sufficiently equipped to have deep discussions about), but I based on my experience and observations giving buttons as much v-padding makes the bottom more prominent that necessary, and thus divert some of the user's attention from the task at hand.

On the other hand, if the main content h-paddings are too small - it makes a dialog visually unpleasant. And hence the recommendations for 12/24 paddings. 
Since a lot of our form are resizable, and generally larger than 640px, making h-paddings 12px doesn't feel like a good UX. Dynamically re-calc paddings feels even worse decision. And since 24px is regarded by few members of the team as "too wide", 18 feels like a good middle ground.

ivangrek:
I thought these tasks were aimed at developing a common approach. And so ok.

gerhardol:
> On the other hand, if the main content h-paddings are too small - it makes a dialog visually unpleasant. And hence the recommendations for 12/24 paddings.
> Since a lot of our form are resizable, and generally larger than 640px, making h-paddings 12px doesn't feel like a good UX. 

This is where we disagre - I do not find 12 unpleasant for large windows

> Dynamically re-calc paddings feels even worse decision. 

yes...

> And since 24px is regarded by few members of the team as "too wide", 18 feels like a good middle ground.

But the recommendation is to have multiple by 4, 18 may scale worse

RussKie:
@msftbot merge in 24 hours

gerhardol:
I can live with this, but still

> 18 may scale worse

RussKie:
Margins are 12px now

gerhardol (approved):

PR #8511, opened 2020-10-03
Don't refresh revision grid invoking toolstrip actions

RussKie:
Appears to be a regression from #8485 which caused the revision grid to be refreshed whenever an action was raised via the toolstrip, e.g. invoking fetch would refresh the grid before the fetch operation was performed, as well as after.

Somehow the toolstrip would pass focus to all its items, that included the Branch Filter which in turn would cause the refresh.
Update the logic to force the grid refresh only when the filter is being changed and applied.

What's worse, multiple refreshes could overlap each other due to peculiar behaviour of the toolstrip, which keeps on passing focus back to its items,
and with that invoke `FilterBranchHelper.toolStripBranches_Leave()`.
There was also another strange behaviour that could lead to a blank revision grid - if a user typed a filter but hasn't applied it, and then invoked
another action via the toolstrip (e.g. fetch) - the revision grid would refresh first after the fetch, and then another refresh operation would
be in flight that attempted to apply the pending filter.

Remove the method and in essence change the user behaviour - now users have to hit ENTER to apply a filter rather than leave the filter textbox.
And move the latch in `ForceRefreshRevisions()` a little higher, to reduce risks of race refreshes.

Also do not refresh the grid when fetching the status from the superproject if the status hasn't changed.

----

### DIFF
--- a/GitUI/FilterBranchHelper.cs
+++ b/GitUI/FilterBranchHelper.cs
@@ -11,6 +11,8 @@ namespace GitUI
     public class FilterBranchHelper : IDisposable
     {
         private bool _applyingFilter;
+        private bool _filterBeingChanged;
+
         private readonly ToolStripComboBox _NO_TRANSLATE_toolStripBranches;
         private readonly RevisionGridControl _NO_TRANSLATE_RevisionGrid;
         private readonly ToolStripMenuItem _localToolStripMenuItem;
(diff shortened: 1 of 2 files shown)


RussKie:
NB: I'm planning to completely remove the branches filter textbox from the toolstrip once we agree on and implement the new branch filtering in the left panel. Related discussions in https://github.com/gitextensions/gitextensions/issues/8441

mstv (approved):
Looks sensible to me, have not run it

RussKie:
> Looks sensible to me, have not run it

I suppose once I merge everyone will run it :D
Some of it is rather a nuanced behaviour, and probably won't affect the great deal of our users, none the less I found it quite annoying.

PR #8512, opened 2020-10-03
Fixed "Show in folder" for paths with comma

m417z:
A simple fix. I didn't test the code, but I tested the command line and it works.

### DIFF
--- a/GitUI/OsShellUtil.cs
+++ b/GitUI/OsShellUtil.cs
@@ -18,7 +18,7 @@ namespace GitUI
 
         public static void SelectPathInFileExplorer(string filePath)
         {
-            Process.Start("explorer.exe", "/select, \"" + filePath + "\"");
+            Process.Start("explorer.exe", $"/select, {filePath.Quote()}");
         }
 
         public static void OpenWithFileExplorer(string filePath)
--- a/GitUI/OsShellUtil.cs
+++ b/GitUI/OsShellUtil.cs
@@ -18,7 +18,7 @@ namespace GitUI
 
         public static void SelectPathInFileExplorer(string filePath)
         {
-            Process.Start("explorer.exe", "/select, " + filePath);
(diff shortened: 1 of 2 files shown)


GitUI/OsShellUtil.cs
```diff
@@ -18,7 +18,7 @@ public static void OpenAs(string file)
 
         public static void SelectPathInFileExplorer(string filePath)
         {
-            Process.Start("explorer.exe", "/select, " + filePath);
+            Process.Start("explorer.exe", "/select, \"" + filePath + "\"");
```
RussKie, line 21: We have an extension method
```suggestion
            Process.Start("explorer.exe", $"/select, {filePath.Quote()}");
```

RussKie:
Please provide your sign off in the contributors.txt

contributors.txt
```diff
@@ -139,3 +139,4 @@ YYYY/MM/DD, github id, Full name, email
 2020/09/09, adamshakhabov, Adam Shakhabov, theadamo86(at)gmail.com
 2020/09/13, hoborm, Mihaly Hobor, mail(at)hobor.xyz
 2020/07/11, blitzmann, Ryan Holmes, holmes.ryan.90(at)gmail.com
+2020/10/03, m417z, ,
```
gerhardol, line 142: Name and email is required (nor use special or ask someone else to use the patch)

gerhardol (changes requested):

RussKie:
I've tested the current release, and I don't believe there is a problem selecting a file:

image
image

m417z:
@RussKie you're right, the bug doesn't happen with a space. It does happen when a path contains a comma, though.
For example, this doesn't work: `explorer.exe /select, C:\test,test`
This works (assuming the folder exists): `explorer.exe /select, "C:\test,test"`

PR #8517, opened 2020-10-04
Disable merge rebase commands when branch is local

### LINKED ISSUE #1248: Disable Merge and Rebase commands when current is local branch
Disable Merge and Rebase commands from Pull menu (or show error message before executing command) if current branch is local (not pushed to remote repository)

GintasS:
Fixes #1248

## Proposed changes

Like the author of the issue has said, I'm disabling these two buttons when the branch is local.

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/SubmodulesDialog/FormAddSubmodule.Designer.cs
+++ b/GitUI/CommandsDialogs/SubmodulesDialog/FormAddSubmodule.Designer.cs
@@ -146,7 +146,6 @@
             this.Controls.Add(this.Browse);
             this.Controls.Add(this.Directory);
             this.Controls.Add(this.label1);
-            this.FormBorderStyle = System.Windows.Forms.FormBorderStyle.SizableToolWindow;
             this.MaximizeBox = false;
             this.MinimizeBox = false;
             this.Name = "FormAddSubmodule";


RussKie (commented):
👍  
Please add tests

GintasS:
> 👍
> Please add tests

Will do :)

GintasS:
Please let me know if the this test is not sufficient, could not figure out how to make a scenario where a branch becomes a remote one, so it would set true to Enabled, Checked properties.

gerhardol:
If the default action is Merge or Rebase, the error message still appears, this should also be handled

The screenshot After does not have the circle checked, it should be set here.

GintasS:
Ok, try to see the cases, I removed the circle so it wouldn't confuse that an option is selected, but somehow is disabled.

gerhardol:
> Ok, try to see the cases, I removed the circle so it wouldn't confuse that an option is selected, but somehow is disabled.

For the disabled options, it is good to disable
I was asking about the remaining enabled option

GintasS:
> > Ok, try to see the cases, I removed the circle so it wouldn't confuse that an option is selected, but somehow is disabled.
> 
> For the disabled options, it is good to disable
> I was asking about the remaining enabled option

Just to be sure, you're talking about "Do not merge, only fetch remote changes"?

gerhardol:
> Just to be sure, you're talking about "Do not merge, only fetch remote changes"?

yes

GitUI/CommandsDialogs/FormPull.cs
```diff
@@ -1163,6 +1164,15 @@ private void PruneTags_CheckedChanged(object sender, EventArgs e)
             AllTags.Checked = AllTags.Checked || PruneTags.Checked;
         }
 
+        private void DisableMergeOptionsWhenCurrentBranchIsLocal()
+        {
+            var isLocalBranch = Module.GetRemoteBranch(localBranch.Text) == string.Empty;
```
RussKie, line 1169: Any reason not to write as follows?
```suggestion
            var isLocalBranch = string.IsNullOrWhiteSpace(Module.GetRemoteBranch(localBranch.Text));
```

RussKie (commented):
LGTM

GintasS:
> > Just to be sure, you're talking about "Do not merge, only fetch remote changes"?
> 
> yes

Fixed this.

GitUI/CommandsDialogs/FormPull.cs
GintasS: I was thinking that method call inside another method is not a good idea, also the introduction of IsNull part. Can GetRemoteBranch return null and how we should handle it? In my opinion, simple equality comparison with empty is simplier.

GintasS:
> **If the default action is Merge or Rebase, the error message still appears, this should also be handled**
> 
> The screenshot After does not have the circle checked, it should be set here.

I'm not sure how to do it properly, I know that you can set default action via:
AppSettings.DefaultPullAction

GitUI/CommandsDialogs/FormPull.cs
RussKie: > call inside another method is not a good idea

There are no issues with that. If you think it may be confusing (which I don't) you could introduce a temp variable, e.g.
```cs
string remoteBranch = Module.GetRemoteBranch(localBranch.Text);
bool isLocalBranch = string.IsNullOrWhiteSpace(remoteBranch);
```

> Can GetRemoteBranch return null and how we should handle it?

It doesn't. But even if it did - what is the concern here?

> In my opinion, simple equality comparison with empty is simplier.

Comparing strings like this is not in a spirit of C#.

GintasS: Sure, I will change it.

GitUI/CommandsDialogs/FormPull.cs
```diff
@@ -1163,6 +1164,17 @@ private void PruneTags_CheckedChanged(object sender, EventArgs e)
...
 
+        private void DisableMergeOptionsWhenCurrentBranchIsLocal()
+        {
+            var isLocalBranch = string.IsNullOrEmpty(Module.GetRemoteBranch(localBranch.Text));
+
+            Merge.Checked = !isLocalBranch;
+            Merge.Enabled = !isLocalBranch;
+            Rebase.Enabled = !isLocalBranch;
+            Fetch.Checked = !isLocalBranch;
+            Fetch.Enabled = !isLocalBranch;
```
RussKie, line 1175: [nit] we can simplify this a little bit by chaining assignments:
```suggestion
            Rebase.Enabled =
                Merge.Checked =
                Merge.Enabled =
                Fetch.Checked =
                Fetch.Enabled = !isLocalBranch;
```

RussKie:
> I know that you can set default action via:
> AppSettings.DefaultPullAction

We shouldn't touch this, this is a user setting.



> > **If the default action is Merge or Rebase, the error message still appears, this should also be handled**
> > The screenshot After does not have the circle checked, it should be set here.
> 
> I'm not sure how to do it properl

I presume what @gerhardol is referring to here is that by disabling the two merge options the 3rd is left unchecked. I'm not saying it has be checked until we understand the new behaviour with your fix.
Have you tested what happens if you click the "Pull" button?

GintasS:
> > I know that you can set default action via:
> > AppSettings.DefaultPullAction
> 
> We shouldn't touch this, this is a user setting.
> 
> > > **If the default action is Merge or Rebase, the error message still appears, this should also be handled**
> > > The screenshot After does not have the circle checked, it should be set here.
> > 
> > 
> > I'm not sure how to do it properl
> 
> I presume what @gerhardol is referring to here is that by disabling the two merge options the 3rd is left unchecked. I'm not saying it has be checked until we understand the new behaviour with your fix.
> Have you tested what happens if you click the "Pull" button?

So yes, I disabled the 3rd option with my new fix. Now regarding what happens when I press Pull button:

1.Default action is set to Pull - merge
image


2.Default action is set to Pull - rebase
image

GintasS:
Updated "After" image on the first comment.

RussKie:
@gerhardol what do you think of this? To me this feels acceptable.

GitUI/CommandsDialogs/FormPull.cs
```diff
@@ -1163,6 +1164,17 @@ private void PruneTags_CheckedChanged(object sender, EventArgs e)
             AllTags.Checked = AllTags.Checked || PruneTags.Checked;
         }
 
+        private void DisableMergeOptionsWhenCurrentBranchIsLocal()
```
gerhardol, line 1167:
```suggestion
        private void DisableMergeOptionsWhenCurrentBranchIsNotTracking()
```

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPullTests.cs
```diff
@@ -207,6 +207,27 @@ public void Should_remember_user_choice_upon_pull(bool autoStashChecked)
                 AppSettings.PullAction.Merge);
         }
 
+        [Test]
+        public void Should_disable_enable_merge_options_when_branch_local()
```
gerhardol, line 211:
```suggestion
        public void Should_disable_enable_merge_options_when_no_tracking()
```

GitUI/CommandsDialogs/FormPull.cs
gerhardol, line 1175: if (!isLocalBranch)
{
  "Do not merge".Checked = true;
}

gerhardol (commented):
A few naming suggestions (not commented all occurrences) and setting of checked
Otherwise fine

GitUI/CommandsDialogs/FormPull.cs
GintasS: What you mean by this?

GintasS:
Why does the build fail, integration tests are fine from my end?

gerhardol:
> Why does the build fail, integration tests are fine from my end?

They fail for me too, all on
`accessor.Fetch.Checked.Should().Be(fetchChecked);`
Some defaults maybe?

RussKie:
I'll have a look in a few days

RussKie:
image
Please done do that, rebase cleanly on HEAD instead.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPullTests.cs
```diff
@@ -207,6 +207,27 @@ public void Should_remember_user_choice_upon_pull(bool autoStashChecked)
...
 
+        [Test]
+        public void Should_disable_enable_merge_options_when_no_tracking()
+        {
+            RunFormTest(
+                form =>
+                {
+                    var accessor = form.GetTestAccessor();
+
+                    accessor.DisableMergeOptionsWhenCurrentBranchIsNotTracking();
```
RussKie, line 218: What's state of Merge, Rebase, Fetch controls before `DisableMergeOptionsWhenCurrentBranchIsNotTracking` is run? Is their state expected?

GintasS: Since we're using **_commands.StartPullDialog**, later **FormPullLoad** gets called and in it we call  **DisableMergeOptionsWhenCurrentBranchIsNotTracking()** , so yes, no need to call it again after the accessor initialization.

Thus, states of the controls are:
Rebase.Enabled = false
Merge.Enabled  = false
Fetch:.Enabled   = true
Fetch.Checked = true

(Also fixed !isLocalBranch in if statement because Fetch.Checked was set to true when the branch is **not** the local one, if it's available in remote, we should set Merge to true, not Fetch)

RussKie: > Since we're using **_commands.StartPullDialog**, later **FormPullLoad** gets called and in it we call **DisableMergeOptionsWhenCurrentBranchIsNotTracking()** , so yes, no need to call it again after the accessor initialization.
> 
> Thus, states of the controls are:
> Rebase.Enabled = false
> Merge.Enabled = false
> Fetch:.Enabled = true
> Fetch.Checked = true

Please add this as a comment before line:216, so we don't have to scratch our heads next time we look at the test.
Thank you

GintasS: Fixed, thank you for your comments 👍

RussKie:
There's something wrong, you have too many changes, and most aren't yours.
Please cleanly rebase **only your** changes on top of the master branch, and squash. https://github.com/gitextensions/gitextensions/wiki/How-To%3A-Squash-and-Rebase-your-changes

GintasS:
> There's something wrong, you have too many changes, and most aren't yours.
> Please cleanly rebase **only your** changes on top of the master branch, and squash. https://github.com/gitextensions/gitextensions/wiki/How-To%3A-Squash-and-Rebase-your-changes

I removed the changes, but the rebase thing is till valid.

Issue #7726, opened 2020-02-04
Latest version not running under new install of windows 10


MarkusRissmann:
I have a brand new install of windows 10 in VMWare workstations.

I am installing "GitExtensions-3.3.1.7897.msi".

After installing I launch GitExtension from the desktop shortcut. Nothing appears to happen. If I view in Task Manager I see "Git Extentions" being launched 20-30+ times and growing.

I had to disable windows defender to get it to run in the first place else I would get a message saying git exe not found.

RussKie:
Hmm that's nasty.
If git isn't present the app should show a popup saying just that.

MarkusRissmann:
ahhh, so that is what it was telling me. The message wasn't clear at all. This is a new environment so I was trying to install minimal apps. I guess to minimal.

RussKie:
How would you reword the message to be clearer?

MarkusRissmann:
something like 

"Please make sure that the GIT library has been installed" 

or more verbose  

"GIT is a prerequisite of GitExtentions. Please make sure it is installed". 

My confusion was that I was already having issues installing GitExtensions with windows defender so I thought the error had to do with the current installation not a missing resource,

RussKie:
Is it not what you saw?
![image](https://user-images.githubusercontent.com/4403806/73831061-d0c3bf00-4859-11ea-84b7-1a976a492cdc.png)

MarkusRissmann:
Yes, but as I said, I was confused and thought something was wrong with the current installation of GitExtensions because of the previous warnings and blocking from windows defender. It didn't dawn on me (but is very apparent now) that this error was referring to a prerequisite.

RussKie:
@gitextensions/git-extensions-team thoughts on the proposed wording?

NikolayXHD:
The Git executable could not be located on your system, Git is a prerequisite to GitExtensions.

MarkusRissmann:
Sounds good! Thx

JonathanLindsey:
In my opinion the words 'located' and 'find' as used in the current dialog are confusing. I think it should be worded something like this:

> Git Extensions requires a Git executable to be on your computer which does not appear to be the case. Please install Git or click Browse to find Git on your computer manually.
> 
> **Install Git**
> _Go to git-scm.com_
> 
> **Browse to Git**
> _Find Git on your computer_
> 
> More info

The **bolded lines** should be large text and the _italicized lines_ should be smaller text on the same button.

[...]

PR #8519, opened 2020-10-04
Fix 7726 - "Latest version not running under new install of windows 10"

hoborm:
Fixes #
Rewritten UI text with the wording discussed in the Issue comment section.

## Proposed changes

- New Find Git text
- New Install Git text
- New alert text
- Simple replacement in translation files (*.xlf)

### Before

[In this reply](https://github.com/gitextensions/gitextensions/issues/7726#issuecomment-582328310)

### After

image

## Test methodology 

- Ran the UI tests
- No new tests though

## Test environment(s) 

- GIT git version 2.28.0.windows.1
- Windows Version	10.0.19041 Build 19041

I could not change the wording in the translated languages, all of the changes should be translated again.

----

### DIFF
--- a/ResourceManager/Strings.cs
+++ b/ResourceManager/Strings.cs
@@ -27,10 +27,9 @@ You can change your mind at any time.
 
 Yes, I allow telemetry!");
 
-        private readonly TranslationString _installGitInstructions = new TranslationString("Install git...");
-        private readonly TranslationString _findGitExecutable = new TranslationString("Find git...");
-        private readonly TranslationString _gitExecutableNotFoundText =
-            new TranslationString("The Git executable could not be located on your system.");
+        private readonly TranslationString _installGitInstructions = new TranslationString("Install Git\r\nGo to GitExtensions wiki");
+        private readonly TranslationString _findGitExecutable = new TranslationString("Browse to Git\r\n Find Git on your computer");
+        private readonly TranslationString _gitExecutableNotFoundText = new TranslationString("Git Extensions requires a Git executable to be on your computer which does not appear to be the case. Please install Git or click Browse to find Git on your computer manually.");
         private readonly TranslationString _authorDateText = new TranslationString("{0:Author date|Author dates}");
         private readonly TranslationString _committerText = new TranslationString("Committer");
         private readonly TranslationString _commitDateText = new TranslationString("{0:Commit date|Commit dates}");
--- a/GitUI/Translation/Czech.xlf
+++ b/GitUI/Translation/Czech.xlf
@@ -11017,7 +11017,8 @@ Select this commit to populate the full message.</source>
         
       </trans-unit>
       <trans-unit id="_findGitExecutable.Text">
-        <source>Find git...</source>
+        <source>Browse to Git
+Find Git on your computer</source>
         <target>Najít git...</target>
         
       </trans-unit>
@@ -11034,7 +11035,7 @@ Select this commit to populate the full message.</source>
         
       </trans-unit>
       <trans-unit id="_gitExecutableNotFoundText.Text">
-        <source>The Git executable could not be located on your system.</source>
+        <source>Git Extensions requires a Git executable to be on your computer which does not appear to be the case. Please install Git or click Browse to find Git on your computer manually.</source>
         
       </trans-unit>
       <trans-unit id="_hoursAgo.Text">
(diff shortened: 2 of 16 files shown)


gerhardol (commented):
Text is OK

However, only Strings.cs need manual changes, then English.xlf is updated by the build script:
https://github.com/gitextensions/gitextensions/wiki/Translations

The translations are updated from Transifex (every  changed text will require the translators to update again.)
You can squash the changes directly for something like this, so a squash is not required before the merge.

ResourceManager/Strings.cs
```diff
@@ -27,10 +27,9 @@ public sealed class Strings : Translate
 
 Yes, I allow telemetry!");
 
-        private readonly TranslationString _installGitInstructions = new TranslationString("Install git...");
-        private readonly TranslationString _findGitExecutable = new TranslationString("Find git...");
-        private readonly TranslationString _gitExecutableNotFoundText =
-            new TranslationString("The Git executable could not be located on your system.");
+        private readonly TranslationString _installGitInstructions = new TranslationString("Install Git\r\nGo to GitExtensions wiki");
```
RussKie, line 30: `TaskDialogCommandLink` has an overload that takes 3 parameters: `TaskDialogCommandLink(string name, string text, string instruction)`.

```suggestion
        private readonly TranslationString _installGitInstructions = new TranslationString("Install git...");
        private readonly TranslationString _installGitInstructionsDetails = new TranslationString("Click here to read instructions");
```

ResourceManager/Strings.cs
```diff
@@ -27,10 +27,9 @@ public sealed class Strings : Translate
...
+        private readonly TranslationString _findGitExecutable = new TranslationString("Browse to Git\r\n Find Git on your computer");
```
RussKie, line 31:
```suggestion
        private readonly TranslationString _findGitExecutable = new TranslationString("Find git...");
        private readonly TranslationString _findGitExecutableDetails = new TranslationString("Click here to find git on your computer");
```

ResourceManager/Strings.cs
```diff
@@ -27,10 +27,9 @@ public sealed class Strings : Translate
...
+        private readonly TranslationString _gitExecutableNotFoundText = new TranslationString("Git Extensions requires a Git executable to be on your computer which does not appear to be the case. Please install Git or click Browse to find Git on your computer manually.");
```
RussKie, line 32: These many words don't make it more the dialog user friendly. In fact I feel it is the opposite.

```
The Git executable could not be located on your system.
```
vs
```
Git Extensions requires a Git executable to be on your computer which does not appear to be the case. Please install Git or click Browse to find Git on your computer manually.
```


What's in the original message isn't sufficiently clear in your opinion?

RussKie (changes requested):

ResourceManager/Strings.cs
hoborm: I just created the modifications based on the discussion over the issue #7726  itself. I don't have any strong opinion on either of these, I just tried to contribute to the application.

RussKie: Let's keep the short version, thank you.

mstv: Why have you changed your mind, @RussKie? https://github.com/gitextensions/gitextensions/issues/7726#issuecomment-586212844
I think the first sentence will help users to understand the dependency better.
I agree, the second sentence is kind of redundant with the TaskDialog buttons.

RussKie: > Why have you changed your mind, @RussKie? [#7726 (comment)](https://github.com/gitextensions/gitextensions/issues/7726#issuecomment-586212844)

Because of how the dialog will look - putting myself in a shoes of an average user to me it looks scary with so much text:
image

With UI quite often it is hard to know what a result would be, and with that a number of iterations may be required... And according to Jeff Bezos [The Smart People Change Their Minds](https://techcrunch.com/2012/10/19/jeff-bezos-the-smart-people-change-their-minds/) 😄 

Also to me it feels the message has a lot of duplicating information, which can be removed without affecting the comprehension.

I'm happy to iterate over the wording, if you have ideas.

mstv:
```suggestion
        private readonly TranslationString _gitExecutableNotFoundText = new TranslationString("Git Extensions requires a Git executable but could not locate it.");
```
?

RussKie: 👍

GintasS:

RussKie:
We should probably just reword it..

PR #8522, opened 2020-10-05
Migrate to .NET 5.0

RussKie:
## Proposed changes

- Convert to .NET 5.0

## Done:

* [x] Can build and run (almost all) tests using `dotnet build` and `dotnet test`
* [x] Can run the app
* [x] Commit info RTF links rendering
* [x] basic CI build via GitHub action

## Outstanding:

* [ ] Theming / colours - blocked / depends on https://github.com/EasyHook/EasyHook/issues/278
* [ ] Plugins load - https://github.com/dotnet/runtime/issues/46066#issuecomment-760526283
* [x] Local dev experience: `dotnet restore`, `dotnet build`, `dotnet test`, `dotnet run`
* [ ] Packaging
* [ ] Publishing / signing
* [x] Several tests deadlock -> #9158

Follow up:
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621327401 -> #9150
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621338341 -> #9151
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621346302 -> #9152 #9153 #9154 #9155 #9156 #9157
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621360978 and https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621368145 -> #9158
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621392829 -> #9159
* [x] https://github.com/gitextensions/gitextensions/pull/8522#discussion_r621389179 -> #9160

----

Adjust code with the newer code analyzers rules.


RussKie:
@gitextensions/git-extensions-team you know what, I think the big regression in the editorcontrol was solved in 5.0 RTM, so it looks like we have a take off...

If you're curious - clone with submodule into a clean folder (or git clean -xdf your current working directory), `.\start-code.cmd`, `.\build -launch` from the terminal in VS Code.

The theming is all busted, subject to discussions in https://github.com/gitextensions/gitextensions/pull/8424#discussion_r500004536. BUT IT RUNS!

ivangrek:
Application.EnableVisualStyles();
Application.SetCompatibleTextRenderingDefault(false);

to

**Application.SetHighDpiMode(HighDpiMode.SystemAware);**
Application.EnableVisualStyles();
Application.SetCompatibleTextRenderingDefault(false);

ivangrek:
What are you planning to do with CodeFactor errors?
I fixed one in #8623.

RussKie:
Ah thanks! ;)

Re: CodeFactor - in this work I'm not paying a lot of attention to it.
It is generally useful for the new work, to alert if we're adding tech debt.

UnitTests/GitUI.Tests/UserControls/CommitInfo/BranchComparerTests.cs
```diff
@@ -29,8 +29,8 @@ public void BranchComparer([Values(null, "current")] string currentBranch)
 
                 // local branches
                 "1234_issue",
-                "current/2",
                 "current_2",
+                "current/2",
```
RussKie, line 33: See: https://github.com/dotnet/runtime/discussions/46265

RussKie:
@NikolayXHD if you have some spare cycles do you think you could give me a hand with (re-)enabling colours at least in the default light theme? This way we'll get the app usable.

weltkante:
> Commit info RTF links rendering

I don't know what the status (and requirements) of the RTF link handling is here, but I did some investigations as part of dotnet/winforms#3632 and a porting attempt of NLog.Windows.Forms which had similar functionality, feel free to ping me if there are further questions. If you need to react to `LinkClicked` events you may be blocked by dotnet/winforms#4418

ivangrek:
I think we need .NET 6 as a priority. If it is possible to move partially to .NET 5, let's do it!

RussKie:
> I think we need .NET 6 as a priority.

.NET 6 will released in 6 months.

ivangrek:
> .NET 6 will released in 6 months.

And this is the main point! See how much time we spent on this problem. 😃

RussKie:
Since Sep 2019 😣

Few tests are failing, but it's building and runs.
There are issues to sort, but it's almost at the "ready to merge" state.

RussKie:
@gitextensions/git-extensions-team: this is ready review. Few tests are failing, which are easy to fix.
Outstanding items, such as plugins loading, packaging, etc. will be dealt with separately.

@sharwell may I please get your to review https://github.com/gitextensions/gitextensions/pull/8522/commits/2971414d185bd86de8b16db85ad3f051146cfb55 and https://github.com/gitextensions/gitextensions/pull/8522/commits/53db91514b860795b462e378632b623b82d793b1
Also [handing tests in `ThreadHelperTests`](https://github.com/gitextensions/gitextensions/pull/8522/commits/2f8e84cf424b8e5c305ed29b0b8e3ee7e3b81ed6#diff-c12e3716357e7ef021e61eaa0cde3bdd3c62790557e6b1436b11cc48e18be480) in https://github.com/gitextensions/gitextensions/pull/8522/commits/2f8e84cf424b8e5c305ed29b0b8e3ee7e3b81ed6

@drewnoakes is you can please skim over the NRT fixes.

GitCommands/GitCommands.csproj
```diff
@@ -1,19 +1,21 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
     <UseWindowsForms>true</UseWindowsForms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
```
drewnoakes, line 7: Is this needed? I thought this was enabled at a higher level via `Directory.Build.props`.

GitUI/GitUI.csproj
```diff
@@ -3,15 +3,14 @@
   <PropertyGroup>
     <OutputType>Library</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
```
drewnoakes: Ditto.

Plugins/GitUIPluginInterfaces/GitUIPluginInterfaces.csproj
```diff
@@ -4,13 +4,12 @@
     <OutputType>Library</OutputType>
     <UseWindowsForms>true</UseWindowsForms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
```
drewnoakes, line 7: Ditto.

GitCommands/PathEqualityComparer.cs
```diff
@@ -7,10 +7,18 @@ namespace GitCommands
...
     public class PathEqualityComparer : IEqualityComparer<string>
     {
-        public bool Equals(string path1, string path2)
+        public bool Equals(string? path1, string? path2)
         {
-            path1 = Path.GetFullPath(path1).TrimEnd('\\');
-            path2 = Path.GetFullPath(path2).TrimEnd('\\');
+            if (path1 is not null)
+            {
+                path1 = Path.GetFullPath(path1!).TrimEnd('\\');
```
drewnoakes, line 14: Is the `!` really needed here (and below)? That seems like a Roslyn bug if so.

GitCommands/UserRepositoryHistory/RecentRepoInfo.cs
```diff
@@ -181,13 +182,15 @@ private static void AddToOrderedSignDir(SortedList<string, List<RecentRepoInfo>>
...
 
-            var existsShortName = orderedRepos.TryGetValue(repoInfo.Caption!, out var list);
+            bool existsShortName = orderedRepos.TryGetValue(repoInfo.Caption!, out List<RecentRepoInfo>? list);
             if (!existsShortName)
             {
                 list = new List<RecentRepoInfo>();
                 orderedRepos.Add(repoInfo.Caption!, list);
             }
 
+            Assumes.NotNull(list);
```
drewnoakes, line 192: `Validates.NotNull`.

ResourceManager/GitExtensionsControlInitialiser.cs
```diff
@@ -27,11 +27,11 @@ public GitExtensionsControlInitialiser(GitExtensionsControl control)
             _translate = control;
         }
 
-        private void LoadHandler(object control, EventArgs e)
+        private void LoadHandler(object? sender, EventArgs e)
         {
             if (!_initialiseCompleteCalled && !IsDesignModeActive)
             {
-                throw new Exception($"{control.GetType().Name} must call {nameof(InitializeComplete)} in its constructor, ideally as the final statement.");
+                throw new Exception($"{((Control)sender!).GetType().Name} must call {nameof(InitializeComplete)} in its constructor, ideally as the final statement.");
```
drewnoakes, line 34: This looks wrong. The method advertises it accepts null but then throws if it receives it. Either change the signature and suppress usages of the method, or handle null more gracefully here.

ResourceManager/Xliff/TranslationUtil.cs
```diff
@@ -276,7 +281,7 @@ private static bool IsTranslatableItemInComponent(PropertyInfo property)
         private static bool IsTranslatable(this Assembly assembly)
         {
             bool isInvalid = UnTranslatableDLLs.Any(
-                asm => assembly.FullName.StartsWith(asm, StringComparison.OrdinalIgnoreCase));
+                asm => assembly.FullName!.StartsWith(asm, StringComparison.OrdinalIgnoreCase));
```
drewnoakes, line 284:
```c#
assembly.FullName?.StartsWith(asm, StringComparison.OrdinalIgnoreCase) == true
```

Plugins/Bitbucket/GetPullRequest.cs
```diff
@@ -110,7 +112,7 @@ public GetPullRequest(string projectKey, string repoName, Settings settings)
         protected override List<PullRequest> ParseResponse(JObject json)
         {
             var result = new List<PullRequest>();
-            foreach (JObject val in json["values"])
+            foreach (JObject val in json["values"]!)
```
drewnoakes, line 115:
```c#
foreach (JObject val in json["values"] ?? Enumerable.Empty<JObject>())
```

Plugins/Bitbucket/GetRelatedRepoRequest.cs
```diff
@@ -41,7 +43,7 @@ public GetRelatedRepoRequest(Settings settings) : base(settings)
         protected override List<Repository> ParseResponse(JObject json)
         {
             var result = new List<Repository>();
-            foreach (JObject val in json["values"])
+            foreach (JObject val in json["values"]!)
```
drewnoakes, line 46: Ditto.

Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
```diff
@@ -185,7 +187,8 @@ private IEnumerable<AppVeyorBuildInfo> QueryBuildsResults(string projectId)
             }
         }
 
-        internal IEnumerable<AppVeyorBuildInfo> ExtractBuildInfo(string projectId, string? result)
+#nullable disable
```
drewnoakes, line 190: Better to use `#nullable disable warnings` here (and elsewhere) so that you can keep annotations in code (eg. line 263).

Plugins/BuildServerIntegration/TeamCityIntegration/TeamCityAdapter.cs
```diff
@@ -568,7 +575,8 @@ public void Dispose()
             _httpClient?.Dispose();
         }
 
-        public Project? GetProjectsTree()
+ #nullable disable
+        public Project GetProjectsTree()
```
drewnoakes, line 579: This can return null. Keep the annotation? Use `#nullable disable warnings` instead?

GitExtUtils/MruCache.cs
```diff
@@ -62,6 +65,8 @@ public void Add(TKey key, TValue value)
                 if (_entries.Count == Capacity)
                 {
                     var last = _entries.Last;
+                    Assumes.NotNull(last);
```
drewnoakes, line 68: `Validates.NotNull`. I suggest searching for usages of `Assumes.NotNull` and replacing them.

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
```diff
@@ -98,10 +98,9 @@ private static IEnumerable<string> ReadOrInitializeAutoCompleteRegexes()
             Validates.NotNull(appDataPath);
 
             var path = PathUtil.Combine(appDataPath, "AutoCompleteRegexes.txt");
-
             if (File.Exists(path))
             {
-                return File.ReadLines(path);
+                return File.ReadLines(path!);
```
drewnoakes, line 103: I'm surprised this `!` is needed. I'd think `File.Exists` would be annotated such that a true return would imply `path` is non-null.

GitExtensions/Program.cs
```diff
@@ -247,10 +247,10 @@ private static void HandleConfigurationException(ConfigurationException ce)
             try
             {
                 // perhaps this should be checked for if it is null
-                var in3 = ce.InnerException.InnerException;
+                Exception? in3 = ce.InnerException?.InnerException;
```
drewnoakes, line 250: Is the comment redundant now?

drewnoakes:
This is a Hurculean effort @RussKie! 💪 

In general I'd like to avoid using `!`. Most cases can be replaced with `Validates.NotNull` to surface bugs more easily. There are valid scenarios, but they're uncommon.

I understand there are a lot of these cases and it's easier to just suppress. Unfortunately it's hard to search for these suppressions to fix them later as `!` is used for negation as well.

.editorconfig
```diff
@@ -91,10 +91,10 @@ dotnet_code_quality_unused_parameters = all:suggestion
 
 #### C# Coding Conventions ####
 
-# var preferences
-csharp_style_var_elsewhere = false:silent
-csharp_style_var_for_built_in_types = false:silent
-csharp_style_var_when_type_is_apparent = true:silent
+# Avoid var
```
sharwell, line 94: 😕 Why?

BugReporter/BugReportForm.cs
```diff
@@ -208,7 +209,7 @@ protected void TranslateItem(string itemName, object item)
                 return;
             }
 
-            var itemsToTranslate = new[] { (itemName, item) };
+            IEnumerable<(string name, object? item)> itemsToTranslate = new (string name, object? item)[] { (itemName, item) };
```
sharwell, line 212: It looks like the old form was better here

BugReporter/Info/Report.cs
```diff
@@ -40,7 +40,7 @@ public override string ToString()
             serializer.Serialize(stream, this);
             stream.Position = 0;
             var doc = XDocument.Load(stream);
-            return doc.Root.ToString();
+            return doc.Root!.ToString();
```
sharwell, line 43: 💡 Validate instead of suppress

BugReporter/Serialization/SerializableDictionary.cs
```diff
@@ -69,7 +69,7 @@ public void WriteXml(XmlWriter writer)
         {
             foreach (var key in Keys)
             {
-                writer.WriteStartElement(key.ToString().Replace(" ", ""));
+                writer.WriteStartElement(key.ToString()!.Replace(" ", ""));
```
sharwell, line 72: 📝 This appears to be a valid suppression due to a ["design choice"](https://github.com/dotnet/runtime/issues/1318) for `object.ToString()`

BugReporter/Serialization/SerializableException.cs
```diff
@@ -170,14 +171,14 @@ public string ToXmlString()
             serializer.Serialize(stream, this);
             stream.Position = 0;
             var doc = XDocument.Load(stream);
-            return doc.Root.ToString();
+            return doc.Root!.ToString();
```
sharwell, line 174: 💡 validate

BugReporter/Serialization/SerializableException.cs
```diff
@@ -170,14 +171,14 @@ public string ToXmlString()
...
         }
 
         public static SerializableException FromXmlString(string xml)
         {
             var serializer = new XmlSerializer(typeof(SerializableException));
             using StringReader reader = new(xml);
-            SerializableException exception = (SerializableException)serializer.Deserialize(reader);
+            SerializableException exception = (SerializableException)serializer.Deserialize(reader)!;
```
sharwell, line 181: 💡 validate

BugReporter/Serialization/SerializableException.cs
```diff
@@ -201,9 +202,9 @@ public static SerializableException FromXmlString(string xml)
             {
                 var extendedInformation = new SerializableDictionary<string, object>();
 
-                foreach (var property in extendedProperties.Where(property => property.GetValue(exception, null) is not null))
+                foreach (PropertyInfo property in extendedProperties.Where(property => property.GetValue(exception, null) is not null))
```
sharwell, line 205: 💡 Previous code was better

sharwell (dismissed):
Stopping my review pending revert of whichever commit changed the `var` preferences.

I recommend moving the entire Nullable Reference Types change to a separate pull request, as it's not necessary for the .NET 5.0 migration and is the source of the overwhelming majority of the PR complexity.

It might be easier to just create a new PR that cherry picks the changes required for .NET 5.0, and then update the current branch afterward to apply the style changes as a follow-up.

drewnoakes:
I agree with @sharwell that it'd be convenient to pull out the NRT changes, to make reviewing this easier. But that won't be straightforward.

@RussKie one idea that might help is to make another PR that replaces `object sender, ` with `object? sender, ` across the codebase, and get that merged first.

BugReporter/BugReportForm.cs
drewnoakes: I suspect the intention was this:

```suggestion
            var itemsToTranslate = new[] { (itemName, (object?)item) };
```

RussKie:
A lot NRTA are introduced in .NET 5.0, so those can't be applied prior to the migration.

The "var" rule matches the rule in dotnet/runtime and dotnet/winforms repos.
A number of changes to "var" are manual to increase the comprehension of the code.

drewnoakes:
> A lot NRTA are introduced in .NET 5.0, so those can't be applied prior to the migration.

They can. Prior to .NET 5, the APIs we were compiling against weren't annotated (oblivious types). You will find that the annotations you added work fine even on .NET Framework.

RussKie:
I removed all additional NRT-related changes except for removing NRT workarounds

.editorconfig
RussKie: Because it makes it harder to reason about types. Especially so when the code is reviewed outside VS.

ivangrek (approved):
👍

GitCommands/CommitTemplateItem.cs
```diff
@@ -90,8 +90,10 @@ private static string SerializeCommitTemplates(CommitTemplateItem[]? items)
 
                     byte[] memoryData = Convert.FromBase64String(serializedString.Substring(p + 1));
                     using var rs = new MemoryStream(memoryData, 0, length);
+#pragma warning disable SYSLIB0011 // Type or member is obsolete
```
sharwell, line 93: 📝 We should consider filing issues to migrate these to another serializer

GitExtUtils/ArrayExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static T[] Subsequence<T>(this T[] array, int index, int length)
         }
 
         [MustUseReturnValue]
-        public static T[] Append<T>(this T[] array, T element)
+        public static T[] AppendTo<T>(this T[] array, T element)
```
sharwell, line 17: I'm guessing there is a new extension method?

GitExtensions/Project.Build.targets
```diff
@@ -8,12 +8,12 @@
     ============================================================
     -->
   <Target Name="_CopyTranslations" AfterTargets="AfterBuild">
-    <_GetAvailableTranslationsCodeTask
+    <!-- <_GetAvailableTranslationsCodeTask
```
sharwell, line 11: 💡 Can you link commented-out code to specific issues in cases where behavior changed?

GitUI/CommandsDialogs/FormEditor.cs
```diff
@@ -6,6 +6,7 @@
 using GitExtUtils.GitUI.Theming;
 using GitUI.Editor;
 using GitUI.Theming;
+using JetBrains.Annotations;
```
sharwell, line 9: ❔ Was this one needed?

GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -15,6 +15,8 @@
 using Microsoft;
 using Microsoft.WindowsAPICodePack.Dialogs;
 using ResourceManager;
+using TaskDialog = Microsoft.WindowsAPICodePack.Dialogs.TaskDialog;
+using TaskDialogStartupLocation = Microsoft.WindowsAPICodePack.Dialogs.TaskDialogStartupLocation;
```
sharwell, line 19: 💡 There are some other files that would benefit from this approach. And I was literally just mentioning this to a team member in context of https://github.com/dotnet/roslyn/issues/38095. 😄

GitUI/CommandsDialogs/FormSparseWorkingCopyViewModel.cs
```diff
@@ -162,7 +162,7 @@ public void SaveChanges()
         {
             // Don't abort if !IsWithUnsavedChanges because we have to run IsRefreshWorkingCopyOnSave in either case (e.g. if edited by hand or got outdated)
 
-            // Special case: turning off sparse for a repo � this won't just go smoothly, looks like git still reads the sparse checkout rules, so emptying or deleting them with turning off will just leave you with what you had before
+            // Special case: turning off sparse for a repo � this won't just go smoothly, looks like git still reads the sparse checkout rules, so emptying or deleting them with turning off will just leave you with what you had before
```
sharwell, line 165: ⚠️ An encoding error occurred and needs to be manually fixed. Don't worry about changing .editorconfig or looking for cases outside this PR (I have a script that can find/fix those separately).

GitUI/CommandsDialogs/FormSparseWorkingCopyViewModel.cs
```diff
@@ -231,7 +231,7 @@ private void SaveChangesTurningOffSparseSpecialCase()
             }
 
             // Adjust the rules
-            // Comment out all existing nonempty lines, add the single �/*� line to make a total pass filter
+            // Comment out all existing nonempty lines, add the single �/*� line to make a total pass filter
```
sharwell, line 234: ⚠️ An encoding error occurred, but in this case I recommend changing the comment to use `"` instead of curly-quotes.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -138,6 +138,10 @@ public CommitInfo()
                     .Subscribe(_ => handler(_.EventArgs));
 
             commitInfoHeader.SetContextMenuStrip(commitInfoContextMenuStrip);
+
+            // at this point rtbxCommitMessage.Bounds = {X = 8 Y = 8 Width = 440 Height = 0}
+            // and with Height=0 we won't be receiving any ContentsResizedEvents
+            rtbxCommitMessage.Height = 1;
```
sharwell, line 144: 💭 Just to make sure, this was intentional?

GitUI/GitUIExtensions.cs
```diff
@@ -117,7 +117,11 @@ void OpenWithDiffTool()
                     && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
                 {
                     // Patch already evaluated
+#pragma warning disable VSTHRD103 // Call async methods when in an async method
+#pragma warning disable VSTHRD003 // Avoid awaiting foreign Tasks
```
sharwell, line 121: 💡 Recommend filing follow-up issues for each of the VSTHRD suppressions (i.e. one issue covering all VSTHRD003, one issue covering all VSTHRD103, etc.).

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -462,7 +462,7 @@ private async Task RunBackgroundAsync(CancellationToken cancellationToken)
 
                 try
                 {
-                    CancellationToken timeoutToken;
+                    CancellationToken timeoutToken = default;
```
sharwell, line 465:
```suggestion
                    CancellationToken timeoutToken = CancellationToken.None;
```

sharwell (commented):
Completed 150/240 files

UnitTests/GitUI.Tests/ThreadHelperTests.cs
```diff
@@ -27,6 +27,7 @@ private static async Task ThrowExceptionAsync(Exception ex)
         }
 
         [Test]
+        [Ignore("Hangs")]
```
sharwell, line 30: 💡 You can create an issue and assign it to me for this one

UnitTests/ResourceManager.Tests/CommitDataRenders/CommitDataHeaderRendererTests.cs
```diff
@@ -16,6 +16,7 @@ namespace ResourceManagerTests.CommitDataRenders
     [TestFixture]
     public class CommitDataHeaderRendererTests
     {
+        private const int MultipleCommits = 2;
```
sharwell, line 19: 💭 Is this value used?

Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs
```diff
@@ -98,13 +98,7 @@ private static ExportProvider CreateExportProvider(string? applicationDataFolder
...
             try
             {
-                var assemblyName = AssemblyName.GetAssemblyName(file.FullName);
-                if (assemblyName is null)
-                {
-                    return null;
-                }
-
-                return Assembly.Load(assemblyName);
+                return Assembly.LoadFile(file.FullName);
```
sharwell, line 101: 💡 `LoadFrom` would be a closer match to the original behavior here

ResourceManager/LinkFactory.cs
```diff
@@ -48,7 +48,7 @@ public string CreateLink(string? caption, string uri)
         private string AddLink(string? caption, string uri)
         {
             string htmlUri = WebUtility.HtmlEncode(uri);
-            string rtfLinkText = caption + "#" + htmlUri;
+            string rtfLinkText = caption;
```
sharwell, line 51: 💭 Oh interesting. Is this going to change any observable behavior?

UnitTests/GitUI.Tests/ControlThreadingExtensionsTests.cs
```diff
@@ -170,6 +170,7 @@ public async Task TokenCancelledBeforeSwitchOnBackgroundThread()
         }
 
         [Test]
+        [Ignore("Hangs")]
```
sharwell, line 173: 💡 Can link to same issue mentioned previously

GitCommands/EnvironmentPathsProvider.cs
```diff
@@ -73,22 +73,32 @@ public IEnumerable<string> GetEnvironmentValidPaths()
...
+
+                return true;
             }
             catch (ArgumentException)
             {
             }
-            catch (PathTooLongException)
+            catch (IOException)
             {
+                // Querying attribures for UNC paths results in IOException
```
sharwell, line 91: 💭 It seems like we could use a `path.StartsWith` check to avoid relying on exceptions for flow control, but I'm not sure it matters.

GitCommands/EnvironmentPathsProvider.cs
```diff
@@ -73,22 +73,32 @@ public IEnumerable<string> GetEnvironmentValidPaths()
...
         {
-            FileInfo? fi = null;
             try
             {
-                fi = new FileInfo(path);
+                // NOTE:
+                // Path APIs don't throw an exception for invalid characters
+                // https://docs.microsoft.com/dotnet/core/compatibility/2.1#path-apis-dont-throw-an-exception-for-invalid-characters
+
+                _ = new FileInfo(path).Attributes;
```
sharwell, line 82: ❔ Instead of this, would it be better to manually check `path` against `Path.GetInvalidPathChars`?

GitExtensions/Program.cs
```diff
@@ -44,16 +43,16 @@ private static void Main()
 
             AppSettings.SetDocumentationBaseUrl(ThisAssembly.Git.Branch);
 
-            ThemeModule.Load();
-            Application.ApplicationExit += (s, e) => ThemeModule.Unload();
+            ////ThemeModule.Load();
+            ////Application.ApplicationExit += (s, e) => ThemeModule.Unload();
```
sharwell, line 47: 💭 I worry that it will be particularly difficult to track down the commented theming code. I almost wish it was wrapped in `#if SUPPORT_THEMES` instead of just commented.

GitUI/CommandsDialogs/FormCleanupRepository.cs
```diff
@@ -110,21 +110,16 @@ private void checkBoxPathFilter_CheckedChanged(object sender, EventArgs e)
 
         private void AddPath_Click(object sender, EventArgs e)
         {
-            var dialog = new CommonOpenFileDialog
+            var dialog = new FolderBrowserDialog()
```
sharwell, line 113: ❔ Are we losing any functionality by this change?

GitUI/CommandsDialogs/FormCleanupRepository.cs
```diff
@@ -110,21 +110,16 @@ private void checkBoxPathFilter_CheckedChanged(object sender, EventArgs e)
...
+                SelectedPath = Module.WorkingDir,
             };
 
-            var result = dialog.ShowDialog();
-            if (result == CommonFileDialogResult.Ok)
+            var result = dialog.ShowDialog(this);
+            if (result == DialogResult.OK)
             {
-                var subFoldersToClean = dialog.FileNames
+                // TODO: fix
```
sharwell, line 121: 💡 Link to issue

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -27,29 +27,31 @@ public static void Initialize()
...
+                ////            GitHosters.Add(repositoryHostPlugin);
+                ////        }
+
+                ////        Plugins.Add(plugin);
+                ////    }
+                ////}
+                ////catch
+                ////{
+                ////    // no-op
+                ////}
```
sharwell, line 54: ❔ Does this disable all extensions functionality, or just a subset? Can we link it to a follow-up issue?

GitUI/Theming/LocalHook.cs
```diff
@@ -0,0 +1,497 @@
+﻿#pragma warning disable SA1121 // Use built-in type alias
+#pragma warning disable SA1131 // Use readable conditions
+#pragma warning disable SA1308 // Variable names should not be prefixed
+#pragma warning disable SA1312 // Variable names should begin with lower-case letter
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+#pragma warning disable IDE1006 // Naming Styles
+#pragma warning disable SX1309 // Field names should begin with underscore
```
sharwell, line 9: 💭 A this point you might as well use this:

```
// <auto-generated>Sourced externally; not subject to local code style</auto-generated>
```

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
```diff
@@ -92,7 +93,7 @@ public int EnsureScoreIsAbove(int minimalScore)
 
         public ImmutableStack<RevisionGraphRevision> Parents => _parents;
         public ImmutableStack<RevisionGraphRevision> Children => _children;
-        public SynchronizedCollection<RevisionGraphSegment> StartSegments { get; }
+        public ConcurrentBag<RevisionGraphSegment> StartSegments { get; }
```
sharwell, line 96: 💭 I believe `ConcurrentBag` is unordered. Is that going to cause a problem here?

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormInitTests.cs
```diff
@@ -65,7 +65,6 @@ public void Should_show_current_GitModuleWorkingDir_if_supplied_path_null()
         [TestCase("")]
         [TestCase(null)]
         [TestCase("    ")]
-        [TestCase(@"c:\foo*\bar?\")]
```
sharwell, line 68: 💡 If you switch to using `GetInvalidPathChars` (per a previous comment), it may make sense to restore several of the test inputs that are removed.

UnitTests/GitCommands.Tests/EnvironmentPathsProviderTests.cs
```diff
@@ -30,7 +30,7 @@ public void Setup()
         public void GetEnvironmentValidPaths()
         {
             string pathVariable = string.Join(_separator, GetValidPaths().Concat(GetInvalidPaths()));
-            _environment.GetEnvironmentVariable("PATH").Returns(pathVariable);
+            _environment.GetEnvironmentVariable("PATH").Returns(x => pathVariable);
```
sharwell, line 33:
```suggestion
            _environment.GetEnvironmentVariable("PATH").Returns(_ => pathVariable);
```

UnitTests/GitCommands.Tests/Git/Gpg/GitGpgControllerTests.cs
```diff
@@ -161,9 +161,6 @@ public void Validate_GetTagVerifyMessage(int usefulTagRefNumber, string expected
                         var gitRef = new GitRef(_module, objectId, "refs/tags/TagName");
                         revision.Refs = new[] { gitRef };
 
-                        var args = new GitArgumentBuilder("verify-tag") { gitRef.LocalName };
-                        validate = _executable.StageOutput(args.ToString(), "");
```
sharwell, line 165: 📝 I'm not sure what this was testing, so not sure what impact removing it has

Plugins/Statistics/GitStatistics/LineCounter.cs
```diff
@@ -21,7 +21,7 @@ public sealed class LineCounter
 
         public void FindAndAnalyzeCodeFiles(string filePattern, string directoriesToIgnore, IEnumerable<string> filesToCheck)
         {
-            var extensions = filePattern.Replace("*", "").Split(';').ToHashSet(StringComparer.InvariantCultureIgnoreCase);
+            var extensions = LinqExtensions.ToHashSet(filePattern.Replace("*", "").Split(';'), StringComparer.InvariantCultureIgnoreCase);
```
sharwell, line 24: ❔ Intentional change? Do we not want to just remove `LinqExtensions` now that `System.Linq.Enumerable` has this method?

GitUI/CommandsDialogs/FormResolveConflicts.cs
RussKie: These will be replaced with [`System.Windows.Forms.TaskDialog`](https://docs.microsoft.com/dotnet/api/system.windows.forms.taskdialog)

ResourceManager/LinkFactory.cs
RussKie: We had to do a bit of work in dotnet/winforms to address issues in this area. 
E.g. https://github.com/dotnet/winforms/issues/1644 and https://github.com/dotnet/winforms/issues/3632

I've done some testing in the app, and - no, no observable change in behaviours. `RichTextBox.DetectUrls = false` does the magic (https://github.com/dotnet/winforms/issues/3632#issuecomment-752543714)

GitUI/GitUIExtensions.cs
RussKie: 👍  I will raise a separate issue for each once this PR is merged.

UnitTests/GitUI.Tests/ThreadHelperTests.cs
RussKie: 👍 will do once this is merged

GitUI/CommitInfo/CommitInfo.cs
RussKie: Yes, as per the comment. Setting the height to a positive number makes it visible, and it starts triggering resize event.

GitUI/CommandsDialogs/FormEditor.cs
RussKie: Apparently not....

GitCommands/CommitTemplateItem.cs
RussKie: Yes, will do once this is merged.
I had a brief look at the code, and looks like this serialization may have been used originally before it was updated to json-based.

UnitTests/ResourceManager.Tests/CommitDataRenders/CommitDataHeaderRendererTests.cs
RussKie: Apparently not

Plugins/Statistics/GitStatistics/LineCounter.cs
RussKie: This is outside the scope of this PR. But will certainly raise a follow up issue.

UnitTests/GitCommands.Tests/Git/Gpg/GitGpgControllerTests.cs
RussKie: image

Looks a combination of a) badly arranged test - for non-deref git command isn't invoked, and b) update in NUnit, that now starts asserting where previously it wouldn't.

GitUI/Plugin/PluginRegistry.cs
RussKie: Plugin loading is currently disabled because I couldn't get it working. Partially because the build props and targets need to be reviewed and rewritten for `dotnet build` instead of the dotnet/arcade scripts. 
This is will be a follow up work

GitUI/CommandsDialogs/FormCleanupRepository.cs
RussKie: No, the original code was creating the modern folder browser dialog, which Windows Forms received in .NET Core timeframe only (previously it was old pre-Vista style). See https://github.com/dotnet/winforms/pull/57
Replacing with the Windows Forms variant now.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
RussKie: I haven't observed any side-effects. 
We don't appear to be relying on order of items in all but one case:
https://github.com/gitextensions/gitextensions/blob/f840a0ab83a461035f816951fe39c752a92abc4c/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs#L314-L318

@mstv you've looked at this code recently. Any thought?

GitExtensions/Project.Build.targets
RussKie: This is part of outstanding "Packaging" task (in the 1st post)

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormInitTests.cs
RussKie: I don't want to tinker with this in the context of the PR. I'll raise a new issue once this is merged.

RussKie:
@sharwell I believe I addressed the actionable feedback, and captured the follow up items in the 1st post.
Ready for more.

RussKie:
image

🎉

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormInitTests.cs
sharwell: Does `TestCase` have a `Skip` property (like `Fact` has in XUnit)?

gerhardol: See above. These are Windows/NTFS-ish specific use cases.

Directory.Build.props
```diff
@@ -1,20 +1,25 @@
 <Project>
-  <Import Project=".\scripts\sdk\Sdk.props" />
```
sharwell, line 2: ❔ Is this file still used elsewhere? If not, was it deleted?

Directory.Build.targets
```diff
@@ -1,6 +1,4 @@
 <Project>
-  <Import Project=".\scripts\sdk\Sdk.targets" />
```
sharwell, line 2: ❔ Is this file still used elsewhere? If not, was it deleted?

GitCommands/GitCommands.csproj
```diff
@@ -1,39 +1,40 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <UseWindowsForms>true</UseWindowsForms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <Nullable>enable</Nullable>
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
+    <RunAnalyzersDuringLiveAnalysis>false</RunAnalyzersDuringLiveAnalysis>
```
sharwell, line 8: ❔ Why are we disabling analyzers here? Can we have a bug filed to enabled them again?

GitCommands/GitCommands.csproj
```diff
@@ -1,39 +1,40 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
+    <OutputType>Library</OutputType>
```
sharwell, line 4:
```suggestion
```

GitExtUtils/GitExtUtils.csproj
```diff
@@ -2,22 +2,19 @@
 
   <PropertyGroup>
     <Nullable>enable</Nullable>
```
sharwell, line 4: This is a duplicate of an inherited property

```suggestion
```

GitExtUtils/GitExtUtils.csproj
```diff
@@ -2,22 +2,19 @@
...
-  </PropertyGroup>
-
-  <ItemGroup>
-    <Compile Include="..\CommonAssemblyInfo.cs" Link="Properties\CommonAssemblyInfo.cs" />
-  </ItemGroup>
+    <OutputType>Library</OutputType>
```
sharwell, line 5:
```suggestion
```

GitExtensions/GitExtensions.csproj
```diff
@@ -28,21 +28,19 @@
...
     <ProjectReference Include="..\GitExtUtils\GitExtUtils.csproj" />
+    <ProjectReference Include="..\GitUI\GitUI.csproj" />
     <ProjectReference Include="..\Plugins\GitUIPluginInterfaces\GitUIPluginInterfaces.csproj" />
-    <ProjectReference Include="..\GitCommands\GitCommands.csproj" />
     <ProjectReference Include="..\ResourceManager\ResourceManager.csproj" />
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="app.config" />
-    <None Include="app.manifest" />
```
sharwell, line 44: ❔ Are these files still used elsewhere? If not, were they deleted?

GitExtensions/GitExtensions.csproj
```diff
@@ -28,21 +28,19 @@
...
+    <None Remove="AutoCompleteRegexes.txt" />
```
sharwell, line 42: 💭 I never really saw this as necessary. Including the item in both groups seems acceptable, and is less noisy.

```suggestion
```

GitExtensions/Program.cs
```diff
@@ -44,16 +43,18 @@ private static void Main()
...
+#endif
 
             SystemEvents.UserPreferenceChanged += (s, e) =>
             {
                 // Whenever a user changes monitor scaling (e.g. 100%->125%) unload and
                 // reload the theme, and repaint all forms
                 if (e.Category == UserPreferenceCategory.Desktop || e.Category == UserPreferenceCategory.VisualStyle)
                 {
-                    ThemeModule.ReloadWin32ThemeData();
+                    ////ThemeModule.ReloadWin32ThemeData();
```
sharwell, line 57:
```suggestion
#if SUPPORT_THEMES
                    ThemeModule.ReloadWin32ThemeData();
#endif
```

GitUI/GitUI.csproj
```diff
@@ -1,50 +1,34 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <OutputType>Library</OutputType>
```
sharwell, line 4:
```suggestion
```

GitUI/Theming/LocalHook.cs
```diff
@@ -0,0 +1,480 @@
...
+
+            Marshal.WriteIntPtr(Result.m_Handle, IntPtr.Zero);
+
+            try
+            {
+                ////NativeAPI.LhInstallHook(
+                ////    InTargetProc,
+                ////    Marshal.GetFunctionPointerForDelegate(Result.m_HookProc),
+                ////    GCHandle.ToIntPtr(Result.m_SelfHandle),
+                ////    Result.m_Handle);
```
sharwell, line 411: All commented code in this file should be updated to this pattern:

```suggestion
#if SUPPORT_THEMES
                NativeAPI.LhInstallHook(
                    InTargetProc,
                    Marshal.GetFunctionPointerForDelegate(Result.m_HookProc),
                    GCHandle.ToIntPtr(Result.m_SelfHandle),
                    Result.m_Handle);
#endif
```

Plugins/GitUIPluginInterfaces/GitUIPluginInterfaces.csproj
```diff
@@ -1,25 +1,25 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <OutputType>Library</OutputType>
```
sharwell, line 4:
```suggestion
```

RepoLayout.props
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information. -->
+<Project>
```
sharwell, line 3: ❔ Can we move this into an **eng** folder?

ResourceManager/ResourceManager.csproj
```diff
@@ -1,30 +1,31 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Nullable>enable</Nullable>
+    <OutputType>Library</OutputType>
```
sharwell, line 4:
```suggestion
```

Solution.Versions.props
```diff
@@ -1,41 +1,44 @@
 <Project>
   <PropertyGroup>
+    <MicrosoftNETCoreAppPackageVersion>5.0.0-rtm.20480.9</MicrosoftNETCoreAppPackageVersion>
```
sharwell, line 3: 💭 Is this necessary? It seems like this would be specified in **global.json**

Solution.Versions.props
```diff
@@ -1,41 +1,44 @@
 <Project>
```
sharwell, line 1: 💡 Consider filing a follow-up issue to migrate to [CentralPackageVersions](https://github.com/microsoft/MSBuildSdks/tree/main/src/CentralPackageVersions), which appears to be quite easy to adopt in non-arcade scenarios.

UnitTests/Directory.Build.props
```diff
@@ -1,5 +1,6 @@
 <Project>
   <Import Project="..\Directory.Build.props" />
```
sharwell, line 2: 💡 This works, but the most "copy-paste-friendly" way to add this line in non-root files is this:

```
<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)\.., Directory.Build.props))\Directory.Build.props" />
```

GitUI/Theming/ThemeLoader.cs
```diff
@@ -135,25 +143,47 @@ private static string[] GetClassNames(string themeFileName, StyleRule rule)
...
-            if (style.Name != ColorProperty || !(style.Term is HtmlColor htmlColor))
+            // Wish we could write the following:
+            //
+            //      Property colorProperty = rule.Style.Declarations.FirstOrDefault(d => d.Name == ColorProperty);
+            //      Color color = ColorTranslator.FromHtml(colorProperty.DeclaredValue.Original.Text);
+            //
+            // colorProperty.DeclaredValue.Original.Text contains the required information, e.g. #f9f9f9
+            // but none of the properties nor types are public, thus leaving us with either atempting to
+            // access the data via the reflection, or parsing the raw text.
+            // Prefer the latter option - less magic.
```
sharwell, line 164: 💡 It looks like better options are available in newer versions of ExCSS. For example:
https://github.com/TylerBrinks/ExCSS/blob/e6fec79cf7ba84c3a7ec2012aaeb2955a790ad98/src/ExCSS/Values/Color.cs

sharwell (commented):
At this point I've reviewed everything except the submodule changes and two CI build scripts

.editorconfig
```diff
@@ -240,3 +240,6 @@ dotnet_diagnostic.SA1135.severity = none
 
 # SA1629: Documentation text should end with a period
 dotnet_diagnostic.SA1629.severity = none
+
+# CA1416: Validate platform compatibility
+dotnet_diagnostic.CA1416.severity = none
```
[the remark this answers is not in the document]
RussKie: https://github.com/dotnet/roslyn-analyzers/issues/4983

Directory.Build.props
RussKie: This is a part of the current arcade build scripts. I plan to review the whole build pipeline once this is merged.
Some of these implementations can surely be removed when the remaining tasks (such as plugin loading, packaging, etc.) are resolved.

RepoLayout.props
RussKie: https://github.com/gitextensions/gitextensions/pull/8522#discussion_r622826901

GitUI/Theming/ThemeLoader.cs
RussKie: I struggled to upgrade the code to the new version... Happy to look over it again once this is merged.

sharwell: I meant to say this would be a good follow-up issue. 👍

mstv:
It works! 🚀
I could rebase my work branchs `all5` and `fix/6527_git_errors` without problems as far as I can see. 🎉

I noticed three minor glitches:

1. `FileDiff` encoding is unset (I guess I must allow to store it once).
 
FileDiff toolbar

2. `FileStatusList` group headings consume (too) much additional space

FileStatusList

3. `CommitInfo` has different indents. For `Parents:` it even depends on the commit.

CommitInfo

GitExtensions/GitExtensions.csproj
RussKie: This is what VS writes out. I'm going to leave it as is, otherwise it would create a noise if the csproj in modified in VS.

RussKie:
@msftbot merge in 4 days

GitCommands/CommitTemplateItem.cs
RussKie: https://github.com/gitextensions/gitextensions/issues/9150

RussKie:
> I noticed three minor glitches:

Please raise them as separate issue.

Solution.Versions.props
RussKie: * https://github.com/gitextensions/Git.hub/pull/20
* https://github.com/gitextensions/conemu-inside/pull/30
* https://github.com/gitextensions/gitextensions/pull/9161

gerhardol (commented):
A lot of functionality is lost, but this is a big step anyway

RussKie:
The only functionality which is truly lost at this stage is theming.
Everything else is being restored, I'm working on it.

ivangrek:
👍

NikolayXHD:
> The only functionality which is truly lost at this stage is theming.

Started working on it 
cc: @RussKie

RussKie:
Awesome! I'm currently having troubles building VC projects with
`dotnet`... Maybe you have pointers.

Also maybe worth looking into Detours for rerouting Win32 calls

NikolayXHD:
Currently not much luck with porting the theming.

Easy hook fails to install some hooks, Other hooks do install, but cause [System.ExecutionEngineException](https://docs.microsoft.com/en-us/dotnet/api/system.executionengineexception?view=net-5.0) at Managed to Native Transitions. Which native call raises the exceptions varies, depending on what controls are displayed.

I can tell, it's not the issue of EasyHook himself, rather with .dll's built for net50.

1. If i use the same managed `EasyHook.dll` built against netstandard20 + unmanaged `EasyHook32.dll`, `EasyHook64.dll` with net461, they work just fine, like the original EasyHook.

2. I tried alternative hooking library Reloaded.Hooks

It does install the hooks, but `System.ExecutionEngineException` stays exactly the same.

I do the experiments in `net50` branch in a small repository https://github.com/NikolayXHD/TestEasyHook

What prevents me from moving forward at the moment, is I don't have a question specific enough to ask hooking library authors for help.

> Also maybe worth looking into Detours for rerouting Win32 calls
will certainly try it.

NikolayXHD:
> Awesome! I'm currently having troubles building VC projects with `dotnet`... Maybe you have pointers.

Building EasyHook VC projects from Project.Build.targets fails to find the *.h files.
I was able to build from Visual Studio. Open Externals\EasyHook\EasyHook.sln, from open solutions it builds fine.

drewnoakes:
@NikolayXHD if you run with the [Managed Debug Assistants](https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/diagnosing-errors-with-managed-debugging-assistants) turned on in VS, do they flag any kind of interop issues? It may be that a calling convention or argument marshalling is incorrectly specified, and that it worked under .NET Framework but fails under .NET Core.

RussKie:
Let's start a new thread about this.

mstv:
> 2. `FileStatusList` group headings consume (too) much additional space
>
> FileStatusList

This has been fixed with an update on the recent Patch Day.

RussKie:
Yes, it was fixed in .NET 5.0.9, IIRC.

PR #8526, opened 2020-10-08
Disabled "Reset" mnemonic key in "Reset changes" form

### LINKED ISSUE #8467: Set "Cancel" button as the Reset prompt default
## Feature description

So, I use "R" a lot to open the "Reset Changes" prompt and resetting some of the files that I didn't want to change, but I've ran into a problem today, right on the Commits window, where the button labeled as "Reset" on the "Reset Changes" prompt is focused instead of the button labeled as "Cancel" 

Pressing "R" pops up this prompt asking you if you want to reset your changes, but double tapping "R" accepts which ever is highlighted, creating a mistake-prone behaviour

So, I pressed "R" but my finger slipped away a second time, hitting "R" again, successfully aproving the prompt and consequely, resetting my changes.

And so, I reset all the changes from my files, without even noticing

I wish that this behaviour is set to the opposite action, cancel, rather than a quick way of resetting changes cause, it's easier to click it again and say "I want to reset my changes" rather than "Now I have to redo everything because of my mistake"

Thank you, very much!

Screenshot_1

## Environment

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

oriash93:
Fixes #8467 

## Proposed changes

- Disabled "Reset" button mnemonic key feature in "Reset changes" form

----

### DIFF
--- a/GitUI/CommandsDialogs/FormResetChanges.Designer.cs
+++ b/GitUI/CommandsDialogs/FormResetChanges.Designer.cs
@@ -53,9 +53,10 @@
             this.btnReset.Name = "btnReset";
             this.btnReset.Size = new System.Drawing.Size(95, 25);
             this.btnReset.TabIndex = 1;
-            this.btnReset.Text = "&Reset";
+            this.btnReset.Text = "Reset";
             this.btnReset.UseVisualStyleBackColor = true;
             this.btnReset.Click += new System.EventHandler(this.btnReset_Click);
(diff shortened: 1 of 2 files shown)


RussKie:
Thanks for having a go at it. It is generally a good idea to post on the issue that you want to work on and have it assigned to you to avoid situations like this.

PR #8529, opened 2020-10-08
Fix #8188: cannot force push tags

### LINKED ISSUE #8188: Unable force push tag
## Current behaviour

```
"C:\Program Files\Git\bin\git.exe" push --force-with-lease --progress "remote" tag v3.4
To https://github.com/gitextensions/gitextensions.git
 ! [rejected]            v3.4 -> v3.4 (stale info)
error: failed to push some refs to 'https://github.com/gitextensions/gitextensions.git'
```

This however works:
```
$ git push --force --progress "remote" tag v3.4
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
Writing objects: 100% (1/1), 160 bytes | 80.00 KiB/s, done.
Total 1 (delta 0), reused 0 (delta 0), pack-reused 0
To https://github.com/gitextensions/gitextensions.git
 + 699c02c10...dd6d6ddce v3.4 -> v3.4 (forced update)
```

For tags `--force-with-lease` doesn't work, it has to be `--force`.

## Steps to reproduce

1. Create a tag and push to remote
2. Delete the local tag
3. Re-create the tag at a different commit
4. Force push the tag via `FormPush` (tags tab) with `Force` option.

No idea, never tried it before.

## Environment

- Git Extensions 3.4.0.9381
- Git 2.26.0.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

lhiginbotham:
Fix #8188: cannot force push tags
- Checking the "force push" option in the tags tab of the push form resulted in passing ForcePushOptions.ForceWithLease to the git push command builder
- Add a check for the force push checkbox for the tags tab in order to pass the correct ForcePushOptions.Force value to the command builder

## Test methodology 

- I've tested this out with a dummy remote repo on GitHub and verified that I am now able to force push a tag from one commit to another

## Test environment(s) 

- GIT 2.28.0.windows.1
- Windows 10

----

### DIFF
--- a/GitUI/CommandsDialogs/FormPush.cs
+++ b/GitUI/CommandsDialogs/FormPush.cs
@@ -473,7 +473,7 @@ namespace GitUI.CommandsDialogs
 
         private ForcePushOptions GetForcePushOption()
         {
-            if (ForcePushBranches.Checked)
+            if (ForcePushBranches.Checked || ForcePushTags.Checked /* tags cannot be pushed using --force-with-lease */)
             {
                 return ForcePushOptions.Force;
             }
@@ -1262,6 +1262,26 @@ namespace GitUI.CommandsDialogs
 
                 pushCheckBox.Value = willPush(row);
             }
+        }
+
+        internal TestAccessor GetTestAccessor() => new TestAccessor(this);
+
+        internal readonly struct TestAccessor
+        {
+            private readonly FormPush _form;
+
+            public TestAccessor(FormPush form)
+            {
+                _form = form;
+            }
+
+            public CheckBox ckForceWithLease => _form.ckForceWithLease;
+
+            public CheckBox ForcePushBranches => _form.ForcePushBranches;
+
+            public CheckBox ForcePushTags => _form.ForcePushTags;
+
+            public ForcePushOptions GetForcePushOption() => _form.GetForcePushOption();
         }
     }
 }
--- a/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPushTests.cs
+++ b/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPushTests.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using CommonTestUtils;
+using FluentAssertions;
+using GitCommands.Git;
+using GitUI;
+using GitUI.CommandsDialogs;
+using NUnit.Framework;
+
+namespace GitExtensions.UITests.CommandsDialogs
+{
+    [Apartment(ApartmentState.STA)]
+    public class FormPushTests
+    {
+        // Created once for the fixture
+        private ReferenceRepository _referenceRepository;
+
+        // Created once for each test
+        private GitUICommands _commands;
+
+        [SetUp]
+        public void SetUp()
+        {
+            if (_referenceRepository == null)
+            {
+                _referenceRepository = new ReferenceRepository();
+            }
+            else
+            {
+                _referenceRepository.Reset();
+            }
+
+            _commands = new GitUICommands(_referenceRepository.Module);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+        }
+
+        [OneTimeTearDown]
+        public void OneTimeTearDown()
+        {
+            _referenceRepository.Dispose();
+        }
+
+        // Note: the DataBindings between ForcePushTags and ForcePushBranches or ckForceWithLease (depending on Git version) do not function in this test environment
+        [TestCase(false, false, false, ForcePushOptions.DoNotForce)]
+        [TestCase(false, true, false, ForcePushOptions.Force)]
+        [TestCase(false, false, true, ForcePushOptions.Force)] // ForcePushTag requires normal force as with-lease is not allowed for tags
+        [TestCase(false, true, true, ForcePushOptions.Force)]
+        [TestCase(true, false, false, ForcePushOptions.ForceWithLease)] // would be ForcePushOptions.DoNotForce if DataBindings were working
+        [TestCase(true, true, false, ForcePushOptions.Force)]
+        [TestCase(true, false, true, ForcePushOptions.Force)] // ForcePushBranches and ForcePushTags take precedence over ckForceWithLease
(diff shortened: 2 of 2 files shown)


vbjay:
Force push with lease is there to prevent code stomps.

If you want to make it work. Do a fetch all and review history and make
sure you are ok with overwriting stuff. Rebase as needed and then push.

RussKie:
> Force push with lease is there to prevent code stomps.

You can't push tags with lease.

RussKie (commented):
👍 
Can you please add a test for `GetForcePushOption` as well?

vbjay:
Got it. Tags vs commits.

lhiginbotham:
@RussKie 
> Can you please add a test for `GetForcePushOption` as well?

I am not sure how I would go about adding a test for this method:  1) it is a `private` method of `GitUI.CommandsDialogs.FormPush` and 2) there do not appear to be _any_ tests for the `FormPush` class that I can see...  Are you suggesting I write some tests that somehow execute the form for the various possible "force push" button states and then somehow check what git command was generated?  Otherwise, I can change the code to expose `GetForcePushOption()` as a `public static` method of `FormPush`, taking in the three button states as parameters and then just simply test this new `public` method in a simple test suite.

RussKie:
> 1) it is a `private` method of `GitUI.CommandsDialogs.FormPush`

We test non-public API with a help of `TestAccessor` struct. There are few examples, you can have a look at `FormPull.TestAccessor`, this would probably be the closest to what you'd need.


> 2) there do not appear to be _any_ tests for the `FormPush`

Bummer! But we can fix this :)
I think you can pretty much copy `FormPullTest` class, remove all but one test that would toggle checkboxes and assert that `form.GetTestAccessor().GetForcePushOption()` returns the expected.

This should be fairly straight forward, but holler if you get stuck.
Thank you

lhiginbotham:
@RussKie 
Done!  I hope this is what you were looking for in terms of tests.

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -1254,5 +1254,25 @@ private void SetBranchesPushCheckboxesState(Func<DataGridViewRow, bool> willPush
...
+            public TestAccessor(FormPush form)
+            {
+                _form = form;
+            }
+
+            public CheckBox ForcePushBranches() => _form.ForcePushTags;
+
+            public CheckBox ckForceWithLease() => _form.ckForceWithLease;
+
+            public CheckBox ForcePushTags() => _form.ForcePushTags;
```
RussKie, line 1273: No need for methods, use readonly properties:
```suggestion
            public CheckBox ForcePushTags => _form.ForcePushTags;

            public CheckBox ckForceWithLease => _form.ckForceWithLease;
```

Why is `_form.ForcePushTags` referenced twice under different names?

RussKie (changes requested):
👍

GitUI/CommandsDialogs/FormPush.cs
lhiginbotham: @RussKie 😦 good catch, this is an error.  It should be `_form.ForcePushBranches`...

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPushTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        [OneTimeTearDown]
+        public void OneTimeTearDown()
+        {
+            _referenceRepository.Dispose();
+        }
+
+        [TestCase(false, false, false, ForcePushOptions.DoNotForce)]
+        [TestCase(false, true, true, ForcePushOptions.Force)]
+        [TestCase(false, true, false, ForcePushOptions.DoNotForce)] // ForcePushBranches state is bound to ForcePushTags state
```
lhiginbotham, line 50: The state bounding actually depends on whether or not `GitVersion.Current.SupportPushForceWithLease`, so this needs modified in some way.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormPushTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        [TestCase(false, false, true, ForcePushOptions.Force)] // ForcePushTag requires normal force as with-lease is not allowed for tags
+        [TestCase(true, false, false, ForcePushOptions.ForceWithLease)]
+        [TestCase(true, true, false, ForcePushOptions.ForceWithLease)] // ForcePushBranches state is bound to ForcePushTags state
+        [TestCase(true, false, true, ForcePushOptions.Force)] // ForcePushBranches and ForcePushTags take precedence over ckForceWithLease
+        [TestCase(true, true, true, ForcePushOptions.Force)] // ForcePushBranches and ForcePushTags take precedence over ckForceWithLease
```
lhiginbotham, line 55: @RussKie 
When I changed `public CheckBox ForcePushBranches() => _form.ForcePushTags;` to `public CheckBox ForcePushBranches => _form.ForcePushBranches;`, I got some test failures, as I had expected.  When fixing up the test cases and reviewing them once they all were passing, I had discovered that it seems like databinding is not actually working in this test.  I expected that if you had `forcePushBranchWithLeaseChecked = true`, `forcePushBranchChecked = false`, `forcePushTagChecked = false`, we would get the result of `ForcePushOptions.DoNotForce`, since setting the force push tag to false should have resulted in force push branch with lease to be false...

I did a little bit of searching on the internet to see how to resolve this and came across this [StackOverflow answer](https://stackoverflow.com/a/10040854) and tried a few solutions from the linked answer's comments but could not get databinding to work in the test.  I will remove the comments about bound state and make a comment, something like that the databinding is not working for the test and that the expected test case results reflect this.

GitUI/CommandsDialogs/FormPush.cs
lhiginbotham: Fixed!

RussKie (approved):

RussKie:
Please squash and it is ready to go.

lhiginbotham:
@RussKie this is good to be merged, thanks!

RussKie (approved):

RussKie:
Thank you

PR #8532, opened 2020-10-09
Detect when Azure DevOps token is invalid/expired

pmiossec:
## Proposed changes

- Detect when Azure DevOps token is invalid/expired
and display an explicit error message.

Better than the current behavior that is displaying nothing
so the user doesn't understand what is happening (even me that developped the plugin 🤣  )

It happens sometimes and is difficult to understand/detect
when the token has expired (because maximum duration is 1 year)

### Before

Nothing. Azure Devops pipeline results are just not displayed.

### After

An error message is displayed **once**:
image

A tooltip is always displayed on the "Working directory" pseudo commit result (in case the user didn't carefully read the popup message):
image

## Test methodology 

- Manual (change the token to set a bad one to see the new behavior)

## Test environment(s) 

- Git Extensions 33.33.33
- Build 905403fe2edfa879c96e293973afebe555eb6e8e
- Git 2.28.0.windows.1
- .NET Framework 4.8.4240.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
+++ b/Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
@@ -13,6 +13,7 @@ using GitUI;
 using GitUIPluginInterfaces;
 using GitUIPluginInterfaces.BuildServerIntegration;
 using Microsoft.VisualStudio.Threading;
+using Microsoft.WindowsAPICodePack.Dialogs;
 using ResourceManager;
 
 namespace AzureDevOpsIntegration
@@ -61,9 +62,11 @@ The build server integration has been disabled for this session.");
 As a consequence, the build server integration has been disabled for this session.
 
 Detail of the error:");
+
+        private Action _openSettings;
         private string CacheKey => _projectUrl + "|" + _settings.BuildDefinitionFilter;
 
-        public void Initialize(IBuildServerWatcher buildServerWatcher, ISettingsSource config, Func<ObjectId, bool> isCommitInRevisionGrid = null)
+        public void Initialize(IBuildServerWatcher buildServerWatcher, ISettingsSource config, Action openSettings, Func<ObjectId, bool> isCommitInRevisionGrid = null)
         {
             if (_buildServerWatcher != null)
             {
@@ -71,6 +74,7 @@ Detail of the error:");
             }
 
             _buildServerWatcher = buildServerWatcher;
+            _openSettings = openSettings;
             _settings = IntegrationSettings.ReadFrom(config);
 
             if (!_settings.IsValid())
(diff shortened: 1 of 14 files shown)


Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -48,6 +50,19 @@ internal class AzureDevOpsAdapter : IBuildServerAdapter
         // * there is only one instance at every given time (link to the revision grid and recreated on revision grid refresh)
         // * data is created by the first instance and used in readonly by the later instances
         private static CacheAzureDevOps CacheAzureDevOps = null;
+        private static string ProjectOnErrorKey = null;
+
+        private readonly TranslationString _buildIntegrationErrorCaption = new TranslationString("Build Server Integration results error");
```
RussKie, line 55:
```suggestion
        private readonly TranslationString _buildIntegrationErrorCaption = new TranslationString("Azure DevOps error");
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -48,6 +50,19 @@ internal class AzureDevOpsAdapter : IBuildServerAdapter
...
+        private readonly TranslationString _badTokenErrorMessage = new TranslationString(@"The personal access token used to request build server results is invalid or has expired.
+
+Please update it in the 'Build server integration' settings.
+
+As a consequence, the build server integration has been disabled for this session.");
```
RussKie, line 60:
```suggestion
        private readonly TranslationString _badTokenErrorMessage = new TranslationString(@"The personal access token is invalid or has expired. Update it in the 'Build server integration' settings.

The build server integration has been disabled for this session.");
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -122,15 +137,41 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                {
+                    _apiClient = null;
+                    observer.OnNext(new BuildInfo { CommitHashList = new[] { ObjectId.WorkTreeId }, Description = errorMessage, Status = BuildInfo.BuildStatus.Failure });
+                    observer.OnCompleted();
 
-                    CacheAzureDevOps = new CacheAzureDevOps { Id = CacheKey, BuildDefinitions = _buildDefinitions };
+                    if (ProjectOnErrorKey == null || ProjectOnErrorKey != CacheKey)
+                    {
+                        ProjectOnErrorKey = CacheKey;
+                        MessageBox.Show(errorMessage, _buildIntegrationErrorCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
```
RussKie, line 173: I think we could do a little better with the use of Task Dialog - we show the message (in the same way we do it now) and have two buttons: `[Fix PAT]` and `[Continue]` (the buttons text is up for a discussion).
The first button will take a user to settings where a new PAT can be inserted, and the other button basically ignores the error.

RussKie:
This is a great idea 🚀

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
pmiossec: >The first button will take a user to settings where a new PAT can be inserted

I had a look but unfortunately it seems that we have no way to open the user settings from a build integration plugin...

RussKie: Other plugins can do it:
https://github.com/gitextensions/gitextensions/blob/588fc0f5b0666855a225d9507ff17ee78f2fe27e/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs#L225
https://github.com/gitextensions/gitextensions/blob/588fc0f5b0666855a225d9507ff17ee78f2fe27e/Plugins/JiraCommitHintPlugin/JiraCommitHintPlugin.cs#L60

Do you think you could spend a little more time here to see what can be done?

pmiossec: Working WIP pushed...

image

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -158,6 +229,22 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+        private void BtnUpdateToken_Click(object sender, EventArgs e)
+        {
+            CloseTaskDialogWithResult(sender, TaskDialogResult.Yes);
+        }
+
+        private static void CloseTaskDialogWithResult(object sender, TaskDialogResult taskDialogResult)
+        {
+            var taskDialog = (TaskDialog)((TaskDialogButton)sender).HostingDialog;
+            taskDialog.Close(taskDialogResult);
+        }
```
RussKie, line 246: Please use local functions and capture locals, the same way as used in other places. This will make it simpler to migrate to the Windows Forms Task Dialog.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -122,15 +139,69 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                        string errorMessage = _badTokenErrorMessage.Text;
+                        ManageError(errorMessage);
+
+                        if (ProjectOnErrorKey == null || ProjectOnErrorKey != CacheKey)
+                        {
+                            ProjectOnErrorKey = CacheKey;
+
+                            var btnUpdateToken = new TaskDialogButton("btnUpdateToken", "Update token");
+                            btnUpdateToken.Click += BtnUpdateToken_Click;
+                            var btnIgnoreError = new TaskDialogButton("btnIgnoreError", "Ignore error");
```
RussKie, line 165:
```suggestion
                            var btnIgnoreError = new TaskDialogButton("btnIgnoreError", "Ignore");
```

RussKie (commented):
👍

pmiossec:
done.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
pmiossec: I have renamed the "Update token" in "Open settings" (see the updated screenshot in the first message) because:
1. That's what is done in fact
2. That allows the taskdialog to be reused for other reasons where we detect a bad configuration. For example if no build definitions are found...

RussKie (approved):

RussKie:
Thank you

PR #8537, opened 2020-10-10
Fixes #8146 "Keyboard navigation : Enter Key should open a recent repository"

### LINKED ISSUE #8146: Keyboard navigation : Enter Key should open a recent repository
## Feature description

When on the startup screeen, 
and navigating through the recent repositories 
should be able to press enter to open a repository

## Environment
- GitExtensions version:  3.3.1.7897
- GIT version:  2.22.0
- OS version: MSWNT 10.0
- .NET version: 4.8.4180

lhiginbotham:
Fixes #8146 "Keyboard navigation : Enter Key should open a recent repository"

## Proposed changes

- intercept "Enter" key event and attempt to open the selected repository, if anything is selected

## Test methodology 

- I've attempted to use the enter key in a variety of scenarios, including when there is no repository selected, when there is a repository selected, when there is no repository selected but the list group is collapsed, and when there is a repository selected but the list group is collapsed
- - Out of these scenarios, there seems to be an issue only when pressing enter when a repository is selected but the list group is collapsed.  Since the list group was collapsed, I would have expected a no-op, since we cannot visually tell what repository is selected.  I cannot fix this, however, as it seems like the `ListViewGroup.CollapsedState` property is not accessible (maybe we need to use a newer version of .NET in order to see the API implemented and exposed in [this WinForms PR](https://github.com/dotnet/winforms/pull/3155)? @RussKie )

## Test environment(s) 

- Windows 10

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -330,6 +330,20 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
             handler?.Invoke(this, args);
         }
 
+        protected override bool ProcessDialogKey(Keys keyData)
+        {
+            if (keyData == Keys.Enter)
+            {
+                // .NET 5.0 introduced collapsible `ListViewGoup`s, which we do not yet have API access to but still get rendered in the UI.
+                // Whenever the list of repos is collapsed but we still have a repo selected (and hidden), if we hit enter, the selected repo will
+                // be opened. This should be a no-op instead, however, since we cannot visually tell what repo is selected. When we upgrade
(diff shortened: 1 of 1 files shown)


RussKie:
> * it seems like the `ListViewGroup.CollapsedState` property is not accessible

Yes, the `ListView` control was enhanced in .NET 5.0, so we'll have access to new features when we upgrade.

> * Out of these scenarios, there seems to be an issue only when pressing enter when a repository is selected but the list group is collapsed.

I suppose we can live with that until we upgrade to .NET 5.0.
The only thing I'd suggest is adding a comment in the code noting this behaviour.

RussKie (commented):
🚀

lhiginbotham:
@RussKie I've added the comment

RussKie (approved):

RussKie:
Thank you

PR #8542, opened 2020-10-14
fix german translation for Repositories

forki:
Small change to german translation. "Repositorys" looks really weird


gerhardol:
Can you please make the changes to the sources in Transifex?
https://github.com/gitextensions/gitextensions/wiki/Translations

The translated files in GitHub are updated at import so no point in updating separately

PR #8546, opened 2020-10-15
Save Settings When User ran with args

### LINKED ISSUE #6142: Savesettings when user ran with args
## Current behaviour

See https://github.com/gitextensions/gitextensions/issues/6060#issuecomment-455914819
We save settings after running a command.

```c#
var commands = new GitUICommands(GetWorkingDir(args));
if (args.Length <= 1)
            {
                commands.StartBrowseDialog();
            }
else
            {
                // if we are here args.Length > 1
                commands.RunCommand(args);
            }

            AppSettings.SaveSettings(); // Right here.  If we run a command we shouldn't save settings.
        }
```

## Expected behaviour

RunCommand should 
- return a value telling us to save settings or not 
- skip saving and in the command running code force a save if desired.

## Steps to reproduce

Based on the comment I linked, I am thinking they ran ```gitext.cmd pull``` and it turned --autostash off for pull dialogs.  
## Environment

- Git Extensions 3.1.0.4945
- Build 231592f2e9e9bce58ebfb02cb124be86e4b4e563 (Dirty)
- Git 2.20.1.windows.1
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

racecrew:
Fixes #6142

## Proposed changes

- Use an out bool variable which indicates whether the app. settings should be saved or not
- Pass the variable as an argument to the method which run the commands

## Test methodology 

- Manual

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -84,6 +84,7 @@ namespace GitExtensions
                 ThreadHelper.JoinableTaskContext = new JoinableTaskContext();
             }
 
+            bool saveAppSettings;
             AppSettings.LoadSettings();
 
             if (EnvUtils.RunningOnWindows())
@@ -100,11 +101,13 @@ namespace GitExtensions
                 }
             }
 
+            bool telemetrySettingChanged = false;
             if (!AppSettings.TelemetryEnabled.HasValue)
             {
                 AppSettings.TelemetryEnabled = MessageBox.Show(null, Strings.TelemetryPermissionMessage,
                                                                Strings.TelemetryPermissionCaption, MessageBoxButtons.YesNo,
                                                                MessageBoxIcon.Question) == DialogResult.Yes;
+                telemetrySettingChanged = true;
             }
 
             try
@@ -156,6 +159,7 @@ namespace GitExtensions
 
             if (args.Length <= 1)
             {
+                saveAppSettings = true;
                 commands.StartBrowseDialog();
             }
             else
@@ -165,13 +169,16 @@ namespace GitExtensions
                 // Avoid replacing the ExitCode eventually set while parsing arguments,
                 // i.e. assume -1 and afterwards, only set it to 0 if no error is indicated.
                 Environment.ExitCode = -1;
-                if (commands.RunCommand(args))
+                if (commands.RunCommand(args, out saveAppSettings))
                 {
                     Environment.ExitCode = 0;
                 }
             }
 
-            AppSettings.SaveSettings();
+            if (saveAppSettings || telemetrySettingChanged)
+            {
+                AppSettings.SaveSettings();
+            }
         }
 
         [CanBeNull]
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1381,8 +1381,10 @@ namespace GitUI
                 });
         }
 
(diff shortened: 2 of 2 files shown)


racecrew:
How is the complex error should be solved?! Could you help pleae? Thank you in advance!

gerhardol:
> How is the complex error in CodeFactor Check could be solved?! Could anyone help somehow?

The PR has not changed the complexity to be much worse so a refactoring should be done in a separate PR anyway.
For me this is OK, I would not request a refactoring

There are a lot of issues in CodeFactor for anyone that is interested... 

@racecrew Thanks for your efforts. We are few maintainers with mostly limited time, not even @RussKie can do everything. Answers may take time...

racecrew:
@gerhardol Thank you for your answer! And I am sorry for my impatience, I am a newbie here and don't know how fast the answers should come... and I am worried a bit because I am not sure if my comments are seen or not. I will have more patience in future :-)

So, if I understand you correctly, the CodeFactor issue could be ignored, right? How does it work, one of the reviewers decides the PR is ok or not despite of the failed check?

gerhardol:
> I am not sure if my comments are seen or not. I will have more patience in future :-)

Most are seen. But for myself, I am often out of time after reading comments and may not react on all.
There are other things in my life and I want to spend time on _my_ features too.

> So, if I understand you correctly, the CodeFactor issue could be ignored, right? How does it work, one of the reviewers decides the PR is ok or not despite of the failed check?

For me, that is OK. It is our mighty leader @RussKie that has the final saying - he listen to arguments though.
In this case the reported reported two issues, but one was the same as before and second was not serious.
Complexity is just a measure, if you have a lot of resources you may be strict on keeping this down too.
(Personally I find complexity too dull to really bother with, but there are few measures in this area.)

racecrew:
Thank you for the clarification @gerhardol

GitExtensions/Program.cs
```diff
@@ -100,11 +101,13 @@ private static void RunApplication()
                 }
             }
 
+            bool telemetrySettingChanged = false;
             if (!AppSettings.TelemetryEnabled.HasValue)
             {
                 AppSettings.TelemetryEnabled = MessageBox.Show(null, Strings.TelemetryPermissionMessage,
                                                                Strings.TelemetryPermissionCaption, MessageBoxButtons.YesNo,
                                                                MessageBoxIcon.Question) == DialogResult.Yes;
+                telemetrySettingChanged = true;
```
RussKie, line 110: I would save settings here to reduce a chance of showing the same popup to a user if another instance of GE is started before the first instance is closed.

GitExtensions/Program.cs
```diff
@@ -165,13 +169,16 @@ private static void RunApplication()
                 // Avoid replacing the ExitCode eventually set while parsing arguments,
                 // i.e. assume -1 and afterwards, only set it to 0 if no error is indicated.
                 Environment.ExitCode = -1;
-                if (commands.RunCommand(args))
+                if (commands.RunCommand(args, out saveAppSettings))
```
RussKie, line 172: So the only way to have `saveAppSettings=true` here is if `args.Contains("browse")` right? 
Why not make this decision here rather than complicate the code in layers below?

GintasS:

PR #8549, opened 2020-10-17
FormPush multiple: force-with-lease if rejected

### LINKED ISSUE #8548: Push for non current rejected: force-with-lease should be offered
## Feature description

If git-push fails, GE offers to pull-merge/rebase or push-force-with-lease if the current branch is rejected. This is the same if the current-only or multiple tab is used.

push-current

(ahead/behind in this screenshot, that is another PR)
push-current-mult

push-failed

push-rejected-current

If any other branch is pushed, the push is just rejected. pull is not relevant for other than the current (not entirely, but it will be too complicated here).
Force box could be ticked (there are some limitations for force-with-lease per branch), but that could cause commits to be lost unintentionally.

The form should therefore offer force-with-lease also for non current branches.

force-only

### Background

I often work with one or more official branches on top of master but other commits on top of that with non official patches.
When rebasing or applying fixup to "early" commits, I now just have to "rebase another branch" and push multiple branches (after #8515). The push requires force and I want that to be safer than the raw --force.

## Environment

GE master a573768a5b2abebf0515fd95ec35291978f09219, after #8515

gerhardol:
Fixes #8548

## Proposed changes

- Slightly safer regex to check for current branch
- If push is rejected for other than current, offer a slim-down variant of the error dialog. Currently, this is a subset of the "full" dialog, this could change

Note that the full dialog is displayed if current branch is pushed at the same time as any other branch. This is OK, the pull actions have no effect on the non current branches

### Before

none - see #8548

### After

force-only

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormPush.cs
+++ b/GitUI/CommandsDialogs/FormPush.cs
@@ -66,10 +66,13 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _updateTrackingReference =
             new TranslationString("The branch {0} does not have a tracking reference. Do you want to add a tracking reference to {1}?");
 
-        private readonly TranslationString _pullRepositoryMainInstruction = new TranslationString("Pull latest changes from remote repository");
-        private readonly TranslationString _pullRepository =
+        private readonly TranslationString _pullRepositoryMainMergeInstruction = new TranslationString("Pull latest changes from remote repository");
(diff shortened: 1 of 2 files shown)


RussKie (approved):

gerhardol:
@msftbot merge in 24 hours

Issue #8553, opened 2020-10-18
Ahead/Behind misses information


gerhardol:
## Current behaviour

Ahead/behind misses information in some situations:
* If the local is up-to-date with the remote, nothing is shown (similar to no tracking branch)
(this is due to how ahead/behind is presented and can be fixed)
=> This can be fixed by checking if there is a tracking branch (or %(upstream:trackshort), but that adds runtime)

* If a branch has setup a tracking branch (remote removed for instance, many other ways...), you do not get any information.
=> for-each-ref displays this as "gone" should be presented to the user

These two are easy to handle

* If a branch do not have a tracking branch, for instance pushed in "multiple" tab in FormPush or .git/config edited, no information is displayed.
=> This could be partly fixed by examining hashes and refs, similar to what ParseRefs() and Push multiple does (both these implementations have limitations/bugs though, to be investigated). Full ahead/behind info will require separate Git calls, but there could be a separate marker

## Expected behaviour

Show Ahead/Behind info where possible

The AheadBehind info could be retrieved together with the branches, but it will be slower to list the branches in the sidepanel. For me with 1962 branches and 92 tracked, getting tracked info increases from 0.28 s to 0,50 s, so AheadBehind should be separate still.

## Steps to reproduce

The main scenarios
* A tracked branch in synch with the remote branch
* A pushed branch, the the remote branch is removed
* A pushed branch, remove the track information (eg edit .git/config)

## Did this work in previous version of GitExtensions

No

## Environment

- Git Extensions 33.33.33
- Build 5c23aa6d98a592bf9f33235be0236fd1287bf75d
- Git 2.28.0.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

gerhardol:
First two closed in #8562
Third closed in #8573

PR #8562, opened 2020-10-20
Ahead/behind show up-to-date and gone

gerhardol:
First part of #8553 

## Proposed changes

- Show ahead/behind if remote is up to date
- Show if the tracking branch no longer exists "gone"

### Before

Only the branch with differences displays the count
branch-before

No indication if the branches are up to date
push-none

### After

feature/i8553-ahead-zero-gone is not yet pushed
feature/i8548-multiple-force-push is deleted at the remote
feature/i8448-presentation is pushed and in sync with the remote
feature/i8448-range-diff-3 same as above (not displayed)
branch-after

After a branch is pushed
push-after

## Test methodology 

Tests adapted and expanded

----

### DIFF
--- a/GitCommands/Git/AheadBehindData.cs
+++ b/GitCommands/Git/AheadBehindData.cs
@@ -2,15 +2,22 @@
 {
     public struct AheadBehindData
     {
+        // gone: "plumbing" expression, see https://git-scm.com/docs/git-for-each-ref#Documentation/git-for-each-ref.txt-upstream
+        public static readonly string Gone = "gone";
         public string Branch { get; set; }
         public string AheadCount { get; set; }
         public string BehindCount { get; set; }
 
         public string ToDisplay()
         {
-            return (!string.IsNullOrEmpty(AheadCount) ? AheadCount + "↑" : string.Empty)
-                   + (!string.IsNullOrEmpty(AheadCount) && !string.IsNullOrEmpty(BehindCount) ? " " : string.Empty)
-                   + (!string.IsNullOrEmpty(BehindCount) ? BehindCount + "↓" : string.Empty);
+            return AheadCount == Gone
+                ? AheadBehindData.Gone
+                : AheadCount == "0" && string.IsNullOrEmpty(BehindCount)
+                ? "0↑↓"
+                : (!string.IsNullOrEmpty(AheadCount) && AheadCount != "0"
+                    ? AheadCount + "↑" + (!string.IsNullOrEmpty(BehindCount) ? " " : string.Empty)
+                    : string.Empty)
+                + (!string.IsNullOrEmpty(BehindCount) ? BehindCount + "↓" : string.Empty);
         }
     }
 }
--- a/GitCommands/Git/AheadBehindDataProvider.cs
+++ b/GitCommands/Git/AheadBehindDataProvider.cs
@@ -17,10 +17,15 @@ namespace GitCommands.Git
     {
         private readonly Func<IExecutable> _getGitExecutable;
 
-        // TODO handle [gone] status to show that remote branch no longer exists
+        // Parse info about remote branches, see below for explanation
+        // This assumes that the Git output is not localised
         private readonly Regex _aheadBehindRegEx =
-            new Regex(@"^(\[(ahead (?<ahead_p>\d+))?(, )?(behind (?<behind_p>\d+))?\])?::(\[(ahead (?<ahead_u>\d+))?(, )?(behind (?<behind_u>\d+))?\])?::(?<branch>.*)$",
-                RegexOptions.Compiled | RegexOptions.Multiline);
+            new Regex(
+                @"^((?<gone_p>gone)|((ahead\s(?<ahead_p>\d+))?(,\s)?(behind\s(?<behind_p>\d+))?)|(?<unk_p>.*?))::
+                   ((?<gone_u>gone)|((ahead\s(?<ahead_u>\d+))?(,\s)?(behind\s(?<behind_u>\d+))?)|(?<unk_u>.*?))::
+                   (?<remote_p>.*?)::(?<remote_u>.*?)::(?<branch>.*)$",
+                RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.IgnorePatternWhitespace | RegexOptions.ExplicitCapture);
+        private readonly string _refFormat = @"%(push:track,nobracket)::%(upstream:track,nobracket)::%(push)::%(upstream)::%(refname:short)";
 
         public AheadBehindDataProvider(Func<IExecutable> getGitExecutable)
         {
@@ -46,14 +51,15 @@ namespace GitCommands.Git
                 throw new ArgumentException(nameof(branchName));
             }
 
-            if (branchName == "(no branch)")
+            if (branchName == DetachedHeadParser.DetachedBranch)
             {
                 return null;
             }
 
             var aheadBehindGitCommand = new GitArgumentBuilder("for-each-ref")
             {
-                "--format=\"%(push:track)::%(upstream:track)::%(refname:short)\"",
+                $"--color=never",
+                $"--format=\"{_refFormat}\"",
                 "refs/heads/" + branchName
             };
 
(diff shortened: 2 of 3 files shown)


GitCommands/Git/AheadBehindData.cs
```diff
@@ -2,15 +2,21 @@
 {
     public struct AheadBehindData
     {
+        public static readonly string Gone = "gone";
```
mstv, line 5: Make translatable or replace with a character, e.g. a dash.

GitCommands/Git/AheadBehindData.cs
```diff
@@ -2,15 +2,21 @@
...
-                   + (!string.IsNullOrEmpty(AheadCount) && !string.IsNullOrEmpty(BehindCount) ? " " : string.Empty)
-                   + (!string.IsNullOrEmpty(BehindCount) ? BehindCount + "↓" : string.Empty);
+            return AheadCount == Gone
+                ? Gone
+                : AheadCount == "0" && string.IsNullOrEmpty(BehindCount)
+                ? "0↑↓"
+                : (!string.IsNullOrEmpty(AheadCount) && AheadCount != "0"
+                    ? AheadCount + "↑" + (!string.IsNullOrEmpty(BehindCount) ? " " : string.Empty)
+                    : string.Empty)
+                    + (!string.IsNullOrEmpty(BehindCount) ? BehindCount + "↓" : string.Empty);
```
mstv, line 19: Replace with a local function using StringBuilder or
```suggestion
                : (!string.IsNullOrEmpty(AheadCount) && AheadCount != "0"
                   ? AheadCount + "↑" + (!string.IsNullOrEmpty(BehindCount) ? " " : string.Empty)
                   : string.Empty)
                  + (!string.IsNullOrEmpty(BehindCount) ? BehindCount + "↓" : string.Empty);
```

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -64,22 +69,39 @@ public AheadBehindDataProvider(Func<IExecutable> getGitExecutable)
...
+                            : match.Groups["ahead_u"].Success
+                            ? match.Groups["ahead_u"].Value
+
+                            // No information about the remote branch, it is gone
+                            : match.Groups["gone_p"].Success || match.Groups["gone_u"].Success
+                            ? AheadBehindData.Gone
+
+                            // A remote exists, but "track" does not display the count if ahead/behind match
+                            : "0",
+
```
mstv, line 103:
```suggestion
                        AheadCount =
                            // Prefer push to upstream for the count
                            match.Groups["ahead_p"].Success
                            ? match.Groups["ahead_p"].Value
                            : match.Groups["ahead_u"].Success
                              ? match.Groups["ahead_u"].Value
                              : // No information about the remote branch, it is gone
                                match.Groups["gone_p"].Success || match.Groups["gone_u"].Success
                                ? AheadBehindData.Gone
                                : // A remote exists, but "track" does not display the count if ahead/behind match
                                  "0",
```

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -64,22 +69,39 @@ public AheadBehindDataProvider(Func<IExecutable> getGitExecutable)
...
-                return null;
-            }
-
             var aheadBehindForBranchesData = new Dictionary<string, AheadBehindData>();
             foreach (Match match in matches)
             {
+                var branch = match.Groups["branch"].Value;
+                var remoteRef = (match.Groups["remote_p"].Success && !string.IsNullOrEmpty(match.Groups["remote_p"].Value))
+                            ? match.Groups["remote_p"].Value
+                            : match.Groups["remote_u"].Value;
```
mstv, line 78:
```suggestion
                var remoteRef = (match.Groups["remote_p"].Success && !string.IsNullOrEmpty(match.Groups["remote_p"].Value))
                                ? match.Groups["remote_p"].Value
                                : match.Groups["remote_u"].Value;
```

mstv (commented):
Looks sensible, haven't checked in detail. the logic is checked by tests.

GitCommands/Git/AheadBehindData.cs
gerhardol: (What I was most unsure about.)
"gone" is what the Git command writes, but I prefer "-" myself too.
(The change is on line 13, the const is still needed.)

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: There must be a newline before a comment....

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: why one more level?

RussKie:
I was going to ask what git uses, but you've already said. If "gone" is
git's lingo I'm ok with it. Though my first reaction was to have "deleted"
or "removed".
"-" to me is akin "n/a", which is probably not really suitable in here. I
think this may be appropriate for "not pushed" case....

GitCommands/Git/AheadBehindData.cs
gerhardol: Changing the indentation

GitCommands/Git/AheadBehindData.cs
gerhardol: > I was going to ask what git uses, but you've already said. If "gone" is
git's lingo I'm ok with it. Though my first reaction was to have "deleted"
or "removed".
"-" to me is akin "n/a", which is probably not really suitable in here. I
think this may be appropriate for "not pushed" case....

@RussKie 
Changing to "-" right now
Should we have this translatable if sticking with "gone"?

For branches that have lost the tracking (second part of #8553), should we reserve something?
(I want to display this in FormPush at least)

RussKie: If it's a git's string I think we can have it as is. Need to have a comment in the code why it is not translated though.

GitCommands/Git/AheadBehindDataProvider.cs
RussKie: Right now it is not very readable... You can suppress the warning using #pragma

mstv: > There must be a newline before a comment....

Or use `////`.

GitCommands/Git/AheadBehindDataProvider.cs
mstv: I'd either `?` align it with the start of the condition or else use one indent level.

GitCommands/Git/AheadBehindData.cs
gerhardol: for-each-ref is a plumbing command and should not be translated, otherwise also ahead/behind could be incorrect.
I added one more check (made the parsing even more complicated) to handle garbage, now it could be interpreted as "0" changes.

I checked the Git code and 'gone' as well as 'ahead'/'behind' can be translated in other printouts. I tried to get command cli to be presented in something else than en, but that failed (Just "LANGUAGE=de_DE git status" should be enough). Maybe not included with git-for-windows? So maybe it should be translated later. But the translated "expression" can be harder to understand, git gone yields some Google hits.

Not sure if the comment is good though...

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: I used #pragma instead of misusing ////
Even more conditions now, I prefer to not right shift (should be 4 spaces if done)

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: One indent from 'var'

gerhardol:
The parsing/interpretation of the Git output is a little too complicated right now
Adding properties for gone and zero changes will simplify, or maybe set with fewer conditions and then correct after (if zero changes, overwrite afterwards).

gerhardol:
> The parsing/interpretation of the Git output is a little too complicated right now

Rewriting will not really improve, I suggest it is kept as is with comments

Squashed and rebased without changes

RussKie (approved):

PR #5525, opened 2018-10-05
Remove currents repository and branch from recent lists

pmiossec:
Changes proposed in this pull request:
- Remove current repository from recent lists
- Remove current branch from list of branches

because that make no sense to select it (and it clutter the GUI as it is the first in the list)
 
Screenshots before and after (if PR changes UI):
- The current repository is no more in the list (on top)
image

- The current branch is no more in the list (on top)
image

Has been tested on (remove any that don't apply):
- GIT 2.17
- Windows 10

Issue #4952, opened 2018-05-15
Fluent UI "Ribbon" interface 


gerhardol:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
GitExtensions uses an old style WinForms interface

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
A refreshed interface based on Fluent UI aka Ribbon

GE 2.x supported Mono and had many design limitations. GE in master (to be released as 3.0) breaks that compatibilitity and has many changes under the hood as well as some new elements (left panel, redesigned dashboard for instance). However, it would be nice if at least Browse got a new shiny look.

There are a number of discussions about the interfaces already, they to large extent discuss Mono support or not, not so much what should be done.

If you are interested, comment to this issue or open a new.
(I am not an UI designer myself and will not take a lead.)

**Environment you encounter the issue:**

- GitExtensions version: master
- OS version: >= Win7

**Did this work in previous version of GitExtensions (which)?**
no

vbjay:
I've liked the devexpress ribon you. Could give them a try.

drewnoakes:
I'm not familiar with any ribbon controls. Personally I'm not a huge fan of ribbons, though I'll admit I haven't used them very much. Something about how items move around seems constantly surprising to me.

There's a WPF option: https://github.com/fluentribbon/Fluent.Ribbon/wiki/Screenshots

While we don't currently do so, it's possible (and in my experience easy and stable) to host WPF controls in WinForms and vice-versa. Something to consider.

Quppa:
It's a shame that there seems to be no wrapper around the [native control](https://msdn.microsoft.com/en-us/library/windows/desktop/dd316910(v=vs.85).aspx), since that would be my first preference.

I don't have any strong opinions about which third-party control to pick.

[...]

PR #8572, opened 2020-10-24
Break the toolstrip

RussKie:
Relates to #5525

## Proposed changes

- Break the main toolbar into two - general operations and filters. This is arbitrary, filters were the easiest to break out. The main point of the PR was to attempt the split and see how well it works. This then maybe used as a foundation for further refactoring of the toolstrip area, such as moving user scripts to own toolstrip, view-related, repo-related... Possibilities are endless.
- Save/restore positions of the toolbars

### Before

image

### After

image

image

image

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
@@ -44,6 +44,8 @@ namespace GitUI.CommandsDialogs
 
             CreateMenuItems();
             Translate();
+
+            _toolStripContextMenu.Opening += (s, e) => RefreshToolbarsMenuItemCheckedState(_toolStripContextMenu.Items);
         }
 
         public void Dispose()
@@ -89,7 +91,7 @@ namespace GitUI.CommandsDialogs
 
             foreach (ToolStrip toolStrip in toolStrips)
             {
-                Debug.Assert(!string.IsNullOrEmpty(toolStrip.Text), "Toolstrip must specify its name via Text property");
+                Debug.Assert(!string.IsNullOrEmpty(toolStrip.Text), "Toolstrip must specify its name via Text property.");
 
                 _toolStripContextMenu.Items.Add(CreateItem(toolStrip));
                 _toolbarsMenuItem.DropDownItems.Add(CreateItem(toolStrip));
@@ -100,9 +102,10 @@ namespace GitUI.CommandsDialogs
                 var toolStripItem = new ToolStripMenuItem(senderToolStrip.Text)
                 {
                     Checked = senderToolStrip.Visible,
-                    CheckOnClick = true
+                    CheckOnClick = true,
+                    Tag = senderToolStrip
                 };
-                toolStripItem.CheckedChanged += (s, e) =>
+                toolStripItem.Click += (s, e) =>
                 {
                     senderToolStrip.Visible = !senderToolStrip.Visible;
                 };
@@ -223,6 +226,7 @@ namespace GitUI.CommandsDialogs
                     Name = "toolbarsMenuItem",
                     Text = "Toolbars",
                 };
+                _toolbarsMenuItem.DropDownOpening += (s, e) => RefreshToolbarsMenuItemCheckedState(_toolbarsMenuItem.DropDownItems);
             }
         }
 
@@ -298,6 +302,15 @@ namespace GitUI.CommandsDialogs
             else
             {
                 throw new ApplicationException("this case is not allowed");
+            }
+        }
+
+        private void RefreshToolbarsMenuItemCheckedState(ToolStripItemCollection toolStripItems)
+        {
+            foreach (ToolStripMenuItem item in toolStripItems)
+            {
+                Debug.Assert(item.Tag is ToolStrip, "Toolbars context menu items must reference Toolstrips via Tag property.");
+                item.Checked = ((ToolStrip)item.Tag).Visible;
             }
         }
     }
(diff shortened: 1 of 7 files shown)


NikolayXHD (approved):

gerhardol:
> Relates to #5525

Is this the correct reference?

I like the change, but I rather like to remove/add specific buttons:
https://github.com/gitextensions/gitextensions/issues/4952#issuecomment-715327377 (or some other issue)

I hardly use the filters, but most buttons in the (new) main toolstrip (except toggle panel and hardly branches (but submodule "up").
However, the filters do not occupy any used space for me, they are just hidden in windowed mode.

The code looks good and works as expected, but I believe that was not what you asked for.

Will continue discussion in #4952

ivangrek:
👍 
How can I hide / show the panel?

ivangrek (approved):

RussKie:
> > Relates to #5525
> 
> Is this the correct reference?

Yes, there was a discussion pertaining to breaking the main toolbar into different ones (e.g. https://github.com/gitextensions/gitextensions/pull/5525#issuecomment-711440403).

> I like the change, but I rather like to remove/add specific buttons:
> [#4952 (comment)](https://github.com/gitextensions/gitextensions/issues/4952#issuecomment-715327377) (or some other issue)

This one is a good one too, 👍 

> The code looks good and works as expected, but I believe that was not what you asked for.

This is the first step to allow users customise the toolbar - either via a coarser solution of showing/hiding toolbars, or via a fine-grainer one of showing/hiding individual buttons.
This is something has been on my mind, for at least 3 years now judging by commits in https://github.com/RussKie/gitextensions/tree/rework_toolbars

gerhardol:
> This is the first step to allow users customise the toolbar - either via a coarser solution of showing/hiding toolbars, or via a fine-grainer one of showing/hiding individual buttons.

I am not opposing the change in any way. It also requires a way to hide/show the toolbars to be really useful, as well as splitting up the toolbars further. 
However, if someone puts time on this, I rather see that the effort is on the individual buttons rather than the toolbars.

ivangrek:
I see several strategies here.
One panel and customizable buttons.
Multiple panels and the ability to disable a panel.
Both options.

ivangrek:
Plugins could also register their buttons.

RussKie:
Since it bothered me, and you asked for it - there you go:
image
image

NB: `FormFileHistory` is unaffected, though it uses the same `FormBrowseMenus` API.

RussKie:
Btw, the toolbars settings are stored in a local user config located somewhere like this: `C:\Users\<user>\AppData\Local\Git_Extensions\GitExtensions.exe_Url_<some crazy string>\33.33.33.0\user.config`

drewnoakes:
If you arrange the toolbars on the same row, and the leftmost one reduces its width, does the right one stay still or move to the left?

Or in other words, if you don't want to give up any vertical screen height and you want both toolbars, will this make the experience different/worse?

RussKie:
Thank you @ivangrek, can you please explain the nature of your changes? What did I miss?

Also your changes introduce a number of perf regressions - LINQ is significantly more expensive and requires more allocations ([for example](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyAMABMugHQAyYAdgI4DcOOyAzEagQMIEDeOBfRLZABYCAWQAUASm69+cgPYAjAFYBTAMbAA2gF0CYYKoC2AZwIBeApVUB3AkrWatAVh31scubM8EAZvIAnVQBDdQALAnEAN2CA+31KfUNTaR4PH09iEgARMGCAc0p5E2AwdRMc1UUAV3ySAEETE1UA4HF5AgBCS0pqiAhJWgzPb2GOeUoTeQhVEgB1AINVdsHRuQBfUc30vmYiEVFUKRkdzwcNbT0l0wsrW3sVC5c3NbG+V74YuIgqalvripzMItZZgCwAPn0XR6fQGJAAyqoZppxGDzJCwJISAAVeT1AIBYIATyk7g26yAA=)).

RussKie:
> If you arrange the toolbars on the same row, and the leftmost one reduces its width, does the right one stay still or move to the left?

This would be the default behaviour of Windows Forms. If toolbars are sequentially positioned (i.e. no gaps) they remain together when the left one contracts or expands - this is easy to verify by switching repositories (the repositories toolbar dropdown changes width).

. . .

I'm currently unhappy with the visual tearing when the toolbars are refreshed, something is missing somewhere...

RussKie:
What I see as immediate low hanging fruits from here 
- breaking user scripts into own toolbar, and 
- adding the following items to the 'Filters' toobar:
    * Show all / current branch only / filtered - either as 3 different buttons or a dropdown list (requires POC to see what works best for UI/UX)
    * Show reflogs
    * ❔ Highlight the current branch - whilst not filtering, I find myself using it often, going to the menu feels distracting.

ivangrek:
> Also your changes introduce a number of perf regressions - LINQ is significantly more expensive and requires more allocations (for example).

needs to choice

ivangrek:
> If you arrange the toolbars on the same row, and the leftmost one reduces its width, does the right one stay still or move to the left?

Would be nice to "float left"

ivangrek:
image

RussKie:
> image

This is how a user positioned the toolbars, the runtime obeys the user's choice.
We could maybe add "Reset" option to reset back to the default, but this is outside the scope of this change.

ivangrek:
Just VS UX.
We could create new issue when ok.

pmiossec:
@RussKie 

> Save/restore positions of the toolbars

I see in the description of the PR that the positions of the toolbars are saved and restored. Should it be the case or is it in fact not yet implemented?

Because I always end up with this configuration (that I hate):
image

Only hidding of toolbar is saved for me.
Tested with custom version, version build from HEAD and the portable version from AppVeyor.

Tell me and I will open an issue if that's already implemented...

PR #8564, opened 2020-10-21
Implement custom avatar providers and refactor avatar handling.

wischi-chr:
## Proposed changes
- Implement a template based custom avatar provider (as proposed in #4226)
- Refactor current avatar handling according to single responsibility principle
- PR might close #4226 _(or at least large parts of it)_

### Before
image

### After
image
image

## Test methodology
- Implemented changes in a (unit) testable manner
- Tried to keep it as simple and maintainable as possible
- Tried to keep a lot of the existing (already field tested) code-base
- Wrote unit tests for relevant cases to ensure functionality and prevent later regressions
- Used LiveUnitTesting, OpenCover and Stryker.NET to check for missing test cases and implemented them if applicable
- Manually tested a lot of different provider combinations

## Test environment(s) 

- GIT 2.29.0.windows.1
- Windows 10

# Update 0 - Original proposal in detail

### Translations
First things first, let's start with the changes I'm least confident about - Translations - and should probably be double-checked if I got it right.

I read the wiki article about it, but I'm not sure if I implemented the UI parts correctly to be translatable, namely the `DescriptionAttribute` in `GitCommands.AvatarProvider` for the new `Custom` Option and UI changes in `AppearanceSettingsPage.cs`.

Is there something else I have to do? Add the keys to some resource file or something?

### UI Updates
For some reason the Visual Studio Forms Designer kept messing up everything inside the Designer.cs files (reordered controls in code, changed coordinates, etc.) - Because I didn't wanna mess up everything and to keep the differences clean for this PR , I wrote the UI changes manually by adding the controls in code.

Is that how you do this as well? Or is there some kind of trick to prevent the Designer from rewriting everything the moment the form/control is opened?

### User preferences
There are two new avatar providers `GitHub` and `Custom` (more on that later) and a new AppSetting `CustomAvatarTemplate` used to persist (as the name suggests) the template for custom avatar providers (more about that, including syntax, later).

#### Split IAvatarProvider into two interfaces
Currently all `IAvatarProvider` have to know something about caching and implement the caching stuff even though most providers shouldn't know anything about caching - and the current code also reflects that. Invoking `CacheCleared` in `AvatarDownloader.cs` without doing anything (because there is no cache) and passing through calls to inner providers etc.

The new interfaces are `IAvatarProvider` - which knows nothing about caching at all, and `ICachedAvatarProvider` which is basically the old version of `IAvatarProvider`.

#### Refactor AvatarDownloader.cs
[...]

PR #8574, opened 2020-10-24
Cleanup the plugin interface

ivangrek:
Dependent on:

#8564
#8278
#8628

## Proposed changes

- Cleanup the plugin interface
- Split into multiple simple interfaces
- Eliminate dependence on translation infrastructure

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

----


ivangrek:
Currently only one plugin has changed

ivangrek:
The purpose is not entirely clear for Register and Unregister.
It is similar to loading / unloading events.
I would like to rename these methods.

gerhardol (commented):
No deep opinion about plugin interfaces. I do not see anything that I have an opinion about, compatibility seem to be maintained with some quirks not fun but OK to handle. If the improvement is big enough to motivate a change is another question.

ivangrek:
@gerhardol, thanks for comment.

The ultimate goal is to have a simple interface. Infrastructure agnostic.

`Big Things Start Small.`

Plugins/GitUIPluginInterfaces/IGitPlugin.cs
```diff
@@ -28,4 +28,34 @@ public interface IGitPlugin
         /// </summary>
         bool Execute(GitUIEventArgs args);
     }
-}
\ No newline at end of file
+
+    public interface IGitPluginV2
```
mstv, line 32: How many plugins are maintained independently of the GE repo? Could we drop supporting the initial interface?
I guess most of them are still there. (I'll celebrate the day they will be removed from the standard build.)

Plugins/GitUIPluginInterfaces/IGitPlugin.cs
```diff
@@ -28,4 +28,34 @@ public interface IGitPlugin
...
+
+    public interface IGitPluginRegistrable
+    {
+        void Register(IGitUICommands gitUiCommands, IGitPluginSettingsContainer settingsContainer);
+    }
+
+    public interface IGitPluginUnregistrable
+    {
+        void Unregister();
+    }
```
mstv, line 55: This pair of functions should be in one interface, shouldn't it?

> The purpose is not entirely clear for Register and Unregister.
It is similar to loading / unloading events.
I would like to rename these methods.

The interfaces should have a documentation as a basic principle.
These functions are called by the Plugin**Registry**. Though "Load / Unload" might be more clear anyway.

GitUI/CommandsDialogs/SettingsDialog/Plugins/PluginSettingsPage.cs
```diff
@@ -48,24 +65,39 @@ protected override ISettingsSource GetCurrentSettings()
...
+            if (_gitPlugin == null && _gitPluginV2 == null)
             {
                 throw new ApplicationException();
             }
 
-            return _gitPlugin.HasSettings ? _gitPlugin.GetSettings() : Array.Empty<ISetting>();
+            if (_gitPlugin != null)
+            {
+                return _gitPlugin.HasSettings ? _gitPlugin.GetSettings() : Array.Empty<ISetting>();
+            }
```
mstv, line 83:
```suggestion
            if (_gitPlugin?.HasSettings == true)
            {
                return _gitPlugin.GetSettings();
            }
```

pmiossec:
What I want to see in the plugin interface if that should be redesigned is the version of gitextensions it is built against. The best would be that this information is filled automatically at build time.

That way, we will be able to check if the plugin is built for the current version and if needed disable it if there are problems or by default , propose to enable it or best, propose to see if there is an update that better suit the current version... 

But I should admit that won't solve the case where the plugin doesn't load and you won't be able to check this version number...

Plugins/GitUIPluginInterfaces/IGitPlugin.cs
ivangrek: > This pair of functions should be in one interface, shouldn't it?

Not all plugins use both of them.

ivangrek:
Could anyone comment on why null is installed here?

GitPluginBase.cs

```
public virtual void Unregister(IGitUICommands gitUiCommands)
{
    SettingsContainer.SetSettingsSource(null);
}
```

GitPluginSettingsContainer.cs

```
public void SetSettingsSource(ISettingsSource settingsSource)
{
    _settingsSource = settingsSource;
}

private ISettingsSource ExternalSettings => _settingsSource ?? AppSettings.SettingsContainer;

public override T GetValue<T>(string name, T defaultValue, Func<string, T> decode)
{
    return ExternalSettings.GetValue(_pluginName + name, defaultValue, decode);
}
```

Should the plugin still work with the AppSettings.Settings container after unregistering?

RussKie:
> * Cleanup the plugin interface
> * Split into multiple simple interfaces

Thank you, but could you please provide more reasoning behind this, and what we're achieving. E.g. why do we need more interfaces? And how does it impact plugin development?



> * Eliminate dependence on translation infrastructure

This is probably a good thing, though we're yet to explore and understand what it means to localise a plugin, and how it would integrate into the app.

ivangrek:
Thanks for the comments.

> E.g. why do we need more interfaces?

More interfaces will provide the ability to use only what the plugin needs. 

> And how does it impact plugin development?

This will also make it easy to extend the API in the future by adding a new interface. This is in general terms.

> This is probably a good thing, though we're yet to explore and understand what it means to localise a plugin, and how it would integrate into the app.

I agree. At first glance, after the plugins are moved, the translation should be done by the plugin itself.

These are pilot changes for the opportunity to start a discussion. I plan to provide more changes in this direction.

RussKie:
Right now I am not convinced there is a pressing need for more interfaces.
We've lived with the current model for the past 7+ years, and I don't
believe we have felt any constraints so far.
I would think that moving the common extensibility points into the
extensibility project is more important at this stage.

ivangrek:
image


It remains to resolve:

ResourceManager
---
    GitPluginBase
    GitExtensionsFormBase
    TranslationString

GitExtUtils
---
    ClipboardUtil
    GitArgumentBuilder
    ThreadHelper
    DpiUtil

Dependencies
---

ResourceManager to remove.

image

It is not yet clear what to do with the base form. Translation is also used here.

image

Next
---

Change GitPluginBase to IGitPlugin.
Remove GitPluginBase.
Change translations to local resources like Strings.en-US.resx.
Remove TranslationString from plugins.
Move utils classes to GitExtensions.Core.Utils.
...
Remove ResourceManager from references.

ivangrek:
> It is not yet clear what to do with the base form. Translation is also used here.

I think we can use the standard form.
Thoughts?

ivangrek:
ReleaseNotesGenerator plugin ready to move.
Please review.

GitUI/CommandsDialogs/SettingsDialog/Plugins/PluginSettingsPage.cs
```diff
@@ -59,14 +60,14 @@ private IEnumerable<ISetting> GetSettings()
...
-            return _gitPlugin.HasSettings ? _gitPlugin.GetSettings() : Array.Empty<ISetting>();
+            return _gitPlugin.GetSettings() ?? Array.Empty<ISetting>();
         }
 
         public override SettingsPageReference PageReference => new SettingsPageReferenceByType(_gitPlugin.GetType());
 
         protected override ISettingsLayout CreateSettingsLayout()
         {
-            labelNoSettings.Visible = !_gitPlugin.HasSettings;
+            labelNoSettings.Visible = !_gitPlugin.GetSettings()?.Any() ?? false;
```
RussKie: Loading settings maybe an expensive exercise (e.g. https://github.com/gitextensions/gitextensions/issues/8339), this can cause an observable perf regressions. 
This is one of reasons why we added a flag - a plugin author is responsible for indicating whether a plugin has settings or not.

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -73,7 +73,10 @@ public static void Register(IGitUICommands gitUiCommands)
 
             lock (Plugins)
             {
-                Plugins.ForEach(p => p.Register(gitUiCommands));
+                Plugins.ForEach(x => x.SettingsContainer.SetSettingsSource(gitUiCommands.GitModule.GetEffectiveSettings()));
+
+                Plugins.OfType<ILoadHandler>()
+                    .ForEach(x => x.OnLoad(gitUiCommands));
```
RussKie: We enumerating the collection multiple times, as well as performing casts. This is inefficient.

GitExtensions.Extensibility/GitExtensions.Extensibility.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
RussKie, line 1: We already have the https://github.com/gitextensions/gitextensions.extensibility, and should not be creating another project in the main solution. That project should contain all base interfaces and extensibility points.
The main solution as well as all Git Extensions plugins must depend on that project and install it as a NuGet dependency.

GitExtensions.Core/GitExtensions.Core.csproj
```diff
@@ -0,0 +1,5 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
RussKie, line 1: Why do we need this project?
Any new projects within the solution must be added to all the installer scripts. This is an overhead which needs a good justification.

GitUI/Plugin/PluginRegistry.cs
```diff
@@ -86,9 +95,63 @@ public static void Unregister(IGitUICommands gitUiCommands)
...
             PluginsRegistered = false;
         }
+
+        #region Events
+
+        private static void OnPostCommit(object sender, GitUIPostActionEventArgs e)
+        {
+            Plugins.OfType<IPostCommitHandler>()
+                .ForEach(x => x.OnPostCommit(e));
+        }
```
RussKie, line 130: I still can't see a good reason to introduce a bunch of new interfaces. 
I can certainly see benefits in removing subscriptions from plugin implementations, and allowing them to plug into the extensions point via overrides, but we don't need to define multiple interfaces for this.

```suggestion
        private static void OnPostCommit(object sender, GitUIPostActionEventArgs e)
        {
            foreach (IPlugin plugin in Plugins)
                plugin.OnPostCommit(e);
        }
```

RussKie (commented):
We need to work more in this.

GitUI/CommandsDialogs/SettingsDialog/Plugins/PluginSettingsPage.cs
ivangrek: Last code is

`labelNoSettings.Visible = !(_gitPlugin is IGitPluginConfigurable);`

GitExtensions.Extensibility/GitExtensions.Extensibility.csproj
ivangrek: This project to migrate to https://github.com/gitextensions/gitextensions.extensibility.
Do you think this should be done right away?

GitExtensions.Core/GitExtensions.Core.csproj
ivangrek: This project is for basic interfaces and application mechanisms.
This does not apply directly to plugins.
But we can place this project in https://github.com/gitextensions/gitextensions.extensibility too.
It seems to me best not to mix this directly with the plugin interfaces.

GitUI/Plugin/PluginRegistry.cs
ivangrek: Plugins list is of type List<IGitPlugin>. Collection in memory.
I don’t think that going through the collection several times in memory at this point is critical, we can optimize it if necessary, sacrificing the readability.
Are you suggesting replacing this with "foreach" and "is"?

GitUI/Plugin/PluginRegistry.cs
ivangrek: - SOLID
- removing subscriptions from plugin and easier handling
- less opportunity to make a mistake
- no need to have an empty implementation for the method used

First version of implementation

image

And then we add some new feature

image

The first plugin works as it did.
The second implements the new function.
The main code hasn't changed.

ivangrek:
CreateLocalBranches plugin ready to move.

ivangrek:
Could we do a pre-release package for https://github.com/gitextensions/gitextensions.extensibility?
I can upload projects to this repository, but I don't know how to make a package?

mast-eu:
> Could we do a pre-release package for https://github.com/gitextensions/gitextensions.extensibility?

Yes, this can be done. Actually, I even think it is necessary to finalize this PR.

> I can upload projects to this repository, but I don't know how to make a package?

The release process is described here: https://github.com/gitextensions/gitextensions.pluginmanager/wiki/How-to:-Prepare-a-release
It is very similar for all plugin-related repos, except for the signing part (which is, at the moment, specific for the Plugin Manager).

> And also where to upload plugins?

They go to nuget.org. We have an organisation account over there. In particular, the repo https://github.com/gitextensions/gitextensions.extensibility is linked to https://www.nuget.org/packages/GitExtensions.Extensibility/

ivangrek:
@mast-eu, thanks.

> They go to nuget.org. We have an organisation account over there. In particular, the repo https://github.com/gitextensions/gitextensions.extensibility is linked to https://www.nuget.org/packages/GitExtensions.Extensibility/

I mean a plugin project like GitExtensions.ReleaseNotesGenerator

GitUI/Plugin/PluginRegistry.cs
RussKie: > * SOLID

I am a proponent of SOLID, but I don't see how it applies here. This is an extension point by which the main app interacts with plugins. SOLID is not just SR, it is other things that are completely inappropriate here. The main app isn't loading different plugins based on interfaces, it loads all plugins.
The extensibility surface includes operations "register" and "unregister", as well other events that plugins may be interested in reacting to. It is up to each plugin author to plug into which ever flow they feel necessary.

Multiple types (i.e. interfaces) have associated costs to them - both at the runtime level (types need to loaded into memory, type resolution, etc.), and more importantly at author level - instead of dealing with a single type which has few defined and documented extension points, an plugin author must _know_ to implement specific implementations. There's too much overhead both from the project to maintain, and from an author to deal with.

RussKie:
> Could we do a pre-release package for [gitextensions/gitextensions.extensibility](https://github.com/gitextensions/gitextensions.extensibility)?
> I can upload projects to this repository, but I don't know how to make a package?
> And also where to upload plugins?

Ugh... It's been many months since I looked at this.
Looking at the repo, we publish a NuGet package on the AppVeyor ([example](https://ci.appveyor.com/project/gitextensions/gitextensions-extensibility/build/artifacts)), but we don't push to the NuGet feed (yet).
I think at this stage we're uploading packages manually. @mast-eu is this correct?

For a local development you can generate a package and restore it from a [local feed](https://medium.com/@churi.vibhav/creating-and-using-a-local-nuget-package-repository-9f19475d6af8), and when you think it is all done - push changes, build, and publish the package.

GitExtensions.Core/GitExtensions.Core.csproj
RussKie: > This project is for basic interfaces and application mechanisms.

If this is app-specific, let's hold off creating new projects, and keep it in on of the existing assemblies. Once we switch to .NET we can create new projects, I expect our deployment story will change dramatically with the migration.

GitExtensions.Extensibility/GitExtensions.Extensibility.csproj
RussKie: I'd rather move it, then create noise in the history.

GitUI/Plugin/PluginRegistry.cs
RussKie: We should iterate in one go:
```
foreach (IPlugin plugin in Plugins)
{
    plugin.SettingsContainer.SetSettingsSource(gitUiCommands.GitModule.GetEffectiveSettings());

    plugin.OnLoad(gitUiCommands);

    // ...
}
```

GitUI/Plugin/PluginRegistry.cs
ivangrek: > I am a proponent of SOLID, but I don't see how it applies here.

I agree.

> SOLID is not just SR, it is other things that are completely inappropriate here.

I agree. Could you please clarify what specific things are you talking about?

> Multiple types (i.e. interfaces) have associated costs to them - both at the runtime level (types need to loaded into memory, type resolution, etc.), and more importantly at author level - instead of dealing with a single type which has few defined and documented extension points, an plugin author must know to implement specific implementations. There's too much overhead both from the project to maintain, and from an author to deal with.

I agree. Magic, but I think plugins will run faster.

ivangrek:
> we can use gitextensions/gitextensions.extensibility as a submodule

Can we do this now in a separate PR?

GitExtensions.Core/GitExtensions.Core.csproj
ivangrek: These projects are migration only and will be removed from the main repository.

ivangrek:
image

Will be ok?

mast-eu:
> I mean a plugin project like GitExtensions.ReleaseNotesGenerator

Once done, there will be a newly created repo https://github.com/gitextensions/GitExtensions.ReleaseNotesGenerator. It should contain the extracted history from the main repo (see https://github.com/gitextensions/GitExtensions.GerritPlugin/pull/1 for an example).
While working on it, in order to simplify work, you might as well create the repo in your own user space and we'll transfer it later.

> but we don't push to the NuGet feed (yet).
> I think at this stage we're uploading packages manually. @mast-eu is this correct?

Yes, that's correct.

Plugins/GitUIPluginInterfaces/IGitPlugin.cs
ivangrek: > The interfaces should have a documentation as a basic principle.

Added documentation for interfaces.

GitUI/Plugin/PluginRegistry.cs
ivangrek: Changed to use "foreach".

RussKie:
Wow, a massive changeset. 

I've looked through it and I can see three distinct themes here:
- breaking up the dependency chains,
- refactor (or better to say, a rewrite) of `IPlugin`
- break the plugins away.

In general I'm supportive of the changes overall, but as previously noted I am not convinced of the necessity to break down the `IPugin` interface. 
Also the removal of `HasSettings` property will have a negative impact on performance, and it needs to be reinstated.

Another aspect that we'll need to figure out is plugins translation (i.e. procedures, CI/CD, etc.) and loading of their translations.

@gitextensions/git-extensions-team I'd want to have a release before we accept this change, as it will likely cause various issues, and we'll need a runway to sort them out.

pmiossec:
>Also the removal of HasSettings property will have a negative impact on performance, and it needs to be reinstated.

In fact the goal of this property was to avoid the call to the method `GetSettings()` multiple times when displaying the plugin settings popup. Due to how the settings controls are created and disposed, it happened, in some cases, that some exceptions were thrown because the control was already disposed when trying to display it.

For that reason, the controls must be displayed before they were displayed and for that with the result of the first time call to `GetSettings()`. The property was created for that and from what I see, the refactoring, even without the property kept the same behavior.

So for me, from that point of view, that's ok...

ivangrek:
Also "Any" only triggers the first "yield".

image

ivangrek:
First approximation changes are ready.
Plugins ready to 🚀 to the 🌔.
😃

ivangrek:
> It should contain the extracted history from the main repo (see gitextensions/GitExtensions.GerritPlugin#1 for an example).

I tried to extract history, but I only see the latest changes.

git filter-branch --prune-empty --force --subdirectory-filter GitExtensions.Core GitExtensions_Core
or
git subtree split --prefix=GitExtensions.Core --annotate="(split from main repo) " --branch GitExtensions_Core

https://github.com/ivangrek/gitextensions.extensibility/tree/GitExtensions_Core

RussKie:
Some plugins may be doing IO bound operations as part of GetSettings call,
so this will have a negative effect.

The proposed design makes this risk significantly more material.
We had to deal with this issue in the past already.

PR #8550, opened 2020-10-17
gray additional commit lines in revision grid

NikolayXHD:
## Proposed changes
As [promised](https://github.com/gitextensions/gitextensions/pull/8371#issuecomment-694135580)

- Render commit body using gray text color in revision graph
to separate it from commit subject

- Apply draw non relatives text gray appearance setting to selected row
since now we have a ready-to use mechanism to produce a gray text color for highlighted text

- Split commit message rendering in Revision Grid to smaller methods

## Screenshots

unselected row
image

selected row
image

draw non-relative text gray in selected row
image

## Test environment(s)
Tested manually in Windows10
----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8576, opened 2020-10-25
fix background / foreground contrast for gray text

NikolayXHD:
Fix too low contrast between commit body and background in dark theme.

Problem was introduced in #8550 and [pointed out](https://github.com/gitextensions/gitextensions/pull/8550#issuecomment-716208279) by @gerhardol

## Before fix

image

## Screenshot after fix:

image


RussKie (approved):

RussKie:
Can you please check the other official themes as well?
If all good - merge at will.

gerhardol:
Looks good
There is now very little contrast between headline and body, not much that can be done about it

image

image

gerhardol (approved):

PR #8593, opened 2020-11-04
FormGoToCommit: set focus on commit textbox when no control has focus

### LINKED ISSUE #8591: Can't find commit with entire hash
## Current behaviour

"Go to commit..." dialog doesn't find a commit when the entire hash is used (works when one char is removed)

## Expected behaviour

The entire hash could be used to find a commit.

## Steps to reproduce

1. Copy hash of a commit
2. Select another commit
3. Ctrl+shift+G (the popup open with the hash already filled)
4. Click "Go" button
5. "No revision found" message box

## Screenshots

bug_go_to_commit

Yes.

## Environment

- Git Extensions 33.33.33
- Build bc261f585a6a67b9b3c2340ff867553ec1fa437e (custom build from ` 0b5ecd2a5fe53fd5b76209e8aea7465a90fa2357` from recent `master` )
- Git 2.28.0.windows.1
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

## Diagnostics

Don't know

pmiossec:
Fixes #8591 

## Proposed changes

- When opening the form, the focus is on the commit textbox and the text is selected if textbox if filled with the clipboard content.

Perhaps not the best way to fix but it works...

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7a52bdf65744b5d760d9edb9afc9a1863b88d2e7
- Git 2.28.0.windows.1
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormGoToCommit.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormGoToCommit.cs
@@ -131,6 +131,10 @@ namespace GitUI.CommandsDialogs.BrowseDialog
             {
                 _selectedRevision = _selectedBranch != null ? _selectedBranch.Guid : "";
             }
+            else
+            {
+                textboxCommitExpression.Focus();
+            }
         }
 
         private void comboBoxTags_TextChanged(object sender, EventArgs e)


RussKie:
Sorry, how is this fixing #8591?

pmiossec:
>Sorry, how is this fixing #8591?

This form is a little strange but it is the focus on one of the 3 input fields that determine which ref-spec value (commit, tag or branch) that you will use to find the commit. You could see that here:

https://github.com/gitextensions/gitextensions/blob/7a52bdf65744b5d760d9edb9afc9a1863b88d2e7/GitUI/CommandsDialogs/BrowseDialog/FormGoToCommit.cs#L122

But because there is a bad behavior when loading the form (that I didn't exactly find out which one) and so the focus is on none of the inputs. So when you click "go", it search for the commit `null` and so display the error message.

My change ensure that the commit field has the focus, so the hash value is set and used and so it found the commit and select it \o/

RussKie (approved):

PR #8604, opened 2020-11-08
RevDiff context menu for cherry-pick

### LINKED ISSUE #8588: Cannot cherry pick the creation of a file
I'm not even sure if this is supposed to work, or if there's some logical reason it can't.

I've noticed that I cannot cherry pick the creation of a file. In other words, if there's a new file in another branch, I cannot add that file to my uncommitted local changes by cherry picking it. The context menu option is available, and there's no error when it's used, but the file is not created.

If this is not supposed to work, then take this as a suggestion to disable the "cherry pick" context menu item on created files.

Git 2.28.0.windows.1, GitExtensions 3.4.3.9999.

gerhardol:
Fixes #8588 

## Proposed changes

- Do not show cherry-pick in the file status list if the file is added or if the file does not exist 
Note that the FileViewer check for support patches is quite incomplete and fails for instance for 8a8985d78edb997d094ad9642220a8457b2c7f9b GitUI\Editor\Diff\RangeDiffHighlightService.cs
This is fixed in #7825 (that is blocked for other reasons.) Will not be handled here.

- Do not show _Show in File tree_ if the file has been deleted (as it will not exist in the tree)

### Before

image

image

### After

image

image

## Test methodology 

Tests updated

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -335,6 +335,7 @@ namespace GitUI.CommandsDialogs
             bool isAnyTracked = selectedItems.Any(item => item.Item.IsTracked);
             bool isAnyIndex = selectedItems.Any(item => item.Item.Staged == StagedStatus.Index);
             bool isAnyWorkTree = selectedItems.Any(item => item.Item.Staged == StagedStatus.WorkTree);
+            bool isDeleted = selectedItems.Any(item => item.Item.IsDeleted);
             bool isAnySubmodule = selectedItems.Any(item => item.Item.IsSubmodule);
             (bool allFilesExist, bool allFilesOrUntrackedDirectoriesExist) = FileOrUntrackedDirExists(selectedItems, _fullPathResolver);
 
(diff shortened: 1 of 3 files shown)


RussKie (approved):

gerhardol:
@msftbot merge in 1 day

mstv:
Why is `Show in folder` not disabled in your screenshot with the deleted file?
It is disabled (for me) with the portable build (with commit 19f3a34).

Why shouldn't we keep the previous functionality?
grafik
grafik

gerhardol:
> Why is `Show in folder` not disabled in your screenshot with the deleted file?
> It is disabled (for me) with the portable build as well as with previous builds (tested with the `Diff` of commit [19f3a34](https://github.com/gitextensions/gitextensions/commit/19f3a34e862fe43bc55c85672146147afe73e766)).

Because the the file compared exists in the file system (I compared using the child commit as first and parent as second commit).

> Why shouldn't we keep the previous functionality?

We could, but this is similar to the issue: A menu item that will not have any effect

mstv:
> We could, but this is similar to the issue: A menu item that will not have any effect

As the message says, it selects the parent folder of the deleted file.
I don't have a strong opinion whether we need this functionality. If we remove it, it should be removed completely.

gerhardol:
> As the message says, it selects the parent folder of the deleted file.
I don't have a strong opinion whether we need this functionality. If we remove it, it should be removed completely.

So you see selecting the directory as a feature?
I can drop this commit, just a step in the ever continuing changes here....

gerhardol:

gerhardol:
I did not want this to be merged before I agreed with @mstv if show in file tree parent is a feature...
The label need to be removed before asking to forget the merge.

Should it be restored?

mstv:
I was waiting for other votes.
Now that it is gone, just remove the unreachable code, too.

PR #8619, opened 2020-11-19
Fix left panel broken sort and filtering

RussKie:
## Proposed changes

- Correct sorting in the left panel without enabled filters, as the current behaviour doesn't requery data, nor refresh the content.
    There are two distinct, yet connected, operations:
    - sort (with and without a filter), and
    - filtering.

    The "filter" is primarily concerned with rendering, i.e. visually indicate  which refs are accessible and which are not. Whilst the "sort"'s purpose    to requery data from git, and refresh the tree applying any existing filters.

    To facilitate the above two distinct API are provided:
    - `ToggleFilterMode(bool)` to turn filter-mode on/off
    - `Refresh()` which breaks local caches for each participating tree, and    refreshes data.

- Whenever the refs tree is refreshed due to changes in refs order    the tree would retain the `SelectedNode` by index. E.g. if before     a refresh the 3rd node was seected, after the refresh the     3rd node would remain selected.
    However after the reorder the currently selected ref (in the revision graph)     would very likely be at a different index, and the fix ensures the    originally selected ref node remains selected.

    This re-selection occurs before participating trees apply custom    node selection logic.

### Before

* Before 406c6c390f5e3f41767a1554b5fa4c27ad772160
sort1
* Before 9712e683b2882305642af6452d2e9c257876bcd1
sort2

### After

* No filters
sort3
* With filters
sort4

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
@@ -189,7 +189,7 @@ namespace GitUI.BranchTreePanel
             }
         }
 
-        [DebuggerDisplay("(Local) FullPath = {FullPath}, Hash = {ObjectId}, Visible: {IsVisible}")]
+        [DebuggerDisplay("(Local) FullPath = {FullPath}, Hash = {ObjectId}, Visible: {Visible}")]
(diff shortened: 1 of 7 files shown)


RussKie:
Merging, will attend feedback later.

Issue #8621, opened 2020-11-20
Incorrect refs sorting


RussKie:
## Current behaviour

Refs sort is incorrect depending on whether the resulting list contains dereferences (i.e. tags).
The current sort order is "ref, then deref":
https://github.com/gitextensions/gitextensions/blob/2d52621004c9b72b17915271b26e49de9b6cd3f7/GitCommands/Git/Commands/GitCommandHelpers.cs#L133-L134

This works for tags:

<details>
<summary>tags: deref, ref</summary>

```
d4b0f48bbf3e39a71f5d7ff5231be5678d4aa0f1  refs/tags/v3.4.3^{}
082cfd97d09c47346638b1873f5c8cf6ed139382  refs/tags/v3.4.2^{}
f49b4d059c5b135b4b2bf9eae4f443f19f6ac5df  refs/tags/v3.4.1^{}
a4f5744118651776ec26d570b462e5ba8fc18afc  refs/tags/v3.4^{}
096b4d56709e15a67b42a471bbe81ce3c95ae7f5  refs/tags/v3.4-RC1^{}
5a97671645532bcedc443bac7b727f40db47cb5c  refs/tags/v3.3.1^{}
e52f5178e8d3c9899efbe6fae66026865afa3fe1  refs/tags/v3.3^{}
7c0cfd75b3778b5b832fa210c1a2acc5850b91d4  refs/tags/v3.3-RC1^{}
bc07663c3c893f1d896520e0bdcfbe35cfe4b729  refs/tags/v3.2.1^{}
abac1b4564a5166f0453fa3ee459769e8a199b6f  refs/tags/v3.2^{}
2f872105b4eb42421d9fdafd68e90a9e27d20d96  refs/tags/v3.1.1^{}
bb1687f6bd0450d0c1a2bae10a37c65493fc1294  refs/tags/v3.1^{}
497059b2205086d8a44a33dd29f2cae2b4bfffc2  refs/tags/v3.1-RC1^{}
23b6f51905006ccdde8701591df706284b4155dc  refs/tags/v3.0.2^{}
a269358b22a5dc1ea36928bb87f9e535de76a058  refs/tags/v3.0.1^{}
fca7cf228b481ee8c1b779cf7b882ccdfbdcd1bc  refs/tags/v3.00.00^{}
8cbcd2b25223c73fbe9f7dad08e498cf82734bfa  refs/tags/v3.00.00-rc2^{}
[...]

PR #8622, opened 2020-11-20
Prefer correct local and remote branches sort to tags

RussKie:
Relates to #8621

## Proposed changes

Git tags may be concrete refs and ephemeral (i.e. dereferences), which
adds a layer of complexity to sorting refs.

The original sort mechanism ordered by refs first, followed by derefs.
However it caused local and remote branches to be sorted incorrectly.

Invert the sort order to ensure branches are sorted correctly at the expense
of correct tags order.

----


gerhardol (approved):

mstv (approved):
The sorting A-Z seems to be OK for the GE repo. It is identical as with my older own build.
Though I do not get the issue regarding tags.

PR #8623, opened 2020-11-20
Git Extension commands refactoring

ivangrek:
## Proposed changes

- Move GE UI commands to a separate classes
- Easier code maintenance 😉 
- ~~Change boolean command result to: true -> works without changes, false -> throws exception~~ (maybe another time)

## Test methodology 

- manual
- unit test

## Test environment(s) 

- Git Extensions 33.33.33
- Build 1f168ab4faad05c38f0ed3c1196b35ccf08e427f
- Git 2.28.0.windows.1
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Commands/GitExtensionCommandFactory.cs
+++ b/GitUI/Commands/GitExtensionCommandFactory.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using GitCommands;
 
 namespace GitUI.Commands
 {
@@ -10,6 +11,7 @@ namespace GitUI.Commands
         private const string AddFilesCommandName = "addfiles";
         private const string ApplyCommandName = "apply";
         private const string ApplyPatchCommandName = "applypatch";
+        private const string BlameCommandName = "blame";
         private const string BrowseCommandName = "browse";
 
         private readonly string[] _arguments;
@@ -34,6 +36,9 @@ namespace GitUI.Commands
                 // [filename]
                 [ApplyCommandName] = CreateApplyPatchCommand,
                 [ApplyPatchCommandName] = CreateApplyPatchCommand,
+
+                // [filename]
+                [BlameCommandName] = CreateBlameCommand,
 
                 // [path] [-filter]
                 [BrowseCommandName] = CreateBrowseCommand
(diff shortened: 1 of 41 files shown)


GitUI/Commands/GitExtensionCommandFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+        public GitExtensionCommandFactory(string[] arguments)
+        {
+            _arguments = arguments;
+        }
+
+        public IGitExtensionCommand Create()
+        {
+            if (_arguments.Length <= 1)
+            {
+                return null;
```
gerhardol, line 23: In the current setup, this should be handling Browse
(a comment would be sufficient at this stage)

gerhardol (commented):
Proposal looks OK to me, but I do not have much opinions here.

Surprisingly, there are users for other commands than 'browse'.
(and the commands for MSVS, shellext - at least shellext is in sync with GE so that can be changed.)
There are occasional issues and requests for commands. 
What features that really are used is hard to say.
So the importance is low for me.

There are tests already, that @mstv have fun memories from...

GitUI/Commands/GitExtensionCommandFactory.cs
ivangrek: In Program.cs.

```
// until we complete the migration
if (сommand is null)
```

gerhardol: Yes, but starting without parameters is not doing anything at all now, this comment is to remove the check when everything is complete.
You can add the comment there instead if you prefer or keep the args check for now

ivangrek: Thanks! I understood. I made these commands, but I filled in only one. So yes, it won't work.
If there are objections, I think there will be no point in continuing.

ivangrek:
DoActionOnRepo 🎯 

```
/// <param name="requiresValidWorkingDir">If action requires valid working directory</param>
/// <param name="owner">Owner window</param>
/// <param name="changesRepo">if successfully done action changes repo state</param>
/// <param name="preEvent">Event invoked before performing action</param>
/// <param name="postEvent">Event invoked after performing action</param>
/// <param name="action">Action to do. Return true to indicate that the action was successfully done.</param>
/// <returns>true if action was successfully done, false otherwise</returns>
public bool DoActionOnRepo(
    [CanBeNull] IWin32Window owner,
    [InstantHandle] Func<bool> action,
    bool requiresValidWorkingDir = true,
    bool changesRepo = true,
    EventHandler<GitUIEventArgs> preEvent = null,
    EventHandler<GitUIPostActionEventArgs> postEvent = null)
```

ivangrek:
Note. The "Pull" and "Synchronize" commands that use "Pull" update the settings.
AppSettings.SaveSettings () seems to be used only for these commands. 
This method seems strange to me.

```
private static void UpdateSettingsBasedOnArguments(IReadOnlyDictionary<string, string> arguments)
{
    if (arguments.ContainsKey("merge"))
    {
        AppSettings.DefaultPullAction = AppSettings.PullAction.Merge;
    }

    if (arguments.ContainsKey("rebase"))
    {
        AppSettings.DefaultPullAction = AppSettings.PullAction.Rebase;
    }

    if (arguments.ContainsKey("fetch"))
    {
        AppSettings.DefaultPullAction = AppSettings.PullAction.Fetch;
    }

    if (arguments.ContainsKey("autostash"))
    {
        AppSettings.AutoStash = true;
    }
}
```

gerhardol (commented):
+1
A very big PR, I have tried to review for some days...
The tests are mostly not changed, so if the tests were comprehensive, we should trust the refactoring.

A good refactoring, the code is broken down (and GitUI\Commands\GitExtensionCommandFactory.cs makes use of constants in a good way as discussed).
Will try to look some more on GitExtensionCommandFactory.cs and compare to documentation.

PR #8627, opened 2020-11-21
FileHistory: Git command log

gerhardol:
## Proposed changes

Add button to show _Git command log_ in FormFileHistory
#8076 changed FormFileHistory to run in a separate instance and Git command log is no longer including those commands.

There are no appropriate menus to put this is in, so a button on the top level is added
An alternative is maybe to only use a shortcut key.

### Before

image

### After

image

## Test methodology 

Manual

----


RussKie (approved):
Should we bind the same <kbd>F12</kbd> to it?

gerhardol:
> Should we bind the same F12 to it?

We are maybe 10 people with real use of this form, so lets not spend too much time on it.
I want this to be recognized by these 10 users though so it is easily enough recognized and believe a button is better (also for the 100 users or so that will play around with this and maybe learn something).
So I did not investigate this (but can do if requested.)

PR #8628, opened 2020-11-21
RangeDiff: Commit count in name

gerhardol:
## Proposed changes

#8516 etc added git-range-diff. After some usage, I would like to change the presentation a little.

Currently the first and second commits are presented. I have found that I often want to know the commit count for first/second  to the merge-base, similar to ahead/behind info. The first/second is (normally, see below) listed anyway in other groups and the second commit is not often seen as as the group name is too long.

If one of the first/second selected is artificial, the range diff uses HEAD. This is not listed in the other group names, but should be obvious.
To avoid a second call to git-rev-list (adding about 40 ms on my PC), a modified git-rev-list call was added.

### Before

FileStatusList must be wide to see the second revision

image

image

### After

image

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -818,11 +818,11 @@ namespace GitCommands
             return null;
         }
 
-        public int? GetCommitDiffCount(string parentHash, string childHash)
+        public int? GetCommitDiffCount(ObjectId baseId, ObjectId parentId)
         {
             var args = new GitArgumentBuilder("rev-list")
             {
-                $"{parentHash}...{childHash}",
+                $"{baseId} {parentId}",
                 "--count"
             };
             var output = _gitExecutable.GetOutput(args);
@@ -833,6 +833,25 @@ namespace GitCommands
             }
 
             return null;
+        }
+
+        public (int? first, int? second) GetCommitRangeDiffCount(ObjectId firstId, ObjectId secondId)
+        {
+            var args = new GitArgumentBuilder("rev-list")
+            {
+                $"{firstId}...{secondId}",
+                "--count",
+                "--left-right"
+            };
+            var output = _gitExecutable.GetOutput(args);
+
+            var counts = output.Split('\t');
+            if (counts.Length == 2 && int.TryParse(counts[0], out var first) && int.TryParse(counts[1], out var second))
+            {
+                return (first, second);
+            }
+
+            return (null, null);
         }
 
         public string GetCommitCountString(string from, string to)
--- a/GitUI/UserControls/FileStatusDiffCalculator.cs
+++ b/GitUI/UserControls/FileStatusDiffCalculator.cs
@@ -168,12 +168,13 @@ namespace GitUI
                 statuses: commonBaseToAandB));
 
             // Add rangeDiff as a separate group (range is not the same as diff with artificial commits)
-            List<GitItemStatus> statuses = new List<GitItemStatus> { new GitItemStatus { Name = Strings.DiffRange, IsRangeDiff = true } };
-
-            var desc = Strings.DiffRange + ": " + GetDescriptionForRevision(describeRevision, firstRevHead) + " -> " +
-                       GetDescriptionForRevision(describeRevision, selectedRevHead);
+            var statuses = new List<GitItemStatus> { new GitItemStatus { Name = Strings.DiffRange, IsRangeDiff = true } };
             var first = firstRev.ObjectId == firstRevHead ? firstRev : new GitRevision(firstRevHead);
             var selected = selectedRev.ObjectId == selectedRevHead ? selectedRev : new GitRevision(selectedRevHead);
+            var (baseToFirstCount, baseToSecondCount) = module.GetCommitRangeDiffCount(first.ObjectId, selected.ObjectId);
+            const int rangeDiffCommitLimit = 100;
+            var desc = $"{Strings.DiffRange} {baseToFirstCount ?? rangeDiffCommitLimit}↓ {baseToSecondCount ?? rangeDiffCommitLimit}↑";
+
             var rangeDiff = new FileStatusWithDescription(
                 firstRev: first,
                 secondRev: selected,
@@ -185,13 +186,12 @@ namespace GitUI
 
             // Git range-diff has cubic runtime complexity and can be slow and memory consuming, so just skip if diff is large
             // to avoid that GE seem to hang when selecting the range diff
-            const int maxRangeDiffCommits = 100;
             int count = (baseA == null || baseB == null
-                ? module.GetCommitDiffCount(firstRevHead.ToString(), selectedRevHead.ToString())
-                : module.GetCommitDiffCount(baseA.ToString(), firstRevHead.ToString())
-                + module.GetCommitDiffCount(baseB.ToString(), selectedRevHead.ToString()))
-                ?? maxRangeDiffCommits + 1;
-            if (!GitVersion.Current.SupportRangeDiffTool || count > maxRangeDiffCommits)
+                ? baseToFirstCount + baseToSecondCount
+                : module.GetCommitDiffCount(baseA, firstRevHead)
+                + module.GetCommitDiffCount(baseB, selectedRevHead))
+                ?? rangeDiffCommitLimit;
+            if (!GitVersion.Current.SupportRangeDiffTool || count >= rangeDiffCommitLimit)
             {
                 var range = baseA is null || baseB is null
                     ? $"{first.ObjectId}...{selected.ObjectId}"
@@ -199,11 +199,12 @@ namespace GitUI
                 statuses[0].IsStatusOnly = true;
 
                 // Message is not translated, considered as an error message
-                statuses[0].ErrorMessage = $"# The symmetric difference from {first.ObjectId.ToShortString()} to {selected.ObjectId.ToShortString()} is {count} which is higher than the limit {maxRangeDiffCommits}\n" +
-                    "# Git range-diff may take a long time and Git Extensions seem to hang during execution, why the command is not executed\n" +
-                    "# You can still run the command in a Git terminal\n" +
-                    "# Remove '--no-patch' to see changes to files too\n" +
-                    $"git range-diff {range} --no-patch";
+                statuses[0].ErrorMessage =
+                    $"# The symmetric difference from {first.ObjectId.ToShortString()} to {selected.ObjectId.ToShortString()} is {count} >= {rangeDiffCommitLimit}\n"
+                    + "# Git range-diff may take a long time and Git Extensions seem to hang during execution, why the command is not executed.\n"
+                    + "# You can still run the command in a Git terminal.\n"
+                    + "# Remove '--no-patch' to see changes to files too.\n"
+                    + $"git range-diff {range} --no-patch";
             }
 
             return fileStatusDescs;


RussKie (commented):
The actual implementation LGTM with a comment

GitUI/UserControls/FileStatusDiffCalculator.cs
```diff
@@ -168,12 +168,13 @@ public FileStatusDiffCalculator(Func<GitModule> getModule)
...
 
             // Add rangeDiff as a separate group (range is not the same as diff with artificial commits)
-            List<GitItemStatus> statuses = new List<GitItemStatus> { new GitItemStatus { Name = Strings.DiffRange, IsRangeDiff = true } };
-
-            var desc = Strings.DiffRange + ": " + GetDescriptionForRevision(describeRevision, firstRevHead) + " -> " +
-                       GetDescriptionForRevision(describeRevision, selectedRevHead);
+            var statuses = new List<GitItemStatus> { new GitItemStatus { Name = Strings.DiffRange, IsRangeDiff = true } };
             var first = firstRev.ObjectId == firstRevHead ? firstRev : new GitRevision(firstRevHead);
             var selected = selectedRev.ObjectId == selectedRevHead ? selectedRev : new GitRevision(selectedRevHead);
+            var (baseToFirstCount, baseToSecondCount) = module.GetCommitRangeDiffCount(first.ObjectId, selected.ObjectId);
```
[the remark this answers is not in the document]
gerhardol: There are some more methods needed for that to happen.
Also, I would prefer to not share this with plugins. If so, the methods should change to strings instead of ObjectId (as there may be situations where treeish identifiers like HEAD could be used.)

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1061	'IGitModule' does not contain a definition for 'GetTreeFiles' and no accessible extension method 'GetTreeFiles' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	40	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetCombinedDiffFileList' and no accessible extension method 'GetCombinedDiffFileList' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	61	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetCommitDiffCount' and no accessible extension method 'GetCommitDiffCount' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	191	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetCommitDiffCount' and no accessible extension method 'GetCommitDiffCount' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	192	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetCommitRangeDiffCount' and no accessible extension method 'GetCommitRangeDiffCount' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	173	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetDiffFilesWithSubmodulesStatus' and no accessible extension method 'GetDiffFilesWithSubmodulesStatus' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	54	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetDiffFilesWithSubmodulesStatus' and no accessible extension method 'GetDiffFilesWithSubmodulesStatus' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	87	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetDiffFilesWithSubmodulesStatus' and no accessible extension method 'GetDiffFilesWithSubmodulesStatus' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	138	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetDiffFilesWithSubmodulesStatus' and no accessible extension method 'GetDiffFilesWithSubmodulesStatus' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	147	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetDiffFilesWithSubmodulesStatus' and no accessible extension method 'GetDiffFilesWithSubmodulesStatus' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	148	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetMergeBase' and no accessible extension method 'GetMergeBase' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	103	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetMergeBase' and no accessible extension method 'GetMergeBase' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	108	Active
Error	CS1061	'IGitModule' does not contain a definition for 'GetMergeBase' and no accessible extension method 'GetMergeBase' accepting a first argument of type 'IGitModule' could be found (are you missing a using directive or an assembly reference?)	GitUI	F:\dev\gc\gitextensions\GitUI\UserControls\FileStatusDiffCalculator.cs	111	Active
```

mstv:
I can follow the reasoning for this change.
Just curious: What is such interesting in *how many* commits a branch is ahead & behind? I just want to know *whether* it is ahead & behind. BTW: "(gone)" as text is distracting.
range-diff is a little special because of its limitations (max. distance in commits; commits must match, i.e. not usable for squashed branches -- or have I missed it?)
So finally, I don't have objections.

GitCommands/Git/GitModule.cs
```diff
@@ -835,6 +835,25 @@ private IGitItem GetSubmoduleCommitHash(string filename, string refName)
...
+            var args = new GitArgumentBuilder("rev-list")
+            {
+                $"{firstId}...{secondId}",
+                "--count",
+                "--left-right"
+            };
+            var output = _gitExecutable.GetOutput(args);
+
+            var counts = output.Split('\t');
+            if (counts.Length == 2 && int.TryParse(counts[0], out var first) && int.TryParse(counts[1], out var second))
```
mstv, line 849: This is another reason for adding a generic detection of git error and warning messages to `GetOutput` etc.

GitUI/UserControls/FileStatusDiffCalculator.cs
```diff
@@ -168,12 +168,13 @@ public FileStatusDiffCalculator(Func<GitModule> getModule)
...
+            const int maxRangeDiffCommits = 100;
+            var desc = $"{Strings.DiffRange} {baseToFirstCount ?? maxRangeDiffCommits + 1}↓ {baseToSecondCount ?? maxRangeDiffCommits + 1}↑";
```
mstv, line 176: "100+" or ">100" instead of "101"?

mstv (approved):

RussKie:
100+ has my vote

GitUI/UserControls/FileStatusDiffCalculator.cs
gerhardol: changed to `angeDiffCommitLimit`

gerhardol:
> Just curious: What is such interesting in _how many_ commits a branch is ahead & behind? I just want to know _whether_ it is ahead & behind. 

To some extent, this change was initiated by that First->Second is too long and therefore not really useful.
The exact number is not so interesting, but to see where the branches are attached. Like if my local branch is about 4 commits and I compare to another branch with about 4 commits and the count is 8-9, there are other commits involved to, if it is 4-4 there are likely no other commits involved.

> BTW: "(gone)" as text is distracting.

I agree but this is Git name. May be changed to "-".

> range-diff is a little special because of its limitations (max. distance in commits; commits must match, i.e. not usable for squashed branches -- or have I missed it?)

Correct, the actual diff is hard to pretend (and even not always correct).
But for rebased branches it is very useful: Has someone else changed anything when rebasing, was the rebase correct?
I often have branches with 30 commits for review that I rearrange before creating a PR, with range-diff I get a better overview.

GitCommands/Git/GitModule.cs
```diff
@@ -835,6 +835,25 @@ private IGitItem GetSubmoduleCommitHash(string filename, string refName)
...
         }
 
+        public (int? first, int? second) GetCommitRangeDiffCount(ObjectId firstId, ObjectId secondId)
+        {
+            var args = new GitArgumentBuilder("rev-list")
+            {
+                $"{firstId}...{secondId}",
+                "--count",
+                "--left-right"
+            };
```
ivangrek, line 845: Could we use constants for these commands?

```
public static class GitCommands
{
    public static class RevList
    {
        public const string Command = "rev-list";
        public const string CountParameter = "--count";
        public const string LeftRightParameter = "--left-right";
    }
}
```

or

```
public static class GitCommands
{
    public const string RevList = "rev-list";

    public static class RevListParameters
    {
        public const string Count = "--count";
        public const string LeftRight = "--left-right";
    }
}
```

Most likely there will be many common parameters.

gerhardol: If this is changed, it should be a generic change for all commands, not in this PR.

Personally I see just one reason to make constants for commands: It is possible to search where they are used.
"rev-list" is unique though, but "log" is not.

Other than that, for Git commands you get down to a text string anyway. Using constants will just add another layer when viewing the code.

Other than that, I would like to merge this.
Changes to IGitModule will affect more commands than GetCommitRangeDiffCount, like GetDiffFilesWithSubmodulesStatus.

ivangrek: > Personally I see just one reason to make constants for commands: It is possible to search where they are used.
"rev-list" is unique though, but "log" is not.

The search itself is important.

I use the rule: "Everything that can be calculated at the stage of creating a program must be calculated."
Example:

image

image

> Other than that, for Git commands you get down to a text string anyway. Using constants will just add another layer when viewing the code.

And that's the key.
Why not use it like that?

`var args = new GitArgumentBuilder($"rev-list {firstId}...{secondId} --count --left-right");`

Or so?

`var args = $"rev-list {firstId}...{secondId} --count --left-right";`

I think we could find this door if we looked, the key is already in our hands. 😉

> Other than that, I would like to merge this.

Well, Ok.

gerhardol: > Why not use it like that?

The Git command (like "rev-list") is a hook for options etc so that must be separate.
The (generic) argument builder handles some types native and simplifies use of booleans etc.

Another layer will require me too look up the constant when browsing the code which is not an advantage to me.

ivangrek:
```
var args = new GitArgumentBuilder("rev-list");
var args = new GitArgumentBuilder(GitCommands.RevList);
```

I understand correctly, do you think that the second method may require additional actions?

gerhardol: > I understand correctly, do you think that the second method may require additional actions?

No additional actions, but I do not see the actual commands at a glance.

GitCommands/Git/GitModule.cs
RussKie: How would it help?

GitUI/UserControls/FileStatusDiffCalculator.cs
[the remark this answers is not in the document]
RussKie: This is something we'll need to fix going forward.

RussKie (approved):

RussKie:
@gerhardol squash and merge

Issue #7152, opened 2019-09-16
Rebase re-apply commit status value is incorrect


AdamWillden:
## Current behaviour

As noted in the latest release note a fix was attempted for this very issue in [#6351](https://github.com/gitextensions/gitextensions/pull/6531) by @pmiossec (thank you for looking into it).

While the text was better it still doesn't display nor function correctly (for me at least). Please see GIF below.

To put the GIF in in text form:

1. The commits to re-apply never shows the history of the changes (never has for me). i.e. the list gets truncated as it gets procesed.
2. The commits to re-apply status box always shows the commit AFTER the one being applied.
3. Once I skipped the undesired commit the status wrongly shows `Skipped` but against the commit AFTER the one to be applied (see issue 2).

## Expected behaviour

1. Retain the history of commits
2. Show the commit still to be applied
3. Show correct value against correct commit.

## Steps to reproduce

See gif!

## Screenshots

![000cea13-86f3-496c-9e97-3c33dc890759](https://user-images.githubusercontent.com/4629144/64974218-48da1a00-d8a4-11e9-9a1b-c72571085667.gif)

## Did this work in previous version of GitExtensions

No

## Environment

- Git Extensions 3.2.1.6628
- Build bc07663c3c893f1d896520e0bdcfbe35cfe4b729
- Git 2.23.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3460.0
- DPI 96dpi (no scaling)

gerhardol:
Valid points

pmiossec:
> 1. The commits to re-apply never shows the history of the changes (never has for me). i.e. the list gets truncated as it gets processed.

I would have loved to have that but that is not possible with how the code actually work without redoing all the feature from scratch. Because what are displayed are the patch files generated by git (and the patchs for the already applied commits are not generated). That's a big refactoring without a great added value so I decided to not changes that.

> 2. The commits to re-apply status box always shows the commit AFTER the one being applied.

It feel strange to me because my work in #6351 was about that to maintain the first list of patches instead of reading patch files on the drive to always have the same list of patches with the status updated. There should be a bug but that I never encountered 😢 

  > 3. Once I skipped the undesired commit the status wrongly shows Skipped but against the commit AFTER the one to be applied (see issue 2).

Same here.

I will try to reproduce...

pmiossec:
>I will try to reproduce...

I don't reproduce 😭 It is working well for me...

[...]

Issue #6350, opened 2019-03-07
Rebase dialog should show state of every commit in the rebase


johndog:
## Feature description

The rebase window shows progress on the rebase and offers commands for interacting with it:

![image](https://user-images.githubusercontent.com/22213780/53944805-fb6b2d80-4074-11e9-84e3-534eb3736cbe.png)

The "Status" column shows the state of each commit in the rebase.  However, it doesn't actually:  it seems to skip over commits that were fast-forward-able.  The suggestion is to show the state of every commit listed in the todo script.

## Environment

- Git Extensions 3.0.2.5232
- Build 23b6f51905006ccdde8701591df706284b4155dc
- Git 2.20.1.vfs.1.1.97.g8af1f39
- Microsoft Windows NT 10.0.18346.0
- .NET Framework 4.8.3745.0
- DPI 96dpi (no scaling)

vbjay:
First column.

pmiossec:
>First column.

With such a bad name.... We should change it for "Command" (to stick for the appellation used in the interactive file) or "Action".

Other than that, I also don't understand the request.
@johndog could you be clearer?
I myself don't really like the UX of the popup but never be able to find out what was really wrong (except the fact to show the list empty at the beginning!)

pmiossec:
I think that this form should be cut in 2. One to select the rebase option and one to manage the rebase once started...

pmiossec:
And also the algorithm used to determine the status is false and we always display "Skipped" for commits applied.

johndog:
I support larger efforts to improve the rebase experience, I just wanted to zero in on problems with accuracy.

Here's a concrete scenario:

1. Pick a change 7 commits from head, right click and choose "rebase interactively"
2. Observe: 6 commits should show up in the rebase script.
3. Half way down, insert a "break" command, between the 3rd and 4th commit
4. Let the rebase continue.

**Result**: When the Rebase UI appears, only 3 commits will be listed, and the first will show "Next to apply" as status

**Expect**:  The commits in the script should all show with status "Applied", _then_ the next commit should show "next to apply".  Furthermore, if a commit has actually been partially applied and halted due to merge conflicts, status should show "In Progress".  There should be no doubt about what happens when "continue" is clicked, be that committing what's in the index or beginning to apply the next commit.

gerhardol:
The focus should be on the initial form in my opinion.
See #5161

johndog:
Two issues with that thought:

[...]

Issue #5161, opened 2018-07-12
Extend UI of interactive rebase editor


drewnoakes:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

Git Extensions editor is registered for the editing of todo lists for interactive rebases.

We recently added syntax highlighting here, but there is more we could do.

I propose loading additional UI when we detect we're editing a todo script, via an optional split view:

![image](https://user-images.githubusercontent.com/350947/42644640-fa66bad0-85f3-11e8-9152-e06454a73bda.png)

The lower section shows information about the commit of the selected line in the editor above.

I hope it would be straightforward enough to reuse these controls in a new context.

Existing syntax highlighting code already knows how to detect these scripts.

Visibility of the panel to be stored in `AppSettings`.

Instead of closing this window to start the rebase, It'd be interesting to keep it open throughout the rebase. Maybe run the rebase in a console tab. Maybe highlight progress during the rebase in the (now readonly) editor.

vbjay:
:1st_place_medal:  :100:   :+1:

vbjay:
Hide it if it is not a todo line?

gerhardol:
Great
The current workaround is to open another GE instance

crypto-rsa:
Or perhaps avoid the editor completely and use a simple list control instead? After all, there are just a few operations needed for the interactive rebase - reorder rows and change the action in each row. Using a full text editor introduces unnecessary "degrees of freedom" for that. For example you can accidentally change the commit hash in a row which is probably not what you want.
The commit info at the bottom would be super useful.

vbjay:
I like it. A checked list box that would let you uncheck to remove a line
but still see it. A drop-down for the actions. Up/down arrow buttons to
move the items.

RussKie:
Very nice!
But I wouldn't change or add UI controls, not at this stage anyway.

drewnoakes:
It would be a nice feature for 3.1.

Has anyone seen a good UI for interactive rebasing?

nick-bulleid:
TortoiseGit has a nice UI for this kind of thing.

Seems like you could just extend the current UI in the rebase dialog, as that already shows a list of commits to rebase in the case of conflicts.

RussKie:
Any screenshots or links?

nick-bulleid:
https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html

Different options are available on right click of the commit.

[...]

PR #8638, opened 2020-11-25
Fix rebase status displayed during a rebase

### LINKED ISSUE #8540: Wrong commit displayed as "Applying..." when there is a conflict
## Current behaviour

In the rebase popup, when solving a conflict, the wrong commit is displayed as "Applying".
That's the next one that is displayed as "Applying" instead of the current one which is not displayed.

## Expected behaviour

The current commit should be displayed as "Applying..."

## Screenshots

image

Yes. It was fixed for v3.2 ( #6531 ) 
But it could be a git behavior change more than a GitExtensions regression...

## Environment

- Git Extensions 33.33.33
- Build 905403fe2edfa879c96e293973afebe555eb6e8e
- Git 2.28.0.windows.1
- .NET Framework 4.8.4240.0
- DPI 192dpi (200% scaling)

## Diagnostics

GitExtensions logic is to use the patch files generated by git to detect where we are in the rebase.

Previous behavior was:
image

It seems that now, only the remaining patches are displayed :(

pmiossec:
Fixes #8540, #7152 and #6350

## Proposed changes

- Fix the list of displayed commits that are rebased to always display all of them and with the good `Status` displayed
- other adjustments:
    * Move the important 'Status' column to be the first one (and not lost after author and date so that it could be more easily compare to the `Action` and `Subject`)
    * Display commit hash
    * Rename "Name" header to "Action" (because it display the define action. Especially when doing an interactive rebase)
    * Being able to open a commit diff window by double click (like it is possible to see the patch file for the patches)
    * Set a minimum height for a row (to avoid possible strange behavior)
    * Resize some columns

### Before

image

### After

* New list:
image

* Opening a commit:
image

## Test methodology 

- Manual

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1985,8 +1985,9 @@ namespace GitCommands
                     patchFiles.Add(new PatchFile
                     {
                         Author = error ?? data.Author,
+                        ObjectId = data?.ObjectId,
                         Subject = error ?? data.Body,
-                        Name = parts[0],
+                        Action = parts[0],
                         Date = error ?? data.CommitDate.LocalDateTime.ToString(),
                         IsNext = isApplying,
                         IsApplied = !isCurrentFound,
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1945,14 +1945,14 @@ namespace GitCommands
         private static readonly Regex HeadersMatch = new Regex(@"^(?<header_key>[-A-Za-z0-9]+)(?::[ \t]*)(?<header_value>.*)$", RegexOptions.Compiled);
         private static readonly Regex QuotedText = new Regex(@"=\?([\w-]+)\?q\?(.*)\?=$", RegexOptions.Compiled);
 
-        public bool InTheMiddleOfInteractiveRebase()
-        {
-            return File.Exists(GetRebaseDir() + "git-rebase-todo");
-        }
+        private string RebaseTodoFilePath => GetRebaseDir() + "git-rebase-todo.backup";
+        private string CurrentFilePath => GetRebaseDir() + "stopped-sha";
+
+        public bool InTheMiddleOfInteractiveRebase() => File.Exists(RebaseTodoFilePath);
 
         public IReadOnlyList<PatchFile> GetInteractiveRebasePatchFiles()
         {
-            string todoFile = GetRebaseDir() + "git-rebase-todo";
+            string todoFile = RebaseTodoFilePath;
             string[] todoCommits = File.Exists(todoFile) ? File.ReadAllText(todoFile).Trim().Split(new[] { '\n', '\r' }, StringSplitOptions.RemoveEmptyEntries) : null;
 
             var patchFiles = new List<PatchFile>();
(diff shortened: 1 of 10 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -1926,14 +1926,14 @@ private string GetNextRebasePatch()
         private static readonly Regex HeadersMatch = new Regex(@"^(?<header_key>[-A-Za-z0-9]+)(?::[ \t]*)(?<header_value>.*)$", RegexOptions.Compiled);
         private static readonly Regex QuotedText = new Regex(@"=\?([\w-]+)\?q\?(.*)\?=$", RegexOptions.Compiled);
 
-        public bool InTheMiddleOfInteractiveRebase()
-        {
-            return File.Exists(GetRebaseDir() + "git-rebase-todo");
-        }
+        private string RebaseTodoFilePath => GetRebaseDir() + "git-rebase-todo.backup";
+        private string CurrentFilePath => GetRebaseDir() + "stopped-sha";
```
ivangrek, line 1930: Could we use constants for strings?

GitCommands/Git/GitModule.cs
```diff
@@ -1951,19 +1953,28 @@ public IReadOnlyList<PatchFile> GetInteractiveRebasePatchFiles()
...
-                            Date = error ?? data.CommitDate.LocalDateTime.ToString(),
-                            IsNext = patchFiles.Count == 0
-                        });
+                    string commitHash = parts[1];
+                    CommitData data = _commitDataManager.GetCommitData(commitHash, out var error);
+                    var isApplying = currentCommitShortHash != null && commitHash.StartsWith(currentCommitShortHash);
+                    if (!isCurrentFound && isApplying)
+                    {
+                        isCurrentFound = isApplying;
                     }
```
ivangrek, line 1967:
``` suggestion
   isCurrentFound |= isApplying
```

ivangrek: false false => false
false true => true
true false => true
true true => true

OR

GitCommands/Git/GitModule.cs
RussKie: In which way?

gerhardol: In this case the string is used only once (related to this routine) and the statement is is the accessor so it is used as a constant with a dynamic part.
I lean on not having a constant here, as it adds another layer. (Me and @ivangrek has had that discussion in #8628.) At the same time I see the point when seeing a stray string in code.

In a wider usage, "git-rebase-todo" is hardcoded also in GitUI\Editor\FileViewerInternal.cs SetHighlightingForFile() to highlight in the editor. This file does not need to be syntax highlighted what I see.

To use constants for files GE uses should be a TODO.

gerhardol (approved):
Great change
Also a step towards #5161?

gerhardol:
This should be included, I added to next release milestone

pmiossec:
Updated to better handle skipped commits/patches. Now, if the popup is closed and opened again, the skipped commit are still displayed as "Skipped" (instead of "Applied").

And better highlight the currently applied commit/patch (with conflicts) with some color:
image

And fixed a bug so that the currently applied patch line was not selected (due to the selecting was not made in the UI thread).

RussKie:
@msftbot merge in 24 hours

jrfnl:
@pmiossec Thanks so much! This has been bugging me for quite a while, but never got round to opening an issue. Awesome to see this fixed!

PR #8648, opened 2020-11-30
Interactive rebase: Highlight 4 byte sha

gerhardol:
## Proposed changes 

TryHighlightInteractiveRebaseCommand() assumes that Git sha are longer than 5 chars when highlighting.
The minimum is 4, can be set:
`git config --global core.abbrev 4`
(Separate note: I would like to read and use this value in the GUI to limit printouts, separate.)

Also a minor optimization to use a compiled RegEx. (One of the few situations where `RegexOptions.Compiled` makes sense, should slow down in most other situations.)
(Seen in other situations, no good PR to attach it to and not worth a PR on its own.)

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3017,7 +3017,7 @@ namespace GitCommands
             //
             // Lines may also use \t as a column delimiter, such as output of "ls-remote --heads origin".
 
-            var regex = new Regex(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline);
+            var regex = new Regex(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline | RegexOptions.Compiled);
 
             var matches = regex.Matches(refList);
 
--- a/GitUI/Editor/RebaseTodoHighlightingStrategy.cs
+++ b/GitUI/Editor/RebaseTodoHighlightingStrategy.cs
@@ -127,7 +127,7 @@ namespace GitUI.Editor
                         {
                             var idLength = index - idStartIndex;
 
-                            if (idLength <= 5)
+                            if (idLength < 4)
                             {
                                 return false;
                             }


RussKie (approved):

GitUI/Editor/RebaseTodoHighlightingStrategy.cs
```diff
@@ -127,7 +127,7 @@ private static bool TryHighlightInteractiveRebaseCommand(IDocument document, Lin
                         {
                             var idLength = index - idStartIndex;
 
-                            if (idLength <= 5)
+                            if (idLength < 4)
```
drewnoakes, line 130: This will highlight a 3 byte SHA too. Is that intentional?

GitCommands/Git/GitModule.cs
```diff
@@ -3017,7 +3017,7 @@ public IReadOnlyList<IGitRef> ParseRefs([NotNull] string refList)
             //
             // Lines may also use \t as a column delimiter, such as output of "ls-remote --heads origin".
 
-            var regex = new Regex(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline);
+            var regex = new Regex(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline | RegexOptions.Compiled);
```
drewnoakes, line 3020: If compiling the Regex, it should ideally be cached so that it's not compiled again the next time `refList` is called.

drewnoakes: Regex compilation is fairly slow. It does give slightly faster run time, but the main benefit is a reduction in memory allocation while matching, in my experience. If it's being recompiled each time then my hunch is the overhead likely outweighs the benefit.

GitUI/Editor/RebaseTodoHighlightingStrategy.cs
gerhardol: The minimum length is 4 so 3 should have been excluded. But it makes not much difference, should be changed if updating again

GitCommands/Git/GitModule.cs
gerhardol: I did not try to measure (I expect it to be hard to see the difference), but this situation seem to be parsing a larger amount of data and I assume it to be faster.

In most other situations where Compiled is used, it should be changed.

GitUI/Editor/RebaseTodoHighlightingStrategy.cs
gerhardol: This is correct, a sha with 4 chars is highlighted, 3 is not

Issue #7598, opened 2020-01-05
Redesign of FormFileHistory


mstv:
## Current behaviour

FormFileHistory has several redundant controls. The layout does not use the available space well.
Many functions of FormBrowse which could be applied, are not provided to the user (refer to #6483).

## Expected behaviour

```
FormFileHistory
  = FormBrowse
    - LeftPanel
    - left part of toolbar and menu
    - tab "Console"
    + tab "Blame"
```
* single `CommitInfo` independent from tab "Commit" (`CommitDiff`) and tab "Blame" (`BlameControl`)
  placed by user left or right of the `RevisionGridControl` or -- if really used -- below as tab "Commit"
* tab "Diff" with `RevisionDiffControl` instead of plain `FileViewer`
* no tab "View" (`FileViewer`) because it is the same as on tab "Blame"

## Screenshots

![grafik](https://user-images.githubusercontent.com/36601201/71786099-35e08680-3008-11ea-8f41-50ccabc41a6f.png)
![grafik](https://user-images.githubusercontent.com/36601201/71786161-efd7f280-3008-11ea-996f-6fcb15c7d3a8.png)
![grafik](https://user-images.githubusercontent.com/36601201/71786212-33326100-3009-11ea-8240-2137a1f60d9a.png)
![grafik](https://user-images.githubusercontent.com/36601201/71786375-15fe9200-300b-11ea-9a8a-16706d8e3f1e.png)

## Environment

- GitExtensions version: 3.3.1

gerhardol:
CommitInfo: That is very useful to find other files changed in that commit (I occasionally need to get related commits in e.g. for merges in Browse though).

View: Not strictly needed, but blame is a little sluggish (diff could be used too, view complete file). But this is not important for me.
A reason to keep is though that that the FileHistory vs Blame (H/B) expects a View - affects other shortcuts.

Diff: Do you propose to remove that?
Also with the CommitInfo, it is simpler to see files (more important than View).

Regarding #6483 a Blame tab (or Diff with a separate Blame viewer) could be added to Browse, but then the revision grid need to be filtered to the current file too, which will be a little strange.

mstv:
At least from the user's perspective, I really think of a reused FormBrowse just with filtered revisions.
(Red highlighted menu items and toolbar to be hidden/removed)

![grafik](https://user-images.githubusercontent.com/36601201/71848148-08124500-30cf-11ea-89ba-9ebceff9aecc.png)

[...]

PR #8660, opened 2020-12-06
Custom mergetool

gerhardol:
#8194 for merge tools

## Proposed changes

Adds a dropdown menu in FormResolve to choose the merge tool without changing the defaults.
Some tools work better in some situations.
(a separate fixup to remove unused code, will be squashed)

There are some alternatives not handled hare:
* Expose this in the split button
* Implement this in the "Git Extensions" workflow where handling of files changed is asked in a terminal instead of a popup, 2way merge filters etc is handled. 

I just do not think it is worth it to implement this.

### Before

image

### After

image

## Test methodology 

Tests were added in #8194

----

### DIFF
--- a/GitUI/CommandsDialogs/FormResolveConflicts.Designer.cs
+++ b/GitUI/CommandsDialogs/FormResolveConflicts.Designer.cs
@@ -37,6 +37,7 @@ namespace GitUI.CommandsDialogs
             this.authorDataGridViewTextBoxColumn1 = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.ConflictedFilesContextMenu = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.OpenMergetool = new System.Windows.Forms.ToolStripMenuItem();
+            this.customMergetool = new System.Windows.Forms.ToolStripMenuItem();
             this.ContextMarkAsSolved = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStripSeparator3 = new System.Windows.Forms.ToolStripSeparator();
             this.ContextChooseLocal = new System.Windows.Forms.ToolStripMenuItem();
@@ -141,6 +142,7 @@ namespace GitUI.CommandsDialogs
             // 
             this.ConflictedFilesContextMenu.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
             this.OpenMergetool,
+            this.customMergetool,
             this.ContextMarkAsSolved,
             this.toolStripSeparator3,
             this.ContextChooseLocal,
@@ -169,6 +171,13 @@ namespace GitUI.CommandsDialogs
             this.OpenMergetool.Size = new System.Drawing.Size(195, 22);
             this.OpenMergetool.Text = "Open in mergetool";
             this.OpenMergetool.Click += new System.EventHandler(this.OpenMergetool_Click);
+            // 
+            // customMergetool
+            // 
+            this.customMergetool.Name = "customMergetool";
+            this.customMergetool.Size = new System.Drawing.Size(195, 22);
+            this.customMergetool.Text = "Open in &mergetool";
+            this.customMergetool.Click += new System.EventHandler(this.customMergetool_Click);
             // 
             // ContextMarkAsSolved
             // 
@@ -594,6 +603,7 @@ namespace GitUI.CommandsDialogs
         private System.Windows.Forms.ToolStripMenuItem ContextOpenRemoteWith;
         private System.Windows.Forms.ToolStripSeparator toolStripSeparator2;
         private System.Windows.Forms.ToolStripMenuItem OpenMergetool;
+        private System.Windows.Forms.ToolStripMenuItem customMergetool;
         private System.Windows.Forms.ToolStripSeparator toolStripSeparator3;
         private System.Windows.Forms.ToolStripMenuItem ContextSaveBaseAs;
         private System.Windows.Forms.ToolStripMenuItem ContextSaveLocalAs;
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormResolveConflicts.Designer.cs
```diff
@@ -170,6 +172,13 @@ private void InitializeComponent()
             this.OpenMergetool.Text = "Open in mergetool";
             this.OpenMergetool.Click += new System.EventHandler(this.OpenMergetool_Click);
             // 
+            // customMergetool
+            // 
+            this.customMergetool.Name = "customMergetool";
+            this.customMergetool.Size = new System.Drawing.Size(195, 22);
+            this.customMergetool.Text = "Open in mergetool";
```
mstv, line 179:
```suggestion
            this.customMergetool.Text = "Open in &mergetool";
```

GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -1243,6 +1256,28 @@ private void OpenMergetool_Click(object sender, EventArgs e)
             OpenMergeTool();
         }
 
+        private void customMergetool_Click(object sender, EventArgs e)
+        {
+            if (sender == customMergetool)
+            {
+                // Cancel dropdown manually if "main menu" is selected
+                customMergetool.HideDropDown();
+                ConflictedFilesContextMenu.Hide();
```
mstv, line 1265: Shouldn't we exit here? `customTool` will be `null`, won't it?

GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -295,6 +295,24 @@ private static string FixPath(string path)
             return (path ?? "").ToNativePath();
         }
 
+        public void ReloadCustomMergetools()
```
RussKie, line 298: Why "reload"?

RussKie, line 298: Why "public"?

GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -1243,6 +1256,28 @@ private void OpenMergetool_Click(object sender, EventArgs e)
...
+        private void customMergetool_Click(object sender, EventArgs e)
+        {
+            if (sender == customMergetool)
+            {
+                // Cancel dropdown manually if "main menu" is selected
+                customMergetool.HideDropDown();
+                ConflictedFilesContextMenu.Hide();
+            }
+
+            string customTool = (sender as ToolStripMenuItem)?.Tag as string;
```
RussKie, line 1268: Should we assert it is not null or empty?
```suggestion
            string customTool = (sender as ToolStripMenuItem)?.Tag as string;
            Debug.Assert(!string.IsNullOrWhitespace(customTool));
```

RussKie (commented):
👍

GitUI/CommandsDialogs/FormResolveConflicts.cs
gerhardol: Cut&paste from revdiff. Will change.

gerhardol: Cut&paste from revdiff. Will change.

GitUI/CommandsDialogs/FormResolveConflicts.cs
gerhardol: if customTool is null or empty, the default mergetool will be used so this works fine.

In this case it is possible to select both the top level "Open in mergetool" and the menu. Maybe mostly used while the list of tools is building (can takes ten seconds on PC with virus programs). If top is selected, the menus must be closed directly. An alternative is to ignore the top level selection once there is a dropdown created.

For the custom difftool action in revdiff, top level has no effect as there is a separate menu item.

image

A side discussion, I would like to use custom difftool in more scenarios (like History, diff any two files), then adding a second menu item will duplicating several items and the top level with dropdown should be considered.

GitUI/CommandsDialogs/FormResolveConflicts.cs
gerhardol: null and empty is OK, see comment to mstv.

GitUI/CommandsDialogs/FormResolveConflicts.cs
mstv: > A side discussion, I would like to use custom difftool in more scenarios (like History, diff any two files)

FormHistory is missing too much. And I do not see a real difference to FormBrowse. That's why I would not maintain FormHistory but replace it with a variant of FormBrowse as proposed in #7598, (which has become even easier since the history is shown in a dedicated GE instance).

gerhardol: History discussed in separate issue

If other forms are changed is also not in this PR, except that I would like to use the same behavior if changed elsewhere: action on the main menu and submenu

GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -295,6 +295,24 @@ private static string FixPath(string path)
             return (path ?? "").ToNativePath();
         }
 
+        private void LoadCustomMergetools()
+        {
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                var tools = await Module.GetCustomDiffMergeTools(isDiff: false);
```
RussKie, line 302: Are we running on UI thread here?

gerhardol: Yes (similar in RevDiff).

RussKie:
Merge at will.

PR #8557, opened 2020-10-20
Add paddings and borders for panels

ivangrek:
## Proposed changes

- Add paddings and borders for panels.
- Adds expressiveness and allows to clearly delineate the main panels.

### Before

image

### After

image
image
image

It would be nice to make the colors softer. Like this.

## Test methodology 

- Manual
- Visually

## Test environment(s) 

- GIT 
- Windows 

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8667, opened 2020-12-09
Left panel enhancement

ivangrek:
ref #8557

## Proposed changes

- Add textbox with ability to change border color
- Add toolbar to the left panel

### Before

image

image

### After

image

image

### Without "Search" label (discussion)

Maybe replace the button too.

image

image

## Test methodology 

- manually
- visually

## Test environment(s) 

- Git Extensions 3.4.3.9999
- Git 2.28.0.windows.1
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
@@ -99,7 +99,6 @@ namespace GitUI.BranchTreePanel
             this.tsbShowTags = new System.Windows.Forms.ToolStripButton();
             this.tsbShowSubmodules = new System.Windows.Forms.ToolStripButton();
             this.branchSearchPanel = new System.Windows.Forms.TableLayoutPanel();
-            this.lblSearchBranch = new System.Windows.Forms.Label();
             this.btnSearch = new System.Windows.Forms.Button();
             this.menuSettings = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.toolTip = new System.Windows.Forms.ToolTip(this.components);
@@ -515,7 +514,6 @@ namespace GitUI.BranchTreePanel
             this.branchSearchPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
             this.branchSearchPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
             this.branchSearchPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
-            this.branchSearchPanel.Controls.Add(this.lblSearchBranch, 0, 0);
             this.branchSearchPanel.Controls.Add(this.btnSearch, 2, 0);
             this.branchSearchPanel.Dock = System.Windows.Forms.DockStyle.Fill;
             this.branchSearchPanel.Location = new System.Drawing.Point(0, 25);
@@ -524,16 +522,6 @@ namespace GitUI.BranchTreePanel
             this.branchSearchPanel.RowStyles.Add(new System.Windows.Forms.RowStyle());
             this.branchSearchPanel.Size = new System.Drawing.Size(300, 26);
             this.branchSearchPanel.TabIndex = 4;
-            // 
-            // lblSearchBranch
-            // 
-            this.lblSearchBranch.Anchor = System.Windows.Forms.AnchorStyles.Left;
-            this.lblSearchBranch.AutoSize = true;
-            this.lblSearchBranch.Location = new System.Drawing.Point(3, 6);
-            this.lblSearchBranch.Name = "lblSearchBranch";
-            this.lblSearchBranch.Size = new System.Drawing.Size(44, 13);
-            this.lblSearchBranch.TabIndex = 0;
-            this.lblSearchBranch.Text = "Search:";
             // 
             // btnSearch
             // 
@@ -651,7 +639,6 @@ namespace GitUI.BranchTreePanel
         private ToolStripMenuItem mnubtnFetchOneBranch;
         private TableLayoutPanel repoTreePanel;
         private TableLayoutPanel branchSearchPanel;
-        private Label lblSearchBranch;
         private Button btnSearch;
         private ContextMenuStrip menuSettings;
         private ToolStripSeparator toolStripSeparator1;
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
@@ -60,12 +60,16 @@ namespace GitUI.BranchTreePanel
             this.tsmiMainMenuSpacer2 = new System.Windows.Forms.ToolStripSeparator();
             this.mnubtnCollapseAll = new System.Windows.Forms.ToolStripMenuItem();
             this.mnubtnExpandAll = new System.Windows.Forms.ToolStripMenuItem();
+            this.mnubtnMoveUp = new System.Windows.Forms.ToolStripMenuItem();
+            this.mnubtnMoveDown = new System.Windows.Forms.ToolStripMenuItem();
             this.menuBranch = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.mnubtnFilterLocalBranchInRevisionGrid = new System.Windows.Forms.ToolStripMenuItem();
+            this.runScriptToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.mnubtnFilterRemoteBranchInRevisionGrid = new System.Windows.Forms.ToolStripMenuItem();
             this.menuTags = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.menuRemotes = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.mnuBtnManageRemotesFromRootNode = new System.Windows.Forms.ToolStripMenuItem();
+            this.mnuBtnFetchAllRemotes = new System.Windows.Forms.ToolStripMenuItem();
             this.mnuBtnPruneAllRemotes = new System.Windows.Forms.ToolStripMenuItem();
             this.menuRemote = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.mnubtnRemoteBranchFetchAndCheckout = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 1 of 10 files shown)


RussKie:
Relates to https://github.com/gitextensions/gitextensions/issues/4814.

A big question - would we plan to add any more functionality to the new panel? For just two buttons it looks kind of like a wasted space.

GitUI/Win32/MessageCode.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace GitUI.Win32
+{
+    public enum MessageCode : int
```
RussKie, line 3: Win32 messages live here: https://github.com/gitextensions/gitextensions/blob/master/GitUI/Interops/Messages.cs

GitUI/Win32/NativeMethod.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace GitUI.Win32
+{
+    internal static class NativeMethod
```
RussKie, line 6: Move to https://github.com/gitextensions/gitextensions/tree/master/GitUI/Interops

RussKie (changes requested):

ivangrek:
> A big question - would we plan to add any more functionality to the new panel? For just two buttons it looks kind of like a wasted space.

A place for a flight of fantasy. 😄 
You might also wonder why they are not in the context menu.

GitUI/UserControls/ExTextBox.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            get => _borderFocusedColor;
+
+            set
+            {
+                _borderFocusedColor = value;
+
+                Invalidate();
+            }
+        }
```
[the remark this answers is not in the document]
ivangrek: Ok, I'm trying to separate blocks of code and variable assignments by one line.

RussKie:
> A place for a flight of fantasy. 😄
> You might also wonder why they are not in the context menu.

Haha, yes. But on a serious note, what would be a good idea to add to this menu? @gitextensions/git-extensions-team 

As for the contextmenu - this requires extra actions (like a right click), and some actions like "expand/collapse all" are not _really_ relevant on individual items, and are better placed at common locations such as this toolpanel.

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -77,50 +81,45 @@ private void InitializeComponent()
...
             this.mnuBtnPruneAllBranchesFromARemote = new System.Windows.Forms.ToolStripMenuItem();
+            this.mnuBtnOpenRemoteUrlInBrowser = new System.Windows.Forms.ToolStripMenuItem();
             this.repoTreePanel = new System.Windows.Forms.TableLayoutPanel();
+            this.leftPanelToolStrip = new GitUI.ToolStripEx();
+            this.collapseAllToolStripButton = new System.Windows.Forms.ToolStripButton();
+            this.settingsToolStripDropDownButton = new System.Windows.Forms.ToolStripDropDownButton();
+            this.showBranchesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.showRemotesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.showTagsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.showSubmodulesToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
```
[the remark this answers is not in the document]
ivangrek: Are these prefixes so important? Microsoft does not recommend using them.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions?redirectedfrom=MSDN

mast-eu:
> what would be a good idea to add to this menu?

A useful feature would be the ability to change the sorting order, especially for Tags.
When you have tagged many versions following semantic versioning, it can easily happen that you have to scroll down to see the latest versions' tags. I need, however, much more often quick access to the latest versions, rather than the first ones. Inverting the sorting would show the most recent version on top.

vbjay:
I like martin's idea. To add to it. It would be nice if possible to sort by
how far down the tag is( tag nearest head first, next nearest....).  If
this is not possible then by date of commit. Although date isn't reliable
with rebases involved.

RussKie:
E.g. expose this functionality in the toolbar?
image

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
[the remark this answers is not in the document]
RussKie: Yes. Firstly for consistency purposes. Secondly, when a form/control has a lot of child controls having controls grouped by prefix makes it significantly easier to locate the necessary control. And reason about control type when looking through the list.

GitUI/UserControls/ExTextBox.cs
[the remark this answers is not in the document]
RussKie: I don't think in this context "grouping" matters, but creates a lot of dead space, making harder to grasp the code.

ivangrek:
These buttons.

image

RussKie:
Great, looks like we can fill the toolbar ;)
Let's fix nits, and get this merged.

gerhardol:
I still prefer to not have an extra row. Remove the Search label and a few more buttons can fit...

ivangrek:
> I still prefer to not have an extra row. Remove the Search label and a few more buttons can fit...

Tried to follow the UX. In the future, this will make it very easy to add new buttons.

image

We also have our own UX. I think we will come to this someday.

image

ivangrek:
image

image

pmiossec:
With a placehoder text it would have been perfect. But perhaps we would have to wait after the .net core migration to have it...

gerhardol:
> With a placehoder text it would have been perfect. But perhaps we would have to wait after the .net core migration to have it...

We have that in many other boxes already, should be possible.
But this is the same as right now for me, the "Search" label do not add any benefit to the user, there is a search button to the right

pmiossec:
>But this is the same as right now for me, the "Search" label do not add any benefit to the user, there is a search button to the right

It doesn't add great value, I could admit but for me it improve the UX by letting the user know clearly what is the operation that will be done.
When I see an textbox like that I always wonder if it will do a filter or a search.

And I think that in term of UX, it's better to have a reminder when you enter the field instead than having to analyze the icon and  hoped that the icon has been well chosen...

RussKie:
> 
> image
> 
> image

I think I like these more. it aligns with VS design. And this layout is more open to hosting more buttons (that we discussed above).
My 2c.

RussKie:
Sent too early.

This design is also consistent with other forms we have, e.g.:
image

ivangrek:
> This design is also consistent with other forms we have, e.g.

Yes. I talked about this. Search field can be replaced for consistency UI.
At this stage, both options suit me.

Also fixed this case.

image

image

RussKie:
> > image
> > image
> 


I propose we keep this implementation

ivangrek:
🚀

pmiossec:
> I propose we keep this implementation

 Are you sure you don't want to keep all on one line until some new icons are added? 😁 (because it looks empty and I find it weird in terms of UI... )

gerhardol:
> Are you sure you don't want to keep all on one line until some new icons are added? 😁 (because it looks empty and I find it weird in terms of UI... )

I agree, wasted space

RussKie:
> Are you sure you don't want to keep all on one line until some new icons are added? 😁 (because it looks empty and I find it weird in terms of UI... )

Yes, per the discussion above - there are a number of controls that can/should be placed on the toolbar, e.g.
* https://github.com/gitextensions/gitextensions/pull/8667#issuecomment-742811431
* https://github.com/gitextensions/gitextensions/pull/8667#issuecomment-743000461

ivangrek:
Roll back the last commit?

ivangrek:
image

image

GitUI/BranchTreePanel/RepoObjectsTree.SettingsContextMenu.cs
```diff
@@ -131,16 +133,17 @@ private void ShowEnabledTrees()
...
 
-            if (tsmiShowSubmodules.Checked)
+            if (tsmiShowSubmodule.Checked)
             {
                 AddTree(_submoduleTree);
             }
         }
 
-        private void tsmiShowBranches_Click(object sender, EventArgs e)
+        private void showBranchesToolStripMenuItem_Click(object sender, EventArgs e)
```
RussKie, line 142: Event handlers names must match controls

RussKie:
> image
> image

This has move vote as it allows us to expose more functionality in the toolbar and is consistent with the FormCommit. Judging by comments @mast-eu and @vbjay lean this way as well. 

@gerhardol and @pmiossec I appreciate you are not very enthusiastic about this layout because right now it has empty space (so am I). Once we fill the space, how would you feel about it?

I'd be interested in opinions from @mstv and @NikolayXHD.

gerhardol:
> @gerhardol and @pmiossec I appreciate you are not very enthusiastic about this layout because right now it has empty space (so am I). Once we fill the space, how would you feel about it?

Fine when we fill the space, just right now it looks weird (up to 4 buttons should be fine, after that they need a separate row)
You have the final decision....

RussKie:
In the interest of reducing the code churn and incentify us to fill in the glaring void let's move with "the toolbar above the search bar" option.

mstv:
I have no strong opinion. Since more and more functionality is added to the Left Panel, this seems to be a good preparation.

ivangrek:
https://user-images.githubusercontent.com/3169265/102894773-49c90680-446c-11eb-921c-37f00e2dc847.mp4

ivangrek:
🚀

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -337,74 +364,151 @@ private void InitializeComponent()
...
+            this.leftPanelToolStrip.ClickThrough = true;
+            this.leftPanelToolStrip.GripStyle = System.Windows.Forms.ToolStripGripStyle.Hidden;
+            this.leftPanelToolStrip.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.tsbCollapseAll,
+            this.tsbSettings});
+            this.leftPanelToolStrip.Location = new System.Drawing.Point(0, 0);
+            this.leftPanelToolStrip.Name = "leftPanelToolStrip";
+            this.leftPanelToolStrip.Size = new System.Drawing.Size(300, 25);
+            this.leftPanelToolStrip.TabIndex = 5;
+            this.leftPanelToolStrip.Text = "toolStrip1";
```
RussKie, line 450: Redundant

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -337,74 +364,151 @@ private void InitializeComponent()
...
+            // 
+            // tsbCollapseAll
+            // 
+            this.tsbCollapseAll.DisplayStyle = System.Windows.Forms.ToolStripItemDisplayStyle.Image;
+            this.tsbCollapseAll.Image = global::GitUI.Properties.Images.CollapseAll;
+            this.tsbCollapseAll.ImageTransparentColor = System.Drawing.Color.Magenta;
+            this.tsbCollapseAll.Name = "tsbCollapseAll";
+            this.tsbCollapseAll.Size = new System.Drawing.Size(23, 22);
+            this.tsbCollapseAll.Text = "toolStripButton1";
```
RussKie, line 459: ditto

GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
```diff
@@ -337,74 +364,151 @@ private void InitializeComponent()
...
+            this.tsbCollapseAll.Click += new System.EventHandler(this.btnCollapseAll_Click);
+            // 
+            // tsbSettings
+            // 
+            this.tsbSettings.DisplayStyle = System.Windows.Forms.ToolStripItemDisplayStyle.Image;
+            this.tsbSettings.DropDownItems.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.tsmiShowBranches,
+            this.tsmiShowRemotes,
+            this.tsmiShowTags,
+            this.tsmiShowSubmodules});
```
RussKie, line 469: I don't think we need this dropdown button, suggest removing it and placing the 4 buttons it contains directly on the toolbar

GitUI/Translation/English.xlf
```diff
@@ -8136,6 +8136,14 @@ To show all branches, right click the revision grid, select 'view' and then the
...
         <target />
       </trans-unit>
+      <trans-unit id="tsbCollapseAll.Text">
+        <source>toolStripButton1</source>
+        <target />
+      </trans-unit>
+      <trans-unit id="tsbSettings.Text">
+        <source>toolStripDropDownButton1</source>
+        <target />
+      </trans-unit>
```
RussKie, line 8146: None of these should be here

GitUI/Interops/NativeMethod.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Runtime.InteropServices;
```
RussKie, line 1: This must reside in https://github.com/gitextensions/gitextensions/tree/master/GitUI/Interops/User32 and follow the file and class naming schemes

GitUI/UserControls/ExTextBox.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System;
+using System.Drawing;
+using System.Windows.Forms;
+
+namespace GitUI.UserControls
+{
+    internal sealed class ExTextBox : TextBox
```
RussKie, line 7: We have an established naming scheme (for better or for worse):
```suggestion
    internal sealed class TextBoxEx : TextBox
```

RussKie (commented):
👍 

Please make up the history as well.

ivangrek:
image

image

RussKie (approved):

PR #8678, opened 2020-12-15
Improve handling of Drag and drop

### LINKED ISSUE #8669: D&D to commit graph change current tab to TreeTabPage
## Current behaviour

It's possible to D&D selected text from the diffViewer to the "commit graph" viewer.

After drop, current tab becomes TreeTab instead of the current tab: DiffViewer.

## Expected behaviour

I should not be able to drop text to the commit graph viewer.

When drop to commit graph, it should not change current tab.

## Steps to reproduce

Open Git Extensions
Open a repository
Click on a commit in the commit graph viewer.
Click on Diff tab
Select some text in the diff viewer
Drag this text and drop it to the commit graph viewer

## Screenshots

Not necessary

I don't think so.

## Environment

- Git Extensions 3.4.3.9999
- Git 2.22.0.windows.1 (recommended: 2.25.1 or later)
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

## Diagnostics

1. I think it should not be possible to drop to commit graph.
2. About the change of the current tab to Tree Tab Page : I think it's in the file ``gitextensions\GitUI\CommandsDialogs\FormBrowse.cs`` and the function ``private void FormBrowse_DragDrop(object sender, DragEventArgs e)``. In this function, there is a hardcoded default tab: ``CommitInfoTabControl.SelectedTab = TreeTabPage;``. The code is executed at every drop on the commit graph even if the TreeTabPage is not visible.

pmiossec:
Fixes #8669

## Proposed changes

* Don't try to select "File tree" if it is not displayed
* Switch to "File tree" after some content check (to prevent inopportune switch to the tab)

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build f5362ec637de01d40be55f5d0abfbbc7b6e9d828
- Git 2.28.0.windows.1
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3227,15 +3227,20 @@ namespace GitUI.CommandsDialogs
 
         private void FormBrowse_DragDrop(object sender, DragEventArgs e)
         {
-            CommitInfoTabControl.SelectedTab = TreeTabPage;
             HandleDrop(e);
         }
 
         private void HandleDrop(DragEventArgs e)
         {
-            var itemPath = (e.Data.GetData(DataFormats.Text) ?? e.Data.GetData(DataFormats.UnicodeText)) as string;
-            if (itemPath != null && (File.Exists(itemPath) || Directory.Exists(itemPath)))
+            if (TreeTabPage.Parent == null)
             {
+                return;
+            }
+
+            var itemPath = (e.Data.GetData(DataFormats.Text) ?? e.Data.GetData(DataFormats.UnicodeText)) as string;
+            if (IsFileExistingInRepo(itemPath))
+            {
+                CommitInfoTabControl.SelectedTab = TreeTabPage;
                 fileTree.SelectFileOrFolder(itemPath);
                 return;
             }
@@ -3248,11 +3253,25 @@ namespace GitUI.CommandsDialogs
 
             foreach (string path in paths)
             {
+                if (!IsFileExistingInRepo(path))
+                {
+                    continue;
+                }
+
+                if (CommitInfoTabControl.SelectedTab != TreeTabPage)
+                {
+                    CommitInfoTabControl.SelectedTab = TreeTabPage;
+                }
+
                 if (fileTree.SelectFileOrFolder(path))
                 {
                     return;
                 }
             }
+
+            bool IsPathExists(string path) => path != null && (File.Exists(path) || Directory.Exists(path));
+
+            bool IsFileExistingInRepo(string path) => IsPathExists(path) && path.StartsWith(Module.WorkingDir, StringComparison.InvariantCultureIgnoreCase);
         }
 
         private void FormBrowse_DragEnter(object sender, DragEventArgs e)


RussKie:
If this works, and addresses the issue - go ahead. 👍

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3217,15 +3217,20 @@ public void PopulateFavouriteRepositoriesMenu(ToolStripDropDownItem container, I
...
 
         private void HandleDrop(DragEventArgs e)
         {
+            if (TreeTabPage.Parent == null)
+            {
+                return;
+            }
+
             var itemPath = (e.Data.GetData(DataFormats.Text) ?? e.Data.GetData(DataFormats.UnicodeText)) as string;
             if (itemPath != null && (File.Exists(itemPath) || Directory.Exists(itemPath)))
```
gerhardol, line 3231: This check should probably be changed so it checks that the file is in this repo. Right now it switches to the Tree tab if the file exist in another repo

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3236,6 +3241,7 @@ private void HandleDrop(DragEventArgs e)
                 return;
             }
 
+            CommitInfoTabControl.SelectedTab = TreeTabPage;
```
gerhardol, line 3244: Similar to text drop, should check that the file is in this repo

gerhardol (commented):
I would like to have some kind of feedback if drag&drop cannot select a file, so the user can get a hint what it is used for

gerhardol:
An enhancement, to motivate rejecting files outside the rpo and maybe add a popup:
If a file is dropped on GE that is not in this repo, ask if that repo should be opened.
If not a file in repo, display error?
If not detected as a file, display popup

GitUI/CommandsDialogs/FormBrowse.cs
bansan85: Right, simply move this line to the few lines below, just after ``if (fileTree.SelectFileOrFolder(path)) {`` and before the ``return``

GitUI/CommandsDialogs/FormBrowse.cs
bansan85: Check the return value of fileTree.SelectFileOrFolder and change the SelectedTab if the return is true.

bansan85:
I think that the drag from the text viewer from "File Tree" should be disabled. I don't see what the point to drag a filename from a text to the commit graph to open a file.

It's the same problem for the diff viewer from "Diff". I think it "could" be fine to select the path in the beginning of the diff :
```
--- a/GitUI/CommandsDialogs/FormDeleteRemoteBranch.Designer.cs
+++ b/GitUI/CommandsDialogs/FormDeleteRemoteBranch.Designer.cs
```
but even if I drag "GitUI/CommandsDialogs/FormDeleteRemoteBranch.Designer.cs" to the commit graph, the function ``File.Exists`` in ``FormBrowse.HandleDrop`` fails.

So I think that drag from "File Tree Viewer" and "Diff Viewer" should be disabled.

What do you think about it ?

gerhardol:
>  think that the drag from the text viewer from "File Tree" should be disabled. I don't see what the point to drag a filename from a text to the commit graph to open a file.

This is intended to drag&drop text from GE for instance to a text editor

bansan85:
> 
> 
> > think that the drag from the text viewer from "File Tree" should be disabled. I don't see what the point to drag a filename from a text to the commit graph to open a file.
> 
> This is intended to drag&drop text from GE for instance to a text editor

Ok thanks. So keep it. Maybe it could be more user friendly to allow or not the drop during the DragEnter event based on the content of the DragEvent variable. But this new behavior will make impossible to show error message if drop fails.

gerhardol:
I am OK with the changes here, but would like to see the popups when failing (popup to open a repo if the file is outside the repo would be an enhancement). @bansan85  issue and subsequent questions show that this is not obvious how it should work.
@pmiossec This or a separate PR?

pmiossec:
I forgot this PR 🤣 

I have pushed an update. See if it still suits you. If yes, it could be merged...

>This or a separate PR?

I started to have a look at opening a new repo but it will required a lot more work (to find the exact path of the repo from the file path with cmd `git  rev-parse --show-toplevel`) not really linked to this PR.
So I think it will be in another one (with the warning popup)...

gerhardol (approved):
One suggestion, avoid switching to tree if the file is not in the tree (for this revision)

I do not believe that this really resolves #8669, that issue is really about what drag and drop is to be used for. Popup is what I propose.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3246,13 +3253,22 @@ private void HandleDrop(DragEventArgs e)
                 return;
             }
 
-            foreach (string path in paths)
+            foreach (string path in paths.Where(IsFileExistingInRepo))
```
bansan85, line 3256: Are you sure this is needed ? fileTree.SelectFileOrFolder already do a ``filePath.IndexOf(Module.WorkingDir) != 0``

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3246,13 +3253,22 @@ private void HandleDrop(DragEventArgs e)
...
             {
+                if (CommitInfoTabControl.SelectedTab != TreeTabPage)
```
bansan85, line 3258: Why check before modification ? According to [TabControl.cs](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/TabControl.cs,737), the .net framework already do it.

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: Avoid LINQ where possible, it has perf overhead. We're already traversing the collection, so just check in each iteration.

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: No it does check here: https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/TabControl.cs,795, nor here: https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/TabControl.cs,811. So this checks avoid executing unnecessary operations.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3246,13 +3253,22 @@ private void HandleDrop(DragEventArgs e)
...
+                {
+                    CommitInfoTabControl.SelectedTab = TreeTabPage;
+                }
+
                 if (fileTree.SelectFileOrFolder(path))
                 {
                     return;
                 }
```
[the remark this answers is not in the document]
pmiossec: I don't think that will work because `fileTree.SelectFileOrFolder(path)` doesn't select the file if the file tree is not loaded so you have to select the tab before.

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: @bansan85 yes, it is needed to filter before trying to select it otherwise there are case where you will select the tabs even of the content dropped has nothing to do with file paths belonging to this repository (something you complain in your issue...)

pmiossec: @RussKie I will remove the use of LINQ...

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: @bansan85 I had a quick look at the decompiled code and I saw that there are some other things done before checking if the tab is already the one selected, as @RussKie pointed out, so I decided to do the test to avoid unnecessary operations.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3227,15 +3227,22 @@ public void PopulateFavouriteRepositoriesMenu(ToolStripDropDownItem container, I
...
         }
 
         private void HandleDrop(DragEventArgs e)
         {
+            if (TreeTabPage.Parent == null)
+            {
+                return;
+            }
+
+            var repoDir = Module.WorkingDir;
```
RussKie, line 3240: This is only used in `IsFileExistingInRepo`, so probably worth inlining.

RussKie (commented):
The code LGTM. :shipit:

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: Done.

One of the reason (and after refactoring the only remaining one) why I created this variable is to have a better naming.
I hate this `WorkingDir` name which I find too unclear.
Can I do a PR to rename it `RepoPath` (or `RepositoryPath`)?

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
gerhardol: It did work when I tried it, but the comment above indicates that there are other situations too

gerhardol (approved):

RussKie:
> Can I do a PR to rename it `RepoPath` (or `RepositoryPath`)?

I'm open to a change. Please open a new issue where we all can discuss. It will be more productive than you making a change and then we all arguing about the name :)

gerhardol:
I do not believe that this PR addressed the reason for #8669, explaining why d&d is supported
It is an improvement though

PR #8679, opened 2020-12-16
Don't persist or restore the state of the main menu

### LINKED ISSUE #8636: Main menu items permuted randomly
## Current behaviour

AppVeyor build of master:

grafik

AppVeyor build of PR:

grafik

Older local build:

grafik

Older local build restarted:

grafik

## Expected behaviour

at least no random

Yes, I guess with my own local release build before the "older local build". I don't have access to the PC running the latter ATM.

RussKie:
Resolves #8636


mstv (approved):
This workaround works. And it is not too intrusive I think.

gerhardol (approved):
Seem to work
When applying this PR on my private builsd I get a crash during startup, but I currently say that is something else

PR #8682, opened 2020-12-16
A variety of accessibility-related fixes for the contents of the main pane on the Config page 

guybark:
1. Update app.config to pick up some .NET Framework accessibility fixes up to 4.8, depending on what version of .NET Framework is available on the device at runtime.

2. Add unique accessible names for the Browse and Suggest buttons to avoid duplicate names being exposed programmatically on sibling buttons. (If the UI is ever localized, the accessible names must be localized too.)

3. Change the order in which controls are added to their containers, in order to have the programmatic order (as exposed through the UI Automation API) of the controls match their logical order.

4. Remove the TabIndex property being set on the controls, because that interferes with WinForms setting accessible names on TextBox and ComboBox controls based on the text from the Labels that programmatically precedes them. The default tab order is now based on the order in which the controls are added to their containers.

### Before

The Accessibility Insights for Windows tool reports 31 failures against the UI, as shown in the following image

AIWinBefore

### After

The Accessibility Insights for Windows tool reports 7 failures against the UI, as shown in the following image. Note that today the remaining failures cannot be avoided while building for .NET Framework.

AIWinAfter

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
@@ -80,7 +80,6 @@
             this.label60.Location = new System.Drawing.Point(3, 417);
             this.label60.Name = "label60";
             this.label60.Size = new System.Drawing.Size(114, 13);
-            this.label60.TabIndex = 79;
             this.label60.Text = "Files content encoding";
             // 
             // Global_FilesEncoding
@@ -90,15 +89,14 @@
             this.Global_FilesEncoding.Location = new System.Drawing.Point(129, 413);
             this.Global_FilesEncoding.Name = "Global_FilesEncoding";
             this.Global_FilesEncoding.Size = new System.Drawing.Size(241, 21);
-            this.Global_FilesEncoding.TabIndex = 17;
             // 
             // btnCommitTemplateBrowse
             // 
+            this.btnCommitTemplateBrowse.AccessibleName = "Browse Path to commit template";
             this.btnCommitTemplateBrowse.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Left | System.Windows.Forms.AnchorStyles.Right)));
             this.btnCommitTemplateBrowse.Location = new System.Drawing.Point(1037, 260);
             this.btnCommitTemplateBrowse.Name = "btnCommitTemplateBrowse";
             this.btnCommitTemplateBrowse.Size = new System.Drawing.Size(123, 25);
-            this.btnCommitTemplateBrowse.TabIndex = 15;
             this.btnCommitTemplateBrowse.Text = "Browse";
             this.btnCommitTemplateBrowse.UseVisualStyleBackColor = true;
             this.btnCommitTemplateBrowse.Click += new System.EventHandler(this.btnCommitTemplateBrowse_Click);
@@ -110,7 +108,6 @@
             this.lblCommitTemplatePath.Location = new System.Drawing.Point(3, 266);
             this.lblCommitTemplatePath.Name = "lblCommitTemplatePath";
             this.lblCommitTemplatePath.Size = new System.Drawing.Size(120, 13);
-            this.lblCommitTemplatePath.TabIndex = 76;
             this.lblCommitTemplatePath.Text = "Path to commit template";
             // 
             // txtCommitTemplatePath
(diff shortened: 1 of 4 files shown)


gerhardol:
can you please sign off in contributors.txt?

This is just one form - is it possible to make an estimate of what it would take to apply to all?

vbjay:
Also try adding a controll and seeing if it breaks things.  This is why we
have tabindex. Do it needs to be tested well. Try removing a control and
then adding it again.

mstv (commented):
LGTM, have not run it.

RussKie (commented):
This is interesting, so we don't need tabindex to be accessible? Seems counter-intuitive...

vbjay:
I can see it breaking when someone adds a control or removes a control and
then adds a control.

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
```diff
@@ -90,15 +89,14 @@ private void InitializeComponent()
             this.Global_FilesEncoding.Location = new System.Drawing.Point(129, 413);
             this.Global_FilesEncoding.Name = "Global_FilesEncoding";
             this.Global_FilesEncoding.Size = new System.Drawing.Size(241, 21);
-            this.Global_FilesEncoding.TabIndex = 17;
             // 
             // btnCommitTemplateBrowse
             // 
+            this.btnCommitTemplateBrowse.AccessibleName = "Browse Path to commit template";
```
gerhardol, line 95: These strings should probably be translated.
The translation app is not picking this up right now (this is the first use of AccessibleName in this project).
Short term, translation strings could be created separately (but how hard can it be to fix the translationapp, to make sure the build fails if not translated?)

gerhardol (commented):
Tab order the same as before

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
gerhardol: AccessibleName (and AccessibleDescription) added to the translation app in #8713.
@guybark any comments on the items to translate?
#8713 should be merged soon, then this PR should be rebased on master and translation run once more.

gerhardol:
Can you please rebase on master, you will have to run translation too:
https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions#update-english-translations

cla too: https://github.com/gitextensions/gitextensions/blob/master/contributors.txt

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
guybark: That's a good point, thanks for raising it. The AccessibleName properties needs to be as translated as visual text labels are. (So by default, translated, except for things like branding.) The AccessibleName is what gets announced by screen reader for customers who are blind or have low vision, so needs to be in a language that the customer wants. I hard-coded the strings because in the same file there seemed to be hard-coded Text properties, and I assumed the app wasn't localized. It sounds like I made the wrong assumption. If the AccessibleName properties won't be translated, the change should be canceled until it's possible to have the strings translated.

guybark: Regarding the removal of the use of TabIndex, WinForms apps are vulnerable to both the tab order and the programmatic order of the UI not being a good match for the use of the UI. The tab order is relatively straightforward to fix in VS through use of its "Tab Order" feature (in the View menu) but that doesn't fix the programmatic order. The only way to fix the programmatic order is to ensure that the calls to Add() in the designer.cs file match the logical order of the UI. The existing TabIndex use overrides that, so if the TabIndex use is removed the Add() order will dictate the tab order and programmatic order. If the form is later opened up in the VS designer, the TabIndex use will be added back but in an order that matches the Add() order, so the tab order and programmatic order remain good. A point was raised above that things may break later when controls are added or removed, and that's right. All WinForms apps are vulnerable, and the only way for things to work well for customers is for app builders to include the maintenance of the tab order and programmatic order in the list of things they do as their app evolves over time.

mstv:
Could you give your signoff to the contributors list in a dedicated PR, please, in order to avoid merge conflicts by merging it quickly?
We want to avoid merges into feature branches. So rebase or cherry-pick the main commit and force-push this branch, please.

GitUI/CommandsDialogs/SettingsDialog/Pages/GitConfigSettingsPage.Designer.cs
gerhardol: AccessibleName will be translated when you rebase, the TranslationApp will pick up this field and add it to the source English.xlf, translated on Transifex.
https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions#update-english-translations

So for order: The easiest way around seem to be to just make sure that there is no TabIndex

RussKie (approved):

RussKie:
Thank you

guybark:
I've run the built artifacts and verified that the changes are good. The image below shows the Accessibility Insights for Windows tool reporting that all the controls in the main area of this particular config UI have appropriate UIA Name properties, and they appear in the logical order in the UIA hierarchy. I manually verified that when tabbing through the controls, keyboard focus moves in the expected order.

GitExtensions_Config_UIAHierarchy

RussKie:
Thank you!

PR #8687, opened 2020-12-19
GitStatusMonitor: Display when inactive

gerhardol:
## Proposed changes

Display when the commit count is inactive or not yet updated.
Currently the status is shown as zero changes or shown the changes before for instance Commit or minimize.
Update is scheduled immediately after running again.
- When refreshing
- After switching to a new repo
- Minimizing GE, then activating
- Open a modal dialog, like FormCommit

@mstv has asked about about status being incorrect, for instance when committing. 
The modal dialog ("lock the repo") could still keep GitStatusMonitor active, even if I believe I prefer this.

--
A slightly separate change that builds on the change above, could be separated
@pmiossec  asked about this

Run first update with locks
GitStatusMonitor normally do not try to lock the git-status, to not interfere with Commit.
However, a lock will speed up subsequent git-status.
If for instance you are building while opening GE, a lot of file changed events will fire
and git-status will run as often as allowed (max 1/3 of the time, but git-status can take some seconds).

If Commit button is pressed while still gray there may be a conflict though.
(But next FormCommit git-status should be faster.)

### Before

Shown while operations like Commit are ongoing
image

Shown after a repo is opened, before status is available
image

### After

Until git-status has run
image

## Test methodology

At work with Symantec virus git-status may require 10s (at least the initial updated.
On my PC at home, git-status requires about 200 ms for the GE repo. This change can be hard to see. To simulate a slow git-status execution, you may simulate a delay:

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs line 437
Thread.Delay(5000);

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -281,6 +281,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                     oldCommands.PreCheckoutRevision -= GitUICommands_PreCheckout;
                     oldCommands.PostCheckoutBranch -= GitUICommands_PostCheckout;
                     oldCommands.PostCheckoutRevision -= GitUICommands_PostCheckout;
+                    oldCommands.PostRepositoryChanged -= GitUICommands_PostRepositoryChanged;
                 }
 
                 commandsSource_activate(sender as IGitUICommandsSource);
@@ -293,6 +294,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                 newCommands.PreCheckoutRevision += GitUICommands_PreCheckout;
                 newCommands.PostCheckoutBranch += GitUICommands_PostCheckout;
                 newCommands.PostCheckoutRevision += GitUICommands_PostCheckout;
+                newCommands.PostRepositoryChanged += GitUICommands_PostRepositoryChanged;
 
                 var module = newCommands.Module;
                 StartWatchingChanges(module.WorkingDir, module.WorkingDirGitDir);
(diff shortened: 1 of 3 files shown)


RussKie (approved):

gerhardol:
@gitextensions/git-extensions-team comments (especially) on the functionality change?

RussKie:
@msftbot merge in 72 hours

RussKie:
If someone is deeply concerned they have 72 hours to voice their concerns. If not, we can revisit this later.
Thank you.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -263,6 +283,7 @@ void commandsSource_GitUICommandsChanged(object sender, GitUICommandsChangedEven
                     oldCommands.PreCheckoutRevision -= GitUICommands_PreCheckout;
                     oldCommands.PostCheckoutBranch -= GitUICommands_PostCheckout;
                     oldCommands.PostCheckoutRevision -= GitUICommands_PostCheckout;
+                    oldCommands.PostRepositoryChanged -= GitUICommands_PreRepositoryChanged;
```
mstv, line 286: I have not digged into the code. But where is `+= GitUICommands_PreRepositoryChanged`?

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -104,6 +105,7 @@ public GitStatusMonitor(IGitUICommandsSource commandsSource)
             Init(commandsSource);
 
             _getAllChangedFilesOutputParser = new GetAllChangedFilesOutputParser(() => commandsSource.UICommands.GitModule);
+            _isFirstPostRepoChanged = false;
```
mstv, line 108: Can be omitted because it is the default value, isn't it?

gerhardol: Yes, it was added for clarity but I will remove it

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: GitUICommands_PreRepositoryChanged is never executed because there is no PreRepositoryChanged event...
Fixed the handling

gerhardol:
> If someone is deeply concerned they have 72 hours to voice their concerns. If not, we can revisit this later.

That is what I believe as well. Try to avoid rework, but it is better to go forward if in doubt
Since it was a functionality change I do not mind a few days extra to merge
This time mstv and pmiossec had raised comments, so I wanted them to join.
Will squash fixup! and rebase tomorrow or so.

PR #8688, opened 2020-12-20
Move NetSpell.SpellChecker to Externals

RussKie:
We haven't authored NetSpell.SpellChecker project rather have copied
it presumably from https://github.com/loresoft/NetSpell in
fff5303cb2075a11666648f592b5c659421e1bab.

It would have been been a better idea to fork the original project, and
send patches upstream.
However the original project appears to be dormant, and its NuGet package
has virtually no consumption, so we're stuck with our copy of the codebase.

Move to Externals folder to better reflect the ownership model.


gerhardol (approved):
Let us settle with this change.
A subtree could be used too, but let us avoid introducing that concept. A submodule is possible too.

RussKie:
I'm cherrypicking changes from .NET 5.0 branch to reduce noise on it.

I've briefly scanned through our changed code, and then through the code in the original repo, I don't think bringing in a submodule would be of any value to us - the upstream repo hasn't changed in 3 years, and on NuGet it had 57K downloads since 2014...

PR #8697, opened 2020-12-23
Handle multi-part names for author initials

gerhardol:
## Proposed changes

Tune the author initials to handle middle names and additional info in the name and email addresses.
See test cases for details.

I initially wanted it to handle "Lastname Firstname" as "FL", but that is too hard to find out.
The French convention to use uppercase LASTNAME could be useful to detect lastnames...

## Test methodology 

Tests expanded and updated.

----

### DIFF
--- a/GitUI/Avatars/InitialsAvatarGenerator.cs
+++ b/GitUI/Avatars/InitialsAvatarGenerator.cs
@@ -32,33 +32,59 @@ namespace GitUI.Avatars
 
         protected internal (string initials, int hashCode) GetInitialsAndHashCode(string email, string name)
         {
-            string initials;
-            int hashCode;
+            string initials = null;
+            int hashCode = _unkownCounter;
             if (!string.IsNullOrWhiteSpace(name))
             {
-                var trimmedName = name.Trim();
-                var indexSpace = trimmedName.IndexOf(' ');
-                initials = trimmedName[0].ToString().ToUpper();
-                if (indexSpace != -1)
+                initials = GetInitials(name.Trim().Split());
+                if (!string.IsNullOrWhiteSpace(initials))
                 {
-                    initials += trimmedName[indexSpace + 1].ToString().ToUpper();
+                    hashCode = name.GetHashCode();
+                    return (initials, hashCode);
+                }
+            }
+
+            if (!string.IsNullOrWhiteSpace(email))
+            {
+                var e = email.TrimStart();
+                if (e.Contains("@"))
+                {
+                    e = e.Split('@')[0];
                 }
 
-                hashCode = string.IsNullOrWhiteSpace(email) ? name.GetHashCode() : email.GetHashCode();
-            }
-            else if (!string.IsNullOrWhiteSpace(email))
-            {
-                initials = ("" + email.TrimStart().First()).ToUpper();
-                hashCode = email.GetHashCode();
-            }
-            else
-            {
-                initials = "?";
-                hashCode = _unkownCounter;
-                _unkownCounter++;
+                initials = GetInitials(e.Split('.', '-', '_'));
+                if (!string.IsNullOrWhiteSpace(initials))
+                {
+                    hashCode = email.GetHashCode();
+                    return (initials, hashCode);
+                }
             }
 
+            initials = "?";
+            hashCode = _unkownCounter;
+            _unkownCounter++;
+
             return (initials, hashCode);
+
+            static string GetInitials(string[] names)
+            {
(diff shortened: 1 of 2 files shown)


UnitTests/GitUI.Tests/Avatars/InitialsAvatarGeneratorTests.cs
```diff
@@ -10,36 +10,22 @@ namespace GitUITests.Avatars
...
-        public void GetInitialsAndHashCode_return_initials_of_a_user()
+        [TestCase("albert.einstein@noreply.com", "albert einstein", "AE")]
+        [TestCase("albert.einstein@noreply.com", "albert z middlename einstein (k)", "AE")]
+        [TestCase("albert.einstein@noreply.com", "albert einstein (Nobel_price_winner)", "AE")]
+        [TestCase("albert.einstein@noreply.com", "EINSTEIN, Albert middlename (Nobel_price_winner)", "EM")]
+        [TestCase("albert.einstein@noreply.com", "albert", "AL")]
+        [TestCase("albert.einstein@noreply.com", "", "AE")]
+        [TestCase("albert.einstein@noreply.com", null, "AE")]
+        [TestCase("albert.z.einstein@noreply.com", null, "AE")]
+        [TestCase("albert", "", "AL")]
```
pmiossec, line 21: Perhaps in this case, it's better to have the 2nd letter lowercase to indicate that it was not the first letter of a name...

GitUI/Avatars/InitialsAvatarGenerator.cs
```diff
@@ -32,33 +32,59 @@ public Image GenerateAvatarImage(string email, string name, int imageSize)
...
-                initials = "?";
-                hashCode = _unkownCounter;
-                _unkownCounter++;
+                var e = email.TrimStart();
+                if (e.Contains("@"))
+                {
+                    e = e.Split('@')[0];
+                }
+
+                initials = GetInitials(e.Split('.'));
```
pmiossec, line 55: Should we split only on '.'? or should we also support '_' and '-' (and probably some others)?

pmiossec:
> The French convention to use uppercase LASTNAME could be useful to detect lastnames...

I'm not sure we use it always...

pmiossec (approved):

GitUI/Avatars/InitialsAvatarGenerator.cs
gerhardol: I have seen uses like abc_sd.fgfhhf@ and thought that was inconsistent. But it is maybe better to use first, last also here?
Will change to .-_

gerhardol:
> 
> 
> > The French convention to use uppercase LASTNAME could be useful to detect lastnames...
> 
> I'm not sure we use it always...

With all different nationalities at the workplace, I wish we used it here too. Then I would add heuristics to detect the last name. But I do not have that need or knowledge right now.

GitUI/Avatars/InitialsAvatarGenerator.cs
gerhardol: updated. This is a fallback case, most users have some kind of  name set. The heuristics to select the initials from the name couls possibly be improved.

gerhardol:
@msftbot merge in 3 days

UnitTests/GitUI.Tests/Avatars/InitialsAvatarGeneratorTests.cs
```diff
@@ -10,36 +10,24 @@ namespace GitUITests.Avatars
     [TestFixture]
     public sealed class InitialsAvatarGeneratorTests
     {
-        [Test]
-        public void GetInitialsAndHashCode_return_initials_of_a_user()
+        [TestCase("albert.einstein@noreply.com", "albert einstein", "AE")]
+        [TestCase("albert.einstein@noreply.com", "albert z middlename einstein (k)", "AE")]
+        [TestCase("albert.einstein@noreply.com", "albert einstein (Nobel_price_winner)", "AE")]
+        [TestCase("albert.einstein@noreply.com", "EINSTEIN, Albert middlename (Nobel_price_winner)", "EM")]
```
RussKie, line 16: I'd expect the initials to be "EAM"... 🤔

gerhardol: More than two looks weird and some international names have up to five parts, where to draw the limit? But it would still be something to identify them with. I chose esthetics over functionality.

PR #8700, opened 2020-12-26
Custom difftool in RevDiff, FileHistory, Commit, RevGrid

gerhardol:
Followup to #8194 (and #8660)

## Proposed changes

Make all configured difftools available in more situations

The default behavior is not changed, it is still possible to click the
dropdown "parent" item to enable the functionality.

The dropdown is not shown by default for entries existing in GE3.4 to not
confuse users and to show the shortcut (that is hidden when a dropdown
is enabled).
The dropdown is toggled with a rightclick.
A tooltip is added to enable discovery.

Note that for FormFileHistory the F3 shortcut is disabled as the
shortcut is not handled as a hotkey and a dropdown disables the shortcut.

Show the default tool first and checked in the dropdown.
This is also changed for FormResolveConflicts.

Remove the separate custom difftool menu in RevDiff.

Some notes

* ~~Submodule diff for Commit Unstaged is unchanged. Comparing commit sha in a difftool is quite meaningless (same as inFileViewer) and should be replaced with a diff _within_ the submodule (like update, reset).~~
* I would have preferred to enable this functionality by default, but there will be users that will be annoyed by the extra dropdowns.
* FormFileHistory could allow hotkeys to support F3 also when dropdown is shown. 
* It could be possible to show the shortcut key by overriding the ToolStripMenuItem (something like https://stackoverflow.com/questions/30391181/positioning-of-shortcut-key-of-toolstripmenu-item-properly). Not worth it.
* It takes some time to get the custom tools, 10s on my private PC, more than double that on corporate PCs with viruses like Symantec. RevDiff is not really an issue, the list is loaded at startup. For Commit, History and Resolve the functionality would be improved if the lists were cached.

## Screenshots 

All screenshots except where noted) done after custom tools are loaded

Marking additions since 3.4

Before|After
---|---
image|image
File remembered (similar if two files had been selected)|
image|image
Similar for Index (Staged). Submodule is handled in a separate menu for worktree.|diff submodule removed, F3 still possible (in other situations the normal menu is used, difftool is still available)
image|image
image|image
image|image
image|image
image|image

After: Example where tools are not yet loaded, no dropdown menu yet

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -1,0 +1,147 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GitCommands.Config;
+using GitUI;
+using Microsoft.VisualStudio.Threading;
+
+namespace GitCommands
+{
+    public class CustomDiffMergeToolCache
+    {
+        private CustomDiffMergeToolCache(bool isDiff)
+        {
+            _isDiff = isDiff;
+        }
+
+        private bool _isDiff { get; set; }
+        private SemaphoreSlim _mutex = new(1);
+        private IEnumerable<string>? _tools = null;
+
+        public static CustomDiffMergeToolCache DiffToolCache { get; } = new(true);
+        public static CustomDiffMergeToolCache MergeToolCache { get; } = new(false);
+
+        /// <summary>
+        /// Clear the existing caches
+        /// </summary>
+        public void Clear()
+        {
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await _mutex.WaitAsync().ConfigureAwait(false);
+
+                try
+                {
+                    _tools = null;
+                }
+                finally
+                {
+                    _mutex.Release();
+                }
+            }).FileAndForget();
+        }
+
+        /// <summary>
+        /// Load the availble DiffMerge tools and apply to the menus
+        /// </summary>
+        /// <param name="module">The Git module</param>
+        /// <param name="delay">The delay before starting the operation</param>
+        public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay)
+        {
+            if (_tools is not null)
+            {
+                return _tools;
+            }
+
+            // The command will compete with other resources, avoid delaying startup
+            // Do not bother with cancel tokens
+            await Task.Delay(delay);
+
+            await _mutex.WaitAsync().ConfigureAwait(false);
+            try
+            {
+                if (_tools is null)
+                {
+                    await TaskScheduler.Default;
+                    var toolKey = _isDiff ? SettingKeyString.DiffToolKey : SettingKeyString.MergeToolKey;
+                    var defaultTool = module.GetEffectiveSetting(toolKey);
+                    string output = module.GetCustomDiffMergeTools(_isDiff);
+                    _tools = ParseCustomDiffMergeTool(output, defaultTool);
+                }
+            }
+            finally
+            {
+                _mutex.Release();
+            }
+
+            return _tools;
+        }
+
+        /// <summary>
+        /// Parse the output from 'git difftool --tool-help'.
+        /// </summary>
+        /// <param name="output">The output string.</param>
+        /// <returns>list with tool names.</returns>
+        private IEnumerable<string> ParseCustomDiffMergeTool(string output, string defaultTool)
+        {
+            var tools = new List<string>();
+
+            // Simple parsing of the textual output opposite to porcelain format
+            // https://github.com/git/git/blob/main/git-mergetool--lib.sh#L298
+            // An alternative is to parse "git config --get-regexp difftool'\..*\.cmd'" and see show_tool_names()
+
+            // The sections to parse in the text has a 'header', then break parsing at first non match
+
+            foreach (var l in output.Split('\n'))
+            {
+                if (l == "The following tools are valid, but not currently available:")
+                {
+                    // No more usable tools
+                    break;
+                }
+
+                if (!l.StartsWith("\t\t"))
+                {
+                    continue;
+                }
+
+                // two tabs, then toolname, cmd (if split in 3) in second
+                // cmd is unreliable for diff and not needed but could be used for mergetool special handling
+                string[] delimit = { " ", ".cmd" };
+                var tool = l.Substring(2).Split(delimit, 2, StringSplitOptions.None);
+                if (tool.Length == 0)
+                {
+                    continue;
+                }
+
+                // Ignore (known) tools that must run in a terminal
+                string[] ignoredTools = { "vimdiff", "vimdiff2", "vimdiff3" };
+                var toolName = tool[0];
+                if (!string.IsNullOrWhiteSpace(toolName) && !tools.Contains(toolName) && !ignoredTools.Contains(toolName))
+                {
+                    tools.Add(toolName);
+                }
+            }
+
+            // Return the default tool first
+            return tools.OrderBy(tool => tool != defaultTool).ThenBy(tool => tool);
+        }
+
+        internal TestAccessor GetTestAccessor() => new(this);
+
+        internal readonly struct TestAccessor
+        {
+            private readonly CustomDiffMergeToolCache _cache;
+
+            public TestAccessor(CustomDiffMergeToolCache cache)
+            {
+                _cache = cache;
+            }
+
+            public IEnumerable<string> ParseCustomDiffMergeTool(string output, string defaultTool)
+                => _cache.ParseCustomDiffMergeTool(output, defaultTool);
+        }
+    }
+}
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1239,6 +1239,12 @@ namespace GitCommands
             set => SetBool("showdiffforallparents", value);
         }
 
+        public static bool ShowAvailableDiffTools
+        {
+            get => GetBool("difftools.showavailable", true);
+            set => SetBool("difftools.showavailable", value);
+        }
+
         public static int DiffVerticalRulerPosition
         {
             get => GetInt("diffverticalrulerposition", 0);
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1593,7 +1593,9 @@ namespace GitUI.CommandsDialogs
             RevisionGrid.ReloadTranslation();
             fileTree.ReloadHotkeys();
             revisionDiff.ReloadHotkeys();
-            revisionDiff.ReloadCustomDifftools();
+            new CustomDiffMergeToolProvider().Clear();
+            revisionDiff.LoadCustomDifftools();
+            RevisionGrid.LoadCustomDifftools();
 
             _dashboard?.RefreshContent();
 
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -71,8 +71,6 @@ namespace GitUI.CommandsDialogs
             this.toolStripSeparator15 = new System.Windows.Forms.ToolStripSeparator();
             this.unstagedSubmoduleStageToolStripSeparator = new System.Windows.Forms.ToolStripSeparator();
             this.openFolderMenuItem = new System.Windows.Forms.ToolStripMenuItem();
-            this.openDiffMenuItem = new System.Windows.Forms.ToolStripMenuItem();
-            this.toolStripSeparator16 = new System.Windows.Forms.ToolStripSeparator();
             this.copyFolderNameMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.gitItemStatusBindingSource = new System.Windows.Forms.BindingSource(this.components);
             this.Cancel = new System.Windows.Forms.Button();
@@ -263,10 +261,9 @@ namespace GitUI.CommandsDialogs
             //
             this.openWithDifftoolToolStripMenuItem.Image = global::GitUI.Properties.Images.Diff;
             this.openWithDifftoolToolStripMenuItem.Name = "openWithDifftoolToolStripMenuItem";
-            this.openWithDifftoolToolStripMenuItem.ShortcutKeys = System.Windows.Forms.Keys.F3;
             this.openWithDifftoolToolStripMenuItem.Size = new System.Drawing.Size(232, 22);
             this.openWithDifftoolToolStripMenuItem.Text = "Open with difftool";
-            this.openWithDifftoolToolStripMenuItem.Click += new System.EventHandler(this.OpenWithDifftoolToolStripMenuItemClick);
+            this.openWithDifftoolToolStripMenuItem.Click += new System.EventHandler(this.openWithDifftoolToolStripMenuItem_Click);
             //
             // toolStripSeparator9
             //
@@ -460,7 +457,6 @@ namespace GitUI.CommandsDialogs
             //
             this.stagedOpenDifftoolToolStripMenuItem9.Image = global::GitUI.Properties.Images.Diff;
             this.stagedOpenDifftoolToolStripMenuItem9.Name = "stagedOpenDifftoolToolStripMenuItem9";
-            this.stagedOpenDifftoolToolStripMenuItem9.ShortcutKeys = System.Windows.Forms.Keys.F3;
             this.stagedOpenDifftoolToolStripMenuItem9.Size = new System.Drawing.Size(232, 22);
             this.stagedOpenDifftoolToolStripMenuItem9.Text = "Open with difftool";
             this.stagedOpenDifftoolToolStripMenuItem9.Click += new System.EventHandler(this.stagedOpenDifftoolToolStripMenuItem9_Click);
@@ -504,8 +500,6 @@ namespace GitUI.CommandsDialogs
             this.toolStripSeparator15,
             this.stageSubmoduleToolStripMenuItem,
             this.unstagedSubmoduleStageToolStripSeparator,
-            this.openDiffMenuItem,
-            this.toolStripSeparator16,
             this.copyFolderNameMenuItem,
             this.openFolderMenuItem,
             this.toolStripSeparator13,
@@ -586,20 +580,6 @@ namespace GitUI.CommandsDialogs
             this.openFolderMenuItem.Size = new System.Drawing.Size(228, 22);
             this.openFolderMenuItem.Text = "Show in folder";
             this.openFolderMenuItem.Click += new System.EventHandler(this.OpenToolStripMenuItemClick);
-            //
-            // openDiffMenuItem
-            //
-            this.openDiffMenuItem.Image = global::GitUI.Properties.Images.Diff;
-            this.openDiffMenuItem.Name = "openDiffMenuItem";
-            this.openDiffMenuItem.ShortcutKeys = System.Windows.Forms.Keys.F3;
-            this.openDiffMenuItem.Size = new System.Drawing.Size(228, 22);
-            this.openDiffMenuItem.Text = "Open with difftool";
-            this.openDiffMenuItem.Click += new System.EventHandler(this.OpenWithDifftoolToolStripMenuItemClick);
-            //
-            // toolStripSeparator16
-            //
-            this.toolStripSeparator16.Name = "toolStripSeparator16";
-            this.toolStripSeparator16.Size = new System.Drawing.Size(225, 6);
             //
             // copyFolderNameMenuItem
             //
@@ -1635,8 +1615,6 @@ namespace GitUI.CommandsDialogs
         private ToolStripSeparator unstagedSubmoduleStageToolStripSeparator;
         private ToolStripSeparator toolStripSeparator15;
         private ToolStripMenuItem openFolderMenuItem;
-        private ToolStripMenuItem openDiffMenuItem;
-        private ToolStripSeparator toolStripSeparator16;
         private ToolStripMenuItem copyFolderNameMenuItem;
         private ToolStripMenuItem resetSubmoduleChanges;
         private ToolStripMenuItem commitSubmoduleChanges;
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -244,6 +244,8 @@ namespace GitUI.CommandsDialogs
             openToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenFile);
             openWithToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenFileWith);
             editFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.EditFile);
+            openWithDifftoolToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenWithDifftool);
+            stagedOpenDifftoolToolStripMenuItem9.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.OpenWithDifftool);
 
             stageSubmoduleToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.StageSelectedFile);
 
@@ -447,6 +449,7 @@ namespace GitUI.CommandsDialogs
         {
             showUntrackedFilesToolStripMenuItem.Checked = Module.EffectiveConfigFile.GetValue("status.showUntrackedFiles") != "no";
             MinimizeBox = Owner is null;
+            LoadCustomDifftools();
             base.OnLoad(e);
         }
 
@@ -576,6 +579,17 @@ namespace GitUI.CommandsDialogs
             }
 
             return base.ProcessCmdKey(ref msg, keyData);
+        }
+
+        public void LoadCustomDifftools()
+        {
+            List<CustomDiffMergeTool> menus = new()
+            {
+                new(openWithDifftoolToolStripMenuItem, openWithDifftoolToolStripMenuItem_Click),
+                new(stagedOpenDifftoolToolStripMenuItem9, stagedOpenDifftoolToolStripMenuItem9_Click),
+            };
+
+            new CustomDiffMergeToolProvider().LoadCustomDiffMergeTools(Module, menus, components, isDiff: true);
         }
 
         private void FileViewer_TopScrollReached(object sender, EventArgs e)
@@ -2498,23 +2512,37 @@ namespace GitUI.CommandsDialogs
             ClipboardUtil.TrySetText(fileNames.ToString());
         }
 
-        private void OpenFilesWithDiffTool(IEnumerable<FileStatusItem> items)
+        private void OpenFilesWithDiffTool(IEnumerable<FileStatusItem> items, object sender)
+        {
+            var item = sender as ToolStripMenuItem;
+            if (item?.DropDownItems != null)
+            {
+                // "main menu" clicked, cancel dropdown manually, invoke default mergetool
+                item.HideDropDown();
+                item.Owner.Hide();
+            }
+
+            string toolName = item?.Tag as string;
+            OpenFilesWithDiffTool(items, toolName);
+        }
+
+        private void OpenFilesWithDiffTool(IEnumerable<FileStatusItem> items, string toolName = null)
         {
             foreach (var item in items)
             {
                 GitRevision[] revs = new[] { item.SecondRevision, item.FirstRevision };
-                UICommands.OpenWithDifftool(this, revs, item.Item.Name, item.Item.OldName, RevisionDiffKind.DiffAB, item.Item.IsTracked);
+                UICommands.OpenWithDifftool(this, revs, item.Item.Name, item.Item.OldName, RevisionDiffKind.DiffAB, item.Item.IsTracked, customTool: toolName);
             }
         }
 
-        private void OpenWithDifftoolToolStripMenuItemClick(object sender, EventArgs e)
+        private void openWithDifftoolToolStripMenuItem_Click(object sender, EventArgs e)
         {
-            OpenFilesWithDiffTool(Unstaged.SelectedItems);
+            OpenFilesWithDiffTool(Unstaged.SelectedItems, sender);
         }
 
         private void OpenWithDiffTool()
         {
-            (_currentItemStaged ? stagedOpenDifftoolToolStripMenuItem9 : openWithDifftoolToolStripMenuItem).PerformClick();
+            OpenFilesWithDiffTool(_currentItemStaged ? Staged.SelectedItems : Unstaged.SelectedItems);
         }
 
         private void ResetPartOfFileToolStripMenuItemClick(object sender, EventArgs e)
@@ -3083,7 +3111,7 @@ namespace GitUI.CommandsDialogs
 
         private void stagedOpenDifftoolToolStripMenuItem9_Click(object sender, EventArgs e)
         {
-            OpenFilesWithDiffTool(Staged.SelectedItems);
+            OpenFilesWithDiffTool(Staged.SelectedItems, sender);
         }
 
         private void openFolderToolStripMenuItem10_Click(object sender, EventArgs e)
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -138,9 +138,9 @@ namespace GitUI.CommandsDialogs
                 case Command.DeleteSelectedFiles: diffDeleteFileToolStripMenuItem.PerformClick(); break;
                 case Command.ShowHistory: fileHistoryDiffToolstripMenuItem.PerformClick(); break;
                 case Command.Blame: blameToolStripMenuItem.PerformClick(); break;
-                case Command.OpenWithDifftool: firstToSelectedToolStripMenuItem.PerformClick(); break;
-                case Command.OpenWithDifftoolFirstToLocal: firstToLocalToolStripMenuItem.PerformClick(); break;
-                case Command.OpenWithDifftoolSelectedToLocal: selectedToLocalToolStripMenuItem.PerformClick(); break;
+                case Command.OpenWithDifftool: OpenFilesWithDiffTool(RevisionDiffKind.DiffAB); break;
+                case Command.OpenWithDifftoolFirstToLocal: OpenFilesWithDiffTool(RevisionDiffKind.DiffALocal); break;
+                case Command.OpenWithDifftoolSelectedToLocal: OpenFilesWithDiffTool(RevisionDiffKind.DiffBLocal); break;
                 case Command.EditFile: diffEditWorkingDirectoryFileToolStripMenuItem.PerformClick(); break;
                 case Command.OpenAsTempFile: diffOpenRevisionFileToolStripMenuItem.PerformClick(); break;
                 case Command.OpenAsTempFileWith: diffOpenRevisionFileWithToolStripMenuItem.PerformClick(); break;
@@ -173,22 +173,18 @@ namespace GitUI.CommandsDialogs
             DiffText.ReloadHotkeys();
         }
 
-        public void ReloadCustomDifftools()
+        public void LoadCustomDifftools()
         {
-            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            List<CustomDiffMergeTool> menus = new()
             {
-                var tools = await Module.GetCustomDiffMergeTools(isDiff: true);
-                openWithCustomDifftoolToolStripMenuItem.DropDown = null;
-                ContextMenuStrip customDiffToolDropDown = new();
-                foreach (var tool in tools)
-                {
-                    var toolStripItem = new ToolStripMenuItem(tool) { Tag = tool };
-                    toolStripItem.Click += openWithDifftoolToolStripMenuItem_Click;
-                    customDiffToolDropDown.Items.Add(toolStripItem);
-                }
+                new(firstToSelectedToolStripMenuItem, firstToSelectedToolStripMenuItem_Click),
+                new(selectedToLocalToolStripMenuItem, selectedToLocalToolStripMenuItem_Click),
+                new(firstToLocalToolStripMenuItem, firstToLocalToolStripMenuItem_Click),
+                new(diffWithRememberedDifftoolToolStripMenuItem, diffWithRememberedDiffToolToolStripMenuItem_Click),
+                new(diffTwoSelectedDifftoolToolStripMenuItem, diffTwoSelectedDiffToolToolStripMenuItem_Click)
+            };
 
-                openWithCustomDifftoolToolStripMenuItem.DropDown = customDiffToolDropDown;
-            }).FileAndForget();
+            new CustomDiffMergeToolProvider().LoadCustomDiffMergeTools(Module, menus, components, isDiff: true);
         }
 
         private string GetShortcutKeyDisplayString(Command cmd)
@@ -239,7 +235,7 @@ namespace GitUI.CommandsDialogs
             DiffText.SetFileLoader(GetNextPatchFile);
             DiffText.Font = AppSettings.FixedWidthFont;
             ReloadHotkeys();
-            ReloadCustomDifftools();
+            LoadCustomDifftools();
 
             base.OnRuntimeLoad();
         }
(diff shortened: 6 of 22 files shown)


gerhardol:
Got some hanged tests related to left panel that should be unrelated to the changes.
The hangs pointed to another printout for View_refect_reset_branch_filter() that existed before but are not listed as an error.
Probably a coincidence that the test did not hang when fixing that.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -590,6 +593,40 @@ private void SelectedDiff_ContextMenuOpening(object sender, System.ComponentMode
...
+        {
+            openWithDifftoolToolStripMenuItem.DropDown = null;
+            stagedOpenDifftoolToolStripMenuItem9.DropDown = null;
+
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                var difftool = Module.GetEffectiveSetting(GitCommands.Config.SettingKeyString.DiffToolKey);
+                var tools = await Module.GetCustomDiffMergeTools(isDiff: true);
+                _unstagedOpenDifftoolToolStripMenuItem = new ContextMenuStrip();
+                _stagedOpenDifftoolToolStripMenuItem = new ContextMenuStrip();
```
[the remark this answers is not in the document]
gerhardol: The menuitem could be statically created and cleared if the menu is reloaded (done in RevDiff()).
The menuitem is assigned to the DropDown after creation (for other items) or with right-click (this example, not changing the current looks) so the dropdown arrow is not immediately available.
If changed, it should be done in all five forms using this functionality.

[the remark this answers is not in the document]
gerhardol: Moved variables to the Designer file, added them to the component (so they are disposed of).

gerhardol:
Added caching of the `git-difftool --tool-help`to speedup showing the custom difftool in FormCommit (as RevDiff initiates the cache) and ResolveConflicts (second time).
Unfortunately not in FileHistory, as that runs as a separate instance.

mstv:
This one is huge. Even understanding the usage of the changes takes time. I have it in my private build.

You mentioned Commit (FormCommit?) in the PR title. Something for another PR?

There seems to be no intuitive replacement for "First -> Second...".

gerhardol:
> This one is huge. Even understanding the usage of the changes takes time. I have it in my private build.

The basis has already been implemented in RevDiff, this is an adjustment to how it works there and adding to more diffs.
The changes are quite repetive - adding the same changes to all GUI forms.
Something like LoadCustomDifftools() could be shared, but the usages are somehow different, so it would be more complicated in my opinion.
Splitting it would not be much different, the same changes in many forms.

I could separate the last minor commit, for caching (if this PR is delayed, I will probably do that, to get faster custom mergetools).
Similar for the difftools that are new since 3.4 should not be contriversal. (RevGrid is a special case too.)

> You mentioned Commit (FormCommit?) in the PR title. Something for another PR?

It is there. To not change the defaults, you have to right-click to see the dropdown. There is a tooltip.

> There seems to be no intuitive replacement for "First -> Second...".

What is changed in this PR is that the additional First->Second is removed, just the right-click menu.

mstv:
>> This one is huge. Even understanding the usage of the changes takes time. I have it in my private build.

> I could separate the last minor commit, for caching

This was just an excuse why I shifted the review - not requesting a split.

I played with it 🚀 and think I got it. I can follow the good reasoning 👍, though have some thoughts and nitpicks.

> I would have preferred to enable this functionality by default, but there will be users that will be annoyed by the extra dropdowns.
> There is a tooltip.

Good idea, will help several users. But I don't use to hover menu items.
The submenu cannot be activated using the keyboard.
This makes it hard to discover / use the feature.
The additonal submenu does not hurt / disturb if users recognize that the parent menu item can be clicked.
Since this is a common UX of GE menus, we could give it a try and enable it by default - perhaps with a regarding tooltip for the parent and for the default item.

If we keep the de/activation by right click,
* the right click should not close the menu,
* the submenu for "Diff with 'remembered'" should be toggleable, too,
* the toggled state should be remembered and apply to all regarding menu items - at least for the runtime of the GE instance

The default item should be marked with bold font instead of a checkbox. The checkbox makes me think I could/would change the default.

Nit: RevDiff has "First -> Second" (good), FileHistory has "< - >" with spaces, should be "<->".

Caching will be very helpful. (The addional delay in FormHistory shouldn't be an issue because opening a specific diff will not be the first action I think.)

gerhardol:
> Since this is a common UX of GE menus, we could give it a try and enable it by default - perhaps with a regarding tooltip for the parent and for the default item.

Should I change it this way?
If there is only one difftool, the dropdown do not need to be added so it will not annoy users with only one difftool.
We could add a configuration too. (Detailed-DiffViewer is the best fit, even if the configuration is used elsewhere too.)
(Or should right click change the global configuration?)

> If we keep the de/activation by right click,
>     * the right click should not close the menu,

I was not able to get this functionality. 
Is select (default) and close OK?

>     * the submenu for "Diff with 'remembered'" should be toggleable, too,

One reason to add it for "remember" is to improve how to discover this.
If it is on by default, it is still giving the hint.

Added by default in RevGrid and for mergetool, those should be kept.

>     * the toggled state should be remembered and apply to all regarding menu items - at least for the runtime of the GE instance

For this form if so.

> The default item should be marked with bold font instead of a checkbox. The checkbox makes me think I could/would change the default.

Will change (did not realize it could be done this way - I have done it before myself though...)

> Nit: RevDiff has "First -> Second" (good), FileHistory has "< - >" with spaces, should be "<->".

This was not changed, but there are many inconsistent menus. There are quite some more diff options in RevDiff...
Later - maybe in this PR though

gerhardol:
Rebased and squashed, there were merge conflicts
(also reverted a minor change in openWithCustomDifftoolToolStripMenuItem_MouseDown() that should be discussed first)

for the previous branch, see following (that will be deleted in some time...)
https://github.com/gerhardol/gitextensions/pull/new/feature/revdiff-custom-difftool-2

mstv:
@gitextensions/git-extensions-team, should the submenus with the difftool selection always be shown?
At the moment, they are only shown after right clicking the difftool menu item.
When the submenu is added,the parent menu item can still be clicked and starts the difftool chosen in the settings.

> If there is only one difftool, the dropdown do not need to be added so it will not annoy users with only one difftool.

👍 

---

> > If we keep the de/activation by right click, the right click should not close the menu,

> I was not able to get this functionality.

I mean the second right click which deactivates the submenu.

pmiossec:
Right click to avoid contextual menu is perhaps a too hidden feature that will only be known and used by us.

But I can't think of a more discoverable way to do.

Perhaps a just little better, also displaying the subitem when 'shift' is pressed when right clicking like it is done in the file explorer.

But that doesn't solve the problem of discoverability.
Adding a tool tip that inform about the feature?

gerhardol:
> 
> 
> Right click to avoid contextual menu is perhaps a too hidden feature that will only be known and used by us.
> 
> But I can't think of a more discoverable way to do.
> 
> Perhaps a just little better, also displaying the subitem when 'shift' is pressed when right clicking like it is done in the file explorer.
> 
> But that doesn't solve the problem of discoverability.
> Adding a tool tip that inform about the feature?

There is a tooltip already.
I can add shift, but would prefer to have that toggle the appearance as rightclick does - or that rightclick is required to expand? not my preference)
Maybe add a new setting after all?

If rightclick to expand, should rightclick when there are no alternatives select directly (as now) or do nothing?

> I mean the second right click which deactivates the submenu.

I understood, but I did not get that working

gerhardol:
Rebased and changed implementation to enable the custom difftools with a setting in RevDiff, Commit and History. RevGrid is special (compare directories) so always enabled there, similar for mergetools (new functionality).

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1563,7 +1563,8 @@ private void OnShowSettingsClick(object sender, EventArgs e)
             RevisionGrid.ReloadTranslation();
             fileTree.ReloadHotkeys();
             revisionDiff.ReloadHotkeys();
-            revisionDiff.ReloadCustomDifftools();
+            revisionDiff.LoadCustomDifftools();
+            GitModule.GitCommandCache.Clear();
```
mstv, line 1567: Shall the cache be cleared after loading? 🤔

mstv (commented):
LGTM!

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: Not really, the entries in the cache should be unique from repo to repo (assuming shas are unique). Tools could be per repo too, that will require a reload, it is a reasonable limitation in my opinion.

gerhardol:
How to handle shortcut (hotkey) hints?
If dropdown is always available the shortcut is not visible. The right click to show dropdown was a way to get around that.
But a configuration is probably better, let us solve this in another way:
* Just ignore that the shortcut is not visible if dropdown is available. If so, the default for the dropdown should probably be off.
* Use a tooltip. Simple to implement, but the tooltip is not easy to see here
* Add a custom renderer and show the shortcut manually. Some hints: https://stackoverflow.com/a/33625362
* Find some secret setting in WinForms to achieve this anyway

mstv:
The hotkey can be displayed on the default entry of the submenu.
I prefer the setting to be on by default. If one has configured multiple diff tools, they will be please I guess.
The setting can have a tooltip (we have a special control for this) which tells that the parent menu item is clickable.

gerhardol:
Rebased (squashed one commit) on master (there were conflicts)

> The hotkey can be displayed on the default entry of the submenu.

This requires a custom renderer

> I prefer the setting to be on by default. If one has configured multiple difftools, they will be pleased I guess.

Agree, off was an alternative for showing the shortcut

> The setting can have a tooltip (we have a special control for this) which tells that the parent menu item is clickable.

Will add

GitUI/CommandsDialogs/FormBrowse.cs
mstv: This is executed after closing FormSettings. Why not clear the cache before loading the difftools?
Why have you added `.Clear()` at all? Is it fixing another issue? I agree to "should be unique from repo to repo (assuming shas are unique)". How is it related to difftools?

---

Can the initial loading of the difftools be postponed until the RevisionGrid has been populated?
My AV virus becomes very active and slows down the more important git commands.

gerhardol: Without a cache, it can take over 10s for me (with Symantec virus) before the tools are available.
This is OK for FormBrowse, but not for FormCommit or ResolveConflicts.

If a user add or remove a tool it is not updating the list unless the cache is cleared.
We can of course require that the user close/open GE to refresh the list, it is not unreasonable.
But with a GE centric perspective, difftools are added and removed in Settings, I believe the user expects that a newly added tool is available in th lst if added in settings.
This can of course be described in the settings tooltip (that is to be added).

mstv: I am not arguing neither against the cache nor against clearing it after settings have changed.

But I have not got yet why not rather use the following sequence: 
```
            GitModule.GitCommandCache.Clear();
            revisionDiff.LoadCustomDifftools();
```

gerhardol: Of course, this order will be changed

mstv:
`firstToSelectedItem.ShortcutKeyDisplayString = "F3";`

grafik

gerhardol:
Tooltip is not translated, otherwise I consider this finished
The shortcut displayed were hardcoded in FormCommit (which means that F3 was always available regardless if hotkey was assigned but you could not see the hotkey).

mstv (approved):
Thank you!
The settings tooltips should be detected by the Translation App.

pmiossec:
>I would have preferred to enable this functionality by default, but there will be users that will be annoyed by the extra dropdowns.


I highly prefer a functionality enabled by default that I can disable easily than a functionality that I will never know it exists...

That's why I think this feature should be enabled by default.

The only thing is that I would add at the bottom of the list diff tool, after a separator, a toolstripitem "Disable this drop-down" that show a message like (in a lot better English) :

 "you will only keep the comparison with default diff tool.

You could still enable it again by right clicking on the menu entry

Do you want to disable this functionality? 

Yes/no"

gerhardol:
Rebased on master so the ToolTip is translated. Plan to squash tomorrow as mstv has approve and merge a couple of days later unless there are comments (assuming we resolve the one below).

> I highly prefer a functionality enabled by default that I can disable easily than a functionality that I will never know it exists...
> That's why I think this feature should be enabled by default.

Yes, it is on now.  
But functionality that are useful to some and annoying for other should be disabled by default.

> The only thing is that I would add at the bottom of the list diff tool, after a separator, a toolstripitem "Disable this drop-down" that show a message like (in a lot better English) :

If this is added, I would like to disable the functionality globally by changing the setting (and refreshing).
No popup or tooltip or rightclick to reenable, to not make it too complicated.
I expect this functionality to be on or off permanently for each user.

RussKie:
> Presentation of mergetools updated
> 
> image

Can we sort the difftools?

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -578,6 +581,54 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
...
+            stagedOpenDifftoolToolStripMenuItem9.DropDown = null;
+            if (!AppSettings.ShowAllCustomDiffTools)
+            {
+                return;
+            }
+
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                var difftool = Module.GetEffectiveSetting(SettingKeyString.DiffToolKey);
+                var tools = await Module.GetCustomDiffMergeTools(isDiff: true);
```
RussKie, line 596: Please use explicit type instead of `var`, helps understanding the code

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -578,6 +581,54 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
...
+                        openWithDifftoolToolStripMenuItem.DropDown.Items.Insert(0, unstagedItem);
+                        stagedOpenDifftoolToolStripMenuItem9.DropDown.Items.Insert(0, stagedItem);
+                    }
+                    else
+                    {
+                        openWithDifftoolToolStripMenuItem.DropDown.Items.Add(unstagedItem);
+                        stagedOpenDifftoolToolStripMenuItem9.DropDown.Items.Add(stagedItem);
+                    }
+                }
+            }).FileAndForget();
```
RussKie, line 629: We have a somewhat similar code repeated in several places... Can we generalise it somehow?

gerhardol:
> Can we sort the difftools?

The tools are sorted, the default tool is placed first (bold in the current version).

> We have a somewhat similar code repeated in several places... Can we generalise it somehow?

As in the initial proposal it was not really possible as the handling differed too much.
With a global setting to control this the handling is more uniform and it is possible.
(I do not know where to place the code though...)
So if a global setting is preferred this can be changed. 
Note though that there are as much code in all of the event handlers so there still some repeated code.

Should there be an alternative to disable the dropdown in the dropdown and should it be simple off or allow to reenable from the menu?
https://github.com/gitextensions/gitextensions/pull/8700#issuecomment-761245476

Similar with updating the screenshots, awaiting some consensus (also known as a RussKie decision).

Tests fail occasionally for this PR, seem to be unrelated to the PR.
1) Failed : GitExtensions.UITests.CommandsDialogs.FormBrowse_LeftPanel_RemotesTests.RepoObjectTree_should_load_active_remotes
  Test exceeded Timeout value of 90000ms

Seem to be a failure to upload to CodeCov:

2021-01-15 23:26:06 FTL Failed to upload the report.
     at Codecov.Upload.Uploads.Uploader()
     at Codecov.Program.UploadFacade.Uploader()
     at Codecov.Program.Run.Uploader()
     at Codecov.Program.Run.Runner(IEnumerable`1 args)

GitCommands/Settings/AppSettings.cs
```diff
@@ -1158,6 +1158,12 @@ public static bool ShowDiffForAllParents
             set => SetBool("showdiffforallparents", value);
         }
 
+        public static bool ShowAllCustomDiffTools
+        {
+            get => GetBool("showallcustomdifftools", true);
+            set => SetBool("showallcustomdifftools", value);
+        }
```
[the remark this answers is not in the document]
gerhardol: "show all available custom tools"
I can change, but I prefer to not have a too long name...

GitUI/CommandsDialogs/FormCommit.cs
gerhardol: ok

RussKie:
> > Can we sort the difftools?
> 
> The tools are sorted, the default tool is placed first (bold in the current version).

It took me a bit to realise that. My preference would be to keep it simple and sort it A-Z.

> image
> 
> I can add shift, but would prefer to have that toggle the appearance as rightclick does - or that rightclick is required to expand? not my preference)
> Maybe add a new setting after all?
> 
> If rightclick to expand, should rightclick when there are no alternatives select directly (as now) or do nothing?

Right click on a context menus is really foreign concept to Windows (and any other OS) users. So this doesn't feel like a good idea.


> 
> image

What if the dpropdown menu item said "Open in <currently selected tool>", and the dropdown list is A-Z sorted with the currently selected tool checked? If a user checks another tool the parent menu item is updated with the new text.
I would then apply the same concept to all dropdowns.

RussKie:
> 
> image

What if the dpropdown menu item said "Open in \<currently selected tool\>", and the dropdown list is A-Z sorted with the currently selected tool checked? If a user checks another tool the parent menu item is updated with the new text.
I would then apply the same concept to all dropdowns.


> customdiff

E.g. on this animation for single and multi file selection the top-level menu would say: "Open in torteisediff". The dropdown has the new menu as the last entry (like you have it) "Set default diff (or merge) tool", which shows the list of available diff or merge tools (depending on a situation).

GitCommands/Settings/AppSettings.cs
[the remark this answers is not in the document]
RussKie: In this case how about this?
```suggestion
        public static bool ShowAvailableDiffTools
        {
            get => GetBool("showcustomdifftools", true);
            set => SetBool("showcustomdifftools", value);
        }
```

gerhardol:
> What if the dpropdown menu item said "Open in <currently selected tool>", and the dropdown list is A-Z sorted with the currently selected tool checked? If a user checks another tool the parent menu item is updated with the new text.
> I would then apply the same concept to all dropdowns.

In ResolveConflicts naming the tool would work, but there is a difference in workflow from "GE resolve" vs "mergetool resolve", so that need to be explained too.
For instance `Mergetool in kdiff3`
For RevDiff with several diff alternatives, this could be "First -> second in kdiff3"

I would prefer to not implement a dynamic selection. A fixed default is easier to explain.
A dynamic tool is OK when selecting in the menu but not when using a shortcut.
Also, would the Git default be changed, the default in this menu, for all in the form or all forms until restart?
If multiple dropdowns and multiple forms, how to synchronize?

Also without "dynamic" change, I do not think that it is easier to understand the functionality with adding the tool to the text.

> E.g. on this animation for single and multi file selection the top-level menu would say: "Open in torteisediff". The dropdown has the new menu as the last entry (like you have it) "Set default diff (or merge) tool", which shows the list of available diff or merge tools (depending on a situation).

That is partly a different suggestion. I choose to interpret it as a menu to change the default difftool.
That is an alternative solution to get the functionality. This is not my preference as I find it more predictable to have one tool, it is in specific situations I want to use alternative tools.

GitCommands/Settings/AppSettings.cs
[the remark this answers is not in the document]
gerhardol: ok

gerhardol:
Pushed most review comments

>> The tools are sorted, the default tool is placed first (bold in the current version).

> It took me a bit to realise that. My preference would be to keep it simple and sort it A-Z.

The default difftool is on top of the menu and in bold as the standard. I would like to keep it as that.

Common code is a little fishy sharing components etc but I do not know a better way to solve this.
The last commit disables the dropdown. There are some limitations, the list is not refreshed until a restart if invoked from FormDiff or FileHistory. I believe that is a reasonable limitation.

image

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -578,6 +581,17 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
...
 
+        public void LoadCustomDifftools()
+        {
+            List<CustomDiffMergeTool> menus = new ()
+            {
+                new (openWithDifftoolToolStripMenuItem, UnstagedOpenWithDifftoolToolStripMenuItem_Click),
+                new (stagedOpenDifftoolToolStripMenuItem9, StagedOpenDifftoolToolStripMenuItem_Click),
+            };
+
+            CustomDiffMergeTool.LoadCustomDiffMergeTools(Module, menus, components, true, true);
```
mstv, line 592: For GH and those without Re#:
```suggestion
            CustomDiffMergeTool.LoadCustomDiffMergeTools(Module, menus, components, isDiff: true, useSettings: true);
```

mstv (approved):
just a nit from my side

GitUI/CommandsDialogs/FormCommit.cs
gerhardol: Thanks!

gerhardol:
squashed and rebased and added a few touchups (including one existing where RevGrid.SetShortcutKeys() ran many times, removed one of them)
I noticed an existing issue that if F3 is set as the shortcut for diffing commits in RevGrid then the ShortCut in History will not work (will not do anything about it right now).

gerhardol:
Build fails for same reason as above, seem to be CodeCov problem
This PR is not that big, maybe it still pushes over a limit?

```
  2021-01-21 22:56:19 WRN Total file size of reports is above 20MB, this may prevent report being shown on codecov.io
  2021-01-21 22:56:19 WRN Reduce the total upload size if this occurs
  2021-01-21 22:56:22 WRN Unable to upload coverage report to Codecov. Server returned: (400) Bad Request
  2021-01-21 22:56:22 WRN Unknown reason. Possible reason being invalid parameters.
  2021-01-21 22:56:22 WRN Failed to upload the report with CodecovUploader.
  2021-01-21 22:56:22 INF Uploading to Codecov
```

mstv:
Ready for merge IMO.

And in a follow-up PR, please delay the initial loading of the difftools:
GE startup times in seconds | commits visible | CPU idle
-|-|-
with difftools off | 7.5 | 27.5
with difftools on | 28 | 38

Release build started with all plugins from `artifacts\bin\GitExtensions\Release\net461`, lengthy `%PATH%` and non-deactivatable AV

diff/merge sections in my `.gitconfig`:
```
[difftool]
	prompt = false
[mergetool]
	prompt = false
	keepbackup = false
[merge]
	guitool = BeyondCompare4
[diff]
	guitool = BeyondCompare4
[difftool "BeyondCompare4"]
	path = C:/Program Files/Beyond Compare 4/bcomp.exe
	cmd = \"C:/Program Files/Beyond Compare 4/bcomp.exe\" \"$REMOTE\" \"$LOCAL\"
[mergetool "BeyondCompare4"]
	path = C:/Program Files/Beyond Compare 4/bcomp.exe
	cmd = \"C:/Program Files/Beyond Compare 4/bcomp.exe\" \"$LOCAL\" \"$REMOTE\" \"$BASE\" \"$MERGED\"
[difftool "Meld"]
	path = C:/Program Files (x86)/Meld/meld.exe
	cmd = \"C:/Program Files (x86)/Meld/meld.exe\" \"$REMOTE\" \"$LOCAL\"
[difftool "P4Merge"]
	path = C:/Program Files/Perforce/p4merge.exe
	cmd = \"C:/Program Files/Perforce/p4merge.exe\" \"$REMOTE\" \"$LOCAL\"
```

gerhardol:
> Ready for merge IMO.

The issue I have seen is with the test that fails
So the delay may be required

> And in a follow-up PR, please delay the initial loading of the difftools:

Feel I have to do that already now to pass the tests....

> Release build started with all plugins from `artifacts\bin\GitExtensions\Release\net461`, lengthy `%PATH%` and non-deactivatable AV

You really should run a build run with publish:
I build with `./build.cmd -pub -ci`  and run the build in `artifacts/publish/Debug` instead.
The startup decreases from 12s to 4s for me. (no custom difftools)

--

How long time for `git difftool --tool-help`
7s on my home PC (13 tools, most time is probably for built in tools), up to 20s on the work PC (Symantec as virus).

--

Reworked how the tools are setup

fixup! Use the setting for showing the available difftools also in RevGrid
Even if the command may be useful there, a user may want to avoid the extra CPU time

squash! Keep a common cache for the parsed tool, to only execute the command once
Cache git-difftool --tool-help
Speedup especially FormCommit and ResolveConflicts
The cache is reset when closing settings.

fixup! delay before parsing difftools, to let other operations start
RevDiff and RevGrid runs when the Form start (git-log runs),
so longer delay there
Note that FormCommit should have a cache available so the
delay there should not make a difference.

fixup! No longer cache the Git command
The shared mutex made the handling hidden, so all handling moved to
CustomDiffMergeTool

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Drawing;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using GitCommands;
+
+namespace GitUI
+{
+    public class CustomDiffMergeTool
```
RussKie, line 12: I have multiple issues with the API naming and coherency in this type.

E.g. `CustomDiffMergeTool` implies a singular item. `Clear()` generally implies a collection.
I also struggle with `LoadCustomDiffMergeTools(...)` method in this class, it feels misplaced.

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        /// <summary>
+        /// Load the availble DiffMerge tools and apply to the menus
+        /// </summary>
+        /// <param name="module">The Git module</param>
+        /// <param name="menus">The menus to update</param>
+        /// <param name="components">The calling Form components, to dispose correctly</param>
+        /// <param name="isDiff">True if diff, false if merge</param>
+        /// <param name="delay">The delay before starting the operation</param>
+        public static void LoadCustomDiffMergeTools(GitModule module, IList<CustomDiffMergeTool> menus, IContainer components, bool isDiff, int delay = 1000)
```
RussKie, line 67: Why do you feel we need to re-run this method instead of loading all available diff/merge tools once at the app startup? 
It isn't a common scenario where a user constantly installs/uninstalls diff tools. Loading the tools only once we will greatly simplify the code and associated plumbing. We also will need to build and maintain a single instance of `ContextMenuStrip`, and we won't need to maintain its lifetime.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2498,23 +2512,37 @@ private void FilenameToClipboardToolStripMenuItemClick(object sender, EventArgs
             ClipboardUtil.TrySetText(fileNames.ToString());
         }
 
-        private void OpenFilesWithDiffTool(IEnumerable<FileStatusItem> items)
+        private void OpenWithCustomDifftoolToolStripMenuItem_Click(IEnumerable<FileStatusItem> items, object sender)
```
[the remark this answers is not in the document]
gerhardol: Renaming to OpenFilesWithDiffTool
Invoked when top or dropdown is clicked
If top is invoked, Tag is not set and no difftool is set (null) so Git uses the default difftool

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2498,23 +2512,37 @@ private void FilenameToClipboardToolStripMenuItemClick(object sender, EventArgs
...
+                // "main menu" clicked, cancel dropdown manually, invoke default mergetool
+                item.HideDropDown();
+                item.Owner.Hide();
+            }
+
+            string toolName = item?.Tag as string;
+            OpenFilesWithDiffTool(items, toolName);
+        }
+
+        private void OpenFilesWithDiffTool(IEnumerable<FileStatusItem> items, string toolName = null)
```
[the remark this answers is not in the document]
gerhardol: No difftool specified, no request to Git -> default difftool
A specific constant would be confusing (may clash with what a user uses too). Empty string could be used, but the Tag must be init in many situations.

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
gerhardol: `CustomDiffMergeTool` is a data class
`LoadCustomDiffMergeTools` is a shared method for the GUI. You asked for a common implementation so I put it as a static method there. The cache was added in the latest change (I did not want to add such functionality here and adding to GitModule or the cache would be simpler but we do not want to add functionality there.

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
gerhardol: `LoadCustomDiffMergeTools` is called at startup of the form. The run is slightly delayed to not compete with CPU for git-log.
Each form need to update each of the menu items that handle difftools (so RevDiff has five menu items).

There is a minimal extra plumbing to reload. This is not a major functionality, I just felt that a user expected the list was refreshed after closing Settings, when `Clear` is called and the difftools are reloaded. (This ignores that it is possible to add diff/mergetools to repo settings).

GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -460,10 +457,9 @@ private void InitializeComponent()
             //
             this.stagedOpenDifftoolToolStripMenuItem9.Image = global::GitUI.Properties.Images.Diff;
             this.stagedOpenDifftoolToolStripMenuItem9.Name = "stagedOpenDifftoolToolStripMenuItem9";
-            this.stagedOpenDifftoolToolStripMenuItem9.ShortcutKeys = System.Windows.Forms.Keys.F3;
             this.stagedOpenDifftoolToolStripMenuItem9.Size = new System.Drawing.Size(232, 22);
             this.stagedOpenDifftoolToolStripMenuItem9.Text = "Open with difftool";
-            this.stagedOpenDifftoolToolStripMenuItem9.Click += new System.EventHandler(this.stagedOpenDifftoolToolStripMenuItem9_Click);
+            this.stagedOpenDifftoolToolStripMenuItem9.Click += new System.EventHandler(this.StagedOpenDifftoolToolStripMenuItem_Click);
```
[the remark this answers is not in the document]
gerhardol: Renamed after a review comment.. 
stagedOpenDifftoolToolStripMenuItem9_Click

mstv:
GE startup times in seconds | commits visible | Diff & CommitInfo | CPU idle
-|-|-|-
with difftools off | 9.5 | 12 | 21
with difftools on | 10 | 13 | 38

This matches the 17 seconds of `time git difftool --tool-help`.

gerhardol:
> with difftools off 	9.5 	12 	21
> with difftools on 	10 	13 	38

The annoying for the user would be Diff&CommitInfo delay
I added a 5s delay before retrieving the tools, should that be increased to 15s or so?

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+                finally
+                {
+                    _mergeToolMutex.Release();
+                }
+            }).FileAndForget();
+        }
+
+        /// <summary>
+        /// Load the availble DiffMerge tools and apply to the menus
```
mstv, line 60:
```suggestion
        /// Load the available DiffMerge tools and apply to the menus
```

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -173,22 +173,18 @@ public void ReloadHotkeys()
...
+                new(firstToSelectedToolStripMenuItem, firstToSelectedToolStripMenuItem_Click),
+                new(selectedToLocalToolStripMenuItem, selectedToLocalToolStripMenuItem_Click),
+                new(firstToLocalToolStripMenuItem, firstToLocalToolStripMenuItem_Click),
+                new(diffWithRememberedDifftoolToolStripMenuItem, diffWithRememberedDiffToolToolStripMenuItem_Click),
+                new(diffTwoSelectedDifftoolToolStripMenuItem, diffTwoSelectedDiffToolToolStripMenuItem_Click)
+            };
 
-                openWithCustomDifftoolToolStripMenuItem.DropDown = customDiffToolDropDown;
-            }).FileAndForget();
+            CustomDiffMergeTool.LoadCustomDiffMergeTools(Module, menus, components, isDiff: true, delay: 5000);
```
mstv, line 187: This magic number of 5 seconds is used twice.
I'd add a constant to `CustomDiffMergeTool`.

Yes, 15 s might be a good compromise for all users.

mstv (commented):
I have reviewed and run `gerhardol/tmp/revdiff-custom-difftool-1` which has not been pushed to this PR yet.
Feels better with the delay, thank you.

A few nits:
- The "9" in `stagedOpenDifftoolToolStripMenuItem9` lets me stumble every time I read it. Remove it?
- Rename `UnstagedOpenWithDifftoolToolStripMenuItem_Click` to `openWithDifftoolToolStripMenuItem_Click`?

gerhardol:
> I have reviewed and run `gerhardol/tmp/revdiff-custom-difftool-1` which has not been pushed to this PR yet.

I realized that about the time you posted the review...

>     * The "9" in `stagedOpenDifftoolToolStripMenuItem9` lets me stumble every time I read it. Remove it?

Separate cleanup of the menu items should be done, they all have suffixes.

>     * Rename `UnstagedOpenWithDifftoolToolStripMenuItem_Click` to `openWithDifftoolToolStripMenuItem_Click`?

RussKie requested that I renamed the staged eventhandler, renamed the unstaged eventhandler back too.
I had renamed them because I was confused too..

--
The update separates the tool cache to a separate class too

gerhardol:
@RussKie Any further comments?

RussKie:
Thank you for the nudge. I will look over it again over the weekend.

GitUI/CommandsDialogs/CustomDiffMergeTool.cs
RussKie: I really struggle with this API. It mixes multiple concerns - it is a data class, and it also responsible for data load, data parsing, UI binding, and cache busting.
It also appears completely divorced from diff/merge tool API that live in GitCommands.DiffMergeTools namespace.

I spent some time (I'm quite time poor atm) thinking how I would approach the plumbing, and come up with this draft: https://github.com/RussKie/gitextensions/commit/caadf00366bd31330330c679e62ac2f6390101f0

In a nutshell, I propose moving data load and parsing into a dedicated class in GitCommands.DiffMergeTools namespace: `CustomDiffMergeToolProvider`. It can also provide a caching mechanism, as caching isn't something its callers should be concerned with. I propose we do a simple caching based in a working directory, as I can't think of a scenario where we have multiple concurrent calls from different repositories in the same process (those calls ultimately will be coming from a UI). NB: I can be totally overlooking some case without realising it.
The UI layer is only responsible for binding the list of custom items, nothing else.

gerhardol: > I really struggle with this API. It mixes multiple concerns - it is a data class, and it also responsible for data load, data parsing, UI binding, and cache busting.
> It also appears completely divorced from diff/merge tool API that live in GitCommands.DiffMergeTools namespace.
> 
> I spent some time (I'm quite time poor atm) thinking how I would approach the plumbing, and come up with this draft: [RussKie@caadf00](https://github.com/RussKie/gitextensions/commit/caadf00366bd31330330c679e62ac2f6390101f0)
> 
> In a nutshell, I propose moving data load and parsing into a dedicated class in GitCommands.DiffMergeTools namespace: `CustomDiffMergeToolProvider`. It can also provide a caching mechanism, as caching isn't something its callers should be concerned with. I propose we do a simple caching based in a working directory, as I can't think of a scenario where we have multiple concurrent calls from different repositories in the same process (those calls ultimately will be coming from a UI). NB: I can be totally overlooking some case without realising it.
> The UI layer is only responsible for binding the list of custom items, nothing else.

This _is_ separate from GE DiffMergeTools that offer tools to Git. The custom tolls lists the tools Git provides.

I would prefer to not use `CustomDiffMergeToolProvider`, the menu setup in LoadCustomDiffMergeTools() is handling `IContainer components` to make sure the ContextMenuStrip is disabled. This is why it is in GitUI. But I can change that if you prefer.

For caching, it takes some time to run the Git command (as discussed with mstv above). This need to be changed in master also if this PR is not merged (by moving CustomDiffMergeTool implementation to RevDiff and Resolve conflicts). As this is such a slow command, I do not want to add this to every time a new repo is opened. The possibility to add a difftool for a repo is something I would like to ignore.
However, as a user I would expect that adding a difftool in GE would take effect immediately.

Locking is needed as several forms wants access the same information. Of course that can be documented too.
(I would like to implement this in the GitCommandCache instead, to have a semaphore on a command with cache requested already running, but there are some caveeats there.)

gerhardol:
@RussKie I pushed a branch that is not fully tested:
https://github.com/gerhardol/gitextensions/pull/new/feature/revdiff-custom-difftool-3

If that is not OK I should make changes in master instead, which then will be what is in this PR except at fewer Forms

gerhardol:
@RussKie Some adjustments, hope this is sufficient

RussKie:
I'm currently overworked and underslept..
I'll try to review over the weekend. Thank you for your patience.

gerhardol:
@RussKie 

If this is not OK I should make changes in master and 3.5 instead, which then will be what is in this PR except at fewer Forms

RussKie:
I'll try to review tomorrow.

GitUI/CustomDiffMergeToolProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                return;
+            }
+
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                // get the tools, possibly with a delay as requesting requires considerable time
+                // cache is shared
+                IEnumerable<string> tools = await (isDiff ? CustomDiffMergeToolCache.DiffToolCache : CustomDiffMergeToolCache.MergeToolCache)
+                    .GetToolsAsync(module, delay);
```
RussKie, line 49: ❗ `tools` enumerated multiple times, especially in the loop.
The collection must be materialised in a form or a list or an array

GitCommands/CustomDiffMergeToolCache.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Load the availble DiffMerge tools and apply to the menus
+        /// </summary>
+        /// <param name="module">The Git module</param>
+        /// <param name="delay">The delay before starting the operation</param>
+        public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay)
+        {
+            if (_tools != null)
```
RussKie, line 52: [nit]
```suggestion
            if (_tools is not null)
```

Few other instances throughout

RussKie (commented):
This looks better 👍 
I'd push for more restructure, but given my current crazy workload, and fact we want to ship it as part of v3.5, I can live with that.

GitUI/CustomDiffMergeToolProvider.cs
gerhardol: CustomDiffMergeToolCache could return a List instead
Changed to ToList()

gerhardol:
> I'd push for more restructure, but given my current crazy work load, and fact we want to ship it as part of v3.5, I can live with that.

Any hints to do before an  approval and a merge is initiated?

GitCommands/CustomDiffMergeToolCache.cs
gerhardol: old habits die slow

RussKie (approved):

gerhardol:
fixed a few changes required after the last changes, squashed and rebased

gerhardol:
@msftbot merge in 48 hours

mstv:
The `GitExtensionsDoc` update to 3.5 slipped in.

gerhardol:
> The `GitExtensionsDoc` update to 3.5 slipped in.

...fixed....

gerhardol:
After the latest restructurings, this PR caused some hangings in the UI while git-difftool --tool-help runs.
The reason for that seem to be that the Git command is running on the UI thread.
This is not what I expected, the await just before has ConfigureAwait(false)
However, ConfigureAwait(false) on the await prior to that await causes the git command to be scheduled on a worker instead.
(And ConfigureAwait(false) on the caller does not schedule the git command off the UI thread.)
It is something with async that I have not understood.
There are other git commands that run on the UI thread, there should probably be other similar changes for other git commands.
@mstv mentioned something about this lately.

Anyway, pushed this change as well as increasing the delay time for ResolveConflicts (the result is cached though, so only first time affected) as well as for FileHistory (separate instance, so no cache).
The delay is not really needed from a responsiveness aspect, but it still competes for resources, so it is probably better to to delay to after normal start is finished.

Last time I tried to abort an automerge it was merged anyway so I squashed and rebased.
The diff command is available here:
https://github.com/gerhardol/gitextensions/commit/4c74effa5cc44e260703782a7ae4129322e59077

mstv (approved):
👍 Now, one can use the GUI, e.g. open the Git Command Log, while the difftools are queried, which did not work before.

gerhardol:
@msftbot merge in 24 hours

gerhardol:
A separate PR for release/3.5 to be created

RussKie:
This change adds ~30sec to integration tests
image

gerhardol:
> This change adds ~30sec to integration tests

This is intentionally run with a delay, the Git command requires some time

Can we run these tests with the functionality off?

PR #8703, opened 2020-12-29
 Submodule browse: Select both first/second commits

gerhardol:
Related to discussions in #8700 how to handle diffs for submodules

## Proposed changes

GE has a decent overview of changes in a submodule, but sometimes I want to see the actual changes in files or more than first/selected commit message. A git-diff --dir-diff will not work with all difftools, not show working dir (unstaged) changes and no intermediate commit messages.

The best difftool is GE itself, when a subrepo is currently opened only the selected commit is selected though. It can be difficult to find the first submodule commit. In the screenshot it was at least the same page, harder if there are more changes.

## Screenshots 

Open submodule in e9eff4bb646ccd355f221c81cbf98e1eeab5a3b4 (or doubleclick).
If the WorkTree was browsed (Commit, RevDiff, side panel later), OldCommit->Worktree would be displayed instead after the change.

image

Before | After
-------|---------
image | image

## Test methodology 

Test updated.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -122,7 +122,14 @@ namespace GitUI.CommandsDialogs
             InitializeComplete();
         }
 
-        public FormBrowse([NotNull] GitUICommands commands, string filter, ObjectId selectCommit = null)
+        /// <summary>
+        /// Open Browse - main GUI including dashboard
+        /// </summary>
+        /// <param name="commands">commands in the current form</param>
+        /// <param name="filter">filter to apply to browse</param>
+        /// <param name="selectedId">Currently (last) selected commit id</param>
+        /// <param name="firstId">First selected commit id (as in a diff)</param>
+        public FormBrowse([NotNull] GitUICommands commands, string filter, ObjectId selectedId = null, ObjectId firstId = null)
             : base(commands)
         {
             InitializeComponent();
@@ -294,10 +301,8 @@ namespace GitUI.CommandsDialogs
                 control.DragDrop += FormBrowse_DragDrop;
             }
 
-            if (selectCommit != null)
-            {
-                RevisionGrid.InitialObjectId = selectCommit;
-            }
+            RevisionGrid.SelectedId = selectedId;
+            RevisionGrid.FirstId = firstId;
 
             InitializeComplete();
             UpdateCommitButtonAndGetBrush(null, AppSettings.ShowGitStatusInBrowseToolbar);
--- a/GitUI/CommandsDialogs/FormCommandlineHelp.resx
+++ b/GitUI/CommandsDialogs/FormCommandlineHelp.resx
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="_NO_TRANSLATE_commands.Text" xml:space="preserve">
-    <value>browse [path] [-filter=] [-commit=]
+    <value>browse [path] [-filter=] [-commit=&lt;selectedSha&gt;[,&lt;firstSha&gt;]]
 about
 add [filename]
 addfiles [filename]
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1160,9 +1160,16 @@ namespace GitUI
             return StartSettingsDialog(owner, CommandsDialogs.SettingsDialog.Pages.GitConfigSettingsPage.GetPageReference());
         }
 
-        public bool StartBrowseDialog(IWin32Window owner = null, string filter = "", ObjectId selectedCommit = null)
+        /// <summary>
+        /// Open Browse - main GUI including dashboard
+        /// </summary>
+        /// <param name="owner">curent window owner</param>
+        /// <param name="filter">filter to apply to browse</param>
+        /// <param name="selectedId">Currently (last) selected commit id</param>
+        /// <param name="firstId">First selected commit id (as in a diff)</param>
+        public bool StartBrowseDialog(IWin32Window owner = null, string filter = "", ObjectId selectedId = null, ObjectId firstId = null)
         {
-            var form = new FormBrowse(this, filter, selectedCommit);
+            var form = new FormBrowse(this, filter, selectedId, firstId);
 
             if (Application.MessageLoop)
             {
@@ -1609,13 +1616,40 @@ namespace GitUI
                 return StartBrowseDialog(null, GetParameterOrEmptyStringAsDefault(args, "-filter"));
             }
 
-            if (Module.TryResolvePartialCommitId(arg, out var objectId))
+            if (TryGetObjectIds(arg, Module, out var selectedId, out var firstId))
             {
-                return StartBrowseDialog(null, GetParameterOrEmptyStringAsDefault(args, "-filter"), objectId);
+                return StartBrowseDialog(null, GetParameterOrEmptyStringAsDefault(args, "-filter"), selectedId, firstId);
             }
 
             Console.Error.WriteLine($"No commit found matching: {arg}");
             return false;
+
+            static bool TryGetObjectIds(string arg, GitModule module, out ObjectId selectedId, out ObjectId firstId)
+            {
+                selectedId = null;
+                firstId = null;
+                foreach (string part in arg.Split(','))
+                {
+                    if (!module.TryResolvePartialCommitId(part, out var objectId))
+                    {
+                        return false;
+                    }
+
+                    if (selectedId == null)
+                    {
+                        selectedId = objectId;
+                    }
+                    else if (firstId == null)
+                    {
+                        firstId = objectId;
+
+                        // just ignore further commits
+                        break;
+                    }
+                }
+
+                return true;
+            }
         }
 
         private static string GetParameterOrEmptyStringAsDefault(IReadOnlyList<string> args, string paramName)
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -790,18 +790,29 @@ namespace GitUI
             return text;
         }
 
-        private async Task OpenSubmoduleAsync()
+        /// <summary>
+        /// Open the currently selected submodule (no checks done) in a new Browse instance
+        /// If the submodule is a diff, both first and currently selected commits are initially selected
+        /// </summary>
+        /// <returns>async Task</returns>
+        public async Task OpenSubmoduleAsync()
         {
             var submoduleName = SelectedItem.Item.Name;
 
             var status = await SelectedItem.Item.GetSubmoduleStatusAsync().ConfigureAwait(false);
+            var selected = SelectedItem.SecondRevision.ObjectId == ObjectId.WorkTreeId
+                ? SelectedItem.SecondRevision.ObjectId.ToString()
+                : status?.Commit.ToString() ?? string.Empty;
+            var first = string.IsNullOrWhiteSpace(status?.OldCommit?.ToString())
+                ? string.Empty
+                : $",{status.OldCommit}";
 
             var process = new Process
             {
                 StartInfo =
                 {
                     FileName = Application.ExecutablePath,
-                    Arguments = "browse -commit=" + status.Commit,
+                    Arguments = $"browse -commit={selected}{first}",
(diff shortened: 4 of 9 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -795,13 +795,16 @@ private async Task OpenSubmoduleAsync()
...
+                ? SelectedItem.SecondRevision.ObjectId
+                : status.Commit;
 
             var process = new Process
             {
                 StartInfo =
                 {
                     FileName = Application.ExecutablePath,
-                    Arguments = "browse -commit=" + status.Commit,
+                    Arguments = $"browse -commit={selected},{status.OldCommit}",
```
mstv, line 807: This code is identical to RevisionDiffControl.cs L459-474. Should we factor out a new method?

GitUI/GitUICommands.cs
```diff
@@ -1609,13 +1609,35 @@ private bool RunBrowseCommand(IReadOnlyList<string> args)
...
+            if (tryGetObjectIds(arg, Module, out var objectIds))
             {
-                return StartBrowseDialog(null, GetParameterOrEmptyStringAsDefault(args, "-filter"), objectId);
+                return StartBrowseDialog(null, GetParameterOrEmptyStringAsDefault(args, "-filter"), objectIds.ToArray());
             }
 
             Console.Error.WriteLine($"No commit found matching: {arg}");
             return false;
+
+            static bool tryGetObjectIds(string arg, GitModule module, out List<ObjectId> objectIds)
```
mstv, line 1620: ?
```suggestion
            static bool TryGetObjectIds(string arg, GitModule module, out List<ObjectId> objectIds)
```

GitUI/CommandsDialogs/FormCommandlineHelp.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="_NO_TRANSLATE_commands.Text" xml:space="preserve">
-    <value>browse [path] [-filter=] [-commit=]
+    <value>browse [path] [-filter=] [-commit=&lt;sha1&gt;[,&lt;sha2&gt;]]
```
mstv, line 121: Unfortunately, the argument option `-filter=` from <= 2012 is inconsistent with most of the others.
So rather?
```suggestion
    <value>browse [path] [-filter=] [--commit &lt;sha1&gt;[,&lt;sha2&gt;]]
```

gerhardol: -commit is also an old option, cannot be changed (probably used from shell extension too)

GitUI/UserControls/FileStatusList.cs
gerhardol: I considered that, but settled with aligning the code
Where to put it?

```
        public async Task OpenSubmoduleAsync(FileStatusItem item, IFullPathResolver fullPathResolver)
        {
            var submoduleName = item.Item.Name;

            var status = await item.Item.GetSubmoduleStatusAsync().ConfigureAwait(false);
            var selected = item.SecondRevision.ObjectId == ObjectId.WorkTreeId
                ? item.SecondRevision.ObjectId
                : status.Commit;

            var process = new Process
            {
                StartInfo =
                {
                    FileName = Application.ExecutablePath,
                    Arguments = $"browse -commit={selected},{status.OldCommit}",
                    WorkingDirectory = _fullPathResolver.Resolve(submoduleName.EnsureTrailingPathSeparator())
                }
            };

            process.Start();
        }
```

GitUI/GitUICommands.cs
gerhardol: OK

GitUI/UserControls/FileStatusList.cs
RussKie: ❗  `IFullPathResolver fullPathResolver` must not be passed as an input argument

GitUI/CommandsDialogs/FormCommandlineHelp.resx
RussKie: I think `--commit=abc123,09982039cbs00808dsfsd08978s` is totally fine

GitUI/GitUICommands.cs
RussKie, line 1620: We unnecessary allocate memory here - we expect maximum of two items, yet we allocate a dynamic list, and then re-allocate the list into an array.
Allocate and return `ObjectId[2]`

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -162,7 +162,7 @@ private RefFilterOptions RefFilterOptions
         internal bool DoubleClickDoesNotOpenCommitInfo { get; set; }
 
         [CanBeNull]
-        internal ObjectId InitialObjectId { private get; set; }
+        internal ObjectId[] InitialObjectIds { private get; set; }
```
RussKie, line 165: Passing arrays around is inherently dangerous, do we expect more than 2 commits to be selected? If not, should we have two properties instead?

gerhardol: More than two is an edge case, like if you want to see a submodule selected with the same commits you have multiselected or to select multiple commits from the cmd line. But that would be separate extensions.

Would passing a list around be OK?

Otherwise new arguments --selected-commit and --first-commit should be added (with backwards compatibility with -commit) should be added

GitUI/CommandsDialogs/FormCommandlineHelp.resx
gerhardol: If --commit (or --selected-commit and --first-commit) is added then -commit should be handled still. Maybe shell extension could be synchronized (not for me that only run private builds...) but vsix and other users could be affected, I do not see all risks. It is hard to change published interfaces.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: If automatically selecting more than two commits is an edge case I propose we have two `ObjectId` properties, thus removing the need to pass arrays around.
I propose we keep `InitialObjectId` as is (at least its functionality, but happy to have it renamed to something like `FirstSelectedCommit`), and add another property  for the second selected commit.

RussKie:
@drewnoakes was there any reason not to make `ObjectId` a `struct`? After all it looks like 5 uints, that can leave on the stack.

GitUI/CommandsDialogs/FormCommandlineHelp.resx
RussKie: This `[-commit=&lt;sha1&gt;[,&lt;sha2&gt;]]` looks like an additive change - i.e. the original functionality should be unaffected. We just allow to append another hash to it.
Am I overlooking something else?

gerhardol: Modifying arguments for -commit is compatible.
@mstv remarked on the other arguments using GNU/Posix standard that single char options have one dash and can often be combined (ls -ar is the same as ls -a -r) and multicharacter options has two dashes (that cannot be combined) (ls --all --reverse).

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: I changed this even if I find that this makes the code harder to maintain.

GitUI/CommandsDialogs/FormCommandlineHelp.resx
mstv: > @mstv remarked on the other arguments using GNU/Posix standard that ... and multicharacter options has two dashes.

But I missed that `-commit=...` has already existed.
Since GE is mainly a GUI tool, I think a cleanup of the sligthly inconsistent format of the CLI arguments is not worth to be cleaned up.

mstv (approved):

RussKie (changes requested):
My only concern is that we are extremely inconsistent in naming: `InitialObjectId`, `FirstObjectId`, `selectedId`, `firstId`, `selectedCommit`, `firstCommit`, `selectedSha1`, `firstSha2`...

We need to be consistent, otherwise this will turn into a undecipherable goo very quickly.
Xml-docs annotating input args would help as well.

gerhardol:
squashed and rebased on master

* Updated review comments
* Changed handling of displaying new submodules (would show diff from worktree to the current commit which may not be expected)
* Removed handling to use empty commits as worktree (use the artificial sha if needed)

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -122,7 +122,14 @@ private FormBrowse()
...
         }
 
-        public FormBrowse([NotNull] GitUICommands commands, string filter, ObjectId selectCommit = null)
+        /// <summary>
+        /// Open Browse - main GUI including dashboard
+        /// </summary>
+        /// <param name="commands">commands in the current form</param>
+        /// <param name="filter">filter to apply to browse</param>
+        /// <param name="selectedId">Currently selected commit id</param>
+        /// <param name="firstId">First commit id in a diff</param>
```
RussKie, line 131: Should we call it `diffId` instead?

RussKie (approved):

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: (selected|first)(Id|Commit) is the common name, except in fileStatusList where First/Second is used (as Selected is not always Second). (Id for ObjectId and Commit for GitRevision).

"selected" is the primary, last selected commit which normally is the second in a diff (to or B)
"first" is the first in a diff, there can be several first before the selected though.

RussKie:
FirstSelectedCommitId and LastSelectedCommitId?

A bit verbose perhaps but provide a good hint to developers.

drewnoakes:
> @drewnoakes was there any reason not to make ObjectId a struct? After all it looks like 5 uints, that can leave on the stack.

Only reason I can think of is the potential for excessive copying. I don't think that'd be an issue in practice, and can be amerliorated by newer C# features such as `readonly struct` and `in` parameter modifiers.

Conversion from class to struct would need to take nulls into account.

drewnoakes:
If someone makes this conversion, please tag me on the PR so I can take a look 👍

gerhardol:
> FirstSelectedCommitId and LastSelectedCommitId? A bit verbose perhaps but provide a good hint to developers.

FirstSelectedId and LastSelectedId in that case as Commit is normallt used for GitRevision.
This will deviate from the normal use use of selectedId/firstId, maybe OK as this is primarily for selection

gerhardol:
Added the handling to side panel too.
If someone agrees, I will add that to this PR too, otherwise it will be a new PR (datastructures changes slightly for submodules in DetailedSubmoduleInfo). Of course, naming depends on the discussion here.

https://github.com/gerhardol/gitextensions/pull/new/feature/browse-submodules-2

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -162,7 +162,8 @@ private RefFilterOptions RefFilterOptions
         internal bool DoubleClickDoesNotOpenCommitInfo { get; set; }
 
         [CanBeNull]
-        internal ObjectId InitialObjectId { private get; set; }
+        internal ObjectId SelectedId { private get; set; }
+        internal ObjectId FirstId { private get; set; }
```
RussKie, line 166: Please add xml-docs here as well

RussKie (approved):

RussKie:
> Added the handling to side panel too.
> If someone agrees, I will add that to this PR too, otherwise it will be a new PR (datastructures changes slightly for submodules in DetailedSubmoduleInfo). Of course, naming depends on the discussion here.
> 
> [gerhardol/gitextensions/pull/new/feature/browse-submodules-2](https://github.com/gerhardol/gitextensions/pull/new/feature/browse-submodules-2)

I think it'd be easier if we merge this, and then look at the next one. Otherwise we may bog down.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Done, squashed

gerhardol:
> I think it'd be easier if we merge this, and then look at the next one. Otherwise we may bog down.

I agree. In some other changes the follow up PR could change the original PR, but this is even separate files.

gerhardol:
@msftbot merge in 48 hours

PR #8707, opened 2020-12-31
FormCommit: remove context menu to copy to message

gerhardol:
## Proposed changes

The "Add selection to commit message" command is inserted to the
FileViewer menu, which is confusing.
Other inserted entries are removed in #7825.
A proper redesign could be done with enabling and eventhandlers,
but usage is considered low and workaround is to copy&paste.

Hotkey could be handled independent of the menu item.

### Before

image

### After

image

## Test methodology 

manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -104,7 +104,6 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _stageFiles = new TranslationString("Stage {0} files");
         private readonly TranslationString _selectOnlyOneFile = new TranslationString("You must have only one file selected.");
 
-        private readonly TranslationString _addSelectionToCommitMessage = new TranslationString("Add selection to commit message");
         private readonly TranslationString _stageSelectedLines = new TranslationString("Stage selected line(s)");
         private readonly TranslationString _unstageSelectedLines = new TranslationString("Unstage selected line(s)");
         private readonly TranslationString _resetSelectedLines = new TranslationString("Reset selected line(s)");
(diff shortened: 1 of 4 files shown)


RussKie (approved):

mstv:
> The "Add selection to commit message" command is inserted to the FileViewer menu, which is confusing.

How does it confuse you?

> workaround is to copy&paste

This feature saves many keypresses / clicks, especially if more than one identifier is used in the commit message.

> usage is considered low
Hotkey could be handled independent of the menu item.

Without a menu item, this feature would become almost undiscoverable.

RussKie:
My 2c.

I've been actively using the app for the past 7-8 years I didn't know we
had this feature until this PR.
To me *personally* this feature looks like yet another feature that was
added for some corner case, without assessing practicality for the majority
of our users.

I'm really curious about use cases in which it is *really* needed, and
where it makes a ctrl+c/ctrl+v combination such a significant effort to
achieve the desired result.

gerhardol:
> > The "Add selection to commit message" command is inserted to the FileViewer menu, which is confusing.
> 
> How does it confuse you?

The implementation where one module adds to the menus of another module. This will be the sole user of this infrastructure in FileViewer. (There are other cross dependencies for FileViewer.)

> > workaround is to copy&paste
> 
> This feature saves many keypresses / clicks, especially if more than one identifier is used in the commit message.

So you are actually using this?
I rather clean this up and remove the corner case with this functionality but can live with it staying.
I intended to but forgot to put the discussion label for the PR.

mstv:
For [subversion-style commit messages](https://subversion.apache.org/docs/community-guide/conventions.html#log-messages), one must jump between the patch view and the commit message very often.

I use it often - also for less extensive commit messages. _I_ don't need a menu item. ;-)

gerhardol:
There is at least one user - I suggest keeping it for now and maybe cleanup later.
It is hard to remove such special functionality...

PR #8708, opened 2020-12-31
FileViewer: Consistently use Stage/Reset  Lines

gerhardol:
## Proposed changes

Current text when staging changes for specific lines is cherry picking and reverting for normal commits, for worktree (index) staging (unstaging) and resetting is used. The Git commands are slightly different but this is unknown for the user, the action is the same.
Especially, the hotkey must be the same. It is very confusing to identify the view as a "reset" or "revert" with separate hotkeys.

It is possible to have separate menu items with separate texts for e.g. reset/revert, but if the hotkey is to be the same, the hotkey handler must be identical so the action in the menu and hotkey will differ.

The names are currently chosen in analogy with cherry-picking/reverting commits, giving the "different name for the same action" for the line manipulation.

See also #7825 that updates the context menu for stage/unstage (including hotkeys not yet available for revert/cherry-pick) as well as #8707

Current usage of cherry-picking and reverting is expected to be relative limited, it is more important when staging commits. 

This PR also moves a few other strings to Strings.cs, preparing for #7825 (but still an improvement if #7825 would be rejected).
The printouts are slightly changed, to be consistent.

### Before

image

image

### After

image

image

## Test methodology 

UI changes only, screenshots

----

### DIFF
--- a/GitUI/Editor/FileViewer.Designer.cs
+++ b/GitUI/Editor/FileViewer.Designer.cs
@@ -99,7 +99,7 @@ namespace GitUI.Editor
             this.cherrypickSelectedLinesToolStripMenuItem.Image = global::GitUI.Properties.Images.CherryPick;
             this.cherrypickSelectedLinesToolStripMenuItem.Name = "cherrypickSelectedLinesToolStripMenuItem";
             this.cherrypickSelectedLinesToolStripMenuItem.Size = new System.Drawing.Size(243, 22);
-            this.cherrypickSelectedLinesToolStripMenuItem.Text = "Cherry pick selected lines";
+            this.cherrypickSelectedLinesToolStripMenuItem.Text = Strings.StageSelectedLines;
             this.cherrypickSelectedLinesToolStripMenuItem.Click += new System.EventHandler(this.cherrypickSelectedLinesToolStripMenuItem_Click);
             // 
             // copyPatchToolStripMenuItem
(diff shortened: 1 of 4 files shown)


GitUI/Translation/English.xlf
```diff
@@ -1337,7 +1337,7 @@ Please make sure git (Git for Windows or cygwin) is installed or set the correct
         <target />
       </trans-unit>
       <trans-unit id="cherrypickSelectedLinesToolStripMenuItem.Text">
-        <source>Cherry pick selected lines</source>
+        <source>Stage selected line(s)</source>
```
gerhardol, line 1340: The translation app should not add these strings to the .xlf (as the strings are translated in Strings.cs).
Similar changes in TranslationStrings are handled correctly

GitUI/Editor/FileViewer.Designer.cs
```diff
@@ -99,7 +99,7 @@ private void InitializeComponent()
             this.cherrypickSelectedLinesToolStripMenuItem.Image = global::GitUI.Properties.Images.CherryPick;
             this.cherrypickSelectedLinesToolStripMenuItem.Name = "cherrypickSelectedLinesToolStripMenuItem";
             this.cherrypickSelectedLinesToolStripMenuItem.Size = new System.Drawing.Size(243, 22);
-            this.cherrypickSelectedLinesToolStripMenuItem.Text = "Cherry pick selected lines";
+            this.cherrypickSelectedLinesToolStripMenuItem.Text = Strings.StageSelectedLines;
```
RussKie, line 102: If you move this out, then it won't get translated second time.

RussKie (approved):

mstv (commented):
LGTM but: Should we consider to `MessageBox.Show` by `MessageBoxes.Confirm / Show`?

GitUI/Editor/FileViewer.Designer.cs
gerhardol: Move out how?
I looked a little at the TranslationApp, did not find an obvious way to fix this.

gerhardol:
@msftbot merge in 48 hours

PR #8713, opened 2021-01-01
TranslationApp: Include "AccessibleName"

gerhardol:
## Proposed changes

#8682 adds use of "AccessibleName" which should be translated.
(No such properties right now).

## Test methodology

Manual

----

### DIFF
--- a/ResourceManager/Xliff/TranslationUtil.cs
+++ b/ResourceManager/Xliff/TranslationUtil.cs
@@ -248,6 +248,8 @@ namespace ResourceManager.Xliff
 
         private static readonly HashSet<string> _translatableItemInComponentNames = new HashSet<string>(StringComparer.Ordinal)
         {
+            "AccessibleDescription",
+            "AccessibleName",
             "Caption",
             "Text",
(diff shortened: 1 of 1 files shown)


ResourceManager/Xliff/TranslationUtil.cs
```diff
@@ -248,6 +248,7 @@ private static bool IsTranslatableItemInBox(PropertyInfo property, object itemOb
 
         private static readonly HashSet<string> _translatableItemInComponentNames = new HashSet<string>(StringComparer.Ordinal)
         {
+            "AccessibleName",
```
RussKie, line 251: We should add [`AccessibleDescription `](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.accessibledescription) as well

RussKie (approved):
👍

ResourceManager/Xliff/TranslationUtil.cs
gerhardol: added and rebased

PR #8716, opened 2021-01-02
Fix high CPU usage of TeamCityAdapter

### LINKED ISSUE #3646: TeamCity integration can cause OutOfMemoryException
I've got a strange crash of GE while staying idle, just leaved the GE's browser window open and came back next day. GE has crashed with OOM exception, but it doesn't look like a regular memory leak:
```
System.OutOfMemoryException: Array dimensions exceeded supported range.
   System.Collections.Generic.List`1[[System.__Canon, mscorlib]].set_Capacity(Int32) 
   System.Collections.Generic.List`1[[System.__Canon, mscorlib]].EnsureCapacity(Int32) 
   System.Collections.Generic.List`1[[System.__Canon, mscorlib]].Add(System.__Canon) 
   System.AggregateException.Flatten() 
   System.Threading.Tasks.TaskExceptionHolder.Finalize()
```

It seems that there is a huge tree of AggregateExceptions of AggregateExceptions of AggregateExceptions..., where each AE contains a bunch of another AEs as inner exceptions and that tree cannot be flattened. I've checked just one path in that tree starting from the root and digging to first inner exception for each node and after 110 intermediate nodes I've finally reached a leaf exception:
```
HttpRequestException: Internal Server Error
TeamCityIntegration.TeamCityAdapter.GetStreamFromHttpResponseAsync(System.Threading.Tasks.Task`1<System.Net.Http.HttpResponseMessage>, System.String, System.Threading.CancellationToken) 
   System.Threading.Tasks.ContinuationResultTaskFromResultTask`2[[System.__Canon, mscorlib],[System.__Canon, mscorlib]].InnerInvoke() 
   System.Threading.Tasks.Task.Execute()
```
Hereby depth of the tree is 110 at least. It's also extremely wide, size of the root AE's inner exceptions list is 134217728.

The repo which I leaved open in GE has TeamCity integration turned on, I know that this TeamCity instance has been under maintenance for a few hours and therefore could be unavailable for a while.

I've saved full process memory dump just in case, so you may request for additional info needed for investigation. Sorry, I cannot just share the dump due to the sensitive content, but I can run some windbg command and share it's output, for example.

GE 2.49.03, CLR 4.6.1087.0, Windows 6.1.7601.

### LINKED ISSUE #8167: Падение приложения (Out Off Memory)
## Current behaviour

В режиме простоя с открытым репозиторием приложение падает с ошибкой OutOffMemory.
Перед падением наблюдается непрерывный быстрый рост занимаемой памяти, вплоть до 1.5Гб

## Expected behaviour

Нормальная работа

## Steps to reproduce

Запустить приложение
Открыть репозиторий
Ждать (от нескольких минут до нескольких суток, так как проблема плавающая)

## Environment

- Git Extensions 3.3.1.7897
- Git 2.26.0.windows.1
- .NET Framework 4.8.3928.0
- DPI 96dpi (no scaling)

[msinfo32_2.zip](https://github.com/gitextensions/gitextensions/files/4710755/msinfo32_2.zip)

Dump падения не приложить. в Архиве 7z он 65Мб

### LINKED ISSUE #8281: Constant CPU  usage and thread leak when TeamCityAdapter fails to connect to TC server
## Current behaviour

I noticed that, with recent versions, GitExtensions.exe usually consumes a constant fraction of cpu when idle after being opened a while. On my machine (24c/48t) that corresponds to 5.9% or ~2.8 cpu threads.

I found similar issues that seem to be caused by issues with GPU drivers but it did not seem to match my behaviour (running on AMD ThreadRipper, which does not have a integrated GPU).

Using dotTrace to capture a timeline trace, I find that there are multiple thousand threads all calling TeamCityAdapter.GetBuilds() every 2 sec (see below for more details)

## Expected behaviour

GitExtensions not spawning thousands of threads and using my CPU :)

## Steps to reproduce

- GitExtensions configured with TeamCity build server integration in `GitExtensions.Settings` file
- TeamCity uses NTLM auth.
- Open GitExtension on a repo with that file.
- I DO NOT see the build status and label indicators in the main GE view (even though they are checked in the View menu), nor do I see the TeamCity tab when view a commit.
- Let it run idle for a few minutes.
- GitExtensions.exe process consumes a fixed amount of cpu cores on my machine.
- Use JetBrains dotTrace to capture a Timeline of the running git-extensions.exe process (without touching it)
- Observe thousands of CLR Worker threads calling TeamCityAdpater.GetBuilds() every 2 sec (see below)

## Screenshots

Constant cpu use (5.9% on my machine is 2.8 cores, 0.059*48)
image

Build server integration settings: (verified project name and build if filter they are valid)
image

View of dotTrace, filtering on `GetBuilds()`. The scroll bar is so small that this screenshot represents less than 1% of the total number of threads.
image

Zoomed view of the callstack:
image

I don't remember this being an issue with older version (> 6 months old).

## Environment

- GitExtensions version: 3.4.2.9737 (but same issue with recent versions)
- GIT version: 2.24.1.windows.2
- OS version: Windows 10 2004
- .NET version: .NET 4.8, VS 2019 preview, .NET Core SDK 5.0, etc..
- AMD ThreadRipper 3960X, 64 GB RAM

## Diagnostics

I used JetBrains dotTrace to capture a trace of GE while it was running, and ... I found that there are literally multiple thousand threads all in a 2 second loop trying to call TeamCityAdapter.GetBuilds(). The dotTrace snapshot for about 30 sec of capture takes about 1GB and the dotTrace Timeline Viewer was almost not able to display the trace.

All the threads are similar:
- They are CLR Worker threads
- They all spend their time waiting, but wake up every ~2 seconds.
- The have the same callstack: `Scheduler.Invoke()` => `TeamCityAdapter.GetBuilds()` => `TeamCityAdapter.ObserveBuilds()` => `OnError()`

I *think* that one issue could be that the NTLM auth with the TeamCity server fails, and the code does not properly handle the case and respawn another worker thread ?

The full dotTrace is 1GB, compressed down to 130 MB with 7zip. Too big to attach here.

matyasbach:
Fixes #8281 and possibly also fixes #3646 and fixes #8167

Exceptions thrown from TeamCityAdapter caused immediate retry of failing operations -> spawning lot of new threads without any delay.

## Before
These cases caused high CPU usage:
- incorrect (nonexistent) TC build ID
- network or TC server outage

## After
CPU usage stays minimal. Build server is queried in defined time intervals even when responding errors or not communicating. 

## Test methodology 
Manually testing TeamCity integration and:
- simulating network/VPN outages
- set incorrect build ID

Other build server types could be affected and were not tested.

----

### DIFF
--- a/GitUI/BuildServerIntegration/BuildServerWatcher.cs
+++ b/GitUI/BuildServerIntegration/BuildServerWatcher.cs
@@ -110,9 +110,9 @@ namespace GitUI.BuildServerIntegration
                                                         anyRunningBuilds = true;
                                                         shouldLookForNewlyFinishedBuilds = true;
                                                     })
-                                               .Retry()
-                                               .Concat(delayObservable)
+                                               .OnErrorResumeNext(delayObservable)
                                                .Finally(() => anyRunningBuilds = false)
+                                               .Retry()
                                                .Repeat()
                                                .ObserveOn(MainThreadScheduler.Instance)
                                                .Subscribe(OnBuildInfoUpdate)
--- a/contributors.txt
+++ b/contributors.txt
@@ -146,3 +146,4 @@ YYYY/MM/DD, github id, Full name, email
 2020/10/07, BrianFreemanAtlanta, Brian Freeman, freeman(at)carnegie.com
 2020/10/21, wischi-chr, Christian Wischenbart, christian.wischenbart@gmail.com
 2020/11/05, cristianst85, Cristian Stoica, cristianstoica85(at)gmail.com
+2021/01/02, matyasbach, Matyáš Bach, matyasbach(at)centrum.cz


RussKie (approved):

RussKie:
Thank you.

Will wait for a feedback from users reported the issue before merging.

alastairtree:
Have tested the teamcity integration and it still works. It prompted me for credentials correctly (for the first time in ages) and handles network connection/dissconnection gracefully on refresh where it seemed more flakey beforehand. So no regressions and and looks like a promising fix. Thank you!

RussKie:
@msftbot merge in 48 hours

gerhardol:
For Jenkins, this PR causes a regression: New builds are not detected.
Quite hard to debug, there are few public Jenkins servers.

RussKie:
> this PR causes a regression: New builds are not detected.

Do we need to roll this back for 3.5?

gerhardol:
You can manually refresh so there is a workaround.
The share of Jenkins vs TeamCity users decides the urgency... 

I will try to investigate and create an issue

matyasbach:
Ups, sorry for regression :'(

And yes, it's really hard to debug. I spent another few hours now trying to debug/fix high CPU bug another way without success. 

It (#8281) was introduced by one or both of these PRs - #7452, #7485. I'm not experienced with neither Rx nor Microsoft.VisualStudio.Threading so if anyone can spot what's wrong in _BuildServerWatcher.cs_ or point me in the right direction I'll be happy to try it and test it against TC build server.

gerhardol:
AppVeyor likely have the same problem as Jenkins (GetFinishedBuildsSince() is always empty, all builds retrieved in GetRunningBuilds()).
TC and others have the fullDayObservable as a backup.

I do not get any AppVeyor builds with 3.4.3 or 3.3 either, (I have had that for some time, I thought it was after som experiments) and cannot get it working.
Not really any debug possibilities with Jenkins, will try something.

RussKie:
I have no problems with AppVeyor:
image

gerhardol:
> I have no problems with AppVeyor:

See #8828

PR #8718, opened 2021-01-03
Handle external I/O exceptions that user can fix

RussKie:
Relates to #8358

## Proposed changes

- Move user-facing exceptions to UI layer because these exceptions include localisable strings, that are only available at the UI layer.
- Handle non-app related I/O exceptions that a user can resolve

### Before

image

### After

image

NOTE: the current implementation of Task Dialog can't size itself to content, the .NET 5.0 version provided by Windows Forms SDK can.

## Test methodology 

Manually by:
- creating an index.lock file,
- locking the file (e.g. open in MS Excel)
- invoke the functionality through _Repository > Git Maintenance > Delete index.lock_
----

### DIFF
--- a/GitUI/NBugReports/BugReporter.cs
+++ b/GitUI/NBugReports/BugReporter.cs
@@ -18,24 +18,24 @@ namespace GitExtensions
         private static IntPtr OwnerFormHandle
             => OwnerForm?.Handle ?? IntPtr.Zero;
 
-        private static string FormatText(ExternalOperationException ex, bool canRaiseBug)
+        private static string FormatText(ExternalOperationException exception, bool canRaiseBug)
         {
             StringBuilder sb = new StringBuilder();
 
             // Command: <command>
-            if (!string.IsNullOrWhiteSpace(ex.Command))
+            if (!string.IsNullOrWhiteSpace(exception.Command))
             {
-                sb.AppendLine($"{Strings.Command}: {ex.Command}");
+                sb.AppendLine($"{Strings.Command}: {exception.Command}");
             }
 
             // Arguments: <args>
-            if (!string.IsNullOrWhiteSpace(ex.Arguments))
+            if (!string.IsNullOrWhiteSpace(exception.Arguments))
             {
-                sb.AppendLine($"{Strings.Arguments}: {ex.Arguments}");
+                sb.AppendLine($"{Strings.Arguments}: {exception.Arguments}");
             }
 
             // Working directory: <working dir>
-            sb.AppendLine($"{Strings.WorkingDirectory}: {ex.WorkingDirectory}");
+            sb.AppendLine($"{Strings.WorkingDirectory}: {exception.WorkingDirectory}");
 
             if (canRaiseBug)
             {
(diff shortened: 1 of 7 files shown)


mstv (commented):
I think we do not need another exception type.
File I/O _is_ an external operation, hence `UserExternalOperationException` with e.g.:
Context: "delete index.lock" (could be omitted in this case because this example is trivial)
Command: "DeleteFile"
Arguments: "path/index.lock"
WorkDir: "path/to/repo" (can be different in case of worktrees)

If you want to add `ThreadHelper.AssertOnUIThread();` everywhere, it should be done in the contructor or in a factory method.

RussKie:
> If you want to add `ThreadHelper.AssertOnUIThread();` everywhere, it should be done in the contructor or in a factory method.

I don't think this will be correct - exceptions can be thrown on any thread at any layer. But at the UI layer in order to keep the app alive (where practical) we must ensure the exception is (re-)thrown at the UI thread.

RussKie:
Reworked:

image

GitUI/NBugReports/BugReporter.cs
```diff
@@ -46,85 +47,73 @@ private static string FormatText(ExternalOperationException ex, bool canRaiseBug
...
-                InstructionText = ex.Context,
-                Caption = Strings.CaptionFailedExecute,
+                Text = text,
+                InstructionText = instructionText,
+                Caption = Strings.Error,
                 Icon = TaskDialogStandardIcon.Error,
                 Cancelable = true,
             };
+
+            if (!(exception is null))
```
mstv, line 105: Nit:
```suggestion
            if (exception != null)
```
will be easier to replace when enabling C#9 features.

mstv (approved):
Looks good.
In a follow-up PR  (or here), we could eliminate `FileDeleteException` and substitute `ExternalOperationException` for it.

RussKie:
> we could eliminate `FileDeleteException` and substitute `ExternalOperationException` for it.

No, I don't believe we can.
`ExternalOperationException` is UI-layer user-facing exception that contains localisable strings, pushing this exception up the stack will require us have localisable resources pushed up the stack as well.

mstv:
> `ExternalOperationException` is UI-layer user-facing exception that contains localisable strings, pushing this exception up the stack will require us have localisable resources pushed up the stack as well.

Fair point. I agree that the `UserExternalOperationException` with its `Context` belongs to the UI layer and is bound to the specific operation.
But what would be the implication of not to use `ExternalOperationException`? We add exception classes just to be able to localize them in the UI layer afterwards? So rather eliminate the `FileDeleteException` and throw the original exception up the stack, whose message *is* localized.
This raises the next point which I missed to notice in the review: We do not display the original exception message. It will be helpful to know whether the file was just missing or locked or inaccessible.
I still think it would be good to wrap any exception from a file operation into an `ExternalOperationException`. Perhaps we do not need to localise its `Command` (would be e.g. "DeleteFile" in this case) or else do it in the `BugReporter`.
I will analyze why we have the `FileDeleteException` and prepare a proposal for the further discussion.

Issue #8698, opened 2020-12-24
Incorrect borders when app run in >100% scaling factor


RussKie:
## Current behaviour

https://github.com/gitextensions/gitextensions/pull/8557 introduced new margins and borders, however those margins and borders are not correctly calculated in scaled environments, i.e. scale factor >100%.
For example:
* 100%
![image](https://user-images.githubusercontent.com/3169265/101113313-cabc7d00-35e7-11eb-8d76-a33e45215dca.png)

* 200%
![image](https://user-images.githubusercontent.com/460196/103076153-6597ee00-45cd-11eb-8af6-d92edb9d6f55.png)

## Environment

- Git Extensions 33.33.33
- Build f5362ec637de01d40be55f5d0abfbbc7b6e9d828
- Git 2.28.0.windows.1
- Microsoft Windows NT 10.0.17134.0
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

ivangrek:
If I delete this line, it works as it should.
Will it be right?

`this.AutoScaleDimensions = new System.Drawing.SizeF(96F, 96F);`

RussKie:
If it works...

pmiossec:
Tested. It indeed fix the boders width. But a lot of forms have the same value....

RussKie:
I tried a blanket removal of `this.AutoScaleDimensions = new System.Drawing.SizeF(x, y)`, and it doesn't appear to be a safe change. Some dialogs become totally unusable (e.g. `FormCreateWorktree`).
There has to be another way, investigating.

RussKie:
There's definitely something in our code that scales paddings 4(!) times: 1 -> 2 -> 3 -> 4 -> 6.
```cs
this.MainSplitContainer.Panel1.ResumeLayout(false); // 1 -> 2px
...
this.MainSplitContainer.ResumeLayout(false); // 2 -> 3px
...
this.toolPanel.ResumeLayout(false); // 3 -> 4px
...
this.ResumeLayout(false); // 4 -> 6px
```

I have created a brand new app, copied the `FormBrowse` designer file (removing or renaming non Windows Forms controls), and observed only a single padding scaling: 1 -> 2px at `this.ResumeLayout(false);` call.

ivangrek:
> There's definitely something in our code that scales paddings 4(!) times: 1 -> 2 -> 3 -> 4 -> 6.

It is interesting!

ivangrek:
[...]

PR #8719, opened 2021-01-03
Revert ability for move and show/hide toolbars

### LINKED ISSUE #8680: Toolbar position is not saved or restored
## Current behaviour

Since #8572, I [always](https://github.com/gitextensions/gitextensions/pull/8572#issuecomment-734340761) end up with this toolbar at startup:
image

Even if I move the toolbars. Even when only on instance is opened.

Even with #8679 fix (to only save the toolbar settings)

## Expected behaviour

I can configure the toolbar position and it is saved (or restored)  well.

## Steps to reproduce

1. Get a dev version of GE (portable from CI or your custom version) and launch it
2. Change toolbar setup
3. Restart
4. See toolbar position configuration is the same as initally

## Screenshots

See above.

Never since the feature was introduced... 

## Environment

- Git Extensions 33.33.33
- Build 3245408651bd658ee636e7c3b57f3a02bbe4abf4
- Git 2.28.0.windows.1
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

## Diagnostics

The main goal to opening this issue is:

How can I debug it easily?
Where the configuration data are stored?

RussKie:
Due to discovered layout and placement instability of toolbars
reverting most of the functionality added in #8572.

Closes #8680

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -270,6 +270,8 @@ namespace GitUI.CommandsDialogs
 
             ManageWorktreeSupport();
 
+            WorkaroundToolbarLocationBug();
+
             var toolBackColor = SystemColors.Window;
             var toolForeColor = SystemColors.WindowText;
             BackColor = toolBackColor;
@@ -311,6 +313,16 @@ namespace GitUI.CommandsDialogs
             RevisionGrid.ToggledBetweenArtificialAndHeadCommits += (s, e) => FocusRevisionDiffFileStatusList();
 
             return;
+
+            void WorkaroundToolbarLocationBug()
+            {
+                // Layout engine bug (?) which may change the order of toolbars
+                // if the 1st one becomes longer than the 2nd toolbar's Location.X
+                // the layout engine will be place the 2nd toolbar first
+                toolPanel.TopToolStripPanel.Controls.Clear();
+                toolPanel.TopToolStripPanel.Controls.Add(ToolStripFilters);
+                toolPanel.TopToolStripPanel.Controls.Add(ToolStripMain);
+            }
 
             void FocusRevisionDiffFileStatusList()
             {
@@ -615,18 +627,7 @@ namespace GitUI.CommandsDialogs
 
         protected override void OnLoad(EventArgs e)
         {
-            // Removing the main menu before restoring the toolstrip layouts
-            // as it *does* randomly change the order of the defined items in the menu
-            Controls.Remove(mainMenuStrip);
-
-            toolPanel.TopToolStripPanel.SuspendLayout();
-            ToolStripManager.LoadSettings(this, "toolsettings");
-            toolPanel.TopToolStripPanel.ResumeLayout();
-
-            // Add the menu back
-            Controls.Add(mainMenuStrip);
-
-            _formBrowseMenus.CreateToolbarsMenus(ToolStripMain, ToolStripFilters);
+            _formBrowseMenus.CreateToolbarsMenus(ToolStripMain, ToolStripFilters, toolStripScripts);
 
             HideVariableMainMenuItems();
             RefreshSplitViewLayout();
@@ -689,10 +690,6 @@ namespace GitUI.CommandsDialogs
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
-            // Remove the main menu to prevent its state being persisted
-            Controls.Remove(mainMenuStrip);
-
-            ToolStripManager.SaveSettings(this, "toolsettings");
             SaveApplicationSettings();
 
             foreach (var control in this.FindDescendants())
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -320,6 +320,7 @@ namespace GitUI.CommandsDialogs
                 // if the 1st one becomes longer than the 2nd toolbar's Location.X
                 // the layout engine will be place the 2nd toolbar first
                 toolPanel.TopToolStripPanel.Controls.Clear();
+                toolPanel.TopToolStripPanel.Controls.Add(ToolStripScripts);
                 toolPanel.TopToolStripPanel.Controls.Add(ToolStripFilters);
                 toolPanel.TopToolStripPanel.Controls.Add(ToolStripMain);
             }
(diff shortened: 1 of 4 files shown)


gerhardol (approved):
works for me

pmiossec:
In my case, that doesn't solve the main problem I introduced with #8572, that is that the "Branch/filter" toolbar is displayed before the "Main" toolbar:

image

So, with this PR, this is somewhat worse because this layout is now no more changeable...

@RussKie Do you know where the toolbar positions are saved? I would like to see if the values are well saved...

RussKie:
> Do you know where the toolbar positions are saved? I would like to see if the values are well saved...

I don't think this has anything to do with the saved data, as it is no longer used - I've removed all associated code.
This may be an artifact of changes in the designer though

RussKie:
@pmiossec I've run this on two different PCs and haven't observed any issues.
Are you having issues in a dev version launched from VS? Can you try run `git clean -xdf` and then `.\build.cmd -launch` and see if you still get issues?
The settings are stored in a local user config located somewhere like this: C:\Users\<user>\AppData\Local\Git_Extensions\GitExtensions.exe_Url_<some crazy string>\33.33.33.0\user.config but I don't believe it has any relevance now, as we no longer persist/load the data from it.

pmiossec:
>Are you having issues in a dev version launched from VS? Can you try run git clean -xdf and then .\build.cmd -launch and see if you still get issues?

Created a new worktree (so clean workspace) and I have this behavior (still the same):

bug_toolbar

>The settings are stored in a local user config located somewhere like this: C:\Users\<user>\AppData\Local\Git_Extensions\GitExtensions.exe_Url_<some crazy string>\33.33.33.0\user.config but I don't believe it has any relevance now, as we no longer persist/load the data from it.

I will still have a look...

pmiossec:
>The settings are stored in a local user config located somewhere like this: C:\Users<user>\AppData\Local\Git_Extensions\GitExtensions.exe_Url_\33.33.33.0\user.config but I don't believe it has any relevance now, as we no longer persist/load the data from it.

What didn't work:
* removing this `user.config` file
* removing the `WindowPositions.xml` file

But changing the DPI scale to 100% instead of 200% fix the toolbar positioning.
Another DPI scale problem 😭

RussKie:
So any latest build renders correctly on <200%, and incorrect on 200%?

pmiossec:
> So any latest build renders correctly on <200%, and incorrect on 200%?

No, it took me some time to find the reason, so I just had some time to test 100% and 200% (that I use every days) with the code corresponding to this PR. 

I can just assert that every build since  the introduction of the 2 toolbars (with this revert or not) display badly 200%.

So before merging this revert, we have to test different dpi scaling to better understand. 

And if doesn't work, perhaps we will just have to revert the cut in 2 toolbars (until we find a fix)

pmiossec:
Quick test: only 100% display toolbars well. 125%, 150%, 175%, 200% and 225% failed (and I stopped here)

RussKie:
I can repro the issue, thank you. Investigating

RussKie:
Pushed a possible workaround, please try it.

RussKie:
I can't seem to repro the issue in a standalone repro, so quite likely the origin of the bug is same as for https://github.com/gitextensions/gitextensions/pull/8572 and https://github.com/gitextensions/gitextensions/issues/8698.

RussKie:
It took me awhile, but I got a clean repro: https://github.com/RussKie/Test-ToolStripOrder

RussKie:
@pmiossec have had a chance to test?

pmiossec:
>It took me awhile, but I got a clean repro: https://github.com/RussKie/Test-ToolStripOrder
> @pmiossec have had a chance to test?

Yes, I reproduce with this repository

RussKie:
Were you able to test the latest fix?

pmiossec:
> 
> Were you able to test the latest fix?

Which fix?

If that's the 2nd commit (f362cd397fea0972624765e95118222ba9edcb25), after a `msbuild` and run, that doesn't work. Result:

image

RussKie:
> Which fix?
> 
> If that's the 2nd commit (f362cd397fea0972624765e95118222ba9edcb25), after a `msbuild` and run, that doesn't work. Result:

There are only 2 commits in this PR, which one were you testing?
image

I've tested the fix at 150% and 200% without any issues.

pmiossec (approved):
After a new test, work as expected.

pmiossec:
2 things about the cutting of the toolbars.

1. I find it strange to have the user scripts, that are added to the "Standard" toolbar, to be displayed before the "Filters" toolbar, like that:

image

I think we should create a new toolbar and display it after the 2 other toolbars to get something like that (so that we kept the toolbar layout in the current release):
image


2. I don't really understand why this PR revert also the hiding of the toolbar that works well (now that we discovered what was the cause of the toolbar bad placement and that you provided a workaround).
As you kept the cutting of the toolbar in 2 (i.e. that you didn't revert the whole development), perhaps we could at least kept one feature allowed by this cutting, no!?!

Please have a look at my branch that did this 2 things:

https://github.com/pmiossec/gitextensions/tree/new_toolbar_for_scripts

RussKie:
> Please have a look at my branch that did this 2 things:
> 
> [https://github.com/pmiossec/gitextensions/tree/new_toolbar_for_scripts](https://github.com/pmiossec/gitextensions/tree/new_toolbar_for_scripts?rgh-link-date=2021-01-17T00%3A08%3A19Z)

Thank you, works for me :)
Though the new toolbar must look the same as the other two:
image

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -312,6 +314,17 @@ public FormBrowse([NotNull] GitUICommands commands, string filter, ObjectId sele
 
             return;
 
+            void WorkaroundToolbarLocationBug()
+            {
+                // Layout engine bug (?) which may change the order of toolbars
+                // if the 1st one becomes longer than the 2nd toolbar's Location.X
+                // the layout engine will be place the 2nd toolbar first
+                toolPanel.TopToolStripPanel.Controls.Clear();
+                toolPanel.TopToolStripPanel.Controls.Add(ToolStripScripts);
```
gerhardol, line 323: If the order is left-to-right display order, should the order consistently be Filters, Sscripts, Main?

and I would prefer to remove/add button by button instead, splitting the toolbar is not adding much...

gerhardol (commented):
+1

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: > splitting the toolbar is not adding much

It is adding if a) toolbar can be moved, and b) toolbars can be shown/hidden. "a" obviously doesn't work at this stage, but "b" is working just fine.

RussKie: > should the order consistently be Filters, Sscripts, Main?

The order is: Main, Filters, Scripts.

RussKie:
> 1. I don't really understand why this PR revert also the hiding of the toolbar that works well (now that we discovered what was the cause of the toolbar bad placement and that you provided a workaround).
>    As you kept the cutting of the toolbar in 2 (i.e. that you didn't revert the whole development), perhaps we could at least kept one feature allowed by this cutting, no!?!

We also need to restore persistence of shown/hidden state as well.

RussKie:
Merging, we can deal with saving/restoring visibility of the toolbars separately.

PR #8721, opened 2021-01-05
Process.Start with ExternalOperationException

mstv:
Contributes to #7795, extracted from #8278

## Proposed changes

- Use class `Executable` for `OsShellUtil` implementation
- Use `OsShellUtil` functions instead of plain `Process.Start` where applicable

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build cca84cd8551e5be614bf215bd77d93664b69ece6
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1829,14 +1829,8 @@ namespace GitUI.CommandsDialogs
 
         private void UserManualToolStripMenuItemClick(object sender, EventArgs e)
         {
-            try
-            {
-                // Point to the default documentation, will work also if the old doc version is removed
-                Process.Start("https://git-extensions-documentation.readthedocs.org");
-            }
-            catch (Win32Exception)
-            {
-            }
+            // Point to the default documentation, will work also if the old doc version is removed
+            OsShellUtil.OpenUrlInDefaultBrowser(@"https://git-extensions-documentation.readthedocs.org");
         }
 
         private void CleanupToolStripMenuItemClick(object sender, EventArgs e)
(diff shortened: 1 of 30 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1990,19 +1990,12 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
 
         private void TranslateToolStripMenuItemClick(object sender, EventArgs e)
         {
-            Process.Start("https://github.com/gitextensions/gitextensions/wiki/Translations");
+            OsShellUtil.OpenUrlInDefaultBrowser(@"https://github.com/gitextensions/gitextensions/wiki/Translations");
         }
 
         private void FileExplorerToolStripMenuItemClick(object sender, EventArgs e)
         {
-            try
```
gerhardol, line 1998: Any reasoning to remove this try-catch?
I am lost in the previous discussions...

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1832,7 +1832,7 @@ private void UserManualToolStripMenuItemClick(object sender, EventArgs e)
             try
```
gerhardol, line 1832: Other try-catch removed, this kept?

gerhardol (commented):
Just questions about try-catch changes
Open, OpenAs are handled in OsShellUtil

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: We ultimately want all exceptions to be routed via the `BugReporter`, which will present a user with a message about the error. One caveat - we want exceptions to be on the UI thread, else exceptions are considered fatal and will lead to the app closed.

So generally speaking the only `try..catch` scenario we should have in the code is catch an exception which we can 
- ignore, or
- wrap as a BugReporter-known exception and rethrow.

GitUI/OsShellUtil.cs
```diff
@@ -33,7 +48,7 @@ public static void OpenUrlInDefaultBrowser(string url)
         {
             if (!string.IsNullOrWhiteSpace(url))
             {
-                Process.Start(url);
+                new Executable(url).Start(useShellExecute: true);
```
RussKie, line 51: This may not work in .NET, despite [the docs](https://docs.microsoft.com/en-us/troubleshoot/dotnet/csharp/start-internet-browser) saying it should, and having it working in .NET Framework, I stumble across it yesterday in fact.

I had to add the `verb` option to get it working:

```diff
-            using var process = new Process
-            {
-                EnableRaisingEvents = false,
-                StartInfo = { FileName = uri.AbsoluteUri }
-            };
-            process.Start();
+            var ps = new ProcessStartInfo(uri.AbsoluteUri)
+            {
+                UseShellExecute = true,
+                Verb = "open"
+            };
+            Process.Start(ps);
         }
 }
```

RussKie (approved):

GitUI/CommandsDialogs/FormBrowse.cs
mstv: Good catch. This one slipped through. I have removed it.

mstv:
@msftbot merge in 1 day

gerhardol (approved):

PR #8727, opened 2021-01-07
ResolveConflicts: Incorrect evaluation of mergetool path

gerhardol:
Regression: https://github.com/gitextensions/gitextensions/pull/8690#issuecomment-756021266

## Proposed changes

Correct annotation for PathUtil.TryFindFullPath(), revert change to evaluate path in #8690

The mergetool path was always set to empty so the Git mergetool handling was used (not the GE special implementation to check results etc).

## Test methodology 

manual

----


mstv (approved):
Works for me.

gerhardol:
@msftbot merge in 24 hours

Issue #8728, opened 2021-01-07
Use C# 7/8/9 features


gerhardol:
## Feature description

The source code should use new C3 features (where possible, some features requires .NET5 etc)
Changes should be globally applied to reduce noise.

### https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7
* [x] is null (>700 occurrences)
Allows static analysis and avoids issues overriding assignment: https://intellitect.com/check-for-null-not-null/

### https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8
* [x] Using declarations, to reduce nesting (About 150 occurrences)
* [x] Null-coalescing assignment (about 21 occurrences)

### https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9
Requires `Directory.Build.props` update first:
```
<LangVersion>9</LangVersion>
```
* [x] is not null (about 900 occurrences)
* [ ] Pattern matching enhancements (other than is not null) (about 70 occurrences) #8738

## Environment

- Git Extensions 33.33.33
- Build c0cf25be75a1c69942508118f625c41126f6b005
- Git 2.30.0.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

drewnoakes:
I'd add to this list: Annotate code for nullability.

gerhardol:
> I'd add to this list: Annotate code for nullability.

Can this be done in a bulk change somehow?
Replace ReSharper annotations?

vbjay:
Regex find and replace in all cs files.

drewnoakes:
> Can this be done in a bulk change somehow?
> Replace ReSharper annotations?

It would be a good place to start, however I would not advise trying to annotate everything in bulk. I have tried this before on a few occasions and it gets too confusing. The C# NRT language feature is more strict than the R# feature, and will generate plenty of warnings.

drewnoakes:
I annotated the `GitCommands` project.

> there will be a number of bugs uncovered that will lead to changes.

Agreed. Also it highlights places where the logic is non-obvious or overly permissive of null values (which happens a lot in GE). I've fixed a few simple bugs, but mostly wanted to capture the current state of null values.

PR #8729, opened 2021-01-07
C#8 using to reduce nesting

gerhardol:
Part of #8728

## Proposed changes
 
using declarations, to reduce nestings
A few variables have slightly wider declaration now
A few uses of using cannot be changed, like in switch statements.

## Test methodology 

manual, code review
Directory diff with kdiff3 works OK

----

### DIFF
--- a/GitUI/Theming/Renderers/ButtonRenderer.cs
+++ b/GitUI/Theming/Renderers/ButtonRenderer.cs
@@ -163,11 +163,11 @@ namespace GitUI.Theming
                         new Point(x2, y3),
                         new Point(x3, y1),
                     };
-                    using (var checkPen = new Pen(foreColor, DpiUtil.Scale(1.5f)))
-                        using (ctx.HighQuality())
-                        {
-                            ctx.Graphics.DrawLines(checkPen, points);
-                        }
+                    using (ctx.HighQuality())
+                    {
+                        using var checkPen = new Pen(foreColor, DpiUtil.Scale(1.5f));
+                        ctx.Graphics.DrawLines(checkPen, points);
+                    }
 
                     break;
             }
--- a/GitUI/Theming/Renderers/ButtonRenderer.cs
+++ b/GitUI/Theming/Renderers/ButtonRenderer.cs
@@ -164,10 +164,10 @@ namespace GitUI.Theming
                         new Point(x3, y1),
                     };
                     using (var checkPen = new Pen(foreColor, DpiUtil.Scale(1.5f)))
-                    using (ctx.HighQuality())
-                    {
-                        ctx.Graphics.DrawLines(checkPen, points);
-                    }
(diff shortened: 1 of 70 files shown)


GitUI/Theming/Renderers/HeaderRenderer.cs
```diff
@@ -11,17 +11,16 @@ internal class HeaderRenderer : ThemeRenderer
...
         {
-            using (var ctx = CreateRenderContext(hdc, pcliprect))
+            using var ctx = CreateRenderContext(hdc, pcliprect);
+            switch ((Parts)partId)
             {
-                switch ((Parts)partId)
-                {
-                    case Parts.None:
+                case Parts.None:
                     {
```
mstv, line 18: Why not unindent this code block, too - and a few more below?

GitUI/Theming/Renderers/ButtonRenderer.cs
```diff
@@ -164,10 +164,10 @@ private int RenderCheckBox(Context ctx, State.CheckBox stateid, Rectangle prect)
...
-                    using (ctx.HighQuality())
-                    {
-                        ctx.Graphics.DrawLines(checkPen, points);
-                    }
+                        using (ctx.HighQuality())
+                        {
+                            ctx.Graphics.DrawLines(checkPen, points);
+                        }
 
                     break;
```
mstv, line 172: For consistency with below changes:
```
                    using (ctx.HighQuality())
                    {
                        using var checkPen = new Pen(foreColor, DpiUtil.Scale(1.5f));
                        ctx.Graphics.DrawLines(checkPen, points);
                    }
                    break;
```

mstv (approved):

GitUI/Theming/Renderers/HeaderRenderer.cs
gerhardol: I only left this to make sure someone reviewed - or it was something where autoindent is not working properly and I missed to to fix this manually after reverting.
Good catch, 
Pushed fixup.

GitUI/Theming/Renderers/ButtonRenderer.cs
gerhardol: I missed this snippet, thanks.

mstv:
> A few variables have slightly wider declaration now

I agree that these should be uncritical.

RussKie (approved):

RussKie:
Thank you

gerhardol:
@RussKie Merged with a fixup... Is it possible to block that?

pmiossec:
https://github.com/marketplace/actions/block-fixup-commit-merge

RussKie:
Yes, sorry, realised after hit the merge button.

RussKie:
Thank you @pmiossec. I installed the check:
image

PR #8730, opened 2021-01-07
Improved text and icon on StageAll and commit when Filter active #8596

keco2:
Followup on #8596 

The change affects only visual elements (i.e. no behavioural change) and is related only to the CommitForm when there are no staged files and the filefilter is active.

## Proposed changes

- 2 misleading texts updated to display either "All" or "Filtered" words based on the filter:
  -  if the Filter is empty (i.e. inactive), use the "All" word
  - otherwise use the "Filtered" word
- change the icon on the StageAll button interactively based on the filefilter:
  -  if the Filter is empty (i.e. inactive), use the classic StageAll picture
  - otherwise use another (new) picture with funnel

### Before

image
image

### After

StageAll (TooltipText/icon) - without filefilter:
image

StageAll (TooltipText/icon) - with filefilter:
image

OnCommitClick (message) - without filefilter:
image

OnCommitClick (message) - with filefilter:
image

## Test methodology 

- change of text in the MessageBox on Commit: Manual
- change of ToolTipText on StageAll button: Manual and 1x IntegrationTest
- change of StageAll picture: Manual

## Test environment(s) 

- GIT 3.4.2.9999+
- GIT 2.30.0.windows.1
- Windows 10 20H2

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #8596, opened 2020-11-04
"Stage and commit all unstaged files" may not stage all unstaged files


manuelspezzani:
## Current behaviour

When no files are staged, a "There are no files staged for this commit. Stage and commit all unstaged files?" dialog is opened.
If the user select "yes", only the files matching the currently specified filters are actually staged.

## Expected behaviour

All files are staged, or the message is changed to better describe what is about to happen.
Personally I would prefer the first option. 

## Steps to reproduce

- Start from a repo in a "clean state" (aka no pending changes and no files staged)
- Edit two files (let's say "foo.txt" and "bar.txt") and open the commit dialog
- type "foo" in the filter textbox, on the top left corner of the commit dialog: the list of changes now shows only "foo.txt"
- enter a commit message and click on the commit button (make sure no files has been staged)
- the "There are no staged changes" dialog is shown, answer "yes"

A new commit is created containing only the changes to "foo.txt, while "bar.txt" has not been staged/committed.

## Did this work in previous version of GitExtensions

I didn't tried.

## Environment

- Git Extensions 3.4.3.9999
- Build d4b0f48bbf3e39a71f5d7ff5231be5678d4aa0f1
- Git 2.27.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3661.0
- DPI 96dpi (no scaling)

gerhardol:
I suggest changing the text:
_There are no files staged for this commit. Stage and commit all (filtered) unstaged files?_

RussKie:
May be just stage all files?

keco2:
Hi. I've analyzed this and it seems to me that the behaviour might be quasi by design (as the method is called _StageAllAccordingToFilter_ so it was the intention to stage only filtered files).

So there are 2 texts misleading:
- the **ToolTipText** on StageAll button ("Stage All" instead of e.g. "Stage Filtered" ... but then the "Filtered" would be improper if the filter was not active)
- the **MessageBox** text at a Commit ("There are no files staged for this commit. Stage and commit all unstaged files?" instead of ".... Stage and commit all (filtered) unstaged files?" -- but again, in that case the "(filtered)" would be kind of improper if the filter was not active)


My suggestion: fix both misleading texts and display either "All" or "Filtered" words based on the filter:
- if the Filter is empty, use the "All" word
- otherwise use the "Filtered" word


I've prepared this accordingly:
[...]

PR #8731, opened 2021-01-08
StageAll and UnstageAll improvement when Filter active #8596

keco2:
Followup on #8596 

Note: The first PR #8730 closed (because of I've pushed my changes on master instead of a separate branch and used merge-commit,...)

## Proposed changes

- update 2 misleading texts to display either "All" or "Filtered" words based on the filter:
  -  if the Filter is empty (i.e. inactive), use the "All" word
  - otherwise use the "Filtered" word
- change the icon on the StageAll button interactively based on the filefilter:
  -  if the Filter is empty (i.e. inactive), use the classic StageAll picture
  - otherwise use another (new) picture with funnel
- in addition to the above mentioned changes of StageAll button from #8596 there is a change related to UnstageAll button as well (as agreed in PR #8730):
  - visual change (just like on StageAll button: icon and ToolTipText changes when Filter is active)
  - behavioural change (to get it consistent with the StageAll button): if the Filter is not empty (i.e. it's active) and UnstageAll is clicked: unstage only those files matching the filter (instead of unstaging everything)

### After

**StageAll (TooltipText/icon) - with filefilter:**
Before: image
After: image

**UnstageAll (TooltipText/icon) - with filefilter:**
Before: image
After: image

**OnCommitClick (message) - with filefilter:**
Before: image
After: image

## Test methodology 

- change of text in the MessageBox on Commit: **Manual**
- change of ToolTipText on StageAll/UnstageAll buttons: **Manual and IntegrationTest**
- change of StageAll/UnstageAll button-picture: **Manual**
- change of UnstageAll behaviour: **Manual and IntegrationTest**

## Test environment(s) 

- GIT 3.4.2.9999+
- GIT 2.30.0.windows.1
- Windows 10 20H2

----


keco2:
Btw, the icon with funnel - I'm open to suggestions - if it is not that good. I've created just using VS picture editor. Maybe someone has idea how to do it better.

I've tried funnel-up vs. funnel-down:
image

Currently the funnel-down is implemented (because I think the arrows are more visible so).

pmiossec:
> Currently the funnel-down is implemented (because I think the arrows are more visible so).

I think bottom **right** would be better. What do you think?

keco2:
I've updated the picture with that option. I still like the B version a bit more :) but the C is Ok to me too. Let's wait for further votes - if no more vote comes, I'll update it with C version! :1st_place_medal:

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1345,7 +1354,9 @@ bool ConfirmAndStageAllUnstaged()
                     }
 
                     // there are no staged files, but there are unstaged files. Most probably user forgot to stage them.
-                    if (MessageBox.Show(this, _noFilesStagedButSuggestToCommitAllUnstaged.Text, _noStagedChanges.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) != DialogResult.Yes)
+                    if (MessageBox.Show(this, Unstaged.IsFilterActive
+                        ? _noFilesStagedButSuggestToCommitAllFilteredUnstaged.Text
+                        : _noFilesStagedButSuggestToCommitAllUnstaged.Text, _noStagedChanges.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) != DialogResult.Yes)
```
RussKie, line 1359:
```suggestion
                    string message = Unstaged.IsFilterActive ? _noFilesStagedButSuggestToCommitAllFilteredUnstaged.Text : _noFilesStagedButSuggestToCommitAllUnstaged.Text;
                    if (MessageBox.Show(this, message, _noStagedChanges.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) != DialogResult.Yes)
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1654,6 +1665,13 @@ void StageAreaLoaded()
                 Unstage(canUseUnstageAll: false);
             }
             else
+            if (Staged.IsFilterActive)
```
RussKie, line 1668:
```suggestion
            else if (Staged.IsFilterActive)
```

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -102,6 +104,11 @@ public sealed partial class FormCommit : GitModuleForm
         private readonly TranslationString _stageFiles = new TranslationString("Stage {0} files");
         private readonly TranslationString _selectOnlyOneFile = new TranslationString("You must have only one file selected.");
 
+        private readonly TranslationString _stageAll = new TranslationString("Stage All");
+        private readonly TranslationString _stageFiltered = new TranslationString("Stage Filtered");
+        private readonly TranslationString _unstageAll = new TranslationString("Unstage All");
+        private readonly TranslationString _unstageFiltered = new TranslationString("Unstage Filtered");
```
RussKie, line 110:
```suggestion
        private readonly TranslationString _stageAll = new TranslationString("Stage all");
        private readonly TranslationString _stageFiltered = new TranslationString("Stage filtered");
        private readonly TranslationString _unstageAll = new TranslationString("Unstage all");
        private readonly TranslationString _unstageFiltered = new TranslationString("Unstage filtered");
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1534,13 +1537,23 @@ private void FileStatusList_Enter(object sender, EventArgs e)
...
 
         public void SetFilter(string value)
         {
             FilterComboBox.Text = value;
             FilterFiles(value);
         }
 
+        private void OnFilterChanged()
+        {
+            if (_filterActive != IsFilterActive)
```
RussKie, line 1550: Please invert to reduce nesting

UnitTests/CommonTestUtils/ReferenceRepository.cs
```diff
@@ -23,6 +23,8 @@ public ReferenceRepository()
 
         public string CommitHash => _commitHash;
 
+        public string CreateRepoFile(string fileName, string fileContent) => _moduleTestHelper.CreateRepoFile(fileName, fileContent);
```
RussKie, line 26: Please move before `CreateTag`

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCommitTests.cs
```diff
@@ -207,6 +208,91 @@ public void Should_handle_well_commit_message_in_commit_message_menu()
...
+            _referenceRepository.CreateRepoFile("file2.txt", "Test");
+
+            RunFormTest(async form =>
+            {
+                using (var cts = new CancellationTokenSource(AsyncTestHelper.UnexpectedTimeout))
+                {
+                    await ThreadHelper.JoinPendingOperationsAsync(cts.Token);
+                }
+
+                Assert.AreEqual("Stage All", form.GetTestAccessor().StageAllToolItem.ToolTipText);
```
RussKie, line 226: Asserting against strings isn't a good idea, a lot of our contributors run on non-English environments, and may have different locale and localisation settings.
If you need to assert against default strings, please make sure to run tests in en-US culture:
```cs
    [SetCulture("en-US")]
    [SetUICulture("en-US")]
```

RussKie (changes requested):
👍 

Please provide your sign off as well.

GitUI/CommandsDialogs/FormCommit.cs
keco2: Thx @RussKie ! In addition I'd fix this in _FormCommit.Designer_ as well
(https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormCommit.Designer.cs) :
            this.toolStageAllItem.Text = "Stage **a**ll";
or 
            this.toolStageAllItem.Text = **_stageAll.Text**;

...or leave it unchanged and do it in the **constructor** of _FormCommit_ ?
toolStageAllItem.ToolTipText = _stageAll.Text;

Obviously, the same for _unstageFiltered.

gerhardol: > this.toolStageAllItem.Text = **_stageAll.Text**;

yes

keco2: Ouch, that'll break the VS form designer and throw an error "The designer loader did not provide a root component but has not indicated why." 

So I change instead as follows:
_FormCommit.Designer.cs:_ this.toolStageAllItem.Text = "Stage **a**ll";
_FormCommit.cs (new line in the constructor):_ toolStageAllItem.ToolTipText = _stageAll.Text;

keco2:
Great, thank you for the review! I've pushed the changes!

> Please provide your sign off as well.
Done. Separate PR.

> > Currently the funnel-down is implemented (because I think the arrows are more visible so).
> 
> I think bottom **right** would be better. 

Bottom right commited as promised ;)

UnitTests/CommonTestUtils/ReferenceRepository.cs
keco2: Makes sense, thanks! May I ask you if you use some tool to check this or just common sense? 
I've no ReSharper unfortunately but I've tried a free tool _CodeMaid_ to reorganize the file but it has found much more codeblocks to move/reorganize.

RussKie: No tool unfortunately. I strive to follow the old [.NET Framework guidelines](https://docs.microsoft.com/en-us/archive/blogs/brada/internal-coding-guidelines#28-file-organization) but unless it is enforced by tools it is an uphill battle.

GitUI/CommandsDialogs/FormCommit.cs
RussKie: You mustn't change the designer file manually, as it either likely be undone by the designer itself, or will break the designer.
The strings in the designer do not really make any difference as at runtime we'll load the translated strings anyway.

RussKie (approved):

RussKie:
@msftbot merge in 24 hours

UnitTests/CommonTestUtils/ReferenceRepository.cs
keco2: I see, thanks anyway for the link! Actually, the above mentioned VS extension seems to do a good job in reordering the code (and it's configurable) but I miss the option to sort the new code blocks only (kind of git-diff based). Hmm, could be a nice extension for VS or even GE! :thinking:

GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -967,7 +967,7 @@ private void InitializeComponent()
             this.toolStageAllItem.ImageTransparentColor = System.Drawing.Color.Magenta;
             this.toolStageAllItem.Name = "toolStageAllItem";
             this.toolStageAllItem.Size = new System.Drawing.Size(23, 23);
-            this.toolStageAllItem.Text = "Stage All";
```
gerhardol, line 970: These should be removed here and made sure they are explicitly assigned in init or load
The text is translated twice now

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1534,13 +1537,25 @@ private void FileStatusList_Enter(object sender, EventArgs e)
         private readonly Subject<string> _filterSubject = new Subject<string>();
         [CanBeNull] private Regex _filter;
         private bool _filterVisible = false;
+        private bool _filterActive = false;
```
gerhardol, line 1540: Why is this used?
OnFilterChanged() may fire the event more than once, is that so bad?
(these status variablles are hard to maintain, someon else reuses it for a slightly separate use case or add a parallel variable...)

keco2: It fires on every character added/removed. But you're right, it would be no problem here without it so I can remove that check. Or maybe just rename it to something more meaningful like _filterActiveLastState or _filterActiveCheck? (and sorry for my delayed answer)

GitUI/CommandsDialogs/FormCommit.Designer.cs
keco2: Agree. Good point. I can fix it together with the other change below.

GitUI/UserControls/FileStatusList.cs
keco2: Just to be sure - Which one do you prefer?
- remove the unnecessary check? or
- rename to something more meaningful?

gerhardol: _I_ prefer maintainability, so removing the variable if there is no noticeable effect is what I prefer.

PR #8734, opened 2021-01-09
C#9 language features

### LINKED ISSUE #8728: Use C# 7/8/9 features

gerhardol:
Part of #8728

## Proposed changes

Enable C#9 features

Applies C#9 'syntactic sugar' to simplify code and improve compiler statistic analysis.
Mostly regex replace, ~intellicode for pattern matching and a few manual changes~.
A few printouts changed as well.

Review commit by commit. (Easiest to see using automatic continuous scroll in GE, alternative use gitk).

These are default options when not added to .editorconfig, not enforced though.

## Test methodology 

review

----


GitCommands/Git/GitBranchNameNormaliser.cs
```diff
@@ -107,8 +107,7 @@ public string Normalise(string branchName, GitBranchNameOptions options)
         /// </summary>
         public static bool IsValidChar(char c)
         {
-            return (c >= 32 && c < 127) &&
-                    c != ' ' && c != '~' && c != '^' && c != ':' &&
+            return (c is > ' ' and < '~' and not '^' and not ':') &&
```
gerhardol, line 110: manually adjusted

GitUI/Editor/FileViewer.cs
```diff
@@ -768,7 +768,7 @@ void DetectDefaultEncoding()
 
         private static bool IsDiffView(ViewMode viewMode)
         {
-            return viewMode == ViewMode.Diff || viewMode == ViewMode.FixedDiff || viewMode == ViewMode.RangeDiff;
+            return viewMode is ViewMode.Diff or ViewMode.FixedDiff or ViewMode.RangeDiff;
```
gerhardol, line 771: Example where pattern matching improves readability

mstv (approved):
👍 This night I will dream of endless scrolling.

Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <Import Project=".\scripts\sdk\Sdk.props" />
 
   <PropertyGroup>
-    <LangVersion>8</LangVersion>
+    <LangVersion>9</LangVersion>
```
RussKie, line 5: We need to update the readme as well - C# 9 requires at least VS 16.7, though for .NET 5.0 it has to be 16.8.

GitCommands/Git/GitModule.cs
```diff
@@ -746,7 +746,7 @@ public async Task<List<ConflictData>> GetConflictsAsync(string filename = "")
                 string hash = findSecondWhitespace >= 0 ? fileStage.Substring(0, findSecondWhitespace).Trim() : "";
                 fileStage = findSecondWhitespace >= 0 ? fileStage.Substring(findSecondWhitespace).Trim() : "";
 
-                if (fileStage.Length > 2 && int.TryParse(fileStage[0].ToString(), out var stage) && stage >= 1 && stage <= 3)
+                if (fileStage.Length > 2 && int.TryParse(fileStage[0].ToString(), out var stage) && stage is >= 1 and <= 3)
```
RussKie, line 749: I don't find this syntax particularly better than the old one. I'm finding it requires me to pause and think about it...

ResourceManager/GitExtensionsControl.cs
```diff
@@ -123,7 +123,7 @@ public static bool IsTextEditKey(Keys keys, bool multiLine = false)
             keys &= ~Keys.Shift; // ignore the SHIFT key as modifier
             switch (keys)
             {
-                case Keys key when (key >= Keys.A && key <= Keys.Z) || (key >= Keys.D0 && key <= Keys.D9) || (key >= Keys.Oem1 && key <= Keys.Oem102):
+                case Keys key when key is >= Keys.A and <= Keys.Z or >= Keys.D0 and <= Keys.D9 or >= Keys.Oem1 and <= Keys.Oem102:
```
RussKie, line 126: I had to go and look for [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/patterns3#pattern-combinators). It is legal, but IMO this is very error prone. `and` has a higher priority than `or`, but it is very subtle, and can lead to hard to find bugs. Please retain parens to enhance the readability

RussKie (commented):
Impossible to review the change of this size. I had a cursory look throughout.
Few comments to address, then merge ASAP.

GitUI/Editor/FileViewer.cs
RussKie: At the moment I'm not finding it easier, but perhaps with time I'll get used to it.

Directory.Build.props
gerhardol: Setting version to 16.8+ in wiki and README 
(for simplicity. 16.7 is a servicing baseline though, otherwise just upgrading in itself is the problem.)

GitCommands/Git/GitModule.cs
gerhardol: I learned to like this quite fast. Initially I looked for || and && but this is shorter so it is faster to read and to understand (even better if parens were allowed).

ResourceManager/GitExtensionsControl.cs
gerhardol: Unfortunately StyleCop complains with parenthesis here...
For this specific lin

```
#pragma warning disable SA1008 // Opening parenthesis should be spaced correctly
                case Keys key when key is (>= Keys.A and <= Keys.Z) or (>= Keys.D0 and <= Keys.D9) or (>= Keys.Oem1 and <= Keys.Oem102):
#pragma warning restore SA1008 // Opening parenthesis should be spaced correctly
```

But I prefer separate lines:
```
                case Keys key when key is >= Keys.A and <= Keys.Z
                    or >= Keys.D0 and <= Keys.D9
                    or >= Keys.Oem1 and <= Keys.Oem102:
```

gerhardol:
> Few comments to address, then merge ASAP.

Updated doc only now.

For pattern matching, parenthesis would be helpful. It does not seem to be configurable in stylecop.
It seems like @sharwell has done a change here, but StyleCop 1.2 is not released yet
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3147

and it is not possible to configure the space before the paren:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0047-ide0048

@RussKie Do you request that pattern matching is reverted in genral or for the examples?
In FileViewer.cs I really refer the change though...
Will squash 8ed0913bae6e02860693d8bc1284170a9a0ba101 keep the other commits

RussKie:
No, it was a general comment regarding the language feature.

gerhardol:
> No, it was a general comment regarding the language feature.

Then I separate that commit from this PR

gerhardol:
@msftbot merge in 6 hours

PR #8737, opened 2021-01-10
C#9 pattern enhancements

### LINKED ISSUE #8728: Use C# 7/8/9 features

gerhardol:
Fixes #8728
Separated from #8734 (based on this)

## Proposed changes

Enable C#9 'syntactic sugar' for pattern matching.
Mostly changed with intellicode suggestions and a few manual changes.

Separated from #8734 to merge that quickly and discuss the changes here.
I would like to apply the changes in FileViewer.cs at least.
Maybe StyleCop changes are required toadd parenthesis to improve readability to merge this.

## Test methodology 

manual

----

### DIFF
--- a/GitCommands/Git/GitBranchNameNormaliser.cs
+++ b/GitCommands/Git/GitBranchNameNormaliser.cs
@@ -14,7 +14,7 @@ namespace GitCommands.Git
     {
         /// <summary>
         /// Ensures that the branch name meets the GIT branch naming conventions.
-        /// For more details refer to <see href="https://www.git-scm.com/docs/git-check-ref-format/1.8.2"/>.
+        /// For more details refer to <see href="https://git-scm.com/docs/git-check-ref-format"/>.
         /// </summary>
         /// <param name="branchName">Name of the branch.</param>
         /// <param name="options">The options.</param>
@@ -37,7 +37,7 @@ namespace GitCommands.Git
 
         3. They cannot have two consecutive dots '..' anywhere.
 
-        4. They cannot have ASCII control characters (i.e. bytes whose values are lower than \040, or \177 DEL), space, tilde '~', caret '^', or colon ':' anywhere.
+        4. They cannot have ASCII control characters (i.e. bytes whose values are lower than \040, or \177 DEL), space ' ', tilde '~', caret '^', or colon ':' anywhere.
 
         5. They cannot have question-mark '?', asterisk '*', or open bracket '[' anywhere. See the --refspec-pattern option below for an exception to this rule.
 
@@ -47,7 +47,9 @@ namespace GitCommands.Git
 
         8. They cannot contain a sequence '@{'.
 
-        9. They cannot contain a '\'.
+        9. They cannot be the single character '@'.
+
+        10. They cannot contain a '\'.
 
         These rules make it easy for shell script based tools to parse reference names, pathname expansion by the shell when a reference name is used unquoted (by mistake), and also avoids ambiguities in certain reference name expressions (see gitrevisions(7)):
 
@@ -88,6 +90,7 @@ namespace GitCommands.Git
             }
 
             // run rules in reverse order
+            branchName = Rule10(branchName, options);
             branchName = Rule09(branchName, options);
             branchName = Rule08(branchName, options);
             branchName = Rule07(branchName, options);
@@ -108,8 +111,7 @@ namespace GitCommands.Git
         /// </summary>
         public static bool IsValidChar(char c)
         {
-            return (c >= 32 && c < 127) &&
-                    c != ' ' && c != '~' && c != '^' && c != ':' &&
+            return c is (> ' ' and < '~' and not ('^' or ':')) &&
                     Array.IndexOf(Path.GetInvalidPathChars(), c) < 0;
         }
 
@@ -229,12 +231,23 @@ namespace GitCommands.Git
         }
 
         /// <summary>
-        /// Branch name cannot contain a '\'.
+        /// Branch name cannot be the single character '@'.
         /// </summary>
         /// <param name="branchName">Name of the branch.</param>
         /// <param name="options">The options.</param>
         /// <returns>Normalised branch name.</returns>
         internal string Rule09(string branchName, GitBranchNameOptions options)
+        {
+            return branchName != "@" ? branchName : options.ReplacementToken;
+        }
+
+        /// <summary>
+        /// Branch name cannot contain a '\'.
+        /// </summary>
+        /// <param name="branchName">Name of the branch.</param>
+        /// <param name="options">The options.</param>
+        /// <returns>Normalised branch name.</returns>
+        internal string Rule10(string branchName, GitBranchNameOptions options)
         {
             return Regex.Replace(branchName, @"(\\{1,})", options.ReplacementToken);
         }
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1459,7 +1459,7 @@ namespace GitUI
                 return false;
             }
 
-            if ((command == BlameHistoryCommand || command == FileHistoryCommand) && args.Count <= 2)
+            if (command is (BlameHistoryCommand or FileHistoryCommand) && args.Count <= 2)
             {
                 MessageBox.Show("Cannot open blame / file history, there is no file selected.", "Blame / file history", MessageBoxButtons.OK, MessageBoxIcon.Error);
                 return false;
--- a/GitUI/Script/SplitButton.cs
+++ b/GitUI/Script/SplitButton.cs
@@ -321,7 +321,7 @@ namespace GitUI.Script
                               bounds.Width - _dropDownRectangle.Width - internalBorder,
                               bounds.Height - (internalBorder * 2) + 2);
 
-            bool drawSplitLine = State == PushButtonState.Hot || State == PushButtonState.Pressed || !Application.RenderWithVisualStyles;
+            bool drawSplitLine = State is (PushButtonState.Hot or PushButtonState.Pressed) || !Application.RenderWithVisualStyles;
 
             if (RightToLeft == RightToLeft.Yes)
             {
@@ -619,7 +619,7 @@ namespace GitUI.Script
             {
                 offset = excess_width;
             }
-            else if (h_image == HorizontalAlignment.Center && (h_text == HorizontalAlignment.Left || h_text == HorizontalAlignment.Center))
+            else if (h_image == HorizontalAlignment.Center && h_text is (HorizontalAlignment.Left or HorizontalAlignment.Center))
             {
                 offset += excess_width / 3;
             }
@@ -682,7 +682,7 @@ namespace GitUI.Script
             {
                 offset = excess_height;
             }
-            else if (v_image == VerticalAlignment.Center && (v_text == VerticalAlignment.Top || v_text == VerticalAlignment.Center))
+            else if (v_image == VerticalAlignment.Center && v_text is (VerticalAlignment.Top or VerticalAlignment.Center))
             {
                 offset += excess_height / 3;
             }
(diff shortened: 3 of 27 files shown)


RussKie (commented):
I'm not going to block this work, but so far this syntax makes me think more. Perhaps with time it'll get easier.

pmiossec:
No problem for me to read it. I like it.
But I think it will take a lot of time to think about using it...

gerhardol:
Rebased to resolve conflicts
An alternative supported by StyleCop is the following. 

Before (slightly tweaked)
```C#
        public static bool IsValidChar(char c)
        {
            return c > ' ' && c < '~' && c != '^' && c != ':' &&
                    Array.IndexOf(Path.GetInvalidPathChars(), c) < 0;
        }

```

Currently in the PR
```C#
        public static bool IsValidChar(char c)
        {
            return (c is > ' ' and < '~' and not '^' and not ':') &&
                    Array.IndexOf(Path.GetInvalidPathChars(), c) < 0;
        }
```

Alternative (must be is())
```C#
        public static bool IsValidChar(char c)
        {
            return c is(> ' ' and < '~' and not '^' and not ':') &&
                    Array.IndexOf(Path.GetInvalidPathChars(), c) < 0;
        }
```

ResourceManager/GitExtensionsControl.cs
```diff
@@ -123,7 +123,7 @@ public static bool IsTextEditKey(Keys keys, bool multiLine = false)
             keys &= ~Keys.Shift; // ignore the SHIFT key as modifier
             switch (keys)
             {
-                case Keys key when (key >= Keys.A && key <= Keys.Z) || (key >= Keys.D0 && key <= Keys.D9) || (key >= Keys.Oem1 && key <= Keys.Oem102):
+                case Keys key when key is >= Keys.A and <= Keys.Z or >= Keys.D0 and <= Keys.D9 or >= Keys.Oem1 and <= Keys.Oem102:
```
RussKie, line 126: How about this?
```diff
diff --git a/ResourceManager/GitExtensionsControl.cs b/ResourceManager/GitExtensionsControl.cs
index 2c23fc061..f2301ec8c 100644
--- a/ResourceManager/GitExtensionsControl.cs
+++ b/ResourceManager/GitExtensionsControl.cs
@@ -123,7 +123,10 @@ public static bool IsTextEditKey(Keys keys, bool multiLine = false)
             keys &= ~Keys.Shift; // ignore the SHIFT key as modifier
             switch (keys)
             {
-                case Keys key when key is >= Keys.A and <= Keys.Z or >= Keys.D0 and <= Keys.D9 or >= Keys.Oem1 and <= Keys.Oem102:
+                case Keys key when key is
+                    (>= Keys.A and <= Keys.Z) or
+                    (>= Keys.D0 and <= Keys.D9) or
+                    (>= Keys.Oem1 and <= Keys.Oem102):
                 case Keys.Space:
                 case Keys.Back:
                 case Keys.Delete:
```

GitUI/Script/SplitButton.cs
```diff
@@ -321,7 +321,7 @@ protected override void OnPaint(PaintEventArgs pevent)
                               bounds.Width - _dropDownRectangle.Width - internalBorder,
                               bounds.Height - (internalBorder * 2) + 2);
 
-            bool drawSplitLine = State == PushButtonState.Hot || State == PushButtonState.Pressed || !Application.RenderWithVisualStyles;
+            bool drawSplitLine = State is PushButtonState.Hot or PushButtonState.Pressed || !Application.RenderWithVisualStyles;
```
RussKie, line 324:
```suggestion
            bool drawSplitLine = !Application.RenderWithVisualStyles || (State is PushButtonState.Hot or PushButtonState.Pressed);
```

GitUI/GitUICommands.cs
```diff
@@ -1431,7 +1431,7 @@ public bool RunCommand(IReadOnlyList<string> args)
                 return false;
             }
 
-            if ((command == BlameHistoryCommand || command == FileHistoryCommand) && args.Count <= 2)
+            if ((command is BlameHistoryCommand or FileHistoryCommand) && args.Count <= 2)
```
RussKie, line 1434:
```suggestion
            if (args.Count <= 2 && (command is BlameHistoryCommand or FileHistoryCommand))
```

RussKie (commented):
I'm not a big fan, but I can live with this

GitCommands/Git/GitBranchNameNormaliser.cs
```diff
@@ -107,8 +107,7 @@ public string Normalise(string branchName, GitBranchNameOptions options)
         /// </summary>
         public static bool IsValidChar(char c)
         {
-            return (c >= 32 && c < 127) &&
-                    c != ' ' && c != '~' && c != '^' && c != ':' &&
+            return (c is > ' ' and < '~' and not '^' and not ':') &&
```
sharwell, line 110: 💭 No way I would have known that `'~'` is 0x7E. We should document why this range has been chosen.

GitUI/GitUICommands.cs
sharwell: I prefer the form already in code. The condition which restricts the argument count is the command, so it should be checked first unless fast-path performance dictates otherwise (which is not a problem in this case).

GitUI/Script/SplitButton.cs
sharwell: Intensely agree with the suggestion here. Parentheses are essential.

Plugins/Statistics/GitStatistics/PieChart/EdgeColor.cs
```diff
@@ -51,7 +51,7 @@ public static Color GetRenderingColor(EdgeColorType edgeColorType, Color color)
 
         private static EdgeColorType GetContrastColorType(Color color, EdgeColorType colorType)
         {
-            Debug.Assert(colorType == EdgeColorType.Contrast || colorType == EdgeColorType.EnhancedContrast, "colorType == EdgeColorType.Contrast || colorType == EdgeColorType.EnhancedContrast");
+            Debug.Assert(colorType is EdgeColorType.Contrast or EdgeColorType.EnhancedContrast, "colorType == EdgeColorType.Contrast || colorType == EdgeColorType.EnhancedContrast");
```
sharwell, line 54:
```suggestion
            Debug.Assert(colorType is EdgeColorType.Contrast or EdgeColorType.EnhancedContrast, "colorType is EdgeColorType.Contrast or EdgeColorType.EnhancedContrast");
```

Plugins/Statistics/GitStatistics/PieChart/PieChart3D.cs
```diff
@@ -851,7 +851,7 @@ protected virtual void InitializePieSlices()
         /// </returns>
         protected SizeF GetSliceDisplacement(float angle, float displacementFactor)
         {
-            Debug.Assert(displacementFactor > 0F && displacementFactor <= 1F, "displacementFactor > 0F && displacementFactor <= 1F");
+            Debug.Assert(displacementFactor is > 0F and <= 1F, "displacementFactor > 0F && displacementFactor <= 1F");
```
sharwell, line 854:
```suggestion
            Debug.Assert(displacementFactor is > 0F and <= 1F, "displacementFactor is > 0F and <= 1F");
```

sharwell, line 854: 💡 Consider using `IsDisplacementValid` instead of duplicating the check.

```suggestion
            Debug.Assert(IsDisplacementValid(displacementFactor), "IsDisplacementValid(displacementFactor)");
```

ResourceManager/GitExtensionsControl.cs
sharwell: Considering what the code on the left is, I'm fine with keeping one line as long as the parentheses are returned. If wrapping is applied, I would suggest placing the `or` operators at the beginning of lines instead of the end.

sharwell (commented):
Need documentation for one of the changes, and parentheses added for cases where `and` and `or` are both used, and cases where `and`/`or` are used with `&&`/`||`.

ResourceManager/GitExtensionsControl.cs
RussKie: > suggest placing the `or` operators at the beginning of lines instead of the end.

Works for me

Plugins/Statistics/GitStatistics/PieChart/PieChart3D.cs
gerhardol: IsDisplacementValid() is not the same but ==0F is handled anyway so I will change

gerhardol: ok

GitCommands/Git/GitBranchNameNormaliser.cs
gerhardol: This is documented in the top of the file. I updated the link and compared to current (adding one rule).

GitUI/Script/SplitButton.cs
gerhardol: Added parenthesis in general. Did not change the order here (kept as below)

```C#
bool drawSplitLine = State is(PushButtonState.Hot or PushButtonState.Pressed) || !Application.RenderWithVisualStyles;
```

gerhardol:
Added parenthesis to all changes (as it was only statements with && or || that was changed).
I was not completely sure what paren style that was suggested, I added my last suggestion, also suggested here:
https://github.com/gitextensions/gitextensions/pull/8748#discussion_r556874914
That can be changed in this PR (and StyleCop should allow a space after is before paren).

RussKie:
@sharwell do you think it is feasible to enhance [SA1008](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1008.md) to play nicely with the new syntax?
Do you know if this issue is tracked anywhere, or do I need to raise a new issue?

sharwell:
Can you file a new issue for it?

RussKie:
@gerhardol can you try updating the package to the latest beta https://www.nuget.org/packages/StyleCop.Analyzers?

gerhardol:
Updated to latest beta package of StyleCop. We were already using a beta version, so not a big difference
Also rewrote a few statements
One new check had to be addressed too

Separate commits, for reviews

mstv (approved):

RussKie (approved):

RussKie:
```
1) Error : GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder
  System.AggregateException : One or more errors occurred.
    ----> System.ArgumentException : Unable to sort because the IComparer.Compare() method returns inconsistent results. Either a value does not compare equal to itself, or one value repeatedly compared to another value yields different results. IComparer: 'System.Array+FunctorComparer`1[GitUI.UserControls.RevisionGrid.Graph.RevisionGraphRevision]'.
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout)
     at System.Threading.Tasks.Task.WaitAll(Task[] tasks)
     at GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder() in C:\projects\gitextensions\UnitTests\GitUI.Tests\UserControls\RevisionGrid\Graph\RevisionGraphMultiThreadingTests.cs:line 62
  --ArgumentException
     at System.Collections.Generic.IntrospectiveSortUtilities.ThrowOrIgnoreBadComparer(Object comparer)
     at System.Collections.Generic.ArraySortHelper`1.Sort(T[] keys, Int32 index, Int32 length, IComparer`1 comparer)
     at System.Array.Sort[T](T[] array, Int32 index, Int32 length, IComparer`1 comparer)
     at System.Collections.Generic.List`1.Sort(Comparison`1 comparison)
     at GitUI.UserControls.RevisionGrid.Graph.RevisionGraph.BuildOrderedNodesCache(Int32 currentRowIndex) in C:\projects\gitextensions\GitUI\UserControls\RevisionGrid\Graph\RevisionGraph.cs:line 336
     at GitUI.UserControls.RevisionGrid.Graph.RevisionGraph.GetNodeForRow(Int32 row) in C:\projects\gitextensions\GitUI\UserControls\RevisionGrid\Graph\RevisionGraph.cs:line 130
     at GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.Render() in C:\projects\gitextensions\UnitTests\GitUI.Tests\UserControls\RevisionGrid\Graph\RevisionGraphMultiThreadingTests.cs:line 113
     at GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.<ShouldReorderInTopoOrder>b__5_2() in C:\projects\gitextensions\UnitTests\GitUI.Tests\UserControls\RevisionGrid\Graph\RevisionGraphMultiThreadingTests.cs:line 56
     at System.Threading.Tasks.Task.InnerInvoke()
     at System.Threading.Tasks.Task.Execute()
```

gerhardol:
> 1) Error : GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder

Seem to be a flaky test. I do not see any of my changes related to this and it runs OK locally.

gerhardol:
@msftbot merge in 24 hours

PR #8726, opened 2021-01-06
Sidepanel submodule improvements

gerhardol:
Closes #8714 
Second part of #8703
Based on #8703 that is set to automerge so not to be merged yet. Fully reviewable.

Best reviewed commit by commit

## Proposed changes

* Open submodule in left panel selects first/selected commit as in #8703
* Add actions in Commit/RevDiff worktree to submodules: Reset/Stash/Commit
* Add Open in new instance for submodules
* Tooltip with details for changed submodules. This adds extra git-diff for modules that are dirty only (for others the command were executed anyway). The number of changes reported are limited to five changes (similar to how the RevGrid tooltip is limited already)
* For topmodule there is no submodule status message available, display the same tooltip as for the RevGrid WorkTree commit.
* A few minor refactorings were required

Review commit by commit.
Several of the changes will be squashed at approval.

## Screenshots

Using a testrepo with GE modules recursively added (originally to test performance)

Before | After
------ | -----
image | image
image | image
(all submodules do not have tooltips) | image
Commits selected when opening a submodule from left panel
image | image

## Test methodology 

Manual
Mostly UI changes, using already existing 

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8738, opened 2021-01-10
Use Executable for launching new GE instances

mstv:
Contributes to #7795
Fixes https://github.com/gitextensions/gitextensions/pull/8726#discussion_r554511688

## Proposed changes

- Factor out `GitUICommands.Spawn` and `GitUICommands.SpawnBrowse` using `Executable` in order to throw `ExternalOperationException` on failure.

## Screenshots 

N/A

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build a5acbea9e7a67ce50156e1e4ff38f6b5d002e6e8
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1,9 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Git;
@@ -355,6 +355,44 @@ namespace GitUI
         }
 
         #endregion
+
+        /// <summary>
+        /// Launches a new GE instance.
+        /// </summary>
+        /// <param name="arguments">The command line arguments.</param>
+        /// <param name="workingDir">The working directory for the new process.</param>
+        /// <returns>The <see cref="IProcess"/> object for controlling the launched instance.</returns>
+        public static IProcess Launch(string arguments, string workingDir = "")
+            => new Executable(Application.ExecutablePath, workingDir).Start(arguments);
+
+        /// <summary>
+        /// Launch FormBrowse in a new GE instance.
+        /// </summary>
+        /// <param name="workingDir">The working directory for the new process.</param>
+        /// <param name="selectedId">The optional commit to be selected.</param>
+        /// <param name="firstId">The first commit to be selected, the first commit in a diff.</param>
+        /// <returns>The <see cref="IProcess"/> object for controlling the launched instance.</returns>
+        public static IProcess LaunchBrowse(string workingDir = "", ObjectId selectedId = null, ObjectId firstId = null)
+        {
+            StringBuilder arguments = new StringBuilder("browse");
+
+            if (selectedId is null)
+            {
+                selectedId = firstId;
+                firstId = null;
+            }
+
+            if (selectedId is not null)
+            {
+                arguments.Append(" -commit=").Append(selectedId);
+                if (firstId is not null)
+                {
+                    arguments.Append(',').Append(firstId);
+                }
+            }
+
+            return Launch(arguments.ToString(), workingDir);
+        }
 
         public bool StartCompareRevisionsDialog(IWin32Window owner = null)
         {
@@ -1186,17 +1224,7 @@ namespace GitUI
         public void StartFileHistoryDialog(IWin32Window owner, string fileName, GitRevision revision = null, bool filterByRevision = false, bool showBlame = false)
         {
             string arguments = $"{(showBlame ? BlameHistoryCommand : FileHistoryCommand)} {fileName.Quote()} {revision?.ObjectId} {(filterByRevision ? FilterByRevisionArg : string.Empty)}";
-            var process = new Process
-            {
-                StartInfo =
-                {
-                    FileName = Application.ExecutablePath,
-                    Arguments = arguments,
-                    WorkingDirectory = Module.WorkingDir
-                }
-            };
-
-            process.Start();
+            Launch(arguments, Module.WorkingDir);
         }
 
         public void OpenWithDifftool(IWin32Window owner, IReadOnlyList<GitRevision> revisions, string fileName, string oldFileName, RevisionDiffKind diffKind, bool isTracked, string customTool = null)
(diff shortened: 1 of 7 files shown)


GitUI/GitUICommands.cs
```diff
@@ -356,6 +356,44 @@ public bool StartCheckoutRemoteBranch(IWin32Window owner, string branch)
...
+        /// <summary>
+        /// Spawns a new GE instance.
+        /// </summary>
+        /// <param name="workingDir">The working directory for the new process.</param>
+        /// <param name="selectedId">The optional commit to be selected.</param>
+        /// <param name="firstId">The first commit to be selected, the first commit in a diff.</param>
+        /// <returns>The <see cref="IProcess"/> object for controlling the spawned instance.</returns>
+        public static IProcess SpawnBrowse(string workingDir = "", ObjectId selectedId = null, ObjectId firstId = null)
+        {
+            StringBuilder arguments = new StringBuilder("browse");
```
drewnoakes, line 377: Consider using [`ArgumentBuilder`](https://github.com/gitextensions/gitextensions/blob/master/GitExtUtils/ArgumentBuilder.cs).

RussKie (approved):
LGT, though I don't like "spawn" (it makes me feel uneasy), and prefer "launch" instead.

GitUI/GitUICommands.cs
RussKie: Yes

mstv: `ArgumentBuilder` inserts unwanted spaces

```
            var arguments = new ArgumentBuilder
            {
                "browse",
                { selectedId is not null, $"-commit={selectedId}" },
                { firstId is not null, $",{firstId}" }
            };
```

mstv:
> though I don't like "spawn" (it makes me feel uneasy), and prefer "launch" instead.

https://en.wikipedia.org/wiki/Spawn_(computing)
"launch" is a little more generic. I do not mind renaming it.

GitUI/GitUICommands.cs
gerhardol: The argument to -commit could be quoted or we add another argument `--firstCommit` too
(I prefer it like it is)
Should be written as:
```C#
var arguments = new ArgumentBuilder
            {
                "browse",
                { string.IsNullOrEmpty(selectedId), $"-commit={selectedId}" },
                { string.IsNullOrEmpty(selectedId) && string.IsNullOrEmpty(firstId), $",{firstId}" }
            };
```

GitUI/UserControls/FileStatusList.cs
```diff
@@ -800,24 +800,12 @@ public async Task OpenSubmoduleAsync()
...
 
             var status = await SelectedItem.Item.GetSubmoduleStatusAsync().ConfigureAwait(false);
-            var selected = SelectedItem.SecondRevision.ObjectId == ObjectId.WorkTreeId
-                ? SelectedItem.SecondRevision.ObjectId.ToString()
-                : status?.Commit.ToString() ?? string.Empty;
-            var first = string.IsNullOrWhiteSpace(status?.OldCommit?.ToString())
-                ? string.Empty
-                : $",{status.OldCommit}";
+            ObjectId selectedId = SelectedItem.SecondRevision?.ObjectId == ObjectId.WorkTreeId
+                ? SelectedItem.SecondRevision.ObjectId
```
gerhardol, line 804:
```suggestion
                ? ObjectId.WorkTreeId
```
(yes this is following what I wrote,...)

gerhardol (approved):

GitUI/GitUICommands.cs
mstv: Can `ObjectId.ToString()` return an empty string for a non-null instance? 

RE first variant:
- It will insert a space in front of the comma.
- No need to check both `selectedId` and `firstId`. This handled in the code above. They are of type `ObjectId` not `string` here.

RE second variant:
- Neither the nested `$"{$"{}"}"` nor `$"{cond, optionalValue}"` do compile.
- Adding another instance of `ArgumentBuilder` will complicate the code more than the use of `StringBuilder`.

GitUI/UserControls/FileStatusList.cs
mstv: done

GitUI/GitUICommands.cs
gerhardol: > Can `ObjectId.ToString()` return an empty string for a non-null instance?

Not in the current implementation, so it is safe in the reality.
It is hard to explicitly assign null too, you have to try really hard: TryParse("xxx", out id)
So "is null" sufficient.

But the problem with the examples was that firstCommit must nut be included if secondCommit is not set
(The alternatives were not meant to be complete)
The PR is correct and what I prefer

RussKie: > `ArgumentBuilder` inserts unwanted spaces
> 
> ```
>             var arguments = new ArgumentBuilder
>             {
>                 "browse",
>                 { selectedId is not null, $"-commit={selectedId}" },
>                 { firstId is not null, $",{firstId}" }
>             };
> ```

Does it matter if the command is like `browse -commit=234235,`?

```cs
            var arguments = new ArgumentBuilder
            {
                "browse",
                { selectedId is not null, $"-commit={selectedId},{firstId}" }
            };
```

GitUI/GitUICommands.cs
```diff
@@ -356,6 +356,44 @@ public bool StartCheckoutRemoteBranch(IWin32Window owner, string branch)
 
         #endregion
 
+        /// <summary>
+        /// Launches a new GE instance.
+        /// </summary>
+        /// <param name="arguments">The command line arguments.</param>
+        /// <param name="workingDir">The working directory for the new process.</param>
+        /// <returns>The <see cref="IProcess"/> object for controlling the launched instance.</returns>
+        public static IProcess Launch(string arguments, string workingDir = "")
```
RussKie, line 365: A better place for this general purpose functionality is probably in GitExtUtils, so plugins can leverage it as well.
But this would be a separate piece of work.

RussKie (approved):

GitUI/GitUICommands.cs
gerhardol: > Does it matter if the command is like browse -commit=234235,?

I believe the parser must be tweaked

mstv:
@msftbot merge in 1 day

PR #8740, opened 2021-01-10
Set one style (icon height and size) for Left Pane tree row and Grid row height

ivangrek:
## Proposed changes

- Row height 24px: 16px icon + 2 * 4px paddings

### Before

image

### After

image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build e361fdeae9305b54fe1458345c49b339d9629586
- Git 2.28.0.windows.1
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Designer.cs
@@ -133,6 +133,7 @@ namespace GitUI.BranchTreePanel
             this.treeMain.ShowNodeToolTips = true;
             this.treeMain.Size = new System.Drawing.Size(300, 324);
             this.treeMain.TabIndex = 3;
+            this.treeMain.ShowLines = false;
             // 
             // menuMain
             // 
--- a/GitUI/BranchTreePanel/RepoObjectsTree.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.cs
@@ -71,6 +71,7 @@ namespace GitUI.BranchTreePanel
             mnuBtnPruneAllRemotes.AdaptImageLightness();
             mnubtnFetchCreateBranch.AdaptImageLightness();
             mnubtnPullFromRemoteBranch.AdaptImageLightness();
+            treeMain.ItemHeight = DpiUtil.Scale(24);
             InitializeComplete();
 
             RegisterContextActions();
--- a/GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
@@ -46,7 +46,7 @@ namespace GitUI.UserControls.RevisionGrid.Columns
 
             Column.Width = e.CellBounds.Height;
 
-            var padding = DpiUtil.Scale(2);
+            var padding = DpiUtil.Scale(4);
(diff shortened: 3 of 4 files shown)


mstv:
Why do you want to display less?

(I have no strong opinion regarding the highlighting of the full row, since the full row was clickable before.)

ivangrek:
> Why do you want to display less?

No less, no more, but enough. 😉

mstv:
The number of branches visible at a time has decreased. (Of course, one can scroll to see them all...)

ivangrek:
Yes. Decreased slightly. But perception and search have improved.
Here one can rely only on feelings, since it is not always possible to unequivocally explain why this is so.

ivangrek:
@mstv, do you know how to disable these colors? As far as they do not carry any information for me and complicate perception.

image

mstv:
grafik

pmiossec:
For the yellow, that's in the settings, you have a color for same author commits.

I don't remember if there is a setting to disable it.

RussKie:
> @mstv, do you know how to disable these colors? As far as they do not carry any information for me and complicate perception.

They do for me. Without the alternating colours the graph becomes hard to navigate:

| Alternating colours on | Alternating colours off |
| --- | --- |
| image | image |

RussKie:
>  After
> image

I don't mind the look, but I'd probably the vertical tree lines to remain.  Can you share a screenshot of how it would look like with them?

RussKie:
Have you checked for visual regressions at 150-200% scale factor?

ivangrek:
> I don't mind the look, but I'd probably want the vertical tree lines to remain. Can you share a screenshot of how it would look like with them?

Not compatible.

> Have you checked for visual regressions at 150-200% scale factor?

Add using DpiUtil.Scale.

RussKie:
> > I don't mind the look, but I'd probably want the vertical tree lines to remain. Can you share a screenshot of how it would look like with them?
> 
> Not compatible.

Why?
image

I suppose I can live with the gridlines = off.

ivangrek:
> Why?

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.treeview.fullrowselect?view=net-5.0#remarks

pmiossec:
> I suppose I can live with the gridlines = off.

I'm not sure I will like it...

I agree that the indentation is quite important so the need of grid is not absolutely needed in the current UI.
But what we are doing is, for me:
- increase the padding, so for a long list, the gridline will be more needed...
- choose to improve selected line readability over tree readability even if most of the time, selected line doesn't correspond to something interesting so is not important. And if you loose your selection, you just have to click on a node. I was never bothered by the lost of selection...

pmiossec:
To be clearer, I really like the full line selection but I will prefer to keep the gridlines...

pmiossec:
For example, I hate how they remove the gridlines in the file explorer with a very small indentation and I spend times looking carefully at which parent a folder belong!

ivangrek:
> I'm not sure I will like it...

Exactly. The lines create the illusion of control. But this is just an illusion.

mstv:
> The lines create the illusion of control. But this is just an illusion.

Convince with arguments, please.

gerhardol:
I prefer the compact view, if to align, change the grid...
I prefer the gridlines but that is lower priority for me.

PR #8748, opened 2021-01-13
fix: Select diff when there are staged and unstaged files

RussKie:
If there are both stages and unstaged files the commit dialog fails to select the first unstaged file, thus showing no diff. If there are only unstaged or staged files everything works as expected.
When the form opens it expects "Commit" button to have focus (i.e. be an active control), however it is not happening, and the active control remains null.

This appears as a regression form 3.4.3, but I was unable to see any changes that could lead to this bug. I haven't bisected due to limited available time.

## Proposed changes

Treat "no active control" the same way as "active control == Commit" to force the selection.

### Before

image
image

### After

image

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1178,7 +1178,7 @@ namespace GitUI.CommandsDialogs
             {
                 var fc = this.FindFocusedControl();
 
-                if (fc == Ok)
+                if (fc is null || fc == Ok)
                 {
                     if (Unstaged.GitItemStatuses.Any())
                     {


GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1178,7 +1178,7 @@ private void LoadUnstagedOutput(IReadOnlyList<GitItemStatus> allChangedFiles)
             {
                 var fc = this.FindFocusedControl();
 
-                if (fc == Ok)
+                if (fc is null || fc == Ok)
```
ivangrek, line 1181: If C# 9 will be prepared.

`if (fc is null or Ok)`

gerhardol: C#9 can be used
This specific line kind of depends on #8737 (that applies pattern enhancements to many existing snippets)
@RussKie was negative to the change, voting is needed
I prefer it (submitted the PR...) but would still have preferred another parenthesis:
```C#
if (fc is (null or Ok))
```

mstv: I like the fact that the variable does not need to be repeated anymore.
`if (fc is (null or Ok))` would be most clear for me.
`if (fc is(null or Ok))` would be a compromise for StyleCop and can be easily replaced after an StyleCop update.

mstv (approved):
I cannot reproduce the issue. So it might depend on the timing.
The change will be OK anyway.

GitUI/CommandsDialogs/FormCommit.cs
gerhardol: > I like the fact that the variable does not need to be repeated anymore.

That is my primary reason as well - normally shorter too.

This is what I asked in #8737 - review that and convince @RussKie. He has the majority vote but would not block a common agreement.

RussKie: > was negative to the change, voting is needed

I'm finding it somewhat harder to read, but I won't oppose the change, if we can enforce parens to state the intent clearly.

PR #8751, opened 2021-01-14
Show both bisect and git operation notifications

RussKie:
## Proposed changes

 It is possible for a repo to be in a middle of a bisect operation and be in a conflicted state. Hence detect bisect separately from the rest of git actions.

### Before

image

### After

image

## Test methodology 

- manual

----


mstv (approved):
LGTM, have not run it but this cannot be wrong

RussKie:
@msftbot merge in 24 hours

PR #8761, opened 2021-01-17
Change TextImageRelation For "Reset All" Button

### LINKED ISSUE #6548: Some texts in german are too big for some buttons
## Current behaviour

Some text fields are too big, see screenshots

I also collected some other ui things...

## Expected behaviour

you know what I mean :)

## Screenshots

In commit window:

image

missing translation:
image

In Setting text not in the middle:

image

A little bit more space would be nice here in stash window:

image

Clean Working directory, missing translations:
image

Submodules the space is very low

image

and if I shrink the window

image

manage remotes

image

manage remotes other tag

image

Open Repository:

image

## Environment

- Git Extensions 3.1.0.5766-RC1
- Build 497059b2205086d8a44a33dd29f2cae2b4bfffc2 (Dirty)
- Git 2.19.0.windows.1
- .NET Framework 4.7.3394.0
- DPI 96dpi (no scaling)

thimmy687:
Change TextImageRelation

For some languages e.g. german the button is to small and
the icon was cutted and moved a little bit down.
For this the TextImageRelation was changed to overlay to get
the same behavior as the other buttons.

Fixes #6548 (partially)

## Proposed changes

- change the TextImageRelation to avoid the cutting of the icon in front of the button

### Before

image

### After

image

## Test methodology 

- Manual test

## Test environment(s) 

- Windows 10

----


mstv (commented):
Could you manually apply the _essential_ changes to the designer file, please?
It seems that the `modifyCommitMessageButton` or its `Text` got lost somehow.
Feel free to force-push to this branch.

RussKie (changes requested):
The diff contains too many unrelated changes that makes it impossible to reason about the proposed change.

Please remove all irrelevant changes and keep the actual change.

thimmy687:
the designer is doing so much things in the background :/

But I removed the spam now

pmiossec:
> the designer is doing so much things in the background :/

Even when running VisualStudio in no scale mode? https://github.com/gitextensions/gitextensions/wiki/Guidelines:-UI-design

Side note for dev team: I don't know how to make contributors aware that it is really important to update forms in no scale mode before they start to update the forms....

pmiossec:
> But I removed the spam now

Sorry to have to tell you but there are still a lot of whitespaces added after `//` comments that need to be removed 😞

RussKie:
I've run it locally, and I don't think there is a little more to be done, the text still looks squashed:
image

The layouts of the panel and buttons need to be changed to be fluent and autosize based on content.

thimmy687:
> I've run it locally, and I think there is a little more to be done, the text still looks squashed:
> image
> 
> The layouts of the panel and buttons need to be changed to be fluent and autosize based on content.

I think in the released version other text strings are in place, for this is it fitting

image

RussKie:
> I think in the released version other text strings are in place, for this is it fitting
> 
> image

No, it isn't. You can see the icon on the 4th button is lowered, meaning the text is wrapped.

thimmy687:
> > I think in the released version other text strings are in place, for this is it fitting
> > image
> 
> No, it isn't. You can see the icon on the 4th button is lowered, meaning the text is wrapped.

The wrapping is no longer the case, because I changed the TextImageRelation to default (overlay)

can you tell me what will be the German text in the next release?

If it is "Ungestagte Änderungen" everything is fine, if it is " "Ungestagte Änderungen zurücksetzen" we will need much more space for this button, in this case I would add a additional change.

gerhardol:
> But I guess in the 3rd button the word "zurücksetzen" will be removed, like in version 3.4.3 and than it looks good

The current translation "Reset Unstaged Changes" is "Alle ungestagten Änderungen zurücknehmen" (and for "Reset All Changes" "Alle Änderungen zurücknehmen". You can find this in the .xlf files.
This is updated by translators on Transifex.
The translation soures for next version (3.5?) is not pushed yet to Transifex, usually done some weeks before the release.

RussKie:
> The translation soures for next version (3.5?) is not pushed yet to Transifex, usually done some weeks before the release.

The translations are pushed to transifex constantly, it just we don't pull updated translations that often.

> The current translation "Reset Unstaged Changes" is "Alle ungestagten Änderungen zurücknehmen" (and for "Reset All Changes" "Alle Änderungen zurücknehmen".

Wow, that's way too long to comfortably present on the button without making the button take half the screen. 🤔

RussKie:
https://github.com/gitextensions/gitextensions/pull/8791 brings the latest translations

RussKie:
> But I guess in the 3rd button the word "zurücksetzen" will be removed, like in version 3.4.3 and than it looks good

@mstv what do you think? 

@thimmy687 we don't generally control translations, those are done by the community at [Transifex](https://www.transifex.com/git-extensions/git-extensions/). You are welcome to fix it, though there's no guarantee someone won't attempt to change it later.

mstv:
> But I guess in the 3rd button the word "zurücksetzen" will be removed, like in version 3.4.3 and than it looks good

It is a valid option to remove the verb, although this will only be understandable in combination with the icon.
I agree that "zurücknehmen" would have been the better translation.
Perhaps we should add a tooltip for translations which must be truncated.

**But** *I* strongly prefer all development tools in English.
Amongst others, Microsoft has often chosen the wrong translation, which then must be translated back in order to understand the real meaning. There are several permanent irritants.
And there is no real translation for e.g. "unstaged", just a German declension of the English word: *un*ge*stag*t.
Of course, just my 2 ct.

thimmy687:
but the TextImageRelation change is still valid ... this avoid the cutting/wrapping of the icon in the button

So for me we could merge this still

RussKie:
> but the TextImageRelation change is still valid ... this avoid the cutting/wrapping of the icon in the button
> 
> So for me we could merge this still

No, the change is not correct because the last button isn't showing the full text:
https://github.com/gitextensions/gitextensions/blob/fde9a6ef8a314a122b8453509a530a16f3955317/GitUI/Translation/German.xlf#L3872-L3874

This is how it supposed to look in German language:
image

But the buttons must not be as wide for English, or any other language. The layouts of the panel and buttons need to be changed to be fluent and autosize based on content.

pmiossec:
> But the buttons must not be as wide for English, or any other language. The layouts of the panel and buttons need to be changed to be fluent and autosize based on content.

I'm far to be an expert in winforms, so I'm really interested to know how to do that (without code behind) because that have always been very difficult to manage autosizing of button containing an image. I always thought that the handling of this case should have been better handled by winforms....

thimmy687:
we can also keep the origin size and just change the TextImageRelation

RussKie:
Let's fix the relationship issue. We can look at the sizing later.

RussKie (approved):

RussKie:
Thank you

PR #8763, opened 2021-01-17
Review feedback on PR #8731 (StageAll,UnstageAll)

keco2:
FollowUp on PR #8731

After implementing a minor improvement on FormCommit regarding StageAll and UnstageAll ToolStripButtons when the file-filter is active (PR #8731 / Issue #8596) @gerhardol highlighted 2 potential improvements in his review.
As the original PR has been already merged, I've opened this one.

## Proposed changes

Reason: Maintainablity
 * remove status variable _filterActive (as Gerhard mentioned: someone else may reuse the status variable for a slightly separate use case or add a parallel variable)
  Thus, FilterChanged will now fire on every character added/removed (instead of only when activated+deactivated) but this won't cause any issue.
 * Unset tool(Un)StageAllItem.Text : to avoid unecessary double translation text as this (in fact ToolTipText) is intialized anyway (in the constructor).

### Before
N/A 

### After
N/A (unchanged)

## Test methodology 

- Manual and existing IntegrationTest

## Test environment(s) 

-    GitExtensions 33.33.33 (Build 8adeae7b)
-    GIT 2.30.0.windows.1
-    Windows 10 20H2

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -967,7 +967,6 @@ namespace GitUI.CommandsDialogs
             this.toolStageAllItem.ImageTransparentColor = System.Drawing.Color.Magenta;
             this.toolStageAllItem.Name = "toolStageAllItem";
             this.toolStageAllItem.Size = new System.Drawing.Size(23, 23);
-            this.toolStageAllItem.Text = "Stage all";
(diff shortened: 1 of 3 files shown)


gerhardol (approved):
With this change, the text and image is updated every time something is typed, so no deep calculation (the actual filtering uses more cpu)
(I had a late comment and this was not blocking.)

PR #8774, opened 2021-01-18
Add mnemonics to RevisionGrid context menu

mstv:
## Proposed changes

- Add mnemonics to the RevisionGrid's context menu in order to speed-up the selection of the menu items
  <kbd>Right Click</kbd>+<kbd>letter</kbd>! (of course also with the dedicated context menu key or <kbd>Shift</kbd>+<kbd>F10</kbd>)

## Screenshots 

Before | After
-|-
grafik|grafik
grafik|grafik
grafik|grafik
grafik|grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 5a43ed39e3ec1bbe8184e81458c8691e3a7d3446
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -8876,7 +8876,7 @@ See the changes in the commit form.</source>
         <target />
       </trans-unit>
       <trans-unit id="archiveRevisionToolStripMenuItem.Text">
-        <source>Archive this commit...</source>
+        <source>Arch&amp;ive this commit...</source>
         <target />
       </trans-unit>
       <trans-unit id="bisectSkipRevisionToolStripMenuItem.Text">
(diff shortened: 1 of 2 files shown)


gerhardol (approved):

RussKie:
> `Revert commit` and `View build report`. Just hit it multiple times, then Enter.

This isn't good - one affects the state of the codebase, the other is a utility function. Propose using a different shortcut for `View build report`

mstv:
I had if there was one left. But the remaining "F, J, Q, X, Z" do not occur at all.
So I presume the winner will be `View build report`, won't it?
But why chose a victim?
My intention is *not* to share mnemonics for similar functions.
The `Revert` dialog is not dangerous: It cannot be committed just with `Enter` by mistake.

RussKie:
Thinking out loud, but I can't really see the need for the "View" menu in the context menu.

mstv:
> I can't really see the need for the "View" menu in the context menu.

That's a valid question. So I removed its mnemonic again, rebased and updated the screenshot in the PR description.

RussKie:
> > I can't really see the need for the "View" menu in the context menu.
> 
> That's a valid question. So I removed its mnemonic

👍 Though the point I was trying to make - I don't think "View"a or "Navigate" are appropriate in this menu at all. But we can tackle this separately.

Where do these menu come from?
image
"View" still has a mnemonic in those.

mstv:
>  I was trying to make - I don't think "View"a or "Navigate" are appropriate in this menu at all.

Long time ago, somebody decided to add `Navigate` and `View` to both the main menu and to the context menu.

> But we can tackle this separately.

Yes, please.

> Where do these menu come from?

They are submenus of the RevGrid context menu. I forgot to update them. Caught up on.

RussKie (approved):

PR #8789, opened 2021-01-23
C#9 omit type in new

### LINKED ISSUE #8788: C# 9 omit type in new()
Follow up to #8728

## Feature description

https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features

From the example:
private List<WeatherObservation> _observations = new();

Occasionally used already. Could be globally replaced.

* [x] Simple new #8729 
About 700 occurrences (with my search `(\s(\w+)\s.*\w+.*= new) \2\(`), a lot of the matches are TranslationString

* [ ] Avoid type when arguments are known, like in lists constructors
Probably OK in case by case, no global replace?

```c#
var list = new List<ConflictData>();
list.Add(new(item[0], item[1], item[2]));
```

* [ ] Assign where type is the same, see https://github.com/gitextensions/gitextensions/pull/8789#pullrequestreview-574884866
Probably not global (reverted from first WIP)
Some hundred where the variable name is the same as the type (that was in the first WIP)

* [ ] Replace var with types as below
About 1300 occurrences `var (\w+) = new (\w+)\(` 
However chained calls are not handled so manual reverts are required
About 1100 occurrences with `var (\w+) = new (\w+)\(([^\)\(]*\)[^.])`, a handful manual reverts needed (missing someoccurrences)

```c#
var name = new Type();
```
with
```c#
Type name = new();
```

## Environment

master

gerhardol:
Part of #8788

## Proposed changes

Regex replace (with a few manual changes)
The issue has a few other possible changes

## Test methodology 

manual

----


GitUI/CommandsDialogs/AboutBoxDialog/FormContributors.cs
```diff
@@ -80,7 +80,7 @@ TabPage GetNewTabPage(TextBox textBox, string caption)
                     {
                         BorderStyle = BorderStyle.None,
                         Margin = new Padding(0),
-                        Padding = new Padding(0),
+                        Padding = new(0),
```
mstv, line 83: I like the feature because one does not need to repeat the type. But I must get used to it if the ctor has arguments.
If `Margin = new(0)` would work, I would not like it.
`Padding = new(0)` is on the borderline already because the type is not visible explicitely.

gerhardol: I can revert this type of change (really reapply from scratch).

gerhardol:
Sorry @drewnoakes I should have set this to WIP to not complicate your changes...
I believe your changes should mostly apply clean on these changes though.

drewnoakes:
I will see. Probably many conflicts. Don't worry. How much manual work went into this? I may revert it if needed.

I'm down from 700 errors to about 80 on my branch tonight. Hope to get a PR in soon.

drewnoakes:
I believe both Roslyn and ReSharper can apply the target typed new clean up automatically.

gerhardol:
> I will see. Probably many conflicts. Don't worry. How much manual work went into this? I may revert it if needed.

This was a regex, so no effort

> I'm down from 700 errors to about 80 on my branch tonight. Hope to get a PR in soon.

Great!

PR #8790, opened 2021-01-23
Cache azuredevops build results

pmiossec:
## Proposed changes

- Azure DevOps build results are kept in memory to:
   - display result for finished builds a lot quicker
   - is server friendly (requests return less results i.e. most of the times none...

I use it for months and is working well.
I would like, with this PR, open a discussion.
* Do you think that this feature is interesting? Because I think that would be not too difficult to generalize the feature for all the Build Server Integration plugins.
* I'm not fully satisfied by the way to handle repository change and so cache cleaning. It's somewhat a dirty hack. I already found it not perfect when I did  #7477 and I would prefer to have a new method in `BuildServerWatcher` called when closing the current open repository...

What do you think? If I can expect to have this feature merged, I could try to work on these improvements...

## Test methodology 

- Manual (I use it since months)

## Test environment(s) 

- Git Extensions 33.33.33
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)


gerhardol:
So cache is kept until switching repo?
What if new builds are done on the same commit?

RussKie:
I often flip between repos back and forth, so in my case caching build results for the lifetime of the app is desirable.

How may build results do we normally store per repo? How much memory does it take? 
Having these number may help us better determine how much memory we can allocate to be both useful and not too wasteful.

pmiossec:
> So cache is kept until switching repo?

That's what this PR is doing, yes. 

> What if new builds are done on the same commit?

Like before. The last one is displayed.
I should have explained a little what is it doing. 
Because some builds has been kept in cache, when refreshing the revision grid, instead of doing a request to get ALL the finished builds, just requesting the ones after the last build result we have in the cache. 



> I often flip between repos back and forth, so in my case caching build results for the lifetime of the app is desirable.

I understand this point of view but that's another work because if we do that, we should let the user empty the cache, also add a setting to enable or disable it,....

But that is indeed something to keep in mind in the PR to not do something that will prevent it. 

> How may build results do we normally store per repo?

I did not changed the behavior so it depends of the builds. The request retrieve all the builds and we cache them after a de duplication by commits. So it depends how many pipelines you track and your retention policy for these pipelines. 

> How much memory does it take?

I really don't know and didn't had a look because I think this PR perhaps use less memory. But if we will cache closed repositories, that is another thing.... 

> Having these number may help us better determine how much memory we can allocate to be both useful and not too wasteful.

Yes, it will be interesting to know...

gerhardol:
> Because some builds has been kept in cache, when refreshing the revision grid, instead of doing a request to get ALL the finished builds, just requesting the ones after the last build result we have in the cache.

Then I have no objections (understatement, a good change)
Note that Jenkins will not benefit from this. You can query when a build finished, but if you request builds you get all so there is no big benefit with caching.

GintasS:

RussKie:
@pmiossec is this something we can bring to the mergrable state?

pmiossec:
> @pmiossec is this something we can bring to the mergrable state?

I don't know the state of this branch but I use it since 2years without problems in my own build.
So it should not be a problem to update this branch.

GitUI/BuildServerIntegration/BuildServerWatcher.cs
```diff
@@ -39,6 +39,7 @@ public sealed class BuildServerWatcher : IBuildServerWatcher, IDisposable
         private readonly IRepoNameExtractor _repoNameExtractor;
         private IDisposable? _buildStatusCancellationToken;
         private IBuildServerAdapter? _buildServerAdapter;
+        private readonly object _observerLock = new object();
```
mstv, line 42: Can be shortened in the meantime:
```suggestion
        private readonly object _observerLock = new();
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -219,13 +222,44 @@ void ManageError(string errorMessage)
...
+                        observer.OnNext(CreateBuildInfo(build));
+                    }
+                }
+                else
+                {
+                    var buildsByCommit = builds.GroupBy(b => b.SourceVersion);
+                    foreach (var buildsForACommit in buildsByCommit)
+                    {
+                        var buildToDisplay = buildsForACommit.OrderByDescending(b => b.FinishTime).First();
+                        var buildInfo = CreateBuildInfo(buildToDisplay);
```
mstv, line 252: Ease reading by explicitly specifying types, please.

mstv (commented):
Looks sensible.
Though `CacheAzureDevOps` should be named correctly: `AzureDevOpsCache`.
Nested functions like `ManageError` should be moved to the end of the function.
This refactoring can be done in a separate PR.

pmiossec:
> @pmiossec is this something we can bring to the mergrable state?

done. Could be reviewed. I will most likely squash most of the commits.

Sorry @mstv you have started you review a little too early and I pushed some refactoring cleaning after (luckily fixing most of your comments 😉 )

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +145,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                ManageError(errorMessage);
 
-                foreach (var build in builds)
+                if (_projectOnErrorKey is null || _projectOnErrorKey != CacheKey)
                 {
-                    observer.OnNext(CreateBuildInfo(build));
+                    _projectOnErrorKey = CacheKey;
+
+                    TaskDialogButton btnOpenSettings = new("Open settings");
+                    TaskDialogButton btnIgnore = new("Ignore");
```
RussKie, line 234: Do we need to loc these?

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +145,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                {
+                    _projectOnErrorKey = CacheKey;
+                    MessageBox.Show(errorMessage, _buildIntegrationErrorCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                }
+
+                return false;
             }
-            catch (Exception ex)
+
+            void ManageError(string errorMessage)
```
RussKie, line 269: Do you think we can come up with a bit more descriptive name?

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +145,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                    }
                 }
+
+                return false;
             }
-            catch (OperationCanceledException)
+            catch (Exception ex)
             {
-                // Do nothing, the observer is already stopped
+                string errorMessage = _genericErrorMessage.Text + ex.Message;
```
RussKie, line 257: String interpolation is getting perf optimisations, so it's better to do this:
```suggestion
                string errorMessage = $"{_genericErrorMessage.Text}{ex.Message}";
```

Shouldn't there be some sort of a separator? E.g., like a new line or something like that...

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +145,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
-                        string errorMessage = _genericErrorMessage.Text + ex.Message;
-                        ManageError(errorMessage);
 
-                        if (ProjectOnErrorKey is null || ProjectOnErrorKey != CacheKey)
+                    IEnumerable<Build>? builds = await _apiClient.QueryFinishedBuildsAsync(_buildDefinitions, sinceDate);
+                    IEnumerable<IGrouping<string?, Build>>? buildsByCommit = builds.GroupBy(b => b.SourceVersion);
+
+                    foreach (IGrouping<string?, Build>? buildsForACommit in buildsByCommit)
+                    {
+                        Build? buildToDisplay = buildsForACommit.OrderByDescending(b => b.FinishTime).First();
```
RussKie, line 181: I see below you're using `MinBy(...)`. Wouldn't it work here too?

Plugins/BuildServerIntegration/AzureDevOpsIntegration/BuildsCache.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using GitUIPluginInterfaces.BuildServerIntegration;
+
+namespace AzureDevOpsIntegration
+{
+    public class BuildsCache
+    {
+        public string? Id { get; set; }
+        public string? BuildDefinitions { get; set; }
+        public List<BuildInfo> FinishedBuilds { get; set; } = new List<BuildInfo>();
```
RussKie, line 11: Should this be?
```suggestion
        public List<BuildInfo> FinishedBuilds { get; init; } = new List<BuildInfo>();
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/BuildsCache.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using GitUIPluginInterfaces.BuildServerIntegration;
+
+namespace AzureDevOpsIntegration
+{
+    public class BuildsCache
```
RussKie, line 7: Are any of the fields get mutated after the initial initialisation? I couldn't quite see it in the browser.
If not, should we change this into a [record](https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/record)?

Plugins/GitUIPluginInterfaces/BuildServerIntegration/IBuildServerAdapter.cs
```diff
@@ -15,5 +15,14 @@ public interface IBuildServerAdapter : IDisposable
         IObservable<BuildInfo> GetFinishedBuildsSince(IScheduler scheduler, DateTime? sinceDate = null);
 
         IObservable<BuildInfo> GetRunningBuilds(IScheduler scheduler);
+
+        /// <summary>
+        /// Called when the current repository is closed
+        /// </summary>
+        void RepositoryClosed()
```
RussKie, line 22: Do you think we could align with the .NET API semantics?
```suggestion
        void OnRepositoryChanged()
```

Plugins/GitUIPluginInterfaces/BuildServerIntegration/IBuildServerAdapter.cs
```diff
@@ -15,5 +15,14 @@ public interface IBuildServerAdapter : IDisposable
         IObservable<BuildInfo> GetFinishedBuildsSince(IScheduler scheduler, DateTime? sinceDate = null);
 
         IObservable<BuildInfo> GetRunningBuilds(IScheduler scheduler);
+
+        /// <summary>
+        /// Called when the current repository is closed
```
RussKie, line 20:
```suggestion
        ///  Provides an extension point for handling the switch of repositories.
        ///  For example, it could be used to clear build changes.
```

RussKie (commented):
LGTM with few questions and suggestions.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
pmiossec: `By`s methods are equivalent of `.FirstOrDefault()` so I prefer to keep it like that.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
pmiossec: Done even if the translation is not supported for plugins...

Plugins/BuildServerIntegration/AzureDevOpsIntegration/BuildsCache.cs
pmiossec: Yes, at least one is mutated so not possible.

pmiossec:
Done. Fixed everything I think was doable.
And I squashed all the commits too.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +147,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
-                            return;
-                        }
+                    return;
+                }
 
-                        CacheAzureDevOps = new CacheAzureDevOps { Id = CacheKey, BuildDefinitions = _buildDefinitions };
+                if (running)
+                {
+                    IEnumerable<Build>? builds = FilterRunningBuilds(await _apiClient.QueryRunningBuildsAsync(_buildDefinitions));
+                    foreach (Build? build in builds)
```
mstv, line 159: ❔ really necessary
```suggestion
                    foreach (Build build in builds)
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +147,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                else
+                {
+                    // Display cached builds results
+                    if (!sinceDate.HasValue && _buildsCache.FinishedBuilds.Any())
                     {
-                        string errorMessage = _badTokenErrorMessage.Text;
-                        ManageError(errorMessage);
-
-                        if (ProjectOnErrorKey is null || ProjectOnErrorKey != CacheKey)
+                        foreach (BuildInfo? buildInfo in _buildsCache.FinishedBuilds)
```
mstv, line 169: ❔ ditto
```suggestion
                        foreach (BuildInfo buildInfo in _buildsCache.FinishedBuilds)
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -144,97 +147,132 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                    TaskDialogButton result = TaskDialog.ShowDialog(page);
+                    if (result == btnOpenSettings)
+                    {
+                        _projectOnErrorKey = null;
+                        Validates.NotNull(_openSettings);
+                        _openSettings.Invoke();
+                    }
                 }
+
+                return false;
```
mstv, line 255: Nit: Same as at the end of the `catch`, i.e. can be moved behind the `catch` block.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -245,14 +283,14 @@ private IEnumerable<Build> FilterRunningBuilds(IList<Build> runningBuilds)
...
             }
 
-            var byCommitBuilds = runningBuilds.GroupBy(b => b.SourceVersion);
+            IEnumerable<IGrouping<string?, Build>>? byCommitBuilds = runningBuilds.GroupBy(b => b.SourceVersion);
             runningBuilds = new List<Build>();
 
             // Filter running builds to display the best build as we can only display one build for a commit
             // by selecting the first started or if none, one that is waiting to start
-            foreach (var commitBuilds in byCommitBuilds)
+            foreach (IGrouping<string?, Build>? commitBuilds in byCommitBuilds)
```
mstv, line 291: ❔ ditto
```suggestion
            foreach (IGrouping<string?, Build> commitBuilds in byCommitBuilds)
```

mstv (approved):
Just nits

pmiossec:
thanks @mstv. DOne.

gerhardol:
+1
Brief review, but works for me in AppVeyor and Jenkins still

RussKie (approved):

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
RussKie: If I understand it correctly, the plugins that are bundled with the app are generally translated. Are we talking about the same thing?

PR #8799, opened 2021-01-25
Revert "Add test accessors borrowed from https://github.com/dotnet/winforms/"

sharwell:
The test accessors are design to preserve encapsulation in production code without interfering with the ability to use IDE navigation tools to see test code that interacts with production code internals. Pull request #8787 (unintentionally) defeated this purpose.

Note that dotnet/winforms cannot use the TestAccessor pattern from GitExtensions (even though it provides a better developer experience) due to particularly strict limitations on accessibility and production code internals. Moving even one layer away from the BCL (e.g. dotnet/roslyn) restores the ability to use the full TestAccessor pattern.

Note that refactorings to assist with the creation and maintenance of TestAccessor types is available through Roslyn.Diagnostics.Analyzers (https://github.com/dotnet/roslyn-analyzers/pull/2164).

Reverts gitextensions/gitextensions#8787


RussKie:
Yes, I see your point. The reflection makes references opaque.

PR #8801, opened 2021-01-26
Fix filter by branch with globbing characters *?[]

### LINKED ISSUE #8800: Filter by branches does not apply if contains globbing characters [?*
## Current behaviour

- Type `branch_name_prefix_*` into branches combobox text area, press Enter.

Actual result: empty revision grid

## Expected behaviour
Revision grid is filtered to show commits from any branch with name starting with `branch_name_prefix_`

## Screenshots

Before I input the filter, you can see there is a branch named `CSMS-9313_migrate_sender_fee`

image

Then I input a filter with glob character by `CSMS-9313_*`

image

## Related issues

While fixing this, maybe we can support filter by multiple branches as well

https://github.com/gitextensions/gitextensions/issues/491
https://github.com/gitextensions/gitextensions/issues/3546

## Diagnostics

The actual code is aware of globbing characters, the branches filter with globbing is converted to `--branches=...` parameter of `git log` command

https://github.com/gitextensions/gitextensions/blob/fb0aa6d06a2c78a5f25ea6a9e351d505e85be638/GitCommands/RevisionReader.cs#L194

However `refFilterOptions.HasFlag(RefFilterOptions.Branches)` is never `true`, so the option `--branches` is never added.

NikolayXHD:
Fixes #8800

## Test methodology

Manually

## Test environment(s)

Windows 10

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -172,43 +172,45 @@ namespace GitCommands
             string revisionFilter,
             string pathFilter)
         {
-            return new GitArgumentBuilder("log")
+            var builder = new GitArgumentBuilder("log");
+            builder.Add("-z");
+            builder.Add($"--pretty=format:\"{FullFormat}\"");
+            if (!string.IsNullOrWhiteSpace(branchFilter) && isBranchFilterSimple(branchFilter))
             {
-                "-z",
-                branchFilter,
-                $"--pretty=format:\"{FullFormat}\"",
+                builder.Add(branchFilter);
+            }
+
+            builder.Add(refFilterOptions.HasFlag(RefFilterOptions.FirstParent), "--first-parent", new ArgumentBuilder
+            {
+                { refFilterOptions.HasFlag(RefFilterOptions.Reflogs), "--reflog" },
+                { AppSettings.SortByAuthorDate, "--author-date-order" },
                 {
-                    refFilterOptions.HasFlag(RefFilterOptions.FirstParent),
-                    "--first-parent",
+                    refFilterOptions.HasFlag(RefFilterOptions.All),
+                    "--all",
                     new ArgumentBuilder
                     {
-                        { refFilterOptions.HasFlag(RefFilterOptions.Reflogs), "--reflog" },
-                        { AppSettings.SortByAuthorDate, "--author-date-order" },
                         {
-                            refFilterOptions.HasFlag(RefFilterOptions.All),
-                            "--all",
-                            new ArgumentBuilder
-                            {
-                                {
-                                    refFilterOptions.HasFlag(RefFilterOptions.Branches) &&
-                                    !string.IsNullOrWhiteSpace(branchFilter) && branchFilter.IndexOfAny(new[] { '?', '*', '[' }) != -1,
-                                    "--branches=" + branchFilter
-                                },
-                                { refFilterOptions.HasFlag(RefFilterOptions.Remotes), "--remotes" },
-                                { refFilterOptions.HasFlag(RefFilterOptions.Tags), "--tags" },
-                            }.ToString()
+                            refFilterOptions.HasFlag(RefFilterOptions.Branches) &&
+                            !string.IsNullOrWhiteSpace(branchFilter) && !isBranchFilterSimple(branchFilter),
+                            "--branches=" + branchFilter
                         },
-                        { refFilterOptions.HasFlag(RefFilterOptions.Boundary), "--boundary" },
-                        { refFilterOptions.HasFlag(RefFilterOptions.ShowGitNotes), "--not --glob=notes --not" },
-                        { refFilterOptions.HasFlag(RefFilterOptions.NoMerges), "--no-merges" },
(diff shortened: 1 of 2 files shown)


GitCommands/RevisionReader.cs
```diff
@@ -172,44 +172,46 @@ public sealed class RevisionReader : IDisposable
...
+            var builder = new GitArgumentBuilder("log");
+            builder.Add("-z");
+            builder.Add($"--pretty=format:\"{FullFormat}\"");
+            if (!string.IsNullOrWhiteSpace(branchFilter) && isBranchFilterSimple(branchFilter))
             {
-                "-z",
-                branchFilter,
-                $"--pretty=format:\"{FullFormat}\"",
+                builder.Add(branchFilter);
+            }
```
RussKie, line 181: The original syntax was terse, and thus I'm finding it a bit more readable.

What's wrong with the following?
```cs
            return new GitArgumentBuilder("log")
            {
                "-z",
                branchFilter,
                { !string.IsNullOrWhiteSpace(branchFilter) && isBranchFilterSimple(branchFilter), branchFilter }
            }
```

GitCommands/RevisionReader.cs
```diff
@@ -172,44 +172,46 @@ public sealed class RevisionReader : IDisposable
...
+                { refFilterOptions.HasFlag(RefFilterOptions.NoMerges), "--no-merges" },
+                { refFilterOptions.HasFlag(RefFilterOptions.SimplifyByDecoration), "--simplify-by-decoration" }
+            }.ToString());
+            builder.Add(revisionFilter);
+            builder.Add("--");
+            builder.Add(pathFilter);
+            return builder;
         }
 
+        private static bool isBranchFilterSimple(string branchFilter) =>
```
RussKie, line 212: [naming]
```suggestion
        private static bool IsSimpleBranchFilter(string branchFilter) =>
```

GitCommands/RevisionReader.cs
NikolayXHD: The sequence `"-z", branchFilter` is only valid if the branch filter is simple

```bash
# valid
git log -z full_branch_name
git log --branches=*_name

#invalid
git log -z *-name
git log -z "*-name"
```

NikolayXHD: thus I need the `if` which breaks initializer-like syntax:
```
if (is_simple_branch_filter)
  builder.Add(branchFilter);
```

NikolayXHD: Somehow I missed the possibility to
```
{boolean_condition, the_flag}
```
I was under impression, that the only possibility of conditional flag was ternary semantics
```
{boolean_condition, flag_if_true, flat_if_false}
```

GitCommands/RevisionReader.cs
```diff
@@ -172,44 +172,46 @@ public sealed class RevisionReader : IDisposable
...
-            return new GitArgumentBuilder("log")
+            var builder = new GitArgumentBuilder("log");
+            builder.Add("-z");
+            builder.Add($"--pretty=format:\"{FullFormat}\"");
+            if (!string.IsNullOrWhiteSpace(branchFilter) && isBranchFilterSimple(branchFilter))
             {
-                "-z",
-                branchFilter,
-                $"--pretty=format:\"{FullFormat}\"",
+                builder.Add(branchFilter);
```
gerhardol, line 180: --branches accepts wildcards. Is it not sufficient to remove the check for special chars?

NikolayXHD: --branches option [has a weird behavior](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---branchesltpatterngt), when no wildcards are present in the branchFilter:

>  If pattern lacks ?, *, or [, /* at the end is implied

This is why we have to check if these characters are in branch filter, and where they are not, we use `git log branchFilter` instead

GitCommands/RevisionReader.cs
NikolayXHD: fixed

GitCommands/RevisionReader.cs
NikolayXHD: ok

GitCommands/RevisionReader.cs
```diff
@@ -175,8 +175,11 @@ public sealed class RevisionReader : IDisposable
             return new GitArgumentBuilder("log")
             {
                 "-z",
-                branchFilter,
                 $"--pretty=format:\"{FullFormat}\"",
+                {
```
gerhardol, line 179: Any reason to switch order?

I would pprefer to move this option close to the other where branchFilter is handled (either here or to line 194)

gerhardol (commented):
+1
You get "no results" with wildcards. Not really a problem.
Useful.

image

GitCommands/RevisionReader.cs
NikolayXHD: > Any reason to switch order?

No reason, my mistake. Moved back to where it was.

PR #8802, opened 2021-01-26
Io exception

mstv:
Follow-up to #8718

## Proposed changes

- Improve display of exceptions
- Make `ExternalOperationException` slightly more generic
- Draft `FileSystemWrapper`
- Substitute `ExternalOperationException` for `FileDeleteException`

## Screenshots

Exception|Before|After
-|-|-
`Exception`|grafik|grafik
`ExternalOperationException`|grafik|grafik
`UserExternalOperationException`|grafik|grafik
`UserExternalOperationException`|grafik|grafik

## Test methodology 

- added unit tests
- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7efa007ce6cb45c97745fafe0593f030b1e69b2d
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----


mstv:
I'll rebase and resolve the conflicts after a first review round.

gerhardol:
The description:
Intermediate error, Wrapping exception; can we explain that better and not duplicate? _Inner error_ _first error_?
If Command, Arguments or Workdir is not available, skip displaying?
Why Root error? Is that part of the exception message?

mstv:
Yep, I forgot to mention: The screenshots show dummy exceptions generated using https://github.com/mstv/gitextensions/commit/7efa007ce6cb45c97745fafe0593f030b1e69b2d just for comparison:
```
throw new ExternalOperationException("<command>", "<arguments>", "<workdir>",
    new Exception("Wrapping exception.", new Exception("Intermediate error.", new Exception("Root error, e.g. file locked."))));
```
and using the script:
grafik

mstv:
I think in case of a non-`ExternalOperationException` we should **skip** the `TaskDialog`
Exception|Before|After
-|-|-
`Exception`|grafik|grafik

but open the `BugReportForm` directly - as before #8718:

grafik grafik

Thoughts?

gerhardol:
> Thoughts?

I suggest  to keep the current implementation, it is less in-your-face and makes users less irritated.
I doubt there is a significant difference in useful issues.

gerhardol:
I have tried to review this and have no comments but may come back some other day when I have found my lost concentration.

GitExtUtils/FileSystemUtility.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.IO;
+
+namespace GitExtUtils
+{
+    public static class FileSystemUtility
+    {
+        /// <summary>
+        /// Returns the current working directory for display purposes or the exception message in the unlikely case of inability.
+        /// </summary>
+        public static string GetWorkingDirectoryNoEx()
```
RussKie, line 11: Please remove it. This has a single call site, and the return value of this method incoherent - "_get current directory or exception message_".

GitCommands/Git/IndexLockManager.cs
```diff
@@ -100,7 +95,7 @@ private void DeleteIndexLock(string fileName)
             }
             catch (Exception ex)
             {
-                throw new FileDeleteException(fileName, ex);
+                throw new ExternalOperationException(command: "File.Delete", arguments: fileName, FileSystemUtility.GetWorkingDirectoryNoEx(), ex);
```
RussKie, line 98: I struggle to see any value in `command: "File.Delete"` for a user. Also `Directory.GetCurrentDirectory()` isn't appropriate here, it must be the the current repo working directlry.
```suggestion
                throw new ExternalOperationException(command: null, arguments: fileName, workingDir, ex)
```

GitUI/Strings.cs
```diff
@@ -98,10 +95,13 @@ internal sealed class Strings : Translate
 
         private readonly TranslationString _rotInactive = new TranslationString("[ Inactive ]");
 
-        private readonly TranslationString _argumentsText = new TranslationString("Arguments");
-        private readonly TranslationString _commandText = new TranslationString("Command");
-        private readonly TranslationString _workingDirectoryText = new TranslationString("Working directory");
-        private readonly TranslationString _reportBugText = new TranslationString("If you think this was caused by Git Extensions, you can report a bug for the team to investigate.");
+        private readonly TranslationString _userExternalOperationFailed = new TranslationString("User-defined external operation failed");
+        private readonly TranslationString _externalOperationFailed = new TranslationString("External operation failed");
```
RussKie, line 99: Most users will likely find this confusing - "user-defined" and "external" errors is something we understand, but users likely wouldn't know a difference - _the app failed, omg, help!_
That was the reason to word the original messages as I did.

RussKie:
> `UserExternalOperationException`	grafik	grafik

More information doesn't necessarily make a better user experience - too much information will likely scare a user. Please have a look at https://docs.microsoft.com/en-us/windows/win32/uxguide/mess-error

I'm not finding the proposed dialog better, there are a lot of arguably irrelevant information here - the user will have to stop and _really_ read all of it to try to understand what happened (which is "the app couldn't delete the index.lock file"). It is pretty obvious in the furst screenshot and totally not in the second.

GitUI/TaskDialogExtensions.cs
```diff
@@ -0,0 +1,21 @@
+#nullable enable
+
+using System;
+using Microsoft.WindowsAPICodePack.Dialogs;
+
+namespace GitUI
+{
+    public static class TaskDialogExtensions
+    {
+        public static void AddButton(this TaskDialog taskDialog, string text, Action? action = null, string? name = null)
```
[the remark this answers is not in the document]
mstv: I count more than 20! To be replaced in a follow-up PR.

GitExtUtils/FileSystemUtility.cs
mstv: It will also be reused when we create a wrapper for `FileSystem` which throws `ExternalOperationExceptions`.
I am always open for better names (not my forte).

GitCommands/Git/IndexLockManager.cs
mstv: >  Also Directory.GetCurrentDirectory() isn't appropriate here, it must be the the current repo working directlry.

I cannot agree. Filesystem operations are relative to the CWD not to the repo.
The CWD does not matter only if the `argument` contains the full path.
The repo path - or the worktree folder in case of the screenshot - is contained in the `argument`.

> I struggle to see any value in command: "File.Delete" for a user.

As already proposed, these strings could be translated by the `BugReporter`, if you do not want translatable strings in the `FileSystem` wrapper to be created.
I did not want to put too much into one PR.

GitUI/Strings.cs
mstv: OK

mstv:
> More information doesn't necessarily make a better user experience - too much information will likely scare a user. Please have a look at https://docs.microsoft.com/en-us/windows/win32/uxguide/mess-error

I see the point. That's why I have already omitted the exception types.
What does it mean in the example?
- Uncontentious, the working directory does not add any value if the argument contains a full path.
- "Delete" and "index.lock" are shown twice. This can be improved.
- Disputable: The `Message`s of the wrapping exceptions might not add much information. GE is not such complex.
  (The screenshot shows a synthetic example with three nested exceptions.)

> I'm not finding the proposed dialog better, there are a lot of arguably irrelevant information here - the user will have to stop and really read all of it to try to understand what happened (which is "the app couldn't delete the index.lock file"). It is pretty obvious in the furst screenshot and totally not in the second.

I cannot agree because two important pieces of information are missing:
- Why?: Message from the OS.
- Where?: Full path to the affected file.

I agree, the presentation is not perfect and can be improved. Perhaps:

```
Error
-----

The file is locked.

Operation: Delete index.lock
Directory: F:\Build\gitextensions3\.git\worktrees\gitextensions3_dev 

| --> Ignore |
```

Though we do not have the manpower to catch, edit and translate the exceptions everywhere they can come along in the UI.
And I strongly prefer to handle this aspect in one place.

I think it can be mapped to a slightly modified `ExternalOpertionException`:
Failing function | .Operation | .Arguments | .Directory
-|-|-|-
Process.Start | $"Start {command}" | arguments | workingDir
File.Delete("D:\path\filename") | "Delete filename" | null | "D:\path"
File.Delete("relpath\filename") | "Delete filename" | null | $"{GetWorkingDirNoEx()}\relpath"

The `FileSystem` wrapper must be kept simple and *must not* throw while preparing the details of the `ExternalOpertionException`.
The distinction of case for the `File.Delete` might already be too much logic in the wrapper.

mstv:
I have adressed the review comments and updated the issue description.
Screenshots were created using https://github.com/mstv/gitextensions/commit/ba53e5922231f4b669f6afddeb35cf3fc023fc28.
If I should extract undisputed parts to a separate PR, please advise.
TODO: Add tests for the draft of `FileSystemWrapper` after next review.

mstv:
Rebased and squashed
I would like to extract the more informative error dialog for 3.5 (without changing the `ExternalOperationException`).

GintasS:
This is an old PR and should be closed. @gerhardol

mstv:
A few things from this PR are still required to finish #7795.

PR #8804, opened 2021-01-26
Introduce new interface for named git refs

drewnoakes:
## Proposed changes

This comes ahead of work to null-annotate these APIs. Not all implementations will populate this property, so in the old factoring it should be marked as nullable. However the Name property is used in many places where it is assumed non-null.

This factoring splits out those usages behind a new interface, making it easier to correctly annotate these APIs.

## Test methodology 

- Manual testing
- It compiles

----

### DIFF
--- a/Plugins/GitUIPluginInterfaces/IGitItem.cs
+++ b/Plugins/GitUIPluginInterfaces/IGitItem.cs
@@ -6,6 +6,10 @@ namespace GitUIPluginInterfaces
     {
         ObjectId? ObjectId { get; }
         string? Guid { get; }
+    }
+
+    public interface INamedGitItem : IGitItem
+    {
         string Name { get; }
     }
 }
--- a/GitCommands/Git/GitItem.cs
+++ b/GitCommands/Git/GitItem.cs
@@ -4,7 +4,7 @@ using GitUIPluginInterfaces;
 namespace GitCommands.Git
 {
     [DebuggerDisplay("GitItem( {" + nameof(FileName) + "} )")]
-    public class GitItem : IGitItem
+    public class GitItem : INamedGitItem
     {
         public GitItem(int mode, GitObjectType objectType, ObjectId objectId, string name)
         {
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3188,7 +3188,7 @@ namespace GitCommands
                 .Split('\0', '\n');
         }
 
-        public IEnumerable<IGitItem> GetTree(ObjectId commitId, bool full)
+        public IEnumerable<INamedGitItem> GetTree(ObjectId commitId, bool full)
         {
             var args = new GitArgumentBuilder("ls-tree")
             {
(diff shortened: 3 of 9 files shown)


Plugins/GitUIPluginInterfaces/IGitItem.cs
```diff
@@ -6,6 +6,10 @@ public interface IGitItem
     {
         ObjectId? ObjectId { get; }
         string? Guid { get; }
+    }
+
+    public interface INamedGitItem : IGitItem
```
RussKie, line 11: Please move to own file.

RussKie (approved):

Plugins/GitUIPluginInterfaces/IGitItem.cs
drewnoakes: If this is required, let's add an analyzer for this. There are loads of violations of this today.

gerhardol (approved):

Plugins/GitUIPluginInterfaces/IGitItem.cs
RussKie: There's plenty of legacy code that doesn't meet the current standards. But it doesn't mean we should add to the tech debt pile.

As far as analyzers are concerned I'm not sure there's a good one for this case, it is a bit more nuanced. In my experience an interface should be in the same file with a concrete class, unless there is more than one concrete implementation (this is for a typical IoC scenario). Having an interface in the same file significantly reduces maintenance as generally an interface and class change together.
Re# by default offer to do that when extracting an interface, whilst VS by default would place an interface in its own file, and every time I had to use this approach it was a constant context switching nightmare.

On a flipside, if an interface not in the same file, I'd typically look for a file with the same name as the interface. Hence having an interface (or any type for that matter) hidden inside some other file becomes a maintenance overhead because I don't have a luxury of browsing code in an IDE all the time.

drewnoakes: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1402.md will enforce this.

I'd vote for enforcing this with that analyzer to save time during reviews. I've moved my branch on from this PR and the cost of switching back to tweak it is an (admittendly minor) barrier to contribution.

gerhardol: About 90 occurrences right now, maybe 50 files.
I am not in favor of this rule at all though. It causes code to move around so blame is much harder.
It especially adds an hurdle in adding data classes to group information causing a lot of separate but related status variables or tuples.

But if this is a code style recommendation, I rather use an enforcing rule.

RussKie: > [https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1402.md](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1402.md?rgh-link-date=2021-01-31T11%3A13%3A50Z) will enforce this.
> 
> I'd vote for enforcing this with that analyzer to save time during reviews.

Works for me 👍

PR #8805, opened 2021-01-29
Increase the number of commit templates

pmiossec:
## Proposed changes

- Increase to 10 the number of commit templates and migrate the existing ones when increasing the size

Reason behind the change:

- Because only the not empty lines are displayed in the menu, there is no drawback to allow more commit templates
- I need more lines 😉 (so perhaps someone else also...)

### Before

I can't produce it anymore (but there are only 5 lines 😉  )

### After

image

## Test methodology 

- Manual 

## Test environment(s) 

- Git Extensions 33.33.33
- .NET Framework 4.8.4261.0
- DPI 192dpi (200% scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/CommitDialog/FormCommitTemplateSettings.cs
+++ b/GitUI/CommandsDialogs/CommitDialog/FormCommitTemplateSettings.cs
@@ -11,7 +11,7 @@ namespace GitUI.CommandsDialogs.CommitDialog
 
         private CommitTemplateItem[] _commitTemplates;
 
-        private const int _maxCommitTemplates = 5;
+        private const int _maxCommitTemplates = 10;
         private const int _maxShownCharsForName = 50;
         private const int _maxUsedCharsForName = 80;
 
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/CommitDialog/FormCommitTemplateSettings.cs
```diff
@@ -43,6 +43,15 @@ private void LoadSettings()
                     _commitTemplates[i] = new CommitTemplateItem();
                 }
             }
+            else if (_commitTemplates.Length < _maxCommitTemplates)
+            {
```
gerhardol, line 47: Add a comment that this is for migration

gerhardol (approved):

GitUI/CommandsDialogs/CommitDialog/FormCommitTemplateSettings.cs
pmiossec: done.

PR #8807, opened 2021-01-30
Fix toolbar order (revisited)

RussKie:
## Proposed changes

The designer keeps on wrecking havoc with toolbars - i.e. the order appears to be semi random and not [expected](https://github.com/gitextensions/gitextensions/pull/8719#discussion_r559495096). 
Revisit the toolbar layout workaround, and add assertions to validate the outcome.

Related #8719

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -320,17 +320,6 @@ namespace GitUI.CommandsDialogs
 
             return;
 
-            void WorkaroundToolbarLocationBug()
-            {
-                // Layout engine bug (?) which may change the order of toolbars
-                // if the 1st one becomes longer than the 2nd toolbar's Location.X
-                // the layout engine will be place the 2nd toolbar first
-                toolPanel.TopToolStripPanel.Controls.Clear();
-                toolPanel.TopToolStripPanel.Controls.Add(ToolStripScripts);
-                toolPanel.TopToolStripPanel.Controls.Add(ToolStripFilters);
-                toolPanel.TopToolStripPanel.Controls.Add(ToolStripMain);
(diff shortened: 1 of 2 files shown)


gerhardol (commented):
+1

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -511,6 +500,35 @@ Brush UpdateCommitButtonAndGetBrush(IReadOnlyList<GitItemStatus> status, bool sh
...
+                }
+
+#if DEBUG
+                // 3. Assert the layout is expected
+                for (int i = toolStrips.Length - 1; i > 0; i--)
+                {
+                    // Assert all toolbars on the same row
+                    Debug.Assert(toolStrips[i].Top == 0, $"{toolStrips[i].Name} must be on placed on the 1st row");
+
+                    // Assert the correct order of toolbars
```
mstv, line 527:
```suggestion
                // 3. Assert all toolbars on the same row
                foreach (ToolStrip toolStrip in toolStrips)
                {
                    Debug.Assert(toolStrip.Top == 0, $"{toolStrip.Name} must be placed on the 1st row");
                }

                // 4. Assert the correct order of toolbars
                for (int i = toolStrips.Length - 1; i > 0; i--)
                {
```

mstv (approved):
👍 Should work. Two nits in the assertions.

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: Why is it better?

mstv: Because it asserts *all* toolbars as the comment promises.

PR #8812, opened 2021-02-01
Prioritize main as master in CommitInfo

### LINKED ISSUE #8811: Support "main" branch
## Feature description

A number of repos (including high profile like [Microsoft .NET org](https://github.com/dotnet/announcements/issues/172)) are renaming "master" branch into "main". We must update the code to include "main" alongside "master".

gerhardol:
Fixes #8811

## Proposed changes

sort main branch first as master

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -769,7 +769,7 @@ namespace GitUI.CommitInfo
 
             private static readonly Regex _importantRepoRegex = new($"^{RemoteBranchPrefix}(origin|upstream)/",
                 RegexOptions.Compiled | RegexOptions.CultureInvariant);
-            private static readonly Regex _remoteMasterRegex = new($"^{RemoteBranchPrefix}.*/master[^/]*$",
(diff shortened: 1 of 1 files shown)


pmiossec:
We could imagine some other names : Trunk, stable, dev,...

So what do you think of reading `init.defaultBranch` setting and adding also to the list (if different of main or master)

gerhardol:
> So what do you think of reading `init.defaultBranch` setting and adding also to the list (if different of main or master)

should probably be done

but I would prefer to prioritize branches in more situations (matching a pattern)

This was mostly submitted to show that are not much to do by default

RussKie (approved):
Thank you!

I'm genuinely surprised we have only 2 instances in the code (plenty in comments).

RussKie:
> > So what do you think of reading `init.defaultBranch` setting and adding also to the list (if different of main or master)
> 
> should probably be done
> 
> but I would prefer to prioritize branches in more situations (matching a pattern)
> 
> This was mostly submitted to show that are not much to do by default

Thank you @pmiossec for the suggestion. It can be a good FR.

PR #8732, opened 2021-01-09
Fix borders when app run in >100% scaling factor

ivangrek:
Fixes #8698

## Proposed changes

- Use WndProc to draw borders

### After

image

## Test methodology 

- Visual
- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build a098b1ebba44b3325089e3e0639b4fc5e7e30957 (Dirty)
- Git 2.28.0.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4300.0
- DPI 144dpi (150% scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8815, opened 2021-02-03
Fix tab control border in high dpi

NikolayXHD:
Fixes the issue discussed in https://github.com/gitextensions/gitextensions/pull/8732#issuecomment-757439317

## Proposed changes

Let system draw tab control. To remain compatible with dark theme use Win32 api overrides.

Unfortunately tab control now flickers on mouse move. `.SetStyle(DobuleBuffer, true)` would not help,
because double buffer only works with `UserPaint | AllPaintingInWmPaint`

### Before
image

### After
image

## Test environment(s) 

- Windows 10, dpi 150%

----


ivangrek (commented):
image

Different colors

ivangrek (approved):

gerhardol (commented):
+1
looks fine to me. The implementation seems easier to me too - the best practices for overriding is not my speciality

indent issue in Win32ThemeHooks:283

NikolayXHD:
> indent issue in Win32ThemeHooks:283

Fixed

NikolayXHD:
> image
> 
> Different colors

Fixed

GitUI/Theming/Renderers/TabRenderer.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        private static void RenderPane(Context ctx, Rectangle prect)
+        {
+            var borderPen = GetBorderPen();
+            ctx.Graphics.DrawRectangle(borderPen, prect.Inclusive());
+        }
+
+        private static void RenderTab(Parts partId, States stateId, Rectangle prect, Context ctx)
+        {
+            Brush backBrush = GetTabBackgroundBrush(stateId);
```
RussKie, line 58: Doesn't it need to be disposed of?

RussKie:
The tab highlighting looks fine in the light version, acceptable in the dark version but doesn't look good in the darksilver version:

* light
image

* darksilver
image

Any pointers to what should be tweaked?

GitUI/Theming/Renderers/TabRenderer.cs
NikolayXHD: Here we use on of SystemBrushes, since we do not create it, we should not dispose.

mdonatas:
before
tabs-not-flickering

after
tabs-flickering

This change has caused a very undesirable flickering. Tested at 200% scaling

Issue #8711, opened 2021-01-01
No padding for maximized window


mstv:
## Current behaviour

#8557 has added padding around all controls of FormBrowse. So the maximized window looks like it was resizeable and it does not use the full screen.

## Expected behaviour

No padding if the window is maximized.

RussKie:
Can you please provide screenshots?

ivangrek:
What padings are you talking about?

mstv:
Looks like the window borders added by former Windows versions, which could be dragged for resizing.
This makes me think, the window was not maximized because the controls of maximized applications use the full client area. Perhaps there are some exotic applications. but GE shouldn't be one of them.

![current FormBrowse](https://user-images.githubusercontent.com/36601201/103467264-ad491300-4d4d-11eb-8137-c3c28c3c6194.png)
(I know that this window is not maximized. It looks the same.)

ivangrek:
Like VS, SSMS and many other Win Forms application.

![image](https://user-images.githubusercontent.com/3169265/103467988-63682900-4d5d-11eb-8a40-eaaba36ce1cc.png)

I also suggest moving the tabs to the bottom to eliminate the void in the center.

![image](https://user-images.githubusercontent.com/3169265/103468033-d96c9000-4d5d-11eb-9259-054aea48cfc3.png)

RussKie:
I wouldn't flip, it becomes an accessibility issue - these tabs are high
traffic elements that would require significantly more mouse move,
especially on hdpi monitors.

mstv:
VS is special. It adds space between the client area and the "side tabs" and the status bar.
But it has no space below the status bar as VS Code, as the Windows Explorer, etc.

![VS](https://user-images.githubusercontent.com/36601201/103487019-6b2dd900-4e02-11eb-9411-61c67ea9696e.png) ![VSC](https://user-images.githubusercontent.com/36601201/103487236-2c008780-4e04-11eb-953b-faecf82c2a82.png)

ivangrek:
> But it has no space below the status bar as VS Code, as the Windows Explorer, etc.

I agree.
We don't have that either.

Indentation uniquely identifies work areas. And this design only emphasizes this.

If you notice, VC uses a different way of highlighting work areas with color, with indents inside the colored area. There are also zero margins between panels. All this rests on the complexity of easily changing the design of elements.

If you have any other idea about the desired design, please provide an example.

[...]

PR #8819, opened 2021-02-04
Correct padding increase

### LINKED ISSUE #8698: Incorrect borders when app run in >100% scaling factor

RussKie:
#8557 introduced new margins and borders, however those margins and borders are not correctly calculated in scaled environments, i.e. scale factor >100%.

@ivangrek in #8732 has provided probably a more "correct" fix, however this is a smaller and more targeted fix.

Resolves #8698
Closes #8732

### Before

image

### After

image

## Test methodology 

- manual

image

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -279,6 +279,7 @@ namespace GitUI.CommandsDialogs
             ManageWorktreeSupport();
 
             WorkaroundToolbarLocationBug();
+            WorkaroundPaddingIncreaseBug();
 
             var toolBackColor = SystemColors.Window;
             var toolForeColor = SystemColors.WindowText;
@@ -530,6 +531,13 @@ namespace GitUI.CommandsDialogs
                     Debug.Assert(toolStrips[i].Left < toolStrips[i - 1].Left, $"{toolStrips[i - 1].Name} must be placed before {toolStrips[i].Name}");
                 }
 #endif
+            }
+
+            void WorkaroundPaddingIncreaseBug()
+            {
+                MainSplitContainer.Panel1.Padding = new Padding(1);
+                RevisionsSplitContainer.Panel1.Padding = new Padding(1);
+                RevisionsSplitContainer.Panel2.Padding = new Padding(1);
             }
         }
 


ivangrek:
Why does it work? 😄

RussKie:
> Why does it work? 😄

As described in https://github.com/gitextensions/gitextensions/issues/8698#issuecomment-753476283 something in the layout engine triggers multiple padding scaling:

> There's definitely something in our code that scales paddings 4(!) times: 1 -> 2 -> 3 -> 4 -> 6.
> 
> ```cs
> this.MainSplitContainer.Panel1.ResumeLayout(false); // 1 -> 2px
> ...
> this.MainSplitContainer.ResumeLayout(false); // 2 -> 3px
> ...
> this.toolPanel.ResumeLayout(false); // 3 -> 4px
> ...
> this.ResumeLayout(false); // 4 -> 6px
> ```
> 

In this change I force the padding to 1 after all the layout is done, thus making it expected.

gerhardol (commented):
+1
I would prefer #8732 but can live with this fix

RussKie:
I'd like to reduce amount of changes until after we migrate to .NET 5. I expect some thing may actually change, and some things may require an update.

RussKie:
We can revisit the implementation after the migration.

mdonatas:
image

this seems like the popular case of "fix one, break another" :)
there's now fat padding around main controls

ivangrek:
> there's now fat padding around main controls

This is conceived, but I agree, we can reduce by 2 pixels to 8 pixels like in VS.

mdonatas:
to 8px? why not 0px? I'm running an older build where IMO it looks the best.
Even in VS you don't get a thick padding on all sides and what's more important in VS say Solution Explorer is docked (you can drag it out) so you kinda need to show that visually. I've just checked Slack, Notepad++, Windows Terminal, Windows Explorer and in there controls go to the sides of the window - no wasted space.

image

Also in VS when you don't dock, there is no wasteful padding
image

ivangrek:
image

ivangrek:
> to 8px? why not 0px? I'm running an older build where IMO it looks the best.

In this case, it is necessary to delete everywhere. And use one line as separator and use color for selection.

mstv:
(This issue deals with the DPI scaling.)

We have started a very similar discussion in #8711.

MSVS vertical tab
There is no additional border left to the vertical tabs in MSVS. The bars to the left belong to the tabs (maybe in analogy to the Windows task bar).
taskbar
Perhaps MSVS is somewhat too special to be taken as _the_ example UI.

PR #8833, opened 2021-02-08
Jenkins: requery new running jobs

### LINKED ISSUE #8829: Jenkins build: No queries when builds have finished
## Current behaviour

Regression in #8716 
(some discussions there)

Jenkins builds are updated when first opening and while a build is running. After that, no new builds are displayed.
You may refresh to trigger the status again.

Jenkins and Appveyor work in the same way, they always query all builds (Jenkins has a check for the latest build to find if a full query is needed).
However, AppVeyor only loads builds when refreshing, so the behavior is unchanged.
TeamCity etc can query for builds since a current time so they are not affected.

## Expected behaviour

New builds are updated in the grid

## Steps to reproduce

* Open a repo with Jenkins. 
* Open and see status of builds
* (patiently for) new builds will only be displayed at refresh

Few public Jenkins repos where this can be tested (especially with pipelines). One public where status can be seen (but not updated https://jenkins.mono-project.com/

Before #8716

## Environment

- Git Extensions 33.33.33
- Build db4a7087e673733daa25e1026f7b5cab603a74d1
- Git 2.30.0.windows.2
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8829

## Proposed changes

Jenkins was not requeried since a1fb6669dc20b10d6b65fffebeea9b532e60be5a
This does not seem to be a problem for other buildservers,
AppVeyor requires a manual refresh anyway.

This is not an issue if the build starts immediately after a push, but there could be a queue to start the build, then refresh is needed.

## Test methodology 

* Create a new commit, refresh grid but do not push it
* push the commit from command line
* check that the build status is updated

----

### DIFF
--- a/GitUI/BuildServerIntegration/BuildServerWatcher.cs
+++ b/GitUI/BuildServerIntegration/BuildServerWatcher.cs
@@ -111,8 +111,8 @@ namespace GitUI.BuildServerIntegration
                                                         shouldLookForNewlyFinishedBuilds = true;
                                                     })
                                                .OnErrorResumeNext(delayObservable)
-                                               .Finally(() => anyRunningBuilds = false)
                                                .Retry()
+                                               .Finally(() => anyRunningBuilds = false)
                                                .Repeat()
                                                .ObserveOn(MainThreadScheduler.Instance)
                                                .Subscribe(OnBuildInfoUpdate)


RussKie (commented):
:shipit:

RussKie (approved):
If it works - :shipit:

gerhardol:
@msftbot merge in 1 day

PR #8836, opened 2021-02-09
Jira plugin configuration - Password renamed to Password/API token

### LINKED ISSUE #8825: Jira Plugin-Auth
## Feature description

Jira plugin uses user name and password to authenticate.  Jira supports api tokens.  We should switch to using that.  My work jira account authorizes through a Microsoft account and not user name/password.  

image

## Environment

- Git Extensions 3.4.3.9999
- Git 2.30.0.windows.2
- .NET Framework 4.8.4311.0
- DPI 96dpi (no scaling)

matyasbach:
Fixes #8825.

## Proposed changes
API token is required instead of user password for Jira cloud and possibly for newer versions of Jira on-premise versions.
This PR only changes label from _Password_ to _Password/API token_ to make it clearer that token can (or must) be filled there.

### Before
image

### After
image

----

### DIFF
--- a/Plugins/GitUIPluginInterfaces/UserControls/CredentialsControl.Designer.cs
+++ b/Plugins/GitUIPluginInterfaces/UserControls/CredentialsControl.Designer.cs
@@ -98,7 +98,7 @@
             this.passwordLabel.Name = "passwordLabel";
             this.passwordLabel.Size = new System.Drawing.Size(53, 17);
             this.passwordLabel.TabIndex = 1;
-            this.passwordLabel.Text = "Password";
+            this.passwordLabel.Text = "Password/API token";
             // 
             // CredentialsControl
             // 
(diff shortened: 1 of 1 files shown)


pmiossec:
> API token is required instead of user password for Jira cloud and possibly for newer versions of Jira on-premise versions.

So perhaps "API token/Password" would be better, no?

matyasbach:
> > API token is required instead of user password for Jira cloud and possibly for newer versions of Jira on-premise versions.
> 
> So perhaps "API token/Password" would be better, no?

Agreed and done:
image

pmiossec (approved):

gerhardol:
I would have preferred an info icon (see settings) that explained this instead, text like this make translated GUI a mess.
But this can pass

RussKie:
Is this not localised?

gerhardol:
> Is this not localised?

That is the problem: You may get a very long label in between two textboxes

PR #8838, opened 2021-02-09
Remove /en/latest from Doc links

gerhardol:
## Proposed changes

Requests are directed to / anyway

See discussion in https://github.com/gitextensions/GitExtensionsDoc/issues/131

PR will be updated when submodule is merged

## Test methodology 

review

----

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "GitExtensionsDoc"]
-	path = GitExtensionsDoc
-	url = ../../gitextensions/GitExtensionsDoc.git
 [submodule "Externals/Git.hub"]
 	path = Externals/Git.hub
 	url = ../../gitextensions/Git.hub.git
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/AppearanceSettingsPage.cs
@@ -13,8 +13,8 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
 {
     public partial class AppearanceSettingsPage : SettingsPageWithHeader
     {
-        private const string _customAvatarTemplateURL = "https://git-extensions-documentation.readthedocs.io/en/latest/settings.html#git-extensions-appearance-author-images-avatar-provider";
-        private const string _gravatarDefaultImageURL = "https://git-extensions-documentation.readthedocs.io/en/latest/settings.html#git-extensions-appearance-author-images-avatar-fallback";
+        private const string _customAvatarTemplateURL = "https://git-extensions-documentation.readthedocs.io/settings.html#git-extensions-appearance-author-images-avatar-provider";
+        private const string _gravatarDefaultImageURL = "https://git-extensions-documentation.readthedocs.io/settings.html#git-extensions-appearance-author-images-avatar-fallback";
         private const string _spellingWikiURL = "https://github.com/gitextensions/gitextensions/wiki/Spelling";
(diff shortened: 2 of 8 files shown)


gerhardol:
GitExtensionsDoc is a submodule without any real dependency to the app, just to make sure it is updated.
On the other hand the GitHub Pages site http://gitextensions.github.io/ is not a submodule and in an even poorer state than the doc.

Should we keep the GitHub pages as is (or redirect the repo README)?
Should the sites be treated the same and both be submodules or both be removed?

RussKie (changes requested):
❗ This will break the release script
https://github.com/gitextensions/gitextensions/blob/af0132b3cb31927d3a47613bfde93bdae9775c21/scripts/set_version_to.py#L38-L45

gerhardol:
> ❗ This will break the release script
> https://github.com/gitextensions/gitextensions/blob/af0132b3cb31927d3a47613bfde93bdae9775c21/scripts/set_version_to.py#L38-L45

What link do we want here?
Should we update the buildscript instead?
(it is maybe better to point to the correct version, but is the complexity worth it?)

RussKie:
I'm open to suggestions.

I don't quite understand the vision with the docs atm. Is the intent to
have the dev version to point to the "branchless" URL (e.g.
http://gitextensions.github.io/), and the release branches to point to
http://gitextensions.github.io/[branch]?

Should the dev version point to http://gitextensions.github.io/dev and the
latest official to http://gitextensions.github.io/latest?

How do you see it?

gerhardol:
> 
> I'm open to suggestions. I don't quite understand the vision with the docs atm. Is the intent to have the dev version to point to the "branchless" URL (e.g. http://gitextensions.github.io/), and the release branches to point to [http://gitextensions.github.io/[branch]](http://gitextensions.github.io/%5Bbranch%5D)? Should the dev version point to http://gitextensions.github.io/dev and the latest official to http://gitextensions.github.io/latest? How do you see it?

Github Pages at gitextensions.github.io only have one version and should describe the latest released version.
The question here is if that should be added as a submodule or GitExtensionsDoc be removed?

For the doc the application could point to a specific version as several exist in parallel.
But that of course assumes that we retain the old versions.
I believe we discussed that the dev version should point to "latest" version (which should have been set to master), but there are so few changes unfortunately that it is hardly any point.

My proposal:
* Remove submodule for doc, add links to thepos in the main readme (together  with GitHub Pages).
* point everything to / to simplify
* keep old versions (as now) with a few latest versions and master not as hidden (as now)

The submodules are not approved

RussKie:
> My proposal:
> 
> * Remove submodule for doc, add links to thepos in the main readme (together  with GitHub Pages).
> * point everything to / to simplify
> * keep old versions (as now) with a few latest versions and master not as hidden (as now)
> 
> The submodules are not approved

Sounds good 👍 We should do it before v3.5 is out. 

The last question - would it work for the released versions currently in use (2.x, 3.x)?

gerhardol:
> The last question - would it work for the released versions currently in use (2.x, 3.x)?

As long as we keep the builds for the existing releases that point to something the links will work. `latest` was accidentally removed.

... I cannot believe I suggest to remove the module that is good at changing when testing submodules, now I will have to use dummy commits in the code submodules and add remotes to file locations. But removing still makes sense...

RussKie:
> .. I cannot believe I suggest to remove the module that is good at changing when testing submodules, now I will have to use dummy commits in the code submodules and add remotes to file locations. But removing still makes sense...

🤣 
We can keep it, if you want. Makes no difference to me.

gerhardol:
rebased and changed as suggested
readme seen here (could separate the website and Git repo link some more?)

RussKie (approved):

PR #8841, opened 2021-02-10
Ugly hottrack color of tab header in dark theme

NikolayXHD:
Fix ugly hottrack color of tab header in dark theme introduced in #8815

https://github.com/gitextensions/gitextensions/pull/8815#issuecomment-773218652

### Before

image

### After

image

## Test environment(s)

Windows10

----

### DIFF
--- a/GitExtUtils/GitUI/ControlUtil.cs
+++ b/GitExtUtils/GitUI/ControlUtil.cs
@@ -8,9 +8,6 @@ namespace GitUI
 {
     public static class ControlUtil
     {
-        private static readonly MethodInfo SetStyleMethod = typeof(TabControl)
-            .GetMethod("SetStyle", BindingFlags.Instance | BindingFlags.NonPublic);
-
         /// <summary>
         /// Enumerates all descendant controls.
         /// </summary>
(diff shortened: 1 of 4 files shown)


RussKie:
@mdonatas could you give this one a try? I can see a little flickering (before and after) but not anywhere as bad as you posted.

mdonatas:
This is on a laptop with Intel integrated graphics with an external 4K monitor connected to it.
It's worse on a 4K screen but FHD is still barely acceptable. And I use GE 99% of the time on a 4K screen.

On FHD screen:
on-fhd-screen

On 4K screen:
on-4k-screen

mdonatas:
It seems that flickering is worse when a cursor comes in/out from the bottom/top rather than left/right

NikolayXHD:
I tried a bunch of things to see if there is something specific triggering the flicker

- painting between tabs
- force graphics.clip
- different graphics quality
- ...

Nothing helps.

Currently we are left with 2 options
- tolerate the flicker
- go back to my previous code which was rendering tab control manually and fix high DPI issues there.

When the control is painted manually, `SetStyle(DoubleBuffer)` can be used to prevent flickering.

RussKie:
Your previous solution was stable, maybe we can fix the rendering issues in
it?

NikolayXHD:
ok

PR #8844, opened 2021-02-10
Experimental flipping of branch colors

ivangrek:
## Proposed changes

- Ref to #8767
- Green indicates the addition of a new one and the safety of deletion and vice versa.
- ~Add experimental settings for controversial cases to speed up implementation.~

### Before
image

image

### After

image

image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build 2e5075edb1bd67211307cb9334f55056bde7140d
- Git 2.30.0.windows.2
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----


gerhardol:
What about adding a new theme instead?

ivangrek:
> What about adding a new theme instead?

I thought about it. I see this as a fix, but with a soft transition. New thema touches on more aspects.

RussKie:
I don't mind the idea of experimental settings (it crossed my mind few times over the years), but we need to have a discussion about logistics and lifetime of such settings. 

So far our release cadence was pretty slow - 2-3 releases a year, and there is generally a sufficient long runway for a feature to bake before it gets released. If we released more often - then experimental features would totally make sense. 

* How long would you propose to keep a setting as experimental? 
* How do we detect whether a feature is success or a total flop?

With regards to the settings placement:

* Do we expect to put all experimental settings on the same page? If so the settings should be grouped by type - e.g. "Appearance / Colors", "Performance", etc.
* If we have more settings than fit on a page, what do we do?

ivangrek:
> Do we expect to put all experimental settings on the same page? If so the settings should be grouped by type - e.g. "Appearance / Colors", "Performance", etc.

On this page, only on/off, everything else will be where it should be in the case of use.

> If we have more settings than fit on a page, what do we do?

Add scrolling.

ivangrek:
>How long would you propose to keep a setting as experimental?
How do we detect whether a feature is success or a total flop?

Don't know. Maybe telemetry. 
I would like to be less stuck as we can discuss endlessly. 😄

gerhardol:
I prefer experimental settings related to what they test.
Also, we already have Advanced settings (the page should probably be under Advanced regardless).
So for switching colors for branches have the setting at Colors, collapsed by default, like advanced settings are handled in many forms.
(For branch colors I prefer a new experimental theme instead, but that is another matter.)

But no strong preference, I can live with this.

ivangrek:
> Also, we already have Advanced settings (the page should probably be under Advanced regardless).

This is only on/off. After enabling, everything appears in those sections where it should be.

RussKie:
Is there anything stopping us from fixing the themes we have? 
The current colour selection looks like a genuine bug, which wasn't noticed for some time (years). Still it looks inconsistent to how git does it.

ivangrek:
> Is there anything stopping us from fixing the themes we have? I'd rather fix it once and for all instead of introducing and maintaining extra settings.

The current colour selection looks like a genuine bug, which wasn't noticed for some time (years). Still it looks inconsistent to how git does it.

👍

gerhardol:
> Is there anything stopping us from fixing the themes we have? I'd rather fix it once and for all instead of introducing and maintaining extra settings.

My point was that if we change the theme, we could have a legacy theme for those that want the old behavior back.

RussKie:
...or tell users how to make a local theme ;)

NikolayXHD (approved):

RussKie (approved):

PR #8846, opened 2021-02-11
Annotate GitExtensions and GitUI projects

drewnoakes:
Follows #8776.

Adds nullable reference type annotations to the `GitExtensions` and `GitUI` projects, and extends some annotations in previously annotated code based on usage observations made while annotating these two projects.

General principles applied in this work:

1. Annotate what actually happens, not what makes sense.
   While making these annotations there are places where `null` appears to flow by accident rather than by design. Such recipients of references would likely have been marked as non-null had this language feature existed at the time. I have intentionally not updated many of these cases, as the PR is already large enough, and that strays into behaviour change territory. They can be addressed in future.
2. Use `Validates.NotNull` in places where allowing a `null` to propagate would result in an exception.
    This validation method is annotated such that Roslyn understands a value that may be null before the call will not be null after the call (because a null would result in an exception). To avoid behaviour changes, I have attempted to only add these throwing helper methods in places where an exception would definitely be raised further down the line.
3. Make minor refactorings where it helps.
    For example `GitItemStatus` only had a default constructor, and also has a property `Name` which was always set to a non-null value via object initializer. Unfortunately Roslyn does not currently support required init properties (perhaps in C# 10). Without a guarantee that the value is assigned during construction, the property must be marked as nullable. However in practice, the value is never null. The simple fix was to add a constructor that requires a non-null `Name` value and assign the property, after which it may be marked as non-null. This is much simpler than updating consuming code to defend against a null value that would never be observed in practice.

Notes:

- In the first annotation PR I introduced a `Strings` class which includes a properly annotated version of `string.IsNullOr*` methods (as the .NET Framework versions are not annotated). That worked well in other projects, however the GitUI project already has a class called `Strings`, containing resource file strings. Rather than rename either `Strings` class, I used fully qualified names. Perhaps in a follow up PR we can sort this out. That should be easier to review.
- There are bug fixes in this PR. I'll go through and comment on these as I see them.
- I've been using the code in this PR all day and haven't hit any bugs, but of course I haven't used every single feature in the app.

Feel free to hit me with any questions about the PR and/or language feature.


GitCommands/Git/GitItemStatus.cs
```diff
@@ -46,7 +47,13 @@ private enum Flags
...
         private Flags _flags;
 
-        public string? Name { get; set; }
+        public GitItemStatus(string name)
+        {
+            Requires.NotNull(name, nameof(name));
+            Name = name;
+        }
+
+        public string Name { get; set; }
```
drewnoakes, line 56: Here is the example I called out in the PR description, of introducing a constructor to avoid marking a property as nullable.

GitCommands/Git/GitItemStatus.cs
```diff
@@ -170,9 +177,14 @@ private void SetFlag(bool isSet, Flags flag)
...
-        public Task<GitSubmoduleStatus?>? GetSubmoduleStatusAsync()
+        public async Task<GitSubmoduleStatus?> GetSubmoduleStatusAsync()
         {
-            return _submoduleStatus?.JoinAsync();
+            if (_submoduleStatus is null)
+            {
+                return null;
+            }
+
+            return await _submoduleStatus.JoinAsync();
```
drewnoakes, line 187: Here's a bug that this PR fixes. Previously this method would have returned a null _task_ when `_submoduleStatus` was null. Trying to await a null task results in an exception. The consuming code expected a task containing a null _value_ instead, which the new code now provides.

GitCommands/Git/GitModule.cs
```diff
@@ -980,15 +980,15 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
             revision.HasNotes = !shortFormat;
             if (shortFormat)
             {
-                revision.Subject = ReEncodeCommitMessage(lines[10], revision.MessageEncoding);
+                revision.Subject = ReEncodeCommitMessage(lines[10], revision.MessageEncoding) ?? "";
```
drewnoakes, line 983: In a few places, null is substituted for a default value. In this case, the empty string.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1910,7 +1910,7 @@ public static Color GetColor(AppColor name)
             return SettingsContainer.GetColor(name.ToString().ToLowerInvariant() + "color", AppColorDefaults.GetBy(name));
         }
 
-        public static void SetEnum<T>(string name, T value)
+        public static void SetEnum<T>(string name, T value) where T : Enum
```
drewnoakes, line 1913: Added an `Enum` type constraint here, which tells the analysis that values of `T` can never be null.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1935,7 +1935,8 @@ public static void SetString(string name, string value)
             SettingsContainer.SetString(name, value);
         }
 
-        public static string GetString(string name, string? defaultValue)
+        [return: NotNullIfNotNull("defaultValue")]
+        public static string? GetString(string name, string? defaultValue)
```
drewnoakes, line 1939: This annotation means that the return value is only null when `defaultValue` itself is null. This avoids a bunch of errors in consuming code.

GitCommands/Submodules/SubmoduleInfo.cs
```diff
@@ -8,14 +8,22 @@ public class SubmoduleInfo
...
 
         public DetailedSubmoduleInfo? Detailed { get; set; }
-        public bool Bold { get; set; }
+
+        public SubmoduleInfo(string text, string path, bool bold)
+        {
+            Text = text;
+            Path = path;
+            Bold = bold;
+        }
```
drewnoakes, line 27: Another constructor to prevent having to make properties nullable.

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -91,11 +91,10 @@ public async Task UpdateSubmodulesStructureAsync(string workingDirectory, string
             var result = GetSuperProjectRepositorySubmodulesStructure(currentModule, noBranchText);
 
             // Prepare info for status updates
-            Assumes.NotNull(result.TopProject?.Path);
+            Assumes.NotNull(result.TopProject);
```
drewnoakes, line 94: `Path` can never be null, so remove it. `TopProject` can though.

GitExtUtils/ArgumentString.cs
```diff
@@ -15,9 +15,9 @@ private ArgumentString(string arguments)
             Arguments = arguments ?? throw new ArgumentNullException(nameof(arguments));
         }
 
-        public static implicit operator ArgumentString(string args) => new ArgumentString(args);
+        public static implicit operator ArgumentString(string? args) => new ArgumentString(args ?? "");
         public static implicit operator ArgumentString(ArgumentBuilder args) => new ArgumentString(args.ToString());
-        public static implicit operator string?(ArgumentString args) => args.Arguments;
+        public static implicit operator string(ArgumentString args) => args.Arguments ?? "";
         public override string ToString() => Arguments ?? "";
```
drewnoakes, line 21: Remove null values from implicit conversions. The previous annotations introduced errors in consuming code that looked weird. It appeared as though a struct (`ArgumentString`) was somehow nullable.

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
```diff
@@ -73,7 +73,6 @@ public async Task<IEnumerable<AutoCompleteWord>> GetAutoCompleteWordsAsync(Cance
             return autoCompleteWords.Select(w => new AutoCompleteWord(w));
         }
 
-        [NotNull]
```
drewnoakes, line 76: When we're done with NRT annotations, we won't need `[NotNull]`, `[CanBeNull]` or `[ContractAnnotation]` any more.

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
```diff
@@ -137,8 +139,7 @@ private static IEnumerable<string> ReadOrInitializeAutoCompleteRegexes()
             return regexes;
         }
 
-        [CanBeNull]
-        private static async Task<string> GetChangedFileTextAsync(GitModule module, GitItemStatus file)
+        private static async Task<string?> GetChangedFileTextAsync(GitModule module, GitItemStatus file)
```
drewnoakes, line 142: This annotation was previously incorrect. It should have been `ItemCanBeNull`. With NRT the compiler will enforce correctness of annotations.

GitUI/Avatars/AvatarMemoryCache.cs
```diff
@@ -40,12 +39,15 @@ public async Task<Image> GetAvatarAsync(string email, string name, int imageSize
...
-                _cache.Add((email, imageSize), image);
-
-                return image;
+                lock (_cache)
+                {
+                    _cache.Add((email, imageSize), image);
+                }
             }
+
+            return image;
```
drewnoakes, line 50: This fixes a bug where we could attempt to add a null image into the cache, which results in an exception.

GitUI/Avatars/AvatarService.cs
```diff
@@ -55,7 +55,7 @@ public static void UpdateAvatarProvider()
                 BuildMainProvider(),
                 BuildFallbackProvider()
             }
-            .Where(p => p != null)
+            .WhereNotNull()
```
drewnoakes, line 58: Unfortunately, `Enumerable.Where<T?>` with a predicate that guarantees non-null values still returns `IEnumerable<T>`, not `IEnumerable<T?>`. We have a `WhereNotNull` extension method in the library that handles this case, avoiding use of `!` operator.

GitUI/Avatars/AvatarMemoryCache.cs
gerhardol: `if (image is not null)`

GitUI/Avatars/GithubAvatarProvider.cs
```diff
@@ -50,7 +50,7 @@ public async Task<Image> GetAvatarAsync([NotNull] string email, string name, int
             // GitHub are never scaled and always 420 x 420 - even if a different size was requested.
 
             // We exploit that fact to filter out identicons.
-            var isIdenticon = imageSize != 420 && image.Size.Width == 420;
+            var isIdenticon = imageSize != 420 && image != null && image.Size.Width == 420;
```
drewnoakes, line 53: Bug fix. The download can fail and return a null `image` which would have thrown here in the past.

GitUI/BranchTreePanel/ContextMenu/MenuItemsGenerator.cs
```diff
@@ -70,23 +70,23 @@ private IEnumerable<ToolStripItemWithKey> CreateGitRefItems()
             }
 
             yield return _menuItemFactory.CreateMenuItem<ToolStripMenuItem, TNode>(
-                node => (node as IGitRefActions).Checkout(), Strings.Checkout, GetTooltip(MenuItemKey.GitRefCheckout), Properties.Images.BranchCheckout)
+                node => ((IGitRefActions)node).Checkout(), Strings.Checkout, GetTooltip(MenuItemKey.GitRefCheckout), Properties.Images.BranchCheckout)
```
drewnoakes, line 73: Replace NullReferenceException with InvalidCastException.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Branches.cs
```diff
@@ -53,7 +53,7 @@ protected BaseBranchNode(Tree tree, string fullPath, bool visible)
             /// <summary>
             /// Full path of the branch. <example>"issues/issue1344"</example>.
             /// </summary>
-            public string FullPath => ParentPath.Combine(PathSeparator.ToString(), Name);
+            public string FullPath => ParentPath.Combine(PathSeparator.ToString(), Name)!;
```
drewnoakes, line 56: This `Combine` extension method (it is not `System.IO.Path.Combine`) has pretty weird behaviour and cannot be annotated. It should probably be replaced with something less surprising and easier to annotate. I've suppressed the warning here with `!`.

GitUI/Avatars/AvatarMemoryCache.cs
drewnoakes: @gerhardol updated.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -342,7 +350,7 @@ private void CreateSubmoduleNodes(SubmoduleInfoResult result, GitModule threadMo
                 string GetSubmoduleSuperPath(string submodulePath)
                 {
                     var superPath = modulePaths.Find(path => submodulePath != path && submodulePath.Contains(path));
-                    Trace.Assert(superPath is not null);
+                    Assumes.NotNull(superPath);
```
drewnoakes, line 353: Various 'soft' assertions (i.e. showing debug assertion messages, only in debug builds) have been changed into hard throws in release builds.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Tags.cs
```diff
@@ -138,7 +136,7 @@ private Nodes FillTagTree(IReadOnlyList<IGitRef> tags, CancellationToken token)
                 {
                     token.ThrowIfCancellationRequested();
 
-                    bool isVisible = !IsFiltering.Value || _refsSource.Contains(tag.ObjectId);
+                    bool isVisible = !IsFiltering.Value || (tag.ObjectId is not null && _refsSource.Contains(tag.ObjectId));
```
drewnoakes, line 139: Bug fix: If the `ObjectId` was null, it would throw when passed to the `Contains` method (null key exception).

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -153,7 +153,11 @@ protected Tree(TreeNode treeNode, IGitUICommandsSource uiCommands)
                     _firstReloadNodesSinceModuleChanged = true;
 
                     // Rebind callbacks
-                    e.OldCommands.PostRepositoryChanged -= UICommands_PostRepositoryChanged;
+                    if (e.OldCommands is not null)
+                    {
+                        e.OldCommands.PostRepositoryChanged -= UICommands_PostRepositoryChanged;
+                    }
```
drewnoakes, line 159: Bug fix.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -366,7 +367,16 @@ static void EnsureNodeVisible(IntPtr hwnd, TreeNode node)
...
-            protected Tree Tree => Nodes.Tree;
+
+            protected Tree Tree
+            {
+                get
+                {
+                    Assumes.NotNull(Nodes.Tree);
+                    return Nodes.Tree;
+                }
+            }
```
drewnoakes, line 378: One usage needs to set a null `Tree` property, however we never want to allow reading a null from here. Changed to allow null in the ctor but throw if it's read.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -161,8 +162,8 @@ SearchControl<string> CreateSearchBox()
                 };
                 search.OnTextEntered += () =>
                 {
-                    OnBranchCriterionChanged(null, null);
-                    OnBtnSearchClicked(null, null);
+                    OnBranchCriterionChanged(this, EventArgs.Empty);
+                    OnBtnSearchClicked(this, EventArgs.Empty);
```
drewnoakes, line 166: This happens a fair bit in the PR.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -621,6 +622,11 @@ private void listView1_GroupMouseUp(object sender, ListViewGroupMouseEventArgs e
             {
                 var groupHitInfo = listView1.GetGroupHitInfo(e.Location);
 
+                if (groupHitInfo is null)
+                {
+                    return;
+                }
+
                 bool isRecentRepositoriesGroup = groupHitInfo.Group == _lvgRecentRepositories;
```
drewnoakes, line 630: Bug fix. `groupHitInfo` can be null.

GitUI/CommandsDialogs/BrowseDialog/FormBisect.cs
```diff
@@ -18,22 +18,19 @@ public sealed partial class FormBisect : GitModuleForm
         private readonly RevisionGridControl _revisionGrid;
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         private FormBisect()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         {
             InitializeComponent();
         }
```
drewnoakes, line 26: We have several forms/controls where a private constructor is added only for the designer/translator. Because these ctors don't assign all fields with non-null values, we'd usually see CS8618. In each case I suppress the diagnostic.

GitUI/CommandsDialogs/BuildReportTabPageExtension.cs
```diff
@@ -96,9 +98,11 @@ public void FillBuildReport([CanBeNull] GitRevision revision)
 
         private void LoadReportContent(GitRevision revision, bool isFavIconMissing)
         {
+            Assumes.NotNull(_buildReportWebBrowser);
+
             try
             {
-                if (revision.BuildStatus.ShowInBuildReportTab)
+                if (revision.BuildStatus?.ShowInBuildReportTab == true)
```
drewnoakes, line 105: NRE bug fix.

GitUI/CommandsDialogs/FormArchive.cs
```diff
@@ -117,7 +117,7 @@ private void Save_Click(object sender, EventArgs e)
                 return;
             }
 
-            string revision = SelectedRevision.Guid;
+            string? revision = SelectedRevision?.Guid;
```
drewnoakes, line 120: NRE bug fix.

GitUI/CommandsDialogs/FormCherryPick.cs
```diff
@@ -73,7 +71,7 @@ private void OnRevisionChanged()
 
             panelParentsList.Visible = _isMerge;
 
-            if (_isMerge)
+            if (_isMerge && Revision is not null)
```
drewnoakes, line 74: NRE bug fix.

drewnoakes, line 74: NRE bug fix (ditto below).

GitUI/CommandsDialogs/FormClone.cs
```diff
@@ -405,10 +407,16 @@ private void ToSelectedIndexChanged(object sender, EventArgs e)
...
-        private void UpdateBranches(RemoteActionResult<IReadOnlyList<IGitRef>> branchList)
+        private void UpdateBranches(RemoteActionResult<IReadOnlyList<IGitRef>>? branchList)
         {
             Cursor = Cursors.Default;
 
+            if (branchList == null)
+            {
+                // cancelled
+                return;
+            }
```
drewnoakes, line 418: This pattern occurs in several places. This `UpdateBranches` method is the callback passed to an instance of `AsyncLoader` and the argument is almost never null.

`AsyncLoader` has a behaviour where if its load is cancelled, it invokes the callback with a null argument. A future change should review this, and probably remove that behaviour. It results in lots of these checks, which could be avoided.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1902,16 +1929,41 @@ private void Stage(IReadOnlyList<GitItemStatus> items)
...
+                            }
+
+                            GitSubmoduleStatus? gitSubmoduleStatus = ThreadHelper.JoinableTaskFactory.Run(() => item.GetSubmoduleStatusAsync());
+
+                            if (gitSubmoduleStatus is null || !gitSubmoduleStatus.IsDirty)
+                            {
+                                continue;
+                            }
+
+                            gitSubmoduleStatus.Status = SubmoduleStatus.Unknown;
```
drewnoakes, line 1966: There was a NRE bug in here, but I can't see it now. The previous code was pretty hard to unpack. Changed to be more verbose, but clearer. This form also allows the compiler to validate null value flow.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2282,7 +2335,8 @@ private void generateListOfChangesInSubmodulesChangesToolStripMenuItem_Click(obj
                 .ToDictionary(localPath =>
                 {
                     var submodule = configFile.ConfigSections.FirstOrDefault(section => section.GetValue("path").Trim() == localPath);
-                    return submodule?.SubSection?.Trim();
+                    Assumes.NotNull(submodule?.SubSection);
+                    return submodule.SubSection.Trim();
```
drewnoakes, line 2339: This lambda is returning the key for the dictionary. Null keys are not allowed. Swaps one kind of exception for another, but is more explicit about the issue.

GitUI/CommandsDialogs/FormFileHistory.cs
```diff
@@ -32,17 +32,27 @@ public sealed partial class FormFileHistory : GitModuleForm
...
-        private FormFileHistory([NotNull] GitUICommands commands)
+        /// <summary>
+        /// Open FileHistory form.
+        /// </summary>
+        /// <param name="commands">commands in the current form.</param>
+        /// <param name="fileName">name in repo of file to view.</param>
+        /// <param name="revision">initial selected commit.</param>
+        /// <param name="filterByRevision">add filter.</param>
+        /// <param name="showBlame">show blame initially instead of diff view.</param>
+        public FormFileHistory(GitUICommands commands, string fileName, GitRevision? revision = null, bool filterByRevision = false, bool showBlame = false)
```
drewnoakes, line 55: Merged two constructors here. Only one was used. Allows compiler verification of definite assignment to class fields.

GitUI/CommandsDialogs/FormMailMap.cs
```diff
@@ -55,7 +58,7 @@ private void LoadFile()
                 var path = _fullPathResolver.Resolve(".mailmap");
                 if (File.Exists(path))
                 {
-                    _NO_TRANSLATE_MailMapText.ViewFileAsync(path);
+                    _NO_TRANSLATE_MailMapText.ViewFileAsync(path!);
```
drewnoakes, line 61: `File.Exists` is not annotated in .NET Framework. However if it returns `true`, then `path` is definitely not null. Suppress the warning here. When moving to .NET 5+, this will not be necessary.

Plugins/GitUIPluginInterfaces/INamedGitItem.cs
```diff
@@ -0,0 +1,9 @@
+﻿#nullable enable
+
+namespace GitUIPluginInterfaces
+{
+    public interface INamedGitItem : IGitItem
```
drewnoakes, line 5: @RussKie I moved this to its own type as requested in last PR.

drewnoakes:
> Merging [#8846](https://codecov.io/gh/gitextensions/gitextensions/pull/8846?src=pr&el=desc) ([bc58fba](https://github.com/gitextensions/gitextensions/commit/bc58fba894f9f378ad75f1d0e7bc5051d9bbff3e)) into [master](https://codecov.io/gh/gitextensions/gitextensions/commit/18616edc8b12d048d3d54c9f41c93a4c9ffb03ea?el=desc) ([18616ed](https://github.com/gitextensions/gitextensions/commit/18616edc8b12d048d3d54c9f41c93a4c9ffb03ea)) will **decrease** coverage by `28.66%`.

I find this very suspect. My guess is that codecov doesn't know how to deal with nullable reference types properly.

gerhardol (commented):
I have browsed about a third of the files, as well as some files I have worked in.
Closer looks at changes other than just annotation.
With the help of the explanations I have few comments.
Will browse the remaining files too, do not expect to find anything...

GE3.5 should go out before this though, too many changes...
Should the found bugs be applied first to master (or the release/3.5 branch?

Will use this myself.

GitUI/UserControls/FileStatusItem.cs
```diff
@@ -19,34 +17,29 @@ public FileStatusItem([CanBeNull] GitRevision firstRev, [NotNull] GitRevision se
 
         /// <summary>
         /// First (Parent or A in diff)
-        /// Can be null for the initial commit
+        /// Can be null for the initial commit.
         /// </summary>
-        [CanBeNull]
-        public GitRevision FirstRevision { get; }
+        public GitRevision? FirstRevision { get; }
```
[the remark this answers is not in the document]
drewnoakes: A sentinel value here seems like a good idea. My sense is to do this in a follow-up PR. Feel free to push to this PR too btw.

drewnoakes:
Thanks for taking a look @gerhardol. I know it's a daunting PR to review.

> GE3.5 should go out before this though, too many changes...
> Should the found bugs be applied first to master (or the release/3.5 branch?

Makes sense. I see there's a release/3.5 branch. Is master not separate from 3.5, or is the plan to merge master into it in future?

GitCommands/PathUtil.cs
```diff
@@ -78,19 +78,19 @@ public static bool IsLocalFile(string fileName)
...
                  || path.StartsWith("git:", StringComparison.CurrentCultureIgnoreCase)
                  || path.StartsWith("ssh:", StringComparison.CurrentCultureIgnoreCase)
                  || path.StartsWith("file:", StringComparison.CurrentCultureIgnoreCase));
         }
 
-        public static bool CanBeGitURL(string url)
+        public static bool CanBeGitURL(string? url)
         {
-            if (string.IsNullOrWhiteSpace(url))
+            if (Strings.IsNullOrWhiteSpace(url))
```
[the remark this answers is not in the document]
drewnoakes: Thanks, pushed a commit to address all instances of `== null` and `!= null` in the core source.

It makes no difference to null analysis. It's mostly style. Academically, it can make a difference if a type overrides the `==` operator to do something suprising, which I've never seen in practice.

I don't mind the style change, but there's a lot of muscle memory here. Ideally we find an analyzer to enforce this to reduce friction.

GitUI/UserControls/FileStatusItem.cs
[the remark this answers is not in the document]
drewnoakes: `RememberFileContextMenuController.CreateFileStatusItem` also passes `null` for `firstRevision`.

drewnoakes:
@RussKie is master ready for work after the 3.5 release?

RussKie:
The master is open, I cherry pick changes into the 3.5 as necessary.

Though any changes that come after this change will likely require
modifications.

RussKie:
@gerhardol has the difftools change that we'll need in the 3.5, it has a
large surface.
Not expecting anything substantial besides that.

drewnoakes:
Thanks @RussKie. Let me know if I can do anything else to help get this merged. I have another smaller PR with more annotations ready to go after this.

gerhardol:
If the findings should go into 3.5, I suggest to separate those changes to a PR to master first.
There are not so many changes to master yet (I will create a few PRs).

gerhardol:
History does not show any differences
```
System.InvalidOperationException: GitRevision does not have a Name.
   at GitUIPluginInterfaces.GitRevision.get_Name()
   at GitUI.CommandsDialogs.FormFileHistory.UpdateSelectedFileViewers(Boolean force)
   at GitUI.CommandsDialogs.FormFileHistory.FileChangesSelectionChanged(Object sender, EventArgs e)
   at GitUI.RevisionGridControl.<.ctor>b__123_0(Object _, EventArgs e)
   at System.Windows.Forms.Timer.OnTick(EventArgs e)
   at System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
```

```
- Git Extensions 33.33.33
- Build 6bf9eea2045c2d6971478ff5b6b65dc09737c9ff
- Git 2.30.1.windows.1
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)
```

GitUI/CommandsDialogs/FormStash.cs
```diff
@@ -208,10 +213,11 @@ private void LoadGitItemStatuses(IReadOnlyList<GitItemStatus> gitItemStatuses)
                 var firstId = Module.RevParse(gitStash.Name + "^");
                 var selectedId = Module.RevParse(gitStash.Name);
                 var firstRev = firstId is null ? null : new GitRevision(firstId);
-                var secondRev = selectedId is null ? null : new GitRevision(selectedId)
+                var secondRev = selectedId is null || firstId is null ? null : new GitRevision(selectedId)
                 {
                     ParentIds = new[] { firstId }
                 };
+                Assumes.NotNull(secondRev);
```
gerhardol, line 220: this looks strange

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -535,14 +537,14 @@ private void FileTreeContextMenu_Opening(object sender, System.ComponentModel.Ca
...
             var itemSelected = gitItem is not null;
-            var isFile = itemSelected && gitItem.ObjectType == GitObjectType.Blob;
-            var isFolder = itemSelected && gitItem.ObjectType == GitObjectType.Tree;
+            var isFile = gitItem?.ObjectType == GitObjectType.Blob;
+            var isFolder = gitItem?.ObjectType == GitObjectType.Tree;
             var isFileOrFolder = isFile || isFolder;
 
             // Many items does not make sense if a local file does not exist, why this is used for Enabled
-            var isExistingFileOrDirectory = itemSelected && FormBrowseUtil.IsFileOrDirectory(_fullPathResolver.Resolve(gitItem.FileName));
+            var isExistingFileOrDirectory = gitItem is not null && FormBrowseUtil.IsFileOrDirectory(_fullPathResolver.Resolve(gitItem.FileName));
```
gerhardol, line 545: The variable is still available, why change this only?

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -655,7 +662,7 @@ private void openWithToolStripMenuItem_Click(object sender, EventArgs e)
                 var fileName = _fullPathResolver.Resolve(gitItem.FileName);
                 if (File.Exists(fileName))
                 {
-                    OsShellUtil.OpenAs(fileName.ToNativePath());
+                    OsShellUtil.OpenAs(fileName!.ToNativePath());
```
gerhardol, line 665: This should not be needed with .NET5. Comment somehow?
Maybe File.Exists should be wrapped?

gerhardol (commented):
+1
Except for some minor comments and FileHistory, all looks fine to me.
I would like RussKie to sign off, he is oon a break for some days though. I hope he do something more productive than reviewing every file in this PR though...

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
drewnoakes: Roslyn doesn't (currently) track the fact that the bool `itemSelected` implies that `gitItem` is not null.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
drewnoakes: I've added a comment. Perhaps all file system access should occur through an interface for testability (and diagnostics, potentially), and that interface could be correctly annotated. Probably a follow up action however.

There's also a `Microsoft.IO` package which replaces several `System.IO` operations with better performance (especially around path operations). The implementation of that interface could be implemented in terms of that, in places.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -459,6 +456,7 @@ private static (Image image, string prefix, string text, string suffix, int pref
             var prefixTextStartX = itemLeft + (image?.Width ?? 0);
             var textMaxWidth = itemWidth - prefixTextStartX;
             var (prefix, text, suffix, textWidth) = formatter.FormatTextForDrawing(textMaxWidth, gitItemStatus.Name, gitItemStatus.OldName);
+            text ??= "";
             text = AppendItemSubmoduleStatus(text, gitItemStatus);
```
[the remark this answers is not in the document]
drewnoakes: Done. It will now return a null prefix instead of an empty string. There is no behaviour change.

GitUI/UserControls/NativeTreeViewDoubleClickDecorator.cs
```diff
@@ -65,19 +64,13 @@ private void OnMouseDown(object sender, MouseEventArgs e)
 
             if (_mouseClicks == 2)
             {
-                int delta = (int)now.Subtract(_lastMouseDown).TotalMilliseconds;
```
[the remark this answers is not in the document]
drewnoakes: The value is unused.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2758,11 +2759,6 @@ Point IScriptHostControl.GetQuickItemSelectorLocation()
 
         #endregion
 
-        /// <summary>
```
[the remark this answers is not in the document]
drewnoakes: It is an exact copy of the base interface's documentation. These days, VS shows the base docs in intellisense, quick info, etc. Duplicating the documentation allows it to go out of date when one level is changed, but not the other.

GitUI/CommandsDialogs/FormStash.cs
```diff
@@ -208,10 +213,11 @@ private void LoadGitItemStatuses(IReadOnlyList<GitItemStatus> gitItemStatuses)
...
                 Stashed.SetDiffs(firstRev, secondRev, gitItemStatuses);
```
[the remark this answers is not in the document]
drewnoakes: Applied and tested.

drewnoakes:
Fixed merge conflicts.

> History does not show any differences...

Fixed in 6faa13d61d5cc69bf511d19c9b66e7594c5c9a7c. I couldn't repro it, but the validation appears unnecessary. I've made the property nullable, as all usages seemed to expect that anyway. That exception will not happen any more.

GitUI/CommandsDialogs/FormStash.cs
```diff
@@ -206,12 +211,17 @@ private void LoadGitItemStatuses(IReadOnlyList<GitItemStatus> gitItemStatuses)
...
+                Assumes.NotNull(selectedId);
+                var secondRev = new GitRevision(selectedId);
+                if (firstId is not null)
                 {
-                    ParentIds = new[] { firstId }
-                };
+                    secondRev.ParentIds = new[] { firstId };
+                }
+
+                Assumes.NotNull(secondRev);
```
gerhardol, line 224: second check...

gerhardol:
In a private build that was OK on one PC but not another, I got an exception at startup (assume it it is me until I can check, it willl take some hours though). The popup need to present these messages better, this is an internal GE issue...

Special handling of Microsoft.Assumes exception?

image

```
Microsoft.Assumes+InternalErrorException: An internal error occurred. Please contact Microsoft Support.
   at Microsoft.Assumes.Fail(String message)
   at Microsoft.Assumes.NotNull[T](T value)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__CreateLink|15_7(Control container, String text, Image icon, EventHandler handler)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>b__15_2(Panel panel)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__AddLinks|15_6(Panel panel, Func`2 addLinks, Action`2 onLayout)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__InitDashboardLayout|15_1()
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.RefreshContent()
   at GitUI.CommandsDialogs.FormBrowse.ShowDashboard()
   at GitUI.CommandsDialogs.FormBrowse.InternalInitialize(Boolean hard)
   at GitUI.CommandsDialogs.FormBrowse.OnLoad(EventArgs e)
   at System.Windows.Forms.Form.WndProc(Message& m)
```

RussKie (approved):
Cursory review 👍

RussKie:
Converting to draft until we iron out the kinks.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/Dashboard.cs
```diff
@@ -161,6 +162,8 @@ void AddLinks(Panel panel, Func<Panel, Control> addLinks, Action<Panel, Control>
 
                 Control CreateLink(Control container, string text, Image icon, EventHandler handler)
                 {
+                    Assumes.NotNull(_selectedTheme);
```
gerhardol, line 165: I am submitting a PR especially for this, candidate for 3.5

GitUI/CommandsDialogs/FormStash.cs
drewnoakes: Removed.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/Dashboard.cs
drewnoakes: Replicated in 0240f7e97f634c054a39998552e14ae8f6d5c7fc.

drewnoakes:
@gerhardol thank you for the further testing and bug report.

> The popup need to present these messages better, this is an internal GE issue...

I agree. I had filed the same issue in https://github.com/microsoft/vs-validation/issues/78, so if you have any thoughts please add them there.

In the meantime perhaps we add our own equivalent function for use within GE. Perhaps `Validate.NotNull` etc.

gerhardol:
> > The popup need to present these messages better, this is an internal GE issue...
> 
> I agree. I had filed the same issue in [microsoft/vs-validation#78](https://github.com/microsoft/vs-validation/issues/78), so if you have any thoughts please add them there.

Not more to add than "me too" in the issue.

> In the meantime perhaps we add our own equivalent function for use within GE. Perhaps `Validate.NotNull` etc.

I had hoped that just catching the exception for NBug would be sufficient, if not a separate library should be used.

drewnoakes:
> if not a separate library should be used.

Note that both vs-mef and vs-threading (used by GE) also use vs-validation, so ideally we fix the upstream library.

drewnoakes:
I've introduced `Validates.NotNull` which replaces `Assumes.NotNull` but with a more sensible exception message.

drewnoakes:
> Converting to draft until we iron out the kinks.

Are you talking about stability? Not sure exactly what kinks you mean.

If it is stability (which is fair enough) then I'm a little confused about the branching strategy. It seems like 3.5 has snapped to a release branch, in which case master is open for new work and it's the ideal time for these kinds of changes as they have a long time to bed in before getting near a release.

Note that null validation exceptions tend to be easy to fix, and the fix is obvious in most cases meaning someone who hits one can patch it and get unblocked quickly.

gerhardol:
> > Converting to draft until we iron out the kinks.
> 
> Are you talking about stability? Not sure exactly what kinks you mean.

I guess that was about the dashboard problem and the exception message.
It will be easier to diff what is in master and release/3.5 if there are not too many changes between the branches. Right now it is basically build changes and INamedGitItem that differs so they are quite the same. That is why I submitted #8860 separately and asked for a separation of the actual findings in this PR.

> Note that both vs-mef and vs-threading (used by GE) also use vs-validation, so ideally we fix the upstream library.

This must be fixed from a Microsoft perspective and it is maybe not so bad to have the Microsoft popup for a while. But since the change is done, I suggest it is kept.

(lock was accidental...)

drewnoakes:
Thanks for the explanation @gerhardol.

> It will be easier to diff what is in master and release/3.5 if there are not too many changes between the branches.

I don't understand why it's helpful to be able to diff these.

BTW I'm not advocating any particular way. Just been out of the project for a while and looking to pick up some tasks that I find interesting. It seems like feature development is on hold for 3.5. Perhaps that makes sense for this team. At work we snap the branch and keep working on main on new features without really looking back, so I'm used to that.

gerhardol:
> Thanks for the explanation @gerhardol.
> 
> > It will be easier to diff what is in master and release/3.5 if there are not too many changes between the branches.
> 
> I don't understand why it's helpful to be able to diff these.

To make sure all intended corrections are included in either branch. release/3.5 was created a little to early so there are many changes that are in both branches. If there are changes we know right now that should be included in both, submitt them before.

> BTW I'm not advocating any particular way. Just been out of the project for a while and looking to pick up some tasks that I find interesting. It seems like feature development is on hold for 3.5. Perhaps that makes sense for this team. At work we snap the branch and keep working on main on new features without really looking back, so I'm used to that.

It is just RussKie taking a minor break before he release 3.5. There are discussions about some GUI appearance changes for 3.5 too, maybe OK after the first public (beta) release.
I believe this PR should be merged soon.

RussKie:
By kinks I meant the bugs that Gerhard found. I'm sure you wouldn't want an
incomplete or non-working code merged ;)

Once the general stability of this change is confirmed - let's merge.

drewnoakes:
All known bugs are currently addressed. I've been using it for the last week, but obviously don't use all features in the application.

gerhardol:
I get this when leaving settings (not seen it before (private build, not investigated)

```
Microsoft.Assumes+InternalErrorException: An internal error occurred. Please contact Microsoft Support.
   at Microsoft.Assumes.Fail(String message)
   at Microsoft.Assumes.NotNull[T](T value)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__CreateLink|15_7(Control container, String text, Image icon, EventHandler handler)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>b__15_2(Panel panel)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__AddLinks|15_6(Panel panel, Func`2 addLinks, Action`2 onLayout)
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.<RefreshContent>g__InitDashboardLayout|15_1()
   at GitUI.CommandsDialogs.BrowseDialog.DashboardControl.Dashboard.RefreshContent()
   at GitUI.CommandsDialogs.FormBrowse.OnShowSettingsClick(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripButton.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.FireEventInteractive(EventArgs e, ToolStripItemEventType met)
   at System.Windows.Forms.ToolStripItem.FireEvent(EventArgs e, ToolStripItemEventType met)
   at GitUI.ToolStripEx.WndProc(Message& m)
```

drewnoakes:
> I get this when leaving settings (not seen it before (private build, not investigated)
> (also not on latest commit for this branch)

@gerhardol I believe that was already fixed in 0240f7e97f634c054a39998552e14ae8f6d5c7fc.

GitExtUtils/Validates.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft
```
mstv, line 4: Have you chosen this namespace by intention?
```suggestion
namespace GitExtUtils
```

gerhardol: Microsoft.Validates replaces Microsoft.Assumes so changing the namespace would require a manual change in every file using it.
Microsoft.Assumes should change the exception message (see discussion above) as it is used in other products, then we could switch back to Assumes. 
But it should be commented at least.

drewnoakes: I can add a comment to explain the namespace choice.

Off topic, but: has there been any discussion about consolidating the namespaces and assembly names used across Git Extensions? They're fairly inconsistent.

drewnoakes: Added comment and fixed merge conflicts.

gerhardol: > Off topic, but: has there been any discussion about consolidating the namespaces and assembly names used across Git Extensions? They're fairly inconsistent.

Not what I have seen

mstv:
Works for me.

drewnoakes:
Spoke with @RussKie offline about this PR. Merging now. Will get another together soon with further annotations.

GitUI was the biggest project, so future PRs will be more focussed.

pmiossec:
@drewnoakes I have this popup every time I open the File History form on a file:

image

Is it already fixed in a PR?

drewnoakes:
> @drewnoakes I have this popup every time I open the File History form on a file:
> 
> image
> 
> Is it already fixed in a PR?

I don't know without seeing a stack trace. Can you file an issue with more details and assign it to me?

drewnoakes:
Filed the issue as https://github.com/gitextensions/gitextensions/issues/8966

RussKie:
These annotations have really screwed me over in my .NET 5.0 conversion branch... I now have to deal with thousands of NRT-related errors all across the solution.

drewnoakes:
@RussKie that sucks. Why so many warnings though? Is it because we're picking up new framework annotations? If so, then those are potential real bugs to address. A bunch of `File`/`Directory` static methods can return null in ways that we rarely handle, for example.

RussKie:
Ignoring [hundreds of CA1416](https://github.com/dotnet/roslyn-analyzers/issues/4983), I get 377 of NRT-related issues in GitUI project alone
image

drewnoakes:
But aren't these legit potential errors? That is, I know this sucks but don't we ultimately want to know about these?

If it's too much to take them on at once (it probably is) consider baselining them in the global suppression file. We can pick them off later.

Most of these will be very quick to fix however.

RussKie:
Given my limited bandwidth fixing those is a significant overhead for me.
If you have time - feel free to push fixes into
https://github.com/RussKie/gitextensions/tree/Convert_to_.NET_5.0_build1

Issue #8745, opened 2021-01-11
Splitter positions are not restored correctly


mstv:
## Current behaviour

The splitter positions of FormBrowse are not re/stored correctly. At least the CommitInfo panel is shrinked by 8 pixels if displayed beside the RevisionGrid.

## Expected behaviour

CommitInfo width is restored - all splitter positions of course.

## Steps to reproduce

- position CommitInfo beside RevisionGrid
- restart GE

## Did this work in previous version of GitExtensions

yes, before borders were added

## Environment

- Git Extensions 33.33.33
- Build d2830009cd34a660040c66e49a68e7717cd3b2d4
- Git 2.27.0.windows.1 (recommended: 2.28.0 or later)
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

ivangrek:
In the current version, this also doesn't work for me.

gerhardol:
There has been related issues (in FormCommit?) maybe a year ago with similar behavior. (have not searched).

ivangrek:
Saving/restoring the splitter position is implemented incorrectly, which leads to breakdown of tests.
The restoration of the position should take place on Form_Show event, and not before that.

Splitter position is work incorrectly. Splitters affect the position of each other in sequential recovery.

From # 8849, where this ability is disabled.

RussKie:
> on Form_Show event

We should use `override OnShow(...)` to reduce unnecessary allocations. There's also `Application.Idle` event. 🤔

mstv:
We should mention in the release notes of v3.5 that this has been broken by force in #8849!

yegor-mialyk:
#8072

PR #8849, opened 2021-02-12
Add borders in Commit Form and set splitter width like in VS

### LINKED ISSUE #8712: Missing borders in FormCommit
## Current behaviour

#8557 has added borders to FormBrowse. This affected FormCommit, too, which is lacking borders now.

## Expected behaviour

At least add bottom borders to the list of unstaged files and to the patch view.

## Screenshots

current FormCommit

former FormCommit

## Environment

- Git Extensions 33.33.33
- Build 468d7955a25faad218e16268ba2a56b07da5508c
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

ivangrek:
Fixes #8712

## Proposed changes

- Add bordes
- Set splitter width like in VS

### Before

image

### After

image

image

### Like
image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build aabe00bb81fba440c57170f601384418a882a06d (Dirty)
- Git 2.30.0.windows.2
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/SplitterManager.cs
+++ b/GitUI/SplitterManager.cs
@@ -95,7 +95,8 @@ namespace GitUI
                                 break;
                             case FixedPanel.Panel2:
                                 int panel2PrevSize = DpiUtil.Scale(prevSize, prevDpi) - DpiUtil.Scale(prevDistance, prevDpi);
-                                SetSplitterDistance(SplitterSize - panel2PrevSize + splitterWidth);
+                                const int paddingOffset = 2; // Refer to FormCommit.ctor+WorkaroundPaddingIncreaseBug
+                                SetSplitterDistance(SplitterSize - panel2PrevSize - paddingOffset + splitterWidth);
                                 break;
                         }
                     }
--- a/GitUI/SplitterManager.cs
+++ b/GitUI/SplitterManager.cs
@@ -27,18 +27,20 @@ namespace GitUI
 
         public void RestoreSplitters()
         {
-            foreach (var splitter in _splitters)
-            {
-                splitter.RestoreFromSettings(_settings);
-            }
+            // TODO: Disabled as splitters have become unstable. Refer to #8745 for more details.
+            ////foreach (var splitter in _splitters)
+            ////{
+            ////    splitter.RestoreFromSettings(_settings);
+            ////}
         }
 
         public void SaveSplitters()
         {
-            foreach (var splitter in _splitters)
-            {
-                splitter.SaveToSettings(_settings);
-            }
+            // TODO: Disabled as splitters have become unstable. Refer to #8745 for more details.
+            ////foreach (var splitter in _splitters)
+            ////{
+            ////    splitter.SaveToSettings(_settings);
+            ////}
         }
 
         private sealed class SplitterData
--- a/GitExtUtils/GitUI/ControlDpiExtensions.cs
+++ b/GitExtUtils/GitUI/ControlDpiExtensions.cs
@@ -70,7 +70,7 @@ namespace GitUI
 
                     case SplitContainer splitContainer:
                     {
-                        const int splitterWidth = 8;
+                        const int splitterWidth = 6;
 
                         if (!isDpiScaled)
                         {
--- a/GitUI/CommandsDialogs/FormBrowse.Designer.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.Designer.cs
@@ -873,7 +873,7 @@ namespace GitUI.CommandsDialogs
             this.revisionGpgInfo1.Location = new System.Drawing.Point(0, 0);
             this.revisionGpgInfo1.Margin = new System.Windows.Forms.Padding(0);
             this.revisionGpgInfo1.Name = "revisionGpgInfo1";
-            this.revisionGpgInfo1.Padding = new System.Windows.Forms.Padding(8);
+            this.revisionGpgInfo1.Padding = new System.Windows.Forms.Padding(6);
             this.revisionGpgInfo1.Size = new System.Drawing.Size(646, 264);
             this.revisionGpgInfo1.TabIndex = 0;
             // 
@@ -1758,7 +1758,7 @@ namespace GitUI.CommandsDialogs
             // 
             this.toolPanel.ContentPanel.Controls.Add(this.MainSplitContainer);
             this.toolPanel.ContentPanel.Margin = new System.Windows.Forms.Padding(0);
-            this.toolPanel.ContentPanel.Padding = new System.Windows.Forms.Padding(8);
+            this.toolPanel.ContentPanel.Padding = new System.Windows.Forms.Padding(6);
             this.toolPanel.ContentPanel.Size = new System.Drawing.Size(1846, 1023);
             this.toolPanel.Dock = System.Windows.Forms.DockStyle.Fill;
             this.toolPanel.LeftToolStripPanelVisible = false;
--- a/GitUI/CommandsDialogs/FormCommit.Designer.cs
+++ b/GitUI/CommandsDialogs/FormCommit.Designer.cs
@@ -616,6 +616,7 @@ namespace GitUI.CommandsDialogs
             //
             this.splitMain.Panel1.Controls.Add(this.splitLeft);
             this.splitMain.Panel1.Controls.Add(this.Ok);
+            this.splitMain.Panel1.Padding = new System.Windows.Forms.Padding(6, 6, 0, 6);
             //
             // splitMain.Panel2
             //
@@ -624,6 +625,7 @@ namespace GitUI.CommandsDialogs
             this.splitMain.SplitterDistance = 397;
             this.splitMain.TabIndex = 0;
             this.splitMain.TabStop = false;
+            this.splitMain.Panel2.Padding = new System.Windows.Forms.Padding(0, 6, 6, 6);
             //
             // splitLeft
             //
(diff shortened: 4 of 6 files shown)


mstv (commented):
@ivangrek, please add screenshots of "before" in order to ease reviews.
Please consider maximized windows.
I expected only https://github.com/gitextensions/gitextensions/issues/8712#issuecomment-753682476.
@drewnoakes specifically removed several borders / sunken panels in the previous UI rework.

mstv:
My 2ct: I don't think MS Visual Studio should be taken as the reference template.

ivangrek:
> My 2ct: I don't think MS Visual Studio should be taken as the reference template.

I agree, it doesn't look as modern as we can see in many other applications.

But I have these thoughts.
Now it is more important to come to an unambiguous interface in all dialogues. The simplest and most understandable will be to take as a basis what is already there, used and understandable. We could consider this as an intermediate goal for the transition to a more modern design. There are many design nuances and we can get bogged down in that.

ivangrek:
> I expected only #8712 (comment).

Anyway, if it causes such resistance, I can suggest using the experimental setting to enable this mode.
Otherwise, we can roll everything back, since the movement in this direction takes too much time and effort.

drewnoakes:
I removed lots of borders, mostly trying to simplify things. We had so many 1px borders, nested, with inconsistent padding. Some places they would have no padding, leading to nested controls producing visible 2px or 3px borders. Removing everything seemed sensible. I'm not opposed to reintroducing some borders in an intentional and controlled manner.

Consider too that we have many more dialogs that would benefit from consistent styling (push, pull, rebase...). Whatever style design evolves here should be applicable across the board.

Keep up the great work @ivangrek!

RussKie:
There's a work item that track consistency of dialogs layouts.
Few dialogs have been modernised, but there's still a long way to go.

ivangrek:
Ref to #8745.

Saving/restoring the splitter position is implemented incorrectly, which leads to breakdown of tests.
The restoration of the position should take place on Form_Show event, and not before that.

pmiossec:
Do you think it could be a good idea to remove the margin and the 2 borders between the unstaged and staged files control?

image

Because I don't know why the stage/unstage toolbar belong more to the bottom control than the top control....

ivangrek:
> Do you think it could be a good idea to remove the margin and the 2 borders between the unstaged and staged files control?

That's a good question. Now it is so implemented. I do not want to dive into the details of each, but focused on the main idea of bringing to an unambiguous form. As far as this raises other issues as well.

Perhaps in the current situation it will be easier to transfer buttons to another panel.
But on the other hand, these buttons are in the area of responsibility of the lower panel, while the upper one is for viewing.

I could suggest moving towards a similar design as the left panel, but we're not ready for that yet. 😄

RussKie:
My 20c - I _do_ like the new main window layout, to me personally it feels significantly more modern and closer to the current UI/UX paradigms of Windows and Office than the current borderless/marginless layout.

With that I _do_ like the proposed layout 👍

RussKie:
> image
> 
> Because I don't know why the stage/unstage toolbar belong more to the bottom control than the top control....

Good observation @pmiossec!
I don't have an answer here either, and I'd split this into a separate UI design discussion. And I throw a more radical idea into the mix - "Reset all changes" and "Reset unstaged" buttons probably belong more to the same panel as "Stage" and "Unstage" buttons :D

ivangrek:
Can we merge this PR?

mstv:
> Can we merge this PR?

Not yet because of just disabling broken splitter positions in https://github.com/gitextensions/gitextensions/pull/8849/commits/d83a7939c9d93a9add5a313766cc782a96ac49a3. Don't forget #8745, too!

BTW: Much less noisy:
```
#if false
    disabled code
#endif
```

RussKie (commented):
What's the deal with the commented out code?

ivangrek:
> What's the deal with the commented out code?

Splitter position is work incorrectly. Splitters affect the position of each other in sequential recovery.

RussKie:
The commit message doesn't explain this, there is brief mention of this on
GitHub. Please update the commit message with an explanation to make it
easier to appreciate the change.
Also I'd follow @mstv suggestion to use #if..#endif or /* */ to reduce the
noise.

So we're fixing one thing and breaking another... Are you working on
getting the splitters working? We can't take this into v3.5 without them.

ivangrek:
> Also I'd follow @mstv suggestion to use #if..#endif or /* */ to reduce the
noise.

In general, or instead of experimental settings?

ivangrek:
> Are you working on getting the splitters working?

I am researching what can be done, but do not see a solution yet.

RussKie:
> > Also I'd follow @mstv suggestion to use #if..#endif or /* */ to reduce the
> > noise.
> 
> In general, or instead of experimental settings?

In this PR :)

RussKie:
> > Are you working on getting the splitters working?
> 
> I am researching what can be done, but do not see a solution yet.

I'll try to find some time this week to pull it down and have a look.

ivangrek:
> In this PR :)

Not quite sure how to enable / disable this?

GitUI/SplitterManager.cs
```diff
@@ -29,18 +29,18 @@ public void AddSplitter(SplitContainer splitter, string settingName, int? defaul
...
         public void RestoreSplitters()
         {
-            foreach (var splitter in _splitters)
-            {
-                splitter.RestoreFromSettings(_settings);
-            }
+            ////foreach (var splitter in _splitters)
+            ////{
+            ////    splitter.RestoreFromSettings(_settings);
+            ////}
```
RussKie, line 35:
```suggestion
            /*
             Disabled as splitters have become unstable. Refer to #123456 for more details.
             
            foreach (var splitter in _splitters)
            {
                splitter.RestoreFromSettings(_settings);
            }
            */
```

Please create a new issue detailing the problem, and reference it here and everywhere else where the code is commented.

ivangrek: #8745

RussKie:
I've run this branch locally, and here are my observations:

1. I can see a small vertical creep of the horizontal splitter (the orange arrow). I couldn't see any splitter instabilities in the `FormBrowse`. I would classify this as minor mildly annoying issue.
2. Visual artifacts marked by the red and cyan colours, these are significantly more concerning. The left panel's top panel exhibits the same issue btw.

image

RussKie:
Looks like we're rendering the toolbar with rounded corners - @NikolayXHD any ideas?
image

ivangrek:
I think Professional Renderer try to render shadows.

RussKie:
The doubled right border comes from `FileStatusList` from `FileStatusListView`.
image

RussKie:
> 2\. Visual artifacts marked by the red and cyan colours, these are significantly more concerning. The left panel's top panel exhibits the same issue btw.

This looks like a theming-related issue:

| Default | Dark |
--|--
image | image

In the default (system) theme the column vertical line is of a faint light blue colour. In the dark theme - the column line's colour is significantly more prominent, and looks the same as the colour as the border.


The toolbar rendering issue only appears in the "themed" scenarios too.

RussKie:
I propose we toned down the borders a little:
```diff
-            splitLeft.Panel1.BackColor = Color.LightGray.AdaptBackColor();
-            splitLeft.Panel2.BackColor = Color.LightGray.AdaptBackColor();
-            splitRight.Panel1.BackColor = Color.LightGray.AdaptBackColor();
-            splitRight.Panel2.BackColor = Color.LightGray.AdaptBackColor();
+            Color borderColor = KnownColor.ControlLightLight.MakeBackgroundDarkerBy(0.08);
+            splitLeft.Panel1.BackColor = borderColor;
+            splitLeft.Panel2.BackColor = borderColor;
+            splitRight.Panel1.BackColor = borderColor;
+            splitRight.Panel2.BackColor = borderColor;
```

<details>
<summary>Current proposal</summary>

* Default
image
* Dark
image
* Dark silver
image | 


</details>

<details>
<summary>New proposal</summary>

* Default
image
* Dark
image
* Dark silver
image



</details>

<details>
<summary>Bonus</summary>





Applied to the main window, it looks like this:
* Default
image
* Dark
image
* Dark silver
image

</details>

ivangrek:
> I propose we tone down the borders a little

👍 but suggest ` #dddddd`

ivangrek:
Why tests locally and on AppVeyor may differ?

GitUI/SplitterManager.cs
```diff
@@ -95,7 +95,8 @@ public void RestoreFromSettings(ISettingsSource settings)
                                 break;
                             case FixedPanel.Panel2:
                                 int panel2PrevSize = DpiUtil.Scale(prevSize, prevDpi) - DpiUtil.Scale(prevDistance, prevDpi);
-                                SetSplitterDistance(SplitterSize - panel2PrevSize + splitterWidth);
+                                const int paddingOffset = 2; // Refer to FormCommit.ctor+WorkaroundPaddingIncreaseBug
```
RussKie, line 98: @ivangrek this makes the splitter stable at the runtime again, however the tests on the CI are still unstable

ivangrek: I don't notice any shifts for the "Commit Form", but sometimes there are changes for the main form.

RussKie:
In essence the change is happening for
splitcontainers with FixedPanel=FixedPanel.Panel2.

How to repro:
1. Open FormCommit
2. Take a screenshot1
3. Close FormCommit
4. Repeat 1-3
5. Compare screenshot1 with screenshot2

You can also attach a debugger and set a breakpoint at SetSplitterDistance
method and see the value changing for "splitRight_" component.

ivangrek:
I marked the points with a pencil on the monitor and just reopen the window and see if there are any shifts. 😄

RussKie:
Not going to work. Use a permanent marker 😁

PR #8851, opened 2021-02-12
Git commands: Switch from the UI thread

gerhardol:
## Proposed changes

Switch from the UI thread before running Git commands.
This is to avoid GUI lock as described https://github.com/gitextensions/gitextensions/pull/8700#issuecomment-778493315
(This was initially for git-status for commit count but expanded.)

The async handling works in a way I did not expect, clarified by sharwell below.
This PR changes the behavior for common commands in Browse.
I checked some other forms too, nothing remarkable as well as FileAndForget().
Git commands still run on the UI thread after this PR, but only those in Browse OnActivation, needed anyway.

The Git commands running on the UI thread causes the UI to be less responsive and seem to be hanging.
In most setups, this is hardly noticable, the repo must be big and a virus run to slow Git down to notice this.
This was especially seen for the left panel.

## Screenshots 

Note: some changes are reverted after review comments.

Before|After
--|--
image|image

## Test methodology 

manual, a big repo is required to see a difference.

----

### DIFF
--- a/GitCommands/CommitDataManager.cs
+++ b/GitCommands/CommitDataManager.cs
@@ -4,9 +4,12 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 using GitCommands.Git.Extensions;
 using GitExtUtils;
+using GitUI;
 using GitUIPluginInterfaces;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitCommands
 {
@@ -200,7 +203,11 @@ namespace GitCommands
             };
 
             // Do not cache this command, since notes can be added
-            data = GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
+            data = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
+            });
 
             if (GitModule.IsGitErrorMessage(data))
             {
--- a/GitCommands/CommitDataManager.cs
+++ b/GitCommands/CommitDataManager.cs
@@ -4,12 +4,9 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
 using GitCommands.Git.Extensions;
 using GitExtUtils;
-using GitUI;
 using GitUIPluginInterfaces;
-using Microsoft.VisualStudio.Threading;
 
 namespace GitCommands
 {
@@ -203,11 +200,7 @@ namespace GitCommands
             };
 
             // Do not cache this command, since notes can be added
-            data = ThreadHelper.JoinableTaskFactory.Run(async () =>
-            {
-                await TaskScheduler.Default;
-                return GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
-            });
+            data = GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
 
             if (GitModule.IsGitErrorMessage(data))
             {
--- a/GitCommands/Git/AheadBehindDataProvider.cs
+++ b/GitCommands/Git/AheadBehindDataProvider.cs
@@ -2,8 +2,11 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using GitExtUtils;
+using GitUI;
 using GitUIPluginInterfaces;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitCommands.Git
 {
@@ -61,7 +64,11 @@ namespace GitCommands.Git
                 "refs/heads/" + branchName
             };
 
-            var result = GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
+            var result = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
+            });
             if (string.IsNullOrEmpty(result))
             {
                 return null;
--- a/GitCommands/Git/AheadBehindDataProvider.cs
+++ b/GitCommands/Git/AheadBehindDataProvider.cs
@@ -2,11 +2,8 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
 using GitExtUtils;
-using GitUI;
 using GitUIPluginInterfaces;
-using Microsoft.VisualStudio.Threading;
 
 namespace GitCommands.Git
 {
@@ -64,11 +61,7 @@ namespace GitCommands.Git
                 "refs/heads/" + branchName
             };
 
-            var result = ThreadHelper.JoinableTaskFactory.Run(async () =>
-            {
-                await TaskScheduler.Default;
-                return GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
-            });
+            var result = GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
             if (string.IsNullOrEmpty(result))
             {
                 return null;
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2950,7 +2950,11 @@ namespace GitCommands
             // Assume that all GetRefs() are done in the background, which may not be correct in the future.
             const bool noLocks = true;
 
-            var cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            var cmd = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            });
             var refList = _gitExecutable.GetOutput(cmd);
             return ParseRefs(refList);
         }
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2652,7 +2652,7 @@ namespace GitCommands
                         async () =>
                         {
                             await TaskScheduler.Default.SwitchTo(alwaysYield: true);
-                            return SubmoduleHelpers.GetCurrentSubmoduleChangesAsync(this, item.Name, item.OldName, firstId, secondId);
+                            return SubmoduleHelpers.GetCurrentSubmoduleChanges(this, item.Name, item.OldName, firstId, secondId);
                         }));
             }
         }
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2950,11 +2950,7 @@ namespace GitCommands
             // Assume that all GetRefs() are done in the background, which may not be correct in the future.
             const bool noLocks = true;
 
-            var cmd = ThreadHelper.JoinableTaskFactory.Run(async () =>
-            {
-                await TaskScheduler.Default;
-                return GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
-            });
+            var cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
             var refList = _gitExecutable.GetOutput(cmd);
             return ParseRefs(refList);
         }
--- a/GitCommands/Git/SubmoduleHelpers.cs
+++ b/GitCommands/Git/SubmoduleHelpers.cs
@@ -23,9 +23,9 @@ namespace GitCommands.Git
             return ParseSubmodulePatchStatus(patch, module, fileName);
         }
 
-        public static Task<GitSubmoduleStatus?> GetCurrentSubmoduleChangesAsync(GitModule module, string submodule, bool noLocks = false)
+        public static async Task<GitSubmoduleStatus?> GetCurrentSubmoduleChangesAsync(GitModule module, string submodule, bool noLocks = false)
         {
-            return GetCurrentSubmoduleChangesAsync(module, submodule, submodule, false, noLocks: noLocks);
+            return await GetCurrentSubmoduleChangesAsync(module, submodule, submodule, false, noLocks: noLocks);
         }
 
         private static GitSubmoduleStatus? ParseSubmodulePatchStatus(Patch? patch, GitModule module, string? fileName)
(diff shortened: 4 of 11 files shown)


gerhardol:
@sharwell Can you provide any insight of the best practices here?

sharwell:
> The reason for that seem to be that the Git command is running on the UI thread.

Process execution supports true asynchronous coding, but last time I looked it's not implemented "all the way down" through `GitModule`.

> I also thought that FileAndForget() would ensure that the UI thread was not used.

`JoinableTaskFactory.RunAsync` _always_ runs synchronously up to the point of the first yield. You can ensure a background thread is used by explicitly switching to the thread pool, or you can explicitly yield to queue operation on the current synchronization context.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -431,6 +432,7 @@ private void Update()
                         bool isSuccess = false;
                         try
                         {
+                            await Task.Delay(1).ConfigureAwait(false);
```
sharwell, line 435: Projects using vs-threading (including Git Extensions) would use this instead:

```csharp
await TaskScheduler.Default;
```

gerhardol: Great
There are some further situations to investigate...

drewnoakes:
@gerhardol do we track when a process is launched on the UI thread, perhaps in our executable infrastructure?

gerhardol:
> @gerhardol do we track when a process is launched on the UI thread, perhaps in our executable infrastructure?

It is displayed in the *Git Command Log* window. I am preparing an update to this PR moving most Git commands from the UI thread.

gerhardol:
This should be included in release/3.5

GitCommands/CommitDataManager.cs
```diff
@@ -200,7 +203,11 @@ private bool TryGetCommitLog(string commitId, string format, [NotNullWhen(return
             };
 
             // Do not cache this command, since notes can be added
-            data = GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
+            data = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
+            });
```
sharwell, line 210: Why is this changing? `Run` blocks the current thread exactly the same way that calling `GetOutput` directly blocks the current thread. This form is both more difficult to read and requires two threads to do the work that previously only needed one.

sharwell (changes requested):

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -278,6 +280,7 @@ public void SelectionChanged(IReadOnlyList<GitRevision> selectedRevisions)
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
                 cancellationToken.ThrowIfCancellationRequested();
+                await TaskScheduler.Default;
```
sharwell, line 283: ❔ Why are we adding this? The call to `GetMergedBranchesAsync` should be properly asynchronous in the case where it's used, and for other cases we're just going to switch right back to the main thread. The previous code was more efficient.

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -61,7 +64,11 @@ public AheadBehindDataProvider(Func<IExecutable> getGitExecutable)
                 "refs/heads/" + branchName
             };
 
-            var result = GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
+            var result = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GetGitExecutable().GetOutput(aheadBehindGitCommand, outputEncoding: encoding);
+            });
```
sharwell, line 71: The previous code was better.

GitCommands/Git/GitModule.cs
```diff
@@ -2950,7 +2950,11 @@ public IReadOnlyList<IGitRef> GetRefs(bool tags = true, bool branches = true)
             // Assume that all GetRefs() are done in the background, which may not be correct in the future.
             const bool noLocks = true;
 
-            var cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            var cmd = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            });
```
sharwell, line 2957: The previous code was better.

GitCommands/Git/SubmoduleHelpers.cs
```diff
@@ -23,9 +23,9 @@ public static class SubmoduleHelpers
             return ParseSubmodulePatchStatus(patch, module, fileName);
         }
 
-        public static Task<GitSubmoduleStatus?> GetCurrentSubmoduleChangesAsync(GitModule module, string submodule, bool noLocks = false)
+        public static async Task<GitSubmoduleStatus?> GetCurrentSubmoduleChangesAsync(GitModule module, string submodule, bool noLocks = false)
```
sharwell, line 26: Why is this changing?

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -139,11 +139,15 @@ protected override void ApplyStyle()
...
-                    TreeViewNode.ToolTipText = LocalizationHelpers.ProcessSubmoduleStatus(
+                    TreeViewNode.ToolTipText = ThreadHelper.JoinableTaskFactory.Run(async () =>
+                    {
+                        await TaskScheduler.Default;
+                        return LocalizationHelpers.ProcessSubmoduleStatus(
                         new GitModule(Info.Path),
                         Info.Detailed.RawStatus,
                         moduleIsParent: false,
                         limitOutput: true);
+                    });
```
sharwell, line 150: The previous code was better.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -431,6 +434,7 @@ private void Update()
                         bool isSuccess = false;
                         try
                         {
+                            await TaskScheduler.Default;
```
sharwell, line 437: ❔ Why is this changing? I would expect `GitCommandHelpers.GetAllChangedFilesCmd` to return effectively immediately, and `GetOutputAsync` to be properly asynchronous. If part of this assumption does not hold, can you clarify which part and why?

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2866,7 +2866,8 @@ private async Task PopulateToolbarAsync(SubmoduleInfoResult result, Cancellation
             // then refresh all items at once with a single switch to the main thread
             ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
             {
-                var loadDetails = newItems.Select(e => e.loadDetails).Where(e => e is not null);
+                await TaskScheduler.Default;
+                var loadDetails = newItems.Select(e => e.loadDetails).WhereNotNull();
```
sharwell, line 2870: The previous code would be better for all cases except ones where `loadFunc` is improperly written. Were there any such cases?

GitUI/GitUIExtensions.cs
```diff
@@ -82,12 +68,16 @@ public static class GitUIExtensions
...
+                string output = ThreadHelper.JoinableTaskFactory.Run(async () =>
+                {
+                    await TaskScheduler.Default;
+                    return fileViewer.Module.GetRangeDiff(
+                            firstId,
+                            item.SecondRevision.ObjectId,
+                            item.BaseA,
+                            item.BaseB,
+                            fileViewer.GetExtraDiffArguments(isRangeDiff: true));
+                });
```
sharwell, line 80: The previous code was better.

GitUI/GitUIExtensions.cs
```diff
@@ -96,8 +86,12 @@ public static class GitUIExtensions
...
             }
 
-            string selectedPatch = GetSelectedPatch(fileViewer, firstId, item.SecondRevision.ObjectId, item.Item)
-                ?? defaultText;
+            string selectedPatch = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return GetSelectedPatch(fileViewer, firstId, item.SecondRevision.ObjectId, item.Item)
+                    ?? defaultText;
+            });
```
sharwell, line 94: The previous code was better.

GitUI/GitUIExtensions.cs
```diff
@@ -144,6 +138,20 @@ void OpenWithDiffTool()
                 return file.IsSubmodule
                     ? LocalizationHelpers.ProcessSubmodulePatch(fileViewer.Module, file.Name, patch)
                     : patch?.Text;
+
+                static Patch GetItemPatch(
```
sharwell, line 142: ❔ Is this used anywhere?

GitUI/UserControls/FileStatusList.cs
```diff
@@ -643,7 +645,11 @@ public void SetDiffs(IReadOnlyList<GitRevision> revisions, Func<ObjectId, GitRev
         {
             _revisions = revisions;
             _getRevision = getRevision;
-            GitItemStatusesWithDescription = _diffCalculator.SetDiffs(revisions, DescribeRevision, getRevision);
+            GitItemStatusesWithDescription = ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                await TaskScheduler.Default;
+                return _diffCalculator.SetDiffs(revisions, DescribeRevision, getRevision);
+            });
```
sharwell, line 652: The previous code was better.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1249,7 +1255,11 @@ private void FileStatusListView_ContextMenu_Opening(object sender, CancelEventAr
                 showAllDiferencesItem.CheckedChanged += (s, e) =>
                 {
                     AppSettings.ShowDiffForAllParents = showAllDiferencesItem.Checked;
-                    GitItemStatusesWithDescription = _diffCalculator.SetDiffs(_revisions, DescribeRevision, _getRevision);
+                    GitItemStatusesWithDescription = ThreadHelper.JoinableTaskFactory.Run(async () =>
+                    {
+                        await TaskScheduler.Default;
+                        return _diffCalculator.SetDiffs(_revisions, DescribeRevision, _getRevision);
+                    });
```
sharwell, line 1262: The previous code was better.

GitCommands/CommitDataManager.cs
gerhardol: reverting this and similar changes. 
I expected the UI thread not to be blocked by these changes, available for other actions.

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: reverting

GitCommands/Git/GitModule.cs
gerhardol: reverting

GitUI/GitUIExtensions.cs
gerhardol: reverting

GitUI/UserControls/FileStatusList.cs
gerhardol: reverting

GitUI/UserControls/FileStatusList.cs
gerhardol: reverting

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: I expected it to work as you described too, but the Git command is executing on the UI thread according to the GE Git Command Log
(is that presentation incorrect?)

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: loadFunc launches Git commands that will run on the UI thread when setting up this menu, so I believe this makes sense

GitUI/GitUIExtensions.cs
gerhardol: reverting

GitUI/GitUIExtensions.cs
gerhardol: GetItemPatch is used in GetSelectedPatch (why it was made into a local function)

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
gerhardol: reverting

GitCommands/Git/SubmoduleHelpers.cs
gerhardol: The Async suffix indicated that the method should be async. I could revert this

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: Git Command Log reports that the command executes on the UI thread
The call is like `.GetOutputAsync(GitCommand).ConfigureAwait(false))`

gerhardol:
@sharwell thanks for the fast review
Some of the reverted changes are for parts that occasionally seem to block the UI, other changes need to be done instead.

GitCommands/Git/SubmoduleHelpers.cs
sharwell: A method is asynchronous or not based on the use of a Task-like return type (e.g. `Task`, `Task<TResult>`, or `ValueTask<TResult>`). The `async` modifier has no impact on this determination.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
sharwell: > but the Git command is executing on the UI thread

The git command _starts_ execution on the UI thread, but as long as the process is executed asynchronously the UI thread will not wait for the process to complete. Starting a process from the UI thread should not cause an observable delay. Waiting for a process to complete on the the UI thread is the problematic operation.

GitUI/BranchTreePanel/RepoObjectsTree.cs
sharwell: See https://github.com/gitextensions/gitextensions/pull/8851/files#r575733576

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: Can the presentation be changed to include this, is it available for the command?
ThreadHelper.JoinableTaskContext.IsOnMainThread is always false after execution also if run synchronously.

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: reverting, loadfunc is normally null

gerhardol:
Thanks for the review @sharwell 
The only remaining for this PR are some refactorings (saving a git call during startup).

Findings:
* Most Git commands in this PR affects the UI, but if possible, changes should be done on a more fundamental state.
* Git Command Log should maybe be presenting not only if the command starts but also if the command finishes on the UI thread.
* I have learnt some about async and threading...

PR #8854, opened 2021-02-14
Browse: Avoid extra git-rev-parse at start

gerhardol:
Remains from #8851

## Proposed changes

- Browse: Avoid extra git-rev-parse at start

There are two calls to `git ls-files -z --unmerged` due to OnActivate()  called from both OnActivated() and InternalInitialize().
This is hard to get around as OnActivated() is needed when the form is activated (from backround, modal window closed)
and InternalInitialize is needed after the module is loaded.

Refactorings:
* GitUIExtensions: GetItemPatch to local function
* Remove Async suffix for GetCurrentSubmoduleChangesAsync

## Screenshots 

See #8851, two calls to `git rev-parse HEAD` quite early.

## Test methodology 

manual

----

### DIFF
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -21,21 +21,6 @@ namespace GitUI
 {
     public static class GitUIExtensions
     {
-        [CanBeNull]
-        private static Patch GetItemPatch(
-            [NotNull] GitModule module,
-            [NotNull] GitItemStatus file,
-            [CanBeNull] ObjectId firstId,
-            [CanBeNull] ObjectId secondId,
-            [NotNull] string diffArgs,
-            [NotNull] Encoding encoding)
-        {
-            // Files with tree guid should be presented with normal diff
-            var isTracked = file.IsTracked || (file.TreeGuid is not null && secondId is not null);
-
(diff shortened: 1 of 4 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1233,7 +1233,7 @@ private void SelectInitialRevision()
                 }
                 else
                 {
-                    toBeSelectedObjectIds = new ObjectId[] { Module.GetCurrentCheckout() };
+                    toBeSelectedObjectIds = currentCheckout is null ? Array.Empty<ObjectId>() : new ObjectId[] { currentCheckout };
```
gerhardol, line 1236: This change is from #8846 (where I first applied the changes) for null annotation but makes sense here.

GitUI/GitUIExtensions.cs
```diff
@@ -15,27 +15,13 @@
 using GitUI.UserControls.RevisionGrid;
 using GitUIPluginInterfaces;
 using JetBrains.Annotations;
+using Microsoft.VisualStudio.Threading;
```
ivangrek, line 18: Redundant?

gerhardol: thanks, removing

mstv (approved):

gerhardol:
@msftbot merge in 24 hours

PR #8860, opened 2021-02-15
Link style was not applied to the dashboard

gerhardol:
Fixes #

## Proposed changes

Apply theme to dashboard links.
theme is null when applying the link theme. No exception, but no effect.
This could be considered for 3.5 too.

An alternative to applying the theme is to remove the theme handling here.

### Before

image

image

### After

image

image

## Test methodology 

Manual

----


drewnoakes (approved):

drewnoakes:
You could convert it from a field to a local. It's only used within `RefreshContent`. Then it can be declared as non-null.

RussKie (approved):

RussKie:
Merge at will.

PR #8861, opened 2021-02-17
Fix ownership of FormUpdates

RussKie:
`FormBrowse` is the main form, and hence it doesn't have `Owner` set. This leads to `FormUpdates` being created as an orphan modal form, that can be hidden behind `FormBrowse` and making the app look unresponsive.

image

Correct the ownership and, thus, force the new form to show above the main form.

## Test methodology 

- manual

----


drewnoakes (approved):

ivangrek:
Maybe just use null? 'FormBrowse' is always owner.

RussKie:
`null` leads to the problem I'm fixing. All dialogs we open from `FormBrowse` within the same process must have correct owner (i.e. non `null`) set to prevent child windows getting lost behind and locking the app.

Issue #8862, opened 2021-02-18
Sidepanel slows down UI


gerhardol:
## Current behaviour

The side panel  adds information, especially for changed submodules in master/3.5. #8726 aded most info, there are others too like ahead/behind.
Note that the heavy information about changes is not retrieved but some extra info is needed.
All info (but not only the info added in 3.5) can be disabled.

Note: Side panel changes causes a slowdown every 15th second (see below for an explanation).
There are also slowdowns due to other changes every 5th second.
This issue mostly ignore the "5 second problem", the solution is mostly the same.

The added info runs on the UI thread and can slow down the UI experience occasionally.
#8851 was a misguided attempt to address that...

The screenshots below shows the command log when one submodule is changed for the GE repo on my home PC (with MS Essentials - whatever it is called now).
At work I may have four changed submodules and each command takes five times the example. I occasionally get spinning cursor or sluggish performance when the sidepanel refreshes.

git-status runs every fifth second if there are changes (for instance at a build, as in the screenshots).
The submodule is rate limited to be updated at most every 15th second.

## Expected behaviour

The extra information using some CPU cycles, but UI performance is not affected.
The first load may use a little more CPU, but not so much at refreshes.

There are many ways to improve the behavior:

* Revert the update - I actually rather keep the slowdowns
* Cache the commands, avoid duplicate requests.
* Wait with requesting the info for the tool tip until needed as some info is related to that  - will delay the tool tip though
* Make sure the submodule update is not done on the UI - TBD

The caching and tooltip-on-demand is a simple way to improve also in 3.5.
Maybe increase the cache size slightly too.

## Steps to reproduce

* Make a modification to one submodule
* Start a build for a repo on the command line or start a script like the following in bash `while [ 1 ]; do touch appveyor.yml ; sleep 3; done`
* You may get occasional slowdowns

## Screenshots

3.4
![image](https://user-images.githubusercontent.com/6248932/108428923-9b5cfa00-723f-11eb-8835-3fe5a2139359.png)

master
![image](https://user-images.githubusercontent.com/6248932/108429974-f6432100-7240-11eb-8142-d3b349ff4b58.png)

## Did this work in previous version of GitExtensions

Less requests in 3.4

## Environment

- Git Extensions 33.33.33
- Build 09f1f48caafb58322856ed80e91c10195426ff01
- Git 2.30.1.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4300.0
[...]

PR #8866, opened 2021-02-20
Cache Git commands related to submodules

gerhardol:
Part of #8862 

## Proposed changes

* Cache Git commands related to submodules
Cache commands with sha input and no variable input like Git Notes
Only fetch Git Notes for commands that are not cached (Notes are not displayed in Submodule scenarios anyway).

* Cache GetCommitRangeDiffCount
Used when selecting multiple commits in RevDiff

*  Increase Git command cache to 50
Not strictly needed, 40 is OK also for my use after this, but it seems like a reasonable change.

## Screenshots 

Example with one submodule changed, on a relative fast PC (Git at work is 5 times slower).

Startup followed by refresh (similar to build), marked with red line.
Commands that are cached are marked with green (so cache have effect already at first display).

Note: The refresh was done a little early, a diff command will normally be required also after the update, but for current master four more commands would be required.

### Before

image

### After

image

## Test methodology 

review and manual

----

### DIFF
--- a/GitCommands/CommitDataManager.cs
+++ b/GitCommands/CommitDataManager.cs
@@ -24,9 +24,12 @@ namespace GitCommands
         CommitData CreateFromRevision(GitRevision revision, IReadOnlyList<ObjectId> children);
 
         /// <summary>
-        /// Gets <see cref="CommitData"/> for the specified <paramref name="sha1"/>.
+        /// Gets <see cref="CommitData"/> for the specified <paramref name="commitId"/>.
         /// </summary>
-        CommitData? GetCommitData(string sha1, out string? error);
+        /// <param name="commitId">The sha or Git reference.</param>
+        /// <param name="error">error message for the Git command</param>
+        /// <param name="cache">Allow caching of the Git command, should only be used if commitId is a sha and Notes are not used.</param>
+        CommitData? GetCommitData(string commitId, out string? error, bool cache = false);
 
         /// <summary>
         /// Updates the <see cref="CommitData.Body"/> (commit message) property of <paramref name="commitData"/>.
@@ -36,7 +39,8 @@ namespace GitCommands
 
     public sealed class CommitDataManager : ICommitDataManager
     {
-        private const string CommitDataFormat = "%H%n%T%n%P%n%aN <%aE>%n%at%n%cN <%cE>%n%ct%n%e%n%B%nNotes:%n%-N";
+        private const string CommitDataFormat = "%H%n%T%n%P%n%aN <%aE>%n%at%n%cN <%cE>%n%ct%n%e%n%B";
+        private const string CommitDataWithNotesFormat = "%H%n%T%n%P%n%aN <%aE>%n%at%n%cN <%cE>%n%ct%n%e%n%B%nNotes:%n%-N";
         private const string BodyAndNotesFormat = "%H%n%e%n%B%nNotes:%n%-N";
 
         private readonly Func<IGitModule> _getModule;
@@ -49,7 +53,7 @@ namespace GitCommands
         /// <inheritdoc />
         public void UpdateBody(CommitData commitData, out string? error)
         {
-            if (!TryGetCommitLog(commitData.ObjectId.ToString(), BodyAndNotesFormat, out error, out var data))
+            if (!TryGetCommitLog(commitData.ObjectId.ToString(), BodyAndNotesFormat, out error, out var data, cache: false))
             {
                 return;
             }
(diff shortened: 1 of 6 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -790,15 +790,15 @@ public async Task<List<ConflictData>> GetConflictsAsync(string filename = "")
...
             var args = new GitArgumentBuilder("rev-list")
             {
-                parentHash,
-                $"^{childHash}",
+                parent,
+                $"^{child}",
                 "--count"
             };
-            var output = _gitExecutable.GetOutput(args);
+            var output = _gitExecutable.GetOutput(args, cache: cache ? GitCommandCache : null);
```
ivangrek, line 801: How about this option?
To separate by parameters.

```
var output = cache switch
{
    true => _gitExecutable.GetOutput(args, cache: GitCommandCache),
    false => _gitExecutable.GetOutput(args)
};
```

gerhardol: For this PR, I suggest to follow the current code style (i.e. not change).

But this could be changed in general.
I prefer the current ?: syntax, of course partly because I am used to that but I see no benefit with the switch here.
The current is easier to read for me and less duplicated code. Long commands where options wrap over several lines can be hard to compare.
This scales better when there are more options affecting a command too.

ivangrek: > This scales better when there are more options affecting a command too.

Could you please to show example there is you mean.

gerhardol:
```c#
            return _gitExecutable.GetOutput(
                new GitArgumentBuilder("diff")
                {
                    "--no-color",
                    "-M -C",
                    "--name-status",
                    { nullSeparated, "-z" },
                    _revisionDiffProvider.Get(firstRevision, secondRevision)
                },
                cache: noCache ? null : GitCommandCache);
```

GitCommands/Git/SubmoduleHelpers.cs
```diff
@@ -126,8 +126,8 @@ public static class SubmoduleHelpers
                 else
                 {
                     var submodule = module.GetSubmodule(fileName);
-                    addedCommits = submodule.GetCommitCount(commitId.ToString(), oldCommitId.ToString());
-                    removedCommits = submodule.GetCommitCount(oldCommitId.ToString(), commitId.ToString());
+                    addedCommits = submodule.GetCommitCount(commitId.ToString(), oldCommitId.ToString(), true);
+                    removedCommits = submodule.GetCommitCount(oldCommitId.ToString(), commitId.ToString(), true);
```
mstv, line 130:
```suggestion
                    addedCommits = submodule.GetCommitCount(commitId.ToString(), oldCommitId.ToString(), cache: true);
                    removedCommits = submodule.GetCommitCount(oldCommitId.ToString(), commitId.ToString(), cache: true);
```

mstv (approved):
👍

GitCommands/Git/GitModule.cs
ivangrek: In the case of the switch, the state machine is clearer.
And the appearance of "null" can signal a potential error.

ivangrek (approved):

gerhardol:
@msftbot merge in 72 hours

gerhardol:
Will of course squash before merge, will create a PR for 3.5 too

GitCommands/CommitDataManager.cs
```diff
@@ -199,8 +203,13 @@ private bool TryGetCommitLog(string commitId, string format, [NotNullWhen(return
                 commitId
             };
 
-            // Do not cache this command, since notes can be added
-            data = GetModule().GitExecutable.GetOutput(arguments, outputEncoding: GitModule.LosslessEncoding);
+            // This command can be cached if commitId is a git sha and Notes are ignored
+#if DEBUG
+            Debug.Assert(!cache || ObjectId.TryParse(commitId, out _), $"Cache should be used only for sha");
+#endif
```
RussKie, line 209: `Debug.Assert` will only work in debug
```suggestion
            Debug.Assert(!cache || ObjectId.TryParse(commitId, out _), $"Cache should be used only for sha");
```

RussKie (approved):

GitCommands/CommitDataManager.cs
gerhardol: I originally had more asserts, but even this is maybe more CPU intensive then required
It may be removed later.

gerhardol:
One fixup, text summary for new submodules could be cached
(common scenario: submodule exists in one branch but not next checked out branch without deleting the submodule directory).

PR #8867, opened 2021-02-20
Submodule SidePanel updated too often

gerhardol:
Part of #8862

## Proposed changes

This PR handles the side panel for submodules being updated too often.
The PR has three parts that could be separated, but the discussion is easier if the combined result is presented.
Similarily, the results should be seen together with #8866 as that also improves the handling. That also reduces clutter to show what this PR really changes.

1. GitStatus updates were handled every 5th second

Regression in master after 3.4
The git-status updates are limited to at most every 5th second (or longer if git-status is really slow).
Submodule updates are limited even more, to every 15th second.
In 3.4 the extra updates were just thrown away.
In 3.5 some code to not require tests to wait these extra 15s were incorrect and updated the sidepanel too.
So the sidepanel as well as the menu dropdown were updated every 5th second, with two updates every 15th second.

2. Submodule tooltip were set twice

The sidepanel recreates the submodule tree at every status update (could be improved, not done here).
The status tooltip added in 3.5 were updated in both of these runs, just ms in between.
The tooltip is set to the default in the first *structure* update and the status update will set the specific status text only once.

3. Submodule tooltip were calculated in the UI thread

The actual diff is calculated by SubmoduleStatusProvider (and are cached by #8866), but *rev-parse --git-common-dir* must run to decode the Git output to present output. (This could maybe be guessed or the processed status changed, not done in this PR).
rev-parse requires about 40 ms on my PC, but at work this requires 200 ms or more. (Processing of the text requires an additional 10ms or so). As the job runs on the UI thread, the UI may get occasional slowdowns, especially if there are many submodules that are changed. The submodule list flicker when the list is refreshed.
The tooltip handling it therefore changed to be done on a worker thread.

Note: Part 1. and 2. are the most important here.
For 3. I tried to change the handling to calculate the tooltip on request too, but did not fully succeed. This solution is intended for 3.5, so I do not want to make it too complicated.

## Screenshots 

Two 15s cycles for each

### Before

Flicker when updating (more visible in real life)

subm-flicker

image

### After

image

## Test methodology 

Manual, inserting delays in the handling.

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
@@ -78,8 +78,7 @@ namespace GitUI.BranchTreePanel
             {
                 if (Info.Detailed is not null && Tree.TreeViewNode.TreeView is not null)
                 {
-                    ApplyText();
-                    ApplyStyle();
+                    ApplyStatus();
                 }
             }
 
@@ -135,26 +134,14 @@ namespace GitUI.BranchTreePanel
                     TreeViewNode.NodeFont = new Font(AppSettings.Font, FontStyle.Bold);
                 }
 
-                if (Info.Detailed?.RawStatus is not null)
-                {
-                    // Prefer submodule status, shows ahead/behind
-                    TreeViewNode.ToolTipText = LocalizationHelpers.ProcessSubmoduleStatus(
-                        new GitModule(Info.Path),
-                        Info.Detailed.RawStatus,
-                        moduleIsParent: false,
-                        limitOutput: true);
-                }
-                else if (GitStatus is not null)
-                {
-                    var changeCount = new ArtificialCommitChangeCount();
-                    changeCount.Update(GitStatus);
-                    TreeViewNode.ToolTipText = changeCount.GetSummary();
-                }
-                else
-                {
-                    TreeViewNode.ToolTipText = DisplayText();
-                }
+                TreeViewNode.ToolTipText = DisplayText();
 
+                // Note that status is applied also after the tree is created, when status is applied
+                ApplyStatus();
+            }
+
+            private void ApplyStatus()
+            {
                 TreeViewNode.ImageKey = GetSubmoduleItemImage(Info.Detailed);
                 TreeViewNode.SelectedImageKey = TreeViewNode.ImageKey;
 
@@ -191,6 +178,35 @@ namespace GitUI.BranchTreePanel
                     // Unknown
                     return details.IsDirty ? nameof(Images.SubmoduleDirty) : nameof(Images.FileStatusModified);
                 }
+            }
+
+            internal async Task SetStatusToolTipAsync(CancellationToken token)
+            {
+                string toolTip;
+                if (Info.Detailed?.RawStatus is not null)
+                {
+                    // Prefer submodule status, shows ahead/behind
+                    toolTip = LocalizationHelpers.ProcessSubmoduleStatus(
+                        new GitModule(Info.Path),
+                        Info.Detailed.RawStatus,
+                        moduleIsParent: false,
+                        limitOutput: true);
+                }
+                else if (GitStatus is not null)
+                {
+                    var changeCount = new ArtificialCommitChangeCount();
+                    changeCount.Update(GitStatus);
+                    toolTip = changeCount.GetSummary();
+                }
+                else
+                {
+                    // No data need to be set
+                    return;
+                }
+
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
+                token.ThrowIfCancellationRequested();
+                TreeViewNode.ToolTipText = toolTip;
             }
         }
 
@@ -254,7 +270,8 @@ namespace GitUI.BranchTreePanel
                     if (cts is not null && loadNodesTask is not null)
                     {
                         var loadedNodes = await loadNodesTask;
-                        await LoadNodeDetailsAsync(cts.Token, loadedNodes).ConfigureAwaitRunInline();
+                        await LoadNodeDetailsAsync(loadedNodes, cts.Token).ConfigureAwaitRunInline();
+                        LoadNodeToolTips(loadedNodes, cts.Token);
                     }
 
                     Interlocked.CompareExchange(ref _currentSubmoduleInfo, null, e);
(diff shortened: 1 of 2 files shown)


GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -193,6 +175,37 @@ string GetSubmoduleItemImage(DetailedSubmoduleInfo details)
                     return details.IsDirty ? nameof(Images.SubmoduleDirty) : nameof(Images.FileStatusModified);
                 }
             }
+
+            public async Task<bool> SetStatusTooltip(CancellationToken token)
```
gerhardol, line 179:
```suggestion
            public async Task<bool> SetStatusTooltipAsync(CancellationToken token)
```

gerhardol:
@mstv Do you have the possibility to review this PR?

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -193,6 +180,37 @@ string GetSubmoduleItemImage(DetailedSubmoduleInfo details)
                     return details.IsDirty ? nameof(Images.SubmoduleDirty) : nameof(Images.FileStatusModified);
                 }
             }
+
+            internal async Task<bool> SetStatusToolTipAsync(CancellationToken token)
```
mstv, line 184: Do you have plans for further usage? Else:
```suggestion
            private async Task SetStatusToolTipAsync(CancellationToken token)
```

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -284,6 +303,24 @@ private async Task LoadNodeDetailsAsync(CancellationToken token, Nodes loadedNod
...
+                        .ForEach(async node =>
+                    {
+                        try
+                        {
+                            await node.SetStatusToolTipAsync(token).ConfigureAwait(false);
+                        }
+                        catch (OperationCanceledException)
+                        {
+                        }
+                    });
```
mstv, line 320: I'd indent this by 1 level.

mstv (commented):
Looks sensible. I have only run the first commit yet (just displaying submodules).

I haven't digged into: Where is the switch to the background thread, i.e. `await TaskScheduler.Default`?

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -193,6 +176,29 @@ string GetSubmoduleItemImage(DetailedSubmoduleInfo details)
...
+                }
+                else if (GitStatus is not null)
+                {
+                    var changeCount = new ArtificialCommitChangeCount();
+                    changeCount.Update(GitStatus);
+                    TreeViewNode.ToolTipText = changeCount.GetSummary();
+                }
+                else
+                {
+                    Debug.Assert(true, "Unexpected default status for SetStatusTooltip");
```
mstv: Nit: I think `Debug.Assert(true, ...)` does nothing at all.
Will be hidden by if all commits are squashed because replaced in a later commit.

gerhardol: This was mostly intended as a comment but should have been false...
Intentionally removed.  The first change to RepoObjectsTree.Nodes.Submodules.cs could be separated, then this Assert has some meaning. The changes should be squashed though.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
gerhardol: ok

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
gerhardol: ok

gerhardol:
> I haven't digged into: Where is the switch to the background thread, i.e. `await TaskScheduler.Default`?

`private void LoadNodeToolTips(CancellationToken token, Nodes loadedNodes)`
uses .ConfigureAwait(false) so the call does not continue on the UI thread

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
mstv: Why not `private`?
The final `return` is superfluous.

gerhardol: removing the last return

`internal` as it is called from SubmoduleTree

mstv:
Nit: The vertical scroll position of the left panel changes when the submodule status is updated on refresh with `F5`.
It flickers for nearly every single submodule both when the status is invalidated and when it is displayed, especially if the Git Command Log is open.
Tested with missing submodules after coming back from one of the original "keepyourlane" branches.

gerhardol:
> Nit: The vertical scroll position of the left panel changes when the submodule status is updated on refresh with `F5`.
> It flickers for nearly every single submodule both when the status is invalidated and when it is displayed, especially if the Git Command Log is open.
> Tested with missing submodules after coming back from one of the original "keepyourlane" branches.

Yes, this is not fixed but instead of refreshing every 5th second (and a double refresh every third time) the refresh is every 15th second (a smaller refresh too).
Most of the time the info is unchanged so there should not be a need to recreate the tree at every status update.
There is currently no difference between a *structure* and a *status* update.

I want to have this in 3.5 though, then a safer solution is needed - these changes are needed anyway.

I would like to squash (to two commits) with the current functionality and prepare for merge...

mstv (approved):

gerhardol:
@msftbot merge in 26 hours

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -288,6 +305,24 @@ private async Task LoadNodeDetailsAsync(CancellationToken token, Nodes loadedNod
                 }
             }
 
+            private void LoadNodeToolTips(CancellationToken token, Nodes loadedNodes)
```
RussKie, line 308:
```suggestion
            private void LoadNodeToolTips(Nodes loadedNodes, CancellationToken token)
```

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -288,6 +305,24 @@ private async Task LoadNodeDetailsAsync(CancellationToken token, Nodes loadedNod
...
+            {
+                if (TreeViewNode.TreeView is not null)
```
RussKie, line 310: [nit] please invert

RussKie (commented):
👍 with 2 small suggestions

gerhardol:
squashed the suggestions, rebased

PR #8869, opened 2021-02-22
Correct FormCherryPick ownership

RussKie:
Still trying to figure out intermittent flakiness reported in #6220. Very subjective, but this little tweak _seems_ to make some difference in local tests. 
Obviously a sweep across the codebase would be ideal but I don't have enough cycles to perform it, hence the piecemeal approach.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCherryPick.cs
+++ b/GitUI/CommandsDialogs/FormCherryPick.cs
@@ -97,7 +97,7 @@ namespace GitUI.CommandsDialogs
             }
         }
 
-        private void Revert_Click(object sender, EventArgs e)
+        private void Pick_Click(object sender, EventArgs e)
         {
             var args = new ArgumentBuilder();
             var canExecute = true;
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/FormCherryPick.cs
```diff
@@ -126,9 +126,9 @@ private void Revert_Click(object sender, EventArgs e)
 
                 // Don't verify whether the command is successful.
                 // If it fails, likely there is a conflict that needs to be resolved.
-                FormProcess.ShowDialog(this, process: null, arguments: command, Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(Owner, process: null, arguments: command, Module.WorkingDir, input: null, useDialogSettings: true);
 
-                MergeConflictHandler.HandleMergeConflicts(UICommands, this, AutoCommit.Checked);
+                MergeConflictHandler.HandleMergeConflicts(UICommands, Owner, AutoCommit.Checked);
```
mstv, line 131: It is not obvious to me why this time, the `Owner` must be used instead of `this` in contrast to 26361af.
FormCherryPick is visible as long as the git command is run.

If this change is correct, it must be applied to the MessageBox in line 109, too.

gerhardol (commented):
+1

GintasS:
This is an old PR and should be closed. @gerhardol

PR #8870, opened 2021-02-22
Cache Git commands related to submodules-3.5

gerhardol:
#8866 for 3.5


mstv (approved):
The commit is identical.

RussKie (approved):

gerhardol:
Squashed the same fixup as in #8866

gerhardol:
Only applied the same change as in master yesterday, squashed today so the commits are identical in master and 3.5

PR #8875, opened 2021-02-24
Remove double execution of GetSubmoduleStatusAsync

drewnoakes:
## Proposed changes

The previous code would call `GetSubmoduleStatusAsync` twice, only using the second call's result.

As the returned task is never null, this check is redundant and does redundant work.

Checking the task's result occurs below.

## Test methodology 

- Manual testing

----

### DIFF
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -112,7 +112,7 @@ namespace GitUI
                         : diffOfConflict;
                 }
 
-                if (file.IsSubmodule && file.GetSubmoduleStatusAsync() is not null)
+                if (file.IsSubmodule)
                 {
                     // Patch already evaluated
                     var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync);


RussKie (approved):

gerhardol (changes requested):
The check is incorrect and should be modified instead.

If _submoduleStatus status is not set, it must be evaluated, GetSubmoduleStatusAsync() does not force that.
If the status is always set, the submodule check a few lines below should be removed too.

gerhardol:
Why did msftbot merge when I requested changes? author feedback is set too (that may not be checked though)

drewnoakes:
> Why did msftbot merge when I requested changes? author feedback is set too (that may not be checked though)

No idea.

> If _submoduleStatus status is not set, it must be evaluated, GetSubmoduleStatusAsync() does not force that.
> If the status is always set, the submodule check a few lines below should be removed too.

I see, you're right. There is more to be tidied up here.

`GetSubmoduleStatusAsync` will never return `null`. So this PR doesn't make anything worse than they were before. I don't think the condition below on line 127 would ever hit the `ProcessSubmodulePatch` call. That is the only usage of the method too.

I don't fully understand the goal of this code. My guess is that, at some point, this code checked whether anything had called `SetSubmoduleStatus` on the item, and that's what the condition I removed used to achieve. Since then it was refactored (possibly during the introduction of JTF) and this bug introduced.

I'd like to repoduce this to see whether it's needed. If not, we can remove some code.

gerhardol:
> I don't fully understand the goal of this code. My guess is that, at some point, this code checked whether anything had called SetSubmoduleStatus on the item, and that's what the condition I removed used to achieve. Since then it was refactored (possibly during the introduction of JTF) and this bug introduced.

Very likely so

There are many ways to create the submodule status. I have tried to clean this up some but it is not clear.
In some sense, if this code works, then the status is always calculated and code can be simplified

PR #8878, opened 2021-02-24
Annotate GitUIPluginInterfaces

drewnoakes:
Follows #8771, #8776, #8846

## Proposed changes

- Annotate `GitUIPluginInterfaces` project
- Update some annotations in previously-annotated code (`GitCommands` and `GitUI` projects)

## Test methodology 

- Manual testing

----

### DIFF
--- a/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
+++ b/GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using GitCommands.Config;
 using GitExtUtils;
 using GitUIPluginInterfaces;
+using Microsoft;
 
 namespace GitCommands.Remotes
 {
@@ -164,12 +165,19 @@ namespace GitCommands.Remotes
         /// </summary>
         public IReadOnlyList<Remote> GetDisabledRemotes()
         {
-            var disabledRemotes = GetModule().LocalConfigFile.GetConfigSections()
-                .Where(s => s.SectionName == $"{DisabledSectionPrefix}remote")
-                .Select(s => new Remote(s.SubSection, s.GetValue("url"), s.GetValue("pushurl", s.GetValue("url"))))
-                .ToList();
+            return GetDisabledRemotes().ToList();
 
-            return disabledRemotes;
+            IEnumerable<Remote> GetDisabledRemotes()
+            {
+                foreach (IConfigSection section in GetModule().LocalConfigFile.GetConfigSections())
+                {
+                    if (section.SectionName == $"{DisabledSectionPrefix}remote")
+                    {
+                        Validates.NotNull(section.SubSection);
+                        yield return new Remote(section.SubSection, section.GetValue("url"), section.GetValue("pushurl", section.GetValue("url")));
+                    }
+                }
+            }
         }
 
         /// <summary>
@@ -245,6 +253,7 @@ namespace GitCommands.Remotes
             var module = GetModule();
             if (!remote.Disabled)
             {
+                Validates.NotNull(remote.Name);
                 return module.RemoveRemote(remote.Name);
             }
 
@@ -260,7 +269,7 @@ namespace GitCommands.Remotes
         /// <returns>True if input remote exists and is enabled.</returns>
         public bool EnabledRemoteExists(string remoteName)
         {
-            return GetEnabledRemoteNames().FirstOrDefault(r => r == remoteName) is not null;
+            return GetEnabledRemoteNames().Any(r => r == remoteName);
         }
 
         /// <summary>
@@ -270,7 +279,7 @@ namespace GitCommands.Remotes
         /// <returns>True if input remote exists and is disabled.</returns>
         public bool DisabledRemoteExists(string remoteName)
         {
-            return GetDisabledRemoteNames().FirstOrDefault(r => r == remoteName) is not null;
+            return GetDisabledRemoteNames().Any(r => r == remoteName);
         }
 
         /// <summary>
@@ -330,6 +339,8 @@ namespace GitCommands.Remotes
                 remoteDisabled = remote.Disabled;
                 if (!string.Equals(remote.Name, remoteName, StringComparison.Ordinal))
                 {
+                    Validates.NotNull(remote.Name);
+
                     // the name of the remote changed - perform rename
                     output = module.RenameRemote(remote.Name, remoteName);
                 }
--- a/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
@@ -409,7 +409,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
 
             var module = new GitModule(targetDir);
 
-            if (addUpstreamRemoteAsCB.Text.Trim().Length > 0 && !string.IsNullOrEmpty(repo.ParentReadOnlyUrl))
+            if (addUpstreamRemoteAsCB.Text.Trim().Length > 0 && !GitExtUtils.Strings.IsNullOrEmpty(repo.ParentReadOnlyUrl))
             {
                 var error = module.AddRemote(addUpstreamRemoteAsCB.Text.Trim(), repo.ParentReadOnlyUrl);
                 if (!string.IsNullOrEmpty(error))
--- a/GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
@@ -183,7 +183,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
                 .First(r => string.IsNullOrEmpty(currentRemote) || r.Name == currentRemote).FetchUrl.IsUrlUsingHttp() ? GitProtocol.Https : GitProtocol.Ssh);
             var hostedRemote = _selectHostedRepoCB.Items.
                 Cast<IHostedRemote>().
-                FirstOrDefault(remote => remote.Name.Equals(currentRemote, StringComparison.OrdinalIgnoreCase));
+                FirstOrDefault(remote => string.Equals(remote.Name, currentRemote, StringComparison.OrdinalIgnoreCase));
 
             if (hostedRemote is null)
             {
@@ -366,8 +366,8 @@ namespace GitUI.CommandsDialogs.RepoHosting
             var giss = new List<GitItemStatus>();
 
             // baseSha is the sha of the merge to ("master") sha, the commit to be firstId
-            GitRevision? firstRev = ObjectId.TryParse(baseSha, out ObjectId firstId) ? new GitRevision(firstId) : null;
-            GitRevision? secondRev = ObjectId.TryParse(secondSha, out ObjectId secondId) ? new GitRevision(secondId) : null;
+            GitRevision? firstRev = ObjectId.TryParse(baseSha, out ObjectId? firstId) ? new GitRevision(firstId) : null;
+            GitRevision? secondRev = ObjectId.TryParse(secondSha, out ObjectId? secondId) ? new GitRevision(secondId) : null;
             if (secondRev is null)
             {
                 MessageBox.Show(this, _strUnableUnderstandPatch.Text, Strings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
--- a/GitUI/Shells/ConEmuControlExtensions.cs
+++ b/GitUI/Shells/ConEmuControlExtensions.cs
@@ -5,9 +5,9 @@ namespace GitUI.Shells
 {
     public static class ConEmuControlExtensions
     {
-        public static void ChangeFolder(this ConEmuControl terminal, IShellDescriptor shell, string path)
+        public static void ChangeFolder(this ConEmuControl? terminal, IShellDescriptor? shell, string? path)
         {
-            if (terminal?.RunningSession is null || shell is null || string.IsNullOrWhiteSpace(path))
+            if (terminal?.RunningSession is null || shell is null || GitExtUtils.Strings.IsNullOrWhiteSpace(path))
             {
                 return;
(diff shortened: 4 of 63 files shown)


GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
```diff
@@ -164,12 +165,19 @@ bool IsSettingForBranch(string setting, string branchName)
...
+            {
+                foreach (IConfigSection section in GetModule().LocalConfigFile.GetConfigSections())
+                {
+                    if (section.SectionName == $"{DisabledSectionPrefix}remote")
+                    {
+                        Validates.NotNull(section.SubSection);
+                        yield return new Remote(section.SubSection, section.GetValue("url"), section.GetValue("pushurl", section.GetValue("url")));
+                    }
+                }
+            }
```
drewnoakes, line 180: Rewrote this to allow validation of `SubSection` as non-null. Roslyn should probably detect this automatically, based upon it being `==` to a non-null value. However `==` is overridable, so it'd have to special-case `string` which seems narrow in scope.

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
```diff
@@ -260,7 +269,7 @@ public string RemoveRemote(ConfigFileRemote remote)
         /// <returns>True if input remote exists and is enabled.</returns>
         public bool EnabledRemoteExists(string remoteName)
         {
-            return GetEnabledRemoteNames().FirstOrDefault(r => r == remoteName) is not null;
+            return GetEnabledRemoteNames().Any(r => r == remoteName);
```
drewnoakes, line 272: I believe this is equivalent.

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -183,7 +183,7 @@ private void SelectHostedRepositoryForCurrentRemote()
                 .First(r => string.IsNullOrEmpty(currentRemote) || r.Name == currentRemote).FetchUrl.IsUrlUsingHttp() ? GitProtocol.Https : GitProtocol.Ssh);
             var hostedRemote = _selectHostedRepoCB.Items.
                 Cast<IHostedRemote>().
-                FirstOrDefault(remote => remote.Name.Equals(currentRemote, StringComparison.OrdinalIgnoreCase));
+                FirstOrDefault(remote => string.Equals(remote.Name, currentRemote, StringComparison.OrdinalIgnoreCase));
```
drewnoakes, line 186: Bug fix. `Name` can be null.

GitUI/Shells/ShellDescriptor.cs
```diff
@@ -7,13 +7,13 @@ namespace GitUI.Shells
     public abstract class ShellDescriptor : IShellDescriptor
     {
         /// <inheritdoc/>
-        public string? ExecutableCommandLine { get; protected set; } = null!;
+        public string? ExecutableCommandLine { get; protected set; }
```
drewnoakes, line 10: These assignments are redundant. This pattern is commonly used for _non_-nullable fields, where assignment is guaranteed in code that the compiler cannot see (i.e. set via reflection, in the case of DI frameworks such as MEF). In this case the properties are nullable, so the assignment is misleading and redundant.

GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
```diff
@@ -185,7 +185,7 @@ public override bool TryGetToolTip(DataGridViewCellMouseEventArgs e, GitRevision
             if (revision.BuildStatus is not null)
             {
                 toolTip = revision.BuildStatus.Tooltip ?? revision.BuildStatus.Description;
-                return true;
+                return toolTip is not null;
```
drewnoakes, line 188: `Description` can be null, so it was possible before to return null and true, which is not expected by callers.

GitUI/UserControls/RevisionGrid/RevisionGridControl.RevisionGridInMemFilter.cs
```diff
@@ -66,8 +66,8 @@ static bool CheckCondition(string? filter, Regex? regex, string? value)
             {
                 if (string.IsNullOrEmpty(authorFilter) &&
                     string.IsNullOrEmpty(committerFilter) &&
-                    string.IsNullOrEmpty(messageFilter) &&
-                    !ObjectId.IsValidPartial(messageFilter, minLength: 5))
+                    (GitExtensions.Strings.IsNullOrEmpty(messageFilter) ||
+                    !ObjectId.IsValidPartial(messageFilter, minLength: 5)))
```
drewnoakes, line 70: Please review this. The previous logic seems broken. If `messageFilter` is null or empty, it would never be a valid partial. I'm pretty sure this was supposed to be `||`.

Plugins/BuildServerIntegration/AzureDevOpsIntegration/Settings/ProjectUrlHelper.cs
```diff
@@ -101,7 +102,7 @@ private static (bool success, string result) TryTransformString(string value, Di
         /// <returns>
         /// A tuple that contains whether a Azure DevOps / TFS project could be recognized from the given url and the resulting home page url of the project.
         /// </returns>
-        public static (bool success, string projectUrl) TryDetectProjectFromRemoteUrl(string remoteUrl)
+        public static (bool success, string projectUrl) TryDetectProjectFromRemoteUrl([CanBeNull] string remoteUrl)
```
drewnoakes, line 105: Added instances of the JetBrains annotations here (and elsewhere) for projects which are not yet annotated, to make annotating them easier in future.

Plugins/GitUIPluginInterfaces/BuildServerIntegration/BuildInfo.cs
```diff
@@ -15,15 +15,15 @@ public enum BuildStatus
...
-        public string Id { get; set; }
+        public string? Id { get; set; }
         public DateTime StartDate { get; set; }
         public long? Duration { get; set; }
         public BuildStatus Status { get; set; }
-        public string Description { get; set; }
-        public IReadOnlyList<ObjectId> CommitHashList { get; set; }
-        public string Url { get; set; }
+        public string? Description { get; set; }
+        public IReadOnlyList<ObjectId> CommitHashList { get; set; } = Array.Empty<ObjectId>();
```
drewnoakes, line 23: Added assignment here, so the property could be marked as non-null.

Plugins/GitUIPluginInterfaces/CredentialsManager.cs
```diff
@@ -36,18 +37,23 @@ public void Save()
 
             Credentials.Clear();
 
-            foreach (var networkCredentials in credentials.Where(c => c.Value is not null))
+            foreach (var networkCredentials in credentials)
             {
+                if (networkCredentials.Value is null)
+                {
+                    continue;
+                }
```
drewnoakes, line 45: Nullability doesn't flow well through linq. Make this explicit here.

Plugins/GitUIPluginInterfaces/GitUIPluginInterfaces.csproj
```diff
@@ -1,10 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
```
drewnoakes, line 4: Enable at the project level.

Plugins/GitUIPluginInterfaces/Settings/ChoiceSetting.cs
```diff
@@ -50,11 +50,11 @@ public override ComboBox CreateControl()
 
             public override void LoadSetting(ISettingsSource settings, ComboBox control)
             {
-                string settingVal = settings.SettingLevel == SettingLevel.Effective
+                string? settingVal = settings.SettingLevel == SettingLevel.Effective
                     ? Setting.ValueOrDefault(settings)
                     : Setting[settings];
 
-                control.SelectedIndex = Setting.Values.IndexOf(settingVal);
+                control.SelectedIndex = settingVal is null ? -1 : Setting.Values.IndexOf(settingVal);
```
drewnoakes, line 57: Bug fix. `IndexOf(null)` will throw.

GitUI/UserControls/RevisionGrid/RevisionGridControl.RevisionGridInMemFilter.cs
gerhardol: change seem OK

GitCommands/Remotes/ConfigFileRemoteSettingsManager.cs
gerhardol: agree

gerhardol (commented):
+1
No deep review and no test run, but am confident enough to approve
Let someone else review too.

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
```diff
@@ -409,7 +409,7 @@ private void Clone(IHostedRepository repo)
 
             var module = new GitModule(targetDir);
 
-            if (addUpstreamRemoteAsCB.Text.Trim().Length > 0 && !string.IsNullOrEmpty(repo.ParentReadOnlyUrl))
+            if (addUpstreamRemoteAsCB.Text.Trim().Length > 0 && !GitExtensions.Strings.IsNullOrEmpty(repo.ParentReadOnlyUrl))
```
RussKie, line 412:
```suggestion
            if (addUpstreamRemoteAsCB.Text.Trim().Length > 0 && !Strings.IsNullOrEmpty(repo.ParentReadOnlyUrl))
```

GitUI/Shells/ConEmuControlExtensions.cs
```diff
@@ -7,9 +7,9 @@ namespace GitUI.Shells
 {
     public static class ConEmuControlExtensions
     {
-        public static void ChangeFolder(this ConEmuControl terminal, IShellDescriptor shell, string path)
+        public static void ChangeFolder(this ConEmuControl? terminal, IShellDescriptor? shell, string? path)
         {
-            if (terminal?.RunningSession is null || shell is null || string.IsNullOrWhiteSpace(path))
+            if (terminal?.RunningSession is null || shell is null || GitExtensions.Strings.IsNullOrWhiteSpace(path))
```
RussKie, line 12:
```suggestion
            if (terminal?.RunningSession is null || shell is null || Strings.IsNullOrWhiteSpace(path))
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.RevisionGridInMemFilter.cs
```diff
@@ -66,8 +66,8 @@ static bool CheckCondition(string? filter, Regex? regex, string? value)
             {
                 if (string.IsNullOrEmpty(authorFilter) &&
                     string.IsNullOrEmpty(committerFilter) &&
-                    string.IsNullOrEmpty(messageFilter) &&
-                    !ObjectId.IsValidPartial(messageFilter, minLength: 5))
+                    (GitExtensions.Strings.IsNullOrEmpty(messageFilter) ||
```
RussKie, line 69:
```suggestion
                    (Strings.IsNullOrEmpty(messageFilter) ||
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.RevisionGridInMemFilter.cs
RussKie: Can we please have this change in a separate commit for ease of tracking/bisecting, if that becomes necessary.

Plugins/GitUIPluginInterfaces/CredentialsManager.cs
RussKie: This way is faster too :)

Plugins/GitUIPluginInterfaces/ObjectId.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
```
RussKie, line 1: @drewnoakes could you see some time, if this could be turned into a struct?

Plugins/GitUIPluginInterfaces/Settings/ChoiceSetting.cs
RussKie: Please move it own commit as well

RussKie (commented):
LGTM 👍

GitUI/CommandsDialogs/RepoHosting/ForkAndCloneForm.cs
drewnoakes: The `Strings` utility method has a naming conflict with a `Strings` resource dictionary, hence the fully qualified name here and elsewhere. In future perhaps we can rename one of the files to avoid this.

Plugins/GitUIPluginInterfaces/CredentialsManager.cs
drewnoakes: It can make a difference on a hot path, especially on enumeration and especially if the value being foreached has a struct enumerator. I doubt we have many credendials here though.

GitUI/UserControls/RevisionGrid/RevisionGridControl.RevisionGridInMemFilter.cs
drewnoakes: 0981ff16676be8461c3e8e80050630857ff6f0d1

Plugins/GitUIPluginInterfaces/Settings/ChoiceSetting.cs
drewnoakes: 4c4a428aa87fcee9c0687e84ca929321e085c0bc

PR #8881, opened 2021-02-24
Fix submodule status loading error

drewnoakes:
Fixes @gerhardol's comments on #8875

## Proposed changes

Refactoring during annotation identified a potential issue here, but the change was not correct. This change allows the task itself to be null (an uncommon pattern) which indicates that `SetSubmoduleStatus` has not been called.

## Test methodology 

- Manual testing

To trigger the broken behaviour:
1. Open the GE repo
2. Select the "File Tree" tab
3. Select a submodule in the tree
4. Press H to open the history

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1947,14 +1947,10 @@ namespace GitUI.CommandsDialogs
                         unstagedFiles.RemoveAll(
                             item =>
                             {
-                                if (!item.IsSubmodule)
-                                {
-                                    return false;
-                                }
-
-                                Task<GitSubmoduleStatus?> statusTask = item.GetSubmoduleStatusAsync();
-
-                                if (!statusTask.IsCompleted)
+                                if (!item.IsSubmodule
+                                    || item.GetSubmoduleStatusAsync() is not Task<GitSubmoduleStatus> statusTask
+                                    || statusTask is null
+                                    || !statusTask.IsCompleted)
                                 {
                                     return false;
                                 }
@@ -1970,7 +1966,10 @@ namespace GitUI.CommandsDialogs
                                 continue;
                             }
 
-                            GitSubmoduleStatus? gitSubmoduleStatus = ThreadHelper.JoinableTaskFactory.Run(() => item.GetSubmoduleStatusAsync());
+                            GitSubmoduleStatus? gitSubmoduleStatus = ThreadHelper.JoinableTaskFactory.Run(() =>
+                            {
+                                return item.GetSubmoduleStatusAsync() ?? Task.FromResult<GitSubmoduleStatus?>(null);
+                            });
 
                             if (gitSubmoduleStatus is null || !gitSubmoduleStatus.IsDirty)
                             {
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -112,10 +112,11 @@ namespace GitUI
                         : diffOfConflict;
                 }
 
-                if (file.IsSubmodule)
+                if (file.IsSubmodule
+                    && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
                 {
                     // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync);
+                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
                     return status is not null
                         ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                         : $"Failed to get status for submodule \"{file.Name}\"";
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -780,14 +780,13 @@ namespace GitUI
 
         private static string AppendItemSubmoduleStatus(string text, GitItemStatus item)
         {
-            if (item.IsSubmodule)
+            if (item.IsSubmodule
+                && item.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task
+                && task is not null
+                && task.IsCompleted
+                && task.CompletedResult() is not null)
             {
-                Task<GitSubmoduleStatus?> task = item.GetSubmoduleStatusAsync();
-
-                if (task.IsCompleted && task.CompletedResult() is not null)
-                {
-                    text += task.CompletedResult()!.AddedAndRemovedString();
-                }
+                text += task.CompletedResult()!.AddedAndRemovedString();
             }
 
             return text;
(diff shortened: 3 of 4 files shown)


GitUI/GitUIExtensions.cs
```diff
@@ -115,10 +115,15 @@ void OpenWithDiffTool()
                 if (file.IsSubmodule)
                 {
                     // Patch already evaluated
```
gerhardol, line 117: Move comment to inside `if (task`

or rather move all submodule handling here

something like
````c#
if (task is null)
{
    var patch = GetItemPatch(fileViewer.Module, file, firstId, selectedId,
                    fileViewer.GetExtraDiffArguments(), fileViewer.Encoding);
    return LocalizationHelpers.ProcessSubmodulePatch(fileViewer.Module, file.Name, patch);
}
else
{
    // Patch already evaluated
    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
    return status is not null
                            ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                            : $"Failed to get status for submodule \"{file.Name}\"";
                    }
```

gerhardol (commented):
quick check only

GitUI/GitUIExtensions.cs
gerhardol: Some good use of pattern matching declarations...

```c#
                if (file.IsSubmodule)
                {
                    if (file.GetSubmoduleStatusAsync() is not Task<GitSubmoduleStatus> task)
                    {
                        var diff = GetItemPatch(fileViewer.Module, file, firstId, selectedId,
                            fileViewer.GetExtraDiffArguments(), fileViewer.Encoding);
                        return LocalizationHelpers.ProcessSubmodulePatch(fileViewer.Module, file.Name, diff);
                    }

                    // Patch already evaluated
                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
                    return status is not null
                        ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                        : $"Failed to get status for submodule \"{file.Name}\"";
                }
```

gerhardol (commented):
I pushed a proposal to a local branch, using pattern matching declarations where possible and find the code much more easy to read

GitUI/UserControls/FileStatusList.cs
```diff
@@ -782,9 +782,9 @@ private static string AppendItemSubmoduleStatus(string text, GitItemStatus item)
         {
             if (item.IsSubmodule)
             {
-                Task<GitSubmoduleStatus?> task = item.GetSubmoduleStatusAsync();
+                Task<GitSubmoduleStatus?>? task = item.GetSubmoduleStatusAsync();
 
-                if (task.IsCompleted && task.CompletedResult() is not null)
+                if (task != null && task.IsCompleted && task.CompletedResult() is not null)
```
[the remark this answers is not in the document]
drewnoakes: Looks good. Please don't hesitate to push to this PR, or create a PR against this branch.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1952,8 +1952,8 @@ private void Stage(IReadOnlyList<GitItemStatus> items)
                                     return false;
                                 }
 
-                                Task<GitSubmoduleStatus?> statusTask = item.GetSubmoduleStatusAsync();
-
+                                Task<GitSubmoduleStatus?>? statusTask = item.GetSubmoduleStatusAsync();
+                                Validates.NotNull(statusTask);
                                 if (!statusTask.IsCompleted)
```
RussKie, line 1957: I'm struggling with this - `statusTask` can be `null`, and in case when it is will it throw on line:1956 or line:1957?

gerhardol: I rewrote most of the checks using pattern matching declarations, I find this much easier to read

GitUI/UserControls/FileStatusList.cs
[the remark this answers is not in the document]
gerhardol: I could not create a PR to your repo, so I pushed the commit

GitUI/UserControls/FileStatusList.cs
```diff
@@ -804,7 +803,9 @@ public async Task OpenSubmoduleAsync()
 
             var submoduleName = SelectedItem.Item.Name;
 
-            var status = await SelectedItem.Item.GetSubmoduleStatusAsync().ConfigureAwait(false);
+            Task<GitSubmoduleStatus?>? task = SelectedItem.Item.GetSubmoduleStatusAsync();
+            Validates.NotNull(task);
```
gerhardol, line 807: This could probably be rewritten to not validate, but it got too complicated to bother...

GitUI/UserControls/FileStatusList.cs
```diff
@@ -924,14 +925,17 @@ private void UpdateFileStatusListView(bool updateCausedByFilter = false)
                         listItem.ImageIndex = GetItemImageIndex(item);
                     }
 
-                    if (item.GetSubmoduleStatusAsync() is not null && !item.GetSubmoduleStatusAsync().IsCompleted)
+                    if (item.IsSubmodule
+                        && item.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task
+                        && task is not null
+                        && !task.IsCompleted)
```
gerhardol, line 931: need to remove this check to set status om L942

gerhardol (changes requested):

GitUI/UserControls/FileStatusList.cs
gerhardol: Removed the Validates (if task or status is null there are no commits to select in the subrepo)

GitUI/UserControls/FileStatusList.cs
gerhardol: removed the check in a separate push

gerhardol (approved):

GitUI/UserControls/FileStatusList.cs
```diff
@@ -924,19 +927,21 @@ private void UpdateFileStatusListView(bool updateCausedByFilter = false)
                         listItem.ImageIndex = GetItemImageIndex(item);
                     }
 
-                    if (item.GetSubmoduleStatusAsync() is not null && !item.GetSubmoduleStatusAsync().IsCompleted)
+                    if (item.IsSubmodule
+                        && item.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task
+                        && task is not null)
```
drewnoakes, line 932: @gerhardol the type check is essentially a null check too.

```suggestion
                        && item.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
```

drewnoakes: Another way of writing this is:

```suggestion
                        && item.GetSubmoduleStatusAsync() is { } task)
```

But is that actually better than assigning the task to a local and doing an explicit `is not null` test? The only downside is a bit of nesting, which is not the end of the world.

mstv: We had this discussion before: This is a hard to read. One does not need to use every new C# syntax invention.

gerhardol: The `task is not null` checks in FileStatusList.cs are not needed.
I changed to that style as it is required in FormCommit

I prefer to have a minimal code why my preference is to just removing is not null where possible.

RussKie (approved):

PR #8882, opened 2021-02-24
Avoid boxing while unpacking flags

drewnoakes:
## Proposed changes

Eliminate boxing when in `GitItemStatus` bool property getters.

`Enum.HasFlag` boxes its argument. Replace it with a method that does not.

## Test methodology 

- Manual testing

----

### DIFF
--- a/GitCommands/Git/GitItemStatus.cs
+++ b/GitCommands/Git/GitItemStatus.cs
@@ -65,13 +65,13 @@ namespace GitCommands
 
         public bool IsTracked
         {
-            get => _flags.HasFlag(Flags.IsTracked);
+            get => HasFlag(Flags.IsTracked);
             set => SetFlag(value, Flags.IsTracked);
         }
 
         public bool IsDeleted
         {
-            get => _flags.HasFlag(Flags.IsDeleted);
+            get => HasFlag(Flags.IsDeleted);
             set => SetFlag(value, Flags.IsDeleted);
         }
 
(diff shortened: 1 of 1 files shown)


GitCommands/Git/GitItemStatus.cs
```diff
@@ -159,10 +159,16 @@ public bool IsStatusOnly
...
-            get => _flags.HasFlag(Flags.IsRangeDiff);
+            get => HasFlag(Flags.IsRangeDiff);
             set => SetFlag(value, Flags.IsRangeDiff);
         }
 
+        private bool HasFlag(Flags flags)
+        {
+            // NOTE Enum.HasFlag boxes its argument
+            return (flags & _flags) == flags;
+        }
```
mstv, line 170: BTW: The name of the enum `Flags` is misleading. It should be in singular.
I can read this better:
```suggestion
        private bool HasFlag(Flags flag)
        {
            // NOTE Enum.HasFlag boxes its argument
            return (_flags & flag) == flag;
        }
```

gerhardol (commented):
+1
Agree with mstv

GitCommands/Git/GitItemStatus.cs
drewnoakes: The logic works equally well with multiple flags. The type is a flags enum, and has a plural name.

RussKie (commented):
IIRC .NET Core has resolved this allocations, hasn't it? #8522 makes this change obsolete.

drewnoakes:
@RussKie what change are you looking for here?

gerhardol (approved):

gerhardol:
> @RussKie what change are you looking for here?

I believe it was a comment that the change is not necessary as .NET5 should be coming soon
Suggest this is merged now, an improvement regardless

GitCommands/Git/GitItemStatus.cs
mstv: > The logic works equally well with multiple flags.

This was clear to me. But `Enum.HasFlag` uses only singular. And nothing else is used here.

Better readability (as in other recent PRs):
```suggestion
        private bool HasFlag(Flags flags)
        {
            // NOTE Enum.HasFlag boxes its argument
            return (_flags & flags) == flags;
        }
```

RussKie (approved):

PR #8883, opened 2021-02-24
Use pattern matching

drewnoakes:
## Proposed changes

- Use C# 9 pattern matching in a few places

## Test methodology 

- Manual testing and review

----

### DIFF
--- a/GitUI/CommandsDialogs/FormPush.cs
+++ b/GitUI/CommandsDialogs/FormPush.cs
@@ -1282,10 +1282,10 @@ namespace GitUI.CommandsDialogs
             SetBranchesPushCheckboxesState(row =>
                 {
                     // Check if the branch is tracked (i.e. not new)
-                    var localColumn = row.Cells[LocalColumn.Name] as DataGridViewTextBoxCell;
-                    var remoteColumn = row.Cells[RemoteColumn.Name] as DataGridViewTextBoxCell;
-                    return localColumn is not null && remoteColumn is not null
-                           && !string.IsNullOrEmpty(localColumn.Value.ToString()) && !string.IsNullOrEmpty(remoteColumn.Value.ToString());
+                    return row.Cells[LocalColumn.Name] is DataGridViewTextBoxCell localColumn &&
+                           row.Cells[RemoteColumn.Name] is DataGridViewTextBoxCell remoteColumn &&
+                           !string.IsNullOrEmpty(localColumn.Value.ToString()) &&
+                           !string.IsNullOrEmpty(remoteColumn.Value.ToString());
                 });
         }
 
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -739,7 +739,7 @@ namespace GitUI.CommitInfo
 
         private void RichTextBox_KeyDown(object sender, KeyEventArgs e)
         {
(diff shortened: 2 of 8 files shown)


GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -740,8 +739,7 @@ protected override void OnLayout(LayoutEventArgs e)
 
         private void RichTextBox_KeyDown(object sender, KeyEventArgs e)
         {
-            var rtb = sender as RichTextBox;
-            if (rtb is null || !e.Control || e.KeyCode != Keys.C)
+            if (sender is not RichTextBox rtb || !e.Control || e.KeyCode != Keys.C)
```
gerhardol, line 742: I guess that the pattern matching declaration style is something I have to get used to. In many situations it is probly better too.
I do not like declarations in is not checks though.

gerhardol (approved):

GitUI/CommitInfo/CommitInfo.cs
drewnoakes: > I do not like declarations in is not checks though.

It took me a while to get used to it. It's clearer than the alternative, in my opinion, such as:

```c#
if (!(sender is RichTextBox rtb) || ...)
```

RussKie: How about this?
```
if (!e.Control || e.KeyCode != Keys.C || sender is not RichTextBox rtb)
```

RussKie (commented):
LGTM

GitUI/CommitInfo/CommitInfo.cs
gerhardol: that order is better for my brain, it is kind of the last result that survives the if()

RussKie: @drewnoakes can you please reorder (few places), then it is good to merge.
Thanks

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -1282,10 +1282,10 @@ private void selectTrackedToolStripMenuItem_Click(object sender, EventArgs e)
...
                 {
                     // Check if the branch is tracked (i.e. not new)
-                    var localColumn = row.Cells[LocalColumn.Name] as DataGridViewTextBoxCell;
-                    var remoteColumn = row.Cells[RemoteColumn.Name] as DataGridViewTextBoxCell;
-                    return localColumn is not null && remoteColumn is not null
-                           && !string.IsNullOrEmpty(localColumn.Value.ToString()) && !string.IsNullOrEmpty(remoteColumn.Value.ToString());
+                    return row.Cells[LocalColumn.Name] is DataGridViewTextBoxCell localColumn &&
+                           row.Cells[RemoteColumn.Name] is DataGridViewTextBoxCell remoteColumn &&
+                           !string.IsNullOrEmpty(localColumn.Value.ToString()) &&
+                           !string.IsNullOrEmpty(remoteColumn.Value.ToString());
```
mstv, line 1288: Left Reading works better with operators at the begin of the next lines.

drewnoakes: Is there an analyzer we could use to control this uniformly across the code?

RussKie: I'm not aware of any. 
@sharwell is the Master of Analyzers, maybe he knows :)

PR #8887, opened 2021-02-25
Eliminate closure allocations

drewnoakes:
## Proposed changes

This change removes 427,688 bytes of closure allocations while loading GE with the gitextensions repo.

Passing the argument on the stack allows the delegate to be cached by the compiler, rather than allocating a new instance for each call.

## Test methodology 

- Manual testing

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -219,7 +219,7 @@ namespace GitUI.UserControls.RevisionGrid.Graph
             }
 
             // Ensure all parents are loaded before adding it to the _nodes list. This is important for ordering.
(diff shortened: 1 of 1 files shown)


GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -219,7 +219,7 @@ public void Add(GitRevision revision, RevisionNodeFlags types)
             }
 
             // Ensure all parents are loaded before adding it to the _nodes list. This is important for ordering.
-            ImmutableInterlocked.Update(ref _nodes, list => list.Add(revisionGraphRevision));
+            ImmutableInterlocked.Update(ref _nodes, (list, rev) => list.Add(rev), revisionGraphRevision);
```
RussKie, line 222:
```suggestion
            ImmutableInterlocked.Update(ref _nodes, (list, revision) => list.Add(revision), revisionGraphRevision);
```

RussKie (approved):

PR #8888, opened 2021-02-25
Eliminate boxing allocation

drewnoakes:
## Proposed changes

This change removes 537,404 bytes of boxed enum allocations while loading GE with the gitextensions repo.

## Test methodology 

- Manual testing
 
----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8889, opened 2021-02-25
[3.5] Eliminate boxing allocation

drewnoakes:
Port of #8888 for `release/3.5`.

This change removes 537,404 bytes of boxed enum allocations while loading GE with the gitextensions repo.


gerhardol (commented):
+1

PR #8886, opened 2021-02-24
GitStatusMonitor increase minimum time between updates

gerhardol:
Fixes #8871

## Proposed changes

Increase the time between git-status runs from 5s to 30s

So for instance when building, git-status runs every 30s

If there has been no changes and git-status has not run in 30s,
git-status will run in about one second as now

This affects:
* Commit button count
* artificial commit summary and tooltip
* left panel status and tooltips (added in 3.5)
* submodule menu drop down

The use case with most impact as I see it is if you save one file in the editor (git-status is triggered quickly), then you select another file in the editor already edited and save that, then the status will be updated in about 30s instead of about 5s.

## Test methodology 

manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8894, opened 2021-02-26
RevDiff: Request GitStatus updates at file manipulations

gerhardol:
https://github.com/gitextensions/gitextensions/pull/8886#issuecomment-786868520

## Proposed changes

Whenever files are manipulated in RevDiff, request GitStatusMonitor refresh
- If the change was just worktree <-> index, git-status is not triggered at all (until next file system change or every fifth minute)
- If the change is throttled (#8886 increases the minimal time for updates to 30 s), an interactive request is done (delayed 200 ms to possibly bundle changes)

I considered to request the update only at (likely) worktree <-> index changes, but that got more complicated and a faster update is desired after manual changes.
If you stage files one by one, if may cause git-status to run more or less continously (where Git is slow). I do not believe this is a problem, but the request could be done with some delay too.
@mstv?

Note that FormCommit is not affected, it always rescans (not directly with git-status) when a change is done.

Candidate for 3.5 too 

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -201,7 +201,8 @@ namespace GitUI.CommandsDialogs
 
             repoObjectsTree.Initialize(_aheadBehindDataProvider, _filterBranchHelper, RevisionGrid, RevisionGrid, RevisionGrid);
             toolStripBranchFilterComboBox.DropDown += toolStripBranches_DropDown_ResizeDropDownWidth;
-            revisionDiff.Bind(RevisionGrid, fileTree);
+            revisionDiff.Bind(RevisionGrid, fileTree, () => RequestRefresh());
+            fileTree.Bind(() => RequestRefresh());
 
             var repositoryDescriptionProvider = new RepositoryDescriptionProvider(new GitDirectoryResolver());
             _appTitleGenerator = new AppTitleGenerator(repositoryDescriptionProvider);
@@ -813,7 +814,7 @@ namespace GitUI.CommandsDialogs
             }
 
             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
-            _gitStatusMonitor.RequestRefresh();
+            RequestRefresh();
 
             if (_dashboard is null || !_dashboard.Visible)
             {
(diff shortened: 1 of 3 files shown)


mstv (approved):
Thank you! Feels much more responsive.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -382,6 +384,20 @@ private ContextMenuSelectionInfo GetSelectionInfo()
...
+        private void RequestRefresh()
+        {
+            // Request immediate update of commit count, no delay due to backoff
+            // If a file system change was triggered too, the requests should be merged
+            // (this will also update the count if only worktree<->index is changed)
+            // This may trigger a second RefreshArtificial()
+            if (_refreshGitStatus is not null)
+            {
+                _refreshGitStatus();
+            }
```
RussKie, line 396:
```suggestion
            _refreshGitStatus?();
```
or `_refreshGitStatus?.Invoke();`

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -825,6 +825,11 @@ private void RefreshRevisions()
             toolStripButtonPush.DisplayAheadBehindInformation(Module.GetSelectedBranch());
         }
 
+        private void RequestRefresh()
+        {
+            _gitStatusMonitor.RequestRefresh();
+        }
```
RussKie, line 831:
```suggestion
        private void RequestRefresh() => _gitStatusMonitor.RequestRefresh();
```

RussKie (approved):
👍 with minor suggestions

GitUI/CommandsDialogs/RevisionDiffControl.cs

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: ok

gerhardol:
review comments and a similar changes (few) for revfile, browse. (I ignored editing of .gitignore and similar, there are maybe other situations too that will be affected by edit+edit if #8886 is included).

Tested with patched GitStatusMonitor so it runs continuously, works OK. The UI flashes a little but that is separate.

Will squash for merge in about a day (or at feedback).

mstv (approved):
👍

gerhardol:
A scenario that is not handled is if you update a submodule, without any filesystem changes (for example for a merge commit).
RevDiff could be handled, but more difficult for the left panel.
Ignored for now.

gerhardol:
@msftbot merge in 24 hours

PR #8897, opened 2021-02-27
Refactor PaintGraphCell

mstv:
## Proposed changes

- Always restore the clip region and the `SmoothingMode`
- Remove return value which is always `true`
- Reduce `if` nesting
- Replace `var`

## Screenshots 

N/A

## Test methodology 

- review, manual and existing tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build 11f8f3ec566658edbf1a41d166bc7a4144024f1c
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
@@ -106,7 +106,7 @@ namespace GitUI.UserControls.RevisionGrid.Columns
                     }
                 }
 
-                var newRows = _graphCache.Count < _graphCache.Capacity
+                int newRows = _graphCache.Count < _graphCache.Capacity
                     ? (rowIndex - _graphCache.Count) + 1
                     : 0;
 
@@ -143,17 +143,13 @@ namespace GitUI.UserControls.RevisionGrid.Columns
                     return true;
                 }
 
-                if (!DrawVisibleGraph())
-                {
-                    return false;
-                }
-
+                DrawVisibleGraph();
                 CreateRectangle();
                 return true;
 
                 void CreateRectangle()
                 {
-                    var cellRect = new Rectangle(
+                    Rectangle cellRect = new(
                         0,
                         ((_graphCache.HeadRow + rowIndex - _graphCache.Head) % _graphCache.Capacity) * rowHeight,
                         width,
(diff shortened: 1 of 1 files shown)


RussKie:
Seems to work 👍

mstv:
@drewnoakes, could you review, please?

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
```diff
@@ -199,52 +198,51 @@ bool DrawVisibleGraph()
...
+                    SmoothingMode oldSmoothingMode = g.SmoothingMode;
                     Region oldClip = g.Clip;
-                    var newClip = new Region(laneRect);
+
+                    int top = g.RenderingOrigin.Y;
+                    Rectangle laneRect = new(0, top, width, rowHeight);
+                    Region newClip = new(laneRect);
                     newClip.Intersect(oldClip);
                     g.Clip = newClip;
                     g.Clear(Color.Transparent);
```
drewnoakes, line 217: Why do we need to use `Clear` vs. `FillRegion`? The latter would avoid needing to set/restore the `Clip`.

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
```diff
@@ -199,52 +198,51 @@ bool DrawVisibleGraph()
...
-                    {
-                        return true;
-                    }
-
-                    // Getting RevisionGraphDrawStyle results in call to AppSettings. This is not very cheap, cache.
-                    _revisionGraphDrawStyleCache = RevisionGraphDrawStyle;
+                    DrawItem();
 
-                    var oldSmoothingMode = g.SmoothingMode;
+                    // Reset graphics options
```
drewnoakes, line 221:
```suggestion
                    // Restore graphics options
```

drewnoakes (approved):
Looks good.

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
mstv: :+1 done

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
mstv: `Clip` is needed anyway in order to create the impression of drawn through graph lines by not letting end the lines at the row borders.

PR #8899, opened 2021-02-27
Fix tool strip border in dark theme

### LINKED ISSUE #8864: [Theming] ToolStrip visual artifacts
## Current behaviour

| Default | Dark |
--|--
image | image

In the default (system) theme the toolstrip is uniformly filled, and all borders are square. In the dark theme - the both ends show visual artifacts, that look like a rounded shadow.
image

## Steps to reproduce

Checkout 983b0c3 (or the commit from #8849)

NikolayXHD:
Fixes #8864

### Before

image
image

### After

image
image

## Proposed changes

At the moment there is a couple of cases where a custom `ToolStrip.Renderer` solves a task not related to dark theme.
1. emulate context menu "caption" item, which serves as an item group header
image
image
2. remove ToolStrip borders (`ToolStripEx` class)

Current implementation of dark theme does not consider the cases above. It does
```
toolStrip.Renderer = new ThemeAwareToolStripRenderer();
```
to all toolstrips, which breaks the cases 1. 2. above.

This PR does such that if a ToolStrip  needs custom rendering, it uses on of 2 predefined renderers,
`ToolStripExRenderer` or `ToolStripExThemeAwareRenderer`, both of which support scenarios 1. and 2. from above.

## Why not use 1 renderer per 1 solved problem?

Because `ToolStripRenderer` design makes it impossible.

It offers 2 ways of customization:

- subscribe to an event such as `RenderMenuItemBackground`
- override a virtual method such as `OnRenderMenuItemBackground`

The event mechanism fails because it cannot suppress default painting, except the few cases when ther is an `eventArgs.Handled` field.

The inheritance mechanism does not allow you to compose different renderers so that each is responsible for solving 1 problem.

Amusingly, `ToolStripRenderer` does offer an ok way to compose renderers, which is a virtual property `ToolStripRenderer.RendererOverride`, but guess what, it's internal.

Every time I stumble upon `virtual internal` pattern in a library code, I cannot help feel that library developer considers other developers, especially me, second class.

## Test environment(s)

- Windows10
- default dpi, 100%

----


RussKie:
Awesome work!

I pushed few tweaks to API names, reordered methods, etc.

ivangrek (approved):
👍

PR #8906, opened 2021-02-28
Show error message if checkout fails

### LINKED ISSUE #8632: Show error message when Reset file fails
## Feature description

Related to #8630 

Due to the missing file, I tryied to reset the file to get it.

Of course, git fails but Git Extensions don't complain.

I suggest that GitModule.CheckoutFiles should return a bool and return the result of ``_gitExecutable.RunBatchCommand``. So caller (like ``RevisionDiffControl.ResetSelectedItemsTo``) should be able to display a MessageBox.

Are you okay before I implement it ?

## Environment

- Git Extensions 33.33.33
- Build 3737099926a138c34433f6cd1de29b5d9580968f
- Git 2.22.0.windows.1 (recommended: 2.28.0 or later)
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

bansan85:
Fixes #8632 

It's a WIP to have first feedback. In the UI message, I can't get either executable nor working directory.

## Proposed changes

- throw an ExternalOperationException exception instead of ignoring the return value of Module.CheckoutFiles

## Screenshot

image

### Before

If git checkout fails, GE does not say a thing.

### After

If git checkout fails, GE does say a thing.

## Test methodology 

- Update test : yes.

----


vbjay:
Looks like a file lock issue. Pointing out that when appropriate would be a
good idea. Maybe even gathering lock process and saying it is open in
notepad or whatever.

bansan85:
> Looks like a file lock issue. Pointing out that when appropriate would be a good idea. Maybe even gathering lock process and saying it is open in notepad or whatever.

In this very special case, the filename is the problem. On Windows, you can't create filename "aux". But you're right, the most common case is a file lock issue.

mstv:
> GE does not say a thing.

My words, #8802.

bansan85:
@mstv I created this pull request because #8278 and #8358 (ref in #8632) was closed. If #8802 may impact this PR, I will wait until its closure and see if it solves my issue.

gerhardol:
@bansan85 You are welcome to review @8802 to speed up inclusion of that PR (and to see that your usecase is handled).

bansan85:
@gerhardol I don't know how to submit a change to a pull request. So I rebased my changes with pull #8802

bansan85:
At this step, it will show (in French):

image

I'm not very fan of the message but at least, GE is not mute anymore.

vbjay:
You just push more commits to the same branch.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -200,9 +199,12 @@ string ComposeOutput()
 
                 // Invoke batch progress callback
                 action?.Invoke(new BatchProgressEventArgs(item.BatchItemsCount, result));
-            }
 
-            return result;
+                if (!result)
+                {
+                    throw new ExternalOperationException(arguments: item.Argument);
```
mstv, line 205: One can specify an error message by providing an inner exception:
`throw new ExternalOperationException(arguments: item.Argument, innerException: new Exception("Git command failed."));`

Open items:
1. There are ongoing discussions regarding localization (in non-UI modules).
2. Errors from git commands should be reported centrally #6527.

UnitTests/GitCommands.Tests/Git/ExecutableExtensionsTests.cs
```diff
@@ -108,12 +108,21 @@ public void GetOutput_with_cache_miss()
...
             {
                 Assert.IsTrue(eventArgs.ExecutionResult);
                 Assert.AreEqual(expectedProcessedCounts[index], eventArgs.ProcessedCount);
                 index++;
-            }));
+            });
+
+            if (expectedResult)
+            {
+                function();
```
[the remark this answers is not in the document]
bansan85: There is no need. Now _gitExecutable.RunBatchCommand don't return anything. An exception is thrown instead of using the return value.

GitCommands/Git/ExecutableExtensions.cs
bansan85: Good idea. With an inner exception with the output of stderr is definitely better.

bansan85: image

vbjay:
If that is the case, you need to remove the index lock. Maybe a task button
that lets you remove it with a reasonable warning on doing that. Something
like try again first after waiting. If the index pick is still there then
you can remove it by ...

bansan85:
> 
> If that is the case, you need to remove the index lock. Maybe a task button that lets you remove it with a reasonable warning on doing that. Something like try again first after waiting. If the index pick is still there then you can remove it by ...

The problem is that I need to inspect the string. As I said before, I can have other problem like:

image

(aux filename is forbidden in Windows).

My main problem was that GE was mute.

Of course I could inspect the string and add a button to the error message to remove the index.lock. But is it really necessary ? @mstv

mstv:
The result looks good. Perhaps `operation: "git"` should be added. I Will review the code in the next days.

> Of course I could inspect the string and add a button to the error message to remove the index.lock. But is it really necessary?

IMO it suffices to show the error message from git and let the user handle it.
Especially a dangling index.lock is a rare issue (one or two times ever).

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -211,21 +216,28 @@ string ComposeOutput()
...
             if (input is not null)
             {
                 await process.StandardInput.BaseStream.WriteAsync(input, 0, input.Length);
                 process.StandardInput.Close();
             }
 
+            if (error is not null)
+            {
+                await process.StandardError.BaseStream.ReadAsync(error, 0, error.Length);
+            }
```
bansan85, line 239: I can't ``process.StandardError.Close();`` either tests crashes.

mstv: Sorry for the delay.
The output should be read *after* the process has finished / terminated, i.e. in `RunCommandAsync` after `await process.WaitForExitAsync()`.
As suggested in #6527, an argument `bool throwOnErrorOutput = true` should be added to `RunCommandAsync`. If `true`, the error output of the command shall be read and if not empty, be wrapped into a new `CommandErrorException` derived from `ExternalOperationException`. This exception does only need to be handled up the callstack unless the remaining output can be displayed despite the exception.

bansan85: I fixed the order to read the error output stream.
I suggest to leave throwOnErrorOutput to #6527 to reduce size and age of commit.
At the current time, my commit is based on #8802 and the original branch can't be merged anymore (BugReporter has been heavily refactored). I will try to rebase my commit based on master and not on #8802 if it's possible.

PR #8928, opened 2021-03-05
Use null propagation

drewnoakes:
## Proposed changes

- Use null propagation

## Test methodology 

- Manual review. It compiles. CI will run tests.

----

### DIFF
--- a/ResourceManager/Xliff/TranslationUtil.cs
+++ b/ResourceManager/Xliff/TranslationUtil.cs
@@ -197,12 +197,7 @@ namespace ResourceManager.Xliff
 
         public static void TranslateProperty(string category, object obj, string propName, ITranslation translation)
         {
-            if (obj is null)
-            {
-                return;
-            }
-
-            var property = obj.GetType().GetProperty(propName, _propertyFlags);
+            var property = obj?.GetType().GetProperty(propName, _propertyFlags);
 
             if (property is null)
             {
--- a/GitUI/CommandsDialogs/FormClone.cs
+++ b/GitUI/CommandsDialogs/FormClone.cs
@@ -467,10 +467,7 @@ namespace GitUI.CommandsDialogs
             {
                 _branchListLoader.Dispose();
 
-                if (components is not null)
-                {
-                    components.Dispose();
-                }
+                components?.Dispose();
             }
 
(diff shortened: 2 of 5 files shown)


ResourceManager/Xliff/TranslationUtil.cs
```diff
@@ -197,12 +197,7 @@ public static void TranslateItemsFromList(string category, ITranslation translat
...
         {
-            if (obj is null)
-            {
-                return;
-            }
-
-            var property = obj.GetType().GetProperty(propName, _propertyFlags);
+            var property = obj?.GetType().GetProperty(propName, _propertyFlags);
 
             if (property is null)
```
ivangrek, line 202: It looks like a case.

drewnoakes: Do you mean a switch? I'm not sure what you mean exactly, sorry.

ivangrek: Designated case: "Use null propagation"

ivangrek: `property?.SetValue(obj, value, null);`

drewnoakes: Thanks, I see what you mean now.

gerhardol (approved):
OK, just agree with ivangrek

ResourceManager/Xliff/TranslationUtil.cs
ivangrek: I found this option. But... 😄 

```
string value = translation.TranslateItem(category, propName, "Text", ProvideDefaultValue);

if (!string.IsNullOrEmpty(value))
{
    return;
}

var property = obj?.GetType()
    .GetProperty(propName, _propertyFlags);

if (property?.CanWrite is true)
{
    property.SetValue(obj, value, null);
}
```

drewnoakes: I'm not sure whether it's worth changing this. The current behaviour bails early when property is null, avoiding redundant work. Feel free to push to this branch if you'd like to see a change. For now, I can't see one that I like.

ivangrek (approved):

PR #8929, opened 2021-03-05
Remove redundant property setters

drewnoakes:
## Proposed changes

Remove redundant property setters. Read only properties have the stronger guarantee that their value will not change over time, making reasoning about the containing type's behaviour easier.

## Test methodology 

- Manual review. Couldn't find any late binding to these properties. CI will run tests.

----

### DIFF
--- a/GitCommands/Config/ConfigSection.cs
+++ b/GitCommands/Config/ConfigSection.cs
@@ -63,7 +63,7 @@ namespace GitCommands.Config
 
         public string SectionName { get; set; }
         public string? SubSection { get; set; }
-        public bool SubSectionCaseSensitive { get; set; }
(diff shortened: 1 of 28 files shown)


RussKie (commented):
Looks sensible from ☎️

thimmy687:
Hi,

this change breaks the Gerrit Plugin again :/

But I'm not totally sure if we can replace or remove it in the plugin, so maybe I can get a hint for this

 -> [PR 29](https://github.com/gitextensions/GitExtensions.GerritPlugin/pull/29)

drewnoakes:
@thimmy687 can you open a new issue with the stack trace of the failure? We may be able to preserve the old behaviour.

PR #8938, opened 2021-03-06
Stop AsyncLoader passing null on cancellation

drewnoakes:
## Proposed changes

Stop AsyncLoader passing null on cancellation. This strange pattern was identified during recent annotation work. It isn't particularly useful to be called back with a default value when the original load action was cancelled.

## Test methodology 

- Manual testing.

----

### DIFF
--- a/GitCommands/AsyncLoader.cs
+++ b/GitCommands/AsyncLoader.cs
@@ -30,20 +30,20 @@ namespace GitCommands
         public async Task LoadAsync(Action<CancellationToken> loadContent, Action onLoaded)
         {
             await LoadAsync(
-                (token) =>
+                token =>
                 {
                     loadContent(token);
-                    return string.Empty;
+                    return (object?)null;
                 },
                 _ => onLoaded());
         }
 
-        public Task<T?> LoadAsync<T>(Func<T> loadContent, Action<T?> onLoaded)
+        public Task<T?> LoadAsync<T>(Func<T> loadContent, Action<T> onLoaded)
         {
             return LoadAsync(token => loadContent(), onLoaded);
         }
 
-        public async Task<T?> LoadAsync<T>(Func<CancellationToken, T> loadContent, Action<T?> onLoaded)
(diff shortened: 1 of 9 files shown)


GitCommands/AsyncLoader.cs
```diff
@@ -73,13 +73,11 @@ public async Task LoadAsync(Action<CancellationToken> loadContent, Action onLoad
                 // Bail early if cancelled, returning default value for type
                 if (token.IsCancellationRequested)
                 {
-                    result = default;
-                }
-                else
-                {
-                    // Load content
-                    result = loadContent(token);
+                    return default;
```
drewnoakes, line 76: This is the key change. Now, if cancellation is requested, we just return a default value here. Previously we would still return a default value, but call `onLoaded` as well with that default value.

gerhardol: But this changes the behavior, previously the callers could detect the cancel scenario. In FileViewer for instance, no text would be shown, now an empty string will be displayed.
I believe the change is fine anyway.

gerhardol (commented):
+1
Default behavior is slightly changed, probably  not an issue

drewnoakes:
> Default behavior is slightly changed, probably not an issue

Yes, in the case of an operation being cancelled, slightly less work would be performed. I made a review of usages to try and detect any important behaviour changes but didn't find any.

RussKie (commented):
LGTM

GitCommands/AsyncLoader.cs
drewnoakes: I cannot see the behaviour change in FileViewer. Can you help me to see it? All the callbacks I see performed a no-op when passed null.

vbjay:
The idea is we should be able to walk the token anywhere and use that to
find out if it is canceled.

drewnoakes:
@vbjay that's what happens currently. There's no change in that regard within this PR.

PR #8942, opened 2021-03-06
FormBrowse Toolbar: Being able to hide each control separately

pmiossec:
Fixes "One or two comment asking for this feature but that I can't find anymore..."

Follow up of #8572

## Proposed changes

- Being able to choose the buttons in the browse form toolbar
- Preventing the closing of the menu to easily unselect multiple items

### Before

image

### After

image 
image

Gif (with not updated with last separator handling. See screenshot above)
toolbar_configuration

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 132668ee18374ce1ddccb8640e717ae8b70087a3 (Dirty)
- .NET Framework 4.8.4311.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
@@ -106,12 +106,148 @@ namespace GitUI.CommandsDialogs
                     CheckOnClick = true,
                     Tag = senderToolStrip
                 };
+
+                toolStripItem.DropDown.Closing += (sender, e) =>
+                {
+                    if (e.CloseReason == ToolStripDropDownCloseReason.ItemClicked)
+                    {
+                        // Cancel closing menu to allow selecting/unselecting multiple items.
+                        e.Cancel = true;
+                    }
+                };
+
+                CreateToolStripSubMenus(senderToolStrip, toolStripItem);
+
                 toolStripItem.Click += (s, e) =>
                 {
                     senderToolStrip.Visible = !senderToolStrip.Visible;
                 };
 
                 return toolStripItem;
+            }
+        }
+
+        private static void CreateToolStripSubMenus(ToolStrip senderToolStrip, ToolStripMenuItem toolStripItem)
+        {
+            const string toolbarSettingsPrefix = "formbrowse_toolbar_visibility_";
+            string? currentGroup = null;
+            foreach (ToolStripItem toolbarItem in senderToolStrip.Items)
+            {
+                if (toolbarItem is ToolStripSeparator)
+                {
+                    toolStripItem.DropDownItems.Add(new ToolStripSeparator());
+                    continue;
+                }
+
+                bool belongToAGroup = BelongToAGroup(toolbarItem, out string groupName);
+                string key;
+                if (belongToAGroup)
+                {
+                    if (currentGroup == groupName)
+                    {
+                        toolbarItem.Visible = LoadVisibilitySetting(groupName);
+                        continue;
+                    }
+
+                    currentGroup = (string)toolbarItem.Tag;
+                    key = groupName;
+                }
+                else
+                {
+                    key = toolbarItem.Name;
+                }
+
+                toolbarItem.Visible = LoadVisibilitySetting(key);
+
+                ToolStripMenuItem menuToolbarItem = new(toolbarItem.ToolTipText)
+                {
+                    Checked = toolbarItem.Visible,
+                    CheckOnClick = true,
+                    Tag = toolbarItem,
+                    Image = toolbarItem.Image
+                };
+
+                menuToolbarItem.Click += (s, e) =>
+                {
+                    toolbarItem.Visible = !toolbarItem.Visible;
+
+                    if (!BelongToAGroup(toolbarItem, out var group))
+                    {
+                        SaveVisibilitySetting(toolbarItem.Name, toolbarItem.Visible);
+                    }
+                    else
+                    {
+                        SaveVisibilitySetting(group, toolbarItem.Visible);
+                        foreach (ToolStripItem item in senderToolStrip.Items)
+                        {
+                            if (item.Tag == (object)group)
+                            {
+                                item.Visible = toolbarItem.Visible;
+                            }
+                        }
+                    }
+
+                    AdaptSeparatorsVisibility(senderToolStrip);
+                };
+
+                toolStripItem.DropDownItems.Add(menuToolbarItem);
+            }
+
+            AdaptSeparatorsVisibility(senderToolStrip);
+
+            return;
+
+            static void SaveVisibilitySetting(string key, bool visible) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible ? null : false);
+            static bool LoadVisibilitySetting(string key) => AppSettings.GetBool(toolbarSettingsPrefix + key, true);
+
+            static bool BelongToAGroup(ToolStripItem toolbarItem, out string groupName)
+            {
+                const string groupPrefix = "ToolBar_group:";
+                if (toolbarItem.Tag is string group && group.StartsWith(groupPrefix))
+                {
+                    groupName = group;
+                    return true;
+                }
+
+                groupName = string.Empty;
+                return false;
+            }
+        }
+
+        private static void AdaptSeparatorsVisibility(ToolStrip senderToolStrip)
+        {
+            // First pass: toolbar items from left to right
+            bool shouldHideNextSeparator = true;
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -104,8 +105,52 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+                        text = "<Separator>";
+                    }
+                    else
+                    {
+                        text = !string.IsNullOrWhiteSpace(toolbarItem.Text)
+                            ? toolbarItem.Text
+                            : toolbarItem.ToolTipText;
+                    }
+
+                    var menuToolbarItem = new ToolStripMenuItem(text)
```
mstv, line 139: Nit:
```suggestion
                    ToolStripMenuItem menuToolbarItem = new(text)
```

mstv (commented):
Suprisingly litte code.

But for a non-obvious reason, the last three elements of the filter toolbar have inverted visibility:
grafik

I guess there is a conflict with existing settings. A common prefix or suffix could help.

RussKie:
👀 at screenshots - this is soo cool! 🚀 

I'd replace "[x] Separator" with an actual separator as the separator should be shown/hidden based on whether any item in its group is visible.

pmiossec:
> But for a non-obvious reason, the last three elements of the filter toolbar have inverted visibility:
> I guess there is a conflict with existing settings. A common prefix or suffix could help.

Done. Could you give another try?


> I'd replace "[x] Separator" with an actual separator as the separator should be shown/hidden based on whether any item in its group is visible.

Done also.

RussKie:
Nice!
image image

RussKie:
A little issue - the highlighted items each represent a single unit:
image


E.g. this makes no sense:
image

pmiossec:
> A little issue - the highlighted items each represent a single unit:
> image
> 
> E.g. this makes no sense:
> image

I was aware of this fact but I was betting on the cleverness of the user to handle this case to not clutter a lot the source code for something that should not be used very often. I also was thinking that the user could also want to remove the labels. 

And as I don't think of a clever way to handle it...

mstv:
The issue is reproducible in the first run of a portable build. After a restart, it works.

RussKie:
> I was aware of this fact but I was betting on the cleverness of the user to handle this case to not clutter a lot the source code for something that should not be used very often. I also was thinking that the user could also want to remove the labels.

I can buy into this :) 

This proposal is certainly a significant improvement 👍

pmiossec:
> I can buy into this :)

Too late, I did something 🤣 

image

Surely need some code cleaning...

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
             }
         }
 
+        private static void CreateToolStripSubMenus(ToolStrip senderToolStrip, ToolStripMenuItem toolStripItem)
+        {
+            const string toolbarPrefix = "formbrowse_toolbar_visibility_";
```
RussKie, line 137: This is greatly bloat the settings 🤔 
I wonder whether we need to store this information in the appsettings, or store it separately like we store windows positions (`WindowPositionList`)?

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+                        item.Visible = visible;
+                        AppSettings.SetBool(toolbarPrefix + item.Name, item.Visible);
+                    }
+                }
+            }
+
+            void AdaptSeparatorsVisibility()
+            {
+                // First pass: toolbar items from left to right
+                var shouldHideNextSeparator = true;
```
RussKie, line 153:
```suggestion
                bool shouldHideNextSeparator = true;
```

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+                    if (currentGroup == groupname)
+                    {
+                        toolbarItem.Visible = AppSettings.GetBool(toolbarPrefix + toolbarItem.Name, true);
+                        continue;
+                    }
+
+                    currentGroup = (string)toolbarItem.Tag;
+                }
+
+                var menuToolbarItem = new ToolStripMenuItem(text) { Checked = toolbarItem.Visible, CheckOnClick = true, Tag = toolbarItem, Image = toolbarItem.Image };
```
RussKie, line 216:
```suggestion
                ToolStripMenuItem menuToolbarItem = new(text) 
                { 
                    Checked = toolbarItem.Visible, 
                    CheckOnClick = true, 
                    Tag = toolbarItem, 
                    Image = toolbarItem.Image 
                };
```

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+
+                    AdaptSeparatorsVisibility();
+                };
+
+                toolStripItem.DropDownItems.Add(menuToolbarItem);
+            }
+
+            AdaptSeparatorsVisibility();
+
+            bool BelongToAGroup(ToolStripItem toolbarItem, out string groupname)
```
RussKie, line 237:
```suggestion
            static bool BelongToAGroup(ToolStripItem toolbarItem, out string groupName)
```

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+            {
+                const string groupPrefix = "ToolBar_group:";
+                if (toolbarItem.Tag is string group && @group.StartsWith(groupPrefix))
```
RussKie, line 240: Why "@group" and not "group"?

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -115,6 +132,122 @@ static ToolStripItem CreateItem(ToolStrip senderToolStrip)
...
+                        SetVisibilityForSameGroupToolbarItem(toolbarItem.Tag, toolbarItem.Visible);
+                    }
+
+                    AdaptSeparatorsVisibility();
+                };
+
+                toolStripItem.DropDownItems.Add(menuToolbarItem);
+            }
+
+            AdaptSeparatorsVisibility();
```
RussKie, line 235: Keep all local methods at the end of the containing method, and after `return;` statement.

RussKie (commented):
Few comments

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
pmiossec: Ok. I already doing some more refactoring to please CodeFactor....

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
pmiossec: Using the appsettings is very convenient so I prefer to continue to do this way.
I changed a little the way to save the value so now only when an item is hidden, a setting with the value `false` is added to the settings. Otherwise nothing is store.
So a lot less bloat.

RussKie (approved):
:rocket:

Haven't run

mstv:
> The issue is reproducible in the first run of a portable build. After a restart, it works.

Has not changed 🤔
Can you reproduce it with the portable build?

RussKie:
I downloaded the build from https://ci.appveyor.com/project/gitextensions/gitextensions/builds/38170891/artifacts and all seems good. Except for not persisting the toolbar visibility (which is a known issue, of course), but now it looks especially odd. 

May be we should remove checkboxes from the toolbar selection?
image

mstv:
Delete / rename the file `c:\Users\<user>\AppData\Local\GitExtensions\WindowPositionsx.xml` in order to get the weird behavior.

PR #8923, opened 2021-03-05
Fix hang when requesting submodule status

drewnoakes:
- Fixes #8914 so that the application doesn't hang when submodule results have not completed at the time they're requested.
- Makes the UI responsive while the submodule operation is underway.
- Discards some unawaited tasks, as these triggered compiler diagnostics after making the above changes (unsure why).

This required making methods `GitModule.GetSingleDiff` and `SubmoduleHelpers.GetCurrentSubmoduleChanges` async.

## Test methodology

- Manual testing

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8952, opened 2021-03-07
range-diff: Incorrect limit when selecting four commits

gerhardol:
Candidate for 3.5

## Proposed changes

* range-diff: Incorrect limit when selecting four commits

Git diff-range can be used to show the difference in 'slices' of branches.
This is advanced usage, my normal usage is just to select two heads.
But if you select baseA, A, baseB, B then range-diff shows the difference between the commits,
(see example after, just with one commit in each 'slice').
This is helpful in comparing a patchset in two separate branches (master and release for instance).

range-diff can be slow with many commits and is limited to 100 commit differences.
This calculation was incorrect and did not count the differences but the union for the "base" and the "head" why the command was normally refused.
With this, also a routine in GitModule could be removed.

* FileStatusList: avoid getRevision() for index commits

Just avoid some CPU cycles

* rev-list, merge-base: Avoid Git call when arguments are identical

No need to invoke Git in these situations, even if the commands are cached.

### Before

image

### After

image

## Test methodology 

manual

---

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -788,23 +788,6 @@ namespace GitCommands
             return null;
         }
 
-        public int? GetCommitDiffCount(ObjectId baseId, ObjectId parentId)
-        {
-            var args = new GitArgumentBuilder("rev-list")
-            {
-                $"{baseId} {parentId}",
-                "--count"
-            };
-            var output = _gitExecutable.GetOutput(args);
-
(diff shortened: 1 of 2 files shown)


gerhardol:
With #8923, the check in diffCalculator can be removed as the request no longer blocks the UI.
I still would like to merge this PR, also for 3.5.

RussKie (approved):

mstv (approved):

gerhardol:
@msftbot merge

PR #8953, opened 2021-03-07
RevDiff: Show worktree/index changes similar to FormCommit

### LINKED ISSUE #8054: Commit count do not match files in RevDiff WorkTree

gerhardol:
Fixes #8054

## Proposed changes

This PR has two purposes:
* Get the output in RevDiff panel to be the same as in the tool tip and in FormCommit (see #8054 for details, not much discussed here)
* Avoid unnecessary git-diff command for artificial commits

git-status is required to see new files, but git-diff was used to get other files. When selecting worktree or index only, the git-diff command is not needed, all info is in git-status. 
(Longer explanation: same as HEAD->Index and Index->Worktree, any other combination involving artificial diffs will require git-diff too.)

The git-diff can be considerably slower if viruses like Symantec are installed, delays showing the actual diff.

Note that git-status normally is slower than git-diff.

### Before

image

image

### After

image

image

## Test methodology 

Change just line endings for a file, for instance like:
`perl -pi -e 's/\r\n/\n/;' GitExtensions.settings`
The examples also used a dummy change to a file

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2428,6 +2428,12 @@ namespace GitCommands
 
         public IReadOnlyList<GitItemStatus> GetDiffFilesWithUntracked(string? firstRevision, string? secondRevision, StagedStatus stagedStatus, bool noCache = false)
         {
+            if (stagedStatus is StagedStatus.WorkTree or StagedStatus.Index)
+            {
+                var status = GetAllChangedFilesWithSubmodulesStatus();
+                return status.Where(x => x.Staged == stagedStatus).ToList();
+            }
+
             var output = GetDiffFiles(firstRevision, secondRevision, noCache: noCache, nullSeparated: true);
             var result = GetDiffChangedFilesFromString(output, stagedStatus);
 


RussKie (approved):

mstv (approved):

gerhardol:
I will push this for 3.5 too

Issue #8955, opened 2021-03-08
Documentation links are broken


RussKie:
## Current behaviour

Documentation pages are inaccessible from the app.

E.g. the Settings dialog:
![image](https://user-images.githubusercontent.com/4403806/110273861-9f4f9100-8021-11eb-8a2e-c66367203b52.png)

Click on the blue ℹ in this dialog will try to open the following URLs:
* https://git-extensions-documentation.readthedocs.io/settings.html#git-extensions-appearance-author-images-avatar-provider
* https://git-extensions-documentation.readthedocs.io/settings.html#git-extensions-appearance-author-images-avatar-fallback
And a user is greeted with the following:
![image](https://user-images.githubusercontent.com/4403806/110273943-cd34d580-8021-11eb-89d6-ea8e5f6b6040.png)

Also open `FormClone` dialog and click on [?] button:
![image](https://user-images.githubusercontent.com/4403806/110274001-f190b200-8021-11eb-8c8d-3661a64d3a70.png)
It opens https://git-extensions-documentation.readthedocs.io/getting_started.html#clone-repository, which is also a 404.

## Expected behaviour

All links are correctly resolved and opened for the _latest_ version.

## Did this work in previous version of GitExtensions

Very likely a regression introduced in https://github.com/gitextensions/gitextensions/pull/8838 and https://github.com/gitextensions/gitextensions/pull/8893.

## Environment

- GitExtensions version: 12615869044a4db1f95609c05af93665045eb181

RussKie:
@gerhardol with https://github.com/gitextensions/GitExtensionsDoc/issues/133 fixed, how do you think we can fix this one?
I can't see another solution apart from restoring restoring `/en/latest`. (In fact it should be `/<current user language>/latest`, but that's a separate issue).

gerhardol:
I played around yesterday and have more or less the same conclusion. (I have been sick for some days and it took quite some hours of reading and fixing minor things until I even read this.)

Plan to make PR sometime tonight (either /en/master or an redirect which would handle translations, but release should probably be updated as it was before, just not / that is automatically redirected).

RussKie:
Please look after yourself first, and get well soon.
I've been burning a lot of midnight oil too, hence the delays with the
release and reviews...

gerhardol:
I added a redirect again and now it works.... As strange as the build on the 3.5 commit that displayed as 3.3.

[...]

PR #8956, opened 2021-03-08
Dialogs layout alignment - `FormDeleteBranch`

RussKie:
Relates to #6183

## Proposed changes

- Apply the same design as for other updated dialogs
- Remove verbose "help" text, replaced with [?] button that links to [the docs](https://git-extensions-documentation.readthedocs.io/en/release-3.4/branches.html#delete-branch) (the link is currently broken, see: https://github.com/gitextensions/gitextensions/issues/8955)
- Remove nagging confirmations - this is NOT a data loss situation, local branches can be easily restored
- Set focus to the branches edit control

### Before

image
image
image

### After

image
image

## Test methodology 

- manual

----

### DIFF
--- a/GitCommands/Git/Commands/GitDeleteBranchCmd.cs
+++ b/GitCommands/Git/Commands/GitDeleteBranchCmd.cs
@@ -14,6 +14,11 @@ namespace GitCommands.Git.Commands
         public GitDeleteBranchCmd(IReadOnlyCollection<IGitRef> branches, bool force)
         {
             _branches = branches ?? throw new ArgumentNullException(nameof(branches));
+            if (_branches.Count == 0)
+            {
+                throw new ArgumentException("At least one branch is required.", nameof(branches));
+            }
+
             _force = force;
         }
 
@@ -22,14 +27,15 @@ namespace GitCommands.Git.Commands
 
         protected override ArgumentString BuildArguments()
         {
-            var hasRemoteBranch = _branches.Any(branch => branch.IsRemote);
-            var hasNonRemoteBranch = _branches.Any(branch => !branch.IsRemote);
+            bool hasRemoteBranch = _branches.Any(branch => branch.IsRemote);
+            bool hasNonRemoteBranch = _branches.Any(branch => !branch.IsRemote);
 
             return new GitArgumentBuilder("branch")
             {
-                { _force, "-D", "-d" },
-                { hasRemoteBranch && hasNonRemoteBranch, "-a" },
-                { hasRemoteBranch && !hasNonRemoteBranch, "-r" },
+                { "--delete" },
+                { _force, "--force" },
+                { hasRemoteBranch && hasNonRemoteBranch, "--all" },
+                { hasRemoteBranch && !hasNonRemoteBranch, "--remotes" },
                 _branches.Select(branch => branch.Name.Quote())
             };
         }
--- a/GitUI/CommandsDialogs/FormDeleteBranch.cs
+++ b/GitUI/CommandsDialogs/FormDeleteBranch.cs
@@ -1,24 +1,24 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
+using System.Drawing;
 using System.Linq;
 using System.Windows.Forms;
+using GitCommands;
 using GitCommands.Git;
 using GitCommands.Git.Commands;
+using GitExtUtils.GitUI;
 using GitUIPluginInterfaces;
 using ResourceManager;
 
 namespace GitUI.CommandsDialogs
 {
-    public sealed partial class FormDeleteBranch : GitModuleForm
+    public sealed partial class FormDeleteBranch : GitExtensionsDialog
     {
-        private readonly TranslationString _deleteBranchCaption = new("Delete branches");
-        private readonly TranslationString _deleteBranchQuestion = new(
-            "Are you sure you want to delete selected branches?" + Environment.NewLine + "Deleting a branch can cause commits to be deleted too!");
-        private readonly TranslationString _deleteUnmergedBranchForcingSuggestion =
-            new("You cannot delete unmerged branch until you set “force delete” mode.");
-        private readonly TranslationString _cannotDeleteCurrentBranchMessage =
-            new("Cannot delete the branch “{0}” which you are currently on.");
+        private readonly TranslationString _deleteBranchCaption = new("Delete Branches");
+        private readonly TranslationString _cannotDeleteCurrentBranchMessage = new("Cannot delete the branch “{0}” which you are currently on.");
+        private readonly TranslationString _deleteBranchConfirmTitle = new("Delete Confirmation");
+        private readonly TranslationString _deleteBranchQuestion = new("The selected branch(es) have not been merged into HEAD.\r\nProceed?");
+        private readonly TranslationString _useReflogHint = new("Did you know you can use reflog to restore deleted branches?");
 
         private readonly IEnumerable<string> _defaultBranches;
         private readonly HashSet<string> _mergedBranches = new();
@@ -33,24 +33,31 @@ namespace GitUI.CommandsDialogs
         }
 
         public FormDeleteBranch(GitUICommands commands, IEnumerable<string> defaultBranches)
-            : base(commands)
+            : base(commands, enablePositionRestore: false)
         {
             _defaultBranches = defaultBranches;
 
             InitializeComponent();
+
+            // work-around the designer bug that can't add controls to FlowLayoutPanel
+            ControlsPanel.Controls.Add(Delete);
+            AcceptButton = Delete;
+
             InitializeComplete();
         }
 
-        private void FormDeleteBranchLoad(object sender, EventArgs e)
+        protected override void OnRuntimeLoad(EventArgs e)
         {
+            base.OnRuntimeLoad(e);
+
             Branches.BranchesToSelect = Module.GetRefs(RefsFilter.Heads).ToList();
-            foreach (var branch in Module.GetMergedBranches())
+            foreach (string branch in Module.GetMergedBranches())
             {
                 if (!branch.StartsWith("* "))
                 {
                     _mergedBranches.Add(branch.Trim());
                 }
-                else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch.Substring(2)))
+                else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
                 {
                     _currentBranch = branch.Trim('*', ' ');
                 }
@@ -62,50 +69,69 @@ namespace GitUI.CommandsDialogs
             }
         }
 
-        private void OkClick(object sender, EventArgs e)
+        protected override void OnShown(EventArgs e)
         {
-            try
+            RecalculateSizeConstraints();
+            base.OnShown(e);
+            Branches.Focus();
+        }
+
+        private void RecalculateSizeConstraints()
+        {
+            SuspendLayout();
+            MinimumSize = MaximumSize = Size.Empty;
+
+            int height = ControlsPanel.Height + MainPanel.Padding.Top + MainPanel.Padding.Bottom
+                       + tlpnlMain.Height + tlpnlMain.Margin.Top + tlpnlMain.Margin.Bottom + DpiUtil.Scale(42);
+
+            MinimumSize = new Size(tlpnlMain.PreferredSize.Width + DpiUtil.Scale(70), height);
+            MaximumSize = new Size(Screen.PrimaryScreen.Bounds.Width, height);
+            Size = new Size(Width, height);
+            ResumeLayout();
+        }
+
+        private void Delete_Click(object sender, EventArgs e)
+        {
+            var selectedBranches = Branches.GetSelectedBranches().ToArray();
+            if (!selectedBranches.Any())
             {
-                var selectedBranches = Branches.GetSelectedBranches().ToArray();
-
-                if (!selectedBranches.Any())
-                {
-                    return;
-                }
-
-                if (_currentBranch is not null && selectedBranches.Any(branch => branch.Name == _currentBranch))
-                {
-                    MessageBox.Show(this, string.Format(_cannotDeleteCurrentBranchMessage.Text, _currentBranch), _deleteBranchCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
-                    return;
-                }
-
-                // always treat branches as unmerged if there is no current branch (HEAD is detached)
-                var hasUnmergedBranches = _currentBranch is null || selectedBranches.Any(branch => !_mergedBranches.Contains(branch.Name));
-
-                // we could show yes/no dialog and set forcing checkbox automatically, but more safe way is asking user to do it himself
-                if (hasUnmergedBranches && !ForceDelete.Checked)
-                {
-                    MessageBox.Show(this, _deleteUnmergedBranchForcingSuggestion.Text, _deleteBranchCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
-                    return;
-                }
-
-                // ask for confirmation to delete unmerged branch that may cause loosing commits
-                // (actually we could check if there are another branches pointing to that commit)
-                if (hasUnmergedBranches
-                    && MessageBox.Show(this, _deleteBranchQuestion.Text, _deleteBranchCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) != DialogResult.Yes)
-                {
-                    return;
-                }
-
-                GitDeleteBranchCmd cmd = new(selectedBranches, ForceDelete.Checked);
-                UICommands.StartCommandLineProcessDialog(this, cmd);
-            }
-            catch (Exception ex)
-            {
-                Trace.WriteLine(ex.Message);
+                return;
             }
 
-            Close();
+            if (_currentBranch is not null && selectedBranches.Any(branch => branch.Name == _currentBranch))
+            {
+                MessageBox.Show(this, string.Format(_cannotDeleteCurrentBranchMessage.Text, _currentBranch), _deleteBranchCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return;
+            }
+
+            // always treat branches as unmerged if there is no current branch (HEAD is detached)
+            bool hasUnmergedBranches = _currentBranch is null || selectedBranches.Any(branch => !_mergedBranches.Contains(branch.Name));
+            if (hasUnmergedBranches && !AppSettings.DontConfirmDeleteUnmergedBranch)
+            {
+                TaskDialogPage page = new()
+                {
+                    Text = _deleteBranchQuestion.Text,
+                    Caption = _deleteBranchConfirmTitle.Text,
+                    Icon = TaskDialogIcon.Warning,
+                    Buttons = { TaskDialogButton.Yes, TaskDialogButton.No },
+                    DefaultButton = TaskDialogButton.No,
+                    Footnote = _useReflogHint.Text,
+                    SizeToContent = true,
+                };
+
+                bool isConfirmed = TaskDialog.ShowDialog(Handle, page) == TaskDialogButton.Yes;
+                if (!isConfirmed)
+                {
+                    return;
+                }
+            }
+
+            GitDeleteBranchCmd cmd = new(selectedBranches, force: hasUnmergedBranches);
+            bool success = UICommands.StartCommandLineProcessDialog(Owner, cmd);
+            if (success)
+            {
+                Close();
+            }
         }
     }
 }
--- a/UnitTests/GitCommands.Tests/Git/Commands/GitDeleteBranchCmdTests.cs
+++ b/UnitTests/GitCommands.Tests/Git/Commands/GitDeleteBranchCmdTests.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Collections.Generic;
+using GitCommands;
+using GitCommands.Git.Commands;
+using GitUIPluginInterfaces;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace GitCommandsTests.Git.Commands
+{
+    [TestFixture]
+    public sealed class GitDeleteBranchCmdTests
+    {
+        [Test]
+        public void ctor_should_throw_if_branches_is_null()
+        {
+            Assert.Throws<ArgumentNullException>(() => new GitDeleteBranchCmd(branches: null, force: false));
+        }
+
+        [Test]
+        public void ctor_should_throw_if_branches_is_empty()
+        {
+            Assert.Throws<ArgumentException>(() => new GitDeleteBranchCmd(Array.Empty<IGitRef>(), force: false));
+        }
+
+        [Test]
+        public void ctor_should_have_expected_values()
+        {
+            const string remoteName = "origin";
+            string completeName = $"refs/remotes/{remoteName}/branch_name";
+            GitRef remoteBranchRef = SetupRawRemoteRef(remoteName, completeName);
+
+            GitDeleteBranchCmd cmd = new(new IGitRef[] { remoteBranchRef }, force: false);
+
+            Assert.IsFalse(cmd.AccessesRemote);
+            Assert.IsTrue(cmd.ChangesRepoState);
+        }
+
+        private static IEnumerable<TestCaseData> GetRefsCommandTestData
+        {
+            get
+            {
+                const string name = "local_branch";
+
+                // Test local branches only
+                List<IGitRef> localRefs = new();
+                var localGitModule = Substitute.For<IGitModule>();
+                var localConfigFileSettings = Substitute.For<IConfigFileSettings>();
+                localConfigFileSettings.GetValue($"branch.local_branch.merge").Returns(string.Empty);
+                localConfigFileSettings.GetValue($"branch.local_branch.remote").Returns(string.Empty);
+                localGitModule.LocalConfigFile.Returns(localConfigFileSettings);
+                GitRef localBranchRef = new(localGitModule, ObjectId.Random(), $"refs/heads/{name}");
+
+                localRefs.Add(localBranchRef);
+
+                yield return new TestCaseData(localRefs, /* force */ false, /* expected */ $"branch --delete \"{name}\"");
+                yield return new TestCaseData(localRefs, /* force */ true, /* expected */ $"branch --delete --force \"{name}\"");
+
+                // Test local and remote branches
+                const string remoteName = "origin";
+                string completeName = $"refs/remotes/{remoteName}/{name}";
+                GitRef remoteBranchRef = SetupRawRemoteRef(remoteName, completeName);
+
+                List<IGitRef> mixedRefs = new();
+                mixedRefs.Add(localBranchRef);
+                mixedRefs.Add(remoteBranchRef);
+
+                yield return new TestCaseData(mixedRefs, /* force */ false, /* expected */ $"branch --delete --all \"{name}\" \"origin/{name}\"");
+                yield return new TestCaseData(mixedRefs, /* force */ true, /* expected */ $"branch --delete --force --all \"{name}\" \"origin/{name}\"");
+
+                // Test remote branches only
+                List<IGitRef> remoteRefs = new();
+                remoteRefs.Add(remoteBranchRef);
+
+                yield return new TestCaseData(remoteRefs, /* force */ false, /* expected */ $"branch --delete --remotes \"origin/{name}\"");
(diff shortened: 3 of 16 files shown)


mstv (commented):
👍

The labels for input fields should end with a colon in order to indicate that is not just information but input.
A mnemonic would be nice, too.
I.e. "Select &branches:".

The text baseline of the label should be aligned with the input text's.
The label should have the same distance from the left borders as the buttons at the right.

RussKie:
> The labels for input fields should end with a colon in order to indicate that is not just information but input.

In the vast majority of our dialogs we don't do this, please have a look at https://github.com/gitextensions/gitextensions/issues/6183

> A mnemonic would be nice, too.
> I.e. "Select &branches:".

Updated

> The text baseline of the label should be aligned with the input text's.
> The label should have the same distance from the left borders as the buttons at the right.

Overlaid over `FormCreateBranch` - the positions are the same
image

The combo is a part of `BranchComboBox` and I don't want to mess with it currently.

RussKie:
@mstv are you happy with the current implementation?

mstv:
> The combo is a part of BranchComboBox. Looks like I'm changing it anyway

I am waiting for this.

> Overlaid over `FormCreateBranch` - the positions are the same
> image

The ComboBoxes do *not* have the same vertical position - neither is correct.
The `Create branch at this revision` is perfectly aligned with the input field.

drewnoakes:
> The labels for input fields should end with a colon in order to indicate that is not just information but input.

The guidance for development of Visual Studio has actually moved away from the use of colons in labels. I'm not saying that's a gold standard or that we have to do the same, but I thought I'd point it out here.

RussKie:
image

RussKie:
Looks like I forgot to push... 😩

drewnoakes:
Why remove the checkbox for force? Is this automatic now?

gerhardol:
> 
> 
> Why remove the checkbox for force? Is this automatic now?

That should not be removed.
You need to force if the brach is not merged, should be an explicit decision.

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -61,50 +66,47 @@ private void FormDeleteBranchLoad(object sender, EventArgs e)
...
+
+            if (_currentBranch is not null && selectedBranches.Any(branch => branch.Name == _currentBranch))
             {
-                Trace.WriteLine(ex.Message);
+                MessageBox.Show(this, string.Format(_cannotDeleteCurrentBranchMessage.Text, _currentBranch), _deleteBranchCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return;
             }
 
-            Close();
+            GitDeleteBranchCmd cmd = new(selectedBranches, force: true);
```
gerhardol, line 104: Add the checkbox again

gerhardol (dismissed):

RussKie:
Deletion of a local branch **is not** a destructive or code-loss event. 

Deleted a branch mistake? No worries, show reflogs, and recreate it.

I'm sick of multiple confirmation boxes, which carry no meaningful function.
I switch between different branches, including contributors' remotes all day every day. I get hundreds of branches I don't need. Deleting a branch becomes a mammoth undertaking.

gerhardol:
> Deletion of a local branch **is not** a destructive or code-loss event.
> 
> Deleted a branch mistake? No worries, show reflogs, and recreate it.
> 
> I'm sick of multiple confirmation boxes, which carry no meaningful function.
> I switch between different branches, including contributors' remotes all day every day. I get hundreds of branches I don't need. Deleting a branch becomes a mammoth undertaking.

Setting force that deviate from Git defaults which is not something GE normally do.

If we are deviating from Git defaults, what more should we change?
Reset is similar to delete.
If this is something that should be done, I would like to remove all variants of reset, those annoys me.

If GE is not using Git defaults this is normally handled by overrides.
In this case the multi select handling could be improved, just having one box to check.

RussKie:
> Setting force that deviate from Git defaults which is not something GE normally do.

Sorry Gerhard, you completely lost me here. Which git default are you referring to? I can create a branch, make commits, switch to another branch and delete the first branch from the command line. No warnings, no nothing... Whilst GE makes me jump through hoops to delete a branch.
The sole purpose of the app is to make a user's life easier and automate git interactions. And the branch deletion dialog is doing the completely opposite of that.

> Reset is similar to delete.
> If this is something that should be done, I would like to remove all variants of reset, those annoys me.

Reset operation **can be** lossy if a user has uncommitted changes. A soft/mixed reset can also be very intrusive if a user has unstaged/staged changes. 
But the user can't lose uncommitted changes deleting a branch.

gerhardol:
> > Setting force that deviate from Git defaults which is not something GE normally do.
> 
> Sorry Gerhard, you completely lost me here. Which git default are you referring to?

--force is not set by default, so I would not expect the option to be set by default

https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---force

pmiossec:
It was also in my todo-list to change the behavior of this form because I find painful the message "You can't delete umerged branch...." (2nd screenshot) that appears when you validate because most of the times you didn't even know if the branch is merged or not.

But I was thinking more by detecting unmerged state of the branch(es) before the user click validate and enabling/disabling the button  if you can't delete the unmerged branch(es). Then you will have to check the checkbox (renamed "Delete unmerged"?) to be able to validate.

Note: I think also that we could do better than git by checking that the branch is merged in another ref instead just HEAD (like Git does) so that the user has to check the box **really only** when commits will be only reachable by the reflog...

> Remove verbose "help" text, replaced with [?] button that links to the docs 

I don't like this kind of things where you have to open a documentation to understand something (because nearly everyone hate reading documentation). We end up with a worst UX, for me.
I can't count anymore the number of times where I said to beginner asking "why we have to check 'force'?" that the explanation  is already clearly written and they just have to read it.

>Deleted a branch mistake? No worries, show reflogs, and recreate it.

If you really want to go with this redesign, I think you should display clearly somewhere that they could use the reflog. Because a lot of user doesn't even know this feature or don't use enough often to think about it.


>Reset operation can be lossy if a user has uncommitted changes. A soft/mixed reset can also be very intrusive if a user has unstaged/staged changes.

Talking about the "Reset" command, I would like to set the default at "Soft" (instead of "Mixed" currently) because that's the one where you loose the less the current state... I will do a PR to start the discussion.

RussKie:
Thank you, this is very helpful.
In all the years I've been using git I always used `-D` option.

Can you give me a scenario in which we would want to stop a user, wishing
to delete a branch, from deleting the branch?

A fair point about documentation and reflog in general. I do agree to a
degree, but having random poorly articulated blurbs of text on random
dialogs in random spots is a UX disaster.

IMO the biggest impact in this area we can by having a good documentation,
which is uniformly accessible from each of our dialog.

RussKie:
WRT: making reflog more visible - I've been meaning to add "Show all branches", "Show current branch" and "Show reflog" to the Filters toolbar (like we have "Show first parents").

OT: Also "Show reflogs" should reset to "Show all branches", if only current or filtered branches are shown.

pmiossec:
> WRT: making reflog more visible - I've been meaning to add "Show all branches", "Show current branch" and "Show reflog" to the Filters toolbar (like we have "Show first parents").

It's a good idea to give some visibility to these features.

But for me it is not more a problem of visibility than one than understanding what to do with it and use cases when to use it.

You removed an informing message and a check that could prevent a user to do a mistake by make him think about why he must check it without giving him a clue on how to fix the error.

Your design rework is pushing a little more GE toward a power user usage.
And seeing how dev struggle to get how to use git right or more advanced features that could make your job easier, I don't know if it's the good direction.

I would add a message like: 
"Tip: you could later restore a mistakenly deleted local branch by using the 'reflog' feature". and eventually "See documentation for more details".

By this simple tip we could learn something to beginners and open them a new horizon :D

mstv:
If you don't want to see this tip text all the day, you could hide it if the setting "Don't show help images" is ticked.

RussKie:
I'm aware of those. These "help images" and "help text" exist on very few
forms, and, let's be honest, they look awkward.

They also present a UI challenge because it is difficult to provide a good
concise meaningful help text, and present this text in every different
language we support. Not to mention translation challenges.

Instead I want a unified experience across the app where each dialog has a
link to docs, where we are not constrained in length, can have images and
external links, if necessary.

drewnoakes:
My personal take on the discussion here.

In general, I'm in favour of simplifying dialogs. I feel like a lot of our dialogs give the user all the control, at the cost of all the complexity. I _regularly_ feel as though GE should pre-populate or show/hide options based on its own understanding of the repo/branch/etc.

- Delete branch dialog
  - If the branch is merged to HEAD, don't show "force"
  - If the branch has remote peers, offer to delete them too
- Push branch dialog
  - If the branch is untracked, don't show "force"
  - If the branch has diverged such that force will be required, show me a warning and pre-select "force".
    - This would use "force with lease" internally. If that failed, it'd progress me to another option with an explanation such as "The remote branch has moved since you last fetched it. Force pushing may lose remote changes you haven't seen yet. Are you sure you want to force push?"
- Commit dialog
  - Disable "Push" button when a branch has a tracking branch which is in sync with the remote (or at least make state visible)
- Rebase dialog
  - If the branch has `fixup!` or `squash!` commits, pre-select "interactive"
- Merge dialog
  - If the branch can be fast-forwarded, show me that fact. Show an option to create a merge commit.
  - If the branch cannot be fast-forwarded, explain that a merge commit will be made. Give the option to rebase instead, which swaps to another dialog.

These scenarios are based around thinking through what the user is trying to achieve, rather than how to map the set of options that git supports to dialogs. GE historically has not take much opinion about these things, but the consequence of this is that we have so many dials and knobs everywhere. I would like to streamline workflows and make the tooling for welcoming by adding contextual assistance to hide complexity without removing any of the power. This isn't about forcing particular workflows and being "opinionated", it's about hiding options that make no sense and emphasing those which make the most sense in given situations.

gerhardol:
While I agree in general, it is not always so simple.
For remotes, the local may not be synced (there are likely local status that cannot be determined too).

> * Push branch dialog
>   
>   * If the branch is untracked, don't show "force"
>   * If the branch has diverged such that force will be required, show me a warning and pre-select "force".

You may not always track remote branches and also if you do the status on the server may differ from what is seen at the local.
Also, it is possible to change the branch to push to.

So many check boxes should stay.

Git defaults may not be the best in all scenarios, but it is easier to explain to users.

mstv:
This PR is about deleting local branches.
Since I rarely checkout the master branch, the check feels arbitrary and unpredictable very often.
Only checks for containment in `upstream/master` .. `origin/main` would make a little sense for me. Would need configuration...

I vote for removal of the checkbox - perhaps with a hint text pointing the user to the reflog. The necessary confirmation of the  popup should provide enough safety.
The minimum we "must" do, is to add a mnemonic to the `Force` checkbox.

RussKie:
> The necessary confirmation of the popup should provide enough safety.

This is the purpose of the [?] button - it points to our docs. I'm happy to extend this section to mention the reflogs.
image

RussKie:
Added the ability to suppress nagging confirmations.
Once https://github.com/gitextensions/gitextensions/pull/9372 is merged, I'll update the confirmation settings page to include the new setting.

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -1,24 +1,25 @@
...
+    public sealed partial class FormDeleteBranch : GitExtensionsDialog
     {
         private readonly TranslationString _deleteBranchCaption = new("Delete branches");
-        private readonly TranslationString _deleteBranchQuestion = new(
-            "Are you sure you want to delete selected branches?" + Environment.NewLine + "Deleting a branch can cause commits to be deleted too!");
-        private readonly TranslationString _deleteUnmergedBranchForcingSuggestion =
-            new("You cannot delete unmerged branch until you set “force delete” mode.");
         private readonly TranslationString _cannotDeleteCurrentBranchMessage =
             new("Cannot delete the branch “{0}” which you are currently on.");
+        private readonly TranslationString _deleteBranchConfirmTitle = new("Delete Confirmation");
```
mstv, line 20: ❔
```suggestion
        private readonly TranslationString _deleteBranchConfirmTitle = new("Delete confirmation");
```

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -1,24 +1,25 @@
...
+        private readonly TranslationString _deleteBranchQuestion = new("Are you sure you want to delete selected branches?");
```
mstv, line 21: ❔
```suggestion
        private readonly TranslationString _deleteBranchQuestion = new("Are you sure you want to delete the selected branch(es), which have not been merged into HEAD?");
```

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -62,50 +70,67 @@ private void FormDeleteBranchLoad(object sender, EventArgs e)
...
+                    Icon = TaskDialogIcon.Information,
+                    Buttons = { TaskDialogButton.Yes, TaskDialogButton.No },
+                    Footnote = _useReflogHint.Text,
+                    Verification = new(TranslatedStrings.DontShowAgain),
+                    SizeToContent = true
+                };
 
-                // ask for confirmation to delete unmerged branch that may cause loosing commits
-                // (actually we could check if there are another branches pointing to that commit)
-                if (hasUnmergedBranches
```
mstv, line 94: I would not show the confirmation dialog if all branches have been merged into HEAD.

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -62,50 +70,67 @@ private void FormDeleteBranchLoad(object sender, EventArgs e)
...
-                    return;
+                    AppSettings.DontConfirmDeleteBranch = page.Verification.Checked;
                 }
-
-                GitDeleteBranchCmd cmd = new(selectedBranches, ForceDelete.Checked);
-                UICommands.StartCommandLineProcessDialog(this, cmd);
             }
-            catch (Exception ex)
+
+            GitDeleteBranchCmd cmd = new(selectedBranches, force: true);
```
mstv, line 128:
```suggestion
            GitDeleteBranchCmd cmd = new(selectedBranches, force: hasUnmergedBranches);
```

mstv (commented):
Align the text of the Label and of the ComboBox as well as the frame of the ComboBox and the Button, please.

GitUI/CommandsDialogs/FormDeleteBranch.cs
RussKie: This is consistent with other titles. If we want to change them it needs to be done separately across the codebase.

RussKie:
`BranchComboBox` is a difficult control to deal with, here's the best I can do. Only tested on Q11, as I don't currently have W10 to test it.
image

GitUI/CommandsDialogs/FormDeleteBranch.cs
RussKie: I want to be able to easily delete local branches irrespective of their merge status. Applying this condition will prevent this.
This is a 'no loss of data' scenario, as it can be restored via reflogs.

GitUI/CommandsDialogs/FormDeleteBranch.cs
RussKie: image

RussKie: image

GitUI/CommandsDialogs/FormDeleteBranch.cs
RussKie: Done

GitUI/CommandsDialogs/FormDeleteBranch.cs
mstv: How should we capitalize window titles?
At least for FormDeleteBranch and its TaskDialog it should be consistent (see screenshots), shouldn't it?

GitUI/CommandsDialogs/FormDeleteBranch.cs
mstv: > I want to be able to easily delete local branches irrespective of their merge status.

Yes, of course. So do I.

> Applying this condition will prevent this.

Why do you think this?
It will add "--force" only if necessary - as @gerhardol requested IIUIC.
(The omitting of "--force" works for checked-out branches but not yet for `detached HEAD` because our parser is buggy: #9279 - to be done after https://github.com/gitextensions/gitextensions/pull/9254#discussion_r649489487. This will not disturb you.)

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -62,50 +69,75 @@ private void FormDeleteBranchLoad(object sender, EventArgs e)
...
-                    return;
-                }
+            if (_currentBranch is not null && selectedBranches.Any(branch => branch.Name == _currentBranch))
+            {
+                MessageBox.Show(this, string.Format(_cannotDeleteCurrentBranchMessage.Text, _currentBranch), _deleteBranchCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return;
+            }
 
-                if (_currentBranch is not null && selectedBranches.Any(branch => branch.Name == _currentBranch))
+            // always treat branches as unmerged if there is no current branch (HEAD is detached)
```
mstv, line 107: The `null` check for `_currentBranch` is not necessary and causing #9279.

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -62,50 +69,75 @@ private void FormDeleteBranchLoad(object sender, EventArgs e)
...
                 }
 
-                // ask for confirmation to delete unmerged branch that may cause loosing commits
-                // (actually we could check if there are another branches pointing to that commit)
-                if (hasUnmergedBranches
-                    && MessageBox.Show(this, _deleteBranchQuestion.Text, _deleteBranchCaption.Text, MessageBoxButtons.YesNo, MessageBoxIcon.Question) != DialogResult.Yes)
+                if (!isConfirmed)
                 {
                     return;
                 }
```
mstv, line 132: Could be moved before `if (isConfirmed)`, which can be removed then.

mstv (commented):
> BranchComboBox is a difficult control to deal with, here's the best I can do.

Thank you. Looks much better.

pmiossec:
I prefer to tell it now again but I really don't link this behavior change.
As I explained it [previously](https://github.com/gitextensions/gitextensions/pull/8956#issuecomment-799439498), I also find that the current behavior is not good but I don't link the direction we take here 😟

Requiring the user to read the documentation without any nudge that could give him a hint that he should know all the consequences and understand how delete and reflog works before continuing
and relying only on the reflog to fix a mistake 
make me think that now this UI change will be really understandable only by power user and will be very dangerous for the beginners.
If a beginner check the "don't ask me again" checkbox, consequences could be very painful to fix.

And using the reflog is not an easy thing to do in this specific case if you discovered after some times that the branch was unmerged (because the reflog of this local branch has been deleted and so the reflog of HEAD will be difficult to read and use).

I'm still convinced that displaying a message warning message unmerged branches selected directly in the form would be better.

**And that once at least one branch unmerged has been selected, disabling the delete button and forcing the user to check a "force deleting unmerged branch" to enable again the button would be a better idea.**

I think it would give a much better feedback to the user while allowing to remove the annoying confirmation.

RussKie:
But why are you insisting on continuing to making it hard for non novice users?

And if a user was warned, chose to proceed and asked not to show the warning - well, I take no excuse here - the user **has** been warned.

pmiossec:
> But why are you insisting on continuing to making it hard for non novice users?

Because I consider that having to click on a checkbox when you see that the button is disabled (and a little warning message) is not making things harder. The power user will click on the checkbox and that will take him what 0.5s?

And in that case I personally highly value this feedback **BEFORE** having to click...

> And if a user was warned, chose to proceed and asked not to show the warning - well, I take no excuse here - the user **has** been warned.

I don't take your point. A good UX is one that prevent you to do a mistake (if the cost is indeed not to high to pay) and not just one that tell you "this feature is dangerous, use it at your own risk".

For me that's the problem of the checkbox. You have been warned one time. But that worth for all the other future times (even 1 year after). And here, I'm not even sure that everyone will understand the consequences of that warning.

And if you have to rely on the reflog to retrieve your branch, you will spend a lot more time that all the little time accumulated to just click on a checkbox... 

The solution I proposed was to safely remove the frustration of this confirmation.

The main thing is that if a user click the "don't ask me again" checkbox he will never know that a selected branch is not merged again. And even if he is a power user he could discover it long after and using a reflog will be painful.

It sometimes happened to me that a branch that I thought was merged is not and I give it a second look to see that there is a remaining commit for whatever reason.

I follow GE development so I understand that if this PR is merged, I have to not forget to **never** click this checkbox. But for other users?

What I feel is that PR improve a lot UX for the ones that have the same workflow than you but not the one of everyone.
I was expecting to improve everyone workflow (even if I understand that with what I propose, you will have one more click to do).

But don't get me wrong, this PR is an improvement over the current behavior. That's just that I think that giving this feedback earlier and preventing the user to do a mistake would bring more value...

I also understand that maybe I can't asked someone else to do what I think is good. I see all the good work and effort you do for the project.

That's just that after this PR merged, there will be no coming back...

From my point of view, I just regret not having as enough time as I would like to implement it.

That's why I'm interested by other point of view.

If that's just me, we could go with that, I would just ask to display the warning message "The selected branch(es) not been merged into HEAD" directly in `FormDeleteBranch` to have the information before clicking.

mstv:
Valid points, @pmiossec.
Disabling the button - best accompanied with a banner explaining why - gives much better UX than the popup afterwards.

I want an option to automatically force-delete if required by git - not necessarily in FormDeleteBranch.
1a. I do not want to checkout `master` in order to get a working / meaningful is-merged check.
1b. Since I often work with detached HEAD, this check is disabled in GE - I guess in order to not loose commits.
  So, this algorithm should be more intelligent. The current approach is stricter than git and feels arbitrary, especially with misleading messages if no branch is checked out.
2. If I think I want to delete a branch, I will do it also if must check `force`.
3. right click, `d`, `enter` is much more fluent than with `Alt+F` in-between or even aiming the checkbox with the mouse.
This was only the explanation why I do not want extra interaction or popups than FormDeleteBranch itself - enabled by an option.

The warning banner should be displayed only if there are selected "unmerged" branches.
The option should automatically set the `force` checkbox (which enables the `Delete` button).

gerhardol:
I agree with @pmiossec here as in my review. 
A serial interface with many separate popups is easier to implement than one where all info is available in the first presentation but it is not unreasonable here.

What if the Force delete checkbox is sticky, then it is visual queue before deleting, no additional setting?

RussKie:
Thank you for your input, appreciate it.

RussKie:
One thing everyone appears to completely overlook is that the check whether branch is merged or not is flawed, and it doesn't detect squash-merged branches. In repos whether squash-merge is a common practice (such as dotent/winforms) cleaning up a repo becomes a laborious exercise, where even clicking additional checkbox is an overhead.
Checking a squash-merged branch is a non trivial exercise, https://stackoverflow.com/a/60501712.

RussKie:
I don't feel the "force" button is require at all. 
* If the branch is merged - just delete it. 
* If the branch isn't merged (as per our naive check) - confirm with the user to proceed. If "yes" - just delete it. Else, do nothing, remain at the dialog.

In both cases, run `git branch -D`.

I'll remove the "Don't show again" prompt, and will make it available on the settings page only. Default - "confirm delete: true".

gerhardol:
> One thing everyone appears to completely overlook is that the check whether branch is merged or not is flawed, and it doesn't detect squash-merged branches. In repos whether squash-merge is a common practice (such as dotent/winforms) cleaning up a repo becomes a laborious exercise, where even clicking additional checkbox is an overhead.

That is why --ff-only should be used, it is then trivial to find and delete obsolete remote branches too.
(A good enough reason why GH was not considered at work.)

> These branches have been merged, yet we're reporting those as unmerged.

In this case you could have deleted local branches where remotes no longer exists (the _Delete obsolete branches_ should be updated to support that).

> In both cases, run git branch -D.

The full options should be used, which is `--delete --force`

> I don't feel the "force" button is require at all.
> I'll remove the "Don't show again" prompt, and will make it available on the settings page only. Default - "confirm delete: true".

If the checkbox is kept and only keep this in settings, I agree.
But no blocker for me.

A colleague of mine asked for branch multiselection to make such cleanups.

RussKie:
> A colleague of mine asked for branch multiselection to make such cleanups.

The dialog allows you to manually select multiple branches, though I find
it cumbersome.
It would be great to select multiple branches in the revision graph and the
left panel.

RussKie:
Done

RussKie:
@msftbot merge in 48 hours

GitCommands/Git/Commands/GitDeleteBranchCmd.cs
```diff
@@ -22,14 +22,14 @@ public GitDeleteBranchCmd(IReadOnlyCollection<IGitRef> branches, bool force)
...
-            var hasNonRemoteBranch = _branches.Any(branch => !branch.IsRemote);
+            bool hasRemoteBranch = _branches.Any(branch => branch.IsRemote);
+            bool hasNonRemoteBranch = _branches.Any(branch => !branch.IsRemote);
 
             return new GitArgumentBuilder("branch")
             {
-                { _force, "-D", "-d" },
-                { hasRemoteBranch && hasNonRemoteBranch, "-a" },
-                { hasRemoteBranch && !hasNonRemoteBranch, "-r" },
+                { _force, "--delete --force", "--delete" },
```
gerhardol, line 30: Should be split in two

gerhardol (commented):
If hints like this is added, it should be above buttons

image

...and as stated before, I prefer an explicit force checkbox, not blocking though...

RussKie:
> If hints like this is added, it should be above buttons
> 
> image

Why? IMO this is a secondary information which has not direct relation to the performed operation; the docs is the place for this information. I added only as a response to comments that most users aren't likely aware of reflog operation. 

> ...and as stated before, I prefer an explicit force checkbox, not blocking though...

The checkbox is superficial and presents an additional action that a user needs to take without adding a significant value. E.g. 
_A user tries to delete a branch, which isn't merged. The checkbox isn't checked, and the user is present with a dialog saying the branch can't be deleted unless the checkbox is checked. The user is then has to check the checkbox, to get another prompt to confirm the deletion._
The new behaviour is _to prompt the user to confirm deletion of an unmerge branch_.

Also in the case of all branches are merged the checkbox is totally unnecessary. And if we retain the checkbox we need to hide it, which requires extra code and extra layout related logic.
And to top it off, if the user disables confirmation prompts the checkbox shouldn't be shown as well.

gerhardol:
> > If hints like this is added, it should be above buttons
> 
> Why? IMO this is a secondary information which has not direct relation to the performed operation; the docs is the place for this information. I added only as a response to comments that most users aren't likely aware of reflog operation.

Most GUI dialogues have the buttons in the bottom.
This is info like the hints you get when starting applications that you can opt out from.
I do not delete branches so often that I would get bothered in this form, but will object if it is added to e.g. Push/Pull.

> > ...and as stated before, I prefer an explicit force checkbox, not blocking though...
> 
> The checkbox is superficial and presents an additional action that a user needs to take without adding a significant value. E.g.
> _A user tries to delete a branch, which isn't merged. The checkbox isn't checked, and the user is present with a dialog saying the branch can't be deleted unless the checkbox is checked. The user is then has to check the checkbox, to get another prompt to confirm the deletion._
> The new behaviour is _to prompt the user to confirm deletion of an unmerge branch_.
> 
> Also in the case of all branches are merged the checkbox is totally unnecessary. And if we retain the checkbox we need to hide it, which requires extra code and extra layout related logic.
> And to top it off, if the user disables confirmation prompts the checkbox shouldn't be shown as well.

My preference is to stay closer to Git, I see no benefit in hiding the behavior here.
The additional "Are you sure" popup is not in the discussion.

Prefill Force from settings and you are OK. 
In my case I would not set this and just select the checkbox when I am sure, avoiding the extra popup that I get now.

The checkbox could be grayed out if the branch is merged, similar to "--delete" is set from merged status (the implementation could always set "--delete").
I would delete such logic though and rely on reporting when deleting fails.

UnitTests/GitCommands.Tests/Git/Commands/GitDeleteBranchCmdTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        public void ctor_should_throw_if_branches_is_empty()
+        {
+            Assert.Throws<ArgumentException>(() => new GitDeleteBranchCmd(Array.Empty<IGitRef>(), force: false));
+        }
+
+        [Test]
+        public void ctor_should_have_expected_values()
+        {
+            const string remoteName = "origin";
+            string completeName = $"refs/remotes/{remoteName}/local_branch";
```
gerhardol, line 30:
```suggestion
            string completeName = $"refs/remotes/{remoteName}/branch_name";
```

PR #8958, opened 2021-03-08
Rename translation strings to avoid conflict

drewnoakes:
## Proposed changes

Rename translation strings to avoid conflict.

So long as we target .NET Framework, we do not have annotated versions of `string.IsNullOrEmpty` and friends. We've worked around this by introducing a `Strings` class that has annotated versions of these, but that class had a naming conflict with existing translation string resources.

## Test methodology 

- Manual testing.

----

### DIFF
--- a/GitCommands/ExternalLinks/ExternalLinksStorage.cs
+++ b/GitCommands/ExternalLinks/ExternalLinksStorage.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Xml;
 using System.Xml.Serialization;
 using GitCommands.Settings;
+using GitExtUtils;
 
 namespace GitCommands.ExternalLinks
 {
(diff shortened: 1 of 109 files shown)


RussKie:
image

Is this change modifying translations?

drewnoakes:
> Is this change modifying translations?

Not intentionally. I'll take a look.

drewnoakes:
@RussKie I just had to run a command to regenerate some files.

Rebased to fix merge conflict. There was no actual conflict when rebasing, so the diff is identical, just a different base now.

RussKie (approved):

drewnoakes:
Did you squash merge this? The commit description was lost.

RussKie:
Not sure... maybe...

PR #8959, opened 2021-03-08
Remove redundant await

ivangrek:
## Proposed changes

- Remove redundant await where is no needed

## Test methodology 

- Manually

## Test environment(s) 

- Git Extensions 33.33.33
- Build 22f3b681ffff7beec4a789bda7e9fa915cd62d45
- Git 2.30.1.windows.1
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -405,9 +405,9 @@ namespace GitUI.Editor
         /// <param name="item">The gitItem to present.</param>
         /// <param name="text">The patch text.</param>
         /// <param name="openWithDifftool">The action to open the difftool.</param>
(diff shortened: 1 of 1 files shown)


RussKie (approved):

RussKie:
@msftbot merge in 48 hours

drewnoakes (approved):
This kind of change:
- Avoids allocating some `Task` objects, but also
- Can make debugging more difficult as it means intermediate methods will not appear in stack traces.

ivangrek:
It also avoids creating a state machine

drewnoakes:
The overhead is minimal so I only recommend making this change in hot paths. It can be much harder to debug async code that is missing stack frames. This guidance has come out in other places too.

PR #8535, opened 2020-10-10
Fixes #8263 by catching file IO exception and showing message box

lhiginbotham:
Fixes #8263, a crash due to not being able to save a file to a locked filepath

## Proposed changes

- catch file IO exception and display the error to the user without crashing

## Screenshots 

Example message pop-up
image

## Test methodology 

- tested by attempting to save a file from the commit diff list to a file that was locked by opening an archive using 7-zip

## Test environment(s) 

- Windows 10

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8961, opened 2021-03-08
Report `SaveBlobAs` errors without crashing the app

### LINKED ISSUE #8263: [NBug] Der Prozess kann nicht auf die Datei "C:\Projects\Tropico6\...
## Current behaviour

GitExtension crashed when trying to save over a file via "Save As" where a file lock is hold by another process

## Expected behaviour

Some error popup telling that GitExtensions could not save the file.

## Steps to reproduce

- Grab any file with a history
- Open another program that holds an write-exclusive lock on the file
- Try to use "Save as" on the file, attempting to overwrite it.

## Error Details
```
System.IO.IOException: Der Prozess kann nicht auf die Datei "C:\Projects\Tropico6\Tropico6\Source\Tropico6\Game\Building\T6Building.cpp" zugreifen, da sie von einem anderen Prozess verwendet wird.
   bei System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   bei System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   bei System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
   bei GitCommands.GitModule.SaveBlobAs(String saveAs, String blob)
   bei GitUI.CommandsDialogs.FormFileHistory.saveAsToolStripMenuItem_Click(Object sender, EventArgs e)
   bei System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   bei System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   bei System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   bei System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   bei System.Windows.Forms.ToolStripItem.FireEventInteractive(EventArgs e, ToolStripItemEventType met)
   bei System.Windows.Forms.ToolStripItem.FireEvent(EventArgs e, ToolStripItemEventType met)
   bei System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   bei System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   bei System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   bei System.Windows.Forms.Control.WndProc(Message& m)
   bei System.Windows.Forms.ScrollableControl.WndProc(Message& m)
   bei System.Windows.Forms.ToolStrip.WndProc(Message& m)
   bei System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
   bei System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   bei System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

## Additional information
"Save as" from history on an older version over a file where another process holds an exclusive file lock.

I can understand, that GitExtension can't write over the file, but that isn't a reason to crash.. ;)

Note: I just got https://github.com/gitextensions/gitextensions/issues/8262 closed while I was editing the description text, so I hope that this here won't be closed as a duplicate of 8262.. lets see.

## Environment
- Git Extensions 3.4.1.9675
- Git 2.24.1.windows.2 (recommended: 2.25.1 or later)
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

RussKie:
Resolves #8263
Closes #8535 as obsolete

### Before

App crash

### After

image

## Test methodology 

- manual

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -824,7 +824,6 @@ namespace GitUI.CommandsDialogs
                 await TaskScheduler.Default;
 
                 var blob = Module.GetFileBlobHash(item.Item.Name, item.SecondRevision.ObjectId);
-
                 if (blob is null)
                 {
                     return;
@@ -832,7 +831,15 @@ namespace GitUI.CommandsDialogs
 
                 var fileName = PathUtil.GetFileName(item.Item.Name);
                 fileName = (Path.GetTempPath() + fileName).ToNativePath();
-                Module.SaveBlobAs(fileName, blob.ToString());
+                try
+                {
+                    Module.SaveBlobAs(fileName, blob.ToString());
+                }
+                catch (Exception ex)
+                {
+                    await this.SwitchToMainThreadAsync();
+                    ThrowUserExternalOperationException(ex.Message, command: null, arguments: item.Item.Name, ex);
+                }
 
                 onSaved(fileName);
             }).FileAndForget();
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -498,7 +498,14 @@ namespace GitUI.CommandsDialogs
                     "|All files (*.*)|*.*";
                 if (fileDialog.ShowDialog(this) == DialogResult.OK)
                 {
-                    Module.SaveBlobAs(fileDialog.FileName, selectedRows[0].Guid + ":\"" + orgFileName + "\"");
+                    try
+                    {
+                        Module.SaveBlobAs(fileDialog.FileName, selectedRows[0].Guid + ":\"" + orgFileName + "\"");
+                    }
+                    catch (Exception ex)
+                    {
+                        ThrowUserExternalOperationException(ex.Message, command: null, arguments: orgFileName, ex);
+                    }
                 }
             }
         }
--- a/GitUI/CommandsDialogs/FormVerify.cs
+++ b/GitUI/CommandsDialogs/FormVerify.cs
@@ -492,7 +492,14 @@ namespace GitUI.CommandsDialogs
                     };
                 if (fileDialog.ShowDialog(this) == DialogResult.OK)
                 {
-                    Module.SaveBlobAs(fileDialog.FileName, lostObject.ObjectId.ToString());
+                    try
+                    {
+                        Module.SaveBlobAs(fileDialog.FileName, lostObject.ObjectId.ToString());
+                    }
+                    catch (Exception ex)
+                    {
+                        ThrowUserExternalOperationException(ex.Message, command: null, arguments: null, ex);
+                    }
                 }
             }
         }
--- a/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
+++ b/GitUI/CommandsDialogs/RevisionFileTreeControl.cs
@@ -12,6 +12,7 @@ using GitExtUtils;
 using GitExtUtils.GitUI;
 using GitUI.CommandsDialogs.BrowseDialog;
 using GitUI.Hotkey;
+using GitUI.NBugReports;
 using GitUI.Properties;
 using GitUIPluginInterfaces;
 using Microsoft;
@@ -329,10 +330,17 @@ See the changes in the commit form.");
                 !string.IsNullOrWhiteSpace(gitItem.FileName))
             {
                 var fileName = gitItem.FileName.SubstringAfterLast('/').SubstringAfterLast('\\');
-
                 fileName = (Path.GetTempPath() + fileName).ToNativePath();
-                Module.SaveBlobAs(fileName, gitItem.Guid);
-                return fileName;
+
+                try
+                {
+                    Module.SaveBlobAs(fileName, gitItem.Guid);
+                    return fileName;
+                }
+                catch (Exception ex)
+                {
+                    ThrowUserExternalOperationException(ex.Message, command: null, arguments: gitItem.FileName, ex);
+                }
             }
 
             return null;
@@ -736,7 +744,14 @@ See the changes in the commit form.");
                 fileDialog.Filter = $@"{_saveFileFilterCurrentFormat.Text}(*{extension})|*{extension}| {_saveFileFilterAllFiles.Text} (*.*)|*.*";
                 if (fileDialog.ShowDialog(this) == DialogResult.OK)
                 {
-                    Module.SaveBlobAs(fileDialog.FileName, gitItem.Guid);
+                    try
+                    {
+                        Module.SaveBlobAs(fileDialog.FileName, gitItem.Guid);
+                    }
+                    catch (Exception ex)
+                    {
+                        ThrowUserExternalOperationException(ex.Message, command: null, arguments: gitItem.FileName, ex);
+                    }
                 }
             }
         }
--- a/GitUI/GitModuleControl.cs
+++ b/GitUI/GitModuleControl.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using System.Threading;
 using System.Windows.Forms;
 using GitCommands;
+using GitExtUtils;
+using GitUI.NBugReports;
 using ResourceManager;
 
 namespace GitUI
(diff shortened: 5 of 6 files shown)


mstv (commented):
This is a well chosen example for your UI-driven approach. It demonstrates some difficulties which must be considered.

pro:
- easy tailoring of additional error information
- no need for localization in core code - at least in this example, which is simple in this regard
- quick solution for the issue to be fixed by this PR

contra:
- Every event handler would need such wrapping!
- Technical issues like switching to the UI thread should be addressed in a single place.
- no info *which* operation failed: Triggered operations can be more complex and less obvious.
  Example: The commit message and the amend state is stored when closing FormCommit.
- no distinction between failed external operations and logic errors in called functions

to consider:
- distinguish between `UserExternalOperationException` and `ExternalOperationException` - https://github.com/gitextensions/gitextensions/pull/8906#discussion_r588929609
  The only thing the user can influence in these terms, is the filename - which should already have been checked by the FileDialog.
- GE is not Windows: Messages can be slightly more technical.
- The `ThrowEx` method is rather related to `GitModule` than to a Form or Control.
- SRP

lhiginbotham (approved):
@RussKie 
Looks good and seems simple, thanks!

RussKie:
> * Every event handler would need such wrapping!

Yes, and I'll explain the reasons why. On the flip side I don't believe we have too many cases where we must present exceptions in this way.

The way I look at the app is we have several distinct layers (that we have been trying to untangle for a few releases now) that live in the same repo, and thus make it harder to think of them as such.
We have:
1. API contracts - interfaces and base implementations, that must be used by the app and plugins
2. DAL/business logic layer - this wraps git commands and everything that can be shared between multiple UI clients (should we wish to build a WPF/Blazor/Avlaonia/etc. version)
3. Windows Forms UI app

The first two layers must be thread agnostic, and exceptions thrown from these layers can be anything. The UI layer is bound to UI thread, and must present any exceptions bubbled from the lower layers in a user-friendly manner. This involves
1. marshalling to the UI thread *if necessary*
2. wrapping exceptions in a unified manner, and
3. (re-)throwing the wrapped exception for the `BugReporter` to handle.

Step 1. must be done as the developer knows the threading situation
Step 2. must be done because different lower-layer exceptions may represent different classes of "user" exceptions, and thus can be handled differently. Also the lower-layer components can't be involved in localisation, it is the function of the UI layer.
Step 3. must be done **only** after 1\. and 2\. steps are done, otherwise the exception will be handled as "unhandled" exception leading to the termination of the app.


> * Technical issues like switching to the UI thread should be addressed in a single place.

Exceptions must be thrown on the UI thread, otherwise they will lead to termination of the app.

> * no info _which_ operation failed: Triggered operations can be more complex and less obvious.
>   Example: The commit message and the amend state is stored when closing FormCommit.

This is certainly fixable. And you've been actively working on improving this. :)

> * no distinction between failed external operations and logic errors in called functions

Perhaps I'm misunderstanding your point, but I don't think there's a blanket solution for this problem.

gerhardol (changes requested):
Missing RevDiff: saveAsToolStripMenuItem1_Click()

Have not run

I have no opinion on the mstv/RussKie solution, but suggest this is included in 3.5 now, adjust later if needed

mstv:
TL;DR: I guess the switch to the UI thread can and should be done by a wrapped `ThreadHelper.JoinableTaskFactory.RunAsync`. Is it worth that I propose an according implementation for 3.6?
In the meantime, this PR could be merged into 3.5 but not into the main branch IMO.

---

I would distinguish 2 1/2 types of exceptions:
1. logic error
2. external operation failure
2.5 failure of a user-defined external operation (https://github.com/gitextensions/gitextensions/pull/8906#discussion_r588929609)

Though all (these) types of exceptions can be thrown by any layer. Whether it is a logic error or a failure, is best known where the exception is thrown.
A `UserExternalOperationException` is *not* a `UIException`. Only the reaction in the `BugReporter` differs slightly, namely whether it should / might / not be reported as a bug.

When testing #8988, I created an `index.lock` and tried to stage a file. This failed silently, similar to #8632. Git operations must throw much more exceptions: #6527!

> The first two layers must be thread agnostic.

This is a sensible decision because there is no need in GE. (But it is not an a priori "must".)

> The UI layer is bound to UI thread, and must present any exceptions bubbled from the lower layers in a user-friendly manner. This involves
> 1. marshalling to the UI thread if necessary
> 2. wrapping exceptions in a unified manner, and
> 3. (re-)throwing the wrapped exception for the BugReporter to handle.

I see no need for point 2. What would be the purpose and the benefit of a unified wrapping?
Re-throwing - spread over many event handlers - annoys me.
So rather the `ThreadHelper` should create a wrapper instance of the `JoinableTaskFactory` whose `RunAsync` implementation switches to the UI thread on exception (what is not done by the library implementation for whatever reason). But this should work well for the case of GE.

> Also the lower-layer components can't be involved in localisation, it is the function of the UI layer.

I know your opinion but not the rationale other than saving a dependency. This might be an approach if it works well. But exceptions thrown by library functions *are* localized (and can contain filenames already).

> distinguish between `UserExternalOperationException` and `ExternalOperationException`

```
    /// <summary>
    /// Represents errors that occur during execution of user-configured operation, e.g. a script.
    /// </summary>
    public class UserExternalOperationException : ExternalOperationException
```

Regarding `SaveBlobAs`, the user usually is not responsible for the failed operation. Even if the user chose the filename, the `FileDialog` should have checked the validity. Hence just an `ExternalOperationException` should be thrown - on a git error or by the file operation(s).

RussKie:
> the switch to the UI thread can and should be done by a wrapped `ThreadHelper.JoinableTaskFactory.RunAsync`. Is it worth that I propose an according implementation for 3.6?

It is worth investigating definitely. Whether it will be a good change or not I can't say. @sharwell may have thoughts and opinions on this subject.

RussKie:
> But exceptions thrown by library functions _are_ localized (and can contain filenames already).


I don't believe it is the case. I may have missed something, but I've searched, and we only translate at the GitUI layer.
That was precisely the reason to introduce specific exceptions (e.g. `FileDeleteException`) that carried only data necessary to generate a loc'able user error message.

RussKie:
> Regarding `SaveBlobAs`, the user usually is not responsible for the failed operation. Even if the user chose the filename, the `FileDialog` should have checked the validity. Hence just an `ExternalOperationException` should be thrown - on a git error or by the file operation(s).

I'm of an opposite opinion - an error is likely outside the app, e.g. if a target file is locked, or the target drive is unavailable - it is the user-level issue, and the user must be able to solve it. 
No need to crash the app.

mstv:
> > But exceptions thrown by library functions _are_ localized (and can contain filenames already).

> I don't believe it is the case. I may have missed something, but I've searched, and we only translate at the GitUI layer.

I mean exceptions from the .NET libraries, e.g. from `FileSystem`. (Have a look at the exception message in the PR description.)

mstv:
>> Regarding `SaveBlobAs`, the user usually is not responsible for the failed operation. Even if the user chose the filename, the `FileDialog` should have checked the validity. Hence just an `ExternalOperationException` should be thrown - on a git error or by the file operation(s).

> I'm of an opposite opinion - an error is likely outside the app, e.g. if a target file is locked, or the target drive is unavailable - it is the user-level issue, and the user must be able to solve it. 
No need to crash the app.

Why do you think I want to crash the app? I just mean `ExternalOperationException` is the correcter choice (not `UserExternalOperationException`). See the different button and descriptive text in the TaskDialog (screenshots in the description of #8988).

sharwell:
> So rather the ThreadHelper should create a wrapper instance of the JoinableTaskFactory whose RunAsync implementation switches to the UI thread on exception (what is not done by the library implementation for whatever reason). But this should work well for the case of GE.

I don't see why this would be desirable.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -824,15 +824,22 @@ private void SaveSelectedItemToTempFile(Action<string> onSaved)
...
                 fileName = (Path.GetTempPath() + fileName).ToNativePath();
-                Module.SaveBlobAs(fileName, blob.ToString());
+                try
+                {
+                    Module.SaveBlobAs(fileName, blob.ToString());
+                }
+                catch (Exception ex)
+                {
+                    await this.SwitchToMainThreadAsync();
+                    ThrowUserExternalOperationException(ex.Message, command: null, arguments: item.Item.Name, ex);
```
sharwell, line 841: This shouldn't be necessary. Rather, you can update `FileAndForget` to catch this exception and report it on a UI thread (this is the purpose of `FileAndForget` as opposed to a plain `Forget` method).

mstv: Thank you for pointing this out, @sharwell! It works right out of the box because `FileAndForget` already catches exceptions, switches to the UI thread and calls `Application.OnThreadException`.

@RussKie, in which context must an exception been thrown so that it is reported via `AppDomain.CurrentDomain.UnhandledException` only?

RussKie: If an exception is thrown from a non-UI thread it will get handled by `AppDomain.CurrentDomain.UnhandledException` handler. 
Docs: https://docs.microsoft.com/dotnet/api/system.windows.forms.unhandledexceptionmode

PR #8962, opened 2021-03-08
Optimise string split operations

drewnoakes:
## Proposed changes

- Optimise string split operations
  - Avoid allocating a params array for most calls by keeping cached instances on `Delimiters` class.
  - Use `LazySplit` where possible to do so, to avoid allocating array for the result and internal arrays during the split operation.

## Test methodology 

- Manual testing

----

### DIFF
--- a/GitExtUtils/Delimiters.cs
+++ b/GitExtUtils/Delimiters.cs
@@ -2,16 +2,23 @@ using System.IO;
 
 namespace System
 {
+    /// <summary>
+    /// Singleton instances of commonly used character arrays.
+    /// </summary>
+    /// <remarks>
+    /// Using these instances avoids allocating an array for each invocation of methods
+    /// such as <c>string.Split</c>.
+    /// </remarks>
     internal static class Delimiters
     {
-        public static readonly char[] Newline = { '\n' };
+        public static readonly char[] LineFeed = { '\n' };
         public static readonly char[] Space = { ' ' };
         public static readonly char[] Tab = { '\t' };
         public static readonly char[] Null = { '\0' };
         public static readonly char[] TabAndSpace = { '\t', ' ' };
-        public static readonly char[] TabAndNewlineAndCarriageReturn = { '\t', '\n', '\r' };
-        public static readonly char[] NullAndNewline = { '\0', '\n' };
-        public static readonly char[] NewlineAndCarriageReturn = { '\n', '\r' };
+        public static readonly char[] TabAndLineFeedAndCarriageReturn = { '\t', '\n', '\r' };
+        public static readonly char[] NullAndLineFeed = { '\0', '\n' };
+        public static readonly char[] LineFeedAndCarriageReturn = { '\n', '\r' };
         public static readonly char[] ForwardSlash = { '/' };
         public static readonly char[] Colon = { ':' };
         public static readonly char[] Comma = { ',' };
--- a/GitExtUtils/Delimiters.cs
+++ b/GitExtUtils/Delimiters.cs
@@ -1,0 +1,22 @@
+using System.IO;
+
+namespace System
+{
+    internal static class Delimiters
+    {
+        public static readonly char[] Newline = { '\n' };
+        public static readonly char[] Space = { ' ' };
+        public static readonly char[] Tab = { '\t' };
+        public static readonly char[] Null = { '\0' };
+        public static readonly char[] TabAndSpace = { '\t', ' ' };
+        public static readonly char[] TabAndNewlineAndCarriageReturn = { '\t', '\n', '\r' };
+        public static readonly char[] NullAndNewline = { '\0', '\n' };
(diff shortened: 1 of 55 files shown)


RussKie:
Would `Span<T>` be better for splicing?

drewnoakes:
> Would `Span<T>` be better for splicing?

Possibly. I haven't investigated. The biggest saving here comes from the lack of arrays -- not just for the result of the split operation but also for internal working data that the .NET Framework implementation constructs.

For me, `Span<>` becomes more attractive once we move off .NET Framework, as many more APIs support it. The implementation of Span is faster on .NET Core+ too.

GitExtUtils/Delimiters.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal static class Delimiters
+    {
+        public static readonly char[] Newline = { '\n' };
+        public static readonly char[] Space = { ' ' };
+        public static readonly char[] Tab = { '\t' };
+        public static readonly char[] Null = { '\0' };
+        public static readonly char[] TabAndSpace = { '\t', ' ' };
+        public static readonly char[] TabAndNewlineAndCarriageReturn = { '\t', '\n', '\r' };
+        public static readonly char[] NullAndNewline = { '\0', '\n' };
+        public static readonly char[] NewlineAndCarriageReturn = { '\n', '\r' };
```
RussKie, line 14:
```suggestion
        public static readonly char[] CrLf = { '\n', '\r' };
```

GitExtUtils/Delimiters.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace System
+{
+    internal static class Delimiters
+    {
+        public static readonly char[] Newline = { '\n' };
+        public static readonly char[] Space = { ' ' };
+        public static readonly char[] Tab = { '\t' };
+        public static readonly char[] Null = { '\0' };
+        public static readonly char[] TabAndSpace = { '\t', ' ' };
+        public static readonly char[] TabAndNewlineAndCarriageReturn = { '\t', '\n', '\r' };
```
RussKie, line 12:
```suggestion
        public static readonly char[] TabAndCrLf = { '\t', '\n', '\r' };
```

GitExtUtils/LazyStringSplit.cs
```diff
@@ -0,0 +1,173 @@
...
+                Current = null!;
+            }
+
+            void IDisposable.Dispose()
+            {
+            }
+        }
+    }
+
+    internal static class LazyStringSplitExtensions
```
RussKie, line 116: Own file please

GitExtUtils/Delimiters.cs
```diff
@@ -0,0 +1,22 @@
...
+        public static readonly char[] NullAndNewline = { '\0', '\n' };
```
RussKie, line 13:
```suggestion
        public static readonly char[] NullAndLf = { '\0', '\n' };
```

RussKie (commented):
Overall LGTM
Have you done any perf benchmarks?

GitExtUtils/LazyStringSplit.cs
drewnoakes: This was intentional. The file is linked into multiple projects. Splitting it out will create more linked files, and create the opportunity for future usages to only link one file, where they then wouldn't benefit from these optimised replacements for common methods that would otherwise allocate.

GitExtUtils/Delimiters.cs
drewnoakes: I feel like "new line" is more commonly understood than "line feed", although "line feed" is more specific.

GitExtUtils/LazyStringSplit.cs
RussKie: Alright, leave it as is.

GitExtUtils/Delimiters.cs
RussKie: "New line" means different thing on different platforms.

GitExtUtils/Delimiters.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// such as <c>string.Split</c>.
+    /// </remarks>
+    internal static class Delimiters
+    {
+        public static readonly char[] LineFeed = { '\n' };
+        public static readonly char[] Space = { ' ' };
+        public static readonly char[] Tab = { '\t' };
+        public static readonly char[] Null = { '\0' };
+        public static readonly char[] TabAndSpace = { '\t', ' ' };
+        public static readonly char[] TabAndLineFeedAndCarriageReturn = { '\t', '\n', '\r' };
```
RussKie, line 19: Personally I'm finding this less readable than a shorter `TabAndCrLf`. Both `Cr` and `Lf` are established acronyms and both are present in git vocabulary as well.

RussKie (approved):
👍 overall but I do have naming concerns.

PR #8963, opened 2021-03-08
Improve revision graph drawing

ivangrek:
refers to #8941

## Proposed changes

- Remove dot filling to show that is not applied yet;
- Add a dot to the circle to indicate that it is done;
- Designate three states:  current, in progress, done;
- Remove redundant checks.

### Before

image

### After

image

image

image

image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build 3f977c1e31284d64d948328b5acca7bbfc8979ed
- Git 2.30.1.windows.1
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
@@ -69,12 +69,9 @@ namespace GitUI.UserControls.RevisionGrid.Columns
 
         public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevision revision, int rowHeight, in CellStyle style)
         {
-            if (AppSettings.ShowRevisionGridGraphColumn &&
-                e.State.HasFlag(DataGridViewElementStates.Visible) &&
-                e.RowIndex >= 0 &&
-                _revisionGraph.Count != 0 &&
-                _revisionGraph.Count > e.RowIndex &&
-                PaintGraphCell(e.RowIndex, e.CellBounds, e.Graphics))
+            if (_revisionGraph.Count > e.RowIndex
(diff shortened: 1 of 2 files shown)


mstv:
My 2ct: The dots inside the commit circles and squares attract my attention too much.
I'd rather fill it with a pale nuance of the color.

drewnoakes:
It's not clear to me what the dots are for. They look cool as a way of drawing attention to a node, but used this much are a bit distracting IMO.

Could this presentation be used for HEAD, for example? Or nodes involved in a bisect? Or fixup/squash nodes? These all require attention.

Another idea is to remove the centre of nodes that are not parents or children of the selected node. That might make it easier to follow ancestry as you click around the graph. It wouldn't need the dot though.

drewnoakes:
>> Could this presentation be used for HEAD
>
> I don't quite understand which section we are talking about.

I mean that only the checked out commit (HEAD) would be shown with a dot.

I feel that the dot is so visible &mdash; even moreso than the square &mdash; that it would help bring attention to the checked out commit.

>> Another idea is to remove the centre of nodes that are not parents or children of the selected node. That might make it easier to follow ancestry as you click around the graph. It wouldn't need the dot though.
>
> This is a great idea! Though I would invert it - show center dots only for those commits that are part of the selected tree, e.g. click on a node and all its parents get the dot. (Though I often want to trace the children too).

The hollow node is a nice alternative to desaturation (i.e. showing all non-reachable commits in grey). Dots add emphasis rather than taking it away. In the case of unreachable stuff, I think we want to de-emphasise parts of the graph, so would avoid using dots for that, personally.

gerhardol:
My proposal
* Unfilled for artificial as well as stash
* Filled for relatives of current checkout
* dot for not relatives (that you can draw gray now)

PR #8964, opened 2021-03-10
Annotate plugins

drewnoakes:
## Proposed changes

- Annotate plugins. This makes all non-test code in GE null annotated.
- Rework some project properties to reduce duplication and increase correctness.

## Test methodology 

- Manual testing. CI will run tests.

----

### DIFF
--- a/Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
+++ b/Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
@@ -37,7 +37,7 @@ namespace AppVeyorIntegration
                     return null;
                 }
 
-                return ".Net 4 full framework required";
+                return ".NET Framework 4 or higher required";
             }
         }
     }
--- a/Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
+++ b/Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
@@ -444,7 +444,7 @@ namespace AppVeyorIntegration
             };
         }
 
-        private Task<Stream> GetStreamAsync(HttpClient httpClient, string restServicePath, CancellationToken cancellationToken)
+        private Task<Stream?> GetStreamAsync(HttpClient httpClient, string restServicePath, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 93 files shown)


Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
```diff
@@ -458,7 +473,7 @@ private Task<Stream> GetStreamFromHttpResponseAsync(HttpClient httpClient, Task<
                 return task.CompletedResult().Content.ReadAsStreamAsync();
             }
 
-            return null;
+            return null; // TODO NULLABLE returning a null Task is a no no
```
RussKie, line 476: Should this be `Task.FromResult(null)`?

Plugins/Gource/GourceStart.cs
```diff
@@ -116,7 +116,7 @@ async Task DownloadImage((string email, string name) author)
                     var image = await AvatarService.DefaultProvider.GetAvatarAsync(author.email, author.name, imageSize: 90);
                     var filename = author.name + ".png";
 
-                    if (filename.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
+                    if (image == null || filename.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
```
RussKie, line 119:
```suggestion
                    if (image is null || filename.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
```

Plugins/Statistics/GitStatistics/FormGitStatistics.cs
```diff
@@ -82,7 +83,7 @@ public FormGitStatistics(IGitModule module, string codeFilePattern, bool countSu
             InitializeComplete();
         }
 
-        private void FormGitStatisticsSizeChanged(object sender, EventArgs e)
+        private void FormGitStatisticsSizeChanged(object? sender, EventArgs? e)
```
RussKie, line 86: I don't think this is correct. Handlers generally have the following annotation: `(object? sender, EventArgs e)`

drewnoakes: Agree this looks wrong. As before, I've just been annotating the current reality and fixing things up only when needed. I'll see if this is ready to change.

Plugins/BuildServerIntegration/AppVeyorIntegration/AppVeyorAdapter.cs
drewnoakes: Looks like it. Requires some other changes.

RussKie (approved):
If it works - 👍

drewnoakes:
Thanks @ivangrek

PR #8972, opened 2021-03-11
Use pattern matching

drewnoakes:
## Proposed changes

- Another usage of pattern matching.

## Test methodology 

- Manual testing.

----


RussKie (commented):
I don't necessarily find it clearer, but I won't block this

gerhardol (approved):
I believe I prefer this way

PR #8971, opened 2021-03-11
Support Bitbucket repos with dots in the name

AntonKedrov:
Fixes #8965

## Proposed changes

- Changed HTTP and SSH regex patterns to match URLs, which might contain a dot in the repo name

## Test methodology 

- Manual testing

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8974, opened 2021-03-11
Sign CLA

### LINKED ISSUE #8965: Bitbucket Plugin doesn't allow repositories with dot in the name
Hi! I'm switching from Sourcetree to Git Extensions. So far I really like it, but there is a small issue with Bitbucket Plugin.

When launching Bitbucket Plugin, it displays error "Your repository is not hosted in BitBucket Server." I think this is because of Regex pattern does not allow dots in the repository name.
https://github.com/gitextensions/gitextensions/blob/082d8c27da3e527b0d71019430ccecd995a8d878/Plugins/Bitbucket/Settings.cs#L13

Our repositories have at least one dot in the name, so URL looks like this:
`https://stash.company.com/scm/project_name/repo_type.project_group.app_name.app_type.git`
Could you please change pattern from `(?<repo>([\w\-]+))` to `(?<repo>([\w\-\.]+))`.

I have tried Bitbucket Plugin on the repo, which doesn't have dots in the name and it worked just fine.

AntonKedrov:
Fixes #8965

## Proposed changes

- Signed contributors.txt

----

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -152,3 +152,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/01/24, bcolaco, Bruno Colaço, bcolaco[@)gmail.com
 2021/01/25, guybark, Guy Barker, guybark(at)microsoft.com
 2021/02/10, mabako, Marcus Bauer, mabako(at)gmail.com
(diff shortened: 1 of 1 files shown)


AntonKedrov:
Added this change into #8971

PR #8990, opened 2021-03-13
Include stash, bisect in grid if tags are shown

### LINKED ISSUE #8879: "bisect" labels no longer displayed
## Current behaviour

image

## Expected behaviour

Git Extensions 3.4.3.9999
image

## Steps to reproduce

Start bisect and compare 3.4.3 version with the current dev

Git Extensions 3.4.3.9999

## Environment

- Git Extensions 33.33.33
- Build 221fc396b5db296c45d566a35a76facd6fbab38b (Dirty)

gerhardol:
Fixes #8879

## Proposed changes

Revert behavior to be same as in 3.4

If tag labels are shown, also show bisect, notes, stash etc

Note that the labels were orange in 3.4, changed in #8424

### Before

(slightly older than master)
image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -79,10 +79,11 @@ namespace GitCommands.Git.Commands
                 };
         }
 
-        public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
+        public static ArgumentString GetRefsCmd(GetRefsEnum getRef, bool noLocks, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
         {
             string format;
-            if (!tags)
+            bool hasTags = (getRef & GetRefsEnum.All) != 0 || (getRef & GetRefsEnum.Tags) != 0;
+            if (!hasTags)
             {
                 // If we don't need tags, it is easy.
                 format = @"--format=""%(objectname) %(refname)""";
@@ -107,14 +108,17 @@ namespace GitCommands.Git.Commands
                     ? (ArgumentString)"--no-optional-locks"
                     : default)
             {
-                { sortBy != GitRefsSortBy.Default, GetSortCriteria(tags, sortBy, sortOrder), string.Empty },
+                { sortBy != GitRefsSortBy.Default, GetSortCriteria(hasTags, sortBy, sortOrder), string.Empty },
                 format,
-                { branches, "refs/heads/", string.Empty },
-                { branches && tags, "refs/remotes/", string.Empty },
-                { tags, "refs/tags/", string.Empty },
+                { HasFlag(getRef, GetRefsEnum.Branches), "refs/heads/", string.Empty },
+                { HasFlag(getRef, GetRefsEnum.Remotes), "refs/remotes/", string.Empty },
+                { HasFlag(getRef, GetRefsEnum.Tags), "refs/tags/", string.Empty },
             };
 
             return cmd;
+
+            static bool HasFlag(GetRefsEnum requested, GetRefsEnum flag)
+                => (requested & GetRefsEnum.All) == 0 && (requested & flag) != 0;
 
             static string GetSortCriteria(bool needTags, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
             {
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2902,7 +2902,26 @@ namespace GitCommands
             return result;
         }
 
+        /// <summary>
+        /// Get the Git refs.
+        /// </summary>
+        /// <param name="tags">Include tags.</param>
+        /// <param name="branches">Include local branches, also remote branches if <see paramref="tags"/> is set.</param>
+        /// <returns>All Git refs.</returns>
         public IReadOnlyList<IGitRef> GetRefs(bool tags = true, bool branches = true)
+        {
+            return GetRefs(
+                (tags ? GetRefsEnum.Tags : GetRefsEnum.None)
+                | (branches ? GetRefsEnum.Branches : GetRefsEnum.None)
+                | (tags && branches ? GetRefsEnum.Remotes : GetRefsEnum.None));
+        }
+
+        /// <summary>
+        /// Get the Git refs.
+        /// </summary>
+        /// <param name="getRef">Combined refs to search for.</param>
+        /// <returns>All Git refs.</returns>
+        public IReadOnlyList<IGitRef> GetRefs(GetRefsEnum getRef)
         {
             // We do not want to lock the repo for background operations.
             // The primary use of 'noLocks' is to run git-status the commit count as a background operation,
@@ -2911,7 +2930,7 @@ namespace GitCommands
             // Assume that all GetRefs() are done in the background, which may not be correct in the future.
             const bool noLocks = true;
 
-            var cmd = GitCommandHelpers.GetRefsCmd(tags: tags, branches: branches, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
+            var cmd = GitCommandHelpers.GetRefsCmd(getRef, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
             var refList = _gitExecutable.GetOutput(cmd);
             return ParseRefs(refList);
         }
--- a/GitCommands/Git/Commands/GitRefsEnum.cs
+++ b/GitCommands/Git/Commands/GitRefsEnum.cs
@@ -1,0 +1,21 @@
+﻿using System.Diagnostics;
+using GitUIPluginInterfaces;
+
+namespace GitCommands.Git.Commands
+{
+    /// <summary>
+    /// Enums requestable in GitRefs() (multiple names can be appended)
+    /// Compare to <see ref="GitRefType"/> for actual values of parsed GitRefs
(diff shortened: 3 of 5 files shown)


RussKie:
Thank you.
Please let me tinker with it.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -109,9 +109,7 @@ public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks,
             {
                 { sortBy != GitRefsSortBy.Default, GetSortCriteria(tags, sortBy, sortOrder), string.Empty },
                 format,
-                { branches, "refs/heads/", string.Empty },
-                { branches && tags, "refs/remotes/", string.Empty },
-                { tags, "refs/tags/", string.Empty },
+                { !tags, "refs/remotes/ refs/heads/", string.Empty },
```
RussKie, line 112: This is incorrect, since this would never be true.

Since we can't filter at `` commands, we can fix it by delegating the filter to the bash:
```diff
                 format,
-                { branches, "refs/heads/", string.Empty },
-                { branches && tags, "refs/remotes/", string.Empty },
-                { tags, "refs/tags/", string.Empty },
+                { !branches, "| sed 'refs\\/heads\\//d", string.Empty },
+                { !(branches && tags), "| sed 'refs\\/remotes\\//d", string.Empty },
+                { !tags, "| sed 'refs\\/tags\\//d", string.Empty },
             };
```

RussKie: Here's the fix for the tests:
```diff
diff --git a/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs b/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
index ec2ab688c..62e35dee9 100644
--- a/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
+++ b/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
@@ -664,13 +664,14 @@ private static IEnumerable<TestCaseData> GetRefsCommandTestData
                         }
 
                         yield return new TestCaseData(/* tags */ true, /* branches */ true, /* noLocks */ false, sortBy, sortOrder,
-                            /* expected */ $@"for-each-ref{sortConditionRef}{sortCondition}{format} refs/heads/ refs/remotes/ refs/tags/");
+                            /* expected */ $@"for-each-ref{sortConditionRef}{sortCondition}{format}");
                         yield return new TestCaseData(/* tags */ true, /* branches */ false, /* noLocks */ false, sortBy, sortOrder,
-                            /* expected */ $@"for-each-ref{sortConditionRef}{sortCondition}{format} refs/tags/");
+                            /* expected */ $@"for-each-ref{sortConditionRef}{sortCondition}{format} | sed 'refs\/heads\//d | sed 'refs\/remotes\//d");
                         yield return new TestCaseData(/* tags */ false, /* branches */ true, /* noLocks */ false, sortBy, sortOrder,
-                            /* expected */ $@"for-each-ref{sortCondition} --format=""%(objectname) %(refname)"" refs/heads/");
+                            /* expected */ $@"for-each-ref{sortCondition} --format=""%(objectname) %(refname)"" | sed 'refs\/remotes\//d | sed 'refs\/tags\//d");
+
                         yield return new TestCaseData(/* tags */ false, /* branches */ true, /* noLocks */ true, sortBy, sortOrder,
-                            /* expected */ $@"--no-optional-locks for-each-ref{sortCondition} --format=""%(objectname) %(refname)"" refs/heads/");
+                            /* expected */ $@"--no-optional-locks for-each-ref{sortCondition} --format=""%(objectname) %(refname)"" | sed 'refs\/remotes\//d | sed 'refs\/tags\//d");
                     }
                 }
             }
```

gerhardol: The argument need to be added twice, in some other way. But the matching should be done on a positional base, sed will match in the middle of the refs too

gerhardol: I changed the solution to handle the RevGrid usage specially and explicitly, to really show that this is a special usage.
I looked at the usage for when both tags and branches were retrieved, they were all to get remotes - that should be a separate option too, but I do not want to change the IGitModule interface. Added some xml doc.

Note that stash, notes, bisect labels are always shown.

GitCommands/Git/GitModule.cs
```diff
@@ -2903,6 +2903,18 @@ public RemoteActionResult<IReadOnlyList<IGitRef>> GetRemoteServerRefs(string rem
...
+        }
+
+        /// <summary>
+        /// Get the Git refs.
+        /// </summary>
+        /// <param name="tags">Include tags.</param>
+        /// <param name="branches">Include local branches, also remote branches if <see paramref="tags"/> is set.</param>
+        /// <param name="all">Include all refs (override <see paramref="tags"/>, <see paramref="branches"/>). This includes stash, bisect, stash.</param>
+        /// <returns>All Git refs.</returns>
+        public IReadOnlyList<IGitRef> GetAllRefs(bool tags = true, bool branches = true, bool all = false)
```
RussKie, line 2917: If we're going down this path then we ought to create an enum:

Local
Remote
Tag
All = Local | Remote | Tag

GitCommands/Git/Commands/GitCommandHelpers.cs
RussKie: > The argument need to be added twice, in some other way. But the matching should be done on a positional base, sed will match in the middle of the refs too

To match the start of line use `^`

https://www.howtogeek.com/666395/how-to-use-the-sed-command-on-linux/

RussKie: > The argument need to be added twice, in some other way. 

Can you please clarify this bit?

RussKie:
The more pocessing we can offload to the underlying cli - the better, we'll do less work, less allocations, which is a win for our users.

GitCommands/Git/GitModule.cs
gerhardol: All is more than that, it was the problem here...
The handling of remotes is quite gross too. 

Can we make the change like this now and add an enum later in master?

RussKie:
Sure

GitCommands/Git/GitModule.cs
gerhardol: Can we change the IGitModule interface here?

RussKie: Yes

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -107,11 +108,11 @@ public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks,
...
             {
-                { sortBy != GitRefsSortBy.Default, GetSortCriteria(tags, sortBy, sortOrder), string.Empty },
+                { sortBy != GitRefsSortBy.Default, GetSortCriteria(hasTags, sortBy, sortOrder), string.Empty },
                 format,
-                { branches, "refs/heads/", string.Empty },
-                { branches && tags, "refs/remotes/", string.Empty },
-                { tags, "refs/tags/", string.Empty },
+                { (getRef & GetRefsEnum.All) == 0 && (getRef & GetRefsEnum.Branches) != 0, "refs/heads/", string.Empty },
+                { (getRef & GetRefsEnum.All) == 0 && (getRef & GetRefsEnum.Remotes) != 0, "refs/remotes/", string.Empty },
+                { (getRef & GetRefsEnum.All) == 0 && (getRef & GetRefsEnum.Tags) != 0, "refs/tags/", string.Empty },
```
RussKie, line 115: We're duplicating this code, so it is worth considering extracting it in a local static method:
```cs
static bool HasFlag(GetRefsEnum requested, GetRefsEnum flag)
    => (requested & GetRefsEnum.All) == 0 && (requested & flag) != 0;
```

RussKie (approved):
👍

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: changed

gerhardol:
Note that GetRefs are called six times during a refresh (probably four from the sidepanel).
The startup can be optimized...

Remotes were incorrectly handled in the latest commit, rebased and squashed.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -79,10 +79,11 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
                 };
         }
 
-        public static ArgumentString GetRefsCmd(bool tags, bool branches, bool noLocks, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
+        public static ArgumentString GetRefsCmd(GetRefsEnum getRef, bool noLocks, GitRefsSortBy sortBy, GitRefsSortOrder sortOrder)
         {
             string format;
-            if (!tags)
+            bool hasTags = (getRef & GetRefsEnum.All) != 0 || (getRef & GetRefsEnum.Tags) != 0;
```
RussKie, line 85: Did you miss this one?

RussKie (approved):

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: This is different - I can add a special if you prefer

RussKie:
No worries then

PR #8991, opened 2021-03-14
Inject current branch slug into the documentation link

### LINKED ISSUE #8955: Documentation links are broken

RussKie:
Resolves #8955

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using GitCommands.Settings;
 using GitCommands.Utils;
@@ -62,6 +63,7 @@ namespace GitCommands
         public static readonly string UserPluginsDirectoryName = "UserPlugins";
         private static string _applicationExecutablePath = Application.ExecutablePath;
         private static readonly ISshPathLocator SshPathLocatorInstance = new SshPathLocator();
+        private static string? _documentationBaseUrl;
 
         public static Lazy<string?> ApplicationDataPath { get; private set; }
         public static readonly Lazy<string?> LocalApplicationDataPath;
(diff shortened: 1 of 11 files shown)


gerhardol:
Better, but prefix need to be either /en/master or /en/release-xxx so logic (and XML doc) need to be adjusted.

RussKie:
> Better, but prefix need to be either /en/master or /en/release-xxx so logic (and XML doc) need to be adjusted.

Why?

You said you fixed the redirection, so https://git-extensions-documentation.readthedocs.io/settings.html#git-extensions-appearance-author-images-avatar-provider now correctly resolves to https://git-extensions-documentation.readthedocs.io/en/release-3.4/settings.html#git-extensions-appearance-author-images-avatar-provider.
So the master branch points to the latest.

Meanwhile this change makes a release build point to the respective doc's branch.

Am I overlooking something?

gerhardol:
> Am I overlooking something?

With the issues I saw with the redirection, I believe it is better to not rely on it.

If / points to the development version, that is what users will see by default too (currently 3.5 but it should be 3.6 or so).

RussKie:
Ok, added "en/latest" for all non "release/x.y" branches

GitCommands/Settings/AppSettings.cs
```diff
@@ -117,6 +119,41 @@ static AppSettings()
...
+        }
+
+        internal static void SetDocumentationBaseUrl(string currentGitBranch)
+        {
+            if (_documentationBaseUrl is not null)
+            {
+                throw new InvalidOperationException("Documentation base URL can only be set once");
+            }
+
+            string? docVersion = "en/latest/";
```
gerhardol, line 142: (not null)
Suggest we use master to point out that this is dev version of the doc (where the version could be updated)
We have used 'latest' for the released version so far

```suggestion
            string docVersion = "en/master/";
```

RussKie: Given the controversies around the "master", let's have it "en/dev"

gerhardol: Then we chanhange the name of the "master" branch too.
I prefer "main" though, just as GH changed (and some longtime ClearCase usage).

RussKie:
sure

gerhardol (approved):

gerhardol:
I reordered Settings-Scripts so I could push main branch and build on read-the-docs so links in the app works.

PR #8992, opened 2021-03-14
Include spelling dictionaries

### LINKED ISSUE #8338: Spelling dictionaries missing from portable zip
The Dictionaries\ folder is missing from the portable zip release breaking spell check in editor like the commit dialogue.

## Expected behaviour
Spelling check works in the commit dialogue.

## Steps to reproduce

Try to spell check in the commit message dialogue with a portable zip release.

The last zip with a Dictionaries\ folder was v3.3.1

RussKie:
Resolves #8338


gerhardol (approved):
Files are included in the portable, so verified.

Unfortunately, the portable package is 5 MB bigger and unpacked 20 MB larger - could be an issue for those running from an USB stick (where the workaround is to delete dictionaries).

RussKie:
USB stick capacities are long measured in tens of GBs, and some in TB, so
20MB is hardly an issue. ;)

PR #8995, opened 2021-03-14
Resolve PluginSettings key names issue

### LINKED ISSUE #5797: PluginSettings key names
Bug  related: #5793
Plugin settings are stored with keys like this

```xml
<item>
    <key>
      <string>Periodic background fetchArguments of git command to run</string>
    </key>
    <value>
      <string>fetch --all</string>
    </value>
  </item>
```
The issue is someone could develop a plugin and name their plugin the same as another plugin that we don't know even exists. 

Use a unique value per plugin for the settings.  

I recommend each plugin use a guid that NEVER changes to retrieve its settings.  

```xml
<item>
    <key>
      <string>2ec6be62-d9e2-4b2b-a06b-e037b580e201_Arguments of git command to run</string>
    </key>
    <value>
      <string>fetch --all</string>
    </value>
  </item> 
```

- Git Extensions 3.00.00.03-RC2
- 8cbcd2b25223c73fbe9f7dad08e498cf82734bfa  (Dirty)
- Git 2.19.1.windows.1
- .NET Framework 4.7.3221.0
- DPI X:100.00 % Y:100.00 %

ivangrek:
Fixes #5797, https://github.com/gitextensions/gitextensions.extensibility/issues/14

## Proposed changes

- Don't use the plugin description field and switch to using the name field;
- Use Guid Id as plugin settings key.

### Before

image

### After

image

image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build f1944e19892542fa61f2efd29cb01ba1b9f114b4
- Git 2.30.2.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -117,7 +117,7 @@ namespace GitUI.Script
                 {
                     foreach (var plugin in PluginRegistry.Plugins)
                     {
-                        if (string.Equals(plugin.Description, command, StringComparison.CurrentCultureIgnoreCase))
+                        if (string.Equals(plugin.Name, command, StringComparison.CurrentCultureIgnoreCase))
                         {
                             var eventArgs = new GitUIEventArgs(owner, uiCommands);
                             return new CommandStatus(executed: true, needsGridRefresh: plugin.Execute(eventArgs));
--- a/ResourceManager/GitPluginBase.cs
+++ b/ResourceManager/GitPluginBase.cs
@@ -12,6 +12,8 @@ namespace ResourceManager
     [UsedImplicitly]
     public abstract class GitPluginBase : IGitPlugin, ITranslate
     {
+        public Guid Id { get; protected set; }
+
         public string? Description { get; protected set; }
         public string? Name { get; protected set; }
         public Image? Icon { get; protected set; }
--- a/ResourceManager/GitPluginBase.cs
+++ b/ResourceManager/GitPluginBase.cs
@@ -30,7 +30,6 @@ namespace ResourceManager
         protected void SetNameAndDescription(string name)
         {
             Name = name;
-            Description = name;
         }
 
         void IDisposable.Dispose()
--- a/ResourceManager/GitPluginBase.cs
+++ b/ResourceManager/GitPluginBase.cs
@@ -79,6 +79,8 @@ namespace ResourceManager
 
         protected void Translate()
         {
+            // Description for old plugin setting processing as key
+            Description = Name;
             Translator.Translate(this, AppSettings.CurrentTranslation);
         }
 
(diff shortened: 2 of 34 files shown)


RussKie (approved):

RussKie:
Now I have an itch to convert our xml-based configs into json-based, so we can have a little more structure and hierarchy... 🤔

GitUI/Script/ScriptRunner.cs
```diff
@@ -117,7 +117,7 @@ private static CommandStatus RunScriptInternal(IWin32Window owner, IGitModule mo
                 {
                     foreach (var plugin in PluginRegistry.Plugins)
                     {
-                        if (string.Equals(plugin.Description, command, StringComparison.CurrentCultureIgnoreCase))
+                        if (string.Equals(plugin.Name, command, StringComparison.CurrentCultureIgnoreCase))
```
ivangrek, line 120: The search will be based on the localized name.

ResourceManager/GitPluginBase.cs
```diff
@@ -30,7 +30,6 @@ protected GitPluginBase()
         protected void SetNameAndDescription(string name)
         {
             Name = name;
-            Description = name;
```
ivangrek, line 33: I think it should be left for the settings. As a non-localized key. For legacy plugins to work.

RussKie:
This is a breaking change, all plugins will have to be updated.

ivangrek:
> This is a breaking change, all plugins will have to be updated.

Removing support for outdated plugins and moving on?

```
// for old plugin setting processing
if (value is null || value.Equals(defaultValue))
{
    value = ExternalSettings.GetValue($"{_pluginName}{name}", defaultValue, decode);
}
```

GitUI/Script/ScriptRunner.cs
ivangrek: Ok.

ivangrek:
Left the Description as a key for deprecated plugins.

RussKie (approved):

RussKie:
Thank you!

Changes like this highlight the urgency of moving the shared contracts to https://github.com/gitextensions/gitextensions.extensibility.
I know you've attempted to do that, though the changes you proposed were a little too radical.

mast-eu (approved):
Looks good to me.

https://github.com/gitextensions/gitextensions.pluginmanager and https://github.com/gitextensions/GitExtensions.GerritPlugin should be updated accordingly.

CC: @maraf , because I'm not sure whether he has some other plugin which saves settings.

maraf:
> CC: @maraf , because I'm not sure whether he has some other plugin which saves settings.

Thank you for the notice. I have plugins that save settings and will update them accordingly.

PR #9001, opened 2021-03-16
Check BeforeCommit script failure to abort commit

### LINKED ISSUE #2479: GitExtensions apparently ignoring script failure
Had a bit of a look through previous issues, didn't find anything similar.  I apologise if I've missed something.

I've been trying to add a pre-commit hook to lint php files before committing - the idea being if at least one file has a syntax error, abort the commit, don't hit the repo, and I avoid embarrassing myself by letting my co-devs see that I've let them see such a simple error.

On its own, the php lint file (see below) ran fine.

However, integration doesn't seem to work so well - when the hook fires and fails (nonzero return value), I first get a dialog with the output (telling me what file blew up) and a red cross in circle, which lead me to expect that, since the hook failed, nothing further would happen.  Nope.

After clicking 'OK', another dialog comes up telling me that, despite the earlier failure, it's busily committing the change that contains a syntax error.

I'm using GitExtensions 2.47.3 on a Windows Server 2012 box (don't ask).  I've probably missed something - does anyone have an idea _what_?  What's the bet it's something quite simple?

Script details, via Tools > Settings > Git Extensions > Scripts:
Name: PHP Lint
Command: php
Arguments: -d safe_mode=Off "C:\stuff\phplint_precommit.php"
Event: BeforeCommit

php file I'm using for the precommit hook:

``` php
<?php

$output = array();
$return = 0;
$php = 'c:/php/php.exe';

exec('git rev-parse --verify HEAD', $output, $return);
$against = $return == 0 ? 'HEAD' : '4b825dc642cb6eb9a060e54bf8d69288fbee4904';

exec("git diff-index --name-only {$against}", $output);

$filename_pattern = '/\.php$/';
$exit_status = 0;

foreach ($output as $file) {

    if (!preg_match($filename_pattern, $file)) {
        // don't check files that aren't PHP
        continue;
    }

    $lint_output = array();
    exec("{$php} -lf " . escapeshellarg($file), $lint_output, $return);
    if ($return == 0) {
        continue;
    }
    echo implode("\n", $lint_output), "\n";
    $exit_status = 1;
}

exit($exit_status);
```

JuPrgn:
Should fix #2479

## Intro
I am not familiar with C # programming (first C# coding). And I don't have the environment to compile and test this modification but since it seems like a simple fix with few lines, I guess it can be easily checked.

## Proposed changes
My proposal is that when a `BeforeCommit `script is called and returns a failure, the commit action should be aborted.
This script failure detection should probably be applicable to other cases (`BeforePush`... ?).

I have a doubt on the fact that this function : https://github.com/gitextensions/gitextensions/blob/e721aadada0ed6c0df8435c36df7357d2c8e957c/GitUI/Script/ScriptRunner.cs#L48
returns `Executed : false` when a script failure happen (`exit 1`) or if this function just launch the script and the output information should be found somewhere else. 

## Test methodology 
I proposed test steps on https://github.com/gitextensions/gitextensions/issues/2479#issuecomment-800172413

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1320,7 +1320,12 @@ namespace GitUI.CommandsDialogs
                                                                        ensureCommitMessageSecondLineEmpty: AppSettings.EnsureCommitMessageSecondLineEmpty);
                     }
 
-                    ScriptManager.RunEventScripts(this, ScriptEvent.BeforeCommit);
+                    bool success = ScriptManager.RunEventScripts(this, ScriptEvent.BeforeCommit);
+
+                    if (!success)
+                    {
+                        return;
+                    }
 
                     var commitCmd = Module.CommitCmd(
                         amend,
@@ -1331,7 +1336,7 @@ namespace GitUI.CommandsDialogs
                         gpgSignCommitToolStripComboBox.SelectedIndex > 0,
                         toolStripGpgKeyTextBox.Text);
 
-                    bool success = FormProcess.ShowDialog(this, process: null, arguments: commitCmd, Module.WorkingDir, input: null, useDialogSettings: true);
+                    success = FormProcess.ShowDialog(this, process: null, arguments: commitCmd, Module.WorkingDir, input: null, useDialogSettings: true);
 
                     UICommands.RepoChangedNotifier.Notify();
 
--- a/GitUI/Script/ScriptManager.cs
+++ b/GitUI/Script/ScriptManager.cs
@@ -54,12 +54,18 @@ namespace GitUI.Script
             return null;
         }
 
-        public static void RunEventScripts(GitModuleForm form, ScriptEvent scriptEvent)
+        public static bool RunEventScripts(GitModuleForm form, ScriptEvent scriptEvent)
         {
             foreach (var script in GetScripts().Where(scriptInfo => scriptInfo.Enabled && scriptInfo.OnEvent == scriptEvent))
             {
-                ScriptRunner.RunScript(form, form.Module, script.Name, form.UICommands, revisionGrid: null);
+                var result = ScriptRunner.RunScript(form, form.Module, script.Name, form.UICommands, revisionGrid: null);
+                if (!result.Executed)
+                {
+                    return false;
+                }
             }
+
+            return true;
         }
 
         public static string? SerializeIntoXml()
--- a/GitUI/Script/ScriptRunner.cs
+++ b/GitUI/Script/ScriptRunner.cs
@@ -151,7 +151,11 @@ namespace GitUI.Script
 
             if (!scriptInfo.RunInBackground)
             {
-                FormProcess.ShowDialog(owner, command, argument, module.WorkingDir, null, true);
+                bool success = FormProcess.ShowDialog(owner, command, argument, module.WorkingDir, null, true);
+                if (!success)
+                {
+                    return false;
+                }
             }
             else
             {


JuPrgn:
Testing using build artefacts has no effect commit is not aborted.

pmiossec:
>I am not familiar with C # programming (first C# coding). And I don't have the environment to compile and test this modification but since it seems like a simple fix with few lines, I guess it can be easily checked.

If you look at the APpVeyor build result, you will be able to download a portable version and test it.
The last artifact here:
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/38252181/artifacts

At some point, if you want to see this PR merges, you will have to squash the commits.

Note: Create a PR just for the contributor.txt change could be a good idea to merge it quickly and reduce the risk of conflicts when we would like to merge your PR...

JuPrgn:
Thanks I figured it out with the checks and tested it without success.

Ok I will do that !

pmiossec:
> Thanks I figured it out with the checks and tested it **without** success.


Not sure to fully understand. You tested it **with** or **without** success?

JuPrgn:
Without success. Script failure does not cancel commit as I expected...

ivangrek (approved):
🚀

RussKie:
You can build and run locally:
`.\build.cmd -launch`

You can edit the code and VS Code or any other text editor you fancy.

JuPrgn:
This is working now ! 

I tested with a script containing : 
- `exit 1` the commit is aborted and the "commit stage dialog" is kept open.
- `exit 0` the commit is done

Please note that you may want to apply the same functionality to other `Before` script events. I do not yet have practical cases for these other events but please let me know if you want to apply this on these too.

mstv (approved):

RussKie:
Thank you!

gerhardol:
This could be a breaking change for users where the script happens to fail...
If added to pre-commit, it should be handled to other states though, hard enough to document why pre-commit is different.

JuPrgn:
You are welcome !
Ok I will add this also to others `Before` states thank you for your feedback.

JuPrgn:
> This could be a breaking change for users where the script happens to fail...
Maybe we could add a feature to enable/disable script failure on script settings ?

gerhardol:
> > This could be a breaking change for users where the script happens to fail...
> 
> Maybe we could add a feature to enable/disable script failure on script settings ?

GE has too many options already....

The change makes sense though, so I believe it is not strange to enforce this.
More natural in BeforeCommit than for instance BeforeFetch, but it is still something that is easy to explain, then it is often the way to go.

So I would like to see this for all Before.

Should also be documented in https://git-extensions-documentation.readthedocs.io/en/release-3.4/settings.html#git-extensions-scripts
(That is also missing the new grouping added in 3.4.)

JuPrgn:
@gerhardol this is implemented for all `Before` in #9011 
I will add documentation but I do not know what is the new grouping added in 3.4 ?

gerhardol:
> @gerhardol this is implemented for all `Before` in #9011

great

> I will add documentation but I do not know what is the new grouping added in 3.4 ?

Script behavior for instance is grouping some options

image

JuPrgn:
An important feature is missing, script are configured as `Global for all repositories` there is no option to configure them as `Local for current repository`. Users may have to enable and disable scripts every time they work on another repository.
I do not know the amount of work required to add this setting option but this would be really helpfull especially now that a failing/missing script is blocking the action.

pmiossec:
@JuPrgn It's better to open a new issue to discuss about that...

gerhardol:
> Script behavior for instance is grouping some options

I changed this (since I needed to make a push anyway to trigger branch updates).

mstv:
> An important feature is missing, script are configured as Global for all repositories there is no option to configure them as Local for current repository.

Git itself supports hook scripts. Why would one want to tie this to an exchangable Git UI client (despite nobody really wants to switch to an other than GE 😄)?

Issue #8890, opened 2021-02-25
GitConfigurationException: Cannot open the application


gerhardol:
## Current behaviour

(opened for colleague, not having the problem myself)

The default setup at work is to set HOME to H: that is synced to the local PC with DFS.
HOMEDRIVE\HOMEPATH is default like C:\Users\username

In some situation the sync fails and the files on H: cannot be accessed, the user can see that files exist but not access them.
This applies for both .settings, .gitconfig and ssh keys.

When starting up on 3.4 or latest portable version, you get multiple pops about inaccessible settings, eventually you get the NBug with the info below. The user had problems getting into the settings, succeeeded eventually but could not change the settings, it seemed to reset back to H: This was also the case for a fresh portable installation.

The eventual workaround was to edit the .settings file and provide another path with a .gitconfig file.

```xml
  <item>
    <key>
      <string>customhomedir</string>
    </key>
    <value>
      <string>C:\cygwin\home\userid</string>
    </value>
  </item>
```

## Expected behaviour

Limited number of popups and at least a possibility to change the settings.

Many users would not even understand that the .gitconfig setting need to be changed, so some automatic fallback would be appreciated.

## Steps to reproduce

I have not succeeded to get the sync in this state myself other than by inserting an error:
```c#
public ConfigFileParser(ConfigFile configFile)
            {
                _configFile = configFile;
            }

            public void Parse(string? fileContent = null)
            {
                _fileContent = fileContent ?? throw new IOException("xxx");
```

A partial reproduce:

* Copy .gitconfig to another directory
* Change GE Git dir to this directory
* Make the directory inaccessible (set deny or something)

## Error Details

<details>
  <summary>Click to expand!</summary>

```
GitCommands.Config.GitConfigurationException: Exception of type 'GitCommands.Config.GitConfigurationException' was thrown. ---> System.IO.IOException: This file is currently not available for use on this computer.

 at void System.IO.__Error.WinIOError(int errorCode, string maybeFullPath)
at void System.IO.FileStream.Init(string path, FileMode mode, FileAccess access, int rights, bool useRights, FileShare share, int bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, string msgPath, bool bFromProxy, bool useLongPath, bool checkHost)
at new System.IO.FileStream(string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options, string msgPath, bool bFromProxy, bool useLongPath, bool checkHost)
[...]

PR #9003, opened 2021-03-16
FileSettingsCache: No exception if Git config is inaccessible

gerhardol:
Part of #8890

## Proposed changes

If the .gitconfig file is not accessible at startup you should get a prompt asking to change.
This was not happening in #8890 though, the GE config had to be edited manually.
This PR eliminates some additional exceptions at many operations due to that the FileSettingsCache cannot enable the FileSystemWatcher.
Most normal operation after this.

Note that it is still not possible to store settings, even if .gitconfig is not changed at all.

### Before

image

### After

Normal operation - mostly

## Test methodology 

Manual
Repro (part of the original issue) in the issue

----

### DIFF
--- a/GitCommands/Settings/FileSettingsCache.cs
+++ b/GitCommands/Settings/FileSettingsCache.cs
@@ -46,7 +46,7 @@ namespace GitCommands.Settings
                 dir = null;
             }
 
-            if (Directory.Exists(dir))
+            if (Directory.Exists(dir) && File.Exists(SettingsFilePath))
             {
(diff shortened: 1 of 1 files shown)


RussKie (approved):

gerhardol:
@RussKie will you cherry-pick to 3.5 or should I make a PR?

RussKie:
Either way will work.

There's a 3rd option too - you can cherrypick it and push ;)

gerhardol:
done

RussKie:
Champion!

PR #9010, opened 2021-03-17
Allow opening files in Visual Studio

drewnoakes:
Adds a context menu item in `FileStatusList` that opens the selected file in Visual Studio.

- If Visual Studio is not running (or installed), the item is not shown.
- If Visual Studio is running, but the file is not present on disk, the item is disabled (matching "Show in folder" in the same menu)
- If no running Visual Studio instance has a project containing the file, a message is displayed to the user explaining that fact.

## Screenshots 

image

image

image

When the file is not on disk:

image

## Test methodology 

- Manual testing

## Test environment(s)

- Windows 10
- 150% scaling

----


GitUI/VisualStudioIntegration.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    internal static class VisualStudioIntegration
+    {
+        public static bool TryOpenFile(string filePath)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            if (!File.Exists(filePath))
+            {
+                // When opening the context menu, we disable this item if the file does not exists.
```
drewnoakes, line 19:
```suggestion
                // When opening the context menu, we disable this item if the file does not exist.
```

RussKie:
This is an awesome feature! 
But at the same time it gives me an easy feeling - we'll be baking in and taking a dependency on  a product-specific functionality, which also needs to be bundled up in the setup.
I don't have data to back my claim, but I don't think all our users use VS, and as a result we'll be serving a subset of our users.

 I'd be very much interested in a plugin that enables a tighter VS integration, e.g. I want to open sln and csproj files.

pmiossec:
First a question: Why is it not in the "open section"? Like just after "Open this revision with..."

> I don't have data to back my claim, but I don't think all our users use VS, and as a result we'll be serving a subset of our users.

Yes, I also think that. I could be very convenient for us dotnet guys but I wonder why more Visual Studio than any other tool.
I would find much more interesting to have a "My favorite editor" feature but looking at your PR it would be difficult to do it because it is looking for an open instance...

> I'd be very much interested in a plugin that enables a tighter VS integration, e.g. I want to open sln and csproj files.

There is this extension that does it for sln files at the moment: https://www.neptuo.com/project/desktop/gitextensions.solutionrunner/

Perhaps it's also not as convenient but you could drag and drop into VisualStudio...

RussKie:
A bit of OT:

GE has almost become the hub-app for me - I do spend a lot (and I mean a
lot) of time in GE switching between different repos and branches, and I
use the console tab to run a lot of builds and scripts from it (hurray for
powershell, `dotnet` and its tools). But every now and then I find myself
wanting to open a sln file (not from a cli) and opening Explorer just to
double click on a sln file. Doing so from GE would be great.

At the same time I do recall a discussion in
https://github.com/gitextensions/gitextensions/pull/6135 which I didn't
like for the same reason - that time adding OS specific implementations.

We should be able to leverage scripts to achieve a significant part of the
desired functionality, but the scripting engine needs to be extended to the
File Tree control.

ivangrek:
👍 
This adds value. But to understand this, we need to look not at the finger, but where it points.

drewnoakes:
> we'll be baking in and taking a dependency on a product-specific functionality, which also needs to be bundled up in the setup.

This is true. I don't think this is a problem, per se.

> I don't have data to back my claim, but I don't think all our users use VS, and as a result we'll be serving a subset of our users.

Also true. GE has many features. Most users only use a subset of them.

The way I implemented it, if you aren't running VS, the item does not even appear. Users who don't use VS won't see any change.

> Why is it not in the "open section"? Like just after "Open this revision with..."

Ideally it would live there. I couldn't get that to work cleanly because the context menu handling is a mess. There are loads of users of this `FileStatusList` and they all provide their own context menus to it. I would have to copy/paste this code in lots of places to make it work. I didn't have the time to try and rationalise all the various menus either.

> I would find much more interesting to have a "My favorite editor" feature but looking at your PR it would be difficult to do it because it is looking for an open instance...

Nothing stopping us adding equivalent menu items for other editors.

For example, we could add an "Open external" item that would launch the file in the shell, to pick up the system-default editor. There are possible concerns here though, if the file is executable for example.

> every now and then I find myself wanting to open a sln file (not from a cli) and opening Explorer just to double click on a sln file. Doing so from GE would be great.

It would be possible to extend this PR so that when it doesn't find a running instance, it walks up the repo looking for a .sln file and offers to launch that instead. Doing this would mean we need to determine if VS is installed, as otherwise we show the menu item to people who don't even have VS installed. Right now we avoid this by not showing it if it's not running.

drewnoakes:
> This adds value. But to understand this, we need to look not at the finger, but where it points.

Could you elaborate?

drewnoakes:
Also, I encourage you to try this branch out locally. It feels pretty magical 😃

gerhardol:
Does not compile for me - what do I need to change?

Error	CS0177	The out parameter 'prot' must be assigned to before control leaves the current method	GitUI	C:\dev\gc\gitextensions\GitUI\VisualStudioIntegration.cs	95	Active

drewnoakes:
@gerhardol what version of VS are you using? I can't see why you'd be seeing this error, other than using an older version of the C# compiler that doesn't support extern local functions.

drewnoakes:
@gerhardol looks like you're hitting https://github.com/dotnet/roslyn/issues/49500 which was fixed in VS 16.9.

gerhardol:
> @gerhardol looks like you're hitting [dotnet/roslyn#49500](https://github.com/dotnet/roslyn/issues/49500) which was fixed in VS 16.9.

Yes - I was on 1.8.2, thought I had just updated.

Works well for me. I just wish this was available in FileTree too.

VS Code would be even nicer. I normally have at least three VSCode windows open, this would simplify open a file in the correct workspace.

To open solution (projects) and VSC workspaces would be nice, but I do not see how you can find the 'right' project to open.
VSC could open a folder, but in my case the workspace has specific settings, it is generated too. So that will have to be done from the command line.

pmiossec:
> 
> 
> @gerhardol looks like you're hitting [dotnet/roslyn#49500](https://github.com/dotnet/roslyn/issues/49500) which was fixed in VS 16.9.

Not blocking but I think that this dependency to the very last version of VS will be a pain to handle.
All occasional contributors will have to update to contribute and we will have to remind that often...

GitUI/Translation/English.xlf
```diff
@@ -1358,6 +1358,10 @@ The primary difftool can still be selected by clicking the main menu entry.</sou
         <source>No changes</source>
         <target />
       </trans-unit>
+      <trans-unit id="_openInVisualStudioMenuItem.Text">
```
gerhardol, line 1361: Unused translation item - should not be needed

GitUI/UserControls/FileStatusList.cs
```diff
@@ -160,6 +163,57 @@ static Bitmap ScaleHeight(Bitmap input)
...
+                            TranslatedStrings.OpenInVisualStudioFailureCaption,
+                            MessageBoxButtons.OK,
+                            MessageBoxIcon.Warning);
+                    }
+                };
+                return item;
+            }
+        }
+
+        private string? SelectedItemAbsolutePath
```
gerhardol, line 204: Suggest using _fullPathResolver instead

gerhardol (commented):
+1
Works well.
Can we set the minimal MSVS version to avoid  the strange compile error?

drewnoakes:
> Not blocking but I think that this dependency to the very last version of VS will be a pain to handle.
> All occasional contributors will have to update to contribute and we will have to remind that often...

Yes it's easy enough to avoid the dependency. I can back out the local extern methods.

GitUI/Translation/English.xlf
drewnoakes: This is also true of the sibling `_openSubmoduleMenuItem` translation. I've renamed to prefix with `_NO_TRANSLATE_`.

drewnoakes:
I've backed out use of language features that require recent VS versions in order to compile.

gerhardol (approved):

PR #9027, opened 2021-03-21
#9013 Use Close in Clean working directory

### LINKED ISSUE #9013: Remove "Close"/"Cancel" button in Form that just close the windows
## Feature description

Lost of windows don't have a close button. So to close the window, I need to click on the red cross.

Example with bisect form
image

But there is a few window that have a dedicated close button.

Example with "Clean working directory" form
image

More important, in the "Clean working directory" form, the "Cancel" button only close the window. This could confuse the user because you can not cancel anything (like if you have clicked on the "Cleanup" button).

## New behavior

I suggest to remove the close/cancel button in all windows where the button just close the UI.

After a quick review, the forms are "Clean working directory", "Reset changes", "Recover lost objects".

bansan85:
Fixes #9013 

## Proposed changes

- Replace Cancel by Close to avoid confusion.

### Before

image

### After

image

## Test methodology 

Not needed

## Test environment(s) 

- Git Extensions 33.33.33
- Build dd34ed04f62d67884675667f83493ac36e88a954 (Dirty)
- Git 2.30.0.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -1680,10 +1680,6 @@ You will have push access. {2}</source>
         <source>Clone</source>
         <target />
       </trans-unit>
-      <trans-unit id="closeBtn.Text">
-        <source>Close</source>
-        <target />
-      </trans-unit>
       <trans-unit id="columnHeaderMyReposForks.Text">
         <source># Forks</source>
         <target />
@@ -3012,10 +3008,6 @@ revision:</source>
       </trans-unit>
       <trans-unit id="AddPath.Text">
         <source>Add a path...</source>
-        <target />
-      </trans-unit>
-      <trans-unit id="Cancel.Text">
-        <source>Cancel</source>
         <target />
       </trans-unit>
       <trans-unit id="Cleanup.Text">
@@ -4965,10 +4957,6 @@ command "git help shortlog"</source>
       </trans-unit>
       <trans-unit id="_switchWorktreeTitle.Text">
         <source>Open a worktree</source>
-        <target />
-      </trans-unit>
-      <trans-unit id="buttonClose.Text">
-        <source>Close</source>
         <target />
       </trans-unit>
       <trans-unit id="buttonPruneWorktrees.Text">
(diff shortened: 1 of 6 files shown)


GitUI/Translation/English.xlf
```diff
@@ -3050,6 +3046,10 @@ revision:</source>
         <source>Cleanup</source>
         <target />
       </trans-unit>
+      <trans-unit id="buttonClose.Text">
```
gerhardol, line 3049: There is some way to avoid that this text (as well as the other using the new translated text), I do not recall how...

gerhardol (commented):
Looks good, just there are unneeded translation items...

GitUI/Translation/English.xlf
bansan85: Maybe remove the xx.Text="..." in FormCleanupRepository.Designer.cs and move it in the constructor of the form?

gerhardol: Should be OK, I believe there is some other trick too.
(This could maybe be fixed in a follow up, there are likely other similar strings.)

pmiossec: In in front of my computer, so you have to have a look at another form but to prevent the string to be translated, you should rename the button with the prefix looking link `__no_translate`

bansan85: I just tried, it doesn't work :/

mstv: The prefix is `_NO_TRANSLATE_`.

vbjay:
Look at  control names with no translate in name.

bansan85:
Thanks, I changed names.

RussKie:
There's no need for the Close or Cancel button.

bansan85:
@RussKie Are you sure ? According to https://github.com/gitextensions/gitextensions/issues/9013#issuecomment-802773243 the question looks not answered. Would you like I remove the button only in "Clean working directory" ? Or also in "Reset changes" and "Recover lost objects" ?

RussKie:
I don't believe this dialog requires a Close button.

gerhardol:
For a dialog like this I believe that there should be a close button, nicr for the user than using the upper right close.

RussKie:
Other dialogs don't have "Close" button, and I can't see why this dialog needs it either.
"Cancel" button could be useful _only_ if the clean operation could take significant time _and_ can be cancelled. Even in this situation "Cancel" button must either be hidden or disabled until the operation is in flight.

RussKie:
Thank you

bansan85:
Thanks, I didn't know what to do with this PR since some people wanted to remove this button and some other don't wanted.

RussKie:
I still want it removed ;)

PR #9030, opened 2021-03-22
New ISettingsSource interface

ivangrek:
## Proposed changes

- Add and implement new ISettingsSource interface for preparing to move to Extensibility.

## Test methodology 

- Manually
- Visually
- Current Unit tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build f9bed12a3a03ea80edec4b9a3d0fa890ef780604
- Git 2.30.2.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
+++ b/Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
@@ -1,0 +1,46 @@
+﻿namespace GitUIPluginInterfaces.Extensibility
+{
+    public interface ISettingsSource
+    {
+        /// <summary>
+        /// Extracts the string value with the specified key as is.
+        /// </summary>
+        /// <param name="key">The key of the setting value.</param>
+        /// <returns>The string value.</returns>
+        string GetValue(string key);
+
+        /// <summary>
+        /// Extracts the string value with the specified key as is or use default value.
+        /// </summary>
+        /// <param name="key">The key of the setting value.</param>
+        /// <param name="defaultValue">The default value to use if no value is found.</param>
+        /// <returns>The string value.</returns>
+        string GetValue(string key, string defaultValue);
+
+        /// <summary>
+        /// Extracts the value with the specified key and converts it to type T or create new.
+        /// </summary>
+        /// <typeparam name="T">The type to convert the value to.</typeparam>
+        /// <param name="key">The key of the setting value.</param>
+        /// <returns>The converted value to T or new T().</returns>
+        T GetValue<T>(string key)
+            where T : new();
+
+        /// <summary>
+        /// Extracts the value with the specified key and converts it to type T or use default value.
+        /// </summary>
+        /// <typeparam name="T">The type to convert the value to.</typeparam>
+        /// <param name="key">The key of the setting value.</param>
+        /// <param name="defaultValue">The default value to use if no value is found.</param>
+        /// <returns>The converted value to T.</returns>
+        T GetValue<T>(string key, T defaultValue);
+
+        /// <summary>
+        /// Stores the value with the specified key.
+        /// </summary>
+        /// <typeparam name="T">The type to convert the value from.</typeparam>
+        /// <param name="key">The key of the setting value.</param>
+        /// <param name="value">The value to save.</param>
+        void SetValue<T>(string key, T value);
+    }
+}
--- a/Plugins/GitUIPluginInterfaces/ISettingsSource.cs
+++ b/Plugins/GitUIPluginInterfaces/ISettingsSource.cs
@@ -21,6 +21,7 @@ namespace GitUIPluginInterfaces
 
         public abstract void SetValue<T>(string name, T value, Func<T, string?> encode);
 
+        [Obsolete("Use new GetValue<T> method.")]
         public bool? GetBool(string name)
         {
             return GetValue<bool?>(name, null, x =>
@@ -39,21 +40,25 @@ namespace GitUIPluginInterfaces
             });
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public bool GetBool(string name, bool defaultValue)
         {
             return GetBool(name) ?? defaultValue;
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetBool(string name, bool? value)
         {
             SetValue(name, value, b => b.HasValue ? (b.Value ? "true" : "false") : null);
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetInt(string name, int? value)
         {
             SetValue(name, value, b => b.HasValue ? b.ToString() : null);
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public int? GetInt(string name)
         {
             return GetValue<int?>(name, null, x =>
@@ -67,11 +72,13 @@ namespace GitUIPluginInterfaces
             });
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetFloat(string name, float? value)
         {
             SetValue(name, value, b => b.HasValue ? b.ToString() : null);
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public float? GetFloat(string name)
         {
             return GetValue<float?>(name, null, x =>
@@ -85,16 +92,19 @@ namespace GitUIPluginInterfaces
             });
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public DateTime GetDate(string name, DateTime defaultValue)
         {
             return GetDate(name) ?? defaultValue;
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetDate(string name, DateTime? value)
         {
             SetValue(name, value, b => b?.ToString("yyyy/M/dd", CultureInfo.InvariantCulture));
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public DateTime? GetDate(string name)
         {
             return GetValue<DateTime?>(name, null, x =>
@@ -108,31 +118,37 @@ namespace GitUIPluginInterfaces
             });
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public int GetInt(string name, int defaultValue)
         {
             return GetInt(name) ?? defaultValue;
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetFont(string name, Font value)
         {
             SetValue(name, value, x => x.AsString());
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public Font GetFont(string name, Font defaultValue)
         {
             return GetValue(name, defaultValue, x => x.Parse(defaultValue));
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public Color GetColor(string name, Color defaultValue)
         {
             return GetValue<Color?>(name, null, x => ColorTranslator.FromHtml(x)) ?? defaultValue;
         }
 
+        [Obsolete("Use new SetValue<T> method.")]
         public void SetEnum<T>(string name, T value) where T : Enum
         {
             SetValue(name, value, x => x.ToString());
         }
 
+        [Obsolete("Use new GetValue<T> method.")]
         public T GetEnum<T>(string name, T defaultValue) where T : struct, Enum
         {
             return GetValue(name, defaultValue, x =>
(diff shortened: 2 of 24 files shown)


Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace GitUIPluginInterfaces.Extensibility
+{
+    public interface ISettingsSource
+    {
+        T GetValue<T>(string key, T defaultValue = default!);
```
drewnoakes, line 5: The `!` here seems like a problem. Can you experiment with some different options here?

What will the implementation return if `defaultValue` is `null`? 

It's possible to annotate this as (from memory):

```suggestion
        [return: NotNullIfNotNull("defaultValue")]
        T? GetValue<T>(string key, T? defaultValue = default);
```

`GetValue` and `SetValue` should be documented.

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
```diff
@@ -165,15 +183,126 @@ public Color GetColor(string name, Color defaultValue)
...
 
+        [Obsolete("Use new GetValue<T> method.")]
         [return: NotNullIfNotNull("defaultValue")]
         public string? GetString(string name, string? defaultValue)
         {
             return GetValue(name, defaultValue, x => x);
         }
+
+#pragma warning disable CS0618 // Type or member is obsolete
+        public T GetValue<T>(string key, T defaultValue = default!)
```
drewnoakes, line 200: Why move to an interface that loses type information (pushing it to `T`) then introduce dispatch on the value of `T`?

Why not just move `GetBool`, `GetInt`, ... to the interface?

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
ivangrek: I think this operator for this case.

`var person = new Person(null!);`

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

> What will the implementation return if defaultValue is null?

In runtime will be ok. With null as value will be ok.

ivangrek: > GetValue and SetValue should be documented.

👍

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
ivangrek: > Why move to an interface that loses type information (pushing it to T) then introduce dispatch on the value of T?

As far as I understand, we are not losing information. I am planning further steps and the current implementation will be changed.

> Why not just move GetBool, GetInt, ... to the interface?

After adding new types, we will have to change the interface. In the current case, we only add the implementation.
As I understand it, the generic types are for this.

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
drewnoakes: I'm familiar with the `!` operator, and we should avoid using it as it can hide bugs. The Nullable Reference Types feature is design-time only, so it's true that this code might be fine, but consuming code will assume the value cannot be null, and then may receive a null at runtime and blow up.

I think the annotations here have to be changed.

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
drewnoakes: > As far as I understand, we are not losing information

I was referring to runtime information. The `switch` over `typeof(T)` will happen at runtime and has a cost associated with it. This cost did not exist previously.

Is there a benefit to changing this that I don't see?

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
ivangrek: 👍

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
ivangrek: I understand that this adds transformations that were not there, but this is temporary.
This is required to work with the current implementation.
I announced further thoughts in #9032.

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Extracts the value with the specified key and converts it to type T.
+        /// </summary>
+        /// <typeparam name="T">The type to convert the value to.</typeparam>
+        /// <param name="key">The key of the setting value.</param>
+        /// <param name="defaultValue">The default value to use if no value is found.</param>
+        /// <returns>The converted value to T.</returns>
+        [return: NotNullIfNotNull("defaultValue")]
+        T? GetValue<T>(string key, T? defaultValue = default);
```
RussKie, line 15:
```suggestion
        T? Get<T>(string key, T? defaultValue = default);
```

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        /// <summary>
+        /// Stores the value with the specified key.
+        /// </summary>
+        /// <typeparam name="T">The type to convert the value from.</typeparam>
+        /// <param name="key">The key of the setting value.</param>
+        /// <param name="value">The value to save.</param>
+        void SetValue<T>(string key, T value);
```
RussKie, line 23:
```suggestion
        void Set<T>(string key, T value);
```

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
```diff
@@ -165,15 +191,127 @@ public Color GetColor(string name, Color defaultValue)
...
+            {
+                return (T)(object)GetFont(key, (Font)(object)defaultValue!);
+            }
+
+            throw new InvalidOperationException($"Unknown type: {type}.");
+        }
+
+        public void SetValue<T>(string key, T value)
+        {
+            var type = typeof(T);
```
RussKie, line 268: Can't we write as follows?
```cs
if (value is bool boolValue) 
    SetBool(key, boolValue); return;
if (value is int intValue) 
    SetInt(key, intValue); return;
...
```

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
RussKie: > Why move to an interface that loses type information (pushing it to `T`) then introduce dispatch on the value of `T`?
> 
> Why not just move `GetBool`, `GetInt`, ... to the interface?

I personally find `GetBool` et al. rather distasteful, and an unnecessary cognitive load. All changes outside this class look great - the public API surface is simple and clean. The changes here do look ugly and may result in additional boxing.

We should be able to provide additional overloads for value-types, using `struct` and `class` constraints. 🤔

drewnoakes: > I personally find GetBool et al. rather distasteful, and an unnecessary cognitive load.

Can you elaborate? This pattern is not common in the BCL, for example. Also accepting `T` means you might expect `GetValue<MyCustomType>` to work, but it fails at runtime instead of compile time.

ivangrek: > Also accepting T means you might expect GetValue<MyCustomType> to work, but it fails at runtime instead of compile time.

Are you suggesting returning null instead of throwing an exception?

drewnoakes: No, I'm suggesting that the API shouldn't allow using invalid types.

Look at the methods on [`BinaryReader`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binaryreader?view=net-5.0#methods) for example. It has a method per type.

[`BinaryWriter`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binarywriter?view=net-5.0#methods) does as well, though they're overloads with the same name.

Not just in the BCL either. Check out JSON.NET's [`JsonReader`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonReader.htm) class too.

ivangrek:
> We should be able to provide additional overloads for value-types, using struct and class constraints. 🤔

I don't see the need to make separate methods for struct and class.

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
ivangrek: 👍

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
ivangrek: I think "GetValue" will be clearer since we have a source.

`Source.GetValue` 

It's not as clear what we are getting as it seems to me.

`Source.Get`

Plugins/GitUIPluginInterfaces/Extensibility/ISettingsSource.cs
ivangrek: https://github.com/gitextensions/gitextensions/pull/9030#discussion_r600408121

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
ivangrek: Thank. I'll see.
But this interface will work with all types. This was not required to integrate with the current implementation, so I did not extend it. Since I am planning further steps.

drewnoakes: Will it actually be extensible? All types would have to be in that switch statement, no?

ivangrek: If we talk about the current implementation, then yes. For objects, we could add.

`JsonConvert.DeserializeObject<T>(value);`

or from Microsoft

`JsonSerializer.Serialize<T>(value);`

In any case, consumers are implementation independent.

ivangrek: > Look at the methods on BinaryReader for example. It has a method per type.
>
>BinaryWriter does as well, though they're overloads with the same name.
>
>Not just in the BCL either. Check out JSON.NET's JsonReader class too.

BinaryReader and BinaryWriter.
I think they are done this way, because they work only for these types.

JsonReader.
Because the JSON format has these types in the specification or so.

ivangrek: Please see last commit. I think this will be more clear interface.

ivangrek:
Changed ISettingsSource interface.

Plugins/GitUIPluginInterfaces/ISettingsSource.cs
ivangrek: Did such a test.

image

image

https://github.com/dotnet/runtime/blob/7795971839be34099b07595fdcf47b95f048a730/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs#L124-L147

I don't know how it works correctly, but null returns. Perhaps we shouldn't complicate things too? 😄

RussKie: > Look at the methods on [`BinaryReader`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binaryreader?view=net-5.0#methods) for example. It has a method per type.
> 
> [`BinaryWriter`](https://docs.microsoft.com/en-us/dotnet/api/system.io.binarywriter?view=net-5.0#methods) does as well, though they're overloads with the same name.
> 
> Not just in the BCL either. Check out JSON.NET's [`JsonReader`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonReader.htm) class too.

This is a good point. I didn't consider these.

RussKie: > [dotnet/runtime@`7795971`/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs#L124-L147](https://github.com/dotnet/runtime/blob/7795971839be34099b07595fdcf47b95f048a730/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs#L124-L147)
> 
> I don't know how it works correctly, but null returns. Perhaps we shouldn't complicate things too? 😄

To use this I'd guess we'd need to provide same kind of plumbing ASP.NET Core uses... It's been awhile since I used ASP.NET Core last ime, and haven't really looked in how it plumbs configs.

I think it would be prudent of us to retain `ReadXyz` methods, where `Xyz` denotes a known type, and have a single `Write(Xyz)` method with required overloads.

drewnoakes: Consider whether extension methods could play a role here too. It looks like the underlying data is `string`. If that's the case, the interface could work only with `string` and all the typed get/set methods could be extensions.

Issue #4631, opened 2018-03-14
Support for --allow-empty option in "git commit"


pankajazad:
**Do you want to request a *feature* or report a *bug*?**
Feature
**What is the current behavior?**
Non-existent
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
Capability of committing with no changes in files 
i.e. $git commit --allow-empty -m "Message..."
Adding a checkbox  "allow empty" in commit dialog besides the amend commit will suffice.

RussKie:
Please clarify the use case

vbjay:
If the files are not changing then why not just amend this commit or if you
want to modify another commit, reword that commit.  We have a reword menu
option in advanced now but you can get the idea here.
http://git-extensions-documentation.readthedocs.io/en/latest/modify_history.html?highlight=reword

pankajazad:
Sometimes you need to perform a GIT commit without changing any file in the repository:
1. For triggering a new build
2. For testing the workflow (GIT hooks, builds, releases, etc.)
3. My special use cases - It can be used for bare minimum problem/bug reporting with using specialized bug tracking tools.

pmme:
I often like to start a repo with an empty commit in case I want to make an interactive rebase or other easily able to modify the first real commit. 
Similar to this article scrounged from google's results: https://coderwall.com/p/m_pgbg/start-git-repos-with-an-empty-commit

vbjay:
https://git-scm.com/docs/git-rebase#git-rebase---root

vbjay:
You don't need to do that.  Most build servers have a link on a commit that
allows you to request a new build.  Notice the new build button.
https://www.appveyor.com/assets/img/posts/nuget-support/nuget-project-feed1.01559bf8.png.
Better way is to generate a file that has info like current sha, date/time
and commit that file.  No need for empty commits.

pmme:
Can you easily run a ``git rebase --root`` from GE?

vbjay:
![image](https://user-images.githubusercontent.com/3629489/37445171-9de92834-27ec-11e8-894d-7d0b33ae4974.png)

vbjay:
![image](https://user-images.githubusercontent.com/3629489/37445423-5203ecea-27ee-11e8-944e-18519217d6c5.png)
[...]

PR #8563, opened 2020-10-21
Draw a separator line at the top of the footer panel, similar to what Task Dialog does

RussKie:
# Screenshots 

* Task Dialog
image

* GitExtensionsDialog

| Before | After |
|---|---|
| image | image |
| image  | image |

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9044, opened 2021-03-31
Added 'Allow Empty' checkbox to commit form, and updated command logic

darkcamper:
Implements #4631

## Proposed changes

-  Add new option to the commit dialog to allow commits with no changes.

### Before

before-4631

### After

after-4631

## Test methodology 
Manual testing

## Test environment(s) 
- GIT 2.29
- Windows 10, 20H2

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -3364,16 +3364,25 @@ i.e. it may no longer be a fixup or an autosquash commit.</source>
 Are you sure you want to commit?</source>
         <target />
       </trans-unit>
-      <trans-unit id="_noFilesStagedAndNothingToCommit.Text">
-        <source>There are no files staged for this commit.</source>
+      <trans-unit id="_noFilesStagedCommitAllFilteredUnstagedOption.Text">
+        <source>Stage and commit the unstaged files that match your filter</source>
         <target />
       </trans-unit>
-      <trans-unit id="_noFilesStagedButSuggestToCommitAllFilteredUnstaged.Text">
-        <source>There are no files staged for this commit. Stage and commit the unstaged files that match your filter?</source>
+      <trans-unit id="_noFilesStagedCommitAllUnstagedOption.Text">
+        <source>Stage and commit all unstaged files</source>
         <target />
       </trans-unit>
-      <trans-unit id="_noFilesStagedButSuggestToCommitAllUnstaged.Text">
-        <source>There are no files staged for this commit. Stage and commit all unstaged files?</source>
+      <trans-unit id="_noFilesStagedCommitCaption.Text">
+        <source>Confirm commit</source>
+        <target />
+      </trans-unit>
+      <trans-unit id="_noFilesStagedCommitInstructions.Text">
+        <source>There aren't any changes in the staging area.
+How do you want to proceed?</source>
+        <target />
+      </trans-unit>
+      <trans-unit id="_noFilesStagedMakeEmptyCommitOption.Text">
+        <source>Make an empty commit</source>
         <target />
       </trans-unit>
       <trans-unit id="_noStagedChanges.Text">
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2067,7 +2067,7 @@ namespace GitCommands
             }
         }
 
-        public ArgumentString CommitCmd(bool amend, bool signOff = false, string author = "", bool useExplicitCommitMessage = true, bool noVerify = false, bool gpgSign = false, string gpgKeyId = "")
+        public ArgumentString CommitCmd(bool amend, bool signOff = false, string author = "", bool useExplicitCommitMessage = true, bool noVerify = false, bool gpgSign = false, string gpgKeyId = "", bool allowEmpty = false)
         {
             return new GitArgumentBuilder("commit")
             {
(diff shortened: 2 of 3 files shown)


pmiossec:
We need your update on the `contributors.txt` file, if not done previously, preferably in another pull request.

And you need to update the translation: https://github.com/gitextensions/gitextensions/wiki/Translations#v34-and-newer

darkcamper:
Ok! I've updated this PR with the updated translation file and I'll open another PR for the contributors file

GitUI/Translation/English.xlf
```diff
@@ -3201,6 +3201,10 @@ gitex.cmd / gitex (located in the same folder as GitExtensions.exe):</source>
         <source>Commit</source>
         <target />
       </trans-unit>
+      <trans-unit id="AllowEmpty.Text">
+        <source>Allow &amp;Empty</source>
```
mstv, line 3205:
```suggestion
        <source>Allow &amp;empty</source>
```
Could you also align the `&Amend commit` checkbox to our capitalization convention, please?

mstv (commented):
Please revert any unnecessary changes to `FormCommit.Designer.cs`. Or else create a separate commit with how the designer saves the file without UI changes. Then add the new checkbox.
Have you adapted the tab order?
Force-pushing this feature branch is welcome.

The check state of the `Allow empty` must affect the `CommitAndPush` button text in a similar way as `Amend commit` does.

RussKie:
This looks to me as a rather narrow use case, and with that I oppose to the
checkbox being added to the main screen.

It can probably be added to the dropdown menu that hosts additional options
(e.g. "refresh on focus", "keep open", etc.)

drewnoakes:
This adds too much complexity to the UI in my opinion. I prefer allowing the user to commit an empty message without the checkbox. Perhaps a dialog box is shown saying "No commit message has been entered. Commit anyway?" With ok/cancel.

RussKie:
I could agree to this.

darkcamper:
@drewnoakes @RussKie, just to clarify, this feature would allow you to make a commit with no changes (no file additions/modifications), but the message would still be mandatory.

I think the dialog box would work really well when there are no changes in the working directory:

`'There are no files staged for this commit' (Ok) => 'There are no files staged for this commit. Commit anyway?' (Yes/No)`

But when there are pending changes (not staged), there is already a dialog asking if you want to stage everything, and commit.

`
'There are no files staged for this commit. Stage and commit all unstaged files?' (Yes/No)`

One possibility would be adding a "Cancel" option

```
'There are no files staged for this commit. Stage and commit all unstaged files?' (Yes/No/Cancel)
--> Yes: Old 'yes' behavior
--> No: 'There are no files staged for this commit. Commit anyway? (Yes/No)' dialog
--> Cancel: Old 'no' behavior
```

Although I'm not sure if it's a bit confusing. What do you think?

pmiossec:
Perhaps you could use a task dialog (like in https://github.com/gitextensions/gitextensions/pull/8563 ) when their is no files staged with 2 possible actions "Stage all files and commit"(disabled if no files changed) and "Create an empty commit". And indeed a cancel button.

RussKie:
That could work

darkcamper:
It seems like a better idea. I'm giving it a try

darkcamper:
taskdialog-emptycommit_disablestage

Well, this is the result. Anything you'd change?

pmiossec:
Was there an icon in the message box you replaced?
If yes, I would display the same...
If not, I propose (wait for other advices...) to add the warning icon.

darkcamper:
There was a question mark icon, but it's not available in TaskDialog (and it seems that Microsoft is advising against using question icons outside 'Help entry points'...)  

If having an icon is preferred I'd use the warning icon too

drewnoakes:
I like the improvements here 👍 

Please also verify that the <kbd>Esc</kbd> key is equivalent to cancelling the dialog. (I think that is the default behaviour with `TaskDialog`, but please check.)

GitUI/CommandsDialogs/FormCommit.Designer.cs
```diff
@@ -23,61 +23,62 @@ private void InitializeComponent()
         {
```
RussKie, line 23: The designer is evil... 
Please undo _all_ unrelated changed.

RussKie (commented):
> taskdialog-emptycommit_disablestage
> 
> Well, this is the result. Anything you'd change?

Nice!

I'd like to ask for a few changes:
1. Please remove the trailing dots in the command links.
2. Use the same icon as in the dialog below.
3. I'd change the heading to something like "Confirm commit"

E.g. 
image

darkcamper:
@RussKie Ok to those changes! (they'll certainly make the new dialog more consistent with the rest of the application). I'll also make sure that all the mess that the winforms designer introduced is cleaned up.

darkcamper:
While doing some tests I came up with a question about the "Push/Commit & Push" button. Should this new feature affect in any way that button, when there aren't unstaged changes? (the case when there are unstaged changes is already covered by the new dialog)

Previously, when you couldn't make empty commits, it made sense for the button to only allow "pushing to remote" in that situation. But now, I wonder if we should ask the user if they want to "make an empty commit & push/push" or just leave that button as is.

gerhardol:
> 
> 
> While doing some tests I came up with a question about the "Push/Commit & Push" button. Should this new feature affect in any way that button, when there aren't unstaged changes? (the case when there are unstaged changes is already covered by the new dialog)
> 
> Previously, when you couldn't make empty commits, it made sense for the button to only allow "pushing to remote" in that situation. But now, I wonder if we should ask the user if they want to "make an empty commit & push/push" or just leave that button as is.

A special case to a special case, not so important to me.
If you bother, "make an empty commit & push/push" seems more complete to me.

RussKie:
How would it look like in English and other languages we support?

darkcamper:
After some consideration, I think that overloading the 'Push/Commit & Push' button with yet another option could be confusing and would bring little benefit. After all, If you want to create & push an empty commit just can just use the 'Commit' button, choose the 'Make empty commit' in the new dialog, and then press the 'Push' button.

Besides I can't really think of a good caption/flow for that button, if we finally try to squeeze the 'make empty commit' into it.

Naturally, if anyone cames up with a good idea, I'd be willing to make the necessary changes.

darkcamper:
I've force pushed all the changes we've talked about (well, I did it a week ago but forgot to post a comment 😓).

RussKie (approved):

RussKie:
Thank you

darkcamper:
Thank you all for your help with my first contribution!

PR #9048, opened 2021-04-03
Suppress `ThrowIfNotOnUIThread` for `BugReportForm`

### LINKED ISSUE #9047: `BugReportForm` not shown for exceptions in background threads
## Current behaviour

GE terminates after the TaskDialog displayed for an exception from a background thread
due to a check for UI thread in `GitExtensionsFormBase`, which is wrong / not necessary in this case

## Expected behaviour

GE terminates after `BugReportForm`

## Steps to reproduce

- throw an exception from e.g. `ProcessWrapper.OnProcessExit`
- click `Report bug!` --> GE terminates
## Screenshots

grafik

I doubt it

## Environment

- Git Extensions 33.33.33
- Build ac6487fd05262fc38490cd18b0eabe6bfd321edc
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

mstv:
Fixes #9047

## Proposed changes

- Suppress `ThrowIfNotOnUIThread()` in `GitExtensionsControlInitialiser` for `BugReportForm`

### Before

GE terminated

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 4735576e5f1744068d25760d459b58ed29d9cdac
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/ResourceManager/GitExtensionsControlInitialiser.cs
+++ b/ResourceManager/GitExtensionsControlInitialiser.cs
@@ -15,11 +15,7 @@ namespace ResourceManager
 
         public GitExtensionsControlInitialiser(GitExtensionsFormBase form)
         {
-            if (form.GetType().Name != "BugReportForm")
-            {
-                ThreadHelper.ThrowIfNotOnUIThread();
-            }
-
+            ThreadHelper.ThrowIfNotOnUIThread();
             form.Load += LoadHandler;
             _translate = form;
         }
--- a/ResourceManager/GitExtensionsControlInitialiser.cs
+++ b/ResourceManager/GitExtensionsControlInitialiser.cs
@@ -13,9 +13,13 @@ namespace ResourceManager
         /// <summary>indicates whether the initialisation has been signalled as complete.</summary>
         private bool _initialiseCompleteCalled;
 
-        public GitExtensionsControlInitialiser(GitExtensionsFormBase form)
+        public GitExtensionsControlInitialiser(GitExtensionsFormBase form, bool allowNonUiThread)
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            if (!allowNonUiThread)
+            {
+                ThreadHelper.ThrowIfNotOnUIThread();
+            }
+
             form.Load += LoadHandler;
             _translate = form;
         }
--- a/ResourceManager/GitExtensionsControlInitialiser.cs
+++ b/ResourceManager/GitExtensionsControlInitialiser.cs
@@ -13,9 +13,9 @@ namespace ResourceManager
         /// <summary>indicates whether the initialisation has been signalled as complete.</summary>
         private bool _initialiseCompleteCalled;
 
-        public GitExtensionsControlInitialiser(GitExtensionsFormBase form, bool allowNonUiThread)
+        public GitExtensionsControlInitialiser(GitExtensionsFormBase form)
         {
-            if (!allowNonUiThread)
+            if (form.GetType().Name != "BugReportForm")
             {
                 ThreadHelper.ThrowIfNotOnUIThread();
             }
--- a/GitUI/GitExtensionsForm.cs
+++ b/GitUI/GitExtensionsForm.cs
@@ -29,8 +29,7 @@ namespace GitUI
         /// <summary>Creates a new <see cref="GitExtensionsForm"/> indicating position restore.</summary>
         /// <param name="enablePositionRestore">Indicates whether the <see cref="Form"/>'s position
         /// will be restored upon being re-opened.</param>
-        protected GitExtensionsForm(bool enablePositionRestore, bool allowNonUiThread = false)
-            : base(allowNonUiThread)
+        protected GitExtensionsForm(bool enablePositionRestore)
         {
             var needsPositionSave = enablePositionRestore;
             _needsPositionRestore = enablePositionRestore;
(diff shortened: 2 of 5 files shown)


gerhardol (commented):
Seem OK
Can you add how you provoked the problem to the description?

RussKie:
This feels wrong to me...

mstv:
> Can you add how you provoked the problem to the description?

> #9047: throw an exception from e.g. ProcessWrapper.OnProcessExit

d9fe6ba2939aba6507a5321860c0c2b26acc7ea8

---

> This feels wrong to me...

Then the plumbed but pragmatic variant.

> Or move all strings to TranslatedStrings class (which I think is a more
strategic solution).

Tinkering the `BugReportForm` feels wrong to me.

ResourceManager/GitExtensionsControlInitialiser.cs
```diff
@@ -15,7 +15,11 @@ internal sealed class GitExtensionsControlInitialiser
 
         public GitExtensionsControlInitialiser(GitExtensionsFormBase form)
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            if (form.GetType().Name != "BugReportForm")
```
drewnoakes, line 18: Checking for a type by name like this is a code smell. What is wrong with passing a boolean around to control this?

RussKie:
This a very specific case. Why should this be exposed to other forms?

Feels like the bug report form should be invoked on the UI thread.

RussKie:
I played with your proposal locally using the suggestion to throw an exception from `ProcessWrapper.OnProcessExit`, and... it doesn't quite work.

```diff
diff --git a/GitCommands/Git/Executable.cs b/GitCommands/Git/Executable.cs
index 30d6ae33f..9b3a84bdc 100644
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -118,8 +118,16 @@ private sealed class ProcessWrapper : IProcess
                 }
             }
 
+            private static int count = 0;
+
             private void OnProcessExit(object sender, EventArgs eventArgs)
             {
+                if (count > 50)
+                {
+                    throw new DivideByZeroException("BOOM!");
+                }
+
+                count++;
                 lock (_syncRoot)
                 {
                     // The Exited event can be raised after the process is disposed, however
```

```diff
diff --git a/ResourceManager/GitExtensionsControlInitialiser.cs b/ResourceManager/GitExtensionsControlInitialiser.cs
index 782aa28b6..834979733 100644
--- a/ResourceManager/GitExtensionsControlInitialiser.cs
+++ b/ResourceManager/GitExtensionsControlInitialiser.cs
@@ -15,14 +15,14 @@ internal sealed class GitExtensionsControlInitialiser
 
         public GitExtensionsControlInitialiser(GitExtensionsFormBase form)
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            ////ThreadHelper.ThrowIfNotOnUIThread();
             form.Load += LoadHandler;
             _translate = form;
         }
 
         public GitExtensionsControlInitialiser(GitExtensionsControl control)
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            ////ThreadHelper.ThrowIfNotOnUIThread();
             control.Load += LoadHandler;
             _translate = control;
         }
```
With some chance it will show the bug report form, but if you click on [Quit] the app will indefinitely lock up. The worker thread that will show the form will deadlock the main thread which will be waiting on the worker thread...

Other times the bug report form will never show up due to a race condition between multiple bug reports wanting to show a task dialog:
image


I toyed with an idea of serialising reports, allowing only one bug report dialog at a time, and POSTing other reports to the message queue. But ran into an issue where if a worker thread was aborted it would bring the app down before a user could finish reporting an error...

ResourceManager/GitExtensionsControlInitialiser.cs
```diff
@@ -15,7 +15,11 @@ internal sealed class GitExtensionsControlInitialiser
...
+            {
+                ThreadHelper.ThrowIfNotOnUIThread();
+            }
```
RussKie, line 21: As a quick fix - may be wrap it like this? 
```suggestion
#if DEBUG
            if (form.GetType().Name != "BugReportForm")
            {
                ThreadHelper.ThrowIfNotOnUIThread();
            }
#endif
```

mstv:
> I played with your proposal locally using the suggestion to throw an exception from ProcessWrapper.OnProcessExit, and... it doesn't quite work.

... because the test is too heavy. You are right: If things go wrong, it is likely that other errors will follow.
Though the background operations shall catch exceptions (and already do in the majority). For the case we missed it somewhere, we want a bug report.

If multiple uncaught exceptions are thrown from background threads, a variety of thread / task constellations can arise: dead-locks, multitudes of worker threads without calling the event handler, ... This cannot be handled properly in the dying process.
The safest approach would be to forward the info to a new process via command line.
Alternatively, we could just skip the BugReportForm and open the browser right after the TaskDialog (although it works well when I fire only one exception).

> I toyed with an idea of serialising reports, allowing only one bug report dialog at a time, and POSTing other reports to the message queue. But ran into an issue where if a worker thread was aborted it would bring the app down before a user could finish reporting an error...

Yes, all event handlers must block, else the process exits. Usually, this blocks the UI thread, too.
The first exception is most important. I would not invest much to get the others.

ResourceManager/GitExtensionsControlInitialiser.cs
mstv: > As a quick fix - may be wrap it like this?
> `#if DEBUG`

Why should we omit this in the Release build? What would it quick-fix? Did you mean it without the `if`?
This would replace one code smell with another (slightly less bad).

RussKie:
> The safest approach would be to forward the info to a new process via command line

I kind of arrived to the same conclusion...

ResourceManager/GitExtensionsControlInitialiser.cs
RussKie: Yes, you're right. I am not sure what I was thinking...

mstv:
Superseded by #9063

PR #9049, opened 2021-04-04
Fix tab control flickering in dark theme

NikolayXHD:
## Fixes
tab control flickering introduced by #8815

[screenshot](https://github.com/gitextensions/gitextensions/pull/8815#issuecomment-776006743)
## Proposed change

as discussed https://github.com/gitextensions/gitextensions/pull/8841#issuecomment-777010472

1. Revert #8815
2. Fix high DPI issues in the code which renders TabControl manually.

## Screenshots

image

## Test methodology

manually

## Test environment(s) 

- Windows 10
- UI scale 150%

----


GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
```diff
@@ -0,0 +1,309 @@
...
+            _graphics.FillRectangle(canvasBrush, _clipRectangle);
+
+            RenderSelectedPageBackground();
+
+            IEnumerable<int> pageIndices;
+            if (_selectedIndex.IsWithin(0, _tabCount))
+            {
+                // render tabs in pyramid order with selected on top
+                pageIndices = Enumerable.Range(0, _selectedIndex)
+                    .Concat(Enumerable.Range(_selectedIndex, _tabCount - _selectedIndex).Reverse());
```
RussKie, line 92: Linq is known to have perf-related overhead. Paint operations we'd want to be as lean and mean as possible.
Perhaps worth considering switching this to a plain `for` or `foreach` loop.

GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+
+        private void RenderTabBackground(int index)
+        {
+            using var borderPen = CreateBorderPen();
+            var outerRect = GetOuterTabRect(index);
+            _graphics.FillRectangle(GetBackgroundBrush(index), outerRect);
+
+            var points = new List<Point>(4);
```
RussKie, line 113: `stackalloc Point[4]`?

mdonatas:
The flickering is gone 👍 but something felt wrong so I tried an older build.

This PR:
image

Some build from 3.3 days I had at hand:
image

In 3.3 days the top border would get the color of the active tab.. I think that's the reason something seemed of with tabs in this PR.
Alternatively in 3.3 there might've been no borders.. since their color is same as surrounding background it's hard to tell.

gerhardol (commented):
+1
Run fine with 150%, 200% (no restart though). I do not really see the problem also without this change though.
I normally only use 100% myself, so I do not encounter these issues in normal usage.
I tried to look at border values etc for a few situations, looked fine to me.

GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
NikolayXHD: I cannot imagine loop -> linq overhead per 1 iteration to be to so huge to become comparable with painting ~ 20 x 100 px rectangle of background + icon + text.

GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
NikolayXHD: > stackalloc Point[4]

can't do, the actual size is 2-4 depending on `index`

> `List<Point> points = new(4);`

ok

NikolayXHD:
> The flickering is gone  but something felt wrong so I tried an older build.
> 
> This PR:
> image
> 
> Some build from 3.3 days I had at hand:
> image
 
changed hot tracked tab color

mdonatas:
bottom-border-color

Watch the bottom border of the tab button, it changes color on/after mouse over so the buttons that had been mouse-overed have a different bottom border color than those that haven't.

image
See that the one that hasn't been hovered has a white bottom border.

NikolayXHD:
> bottom-border-color
> 
> Watch the bottom border of the tab button, it changes color on/after mouse over so the buttons that had been mouse-overed have a different bottom border color than those that haven't.
> 
> image
> See that the one that hasn't been hovered has a white bottom border.

@mdonatas Looks to me the visual artifact happens with "Use system default visual style" flag enabled.
I can tell it from the shape of scrollbar arrow, which makes appearance in the gif for a brief moment.

With that flag it's the system drawing the tab control, our (my) code has nothing to do with this.
https://github.com/gitextensions/gitextensions/blob/dc4734a6211dba31dbeb1881f16f27515a225061/GitExtUtils/GitUI/Theming/ThemeFix.cs#L23

mdonatas:
What controls that flag?

In any case, I have three builds
Older 3.3 (I can also see that the arrow is different here)
image

Flickery master
image

This PR
image

And only in this PR is there an underline in places where it should not be... so maybe it's not your code but it's because of the changes made in this PR.

One other thing, `Console` tab is the only that doesn't change color when activated... weird
image

mdonatas:
Hmm... ok, so there is something related to this `Console` tab as it the only one which gets different active color and paints the area beneath it with it. If I don't touch it then there are no artifacts.

Also same here with 175% DPI, couldn't reproduce this underline.

mdonatas:
Uuuu.... thanks for showing me this `Use system default visual style` setting! GE looks better with it!
Look at the tabs! Console tab has correct color and no underline.
image

A blue hover color is back!

And this area in the toolbars looks way better!
This looked like a mess
image

This looks a little bit less like a mess :D
image

NikolayXHD:
> Would it be wrong to argue that unless a user wants black theme color we should have this setting on by default?

Right, and it matches what current behavior does. 

When user selects "default" theme from the list, "Use system defined visual style" checkbox is disabled, controls are rendered by operating system. If OS is dark, GE is dark and vice versa.

With non-default theme "Use system defined visual style" is unchecked and becomes editable. This is because we need to render colors different from current OS theme.

mdonatas:
> 
> 
> Right, and it matches what current behavior does.
> 

Oh, sorry, I didn't notice that.

Can you reproduce `Console` tab being of a different color when active than any other tab? I think that might be a clue for why there is an underline when it's activated.

gerhardol:
I suggest this is merged as is both in master and 3.5 and further discussions are handled separately.
The remaining visual glitches are minor (if I have to look this hard for them they I personally do not care at all, but others are welcome to have other opinions).
Open a new issue instead.

NikolayXHD:
I plan to look at this tomorrow. If I can't fix it tomorrow, i'll do as @gerhardol suggests, merge this and separate the remnant to another issue.

RussKie:
:+1:

NikolayXHD:
> Can you reproduce Console tab being of a different color when active than any other tab? I think that might be a clue for why there is an underline when it's activated.

@mdonatas what you said turns right, it was enough to make console tab render the same background color, to get rid of the artifact.

RussKie (approved):
Thank you 

Squash and merge

mdonatas:
Great! Thanks @NikolayXHD

PR #9028, opened 2021-03-21
Straighten graph lanes

mstv:
Fixes #5782

## Proposed changes

- Move lanes to the right if a branch would go to one lane left for one row
- Use array instead of `List<RevisionGraphRevision>` in `RevisionGraph`

### Before

grafik

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build b737b5330fc78f10310d5d4aec83f94db23a1e52
- Git 2.27.0.windows.1 (recommended: 2.30.0 or later)
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #8498, opened 2020-09-30
Graph column width is too narrow


KarolisMil:
## Current behaviour

If repository is very big, sometimes graph view column is too narrow to fit whole view. It is difficult to see child/parent commits visually, diff-view commits etc.

## Expected behaviour

It should be able to change graph column width freely to use as much screen space as needed.

## Steps to reproduce

To reproduce bug exactly, you need to have many branches in parallel at some point in history.

## Screenshots

![image](https://user-images.githubusercontent.com/13799557/94681729-95b45580-032c-11eb-9b40-5947057256ac.png)

## Did this work in previous version of GitExtensions

This issue is present for at least 1 year now in spite of software versions.

## Environment

- Git Extensions 3.4.3.9999
- Build d4b0f48bbf3e39a71f5d7ff5231be5678d4aa0f1
- Git 2.22.0.windows.1 (recommended: 2.25.1 or later)
- Microsoft Windows NT 10.0.18362.0
- .NET Framework 4.8.4220.0
- DPI 96dpi (no scaling)

Treneris2:
+1

KarolisMil:
Any news?

mstv:
The const `RevisionGraph.MaxLanes` could be turned into an `AppSetting`.

RussKie:
Closing as stale

PR #9050, opened 2021-04-04
Straighten graph lanes over multiple rows

### LINKED ISSUE #5782: Readability revision graph
Discussion/feature

The revision graph is very curly. I think the readability could be improved easily. I propose to make the change as visible in this screenprint (left is new):
image

POC: #5783

- Git Extensions 3.00.rc2
- d6496b811db20f8890cb3a50caa14937f3e0b650  (Dirty)
- Git 2.19.1.windows.1
- .NET Framework 4.7.3110.0
- DPI X:100,00% Y:100,00%

mstv:
Fixes #5782
Follow-up to / generalization of #9028

## Proposed changes

- Straighten lanes over multiple rows
- Fix-up missing handling of `RevisionGraphRow._revisionLane` and `RevisionGraph.GetCachedCount`

## Screenshots 

Before|After
-|-
grafik|grafik

Remark: The part of the red branch starting at `cff2c15` ending at the line with `03d75aa` could be straightened, too. Though this is not detected by the current implementation because this part consists of multiple segments. 

## Test methodology 

- manual with GE and Linux repos
- added NUnit test case

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7835202d0aefbcfea65cad1ad884d6c410b978ec
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -15,8 +15,8 @@ namespace GitUI.UserControls.RevisionGrid.Graph
     // The RevisionGraph contains all the basic structures needed to render the graph.
     public class RevisionGraph : IRevisionGraphRowProvider
     {
-        public const int MaxLanes = 40;
-        private const int _straightenLanesLookAhead = 1;
+        internal const int MaxLanes = 40;
+        private const int _straightenLanesLookAhead = 20;
 
         // Some unordered collections with raw data
         private ConcurrentDictionary<ObjectId, RevisionGraphRevision> _nodeByObjectId = new ConcurrentDictionary<ObjectId, RevisionGraphRevision>();
@@ -72,7 +72,8 @@ namespace GitUI.UserControls.RevisionGrid.Graph
                 return 0;
             }
 
-            return _orderedRowCache.Count;
+            int cachedCount = _orderedRowCache.Count;
+            return cachedCount == Count ? cachedCount : cachedCount - _straightenLanesLookAhead;
         }
 
         /// <summary>
@@ -89,6 +90,8 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         public void CacheTo(int currentRowIndex, int lastToCacheRowIndex)
         {
             currentRowIndex += _straightenLanesLookAhead;
+            lastToCacheRowIndex += _straightenLanesLookAhead;
+
             RevisionGraphRevision[] orderedNodesCache = BuildOrderedNodesCache(currentRowIndex);
 
             BuildOrderedRowCache(orderedNodesCache, currentRowIndex, lastToCacheRowIndex);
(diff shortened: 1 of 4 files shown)


GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -15,8 +15,8 @@ internal interface IRevisionGraphRowProvider
     // The RevisionGraph contains all the basic structures needed to render the graph.
     public class RevisionGraph : IRevisionGraphRowProvider
     {
-        public const int MaxLanes = 40;
-        private const int _straightenLanesLookAhead = 1;
+        internal const int MaxLanes = 40;
+        private const int _straightenLanesLookAhead = 20;
```
mstv, line 19: These two constants could be turned into settings, which might solve #8498, too.

gerhardol: While I personally like options, I believe GE has too many options for most users so I will by default vote no to new options.
If this can be described well it could be added.

ivangrek:
How do you like the idea to mirror the graph vertically?
There will be one line on the right.
On the left, hide, scroll or expand the graph in a separate panel.

mstv:
> How do you like the idea to mirror the graph vertically?

Pro: The main branch would be located directly beside the commit subject.
Con: Would be very nervous when the number of graph lanes changes and the graph width is adapted automatically.
Con: I read from left to right - also the graph.

> There will be one line on the right.

I do not see the benefit of showing only one line of the graph.

> On the left, hide, scroll or expand the graph in a separate panel.

The whole graph can be hidden using the `View` menu.
Instead of another panel, the column could be made resizable. I am happy with the dynamic resizing.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
gerhardol: > Instead of another panel, the column could be made resizable. I am happy with the dynamic resizing.

This could be the configuration for MaxLanes

RussKie: Can we detect automatically how many lanes required?

RussKie (approved):
:shipit:

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
mstv: > Can we detect automatically how many lanes required?

For this, it would not suffice to load the revisions: The graph segments must be created, which is an expensive operation. That's why it is done in chunks of at most 1500 rows on user request.

`MaxLanes` is a hard-coded limit. (I have not introduced it. I moved it here from `RevisionGraphColumnProvider` in the predecessor PR.)

RussKie:
This looks awkward:
image

PR #9056, opened 2021-04-08
Throw on non-zero exit code and on stderror output

### LINKED ISSUE #6527: Generic handling of Git exit codes, errors and warnings

### LINKED ISSUE #8263: [NBug] Der Prozess kann nicht auf die Datei "C:\Projects\Tropico6\...

### LINKED ISSUE #8632: Show error message when Reset file fails

mstv:
Fixes #6527
Fixes #8263
Fixes #8632
Closes #8906 as obsolete
Closes #8961 as obsolete

## Proposed changes

- Throw `ExternalOperationException` if a started process exits with a non-0 exit code or if the process printed output to `StandardError`
- Suppress this exception for commands whose errors are handled

## Test methodology 

- manual
- existing NUnit tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build 3bed1c81024391ba2f58c2b005f915d6fac21260
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -33,7 +33,8 @@ namespace GitCommands
                               bool redirectInput = false,
                               bool redirectOutput = false,
                               Encoding? outputEncoding = null,
-                              bool useShellExecute = false)
+                              bool useShellExecute = false,
+                              bool throwOnErrorOutput = true)
         {
             // TODO should we set these on the child process only?
             EnvironmentConfiguration.SetEnvironmentVariables();
@@ -42,7 +43,7 @@ namespace GitCommands
 
             var fileName = _fileNameProvider();
 
-            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute);
+            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute, throwOnErrorOutput);
         }
 
         #region ProcessWrapper
@@ -62,6 +63,10 @@ namespace GitCommands
             private readonly ProcessOperation _logOperation;
             private readonly bool _redirectInput;
             private readonly bool _redirectOutput;
+            private readonly bool _throwOnErrorOutput;
+
+            private MemoryStream? _emptyStream;
+            private StreamReader? _emptyReader;
 
             private bool _disposed;
 
@@ -72,11 +77,19 @@ namespace GitCommands
                                   bool redirectInput,
                                   bool redirectOutput,
                                   Encoding? outputEncoding,
-                                  bool useShellExecute)
+                                  bool useShellExecute,
+                                  bool throwOnErrorOutput)
             {
                 Debug.Assert(redirectOutput == (outputEncoding is not null), "redirectOutput == (outputEncoding is not null)");
                 _redirectInput = redirectInput;
                 _redirectOutput = redirectOutput;
+                _throwOnErrorOutput = throwOnErrorOutput;
+
+                Encoding errorEncoding = outputEncoding;
+                if (throwOnErrorOutput)
+                {
+                    errorEncoding ??= Encoding.Default;
+                }
 
                 _process = new Process
                 {
@@ -89,9 +102,9 @@ namespace GitCommands
                         CreateNoWindow = !createWindow,
                         RedirectStandardInput = redirectInput,
                         RedirectStandardOutput = redirectOutput,
-                        RedirectStandardError = redirectOutput,
+                        RedirectStandardError = redirectOutput || throwOnErrorOutput,
                         StandardOutputEncoding = outputEncoding,
-                        StandardErrorEncoding = outputEncoding,
+                        StandardErrorEncoding = errorEncoding,
                         FileName = fileName,
                         Arguments = arguments,
                         WorkingDirectory = workDir
@@ -134,8 +147,25 @@ namespace GitCommands
                     {
                         try
                         {
-                            var exitCode = _process.ExitCode;
+                            int exitCode = _process.ExitCode;
                             _logOperation.LogProcessEnd(exitCode);
+
+                            if (_throwOnErrorOutput)
+                            {
+                                string errorOutput = _process.StandardError.ReadToEnd().Trim();
+                                if (exitCode != 0 || errorOutput.Length > 0)
+                                {
+                                    ExternalOperationException ex
+                                        = new(command: _process.StartInfo.FileName,
+                                              _process.StartInfo.Arguments,
+                                              _process.StartInfo.WorkingDirectory,
+                                              exitCode,
+                                              new Exception(errorOutput));
+                                    _logOperation.LogProcessEnd(ex);
+                                    _exitTaskCompletionSource.TrySetException(ex);
+                                }
+                            }
+
                             _exitTaskCompletionSource.TrySetResult(exitCode);
                         }
                         catch (Exception ex)
@@ -180,12 +210,19 @@ namespace GitCommands
             {
                 get
                 {
-                    if (!_redirectOutput)
+                    if (!_redirectOutput && !_throwOnErrorOutput)
                     {
                         throw new InvalidOperationException("Process was not created with redirected output.");
                     }
 
-                    return _process.StandardError;
+                    if (!_throwOnErrorOutput)
+                    {
+                        return _process.StandardError;
+                    }
+
+                    _emptyStream ??= new();
+                    _emptyReader ??= new(_emptyStream);
+                    return _emptyReader;
                 }
             }
 
@@ -221,6 +258,9 @@ namespace GitCommands
                 _process.Dispose();
 
                 _logOperation.NotifyDisposed();
+
+                _emptyReader?.Dispose();
+                _emptyStream?.Dispose();
             }
         }
 
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -340,9 +340,9 @@ namespace GitCommands
                     if (_gitCommonDirectory is null)
                     {
                         GitArgumentBuilder args = new("rev-parse") { "--git-common-dir" };
-                        var result = _gitExecutable.Execute(args);
+                        ExecutionResult result = _gitExecutable.Execute(args, throwOnErrorOutput: false);
 
-                        var dir = result.StandardOutput.Trim().ToNativePath();
+                        string dir = result.StandardOutput.Trim().ToNativePath();
 
                         if (!result.ExitedSuccessfully || dir == ".git" || dir == "." || !Directory.Exists(dir))
                         {
@@ -385,9 +385,9 @@ namespace GitCommands
                 "status",
                 submodulePath
             };
-            var result = _gitExecutable.Execute(args);
+            ExecutionResult result = _gitExecutable.Execute(args, throwOnErrorOutput: false);
 
-            return result.ExitCode == 0 || IsSubmoduleRemoved();
+            return result.ExitedSuccessfully || IsSubmoduleRemoved();
 
             bool IsSubmoduleRemoved()
                 => result.StandardOutput.StartsWith("No submodule mapping found in .gitmodules for path");
@@ -694,7 +694,7 @@ namespace GitCommands
             };
 
             var unmerged = (await _gitExecutable
-                .GetOutputAsync(args)
+                .GetOutputAsync(args, throwOnErrorOutput: false) // ignore non-zero exit code, e.g. in case of missing submodule
                 .ConfigureAwait(false))
                 .Split(Delimiters.NullAndLineFeed, StringSplitOptions.RemoveEmptyEntries);
 
@@ -765,7 +765,7 @@ namespace GitCommands
             return _gitTreeParser.ParseSingle(output);
         }
 
-        public int? GetCommitCount(string parent, string child, bool cache = false)
+        public int? GetCommitCount(string parent, string child, bool cache = false, bool throwOnErrorOutput = true)
         {
             if (parent == child)
             {
@@ -778,7 +778,7 @@ namespace GitCommands
                 $"^{child}",
                 "--count"
             };
-            var output = _gitExecutable.GetOutput(args, cache: cache ? GitCommandCache : null);
+            string output = _gitExecutable.GetOutput(args, cache: cache ? GitCommandCache : null, throwOnErrorOutput: throwOnErrorOutput);
 
             if (int.TryParse(output, out var commitCount))
             {
@@ -880,7 +880,7 @@ namespace GitCommands
                 { !string.IsNullOrWhiteSpace(fileName), "--" },
                 fileName.ToPosixPath().QuoteNE()
             };
-            using var process = _gitExecutable.Start(args, createWindow: true);
+            using var process = _gitExecutable.Start(args, createWindow: true, throwOnErrorOutput: true);
             process.WaitForExit();
         }
 
@@ -1049,9 +1049,9 @@ namespace GitCommands
         public ObjectId? GetCurrentCheckout()
         {
             GitArgumentBuilder args = new("rev-parse") { "HEAD" };
-            var result = _gitExecutable.Execute(args);
+            ExecutionResult result = _gitExecutable.Execute(args, throwOnErrorOutput: false);
 
-            return result.ExitCode == 0 && ObjectId.TryParse(result.StandardOutput, offset: 0, out var objectId)
+            return result.ExitedSuccessfully && ObjectId.TryParse(result.StandardOutput, offset: 0, out var objectId)
                 ? objectId
                 : null;
         }
@@ -1070,7 +1070,7 @@ namespace GitCommands
                 "--quiet",
                 $"{objectIdPrefix}^{{commit}}"
             };
-            var output = _gitExecutable.GetOutput(args).Trim();
+            string output = _gitExecutable.GetOutput(args, throwOnErrorOutput: false).Trim();
 
             if (output.StartsWith(objectIdPrefix) && ObjectId.TryParse(output, out objectId))
             {
@@ -1127,7 +1127,8 @@ namespace GitCommands
                 $"{startRev}..{endRev}"
             };
 
-            return _gitExecutable.GetOutputLines(args).Any();
+            // Could fail if pulling interactively from remote where the specified branch does not exist
+            return _gitExecutable.GetOutputLines(args, throwOnErrorOutput: false).Any();
         }
 
         public ConfigFile GetSubmoduleConfigFile()
@@ -1599,7 +1600,7 @@ namespace GitCommands
                 return "";
             }
 
-            var execution = _gitExecutable.Execute(
+            ExecutionResult execution = _gitExecutable.Execute(
                 new GitArgumentBuilder("update-index")
                 {
                     { assumeUnchanged ? "--assume-unchanged" : "--no-assume-unchanged" },
@@ -1612,7 +1613,8 @@ namespace GitCommands
                         UpdateIndex(inputWriter, file.Name);
                     }
                 },
-                SystemEncoding);
+                SystemEncoding,
+                throwOnErrorOutput: false);
 
             wereErrors = !execution.ExitedSuccessfully;
             return execution.AllOutput;
@@ -1660,7 +1662,7 @@ namespace GitCommands
 
             if (nonDeletedFiles.Count != 0)
             {
-                var execution = _gitExecutable.Execute(
+                ExecutionResult execution = _gitExecutable.Execute(
                     UpdateIndexCmd(AppSettings.ShowErrorsWhenStagingFiles),
                     inputWriter =>
                     {
@@ -1669,7 +1671,8 @@ namespace GitCommands
                             UpdateIndex(inputWriter, file.Name);
                         }
                     },
-                    SystemEncoding);
+                    SystemEncoding,
+                    throwOnErrorOutput: false);
 
                 wereErrors |= !execution.ExitedSuccessfully;
                 output.AppendLine(execution.AllOutput);
@@ -1677,7 +1680,7 @@ namespace GitCommands
 
             if (deletedFiles.Count != 0)
             {
-                var execution = _gitExecutable.Execute(
+                ExecutionResult execution = _gitExecutable.Execute(
                     new GitArgumentBuilder("update-index")
                     {
                         "--remove",
@@ -1690,7 +1693,8 @@ namespace GitCommands
                             UpdateIndex(inputWriter, file.Name);
                         }
                     },
-                    SystemEncoding);
+                    SystemEncoding,
+                    throwOnErrorOutput: false);
 
                 wereErrors |= !execution.ExitedSuccessfully;
                 output.Append(execution.AllOutput);
@@ -2136,7 +2140,7 @@ namespace GitCommands
 
         public async Task<IReadOnlyList<Remote>> GetRemotesAsync()
         {
-            ExecutionResult result = await _gitExecutable.ExecuteAsync("remote -v");
+            ExecutionResult result = await _gitExecutable.ExecuteAsync("remote -v", throwOnErrorOutput: false);
             ////TODO: Handle non-empty result.StandardError if not result.ExitedSuccessfully
             return result.ExitedSuccessfully
                 ? ParseRemotes(result)
@@ -2289,7 +2293,7 @@ namespace GitCommands
                 diffOptions
             };
 
-            var cache = cacheResult &&
+            CommandCache? cache = cacheResult &&
                         !string.IsNullOrEmpty(secondRevision) &&
                         !string.IsNullOrEmpty(firstRevision) &&
                         !secondRevision.IsArtificial() &&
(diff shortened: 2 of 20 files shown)


gerhardol:
There are some more situations where errors are handled, where IsGitErrorMessage() parses the output.
Those should be changed though
(as well as `ExitCode == 0` should be replaced with `ExitedSuccessfull`)

Some situations like "describe" and in RevParse() (that should have --quiet --verify and check status anyway) are handled in this PR.

```patch
diff --git a/GitCommands/Git/GitModule.cs b/GitCommands/Git/GitModule.cs
index 6aace..9124d 100644
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2507,14 +2507,12 @@ public IReadOnlyList<GitItemStatus> GetTreeFiles(ObjectId commitId, bool full)
             bool excludeAssumeUnchangedFiles = true, bool excludeSkipWorktreeFiles = true,
             UntrackedFilesMode untrackedFiles = UntrackedFilesMode.Default)
         {
-            var output = _gitExecutable.GetOutput(
-                GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
-
-            List<GitItemStatus> result = _getAllChangedFilesOutputParser.Parse(output).ToList();
-            if (IsGitErrorMessage(output))
+            var exec = _gitExecutable.Execute(GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
+            List<GitItemStatus> result = _getAllChangedFilesOutputParser.Parse(exec.StandardOutput).ToList();
+            if (!exec.ExitedSuccessfully)
             {
                 // No simple way to pass the error message, create fake file
-                result.Add(createErrorGitItemStatus(output));
+                result.Add(createErrorGitItemStatus(exec.StandardError));
             }
  
             if (!excludeAssumeUnchangedFiles || !excludeSkipWorktreeFiles)
```

GitCommands/Git/Executable.cs
```diff
@@ -179,7 +203,7 @@ public StreamReader StandardError
                         throw new InvalidOperationException("Process was not created with redirected output.");
                     }
 
-                    return _process.StandardError;
+                    return _throwOnErrorOutput ? new StreamReader(new MemoryStream()) : _process.StandardError;
```
RussKie, line 206: Please verify we're not leaking memory here, I can't see where we're disposing the `StreamReader` created here. `_process` gets disposed, and it will dispose its `StandardError` reader, but the new will be left dangling.

GitCommands/Git/Executable.cs
```diff
@@ -128,8 +138,22 @@ private void OnProcessExit(object sender, EventArgs eventArgs)
...
                             _logOperation.LogProcessEnd(exitCode);
+
+                            if (_throwOnErrorOutput)
+                            {
+                                string errorOutput = _process.StandardError.ReadToEnd().Trim();
+                                if (exitCode != 0 || errorOutput.Length > 0)
+                                {
+                                    ExternalOperationException ex
+                                        = new(command: _process.StartInfo.FileName, _process.StartInfo.Arguments, _process.StartInfo.WorkingDirectory,
+                                                new Exception($"exit code {exitCode}{Environment.NewLine}{errorOutput}"));
```
RussKie, line 151:
```suggestion
                                                new Exception($"Exit code: {exitCode}{Environment.NewLine}{errorOutput}"));
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -275,11 +276,20 @@ public void SelectionChanged(IReadOnlyList<GitRevision> selectedRevisions)
...
+                if (selectedGuid is not null)
+                {
+                    try
+                    {
+                        mergedBranches = (await Module.GetMergedBranchesAsync(includeRemote: true, fullRefname: true, commit: selectedGuid)).ToHashSet();
+                    }
+                    catch (ExternalOperationException)
+                    {
+                        // ignore not parsable "HEAD"
+                    }
```
RussKie, line 289:
```suggestion
                HashSet<string> mergedBranches;
                if (selectedGuid is not null)
                {
                    try
                    {
                        mergedBranches = (await Module.GetMergedBranchesAsync(includeRemote: true, fullRefname: true, commit: selectedGuid)).ToHashSet();
                    }
                    catch (ExternalOperationException)
                    {
                        // ignore not parsable "HEAD"
                        mergedBranches = new();
                    }
```

RussKie (commented):
👍 with few comments

GitCommands/Git/Executable.cs
mstv: Done. But can't the GC do it? It is very unlikely that this code path is reached, since the exception already contains the consumed `StandardError` output.

GitCommands/Git/Executable.cs
mstv: I will move the exit code to the `ExternalOperationException`. The user will not be interested very much in the specific value.

GitUI/BranchTreePanel/RepoObjectsTree.cs
mstv: `mergedBranches` shall not be nullable, and must be an empty `HashSet` if `selectedGuid is null`.

GitCommands/Git/GitModule.cs
```diff
@@ -2467,9 +2473,9 @@ public IReadOnlyList<GitItemStatus> GetStashDiffFiles(string stashName)
             // shows untracked files
             var args = new GitArgumentBuilder("log")
             {
-                $"{stashName}^3",
                 "--pretty=format:\"%T\"",
-                "--max-count=1"
+                "--max-count=1",
+                $"-- {stashName}^3"
```
gerhardol, line 2478: -- separates paths, stashNane is a ref
The order is correct now though

GitCommands/Git/GitModule.cs
```diff
@@ -2289,20 +2292,22 @@ public IReadOnlyList<GitStash> GetStashes(bool noLocks = false)
...
+            CommandCache? cache = cacheResult &&
                         !string.IsNullOrEmpty(secondRevision) &&
                         !string.IsNullOrEmpty(firstRevision) &&
                         !secondRevision.IsArtificial() &&
                         !firstRevision.IsArtificial()
                 ? GitCommandCache
                 : null;
 
-            var patch = _gitExecutable.GetOutput(
+            bool nonZeroGitExitCode = firstId == ObjectId.WorkTreeId && secondId is not null && !isTracked;
```
gerhardol, line 2303: Should not be here but at the caller (may check arguments).
!isTracked cannot use git-diff (regardless of first/second commit).

gerhardol (commented):
+0
In general very much needed, good time to add this now.
Will use myself.
(only briely looked at the code now.)

GitCommands/Git/GitModule.cs
mstv: The commit message contains the git error:

```
Fixup git arguments for GetStashDiffFiles

fatal: ambiguous argument 'stash@{0}^3': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions
```

sounds a bit confusing to me; but works well

GitCommands/Git/GitModule.cs
mstv: repro:
- create a new file (untracked)
- select `Working directory`, then `Ctrl`+`click` the checked out commit, which turns the added (exit code 0) into a removed file from this perspective (exit code 1)

mstv:
> I get an exception now when starting up, some dashboard repo does not exist I assume

cannot reproduce

mstv:
> Missing Working directory for exceptions

How did you get this?
Usually the directory is there (renamed to just "Directory":

exception with wdir

GitCommands/Git/GitModule.cs
gerhardol: The caller should not request this information, it is meaningless.

GitCommands/Git/GitModule.cs
gerhardol: stashName is not a path, so it should not be separated with `--`

mstv: Without "--", git outputs the above _fatal_ error with non-zero exit code.

gerhardol:
Failure in repo without tags (similar to git-describe).
```
       public IReadOnlyList<IGitRef> GetRefs(GetRefsEnum getRef)
        {
            // We do not want to lock the repo for background operations.
            // The primary use of 'noLocks' is to run git-status the commit count as a background operation,
            // but to run the same in a foreground for FormCommit.
            //
            // Assume that all GetRefs() are done in the background, which may not be correct in the future.
            const bool noLocks = true;

            ArgumentString cmd = GitCommandHelpers.GetRefsCmd(getRef, noLocks, AppSettings.RefsSortBy, AppSettings.RefsSortOrder);
            ExecutionResult exec = _gitExecutable.Execute(cmd, throwOnErrorOutput: false);
            return !exec.ExitedSuccessfully
                ? ParseRefs(exec.StandardOutput)
                : new List<IGitRef>();
        }
```

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
```diff
@@ -178,7 +178,7 @@ bool TestGitCommand(string command)
             {
                 try
                 {
-                    string output = new Executable(command).GetOutput();
+                    string output = new Executable(command).GetOutput(throwOnErrorOutput: false);
```
gerhardol, line 181: Avoid commands that fail when they do not have to
```suggestion
                    string output = new Executable(command).GetOutput(arguments: " --version");
```

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
```diff
@@ -237,7 +237,7 @@ public static bool CheckIfFileIsInPath(string fileName)
 
         public bool CanFindGitCmd()
         {
-            return !string.IsNullOrEmpty(Module?.GitExecutable.GetOutput(""));
+            return !string.IsNullOrEmpty(Module?.GitExecutable.GetOutput(arguments: string.Empty, throwOnErrorOutput: false));
```
gerhardol, line 240:
```suggestion
            return !string.IsNullOrEmpty(Module?.GitExecutable.GetOutput(arguments: "--version"));
```

gerhardol:
I have experienced a few more situations where Git exists with non-zero.
Pushed to a private branch, I believe you can pick them from there
(if you want a motivation let me know, I logged with stack trace while working today and have not investigated the details).

mstv:
> I have experienced a few more situations where Git exists with non-zero.
Pushed to a private branch, I believe you can pick them from there

Thank you. I have cherry-picked them to my work branch `all`. I will look at the details later.

mstv:
> (if you want a motivation let me know, I logged with stack trace while working today and have not investigated the details)

Yes, please, especially the exception text taken from `StandardError`. I find warnings or errors from these commands quite surprising.

gerhardol:
One more push with a fix for git-diff with core.safecrlf

gerhardol:
I still have an exception at startup. Just pushed a commit to be able to work.
 
The sidepanel is init without the working directory is set. That should be disallowed for basically all commands.
Have not found what causes this.

image

mstv:
> The sidepanel is init without the working directory is set. That should be disallowed for basically all commands.

Good catch! This might also be the cause why some of the ROT tests work unstable (and have been disabled).

mstv:
> > (if you want a motivation let me know, I logged with stack trace while working today and have not investigated the details)
> 
> Yes, please, especially the exception text taken from `StandardError`. I find warnings or errors from these commands quite surprising.

I could not reproduce these exceptions. Could you provide details for the first four `!fixup` commits, please?

gerhardol:
Added two commits where error is OK (this was the remaining fatal: reference I saw too)
Maybe all GetOutputLines() should be ignored

The commit messages has descriptions

gerhardol:
42e27c794fd4658502b303fd2c1f592b0cd20682

Line patches exit with error and error text if patches cannot be applied
To use Execute() instead of GetOutput() the input array need to be made to a stream.

gerhardol:
git-mergetool can exit with error if user do not complete the merge
(pushed as well but my private branch is getting out of date)

```patch
diff --git a/GitCommands/Git/GitModule.cs b/GitCommands/Git/GitModule.cs
index e6ca8..6f2dd 100644
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -880,7 +880,7 @@ public void RunMergeTool(string? fileName = "", string? customTool = null)
                 { !string.IsNullOrWhiteSpace(fileName), "--" },
                 fileName.ToPosixPath().QuoteNE()
             };
-            using var process = _gitExecutable.Start(args, createWindow: true);
+            using var process = _gitExecutable.Start(args, createWindow: true, throwOnErrorOutput: false);
             process.WaitForExit();
         }
 
```

RussKie (commented):
👍

GitCommands/Git/GitModule.cs
```diff
@@ -880,7 +880,7 @@ public void RunMergeTool(string? fileName = "", string? customTool = null)
                 { !string.IsNullOrWhiteSpace(fileName), "--" },
                 fileName.ToPosixPath().QuoteNE()
             };
-            using var process = _gitExecutable.Start(args, createWindow: true);
+            using var process = _gitExecutable.Start(args, createWindow: true, throwOnErrorOutput: false);
```
mstv, line 883: In my tests, I could not see any errors from git.
How to reproduce, @gerhardol?

gerhardol: You need to set https://git-scm.com/docs/git-mergetool#Documentation/git-mergetool.txt-mergetoollttoolgttrustExitCode
The tool need to set the exit code too. At least Beyond Compare does that, not checked others.

Then you need to make a merge with a conflict and not resolve the conflict, but that part I am sure you follow.

mstv: Thank you for the explanation.

gerhardol (commented):
+1
I have reviewed before and used this PR for quite some time.
I doubt I will have many comments when reviewing again.
Just as I detected some commands that required an adjustment (as mstv had done before of course for many uses), other users will see new edge cases. This will depend on Git versions to some extent too.

So I recommend merge.

gerhardol:
One more exclusion for throwing in #9413 
`git-log --follow` accepts exactly one argument only. Manual path filters may contain more arguments.
There are basic checks for more than one argument, but the final check is the Git command.

mstv:
> git-log --follow accepts exactly one argument only. Manual path filters may contain more arguments.
There are basic checks for more than one argument, but the final check is the Git command.

Are these errors displayed?

gerhardol:
> Are these errors displayed?

Not seen, handled by ignoring the output.
Not a problem before #9413 beacause rename detection is not implemented for Advanced filter and FileHistory always have only one argument.

GitCommands/Git/GitModule.cs
mstv: Since `RunMergeTool` does not check the results, I have reverted this from this PR.
If errors must really be ignored, I'd like to handle it in a dedicated PRs on principle.

I have not seen any popups when aborting the merge conflict resolution in BC - with `git config mergetool.bc.trustExitCode true`.

mstv:
@msftbot merge in 5 days

gerhardol (approved):
Lets get this merged, there will be adjustments needed...
(Both due to incorrect GE code and to allow error exits)

GitCommands/Git/GitModule.cs
RussKie: We need to have at least a comment explaining this. It is not obvious what this check is and why it is needed.

GitCommands/Git/Executable.cs
```diff
@@ -33,7 +33,8 @@ public Executable(Func<string> fileNameProvider, string workingDir = "")
                               bool redirectInput = false,
                               bool redirectOutput = false,
                               Encoding? outputEncoding = null,
-                              bool useShellExecute = false)
+                              bool useShellExecute = false,
+                              bool throwOnErrorOutput = true)
```
RussKie, line 37: How many callsites with `throwOnErrorOutput = true` vs `throwOnErrorOutput = false` do we have?

GitCommands/Git/Executable.cs
```diff
@@ -134,8 +147,25 @@ private void OnProcessExit(object sender, EventArgs eventArgs)
...
+                                    ExternalOperationException ex
+                                        = new(command: _process.StartInfo.FileName,
+                                              _process.StartInfo.Arguments,
+                                              _process.StartInfo.WorkingDirectory,
+                                              exitCode,
+                                              new Exception(errorOutput));
+                                    _logOperation.LogProcessEnd(ex);
+                                    _exitTaskCompletionSource.TrySetException(ex);
+                                }
+                            }
```
RussKie, line 167: Do we have tests for this?

RussKie:

RussKie:
I have few questions, disabling the auto-merge for now.

GitCommands/Git/Executable.cs
gerhardol: 42 (and 18 tests) that do not throw. Some of those could probably be tightened to not return errors.
Other about 69 (and 22 plugin and 24 tests)

mstv: This is what this PR is about: Let the user see the errors *by default*.
I want to have a justification for all exceptions from this default.

GitCommands/Git/GitModule.cs
mstv: > Should not be here but at the caller (may check arguments).
!isTracked cannot use git-diff (regardless of first/second commit).

> The caller should not request this information, it is meaningless.

Explain it in other words, please.

> It is not obvious what this check is and why it is needed.

Would you better understand it with `nonZeroGitExitCodeExpected`?
It is used once - in the next statement.

GitCommands/Git/Executable.cs
mstv: This is an integration. What would be the focal point of this test?

RussKie (approved):

GitCommands/Git/Executable.cs
RussKie: Now looking at it, I don't think it is correct.

I want to fail on "exit code != 0", but I don't want to fail on "errorOutput.Length > 0" because it contains warnings.
If anything it has to be 
```cs
if (exitCode != 0 || (_throwOnErrorOutput && errorOutput.Length > 0))
```

As far as a test goes, we have integration tests in `ExecutableExtensionsTests` that allow testing this functionality.

gerhardol: The intention with this PR is that errors and warnings should be handled, so checking the error output is in line with this.
In #9468 the throw will suppress both exit code and stderr, only show popup for exit code - could be popup on filtered stderr too.)
The origin of #9468 was that Git added printout to stderr in situations where the user did not expect them or see them before, so this should maybe be removed before a public release.
I suggest open an issue in milestone 3.6 for this and we catch warnings until then.

gerhardol: Throwing on only stderr could be a separate argument too.

RussKie: By fixing one thing we must not break existing working things.

And I maintain this implementation is not correct - "exit code=0" implies the operation completed successfully. `errorOutput` may contain warnings, but they are not fatal, otherwise they'd been made errors.

gerhardol: > By fixing one thing we must not break existing working things.

This PR **will** break functionality that happened to work before. See my conversation in this PR and PRs pointing to this as examples. (Of course, mstv fixed most problems in the PR but there is no history of that other than the code changes).

I find this breaking change important to find issues (at least during development), some to be suppressed, some to be suppressed, others to be fixed.
For production the popup could be suppressed (I currently want to have it enabled though.)

> And I maintain this implementation is not correct - "exit code=0" implies the operation completed successfully. errorOutput may contain warnings, but they are not fatal, otherwise they'd been made errors.

The warnings on stderr could be fatal, it is case by case. This could be yet another configuration.
But letting the caller make the choice is likely sufficient, similar to all other commands.

`throwOnErrorOutput` should not be allowed with `GetOutput*`, stderr should never be parsed as stdout.
Also, `includeErrorOutput` should be removed as these commands should fail.
This affects like 7 methods

`GetOutputLinesAsync` should be removed too (9 usages), will not be usable in #9476 anyway.
Not sure how the async yielding should be handled though.

gerhardol: Draft cleanup in 3d320d33b8987b522b12dc88ea492ebc7ed84e80
The only actual change is that stderr is not parsed as an output (except SystemEncodingReader where I was unsure).

`GetOutput` is the simple flow where neither error exit nor stderr is allowed. 
"Advanced" usage with `Execute`

mstv: > this implementation is not correct

No, it is. But the definition seems to be too strict.

> "exit code=0" implies the operation completed successfully. errorOutput may contain warnings, but they are not fatal, otherwise they'd been made errors.

I agree, this is a good definition.

Anyway, I want to get noticed of git warnings and non-fatal errors. They do not need to popup interrupting the command execution. The `ErrorOutput` should be added to the Git Command Log. New warnings should be indicated e.g. in the toolbar of FormBrowse or might be shown as Windows notifications. (Might need a configurable filter.)

PS: Regarding my status: It means tomorrow.

gerhardol: `GetOutput()` as a simple method could well fail on any error output. stderr should normally not be parsed in the expected output.
I reviewed the use of stderr/AllOuput in #9488.
If `GetOutput` is changed to not throw if error exit is set, the method should only return stderr, not merge it with stderr.
`GetOutput` could be replaced globally too.

PR #9058, opened 2021-04-10
AutoComplete: Avoid Git for non tracked files

gerhardol:
## Proposed changes

Requirement for #9056 

AutoComplete in FormCommit tries to list all files that are in worktree/index and parse them, to find names of functions etc to use in the autocompletion.
(Extensions and regexp in AutoCompleteRegexes.txt)
First git-diff is tried (OK, but empty), but git-show will raise a git error status. Better to not call Git here than to handle the error in #9056 

### Before

image

### After

No Git calls

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2725,10 +2725,20 @@ namespace GitCommands
                 : null;
         }
 
+        /// <summary>
+        /// Get the file contents for the HEAD commit
+        /// </summary>
+        /// <param name="file">The Git status item</param>
+        /// <returns>An awaitable task with the requested file contents.</returns>
         public async Task<string?> GetFileContentsAsync(GitItemStatus file)
         {
-            var contents = new StringBuilder();
+            if (!file.IsTracked || (file.Staged == StagedStatus.Index && file.IsNew))
+            {
+                // File is not in Git
+                return null;
+            }
 
+            StringBuilder contents = new();
             string? currentContents = await GetFileContentsAsync(file.Name).ConfigureAwaitRunInline();
             if (currentContents is not null)
             {
--- a/GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
+++ b/GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
@@ -149,19 +149,21 @@ namespace GitUI.AutoCompletion
 
         private static async Task<string?> GetChangedFileTextAsync(GitModule module, GitItemStatus file)
         {
-            var changes = await module.GetCurrentChangesAsync(file.Name, file.OldName, file.Staged == StagedStatus.Index, "-U1000000")
+            if (file.IsTracked)
+            {
+                var changes = await module.GetCurrentChangesAsync(file.Name, file.OldName, file.Staged == StagedStatus.Index, "-U1000000")
                 .ConfigureAwait(false);
 
-            if (changes is not null)
-            {
-                return changes.Text;
-            }
+                if (changes is not null)
+                {
+                    return changes.Text;
+                }
 
-            var content = await module.GetFileContentsAsync(file).ConfigureAwaitRunInline();
-
-            if (content is not null)
-            {
-                return content;
+                var content = await module.GetFileContentsAsync(file).ConfigureAwaitRunInline();
+                if (content is not null)
+                {
+                    return content;
+                }
             }
 
             // Try to read the contents of the file: if it cannot be read, skip the operation silently.


mstv (approved):
Thank you!

GitCommands/Git/GitModule.cs
```diff
@@ -2725,9 +2725,19 @@ public bool IsDirtyDir()
                 : null;
         }
 
+        /// <summary>
+        /// Get the file contents in HEAD for the file
+        /// </summary>
+        /// <param name="file">The Git status item</param>
+        /// <returns>async Task with contents</returns>
```
RussKie, line 2732:
```suggestion
        /// <returns>An awaitable task with the requested file content.</returns>
```

GitCommands/Git/GitModule.cs
```diff
@@ -2725,9 +2725,19 @@ public bool IsDirtyDir()
                 : null;
         }
 
+        /// <summary>
+        /// Get the file contents in HEAD for the file
```
RussKie, line 2729: I'm struggling to parse the meaning
```suggestion
        /// Get the file content.
```

GitCommands/Git/GitModule.cs
```diff
@@ -2725,9 +2725,19 @@ public bool IsDirtyDir()
...
+        /// </summary>
+        /// <param name="file">The Git status item</param>
+        /// <returns>async Task with contents</returns>
         public async Task<string?> GetFileContentsAsync(GitItemStatus file)
```
RussKie, line 2733: This should be:
```suggestion
        public async Task<string?> GetFileContentAsync(GitItemStatus file)
```

GitCommands/Git/GitModule.cs
```diff
@@ -2725,9 +2725,19 @@ public bool IsDirtyDir()
...
         {
             var contents = new StringBuilder();
```
RussKie, line 2735:
```suggestion
            var content = new StringBuilder();
```

RussKie (commented):
👍

GitCommands/Git/GitModule.cs
gerhardol: Get the file contents for the HEAD commit
```suggestion
        /// Get the file contents for the HEAD commit
```

GitCommands/Git/GitModule.cs
gerhardol: I suggest to keep the s
https://english.stackexchange.com/questions/56831/file-content-vs-file-contents

GitCommands/Git/GitModule.cs
gerhardol: As discussed below (currently) changed to
```suggestion
        /// <returns>An awaitable task with the requested file contents.</returns>
```

GitCommands/Git/GitModule.cs
```diff
@@ -2725,10 +2725,20 @@ public bool IsDirtyDir()
...
         public async Task<string?> GetFileContentsAsync(GitItemStatus file)
         {
-            var contents = new StringBuilder();
+            if (!file.IsTracked || (file.Staged == StagedStatus.Index && file.IsNew))
+            {
+                // File is not in Git
+                return null;
+            }
 
+            var contents = new StringBuilder();
```
mstv, line 2741:
```suggestion
            StringBuilder contents = new();
```

gerhardol: We could change this globally too
(this was just a moved line, I do normally not updated unrelated changes to keep diffs reviewable.)
Updated, no other changes than this and the squash

gerhardol:
@msftbot merge in 24 hours

GitCommands/Git/GitModule.cs
```diff
@@ -2725,10 +2725,20 @@ public bool IsDirtyDir()
...
+        /// </summary>
+        /// <param name="file">The Git status item</param>
+        /// <returns>An awaitable task with the requested file contents.</returns>
         public async Task<string?> GetFileContentsAsync(GitItemStatus file)
         {
-            var contents = new StringBuilder();
+            if (!file.IsTracked || (file.Staged == StagedStatus.Index && file.IsNew))
+            {
+                // File is not in Git
+                return null;
```
ivangrek, line 2738: Task is null there.

gerhardol: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/async-return-types

The Task is async, not the same as not async.
The task is not null, it returns a null string.

Issue #8914, opened 2021-03-03
GE hangs when viewing submodules in diff viewer


gerhardol:
## Current behaviour

Regression in master in the last days. 

If a submodule is selected in worktree (at least - seem OK for HEAD though), GE hangs and stops responding. Status is never changed from the default. Mouse just spinning.

## Expected behaviour

Submodule is displayed.

## Steps to reproduce

Select modified submodule in worktree in the diff viewer.

## Screenshots

![image](https://user-images.githubusercontent.com/6248932/109844938-e79b3780-7c4c-11eb-9443-1ddee6a2fa6d.png)

## Did this work in previous version of GitExtensions

- Git Extensions 3.6.0.11114
- Build f54a727a95393dcacfbbd8603dfd6daf931618c4 (Dirty)

## Environment

- Git Extensions 3.6.0.11170
- Build 273a0f6fd3e07858f837cdc19d50827871e32319 (Dirty)
- Git 2.30.0.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

gerhardol:
Not in release/3.5

Introduced in 7f7a6f2be079f123d5fdbe10c41489ddc7b54ea7
https://github.com/gitextensions/gitextensions/commit/7f7a6f2be079f123d5fdbe10c41489ddc7b54ea7#diff-c9456496ac47efd017c5bf895c05418cbe001c5a471a5bcd33104d6f9a42d513
```c#
diff --git a/GitUI/GitUIExtensions.cs b/GitUI/GitUIExtensions.cs
index 5359d6..195a1a 100644
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -112,10 +112,11 @@ void OpenWithDiffTool()
                         : diffOfConflict;
                 }
 
-                if (file.IsSubmodule)
+                if (file.IsSubmodule
+                    && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
                 {
                     // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync);
+                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
                     return status is not null
                         ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
[...]

PR #9059, opened 2021-04-10
Use exit status instead of IsGitErrorMessage()

gerhardol:
Preparation for #9056 

## Proposed changes

ExecutableExtensions: CommandCache for Execute()
Previously caching was possible only for GetOutput()

Instead of parsing the combined standard/error output for 'fatal' or 'error:' in IsGitErrorMessage(), use the executable exit status.
The standard usage is to add a a fake GitStatusItem to show the failed command for the user in FileStatusList.
This handling is done _after_ standard output is parsed, as it is currently. 

Note that changes like GitConfigurationException makes it hard to reach these codepaths (you get constant popups, cannot do anything) so testing this is hard.

## Test methodology 

Manual, some commands are tested though

----

### DIFF
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -294,9 +294,10 @@ namespace GitCommands
             ArgumentString arguments = default,
             Action<StreamWriter>? writeInput = null,
             Encoding? outputEncoding = null,
+            CommandCache? cache = null,
             bool stripAnsiEscapeCodes = true)
         {
-            var result = await executable.ExecuteAsync(arguments, writeInput, outputEncoding, stripAnsiEscapeCodes);
+            var result = await executable.ExecuteAsync(arguments, writeInput, outputEncoding, cache, stripAnsiEscapeCodes);
             return result.StandardOutput.SplitLines().Concat(result.StandardError.SplitLines());
         }
 
@@ -320,10 +321,11 @@ namespace GitCommands
             ArgumentString arguments,
             Action<StreamWriter>? writeInput = null,
             Encoding? outputEncoding = null,
+            CommandCache? cache = null,
             bool stripAnsiEscapeCodes = true)
         {
             return GitUI.ThreadHelper.JoinableTaskFactory.Run(
-                () => executable.ExecuteAsync(arguments, writeInput, outputEncoding, stripAnsiEscapeCodes));
+                () => executable.ExecuteAsync(arguments, writeInput, outputEncoding, cache, stripAnsiEscapeCodes));
         }
 
         /// <summary>
@@ -340,9 +342,18 @@ namespace GitCommands
             ArgumentString arguments,
             Action<StreamWriter>? writeInput = null,
             Encoding? outputEncoding = null,
+            CommandCache? cache = null,
             bool stripAnsiEscapeCodes = true)
         {
             outputEncoding ??= _defaultOutputEncoding.Value;
+
+            if (cache is not null && cache.TryGet(arguments, out var cachedOutput, out var cachedError))
+            {
+                return new ExecutionResult(
+                    CleanString(stripAnsiEscapeCodes, EncodingHelper.DecodeString(cachedOutput, error: null, ref outputEncoding)),
+                    CleanString(stripAnsiEscapeCodes, EncodingHelper.DecodeString(output: null, cachedError, ref outputEncoding)),
+                    exitCode: 0);
+            }
 
             using var process = executable.Start(arguments, createWindow: false, redirectInput: writeInput is not null, redirectOutput: true, outputEncoding);
             var outputBuffer = new MemoryStream();
@@ -363,8 +374,12 @@ namespace GitCommands
 
             var output = outputEncoding.GetString(outputBuffer.GetBuffer(), 0, (int)outputBuffer.Length);
             var error = outputEncoding.GetString(errorBuffer.GetBuffer(), 0, (int)errorBuffer.Length);
-
             var exitCode = await process.WaitForExitAsync();
+
+            if (cache is not null && exitCode == 0)
+            {
+                cache.Add(arguments, outputBuffer.ToArray(), errorBuffer.ToArray());
+            }
 
             return new ExecutionResult(
                 CleanString(stripAnsiEscapeCodes, output),
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2363,11 +2363,11 @@ namespace GitCommands
             });
         }
 
-        public string GetDiffFiles(string? firstRevision, string? secondRevision, bool noCache = false, bool nullSeparated = false)
+        public ExecutionResult GetDiffFiles(string? firstRevision, string? secondRevision, bool noCache = false, bool nullSeparated = false)
         {
             noCache = noCache || firstRevision.IsArtificial() || secondRevision.IsArtificial();
 
-            return _gitExecutable.GetOutput(
+            return _gitExecutable.Execute(
                 new GitArgumentBuilder("diff")
                 {
                     "--no-color",
@@ -2427,8 +2427,12 @@ namespace GitCommands
                 return status.Where(x => x.Staged == stagedStatus).ToList();
             }
 
-            var output = GetDiffFiles(firstRevision, secondRevision, noCache: noCache, nullSeparated: true);
-            var result = GetDiffChangedFilesFromString(output, stagedStatus);
+            ExecutionResult exec = GetDiffFiles(firstRevision, secondRevision, noCache: noCache, nullSeparated: true);
+            var result = GetDiffChangedFilesFromString(exec.StandardOutput, stagedStatus);
+            if (!exec.ExitedSuccessfully)
+            {
+                result.Add(createErrorGitItemStatus(exec.StandardError));
+            }
 
             if (firstRevision == GitRevision.WorkTreeGuid || secondRevision == GitRevision.WorkTreeGuid)
             {
@@ -2510,14 +2514,12 @@ namespace GitCommands
             bool excludeAssumeUnchangedFiles = true, bool excludeSkipWorktreeFiles = true,
             UntrackedFilesMode untrackedFiles = UntrackedFilesMode.Default)
         {
-            var output = _gitExecutable.GetOutput(
-                GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
-
-            List<GitItemStatus> result = _getAllChangedFilesOutputParser.Parse(output).ToList();
-            if (IsGitErrorMessage(output))
+            ExecutionResult exec = _gitExecutable.Execute(GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
+            List<GitItemStatus> result = _getAllChangedFilesOutputParser.Parse(exec.StandardOutput).ToList();
+            if (!exec.ExitedSuccessfully)
             {
                 // No simple way to pass the error message, create fake file
-                result.Add(createErrorGitItemStatus(output));
+                result.Add(createErrorGitItemStatus(exec.StandardError));
             }
 
             if (!excludeAssumeUnchangedFiles || !excludeSkipWorktreeFiles)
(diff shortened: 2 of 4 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -2510,14 +2510,12 @@ public IReadOnlyList<GitItemStatus> GetTreeFiles(ObjectId commitId, bool full)
             bool excludeAssumeUnchangedFiles = true, bool excludeSkipWorktreeFiles = true,
             UntrackedFilesMode untrackedFiles = UntrackedFilesMode.Default)
         {
-            var output = _gitExecutable.GetOutput(
-                GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
-
-            List<GitItemStatus> result = _getAllChangedFilesOutputParser.Parse(output).ToList();
-            if (IsGitErrorMessage(output))
+            var exec = _gitExecutable.Execute(GitCommandHelpers.GetAllChangedFilesCmd(excludeIgnoredFiles, untrackedFiles));
```
RussKie, line 2513: Please use type instead of var

RussKie (commented):
LGTM
How did you test these?

gerhardol:
> LGTM
> How did you test these?

Some commands have tests (but not really related to errors).

Manual usage, adding errors (bad config like when IsGitError was added), for some commands. (GitConfigurationException makes this much harder to use, you do not get much functionality.)
The handling is unchanged, some processing also if there are errors - this could be changed but are harder to review.
It may be that we should remove some of the handling with #9056 too, then GE is unusable if there is any unexpected Git error.

RussKie:
👍

GitCommands/Git/GitModule.cs

gerhardol:
Updated with: ExecutableExtensions: CommandCache for Execute()

RussKie (approved):

gerhardol:
@mstv This affects #9056, I suggest this is merged first

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -363,9 +374,13 @@ string ComposeOutput()
 
             var output = outputEncoding.GetString(outputBuffer.GetBuffer(), 0, (int)outputBuffer.Length);
             var error = outputEncoding.GetString(errorBuffer.GetBuffer(), 0, (int)errorBuffer.Length);
-
             var exitCode = await process.WaitForExitAsync();
 
+            if (cache is not null && await exitTask == 0)
+            {
+                cache.Add(arguments, outputBuffer.ToArray(), errorBuffer.ToArray());
```
mstv, line 381: We must be careful not to cache too much.
I guess that I have seen diffs with Index and WD not updated correctly until I reloaded everything in the past month.

GitCommands/Git/GitModule.cs
```diff
@@ -3032,21 +3022,20 @@ public IReadOnlyList<string> GetAllBranchesWhichContainGivenCommit(ObjectId obje
...
                     objectId
-                });
-
-            if (IsGitErrorMessage(output))
+                };
+            ExecutionResult exec = _gitExecutable.Execute(args);
+            if (!exec.ExitedSuccessfully)
             {
                 return Array.Empty<string>();
             }
```
mstv, line 3036: What do we do with the error message in this case?
After #9056, we should throw an `ExternalOperationException` here, which must be caught by the caller and be displayed in the UI element.

GitCommands/Git/GitModule.cs
```diff
@@ -3074,14 +3063,13 @@ public IReadOnlyList<string> GetAllBranchesWhichContainGivenCommit(ObjectId obje
...
         public IReadOnlyList<string> GetAllTagsWhichContainGivenCommit(ObjectId objectId)
         {
-            var output = _gitExecutable.GetOutput($"tag --contains {objectId}");
-
-            if (IsGitErrorMessage(output))
+            ExecutionResult exec = _gitExecutable.Execute($"tag --contains {objectId}");
+            if (!exec.ExitedSuccessfully)
             {
                 return Array.Empty<string>();
             }
```
mstv, line 3070: ditto

GitCommands/Git/GitModule.cs
```diff
@@ -3115,11 +3103,12 @@ public IReadOnlyList<string> GetAllTagsWhichContainGivenCommit(ObjectId objectId
              *     Format path dialog improved
              */
 
-            if (IsGitErrorMessage(output))
+            if (!exec.ExitedSuccessfully)
             {
                 return null;
             }
```
mstv, line 3109: ditto

GitCommands/Git/GitModule.cs
```diff
@@ -4146,24 +4135,11 @@ public bool StopTrackingFile(string filename)
...
-        /// </summary>
-        /// <param name="gitOutput">The output from the git command, to inspect.</param>
-        /// <returns><c>true</c> if the command detailed an error, otherwise <c>false</c>.</returns>
-        public static bool IsGitErrorMessage(string gitOutput)
-        {
-            return Regex.IsMatch(gitOutput, @"^\s*(error:|fatal)");
+            ExecutionResult exec = _gitExecutable.Execute(args);
+            return !exec.ExitedSuccessfully
+                ? exec.StandardOutput.TrimEnd()
+                : null;
```
mstv, line 4142: ditto

mstv (commented):
Looks good, Just a few nits.
How ´do you trigger the git errors?

GitCommands/Git/ExecutableExtensions.cs
gerhardol: Sidenote to this PR:
If something is cached, a refresh will not update automatically, the cache contents is kept.
So what you saw is likely that the file lists or the contents of a diff is not updated automatically at changes.
I would like to have notification about changes in the diff tab and FormCommit similar to the Browse refresh (FileStatusMonitor) to detect this.

However, I want a cache for git-status too, but that is a new functionality, the cache will need quite heavy invalidation policy...

gerhardol:
> How ´do you trigger the git errors?

You cannot easily trigger most of them.
See comment https://github.com/gitextensions/gitextensions/pull/9059#issuecomment-817218537

This PR is not really trying to improve the handling, it is a preparation for #9056 where commands currently handling Git errors but the handling was not easily visible and was not handled in the draft #9056.
 
The review comments about error messages not displayed are like that, the messages were not displayed before, at some point it was not fatal, compare to git-rev-parse where an error is normal. Maybe these situations should be reconsidered. That was not my goal for this PR though.

gerhardol:
Rebased (on #9058) and added a fixup with an explanation where the error message should be presented (removing it would have been a regression).
See below, the "fatal:" info is useful information.
(This is something I see quite often when working with submodules and worktrees).

```
Submodule Externals/EasyHook Change

From:	5183aa53a286771f10b5828053e9474a30dc4db1

To:	0296f95158a7da669b44aaa264d8793c478ef648
					1 year ago (2020-01-16 11:56:13)
		Remove targets, replaced by super's targets/props

Type: Unknown

Differences:

fatal: bad object 5183aa53a286771f10b5828053e9474a30dc4db1
```

I need https://github.com/gitextensions/gitextensions/issues/8914#issuecomment-790080907
to run this also on my home PC

GitCommands/Git/GitModule.cs
gerhardol: Exit > 0 is normal if no tag exists where the command was started, must not raise errors

GitCommands/Git/GitModule.cs
gerhardol: adding comment

GitCommands/Git/GitModule.cs
gerhardol: adding comment

GitCommands/Git/GitModule.cs
gerhardol: adding comment

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -340,10 +342,19 @@ string ComposeOutput()
             ArgumentString arguments,
             Action<StreamWriter>? writeInput = null,
             Encoding? outputEncoding = null,
+            CommandCache? cache = null,
             bool stripAnsiEscapeCodes = true)
         {
             outputEncoding ??= _defaultOutputEncoding.Value;
 
+            if (cache is not null && cache.TryGet(arguments, out var cacheOutput, out var cacheError))
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -340,10 +342,19 @@ string ComposeOutput()
...
+            {
+                return new ExecutionResult(
+                    CleanString(stripAnsiEscapeCodes, EncodingHelper.DecodeString(cacheOutput, null, ref outputEncoding)),
+                    CleanString(stripAnsiEscapeCodes, EncodingHelper.DecodeString(null, cacheError, ref outputEncoding)),
+                    0);
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -363,9 +374,13 @@ string ComposeOutput()
 
             var output = outputEncoding.GetString(outputBuffer.GetBuffer(), 0, (int)outputBuffer.Length);
             var error = outputEncoding.GetString(errorBuffer.GetBuffer(), 0, (int)errorBuffer.Length);
-
             var exitCode = await process.WaitForExitAsync();
 
+            if (cache is not null && await exitTask == 0)
```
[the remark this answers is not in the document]
gerhardol: ok

mstv (approved):
Thank you.

GitCommands/Git/GitModule.cs
```diff
@@ -2725,10 +2725,20 @@ public bool IsDirtyDir()
...
+        /// </summary>
+        /// <param name="file">The Git status item</param>
+        /// <returns>An awaitable task with the requested file contents.</returns>
         public async Task<string?> GetFileContentsAsync(GitItemStatus file)
         {
-            var contents = new StringBuilder();
+            if (!file.IsTracked || (file.Staged == StagedStatus.Index && file.IsNew))
+            {
+                // File is not in Git
+                return null;
```
ivangrek, line 2728: return Task.FromResult(null)

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
```diff
@@ -149,19 +149,21 @@ private static IEnumerable<string> ReadOrInitializeAutoCompleteRegexes()
...
+                if (changes is not null)
+                {
+                    return changes.Text;
+                }
 
-            if (content is not null)
-            {
-                return content;
+                var content = await module.GetFileContentsAsync(file).ConfigureAwaitRunInline();
+                if (content is not null)
```
ivangrek, line 163: Why?

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -376,7 +376,7 @@ string ComposeOutput()
             var error = outputEncoding.GetString(errorBuffer.GetBuffer(), 0, (int)errorBuffer.Length);
             var exitCode = await process.WaitForExitAsync();
```
ivangrek, line 377: I think we need to use `await exitTask` there. 
To prevent second run.

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
mstv: This is an early exit. Else it is tried another way below. So why "Why?"?

GitCommands/Git/ExecutableExtensions.cs
mstv: Done in line 373: `await Task.WhenAll(outputTask, errorTask, exitTask);`.

GitUI/AutoCompletion/CommitAutoCompleteProvider.cs
ivangrek: Ok.

GitCommands/Git/ExecutableExtensions.cs
ivangrek: image

ivangrek: Also.
https://www.meziantou.net/process-waitforexitasync-doesn-t-behave-like-process-waitforexit.htm

GitCommands/Git/GitModule.cs
gerhardol: The task is not null, it returns a null string. What is the point of writing that with more text?

ivangrek: If task is not null it is ok.
I'll take a closer look. 😄

ivangrek: I don't know why we are using `ConfigureAwaitRunInline()` there. But we could try to get the content in parallel for two files.

GitCommands/Git/ExecutableExtensions.cs
mstv: > I think we need to use `await exitTask` there.
To prevent second run.

Thank you, I got it now. (I think in this case, it will not do harm and will not execute very much twice. Though beware of early signs, i.e. it should be changed.)

The .NET 5 issue does not affect this code because the output tasks are awaited explicitly.

GitCommands/Git/GitModule.cs
gerhardol: This is in #9058, I based this PR on #9058 to prepare the merge of this, so discussed there.

GitCommands/Git/ExecutableExtensions.cs
gerhardol: This was not changed in this PR, hesitant to change...

GitCommands/Git/GitModule.cs
RussKie: My first reaction was too to suggest `Task.FromResult(null)`. But since there's an `await` below, this is not necessary, and distorts the intent.

RussKie:
@gerhardol merge at will

gerhardol:
Rebased and squashed with an indentation change in GitModule.
Merging despite two partly open discussions, not really this PR:
* Task null return that is part of #9058 (mstv tried to stop the bot merge so I did not try that myself, but it failed...). RussKie resolved the conversation.
* Execute() await handling. Existing code

gerhardol:
@msftbot merge in 12 hours

mstv:
> mstv tried to stop the bot merge

I stopped it because of the new review comment - by removing the `auto-merge` label, which worked.
After you had rebuted it, I re-added the label and #9058 was merged as planned.

PR #9063, opened 2021-04-11
Report bugs in a separate process

RussKie:
## Proposed changes

- If app is getting terminated due to an exception on a background thread launch the bug reporter app as a separate process as mentioned in https://github.com/gitextensions/gitextensions/pull/9048#issuecomment-814942209

- If an unhandled exception is throw on the main thread, show the error within the app process. The error may or may not be fatal.
This code path is for the most part unchanged from the existing behaviour.

- If an unhandled exception is thrown from a background thread it will crash the app. Serialize and encode the error information as base64 string, and launch the bug reporter app before terminating the app.

- The bug report exception handling is updated to use SerializableException type to accommodate both -in-app and out-of-proc exception reporting.

## Test methodology 

- manual
- unit tests
- integration tests

----


RussKie:
Outstanding:
- `SerializableException` doesn't correctly serialize nested exceptions, thus losing information
- Correct "copy exception" functionality
- Tests

mstv:
The `BugReporter.exe` itself crashes and starts another instance. 🤣

grafik

I have only debugged GE, not the BugReporter yet. I'll continue later.
I am still a little busy, a.o. struggling with MSVS 16.9.3 crashing partially.

mstv:
Found it: The BugReporter uses `Executable` to call `git --version` etc.
Works now with the dummy exception thrown on a later git call.

mstv:
I think we should not use the `BugReporter.exe` for non-background exceptions - also not in the future.
Terminating exceptions *are* bugs regardless of their type. Hence we do not need different arguments, just the serialized exception.

BugReporter/UserEnvironmentInformation.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using GitCommands;
+using GitExtUtils;
+using GitExtUtils.GitUI;
+
+namespace BugReporter
+{
+    internal static class UserEnvironmentInformation
```
mstv, line 10: Do we still need the `GitUI\UserEnvironmentInformation.cs`?

IntegrationTests/UI.IntegrationTests/NBugReports/BugReportFormTests.cs
```diff
@@ -9,26 +9,26 @@ namespace GitExtensions.UITests.NBugReports
     [TestFixture]
     public class BugReportFormTests
     {
-        private BugReportForm _form;
+        ////private BugReportForm _form;
```
mstv, line 12: What shall happen with these tests?

UnitTests/GitUI.Tests/NBugReports/BugReporterTests.cs
```diff
@@ -16,7 +15,7 @@ public sealed class BugReporterTests
         public void Append(Exception exception, string expectedRootError, string expectedText)
         {
             StringBuilder text = new();
-            string rootError = BugReporter.Append(text, exception);
+            string rootError = GitExtensions.BugReporter.Append(text, exception);
```
mstv, line 18: Is this necessary?

mstv (commented):
👍 Thank you.

The linebreaks in the stack trace get modified by the XML serialization(?). 
grafik
The bug report and the internal display for normal exceptions are not affected. So it is good enough IMO.

RussKie:
> The `BugReporter.exe` itself crashes and starts another instance. 🤣
> 
> grafik

🤔  I thought I solved it... I'll need to double check on my personal workstation that I have pushed all changes. 


> I think we should not use the `BugReporter.exe` for non-background exceptions - also not in the future.
> Terminating exceptions _are_ bugs regardless of their type. Hence we do not need different arguments, just the serialized exception.

Yes, that's the intent - if it isn't fatal - we will show the form, else we'll launch BugReporter.exe and terminate the app.

IntegrationTests/UI.IntegrationTests/NBugReports/BugReportFormTests.cs
RussKie: Once the whole thing is finished I'll see whether these tests are still relevant, and, if they are, where they belong.

RussKie:
> The linebreaks in the stack trace get modified by the XML serialization(?).

Good catch! I think `\r` may get stripped somewhere along. Will check.

BugReporter/UserEnvironmentInformation.cs
RussKie: Yes, because it populates information for other purposes too (e.g. the About dialog).
I wanted to reuse it across the projects, but it didn't quite work.

UnitTests/GitUI.Tests/NBugReports/BugReporterTests.cs
RussKie: Yes, due to namespace clash. Reworking

BugReporter/Serialization/SerializableException.cs
```diff
@@ -177,7 +177,15 @@ public static SerializableException FromXmlString(string xml)
...
             var serializer = new XmlSerializer(typeof(SerializableException));
             using StringReader reader = new(xml);
-            return (SerializableException)serializer.Deserialize(reader);
+            SerializableException exception = (SerializableException)serializer.Deserialize(reader);
+
+            if (exception.StackTrace is not null && exception.StackTrace.IndexOf("\r\n") < 0)
+            {
+                // Presume that the payload was serialized with \n only
+                exception.StackTrace = exception.StackTrace.Replace("\n", Environment.NewLine);
+            }
```
mstv, line 186: Nit: I think `Environment.NewLine` should be used everywhere. Should we shorten the condition?
```suggestion
            if (exception.StackTrace?.IndexOf(Environment.NewLine) < 0)
            {
                // Presume that the payload was serialized with \n only
                exception.StackTrace = exception.StackTrace.Replace("\n", Environment.NewLine);
            }
```

GitUI/NBugReports/BugReporter.cs
```diff
@@ -1,17 +1,15 @@
...
 using GitExtUtils;
-using GitUI;
-using GitUI.NBugReports;
 using Microsoft.WindowsAPICodePack.Dialogs;
 
-namespace GitExtensions
+namespace GitUI.NBugReports
 {
-    public static class BugReporter
+    public static class BugReportInvoker
```
mstv, line 12: The file should be renamed, too.

UnitTests/GitUI.Tests/NBugReports/BugReporterTests.cs
```diff
@@ -2,21 +2,20 @@
...
-using GitExtensions;
 using GitExtUtils;
 using GitUI.NBugReports;
 using NUnit.Framework;
 
 namespace GitUITests.NBugReports
 {
     [TestFixture]
-    public sealed class BugReporterTests
+    public sealed class BugReportInvokerTests
```
mstv, line 12: The file should be renamed, too.

mstv (commented):
Works very well.
Open: The nits below and the `BugReportFormTests`, which is rather a unit test.

BugReporter/Serialization/SerializableException.cs
RussKie: Yes, thank you.

Btw, https://github.com/dotnet/runtime/issues/40315 would have made this a non-issue

RussKie:
I think it is ready now.

mstv (approved):
Thank you

PR #9064, opened 2021-04-11
Do not reset a valid `AppSettings.GitBinDir`

mstv:
## Proposed changes

- Do not reset a valid `AppSettings.GitBinDir` if `sh` is found in - GE's extended private - `PATH`
- Really try "usr\bin" *and* "bin" subfolders - in changed sequence (as recommended by @vbjay)
- Fixup false friend "gitbindir" in setup of `ExecutableExtensionsTests.SetUp`

### Before

This AppSetting was reset in the second run, and `Repair` did not work:
 
grafik
grafik

### After

grafik

## Test methodology 

- have git in the `PATH`, namely "C:\Program Files\Git\cmd"
- do not specify the path to the git executable
- set Linux tools path to "C:\Program Files\Git\bin"
- restart twice

## Test environment(s) 

- Git Extensions 33.33.33
- Build 639f83d6ffd9b756a61a7e433284e70221581047
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
@@ -65,15 +65,15 @@ namespace GitUI.CommandsDialogs.SettingsDialog
                 gitpath = possibleNewPath.Trim();
             }
 
-            foreach (string toolsPath in new[] { @"bin\", @"usr\bin\" })
+            foreach (string toolsPath in new[] { @"usr\bin\", @"bin\" })
             {
-                gitpath = gitpath.Replace(@"\cmd\git.exe", @"\" + toolsPath)
+                string linuxToolsPath = gitpath.Replace(@"\cmd\git.exe", @"\" + toolsPath)
                     .Replace(@"\cmd\git.cmd", @"\" + toolsPath)
                     .Replace(@"\bin\git.exe", @"\" + toolsPath);
 
-                if (ContainsSh(gitpath))
+                if (ContainsSh(linuxToolsPath))
                 {
-                    AppSettings.GitBinDir = gitpath;
+                    AppSettings.GitBinDir = linuxToolsPath;
                     return true;
                 }
 
@@ -90,7 +90,7 @@ namespace GitUI.CommandsDialogs.SettingsDialog
 
                 foreach (string path in GetGitLocations())
                 {
-                    gitpath = path + toolsPath;
+                    linuxToolsPath = path + toolsPath;
                     if (ContainsSh(gitpath))
                     {
                         AppSettings.GitBinDir = gitpath;
--- a/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
@@ -79,6 +79,11 @@ namespace GitUI.CommandsDialogs.SettingsDialog
 
                 if (CheckIfFileIsInPath("sh.exe") || CheckIfFileIsInPath("sh"))
                 {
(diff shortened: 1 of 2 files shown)


vbjay:
Linux sh tools isn't that path.  It's the folder that has the linux tools
which has the sh file.

gerhardol (approved):
The behavior matches the check in CheckGitExe()
(have not run)
3.5?

mstv:
> 3.5?

Yes, I will cherry-pick this commit after it has been merged.

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
```diff
@@ -82,6 +82,15 @@ public bool SolveLinuxToolsDir(string? possibleNewPath = null)
 
                 if (CheckIfFileIsInPath("sh.exe") || CheckIfFileIsInPath("sh"))
                 {
+                    string configuredPath = AppSettings.GitBinDir;
+                    if (Directory.Exists(configuredPath))
+                    {
+                        if (File.Exists(configuredPath + "sh.exe") || File.Exists(configuredPath + "sh"))
+                        {
```
RussKie, line 89: Can join and reduce nesting

RussKie (approved):

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
gerhardol: In that case the checks on line 74, 93/102 should be changed too, they all follow the same pattern.
(also include line 83).
Adds noise to the change.

vbjay:
Frankly. I think the check should be recursive. Ask them to point to linux
sh tools iterate over the folders in the directory selected and find the
sh.exe path.  We could hint that they can find it where git is installed.

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
mstv: I have inserted a refactoring commit.

mstv:
> Frankly. I think the check should be recursive. Ask them to point to linux sh tools iterate over the folders in the directory selected and find the sh.exe path. We could hint that they can find it where git is installed.

This is kind of what this function tries to do. But it works only if the full path to the git executable is specified.
You are welcome to improve the logic. With this PR, I just want to fix the bug.

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
```diff
@@ -65,41 +65,43 @@ public bool SolveLinuxToolsDir(string? possibleNewPath = null)
                 gitpath = possibleNewPath.Trim();
             }
 
-            foreach (var toolsPath in new[] { @"bin\", @"usr\bin\" })
+            foreach (string toolsPath in new[] { @"bin\", @"usr\bin\" })
```
gerhardol, line 68: If you rewrite this, this loop could be moved down to where toolspath is used, no need to do the first two checks twice.
(Not requesting changes - I would have preferred the previous version for 3.5 even if this is a nice refactor.)

gerhardol (approved):
(have not run)

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
mstv: `toolsPath` is used for replacing in `gitpath` on the next line.
But there is another bug: The replacing of `gitpath` cannot work in the subsequent loops.

PR #9067, opened 2021-04-13
Fix tab control flickering in dark theme

NikolayXHD:
repeat #9049 in release branch

----


gerhardol (approved):
Same patch as in master (except for two lines due to annotation).
The tabs look OK (but I do not really see the problem normally...)

PR #9072, opened 2021-04-14
GitHub token: Add a link to let the user generate its Personal Access Token

### LINKED ISSUE #8850: Unable to generate GitHub OAuth token using plugin
## Current behaviour

I am attempting to integrate GE with GH and enter my GH username and password. I click "OK" , but see an error when generating a token. Of note, this is the first time I am attempting to authorize GE as an OAuth client.

I don't have 2FA enabled for my GH account. I've tried creating a PAT on GH and entering that as the password, but that doesn't change the error message.

Entering a PAT in the "OAuth Token" section of the GH plugin settings results in a different error when attempting any GitHub plugin commands, possibly because a PAT isn't the right token to pass.

On further debugging, it appears that the API endpoint for authorizations ~has either moved to a different location~ or has been deprecated:

> POST https://api.github.com/authorizations HTTP/1.1
> 
> Authorization: Basic ********
> User-Agent: GitExtensions
> Content-Type: application/json
> Host: api.github.com
> Content-Length: 104
> Expect: 100-continue
> 
> {"scopes":["repo", "public_repo"],"note":"Token for Git Extensions on *** at 2/12/2021 15:07"}
> 
> HTTP/1.1 404 Not Found
> Date: Fri, 12 Feb 2021 23:07:35 GMT
> Content-Type: application/json; charset=utf-8
> Content-Length: 74
> Server: GitHub.com
> X-GitHub-Media-Type: github.v3; format=json
> X-RateLimit-Limit: 60
> X-RateLimit-Remaining: 48
> X-RateLimit-Reset: 1613173499
> x-ratelimit-used: 12
> Access-Control-Expose-Headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset
> Access-Control-Allow-Origin: *
> Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
> X-Frame-Options: deny
> X-Content-Type-Options: nosniff
> X-XSS-Protection: 1; mode=block
> Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin
> Content-Security-Policy: default-src 'none'
> Vary: Accept-Encoding, Accept, X-Requested-With
> X-GitHub-Request-Id: A989:6F61:284CD6:2E9C86:60270A37
> 
> {"message":"Not Found","documentation_url":"https://docs.github.com/rest"}

## Expected behaviour
The GitHub plugin can either create a token successfully or prompts me to authorize GE in browser (similar to #7706, and how other apps I use do it)

## Steps to reproduce

1. Ensure that you don't already have a GitHub OAuth token configured in plugin settings
2. Launch GitHub from Plugins/GitHub
3. Login with your GitHub username and password (and 2FA if applicable)
4. Click on OK to attempt authentication/authorization.

## Screenshots

image

Not sure; this is the first time I'm trying this out

## Environment

- Git Extensions 3.4.3.9999
- Git 2.29.2.windows.2
- .NET Framework 4.8.4250.0
- DPI 120dpi (125% scaling)

## Diagnostics
Did not see any steps on how to collect logs for GitHub plugin.

pmiossec:
instead of using the GitHub api to generate one
because it no longer work since GitHub deprecated this api

Fixes #8850

### Before

See #8850

### After
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build d4eaf22a0fa2ec889694a8d7de8f4d5835edba4f
- .NET Framework 4.8.4341.0
- DPI 192dpi (200% scaling)


pmiossec:
Side notes: 
1. I kept the field named "Oauth token" because otherwise all the users that have a token configured will loose it. But technically it's a "Personnal access token"

image

2. Perhaps we should also add a link toward https://github.com/settings/tokens . What do you think?
3. Once everyone is happy, I think that should be backported to 3.5

pmiossec:
@DamitSe @clayton-travlr if you want to test it and give your feedback?

DamitSe:
Tried using a PAT, it throws an exception in the GH plugin (maybe PAT is not the right kind of token?)
https://github.com/gitextensions/gitextensions/issues/8850#issuecomment-819678693

vbjay:
Pat would work. Just need to give it the right scope.

Plugins/GitHub3/GitHub3Plugin.cs
```diff
@@ -102,6 +104,22 @@ public GitHub3Plugin() : base(true)
...
+            yield return new PseudoSetting(generateTokenLink);
+        }
+
+        private void GenerateTokenLink_Click(object sender, EventArgs e)
+        {
+            try
+            {
+                Process.Start($"https://{GitHubHost.ValueOrDefault(Instance.Settings)}/settings/tokens/new?description=Token%20for%20GitExtensions&scopes=repo,public_repo");
+            }
+            catch (Exception)
```
ivangrek, line 119:
```
catch
{}
```

RussKie: We must not swallow exceptions, especially those that a user can fix.

Plugins/GitHub3/GitHub3Plugin.cs
```diff
@@ -72,12 +72,15 @@ public static string OAuthToken
     public class GitHub3Plugin : GitPluginBase, IRepositoryHostPlugin
     {
         private readonly TranslationString _viewInWebSite = new("View in {0}");
-        private readonly TranslationString _tokenAlreadyExist = new("You already have an OAuth token. To get a new one, delete your old one in Plugins > Settings first.");
+        private readonly TranslationString _tokenAlreadyExist = new("You already have an personal access token. To get a new one, delete your old one in Plugins > Plugin Settings first.");
+        private readonly TranslationString _generateToken = new("Generate a GitHub personal access token");
+        private readonly TranslationString _manageToken = new("Manage GitHub personal access token");
+        private readonly TranslationString _openLinkFailed = new TranslationString("Fail to open the link");
```
RussKie, line 78: Should the English.Plugins.xlf be updated? 🤔

RussKie (commented):
LGTM modulo the comment

RussKie:
Worth taking into 3.5 too.

Plugins/GitHub3/GitHub3Plugin.cs
pmiossec: That surprised me also but I launched to localization command and none of the string were translated. The build doesn't complained also.
So I think that the plugins are not translated....

Plugins/GitHub3/GitHub3Plugin.cs
pmiossec: I don't get your points...
I put a catch here, in case, to avoid a crash if I miss something because I know that `Process.Start()` could throw exception.
But none of these types of exceptions seems to be thrown when opening an url.
They are all related to opening a file or a process.

That's why I did just a basic handling and not something that would add a lot of lines of code for nothing.

I did a small change to display the exception message.
Is it enough or were you thinking about something else?

Plugins/GitHub3/GitHub3Plugin.cs
```diff
@@ -101,7 +104,37 @@ public GitHub3Plugin() : base(true)
...
+        private void ManageTokenLink_Click(object sender, EventArgs e)
+        {
+            OpenLink($"https://{GitHubHost.ValueOrDefault(Instance.Settings)}/settings/tokens");
+        }
+
+        private void OpenLink(string url)
+        {
+            try
+            {
+                Process.Start(url);
```
mstv, line 132: At least for .NET 5, `useShellExecute: true` and `Verb = useShellExecute` are needed.

pmiossec: @mstv I did something. Is that good? Because I don't understand the `Verb = useShellExecute` part and didn't found something about it. Do you have a documentation?

vbjay:
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute

pmiossec:
> 
> https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute

This one is set @vbjay . That's the part I changed. I was asking for the "Verb" property.

vbjay:
Pretty sure verb should be open

pmiossec:
> 
> Pretty sure verb should be open

The only thing I found: https://stackoverflow.com/questions/7693429/process-start-to-open-an-url-getting-an-exception#comment117440865_53245993

Plugins/GitHub3/GitHub3Plugin.cs
mstv: 👍, just add the `Verb` like here:
https://github.com/gitextensions/gitextensions/blob/d488e497cc6244483794abedb8e4779495ba861c/GitCommands/Git/Executable.cs#L83-L101

mstv: Sorry, I have corrected a copy & paste mistake in my initial comment.

pmiossec: @mstv Done. Even if it works without...

mstv (commented):
+1
have not run

pmiossec:
@msftbot merge in 3 days

Plugins/GitHub3/GitHub3Plugin.cs
```diff
@@ -72,12 +72,15 @@ public static string OAuthToken
     public class GitHub3Plugin : GitPluginBase, IRepositoryHostPlugin
     {
         private readonly TranslationString _viewInWebSite = new("View in {0}");
-        private readonly TranslationString _tokenAlreadyExist = new("You already have an OAuth token. To get a new one, delete your old one in Plugins > Settings first.");
+        private readonly TranslationString _tokenAlreadyExist = new("You already have an personal access token. To get a new one, delete your old one in Plugins > Plugin Settings first.");
+        private readonly TranslationString _generateToken = new("Generate a GitHub personal access token");
+        private readonly TranslationString _manageToken = new("Manage GitHub personal access token");
+        private readonly TranslationString _openLinkFailed = new TranslationString("Fail to open the link. Reason: ");
```
gerhardol, line 78: Should we have parameters in texts like these?

Plugins/GitHub3/GitHub3Plugin.cs
```diff
@@ -101,7 +104,43 @@ public GitHub3Plugin() : base(true)
...
+            yield return new PseudoSetting(generateTokenLink);
+
+            var manageTokenLink = new LinkLabel { Text = _manageToken.Text };
+            manageTokenLink.Click += ManageTokenLink_Click;
+            yield return new PseudoSetting(manageTokenLink);
+        }
+
+        private void GenerateTokenLink_Click(object sender, EventArgs e)
+        {
+            OpenLink($"https://{GitHubHost.ValueOrDefault(Instance.Settings)}/settings/tokens/new?description=Token%20for%20GitExtensions&scopes=repo,public_repo");
```
gerhardol, line 120: This should be using OsShellUtil.OpenUrlInDefaultBrowser

pmiossec: It seems not possible because `OsShellUtil`is part of `GitUI` project that is not a dependency to the plugins.

Plugins/GitHub3/GitHub3Plugin.cs
pmiossec: I let you decide. Even if I think about it when writing it, I don't have a strong opinion. I though it was easier for translators (even if these strings are not translated).

gerhardol: Not sure how this is handled in other languages.
Not requesting changes here, this is the common pattern now.

PR #9075, opened 2021-04-14
Some GitHub plugin fixes

pmiossec:
## Proposed changes

-  ViewPullRequestsForm: 
    - better handle error on one remote (form is now still usable when a remote is on error)
    - fix selection of first remote found with at least a pull request
- CreatePullRequestForm: 
    - better handle error on one remote (form is now still usable when a remote is on error)
    - CreatePullRequestForm: increase width of dropdown 

### Before

Unable to do one easily but red arrow in 'after' screenshot show width increase to display well long branch name

### After

image

- ViewPullRequestsForm:

image

- CreatePullRequestForm: 

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 47bb7083c58cafdb7d180c425a88389b116b6f44
- .NET Framework 4.8.4341.0
- DPI 192dpi (200% scaling)

### DIFF
--- a/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
@@ -6,6 +6,7 @@ using System.Windows.Forms;
 using GitCommands;
 using GitUIPluginInterfaces.RepositoryHosts;
 using Microsoft.VisualStudio.Threading;
+using Microsoft.WindowsAPICodePack.Dialogs;
 using ResourceManager;
 
 namespace GitUI.CommandsDialogs.RepoHosting
@@ -19,6 +20,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
         private readonly TranslationString _strFailedToCreatePullRequest = new("Failed to create pull request.");
         private readonly TranslationString _strPleaseCloneGitHubRep = new("Please clone GitHub repository before pull request.");
         private readonly TranslationString _strDone = new("Done");
+        private readonly TranslationString _strRemoteFailToLoadBranches = new("Fail to load target branches");
         #endregion
 
         private readonly IRepositoryHostPlugin _repoHost;
@@ -28,6 +30,7 @@ namespace GitUI.CommandsDialogs.RepoHosting
         private string? _currentBranch;
         private string? _prevTitle;
         private readonly AsyncLoader _remoteLoader = new();
+        private bool _ignoreFirstRemoteLoading = true;
 
         [Obsolete("For VS designer and translation test only. Do not remove.")]
 #pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
@@ -96,11 +99,18 @@ namespace GitUI.CommandsDialogs.RepoHosting
                 _pullReqTargetsCB.SelectedIndex = 0;
             }
 
+            _ignoreFirstRemoteLoading = false;
+
             _pullReqTargetsCB_SelectedIndexChanged(this, EventArgs.Empty);
         }
 
         private void _pullReqTargetsCB_SelectedIndexChanged(object sender, EventArgs e)
         {
+            if (_ignoreFirstRemoteLoading)
+            {
+                return;
+            }
+
             _currentHostedRemote = (IHostedRemote)_pullReqTargetsCB.SelectedItem;
 
             _remoteBranchesCB.Items.Clear();
@@ -160,8 +170,15 @@ namespace GitUI.CommandsDialogs.RepoHosting
 
                             _createBtn.Enabled = true;
                         }
-                        catch (Exception)
+                        catch (Exception ex)
                         {
+                            System.Windows.Forms.TaskDialog.ShowDialog(new TaskDialogPage
+                                {
+                                    Icon = TaskDialogIcon.Error,
+                                    Caption = _strRemoteFailToLoadBranches.Text,
+                                    Text = string.Format(TranslatedStrings.RemoteInError, ex.Message, remote.DisplayData),
+                                    Buttons = { System.Windows.Forms.TaskDialogButton.OK },
+                                });
                         }
                     })
                 .FileAndForget();
--- a/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
+++ b/GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
@@ -132,31 +132,37 @@ namespace GitUI.CommandsDialogs.RepoHosting
                     {
                         await TaskScheduler.Default;
 
-                        IHostedRepository hostedRepository = remote.GetHostedRepository();
-                        var branches = hostedRepository.GetBranches();
-
-                        await this.SwitchToMainThreadAsync();
-
-                        comboBox.Items.Clear();
-
-                        var selectItem = 0;
-                        var defaultBranch = hostedRepository.GetDefaultBranch();
-                        for (var i = 0; i < branches.Count; i++)
+                        try
                         {
-                            if (branches[i].Name == defaultBranch)
+                            IHostedRepository hostedRepository = remote.GetHostedRepository();
+                            var branches = hostedRepository.GetBranches();
+
+                            await this.SwitchToMainThreadAsync();
+
+                            comboBox.Items.Clear();
+
+                            var selectItem = 0;
+                            var defaultBranch = hostedRepository.GetDefaultBranch();
+                            for (var i = 0; i < branches.Count; i++)
                             {
-                                selectItem = i;
+                                if (branches[i].Name == defaultBranch)
+                                {
+                                    selectItem = i;
+                                }
+
+                                comboBox.Items.Add(branches[i].Name);
                             }
 
-                            comboBox.Items.Add(branches[i].Name);
-                        }
+                            if (branches.Count > 0)
+                            {
+                                comboBox.SelectedIndex = selectItem;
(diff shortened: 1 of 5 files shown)


pmiossec:
@DamitSe I don't know exactly the reason of your error describe in comment https://github.com/gitextensions/gitextensions/issues/8850#issuecomment-819678693 but this PR could perhaps fix your issue and let you use the form...

DamitSe:
> @DamitSe I don't know exactly the reason of your error describe in comment [#8850 (comment)](https://github.com/gitextensions/gitextensions/issues/8850#issuecomment-819678693) but this PR could perhaps fix your issue and let you use the form...

I opened a new issue for this #9076; the issue was that I hadn't authorized the PAT but the dialog box didn't tell me that.

GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
```diff
@@ -132,31 +132,37 @@ private void PopulateBranchesComboAndEnableCreateButton(IHostedRemote remote, Co
...
+                                {
+                                    selectItem = i;
+                                }
+
+                                comboBox.Items.Add(branches[i].Name);
                             }
 
-                            comboBox.Items.Add(branches[i].Name);
-                        }
+                            if (branches.Count > 0)
```
ivangrek, line 156: `branches.Any()`

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -73,22 +73,27 @@ private void ViewPullRequestsForm_Load(object sender, EventArgs e)
...
+
+                    // load all hosted repositories.
+                    foreach (var hostedRemote in hostedRemotes)
                     {
-                        el.GetHostedRepository(); // We do this now because we want to do it in the async part.
+                        try
+                        {
+                            hostedRemote.GetHostedRepository(); // We do this now because we want to do it in the async part.
+                        }
+                        catch (Exception)
```
ivangrek, line 85:
```
catch
{}
```

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
```diff
@@ -99,7 +104,17 @@ private void _selectedOwner_SelectedIndexChanged(object sender, EventArgs e)
...
             var hostedRemote = _selectHostedRepoCB.SelectedItem as IHostedRemote;
 
-            var hostedRepo = hostedRemote?.GetHostedRepository();
+            _pullRequestsList.Items.Clear();
+            IHostedRepository? hostedRepo;
+            try
+            {
+                hostedRepo = hostedRemote?.GetHostedRepository();
+            }
+            catch (Exception)
```
ivangrek, line 113:
```
catch
{}
```

GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
drewnoakes: Why? This will be slower.

ivangrek: It is more readable, more flexible, and uses a named method to denote an operation.
But I agree, this is subjective.

ivangrek: Maybe extension method `Any()` for list will be better. 😄

GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
```diff
@@ -132,31 +132,37 @@ private void PopulateBranchesComboAndEnableCreateButton(IHostedRemote remote, Co
...
+                                comboBox.SelectedIndex = selectItem;
+                            }
 
-                        if (branches.Count > 0)
+                            _createBtn.Enabled = true;
+                        }
+                        catch (Exception)
                         {
-                            comboBox.SelectedIndex = selectItem;
                         }
```
RussKie, line 165: This isn't good. @mstv and I have been working to remove these silent catch blocks. 
The error must be bubbled up to the bug reporter, and explain the user what can be done to resolve it, if possible.

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
RussKie: Why?

RussKie, line 85: ditto

GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
RussKie: Having worked on the .NET SDK for the past few years I've learnt to avoid Linq.

drewnoakes: To be clear, my original objection wasn't really about performance. I guess it's what you're used to. I'm perfectly used to seeing a count check. My brain understands immediately, and also doesn't worry about allocations or overhead. If I see `Any` I worry about those things slightly, and worry about multiple enumeration and so forth more. But it's a matter of style.

I find these kinds of style things in PR reviews can provide a micro-barrier to contribution. In general I think projects at this kind of scale should use analyzers to enforce these kinds of preferences to save everyone's time and reduce friction. Of course that sometimes means writing an analyzer :)

GitUI/CommandsDialogs/RepoHosting/ViewPullRequestsForm.cs
drewnoakes: In theory in IL it's possible to throw an arbitrary object. Catching `Exception` and an unspecified catch are technically different. I have never seen it make a difference in practice.

This is another example of something where an analyzer should step in to enfore this kind of thing according to the project's preferences IMO.

RussKie: I think here we'd want to let the user know that we couldn't reach certain remotes. Otherwise if the user won't see some of the expected remotes they may not be able to immediately reconcile it.
Perhaps show a task dialog-based message:
```
We couldn't fetch the following repositories:
  - https://bla/bla/bla.git
  - https://foo/foo/foo.git
```

pmiossec: I did something like that first but it will be more difficult to achieve: the error is displayed twice because the method is called from 2 places and I was not able to understand if that's normal and what will be the consequences to remove the first call :(

RussKie:
Ok, leave it like this but please add a comment

pmiossec:
> Ok, leave it like this but please add a comment
> 

I also think it is better to display a message so I can introduce a boolean to avoid the display of message when initializing on form load. What do you think?

pmiossec:
Added display of errors to the user. Added screenshots in the first post.

GitUI/CommandsDialogs/RepoHosting/CreatePullRequestForm.cs
```diff
@@ -132,31 +141,38 @@ private void PopulateBranchesComboAndEnableCreateButton(IHostedRemote remote, Co
...
+                                comboBox.SelectedIndex = selectItem;
+                            }
 
-                        if (branches.Count > 0)
+                            _createBtn.Enabled = true;
+                        }
+                        catch (Exception ex)
                         {
-                            comboBox.SelectedIndex = selectItem;
+                            MessageBox.Show(string.Format(_strRemoteInError.Text, remote.DisplayData) + Environment.NewLine + Environment.NewLine + ex.Message, TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Warning);
```
RussKie, line 174: A better error presentation:

```
[Error]

Remote: <xyz/uvw>
Error: Failed to load branches.
<...>
```

```
[Error]

Remote: <xyz/uvw>
Error: Remote ignored.
<...>
```

pmiossec: @RussKie Is that what you have in mind?

image

image

RussKie: Yes, though it looks a little awkward. 

I think it may look better if tweak it a little more:

* use TaskDialog (preferably from System.Windows.Forms namespace)
* Reformat as:
```
[ Remote ignored ] (<-- caption)
---------------------
The GitHub authentica..... (<-- main instruction aka Heading)

Remote: xx/yy/zz (Text)
```

Do you think you could try that?

pmiossec: I did something like that using `Microsoft.WindowsAPICodePack.Dialogs`.
Indeed it was looking better.
I rebased and covert it to `System.Windows.Forms.TaskDialog`.
Unfortunately, the plugins no more load in .net 5 version.

So even if it compiles, we are unable to test it and we will have to wait for plugins loading....

RussKie:
Yep.
I'm working on reinstating all things back. I'm under a massive pump at
work atm, so have to prioritise things.

RussKie:
Let's merge this, and revisit when plugins are working again

pmiossec:
> Let's merge this, and revisit when plugins are working again

Yes, it won't change the face of the world.
I have rebased and squashed some commits...

RussKie:
Merge at will

PR #9082, opened 2021-04-17
Fix settings saving for "Detailed" section by selected setting source

ivangrek:
## Proposed changes

- Fix settings saving for "Detailed" section by selected setting source.
- Provide a way to work with settings without being tied to a specific implementation and source.
- There is plan to change the CurrentSettings type to ISettingsSource.
- Small interface for each type of setting.

~For discussion without code formatting~.

## Test methodology 

- manually

## Test environment(s) 

- Git Extensions 33.33.33
- Build f840a0ab83a461035f816951fe39c752a92abc4c
- Git 2.30.2.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
@@ -1,7 +1,6 @@
 ﻿using System;
-using GitCommands;
 using GitCommands.Settings;
-using Microsoft;
+using GitUIPluginInterfaces.Settings;
 
 namespace GitUI.CommandsDialogs.SettingsDialog.Pages
 {
@@ -14,37 +13,9 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
             InitializeComplete();
         }
 
-        private DetailedGroup DetailedSettings
-        {
-            get
-            {
-                Validates.NotNull(CurrentSettings);
-                return CurrentSettings.Detailed;
-            }
-        }
-
         public static SettingsPageReference GetPageReference()
         {
             return new SettingsPageReferenceByType(typeof(DetailedSettingsPage));
-        }
-
-        protected override void Init(ISettingsPageHost pageHost)
-        {
-            base.Init(pageHost);
-            BindSettingsWithControls();
-        }
-
-        protected override void PageToSettings()
-        {
-            AppSettings.SmtpServer = SmtpServer.Text;
-            if (int.TryParse(SmtpServerPort.Text, out var port))
-            {
-                AppSettings.SmtpPort = port;
-            }
-
-            AppSettings.SmtpUseSsl = chkUseSSL.Checked;
-
-            base.PageToSettings();
         }
 
         protected override void OnRuntimeLoad()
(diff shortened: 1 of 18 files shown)


GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
```diff
@@ -36,13 +36,22 @@ protected override void Init(ISettingsPageHost pageHost)
...
+                if (int.TryParse(SmtpServerPort.Text, out var port))
+                {
+                    detailedSettings.SmtpPort = port;
+                }
 
-            AppSettings.SmtpUseSsl = chkUseSSL.Checked;
+                detailedSettings.SmtpUseSsl = chkUseSSL.Checked;
+
+                SmtpServerPort.Text = detailedSettings.SmtpPort.ToString();
+                chkUseSSL.Checked = detailedSettings.SmtpUseSsl;
```
gerhardol, line 53: Any reason for this?
(at least checked should be redundant)

GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
```diff
@@ -36,13 +36,22 @@ protected override void Init(ISettingsPageHost pageHost)
 
         protected override void PageToSettings()
         {
-            AppSettings.SmtpServer = SmtpServer.Text;
-            if (int.TryParse(SmtpServerPort.Text, out var port))
+            if (CurrentSettings is ISettingsSource settingsSource)
             {
-                AppSettings.SmtpPort = port;
-            }
+                var detailedSettings = settingsSource.Detailed();
```
gerhardol, line 41: nit: Using type is the preferred setting (I believe it is OK here)

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
```diff
@@ -0,0 +1,52 @@
...
+    public interface IDetailedSettings
+    {
+        string SmtpServer { get; set; }
+
+        int SmtpPort { get; set; }
+
+        bool SmtpUseSsl { get; set; }
+    }
+
+    public static class DetailedSettingsDefaults
```
gerhardol, line 12: There is a one class per file rule (no enforcement)

gerhardol (commented):
What is the motivation for this change?
Not clear for me from the description.
Is this planned for all settings?
The change look good to me, but it is more code for the same functionality

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public const string SmtpServerDefault = "smtp.gmail.com";
+        public const int SmtpPortDefault = 465;
+        public const bool SmtpUseSslDefault = true;
```
RussKie, line 16: These are not "constants" in the true meaning of the word, and can invariably change.
```suggestion
        public static readonly string SmtpServerDefault = "smtp.gmail.com";
        public static readonly int SmtpPortDefault = 465;
        public static readonly bool SmtpUseSslDefault = true;
```

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
```diff
@@ -0,0 +1,52 @@
+﻿namespace GitUIPluginInterfaces
```
RussKie, line 1: This adds several layers of abstraction, and I personally can't see any benefits for doing this. I would appreciate it if you could detail what the pros and cons of your design proposal, and why it is better than, say, something like this:
```cs
    public class DetailedSettings
    {
        public static readonly string SmtpServerDefault = "smtp.gmail.com";
        public static readonly int SmtpPortDefault = 465;
        public static readonly bool SmtpUseSslDefault = true;

        public DetailedSettings(ISettingsSource settingsSource)
        {
            ...
        }

        public string SmtpServer { get => ...; set => ...; } = SmtpServerDefault;
        public int SmtpPort { get => ...; set => ...; } = SmtpPortDefault;
        public bool SmtpUseSsl { get => ...; set => ...; } = SmtpUseSslDefault;
    }
```

GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
ivangrek: Existing code.

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
ivangrek: These values will not change.
What are constants for you?

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
ivangrek: I thought so. 
I left the constants separately, if they are needed for access, since the implementation will be internal.

`internal sealed class DetailedSettings : IDetailedSettings`

GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
ivangrek: What would you like to change?

ivangrek:
`What is the motivation for this change?`

Reduce the coherence of the code, simplify the implementation, prepare the code for further transformations.

`Is this planned for all settings?`

Yes for all. 
Also reducing the use of the static AppSettings class.

`The change look good to me, but it is more code for the same functionality.`

At the next stages, all unnecessary will be reduced.

ivangrek:
@gerhardol, when using AppSettings here, we only save the global settings. Try changing these values in the current implementation.

image

GitUI/CommandsDialogs/SettingsDialog/Pages/DetailedSettingsPage.cs
ivangrek: 👍

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
ivangrek: Moved to the DetailedSettings class.

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
mstv: `public` constants are compiled into every library. Hence on change, all dependent DLLs would need to be rebuilt and updated together.

ivangrek: Class is internal.

`internal sealed class DetailedSettings : IDetailedSettings`

Subsequently, all constants can go to the default configuration and, when used in a cascade, override the value that does not exist.

ivangrek:
Please do you have any objections?

gerhardol:
> Please do you have any objections?

No strong objections, even if I prefer the settings with only global scope to only show that in settings (no local).
I did a quick check and this is the only situation that seem to mix global/local settings (other than smtp could be global too).

If this is the first step to add local for all settings, I would vote no.

Plugins/GitUIPluginInterfaces/ApplicationSettings.cs
RussKie: `SmtpPortDefault = 465` this is not a constant. Google can change it any day.

`const` is a compile time artifact, it is burned into the code. So any assembly referencing a constant must be recompiled if `const` value ever changed. That's why `public const` must be avoided at all cost.
`readonly` is a runtime artifact.

ivangrek: > SmtpPortDefault = 465 this is not a constant. Google can change it any day.

From this point of view, contant does not exist at all. 😄 

For us, this is the default and is constant. If the port changes, then we change the setting, but the default value will be the same.

PR #9095, opened 2021-04-21
c# new var

gerhardol:
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features

Part of #8788, #8728
Follow up to #8729, covering more cases

Done after getting requests to change existing code in some PRs

## Proposed changes

Use C#9 new() handling and remove var types.

Submitted in commits with the regex I used to replace the types.
Some hundreds of manual changes too.
To be squashed.

This could be mostly "replayed" but I prefer to merge this now...

```
^(\s*(using )?)var(?<eq> \w+\s+=\s+)new\s+(((?<ga>GitArgumentBuilder)(?<gaa>.*[^;]*;$))|((?<in>\w+<.*>)(?<ina>\(\);$))|((?<st>\w+(<[^>]+>)?)(?<sta>(\([^\.]*\));?\s*$)))
 $1${ga}${in}${st}${eq}new${gaa}${ina}${sta}
Some incorrect changes manually converted

^(\s*(using )?)var(?<eq>\s+\w+\s+=\s+)new\s+(?<st>\w+(<[^>]+>)?)(?<sta>\s*$)
$1${st}${eq}new()${sta}

^(\s*(using )?)var(?<eq> \w+\s+=\s+)new\s+(((?<ga>GitArgumentBuilder)(?<gaa>.*[^;]*;$))|((?<in>\w+<.*>)(?<ina>\(\);$))|((?<st>\w+(<[^>]+>)?)(?<sta>(\(.*\));?\s*$)))
$1${ga}${in}${st}${eq}new${gaa}${ina}${sta}
Slightly to aggressive, some reverted

Manual changes, search using
^(\s*(using )?)var\s+(?<eq>\w+\s+=\s+)new\s+
```

## Test methodology 

Manual - currently little runtime but this should only be a reformatting

----

### DIFF
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -18,7 +18,7 @@ namespace GitCommands
     public static class ExecutableExtensions
     {
         private static readonly Regex _ansiCodePattern = new(@"\u001B[\u0040-\u005F].*?[\u0040-\u007E]", RegexOptions.Compiled);
-        private static readonly Lazy<Encoding> _defaultOutputEncoding = new Lazy<Encoding>(() => GitModule.SystemEncoding, false);
+        private static readonly Lazy<Encoding> _defaultOutputEncoding = new(() => GitModule.SystemEncoding, false);
 
         /// <summary>
         /// Launches a process for the executable and returns its output.
@@ -70,7 +70,7 @@ namespace GitCommands
             CommandCache? cache = null,
             bool stripAnsiEscapeCodes = true)
         {
-            var sb = new StringBuilder();
+            StringBuilder sb = new();
             foreach (var batch in batchArguments)
             {
                 sb.Append(executable.GetOutput(batch.Argument, input, outputEncoding, cache, stripAnsiEscapeCodes));
@@ -116,8 +116,8 @@ namespace GitCommands
                 process.StandardInput.Close();
             }
 
-            var outputBuffer = new MemoryStream();
-            var errorBuffer = new MemoryStream();
+            MemoryStream outputBuffer = new();
+            MemoryStream errorBuffer = new();
             var outputTask = process.StandardOutput.BaseStream.CopyToAsync(outputBuffer);
             var errorTask = process.StandardError.BaseStream.CopyToAsync(errorBuffer);
             var exitTask = process.WaitForExitAsync();
@@ -356,16 +356,16 @@ namespace GitCommands
             }
 
             using var process = executable.Start(arguments, createWindow: false, redirectInput: writeInput is not null, redirectOutput: true, outputEncoding);
-            var outputBuffer = new MemoryStream();
-            var errorBuffer = new MemoryStream();
+            MemoryStream outputBuffer = new();
+            MemoryStream errorBuffer = new();
             var outputTask = process.StandardOutput.BaseStream.CopyToAsync(outputBuffer);
             var errorTask = process.StandardError.BaseStream.CopyToAsync(errorBuffer);
 
             if (writeInput is not null)
             {
 #if DEBUG
-                using MemoryStream mem = new MemoryStream();
-                using StreamWriter sw = new StreamWriter(mem);
+                using MemoryStream mem = new();
+                using StreamWriter sw = new(mem);
                 writeInput(sw);
 
                 System.Diagnostics.Debug.WriteLine($"git {arguments} {Encoding.UTF8.GetString(mem.ToArray(), 0, (int)mem.Length)}");
--- a/GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
+++ b/GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
@@ -13,7 +13,7 @@ namespace GitUI.Editor.Diff
 
         public DiffLinesInfo Analyze(string diffContent)
         {
-            var ret = new DiffLinesInfo();
+            DiffLinesInfo ret = new();
             var isCombinedDiff = PatchProcessor.IsCombinedDiff(diffContent);
             var lineNumInDiff = 0;
             var leftLineNum = DiffLineInfo.NotApplicableLineNum;
@@ -31,7 +31,7 @@ namespace GitUI.Editor.Diff
                 lineNumInDiff++;
                 if (line.StartsWith("@"))
                 {
-                    var meta = new DiffLineInfo
+                    DiffLineInfo meta = new()
                     {
                         LineNumInDiff = lineNumInDiff,
                         LeftLineNumber = DiffLineInfo.NotApplicableLineNum,
@@ -48,7 +48,7 @@ namespace GitUI.Editor.Diff
                 }
                 else if (isHeaderLineLocated && isCombinedDiff)
                 {
-                    var meta = new DiffLineInfo
+                    DiffLineInfo meta = new()
                     {
                         LineNumInDiff = lineNumInDiff,
                         LeftLineNumber = DiffLineInfo.NotApplicableLineNum,
@@ -76,7 +76,7 @@ namespace GitUI.Editor.Diff
                 }
                 else if (isHeaderLineLocated && IsMinusLine(line))
                 {
-                    var meta = new DiffLineInfo
+                    DiffLineInfo meta = new()
                     {
                         LineNumInDiff = lineNumInDiff,
                         LeftLineNumber = leftLineNum,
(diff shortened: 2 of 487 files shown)


mstv (commented):
Comprehensive - thank you

mstv (approved):

gerhardol:
@msftbot merge in 4 days

RussKie:
This will likely completely stall the migration to .NET 5. I'm already at the point of giving up due to all the conflicts I have to deal with...

gerhardol:
> This will likely completely stall the migration to .NET 5. I'm already at the point of giving up due to all the conflicts I have to deal with...

This is not hard to repeat, the commit messages contains the changes done, except the final (not fixup) commit that was done manually (the search phrase is included, the change will require a couple of hours).
For this to work, the .NET5 branch must be rebased on the commit prior to this is merged. I can fix the rebase on this PR if you prefer.
If this is not merged, I expect several of these will contribute to changes later.

While this is quite large (about 1700 changes), the combination is smaller than the previous c#7/8/9 changes and that merge conflicts are much smaller than the other changes.

Feel free to stop the merge if you prefer.

Rebasing to resolve merge conflicts.

RussKie:
Please base this work on .NET branch.

gerhardol:
To be reapplied after .NET5

gerhardol:
"Replayed" (regex and manual) the changes on master

GitUI/Editor/FileViewer.cs
```diff
@@ -1027,7 +1027,7 @@ private Task ViewItemAsync(string fileName, bool isSubmodule, Func<Image?> getIm
                                     ResetView(ViewMode.Text, null);
 
                                     var text = getFileText();
-                                    var summary = new StringBuilder()
+                                    StringBuilder summary = new StringBuilder()
```
drewnoakes, line 1030: Intentional?

GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
```diff
@@ -13,7 +13,7 @@ public class DiffLineNumAnalyzer
 
         public DiffLinesInfo Analyze(string diffContent)
         {
-            var ret = new DiffLinesInfo();
+            DiffLinesInfo ret = new();
             var isCombinedDiff = PatchProcessor.IsCombinedDiff(diffContent);
             var lineNumInDiff = 0;
             var leftLineNum = DiffLineInfo.NotApplicableLineNum;
```
drewnoakes, line 19: Discussion point: For projects that prefer var, I think a change like this makes the code look a bit inconsistent, as the variables no longer line up. It's a minor point, and of course any code style can be learned, but it does seem a bit harder to read with this change.

I feel much more strongly about the case where the type name is completely removed (e.g. `foo.Bar(new())`), but that does not appear to be present in this PR.

GitUI/Editor/FileViewer.cs
drewnoakes: Also, was there not a Roslyn codefix to make these changes? If one exists, why didn't you use it? If it has issues, I'd like to report them.

gerhardol: This was intentional. new() cannot be used here and I reverted the change. In this case I followed IDE0008 and kept the type. Will remove that from this PR.

It is not possible to apply IDE0090 without removing `var` first. So IDE0008 has to be applied which removes all var declarations before IDE0090 can be applied.
Adding a commit to apply IDE0090, there are some usages, especially in definitions.
Who could imagine a compiler was better than a regex?

GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
drewnoakes: I like the changes in e1168fb4885c7c1d0b649c972551a71c7aa2593a.

RussKie: I don't like `var` because more often than not it makes the code opaque (e.g. `var result = GetSomething();`), and prefer to avoid it.
The new `new` syntax makes it better in my opinion.

Also I think I rather like `foo.Bar(new())` instead of `foo.Bar(new Some.CrazyLongTypeName.ThatIDontCareAbout())` because it provides enough information for me. Obviously, like with anything else, a common sense must be applied as this may become unreadable `foo.Bar(new(), new(new, new()), new())`

RussKie:
> Should I change this to apply IDE0008 (remove all var declarations) followed by IDE0090 (new())?

Sure, if you think it's worth it.

I'm happy for this to merge (once the build is green), and then have a follow up.

gerhardol:
> > Should I change this to apply IDE0008 (remove all var declarations) followed by IDE0090 (new())?
> 
> Sure, if you think it's worth it.
> 
> I'm happy for this to merge (once the build is green), and then have a follow up.

I personally am no fan of removing all var, type is often obvious and I prefer shorter code.
But if IDE0008 is the preference it may be better to apply it universally.

A little surprised that Codefactor found 8 fixed issues and 10 new after this change...

Would like to squash and merge this PR a couple of days after next approval.

GitUI/HelperDialogs/FormResetCurrentBranch.cs
```diff
@@ -24,7 +24,7 @@ public enum ResetType
         }
 
         public static FormResetCurrentBranch Create(GitUICommands commands, GitRevision revision, ResetType resetType = ResetType.Mixed)
-            => new FormResetCurrentBranch(commands, revision ?? throw new NotSupportedException(TranslatedStrings.NoRevision), resetType);
+            => new(commands, revision ?? throw new NotSupportedException(TranslatedStrings.NoRevision), resetType);
```
mstv, line 27: This omission is located at the edge of readability *for me*. Not requesting a change.

GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
drewnoakes: It seems contradictory to dislike var due to invisible types, while also preferring new Foo(new())). Am I missing something?

One issue with new Foo(new())) is that if you add an overload of Foo, your code no longer compiles.

gerhardol: > It seems contradictory to dislike var due to invisible types, while also preferring new Foo(new())). Am I missing something?

I find something like the following fine, then the type is "close".
```c#
List<DateTime> l = new() { new() };
```

> One issue with new Foo(new())) is that if you add an overload of Foo, your code no longer compiles.

Need to be handled by the designer adding the overload or updating a library.

But I do not want to use `new()` much more than in this PR or really remove `var` generally.

mstv (approved):
Thank you

---

> A little surprised that Codefactor found 8 fixed issues and 10 new after this change...

Codefactor seems to be commenting every touched function regardless of the change (seen in other PR, too).

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -116,8 +116,8 @@ public static class ExecutableExtensions
                 process.StandardInput.Close();
             }
 
-            var outputBuffer = new MemoryStream();
-            var errorBuffer = new MemoryStream();
+            MemoryStream outputBuffer = new();
+            MemoryStream errorBuffer = new();
```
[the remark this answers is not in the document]
mstv: For another PR: applies at least to `MemoryStream` and `*Form` IIUIC.

GitUI/HelperDialogs/FormResetCurrentBranch.cs
gerhardol: I see your point in this case, even I prefer the change also here.
It is hard to have consistent rules without reasonable exceptions. I do not really want IDE0008 to be enforced for instance but to keep it as a recommendation where needed.

gerhardol:
https://github.com/gitextensions/gitextensions/pull/9182#discussion_r631432160

Adding warning if not followed (no compile error)
image

GitUI/Editor/Diff/DiffLineNumAnalyzer.cs
RussKie: > It seems contradictory to dislike var due to invisible types, while also preferring new Foo(new())). Am I missing something?

I see no contradictions here.
I want to see what type is being returned from a call so I can reason about how it is used. But when I'm see a type constructed I don't _really_ care what goes in (the compiler will check).

GitCommands/Git/ExecutableExtensions.cs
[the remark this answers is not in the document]
drewnoakes: Disposing a `MemoryStream` does nothing, so this change is technically a pessimisation. But it's not a big deal either way.

gerhardol:
Squashed with merge conflict resolve on master. Left a tmp/ branch behind
Set to merge (will need a conflict resolve rebase I assume) in four days.

PR #9097, opened 2021-04-22
Reduce allocations for RevisionGraph lane straightening

mstv:
Follow-up to #9050

## Proposed changes

- Allocate `HashSet<int> _gaps` only if lanes are straightened in a `RevisionGraph` row

## Screenshots 

N/A

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build d488e497cc6244483794abedb8e4779495ba861c (Dirty)
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
@@ -40,7 +40,7 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         /// <summary>
         /// Contains the gaps created by <cref>MoveLanesRight</cref>
         /// </summary>
-        private HashSet<int> _gaps = new();
+        private HashSet<int>? _gaps;
 
         /// <summary>
         /// The cached lanecount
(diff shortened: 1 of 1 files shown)


RussKie (approved):

gerhardol (approved):

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
```diff
@@ -185,6 +185,7 @@ public int GetLaneIndexForSegment(RevisionGraphSegment revisionGraphRevision)
 
         public void MoveLanesRight(int fromLane)
         {
+            _gaps ??= new();
             int nextGap = _gaps.Min(lane => lane > fromLane ? lane : null) ?? int.MaxValue;
```
drewnoakes, line 189:
```suggestion
            int nextGap = _gaps?.Min(lane => lane > fromLane ? lane : null) ?? int.MaxValue;
```

And move:

```c#
            _gaps ??= new();
```

Down before the first actual write on line 205.

mstv: Done because it is clearer. Since this function is called only if there are lanes to be moved, it has no effect on the number of allocations.

PR #9107, opened 2021-04-24
Start using MEF

RussKie:
## Proposed changes

Use MEF tp compose and reference the following types:
- `AppTitleGenerator`
- `RepositoryDescriptionProvider`
- `WindowsJumpListManager`

## Test methodology 

- manual
- unit tests

----

### DIFF
--- a/GitCommands/AppTitleGenerator.cs
+++ b/GitCommands/AppTitleGenerator.cs
@@ -1,3 +1,5 @@
+using System;
+using System.ComponentModel.Composition;
 using GitCommands.UserRepositoryHistory;
 using GitUIPluginInterfaces;
 
@@ -20,17 +22,18 @@ namespace GitCommands
     /// <summary>
     /// Generates application title.
     /// </summary>
+    [Export(typeof(IAppTitleGenerator))]
     public sealed class AppTitleGenerator : IAppTitleGenerator
     {
+        private readonly IRepositoryDescriptionProvider _descriptionProvider;
 #if DEBUG
         private static string? _extraInfo;
 #endif
 
-        private readonly IRepositoryDescriptionProvider _description;
-
-        public AppTitleGenerator(IRepositoryDescriptionProvider description)
+        [ImportingConstructor]
+        public AppTitleGenerator(IRepositoryDescriptionProvider descriptionProvider)
         {
-            _description = description;
+            _descriptionProvider = descriptionProvider;
         }
 
         /// <summary>
@@ -48,7 +51,7 @@ namespace GitCommands
 
             branchName = branchName?.Trim('(', ')') ?? "no branch";
 
-            var description = _description.Get(workingDir);
+            var description = _descriptionProvider.Get(workingDir);
 
 #if DEBUG
             return $"{description} ({branchName}) - {AppSettings.ApplicationName}{_extraInfo}";
--- a/GitCommands/Git/GitDirectoryResolver.cs
+++ b/GitCommands/Git/GitDirectoryResolver.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.Composition;
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
@@ -21,6 +22,7 @@ namespace GitCommands.Git
     /// <summary>
     /// Resolves the location of .git folder.
     /// </summary>
+    [Export(typeof(IGitDirectoryResolver))]
     public sealed class GitDirectoryResolver : IGitDirectoryResolver
     {
         private readonly IFileSystem _fileSystem;
--- a/GitUI/WindowsJumpListManager.cs
+++ b/GitUI/WindowsJumpListManager.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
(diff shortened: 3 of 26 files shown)


ivangrek (commented):
Maybe like this:

image

Or using WinForm as service.

https://www.thecodebuzz.com/dependency-injection-windows-form-desktop-app-net-core/

We could start simple.

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
```diff
@@ -16,10 +18,17 @@ public class AppTitleGeneratorTests
         [SetUp]
         public void Setup()
         {
+            // Create the test instance
+            _appTitleGenerator = new();
```
ivangrek, line 22: Think we need to use full name of class istead `new()`.

RussKie:
We already are using MEF, so I'm finding it hard to justify introduction of another DI framework.

> Or using WinForm as service.
> 
> [thecodebuzz.com/dependency-injection-windows-form-desktop-app-net-core](https://www.thecodebuzz.com/dependency-injection-windows-form-desktop-app-net-core/)
> 

This is really cool! Thanks for the link.

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
RussKie: In cases like this (and in non-prod code) I dont think it particular matters. 

That is, "_new up the known instance_" vs "_get a response of <?> type by calling some method_".
```cs
private AppTitleGenerator _appTitleGenerator;

// Is it really that much more (un)readable?
 _appTitleGenerator = new();
 _appTitleGenerator = new AppTitleGenerator();
```
vs
```cs
var appTitleGenerator = someType.ExecuteSomeMethod();
AppTitleGenerator appTitleGenerator = someType.ExecuteSomeMethod();
```

GitCommands/AppTitleGenerator.cs
```diff
@@ -21,18 +22,19 @@ public interface IAppTitleGenerator
...
-        private readonly IRepositoryDescriptionProvider _description;
-
-        public AppTitleGenerator(IRepositoryDescriptionProvider description)
-        {
-            _description = description;
-        }
+#pragma warning disable CS0649
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        [Import(typeof(IRepositoryDescriptionProvider))]
+        private IRepositoryDescriptionProvider _description;
```
sharwell, line 35: 💡 I highly recommend using constructor injection instead of property/field injection.

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
```diff
@@ -16,10 +18,17 @@ public class AppTitleGeneratorTests
...
+
+            // Create mocks for dependencies
             _repositoryDescriptionProvider = Substitute.For<IRepositoryDescriptionProvider>();
             _repositoryDescriptionProvider.Get(Arg.Any<string>()).Returns(ShortName);
 
-            _appTitleGenerator = new AppTitleGenerator(_repositoryDescriptionProvider);
+            // Create the test MEF container and compose the mocks into the test instance
+            CompositionContainer container = new();
+            container.ComposeExportedValue(_repositoryDescriptionProvider);
+            container.ComposeParts(_appTitleGenerator);
```
sharwell, line 31: 💡 I highly recommend using `TestComposition` from dotnet/roslyn for constructing a MEF container for testing.

https://github.com/dotnet/roslyn/blob/main/src/Workspaces/CoreTestUtilities/MEF/TestComposition.cs

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
ivangrek: I am proceeding from line-by-line reading, and the type must occur once in a line.
Otherwise I don't understand when we use it.

ivangrek:
> This is really cool! Thanks for the link.

The form can be thought of as a request to a remote system, where the response is generated not by a machine, but by a person.

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
RussKie: This wasn't a straight forward thing, but I think I copied the necessary bits. Appreciate your review again.

GitCommands/AppTitleGenerator.cs
```diff
@@ -20,17 +22,18 @@ public interface IAppTitleGenerator
...
 #endif
 
-        private readonly IRepositoryDescriptionProvider _description;
-
-        public AppTitleGenerator(IRepositoryDescriptionProvider description)
+        [ImportingConstructor]
+        public AppTitleGenerator(IRepositoryDescriptionProvider descriptionProvider)
         {
-            _description = description;
+            _descriptionProvider = descriptionProvider ?? throw new ArgumentNullException(nameof(descriptionProvider));
```
sharwell, line 36: 💡 This null check is not necessary. MEF will not instantiate this part if a non-null instance of `IRepositoryDescriptionProvider` is not available.

GitCommands/Git/GitDirectoryResolver.cs
```diff
@@ -15,12 +16,14 @@ public interface IGitDirectoryResolver
...
         /// <param name="repositoryPath">The repository working folder.</param>
         /// <returns>The resolved location of .git folder.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="repositoryPath"/> is <see langword="null"/>.</exception>
         string Resolve(string repositoryPath);
     }
 
     /// <summary>
     /// Resolves the location of .git folder.
     /// </summary>
+    [Export(typeof(IGitDirectoryResolver))]
```
sharwell, line 26: 💡 I recommend making the following changes:

1. Eliminate all but one constructor
2. Mark the remaining constructor with `[ImportingConstructor]`

GitUI/WindowsJumpListManager.cs
```diff
@@ -18,24 +19,31 @@ namespace GitUI
...
             }
         }
 
+        [ImportingConstructor]
+        public WindowsJumpListManager(IRepositoryDescriptionProvider repositoryDescriptionProvider)
+        {
+            _repositoryDescriptionProvider = repositoryDescriptionProvider;
+        }
+
         ~WindowsJumpListManager()
```
sharwell, line 47: 💡 MEF will dispose instantiated parts when the container is disposed. There is no need to include a finalizer here.

UnitTests/GitCommands.Tests/AppTitleGeneratorTests.cs
```diff
@@ -1,80 +1,95 @@
...
         {
-            _repositoryDescriptionProvider = Substitute.For<IRepositoryDescriptionProvider>();
-            _repositoryDescriptionProvider.Get(Arg.Any<string>()).Returns(ShortName);
+            TestComposition composition = TestComposition.Empty
+                .AddParts(typeof(MockRepositoryDescriptionProvider))
+                .AddParts(typeof(IAppTitleGenerator))
+                .AddParts(typeof(AppTitleGenerator));
 
-            _appTitleGenerator = new AppTitleGenerator(_repositoryDescriptionProvider);
+            _mefExportProvider = composition.ExportProviderFactory.CreateExportProvider();
```
sharwell, line 26: ❗ Do not reuse a MEF container across tests. The `TestComposition` can be reused, but the result of a call to `CreateExportProvider` should never appear in a OneTimeSetUp method.

GitCommands/Git/GitDirectoryResolver.cs
RussKie: Can you please help me understanding how to register `FileSystem -> IFileSystem`?

```diff
diff --git a/GitCommands/Git/GitDirectoryResolver.cs b/GitCommands/Git/GitDirectoryResolver.cs
index 1d81f491f..d3610c9fc 100644
--- a/GitCommands/Git/GitDirectoryResolver.cs
+++ b/GitCommands/Git/GitDirectoryResolver.cs
@@ -28,16 +28,12 @@ public sealed class GitDirectoryResolver : IGitDirectoryResolver
     {
         private readonly IFileSystem _fileSystem;
 
+        [ImportingConstructor]
         public GitDirectoryResolver(IFileSystem fileSystem)
         {
             _fileSystem = fileSystem;
         }
 
-        public GitDirectoryResolver()
-            : this(new FileSystem())
-        {
-        }
-
```

RussKie:
@msftbot merge in 48 hours

PR #9018, opened 2021-03-20
Ignore git clone prefix when checking clipboard for source URL

nyankoframe:
Fixes #6260 

## Proposed changes
 
-  Detects whether clipboard contents contain a potential URL using regex; if so, clipboard contents are provided as-is allowing user to modify contents prior to clone operation.

## Test methodology 

- Manual testing
- Added new tests to PathUtilTest.IsUrl and PathUtilTest.CanBeGitURL

## Test environment(s) 

- GIT 2.31.0.windows.1
- Windows 10.0.19042.868

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9116, opened 2021-04-26
Update Clone dialog's handling of clipboard text

### LINKED ISSUE #6260: Detect 'git clone' prefix and ignore
## Current behaviour

I copy a branch from gerrit code review server (git clone ssh://...), want to clone a new repo and got and error ( it tries to execute git clone git clone ... )

## Expected behaviour
The prefix git clone is detected automatically and will be ignored

## Steps to reproduce

Copy and paste a link from gerrit

Current (3.xx)

## Environment

- Git Extensions 3.1.0.5877
- Git 2.19.0.windows.1
- .NET Framework 4.7.3416.0
- DPI 96dpi (no scaling)

nyankoframe:
Fixes #6260 

## Proposed changes

- Adds a new method in FormClone to extract a URL from a text string.
- Updates Clone form to use new method

## Test methodology 

- Manual testing
- Added new integration test in FormCloneTests.cs

## Test environment(s) 

- GIT 2.31.0.windows.1
- Windows 10.0.19042.868

----

### DIFF
--- a/GitCommands/PathUtil.cs
+++ b/GitCommands/PathUtil.cs
@@ -71,23 +71,6 @@ namespace GitCommands
             return !Regex.IsMatch(fileName, @"^(\w+):\/\/([\S]+)");
         }
 
-        /// <summary>
-        /// A naive way to check whether the given path is a URL by checking
-        /// whether it starts with either 'http', 'ssh' or 'git'.
-        /// </summary>
-        /// <param name="path">A path to check.</param>
-        /// <returns><see langword="true"/> if the given path starts with 'http', 'ssh' or 'git'; otherwise <see langword="false"/>.</returns>
-        [Pure]
-        public static bool IsUrl(string path)
-        {
-            return !string.IsNullOrEmpty(path)
-                && (path.StartsWith("http:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("https:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("git:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("ssh:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("file:", StringComparison.CurrentCultureIgnoreCase));
-        }
-
         public static bool CanBeGitURL(string url)
         {
             if (string.IsNullOrWhiteSpace(url))
@@ -95,7 +78,7 @@ namespace GitCommands
                 return false;
             }
 
-            return IsUrl(url)
+            return Uri.IsWellFormedUriString(url, UriKind.Absolute)
                    || url.EndsWith(".git", StringComparison.CurrentCultureIgnoreCase)
                    || GitModule.IsValidGitWorkingDir(url);
         }
--- a/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCloneTests.cs
+++ b/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCloneTests.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using CommonTestUtils;
+using FluentAssertions;
+using GitCommands.Git;
+using GitUI;
+using GitUI.CommandsDialogs;
+using NUnit.Framework;
+
+namespace GitExtensions.UITests.CommandsDialogs
+{
+    [Apartment(ApartmentState.STA)]
+    public class FormCloneTests
+    {
+        // Created once for the fixture
+        private ReferenceRepository _referenceRepository;
+
+        // Created once for each test
+        private GitUICommands _commands;
+
+        [SetUp]
+        public void SetUp()
+        {
+            if (_referenceRepository is null)
+            {
+                _referenceRepository = new ReferenceRepository();
+            }
+            else
+            {
+                _referenceRepository.Reset();
+            }
+
+            _commands = new GitUICommands(_referenceRepository.Module);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+        }
+
+        [OneTimeTearDown]
+        public void OneTimeTearDown()
+        {
+            _referenceRepository.Dispose();
+        }
+
+        [TestCase(null, false, "")]
+        [TestCase("", false, "")]
+        [TestCase(" ", false, "")]
+        [TestCase("blah", false, "")]
+        [TestCase("git clone https://github.com/gitextensions/gitextensions && cd gitextensions", true, "https://github.com/gitextensions/gitextensions")]
+        [TestCase("git clone ssh://username@gerrit-server:/PROJECT", true, "ssh://username@gerrit-server:/PROJECT")]
+        [TestCase("git clone https://github.com/gitextensions/gitextensions && git clone https://github.com/gitextensions/git.hub", true, "https://github.com/gitextensions/gitextensions")]
+        public void Test_Url_extract_from_string(
+            string text, bool expected, string expectedUrl)
+        {
+            RunFormTest(
+                form =>
+                {
+                    var accessor = form.GetTestAccessor();
+
+                    accessor.TryExtractUrl(text, out var url).Should().Equals(expected);
+
+                    // No need to compare URL if the result was expected to be false
+                    if (expected)
+                    {
+                        url.Should().Equals(expectedUrl);
+                    }
+                });
+        }
+
+        private void RunFormTest(Action<FormClone> testDriver)
+        {
+            RunFormTest(
+                form =>
+                {
+                    testDriver(form);
+                    return Task.CompletedTask;
+                });
+        }
+
+        private void RunFormTest(Func<FormClone, Task> testDriverAsync)
+        {
+            UITest.RunForm(
+                () =>
+                {
+                    _commands.StartCloneDialog(owner: null, url: null);
+                },
+                testDriverAsync);
+        }
+    }
+}
--- a/UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
+++ b/UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
@@ -101,20 +101,6 @@ namespace GitCommandsTests.Helpers
             Assert.AreEqual(PathUtil.IsLocalFile("ssh://domain\\user@serverip/cache/git/something/something.git"), false);
         }
 
-        [TestCase(null, false)]
-        [TestCase("", false)]
-        [TestCase("    ", false)]
-        [TestCase("http://", true)]
-        [TestCase("HTTPS://www", true)]
-        [TestCase("git://", true)]
-        [TestCase("file:", true)]
-        [TestCase("SSH:", true)]
-        [TestCase("SSH", false)]
-        public void IsUrl(string path, bool expected)
-        {
-            PathUtil.IsUrl(path).Should().Be(expected);
-        }
-
         [Test]
         public void GetFileNameTest()
         {
@@ -347,6 +333,17 @@ namespace GitCommandsTests.Helpers
         [TestCase("https://github.com/gitextensions/gitextensions", true)]
         [TestCase("https://github.com/gitextensions/gitextensions.git", true)]
         [TestCase("github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("HTTPS://MYPRIVATEGITHUB.COM:8080/LOUDREPO.GIT", true)]
+        [TestCase("git://myurl/myrepo.git", true)]
+        [TestCase("github.com/gitextensions", false)]
+        [TestCase("github.com/gitextensions/gitextensions/pull/9018", false)]
+        [TestCase("http://", false)]
+        [TestCase("HTTPS://www", true)]
+        [TestCase("git://", true)]
+        [TestCase("file:", false)]
+        [TestCase("SSH:", true)]
+        [TestCase("SSH", false)]
+        [TestCase("https://myhost:12368/", true)]
         public void CanBeGitURL(string url, bool expected)
         {
             Assert.AreEqual(expected, PathUtil.CanBeGitURL(url));
--- a/contributors.txt
+++ b/contributors.txt
@@ -152,3 +152,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/01/24, bcolaco, Bruno Colaço, bcolaco[@)gmail.com
 2021/01/25, guybark, Guy Barker, guybark(at)microsoft.com
 2021/02/10, mabako, Marcus Bauer, mabako(at)gmail.com
+2021/03/20, nyankoframe, Damit Senanayake, damit(at)outlook.com
--- a/GitCommands/UserRepositoryHistory/LocalRepositoryManager.cs
+++ b/GitCommands/UserRepositoryHistory/LocalRepositoryManager.cs
@@ -85,7 +85,7 @@ namespace GitCommands.UserRepositoryHistory
                 throw new ArgumentException(nameof(repositoryPath));
             }
 
-            if (PathUtil.IsUrl(repositoryPath))
+            if (Uri.IsWellFormedUriString(repositoryPath, UriKind.Absolute))
             {
                 // TODO: throw a specific exception
                 throw new NotSupportedException();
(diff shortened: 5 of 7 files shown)


GitCommands/PathUtil.cs
```diff
@@ -72,20 +77,20 @@ public static bool IsLocalFile(string fileName)
...
         [Pure]
         public static bool IsUrl(string path)
         {
             return !string.IsNullOrEmpty(path)
-                && (path.StartsWith("http:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("https:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("git:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("ssh:", StringComparison.CurrentCultureIgnoreCase)
-                 || path.StartsWith("file:", StringComparison.CurrentCultureIgnoreCase));
+                && UrlRegex.IsMatch(path);
```
drewnoakes, line 93: RegEx parsing of this kind of thing makes me a bit uneasy, as the format is generally quite involved. Why not use `Url.TryParse` instead?

GitCommands/PathUtil.cs
```diff
@@ -72,20 +77,20 @@ public static bool IsLocalFile(string fileName)
         }
 
         /// <summary>
-        /// A naive way to check whether the given path is a URL by checking
-        /// whether it starts with either 'http', 'ssh' or 'git'.
+        /// A slightly less naive way to check whether the given path is a URL by checking
```
drewnoakes, line 80: It's unclear what this is less naive than. I mean, it makes sense in the diff, but won't in the resulting code.

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
```diff
@@ -104,12 +104,14 @@ public void IsLocalFileTest()
...
+        [TestCase("http://", false)]
         [TestCase("HTTPS://www", true)]
-        [TestCase("git://", true)]
-        [TestCase("file:", true)]
-        [TestCase("SSH:", true)]
+        [TestCase("git://", false)]
+        [TestCase("file:", false)]
+        [TestCase("SSH:", false)]
         [TestCase("SSH", false)]
+        [TestCase("rtsp://probably.not.a.git.url/default.mp4", false)]
```
drewnoakes, line 113: The method `IsUrl` suggests it will return true for URLs. This actually is a URL, even if not a git one.

GitCommands/PathUtil.cs
nyankoframe: I'm not sure if there is a Url.TryParse method (unless I'm looking in the wrong place?), but is this what you're referring to?
[System.UriParser.IsWellFormedOriginalString](https://docs.microsoft.com/en-us/dotnet/api/system.uriparser.iswellformedoriginalstring)

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
nyankoframe: You're right, I was trying to be too clever. :) I'll remove this test case from this branch and think about how to update it in master (either remove the test case or rename the method)

GitCommands/PathUtil.cs
drewnoakes: Ah, yes, `Uri.TryCreate` seems like a better match: https://docs.microsoft.com/en-us/dotnet/api/system.uri.trycreate?view=net-5.0#System_Uri_TryCreate_System_String_System_UriKind_System_Uri__

nyankoframe: I've tried using Uri.TryCreate and it returns false if the string has any characters before or after the URL, which was the original problem - someone had a string in clipboard that of the form "git clone https://...." that wasn't getting detected as a valid potential URL because of the "git clone" text before the URL. I feel that using the regex to detect if the string contains a valid URL allows the user to manually remove non-URL text without having GE guess and potentially get it wrong.

nyankoframe:
It looks like #9018 was already merged into the 3.5 milestone, so this PR might be redundant. @RussKie could you confirm? If so, I'll go ahead and close it.

RussKie:
#9018 was merged into the master only

GitCommands/PathUtil.cs
drewnoakes: Given a URL cannot contain spaces, you could take the input string, split it by spaces and check each value.

My concern is that the full URL format is complex, and this regex will not cover all valid cases. The Uri class is far more likely to be correct.

contributors.txt
```diff
@@ -152,3 +152,7 @@ YYYY/MM/DD, github id, Full name, email
 2021/01/24, bcolaco, Bruno Colaço, bcolaco[@)gmail.com
 2021/01/25, guybark, Guy Barker, guybark(at)microsoft.com
 2021/02/10, mabako, Marcus Bauer, mabako(at)gmail.com
+2021/03/11, AntonKedrov, Anton Kedrov, a-k(at)gmx.com
+2021/03/16, JuPrgn, Julien Peyregne, julienprgn(at)gmail.com
+2021/03/20, nyankoframe, Damit Senanayake, damit(at)outlook.com
+2021/03/31, darkcamper, Rubén Ruiz, darkcamper(at)gmail.com
```
RussKie, line 158: This is incorrect, please only add yourself

RussKie (changes requested):

GitCommands/PathUtil.cs
nyankoframe: Thanks for the feedback, I've switched to using `Uri.IsWellFormedUriString` as that allows file paths to not be treated as URIs (`Uri.TryCreate` would consider them as URIs resulting in LocalRepositoryManager methods throwing an exception)

contributors.txt
nyankoframe: Thank you for the clarification. Updated in latest revision.

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
```diff
@@ -347,6 +348,13 @@ public void DeleteWithExtremePrejudice_should_return_true_if_delete_successful()
         [TestCase("https://github.com/gitextensions/gitextensions", true)]
         [TestCase("https://github.com/gitextensions/gitextensions.git", true)]
         [TestCase("github.com/gitextensions/gitextensions.git", true)]
+        [TestCase("git clone https://github.com/gitextensions/gitextensions", true)]
+        [TestCase("HTTPS://MYPRIVATEGITHUB.COM:8080/LOUDREPO.GIT", true)]
+        [TestCase("git://myurl/myrepo.git", true)]
+        [TestCase("git clone https://github.com/gitextensions/gitextensions && cd gitextensions", true)]
+        [TestCase("github.com/gitextensions", false)]
+        [TestCase("github.com/gitextensions/gitextensions/pull/9018", false)]
+        [TestCase("git clone ssh://username@gerrit-server:/PROJECT", true)]
```
RussKie, line 357: Neither of these are correct inputs for `CanBeGitURL` method - because the answer to the question "_Can either of those be a git URL?_" is "_no_".
```cs
        [TestCase("git clone https://github.com/gitextensions/gitextensions", true)]
        [TestCase("git clone https://github.com/gitextensions/gitextensions && cd gitextensions", true)]
        [TestCase("github.com/gitextensions", false)]
        [TestCase("github.com/gitextensions/gitextensions/pull/9018", false)]
        [TestCase("git clone ssh://username@gerrit-server:/PROJECT", true)]
```

We can consider renaming the method to `ContainsGitUrl`, because then the answer becomes "yes". But then the signature must be changed as well to return the valid git URL... So we'll have to make it something like:
```cs
        public static bool TryExtractGitUrl(string url, out string gitUrl)
```

GitCommands/PathUtil.cs
```diff
@@ -72,20 +72,32 @@ public static bool IsLocalFile(string fileName)
...
+            {
+                return false;
+            }
+
+            string[] pathElements = path.Split(' ');
+            bool validUrl = false;
+            foreach (string str in pathElements)
+            {
+                validUrl |= Uri.IsWellFormedUriString(str, UriKind.Absolute);
+            }
```
RussKie, line 98: Likewise this method's purpose is to validate the input string, and not break it apart and check each part. The proposal makes the behaviour surprising - as now the question "Is 'git clone https://github.com/gitextensions/gitextensions' a URL?" results in "Yes", which is unexpected to anyone looking at the signature.

RussKie (changes requested):
Thank you, but the API needs a little more work.

RussKie:
Please rebase the PR on to the master branch. This won't go into 3.5.1

gerhardol:
@nyankoframe Any update? 3.5.1 is hopefully released quite soon

GitCommands/PathUtil.cs
drewnoakes: Agree. If the caller wants to check individual words, they can split the string and call this on each.

nyankoframe:
> @nyankoframe Any update? 3.5.1 is hopefully released quite soon

Sorry for the delay; I was a bit buried at work, but should have some time in the next few days to make the proposed changes.

GitCommands/PathUtil.cs
```diff
@@ -72,20 +72,44 @@ public static bool IsLocalFile(string fileName)
...
-                 || path.StartsWith("file:", StringComparison.CurrentCultureIgnoreCase));
+            if (string.IsNullOrEmpty(path))
+            {
+                return false;
+            }
+
+            return Uri.IsWellFormedUriString(path, UriKind.Absolute);
+        }
+
+        public static bool ExtractURLsFromString(string contents, List<string> urls)
```
drewnoakes, line 96: A small change to the API here seems more idiomatic, somehow.

```suggestion
        public static bool TryExtractURLsFromString(string contents, out List<string> urls)
```

GitCommands/PathUtil.cs
```diff
@@ -72,20 +72,44 @@ public static bool IsLocalFile(string fileName)
...
+        {
+            if (string.IsNullOrEmpty(contents))
+            {
+                return false;
+            }
+
+            string[] parts = contents.Split(' ');
```
drewnoakes, line 103: Could use `LazyStringSplit` here (I think the PR where I added that was merged) to avoid allocating a `string[]`.

GitCommands/PathUtil.cs
```diff
@@ -72,20 +72,44 @@ public static bool IsLocalFile(string fileName)
...
+        /// <remarks>
+        /// Uri.IsWellFormedUriString is used instead of Uri.TryCreate as file URIs
+        /// of the form X:\\directory\\filename should not be treated as URLs (doing so causes
+        /// LocalRepositoryManager.AddAsMostRecentAsync to throw a NotSupportedException)
+        /// </remarks>
         /// <param name="path">A path to check.</param>
-        /// <returns><see langword="true"/> if the given path starts with 'http', 'ssh' or 'git'; otherwise <see langword="false"/>.</returns>
+        /// <returns><see langword="true"/> if the given path contains a URL; otherwise <see langword="false"/>.</returns>
         [Pure]
         public static bool IsUrl(string path)
```
drewnoakes, line 86: This method doesn't add much value any more. It's basically just `Uri.IsWellFormedUriString`. Do we still need it?

GitCommands/PathUtil.cs
nyankoframe: I don't see LazyStringSplit in the 3.5 branch, but will use it when I replace the original implementation in the main branch.

GitCommands/PathUtil.cs
nyankoframe: Thanks for the feedback; will update in the next iteration.

GitCommands/PathUtil.cs
nyankoframe: I've removed this method and replaced references to it by calling `Uri.IsWellFormedUriString`.

nyankoframe:
I don't understand why the submodule tests are failing in AppVeyor, but pass when I run them locally. :(

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
```diff
@@ -347,9 +333,36 @@ public void DeleteWithExtremePrejudice_should_return_true_if_delete_successful()
...
+        [TestCase("github.com/gitextensions", false)]
+        [TestCase("github.com/gitextensions/gitextensions/pull/9018", false)]
+        [TestCase("http://", false)]
+        [TestCase("HTTPS://www", true)]
+        [TestCase("git://", true)]
+        [TestCase("file:", false)]
+        [TestCase("SSH:", true)]
+        [TestCase("SSH", false)]
+        [TestCase("https://myhost:12368/", true)]
         public void CanBeGitURL(string url, bool expected)
```
gerhardol, line 347: There are no tests for the second tests

```c#
                   || url.EndsWith(".git", StringComparison.CurrentCultureIgnoreCase)
                   || GitModule.IsValidGitWorkingDir(url);
```

gerhardol (approved):
I would like to see the test being updated, but otherwise OK for me

gerhardol:
> I don't understand why the submodule tests are failing in AppVeyor, but pass when I run them locally. :(

Tests require the as much effort as the code...
This is likely not caused by this PR though.
Updated build is OK.

I cannot see what goes wrong, it seems a little like tests run in parallel (which is not possible for these tests).
The test timed out, likely due to a popup in the test sever, that will need to be investigated by logging in to the server before the timeout fires.

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
gerhardol: I believe the comments are addressed

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
nyankoframe: There are some tests for paths/URLs that end with .git but I guess there are no tests for IsValidGitWorkingDir since it needs a specific directory structure.

GitCommands/PathUtil.cs
```diff
@@ -72,20 +72,37 @@ public static bool IsLocalFile(string fileName)
...
+        /// <remarks>
+        /// Uri.IsWellFormedUriString is used instead of Uri.TryCreate as file URIs
+        /// of the form X:\\directory\\filename should not be treated as URLs.
+        /// Each URL extracted from <paramref name="contents"/> is added as an entry in
+        /// <paramref name="urls"/>.
+        /// </remarks>
+        /// <param name="contents">A string to attempt to extract URLs from.</param>
+        /// <param name="urls">A <see cref="List{T}"/> that contains any URLs extracted from <paramref name="contents"/>.</param>
+        /// <returns><see langword="true"/> if the given string contains one or more URLs; otherwise <see langword="false"/>.</returns>
+        public static bool TryExtractURLsFromString(string contents, out List<string> urls)
```
RussKie, line 87: 💭  To clarify - we now have a single callsite for this method in `FormClone`, right? If this is so, I'd turn this method into a private in `FormClone`, and test it through a test accessor. And since the only use case for this method is to use the first detected URL, there are questions why we need to allocate a list, etc.

gerhardol:
Note: 3.5.1 is hopefully shipped soon...

GitCommands/PathUtil.cs
nyankoframe: Thank you for the feedback; I've changed this method to a private in `FormClone` and have added a test for it.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCloneTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            string text, bool expected, string expectedUrl)
+        {
+            RunFormTest(
+                form =>
+                {
+                    var accessor = form.GetTestAccessor();
+
+                    accessor.TryExtractUrl(text, out var url).Should().Equals(expected);
+
+                    url.Should().Equals(expectedUrl);
```
gerhardol, line 65: Should expected URL be tested if `expected` is false?

RussKie (approved):

RussKie:
Thank you!

RussKie:
@nyankoframe can you please send a matching PR to the master branch?

PR #9122, opened 2021-04-26
View changed submodule could hang GE

gerhardol:
Temporary fix for #8914 (not closing #8914)

The proper fix is in #8923, but that PR introduces some test instabilities so the app code is patched instead.

## Proposed changes

Do not hang the application when submodule status is not evaluated when viewing.

## Test methodology 

Manual.
You may have to insert the error by installing Symantec or adding a delay to submodule status calculation.

----

### DIFF
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -115,9 +115,7 @@ namespace GitUI
 
                 if (file.IsSubmodule)
                 {
-                    // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(
-                        () => file.GetSubmoduleStatusAsync() ?? Task.FromResult<GitSubmoduleStatus?>(null));
+                    var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync!);
                     return status is not null
                         ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                         : $"Failed to get status for submodule \"{file.Name}\"";
--- a/GitUI/GitUIExtensions.cs
+++ b/GitUI/GitUIExtensions.cs
@@ -113,11 +113,11 @@ namespace GitUI
                         : diffOfConflict;
                 }
 
-                if (file.IsSubmodule
-                    && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
+                if (file.IsSubmodule)
                 {
                     // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
+                    var status = ThreadHelper.JoinableTaskFactory.Run(
+                        () => file.GetSubmoduleStatusAsync() ?? Task.FromResult<GitSubmoduleStatus?>(null));
                     return status is not null
                         ? LocalizationHelpers.ProcessSubmoduleStatus(fileViewer.Module, status)
                         : $"Failed to get status for submodule \"{file.Name}\"";


GitUI/GitUIExtensions.cs
```diff
@@ -113,11 +113,11 @@ void OpenWithDiffTool()
...
                 }
 
-                if (file.IsSubmodule
-                    && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
+                if (file.IsSubmodule)
                 {
                     // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
+                    var status = ThreadHelper.JoinableTaskFactory.Run(
+                        () => file.GetSubmoduleStatusAsync() ?? Task.FromResult<GitSubmoduleStatus?>(null));
```
mstv, line 120: Perhaps:
```suggestion
                    GitSubmoduleStatus? status = file.GetSubmoduleStatusAsync()?.GetAwaiter().GetResult();
```

BTW: `GetSubmoduleStatusAsync` turned into a property  would avoid the misleading call `()`.

gerhardol: I used what @drewnoakes suggested
This is hopefully a short term fix only
Will test this some time...

drewnoakes (approved):

GitUI/GitUIExtensions.cs
gerhardol: That change hangs GE so I will not change

gerhardol:
@msftbot merge in 24 hours

gerhardol:
The tests timed out similar to #8923 so I reverted to my original workaround that seem to be OK.

gerhardol:
I have rebuilt this on AppVeyor six times now, with #8923 ish solutions the builds failed 2 out of 3 so I believe this solution is OK (it is similar to pre #8881 that introduced the issue).

Note that the check is changed, always entering the if(isSubmodule), not checking for task.
This is how the current codepaths are implemented, the submodule status tasks are always started (with a minor delay), so the task is never null.
If task is not started, the patch will evaluate the task, currently unused.

The proper change should restore the assumption that the task is not evaluated though.

GitUI/GitUIExtensions.cs
```diff
@@ -113,11 +113,9 @@ void OpenWithDiffTool()
                         : diffOfConflict;
                 }
 
-                if (file.IsSubmodule
-                    && file.GetSubmoduleStatusAsync() is Task<GitSubmoduleStatus> task)
+                if (file.IsSubmodule)
                 {
-                    // Patch already evaluated
-                    var status = ThreadHelper.JoinableTaskFactory.Run(() => task);
+                    var status = ThreadHelper.JoinableTaskFactory.Run(file.GetSubmoduleStatusAsync!);
```
mstv, line 118: 1. Rant: The syntax of `!` is weird in this case. It applies to the _return value_ of the function `file.GetSubmoduleStatusAsync`, which is _not called_ here - but later by `Run`.
2. Are we sure that `file._submoduleStatus` is never `null`?

--- 

3. Does `JoinAsync()` always return the same `Task` instance? `GetSubmoduleStatusAsync()` is called pretty often.

mstv (commented):
Looks better.
Since this will be reworked anyway, we could continue the discussion after merging.

GitUI/GitUIExtensions.cs
gerhardol: 1. This is compiler info to ignore that the value may be null, it does not matter that function is not called
2. GetSubmoduleStatus is _currently_ always set. Anyway, this is a fix for the current solution that hangs if status is not yet ready
3. Do not know, I assume so

gerhardol:
Annoying that the branch was merged also when auto merge was removed, a fixup! was merged. (block-fixup should be enough to stop this)
But it was just review and squash that was missing

RussKie:
I'll check with the msftbot team.

PR #9132, opened 2021-04-28
Sanitize localized stack trace in exception tests

mstv:
## Proposed changes

- Use fixed culture `en-US` in `SerializableExceptionTests`
- Simplify regex

### Before

image

### After

image

## Test methodology 

- adapt NUnit test

## Test environment(s) 

- Git Extensions 33.33.33
- Build a294965deda0a16bb72cf31556c9cbc1bad20362
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/UnitTests/BugReporter.Tests/SerializableExceptionTests.cs
+++ b/UnitTests/BugReporter.Tests/SerializableExceptionTests.cs
@@ -12,6 +12,8 @@ using NUnit.Framework;
 namespace BugReporterTests
 {
     [TestFixture]
+    [SetCulture("en-US")]
+    [SetUICulture("en-US")]
     public sealed class SerializableExceptionTests
     {
         [MethodImpl(MethodImplOptions.NoInlining)]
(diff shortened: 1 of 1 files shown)


gerhardol (approved):
I would like to have this commented, explaining the three manipulations that are done.
It feels like these hardcoded assumptions has a risk of failing in the future (there may be languages that use two words for "at", right-to-left etc).

Draft?

Adjust the log for languages
1. Replace Inner exception divider
2. Replace the stack trace preposition
3. Remove file path (to make compare results)

mstv:
I have added the comments and noticed that the previous regex contained the preposition "in".

gerhardol (approved):

RussKie:
I want to look at it.

RussKie:
Did you try applying the following to the fixture?
```cs
    [SetCulture("en-US")]
    [SetUICulture("en-US")]
```

mstv:
> Did you try applying the following to the fixture?
> 
> ```cs
>     [SetCulture("en-US")]
>     [SetUICulture("en-US")]
> ```

Thank you, it works.
I kept the - IMO simplified - regex.

RussKie:
We set culture for locale sensitive tests. We ought to do the same here,
and avoid custom logic.

mstv:
> We set culture for locale sensitive tests. We ought to do the same here, and avoid custom logic.

I did so yesterday.
In addition, I simplified the existing replacement code.

RussKie (approved):

RussKie:
I didn't quite grasp the response. Should've opened the diff.

Thank you

PR #9140, opened 2021-05-02
feat: add dot command to GitUICommands

MrJithil:
Fixes #

## Proposed changes

- dot command is widespread in most applications to open the current working directory. I have added a `private` function to handle the `.` command in this pull request. For Eg: If a user wants to open their current repository in GitExtensions, he should trigger the command `GitExtensions .`

## Screenshots 
DotCommand

### Before
A user needs to enter different commands to open his current repository.
`GitExtensions browse .` is an example.

### After
By these changes, a user can open the repository in GitExtensions by just triggering the command `GitExtensions .`

### TODO
Need documentation for this command once this PR got sufficient approvals. 

## Test methodology 

- I have tested the changes by modifying the project launch setting property.
- Also, to ensure functionality, I have run the executables from a different git repository.
- Both negative and positive cases were tested.
- Since the added function is a private one, we don't need any unit tests.
- I captured the evidence from my local machine in this pull request.

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCommandlineHelp.resx
+++ b/GitUI/CommandsDialogs/FormCommandlineHelp.resx
@@ -118,7 +118,8 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="_NO_TRANSLATE_commands.Text" xml:space="preserve">
-    <value>browse [path] [-filter=] [-commit=&lt;selectedSha&gt;[,&lt;firstSha&gt;]]
+    <value>[path]
+browse [path] [-filter=] [-commit=&lt;selectedSha&gt;[,&lt;firstSha&gt;]]
 about
 add [filename]
 addfiles [filename]
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1499,6 +1499,14 @@ namespace GitUI
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied a path. Open the repository if its a valid path
+                    string dir = !string.IsNullOrWhiteSpace(command) && File.Exists(command) ? Path.GetDirectoryName(command) : command;
+                    if (args.Count == 2 && Directory.Exists(dir))
+                    {
+                        LaunchBrowse(dir);
+                        return true;
+                    }
+
                     break;
             }
 #pragma warning restore SA1025 // Code should not contain multiple whitespace in a row


GitUI/GitUICommands.cs
```diff
@@ -1471,6 +1471,8 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                     return StartViewPatchDialog(args.Count == 3 ? args[2] : "");
                 case "uninstall":
                     return UninstallEditor();
+                case ".":
```
gerhardol, line 1474: see comment on line 1383: Update FormCommandlineHelp too

gerhardol, line 1474: Should this be handled in default:?
A specific case for handling paths.

gerhardol (commented):
seem ok
The normal way is to use "browse <path>" or "gitex.cmd <path>"

GitUI/GitUICommands.cs
MrJithil: Yes, I have added that. Thanks for the knowledge share.

MrJithil: Sure. Thanks for the great suggestion.
I have added that.

MrJithil:
@gerhardol I have added the suggested changes. Please review.

GitUI/GitUICommands.cs
```diff
@@ -1914,5 +1919,10 @@ internal TestAccessor(GitUICommands commands)
 
             internal bool RunCommandBasedOnArgument(string[] args) => _commands.RunCommandBasedOnArgument(args, InitializeArguments(args));
         }
+
+        private bool RunDotCommand()
+        {
+            return StartBrowseDialog();
+        }
```
mstv, line 1926: Please remove. I do not see any benefit of adding an alias. Furthermore "." is rather a path than a command.

mstv (commented):
> The normal way is to use "browse " or "gitex.cmd "

So if we start to support ".", why not any path?
I don't know the rationale why the usage of `browse` was forced when the CLI was designed.

GitUI/GitUICommands.cs
mstv: The `default:` case is dealing with clone URLs.
Without a comment, the added code will be hard to understand next time we will read it - basically, i.e. not only if not handled using `case ".":`.

GitUI/GitUICommands.cs
gerhardol: Agree. No alias, add a comment that this is interpreted as a path
Also add a test, similar to RunCommandBasedOnArgument_browse()

GitUI/GitUICommands.cs
MrJithil: I have added the comment as discussed. The changes will be reflected in the next push.

GitUI/GitUICommands.cs
MrJithil: Sure. I have added the test case `RunCommandBasedOnArgument_dot()`.
Thank you much for the sample reference.
The changes will be in the next pull.
CC: @mstv @gerhardol

gerhardol (approved):

IntegrationTests/UI.IntegrationTests/GitUICommands/RunCommandTests.cs
```diff
@@ -363,5 +363,14 @@ private static void ClickButton(Form form, string buttonName)
 
         private static void SetText(Form form, string controlName, string value)
             => form.FindDescendantOfType<Control>(control => control.Name == controlName).Text = value;
+
+        [Test]
+        public void RunCommandBasedOnArgument_dot()
+        {
+            var selected = ObjectId.Random();
+            var first = ObjectId.Random();
+            var otherIgnored = ObjectId.Random();
```
mstv, line 372: What is the purpose of those unused instances?

GitUI/GitUICommands.cs
```diff
@@ -1487,6 +1487,12 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied `.` as param. So, we need to open the current working directory.
+                    if (args[1] == ".")
+                    {
+                        return StartBrowseDialog();
+                    }
```
mstv, line 1494: If we start to support ".", why not any path? Thoughts, @gitextensions/git-extensions-team?

Then `Program.GetWorkingDir` should try `args[1]`, too. Currently, it is hard-coded to `args[2]`.

mstv (commented):
@MrJithil, file another PR with your sign off in `contributors.txt`, please.

IntegrationTests/UI.IntegrationTests/GitUICommands/RunCommandTests.cs
MrJithil: Thought to extend the unit tests. But, since it's handled with a function, removed it.

MrJithil: Removed in the latest push.

MrJithil:
> @MrJithil, file another PR with your sign off in `contributors.txt`, please.

I have created #9148 as my sign-off. Thanks @mstv

GitUI/GitUICommands.cs
gerhardol: Why not?

RussKie: I'm with @mstv - "." denotes a path, so its handling should be generic to any path.

MrJithil: So, do we need to create a separate PR for that? Or do I need to make changes to this one?
CC: @mstv

MrJithil: Anyway, I started working on it.

MrJithil: Implemented it.

MrJithil:
@mstv @gerhardol @RussKie I have implemented the suggestion to take care `path` along with the `.` param.

Please see the evidence here. Kindly review, approve and merge if you are okay to proceed. Please feel free to revert back in case of any additional suggestions.

New 

Evidence:


https://user-images.githubusercontent.com/26359740/117197332-52135400-ae05-11eb-9c31-904b6c76ead9.mp4

GitUI/CommandsDialogs/FormCommandlineHelp.resx
```diff
@@ -118,7 +118,9 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="_NO_TRANSLATE_commands.Text" xml:space="preserve">
-    <value>browse [path] [-filter=] [-commit=&lt;selectedSha&gt;[,&lt;firstSha&gt;]]
+    <value>.
```
gerhardol, line 121: remove .

GitUI/GitUICommands.cs
```diff
@@ -1487,6 +1487,12 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied a path. Open the repository if its a valid path
+                    if (Directory.Exists(args[1]))
+                    {
+                        return RunOpenRepoFromPath(args[1]);
```
gerhardol, line 1493: Suggest reducing the code (untested)
```suggestion
                        List<string> args2 = new() { "browse" };
                        args2.AddRange(args);
                        return RunBrowseCommand(args2);
```

gerhardol (commented):
+1

GitUI/CommandsDialogs/FormCommandlineHelp.resx
mstv: Or rather turn into
```suggestion
    <value>[path]
```
?

GitUI/GitUICommands.cs
mstv: I am afraid `RunBrowseCommand` will not re-evaluate the path because it has already been checked and fed to the `Module`.

`RunOpenRepoFromPath` is less flexible than the existing logic in `Program.GetWorkingDir`, which should be re-used definitely.
I suggest to just call `LaunchBrowse(args[1])` - also if `File.Exists(args[1])`.
This has the drawback that a new instance is started. But  this should not be too expensive nowadays.

MrJithil: We may not do this approach for reducing the code. Because `browse` or `openrepo` commands are trying to open a repository that is set during the instantiation (`GitModule`). 
So, instead of proceeding with a workaround, my opinion is to treat it as a separate function as now.

Please suggest if you have disagreements.

GitUI/CommandsDialogs/FormCommandlineHelp.resx
MrJithil: Removed.

GitUI/GitUICommands.cs
mstv: > So, instead of proceeding with a workaround, my opinion is to treat it as a separate function as now.

I disagree to duplicating or - even worse - creating slightly different logic.

You can ignore the returned `IProcess` instance (as all other callers do, too). The new GE instance shall run detached.
Just `return true` in order to exit the first GE instance.

gerhardol: mstv suggestion avoids the additional method
(I will not force checking if args[1] is a file to use the directory).
```suggestion
LaunchBrowse(args[1]);
```

MrJithil: Sure. I have added those changes. I kept the Directory Exisist checks there, to not break the code and avoid the malformed commands. 

@mstv @gerhardol Please review.

RussKie (approved):
Works for me 👍

GitUI/GitUICommands.cs
```diff
@@ -1487,6 +1487,13 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied a path. Open the repository if its a valid path
+                    if (Directory.Exists(args[1]))
```
mstv, line 1491: Browse supports files, too. Do you have an example where we would not see a relevant error message when using:
```suggestion
                    if (args.Length == 2 && (Directory.Exists(args[1]) || File.Exists(args[1])))
```
?

MrJithil: `LaunchBrowse` only supports File. So, get the directory if it's a path, then open it up.

MrJithil: Added

GitUI/GitUICommands.cs
```diff
@@ -1487,6 +1487,14 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied a path. Open the repository if its a valid path
+                    string dir = !string.IsNullOrWhiteSpace(args[1]) && File.Exists(args[1]) ? Path.GetDirectoryName(args[1]) : args[1];
+                    if (args.Count == 2 && Directory.Exists(dir))
+                    {
+                        LaunchBrowse(dir);
```
gerhardol, line 1494: Exception if no argument...
untested

```suggestion
                    string dir = args.Count != 2 
                        ? ""
                        :  File.Exists(args[1]) ? Path.GetDirectoryName(args[1]) : args[1];
                    if (Directory.Exists(dir))
                    {
                        LaunchBrowse(dir);
```

GitUI/GitUICommands.cs
mstv: > `LaunchBrowse` only supports File. So, get the directory if it's a path, then open it up.

Why do you think this? It supports the repo dir as well as any dir or file below. Just play around with `GitExtensions.exe browse <path>`. It opens the repo. The tail is ignored.

mstv: Ah, I forgot about the (arguable) design of `.Exists`. So exceptions from the `.Exists` functions must be silently caught.

MrJithil: The same changes already added in the MR. please test it if necessary

MrJithil: And, I request the approval as the earliest since its a been a while.

mstv: > Exception if no argument...

as @gerhardol wrote.
Avoid too much logic, just ignore the exceptions but catch them.

GitUI/GitUICommands.cs
MrJithil: I don't see any exception if I am passing no args. Could you please give evidence?

Since we are `string.IsNullOrWhiteSpace(args[1])` invalid cases will be ignored.

GitUI/GitUICommands.cs
MrJithil: We will not get any exceptions. Also, I just add only one line code to avoid a try and catch. It will not even affect the complete flow.
Do you think, the try-catch is necessary here? If yes, do we need an extra `function` to handle it? Because the if condition will be exhausted with too many lines.

MrJithil: @mstv

GitUI/GitUICommands.cs
```diff
@@ -1487,6 +1487,14 @@ private bool RunCommandBasedOnArgument(IReadOnlyList<string> args, IReadOnlyDict
                         return StartCloneDialog(null, args[1].Replace("github-mac://openRepo/", ""), true);
                     }
 
+                    // User supplied a path. Open the repository if its a valid path
+                    string dir = !string.IsNullOrWhiteSpace(args[1]) && File.Exists(args[1]) ? Path.GetDirectoryName(args[1]) : args[1];
```
RussKie, line 1491: [nit] put `args[1]` into a variable, and use it across.

MrJithil: We discussed reducing the code before. Now, are we okay to add more lines?
Please confirm.

MrJithil: I appreciate it. Instead of creating a variable for `args[1]`, I reused the variable `command` here.

RussKie:
Please [squash](https://github.com/gitextensions/gitextensions/wiki/How-To%3A-Squash-and-Rebase-your-changes) and it is ready to go.

MrJithil:
> Please [squash](https://github.com/gitextensions/gitextensions/wiki/How-To%3A-Squash-and-Rebase-your-changes) and it is ready to go.

Rebased the branch with `master`. @RussKie Please proceed.

RussKie:
Thank you

PR #9141, opened 2021-05-02
`FileViewer`: Remove decoration of default encoding

mstv:
Fixes 1. of https://github.com/gitextensions/gitextensions/pull/8522#issuecomment-829594076 - unrelated to .NET 5.0

## Proposed changes

`FileViewer` toolbar: Remove the decoration of default `Encoding`
because of
- incomplete implementation: ComboBox empty due to not matching name
- inconsistent `Encoding` naming: `EncodingName` vs. `HeaderName`

### Before

image

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7213de011d4ae508fc819473c68f27a6c1f702e8
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -709,26 +709,9 @@ namespace GitUI.Editor
             ReloadHotkeys();
             Font = AppSettings.FixedWidthFont;
 
-            DetectDefaultEncoding();
-            return;
-
-            void DetectDefaultEncoding()
-            {
-                var encodings = AppSettings.AvailableEncodings.Values.Select(e => e.EncodingName).ToArray();
-                encodingToolStripComboBox.Items.AddRange(encodings);
-                encodingToolStripComboBox.ResizeDropDownWidth(50, 250);
-
-                var defaultEncodingName = Encoding.Default.EncodingName;
-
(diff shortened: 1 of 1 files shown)


RussKie (approved):

RussKie:
Do we need to apply the same in the Settings dialog?

mstv:
> Do we need to apply the same in the Settings dialog?

The default encoding is handled more straight-forward in the Settings dialog - it seems to be complete and does only affect the settings file(s).

With .NET 5.0, the users must reconfigure their preferred encodings.
.NET 5.0 supports much less of them and has a different default:

image

GE 3.5:

image

gerhardol:
> With .NET 5.0, the users must reconfigure their preferred encodings.
> .NET 5.0 supports much less of them and has a different default:

Should migration set something (Unicode UTF-8 works fine in Western languages) if no value available in .NET5?

gerhardol (commented):
+1
Seem ok and works for me
I am not sure how the default encoding should be handled. There are some handling of default encoding for Git (read at startup) and for a GitModule (a separate Git command to find Git dir to find encoding required for many commands).
But this change should only remove the "Default()" prefix.

Issue #9126, opened 2021-04-26
Dialog placement does not consider the Taskbar


mstv:
## Current behaviour

Not fitting GE dialogs are moved to `Left = 0` although the area is occupied by the - not auto-hiding - Windows Taskbar.

## Expected behaviour

Dialogs are placed in the free working area.

This is a minor issue for me. Just noticed it when testing the installer in a VM.
I expected that `Screen.WorkingArea` would cut out the Taskbar.

## Steps to reproduce

- move the Windows Taskbar to the left
- move the GE dashboard near the top of desktop
- open FormSettings

## Screenshots

![image](https://user-images.githubusercontent.com/36601201/116149947-08c75400-a6e3-11eb-8d69-8d34cab78887.png)

## Did this work in previous version of GitExtensions

No, it works better than in earlier versions.

## Environment

- Git Extensions 33.33.33
- Build 6d07a7312a5b84b81f7ae1ff25e505c2fdd7b31c
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

PR #9143, opened 2021-05-03
fix: position setting window always in the window center screen

MrJithil:
Fixes #
Settings window popup is not positioned correctly while GitExtension not in the center of the window or in maximized state. 
This issue is stated in issue: #9126 
## Proposed changes

- Always start the settings window pop-up from the center of the screen. 

### Before
The settings window will start from the center of the GitExtension. So, if the GitExtension is in a corner ( especially on the left side of the window) a partial view of the settings pop-up will appear. The example could be found below: 
GitExtSettingsWindowsPosIssue

### After
The settings window will start from the center of the window. It will ensure, the settings pop-up will be displayed correctly to the user regardless of the GitExtension position.

Evidence:

GitExtSettingsWindowsPosIssueFixed

## Test methodology 

- Positioned the GitExtension in different parts of the window and opened the settings dialogue. Verified the settings pop-up is correctly displayed.

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/CommandsDialogs/FormSettings.Designer.cs
+++ b/GitUI/CommandsDialogs/FormSettings.Designer.cs
@@ -219,7 +219,7 @@ namespace GitUI.CommandsDialogs
             this.MinimumSize = new System.Drawing.Size(966, 785);
             this.Name = "FormSettings";
             this.SizeGripStyle = System.Windows.Forms.SizeGripStyle.Show;
-            this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
+            this.StartPosition = System.Windows.Forms.FormStartPosition.CenterScreen;
             this.Text = "Settings";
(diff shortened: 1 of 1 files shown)


mstv (commented):
A simple and working solution.
Though I am afraid, users with high-res displays will not like the dialog centered on the screen.

We have a positioning logic, which is aware of multi-screen setups.
I would rather extend this in order to take the left border of the working area into account correctly.

RussKie:
Thank you, but the current proposal will not work.
The fix will need to be an addendum to https://github.com/gitextensions/gitextensions/pull/8341 and account for the taskbar position (whichever side it is docked to).

RussKie (changes requested):
Thank you, but the current proposal will not work.
The fix will need to be an addendum to #8341 and account for the taskbar position (whichever side it is docked to).

PR #8303, opened 2020-07-08
Rework FormStatus, FormProgress and FormRemoteProgress

RussKie:
- Move FormStatus, FormProgress and FormRemoteProgress into GitUI.HelperDialogs
- Fix inheritance chain for the above forms
- Consolidate and remove multiple `ShowDialog` overloads to simplify the API and get a coherent view on the usage
  * `FormStatus` us used 3 times directly (via .ctor)
  * `FormProcess` is used once directly (via .ctor), 66 times indirectly (61 via `ShowDialog`, 5 times via `ReadDialog`)
  * `FormRemoteProcess` is used 13 times directly (via .ctor), and 1 indirectly
 

This is an intermediate step to unblock further work in https://github.com/gitextensions/gitextensions/pull/8292

PR #9145, opened 2021-05-03
Stash pop/apply: Refresh grid if command was successful

### LINKED ISSUE #9121: Stash pop is not refreshing the window
## Current behaviour

The window is not being refresh after a stash pop operation and so still shows the previous stashed code, even when it is not there anymore.

## Expected behaviour

The window should be refreshed after a stash pop operation

## Steps to reproduce

1. Stash some code
2. Observe that the window shows the stashed code with the other commits
3. Do a stash pop directly via menu option on toolbar
4. Observe commits in window

Yes, this was working fine just on previous stable version to current 3.5.0.11713

## Environment

- Git Extensions 3.5.0.11713
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 192dpi (200% scaling)

gerhardol:
Fixes #9121

## Proposed changes

Refresh revision grid after Stash pop.
Also changed after stash-apply, but that makes no practical difference as the command is only invoked from FormStash (and not update the grid...)

Regression, to be cherry-picked for 3.5

## Test methodology 

See the issue. Stash, stash-pop and see that the grid refreshes.

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -191,7 +191,10 @@ namespace GitUI
             bool Action()
             {
                 var arguments = GitCommandHelpers.StashSaveCmd(includeUntrackedFiles, keepIndex, message, selectedFiles);
-                return FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+
+                // git-stash may have changed commits also if aborted, the grid must be refreshed
+                return true;
             }
 
             return DoActionOnRepo(owner, Action);
@@ -201,8 +204,11 @@ namespace GitUI
         {
             bool Action()
             {
-                bool success = FormProcess.ShowDialog(owner, process: null, arguments: "stash pop", Module.WorkingDir, input: null, useDialogSettings: true);
-                return success && MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
+                FormProcess.ShowDialog(owner, process: null, arguments: "stash pop", Module.WorkingDir, input: null, useDialogSettings: true);
+                MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
+
+                // git-stash may have changed commits also if aborted, the grid must be refreshed
+                return true;
             }
 
             return DoActionOnRepo(owner, Action);
@@ -212,7 +218,10 @@ namespace GitUI
         {
             bool Action()
             {
-                return FormProcess.ShowDialog(owner, process: null, arguments: $"stash drop {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(owner, process: null, arguments: $"stash drop {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+
+                // git-stash may have changed commits also if aborted, the grid must be refreshed
+                return true;
             }
 
             return DoActionOnRepo(owner, Action);
@@ -222,8 +231,11 @@ namespace GitUI
         {
             bool Action()
             {
-                bool success = FormProcess.ShowDialog(owner, process: null, arguments: $"stash apply {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
-                return success && MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
+                FormProcess.ShowDialog(owner, process: null, arguments: $"stash apply {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+                MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
+
+                // git-stash may have changed commits also if aborted, the grid must be refreshed
+                return true;
             }
 
             return DoActionOnRepo(owner, Action);


ivangrek:
Why update if merge conflicts are false?

gerhardol:
> Why update if merge conflicts are false?

The stash commit is popped, no longer visible in the grid.

GitUI/GitUICommands.cs
```diff
@@ -202,7 +202,12 @@ public bool StashPop(IWin32Window? owner)
             bool Action()
             {
                 bool success = FormProcess.ShowDialog(owner, process: null, arguments: "stash pop", Module.WorkingDir, input: null, useDialogSettings: true);
-                return success && MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
+                if (success)
+                {
+                    MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
```
RussKie, line 207: So what happens if conflicts aren't resolved?

gerhardol: The merge conflict banner is shown similar to if a patch cannot be applied, rebase conflicts etc.
The grid needs to be updated regardless

The old handling (always run HandleMergeConflicts(), return true to always refresh the grid) was safer, I believe this is OK though.

mstv: I must have missed these two when approving #8303. Since we cannot know how much git had done before the error, we should always request a refresh, i.e. always `return true`, please.

mstv (commented):
Applies to `StashSave`, `StashDrop`, too.

gerhardol:
> I must have missed these two when approving #8303. Since we cannot know how much git had done before the error, we should always request a refresh, i.e. always return true, please.

I choose to make the update for the likely change, but if the other situations already makes the change unconditionally I will revert the behavior to prior #8303

GitUI/GitUICommands.cs
```diff
@@ -227,13 +228,10 @@ public bool StashApply(IWin32Window? owner, string stashName)
         {
             bool Action()
             {
-                bool success = FormProcess.ShowDialog(owner, process: null, arguments: $"stash apply {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
-                if (success)
-                {
-                    MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
-                }
+                FormProcess.ShowDialog(owner, process: null, arguments: $"stash apply {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+                MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
```
RussKie, line 232: Please add a comment explaining why we can't chain the operations like we did.

GitUI/GitUICommands.cs
```diff
@@ -217,7 +216,9 @@ public bool StashDrop(IWin32Window? owner, string stashName)
         {
             bool Action()
             {
-                return FormProcess.ShowDialog(owner, process: null, arguments: $"stash drop {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(owner, process: null, arguments: $"stash drop {stashName.Quote()}", Module.WorkingDir, input: null, useDialogSettings: true);
+
+                return true;
```
RussKie, line 221: ditto here

GitUI/GitUICommands.cs
```diff
@@ -201,13 +203,10 @@ public bool StashPop(IWin32Window? owner)
...
-                if (success)
-                {
-                    MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
-                }
+                FormProcess.ShowDialog(owner, process: null, arguments: "stash pop", Module.WorkingDir, input: null, useDialogSettings: true);
+                MergeConflictHandler.HandleMergeConflicts(this, owner, false, false);
 
-                return success;
+                return true;
             }
```
RussKie: ditto

GitUI/GitUICommands.cs
```diff
@@ -191,7 +191,9 @@ public bool StashSave(IWin32Window? owner, bool includeUntrackedFiles, bool keep
             bool Action()
             {
                 var arguments = GitCommandHelpers.StashSaveCmd(includeUntrackedFiles, keepIndex, message, selectedFiles);
-                return FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+
+                return true;
```
RussKie, line 196: ditto

RussKie (approved):

GitUI/GitUICommands.cs
```diff
@@ -191,7 +191,10 @@ public bool StashSave(IWin32Window? owner, bool includeUntrackedFiles, bool keep
             bool Action()
             {
                 var arguments = GitCommandHelpers.StashSaveCmd(includeUntrackedFiles, keepIndex, message, selectedFiles);
-                return FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+                FormProcess.ShowDialog(owner, process: null, arguments, Module.WorkingDir, input: null, useDialogSettings: true);
+
+                // git-stash may have changed commits also if aborted, the grid must be refereshed
```
mstv, line 196:
```suggestion
                // git-stash may have changed commits also if aborted, the grid must be refreshed
```

mstv (approved):
Would you fix the typo in all comments?

gerhardol:
@msftbot merge in 24 hours

Issue #9112, opened 2021-04-25
SSH broken after upgrade to 3.5.0 (due to breaking setting change)


jrfnl:
## Current behaviour

Any repo for which the remote is set as `git@github.com:....` can no longer communicate with remotes.

## Expected behaviour

That essential settings do not change value in a functionality breaking way when updating to a new minor.

## Steps to reproduce

1. Have GitExtensions 3.4.3
2. In Settings -> SSH -> make sure it is set to PuTTY
3. Have a repo set up with a `git@github.com:...` remote URL and a link to the Private key saved in the settings for the remote repository.
4. Upgrade to GitExtensions 3.5.0.
5. Try to `Fetch all` for the repo and see the below error message.
6. Go to Settings -> SSH and conclude that it is no longer set to PuTTY, but to the `ssh` command in the system `git` directory.
7. Change the setting back to PuTTY to get things working again.

## Screenshots

![image](https://user-images.githubusercontent.com/663378/115997983-71300b80-a5e5-11eb-8907-8930ef90611f.png)

## Did this work in previous version of GitExtensions

Yup, worked fine in 3.4.3.

## Environment

- Git Extensions 3.5.0.11713
- Build 0a5ef9ca6ce87754a510aa02c7a96d8b915714ac (Dirty)
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4341.0
- DPI 144dpi (150% scaling)

gerhardol:
This should be changed in 3.5 (as well as master)

gerhardol:
I saw the same problem when running the installer. Cannot recreate the problem though.

I did not check the contents in the registry key where ssh setting is stored: Computer\HKEY_CURRENT_USER\SOFTWARE\GitExtensions\gitssh
The path after running the installer is changed to "C:\Program Files (x86)\Git\usr\bin\ssh.exe" 
The Git installation is x64, no 32 bit Git installed.
So the SSH path must be adjusted after running the installer.

Note that the presentation in settings is incorrect, see #9149
OpenSSH and other are mostly handled the same. Before som refactoring, OpenSSH was related to the Git installation dir and always empty, now it is fixed to the Git installation when updating the settings. If you mess with Git installation paths, you have to change OpenSSH too. Should be an OK limitation.

The OpenSSH handling in the installer does not seem to be changed in 3.5

jrfnl:
Thanks for looking into this @gerhardol !

[...]

PR #9149, opened 2021-05-04
OpenSSH presented as OtherSSH in settings

gerhardol:
Part of #9112

## Proposed changes

Some SSH refactoring has changed the SSH setting to not always be stored as "" (stored in registry).
The OpenSSH setting is therefore always presented as other.

This is part of what described in #9112, but not the core problem.

### Before

OpenSSH is not set, but the path is correct

image

### After

image

## Test methodology 

manual

---

### DIFF
--- a/GitCommands/Git/GitSshHelpers.cs
+++ b/GitCommands/Git/GitSshHelpers.cs
@@ -25,19 +25,11 @@ namespace GitCommands
             }
         }
 
-        /// <summary>Un-sets the git SSH command path.</summary>
-        public static void UnsetSsh()
-        {
-            Environment.SetEnvironmentVariable("GIT_SSH", "", EnvironmentVariableTarget.Process);
-        }
-
         /// <summary>Sets the git SSH command path.</summary>
         public static void SetSsh(string? path)
         {
-            if (!string.IsNullOrEmpty(path))
-            {
-                Environment.SetEnvironmentVariable("GIT_SSH", path, EnvironmentVariableTarget.Process);
-            }
+            // Git will use the embedded OpenSSH ssh.exe if empty/unset
+            Environment.SetEnvironmentVariable("GIT_SSH", path ?? "", EnvironmentVariableTarget.Process);
         }
 
         public static bool Plink()
--- a/GitCommands/SshPathLocator.cs
+++ b/GitCommands/SshPathLocator.cs
@@ -8,6 +8,7 @@ namespace GitCommands
     public interface ISshPathLocator
     {
         string Find(string gitBinDirectory);
+        public string? GetSshFromGitDir(string gitBinDirectory);
     }
 
     public sealed class SshPathLocator : ISshPathLocator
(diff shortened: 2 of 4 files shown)


GitCommands/SshPathLocator.cs
```diff
@@ -68,5 +74,13 @@ public string Find(string gitBinDirectory)
                 return null;
             }
         }
+
+        /// <summary>
+        /// Check if a path is a known OpenSSH executable
```
RussKie, line 79: The implementation is quite different from what the doc states.
```suggestion
        /// Checks if OpenSSH is accessible via %PATH%, or comes from the git's bin folder.
```

gerhardol:
I will change this to an `out` parameter in Find() instead.

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -447,7 +447,8 @@ private bool CheckSSHSettings()
             }
 
             var ssh = _sshPathLocator.Find(AppSettings.GitBinDir);
-            RenderSettingSet(SshConfig, SshConfig_Fix, string.IsNullOrEmpty(ssh) ? _opensshUsed.Text : string.Format(_unknownSshClient.Text, ssh));
+            var sshPath = _sshPathLocator.Find(AppSettings.GitBinDir);
```
RussKie, line 450: `_sshPathLocator.Find` return a valid path or an empty string, if SSH is not found. Why do we need to detect it the second time?

RussKie (changes requested):

RussKie:
> Some SSH refactoring has changed the SSH setting to not always be stored as "" (stored in registry).

I think we need to fix this. Are you able to provide more information?

gerhardol:
> > Some SSH refactoring has changed the SSH setting to not always be stored as "" (stored in registry).
> 
> I think we need to fix this. Are you able to provide more information?

If registry key (gitssh) is empty, the GIR_SSH environment variable is not set and Git will use the bundled ssh.exe.
If the configured path is empty, OpenSSH is used.

The refactoring will however try to set a value in the registry and within the application.
ssh works mostly fine anyway, a few annoyances:
* The setting always show "Other", never OpenSSH (the path is OK though)
* If you switch Git path, the ssh path is stuck on the old Git path

This was a fix for presentation primarily, to include in 3.5 and to simplify the more severe issue with SSH path being modified incorrectly when installing 3.5.

GitCommands/SshPathLocator.cs
```diff
@@ -43,6 +44,11 @@ public string Find(string gitBinDirectory)
             return ssh ?? "";
         }
 
+        /// <summary>
+        /// Get ssh path from Git installation
+        /// </summary>
+        /// <param name="gitBinDirectory">Git installation directory</param>
+        /// <returns>Path to ssh.exe or null</returns>
```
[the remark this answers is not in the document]
gerhardol: will update

GitCommands/SshPathLocator.cs
gerhardol: Will remove this method, add an out parameter to Find() instead to inform  if this is an OpenSSH - or we go the full way and set it to empty if matching OpenSSH in the Git bindir. (The last alternative may be OK, I was not sure when I did this initially.)

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
gerhardol: To find if this is OpenSSH or not.
This will be changed (see above), probably returning empty for OpenSSH.

GitCommands/SshPathLocator.cs
RussKie: That seems like a better solution.

gerhardol:
Changed back to use "" for OpenSSH. As most original changes were reverted, I squashed on master.
The original commit for SSH Locator mentioned that the change was done to always set GIT_SSH env var, not properly done (variable unset when saving settings). A second commit corrects that. As the env var is set regardless, there is no change for the binary.

GitCommands/Git/GitSshHelpers.cs
```diff
@@ -26,19 +26,16 @@ bool DoArgumentsRequireSsh()
...
-
         /// <summary>Sets the git SSH command path.</summary>
         public static void SetSsh(string? path)
         {
-            if (!Strings.IsNullOrEmpty(path))
+            if (Strings.IsNullOrEmpty(path))
             {
-                Environment.SetEnvironmentVariable("GIT_SSH", path, EnvironmentVariableTarget.Process);
+                // GIT_SSH is not required by Git to use OpenSSH, but supposedly by Gerrit plugin
+                path = _sshPathLocatorInstance.GetSshFromGitDir(AppSettings.GitBinDir) ?? "";
```
RussKie, line 35: 🤔 If this is required by Gerrit, should this be in Gerrit plugin?

RussKie (approved):

GitCommands/Git/GitSshHelpers.cs
gerhardol: Maybe
It does not seem like Gerrit uses GIT_SSH directly but relies on SshPathLocator.Find().
It could probably be OK to unset the variable if empty instead.
Pushing an update.

Gerrit maybe needs to implement SshPathLocator.GetSshFromGitDir() too, not sure how an empty path is handled. (There is code, not sure if that is working.)
The plugin imports GitUI so it need to be rebuilt on most GE changes anyway.

mstv (approved):
LGTM, have not run

RussKie:
Rebase and merge

PR #9167, opened 2021-05-06
Add `BugReporter.exe` to installer

mstv:
Related to #9166

## Proposed changes

- Add `BugReporter.exe` to installer

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build fec693532c9cc0e23e0e91c841b9a343c64f86c2
- Git 2.31.1.windows.1
- .NET Framework 4.7.2633.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/Setup/Product.wxs
+++ b/Setup/Product.wxs
@@ -125,6 +125,9 @@
       <Component Id="GitExtensions.exe.config" Guid="*">
         <File Source="$(var.ArtifactsPublishPath)\GitExtensions.exe.config" />
       </Component>
+      <Component Id="BugReporter.exe" Guid="*">
+        <File Source="$(var.ArtifactsPublishPath)\BugReporter.exe" />
+      </Component>
       <Component Id="System.IO.Abstractions.dll" Guid="*">
         <File Source="$(var.ArtifactsPublishPath)\System.IO.Abstractions.dll" />
       </Component>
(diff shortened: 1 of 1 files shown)


gerhardol (approved):
So this will be cherrypicked to master?

This explains why there are so few NBug for 3.5...

mstv:
> So this will be cherrypicked to master?

IIUC, the installer will be replaced for .NET 5.0.
BTW: Is there a shorter abbreviation which is easier to type?

> This explains why there are so few NBug for 3.5...

Only uncaught exceptions from background threads are affected.

RussKie:
OMG facepalm.... 
Thank you!

RussKie:
I don't believe there is a config.... Is there?

mstv:
Juliette mentioned `BugReporter.exe.config`, and yes, it is copied to several artifacts folders.

mstv:
@RussKie:
1. There is created a `BugReporter.exe.config`. In my test install, it works without. Should I add it to the 3.5 installer yet?
2. For .NET 5.0, I presume the planned installer adaption will include `BugReporter.*`, too. Or else should I prepare a PR for master?

RussKie:
Sorry. Yes, I had a look, the config is probably needed - it has the
binding redirects.

In .NET redirects are stored in runtime.configs IIRC. We'll cross that
bridge when we get there.

PR #9182, opened 2021-05-07
Set GIT_SSH only if path exists

### LINKED ISSUE #9112: SSH broken after upgrade to 3.5.0 (due to breaking setting change)

gerhardol:
Fixes #9112

Presentation handled in #9149
A second part separated to #9188 

Also for 3.5, but translation changes to be null

## Proposed changes

If the SSH path does not exist, do not set GIT_SSH env var. 
Git will then attempt to use OpenSSH included with Git.

In 3.4 or earlier, the OpenSSH path was set in the registry, even if it was supposed to be empty (handled in #9149).
If the OpenSSH path is corrupted (as reported and discussed in #9112 when upgrading GE) it will just be ignored.
This will also recover from issues if the Git installation is changed.

The settings will check and report an illegal path too.

Note that viewing the SSH setting will change the presented, then saved path if the path does not exist. A follow up PR will address this.

## Test methodology 

Modify the gitssh key in the registry and start GE and make sure that OpenSSH can still be used.

----

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
@@ -27,8 +27,9 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         private readonly TranslationString _gitVersionFound =
             new TranslationString("Git {0} is found on your computer.");
 
-        private readonly TranslationString _unknownSshClient =
-            new TranslationString("Unknown SSH client configured: {0}.");
+        private readonly TranslationString _sshClientNotFound = new("SSH client not found: {0}.");
+
+        private readonly TranslationString _otherSshClient = new("Other SSH client configured: {0}.");
 
         private readonly TranslationString _linuxToolsSshNotFound =
             new TranslationString("Linux tools (sh) not found. To solve this problem you can set the correct path in settings.");
@@ -196,6 +197,10 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
                 {
                     PageHost.GotoPage(SshSettingsPage.GetPageReference());
                 }
+            }
+            else
+            {
+                PageHost.GotoPage(SshSettingsPage.GetPageReference());
             }
         }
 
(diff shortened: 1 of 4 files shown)


RussKie (approved):

gerhardol:
Added then removed cleanup of handling registry vs GIT_SSH env var. That part could be skipped for 3.5

gerhardol:
@msftbot merge in 1 day

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -27,8 +27,11 @@ public partial class ChecklistSettingsPage : SettingsPageWithHeader
         private readonly TranslationString _gitVersionFound =
             new TranslationString("Git {0} is found on your computer.");
 
-        private readonly TranslationString _unknownSshClient =
-            new TranslationString("Unknown SSH client configured: {0}.");
+        private readonly TranslationString _sshClientNotFound =
+            new TranslationString("SSH client not found: {0}.");
+
+        private readonly TranslationString _otherSshClient =
+            new TranslationString("Other SSH client configured: {0}.");
```
mstv, line 34: Nit:
```suggestion
        private readonly TranslationString _sshClientNotFound = new("SSH client not found: {0}.");

        private readonly TranslationString _otherSshClient = new("Other SSH client configured: {0}.");
```
Should we activate the regarding code analyzer?

mstv (approved):
LGTM, have not run

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
gerhardol: Added a commit for IDE0090
https://github.com/gitextensions/gitextensions/pull/9095#issuecomment-840428715

This was on my mind, but it would have to be handled when rebasing c#9 new()

PR #9185, opened 2021-05-08
fix: remove unused functions

MrJithil:
Fixes #

## Proposed changes

- 
- 
- 

## Test methodology 

- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

----

### DIFF
--- a/GitUI/Theming/ThemeModule.cs
+++ b/GitUI/Theming/ThemeModule.cs
@@ -26,23 +26,6 @@ namespace GitUI.Theming
             Win32ThemeHooks.ThemeSettings = Settings;
         }
 
-        private static void InstallHooks(Theme theme)
-        {
-            Win32ThemeHooks.WindowCreated += Handle_WindowCreated;
-
-            try
-            {
-                Win32ThemeHooks.InstallHooks(theme, new SystemDialogDetector());
-            }
-            catch (Exception)
-            {
-                Win32ThemeHooks.Uninstall();
-                throw;
-            }
-
-            ResetGdiCaches();
-        }
-
         private static ThemeSettings LoadThemeSettings(IThemeRepository repository)
         {
             Theme invariantTheme;
--- a/GitUI/Theming/ThemeLoader.cs
+++ b/GitUI/Theming/ThemeLoader.cs
@@ -16,9 +16,6 @@ namespace GitUI.Theming
 
     public class ThemeLoader : IThemeLoader
     {
-        private const string ClassSelector = ".";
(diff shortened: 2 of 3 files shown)


GitUI/Theming/ThemeModule.cs
```diff
@@ -26,23 +26,6 @@ public static void Load()
             Win32ThemeHooks.ThemeSettings = Settings;
         }
 
-        private static void InstallHooks(Theme theme)
```
gerhardol, line 29: This function (and probably the other) had to be disabled in the .NET5 merge with `#if SUPPORT_THEMES`, as there is no solution for EasyHook yet.
I really hope that we can restore this functionality soon, so I do not want to loose theming.

gerhardol (commented):
-1

GitUI/Theming/ThemeModule.cs
MrJithil: Why don't we added TODO or just a comment in that case? I strongly recommend adding comments to the preserved code for future purposes, so people start contributing tomorrow will not lose their efforts.

gerhardol: The .NET5 is quite a big change that RussKie has spent a lot of time on, rebaseing many times. 
Great work but he had to merge to get going and everything is not perfect. (there are some tasks in the PR). Be easy on RussKie, he spends a lot of time on GE.

The usage of the functions are marked.
I would vote for a PR setting the `#if` around these functions (assuming all are for theming).

RussKie (changes requested):
Theme-related functionality must be commented out with `#if SUPPORT_THEMES`

GitUI/Theming/ThemeModule.cs
MrJithil: I couldn't see any #if blocks for the functions I removed. BTW, since this PR is drafted, we can ignore the causes.

MrJithil: > Theme-related functionality must be commented out with `#if SUPPORT_THEMES`

Yes. That's my point. If there is no `#if ` block for the functions, it's hard to predict the future purposes especially for the new contributors.

MrJithil:
> Theme-related functionality must be commented out with `#if SUPPORT_THEMES`

Yes. That's my point. If there is no #if block for the functions, it's hard to predict the future purposes especially for the new contributors.

gerhardol:
It was -1 for the merge, not the effort...

Issue #7662, opened 2020-01-20
No supported authentication methods available (server sent: publickey)


Kundan22:
## Current behaviour
I am using visual studio 2019 and recently upgraded git version 2.25.0.windows.1. 
All the git operations are working fine with gitbash and tortoise windows extensions. I've my ssh public key file kept in C:\Users\user-name\.ssh & C:\Program Files\Git\etc\ssh

but when I am trying to push my changes using visual studio I am getting below error[ScreenShot].

##am I missing any configuration or place where public key should be present?

## Expected behaviour
As git bash is and other plugin tortoise git is working fine visual studio github extension should also work for me

## Steps to reproduce
Open any local github repository , make changes and try to push it to master branch. and it throws similar error
## Screenshots

![image](https://user-images.githubusercontent.com/31039343/72727134-786a9d00-3b8a-11ea-8964-377a0cc9560b.png)

## Did this work in previous version of GitExtensions
Yes It was working fine with earlier version v2.17.1(2)

## Environment

- GitExtensions version: 2.10.8.8121
- GIT version: 2.25.0
- OS version: Windows10, VS2019
- .NET version: 4.7.2
- VisualStudio: VS2019

gerhardol:
Do you have a PuTTY public key setup?
From the paths you refer to, it could be that you have switched from OpenSSH to PuTTY in the settings.

Kundan22:
where do I check that, in Visual Studio or somewhere else?

vbjay:
run 

```bash
git config --global --edit
git config --edit
```
in a git bash console pointing to a project that has this issue and reply with the contents of each removing any sensitive data.

Kundan22:
running git command from same project directory with git bash I am getting error 
`bash: $'\302\226git': command not found` in both cases

vbjay:
Sounds like you might have environment issues. Try uninstalling git and
then reinstalling it or repairing the installation.

vbjay:
https://git-scm.com/

Kundan22:
re-installed git from https://git-scm.com/ and same issue. Git bash & Tortoise git works fine and asks for ssh key credential as soon as I Push the changes.

But Visual Studio and VSCode git bash gives this error when I tries to push the changes

![image](https://user-images.githubusercontent.com/31039343/72823014-56007e80-3c73-11ea-9e6d-82355772e947.png)

**after re-installing following commands gives this output now**

git config --global --edit

```
[gui]
[user]
	name = username
[...]

PR #9188, opened 2021-05-09
Keep persistent ssh settings in registry

### LINKED ISSUE #1663: SSH setting is not saved
If I change SSH setting to OpenSSH and click to OK/Apply it's saved but only until I close GitExt. If GitExt is opened again, SSH is set to PuTTY.
Reinstall, register cleaning doesn't help. 
Tested in versions 2.43 and 2.44 too on Windows 8 64 bit

### LINKED ISSUE #8162: Changing SSH setting to use OpenSSH does not persist
## Current behaviour

When I go into settings -> SSH and try to change it from PuTTY to OpenSSH, it changes it while GitExtensions is opened. After closing GitExtensions and reopening it, the setting is changed back to PuTTY.

## Expected behaviour

Changing the setting from PuTTY to OpenSSH should persist after the GitExtensions application is reopened.

## Screenshots

Here is a GIF screen recording showing the issue: https://i.imgur.com/9Neh2BX.gifv

I don't recall for sure. But I'm pretty sure it used to work.

## Environment

- Git Extensions 3.3.1.7897
- Git 2.26.2.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)
- I have tried running GitExtensions as administrator to no avail.
- Also note that I have installed PuTTY manually, before I installed GitExtensions.
- I also tried reinstalling GitExtensions, making sure that OpenSSH is selected during install, to no avail. That's almost it's own bug, the fact that I have OpenSSH selected during install, and right after installing and running GitExtensions, PuTTY is selected. But most likely related.

gerhardol:
Fixes #8162
Fixes #1663
Also see #7662 and a few others

See #9202 for a partial change for 3.5

## Proposed changes

Just read path from AppSettings.SshPath
Set GIT_SSH env var only set, not read.

Migrate explicit OpenSSH in settings to empty

Previously, the app first read registry for gitssh path and set environment variable GIT_SSH unless it was empty. This was maybe a feature for some, as you could share SSH settings.
However, when setting SSH in settings, GIT_SSH was always set which was inconsistent.
The behavior is changed to only set GIT_SSH, never set it.

## Test methodology 

Tests updated
As logic is simplified, tests are simplified too

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -62,7 +62,6 @@ namespace GitCommands
         public static readonly string SettingsFileName = ApplicationId + ".settings";
         public static readonly string UserPluginsDirectoryName = "UserPlugins";
         private static string _applicationExecutablePath = Application.ExecutablePath;
-        private static readonly ISshPathLocator SshPathLocatorInstance = new SshPathLocator();
         private static string? _documentationBaseUrl;
 
         public static Lazy<string?> ApplicationDataPath { get; private set; }
@@ -117,6 +116,7 @@ namespace GitCommands
             }
 
             MigrateAvatarSettings();
+            MigrateSshSettings();
         }
 
         /// <summary>
@@ -301,10 +301,9 @@ namespace GitCommands
             set => WriteStringRegValue("CascadeShellMenuItems", value);
         }
 
-        [MaybeNull]
         public static string SshPath
         {
-            get => ReadStringRegValue("gitssh", null);
+            get => ReadStringRegValue("gitssh", "");
             set => WriteStringRegValue("gitssh", value);
         }
 
@@ -660,6 +659,20 @@ namespace GitCommands
             if (providerChanged || fallbackImageChanged)
             {
                 SaveSettings();
+            }
+        }
+
+        private static void MigrateSshSettings()
+        {
+            string ssh = AppSettings.SshPath;
+            if (!string.IsNullOrEmpty(ssh))
+            {
+                // OpenSSH uses empty path, compatibility with path set in 3.4
+                string path = new SshPathLocator().GetSshFromGitDir(GitBinDir);
+                if (path == ssh)
+                {
+                    AppSettings.SshPath = "";
+                }
             }
         }
 
@@ -1529,7 +1542,6 @@ namespace GitCommands
             {
                 SettingsContainer.LockedAction(() =>
                 {
-                    SshPath = SshPathLocatorInstance.Find(GitBinDir);
                     SettingsContainer.Save();
                 });
 
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
@@ -136,7 +136,6 @@ namespace GitUI.CommandsDialogs.SettingsDialog.Pages
         #endregion
 
         private const string _putty = "PuTTY";
-        private readonly ISshPathLocator _sshPathLocator = new SshPathLocator();
         private DiffMergeToolConfigurationManager? _diffMergeToolConfigurationManager;
 
         /// <summary>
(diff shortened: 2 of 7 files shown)


GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -450,7 +450,7 @@ private bool CheckSSHSettings()
                                         _puttyConfigured.Text);
             }
 
-            string ssh = _sshPathLocator.Find(AppSettings.GitBinDir);
+            string ssh = AppSettings.SshPath;
```
mstv, line 453: `_sshPathLocator` can be removed.

GitUI/CommandsDialogs/SettingsDialog/Pages/SshSettingsPage.cs
```diff
@@ -34,7 +34,7 @@ protected override void SettingsToPage()
             PageantPath.Text = AppSettings.Pageant;
             AutostartPageant.Checked = AppSettings.AutoStartPageant;
 
-            var sshPath = _sshPathLocator.Find(AppSettings.GitBinDir);
+            var sshPath = AppSettings.SshPath;
```
mstv, line 37: ditto

GitUI/Infrastructure/Telemetry/AppEnvironmentTelemetryInitializer.cs
```diff
@@ -11,7 +11,7 @@ internal class AppEnvironmentTelemetryInitializer : ITelemetryInitializer
         public void Initialize(ITelemetry telemetry)
         {
             string sshClient;
-            var sshPath = _sshPathLocator.Find(AppSettings.GitBinDir);
+            var sshPath = AppSettings.SshPath;
```
mstv, line 14: ditto

mstv (commented):
Should work, have not run.
A little cleanup can be done:

GitCommands/Settings/AppSettings.cs
```diff
@@ -663,6 +664,22 @@ private static void MigrateAvatarSettings()
...
 
+        private static void MigrateSshSettings()
+        {
+            var sshPathLocator = new SshPathLocator();
+            string ssh = AppSettings.SshPath;
+
+            if (!string.IsNullOrEmpty(ssh))
+            {
+                // OpenSSH uses empty path, compatibility with path set in 3.4
+                string path = sshPathLocator.GetSshFromGitDir(GitBinDir);
```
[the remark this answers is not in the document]
gerhardol: Removed the variable

mstv (approved):

RussKie:
👍 

> Just read path from AppSettings.SshPath
> Set GIT_SSH env var only set, not read.
> 
> Migrate explicit OpenSSH in settings to empty
> 
> Previously, the app first read registry for gitssh path and set environment variable GIT_SSH unless it was empty. This was maybe a feature for some, as you could share SSH settings.
> However, when setting SSH in settings, GIT_SSH was always set which was inconsistent.
> The behavior is changed to only set GIT_SSH, never set it.

Please put it in the commit message infull.

gerhardol:
@msftbot merge in 14 hours

gerhardol:
The UI tests on AppVeyor occasionally timed out with this PR. They work fine locally.
The reason is that you may get null exceptions due to registry settings missing and the access function was annotated with `[MaybeNull]` and null was not handled for the plink check. Should be OK elsewhere, not sure about Telemetry though.
Changing the default to "" to follow null annotation. We are not interested in knowing if the registry key exists or not.

Force pushing if the automatic merge still kicks in...

````patch
diff --git a/GitCommands/Settings/AppSettings.cs b/GitCommands/Settings/AppSettings.cs
index ca7f5..b05a1 100644
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -301,10 +301,9 @@ public static string CascadeShellMenuItems
             set => WriteStringRegValue("CascadeShellMenuItems", value);
         }
 
-        [MaybeNull]
         public static string SshPath
         {
-            get => ReadStringRegValue("gitssh", null);
+            get => ReadStringRegValue("gitssh", "");
             set => WriteStringRegValue("gitssh", value);
         }
 
```

mstv (approved):

gerhardol:
@msftbot merge in 24 hours

PR #9202, opened 2021-05-13
Set GIT_SSH only if path exists 3.5

gerhardol:
#9182 and part of #9188 for 3.5
(#9188 could be applied but it introduces plugin interface changes)
#9182 Translation changes in #9182 was reverted too (stating "unknown" instead of "other")

## Proposed changes

Set GIT_SSH only if path exists

Just read path from AppSettings.SshPath
Set GIT_SSH env var only, not read.

Report if ssh path is not found in Settings

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitSshHelpers.cs
+++ b/GitCommands/Git/GitSshHelpers.cs
@@ -1,5 +1,5 @@
 using System;
-using GitExtensions;
+using System.IO;
 
 namespace GitCommands
 {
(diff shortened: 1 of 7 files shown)


RussKie (approved):
:shipit:

GitCommands/Git/GitSshHelpers.cs
```diff
@@ -26,19 +26,26 @@ bool DoArgumentsRequireSsh()
...
         {
-            if (!Strings.IsNullOrEmpty(path))
+            // Git will use the embedded OpenSSH ssh.exe if empty/unset
+            if (!string.IsNullOrEmpty(path) && !File.Exists(path))
             {
-                Environment.SetEnvironmentVariable("GIT_SSH", path, EnvironmentVariableTarget.Process);
+                path = "";
             }
+
+            if (!string.IsNullOrEmpty(path) && !File.Exists(path))
```
gerhardol, line 38: Force pushing a backporting error (just unneeded, no negative effect as this is the path Git will use anyway).
```suggestion
            if (!string.IsNullOrEmpty(path))
```

gerhardol:
Note: I would like to merge #9188 first, so the manual cherry-pick can refer to the correct commit

gerhardol:
Most changes in master was eventually added to 3.5 as well

RussKie (approved):

PR #9203, opened 2021-05-13
LogCaptureCallStacks in AppSettings

gerhardol:
## Proposed changes

Related to Git issues, like debugging for #9056 it is very helpful to have call stacks always available in the Command Log. This is especially important for commands during the startup phase, before the capture can be enabled.
It can be hard to recreate the situation a reporter experiences.

I often have this enabled in my private builds, but with #9056 I expect at least developers to require this.
(#9056 is a good change to avoid just swallowing Git errors.)

## Screenshots 

No UI change, just showing what is made persistent

image

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1863,6 +1863,12 @@ namespace GitCommands
             set => SetInt("DiffViewer.AutomaticContinuousScrollDelay", value);
         }
 
+        public static bool LogCaptureCallStacks
+        {
+            get => GetBool("Log.CaptureCallStacks", false);
+            set => SetBool("Log.CaptureCallStacks", value);
+        }
+
         public static bool IsPortable()
         {
             return Properties.Settings.Default.IsPortable;
--- a/GitCommands/Logging/CommandLog.cs
+++ b/GitCommands/Logging/CommandLog.cs
@@ -154,8 +154,6 @@ namespace GitCommands.Logging
 
         private static ConcurrentQueue<CommandLogEntry> _queue = new ConcurrentQueue<CommandLogEntry>();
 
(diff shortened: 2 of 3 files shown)


GitCommands/Settings/AppSettings.cs
```diff
@@ -1863,6 +1863,12 @@ public static int AutomaticContinuousScrollDelay
             set => SetInt("DiffViewer.AutomaticContinuousScrollDelay", value);
         }
 
+        public static bool LogCaptureCallStacks
+        {
+            get => GetBool("logcapturecallstacks", false);
+            set => SetBool("logcapturecallstacks", value);
```
mstv, line 1869:
```suggestion
            get => GetBool("Log.CaptureCallStacks", false);
            set => SetBool("Log.CaptureCallStacks", value);
```
?

mstv (commented):
So, we should activate it manually in the settings file.
Perhaps rename - though the setting name style varies.

GitCommands/Settings/AppSettings.cs
gerhardol: The value is persistent in settings, by changing in Command Log.
Sure, I could make it not persisted but that would be more code.

For name I could change, but lowercase seem more common

RussKie: +1 to @mstv suggestion

RussKie (approved):
👍 modulo @mstv suggestion

GitCommands/Settings/AppSettings.cs
gerhardol: Changed settings name and force pushed

gerhardol:
@msftbot merge in 14 hours

PR #9206, opened 2021-05-13
Use enum interface for GetRefs()

gerhardol:
https://github.com/gitextensions/gitextensions/pull/8990#discussion_r594314175

## Proposed changes

Replace the confusing GetRefs() interface with the enum interface introduced in #8990.
(A little too late change for 3.5).
Some optimizations for getting remotes, that had a special handling before.

## Test methodology 

Tests updated
Code review

----


mstv (commented):
A very sensible refactoring.

Untouched code cannot be commented inline at GH - so here:

`GetRefsEnum` should be renamed to `GetRefsFlags`.

For the enum values, either stick to the (arguable) git naming (also used for `GitRefType`) or be descriptive:
`.Branches` means `.Heads` or `.LocalBranches`
`.Remotes` means `.RemoteBranches`

Typo in comment of `.All`: "abve".

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -42,7 +43,7 @@ public FormDeleteBranch(GitUICommands commands, IEnumerable<string> defaultBranc
 
         private void FormDeleteBranchLoad(object sender, EventArgs e)
         {
-            Branches.BranchesToSelect = Module.GetRefs(tags: true, branches: true).Where(h => h.IsHead && !h.IsRemote).ToList();
+            Branches.BranchesToSelect = Module.GetRefs(GetRefsEnum.Branches | GetRefsEnum.Tags).Where(h => h.IsHead).ToList();
```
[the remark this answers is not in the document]
gerhardol: Except for Remotes, I did not try to optimize the handling, just get something that worked as before.
I noticed this too, but concentrated on something that worked the same.
Will change.

GitUI/CommandsDialogs/FormMergeBranch.cs
```diff
@@ -69,7 +70,7 @@ private void FormMergeBranchLoad(object sender, EventArgs e)
             var selectedHead = Module.GetSelectedBranch();
             currentBranchLabel.Text = selectedHead;
 
-            Branches.BranchesToSelect = Module.GetRefs(true, true);
+            Branches.BranchesToSelect = Module.GetRefs(GetRefsEnum.All);
```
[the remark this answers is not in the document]
gerhardol: If changed, Tags is not needed. There may be a few other such changes too (that was hard to understand with the previous API).
These changes should probably be separated to a separate commit.

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -135,7 +135,7 @@ public FormPush(GitUICommands commands)
 
             void Init()
             {
-                _gitRefs = Module.GetRefs();
+                _gitRefs = Module.GetRefs(GetRefsEnum.All);
```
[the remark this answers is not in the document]
gerhardol: same as above, resolving

GitUI/FilterBranchHelper.cs
```diff
@@ -117,40 +118,12 @@ public void InitToolStripBranchFilter()
...
-            if (tag)
-            {
-                list.AddRange(GetTagsRefs());
-            }
+            GetRefsEnum refs = (local ? GetRefsEnum.Branches : GetRefsEnum.None)
+                | (tag ? GetRefsEnum.Tags : GetRefsEnum.None)
+                | (remote ? GetRefsEnum.Remotes : GetRefsEnum.None);
+            var list = Module.GetRefs(refs).Select(branch => branch.Name).ToList();
 
             return list;
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/CommandsDialogs/FormPush.cs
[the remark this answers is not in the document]
gerhardol: Only IsHead, IsRemote is used, changing

GitUI/CommandsDialogs/FormMergeBranch.cs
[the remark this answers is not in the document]
gerhardol: Similar in FormRebase (changing or commenting all GetRefsEnum.All occurrences).
So Tags should be included.
Rebasing on stash works for Git and can still be done in the revision grid (no need to block but also not a feature).
```suggestion
            // Offer rebase on refs also for tags (but not stash, notes etc)
            Branches.BranchesToSelect = Module.GetRefs(GetRefsEnum.Branches | GetRefsEnum.Remotes | GetRefsEnum.Tags);
```

gerhardol:
> Untouched code cannot be commented inline at GH - so here:
> 
> `GetRefsEnum` should be renamed to `GetRefsFlags`.

GitRefsFlags

> 
> For the enum values, either stick to the (arguable) git naming (also used for `GitRefType`) or be descriptive:
> `.Branches` means `.Heads` or `.LocalBranches`
> `.Remotes` means `.RemoteBranches`

I would like to stick with the Git refs/ names so Heads, Remotes and Tags (and clarify at the definition)

> Typo in comment of `.All`: "abve".

OK

GitUI/CommandsDialogs/FormDeleteBranch.cs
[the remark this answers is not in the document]
gerhardol: I may separate the Remotes changes to a separate commit too

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -958,7 +958,8 @@ public void ForceRefreshRevisions()
 
                 _revisionReader ??= new RevisionReader();
 
-                var refs = Module.GetRefs(GetRefsEnum.All);
+                // Find all ambigous refs (including stash, notes etc)
```
mstv, line 961:
```suggestion
                // Find all ambiguous refs (including stash, notes etc)
```

mstv (approved):
👍

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
```diff
@@ -1,18 +1,21 @@
-﻿namespace GitCommands.Git.Commands
+﻿namespace GitUIPluginInterfaces
 {
     /// <summary>
     /// Enums requestable in GitRefs() (multiple names can be appended)
     /// Compare to <see ref="GitRefType"/> for actual values of parsed GitRefs
     /// </summary>
-    public enum GetRefsEnum
+    public enum GitRefsFlags
```
RussKie, line 7: https://docs.microsoft.com/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations
```suggestion
    [Flags]
    public enum GitRefs
```

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
```diff
@@ -1,18 +1,21 @@
...
+        // Local Branches
+        Heads = 1 << 0,
+
+        // Remote branches
         Remotes = 1 << 1,
         Tags = 1 << 2,
 
-        // All refs, including those abve but also (at least) stash, notes and bisect
+        // All refs, including those above but also (at least) stash, notes and bisect
         All = 1 << 3
```
RussKie, line 19:
```suggestion
        Everything = 1 << 3 | Heads | Remotes | Tags
```
...or rewrite the enum as
```cs
 Heads      = 0x0001,
 Remotes    = 0x0010,
 Tags       = 0x0100,
 Everything = 0x1111
```

RussKie, line 19: Should we add "stash, notes and bisect" as individual entries as well?

gerhardol: Only add those we specifically need to get separately.

Any namespace can be added below refs/. 
Gerrit uses at least refs/for/ and refs/changes/ so all cannot be added.

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
gerhardol: Can the guidelines be summarized somewhere of the current codestyle?
Even more important for new contributors.

Changing this, but the name is very close to the existing class GitRef, this is confusing.

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
gerhardol: Prefer not to change this, then a separate flag is needed for All anyway.
The reason is that HasFlagButNotAll() need to be able to check if All bit is set, then the arguments for git-for-each-ref should be empty and that check cannot be done.

Sure, the most significant bit for All can be dynamically calculated in GetRefsCmd() too.
So the following (and another minor change) is required.
```c#
            static bool HasFlagButNotAll(GitRefs requested, GitRefs flag)
            {
                const GitRefs allOnly = GitRefs.All & ~(GitRefs.Heads | GitRefs.Remotes | GitRefs.Tags);
                return (requested & allOnly) == 0 && (requested & flag) != 0;
            }
```

gerhardol: I refactored GetRefsCmd() some to make this handling clearer.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -79,49 +79,29 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
...
                     : default)
             {
-                { sortBy != GitRefsSortBy.Default, GetSortCriteria(hasTags, sortBy, sortOrder), string.Empty },
-                format,
-                { HasFlag(getRef, GetRefsEnum.Branches), "refs/heads/", string.Empty },
-                { HasFlag(getRef, GetRefsEnum.Remotes), "refs/remotes/", string.Empty },
-                { HasFlag(getRef, GetRefsEnum.Tags), "refs/tags/", string.Empty },
+                sortCriteria(hasTags, sortBy, sortOrder),
+                gitRefsFormat(getRef, hasTags),
+                gitRefsPattern(getRef)
```
mstv, line 93:
```suggestion
                GetSortCriteria(hasTags, sortBy, sortOrder),
                GetGitRefsFormat(getRef, hasTags),
                GetGitRefsPattern(getRef)
```
?
Or else new convention for local functions?

mstv (approved):

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: I had a squiggle to adjust the name that I acted on but do not see that now, changing back...

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
RussKie: > Can the guidelines be summarized somewhere of the current codestyle?

For many years in some shape or form I've been using Microsoft .NET design guidelines - originally [posted by Brad Abrams](https://docs.microsoft.com/en-us/archive/blogs/brada/internal-coding-guidelines) and the [current official](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/). 

A lot of styling guidelines can now be enforced with the help of StyleCop and Roslyn analyzers (and that what we've been doing for some time), though some things are hard to enforce automatically (like naming).

RussKie: > Changing this, but the name is very close to the existing class GitRef, this is confusing.

Perhaps we need to revisit the naming of the class or both. E.g. the enum can also be named `GitRefTypes`...

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
RussKie: 🤦‍♂️ Looks like I've made a clown of myself. Here's what I meant to write above:
```cs
[Flags]
public enum GitRefs : byte
{
	None       = 0,
	Heads      = 0x01,          // 00000001
	Remotes    = 0x02,          // 00000010
	Tags       = 0x04,          // 00000100
	Everything = byte.MaxValue  // 11111111
}
```


My point was that `All` makes no sense as a single bit. E.g. `GitRefs.Remotes | GitRefs.Tags` means "_remotes and tags_", but `GitRefs.All | GitRefs.Remotes | GitRefs.Tags` means "_everything_" with "remotes and tags" being totally superficial.

Since the members in the enum aren't representing the exhaustive set of values (which we may consider adding in the future), `All` (or `Everything`) must essentially be represented by "all bits set" value. And to assert whether it is set we just need to read the top bit, e.g. `((int)requested >> 7) == 1`

gerhardol: In this scenario, `Other` will not get any data at all or be the same as `Everything` which is confusing for an API usage.
The enums is the API, should be clear from a usage point of view.
This is also why I do not want to have a `None` value. I that value is used alone, the Git command should throw as it is meaningless, and how should `None|Tags` be interpreted?

All is a special mode comparable to Heads/Remotes/Tags even if All contains the latter. So `All|Tags` is the same as `All` regardless if all bits are set or not. Sure, that All is not used with other could be asserted, but that will be a runtime check without any real meaning.

I can add all bits to All but the internal check is a little weird in my opinion, the check must mask the enum to isolate the All bit.

Personally I find 1>>1, 1>>2 etc be easier to understand than listing hexvalues. (binary values work well here, harder when there are many values).

mstv: `GetRefs` uses this argument to control the filtering, where `All` means `Unfiltered`. Hence the enum value should be defined as `Unfiltered = 0`.

gerhardol: > `GetRefs` uses this argument to control the filtering, where `All` means `Unfiltered`. Hence the enum value should be defined as `Unfiltered = 0`.

Good suggestion, that should work
The name of the type should then be `GitRefsFilter`

gerhardol: `Unfiltered|Tags` will be the same as `Tags` which could be confusing, that is API documentation.

mstv: > `Unfiltered|Tags` will be the same as `Tags` which could be confusing, that is API documentation.

Then `GitRefsFilter.None = 0`.

GitUI/FilterBranchHelper.cs
```diff
@@ -117,42 +119,15 @@ public void InitToolStripBranchFilter()
...
-        }
-
         private List<string> GetBranchAndTagRefs(bool local, bool tag, bool remote)
         {
-            var list = GetBranchHeads(local, remote);
-            if (tag)
-            {
-                list.AddRange(GetTagsRefs());
-            }
+            // Options are interpreted as the refs the search should be limited too
```
gerhardol, line 124: This is a scope creep compared to master, I opted to add a comment instead of code to limit the functionality

gerhardol:
Squashed without further changes.
I had planned to separate the optimizations to a separate commit, but after the last changes there will be more work than added value.

gerhardol:
@msftbot squash in 3 days

mstv (approved):

Plugins/GitUIPluginInterfaces/GitRefsFlags.cs
RussKie: Thank you @mstv - this makes sense :)

RussKie:
MC

Plugins/GitUIPluginInterfaces/RefsFilter.cs
```diff
@@ -0,0 +1,23 @@
...
+        // No filter, include those below but also (at least) stash, notes and bisect
+        // Note that if NoFilter is combined with a filter, the filter takes precedence
+        NoFilter = 0,
+
+        // Local Branches
+        Heads = 1 << 0,
+
+        // Remote branches
+        Remotes = 1 << 1,
+        Tags = 1 << 2
```
RussKie, line 21:
```suggestion

        // Tags
        Tags = 1 << 2
```

RussKie (approved):

Plugins/GitUIPluginInterfaces/RefsFilter.cs
gerhardol: Squashed this change (no other real merge conflict) when rebasing

RussKie:
Merge at will

Issue #9111, opened 2021-04-25
Accessibility: contrast too low


jrfnl:
## Current behaviour

Since upgrading to 3.5.0, it appears the default theme has been adjusted and the background for branch name labels has been changed.

This makes the text contrast too low, which makes reading these for anyone over 40 much harder.

## Expected behaviour

That the `default` theme is not changed on a whim, but put in a new theme. Or at the very least that changes take accessibility into account.

## Steps to reproduce

Upgrade from 3.4.3 to 3.5.0 with `default` theme.

## Screenshots

![image](https://user-images.githubusercontent.com/663378/115997680-590bbc80-a5e4-11eb-9526-a1935b749eeb.png)

## Did this work in previous version of GitExtensions

Yes.

## Environment

- Git Extensions 3.5.0.11713
- Build 0a5ef9ca6ce87754a510aa02c7a96d8b915714ac (Dirty)
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4341.0
- DPI 144dpi (150% scaling)

ivangrek:
Contrast or brightness?

jrfnl:
I'd call it contrast, but then, I'm not a designer...

mstv:
For me it is the contrast in the rows with the gray background and especially that it differs from row to row. The frame is distracting, too.

jrfnl:
I think the removal/toning down of the background colour of the branch labels has a lot to do with it and I can't really see how I could customize this via a custom Theme (which I've started to create to undo the bug reported here: https://github.com/gitextensions/gitextensions/issues/9113).

jrfnl:
Looks like this is related to/caused by #8900.

RussKie:
@ivangrek the current colour scheme/contrast are likely failing the [minimum contrast test](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html).

For accessibility tests we use [Accessibility Insights](https://accessibilityinsights.io/docs/en/windows/overview/).
Use the "Color Contrast Analyzer" (3rd items in the menu) to find correct colours:
![image](https://user-images.githubusercontent.com/4403806/116024792-54054680-a692-11eb-84cb-d545e08111a8.png)

ivangrek:
Thank you. I am research this situation.

ivangrek:
### Base color
[...]

PR #9207, opened 2021-05-14
Accessibility: revision grid branch colors

gerhardol:
Part of #9111 for 3.5
(no PR for master yet, no theme handling even if this commit will not change themes)

This is based of what @ivangrek investigated, I take no credit . Even if this is not final, something should probably be done for 3.5.1 that hopefully can be released soon.
Note especially that no extra investigations for selected commits was done. There are also reports for tag labels.
Note: The tuning was adjusting red to green only, not branch colors to the background which is what matters here.

Edit: Also changed to revert colors to same as in 3.4 (Branch/RemoteBranch still switched compared to 3.4).

## Proposed changes

Adjust colors for branches in the default theme

## Screenshots 

release/3.5 | first tuning | reverted
--|--|--
image | image | image

## Test methodology 

manual

----


GitUI/Themes/dark.css
```diff
@@ -38,9 +38,9 @@
 .HighlightAllOccurences      { color: #606020; }
 .Tag                         { color: #408080; }
 .Graph                       { color: #882d00; }
-.Branch                      { color: #00d76b; }
+.Branch                      { color: #3d9967; }
 .Branch.colorblind           { color: #0080ff; } /* hsl(210, 100%, 50%) */
-.RemoteBranch                { color: #ff8080; } /* hsl(0,   100%, 75%) */
+.RemoteBranch                { color: #de6459; } /* hsl(0,   100%, 75%) */
```
mstv, line 43:
```suggestion
.RemoteBranch                { color: #de6459; }
```

GitUI/Themes/darksilver.css
```diff
@@ -38,12 +38,12 @@
 .HighlightAllOccurences  { color: #606020; }
 .Tag                     { color: #408080; }
 .Graph                   { color: #882d00; }
-.RemoteBranch            { color: #ff8080; } /* hsl(0,   100%, 75%) */
+.RemoteBranch            { color: #de6459; } /* hsl(0,   100%, 75%) */
```
mstv, line 41:
```suggestion
.RemoteBranch            { color: #de6459; }
```

mstv (commented):
Have not run myself.
+1, only if you have swapped the "before" and "after" screenshots. 😉

gerhardol:
> +1, only if you have swapped the "before" and "after" screenshots. 😉

The order was correct, but not the changes that should be done to the invariant theme. Hard to see the changes though, I cannot say that this is better...
Also, the Current branch point is missing for unknown reason

RussKie:
> ## Screenshots
> before	after
> image	image

Sorry, what's the difference besides lost triangles? Why are the triangles are gone?

gerhardol:
> Sorry, what's the difference besides lost triangles? Why are the triangles are gone?

The colors are subtly changed to improve contrast (atleast with standard background).
Triangles lost for unknown reason.

This is a state of how far the issue is resolved. So it is not just to apply the investigation so far...
I am not good at designing and choosing colors and will not take this further...

mstv:
> Triangles lost for unknown reason.

🤔 Might the drawing be denied by `ThreadHelper.AssertOnUIThread()` in `DrawArrow` sometimes? But pretty unlikely. Just came across this while reviewing another PR. I have never noticed missing arrows.

gerhardol:
Updated proposal, new screenshots and arrow visible (using correct alpha).

The original colors look better in my opinion.

@jfrnl also wants to switch back the colors. I was voted down in the original change (that was merged to the 3.5 branch too late in the release process...), I expect that to stay.

RussKie:
> @jfrnl also wants to switch back the colors. I was voted down in the original change (that was merged to the 3.5 branch too late in the release process...), I expect that to stay.

I appreciate it may cause discomfort to some users, but I am of an opinion we made the right call to align with git's colour scheme.

RussKie (approved):

gerhardol:
Is this better? Should we merge it?

RussKie:
👍from me

RussKie:
Let's merge this

nolanhergert:
Before my "origin/master" branch was green, and now it is red. Also, local branches that were red are now green. Was this intended? Quite confusing when I'm trying to ramp up new people, and I would argue unintuitive since I want to encourage changes to be synced to remote.

gerhardol:
> Before my "origin/master" branch was green, and now it is red. Also, local branches that were red are now green. Was this intended? Quite confusing when I'm trying to ramp up new people, and I would argue unintuitive since I want to encourage changes to be synced to remote.
> 
> EDIT: I less want to complain but more confirm that this is what is intended. I now see the code change is very clearly doing this (as opposed to the initial screenshots above), so it's not accidental. I disagree with the decision, but it's not my project.

Yes, this was intended, changed to use the same color scheme as git-log command line.
This was much debate about this change both before and after the release and you are not the first to ask...

PR #9212, opened 2021-05-16
Preparations for exceptions on git error

mstv:
Preparations for #9056

## Proposed changes

- Handle more exceptions as "failed external operation" (same as #9094 for release/3.5)
- Do not crash on exception if debugger is attached
- Check `ExitedSuccessfully` in `GetStashDiffFiles`
- Avoid exit code 1 when checking for valid git executable
- `ExternalOperationException`: Add property `ExitCode`
- Remove unnecessary `ThreadHelper.AssertOnUIThread` before throwing `UserExternalOperationException`

### Before

image

### After

image

## Test methodology 

- extended existing NUnit tests
- manual tests with FormStash and `Include tracked` ticked

## Test environment(s) 

- Git Extensions 33.33.33
- Build 06582e3d0114e17540b03dbaa6f9388a0a0e07a2
- Git 2.31.1.windows.1
- .NET 5.0.6
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2457,16 +2457,15 @@ namespace GitCommands
         {
             var resultCollection = GetDiffFilesWithUntracked(stashName + "^", stashName, StagedStatus.None, true).ToList();
 
-            // shows untracked files
+            // add - optionally stashed - untracked files
             GitArgumentBuilder args = new("log")
             {
                 $"{stashName}^3",
                 "--pretty=format:\"%T\"",
                 "--max-count=1"
             };
-            var untrackedTreeHash = _gitExecutable.GetOutput(args);
-
-            if (ObjectId.TryParse(untrackedTreeHash, out var treeId))
+            ExecutionResult executionResult = _gitExecutable.Execute(args);
+            if (executionResult.ExitedSuccessfully && ObjectId.TryParse(executionResult.StandardOutput, out ObjectId? treeId))
             {
                 var files = GetTreeFiles(treeId, full: true);
 
--- a/GitExtUtils/ExternalOperationException.cs
+++ b/GitExtUtils/ExternalOperationException.cs
@@ -15,13 +15,20 @@ namespace GitExtUtils
         /// <param name="command">The command that led to the exception.</param>
         /// <param name="arguments">The command arguments.</param>
         /// <param name="workingDirectory">The working directory.</param>
+        /// <param name="exitCode">The exit code of an executed process.</param>
         /// <param name="innerException">The exception that is the cause of the current exception.</param>
-        public ExternalOperationException(string? command, string? arguments, string workingDirectory, Exception? innerException)
+        public ExternalOperationException(
+            string? command = null,
+            string? arguments = null,
+            string? workingDirectory = null,
+            int? exitCode = null,
+            Exception? innerException = null)
             : base(innerException?.Message, innerException)
         {
             Command = command;
             Arguments = arguments;
             WorkingDirectory = workingDirectory;
+            ExitCode = exitCode;
         }
 
         /// <summary>
@@ -38,5 +45,10 @@ namespace GitExtUtils
         /// The working directory.
         /// </summary>
         public string WorkingDirectory { get; }
+
+        /// <summary>
+        /// The exit code of an executed process.
+        /// </summary>
+        public int? ExitCode { get; }
     }
 }
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -67,7 +67,8 @@ namespace GitExtensions
             {
                 DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
 
-                if (!Debugger.IsAttached)
+                // If you want to suppress the BugReportInvoker when debugging and exit quickly, uncomment the condition:
+                ////if (!Debugger.IsAttached)
                 {
                     AppDomain.CurrentDomain.UnhandledException += (s, e) => BugReportInvoker.Report((Exception)e.ExceptionObject, e.IsTerminating);
                     Application.ThreadException += (s, e) => BugReportInvoker.Report(e.Exception, isTerminating: false);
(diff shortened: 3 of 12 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -2460,9 +2460,9 @@ public IReadOnlyList<GitItemStatus> GetStashDiffFiles(string stashName)
             // shows untracked files
             var args = new GitArgumentBuilder("log")
             {
-                $"{stashName}^3",
                 "--pretty=format:\"%T\"",
-                "--max-count=1"
+                "--max-count=1",
+                $"-- {stashName}^3"
```
gerhardol, line 2465: https://github.com/gitextensions/gitextensions/pull/9056#discussion_r616185182
https://git-scm.com/docs/git-log#_description

stashName is a revision, not a path and should not be after "--"
The "fatal:" outputs need to be handled

GitExtensions/Program.cs
```diff
@@ -67,7 +67,7 @@ private static void Main()
             {
                 DiagnosticsClient.Initialize(ThisAssembly.Git.IsDirty);
 
-                if (!Debugger.IsAttached)
+                ////if (!Debugger.IsAttached)
```
gerhardol, line 70: Why just comment out?
Remove and add comment instead?

gerhardol (changes requested):
Fine except for git-log stashName.
(I have been using most of these patches for some time as they are a subset of #9056).

GitCommands/Git/GitModule.cs
mstv: I noticed your argument. But I guess git might have a different definition of _paths_ because it works now without error.
Do you have repro where this changes causes wrong behavior?

GitExtensions/Program.cs
mstv: I left it for whom has added this arguable behavior in order make them be able to easily restore it temporarily.

GitCommands/Git/GitModule.cs
gerhardol: No repro, only Git documentation. 
All other commands with "--" are supposed to be followed by paths, so also if this works, it could be changed.
If I am voted down, this must at least be clearly commented.

GitExtensions/Program.cs
gerhardol: Then describe in a comment it, no ////

GitExtUtils/ExternalOperationException.cs
```diff
@@ -12,16 +12,23 @@ public class ExternalOperationException : Exception
...
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
         /// <param name="command">The command that led to the exception.</param>
         /// <param name="arguments">The command arguments.</param>
-        /// <param name="workingDirectory">The working directory.</param>
-        /// <param name="innerException">The exception that is the cause of the current exception.</param>
-        public ExternalOperationException(string? command, string? arguments, string workingDirectory, Exception? innerException)
+        /// <param name="directory">The directory of the operation, e.g. the working directory or the git repo.</param>
+        /// <param name="exitCode">The exit code of an executed process.</param>
+        public ExternalOperationException(
+            Exception? innerException = null,
```
RussKie, line 21: It is a convention to have inner exception as the last argument

GitExtensions/Program.cs
RussKie: This is done so when a developer runs with a debugger attached (e.g. from a VS) and there is an exception the debugger breaks on the exception, and not terminates the process.
Please revert this change.

GitUI/NBugReports/BugReportInvoker.cs
```diff
@@ -42,14 +45,17 @@ internal static string Append(StringBuilder text, Exception exception)
...
+
                 // Command: <command>
                 AppendIfNotEmpty(externalOperationException.Command, TranslatedStrings.Command);
 
                 // Arguments: <args>
                 AppendIfNotEmpty(externalOperationException.Arguments, TranslatedStrings.Arguments);
 
                 // Directory: <dir>
-                AppendIfNotEmpty(externalOperationException.WorkingDirectory, TranslatedStrings.WorkingDirectory);
+                AppendIfNotEmpty(externalOperationException.Directory, TranslatedStrings.Directory);
```
RussKie, line 58: Why are we departing from the notion of working directory? Git Extensions and git work with working directories, and don't work with arbitrary directories.

GitExtensions/Program.cs
mstv: It is opposite: If this condition is active, the application just terminates. One cannot debug the `BugReportInvoker` etc.

mstv: You are absolute right. I got the git error wrong. I have replaced this with a check for `ExitedSuccessfully`.
Else stashed untracked files were not displayed anymore.
grafik

GitExtUtils/ExternalOperationException.cs
mstv: Done

GitUI/NBugReports/BugReportInvoker.cs
mstv: Reverted this leftover of the planned generalized use of `ExternalOperationException` e.g. for file operations, which might be handled sufficiently since #9094.

GitCommands/Git/GitModule.cs
mstv: You are absolute right. I got the git error wrong. I have replaced this with a check for `ExitedSuccessfully`.
Else stashed untracked files were not displayed anymore.
grafik

GitExtensions/Program.cs
mstv: > This is done so when a developer runs with a debugger attached (e.g. from a VS) and there is an exception the debugger breaks on the exception, and not terminates the process.

🤔 In order to achieve _this_, I tick all exceptions but `TaskCanceledException`:

grafik

The `BugReportInvoker` kicks in later when the exception has bubbled up.

gerhardol (approved):

RussKie:
MC

mstv:
MC resolved

RussKie (approved):

Issue #9213, opened 2021-05-16
Content gets loaded from unrelated repo???


RussKie:
## Current behaviour

I have a long running GE instance, switching between repos back and forth. In a repo I added a blank new README.md, however GE shows me content from a README.md that belong to a different totally unrelated repo (that has previously been opened in this GE instance).

## Steps to reproduce

Not quite sure...

## Screenshots

The unstaged file content is **wrong**! Once stages the content is shown correctly.

![ge-file-content](https://user-images.githubusercontent.com/4403806/118416962-67398e00-b6f5-11eb-93ac-18bfd3e03c05.gif)

## Environment

- Git Extensions 3.5.0.11713
- Build 0a5ef9ca6ce87754a510aa02c7a96d8b915714ac (Dirty)
- Git 2.29.2.windows.2 (recommended: 2.30.0 or later)
- Microsoft Windows NT 10.0.19043.0
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

RussKie:
Looks like the problem lies here:
![image](https://user-images.githubusercontent.com/4403806/118417981-e3ce6b80-b6f9-11eb-9912-8adc24480b77.png)

...because of how the GE instance was started 
![image](https://user-images.githubusercontent.com/4403806/118418015-08c2de80-b6fa-11eb-8b3c-6bd7d5aa690e.png)

It resolve the file from the current directory, which is "C:\Development\winforms\".

gerhardol:
In this case the File.Exists() in ViewFileAsync() should be removed, the path is already evaluated.

However, if relative filepaths cannot be used, there will likely be a lot more situations than this specific case that do not work.

I can reproduce only when using openrepo argument, so current directory is not set then for unknown reason.

PR #9216, opened 2021-05-17
Use fullPathResolver for viewing local files

gerhardol:
Part of #9213

## Proposed changes

Refactor: Use existing FullPathResolve instead of multiple invocations.
This removes a File.Exists() that is resolved incorrectly.
If opening GE with openrepo arg (and more?), open another and viewing a new file in Worktree commit that exist in the first repo, the file in the existing repo was shown.

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.IO;
@@ -570,26 +571,32 @@ namespace GitUI.Editor
         public Task ViewFileAsync(string fileName, bool isSubmodule = false, Action? openWithDifftool = null)
         {
             string? fullPath = _fullPathResolver.Resolve(fileName);
-
-            if (isSubmodule && !GitModule.IsValidGitWorkingDir(fullPath))
-            {
-                return ViewTextAsync(fileName, "Invalid submodule: " + fileName);
-            }
-
             Validates.NotNull(fullPath);
+            Debug.Assert(Path.IsPathFullyQualified(fullPath), "Path must be resolved and fully qualified");
 
-            if (!isSubmodule && (fileName.EndsWith("/") || Directory.Exists(fullPath)))
+            if (!isSubmodule)
             {
-                if (!GitModule.IsValidGitWorkingDir(fullPath))
+                if (fileName.EndsWith("/") || Directory.Exists(fullPath))
                 {
-                    return ViewTextAsync(fileName, "Directory: " + fileName);
-                }
+                    if (!GitModule.IsValidGitWorkingDir(fullPath))
+                    {
+                        return ViewTextAsync(fileName, "Directory: " + fileName);
+                    }
 
-                isSubmodule = true;
+                    isSubmodule = true;
+                }
+                else if (!File.Exists(fullPath))
+                {
+                    return ViewTextAsync(fileName, $"File {fullPath} does not exist");
+                }
+            }
+            else if (!GitModule.IsValidGitWorkingDir(fullPath))
+            {
+                return ViewTextAsync(fileName, $"Invalid submodule: {fileName}");
             }
 
             return ShowOrDeferAsync(
-                fileName,
+                fullPath,
                 () => ViewItemAsync(
                     fileName,
                     isSubmodule,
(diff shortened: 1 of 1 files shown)


mstv (approved):
LGTM

GitUI/Editor/FileViewer.cs
```diff
@@ -607,14 +607,12 @@ public Task ViewFileAsync(string fileName, bool isSubmodule = false, Action? ope
             {
                 try
                 {
-                    var path = _fullPathResolver.Resolve(fileName);
-
-                    if (!File.Exists(path))
+                    if (!File.Exists(fullPath))
```
RussKie, line 610: This should be the way around - the path must be resolved via `_fullPathResolver`, i.e. resolved for the _current_ working directory, and not some arbitrary path.

RussKie (changes requested):

GitUI/Editor/FileViewer.cs
gerhardol: The full path is already resolved, no need to do it again

gerhardol:
Rebased without changes

This solves what is reported in #9213.
I have investigated som, not found similar issues
But also not found why this occurs when starting with command line arguments

GitUI/Editor/FileViewer.cs
RussKie, line 610: Let's add an assert
```suggestion
                    Debug.Assert(Path.IsPathFullyQualified(path), "Path must be resolved and fully qualified");
                    if (!File.Exists(fullPath))
```

GitUI/Editor/FileViewer.cs
```diff
@@ -625,16 +623,12 @@ public Task ViewFileAsync(string fileName, bool isSubmodule = false, Action? ope
 
             string GetFileText()
             {
-                var path = File.Exists(fileName)
-                    ? fileName
-                    : _fullPathResolver.Resolve(fileName);
-
-                if (!File.Exists(path))
+                if (!File.Exists(fullPath))
```
RussKie, line 626: ditto here

GitUI/Editor/FileViewer.cs
```diff
@@ -843,25 +837,22 @@ private void OnExtraDiffArgumentsChanged()
...
             return ShowOrDeferAsync(GetFileLength(), showFunc);
 
             long GetFileLength()
             {
                 try
                 {
-                    var resolvedPath = _fullPathResolver.Resolve(fileName);
-
-                    if (File.Exists(resolvedPath))
+                    if (File.Exists(fullPath))
```
RussKie, line 848: ditto here

RussKie (approved):
Tested locally - 👍

GitUI/Editor/FileViewer.cs
gerhardol:
```suggestion
                    if (!File.Exists(fullPath))
                    Debug.Assert(Path.IsPathFullyQualified(fullPath), "Path must be resolved and fully qualified");
```

GitUI/Editor/FileViewer.cs
gerhardol: Refactoring the code instead.

GitUI/Editor/FileViewer.cs
RussKie: No, we want to assert before checking file.

gerhardol: Made a minor refactoring of the handling with properly ordered checks (a few lines removed)

GitUI/Editor/FileViewer.cs
```diff
@@ -575,26 +576,32 @@ string GetFileTextIfBlobExists()
...
         {
             string? fullPath = _fullPathResolver.Resolve(fileName);
-
-            if (isSubmodule && !GitModule.IsValidGitWorkingDir(fullPath))
-            {
-                return ViewTextAsync(fileName, "Invalid submodule: " + fileName);
-            }
-
             Validates.NotNull(fullPath);
+            Debug.Assert(Path.IsPathFullyQualified(fullPath), "Path must be resolved and fully qualified");
```
mstv, line 580: Is `fullPath` really not `null` here now? Then its type could be (non-nullable) `string`.

gerhardol: _fullPathResolver.Resolve() is string?, but the result should not be null in our case (and null was actually OK).
ValidatesNotNull() must check the result, to follow the contract of other methods.

gerhardol:
@RussKie Do you want to review the latest changes (after your comments) or can I merge?
You set the 3.5.1 MS too, need to reformat for 3.5.

RussKie:
Yep, thank you.
I think it's worth servicing in 3.5

PR #9222, opened 2021-05-19
Git commands should never have colors

gerhardol:
A collection of some smaller issues

## Proposed changes

Align git-diff -M -C arguments

GetCurrentChangesCmd() only had the options for Index/staged files but not worktree.
This is not really an error but different from all other diff commands.

git-diff options for rename and copy of files were put in various order which made me confused.
There should be no significance in the ordering, this is a cleanup to avoid dong the investigation again
Order is set to beconsistent and more aligned with Git documentation.

Also, "-M -C" was at one occasion on separate lines, which made this harder to search for.
(The alternative is to change all options to be separated.)

--
Git commands should never have colors

Colors are excluded by default in pipe/file output
However, the user can override the settings
Most common command to override for users seem to be "diff",
why most diff commands have --no-colors arguments.
Also for range-diff
This was also set for for-each-ref, this is not needed

To ensure this is set for all commands use this as a configuration
for the Git commands instead

--
AutoCompletion for .groovy

In commit info, similar as for .cs etc

--
FileViewer: Remove unnecessary null check

Should have been removed at NRT changes

## Test methodology 

Tests updated

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -69,9 +69,11 @@ namespace GitCommands.Git.Commands
                         ? (ArgumentString)"--no-optional-locks"
                         : default)
                 {
-                    { staged, "-M -C --cached" },
-                    extraDiffArguments,
+                    "--find-renames",
+                    "--find-copies",
                     { AppSettings.UseHistogramDiffAlgorithm, "--histogram" },
+                    extraDiffArguments,
+                    { staged, "--cached" },
                     "--",
                     fileName.ToPosixPath().Quote(),
                     { staged, oldFileName?.ToPosixPath().Quote() }
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -69,7 +69,6 @@ namespace GitCommands.Git.Commands
                         ? (ArgumentString)"--no-optional-locks"
                         : default)
                 {
-                    "--no-color",
                     { staged, "-M -C --cached" },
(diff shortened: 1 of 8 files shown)


GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -69,10 +69,10 @@ public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath
                         ? (ArgumentString)"--no-optional-locks"
                         : default)
                 {
-                    "--no-color",
-                    { staged, "-M -C --cached" },
-                    extraDiffArguments,
+                    { "-M -C" },
```
pmiossec, line 72: For me, the shortcut options (with one '-') should be kept for humans in a terminal.

While you are at it, is it possible to replace it by the full option name so that reading code and also debugging could be made easier? 

I had to look at the documentation for these obscure options and it seems to be the shortcuts for '--find-renames' and '--find-copies' which is much more comprehensive, right?

RussKie:
LGTM in principle, just confused why unrelated things bundled togerher.

gerhardol:
> LGTM in principle, just confused why unrelated things bundled togerher.
> 

Because I did not want to open 4 tiny pull requests.

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: will change

gerhardol: done

mstv (approved):
LGTM, have run it for a few days

gerhardol:
@msftbot merge in 24 hours

PR #9231, opened 2021-05-23
Preparations for exceptions on git error (part 2)

mstv:
Some of the preparations for #9056 contributed by @gerhardol

## Proposed changes

- Suppress git warning for safecrlf
  occured if:
  * core.safecrlf is warn or true
  * core.autocrlf is true on Windows
  * a file stored in crlf is edited to lf
  * File is selected in Diff tab
- Let git filter for merge commits instead with own regex
- Avoid git error in Left Panel
  occurred when the ROT is initialized in the Dashboard without a working dir, of course

## Screenshots 

not affected

## Test methodology 

- NUnit tests
- manual

## Test environment(s) 

- Git Extensions 0.0.0.0
- Build b430c3f01860c5cfaeeee18e72c2a1db456a449a
- Git 2.31.1.windows.1
- .NET 5.0.6
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1116,19 +1116,11 @@ namespace GitCommands
             {
                 "--parents",
                 "--no-walk",
+                "--min-parents=2",
                 $"{startRev}..{endRev}"
             };
 
-            return _gitExecutable
-                .GetOutputLines(args)
-                .Any(IsTwoSha1Hashes);
-
-            static bool IsTwoSha1Hashes(string parents)
-            {
-                // TODO use Regex here to avoid allocations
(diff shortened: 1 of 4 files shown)


gerhardol (approved):

GitCommands/Git/GitModule.cs
```diff
@@ -1116,19 +1116,11 @@ public bool ExistsMergeCommit(string? startRev, string? endRev)
             {
                 "--parents",
                 "--no-walk",
+                "--min-parents=2",
```
gerhardol, line 1119: This option is available in Git 2.1 at least (but maybe not when the command was originally added?), so no version check required

RussKie (approved):

PR #9235, opened 2021-05-25
Use System.Windows.Forms.TaskDialog

### LINKED ISSUE #9151: Use `System.Windows.Forms.TaskDialog`
## Current behaviour

Historically we used `Microsoft.WindowsAPICodePack.Dialogs.TaskDialog`, which can now be replaced with [`System.Windows.Forms.TaskDialog`](https://docs.microsoft.com/dotnet/api/system.windows.forms.taskdialog).

DmitryZhelnin:
Fixes #9151

## Proposed changes

- use `System.Windows.Forms.TaskDialog` instead of `Microsoft.WindowsAPICodePack.Dialogs.TaskDialog`

## Test methodology 

- Almost all dialogs are tested manually

## Screenshots

| Old | New |
| :---: | :---: |
BugReportInvoker_old | BugReportInvoker_new
FormPull_old_1 | FormPull_new_1
FormPull_old_2 | FormPull_new_2
FormPull_old_3 | FormPull_new_3
FormPull_old_4 | FormPull_new_4
FormPull_old_5 | FormPull_new_5
FormCheckoutBranch_old | FormCheckoutBranch_new
FormCommit_old_1 | FormCommit_new_1 
FormCommit_old_2 | FormCommit_new_2
FormPush_old | FormPush_new
FormResolveConflict_old | FormResolveConflict_new
FormSettings_old | FormSettings_new
FormStash_old | FormStash_new
InvalidRepositoryRemover_old | InvalidRepositoryRemover_new
MessageBoxes_old | MessageBoxes_new
Program_old | Program_new
RevisionGridControl_old | RevisionGridControl_new

## Test environment(s) 

- Windows 7 SP1, DPI Scaling 125%

----

### DIFF
--- a/GitUI/CommandsDialogs/FormCheckoutBranch.cs
+++ b/GitUI/CommandsDialogs/FormCheckoutBranch.cs
@@ -393,7 +393,8 @@ namespace GitUI.CommandsDialogs
                             Verification = new TaskDialogVerificationCheckBox
                             {
                                 Text = _dontShowAgain.Text
-                            }
+                            },
+                            SizeToContent = true
                         };
 
                         messageBoxResult = TaskDialog.ShowDialog(Handle, page) == TaskDialogButton.Yes;
--- a/GitUI/CommandsDialogs/FormCheckoutBranch.cs
+++ b/GitUI/CommandsDialogs/FormCheckoutBranch.cs
@@ -13,7 +13,6 @@ using GitUI.Script;
 using GitUIPluginInterfaces;
 using Microsoft;
 using Microsoft.VisualStudio.Threading;
-using Microsoft.WindowsAPICodePack.Dialogs;
 using ResourceManager;
 
 namespace GitUI.CommandsDialogs
@@ -385,21 +384,21 @@ namespace GitUI.CommandsDialogs
                     bool? messageBoxResult = AppSettings.AutoPopStashAfterCheckoutBranch;
                     if (messageBoxResult is null)
                     {
-                        using Microsoft.WindowsAPICodePack.Dialogs.TaskDialog dialog = new()
+                        TaskDialogPage page = new()
                         {
-                            OwnerWindowHandle = Handle,
                             Text = _applyStashedItemsAgain.Text,
                             Caption = _applyStashedItemsAgainCaption.Text,
-                            Icon = TaskDialogStandardIcon.Information,
-                            StandardButtons = TaskDialogStandardButtons.Yes | TaskDialogStandardButtons.No,
-                            FooterCheckBoxText = _dontShowAgain.Text,
-                            FooterIcon = TaskDialogStandardIcon.Information,
-                            StartupLocation = Microsoft.WindowsAPICodePack.Dialogs.TaskDialogStartupLocation.CenterOwner
+                            Icon = TaskDialogIcon.Information,
+                            Buttons = { TaskDialogButton.Yes, TaskDialogButton.No },
+                            Verification = new TaskDialogVerificationCheckBox
+                            {
+                                Text = _dontShowAgain.Text
+                            }
                         };
 
-                        messageBoxResult = dialog.Show() == TaskDialogResult.Yes;
+                        messageBoxResult = TaskDialog.ShowDialog(Handle, page) == TaskDialogButton.Yes;
 
-                        if (dialog.FooterCheckBoxChecked == true)
+                        if (page.Verification.Checked)
                         {
                             AppSettings.AutoPopStashAfterCheckoutBranch = messageBoxResult;
                         }
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1218,7 +1218,8 @@ namespace GitUI.CommandsDialogs
                         Caption = _noFilesStagedCommitCaption.Text,
                         Icon = TaskDialogIcon.Error,
                         Heading = _noFilesStagedCommitInstructions.Text,
-                        Buttons = { TaskDialogButton.Cancel }
+                        Buttons = { TaskDialogButton.Cancel },
+                        SizeToContent = true
                     };
 
                     // Option 1: there are no staged files, but there are unstaged files. Most probably user forgot to stage them.
@@ -1284,6 +1285,7 @@ namespace GitUI.CommandsDialogs
                         Buttons = { TaskDialogButton.Cancel },
                         Icon = TaskDialogIcon.Error,
                         AllowCancel = true,
+                        SizeToContent = true
                     };
                     TaskDialogCommandLinkButton btnCheckout = new(TranslatedStrings.ButtonCheckoutBranch);
                     TaskDialogCommandLinkButton btnCreate = new(TranslatedStrings.ButtonCreateBranch);
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -1223,10 +1223,7 @@ namespace GitUI.CommandsDialogs
 
                     // Option 1: there are no staged files, but there are unstaged files. Most probably user forgot to stage them.
                     string text = Unstaged.IsFilterActive ? _noFilesStagedCommitAllFilteredUnstagedOption.Text : _noFilesStagedCommitAllUnstagedOption.Text;
-                    TaskDialogCommandLinkButton lnkStageAndCommit = new(text, enabled: !Unstaged.IsEmpty)
-                    {
-                        Tag = 1
-                    };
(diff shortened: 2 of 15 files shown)


GitUI/CommandsDialogs/FormCheckoutBranch.cs
```diff
@@ -385,21 +384,21 @@ private DialogResult PerformCheckout(IWin32Window? owner)
...
+                        TaskDialogPage page = new()
                         {
-                            OwnerWindowHandle = Handle,
                             Text = _applyStashedItemsAgain.Text,
                             Caption = _applyStashedItemsAgainCaption.Text,
-                            Icon = TaskDialogStandardIcon.Information,
-                            StandardButtons = TaskDialogStandardButtons.Yes | TaskDialogStandardButtons.No,
-                            FooterCheckBoxText = _dontShowAgain.Text,
-                            FooterIcon = TaskDialogStandardIcon.Information,
-                            StartupLocation = Microsoft.WindowsAPICodePack.Dialogs.TaskDialogStartupLocation.CenterOwner
```
gerhardol, line 397: `StartupLocation` is not specified in the update. Is this the default?

gerhardol (commented):
Looks fine, just tested one dialog at startup, reviewed arguments for the first files

GitUI/CommandsDialogs/FormCheckoutBranch.cs
DmitryZhelnin: Yes, `CenterOwner` is the default location

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1213,46 +1212,36 @@ bool ConfirmEmptyMergeCommit()
...
-                            lnkStageAndCommit.Enabled = false;
-                        }
+                        Tag = 2
                     };
+                    page.Buttons.Add(lnkEmptyCommit);
 
-                    var dialogResult = dialog.Show();
-                    if (dialogResult == TaskDialogResult.Cancel)
+                    TaskDialogButton result = TaskDialog.ShowDialog(Handle, page);
+                    if (result.Tag is not int)
```
RussKie, line 1244: `result` is the button instance that was clicked. So this should be something like:
```suggestion
                    if (result != lnkEmptyCommit)
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -172,22 +169,20 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
...
+                                Buttons = { btnOpenSettings, btnIgnore }
                             };
 
-                            btnOpenSettings.Click += (sender, e) => errorDialog.Close(TaskDialogResult.Yes);
-                            btnIgnore.Click += (sender, e) => errorDialog.Close(TaskDialogResult.No);
-
-                            var result = errorDialog.Show();
-                            if (result == TaskDialogResult.Yes)
+                            TaskDialogButton result = TaskDialog.ShowDialog(page);
+                            if (result.Tag is yes)
```
RussKie, line 185:
```suggestion
                            if (result == btnOpenSettings)
```

Plugins/BuildServerIntegration/AzureDevOpsIntegration/AzureDevOpsAdapter.cs
```diff
@@ -172,22 +169,20 @@ private async Task ObserveBuildsAsync(DateTime? sinceDate, bool running, IObserv
                         {
                             ProjectOnErrorKey = CacheKey;
 
-                            TaskDialogButton btnOpenSettings = new("btnOpenSettings", "Open settings");
-                            TaskDialogButton btnIgnore = new("btnIgnoreError", "Ignore");
-                            using TaskDialog errorDialog = new()
+                            const int yes = 1;
+                            TaskDialogButton btnOpenSettings = new("Open settings") { Tag = yes };
```
RussKie, line 173:
```suggestion
                            TaskDialogButton btnOpenSettings = new("Open settings");
```

RussKie (commented):
👍 

We need to set [`SizeToContent=true`](https://docs.microsoft.com/dotnet/api/system.windows.forms.taskdialogpage.sizetocontent#System_Windows_Forms_TaskDialogPage_SizeToContent) because a number of dialogs are currently squashed. E.g. as shown in https://github.com/gitextensions/gitextensions/pull/9212:
image

Also it'd be great to see screenshots of the old/new dialogs for sanity check.

DmitryZhelnin:
Made some changes in dialog result handling. Will add screenshots a bit later.

DmitryZhelnin:
Added screenshots of dialogs that I've been able to trigger

ivangrek:
Looks good.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1223,25 +1223,19 @@ bool ConfirmAndStageAllUnstaged()
...
-                    TaskDialogCommandLinkButton lnkEmptyCommit = new(_noFilesStagedMakeEmptyCommitOption.Text)
-                    {
-                        Tag = 2
-                    };
+                    TaskDialogCommandLinkButton lnkEmptyCommit = new(_noFilesStagedMakeEmptyCommitOption.Text);
                     page.Buttons.Add(lnkEmptyCommit);
 
                     TaskDialogButton result = TaskDialog.ShowDialog(Handle, page);
-                    if (result.Tag is not int)
+                    if (result == TaskDialogButton.Cancel)
```
ivangrek, line 1239: `if (result == TaskDialogButton.Cancel)`
to
`if (result is TaskDialogButton.Cancel)`

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1294,45 +1286,32 @@ void DoCommit()
...
-                    };
                     page.Buttons.Add(btnCheckout);
                     page.Buttons.Add(btnCreate);
                     page.Buttons.Add(btnContinue);
 
-                    TaskDialog.ShowDialog(Handle, page);
-
-                    switch (dialogResult)
+                    TaskDialogButton result = TaskDialog.ShowDialog(Handle, page);
+                    if (result == TaskDialogButton.Cancel)
```
ivangrek, line 1298: `if (result == TaskDialogButton.Cancel)`
to
if (result is TaskDialogButton.Cancel)

GitUI/CommandsDialogs/FormCommit.cs
DmitryZhelnin: It will not compile. `TaskDialogButton.Cancel` is not a constant value.

GitUI/CommandsDialogs/FormCommit.cs
DmitryZhelnin: It will not compile. `TaskDialogButton.Cancel` is not a constant value.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1293,61 +1277,43 @@ void DoCommit()
...
-                            if (!UICommands.StartCreateBranchDialog(this, _editedCommit?.ObjectId))
-                            {
-                                return;
-                            }
+                        return;
+                    }
 
-                            break;
-                        case -1:
+                    if (result == btnCheckout)
```
gerhardol, line 1303: I prefer a case with this many options.
In a few other places too

gerhardol (commented):
+1
Code and screenshots look good
No detailed review on the last 3/4 of the code

GitUI/CommandsDialogs/FormCommit.cs
DmitryZhelnin: Do you want to use separate variable for dialog result, like `int dialog result = -1` or maybe to use `Tag` property of each button? I thought that it would be good to use the same approach in all places. For me switch/case with values like `1` , `2` is not very intuitive. On the other hand the `if` statements with checks like `result == btnCheckout` is more self explanatory.
But I don't mind to make changes. What option is more preferable - `int` variable or `Tag` ?

gerhardol: No change requested. 
It is better to avoid the additional tags etc. In the back of my mind I was thinking about C#8 switch expressions, but that is still requiring constants...
Good work.

GitUI/CommandsDialogs/FormCommit.cs
RussKie: Are you sure? We designed it to allow this scenario

GitUI/CommandsDialogs/FormCommit.cs
RussKie: IMO `is null` is ok, for everything else it looks awkward.

RussKie (approved):

RussKie:
Thank you

Issue #9229, opened 2021-05-23
Reduce Git commands


gerhardol:
## Feature description

A number of Git commands can be combined or made more efficient to reduce execution time.
Some commands are remarkably slowed down by virus programs like Symantec, `git status` can be at least seven times slower.

A command like `git for-each-ref` can take about 300 ms with virus programs in repos at work, while the same command at home with  a faster PC and GE repo requires 30 (but occasionally 120 ms) at home.
Also `git for-each-ref` requires about the same time regardless of the filter in the Git command so a command combining fetching branches/remotes/tags at the same time and filter in code is almost three times faster than three separate Git calls.

This is currently a brief list of what I recall right now, I suggest the list is improved (add comments if you are not a member of the project and can edit the list).

### Loading

Affects the time to load or refresh a repo

* [x] #9246 `git --version` is called twice in the startup (to check version and to check existence)
* [x] `git for-each-ref` is called three times to load the sidepanel, one is enough
* [x] #9317 `git log` uses a GUID to separate branch from further data. A double \0\0 is probably enough, but the separation of commits uses null termination right now. (there were discussions in the PR where 1DEA7CC4-FB39-450A-8DDF-762FCEA28B05 was introduced). 
* [x] #9243 `git log` uses Git to separate and encode the the header and body, but still have special handling for the handling. There was also an issue some years ago that stated that the Git handling was incompatible with the GE internal handling. It is therefore maybe better to load the raw commit message body and process that. The processing can be improved in general.

Caching can be improved (switching commits)

* [ ] `git status` could be shared for GitStatusMonitor and for RevDiff listing (and GitStatusMonitor is aware of when the cache is potentially changed).
* [ ] `git diff` for file work tree items could be cached, invalidated by GitStatusMonitor
* [ ] `git for-each-ref` is needed by many different entities and could be cached, invalidated by GitStatusMonitor.

### Background

Not directly affecting interactive commands but CPU load.

* [x] #9256 `git ls-tree` lists superproject branches in subrepos. This is configurable, if enabled this loads a large number of Git commands. (Very annoying when viewing the Git Command Log, not even cancelled, the listing goes on for a minute or so). All of these are not interesting, should be limited to a number of superproject branches (suggest a setting that only can be read right now, so the value can be adjusted in the settings file if there is a need to tweak it).
[...]

Issue #6139, opened 2019-01-20
Git log parsing with -z --name-only arguments


vbjay:
@drewnoakes  Mind taking a look?
## Current behaviour

--name-only and -z create log output like the following.
![image](https://user-images.githubusercontent.com/3629489/51441678-aed4ba00-1ca1-11e9-8a78-92257793e04a.png)
Note the red dots.  That is where the stream reading stops the chunk to process one commit log entry.

![image](https://user-images.githubusercontent.com/3629489/51443932-adfe5100-1cbe-11e9-9ac4-a4ff4942450b.png)
Note how my OtherOutput property only contains the first file and not all files in the commit.  

![image](https://user-images.githubusercontent.com/3629489/51443958-f61d7380-1cbe-11e9-85f4-18a639cbbe1e.png)

## Expected behaviour
Handle --name-only output correctly.  I am adding a property to Revision that will contain the additional output of any git command (planning for the future of any additional output from any git command) in #6132.  We need to handle null vs double null characters in the stream.

## Steps to reproduce

https://github.com/gitextensions/gitextensions/blob/258b651af120cf49c250c5832bdc1b8d737120e5/GitCommands/StreamExtensions.cs#L12  Debug code here.
## Did this work in previous version of GitExtensions

Nope

## Environment

Master

- Git Extensions 3.01.00.0
- Build 258b651af120cf49c250c5832bdc1b8d737120e5 (Dirty)
- Git 2.20.1.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3260.0
- DPI 96dpi (no scaling)

## Diagnostics

```bash

git -c log.showSignature=false log master -z --pretty=format:"%H%T%P%n%at%n%ct %n%e%n%aN%n%aE%n%cN%n%cE%n%s%n%n%b" --name-only -n3 --no-merges>history.log
```
[history.log](https://github.com/gitextensions/gitextensions/files/2776787/history.log)
Examine history.log in notepad++ showing all characters or your favorite hex editor.

drewnoakes:
As you've identified, that parsing code assumes that null bytes separate commits in the response stream.

Is it not possible to reformulate the git command to preserve that assumption?

If not, please be mindful of performance. This was done with the goal of keeping revision parsing light and fast with minimal data copying.

drewnoakes:
Just seen your PR.

vbjay:
[...]

PR #9243, opened 2021-05-28
RevisionReader: Parse raw commit body

gerhardol:
Part of #7635 and #9229
                                      
Parse raw body %B instead of combining %s%b to avoid extra processing
in both Git and GE.

## Proposed changes

See #7635 and #9229 for background.
This decreases the time to list Git revisions with a few percent, as well as cleaning up the code and tests.

## Test methodology 

Tests were adapted

----


RussKie (commented):
👍

GitCommands/RevisionReader.cs
```diff
@@ -448,32 +445,40 @@ DateTime ParseUnixDateTime()
...
-                return (body: subject, additionalData: null);
+                // TODO log this error
+                Debug.Fail("Subject is empty in the log -- this should not happen");
+                return (subject: null, body: null, additionalData: null);
             }
 
-            string tail = reader.ReadToEnd() ?? "";
-            int indexOfEndOfBody = tail.LastIndexOf(EndOfBody, StringComparison.InvariantCulture);
+            // Ignore leading empty or whitespace lines, see Git pretty.c
+            string tail = reader.ReadToEnd().TrimStart() ?? "";
```
drewnoakes, line 458: Is there a way to prevent allocating this tail string? Perhaps by extending the reader? For large repositories, this may create a lot of objects for the GC.

Eg:

reader.SkipWhiteSpace();
var body = reader.ReadTo(EndOfBody);

GitCommands/RevisionReader.cs
```diff
@@ -52,8 +52,7 @@ public sealed class RevisionReader : IDisposable
               /* Author email    */ "%aE%n" +
               /* Committer name  */ "%cN%n" +
               /* Committer email */ "%cE%n" +
-              /* Commit subject  */ "%s%n%n" +
-              /* Commit body     */ "%b" + EndOfBody;
+              /* Commit subject  */ "%B" + EndOfBody;
```
drewnoakes, line 55: Comment is now a little misleading. This is the body too, no?

drewnoakes (approved):

RussKie:
We can now use Span<char> and ROS<char>, and avoid allocations.

GitCommands/RevisionReader.cs
gerhardol: Will change to raw body

GitCommands/RevisionReader.cs
gerhardol: For end of body, I would like to remove that with a single null, and read additional data separately. Not finished yet. 

This implementation should have the same allocations as the current.

GitCommands/RevisionReader.cs
gerhardol: Also updated a few other comments

GitCommands/RevisionReader.cs
gerhardol: Changing to something like the suggested to simplify further, will simplify updated parsing of the reader. 

ParseCommitBody() is no longer needed, removing it and the tests for it. 
This may be be replaced with parsing the git-log output instead eventually.

gerhardol:
Will add a few more changes to not allocate commit body that need to be garbage collected, maybe change endofbody too

gerhardol:
Preliminary improvements: Parsing about half time from before
I just ran a few times, noting "typical" times, no deep analysis.
It depends a little on when GC runs, should probably run the test at least 10 times.
But the difference is significant.

GE: 1.82-2.22s-> 0.99-1.12s
Linux: 12.0-13.0s -> 3.9-5.82s

drewnoakes (approved):

GitCommands/RevisionReader.cs
drewnoakes: Nice.

mstv:
🎉 The loading times feel to have improved visibly. I will review in the next days.

gerhardol:
Two minor updates:

* Only store Body if different from Subject.
This makes no major difference for load times and memory usage. Maybe the string interning is helping here.
However, the code is clearer to me.
If this is not changed, the situations where Body is null (old commit where it is not stored) and MultiLine is false should be handled specially, no need to fetch the commit body.
See CommitInfo.cs:329, UpdateCommitMessage()

* Align use of %B vs %s%b and trimming subject/body
No real functional change

@mstv Can I squash before your review?

mstv:
> Can I squash before your review?

Yes, it is OK for me.

gerhardol:
Two first commits are same as #9253 
Last deals with other call sites and will be squashed
Previous commits are available in `tmp/log-zero-0`

mstv (commented):
Looks sensible.
A few existing NUnit test need to be adapted.

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,82 @@ public StringLineReader(string s)
...
+                {
+                    _index++;
+                }
+
+                return this;
+            }
+
+            private void TrimEnd(ref int endIndex)
+            {
+                while (endIndex > _index && char.IsWhiteSpace(_s[endIndex - 1]))
```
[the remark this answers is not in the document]
gerhardol: This is a private method and endIndex is found in _s, so the check will not do anything.
(I made it private instead of a submethod to `ReadToEndOfBody()` to keep it next to TrimStart()

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,82 @@ public StringLineReader(string s)
...
+            }
+
+            public string? ReadToEnd()
+            {
+                if (_index >= _s.Length)
+                {
+                    return null;
+                }
+
+                return _s.Substring(_index);
```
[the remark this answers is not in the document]
gerhardol: I optimized that, but it is clearer, will change

GitCommands/RevisionReader.cs
```diff
@@ -33,27 +33,27 @@ public enum RefFilterOptions
 
     public sealed class RevisionReader : IDisposable
     {
-        private const string EndOfBody = "1DEA7CC4-FB39-450A-8DDF-762FCEA28B05";
+        private const char EndOfBody = (char)0x3;
```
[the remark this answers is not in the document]
gerhardol: additionalData is just filenames so the change should be safe
But the split should maybe be done in line 386 instead, encode additional privately (are evn filenames encoded here?)

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,82 @@ public StringLineReader(string s)
...
-                    _index = _s.Length;
+                    endSubjectIndex = endBodyIndex;
                 }
+                else
+                {
+                    TrimEnd(ref endSubjectIndex);
+                }
+
+                // subject can also be defined as the contents before empty line (%s for --pretty),
+                // this uses the the alternative definition of first line in body.
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,82 @@ public StringLineReader(string s)
...
+                var subject = _s.Substring(_index, endSubjectIndex - _index);
+                var body = skipBody || endSubjectIndex == endBodyIndex
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,82 @@ public StringLineReader(string s)
...
+                    ? null
+                    : _s.Substring(_index, endBodyIndex - _index);
+                bool hasMultiLineMessage = endSubjectIndex < endBodyIndex;
```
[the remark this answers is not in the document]
gerhardol: I had already changed that in a fix (I intended to change that earlier)

GitCommands/RevisionReader.cs
```diff
@@ -403,27 +392,21 @@ DateTime ParseUnixDateTime()
             var committer = reader.ReadLine(stringPool);
             var committerEmail = reader.ReadLine(stringPool);
 
-            var subject = reader.ReadLine(advance: false);
+            // We keep full multiline message bodies within the last six months.
+            // Note also that if body and subject are identical (single line), the subject never need to be stored
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
```diff
@@ -407,6 +409,7 @@ void DrawSuperProjectRef(string label, ref int currentOffset)
 
             if (lines.Length == 1)
             {
+                indicator.Render();
                 return;
             }
```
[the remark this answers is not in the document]
gerhardol: ok
Also in #9253

GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
```diff
@@ -101,11 +101,13 @@ public override bool TryGetToolTip(DataGridViewCellMouseEventArgs e, GitRevision
...
-                var bodySummary = _gitRevisionSummaryBuilder.BuildSummary(revision.Body);
-                var initialLength = (bodySummary?.Length ?? 50) + 10;
+                // The body is not stored for older commits (to save memory)
+                string bodySummary = _gitRevisionSummaryBuilder.BuildSummary(revision.Body)
+                    ?? revision.Subject + (revision.HasMultiLineMessage ? TranslatedStrings.BodyNotLoaded : "");
+                int initialLength = bodySummary.Length + 10;
                 _toolTipBuilder.EnsureCapacity(initialLength);
 
-                _toolTipBuilder.Append(bodySummary ?? revision.Subject + TranslatedStrings.BodyNotLoaded);
+                _toolTipBuilder.Append(bodySummary);
```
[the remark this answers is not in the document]
gerhardol: I find this way better, will not change unless requested

GitCommands/Git/GitModule.cs
```diff
@@ -950,11 +950,13 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
...
-                string message = ProcessDiffNotes(10);
+                string message = ProcessDiffNotes(startIndex: 10);
 
                 // commit message is not re-encoded by git when format is given
-                revision.Body = ReEncodeCommitMessage(message, revision.MessageEncoding);
-                revision.Subject = revision.Body?.Substring(0, revision.Body.IndexOfAny(new[] { '\r', '\n' })) ?? "";
+                // See also RevisionReader for parsing commit body
+                revision.Body = ReEncodeCommitMessage(message, revision.MessageEncoding).TrimStart().TrimEnd();
+                int endSubjectIndex = revision.Body.IndexOf('\n');
+                revision.Subject = endSubjectIndex < 0 ? revision.Body : revision.Body.Substring(0, endSubjectIndex).TrimEnd();
```
mstv, line 959:
```suggestion
                int? endSubjectIndex = revision.Body?.IndexOf('\n');
                revision.Subject = endSubjectIndex is not >= 0 ? revision.Body : revision.Body.Substring(0, endSubjectIndex).TrimEnd();
```

GitCommands/RevisionReader.cs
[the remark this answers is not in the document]
mstv: Unicode filenames could contain bytes with this value.
No need for a guid but perhaps `"\x03\x03\x03\x03"`.

UnitTests/GitUI.Tests/UserControls/RevisionGrid/Graph/LaneInfoProviderTests.cs
```diff
@@ -234,7 +234,7 @@ public void GetLaneInfo_should_display_the_subject_if_singleline_body_null()
             _realCommitNode.GitRevision.HasMultiLineMessage = false;
             _laneNodeLocator.FindPrevNode(Arg.Any<int>(), Arg.Any<int>()).Returns(x => (_realCommitNode, isAtNode: false));
 
-            GetLaneInfo_should_display(_realCommitNode, suffix: _realCommitNode.GitRevision.Subject);
+            GetLaneInfo_should_display(_realCommitNode);
```
mstv, line 237: No, the `LaneInfoProvider` must be adapted.

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +466,86 @@ public StringLineReader(string s)
...
+                if (_index >= _s.Length)
                 {
-                    _index = _s.Length;
+                    return null;
                 }
 
-                return s;
+                _index = _s.Length;
+
+                return _s.Substring(_index);
```
mstv, line 548: The substring must be taken before setting `_index = _s.Length`. The latter must be done for the `return null`, too.

mstv (commented):
There seems to be a need for NUnit tests for the `StringLineReader`.

GitCommands/Git/GitModule.cs
gerhardol: It seems like `ReEncodeCommitMessage()` is incorrectly annotated.
A few other changes are required too.

GitCommands/RevisionReader.cs
gerhardol: One reason I removed the advance of index here.
Renaming to `PeakToEnd()`, reversing the advancing

UnitTests/GitUI.Tests/UserControls/RevisionGrid/Graph/LaneInfoProviderTests.cs
gerhardol: Implementation in `LaneInfoProvider()` is OK. 
The difference now is that even if Body is null, Subject is used if `HasMultiLineMessage == false`

mstv: I will check myself when MS VS works again.

GitCommands/RevisionReader.cs
mstv: I agree, but `PeekToEnd`.

gerhardol:
> Unicode filenames could contain bytes with this value.
No need for a guid but perhaps "\x03\x03\x03\x03".

It turns out that using a string to separate is considerably slower than a char. The total command is more than twice as long if a string is searched instead of a char like `_s.LastIndexOf((char)3`
To fix this, the parsing is aware of if there is a path filter or not 
The performance seem toimprove some tenths of percent further by this.

> There seems to be a need for NUnit tests for the `StringLineReader`.

The test should be for `TryParseRevision()` so not the tests are useless again at next refactoring.

I did not intend to make a change on the internals and originally just adopted the tests (the old tests were incorrect though). But the benefit in ditching the old implementation and avoid allocations were so big that I could not resist. I hate writing new tests though...

gerhardol:
> It turns out that using a string to separate is considerably slower than a char. The total command is more than twice as long if a string is searched instead of a char like `_s.LastIndexOf((char)3`
> To fix this, the parsing is aware of if there is a path filter or not
> The performance seem toimprove some tenths of percent further by this.

@RussKie 3.5 is has less execution time and less memory usage also compared to the optimized version in this PR. (This PR is maybe 25% worse than 3.5, master 400% worse). 
So there is a major regression with .NET5, not just for `LastIndexOf()`.

I was looking into backporting the change to not add endOfBody marker also to 3.5 but there is no major reason...

GitCommands/Git/GitModule.cs
```diff
@@ -950,11 +950,15 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
...
             else
             {
-                string message = ProcessDiffNotes(10);
+                string message = ProcessDiffNotes(startIndex: 10);
 
                 // commit message is not re-encoded by git when format is given
-                revision.Body = ReEncodeCommitMessage(message, revision.MessageEncoding);
-                revision.Subject = revision.Body?.Substring(0, revision.Body.IndexOfAny(new[] { '\r', '\n' })) ?? "";
+                // See also RevisionReader for parsing commit body
+                revision.Body = ReEncodeCommitMessage(message, revision.MessageEncoding)?.TrimStart().TrimEnd();
```
RussKie, line 957: Why not 
```suggestion
                revision.Body = ReEncodeCommitMessage(message, revision.MessageEncoding)?.Trim();
```

Also why not trim inside `ReEncodeCommitMessage`?

GitCommands/Git/GitModule.cs
```diff
@@ -950,11 +950,15 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
...
+                int endSubjectIndex = revision.Body?.IndexOf('\n') ?? -1;
+                revision.Subject = endSubjectIndex >= 0
+                    ? revision.Body.Substring(0, endSubjectIndex).TrimEnd()
+                    : revision.Body ?? "";
```
RussKie, line 961: We now have access to `ReadOnlySpan<char>`, so it would be significantly more performant to use it and splice as necessary. E.g. pseudo code:
```suggestion
                // See also RevisionReader for parsing commit body
                string body = ReEncodeCommitMessage(message, revision.MessageEncoding); // don't return null, return "" instead
                revision.Body = body;
                
                ReadOnlySpan<char> span = body.AsSpan();
                int endSubjectIndex = span.IndexOf('\n');
                revision.Subject = endSubjectIndex >= 0
                    ? span.Substring(0, endSubjectIndex).TrimEnd()
                    : "";
```

GitCommands/RevisionReader.cs
```diff
@@ -511,34 +470,101 @@ public StringLineReader(string s)
...
-                }
+                _index = endIndex + 1;
 
                 return pool is not null
                     ? pool.Intern(_s, startIndex, endIndex - startIndex)
                     : _s.Substring(startIndex, endIndex - startIndex);
             }
 
-            public string? ReadToEnd(bool advance = true)
+            public StringLineReader TrimStart()
```
RussKie, line 484: This should be leveraging .NET implementations instead of our custom ones (e.g. `ROS<char>`).
/cc: @drewnoakes

GitUI/CommandsDialogs/FormVerify.LostObject.cs
```diff
@@ -112,7 +112,7 @@ private LostObject(LostObjectType objectType, string rawType, ObjectId objectId)
                     {
                         result.Author = module.ReEncodeStringFromLossless(logPatternMatch.Groups[1].Value);
                         string encodingName = logPatternMatch.Groups[2].Value;
-                        result.Subject = module.ReEncodeCommitMessage(logPatternMatch.Groups[3].Value, encodingName);
+                        result.Subject = module.ReEncodeCommitMessage(logPatternMatch.Groups[3].Value, encodingName) ?? "";
```
RussKie, line 115: `ReEncodeCommitMessage` mustn't return `null`.

RussKie:
> @RussKie 3.5 is has less execution time and less memory usage also compared to the optimized version in this PR. (This PR is maybe 25% worse than 3.5, master 400% worse).
> So there is a major regression with .NET5, not just for `LastIndexOf()`.

If you think there's a regression please raise an issue in https://github.com/dotnet/runtime

GitUI/CommandsDialogs/FormVerify.LostObject.cs
gerhardol: The method was previously not annotated `string?` but returned `ReEncodeStringFromLossless()` that returns `string?`. Some calls assumed `string?`, other not.

Do you want to change the implementation in `ReEncodeCommitMessage()` to return `?? ""`?

GitCommands/Git/GitModule.cs
gerhardol: This is not performance critical, but can change.
Adopting slightly (for instance not changing `ReEncodeCommitMessage()` - yet at least)

GitCommands/Git/GitModule.cs
gerhardol: That affects a number of other calls too. I believe this is OK, changing.
It though means that GE _never_ present commit body as entered by the user.
But it is better that handling is consistent, this is what this part of the PR is all about.

gerhardol:
> > @RussKie 3.5 is has less execution time and less memory usage also compared to the optimized version in this PR. (This PR is maybe 25% worse than 3.5, master 400% worse).
> > So there is a major regression with .NET5, not just for `LastIndexOf()`.
> 
> If you think there's a regression please raise an issue in https://github.com/dotnet/runtime

I do not have a good testcase for the general problem.
`LastIndexOf()` is quite clear to me, but I assume that refering to this PR is still not good enough.

Will not report anything right now.

GitCommands/RevisionReader.cs
gerhardol: I implemented this in RevisionReader, see branch `tmp/log-zero-0` (tests are not adapted)
However, it seem to be slower. linux repo loads in 6-7 seconds instead of 5-6 seconds.

This could be due to the adaptions I made to StringPool. This implementation could maybe be replaced with https://docs.microsoft.com/en-us/windows/communitytoolkit/high-performance/stringpool

mstv:
There is a nasty "%" being appended (both by 128949f and 90a9040):

grafik

UnitTests/GitUI.Tests/UserControls/RevisionGrid/Graph/LaneInfoProviderTests.cs
mstv: You are right.

vbjay:
End of body is needed.

https://github.com/gitextensions/gitextensions/issues/6139

gerhardol:
> End of body is needed. #6139

EndOfBody is included with path filter (and could be included with --name-only too if that was in RevisionReader).
In the normal case it is not needed (and will affect performance also if the .NET5 issue i solved).

mstv:
FileHistory / Blame seems to be broken.

vbjay:
File history was my point. They provide file names at the end. End of body
IS required.

gerhardol:
> There is a nasty "%" being appended (both by 128949f and 90a9040)

Fixed

>                    FileHistory / Blame seems to be broken.

Worked for me, but there are artifacts, the body was not parsed correctly.
I do not get (Last)IndexOf working with a string of unprintable right now, something else need to be done.
As EndOfBody is only used when there is a file filter, this is not performance critical, so I leave it at this for now.

I suggest that this is rebased and squashed again...

mstv:
> I suggest that this is rebased and squashed again...

Yes, please.

gerhardol:
> > I suggest that this is rebased and squashed again...
> 
> Yes, please.

Fixed EndOfBody in tests, not changed in the commit yesterday:

Summary of open items as I interpret it:

* mstv wants more tests
* RussKie wants to use span in RevisionReader pushed to tmp/log-zero-0. Seem to slow down though, maybe due to my adaption in StringPool. Can we use high-performance toolkit instead?
* RussKie wants to change ReEncodeCommitMessage

The .NET5 implementation is significantly slower than .NET4 also with these optimizations. As the changes makes sense to me regardless, I suggest that is a later investigation.
(It seems like very few people actively use master right now or this would have been reported.)
This piece of code is probably the most important to optimize, it is a large part of the time opening a repo.

gerhardol:
I have tried `Microsoft.Toolkit.HighPerformance` `StringPool`, pushed to `tmp/log-zero-0`.
It is impossible to say which handling is the most efficient, The time varies too much.
Also if a GC is forced prior to making a refresh in the repo, the time `TryParseRevision()` requires to parse a repo varies from 2.4s to 4.3s over 6 refreshes, so it is not possible to determine the efficiency unless a very large number of refreshes are done.

If something, the current implementation seem to have lower low values compared to using ReadOnlySpan and `Microsoft.Toolkit.HighPerformance`. ReadOnlySpan with current StringPool seem to be slowest.
.NET5 is much slower too.
ReadOnlySpan should be faster so going that way should be a change for the better in the long run.

@drewnoakes any thoughts about the optimizations?

drewnoakes:
Sorry @gerhardol it's hard to say without taking a deeper look and I don't have time to dedicate to this right now. I wonder if the variability comes from git itself. Perhaps we can exclude that by having a perf test that consumes ready made data, so we are only measuring our code. That is, pipe the output of git to a byte array in memory, then use that in GE for testing/benchmarking (rather than streaming from the git process). I would hope that would reduce the variability and make it easier to see where the overhead is on the GE side.

gerhardol:
> Sorry @gerhardol it's hard to say without taking a deeper look and I don't have time to dedicate to this right now. I wonder if the variability comes from git itself. Perhaps we can exclude that by having a perf test that consumes ready made data, so we are only measuring our code. That is, pipe the output of git to a byte array in memory, then use that in GE for testing/benchmarking (rather than streaming from the git process). I would hope that would reduce the variability and make it easier to see where the overhead is on the GE side.

The time intervall is for `TryParseRevision()` only (start/stop StopWatch), so Git is not running and other GE activity should have minimal impact (assuming `subject.OnNext()` is not doing too much async which may be incorrect will try to just ignore that).
It is likely GC that causes the variability, but I do not know anything else than to force a GC before parsing the first parsing.
Maybe Profiler 'ticks' is more accurate to estimate what is taking time, that will ignore the GC.

I have identified a few other tweaks using Performance profiler (inlining, time parsing etc). But that is not changing much.

GitUI/CommandsDialogs/FormVerify.LostObject.cs
RussKie: I'd look at changing the implementation of `ReEncodeCommitMessage` to never return `null` if inputs aren't `null`. May be we should add an attribute [`NotNullIfNotNull`](https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/nullable-analysis#specify-conditional-post-conditions-notnullwhen-maybenullwhen-and-notnullifnotnull)?

GitCommands/RevisionReader.cs
RussKie: First of all I'd look at rewriting the parsing logic to use ROS<char>, and then benchmark it. Then would add `StringPool`, and benchmark it again.
I'm finding it difficult to believe that our current implementation is less performant than with the use of ROS, but I haven't studied the code deeply enough to be sure.

RussKie: This optimisation work is somewhat orthogonal to this change though.

RussKie:
> It is impossible to say which handling is the most efficient, The time varies too much.

The best way to measure is to write benchmarks with https://github.com/dotnet/BenchmarkDotNet/.

RussKie:
I'm too very busy with work, and struggling to dedicate enough time to this atm. We can take the current change as is (if it is stable), and look at optimisations in the future.

drewnoakes:
> It is likely GC that causes the variability

You could verify this with PerfView. Use it to "Collect" a trace while you run the test. You'll be able to see all GC events. PerfView is not very friendly to use, but it's very powerful.

gerhardol:
> We can take the current change as is (if it is stable), and look at optimisations in the future.

I find this stable (if the code runs without hitting error exits a testcase would not find much more - the test cases in master is incorrect too.
I will try to add new test cases but have failed so far.
A benchmark would likely give the same problem as the actual run, so not attempted.

Using Diagnostic Tools sampling gives less variance than only StopWatch for TryParseRevision.
That seem to confirm that the use of ReadOnlySpan and HighPerformance StringBuffer decreases execution time so I go with that.

Identified a change that improve 15-30% (for TryParseRevision() runtime): Instead of converting commit unix time to DateTime when parsing do it when reading info for the commits.
The creation of the DateTime objects is not that bad (it requires allocation though), it is the calculation of LocalTime.
Not too much changes in other modules.
This part could be separated, but merges will be a pain.

ReadOnlySpan makes a minor improvement, not statistically proven but it should be better.
Surprisingly, it seems like ROS in the core function is more effective than in StringLineReader

With the total changes, the performance in master with these optimizations is comparable to 3.5
(So 3.5 has a lot more potential...)

The major hotspot with around 1/3 of the parsing is to parse ObjectId. I do not see much improvement possible there.
The second hotspot with around 1/4 is decoding of the output, which is hard to avoid.
The third hotspot is the StringPool and creating strings in general.

The two first commits in the PR is now included in #9300 (as they have been merged in 3.5)

RussKie (approved):

GitCommands/RevisionReader.cs
```diff
@@ -351,13 +344,12 @@ DateTime ParseUnixDateTime()
...
                     {
-                        return DateTimeUtils.UnixEpoch.AddTicks(unixTime * TimeSpan.TicksPerSecond).ToLocalTime();
+                        return unixTime;
                     }
 
                     unixTime = (unixTime * 10) + (c - '0');
                 }
             }
-
             #endregion
```
mstv, line 353: Nit: restore empty line before `#endregion`

GitCommands/RevisionReader.cs
```diff
@@ -131,33 +135,19 @@ public sealed class RevisionReader : IDisposable
                 {
                     token.ThrowIfCancellationRequested();
 
-                    if (TryParseRevision(module, chunk, stringPool, logOutputEncoding, out var revision))
+                    if (TryParseRevision(module, chunk, logOutputEncoding, sixMonths, hasPathFilter, out var revision))
                     {
-                        if (revisionPredicate is null || revisionPredicate(revision))
```
mstv, line 136: `revisionPredicate` is not used / applied.

GitCommands/RevisionReader.cs
```diff
@@ -246,34 +237,35 @@ private static bool TryParseRevision(GitModule module, ArraySegment<byte> chunk,
...
             {
                 revision = default;
                 return false;
             }
 
-            var array = chunk.Array;
-            var offset = chunk.Offset + (ObjectId.Sha1CharCount * 2);
-            var lastOffset = chunk.Offset + chunk.Count;
+            var offset = ObjectId.Sha1CharCount * 2;
+            var lastOffset = chunk.Count;
```
mstv, line 262: Should be renamed to `endOffset` in a separate commit.

GitCommands/RevisionReader.cs
```diff
@@ -338,10 +330,11 @@ int CountParents(int baseOffset)
             #region Timestamps
 
             // Lines 2 and 3 are timestamps, as decimal ASCII seconds since the unix epoch, each terminated by `\n`
-            var authorDate = ParseUnixDateTime();
-            var commitDate = ParseUnixDateTime();
+            int startIndex = offset;
```
mstv, line 333: `startIndex` is not used, is it?

GitCommands/RevisionReader.cs
```diff
@@ -33,28 +34,29 @@ public enum RefFilterOptions
...
+            /* Author name     */ "%aN%n" +
+            /* Author email    */ "%aE%n" +
+            /* Committer name  */ "%cN%n" +
+            /* Committer email */ "%cE%n" +
+
+            // Raw body, followed by additional data if path filter
+            /* Commit raw body */ "%B";
 
+        // Separate commit message body and filtered file names
+        private const string EndOfBody = "1DEA7CC4-FB39-450A-8DDF-762FCEA28B05";
```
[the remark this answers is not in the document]
gerhardol: OK

GitCommands/RevisionReader.cs
```diff
@@ -246,34 +237,35 @@ private static bool TryParseRevision(GitModule module, ArraySegment<byte> chunk,
             // at the beginning of the chunk. The latter part of the chunk will require
             // decoding as a string.
 
-            if (chunk.Count == 0)
+            ReadOnlySpan<byte> array = chunk.AsSpan();
```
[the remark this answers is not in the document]
gerhardol: OK

GitCommands/RevisionReader.cs
gerhardol: checked the file

GitCommands/RevisionReader.cs
```diff
@@ -395,19 +395,23 @@ long ParseUnixDateTime(ref ReadOnlySpan<byte> array)
...
                 revision = default;
                 return false;
             }
 
+            // We keep a full multiline message body within the last six months.
+            // Note also that if body and subject are identical (single line), the subject never need to be stored
+            if (sixMonths > authorUnixTime || !hasMultiLineMessage)
+            {
+                body = null;
+            }
```
mstv: Why have you switched back to GC'ing the `body`?

GitCommands/RevisionReader.cs
```diff
@@ -476,57 +480,41 @@ public StringLineReader(ReadOnlySpan<char> s)
...
-                    hasMultiLineMessage = true;
-                    subject = s.Slice(0, lengthSubject).TrimEnd().ToString();
-                }
-
-                // Skip body if identical to subject
-                string? body = skipBody || !hasMultiLineMessage
-                    ? null
-                    : _s.Slice(_index, lengthBody).ToString();
+                int lengthSubject = bodySlice.IndexOf('\n');
+                bool hasMultiLineMessage = lengthSubject > 0;
```
mstv, line 514: ❓
```suggestion
                bool hasMultiLineMessage = lengthSubject >= 0;
```

GitCommands/RevisionReader.cs
gerhardol: reapplied

GitCommands/RevisionReader.cs
gerhardol: why separate?
applied now

GitCommands/RevisionReader.cs
gerhardol: removed
(I had implemented comparing the time slices to optimize, removed it to keep complexity down but missed this)

GitCommands/RevisionReader.cs
gerhardol: It This was to simplify, but I will revert so my previous version...

GitCommands/RevisionReader.cs
gerhardol: Yes, formally better

GitCommands/RevisionReader.cs
```diff
@@ -395,23 +398,21 @@ long ParseUnixDateTime(ref ReadOnlySpan<byte> array)
             var committer = reader.ReadLine();
             var committerEmail = reader.ReadLine();
 
-            (string? subject, string? body, bool hasMultiLineMessage) = reader.ReadToEndOfBody(hasPathFilter);
+            bool skipBody = sixMonths > authorUnixTime;
+            (string? subject, string? body, bool hasMultiLineMessage) = reader.ReadToEndOfBody(hasPathFilter, skipBody);
+
+            // We keep a full multiline message body within the last six months.
+            // Note also that if body and subject are identical (single line), the subject never need to be stored
```
mstv, line 405:
```suggestion
            // Note also that if body and subject are identical (single line), the body never need to be stored
```

GitCommands/RevisionReader.cs
```diff
@@ -395,23 +398,21 @@ long ParseUnixDateTime(ref ReadOnlySpan<byte> array)
...
+            skipBody = skipBody || !hasMultiLineMessage;
 
-            if (author is null || authorEmail is null || committer is null || committerEmail is null || subject is null || body is null)
+            if (author is null || authorEmail is null || committer is null || committerEmail is null || subject is null || (!skipBody && body is null))
```
mstv, line 408: ❓
```suggestion
            if (author is null || authorEmail is null || committer is null || committerEmail is null || subject is null || (skipBody != (body is null)))
```

mstv (commented):
Yet another two nits

GitCommands/RevisionReader.cs
gerhardol: This is on the limit for pattern also for me, but changing as I assume I will prefer this in some time...

GitCommands/RevisionReader.cs
gerhardol: OK
(Also added additional comment on line 520)

mstv:
👍 Seems to work well for me. Rebase will solve the false translation issue. Feel free to squash.

gerhardol:
> 👍 Seems to work well for me. Rebase will solve the false translation issue. Feel free to squash.

Kept the storing of dates separate, squashed the other changes.
(Two lines changed, cleanups.)

I am struggling with creating tests for TryParseRevision, see if I succeed.

mstv (approved):

RussKie:
Ready to merge?

gerhardol:
> Ready to merge?

I am writing tests...
(So basically, not being productive at all the last week.)
But I got a grip on how to do it yesterday, will add something.

gerhardol:
A few minor tweaks, preparations for testability and some tests for TryParseRevision()
Some special care for this section, this is probably the most important to optimize for performance.
This involves minimizing calling methods.

For the testability, I would have preferred to put TryParseRevision() in an internal class. It would add a class method call but skip adding a few variables on the stack so the same performance.
To split RevisionReader.cs is not a good option as the parsing must follow the arguments. It could be possible to add  the variables as private variables but the scope should be limited.
I find the current implementation the best compromise.

The tests are not great but relative complete. The code was rewritten some to make the tests pass in teorethic failures (but the flow is improved).
However, the tests fail on AppVeyor due to carrige return issues. 
I cannot figure out why, the tests look the same as similar Approval tests.
(The Json serialization of the objects is strange too, like that the actual sha in `ObjectId` is not included...)
Worst case, the tests could be ignored in AppVeyor.
(I have tried updating ApprovalTests.NET to the latest version, still fails.)

I have not done a formal benchmark.

The parsing when there is a pathFilter is not great (like parsing errors must be ignored), there are some other ways to achieve what we are after. But this is for a future PR, the parsing itself should be relative untouched by this.

gerhardol:
Will investigate #9315 or a followup before continuing this PR.
(It is complete, but simplifications should maybe be done prior to this PR.)

gerhardol:
Rebased on #9317 (required some cleanup) and a few squashed other fixups (see `tmp/log-zero-0` for details).
Tests are fixed (format and LocalTime complicates unit tests).
#9317 should be merged soon.

mstv (approved):

gerhardol:
@msftbot merge in 48 hours

UnitTests/GitCommands.Tests/RevisionReaderTests.cs
```diff
@@ -78,70 +93,59 @@ public void BuildArguments_check_parameters(RefFilterOptions refFilterOptions, s
...
+        [TestCase("bad_sha", false)]
+        [TestCase("empty", false)]
+        [TestCase("illegal_timestamp", true, true)]
+        [TestCase("multi_pathfilter", true)]
+        [TestCase("no_encoding", false)]
+        [TestCase("no_subject", true)]
+        [TestCase("normal", true)]
+        [TestCase("short_sha", false)]
+        [TestCase("simple_pathfilter", true)]
+        [TestCase("subject_no_nody", true)]
```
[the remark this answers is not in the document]
gerhardol: done, 
squashed

GitCommands/RevisionReader.cs
gerhardol: ROS was implemented - one reason this PR had so many feature creeps...

PR #9246, opened 2021-05-30
Use static/cached GitVersion to resolve Git path

gerhardol:
Part of #9229 

## Proposed changes

- Reset GitVersion cache when setting Git command path
Report new GitVersion after changes

- Use static/cached GitVersion to resolve Git path
This eliminates one `git --version` command during startup. At home this is about 35ms, at work 230ms (probably serialized, so delaying the startup considerably).

## Test methodology 

Tests exist
Manual
See Git Command Log for executed command, modify gitcommand path in registry to simulate issues

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -336,6 +336,7 @@ namespace GitCommands
             }
             set
             {
+                GitVersion.ResetVersion();
                 if (IsPortable())
                 {
                     SetString("gitcommand", value);
--- a/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
@@ -177,6 +177,12 @@ namespace GitUI.CommandsDialogs.SettingsDialog
             {
                 try
                 {
+                    // Use cached version if possible
+                    if (AppSettings.GitCommand == command && !(GitVersion.Current?.IsUnknown ?? true))
+                    {
+                        return true;
+                    }
+
                     string output = new Executable(command).GetOutput(arguments: "--version");
                     if (!string.IsNullOrEmpty(output))
                     {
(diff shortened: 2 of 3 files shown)


GitCommands/Settings/AppSettings.cs
```diff
@@ -336,6 +336,7 @@ public static string GitCommandValue
             }
             set
             {
+                GitVersion.ResetVersion();
```
drewnoakes, line 339: Only if value has changed?

drewnoakes (approved):

GitCommands/Settings/AppSettings.cs
gerhardol: I considered that, but wanted to keep it simple.
It is not the end of the world to run the command again....
Can change

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
```diff
@@ -177,6 +177,12 @@ bool TestGitCommand(string command)
             {
                 try
                 {
+                    // Use cached version if possible
+                    if (AppSettings.GitCommand == command && !(GitVersion.Current?.IsUnknown ?? true))
```
mstv, line 181: Does this
```suggestion
                    if (AppSettings.GitCommand == command && GitVersion.Current?.IsUnknown is false)
```
work, too?

mstv (approved):
👍

GitUI/CommandsDialogs/SettingsDialog/CheckSettingsLogic.cs
gerhardol: works fine, updating

gerhardol:
@msftbot merge in 24 hours

gerhardol:
Missed to squash the commit, automerge should fail...

RussKie:
This issue was raised to the bot owners.

PR #9253, opened 2021-06-01
RevGrid multiline indicators 3.5

gerhardol:
Two issues from #9243 for 3.5
(unknown when the regressions occurred)

## Proposed changes

These issues relate to that the commit body is not stored for commits older than 6 months.
The subject and info if the commit has multiline info is always available.
If you select the commit in the grid, the full body is loaded (again).

- RevGrid tooltip: Only show "click to load" for multiline
This occurred if the commit had any related refs.

- Multiline indicator only shown if Body is available
Note that there is no tooltip if there are no refs or multiline. Without the multiline indicator it is hard to understand why.

## Screenshots 

before|after
--|--
image | image

## Test methodology 

manual

----


mstv (approved):
👍 Works well

RussKie (approved):

PR #9256, opened 2021-06-03
Limit superproject refs in rev grid

gerhardol:
Part of #9229

## Proposed changes

Limit the number of super project refs shown in a submodule.
The value is hardcoded to 100, could be a default in AppSettings so users could override if they want to see more even if this is not exposed in settings.
Maybe the number of references should be tweaked to an even lower value than 100.

image

image

The superproject branches can be interesting in a subrepo, but with a large number of refs (I have 1200 in my GE repo) the listing of superproject branches/remotes/tags require some time, so the update takes time. Quite some CPU load is used (in background) and the Git Command log is stalling the application if open.

This also cleans up the code, an unnecessary command to candle git-for-each-ref was removed. Note that the call was incorrect for tags, fixed in the "common" variant for 3.3 or so.

## Test methodology 

Tests are updated

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -801,26 +801,6 @@ namespace GitUI
             Translator.Translate(this, AppSettings.CurrentTranslation);
         }
 
-        internal static bool ShowRemoteRef(IGitRef r)
-        {
-            if (r.IsTag)
-            {
-                return AppSettings.ShowSuperprojectTags;
-            }
-
-            if (r.IsHead)
-            {
-                return AppSettings.ShowSuperprojectBranches;
-            }
-
-            if (r.IsRemote)
-            {
-                return AppSettings.ShowSuperprojectRemoteBranches;
-            }
-
-            return false;
-        }
-
         private void ShowLoading(bool sync = true)
         {
             _loadingControlSync.Visible = sync;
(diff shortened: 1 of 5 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -995,7 +995,7 @@ public void ForceRefreshRevisions()
                 _superprojectCurrentCheckout = null;
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    var scc = await GetSuperprojectCheckoutAsync(ShowRemoteRef, capturedModule, noLocks: true);
+                    var scc = await GetSuperprojectCheckoutAsync(capturedModule, noLocks: true);
```
mstv, line 998: Can `RevisionGridControl.ShowRemoteRef` be removed, too?

mstv (commented):
+1 LGTM, have not run yet

mstv:
> The value is hardcoded to 100, could be a default in AppSettings so users could override if they want to see more even if this is not exposed in settings.

I have no strong opinion.

> Maybe the number of references should be tweaked to an even lower value than 100.

👍 if only the RevisionGrid is affected.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Yes, the method was used in one other occasion, that filtered the refs a second time

RussKie (approved):
LGTM

RussKie (approved):
👍

PR #9274, opened 2021-06-07
Handle `laneInfo` being `null`

mstv:
## Proposed changes

- Handle `laneInfo` being `null` when determining the color of commits without neither parents nor children
- Consume exceptions from `RevisionGraphColumnProvider.OnCellPainting` since they do not bubble up to our handlers, and write them to the `Debug` output and - in debug build - also to `%TEMP%`
- Remove redundant condition in `RevisionGraphColumnProvider.OnCellPainting`

### Before

Graph rows drawn black completely due to throwing `Validates.NotNull(laneInfo)`

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build f89e7963407fb4b4cbc1229ff6db32332ed929c9
- Git 2.31.1.windows.1
- .NET 5.0.6
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Drawing.Drawing2D;
@@ -7,6 +8,7 @@ using System.Linq;
 using System.Windows.Forms;
 using GitCommands;
 using GitExtUtils.GitUI;
+using GitUI.NBugReports;
 using GitUI.UserControls.RevisionGrid.Graph;
 using GitUIPluginInterfaces;
 using Microsoft;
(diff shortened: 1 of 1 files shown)


RussKie (approved):

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
```diff
@@ -216,7 +217,15 @@ void DrawItem(Graphics g, int index)
...
-                    DrawItem();
+                    try
+                    {
+                        DrawItem();
+                    }
+                    catch (Exception ex)
+                    {
+                        // consume since it does not bubble up to our handlers
+                        Debug.Write(ex);
+                    }
```
drewnoakes, line 228: try/catch is not free. this may regress scrolling perf. i don't know often this is called. can this be done at a higher level?

drewnoakes (approved):

GitUI/UserControls/RevisionGrid/Columns/RevisionGraphColumnProvider.cs
mstv: Now, with one try/catch per graph row.

gerhardol (commented):
+1

RussKie (approved):

PR #9306, opened 2021-06-22
Fix#8742 - 'Branch search selects wrong branch when a later branch that starts with its name exists'

### LINKED ISSUE #8742: Branch search selects wrong branch when a later branch that starts with its name exists
## Current behaviour

When I create two branches where one's name is a prefix of the other's and try to select the shorter branch name the other branch gets selected.

## Expected behaviour

The correct branch gets selected.

## Steps to reproduce

1. Create a branch named "testing-with-this-quite-a-long-branch-name" from some commit.
2. Create a branch named "testing-with-this-quite-a-long-branch-name2" from a later commit.
3. In the branch search box (above the treeview) type or paste "testing-with-this-quite-a-long-branch-name".
4. Press down to select "testing-with-this-quite-a-long-branch-name" in the dropdown (not "testing-with-this-quite-a-long-branch-name2") and press Enter.
5. Observe that the "testing-with-this-quite-a-long-branch-name2" branch is selected instead.

## Screenshots

image

Don't know.

## Environment

- Git Extensions 3.4.3.9999
- Git 2.29.2.windows.2
- .NET Framework 4.7.3701.0
- DPI 120dpi (125% scaling)

AmTrJ:
Fixes #8742

## Proposed changes

- List of nodes that are filtered based on the search text, is reordered such that, the first element is the best match to search text.
- Order of Priority:
  - Full Match to search text
  - Match of text after the '/' for a branched node.
  - Partial Match

### After
image

## Test methodology 

- manual

## Test environment(s) 

- GIT 33.33.33
- Windows 10
- .net 4.8

----


gerhardol:
general recommendation: Use another branch than master in your repo, rebase before submitting a PR.
Also use a descriptive name for the PR and the commit.

This is only when pressing the search glass?
I have never used that myself, only the dropdown (that should be improved).
Before a partial match just jumped from node to node, now it cycles though all matches, which is an improvement.
A full match is now selected first I believe, but a multipath match like mp/branch may match tmp/branch before a/mp/branch.
I guess that is OK.

The dropdown when typing would benefit more from similar changes. That would also make it easier to find what is going on, I am not sure from playing around.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -474,13 +475,41 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                            }
+
                             AddTreeNodeToSearchResult(ret, n);
                         }
                     }
                     else
                     {
                         if (n.Text.IndexOf(text, StringComparison.InvariantCultureIgnoreCase) != -1)
                         {
+                            if (!isFullMatch)
```
gerhardol, line 499: This is duplication of code.
Please use a local function (or maybe a variable for branch/n.Text)

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -474,13 +475,41 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                            {
+                                if (string.Compare(text, n.Text, StringComparison.InvariantCultureIgnoreCase) == 0)
+                                {
+                                    isFullMatch = true;
+                                    matchIndex = ret.Count;
+                                }
+                                else if (matchIndex == -1 && n.Text.Contains('/') &&
+                                    string.Compare(text, n.Text.Split('/').Last(), StringComparison.InvariantCultureIgnoreCase) == 0)
+                                {
+                                    matchIndex = ret.Count;
```
gerhardol, line 509: Add a comment with something like:
// First partial match and no full match, add to first in queue
// Further partial matches will not be sorted

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,10 +520,28 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
             }
         }
 
+        private void rotateListToSetFirstAsMatchIndex(List<TreeNode> ret, int matchIndex)
+        {
+            int rotationIndex = 0;
+
+            while (rotationIndex < matchIndex)
+            {
+                TreeNode firstElement = ret.FirstOrDefault();
```
gerhardol, line 538: Why rotate?
This changes the order in the list, this should insert the matchIndex first to keep the order

AmTrJ: rotation was done to maintain the tree order, in list. On click of search button multiple times, the selected node moves down  the tree, to further matches available in the list. So if I move only the match Index to first, there will be random jumping in the tree.

AmTrJ:
Thank you for your inputs and review comments. 

This is only when pressing the search glass?- Also on pressing enter from dropdown list.

Can you provide me inputs on how I can resolve the review comments? Should I create a separate PR?

gerhardol:
> 
> 
> Thank you for your inputs and review comments.
> 
> This is only when pressing the search glass?- Also on pressing enter from dropdown list.
> 
> Can you provide me inputs on how I can resolve the review comments? Should I create a separate PR?

The review comments on this PR should be resolved with fixup! commits

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: At least add comments in the code to describe this

GitUI/BranchTreePanel/RepoObjectsTree.cs
AmTrJ: Moved it to a local function

GitUI/BranchTreePanel/RepoObjectsTree.cs
AmTrJ: Sure. I have added detailed comments.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -529,6 +504,49 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+        /// <summary>
+        /// A function to see if the node text or full path of a node (for branch node) is matching with search text
+        /// </summary>
+        /// <param name="nodeText">nodeText/full path</param>
+        /// <param name="searchText">text entered in search box</param>
+        /// <param name="isFullMatch">is full match found</param>
+        /// <param name="matchIndex">index of the closest match found</param>
+        /// <param name="ret">List of matching nodes</param>
+        /// <returns>true if match found</returns>
+        private bool searchForMatchingText(string nodeText, string searchText, ref bool isFullMatch, ref int matchIndex, List<TreeNode> ret)
```
gerhardol, line 516: Use a local method, no need to expose to other methods in this module.
Then it would not be required to use ref  - but I suggest keeping that as it improves readability (that is the bad part of local functions, they have access to all method variables). Local methods should be static if reasonable.

Similar for rotate

AmTrJ: I haven't used local methods before. It's an interesting feature.
To improve readability, I will retain all the method parameters including ref as before. 
There is no reason for both of these local methods to be static. Changes made to variables inside the function are required in the containing function also. 
Is my understanding correct?

gerhardol: If the local method is static, it means it cannot change the main method variables, so that is a good documentation.
Visual Studio will suggest this if the function has no side effects.

Local methods is something I have missed since working with Pascal a long time ago. Not much else I miss from Pascal though.

AmTrJ: Thank you for your inputs. I have moved the functions to local methods.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                    while (rotationIndex < matchIndex)
+                    {
+                        TreeNode firstElement = ret.FirstOrDefault();
+                        ret.RemoveAt(0);
+                        ret.Add(firstElement);
+                        rotationIndex++;
+                    }
+                }
+
                 return ret;
```
gerhardol, line 547: Move before the local methods.
(This convention need to be documented - a few situations where a local method fits better in the flow though.)

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                        }
+
+                        return true;
+                    }
+
+                    return false;
+                }
+
+                // Rotate the list to set the match with high priority as first item
+                void rotateListToSetFirstAsMatchIndex(List<TreeNode> ret, int matchIndex)
```
gerhardol, line 534:
```suggestion
                static void rotateListToSetFirstAsMatchIndex(List<TreeNode> ret, int matchIndex)
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                // if matches with high priority are found bring those to the first of the list by rotating the list
+                // Rotation is done to maintain the tree order, so that on click of search button multiple times,
+                // the selected node passes down the tree to other matches.
+                if (matchIndex > 0)
+                {
+                    rotateListToSetFirstAsMatchIndex(ret, matchIndex);
+                }
+
+                // A function to see if the node text or full path of a node (for branch node) is matching with search text
+                bool searchForMatchingText(string nodeText, string searchText, ref bool isFullMatch, ref int matchIndex, List<TreeNode> ret)
```
gerhardol, line 504:
```suggestion
                static bool searchForMatchingText(string nodeText, string searchText, ref bool isFullMatch, ref int matchIndex, List<TreeNode> ret)
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -465,21 +465,22 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                bool isFullMatch = false;
+                int matchIndex = -1;
                 while (queue.Count != 0)
                 {
                     var n = queue.Dequeue();
 
                     if (n.Tag is BaseBranchNode branch)
                     {
-                        if (branch.FullPath.IndexOf(text, StringComparison.InvariantCultureIgnoreCase) != -1)
+                        if (searchForMatchingText(branch.FullPath, text, ref isFullMatch, ref matchIndex, ret))
```
gerhardol, line 476: Still duplicated code
```suggestion
                    string nodeText = n.Tag is BaseBranchNode branch ? branch.FullPath : n.Text;
                    if (searchForMatchingText(nodeText, text, ref isFullMatch, ref matchIndex, ret))
                    {
                        AddTreeNodeToSearchResult(ret, n);
                    }
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                if (matchIndex > 0)
+                {
+                    rotateListToSetFirstAsMatchIndex(ret, matchIndex);
+                }
+
+                // A function to see if the node text or full path of a node (for branch node) is matching with search text
+                bool searchForMatchingText(string nodeText, string searchText, ref bool isFullMatch, ref int matchIndex, List<TreeNode> ret)
+                {
+                    if (nodeText.IndexOf(searchText, StringComparison.InvariantCultureIgnoreCase) != -1)
+                    {
```
gerhardol, line 507: Invert 
```suggestion
                    if (nodeText.IndexOf(searchText, StringComparison.InvariantCultureIgnoreCase) < 0)
                    {
                       return false;
```

gerhardol (commented):
I am OK after this I believe
I suggest you rebase and squash your changes on master and someone else should take a look too
The ordering of results in the interactive search should be fixed

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                        // If full match is already found no need to search for furthur matches as full match has highest priority
```
mstv, line 508:
```suggestion
                        // If full match is already found no need to search for further matches as full match has highest priority
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                {
+                    int rotationIndex = 0;
+
+                    while (rotationIndex < matchIndex)
+                    {
+                        TreeNode firstElement = ret.FirstOrDefault();
+                        ret.RemoveAt(0);
+                        ret.Add(firstElement);
+                        rotationIndex++;
+                    }
```
mstv, line 544: This *is* a classical `for` loop - no need for ``while`.
```suggestion
                    for(int rotationIndex = 0; rotationIndex < matchIndex; ++rotationIndex)
                    {
                        TreeNode firstElement = ret.FirstOrDefault();
                        ret.RemoveAt(0);
                        ret.Add(firstElement);
                    }
```

mstv (commented):
Two nitpicks in addition to @gerhardol's comments. Have not run it yet.

GitUI/BranchTreePanel/RepoObjectsTree.cs
AmTrJ: Yes, I had some other logic for rotation in beginning, which required while loop. Missed to change it to for loop.
Thank you.
How do I proceed with these commit suggestions? Do I commit it from here or should I make change in my local and push to fork? (This is my first PR so trying to understand the flow)

gerhardol: The normal is to make fixup commits until approved.
In this case it is OK to squash and rebase on master

AmTrJ: I had made a merge from remote/branch (5ce4cbc) after my first commit. Is it okay to squash this also?

mstv: > I had made a merge from remote/branch ([5ce4cbc](https://github.com/gitextensions/gitextensions/commit/5ce4cbc6e37a2eb755a47de79135e6bf0ccabd8f)) after my first commit. Is it okay to squash this also?

Yes, please. We use to rebase in order to avoid all merges except PR to master - (and even those if supported by GH).

mstv:
> Also use a descriptive name for the PR and the commit.

Top right, there is an `Edit` button for changing the PR title.

GitUI/BranchTreePanel/RepoObjectsTree.cs
AmTrJ: I just edited my comment above on how to do the rebase in this scenario. Can you please give your inputs on the same.

gerhardol: There are many ways to rebase. Some info in the wiki https://github.com/gitextensions/gitextensions/wiki/How-To%3A-Squash-and-Rebase-your-changes

gerhardol:
There is one problem with this change: The search button on cycles over some matches only.
The correction is probably more important than this behavior change.

AmTrJ:
> There is one problem with this change: The search button on cycles over some matches only.
- Is it skipping the base nodes like 'Remotes','Branches' etc? I observed there is something wrong in highlighting of the same (selection is not appearing). This was after merging from remote.

mstv:
> I observed there is something wrong in highlighting of the same (selection is not appearing). This was after merging from remote.

This was changed on purpose - but did not respect the search use case.
`treeMain.HideSelection` must be turned off as long as the search box is focused.

AmTrJ:
> `treeMain.HideSelection` must be turned off as long as the search box is focused.
In that case, will it work on click of search button?

mstv:
I think yes because of `_txtBranchCriterion.Focus();`.

GitUI/BranchTreePanel/RepoObjectsTree.cs
RussKie: 1. Incorrect casing, all members except for fields must be in PascalCase.
2. Methods' names that return `bool` must start with `Is`, `Has`, `Does`, etc. verb.
```suggestion
                static bool HasMatchingText(string nodeText, string searchText, ref bool isFullMatch, ref int matchIndex, List<TreeNode> ret)
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                }
```
RussKie, line 545: This creates an unnecessary churn. I don't think this is a good idea.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -491,6 +492,58 @@ List<TreeNode> SearchTree(string text, IEnumerable<TreeNode> nodes)
...
+                            {
+                                matchIndex = ret.Count;
+                            }
+                        }
+
+                        return true;
+                    }
+
+                    return false;
+                }
```
RussKie, line 531: 1. `ret` parameter is only needed to provide a return for `matchIndex = ret.Count`. This makes it superficial, and redundant. Please restructure the code to not require this parameter at all.
2. Suggest update the signature to return `(bool IsMatch, bool IsFullMatch)`

GitUI/BranchTreePanel/RepoObjectsTree.cs
AmTrJ: I thought swapping match to the first position. but this will cause jumping across tree on repeated click of search. Please provide your suggestions for the same.

PR #9308, opened 2021-06-25
Stash: Keyboard navigation

### LINKED ISSUE #8108: Current stash list UI control disallow quickily navigate between the changes of all stashes
## Current behavior

The current "Manage stashes" UI is using combo-box to select which stash to show its changes. However, after user clicked or press ENTER on a selected stash, the focus on the combo-box control (with label "Show:") is lost. This makes the user cannot use the arrow key in the keyboard to quickly switch to another stash (they must use mouse with at least two clicks and one movement, to jump to the next stash).

Consequence: This is inconvenience if I want to inspect all stashes to see if they are still useful, so as to clean-up some useless ones.

P.S. this is so far the worst UI flow, which pushes me to find other Git GUI client to help.

## Expected behavior

After the user selected a stash in the combo-box of "Show:", in the "Manage stashes" view, the focus should be kept. And allow user to use up-arrow/down-arrow to directly switch to previous/next stash to show the changes immediately, with one keystroke only.

## Steps to reproduce

Choose Main Menu > Commands > "Manage stashes" that jump to the UI view which inspects stashes and their changes. Click the combo-box labeled by "Show:", and click one item. Then, try using up-arrow or down-array key to navigate to previous/next stash --> it is impossible, must use mouse with multi-clicks and movement (inconvenience).

## Screenshots

(focus of the "Show:" combo-box is lost after selected an item by mouse-clicking or keyboard ENTER of the combo-box list)

image

No.

## Environment

- Git Extensions 3.3.1.7897
- Git 2.26.2.windows.1
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8108

## Proposed changes

- HotKeys for next/previous stash and refresh
Ctrl-P navigates to next older stash item, similar to how navigation in RevGrid.

- Mnemonics for the form. 
As `Stashes` control gets a specific hotkey, it is possible to navigate in the stashes list with Alt-H arrow /PageUp/etc too
"&Apply stash" but "S&tash all changes" as I expect Ctrl-Alt-Up to be used for stash in current worktree.

- Stash: Keep position when dropping a stash 
Simplify deleting stashes, unless deleting the first, you would have to search for the stash close to where you just deleted the stash.

- Remove button for custom message, just make it writeable for worktree.
Add label for mnemonics 

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormStash.cs
+++ b/GitUI/CommandsDialogs/FormStash.cs
@@ -183,10 +183,10 @@ namespace GitUI.CommandsDialogs
             Loading.IsAnimating = true;
             Stashes.Enabled = false;
             refreshToolStripButton.Enabled = false;
-            toolStripButton_customMessage.Enabled = false;
+            StashMessage.ReadOnly = true;
             if (gitStash == _currentWorkingDirStashItem)
             {
-                toolStripButton_customMessage.Enabled = true;
+                StashMessage.ReadOnly = false;
                 _asyncLoader.LoadAsync(() => Module.GetAllChangedFiles(), LoadGitItemStatuses);
                 Clear.Enabled = false; // disallow Drop  (of current working directory)
                 Apply.Enabled = false; // disallow Apply (of current working directory)
@@ -295,7 +295,7 @@ namespace GitUI.CommandsDialogs
 
         private void ResizeStashesWidth()
         {
-            Stashes.Size = new Size(toolStrip1.Width - 15 - refreshToolStripButton.Width - showToolStripLabel.Width - toolStripButton_customMessage.Width, Stashes.Size.Height);
+            Stashes.Size = new Size(toolStrip1.Width - 15 - refreshToolStripButton.Width - showToolStripLabel.Width, Stashes.Size.Height);
         }
 
         private void StashedSelectedIndexChanged(object sender, EventArgs e)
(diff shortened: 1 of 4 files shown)


mstv (approved):
LGTM
Hotkey(s) for `Write custom stash message` (and for `Refresh`) would be nice.

gerhardol:
> LGTM
> Hotkey(s) for `Write custom stash message` (and for `Refresh`) would be nice.

Implemented that too (also just showing refresh when it has effect, for worktree).

For the custom text mnemonics would be more natural than a hotkey, but then the logic for the custom message would need to be changed (remove button, add label and add to command if not empty) and I am not sure exactly how to handle.

mstv:
> Implemented that too (also just showing refresh when it has effect, for worktree).

👍 Thank you!

> For the custom text mnemonics would be more natural than a hotkey, but then the logic for the custom message would need to be changed (remove button, add label and add to command if not empty) and I am not sure exactly how to handle.

I agree, it is not perfect. But I think this would be too much effort for little gain.
Though the hotkey should be displayed in the tooltip of `toolStripButton_customMessage`.
(BTW: The message can be double-clicked in order to enable editing it.)

GitUI/CommandsDialogs/FormStash.cs
```diff
@@ -192,6 +211,60 @@ private void FileViewer_BottomScrollReached(object sender, EventArgs e)
...
+
+        private bool EnableCustomMessage()
+        {
+            if (Stashes.SelectedIndex != 0)
+            {
+                // Worktree not select, not relevant
+                return false;
+            }
+
+            toolStripButton_customMessage.Enabled = true;
```
mstv, line 248:
```suggestion
            if(!toolStripButton_customMessage.Enabled)
            {
                return false;
            }
```

gerhardol: changed implementation

gerhardol:
> I agree, it is not perfect. But I think this would be too much effort for little gain.

Replaced button with label, could also simplify the logic.
To some extent, also the label could be removed, the box is focused by default.
(Some struggling with WinForms layout as usual...)

RussKie (commented):
:shipit:

mstv (approved):
Works for me. (The failing test is unrelated.)

gerhardol:
rebased without further changes

gerhardol:
@msftbot merge in 24 hours

PR #9286, opened 2021-06-13
Fixup partial graph rendering

mstv:
Fixes #9108

## Proposed changes

- Fixup `RevisionGraph.GetCachedCount`: return full `cachedCount` only if loading has really finished
- Fixup condition in `DrawItem` of `RevisionGraphColumnProvider`
- Fixup theoretical race condition with `RevisionGridControl._isReadingRevisions`

## Screenshots 

see issue

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7287ca7ff5688e23ef3b0a0d242a80e744996e89
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET 5.0.6
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9309, opened 2021-06-26
Increase test robustness

mstv:
Follow-up to #9286

## Proposed changes

- remove the failing assertion (failing only on AppVeyor) because it is not really part of what the test shall verify

----


RussKie (approved):

mstv:
The test passed only once on AppVeyor.
Perhaps we should remove the failing assertion because it is not really part of what the test shall verify.

RussKie (approved):

PR #9311, opened 2021-06-28
Reduce window flicker and loss of focus

RussKie:
## Proposed changes

Reduce window flicker and loss of focus

1. Force child window ownership to FormBrowse

2. Show the rebase dialog *before* starting any operations. This is what appears to be the main cause of the loss of focus. The dialog isn't yet shown when we invoke the progress dialog, which is then gets closed but the window that spawned it is not visible. MS Windows gets confused and pushes another window to the foreground, thus causing the app to lose focus.

### Before

rebase-before

### After

rebase-after

----

### DIFF
--- a/GitUI/CommandsDialogs/FormRebase.Designer.cs
+++ b/GitUI/CommandsDialogs/FormRebase.Designer.cs
@@ -601,7 +601,6 @@ namespace GitUI.CommandsDialogs
             this.Name = "FormRebase";
             this.StartPosition = System.Windows.Forms.FormStartPosition.CenterParent;
             this.Text = "Rebase";
-            this.Load += new System.EventHandler(this.FormRebaseLoad);
             this.flowLayoutPanel2.ResumeLayout(false);
             this.flowLayoutPanel2.PerformLayout();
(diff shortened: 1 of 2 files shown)


gerhardol:
If using the quick commands, I do not expect the dialog to be opened unless there are conflicts.

RussKie:
This is what causes the flicker and the loss of focus.

gerhardol:
> This is what causes the flicker and the loss of focus.

But I find the opening of the window more annoying...

RussKie:
Loss of focus is an accessibility issue. We are showing the progress dialog
anyway, so can't see your argument here.

gerhardol (approved):
I can get used to the extra popup, then the fixed loss of focus for rebase is an improvement

mstv (approved):
🎉 Rebasing without `Alt+Tab`. Thank you!

RussKie:
I rolled all "Owner" related changes as in some scenarios those could lead to the active top window could get under another window with the same parent. Since now the rebase dialog is visible and lives longer any of its children, the existing ownership model is correct.

PR #9314, opened 2021-06-28
Exclude vimdiff1 from custom difftools

gerhardol:
## Proposed changes

vimdiff1 is a console only application included since Git 2.31
It should be excluded from the default difftools

This is a little annoying if you only have one difftool

This could be a hidden setting too, to allow the user to override the GE defaults.
For now this is good enough, new tools are not added that often.

To be cherry-picked for 3.5.1

### Before

image

### After

With one tool only, the rightmost submenu is not added

## Test methodology 

Manual

----


mstv (approved):

gerhardol:
Also pushed to release/3.5

PR #9315, opened 2021-06-29
StreamExtensions: Read double null terminated records

gerhardol:
Related to #9243 

I actually want this to be merged before #9243, it eliminates a parse error for git-log.
There is a merge conflict for these PRs.

## Proposed changes

FormFileHistory adds "--name-info" for "git-log -z". This adds all file names matched in the commit, all separated by additional null characters.
A record with --name-status is therefore terminated with double null.
ReadNullTerminatedChunks() tried to deliver all null terminated records as results. If only one filename matched, this was mostly OK
as a zero size buffer was parsed. With more than one filename matched, the parsing would silently fail.

With "--name-status" a double null end the record, but the 'null' terminating each filename is still included.
Only the first filename is used (as before), others are ignored.

Note: EndOfNull handling can be simplified, the special marker can be removed. That will be done in #9243 or a follow up.
This is actually the primary reason for the PR.

git-log also offers "--log-size" that will include a line “log size <number>” to inform of the size of the header, without files.
This will still not help to find the end of the filenames though, not included in the size.
So most of the complexity of the parsing, including the extra logic added here will have to remain.

Also, there must be a pathfilter to be able to parse the output. For instance stashes may not have any files at all, then the record will be ended by a single null.
That could maybe be fixed with a double null StartofRecord (that will appear before "log size" if that is implemented), that will be quite complex for something without a usecase.

## Test methodology 

Added tests for ReadNullTerminatedChunks() and debug printouts when parsing fails. (GE and linux repos parse without failure.)
The testcases for RevisionReader is not updated, they are slightly broken currently in master anyway, #9243 adds tests for TryParseRevision().

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -65,10 +65,11 @@ namespace GitCommands
             string branchFilter,
             string revisionFilter,
             string pathFilter,
+            bool hasNameInfo,
             Func<GitRevision, bool>? revisionPredicate)
         {
             ThreadHelper.JoinableTaskFactory
-                .RunAsync(() => ExecuteAsync(module, refs, subject, refFilterOptions, branchFilter, revisionFilter, pathFilter, revisionPredicate))
+                .RunAsync(() => ExecuteAsync(module, refs, subject, refFilterOptions, branchFilter, revisionFilter, pathFilter, hasNameInfo, revisionPredicate))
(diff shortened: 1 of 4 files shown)


gerhardol:
There are no files for some commits like stash. This change still works as the --name-only is only used with filters, but it is not general enough. So --log-size should be used.

PR #9317, opened 2021-06-30
GitRevision: Remove Name property

gerhardol:
Replaces #9315 
Related to #9243 (will be simplified by this PR so this should be merged first).

## Proposed changes

Name was used in FormFileHistory for a specific file only,
not fitting the general use in RevisionReader.
Instead just parse this data in FormFileHistory where it is used.

* Displaying a file.
  - Set title (if it differs).
  - Check if the file exists on the filesystem (affects operations on submodules and line patching).
* Diff a file, select the old filename in an older revision.
* Use old filename in SaveAs dialog.

The parsing in StreamExtensions was not reliable for git-log -z --name-only,
requiring searches for  double null terminated records
as well as some guessing for commits without filnames (like stash).
Parse errors had to be silently ignored.

When skipping the generic case (and variable length commit message),
this is much simpler.

This allowed for removing EndOfBody handling in RevisionReader.

## Test methodology 

As functionality in RevisionReader is simplified and code were removed, some testcases were removed.
#9243 adds testcases for the complete TryParseRevision() that are more relevant, the testcases fails on AppVeyor though.

Changes in FormFileHistory.cs are tested manually (affects UI primarily).
One example for a renamed file is `GitUI\CommandsDialogs\RevisionDiffControl.cs`

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9320, opened 2021-07-01
Quick search in whole commit message if loaded

mstv:
## Proposed changes

- Quick search `Body` - instead of just `Subject` - if available

## Test methodology 

- manual:
  - focus RevisionGrid
  - type "cherry"
  - hit <kbd>Alt</kbd>+<kbd>Down</kbd>
    should find cherry-picks

## Test environment(s) 

- Git Extensions 33.33.33
- Build 2e3fce40b207c8ff4c3ee5c4b442e58a3f812b7b
- Git 2.31.1.windows.1
- .NET 5.0.7
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/GitRevisionTester.cs
+++ b/GitCommands/Git/GitRevisionTester.cs
@@ -108,7 +108,7 @@ namespace GitCommands.Git
             }
 
             return revision.Author?.IndexOf(criteria, StringComparison.CurrentCultureIgnoreCase) >= 0 ||
-                   revision.Subject?.IndexOf(criteria, StringComparison.OrdinalIgnoreCase) >= 0;
+                   (revision.Body ?? revision.Subject)?.IndexOf(criteria, StringComparison.OrdinalIgnoreCase) >= 0;
         }
     }
 }


gerhardol:
Is this the second part of https://github.com/gitextensions/gitextensions/pull/9317#issuecomment-871772235?

mstv:
Yes, it is.

gerhardol (commented):
I expect that this quick search applies to what you see in the grid. This should maybe be related to the new setting to see the body in the grid.

RussKie (approved):

mstv:
> I expect that this quick search applies to what you see in the grid.

That would be a clear definition. But the result would be counterintuitive to me since the whole message of the selected commit is visible always - located right of the RevisionGrid, (where it should be displayed by default - perhaps left for RTL languages).

It should behave similar as
grafik

gerhardol:
I meant if this is set:
image

Like searching in a document

mstv:
> I expect that this quick search applies to what you see in the grid.
> Like searching in a document

I think I had understood what you mean.
What about "quick search applies to what you _can_ see in the grid - including the tooltip"?

gerhardol:
> > I expect that this quick search applies to what you see in the grid.
> > Like searching in a document
> 
> I think I had understood what you mean.
> What about "quick search applies to what you _can_ see in the grid - including the tooltip"?

Ok

Issue #9273, opened 2021-06-07
Repair Translation Check


mstv:
The translation check has regressed again: fails, if the master has newer translations.

STR:

- PR with translation change
- concurrent merge with translation change to `master`
- build PR

RussKie:
I'll have a look

RussKie:
I need a repro, if you see one please let me know.

RussKie:
Got a repro https://github.com/gitextensions/gitextensions/commit/f5c663c92357a1673357d7a90d5a32c1546ab6ea

Here's what's happening:
1. The master has a changed translation, when we perform a soft reset the index it contains xlf files:
![image](https://user-images.githubusercontent.com/4403806/125015340-ac08f200-e0b2-11eb-918f-97228f074f84.png)
2. Then we build and run the loc check, which detects the changed xlf

In 3.5 branch we would
1. Reset
2. Build
3. Undo reset
4. Run loc check

mstv:
I think you should have a look at the code removed (again) from `appveyor.yml` in 3eea46688b0398557f0a8423f48c4fb4d95c9ad2.
It did/does only work this way.

RussKie:
Tested in https://ci.appveyor.com/project/gitextensions/gitextensions/builds/39945061

PR #9332, opened 2021-07-04
Add support for setting console emulator font name

dasiths:
# Add Support For Setting Console Emulator Font Name

This is the 2nd part supplementing the PR to the ConEmu fork https://github.com/gitextensions/conemu-inside/pull/28. The idea  is to allow powerline supported fonts.

As shown on the screenshot below, setting the font name to a powerline supported font resulted in this nice rendering.
image

image

I could not get the design view for `ConsoleStyleSettingsPage` to load in Visual Studio 2019 but I hand edited the `ConsoleStyleSettingsPage.designer.cs` file directly. It build, runs and renders fine but as an improvement I would like to use the standard button based font selection and combine font name + size. I'll do it as a separate PR next.

### DIFF
--- a/GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
+++ b/GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
@@ -29,10 +29,11 @@
         private void InitializeComponent()
         {
             this.groupBoxConsoleSettings = new System.Windows.Forms.GroupBox();
-            this.label3 = new System.Windows.Forms.Label();
-            this.cboFontSize = new System.Windows.Forms.ComboBox();
+            this.consoleFontChangeButton = new System.Windows.Forms.Button();
+            this.lblFontName = new System.Windows.Forms.Label();
             this.label1 = new System.Windows.Forms.Label();
             this._NO_TRANSLATE_cboStyle = new System.Windows.Forms.ComboBox();
+            this.consoleFontDialog = new System.Windows.Forms.FontDialog();
             this.groupBoxConsoleSettings.SuspendLayout();
             this.SuspendLayout();
             // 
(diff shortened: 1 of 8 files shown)


RussKie:
Please make sure the solution builds.
```
C:\Program Files\dotnet\sdk\5.0.301\Microsoft.Common.CurrentVersion.targets(1717,5): warning NU1702: ProjectReference 'C:\projects\gitextensions\Externals\ICSharpCode.TextEditor\Project\ICSharpCode.TextEditor.csproj' was resolved using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v5.0'. This project may not be fully compatible with your project. 
```

mstv:
> Please make sure the solution builds.

It might be caused by the changed other submodules. Perhaps the merge went wrong.

Basically, please do not merge anything into feature branches - rebase instead.

dasiths:
@RussKie @mstv I've reverted the changes for other submodules. It builds locally for both previously and also now.

GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
```diff
@@ -49,17 +53,39 @@ private void InitializeComponent()
...
+            this.label2.TabIndex = 4;
+            this.label2.Text = "Font name";
+            // 
+            // cboFontName
+            // 
+            this.cboFontName.FormattingEnabled = true;
+            this.cboFontName.Items.AddRange(new object[] {
+            "Consolas",
+            "Firacode",
+            "Cascadia Code PL"});
```
RussKie, line 76: These fonts are non standard, and must be installed by a user prior to their use. Hence we can't include them like this.
In general this isn't a good way to facilitate user settings because some users may want to user another arbitrary font. It should be done similar to how we configure other fonts, refer to the Appearance>Fonts tab.

GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
```diff
@@ -152,6 +178,8 @@ private void InitializeComponent()
         private System.Windows.Forms.GroupBox groupBoxConsoleSettings;
         private System.Windows.Forms.Label label3;
         private System.Windows.Forms.ComboBox cboFontSize;
+        private System.Windows.Forms.Label label2;
```
RussKie, line 181: Please give it a good name

RussKie (changes requested):
```
Build FAILED.
C:\projects\gitextensions\GitExtensions\Project.Loc.targets(89,5): error : Please update English translations and re-submit the pull-request. Refer to https://github.com/gitextensions/gitextensions/wiki/Translations [C:\projects\gitextensions\GitExtensions\GitExtensions.csproj]
    0 Warning(s)
    1 Error(s)
```

GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
dasiths: Thanks for the review. The combo box is currently editable so the combo box isn't only limited to items in the list. Having said that I wanted to use the standard "button based approach" for font selection but ran into a problem with the designer.

image

image

This means I have a really hard time adding a change like this. I've been hand editing the designer.cs thus far. I was hoping to land this PR and do it as another PR once I figured out the issues with the designer views.

dasiths: @RussKie I implemented the change using the font selection dialog.

dasiths:
Not entirely sure what I need to do to get this resolved?
```
C:\projects\gitextensions\GitExtensions\Project.Loc.targets(89,5): error : Please update English translations and re-submit the pull-request. Refer to https://github.com/gitextensions/gitextensions/wiki/Translations
```

GitUI/CommandsDialogs/SettingsDialog/Pages/ConsoleStyleSettingsPage.Designer.cs
dasiths: @RussKie I've made the required changes now and used the font dialog as requested.

GitUI/Translation/English.xlf
```diff
@@ -8701,6 +8705,14 @@ This will just checkout on the submodule the commit determined by the superproje
...
     </body>
   </file>
+  <file datatype="plaintext" original="Plugin" source-language="en">
+    <body>
+      <trans-unit id="Description.Text">
+        <source>Plugin Manager</source>
+        <target />
+      </trans-unit>
+    </body>
+  </file>
```
RussKie, line 8715: Please revert this piece

RussKie (commented):
This is radical, but I think this is a good direction. However now the "Console style" page is pretty empty, and perhaps we could consider moving the console font setting to the "Fonts" page, and the console style dropdown to the "Colors" page (under its own group, like "Console theme"?).
What do you think?

GitUI/Translation/English.xlf
dasiths: > This is radical, but I think this is a good direction. However now the "Console style" page is pretty empty, and perhaps we could consider moving the console font setting to the "Fonts" page, and the console style dropdown to the "Colors" page (under its own group, like "Console theme"?).
> What do you think?

I could be awayed either way. I like that the console related stuff is in one sub page but I get your thinking as well. 

I was contemplating the fact that the entire ConEmu settings xml can be loaded externally as well and that maybe another feature I could work on next. If we want to try that, then we would need the console sub page to place the xml file path picker for example.

i.e. These are all the settings embedded in that xml file
image

This isn't a reason to keep the console sub page though. Just something that came to my mind. Let me know how you want to proceed. I can go with you suggested approach if you want.

dasiths:
> ```
> Build FAILED.
> C:\projects\gitextensions\GitExtensions\Project.Loc.targets(89,5): error : Please update English translations and re-submit the pull-request. Refer to https://github.com/gitextensions/gitextensions/wiki/Translations [C:\projects\gitextensions\GitExtensions\GitExtensions.csproj]
>     0 Warning(s)
>     1 Error(s)
> ```

I've added the translations but the build still keep failing :(

mstv:
> I've added the translations but the build still keep failing :(

This is an open issue if the master has translation changes.
So it is a perfect time to rebase on master and squash.

RussKie:
> This is an open issue if the master has translation changes.

I don't believe this has anything to do with #9273

This is what's causing the issue:
image
image

dasiths:
> > This is an open issue if the master has translation changes.
> 
> I don't believe this has anything to do with #9273
> 
> This is what's causing the issue:
> image
> image

But that's the change you wanted me to revert right? It fails translation before and after reverting that change.

GitUI/Translation/English.xlf
RussKie: > I was contemplating the fact that the entire ConEmu settings xml can be loaded externally as well and that maybe another feature I could work on next. If we want to try that, then we would need the console sub page to place the xml file path picker for example.
> 
> i.e. These are all the settings embedded in that xml file
> image

Instead of duplicating the ConEmu's settings page, I'd rather see us providing a mechanism to launch the ConEmu configurator. Though personally I don't think it adds any value, you're one of a handful  contributors who expressed the desire to tweak it.

RussKie:
Yes. You're introduced trailing spaces, which must be undone for the translation check to pass.

GitUI/Translation/English.xlf
dasiths: Yeah I wouldn't wanna duplicate it. Rather I was thinking let us point to a xml file from our settings. But like you said, it might be overkill.

So provided we want to continue with the reorg or pages, should I tackle the sub page move as a separate PR?

dasiths:
> Yes. You're introduced trailing spaces, which must be undone for the translation check to pass.

Fixed.

RussKie:
Please provide an updated screenshot

RussKie (approved):
👍 
Please squash

dasiths:
> 👍
> Please squash

@RussKie done and pushed.

RussKie:
Thank you

PR #9337, opened 2021-07-04
Use CsWin32 source generator for interop in GitUI

sharwell:


RussKie (commented):
Overall looks straight forward.

Issue #9354, opened 2021-07-09
GE fails on a commit with empty subject (or doesn't display it at all)


crypto-rsa:
## Current behaviour

When there is a commit with an empty message in the repository, GE shows an error message and hangs/crashes (in Debug build). In a Release build, such commits are not displayed at all.

## Expected behaviour

Commits with an empty message should be displayed normally. They are legitimate - one can be created by `git commit -a --allow-empty-message -m ''`. In my repository, there is a number of old commits (from pre-`git` version control system) which were imported automatically and which have no subject.

## Steps to reproduce

1. Create a repository
2. Make a change
3. Commit using `git commit -a --allow-empty-message -m ''`
4. Open the repository in GE

## Did this work in previous version of GitExtensions

Yes, this behaviour has been introduced in c8396d8d94e06c8c24d23bad33554c92a2754fad.

## Environment

- Git Extensions 33.33.33
- Build c8396d8d94e06c8c24d23bad33554c92a2754fad
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19043.0
- .NET 5.0.7
- DPI 96dpi (no scaling)

gerhardol:
Created a PR for this case

The parsing ignores all parse errors, this Debug.Fail is an exception. It was not possible to call Debug.Fail on the other parse failures as FileHistory use case would feed incorrect data. 
Parse failures will likely come in multiples so the assert should probably turn off after the first.

To display parse errors in Release builds is more tricky, it is likely less disruptive to just continue than to abort here.

gerhardol:
@crypto-rsa Please review the pull requests

PR #9359, opened 2021-07-10
Log RevisionReader parse errors

gerhardol:
Followup to #9354 #9243

## Proposed changes

* A few minor optimizations, mostly using `in` parameters (likely not noticeable)
The log parsing is probably the most important code to optimize for opening a repo.

* Log parse errors (with limitations)

Before #9243 parse errors had to be ignored as the input data was incorrectly split for FileHistory.
In #9354 the parsing failed (introduced in #9243), only seen in DEBUG though.

If there are other parse errors (likely existing from before, not much of the actual parsing logging has been changed), the visibility is improved by Debug.Assert the first occurrence and Trace.Assert the first 10 errors. Printouts include the commit sha if available.
This will notify DEBUG users and give a way for production users to provide information, but avoid that the log or popups overwhelming the user.

Also removed the trace printout about time to parse the log information.

Note: TRACE is set by default in GE.

Note 2: Debug.Assert seem to abort execution in master, not giving the popup to allow continuing as before (.NET5 or BugReporter?)

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -54,6 +54,8 @@ namespace GitCommands
             /* Committer email */ "%cE%n" +
             /* Commit raw body */ "%B";
 
+        // Trace info for parse errors
+        private static int _noOfParseError = 0;
         private readonly CancellationTokenSequence _cancellationTokenSequence = new();
 
         public void Execute(
@@ -111,9 +113,8 @@ namespace GitCommands
 
             var arguments = BuildArguments(maxCount, refFilterOptions, branchFilter, revisionFilter, pathFilter);
 
-#if TRACE
+#if DEBUG
             var sw = Stopwatch.StartNew();
-            int parseErrors = 0;
 #endif
 
             var logOutputEncoding = module.LogOutputEncoding;
@@ -140,16 +141,11 @@ namespace GitCommands
 
                         subject.OnNext(revision);
                     }
-#if TRACE
-                    else
-                    {
-                        parseErrors++;
-                    }
-#endif
                 }
 
-#if TRACE
-                Trace.WriteLine($"**** [{nameof(RevisionReader)}] Emitted {revisionCount} revisions in {sw.Elapsed.TotalMilliseconds:#,##0.#} ms. bufferSize={buffer.Length} parseErrors={parseErrors}");
+#if DEBUG
+                // TODO Make it possible to explicitly activate Trace printouts like this
+                Debug.WriteLine($"**** [{nameof(RevisionReader)}] Emitted {revisionCount} revisions in {sw.Elapsed.TotalMilliseconds:#,##0.#} ms. bufferSize={buffer.Length} parseErrors={_noOfParseError}");
 #endif
             }
 
@@ -254,6 +250,7 @@ namespace GitCommands
 
             if (chunk.Count < ObjectId.Sha1CharCount * 2)
             {
+                ParseAssert($"Log parse error, not enough data: {chunk.Count}");
                 revision = default;
                 return false;
             }
@@ -266,6 +263,7 @@ namespace GitCommands
             if (!ObjectId.TryParseAsciiHexReadOnlySpan(array.Slice(0, ObjectId.Sha1CharCount), out var objectId) ||
                 !ObjectId.TryParseAsciiHexReadOnlySpan(array.Slice(ObjectId.Sha1CharCount, ObjectId.Sha1CharCount), out var treeId))
             {
+                ParseAssert($"Log parse error, object id: {chunk.Count}({array.Slice(0, ObjectId.Sha1CharCount).ToString()}");
                 revision = default;
                 return false;
             }
@@ -276,7 +274,7 @@ namespace GitCommands
             int noParents = CountParents(in array, offset);
             if (noParents < 0)
             {
-                // Parse issue
+                ParseAssert($"Log parse error, {noParents} no of parents for {objectId}");
                 revision = default;
                 return false;
             }
@@ -288,7 +286,8 @@ namespace GitCommands
 
                 while (baseOffset < array.Length && array[baseOffset] != '\n')
                 {
-                    Debug.Assert(count == 0 || array[baseOffset] == ' ', $"Unexpected contents in the parent array: {array[baseOffset]}/{count}");
+                    // Parse error, not using ParseAssert (or increasing _noOfParseError)
+                    Debug.Assert(count == 0 || array[baseOffset] == ' ', $"Log parse error, unexpected contents in the parent array: {array[baseOffset]}/{count} for {objectId}");
                     baseOffset += ObjectId.Sha1CharCount;
                     if (count > 0)
                     {
@@ -313,7 +312,7 @@ namespace GitCommands
             {
                 if (!ObjectId.TryParseAsciiHexReadOnlySpan(array.Slice(offset, ObjectId.Sha1CharCount), out ObjectId parentId))
                 {
-                    // TODO log this parse problem
+                    ParseAssert($"Log parse error, parent {parentIndex} for {objectId}");
                     revision = default;
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


GitCommands/RevisionReader.cs
```diff
@@ -428,6 +423,13 @@ long ParseUnixDateTime(ref ReadOnlySpan<byte> array)
             };
 
             return true;
+
+            static void ParseAssert(string? message)
+            {
+                _noOfParseError++;
+                Debug.Assert(_noOfParseError > 1, message);
```
mstv, line 430: Shall the first error be ignored? Or is the identifier misleading as I guessed above?

GitCommands/RevisionReader.cs
```diff
@@ -266,29 +262,30 @@ private static bool TryParseRevision(ArraySegment<byte> chunk, Func<string?, Enc
...
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            int CountParents(ref ReadOnlySpan<byte> array, int baseOffset)
+            int CountParents(in ReadOnlySpan<byte> array, int baseOffset)
             {
                 int count = 0;
 
                 while (baseOffset < array.Length && array[baseOffset] != '\n')
                 {
-                    Debug.Assert(count == 0 || array[baseOffset] == ' ', $"Unexpected contents in the parent array: {array[baseOffset]}/{count}");
+                    Debug.Assert(count == 0 || array[baseOffset] == ' ' || _noOfParseError++ > 0, $"Unexpected contents in the parent array: {array[baseOffset]}/{count} for {objectId}");
```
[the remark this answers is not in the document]
gerhardol: We can remove increasing the counter here too, this has not raised any issues what I know.
Alternative wrap this in #if DEBUG, use ParseAssert() (then it is same as currently in the PR)

GitCommands/RevisionReader.cs
gerhardol: The first error fails to be asserted, so it is displayed. The other are ignored.

mstv: Thank you. You are right.

GitCommands/RevisionReader.cs
```diff
@@ -111,9 +113,8 @@ public sealed class RevisionReader : IDisposable
 
             var arguments = BuildArguments(maxCount, refFilterOptions, branchFilter, revisionFilter, pathFilter);
 
-#if TRACE
+#if DEBUG
```
RussKie, line 116: `TRACE` is available in release builds, and can be viewed in [DebugView](https://docs.microsoft.com/sysinternals/downloads/debugview)

GitCommands/RevisionReader.cs
```diff
@@ -327,11 +325,11 @@ int CountParents(ref ReadOnlySpan<byte> array, int baseOffset)
...
 
             // Lines 2 and 3 are timestamps, as decimal ASCII seconds since the unix epoch, each terminated by `\n`
-            var authorUnixTime = ParseUnixDateTime(ref array);
-            var commitUnixTime = ParseUnixDateTime(ref array);
+            var authorUnixTime = ParseUnixDateTime(array);
+            var commitUnixTime = ParseUnixDateTime(array);
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            long ParseUnixDateTime(ref ReadOnlySpan<byte> array)
+            long ParseUnixDateTime(in ReadOnlySpan<byte> array)
```
RussKie, line 332: ROS is a struct, so it has to be be `ref`, otherwise it will be copied

gerhardol: ROS is a `readonly ref struct` so I interpret it as in will use a reference.
https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#avoid-defensive-copies
`in` should be specified at the call sites to specify the intent. 

So the ObjectId.cs changes does not have any effect.

RussKie: `ref structs` aren't copied, they are passed by reference

RussKie: https://docs.microsoft.com/dotnet/csharp/write-safe-efficient-code#use-ref-struct-types

GitCommands/RevisionReader.cs
gerhardol: Yes, this was intentional. I do not see a need for this in released builds
Of course, performance tests need to manually enable this.
While this is likely the most important piece of code to optimize opening of repos, it is not the only part that need tweaking.
This could be controlled with a app setting too (probably just in the settings file, not in the settings form).

GitCommands/RevisionReader.cs
gerhardol: I interpret the section above that described defensive copies to that `in` should do the same as `ref` here.
If you disagree, I can change, ref will work (but it is not the intent).
If that is changed, I would like to change StringLineReader internal class to local methods.

drewnoakes: > `ref structs` aren't copied, they are passed by reference

Ref structs can be copied on the stack. They just cannot escape the stack, which gives them guarantees around their lifetimes.

Both ref and in can avoid the copy for a readonly struct (ref or not). I suspect in to be the better choice here.

GitCommands/RevisionReader.cs
RussKie: Then we should replace just `Trace` calls with `Debub` calls :)

GitCommands/RevisionReader.cs
RussKie: I stand corrected :)

GitCommands/RevisionReader.cs
```diff
@@ -473,7 +472,7 @@ public StringLineReader(ReadOnlySpan<char> s)
                 return StringPool.Shared.GetOrAdd(_s.Slice(startIndex, lineLength));
             }
 
-            public (string? subject, string? body, bool hasMultiLineMessage) PeakToEnd(bool skipBody)
+            public (string? subject, string? body, bool hasMultiLineMessage) PeakSubjectBody(bool skipBody)
```
crypto-rsa, line 476: I am not quite familiar with this code but I think the first word in the method name should be _Pe**e**k_, not _Pe**a**k_. It also occurs in some comments so ideally these should be updated, too.

GitCommands/RevisionReader.cs
```diff
@@ -285,7 +285,8 @@ int CountParents(in ReadOnlySpan<byte> array, int baseOffset)
 
                 while (baseOffset < array.Length && array[baseOffset] != '\n')
                 {
-                    Debug.Assert(count == 0 || array[baseOffset] == ' ' || _noOfParseError++ > 0, $"Unexpected contents in the parent array: {array[baseOffset]}/{count} for {objectId}");
+                    // Parse error, not increasing  using ParseAssert (or increasing _noOfParseError)
```
crypto-rsa, line 288: Nitpick: double space before _using_.

gerhardol: will change

GitCommands/RevisionReader.cs
gerhardol: will change

GitCommands/RevisionReader.cs
gerhardol: will change this printout
(printouts for ParseAssert will remain)

crypto-rsa:
Hmm, I just noticed an issue which is probably related to this. In the history form for any file, the very first commit (where the file was added) is displayed incorrectly - the subject is replaced by the author e-mail. The commit info box is correct, though.

obrazek

crypto-rsa:
The history issue is indeed fixed by the latest fixup.

gerhardol:
> The history issue is indeed fixed by the latest fixup.

That was a regression in master, since gitext://gotocommit/c8396d8d94e06c8c24d23bad33554c92a2754fad

GitCommands/RevisionReader.cs
gerhardol: Added a comment that this should be possible to activate
#if DEBUG must remain

gerhardol:
@msftbot merge in 3 days

PR #9361, opened 2021-07-11
Fixup localization verification

### LINKED ISSUE #9273: Repair Translation Check

mstv:
Fixes #9273

## Proposed changes

- Restore the working loc verification from 8da72f42cf02b377fbdb9ccccf1747f1b3f53e46
- Stage all files which have been modified by the build process, a.o. by `set_version_to.py`

## Test methodology 

- CI build

## Test environment(s) 

- AppVeyor

----

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -54,10 +54,15 @@ build_script:
     dotnet build -c Release --verbosity q --nologo /bl:.\artifacts\log\build.binlog /p:ContinuousIntegrationBuild=true
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
 - ps: |
-    # if we have reset above we need to reset English xlfs, otherwise the loc verification step will fail
+    # if we have reset above, we need to restore the merge commit, otherwise the loc verification step will fail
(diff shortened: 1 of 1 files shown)


RussKie:
I've been thinking overnight - we want to build with the has of the original PR. So perhaps we want to run the loc check first, and only if it passes perform soft reset and build. This way we don't need to do gymnastics of resetting back, etc.
What do you think?

mstv:
It is not about resetting but about having committed the `.xlf` files, which have been merged with the master.

Building twice is also gymnastics and will take too much time IMO.

So rather make the script which injects the git info, aware of the CI build and override the hash with the `APPVEYOR_PULL_REQUEST_HEAD_COMMIT`.

RussKie (approved):

RussKie:
It is working, though is giving strange output
image

mstv:
> It is working, though is giving strange output

That's why I must reset `LASTEXITCODE`.

Issue #9360, opened 2021-07-11
After update to 3.5.1 PuTTY key based authentication broken


andrew-adamson:
## Current behaviour

I had working ssh key auth in version 3.5.0.11713 (3.5.0 release) but then it broke upon upgrading to 3.5.1.12196 (3.5.1 release). Instead of authenticating to the remote repo with the key it would ask me for a password, which I do not have.

Rolling back to 3.5.0 corrected the issue. 

## Expected behaviour
Working key based auth to a remote repository.

## Steps to reproduce

I have putty+pageant fully installed separately to Git Extensions, so have the path to pageant set as C:\Program Files\PuTTY\pageant.exe (and the other binaries also set to the same directory). "Automatically start authentication client when a private key is configured for a remote" is also ticked. On the remote repository settings I have the same private key file path set as the key loaded in pageant, and the "test connection" function passes the test in 3.5.1.

To reproduce, simply try and push changes to a remote repo, where it will prompt for a password instead of authenticating with the key.

## Did this work in previous version of GitExtensions

Yes, was working in the 3.5.0 release.

## Environment

- Git Extensions 3.5.0.11713 (was 3.5.1.12196 before rollback)
- Build 0a5ef9ca6ce87754a510aa02c7a96d8b915714ac (Dirty)
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19043.0
- .NET Framework 4.8.4390.0
- DPI 96dpi (no scaling)

gerhardol:
For explanations to changes in SSH in 3.5.1 see https://github.com/gitextensions/gitextensions/issues/9355#issuecomment-877706848
Please check the contents in the registry
Also check if any environment variables like GITEXT_PLINK are set, that overrides the ssh setting.
It may also be similar to the other reports that GE cannot access the plink path, then the fallback is to use an empty GIT_SSH setting, so Git tries to use OpenSSH.

This works for me.
(If pageant path is incorrect, I get a popup that pageant cannot be started, but some compatibility code tries to find pageant so it is not necessary normally.)

gerhardol:
Test build available, see https://github.com/gitextensions/gitextensions/issues/9355#issuecomment-878650413

andrew-adamson:
[...]

Issue #9355, opened 2021-07-09
After update to 3.5.1 OpenSSH keybased authentication broken


dopry:
## Current behaviour
I'm using the OpenSSH newly included in windows with KeepassXC as an SSH Agent. I can use git as expected from the command line, but git extensions fails to authenticate.  

I cannot set the SSH client to c:\Windows\System32\OpenSSH\ssh.exe because the open file dialog does not see the OpenSSH folder.  I've also set GIT_SSH to c:\Windows\System32\OpenSSH\ssh.exe, but git extensions doesn't seem to respect the environment variable. 

## Expected behaviour
1. Git Extensions can use c:\Windows\System32\OpenSSH\ssh.exe
2. Git Extensions will respect GIT_SSH

## Did this work in previous version of GitExtensions
It worked in 3.5.0

## Environment
Sorry I reverted since I need to work, but it was GE 3.5.1.12???

mstv:
> I cannot set the SSH client to c:\Windows\System32\OpenSSH\ssh.exe because the open file dialog does not see the OpenSSH folder.

I can confirm that the FileOpenDialog does not see `c:\Windows\System32\OpenSSH\` - only for 3.5.1, works for .NET 5. Very weird!

Have you tried entering `c:\Windows\System32\OpenSSH\ssh.exe` directly without the `Browse` button?

gerhardol:
The GE SSH settings is stored in the registry, to work with the installer and so the setting can be shared with multiple installations. (That should probably be changed, just using the GE settings but the installer is tricky.)
See the key HKEY_USERS\SOFTWARE\GitExtensions\gitssh

If the path ends with"plink.exe", PuTTY is assumed, and some special handling is done (starting PageAnt etc).
Otherwise there is no special handling. If the key is empty or the file does not exist, the setting in GE is presented as OpenSSH, otherwise Other. This is just a presentation, no different handling.
The GIT_SSH variable is always set also if the reg key empty. Git uses ssh.exe in the installation if the variable is not set. 
GE does not use GIT_SSH environment variable itself (only for Git), but sets the value to follow the settings (there were issues opened when GIT_SSH was not set). So part 2. should not be changed directly (see below first).
It should be possible to set GIT_SSH_COMMAND though.
https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHCOMMANDcode

Before 3.5.1, the SSH settings were lost in the upgrade. 
GIT_SSH was set regardless if the path existed or not.
[...]

Issue #9348, opened 2021-07-08
After updating to 3.5.1, I always get asked the SSH passphrase every time a pull is done


vittoso:
## Current behaviour

After updating to 3.5.1 from 3.4.3.9999, everytime I click the pull button in GE git asks me for the OpenSSH passphrase. 
With 3.4.3.9999 I had to retype the passphrase rarely as it was remembered, I believe this is the correct behaviour.
I tried both "OpenSSH" and "Other ssh client" (pointing to the ssh.exe from the windows 10 open ssh-agent) in GE settings, but the behaviour is the same

Using command-line git from powershell correctly remembers the openssh passphrase.

## Expected behaviour
Users does not have to retype the passphrase everytime a pull is done

## Steps to reproduce

1. Update from 3.4.3.9999 to 3.5.1 via the "check for updates" button in the application
2. Open any already configured repository (i have OpenSSH as global settings)
3. Click on the pull button
4. You are asked to type the passphrase in. Type the correct passphrase.
5. After pull is finished correctly, click on the pull button again
6. You are asked - again -  to type the passphrase in. Type the correct passphrase.
7. Repeat steps 5-6

## Screenshots

![image](https://user-images.githubusercontent.com/1496592/124876968-e13a1500-dfca-11eb-8429-a83bdd8450b7.png)

## Did this work in previous version of GitExtensions

With 3.4.3.9999  it was working correctly (Ssh passphrase was asked rarely)

## Environment

- Git Extensions 3.5.1.12196
- Build 9099a1e76d01dbbd6896d6b026027db0b68fba96
- Git 2.30.1.windows.1
- Microsoft Windows NT 10.0.19043.0
- .NET Framework 4.8.4390.0
- DPI 120dpi (125% scaling)

skajfes:
I have the same issue, with version 3.5.1.12196 I am asked for the passphrase every time when pushing or pulling from the remote. I have OpenSSH setup and configured on the system and I am not asked for passphrase when pulling from powershell. 

I have noticed that when I open settings there is an error of "Unknown SSH client configured", in red. SSH client is C:\WINDOWS\System32\OpenSSH\ssh.exe, but GitExtensions can't see that file for some reason. 

GitExtensions version 3.5.0.11713 does not have this issue, and behaves correctly.

dopry:
I am having the same issue with 3.5.1.12196. It can not see the folder c:\Windows\System32\OpenSSH.  I can get to the folder in explorer just fine.

[...]

Issue #9353, opened 2021-07-09
Git executable not found in Console


bvarnai:
## Current behaviour

After upgrading to 3.5.1, when I choose the Console tab and type `git --version`, no Git executable found. This was working in version 3.4.3

## Expected behaviour

Git executable should be available in Console (like in Git bash). IMHO without adding `git.exe` to the PATH explicitly.

## Steps to reproduce

Switch to Console and type `git --version`

## Screenshots

![image](https://user-images.githubusercontent.com/7565328/125047075-e36da380-e09e-11eb-9640-a9336732e9c4.png)

## Did this work in previous version of GitExtensions

Yes, in version 3.4.3. 

## Environment

- Git Extensions 3.5.1.12196
- Build 9099a1e76d01dbbd6896d6b026027db0b68fba96
- Git 2.32.0.windows.2
- Microsoft Windows NT 10.0.18363.0
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

mstv:
There had been issues with and fixes for the setting "Path to linux tools (sh)", which is `gitbindir` (old name) in the settings file `%AppData%\Roaming\GitExtensions\GitExtensions\GitExtensions.settings`.
That's why you might need to set this settings again.
Perhaps you had something like "C:\Program Files\Git\bin" and now "C:\Program Files\Git\usr\bin". The latter does not contain `git.exe` usually. The first one uses to contain `sh.exe`.

Only this directory is prepended to the `PATH` in the console (and for user scripts).
The setting "command used to run git"  is used internally only.

bvarnai:
@mstv Thank you! I tried to set `gitbindir` manually but it always restored when the application is started. What did I miss?

vbjay:
Should be usr\bin folder where git is installed.

mstv:
Unfortunately, `git.exe` is located only in `C:\Program Files\Git\bin` not in `...\usr\bin`.

@bvarnai, I guess you have enabled `Check settings at startup`. Its label does not clearly state that it tries to automatically choose the correct paths on every start of GE. 
Since 8eb7f615230cf847860175158bede84204265dc0 (cherry-picked from #9064), `Git\usr\bin` is preferred over `Git\bin` because the first one contains all the tools.

(BTW: I do not configure a global username on purpose.)

![image](https://user-images.githubusercontent.com/36601201/125349977-7425d700-e35e-11eb-8841-c85818c11ea5.png)

[...]

PR #9369, opened 2021-07-12
Do not check if path exists before settings GIT_SSH

gerhardol:
Attempt to handle #9355 #9360 #9348

## Proposed changes

3.4 added a check if the GIT_SSH path exists before setting the variable, to avoid that old incorrect settings are persistent. (This reverts to the Git bundled OpenSSH ssh.exe).

This check may be a reason to why some configurations does not work in 3.5.1.
~~Adding a trace printout to confirm this too.~~
Adding cherry-picked info from master

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitSshHelpers.cs
+++ b/GitCommands/Git/GitSshHelpers.cs
@@ -1,5 +1,4 @@
 using System;
-using System.IO;
 
 namespace GitCommands
 {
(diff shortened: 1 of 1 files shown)


mstv:
Could also be caused by `AutoSolve` never turned off: https://github.com/gitextensions/gitextensions/issues/9353#issuecomment-878575186.

gerhardol:
> 
> 
> Could also be caused by `AutoSolve` never turned off: [#9353 (comment)](https://github.com/gitextensions/gitextensions/issues/9353#issuecomment-878575186).

Was fixing #9348 

Will let a few more users report on this before removing the trace and merge

I do not like that the settings are out of sync with the functionality, but GIT_SSH is something that GE does not use itself, just relayed to Git.

mstv (approved):
Identical to #9385

(Edited because GH/git confused me with "The change was in DoArgumentsRequireSsh()".)

PR #9380, opened 2021-07-15
Recent repo search

### LINKED ISSUE #9302: Recent repository search
## Feature description
Background: I'm working with a lot of different repositories and I need to switch between quite regulary / fast.
It would be great to see some sort of recent repository search where I can just type some part of the path and git extensions will show me a list of all matching repos. Ideally it should be possible to just start typing after opening git extensions.

Maybe just a big searchbar at the top of the start screen?
image

Let me know what you think

davidgiga1993:
Fixes #9302

## Proposed changes

- A new searchbar for filtering the recent / favorite repositories
   - Pressing return will open the first search result
- The default focus has been changed to focus the search textbox once the dashboard gets visible
   -  This allows immediate search when opening gitextensions or when going back (strg+w) to the dashboard
- A new interface has been added for filtering repositories

### Before
image

### After

image

## Test methodology 

- Manual tests:
  - Search for repo, press return
  - Search for random text (no search results), press return
  - Search for non existing repo, press return

## Test environment(s) 

- Windows 10
- UI Scaling 100%

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
@@ -35,6 +35,7 @@
         private void InitializeComponent()
         {
             this.components = new System.ComponentModel.Container();
+            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(UserRepositoriesList));
             System.Windows.Forms.ListViewGroup listViewGroup1 = new System.Windows.Forms.ListViewGroup("Most Recent", System.Windows.Forms.HorizontalAlignment.Left);
             System.Windows.Forms.ListViewGroup listViewGroup2 = new System.Windows.Forms.ListViewGroup("Less Recent", System.Windows.Forms.HorizontalAlignment.Left);
             System.Windows.Forms.ListViewGroup listViewGroup3 = new System.Windows.Forms.ListViewGroup("Other", System.Windows.Forms.HorizontalAlignment.Left);
@@ -42,35 +43,37 @@
             "C:\\Users\\russkie\\AppData\\Roaming\\GitExtensions\\GitExtensions",
             "refs/heads/master",
             "Favourite"}, 0);
-            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(UserRepositoriesList));
             this.toolStripMenuItem1 = new System.Windows.Forms.ToolStripSeparator();
             this.toolStripMenuItem2 = new System.Windows.Forms.ToolStripSeparator();
             this.pnlHeader = new System.Windows.Forms.Panel();
-            this.pnlBody = new System.Windows.Forms.Panel();
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
             this.lblRecentRepositories = new System.Windows.Forms.Label();
+            this.pnlBody = new System.Windows.Forms.Panel();
+            this.tableLayoutPanel2 = new System.Windows.Forms.TableLayoutPanel();
+            this.contextMenuStripRepository = new System.Windows.Forms.ContextMenuStrip(this.components);
+            this.tsmiOpenFolder = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiCategories = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiCategoryNone = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiCategoryAdd = new System.Windows.Forms.ToolStripMenuItem();
-            this.contextMenuStripRepository = new System.Windows.Forms.ContextMenuStrip(this.components);
-            this.contextMenuStripCategory = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.tsmiRemoveFromList = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiRemoveMissingReposFromList = new System.Windows.Forms.ToolStripMenuItem();
-            this.tsmiOpenFolder = new System.Windows.Forms.ToolStripMenuItem();
+            this.imageList1 = new System.Windows.Forms.ImageList(this.components);
+            this.textBoxSearch = new System.Windows.Forms.TextBox();
+            this.contextMenuStripCategory = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.tsmiCategoryRename = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiCategoryDelete = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiCategoryClear = new System.Windows.Forms.ToolStripMenuItem();
+            this.menuStripRecentMenu = new System.Windows.Forms.MenuStrip();
+            this.mnuTop = new System.Windows.Forms.ToolStripMenuItem();
+            this.mnuConfigure = new System.Windows.Forms.ToolStripMenuItem();
             this.listView1 = new GitUI.UserControls.NativeListView();
             this.clmhdrPath = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.clmhdrBranch = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.clmhdrCategory = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
-            this.imageList1 = new System.Windows.Forms.ImageList(this.components);
-            this.menuStripRecentMenu = new System.Windows.Forms.MenuStrip();
-            this.mnuTop = new System.Windows.Forms.ToolStripMenuItem();
-            this.mnuConfigure = new System.Windows.Forms.ToolStripMenuItem();
             this.pnlHeader.SuspendLayout();
-            this.pnlBody.SuspendLayout();
             this.tableLayoutPanel1.SuspendLayout();
+            this.pnlBody.SuspendLayout();
+            this.tableLayoutPanel2.SuspendLayout();
             this.contextMenuStripRepository.SuspendLayout();
             this.contextMenuStripCategory.SuspendLayout();
             this.menuStripRecentMenu.SuspendLayout();
@@ -79,12 +82,12 @@
             // toolStripMenuItem1
             // 
             this.toolStripMenuItem1.Name = "toolStripMenuItem1";
-            this.toolStripMenuItem1.Size = new System.Drawing.Size(221, 6);
+            this.toolStripMenuItem1.Size = new System.Drawing.Size(270, 6);
             // 
             // toolStripMenuItem2
             // 
             this.toolStripMenuItem2.Name = "toolStripMenuItem2";
-            this.toolStripMenuItem2.Size = new System.Drawing.Size(221, 6);
+            this.toolStripMenuItem2.Size = new System.Drawing.Size(270, 6);
             // 
             // pnlHeader
             // 
@@ -97,17 +100,6 @@
             this.pnlHeader.Padding = new System.Windows.Forms.Padding(20, 0, 20, 11);
             this.pnlHeader.Size = new System.Drawing.Size(451, 73);
             this.pnlHeader.TabIndex = 0;
-            // 
-            // pnlBody
-            // 
-            this.pnlBody.Controls.Add(this.listView1);
-            this.pnlBody.Dock = System.Windows.Forms.DockStyle.Fill;
-            this.pnlBody.Location = new System.Drawing.Point(0, 0);
-            this.pnlBody.Margin = new System.Windows.Forms.Padding(0);
-            this.pnlBody.Name = "pnlBody";
-            this.pnlBody.Padding = new System.Windows.Forms.Padding(20);
-            this.pnlBody.Size = new System.Drawing.Size(451, 73);
-            this.pnlBody.TabIndex = 1;
             // 
             // tableLayoutPanel1
             // 
@@ -136,8 +128,56 @@
             this.lblRecentRepositories.Margin = new System.Windows.Forms.Padding(0);
             this.lblRecentRepositories.Name = "lblRecentRepositories";
             this.lblRecentRepositories.Size = new System.Drawing.Size(411, 32);
-            this.lblRecentRepositories.TabIndex = 1;
+            this.lblRecentRepositories.TabIndex = 2;
             this.lblRecentRepositories.Text = "Recent Repositories";
+            // 
+            // pnlBody
+            // 
+            this.pnlBody.Controls.Add(this.tableLayoutPanel2);
+            this.pnlBody.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.pnlBody.Location = new System.Drawing.Point(0, 73);
+            this.pnlBody.Margin = new System.Windows.Forms.Padding(0);
+            this.pnlBody.Name = "pnlBody";
+            this.pnlBody.Padding = new System.Windows.Forms.Padding(20, 18, 20, 3);
+            this.pnlBody.Size = new System.Drawing.Size(451, 210);
+            this.pnlBody.TabIndex = 1;
+            // 
+            // tableLayoutPanel2
+            // 
+            this.tableLayoutPanel2.ColumnCount = 1;
+            this.tableLayoutPanel2.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle());
+            this.tableLayoutPanel2.Controls.Add(this.listView1, 0, 1);
+            this.tableLayoutPanel2.Controls.Add(this.textBoxSearch, 0, 0);
+            this.tableLayoutPanel2.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.tableLayoutPanel2.Location = new System.Drawing.Point(20, 3);
+            this.tableLayoutPanel2.Name = "tableLayoutPanel2";
+            this.tableLayoutPanel2.RowCount = 2;
+            this.tableLayoutPanel2.RowStyles.Add(new System.Windows.Forms.RowStyle(System.Windows.Forms.SizeType.Absolute, 45F));
+            this.tableLayoutPanel2.RowStyles.Add(new System.Windows.Forms.RowStyle());
+            this.tableLayoutPanel2.Size = new System.Drawing.Size(411, 204);
+            this.tableLayoutPanel2.TabIndex = 4;
+            // 
+            // contextMenuStripRepository
+            // 
+            this.contextMenuStripRepository.Items.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.tsmiOpenFolder,
+            this.toolStripMenuItem1,
+            this.tsmiCategories,
+            this.toolStripMenuItem2,
+            this.tsmiRemoveFromList,
+            this.tsmiRemoveMissingReposFromList});
+            this.contextMenuStripRepository.Name = "contextMenuStripRepository";
+            this.contextMenuStripRepository.Size = new System.Drawing.Size(274, 104);
+            this.contextMenuStripRepository.Closed += new System.Windows.Forms.ToolStripDropDownClosedEventHandler(this.contextMenuStrip_Closed);
+            this.contextMenuStripRepository.Opening += new System.ComponentModel.CancelEventHandler(this.contextMenuStrip_Opening);
+            // 
+            // tsmiOpenFolder
+            // 
+            this.tsmiOpenFolder.Image = global::GitUI.Properties.Images.BrowseFileExplorer;
+            this.tsmiOpenFolder.Name = "tsmiOpenFolder";
+            this.tsmiOpenFolder.Size = new System.Drawing.Size(273, 22);
+            this.tsmiOpenFolder.Text = "Show in folder";
+            this.tsmiOpenFolder.Click += new System.EventHandler(this.tsmiOpenFolder_Click);
             // 
             // tsmiCategories
             // 
@@ -145,7 +185,7 @@
             this.tsmiCategoryNone,
             this.tsmiCategoryAdd});
             this.tsmiCategories.Name = "tsmiCategories";
-            this.tsmiCategories.Size = new System.Drawing.Size(224, 22);
+            this.tsmiCategories.Size = new System.Drawing.Size(273, 22);
             this.tsmiCategories.Text = "Categories";
             this.tsmiCategories.DropDownOpening += new System.EventHandler(this.tsmiCategories_DropDownOpening);
             // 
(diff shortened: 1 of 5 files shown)


drewnoakes:
This is a feature I've wanted for ages. Thank you for doing this ❤️ 

Two points:

1. What is the "Actions" text on the right?
2. Can we get a little more space above the text box so it doesn't look as cramped?

image

pmiossec:
>     1. What is the "Actions" text on the right?

Not part of this PR. It's a recent PR #9371 done by @RussKie

davidgiga1993:
> 2. Can we get a little more space above the text box so it doesn't look as cramped?

Like this?

image

drewnoakes:
That does look better. Personally I would increase the space above and below to match the space to the left:

image

This follows more of a "grid system" (a concept in graphic design for creating order in a layout):

image

davidgiga1993:
Thanks for the feedback! Let's try this again 😄 
image

drewnoakes:
LGTM. To get the build working you need to updated the translations as described in https://github.com/gitextensions/gitextensions/wiki/Translations

davidgiga1993:
@drewnoakes Doesn't seem to work? Followed the documentation:
1) Run `dotnet build /p:RunTranslationApp=true`
2) Commit the `English.xlf`

GitUI/Translation/English.xlf
```diff
@@ -8709,6 +8709,14 @@ This will just checkout on the submodule the commit determined by the superproje
...
     </body>
   </file>
+  <file datatype="plaintext" original="Plugin" source-language="en">
+    <body>
+      <trans-unit id="Description.Text">
+        <source>Plugin Manager</source>
+        <target />
+      </trans-unit>
+    </body>
+  </file>
```
RussKie, line 8719: This must not be committed, it is noted in https://github.com/gitextensions/gitextensions/wiki/Translations

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
```diff
@@ -164,74 +204,92 @@ private void InitializeComponent()
...
+            this.imageList1.Images.SetKeyName(0, "source_code.png");
+            // 
+            // textBoxSearch
+            // 
+            this.textBoxSearch.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.textBoxSearch.Location = new System.Drawing.Point(3, 3);
+            this.textBoxSearch.Name = "textBoxSearch";
+            this.textBoxSearch.Size = new System.Drawing.Size(439, 20);
+            this.textBoxSearch.TabIndex = 0;
+            this.textBoxSearch.PlaceholderText = "Search";
```
pmiossec, line 234: I think this string should be set in the constructor and using the `TranslationString` so that the string could be translated.

Also, I don't know what other think but is "Search a repository" clearer?

pmiossec:
@davidgiga1993 I see that you did a sync merge with `master` branch.
We try to avoid that and sync with a rebase.

Do you think you could squash and rebase on top of `master` to keep only 2 commits (the one for the contribution file and the one with your changes). If you could push the contribution file change in another PR, it's even better (that's what we advise).

I can help in doing it if you don't know how to do it...

pmiossec:
@davidgiga1993 The easiest sometimes to solve this situation is in fact to do a 'reset soft' on the commit you synced/merged from `master` and recreate the commits like that:

GE_reset_soft_gif

After that you could compare the local branch and the remote to ensure that you don't have differences before force (with lease) pushing your branch.

RussKie:
We have a rebase guide in the wiki.

GitUI/Translation/English.xlf
davidgiga1993: The wiki states that this only applies to `English.Plugins.xlf` and not `English.xlf`, or am I missing something here?

davidgiga1993:
> @davidgiga1993 I see that you did a sync merge with `master` branch.
> We try to avoid that and sync with a rebase.
> 
> Do you think you could squash and rebase on top of `master` to keep only 2 commits (the one for the contribution file and the one with your changes). If you could push the contribution file change in another PR, it's even better (that's what we advise).
> 
> I can help in doing it if you don't know how to do it...

Sure, no problem! The contrib PR is [here](https://github.com/gitextensions/gitextensions/pull/9384)

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.resx
```diff
@@ -117,18 +117,18 @@
...
-    <value>166, 17</value>
+    <value>422, 17</value>
   </metadata>
   <data name="imageList1.ImageStream" mimetype="application/x-microsoft.net.object.binary.base64">
     <value>
         AAEAAAD/////AQAAAAAAAAAMAgAAAFdTeXN0ZW0uV2luZG93cy5Gb3JtcywgVmVyc2lvbj00LjAuMC4w
         LCBDdWx0dXJlPW5ldXRyYWwsIFB1YmxpY0tleVRva2VuPWI3N2E1YzU2MTkzNGUwODkFAQAAACZTeXN0
         ZW0uV2luZG93cy5Gb3Jtcy5JbWFnZUxpc3RTdHJlYW1lcgEAAAAERGF0YQcCAgAAAAkDAAAADwMAAACy
-        EAAAAk1TRnQBSQFMAwEBAAFoAQABaAEAASABAAEgAQAE/wERAQAI/wFCAU0BNgcAATYDAAEoAwABgAMA
+        EAAAAk1TRnQBSQFMAwEBAAHwAQAB8AEAASABAAEgAQAE/wERAQAI/wFCAU0BNgcAATYDAAEoAwABgAMA
```
pmiossec, line 131: That's strange. Do you know why this image has changed?

davidgiga1993: No, I thought it's just the VS forms editor doing it's thing. But it's not an intentional change.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -8,12 +8,49 @@
...
+        {
+            if (_pattern.Trim().Length == 0)
+            {
+                return;
+            }
+
+            for (var i = repositories.Count - 1; i >= 0; i--)
+            {
+                var repo = repositories[i];
+                if (repo.Path.ToLower().Contains(_pattern))
```
pmiossec, line 38: For me, it's better to lower none of the strings (even the pattern above) and instead use `StringComparison.OrdinalIgnoreCase`

https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -8,12 +8,49 @@
...
+    /// <summary>
+    /// Case insensitive text search based on the repository path
+    /// </summary>
+    public class RepositoryTextFilter : IRepositoryFilter
+    {
+        private readonly string _pattern;
+
+        public RepositoryTextFilter(string pattern)
+        {
+            _pattern = pattern;
```
pmiossec, line 25: Sorry but looking at it again, it seems weird to me...
As long as you trim the string to check the length, the `.Trim()` must be here, no!?!

Either we don't trim or we use the trimmed string as the filter pattern, right?

davidgiga1993: Yes, you're right. It probably would make sense to trim the spaces only from the left (to ignore accidentially added spaces at the beginning of the input) but keep spaces on the right since it's legitim to have a repo path with spaces in it

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
drewnoakes: The comparison should probably use the machine's locale. Not everyone uses Latin characters. If suggest using StringComparison.CurrentCultureIgnoreCase.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
drewnoakes: "Search repositories" seems more idiomatic to me.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
```diff
@@ -136,16 +128,64 @@ private void InitializeComponent()
             this.lblRecentRepositories.Margin = new System.Windows.Forms.Padding(0);
             this.lblRecentRepositories.Name = "lblRecentRepositories";
             this.lblRecentRepositories.Size = new System.Drawing.Size(411, 32);
-            this.lblRecentRepositories.TabIndex = 1;
+            this.lblRecentRepositories.TabIndex = 2;
```
RussKie, line 131: Please check that the tab order is logical

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -21,6 +21,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
     public partial class UserRepositoriesList : GitExtensionsControl
     {
         private readonly TranslationString _groupRecentRepositories = new("Recent repositories");
+        private readonly TranslationString _repositorySearchPlaceholder = new("Search a repository");
```
RussKie, line 24:
```suggestion
        private readonly TranslationString _repositorySearchPlaceholder = new("Search");
```
or
```suggestion
        private readonly TranslationString _repositorySearchPlaceholder = new("Search repositories");
```

image

I don't think we need a mnemonic though.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -96,6 +99,9 @@ public UserRepositoriesList()
 
             DragEnter += OnDragEnter;
             DragDrop += OnDragDrop;
+
+            textBoxSearch.TextChanged += TextBoxSearch_TextChanged;
+            textBoxSearch.KeyDown += TextBoxSearch_KeyDown;
```
RussKie, line 104: Move into the *.designer file

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -857,7 +891,10 @@ private void OnDragEnter(object sender, DragEventArgs e)
             }
         }
 
-        // returns false only if no repository is selected
+        /// <summary>
+        /// Tries to open the currently selected repository
+        /// </summary>
+        /// <returns>False if no repo is selected, true otherwise</returns>
         private bool TryOpenSelectedRepository()
```
RussKie, line 898: * Modify the signature as follows:
    ```suggestion
            private bool TryOpenSelectedRepository(Repository? repository)
    ```
* Update the existing callsites with `TryOpenSelectedRepository(GetSelectedRepository())`
* Remove `OpenRepository` and use `TryOpenSelectedRepository` instead

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.resx
```diff
@@ -117,18 +117,18 @@
   <resheader name="writer">
```
RussKie, line 117: Revert changes to this file

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -633,6 +647,26 @@ private void listView1_MouseClick(object sender, MouseEventArgs e)
             }
         }
 
+        private void TextBoxSearch_TextChanged(object sender, EventArgs e)
+        {
+            ShowRecentRepositories();
```
RussKie, line 652: This is a very handed way of performing the filter. Each time the user will type a character we will be reloading the configuration file, deserializing the list of repositories, recalculating tile sizes and render items.

The filtering must be performed on `recentRepositories` and `favouriteRepositories` collections that are loaded in `ShowRecentRepositories`, adding or removing tiles from the listview as necessary.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -8,12 +8,49 @@
...
+    }
+
+    /// <summary>
+    /// Case insensitive text search based on the repository path
+    /// </summary>
+    public class RepositoryTextFilter : IRepositoryFilter
+    {
+        private readonly string _pattern;
+
+        public RepositoryTextFilter(string pattern)
```
RussKie, line 23: 1. `pattern` is a variable, so it mustn't be set via the .ctor, and pass to `Filter` as an arg instead.
2. In general I'm not convinced this class is needed at all. This functionality can be part of `IUserRepositoriesListController`, e.g. `void IUserRepositoriesListController.FilterRepositories(filter, ListView.ListViewItemCollection repositories)`

RussKie (changes requested):

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
davidgiga1993: I agree, but one question: The `SplitRcentRepos` may trim the length of the repo list, correct? So would it make sense to keep that list in memory (before the split) and apply the filter on that? This way the search will also find items which are not currently shown in the UI. Or would you expect to only filter the items currently visible?

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.Designer.cs
davidgiga1993: Yep, the default focus is on the search box, after that comes the recent repo list, then the left menu. The overall behavior stays the same

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
RussKie: That's a very good Q!
I suppose I could see an expectation to get a result searching for "project" whilst having a repo displayed as `c:\myrepos\so....roject.git` (with the full name `c:\myrepos\somecompany\projects\coolproject.git`). But at the same time I can see why this search result is totally unexpected too.
Personally I'd go for a filter to what's being displayed. That is, if there is not "project" word is visible, then nothing is returned.

@gitextensions/git-extensions-team do you have opnions?

gerhardol: If I understand the question: I would expect to search the full name, not just what happens to be displayed.

davidgiga1993: @RussKie Maybe I maybe formulated the question badly. I meant the number of items shown (defined by `maxRecentRepositories`) might be less than the number of repos known to git extensions.
As I user I would expect the search to search for "any repo known to git extensions matching the patch" rather than "search in the couple of items shown in the UI".
Therefore the search needs to cache the raw list returned by `ThreadHelper.JoinableTaskFactory.Run(() => RepositoryHistoryManager.Locals.LoadRecentHistoryAsync());`.
Do you agree?

RussKie: > RepositoryHistoryManager.Locals.LoadRecentHistoryAsync()

We trim the history size upon save, so the list you get is all we have:
https://github.com/gitextensions/gitextensions/blob/a32fe7a6b79cb3af43fed302a7b548687156d720/GitCommands/UserRepositoryHistory/LocalRepositoryManager.cs#L292-L313

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -857,28 +889,38 @@ private void OnDragEnter(object sender, DragEventArgs e)
...
 
+            OpenRepository(repository);
+            return true;
+        }
+
+        /// <summary>
+        /// Opens the given repository
+        /// </summary>
+        /// <param name="selected">Repository</param>
+        private void OpenRepository(Repository selected)
```
RussKie, line 911:
```suggestion
        private void OpenRepository(Repository repository)
```

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
davidgiga1993: Ah I see, thanks for clarifying. So for now I just cache the result from `LoadRecentHistoryAsync()` to avoid loading the data from disk all the time. The search is still applied on the repo list itself.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -633,6 +645,26 @@ private void listView1_MouseClick(object sender, MouseEventArgs e)
...
+            if (e.KeyCode == Keys.Enter)
+            {
+                // Open the first repo in the list
+                var items = listView1.Items;
+                if (items.Count == 0)
+                {
+                    return;
+                }
+
+                OpenRepository((Repository)items[0].Tag);
```
RussKie, line 664:
```suggestion
                TryOpenRepository(items[0].Tag as Repository);
```

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
RussKie, line 911: There's no need for the new method.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -13,15 +13,21 @@ public interface IUserRepositoriesListController
...
 
     public sealed class UserRepositoriesListController : IUserRepositoriesListController
     {
         private readonly ILocalRepositoryManager _localRepositoryManager;
         private readonly IInvalidRepositoryRemover _invalidRepositoryRemover;
 
+        // Holds the raw, unfiltered list of repositories.
+        // This is done to allow fast filtering of all known repos.
+        private IList<Repository>? _AllRecentRepositories;
+        private IList<Repository>? _AllFavoriteRepositories;
```
RussKie, line 29:
```suggestion
        private IReadOnlyList<Repository>? _allRecentRepositories;
        private IReadOnlyList<Repository>? _allFavoriteRepositories;
```

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -340,6 +343,15 @@ void BindRepositories(IReadOnlyList<RecentRepoInfo> repos, bool isFavourite)
             }
         }
 
+        protected override void OnVisibleChanged(EventArgs e)
+        {
+            base.OnVisibleChanged(e);
+
+            // Reset the recent repo cache and the search
+            _controller.ClearCache();
```
RussKie, line 351: Why is this necessary?

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -85,5 +110,26 @@ public bool IsValidGitWorkingDir(string path)
 
         public bool RemoveInvalidRepository(string path)
            => _invalidRepositoryRemover.ShowDeleteInvalidRepositoryDialog(path);
+
+        private static IList<Repository> Filter(IList<Repository> repositories, string pattern)
```
RussKie, line 114:
```suggestion
        private static IReadOnlyList<Repository> Filter(IReadOnlyList<Repository> repositories, string pattern)
```

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
```diff
@@ -70,11 +85,21 @@ public bool IsValidGitWorkingDir(string path)
                 SortMostRecentRepos = AppSettings.SortMostRecentRepos
             };
 
-            var repositories = ThreadHelper.JoinableTaskFactory.Run(() => RepositoryHistoryManager.Locals.LoadRecentHistoryAsync());
+            if (_AllRecentRepositories == null)
+            {
+                _AllRecentRepositories = ThreadHelper.JoinableTaskFactory.Run(() => RepositoryHistoryManager.Locals.LoadRecentHistoryAsync());
+            }
```
RussKie, line 91:
```suggestion
            _allRecentRepositories ??= ThreadHelper.JoinableTaskFactory.Run(() => RepositoryHistoryManager.Locals.LoadRecentHistoryAsync());
```

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
davidgiga1993: The `LoadFavourite/RecentHistoryAsync` calls return an `IList` not an `IReadOnlyList`

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
davidgiga1993: The intention was to clear the cache when the view is shown again to update the items shown in case something changed. This solution had a couple of flaws which I just resolved by clearing the cache in the `ShowRecentRepositories` (if required)

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
RussKie: This is a covariant conversion (I think this is the right word), i.e. you can go from `List<T>` to `IReadOnlyList<T>`.

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -517,7 +533,7 @@ private void UpdateCategoryName(string originalName, string? newName)
         private void contextMenuStrip_Closed(object sender, ToolStripDropDownClosedEventArgs e)
         {
             _rightClickedItem = null;
-            ShowRecentRepositories();
+            ShowRecentRepositories(false);
```
RussKie, line 536:
```suggestion
            ShowRecentRepositories(reloadData: false);
```

RussKie (commented):
👍 
Few outstanding comments

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesListController.cs
davidgiga1993: Yes you're right for `List<T>` but the existing methods return an `IList<T>` which does not implement `IReadOnlyList<T>`

RussKie:
Please update English translations and re-submit the pull-request. Refer to https://github.com/gitextensions/gitextensions/wiki/Translations

RussKie (approved):
👍

drewnoakes (approved):

RussKie:
Thank you!

davidgiga1993:
Is there already a plan in which version this change will be included?

pmiossec:
There is a milestone to manage the next version:
https://github.com/gitextensions/gitextensions/milestone/60

But I don't think it is up to date.

Because we switched to .net 5 for this version there are things to fix before releasing.
Maybe #9402 (and similar one #9491) should be added.
Fixing the installer need to be done from what I understood.
And some more that are discovered some time to time.

No planned date at the moment.
So if you want to have this feature, you'd better build your own version or use the portable development version built after each PR merged in `master`

davidgiga1993:
Thanks for your replies! Yes I was looking at the milestone and was a bit puzzled that it's missing some stuff, but that explains it 👍

PR #9385, opened 2021-07-16
Do not check if path exists before settings GIT_SSH

gerhardol:
Related to #9355 #9360 #9348

Similar to #9369 that still have trace printouts and therefore is WIP (and should get cherry-pick comment when this merged).

## Proposed changes

The path in GIT_SSH may be accessible for git.exe even if not for GE

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitSshHelpers.cs
+++ b/GitCommands/Git/GitSshHelpers.cs
@@ -28,14 +28,10 @@ namespace GitCommands
         public static void SetSsh(string path)
         {
             // Git will use the embedded OpenSSH ssh.exe if empty/unset
-            if (!string.IsNullOrEmpty(path) && !File.Exists(path))
-            {
-                path = "";
-            }
-
             Environment.SetEnvironmentVariable("GIT_SSH", path, EnvironmentVariableTarget.Process);
         }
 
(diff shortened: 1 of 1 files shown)


RussKie (approved):

GitCommands/Git/GitSshHelpers.cs
```diff
@@ -28,14 +28,10 @@ bool DoArgumentsRequireSsh()
         public static void SetSsh(string path)
         {
             // Git will use the embedded OpenSSH ssh.exe if empty/unset
-            if (!string.IsNullOrEmpty(path) && !File.Exists(path))
-            {
-                path = "";
-            }
-
             Environment.SetEnvironmentVariable("GIT_SSH", path, EnvironmentVariableTarget.Process);
```
mstv, line 31: I would remove also the comment.

mstv (approved):

GitCommands/Git/GitSshHelpers.cs
gerhardol: That comment still applies to setting the environment
(It would be easier to let GIT_SSH override the GE settings and not set the value if empty (no need to override), but that would require more changes for something not so important.)

RussKie:
Merge?

Issue #4011, opened 2017-09-19
Inconsistencies in how path to plink.exe is obtained


CookiePLMonster:
**Do you want to request a *feature* or report a *bug*?**

Most likely a bug, or at least a design issue.

**What is the current behavior?**

There are two places Git Extensions stores paths to plink.exe (as well as other SSH-related tools, but those might be non-issue) - Windows Registry and GitExt Settings.  Different functionalities of the tool use different paths, so in case of desyncs incosistencies can occur. Therefore, it **is** possible to desync registry and GitExt Settings without modifying registry and/or settings manually.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

1. Make sure path to plink.exe in **SSH -> Settings** is set to a non-default path.
2. Uninstall Git Extensions.
3. Install Git Extensions again (in my tests, I used 2.50.02).
4. Open regedit and navigate to `HKEY_CURRENT_USER\Software\GitExtensions` - notice `gitssh` is set to a default path, ie. plink.exe shipped with GitExt.
5. Run Git Extensions. **DO NOT CLICK APPLY NOR OK - click Cancel** - it is explicitly said that "Clicking Cancel will NOT revert any changes made", so it feels like a perfectly valid course of action.
6. Notice paths in registry did not change, but Settings -> SSH still contains custom paths - **these will not be placed in registry until you press Apply**
7. To make testing easier, navigate to GitExt install directory and remove PuTTY from there (so it's obvious when registry path is used).
8. Open any git repo with GitExt, navigate to Remote repositories and click Test Connection on any SSH-based remote - notice test succeeds, and therefore **Test Connection does not use `gitssh` registry entry**.
9. Try to Pull from any SSH-based remote - notice it fails (because it tries to run an inexistant plink.exe), and therefore **Git Pull ignores settings from SSH -> Configure PuTTY and only relies on registry entry**.

Additionally:
Even after applying settings in GitExt, notice that **only** `gitssh` registry entry gets updated - entries `pageant`, `plink`  and `puttygen` still point to default paths no matter what.

**What is the expected behavior?**

It's either not possible to desync registry and GitExt settings **or** registry entries are removed and all functionalities rely solely on GitExt Settings.

**Environment you encounter the issue:**

- GitExtensions version: 2.50.02, however a quick check with a self-compiled GitExt proves this is still an issue
- GIT version: 2.14.1 (64-bit version)
[...]

PR #9386, opened 2021-07-16
Remove PuTTY binaries from installer

gerhardol:
## Proposed changes

The PuTTY binaries distributed with GE are almost two years old now. Users should maintain the version separate from GE itself.
(Similar was changed for Git, KDiff3 some years ago.)

This removes the PuTTY binaries from the installation and changes the searchpaths in Settings when changing to PuTTY to searh "Program Files\PuTTY" first.

TODO set defaults for putty in installer or possibly in startup code and test installer.
No migration is planned, users will have to to settings to change it.
(If this is considered required, there could be a hardcoded check for <ge path>/PuTTY and run the pathfinder for known locations as used in settings, but I do not want to spend too much time on this.)

Is this the way to go?

## Test methodology 

Manual

----


Setup/Product.wxs
```diff
@@ -895,13 +886,13 @@
...
+        <RegistryValue Name="plink" Value="[ProgramFilesFolder]plink.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
       </Component>
       <Component Id="pageant.reg" Guid="*">
-        <RegistryValue Name="pageant" Value="[INSTALLDIR]PuTTY\pageant.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
+        <RegistryValue Name="pageant" Value="[ProgramFilesFolder]pageant.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
       </Component>
       <Component Id="puttygen.reg" Guid="*">
-        <RegistryValue Name="puttygen" Value="[INSTALLDIR]PuTTY\puttygen.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
+        <RegistryValue Name="puttygen" Value="[ProgramFilesFolder]puttygen.exe" Root="HKCU" Key="$(var.AppRegKey)" Type="string" />
       </Component>
```
RussKie, line 896: We don't need this either, since we're no longer ship or install it, we mustn't register those either.

RussKie (commented):
On the settings page we can show the actual settings, if we can find those, or a link to the wiki page with instructions.

Setup/Product.wxs
gerhardol: I thought these set the path to the PuTTY binaries in the GE registry, which still is needed.
There is detection of PuTTY path in the GE settings when setting PuTTY. If PuTTY is removed completely from the installer and the user configures this in the setting, it is easier to handle. But I have been flamed when suggesting that previously. (I still prefer that myself.)

For 3.5 I would like to update the binaries, there is a 0.76 version coming soon that I hope will be in time for a 3.5.2 release.
The tuning of the path should probably be in 3.5.2 too (to select ProgramFiles/PuTTY of the distributed version).

> On the settings page we can show the actual settings, if we can find those,

When selecting PuTTY the hardcoded directories (PUTTY, TortoiseGit, TortoiseSVN) are searched, you can set manually after that.

> or a link to the wiki page with instructions.

A link could be put there regardless
But I want to keep the complexity for the change to a minimum.

RussKie: > For 3.5 I would like to update the binaries, there is a 0.76 version coming soon that I hope will be in time for a 3.5.2 release.
> The tuning of the path should probably be in 3.5.2 too (to select ProgramFiles/PuTTY of the distributed version).

👍 
Yes, we're few things for 3.5.2. With 27K downloads I think we can assume the main lot of bugs has been reported.



> If PuTTY is removed completely from the installer and the user configures this in the setting, it is easier to handle.

That'd be my preference.

gerhardol: Removed PuTTY from the installer (it will be much easier to test this now...).
Documentation will have to be changed too.

I will consider a migration, may open the settings if pointing to the removed binaries, but want to have it really simple.

CookiePLMonster:
Moving forward with this change would probably kind of resolve #4011 by design.

gerhardol:
> Moving forward with this change would probably kind of resolve #4011 by design.

There are a few other issues as well, due to the installer (sometimes?) manipulates the paths.
Waiting for the wix installer to work again with .NET5 before proceeding.

RussKie:
Here's a build to test: https://ci.appveyor.com/api/buildjobs/2svhhdi4ank88rgg/artifacts/GitExtensions-4.0.0.13119-alpha-aabfaf14b.msi

gerhardol:
Upgrade and new install worked for me

PR #9387, opened 2021-07-17
Enable use of the designer

### LINKED ISSUE #9388: How to use the VS Designer? Always getting errors
### Discussed in https://github.com/gitextensions/gitextensions/discussions/9376

<div type='discussions-op-text'>

<sup>Originally posted by **davidgiga1993** July 14, 2021</sup>
Hi,
I'm trying to edit some GUI items but no matter which control I open in the editor I get the following exception:
```
at GitUI.ThreadHelper.ThrowIfNotOnUIThread(String callerMemberName) in D:\Development\Projects\gitextensions\GitExtUtils\GitUI\ThreadHelper.cs:line 59
at ResourceManager.GitExtensionsControlInitialiser..ctor(GitExtensionsControl control) in D:\Development\Projects\gitextensions\ResourceManager\GitExtensionsControlInitialiser.cs:line 25
at ResourceManager.GitExtensionsControl..ctor() in D:\Development\Projects\gitextensions\ResourceManager\GitExtensionsControl.cs:line 20
at GitUI.GitModuleControl..ctor() in D:\Development\Projects\gitextensions\GitUI\GitModuleControl.cs:line 108  
```
Looks like the condition to check if the code is executed by the designer doesn't work as expected?
Or am I missing something here?

I'm using a fresh VS2019 install with the .net desktop development feature installed</div>

RussKie:
This is likely an ongoing work where we'll need to check every form and/or control we edit and apply `IsDesignMode` check to avoid executing unnecessary code in the Windows Forms designer.

Resolves #9388

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1919,10 +1919,9 @@ namespace GitCommands
             set => SetBool("Log.CaptureCallStacks", value);
         }
 
-        public static bool IsPortable()
-        {
-            return Properties.Settings.Default.IsPortable;
-        }
+        // There is a bug in .NET/.NET Designer that fails to execute Properties.Settings.Default call.
+        // Return false whilst we're in the designer.
+        public static bool IsPortable() => !IsDesignMode && Properties.Settings.Default.IsPortable;
 
         public static bool WriteErrorLog
         {
@@ -2126,6 +2125,22 @@ namespace GitCommands
             string availableEncodings = AvailableEncodings.Values.Select(e => e.HeaderName).Join(";");
             availableEncodings = availableEncodings.Replace(Encoding.Default.HeaderName, "Default");
             SetString("AvailableEncodings", availableEncodings);
+        }
+
+        private static bool? _isDesignMode;
+
+        private static bool IsDesignMode
+        {
+            get
+            {
+                if (_isDesignMode is null)
+                {
+                    string processName = Process.GetCurrentProcess().ProcessName.ToLowerInvariant();
+                    _isDesignMode = processName.Contains("devenv") || processName.Contains("designtoolsserver");
+                }
+
+                return _isDesignMode.Value;
+            }
         }
 
         internal static TestAccessor GetTestAccessor() => new();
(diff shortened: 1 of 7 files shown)


GitCommands/Settings/AppSettings.cs
```diff
@@ -2128,6 +2127,22 @@ private static void SaveEncodings()
...
+            {
+                if (_isDesignMode is null)
+                {
+                    string processName = Process.GetCurrentProcess().ProcessName.ToLowerInvariant();
+                    _isDesignMode = processName.Contains("devenv") || processName.Contains("designtoolsserver");
+                }
+
+                return _isDesignMode.Value;
+            }
+        }
```
RussKie, line 2144: @KlausLoeffelmann this seems the only viable way to detect the designer I found.

ResourceManager/GitExtensionsControlInitialiser.cs
```diff
@@ -8,38 +10,60 @@ namespace ResourceManager
 {
     internal sealed class GitExtensionsControlInitialiser
     {
+        private static bool? _isDesignMode;
         private readonly ITranslate _translate;
 
-        /// <summary>indicates whether the initialisation has been signalled as complete.</summary>
+        // Indicates whether the initialisation has been signalled as complete.
         private bool _initialiseCompleteCalled;
```
mstv, line 17: ❔ Shall private members be hidden from xmldoc?

mstv (approved):
LGTM
Thank you!

vbjay:
Not that I use it but what about https://www.jetbrains.com/rider/
detecting devenv only might be an issue.

gerhardol (commented):
Looks OK, have not used it
(but I only accidentally open the designer.)

ResourceManager/GitExtensionsControlInitialiser.cs
RussKie: Xml-docs are generally applied to public API, as they are used to generate documentation. Annotations of non public API is optional. In this specific case I didn't think it was necessary

RussKie:
> Not that I use it but what about [jetbrains.com/rider](https://www.jetbrains.com/rider/) detecting devenv only might be an issue.

No idea. I don't use Rider, so I can't comment on this.

dasiths:
Sorry I'm late. @RussKie @vbjay  the other way to do this is to have a static that gets during app startup. Which indicates proper loading. All other instances can be treated as designer mode. This will support non VS IDE's as well.

PR #8809, opened 2021-02-01
Improve RevisionDataGridView responsiveness

wischi-chr:
If one rapidly scrolls through the revision data grid, the datagrid is no longer updated/painted, so I took a closer look at
`./gitextensions/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs` and found a few things that should probably be changed.

Currently `RunBackgroundAsync` always runs in the background, uses exceptions for control flow in a tight loop (basically throwing an exception 5 times per second) and using a queue to process `BackgroundUpdateAsync`, which causes it to be a lot more executed than really needed.

## Proposed changes
1. Don't use exceptions for control flow
2. Get rid of the background task
3. Replace queue with a mechanism that only updates when needed.
4. Make sure the DataGridView is redrawn even with massive amounts of scroll events.

Let's start with point three. I added a new nested class called `BackgroundUpdater` that is similar to the [rx debounce operator](http://reactivex.io/documentation/operators/debounce.html). It allows background updates to be skipped if there is already one in progress, and also makes sure that the update runs again if a request comes in while the background update is in progress. Multiple requests (while the update is running) don't queue up, the update is just flagged for a rerun. This change also fixes point 1 and 2.

Sadly at that point, point 4 is still not fixed, when scrolling really fast (for example with a mouse that supports free mouse-wheel spinning), because the message pump from the application is flooded with [WM_CTLCOLORSCROLLBAR](https://docs.microsoft.com/en-us/windows/win32/controls/wm-ctlcolorscrollbar) messages which prevent redrawing the control.

To fix that we track the last time the DataGrid was painted and injecting a WM_PAINT message if needed - forcing the control to repaint.

The changes can be tested by rapidly scrolling through the RevisionDataGridView with the current master and with this PR.

## Test methodology

- Manual testing (without rx it would probably be pretty hard to test the current `RevisionDataGridView` implementation automatically
- Manual performance comparison against the current master branch (cbc8c05efeae68786ba7db2cad2ed24daebaeb6f)

## Test environment(s)

- git 2.30.0.windows.1
- Windows 10

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9393, opened 2021-07-18
Show artificial commits also for filtered grid

gerhardol:
Follow up to #9335
See long term goals in #7598 (comment)

## Proposed changes

Artificial commits are added to the HEAD. If commits are filtered, HEAD may not appear.
This occurs both for files filtered in Browse and FileHistory.
Note that FileHistory for a file in HEAD would show Artificial, but not history for any other file.

This change inserts the artificial commits first if not added to the commits.

Also change the confusing diff message (especially when selecting artificial commits without changes).

Note that the first seen parent to HEAD could be queried, the artificial commits be inserted there. This adds another for a call that may delay display which is not worth it. 

~~Note: For filters where the file do exist in HEAD this is a minor regression, the Index is no longer connected to the head.~~
~~If HEAD is not displayed, the artificial commits could be connected to next commit too.~~
~~This was ignored as either all commits have to be parsed before they are displayed or the rendered grid will have to recalculated when the commits are reinserted.~~

~~If "follow renamed files" is moved from FileHistory to RevisionReader (or RevisionGridControl) so it can be used by Browse filters too,~~
~~this info is mostly calculated (even if --follow seem to drop some commits, seen in #9392 so not in every situation).~~
 
### Before

file in HEAD

image

No changes in Index

image

not in HEAD

image

### After

(not updated screenshot, artificial is connected to HEAD)

image

image

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -178,40 +178,70 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         }
 
         /// <summary>
-        /// Add a single revision from the git log.
+        /// Add a single revision from the git log to the graph, including segments to parents.
         /// </summary>
-        public void Add(GitRevision revision, RevisionNodeFlags types)
+        /// <param name="revision">The revision to add.</param>
+        /// <param name="types">The graph node flags.</param>
+        /// <param name="insertAsFirst">Insert the (artificial) revision first in the graph.</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types, bool insertAsFirst = false)
         {
+            bool updateParents = true;
             if (!_nodeByObjectId.TryGetValue(revision.ObjectId, out RevisionGraphRevision revisionGraphRevision))
             {
-                // This revision is added from the log, but not seen before. This is probably a root node (new branch) OR the revisions
-                // are not in topo order. If this the case, we deal with it later.
-                revisionGraphRevision = new RevisionGraphRevision(revision.ObjectId, ++_maxScore);
+                int score = insertAsFirst
 
+                    // Insert first artificial (WorkTree) before first existing
+                    // _maxScore does not need to be updated
+                    ? -1
+
+                    // This revision is added from the log, but not seen before. This is probably a root node (new branch)
+                    // OR the revisions are not in topo order. If this the case, we deal with it later.
+                    : ++_maxScore;
+                revisionGraphRevision = new RevisionGraphRevision(revision.ObjectId, score);
                 _nodeByObjectId.TryAdd(revision.ObjectId, revisionGraphRevision);
             }
             else
             {
                 // This revision was added earlier, but is now found in the log.
                 // Increase the score to the current maxScore to keep the order in tact.
-                revisionGraphRevision.EnsureScoreIsAbove(++_maxScore);
+                if (!insertAsFirst)
+                {
+                    revisionGraphRevision.EnsureScoreIsAbove(++_maxScore);
+                }
+                else
+                {
+                    // Second artificial (Index), score already set
+                    updateParents = false;
+                }
             }
 
             // This revision may have been added as a parent before. Probably only the ObjectId is known. Set all the other properties.
             revisionGraphRevision.GitRevision = revision;
             revisionGraphRevision.ApplyFlags(types);
 
-            // No build the revisions parent/child structure. The parents need to added here. The child structure is kept in synch in
+            // Build the revisions parent/child structure. The parents need to added here. The child structure is kept in synch in
             // the RevisionGraphRevision class.
-            if (revision.ParentIds is not null)
+            if (!updateParents)
+            {
+                // score set already for the commit, force reorder
+                // Parent (HEAD) not in the graph, no segment added
+                _reorder = true;
+            }
+            else if (revision.ParentIds is not null)
             {
                 foreach (ObjectId parentObjectId in revision.ParentIds)
                 {
                     if (!_nodeByObjectId.TryGetValue(parentObjectId, out RevisionGraphRevision parentRevisionGraphRevision))
                     {
-                        // This parent is not loaded before. Create a new (partial) revision. We will complete the info in the revision
-                        // when this revision is loaded from the log.
-                        parentRevisionGraphRevision = new RevisionGraphRevision(parentObjectId, ++_maxScore);
+                        int score = insertAsFirst
+
+                            // After first inserted (as 0)
+                            ? revisionGraphRevision.Score + 1
+
+                            // This parent is not loaded before. Create a new (partial) revision. We will complete the info in the revision
+                            // when this revision is loaded from the log.
+                            : ++_maxScore;
+                        parentRevisionGraphRevision = new RevisionGraphRevision(parentObjectId, score);
                         _nodeByObjectId.TryAdd(parentObjectId, parentRevisionGraphRevision);
                     }
                     else
@@ -442,7 +472,7 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
             // Reset the reorder flag and the orderedUntilScore. This makes sure it isn't marked dirty before we even got to
             // rebuilding it.
-            _orderedUntilScore = 0;
+            _orderedUntilScore = int.MinValue;
             _reorder = false;
 
             // Use a local variable, because the cached list can be reset
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -299,9 +299,16 @@ namespace GitUI.UserControls.RevisionGrid
             e.Handled = true;
         }
 
-        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None)
+        /// <summary>
+        /// Add a single revision from the git log to the graph, including segments to parents.
+        /// Update visible rows if needed.
+        /// </summary>
+        /// <param name="revision">The revision to add.</param>
+        /// <param name="types">The graph node flags.</param>
+        /// <param name="insertAsFirst">Insert the (artificial) revision first in the graph.</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool insertAsFirst = false)
         {
-            _revisionGraph.Add(revision, types);
+            _revisionGraph.Add(revision, types, insertAsFirst);
 
             if (ToBeSelectedObjectIds.Contains(revision.ObjectId))
             {
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -846,7 +846,8 @@ namespace GitUI
             _isRefreshingRevisions = true;
             ShowLoading();
 
-            var firstRevisionReceived = false;
+            bool firstRevisionReceived = false;
+            bool headIsHandled = false;
 
             try
             {
@@ -1131,14 +1132,12 @@ namespace GitUI
                     this.InvokeAsync(() => { ShowLoading(false); }).FileAndForget();
                 }
 
-                var isCurrentCheckout = revision.ObjectId.Equals(CurrentCheckout);
+                bool isCurrentCheckout = revision.ObjectId.Equals(CurrentCheckout);
 
-                if (isCurrentCheckout &&
-                    ShowUncommittedChangesIfPossible &&
-                    AppSettings.RevisionGraphShowWorkingDirChanges &&
-                    !Module.IsBareRepository())
+                if (isCurrentCheckout)
                 {
-                    AddArtificialRevisions(revision.ObjectId);
+                    headIsHandled = true;
+                    AddArtificialRevisions(insertAsFirst: false);
                 }
 
                 var flags = RevisionNodeFlags.None;
@@ -1161,47 +1160,55 @@ namespace GitUI
                 _gridView.Add(revision, flags);
 
                 return;
+            }
 
-                void AddArtificialRevisions(ObjectId filteredCurrentCheckout)
+            void AddArtificialRevisions(bool insertAsFirst)
+            {
+                if (!ShowUncommittedChangesIfPossible
+                    || !AppSettings.RevisionGraphShowWorkingDirChanges
+                    || Module.IsBareRepository())
                 {
-                    _indexChangeCount = new ArtificialCommitChangeCount();
-                    _workTreeChangeCount = new ArtificialCommitChangeCount();
-
-                    var userName = Module.GetEffectiveSetting(SettingKeyString.UserName);
-                    var userEmail = Module.GetEffectiveSetting(SettingKeyString.UserEmail);
-
-                    // Add working directory as virtual commit
-                    GitRevision workTreeRev = new(ObjectId.WorkTreeId)
-                    {
-                        Author = userName,
-                        AuthorUnixTime = 0,
-                        AuthorEmail = userEmail,
-                        Committer = userName,
-                        CommitUnixTime = 0,
-                        CommitterEmail = userEmail,
-                        Subject = ResourceManager.TranslatedStrings.Workspace,
-                        ParentIds = new[] { ObjectId.IndexId },
-                        HasNotes = true
-                    };
-                    _gridView.Add(workTreeRev);
-
-                    // Add index as virtual commit
-                    GitRevision indexRev = new(ObjectId.IndexId)
-                    {
-                        Author = userName,
-                        AuthorUnixTime = 0,
-                        AuthorEmail = userEmail,
-                        Committer = userName,
-                        CommitUnixTime = 0,
-                        CommitterEmail = userEmail,
-                        Subject = ResourceManager.TranslatedStrings.Index,
-                        ParentIds = new[] { filteredCurrentCheckout },
-                        HasNotes = true
-                    };
-                    _gridView.Add(indexRev);
-
-                    UpdateArtificialCommitCount(_artificialStatus, workTreeRev, indexRev);
+                    return;
                 }
+
+                _indexChangeCount = new ArtificialCommitChangeCount();
+                _workTreeChangeCount = new ArtificialCommitChangeCount();
+
+                var userName = Module.GetEffectiveSetting(SettingKeyString.UserName);
+                var userEmail = Module.GetEffectiveSetting(SettingKeyString.UserEmail);
+
+                // Add working directory as an artificial commit
+                GitRevision workTreeRev = new(ObjectId.WorkTreeId)
+                {
+                    Author = userName,
+                    AuthorUnixTime = 0,
+                    AuthorEmail = userEmail,
+                    Committer = userName,
+                    CommitUnixTime = 0,
+                    CommitterEmail = userEmail,
+                    Subject = ResourceManager.TranslatedStrings.Workspace,
+                    ParentIds = new[] { ObjectId.IndexId },
+                    HasNotes = true
+                };
+                _gridView.Add(workTreeRev, insertAsFirst: insertAsFirst);
+
+                // Add index as an artificial commit
+                GitRevision indexRev = new(ObjectId.IndexId)
+                {
+                    Author = userName,
+                    AuthorUnixTime = 0,
+                    AuthorEmail = userEmail,
+                    Committer = userName,
+                    CommitUnixTime = 0,
+                    CommitterEmail = userEmail,
+                    Subject = ResourceManager.TranslatedStrings.Index,
+                    ParentIds = new[] { CurrentCheckout },
+                    HasNotes = true
+                };
+
+                _gridView.Add(indexRev, insertAsFirst: insertAsFirst);
+
+                UpdateArtificialCommitCount(_artificialStatus, workTreeRev, indexRev);
             }
 
             void OnRevisionReaderError(Exception exception)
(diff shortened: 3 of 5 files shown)


GitCommands/RevisionReader.cs
```diff
@@ -79,6 +79,11 @@ public sealed class RevisionReader : IDisposable
                     });
         }
 
+        /// <summary>
+        /// Parents require rewrite, all commits may not be included
+        /// </summary>
+        public bool ParentsAreRewriten { get; private set; } = false;
```
mstv, line 85:
```suggestion
        public bool ParentsAreRewritten { get; private set; } = false;
```

GitCommands/RevisionReader.cs
```diff
@@ -79,6 +79,11 @@ public sealed class RevisionReader : IDisposable
                     });
         }
 
+        /// <summary>
+        /// Parents require rewrite, all commits may not be included
```
mstv, line 83: ❔
```suggestion
        /// Parents require rewrite, not all commits may be included
```

GitUI/CommandsDialogs/FormFileHistory.cs
```diff
@@ -24,6 +24,7 @@ public sealed partial class FormFileHistory : GitModuleForm
         private const string FormBrowseName = "FormBrowse";
 
         private readonly TranslationString _buildReportTabCaption = new("Build Report");
+        private readonly TranslationString _noChanges = new("No changes.");
```
mstv, line 27: `FileStatusList` has this without the trailing dot.
```suggestion
        private readonly TranslationString _noChanges = new("No changes");
```

mstv (commented):
not reviewed completely, just nits

gerhardol:
> not reviewed completely, just nits

You can wait with a formal review as I will need to change internal api some to connect artificial to head in some more situations.

Review the appearance though

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -180,7 +180,7 @@ public void HighlightBranch(ObjectId id)
         /// <summary>
         /// Add a single revision from the git log.
         /// </summary>
-        public void Add(GitRevision revision, RevisionNodeFlags types)
+        public void Add(GitRevision revision, RevisionNodeFlags types, bool addParentGraph = true)
```
RussKie, line 183: Please add xml-doc explaining the new argument.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -299,9 +299,9 @@ private void OnCellPainting(object sender, DataGridViewCellPaintingEventArgs e)
             e.Handled = true;
         }
 
-        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None)
+        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool addParentGraph = true)
```
RussKie, line 302: Ditto. Bonus points for xml-doc'ing everything on this method.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1035,15 +1040,22 @@ void OnRevisionRead(GitRevision revision)
 
                 return;
 
-                void AddArtificialRevisions(ObjectId filteredCurrentCheckout)
+                void AddArtificialRevisionsIfPossible(bool addParentGraph)
```
RussKie, line 1043:
```suggestion
                void AddArtificialRevisions(bool addParentGraph)
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1068,10 +1080,11 @@ void AddArtificialRevisions(ObjectId filteredCurrentCheckout)
                         CommitUnixTime = 0,
                         CommitterEmail = userEmail,
                         Subject = ResourceManager.TranslatedStrings.Index,
-                        ParentIds = new[] { filteredCurrentCheckout },
+                        ParentIds = new[] { CurrentCheckout },
                         HasNotes = true
                     };
-                    _gridView.Add(indexRev);
+
+                    _gridView.Add(indexRev, addParentGraph: addParentGraph);
```
RussKie, line 1087:
```suggestion
                    _gridView.Add(indexRev, addParentGraph);
```

GitCommands/RevisionReader.cs
RussKie: I re-read the xml-doc and the name over and over, and I still can't grasp the intent... 🤕

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
mstv: Gerhard does this only if necessary, namely if default arguments are in the play.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: The common checks are in this method so it adding also for non fatal reasons (null, empty etc) so I believe the name should be changed.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: Will update

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
gerhardol: Will update

GitUI/CommandsDialogs/FormFileHistory.cs
gerhardol: Will use a common string (without a trailing dot)

GitCommands/RevisionReader.cs
gerhardol: ok

GitCommands/RevisionReader.cs
gerhardol: The parents for the commits are rewritten for the filter and all commits may not be included

RussKie:
> The parents for the commits are rewritten for the filter and all commits
may not be included

This is what I'm having troubles understanding :) What does "parents are
rewritten" - to me it sounds like we're rewriting history, which we aren't
obviously. We are filtering commits in a way that may exclude immediate
parents of the selected commit, is that what you mean?

gerhardol:
> > The parents for the commits are rewritten for the filter and all commits
> may not be included
> 
> This is what I'm having troubles understanding :) What does "parents are
> rewritten" - to me it sounds like we're rewriting history, which we aren't
> obviously. We are filtering commits in a way that may exclude immediate
> parents of the selected commit, is that what you mean?

 
https://git-scm.com/docs/git-log#Documentation/git-log.txt---parents 

Git lingo for replacing the true parent

RussKie:
Right! Thank you.
Worth adding that link to the comments. So at least one can look it up.

gerhardol:
Just updating the comments, still plan to rewrite to use `--follow information for Browse as well to find if HEAD is added.
(Some of the changes will probably not be used.)

gerhardol:
Implemented "insert first" without using `--follow`. 
RevisionGraph was not so much more complex.

RussKie (commented):
LGTM

gerhardol:
Reverting one minor cleanup (that will be tweaked in an upcoming PR) and making one performance tweak (really to improve readability)

mstv:
Most likely a git issue:
I have filtered with `Diff contains (SLOW)`. But the git command has not returned for almost 6 hours - I guess due to a child process `astextplain`. `pdf2text.exe` seems to block in conjunction with a pipe to `"C:\Program Files\Git\usr\bin\sed.exe" s/(\^M$)|(^\^L)//`.
After killing `pdf2text.exe`, another one is started for the next PDF.
These PDFs are stored in Git LFS. But the temp files have not been expanded.

This even blocks GE from switching repos.

grafik

- Git Extensions 33.33.33
- Build 825d3803734c4ef29c94eb32feb42614c3f4d2f3
- Git 2.30.1.windows.1
- .NET 5.0.5
- DPI 96dpi (no scaling)

gerhardol:
> I have filtered with `Diff contains (SLOW)`. But the git command has not returned for almost 6 hours - I guess due to a child process `astextplain`. `pdf2text.exe` seems to block in conjunction with a pipe to `"C:\Program Files\Git\usr\bin\sed.exe" s/(\^M$)|(^\^L)//`.

This should be a separate issue. Diff contains `git-log -G` is not working as expected, I do not get the results I expect and the graph is incorrect. (`--parents` are not included by default, can be forced by adding a path filter like `"**/*" "**/*"`).
There are pure GE issues too, like the left panel pumps out `git branch --format=%(refname) -a --merged HEAD` while `git-log -G` runs.

In your case I assume that you have changed .gitattributes for *.pdf. 
I assume `-G` generates the diff to find matches, this will be slow if searching all commits.
It must be be possible to abort these searches...
(A search for "hotkey" in the GE repo requires 24s for me which is reasonable, but that will require many minutes at work...)
https://stackoverflow.com/questions/28146244/what-is-the-difference-between-diff-and-diff-astextplain

Diff contains should maybe be included in Advanced filter, but I would like to finish the "FileHistory" first, RussKie is refactoring too.
But open issues as long as the actual PRs progresses...

mstv:
>  but I would like to finish the "FileHistory" first, RussKie is refactoring too.

I'll review now.

> But open issues as long as the actual PRs progresses...

It was quick a drop of the hang at work. I fear I must report a bug to the git team for the first time.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1004,14 +1005,12 @@ void OnRevisionRead(GitRevision revision)
                     this.InvokeAsync(() => { ShowLoading(false); }).FileAndForget();
                 }
 
-                var isCurrentCheckout = revision.ObjectId.Equals(CurrentCheckout);
+                bool isCurrentCheckout = revision.ObjectId.Equals(CurrentCheckout);
 
-                if (isCurrentCheckout &&
-                    ShowUncommittedChangesIfPossible &&
-                    AppSettings.RevisionGraphShowWorkingDirChanges &&
```
mstv, line 1011: These checks have been moved to the other local function.
(Removing this setting would be OK for me since simpler git GUIs have similar artificial commits by default. Just my 2ct.)

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1034,47 +1033,55 @@ void OnRevisionRead(GitRevision revision)
...
+                GitRevision indexRev = new(ObjectId.IndexId)
+                {
+                    Author = userName,
+                    AuthorUnixTime = 0,
+                    AuthorEmail = userEmail,
+                    Committer = userName,
+                    CommitUnixTime = 0,
+                    CommitterEmail = userEmail,
+                    Subject = ResourceManager.TranslatedStrings.Index,
+                    ParentIds = new[] { CurrentCheckout },
```
mstv, line 1078: What is the difference between `CurrentCheckout` and - the formerly used - `filteredCurrentCheckout`?

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -178,32 +178,58 @@ public void HighlightBranch(ObjectId id)
...
-                revisionGraphRevision = new RevisionGraphRevision(revision.ObjectId, ++_maxScore);
+
+                if (!insertFirst)
+                {
+                    revisionGraphRevision = new RevisionGraphRevision(revision.ObjectId, ++_maxScore);
+                }
+                else
+                {
+                    // Insert first artificial (WorkTree)
+                    // _maxScore do not need to be updated
```
mstv, line 201:
```suggestion
                    // _maxScore does not need to be updated
```
?

GitCommands/RevisionReader.cs
gerhardol: This piece was dropped in this PR after a rewrite, will appear in a follow up though

mstv (commented):
Just some questions and nits

GitUI/Translation/English.xlf
```diff
@@ -8733,6 +8733,14 @@ This will just checkout on the submodule the commit determined by the superproje
       </trans-unit>
     </body>
   </file>
+  <file datatype="plaintext" original="Plugin" source-language="en">
+    <body>
+      <trans-unit id="Description.Text">
+        <source>Plugin Manager</source>
```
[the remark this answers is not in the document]
gerhardol: Must be reverted

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
```diff
@@ -44,6 +44,14 @@ public void ApplyFlags(RevisionNodeFlags types)
         /// </summary>
         public int Score { get; private set; }
 
+        /// <summary>
+        /// Set the score to arbituary value allowing to insert a revision
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
```diff
@@ -44,6 +44,14 @@ public void ApplyFlags(RevisionNodeFlags types)
...
+        /// </summary>
+        public void InsertScore(int score)
```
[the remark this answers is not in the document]
gerhardol: This is an exception. An alternative is to skip setting the score here and set it when adding the segment to the parent below, like:

```
if (!insertFirst)
                {
parentRevisionGraphRevision = new RevisionGraphRevision(parentObjectId, ++_maxScore);
                }
                else
                {
parentRevisionGraphRevision = new RevisionGraphRevision(parentObjectId, -1);
                }

File: GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs [hunk 14]
```diff
@@ -299,9 +299,16 @@ private void OnCellPainting(object sender, DataGridViewCellPaintingEventArgs e)
...
+        /// Add a single revision from the git log to the graph, including segments to parents.
+        /// Update visible rows if needed.
+        /// </summary>
+        /// <param name="revision">The revision to add.</param>
+        /// <param name="types">The graph node flags.</param>
+        /// <param name="insertFirst">Insert the (artificial) revision first in the graph.</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool insertFirst = false)
         {
-            _revisionGraph.Add(revision, types);
+            _revisionGraph.Add(revision, types, insertFirst);
```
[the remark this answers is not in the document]
gerhardol: Changed upwards

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Artificial commits does not make sense in all forms, so at least `ShowUncommittedChangesIfPossible` should stay
In general I am in favor of reducing GE configurations, but it seems like each addition get someone confused, and removing old configurations would be worse.
These should be advanced settings though.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: None, filteredCurrentCheckout was always set to CurrentCheckout.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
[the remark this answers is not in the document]
gerhardol: Changed the implementation to this
Also
```
_orderedUntilScore = int.MinValue;
```
and inerting as -1 instead of -2

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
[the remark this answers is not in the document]
gerhardol: removed this...

gerhardol:
I have seen some situations where the segments are incorrectly rendered directly after loading, the artificial commits have segments to other commits.
Scrolling up/down fixes this.
I would like to merge this still.

mstv:
> I have seen some situations where the segments are incorrectly rendered directly after loading, the artificial commits have segments to other commits.

Perhaps `RevisionGraph.GetCachedCount` must take also these two additional commits into account. Can be done in a follow-up.

mstv:
Nothing is displayed for artificial commits if showing the history of a folder.
(The RevisionGrid looses the focus when switching to an artificial commit.)
Should we suppress them for folder history?

gerhardol:
> 
> 
> > I have seen some situations where the segments are incorrectly rendered directly after loading, the artificial commits have segments to other commits.
> 
> Perhaps `RevisionGraph.GetCachedCount` must take also these two additional commits into account. Can be done in a follow-up.

Need a reproduce - I lost the setup I had.
This is for new functionality and a simple scroll up/down is enough.



> 
> 
> Nothing is displayed for artificial commits if showing the history of a folder.
> (The RevisionGrid looses the focus when switching to an artificial commit.)
> Should we suppress them for folder history?

Folder history is fine(filter in grid).
Folder blame is not possible, that is the same in FileHistory - #9392 

FileTree is not defined for artificial. Some way to view blame from artificial would be nice, mentioned in #9424 To be handled later

mstv:
Works for me.

The test hang is most likely be caused by the missing adaptation of `RevisionGraph.GetCachedCount`.
If debug-running the three testcases in `FormBrowse_LeftPanel_ReorderNodesTest`, the third hangs often.

Another candidate could be `GitCommandsTests.Submodules.SubmoduleStatusProviderTests.Submodule_status_changes_for...`: No execution time at AV tab "tests".

gerhardol:
I am perplexed as if I reverse the activation of artificial commits tests still fail. Also, these tests should not be activated.
The RevGrid background thread is not cancelled, but also if that is done explicitly at the Dispose some tests hangs.

> 
> 
> Works for me.

> The test hang is most likely be caused by the missing adaptation of `RevisionGraph.GetCachedCount`.
> If debug-running the three testcases in `FormBrowse_LeftPanel_ReorderNodesTest`, the third hangs often.

I do not think that  `RevisionGraph.GetCachedCount` need adaption, does not handle Score.

mstv:
> I do not think that RevisionGraph.GetCachedCount need adaption, does not handle Score.

Yes, most likely not directly there. But the already loaded rows may be drawn as they have become "ordered".
(When using a slow filter,) the RevisionGrid can be drawn row by row.
If the artificial commits are inserted later, even rows above might need to be redrawn.
Note that the graph is cached as a bitmap per row. It must be invalidated, too.

gerhardol:
Tests are fixed (primary reason was that artificial was inserted when there were no other results at all, which caused the tests to hang).

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -163,6 +163,11 @@ protected override void Dispose(bool disposing)
                 // Make sure to mark the background queue as complete before disposing the cancellation token sequence.
                 _backgroundQueue.Complete();
                 _backgroundCancellationSequence.Dispose();
+                if (!_backgroundCancellationToken.IsCancellationRequested)
+                {
+                    _backgroundCancellationSequence.CancelCurrent();
```
mstv, line 168: ❔ Is it a good idea / really necessary to call `CancelCurrent` after `Dispose`?

gerhardol: Not strictly necessary, the core problem for the tests were that artificial commits were added without any other commits caused `OnRevisionReadCompleted()` to take the quick path and never quit the loading phase, background thread would never finish.
So I believe it should be included.

mstv: IIUIC, not necessary (and not critical) since `Dispose` does its job (and `CancelCurrent` will not throw after `Dispose` but should not be called because it creates a new token):
```
        /// <summary>
        /// Cancels the current token and disposes this object.
        /// </summary>
        /// <remarks>
        /// Only the first call to this method per instance of this class has any effect.
        /// </remarks>
        public void Dispose()
        {
            var cancellationTokenSource = Interlocked.Exchange(ref _cancellationTokenSource, null);

            if (cancellationTokenSource is null)
            {
                return;
            }

            cancellationTokenSource.Cancel();
            cancellationTokenSource.Dispose();
        }
```

gerhardol: When I debugged, the cancellationToken in RunBackgroundAsync() was never cancelled. So something is not working here.
But the core problem was that the artificial commits were added starting the background queue without the queue being stopped. That is a little error prone. But #8809 is submitted to fix this - should be completed.

So I will revert this.

mstv (approved):

gerhardol:
Plan to squash merge tomorrow

The issue I saw with the grid not refreshed after inserting artificial (requiring a scroll) could not be reproduced after a few minor changes (forcing the redraw), when earlier builds recreated them.
If this is enough or mot, I cannot say. But it is not a major issue.

gerhardol:
> RevGrid is scrolled to the topmost position. The artificial commit `Working directory` is missing:

WorkDir is there, just not shown. Ctrl-N is used to select first commit in list here.
See the missing scrollbar too.

This occurs when artificial is added first, also when not "inserted" as added in this PR.
It does not occur in Browse (used by default for fileHistory in #9445.
No fix yet.

image

> NRE in FormFileHistory:

#9483

PR #9397, opened 2021-07-20
Fix #9239 Drag Items from Selected Commit

### LINKED ISSUE #9239: Drag and drop from File tree copies the work tree version rather than the selected commit's
Draging a file from the File tree tab does not copy the version of the file belonging to the selected commit.

## Current behaviour

When dragging a file from the File tree tab the contents of the file is copied from the working tree rather than the selected commit.

## Expected behaviour

The version of the file belonging to the selected commit is copied.

## Steps to reproduce

* Check out a commit.
* Select another commit in the main GUI.
* Open the File tree tab.
* Identify a file that has been changed between the selected commit and the currently checked out commit.
* Drag a file from the File tree to the desktop.
* Compare the contents of the file on the desktop with the working tree and the selected commit.

## Screenshots

-

Not sure

## Environment

- Git Extensions 3.4.3.9999
- Git 2.29.2.windows.2
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

## Diagnostics

-

ransontech:
Fixes #9239

## Proposed changes
When dragging from File Tree provide a path to a temporary file structure that represents the selected commit instead of the checked-out commit.

## Test methodology 

- Find examples of the three GitObjectTypes.
- Drag those objects from the File Tree to a cmd prompt and File Explorer window.
- Cmd prompt gets path to the temporary stucture.
- File Explorer copies the temporary structure to the target directory.
- Confirmed that the archived structure represents the selected commit, not the checked-out commit.

## Test environment(s) 

- GIT 2.26.2.windows.1
- Windows 10.0.19042

----


GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -45,6 +45,9 @@ public partial class RevisionFileTreeControl : GitModuleControl
 
 See the changes in the commit form.");
 
+        private readonly TranslationString _cantSaveToTemp = new("Unable to save to Temp directory.");
+        private readonly TranslationString _cantExtractTempArchive = new("Unable to extract tempory archive.");
```
pmiossec, line 49: I think you mean `temporary`?
Also commit the translation file if you want to fix the build.
See https://github.com/gitextensions/gitextensions/wiki/Translations#net-50-v36

pmiossec:
It's better if the change in the `contributors.txt` is made in another PR.
Do you think you could do that?

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
ransontech: Thanks for your comments, have fixed the typo and updated the translation file.

ransontech:
> It's better if the change in the `contributors.txt` is made in another PR.
> Do you think you could do that?

No problem at all, will raise another PR for that.

GitCommands/Git/GitModule.cs
```diff
@@ -595,6 +596,30 @@ public void SaveBlobAs(string saveAs, string blob)
...
+            {
+                GitArgumentBuilder args = new("archive")
+                {
+                    $"--format zip",
+                    $"--output \"{archivePath.ToNativePath()}\"",
+                    $"{hash}",
+                    "-0"
+                };
+
+                using var process = _gitCommandRunner.RunDetached(args);
```
mstv, line 613: I think nothing from `GitCommandRunner` is used:
```suggestion
                using IProcess process = _gitExecutable.Start(args);
```

GitCommands/Git/GitModule.cs
```diff
@@ -595,6 +596,30 @@ public void SaveBlobAs(string saveAs, string blob)
             stream.Write(blobData, 0, blobData.Length);
         }
 
+        public void SaveTreeAs(string saveAs, string hash)
+        {
+            string archivePath = Path.Join(saveAs.SubstringUntilLast('\\'), $"{hash}.zip");
```
mstv, line 601: Use a portable function for the extraction of the path part, please.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -339,6 +342,31 @@ private void OnItemActivated()
...
         }
 
+        private string? SaveItemToTempFile(GitItem gitItem)
+        {
+            if (string.IsNullOrWhiteSpace(gitItem.FileName))
+            {
+                return null;
+            }
+
+            string fileName = gitItem.FileName.SubstringAfterLast('/').SubstringAfterLast('\\');
```
mstv, line 352: ditto: use existing function

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -403,15 +431,32 @@ private void tvGitTree_ItemDrag(object sender, ItemDragEventArgs e)
...
+                catch (UnauthorizedAccessException)
+                {
+                    MessageBox.Show(_cantSaveToTemp.Text, TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                    return;
+                }
+                catch (FileNotFoundException)
+                {
+                    MessageBox.Show(_cantExtractTempArchive.Text, TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                    return;
+                }
```
mstv, line 448: There are more types of possible exceptions. And this mapping to messages might match often but not always.
Usually, the exception text is more informative than these slightly generic messages.
You could even let the exception simply bubble up to the `BugReportInvoker`, i.e. without catching.
In order to suppress bug reports, the `Exception` should be wrapped into a `UserExternalOperationException`.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -403,15 +431,32 @@ private void tvGitTree_ItemDrag(object sender, ItemDragEventArgs e)
...
 
-                fileList.Add(fileName.ToNativePath());
+                if (tempFilePath is not null)
```
mstv, line 450: 🤔 `tempFilePath` is not declared as `string?`.
If it really can be `null`, the old variant could be kept as fallback.

mstv (commented):
Elegant solution.
Just requesting little fine-tuning.
Have not run yet.

mstv:
To be discussed:
This could leave lots of files in `%TEMP%` - also if an entry is dragged by mistake instead of clicking.

GitCommands/Git/GitModule.cs
```diff
@@ -595,6 +596,30 @@ public void SaveBlobAs(string saveAs, string blob)
             stream.Write(blobData, 0, blobData.Length);
         }
 
+        public void SaveTreeAs(string saveAs, string hash)
```
RussKie, line 599: This functionality doesn't belong to `GitModule`, it is specific `RevisionFileTreeControl`.

GitCommands/Git/GitModule.cs
```diff
@@ -595,6 +596,30 @@ public void SaveBlobAs(string saveAs, string blob)
...
+
+            if (!File.Exists(archivePath))
+            {
+                GitArgumentBuilder args = new("archive")
+                {
+                    $"--format zip",
+                    $"--output \"{archivePath.ToNativePath()}\"",
+                    $"{hash}",
+                    "-0"
+                };
```
RussKie, line 611: Turn either into a `GitCommand` or move to `GitCommandHelpers`.

GitCommands/Git/GitModule.cs
RussKie, line 599: The whole operation must be run via the `FormProcess`

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
RussKie: Any issues would very likely be user-level failures (i.e. some the user will be in a position to resolve), and must be reported via the BugReporter

RussKie (changes requested):

GitCommands/Git/GitModule.cs
ransontech: Thanks for the comments, I'm struggling a bit with this one. Running via FormProcess results in the dialog stealing focus, which interrupts the drag opperation, I'm not sure how to get around that?

GitCommands/Git/GitModule.cs
ransontech: Thanks, using new method "GetArchivePath" in PathUtils

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
ransontech: Using existing GetFileName of PathUtils

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
ransontech: Thanks both, I can see your point, am not catching these locally now

GitCommands/Git/GitModule.cs
RussKie: 🤦‍♂️ of course! Silly me...

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
ransontech: Sorry, should be `string?`, as it could be null if the GitItem.FileName was empty, but no need for me to check for it as the drag operation is ok to continue with null.

ransontech:
> To be discussed:
> This could leave lots of files in `%TEMP%` - also if an entry is dragged by mistake instead of clicking.

Very good point, I now delete the .zip when it has been extracted, and the extracted file/s once they have been coppied by the drag operation. So should not be anything left as long the execution is not interrupted.

GitCommands/Git/GitModule.cs
ransontech: Have moved to GitCommandHelpers

GitCommands/Git/GitModule.cs
ransontech: Have moved to RevisionFileTreeControl

RussKie (commented):
Thinking more about the proposed implementation I get concerned that we may be approaching it the wrong way. 

For large commits this may turn out to be a very long operation. Blocking the app and likely affecting the OS as well... I don't think `ItemDrag` is the correct event to extract data from git, zip/unzip, copy, delete.

I would imaging the flow to be something like this:
- start to drag, update the mouse cursor, etc.
- finish the drag, perform the drop
- spawn the progress dialog and perform extract/zip/unzip/etc.

This would be consistent with Windows file copy behaviour - where you indicate what files and where need to be copied, then Windows shows the "File Copy" dialog while performing the requested operations.

GitCommands/PathUtil.cs
```diff
@@ -94,6 +95,30 @@ public static string GetFileName(string fileName)
...
+            var pos = fileName.LastIndexOfAny(pathSeparators);
+            if (pos != -1)
+            {
+                fileName = fileName.Substring(0, pos + 1);
+            }
+
+            return fileName;
+        }
+
+        public static string GetTempPath(string fileName)
```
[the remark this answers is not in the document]
ransontech: Agreed, have changed this.

GitCommands/PathUtil.cs
```diff
@@ -94,6 +95,30 @@ public static string GetFileName(string fileName)
             return fileName;
         }
 
+        public static string GetParentDirectoryPath(string fileName)
```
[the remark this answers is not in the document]
ransontech: Good point, done.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -403,15 +429,26 @@ private void tvGitTree_ItemDrag(object sender, ItemDragEventArgs e)
...
+                FileAttributes attr = File.GetAttributes(tempFilePath);
+
+                if (attr.HasFlag(FileAttributes.Directory))
+                {
+                    Directory.Delete(tempFilePath, recursive: true);
+                }
+                else
+                {
+                    File.Delete(tempFilePath);
+                }
```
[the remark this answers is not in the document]
ransontech: DoDragDrop is a blocks until the drag drop operation has completed, so I believe they should be unlocked before the execution continues to the deletion

RussKie:
@ransontech is this something you're still planning to pursue?

PR #9413, opened 2021-07-25
Rename detection for Advanced filter

gerhardol:
See long term goals in #7598

## Proposed changes

- Refactor FileChanges -> RevisionGrid (the normal name for the control in other forms)

- Do not try to `--follow` for folders. Something like `Bin/` is OK, but `GitUI` will give too many files and too long command lines. (This could be solved in other ways but do not bother.)
Note that 3.5 does not follow folders (or Submodules) so not limiting functionality. History for folders is useless in 3.5 as well.
Note: There is no enforcing of using the ending "/" in the Advanced filter, manual input.

- Move detection of renamed files from FileHistory to RevisionGrid, to use the handling also for Advanced filters (with slightly more general rules).

## Test methodology 

Add an advanced filter like "\*\*/Rev\*Diff\*" to see that GitUI\CommandsDialogs\RevisionDiffControl.cs was renamed

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FormRevisionFilter.cs
+++ b/GitUI/UserControls/RevisionGrid/FormRevisionFilter.cs
@@ -116,7 +116,7 @@ namespace GitUI.UserControls.RevisionGrid
 
         public string GetPathFilter()
         {
-            return FileFilterCheck.Checked ? $" {FileFilter.Text.ToPosixPath()}" : "";
+            return FileFilterCheck.Checked ? FileFilter.Text : "";
         }
 
         public string GetInMemAuthorFilter()
@@ -147,6 +147,12 @@ namespace GitUI.UserControls.RevisionGrid
         public void SetBranchFilter(string filter)
         {
             BranchFilter.Text = filter?.Trim();
+        }
+
+        public void SetPathFilter(string filter)
+        {
+            FileFilter.Text = filter?.Trim();
+            FileFilterCheck.Checked = !string.IsNullOrWhiteSpace(filter);
         }
 
         private void OkClick(object sender, EventArgs e)
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -144,8 +144,6 @@ namespace GitUI
         private string? _rebaseOnTopOf;
         private bool _isRefreshingRevisions;
         private IReadOnlyList<ObjectId>? _selectedObjectIds;
-        private string? _fixedRevisionFilter = "";
-        private string _fixedPathFilter = "";
         private string _branchFilter = "";
         private SuperProjectInfo? _superprojectCurrentCheckout;
         private int _latestSelectedRowIndex;
@@ -176,6 +174,13 @@ namespace GitUI
         internal bool IsShowCurrentBranchOnlyChecked { get; private set; }
         internal bool IsShowAllBranchesChecked { get; private set; }
         internal bool IsShowFilteredBranchesChecked { get; private set; }
+
+        /// <summary>
+        /// The (first) seen name for commits, for FileHistory with path filters.
+        /// See BuildFilter() for limitations of commits included.
+        /// The property is explicitly initialized by FileHistory.
+        /// </summary>
+        internal Dictionary<ObjectId, string>? FilePathByObjectId { get; set; } = null;
 
         public RevisionGridControl()
         {
@@ -386,10 +391,26 @@ namespace GitUI
             }
         }
 
-        public void SetFilters((string? revision, string path) filter)
+        // returns " --find-renames=..." according to app settings
+        private static ArgumentString FindRenamesOpt()
         {
-            _fixedRevisionFilter = filter.revision;
-            _fixedPathFilter = filter.path;
+            return AppSettings.FollowRenamesInFileHistoryExactOnly
+                ? " --find-renames=\"100%\""
+                : " --find-renames";
+        }
+
+        // returns " --find-renames=... --find-copies=..." according to app settings
+        private static ArgumentString FindRenamesAndCopiesOpts()
+        {
+            var findCopies = AppSettings.FollowRenamesInFileHistoryExactOnly
+                ? " --find-copies=\"100%\""
+                : " --find-copies";
+            return FindRenamesOpt() + findCopies;
+        }
+
+        public void SetPathFilters(string path)
+        {
+            _revisionFilter.SetPathFilter(path);
         }
 
         private void InitiateRefAction(IReadOnlyList<IGitRef>? refs, Action<IGitRef> action, FormQuickGitRefSelector.Action actionLabel)
@@ -952,6 +973,7 @@ namespace GitUI
                 _gridView.SelectionChanged += OnGridViewSelectionChanged;
                 _gridView.ResumeLayout();
 
+                string pathFilter = BuildPathFilter(_revisionFilter.GetPathFilter());
                 _revisionReader.Execute(
                     Module,
                     refs,
@@ -959,8 +981,8 @@ namespace GitUI
                     _revisionFilter.GetMaxCount(),
                     refFilterOptions,
                     _branchFilter,
-                    _revisionFilter.GetRevisionFilter() + QuickRevisionFilter + _fixedRevisionFilter,
-                    _revisionFilter.GetPathFilter() + _fixedPathFilter,
+                    _revisionFilter.GetRevisionFilter() + QuickRevisionFilter,
+                    pathFilter,
                     predicate);
 
                 if (_initialLoad)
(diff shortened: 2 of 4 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -592,6 +606,17 @@ public new void Load()
             LoadCustomDifftools();
         }
 
+        public new void Load(ref Dictionary<ObjectId, string> filePathByObjectId)
```
RussKie, line 609: `Dictionary` is a class, why is it passed by `ref`?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -814,6 +839,12 @@ private void ShowLoading(bool sync = true)
         }
 
         public void ForceRefreshRevisions()
+        {
+            Dictionary<ObjectId, string> filePathByObject = null;
+            ForceRefreshRevisions(ref filePathByObject);
```
RussKie, line 844: This doesn't make sense to me

gerhardol: `BuildFilter()` is needed for path arguments both from FileHistory and  Advanced filter (this PR removes the separate path filter handling too).
The routine finds the previous filenames and the name(s) in previous commits, mostly used for presentation. The latter part is only needed in FileHistory.
With FileHistory it is possible to "insert" `BuildFilter()` before RevisionGridControl handling, not possible with changes in Advanced filter.
So `filePathByObject` is only needed by FileHistory.

Note that `BuildFilter()` also handles quoting of path arguments to handle user input in Advanced filter better.

There are alternatives to produce `filePathByObject`, some examples:
* Set `filePathByObject` to null when not needed (other than FileHistory requesting) and update the variable if not null. - What is in the PR now, in my view the simplest with the least amount of code.
* Always update the variable, referred by a ref. (in the snippet above, null would be changed to new()) - Similar, just requires a few more CPU cycles.
* Always update the variable, let FileHistory ask for it afterwards. - Requires a few more CPU cycles and yet another property but maybe the overloaded `ForceRefreshRevisions()` is not needed.
* Make producing `filePathByObject` a RevisionGrid configurable, let FileHistory ask for it afterwards. - Similar to above.
* Move the handling of `filePathByObject` to FileHistory, requiring a third git-log invocation. - Straightforward but can add seconds to the FileHistory opening time.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: see below

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: An alternative implementation in e58fa066b09f443de7826f675c2b847fb05f7aff, is that OK?
(similar to alternative 4 above)

RussKie: > An alternative implementation in [e58fa06](https://github.com/gitextensions/gitextensions/commit/e58fa066b09f443de7826f675c2b847fb05f7aff), is that OK?

This looks better 👍

gerhardol: Why set this as draft?

RussKie (approved):
Let's get it in before I get to the history dialog in my clean up.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -986,14 +1008,117 @@ public void ForceRefreshRevisions()
...
+                return;
             }
             catch
             {
                 SetPage(new ErrorControl());
                 throw;
             }
 
-            return;
+            string BuildPathFilter(string? path)
```
RussKie, line 1020: ❗  We really need to test it.

gerhardol: Can we take this later?
There are several PRs that depends on this PR.

The quoting handling is making some guessing, but as the handling ultimately uses Git to ignore bad paths, it should not be critical.
The quoting and --follow handling can be separated.

...and I have raised my opinions about the usefulness about parsing internal methods like this before...

gerhardol:
Suggest squash merging in a day or so.

GitUI/UserControls/RevisionGrid/FormRevisionFilter.cs
```diff
@@ -149,6 +149,12 @@ public void SetBranchFilter(string filter)
             BranchFilter.Text = filter?.Trim();
         }
 
+        public void SetPathFilter(string filter)
+        {
+            FileFilterCheck.Checked = !string.IsNullOrWhiteSpace(filter);
+            FileFilter.Text = filter?.Trim();
```
mstv, line 155: Have you considered filenames with spaces at start/end?

mstv:
A separate commit with the renaming would ease the review a little.

GitUI/UserControls/RevisionGrid/FormRevisionFilter.cs
gerhardol: GE added filenames (for FileHistory and a later PR for a context menu for RevDiff etc) are quouted when adding.
User added paths in Advanced filter must be correctly quoted by the user.
(There is a helper for quoting paths without spaces and quotes, likely the most common case).

> A separate commit with the renaming would ease the review a little.

This was separated in several commits until I squashed it preparing for merge.
But commits are still separate in branch gerhardol/feature/i7598-replace-formfilehistory

mstv (commented):
Almost OK for me.

---

> This was separated in several commits until I squashed it preparing for merge.

My 2ct: I would keep such plain refactoring commits separate.
Since my available time is getting less, I have started my review after the requested changes were done.
(Not requesting changes neither for the edge case with spaces nor for separate commits.)

> But commits are still separate in branch gerhardol/feature/i7598-replace-formfilehistory

This branch does not match. I have split it myself for reviewing.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -986,14 +1008,117 @@ public void ForceRefreshRevisions()
...
+                    $"--format=\"{startOfObjectId}%H\"",
+                    "--name-only",
+                    "--follow",
+                    FindRenamesAndCopiesOpts(),
+                    "--",
+                    path
+                };
+
+                HashSet<string?> setOfFileNames = new();
+                var lines = Module.GitExecutable.GetOutputLines(args, outputEncoding: GitModule.LosslessEncoding/*, throwOnErrorOutput: false*/);
```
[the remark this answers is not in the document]
gerhardol: `git-log --follow` only handle one path argument.
It is not trivial to detect how many arguments the user supplies in the advanced filter (only guessing), so to not block valid user inputs, the Git command must not throw.
If there are error handling added, it must react on the error code and ask if the user wants to continue or not.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -986,14 +1008,117 @@ public void ForceRefreshRevisions()
...
+
+                return;
             }
             catch
             {
                 SetPage(new ErrorControl());
                 throw;
             }
 
-            return;
```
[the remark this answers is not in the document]
gerhardol: Can revert, this was required before a rewrite (after RussKie comments).

gerhardol:
> My 2ct: I would keep such plain refactoring commits separate.

Separated to a refactor commit.
Git supports ignoring some changes in some commits in blame:
https://akrabat.com/ignoring-revisions-with-git-blame/
With no `--ff-only` in GH this requires a separate PR which makes this very cumbersome.
In this case I did not think it was necessary, so the commits was marked with `squash!` in the original commits.

Squashed the two minor changes and modified the HEAD commit message, so RussKie can rebase merge this PR to get going with his changes.

mstv (approved):

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -177,6 +175,13 @@ private RefFilterOptions RefFilterOptions
         internal bool IsShowAllBranchesChecked { get; private set; }
         internal bool IsShowFilteredBranchesChecked { get; private set; }
 
+        /// <summary>
+        /// The (first) seen name for commits, for FileHistory with path filters.
+        /// See BuildFilter() for limitations of commits included.
+        /// The property is explicitly initialized by FileHistory.
+        /// </summary>
+        internal Dictionary<ObjectId, string> FilePathByObjectId { get; set; } = null;
```
RussKie, line 183:
```suggestion
        internal Dictionary<ObjectId, string>? FilePathByObjectId { get; set; } = null;
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: > Can we take this later?

Sure



> ..and I have raised my opinions about the usefulness about parsing internal methods like this before...

If we don't have tests, there's a high change we'll either keep churning here trying to fix unintended regressions, or regress while updating the functionality.

RussKie (approved):

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: changed

gerhardol:
Rebased on master, squashing https://github.com/gitextensions/gitextensions/pull/9413#pullrequestreview-719557101 and activate the comment discussed in https://github.com/gitextensions/gitextensions/pull/9413#discussion_r680241395 due to #9056.

gerhardol:
`Rebase and merge` in 10 hours (manual, no msftbot?)

RussKie:
👍

gerhardol:
(switched order Text/Checked in FormFileHistory.cs as setting Checked triggers `EnableFilters()` - no difference right now though)

PR #9416, opened 2021-07-26
Support `git-bash.exe` again

### LINKED ISSUE #9353: Git executable not found in Console

mstv:
Fixes #9353
not completely - but the regression from #8213
`git-bash.exe` cannot be used for the built-in console.
As I explained in https://github.com/gitextensions/gitextensions/issues/9353#issuecomment-881975596, `bash.exe` ignores the path to `git.exe` added by GE to `PATH`.

## Proposed changes

- `BashShell`: Try `git-bash.exe` first for stand-alone shell
- search also in the 64-bit-programs folder

### Before

image

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build d4b3e13e80fd0f094fecd9f153f0fe135c4d4d09 (Dirty)
- Git 2.31.1.windows.1
- .NET Framework 4.8.4360.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitUI/Shells/BashShell.cs
+++ b/GitUI/Shells/BashShell.cs
@@ -9,7 +9,8 @@ namespace GitUI.Shells
 {
     public class BashShell : ShellDescriptor
     {
-        private const string BashExe = "bash.exe"; // Generic bash, should generally be in the git dir
+        private const string GitBashExe = "git-bash.exe"; // Bash with git in the path, should generally be in the git dir
+        private const string BashExe = "bash.exe"; // Fallback to generic bash, should generally be in the git bin dir
         private const string ShExe = "sh.exe";     // Fallback to SH
         public const string ShellName = "bash";
 
@@ -18,16 +19,32 @@ namespace GitUI.Shells
             Name = ShellName;
             Icon = Images.GitForWindows;
 
-            ExecutableName = BashExe;
-            if (PathUtil.TryFindShellPath(ExecutableName, out var exePath))
+            if (PathUtil.TryFindShellPath(GitBashExe, out var exePath))
             {
+                ExecutableName = GitBashExe;
                 ExecutablePath = exePath;
+
+                // Try to find bash or sh below to set ExecutableCommandLine, as git-bash.exe cannot be connected to the built-in console.
+            }
+
+            if (PathUtil.TryFindShellPath(BashExe, out exePath))
+            {
+                if (ExecutablePath is null)
+                {
+                    ExecutableName = BashExe;
+                    ExecutablePath = exePath;
+                }
+
                 ExecutableCommandLine = $"{exePath.Quote()} --login -i";
             }
             else if (PathUtil.TryFindShellPath(ShExe, out exePath))
             {
-                ExecutableName = ShExe;
-                ExecutablePath = exePath;
+                if (ExecutablePath is null)
+                {
+                    ExecutableName = ShExe;
+                    ExecutablePath = exePath;
+                }
+
                 ExecutableCommandLine = $"{exePath.Quote()} --login -i";
             }
         }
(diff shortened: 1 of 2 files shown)


GitUI/Shells/BashShell.cs
```diff
@@ -18,16 +19,31 @@ public BashShell()
...
-            ExecutableName = BashExe;
-            if (PathUtil.TryFindShellPath(ExecutableName, out var exePath))
+            if (PathUtil.TryFindShellPath(GitBashExe, out var exePath))
             {
+                ExecutableName = GitBashExe;
                 ExecutablePath = exePath;
+                //// do not set ExecutableCommandLine: git-bash.exe cannot be connected to built-in console
+            }
+
+            if (PathUtil.TryFindShellPath(BashExe, out exePath))
```
gerhardol, line 29: ?
```suggestion
            else if (PathUtil.TryFindShellPath(BashExe, out exePath))
```

gerhardol (commented):
OK except for the comment
Have not run

GitUI/Shells/BashShell.cs
mstv: No `else if` on purpose.

Should I reformulate the above comment to
```
//// Do not set ExecutableCommandLine to git-bash.exe since it cannot be connected to the built-in console.
//// For the console, try to find bash or sh below.
```
?

gerhardol: Do not use ////, should only be used for TODO and other that should be changed, add the extra newline

Maybe?
```c#

// Try to find bash or sh below to set ExecutableCommandLine, as git-bash.exe cannot be connected to the built-in console.
```

mstv: Thank you, I took this.

gerhardol (approved):

GitUI/Shells/BashShell.cs
```diff
@@ -18,16 +19,32 @@ public BashShell()
...
             }
             else if (PathUtil.TryFindShellPath(ShExe, out exePath))
             {
-                ExecutableName = ShExe;
-                ExecutablePath = exePath;
+                if (ExecutablePath is null)
+                {
+                    ExecutableName = BashExe;
+                    ExecutablePath = exePath;
+                }
```
RussKie, line 46: Line:30 - look for `BashExe`, line:34 - set to `BashExe`. Line:40 - look for `ShExe`, line:34 - set to `BashExe`? This at least deserves a comment explaining why this is.

Also lines:31-39 and lines:41:49 are identical, and **if** the above is correct, then, I think, this should work:
```suggestion
            if (PathUtil.TryFindShellPath(BashExe, out exePath) || PathUtil.TryFindShellPath(ShExe, out exePath))
            {
                if (ExecutablePath is null)
                {
                    ExecutableName = BashExe;
                    ExecutablePath = exePath;
                }

                ExecutableCommandLine = $"{exePath.Quote()} --login -i";
            }
```

RussKie (commented):
Tests?

GitUI/Shells/BashShell.cs
mstv: Good catch. It was a copy & paste mistake. (The code before #8213 did not have this duplicated code. But it's only 2.5 copies...)

RussKie (approved):

PR #9420, opened 2021-07-28
Added option in the GitFlow plugin to use the Git Flow commands for all submodules as well

TT-Mark-van-Doorn:
## Proposed changes

- When using the GitFlow plugin there now is an option to use the GitFlow options for all submodules as well.
This could be helpfull on a project with a lot of submodules so you don't have to create a feature branch for everysubmodule.

### Before

image

### After

image

## Test methodology 

- Create a repository with multiple submodules
- Use the GitFlow plugin with the new "Include submodules" checkbox enabled to init Git Flow for all submodules
- Make random changes to the main repo and submodules, commit these.
- Use the Git Flow plugin again to Publish, Pull and Finish all submodules and the main repo

## Test environment(s) 

- Git Extensions 33.33.33
- Build 9b2011f707ca142f8ff0609489bf7f0a5a823d2a
- Git 2.32.0.windows.1
- .NET Framework 4.8.4390.0
- DPI 96dpi (no scaling)

----

### DIFF
--- a/Plugins/GitFlow/GitFlowForm.Designer.cs
+++ b/Plugins/GitFlow/GitFlowForm.Designer.cs
@@ -1,6 +1,4 @@
-﻿using System;
-
-namespace GitFlow
+﻿namespace GitFlow
 {
     partial class GitFlowForm
     {
@@ -118,7 +116,7 @@ namespace GitFlow
             // 
             // btnFinish
             // 
-            this.btnFinish.Location = new System.Drawing.Point(490, 72);
+            this.btnFinish.Location = new System.Drawing.Point(490, 65);
             this.btnFinish.Name = "btnFinish";
             this.btnFinish.Size = new System.Drawing.Size(82, 23);
             this.btnFinish.TabIndex = 0;
@@ -130,7 +128,7 @@ namespace GitFlow
             // 
             this.gbManage.Controls.Add(this.pnlManageBranch);
             this.gbManage.Controls.Add(this.cbManageType);
-            this.gbManage.Location = new System.Drawing.Point(13, 149);
+            this.gbManage.Location = new System.Drawing.Point(13, 169);
             this.gbManage.Name = "gbManage";
             this.gbManage.Size = new System.Drawing.Size(628, 162);
             this.gbManage.TabIndex = 6;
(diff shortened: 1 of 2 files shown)


Plugins/GitFlow/GitFlowForm.cs
```diff
@@ -337,9 +351,37 @@ private bool RunCommand(ArgumentString commandText)
 
             txtResult.Text = resultText;
 
+            ApplyCommandToAllSubmodules(commandText);
+
             return result.ExitCode == 0;
         }
 
+        private void RunOutputCommand(ArgumentString commandText)
```
gerhardol, line 359: This is mostly a duplication of RunCommand()
Can this be merged?

pmiossec:
if I see a checkbox named like that "Include submodules", I don't really understand what that means.
Why don't you use the same label "Apply the command to all submodules" than the method you created?

I find it much clearer...

And maybe put the checkbox on its own line because the name of the current  checked out branch could be long and overlap the checkbox

Plugins/GitFlow/GitFlowForm.Designer.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace GitFlow
+﻿using System;
```
gerhardol, line 1: should not be needed

Plugins/GitFlow/GitFlowForm.Designer.cs
```diff
@@ -459,11 +462,22 @@ private void InitializeComponent()
             this.lblRunCommand.TabIndex = 1;
             this.lblRunCommand.Text = "-";
             // 
+            // cbApplyToSubmodules
+            // 
+            this.cbApplyToSubmodules.AutoSize = true;
```
gerhardol, line 467: Should this be in the Start branch group, would it not make sense to have this in next group?

gerhardol (commented):
Can you get another GitFlow user to review this change too, there are no GitFlow users that I know in the core team?

Plugins/GitFlow/GitFlowForm.Designer.cs
TT-Mark-van-Doorn: I don't think it should be inside the Start branch group since the checkbox applies for ALL commands. So it has to be above all steps to follow in my opinion.

Plugins/GitFlow/GitFlowForm.Designer.cs
TT-Mark-van-Doorn: resolved

Plugins/GitFlow/GitFlowForm.cs
TT-Mark-van-Doorn: I don't recommend merging it since the method does seperate things. The RunCommand method doesn't have a result code for example. If duplicate code must be avoided then I suggest putting all the text and drawing actions to a seperate method.

pmiossec:
I'm so sorry @TT-Mark-van-Doorn but thinking about it again, don't you think that the label "**Also**  apply the command to all submodules"  is a little better?

PR #9106, opened 2021-04-24
Restore splitter persistence

RussKie:
Resolves #8745
Resolves #9000
Resolves #9089

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9424, opened 2021-07-28
RevFileTree: Blame instead of View

### LINKED ISSUE #9435: BlameControl: Hide unrelated context menu item
## Current behaviour

image

## Expected behaviour

`Enable automatic continuous scroll` applies only to tab "Diff" of FormBrowse

## Environment

- Git Extensions 33.33.33
- Build 77310a56f66d30afc29717d4060c2bca146d4b0d
- Git 2.31.1.windows.1
- .NET 5.0.7
- DPI 96dpi (no scaling)

### LINKED ISSUE #9437: Improve FileTree / BlameControl updates
## Current behaviour

With #9424, the BlameControl is used for viewing files on tab "File tree" of FormBrowse.
When switching files, the FileTree is invalidated.
After reloading the FileTree, the selected file may not be selected - although it exists in this tree. A folder is selected instead.

## Expected behaviour

When switching files, the FileTree does not need to be invalidated.
After reloading the FileTree when switching commits, scroll the FileTree selection to be visible.
The file selection shall be restored after reloading the FileTree - also following renames if possible.

## Steps to reproduce

- view tab "File tree"
- switch files
- switch commits
- `Blame previous revision` multiple times

## Screenshots

FileTreeBlame

FolderSelected

new feature

## Environment

#9424

gerhardol:
Part of #7598
Closes #9435
Closes #9437 

## Proposed changes

- RevFileTree: Blame instead of View
Show blame for object type blobs (files), as is available in FileHistory.
Note that object type commit (submodules) are shown as text, as before.
It is possible to toggle the view to show file contents, as before.

In FileHistory the commit info is available in a header. This is excluded in RevFileTree. The info for the selected commit can be seen in Commit, Diff tabs, overview is available in tooltip for otyher commits.
This could be configurable.

- RevDiff: HotKey to view FileTree
A menu command existed to view the file, this as a hotkey.

Note that `T` was selected as key, `F` is to be used to add the filepath to the Advanced filter in the grid.

- RevDiff: `B` to view Blame instead of Diff
Note that this is just for the current file, selecting another (or the same) file will show Diff by default.
For artificial, HEAD is blamed, worktree/index changes cannot be shown

- When blaming in RevDiff or FileTree, select the current line in fileviewer in Blame (especially useful in RevDiff)

- Cancellation support in LoadBlame() (in between operations, not in methods)

### Before

image

For reference - FileHistory Blame tab, unchanged by this PR
image

(RevDiff `B` opened FileHistory in a new instance, no screenshot.)

### After

image

Only the hotkey added
image

Context menu and hotkey `B` toggle Diff/Blame.

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -319,8 +319,8 @@ namespace GitUI.CommandsDialogs
             _aheadBehindDataProvider = GitVersion.Current.SupportAheadBehindData ? new AheadBehindDataProvider(() => Module.GitExecutable) : null;
 
             repoObjectsTree.Initialize(_aheadBehindDataProvider, branchFilterAction: ToolStripFilters.SetBranchFilter, RevisionGrid, RevisionGrid, RevisionGrid);
-            revisionDiff.Bind(RevisionGrid, fileTree, () => RequestRefresh());
-            fileTree.Bind(() => RequestRefresh());
+            revisionDiff.Bind(RevisionGrid, fileTree, RequestRefresh);
+            fileTree.Bind(RevisionGrid, RequestRefresh);
 
             UpdateCommitButtonAndGetBrush(null, AppSettings.ShowGitStatusInBrowseToolbar);
             RestorePosition();
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -462,12 +462,14 @@ namespace GitUI.CommandsDialogs
 
         private void BlameSelectedFileDiff()
         {
+            int? line = DiffText.Visible ? DiffText.CurrentFileLine : null;
+
             BlameControl.Visible = true;
             DiffText.Visible = false;
             GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
                 ? _revisionGrid.GetActualRevision(_revisionGrid.CurrentCheckout)
                 : DiffFiles.SelectedItem.SecondRevision;
-            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding);
+            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding, line);
         }
 
         private async Task ShowSelectedFileDiffAsync()
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -62,6 +62,7 @@ namespace GitUI.CommandsDialogs
             DiffText.TopScrollReached += FileViewer_TopScrollReached;
             DiffText.BottomScrollReached += FileViewer_BottomScrollReached;
             DiffText.LinePatchingBlocksUntilReload = true;
+            BlameControl.HideCommitInfo();
         }
 
         private void FileViewer_TopScrollReached(object sender, EventArgs e)
@@ -230,7 +231,7 @@ namespace GitUI.CommandsDialogs
             }
         }
 
-        public void Bind(RevisionGridControl revisionGrid, RevisionFileTreeControl revisionFileTree, Action refreshGitStatus)
+        public void Bind(RevisionGridControl revisionGrid, RevisionFileTreeControl revisionFileTree, Action? refreshGitStatus)
         {
             _revisionGrid = revisionGrid;
             _revisionFileTree = revisionFileTree;
@@ -459,8 +460,20 @@ namespace GitUI.CommandsDialogs
             RequestRefresh();
         }
 
+        private void BlameSelectedFileDiff()
+        {
+            BlameControl.Visible = true;
+            DiffText.Visible = false;
+            GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
+                ? _revisionGrid.GetActualRevision(_revisionGrid.CurrentCheckout)
+                : DiffFiles.SelectedItem.SecondRevision;
+            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding);
+        }
+
         private async Task ShowSelectedFileDiffAsync()
         {
+            BlameControl.Visible = false;
+            DiffText.Visible = true;
             await DiffText.ViewChangesAsync(DiffFiles.SelectedItem,
                 openWithDiffTool: () => firstToSelectedToolStripMenuItem.PerformClick(),
                 cancellationToken: _viewChangesSequence.Next());
@@ -600,7 +613,8 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            UICommands.StartFileHistoryDialog(this, item.Item.Name, item.SecondRevision, true, true);
+            blameToolStripMenuItem.Checked = !blameToolStripMenuItem.Checked;
+            BlameSelectedFileDiff();
         }
 
         private void StageFileToolStripMenuItemClick(object sender, EventArgs e)
@@ -1141,7 +1155,14 @@ namespace GitUI.CommandsDialogs
         {
             if (alreadyContainedFocus && DiffFiles.Focused)
             {
-                DiffText.Focus();
+                if (BlameControl.Visible)
+                {
+                    BlameControl.Focus();
+                }
+                else
+                {
+                    DiffText.Focus();
+                }
             }
             else
             {
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -469,7 +469,7 @@ namespace GitUI.CommandsDialogs
             GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
                 ? _revisionGrid.GetActualRevision(_revisionGrid.CurrentCheckout)
                 : DiffFiles.SelectedItem.SecondRevision;
-            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding, line);
+            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding, line, cancellationToken: _viewChangesSequence.Next());
         }
 
         private async Task ShowSelectedFileDiffAsync()
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -121,6 +121,7 @@ namespace GitUI.CommandsDialogs
             ResetSelectedFiles = 9,
             StageSelectedFile = 10,
             UnStageSelectedFile = 11,
+            ShowFileTree = 12
         }
 
         public CommandStatus ExecuteCommand(Command cmd)
@@ -151,6 +152,7 @@ namespace GitUI.CommandsDialogs
                 case Command.ResetSelectedFiles: return ResetSelectedFilesWithConfirmation();
                 case Command.StageSelectedFile: return StageSelectedFiles();
                 case Command.UnStageSelectedFile: return UnstageSelectedFiles();
+                case Command.ShowFileTree: diffShowInFileTreeToolStripMenuItem.PerformClick(); break;
 
                 default: return base.ExecuteCommand(cmd);
             }
@@ -173,6 +175,7 @@ namespace GitUI.CommandsDialogs
             resetFileToParentToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.ResetSelectedFiles);
             stageFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.StageSelectedFile);
             unstageFileToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.UnStageSelectedFile);
+            diffShowInFileTreeToolStripMenuItem.ShortcutKeyDisplayString = GetShortcutKeyDisplayString(Command.ShowFileTree);
 
             DiffText.ReloadHotkeys();
         }
--- a/GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
+++ b/GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
@@ -59,6 +59,7 @@
             this.expandToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.collapseAllToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.FileText = new GitUI.Editor.FileViewer();
+            this.BlameControl = new Blame.BlameControl();
             ((System.ComponentModel.ISupportInitialize)(this.FileTreeSplitContainer)).BeginInit();
             this.FileTreeSplitContainer.Panel1.SuspendLayout();
             this.FileTreeSplitContainer.Panel2.SuspendLayout();
@@ -81,6 +82,7 @@
             // FileTreeSplitContainer.Panel2
             // 
             this.FileTreeSplitContainer.Panel2.Controls.Add(this.FileText);
+            this.FileTreeSplitContainer.Panel2.Controls.Add(this.BlameControl);
             this.FileTreeSplitContainer.Size = new System.Drawing.Size(793, 303);
             this.FileTreeSplitContainer.SplitterDistance = 300;
             this.FileTreeSplitContainer.SplitterWidth = 6;
@@ -339,6 +341,15 @@
             this.FileText.Size = new System.Drawing.Size(487, 303);
             this.FileText.TabIndex = 0;
             // 
+            // BlameControl
+            // 
+            this.BlameControl.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.BlameControl.Location = new System.Drawing.Point(0, 0);
+            this.BlameControl.Margin = new System.Windows.Forms.Padding(0);
+            this.BlameControl.Name = "blameControl";
+            this.BlameControl.Size = new System.Drawing.Size(487, 303);
+            this.BlameControl.TabIndex = 0;
+            // 
             // RevisionFileTreeControl
             // 
             this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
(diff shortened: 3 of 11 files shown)


GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -67,11 +69,14 @@ public RevisionFileTreeControl()
...
                                                                          new FileAssociatedIconProvider());
+            blameControl.HideCommitInfo();
         }
 
-        public void Bind(Action refreshGitStatus)
+        public void Bind(RevisionGridControl? revisionGrid, RevisionFileTreeControl revisionFileTree, Action refreshGitStatus)
         {
+            _revisionGrid = revisionGrid;
             _refreshGitStatus = refreshGitStatus;
+            _revisionFileTree = revisionFileTree;
```
mstv, line 79: Same sequence as arguments❔
```suggestion
            _revisionFileTree = revisionFileTree;
            _refreshGitStatus = refreshGitStatus;
```

mstv (commented):
👍 Feels quite natural.

Wishes for this PR:
- #9437 Update / paint FileTree independent from BlameControl in order to reduce flicker
- #9437 Make switching commits, e.g. `Blame previous revision`, not loose the file selection
image
image

Wishes for later:
- #9437 Scroll FileTree selection to be visible when switching commits, e.g. on `Blame this / previous`
- #9436 Make Commit Info gutter of BlameControl collapsible (simulating "View"; improves scrolling UX)
- Store / restore splitter position of BlameControl (there is a pending issue for `master` anyway: 13caf92691cbefdf05cebb31454d7a8c0c7c3264)
  should be done in #9106
- #9435 Hide unrelated context menu item
image
- #9434 BlameControl: Correctly align line numbers with text
image

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -206,7 +206,7 @@ public FormBrowse(GitUICommands commands, string filter, ObjectId? selectedId =
             repoObjectsTree.Initialize(_aheadBehindDataProvider, _filterBranchHelper, RevisionGrid, RevisionGrid, RevisionGrid);
             toolStripBranchFilterComboBox.DropDown += toolStripBranches_DropDown_ResizeDropDownWidth;
             revisionDiff.Bind(RevisionGrid, fileTree, () => RequestRefresh());
-            fileTree.Bind(() => RequestRefresh());
+            fileTree.Bind(RevisionGrid, fileTree, () => RequestRefresh());
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
```diff
@@ -360,6 +371,7 @@ private void InitializeComponent()
         private System.Windows.Forms.SplitContainer FileTreeSplitContainer;
         private UserControls.NativeTreeView tvGitTree;
         private Editor.FileViewer FileText;
+        private Blame.BlameControl blameControl;
```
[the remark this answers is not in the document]
gerhardol: will change

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: ok

gerhardol:
> Wishes for this PR:

Most of these are for BlameControl should be a separate issue (#7598 is quite confusing as is...)
Improvements are necessary.

Also if Blame is improved, there may be need for a configurable way to  show Blame/View (or even Blame/Diff).
Some discussions in #7598, this is a first implementation to fond what is really needed.

gerhardol:
Converted this to a draft and based it on #9425 (approved, to be merged soon) and #9426 as there are some functionality required.
The commits from 7037db268d7d61aa3fdffe75f4469cc60321eaf0 *RevFileTree: Blame instead of View* to 1c921d7f27c8e20b7835dfb3ad3450ce5c26b825 *fixup! RevFileTree: Blame instead of View* are unchanged.

Some new additions:
* flicker selecting is removed in c7f0165d0ed6b51077d552cd0ff3c3dd476cdbf0 *fixup! avoid flicker switching files*
* The existing issue is removed 58e63c02e5ac9697c766cb292d5177bd870b975b *Blame; Hide EnableAutomaticContinuousScroll*
* To avoid that slow blame is a blocker, it is possible to View files as before.
The shortcut key `B` is used to toggle View/Blame instead of opening a new instance. (the value is stored in settings.
Note: a similar change should be added to RevDiff, `B` to Blame files.

gerhardol:
Cleaned up commits a little, adding RevDiff Blame
Ready for review again
Further changes are presented in #9445

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -51,26 +51,29 @@ public partial class RevisionFileTreeControl : GitModuleControl
...
                                                                          new FileAssociatedIconProvider());
+            BlameControl.HideCommitInfo();
+            blameToolStripMenuItem1.Checked = AppSettings.RevisionFileTreeShowBlame;
         }
 
-        public void Bind(Action refreshGitStatus)
+        public void Bind(RevisionGridControl? revisionGrid, Action refreshGitStatus)
         {
+            _revisionGrid = revisionGrid;
             _refreshGitStatus = refreshGitStatus;
```
GintasS, line 77: Since you're using nullable reference type for `RevisionGridControl`, I'm wondering if the `Action refreshGitStatus `parameter should be too `Action? refreshGitStatus` here?

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -351,35 +356,59 @@ private void tvGitTree_AfterSelect(object sender, TreeViewEventArgs e)
...
+            switch (gitItem.ObjectType)
             {
-                switch (gitItem.ObjectType)
-                {
-                    case GitObjectType.Blob:
-                    case GitObjectType.Commit:
+                case GitObjectType.Blob:
                     {
-                        GitItemStatus file = new(name: gitItem.FileName)
+                        if (!blameToolStripMenuItem1.Checked)
```
GintasS, line 370: What do you think about not using this logical negation and instead writing `blameToolStripMenuItem1.Checked == false`? Would be more readable, you can miss that `!`

GintasS: Also, in the method `ExecuteCommand` in `RevisionDiffControl,cs`, you use (reuse) one style of switch statements (making cases inline) and here you do in one style, FYI.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -351,35 +356,59 @@ private void tvGitTree_AfterSelect(object sender, TreeViewEventArgs e)
...
                     }
-                }
+
+                default:
+                    return ClearOutputAsync();
+            }
+
+            Task ViewGitItemAsync(GitItem gitItem)
+            {
+                GitItemStatus file = new(name: gitItem.FileName)
```
GintasS, line 392: Just out of curiosity, what is this syntax, this is C# 9.0 feature?

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -428,10 +457,15 @@ private void tvGitTree_KeyDown(object sender, KeyEventArgs e)
...
         private void blameMenuItem_Click(object sender, EventArgs e)
         {
-            if (tvGitTree.SelectedNode?.Tag is GitItem gitItem)
+            if (tvGitTree.SelectedNode?.Tag is not GitItem gitItem)
             {
-                UICommands.StartFileHistoryDialog(this, gitItem.FileName, _revision, true, true);
+                return;
             }
+
+            blameToolStripMenuItem1.Checked = !blameToolStripMenuItem1.Checked;
```
GintasS, line 465: You can miss that `!` so easily here, but it is up to you.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -212,6 +215,8 @@ public void LoadRevision(GitRevision? revision)
 
                 if (tvGitTree.SelectedNode is null)
                 {
+                    BlameControl.Visible = false;
+                    FileText.Visible = true;
                     FileText.Clear();
```
GintasS, line 220: What do you think about having a separate method to control these `BlameControl `and `FileText `Visible states? I can see in a couple of instances you setting these two to false and true respectively.

GintasS (commented):
Wrote a few comments regarding style basically.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features
Compact code - clearer and more compact than `var`

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: That depends on what the caller provides, but in this case it should be changed
(comments on existing code is welcome, bigger changes should be separated commits or even PRs though)

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: > What do you think about not using this logical negation and instead writing blameToolStripMenuItem1.Checked == false? Would be more readable, you can miss that !

In this case, I prefer the compact style

> Also, in the method ExecuteCommand in RevisionDiffControl,cs, you use (reuse) one style of switch statements (making cases inline) and here you do in one style, FYI.

Is this about how Blame is activated?
In that case it is intentional, FileTree remembers Blame/View for those that prefer to only view (that is faster).
For RevDiff it is always a one time activation, next selection is always Diff (or New etc) again.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: I guess pattern matching could be used, but I want to have `!` as our convention

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: Similar in FileTree and RevDiff, just one of the controls are active at a time.
In this case there are only two controls and only true/false switch so I prefer to not hide this in a method.
For RevDiff, I separated `BlameSelectedFileDiffAsync()` to a separate method that was kept close to the existing `ShowSelectedFileDiffAsync()` to show the similarities.
(Good suggestion still...)

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: `RevisionGridControl` should not be NRT, caller is always non null
In RevDiff `refreshGitStatus` should be NRT too.

gerhardol:
Rebased on master to resolve MC (and master eeb8f67ddbf87ccc88741fe31c31ac444948a266 is unusable, fixed in adbb410d5b65b8d23d8bb71541179cbe573478e8).
Also squashed the NRT in https://github.com/gitextensions/gitextensions/pull/9424#discussion_r685302418

GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
```diff
@@ -339,6 +341,15 @@ private void InitializeComponent()
...
             this.FileText.TabIndex = 0;
             // 
+            // BlameControl
+            // 
+            this.BlameControl.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.BlameControl.Location = new System.Drawing.Point(0, 0);
+            this.BlameControl.Margin = new System.Windows.Forms.Padding(0);
+            this.BlameControl.Name = "blameControl";
+            this.BlameControl.Size = new System.Drawing.Size(487, 303);
+            this.BlameControl.TabIndex = 0;
```
mstv, line 351: Conflict with `FileText.TabIndex` (although they are not displayed at the same time).

mstv (commented):
LGTM

GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
gerhardol: That was intentional, really a need to separate them?

gerhardol:
Rebased without changes, to simplify reviews and get the build successful

GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
RussKie: As a general rule: no item should have identical indices. Though this is the keyboard tab order, i.e. used when tab between controls. Accessibility tools don't use it (AFAIK), and use the order in which controls are added into respective parents instead.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -51,26 +51,29 @@ public partial class RevisionFileTreeControl : GitModuleControl
         private readonly Stack<string> _lastSelectedNodes = new();
         private readonly IRevisionFileTreeController _revisionFileTreeController;
         private readonly IFullPathResolver _fullPathResolver;
-        private readonly IFindFilePredicateProvider _findFilePredicateProvider;
+        private readonly IFindFilePredicateProvider _findFilePredicateProvider = new FindFilePredicateProvider();
```
RussKie, line 54: Why is this necessary?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -449,8 +453,21 @@ private void ResetSelectedItemsTo(bool actsAsChild)
...
+        private Task BlameSelectedFileDiffAsync()
+        {
+            BlameControl.Visible = true;
+            DiffText.Visible = false;
+            GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
+                ? _revisionGrid.GetActualRevision(_revisionGrid.CurrentCheckout)
+                : DiffFiles.SelectedItem.SecondRevision;
+            BlameControl.LoadBlame(rev, null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, null, DiffText.Encoding);
+            return Task.CompletedTask;
+        }
```
RussKie, line 465: This is really confusing - there is nothing async in this method, and everything must be done on the UI thread.
Please change it into a sync method, and update the the call site to invoke it async, if necessary.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
RussKie: > What do you think about not using this logical negation and instead writing `blameToolStripMenuItem1.Checked == false`? Would be more readable, you can miss that `!`

This is C# intrinsic (and C-like languages in general). The only time one would write ` == false` when the left operand is a nullable.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -351,35 +356,59 @@ private void tvGitTree_AfterSelect(object sender, TreeViewEventArgs e)
...
-                            TreeGuid = gitItem.ObjectId,
-                            IsSubmodule = gitItem.ObjectType == GitObjectType.Commit
-                        };
+                            return ViewGitItemAsync(gitItem);
+                        }
 
-                        return FileText.ViewGitItemAsync(file, gitItem.ObjectId);
+                        BlameControl.Visible = true;
+                        FileText.Visible = false;
+                        BlameControl.LoadBlame(_revision, null, gitItem.FileName, _revisionGrid, null, FileText.Encoding);
```
RussKie, line 377: Please add argument names for `null`.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: It is not necessary for this PR, it was a change I had done (and believe should be changed) that was refactored to a separate commit

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: This was done in the same way as `ShowSelectedFileDiffAsync()`, the call in `blameToolStripMenuItem_Click()` need to async anyway (even if this is sync).
Cancelation token should be added to LoadBlame too, similar to how `ViewChangesAsync()` was changed.
(that will require Git commands to be cancellable first to be really useful)

So I would prefer to keep it as is.

GitUI/CommandsDialogs/RevisionFileTreeControl.Designer.cs
gerhardol: Do you request to change this?

gerhardol:
Added two more features:
* When blaming in RevDiff or FileTree, select the current line in fileviewer in Blame (especially useful in RevDiff)
* Cancellation support in LoadBlame() (in between operations, not in methods)

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: Changed this - there are operations blocking the UI thread, but not for long.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -51,26 +51,29 @@ public partial class RevisionFileTreeControl : GitModuleControl
...
                                                                          new GitRevisionInfoProvider(() => Module),
                                                                          new FileAssociatedIconProvider());
+            BlameControl.HideCommitInfo();
+            blameToolStripMenuItem1.Checked = AppSettings.RevisionFileTreeShowBlame;
         }
 
-        public void Bind(Action refreshGitStatus)
+        public void Bind(RevisionGridControl revisionGrid, Action? refreshGitStatus)
         {
+            _revisionGrid = revisionGrid;
```
RussKie, line 76: I very much dislike this - this creates further coupling between various controls, and making it harder to reason about each individual control's functionality, and further complicating testing stories.

gerhardol: This is required for BlameControl so the same as in FormFileHistory.
You can see this as a step to remove a lot of code in GE: The follow up #9445 obsoletes FormFileHistory, so this is a step to remove a complete form.

gerhardol:
Rebased and squashed without changes, prepared for rebase merge when approved.

gerhardol:
Rebased to resolve a merge conflict

RussKie:
I ran it locally briefly, seems to work. Need time to form a better opinion, but overall I think it is a good change.
Two things I've noticed:
1. "Blame" isn't sticky, i.e. when change file selection the view changes to the "diff". Feels weird.
2. Attempt to open history "feels" like it getting ignored. At least there is no immediate feedback, the new window doesn't open, or opens after a significant delay.

RussKie (approved):

gerhardol:
>     1. "Blame" isn't sticky, i.e. when change file selection the view changes to the "diff". Feels weird.

Blame/View is sticky in FileTree tab.
Intentional in diff tab, I want to go to diff quickly in diff tab. This is partly as BlameControl is slower, annoying to show first then to blame.
I have used this for a month or so now, this is something I appreciate so I prefer to keep this.

>     2. Attempt to open history "feels" like it getting ignored. At least there is no immediate feedback, the new window doesn't open, or opens after a significant delay.

Do you mean H? That is not changed by this PR.
Or open FileTab? Open FileTree with Blame requires a little longer time to open Blame to View, That is why Blame/View is sticky, stored in settings.

PR #9425, opened 2021-07-28
ToggleBetweenArtificialAndHeadCommits: Go to WorkTree if HEAD is hidden

gerhardol:
Contributes to #7598

## Proposed changes

A few unrelated changes related to filtered commits

- From #9393 change the confusing diff message (especially when selecting artificial commits without changes).

- From #9393 squash! Use default value for ShowUncommittedChangesIfPossible

- Improve performance for GitModule.GetParents
git-rev-parse is faster than git-log
For me in GE it is about 0.18 s vs 0.30 s.
Also check input and cache results

The command will be used when calculating the Diff for filtered commits in a later PR.

- ToggleBetweenArtificialAndHeadCommits: Go to WorkTree if HEAD is hidden
Note: This has no effect until #9393 is merged as artificial commits are not shown

If HEAD is not visible (in a filtered view) nothing happens for Ctrl-< as there is nowhere to go
In this case go to WorkTree commit instead.

- squash! Browse: Remove unused _artificialStatus
_artificialStatus was previously used to cache the commit-count over operations, to quickly display the old info.
Since som time ago, this has no effect as null status is distributed before the cache is distributed.
(This was not a good idea in the first place, possibly incorrect data should not be shown.)

## Test methodology 

Manual

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1004,17 +1004,21 @@ namespace GitCommands
             }
         }
 
-        public IReadOnlyList<ObjectId> GetParents(ObjectId commitId)
+        public IReadOnlyList<ObjectId> GetParents(ObjectId objectId)
         {
-            GitArgumentBuilder args = new("log")
+            if (objectId.IsArtificial)
             {
-                "-n 1",
-                "--format=format:%P",
-                commitId
+                // For WorkTree could Index be returned, but for Index HEAD should be available anyway
+                throw new InvalidOperationException(nameof(objectId));
+            }
+
+            GitArgumentBuilder args = new("rev-parse")
+            {
+                $"{objectId}^@"
             };
-            return _gitExecutable
-                .GetOutput(args)
-                .LazySplit(' ', StringSplitOptions.RemoveEmptyEntries)
+            return _gitExecutable.Execute(args, cache: GitCommandCache)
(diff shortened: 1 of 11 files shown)


mstv (commented):
LGTM, have not run

mstv (approved):

gerhardol:
Plan to cleanup/squash commits and rebase merge in 2 days

GitCommands/Git/GitModule.cs
```diff
@@ -1004,17 +1004,21 @@ string ProcessDiffNotes(int startIndex)
...
+        public IReadOnlyList<ObjectId> GetParents(ObjectId objectId)
         {
-            GitArgumentBuilder args = new("log")
+            if (objectId.IsArtificial)
             {
-                "-n 1",
-                "--format=format:%P",
-                commitId
+                // For WorkTree could Index be returned, but for Index HEAD should be available anyway
+                throw new ArgumentException(nameof(objectId));
```
RussKie, line 1012: I think InvalidOperationException may be more appropriate

RussKie (approved):

GitCommands/Git/GitModule.cs
gerhardol: Will change and rebase merge.

RussKie:
Merge at will

Issue #9411, opened 2021-07-23
Commit-icon in UI does not update on new changed files 


Bejoscha:
## Current behavior

When local files change in a repo, the commit-icon is supposed to turn orange (from green) and list the number of changed files.

However, since the last update to 3.5.1.12196 this functionality isn't as reliable anymore. I find myself often looking at an orange "commit" message with incorrect number (too few files). 

## Expected behavior

Show current situation correctly.

## Steps to reproduce

Have a repo with a few local files changed (not staged). the icon is orange and lists the number of changed files.
Now mass-change several local files in the repo (extended folder-structure in my case). Note that the icon (still organe) does not update the number of different files.
However, pressing the commit button to get to the commit-dialog shows that all files are (now?) recognized, and leaving the dialog without committing will update the icon as well.

The behavior is not 100% reproducible.

## Screenshots

![image](https://user-images.githubusercontent.com/6987920/126757912-5c484525-271d-4070-8666-99c0952cd7e9.png)

## Did this work in previous version of GitExtensions

Yes. In 3.5.0 it worked very reliably. 

## Environment

- Git Extensions 3.5.1.12196
- Build 9099a1e76d01dbbd6896d6b026027db0b68fba96
- Git 2.21.0.windows.1 (recommended: 2.30.0 or later)
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4390.0
- DPI 144dpi (150% scaling)

## Diagnostics

gerhardol:
git-status may require quite some resources, the command may require several seconds, especially with corporate viruses like Symantec.
Therefore, the command should not run immediately when changes are detected in the file system. Especially when building, one core could be completely occupied with running git-status with the same result all over again (caching a list of ignored files was done in GE 2.x, that could use GB of RAM).
On the other hand, the first detection at changes and interactive changes should be fast.

There is therefore some hysteresis delaying git-status. The first change detected as well as after exiting GE forms is done immediately. In 3.4, the wait was up to 5 seconds for subsequent changes outside GE. In 3.5.0 the time was changed to 30 s instead to limit git-status running in the background.
If you force refresh or open Commit the status is refreshed.

The coloring of the button is not changed.

Note that git-for-windows has introduced an internal handling in core.usebuiltinfsmonitor. That improves the runtime, but for me this handling occasionally miss changes, never report them, requiring a manual `git -c core.usebuiltinfsmonitor=false status`.

PR #9432, opened 2021-08-02
RevGrid commit count: Do not reset when creating artificial commits

gerhardol:
## Proposed changes

The commit count handles the GitStatusMonitor that updates the commit count.
This is done in a number of situations, also when refreshing the grid.

RevGrid loading commits could be finished after the git status is finished
and should not reset the commit count.

This probably very seldom a problem in 3.5 or master before #9393, git-status is slightly delayed after the request, so the artificial commit that normally is added very early should be affected.
If you check out the first commit in a repo, it would be more common.
It occurs more often with #9393 that inserts the artificial commits after all other commits if HEAD is filtered out.
(Before #9393 the artificial commits were not displayed at all if HEAD was filtered, so the functionality did not exist.)

Note: Commit-count on the button is not affected by this, just the count in the grid. 
(There is a report in #9411 for something similar not affected.)
The count on the button must match the count in the grid, so the extra status init in the grid removed by this PR has no benefit.

Git-status could also be triggered when revision loading is completed, instead of starting after the short delay and running in parallel with git-log.

This is a candidate for 3.5.2 too.

### Before

image

### After

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -105,6 +105,8 @@ namespace GitUI
         private readonly RevisionGraphColumnProvider _revisionGraphColumnProvider;
         private readonly DataGridViewColumn _maximizedColumn;
         private DataGridViewColumn? _lastVisibleResizableColumn;
+        private readonly ArtificialCommitChangeCount _workTreeChangeCount = new();
+        private readonly ArtificialCommitChangeCount _indexChangeCount = new();
 
         private RefFilterOptions _DONT_USE_ME_DIRECTLY_refFilterOptions = RefFilterOptions.All | RefFilterOptions.Boundary;
 
@@ -1171,9 +1173,6 @@ namespace GitUI
                     return;
                 }
 
-                _indexChangeCount = new ArtificialCommitChangeCount();
-                _workTreeChangeCount = new ArtificialCommitChangeCount();
-
                 var userName = Module.GetEffectiveSetting(SettingKeyString.UserName);
                 var userEmail = Module.GetEffectiveSetting(SettingKeyString.UserEmail);
 
@@ -1207,8 +1206,6 @@ namespace GitUI
                 };
 
                 _gridView.Add(indexRev, insertAsFirst: insertAsFirst);
-
-                UpdateArtificialCommitCount(_artificialStatus, workTreeRev, indexRev);
             }
 
             void OnRevisionReaderError(Exception exception)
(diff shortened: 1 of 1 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2235,10 +2221,12 @@ void UpdateChangeCount(ObjectId objectId, IReadOnlyList<GitItemStatus> status)
             {
                 Debug.Assert(objectId == ObjectId.WorkTreeId || objectId == ObjectId.IndexId,
                     $"Unexpected Git object id {objectId}");
-                var changeCount = GetChangeCount(objectId);
                 var staged = objectId == ObjectId.WorkTreeId ? StagedStatus.WorkTree : StagedStatus.Index;
                 var items = status.Where(item => item.Staged == staged).ToList();
-                changeCount?.Update(items);
+
+                var changeCount = GetChangeCount(objectId);
+                Validates.NotNull(changeCount);
```
mstv, line 2228: will fail on git error

mstv (commented):
I have not grasped what triggers the update of the commit counts when the artificial have been added. How long can it be delayed?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: No, just getting the object

```
        public ArtificialCommitChangeCount? GetChangeCount(ObjectId objectId)
        {
            return objectId == ObjectId.WorkTreeId
                ? _workTreeChangeCount
                : objectId == ObjectId.IndexId
                    ? _indexChangeCount
                    : null;
        }
```

gerhardol:
> I have not grasped what triggers the update of the commit counts when the artificial have been added. How long can it be delayed?

Much to complicated to explain here, but status monitor is stopped prior to opening a new repo and opening Settings, Commit.
Started after opening the repo (and a few other commands that may affect the commit count like reset), then an interactive request is submitted, waiting 200-300 ms (to not interfere so much with other commands, yet be hardly noticable). 
When the filesystem is updated outside GE, the delay is 1s, subsequent changes 30s.

mstv:
> > I have not grasped what triggers the update of the commit counts when the artificial have been added. How long can it be delayed?

> Much to complicated to explain here, but status monitor is stopped prior to opening a new repo and opening Settings, Commit.
Started after opening the repo (and a few other commands that may affect the commit count like reset), then an interactive request is submitted, waiting 200-300 ms (to not interfere so much with other commands, yet be hardly noticable).
When the filesystem is updated outside GE, the delay is 1s, subsequent changes 30s.

Thank you.
This PR seems to be necessary. We will see how long it delays the updates. Can be improved later if necessary.

mstv:
The MC needs to be resolved.

gerhardol:
> The MC needs to be resolved.

Was working on that...
Also made the `ArtificialCommitChangeCount` objects `readonly`

> This PR seems to be necessary. We will see how long it delays the updates. Can be improved later if necessary.

It is necessary to present the improvements in #9393 (before this was probably not an issue).

Hmm, the commit count button is not gray when it should be just after pushing `F5`...

mstv (approved):

gerhardol:
Plan to squash merge tomorrow, also for 3.5

PR #9458, opened 2021-08-08
.NET application bootstrap idea. Fixes #9276

### LINKED ISSUE #9276: .NET application bootstrap idea
Use https://github.com/Tyrrrz/DotnetRuntimeBootstrapper for the app to automatically install necessary .NET runtime.

GintasS:
Fixes #9276

## Proposed changes

- Use DotnetRuntimeBootstrapper NuGet package for bootstrapping the app.
- 
- 
## Test methodology 

- Used existing tests.
- 
- 

## Test environment(s) 

git version 2.32.0.windows.2
Windows 10

----

### DIFF
--- a/GitExtensions/GitExtensions.csproj
+++ b/GitExtensions/GitExtensions.csproj
@@ -21,6 +21,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="DotnetRuntimeBootstrapper">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
     <PackageReference Include="GitInfo">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
--- a/Packages.props
+++ b/Packages.props
@@ -7,6 +7,7 @@
     <PackageReference Update="AppInsights.WindowsDesktop" Version="2.17.10" />
     <PackageReference Update="Ben.Demystifier" Version="0.4.1" />
     <PackageReference Update="ConEmu.Core" Version="21.7.18" />
+    <PackageReference Update="DotnetRuntimeBootstrapper" Version="1.1.2" />
     <PackageReference Update="EnvDTE" Version="16.10.31320.204" />
     <PackageReference Update="ExCSS" Version="4.1.0" />
     <PackageReference Update="GitInfo" Version="2.1.2" />


GintasS:
1. I added the NuGet package to the Startup project GitExtensions. Should I add this NuGet package somewhere else too?
2. When I run the "Release" mode in Visual Studio, I got all the files, run the .exe, but it is somehow very slow. Looks like this might not be related, since I have created a release without the changes and it is too slow.

RussKie:
This is about E2E user experience - e.g. a user performs a fresh install of the app on a workstation, and upon the first launch the app downloads the necessary .NET runtime.
We're still waiting for [WiX Toolset on .NET](https://github.com/wixtoolset/issues/issues/5627) (i.e. use `dotnet` to pack the msi), but I'm pretty sure we can still build the installer using `msbuild`.

GintasS:
I have tried running the new release that I have created inside a Virtual Box Windows VM.
It seems to work as far as I understand.
Part 1:
https://user-images.githubusercontent.com/20372363/128632705-198726d7-1de9-47bb-a83f-451c59a2d93b.mp4

Part2:
https://user-images.githubusercontent.com/20372363/128632803-8e9eeed2-dd52-4ef8-9adb-8ac943b173f6.mp4

mstv:
My 2ct: Regarding code should only be run by the installer. Nothing should be added to every start of the `GitExtensions.exe` - perhaps except a quick check of a setting whether it has run once.
`DotnetRuntimeBootstrapper` does not describe how it tries to minimize the overhead added.

GintasS:
So for now, there are two blockers, first one is the WiX Toolset issue and the overhead that mstv mentioned?

vbjay:
And when you run portable. That's the whole point of it being one the exe.

vbjay:
Having a bootsteapper that gets it pulled down and then runs exe is what I
meant but you get the idea.

GitExtensions/GitExtensions.csproj
```diff
@@ -21,6 +21,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="DotnetRuntimeBootstrapper">
```
RussKie, line 24: We need to package this into the installer (.\Setup\Product.wxs). You should be able to build the installer using `msbuild` (from the develop prompt), e.g. 
```
Setup> msbuild
```

RussKie (commented):
👍

RussKie:
> Regarding code should only be run by the installer.

I disagree, because we have the portable version, which must as well install required components. 
Whilst Windows Forms can be published as self-contained, that'd be hundreds of megabytes. It also isn't trim compatible because it uses reflections quite extensively and has COM dependencies.

GitExtensions/GitExtensions.csproj
GintasS: I have run the command, but seems to fail, any ideas?

image

GintasS: Also, what about just selecting "Release" in Visual Studio and building from there? I have used this for testing in a Windows VM.

RussKie:
I've never done it from VS :) And it's been awhile since I done it manually
from cli too, so dont remember what the build target was defined...

RussKie:
This is how the msi is being created
https://github.com/gitextensions/gitextensions/blob/6bdeda2d958cbf7d311b44ade9fade5c183499be/GitExtensions/Project.Publish.targets#L199

RussKie:
Here's me building it locally against 3.5 branch (with bogus values):
image

NB: I had to install .NET Framework 2.0 (ugh), because WiX Toolset 3.11.2 requires it. There's probably a way to use [4.0.0-preview.0](https://www.nuget.org/packages/wix/4.0.0-preview.0) but it requires an investigation.

PR #9459, opened 2021-08-08
Artificial commits: Invalidate rowcache when inserting

gerhardol:
Followup to #9393

## Proposed changes

Inserting artificial commits first if HEAD was not added to the grid was added in #9393.
In some situations the grid would not be refreshed correctly, requiring a manual scroll.
This normally would occur for bigger repos only.
When working with the PR, I could not get a stable repro and could not address it.
This is a follow up to invalidate the row cache properly to refresh the graph in the grid.

### Before

image

### After

The normal commits are being displayed while the git-log output is created, so initially the artificial commits are not shown.
After a few seconds the graph is updated as below (that would require a scroll before).

image

## Test methodology 

manual

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -185,32 +185,36 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         /// <param name="insertAsFirst">Insert the (artificial) revision first in the graph.</param>
         public void Add(GitRevision revision, RevisionNodeFlags types, bool insertAsFirst = false)
         {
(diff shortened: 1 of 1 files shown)


RussKie (approved):
I don't have a deep understanding of the functionality. The changes look sensible...

GintasS (commented):
Same here, apart from the ! negation, couldn't pick on anything, no knowledge here.

GintasS (approved):

mstv (approved):
> I don't have a deep understanding of the functionality. The changes look sensible...

same for me

gerhardol:
> 
> 
> > I don't have a deep understanding of the functionality. The changes look sensible...
> 
> same for me

You know this better than me...

But I tried to add comments with what I know.

Will squash merge

PR #5654, opened 2018-10-30
Fix for issue "Delete dialog appears when pressing ctrl+alt+del on latest commit of a branch" with id 5636

AbhiAgarwal192:
Fixes #5636

Changes proposed in this pull request:
- On clicking ctrl+alt+del , modifiers ctrl+alt were ignored and it was treated as if del key was fixed. With this fix if modifiers are present then del key is ignored.

 
Screenshots before and after (if PR changes UI):
- None

What did I do to test the code and ensure quality:
- Manual testing of the change.

Has been tested on (remove any that don't apply):
- GIT 2.11 and above: 2.18
- Windows 7 and above : windows 10.

PR #9460, opened 2021-08-09
Don't show delete dialog on CTRL+ALT+DEL

### LINKED ISSUE #5636: Delete dialog appears when pressing ctrl+alt+del on latest commit of a branch
Bug.

If the latest commit of a non-checked out branch is highlighted, and the main Git Extensions window has focus, then when you press ctrl+alt+del, the delete dialog appears.

Edit: The Windows ctrl+alt+del screen comes up and you have to close it before the delete dialog appears.

1. Highlight the latest commit of a branch you do not have checked out.
2. Press ctrl+alt+del.

Delete key press should be ignored when pressing ctrl+alt+del.

- GitExtensions version: 2.51.04
- GIT version: N/A
- OS version: Windows 7 SP1
- .NET version: N/A

Not sure.

RussKie:
Fixes #5636
Closes #5654

## Test methodology 

- manual

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -1572,9 +1572,15 @@ namespace GitUI
 
         private void OnGridViewKeyUp(object sender, KeyEventArgs e)
         {
-            var selectedRevision = LatestSelectedRevision;
+            GitRevision? selectedRevision = LatestSelectedRevision;
 
             if (selectedRevision is null)
+            {
+                return;
+            }
+
+            // https://github.com/gitextensions/gitextensions/issues/5636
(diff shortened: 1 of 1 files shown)


GintasS (approved):
Looks good 👍

mstv (approved):

DamitSe (approved):
Change looks good. However, I wonder if we can ignore modifiers only for Delete (for example, what if I want to assign Ctrl+F2 to an action?)

mstv:
> However, I wonder if we can ignore modifiers only for Delete (for example, what if I want to assign Ctrl+F2 to an action?)

Valid point but `OnGridViewKeyUp` handles just two hard-coded hotkeys `F2` and `Del`.
If we really add another, it can be handled before the check.

gerhardol (approved):

PR #9468, opened 2021-08-09
Avoid popup at successful FileViewer patching

### LINKED ISSUE #9319: UI: clean cherrypick feedback modal presents as error
## Current behaviour

When cherrypicking changes from another commit into the current branch, the UI always shows an error message, even when the cherrypicking was successful and applied cleanly.

## Expected behaviour

No error message when the patch applied cleanly.

## Steps to reproduce

* Have a branch with some changes to File A.
* Cherrypick the changes from File B (which hasn't been touched in the current branch and for which the changes would apply cleanly) from another branch.
* See the error display. 

## Screenshots

image

Yes, IIRC this did not happen prior to 3.5.0.

## Environment

- Git Extensions 3.5.0.11998
- Build 747591af7e0fa8bbacd241994db245b9e9c14266 (Dirty)
- Git 2.32.0.windows.1
- .NET Framework 4.8.4360.0
- DPI 144dpi (150% scaling)

gerhardol:
Fixes #9319

## Proposed changes

Previously, when performing line patches and Git reported a message, GE displayed it.
With more recent Git versions, there are messages also when the lines were applied without problems.

This PR only raises the popup when Git exists with an error.

Note: Git has several strategies applying patches so if one method fails, it tries another.
So the Git message may include "error: " or "warning: " also when patching was OK.
I have not found if any of these situations should be presented anyway.
This PR will not present them.
To be able to debug this also in production releases, a Trace message was added.
Maybe this should be available as a status message somehow.
That is fine tuning, I personally see no need for that.

## Screenshots 

The examples line patches a file that already have the changes, then reverting the same change.

### Before

Applying a patch successfully (or the same patch again)

image

Reverting the same patch - requires a second try by Git before successful

image

### After

The popups do not appear unless there are errors, like if trying to revert the change twice 
(unchanged behavior)

image

## Test methodology 

Manual

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -1430,8 +1430,7 @@ namespace GitUI.Editor
                 { !stage, "--reverse" }
             };
 
-            string output = Module.GitExecutable.GetOutput(args, patch, throwOnErrorOutput: false);
-            ProcessApplyOutput(output, patch, patchUpdateDiff: true);
+            ProcessApplyOutput(args, patch, patchUpdateDiff: true);
         }
 
         /// <summary>
@@ -1500,15 +1499,7 @@ namespace GitUI.Editor
                 { currentItemStaged, "--reverse --index" }
             };
 
-            string output = Module.GitExecutable.GetOutput(args, patch, throwOnErrorOutput: false);
-            if (EnvUtils.RunningOnWindows())
-            {
-                // remove file mode warnings
-                Regex regEx = new("warning: .*has type .* expected .*", RegexOptions.Compiled);
-                output = output.RemoveLines(regEx.IsMatch);
-            }
-
-            ProcessApplyOutput(output, patch, patchUpdateDiff: true);
+            ProcessApplyOutput(args, patch, patchUpdateDiff: true);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


GitUI/Editor/FileViewer.cs
```diff
@@ -1557,18 +1548,33 @@ private void ApplySelectedLines(bool allFile, bool reverse)
...
+                output = output.RemoveLines(regEx.IsMatch);
+            }
+
+            if (!result.ExitedSuccessfully && (patchUpdateDiff || !MergeConflictHandler.HandleMergeConflicts(UICommands, this, false, false)))
+            {
+                MessageBox.Show(this, $"{output}\n\n{Encoding.GetString(patch)}", TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+            }
+            else if (!result.ExitedSuccessfully || output.StartsWith("error: ") || output.StartsWith("warning: "))
+            {
+                // git-apply may fail on first attempt but succeed in subequent attempts
```
mstv, line 1571:
```suggestion
                // git-apply may fail on first attempt but succeed in subsequent attempts
```

GitUI/Editor/FileViewer.cs
```diff
@@ -1557,18 +1548,33 @@ private void ApplySelectedLines(bool allFile, bool reverse)
...
+                // Trace such occurrences that may be interesting, some of these should maybe be presented to the user
+                Trace.WriteLineIf(!string.IsNullOrWhiteSpace(output), $"Patch output: {result.ExitCode}:{output} for: git {args}");
+            }
+            else if (!string.IsNullOrWhiteSpace(output))
+            {
+                Debug.WriteLine($"Patch output: {result.ExitCode}:{output} for: git {args}");
```
mstv, line 1577: Just curious: How do see this output between all the noisy "thread has exited" messages?

mstv: We could add the [error] output to the Git Command Log.

mstv (approved):
LGTM, have not run

GitUI/Editor/FileViewer.cs
gerhardol: > Just curious: How do see this output between all the noisy "thread has exited" messages?

Right now, you dont unless you break in the debugger or view a log.
There is a need to set trace levels, to view some messages only.

> We could add the [error] output to the Git Command Log.

I have also considered that, for messages the user could be interested in (not thread exit or time to parse git-log for revGrid...)

For Debug vs Trace, I would handle the output the same, just not visible in Release.

GitUI/Editor/FileViewer.cs
gerhardol: ok

gerhardol:
Squash merge in two days or after @RussKie approval

GitUI/Editor/FileViewer.cs
RussKie: What's the point of these trace and debug messages? These aren't visible to a user...

`(!result.ExitedSuccessfully || output.StartsWith("error: ") || output.StartsWith("warning: "))` to me implies there was an error - why are we not showing an error/throwing?

gerhardol: The approach to ignoring popups is quite aggressive here. This can be motivated with that there are no subsequent GE processing.
However, if this is too aggressive it is possible to debug this with something like https://docs.microsoft.com/en-us/sysinternals/downloads/debugview
(or edit .ini to print to console)

I would like to clean up the trace printouts and set levels that can be configured.

> to me implies there was an error - why are we not showing an error/throwing?

Git has succeeded to apply the patch with a secondary strategy, so it *should* not be an error

RussKie (approved):

PR #9445, opened 2021-08-06
Blame/filter in Browse, replace formfilehistory

gerhardol:
Closes #7598

Based on #9424.
This PR could be split in several PRs, submitted as is to describe the "initial completion" of #7598 and describe what could be improved.

I consider this to be ready for review, but may tweak the functionality until #9424 is merged.

## Proposed changes

* See screenshots below

* RevDiff and FileTree selects filtered files by default when changing commits, the filtered file is "followed" when switching commits (unless another file is selected and it exists in the next commit, as already implemented).

* Some refactoring to allow for new functionality separated.

Future:
* Present active filters better. Some way to deactivate filters in a simple way. (Some discussion in #7598).
* A few tests could be added (some added)

### After

Show path filter in title (last part is due to Debug build, unchanged)
image

Separator changed as T/F/H/B hotkey commands are related

image

Context command, hotkey `F` to filter a file in the grid
(similar to open a new instance)

image

image

Configuration to open a Browse with filter instead of FileHistory.
Enabled by default.
This is put in Browse as Appearance is too busy already

image

## Test methodology 

manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9477, opened 2021-08-11
Reset revision filter when switching repos

### LINKED ISSUE #3153: Clear "file filter" when I switch repository
I'd like the "file filter" to turn off automatically when I switch repository, because I think it makes sense from a usability point of view. To reproduce my frustration:

1) Cntrl+F
2) Enter a valid file filter and tick it on, to view just changes to those files
3) In the bar at the top, switch to a different repository
4) Your whole view is now empty because despite the directory/file not being found in the new repository, the filter is still active

Preferably, the setting should be stored with the repository, so that I can have one filter active on one repository and still use another repository without losing that information in the first repository I was viewing. An alternative would be to check if the file/folder exists after switching, turning it off if it doesn't match. It would also be fine to just turn off the tick whenever you switch repository.

A similar approach would be nice for branches, where I don't want to filter by a branch that doesn't exist, and would prefer it to be stored with the repository.

RussKie:
Closes #3153

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1979,6 +1979,7 @@ namespace GitUI.CommandsDialogs
             _submoduleStatusProvider.Init();
             ToolStripFilters.SetBranchFilter(string.Empty, refresh: false);
             ToolStripFilters.SetRevisionFilter(string.Empty);
+            RevisionGrid.DisableFilters();
 
             UICommands = new GitUICommands(module);
             if (Module.IsValidGitWorkingDir())
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1978,6 +1978,7 @@ namespace GitUI.CommandsDialogs
             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
             _submoduleStatusProvider.Init();
             ToolStripFilters.SetBranchFilter(string.Empty, refresh: false);
+            ToolStripFilters.SetRevisionFilter(string.Empty);
 
             UICommands = new GitUICommands(module);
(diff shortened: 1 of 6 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1978,6 +1978,7 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
             _submoduleStatusProvider.Init();
             ToolStripFilters.SetBranchFilter(string.Empty, refresh: false);
+            ToolStripFilters.SetRevisionFilter(string.Empty);
```
gerhardol, line 1981: Note for other reviewers: revision filters can be set at FormBrowse startup (as command lind arguments), but SetGitModule() does not execute at startup so filters are not cleared.

gerhardol (commented):
The issue reports on the Advanced filter, paths, but the PR handles the toolbar revision filter. The revision is also relevant though.

However, I have made the path filter change in #9445, will adopt that to all items and push to this branch.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -409,7 +409,12 @@ private static ArgumentString FindRenamesAndCopiesOpts()
             return FindRenamesOpt() + findCopies;
         }
 
-        public void SetPathFilters(string path)
+        public void DisableFilters()
+        {
+            _formRevisionFilter.DisableFilters();
+        }
+
+        public void SetPathFilter(string path)
```
RussKie, line 417: We don't use "disable/clear" for other filters, instead we set those to `""`. Any reason you thought it wouldn't work in this case?

RussKie:
Thank you, I was mostly oblivious about other filter dialogs 👍

PR #9476, opened 2021-08-11
Ignore warnings reading merged remote branches

RussKie:
Fixes #9462

## Test methodology 

- manual
- unit tests

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9488, opened 2021-08-12
Execution interface cleanup

gerhardol:
Follow up to #9056, some discussions in #9476 too

## Proposed changes

* Execution: Adjust use of exit status, alloutput
Review that stderr (included in AllOutput) is not included in parsing of output data

* GetOutput only to be used where exit code is 0 and no error output

* GetOutputlines() removed, all callsites hanged on the call anyway.

Replace calls with throwOnErrorOutput:false with Execute()
Parse only StandardOutput for the replaced calls,
also where throwOnErrorOutput:false was set before.

## Test methodology 

manual

----

### DIFF
--- a/GitCommands/Git/ExecutableExtensions.cs
+++ b/GitCommands/Git/ExecutableExtensions.cs
@@ -34,7 +34,6 @@ namespace GitCommands
         /// <param name="outputEncoding">The text encoding to use when decoding bytes read from the process's standard output and standard error streams, or <c>null</c> if the default encoding is to be used.</param>
         /// <param name="cache">A <see cref="CommandCache"/> to use if command results may be cached, otherwise <c>null</c>.</param>
         /// <param name="stripAnsiEscapeCodes">A flag indicating whether ANSI escape codes should be removed from output strings.</param>
-        /// <param name="throwOnErrorOutput">A flag configuring whether to throw an exception if the exit code is not 0 or if the output to StandardError is not empty.</param>
         /// <returns>The concatenation of standard output and standard error. To receive these outputs separately, use <see cref="Execute"/> instead.</returns>
         [MustUseReturnValue("If output text is not required, use " + nameof(RunCommand) + " instead")]
         public static string GetOutput(
@@ -43,11 +42,10 @@ namespace GitCommands
             byte[]? input = null,
             Encoding? outputEncoding = null,
             CommandCache? cache = null,
-            bool stripAnsiEscapeCodes = true,
-            bool throwOnErrorOutput = true)
+            bool stripAnsiEscapeCodes = true)
         {
             return GitUI.ThreadHelper.JoinableTaskFactory.Run(
-                () => executable.GetOutputAsync(arguments, input, outputEncoding, cache, stripAnsiEscapeCodes, throwOnErrorOutput));
+                () => executable.GetOutputAsync(arguments, input, outputEncoding, cache, stripAnsiEscapeCodes));
         }
 
         /// <summary>
@@ -90,7 +88,6 @@ namespace GitCommands
         /// <param name="outputEncoding">The text encoding to use when decoding bytes read from the process's standard output and standard error streams, or <c>null</c> if the default encoding is to be used.</param>
         /// <param name="cache">A <see cref="CommandCache"/> to use if command results may be cached, otherwise <c>null</c>.</param>
         /// <param name="stripAnsiEscapeCodes">A flag indicating whether ANSI escape codes should be removed from output strings.</param>
-        /// <param name="throwOnErrorOutput">A flag configuring whether to throw an exception if the exit code is not 0 or if the output to StandardError is not empty.</param>
         /// <returns>A task that yields the concatenation of standard output and standard error. To receive these outputs separately, use <see cref="ExecuteAsync"/> instead.</returns>
         public static async Task<string> GetOutputAsync(
             this IExecutable executable,
@@ -98,8 +95,7 @@ namespace GitCommands
             byte[]? input = null,
             Encoding? outputEncoding = null,
             CommandCache? cache = null,
-            bool stripAnsiEscapeCodes = true,
-            bool throwOnErrorOutput = true)
+            bool stripAnsiEscapeCodes = true)
         {
             outputEncoding ??= _defaultOutputEncoding.Value;
 
@@ -114,7 +110,7 @@ namespace GitCommands
                 redirectInput: input is not null,
                 redirectOutput: true,
                 outputEncoding,
-                throwOnErrorOutput: throwOnErrorOutput);
+                throwOnErrorOutput: true);
             if (input is not null)
             {
                 await process.StandardInput.BaseStream.WriteAsync(input, 0, input.Length);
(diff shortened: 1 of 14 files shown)


GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -114,23 +110,21 @@ public static class ExecutableExtensions
                 redirectInput: input is not null,
                 redirectOutput: true,
                 outputEncoding,
-                throwOnErrorOutput: throwOnErrorOutput);
+                throwOnErrorOutput: true);
```
mstv, line 113: This line can be removed, too. Or do you want to emphasize it?

GitCommands/Git/GitModule.cs
```diff
@@ -2152,8 +2155,7 @@ public async Task<IReadOnlyList<Remote>> GetRemotesAsync()
 
             static IReadOnlyList<Remote> ParseRemotes(ExecutionResult result)
             {
-                IEnumerable<string> lines = result.StandardOutput.LazySplit('\n', StringSplitOptions.RemoveEmptyEntries)
-                    .Concat(result.StandardError.LazySplit('\n', StringSplitOptions.RemoveEmptyEntries));
+                IEnumerable<string> lines = result.StandardOutput.LazySplit('\n', StringSplitOptions.RemoveEmptyEntries);
```
mstv, line 2158: I guess the early exit on "not a git repository" below will not be triggered anymore. Should work anyway.

GitUI/Editor/FileViewer.cs
```diff
@@ -1500,15 +1499,7 @@ public void ResetNoncommittedSelectedLines()
                 { currentItemStaged, "--reverse --index" }
             };
 
-            string output = Module.GitExecutable.GetOutput(args, patch, throwOnErrorOutput: false);
-            if (EnvUtils.RunningOnWindows())
-            {
-                // remove file mode warnings
-                Regex regEx = new("warning: .*has type .* expected .*", RegexOptions.Compiled);
```
mstv, line 1507: ❔ Have you removed `RegexOptions.Compiled` on purpose when moving it below?

GitUI/Script/ScriptRunner.cs
```diff
@@ -134,7 +134,8 @@ private static CommandStatus RunScriptInternal(IWin32Window owner, IGitModule mo
                 command = command.Replace(NavigateToPrefix, string.Empty);
                 if (!string.IsNullOrEmpty(command))
                 {
-                    string revisionRef = new Executable(command, module.WorkingDir).GetOutputLines(argument).FirstOrDefault();
+                    ExecutionResult result = new Executable(command, module.WorkingDir).Execute(argument);
+                    string revisionRef = result.StandardOutput.Split('\n').FirstOrDefault();
```
mstv, line 138:
```suggestion
                    string revisionRef = result.StandardOutput.LazySplit('\n').FirstOrDefault();
```

mstv (approved):
Looks sensible to me. Just nits / questions.

GitUI/Script/ScriptRunner.cs
gerhardol: Not available here

GitUI/Editor/FileViewer.cs
gerhardol: Yes, that should increase the startup time a lot more than decrease the runtime.
Most regex are static already, plan to look in to that later.

This was from #9468

GitCommands/Git/ExecutableExtensions.cs
gerhardol: That was the intention, Can remove if you find it confusing.
(I have another argument to remove, then it is only visible when false.)

GitCommands/Git/GitModule.cs
gerhardol: Yes, those lines should be removed

GitCommands/Git/ExecutableExtensions.cs
mstv: I have no preference.

Plugins/GitFlow/GitFlowForm.cs
```diff
@@ -149,7 +149,7 @@ private IReadOnlyList<string> GetBranches(string typeBranch)
             GitArgumentBuilder args = new("flow") { typeBranch };
             var result = _gitUiCommands.GitModule.GitExecutable.Execute(args);
 
-            if (result.ExitCode != 0 || result.StandardOutput is null)
+            if (!result.ExitedSuccessfully || result.StandardOutput is null)
```
RussKie, line 152: Can `result.StandardOutput is null` be ever true?

GitUI/Editor/FileViewer.cs
```diff
@@ -1557,12 +1548,20 @@ private void ApplySelectedLines(bool allFile, bool reverse)
...
+        private void ProcessApplyOutput(GitArgumentBuilder args, byte[] patch, bool patchUpdateDiff = false)
         {
+            ExecutionResult result = Module.GitExecutable.Execute(args, inputWriter => inputWriter.BaseStream.Write(patch), throwOnErrorOutput: false);
+            string output = result.AllOutput.Trim();
+            if (EnvUtils.RunningOnWindows())
+            {
+                // remove file mode warnings
+                Regex regEx = new("warning: .*has type .* expected .*");
+                output = output.RemoveLines(regEx.IsMatch);
+            }
```
RussKie, line 1563: I'm confused about this one... Warnings would be in the error stream, why not ignore those completely?

GitUI/Script/ScriptRunner.cs
RussKie: We'll need to see whether we always split it, and maybe move the split into `ExecuteAsync` changing `ExecutionResult` to something similar:
```cs
    public readonly struct ExecutionResult
    {
        public string[] StandardOutput { get; }
        public string[] StandardError { get; }
```

RussKie (approved):
👍

Plugins/GitFlow/GitFlowForm.cs
gerhardol: The implementations of various `CleanString()` in `ExecutableExtensions` is not annotated to be null, so I guess that can me changed too.

GitUI/Editor/FileViewer.cs
gerhardol: AllOutput is stdout+stderr

RussKie: That's my underlying question - is stderr used for anything useful? Previously the output would contain both of stdout and stderr, so we needed this filtering. Is it still required with this change?

gerhardol: In this case, this is what we want to possibly display to the user.

See GitModule.createErrorGitItemStatus() where an "error item" is created from standard error at failures.

In #9468 stderr is used in popups at failures in the same way. In that PR I am not sure if this is too aggressive or not, so Trace will print a few more printouts (such printouts should be logged some way).

This "clean" could be lazy but the processing is a fraction of the time to patch files so it will not be noticeable.

gerhardol:
> We'll need to see whether we always split it, and maybe move the split into ExecuteAsync changing ExecutionResult to something similar:

Maybe 7 occurrences splits the output, a lot more occurrences (60?) is not.
Of course, this PR could provide a StandardOutputLines property in addition to StandardOutput.

gerhardol:
Rebased after merging of #9468, squashed the review comment.

https://github.com/gitextensions/gitextensions/pull/9488#discussion_r688944874 could be refactored
https://github.com/gitextensions/gitextensions/pull/9488#discussion_r688937769 could be changed too

gerhardol:
Merging later tonight

RussKie:
👍

PR #9489, opened 2021-08-12
Do not throw if process exits with 0

### LINKED ISSUE #9462: Deleting remote branch is broken
## Current behaviour

image

## Steps to reproduce

1. Latest master (adbb410d5b65b8d23d8bb71541179cbe573478e8)
2. Right click > delete remote branch or <kbd>DEL</kbd> on a remote branch

Yes, not sure where we regressed.

mstv:
Fixes #9462

## Proposed changes

Do not throw if process exits with 0
- Ignore `StandardError` output, which was checked since #9056
- Use it only as exception text on non-zero exit code

## Test methodology 

- existing tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build d8ddc1f6e6c72ceef18fc4b17abfefcefa46adf6 (Dirty)
- Git 2.31.1.windows.1
- .NET 5.0.7
- DPI 96dpi (no scaling)

----

### DIFF
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -63,7 +63,7 @@ namespace GitCommands
             private readonly ProcessOperation _logOperation;
             private readonly bool _redirectInput;
             private readonly bool _redirectOutput;
-            private readonly bool _throwOnErrorOutput;
+            private readonly bool _throwOnErrorExit;
 
             private MemoryStream? _emptyStream;
             private StreamReader? _emptyReader;
@@ -78,15 +78,15 @@ namespace GitCommands
                                   bool redirectOutput,
                                   Encoding? outputEncoding,
                                   bool useShellExecute,
-                                  bool throwOnErrorOutput)
+                                  bool throwOnErrorExit)
             {
                 Debug.Assert(redirectOutput == (outputEncoding is not null), "redirectOutput == (outputEncoding is not null)");
                 _redirectInput = redirectInput;
                 _redirectOutput = redirectOutput;
-                _throwOnErrorOutput = throwOnErrorOutput;
+                _throwOnErrorExit = throwOnErrorExit;
 
                 Encoding errorEncoding = outputEncoding;
-                if (throwOnErrorOutput)
+                if (throwOnErrorExit)
                 {
                     errorEncoding ??= Encoding.Default;
                 }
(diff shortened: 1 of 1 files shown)


GitCommands/Git/Executable.cs
```diff
@@ -78,15 +78,15 @@ private sealed class ProcessWrapper : IProcess
                                   bool redirectOutput,
                                   Encoding? outputEncoding,
                                   bool useShellExecute,
-                                  bool throwOnErrorOutput)
+                                  bool throwOnErrorExit)
```
mstv, line 81: Other occurrences of this function argument can be renamed throughout the project. in a follow-up. I guess it would cause too many MCs for pending PRs.

GitCommands/Git/Executable.cs
```diff
@@ -210,12 +207,12 @@ public StreamReader StandardError
...
                 get
                 {
-                    if (!_redirectOutput && !_throwOnErrorOutput)
+                    if (!_redirectOutput && !_throwOnErrorExit)
                     {
                         throw new InvalidOperationException("Process was not created with redirected output.");
                     }
 
-                    if (!_throwOnErrorOutput)
+                    if (!_throwOnErrorExit)
```
gerhardol, line 215: Should this check be removed and error be handled as stdout?
Now there is no stderr if no error exit.

gerhardol (approved):
One comment about existing implementation

GitCommands/Git/Executable.cs
mstv: If there is an error (non-zero exit code), the redirected stderr is used for the exception message.
If there is no error but text on stderr, it will not be evaluated by most of the callers anyway.
In a follow-up, I want to add the stderr output to the Git Command Log. Then I will review which functions evaluate stderr at all.
Perhaps the `StandardError` property can be removed. Or else I will `tee` this stream for logging and for evaluation by the caller.

gerhardol:
> Perhaps the `StandardError` property can be removed. Or else I will `tee` this stream for logging and for evaluation by the caller.

Stderr is used to present result to users, for one command it is parsed too. Much of that is using AllOutput.

GitCommands/Git/Executable.cs
RussKie: Until the follow up please retain the original behavior, i.e. 
```diff
-                           if (_throwOnErrorOutput)
-                           {
                                string errorOutput = _process.StandardError.ReadToEnd().Trim();
-                               if (exitCode != 0 || errorOutput.Length > 0)
+                               if (exitCode != 0 || (errorOutput.Length > 0 && _throwOnErrorExit))
                                {
                                ExternalOperationException ex
                                    = new(command: _process.StartInfo.FileName,
```

gerhardol: This is OK, stderror is evaluated separately.

GitCommands/Git/Executable.cs
gerhardol: I suggest that this is done when the PR is approved and is to be merged.

gerhardol:
Included in #9562

Issue #9464, opened 2021-08-09
ssh.exe not found after update to 3.5.2


tomas-laubr-smart-software:
## Current behaviour

Cannot git push. When pushing to repository, an error appeared. See screenshot.

## Expected behaviour

Git push works

## Steps to reproduce

1. Have properly configured git extensions. 
2. I had set SSH "Other ssh client", path `C:\Program Files\Git\usr\bin\ssh.exe`
3. Perform automatic update to latest 3.5.2 version

## Screenshots

There should be a screenshot

## Environment

- GitExtensions version:
- GIT version: 
- OS version: Windows 10/x64
- .NET version:

tomas-laubr-smart-software:
![ssh](https://user-images.githubusercontent.com/67330171/128671333-67f00330-b82d-447e-8a6e-21d9bafc509c.png)

gerhardol:
What is the value in Settings?

With the OpenSSH in Git the value uses empty value for the corresponding backing registry key gitssh in  [HKEY_USERS\SOFTWARE\GitExtensions]

tomas-laubr-smart-software:
Right after the update, it was set to `C:\Program Files (x86)\Git\usr\bin\ssh.exe` See the screenshot. Then I had to change it to `C:\Program Files\Git\usr\bin\ssh.exe` in order to make it work. That is the correct path and I believe it was set is settings before update.

RussKie:
@mstv did you work in this area lately?

mstv:
> did you work in this area lately?

@gerhardol did. I changed the handling of the Linux Tools dir (a.k.a. `AppSettings.GitBinDir`, which should be renamed.)

gerhardol:
The problem that the ssh path is changed at upgrades is not new, it has occurred for me for at least five years. Problem is hard to pinpoint it seems.

There is something in the installer that occasionally changes the paths, before 3.5.2 reverting to PuTTY. Maybe flipping the setting? Will be "solved" in master when PuTTY binaries are dropped from GE and the stage is removed in the installer. Draft in #9386 (awaiting that WiX installer works with .NET5).
Before 3.5 the path to OpenSSH ssh.exe was always written to the registry and the environment variable GIT_SSH, now it is empty. If the setting matches the expected Git path, it is "cleaned", but if an incorrect setting has been introduced it cannot be fixed.

[...]

PR #9494, opened 2021-08-13
Show ChecklistSettings also for settings without an autosolver

gerhardol:
See https://github.com/gitextensions/gitextensions/issues/9464#issuecomment-898202703

## Proposed changes

CkecklistSettings at startup was not solved if the "autosolver" was successful (see `AutoSolveAllSettings()` for settings like Git).

For settings without an autosolve like SSH the page was not shown.

## Screenshots 

The settings page as seen in the issue is shown after startup.

Note that some paths in %windir% are not checked correctly (see #9355), the user must uncheck "Check at startup" manually.

To be cherry picked to 3.5.

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -156,7 +156,7 @@ namespace GitExtensions
                         using var checklistSettingsPage = SettingsPageBase.Create<ChecklistSettingsPage>(fakePageHost);
                         if (!checklistSettingsPage.CheckSettings())
                         {
-                            if (!checkSettingsLogic.AutoSolveAllSettings())
+                            if (!checkSettingsLogic.AutoSolveAllSettings() && !checklistSettingsPage.CheckSettings())
                             {
                                 uiCommands.StartSettingsDialog();
                             }
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -156,7 +156,7 @@ namespace GitExtensions
                         using var checklistSettingsPage = SettingsPageBase.Create<ChecklistSettingsPage>(fakePageHost);
                         if (!checklistSettingsPage.CheckSettings())
                         {
-                            if (!checkSettingsLogic.AutoSolveAllSettings() && !checklistSettingsPage.CheckSettings())
+                            if (!checkSettingsLogic.AutoSolveAllSettings() || !checklistSettingsPage.CheckSettings())
                             {
                                 uiCommands.StartSettingsDialog();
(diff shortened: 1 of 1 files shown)


RussKie (approved):

pmiossec:
I have the email configuration in error because I use conditional include in git config and that GitExtensions doesn't support.

Does this change will have an effect on it and will open the pop-up at every start-up?

mstv:
> Does this change will have an effect on it and will open the pop-up at every start-up?

I don't think so. It will be shown less often, namely only if `CheckSettings` still fails after `AutoSolveAllSettings` failed.
The PR title sounds different.

mstv:
This changed code is only executed if you have `Check at startup` checked.

gerhardol:
Pushed something that works as intended too...

> > Does this change will have an effect on it and will open the pop-up at every start-up?
> 
> I don't think so. It will be shown less often, namely only if `CheckSettings` still fails after `AutoSolveAllSettings` failed.

The startup check is set after running the installer, so it will appear once and you will have to uncheck it manually.

RussKie:
I'm planning to release 3.5.3 shortly. Do we want to take this in?

gerhardol:
> I'm planning to release 3.5.3 shortly. Do we want to take this in?

Yes

RussKie:
> I have the email configuration in error because I use conditional include in git config and that GitExtensions doesn't support.
> 
> Does this change will have an effect on it and will open the pop-up at every start-up?

@pmiossec would you be able to check, if this change has any unexpected side effects with your setup?

pmiossec:
> > I have the email configuration in error because I use conditional include in git config and that GitExtensions doesn't support.
> > 
> > Does this change will have an effect on it and will open the pop-up at every start-up?
> 
> @pmiossec would you be able to check, if this change has any unexpected side effects with your setup?

I asked because I have no way to check before 2 weeks...

gerhardol:
> 
> 
> > > I have the email configuration in error because I use conditional include in git config and that GitExtensions doesn't support.
> > > Does this change will have an effect on it and will open the pop-up at every start-up?
> > 
> > 
> > @pmiossec would you be able to check, if this change has any unexpected side effects with your setup?
> 
> I asked because I have no way to check before 2 weeks...

Apart from the CheckSettings that appear after running the installer that has to be dismissed manually, this has no no negative feedback.
I will squash merge to master and release/3.5

Issue #9440, opened 2021-08-04
Freeze after exiting from commit diff view


Qrox:
## Current behaviour
If I double-click on a commit to bring up the commit diff view window, closing it causes GitExtensions to freeze. This happens for both small and large repos. The freeze may end after waiting for some time.

## Expected behaviour
It does not freeze.

## Steps to reproduce
1. Open a repository.
2. Double-click on a commit to open the diff view window.
3. Close the diff view window.
4. Freeze.

## Did this work in previous version of GitExtensions
Yes, but it was a very old version I updated from of which I no longer remember the version number.

## Environment

- Git Extensions 3.5.1.12196
- Build 9099a1e76d01dbbd6896d6b026027db0b68fba96
- Git 2.32.0.windows.2
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4390.0
- DPI 96dpi (no scaling)

## Diagnostics

After freeze ends, the command log contains this entry which takes a very long time:
```
File name:   *snip*\git.exe
Arguments:   difftool --tool-help
Working dir: *snip*
Process ID:  16568
Started at:  2021-08-05T00:56:33.0753398+08:00
UI Thread?:  False
Duration:    20424.33 ms
Exit code:   0
Call stack:  not captured
```

mstv:
`git difftool --tool-help` takes 10-20 seconds. It runs in the background. Though after closing FormDiff, it seems to cause a temporary freeze - but only once.
It can be turned off here:
![image](https://user-images.githubusercontent.com/36601201/128245666-ad18fd50-3582-487c-ab20-186718b383b2.png)

RussKie:
Are we failing to cancel a token?

mstv:
> Are we failing to cancel a token?

`git difftool --tool-help` does not fully run in the background. Some operations can be blocked until it is finished.

Waits lilke the following are too unspecific:
```
        protected override void Dispose(bool disposing)
        {
            if (disposing)
            {
                _asyncLoadCancellation.CancelCurrent();
                ThreadHelper.JoinPendingOperations(); // those run with FileAndForget
```

AdamMadrzejewski:
I was experiencing the same freezing issue, in my case, it was taking forever and the program was not responding. I was about to report it because I was able to experience it every time I opened and closed diff windows. Anyway, just before I saw there was an update, so I updated and I'm not experiencing this issue anymore, so it seems like it was fixed:

It's not happening anymore on my updated version:

```
[...]

PR #9505, opened 2021-08-15
CommitDiff: Do not Join() in Dispose()

gerhardol:
The direct cause of #9440 as it was submitted, required for a general solution too.
This is required to be able to add cancellable Git commands (at least to see the effect...).

## Proposed changes

#9405 407928a20cb395d846c6b6c07b00f102aaad199e added `ThreadHelper.JoinPendingOperations()` in Dispose for CommitInfo.
This will cause Dispose in the control to await any tasks started in the application, not just the tasks in this control.

There may be some cancellation tokens required in the existing tasks, I have not seen any obvious candidates.
For now I propose to remove the join

## Test methodology 

There are tests for this.

## Merge strategy

----

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -152,8 +152,7 @@ namespace GitUI.CommitInfo
         {
             if (disposing)
             {
-                _asyncLoadCancellation.CancelCurrent();
-                ThreadHelper.JoinPendingOperations(); // those run with FileAndForget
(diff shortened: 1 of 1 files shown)


RussKie (approved):

gerhardol:
Due to https://github.com/gitextensions/gitextensions/issues/9440#issuecomment-900731067, this should be added to 3.5 too, also if the original intention of the change is not fully handled.

Plan to squash merge later today to master and 3.5

PR #9521, opened 2021-08-19
Fix Clone URL Population

### LINKED ISSUE #9520: Clone no longer detects github ssh url
## Current behaviour

Since #9116 was included a default url provided by github of `git@github.com:gitextensions/gitextensions.git` is no longer seen as a valid url and automatically included in the clone diaglog.

Example URL: `git@github.com:gitextensions/gitextensions.git`

## Expected behaviour

When getting the clone URL from github, gitextensions automatically populates this into the clone diaglog.

## Steps to reproduce

Put `git@github.com:gitextensions/gitextensions.git` on clipboard
Open clone diaglog
URL is not automatically populated on the diaglog

Yes, in any version before #9116 was merged

## NOTE

The reason for this is that .net doesn't see `git@github.com:gitextensions/gitextensions.git` as a well formed uri string.

```
Uri.IsWellFormedUriString("git@github.com:gitextensions/gitextensions.git", UriKind.Absolute)
```
returns `False`

BlythMeister:
Fixes #9520

## Proposed changes

- Re-use the PathUtil.CanBeGitURL when checking parts of the string rather than only looking for a valid URL.

### Before

Clipboard had to contain a valid URL, not a valid Git URL.
This meant that urls like 'git@github.com:gitextensions/gitextensions.git' were not pulled from the clipboard.

### After

URL loaded from clipboard

## Test methodology

- Manual test
- 
- 

## Test environment(s) 

- GIT 2.33.0.windows.1
- Windows 10

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormClone.cs
+++ b/GitUI/CommandsDialogs/FormClone.cs
@@ -468,12 +468,11 @@ namespace GitUI.CommandsDialogs
         }
 
         /// <summary>
-        /// Check whether the given string contains one or more valid URLs and extracts
+        /// Check whether the given string contains one or more valid git URLs and extracts
         /// the first URL that exists, if any.
         /// </summary>
         /// <remarks>
-        /// Uri.IsWellFormedUriString is used instead of Uri.TryCreate as file URIs
-        /// of the form X:\\directory\\filename should not be treated as URLs.
+        /// PathUtil.CanBeGitURL is used as a standard way to detect a git URL.
         /// The first URL extracted from <paramref name="contents"/> is assigned to
         /// <paramref name="url"/>. If <paramref name="contents"/> contains more than one URL,
         /// subsequent URLs are not extracted.
--- a/GitUI/CommandsDialogs/FormClone.cs
+++ b/GitUI/CommandsDialogs/FormClone.cs
@@ -493,7 +493,7 @@ namespace GitUI.CommandsDialogs
             var parts = contents.Split(' ');
             foreach (string s in parts)
             {
-                if (Uri.IsWellFormedUriString(s, UriKind.Absolute))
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormClone.cs
```diff
@@ -493,7 +493,7 @@ private bool TryExtractUrl(string contents, out string url)
             var parts = contents.Split(' ');
             foreach (string s in parts)
             {
-                if (Uri.IsWellFormedUriString(s, UriKind.Absolute))
+                if (PathUtil.CanBeGitURL(s))
```
gerhardol, line 496: This will also include Git directories, which clone cannot handle.

BlythMeister: Fair, but the current logic doesn't even pull valid SSH URLs used by one of the biggest (if not the biggest) centralised git solutions in the world.

This is essentially a revert without losing the new functionality of pulling the git URL from a paragraph of text.

gerhardol: Not arguing bout that, just that `CanBeGitURL` is probably to general. 
If git-clone accepts Git directories, this is OK but the comment should be changed

BlythMeister: You can clone from a bare repository on a file system directory.
Which comment are you referring to?

gerhardol: Line 476

BlythMeister: Got it, thanks.  I've updated.

gerhardol (approved):

RussKie (approved):

RussKie:
Thank you

PR #9522, opened 2021-08-19
Cancellation for Executable

### LINKED ISSUE #9440: Freeze after exiting from commit diff view

### LINKED ISSUE #9501: [NBug] Value cannot be null.Parameter name: source

## Error Details
```
System.ArgumentNullException: Value cannot be null.
Parameter name: source
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at GitUI.CustomDiffMergeToolProvider.<>c__DisplayClass2_0.<<LoadCustomDiffMergeTools>b__0>d.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
```

## Additional information
I'm trying to connect GitHub repository after they have disabled the username/password access and can't get it working, neither can find any information how to setup personal access token or private key to se with GitHub. This crash happened after an attempt to fetch/pull remote repo.

## Environment
- Git Extensions 3.5.2.12435
- Build 77413699a9c25b160a6a622311529372b5d87f5b
- Git 2.32.0.windows.2
- .NET Framework 4.8.4400.0
- DPI 120dpi (125% scaling)

gerhardol:
Fixes #9440
#9440 is mostly closed in #9505, this handles a few effects
Closes #9501

## Proposed changes

Long running process like "git difftool --tool-help" are currently never cancelled, they must run to exit.
I have seen this command require over a minute at work.
After #9505, there is no major problem due to this, it is possible to open the console and close GE quickly after opening, but it could be a potential issue.

- CancellationToken for Executable
Possibility to cancel processes, not hanging until closing.
There may be other ways to do this too.

- CancellationToken handling for CustomDiffMergeTools
Runs in the background and was never cancelled before. This add cancellation token handling and also cancels the Git command.
A few other cleanups, review commit by commit

## Test methodology 

Manual

* Start Settings to clear the cache for merge/diff tools
* F12 for command log
* Open `Solve merge conflicts...` form
* Notice `mergetool --tool-help` in the list
* Close ResolveConflicts
* Notice that the time is updated immediately, no exit code is written (before the command always ran to completion

## Merge strategy

----

### DIFF
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -47,7 +47,7 @@ namespace GitCommands
         /// </summary>
         /// <param name="module">The Git module</param>
         /// <param name="delay">The delay before starting the operation</param>
-        public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay)
+        public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay, CancellationToken cancellationToken)
         {
             if (_tools is not null)
             {
@@ -56,17 +56,20 @@ namespace GitCommands
 
             // The command will compete with other resources, avoid delaying startup
             // Do not bother with cancel tokens
-            await Task.Delay(delay);
+            await Task.Delay(delay, cancellationToken);
 
-            await _mutex.WaitAsync().ConfigureAwait(false);
+            await _mutex.WaitAsync(cancellationToken).ConfigureAwait(false);
+
             try
             {
                 if (_tools is null)
                 {
                     await TaskScheduler.Default;
+                    cancellationToken.ThrowIfCancellationRequested();
+
                     var toolKey = _isDiff ? SettingKeyString.DiffToolKey : SettingKeyString.MergeToolKey;
                     var defaultTool = module.GetEffectiveSetting(toolKey);
-                    string output = module.GetCustomDiffMergeTools(_isDiff);
+                    string output = module.GetCustomDiffMergeTools(_isDiff, cancellationToken);
                     _tools = ParseCustomDiffMergeTool(output, defaultTool);
                 }
             }
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -73,6 +73,14 @@ namespace GitCommands
                     _tools = ParseCustomDiffMergeTool(output, defaultTool);
                 }
             }
+            catch
+            {
+                if (_tools is null)
+                {
+                    // Parsing has failed, just provide an empty list, no user notification
+                    _tools = Array.Empty<string>();
+                }
+            }
             finally
             {
                 _mutex.Release();
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -26,23 +26,20 @@ namespace GitCommands
         public static CustomDiffMergeToolCache MergeToolCache { get; } = new(false);
 
         /// <summary>
-        /// Clear the existing caches
+        /// Clear the existing caches (await current calculation to finsh first)
         /// </summary>
-        public void Clear()
+        public async Task ClearAsync()
         {
-            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
-            {
-                await _mutex.WaitAsync().ConfigureAwait(false);
+            await _mutex.WaitAsync().ConfigureAwait(false);
 
-                try
-                {
-                    _tools = null;
-                }
-                finally
-                {
-                    _mutex.Release();
-                }
-            }).FileAndForget();
+            try
+            {
+                _tools = null;
+            }
+            finally
+            {
+                _mutex.Release();
+            }
         }
 
         /// <summary>
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -2,6 +2,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using GitCommands.Logging;
 using GitExtUtils;
@@ -231,6 +232,12 @@ namespace GitCommands
 
             /// <inheritdoc />
             public Task<int> WaitForExitAsync() => _exitTaskCompletionSource.Task;
+
+            /// <inheritdoc />
+            public Task WaitForProcessExitAsync(CancellationToken token)
+            {
+                return _process.WaitForExitAsync(token);
+            }
 
             /// <inheritdoc />
             public int WaitForExit()
--- a/GitCommands/Submodules/SubmoduleStatusProvider.cs
+++ b/GitCommands/Submodules/SubmoduleStatusProvider.cs
@@ -293,8 +293,8 @@ namespace GitCommands.Submodules
         private async Task UpdateSubmodulesStatusAsync(GitModule module, IReadOnlyList<GitItemStatus>? gitStatus, CancellationToken cancelToken)
         {
             _previousSubmoduleUpdateTime = DateTime.Now;
-            cancelToken.ThrowIfCancellationRequested();
             await TaskScheduler.Default;
+            cancelToken.ThrowIfCancellationRequested();
 
             if (!_submoduleInfos.ContainsKey(module.WorkingDir) || _submoduleInfos[module.WorkingDir] is null)
             {
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
@@ -188,7 +188,6 @@ namespace GitUI.BranchTreePanel
                 }
 
                 await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
-                token.ThrowIfCancellationRequested();
                 TreeViewNode.ToolTipText = toolTip;
             }
         }
@@ -322,7 +321,6 @@ namespace GitUI.BranchTreePanel
             private async Task LoadNodeDetailsAsync(Nodes loadedNodes, CancellationToken token)
             {
                 await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
-                token.ThrowIfCancellationRequested();
 
                 if (TreeViewNode.TreeView is not null)
                 {
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1590,9 +1590,23 @@ namespace GitUI.CommandsDialogs
             RevisionGrid.ReloadTranslation();
             fileTree.ReloadHotkeys();
             revisionDiff.ReloadHotkeys();
-            new CustomDiffMergeToolProvider().Clear();
-            revisionDiff.LoadCustomDifftools();
-            RevisionGrid.LoadCustomDifftools();
+
+            // Clear the separate caches for diff/merge tools
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await new CustomDiffMergeToolProvider().ClearAsync(isDiff: false);
+            }).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                revisionDiff.CancelLoadCustomDifftools();
+                RevisionGrid.CancelLoadCustomDifftools();
+
+                await new CustomDiffMergeToolProvider().ClearAsync(isDiff: true);
+
+                // The tool lists are created when the forms are init, must be redone after clearing the cache
+                revisionDiff.LoadCustomDifftools();
+                RevisionGrid.LoadCustomDifftools();
+            }).FileAndForget();
 
             _dashboard?.RefreshContent();
 
--- a/GitUI/CustomDiffMergeToolProvider.cs
+++ b/GitUI/CustomDiffMergeToolProvider.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Drawing;
 using System.Linq;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
 
@@ -18,10 +19,17 @@ namespace GitUI
         /// <summary>
         /// Clear the existing caches.
         /// </summary>
-        public void Clear()
+        /// <param name="isDiff">True if diff, false if merge.</param>
+        public async Task ClearAsync(bool isDiff)
         {
-            CustomDiffMergeToolCache.DiffToolCache.Clear();
-            CustomDiffMergeToolCache.MergeToolCache.Clear();
+            if (isDiff)
+            {
+                await CustomDiffMergeToolCache.DiffToolCache.ClearAsync();
+            }
+            else
+            {
+                await CustomDiffMergeToolCache.MergeToolCache.ClearAsync();
+            }
         }
 
         /// <summary>
--- a/GitUI/CustomDiffMergeToolProvider.cs
+++ b/GitUI/CustomDiffMergeToolProvider.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Drawing;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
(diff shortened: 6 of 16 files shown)


GitCommands/Git/Executable.cs
```diff
@@ -232,6 +233,9 @@ public StreamReader StandardError
             /// <inheritdoc />
             public Task<int> WaitForExitAsync() => _exitTaskCompletionSource.Task;
 
+            /// <inheritdoc />
+            public Task WaitForExitAsync(CancellationToken token) => _process.WaitForExitAsync(token);
```
drewnoakes, line 237: Can this return the exit code, for parity with sibling methods?

Also, why is this method implemented so differently to the overload that does not support cancellation?

Plugins/GitUIPluginInterfaces/IProcess.cs
```diff
@@ -55,10 +56,17 @@ public interface IProcess : IDisposable
         /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
         Task<int> WaitForExitAsync();
 
+        /// <summary>
+        /// Instructs the process component to wait for the associated process to exit, or for the cancellationToken to be cancelled.
+        /// </summary>
+        /// <param name="token">An optional token to cancel the asynchronous operation.</param>
+        /// <returns>A task that will complete when the process has exited, cancellation has been requested, or an error occurs.</returns>
+        Task WaitForExitAsync(CancellationToken token);
```
drewnoakes, line 64: Is cancellation expected to kill the process, or just stop the wait? It's not clear from the docs here.

gerhardol: Doc copied from the library doc. I can clarify.
Cancellation throws, the statement after this never executes.

GitCommands/Git/Executable.cs
gerhardol: It is likely possible to get exit status etc from the returned task, but exceptions must be handled in the same way the exitTask handles it now.

To be able to cancel the process, the process must be involved. The other interfaces are add-on to return data from the process.
This method is really something different than the overload without cancellation. Maybe the existing method should be renamed to `WaitForExitTaskAsync` to clarify this.

Maybe there are entirely better ways to do this?

Plugins/GitUIPluginInterfaces/IProcess.cs
gerhardol: I believe it is killed - I cannot see the process running after cancellation and `_process.Exited` is invoked

GitCommands/CustomDiffMergeToolCache.cs
```diff
@@ -59,20 +56,31 @@ public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay
 
             // The command will compete with other resources, avoid delaying startup
             // Do not bother with cancel tokens
-            await Task.Delay(delay);
+            await Task.Delay(delay, cancellationToken);
+
+            await _mutex.WaitAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
gerhardol: Not needed, `WaitAsync(cancellationToken)` will throw
Unclear from documentation, but in the debugger the second line is never executed

The implementation mostly follows this pattern already, just two occurrences changed in 176ef8aec30c88773104478558b8fe723aec8908

GitCommands/CustomDiffMergeToolCache.cs
```diff
@@ -59,20 +56,31 @@ public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay
...
+                    string output = module.GetCustomDiffMergeTools(_isDiff, cancellationToken);
                     _tools = ParseCustomDiffMergeTool(output, defaultTool);
                 }
             }
+            catch
+            {
+                if (_tools is null)
+                {
+                    // Parsing has failed, just provide an empty list, no user notification
+                    _tools = new List<string>();
```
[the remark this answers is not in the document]
gerhardol: Not accepted by the compiler , the correct System.Array is referenced
```
Error	CS0426	The type name 'Empty<>' does not exist in the type 'Array'	GitCommands	C:\dev\gc\gitextensions_4\GitCommands\CustomDiffMergeToolCache.cs	81	Active
```

GitCommands/Submodules/SubmoduleStatusProvider.cs
```diff
@@ -293,8 +293,8 @@ private static string GetModuleBranch(string path, string noBranchText)
         private async Task UpdateSubmodulesStatusAsync(GitModule module, IReadOnlyList<GitItemStatus>? gitStatus, CancellationToken cancelToken)
         {
             _previousSubmoduleUpdateTime = DateTime.Now;
-            cancelToken.ThrowIfCancellationRequested();
             await TaskScheduler.Default;
+            cancelToken.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
gerhardol: Not sure

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -188,7 +188,6 @@ internal async Task SetStatusToolTipAsync(CancellationToken token)
                 }
 
                 await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
-                token.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
gerhardol: See https://github.com/gitextensions/gitextensions/pull/9522#discussion_r693329169

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1590,9 +1590,12 @@ private void OnShowSettingsClick(object sender, EventArgs e)
...
             fileTree.ReloadHotkeys();
             revisionDiff.ReloadHotkeys();
-            new CustomDiffMergeToolProvider().Clear();
-            revisionDiff.LoadCustomDifftools();
-            RevisionGrid.LoadCustomDifftools();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await new CustomDiffMergeToolProvider().ClearAsync();
+                revisionDiff.LoadCustomDifftools();
+                RevisionGrid.LoadCustomDifftools();
```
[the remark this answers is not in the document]
gerhardol: LoadCustomDifftools() has cancellations in each form.

Clear would benefit from cancellations, now Clear will first wait for the previous loading to complete.
This is not be a major problem, really just 0:10 to maybe 1:30 after start or Settings last run (and starting merge conflicts), up to twice the time to see changes in the drop down list is not a major difference.
So a revDiff/revGrid (as well as merge conflicts) `CancelLoadCustomDiffTools` is not making much difference
Maybe a comment

```
// Await previous loading of tools to finish (no cancellation) and clear the existing lists
```

GitUI/CustomDiffMergeToolProvider.cs
```diff
@@ -18,10 +20,11 @@ public class CustomDiffMergeToolProvider
         /// <summary>
         /// Clear the existing caches.
         /// </summary>
-        public void Clear()
+        public async Task ClearAsync()
         {
-            CustomDiffMergeToolCache.DiffToolCache.Clear();
-            CustomDiffMergeToolCache.MergeToolCache.Clear();
+            await CustomDiffMergeToolCache.DiffToolCache.ClearAsync();
+            await CustomDiffMergeToolCache.MergeToolCache.ClearAsync();
```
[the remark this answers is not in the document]
gerhardol: In this routine it makes no difference at all as there is virtually no execution time in the methods themselves.
What (could) take time is waiting for the previous custom loading to finish.

So in FormBrowse the load of diff tool do not have to wait on cancel of mergetools.

Will switch the order so conflict merges do not have to await difftools.

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
gerhardol: Edit: Changed to separate async

GitUI/CustomDiffMergeToolProvider.cs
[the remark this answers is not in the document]
gerhardol: Changed this, see https://github.com/gitextensions/gitextensions/pull/9522#discussion_r693333350

gerhardol:
> We need to kill the process when cancellation is requested

See https://github.com/gitextensions/gitextensions/pull/9522#discussion_r693329169

GitCommands/CustomDiffMergeToolCache.cs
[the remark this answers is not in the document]
RussKie: Is correct namespace referenced? https://docs.microsoft.com/dotnet/api/system.array.empty

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1590,9 +1590,17 @@ private void OnShowSettingsClick(object sender, EventArgs e)
...
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await new CustomDiffMergeToolProvider().ClearAsync(isDiff: false);
+            }).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await new CustomDiffMergeToolProvider().ClearAsync(isDiff: true);
+                revisionDiff.LoadCustomDifftools();
+                RevisionGrid.LoadCustomDifftools();
+            }).FileAndForget();
```
RussKie, line 1603: This deserves at least a comment why we can't combine these two.

GitCommands/CustomDiffMergeToolCache.cs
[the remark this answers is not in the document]
RussKie: I pulled down your branch, and it compiles for me
image

GitCommands/Submodules/SubmoduleStatusProvider.cs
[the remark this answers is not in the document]
RussKie: Why was it necessary to change this?
image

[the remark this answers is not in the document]
gerhardol: It is not necessary, it is a separate commit with cleanups to follow the normal patterns

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: added comment

GitCommands/CustomDiffMergeToolCache.cs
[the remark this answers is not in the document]
gerhardol: Now it worked, no clue why it could not be compiled before

> I pulled down your branch, and it compiles for me

This not changed in this PR (compare to the separate commit with cleanup I added..)

gerhardol:
Minor review comments, rebased to add cancellation after #9524 was merged https://github.com/gitextensions/gitextensions/pull/9524#pullrequestreview-735475144

RussKie (approved):
👍

gerhardol:
Rebased and squashed and updated commit messages, rebase merge tomorrow

Renamed one more exitTask method
image

GitCommands/Git/Executable.cs
```diff
@@ -230,12 +231,15 @@ public StreamReader StandardError
...
             /// <inheritdoc />
-            public Task<int> WaitForExitAsync() => _exitTaskCompletionSource.Task;
+            public Task<int> WaitForExitTaskAsync() => _exitTaskCompletionSource.Task;
 
             /// <inheritdoc />
-            public int WaitForExit()
+            public Task WaitForExitAsync(CancellationToken token) => _process.WaitForExitAsync(token);
+
+            /// <inheritdoc />
+            public int WaitForExitTask()
```
RussKie, line 240: This name is confusing and wrong. It returns an `int` and not a `task`. The same goes for the async version.

gerhardol: The method "waits for the exitTask" created when the process exits, including the extra work to add to log etc.
It is not a normal `WaitForExit()` that waits for the process task to exit.

I still think the name is better.
But what about `WaitForProcessExit()`?

drewnoakes: It is the non-async form of `WaitForExitAsync` so should be called `WaitForExit` in order to be consistent with the rest of the framework (e.g. `Read`/`ReadAsync`, `Write`/`WriteAsycn`, ...).

GitCommands/Git/Executable.cs
```diff
@@ -230,12 +231,15 @@ public StreamReader StandardError
             public void WaitForInputIdle() => _process.WaitForInputIdle();
 
             /// <inheritdoc />
-            public Task<int> WaitForExitAsync() => _exitTaskCompletionSource.Task;
+            public Task<int> WaitForExitTaskAsync() => _exitTaskCompletionSource.Task;
```
drewnoakes: Similarly this should be `WaitForExitAsync`. Async methods don't generally contain the word 'task'.

GitCommands/Git/Executable.cs
gerhardol: But there are two variants of WaitForExit() now: One that waits for the process to exit (that I added, cancellable) and the existing that waits for the additional exitTask to exit. I find the new method to follow the intention of the WaitForExit() method better, but I can rename that instead.

GitCommands/Git/Executable.cs
gerhardol: Async or not is the same.
The name is for the exitTask.

RussKie (approved):
👍 

Once on .NET 6.0 we will have this: https://devblogs.microsoft.com/dotnet/new-dotnet-6-apis-driven-by-the-developer-community/#waitasync-improvements

gerhardol:
> 👍

Was that for calling the existing exitTask as `WaitForProcessExit()` or reverting to `WaitForExit()` and rename the new cancellable method (call that `WaitForProcessExit()` instead?`)?

> Once on .NET 6.0 we will have this: https://devblogs.microsoft.com/dotnet/new-dotnet-6-apis-driven-by-the-developer-community/#waitasync-improvements

Unfortunately no change here, the additional exitTask is the problem here it seems.

RussKie:
It's almost a midnight here, and you're asking complex questions :D

Plugins/GitUIPluginInterfaces/IProcess.cs
```diff
@@ -47,18 +48,25 @@ public interface IProcess : IDisposable
...
         /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
-        Task<int> WaitForExitAsync();
+        Task<int> WaitForProcessExitAsync();
+
+        /// <summary>
+        /// Instructs the process component to wait for the associated process to exit, or for the cancellationToken to be cancelled (cancells the process).
+        /// </summary>
+        /// <param name="token">An optional token to cancel the asynchronous operation.</param>
+        /// <returns>A task that will complete when the process has exited, cancellation has been requested, or an error occurs.</returns>
+        Task WaitForExitAsync(CancellationToken token);
```
RussKie, line 64: So now we have:
```cs
/// Blocks the calling thread until the process exits, or when this object is disposed.
int WaitForProcessExit();

/// Returns a task that completes when the process exits, or when this object is disposed.
Task<int> WaitForProcessExitAsync();

/// Instructs the process component to wait for the associated process to exit, or for the cancellationToken to be cancelled (cancells the process).
Task WaitForExitAsync(CancellationToken token);
```

Maybe we can do something like this, and remove the need for the new method?
```cs
            public Task<int> WaitForProcessExitAsync(CancellationToken token)
            {
                token.Register(() => _exitTaskCompletionSource.SetCanceled(token));
                return _exitTaskCompletionSource.Task;
            }
```

gerhardol: Not really an option, exitTask is awaiten together with tasks for stdout and possibly stderr.

gerhardol: I did not get this working, just renaming

gerhardol:
Rebased, renamed `WaitForExitAsync` back and the new cancellation method to `WaitForProcessExitAsync` (was not able to merge them, probable possible but lets settle with this).

RussKie (approved):

RussKie:
We can iterate on this later, if necessary.

gerhardol:
> We can iterate on this later, if necessary.

I am considering setting the priority for the long running and background processes to BelowNormal
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.priorityclass?view=net-5.0

Probably very little difference for users though

RussKie:
How do we tell whether a process is expected to be a long running one?

gerhardol:
> How do we tell whether a process is expected to be a long running one?

By knowledge, those we know may be long (currently these two in the PR).
In the same way GE could decrease prio for `git --no-optional-locks` and maybe no console visible.

But this fit in other PRs. I may look at this after the blame/filter PRs are merged

PR #9524, opened 2021-08-20
GetRangeDiff() should be Async

gerhardol:
Follow up to #8923 

## Proposed changes

#8923 changed the "get patch" routines to be async, except for git-range-diff.
git-range-diff can take a long time (cubic to commits) and before this PR the Git command ran to completion.
On my home PC this takes about 25s, on the (newer) work PC with viruses I assume this will take several minutes.

Note: This is related to #9522, the Executable() should also have a cancellation token so the Git command is actually cancelled too.
I expect that that PR is merged after this, will add it there.

## Test methodology 

- Select commits in branches master and release/3.5
- Select range-diff (in the bottom)
- Select another file, notice that the FileViewer is updated immediately

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -9,6 +9,7 @@ using System.Net.Mail;
 using System.Security;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using GitCommands.Config;
 using GitCommands.Git;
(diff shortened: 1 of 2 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -2314,12 +2315,13 @@ public IReadOnlyList<GitStash> GetStashes(bool noLocks = false)
...
 
-        public string GetRangeDiff(
+        public async Task<string> GetRangeDiffAsync(
             ObjectId firstId,
             ObjectId secondId,
             ObjectId? firstBase,
             ObjectId? secondBase,
-            string extraDiffArguments)
+            string extraDiffArguments,
+            CancellationToken cancellationToken)
```
RussKie, line 2324: We should pass it into `_gitExecutable.ExecuteAsync`, and terminate the process if necessary. Is this something you planning to do separately?

gerhardol: This is exactly the change that should be done with #9522, in the PR that is merged last.

I believe that the Execute() cancellation need to go together with a case where it is used and I implemented CustomMergeDiffTools first.
This change is simpler and could have be done first.

RussKie (approved):

PR #9539, opened 2021-08-29
Simplify the ISettingsSource class

ivangrek:
## Proposed changes

- Prevent forwarding defaults and "decode", "encode" functions through the implementation

## Test methodology 

- unit tests
- mannualy

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7cc873184ae101937b39a788d7a7915f59e712e0
- Git 2.30.2.windows.1
- .NET 5.0.9
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -615,7 +615,7 @@ namespace GitCommands
         /// It's currently a limitation by <see cref="SettingsCache"/> that a given setting can
         /// only ever use GetString/SetString or GetEnum/SetEnum but not both. This is the case
         /// because <see cref="SettingsCache"/> caches a typed/parsed value of the setting and
-        /// crashes at <see cref="SettingsCache.TryGetValue{T}(string, T, Func{string, T}, out T)"/>
+        /// crashes at <see cref="SettingsCache.TryGetValue(string, out string?)"/>
         /// if the type that is requested doesn't match the cached type.
         /// </remarks>
         private static TEnum GetEnumViaString<TEnum>(string settingName, TEnum defaultValue)
@@ -2178,14 +2178,14 @@ namespace GitCommands
         {
         }
 
-        public override T GetValue<T>(string name, T defaultValue, Func<string, T> decode)
+        public override string? GetValue(string name)
         {
-            return AppSettings.SettingsContainer.GetValue(PathFor(name), defaultValue, decode);
+            return AppSettings.SettingsContainer.GetValue(PathFor(name));
         }
 
-        public override void SetValue<T>(string name, T value, Func<T, string?> encode)
+        public override void SetValue(string name, string? value)
         {
-            AppSettings.SettingsContainer.SetValue(PathFor(name), value, encode);
+            AppSettings.SettingsContainer.SetValue(PathFor(name), value);
         }
     }
 }
--- a/GitCommands/Settings/ConfigFileSettings.cs
+++ b/GitCommands/Settings/ConfigFileSettings.cs
@@ -70,7 +70,7 @@ namespace GitCommands.Settings
                 ConfigFileSettingsCache.Create(configPath, false, allowCache), SettingLevel.SystemWide);
         }
 
-        public string GetValue(string setting)
+        public new string GetValue(string setting)
         {
             return GetString(setting, string.Empty);
         }
@@ -80,7 +80,7 @@ namespace GitCommands.Settings
             return SettingsCache.GetValues(setting);
         }
 
-        public void SetValue(string setting, string? value)
+        public new void SetValue(string setting, string? value)
         {
             if (string.IsNullOrEmpty(value))
             {
--- a/GitCommands/Settings/RepoDistSettings.cs
+++ b/GitCommands/Settings/RepoDistSettings.cs
@@ -57,7 +57,7 @@ namespace GitCommands.Settings
 
         #endregion
 
-        public override void SetValue<T>(string name, T value, Func<T, string?> encode)
+        public override void SetValue(string name, string? value)
         {
             bool isEffectiveLevel = LowerPriority?.LowerPriority is not null;
             bool isDetachedOrGlobal = LowerPriority is null;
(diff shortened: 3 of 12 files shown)


RussKie:
MC

ivangrek:
There is no point in passing functions through the entire implementation. 
Any comments?

RussKie (approved):
I _think_ it looks ok, but I'm not intimately familiar with this code. 🤔 

@gitextensions/git-extensions-team any objections?

jbialobr:
The point was to avoid convertions to/from string everytime a setting is accessed.

vbjay:
Yep. Highly against messing with decode/encode.

ivangrek:
> The point was to avoid convertions to/from string everytime a setting is accessed.

If the "SettingsCache" class is caching, then the transformation methods must be in it, and the "ISettingsSource" class should be just a facade. Or a common interface for "SettingsCache" and "ISettingsSource".

The "SettingsCache" class is the source of the parameters.

I suggest leaving it that way for now and returning after changing the code. 🤔

ivangrek:
Responsibilities of the classes are shuffled.

jbialobr:
>If the "SettingsCache" class is caching, then the transformation methods must be in it,

That would be weird. Then it should know how to serialize to string any object of any type. SettingsCache should be responsible for caching only.

I suppose there is an issue I am not familiar with that made you start refactoring this code. Can you tell me about it?

ivangrek:
>That would be weird. Then it should know how to serialize to string any object of any type. SettingsCache should be responsible for caching only.

SettingsCache is a base class, but basically does second-level caching, while derived classes do first-level caching. This confuses the code. Also SettingsCache is a settings provider, although it is designated as a cache.

Ideally, for each type other than a string, we must transfer a separate implementation (converter) to the SettingsCache class and use the general GetValue <T> (string key) method to convert the value to the desired type. There are already ready-made classes for this logic and I would like to use them.

> I suppose there is an issue I am not familiar with that made you start refactoring this code. Can you tell me about it?

I would like to convert to use the configuration. But to do this, it is necessary to unravel the current implementation.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration?view=dotnet-plat-ext-5.0

jbialobr:
>Also SettingsCache is a settings provider, although it is designated as a cache.

Yes, that's true. 

>I would like to convert to use the configuration. But to do this, it is necessary to unravel the current implementation.

Thanks, in that case I have no further questions.

RussKie:
Yesterday I played with IOptions and IConfigurations. The current settings
xml structure can't be deserialized using the .NET 5.0 Bind
implementations, due to a bug which was fixed in .NET 6.0.
Still a serialised dictionary isn't compatible with IOptions, which implies
structured/grouped settings.

If we want to move this route we'd need to restructure our settings and use
a different persistence format (i.e. not a dictionary). And with that we
may too consider a different format (e.g. json), though main doesn't
support comments or has no equivalent to xml's CData.

I can see benefits moving to .NET configurations but it will require
sizable effort and will be a non-backwards compatible, i.e. a breaking
change.

ivangrek:
What do I have at the moment.

```
public sealed class GitExtConfigurationProvider : FileConfigurationProvider
{
    public GitExtConfigurationProvider(GitExtConfigurationSource source)
        : base(source)
    {
    }

    public override void Load(Stream stream)
    {
        var readerSettings = new XmlReaderSettings
        {
            IgnoreWhitespace = true,
            CheckCharacters = false
        };

        using var reader = XmlReader.Create(stream, readerSettings);

        var encodedNameMap = new XmlSerializableDictionary<string, string>();

        encodedNameMap.ReadXml(reader);

        Data = encodedNameMap;
    }

    public override void Set(string key, string value)
    {
        //base.Set(key, value);
    }
}

public sealed class GitExtConfigurationSource : FileConfigurationSource
{
    public override IConfigurationProvider Build(IConfigurationBuilder builder)
    {
        return new GitExtConfigurationProvider(this);
    }
}

public static class GitExtVariablesExtensions
{
    public static IConfigurationBuilder AddGitExtConfiguration(this IConfigurationBuilder configurationBuilder)
    {
        var filePath = @"***\GitExtensions.settings";
        var provider = new PhysicalFileProvider(Path.GetDirectoryName(filePath));
        var path = Path.GetFileName(filePath);

        var source = new GitExtConfigurationSource()
        {
            FileProvider = provider,
            Path = path,
            ReloadOnChange = true,
            Optional = true
        };

        return configurationBuilder.Add(source);
    }
}
```

And then...

```
var configuration = new ConfigurationBuilder()
    .AddGitExtConfiguration()
    .AddInMemoryCollection() // for test
    .Build();
    
var value = configuration
    .GetValue<string>("BuildServer.Type");
```

We may not change the file format for now.

RussKie:
Nice! How do you load/reload repo-specific settings, e.g. team-shared
(<repo>\GitExtensions.settings) and personal (<repo>\.git\
GitExtensions.settings)?

ivangrek:
For .config file.

```
var configuration = new ConfigurationBuilder()
    .AddIniFile(@"***\.git\config")
    .Build();
```

image

https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iniconfigurationextensions.addinifile?view=dotnet-plat-ext-5.0

ivangrek:
> Nice! How do you load/reload repo-specific settings, e.g. team-shared

A separate provider for each file.
FileConfigurationProvider can reload data after file changes. (`ReloadOnChange = true`)

RussKie:
We have 3 layers of configs:
1. global,
2. team-shared, and
3. personal.

The global is loaded once at the app start, and may get reloaded during the lifetime of the app, if it changes (e.g. by another instance of the app).
The other two must be loaded and unloaded whenever a user switches a repo.

So my question was whether you have considered how we would load/reload non-global configs.

ivangrek:
The file will change and everything should work automatically.
But I'm still experimenting, as the complexity of the code makes it harder to understand.

vbjay:
Make sure you handle repo switch. You need to be able to unload settings
too.

ivangrek:
Go? 😃

RussKie:
I'm going to apply the change locally and test it.

PR #9551, opened 2021-08-31
fix: Show current branch resets when switch repos

### LINKED ISSUE #9517: After update "View - Show current branch only" no longer filter out branches and reset to "View - Show all branches" every time I switch repository
## Current behavior

Today I updated to a newer version of GIT Extensions and it stopped filtering branches by using the "View - Show current branch only" option.
Also when I switch repositories from the toolbar it resets to "View - Show all branches".

## Steps to reproduce

1. Open Git extensions having 2 repositories and 2 branches in some of in
2. Open repository with 2 branches
3. Go to View and select "Show current branch only"
Expected: Only the current branch head is shown.
Actual: Multiple head branches are shown
4. In the toolbar select second repository
5. In the toolbar select the first repository again
Expected: "View - Show current branch only" option is selected
Actual: It got reset to "View - Show all branches"

## Screenshots

image

## Did this work in the previous version of GitExtensions

Yes, it was. I'm not sure, I wasn't updating it for a few months.

## Environment

- Git Extensions `3.5.3.12551`
- Build `00604a59a673142a3d34a22aa657466a922977f3`
- Git 2.31.1.windows.1
- .NET Framework 4.8.4300.0
- DPI 120dpi (125% scaling)

RussKie:
Fixes #9517

A regression introduced in #8489 that resets filters when repos are switched.
However it didn't account for the "Show current branch" setting.

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2000,7 +2000,12 @@ namespace GitUI.CommandsDialogs
             RevisionGrid.InvalidateCount();
             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
             _submoduleStatusProvider.Init();
-            _filterBranchHelper.SetBranchFilter(string.Empty, refresh: false);
+
+            // Reset, if we're filtering but not showing the current branch (check SetShowBranches implementation).
+            if (AppSettings.BranchFilterEnabled && !AppSettings.ShowCurrentBranchOnly)
+            {
+                _filterBranchHelper.SetBranchFilter(string.Empty, refresh: false);
+            }
 
             UICommands = new GitUICommands(module);
             if (Module.IsValidGitWorkingDir())


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2000,7 +2000,12 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
             RevisionGrid.InvalidateCount();
             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
             _submoduleStatusProvider.Init();
-            _filterBranchHelper.SetBranchFilter(string.Empty, refresh: false);
+
+            // Reset, if we're filtering but not showing the current branch (check SetShowBranches implementation).
+            if (AppSettings.BranchFilterEnabled && !AppSettings.ShowCurrentBranchOnly)
```
gerhardol, line 2005: Should not this be resetting all but Show Current?

gerhardol: OK

gerhardol (approved):

PR #9553, opened 2021-09-02
fix: Show current branch resets when switch repos

RussKie:
Fixes #9517
Relates to https://github.com/gitextensions/gitextensions/pull/9551

A regression introduced in #8489 that resets filters when repos are switched.
However it didn't account for the "Show current branch" setting.

PR #9555, opened 2021-09-02
Allow hiding stash references on revision grid

### LINKED ISSUE #9554: Option to hide stash branch on revision grid
## Feature description
Add an option to the view menu, to allow hiding of the stash branch on revision grid.

## Environment
- Git Extensions 3.5.3.12551
- Build 00604a59a673142a3d34a22aa657466a922977f3
- Git 2.33.0.windows.2
- .NET Framework 4.8.4400.0
- DPI 96dpi (no scaling)

Kableado:
Fixes #9554

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -194,6 +194,10 @@ namespace GitCommands
                                     !string.IsNullOrWhiteSpace(branchFilter) && !IsSimpleBranchFilter(branchFilter),
                                     "--branches=" + branchFilter
                                 },
+                                {
+                                    refFilterOptions.HasFlag(RefFilterOptions.Branches) && string.IsNullOrWhiteSpace(branchFilter),
+                                    "--branches"
+                                },
                                 { refFilterOptions.HasFlag(RefFilterOptions.Remotes), "--remotes" },
                                 { refFilterOptions.HasFlag(RefFilterOptions.Tags), "--tags" },
                             }.ToString()
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1104,6 +1104,12 @@ namespace GitCommands
             set => SetBool("showReflogReferences", value);
         }
 
+        public static bool ShowLatestStash
+        {
+            get => GetBool("showLatestStash", true);
+            set => SetBool("showLatestStash", value);
+        }
+
         public static bool ShowSuperprojectTags
         {
             get => GetBool("showSuperprojectTags", false);
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.Command.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.Command.cs
@@ -38,7 +38,8 @@
             GoToMergeBaseCommit = 31,
             OpenCommitsWithDifftool = 32,
             ToggleBetweenArtificialAndHeadCommits = 33,
-            ShowReflogReferences = 34
+            ShowReflogReferences = 34,
+            ShowLatestStash = 35,
         }
     }
 }
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -8,7 +8,6 @@ using System.Linq;
 using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
-using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
@@ -1019,6 +1018,11 @@ namespace GitUI
                 if (AppSettings.ShowReflogReferences)
                 {
                     refFilterOptions |= RefFilterOptions.Reflogs;
+                }
+
+                if (!AppSettings.ShowLatestStash)
+                {
+                    refFilterOptions &= ~RefFilterOptions.Stashes;
                 }
 
                 RefFilterOptions = refFilterOptions;
@@ -2274,6 +2278,12 @@ namespace GitUI
 
             ForceRefreshRevisions();
             OnFilterChanged();
+        }
+
+        internal void ToggleShowLatestStash()
+        {
+            AppSettings.ShowLatestStash = !AppSettings.ShowLatestStash;
+            ForceRefreshRevisions();
         }
 
         internal void ToggleShowSuperprojectTags()
--- a/GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
@@ -224,6 +224,14 @@ namespace GitUI.UserControls.RevisionGrid
                     ExecuteAction = () => _revisionGrid.ToggleShowReflogReferences(),
                     IsCheckedFunc = () => AppSettings.ShowReflogReferences
                 },
+                new MenuCommand
+                {
+                    Name = "ShowLatestStash",
+                    Text = "Show latest stash",
+                    ShortcutKeyDisplayString = GetShortcutKeyDisplayStringFromRevisionGridIfAvailable(RevisionGridControl.Command.ShowLatestStash),
+                    ExecuteAction = () => _revisionGrid.ToggleShowLatestStash(),
+                    IsCheckedFunc = () => AppSettings.ShowLatestStash
+                },
 
                 MenuCommand.CreateSeparator(),
 
--- a/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
+++ b/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
@@ -197,6 +197,109 @@ namespace GitExtensions.UITests.CommandsDialogs
                 });
         }
 
+        [Test]
+        public void ShowLatestStash_starting_disabled_should_filter_as_expected()
+        {
+            using ReferenceRepository referenceRepository = new();
+            GitUICommands commands = new(referenceRepository.Module);
+
+            referenceRepository.CreateCommit("Commit1");
+            referenceRepository.Stash("Stash1");
+            referenceRepository.CreateCommit("Commit2");
+            referenceRepository.Stash("Stash2");
+            referenceRepository.CreateCommit("Commit3");
+
+            bool showLatestStash = AppSettings.ShowLatestStash;
+            bool revisionGraphShowArtificialCommits = AppSettings.RevisionGraphShowArtificialCommits;
+
+            AppSettings.ShowLatestStash = false;
+            AppSettings.RevisionGraphShowArtificialCommits = false;
+            RunFormTest(
+                form =>
+                {
+                    try
+                    {
+                        // 1. Check with ShowLatestStash disabled
+                        Console.WriteLine("Scenario 1: set 'Show latest stash' to false");
+                        //// Assert
+                        AppSettings.ShowLatestStash.Should().BeFalse();
+                        form.GetTestAccessor().RevisionGrid.GetTestAccessor().VisibleRevisionCount.Should().Be(4);
+
+                        // 2. Change ShowLatestStash to enabled
+                        Console.WriteLine("Scenario 1: change 'Show latest stash' to enabled");
+                        form.GetTestAccessor().RevisionGrid.ToggleShowLatestStash();
+                        //// Wait for reload to complete
+                        ProcessUntil(() => form.GetTestAccessor().RevisionGrid.GetTestAccessor().IsRefreshingRevisions.ToString(), false.ToString());
+                        //// Assert
+                        AppSettings.ShowLatestStash.Should().BeTrue();
+                        form.GetTestAccessor().RevisionGrid.GetTestAccessor().VisibleRevisionCount.Should().Be(6);
+                    }
+                    finally
+                    {
+                        AppSettings.ShowLatestStash = showLatestStash;
+                        AppSettings.RevisionGraphShowArtificialCommits = revisionGraphShowArtificialCommits;
+                    }
+                },
+                commands);
+        }
+
+        [Test]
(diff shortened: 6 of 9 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2264,6 +2268,14 @@ public void ToggleShowReflogReferences()
             OnFilterChanged();
         }
 
+        public void ToggleHideStashBranch()
+        {
+            AppSettings.HideStashBranch = !AppSettings.HideStashBranch;
+
+            ForceRefreshRevisions();
+            OnFilterChanged();
```
RussKie, line 2276: This is not necessary as there's no matching button on the filter's toolbar.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2264,6 +2268,14 @@ public void ToggleShowReflogReferences()
             OnFilterChanged();
         }
 
+        public void ToggleHideStashBranch()
```
RussKie, line 2271: This only needs to be `public` if it is exposed in `IRevisionGridFilter` interface
```suggestion
        internal void ToggleHideStashBranch()
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.Command.cs
```diff
@@ -38,7 +38,8 @@ internal enum Command
             GoToMergeBaseCommit = 31,
             OpenCommitsWithDifftool = 32,
             ToggleBetweenArtificialAndHeadCommits = 33,
-            ShowReflogReferences = 34
+            ShowReflogReferences = 34,
+            HideStashBranch = 35,
```
gerhardol, line 42: Positive, Stash is not branch

```suggestion
            ShowStashCommit = 35,
```

GitCommands/Settings/AppSettings.cs
```diff
@@ -1097,6 +1097,12 @@ public static bool ShowReflogReferences
             set => SetBool("showReflogReferences", value);
         }
 
+        public static bool HideStashBranch
+        {
+            get => GetBool("hideStashBranch", false);
+            set => SetBool("hideStashBranch", value);
+        }
```
RussKie, line 1104:
```suggestion
        public static bool ShowStashReferences
        {
            get => GetBool("ShowStashReferences", true);
            set => SetBool("ShowStashReferences", value);
        }
```

GitUI/Translation/Dutch.xlf
```diff
@@ -11614,7 +11614,11 @@ Zie de wijzigingen in het commit form.</target>
       </trans-unit>
       <trans-unit id="ShowReflogReferences.Text">
         <source>Show reflog references</source>
-        
+
```
gerhardol, line 11617: Translations are updated from Transifex, only English is the PR

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
```diff
@@ -224,6 +224,14 @@ private IReadOnlyList<MenuCommand> CreateViewMenuCommands()
                     ExecuteAction = () => _revisionGrid.ToggleShowReflogReferences(),
                     IsCheckedFunc = () => AppSettings.ShowReflogReferences
                 },
+                new MenuCommand
+                {
+                    Name = "HideStashBranch",
+                    Text = "Hide stash branch",
```
RussKie, line 230: The text needs to be "Show..." to be consistent with the rest.
Somehow "stash branch" feels weird to me... Should it be "Show stash references"?

RussKie:
Thank you.

Please revert all xlf changes, and follow the translation procedures.

gerhardol (commented):
This only applies to the most recent Stash (commit), others are not visible anyway, may need to be clarified.
(reflogs shows all stashes too, anonymous though).
I would like to see an option to show stashes without all reflogs. (This requires a separate listing of stashes in addition to the standard git-log which will slow down showing the grid (no longer shown when Git returns the first commit), or that stash "labels" are added similar to superproject branches).

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
gerhardol: https://github.com/gitextensions/gitextensions/pull/9555#pullrequestreview-745050308
and
https://github.com/gitextensions/gitextensions/pull/9555#discussion_r701074711

GitUI/UserControls/RevisionGrid/RevisionGridControl.Command.cs
gerhardol: Alternative - this is only one stash
```suggestion
            ShowMostRecentStash = 35,
```

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
Kableado: I agree, it felt weird putting "hide" when all was "show". I is changed to the suggestion.

GitUI/UserControls/RevisionGrid/RevisionGridControl.Command.cs
Kableado: I have changed it to the suggested "ShowBranchReferences"

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
```diff
@@ -224,6 +224,14 @@ private IReadOnlyList<MenuCommand> CreateViewMenuCommands()
                     ExecuteAction = () => _revisionGrid.ToggleShowReflogReferences(),
                     IsCheckedFunc = () => AppSettings.ShowReflogReferences
                 },
+                new MenuCommand
+                {
+                    Name = "ShowStashReferences",
+                    Text = "Show stash references",
```
gerhardol, line 230: This is a single stash, reference**s** implies that all stashes are shown.

gerhardol: As before, I do not believe "stash reference" is something that is Git lingo, but "stash" or "stash entry" is maybe not good enough. 
"Show latest stash" or "Show most recent stash"?

Kableado: "Show stash references" was propossed by @RussKie https://github.com/gitextensions/gitextensions/pull/9555#discussion_r701080238 . And I agree with "stash reference" because it is handled by git and gitextensions as a reference named "stash".

But, this is not a hill I would like to die on, ¿What label should I use?

RussKie:
I dont use stashes at all, so I'm not all that familiar with associated
lingo.

RussKie:
Please add a test to verify we're toggling shashes correctly. Something along the lines of:
1. create a repo, 
2. add few commits,
3. add few stashes,
4. show stashes - assert,
5. hide stashes - assert.

You can see an example of how we assert different filters in https://github.com/gitextensions/gitextensions/pull/9553.

Kableado:
I wasn't sure how to test this feature. Thanks for the tip!
https://github.com/gitextensions/gitextensions/pull/9555#issuecomment-912926285

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
```diff
@@ -63,14 +63,8 @@ public void OneTimeTearDown()
...
-            if (_referenceRepository is null)
-            {
-                _referenceRepository = new ReferenceRepository();
-            }
-            else
-            {
-                _referenceRepository.Reset();
-            }
+            // we need to completely reset each time
+            _referenceRepository = new ReferenceRepository();
```
gerhardol, line 67: This slows downs tests, but only some tenths of a second locally, so it is OK.

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
```diff
@@ -224,6 +224,14 @@ private IReadOnlyList<MenuCommand> CreateViewMenuCommands()
                     ExecuteAction = () => _revisionGrid.ToggleShowReflogReferences(),
                     IsCheckedFunc = () => AppSettings.ShowReflogReferences
                 },
+                new MenuCommand
+                {
+                    Name = "ShowStashReference",
+                    Text = "Show stash reference",
```
gerhardol, line 230: I still want to change this...

gerhardol (commented):
I believe this is ready, but want to change the name.
Will not block a merge, but also not approve.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
Kableado: The tests on that class needs clean repositories, and this is the way to reset it that I found on the code (FormBrowse.LeftPanelTests.cs:72 and FormBrowse.LeftPanelTests.Remotes.cs:61).  Is there a better way?

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
Kableado: I'm not against changing it, but. To what?

gerhardol: https://github.com/gitextensions/gitextensions/pull/9555#discussion_r701216184

I suggest `Show latest stash` after reading the man page again.
@RussKie as a non user of stash was too quick to suggest "stash reference".

Then we are done with this...

GitCommands/RevisionReader.cs
```diff
@@ -194,6 +194,11 @@ public sealed class RevisionReader : IDisposable
                                     !string.IsNullOrWhiteSpace(branchFilter) && !IsSimpleBranchFilter(branchFilter),
                                     "--branches=" + branchFilter
                                 },
+                                {
+                                    refFilterOptions.HasFlag(RefFilterOptions.Branches) &&
+                                    string.IsNullOrWhiteSpace(branchFilter),
```
RussKie, line 199: Please keep on the same line
```suggestion
                                    refFilterOptions.HasFlag(RefFilterOptions.Branches) && string.IsNullOrWhiteSpace(branchFilter),
```

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
RussKie: ...then we should also rename all other things, like settings name, property names, etc.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
RussKie: `_referenceRepository` must be disposed of to clean up after itself.

> The tests on that class needs clean repositories

This is precisely what `_referenceRepository.Reset()` is doing, it performs `reset --hard` on the very first commit.
https://github.com/gitextensions/gitextensions/blob/4fd3843999ab4b25b090ce4eea364c15e180ece2/UnitTests/CommonTestUtils/ReferenceRepository.cs#L92-L112

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
```diff
@@ -197,6 +191,77 @@ public void Filters_should_behave_as_expected()
...
+                        form.GetTestAccessor().RevisionGrid.GetTestAccessor().VisibleRevisionCount.Should().Be(6);
+                    }
+                    finally
+                    {
+                        AppSettings.ShowStashReference = showStashReference;
+                        AppSettings.RevisionGraphShowArtificialCommits = revisionGraphShowArtificialCommits;
+                    }
+                });
+
+            AppSettings.ShowStashReference = true;
```
RussKie, line 235: Please break this into own test

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
Kableado: It is changed to ```Show latest stash```.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
Kableado: The tests needs a completelly clean repository, without other branches or stashes. ```reset --hard``` only cleanups the current branch.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs
RussKie: In this case we must dispose `_referenceRepository` in `TearDown`, which is run for every test. https://docs.nunit.org/articles/nunit/technical-notes/usage/SetUp-and-TearDown.html

RussKie (commented):
We need further work on the test side.

RussKie (approved):

RussKie:
We also need your sign off in the contributors.txt

RussKie:
@gerhardol merge when ready

gerhardol:
Note: When filtering branches, this option has no effect (stashes are not shown).
This is slightly similar to other options that are ignored when reflog is shown, so I suggest we leave it at this.
Squash merge.

gerhardol:
With the latest updates, the test `RepoObjectTree_should_order_should_display_inactive_after_active_nodes` fails.
The test works locally, I do not see why this PR would affect them.
I have tried to restart the tests several times, same result.
So no merge yet.

RussKie:
> With the latest updates, the test `RepoObjectTree_should_order_should_display_inactive_after_active_nodes` fails.
> The test works locally, I do not see why this PR would affect them.
> I have tried to restart the tests several times, same result.

Should be fixed by https://github.com/gitextensions/gitextensions/pull/9607

RussKie:
Thank you

Issue #9373, opened 2021-07-13
[NBug] Already transitioned to the Completed state.


ssokratov:
:warning: The sections below must be filled in and this text must be removed or the issue will be closed.

## Error Details
```
System.InvalidOperationException: Already transitioned to the Completed state.
   at Microsoft.Verify.FailOperation(String message, Object[] args)
   at Microsoft.VisualStudio.Threading.AsyncQueue`1.Enqueue(T value)
   at GitUI.UserControls.RevisionGrid.RevisionDataGridView.UpdateVisibleRowRange()
   at GitUI.UserControls.RevisionGrid.RevisionDataGridView.Add(GitRevision revision, RevisionNodeFlags types)
   at GitUI.RevisionGridControl.<>c__DisplayClass165_0.<ForceRefreshRevisions>g__OnRevisionRead|0(GitRevision revision)
   at System.Reactive.AnonymousSafeObserver`1.OnNext(T value)
   at System.Reactive.Sink`1.ForwardOnNext(TTarget value)
   at System.Reactive.ObserveOnObserverLongRunning`1.Drain()
   at System.Reactive.ObserveOnObserverLongRunning`1.<>c.<.cctor>b__17_0(ObserveOnObserverLongRunning`1 self, ICancelable cancelable)
   at System.Reactive.Concurrency.NewThreadScheduler.<>c__DisplayClass7_0`1.<ScheduleLongRunning>b__0()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()

```

## Additional information
Just pulled the remote brunch

## Environment
- Git Extensions 3.5.1.12196
- Build 9099a1e76d01dbbd6896d6b026027db0b68fba96
- Git 2.25.0.windows.1 (recommended: 2.30.0 or later)
- Microsoft Windows NT 6.2.9200.0
- .NET Framework 4.8.4390.0
- DPI 96dpi (no scaling)

gerhardol:
_backgroundQueue.Enqueue() failed
[...]

PR #9557, opened 2021-09-04
Reduce amount of TaskCanceledException

RussKie:
We're endlessly spamming TaskCanceledException exceptions when `_backgroundQueue` is empty while executing
```cs
    await _backgroundQueue.DequeueAsync(linkedCancellation.Token)
```
Add a check for an empty queue to avoid unnecessary attempts to dequeue, and as a result unnecessary timeout exceptions.

### Before

image

### After

Try to guess at which point I added the check 😄 

image

### DIFF
--- a/GitUI/CommitInfo/CommitInfo.cs
+++ b/GitUI/CommitInfo/CommitInfo.cs
@@ -14,6 +14,7 @@ using GitCommands;
 using GitCommands.ExternalLinks;
 using GitCommands.Git;
 using GitCommands.Remotes;
+using GitCommands.Settings;
 using GitExtUtils;
 using GitExtUtils.GitUI;
 using GitExtUtils.GitUI.Theming;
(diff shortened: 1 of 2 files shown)


gerhardol (approved):
See also #8809 that eliminates the background queue.
There are some crashes reported there too, like #9373

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -310,7 +311,7 @@ private void ReloadCommitInfo()
 
             if (_revision is not null && !_revision.IsArtificial)
             {
-                StartAsyncDataLoad();
+                StartAsyncDataLoad(Module.EffectiveSettings);
```
RussKie, line 314: We need to take a copy of effective settings, otherwise, as tests show, a `Module` can get disassociated from a control (e.g. get disposed) before the async operation gets complete. By making a local copy we can complete the async operation without any issues.

RussKie:
@msftbot merge in 48 hours

gerhardol (commented):
+1
Very brief use, seem OK

PR #9566, opened 2021-09-11
Fix leaking brushes

### LINKED ISSUE #9429: [NBug] An unexpected internal error occurred in the Win32 call: 80...
## Current behaviour

Program crashed sometime during the night.

## Expected behaviour

Shouldn't crash.

## Steps to reproduce

Leave the program running overnight.

## Error Details
```
System.ComponentModel.Win32Exception: An unexpected internal error occurred in the Win32 call: 8007007E ---> System.IO.FileNotFoundException: The specified module could not be found. (Exception from HRESULT: 0x8007007E)

   at Microsoft.WindowsAPICodePack.Dialogs.NativeTaskDialog.NativeShow()
   at Microsoft.WindowsAPICodePack.Dialogs.TaskDialog.ShowCore()
   at GitUI.NBugReports.BugReportInvoker.Report(Exception exception, Boolean isTerminating)
   at GitExtensions.Program.<>c.<Main>b__1_3(Object s, ThreadExceptionEventArgs e)
   at System.Windows.Forms.Application.ThreadContext.OnThreadException(Exception t)
   at System.Windows.Forms.Control.WndProcException(Exception e)
   at System.Windows.Forms.Control.ControlNativeWindow.OnThreadException(Exception e)

```

## Additional information
Left the program running overnight.

## Environment
- Git Extensions 3.5.1.12196
- Git 2.31.1.windows.1
- .NET Framework 4.8.4390.0
- DPI 96dpi (no scaling)

NikolayXHD:
Fixes #9429

## Proposed changes

1. Return brush wrapper in situations when brush consumer is not certain whether it should dispose the brush.
  wrapper.Dispose method will not touch the underlying brush if it is system brush.
2. Cache the native brushes returned by Win32 GetSysColorBrush override
3. Render text input backround explicitly, because 2. somehow breaks default system rendering.

## Test methodology

- run GitExtensions
- scroll revision grid back and forth, click on different revisions
- observe GDI objects count using [GDIView](https://www.nirsoft.net/utils/gdi_handles.html) (thank you @pmiossec)

## Test environment(s)

Windows 10

## Merge strategy

- Rebase merge

----

### DIFF
--- a/GitExtUtils/GitUI/Theming/BrushScope.cs
+++ b/GitExtUtils/GitUI/Theming/BrushScope.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Drawing;
+
+namespace GitExtUtils.GitUI.Theming
+{
+    public class BrushScope : IDisposable
+    {
+        private readonly bool _isSystemBrush;
+
+        public static BrushScope ForSystemBrush(Brush brush) =>
+            new(brush, isSystemBrush: true);
+
+        public static BrushScope ForRegularBrush(Brush brush) =>
+            new(brush, isSystemBrush: false);
+
+        private BrushScope(Brush brush, bool isSystemBrush)
+        {
+            Brush = brush;
+            _isSystemBrush = isSystemBrush;
+        }
+
+        public Brush Brush { get; }
+
+        public void Dispose()
+        {
+            if (!_isSystemBrush)
+            {
+                Brush.Dispose();
+            }
+
+            GC.SuppressFinalize(this);
+        }
+    }
+}
--- a/GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
+++ b/GitExtUtils/GitUI/Theming/TabControlPaintContext.cs
@@ -106,7 +106,10 @@ namespace GitExtUtils.GitUI.Theming
         {
             using var borderPen = CreateBorderPen();
             var outerRect = GetOuterTabRect(index);
-            _graphics.FillRectangle(GetBackgroundBrush(index), outerRect);
+            using (var backgroundBrush = CreateBackgroundBrusScope(index))
+            {
+                _graphics.FillRectangle(backgroundBrush.Brush, outerRect);
+            }
 
             List<Point> points = new(4);
             if (index <= _selectedIndex)
@@ -262,8 +265,12 @@ namespace GitExtUtils.GitUI.Theming
                 return;
             }
 
-            _graphics.FillRectangle(GetBackgroundBrush(_selectedIndex), pageRect);
-            using var borderPen = CreateBorderPen();
+            using (var backgroundBrush = CreateBackgroundBrusScope(_selectedIndex))
+            {
+                _graphics.FillRectangle(backgroundBrush.Brush, pageRect);
+            }
+
+            using (var borderPen = CreateBorderPen())
             {
                 _graphics.DrawRectangle(borderPen, pageRect);
             }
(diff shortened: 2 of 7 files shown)


GitExtUtils/GitUI/Theming/BrushContainer.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Drawing;
+
+namespace GitExtUtils.GitUI.Theming
+{
+    public class BrushContainer : IDisposable
+    {
+        private readonly bool _isSystemBrush;
+
+        public static BrushContainer ForSystemBrush(Brush brush) =>
+            new BrushContainer(brush, isSystemBrush: true);
```
gerhardol, line 11: nit pick, IDE0090 (general)
A general conversion is done in master, suggest this is done here to give cleaner merge
Style preferences should be documented...
```suggestion
            new(brush, isSystemBrush: true);
```

gerhardol (commented):
+1
Seem fine to me, I have only briefly run this.

pmiossec:
>Return brush wrapper in situations when brush consumer is not certain whether it should dispose the brush.
wrapper.Dispose method will not touch the underlying brush if it is system brush.

Clever idea. I have checked with the portable version and from what I see with GdiView, it is working great and fix the issue.

Code seems good. I will give it another look later.

I was just wondering. On the project, do we prefer:

```
using(var foo = new Brush())
{}
```

or 

```
using var foo = new Brush();
```

gerhardol:
> I was just wondering. On the project, do we prefer:
> 
> ```
> using(var foo = new Brush())
> {}
> ```
> 
> or
> 
> ```
> using var foo = new Brush();
> ```

The latter generally, but the case I checked here the first was still required (null check).

RussKie:
I've started to review, but got caught up with work. Will complete this week.

GitUI/Theming/SystemBrushOverrides.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    internal class SystemBrushOverrides
+    {
+        private readonly Dictionary<int, HandleRef> _cache = new();
+
+        public IntPtr GetBrush(int colorref)
+        {
+            if (!_cache.TryGetValue(colorref, out var handle))
+            {
+                var hbrush = NativeMethods.CreateSolidBrush(colorref);
```
RussKie, line 15: ❗ `HBRUSH` must be disposed of when we no longer need it (e.g. when we uninstall the hook), otherwise we'll leak native handles.
https://docs.microsoft.com/windows/win32/api/wingdi/nf-wingdi-createsolidbrush#remarks

This is how we've wrapped (some) native handled in dotnet/winforms, may be you find it useful:
* https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Primitives/src/Interop/Gdi32/Interop.HBRUSH.cs
* https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Primitives/src/Interop/Gdi32/Interop.CreateBrushScope.cs

RussKie (changes requested):
👍 with some comments

RussKie:
@NikolayXHD I've made few tweaks, please have a look.

NikolayXHD:
> @NikolayXHD I've made few tweaks, please have a look.

ok, meged the changes changes to respective commits.

RussKie (approved):

RussKie:
Thank you!

gerhardol:
Thanks!
Will this be merged to master too, even if themes are not possible yet with .NET5?
How should we remember this otherwise?

NikolayXHD:
> Will this be merged to master too, even if themes are not possible yet with .NET5?

Sure thing :)

PR #9412, opened 2021-07-24
Allow to select and delete multiple branches in the revision grid

RussKie:
## Proposed changes

Allow to select and delete multiple local and remote branches in the revision grid.

If there is only a single commit selected - the logic is unchanged.
If there are multiple commits selected then the logic for deciding which dialog to invoke is as follows:
- if all selected commits contain a remote branch - show `FormDeleteRemoteBranch`,
- else show `FormDeleteBranch`

### After

multi-delete-local
multi-delete-remote

## Test methodology 

- TBD
- 
- 

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #8957, opened 2021-03-08
Dialogs layout alignment - `FormDeleteRemoteBranch`

RussKie:
Relates to #6183

## Proposed changes

* Apply the same design as for other updated dialogs
* Remove verbose "help" text, replaced with [?] button that links to the docs
* Set focus to the branches edit control

### Before

image

### After

image

## Test methodology 

- manual

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9585, opened 2021-09-22
SHIFT+DEL = quick delete branch

LukeOwlclaw:
Currently it is a hassle to remove a number of obsolete branches.
There is so much clicking and confirming necessary.
With this PR you simply press SHIFT+DEL it will:
  - choose the first available branch for the current commit for deletion
  - enable the "delete remote branch" checkbox
  - suppress (quietly confirm) confirmation dialogs

Despite not having signed the "The Developer Certificate of Origin", 
I'd like to share my changes. Use them as you may see fit. Thus:

Hereby, I declare that I claim no rights on this commit.
Instead I donate my changes free of charge and free of demands of any kind.
Any developer may use, rebase, or merge these changes.

### DIFF
--- a/GitUI/CommandsDialogs/FormDeleteBranch.cs
+++ b/GitUI/CommandsDialogs/FormDeleteBranch.cs
@@ -21,6 +21,7 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _useReflogHint = new("Did you know you can use reflog to restore deleted branches?");
 
         private readonly IEnumerable<string> _defaultBranches;
+        private readonly bool _forceDeleteUnmergedBranch;
         private readonly HashSet<string> _mergedBranches = new();
         private string? _currentBranch;
 
@@ -32,11 +33,11 @@ namespace GitUI.CommandsDialogs
             InitializeComponent();
         }
 
-        public FormDeleteBranch(GitUICommands commands, IEnumerable<string> defaultBranches)
+        public FormDeleteBranch(GitUICommands commands, IEnumerable<string> defaultBranches, bool forceDeleteUnmergedBranch = false)
             : base(commands, enablePositionRestore: false)
         {
             _defaultBranches = defaultBranches;
-
+            _forceDeleteUnmergedBranch = forceDeleteUnmergedBranch;
             InitializeComponent();
 
             // work-around the designer bug that can't add controls to FlowLayoutPanel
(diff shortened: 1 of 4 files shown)


RussKie:
Have a look at #8956, it allows suppressing nagging prompts, and pretty
much streamlines the deletion process to DEL (to open the dialog) + Enter
(to delete).
The delete remote branch dialog has undergone a similar clean up in #8957,
though we still require the user to confirm the deletion.

LukeOwlclaw:
Thanks for your response. Yes, I have seen the option to suppress the prompt. However, enabling the checkbox "Delete branch(es) from remote repository" is still troublesome (TAB, TAB + SPACE; even with #8957 being merged). And I personally also dislike "Delete remote branches" confirmation (and am wondering why it can be suppressed for local branches but not for remote branches). That is why I added this shortcut to circumvent all safety nets (just like when you use SHIFT+DEL in Windows, it bypasses the recycle bin and the file is gone for good).

However, feel free to reject my idea. But I really think that it helps developers to work more efficiently with GitExtensions.

RussKie:
> am wondering why it can be suppressed for local branches but not for remote branches

This is done on purpose, deleting a remote branch may lead to data loss (unlike a local, that can be easily restored with reflogs); thus this operation requires an explicit user confirmation.
This has been discussed many times over the years, and the team feels pretty strong about this.

Many git hosters (like GitHub, etc.) can auto delete merged branches, and `git fetch --prune` will delete local copies too. So if you main scenario to delete obsolete remote branches locally consider using prune.

vbjay:
Take a look at the plugins menu.

pmiossec:
>And I personally also dislike "Delete remote branches" confirmation (and am wondering why it can be suppressed for local branches but not for remote branches).

Because in your local repository, you could "easily" recover commits lost from a deleted branch.
But for a remote, most of the times, you can't recover and so can loose the work of someone else.

So the rule is more "mess your local repository as much as you wish but don't mess other work"

> However, feel free to reject my idea. But I really think that it helps developers to work more efficiently with GitExtensions.

It's the kind of decision where we should be very careful.
Because with one action you could loose much more work and time than the "more efficient" workflow could help you gain.

RussKie:
As an alternative, if you really wish to be able to delete remote branches
without confirmations you could add a script (Settings > Scripts) with
"Show in ReivisionGrid" and bind a hotkey to it (Settings > Hotkeys).

RussKie:
Thank you for your submission but we can't take it in the current form.

1. Remove branches can only be delete after an explicit user confirmation.
2. Local branches can be deleted without "nagging" confirmations after #8956.
3. It is possible to select and delete multiple branches (local or remote) by selecting those in each respective dialog. The UX can be improved, and this is what  #9412 is planned to address.

PR #9591, opened 2021-09-26
Move the init logic from OnLoad to InternalInitialize

RussKie:
Addendum to #9106

## Proposed changes

Restore the splitter positions when we load the form irrespective of the current view (the dashboard or the repo view).

## Test methodology 

- manual

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -200,6 +200,7 @@ namespace GitUI.CommandsDialogs
             _filterBranchHelper = new FilterBranchHelper(toolStripBranchFilterComboBox, toolStripBranchFilterDropDownButton, RevisionGrid);
             _aheadBehindDataProvider = GitVersion.Current.SupportAheadBehindData ? new AheadBehindDataProvider(() => Module.GitExecutable) : null;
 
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -683,43 +696,19 @@ protected override void OnLoad(EventArgs e)
...
-                        () =>
-                        {
-                            RefreshButton.Image = indexChanged && AppSettings.UseFastChecks && Module.IsValidGitWorkingDir()
-                                ? Images.ReloadRevisionsDirty
-                                : Images.ReloadRevisions;
-                        })
-                    .FileAndForget();
-            };
-            UpdateSubmodulesStructure();
-            UpdateStashCount();
```
gerhardol, line 713: Submodules and stash count are no longer updated when opening a repo directly

gerhardol (changes requested):
regression...
Should we include changes like this in 3.5?
They take some extra time to review and test (changes to master are easier to use).

RussKie:
> Should we include changes like this in 3.5?

No :) I've been sitting on the fence, but I have limited the surface of the change to the splitters only.

Juscc53 (commented):
[SPAM]

vbjay:
Reported.

PR #9593, opened 2021-09-26
Move the init logic from OnLoad to InternalInitialize

RussKie:
## Proposed changes

- Move the init logic from OnLoad to InternalInitialize
If the app is opened with `browse <repo>` we would execute a number of instructions that we wouldn't execute if the app was opened in the dashboard-mode. Those instructions included initialising the "Push" button, the revision grid.
Move these instructions into `InternalInitialize` method, that gets executed whenever a dashboard is shown or a repo is loaded. This way we unify the flow, and contain repo-init instructions in one place. 
At the same time restore the splitter positions when we load the form irrespective of the current view.
Addendum to #9106

- Don't apply empty filters on start up

- Invoke UICommandsChanged event when loading the form
`FormBrowse` subscribed to `UICommandsChanged` events, however this event was not fired when the form initially loaded, only after a repository is switched (either set from the dashboard or selecting a different repository).
This was because `GitModuleForm` would assign the passed in `GitUICommands` to a field instead of the property, which raises the event. 
Correct this, assign the passed in commands to the property, and raise the event. React to the the event in `FormBrowse` via the `OnUICommandsChanged` override.


gerhardol (commented):
+1
Some comments on the first commit, see the 3.5 PR
Have not run in master but the second two seem OK

RussKie:
I don't plan making any further changes. 
Best to review commit-by-commit. 88551a8 contains non-functional changes - the code moved verbatim.

RussKie:
@msftbot merge in 3 days

RussKie:
I did quite a bit of testing making sure all methods got hit in various scenarios - e.g. open from a dashboard, open with a repo, go to dashboard and back, etc. Of course there's a chance I missed or lost something, but we can iterate on it.

RussKie:
Unfortunately looks like I did regress a little in here:
https://github.com/gitextensions/gitextensions/blob/d3edfb4865aaec71bd9721fde60a97c600687659/GitUI/CommandsDialogs/FormBrowse.cs#L447-L463

This method is called for any operation that may change the state of the repo (e.g. after a commit), and this will reset the filter and update the path in the terminal.

Getting it fixed in https://github.com/gitextensions/gitextensions/pull/9580

PR #9608, opened 2021-10-02
Fix branch copy and Open with difftool menu switching back to System style

### LINKED ISSUE #9568: "Open with difftool" menu switches right-click menus to Classic style
## Current behaviour

When I click a commit, switch to the diff tab, right-click a filename, and expand the "Open with difftool" menu, the menu switches to classic theme. Other menus (like right-clicking a commit) may or may not switch to classic theme when expanding a submenu.

## Expected behaviour

Expanding menus does not corrupt theming.

## Steps to reproduce

- Open a repository in Git Extensions, and click a commit in the history.
- In the bottom pane, click the Diff tab.
- In the left bar, right-click a filename.
- Hover the "Open with difftool" menu to expand it.

## Screenshots

GitExtensions_KGY7rdCq8b

Unsure.

## Environment

- Git Extensions 3.5.3.12551
- Build 00604a59a673142a3d34a22aa657466a922977f3
- Git 2.32.0.windows.2
- .NET Framework 4.8.4400.0
- DPI 120dpi (125% scaling)

This *may* be related to #9567.

Greybird:
Fixes #9568

## Proposed changes

- Fix MenuUtil determination of renderer to use for Caption-style menu items

## Screenshots 
OpenWithDiffTool
BranchCopy

### Before

Drop down opening for branches copy and Open with diff tool menu items were making the full menu switch to System style.

### After

The Caption-Style is retaining the current theme used, even if the renderer is replaced by a GitExtension specific implementation to allow for a `IMenuItemBackgroundFilter` to be used.

## Test methodology 

- Manual tests

## Test environment(s) 

- GIT 2.28
- Windows 10

## Merge strategy

----


GitExtUtils/GitUI/ToolStripExtensions.cs
```diff
@@ -56,6 +56,6 @@ internal static void EnableTheming(this ToolStrip toolStrip, bool enable)
         }
 
         internal static bool IsThemingEnabled(this ToolStrip toolStrip) =>
-            toolStrip.Renderer is ToolStripExThemeAwareRenderer;
+            toolStrip.Renderer is ToolStripProfessionalRenderer;
```
gerhardol, line 59: This seems to be working also in 3.5 where theming is possible.

gerhardol (commented):
+0

RussKie:
> +0

Why "+0"?

GitExtUtils/GitUI/ToolStripExtensions.cs
RussKie: I'm not seeing `ToolStripExThemeAwareRenderer` used even in release/3.5 branch... So I think the types all references can be completely deleted.
The only question remains what was the intent for this type? It was introduced in #8899 but even there I can't see any instantiations. @NikolayXHD any thoughts?

Greybird: Hi @RussKie,

To give a little bit of context, it is actually created through reflection, via:
https://github.com/gitextensions/gitextensions/blob/d488e497cc6244483794abedb8e4779495ba861c/GitExtUtils/GitUI/ToolStripExtensions.cs#L43

This code is used when using either `ToolStripEx`, through `UseCustomRenderer`, or through a call to `EnableTheming`, done by the `ThemeFix` class.

The main purpose of these two type is to extend WinForms renderers to allow registering a `IMenuItemBackgroundFilter`.
This feature is used for a single purpose in master, through:
https://github.com/gitextensions/gitextensions/blob/5190ba5c1a5f35b321295c59fb04e2a8ae454c14/GitUI/UserControls/RevisionGrid/MenuUtil.cs#L28-L31

This usage is done in the two locations that were having the bug, due to the `ToolStripExtensions` not considering the fact that if GitExtensions theming is not activated, then the actual type is the WinForms one, as the `ThemeFix` will not call the `EnableTheming`:
https://github.com/gitextensions/gitextensions/blob/3d4ed902fcd13dba41d66dd1e6b918ba78eac3f5/GitExtUtils/GitUI/Theming/ThemeFix.cs#L21-L26

gerhardol:
> Why "+0"?

Gerrit notation: Review without any opinion if this should be merged or not. It seems to be working though.
+/- 1 is recommendation
+/- 2 is approving/requesting changes

Greybird:
@RussKie : added the CLA sign-off, sorry for missing this part.

GitExtUtils/GitUI/ToolStripExtensions.cs
NikolayXHD: What @Greybird says is correct.

In addition to that, in Windows10 `MenuItemBackgroundFilter` appears to be unnecessary, which renders `OnRenderMenuItemBackground` methods obsolete.

I am going to test, if this is also unnecessary in Windows 7. Bases on the outcome, either a bunch of code will be removed, or the documentation will reflect that the code is required to render "caption" menu items correctly in Windows 7

RussKie:
@NikolayXHD I'm going to take this change. Once you have more time to propose a better fix we'll take it.

PR #9616, opened 2021-10-05
Fix #6276: "Show messages of annotated tags" truncates to ten lines without notice

### LINKED ISSUE #6276: "Show messages of annotated tags" truncates to ten lines without notice
## Current behaviour

 * Only the first ten lines of an annotated tag's message are shown in commit pane.
 * There is no indication that the tag message was truncated (which may trick people into thinking that what they see is all what there is).

## Expected behavior

At least, there should be an indicator telling that the message is truncated.
Ideally, the tag message should be shown in its entirety.

## Steps to reproduce

1. Create an annotated git tag with a message that has more than ten lines of text.
2. Review the tagged commit in Git Extensions.

## Screenshots

Not needed as the issue is trivial to reproduce and see.

No. The ability to view annotated tag messages was implemented in #1664, where the ten-line limit was hard-coded right from the start:

        string info = RunGitCmd("tag -l -n10 " + tag, SystemEncoding);

## Environment

- GitExtensions version: 3.0.2.5232
- OS version: Windows NT 10.0.17134.0
- .NET version: 4.7.3260.0
- DPI 96dpi (no scaling)

## Diagnostics

Irrelevant.

lhiginbotham:
Fixes #6276: "Show messages of annotated tags" truncates to ten lines without notice

## Proposed changes

- use `git cat-file -p <tag name>` to get full tag message, preventing the previous 10 line limitation

### Before

image

### After

image

## Test methodology

Tested the following scenarios:
- a non-annotated tag (result: no special treatment, tag name does not show in commit message area)
- a tag with new line characters only (result: same as non-annotated tag)
- a tag with a null message (result: name appears, no message)
- a tag with varying numbers of new lines between lines of text (result: multiple new lines condense into a single new line, message shows)
- a tag with more than ten lines of text (result: all lines of text appear)

## Test environment(s) 

- GIT 2.33.0.windows.1

## Merge strategy

----


RussKie (commented):
👍 
Do you think you could add an non-UI integration test (i.e. with [`ReferenceRepository`](https://github.com/gitextensions/gitextensions/blob/master/UnitTests/CommonTestUtils/ReferenceRepository.cs) but without a UI involved)?

GitCommands/Git/GitModule.cs
```diff
@@ -3109,22 +3109,33 @@ public IReadOnlyList<string> GetAllTagsWhichContainGivenCommit(ObjectId objectId
...
+             *
+             * $ git cat-file -p 1.57
+             * tree e776189859a8d635158dc36a1eb15778337fd977
+             * parent ffec6aabd6ec2e2984e3787d8ad6d626e5da9bc8
+             * author Henk Westhuis <Henk> 1237671104 +0100
+             * committer Henk Westhuis <Henk> 1237671104 +0100
+             *
+             * Minor changes.
+             *
+             * Packed with Git-1.6.1 to avoid a bug in Git-1.6.2
```
gerhardol, line 3131: Remove this irrelevant comment

GitCommands/Git/GitModule.cs
```diff
@@ -3134,18 +3145,24 @@ public IReadOnlyList<string> GetAllTagsWhichContainGivenCommit(ObjectId objectId
             }
 
             string output = exec.StandardOutput;
-            if (!output.StartsWith(tag))
+            string[] messageLines = output.Split(
+                new string[] { "\r\n", "\r", "\n" },
+                StringSplitOptions.None);
+
+            if (messageLines.Length < 6)
```
gerhardol, line 3152: Add a constant for the magic number 5/6

gerhardol (commented):
Long captions are truncated to not overflow the user with information. There has been several changes to limit commit messages in various tooltips for instance.
If number of branches exceeds 10, they are limited in the commit info.

This goes the other way. It could be done as "add ore info" too.
Or just add a comment like: No check for length, also displaying very long annotation messages

lhiginbotham:
@mstv, you said:
> Long captions are truncated to not overflow the user with information. There has been several changes to limit commit messages in various tooltips for instance. If number of branches exceeds 10, they are limited in the commit info.
> 
> **This goes the other way. It could be done as "add [m]ore info" too. Or just add a comment like: No check for length, also displaying very long annotation messages**

Sorry, but are you meaning that I must add a comment in the code to warn devs that we are allowing for potentially large annotation messages..?  Or are you suggesting I expose an option to the user to either display a truncated message (probably with an ellipsis at the end `[...]`) vs the full annotation message?

Or are you suggesting nothing at all, just making a general observation for everyone's information?

lhiginbotham:
@RussKie I wrote the tests that you have asked for.  They helped me catch the fact that I was accidentally appending an extraneous new line character on the end of the `GetTagMessage` string! (Due to calling `AppendLine` on all message rather than on all except the last, where `Append` would have been more appropriate for the last message line).

I left a note saying that we should put in "sad-path" tests to test things like if `GetTagMessage` is passed a non-tag reference, but doing so would require some refactoring, which might be better left for a separate refactoring issue where we can attempt to generalize `GetTagMessage` into something like `GetObjectMessage` (if this is even necessary, seeing as there is already logic _somewhere_ for rendering commit message bodies; at the very least, we can put a `return null;` if `GetTagMessage` is passed a non-tag ref).

If you are fine with leaving things as-is, this PR should be good to go!

lhiginbotham:
@RussKie are there any pending items on this PR?

PR #9623, opened 2021-10-06
Don't update filters or issue CD command to the terminal

RussKie:
Fix a regression introduced in #9593

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -446,20 +446,6 @@ namespace GitUI.CommandsDialogs
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1691,6 +1679,29 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
...
+                    {
+                        Debug.WriteLine($"Log output encoding: {Module.LogOutputEncoding.EncodingName}");
+                    }
+#endif
+
+                    // Reset the filter when switching repos
+
+                    // If we're applying custom branch or revision filters - reset them
+                    RevisionGrid.DisableFilters();
+                    SetPathFilter(pathFilter: null);
```
gerhardol, line 1700: I do not think that is needed

RussKie: Why not?

PR #9629, opened 2021-10-08
Package plugins

### LINKED ISSUE #9614: .net5: dotnet publish does not include plugins
## Current behaviour

`dotnet publish` only includes the main app, not the default plugins.
The portable build in Appveyor (as well as for private builds) does not include the plugins.

Working in builds started from the build path `artifacts\Debug\bin\GitExtensions\net5.0-windows`

## Expected behaviour

Plugins included

## Steps to reproduce

See Appveyor artifacts

This is only in master (.net5).
(May have been OK in earlier .net5 builds, do not remember.)

## Environment

- Git Extensions 33.33.33
- Build 810519e1fbdc4d618414bc4042e008a705164d86
- Git 2.33.0.windows.2
- .NET 5.0.10
- DPI 96dpi (no scaling)

RussKie:
Resolves #9614

NB: `CopyLocalLockFileAssemblies` slightly increases the size of `\artifacts\Release\bin\GitExtensions\net5.0-windows\Plugins\` folder ~78MB -> ~130MB. There are few possible tweaks to reduce the size back, but this isn't blocking.

### DIFF
--- a/GitExtensions/GitExtensions.csproj
+++ b/GitExtensions/GitExtensions.csproj
@@ -25,7 +25,13 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="WiX" />
+    <!-- This is needed for scripts\Publish.targets to resolve .NET references -->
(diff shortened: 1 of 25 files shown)


RussKie:
I'll look into these

PR #9630, opened 2021-10-08
Add support for Del key to RepoObjectsTree nodes

mdonatas:
## Motivation

When cleaning up branches it is very convenient to do that from the repo objects tree panel as the branches are sorted by date and it's easy to choose unnecessary ones. However having to open a context menu and click on `Delete` menu item is a big hurdle when compared to the whole process. It's actually easier to click on a line in a rev-grid and press `del` when focus is there. Enabling `del` key in repo objects tree panel solves that.

## Proposed changes

- Enable triggering `Delete` form for repo objects tree items with a `del` key
- Implementation is exactly the same as is already done for triggering `Rename` form with `F2` key

### Before

image

image

image

### After

image

image

image

## Test methodology 

- Manually. Pressing `del` key on a local branch, remote branch and tag opens up delete form

## Test environment(s) 

- Git Extensions 33.33.33
- Build 28321b06951cfcd0c54683feec189564808e209a
- Git 2.33.0.windows.2
- .NET 5.0.10
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/BranchTreePanel/ContextMenu/MenuItemsGenerator.cs
+++ b/GitUI/BranchTreePanel/ContextMenu/MenuItemsGenerator.cs
@@ -107,9 +107,10 @@ namespace GitUI.BranchTreePanel.ContextMenu
         {
             if (Implements<ICanDelete>())
             {
-                yield return _menuItemFactory.CreateMenuItem<ToolStripMenuItem, TNode>(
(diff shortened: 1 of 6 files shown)


vbjay:
Hurdle is kinda on purpose.

vbjay:
Oh ok. I see what you mean. I'd just make sure it only works when the menu
would show and be enabled.

RussKie:
Thank you

PR #9636, opened 2021-10-09
RefactorI ExternalLinksStorage class

ghost:
## Proposed changes

- Remove implementation dependency and hierarchies

## Test methodology 

- unit tests
- manually

## Test environment(s) 

- Git Extensions 33.33.33
- Build 55adb727e53407c11fd24b97e682f1f8156e1fda
- Git 2.30.2.windows.1
- Microsoft Windows NT 10.0.22000.0
- .NET 5.0.10
- DPI 96dpi (no scaling)

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9637, opened 2021-10-09
Change ConfiguredLinkDefinitionsProvider class

ghost:
## Proposed changes

- Remove implementation dependency 

Apply after #9636

## Test methodology 

- unit tests
- manually

## Test environment(s) 

- Git Extensions 33.33.33
- Build 55adb727e53407c11fd24b97e682f1f8156e1fda
- Git 2.30.2.windows.1
- .NET 5.0.10
- DPI 96dpi (no scaling)

## Merge strategy

----


ivangrek:
I will add here. If specific processing is required here, then it is better if we pass a separate implementation of the container, and take ISettingsSource instead of RepoDistSettings as input. But it is better to do it in the next step.

PR #9650, opened 2021-10-15
Fix revision lane order

mdonatas:
## The Story

I am running custom build periodically rebased off of `master` at work and for some time now I've felt there is something confusing/wrong with revision graph lanes but couldn't put a finger on it. Until today. Today I've created a feature branch off of `master` and merged it in immediately and revision graph looked clearly wrong - I expected `master` branch to be on the left and feature branch commit on the right but that wasn't the case (screenshots below).

What ensued was 1.5h of bisecting 7 months worth of commits with `git clean -xdf` in between some of checkouts as switching between .net 5, .net 3.1 and so on comes heavy on VS.

It turned out (at least I suspect so..) that this got introduced completely by accident here https://github.com/gitextensions/gitextensions/commit/02f6edaba7628987609b2cf5642fa7dd1146e025#diff-e0c72dee0a616951e33a8b95092faa5d18fa54019f17002bbf8150edaa11b3b6L95-L96 as `SynchronizedCollection` is no longer available int .net core

## Proposed changes

- Use `BlockingCollection` instead of `ConcurrentBag`. Before the mentioned commit `SynchronizedCollection` was used there and it is based on `List<T>` which means order is important and that a bag is not a suitable replacement.

### Before

image

### After

image

## Test methodology 

- Manually inspected revision graph

## Test environment(s) 

- Git Extensions 33.33.33
- Build e683a364f1049441e251e21d7ff3b0a552e6d8b5
- Git 2.33.0.windows.2
- .NET 5.0.11
- DPI 120dpi (125% scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -320,13 +320,14 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
                 // The list containing the segments is created later. We can set the correct capacity then, to prevent resizing
                 List<RevisionGraphSegment> segments;
+                RevisionGraphSegment[] revisionStartSegments = revision.GetStartSegments();
 
                 if (nextIndex == 0)
                 {
                     // This is the first row. Start with only the startsegments of this row
-                    segments = new List<RevisionGraphSegment>(revision.StartSegments);
+                    segments = new List<RevisionGraphSegment>(revisionStartSegments);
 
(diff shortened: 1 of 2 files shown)


mdonatas:
@mstv Sure, good point. I've also reduced access to it. Any `foreach`, `.First()` or even `.Count` is relatively expensive as it needs to block everything to make sure collection doesn't change during these operations. Converting it to array once is better.

@RussKie it's fine. The journey is exhausting, but results are rewarding. I kinda like these types of issues ;)

PR #9652, opened 2021-10-16
Fix: Prevent the find-replace dialog from blocking Windows shutdown

### LINKED ISSUE #9649: GitExtensions blocks Windows logout/shutdown
### Environment

- Git Extensions 3.5.4.12724
- Git 2.31.1.windows.1
- .NET Framework 4.8.4420.0
- DPI 96dpi (no scaling)

### Issue description

In themed GitExtensions, or in any session in which one has viewed the `FindAndReplace` form, an attempt to log out or shut down without closing GitExtensions will result in the app preventing the action:

This app is preventing shutdown

### Steps to reproduce

- Open the app. 
- Launch a "find and replace" window:
    - Select an item in the file tree
    - Select the content of the file
    - Press <kbd>Ctrl</kbd>+<kbd>F</kbd>.
- Close the "find and replace" window (optional - it'll fail either way)
- Try to log out or shut down

If you're using a theme (I use the dark theme), it doesn't appear to be necessary to launch the window, as I suspect the form is preloaded by the theme builder

It appears not (the code has been this way since 2009), but the theming may have made it more likely to occur.

Original commit: 76ef6768785f9c71a8d41b26065399471de63b12

### Diagnostics

I have tested using the window messaging test tool [SendMessage.exe](https://community.chocolatey.org/packages/sendmessage), and it's clear that the `FindAndReplaceForm` is the problem. It responds with `0` to the window message `WM_QUERYENDSESSION`, which prevents the form from being closed. This is because the code is deliberately blocking the form close event so the form can be reused:

https://github.com/gitextensions/gitextensions/blob/55f85f44ccfae01437c4340cd0397995cc434a4d/GitUI/Editor/FindAndReplaceForm.cs#L291-L305

The `CloseReason` `enum` looks like this:

```csharp
public enum CloseReason
{
    None = 0,
    WindowsShutDown = 1,
    MdiFormClosing = 2,
    UserClosing = 3,
    TaskManagerClosing = 4,
    FormOwnerClosing = 5,
   ApplicationExitCall = 6
}
```

... and it feels to me that the correct check in the `FormClosing` event should be this:

```csharp
if (e.CloseReason == CloseReason.UserClosing)
```

This is the value I see if I close the form manually, and it seems reasonable that all other reasons should be respected and not blocked. I have verified by browsing around in the app and opening/closing the dialog, opening/closing repos, that the only close event received during normal use appears to be the `UserClosing` event.

When I trigger logout _or_ shutdown, I receive a `WindowsShutDown` event, so the defensive change would just be to allow that event through, but I do feel that `UserClosing` is really the only close event that we would need to block.

I am happy to make this fix myself - I can open a PR in the next few days.

simonmckenzie:
Fixes #9649, where Git Extensions prevents logout/shutdown if the find/replace dialog has been initialised at any time in the user's session (regardless of whether the dialog is visible)

This app is preventing shutdown

## Proposed changes
- Update the "prevent disposal" code in `FindAndReplaceForm` to only block user-initiated close operations. This ensures that when the operating system requests shutdown (specifically, by sending the `WM_QUERYENDSESSION` message), the app will not block shutdown

## Test methodology
- Manual testing - to validate this change via tests would be excessive, as it would require the sending of window messages to the application, which, to my mind, is too low-level. Instead, I performed many manual tests. In all test sessions (many combinations of open/close app, repo, find/replace etc.), the only `CloseReasons` received were `UserClosing`, `FormOwnerClosing`, and `WindowsShutdown`. Even terminating through the task manager still sent a message of `FormOwnerClosing` 🤷. The code change handles all of these options correctly.
- Consideration of _all_ `enum` values for `CloseReason` and what the form should do for each:
    - `None` - this reason is never actually received, it's just a sensible default value, so there's no need for special processing
    - `WindowsShutDown`, `TaskManagerClosing`, `ApplicationExitCall` - the app should definitely terminate
    - `FormOwnerClosing` - the owner is `FormBrowse`, so when the owner closes, the form should close too
    - `MdiFormClosing` - not applicable. There is no `MdiParent`
    - `UserClosing` - this is the only case where the form's `Close` should be blocked for the purpose of later reuse

## Test environment(s)

- GIT `git version 2.33.0.windows.2`
- Windows 10 21H1

## Merge strategy

- Squash merge (maintainer to decide merge message, PR submitter should cleanup commits/messages at PR approval).
The maintainer may still request the contributor to squash and rebase, to make sure that merges and commit messages are clarified.

----

### DIFF
--- a/GitUI/Editor/FindAndReplaceForm.cs
+++ b/GitUI/Editor/FindAndReplaceForm.cs
@@ -291,7 +291,7 @@ namespace GitUI
         private void FindAndReplaceForm_FormClosing(object sender, FormClosingEventArgs e)
         {
             // Prevent dispose, as this form can be re-used
-            if (e.CloseReason != CloseReason.FormOwnerClosing)
+            if (e.CloseReason == CloseReason.UserClosing)
             {
                 Owner?.Select(); // prevent another app from being activated instead
 


simonmckenzie:
Hi @RussKie , yes, I have signed off - I did another PR a few days ago and did it then.
https://github.com/gitextensions/gitextensions/blob/55f85f44ccfae01437c4340cd0397995cc434a4d/contributors.txt#L169

PR #9655, opened 2021-10-16
Add mnemonics to every menu item in FormBrowse

Timwi:
## Proposed changes

- Added mnemonics to a lot of menu items in the main window (FormBrowse, FormBrowseMenus)

### Before

image

### After

image

## Caveats

**Main menu bar:** The “GitHub” menu still doesn’t have a mnemonic. This is because the property `IGitPlugin.Name` is used to generate the menu label. I will investigate this further and maybe propose a course of action in Discussions at a later point.

**View menu:** There are too many menu items to assign mnemonics to all. It could be argued that this menu doesn’t need mnemonics everywhere because most people will set the options once and then leave them. However, it would be a relatively simple fix: we can move the 7 options from “Show revision graph column” to “Show build status text” into a submenu named “Columns”. Perhaps another set of options could also be grouped into a submenu. Please indicate whether I have your go-ahead to do this.

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -8770,14 +8770,6 @@ This will just checkout on the submodule the commit determined by the superproje
       </trans-unit>
     </body>
   </file>
-  <file datatype="plaintext" original="Plugin" source-language="en">
-    <body>
-      <trans-unit id="Description.Text">
-        <source>Plugin Manager</source>
-        <target />
-      </trans-unit>
-    </body>
-  </file>
   <file datatype="plaintext" original="PluginRootIntroductionPage" source-language="en">
     <body>
       <trans-unit id="$this.Text">


Timwi:
Is this build failure something I can address?

Timwi:
> Yes, update the translation file, have a look at that:

I already did that, and doing it again made no additional changes to the file. Anything else?

mstv:
You can check https://ci.appveyor.com/api/buildjobs/rnr6gv9b43ae8x0a/artifacts/English.xlf for differences.

pmiossec:
Looking at the error message, I wonder if it's not a problem of end of line

Timwi:
I’ve removed it and there’s still a build failure... what now?

pmiossec:
> I’ve removed it and there’s still a build failure... what now?

You also rebased your PR on top of master if I guess well.
And master is failing due to an error introduced by a previous PR merged. Just wait for this build error to be fixed.

PR #9661, opened 2021-10-18
Fix loading user plugins.

maraf:
Fixes gitextensions/gitextensions.pluginmanager#57

## Proposed changes

Before this change the user plugins path was set in the PluginRegistry, which was initialized in the FormBrowse and the FormCommit. The MEF catalog is initialized earlier in the Program.RunApplication and so user plugins are never loaded.

This change moves initialization of user plugins path to the MEF initialization.

## Test methodology 

- Manually

## Test environment(s) 

- GIT 2.31.1.windows.1
- Windows 10.0.19043.0

----

### DIFF
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -105,10 +105,11 @@ namespace GitExtensions
             }
 
             ManagedExtensibility.Initialise(new[]
-            {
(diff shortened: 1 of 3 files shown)


RussKie:
This looks sensible, just need to get the build fixed (broken since #8809) before we can merge this.

PR #8834, opened 2021-02-08
Move Build Server integration to plugin

ivangrek:
ref to #8766

## Proposed changes

- Move Build Server integration to plugin to transfer to a separate repo

### Before

image

### After

image

## Test methodology 

- Manually
- Visually

## Test environment(s) 

- Git Extensions 33.33.33
- Build b69f89ef38d59ee8657f8df2451f2ceb9cdc9f59 (Dirty)
- Git 2.30.0.windows.2
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9663, opened 2021-10-18
Move Build Server integration to plugin

ghost:
Discussion started #8834

## Proposed changes

- Move Build Server integration to plugin
- Combine all parts in one place

### Before

image

### After

image

## Test methodology 

- manually
- unit tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build 779e93356fc4e6d52bed44356be185d52b4474f3
- Git 2.30.2.windows.1
- .NET 5.0.11
- DPI 96dpi (no scaling)

## Merge strategy

----


ivangrek:
image

Added replacement of values in the settings file.

`AppSettings.MigrateBuildServerIntegrationSettings();`

```
public static void MigrateBuildServerIntegrationSettings()
{
    try
    {
        const string settingsName = "__MigrateBuildServerIntegrationSettings";

        bool migrated = GetBool(settingsName, false);

        if (!migrated)
        {
            string settingsFileText = File.ReadAllText(SettingsFilePath);
            string result = settingsFileText.Replace("BuildServer.", $"{BuildServerIntegrationPluginId}.");
            File.WriteAllText(SettingsFilePath, result);
        }

        SetBool(settingsName, true);
    }
    catch
    {
        // ignored
    }
}
```

Need help testing other integrations.

ivangrek:
image

No translations, but an error.

RussKie:
The error is here:
image

This test is unstable.

RussKie:
I've tried running your branch but I don't see the build server integration at all... What am I missing?

Current vs this branch:
image

ivangrek:
The assembly may have failed. It was so. I ran the build several times. Perhaps I have not configured something?

ivangrek:
Cleaned up the artifacts folder and launched the application.

image

RussKie:
Doh, I didn't realise the plugin is moved under the Plugins node (where all other plugings are).

I've done a clean build, and I'm still not seeing settings correctly converted - i.e. the project repo is still not showing the AppVeyor configs.

GitCommands/Settings/AppSettings.cs
```diff
@@ -1626,6 +1629,29 @@ public static void LoadSettings()
...
+                    File.WriteAllText(SettingsFilePath, result);
+                }
+
+                SetBool(settingsName, true);
+            }
+            catch
+            {
+                // ignored
+            }
+        }
```
RussKie, line 1653: This is a very blunt approach, and it will break users who run different versions side-by-side, e.g. v3.x and v4.x. Besides in which settings file are we replacing this? We have several layers of configs - the global (`C:\Users\<user>\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.settings`), the team (`<repo root>\GitExtensions.settings`) and the user (`<repo root>\.git\GitExtensions.settings`).

Whenever we open a repo we need to do something like this:
- load the config, 
- check if settings were migrated
- if not:
    -  read the original settings
    - write the new settings (to the settings at the same layer)
    - write the flag indicating that the settings were migrated (to the settings at the same layer)
- if yes - read the new settings

PR #9694, opened 2021-10-27
Experimental support of dashboard details list view

### LINKED ISSUE #8561: Dashboard list view
## Feature description

The current dashboard look and feel is useful, especially when categories are used.
Still, it shows repositories in the view which is similar to "Tile" view in Explorer.
It makes it difficult to find repositories visually since you need to scan from left to right, top to bottom.

It would be good if it is possible to switch to another view, similar to "Details" view in the Explorer. It is a grid where columns might be: Path, Branch, Last Access Date.
Sorting by Path would simplify looking for a repository a lot since all work-related repositories are in the same folder, so it's very quick to find the required one in the alphabetical list.

What would be the opinion on this feature? I'm ready to try to contribute it if it looks appropriate.

## Environment

- Git Extensions 3.4.3.9999
- Git 2.27.0.windows.1
- .NET Framework 4.8.4250.0
- DPI 96dpi (no scaling)

ilabutin:
Fixes #8561

## Proposed changes

- Dashboard uses details view instead of Large Icons view

### Before

image

### After

image

## Test methodology 

- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy

----


ilabutin:
I'm happy to work on this PR to make the dashboard as useful as we can make it.
Let me reiterate all the -requirements- wishes to ensure I correctly understand them:
* make two list views
    * "recent repositories" view is shown on the left
    * "full list of repositories" view is shown on the right
    * @RussKie Does it also apply to the current way of displaying repositories (the grid view with large icons)?
    * How to distribute the space? I assume that it is better to autosize the width of the "recent" list and then use the rest of the space to the right for the "full list of repositories". Do you agree?
* Use proper icon for the repository (in the list view)
* [optional] Use branch icon next to the branch name (in the list view)
* Remove "category" column from the list view
* Find good icons for list/grid switch and place them in the header space.

RussKie:
Below are _personal_ thoughts and views (thank you for asking). Others may have there own views and wishes.

> * make two list views
>   
>   * "recent repositories" view is shown on the left

Yes. 
(Stretch goal) I'd think it could valuable to add grouping by date, e.g. "Today", "This week", "Older".
Similar to how VS has it:
image


>   * "full list of repositories" view is shown on the right

Yes

>   * @RussKie Does it also apply to the current way of displaying repositories (the grid view with large icons)?

I think the layout should apply to all views. This will make it easier to implement and maintain in my view.

>   * How to distribute the space? I assume that it is better to autosize the width of the "recent" list and then use the rest of the space to the right for the "full list of repositories". Do you agree?


Sizing is difficult to determine up front, as there are a number of scenarios and possibilities to address those. E.g., here's a telemetry data from 4K users in the last 90 days:
image
The majority of our users work on 1920x1080 monitors (they may have other monitors too). So we must optimise the dashboard to look good on this resolution. We also should strive to make it usable at 1366x768. And on 4K as well, though this is the opposite problem - the dashboard shouldn't have too much of empty or "wasted" space.
And we can't assume that users always use the app maximised to the full screen, e.g.: I never do even if I work on a laptop with a 1920x1080 screen.

I think we may need to experiment, but have some sort of semi-fluid sizing, e.g. (the numbers are arbitrary): 
1. The recents listview would take _no more than 30% but no less than 500px_. 
2. If a path is long and requires more horizontal space than is available the listview should show scrollbars.
3. We also need to handle situations where the available size is too small to render. E.g. like this:
    image


In the tiles view the recents listview should contain a single column, whilst the favourites listview can contain multiple columns (to make use of the available space).
In the detailed view both listview to contain two colimns - path and branch.

I've noticed that in your screenshots you have short paths, which isn't necessary what others use or prefer. So we'll need to cater for everyone. E.g. here's my dashboard (on my personal computer):
image



> * Use proper icon for the repository (in the list view)

Yes

> * [optional] Use branch icon next to the branch name (in the list view)

Yes

> * Remove "category" column from the list view

Yes, in the detailed view. Use `ListViewGroup` functionality instead.

> * Find good icons for list/grid switch and place them in the header space.

Yes. We use icons from http://p.yusukekamiyamane.com/, I'm sure you can find good icons for this.

RussKie:

PR #9697, opened 2021-10-27
Allow user to move context menu items to the extended context menu

### LINKED ISSUE #7863: Add option to move context menu to Extended menu (accessed by Shift right click on Windows)
## Feature description

My right click context menu is becoming a bit cluttered as various programs have added items to it over time.  It would be nice if GitExtensions added an option to the settings to control how the context menu is shown.  Currently the only options are to pick which items are moved into the sub-menu.  I would like an option to either remove the context menu items, or to only show them in the Extended context menu (which is accessed by Shift right mouse click in Windows).

## Environment

- Git Extensions 3.3.1.7897
- Git 2.20.0.windows.1 (recommended: 2.23.0 or later)
- .NET Framework 4.8.4121.0
- DPI 96dpi (no scaling)

lhiginbotham:
This PR refactors GitExtensionsShellEx's menu population code to use more C++ semantics and to be (hopefully) a little bit cleaner.  In addition, it adds support for #7863, allowing the user to customize which shell menu items will only be shown whenever the shift key is held.

( Fixes #7863 )

## Proposed changes

- use more C++ semantics and clean-up the code in the context menu population method
- add a new control to the shell extensions settings page for selecting which items should be relegated to the extended context menu

### Before

image

### After

image

## Test methodology 

- Interactive testing by toggling different settings (show always button, submenu options, extended shell menu options)

## Test environment(s) 

- Git Extensions 3.5.4.12724
- Git 2.33.1.windows.1
- .NET Framework 4.8.4400.0
- DPI 120dpi (125% scaling)

One thing I am worried about is that no matter if I used my desktop with a "normal" 100% UI scaling and two 1920x1080 monitors or if I used my laptop (125% scaling, smaller screen), the designer code always had some sort of changes _just by opening the settings page designer_.  Hopefully these automatic changes are not a problem (I think I've read somewhere that the designer automatically adjusts to the developer's DPI/UI-scaling settings and that the resulting code will be a similar/same UI display for any sort of display settings).

## Merge strategy

----


RussKie:
Looking at the screenshot I wonder if we could make the UI a little more balanced. Right now the added list (presumably containing all the same information) looks squashed... If we can't make both lists the same size without adversely affecting the usability of the first one (e.g., scrolling won't deliver a good UX), perhaps we can use a tabcontrol with two tabs, the first will contain the existing options, and the second will contain the extended properties.
What do you think?

RussKie (commented):
I'm not a C++ expect, the diff looks sensible.

lhiginbotham:
> Looking at the screenshot I wonder if we could make the UI a little more balanced. Right now the added list (presumably containing all the same information) looks squashed... If we can't make both lists the same size without adversely affecting the usability of the first one (e.g., scrolling won't deliver a good UX), perhaps we can use a tabcontrol with two tabs, the first will contain the existing options, and the second will contain the extended properties. What do you think?

@RussKie I was having trouble trying to get them to be the same size so I will try the tabbed control approach

lhiginbotham:
@pmiossec @RussKie @merriemcgaw @KlausLoeffelmann

On my desktop computer (two monitors, 100% UI scaling), I opened the designer page for the shell extension settings page on the master branch with a Visual Studio session opened via `devenv.exe /noscale`.  Without doing anything, I saved the changes and pushed them to a new branch.  Then, I opened Visual Studio on my single-monitor, 125% scaling laptop with `devenv.exe /noscale` and found that when I was using that new branch, I still was getting automatic changes.

See https://github.com/lhiginbotham/gitextensions/commits/lhiginbotham/no-scale-shell-options-designer

RussKie:
Which version of VS are your using?

lhiginbotham:
image

RussKie:
Unfortunately the .NET WinForms designer does not work well for non-default DPI scenarios, and as a result we are removing the prompt to restart Visual Studio at 100%. In addition, the `/noscale` switch in some scenarios can cause incorrect translation of coordinates between the VS process running at 100% and the DesginToolsServer process. We are acutely aware of the issue, and working on a fix for it. In mean time it is recommended to open the design surface on a monitor running at 100% scaling factor.

The above affects _only_ .NET Core and .NET (5.0+) projects, and has no impact for .NET Framework projects.

vbjay:
Or instead of tabs... Could use a grid like the push multiple branches tab on push dialog where you have an  either or selection.  Two columns of checkboxes.  One for main and one for extended.  

image
So instead of Push and Force, could be Main and Extended.

RussKie:
@lhiginbotham is this something you're still planning to pursue?

RussKie:
Superseded by https://github.com/gitextensions/gitextensions/pull/10512.

PR #9699, opened 2021-10-30
GitStatusMonitor could start multiple commands

### LINKED ISSUE #9615: Git Extension : CPU load 100% - Disk space full - GIT.exe process increases
## Current behaviour

I have different PC on Windows 10, with GIT and GIT Extension. One of them has a problem, when I open GIT Extension and the repository, GIT.ext process increases into the task manager. CPU Load is full after few minutes and disk space is also full after 20 or 30 minutes. PC cannot be used after several minutes due to the CPU Load. Into the task manager, after 20 minutes, I have more than 2000 process ! 

This process is called in loop : 
2021-09-10T12:29:27.4966054+02:00	1606	38644	UI	0	git	--no-optional-locks status --porcelain=2 -z --untracked-files --ignore-submodules=none	C:\HIL Applications\
2021-09-10T12:29:29.3576102+02:00	1607	64000	UI	0	git	--no-optional-locks status --porcelain=2 -z --untracked-files --ignore-submodules=none	C:\HIL Applications\

I read other issues on the same topic but I never found a answer to resolve my problem.
https://github.com/gitextensions/gitextensions/issues/6732
https://github.com/microsoft/vscode/issues/71085
https://github.com/git-for-windows/git/issues/3318
https://github.com/gitextensions/gitextensions/issues/230

I tried to uninstall my version of GIT and GIT Extension and install a old one... in case for this problem came from a regression, but I have still the same problem.

The setup is the same than other PC that I have, but only one PC have this behavior. Moreover, into the GIT Extension settings, I deactivate the "commit update" into the button. I though that this process is called to check which files have been changed into the branch. But nothing works...it's very boring. I don't know how can I resolve it. 

It seems that GIT process is launched by GIT Extension but is never killed automatically...and process increases indefinitely...
## Expected behaviour

Same than other PC. GIT Extension can be opened without any over load for the CPU after several minutes. Process GIT can be launched by GIT Extension but is killed after few seconds after.

## Steps to reproduce

Only on 1 PC, I don't know how to reproduce this behavior. 

## Screenshots

I didn't attach my task manager, but you can see GIT process many, many, many times !

I tried different version, but always the same behavior.
GitExtensions-3.1.0.5877
GitExtensions-3.2.1.6628
GitExtensions-3.4.3.9999
GitExtensions-3.5.3.12551

Git-2.26.2-64-bit
git-2.26.3
Git-2.27.0-64-bit
Git-2.28.0-64-bit
Git-2.33.0.2-64-bit

## Environment

- GitExtensions version: See above.
- GIT version: See above.
- OS version: Windows 10 64bits
- .NET version:  ?

## Diagnostics

Thank you for your help !!

gerhardol:
Fixes #9615

Draft, will be updated with master sha if approved
One intention with this PR is to explain the minimal changes needed for #9700

## Proposed changes

If the monitor was inactivated while a status command was running
(minimizing/restoring the app, start a modal dialog) the current command
is canceled to give the latest data.
lock() will ensure that more than one "active" command is not running.
However, one control variable was reset also if the command was canceled
which allowed the canceled command to be followed by another.

Also changed to not request a new status command if a command is already
 running. (current status should be good enough, if there are pending
 changes they will be requested later).

Cherry picked changes from #9700.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -228,8 +228,14 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                         {
                             if (prevStatus == GitStatusMonitorState.Inactive)
                             {
-                                // Timer is already running, schedule new update
-                                ScheduleNextInteractiveTime();
+                                // Timer is already running, schedule a new command only if a command is not running,
+                                // to avoid that many commands are started (and cancelled) if quickly switching Inactive/Running
+                                // If data has changed when Inactive it should be updated by normal means
+                                if (!_commandIsRunning)
+                                {
+                                    ScheduleNextInteractiveTime();
+                                }
+
                                 break;
                             }
 
@@ -313,8 +319,6 @@ namespace GitUI.CommandsDialogs.BrowseDialog
 
             void GitUICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
             {
-                CurrentStatus = GitStatusMonitorState.Inactive;
-                CurrentStatus = GitStatusMonitorState.Running;
                 _isFirstPostRepoChanged = true;
             }
         }
@@ -463,9 +467,9 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                         }
                         finally
                         {
-                            if (!ModuleHasChanged() && !cancelToken.IsCancellationRequested)
+                            lock (_statusSequence)
                             {
-                                lock (_statusSequence)
+                                if (_commandIsRunning && !ModuleHasChanged() && !cancelToken.IsCancellationRequested)
                                 {
                                     if (isSuccess)
                                     {
@@ -478,7 +482,10 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                                             _nextUpdateTime = _nextEarliestTime;
                                         }
                                     }
+                                }
 
+                                if (!cancelToken.IsCancellationRequested)
+                                {
                                     _commandIsRunning = false;
                                 }
                             }


GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -313,8 +319,6 @@ void GitUICommands_PostCheckout(object sender, GitUIPostActionEventArgs e)
 
             void GitUICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
             {
-                CurrentStatus = GitStatusMonitorState.Inactive;
```
gerhardol, line 316: The refresh is triggered from a RequestRefresh(), this is not needed

RussKie:
I don't think we should further update v3.x, and should concentrate on v4 instead.

gerhardol:
> I don't think we should further updathe PR was tote v3.x, and should concentrate on v4 instead.

As in the description, one intention with the PR was to describe the essential changes in #9700 - but I missed to mention #9700.
I would still like to add this to release/3.5, if there is a release for other reasons.

RussKie:
I'd like to release v4 asapp.

pmiossec:
> I'd like to release v4 asapp.

I'd like also but for me the issues of broken forms with hdpi is a big blocker for me.
It makes at least 4 very useful forms unusable.
I don't know if we could release a version with these problems (because hdpi screen are more and more used).

I don't know if dotnet6 will improve something on that...

I will start to use dotnet5 version starting next week so I will see if other things are broken but to make it usable I had to find a workaround for 2 of the forms.

I will see what to do with that....

RussKie:
The dpi anchoring issue was fixed last week in the .NET 7 branch (https://github.com/dotnet/winforms/pull/6122). I expect the fix to be serviced in older releases as well. 
.NET 6.0 goes GA this week, and I plan to migrate the master to .NET 6.0 shortly after.

gerhardol:
I still would like to merge this even if a 3.5 release is not planned.
(If no, I will decline the PR and keep but forget about the branch, not keeping it open.)

RussKie:
Please merge, if you feel strongly about it.

PR #8010, opened 2020-04-18
Feature/add windows explorer integration in settings

NoahLerner:
Fixes #7982  

## Proposed changes

- Add a checkbox in Tools > Settings > ShellExtensions which will toggle on/off Windows Explorer Integration

### Before

image

### After

image

## Test methodology 

- will update...
- 
- 

## Test environment(s) 

- GIT 64 bit
- Windows 10
- English

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9701, opened 2021-10-31
Feature/i7982 shell integration settings

### LINKED ISSUE #7982: Add "Remove Windows Explorer integration" in Settings Menu
## Feature description

Could you add an option in the settings menu to entirely enable/disable the Windows Explorer integration (Shell extension)? There is the possibility to tweak the context-menu, but not to entirely disable it.
I already have more than enough stuff in my right-click menu in Windows Explorer and want to get rid of GitExtensions there.

During installation, you have the possibility to select/unselect the option "Windows Explorer integration". 
image
However, upon installing, it sounded like a good idea. Now the context menu entry sits there and I hate it, cluttering my context menu even though I never use it.
I could uninstall GitExtensions and reinstall it again without the option. This is like using a cannon to kill a mosquito, thinking how easy it must be to add this option.

## Environment

gassanosh:
Fixes #7982
 

## Proposed changes
- Add Tools > Settings > ShellExtensions > Enable Windows Explorer Integration checkbox

It's similar to https://github.com/gitextensions/gitextensions/pull/8010 which is abandoned

### Before

Before

### After

After

## Test methodology 

- Manual testing

## Test environment(s) 

- GIT 2.33
- Windows 10 21H1

## Merge strategy

----


gassanosh:
How can I fix ci?
It says there are some line ending issues in files I never changed and something about TestsResults artifacts I also didn't change.

gerhardol:
> 
> 
> How can I fix ci? It says there are some line ending issues in files I never changed and something about TestsResults artifacts I also didn't change.

https://ci.appveyor.com/project/gitextensions/gitextensions/builds/41353177#L74
is just a warning

What is failing is translations
https://github.com/gitextensions/gitextensions/wiki/Translations#net-50-v36

gerhardol:
Also see #9697

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+namespace GitUI.CommandsDialogs.SettingsDialog.ShellExtension
+{
+    public static class ShellExtensionManager
+    {
+        /// <summary>
+        /// Checks if necessary for registration files were found
+        /// </summary>
+        /// <returns>false only if all necessary files were found, otherwise true</returns>
+        public static bool IsNotInstalled()
+        {
```
RussKie, line 19:
```suggestion
        /// <summary>
        ///  Checks if the shell extension files are found at the known location.
        /// </summary>
        /// <returns><see langword="true"/> the extension files are found at the known location; otherwise <see langword="false"/>.</returns>
        public static bool CheckFilesFound()
        {
```

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+            string path32 = FindFileInBinFolders(CommonLogic.GitExtensionsShellEx32Name);
+            string path64 = FindFileInBinFolders(CommonLogic.GitExtensionsShellEx64Name);
+            return string.IsNullOrEmpty(path32) || (Environment.Is64BitOperatingSystem && string.IsNullOrEmpty(path64));
+        }
+
+        /// <summary>
+        /// Checks if shell extension is properly registered in system
+        /// </summary>
+        /// <returns>false only if all checks are positive, otherwise true</returns>
+        public static bool IsNotRegistered()
```
RussKie, line 29:
```suggestion
        /// <summary>
        ///  Checks if the shell extension is properly registered in the system.
        /// </summary>
        /// <returns><see langword="true"/> if the extension is registered; otherwise <see langword="false"/>.</returns>
        public static bool CheckIfRegistered()
```

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            return string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.LocalMachine, @"Software\Microsoft\Windows\CurrentVersion\Shell Extensions\Approved",
+                    "{3C16B20A-BA16-4156-916F-0A375ECFFE24}")) ||
+                string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot,
+                    @"*\shellex\ContextMenuHandlers\GitExtensions2", null)) ||
```
RussKie, line 34: Keep on the same line. Reading this multiline statement with each sub-statement split across multiple lines is very difficult.
```suggestion
                string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"*\shellex\ContextMenuHandlers\GitExtensions2", null)) ||
```

RussKie: Specify the argument name for `null`s
```suggestion
                string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot,
                    @"*\shellex\ContextMenuHandlers\GitExtensions2", key: null)) ||
```

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (string.IsNullOrEmpty(path))
+            {
+                throw new FileNotFoundException(null, dllName);
+            }
+
+            ProcessStartInfo pi = BuildRegisterStartupInfo(string.Format(argumentsPattern, path));
+            Process.Start(pi)?.WaitForExit();
+        }
+
+        private static ProcessStartInfo BuildRegisterStartupInfo(string arguments)
```
RussKie, line 109: This method is superficial.

GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
```diff
@@ -46,6 +66,17 @@ protected override void PageToSettings()
...
+            {
+                if (cbEnableIntegration.Checked)
+                {
+                    ShellExtensionManager.RegisterShellExtensionSafe();
+                }
+                else
+                {
+                    ShellExtensionManager.UnregisterShellExtensionSafe();
+                }
+            }
```
RussKie, line 79: What are we doing here?

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                // ignored
+            }
+        }
+
+        /// <summary>
+        /// Unregister shell extensions
+        /// </summary>
+        /// <exception cref="FileNotFoundException">If at least one necessary for registration file wasn't found</exception>
+        /// <exception cref="Win32Exception">If user canceled elevation dialog</exception>
+        public static void UnregisterShellExtension() => RegisterAllShellExtensionDlls("/u {0}");
```
RussKie, line 70: "Unregister" calling "register" doesn't look good.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// </summary>
+        /// <exception cref="FileNotFoundException">If at least one necessary for registration file wasn't found</exception>
+        /// <exception cref="Win32Exception">If user canceled elevation dialog</exception>
+        public static void RegisterShellExtension() => RegisterAllShellExtensionDlls("{0}");
+
+        /// <summary>
+        /// Register shell extensions.
+        /// The same as <see cref="RegisterShellExtension"/> but throws no exceptions
+        /// </summary>
+        public static void RegisterShellExtensionSafe()
```
RussKie, line 53: In .NET lingo this should be look something like the following returning a flag indicating success or failure.
```suggestion
        public static bool TryRegister()
```
Though at this stage I'm not sure I understand the need for this.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        private static void RegisterSingleShellExtensionDll(string dllName, string argumentsPattern)
+        {
+            string path = FindFileInBinFolders(dllName);
+            if (string.IsNullOrEmpty(path))
+            {
+                throw new FileNotFoundException(null, dllName);
+            }
+
+            ProcessStartInfo pi = BuildRegisterStartupInfo(string.Format(argumentsPattern, path));
+            Process.Start(pi)?.WaitForExit();
```
RussKie, line 106: Will this lock the app while we're waiting for the registration to complete?

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                    @"Directory\Background\shellex\ContextMenuHandlers\GitExtensions2",
+                    null));
+        }
+
+        /// <summary>
+        /// Register shell extensions
+        /// </summary>
+        /// <exception cref="FileNotFoundException">If at least one necessary for registration file wasn't found</exception>
+        /// <exception cref="Win32Exception">If user canceled elevation dialog</exception>
+        public static void RegisterShellExtension() => RegisterAllShellExtensionDlls("{0}");
```
RussKie, line 47:
```suggestion
        public static void Register() => RegisterAllShellExtensionDlls("{0}");
```

RussKie: Why do we need to pass "{0}"?

RussKie (commented):
First pass with comments.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: The code was extracted as is from ChecklistSettingsPage.cs so I thought it was preferred code style. Will fix it.

gassanosh: As for the specifying argument name for null, I believe it's better to fix CommonLogic.GetRegistryValue

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: Agree, it was oversimplified after refactoring, will inline it.

GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
gassanosh: Do nothing in following cases:
1. Binaries weren't installed
2. Checkbox and shell registration status match each other
I should fix naming IsNotRegistered => IsRegistered so it will be more clear

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: I don't like to introduce return value just in case someone need it or because of naming. I didn't need return value so it's void. That's why it _Safe_ (in meaning not exceptions will be thrown as described in xml doc).
But if you think it's better to introduce not used return value I'll fix it.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: Will fix naming

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: The better name for RegisterAllShellExtensionDlls is RunRegSvrForShellExtensionDlls. And parameter name is argumentsPattern, so here we should provide valid arguments pattern, which for register command is "{0}".
Before refactoring there was a code with string concatenations and strreplace so I decided to rewrite it.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: The code was extracted as is from ChecklistSettingsPage. I decided not to change the logic.
As for the question - indeed, but don't think its an issue because after the call user will see UAC window so waiting here will be OK.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: P.S. I decided that it's better to get rid of *Safe methods.

gassanosh:
I'm not familiar with your workflow, so please let me know if I've done something wrong or I should do something else.

RussKie:
Apologies, I've been incredibly busy at work. I'll try to get to it within the next week or so...

GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
```diff
@@ -46,6 +66,25 @@ protected override void PageToSettings()
...
+                    }
+                    else
+                    {
+                        ShellExtensionManager.Unregister();
+                    }
+                }
+                catch
+                {
+                    // ignored
+                }
```
mstv, line 86: Why? Either `ShellExtensionManager` throws or it has handled the exceptions.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        ///  Checks if the shell extension is properly registered in the system.
+        /// </summary>
+        /// <returns><see langword="true"/> if the extension is registered; otherwise <see langword="false"/>.</returns>
+        public static bool CheckIfRegistered()
+        {
+            return !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.LocalMachine, @"Software\Microsoft\Windows\CurrentVersion\Shell Extensions\Approved", "{3C16B20A-BA16-4156-916F-0A375ECFFE24}")) &&
+                !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"*\shellex\ContextMenuHandlers\GitExtensions2")) &&
+                !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"Directory\shellex\ContextMenuHandlers\GitExtensions2")) &&
+                !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"Directory\Background\shellex\ContextMenuHandlers\GitExtensions2"));
```
mstv, line 34: Support left reading, please:
```suggestion
            return !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.LocalMachine, @"Software\Microsoft\Windows\CurrentVersion\Shell Extensions\Approved", "{3C16B20A-BA16-4156-916F-0A375ECFFE24}"))
                && !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"*\shellex\ContextMenuHandlers\GitExtensions2"))
                && !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"Directory\shellex\ContextMenuHandlers\GitExtensions2"))
                && !string.IsNullOrEmpty(CommonLogic.GetRegistryValue(Registry.ClassesRoot, @"Directory\Background\shellex\ContextMenuHandlers\GitExtensions2"));
```

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <exception cref="FileNotFoundException">If at least one necessary for registration file wasn't found</exception>
+        /// <exception cref="Win32Exception">If user canceled elevation dialog</exception>
+        public static void Register() => RunRegSvrForShellExtensionDlls("{0}");
+
+        /// <summary>
+        /// Unregister shell extensions
+        /// </summary>
+        /// <exception cref="FileNotFoundException">If at least one necessary for registration file wasn't found</exception>
+        /// <exception cref="Win32Exception">If user canceled elevation dialog</exception>
+        public static void Unregister() => RunRegSvrForShellExtensionDlls("/u {0}");
```
mstv, line 49: These are _possible_ causes for these exceptions. `Win32Exception` could also be thrown for other reasons - as well as other exceptions.
So please do not assert these were the only possible exceptions, and catch any `Exception`.
Perhaps `Win32Exception.NativeErrorCode` can be used to ignore a canceled elevation more specific.

mstv (commented):
LGTM but the exception handling; have not run

RussKie is very busy in RL, please be patient.

You could give your sign off in `contributors.txt` in a separate PR, which can be merged quickly.
This will avoid the merge conflict.

GitUI/CommandsDialogs/SettingsDialog/Pages/ShellExtensionSettingsPage.cs
gassanosh: 1. `ShellExtensionManager` can throw exceptions.
2. In this case the exceptions are not critical so I have to catch them.
3. Moreover as far as I get it in PageToSettings I cannot even show any errors so I can only ignore errors from `ShellExtensionManager`.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: It's just refactoring, I extracted the code from ShellExtensionsRegistered_Click to ShellExtensionManager where the operators also were at the end of line. If you think moving operators to the left side will help I'll do so.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
gassanosh: It's also just refactoring. I just noted the assumptions from code to xmldoc.
As I don't want to change old behavior I don't think it is good idea to change this logic.

gassanosh: P.S. Or did you mean I should encapsulate exception logic in ShellExtensionManager and return something like error code instead?

mstv: > P.S. Or did you mean I should encapsulate exception logic in ShellExtensionManager and return something like error code instead?

No, I do not want error codes.

With this refactoring, you have created a public API,
and added xmldoc - good,
but this xmldoc promises that these functions would throw only the listed exceptions - wrong.

I did not review the original code and would not have accepted it.

gassanosh: My primary intension is not to break working code.
1. I don't know whether author wanted to catch only elevation canceled exception, or any Win32Exception.
2. I'm not sure if for every OS and all possible UAC settings canceling elevation dialog will lead to the same exception.

So I don't understand what should I do. Change xml doc? Change logic? Anyway, please, suggest changes as you more familiar with the code.

mstv: Please do not assert in xmldoc that there were no other possible exceptions. The list must be complete.
Swallowing errors is a bad practice. So catch and display all exceptions except for canceled user elevation.
Done in https://github.com/mstv/gitextensions/commit/61c09dc0b0c54fb02ce437523d1d3969be7cfc65

gassanosh: I introduced ShellExtensionManager as "model" layer api. With your change it became "presentation" layer api (as it can show message boxes). I don't like the idea.
The worse part is that now the message box will be shown during PageToSettings.
So I see at least those issues:
1. When message is shown the settings can be in inconsistent state (as message box will appear at random time during saving settings). So the settings will be inconsistent as long as user don't close message box (if ever).
2. It more UX issue but still, when user changes page and clicks save he will see "irrelevant" error. I call it "irrelevant" as the user can do nothing with this error. The settings window is closed, so he will have to open it again and go to proper page to find what have happened.

That's what I meant, when I said that I don't want to break working code and wouldn't have agreed with the change if I had reviewed it.

As for me I'd suggested rolling back `ShowMessage` call from `ShellExtensionManager`, rewrite `FormSettings::Save` so that it will:
1. Catch exceptions from `settingsPage.SaveSettings();`
2. If any exception were caught, continue saving all other settings (or rollback already saved changes), so that settings will be in consistent state.
3. Show error and navigate to page with the issues without closing the setting dialog, so user can see the problem an fix it.

gassanosh: P.S. Also, if you decided to show all errors to the user I believe you should show elevation cancelled error message to the user. If he rejected elevation he will get clear message why we cannot perform requested action.

RussKie (approved):

gassanosh:
So, when the pr will be merged?

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
mstv: > I introduced ShellExtensionManager as "model" layer api. With your change it became "presentation" layer api (as it can show message boxes). I don't like the idea.

Yes, it was a quick draft. The exception handling could be moved to a dedicated class.
The settings page belongs to the "presentation" layer.

> The worse part is that now the message box will be shown during PageToSettings. So I see at least those issues:
> 
>     1. When message is shown the settings can be in inconsistent state (as message box will appear at random time during saving settings). So the settings will be inconsistent as long as user don't close message box (if ever).
>
>     2. It's more UX issue but still, when user changes page and clicks save he will see "irrelevant" error. I call it "irrelevant" as the user can do nothing with this error. The settings window is closed, so he will have to open it again and go to proper page to find what have happened.

With the current approach, the user must reopen the settings anyway because - due to the swallowed exceptions - they cannot know whether it has worked or not.
IMO, the root cause for the UX issues is: This de/registration of the Explorer integration is *not* an actual GE setting, i.e. it is not stored in the settings file.
The de/registration should be done at once - not when the settings are saved. One or two Buttons instead of the CheckBox will give a clear UX (including the possibility to show error messages).
Errors should be displayed by extending the class `MessageBoxes`.

> P.S. Also, if you decided to show all errors to the user I believe you should show elevation cancelled error message to the user too. If he rejected elevation he will get clear message why we cannot perform requested action.

OK for me.

RussKie:
@mstv any further concerns?

mstv:
> @mstv any further concerns?

Yes, as replied to a code comment above:

With the current approach, the user must reopen the settings anyway because - due to the swallowed exceptions - they cannot know whether it has worked or not.
IMO, the root cause for the UX issues is: This de/registration of the Explorer integration is *not* an actual GE setting, i.e. it is not stored in the settings file.
The de/registration should be done at once - not when the settings are saved. One or two Buttons instead of the CheckBox will give a clear UX (including the possibility to show error messages).
(Errors should be displayed by extending the class `MessageBoxes`.)

RussKie:
Closing as stale.

PR #9702, opened 2021-10-31
Support Git in WSL

### LINKED ISSUE #8172: Improve WSL Performance and Permissions for `\\wsl$`
## Feature description

This is a followup to some of the comments related to strange WSL behavior in #8171 

Would it be possible to add special handling for Git working directories located under the special `\\wsl$` Windows Subsystem for Linux path?  Specifically, rather than running a Windows version of `git.exe` for those directories and passing it `\\wsl$` paths, I think it would be much nicer to run the commands directly in WSL via its version of `git`.

This would involve several steps I've thought of, as well as some I probably haven't.

1. Detect `\\wsl$` in the current path argument for GE to enable the special handling.
2. Extract the WSL distro name from the path.  It should be the first directory name in the path after `\\wsl$\`.
3. Extract the remaining part of the path and replace backslashes with forward slashes.  This will be the working directory in WSL, call it `{wsl_cwd}`.
4. Detect any `\\wsl$\{distro}` paths in any arguments that are normally passed to `git.exe`.  Remove the prefix and replace backslashes with forward slashes.  Maybe replace double quotes around paths with single quotes?  Call the resulting arguments string `{wsl_args}`.
5. Send the final command to WSL git in the form:
`C:\Windows\System32\wsl.exe -d {distro} -- /usr/bin/git -C '{wsl_cwd}' {wsl_args}`

This would lead to significant performance improvements, as well as eliminate any strange issues with file system permissions and execute bits that come from trying to edit things via `\\wsl$\`.

For example, the following takes over ten seconds on my machine for a 166 MiB sample project:
`git.exe -c diff.ignoreSubmodules=none status --porcelain=2 -z --untracked-files=all --ignore-submodules=all \\wsl$\Ubuntu\home\ubuntu\MyProject\`

The following runs almost instantaneously on my machine with the same sample project:
`wsl.exe -d Ubuntu -- /usr/bin/git -C '/home/ubuntu/MyProject/' -c diff.ignoreSubmodules=none status --porcelain=2 -z --untracked-files=all --ignore-submodules=all /home/ubuntu/MyProject/`

Thanks for your consideration!

## Environment

- Git Extensions 3.4.0.9381
- Git 2.22.0.windows.1 (recommended: 2.25.1 or later)
- .NET Framework 4.8.4180.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8172

## Proposed changes

For Git repos stored in WSL2 directories, execute the Git executable in WSL
where possible to improve performance. WSL Git commands are several
times faster when executed by the WSL Git executable instead of the
Git Extensions native (Windows) executable.

The paths internal to Git Extensions are always native (Windows).
Using the WSL executable requires that paths provided to WSL Git
are adjusted from/to Git Extensions native/internal (Windows) format to WSL, like
\\wsl$\Ubunto\repo to /repo and c:\repo to /mnt/c/repo.

The Git Extensions native (Windows) Git executable is still used in some situations:
* All handling and settings related to Git in Settings. This includes
display of Git version as well.
However, if the Git version is too old to be supported, the GitVersion
check added in #9680 will report that the version is too old.
* Custom merge implementation in FormResolveConflicts
* Custom difftool/mergetool, the tools are only extracted in Windows.
Default git-difftool is using the Git executable, which could be separate
for WSL Git.
* Some built-in plugins etc, ScriptRunner and FindLargeFiles always use
 native Git

Some additional known limitations:
* The dialog that offers "--force-with-lease" for push do not appear as the output is not redirected. (So the Push dialog appears and --force-with-lease has to manually selected).
* GitHub plugin do not use the Windows token

Note: All commands are not tried in WSL, there may be a few more changes needed.

### Before

See additions below
image

### After

image

WSL is shown separately in Git command log
image

## Test methodology 

Tests added for PathUtil.
For the few changed commands, something could be added. There are no current tests for the commands.

## Merge strategy

Rebase merge (PR submitter must change the commit message for the last commit).

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -393,7 +393,7 @@ namespace GitCommands.Git.Commands
         }
 
         public static ArgumentString RebaseCmd(
-            string? branch, bool interactive, bool preserveMerges, bool autosquash, bool autoStash, bool ignoreDate, bool committerDateIsAuthorDate, string? from = null, string? onto = null)
+            string? branch, bool interactive, bool preserveMerges, bool autosquash, bool autoStash, bool ignoreDate, bool committerDateIsAuthorDate, string? from = null, string? onto = null, bool supportRebaseMerges = true)
         {
             // TODO-NULLABLE does it make sense for 'branch' to be null here?
 
@@ -421,7 +421,7 @@ namespace GitCommands.Git.Commands
 
                 if (preserveMerges)
                 {
-                    builder.Add(GitVersion.Current.SupportRebaseMerges ? "--rebase-merges" : "--preserve-merges");
+                    builder.Add(supportRebaseMerges ? "--rebase-merges" : "--preserve-merges");
                 }
             }
 
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using GitExtUtils;
 using GitUIPluginInterfaces;
@@ -194,14 +195,17 @@ namespace GitCommands.Git.Commands
         /// </summary>
         /// <param name="gitRef">The branch to move.</param>
         /// <param name="targetId">The commit to move to.</param>
-        /// <param name="repoDir">Directory to the current repo.</param>
+        /// <param name="repoDir">Directory to the current repo in Posix format.</param>
         /// <param name="force">Push the reference also if commits are lost.</param>
         /// <returns>The Git command to execute.</returns>
         public static ArgumentString PushLocalCmd(string gitRef, ObjectId targetId, string repoDir, bool force = false)
         {
+            Debug.Assert(repoDir.IndexOf(PathUtil.NativeDirectorySeparatorChar) >= 0,
+                $"'PushLocalCmd' must be called with 'repoDir' in Posix format");
+
             return new GitArgumentBuilder("push")
             {
-                $@"""file://{repoDir.ToPosixPath()}""",
+                $@"""file://{repoDir}""",
                 $"{targetId}:{gitRef}".QuoteNE(),
                 { force, "--force" }
             };
@@ -210,6 +214,8 @@ namespace GitCommands.Git.Commands
         /// <summary>
         /// Git Clone.
         /// </summary>
+        /// <param name="fromPath">URL or file system path in Posix format.</param>
+        /// <param name="toPath">Directory to the destination path in Posix format.</param>
         /// <param name="central">Makes a bare repo.</param>
         /// <param name="branch">
         /// <para><c>NULL</c>: do not checkout working copy (--no-checkout).</para>
@@ -224,7 +230,10 @@ namespace GitCommands.Git.Commands
         /// </param>
         public static ArgumentString CloneCmd(string fromPath, string toPath, bool central = false, bool initSubmodules = false, string? branch = "", int? depth = null, bool? isSingleBranch = null)
         {
-            var from = PathUtil.IsLocalFile(fromPath) ? fromPath.ToPosixPath() : fromPath;
+            Debug.Assert(fromPath.IndexOf(PathUtil.NativeDirectorySeparatorChar) >= 0,
+               $"'CloneCmd' must be called with 'fromPath' in Posix format");
+            Debug.Assert(toPath.IndexOf(PathUtil.NativeDirectorySeparatorChar) >= 0,
+               $"'CloneCmd' must be called with 'toPath' in Posix format");
 
             return new GitArgumentBuilder("clone")
             {
@@ -237,8 +246,8 @@ namespace GitCommands.Git.Commands
                 "--progress",
                 { branch is null, "--no-checkout" },
                 { !string.IsNullOrEmpty(branch), $"--branch {branch}" },
-                from.Trim().Quote(),
-                toPath.ToPosixPath().Trim().Quote()
+                fromPath.Trim().Quote(),
+                toPath.Trim().Quote()
             };
         }
 
@@ -521,7 +530,9 @@ namespace GitCommands.Git.Commands
                 { squash, "--squash" },
                 { noCommit, "--no-commit" },
                 { allowUnrelatedHistories, "--allow-unrelated-histories" },
-                { !string.IsNullOrWhiteSpace(mergeCommitFilePath), $"-F \"{mergeCommitFilePath}\"" }, // let git fail, if the file doesn't exist
+
+                 // let git fail, if the file doesn't exist
+                { !string.IsNullOrWhiteSpace(mergeCommitFilePath), $"-F \"{mergeCommitFilePath}\"" },
                 { log is not null && log.Value > 0, $"--log={log}" },
                 branch
             };
--- a/GitCommands/Git/Executable.cs
+++ b/GitCommands/Git/Executable.cs
@@ -16,16 +16,18 @@ namespace GitCommands
     {
         private readonly string _workingDir;
         private readonly Func<string> _fileNameProvider;
+        private readonly string _prefixArguments;
 
         public Executable(string fileName, string workingDir = "")
             : this(() => fileName, workingDir)
         {
         }
 
-        public Executable(Func<string> fileNameProvider, string workingDir = "")
+        public Executable(Func<string> fileNameProvider, string workingDir = "", string prefixArguments = "")
         {
             _workingDir = workingDir;
             _fileNameProvider = fileNameProvider;
+            _prefixArguments = prefixArguments;
         }
 
         /// <inheritdoc />
@@ -44,7 +46,7 @@ namespace GitCommands
 
             var fileName = _fileNameProvider();
 
-            return new ProcessWrapper(fileName, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute, throwOnErrorExit);
+            return new ProcessWrapper(fileName, _prefixArguments, args, _workingDir, createWindow, redirectInput, redirectOutput, outputEncoding, useShellExecute, throwOnErrorExit);
         }
 
         #region ProcessWrapper
@@ -72,6 +74,7 @@ namespace GitCommands
             private bool _disposed;
 
             public ProcessWrapper(string fileName,
+                                  string prefixArguments,
                                   string arguments,
                                   string workDir,
                                   bool createWindow,
@@ -107,12 +110,12 @@ namespace GitCommands
                         StandardOutputEncoding = outputEncoding,
                         StandardErrorEncoding = errorEncoding,
                         FileName = fileName,
-                        Arguments = arguments,
+                        Arguments = $"{prefixArguments}{arguments}",
                         WorkingDirectory = workDir
                     }
                 };
 
-                _logOperation = CommandLog.LogProcessStart(fileName, arguments, workDir);
+                _logOperation = CommandLog.LogProcessStart($"{fileName} {prefixArguments}".Trim(), arguments, workDir);
 
                 _process.Exited += OnProcessExit;
 
@@ -134,7 +137,7 @@ namespace GitCommands
                     Dispose();
 
                     _logOperation.LogProcessEnd(ex);
-                    throw new ExternalOperationException(fileName, arguments, workDir, innerException: ex);
+                    throw new ExternalOperationException($"{fileName} {prefixArguments}".Trim(), arguments, workDir, innerException: ex);
                 }
             }
 
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -904,7 +904,9 @@ namespace GitCommands
 
         public void RunMergeTool(string? fileName = "", string? customTool = null)
         {
-            var gui = GitVersion.Current.SupportGuiMergeTool ? "--gui" : string.Empty;
+            // Use Windows Git if custom tool is selected as the list is native to the application.
+            bool isWindowsGit = !string.IsNullOrWhiteSpace(customTool);
+            string gui = (isWindowsGit ? GitVersion.Current : GitVersion).SupportGuiMergeTool ? "--gui" : string.Empty;
             GitArgumentBuilder args = new("mergetool")
             {
                 { string.IsNullOrWhiteSpace(customTool), gui, $"--tool={customTool}" },
@@ -912,8 +914,7 @@ namespace GitCommands
                 fileName.ToPosixPath().QuoteNE()
             };
 
-            // Use native (Windows) Git if custom tool is selected as the list is native
-            using var process = (string.IsNullOrWhiteSpace(customTool) ? _gitExecutable : _gitNativeExecutable).Start(args, createWindow: true);
+            using IProcess process = (isWindowsGit ? _gitWindowsExecutable : _gitExecutable).Start(args, createWindow: true);
             process.WaitForExit();
         }
 
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -514,7 +514,8 @@ namespace GitCommands
                 "-z",
                 "--unmerged"
             };
-            return !string.IsNullOrEmpty(_gitExecutable.GetOutput(args));
+            var result = _gitExecutable.Execute(args);
+            return result.ExitedSuccessfully && !string.IsNullOrEmpty(result.StandardOutput);
         }
 
         public bool HandleConflictSelectSide(string fileName, string side)
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -47,8 +47,18 @@ namespace GitCommands
         private readonly IGitTreeParser _gitTreeParser = new GitTreeParser();
         private readonly IRevisionDiffProvider _revisionDiffProvider = new RevisionDiffProvider();
         private readonly GetAllChangedFilesOutputParser _getAllChangedFilesOutputParser;
+
+        // The executable may use Windows Git (native to the app, always used in special situations) or WSL Git.
         private readonly IGitCommandRunner _gitCommandRunner;
+        private readonly IGitCommandRunner _gitWindowsCommandRunner;
         private readonly IExecutable _gitExecutable;
+        private readonly IExecutable _gitWindowsExecutable;
+
+        /// <summary>
+        /// Name of the WSL distro for the GitExecutable, empty string for the app native Windows Git executable.
+        /// This can be seen as the Git "instance" identifier.
+        /// </summary>
+        private readonly string _wslDistro;
 
         public GitModule(string? workingDir)
         {
@@ -56,9 +66,21 @@ namespace GitCommands
             WorkingDirGitDir = GitDirectoryResolverInstance.Resolve(WorkingDir);
             _indexLockManager = new IndexLockManager(this);
             _commitDataManager = new CommitDataManager(() => this);
-            _gitExecutable = new Executable(() => AppSettings.GitCommand, WorkingDir);
-            _gitCommandRunner = new GitCommandRunner(_gitExecutable, () => SystemEncoding);
             _getAllChangedFilesOutputParser = new GetAllChangedFilesOutputParser(() => this);
+            _gitWindowsExecutable = new Executable(() => AppSettings.GitCommand, WorkingDir);
+            _gitWindowsCommandRunner = new GitCommandRunner(_gitWindowsExecutable, () => SystemEncoding);
+
+            _wslDistro = AppSettings.WslGitEnabled ? PathUtil.GetWslDistro(WorkingDir) : "";
+            if (!string.IsNullOrEmpty(_wslDistro))
+            {
+                _gitExecutable = new Executable(() => AppSettings.WslGitCommand, WorkingDir, $"-d {_wslDistro} git ");
+                _gitCommandRunner = new GitCommandRunner(_gitExecutable, () => SystemEncoding);
+            }
+            else
+            {
+                _gitExecutable = _gitWindowsExecutable;
+                _gitCommandRunner = _gitWindowsCommandRunner;
+            }
 
             // If this is a submodule, populate relevant properties.
             // If this is not a submodule, these will all be null.
@@ -130,25 +152,30 @@ namespace GitCommands
             }
         }
 
-        /// <summary>
-        /// Gets the directory which contains the git repository.
-        /// </summary>
+        /// <inherit/>
         public string WorkingDir { get; }
 
         /// <summary>
-        /// Gets the access to the current git executable associated with this module.
+        /// GitVersion for the default GitExecutable.
         /// </summary>
+        public GitVersion GitVersion => GitVersion.CurrentVersion(_wslDistro);
+
+        /// <inherit/>
         public IExecutable GitExecutable => _gitExecutable;
 
-        /// <summary>
-        /// Gets the access to the current git executable associated with this module.
-        /// </summary>
+        /// <inherit/>
         public IGitCommandRunner GitCommandRunner => _gitCommandRunner;
 
         /// <summary>
         /// Gets the location of .git directory for the current working folder.
         /// </summary>
         public string WorkingDirGitDir { get; private set; }
+
+        /// <inherit/>
+        public string GetGitExecPath(string? path) => PathUtil.GetGitExecPath(path, _wslDistro);
+
+        /// <inherit/>
+        public string GetWindowsPath(string path) => PathUtil.GetWindowsPath(path, _wslDistro);
 
         /// <summary>
         /// If this module is a submodule, returns its name, otherwise <c>null</c>.
@@ -311,7 +338,7 @@ namespace GitCommands
             };
             var gitPath = _gitExecutable.GetOutput(args);
 
-            var systemPath = gitPath.Trim().ToNativePath();
+            var systemPath = GetWindowsPath(gitPath).Trim();
 
             if (systemPath.StartsWith(".git\\"))
             {
@@ -346,7 +373,7 @@ namespace GitCommands
                         GitArgumentBuilder args = new("rev-parse") { "--git-common-dir" };
                         ExecutionResult result = _gitExecutable.Execute(args, throwOnErrorExit: false);
 
-                        string dir = result.StandardOutput.Trim().ToNativePath();
+                        string dir = GetWindowsPath(result.StandardOutput).Trim();
 
                         if (!result.ExitedSuccessfully || dir == ".git" || dir == "." || !Directory.Exists(dir))
                         {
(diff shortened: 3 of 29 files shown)


GitCommands/PathUtil.cs
```diff
@@ -12,6 +12,7 @@ public static class PathUtil
     {
         private static readonly IEnvironmentAbstraction EnvironmentAbstraction = new EnvironmentAbstraction();
         private static readonly IEnvironmentPathsProvider EnvironmentPathsProvider = new EnvironmentPathsProvider(EnvironmentAbstraction);
+        private const string _wslPrefix = @"\\wsl$\";
```
RussKie, line 15:
```suggestion
        private const string WslPrefix = @"\\wsl$\";
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +70,14 @@ public interface IGitModule
         /// <summary>Indicates HEAD is not pointing to a branch (i.e. it is detached).</summary>
         bool IsDetachedHead();
 
+        /// <summary>
+        /// Convert native (Windows) path to the path for the Git executable
+        /// (this could be for instance WSL)
+        /// </summary>
+        /// <param name="path">ge.exe path to repo.</param>
```
RussKie, line 77: What does this mean?

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
```diff
@@ -231,6 +231,52 @@ public void IsWslPath(string path, bool expected)
...
+        {
+            PathUtil.GetRepoPath(path, wslDistro).Should().Be(expected);
+        }
+
+        [TestCase(@"\\wsl$/Ubuntu/work/../GitExtensions", "Ubuntu", @"//wsl$/Ubuntu/work/../GitExtensions")]
+        [TestCase(@"\\wsl$\Ubuntu-20.04\work\..\GitExtensions\", "Ubuntu", @"\\wsl$\Ubuntu-20.04\work\..\GitExtensions\")]
+        public void GetRepoPath_unexpected_usage(string path, string wslDistro, string expected)
+        {
+            PathUtil.GetRepoPath(path, wslDistro).Should().Be(expected);
+        }
```
RussKie, line 278: Looking at inputs and outputs for these tests I see that in some cases we're chaining inputs to POSIX and in other tests we don't. From a caller perspective this feels very arbitrary.
I think we ought to normalise the output to be consistent.

GitUI/HelperDialogs/FormResetAnotherBranch.cs
```diff
@@ -93,7 +94,7 @@ private void Ok_Click(object sender, EventArgs e)
                 return;
             }
 
-            var command = GitCommandHelpers.PushLocalCmd(gitRefToReset.CompleteName, _revision.ObjectId, Module.WorkingDir, ForceReset.Checked);
+            var command = GitCommandHelpers.PushLocalCmd(gitRefToReset.CompleteName, _revision.ObjectId, Module.GetRepoPath(Module.WorkingDir), ForceReset.Checked);
```
RussKie, line 97: We need to clearly define and explain the rules when `GetRepoPath` must be used.

Also any reason not to compute and cache this property when we initialise a `GitModule`?

RussKie (commented):
When we open a repo we can definitively determine whether a repo is a Windows repo or a WSL repo, right? Why do we need the checkbox? Is there a use case where a user wouldn't want to run a native git for WSL repo?

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: To set inherit in GitModule
```suggestion
        /// <param name="path">Git Extensions native (Windows) path.</param>
```

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
gerhardol: OK
These paths are intended for the Git executable, so normalizing is required.

GitUI/HelperDialogs/FormResetAnotherBranch.cs
gerhardol: GetRepoPath is required for paths that the Git executable will use.
Maybe rename to GetPathForGitExe()?

gerhardol:
> When we open a repo we can definitively determine whether a repo is a Windows repo or a WSL repo, right? Why do we need the checkbox? Is there a use case where a user wouldn't want to run a native git for WSL repo?

Only if there are issues in the implementation or if you have a very old Git version in your repo.
Some setups still use Ubuntu-16.04 for instance, not supported directly.
I do not think it is a big issue to not support such distros though, this is new functionality.
This could be a settings file only setting too.

If you have opened the WSL repo and flips this switch you have to reopen the repo for the settings to have effect, so if this can be deprecated I would be happier to not have to document this behavior.
(Should have stated in the description that doc need to be updated.)

GitUI/HelperDialogs/FormResetAnotherBranch.cs
RussKie: > GetRepoPath is required for paths that the Git executable will use.

Should this be pushed from UI layer into the GitCommands layer then? That is, instead of forcing callers to remember/know that paths must be encoded in a specific way, should callee (i.e. git commands, git executable, etc.) perform encoding when appropriate?

RussKie: What I'm getting at - is "use WSL" something a consumer should care, or is it an infra setting that GitModule and Co should take care of based on the characteristics of a repo?

RussKie:
> Only if there are issues in the implementation or if you have a very old Git version in your repo.
> Some setups still use Ubuntu-16.04 for instance, not supported directly.
> I do not think it is a big issue to not support such distros though, this is new functionality.
> This could be a settings file only setting too.

Ubuntu 16.04 LTS reach EOL in Apr 2021 (though its EOS is until Apr 2026). If not _truly_ necessary, I'd rather not complicate our implementations, and offer a simple config-free support, if possible.

vbjay:
Small concern. I set up a windows git config. Then I use wsl. Does it use
its own config seperate from all the stuff I set up in windows version?  If
so, we need to show this in git config settings.

gerhardol:
> Small concern. I set up a windows git config. Then I use wsl. Does it use its own config separate from all the stuff I set up in windows version? If so, we need to show this in git config settings.

Settings in GE is only used for native (Windows) Git. There are a number of hardcoded access to .gitconfig etc in addition to that Settings in started globally without knowledge of WSL repos. I have four different distros but do not use Git in more than two of them, all should not be listed.

So WSL Git is a "secondary citizen", must be documented. I do not believe this is a big limitation.
Personally, I could live with just being able to compare, stage etc in GE and other operations from WSL command line.

GitUI/HelperDialogs/FormResetAnotherBranch.cs
gerhardol: This handles specific arguments only. 
An alternative would be to guess what are file arguments to transform when running the Git command (ths is what the wslgit wrapper does). The problem with that is that it is hard to find what is arguments. You may translate for instance searches for "C:\repo" and it is hard to handle where paths end if the paths contain spaces.

For the global "use WSL" configuration, the user will need to know about this as some settings will have to be done separately in WSL. But there should be no need to know details.

GitCommands/PathUtil.cs
gerhardol: done

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Refreshed the descriptions in general

UnitTests/GitCommands.Tests/Helpers/PathUtilTest.cs
gerhardol: Changed this in general and removed som ToPosixPath() that are no longer needed.

GitUI/HelperDialogs/FormResetAnotherBranch.cs
gerhardol: Renamed the method, hopefully clearer now

gerhardol:
> Ubuntu 16.04 LTS reach EOL in Apr 2021 (though its EOS is until Apr 2026). If not _truly_ necessary, I'd rather not complicate our implementations, and offer a simple config-free support, if possible.

Removed the option, changed the text to refer to WSL.

gerhardol:
Superseded by #9720 (as well as some parts in #9718).

gerhardol:
Squashed and rebased to simplify review
Old commits available in branch `tmp/i8172-wsl-git`

Added a few adjustments

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -194,14 +194,14 @@ public static ArgumentString ResetCmd(ResetMode mode, string? commit = null, str
...
-        /// <param name="repoDir">Directory to the current repo.</param>
+        /// <param name="repoDir">Directory to the current repo in Posix format.</param>
         /// <param name="force">Push the reference also if commits are lost.</param>
         /// <returns>The Git command to execute.</returns>
         public static ArgumentString PushLocalCmd(string gitRef, ObjectId targetId, string repoDir, bool force = false)
         {
             return new GitArgumentBuilder("push")
             {
-                $@"""file://{repoDir.ToPosixPath()}""",
+                $@"""file://{repoDir}""",
```
RussKie, line 204: Is is possible to add an debug assertion that the path is in the posix format? It's good that we updated the comments but there're no guards to ensure the path is in the expected format.

GitCommands/Git/Executable.cs
```diff
@@ -107,12 +110,12 @@ private sealed class ProcessWrapper : IProcess
                         StandardOutputEncoding = outputEncoding,
                         StandardErrorEncoding = errorEncoding,
                         FileName = fileName,
-                        Arguments = arguments,
+                        Arguments = prefixArguments + arguments,
```
RussKie, line 113: Use string interpolation
```suggestion
                        Arguments = $"{prefixArguments}{arguments}",
```

GitCommands/Git/GitModule.cs
```diff
@@ -47,18 +47,40 @@ public sealed class GitModule : IGitModule
         private readonly IGitTreeParser _gitTreeParser = new GitTreeParser();
         private readonly IRevisionDiffProvider _revisionDiffProvider = new RevisionDiffProvider();
         private readonly GetAllChangedFilesOutputParser _getAllChangedFilesOutputParser;
+
+        // executable may use "native" Git (required for UI as well as FormProcess) or Git in WSL
```
RussKie, line 51:
```suggestion
        // The executable may use "native" Git (required for UI as well as FormProcess) or Git in WSL
```

GitCommands/Git/GitModule.cs
```diff
@@ -130,26 +152,35 @@ static bool HasGitModulesFile(string path)
...
 
+        /// <inherit/>
+        public string GetPathForGitExec(string? path) => PathUtil.GetRepoPath(path, _wslDistro);
+
+        /// <summary>
+        /// Convert a path to native (Windows) format.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git executable, possibly WSL Git.</param>
+        /// <returns>Path in in Windows format with native file separators.</returns>
+        public string GetPathForNative(string path) => PathUtil.GetNativePath(path, _wslDistro);
```
RussKie, line 182: This must be added to `IGitModule` as well.

GitCommands/PathUtil.cs
```diff
@@ -159,7 +160,93 @@ private static string ResolveRelativePath(string path, string relativePath)
...
+        }
+
+        /// <summary>
+        /// Convert a path for the Git executable.
+        /// For native (Windows) the path is the same, for WSL Git the path may be transformed.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git Extensions native (Windows) executable.</param>
+        /// <param name="wslDistro">Name of the distro or empty for non WSL paths</param>
+        /// <returns>Posix path if not a WSL distro (Windows Git), WSL path for WSL Git.</returns>
+        public static string GetRepoPath(string? path, string wslDistro)
```
RussKie, line 194: ❗  The annotation is incorrect: if `path` is nullable - this returns `null` on line:198. 
All callers of this method assume the return isn't null, and subsequently may result in NREs.

GitCommands/PathUtil.cs
```diff
@@ -159,7 +160,93 @@ private static string ResolveRelativePath(string path, string relativePath)
...
+            return path.ToPosixPath();
+        }
+
+        /// <summary>
+        /// Convert a path to native (Windows) format.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git executable, possibly WSL Git.</param>
+        /// <param name="wslDistro">Name of the distro or empty for non WSL paths.</param>
+        /// <returns>Path in in Windows format with native file separators.</returns>
+        public static string GetNativePath(string path, string wslDistro)
```
RussKie, line 232: Either incorrect nullable annotations, or absent null checks and redundant checks below.

GitCommands/PathUtil.cs
```diff
@@ -159,7 +160,93 @@ private static string ResolveRelativePath(string path, string relativePath)
...
-            return path.ToLower().StartsWith(@"\\wsl$\");
+            return path.ToLower().StartsWith(WslPrefix);
+        }
+
+        /// <summary>
+        /// Get the name of the distribution (like "Ubuntu-20.04") for WSL2 paths.
+        /// </summary>
+        /// <param name="path">Directory path.</param>
+        /// <returns>Name of the distro or empty for non WSL2 paths</returns>
+        public static string GetWslDistro(string path)
```
RussKie, line 171: Either incorrect nullable annotations, or absent null checks.

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -533,7 +533,8 @@ private bool HandlePushOnExit(ref bool isError, FormProcess form)
                 {
                     if (!form.ProcessArguments.Contains(" -f ") && !form.ProcessArguments.Contains(" --force"))
                     {
-                        Trace.Assert(form.ProcessArguments.StartsWith("push "), "Arguments should start with 'push' command");
+                        // Note that e.g. WSL may add other arguments prior to the actual command so push may not be first.
```
RussKie, line 536:
```suggestion
                        // Note that WSL may add other arguments prior to the actual command so "push" may not be first.
```

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -533,7 +533,8 @@ private bool HandlePushOnExit(ref bool isError, FormProcess form)
...
+                        Debug.Assert(form.ProcessArguments.Contains("push "), "Arguments should start with 'push' command");
```
RussKie, line 537: Is the comment still valid?

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: Adding Debug.Assert

GitCommands/Git/Executable.cs
gerhardol: ok

GitCommands/Git/GitModule.cs
gerhardol: ok

GitCommands/Git/GitModule.cs
gerhardol: I had planned to not add this to the interface, changing.

GitCommands/PathUtil.cs
gerhardol: Changing to not return null

GitCommands/PathUtil.cs
gerhardol: Changing to not return null

GitCommands/PathUtil.cs
gerhardol: I believe this is OK, not called with null and not returning null.
If path is null, an empty string is returned still, so `string? path` is possible

GitUI/CommandsDialogs/FormPush.cs
gerhardol: ok

GitUI/CommandsDialogs/FormPush.cs
gerhardol: Yes, the extra arguments (see first comment) is mostly hidden for the user

GitCommands/PathUtil.cs
RussKie: We either need to annotate the return as `string?` or add `_ = path ?? throw new ANE(nameof(path))`

GitCommands/PathUtil.cs
RussKie: We either need to annotate the args as `string?` or add `_ = path ?? throw new ANE(nameof(path))`

GitCommands/PathUtil.cs
gerhardol: if path is null (despite annotation), GetWslDistroLength() returns -1 and the method returns ""

RussKie: So if `path` _can_ be `null`, then it must be annotated as `string? path`.
If there's no code path to return a `null` then the return is `string`, else we must annotate it as `string?`.

GitCommands/PathUtil.cs
gerhardol: The args are not null and should not be nullable
I will add a null check for path still

GitCommands/PathUtil.cs
gerhardol: If path is null, it is handled, but not expected why annotation should be string.

GitCommands/PathUtil.cs
RussKie, line 194:
```suggestion
        public static string GetRepoPath(string? path, string? wslDistro)
```

GitCommands/PathUtil.cs
```diff
@@ -159,7 +160,93 @@ private static string ResolveRelativePath(string path, string relativePath)
...
+            return path.ToPosixPath();
+        }
+
+        /// <summary>
+        /// Convert a path to native (Windows) format.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git executable, possibly WSL Git.</param>
+        /// <param name="wslDistro">Name of the distro or empty for non WSL paths.</param>
+        /// <returns>Path in in Windows format with native file separators.</returns>
+        public static string GetNativePath(string path, string wslDistro)
```
RussKie, line 232:
```suggestion
        public static string GetNativePath(string? path, string? wslDistro)
```

GitCommands/PathUtil.cs
RussKie, line 171:
```suggestion
        public static string GetWslDistro(string? path)
```

GitCommands/PathUtil.cs
```diff
@@ -159,7 +160,93 @@ private static string ResolveRelativePath(string path, string relativePath)
...
+        /// Get the name of the distribution (like "Ubuntu-20.04") for WSL2 paths.
+        /// </summary>
+        /// <param name="path">Directory path.</param>
+        /// <returns>Name of the distro or empty for non WSL2 paths</returns>
+        public static string GetWslDistro(string path)
+        {
+            int distroLen = GetWslDistroLength(path);
+            return distroLen <= 0 ? "" : path.Substring(WslPrefix.Length, distroLen);
+
+            static int GetWslDistroLength(string path)
```
RussKie, line 176:
```suggestion
            static int GetWslDistroLength(string? path)
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +74,21 @@ public interface IGitModule
...
+        /// <param name="path">Path as seen by the Git Extensions native (Windows) executable.</param>
+        /// <returns>Posix path if Windows Git, WSL path for WSL Git.</returns>
+        public string GetPathForGitExec(string? path);
+
+        /// <summary>
+        /// Convert a path to native (Windows) format.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git executable, possibly WSL Git.</param>
+        /// <returns>Path in in Windows format with native file separators.</returns>
+        public string GetPathForNative(string path);
```
RussKie, line 90:
```suggestion
        public string GetPathForNative(string? path);
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +74,21 @@ public interface IGitModule
         /// <summary>Indicates HEAD is not pointing to a branch (i.e. it is detached).</summary>
         bool IsDetachedHead();
 
+        /// <summary>
+        /// Convert the path for the Git executable.
+        /// For WSL Git, the path will be adjusted.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git Extensions native (Windows) executable.</param>
+        /// <returns>Posix path if Windows Git, WSL path for WSL Git.</returns>
```
RussKie, line 82: Does it return an empty string for an invalid path or throws?

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +74,21 @@ public interface IGitModule
...
+        public string GetPathForGitExec(string? path);
+
+        /// <summary>
+        /// Convert a path to native (Windows) format.
+        /// </summary>
+        /// <param name="path">Path as seen by the Git executable, possibly WSL Git.</param>
+        /// <returns>Path in in Windows format with native file separators.</returns>
```
RussKie, line 89: Does it return an empty string for an invalid path or throws?

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +74,21 @@ public interface IGitModule
         /// <summary>Indicates HEAD is not pointing to a branch (i.e. it is detached).</summary>
         bool IsDetachedHead();
 
+        /// <summary>
+        /// Convert the path for the Git executable.
+        /// For WSL Git, the path will be adjusted.
```
RussKie, line 79:
```suggestion
        ///  Returns the path to the Git executable. For WSL Git the path is adjusted.
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
```diff
@@ -70,6 +74,21 @@ public interface IGitModule
...
+        /// </summary>
+        /// <param name="path">Path as seen by the Git Extensions native (Windows) executable.</param>
```
RussKie, line 81:
```suggestion
        /// <param name="path">The path as seen by the Git Extensions native (Windows) executable.</param>
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
RussKie, line 82:
```suggestion
        /// <returns>The POSIX path for Windows Git, WSL path for WSL Git.</returns>
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
RussKie, line 89:
```suggestion
        /// <summary>
        ///  Returns the path in Windows native format.
        /// </summary>
        /// <param name="path">The path as seen by the Git executable, possibly WSL Git.</param>
        /// <returns>The path in the Windows format with native file separators.</returns>
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
RussKie, line 90: Would it makes sense to name these as follows?
```cs
        public string GetDistroPath(string? path);
        public string GetWindowsPath(string? path);
```

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Posix unless identified as wsl

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: I still have a dream to support more than windows so I prefer the current name.

RussKie: When we get to support some other OS we can rename it, I promise :)

gerhardol: I can change. Will also change GitExec names, to be consistent. (Most Native names/comments).

Will write something in the comments like:
`With Windows path the GitExtensions.exe native path format is meant, this should be renamed if GE is running native on other OSes.`

RussKie:
:+1:

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Updated the naming to `GetWindowsPath` and `GetGitExecPath` (the last is more descriptive to me, the distro may use several Git executables).

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Rewrote the documentation, slightly different

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Rewrote the documentation, slightly different

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Rewrote the documentation, slightly different

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Empty if null path, set path separators to Windows if wslDistro is empty/null, otherwise try to convert the known path formats in WSL to Windows

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: ok

GitCommands/PathUtil.cs
gerhardol: ok

GitCommands/PathUtil.cs
gerhardol: ok

GitCommands/PathUtil.cs
gerhardol: ok

GitCommands/PathUtil.cs
gerhardol: ok

Plugins/GitUIPluginInterfaces/IGitModule.cs
gerhardol: Rewrote the documentation, slightly different

RussKie (approved):

andrej-schaefer:
Great feature thanks,
have anyone an idea how to store the passphrase for the SSH permanently

Tried a lot using keychain, but it seems when using git command with wsl it dont use the keychain 
workaround was to go with putty again by using wsl.localhost instead of wsl$ for the path
someone at the top have written "should it be detected automatically or be configured", here is a reason why it should stay configurable

maybe i missing some config, but was not able to find any solution for now

gerhardol:
See discussion in https://github.com/gitextensions/gitextensions/issues/10274#issuecomment-1302644406
(avoid discussions in closed PRs)

The complexity for this feature had to be limited somehow, I am sure that other can improve on this solution like handle #10173 better.

PR #9703, opened 2021-10-31
revDiff: Separate group for both changed

gerhardol:
Follow up to #7923

## Proposed changes

PR for discussion

If two commits are selected, #7923 (and more) added more diffs than just the diff between the two commits.
The extra categories can be disabled with ShowDiffForAllParents.

This helps me in checking differences between branches, for merging etc.
However, I am often interested in changes done in both branches that are not identical.
In the example in the screenshots it is `ab` that is modified in both that is most interesting, there are conflicts there.
`a` and `b` have unique changes in each of the branches.
`c` has the same change in both.

The following adds yet another section with the "conflict" changes.

Yet another category seem a little much. The files in the first category appears in "Both changed" and the two "Unique diff with BASE" and is not really needed (except that files changed in both branches are in all  three lists).
But removing this category may confuse some users.

If the category headers had some visual indication of the number of changes, the categories could be collapsed by default. That is probably a separate PR but may be discussed here.

Some possibilities:
* Add another category as in this PR. If so, the category need to be named better. What is proposed. 
* Collapse the "duplicate" (current first group) by default. (Of course can `ShowDiffForAllParents` be used to show only the diff). Remembering the collapse state in a session would be good use but requires a visual indicator of number of changes and is a separate PR.
* A new setting in addition to `ShowDiffForAllParents` that skips the common diff. (I say no, too many settings already)

### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/TranslatedStrings.cs
+++ b/GitUI/TranslatedStrings.cs
@@ -77,6 +77,7 @@ namespace GitUI
         private readonly TranslationString _diffWithParent = new("Diff with a/");
         private readonly TranslationString _diffBaseToB = new("Unique diff BASE with b/");
         private readonly TranslationString _diffCommonBase = new("Common diff with BASE a/");
+        private readonly TranslationString _diffBothChanged = new("Both changed to BASE");
         private readonly TranslationString _diffRange = new("Range diff");
         private readonly TranslationString _combinedDiff = new("Combined diff");
 
@@ -208,6 +209,7 @@ namespace GitUI
         public static string DiffWithParent => _instance.Value._diffWithParent.Text;
         public static string DiffBaseToB => _instance.Value._diffBaseToB.Text;
         public static string DiffCommonBase => _instance.Value._diffCommonBase.Text;
+        public static string DiffBothChanged => _instance.Value._diffBothChanged.Text;
         public static string DiffRange => _instance.Value._diffRange.Text;
         public static string CombinedDiff => _instance.Value._combinedDiff.Text;
         public static string ShowDiffForAllParentsText => _instance.Value._showDiffForAllParentsText.Text;
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -10600,6 +10600,10 @@ Select this commit to populate the full message.</source>
         <source>Unique diff BASE with b/</source>
         <target />
       </trans-unit>
+      <trans-unit id="_diffBothChanged.Text">
+        <source>Both changed to BASE</source>
+        <target />
+      </trans-unit>
       <trans-unit id="_diffCommonBase.Text">
         <source>Common diff with BASE a/</source>
         <target />
--- a/GitUI/UserControls/FileStatusDiffCalculator.cs
+++ b/GitUI/UserControls/FileStatusDiffCalculator.cs
@@ -167,8 +167,14 @@ namespace GitUI
 
             GitItemStatusNameEqualityComparer comparer = new();
             var commonBaseToAandB = allBaseToB.Intersect(allBaseToA, comparer).Except(allAToB, comparer).ToList();
+            var bothChangedAandB = allAToB.Intersect(allBaseToB, comparer).Intersect(allBaseToA, comparer).ToList();
 
             GitRevision revBase = new(baseRevGuid);
+            fileStatusDescs.Add(new FileStatusWithDescription(
+                firstRev: firstRev,
+                secondRev: selectedRev,
+                summary: TranslatedStrings.DiffBothChanged,
+                statuses: bothChangedAandB));
             fileStatusDescs.Add(new FileStatusWithDescription(
                 firstRev: revBase,
                 secondRev: selectedRev,


gerhardol:
@mdonatas as you originated this in #4154 do you have any comments?

mdonatas:
First of all, I think this is a move in the right direction. Much closer to what I had in mind when proposing this feature long time ago. For two commit diff I'd have this new view expanded and other 3 views collapsed. I tried to share original idea in this video https://www.loom.com/share/a1114fe3716e4a4db86398e2d4129233 (commenting with screenshots would've taken too much time :D )

Can someone who genuinely finds current `master` implementation useful comment on the scenarios where it helps? In what workflow is it useful to see all the changes between `base-to-a` and `base-to-b` and where looking at those full lists by manually selecting each branches' diff would become harder? This is a somewhat loaded question as if you watch the video you'll know I don't like current implementation one bit :D

gerhardol:
> First of all, I think this is a move in the right direction. Much closer to what I had in mind when proposing this feature long time ago. For two commit diff I'd have this new view expanded and other 3 views collapsed. I tried to share original idea in this video https://www.loom.com/share/a1114fe3716e4a4db86398e2d4129233 (commenting with screenshots would've taken too much time :D )

Visualizing the number of changes in a group and collapsing goes together. Probably another PR that maybe should be done prior to this, or at least with a strong intention of including it after this.
 
> Can someone who genuinely finds current `master` implementation useful comment on the scenarios where it helps? In what workflow is it useful to see all the changes between `base-to-a` and `base-to-b` at once and where looking at those full lists separately by manually selecting each branches' diff would become harder? This is a somewhat loaded question as if you watch the video you'll know I don't like current implementation one bit :D

I had thought about something to compare branches and was inspired by your feature description but made my interpretation. You had done your implementation, but most of the work with a PR is to to take it the whole way (which includes steering other PRs) so it was not done your way...
I can change my mind though.

The main use I have is also compare what is done in separate branches, quite often my combined feature branches with the various feature branches. Merge conflicts are one part of it, but also what changes in general that has been done. That is where the BASE to A/B are useful.
In my use case, what I miss is primarily a list of the files changed in A and B, the data is there.
I do use the extra data about BASE->A or B.

There are three diffs (A->B, BASE->B, BASE->A) that can be used to calculated statuses (all, unique A change, unique B change, both AB changed, same AB change). 
Possible combinations: A->B (all, (unique B change), (unique A change), both AB changed, (no same AB change)), BASE->B (all, (no unique A), unique B, (both AB changed), same AB), BASE->A (all, unique A, (no unique B), (both AB changed), same AB). 
A->B-same is always empty
BASE->B-same and BASE->A-same are identical and called BASE->AB.
A->B-unique B is same as BASE->B-all
A->B-unique A is same as reverse BASE->A-all
So if I get it right, there are 9 more or less interesting views.

This is hard to explain to users and in a UI, what is it that you see?
A->B-all is obvious, should always be first and be expanded by default. (If there is a way to hide this, it user custom and later).
Excluded below.

Current implementation is A->B-all, BASE->B-uniqueB, BASE->A-uniqueA, BASE->AB-same, The PR adds A->B-both
Your implementation is A->B-all, BASE->B-both, BASE->A-both, A->B-both

All 9 or so views should not be displayed by default.
Can we agree on what to show?
The filters could maybe be applied within diff but I often want to see more than one filter at a time (and know which group a file belongs to).
Each diff+filter could be a separate group that is enabled individually.

mstv:
> The following adds yet another section with the "conflict" changes.
> Yet another category seem a little much. The files in the first category appears in "Both changed"

This adds useful information. 👍
Perhaps "... changed to BASE in both" is a better name (with "..." as a kind of indentation) because it is just a filtering of the first category.
@mdonatas likes to filter the other categories in the same way.
Adding all these subcategories would overload the list IMO.
We already have the regex filter. So I propose to add a filter button there, which is applied to all categories and displays the number of concurrently changed files (tooltip e.g. "Filter to files changed to BASE in both" /  "... concurrently").

(I use all the existing categories in my daily work. Please do not make assumptions what is selected as A or B.)

(Whether there's a real merge conflict, is not easy to predict. You will know when you try the merge / rebase, which must be done anyway sometime. I prefer to see conflicts as 3-way diff.)

> If the category headers had some visual indication of the number of changes, the categories could be collapsed by default. That is probably a separate PR but may be discussed here.

Prepending the number of files would be good.
I would collapse the categories only if they occupy more than ~80% of the visible height. This will allow to get most of the information by scrolling without the need for clicking to expand short lists or to collapse crowded ones.

gerhardol:
This PR will not be merged, keeping it open to discuss other solutions.

> We already have the regex filter. So I propose to add a filter button there, which is applied to all categories and displays the number of concurrently changed files (tooltip e.g. "Filter to files changed to BASE in both" / "... concurrently").

If this is changed, probably the separate Unique/Common BASE->A/B should be merged.
The diffs are easier to explain.

The different filters/views that are interesting
* Files with unique changes in the two heads
* Files only changed in one of the HEADs
* Files with the same changes

A global filter is one way but will only show one perspective at a time.
Some filters are not interesting for some diffs. A->B common is empty and the diff will always be empty which is kind of meaningless to filter, but may also be confusing.

An alternative to a filter is to use modified icons (color or bold could be used too). This is always available and could be sortable in addition to the name/type.
This could be overlaying the current icons with `=` (same change), `!` (unique) or `A`/`B` (only).
If the diff is big, you have to scroll some to see files in the different diffs (primarily applicable for `!`).
This is my current suggestion.

> Prepending the number of files would be good. I would collapse the categories only if they occupy more than ~80% of the visible height. This will allow to get most of the information by scrolling without the need for clicking to expand short lists or to collapse crowded ones.

If there is a filter both the current and the total files should be displayed. 
Details about the filter/view could be shown in a tooltip, like number of common/unique files.

With a view (changed icons), I do not see this to be required,
The first part of this is maybe just prepending the group name with `[x]` with number of changes in x.

gerhardol:
Proof of concept
(I have pushed some preliminary changes to a private branch)

image

PR #9708, opened 2021-11-05
RevisionGridControl: prevent exception when switching repository

pmiossec:
Fixes nothing, just improve UX in DEBUG inside Visual Studio

## Proposed changes

- Behavior is not changed, the exception is already try/catched
- This check just prevent the exception to be raised which is really boring when debugging the application

## Test methodology 

- In debug in VS, switch the opened repository

## Test environment(s) 

- Git Extensions 33.33.33
- Build e61325736ad63a00a31f3193bd91a9e946f2b810
- Git 2.28.0.windows.1 (recommended: 2.33.0 or later)
- .NET 5.0.11
- DPI 192dpi (200% scaling)

## Merge strategy

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -579,6 +579,12 @@ namespace GitUI
             {
                 _gridView.Select();
 
+                // Prevent exception when changing of reporsitory because the grid still contains no rows.
+                if (index >= _gridView.Rows.Count)
+                {
+                    return;
+                }
+
                 bool shallSelect;
                 bool wasSelected = _gridView.Rows[index].Selected;
                 if (toggleSelection)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -579,6 +579,11 @@ private void SetSelectedIndex(int index, bool toggleSelection = false)
             {
                 _gridView.Select();
 
+                if (index >= _gridView.Rows.Count)
+                {
+                    return;
+                }
```
RussKie, line 585: Can you please add a comment describing a situation that leads to this path?

RussKie (commented):
+0

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
pmiossec: DOne.

gerhardol (approved):

PR #9718, opened 2021-11-11
RevDiff: Count number of changed files per group

gerhardol:
Discussed in #9703

## Proposed changes

For FileStatusList, can have one or more groups

- Show number of changes in a group first in the group name. Change for all group headers, except RangeDiff.

- Collapse groups RevDiff groups if they do not fulfill the following:
  - All other than comparing branches with a common BASE (the only that has more than two groups, currently five)
  - First group
  - 5 or less files

@mstv wanted the last part to expand groups if they occupy less than 80%, but I want to keep it simple.
 
### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -931,9 +931,15 @@ namespace GitUI
                 ListViewGroup? group = null;
                 if (i.FirstRev is not null)
                 {
-                    group = new ListViewGroup(i.Summary)
+                    string name = i.Statuses.Count == 1 && i.Statuses[0].IsRangeDiff
+                        ? i.Summary
+                        : $"({i.Statuses.Count}) {i.Summary}";
+                    group = new ListViewGroup(name)
                     {
-                        CollapsedState = ListViewGroupCollapsedState.Expanded,
+                        // Collapse some groups for diffs with common BASE
+                        CollapsedState = i.Statuses.Count <= 7 || GitItemStatusesWithDescription.Count < 3 || i == GitItemStatusesWithDescription[0]
+                            ? ListViewGroupCollapsedState.Expanded
+                            : ListViewGroupCollapsedState.Collapsed,
                         Tag = i.FirstRev
                     };
 


mstv:
👍

Round parentheses should be used in order to be consistent with the `Commit` button etc.

I'd set the file limit to 7 (UI guide) or rather 12 (because I find the "7 items rule" overcautious / too restrictive).

gerhardol:
> Round parentheses should be used in order to be consistent with the `Commit` button etc.

I found it harder to understand that (8) was count as it is first, maybe change Commit button too?

> I'd set the file limit to 7 (UI guide) or rather 12 (because I find the "7 items rule" overcautious / too restrictive).

Which UI Guide?
7 seem OK to me

mstv:
> I found it harder to understand that (8) was count as it is first, maybe change Commit button too?

Perhaps yes. I don't have a strong preference - just for consistency.
With filters, it will become "(8/8)", which is quite common.
We had a discussion regarding the count on the Commit button, branches in menu items and window title etc. ~ 1 year ago IIRC.

mstv:
> Which UI Guide?
7 seem OK to me

This seems to be / have been an empirical, generic rule regarding visual perception, which I heard / read of from different sources years ago.

RussKie:
Good idea. I suggest using the subheading in the listview group instead of
overloading the header.

RussKie:
Hit send too fast.
In the subheading we can have text, sometimes along the lines of "Files
changed: 1,234"

gerhardol:
> 
> Hit send too fast. In the subheading we can have text, sometimes along the lines of "Files changed: 1,234"

The header height is a little high, it did was not well received with Win 11

image

gerhardol:
> 
> Round parentheses should be used in order to be consistent with the `Commit` button etc.
> 
> I'd set the file limit to 7 (UI guide) or rather 12 (because I find the "7 items rule" overcautious / too restrictive).

Updated (not screenshot)

RussKie:
If we make the text a little longer it _may_ look a little better. 
There's also [`Footer`](https://docs.microsoft.com/dotnet/api/system.windows.forms.listviewgroup.footer) property available, see if it can be utilised.

gerhardol:
> If we make the text a little longer it _may_ look a little better. There's also [`Footer`](https://docs.microsoft.com/dotnet/api/system.windows.forms.listviewgroup.footer) property available, see if it can be utilised.

For me it is the height in general that is the issue, I doubt that will be considered a good change.
So same with a footer.

mstv (approved):
Thank you

gerhardol:
If the prepending "(<number>)" to "Diff with" is confusing, this could be applied to just branch diffs (with 3 or more groups, so merge diffs with combined diffs will be included too).
(But I like to see the actual number of changes also for simple diffs.)

RussKie:
I won't stand in a way. I don't have a strong opinion about this (at least yet 😄).

gerhardol:
In that case, plan to merge this just after #9720 (in a day and a half or so).

PR #9720, opened 2021-11-11
BASE diff: Icons for unique changes

gerhardol:
If a diff has a common BASE commit show icons that
indicates how the files have changed in first (A) and
second (B) compared to BASE.

Discussed in #9703 

## Proposed changes

This tweaks how the differences between two commits with a common BASE is displayed.
- Instead of separating the diffs, display the groups as "pure" diffs without manipulations. The current groups may not be obvious for all users.
- Show where changes occurred; in first (A), second (B), both same (=) or both unequal (!) with overlayed icons.
Especially the both unique is of interest, what was first suggested in #9703.
- Group the changes if the files are sorted by status
- Change descriptions for Rest/Diff revisions

Some notes what this is not:
- No filter of files as suggested in https://github.com/gitextensions/gitextensions/pull/9703#issuecomment-962691415. Could be added too.
- No separate groups as discussed in #9703 - the needs for all would increase the list to nine groups or so, too much.
- No sorting of the files by this "diff status" only in combination of name or extension. The additional way to group the sorting should only be visible in some situations, which would be complex.
- No "overlays" but separate icons for Added/Modified/Removed/Renamed/Copied only; not submodules as that would add a lot of extra icons.
- The icons could look better, I am not a designer
- Moving RangeDiff into the A->B group (if done, it should always be last there)

The Git documentation uses mostly first/second commit, but A/B can be seen in some commands. I used A/B in the icons as that was easier to identify than 1/2.
 
Documentation will be updated after merge

## Screenshots 

Sorted by icon

### Before

image

image

image

### After

FileStatusCopiedOnlyA FileStatusCopiedOnlyB FileStatusCopiedSame FileStatusCopiedUnequal

FileStatusModifiedOnlyA FileStatusModifiedOnlyB FileStatusModifiedSame FileStatusModifiedUnequal

FileStatusRemovedOnlyA FileStatusRemovedOnlyB FileStatusRemovedSame FileStatusRemovedUnequal

FileStatusRenamedOnlyA FileStatusRenamedOnlyB FileStatusRenamedSame FileStatusRenamedUnequal 

image

image

image

## Test methodology 

Manual.
I will try to add some tests.

## Merge strategy

----


mstv:
👍 This improves this feature significantly.
Its importance and the first use (without prior reading the description) raised wishes:
- "unique" feels misleading to me. It is rather an "unequal" or a "concurrent" change. "A" and "B" are the unique changes IMO. That's why I think `≠` will be clearer as overlay.
- Copied & renamed files should have an overlay, too. Though the need for this will decrease with filtering. A real overlay will be possible when we switch to a TreeView.
- Filter could be three toggles: concurrent/unequal, orphan, equal
- The icons LGTM despite the anti-aliasing. Just remove the nasty gray pixels, please.
- Upper case "A:" & "B:" should not be mixed with "a/" & "b/" in the diff.
- The BASE revision is not identified anymore since the catagory "common" has been removed. It could be added to the RangeDiff, e.g. "Range diff 3↑ 2↓ to BASE 12345678: master".

> Moving RangeDiff into the A->B group (if done, it should always be last there)

👍

gerhardol:
>     * "unique" feels misleading to me. It is rather an "unequal" or a "concurrent" change. "A" and "B" are the unique changes IMO. That's why I think `≠` will be clearer as overlay.

Unequal is fine for me (then the reverse alphabetic order is also a good sort order).
 `≠`  vs `=` can be hard to see in the icons though.
Unique/Unequal is only used in the code now (something like that should be added to the documentation though).
 
>     * Copied & renamed files should have an overlay, too. Though the need for this will decrease with filtering. A real overlay will be possible when we switch to a TreeView.

I rather prefer to wait with that, also for submodules. At least until the existing icons are considered OK, it takes some time to redesign the icons (including remove size with TweakPng).

>     * Filter could be three toggles: concurrent/unequal, orphan, equal

Yes, also not now for me, I do not have a clear view how the filter should be triggered.

>     * The icons LGTM despite the anti-aliasing. Just remove the nasty gray pixels, please.

Which pixels? For `A` the normal sized pixel looks better with the feeble attempt for aliasing.

>     * Upper case "A:" & "B:" should not be mixed with "a/" & "b/" in the diff.

Can you elaborate?
`A`/`B` is only used on the icons. They are the same as a/b in commands.
 
>     * The BASE revision is not identified anymore since the catagory "common" has been removed. It could be added to the RangeDiff, e.g. "Range diff 3↑ 2↓ to BASE 12345678: master".

Good idea, then all three involved commits are visible.

> > Moving RangeDiff into the A->B group (if done, it should always be last there)

Later, I assume

mstv:
> Which pixels? For A the normal sized pixel looks better with the feeble attempt for aliasing.

The "A" is OK for me.
The current "B" could be mixed with "8" easily.
"=" has triggered my request.
"!" should have one column of black pixels only - if it is not replaced with "≠" with perpendicular vertical line - perhaps in red.

>>     * Upper case "A:" & "B:" should not be mixed with "a/" & "b/" in the diff.

> Can you elaborate?
> `A`/`B` is only used on the icons. They are the same as a/b in commands.

image

must be "Diff BASE with b/xxxxxxx: tmp/a" as before this PR, since BASE is "a/" in the diff.

gerhardol:
Changed according to my comments in https://github.com/gitextensions/gitextensions/pull/9720#issuecomment-967707962

> must be "Diff BASE with b/xxxxxxx: tmp/a" as before this PR, since BASE is "a/" in the diff.

Did not change this. I see `a` as the identification here.
I have probably argued for this to change before, but `Second: a/sha desc` is described in the diff (and reset) submenus

RussKie:
What does `!` stand for? Unique? If yes, I'm not finding it obvious...

> * "unique" feels misleading to me. It is rather an "unequal" or a "concurrent" change. "A" and "B" are the unique changes IMO. That's why I think `≠` will be clearer as overlay.

👍 Perhaps make  `≠` red so it stand out more?

. . . 

To make differences between "A" and "B" more obvious, perhaps have "A" in the top LHS corner, and keep "B" in the bottom RHS corner... 🤔

gerhardol:
Updated screenshot

> What does `!` stand for? Unique? If yes, I'm not finding it obvious...

Attention or maybe short for `!=`.

> > "A" and "B" are the unique changes IMO. That's why I think `≠` will be clearer as overlay.
> 
> 👍 Perhaps make `≠` red so it stand out more?
> 
> . . .
> 
> To make differences between "A" and "B" more obvious, perhaps have "A" in the top LHS corner, and keep "B" in the bottom RHS corner... 🤔

The "overlays" should be subtle in my opinion, not breaking too much with the existing icons.
The overlays are like 5x5 pixels (which is still about a quarterof the available space) and I cannot get a good looking  `≠` with that.

A designer can make a better job....

RussKie:
> > What does `!` stand for? Unique? If yes, I'm not finding it obvious...
> 
> Attention or maybe short for `!=`.

Whilst some users (who likely come with C-language background) may understand this, I think a lot of other users may not find this obvious.



> A designer can make a better job....

Yes, probably can... Do you have any designers in mind?

gerhardol:
> 
> 
> > > What does `!` stand for? Unique? If yes, I'm not finding it obvious...
> > 
> > 
> > Attention or maybe short for `!=`.
> 
> Whilst some users (who likely come with C-language background) may understand this, I think a lot of other users may not find this obvious.

Well is primarily attention for me

> > A designer can make a better job....
> 
> Yes, probably can... Do you have any designers in mind?

No

mdonatas:
Tried running your branch and I like what you are doing! This is so much better 👍 

As discussed in #9703 there are several other views which aren't shown now. I think a submenu such as for `Sort by...` listing all options would let users customize what's important for their workflows. What do others think?

image

gerhardol:
> As discussed in #9703 there are several other views which aren't shown now. I think a submenu such as for `Sort by...` listing all options would let users customize what's important for their workflows. What do others think?

I believe a quick way to apply filters is the solution, not changing the diffs.
The other diffs were useful but hard to explain.
https://github.com/gitextensions/gitextensions/pull/9720#issuecomment-967655556
These filters could be applied from these submenus though.
(I still suggest that is done in a follow up to this PR)

gerhardol:
> If you want to emphasize the first / second selected commit ("A" / "B"), the categories could be:
- Diff B with A a/hashA: identA
- Diff BASE with B b/hashB: identB
- Diff BASE with A b/hashA: identA

Then a/ and b/ should be removed in the last two headers.
I suggest the first header is changed.

---

> Proposal for overlays (indeed, `≠` is difficult and not perfect):

I do not like `≠` and if it has to be red I still prefer `!`.
Can you push your changes?

RussKie:
> Proposal for overlays (indeed, `≠` is difficult and not perfect):
> 
> FileStatusModifiedOnlyA FileStatusModifiedOnlyB FileStatusModifiedSame FileStatusModifiedUnequal

I like that

gerhardol:
Adjusted the text and icons.
Keeping `!` but made it red
Screenshot updated

gerhardol:
Adjusted descriptions for Rest/Diff and the first Diff group heading.
If this is approved, I can update copied/renamed icons too (but that is the last time I chenge the icons).

To move RangeDiff into A->B group with keeping the diff last was quite simple, but I want to merge #9718 first (to be able to find the RangeDiff in a large diff).

image

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -30,8 +30,8 @@ public partial class RevisionDiffControl : GitModuleControl
             new("Are you sure you want to delete the selected file(s)?");
         private readonly TranslationString _deleteFailed = new("Delete file failed");
         private readonly TranslationString _multipleDescription = new("<multiple>");
-        private readonly TranslationString _selectedRevision = new("Second: b/");
-        private readonly TranslationString _firstRevision = new("First: a/");
+        private readonly TranslationString _selectedRevision = new("Second(B): ");
+        private readonly TranslationString _firstRevision = new("First(A): ");
```
mstv, line 34:
```suggestion
        private readonly TranslationString _selectedRevision = new("Second: B ");
        private readonly TranslationString _firstRevision = new("First: A ");
```
This matches the FileStatusList headings better.

GitCommands/Git/DiffBranchStatus.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Text;
+using System.Threading.Tasks;
+using GitUIPluginInterfaces;
+using Microsoft;
+using Microsoft.VisualStudio.Threading;
+
```
mstv, line 7: Remove?

GitCommands/Git/DiffBranchStatus.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace GitCommands
+{
+     public enum DiffBranchStatus
+    {
```
mstv, line 11:
```suggestion
    public enum DiffBranchStatus
    {
```

mstv (commented):
>  (but that is the last time I change the icons)

I almost did not dare to ask: The equal sign should have one pixel less at the left, shouldn't it?

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: OK

GitCommands/Git/DiffBranchStatus.cs
gerhardol: yes

GitCommands/Git/DiffBranchStatus.cs
gerhardol: yes

gerhardol:
> > (but that is the last time I change the icons)
> 
> I almost did not dare to ask: The equal sign should have one pixel less at the left, shouldn't it?

Of course you can ask - I can refuse.

This was intentional, to see better but I can change.
Pushing fixup.

mstv (approved):
Thank you

gerhardol:
Added icons for copied/renamed too
Updated descriptions
I cannot think of any tests

If no other comments, I plan to squash merge in a few days

RussKie:
LGTM

PR #9729, opened 2021-11-14
FormBrowse init optimizations

gerhardol:
## Proposed changes

@RussKie has refactored FormBrowse init and filter handling recently.
After this handling some logic like InternalInitialize() was running several times.
This included a few Git commands too, delaying the startup slightly.
There are some test instabilities for FormBrowse, this is an attempt to at least improve investigation.

I initially got som stability issues with this change when testing in isolation but this works better when restructuring commits.
`'Loading Revisions' didn't finish in 100 iterations` can still occur

## Test methodology 

There are tests for this already

## Merge strategy

Rebase merge

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -311,14 +311,11 @@ namespace GitUI.CommandsDialogs
             _aheadBehindDataProvider = new AheadBehindDataProvider(() => Module.GitExecutable);
             toolStripButtonPush.Initialize(_aheadBehindDataProvider);
             repoObjectsTree.Initialize(_aheadBehindDataProvider, branchFilterAction: ToolStripFilters.SetBranchFilter, RevisionGrid, RevisionGrid, RevisionGrid);
-            revisionDiff.Bind(RevisionGrid, fileTree, RequestRefresh);
-            fileTree.Bind(RevisionGrid, RequestRefresh);
-
-            // We're launching the main form, and whilst the module has been set for us we still need to formally switch to it
-            SetGitModule(this, new GitModuleEventArgs(new GitModule(Module.WorkingDir)));
-
+            revisionDiff.Bind(RevisionGrid, fileTree, RefreshGitStatusMonitor);
+            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor);
             RevisionGrid.ResumeRefreshRevisions();
 
+            // Application is init, the repo related operations are triggered in OnLoad()
             return;
 
             void InitCountArtificial(out GitStatusMonitor gitStatusMonitor)
@@ -442,20 +439,20 @@ namespace GitUI.CommandsDialogs
             base.OnApplicationActivated();
         }
 
-        // Contains app init logics ONLY. All repo-specific logic must be placed UICommands.PostRepositoryChanged handler.
         protected override void OnLoad(EventArgs e)
         {
             _formBrowseMenus.CreateToolbarsMenus(ToolStripMain, ToolStripFilters, ToolStripScripts);
 
-            HideVariableMainMenuItems();
             RefreshSplitViewLayout();
             LayoutRevisionInfo();
             SetSplitterPositions();
-            InternalInitialize(false);
 
             base.OnLoad(e);
 
             _formBrowseDiagnosticsReporter.Report();
+
+            // All app init is done, make all repo related similar to switching repos (like triggering RefreshRevisions())
+            SetGitModule(this, new GitModuleEventArgs(new GitModule(Module.WorkingDir)));
         }
 
         protected override void OnActivated(EventArgs e)
@@ -572,22 +569,28 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
-            RequestRefresh();
-
-            if (_dashboard is null || !_dashboard.Visible)
+            bool isDashboard = string.IsNullOrEmpty(Module.WorkingDir) || (_dashboard?.Visible ?? false);
+            if (isDashboard)
             {
-                revisionDiff.RefreshArtificial();
-                RevisionGrid.ForceRefreshRevisions();
-                RevisionGrid.IndexWatcher.Reset();
+                // Explicit call: Title is normally updated on RevisionGrid filter change
+                _appTitleGenerator.Generate();
 
-                InternalInitialize(true);
+                // "Repo" related methods, creates _dashboard
+                InternalInitialize();
+
+                return;
             }
 
-            toolStripButtonPush.DisplayAheadBehindInformation(Module.GetSelectedBranch());
+            Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
+            RevisionGrid.PerformRefreshRevisions();
+
+            InternalInitialize();
+
+            RefreshGitStatusMonitor();
+            revisionDiff.RefreshArtificial();
         }
 
-        private void RequestRefresh() => _gitStatusMonitor?.RequestRefresh();
+        private void RefreshGitStatusMonitor() => _gitStatusMonitor?.RequestRefresh();
 
         private void RefreshSelection()
         {
@@ -658,7 +661,7 @@ namespace GitUI.CommandsDialogs
         private void HideDashboard()
         {
             MainSplitContainer.Visible = true;
-            if (_dashboard is null || !_dashboard.Visible)
+            if (!_dashboard?.Visible ?? true)
             {
                 return;
             }
@@ -714,6 +717,7 @@ namespace GitUI.CommandsDialogs
                     {
                         if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
                         {
+                            _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
                             RefreshRevisions();
                         }
                     };
@@ -772,7 +776,7 @@ namespace GitUI.CommandsDialogs
             mainMenuStrip.Refresh();
         }
 
-        private void InternalInitialize(bool hard)
+        private void InternalInitialize()
         {
             toolPanel.SuspendLayout();
             toolPanel.TopToolStripPanel.SuspendLayout();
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -572,22 +570,33 @@ private void RefreshRevisions()
...
+                // Still loading, the method will be called again
+                return;
             }
 
-            toolStripButtonPush.DisplayAheadBehindInformation(Module.GetSelectedBranch());
+            RevisionGrid.ForceRefreshRevisions();
+
+            InternalInitialize();
+
+            RefreshGitStatusMonitor();
```
gerhardol, line 595: Let the separate git-status calls run after the UI is refreshed

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1800,7 +1810,6 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
 
                     // If we're applying custom branch or revision filters - reset them
                     RevisionGrid.DisableFilters();
-                    SetPathFilter(pathFilter: null);
```
gerhardol, line 1803: This triggers a Grid refresh, so this must not occur (see also comments for the commit). 

Maybe the `CanRefresh` check in `RequestRefresh` should at least ``Debug.Assert`
Maybe add `RevisionGrid.DisableRefreshRevisions()`/`RevisionGrid.ResumeRefreshRevisions()` in `GitSetModule` to permanently avoid this as the refresh is done in `UICommands_PostRepositoryChanged()`

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -572,22 +570,33 @@ private void RefreshRevisions()
                 return;
             }
 
-            _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
-            RequestRefresh();
```
gerhardol, line 576: Renamed to `RefreshGitStatusMonitor`

gerhardol (commented):
Some explanations to changes

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -465,17 +465,12 @@ public void DisableContextMenu()
...
         /// </summary>
         internal void ResumeRefreshRevisions()
         {
             --_updatingFilters;
             Debug.Assert(_updatingFilters >= 0, $"{nameof(ResumeRefreshRevisions)} was called without matching {nameof(SuspendRefreshRevisions)}!");
-
-            if (_updatingFilters == 0)
-            {
-                ForceRefreshRevisions();
-            }
```
RussKie, line 478: Coupled with above comment I'm finding this change the most puzzling. This was meant to guard against nested unbalanced suspensions (i.e., without a resume call), and execute a refresh only on the outer scope.
Now it feels like we may be returning into a realm where we may be calling multiple refreshes.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -311,14 +311,14 @@ public FormBrowse(GitUICommands commands, BrowseArguments args)
...
-            fileTree.Bind(RevisionGrid, RequestRefresh);
+            revisionDiff.Bind(RevisionGrid, fileTree, RefreshGitStatusMonitor);
+            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor);
+            RevisionGrid.ResumeRefreshRevisions();
 
             // We're launching the main form, and whilst the module has been set for us we still need to formally switch to it
+            // Also triggers RefreshRevisions()
             SetGitModule(this, new GitModuleEventArgs(new GitModule(Module.WorkingDir)));
 
-            RevisionGrid.ResumeRefreshRevisions();
```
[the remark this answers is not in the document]
gerhardol: `RevisionGrid.ResumeRefreshRevisions();` should maybe be renamed to `RevisionGrid.EnableRefreshRevisions();`
All refresh is done from SetGitModule() -> PostRepoChanged

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: See above, the refresh is always called explicitly after Suspend/Resume.

GitUI/CommandsDialogs/FormBrowse.cs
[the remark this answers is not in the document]
RussKie: > `RevisionGrid.ResumeRefreshRevisions();` should maybe be renamed to `RevisionGrid.EnableRefreshRevisions();`

The naming was inspired by [`SuspendLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.suspendlayout) and [`ResumeLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.resumelayout). There's also a [`PerformLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.performlayout) and I've been thinking maybe renaming `ForceRefreshRevisions` into `PerformRefreshRevisions`... 🤔 




> All refresh is done from SetGitModule() -> PostRepoChanged

Not quite. There are quite a few calls to `ForceRefreshRevisions` throughout, e.g. applying various filters, and this mechanism guarded against chained refreshes.

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: This is likely because of the changes to `ResumeRefreshRevisions`

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -572,22 +570,28 @@ private void RefreshRevisions()
...
-                InternalInitialize(true);
+                // "Repo" related methods, creates _dashboard
+                InternalInitialize();
+
+                return;
             }
 
-            toolStripButtonPush.DisplayAheadBehindInformation(Module.GetSelectedBranch());
+            Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
+            RevisionGrid.ForceRefreshRevisions();
```
RussKie, line 586: 💭  Generally speaking we should strive to avoid calling this API.

RussKie (commented):
In general the changes look sensible and positive, however I am not quite content with the changes to `ResumeRefreshRevisions` - I spent a considerable time checking various code paths (primarily in the context of filters) to reduce redundant refreshes.

gerhardol:
This PR is required to avoid that the left panel is init twice in #9734, and for #9735 that can save seconds opening a repo.
(After #9735 the tests seem to be more stable too.)

> In general the changes look sensible and positive, however I am not quite content with the changes to `ResumeRefreshRevisions` - I spent a considerable time checking various code paths (primarily in the context of filters) to reduce redundant refreshes.

I also have spent considerable time debugging and trying to understand what the intention was. The handling was not easy before, but the changed handling does not go all the way. I tried to explain the changes in my review, another try here:
The major change is to handle startup in the same way as a repo change, remove the special handling.

The handling after GE has started once is not changed much, just some adjustments.

`SetGitModule()` -> `PostRepositoryChanged()` -> `RefreshRevisions()`
This will then make all repo related changes including `RevisionGrid.EnableRefreshRevisions()`.

The startup had another path:

`FormBrowse()` DisableGuard; init all (including some repo related like init left panel as in #9734), `SetGitModule()` etc except `RevisionGrid.EnableRefreshRevisions()`, EnableGuard that triggers `RevisionGrid.EnableRefreshRevisions()`.
A few repo related then runs (again) in `OnLoad()` to get the init correctly. Especially `InternalInitialize()` does repo related setup.

This PR changes the start:

`FormBrowse()` DisableGuard; init avoiding repo related, EnableGuard then explicit `SetGitModule()` that will run repo related similar to at a repo switch, including `RevisionGrid.EnableRefreshRevisions()`.
`SetGitModule()` has the guard internally too. Before #9733 `RevisionGrid.EnableRefreshRevisions()` was actually run before  `PostRepositoryChanged()` (so in a tiny repo the revisions could have been refreshed twice).
`OnLoad()` has no repo related.

This change was probably not enough,  `SetGitModule()` should be called from `OnLoad()` instead.

> > `RevisionGrid.ResumeRefreshRevisions();` should maybe be renamed to `RevisionGrid.EnableRefreshRevisions();`
> 
> The naming was inspired by [`SuspendLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.suspendlayout) and [`ResumeLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.resumelayout). There's also a [`PerformLayout`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.performlayout) and I've been thinking maybe renaming `ForceRefreshRevisions` into `PerformRefreshRevisions`... 🤔 

These are now pure guards, the refresh is always done in post repo change. 

> > All refresh is done from SetGitModule() -> PostRepoChanged
> 
> Not quite. There are quite a few calls to `ForceRefreshRevisions` throughout, e.g. applying various filters, and this mechanism guarded against chained refreshes.

All revision reading related to repo changes is now in  `PostRepositoryChanged()`. 
There are of course refresh in other situations too. If there happen when updating the repo in  `SetGitModule()`, they are now ignored as there will be an explicit call.

RussKie:
I ran it locally with few different filters, and I couldn't observer any reentrancy I was worried about. So 👍  from me.
Proposed few API tweaks to compliment the change.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -825,11 +823,22 @@ private void ShowLoading(bool sync = true)
...
-        public void ForceRefreshRevisions()
+        /// <summary>
+        ///  Indicates whether the revision grid can be refreshed, i.e. it is not currently being refreshed
+        ///  or it is not in a middle of reconfiguration process guarded by <see cref="SuspendRefreshRevisions"/>
+        ///  and <see cref="ResumeRefreshRevisions"/>.
+        /// </summary>
+        public bool CanRefresh => !_isRefreshingRevisions && _updatingFilters == 0;
+
+        /// <summary>
+        ///  Queries git for the new set of revisions, and refreshed the grid.
```
gerhardol, line 834: Rename is of course fine, will just change this wording.
```suggestion
        ///  Queries git for the new set of revisions and refreshes the grid.
```

OK to squash and merge?
I would like to get #9735 ready to merge, that is both improving startup a lot and seem to improve test stability.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -582,7 +582,7 @@ private void RefreshRevisions()
             }
 
             Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
-            RevisionGrid.ForceRefreshRevisions();
+            RevisionGrid.RefreshRevisions();
```
gerhardol, line 585: This must be `PerformRefreshRevisions()` (that must be public again.

RussKie (approved):

PR #9734, opened 2021-11-16
LeftPanel was initiated twice at startup

gerhardol:
Regression in master

Based on #9729 (may be separated).
Ready for review (first commit is not to be considered) even if draft.

## Proposed changes

The structuring of FormBrowse in master has changed so the left panel is initiated both at creation and at  PostRepositoryChanged.
This could add a couple of seconds to the startup. (Especially at work with virus programs.)

Also eliminate a few Git calls when starting to Dashboard.
(More confusing than a problem.)

### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -153,7 +153,7 @@ namespace GitCommands
         }
 
         /// <inherit/>
-        public string WorkingDir { get; }
+        public string WorkingDir { get; init; }
 
         /// <summary>
         /// GitVersion for the default GitExecutable.
@@ -359,6 +359,13 @@ namespace GitCommands
         {
             get
             {
+                if (string.IsNullOrEmpty(WorkingDir))
+                {
+                    // No directory for this module, so the common directory must be empty too
+                    // (should not be retrieved from the GE start directory)
+                    return "";
+                }
+
                 // Get a cache of the common dir
                 // Lock needed as the command is called rapidly when creating the module
                 if (_gitCommonDirectory is not null)
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
@@ -212,17 +212,6 @@ namespace GitUI.BranchTreePanel
                 SubmoduleStatusProvider.Default.StatusUpdated += Provider_StatusUpdated;
             }
 
-            protected override Task OnAttachedAsync()
-            {
-                var e = _currentSubmoduleInfo;
-                if (e is not null)
-                {
-                    OnStatusUpdated(e);
-                }
-
-                return Task.CompletedTask;
-            }
-
             protected override Task<Nodes> LoadNodesAsync(CancellationToken token)
             {
                 return Task.FromResult(new Nodes(null));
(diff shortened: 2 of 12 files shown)


RussKie (commented):
The last three commits look sensible 👍

gerhardol:
Moved a commit from #9735 to this PR, rebased.

GitCommands/Git/GitModule.cs
```diff
@@ -359,6 +359,11 @@ public string GitCommonDirectory
         {
             get
             {
+                if (string.IsNullOrEmpty(WorkingDir))
+                {
+                    return "";
+                }
```
mstv, line 365: I need a comment here because the relation between `_gitExecutable` and `WorkingDir` is not obvious. Perhaps the property `WorkingDir` can be declared `init;`.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -200,15 +200,13 @@ protected virtual Task PostRepositoryChangedAsync()
             protected bool IsAttached { get; private set; }
             protected virtual bool SupportsFiltering { get; } = false;
 
-            public Task AttachedAsync()
+            public void Attached()
             {
                 IsAttached = true;
-                return OnAttachedAsync();
```
mstv, line 206: I think `OnAttached()` should be kept here.

GitCommands/Git/GitModule.cs
gerhardol: If the GitModule has no directory set, the command should not return any path, even if GE has been started from a Git repo.
So not really related to the executable.
This could be applied to other commands too (but not so many commands are executed when the GitModule has no working dir.)
Do you want a comment?

WorkingDir could be init, how could that improve.

mstv: > Do you want a comment?

Yes, would be good.

> WorkingDir could be init, how could that improve.

Then it will be clearer that `WorkingDir` will not change during runtime. Hence the result of this check will always be the same, i.e. will not have an inconsistent affect on repeated calls of this function.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
gerhardol: Yes
Removed the action in RepoObjectsTree.Nodes.Submodules.cs though

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -214,11 +214,6 @@ public SubmoduleTree(TreeNode treeNode, IGitUICommandsSource uiCommands)
 
             protected override void OnAttached()
             {
-                var e = _currentSubmoduleInfo;
-                if (e is not null)
-                {
-                    OnStatusUpdated(e);
-                }
             }
```
mstv: An empty override can be removed usually. This _is_ the case here since the default implementation is empty.

GitCommands/Git/GitModule.cs

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
gerhardol: Removed the empty function.
The other overrides seem OK to me, not changing now.

mstv (approved):

RussKie (commented):
👍

gerhardol:
Squ7ashed and rebased, rebase merge tomorrow

RussKie:
:+1:

PR #9735, opened 2021-11-16
Share GitRef() at startup

gerhardol:
requires #9734

## Proposed changes

In master, GitRefs() run 13+2 times during the startup when 1+2 would be sufficient. While some of these are due to a change in master handled in #9729, there are 9+1 in 3.5 (sidepanel many of them).
At work each of these may require several hundred ms. Even if they to some extent can run in parallel, the startuptime may increase with several seconds.

After #9734 most GitRef() calls are done from PostRepositoryChanged and the call can be shared.

## Screenshots 

Timing is when running in the debugger, much different from normal operation (except Git command running  time).

### Before

From #9734 (see that PR for current master).

image

### After

This contains a few commands running after the commands seen above.
image

## Test methodology 

Test is added

## Merge strategy

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
@@ -212,7 +212,7 @@ namespace GitUI.BranchTreePanel
                 SubmoduleStatusProvider.Default.StatusUpdated += Provider_StatusUpdated;
             }
 
-            protected override Task<Nodes> LoadNodesAsync(CancellationToken token)
+            protected override Task<Nodes> LoadNodesAsync(CancellationToken token, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
             {
                 return Task.FromResult(new Nodes(null));
             }
@@ -275,12 +275,13 @@ namespace GitUI.BranchTreePanel
 
                     if (_currentNodes is null)
                     {
-                        await ReloadNodesAsync(token =>
+                        // Module.GetRefs() is not used for Submodules
+                        await ReloadNodesAsync((token, _) =>
                         {
                             cts = CancellationTokenSource.CreateLinkedTokenSource(e.Token, token);
                             loadNodesTask = LoadNodesAsync(e.Info, cts.Token);
                             return loadNodesTask;
-                        }).ConfigureAwait(false);
+                        }, null).ConfigureAwait(false);
                     }
 
                     if (cts is not null && loadNodesTask is not null)
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -553,16 +553,20 @@ namespace GitUI.CommandsDialogs
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
-            this.InvokeAsync(RefreshRevisions).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await this.SwitchToMainThreadAsync();
+                RefreshRevisions(e.GetRefs);
+            }).FileAndForget();
 
-            ToolStripFilters.UpdateBranchFilterItems();
+            ToolStripFilters.UpdateBranchFilterItems(e.GetRefs);
             UpdateSubmodulesStructure();
             UpdateStashCount();
 
             revisionDiff.UICommands_PostRepositoryChanged(sender, e);
         }
 
-        private void RefreshRevisions()
+        private void RefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
         {
             if (RevisionGrid.IsDisposed || IsDisposed || Disposing)
             {
@@ -573,7 +577,7 @@ namespace GitUI.CommandsDialogs
             if (isDashboard)
             {
                 // Explicit call: Title is normally updated on RevisionGrid filter change
-                _appTitleGenerator.Generate();
+                Text = _appTitleGenerator.Generate();
 
                 // "Repo" related methods, creates _dashboard
                 InternalInitialize();
@@ -582,7 +586,7 @@ namespace GitUI.CommandsDialogs
             }
 
             Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
-            RevisionGrid.PerformRefreshRevisions();
+            RevisionGrid.PerformRefreshRevisions(getRefs);
 
             InternalInitialize();
 
@@ -718,7 +722,7 @@ namespace GitUI.CommandsDialogs
                         if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
                         {
                             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
-                            RefreshRevisions();
+                            RefreshRevisions(new FilteredGitRefsProvider(UICommands.GitModule).GetRefs);
                         }
                     };
 
--- a/Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
+++ b/Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
@@ -1,15 +1,19 @@
-﻿using System.ComponentModel;
+﻿using System.Collections.Generic;
+using System.ComponentModel;
 using System.Windows.Forms;
 
 namespace GitUIPluginInterfaces
 {
     public class GitUIEventArgs : CancelEventArgs
     {
+        private readonly IFilteredGitRefsProvider _refs;
+
         public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands)
             : base(cancel: false)
         {
             OwnerForm = ownerForm;
             GitUICommands = gitUICommands;
+            _refs = new FilteredGitRefsProvider(GitModule);
         }
 
         public IGitUICommands GitUICommands { get; }
@@ -17,5 +21,7 @@ namespace GitUIPluginInterfaces
         public IWin32Window? OwnerForm { get; }
 
         public IGitModule GitModule => GitUICommands.GitModule;
+
+        public IReadOnlyList<IGitRef> GetRefs(RefsFilter filter) => _refs.GetRefs(filter);
     }
 }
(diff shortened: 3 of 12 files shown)


gerhardol:
This is ready for review. The startup and repo switch is much improved by this PR.
Also, the tests seem to be slightly more stable. As the startup sequence is simplified, it should be simpler to investigate the tests.

gerhardol:
@RussKie @mstv Any comments on this PR?
It improves the performance when switching repos dramatically, I want to make a few similar changes in addition to this (with much smaller difference though).

RussKie (commented):
Overall I support the direction of your thought, but I'm not excited by `LazyGitRefsProvider`. I get a feeling it may bring hard to debug issues...
For quite some time I have been thinking how we could cache certain results, and whilst I haven't come up with anything concrete, my thinking was along the lines of a transparent caching in `GitModule` - i.e. results are cached or disposed of without any user input. 
I think we can rely on the file/folder watchers or git status watchers to detect when there is a change in the refs index and break the cache then. What do you think of this approach?

gerhardol:
> Overall I support the direction of your thought, but I'm not excited by `LazyGitRefsProvider`. I get a feeling it may bring hard to debug issues... For quite some time I have been thinking how we could cache certain results, and whilst I haven't come up with anything concrete, my thinking was along the lines of a transparent caching in `GitModule` - i.e. results are cached or disposed of without any user input. I think we can rely on the file/folder watchers or git status watchers to detect when there is a change in the refs index and break the cache then. What do you think of this approach?

I have been tinkering with using the existing GitCommandCache and make commands like GetRefs() disabled after sometime or when the index is changed. But this was quite complex with the current setup. In addition, the commands must still be changed, to get the unfiltered information from Git and filter in GE to be able to cache efficiently. (The command time to get filtered and unfiltered information is almost the same and the GE processing time is minimal to the Git command time.)

The approach here is to just share the information from the change event, so the cache do not need to be expired. The data presented should be consistent also if refs are changed when reading a repo. So this should probably simplify debugging.

Another command similar to GetRefs() is `for-each-ref --format="%(push:track,nobracket)::%(upstream:track,nobracket)::%(push)::%(upstream)::%(refname:short)"` (at home for the Git repo, this command requires a second both for filtered/unfiltered).

git-status is a special one, that could be handled by GitStatusMonitor that monitor when the command is expired in addition to the "repo changed" phase. Various variants of git-diff and git-submodule-status can be handled here too.
But the logic for such a cache is not simple, the logic has been tuned many times (and #9700 is pending...) so I am not sure I want that.

gerhardol:
Some ideas how a general implementation could be done.

Add separate GitCommandCache in addition to the cache for issues that never change: Cache invalidated by the git index (like git refs and HEAD) and by file system watcher (git-status, diff including Git working tree). Maybe cace for index commits too.
Every time the watchers fires, the cache is expired. Some extra safety could be implemented with checking index when a command is to be executed (not for working tree).
The commands in these caches must be changed to filter in GE to use the cache efficiently. Some options like --no-optional-locks should be ignored too.

To avoid that GE performs parsing of the output for these commands, the cache could be customized for the relative few commands using the functionality.

But this is a much bigger job, the gains will be much smaller compared to this PR.
(Basically clicking around with artificial commits will be faster.)

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
```diff
@@ -10,12 +11,17 @@ public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands)
...
 
         public IGitUICommands GitUICommands { get; }
 
         public IWin32Window? OwnerForm { get; }
 
         public IGitModule GitModule => GitUICommands.GitModule;
+
+        private readonly LazyGitRefsProvider _refs;
+
+        public IReadOnlyList<IGitRef> GetRefs(RefsFilter filter) => _refs.GetRefs(filter);
```
mstv, line 25: Mixing `private` and `public`?

Plugins/GitUIPluginInterfaces/ILazyGitRefsProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+
+namespace GitUIPluginInterfaces
+{
+    /// <summary>
+    /// A lazy provider for GitRefs() that can be shared for instance when a repository is changed.
+    /// </summary>
+    public interface ILazyGitRefsProvider
```
mstv, line 8: ❔
```suggestion
    public interface IFilteredGitRefsProvider
```
This interface stands for itself. It is kind of lazy implicitly. The degree of laziness depends on the implementation.

Plugins/GitUIPluginInterfaces/LazyGitRefsProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using GitExtUtils;
+
+namespace GitUIPluginInterfaces
+{
+    /// <inherit/>
+    public sealed class LazyGitRefsProvider : ILazyGitRefsProvider
```
mstv, line 9: follow up:
```suggestion
    public sealed class LazyFilteredGitRefsProvider : IFilteredGitRefsProvider
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -553,16 +553,20 @@ private bool NeedsGitStatusMonitor()
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
-            this.InvokeAsync(RefreshRevisions).FileAndForget();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await this.SwitchToMainThreadAsync();
+                RefreshRevisions(e.GetRefs);
+            }).FileAndForget();
```
mstv, line 560: `RefreshRevisions` is exectuted by the UI thread now - just delayed. It does several updates which can and should be done in the background.
That's why I suggest to extract an `async` variant of `RefreshRevisions` which switches to the UI thread right before `RevisionGrid.PerformRefreshRevisions(getRefs);`.
A synchronous `RefreshRevisions` which triggers the `async` variant, will be needed, too.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -286,12 +284,13 @@ private void OnStatusUpdated(SubmoduleStatusEventArgs e)
 
                     if (_currentNodes is null)
                     {
-                        await ReloadNodesAsync(token =>
+                        var module = Module;
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
```diff
@@ -286,12 +284,13 @@ private void OnStatusUpdated(SubmoduleStatusEventArgs e)
...
+                        await ReloadNodesAsync((token, _) =>
                         {
                             cts = CancellationTokenSource.CreateLinkedTokenSource(e.Token, token);
                             loadNodesTask = LoadNodesAsync(e.Info, cts.Token);
                             return loadNodesTask;
-                        }).ConfigureAwait(false);
+                        }, Module.GetRefs).ConfigureAwait(false);
```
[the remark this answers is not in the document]
gerhardol: ok

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
gerhardol: What do you want to change?

Plugins/GitUIPluginInterfaces/ILazyGitRefsProvider.cs
gerhardol: ok

Plugins/GitUIPluginInterfaces/LazyGitRefsProvider.cs
gerhardol: ok, as for tests

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Submodules.cs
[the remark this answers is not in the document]
gerhardol: (GitRefs is not even used for submodules)

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: (later)

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
mstv: `private` and `protected` properties should be in a separate section, i.e. not intermixed with `public`.

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
```diff
@@ -10,12 +11,17 @@ public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands)
...
+            _refs = new(GitModule);
         }
 
         public IGitUICommands GitUICommands { get; }
 
         public IWin32Window? OwnerForm { get; }
 
         public IGitModule GitModule => GitUICommands.GitModule;
+
+        private readonly FilteredGitRefsProvider _refs;
```
mstv, line 23: The interface could be used here - though pretty academic because the ctor must create an instance of the implementation class.
```suggestion
        private readonly IFilteredGitRefsProvider _refs;
```

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: The handling is the same, InvokeAsync also switches to main thread. (An argument is needed now so an explicit call is needed).
I believe the event is also executed on the main thread, so no explicit switching.
The only in Refresh not requiring to run on the main thread is the disposing check.
So I am not changing this now.

However, I saw that title in dashboard was not set (actually workin OK anyway, this was a change in another PR that should have set this too).
Also eliminated a special case for refreshing after a plugin has executed (no actual change, no eliminated call).

GetRefs() could be explicitly calculated off the main thread, that would not make any change as the use on the main thread would be waiting anyway.

mstv (approved):

gerhardol:
Rebased and squashed.
@RussKie Any objections merging this?
This decreases the time to switch repos significantly

gerhardol:
Squash merge in three days or when @RussKie approves

PR #9744, opened 2021-11-20
Make the Debug build debuggable

RussKie:
An addendum to #9704.

## Proposed changes

- The .NET bootstrap created a .NET Framework entry point. This tricks the debugger, and it doesn't see the app as a .NET process, and, thus, unable to hit any breakpoints. 
To mitigate this, only apply the bootstrap to the Release configuration.
- Correct path to our executable under DotnetRuntimeBootstrapper.
Resolves #9743

PR #9747, opened 2021-11-21
Fix tests

RussKie:
An addendum to #9744

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1825,7 +1825,18 @@ namespace GitCommands
 
         public static string GetGitExtensionsFullPath()
         {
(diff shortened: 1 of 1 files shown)


GitCommands/Settings/AppSettings.cs
```diff
@@ -1825,7 +1825,18 @@ public static DiffListSortType DiffListSorting
 
         public static string GetGitExtensionsFullPath()
         {
-            Debug.Assert(_applicationExecutablePath.EndsWith("GitExtensions.exe"), $"{_applicationExecutablePath} must point to GitExtensions.exe");
+#if DEBUG
+#pragma warning disable SA1515 // Single-line comment should be preceded by blank line
```
gerhardol, line 1829: Follow the rules even if they are annoying here...
Otherwise the workaround is fine for me

PR #9748, opened 2021-11-21
Ensure toggle buttons are providing visual feedback when activated

### LINKED ISSUE #9715: Toolbar togglebuttons highlight frame only
### Environment

- Git Extensions 33.33.33
- Build 34108e4ef0482849d5f6bc2085008ced85c9fcb7 based on 2bde8f93e9602e57b22872d814fbd28c3dba593b
- Git 2.33.1.windows.1
- .NET 5.0.5
- DPI 96dpi (no scaling)

### Issue description

**Actual**

grafik

**Expected**

grafik

### Steps to reproduce

👀

yes, I guess at least 89b98694dd425ab9fc13281ae2fb613df3eb9c2f (side effect of #9608?)

### Diagnostics

_No response_

Greybird:
Fixes #9715

## Proposed changes

- Use a specific menu item class for toggle button, using the Pressed style when Checked

### Before

before

### After

after

## Test methodology 

- Manual tests

## Test environment(s) 

- Windows 10
- DPI 96dpi (no scaling)
- French language

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FilterToolBar.Designer.cs
+++ b/GitUI/UserControls/FilterToolBar.Designer.cs
@@ -32,8 +32,8 @@ namespace GitUI.UserControls
             this.tslblRevisionFilter = new System.Windows.Forms.ToolStripLabel();
             this.tstxtRevisionFilter = new System.Windows.Forms.ToolStripTextBox();
             this.tsddbtnRevisionFilter = new System.Windows.Forms.ToolStripDropDownButton();
-            this.tsmiShowReflogs = new System.Windows.Forms.ToolStripButton();
-            this.tsmiShowFirstParent = new System.Windows.Forms.ToolStripButton();
+            this.tsmiShowReflogs = new GitUI.CommandsDialogs.ToggleToolStripPushButton();
+            this.tsmiShowFirstParent = new GitUI.CommandsDialogs.ToggleToolStripPushButton();
             this.SuspendLayout();
             // 
             // tsmiCommitFilter
@@ -264,8 +264,8 @@ namespace GitUI.UserControls
         private ToolStripMenuItem tsmiAuthorFilter;
         private ToolStripMenuItem tsmiDiffContainsFilter;
         private ToolStripMenuItem tsmiHash;
-        private ToolStripButton tsmiShowReflogs;
-        private ToolStripButton tsmiShowFirstParent;
+        private GitUI.CommandsDialogs.ToggleToolStripPushButton tsmiShowReflogs;
+        private GitUI.CommandsDialogs.ToggleToolStripPushButton tsmiShowFirstParent;
         private ToolStripTextBox tstxtRevisionFilter;
         private ToolStripLabel tslblRevisionFilter;
         private ToolStripSeparator toolStripSeparator19;
--- a/GitUI/UserControls/FilterToolBar.Designer.cs
+++ b/GitUI/UserControls/FilterToolBar.Designer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Windows.Forms;
+using GitUI.CommandsDialogs;
 
 namespace GitUI.UserControls
 {
@@ -21,7 +22,7 @@ namespace GitUI.UserControls
             this.tsmiHash = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiDiffContainsFilter = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStripLabel1 = new System.Windows.Forms.ToolStripLabel();
-            this.tsbtnAdvancedFilter = new System.Windows.Forms.ToolStripButton();
+            this.tsbtnAdvancedFilter = new GitUI.CommandsDialogs.ToggleToolStripPushButton();
             this.tssbtnShowBranches = new System.Windows.Forms.ToolStripSplitButton();
             this.tsmiShowBranchesAll = new System.Windows.Forms.ToolStripMenuItem();
             this.tsmiShowBranchesCurrent = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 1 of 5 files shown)


Greybird:
Feel free to revert should you see any possible enhancement.
I'm pretty sure there are things to do around `CheckOnClick` property on menu items to simplify code that ensure underlying setting and `Checked` property of the menu item are in sync, but this particular improvement could be done afterwards I think.

Greybird:
@RussKie : unsure about the failures on AppVeyor, tests are successful on my dev pc, would you have guidance about how to fix?

pmiossec:
> @RussKie : unsure about the failures on AppVeyor, tests are successful on my dev pc, would you have guidance about how to fix?

Flaky tests. Passed after rebuild of PR triggered...

GitUI/UserControls/FilterToolBar.Designer.cs
```diff
@@ -264,8 +264,8 @@ private void InitializeComponent()
         private ToolStripMenuItem tsmiAuthorFilter;
         private ToolStripMenuItem tsmiDiffContainsFilter;
         private ToolStripMenuItem tsmiHash;
-        private ToolStripButton tsmiShowReflogs;
-        private ToolStripButton tsmiShowFirstParent;
+        private GitUI.CommandsDialogs.ToggleToolStripPushButton tsmiShowReflogs;
+        private GitUI.CommandsDialogs.ToggleToolStripPushButton tsmiShowFirstParent;
         private ToolStripTextBox tstxtRevisionFilter;
         private ToolStripLabel tslblRevisionFilter;
         private ToolStripSeparator toolStripSeparator19;
```
gerhardol, line 271: `tsbtnAdvancedFilter` requires this too from #9728 (that adds push in #9728)

gerhardol (commented):
For "permanent" pushed buttons like left panel and layout related, I probably prefer the subtle indication in master.
For temp states like Reflog (and Advanced filter when added) and FileViewer this is better.

GitUI/UserControls/FilterToolBar.Designer.cs
Greybird: @gerhardol : good catch, I focused on usages of `Checked` property, but this button is somewhat different. However, it is actually a toggle.
I'll have to work a little bit on determining if it has to be pressed or not, as I do not see a property return a consistent value about the presence of advanced filters if the branch filter is applied or not (the branch filter seems to be integrated in the filters, making the `HasFilter` property `true`, once advanced filter has been activated once with this setting enabled.

Greybird:
> For "permanent" pushed buttons like left panel and layout related, I probably prefer the subtle indication in master. For temp states like Reflog (and Advanced filter when added) and FileViewer this is better.

I'm totally open to modification ; can we just make sure a consensus is present on the matter to prevent back and forth on the subject ?
(I'm not familiar with how you are taking decisions for these matters, if your voice is enough, could you please confirm this ?)

GitUI/UserControls/FilterToolBar.Designer.cs
gerhardol: As this is added in #9728 (plan to merge today), it is not visible yet. Need to be added before merge.

gerhardol:
> > For "permanent" pushed buttons like left panel and layout related, I probably prefer the subtle indication in master. For temp states like Reflog (and Advanced filter when added) and FileViewer this is better.
> 
> I'm totally open to modification ; can we just make sure a consensus is present on the matter to prevent back and forth on the subject ? (I'm not familiar with how you are taking decisions for these matters, if your voice is enough, could you please confirm this ?)

@mstv please comment
There is no defined flow for comments, it is the strongest opinions that get priority. Regardless, there should be a few days for comments.
Of course, what RussKie says is most important. 
In this case, I raised my opinion in the review.

I kind of asked for an inconsistent experience, so I am OK with being turned down too.

The real problem for me is that the selected color is much more accented with this style then it is in 3.5.
This is OK when actually pushing buttons but to visible for my preference.
If this is how the visual style is intended, it should maybe be like this, but I do not like it....

Static screenshots to show this better

3.5
image

master
image

this PR
image

mstv:
> The real problem for me is that the selected color is much more accented with this style then it is in 3.5.
This is OK when actually pushing buttons but to visible for my preference.
If this is how the visual style is intended, it should maybe be like this, but I do not like it....

Fully agree

Regarding #9715, I don't have a strong opinion. That's why I labeled it as "discussion". (The frame is a little thin to me.)

gerhardol:
Can we use the "hover over non selected" instead?
The Tag button is just hovered, the other are selected.

image

Greybird:
> Can we use the "hover over non selected" instead? The Tag button is just hovered, the other are selected.

It is technically possible by using a similar hack:
```
public class ToggleToolStripPushButton : ToolStripButton
{
    private bool _toggled;
    public override bool Selected => base.Selected || Toggled;

    public bool Toggled
    {
        get => _toggled;
        set
        {
            if (value != _toggled)
            {
                _toggled = value;
                if (Parent != null)
                {
                    Parent.Invalidate(Bounds);
                    Parent.Update();
                }
            }
        }
    }
}
```
But is would also require to switch all `Checked` usages to `Toggled`, ans also in code running through items to set the standard Checked property on menus or buttons, to ensure Checked stays false, as, the "hover over non selected" color is in fact "selected non checked" (see [Selected](https://github.com/dotnet/winforms/blob/4e4f13a1100bfcde2760ac0dbfcad6b040363efd/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs#L1749-L1765) doc)

At this point, I suppose the hack is not worth anymore, as it is also error-prone in the future, and it might be better to consider designing a specific `Renderer` to replace the standard `ToolStripProfessionalRenderer`.

RussKie (approved):
LGTM, though I personally would prefer a slightly lighter colour.

gerhardol:
I lean on voting to keep the current looks in master with the thin frame over the dark color. The proposal is too bright for me and the hack seem to intrusive (not so hard to implement, but hard to maintain.)
My private favorite is the "inconsistent" selecting and I believe that this can be explained to the users, but it requires explanation, so probably not a good solution.

gerhardol:
@gitextensions/git-extensions-team Any opinions about this, keep as is in current (not much visibility) or this PR (too much visibility)?

NikolayXHD:
Hmm, strange. With default style there should be no issue:

image

Checked buttons are highlighted properly.

Sorry I cannot confirm it is also true for current master where Theming is not supported, but at least it should be like that.

RussKie:
I prefer the lighter version.

NikolayXHD:
The problem should be solved by #9795 which brings back properly highlighting selected toolstrip buttons as in release-3.5 branch

gerhardol:
@Greybird Thanks for the effort, but the core issue was fixed in #9795
Anything outstanding in this PR or should it be closed?

Greybird:
> @Greybird Thanks for the effort, but the core issue was fixed in #9795
> Anything outstanding in this PR or should it be closed?

Great ! To me, nothing to salvage from the pr. You can close it !

PR #9757, opened 2021-11-26
RevDiff: Move RangeDiff to A->B group

gerhardol:
Suggested in #9720, required #9718 too

## Proposed changes

RangeDiff shows the differences between two branches, identifying and comparing commits.
This was previously in a separate group.

This PR discards the second group and moves it into the First(A)->Second(B) goup.
The RangeDiff is always sorted last.

Note: The RangeDiff always happened to be sorted last also without the sort changes, that could be replaced with a comment.
But this could accidentally break. (A test could be added eventually.)

### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusDiffCalculator.cs
+++ b/GitUI/UserControls/FileStatusDiffCalculator.cs
@@ -200,8 +200,7 @@ namespace GitUI
                 return fileStatusDescs;
             }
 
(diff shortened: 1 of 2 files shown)


mstv (approved):
👍

valdo1990:
Thank you

RussKie (commented):
+0

PR #9759, opened 2021-12-01
RevisionGrid: Option for git log with --topo-order

### LINKED ISSUE #9758: Graph: Option to group commits (below a merge) from incoming branch above the commits from receiving branch
### Feature description

If you run `git log --all --graph --oneline` you get a graph like this:

image

As you can see below the merge commit `9f9d5a1`, the three commits from the incoming branch are grouped above the two commits from the receiving branch.

I prefer it this way. But in GitExtensions, the commits are sorted by committed date or author date, with no regards to the branch they're on:

image

I find it harder to read and follow the history when it's like this, especially in graphs that are more complex than the simple example here. All the commits are jumbled up based on their author/committer date sorting and not sensibly grouped by the branch they were in.

It will be nice to be able to sort by committer date or author date, but there should be another option: **"Group commits by branch in graph"**.

- When this is enabled, the grouping would come first applying globally to the entire graph. The sorting (by committer/author) applies locally to each group (not globally to the entire graph).
- When this is disabled, the sorting (by committer/author) would apply globally to the entire graph.

_Sorry if this feature already exists, but I couldn't find it._

### Environment

N/A

mstv:
Fixes #9758

## Proposed changes

- Explicitly specify the argument `--topo-order` to `git log` if not sorting by author date

### Before

commit date order

### After

topo order

## Test methodology 

- 👀

## Test environment(s) 

- Git Extensions 33.33.33
- Build 624068b81beb2475fa94987dc1e0c50ab60a1618
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -131,7 +131,8 @@ namespace GitCommands
                     new ArgumentBuilder
                     {
                         { refFilterOptions.HasFlag(RefFilterOptions.Reflogs), "--reflog" },
-                        { AppSettings.SortByAuthorDate, "--author-date-order" },
+                        { AppSettings.RevisionSortOrder == RevisionSortOrder.AuthorDate, "--author-date-order" },
+                        { AppSettings.RevisionSortOrder == RevisionSortOrder.Topology, "--topo-order" },
                         {
                             refFilterOptions.HasFlag(RefFilterOptions.All),
                             "--all",
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -19,6 +19,14 @@ using Microsoft.Win32;
 
 namespace GitCommands
 {
+    public enum CommitInfoPosition
+    {
+        // DO NOT RENAME THESE -- doing so will break user preferences
+        BelowList = 0,
+        LeftwardFromList = 1,
+        RightwardFromList = 2
+    }
+
     public enum LocalChangesAction
     {
         // DO NOT RENAME THESE -- doing so will break user preferences
@@ -28,13 +36,12 @@ namespace GitCommands
         Stash
     }
 
-    public enum TruncatePathMethod
+    public enum RevisionSortOrder
     {
         // DO NOT RENAME THESE -- doing so will break user preferences
-        None,
-        Compact,
-        TrimStart,
-        FileNameOnly
+        GitDefault,
+        AuthorDate,
+        Topology
     }
 
     public enum ShorteningRecentRepoPathStrategy
@@ -45,11 +52,13 @@ namespace GitCommands
         MiddleDots
     }
 
-    public enum CommitInfoPosition
+    public enum TruncatePathMethod
     {
-        BelowList = 0,
-        LeftwardFromList = 1,
-        RightwardFromList = 2
+        // DO NOT RENAME THESE -- doing so will break user preferences
+        None,
+        Compact,
+        TrimStart,
+        FileNameOnly
     }
 
     public static class AppSettings
@@ -524,10 +533,10 @@ namespace GitCommands
             set => SetBool("showgitstatusforartificialcommits", value);
         }
 
-        public static bool SortByAuthorDate
+        public static RevisionSortOrder RevisionSortOrder
         {
-            get => GetBool("sortbyauthordate", false);
-            set => SetBool("sortbyauthordate", value);
+            get => GetEnum("RevisionSortOrder", RevisionSortOrder.GitDefault);
+            set => SetEnum("RevisionSortOrder", value);
         }
 
         public static bool CommitInfoShowContainedInBranches => CommitInfoShowContainedInBranchesLocal ||
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -19,48 +19,6 @@ using Microsoft.Win32;
 
 namespace GitCommands
 {
-    public enum CommitInfoPosition
-    {
-        // DO NOT RENAME THESE -- doing so will break user preferences
-        BelowList = 0,
-        LeftwardFromList = 1,
-        RightwardFromList = 2
-    }
-
-    public enum LocalChangesAction
-    {
-        // DO NOT RENAME THESE -- doing so will break user preferences
-        DontChange,
-        Merge,
-        Reset,
-        Stash
-    }
-
-    public enum RevisionSortOrder
-    {
-        // DO NOT RENAME THESE -- doing so will break user preferences
-        GitDefault,
-        AuthorDate,
-        Topology
-    }
-
-    public enum ShorteningRecentRepoPathStrategy
-    {
-        // DO NOT RENAME THESE -- doing so will break user preferences
-        None,
-        MostSignDir,
-        MiddleDots
-    }
-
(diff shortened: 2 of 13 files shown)


mstv:
The time for loading a large repo using `--topo-order` ([PR build](https://ci.appveyor.com/api/buildjobs/om9pv5axs468pch3/artifacts/artifacts%2FRelease%2Fpublish%2FGitExtensions-Portable-4.0.0.13176-b4f18f1a2.zip)) should be compared with the [previous dev build](https://ci.appveyor.com/api/buildjobs/hm4qlxyjl9p77bre/artifacts/artifacts%2FRelease%2Fpublish%2FGitExtensions-Portable-4.0.0.13172-7b2db86c3.zip). @ADTC, could you measure the times?
Depending on the result, it might become the default value.

The `Sort by` setting might be turned into a ComboBox: `standard (mainly by commit date)`, `author-date`, `topology`.

image

GitCommands/RevisionReader.cs
```diff
@@ -181,7 +181,7 @@ public sealed class RevisionReader : IDisposable
                     new ArgumentBuilder
                     {
                         { refFilterOptions.HasFlag(RefFilterOptions.Reflogs), "--reflog" },
-                        { AppSettings.SortByAuthorDate, "--author-date-order" },
+                        { AppSettings.SortByAuthorDate, "--author-date-order", "--topo-order" },
```
gerhardol, line 184: `--topo-order` is a filter too, this cannot be done this way.
Branches not merged to HEAD are not shown.

So this need to be displayed with filters too.

gerhardol (changes requested):

GitCommands/RevisionReader.cs
mstv: > Branches not merged to HEAD are not shown.

Sure? I would have expected an explicit warning in the git documentation. At least, I have not noticed it.

image

gerhardol: Inserting in branches, could it be different with other options?
(I do not see your commit with this option)
Git 2.34.1

`-c log.showSignature=false log --max-count=100000 -z --topo-order --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B"`

I interpret the following as a filter:
```
Show no parents before all of its children are shown, and avoid showing commits on multiple lines of history intermixed.
```

mstv: If added using your tweak, `--all` is omitted. With this PR it is:
`-c log.showSignature=false log --max-count=100000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --topo-order --all --boundary --not --glob=notes --not`
So in combination with filters, we must double-check if `--topo-order` causes side-effects.

gerhardol: `--all` is added so it seem OK.

Then it likely comes down to implementing the dropdown (I prefer the current way to display).

ADTC:
I don't know if this is the right approach, since I don't explicitly mention _any_ order option in my `git log` command. Anyway I'll try it in the advanced filter and let you know.

Thanks for this, though it is not exactly what I though it will be. 🙂

mstv:
https://git-scm.com/docs/git-log#Documentation/git-log.txt---topo-order

Before this PR, we did not specify an order (except `--author-date-order` depending on the setting)
`-c log.showSignature=false log --max-count=100000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary --not --glob=notes --not`

gerhardol:
`--topo-order` (implicit for command line `--graph` see the man page) is a sorting mechanism, can be combined with date/author sorting though.

Note that any sorting will delay when Git returning results for GE to process, so it could take seconds more before the graph is updated (no different from the commandline).
The sorting also increases Git execution time. For Linux the command time increases from 3.0 s to 13.6 s when adding both topo and date, so it could be a significant slowdown.

gerhardol:
image

Suggestion:
`Order commits by topo`
(Not proper English but Git lingo.)

Remove `Sort by author` from Settings?
(The benefit of having it there is to show the warning for that sort delays rendering, I guess that could be kept with a comment refering to the View mwnu).

RussKie:
I agree with the idea of updating the settings from a checkbox to a combo,
and disagree on presence of the option in the View menu - it looks very out
of place there.
If this affects filtering then we need to add it to the filter dialog, an
optionally to the toolbar options.

gerhardol:
> If this affects filtering then we need to add it to the filter dialog, an optionally to the toolbar options.

This could affect filtering but I suggest it is added with `--all` and is lablled as `Order commits by topo`

eternalstarfire:
💥  Just wanted to say YAY - I can't wait for this!! 😍

Also: SourceTree calls this '_Ancestor order_' (the other option being 'Date order')

gerhardol:
The change in cc8071e69d964f72545b87a6c0e2d9d3e6b2933f seems OK

mstv:
(This has healed me from adding any further option to settings pages.)

GitCommands/Settings/AppSettings.cs
```diff
@@ -19,6 +19,14 @@
 
 namespace GitCommands
 {
+    public enum CommitInfoPosition
```
RussKie, line 22: Could you please move this and others into own files, e.g., AppSettings.CommitInfoPosition.cs

RussKie (approved):

GitCommands/Settings/AppSettings.cs
mstv: Done.

GitUI/Translation/English.xlf
```diff
@@ -9732,6 +9732,10 @@ See the changes in the commit form.</source>
         <source>Highlight selected branch (until refresh)</source>
         <target />
       </trans-unit>
+      <trans-unit id="TopoOrder.Text">
+        <source>Arrange commits by t&amp;opo order (ancestor order)</source>
```
RussKie, line 9736:
```suggestion
        <source>Arrange c&amp;ommits by topo order (ancestor order)</source>
```

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
```diff
@@ -432,6 +425,20 @@ VIEW menu
...
+                {
+                    Name = "AuthorDateSort",
+                    Text = "&Sort commits by author date",
+                    ExecuteAction = () => _revisionGrid.ToggleAuthorDateSort(),
+                    IsCheckedFunc = () => AppSettings.RevisionSortOrder == RevisionSortOrder.AuthorDate
+                },
+                new MenuCommand
+                {
+                    Name = "TopoOrder",
+                    Text = "Arrange commits by t&opo order (ancestor order)",
```
RussKie, line 438:
```suggestion
                    Text = "Arrange c&ommits by topo order (ancestor order)",
```

RussKie (commented):
👍

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
mstv: Done.

gerhardol (approved):

PR #9760, opened 2021-12-01
Settings could not be read from \\wsl$

gerhardol:
A few follow up from #9702 

## Proposed changes

A few a follow up related to WSL Git.

* Settings could not be read from \\wsl$

The URI is invalid, why the path has to be read with a StreamReader.
This caused various issues like remotes could not be disabled

This is not related to WSL Git, it also occurs if Windows Git is used.
This is one reason why insiders Windows builds also supports `wsl.localhost`
(\\wsl$ paths will be required to utilize wsl git. This is a feature but must be documented.)

This problem was primarily seen when deactivating remotes, an exception occurred (as a settings file were not refreshed).

* WSL: Remotes for filepaths could not be saved

Paths must be manipulated to match Git executable.

* WSL: Git internal paths were incorrectly resolved

Affected Remote mostly

* WSL: force-push was inserted incorrectly

The `--force-with-lease` offered if a push is refused assumed that the command started with "push".
The option is "inserted" in the command string.
WSL Git has the distribution first, why the insert point differs.

Note: The output from `git push` is not always available for WSL Git (for instance at home).
At work it works though.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -329,6 +329,7 @@ namespace GitCommands
         /// See https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---git-pathltpathgt.
         /// </summary>
         /// <param name="relativePath">A path relative to the .git directory.</param>
+        /// <returns>An absolute path in Windows format.</returns>
         public string ResolveGitInternalPath(string relativePath)
         {
             GitArgumentBuilder args = new("rev-parse")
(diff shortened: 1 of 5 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -328,24 +328,23 @@ public static bool IsValidGitWorkingDir(string? dir)
         /// git special folders are located in different directories depending on the kind of repo: submodule, worktree, main.
         /// See https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---git-pathltpathgt.
         /// </summary>
-        /// <param name="relativePath">A path relative to the .git directory.</param>
+        /// <param name="relativePath">An absolute path relative to the .git directory.</param>
```
mstv, line 331:
```suggestion
        /// <param name="relativePath">A path relative to the .git directory.</param>
```

mstv (approved):
Looks sensible, have not run

GitCommands/Git/GitModule.cs

GitCommands/Git/GitModule.cs
```diff
@@ -329,23 +329,22 @@ public static bool IsValidGitWorkingDir(string? dir)
         /// See https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---git-pathltpathgt.
         /// </summary>
         /// <param name="relativePath">A path relative to the .git directory.</param>
+        /// <returns>An absolute path in Windows format</returns>
```
RussKie, line 332:
```suggestion
        /// <returns>An absolute path in Windows format.</returns>
```

PR #9765, opened 2021-12-05
Add mnemonics to filter toolbar

mstv:
## Proposed changes

- add mnemonics to labels of `ToolStripFilters`

### Before

image

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7c33b6fab9bee4328eb970efa6d50653f5297e74
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FilterToolBar.Designer.cs
+++ b/GitUI/UserControls/FilterToolBar.Designer.cs
@@ -72,7 +72,7 @@ namespace GitUI.UserControls
             this.toolStripLabel1.Name = "toolStripLabel1";
             this.toolStripLabel1.Size = new System.Drawing.Size(58, 22);
             this.toolStripLabel1.Tag = "ToolBar_group:Branch filter";
-            this.toolStripLabel1.Text = "Branches:";
+            this.toolStripLabel1.Text = "&Branches:";
             this.toolStripLabel1.ToolTipText = "Branch filter";
             // 
             // tsbtnAdvancedFilter
@@ -129,7 +129,7 @@ namespace GitUI.UserControls
             this.tsmiShowReflogs.Image = global::GitUI.Properties.Images.Book;
             this.tsmiShowReflogs.Name = "tsmiShowReflogs";
             this.tsmiShowReflogs.Size = new System.Drawing.Size(23, 22);
-            this.tsmiShowReflogs.Text = "Reflogs";
+            this.tsmiShowReflogs.Text = "R&eflogs";
             this.tsmiShowReflogs.ToolTipText = "Show reflogs";
             this.tsmiShowReflogs.Click += new System.EventHandler(this.tsmiShowReflogs_Click);
             // 
@@ -190,7 +190,7 @@ namespace GitUI.UserControls
             this.tslblRevisionFilter.Name = "tslblRevisionFilter";
             this.tslblRevisionFilter.Size = new System.Drawing.Size(36, 22);
             this.tslblRevisionFilter.Tag = "ToolBar_group:Text filter";
-            this.tslblRevisionFilter.Text = "Filter:";
+            this.tslblRevisionFilter.Text = "&Filter:";
             this.tslblRevisionFilter.ToolTipText = "Text filter";
             // 
             // tstxtRevisionFilter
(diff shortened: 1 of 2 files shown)


mstv:
@mdonatas, do you object to remove "your" hotkey `Ctrl+E`?

gerhardol (commented):
+1
OK to remove for me, but need to agree with the potential users

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2019,7 +2019,6 @@ internal enum Command
             FocusCommitInfo = 4,
             FocusDiff = 5,
             FocusFileTree = 6,
-            FocusFilter = 18,
```
RussKie, line 2022: Please mark as deprecated (an example below)

mdonatas:
Ummm... `remove non-intuitive...` what's a more intuitive alternative then?
In `Windows Explorer` `ctrl-e` focuses the search bar
In `Firefox` `ctrl-e` focuses the search bar
In `Chrome` `ctrl-e` focuses the search bar / address bar
In `Outlook` `ctrl-e` focuses the search bar
In `VSCode` `ctrl-e` focuses the search bar

Some of the web apps are using `/` to focus a search/filter box but I haven't noticed this used in any desktop apps.

Can it simply be unset rather than removed? When removed this option is taken away from all the users.

mdonatas:
By the way, why is it not unique? It's not set at all by default.
https://github.com/gitextensions/gitextensions/blob/81828631a90545d01c525bfe98494ac3d9b25787/GitUI/Hotkey/HotkeySettingsManager.cs#L336

valdo1990:
No

mstv:
> Ummm... `remove non-intuitive...` what's a more intuitive alternative then?

`Ctrl+E` has no relation to "Filter" / "Search" IMO.
Mnemonics are much more discoverable.
@gitextensions/git-extensions-team, should I add just the mnemonics (in addition to the hotkey, which will be kept)?

> In `Windows Explorer` `ctrl-e` focuses the search bar
In `Firefox` `ctrl-e` focuses the search bar
In `Chrome` `ctrl-e` focuses the search bar / address bar
In `Outlook` `ctrl-e` focuses the search bar
In `VSCode` `ctrl-e` focuses the search bar
Can it simply be unset rather than removed? When removed this option is taken away from all the users.

I didn't know that it is commonly used. I use `F6` / [`Shift+`]`Ctrl+F`.
Of course, we will keep it then. I can unassign it for my installations if really necessary.

> By the way, why is it not unique? `ShowEntireFile` is not set at all by default.

You are right. I did not remember that I had assigned it manually.

 (I am going to add / use hotkeys for the DiffView settings and to make them available from the RevisionGrid if the DiffView is visible and if the hotkey is not handled by the RevisionGrid itself. This PR was a preparation step.)

RussKie:
I side with @mstv, I don't either see the search in Git Extensions being
equivalent to search in Windows Explorer or a browser. That's my _personal_
opinion (as a user, not as a maintainer).

mdonatas:
Hmm, they are not equivalent in that sense but they are equivalent in "get me to the field that filters/searches". In all these cases (except VSCode) there is a field in UI which is not hidden and which is there to help user find whatever makes sense based on the app.

mstv:
We can have both: mnemonics and hotkey.

GitUI/UserControls/FilterToolBar.Designer.cs
```diff
@@ -129,7 +129,7 @@ private void InitializeComponent()
             this.tsmiShowReflogs.Image = global::GitUI.Properties.Images.Book;
             this.tsmiShowReflogs.Name = "tsmiShowReflogs";
             this.tsmiShowReflogs.Size = new System.Drawing.Size(23, 22);
-            this.tsmiShowReflogs.Text = "Reflogs";
+            this.tsmiShowReflogs.Text = "R&eflogs";
```
gerhardol, line 132: I would like to remove the test label on the button, there is a tooltip

gerhardol (approved):
See comment about removing text on button though, could be combined...

GitUI/UserControls/FilterToolBar.Designer.cs
RussKie: Apologies, I can't parse what you're saying.

gerhardol: Reflogs should only have an image, no text
Only Commit button has that (filter fields could be modified too)

mstv: (I took me a while, too: s/test/text/.)
I have removed the text label and updated the screenshot.

PR #9767, opened 2021-12-06
Update .NET Runtime Bootstrapper

### LINKED ISSUE #9766: Update `DotnetRuntimeBootstrapper` to latest version to make use of added improvements
### Feature description

Hi!

I noticed you were using https://github.com/Tyrrrz/DotnetRuntimeBootstrapper.
In the last major update it was reworked quite drastically and as a result comes with some large improvements. You can read the full list in the changelog or in updated readme, but here are the highlights:

1. Using `hostfxr.dll` now instead of `dotnet` CLI. This means no extra process is created.
2. Optimized hot path. Prerequisites will only be checked for if the app fails to run. Made possible by the previous point.
3. Bootstrapper executable is now only generated on `dotnet publish`. It used to also be generated on `dotnet build` too, but that broke debugging. You can still re-enable that with `<GenerateBootstrapperOnBuild>true</GenerateBootstrapperOnBuild>` if you want for some reason.
4. Strategy used for resolving .NET Runtime, both the installation location as well as the download URL, have been optimized and made more stable. The former now mimics the behavior of `nethost` library.
5. Native resources (mainly application manifest) is now properly inherited from the `exe` file by the application. Previously it wasn't possible because the bootstrapper created a separate process (for `dotnet` CLI), but it doesn't do this anymore.

### Environment

- Git Extensions 3.5.4.12724
- Git 2.31.1.windows.1
- .NET Framework 4.8.4420.0
- DPI 96dpi (no scaling)

Tyrrrz:
Fixes #9766

## Proposed changes

- Update DotnetRuntimeBootstrapper to latest version
- Revert the workaround for getting application path (it should now resolve correctly)
- Remove condition on package reference (bootstrapper is now only generated on publish, not on build)

## Test methodology 

- I tested it manually :)

## Test environment(s) 

- GIT 2.31.1.windows.1
- Windows 10.0.19043.1348

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -61,11 +61,7 @@ namespace GitCommands
         public static readonly string ApplicationId = ApplicationName.Replace(" ", "");
         public static readonly string SettingsFileName = ApplicationId + ".settings";
         public static readonly string UserPluginsDirectoryName = "UserPlugins";
-
-        // Under DotnetRuntimeBootstrapper this points to C:\Program Files\dotnet\dotnet.exe
-        // whereas we need our executable
-        private static string _applicationExecutablePath = Assembly.GetEntryAssembly().Location.Replace(".dll", ".exe");
-
+        private static string _applicationExecutablePath = Application.ExecutablePath;
         private static string? _documentationBaseUrl;
 
         public static Lazy<string?> ApplicationDataPath { get; private set; }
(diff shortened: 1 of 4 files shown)


gerhardol (commented):
+1
Seem OK to me

RussKie (approved):

RussKie:
Thank you

PR #9784, opened 2021-12-30
Load app colors from CSS themes

### LINKED ISSUE #9782: Restore Color settings with .NET5
### Feature description

Theming was unfortunately disabled with .NET5 in #8522, not in the current roadmap for next release.
This is tracked in #9191
There primarily issues with the hooking required to change the view of user controls, but also to some extent with the .css parsing in the ExCSS library used.

However, the theming also included the possibility to change colors for labels etc available before the theming (also with colorpickers etc).
This is required for instance to switch back the colors for local/remote branches, an unfortunate change that left many users dissatisfied.

This functionality can be restored in two ways:
* Migrate back to the old application colors.
* Fix the issues with the ExCSS package with netstandard and only use the "color" part, maybe even remove the references

### Environment

- GitExtensions version: master
- .NET version: 5.0.404

gerhardol:
Fixes #9782

## Proposed changes

Restore the possibility to change application colors from .css files.
While the theme handling is unfortunately disabled after .NET5, the possibility to set colors like branches was available before theming and is expected by users.

@NikolayXHD suggested that ExCSS is removed and replaced with regex parsing, to get rid of a dependency and get parse errors (that must be ignored with ExCSS >= 3).
This is not done due to simplicity, easier to fix this than to reimplement.  (If the implementation is limited to allow only single line comments, one command per line etc, it would be simpler and maybe feasible.)

### Before

image

### After

Weird mix of switch local-remote colors and colorblind
image

## Test methodology 

Renabled and adapted tests

## Merge strategy

----

### DIFF
--- a/GitExtensions/Program.cs
+++ b/GitExtensions/Program.cs
@@ -11,6 +11,7 @@ using GitUI.CommandsDialogs.SettingsDialog;
 using GitUI.CommandsDialogs.SettingsDialog.Pages;
 using GitUI.Infrastructure.Telemetry;
 using GitUI.NBugReports;
+using GitUI.Theming;
 using GitUIPluginInterfaces;
 using Microsoft.VisualStudio.Threading;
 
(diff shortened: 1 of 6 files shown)


NikolayXHD (approved):

RussKie:
👍

[nit] I would skip/ignore the test instead of the pragma.

gerhardol:
Updated comment about no errors parsed (and that this could be rewritten), removed the dummy rule and unnecessary handling.
Plan to merge tomorrow.

Issue about the error parsing in ExCSS:
https://github.com/TylerBrinks/ExCSS/issues/132

RussKie (approved):

PR #9785, opened 2022-01-02
BuildServer: Correct settings paths

### LINKED ISSUE #9135: .NET 5.0: Restore build server integration
## Current behaviour

- No build server status
- `Build server type` not selectable because the ComboBox is empty

## Expected behaviour

CI status is displayed in the RevisionGrid

## Screenshots

NoBuildServerType

## Environment

based on #8522:
- Git Extensions 0.0.0.0
- Build 343a904ba6fbe2896022511039114ee46d441509
- Git 2.31.1.windows.1
- .NET 5.0.5
- DPI 96dpi (no scaling)

gerhardol:
Fixes #9135

## Proposed changes

* Regression in master from #9082:
Settings were read/written to path missing "BuildServer."
BuildServer worked OK if settings were reapplied.

* Cleanup contents in GE bundled GitExtensions.settings
Remove some old TeamCity settings etc

--

Settings: Changing view from effective must not overwrite

Changing from Effective to any other level (local/dist/global) overwrote
settings on that level using effective values.

This change applies to all settings.

--

#9082 tried to improve the settings, not fully working still.
The behavior below is also in 3.5 as far as I can see.
BuildServer: Unset settings were always set to effective

setting when editing a specific level (Local<-Distributed<-User (global)).

Checkboxes now reflects if the bool value is set on the specific level.

For textboxes will an empty string be handled as unset (i.e. null).
Without this will lower priority values always have to be set in
higher priority settings. Order is Local<-Distributed<-User (global).

For BuildServer settings where an empty string could make sense to
override a value (like ignore a setting in Distributed by setting it locally)
adummy value must be used.
This is still considered a better functionality than requiring the value to be set.

* Displayed settings were not updated when changing level

## Screenshots 

Same settings in both, AppVeyorAccountName and AppVeyorAccountToken set in global config, AppVeyorProjectName set in distributed and overridden in local.

### Before

Similar and master and 3.5 (but master do not read the correct path due to first commit)

build-server-354

### After

build-server-settings

## Test methodology 

Manual

## Merge strategy

----


gerhardol:
@ivangrek Can you please review?

gerhardol:
Restructured the commits (separating the generic change and moving some code snippets to be generic, did not bother for CheckBox handling).

RussKie (commented):
Conceptually I understand the fix, but looking at the diff I find myself scratching head looking at the API in `BuildServerSettings`. E.g. I can infer the purpose of `EnableIntegration` property, but I struggle with `EnableIntegrationWithDefault` and `ByType()`.

gerhardol:
> Conceptually I understand the fix, but looking at the diff I find myself scratching head looking at the API in `BuildServerSettings`. E.g. I can infer the purpose of `EnableIntegration` property, but I struggle with `EnableIntegrationWithDefault` and `ByType()`.

I spent considerable time in investigating the issue myself.

The generic build server related settings are in the namespace "BuildServer.". This sets if integration is enabled at all and the type (e.g. `AppVeyor`) of the build server.

The settings for a specific server is in the namespace like "BuildServer.AppVeyor.", that is what what ByType() is extracting. This is what the build server plugins are using, they are not aware of the upper namespaces. (This is the core issue in #9082).

So the following:
```
buildServerAdapter.Initialize(this, buildServerSettings.ByType(),
```

Could be written:
```
buildServerAdapter.Initialize(this, _module().GetEffectiveSettings().ByPath($"BuildServer.{buildServerSettings.Type}")
```

(but "BuildServer" should have a method).

For `EnableIntegration` vs `EnableIntegrationWithDefault` will some users need to distinguish unset values and some should use a global default.

In this case the default handling could be distributed to the users (i.e. instead of `EnableIntegrationWithDefault` use `EnableIntegration ?? _enableIntegrationDefault`),
which enables for different default handling in different part of the application, which I would like to avoid.

gerhardol:
@RussKie Regarding the updates: I do not think it is an improvement that the plugins resolve the settings path. 
I can live with this though.

However, the changes mixes the settings path and the "effective" changes, the commits have to be squashed (unless you want to untangle).
I would have preferred to have two commits for that. However, it was much easier to test the changes in one PR.

RussKie:
I'm still working through this. Please bear with me.

gerhardol:
@RussKie is rewriting this, the latest commits compiles but do not work

RussKie:
I consider my tweaks complete, please have a look.

> I do not think it is an improvement that the plugins resolve the settings path.
> I can live with this though.

It feels less "leaky" as exposing settings source at `IBuildServerSettings` level, and then invoking it at intermediary layers (such as  `BuildServerIntegrationSettingsPage` when calling `control.LoadSettings` and `control.SaveSettings`).

RussKie:
I'm still yet to formally test it though.

RussKie:
I'm working on integration tests now.

UnitTests/GitCommands.Tests/Settings/MockData/level1_repogit_GitExtensions.settings.xml
```diff
@@ -0,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<dictionary>
+  <item>
+    <key>
+      <string>BuildServer.AppVeyor.AppVeyorProjectName</string>
```
gerhardol, line 5: For these tests to have relevance, there should be tests that the effective data is correctly interpreted, i.e. with several build servers and overriding data.
To some extent, each build server internal setting dependency should be handled too...

gerhardol (commented):
+1
A few comments, I can live with this. Should be merged soon.
It feels like you spent some time rewriting this code to understand it, I cannot say that the API is better... (I do this too from time to time.)

UnitTests/GitCommands.Tests/Settings/MockData/level1_repogit_GitExtensions.settings.xml
RussKie: This test has these traits - there are 2 different build servers and 1 override.

The generic infra has no knowledge of specifics of each individual build server and hence can't test those here. And since there're plans to move the build server integration it'd be incorrect to add each build server specific tests here either.

GitCommands/Settings/BuildServerSettings.cs
```diff
@@ -1,40 +1,46 @@
...
+        /// <inheritdoc />
+        public ISettingsSource GetSettings()
+        {
+            string? type = BuildServerType;
+            Validates.NotNull(type);
+
+            return new SettingsPath(_settingsSource, $"{BuildServerGroupName}.{type}");
+        }
+
+        public string? BuildServerType
```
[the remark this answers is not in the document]
RussKie: Upon closer investigation I arrived to a conclusion that this is not a "type" rather a "name" or a "user readable label". Which is a mistake, and we'll need deal with this.

UnitTests/GitCommands.Tests/Settings/MockData/level1_repogit_GitExtensions.settings.xml
gerhardol: The override of BuildServerType should be checked
As the module provides functionality for settings for BuildServerTypes, it should test that.
Like `BuildServer.Type=MyServer` and `BuildServer.Type.MyServer=MySetting`, test overrides.

GitCommands/Settings/BuildServerSettings.cs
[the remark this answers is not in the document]
gerhardol: `ServerName`?

GitCommands/Settings/BuildServerSettings.cs
```diff
@@ -1,40 +1,46 @@
...
         {
             _settingsSource = settingsSource;
         }
 
-        public string? Type
+        /// <inheritdoc />
+        public ISettingsSource GetSettings()
+        {
+            string? type = BuildServerType;
+            Validates.NotNull(type);
```
RussKie, line 22: This is actually incorrect. `type` can be null, e.g. when no prior build server configuration is present. To repro remove/rename all GitExtensions.settings files at all levels (global, team, local).

UnitTests/GitCommands.Tests/Settings/MockData/level1_repogit_GitExtensions.settings.xml
RussKie: > The override of BuildServerType should be checked
> As the module provides functionality for settings for BuildServerTypes, it should test that.
> Like `BuildServer.Type=MyServer` and `BuildServer.Type.MyServer=MySetting`, test overrides.

This is asserted in `RepoDistributed_settings_should_return_expected`.

RussKie:
The integration tests can be extended further to verify the values are correctly persisted and read from the config. Also testing configuration overrides would be great to have (either as a standalone or/and via FormSettings). However this is beyond the scope.

UnitTests/CommonTestUtils/MEF/ExportProviderCache.cs
```diff
@@ -84,7 +81,7 @@ bool IsExcludedPart(ComposablePartDefinition part)
 
         public static IExportProviderFactory CreateExportProviderFactory(ComposableCatalog catalog, bool isRemoteHostComposition)
         {
-            var scope = _localCompositionScope;
+            var scope = new Scope("local");
```
RussKie, line 84: @sharwell I'm not quite sure about this one, though it got me to the green build. I couldn't find the correct way to dispose the test composition export provider.

RussKie:
@gerhardol I'm done with this... Back to 🛌

gerhardol:
Rebased and reordered the commit. One change, dictionary is restored to US English (we do not want commits in other languages). (As one unused using)
I will probably not add to the tests and merge this soon.

To easily see differences between branches like this is my favorite feature in next version...
image

IntegrationTests/UI.IntegrationTests/CommandsDialogs/SettingsDialog/Pages/BuildServerIntegrationSettingsPageTests.WithPlugins.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.VisualStudio.Composition;
+using NUnit.Framework;
+
+namespace UITests.CommandsDialogs.SettingsDialog.Pages
+{
+    [Apartment(ApartmentState.STA)]
+    public class BuildServerIntegrationSettingsPageTests_WithPlugins
+    {
+        // Created once for the fixture
+        ////private TestComposition _composition;
```
RussKie, line 24: I missed this.

RussKie (approved):

IntegrationTests/UI.IntegrationTests/CommandsDialogs/SettingsDialog/Pages/BuildServerIntegrationSettingsPageTests.WithPlugins.cs
gerhardol: will change before merging

gerhardol:
Reordered the test contents so it consistently is global->distributed->local to simplify reading (the path for global and distributed was switched, makes nodifference).
Added a check for reading effective settings as well.
I considered adding write tests for the write effective settings that was fixed in the PR but will not drag on, this need to be done.
Plan to squash and rebase merge tomorrow.

RussKie:
👍

Issue #4816, opened 2018-04-10
extend git notes support


tschlue:
I think the git notes functionality is a really usefull one to add more information or clarify wrong information in commit messages to commits.

At the moment the only support is to add notes.

What is missing is to fetch and to push notes to the remotes.

I didn't found any other issue for this.

drewnoakes:
Can you make a pull request?

tschlue:
My knowledge about the code is none existing and at the moment I didn't find the time to work into it.

GintasS:
@gerhardol Should we close this and your PR as well?

gerhardol:
Further discussions in #6039

PR #9786, opened 2022-01-02
GitNotes: Reduce overhead when Body exists

gerhardol:
Part of #4816 

## Proposed changes

A very minor optimisation to only fetch Notes and not Body when not needed.

There are only a few ms saved from this (so this is still more than most any code efficiency comment about LINQ etc but still hardly noticeable), the primary reason is to see this explicitly in the log and not check again why the Body/Subject is parsed over and over again.

## Screenshots 

First and last Git command is due to diff tab, the other are required when displaying a commit.

### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/CommitDataManager.cs
+++ b/GitCommands/CommitDataManager.cs
@@ -35,14 +35,15 @@ namespace GitCommands
         /// <summary>
         /// Updates the <see cref="CommitData.Body"/> (commit message) property of <paramref name="commitData"/>.
         /// </summary>
-        void UpdateBody(CommitData commitData, out string? error);
+        void UpdateBody(CommitData commitData, bool appendNotesOnly, out string? error);
     }
 
     public sealed class CommitDataManager : ICommitDataManager
     {
         private const string CommitDataFormat = "%H%n%T%n%P%n%aN <%aE>%n%at%n%cN <%cE>%n%ct%n%e%n%B";
         private const string CommitDataWithNotesFormat = "%H%n%T%n%P%n%aN <%aE>%n%at%n%cN <%cE>%n%ct%n%e%n%B%nNotes:%n%-N";
-        private const string BodyAndNotesFormat = "%H%n%B%nNotes:%n%-N";
+        private const string BodyAndNotesFormat = "%B%nNotes:%n%-N";
+        private const string NotesFormat = "%-N";
 
         private readonly Func<IGitModule> _getModule;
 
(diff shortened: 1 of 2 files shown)


GitCommands/CommitDataManager.cs
```diff
@@ -35,14 +35,15 @@ public interface ICommitDataManager
         /// <summary>
         /// Updates the <see cref="CommitData.Body"/> (commit message) property of <paramref name="commitData"/>.
         /// </summary>
-        void UpdateBody(CommitData commitData, out string? error);
+        void UpdateBody(CommitData commitData, bool bodyIsNull, out string? error);
```
RussKie, line 38:
```suggestion
        void UpdateBody(CommitData commitData, bool appendNotesOnly, out string? error);
```

GitCommands/CommitDataManager.cs
```diff
@@ -52,36 +53,35 @@ public CommitDataManager(Func<IGitModule> getModule)
         }
 
         /// <inheritdoc />
-        public void UpdateBody(CommitData commitData, out string? error)
+        public void UpdateBody(CommitData commitData, bool onlyNotes, out string? error)
```
RussKie, line 56:
```suggestion
        public void UpdateBody(CommitData commitData, bool appendNotesOnly, out string? error)
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -328,7 +328,7 @@ void UpdateCommitMessage()
 
                 if (_revision.Body is null || (AppSettings.ShowGitNotes && !_revision.HasNotes))
                 {
-                    _commitDataManager.UpdateBody(data, out _);
+                    _commitDataManager.UpdateBody(data, _revision.Body is not null, out _);
```
RussKie, line 331:
```suggestion
                    _commitDataManager.UpdateBody(data, appendNotesOnly: _revision.Body is not null, out _);
```

gerhardol: ok

GitCommands/CommitDataManager.cs
gerhardol: ok

GitCommands/CommitDataManager.cs
gerhardol: ok

gerhardol:
@RussKie Any more comments?

RussKie (approved):

RussKie:
All good. Thank you for the ping.

PR #9788, opened 2022-01-02
FileHistory: Launch commandline in FormBrowse

gerhardol:
#9445 uses the FormBrowse for displaying file history
(by default, configurable with AppSettings.UseBrowseForFileHistory).
FormFileHistory is deprecated.

However, command line invocation of filehistory still used FormFileHistory,
not using the setting. This is needed for all 'external' use like
Explorer and Visual Studio.

## Screenshots 

FormBrowse with path filter is opened instead of FormFileHistory

## Test methodology 

Tests are adopted to run with FormFileHistory, same as today.
There are other tests already for FormBrowse.

## Merge strategy

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1416,7 +1416,7 @@ namespace GitUI
                     return RunBlameCommand(args);
                 case "branch":
                     return StartCreateBranchDialog();
-                case "browse":      // [path] [-filter]
+                case "browse":      // [path] [--pathFilter=filname] [-filter] [-commit=selected[,first]]
                     return RunBrowseCommand(args);
                 case "checkout":
                 case "checkoutbranch":
@@ -1725,8 +1725,22 @@ namespace GitUI
                 filterByRevision = true;
             }
 
-            ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
+            // Similar to StartFileHistoryDialog(), tests works better with this setup
+            if (AppSettings.UseBrowseForFileHistory.Value)
+            {
+                ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
+                                () => new FormBrowse(commands: this, new BrowseArguments
+                                {
+                                    RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
+                                    PathFilter = fileHistoryFileName,
+                                    SelectedId = revision?.ObjectId
+                                }));
+            }
+            else
+            {
+                ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
                 () => new FormFileHistory(commands: this, fileHistoryFileName, revision, filterByRevision, showBlame));
+            }
 
             return true;
         }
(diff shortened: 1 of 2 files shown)


GitUI/GitUICommands.cs
```diff
@@ -1725,8 +1725,22 @@ private bool RunFileHistoryCommand(IReadOnlyList<string> args, bool showBlame)
                 filterByRevision = true;
             }
 
-            ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
+            // Similar to StartFileHistoryDialog(), tests works better with this setup
```
RussKie, line 1728: Did you mean this?
```suggestion
            // Similar to StartFileHistoryDialog(), tests work better with this setup
```

GitUI/GitUICommands.cs
```diff
@@ -1725,8 +1725,22 @@ private bool RunFileHistoryCommand(IReadOnlyList<string> args, bool showBlame)
...
+                                    RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
+                                    PathFilter = fileHistoryFileName,
+                                    SelectedId = revision?.ObjectId
+                                }));
+            }
+            else
+            {
+                ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
                 () => new FormFileHistory(commands: this, fileHistoryFileName, revision, filterByRevision, showBlame));
+            }
```
RussKie, line 1743:
```suggestion
            else
            {
                ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
                                 () => new FormFileHistory(commands: this, fileHistoryFileName, revision, filterByRevision, showBlame));
            }
```

GitUI/GitUICommands.cs
```diff
@@ -1725,8 +1725,22 @@ private bool RunFileHistoryCommand(IReadOnlyList<string> args, bool showBlame)
...
+            // Similar to StartFileHistoryDialog(), tests works better with this setup
+            if (AppSettings.UseBrowseForFileHistory.Value)
+            {
+                ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
+                                () => new FormBrowse(commands: this, new BrowseArguments
+                                {
+                                    RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
+                                    PathFilter = fileHistoryFileName,
+                                    SelectedId = revision?.ObjectId
+                                }));
```
RussKie, line 1737: Off by one
```suggestion
                                 () => new FormBrowse(commands: this, new BrowseArguments
                                 {
                                     RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
                                     PathFilter = fileHistoryFileName,
                                     SelectedId = revision?.ObjectId
                                 }));
```

RussKie (commented):
LGTM

GitUI/GitUICommands.cs
gerhardol: Removing second part, this should not be launched but be opened as a modeles form

GitUI/GitUICommands.cs
gerhardol: Can we get StyleCop to handle this?

RussKie: I'm not good enough? :)
I'm not aware of any rules for this situation.

RussKie (approved):

PR #9789, opened 2022-01-03
Updated tooltip for ShowDiffForAllParents

gerhardol:
Follow up to #9720 

## Proposed changes

The diffs shown in Diff Tab when multi-selecting commits is slightly updated in master, missed in #9720 

Also, a lot of strings used with TranslatedStrings were translated twice.
No simple way what I see to hardcode this in TranslationApp, so changing this by renaming controls.

### Before

image

### After

image

## Test methodology 

Compiling

## Merge strategy

----

### DIFF
--- a/GitUI/TranslatedStrings.cs
+++ b/GitUI/TranslatedStrings.cs
@@ -84,12 +84,13 @@ namespace GitUI
 
 - For a single selected commit, show the difference with its parent commit.
 - For a single selected merge commit, show the difference with all parents.
-- For two selected commits with a common ancestor (BASE):
-   - Show the difference between the commits.
-   - The difference of unique files from BASE to each of the selected commits.
-   - The difference of common files (identical changes) from BASE to the commits.
-- For multiple selected commits (up to four), show the difference for all the first selected with the last selected commit.
-- For more than four selected commits, show the difference from the first to the last selected commit.");
+- For two selected commits with a common ancestor (BASE), show the difference
+between the commits as well as the difference from BASE to the selected commits.
+See documentation for more details about icons and range diffs.
+- For multiple selected commits (up to four), show the difference for
+all the first selected with the last selected commit.
+- For more than four selected commits, show the difference from the first to
+the last selected commit.");
 
         private readonly TranslationString _stageSelectedLines = new("Stage selected line(s)");
         private readonly TranslationString _unstageSelectedLines = new("Unstage selected line(s)");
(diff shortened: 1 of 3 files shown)


GitUI/TranslatedStrings.cs
```diff
@@ -84,12 +84,13 @@ internal sealed class TranslatedStrings : Translate
...
 - For a single selected commit, show the difference with its parent commit.
 - For a single selected merge commit, show the difference with all parents.
-- For two selected commits with a common ancestor (BASE):
-   - Show the difference between the commits.
-   - The difference of unique files from BASE to each of the selected commits.
-   - The difference of common files (identical changes) from BASE to the commits.
-- For multiple selected commits (up to four), show the difference for all the first selected with the last selected commit.
-- For more than four selected commits, show the difference from the first to the last selected commit.");
+- For two selected commits with a common ancestor (BASE), show the difference
+between the commits as well as the difference from BASE to the selected commits.
```
RussKie, line 88: Do we really need to wrap the lines?

gerhardol: The existing tooltip was too wide already for lower resolution screens.
I find it too wide also on my second screen, 1600x1200 and unusable on my old laptop (1280x720?).

gerhardol:
For strings translated twice: These could be added to list in the translation app instead of prefixing them with `_NO_TRANSLATE_`
Harder to discover that though.

RussKie:
I'd like to move away from our current translation implementation and the
translation app in favour of xliff tasks. This will streamline the process
for contributors, and speed up the build process.
 I made some inroads, but it isn't a straightforward task though, and
requires some effort.

gerhardol:
<< These could be added to list in the translation app instead of prefixing them with _NO_TRANSLATE_

> I'd like to move away from our current translation implementation

How to handle this is a short term solution. Dead strings to translate, _NO_TRANSLATE_ or list in translation app.

RussKie:
There's little overhead to have dups - those get auto translated. I'm not
saying it is not an issue, but I wouldn't stress too much about it.
Your approach with the prefix works too.

gerhardol:
> There's little overhead to have dups - those get auto translated. I'm not saying it is not an issue, but I wouldn't stress too much about it. Your approach with the prefix works too.

You still have to handle each occurrence, even if there is a suggestion in Transifex.

But I am dropping the duplications right now, it should be better to fix the translation app than to use _NO_TRANSLATE_
Changes in the xlf can be blocking.

Adding one minor casing though.

RussKie (approved):

PR #9790, opened 2022-01-03
Do not report Git error for a command seemingly executed in the background

gerhardol:
## Proposed changes

git-ls-files is executed every time GE is reactivated, to determine if "InTheMiddleOfAMerge is to be displayed.
If this fails, there is nothing wrong with GE (and really not Git), no interest to throw a popup for the user.

This is especially annoying with WSL. If the Linux VM is busy will WSL occasionally time out and you get the popup regularily.
After this command, the risk for timeouts is lower, the user will get notification when actually performing actions.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -513,14 +513,16 @@ namespace GitCommands
             return _gitExecutable.RunCommand(arguments, createWindow: createWindow);
         }
 
-        public bool InTheMiddleOfConflictedMerge()
+        public bool InTheMiddleOfConflictedMerge(bool background = false)
         {
             GitArgumentBuilder args = new("ls-files")
             {
                 "-z",
                 "--unmerged"
             };
-            var result = _gitExecutable.Execute(args);
+
+            // Do not report errors for commands called in the background
+            ExecutionResult result = _gitExecutable.Execute(args, throwOnErrorExit: !background);
             return result.ExitedSuccessfully && !string.IsNullOrEmpty(result.StandardOutput);
         }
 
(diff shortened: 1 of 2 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -513,14 +513,16 @@ public bool EditNotes(ObjectId commitId)
             return _gitExecutable.RunCommand(arguments, createWindow: createWindow);
         }
 
-        public bool InTheMiddleOfConflictedMerge()
+        public bool InTheMiddleOfConflictedMerge(bool background = false)
```
RussKie, line 516:
```suggestion
        public bool InTheMiddleOfConflictedMerge(bool throwErrors = true)
```

gerhardol: If it should be aligned to the GitProcess: Renaming to `throwOnErrorExit`

RussKie: I don't think this is necessary. `throwOnErrorExit` has meaning within `GitProcess` implementation, i.e. throw an exception if the process faulted. However in the context of `InTheMiddleOfConflictedMerge` (and especially its callers) there is clear link to the underlaying process. Reading this code I get multiple questions `hasConflicts = Module.InTheMiddleOfConflictedMerge(throwOnErrorExit: false);` - e.g., "exit" of what? Hence I proposed `throwErrors`.

gerhardol: I find `throwErrors` in isolation be more confusing, for what occasion are errors thrown?
`throwOnErrorExit` has information that this is limited to "something".

`throwOnErrorExit: false` is only set in one "background" operation right now, hopefully we can keep it at that.
This is the type of operations where an external operation (WSL command, not even Git) returns an error that GE can handle.
Maybe the user should be notified but not at every occurrence.
The solution in this case is for Microsoft to fix the WSL implementation.

PR #9795, opened 2022-01-05
make ToolStripEx use ToolStripExSystemRenderer by default as before #9608

NikolayXHD:
Achieve the goal of #9608 without unintentionally making `ToolStripEx` use `ProfessionalRenderer` instead of `SystemRenderer`

In particular, this makes `FormBrowse` toolbars selected buttons be highlighted properly, as before.

after #9608
image

before #9608 and after this PR
image

### DIFF
--- a/GitExtUtils/GitUI/ToolStripExSystemRenderer.cs
+++ b/GitExtUtils/GitUI/ToolStripExSystemRenderer.cs
@@ -1,0 +1,24 @@
+﻿using System.Windows.Forms;
+
+namespace GitUI
+{
+    public sealed class ToolStripExSystemRenderer : ToolStripSystemRenderer
+    {
+        protected override void OnRenderMenuItemBackground(ToolStripItemRenderEventArgs e)
+        {
+            if (e.ToolStrip.GetMenuItemBackgroundFilter()?.ShouldRenderMenuItemBackground(e) != false)
+            {
+                base.OnRenderMenuItemBackground(e);
+            }
+        }
+
+        protected override void OnRenderToolStripBorder(ToolStripRenderEventArgs e)
+        {
+            if (e.ToolStrip is not IToolStripEx { DrawBorder: false })
+            {
+                // render border
+                base.OnRenderToolStripBorder(e);
+            }
+        }
+    }
+}
--- a/GitExtUtils/GitUI/Theming/ThemeFix.cs
+++ b/GitExtUtils/GitUI/Theming/ThemeFix.cs
@@ -74,15 +74,14 @@ namespace GitExtUtils.GitUI.Theming
 
         private static void SetupToolStrip(ToolStrip strip)
         {
-            strip.EnableTheming(enable: true);
-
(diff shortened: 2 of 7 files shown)


GitExtUtils/GitUI/ToolStripExSystemRenderer.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace GitUI
 {
-    internal sealed class ToolStripExRenderer : ToolStripSystemRenderer
+    public sealed class ToolStripExSystemRenderer : ToolStripSystemRenderer
```
RussKie, line 5: Does it need to be public?

RussKie (approved):
LGTM, haven't run

gerhardol:
Looks good, but a few buttons are not changed.
image

gerhardol (commented):
+1
No objections to the solution (looks good, have not enough knowledge to question the use of Renderer).
Ont comment is the buttons that ar not changed for me if not system-defined visuals is used.

GitExtUtils/GitUI/ToolStripExSystemRenderer.cs
NikolayXHD: It became public so it could be [called](https://github.com/gitextensions/gitextensions/pull/9795/files#diff-81d35d9561cd05f67e9b83ae32c98f4398fcff5e2377227227e0b77d207f5066R17) it in `GitUI/UserControls/ToolStripEx.cs` construstor

NikolayXHD:
> Looks good, but a few buttons are not changed for me.

Thank you for noticing, this is unexpected to me, looking at it.

NikolayXHD:
> > Looks good, but a few buttons are not changed for me.
> 
> Thank you for noticing, this is unexpected to me, looking at it.

fixed

gerhardol:
If not default or system-defined-visuals is set, buttons look consistent but like the following:

image

It was not what I expected at first, but it is probably how it should be.

Related: `default` theme should set system-defined but disabled to indicate that this is always used.
(colorblind could maybe be enabled too, any reason to disable that button?)

image

NikolayXHD:
> colorblind could maybe be enabled too, any reason to disable that button?

The idea was that "colorblind" works best with "use system visual style" unchecked, so that the color transformation is applied more consistently. On close look, maybe that was not a good idea, because the controls with "system defined style" will never display anything green, because Windows SystemColors already take colorblindness into account.

RussKie:
What about the revision grid where we do use red and green?

gerhardol:
> What about the revision grid where we do use red and green?

This is the application colors where colorblind is defined (the system colors are not changed since themeing was disabled).
It would make sense to allow changing them also for default.

I tried a quick shot at it, it is not trivial and variants are not expected for default themes.
I would have had expected that default values also for the default theme is available in the application, but it is parsed.

NikolayXHD:
#9799

PR #9800, opened 2022-01-07
Adjust BrowseRepoSettings

gerhardol:
## Proposed changes

Minor layout issue.
Opened in designer too

### Before

image

### After

image

## Test methodology 

manual

## Merge strategy

----


RussKie (commented):
I suggest rewording the text to something like "Show file history in the main window".
1. Use sentence case
2. Don't use internal type names. An average user shouldn't know how we name our types/dialogue.

gerhardol:
> I suggest rewording the text to something like "Show file history in the main window".
> 
>     1. Use sentence case
> 
>     2. Don't use internal type names. An average user shouldn't know how we name our types/dialogue.

1. Changed in #9789 

2. FileHistory was used as that is what the user see in 3.5 when opening the file history.
Browse is used in the title for the page too.

Will change in #9789 (using suggestion), just keep the layout change here (there are enough conflicts in English.xlf already).

RussKie:
👍

Issue #9796, opened 2022-01-05
Feedback with new Blame and File history in next version


pmiossec:
### Environment

- Git Extensions 33.33.33
- Build 688dc11364f343a29322b6dfa86450756d74fbe6
- Git 2.34.0.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET 5.0.13
- DPI 192dpi (200% scaling)

### Issue description

I'm really lost with the new behavior of the Blame and file history. So it will be a mix of feedback and questions.

## File History

When I tested it multiple time quickly during the last months, I always thought there was a big regression or that the feature was never finished. I just now noticed that the new GitExtensions instance display the history filtered on the file. So for me that's not really clear (especially if the file is modified often)

What are not as good as before for me:

1. Before it was a clearly identifiable different view. Now it's difficult to distinguish to a normal instance

2. There is a lot less space to see the diff and blame. And you always have to move the split separator. Sometimes a diff was difficult to read, so now that you can't see the whole file and have a blame view, it will be worst :(
3. You loose the "View" tab (that I sometimes I preferred to use because I found the blame aggressive when you move your mouse and the grey selection change)
4. You no more are able to identify quickly which instance is file history (and when you want to switch quickly and often during an investigation or when you want to clean the instances opened, its a pain) lost in all similar windows

![image](https://user-images.githubusercontent.com/460196/148282284-c349e0b8-a156-4c1d-8371-d637bc87a20e.png)

I think we need to reproduce this behavior (ie. changing the title of the application) when the "Path filter" is enabled

5. Maybe when we open a new instance to enable the "Path filter", the left panel should be hidden
6. I don't know if it make sense also to filter the file list also
7. Perf: It takes more time to open the file history in another window. And because it's a full GitExtension instance, it also fetch in background if enable, do the `git status` that could take a lot of CPU on big repositories.
8. Q: Is there a possibility to open the old window? I though we would keep it...
9. I'm pretty sure the "Filtering" icon is not enough visible for legacy user that will have it enabled when doing "File History" as usual (and that we will have some issues opened). And a dropdown menu (like the stash button) to disable quickly the filters could be an improvement...

## Blame 

10. When blame is enabled, it's strange the way to come back to the "diff" view. You need to click again on the file.
11. When you use "Blame this revision" or "Blame previous revision", you come back to the diff view
[...]

PR #9801, opened 2022-01-07
Add Blame Settings page

gerhardol:
Part of #9796

## Proposed changes

Blame settings is currently only in FileHistory. They are a little hidden there, hard to find them.
With master Blame is also in Browse so they need to be accessible.

Another option would have been to add them to the FileViewer context menu.
They would have to be added last in the menu to avoid messing with the FileViewer modes and the logic would have to be added to the blame control too.
Also if that is considered, it is expected to have this in Settings.

Designer opened, tab order tested

### Before

Unchanged

image

### After

image

## Test methodology 

Manual

## Merge strategy

----


RussKie (approved):

mdonatas:
When it was tested was it look at with higher DPI? 200% here
image

gerhardol:
> When it was tested was it look at with higher DPI? 200% here image

125% looks fine
More resources are needed in the project for everything...

PR #9804, opened 2022-01-07
RevDiff: Persist Blame until another file is selected

gerhardol:
Part of #9796

## Proposed changes

#9424 added Blame to RevDiff
The intention was that the Blame should be short term and diff be shown at every change as that is the intention of the Diff tab.
(Longer term in FileTree.)

However @pmiossec (as well as @RussKie somehow in the review) commented on the behavior to be to short, that you cannot follow the file.

The behavior is changed to end Blame view when the selected file is changed.
It does not matter if a file is explicitly selected or if it cannot be "followed" when changing a commit (explicitly or "blame this revision").

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -48,6 +48,10 @@ namespace GitUI.CommandsDialogs
         private readonly RememberFileContextMenuController _rememberFileContextMenuController
             = RememberFileContextMenuController.Default;
         private Action? _refreshGitStatus;
+        private FileStatusItem? _selectedBlameItem;
+        private string? _fallbackFollowedFile;
+        private FileStatusItem? _lastExplicitlySelectedItem;
+        private bool _isImplicitListSelection = false;
 
         public RevisionDiffControl()
         {
@@ -228,7 +232,9 @@ namespace GitUI.CommandsDialogs
         public void DisplayDiffTab(IReadOnlyList<GitRevision> revisions)
         {
             SetDiffs(revisions);
-            if (DiffFiles.SelectedItem is null)
+
+            // Select something by default, except range diff
+            if (DiffFiles.SelectedItem is null && !(DiffFiles.FirstGroupItems.Count() == 1 && DiffFiles.FirstGroupItems.FirstOrDefault().Item.IsRangeDiff))
             {
                 DiffFiles.SelectFirstVisibleItem();
             }
@@ -239,25 +245,47 @@ namespace GitUI.CommandsDialogs
         /// When switching commits, the last selected file is "followed" if available in the new commit,
         /// this file is used as a fallback.
         /// </summary>
-        public string? FallbackFollowedFile { get; set; } = null;
+        public string? FallbackFollowedFile
+        {
+            get => _fallbackFollowedFile;
+            set
+            {
+                _fallbackFollowedFile = value;
+                _lastExplicitlySelectedItem = null;
+            }
+        }
 
         private void SetDiffs(IReadOnlyList<GitRevision> revisions)
         {
             Validates.NotNull(_revisionGrid);
 
-            var item = DiffFiles.SelectedItem;
-            var oldDiffItem = DiffFiles.FirstGroupItems.Contains(item) ? item : null;
+            FileStatusItem prevSelectedItem = DiffFiles.SelectedItem;
+            FileStatusItem prevDiffItem = DiffFiles.FirstGroupItems.Contains(prevSelectedItem) ? prevSelectedItem : null;
             DiffFiles.SetDiffs(revisions, _revisionGrid.CurrentCheckout);
 
-            // Try to restore previous item
-            if (oldDiffItem is not null && DiffFiles.FirstGroupItems.Any(i => i.Item.Name.Equals(oldDiffItem.Item.Name)))
+            _isImplicitListSelection = true;
+
+            // First try the last item explicitly selected
+            if (_lastExplicitlySelectedItem is not null
+                && DiffFiles.FirstGroupItems.FirstOrDefault(i => i.Item.Name.Equals(_lastExplicitlySelectedItem.Item.Name))?.Item is GitItemStatus explicitItem)
             {
-                DiffFiles.SelectedGitItem = oldDiffItem.Item;
+                DiffFiles.SelectedGitItem = explicitItem;
+                return;
             }
-            else if (!string.IsNullOrWhiteSpace(FallbackFollowedFile)
-                && DiffFiles.FirstGroupItems.Where(i => i.Item.Name.Equals(FallbackFollowedFile)).FirstOrDefault()?.Item is GitItemStatus listItem)
+
+            // Second go back to the filtered file
+            if (!string.IsNullOrWhiteSpace(FallbackFollowedFile)
+                && DiffFiles.FirstGroupItems.FirstOrDefault(i => i.Item.Name.Equals(FallbackFollowedFile))?.Item is GitItemStatus fallbackItem)
             {
-                DiffFiles.SelectedGitItem = listItem;
+                DiffFiles.SelectedGitItem = fallbackItem;
+                return;
+            }
+
+            // Third try to restore the previous item
+            if (prevDiffItem is not null
+                && DiffFiles.FirstGroupItems.FirstOrDefault(i => i.Item.Name.Equals(prevDiffItem.Item.Name))?.Item is GitItemStatus prevItem)
+            {
+                DiffFiles.SelectedGitItem = prevItem;
             }
         }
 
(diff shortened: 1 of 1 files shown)


gerhardol:
@pmiossec Any comments?

pmiossec:
> @pmiossec Any comments?

This PR is an improvement (the blame is kept) but because it's still not working well for me, I have difficulties to understand exactly what you want to achieve and give a clear feedback...

>The behavior is changed to end Blame view when the selected file is changed.

1. It seems to not work well **and need to be fixed by disabling the blame**:

blame_switch

2. When the 'Filter file in grid' is selected, the Blame feature seems to work well 👍 

But if this option is not selected, the Blame feature seems completely broken:

* You can't use the revision grid because if the file is not in the commit selected, you will loose the file you are blaming
* "Blame previous revision" is broken also (an acceptable solution could be to filter on the file only when the feature is used before selecting the previous revision)

And that's what worries me because it makes the feature very difficult to use it rightly. I don't know how other use the blame feature but 90% of the time I have to blame previous revision and so need to filter on the file.

And now, you need to filter the file **before** doing the blame and so understand that you must do 2 actions whereas before it was done automatically for you...

For me, when we enable the "Blame mode", we must filter also on the file.

Proposition: **At least, we absolutely need to fix the fact that filtering on the file when blame is already enabled should stay in blame mode**.

Here is the current behavior that is not satisfying for me:
blame_then_filter

Another point making me think that when blaming a file, we should before filter on it is that the behavior is also broken if the filter is already enabled on another file.

3. Another not really satisfying behavior is that I like to switch to the diff and go back to the blame mode when I do a blame to see what is the exact diff made in the selected commit and switch between the diff and the blame is not working well because the file filter get the focus when trying the enable/disable with <kbd>b</kbd>:

blame_filter_file

gerhardol:
Pushed an update for this, still in draft to await #9816
But I believe the functionality is there.
I have not planned other changes for #9796 after this PR

* Goto-blame-line from RevDiff to FileTree implemented (was previously only within the forms). A little more important after #9831 

* Try harder to (keep) selection of the interesting file item. When switching revisions the following file in the first group is selected (based on name) in the following priority order.
  - Last clicked file
  - The file currently filtered
  - The last file currently default selected (from next prio)
  - The first file in the list, unless it is the RangeDiff (always need an explicit click, both to not show a diff at a glance after range-diff was moved to first group in #9757, the command can require some cpu time too).

* Stay on Blame until the file cannot be shown anymore (as the file must be clicked to be blamed, it has the highest priority).

mstv (commented):
"Explicitly clicked" should rather be "explicitly selected", i.e. should work for keyboard, too.
Perhaps it should also be set, if Blame is requested (because there is no need to select a file for single-file commits).

> Stay on Blame until the file cannot be shown anymore (as the file must be clicked to be blamed, it has the highest priority).

It always falls back to diff (at least in combination with the other open PRs).

gerhardol:
> "Explicitly clicked" should rather be "explicitly selected", i.e. should work for keyboard, too. Perhaps it should also be set, if Blame is requested (because there is no need to select a file for single-file commits).

What event provides this information?
(The variable will then be renamed.)

"last clicked/selected" could be set when an action is done on the file, could be seen as "selected".
But not for single file commits, then traversing commits would reset the "last selected" if there happen to be a commit with a single file.
(Well if "last selected" is null it could be set but now it getting complex...)
"last selected" is intentionally not reset when refreshing or opening a new repo, you may still be interested in the last selection... 
 
> > Stay on Blame until the file cannot be shown anymore (as the file must be clicked to be blamed, it has the highest priority).
> 
> It always falls back to diff (at least in combination with the other open PRs).

Embarrassing, fixed..

mstv:
> > "Explicitly clicked" should rather be "explicitly selected", i.e. should work for keyboard, too. Perhaps it should also be set, if Blame is requested (because there is no need to select a file for single-file commits).
> 
> What event provides this information? (The variable will then be renamed.)
> 
> "last clicked/selected" could be set when an action is done on the file, could be seen as "selected".

I think `KeyUp` in the FileStatusList should work (in addition to `Click`).

> But not for single file commits, then traversing commits would reset the "last selected" if there happen to be a commit with a single file.

👍

> > > Stay on Blame until the file cannot be shown anymore (as the file must be clicked to be blamed, it has the highest priority).
> > 
> > It always falls back to diff (at least in combination with the other open PRs).
> 
> Embarrassing, fixed..

Works.

gerhardol:
@pmiossec I believe your comments are handled, can you check?

gerhardol:
@mstv I made a fix for 
https://github.com/gitextensions/gitextensions/pull/9842#pullrequestreview-872433486

There are maybe better ways to do it...

gerhardol:
@pmiossec @mstv Any further comments?

mstv:
`_lastClickedItem` - emphasis on "clicked" - breaks my workflow in unfiltered mode following a file over several commits. The file position is lost if I forget to explicitly click the file.

gerhardol:
> `_lastClickedItem` - emphasis on "clicked" - breaks my workflow in unfiltered mode following a file over several commits. The file position is lost if I forget to explicitly click the file.

I cannot  see how clicking or not affects this, that is just the prio order in the selection.

Currently, only the clicked item, not the position is retained. (that will require feature/i9796-blame-line-to-diff that I am not sure we want.)
Without double checking, I believe FileViewer tries to retain the the line in a file.

pmiossec:
> @pmiossec @mstv Any further comments?

I did had the time to test it extensively (and won't be able to do it I  the next to weeks) but the little I give it a try, it seems to behave better.

I will test it more extensively later once merged to see if there still improvements needed.

Nothing blocking the merge for me...

gerhardol:
> Nothing blocking the merge for me...

But I do not have any approvals...

mstv:
> > `_lastClickedItem` - emphasis on "clicked" - breaks my workflow in unfiltered mode following a file over several commits. The file position is lost if I forget to explicitly click the file.
> 
> I cannot see how clicking or not affects this, that is just the prio order in the selection.

The prioritization itself is the problem. I don't want to loose the file selection file on refresh (nor on selection of another commit).
Only when filtered for a single file, I would reselect that file whenever possible.

`Click`

grafik

`Cursor up` (I use the cursor keys and the scroll wheel when reviewing diffs. Only the pointer fingers move.)

grafik

`F5`

grafik

---

Another issue (while the commit which I ran for these screenshots, is checked out):

grafik

`F5`

grafik

gerhardol:
> > > `_lastClickedItem` - emphasis on "clicked" - breaks my workflow in unfiltered mode following a file over several commits. 

> The prioritization itself is the problem. I don't want to loose the file selection file on refresh (nor on selection of another commit). Only when filtered for a single file, I would reselect that file whenever possible.

So you want move by arrow to be handled as click.
I do not know how to distinguish this from "another file selected as the file was not available in the commit".
So if click handling is removed, follow will still not work in unfiltered mode. (You can make examples where it worked before, but if the intermediate commit does not have the implicitly selected file, a new file will be implicitly selected).

Maybe "select by arrow" can be specially handled too.

The handling in this PR solves part of the handling though.
I can remove the click handling but I do not really believe it improves your use case.
But I rather do that to get the first part of this merged, to get @pmiossec etc feedback on usage too.

> Another issue (while the commit which I ran for these screenshots, is checked out):

Did the commit change?
How is this related?

gerhardol:
> A "follow mode" needs to be made explicit in unfiltered mode IMO. Perhaps explicit activation, e.g. click with modifier key, and an indicator that it is active. It always is very surprising to me when I loose the selected file because I had _clicked_ a file ten commits ago.
> 
> > Maybe "select by arrow" can be specially handled too.
> 
> Yes, all kinds of _manual_ selection should affect the follow mode, i.e. at least on click and on arrow keys.
> 
> > but if the intermediate commit does not have the implicitly selected file, a new file will be implicitly selected).
> 
> I got used to it.

This PR also have some handling for selection focus, that will stay.

The core problem is to distinguish from explicit selections (when afile is clicked (OK) or selected by keys (not known now)).

If the click follow mode is reverted, the first priority for selection will go back to selection, without determining if it is implicit or explicit selected, with "filtered file" as second priority.
For select-by-key the selection is slightly stickier as there is one less alternative if the clicked and selected is in the same commit, but it will still not resume after not finding the file (the implicitly followed file will be followed instead).

@gitextensions/git-extensions-team can I get votes to keep follow more or less as is in master or to do as in this PR and handle click but not selection-by-keys?

RussKie:
> can I get votes to keep follow more or less as is in master or to do as in this PR and handle click but not selection-by-keys?

The Windows Forms teams spends a very significant amount of time to ensure the Windows Forms SDK meets accessibility requirements, and mouse and keyboard operations are equivalent, to ensure a consistent user experience for everyone, including users with accessibility needs (that is, those we can't use a mouse for some reason, e.g., with mobility limitations or vision impairments, etc.). 
I can't agree with different behaviours that depend on the input methods.

gerhardol:
Implemented a hackier solution where the implicit selections after changing revisions and updating artificial is ignored, other selection changes are handled as explicit.

mstv (approved):

gerhardol:
Rebased with a indentation correction.

PR #9805, opened 2022-01-07
Blame: Start on line also when control already exists

gerhardol:
(This was seen when preparing #9424, no separate issue exists.)

## Proposed changes

Blame can start presenting on a line. This is is available in
RevDiff and RevFileTree since #9424.
(Even if the code enabling this has existed for long time.)

The "goto line" was only used if the control had to be init,
not if e.g. toggling to Diff and selecting an new diff line
and then Blame again.

Before #9804 This has to be tested in RevFileTree

### Before

Before line 52 would have been displayed also the second time blame was activated

### After

First opening Blame on line 53, then on 152
Note that the "make visible" line is hidden in this screenshot, should be in the middle of the screen. This is standard FileViewer behavior.

blame-line

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -714,7 +714,7 @@ namespace GitUI.Editor
                 var viewPosition = _currentViewPosition;
                 if (isDiff && _viewer.GetLineText(0) == viewPosition.FirstLine && viewPosition.ActiveLineNum is not null)
                 {
-                    // prefer the RightLineNum that is for the currentl revision
+                    // prefer the RightLineNum that is for the current revision
                     return viewPosition.ActiveLineNum.RightLineNumber != DiffLineInfo.NotApplicableLineNum
                         ? viewPosition.ActiveLineNum.RightLineNumber
                         : viewPosition.ActiveLineNum.LeftLineNumber;
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -713,10 +713,10 @@ namespace GitUI.Editor
                 var viewPosition = _currentViewPosition;
                 if (isDiff && _viewer.GetLineText(0) == viewPosition.FirstLine && viewPosition.ActiveLineNum is not null)
                 {
-                    // prefer the LeftLineNum because the base revision will not change
-                    return viewPosition.ActiveLineNum.LeftLineNumber != DiffLineInfo.NotApplicableLineNum
-                        ? viewPosition.ActiveLineNum.LeftLineNumber
-                        : viewPosition.ActiveLineNum.RightLineNumber;
+                    // prefer the RightLineNum that is for the currentl revision
+                    return viewPosition.ActiveLineNum.RightLineNumber != DiffLineInfo.NotApplicableLineNum
+                        ? viewPosition.ActiveLineNum.RightLineNumber
+                        : viewPosition.ActiveLineNum.LeftLineNumber;
                 }
 
                 // Convert from offset to line number
--- a/GitUI/Editor/FileViewerInternal.cs
+++ b/GitUI/Editor/FileViewerInternal.cs
@@ -452,26 +452,33 @@ namespace GitUI.Editor
 
         public void GoToLine(int lineNumber)
         {
-            TextEditor.ActiveTextAreaControl.Caret.Position = new TextLocation(0, GetCaretLine(lineNumber, rightFile: true));
+            TextEditor.ActiveTextAreaControl.Caret.Position = new TextLocation(0, GetCaretOffset(lineNumber, rightFile: true));
         }
 
-        private int GetCaretLine(int lineNumber, bool rightFile)
+        /// <summary>
+        /// Convert the line number to the offset for the caret.
+        /// </summary>
+        /// <param name="lineNumber">The line number seen in the editor.</param>
+        /// <param name="rightFile">If the line number for the right file is preferred.</param>
+        /// <returns>The caret offset</returns>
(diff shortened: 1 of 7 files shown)


RussKie (commented):
I'd like to have tests for "new blame" functionality in general. It has gaps (like this one) which cause frustration to users (e.g. @pmiossec and myself).

gerhardol:
> I'd like to have tests for "new blame" functionality in general. It has gaps (like this one) which cause frustration to users (e.g. @pmiossec and myself).

This is not new functionality

gerhardol:
@RussKie @mstv I have tried to add tests for the initial line control, but fail on awaiting the internal blame viewer to be updated. Can you please assist?

The commit to make git-blame cancellable (the command requires 10s for English.xlf for me) could be separated to another PR, in here for now.

gerhardol:
The core change in this PR is simple and could be merged with a review ( Blame: Start on line also when control already exists )
Since RussKie asked for tests too, a lot more has to be done
Some async changes required is handled in #9816

RussKie:
I'll try to take a look in coming days.

gerhardol:
> I'll try to take a look in coming days.

I am trying, that why I am grumpy. 
Can push what I fail with

gerhardol:
> > I'll try to take a look in coming days.
> 
> I am trying, that why I am grumpy. Can push what I fail with

Pushed.

Module.Blame need to be mocked which I have failed with. 
The tests is probably what I would like to add though.

mstv (commented):
LGTM if rebased again on #9816

Regarding the tests:
Why do you want to mock `Module.Blame()` and have injected a `MockExecutable` into the `_referenceRepository.Module`?
I would rather create real commits in the test repo.

gerhardol:
> Regarding the tests: Why do you want to mock `Module.Blame()` and have injected a `MockExecutable` into the `_referenceRepository.Module`? I would rather create real commits in the test repo.

The MockExecutable was just a way I was trying to get something working. That is not the way.
I would prefer to mock Module.Blame output rather than setting up a reference repo here, the output should be a little longer than default. But it may come down to a reference repo.

This change is useful when trying other changes for gotoline in Blame and I will open a separate PR with the original change to simplify testing and decrease merge conflicts.

gerhardol:
Tests added, using a dummy repo.
The current implemtation was one line off.

CommitData required ParentId to be not null so the test  would fail.
I changed it to handle null, as it is a valid value. 
(Hiding commit info in the test is possible too, or set ParentIds in the tests, but I find handling null is correct.)

GitUI/UserControls/BlameControl.cs
```diff
@@ -112,23 +112,22 @@ public async Task LoadBlameAsync(GitRevision revision, IReadOnlyList<ObjectId>?
                 return;
             }
 
+            int line = _clickedBlameLine is not null && _clickedBlameLine.Commit.ObjectId == objectId
+                ? _clickedBlameLine.OriginLineNumber
+                : initialLine ?? (fileName == _fileName ? BlameFile.CurrentFileLine : 1);
```
gerhardol, line 117: A change in behavior:
Before the scroll position was kept when opening a new file or revision (unless the caller specified a line, for instance when going from file to file in FileTree). 
When switching files this was confusing, then it is clearer if the new file starts from line 1.
For the same file the scrollposition is the same as the line number. This is a simplification of the code, will probably require some hundreds of micro seconds extra.

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -712,7 +719,8 @@ public int CurrentFileLine(bool isDiff)
                         : viewPosition.ActiveLineNum.RightLineNumber;
                 }
 
-                return viewPosition.CaretPosition.Line;
+                // Convert from offset to line number
+                return viewPosition.CaretPosition.Line + 1;
```
gerhardol, line 723: The reported file number was incorrect (so wrong line was opened in Blame, where this was used).

mstv (commented):
LGTM, will run the last update

mstv:
Sometimes, the switching from Diff to Blame works as expected - sometimes, it is confusing, e.g.:

Select line 179 | 148 of `GitVersion.cs` of 9a59d3d:

image

Hit `B`:

image

I think the `RightLineNumber` would have been the correct choice:

image

gerhardol:
> I think the `RightLineNumber` would have been the correct choice:

Yes! Even more complex CurrentFileLine!
It seem safe to switch to right file, will change.

gerhardol:
Adding a commit to scroll to the middle (instead of bottom) when going to a line.
@mstv mentioned this in another PR, it fits the best here I believe.

GitUI/Editor/FileViewerInternal.cs
```diff
@@ -713,10 +714,10 @@ public int CurrentFileLine(bool isDiff)
                 var viewPosition = _currentViewPosition;
                 if (isDiff && _viewer.GetLineText(0) == viewPosition.FirstLine && viewPosition.ActiveLineNum is not null)
                 {
-                    // prefer the LeftLineNum because the base revision will not change
-                    return viewPosition.ActiveLineNum.LeftLineNumber != DiffLineInfo.NotApplicableLineNum
-                        ? viewPosition.ActiveLineNum.LeftLineNumber
-                        : viewPosition.ActiveLineNum.RightLineNumber;
+                    // prefer the RightLineNum that is for the currentl revision
```
mstv, line 717:
```suggestion
                    // prefer the RightLineNum that is for the current revision
```

mstv (approved):
Works well, thank you.

gerhardol:
Squash merge in three days

PR #9806, opened 2022-01-07
AppTitle: File name in path filter first

gerhardol:
Part of #9796 

## Proposed changes

If a path filter is applied, display the filename path (if not possible use the filter path) first in the app title.
This is to separate path filters similar to how FileHistory is separated but also show the filtered path better.

I am not thrilled by the ⏱ but have no better suggestions myself.

### Before

FileHistory is not default in master but can be enabled

image

image

### After

image

image

## Test methodology 

Updated the app title tests.

## Merge strategy

----

### DIFF
--- a/GitCommands/AppTitleGenerator.cs
+++ b/GitCommands/AppTitleGenerator.cs
@@ -1,5 +1,6 @@
-using System;
+﻿using System;
 using System.ComponentModel.Composition;
+using System.IO;
 using GitCommands.UserRepositoryHistory;
 using GitUIPluginInterfaces;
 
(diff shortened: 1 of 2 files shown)


gerhardol:
From https://github.com/gitextensions/gitextensions/issues/9796#issuecomment-1007885099 by @RussKie 

>> Here are for me useful things:
1. I like the idea of the "File History" text we had before but (1) it is too long and hide the filename in the task bar preview (2) it could be wrong when you display a folder history => Proposition: Use an emoji instead
2. The file name is the more important information (and the path is too long) => Proposition: display filename and after directory path like "filename.cs [directory/path/to/file]"
After a quick test using an emoji, I end up with something like that (when applied on the File History form because it was easier for me) and that I think render quite well

> I _think_ I like the idea with the emoji, but I'm finding the chosen emoji confusing. For file history I'd expect something like these (though I don't know if these are available as emojis):
I also not finding "filename.cs [directory/path/to/file]" over "directory/path/to/file/filename.cs", as it adds an additional layer of mental indirection. Perhaps we should apply shortening strategies we apply in other places.

The status bar is not used for anything useful other than this information, we can as well add extra info...
The file emoji is maybe better than the time emoji.

I did not include the path at all (either full path or only filename). 
The title bar do not contain all info, just enough to separate instances.
Personally I would rather keep the filename in quotes and skip the emoji.

pmiossec:
> I think I like the idea with the emoji, but I'm finding the chosen emoji confusing. For file history I'd expect something like these (though I don't know if these are available as emojis):

For me whatever emoji that let the user think about 'time' or 'history' is good enough because after the first use, the user will have understood. And yes there are not a lot of choice...
And for me the 'file' emoji is not the best one because you could do the history of a folder. So the notion of 'history' is more important than the one of 'file' /'directory'.

I also like the choice to display only the filename so that we are able to see the most important information in the Taskbar. Before, displaying the full path was useless...

gerhardol:
I pushed using quotes only, that is enough to make this stand out for me and it shows one more character of the filename.

pmiossec:
> I pushed using quotes only, that is enough to make this stand out for me and it shows one more character of the filename.

It lacks an emoji 😉 but for me it is good enough.

gerhardol:
> > I pushed using quotes only, that is enough to make this stand out for me and it shows one more character of the filename.
> 
> It lacks an emoji 😉 but for me it is good enough.

@pmiossec 
Can you approve so I can merge this?

pmiossec (approved):

mstv (approved):
Looks sensible, have not run

PR #9807, opened 2022-01-07
AdvancedFilter: Disable filters button dropdown

gerhardol:
Part of #9796 

## Proposed changes

Change Advanced filter to a split button and add "Disable filters"
to the dropdown.

* To indicate that a filter is active, the "Checked" property can no longer be used.
This is indicated with a new icon instead.

### Before

image

image

### After

image

image

## Test methodology 

Manual

## Merge strategy

----


gerhardol:
Changing  the icon when a filter is active, I believe this is good enough to indicate that there is a filter

pmiossec:
What I was thinking is to have in the drop-down a way to disable the one related to the file path because it will be set every time you use the 'file history' feature.

And I find it boring to have to open the pop-up every times to disable it. And 'Disable all' could only be used if you don't have other filters that you want to keep (for example the number of commits to retrieve) 

After that I let you decide if, for homogeneity, we must add every filter in the drop-down. Maybe overkill...

gerhardol:
> After that I let you decide if, for homogeneity, we must add every filter in the drop-down. Maybe overkill...

Adding a special for disabling path filters only.
For the other filters only Branch filter can be activated outside of this dialog and the branch filter can be disabled already in the toolbar.

RussKie:
I propose the following:
image

Though perhaps a funnel with a cross may work better.

gerhardol:
> I propose the following:

Even longer toolbar, but a button with image and tooltip would be shorter.
OK for me, about the same as a tool strip button.

But @pmiossec wanted to have a special for path filters, another for all filters for which a tool strip button is preferred. If that is a requirement I prefer one button for all.

gerhardol:
> I don't want to clear these "permanent filters" when I need to quickly clear the path filter.

One or two buttons for you?
With two buttons, someone need to hand edit an icon...

> Side note: A "since" filter expressed in number of days instead of a date will be a nice improvement for my use case....

Side note: How big are the repos compared to Linux, how long time to open?
Do you notice much changes in latest master?
(There were some severe regressions for a while in master but the the time to open repos should be about half in master compared to 3.5.)

gerhardol:
Fugue has a funnel-minus icon too, if the two button approach is preferred
funnel--minus

pmiossec:
> One or two buttons for you?

Not sure but I tend to think only one. Because, with the workflow I explained with a "permanent filter" enabled it won't change anything, you will have to click twice. The difference is that "Clear all filters" will be in one click but I don't think I will use it often (except if I stop using the "permanent filter").
But if we choose to go for 2, I think the icon you found is good enough to not display text...

I prefer to test one button during some times to see if it's painful or not but that's not as if I will have to do this action as often as it become painful.

Other side note: It is for me much more painful for me to not have 2 dissociated buttons for the fetch and pull (like in all the others git GUIs) and having to click the little arrow if I want to change the action...

>Side note: How big are the repos compared to Linux, how long time to open?

Not that big. Around 25k commit. But enough to feel the difference when the filter is enabled especially when my laptop is busy doing something else.

>Do you notice much changes in latest master?

I didn't paid attention to that as I don't run v3.5 but my custom version based on the last commit before switch to .net5 and I didn't used enough the master version to compare....

gerhardol:
> Not sure but I tend to think only one.

Updated the original proposal but did not overwrite RussKies proposal but pushed to 
https://github.com/gerhardol/gitextensions/pull/new/feature/i9796-adv-filter-disable-go1

image

RussKie:
I think a strategic solution to the problem would be to use
https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstripcontrolhost,
and move the advanced filter dialog into it. This would probably allow us
to remove the quick path filter too, because the advanced filter will be as
readily available.

PS. I've restarted my efforts in making toolbars movable because even on a
4K screen with user scripts the toolstrip becomes unusably long.

gerhardol:
> I think a strategic solution to the problem would be to use https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstripcontrolhost, and move the advanced filter dialog into it.

I believe my last proposal is sufficient. Maybe the icon when a filter is active can be clearer.

pmiossec:
> > Not sure but I tend to think only one.
> 
> Updated the original proposal but did not overwrite RussKies proposal but pushed to https://github.com/gerhardol/gitextensions/pull/new/feature/i9796-adv-filter-disable-go1
> 
> image

What is in this branch seems good for me in terms of behavior....
Maybe one or two adjustments needed around "Disable filters": I don't know if it is on purpose but, it is not enabled when "Limit" is checked and when clicking on it it doesn't disable "Limit" also. I was expecting that it disable **all** the filters...

gerhardol:
> What is in this branch seems good for me in terms of behavior....

I will push this then

> Maybe one or two adjustments needed around "Disable filters": I don't know if it is on purpose but, it is not enabled when "Limit" is checked and when clicking on it it doesn't disable "Limit" also. I was expecting that it disable all the filters...

Limit, IgnoreCase, CurrentBranchOnlyCheck is ignored in disabling and displaying tooltip.
(ShowSimplifyByDecoration was added in #9808)
I rather skip that still...

RussKie:
I want to try ToolStripControlHost (I had a brush with it as part of work investigations). I think it could be an easy win.

gerhardol:
Completed renaming from Disable filters to Reset filters, that RussKie started with
Updated with toolstrip for ResetPathFilter in addition to all
Changed default hotkey to ctrl-i due to https://github.com/gitextensions/gitextensions/issues/9796#issuecomment-1007342576 @pmiossec

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -377,14 +377,14 @@ private static ArgumentString FindRenamesAndCopiesOpts()
...
+            _filterInfo.ResetAllFilters();
         }
 
-        public void DisableRevisionFilter()
+        public void ResetRevisionFilter()
         {
-            _filterInfo.DisableFilters();
+            _filterInfo.ResetAllFilters();
             PerformRefreshRevisions();
         }
```
mstv, line 389: ❓ The only difference is `PerformRefreshRevisions()`. Rename to `ResetAllFiltersAndRefresh`?

mstv (commented):
:+1 for the PR
but a few comments

GitUI/Hotkey/HotkeySettingsManager.cs
```diff
@@ -297,13 +297,14 @@ public static HotkeySettings[] CreateDefaultSettings()
...
                     Hk(RevisionGridControl.Command.CompareToWorkingDirectory, Keys.Control | Keys.D),
                     Hk(RevisionGridControl.Command.CreateFixupCommit, Keys.Control | Keys.X),
-                    Hk(RevisionGridControl.Command.DisableRevisionFilter, Keys.Control | Keys.Shift | Keys.F),
                     Hk(RevisionGridControl.Command.GoToChild, Keys.Control | Keys.N),
                     Hk(RevisionGridControl.Command.GoToCommit, Keys.Control | Keys.Shift | Keys.G),
                     Hk(RevisionGridControl.Command.GoToParent, Keys.Control | Keys.P),
                     Hk(RevisionGridControl.Command.NextQuickSearch, Keys.Alt | Keys.Down),
                     Hk(RevisionGridControl.Command.OpenCommitsWithDifftool, Keys.None),
                     Hk(RevisionGridControl.Command.PrevQuickSearch, Keys.Alt | Keys.Up),
+                    Hk(RevisionGridControl.Command.ResetPathRevisionFilter, Keys.Control | Keys.Alt | Keys.I),
```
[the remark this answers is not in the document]
gerhardol: I got some other action (screenshot?) for ctrl-shift-i so I selected something else. Weird that I do not see that today.
But it does not seem to be a common desktop shortcut, changing.

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -318,9 +325,16 @@ private static void ToolStripSplitButtonDropDownClosed(object sender, EventArgs
...
         {
-            RevisionGridFilter.ShowRevisionFilterDialog();
+            if (!tsmiResetAllFilters.Enabled)
+            {
+                RevisionGridFilter.ShowRevisionFilterDialog();
+            }
+            else
+            {
+                tsbtnAdvancedFilter.ShowDropDown();
+            }
```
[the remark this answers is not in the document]
gerhardol: I expect that if you have a filter, the most likely action is to deactivate it, not modify it.
So the dropdown is shown.
The arrow is a little narrow.

GitUI/UserControls/FilterToolBar.Designer.cs
```diff
@@ -78,11 +82,43 @@ private void InitializeComponent()
...
+            this.tsbtnAdvancedFilter.DropDownItems.AddRange(new System.Windows.Forms.ToolStripItem[] {
+            this.tsmiResetPathFilters,
+            this.tsmiResetAllFilters,
+            this.toolStripSeparatorAdvancedFilter,
+            this.tsmiAdvancedFilter});
             this.tsbtnAdvancedFilter.Image = global::GitUI.Properties.Images.FunnelPencil;
             this.tsbtnAdvancedFilter.Name = "tsbtnAdvancedFilter";
-            this.tsbtnAdvancedFilter.Size = new System.Drawing.Size(23, 22);
+            this.tsbtnAdvancedFilter.Size = new System.Drawing.Size(32, 22);
             this.tsbtnAdvancedFilter.ToolTipText = "Advanced filter";
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/FilterToolBar.Designer.cs
```diff
@@ -78,11 +82,43 @@ private void InitializeComponent()
...
-            this.tsbtnAdvancedFilter.Click += new System.EventHandler(this.tsbtnAdvancedFilter_Click);
+            this.tsbtnAdvancedFilter.ButtonClick += new System.EventHandler(this.tsbtnAdvancedFilter_ButtonClick);
+            this.tsbtnAdvancedFilter.DropDownOpening += new System.EventHandler(this.toolStripButtonLevelUp_DropDownOpening);
+            // 
+            // tsmiResetPathFilters
+            // 
+            this.tsmiResetPathFilters.Name = "tsmiResetPathFilters";
+            this.tsmiResetPathFilters.Size = new System.Drawing.Size(259, 22);
+            this.tsmiResetPathFilters.Text = "Reset path filter";
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: ok

GitUI/UserControls/FilterToolBar.Designer.cs
```diff
@@ -78,11 +82,43 @@ private void InitializeComponent()
...
+            this.tsmiResetAllFilters,
+            this.toolStripSeparatorAdvancedFilter,
+            this.tsmiAdvancedFilter});
             this.tsbtnAdvancedFilter.Image = global::GitUI.Properties.Images.FunnelPencil;
             this.tsbtnAdvancedFilter.Name = "tsbtnAdvancedFilter";
-            this.tsbtnAdvancedFilter.Size = new System.Drawing.Size(23, 22);
-            this.tsbtnAdvancedFilter.ToolTipText = "Advanced filter";
-            this.tsbtnAdvancedFilter.Click += new System.EventHandler(this.tsbtnAdvancedFilter_Click);
+            this.tsbtnAdvancedFilter.Size = new System.Drawing.Size(32, 22);
+            this.tsbtnAdvancedFilter.ToolTipText = "&Advanced filter";
```
mstv, line 93: That's just the tooltip.

gerhardol: fixed...

mstv (approved):

GitUI/Hotkey/HotkeySettingsManager.cs
```diff
@@ -297,13 +297,14 @@ public static HotkeySettings[] CreateDefaultSettings()
...
-                    Hk(RevisionGridControl.Command.DisableRevisionFilter, Keys.Control | Keys.Shift | Keys.F),
                     Hk(RevisionGridControl.Command.GoToChild, Keys.Control | Keys.N),
                     Hk(RevisionGridControl.Command.GoToCommit, Keys.Control | Keys.Shift | Keys.G),
                     Hk(RevisionGridControl.Command.GoToParent, Keys.Control | Keys.P),
                     Hk(RevisionGridControl.Command.NextQuickSearch, Keys.Alt | Keys.Down),
                     Hk(RevisionGridControl.Command.OpenCommitsWithDifftool, Keys.None),
                     Hk(RevisionGridControl.Command.PrevQuickSearch, Keys.Alt | Keys.Up),
+                    Hk(RevisionGridControl.Command.ResetPathRevisionFilter, Keys.Control | Keys.Shift | Keys.I),
+                    Hk(RevisionGridControl.Command.ResetRevisionFilter, Keys.Control | Keys.I),
                     Hk(RevisionGridControl.Command.RevisionFilter, Keys.Control | Keys.F),
```
gerhardol, line 308: With #9827 it is more important that shortcuts are unique in all forms.
The existing `ctrl-f` should probably be changed, which affect the new shortcuts too.
@mstv any better suggestions?
```suggestion
                    Hk(RevisionGridControl.Command.ResetPathRevisionFilter, Keys.Control | Keys.Shift | Keys.H),
                    Hk(RevisionGridControl.Command.ResetRevisionFilter, Keys.Control | Keys.I),
                    Hk(RevisionGridControl.Command.RevisionFilter, Keys.Control | Keys.I),
```

gerhardol:
> I think a strategic solution to the problem would be to use https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstripcontrolhost

@RussKie Any opinions about the PR, is it good-enough like this?

GitUI/Hotkey/HotkeySettingsManager.cs
mstv: Hotkeys _can_ have different behavior depending on the focused control. You are right, if we can avoid it, it will save some manual focus switches.

```suggestion
                    Hk(RevisionGridControl.Command.ResetPathRevisionFilter, Keys.Control | Keys.Shift | Keys.H),
                    Hk(RevisionGridControl.Command.ResetRevisionFilter, Keys.Control | Keys.Shift | Keys.I),
                    Hk(RevisionGridControl.Command.RevisionFilter, Keys.Control | Keys.I),
```
was meant. I like the pair `Ctrl+I` / `Ctrl+Shift+I`.

gerhardol: > Hotkeys _can_ have different behavior depending on the focused control. 
> You are right, if we can avoid it, it will save some manual focus switches.

There will be different actions for different controls. like Ctrl-A has a meaning in the FileViewer and FileStatusList, but we should avoid it so you do not have to focus the right control .

> was meant. I like the pair `Ctrl+I` / `Ctrl+Shift+I`.

Yes...

gerhardol:
Squash merge tomorrow

PR #9816, opened 2022-01-13
BlameControl: cancellation token for git-blame

gerhardol:
## Proposed changes

git-blame may require considerable time, the git executable
must be cancelable.
For instance git-blame of "GitUI/Translation/en_pseudo.xlf" takes about a minute for me.

Clear the blame viewer when processing a new file.
Before the contents of the previous file was not cleared, which was confusing.
(This was not a problem when a separate FileHistory had to be opened for Blame, but is now when Blame can be viewed in FileTree and RevDidd).

Note that cancelling the Git command will not kill the Git process. So clicking around in the file tree can start several long running processes.
Before, they always ran to completion in GitModule, now they are aborted (no exit status and short runtime in command log).
So in some sense cancelling is worse as the command still runs but command cache is not updated if the file is viewed again...
process.Kill() could be used when cancellation is done.

### Before

(You can see that the second time English-plugins.xlf is clicked that it is quickly displayed as git-blame ran to completion before the executable was cancelled).
blame-view-0

### After

blame-view-1

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -340,7 +340,7 @@ namespace GitUI.CommandsDialogs
 
             if (tabControl1.SelectedTab == BlameTab)
             {
-                Blame.LoadBlame(revision, children, fileName, RevisionGrid, BlameTab, Diff.Encoding, force: force);
+                _ = Blame.LoadBlameAsync(revision, children, fileName, RevisionGrid, BlameTab, Diff.Encoding, force: force, cancellationToken: _viewChangesSequence.Next());
             }
             else if (tabControl1.SelectedTab == ViewTab)
             {
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -476,14 +476,15 @@ namespace GitUI.CommandsDialogs
             RequestRefresh();
         }
 
-        private void BlameSelectedFileDiff(int? line = null)
+        private async Task ShowSelectedFileBlameAsync(int? line = null)
         {
             BlameControl.Visible = true;
             DiffText.Visible = false;
             GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
                 ? _revisionGrid.GetActualRevision(_revisionGrid.CurrentCheckout)
                 : DiffFiles.SelectedItem.SecondRevision;
-            BlameControl.LoadBlame(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid, controlToMask: null, DiffText.Encoding, line, cancellationToken: _viewChangesSequence.Next());
+            await BlameControl.LoadBlameAsync(rev, children: null, DiffFiles.SelectedItem.Item.Name, _revisionGrid,
+                controlToMask: null, DiffText.Encoding, line, cancellationToken: _viewChangesSequence.Next());
         }
 
         private async Task ShowSelectedFileDiffAsync()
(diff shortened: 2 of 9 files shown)


GitUI/CommandsDialogs/FormFileHistory.cs
```diff
@@ -340,7 +340,7 @@ private void UpdateSelectedFileViewers(bool force = false)
 
             if (tabControl1.SelectedTab == BlameTab)
             {
-                Blame.LoadBlame(revision, children, fileName, RevisionGrid, BlameTab, Diff.Encoding, force: force);
+                _ = Blame.LoadBlameAsync(revision, children, fileName, RevisionGrid, BlameTab, Diff.Encoding, force: force, cancellationToken: _viewChangesSequence.Next());
```
mstv, line 343: ❔ `_ =`
Why do you explicitly discard the `Task` result sometimes and sometimes not?

GitUI/UserControls/BlameControl.cs
```diff
@@ -118,7 +119,12 @@ public void LoadBlame(GitRevision revision, IReadOnlyList<ObjectId>? children, s
             _fileName = fileName;
             _encoding = encoding;
 
-            _blameLoader.LoadAsync(() => _blame = Module.Blame(fileName, objectId.ToString(), encoding),
+            // Clear the contents of the viewer while loading
+            BlameAuthor.ClearBlameGutter();
+            BlameAuthor.Clear();
+            BlameFile.Clear();
+
+            return _blameLoader.LoadAsync(cancellationToken => _blame = Module.Blame(fileName, objectId.ToString(), encoding, null, cancellationToken),
```
mstv, line 127:
```suggestion
            return _blameLoader.LoadAsync(cancellationToken => _blame = Module.Blame(fileName, objectId.ToString(), encoding, lines: null, cancellationToken),
```

mstv (commented):
+1 LGTM

---

> Kill() could be used when cancellation is done.

Yes, for later. I would rather give git a chance to clean up by sending a `Ctrl+C` signal.
This applies to `git log`, too, particularly if the RevisionGrid filters are active.

GitUI/CommandsDialogs/FormFileHistory.cs
gerhardol: This was required in #8914
Will add to FormBlame

GitUI/UserControls/BlameControl.cs
gerhardol: ok

gerhardol: `Module.Blame(fileName, objectId.ToString(), encoding, cancellationToken: cancellationToken)`

GitUI/CommandsDialogs/FormFileHistory.cs
RussKie: Why not `JTF.RunAsync(() => Blame.LoadBlameAsync(...)).FileAndForget();` or `JTF.Run(() => Blame.LoadBlameAsync(...))`?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -638,7 +638,7 @@ private void blameToolStripMenuItem_Click(object sender, EventArgs e)
             }
 
             blameToolStripMenuItem.Checked = !blameToolStripMenuItem.Checked;
-            BlameSelectedFileDiff();
+            ShowSelectedFileBlameAsync();
```
RussKie, line 641: Ditto - why not use JTF?

RussKie:
LGTM

GitUI/CommandsDialogs/FormFileHistory.cs
gerhardol: I just realized that GitUIExtensions.ViewChangesAsync is `async Task` changing this too (it should work the same).
The same handling should be used, should ViewChangesAsync be changed too?

gerhardol:
squashed and rebased, added one change to clear the file viewer async.

Would like to merge this in two days or so.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -638,7 +639,10 @@ private void blameToolStripMenuItem_Click(object sender, EventArgs e)
             }
 
             blameToolStripMenuItem.Checked = !blameToolStripMenuItem.Checked;
-            BlameSelectedFileDiff();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await ShowSelectedFileBlameAsync();
+            }).FileAndForget();
```
mstv, line 645: ❔
```suggestion
            ThreadHelper.JoinableTaskFactory.RunAsync(ShowSelectedFileBlameAsync).FileAndForget();
```

gerhardol: Yes, but this will be changed with #9804 again.

mstv (approved):

gerhardol:
Squash merging tomorrow.

PR #9827, opened 2022-01-19
Route hotkeys to all visible controls

mstv:
## Proposed changes

- Route hotkeys to the other visible controls in FormBrowse and FormCommit if not handled by the focused control 

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 9916393dc46aaeb1c5a5006d046768b301bd5721
- Git 2.31.1.windows.1 (recommended: 2.33.0 or later)
- Microsoft Windows NT 10.0.19044.0
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9828, opened 2022-01-19
RevisionDiffControl: Add hotkey for selecting all changes (in the first group)

### LINKED ISSUE #9823: RevisionGrid: Avoid temporary hang on `Ctrl+A`
### Feature description

I often hit `Ctrl+A` in order to select all files affected by a commit.
If the RevisionGrid is focused, GE will be very busy for a long time (and the commit is not selected anymore).

Variant A: Ignore `Ctrl+A` in the RevisionGrid.

Variant B: A TaskDialog could be displayed asking for confirmation to select all revisions
and offer to select all files affected by the commit in the FileStatusList (focusing the Diff tab).

Variant C: Add a GE hotkey for the second part of variant B.
With an upcoming PR which routes hotkeys to all visible controls, this hotkey will then take precedence over the built-in handling of the DataGridView.

Variant D: ?

### Environment

GE HEAD

mstv:
Fixes #9823
in combination with the base PR #9827
and main issue of #9861

## Proposed changes

- `IsTextEditKey`: Add missing hotkeys combined with the `Control` modifier key
- Add hotkey `RevisionDiffControl.Command.SelectFirstGroupChanges`, which selects all changes in the first group and focuses the FileStatusList
- Hardcode Ctrl+A handling for RevisionGridControl:
  - Forward this key to the parent form as a `Hotkey`
  - Suppress selecting all revisions

### Before

Hit `Ctrl+A` in RevisionGrid => all revisions selected

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build fb0100d750fd791622389f4e2452f8eb985f2c75
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

I do **not** agree that the maintainer **squash merge** this PR.

----


gerhardol (commented):
+1
LGTM, have not run
This is really WIP right now as base PR is not merged

gerhardol (approved):

GitUI/Hotkey/HotkeySettingsManager.cs
```diff
@@ -306,8 +306,9 @@ public static HotkeySettings[] CreateDefaultSettings()
                     Hk(RevisionGridControl.Command.ResetRevisionPathFilter, Keys.Control | Keys.Shift | Keys.H),
                     Hk(RevisionGridControl.Command.ResetRevisionFilter, Keys.Control | Keys.Shift | Keys.I),
                     Hk(RevisionGridControl.Command.RevisionFilter, Keys.Control | Keys.I),
-                    Hk(RevisionGridControl.Command.SelectCurrentRevision, Keys.Control | Keys.Shift | Keys.C),
                     Hk(RevisionGridControl.Command.SelectAsBaseToCompare, Keys.Control | Keys.L),
+                    Hk(RevisionGridControl.Command.SelectCurrentRevision, Keys.Control | Keys.Shift | Keys.C),
+                    Hk(RevisionGridControl.Command.SelectFirstDiffGroupChanges, Keys.Control | Keys.A),
```
gerhardol, line 311: Why is this needed as a new hotkey?

mstv: > Why is this needed as a new hotkey?

Because of the the case where the RevisionGrid is focused:
`GitExtensionsControl.ProcessCmdKey` does not find a Hotkey for the RevisionGrid and calls the base method `UserControl.ProcessCmdKey`.
This gives `DataGridView` the chance to handle the key (and to select all revisions) before `FormBrowse.ProcessCmdKey` could route it to `revisionDiff`.
But `FormBrowse.ProcessCmdKey` would not route it anyway, because `IsTextEditKey` returns `true` for `Ctrl+A` since this PR.
(Adding explicit Hotkeys for `Ctrl+A`, `Ctrl+C`, `Ctrl+V` to all controls is not an option IMO.)

I have updated the PR description text.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -150,6 +150,11 @@ public sealed partial class RevisionGridControl : GitModuleControl, IScriptHostC
         /// </summary>
         internal Dictionary<ObjectId, string>? FilePathByObjectId { get; set; } = null;
 
+        /// <summary>
+        /// Executed on hotkey Ctrl+A, which is named SelectFirstDiffGroupChanges.
+        /// </summary>
+        internal Action SelectAll { get; set; }
```
mstv, line 156: Any objections?

gerhardol: (Sorry, my brain shortcircuited when trying to review this last time, will try the next days...)
I would prefer to hardcode Ctrl-A in RevGrid rather than adding a hotkey for another form, but has not realized if that can be done.

gerhardol (commented):
Ctrl-A in FileViewer (like diff) also not works. Checking for DiffText.Focused in RevDiff does not seem to help (similar problem as https://github.com/gitextensions/gitextensions/pull/9804#issuecomment-1038450150).
In the RevDiff filter box, ctrl-a, ctrl-x is executing the hotkey too. (That works in 3.5.4.)

I am leaning to just ignoring Ctrl-A in RevGrid if a variant of IsTextEditKey (with only ctrl-A) is not feasible.
But if you find this good enough, I will approve so we can continue with more important issues.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
mstv: > I would prefer to hardcode `Ctrl+A` in RevGrid rather than adding a hotkey for another form, but have not realized if that can be done.

I found different a little less spaghetti variant. FormBrowse still needs to know about `Ctrl+A`.
But the new `GitExtensionsFormBase.ProcessHotkey` is nearly the same as the `GitExtensionsControl.ProcessHotkey` factored out in #9827 - good.

ResourceManager/GitExtensionsControl.cs
```diff
@@ -138,9 +138,18 @@ public static bool IsTextEditKey(Keys keys, bool multiLine = false)
...
                     or (>= Keys.Oem1 and <= Keys.Oem102):
                 case Keys.Space:
+                case Keys.Insert:
+                    return true;
+            }
+
+            keys &= ~Keys.Control; // ignore the CONTROL key as modifier
+            switch (keys)
+            {
+                case Keys.A:
```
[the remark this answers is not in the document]
mstv: > Add Ctrl-Z/Ctrl-Y too?

Yes, as well as `Ctrl+X`.

gerhardol (approved):
I believe I understand the changes in this PR and cannot see any issues in use.

mstv:
Rebased and partially squashed without changes, updated the description.
I want to merge these three commits in two days.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2159,12 +2159,13 @@ private void QuickFetch()
...
             return !IsDesignMode && HotkeysEnabled
-                && (base.ProcessCmdKey(ref msg, keyData) // generic handling of this form's hotkeys (upstream)
-                    || (!GitExtensionsControl.IsTextEditKey(keyData) // downstream (without keys for quick search)
-                        && (RevisionGridControl.ProcessHotkey(keyData)
+                && (base.ProcessHotkey(keyData) // generic handling of this form's hotkeys (upstream)
+                    || ((!GitExtensionsControl.IsTextEditKey(keyData) // downstream (without keys for quick search)
+                         || (keyData == (Keys.Control | Keys.A) && RevisionGridControl.ContainsFocus)) // route Ctrl+A away from RevisionGridControl
+                        && ((keyData != (Keys.Control | Keys.A) && RevisionGridControl.ProcessHotkey(keyData))
                             || (CommitInfoTabControl.SelectedTab == DiffTabPage && revisionDiff.ProcessHotkey(keyData))
                             || (CommitInfoTabControl.SelectedTab == TreeTabPage && fileTree.ProcessHotkey(keyData)))));
```
RussKie, line 2170: Can we please break it in a series of easy to understand/maintain conditions?

ResourceManager/GitExtensionsFormBase.cs
```diff
@@ -41,6 +41,21 @@ public GitExtensionsFormBase()
 #endif
         }
 
+        /// <summary>
+        /// Checks if the form wants to handle the key and executes that hotkey
+        /// (without propagating an unhandled key to the base class function as in <cref>ProcessCmdKey</cref>).
+        /// </summary>
+        public virtual bool ProcessHotkey(Keys keyData)
```
RussKie, line 48: Similar to `ProcessCmdKey`:
```suggestion
        public virtual bool ProcessHotKey(Keys keyData)
```

RussKie (commented):
👍

GitUI/CommandsDialogs/FormBrowse.cs
mstv: yes, done

ResourceManager/GitExtensionsFormBase.cs
mstv: > Similar to `ProcessCmdKey`:

This comment was too short for me. What is your intention, @RussKie?

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2161,13 +2161,29 @@ private void QuickFetch()
...
+
+            // downstream (without keys for quick search)
+            // but allow routing Ctrl+A away from RevisionGridControl in order to not select all revisions
+            if (GitExtensionsControl.IsTextEditKey(keyData)
+                && !(keyData == (Keys.Control | Keys.A) && RevisionGridControl.ContainsFocus))
+            {
+                return false;
+            }
+
+            // route to visible controls which have their own hotkeys
```
gerhardol, line 2183: This restructure makes this easier to read.
I also want to add an explanation.
(CommitInfo should have editable text for artificial commands, this could be changed then. You have added Side Panel hotkeys...)

```suggestion
            // route to visible controls which have their own hotkeys
            // For CommitInfo (as tab or panel), the text hotkeys are ignored
```

mstv: I have added a comment regarding CommitInfo before `IsTextEditKey`.

> CommitInfo should have editable text for artificial commands, this could be changed then.

Yes, we must keep this in mind. When we add Hotkeys for CommitInfo, we must add it here, too.
Though I think "controls which have their own hotkeys" is clear enough.

> You have added Side Panel hotkeys...

... for a future follow-up PR. (The prepared branch does not work yet due to (shortcut use of) `PreviewKeyDown` in the ROT.)
Yes, I will add the routing to the ROT for completeness, although the existing hotkeys (rename, delete, search) have their own meaning for the RevisionGrid, which takes precedence.

ResourceManager/GitExtensionsFormBase.cs
RussKie: We should rename these too :)

mstv: > We should rename these too :)

1. This would break the settings file.
2. (We are neither .NET nor WinForms.)

RussKie: Touche ;)

PR #9829, opened 2022-01-21
LostObjects: better `log` command and regex

pmiossec:
to handle commit subjects (i.e first line) that contains ', '

Instead, use of a non printing ASCII character ('Unit Separator' 1F --made for this purpose--)  as delimiter
to highly decrease the probability to find the delimiter chosen in a commit message.

Example that will fail the regex: WIP on null_fix_github_plugin: f77395162 Try fix #Issue1, #Issue2

### Before

image

### After

image

## Test methodology 

- Manual (Create a dangling commit with the first line of the commit message contains 1 or more ', '

## Test environment(s) 

- Git Extensions 33.33.33
- Build fbc60746e3e67fe7057f390ad7d7d38ba0704d1e
- Git 2.34.0.windows.1
- .NET 5.0.13
- DPI 192dpi (200% scaling)

## Merge strategy

### DIFF
--- a/GitUI/CommandsDialogs/FormVerify.LostObject.cs
+++ b/GitUI/CommandsDialogs/FormVerify.LostObject.cs
@@ -30,7 +30,7 @@ namespace GitUI.CommandsDialogs
             private static readonly string LogCommandArgumentsFormat = (ArgumentString)new GitArgumentBuilder("log")
             {
                 "-n1",
-                "--pretty=format:\"%aN, %s, %ct, %P\" {0}"
+                "--pretty=format:\"%aN\u001F%s\u001F%ct\u001F%P\" {0}"
             };
 
             private static readonly string TagCommandArgumentsFormat = (ArgumentString)new GitArgumentBuilder("cat-file")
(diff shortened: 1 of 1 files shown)


RussKie (approved):

GitUI/CommandsDialogs/FormVerify.LostObject.cs
```diff
@@ -30,7 +30,7 @@ private sealed class LostObject
             private static readonly string LogCommandArgumentsFormat = (ArgumentString)new GitArgumentBuilder("log")
             {
                 "-n1",
-                "--pretty=format:\"%aN, %s, %ct, %P\" {0}"
+                "--pretty=format:\"%aN\u001F%s\u001F%ct\u001F%P\" {0}"
```
gerhardol, line 33: May this sequence appear in the encoded text from Git?
(There were some discussions about that for the RevisionReader changes).
Regardless, this should work much better than  `, `.

gerhardol (commented):
+1

GitUI/CommandsDialogs/FormVerify.LostObject.cs
pmiossec: > May this sequence appear in the encoded text from Git?

Sorry, I don't understand. Can you elaborate more?

I try a probable answer...

When 'Unit Separator' 1F is displayed, it is invisible so the fields seems not be separated even if they are.
I admit it is not the easiest thing to debug even if it works well and this character has been created for that... ( "ASCII includes several control characters that are intended to be used as delimiters. They are: 28 for File Separator, 29 for Group Separator, 30 for Record Separator, and 31 for Unit Separator.") 

image

Utf code used in VS:
image

mstv: The regex for the `subject` group is greedy including possible `\u001F`. Only the `author` group might be truncated by mistake. Then the right part would be added to the `subject`. So, I am pretty confident.

mstv (approved):

GitUI/CommandsDialogs/FormVerify.LostObject.cs
gerhardol: > > May this sequence appear in the encoded text from Git?
> 
> Sorry, I don't understand. Can you elaborate more?
> When 'Unit Separator' 1F is displayed, it is invisible so the fields seems not be separated even if they are. I admit it is not the easiest thing to debug even if it works well and this character has been created for that... ( "ASCII includes several control characters that are intended to be used as delimiters. They are: 28 for File Separator, 29 for Group Separator, 30 for Record Separator, and 31 for Unit Separator.")

A string is not stored in utf8 in C#, so the inserted character could possibly appear in the encoded string and wreck the parsing. If the separator character is ensured to not appear in strings, this will be safe (and it is less of a problem than the current implementation regardless).

PR #9831, opened 2022-01-22
RevDiff: Setting to show Blame in FileTree

gerhardol:
Part of #9796 

This addresses what is maybe the core problem in #9796: Blame in RevDiff
This was added in #9424 and was intended for quick checks. The use
corresponding to FileHistory was intended to be used in FileTree.

(Some other changes planned for Blame in general and specifically for Blame in RevDiff for other concerns.)

## Proposed changes

Blame in RevDiff has certain limitations as revisions for a file may
not exist in the grid and the file to blame may not exist in all
revisions.

Add a setting to explicitly enable blame in RevGrid.

### Before

Blame 'B' in RevDiff by default opened in RevDiff itself

image

### After

Default is only FileTree

image

Blame 'B' in RevDiff by default opens FileTree
Basically the same as selecting filetree in the context menu 'T' but Blame is opened if show file contents was selected ('B' in FileTree).

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -476,10 +476,8 @@ namespace GitUI.CommandsDialogs
             RequestRefresh();
         }
 
-        private void BlameSelectedFileDiff()
+        private void BlameSelectedFileDiff(int? line = null)
         {
-            int? line = DiffText.Visible ? DiffText.CurrentFileLine : null;
-
             BlameControl.Visible = true;
             DiffText.Visible = false;
             GitRevision rev = DiffFiles.SelectedItem.SecondRevision.IsArtificial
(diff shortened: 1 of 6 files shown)


GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -476,10 +476,8 @@ private void ResetSelectedItemsTo(bool actsAsChild)
             RequestRefresh();
         }
 
-        private void BlameSelectedFileDiff()
+        private void BlameSelectedFileDiff(int? line = null)
         {
-            int? line = DiffText.Visible ? DiffText.CurrentFileLine : null;
```
mstv, line 481: Sure? This argument is never used.

gerhardol: fixed the caller
Too many conflicting changes right now...

mstv: > Too many conflicting changes right now...

Be patient, please. You have thought through a lot and have integrated the history and blame functionality pretty smoothly IMO. I was afraid it would become more intrusive / complex.
I don't think it will be too difficult to support the other users' expectations step by step.

> fixed the caller

partially: the call yet

gerhardol: hopefully done...
I have some difficult merges to do...

mstv (approved):

RussKie (commented):
:shipit:

mstv:
For the artificial commits (where FileTree is not available), the Blame action should not just be disabled. Then I'd rather show the Blame in the Diff tab.

gerhardol:
> For the artificial commits (where FileTree is not available), the Blame action should not just be disabled. Then I'd rather show the Blame in the Diff tab.

That would be inconsistent behavior, slightly confusing.

I would have liked to switch to HEAD and showed it there (that is what Blame in artificial does anyway), but that was not so easy.

PR #9832, opened 2022-01-22
Blame: Start on line also when control already exists

gerhardol:
This is the core change in #9805 
To add tests to this, a lot more has to be changed and this change is useful to functional testing other upcoming GoToLine changes,
so it was broken out to a separate PR.

#9805 is kept to handle the tests.

## Proposed changes

Blame can start presenting on a line. This is is available in
RevDiff and RevFileTree since #9424.
(Even if the code enabling this has existed for long time,
it was just used for a special case.)

The "goto line" was only used if the control had to be init,
not if e.g. toggling to Diff and selecting an new diff line
and then Blame again.

## Screenshots 

See #9805 

## Test methodology 

Manual - tests in #9805

## Merge strategy

----


mstv (approved):
Works for switching Diff to Blame.
Though `FileViewerInternal.GoToLine` should center the `Caret.Position` if it is not entirely visible - separate issue.

PR #9833, opened 2022-01-23
RevGrid: Show popup when a revision is not in the revision grid

gerhardol:
Discussed in relation to #9796 

## Proposed changes

If a objectId is expected to be available in the repo (for instance
retrieved with git-rev-parse), but is not included in the grid,
GE will inform the user with a popup instead of just silently ignore.

This is how GE has behaved for a long time, but there were previously less opportunities to 
enter commit sha that were in the repo but not displayed in the grid. 
There are various filters (including path filter) that hides commits and
links are exposed in the side panel and commit info.

All occurrences of `SetSelectedRevision()` are not getting the popup, it is opt-in
even if most occurrences are included. I have tried to verify that there are no popups in 
"automatic" situations and that the ObjectIds are valid Git sha (normally parsed by git-rev-parse or from git-blame etc).

### Before

Navigate - go to commit with input that is not commitish.

image

### After

image

image

## Test methodology 

Manual review and check of situations the pops should occur in

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -2490,20 +2490,16 @@ namespace GitUI.CommandsDialogs
             {
                 case "gotocommit":
                     Validates.NotNull(e.Data);
-                    var found = Module.TryResolvePartialCommitId(e.Data, out var revision);
-
-                    if (found)
+                    if (!Module.TryResolvePartialCommitId(e.Data, out ObjectId commitId) || !RevisionGrid.SetSelectedRevision(commitId))
                     {
-                        found = RevisionGrid.SetSelectedRevision(revision);
-                    }
+                        if (commitId is null)
+                        {
+                            return;
+                        }
 
-                    // When 'git log --first-parent' filtration is used, user can click on child commit
-                    // that is not present in the shown git log. User still wants to see the child commit
-                    // and to make it possible we add explicit branch filter and refresh.
-                    if (AppSettings.ShowFirstParent && !found)
-                    {
-                        ToolStripFilters.SetBranchFilter(revision?.ToString());
-                        RevisionGrid.SetSelectedRevision(revision);
+                        // This may occur at various filters, like AppSettings.ShowFirstParent
+                        // will hide other than the first parent.
+                        MessageBoxes.RevisionFilteredInGrid(this, commitId);
                     }
 
                     break;
@@ -2511,9 +2507,14 @@ namespace GitUI.CommandsDialogs
                 case "gototag":
                     Validates.NotNull(e.Data);
                     CommitData? commit = _commitDataManager.GetCommitData(e.Data, out _);
-                    if (commit is not null)
+                    if (commit is null)
                     {
-                        RevisionGrid.SetSelectedRevision(commit.ObjectId);
+                        break;
+                    }
+
+                    if (!RevisionGrid.SetSelectedRevision(commit.ObjectId))
+                    {
+                        MessageBoxes.RevisionFilteredInGrid(this, commit.ObjectId);
                     }
 
                     break;
--- a/GitUI/CommandsDialogs/FormFileHistory.cs
+++ b/GitUI/CommandsDialogs/FormFileHistory.cs
@@ -553,8 +553,11 @@ namespace GitUI.CommandsDialogs
                 Validates.NotNull(e.Data);
                 if (Module.TryResolvePartialCommitId(e.Data, out var objectId))
                 {
-                    RevisionGrid.SetSelectedRevision(objectId);
-                }
+                    if (!RevisionGrid.SetSelectedRevision(objectId))
+                    {
+                        MessageBoxes.RevisionFilteredInGrid(this, objectId);
+                    }
+               }
             }
             else if (e.Command == "gotobranch" || e.Command == "gototag")
             {
@@ -562,7 +565,10 @@ namespace GitUI.CommandsDialogs
                 CommitData? commit = _commitDataManager.GetCommitData(e.Data, out _);
                 if (commit is not null)
                 {
-                    RevisionGrid.SetSelectedRevision(commit.ObjectId);
+                    if (!RevisionGrid.SetSelectedRevision(commit.ObjectId))
+                    {
+                        MessageBoxes.RevisionFilteredInGrid(this, commit.ObjectId);
+                    }
                 }
             }
             else if (e.Command == "navigatebackward")
--- a/GitUI/MessageBoxes.cs
+++ b/GitUI/MessageBoxes.cs
@@ -2,15 +2,18 @@
 using System.Windows.Forms;
 using GitCommands;
 using GitCommands.Config;
+using GitUIPluginInterfaces;
 using ResourceManager;
 
 namespace GitUI
 {
     public class MessageBoxes : Translate
     {
-        private readonly TranslationString _archiveRevisionCaption = new("Archive revision");
+        private readonly TranslationString _cannotFindRevisionFilter = new(@"Revision ""{0}"" is not visible in the revision grid. Remove the revision filter.");
+        private readonly TranslationString _cannotFindRevisionCaption = new("Cannot find revision");
+        private readonly TranslationString _noRevisionFoundError = new("No revision found.");
 
-        private readonly TranslationString _failedToExecuteScript = new("Failed to execute script");
+        private readonly TranslationString _archiveRevisionCaption = new("Archive revision");
 
         private readonly TranslationString _failedToRunShell = new("Failed to run shell");
 
@@ -58,6 +61,12 @@ namespace GitUI
         private static MessageBoxes? instance;
 
         private static MessageBoxes Instance => instance ??= new();
+
+        public static void RevisionFilteredInGrid(IWin32Window? owner, ObjectId objectId)
+            => ShowError(owner, string.Format(Instance._cannotFindRevisionFilter.Text, objectId.ToShortString()), Instance._cannotFindRevisionCaption.Text);
+
+        public static void CannotFindGitRevision(IWin32Window? owner)
+            => ShowError(owner, Instance._noRevisionFoundError.Text, Instance._cannotFindRevisionCaption.Text);
 
         public static void FailedToRunShell(IWin32Window? owner, string shell, Exception ex)
             => ShowError(owner, $"{Instance._failedToRunShell.Text} {shell.Quote()}.{Environment.NewLine}"
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -499,15 +499,18 @@ namespace GitUI.Blame
                 return;
             }
 
-            if (_revGrid is not null)
+            ObjectId selectedId = _lastBlameLine.Commit.ObjectId;
+            if (_revGrid is null)
             {
-                _clickedBlameLine = _lastBlameLine;
-                _revGrid.SetSelectedRevision(_lastBlameLine.Commit.ObjectId);
-            }
-            else
-            {
-                using FormCommitDiff frm = new(UICommands, _lastBlameLine.Commit.ObjectId);
+                using FormCommitDiff frm = new(UICommands, selectedId);
                 frm.ShowDialog(this);
+                return;
+            }
+
+            _clickedBlameLine = _lastBlameLine;
+            if (!_revGrid.SetSelectedRevision(selectedId))
+            {
+                MessageBoxes.RevisionFilteredInGrid(this, selectedId);
             }
         }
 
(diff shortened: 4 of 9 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -2454,7 +2454,7 @@ private void RevisionInfo_CommandClicked(object sender, ResourceManager.CommandE
 
                     if (found)
                     {
-                        found = RevisionGrid.SetSelectedRevision(revision);
+                        found = RevisionGrid.SetSelectedRevision(revision, showNoRevisionMsg: !AppSettings.ShowFirstParent);
```
gerhardol, line 2457: `RevisionInfo_CommandClicked()` is called when links in the Commit tab is clicked.

GitUI/CommandsDialogs/FormFileHistory.cs
```diff
@@ -553,7 +553,7 @@ private void Blame_CommandClick(object sender, ResourceManager.CommandEventArgs
                 Validates.NotNull(e.Data);
                 if (Module.TryResolvePartialCommitId(e.Data, out var objectId))
                 {
-                    RevisionGrid.SetSelectedRevision(objectId);
+                    RevisionGrid.SetSelectedRevision(objectId, showNoRevisionMsg: true);
```
gerhardol, line 556: Similar to FormBrowse

GitUI/UserControls/BlameControl.cs
```diff
@@ -502,7 +502,7 @@ private void ActiveTextAreaControlDoubleClick(object sender, EventArgs e)
             if (_revGrid is not null)
             {
                 _clickedBlameLine = _lastBlameLine;
-                _revGrid.SetSelectedRevision(_lastBlameLine.Commit.ObjectId);
+                _revGrid.SetSelectedRevision(_lastBlameLine.Commit.ObjectId, showNoRevisionMsg: true);
```
gerhardol, line 505: If a blame is not visible in the grid, the "Blame previous revision" and "Blame this revision" are disabled, so the popup should be redundant.

GitUI/UserControls/RevisionGrid/RevisionGridMenuCommands.cs
```diff
@@ -501,11 +500,11 @@ public void GotoCommitExecute()
 
             if (objectId is not null)
             {
-                _revisionGrid.SetSelectedRevision(objectId);
+                _revisionGrid.SetSelectedRevision(objectId, showNoRevisionMsg: true);
```
gerhardol, line 503: `Goto commit` form

GitUI/CommandsDialogs/FormBrowse.cs
RussKie, line 2457: I think there may be other reasons why a given revision isn't visible, i.e. other filters are applied.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -671,6 +673,11 @@ public bool SetSelectedRevision(ObjectId? objectId, bool toggleSelection = false
                 return true;
             }
 
+            if (showNoRevisionMsg)
```
RussKie, line 676: Why would we not want to show the message?

gerhardol:
Pushed one update. If the revision was not found the selection was still cleared after selecting.
So if the revision grid is filtered and you for instance clicked a branch in the sidepanel that is not visible the selection was just cleared.
Even more confusing when doubleclicking a commit that is not available in Blame (drop down menu is disabled, double click is still possible).

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -658,8 +658,10 @@ private int FindRevisionIndex(ObjectId? objectId)
         /// Returns whether the required revision was found and selected.
         /// </summary>
         /// <param name="objectId">Id of the revision to select.</param>
+        /// <param name="toggleSelection">Toggle if the selected state for the revision.</param>
+        /// <param name="showNoRevisionMsg">Show popup if revision is not visible in grid. Should only be set if the ObjectId is in the repo.</param>
         /// <returns><c>true</c> if the required revision was found and selected, otherwise <c>false</c>.</returns>
-        public bool SetSelectedRevision(ObjectId? objectId, bool toggleSelection = false)
+        public bool SetSelectedRevision(ObjectId? objectId, bool toggleSelection = false, bool showNoRevisionMsg = false)
```
mstv, line 664:
```suggestion
        public bool SetSelectedRevision(ObjectId? objectId, bool toggleSelection = false, bool showNoRevisionMsg = true)
```
(As you know,) *I* would rather show the message by default and justify the cases where it is suppressed.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: If the check is done in an "automatic" flow where you do not realize what went wrong.
Also if you for some reason manage to check a complete ObjectId that is not in the repo at all.

Both of these arguments (especially the second one) is possibly not valid scenarios, but I cannot verify all scenarios so this is taking it safe.

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: The intention is that the message should be interpreted for any filter not just "advanced filter", branchfilters etc but also show first parents etc.
Clarification?

GitUI/MessageBoxes.cs
```diff
@@ -54,6 +57,12 @@ internal MessageBoxes()
 
         private static MessageBoxes Instance => instance ??= new();
 
+        public static void RevisionFilteredInGrid(IWin32Window? owner, ObjectId objectId)
+            => ShowError(owner, string.Format(Instance._cannotFindRevisionFilter.Text, objectId.ToShortString()), Instance._cannotFindRevisionCaption.Text);
```
mstv, line 61: In a follow-up PR, we should switch `MessageBoxes` to `TaskDialog` and add an `Action` in order to turn off the filter directly.
Then this `Action` should retry to locate the revision.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: I choose the defensive choice: to not change the behavior unless you choose to.
The value is set true almost everywhere.
https://github.com/gitextensions/gitextensions/pull/9833#discussion_r792165847

But we can say that new usage of the code must investigate the behavior and change the defaults.

GitUI/MessageBoxes.cs
gerhardol: I do not think we can disable all possible filters. For instance all commits may not be loaded.
But it could be a suggestion, *Advanced filter* has most common occurrences (and could add more).

gerhardol:
rebased to resolve merge conflict

mstv (commented):
+1

gerhardol:
> (As you know,) I would rather show the message by default and justify the cases where it is suppressed.

> Why would we not want to show the message?

Changed the defaults to show the popup by defaults.
Also included most of those situations where popups where not set before. I reviewed those again and could not see any action that are automatic. Some of those I did not include earlier are from key board actions and the other is commented as follows:
```
            // Parent-child navigation can expect that SetSelectedRevision is always successful since it always uses first-parents
            _parentChildNavigationHistory = new ParentChildNavigationHistory(objectId => SetSelectedRevision(objectId));
```

` _parentChildNavigationHistory` is invoked at refreshes etc. Maybe this specific occurrence should suppress pops.

A few popup supersessions remains, mostly to avoid double popus.

RussKie:
Please correct me if I misunderstand the use case.
I understand it as follows:

A user is searching for a commit by hash.
The hash is not found.
The app shows a warning to the user (e.g., "The requested hash '<hash>'
cannot be found.") with an option to reset current filters, if any applied?
(I.e. FilterInfo)

In which case would we not want to show the warning?

gerhardol:
> Please correct me if I misunderstand the use case. I understand it as follows: A user is searching for a commit by hash. The hash is not found. The app shows a warning to the user (e.g., "The requested hash '<hash>' cannot be found.") with an option to reset current filters, if any applied? (I.e. FilterInfo) In which case would we not want to show the warning?

In this scenario, the pop is desired.

But the popup is suppressed for instance in https://github.com/gitextensions/gitextensions/pull/9833/files#diff-613b92e9f0a420f482a14b89d7f9c9d31fb4f0f98e5f343a130136492944f1e1L2457 to avoid two potential popups.

I am not 100% sure about some automated use, see `_parentChildNavigationHistory ` where the command is executed when switching repos. It seems OK.

An example where I do not want a popup (fortunately parent rewriting is used so this is not an issue):

If "goto parent `Ctrl-P`" was implemented with asking Git for the parent for the current selected revision, the sha would be valid but maybe not available in the revision grid. If the implementation first tried the Git direct parent, if that failed asked Git for the "rewritten parent" you would not like to have a popup for the first occurrence.

RussKie:
From the layering perspective the RevisionGridControl shouldn't show any
messages, it should only render the supplied content. If there are any
anomalies (e.g., absent hashes) it should throw an exception. Callers must
handle the exception as necessary.

gerhardol:
> From the layering perspective the RevisionGridControl shouldn't show any messages, it should only render the supplied content. If there are any anomalies (e.g., absent hashes) it should throw an exception. Callers must handle the exception as necessary.

It is not expected to be an error if `SetSelectedRevision` fails to find a revision, it returns false. The method is also used in non interactive situations and "two-step" checks. 
Popups cannot be shown in  every situation. 
This PR is not changing the internal handling, just shows the popup in all but a few situations.

RussKie:
I'm currently away and don't have access to a computer at the moment. I'd
like to have an opportunity to review this proposal in more details.

gerhardol:
> I'm currently away and don't have access to a computer at the moment. I'd like to have an opportunity to review this proposal in more details.

@RussKie comments?

GitUI/CommandsDialogs/FormBrowse.cs
RussKie: Sorry, the meaning of your message is still escaping me... 

We should avoid using `AppSettings.ShowFirstParent` and other filtering options on their own. All filtering decisions or checks should be done only via `FilterInfo` object (which is essentially an abstraction layer).

GitUI/MessageBoxes.cs
RussKie: `FilterInfo` contains all necessary information to know whether filters are applied or not. If there's something missing we need to add those checks there.
If there is no filter applied (as far as `FilterInfo` is concerned) we should show a task dialog with a message "Requested revision cannot be found". Else, if there is a filter applied, we should show a task dialog with an option to reset the filter - if that option is invoked - open the Advanced Filter dialog.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -658,22 +658,27 @@ private int FindRevisionIndex(ObjectId? objectId)
...
+            if (index < 0 || index >= _gridView.RowCount)
             {
-                Validates.NotNull(objectId);
-                SetSelectedIndex(index, toggleSelection);
-                _navigationHistory.Push(objectId);
-                return true;
+                if (showNoRevisionMsg)
+                {
+                    MessageBoxes.RevisionFilteredInGrid(this, objectId);
+                }
```
RussKie, line 673: I am very much against having the grid show messages, it doesn't feel appropriate.
This API returns `false` if we couldn't find the requested objectid. The caller (i.e., `FormBrowse`) must then show the error message, if necessary (in the interactive flow).

GitUI/MessageBoxes.cs
gerhardol: > `FilterInfo` contains all necessary information to know whether filters are applied or not. If there's something missing we need to add those checks there. If there is no filter applied (as far as `FilterInfo` is concerned) we should show a task dialog with a message "Requested revision cannot be found". Else, if there is a filter applied, we should show a task dialog with an option to reset the filter - if that option is invoked - open the Advanced Filter dialog.

 `FilterInfo` is just a subset, there are several others like "--first-parent, --reflog, --simplify-by-decoration" at a quick glance.
Advanced filter will be quite bigger...

Only a few of the uses is from FormBrowse, so the AdvancedFilter is not accissible, an action to reset the filter could be used though.

gerhardol:
Squashed, rebased (to resolve merge conflicts) and distributed the popup to all callers.
(Some cleanup too, showing the popup slightly more.)
Note that before #9864 it could be possible to get old commits in the history so popups could occur.

RussKie (approved):

gerhardol:
squash merge in one day

mstv (approved):
Have not run the latest

gerhardol:
Minor syntax touchups, removed handling for "gotocommit" that did not work (before this PR, exception occurred), the popup is sufficient.

PR #9841, opened 2022-01-28
Pull rebase: Fix detection of merge commits

pmiossec:
When the string is empty (i.e. no merge commits),
`LazySplit('\n').Any()` return `true`.

Replace by a check that the string contains something (i.e. a merge commit has been found)
preventing creation of potentially a lot of unnecessary substrings.

### Before

This popup is always displayed even if there is no merge commits to rebase:
image

### After

The popup is displayed only when a merge commit will be rebased.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build a61328ace4b35d936dbdf2e9127293452f835172
- Git 2.34.0.windows.1
- .NET 5.0.13
- DPI 192dpi (200% scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1177,7 +1177,8 @@ namespace GitCommands
             };
 
             // Could fail if pulling interactively from remote where the specified branch does not exist
-            return _gitExecutable.Execute(args, throwOnErrorExit: false).StandardOutput.LazySplit('\n').Any();
+            string mergeCommitsOutput = _gitExecutable.Execute(args, throwOnErrorExit: false).StandardOutput;
+            return !string.IsNullOrWhiteSpace(mergeCommitsOutput);
         }
 
         public ConfigFile GetSubmoduleConfigFile()


gerhardol:
I do not see this popup. Any other prerequisites?

pmiossec:
> I do not see this popup. Any other prerequisites?

Here are the conditions to call `MergeCommitExists()`:

https://github.com/gitextensions/gitextensions/blob/master/GitUI/CommandsDialogs/FormPull.cs#L652

So I would say "Rebase" and "Pull from remote":

image

GitCommands/Git/GitModule.cs
```diff
@@ -1177,7 +1177,7 @@ public bool ExistsMergeCommit(string? startRev, string? endRev)
             };
 
             // Could fail if pulling interactively from remote where the specified branch does not exist
-            return _gitExecutable.Execute(args, throwOnErrorExit: false).StandardOutput.LazySplit('\n').Any();
+            return !string.IsNullOrWhiteSpace(_gitExecutable.Execute(args, throwOnErrorExit: false).StandardOutput);
```
RussKie, line 1180:
```suggestion
            string output = _gitExecutable.Execute(args, throwOnErrorExit: false).StandardOutput;
            return !string.IsNullOrWhiteSpace(output);
```

mstv (commented):
LGTM
Of course but not obvious at first glance, `[Lazy]Split()` always returns (a part of) the input string - at least one empty string.

GitCommands/Git/GitModule.cs
pmiossec: fixed.

PR #9842, opened 2022-01-29
FileTree: Goto line in Blame from RevDiff

gerhardol:
Follow up to #9831

## Proposed changes

Before #9831 the default behavior in RevDiff was to open Blame in RevDiff,
now the file tree is opened instead but on the first line.
(Blame from RevDiff is basically the same as view in FileTree).
(there are some issues handled in #9805),

In RevDiff the current line in file diff was displayed initially.
When opening in FileTree the first line was displayed, not context.
This therefore restores the context handling.

Note: There is no follow from Blame to FileContents or Diff.
A separate PR may be submitted for that.
(using it locally for now)

## Screenshots 

Starting point:

image

Note the off-by-one fixed in #9805 

### Before

image

### After

Manually scrolled 

image

## Test methodology 

Manual (some related tests in #9805)

## Merge strategy

----


GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -404,7 +407,6 @@ private Task ShowGitItemAsync(GitItem gitItem)
                             return ViewGitItemAsync(gitItem);
                         }
 
-                        int? line = FileText.Visible ? FileText.CurrentFileLine : null;
```
mstv, line 407: ❓ If the following does not suffice
```
line ??= line = FileText.Visible ? FileText.CurrentFileLine : null;
```
`_requestedLine` must be updated if a line is selected in the FileText.

gerhardol: will update this some (mostly to handle repeated RevDiff->FileTree blame), this too
(pushed a tmp/ branch for now)

gerhardol: The line is now set directly when there is knowledge to display a file, so it is slightly simpler.

mstv (commented):
Works well for `B`. But for `T`, it *can* be different. I have not got a clue what toggles the behavior. Either it works, or line 1 is selected always.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -404,9 +418,8 @@ private Task ShowGitItemAsync(GitItem gitItem)
...
-                        BlameControl.Visible = true;
                         FileText.Visible = false;
+                        BlameControl.Visible = true;
```
mstv, line 422: This makes the FileStatusList's filter field get the focus when hitting `B` multiple times (with `Show blame in diff viewer` on).

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -134,18 +132,24 @@ public void ExpandToFile(string filePath, int? line = null, bool requestBlame =
 
             if (foundNode is not null)
             {
-                if (isIncompleteMatch)
+                if (isIncompleteMatch || foundNode.Tag is not GitItem gitItem)
                 {
                     MessageBox.Show(_nodeNotFoundNextAvailableParentSelected.Text, TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                    return;
```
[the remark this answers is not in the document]
gerhardol: > Should ClearOutputAsync() be called?

Yes, adding that

> How can I manually trigger this branch of execution?

Only in debugger what I can think of.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: That is different controls, how can that be related?

gerhardol:
> Works well for B. But for T, it can be different. I have not got a clue what toggles the behavior. Either it works, or line 1 is selected always.

Cannot reproduce now, will try again.
Should not be a blocker.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
mstv: Not directly, you are right. I hoped that BlameControl would be focused automatically but it isn't.

STR:
- focus FileText
- hit `B` twice

So, BlameControl should be focused explicitly if FileText was focused.

gerhardol: This behavior is not changed by this PR, it is a separate issue. (RevDiff is not changed here other than setting line if Blame is done in FileTree).
I have never the less not been able to find a workaround but will continue to look at it in #9804.

gerhardol:
> Works well for B. But for T, it can be different. I have not got a clue what toggles the behavior. Either it works, or line 1 is selected always.

I assume this is when Blame was not selected and the contentents is shown?
GotoLine() is not implemented in that scenario.

I have pushed `feature/i9796-blame-line-to-diff` about that, but have not decided if I want to create a PR for that, the complexity increases without too much benefit.

I hope this PR is done now so the PRs for #9796 can be completed and the complete experience can be tested.

GitUI/UserControls/BlameControl.cs
```diff
@@ -90,6 +90,8 @@ public void UpdateShowLineNumbers()
             BlameAuthor.ShowLineNumbers = AppSettings.BlameShowLineNumbers;
         }
 
+        public void GoToLine(int line) => BlameFile.GoToLine(line);
```
mstv, line 93: This is not used by this PR. Will it be used in a follow-up?

mstv (approved):

GitUI/UserControls/BlameControl.cs
gerhardol: It was in a previous version and may be in a future but as it is not used right now it will be removed.
As the viewing of files is done async, the line to go to should be provided when viewing the file. 
This is why `feature/i9796-blame-line-to-diff` is so complex.

gerhardol:
Plan to merge in two days.

gerhardol:
Reverting [70b0ce0](https://github.com/gitextensions/gitextensions/pull/9842/commits/70b0ce0a49148ad8c04690ab6dff6c2b2edf6d9c), that was a result of 
https://github.com/gitextensions/gitextensions/pull/9842#pullrequestreview-872433486
and some attempts to improve will be done in #9804

PR #9853, opened 2022-02-04
Offload refs and revision loading to background thread

RussKie:
## Proposed changes

- This is a tactical fix aim at increasing the responsiveness of the app during the refresh by moving Module.GetRefs call and revisionReader.Execute calls to a background thread.
This makes the app responsive even while loading huge repos like VS, which has >100K commits and takes more than 10s to parse revisions info, and has close to 180K tags, which also take more than 10s to parse.
NB: It also takes another 10+s to load that many tags into the left panel, and this issue is not being addressed in the context of this change.
A more holistic fix would likely involve rework of Module.GetRefs implementation.

- Also pass a cancellation token to allow cancel the load operation.

- Lastly show the spinner while revisions are being loaded, and clear both the left panel and the commit info panel at the beginning of the refresh.

## Screenshots 

ℹ️  I have disabled tags, because those have a significant effect and add about 25 extra seconds to the load.

### Before

- Git Extensions 4.0.0.13119
- Build b26006c674076a56b1c55a50918f4a65bfd5fb7a (Dirty)
- Git 2.33.0.windows.1
- .NET 5.0.13
- DPI 96dpi (no scaling)

https://user-images.githubusercontent.com/4403806/152541597-ec82ab45-e2c7-4bbb-a9ca-874ee4fabeb0.mp4

### After

https://user-images.githubusercontent.com/4403806/152540529-e11d17a8-3226-4286-a0a1-e36035b74192.mp4

You can still see moment of unresponsiveness but by far and large this is a significant improvement.
⚠️ And this was run with VS debugger attached. 

## Test methodology 

- manually

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -103,6 +103,7 @@ namespace GitUI
         private readonly ArtificialCommitChangeCount _workTreeChangeCount = new();
         private readonly ArtificialCommitChangeCount _indexChangeCount = new();
         private readonly CancellationTokenSequence _customDiffToolsSequence = new();
+        private readonly CancellationTokenSequence _refreshRevisionsSequence = new();
 
         /// <summary>
         /// The set of ref names that are ambiguous.
@@ -265,6 +266,7 @@ namespace GitUI
                 //// _selectionTimer handled by this.components
                 _buildServerWatcher?.Dispose();
                 _customDiffToolsSequence.Dispose();
+                _refreshRevisionsSequence.Dispose();
 
                 if (_indexWatcher.IsValueCreated)
                 {
@@ -291,6 +293,11 @@ namespace GitUI
             base.Dispose(disposing);
         }
 
+        /// <summary>
+        /// Reset the controls to the supplied content.
+        /// This is used to remove spinners added when loading and to replace the gridview at errors.
+        /// </summary>
+        /// <param name="content">The content to show.</param>
         private void SetPage(Control content)
         {
             for (int i = Controls.Count - 1; i >= 0; i--)
@@ -426,15 +433,7 @@ namespace GitUI
 
         private void ResetNavigationHistory()
         {
-            var selectedRevisions = GetSelectedRevisions();
-            if (selectedRevisions.Count > 0)
-            {
-                _navigationHistory.Push(selectedRevisions[0].ObjectId);
-            }
-            else
-            {
-                _navigationHistory.Clear();
-            }
+            _navigationHistory.Clear();
         }
 
         public void NavigateBackward()
@@ -839,7 +838,6 @@ namespace GitUI
         ///  and <see cref="ResumeRefreshRevisions"/>.
         /// </summary>
         public bool CanRefresh => !_isRefreshingRevisions && _updatingFilters == 0;
-        private readonly CancellationTokenSequence _cancellationTokenSequence = new();
 
         /// <summary>
         ///  Queries git for the new set of revisions and refreshes the grid.
@@ -854,22 +852,30 @@ namespace GitUI
                 return;
             }
 
-            _isRefreshingRevisions = true;
+            // Revision info is read in two parallel steps:
+            // 1. Read current commit, refs, prepare grid etc.
+            // 2. Read all revisions with git-log.
+            // Both these operations can take a long time and is done async.
+            // Step 2. requires that most of the information in 1. is prepared when updating the grid.
+            // Git will will provide this information when available (so slower to first info if sorting).
+            // This information from step 1. is therefore protected with Lazy accessosors and a semaphore.
             System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(1, 1);
             semaphoreCurrentCommit.Wait();
+            _isRefreshingRevisions = true;
 
-            bool firstRevisionReceived = false;
-            bool headIsHandled = false;
+            GitModule capturedModule = Module;
 
             // Find all ambiguous refs (including stash, notes etc)
             // Note: GetRefs can be very slow operation for large repos, start in parallel to git-log.
             // refsByObjectId is needed when first output from git-log is handled
             Func<IReadOnlyList<IGitRef>> getUnfilteredRefs = () => (getRefs ?? Module.GetRefs)(RefsFilter.NoFilter);
-            _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(getUnfilteredRefs()));
             Lazy<ILookup<ObjectId, IGitRef>> refsByObjectId = new(() => getUnfilteredRefs().ToLookup(head => head.ObjectId), System.Threading.LazyThreadSafetyMode.ExecutionAndPublication);
+            _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(getUnfilteredRefs()));
 
             // optimize for no notes at all
             Lazy<bool> hasAnyNotes = new(() => AppSettings.ShowGitNotes && getUnfilteredRefs().Any(i => i.CompleteName == GitRefName.RefsNotesPrefix));
+            bool firstRevisionReceived = false;
+            bool headIsHandled = false;
 
             try
             {
@@ -882,21 +888,19 @@ namespace GitUI
 
                 _buildServerWatcher.CancelBuildStatusFetchOperation();
 
-                GitModule capturedModule = Module;
-
                 base.Refresh();
 
                 IndexWatcher.Reset();
 
                 _isReadingRevisions = true;
                 _superprojectCurrentCheckout = null;
-                Subject<GitRevision> revisions = new();
+                Subject<GitRevision> observeRevisions = new();
                 _revisionSubscription?.Dispose();
-                _revisionSubscription = revisions
+                _revisionSubscription = observeRevisions
                     .ObserveOn(ThreadPoolScheduler.Instance)
                     .Subscribe(OnRevisionRead, OnRevisionReaderError, OnRevisionReadCompleted);
 
-                System.Threading.CancellationToken cancellationToken = _cancellationTokenSequence.Next();
+                System.Threading.CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
 
                 _selectionTimer.Enabled = false;
                 _selectionTimer.Stop();
@@ -909,13 +913,14 @@ namespace GitUI
                 _gridView.SelectionChanged += OnGridViewSelectionChanged;
                 _gridView.ResumeLayout();
 
-                // Add back and show the spinner controls, which was removed by SetPage call
+                // Add the spinner controls, removed by SetPage()
                 Controls.Add(_loadingControlSpinner);
                 Controls.Add(_loadingControlText);
                 ShowLoading();
 
                 cancellationToken.ThrowIfCancellationRequested();
 
+                // Evaluate GitRefs and current commit
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
                     await TaskScheduler.Default;
@@ -939,22 +944,17 @@ namespace GitUI
                     }
                     else
                     {
-                        // This is a new checkout, so ensure the variable is cleared out.
+                        // This is a new checkout
                         _selectedObjectIds = null;
+                        ResetNavigationHistory();
                         CurrentCheckout = newCurrentCheckout;
                     }
 
                     UpdateSelectedRef(capturedModule, getUnfilteredRefs, headRef);
                     SelectInitialRevision(newCurrentCheckout);
+                    _superprojectCurrentCheckout = await GetSuperprojectCheckoutAsync(capturedModule, noLocks: true);
+
                     semaphoreCurrentCommit.Release();
-                    await this.SwitchToMainThreadAsync(cancellationToken);
-
-                    SuperProjectInfo scc = await GetSuperprojectCheckoutAsync(capturedModule, noLocks: true);
-
-                    if (_superprojectCurrentCheckout != scc)
-                    {
-                        _superprojectCurrentCheckout = scc;
-                    }
 
                     _selectionTimer.Enabled = true;
                     _selectionTimer.Start();
@@ -971,18 +971,19 @@ namespace GitUI
                     out bool parentsAreRewritten);
                 ParentsAreRewritten = parentsAreRewritten;
 
+                // Get info about all Git commits, update the grid
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
                     await TaskScheduler.Default;
                     await new RevisionReader().ExecuteAsync(
-                        Module,
-                        revisions,
+                        capturedModule,
+                        observeRevisions,
                         args,
                         cancellationToken);
                 }).FileAndForget(
                     ex =>
                     {
-                        revisions.OnError(ex);
+                        observeRevisions.OnError(ex);
                         return false;
                     });
             }
@@ -1219,7 +1220,7 @@ namespace GitUI
                 this.InvokeAsync(() => SetPage(new ErrorControl()))
                     .FileAndForget();
 
-                _cancellationTokenSequence.CancelCurrent();
+                _refreshRevisionsSequence.CancelCurrent();
 
                 // Rethrow the exception on the UI thread
                 this.InvokeAsync(() => throw new AggregateException(exception))
@@ -1260,7 +1261,6 @@ namespace GitUI
                         SetPage(_gridView);
                         _isRefreshingRevisions = false;
                         CheckAndRepairInitialRevision();
-                        ResetNavigationHistory();
                         HighlightRevisionsByAuthor(GetSelectedRevisions());
 
                         if (ShowBuildServerInfo)
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -921,19 +921,15 @@ namespace GitUI
                     await TaskScheduler.Default;
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    string branchName = Module.IsValidGitWorkingDir()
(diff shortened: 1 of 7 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -914,40 +916,61 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+                    await TaskScheduler.Default;
+
+                    // Find all ambiguous refs (including stash, notes etc)
+                    // NB: This can be EXTREMELY slow operation for large repos.
+                    IReadOnlyList<IGitRef> refs = (getRefs ?? Module.GetRefs)(RefsFilter.NoFilter);
+                    _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(refs));
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    string pathFilter = BuildPathFilter(_filterInfo.PathFilter);
+                    _revisionReader.Execute(
```
gerhardol, line 935: cancellationtoken shold be propagated to RevisionReader (an internal token is used now)
I tried this without seeing any difference though

gerhardol (commented):
+1
With the repos I work with, I do not see much difference though, loading was async already.

I tested with Linux repo. There are likely fewer refs though, so the experiences will differ.
Are there other Git commands in `PerformRefreshRevisions()` that require substantial time?
Any of the tags/branches commands that requires "more async"?

RussKie:
> With the repos I work with, I do not see much difference though, loading was async already.

This only really surfaces in repos with huge history and large number of refs, like the VS repo.

> Are there other Git commands in `PerformRefreshRevisions()` that require substantial time?

I haven't really spent that much time investigating, but from the high level I think GetRefs is one of the biggest blockers in this flow, as it is executed synchronously and before the RevisionReader is invoked. The left panel is then getting bound synchronously too, which makes the app appear unresponsive.
There are improvements possible, like trying to load refs asynchronously, and then in continuation load the left panel and the revision grid. The left panel loading can be improved where each node (like Branches, Remotes, Tags) can load itself in async mode. The data loading can also happen on demand (i.e. upon the node expansion). Neither of these are quick fixes though.

And I can't stop thinking that our data load implementations are serious hinderance. It could have been much cleaner to load the data separately from presentation, i.e. the revision grid should get bound to an observable collection that gets populated with commits info. The collection then gets be updated with refs (like current branch, remote branch, tags) info and other aux info, and through the data binding mechanisms this information would be reflected in the grid...
However I don't have any capacity to even POC this 😞

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: I had that thought too. The CT would need to be [linked](https://docs.microsoft.com/dotnet/standard/threading/how-to-listen-for-multiple-cancellation-requests) because we could be different source of cancellation - both in the revision grid and the revision reader.
Right now we have several layers of CTs, that can be considered somewhat excessive but work I left them as-is, as this change will require significantly more testing than I can afford atm.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -914,40 +916,61 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
                 {
-                    var scc = await GetSuperprojectCheckoutAsync(capturedModule, noLocks: true);
-                    await this.SwitchToMainThreadAsync();
-
-                    if (_superprojectCurrentCheckout != scc)
+                    cancellationToken.ThrowIfCancellationRequested();
+                    await TaskScheduler.Default;
+
+                    // Find all ambiguous refs (including stash, notes etc)
+                    // NB: This can be EXTREMELY slow operation for large repos.
```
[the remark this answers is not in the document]
RussKie: `_ambiguousRefs` aren't used in this method at all, and is a placeholder for some other on-demand functionality. Since it isn't actually needed, wrapping it in a Lazy avoids doing unnecessary work.

The VS repo has ~180K tags, which make GetRefs call very expensive. I guess the Linux repo doesn't suffer from the same insanity. In my test scenarios I disabled the Tags node in left panel, and made the revision grid loading (that includes loading refs and revisions) async. This makes the app appear significantly more responsive.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Revision Reader has RunAsync, this should be changed if revgrid is running in a background task.
(That should be simple, I can push an update.)

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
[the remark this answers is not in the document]
gerhardol: Slightly OT: How slow was this at startup in post 3.5 before #9734/#9735 when GefRefs was running like 9 times?
And in 3.5 this is like 5 times (plus running 4 times when opening Dashboard?
(one time in master)
Getting ahead/behind is done in two steps and with many refs this would be improved by sharing similar to GetRefs().

RussKie:
I initially rolled back #9735, and it was really bad. #9735 makes it much
better.

gerhardol:
> > Are there other Git commands in `PerformRefreshRevisions()` that require substantial time?
> 
> I haven't really spent that much time investigating, but from the high level I think GetRefs is one of the biggest blockers in this flow, as it is executed synchronously and before the RevisionReader is invoked. The left panel is then getting bound synchronously too, which makes the app appear unresponsive. There are improvements possible, like trying to load refs asynchronously, and then in continuation load the left panel and the revision grid. The left panel loading can be improved where each node (like Branches, Remotes, Tags) can load itself in async mode. The data loading can also happen on demand (i.e. upon the node expansion). Neither of these are quick fixes though.

GetRefs() could be running in paralell to git-log, the refs are required by the observer updating the grid.
(So lazy methods in the observer should be sufficient).

But I doubt there  would be much difference, maybe even slow down the display of revisions in the grid. git-log output is read when seen which is relative quick.
If the refs are read in parallel then the time to first revision could be increased.
This does not apply (much) if the log is sorted, then start display is delayed.

> And I can't stop thinking that our data load implementations are serious hinderance. It could have been much cleaner to load the data separately from presentation, i.e. the revision grid should get bound to an observable collection that gets populated with commits info. The collection then gets be updated with refs (like current branch, remote branch, tags) info and other aux info, and through the data binding mechanisms this information would be reflected in the grid... However I don't have any capacity to even POC this 😞

It is not so much to change for that...

```bash
/c/dev/gc/gitextensions_4 

$ git log --pretty=format:"%H" --all | wc -l
26213

$ time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1
real    0m0.139s
$ time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | tail -1
real    0m0.405s

$ cd ../linux
$ git log --pretty=format:"%H" --all | wc -l
1072918

$ time git -c log.showSignature=false log --max-count=10000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1
real    0m0.864s
$ time git -c log.showSignature=false log --max-count=10000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | tail -1
real    0m22.326s

$ time git -c log.showSignature=false log --max-count=10000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary --date-order  | head -1
real    0m9.009s
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -840,7 +840,8 @@ private void ShowLoading(bool sync = true)
         ///  Queries git for the new set of revisions and refreshes the grid.
         /// </summary>
         /// <exception cref="AggregateException"></exception>
-        public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs = null)
+        // TODO: pass the cancellation token everywhere
```
gerhardol, line 843: As `RefreshRevisions()` is called from many locations, why not use an internal token as currently in RevisionReader?

RussKie:
```
$ git log --pretty=format:"%H" --all | wc -l
256317

$ time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1
real    0m10.353s
user    0m0.000s
sys     0m0.031s

$ time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | tail -1
real    0m13.065s
user    0m0.265s
sys     0m0.358s
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: I'd appreciate it, thanks

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: I'm not sure what you mean. 

`RevisionReader` is instantiated and used internally by `RevisionGridControl`. The CT can come from outside (e.g., the `FormBrowse`).

gerhardol: I suggest the CT is internal to `PerformRefreshRevisions()`, similar to how the `RevisionReader` currently handles it.
If one refresh is cancelled, a new is normally request immediately.

gerhardol:
> For the VS repo:

Was this the first time or after a few runs (to be comparable, run the commands a couple of times.)

Not much difference between first and last (would be interesting to know why).
And for getting refs?

```
time git --no-optional-locks for-each-ref --sort="-*committerdate" --sort="-committerdate" --format="%(if)%(authordate)
%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)" > /dev/null
```

The effect of running in parallel:

```
time git -c log.showSignature=false log --max-count=10000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1 > /dev/null & time git --no-optional-locks for-each-ref --sort="-*committerdate" --sort="-committerdate" --format="%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)" > /dev/null &  time wait
```

gerhardol:
I pushed an update that separates GetRefs() and git-log.
For me the GetRefs() will not yield unless I `await Task.Delay(1)` but if getting refs is slower, it may make a difference.
Really a wip (you can of course just throw it away).

gerhardol:
@RussKie 
A few more adjustments, now the Git commands run in parallel.

I believe this is ready for review

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -914,16 +913,46 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+                // Add back and show the spinner control, which was removed by SetPage call
+                Controls.Add(_loadingControlSync);
+                ShowLoading();
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // Find all ambiguous refs (including stash, notes etc)
+                // Note: GetRefs can be very slow operation for large repos, start in parallel to git-log.
+                // refsByObjectId is needed when first output from git-log is handled
+                Func<IReadOnlyList<IGitRef>> refs = () => (getRefs ?? Module.GetRefs)(RefsFilter.NoFilter);
```
mstv, line 925:
```suggestion
                Func<IReadOnlyList<IGitRef>> refsFactory = () => (getRefs ?? Module.GetRefs)(RefsFilter.NoFilter);
```
or `getUnfilteredRefs`

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -914,16 +913,46 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+                _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(refs()));
+                refsByObjectId = new(() => refs().ToLookup(head => head.ObjectId));
```
mstv, line 927:
```suggestion
                refsByObjectId = new(() => refs().ToLookup(head => head.ObjectId), LazyThreadSafetyMode.ExecutionAndPublication);
```

This seems to be unnecessary for `_ambiguousRefs` because it looked as it was used from the UI thread only.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -957,6 +989,27 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
 
             return;
 
+            static void UpdateSelectedRef(GitModule module, Func<IReadOnlyList<IGitRef>> refs, string branchName)
```
mstv, line 992:
```suggestion
            static void UpdateSelectedRef(GitModule module, Func<IReadOnlyList<IGitRef>> getRefs, string branchName)
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: getUnfilteredRefs

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: The same Lazy handling is used for GetRefs()
https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization#thread-safe-initialization
I interpret that as LazyThreadSafetyMode.ExecutionAndPublication is not needed, but the rest of the doc is unclear...

Changing this though (not for getRefs())

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: ok

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -817,13 +817,13 @@ public void ReloadTranslation()
...
-        private void ShowLoading(bool sync = true)
+        private void ShowLoading(bool showSpinner = true)
         {
-            _loadingControlSync.Visible = sync;
+            _loadingControlSync.Visible = showSpinner;
             _loadingControlSync.Left = (ClientSize.Width - _loadingControlSync.Width) / 2;
             _loadingControlSync.Top = (ClientSize.Height - _loadingControlSync.Height) / 2;
 
-            _loadingControlAsync.Visible = !sync;
+            _loadingControlAsync.Visible = !showSpinner;
```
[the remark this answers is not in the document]
gerhardol: _loadingControlSync is the spinner shown at creation and when refreshing until first revision
_loadingControlAsync is the "Loading" text while still revisions are read

So showSpinner makes sense, renaming the controls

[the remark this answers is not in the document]
gerhardol: Added documentation

The Async/Text "Loading..." was not shown since some time ago

Some other refresh, obviously waiting some 20h more before squashing for proper review

gerhardol:
Closing, replaced by #9864
fb3dcfacd8ad8bb5c061f2caada83aa450e2478d corresponds to c3d1d7067ff5b6b6f5ea024439351dfd457f607d

gerhardol:
@RussKie 
I am still interested to see the results for
https://github.com/gitextensions/gitextensions/pull/9853#issuecomment-1031191726

and of course what you see with #9864

RussKie:
> 
> ```
> time git --no-optional-locks for-each-ref --sort="-*committerdate" --sort="-committerdate" --format="%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)" > /dev/null
> ```

real    0m19.253s
user    0m0.031s
sys     0m0.000s



> Both commands in sequence:
> 
> ```
> time (time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1 > /dev/null ; time git --no-optional-locks for-each-ref --sort="-*committerdate" --sort="-committerdate" --format="%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)" > /dev/null)`
> ```

real    0m10.276s
user    0m0.000s
sys     0m0.030s

real    0m10.954s
user    0m0.000s
sys     0m0.015s

real    0m21.230s
user    0m0.000s
sys     0m0.045s


> 
> The effect of running in parallel:
> 
> ```
> time (time git -c log.showSignature=false log --max-count=1000000 -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --all --boundary | head -1 > /dev/null & time git --no-optional-locks for-each-ref --sort="-*committerdate" --sort="-committerdate" --format="%(if)%(authordate)%(then)%(objectname) %(refname)%(else)%(*objectname) %(*refname)%(end)" > /dev/null & wait)
> ```


real    0m12.705s
user    0m0.015s
sys     0m0.031s

real    0m13.369s
user    0m0.000s
sys     0m0.000s

real    0m13.392s
user    0m0.015s
sys     0m0.078s

Issue #9792, opened 2022-01-03
RevisionGrid not refreshed after resolving merge conflicts


gerhardol:
### Environment

- GitExtensions version: master 07d87634f5392dc9263f47e0cbd616334362c14d plus a few patches
- Git 2.34.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 5.0.13
- DPI 96dpi (no scaling)

### Issue description

The RevisionGrid not refreshed after resolving merge conflicts, occasionally.
The workaround is not hard (refresh the grid), but this looks bad.

![image](https://user-images.githubusercontent.com/6248932/147977372-3673fba4-7ecb-40d7-a345-ad278c77fe6b.png)

### Steps to reproduce

This is hard to reproduce, I get it relative often at work (where I cannot debug) with complex merges but then I just have to dismiss and "remember for later" (so it is fuzzy when this really started).
I have a problem reproducing at home. One in 10 times "succeeded".

Something like the following reproduced the problem. Since this is likely a timing issue all steps are likely not needed, but this was the scenario when reproducing.

* Rebase interactive of a branch with two commits on one new commit, all of the commits have done conflicting changes to the same file.
* order changed for the commits to rebase.
* Dismiss FormRebase when applying first commit fails
* Start FormResolveConflict from toolbar ![image](https://user-images.githubusercontent.com/6248932/147976368-60bad5b8-5374-49f6-a49d-cbfb9b400d59.png)
* Resolve with mergetool
* Continue in toolbar
* Start FormResolveConflict from toolbar
* Resolve with mergetool
* Continue in toolbar

grid is not refreshed as seen above.

### Did this work in previous version of GitExtensions?

This started sometime around or after .NET5 I believe, maybe some time after.
It could even be triggered by #8809.

RussKie:
Thank you for raising it, I observed this as well but never got around raising an issue or investigating.

mstv:
Duplicate of #9223?

Issue #9223, opened 2021-05-19
RevisionGrid not refreshed correctly after interactive rebase operations


mstv:
## Current behaviour

The RevisionGrid is not refreshed after advanced / interactive rebase operations often.
After `Continue` / `Abort` from the `Rebase is currently in progress` banner, the Revision Graph disappears.

## Expected behaviour

Grid is updated, graph is visible.

## Steps to reproduce

- select the current commit
- navigate to previous
- context menu `Advanced | Edit commit` -> the `Rebase ...` banner appears but the revisions are not updated often 
- `Continue` / `Abort` -> graph disappears

## Screenshots

rows not updated:

![grafik](https://user-images.githubusercontent.com/36601201/118893024-1480f180-b902-11eb-850e-13fe823be12b.png)

OK after `F5`:

![grafik](https://user-images.githubusercontent.com/36601201/118893215-5c077d80-b902-11eb-9daa-b493c87898a6.png)

After `Abort`:

![no graph](https://user-images.githubusercontent.com/36601201/118892377-1dbd8e80-b901-11eb-95e3-3778a8d31d52.png)

## Workaround?

Turn off `FileSystemWatcher` for index

![grafik](https://user-images.githubusercontent.com/36601201/118893900-7f7ef800-b903-11eb-834d-2994c2e28e76.png)

## Did this work in previous version of GitExtensions

at least not for months

## Environment

- Git Extensions 3.6.0.11771
- Build 43a5e53eb416870cd4d1d64e3a56a71c55e9733b (Dirty)
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

pmiossec:
I'm on 3f4ed21998af4cb8bd5e105c5c93943d94ec4ca9 (last commit before .net 5 PR merged) and I just noticed thing strange behavior but not consistent.

On one repo, just a refresh with 'F5' make it disappear (scrolling make it appears):
![fail_refresh_graphlines](https://user-images.githubusercontent.com/460196/118895757-113c3480-b907-11eb-8dbe-babdc480c2e7.gif)

>Workaround
>Turn off FileSystemWatcher for index

I don't have it enabled and have the issue.

I don't think I made a rebase (but not sure)

mstv:
> I don't have it enabled and have the issue.
I don't think I made a rebase (but not sure)

So you can reproduce #9108.

[...]

PR #9864, opened 2022-02-13
RefreshGrid: async reading refs and logs in parallel

gerhardol:
Co-authored by RussKie
This replaces #9853
(changes pushed to #9853 too).

## Proposed changes

Improve the responsiveness in the loading phase and decrease
the revision loading time.

Revision info is read in two parallel steps:

1. Read current commit, refs, prepare grid etc.
2. Read all revisions with git-log.

Both these operations can take a long time and is done async.
Before part 1. was read synchrounously and blocked the interface
(git-log was async).
As the information is needed by step 2. when commits are added to the grid,
the information from step 1. is therefore protected with Lazy accessosors
and a semaphore.

RevisionReader (git-log) was previously also updating other information,
but is now a pure revision reader.

Other:

* The text label informing that loading of revisions still occur was
not displayed.

* NavigationHistory was normally always reset also if the grid was refreshed.
(With a potential race condition to keep existing history also if a new repo was opened).

* Clear side panel and the commit info panel at the beginning of the refresh.

* Avoid a Git call to get current head

* Optimizations for git-notes, assuming no notes exists.

## Screenshots 

Showing the difference for linux, limited number of refs so there are repos that benefits more
(after run in the debugger)

3.0 for loading refs is to load ahead/behind for the sidepanel, done in the background

The time differs from run to run, this is showing something I feel is representative.
The biggest difference will be seen in repos with many refs anyway.

The "sync" phase in the VS repo would be 10 s:
https://github.com/gitextensions/gitextensions/pull/9853#issuecomment-1030990082

### Before

170 ms in the sync phase (33 ms to load current HEAD)
Another 30 ms before logs are loaded
3.1 s to load revisions

3.3 s to load all

image

### After

image

(most of 196 ms after ls-files is time between activating and 
A few ms blocking the UI thread,
110 ms (a few more) in the spinner phase before loading refs, in this case the first rev is shown displayed just after that.
3.3s to load completely - git-log is more or less running from start to end

In this case the total loading time is actually a few ms more, but the git-log

image

image

## Test methodology 

Tests slightly updated, but mostly manual

## Merge strategy

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -5,6 +5,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using GitExtUtils;
 using GitUI;
@@ -33,7 +34,7 @@ namespace GitCommands
     }
 #pragma warning restore SA1025 // Code should not contain multiple whitespace in a row
 
-    public sealed class RevisionReader : IDisposable
+    public sealed class RevisionReader
     {
         private const string FullFormat =
 
@@ -55,66 +56,17 @@ namespace GitCommands
 
         // Trace info for parse errors
         private static int _noOfParseError = 0;
-        private readonly CancellationTokenSequence _cancellationTokenSequence = new();
 
-        public void Execute(
+        public async Task ExecuteAsync(
             GitModule module,
-            IReadOnlyList<IGitRef> refs,
             IObserver<GitRevision> subject,
-            int maxCount,
-            RefFilterOptions refFilterOptions,
-            string branchFilter,
-            string revisionFilter,
-            string pathFilter)
+            ArgumentBuilder arguments,
+            CancellationToken token)
         {
-            ThreadHelper.JoinableTaskFactory
-                .RunAsync(() => ExecuteAsync(module, refs, subject, maxCount, refFilterOptions, branchFilter, revisionFilter, pathFilter))
-                .FileAndForget(
-                    ex =>
-                    {
-                        subject.OnError(ex);
-                        return false;
-                    });
-        }
-
-        /// <summary>
-        /// The parents for commits are replaced with the parent in the graph (as all commits may not be included)
-        /// See https://git-scm.com/docs/git-log#Documentation/git-log.txt---parents
-        /// </summary>
-        public bool ParentsAreRewritten { get; private set; } = false;
-
-        private async Task ExecuteAsync(
-            GitModule module,
-            IReadOnlyList<IGitRef> refs,
-            IObserver<GitRevision> subject,
-            int maxCount,
-            RefFilterOptions refFilterOptions,
-            string branchFilter,
-            string revisionFilter,
-            string pathFilter)
-        {
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-
-            var token = _cancellationTokenSequence.Next();
+            await TaskScheduler.Default;
+            token.ThrowIfCancellationRequested();
 
             var revisionCount = 0;
-
-            await TaskScheduler.Default;
-
-            token.ThrowIfCancellationRequested();
-
-            var branchName = module.IsValidGitWorkingDir()
-                ? module.GetSelectedBranch()
-                : "";
-
-            token.ThrowIfCancellationRequested();
-
-            UpdateSelectedRef(module, refs, branchName);
-            var refsByObjectId = refs.ToLookup(head => head.ObjectId);
-
-            token.ThrowIfCancellationRequested();
-
-            var arguments = BuildArguments(maxCount, refFilterOptions, branchFilter, revisionFilter, pathFilter);
 
 #if DEBUG
             var sw = Stopwatch.StartNew();
@@ -138,9 +90,6 @@ namespace GitCommands
 
                     if (TryParseRevision(chunk, logOutputEncoding, sixMonths, out GitRevision? revision))
                     {
-                        // Look up any refs associated with this revision
-                        revision.Refs = refsByObjectId[revision.ObjectId].AsReadOnlyList();
-
                         revisionCount++;
 
                         subject.OnNext(revision);
@@ -159,13 +108,14 @@ namespace GitCommands
             }
         }
 
-        private ArgumentBuilder BuildArguments(int maxCount,
+        public static ArgumentBuilder BuildArguments(int maxCount,
             RefFilterOptions refFilterOptions,
             string branchFilter,
             string revisionFilter,
-            string pathFilter)
+            string pathFilter,
+            out bool parentsAreRewritten)
         {
-            ParentsAreRewritten = !string.IsNullOrWhiteSpace(pathFilter) || !string.IsNullOrWhiteSpace(revisionFilter);
+            parentsAreRewritten = !string.IsNullOrWhiteSpace(pathFilter) || !string.IsNullOrWhiteSpace(revisionFilter);
             return new GitArgumentBuilder("log")
             {
                 { maxCount > 0, $"--max-count={maxCount}" },
@@ -208,7 +158,7 @@ namespace GitCommands
                 },
                 revisionFilter,
                 {
-                    ParentsAreRewritten,
+                    parentsAreRewritten,
                     new ArgumentBuilder
                     {
                         { AppSettings.FullHistoryInFileHistory, $"--full-history" },
@@ -221,30 +171,9 @@ namespace GitCommands
                 },
                 { !string.IsNullOrWhiteSpace(pathFilter), $"-- {pathFilter}" }
             };
-        }
 
-        private static bool IsSimpleBranchFilter(string branchFilter) =>
-            branchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
-
-        private static void UpdateSelectedRef(GitModule module, IReadOnlyList<IGitRef> refs, string branchName)
-        {
-            var selectedRef = refs.FirstOrDefault(head => head.Name == branchName);
-
-            if (selectedRef is not null)
-            {
-                selectedRef.IsSelected = true;
-
-                var localConfigFile = module.LocalConfigFile;
-                var selectedHeadMergeSource = refs.FirstOrDefault(
-                    head => head.IsRemote
-                         && selectedRef.GetTrackingRemote(localConfigFile) == head.Remote
-                         && selectedRef.GetMergeWith(localConfigFile) == head.LocalName);
-
-                if (selectedHeadMergeSource is not null)
-                {
-                    selectedHeadMergeSource.IsSelectedHeadMergeSource = true;
-                }
-            }
+            static bool IsSimpleBranchFilter(string branchFilter) =>
+                branchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
         }
 
         private static bool TryParseRevision(in ArraySegment<byte> chunk, in Encoding logOutputEncoding, long sixMonths, [NotNullWhen(returnValue: true)] out GitRevision? revision)
@@ -417,11 +346,6 @@ namespace GitCommands
             }
         }
 
-        public void Dispose()
-        {
-            _cancellationTokenSequence.Dispose();
-        }
-
         #region Nested type: StringLineReader
 
         /// <summary>
@@ -498,10 +422,6 @@ namespace GitCommands
             {
                 _revisionReader = revisionReader;
             }
-
-            internal ArgumentBuilder BuildArgumentsBuildArguments(int maxCount, RefFilterOptions refFilterOptions,
-                string branchFilter, string revisionFilter, string pathFilter) =>
-                _revisionReader.BuildArguments(maxCount, refFilterOptions, branchFilter, revisionFilter, pathFilter);
 
             internal static bool TryParseRevision(ArraySegment<byte> chunk, Encoding logOutputEncoding, long sixMonths, [NotNullWhen(returnValue: true)] out GitRevision? revision) =>
                 RevisionReader.TryParseRevision(chunk, logOutputEncoding, sixMonths, out revision);
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -74,7 +74,6 @@ namespace GitUI
         public static readonly string HotkeySettingsName = "RevisionGrid";
 
         private readonly TranslationString _droppingFilesBlocked = new("For you own protection dropping more than 10 patch files at once is blocked!");
-        private readonly TranslationString _cannotHighlightSelectedBranch = new("Cannot highlight selected branch when revision graph is loading.");
         private readonly TranslationString _noRevisionFoundError = new("No revision found.");
         private readonly TranslationString _baseForCompareNotSelectedError = new("Base commit for compare is not selected.");
         private readonly TranslationString _strLoading = new("Loading");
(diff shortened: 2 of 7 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -864,47 +886,24 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
 
-                _revisionReader ??= new RevisionReader();
-
-                // Find all ambiguous refs (including stash, notes etc)
-                IReadOnlyList<IGitRef> refs = (getRefs ?? Module.GetRefs)(RefsFilter.NoFilter);
-                _ambiguousRefs = GitRef.GetAmbiguousRefNames(refs);
+                System.Threading.CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
 
+                _selectionTimer.Enabled = false;
+                _selectionTimer.Stop();
```
mstv, line 906:
```suggestion
                _selectionTimer.Stop();
```
These pairs have been in the existing code but they are synonymous.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -430,15 +433,7 @@ public Point GetQuickItemSelectorLocation()
...
-            var selectedRevisions = GetSelectedRevisions();
-            if (selectedRevisions.Count == 1)
-            {
-                _navigationHistory.Push(selectedRevisions[0].ObjectId);
-            }
-            else
-            {
-                _navigationHistory.Clear();
-            }
+            _navigationHistory.Clear();
```
mstv, line 436: ❓ I have not analyzed how it did / should work, and have not run it yet. At first glance, the following might be what it should (have) be(en):
```
            _navigationHistory.Clear();
            if (selectedRevisions.Count == 1)
            {
                _navigationHistory.Push(selectedRevisions[0].ObjectId);
            }
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -817,16 +812,24 @@ public void ReloadTranslation()
             Translator.Translate(this, AppSettings.CurrentTranslation);
         }
 
-        private void ShowLoading(bool sync = true)
+        /// <summary>
+        /// Show spinner (in the synchronous part of loading revisions
+        /// (and creating the control), or the "Loading..." text when first revision is
+        /// handled and the grid is being updated.
+        /// Note that these controls are removed by SetPage() when the fgrid is loaded.
```
mstv, line 819:
```suggestion
        /// Note that these controls are removed by SetPage() when the grid is loaded.
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -849,9 +852,28 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
                 return;
             }
 
+            // Revision info is read in two parallel steps:
+            // 1. Read current commit, refs, prepare grid etc.
+            // 2. Read all revisions with git-log.
+            // Both these operations can take a long time and is done async.
+            // Step 2. requires that most of the information in 1. is prepared when updating the grid.
+            // Git will will provide this information when available (so slower to first info if sorting).
+            // This information from step 1. is therefore protected with Lazy accessosors and a semaphore.
```
mstv, line 861:
```suggestion
            // Git will provide this information when available (so slower to first info if sorting).
            // This information from step 1. is therefore protected with Lazy accessors and a semaphore.
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -849,9 +852,28 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+            System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(1, 1);
+            semaphoreCurrentCommit.Wait();
```
mstv, line 863:
```suggestion
            // This local semaphore instance is bound in closures below.
            System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(initialCount: 0, maxCount: 1);
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -817,16 +812,24 @@ public void ReloadTranslation()
             Translator.Translate(this, AppSettings.CurrentTranslation);
         }
 
-        private void ShowLoading(bool sync = true)
+        /// <summary>
+        /// Show spinner (in the synchronous part of loading revisions
+        /// (and creating the control), or the "Loading..." text when first revision is
```
mstv, line 817:
```suggestion
        /// Show spinner (in the synchronous part of loading revisions
        /// and creating the control), or the "Loading..." text when first revision is
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -849,9 +852,28 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+            // Both these operations can take a long time and is done async.
+            // Step 2. requires that most of the information in 1. is prepared when updating the grid.
+            // Git will will provide this information when available (so slower to first info if sorting).
+            // This information from step 1. is therefore protected with Lazy accessosors and a semaphore.
+            System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(1, 1);
+            semaphoreCurrentCommit.Wait();
             _isRefreshingRevisions = true;
-            ShowLoading();
 
+            GitModule capturedModule = Module;
```
mstv, line 866: ❔ Should `capturedModule` be used in the following lines, too, shouldn't it?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1098,6 +1159,12 @@ void OnRevisionRead(GitRevision revision)
                     flags |= RevisionNodeFlags.OnlyFirstParent;
                 }
 
+                if (!hasAnyNotes.Value)
+                {
+                    // No notes at all in this repo
+                    revision.HasNotes = true;
```
mstv, line 1165: ❓
```suggestion
                if (!hasAnyNotes.Value)
                {
                    // No notes at all in this repo
                    revision.HasNotes = false;
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: This happened to work before as there were no selected revisions yet when this was running. So the history was always reset. (The risk for a race condition was minimal.)
This call is changed now to reset when HEAD is changed so the navigation history contains commits in the grid.
I realize right now that the handling is not working when a filter is applied, the filter must always be refreshed (or the reset is changed to validate or so and is reset when the history contains commits that is not in the grid).
(#9833 would create popups when the history is invalid.)

I have not used the history so often: Shift-Alt-Left (or right)

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Yes, does not make a difference there. Module was used in the async code before...

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: No, HasNotes means that Notes is checked and added to Body.

RussKie (commented):
Nice!

The app remains responsive during loading the VS repo.
image

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: There is no point in letting the timer run before revisions may be added

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -849,9 +852,28 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
             }
 
+            // Revision info is read in two parallel steps:
+            // 1. Read current commit, refs, prepare grid etc.
+            // 2. Read all revisions with git-log.
+            // Both these operations can take a long time and is done async.
+            // Step 2. requires that most of the information in 1. is prepared when updating the grid.
+            // Git will will provide this information when available (so slower to first info if sorting).
+            // This information from step 1. is therefore protected with Lazy accessosors and a semaphore.
+            System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(1, 1);
```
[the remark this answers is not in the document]
gerhardol: If "using System.Thread" is added, then there is a conflict for Timer (also in System.Windows.Forms).

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -914,40 +913,79 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+
+                // Evaluate GitRefs and current commit
+                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    _selectionTimer.Enabled = false;
-                    _selectionTimer.Stop();
+                    await TaskScheduler.Default;
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    _ = refsByObjectId.Value;
```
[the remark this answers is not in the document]
gerhardol: Force calculation, so it is not done in the git-log thread when handling the first revision.
It is basically getRefs() that takes time and this happens to be used anyway. refsByObjectId is not more than a ms for me (so likely not noticeable in a large repo either).

GitCommands/RevisionReader.cs
```diff
@@ -159,13 +108,14 @@ public sealed class RevisionReader : IDisposable
             }
         }
 
-        private ArgumentBuilder BuildArguments(int maxCount,
+        public static ArgumentBuilder BuildArguments(int maxCount,
```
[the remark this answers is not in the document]
gerhardol: Just the method for `ParentsAreRewritten` could be exposed though, but I found that it was easier to see what the RevisionReader was doing this way.This could be moved to GitCommandHelper.cs but the format is internal to the RevisionReader.

gerhardol:
Set to draft as no diffs are shown and I do not find the issue right now...

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: I will revert this, there are no revisions shown yet anyway, so the change can be in the early phase.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Moved this to just before where it is needed, before RunAsync

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Investigated this some more. Removing this.
This was likely an attempt to force run the selection timer after grid was loaded. Initially it seem to be used at the first load only but was running at every update.
However, as  git-log was async already before this PR (and has been for a long time), it was just forcing an update for no reason as the selection timer would trigger one anyway (maybe even before this timer fired).
It may have been a way to clear info in the tabs similar to what RussKie added in this PR for side panel, but it seem to OK without.
(Will play around some more.)

RussKie:
I'm observing an unexpected behaviour with this change - when I check change between commits (i.e., checkout a different one) in VS repo - the commit panel and the diff panels no longer update. Switching to another commit, selecting another commit or switching repos don't help. The only option is to restart the app. I am yet to find some time to debug this.

gerhardol:
> I'm observing an unexpected behaviour with this change - when I check change between commits (i.e., checkout a different one) in VS repo - the commit panel and the diff panels no longer update. Switching to another commit, selecting another commit or switching repos don't help. The only option is to restart the app. I am yet to find some time to debug this.

Are you using the first commit, tabs was not updating there.

I will revert the reset of the side panel at refresh (but not when new repo is opened), that is just annoying.
The handling of _selectionTimer will be readdded, to Reset tabs (and commit info) while reading revisions.

I may add some other optimizations to indexwatcher etc (that is reset three times right now) - or in another PR if that fits better.
Some init is done many times. This is much less of a performance issue than any unnecessary Git command though.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Adding this, to initiate a refresh of tab info.
This is initiated before Refs are read, previously it could take some time before this occurred.
(Removing the change that refreshed only the Commit info, in tab or in the panel)

gerhardol:
A few more changes, more separated to #9871 (there will be a minor merge conflict.) These PRs belong together, maybe #9871 should be merged first.

I see one issue: The submodules are always collapsed at a refresh.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -912,42 +905,88 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+                ShowLoading();
                 _gridView.ResumeLayout();
 
+                IndexWatcher.Reset();
+
+                // Selected is null: Reset the tabs etc.
+                _selectionTimer.Enabled = false;
+                _selectionTimer.Stop();
+                _selectionTimer.Enabled = true;
+                _selectionTimer.Start();
```
mstv, line 921: Nit: Setting `Timer.Enabled` is equivalent to calling `Timer.Stop()` / `.Start`. The (partly redundant) property should only be used for checking whether a timer is running.
```suggestion
                _selectionTimer.Stop();
                _selectionTimer.Start();
```

gerhardol: also changed the other occurrences in the file

gerhardol:
Squash and rebased to simplify review (#9833 and #9877 made some parallel changes).
See tmp/load-refs-parallel-to-log for previous commits.

RussKie:
👍

gerhardol:
The revgrid for submodules were not rendered until the super refs were displayed, which could take some seconds.

I have identified a few more optimizations for submodules.
If this can be merged first it would be great, but I want reviews to merge.

RussKie:
Merge it 👍

gerhardol:
fixup! Lazy getUnfilteredRefs also when refreshed from grid

When refreshing from Browse getRefs is Lazy (shared with side panel).
This also uses a Lazy accessor when refreshing from grid
(refresh when changing could be slower than a refresh including side panel as several getRefs() was required)

Updated the comments

> Merge it 👍

Will check a detail for UpdateSelectedRef(), but will otherwise merge tomorrow.
(and add some related followups)

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1067,26 +1131,29 @@ void OnRevisionRead(GitRevision revision)
...
-                    headIsHandled = true;
-                    AddArtificialRevisions(insertAsFirst: false);
-                }
-
-                var flags = RevisionNodeFlags.None;
+                    semaphoreCurrentCommit.Wait();
+                    if (revision.ObjectId.Equals(CurrentCheckout))
+                    {
+                        // Insert worktree/index before HEAD (CurrentCheckout)
+                        // If grid is filtred and HEAD not visible, insert first after all are updated
```
mstv, line 1146:
```suggestion
                        // If grid is filtered and HEAD not visible, insert it as first after all are updated
```

mstv (approved):
The changes look sensible to me. I have run the main changes up to 06e5878 without issues (using vanilla repos) for several days.

mstv:
After merge conflicts and after (interactive) rebase, the RevisionGrid is empty since the recent changes. Before, only the graph was missing. Manual refresh works.

gerhardol:
> After merge conflicts and after (interactive) rebase, the RevisionGrid is empty since the recent changes. Before, only the graph was missing. Manual refresh works.

I have seen it but cannot reproduce reliably, cannot reproduce better than #9223 (or #9792 that is closed but maybe describes the problem better).
Merging.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: fixed
Thanks for the review!

RussKie:
🚀

PR #9791, opened 2022-01-03
Replace AppSettings.UseFastChecks with ShowGitStatusInBrowseToolbar

gerhardol:
## Proposed changes

Remove the specific setting for the index watcher, use the GitStatusMonitor
settings instead. UseFastChecks hardly adds any overhead, it adds
FileSystemWatcher to index and refs (GitStatusMonitor has a separate
watcher on index).
If the file system is changed, the top-left refresh button becomes red.
(As well as avoiding refreshing the grid in some situations, may not be working though.)

UseFastChecks was visible in Performance settings and expected by
users (and occasionally maintainers) to make a difference.
Sometimes the setting is mixed up with the GitStatusMonitor setting
that can make a difference (as git-status requires a lot of resources,
not FileSystemWatcher).

### Before

image

## Test methodology 

Manual

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9871, opened 2022-02-17
IndexWatcher: Use for refresh icon only

### LINKED ISSUE #9792: RevisionGrid not refreshed after resolving merge conflicts

gerhardol:
Related to #9864 
Possibly fixes #9792 #9223

## Proposed changes

The IndexWatcher controls the icon for the refresh button in Browse.
Before #9791 the setting could be controlled individually,
now the same setting as for GitStatus (commit count).

Previously it also tried to optimize grid refresh operations in a few
situations, skipping when no changes was detected to the index or any refs.
These checks are now removed, always refreshing the grid.

The check for changes seem to miss updates occasionally (and never in
WSL where FileSystemWatcher do not reeport updates).
This seem to be improved by matching on creation of index.lock
in addition to index, but then also reset of worktree files
are tracked as index changes.

Before the file watcher was initiated at creation of the control and
when a new revision was loaded. This requires a sync call to
git-rev-parse --git-common-dir delaying the revision loading
50-100 ms.
This is changed to just clear the icon when loading a revision
(as well as when a new repo is loaded even if that should not be required)
and resetting the file system watcher after all revisions was loaded.

The IndexWatcher was reset several additional times after the grid
was refreshed.

The IndexWatcher.IndexChanged was used to possibly refresh the revisions
that was already refreshed when the revisions were selected in the grid.
(Also, no changes should have occurred this time.)

An example where FileSystemWatcher does not fire.

```
echo >> appveyor.yml ; git add appveyor.yml
```

## Test methodology 

Manual.

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -874,11 +874,6 @@ namespace GitUI.CommandsDialogs
 
                 SetShortcutKeyDisplayStringsFromHotkeySettings();
 
-                if (hasWorkingDir)
-                {
-                    ShowRevisions();
-                }
-
                 RefreshWorkingDirComboText();
 
                 OnActivate();
@@ -899,7 +894,6 @@ namespace GitUI.CommandsDialogs
                     toolStripButtonPush.DisplayAheadBehindInformation(branchSelect.Text);
 
                     ActiveControl = RevisionGrid;
-                    RevisionGrid.IndexWatcher.Reset();
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


mstv:
Unfortunately, does not fix #9223.

RussKie (approved):

gerhardol:
> Unfortunately, does not fix #9223.

I did not try so hard to recreate #9223, but then #9223 is not a duplicate of #9792, at least the workaround is not the same

Changes still makes sense to me, will merge in two days unless there are comments

PR #9885, opened 2022-03-05
RevGrid: Lazy handling for current branch

gerhardol:
Followup to #9864

## Proposed changes

Avoid Git calls where no branch is checked out and in submodules
This is similar to how HEAD is cached in CurrentCheckout.

For repos where a Git branch is checked out, this is handled by reading files, so some IO is needed.
This adds some ms here and there, nothing critical (but affecting performance a lot more than most programming constructs).

For submodules where normally no branch is checked out, this requires Git commands, both when loading revisions and opening menues etc.

## Screenshots 

No UI changed, just some commands (only startup shown).

### Before

image

### After

image

## Test methodology 

manual
Removed an incorrect test and adopted another.
(A branch name like `(test)` is actually legal and should be displayed as `((test))` in the title.

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -123,6 +123,7 @@ namespace GitUI
         // NOTE internal properties aren't serialised by the WinForms designer
 
         internal ObjectId? CurrentCheckout { get; private set; }
+        internal Lazy<string> CurrentBranch { get; private set; } = new(() => "");
         internal FilterInfo CurrentFilter => _filterInfo;
         internal bool ShowUncommittedChangesIfPossible { get; set; } = true;
         internal bool ShowBuildServerInfo { get; set; }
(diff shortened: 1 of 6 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2616,12 +2620,7 @@ private void CompareWithCurrentBranchToolStripMenuItem_Click(object sender, Even
                 return;
             }
 
-            var headBranchName = Module.RevParse(headBranch);
-
-            if (headBranchName is not null)
-            {
-                UICommands.ShowFormDiff(baseCommit.ObjectId, headBranchName, baseCommit.Subject, headBranch);
-            }
+            UICommands.ShowFormDiff(baseCommit.ObjectId, CurrentCheckout, baseCommit.Subject, CurrentBranch.Value);
```
mstv, line 2623: Should be correct. But I could not test this with detached head because `CompareWithCurrentBranch` is disabled then.

mstv (approved):
LGTM

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: The difference in behavior is that return is early, with a popup. (That should not occur as you mention).

RussKie (approved):

PR #9887, opened 2022-03-06
RevisionGrid: Set accessibility values

### LINKED ISSUE #9852: Accessibility no longer works on main tree
### Environment

- Git Extensions 3.5.4.12724
- Git 2.31.1.windows.1
- .NET Framework 4.8.4420.0
- DPI 96dpi (no scaling)

### Issue description

In previous versions, I could use my screen reader. Now it only says "Row 123" instead of the commit message. See screenshot with inspect.exe.

image

### Steps to reproduce

Run inspect.exe (C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64\inspect.exe) and hover over a commit row. Commit message is not accessible.

Yes, but I don't remember which unfortunately.

### Diagnostics

_No response_

mstv:
Fixes #9852

## Proposed changes

- Set missing accessibility `Value` for `RevisionGrid` columns:
  - message
  - commit id
- Set empty string for accessibility property `Name` of all `RevisionGrid` cells
  because it contained redundant information "Message Row nnn"

### Before

grafik

### After

grafik

## Test methodology 

- manual: run "...\Windows Kits\10\bin\10.0.19041.0\x86\inspect.exe"  and hover the grid cells

## Test environment(s) 

- Git Extensions 33.33.33
- Build 687a8e3de13f4171b17e3ddd6d767225c3237840
- Git 2.35.1.windows.1
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
@@ -60,6 +60,15 @@ namespace GitUI.UserControls.RevisionGrid.Columns
             }
         }
 
+        public override void OnCellFormatting(DataGridViewCellFormattingEventArgs e, GitRevision revision)
+        {
+            // Set the grid cell's accessibility text
+            if (!revision.IsArtificial)
+            {
+                e.Value = revision.ObjectId.ToShortString();
+            }
+        }
+
         public override bool TryGetToolTip(DataGridViewCellMouseEventArgs e, GitRevision revision, [NotNullWhen(returnValue: true)] out string? toolTip)
         {
             if (revision.ObjectId.IsArtificial)
--- a/GitUI/UserControls/AccessibleDataGridViewTextBoxCell.cs
+++ b/GitUI/UserControls/AccessibleDataGridViewTextBoxCell.cs
@@ -1,0 +1,27 @@
+﻿using System.Windows.Forms;
+
+namespace GitUI.UserControls
+{
+    /// <summary>
+    /// A <cref>DataGridViewTextBoxCell</cref> with <cref>CreateAccessibilityInstance</cref> overridden
+    /// in order to suppress the <cref>Name</cref> "Message Row nnn", which is redundant
+    /// because the row and column numbers are provided in dedicated accessibility properties.
+    /// </summary>
+    public class AccessibleDataGridViewTextBoxCell : DataGridViewTextBoxCell
+    {
+        protected class DataGridViewTextBoxCellUnnamedAccessibleObject : DataGridViewTextBoxCellAccessibleObject
+        {
+            public DataGridViewTextBoxCellUnnamedAccessibleObject(DataGridViewCell? owner) : base(owner)
+            {
+            }
+
+            public override string Name
+                => string.Empty;
+        }
+
+        protected override AccessibleObject CreateAccessibilityInstance()
+        {
+            return new DataGridViewTextBoxCellUnnamedAccessibleObject(this);
+        }
+    }
+}
--- a/GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/MessageColumnProvider.cs
@@ -95,6 +95,12 @@ namespace GitUI.UserControls.RevisionGrid.Columns
             }
         }
 
+        public override void OnCellFormatting(DataGridViewCellFormattingEventArgs e, GitRevision revision)
+        {
+            // Set the grid cell's accessibility text
+            e.Value = revision.Subject.Trim();
+        }
+
         public override bool TryGetToolTip(DataGridViewCellMouseEventArgs e, GitRevision revision, [NotNullWhen(returnValue: true)] out string? toolTip)
         {
             _toolTipBuilder.Clear();
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -28,6 +28,8 @@ namespace GitUI.UserControls.RevisionGrid
 
     public sealed partial class RevisionDataGridView : DataGridView
     {
+        private static readonly AccessibleDataGridViewTextBoxCell _accessibleDataGridViewTextBoxCell = new();
+
         private readonly SolidBrush _alternatingRowBackgroundBrush;
         private readonly SolidBrush _authoredHighlightBrush;
 
@@ -205,6 +207,7 @@ namespace GitUI.UserControls.RevisionGrid
             _columnProviders.Add(columnProvider);
 
             columnProvider.Column.Tag = columnProvider;
+            columnProvider.Column.CellTemplate = _accessibleDataGridViewTextBoxCell;
 
             Columns.Add(columnProvider.Column);
         }


GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
```diff
@@ -60,6 +60,15 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
             }
         }
 
+        public override void OnCellFormatting(DataGridViewCellFormattingEventArgs e, GitRevision revision)
+        {
+            // Set the grid cell's accessibility text
+            if (!revision.IsArtificial)
```
gerhardol, line 66: Why limit artificial?
Is this displayed somewhere other than for accessibility?
(I do not see that when commenting out the test.)

gerhardol (approved):
Have not checked accessibility

GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
mstv: > Why limit artificial?

We do not want to display "111...1" anywhere.

> Is this displayed somewhere other than for accessibility? (I do not see that when commenting out the test.)

No, it isn't. `OnCellPainting` draws the cell. I have just restored setting `e.Value`, which was removed by mistake in #8550.

RussKie:
I don't think this change is addressing the original issue as the accessible name value remains "Message Zeile 2".

RussKie:
@vladimir-krestov @SergeySmirnov-Akvelon @guybark  do you think you could give us a hand here? I struggle to find a way to override `DataGridViewTextBoxCell.DataGridViewTextBoxCellAccessibleObject`'s properties such as `Name`, `Description`, etc. Any thoughts on whether it is possible?
Thank you.

RussKie:
I tried the following, and it didn't work:

```diff
diff --git a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
index 51c1561cc..aed592d60 100644
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -345,6 +345,8 @@ internal bool MultiSelect
             set => _gridView.MultiSelect = value;
         }
 
+        internal DataGridViewCell GetCell(int row, int column) => _gridView.Rows[row].Cells[column];
+
         private static void FillMenuFromMenuCommands(IEnumerable<MenuCommand> menuCommands, ToolStripDropDownItem targetItem)
```

```dif
diff --git a/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs b/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
index c1ba632c5..ea3d58350 100644
--- a/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/CommitIdColumnProvider.cs
@@ -37,6 +37,8 @@ public CommitIdColumnProvider(RevisionGridControl grid)
 
         public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevision revision, int rowHeight, in CellStyle style)
         {
+            DataGridViewCell cell = _grid.GetCell(e.RowIndex, e.ColumnIndex);
+
             var monospaceFont = style.MonospaceFont;
             if (!_widthByLengthByFont.TryGetValue(monospaceFont, out var widthByLength))
             {
@@ -51,11 +53,19 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
 
                 if (i == -1 && Column.Width > widthByLength[widthByLength.Length - 1])
                 {
-                    _grid.DrawColumnText(e, revision.ObjectId.ToString(), monospaceFont, style.ForeColor, e.CellBounds, useEllipsis: false);
+                    string hash = revision.ObjectId.ToString();
+                    cell.AccessibilityObject.Name = hash;
+                    cell.AccessibilityObject.Value = hash;
+
+                    _grid.DrawColumnText(e, hash, monospaceFont, style.ForeColor, e.CellBounds, useEllipsis: false);
                 }
                 else if (i > 1)
                 {
-                    _grid.DrawColumnText(e, revision.ObjectId.ToShortString(i - 1), monospaceFont, style.ForeColor, e.CellBounds, useEllipsis: false);
+                    string hash = revision.ObjectId.ToShortString(i - 1);
+                    cell.AccessibilityObject.Name = hash;
+                    cell.AccessibilityObject.Value = hash;
+
+                    _grid.DrawColumnText(e, hash, monospaceFont, style.ForeColor, e.CellBounds, useEllipsis: false);
                 }
             }
         }
```

IIUIC, any descendant of `Control` can set [`AccessibleName` property](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.accessiblename), however `DataGridViewTextBoxCell` is not a descendant of `Control` type.

mstv:
> I don't think this change is addressing the original issue as the accessible name value remains "Message Zeile 2".

The issue description was unclear / confusing: https://github.com/gitextensions/gitextensions/issues/9852#issuecomment-1059969421.

The `Name` property has always been "Message Row xxx". Up to v3.4.3 / #8550, the property `Value` contained the commit message. I have restored this.

@justanotheranonymoususer, please confirm whether https://ci.appveyor.com/api/buildjobs/cqfaxktufd5d845u/artifacts/artifacts%2FRelease%2Fpublish%2FGitExtensions-Portable-4.0.0.13552-f95404ed5.zip works for you.

justanotheranonymoususer:
Yes, now it works as before. I still get "Message Row xxx" which I'd
rather not get, but I also get the actual text which is what's
important.

SergeySmirnov-Akvelon:
> @vladimir-krestov @SergeySmirnov-Akvelon @guybark do you think you could give us a hand here? I struggle to find a way to override `DataGridViewTextBoxCell.DataGridViewTextBoxCellAccessibleObject`'s properties such as `Name`, `Description`, etc. Any thoughts on whether it is possible? Thank you.

Hi @RussKie. I was able to do this using inherited classes.
image

```
        ...
        this.Column1 = new DataGridViewTextBoxColumn() { CellTemplate = new SubDataGridViewTextBoxCell() };
        ...

        public class SubDataGridViewTextBoxCell : DataGridViewTextBoxCell
        {
            protected class SubDataGridViewTextBoxCellAccessibleObject : DataGridViewTextBoxCellAccessibleObject
            {
                public SubDataGridViewTextBoxCellAccessibleObject(DataGridViewCell? owner) : base(owner)
                {
                }

                public override string Name => $"Test Name {Owner.Value}";

                public override string Description => $"Test Description {Owner.Value}";
            }

            protected override AccessibleObject CreateAccessibilityInstance()
            {
                return new SubDataGridViewTextBoxCellAccessibleObject(this);
            }
        }
```

justanotheranonymoususer:
Looking at other elements, e.g. browser tabs, taskbar button, I see
that Name is set, and Value and Description are empty or not
implemented. Thanks.

RussKie:
Thank you @SergeySmirnov-Akvelon!

@mstv @justanotheranonymoususer: according to the docs, [`AccessibleName`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.accessiblename#remarks) describes the element _such as the text in a Button, the name of a MenuItem, or a label displayed next to a TextBox control_, and  [`AccessibleDescription`](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.accessibledescription#remarks) _convey information that is less tangible, such as "A button that shows a picture of a cactus"_.

So based on this information, I'd say for a row
* `AccessibleName` would be something like "Commit Row 123",
* `AccessibleDescription` would be something like "\<Subject\> by \<Author\> made on \<date\>. Hash \<hash\>",

...and each cell would have the following:
* `AccessibleName` would be something like "\<ColumnType\> for Commit Row 123", where _ColumnType_ is Author, Author avatar, Date, Commit hash, Subject, etc.
* `AccessibleDescription`  - I think we could leave it empty, because the name and the value make it redundant.
* `Value` would contains the actual text the cell has, e.g., author name, the subject or the date.

We also have branches and tags information, and I'm not sure how these should be presented, i.e., should they be announced at the row level or the cell level?


The above is a description of a desired state, and I don't expect this to be implemented in this PR. We can start with the select columns and set the properties at the cell level.

Thoughts?

mstv:
`Name` is empty now. `Value` contains the major grid contents.

RussKie (approved):
Works for me if works for users with accessibility needs.
Thank you for doing this.

PR #9888, opened 2022-03-06
Init sidepanel only when loading

gerhardol:
Follow up to #9864 

## Proposed changes

* Previously, the side panel was initiated from Browse (UICommands_PostRepositoryChanged is almost always invoked by Browse changes regardless or not if the repo is changed),
then again after loading the grid if a filter was applied or changed.
(normally the loaded branches etc were reused, so no additional Git call just making this confusing and some additional loading).
The second part is needed if the change was initiated from RevGrid itself (normally when changing filters etc).

The side panel init is now only started when loading the grid.
Update is forced if the repo changes (set for instance for F5 refreshes).

Some parts separated to #9896 and #9897

## Test methodology 

manual

## Merge strategy

Rebase merge 

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.BranchTree.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.BranchTree.cs
@@ -34,17 +34,6 @@ namespace GitUI.BranchTreePanel
 
             protected override bool SupportsFiltering => true;
 
-            protected override void OnAttached()
-            {
-                IsFiltering.Value = false;
-            }
-
-            protected override Task PostRepositoryChangedAsync(GitUIEventArgs e)
-            {
-                IsFiltering.Value = false;
-                return ReloadNodesAsync(LoadNodesAsync, e.GetRefs);
-            }
-
             protected override async Task<Nodes> LoadNodesAsync(CancellationToken token, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
             {
                 await TaskScheduler.Default;
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
@@ -124,12 +124,8 @@ namespace GitUI.BranchTreePanel
             private readonly CancellationTokenSequence _reloadCancellationTokenSequence = new();
             private bool _firstReloadNodesSinceModuleChanged = true;
 
-            // A flag to indicate whether the data is currently being filtered or not.
-            // This helps to reduce unnecessary treeview rebinds.
-            private bool _isCurrentlyFiltering;
-
             // A flag to indicate whether the data is being filtered (e.g. Show Current Branch Only).
-            private protected static AsyncLocal<bool> IsFiltering = new();
+            private protected AsyncLocal<bool> IsFiltering = new();
 
             protected Tree(TreeNode treeNode, IGitUICommandsSource uiCommands)
             {
@@ -151,41 +147,13 @@ namespace GitUI.BranchTreePanel
                     // the same time, we don't want to remove them from the tree as this is visible to the user,
                     // as well as less efficient.
                     _firstReloadNodesSinceModuleChanged = true;
-
-                    // Rebind callbacks
-                    if (e.OldCommands is not null)
-                    {
-                        e.OldCommands.PostRepositoryChanged -= UICommands_PostRepositoryChanged;
-                    }
-
-                    uiCommands.UICommands.PostRepositoryChanged += UICommands_PostRepositoryChanged;
                 };
-
-                uiCommands.UICommands.PostRepositoryChanged += UICommands_PostRepositoryChanged;
             }
 
             public void Dispose()
             {
                 Detached();
                 _reloadCancellationTokenSequence.Dispose();
-            }
-
-            private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
-            {
-                if (!IsAttached)
-                {
-                    return;
-                }
-
-                ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
-                {
-                    await PostRepositoryChangedAsync(e);
-                }).FileAndForget();
-            }
-
-            protected virtual Task PostRepositoryChangedAsync(GitUIEventArgs e)
-            {
-                return Task.CompletedTask;
             }
 
             public TreeNode TreeViewNode { get; }
@@ -208,6 +176,7 @@ namespace GitUI.BranchTreePanel
 
             protected virtual void OnAttached()
             {
+                IsFiltering.Value = false;
             }
 
             public void Detached()
@@ -230,8 +199,6 @@ namespace GitUI.BranchTreePanel
 
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    IsFiltering.Value = _isCurrentlyFiltering;
-
                     await ReloadNodesAsync(LoadNodesAsync, getRefs);
                 });
             }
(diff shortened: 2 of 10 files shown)


GitUI/BranchTreePanel/RepoObjectsTree.BranchTree.cs
```diff
@@ -39,12 +39,6 @@ protected override void OnAttached()
                 IsFiltering.Value = false;
             }
 
-            protected override Task PostRepositoryChangedAsync(GitUIEventArgs e)
-            {
-                IsFiltering.Value = false;
```
RussKie, line 44: How is `IsFiltering` used now?

gerhardol: Not changed at all, it is just that the panel is updated only once.
(Maybe the property can be removed?)

gerhardol: `IsFiltering` is required but `_isCurrentlyFiltering` can be removed.
However, when testing this I get a race condition causing _isCurrentlyFiltering to be set before IsFiltering so if the refresh is initiated from the grid (filter only) or Browse (could be module), the result may differ.
I changed this to only fully refresh the side panel on  actual module changes, so refreshing from grid or browse is the same. 

Cannot see a need to force the refresh.
If so this can be handled in the Loading event.

gerhardol: The loaded structure must be invalidated when changed (like a branch is added) so the latest commit will need to be adjusted.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -862,7 +860,7 @@ private void ShowLoading(bool showSpinner = true)
         ///  Queries git for the new set of revisions and refreshes the grid.
         /// </summary>
         /// <exception cref="AggregateException"></exception>
-        public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs = null)
+        public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs = null, bool updateSidepanel = false)
```
RussKie, line 863: This name implies this control needs to know about some other unrelated control, which is undesirable.
```suggestion
        public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs = null, bool forceRefresh = false)
```

Also please add the xmldoc explaining the parameter.

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
```diff
@@ -9,12 +9,13 @@ public class GitUIEventArgs : CancelEventArgs
     {
         private readonly IFilteredGitRefsProvider _getRefs;
 
-        public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs)
+        public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool isForceUpdate)
```
RussKie, line 12:
```suggestion
        public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)
```

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
```diff
@@ -32,5 +34,6 @@ public GitUIEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands)
         public IGitModule GitModule => GitUICommands.GitModule;
 
         public IReadOnlyList<IGitRef> GetRefs(RefsFilter filter) => _getRefs.GetRefs(filter);
+        public bool IsForceUpdate { get; init; }
```
RussKie, line 37:
```suggestion
        public bool ForceRefresh { get; init; }
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -252,11 +252,11 @@ public void Initialize(IAheadBehindDataProvider? aheadBehindDataProvider, Action
         ///  <see langword="true"/>, if the data is being filtered; otherwise <see langword="false"/>.
         /// </param>
         /// <param name="getRefs">Function to get refs.</param>
-        public void Refresh(bool isFiltering, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
+        public void Refresh(bool isFiltering, bool isForceUpdate, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
```
RussKie, line 255:
```suggestion
        public void Refresh(bool isFiltering, bool forceRefresh, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
```

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -124,12 +124,8 @@ private abstract class Tree : IDisposable
             private readonly CancellationTokenSequence _reloadCancellationTokenSequence = new();
             private bool _firstReloadNodesSinceModuleChanged = true;
 
-            // A flag to indicate whether the data is currently being filtered or not.
-            // This helps to reduce unnecessary treeview rebinds.
-            private bool _isCurrentlyFiltering;
-
             // A flag to indicate whether the data is being filtered (e.g. Show Current Branch Only).
-            private protected static AsyncLocal<bool> IsFiltering = new();
+            private protected AsyncLocal<bool> IsFiltering = new();
```
RussKie, line 128: Why did you remove the `static` modifier? Do we expect more than one instance of the left panel now?

gerhardol: This replaces the `_isCurrentlyFiltering` flag, it would have been cumbersome to keep two flags for the same info.

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
gerhardol: will change this (and other comments)
plan to move this to a separate class as well, this is an odd fit here.

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: done

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
gerhardol: done

Plugins/GitUIPluginInterfaces/GitUIEventArgs.cs
gerhardol: done

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: done

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -736,7 +744,7 @@ private void RegisterPlugins()
                         if (plugin.Execute(new GitUIEventArgs(this, UICommands)))
                         {
                             _gitStatusMonitor.InvalidateGitWorkingDirectoryStatus();
-                            UICommands.RepoChangedNotifier.Notify();
+                            RefreshRevisions();
```
mstv, line 747: A drawback of this might be that a future subscriber of `UICommands.RepoChangedNotifier` might not get updated.

mstv (commented):
Have not noticed side-effects so far.

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: If the repo is changed, they will.
If a notification is required for other reasons, there should be a separate notification (like the existing GridLoading).
The intention of this was to make the handling possible to understand.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -62,7 +62,7 @@ public sealed partial class RevisionGridControl : GitModuleControl, IScriptHostC
     {
         public event EventHandler<DoubleClickRevisionEventArgs>? DoubleClickRevision;
         public event EventHandler<FilterChangedEventArgs>? FilterChanged;
-        public event EventHandler<GitUIEventArgs>? RevisionGraphLoaded;
+        public event EventHandler<GridLoadingEventArgs>? RevisionGraphLoading;
```
RussKie, line 65: I suggest to retain the event to indicate the data is loaded as well. It may be useful for plugins.

```suggestion
        /// <summary>
        ///  Occurs whenever the revision graph has been populated with the data.
        /// </summary>
        public event EventHandler<DataLoadEventArgs>? DataLoaded;

        /// <summary>
        ///  Occurs whenever the revision graph has started loading the data.
        /// </summary>
        public event EventHandler<DataLoadEventArgs>? DataLoading;
```

gerhardol: I fam a little hesitant to add features without a direct need, but can add this.

GitUI/UserControls/RevisionGrid/GridLoadingEventArgs.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Windows.Forms;
+using GitUIPluginInterfaces;
+
+namespace GitUI.UserControls.RevisionGrid
+{
+    public class GridLoadingEventArgs : GitUIEventArgs
+    {
+        public GridLoadingEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)
```
RussKie, line 10:
```suggestion
        public GridLoadEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)
```

RussKie (commented):
👍

GitUI/UserControls/RevisionGrid/GridLoadingEventArgs.cs
gerhardol: done

mstv (approved):

gerhardol:
Rebased and squashed without changes, rebase merge tomorrow

Issue #9884, opened 2022-03-03
RevisionGrid: Restore selections after refresh


mstv:
**Current Behavior**

The checked-out revision (and one of its files) is selected on refresh of the RevisionGrid.

**Expected Behavior**

The selected commit and the selected file are re-selected after refreshing (unless gone).

**StR**

![grafik](https://user-images.githubusercontent.com/36601201/156161313-7ee1fefc-3479-4a3a-8011-ac7431adf024.png)

`F5`

![grafik](https://user-images.githubusercontent.com/36601201/156161532-f6fc3010-6658-44ca-884c-c72d7e583285.png)

**Environment**

- Git Extensions 33.33.33
- Build ade0a61828201cb642e4fb7bc1599441268bbf53
- Git 2.35.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 5.0.12
- DPI 96dpi (no scaling)

PR #9896, opened 2022-03-13
Share gitRefs when reloading after changing sort criterias

gerhardol:
Part of #9888 (that is follow up to #9864)

Leaves the "double side panel init" to later.
#9884 is already approved but have a refresh issue that I do not know how to solve in a good way and therefore separate most of the changes.

## Proposed changes

* Changing filters required 3 separate getRefs() calls (so it could be slower than refreshing the complete grid).
* Share getRefs() after loading the grid when filters are applied (normally, this is not needed though if the structure is not rebuilt).

Getting refs can take >10s in large repos and should only not be done 3 times when only one is needed.
Future: This could possibly be cached too, requiring an "external" refresh to reread the cache.

## Screenshots 

For changing sort criteria

### Before

image

### After

resort-2

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
@@ -224,7 +224,7 @@ namespace GitUI.BranchTreePanel
             /// <summary>
             /// Requests to refresh the data tree and to apply filtering, if necessary.
             /// </summary>
-            protected internal virtual void Refresh()
+            protected internal virtual void Refresh(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
             {
                 // NOTE: descendants may need to break their local caches to ensure the latest data is loaded.
 
@@ -232,7 +232,7 @@ namespace GitUI.BranchTreePanel
                 {
                     IsFiltering.Value = _isCurrentlyFiltering;
 
-                    await ReloadNodesAsync(LoadNodesAsync, new FilteredGitRefsProvider(Module).GetRefs);
+                    await ReloadNodesAsync(LoadNodesAsync, getRefs);
                 });
             }
 
@@ -242,7 +242,7 @@ namespace GitUI.BranchTreePanel
             /// <param name="isFiltering">
             ///  <see langword="true"/>, if the data is being filtered; otherwise <see langword="false"/>.
             /// </param>
-            internal void ToggleFilterMode(bool isFiltering)
+            internal void Refresh(bool isFiltering, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
             {
                 // If we're not currently filtering and no need to filter now -> exit.
                 // Else we need to iterate over the list and rebind the tree - whilst there
@@ -262,7 +262,7 @@ namespace GitUI.BranchTreePanel
                 {
                     IsFiltering.Value = true;
 
-                    await ReloadNodesAsync(LoadNodesAsync, new FilteredGitRefsProvider(Module).GetRefs);
+                    await ReloadNodesAsync(LoadNodesAsync, getRefs);
                 });
             }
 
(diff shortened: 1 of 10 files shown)


RussKie (approved):

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -242,7 +242,7 @@ protected internal virtual void Refresh()
             /// <param name="isFiltering">
             ///  <see langword="true"/>, if the data is being filtered; otherwise <see langword="false"/>.
             /// </param>
-            internal void ToggleFilterMode(bool isFiltering)
+            internal void Refresh(bool isFiltering, Func<RefsFilter, IReadOnlyList<IGitRef>> refs)
```
mstv, line 245: ❔ The argument is not used (in this non-virtual function).

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
```diff
@@ -224,7 +224,7 @@ protected virtual void OnDetached()
             /// <summary>
             /// Requests to refresh the data tree and to apply filtering, if necessary.
             /// </summary>
-            protected internal virtual void Refresh()
+            protected internal virtual void Refresh(Func<RefsFilter, IReadOnlyList<IGitRef>> refs)
```
mstv, line 227: ❔ Shall the new argument be used by overridden functions only?

mstv (commented):
`refs` rather is `getRefs`.

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
gerhardol: No, it will be used on line 235, missed that when backporting
(and will change to getRefs)

GitUI/BranchTreePanel/RepoObjectsTree.Nodes.cs
gerhardol: As above

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -301,7 +301,7 @@ protected override void Dispose(bool disposing)
         /// This is used to remove spinners added when loading and to replace the gridview at errors.
         /// </summary>
         /// <param name="content">The content to show.</param>
-        private void SetPage(Control content)
+        private void SetPage(Control content, Lazy<IReadOnlyList<IGitRef>> getRefs)
```
mstv, line 304:
```suggestion
        private void SetPage(Control content, Lazy<IReadOnlyList<IGitRef>> refs)
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -402,21 +402,21 @@ public void SetAndApplyPathFilter(string filter)
             PerformRefreshRevisions();
         }
 
-        private void InitiateRefAction(IReadOnlyList<IGitRef>? refs, Action<IGitRef> action, FormQuickGitRefSelector.Action actionLabel)
+        private void InitiateRefAction(IReadOnlyList<IGitRef>? getRefs, Action<IGitRef> action, FormQuickGitRefSelector.Action actionLabel)
```
mstv, line 405:
```suggestion
        private void InitiateRefAction(IReadOnlyList<IGitRef>? refs, Action<IGitRef> action, FormQuickGitRefSelector.Action actionLabel)
```

gerhardol: Yes reverting this.

Plan to merge later today.

mstv (approved):

Issue #9899, opened 2022-03-15
Async loading of diffs


mstv:
### Feature description

In my daily work, I often encounter large commits, e.g. merges of platform updates or of vendor branches.
It can take 10-20 seconds to load the diffs.
Then GE is blocked. And Windows considers the application "not reponding", pushes the GE window to the background and moves the GE instance to the end of the list in the Task Bar (strange Windows feature).

```
File name:   C:\Program Files\Git\bin\git.exe
Arguments:   -c color.ui=never -c diff.submodule=short -c diff.noprefix=false -c diff.mnemonicprefix=false -c diff.ignoreSubmodules=none -c core.safecrlf=false diff --find-renames --find-copies --name-status -z "63cd6b96c93a8e1749c245cd35902a9cda644198" "3b92cc568a67444db085ddd8b2f018dba59483cc"
Working dir: d:\xyz\g3\
Process ID:  17248
Started at:  2022-03-14T15:46:49.7024322+01:00
UI Thread?:  True
Duration:    17438,729 ms
Exit code:   0
```

### Environment

- Git Extensions 33.33.33
- Build 2b1ef6c8eb561cdad3be04f1f1f5b11e261163de
- Git 2.35.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 5.0.12
- DPI 96dpi (no scaling)

PR #9906, opened 2022-03-19
Process events before PerformRefreshRevisions

### LINKED ISSUE #9223: RevisionGrid not refreshed correctly after interactive rebase operations

### LINKED ISSUE #9792: RevisionGrid not refreshed after resolving merge conflicts

mstv:
Fixes #9223
Fixes #9792

## Proposed changes

- before calling `RevisionGrid.PerformRefreshRevisions`:
  - process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
- reset `RevisionGridControl._isRefreshingRevisions` before calling `LoadingCompleted` and `SetPage`
- minor tuning of `RevisionGraph.Clear`: clear Dictionary instead of creating a new one

### Before

no graph or empty RevisionGrid after rebase / merge / conflict resolution operations

### After

RevisionGrid is updated without need for manual refresh

## Test methodology 

- manual, e.g.
  - start rebase with conflicts
  - close dialog
  - abort 

## Test environment(s) 

- Git Extensions 33.33.33
- Build 10d51fd4222272d8cb5997b50ad386b7c02f0284
- Git 2.35.1.windows.1
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -568,6 +568,9 @@ namespace GitUI.CommandsDialogs
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
+            // Process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
+            Application.DoEvents();
+
             // Note that this called in most FormBrowse context to "be sure"
             // that the repo has not been updated externally.
             RefreshRevisions(e.GetRefs);
@@ -596,14 +599,15 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
+            InternalInitialize();
+            RefreshGitStatusMonitor();
+
             Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
             RevisionGrid.PerformRefreshRevisions(getRefs);
 
-            InternalInitialize();
             ToolStripFilters.UpdateBranchFilterItems(getRefs);
             UpdateSubmodulesStructure();
 
-            RefreshGitStatusMonitor();
             revisionDiff.RefreshArtificial();
             UpdateStashCount();
         }
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -568,9 +568,6 @@ namespace GitUI.CommandsDialogs
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
-            // Process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
-            Application.DoEvents();
-
             // Note that this called in most FormBrowse context to "be sure"
             // that the repo has not been updated externally.
             RefreshRevisions(e.GetRefs);
@@ -599,15 +596,17 @@ namespace GitUI.CommandsDialogs
                 return;
             }
 
-            InternalInitialize();
-            RefreshGitStatusMonitor();
+            // Process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
+            Application.DoEvents();
 
             Debug.Assert(RevisionGrid.CanRefresh, "Already loading revisions when running RefreshRevisions(). This could cause the commits in the grid to be loaded several times.");
             RevisionGrid.PerformRefreshRevisions(getRefs);
 
+            InternalInitialize();
             ToolStripFilters.UpdateBranchFilterItems(getRefs);
             UpdateSubmodulesStructure();
 
+            RefreshGitStatusMonitor();
             revisionDiff.RefreshArtificial();
             UpdateStashCount();
         }
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -52,7 +52,7 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         {
             _loadingCompleted = false;
             _maxScore = 0;
-            _nodeByObjectId = new ConcurrentDictionary<ObjectId, RevisionGraphRevision>();
+            _nodeByObjectId.Clear();
             _nodes = ImmutableList<RevisionGraphRevision>.Empty;
             _orderedNodesCache = null;
             _orderedRowCache = null;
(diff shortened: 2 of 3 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -568,6 +568,9 @@ private bool NeedsGitStatusMonitor()
 
         private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
         {
+            // Process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
```
gerhardol, line 571: Why not doing this in `RefreshRevisions()`?
I would like to replace all internal invocations of RepoChanged in FormBrowse in #9888

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -596,14 +599,15 @@ private void RefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
                 return;
             }
 
+            InternalInitialize();
```
gerhardol, line 602: This will delay the UI refresh slightly, the Git commands for revisions should be started async early

GitUI/CommandsDialogs/FormBrowse.cs
mstv: Just too many functions with similar names.

GitUI/CommandsDialogs/FormBrowse.cs
mstv: It solved the issue for one of two cases. But `DoEvents` suffices. Reverted this part.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -596,6 +596,9 @@ private void RefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
                 return;
             }
 
+            // Process events e.g. from closing FormProcess, particularly after rebase / conflict resolution actions
+            Application.DoEvents();
```
RussKie, line 600: This is a "smell", meaning we're doing something wrong...
Can `Invoke` instead?

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: The handling should be changed to call RefreshRevisions by default. Should this be moved then?

mstv: > Just too many functions with similar names.

I just wanted to express that I didn't notice in the diff that it was added to a different function.
I inserted it before some refresh call and was happy it worked.

GitUI/CommandsDialogs/FormBrowse.cs
mstv: > This is a "smell", meaning we're doing something wrong...

Yes, but what?

> Can `Invoke` instead?

No, `Invoke((Action)(() => RevisionGrid.PerformRefreshRevisions(getRefs)));` does not suffice.
And this is no surprise because `UICommands_PostRepositoryChanged` is executed in the UI thread already.

gerhardol: The calling was changed in #9897. `RefreshRevisions()` is no longer async, but calculations is in async anyway, except for #9899 if artificial that is worse (still issues before and still not async in other situations) after this - I missed that, the calling structure was too complicated.

RussKie: > And this is no surprise because `UICommands_PostRepositoryChanged` is executed in the UI thread already.

Perhaps this is the underlying problem - we use this event for I/O, and not just for the UI updates. 
I'm not asking for a change, just making an observation.

RussKie: 💭 thinking out loud - may be we need to have two events - one is fired on a background thread for I/O (`UICommands_PostRepositoryChanged`) and another one (`DataRefreshed`) - one the UI thread for the UI updates.

mstv: This PR can wait until after #9888 and perhaps other follow-ups, which might even make it obsolete.
(I just reproduced the addressed issue many times when testing #9907 and had a look.)

vbjay:
Highly recommend not using DoEvents 
https://blog.codinghorror.com/is-doevents-evil/

Issue #9752, opened 2021-11-23
Cannot bring GE to front


gerhardol:
### Environment

- Git Extensions 33.33.33
- Build 4240f4903d3b0d7ec70e73dfc4c5425023f8ad7e
- Git 2.34.0.windows.1
- Microsoft Windows NT 10.0.19043.0
- .NET 5.0.12
- DPI 96dpi (no scaling)

### Issue description

In some situations (normally at work) GE does not respond to being activated. Alt-Tab, clicking in the taskbar has no effect.
Show desktop and other tricks does not help to show the app.
It is possible to close the application from the taskbar. (So it does not seem to be an open dialog.)

### Steps to reproduce

Unknown, occurs mostly at work after some hour usage.

### Did this work in previous version of GitExtensions?

This did not occur prior to .NET5 migration, but it probably started some time after.

vbjay:
Usually caused by owned fork maybe not visible but is there. So it focused
owned form instead but you just can't see it.

gerhardol:
> Usually caused by owned fork maybe not visible but is there. So it focused owned form instead but you just can't see it.

Sure, but what change in master is causing thisto happen more often?
If it occurs to others, it is a blocker for 4.0
(How many are using master for daily work now?)

vbjay:
> > Usually caused by owned fork maybe not visible but is there. So it focused owned form instead but you just can't see it.
> 
> Sure, but what change in master is causing thisto happen more often?
> If it occurs to others, it is a blocker for 4.0
> (How many are using master for daily work now?)

Do you have multiple monitors different scaling?  Might be related to that. Frequency. Master has moved to 6.0 or is still at 5.0 but either way it's not framework any more.  I'm betting dpi handling is a lot different.

 You would think clicking or alt tabbing into a parent window that has an owned form would make sure owned form was not just brought to focus but visible and not lost in out of view area by the framework.

RussKie:
FWIW I'm observing that when the app is minimised the first click on the
task bar icon doesn't restore the app, instead I hear a beep sound. On the
2nd click the app restores... This is happening to the v3.x and v4.x
versions though for me.

gerhardol:
The miniature on the taskbar is OK if there are several instances (it always is).
Sometimes I see the app when clicking the taskbar, it disappears immediately.

Sometimes the problem occurs by just clicking on the taskbar a couple of times.

Found a workaround:
I can restore the app by selecting on the taskbar, then Win-Up.

[...]

PR #9908, opened 2022-03-20
RevDiff: async SetDiff

### LINKED ISSUE #9899: Async loading of diffs

gerhardol:
Fixes #9899

With this I see no more major async needs in the refreshRevision sequence,
(InternalInitialize() calls CheckForUpdates but that is only done occasionally during start.)

## Proposed changes

* Change git-diff file difference calculation to be asynchronous and
display a message indicating that the files are loading (instead of
showing the old diff).

This is also used in FormCommit etc, not much changes there as there is not much you can do until the diff is loaded anyway.

Using the same cancellationToken as is used for showing file contents.

* Browse RefreshRevisions(): Skip RefreshArtificial()

This is redundant as the revision is restored after loading the grid is done.
Duplicate update is normally avoided as revisions are not yet set when
RefreshArtificial() was called.

### Before

The diff before revision is changed, not responding if another revision is selected.

### After

Note: modified to show in the filelist after mstv suggestions, to avoid overlay issues and flickering.

image

## Test methodology 

Manual
Inserting Thread.Sleep(); in FileStatusList.SetDiffs()

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -22,6 +22,7 @@ using GitUI.Properties;
 using GitUI.UserControls;
 using GitUIPluginInterfaces;
 using Microsoft;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitUI
 {
@@ -88,8 +89,11 @@ namespace GitUI
 
             HandleVisibility_NoFilesLabel_FilterComboBox(filesPresent: true);
             NoFiles.Text = TranslatedStrings.NoChanges;
+            LoadingFiles.Text = TranslatedStrings.LoadingData;
             Controls.SetChildIndex(NoFiles, 0);
+            Controls.SetChildIndex(LoadingFiles, 0);
             NoFiles.Font = new Font(NoFiles.Font, FontStyle.Italic);
+            LoadingFiles.Font = new Font(LoadingFiles.Font, FontStyle.Italic);
             FilterWatermarkLabel.Font = new Font(FilterWatermarkLabel.Font, FontStyle.Italic);
             FilterComboBox.Font = new Font(FilterComboBox.Font, FontStyle.Bold);
 
@@ -699,6 +703,21 @@ namespace GitUI
             GitItemStatusesWithDescription = _diffCalculator.SetDiffs(revisions, headId);
         }
 
+        public async Task SetDiffsAsync(IReadOnlyList<GitRevision> revisions, ObjectId? headId, CancellationToken cancellationToken)
+        {
+            _enableDisablingShowDiffForAllParents = true;
+            await this.SwitchToMainThreadAsync(cancellationToken);
+            FileStatusListLoading();
+
+            await TaskScheduler.Default;
+            cancellationToken.ThrowIfCancellationRequested();
+            IReadOnlyList<FileStatusWithDescription> gitItemStatusesWithDescription = _diffCalculator.SetDiffs(revisions, headId);
+
+            await this.SwitchToMainThreadAsync(cancellationToken);
+            cancellationToken.ThrowIfCancellationRequested();
+            GitItemStatusesWithDescription = gitItemStatusesWithDescription;
+        }
+
         /// <summary>
         /// FormStash init for WorkTree and Index.
         /// </summary>
@@ -900,6 +919,17 @@ namespace GitUI
         private void SetFilterWatermarkLabelVisibility()
         {
             FilterWatermarkLabel.Visible = FilterVisibleInternal && !FilterComboBox.Focused && string.IsNullOrEmpty(FilterComboBox.Text);
+        }
+
+        private void FileStatusListLoading()
+        {
+            LoadingFiles.Visible = true;
+            NoFiles.Visible = false;
+
+            FileStatusListView.BeginUpdate();
+            FileStatusListView.Groups.Clear();
+            FileStatusListView.Items.Clear();
+            FileStatusListView.EndUpdate();
         }
 
         private void UpdateFileStatusListView(bool updateCausedByFilter = false)
@@ -1194,6 +1224,7 @@ namespace GitUI
 
         private void HandleVisibility_NoFilesLabel_FilterComboBox(bool filesPresent)
         {
+            LoadingFiles.Visible = false;
             NoFiles.Visible = !filesPresent;
             FilterVisibleInternal = FilterVisible && filesPresent;
         }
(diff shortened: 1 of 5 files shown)


mstv (commented):
This looks promising. Thank you! Will run it the next days.

mstv:
Once, "Loading files" was activated *after* the diffs had been loaded, and remained visible instead of the files.

The file selection is not restored anymore - both with and without #9804.

GitUI/UserControls/FileStatusList.cs
```diff
@@ -902,6 +918,18 @@ private void SetFilterWatermarkLabelVisibility()
             FilterWatermarkLabel.Visible = FilterVisibleInternal && !FilterComboBox.Focused && string.IsNullOrEmpty(FilterComboBox.Text);
         }
 
+        private void FileStatusListLoading()
+        {
+            LoadingFiles.Visible = true;
+            NoFiles.Visible = false;
+            FilterVisibleInternal = false;
```
mstv, line 925: Although hiding the `Filter` ComboBox is correct, I would not do it in order to reduce flicker.

gerhardol: Changed this to only hide watermark. It looks weird when a filter is applied, but less flickering.

image

Also fixed the prematurely submitted PR...

gerhardol:
Rebased to resolve merge conflict

mstv:
This PR brings a great improvement for me.
Though occasionally, "Loading data" is not replaced with the diffs. More often with small diffs on a machine without extra load than with large diffs (and VS Code updating C++ Intellisense).
It can be simplified slightly: 26ed4a9. Doesn't solve this issue.
"Loading data" looks better with eb9b518 IMO.

gerhardol:
> This PR brings a great improvement for me. Though occasionally, "Loading data" is not replaced with the diffs. More often with small diffs on a machine without extra load than with large diffs (and VS Code updating C++ Intellisense). 

I have not seen this myself, cannot see what is causing it.

> It can be simplified slightly: [26ed4a9](https://github.com/gitextensions/gitextensions/commit/26ed4a968d1c50c856b4334efd5f35db968eda77). Doesn't solve this issue. "Loading data" looks better with [eb9b518](https://github.com/gitextensions/gitextensions/commit/eb9b51856cf522b06e4cab1d5de11c6e97bf0c8a) IMO.

Cherry picking part of your suggestions. I find the combined RevDiff.SetDiff() to be slightly less readable than just the core in common and the differences separate.
(But I often make this type of solution myself...)
I would have preferred to have the `Loading` label instead of the box, but keeping it below avoids the flickering and filter overlay issues.

Further, I feel that #9752 occur slightly less often with this.

GitUI/UserControls/FileStatusList.cs
gerhardol: Changed this as suggested in https://github.com/gitextensions/gitextensions/pull/9908#issuecomment-1079298052
Label now shown below the box

mstv:
> > This PR brings a great improvement for me. Though occasionally, "Loading data" is not replaced with the diffs. More often with small diffs on a machine without extra load than with large diffs (and VS Code updating C++ Intellisense).
> 
> I have not seen this myself, cannot see what is causing it.

Just quickly hit `Cursor Down` or `Up` many times in the RevisionGrid showing the GE repo.

image

I think this is a basic rule: Modifications of shared resources (as the UI) by async continuations must be made atomic. If the previous one has already started, the next one must wait until the previous one has cancelled or finished, before the next one *starts* to access the shared resources. (BTW: `lock` would have no effect!) So, after every UI operation which could process the message loop, the CancellationToken must be checked.
Is there a less error-prone approach, @sharwell?
(Kind regards to the developers of Outlook: Try to switch folders in a slow VM.)

> > It can be simplified slightly: [26ed4a9](https://github.com/gitextensions/gitextensions/commit/26ed4a968d1c50c856b4334efd5f35db968eda77). Doesn't solve this issue. "Loading data" looks better with [eb9b518](https://github.com/gitextensions/gitextensions/commit/eb9b51856cf522b06e4cab1d5de11c6e97bf0c8a) IMO.
> 
> Cherry picking part of your suggestions. I find the combined RevDiff.SetDiff() to be slightly less readable than just the core in common and the differences separate. (But I often make this type of solution myself...) I would have preferred to have the `Loading` label instead of the box, but keeping it below avoids the flickering and filter overlay issues.

I feel uneasy adding `ThreadHelper.JoinableTaskFactory.RunAsync` to all callers of `SetDiffsAsync`. I wanted to deal with this in the non-async `SetDiffs`.

> Further, I feel that #9752 occur slightly less often with this.

Luckily, I have seen this only once (with this PR). And I have no need to encounter it again.

gerhardol:
> > > This PR brings a great improvement for me. Though occasionally, "Loading data" is not replaced with the diffs. More often with small diffs on a machine without extra load than with large diffs (and VS Code updating C++ Intellisense).
> > 
> > 
> > I have not seen this myself, cannot see what is causing it.
> 
> Just quickly hit `Cursor Down` or `Up` many times in the RevisionGrid showing the GE repo.

I was not able to reproduce in GE, but in Linux (only once).

But the problem may be caused by the reuse of the cancellation token. If a file is being viewed after next sett is selected, the result would not be shown.
I defined a separate token. (Should have done that from the start, this was a late addition.)
Cannot reproduce after, but had problem reproducing before too.

> I feel uneasy adding `ThreadHelper.JoinableTaskFactory.RunAsync` to all callers of `SetDiffsAsync`. I wanted to deal with this in the non-async `SetDiffs`.

It is only two callers and I find the code easier to understand as is Can change.
@RussKie Any opinion?

> > Further, I feel that #9752 occur slightly less often with this.
> 
> Luckily, I have seen this only once (with this PR). And I have no need to encounter it again.

This issue is extremely annoying. This may improve very little, but not much.

gerhardol:
Added clearing of the fileviewer while calculating revisions

mstv (commented):
+2 Works well! Just a minor tuning yet.

gerhardol:
Squash and rebase tonight, merging a couple of days after mstv formal approval or when Russkie approves.

mstv (approved):

PR #9924, opened 2022-03-29
Correct ownership of Push taskdialog

RussKie:
The dialog is spawned when attempting to push a branch that would result in a rewrite of the remote HEAD.
image

### DIFF
--- a/GitUI/CommandsDialogs/FormPush.cs
+++ b/GitUI/CommandsDialogs/FormPush.cs
@@ -528,11 +528,7 @@ namespace GitUI.CommandsDialogs
             {
                 Debug.Assert(form.Visible, "The progress dialog must be visible.");
 
-                // form is ProgressDialog, and form.Owner is FormPush that spawned the ProgressDialog, so essentially owner == this.
-                IWin32Window owner = form.Owner;
-                Debug.Assert(owner == this, "The progress window must belong to this dialog! This is a bug, please correct and send a pull request with a fix.");
-
-                (AppSettings.PullAction onRejectedPullAction, var forcePush) = AskForAutoPullOnPushRejectedAction(match.Groups["currBranch"].Success);
+                (AppSettings.PullAction onRejectedPullAction, bool forcePush) = AskForAutoPullOnPushRejectedAction(form, match.Groups["currBranch"].Success);
 
                 if (forcePush)
                 {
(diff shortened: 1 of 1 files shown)


mstv (commented):
The GE CLI calls `GitUICommands.StartPushDialog(owner: null, ...)`.
I think it is correct, isn't it?
This PR is about not making a closed form the owner of the TaskDialog.
So, the `Assert`s can be removed.

RussKie:
> The GE CLI calls `GitUICommands.StartPushDialog(owner: null, ...)`.

Thank you. I was trying to think of usecases and I clearly missed this one.

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -527,6 +527,8 @@ private bool HandlePushOnExit(ref bool isError, FormProcess form)
             if (match.Success && !Module.IsBareRepository())
             {
                 IWin32Window owner = form.Owner;
+                Debug.Assert(owner is not null, "Progress window must be owned by another window! This is a bug, please correct and send a pull request with a fix.");
```
RussKie, line 530: At this point `form` is `ProgressDialog` and `form.Owner` is `FormPush`.

GitUI/CommandsDialogs/FormPush.cs
```diff
@@ -643,7 +645,8 @@ private bool HandlePushOnExit(ref bool isError, FormProcess form)
 
                 page.Buttons.Add(btnPushForce);
 
-                TaskDialogButton result = TaskDialog.ShowDialog(Handle, page);
+                Debug.Assert(owner is not null, "The dialog must be owned by another window! This is a bug, please correct and send a pull request with a fix.");
```
RussKie, line 648: Subsequently, this method (`AskForAutoPullOnPushRejectedAction`) is only invoked from `HandlePushOnExit`, so `owner` can never be `null`.

RussKie:
I've run multiple manual verifications testing both "noisy" and "quiet" modes:
```diff
diff --git a/GitExtensions/Properties/launchSettings.json b/GitExtensions/Properties/launchSettings.json
index 3fcc654dd..a5e9f8ed2 100644
--- a/GitExtensions/Properties/launchSettings.json
+++ b/GitExtensions/Properties/launchSettings.json
@@ -2,7 +2,8 @@
   "profiles": {
     "GitExtensions": {
       "commandName": "Project",
-      "commandLineArgs": "browse \"$(MSBuildThisFileDirectory)..\""
+      "commandLineArgs": "push quiet",
+      "workingDirectory": "D:\\Development\\test2\\repo1"
     },
     "Dashboard": {
       "commandName": "Project"
```

mstv (approved):

RussKie:
Thank you

PR #9925, opened 2022-03-30
Blame: open current diff/view line

gerhardol:
Related to #9842 

## Proposed changes

Open the current line in the Blame viewer when switching to
RevDiff diff viewer and FileTree content instead of the
line active before the diff.

This is not important, but can be confusing as diff/view to blame selects the line in blame.

This increases the complexity some in FileViewer though, why this is not an obvious function.
(Can set this as draft or WIP, but it is ready to review.)

## Screenshots 

Also diff -> Tree, RevTree blame-view

### Before

blame-to-diff-old

### After

blame-to-diff-new

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -403,7 +403,7 @@ namespace GitUI.CommandsDialogs
                 DiffFiles.SetSelectedIndex(fileIndex, notify: false);
             }
 
-            loadFileContent = ShowSelectedFileDiffAsync();
+            loadFileContent = ShowSelectedFileDiffAsync(ensureNoSwitchToFilter: false, line: 0);
             return true;
         }
 
@@ -558,7 +558,7 @@ namespace GitUI.CommandsDialogs
         {
             if (DiffFiles.SelectedItem is null)
             {
-                await ShowSelectedFileDiffAsync(ensureNoSwitchToFilter);
+                await ShowSelectedFileDiffAsync(ensureNoSwitchToFilter, line);
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


gerhardol:
The base functionality has a different usecase that is more relevant than switching blame -> diff/view: To stay on line when "following" files.

gerhardol:
@mstv any comments?
Should I go ahead and add "follow line" too?

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -552,7 +552,7 @@ private async Task ShowSelectedFileBlameAsync(bool ensureNoSwitchToFilter, int?
         /// Activate diffviewer if Blame is visible
         /// </summary>
         /// <returns>a task</returns>
-        private async Task ShowSelectedFileDiffAsync(bool ensureNoSwitchToFilter = false)
+        private async Task ShowSelectedFileDiffAsync(bool ensureNoSwitchToFilter, int? line)
```
RussKie, line 555: This does feel wrong to do `int?`.... Maybe test for negativity, e.g., `if (line >= 0)`?

gerhardol: -1 can be used instead of null in the meaning as "there is no line to go to".
This will affect a lot of existing code for blame etc though to avoid to have two implementation so I rather keep nullable.

mstv (approved):
LGTM

> Should I go ahead and add "follow line" too?

I think that will be useful.

RussKie:
Is this something we still wishing to take?

gerhardol:
Yes, I have the follow up to add. Could be a separate pr though.

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -527,8 +527,8 @@ private void blameMenuItem_Click(object sender, EventArgs e)
                 return;
             }
 
+            int? line = FileText.Visible ? FileText.CurrentFileLine : BlameControl.CurrentFileLine;
             blameToolStripMenuItem1.Checked = !blameToolStripMenuItem1.Checked;
-            int? line = FileText.Visible ? FileText.CurrentFileLine : null;
```
mstv, line 531: Why moved? `line` is given to the call below.

mstv (approved):
Just a nit, used for long time

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
gerhardol: I believe that `blameMenuItem_Click()` could change the line number

PR #9907, opened 2022-03-19
"rebase/merge --abort" instead of "reset --hard"

mstv:
Fixes #9575

## Proposed changes

- "rebase/merge --abort" instead "reset --hard"

### Before

image

image

### After

image

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 1e409ce21c9c668b009cfb7f7f2fd4374ac28ac1
- Git 2.35.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 5.0.12
- DPI 96dpi (no scaling)

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #9942, opened 2022-04-06
FormResolveConflicts: Clearer label and confirmation messages for Reset button

mstv:
Replaces #9907, which would be too intrusive and does not work for cherry-picks

## Proposed changes

- FormResolveConflicts: Just clearer label and confirmation messages for `Reset` button
  (The command `git reset --hard` will still be used.)

### Before

image

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build ec324afbbc5dec66a66e58937f98cf2e2130af69
- Git 2.35.1.windows.1
- .NET 6.0.3
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormResolveConflicts.cs
+++ b/GitUI/CommandsDialogs/FormResolveConflicts.cs
@@ -93,7 +93,7 @@ namespace GitUI.CommandsDialogs
                 "All changes since the last commit will be deleted." + Environment.NewLine +
                 Environment.NewLine + "Do you want to reset the changes?");
 
-        private readonly TranslationString _resetCaption = new("Reset changes");
+        private readonly TranslationString _resetCaption = new("Reset");
 
         private readonly TranslationString _areYouSureYouWantDeleteFiles =
             new("Are you sure you want to DELETE all changes?" + Environment.NewLine +
--- a/GitUI/CommandsDialogs/FormResolveConflicts.cs
+++ b/GitUI/CommandsDialogs/FormResolveConflicts.cs
@@ -71,11 +71,9 @@ namespace GitUI.CommandsDialogs
         private readonly TranslationString _openMergeToolItemText = new("Open in");
         private readonly TranslationString _button1Text = new("Open in");
 
-        private readonly TranslationString _resetItemRebaseText = new("Abort rebase");
         private readonly TranslationString _contextChooseLocalRebaseText = new("Choose local (theirs)");
         private readonly TranslationString _contextChooseRemoteRebaseText = new("Choose remote (ours)");
 
-        private readonly TranslationString _resetItemMergeText = new("Abort merge");
         private readonly TranslationString _contextChooseLocalMergeText = new("Choose local (ours)");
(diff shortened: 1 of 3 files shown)


GitUI/CommandsDialogs/FormResolveConflicts.cs
```diff
@@ -91,11 +89,11 @@ static SortableConflictDataList()
...
 
         private readonly TranslationString _abortCurrentOperation =
-            new("You can abort the current operation by resetting changes." + Environment.NewLine +
+            new("You can abort the current conflict resolution by resetting hard." + Environment.NewLine +
                 "All changes since the last commit will be deleted." + Environment.NewLine +
-                Environment.NewLine + "Do you want to reset changes?");
+                Environment.NewLine + "Do you want to reset the changes?");
 
-        private readonly TranslationString _abortCurrentOperationCaption = new("Abort");
+        private readonly TranslationString _resetCaption = new("Reset changes");
```
gerhardol, line 96: I have a minor preference for just Reset, but that may not be clearer...

gerhardol (approved):

RussKie:
> grafik

The new button text is clipped.

mstv:
> Popup title: I have a minor preference for just Reset, but that may not be clearer...

> The new button text is clipped.

I was and am undecided.
Would "Reset to previous" suffice for the button? Or even just "Reset"?
Or should I resize the buttons (with some reserve for lengthy translations)?

gerhardol:
> I was and am undecided. Would "Reset to previous" suffice for the button? Or even just "Reset"? Or should I resize the buttons (with some reserve for lengthy translations)?

I find `Reset` enough, easier to translate and full info in the popup.
(Maybe tooltip)

RussKie:
I think a fix would mean the dialog's updated to inherit from `GitExtensionsDialog` and all the buttons moved to `ControlsPanel` (see https://github.com/gitextensions/gitextensions/issues/6183). This way the buttons can have arbitrary widths. Though I agree that it doesn't mean that we'd want the buttons to contain more text than absolutely necessary. Tooltips can be used to provide extra data. Also each form can/should point to the relevant section in the docs, which should/can contain all the necessary information pertaining to the dialog and individual elements on it.

mstv:
> I find `Reset` enough, easier to translate and full info in the popup.

OK, done.

> (Maybe tooltip)

> Tooltips can be used to provide extra data.

A tooltip would be a good idea. But the designer is broken. It is not worth fixing it right now. I think the updated label and message suffice.

> I think a fix would mean the dialog's updated to inherit from GitExtensionsDialog and all the buttons moved to ControlsPanel (see https://github.com/gitextensions/gitextensions/issues/6183). This way the buttons can have arbitrary widths. Though I agree that it doesn't mean that we'd want the buttons to contain more text than absolutely necessary.

`category: nice to have` Not my cup of tea, anyway. 

> Also each form can/should point to the relevant section in the docs, which should/can contain all the necessary information pertaining to the dialog and individual elements on it.

Already there.

Issue #9935, opened 2022-04-03
Refresh does not re-select the commit


mstv:
### Environment

- Git Extensions 33.33.33
- Build bcd0c2617bdd4c842c9a707888f951af07c65ecd
- Git 2.35.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.3
- DPI 96dpi (no scaling)

### Issue description

Refresh does not re-select the commit

### Steps to reproduce

select an older commit, e.g. the initial one
hit `F5`
--> an arbitrary commit somewhere in the middle is selected but not scrolled to

### Did this work in previous version of GitExtensions?

3.5.4

gerhardol:
For me it selects for instance the initial revision, but you have to scroll slighly manually to get GE to autoscroll.
The revision is in the revisionGraph but not yet in the gridView when revisions are loaded, so there has to be some further waiting.

RussKie:
This is probably the biggest annoyance in the vNext for me atm...

gerhardol:
> This is probably the biggest annoyance in the vNext for me atm...

Occurs for quite old commits only what I see, the gridView is normally loaded with the recent commits when grid has loaded.
This should not be much different from before I believe, grid is loading in a similar way.

RussKie:
For me happening every time I rebase.

gerhardol:
> For me happening every time I rebase.

That is a slightly different case than just a refresh.
The selected commit is cleared if HEAD changes, when switching repo or rebasing.
https://github.com/gitextensions/gitextensions/blob/master/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs#L966
The 'repair' should select HEAD

gerhardol:
A partial fix that always select something in a private branch.
What is needed is to force loading required Rows when all revisions are loaded.

gerhardol:
Pushed an update that seem to work. (Removed more code than I added.)
However, if no HEAD is available and artificial commits are inserted, selection shifts. This is not a new problem though, but will look if that can be handled here.

gerhardol:
I am preparing a PR for this
https://github.com/gerhardol/gitextensions/pull/new/feature/i9935-gridload-handling

The main change is to check if the grid rows that are expected are loaded when the revisions are loaded as that is running separately.
This should also improve the test failures, the checks were often done before the grid was loaded why the results were so flaky.
Some adjustments still to do it seems.

[...]

PR #9947, opened 2022-04-10
Show multi revision diff also with no HEAD

gerhardol:
## Proposed changes

This affects when there are three or four selected revisions in the graph.

If no HEAD it is not possible to show the range-diff and conflict markers.
However, it is possible to see multi-diff, the diff from the (last) selected to the other revisions (not just the first).

Not important change, just simplifies testing #9935 to see that order is unchanged (and to decrease confusion).

### Before

image

### After

image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusDiffCalculator.cs
+++ b/GitUI/UserControls/FileStatusDiffCalculator.cs
@@ -112,21 +112,21 @@ namespace GitUI
                 summary: TranslatedStrings.DiffWithParent + GetDescriptionForRevision(firstRev.ObjectId),
                 statuses: module.GetDiffFilesWithSubmodulesStatus(firstRev.ObjectId, selectedRev.ObjectId, selectedRev.FirstParentId, cancellationToken)));
 
-            if (!AppSettings.ShowDiffForAllParents || revisions.Count > maxMultiCompare || headId is null)
+            if (!AppSettings.ShowDiffForAllParents || revisions.Count > maxMultiCompare)
             {
                 return fileStatusDescs;
             }
 
             // Get base commit, add as parent if unique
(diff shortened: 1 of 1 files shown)


GitUI/UserControls/FileStatusDiffCalculator.cs
```diff
@@ -225,7 +225,7 @@ DiffBranchStatus GetDiffStatus(GitItemStatus f, bool atoBDiff)
             return fileStatusDescs;
 
             static ObjectId GetRevisionOrHead(GitRevision rev, ObjectId headId)
-                => rev.IsArtificial ? headId : rev.ObjectId;
+                => headId is null ? null : rev.IsArtificial ? headId : rev.ObjectId;
```
mstv, line 228: I would rather keep it. I guess the change is not necessary at first glance. It would leave questions due to the mismatch with the function name.

mstv (commented):
Yes, more consistent UX. Have not run yet.

GitUI/UserControls/FileStatusDiffCalculator.cs
gerhardol: If line 113 is changed, this line (228) or 119-1217 to be has to be changed too.
(119-121 may even be clearer but more code).

RussKie: The method name is now out of sync with the implementation. Are calling it from multiple places? If not, may move the added check to the call site?

gerhardol: Changed this
Also changed 124/127, not strictly necessary but easier to check completeness.

gerhardol:
Any more comments?

RussKie:
:shipit:

PR #9948, opened 2022-04-10
Load grid to selected revisions

gerhardol:
Fixes #935

## Proposed changes

Load grid to selected revisions

The grid loads async separately from the revisions, must be awaited
before selecting rows in the grid or executing tests.
(If the revisions are 'recent' they will normally be loaded, but not if
few revisions were loaded (like in tests) or an early commit were
selected in another repo.

If only some of the revisions were found, show those.
`CheckAndRepairInitialRevision` that tried to handle this before were
not running correctly.

Handle grids without HEAD where artificial commits are inserted.
Previously, the selection "shifted" at refresh.
(Occurs after #9393/0954ee6268a6148c53fe855c9ba9451dd4594345 after 3.5)

Abort 'restore' selection during load if manual selection is performed.

GridLoaded event renamed to RevisionsLoaded to reflect usage.

Edit: added:
If HEAD was not seen for artificial commits or the previously selected revision was filtered,
try to find a parent instead

--

GridView is still not threadsafe, this should be unchanged by the PR. The handling is reasonable i my opinion, some explicit selection and scrolling while revisions or grid i loading could cause strange effects.

--

~~'CheckAndRepairInitialRevision' handling could be resurrected and also be used to insert artificial commits at the first non-filtered parent.
I would prefer to do that separately if requested though.~~

--

With this, RevGrid tests should be more stable.
However, I have seen som instability i still both locally and on AppVeyor.

## Test methodology 

Manual, tests exists

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -182,8 +182,9 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         /// </summary>
         /// <param name="revision">The revision to add.</param>
         /// <param name="types">The graph node flags.</param>
-        /// <param name="insertAsFirst">Insert the (artificial) revision first in the graph.</param>
-        public void Add(GitRevision revision, RevisionNodeFlags types, bool insertAsFirst = false)
+        /// <param name="insertScore">Insert the (artificial) revision before the node with this score.</param>
+        /// <param name="insertRange">Number of scores "reserved" in the list when inserting.</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types, int? insertScore = null, int insertRange = 0)
         {
             // The commits are sorted by the score (not contiuous numbering there may be gaps)
             // This commit will be ordered after existing, _maxScore is a preliminary score
@@ -192,29 +193,40 @@ namespace GitUI.UserControls.RevisionGrid.Graph
             bool updateParents = true;
             if (!_nodeByObjectId.TryGetValue(revision.ObjectId, out RevisionGraphRevision revisionGraphRevision))
             {
-                int score = insertAsFirst
+                // This revision is added from the log, but not seen before. This is probably a root node (new branch)
+                // OR the revisions are not in topo order. If this the case, we deal with it later.
+                int score = _maxScore;
 
-                    // Insert first artificial (WorkTree) before first existing
-                    ? -1
+                if (insertScore is not null && _nodeByObjectId is not null)
+                {
+                    // This revision is to be inserted before a certain node
+                    foreach (var (_, graphRevision) in _nodeByObjectId)
+                    {
+                        if (graphRevision.Score < insertScore)
+                        {
+                            // Lower existing scores to reserve the inserted range
+                            graphRevision.OffsetScore(-insertRange);
+                        }
+                    }
 
-                    // This revision is added from the log, but not seen before. This is probably a root node (new branch)
-                    // OR the revisions are not in topo order. If this the case, we deal with it later.
-                    : _maxScore;
+                    score = insertScore.Value - insertRange;
+                }
+
                 revisionGraphRevision = new RevisionGraphRevision(revision.ObjectId, score);
                 _nodeByObjectId.TryAdd(revision.ObjectId, revisionGraphRevision);
             }
             else
             {
                 // This revision was added earlier, but is now found in the log.
-                if (!insertAsFirst)
+                if (insertScore is null)
                 {
                     // Increase the score to the current maxScore to keep the order intact.
                     revisionGraphRevision.EnsureScoreIsAbove(_maxScore);
                 }
                 else
                 {
-                    // Second artificial (Index), score already set, no parent (HEAD not in grid)
-                    // No parent segment to be added
+                    // Second artificial (Index), score already set
+                    // No parent segment to be added (HEAD not in grid)
                     updateParents = false;
                 }
             }
@@ -231,10 +243,10 @@ namespace GitUI.UserControls.RevisionGrid.Graph
                 {
                     if (!_nodeByObjectId.TryGetValue(parentObjectId, out RevisionGraphRevision parentRevisionGraphRevision))
                     {
-                        int score = insertAsFirst
+                        int score = insertScore is not null
 
-                            // After first inserted (as 0)
-                            ? revisionGraphRevision.Score + 1
+                            // Inserted after current revision
+                            ? revisionGraphRevision.Score + 1 + revision.ParentIds.IndexOf(parentId => parentId == parentObjectId)
 
                             // This parent is not loaded before. Create a new (partial) revision. We will complete the info in the revision
                             // when this revision is loaded from the log.
@@ -265,9 +277,9 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
             if (!updateParents)
             {
-                // The rows may already be cached, invalidate
+                // The rows may already be cached, invalidate and request reload of "some" rows
                 _reorder = true;
-                CacheTo(0, 0);
+                CacheTo(0, 99);
             }
         }
 
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
@@ -42,6 +42,13 @@ namespace GitUI.UserControls.RevisionGrid.Graph
         /// </summary>
         public int Score { get; private set; }
 
+        /// <summary>
+        /// Offset the score for the revision.
+        /// If negative decreasing the score, i.e. increasing priority.
+        /// </summary>
+        /// <param name="offset">The offset to the current score.</param>
+        public void OffsetScore(int offset) => Score += offset;
+
         // This method is called to ensure that the score is higher than a given score.
         // E.g. the score needs to be higher that the score of its children.
         public int EnsureScoreIsAbove(int minimalScore)
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -1,7 +1,10 @@
-﻿using System;
+﻿#nullable enable
+
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Linq;
 using System.Threading.Tasks;
@@ -28,6 +31,7 @@ namespace GitUI.UserControls.RevisionGrid
 
     public sealed partial class RevisionDataGridView : DataGridView
     {
+        private const int BackgroundThreadUpdatePeriod = 25;
         private static readonly AccessibleDataGridViewTextBoxCell _accessibleDataGridViewTextBoxCell = new();
 
         private readonly SolidBrush _alternatingRowBackgroundBrush;
@@ -35,10 +39,13 @@ namespace GitUI.UserControls.RevisionGrid
 
         private readonly BackgroundUpdater _backgroundUpdater;
         private readonly Stopwatch _lastRepaint = Stopwatch.StartNew();
+        private readonly List<ColumnProvider> _columnProviders = new();
 
         internal RevisionGraph _revisionGraph = new();
 
-        private readonly List<ColumnProvider> _columnProviders = new();
+        // Set while loading the revisions and data grid, see also ToBeSelectedObjectIds.
+        private Lazy<IList<int>> _toBeSelectedGraphIndexesCache;
+        private int _loadedToBeSelectedRevisionsCount = 0;
 
         private int _backgroundScrollTo;
         private int _consecutiveScrollMessageCnt = 0; // Is used to detect if a forced repaint is needed.
@@ -50,14 +57,37 @@ namespace GitUI.UserControls.RevisionGrid
         private Font _boldFont;
         private Font _monospaceFont;
 
-        public bool UpdatingVisibleRows { get; private set; }
-        public bool IsBackgroundUpdaterActive => _backgroundUpdater.IsExecuting;
+        /// <summary>
+        /// Indicates whether 'interesting' rows in the data grid is currently being loaded.
+        ///
+        /// The property is set to true when all revisions have been read (including
+        /// artificial commits that are inserted before HEAD) and the data grid rows for
+        /// selected revisions have been loaded.
+        ///
+        /// Revisions are handled directly when git-log delivers them.
+        /// All revisions are always saved.
+        ///
+        /// GridRows are loaded when needed, starting when revisions are loaded.
+        /// If the top revision is selected then only the first rows need to be loaded.
+        /// If one of the bottom rows are selected, all rows must be loaded.
+        /// (https://github.com/gitextensions/gitextensions/wiki/Revision-Graph#revisiongraphrow describes this some.)
+        /// So all grid rows are not forced to be loaded.
+        /// The code now forces loading more rows to be able to select them.
+        ///
+        /// This property is intended to be used by tests.
+        /// </summary>
+        public bool IsDataLoadComplete { get; private set; } = true;
 
+        public bool UpdatingVisibleRows { get; private set; }
+
+        // _toBeSelectedGraphIndexesCache is init in Clear()
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         public RevisionDataGridView()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         {
             InitFonts();
 
-            _backgroundUpdater = new BackgroundUpdater(UpdateVisibleRowRangeInternalAsync, 25);
+            _backgroundUpdater = new BackgroundUpdater(UpdateVisibleRowRangeInternalAsync, BackgroundThreadUpdatePeriod);
 
             InitializeComponent();
             DoubleBuffered = true;
@@ -166,14 +196,6 @@ namespace GitUI.UserControls.RevisionGrid
         // Contains the object Id's that will be selected as soon as all of them have been loaded.
         // The object Id's are in the order in which they were originally selected.
         public IReadOnlyList<ObjectId> ToBeSelectedObjectIds { get; set; } = Array.Empty<ObjectId>();
-
-        private int _loadedToBeSelectedRevisionsCount = 0;
-
-        public bool HasSelection()
-        {
-            return ToBeSelectedObjectIds.Any() || SelectedRows.Count > 0;
-        }
-
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         [Browsable(false)]
         public IReadOnlyList<ObjectId>? SelectedObjectIds
@@ -263,7 +285,7 @@ namespace GitUI.UserControls.RevisionGrid
             return SystemBrushes.Window;
         }
 
-        private void OnCellPainting(object sender, DataGridViewCellPaintingEventArgs e)
+        private void OnCellPainting(object? sender, DataGridViewCellPaintingEventArgs e)
         {
             _lastRepaint.Restart();
 
@@ -300,11 +322,43 @@ namespace GitUI.UserControls.RevisionGrid
         /// </summary>
         /// <param name="revision">The revision to add.</param>
         /// <param name="types">The graph node flags.</param>
-        /// <param name="insertAsFirst">Insert the (artificial) revision first in the graph.</param>
-        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool insertAsFirst = false)
+        /// <param name="insertWithMatch">Insert the (artificial) revision with the first match in headParents or first if no match found (or headParents is null).</param>
+        /// <param name="insertRange">Number of scores "reserved" in the list when inserting.</param>
+        /// <param name="parents">Parent ids for the revision to find (and insert before).</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool insertWithMatch = false, int insertRange = 0, IEnumerable<ObjectId>? parents = null)
         {
-            _revisionGraph.Add(revision, types, insertAsFirst);
+            // Where to insert the revision, null is last
+            int? insertScore = null;
+            if (insertWithMatch)
+            {
+                if (_loadedToBeSelectedRevisionsCount == 0
+                    && ToBeSelectedObjectIds.Count == 0
+                    && SelectedRows?.Count is > 0)
+                {
+                    // (GraphIndex) selection in grid was 'premature'
+                    ToBeSelectedObjectIds = SelectedObjectIds ?? Array.Empty<ObjectId>();
+                    _loadedToBeSelectedRevisionsCount = ToBeSelectedObjectIds.Count;
+                    ResetGraphIndices();
+                }
 
+                // Insert first by default (if HEAD not found)
+                // Actual value is ignored if insertRange is 0
+                // (Used when child (like WorkTree) is already inserted when adding parent (like Index))
+                insertScore = -1;
+                if (insertRange > 0 && parents is not null)
+                {
+                    foreach (var parentId in parents)
+                    {
+                        if (_revisionGraph.TryGetNode(parentId, out RevisionGraphRevision parentRev))
+                        {
+                            insertScore = parentRev.Score;
+                            break;
+                        }
+                    }
+                }
+            }
+
+            _revisionGraph.Add(revision, types, insertScore, insertRange);
             if (ToBeSelectedObjectIds.Contains(revision.ObjectId))
             {
                 ++_loadedToBeSelectedRevisionsCount;
@@ -317,10 +371,10 @@ namespace GitUI.UserControls.RevisionGrid
         {
             _backgroundScrollTo = 0;
 
-            // Set rowcount to 0 first, to ensure it is not possible to select or redraw, since we are about te delete the data
+            // Set rowcount to 0 first, to ensure it is not possible to select or redraw, since we are about to delete the data
             SetRowCount(0);
             _revisionGraph.Clear();
-            _loadedToBeSelectedRevisionsCount = 0;
+            ClearToBeSelected();
 
             // The graphdata is stored in one of the columnproviders, clear this last
             foreach (var columnProvider in _columnProviders)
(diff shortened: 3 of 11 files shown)


mstv (commented):
Looks plausible.
Though I could not find the previous handling which selected one of the parents if the previous selection was deleted. (Not very important to me.)

gerhardol:
> Though I could not find the previous handling which selected one of the parents if the previous selection was deleted. (Not very important to me.)

That icould be `CheckAndRepairInitialRevision` (but it should be empty if the revision was found).
GridView is not threadsafe, so there may be something else going on.
It is not for sure that this part of the problem is solved.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,8 +340,70 @@ public void Clear()
...
+
+                        if (GetGraphIndexes().Count > 0)
+                        {
+                            EnsureRowVisible(GetGraphIndexes()[0]);
+                        }
+                   }
+
+                    IsLoadingGrid = false;
+                })
+                    .FileAndForget();
```
RussKie, line 400: nit: indent

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -51,7 +51,8 @@ public sealed partial class RevisionDataGridView : DataGridView
         private Font _monospaceFont;
 
         public bool UpdatingVisibleRows { get; private set; }
-        public bool IsBackgroundUpdaterActive => _backgroundUpdater.IsExecuting;
+        public bool IsGridUpdating => _backgroundUpdater.IsExecuting || IsLoadingGrid;
+        public bool IsLoadingGrid { get; set; } = false;
```
RussKie, line 55: I'm not finding the names of these two properties meaningful or consistent. In the context of this type we seem to use "revisions" not "grid" term.

Also this property feels like an implementation detail. In Windows Forms we use `BeginXyz`/`EndXyz` methods pair for such situations.

RussKie (commented):
I don't have access to a computer to review in depth. Just few comments.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: There is a difference to Revisions (the "Git" data) and Grid (displayed rows). Grid is loaded after Revisions, not the same, so the name should differ.

Both of these could be replaced with a notification when Grid is loaded including the last selected row, but it just got complex when I tried this.

gerhardol: Note that these methods are only used by tests

gerhardol:
> > Though I could not find the previous handling which selected one of the parents if the previous selection was deleted. (Not very important to me.)
> 
> That icould be `CheckAndRepairInitialRevision` (but it should be empty if the revision was found). GridView is not threadsafe, so there may be something else going on. It is not for sure that this part of the problem is solved.

A further explanation: Before, `CheckAndRepairInitialRevision` checked if `ToBeSelectedObjectIds` was cleared when the revisions were loaded, but that variable is cleared when the grid rows including *all* selected commits have loaded. It was therefore possible that `CheckAndRepairInitialRevision` tried to fix something that was not yet updated in the grid.
However, the probability for this to occur in repos with GE size if the commit was relative recent should be low (from what I have seen in the debugger grid rows are loaded quite rapidly.) It is very seldom I have seen "incorrect commit selected after refresh" lately (even it seem like ussKie see it more often, really the same type of repos as before?)

If "select something if no longer available" is to be resurrected, the missing commits should be checked and updated with `_loadedToBeSelectedRevisionsCount` and `GetGraphIndexes()`. 
But I very much prefer to merge this first, some usage is needed.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -961,18 +953,18 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
                     ObjectId? newCurrentCheckout = headRef?.ObjectId ?? capturedModule.GetCurrentCheckout();
 
-                    // If the current checkout changed, don't get the currently selected rows, select the
-                    // new current checkout instead.
+                    // If the current checkout (HEAD) is changed, don't get the currently selected rows,
+                    // select the new current checkout instead.
                     if (newCurrentCheckout != CurrentCheckout)
                     {
-                        selectedObjectIds = null;
+                        currentlySelectedObjectIds = null;
```
gerhardol, line 960: Maybe try to keep selection also when refreshing?

```suggestion
if (currentlySelectedObjectIds is not null && currentlySelectedObjectIds.Count == 1 && currentlySelectedObjectIds[0] == CurrentCheckout)
{
                        currentlySelectedObjectIds = List<ObjectId>{ newCurrentCheckout  };
}
```

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: We should avoid having "test only" API unless those are exposed via `TestAccessor`.

gerhardol: IsGridUpdating  can be moved to test accessor, but how would you handle `IsLoadingGrid`?

RussKie: I pushed an alternative, please let me know what you think of it.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -192,29 +193,40 @@ public void Add(GitRevision revision, RevisionNodeFlags types, bool insertAsFirs
...
+                {
+                    // This revision is to be inserted before a certain node
+                    foreach (var kvp in _nodeByObjectId)
+                    {
+                        if (kvp.Value.Score < insertScore)
+                        {
+                            // Insert reserved scores
+                            kvp.Value.OffsetScore(-insertRange);
+                        }
+                    }
```
RussKie, line 210: You can use a deconstructor, which makes it easier to read. Though you'd have to give it a better variable name 😉:
```suggestion
                    foreach (var (_, revisionGraphRevision1) in _nodeByObjectId)
                    {
                        if (revisionGraphRevision1.Score < insertScore)
                        {
                            // Insert reserved scores
                            revisionGraphRevision1.OffsetScore(-insertRange);
                        }
                    }
```

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
```diff
@@ -42,6 +42,8 @@ public void ApplyFlags(RevisionNodeFlags types)
         /// </summary>
         public int Score { get; private set; }
 
+        public void OffsetScore(int offset) => Score += offset;
```
RussKie, line 45: `Score` has a decent docs, it would be great to have similar docs for this method too.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,12 +366,115 @@ public void Clear()
             Invalidate(invalidateChildren: true);
         }
 
+        public void EndSelectionAtLoad()
```
RussKie, line 369: It would be good to have docs for this method explaining what it does and why it is necessary.
Ditto for other public API.

RussKie: Maybe call this `ClearPreviousSelectionCache`?

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,12 +366,115 @@ public void Clear()
...
+                    }
+                    else
+                    {
+                        // Rows already selected once, reselect and refresh
+                        SelectRowsIfReady(RowCount);
+
+                        if (GetGraphIndexes().Count > 0)
+                        {
+                            EnsureRowVisible(GetGraphIndexes()[0]);
+                        }
```
RussKie, line 448: This is not good, we're calling `GetGraphIndexes()` several times. Please refactor to invoke the method and assign the result to a local variable.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,12 +366,115 @@ public void Clear()
...
+                    if (RowCount - 1 < scrollTo)
+                    {
+                        // Wait for the background thread to load all rows in the grid
+                        while (RowCount - 1 < scrollTo)
+                        {
+                            int maxScroll = Math.Min(RowCount - 1, scrollTo);
+                            EnsureRowVisible(maxScroll);
+
+                            UpdateVisibleRowRange();
+                            await Task.Delay(25);
```
RussKie, line 437: Why do we need this, and where did "25" come from?

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,12 +366,115 @@ public void Clear()
...
+                FirstDisplayedScrollingRowIndex = row;
+            }
+        }
+
+        public ObjectId? GetFirstNotSelectedObjectId()
+        {
+            if (_loadedToBeSelectedRevisionsCount > 0 || ToBeSelectedObjectIds.Count == 0)
+            {
+                // At least one of the revisions was selected
+                return null;
```
RussKie, line 391: This looks kind of contrary to the method name - i.e., the name states the result is the "first not selected revision id". `null` in this case would mean that **all** revisions are selected. But according to the comment, `null` indicates that there is an existing selection.... 🤯

RussKie: How about something like this?
```diff
 
-        public ObjectId? GetFirstNotSelectedObjectId()
-        {
-            if (_loadedToBeSelectedRevisionsCount > 0 || ToBeSelectedObjectIds.Count == 0)
-            {
-                // At least one of the revisions was selected
-                return null;
-            }
-
-            return ToBeSelectedObjectIds[0];
-        }
+        public bool HasExistingSelection => _loadedToBeSelectedRevisionsCount > 0 || ToBeSelectedObjectIds.Count == 0;
 
```

```diff
                     // All revisions are loaded (but maybe not yet the grid)
-                    if (_gridView.GetFirstNotSelectedObjectId() is ObjectId notSelectedId)
+                    if (!_gridView.HasExistingSelection)
                     {
+                        ObjectId notSelectedId = _gridView.ToBeSelectedObjectIds.First();
+
                         IEnumerable<ObjectId> parents = null;
```

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -333,12 +366,115 @@ public void Clear()
...
+            }
+
+            return ToBeSelectedObjectIds[0];
+        }
+
+        public void ToBeSelectFirstFoundParent(IEnumerable<ObjectId> headParents)
```
RussKie, line 397: This method name completely lost me...

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
gerhardol: graphRev

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRevision.cs
gerhardol: added

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: `ClearToBeSelected()`

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: GetGraphIndexes() is doing the caching.
The information may be cleared as well so info may not be instaciated on more than line 426 and 445.
This could have been set as a Lazy variable too, but as there is no method to Clear a Lazy variable, the init would have to be done in several locations.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: Added constant `backgroundPeriod` and updated the comment.
There is no point in polling more often than that.
(changed in tests too.)

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: SetToBeSelectedFromParents

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: Changed this slightly.
Tho match the API, another check had to be added, to make sure the added check was not misused, GetGraphIndexes() needed a Assert too.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -860,17 +865,5 @@ private void InitFonts()
 
         private static Color getGrayTextColor(float degreeOfGrayness = 1f) =>
             ColorHelper.GetGrayTextColor(textColorName: KnownColor.ControlText, degreeOfGrayness);
-
-        internal TestAccessor GetTestAccessor() => new(this);
-
-        internal readonly struct TestAccessor
```
gerhardol: @RussKie As IsDataLoadComplete was bypassing TestAccessor even if it is test-only, I assume you no longer requests that this property should be a TestAccessor as you requested in a previous review.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: This still makes me very uneasy. One of the aspects that worries me is threading - the field can be accessed by multiple threads, and could lead to NRE. In c427e589ceefe978927e215c037d0ab3852ecbab I proposed to make the field non-nullable, so we don't risk NREs, though we still have a risk of a race condition. By looking at the other existing fields/properties in use - the risk is probably not significant (though still real).

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -567,16 +572,18 @@ private void SelectRowsIfReady(int rowCount)
                 return;
             }
 
+            CalculateGraphIndices();
```
gerhardol, line 572: Always calculating the graph indices has a performance impact: This is done every time SetRowCount() is called. (The global variable is not really needed with this setup, just FoundAnyToBeSelected to be changed or renamed).
(Not executing at every row but every "chunk of grid rows loaded").
Revisions are loaded before the grid rows.

So I want to revert to the previous handling.

gerhardol (commented):
In addition to the efficiency, the latest commits added some NRE too, at least when no head is cheecked out and when no row at all could be selected after a refresh.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: > So I want to revert to the previous handling.

Sure thing

gerhardol: I changed the handling to use Lazy<>, to not trigger similar concerns in the future.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -50,14 +55,20 @@ public sealed partial class RevisionDataGridView : DataGridView
         private Font _boldFont;
         private Font _monospaceFont;
 
+        /// <summary>
+        ///  Indicates whether the data grid is currently being loaded.
+        ///  Set to false when the selected revisions are loaded in the grid
+        ///  (or all rows loaded if not all revisions were found).
```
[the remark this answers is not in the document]
gerhardol: Revisions are handled directly when git-log delivers them.
All revisions are always saved.

If all revisions to-be-selected are not found the revisions are searched to find a parent match.

GridRows are loaded when needed, started when revisions are loaded. If the top revision is selected then only the first rows need to be loaded. If one of the bottom rows are selected, all rows must be loaded. This is normally not done until some time after revisions are loaded.
(https://github.com/gitextensions/gitextensions/wiki/Revision-Graph#revisiongraphrow describes this some.)
So all grid rows are not forced to be loaded.
The code now forces loading more rows to be able to select them.

The assumption that thegrid rows were loaded when the revisions were loaded is the major problem in the current implementation.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -28,6 +28,7 @@ public enum RevisionNodeFlags
 
     public sealed partial class RevisionDataGridView : DataGridView
     {
+        private const int backgroundPeriod = 25;
```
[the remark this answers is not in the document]
gerhardol: BackgroundUpdateThreadPeriod?

gerhardol:
The refactorings lately caused some test instability that I missed in my fixups. Should be better again, but all test instability is probably not fixed.
Plan to squash and rebase tomorrow to simplify reviews.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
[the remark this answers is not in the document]
RussKie: Could you please add this comment to the code?

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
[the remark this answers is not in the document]
RussKie: `BackgroundThreadUpdatePeriod` or better yet ``BackgroundThreadUpdatePeriodMilliseconds`

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -395,36 +556,66 @@ private void SetRowCount(int count)
...
+        }
+
+        /// <summary>
+        /// Get the revision graph row indexes for the ToBeSelectedObjectIds.
+        /// (In filtering situations, all may no longer be in the grid).
+        /// </summary>
+        private IList<int> CalculateGraphIndices()
+        {
+            Debug.Assert(_loadedToBeSelectedRevisionsCount == ToBeSelectedObjectIds.Count,
+                "GetGraphIndexes() was called before all expected revisions were loaded.");
```
RussKie, line 574: Is this outdated? Should this be
```suggestion
                $"{nameof(CalculateGraphIndices)}() was called before all expected revisions were loaded.");
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1285,13 +1262,51 @@ void OnRevisionReadCompleted()
                 // This has to happen on the UI thread
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
+                    IEnumerable<ObjectId> headParents = null;
```
RussKie, line 1265: The whole handling of `headParents` makes me concerned. `TryGetParents` executes `rev-list` and the yields elements while those are iterated over. The issue here is that we are iterating over this collection multiple times over and over again, and we're performing multiple allocations (i.e., essentially stressing the GC).

First time in `AddArtificialRevisions() -> _gridView.Add(..., headParents)`. 
Then we are calling `headParents.ToList().IndexOf(...)` or `headParents.Skip(...).ToList()`.

...And we're churning through the memory in https://github.com/gitextensions/gitextensions/pull/9948#discussion_r851716047.

We should consider reading `headParents` collection into some kind of a dictionary (where the lookup is O(1)), and don't reallocating past that point.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -231,10 +243,10 @@ public void Add(GitRevision revision, RevisionNodeFlags types, bool insertAsFirs
                 {
                     if (!_nodeByObjectId.TryGetValue(parentObjectId, out RevisionGraphRevision parentRevisionGraphRevision))
                     {
-                        int score = insertAsFirst
+                        int score = insertScore is not null
 
-                            // After first inserted (as 0)
-                            ? revisionGraphRevision.Score + 1
+                            // Inserted after current revision
+                            ? revisionGraphRevision.Score + 1 + revision.ParentIds.ToList().IndexOf(parentObjectId)
```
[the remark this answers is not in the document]
RussKie: This removes re-alloc, but doesn't solve the multiple rescans. We should consider changing the underlying data structure to provide O(1) lookup.
```suggestion
                            ? revisionGraphRevision.Score + 1 + revision.ParentIds.IndexOf(parentId => parentId == parentObjectId)
```

mstv:
The new merge commit is not automatically selected anymore:

image

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: This is executed only if no revision that were to be selected was found, for instance when filtering.
This tries to find the closest parent where to insert artificial and what to select. (If possible, the data is shared.)
headParents/parents could be long list.

parents must be traversed in order in SetToBeSelectedFromParents so a dictionary would not be of help.

However, the implementation has a very minor impact on the performance, should not be many ms.
If you have concern for performance, it is the git-rev-list call that is the issue, it adds 100 ms. I cannot see anything that requires time in this method (some of the datagrid manipulation requires ms though.)
The solution should optimize for a match in one of the first in the list.
It would be a reasonable limitation to just request maximum 10? 100? revisions from Git to limit memory usage.

Note that the previous implementation would fire the git-rev-list in more situations without any reason and risk messing up the selection of what we actually want.
So this is an improvement regardless.
(To insert artificial just before the first visible HEAD parent is nice but not essential, similiar to select something when filtering. This could be dropped even if I would be annoyed as this was a pain to get efficient in the sense to never request more data than required.)

git-rev-list should be cached (except for artificial), that would be 100 ms improvement in the repeat case.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: changing

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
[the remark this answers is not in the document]
gerhardol: Added a variant of the text above

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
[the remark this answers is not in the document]
gerhardol: BackgroundThreadUpdatePeriod , units should normally not be in the variable name.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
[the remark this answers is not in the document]
gerhardol: A commit very seldom has more than one parent, the scenario to optimize for is one parent. O(n^2) does not matter at all.
Further, it occurs for one commit only if HEAD is filtered and artificial commits are shown.
Changing to the suggestion.

gerhardol:
> The new merge commit is not automatically selected anymore:

The rules were changed to just switch from old HEAD to new HEAD and try to restore everything else.
Previously the new HEAD was selected if HEAD was selected - normally, the selection was 'fixed' before all rows were loaded.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
gerhardol: What exactly?
DataGrid is not thread safe, but the same thread basically updates the data for a certain object, I did not find any potential race conditions.

mstv:
> > The new merge commit is not automatically selected anymore:
> 
> The rules were changed to just switch from old HEAD to new HEAD and try to restore everything else. Previously the new HEAD was selected if HEAD was selected - normally, the selection was 'fixed' before all rows were loaded.

It will take time to get used to this new behavior restoring the selection of the merged or cherry-picked commit - if ever although the new commit is only a `Ctrl+<` press away.
It might ease merging several branches (as I do for my custom test builds, which is a rather exotic usecase I guess).

---

The graph seems to need another trigger for refresh - at least when filtering for a file:

image

Then the graph is drawn if scrolled by one commit or by selecting one or a few other commits.

gerhardol:
> It will take time to get used to this new behavior

I am not opposed to change back to the old behavior to select the new HEAD if HEAD changes.
The selection will be lost in some other situations though.
I find the new behavior to be more persistent, but it is a change.

> The graph seems to need another trigger for refresh - at least when filtering for a file:

I have seen that occasionally both before and after this PR related to resolve conflicts etc.
Cannot reproduce.
(The change to show the rebase form also when rebasing interactively without options solved those occurrences, it feels like what I see is related.)

mstv:
> > The graph seems to need another trigger for refresh - at least when filtering for a file:
> 
> I have seen that occasionally both before and after this PR related to resolve conflicts etc.

I know this. But this PR brings a regression.

> Cannot reproduce.

Filter for `RepoObjectsTree.Designer.cs` (modified in second last commit f3e5c63 of `master`).

With `master` (It really is the HEAD. The branch name in the title is wrong because it is updated only on rebuild from scratch.):

image

With PR:

image

mstv:
> > It will take time to get used to this new behavior restoring the selection of the merged or cherry-picked commit - if ever although the new commit is only a `Ctrl+<` press away.
> > It might ease merging several branches (as I do for my custom test builds, which is a rather exotic usecase I guess).
> 
> I am not opposed to change back to the old behavior to select the new HEAD if HEAD changes. The selection will be lost in some other situations though. I find the new behavior to be more persistent, but it is a change.

Both variants have their pros and contras. Thoughts, @gitextensions/git-extensions-team and watchers?

gerhardol:
> I know this. But this PR brings a regression.

I believe I see it as often as before.

> > Cannot reproduce.
> 
> Filter for `RepoObjectsTree.Designer.cs` (modified in second last commit [f3e5c63](https://github.com/gitextensions/gitextensions/commit/f3e5c63e89f62cca00d92102ae3e68aaad1894fc) of `master`).

Still cannot reproduce, tested 5 times each with 'f' in the second revision respectively selecting HEAD and activating in Advanced filter.
If it is related to inserting artificial commits when HEAD is filtered it is something to start with. GridRow is not threadsafe, there can be race conditions.

One alternative implementation is to always run rev-parse when the grid is filtered and to insert artificial and for selection replacement (so one or two Git commands). This could delay showing revisions and I expect it to be wasted in many situations (as it is likely the file is in the HEAD commit).

mstv:
100% reproducible with the commit and `f` at the file mentioned above - also with a fresh clone of GE. Graph disappears again on refresh (`F5`).

- Git Extensions 33.33.33
- Build 6bb1f4ae46fe81b90dd8759bb015c0d5c57d89f5
- Git 2.35.2.windows.1
- .NET 6.0.3
- DPI 96dpi (no scaling)

gerhardol:
> 100% reproducible

Not for me, but I got it like 1/5 when testing exactly like described.
This is a timing issue, the handling for refreshing is not really changed by this PR. It just happened to occur more often.

Seem to work when explicitly refreshing.
I considered to get the HEAD rev-list always when filtering to avoid inserting in most situations. However insertion is still needed in some situations so this would just add code.
When HEAD is not set, the artificial should be inserted before first commit.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1285,13 +1265,54 @@ void OnRevisionReadCompleted()
                 // This has to happen on the UI thread
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
```
mstv, line 1267: This comment is valid.
`await this.SwitchToMainThreadAsync();` must be done before both `AddArtificialRevisions` and `_gridView.SetToBeSelectedFromParents` at least because they are accessing the `_gridView` UI.

mstv (commented):
Uff, this must have been a lot of puzzling.
Works well - as [you have designed it](https://github.com/gitextensions/gitextensions/pull/9948#issuecomment-1113926075).

In the testcase `ShowLatestStash_starting_disabled_should_filter_as_expected`, none of the expected revisions are loaded yet.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -50,14 +57,34 @@ public sealed partial class RevisionDataGridView : DataGridView
         private Font _boldFont;
         private Font _monospaceFont;
 
+        /// <summary>
+        /// Indicates whether 'interesting' rows in the data grid is currently being loaded.
+        ///
+        /// The property is set to false when all revisions have been read (including
```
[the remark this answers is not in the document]
gerhardol: updated (as well as spelling corrections)

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -300,11 +322,43 @@ private void OnCellPainting(object sender, DataGridViewCellPaintingEventArgs e)
...
+        /// <param name="parents">Parent ids for the revision to find (and insert before).</param>
+        public void Add(GitRevision revision, RevisionNodeFlags types = RevisionNodeFlags.None, bool insertWithMatch = false, int insertRange = 0, IEnumerable<ObjectId>? parents = null)
+        {
+            // Where to insert the revision, null is last
+            int? insertScore = null;
+            if (insertWithMatch)
+            {
+                if (_loadedToBeSelectedRevisionsCount == 0
+                    && ToBeSelectedObjectIds.Count == 0
+                    && (SelectedRows?.Count ?? 0) > 0)
```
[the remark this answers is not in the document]
gerhardol: ok

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -114,6 +141,9 @@ public RevisionDataGridView()
                     .FileAndForget();
             };
 
+            // Suppress CS8618, init in Clear().
+            _toBeSelectedGraphIndexesCache = new(() => CalculateGraphIndices());
```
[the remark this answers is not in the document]
gerhardol: Then cs8618 must be suppressed, the reason for the call
But if the workaround raises questions, explicit suppressing is maybe better

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Removing this comment, not valid in general.
_gridView.Add() is called from worker thread in OnRevisionRead(), same here.

gerhardol:
> Uff, this must have been a lot of puzzling.

Yes, this is tricky
 
> In the testcase `ShowLatestStash_starting_disabled_should_filter_as_expected`, none of the expected revisions are loaded yet.

Adding `WaitForRevisionsToBeLoaded(form);` in a few tests
I though believe that these tests have been stable though

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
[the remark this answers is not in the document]
mstv: I might have got it, if the correct function name `ResetGraphIndices` was mentioned in the comment instead of `Clear`.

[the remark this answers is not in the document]
gerhardol: It is Clear()->ClearToBeSelected()->ResetGraphIndices()

gerhardol:
Refresh in an empty repo asserted in this PR, updated.
Also handled in #9985 (no assert there even if the problem is potentially the same).

mstv (commented):
+2 technically
+0 behavior change: merge commit not selected

gerhardol:
> +0 behavior change: merge commit not selected

I can change back to the current behavior to select HEAD if HEAD changes if that is preferred.
But that will keep the current behavior to select HEAD also when for instance pulling.
No response from others so you have the voting power...

Rebased and squashed on master to resolve changes, left `tmp/i9935-gridload-handling` behind

gerhardol:
Can we get a decision for the PR?
It will not fix the test instability but should handle "random selection after a refresh".
Open issues:
* What to select if HEAD changes? Always new HEAD (as in master) or currently selected (as current in this PR.
* When inserting artificial, should Index commit rewrite the parent? (currently Index has no connection. The diff is a little weird though.)

mstv:
> Can we get a decision for the PR? It will not fix the test instability but should handle "random selection after a refresh". Open issues:
>
>     * What to select if HEAD changes? Always new HEAD (as in master) or currently selected (as current in this PR.

IMO to keep / to restore the selection on refresh (`F5`) is a must.

(a) Selecting the new checked out commit after any git operation feels natural. At least, I am used to it - as many users, too, I guess. Some would complain / need an explanation.
(b) Though keeping the selection of the merged commit / branch, might ease merging multiple branches into a test version.
     This might apply to pull, too, in order to see what's new. I am not sure because I use to `fetch` but do not `pull`.

I don't have a strong opinion. That's why I voted "+0" (but not "-") for the currently proposed implementation in the PR - being (b).

@RussKie, @pmiossec and other watchers, what do you think?


>     * When inserting artificial, should Index commit rewrite the parent? (currently Index has no connection. The diff is a little weird though.)

IMO no IIUIC. The artificial commits should be connected to HEAD only. This emphasizes that commits have been filtered out.

RussKie:
> IMO to keep / to restore the selection on refresh (`F5`) is a must.
> 
> (a) Selecting the new checked out commit after any git operation feels natural. At least, I am used to it - as many users, too, I guess. Some would complain / need an explanation. (b) Though keeping the selection of the merged commit / branch, might ease merging multiple branches into a test version. This might apply to pull, too, in order to see what's new. I am not sure because I use to `fetch` but do not `pull`.

Agree with all of these. And I always fetch too, so no real opinion or thoughts for the pull behaviour.

> The artificial commits should be connected to HEAD only.

Agree as well.

RussKie:
Overall it LGTM, though the tests fail.

gerhardol:
> Overall it LGTM, though the tests fail.

The current test result is OK.

This PR fixes some instability in some tests for asserting the results.
Fixing test instability was why I first looked at this. The test failure is a major pain, I do not trust the result and assume it is instability.
However, test timing out is about the same as before. For some time I thought it was better, probably the same though. 
The core problem for timing out could be similar to #10001 (that has another test problem).
This improves #9223 (that #10001 tries to address) too.

mstv (approved):
Thank you

gerhardol:
@RussKie Plan to merge this tomorrow

RussKie:
Sounds good 👍

PR #9951, opened 2022-04-11
Reduce Linq overhead in ArgumentBuilder

lechu445:
Reduces Linq overhead underneath `.Where(...)` by replacing it with simple `if` statement.

This is a performance optimization. I would also say it is also code readability improvement, but it's matter of taste, I think.

### DIFF
--- a/GitExtUtils/ArgumentBuilder.cs
+++ b/GitExtUtils/ArgumentBuilder.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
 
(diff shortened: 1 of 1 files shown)


lechu445:
the failed test in pipeline is unrelated IMHO

GitExtUtils/ArgumentBuilder.cs
```diff
@@ -69,10 +68,12 @@ public void Add(string? s)
         /// <param name="args">The arguments to add to this builder.</param>
         public void AddRange(IEnumerable<string> args)
         {
-            args = args.Where(a => !string.IsNullOrEmpty(a));
             foreach (string s in args)
             {
-                Add(s);
+                if (!string.IsNullOrEmpty(s))
```
mstv, line 73: This check is already done in `Add`.
Hence, it should be removed here at all. And the handling of empty arguments by `AddRange` should be verified by means of a new testcase in `ArgumentBuilderTests.cs`.

mstv (commented):
Please add your sign-off to the `contributors.txt` file and fill-in the PR template.

PR #9956, opened 2022-04-13
git-fetch --jobs==0 if no configuration

### LINKED ISSUE #9937: Use parallel fetch to improve performance when multiple remotes are used
### Feature description

`git-fetch` has an `--jobs` option that performs fetches from multiple remote concurrently. When you have multiple remotes, this makes fetching much faster.

```
$ time git fetch --all
Fetching A
Fetching B
Fetching C
Fetching D

real    0m11.742s
user    0m0.046s
sys     0m0.046s

$ time git fetch --all --jobs=8
Fetching A
Fetching B
Fetching C
Fetching D

real    0m3.232s
user    0m0.030s
sys     0m0.108s
```

That timing matches what I see in Git Extensions:

image

I've used `--jobs=8` with a CLI alias locally for over a year and never found an issue with it. I wish it was the default behaviour.

It would be nice if Git Extensions passed `--jobs` by default.

https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt--j

There may be other commands that could benefit from the same treatment.

An open question would be the number of jobs to pass. I suspect this is primarily an I/O-bound operation, I believe it would be ok to keep things simple and just pass a number like 8 here. If it were CPU-bound, then basing it on CPU count might make sense.

### Environment

- Git Extensions 3.5.4.12724
- Git 2.34.1.windows.1
- .NET Framework 4.8.4420.0
- DPI 144dpi (150% scaling)

gerhardol:
Fixes #9937

## Proposed changes

Override the Git default of one parallel job.

If the user has set fetch.parallel or submodule.fetchJobs the
the user config value is set (or git default if just one is set).

Config check adds 4ms the first check for me and <1ms the second (when the value is cached).

## Test methodology 

Manual
I guess this could be tested, someone is welcome to contribute.

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1522,9 +1522,14 @@ namespace GitCommands
                 }
             }
 
+            bool remoteJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("fetch.parallel"));
+            bool submoduleJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("submodule.fetchJobs"));
+
             // TODO return ArgumentBuilder and add special case ArgumentBuilder.Add(ArgumentBuilder childBuilder)
             return new ArgumentBuilder
             {
+                { remoteJobs && string.IsNullOrWhiteSpace(branchArguments), "--multiple" },
+                { remoteJobs && submoduleJobs, "--jobs=0" },
                 remote.ToPosixPath()?.Trim().Quote(),
                 branchArguments,
                 { fetchTags == true, "--tags" },
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1522,14 +1522,13 @@ namespace GitCommands
                 }
             }
 
-            bool remoteJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("fetch.parallel"));
-            bool submoduleJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("submodule.fetchJobs"));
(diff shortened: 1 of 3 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -1522,9 +1522,14 @@ private ArgumentString GetFetchArgs(string? remote, string? remoteBranch, string
                 }
             }
 
+            bool remoteJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("fetch.parallel"));
+            bool submoduleJobs = string.IsNullOrWhiteSpace(EffectiveConfigFile.GetValue("submodule.fetchJobs"));
+
             // TODO return ArgumentBuilder and add special case ArgumentBuilder.Add(ArgumentBuilder childBuilder)
             return new ArgumentBuilder
             {
+                { remoteJobs && string.IsNullOrWhiteSpace(branchArguments), "--multiple" },
```
mstv, line 1531: The Git documentation mentions this option. Though it seems to be implicit, i.e. not necessary.

mstv (commented):
LGTM
Although I would have chosen the more active approach: I would have run the regarding `git config` command if the value has not been configured yet.

RussKie (approved):

GitCommands/Git/GitModule.cs
gerhardol: Removed --multiple

gerhardol:
> LGTM Although I would have chosen the more active approach: I would have run the regarding `git config` command if the value has not been configured yet.

You mean runnig git-config? That would be a different way using config from the rest of GE and add some tenths (50?) ms for at least every repo.

Also fixed the tests and added tests for config options.

mstv (approved):

PR #9963, opened 2022-04-17
Migrates FabricBot automation to "Config-as-Code"

RussKie:
The file is a dump of the automation rules that are currently active and doesn't contain any additional changes.
Similar to https://github.com/dotnet/winforms/pull/6538.

This still means that only MS employees may be able to use modify the config via the Fabric Bot portal, however this provides an opportunity for the wider community participation, e.g., updating the milestone or labels directly via the code.

----

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,882 @@
+[
+  {
+    "taskType": "trigger",
+    "capabilityId": "IssueResponder",
+    "subCapability": "PullRequestReviewResponder",
+    "version": "1.0",
+    "config": {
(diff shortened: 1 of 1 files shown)


.github/fabricbot.json
```diff
@@ -0,0 +1,882 @@
...
+        "pull_request",
+        "issues",
+        "project_card"
+      ],
+      "taskName": "Apply milestone 'vNext' to merged PRs on master branch",
+      "actions": [
+        {
+          "name": "addMilestone",
+          "parameters": {
+            "milestoneName": "vNext"
```
RussKie, line 555: Here's the milestone label

PR #9723, opened 2021-11-13
Select and filter for multiple branches and tags from repo objects tree in left panel

h0lg:
This PR enables **de/selecting multiple refs (tags, local or remote branches)** from the **repo objects tree in the left sidebar** using `Ctrl` + `left click` and applying them to the **revision grid branch filter** via `Filter for selected` from the refs' context menu.
`Ctrl` + `Shift` + `left click` toggles descendant tree nodes multi-selected.

It relates to the following feature requests and discussions while not implementing any of them exactly as requested and not implementing all of the features requested in them. Consider this an alternative approach that I hope improves the UX of multi-refs operations nonetheless.

It answers #9495, resolves #9472, resolves #3114, resolves #491, resolves #3546, resolves #3165, and represents a selection alternative for #9412.

PR #9975, opened 2022-04-22
Left panel/overview node inheritance

h0lg:
This PR doesn't change functionality, but merely **moves some types** previously nested in `RepoObjectsTree` into the `GitUI.BranchTreePanel` namespace to make developers' lives easier by

1. enable creating meaningful class diagrams with those types
2. enable opening the code editor instead of the Forms designer by default on double-click from the VS Solution Explorer 

Additionally, it **renames the code files to reflect the move** and **adds a class diagram** overviewing the inheritance and relationships between nodes and trees in the left panel that I used for the context menu rewrite in #9723.

Pick the changes you want, if any. No hard feelings if you don't.

Squash-merge this if you like, but be aware that you'll **lose a lot of file history** because of the file renames. I'd recommend a **rebase-merge**.

----


h0lg:
Added `sealed` type modifiers in the last commit to signal the end of the inheritance hierarchy. Just for some mental load relief when working on those types :)

h0lg:
This last one you might want to cherry-pick onto the master, it's a fix-up for #9723

RussKie:
Thank you

PR #9982, opened 2022-05-02
Reset pathfilter by disabling checkbox

gerhardol:
## Proposed changes

Behave similar to Reset all by disabling the checkbox when the filter is empty.
This makes it simpler to toggle the filter.

### After

image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -400,7 +400,10 @@ namespace GitUI
         public void SetAndApplyPathFilter(string filter)
         {
             _filterInfo.ByPathFilter = !string.IsNullOrWhiteSpace(filter);
-            _filterInfo.PathFilter = filter;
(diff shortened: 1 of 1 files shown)


mstv:
IIUIC, you should have mentioned that this PR is about: `Reset path filter` shall not empty the `Path filter` TextBox in order to be able to quickly restore this filter - as it is already possible after `Reset revision filters`.

image

The behavior inside the `Advanced filter` dialog seems to be not affected.
Have I missed anything?

RussKie (approved):

gerhardol:
> Have I missed anything?

Good summary.
I expected the short description and the screenshot to be enough.

It is actually not just "Disable path filter" that will just disable the checkbox, a few other situations too.
I have been disabling all filters to get this behavior, find it useful to switch on and off.

I would like to have a menu option to enable the disabled filters too, but that may be too hard to describe for users.

mstv (approved):

RussKie (approved):

PR #9992, opened 2022-05-12
Refresh regardless of formProcess results

### LINKED ISSUE #9987: "Reword commit" editor appears twice and "Browse" window is not refreshed
### Environment

- GitExtensions version: master branch
- GIT version: 2.36.1
- OS version: Windows 10 Professional
- .NET version: 

### Issue description

After selecting "Reword commit", the external editor appears twice and "Browse" window is not refreshed.

### Steps to reproduce

- Right click on a commit and select "Reword commit".
- Close the external editor.

Actual behavior:
- The external editor appears twice.
- "Browse" window is not refreshed.

Expected behavior:
- The external editor appears once.
- "Browse" window is refreshed.

Works fine in version 3.5.4.

### Diagnostics

_No response_

gerhardol:
Fixes #9987

## Proposed changes

Refresh also if formProcess returns Cancel, which it supposedly do if autoclosing

No other similar occurrence for FormProcess (or FormRemoteProcess) was seen in a quick review.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/HelperDialogs/FormProcess.cs
+++ b/GitUI/HelperDialogs/FormProcess.cs
@@ -68,6 +68,9 @@ namespace GitUI.HelperDialogs
         {
         }
 
+        // Note that "DialogResult FormProcess.ShowDialog(owner)" may exit when the process (command) finishes,
+        // so that result is other than OK or Cancel.
+
         public static bool ShowDialog(IWin32Window? owner, ArgumentString arguments, string workingDirectory, string? input, bool useDialogSettings, string? process = null)
         {
             Debug.Assert(owner is not null, "Progress window must be owned by another window! This is a bug, please correct and send a pull request with a fix.");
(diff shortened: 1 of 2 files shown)


RussKie (approved):
Please add a comment why we don't check the results so that we don't undo it in the future.

shirayuki:
not 9887 but 9987

gerhardol:
> Please add a comment why we don't check the results so that we don't undo it in the future.

Will add a comment in FormProcess related to that. (Even if this specific method signature is not overridden.)

I do not believe that one specific use should be commented (there are a few other forms that could have been checking the return status instead of unconditionally refreshing).

mstv (commented):
I do not get it. IMO the added comment does not address the change.

I don't use "autoclose". The auto-refresh does not work after rebase operations including reword (both with and without #9948): The graph is not drawn.

gerhardol:
> I do not get it. 

This PR addresses that the graph is not redrawn at reword, you still see the old commit message.
The behavior if Cancel is pressed (before closing, autoclose or not) will refresh the graph, no major drawback (as the Git command is not killed, that should be done anyway.)

If you experience the problem in https://github.com/gitextensions/gitextensions/issues/9987#issuecomment-1124412192 where the display is completely blank, it is something else.
As the loading seem to be OK, it is more likely a timing issue when (un)hiding InteractiveGitActionControl in notificationBarGitActionInProgress.RefreshGitAction(); But I do not see how.

PR #10003, opened 2022-05-18
Use grayscaled icon for "Reset another branch to here"

zvirja:
## Proposed changes

I often use `Reset another branch to here` action as a way to sync an inactive branch. Currently in a context menu we have two very similar options - `Reset current branch to here` and `Reset another branch to here` and both of them use the same color. I found that this way I cannot use icon as a way to quickly identify the menu entry, but have to resort to text reading and remembering the position, which is slow and slightly annoying.

By using a different icon for `Reset another branch to here` now I can quickly pick the action from list

I used a grayscaled version of the existing icon as we reset inactive branch. So grey should nicely represent the "inactive" idea.

### Before

image

### After

image

## Test methodology

Manual testing

## Merge strategy

----


RussKie (approved):
I too periodically pick the wrong reset option ;))

gerhardol:
Please sign the cla
The .png had incorrect attributes and would not scale correctly, similar to #7942
I had to use the tool to check and pushed an update.
I would rather keep the options next to each other.

gerhardol (changes requested):
The icon does not scale correctly

UnitTests/GitUI.Tests/ResourcesTests.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace GitUITests
+{
+    [TestFixture]
+    public class ResourcesTests
+    {
+        [Test]
+        public void PngResources_ShouldNotContainDPIInformation()
+        {
+            // arrange
+            ResourceSet resourceSet = Images.ResourceManager.GetResourceSet(CultureInfo.InvariantCulture, true, false);
```
RussKie, line 24: `ResourceSet` is [disposable](https://docs.microsoft.com/en-us/dotnet/api/system.resources.resourceset?view=net-6.0)
```suggestion
            using ResourceSet resourceSet = Images.ResourceManager.GetResourceSet(CultureInfo.InvariantCulture, true, false);
```

Also please add argument names for `true` and `false`.

zvirja:
@gerhardol Fixed:
- Signed CLA
- Fixed year for previous CLA entry
- Fixed new PNG resource
- Added unit test for all PNG resources in `Images` resource set and also fixed a couple of existing violations.

Please review again 😊

UnitTests/GitUI.Tests/ResourcesTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                ReadOnlySpan<byte> header = stackalloc byte[] { 0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A };
+                if (!header.SequenceEqual(cursor[0..header.Length]))
+                {
+                    throw new ArgumentException("PNG does not contain a valid header");
+                }
+
+                // Skip header
+                cursor = cursor[header.Length..];
+
+                var sectionTypes = new List<string>();
```
RussKie, line 59: [nit]
```suggestion
                List<string> sectionTypes = new();
```

UnitTests/GitUI.Tests/ResourcesTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using GitUI.Properties;
+using NUnit.Framework;
+
+namespace GitUITests
+{
+    [TestFixture]
+    public class ResourcesTests
+    {
+        [Test]
+        public void PngResources_ShouldNotContainDPIInformation()
```
RussKie, line 21: 🚀

RussKie (approved):

UnitTests/GitUI.Tests/ResourcesTests.cs
zvirja: Actually it should not be disposed, as otherwise a lot of tests are failing. Apparently, you retrieve a global resource, but not own it 😕 
Dropped a comment on that

gerhardol (approved):

zvirja:
@gerhardol I've added another unit test to verify the function itself. It's a 1 pixel base64 encoded PNG and grabbed correct list using another tool. Should be enough to be confident 😉

gerhardol (approved):
Thanks!

RussKie:
Thank you

Issue #9830, opened 2022-01-21
Folder of a repository remains locked even if it's not currently opened


montoner0:
### Environment

- Git Extensions 3.5.4.12724
- Build 65f01f39982ccb121c4df7d4b00b506939553958
- Git 2.33.1.windows.1
- Microsoft Windows NT 10.0.18363.0
- .NET Framework 4.8.4420.0
- DPI 115dpi (120% scaling)

### Issue description

A folder of a previously selected repository remains locked even if this repository is no more current in Git Extensions.

### Steps to reproduce

1. Open a repository in Git Extension
2. Open another repository in Git Extension
3. Try to rename a folder of the first repository
4. Get 'folder in use' error

### Did this work in previous version of GitExtensions?

Not sure. I see this behavior for a quite long time.

Issue #7393, opened 2019-11-08
Test RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder fails


mstv:
## Current behaviour

Build https://ci.appveyor.com/project/gitextensions/gitextensions/builds/28729159 failed because of test

```
1) Error : GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder
System.AggregateException : One or more errors occurred.
  ----> System.ArgumentException : Unable to sort because the IComparer.Compare() method returns inconsistent results. Either a value does not compare equal to itself, or one value repeatedly compared to another value yields different results. IComparer: 'System.Array+FunctorComparer`1[GitUI.UserControls.RevisionGrid.Graph.RevisionGraphRevision]'.
   at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout)
   at System.Threading.Tasks.Task.WaitAll(Task[] tasks)
   at GitUITests.UserControls.RevisionGrid.RevisionGraphMultiThreadingTests.ShouldReorderInTopoOrder() in C:\projects\gitextensions\UnitTests\GitUITests\UserControls\RevisionGrid\Graph\RevisionGraphMultiThreadingTests.cs:line 62
--ArgumentException
   at System.Collections.Generic.IntrospectiveSortUtilities.ThrowOrIgnoreBadComparer(Object comparer)
   at System.Collections.Generic.ArraySortHelper`1.Sort(T[] keys, Int32 index, Int32 length, IComparer`1 comparer)
   at System.Array.Sort[T](T[] array, Int32 index, Int32 length, IComparer`1 comparer)
   at System.Collections.Generic.List`1.Sort(Comparison`1 comparison)
   at GitUI.UserControls.RevisionGrid.Graph.RevisionGraph.BuildOrderedNodesCache(Int32 currentRowIndex) in C:\projects\gitextensions\GitUI\UserControls\RevisionGrid\Graph\RevisionGraph.cs:line 329
   at GitUI.UserControls.RevisionGrid.Graph.RevisionGraph.GetNodeForRow(Int32 row) in C:\projects\gitextensions\GitUI\UserControls\RevisionGrid\Graph\RevisionGraph.cs:line 123
[...]

PR #10030, opened 2022-06-07
RevisionGrid: Cleanup PerformRefreshRevisions

mstv:
Fixes #

## Proposed changes

- Raise questions about `_isRefreshingRevisions` and `cancellationToken`
- Avoid dangling tasks waiting for `semaphoreCurrentCommit` (might contribute to #9830)
- Exception-safe locking of `semaphoreCurrentCommit`
- Fixup typo in `UpdateSelectedRef` argument
- TODO: It is highly desirable that loading the revisions becomes cancellable.

## Test methodology 

- 
- 
- 

## Test environment(s) 

- Git Extensions 33.33.33
- Build a66f50670d4b724937f49cb7155b0904660d23dd
- Git 2.35.2.windows.1
- .NET 6.0.3
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using GitCommands;
@@ -104,7 +105,7 @@ namespace GitUI
         private readonly AuthorRevisionHighlighting _authorHighlighting;
         private readonly Lazy<IndexWatcher> _indexWatcher;
         private readonly BuildServerWatcher _buildServerWatcher;
-        private readonly Timer _selectionTimer;
+        private readonly System.Windows.Forms.Timer _selectionTimer;
         private readonly RevisionGraphColumnProvider _revisionGraphColumnProvider;
         private readonly DataGridViewColumn _maximizedColumn;
         private DataGridViewColumn? _lastVisibleResizableColumn;
@@ -188,7 +189,7 @@ namespace GitUI
             // Delay raising the SelectionChanged event for a barely noticeable period to throttle
             // rapid changes, for example by holding the down arrow key in the revision grid.
             // 75ms is longer than the default keyboard repeat rate of 15 keypresses per second.
-            _selectionTimer = new Timer(components) { Interval = 75 };
+            _selectionTimer = new System.Windows.Forms.Timer(components) { Interval = 75 };
             _selectionTimer.Tick += (_, e) =>
             {
                 _selectionTimer.Stop();
@@ -839,7 +840,7 @@ namespace GitUI
         ///  or it is not in a middle of reconfiguration process guarded by <see cref="SuspendRefreshRevisions"/>
         ///  and <see cref="ResumeRefreshRevisions"/>.
         /// </summary>
-        public bool CanRefresh => !_isRefreshingRevisions && _updatingFilters == 0;
+        private bool CanRefresh => !_isRefreshingRevisions && _updatingFilters == 0;
 
         #region PerformRefreshRevisions
 
@@ -854,6 +855,7 @@ namespace GitUI
 
             if (!CanRefresh)
             {
+                Trace.WriteLine("Ignoring refresh as RefreshRevisions() is already running.");
                 return;
             }
 
@@ -867,25 +869,22 @@ namespace GitUI
             // Revision info is read in two parallel steps:
             // 1. Read current commit, refs, prepare grid etc.
             // 2. Read all revisions with git-log.
+            //    Git will provide log information when available (so slower to first revision if sorting).
             // Both these operations can take a long time and is done async in parallel.
-            // Step 2. requires that most of the information in 1. is prepared when updating the grid.
-            // Git will provide log information when available (so slower to first revision if sorting).
-            // The information from step 1. is therefore protected with a semaphore.
-            // This local semaphore instance is bound in closures below.
-            System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(initialCount: 0, maxCount: 1);
+            // Step 2. requires that the information in 1. is available when adding revisions,
+            // and is therefore protected with a semaphore.
+            SemaphoreSlim semaphoreUpdateGrid = new(initialCount: 0, maxCount: 1);
+            CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
             _isRefreshingRevisions = true;
 
-            // Protected by semaphoreCurrentCommit
             ILookup<ObjectId, IGitRef> refsByObjectId = null;
+            bool firstRevisionReceived = false;
             bool headIsHandled = false;
+            bool hasAnyNotes = false;
 
             // getRefs (refreshing from Browse) is Lazy already, but not from RevGrid (updating filters etc)
             Lazy<IReadOnlyList<IGitRef>> getUnfilteredRefs = new(() => (getRefs ?? capturedModule.GetRefs)(RefsFilter.NoFilter));
             _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(getUnfilteredRefs.Value));
-
-            // optimize for no notes at all
-            Lazy<bool> hasAnyNotes = new(() => AppSettings.ShowGitNotes && getUnfilteredRefs.Value.Any(i => i.CompleteName == GitRefName.RefsNotesPrefix));
-            bool firstRevisionReceived = false;
 
             try
             {
@@ -906,8 +905,6 @@ namespace GitUI
                 _revisionSubscription = observeRevisions
                     .ObserveOn(ThreadPoolScheduler.Instance)
                     .Subscribe(OnRevisionRead, OnRevisionReaderError, OnRevisionReadCompleted);
-
-                System.Threading.CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
 
                 IReadOnlyList<ObjectId>? currentlySelectedObjectIds = _gridView.SelectedObjectIds;
                 _gridView.SuspendLayout();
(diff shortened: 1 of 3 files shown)


GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -968,7 +972,7 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
                     }
                 }).FileAndForget();
 
-                cancellationToken.ThrowIfCancellationRequested();
+                // TODO: can never have an effect in combination with _isRefreshingRevisions: cancellationToken.ThrowIfCancellationRequested();
```
RussKie, line 975: I think this may be a detrimental change. Instead, I think we should do this:
```
                cancellationToken.Register(() => _isRefreshingRevisions = false);
```

https://docs.microsoft.com/dotnet/api/system.threading.cancellationtoken.register

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -873,9 +873,15 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
             // The information from step 1. is therefore protected with a semaphore.
             // This local semaphore instance is bound in closures below.
             System.Threading.SemaphoreSlim semaphoreCurrentCommit = new(initialCount: 0, maxCount: 1);
-            _isRefreshingRevisions = true;
 
             // Protected by semaphoreCurrentCommit
+
+            _isRefreshingRevisions = true;
+            //// TODO: What if we do not reach the end of OnRevisionReadCompleted(), where _isRefreshingRevisions is reset?!?
```
RussKie, line 880: Can you think of a scenario?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -873,9 +873,15 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
...
+
+            System.Threading.CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
```
RussKie, line 882: [nit] please use `using System.Threading;`
```suggestion
            CancellationToken cancellationToken = _refreshRevisionsSequence.Next();
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
mstv: > I think this may be a detrimental change. Instead, I think we should do this:
```
                cancellationToken.Register(() => _isRefreshingRevisions = false);
```
🚀 This is an easy way to avoid `_isRefreshingRevisions` remaining `true`.

Though we must basically await the cancellation of the previous async tasks before we start the next updates.
We have already noticed this necessity in tests of Forms.
This could be done by using dedicated instances of `JoinableTaskFactory`.

---

The structure of this function is:
```
if (_isRefreshingRevisions) return;
_isRefreshingRevisions = true;
_refreshRevisionsSequence.Next();
do some sync stuff which never requests cancellation
RunAsync(...).FileAndForget();
```
The cancellation is requested by the async code (on completion / on error). So it cannot have been triggered here yet.

I just wanted to clarify how the code works and to reduce the confusion for future readers about this pointless check for cancellation. Or have I missed anything?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
mstv: `RevisionReader.ExecuteAsync` does call `OnCompleted` only if it reaches the end of the function (and no cancellation was requested).
So if an exception is thrown, `_isRefreshingRevisions` remains `true` and blocks any further updates.
Most likely, this hat not bitten us because this function mainly starts further async operations. Exceptions from these are forwarded to the UI thread and do not interrupt `ExecuteAsync` itself.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
mstv: OK, (I just moved the line)

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: This is probably not the only pointless check for cancellation...
The cancellation could occur early in the async code (not likely)

The solution is not obvious.
The cancellationToken is needed only in the try{}, but this (minor) change could be merged.

mstv: > The cancellation could occur early in the async code (not likely)

Yes, and if the async task is executed by a background thread, it can already trigger the cancellation before this sync function finishes (on the UI thread).
So this check for cancellation should not be removed.

gerhardol:
I rebased and pushed a cleanup for this PR, suggest it is merged

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1250,15 +1253,17 @@ void OnRevisionReadCompleted()
...
                             () =>
                             {
+                                if (!firstRevisionReceived)
+                                {
+                                    semaphoreCurrentCommit.Wait(cancellationToken);
+                                }
+
                                 _gridView.LoadingCompleted();
                                 bool showArtificial = AddArtificialRevisions();
-                                _gridView.LoadingCompleted();
```
gerhardol, line 1261: This is a merge error, causing debug builds to fail for empty repos (the actual reason I returned to this now)
It is annoying that the dialog is no longer shown, hard to get info about this.
Known but closed: https://github.com/dotnet/runtime/issues/40512

RussKie: It's (kind of) tracked under https://github.com/dotnet/winforms/issues/3739.
But debug asserts only work under a debugger, otherwise the app'd crash.

RussKie (commented):
LGTM but haven't run

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: > But debug asserts only work under a debugger, otherwise the app'd crash.

Same for Trace.Assert from what I read.
This makes it very hard to detect these types of problems, the Windows just disappear without you see why. I have many instances open and may just assume that I failed to select a widow. I will probably start using Release builds for private use. Maybe we should remove all asserts.

gerhardol:
> LGTM but haven't run

Except for empty repos, my changes likely have no change at all.

Updated the comments slightly
Intend to merge tomorrow

gerhardol:
Replaced on useless Debug.Assert that only were triggered in a few situations with a Trace.Write that logs in a few more situations.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1250,15 +1253,17 @@ void OnRevisionReadCompleted()
...
-
-                    // This has to happen on the UI thread
+                    // No revisions at all received
                     this.InvokeAsync(
                             () =>
                             {
+                                if (!firstRevisionReceived)
+                                {
+                                    semaphoreUpdateGrid.Wait(cancellationToken);
+                                }
```
mstv, line 1263: Should we really block the UI thread?

mstv (commented):
LGTM except for the moved wait

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: _isRefreshingRevisions is set to false also in OnError now, so it is possible to reload without restarting GE now

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: fixed

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Is this really changed, RevisionReader runs on the default thread?
If so, SwitchToMainThreadAsync() (and default) is needed in several other situations.

...and most of BuildPathFilter() can be changed to run on the default thread too...

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: I don't think removing asserts is a good idea, they have purpose. Hopefully one day Windows Desktop can fix this.
I do generally run private builds in release though.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: I pushed b3db43141cc2d2142d76e1186c423b4fde389b03 to a private branch, not tested enough

gerhardol:
Sporadic test failure, likely not due to this PR
https://ci.appveyor.com/project/gitextensions/gitextensions/builds/44180783

Implement IComparer in RevisionGraphRevision?

mstv:
> Sporadic test failure, likely not due to this PR https://ci.appveyor.com/project/gitextensions/gitextensions/builds/44180783

#7393
Just guessing: Might this be caused by changing the `Score` while sorting already?

> Implement IComparer in RevisionGraphRevision?

What is your intention?

gerhardol:
> > Sporadic test failure, likely not due to this PR https://ci.appveyor.com/project/gitextensions/gitextensions/builds/44180783
> 
> #7393 Just guessing: Might this be caused by changing the `Score` while sorting already?

This occurs in tests where Artificial is disabled.
Also, I do not believe sorting is used in the app itself, it is the test case.
It is more likely that sorting is done before all revisions are inserted.

> > Implement IComparer in RevisionGraphRevision?
> 
> What is your intention?

Not sure how sorting is done without the comparer, sorting should be done on score, then objectid.

Inserting could be simplified by score adjusted +3.
(This is only occurring when no HEAD so basically only filtering by path.)

mstv (commented):
+2 LGTM and works (cannot approve a PR opened by me)

gerhardol:
squashed, merging tomorrow

RussKie:
👍thanks

PR #10073, opened 2022-07-03
Blame: No change of control at double click

gerhardol:
https://github.com/gitextensions/gitextensions/issues/9796#issuecomment-1150520616

## Proposed changes

Changes the default behavior, to change the revision the left gutter
must be clicked.
If changing the default behavior change is not acceptable here, there are altaernatives:
* Add a setting to continue allowing this. (default off - where?)
* Add a popup confirming the behavior at first use.

I prefer the simple solution, as is.

## Test methodology 

Manual

## Merge strategy

----


pmiossec (approved):
Works well.

mstv (approved):
Thank you

PR #10076, opened 2022-07-20
Exclude boundary commits with --grep

### LINKED ISSUE #10074: Text filter: sort search results by date
### Feature description

Not sure is current behaviour a bug or a "feature".. But anyway - it will be great to be able to sort searching results by the date, with recents on top.. Eg atm we have:
Untitled-1

On the screenshot above - fresh commits presented on top and at bottom.. which makes very hard to understand whats going on...

### Environment

- Git Extensions 3.5.4.12724
- Git 2.36.1.windows.1
- .NET Framework 4.8.4515.0
- DPI 144dpi (150% scaling)

gerhardol:
Fixes #10074

## Proposed changes

git-log --boundary and --all is included in many revision listings. (The options are currently used together.)
This basically means that the parent to matches is included. This may be desired in some situations, but not with `--grep`.

It is not obvious that --boundary and --grep should be used where they are used now.
Similarly, -all may still be used with --grep. (I do not see how that is used at all though.)
The options can be tuned and expanded in general...

Also fixed the label for search.

### Before

image

image

### After

image

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -114,19 +114,26 @@ namespace GitUI.UserControls.RevisionGrid
             get
             {
                 RefFilterOptions refFilterOptions;
-                if (!ByBranchFilter)
+                if (!string.IsNullOrWhiteSpace(Message))
                 {
+                    refFilterOptions = RefFilterOptions.None;
+                }
+                else if (!ByBranchFilter)
+                {
+                    // Include parents to matches
                     refFilterOptions = RefFilterOptions.All | RefFilterOptions.Boundary;
                 }
                 else if (ShowCurrentBranchOnly)
                 {
                     refFilterOptions = RefFilterOptions.None;
                 }
+                else if (BranchFilter.Length > 0)
+                {
+                    refFilterOptions = RefFilterOptions.Branches;
+                }
                 else
                 {
-                    refFilterOptions = BranchFilter.Length > 0
-                        ? RefFilterOptions.Branches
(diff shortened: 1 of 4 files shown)


mstv (approved):
👍

RussKie:
Do we tests for this?

gerhardol:
> Do we tests for this?

No. It seems like there is no reasoning to the configuration, which is needed first.

gerhardol:
Rebased, added a stupid test

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -80,6 +80,7 @@ public string DiffContent
 
         public int CommitsLimitDefault => AppSettings.MaxRevisionGraphCommits;
 
+        [System.Diagnostics.DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
RussKie, line 83: What prompted this addition?

RussKie (commented):
👀

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: Msvs could not inspect when debugging FilterInfo, stopped with exception.

RussKie: What exception did you get? Can you share repro steps?

RussKie (commented):
LGTM modulo the question about the attribute

RussKie (approved):

Issue #4942, opened 2018-05-13
Exclude submodules from appearing in Recent submodules


gerhardol:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
GE adds every repo to the submodule list, including submodules

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**
Submodules should not be added to the menu (possibly option).
Normally, the submodule is dependent so you go from one superrepo to another. I have seen users being confused about using the repo menu vs the submodule menu.
A submodule centric use will require a very long (unusable) menu.

I work with about 15 different "products" and I have at least 2 clones (really workingtrees) for each. Something like 30-50 superprojects accessed regularily
Each product has about 5 common submodules that has some common submodules too.
Due to the submodules, the list need to include at least 30*5 instead of just 30-50 records

Origin in #4931 

Note: It could be possible to differ between explicit Open repo and open from menu/doubleclicking where the latter should not be added to recents. If open in another window this information most be included as command line parameter.
It could be possible to just look at the repo (and ignore submodules).

**Environment you encounter the issue:**

- GitExtensions version: all

**Did this work in previous version of GitExtensions (which)?**
no

gerhardol:
An example for GitExtensions below
But instead of 4 superprojects as in this example, I have 30-50 superprojects. Instead of 5 submodules, I may access 10 or more submodules of each subrepo. 
So the recently used is not always very useful. The dashboard can be, but I change superrepo clones quite often and may not update the categories.

![image](https://user-images.githubusercontent.com/6248932/46764252-46f2f880-ccdc-11e8-87ea-fe2c785fbede.png)

So I want an optional way to not add submodules if opened from the submodule menu, to keep recent in the list.

RussKie:
Just to confirm I understand you correctly, are you talking about this menu?
![image](https://user-images.githubusercontent.com/4403806/46787831-5a599e80-cd84-11e8-97de-f51c1a9d33ba.png)

If so, I think it kind of makes sense not to add submodules opened from here to the "recent" list.
Perhaps the menu can be rearranged to something like:
```
Update all submodules
------------------------
Sub1
Sub2
...
SubN
------------------------
[x] Add to recent repositories list
```

gerhardol:
You listed the submodule list
I listed the recent repo list (to the right)
What I want is for submodule (repos) opened from the submodule list to be added to the recent repo list

[...]

PR #10084, opened 2022-07-24
Add most recent repos to all recent list

gerhardol:
Part of #4942 

## Proposed changes

Replace group of "less recent" repositories with "all recent" repositories list.
This simplifies my usage: Quick view at most recent that occasionally is to quick and fail, then a closer look at "all recent".
This can be seen as a step to deprecate "most recent" as suggested by RussKie (I really want to keep it though, even if few users know about it).

"Less recent" renamed to "Recent" in the UI, "All recent" in the code.
"All recent" is still default 0, you have to enable the functionality.
"Most recent" renamed to "Pinned" in the UI and the code.

### Before

image

### After

image

## Test methodology 

Manual

There are no tests for the form, just for the repo handling (that mentions "most recent" when handling the groups, not the same as "most recent repos").

## Merge strategy

First commit is the actual change, second is just cleaning up the use of "less recent repos".

----


GitCommands/UserRepositoryHistory/Repository.cs
```diff
@@ -10,7 +10,7 @@ public class Repository
         public enum RepositoryAnchor
         {
             MostRecent,
```
RussKie, line 12: So would you say these are "pinned"?

RussKie (approved):
I still think we have a lot of unnecessary complexity here, but I'd take it as a step in the right direction :)

GitCommands/UserRepositoryHistory/Repository.cs
pmiossec: +100
I was about to ask for renaming to "pinned" also...

gerhardol: Another massive change...
Rebased too.

GitUI/Translation/English.xlf
```diff
@@ -6562,15 +6562,15 @@ Nothing to rebase.</source>
...
+        <source>Recent repositories</source>
         <target />
       </trans-unit>
       <trans-unit id="lblRecentRepositoriesHistorySize.Text">
         <source>Recent repositories history size</source>
         <target />
       </trans-unit>
       <trans-unit id="maxRecentRepositories.Text">
-        <source>Maximum number of most recent repositories</source>
+        <source>Maximum number of pinned recent repositories</source>
```
RussKie, line 6573: Should be "Maximum number of pinned repositories"

gerhardol: This was intentional: The description is intended to differentiate pinned from favorites

RussKie: This made me question the difference between the pinned and favourites. I do feel that there's a significant overlap between these two, and a lot of added complexity. If we look at VS or other IDEs or productivity apps - most of those have a notion of "recent"s, and some have notion of "pinned" projects/files/etc.

gerhardol: I say the GE repo favorites is a good name, there are categories and such.
In general, Pinned is similar to Favorites but you only select a certain items, no categories.

Most recent is in my opinion a better name than Pinned, as this is not something you select yourself. As both you and @pmiossec suggested the name so I agreed, to have the change ready for the alpha.
Maybe *Top recent* is better name?
I can just revert the renaming to Pinned too if you prefer that.

mstv:
Both the repo dropdown in the toolbar and the Dashboard are based on these lists.
This PR affects the repo dropdown in the way I want it: "less recent repositories" turned into "all recent repositories". 👍
But here "pinned" for "most recent" is / would be arguable. It still works as "most recent" and this is just fine.
(I don't need the Dashboard. I am used to the pinned Windows taskbar icon with its jump list and / or the repo dropdown if the jump list is incomplete.)

image

gerhardol:
> But here "pinned" for "most recent" is / would be arguable. It still works as "most recent" and this is just fine.

Agree. I got two votes to change this and was at that time OK with it but should have argued...
But most recent requires you to think a little.
Personally, I do not feel it is worth it to change the naming.

PR #10090, opened 2022-07-27
fix: Cannot cherry-pick new file from stash

zvirja:
## Proposed changes

Fix bug when cherry-picking from stash a new file. Currently when action is performed nothing happens. 

That happens because we cannot internally resolve diff chunks in a new file.

## Test methodology 

Manual testing

## Merge strategy

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -1539,24 +1539,43 @@ namespace GitUI.Editor
             }
 
             byte[]? patch;
-            if (reverse)
+
+            Validates.NotNull(_viewItem);
+            if (_viewItem.Item.IsNew)
             {
-                patch = PatchManager.GetResetWorkTreeLinesAsPatch(
+                var treeGuid = reverse ? _viewItem.Item.TreeGuid?.ToString() : null;
+                patch = PatchManager.GetSelectedLinesAsNewPatch(
                     Module,
+                    _viewItem.Item.Name,
                     GetText(),
                     selectionStart,
                     selectionLength,
-                    Encoding);
+                    Encoding,
+                    reset: reverse,
+                    FilePreamble,
+                    treeGuid);
             }
             else
             {
-                patch = PatchManager.GetSelectedLinesAsPatch(
-                    GetText(),
-                    selectionStart,
-                    selectionLength,
-                    false,
-                    Encoding,
-                    false);
+                if (reverse)
+                {
+                    patch = PatchManager.GetResetWorkTreeLinesAsPatch(
+                        Module,
+                        GetText(),
+                        selectionStart,
+                        selectionLength,
+                        Encoding);
+                }
+                else
+                {
+                    patch = PatchManager.GetSelectedLinesAsPatch(
+                        GetText(),
+                        selectionStart,
+                        selectionLength,
+                        false,
+                        Encoding,
+                        false);
+                }
             }
 
             if (patch is null || patch.Length == 0)
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -985,14 +985,22 @@ namespace GitUI.Editor
                 && (fileName.EndsWith(".diff", StringComparison.OrdinalIgnoreCase)
                     || fileName.EndsWith(".patch", StringComparison.OrdinalIgnoreCase)))
             {
+                // Override the set view mode
                 _viewMode = ViewMode.FixedDiff;
             }
 
-            SupportLinePatching = ((IsDiffView(_viewMode) && (text?.Contains("@@") ?? false)
+            SupportLinePatching =
+
+                // Diffs, currently requires that the file to update exists
+                ((IsDiffView(_viewMode) && (text?.Contains("@@") ?? false)
                         && File.Exists(_fullPathResolver.Resolve(fileName)))
+
+                // New files, patches only applies for artificial or if the file does not exist
                     || ((item?.Item.IsNew ?? false)
                         && ((item.Item.Staged is StagedStatus.WorkTree or StagedStatus.Index)
                             || !File.Exists(_fullPathResolver.Resolve(fileName)))))
+
+                // No patching allowed if no worktree
                 && !Module.IsBareRepository();
 
             SetVisibilityDiffContextMenu(_viewMode);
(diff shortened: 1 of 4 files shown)


gerhardol:
Can you describe  how to setup a repo where this can be tested?
A new file (viewed) does not support as a "source for line patches, so I cannot see how this can make any change.

zvirja:
Sure! Sorry if I wasn't clear enough in my original description ☺️

- Create a new file
- Stash it
- Go to stashes
- Right click on new file and select "cherry-pick"

Previously nothing happened. Now also nothing visually happens, but once you return back - file is cherry-picked

gerhardol:
That this is in FormStash is key information.

As it is now, the menu item in Stash form should be suppressed: `SupportLinePatching` is false, the item should not be visible.
This should be handled as for  GitUI\CommandsDialogs\RevisionDiffControl.cs:438
(I do not believe this menu item is available elsewhere.)

However, line patching could (probably) be enabled generally for new files. Quite often such patches fail though... (Just adding files that really work.)
```c#
            SupportLinePatching = ((IsDiffView(_viewMode) && (text?.Contains("@@") ?? false)
                    && File.Exists(_fullPathResolver.Resolve(fileName)))
                        || (item?.Item.IsNew ?? false))
                    && !Module.IsBareRepository();
```

There may be some scenarios that I have missed though.
Some other code comments separately

GitUI/Editor/FileViewer.cs
```diff
@@ -1539,24 +1539,43 @@ private void ApplySelectedLines(bool allFile, bool reverse)
...
                     GetText(),
                     selectionStart,
                     selectionLength,
-                    Encoding);
+                    Encoding,
+                    reset: reverse,
+                    FilePreamble,
+                    treeGuid);
             }
             else
```
gerhardol, line 1558: reduce indentation like in ResetNoncommittedSelectedLines
```suggestion
            else if (reverse)
```

GitUI/Editor/FileViewer.cs
```diff
@@ -1539,24 +1539,43 @@ private void ApplySelectedLines(bool allFile, bool reverse)
...
+                patch = PatchManager.GetSelectedLinesAsNewPatch(
                     Module,
+                    _viewItem.Item.Name,
                     GetText(),
                     selectionStart,
                     selectionLength,
-                    Encoding);
+                    Encoding,
+                    reset: reverse,
+                    FilePreamble,
```
gerhardol, line 1555: Validate FilePreamble, this was not set by all forms this was used.

GitUI/Editor/FileViewer.cs
```diff
@@ -1539,24 +1539,43 @@ private void ApplySelectedLines(bool allFile, bool reverse)
             }
 
             byte[]? patch;
-            if (reverse)
+
+            Validates.NotNull(_viewItem);
+            if (_viewItem.Item.IsNew)
```
gerhardol, line 1544: Can this be aligned with ResetNoncommittedSelectedLines etc, almost the same logic?

gerhardol (changes requested):

gerhardol:
I pushed a commit that I believe works.

The actual change is that line patching is enabled from new files if the file does not exist already.
Some additional checks if line patch and cherry picking a file is allowed updated too.

zvirja:
@gerhardol Thanks for your help with enhancing the code. It definitely goes beyond my knowledge of this codebase to make and review possible impact of changes like this.

I've just manually tested the change and seems to work locally fine. I cannot cherry-pick a new file if it already exists as you mentioned. While we can go with a more sophisticated logic of actually trying to merge changes - I think that's a good enough starting point which will remove the confusion we currently have 😊 

Thanks again for your help! Let me know if some more actions from my side are expected, as looks like you fixed everything yourself..

gerhardol:
> I've just manually tested the change and seems to work locally fine. I cannot cherry-pick a new file if it already exists as you mentioned. While we can go with a more sophisticated logic of actually trying to merge changes - I think that's a good enough starting point which will remove the confusion we currently have 😊

To allow patches from a new file is hard to handle: The context may not match.
There are also rules that may need tweaking for patching files already in the index (so the handling may need to be more complex....).
In these scenarios it may be easier to just compare the files.

> Thanks again for your help! Let me know if some more actions from my side are expected, as looks like you fixed everything yourself..

Another maintainer need to approve this now when I did most of the change (even if I will approve).
The actual changes (the first commit and assign of SupportLinePatching) is quite minor expanding, so I hope this is easy.

gerhardol (approved):

GitUI/Editor/FileViewer.cs
```diff
@@ -982,10 +988,12 @@ private void ResetView(ViewMode viewMode, string? fileName, FileStatusItem? item
...
-            SupportLinePatching = ((IsDiffView(_viewMode) && (text?.Contains("@@") ?? false))
-                        || (item is not null && item.Item.IsNew && (item.Item.Staged is StagedStatus.WorkTree or StagedStatus.Index)))
-                    && !Module.IsBareRepository()
-                    && File.Exists(_fullPathResolver.Resolve(fileName));
+            SupportLinePatching = ((IsDiffView(_viewMode) && (text?.Contains("@@") ?? false)
+                        && File.Exists(_fullPathResolver.Resolve(fileName)))
+                    || ((item?.Item.IsNew ?? false)
+                        && ((item.Item.Staged is StagedStatus.WorkTree or StagedStatus.Index)
+                            || !File.Exists(_fullPathResolver.Resolve(fileName)))))
+                && !Module.IsBareRepository();
```
RussKie, line 996: This doesn't look readable at all. Can we please break it up and use temp variables?
```cs
if (!Module.IsBareRepository())
{
    SupportLinePatching = false;
}
else 
{
    bool ... = IsDiffView(_viewMode) && (text?.Contains("@@") ?? false) && File.Exists(_fullPathResolver.Resolve(fileName));
    bool ... = (item?.Item.IsNew ?? false) && ((item.Item.Staged is StagedStatus.WorkTree or StagedStatus.Index) || !File.Exists(_fullPathResolver.Resolve(fileName))));
    SupportLinePatching = ... || ...;
}
```

RussKie: Some comments could help too

gerhardol: Added comments
I prefer one assignment, to make sure the more expensive calls to File.Exists() is just executed once.

RussKie (approved):

PR #10101, opened 2022-07-29
RevDiff: Align rules for hotkey execution to menu items

gerhardol:
Related to #10090

## Proposed changes

The rules to allow reset and staging from hotkeys were evaluated separately from the menu item rules.
Set Enabled property (as Visible cannot be read).

It was only possible to (un)stage multiple files from menu due to the incorrect tests, this has been aligned.
It is also possible to reset multiple files by menu/hotkey.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -677,21 +677,26 @@ namespace GitUI.CommandsDialogs
                 ? new Font(fileHistoryDiffToolstripMenuItem.Font, FontStyle.Regular)
                 : new Font(fileHistoryDiffToolstripMenuItem.Font, FontStyle.Bold);
 
-            diffUpdateSubmoduleMenuItem.Visible =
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -1135,35 +1145,33 @@ private bool CanResetToRevision(ObjectId guid)
                    && guid != ObjectId.CombinedDiffId;
         }
 
-        private void resetFileToToolStripMenuItem_DropDownOpening(object sender, EventArgs e)
+        private void ResetFileToToolStripMenuItem_Init()
```
RussKie, line 1148:
```suggestion
        private void InitResetFileToToolStripMenuItem()
```

gerhardol: squashed the change, no other changes

RussKie (approved):

PR #10118, opened 2022-07-30
FormDiff:  Show range diff with merge base

gerhardol:
Fixes #10081

## Proposed changes

- Refactor base->first, head->second
- FormDiff: Show range diff with set merge base instead of only diff first-second

Note: I consider FormDiff to be obsolete, Browse offers more functionality than these separate forms that are difficult to maintain.

Note: merge base in diff is set initially and matches the ui. it will not change if you select new branches in the form and if the merge base is no longer a base a multi diff without range diff will be shown

### Before

image

### After

image

## Test methodology 

Manual

## Merge strategy

To be decided later.
The refactoring commit could be separate.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormDiff.Designer.cs
+++ b/GitUI/CommandsDialogs/FormDiff.Designer.cs
@@ -27,16 +27,16 @@ namespace GitUI.CommandsDialogs
             this.ckCompareToMergeBase = new System.Windows.Forms.CheckBox();
             this.btnCompareDirectoriesWithDiffTool = new System.Windows.Forms.Button();
             this.diffOptionsPanel = new System.Windows.Forms.FlowLayoutPanel();
-            this.baseCommitGroup = new System.Windows.Forms.GroupBox();
-            this.baseCommitPanel = new System.Windows.Forms.FlowLayoutPanel();
-            this.lblBaseCommit = new System.Windows.Forms.Label();
-            this.btnAnotherBaseBranch = new System.Windows.Forms.Button();
-            this.btnAnotherBaseCommit = new System.Windows.Forms.Button();
-            this.headCommitGroup = new System.Windows.Forms.GroupBox();
-            this.headCommitPanel = new System.Windows.Forms.FlowLayoutPanel();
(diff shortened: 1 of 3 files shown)


gerhardol:
@crypto-rsa Please review and test

crypto-rsa:
I cannot properly review the changes since my knowledge of the code is very limited. However, I have tested the branch and it seems to behave as it did before the functionality was removed in #10072. So OK by me.

> Note: I consider FormDiff to be obsolete, Browse offers more functionality than these separate forms that are difficult to maintain.

I do not insist on using a separate _form_. As I mentioned in #10081, there are two issues preventing me from using the **single** GE window for reviews involving `range-diff`:
1. I cannot browse other commits without losing the `range-diff`
2. I can only open the `range-diff` by manually scrolling the revision grid and selecting the revisions with the mouse and the Ctrl key. In other words, I can't use any keyboard navigation to find the revisions I'm interested in.

As long as there is a way to quickly (!) open a **new** browse window with the two selected revisions (or after running the _Compare to BASE_ command), that would be fine.

gerhardol:
> I cannot properly review the changes since my knowledge of the code is very limited. However, I have tested the branch and it seems to behave as it did before the functionality was removed in #10072. So OK by me.

That is the problem with these forms: Someone need to know how they are used and why, taking some responsibility.
You can always make a formal review and just write what you have reviewed (like usage) so some other reviewer get assistance.

> As long as there is a way to quickly (!) open a **new** browse window with the two selected revisions (or after running the _Compare to BASE_ command), that would be fine.

Not after you have selected the revisions, but you can always start a new revision to do your commit-by-commit diff.

Maybe expanding opening new instance similar to submodules is what you want?
image

gerhardol:
See #10145 and #10146 for doubleclicking in sidepanel.
These are not dependent on this PR.
(For me it is OK functionality in this PR, but I would try to avoid make changes in these forms in general...)

mstv (approved):
LGTM, have not tested
Please do not squash, i.e. keep the separate refactoring commit.

PR #10119, opened 2022-07-31
Restore WindowState when activating

### LINKED ISSUE #9752: Cannot bring GE to front

gerhardol:
Fixes #9752

## Proposed changes

This is a fix, probably not the correct solution, added for review.

Occasionally when restoring GE from minimized, the app fail to stay open.

One drawback with the fix is that rightclick to close on the minature in the taskbar requires that the app is not minimized. OK to select the cross.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1963,7 +1963,12 @@ namespace GitCommands
         public static bool WriteErrorLog
         {
             get => GetBool("WriteErrorLog", false);
-            set => SetBool("WriteErrorLog", value);
+        }
+
+        // Set manually in settings file
+        public static bool WorkaroundRestoreFromMinimize
+        {
+            get => GetBool("WorkaroundRestoreFromMinimize", false);
         }
 
         private static IEnumerable<(string name, string value)> GetSettingsFromRegistry()
--- a/ResourceManager/GitExtensionsFormBase.cs
+++ b/ResourceManager/GitExtensionsFormBase.cs
@@ -97,6 +97,13 @@ namespace ResourceManager
                 if (m.Msg == NativeMethods.WM_ACTIVATEAPP && m.WParam != IntPtr.Zero)
                 {
                     OnApplicationActivated();
+                    if (WindowState == FormWindowState.Minimized && Owner is null && AppSettings.WorkaroundRestoreFromMinimize)
+                    {
+                        // Changed behavior from .NET4 to .NET5, application occasionally requires explicit "restore" in Taskbar.
+                        // See https://github.com/gitextensions/gitextensions/pull/10119.
+                        Trace.WriteLine("WindowState is unexpectedly Minimized in OnApplicationActivated(), restoring.");
+                        WindowState = FormWindowState.Normal;
+                    }
                 }
             }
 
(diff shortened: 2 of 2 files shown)


RussKie:
This does feel like a hack...

gerhardol:
> This does feel like a hack...

It certainly is (both WIP and draft), I could not find the core issue. I assume this is something in .NET.

But this is so annoying, if there is no other solution I want to include this (will use it in my private builds).

RussKie:
If this is addressing the issue - let's take it

gerhardol:
> If this is addressing the issue - let's take it

For me it is and I am using it.
But it is a hack, I have not found any other reports with similar issues. I hope that someone with WinForms knowledge could investigate a little about this...

For the 4.0 alpha, at least a trace printout should be added. It may be good to find how many users that experience the problem.

gerhardol:
> If this is addressing the issue - let's take it

I added a Trace printout and a comment (as well as commit message update).
I suggest we include this in next alpha/beta.

gerhardol:
When this "correction" kicks in, GE is restored also if it was maximized prior to the minimize. It is a little unexpected.
This could maybe be improved by using _windowPositionManager similar to RestorePosition(), but I want to keep this simple and still believe that there is a better solution...

mstv:
> When this "correction" kicks in, GE is restored also if it was maximized prior to the minimize. It is a little unexpected.

more than unexpected -- and occurs several times a day in contrast to #9752, which I cannot reproduce.

Though when trying your steps to reproduce, I noticed that the window positions are not stored correctly when closed minimized, at least the maximized state.
Perhaps, we should not store them at all if minimized (and keep the previously stored positions).

gerhardol:
> more than unexpected -- and occurs several times a day in contrast to #9752, which I cannot reproduce.

Then you may have some part of the problem, but Minimized is changed later?

> Though when trying your steps to reproduce, I noticed that the window positions are not stored correctly when closed minimized, at least the maximized state. Perhaps, we should not store them at all if minimized (and keep the previously stored positions).

seem reasonable

gerhardol:
Maybe the state can be saved when minimizing, I have not found the win call for this

> Though when trying your steps to reproduce, I noticed that the window positions are not stored correctly when closed minimized, at least the maximized state. Perhaps, we should not store them at all if minimized (and keep the previously stored positions).

I do not see this. When saving in Minimized or Maximized state, the RestoreBounds are used, which are OK for me.

gerhardol:
> more than unexpected -- and occurs several times a day in contrast to https://github.com/gitextensions/gitextensions/issues/9752, which I cannot reproduce.

I did not find a reasonable way to get the status before it was minimized.
The behavior was therefore changed to require a setting to be acitivated.
The setting currently requires manually editing of the settings file

%APPDATA%\GitExtensions\GitExtensions\GitExtensions.settings
(e.g. C:\Users\user\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.settings
```
  <item>
    <key>
      <string>WorkaroundRestoreFromMinimize</string>
    </key>
    <value>
      <string>true</string>
    </value>
  </item>
```

Can we merge with this to get a new alpha/beta/rc soon?

mstv (approved):

PR #10125, opened 2022-08-02
Improve menu responsiveness with SuspendLayout

gerhardol:
Followup to #10124

## Proposed changes

Menu dropdowns with many items are slow if Layout is enabled
while adding items. For instance Browse "current branch" may require
several seconds to open after clicking if layout is not suspended.

This is basically situations with `DropDowItems.Clear()` in DropDownOpening.
I have updated most menus with this pattern, just skipped a few I did not feel was relevant.
Cannot really see if SuspendLayout should be done if the parent control is suspended (I assume this is not relevant, some search hits indicate it is though), not changed.

Earlier solution, reverted and changed to above:
~~DropDownItems.AddRange() significantly improves performance when adding to menus.
It requires several seconds to open branches menu in Browse before this, with this change it is more or less instance.
(I have over 100 branches, but max 100 are added to the menu, could be limited further.)
(This menu is not really important for me).~~

~~Similar for repo menu, improving #10124 further. 
(#10124 has the important change).~~

~~A few other situations that may benefit from AddRange (but not as much as those above),
related to adding items when starting Browse.~~

~~* GitUI\CommandsDialogs\BrowseDialog\FormBrowseMenus.cs:194~~
~~* GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:360~~

## Test methodology 

Tests are updated.

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -716,6 +716,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
                 return;
             }
 
+            tsmiCategories.DropDown.SuspendLayout();
             tsmiCategories.DropDownItems.Clear();
 
             var categories = GetCategories();
@@ -732,6 +733,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
             }
 
             tsmiCategories.DropDownItems.Add(tsmiCategoryAdd);
+            tsmiCategories.DropDown.ResumeLayout();
 
             RepositoryContextAction(tsmiCategories, selectedRepositoryItem =>
             {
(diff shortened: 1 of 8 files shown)


RussKie (approved):

RussKie:
> DropDownItems.AddRange() significantly improves performance when adding to menus.

Wow, that's an interesting observation.

gerhardol:
> > DropDownItems.AddRange() significantly improves performance when adding to menus.
> 
> Wow, that's an interesting observation.

This was actually a comment about this in GitUI\CommandsDialogs\FormBrowse.cs:2717
 The additions could be improved further too), there are like 5 Add and 2 AddRanges for recent repos now, but I believe this is good enough as is.

I see that there are DropDown.SuspendLayout() methods too, would maybe be even better, but that is a bigger change.

I found one minor simplification/improvement and one minor bug. Plan to merge tomorrow.

gerhardol:
> I see that there are DropDown.SuspendLayout() methods too, would maybe be even better, but that is a bigger change.

I reverted the first changes and implemented suspend instead, so new review required.
(Cannot see any performance difference, should be easier to review though).

pmiossec (approved):

gerhardol:
Will cherrypick to 4.0 too

gerhardol:
@RussKie Any opinions to the (completely) changed PR and to inclusion in 4.0?

RussKie:
The change looks good to me. If you see an observable improvement - let's take it into the 4.0.

PR #10127, opened 2022-08-03
Fix search not launching/working when changing filter selection

pmiossec:
Fixes this use case:

1. Searching for a string (by default searching in commit message)
2. Selecting "Diff contains" and unselecting "Commit message"
3. Launch a search but search/filtering is not launched :(

Note: I think this fix should also back-ported to v4

## Proposed changes

- Fix filtering when playing with type of filtering

image

### Before

No filter launched (so empty revision grid)
image

### After

The filtering is launched (so get results)
image

## Test methodology 

- Manual
- Unit tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build df80a9de42f264168585fa79e0f07645de433013
- .NET 6.0.6
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -223,6 +223,7 @@ namespace GitUI.UserControls.RevisionGrid
                    ByAuthor ||
                    ByCommitter ||
                    ByMessage ||
+                   ByDiffContent ||
                    ByPathFilter ||
                    ByBranchFilter ||
                    ShowSimplifyByDecoration;
@@ -311,6 +312,7 @@ namespace GitUI.UserControls.RevisionGrid
             ByAuthor = false;
             ByCommitter = false;
             ByMessage = false;
+            ByDiffContent = false;
             ByPathFilter = false;
             ByBranchFilter = false;
             ShowSimplifyByDecoration = false;
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -240,55 +240,68 @@ namespace GitUI.UserControls.RevisionGrid
                 Debug.Fail("Not supported");
             }
 
+            bool searchParametersChanged = filter.FilterByAuthor != ByAuthor
+                                          || filter.FilterByCommitter != ByCommitter
+                                          || filter.FilterByCommit != ByMessage
+                                          || filter.FilterByDiffContent != ByDiffContent;
+
             if (filter.FilterByAuthor)
             {
-                if (ByAuthor && string.Equals(Author, filter.Text, StringComparison.CurrentCulture))
+                if (!string.Equals(Author, filter.Text, StringComparison.CurrentCulture))
                 {
-                    return false;
+                    ByAuthor = !string.IsNullOrWhiteSpace(filter.Text);
+                    Author = filter.Text;
+                    searchParametersChanged = true;
                 }
-
-                ByAuthor = !string.IsNullOrWhiteSpace(filter.Text);
-                Author = filter.Text;
-                return true;
+            }
+            else
+            {
+                ByAuthor = false;
             }
 
             if (filter.FilterByCommitter)
             {
-                if (ByCommitter && string.Equals(Committer, filter.Text, StringComparison.CurrentCulture))
+                if (!string.Equals(Committer, filter.Text, StringComparison.CurrentCulture))
                 {
-                    return false;
+                    ByCommitter = !string.IsNullOrWhiteSpace(filter.Text);
+                    Committer = filter.Text;
+                    searchParametersChanged = true;
                 }
-
-                ByCommitter = !string.IsNullOrWhiteSpace(filter.Text);
-                Committer = filter.Text;
-                return true;
+            }
+            else
+            {
+                ByCommitter = false;
             }
 
             if (filter.FilterByCommit)
             {
-                if (ByMessage && string.Equals(Message, filter.Text, StringComparison.CurrentCulture))
+                if (!string.Equals(Message, filter.Text, StringComparison.CurrentCulture))
                 {
-                    return false;
+                    ByMessage = !string.IsNullOrWhiteSpace(filter.Text);
+                    Message = filter.Text;
+                    searchParametersChanged = true;
                 }
-
-                ByMessage = !string.IsNullOrWhiteSpace(filter.Text);
-                Message = filter.Text;
-                return true;
+            }
+            else
+            {
+                ByMessage = false;
             }
 
             if (filter.FilterByDiffContent)
             {
-                if (ByDiffContent && string.Equals(DiffContent, filter.Text, StringComparison.CurrentCulture))
+                if (!string.Equals(DiffContent, filter.Text, StringComparison.CurrentCulture))
                 {
-                    return false;
+                    ByDiffContent = !string.IsNullOrWhiteSpace(filter.Text);
+                    DiffContent = filter.Text;
+                    searchParametersChanged = true;
                 }
-
-                ByDiffContent = !string.IsNullOrWhiteSpace(filter.Text);
-                DiffContent = filter.Text;
-                return true;
+            }
+            else
+            {
+                ByDiffContent = false;
             }
 
-            return false;
+            return searchParametersChanged;
         }
 
         public void ResetAllFilters()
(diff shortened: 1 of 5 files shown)


pmiossec:
Not exactly linked to this fix, but should be discussed:

Maybe we should uncheck a type of filter if it is "incompatible" (because that is a "AND" logic which is applied between all selected filters) with the one selected.

For example the "Diff contains" is incompatible with all the other filters because we nearly never search the exact same string in the Diffs **AND** the commit message or the committer/authors.

What do you think? (If yes, I will open an issue to fix it later --on day 😉 --)

gerhardol:
> Maybe we should uncheck a type of filter if it is "incompatible" (because that is a "AND" logic which is applied between all selected filters) with the one selected.

Yes
Maybe add an official way to add git-log options too

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -240,55 +240,68 @@ public bool Apply(RevisionFilter filter)
                 Debug.Fail("Not supported");
             }
 
+            var searchParametersChanged = filter.FilterByAuthor != ByAuthor
```
RussKie, line 243: Please don't use `var` for intrinsic types. Also don't use when the type isn't obvious (e.g., `var result = DoWork();`).

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
```diff
@@ -741,5 +741,101 @@ public void FilterInfo_GetRevisionFilter(string author, string committer, string
             filterInfo.GetSummary().Should().Be(expectedSummary);
             filterInfo.GetRevisionFilter().ToString().Should().Be(expectedArgs);
         }
+
+        [Test]
+        public void FilterInfo_Apply_ByCommitMessage()
+        {
+            FilterInfo filterInfo = new();
+            var filterLaunched = filterInfo.Apply(new RevisionFilter("message", true, false, false, false));
```
RussKie, line 749: Please add arg names to help with readability

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
```diff
@@ -741,5 +741,101 @@ public void FilterInfo_GetRevisionFilter(string author, string committer, string
...
+            filterInfo.DiffContent.Should().Be("diff");
+
+            filterInfo.ByMessage.Should().BeFalse();
+            filterInfo.ByAuthor.Should().BeFalse();
+            filterInfo.ByCommitter.Should().BeFalse();
+        }
+
+        [Test]
+        public void FilterInfo_Apply_ByMessage_Then_By_DiffContent()
+        {
```
RussKie, line 807: A unit test should typically perform a single action and the assert the results. This test appears to perform several actions, and I'm not saying this is necessarily wrong, but some comments would help understanding what and why we are doing so.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -240,55 +240,68 @@ public bool Apply(RevisionFilter filter)
...
+                                          || filter.FilterByCommitter != ByCommitter
+                                          || filter.FilterByCommit != ByMessage
+                                          || filter.FilterByDiffContent != ByDiffContent;
+
             if (filter.FilterByAuthor)
             {
-                if (ByAuthor && string.Equals(Author, filter.Text, StringComparison.CurrentCulture))
```
gerhardol, line 245: As it is in master, ByAuthor and similar must be set for Author to be set. If this check is not needed, the explicit setting should be removed. Maybe `public bool ByAuthor { get; private set; }`
FilterByAuthor is related to disbling/enabling options without clearing the contents.

gerhardol:
This PR makes one behavior worse: If you enter text in the searchbox but leave without enter a search is not done (same as before). Search should maybe be done on leaving the box.
However, if you switch to another repo the search still occurs even if the searchbox is empty and the Advanced menu is empty.

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
pmiossec: done.

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
pmiossec: Sorry, resharper helps a lot with that...

GitUI/UserControls/RevisionGrid/FilterInfo.cs
pmiossec: I tend to think that boolean is a type where it is obvious enough but ok.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
pmiossec: Sorry, after reading multiple times carefully, I don't get it.

pmiossec:
> This PR makes one behavior worse: If you enter text in the searchbox but leave without enter a search is not done (same as before). Search should maybe be done on leaving the box. However, if you switch to another repo the search still occurs even if the searchbox is empty and the Advanced menu is empty.

I wil have a look to try to reproduce. But as it is start of vacations, I don't know at the moment when...

GitUI/UserControls/RevisionGrid/FilterInfo.cs
pmiossec: > As it is in master, ByAuthor and similar must be set for Author to be set. If this check is not needed, the explicit setting should be removed. Maybe `public bool ByAuthor { get; private set; }`

In fact, I explicitly take care of setting `ByAuthor` before setting the value (and setting to `false` when it has been unselected).

That's because this test is used for 2 different things ( returning if the filter should be applied and bypassing the set of properties in the following code) that should not behave exactly the same way that cause the bad behavior.
If you return because you are in one of the if, you never update the later properties and do not launch the search (early return `false` even if in fact, you should)

RussKie (commented):
On a surface LGTM, haven't run

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: Works as it should - restores 3.5 behavior, changed below though.

gerhardol:
> Maybe we should uncheck a type of filter if it is "incompatible" (because that is a "AND" logic which is applied between all selected filters) with the one selected.

Change to only allow one option in the combobox

> This PR makes one behavior worse: If you enter text in the searchbox but leave without enter a search is not done (same as before). Search should maybe be done on leaving the box. However, if you switch to another repo the search still occurs even if the searchbox is empty and the Advanced menu is empty.

This was due to "Diff contains" was set which is not visible in the form, only the tooltip. (Change below)

This PR also fixes a regression due to the changed filter handling in 4.0: Only one of the menu options is now used. (So this should be included in 4.0)
However, I believe that change is fine in itself, the Advanced filter form can be used to search more if you really want to - these are AND logic, so it should be better to only allow one by default.
The advanced filter should show these options though.

Updated commit message, will push later, syncing with other changes.

```
squash! No revision filter search when changing filter selection

Handle the case where:
1. Search for a string (by default search in commit message)
2. Select "Diff contains" and unselect "Commit message"
3. Launch a search but search/filtering is not launched

* In 3.5, a revision filter search were done for all checked menu items.
In master only the first checked was used, in the following order: 
Author, Committer, Message, Diff

In general, a user wants to use one of them, multiple options can be
search in the Advanced filter form.
The behavior was therefore changed to only select one of the options.

* ByDiffContent was not part of the criteria for "HasFilter".
Similar should be done for the following, but they should be included in Reset/Summary first:
ShowCurrentBranchOnly, ShowReflogReferences, ShowFirstParent.
(AppSettings.ShowMergeCommits and other handled as settings should probably be excluded.)
These fields are not visible in the Advanced form though

* Remove unused menu item for hash
```

gerhardol:
@pmiossec Are you OK with the changes?

pmiossec:
> @pmiossec Are you OK with the changes?

AFK...
Review of code seems fine.

gerhardol:
Rebased to resolve a merge conflict
Reverted the change for just keeping one checkbox and restored the 3.5 behavior. I still believe this is a good change, but the checkboxes should probably be synced with the Advanced form too.
As there are quite often questions about filters, this need to be improved.
Plan to merge this tomorrow.

gerhardol (approved):

Issue #5371, opened 2018-08-30
Show all stashes in the revision grid


mstv:
**Do you want to request a *feature* or report a *bug*?**
enhancement (low prio, primarily logged as an issue if someone else stumbles on this, too)

**What is the current behavior?**
Multiple stashes (with identical names based on the same revision) are not shown separately in the Revision grid.
In general, Git Extensions shows only the latest stash in the Revision grid (see gerhardol's comment below)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
* checkout (the most recent) branch and modify a file
* stash the changes
![grafik](https://user-images.githubusercontent.com/36601201/44882136-5bda7600-acb2-11e8-81d5-15f4f673dd0b.png)
* modify a file and stash the changes
![grafik](https://user-images.githubusercontent.com/36601201/44882174-8298ac80-acb2-11e8-9cfa-1849973f81c7.png)
* open the stash manager and drop down the list of stashes
![grafik](https://user-images.githubusercontent.com/36601201/44882272-ecb15180-acb2-11e8-9a00-f2921c700182.png)

**What is the expected behavior?**
All stashed should be visible in the Revision grid.
gerhardol: A second enhancement could be to allow for manipulations of stashes in Revision grid.

**Environment you encounter the issue:**

- GitExtensions version: master **and** 2.51.02
- GIT version: 2.17.1.w.2-x64
- OS version: Windows 7
- .NET version: 4.7

**Did this work in previous version of GitExtensions (which)?**
no

gerhardol:
GE shows one stash only, for the latest stash. Other stashes are not visible.
This behavior is not changed in 2.x vs master.
It is kindof a bug, it is shown due to the way Git shows stashes. But the short term fix would be to remove stash in Browse completely, which is not what I want.

So I suggest you change the description to an enhancement:
Show all stashes in the revision grid
(controlled in the same was as e.g. reflog branches)

A second enhancement could be to allow for manipulations of stashes in revision grid.

eric-wieser:
It surprises me that even showing reflog references does not seem to make stashes appear

gerhardol:
TCurrently, a single git-log command is used to get the commits in the revision grid.
It does not seem to be possible to get the stashes in the same command (when searching for it, I found a question from @jbialobr on StackOverflow so this has been attempted before). So stashes without reflog would require a separate command and that the grid is not drawn until the output is merged. So GE would be slower with this option.

[...]

PR #10129, opened 2022-08-04
Cleanup git-log filter option handling

gerhardol:
Follow up to #10076 and #9555
Regression in master

## Proposed changes

Only branches were included with message filter and first stash.
This meant that other refs than branches were not shown in the grid, especially notes/commits. Other exclusions are Gerrit refs like refs/for/ and refs/changes/
A side effect is also that git notes were not shown together with the commits that has notes.

For message filtering this was no big deal, but that deactivating "Show latest stash" also deactivated "Git notes" is a bigger issue.

Note: notes/commits refs should probably never be displayed in the grid (other than with reflog), but that requires more changes.

The filter option handling was very hard to follow, this has been cleaned up.

## Screenshots 

No filter, notes and latest stash shown

image

Latest stash deactivated below

### Before

image

### After

image

## Test methodology 

Tests are updated

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10130, opened 2022-08-04
Eliminate static methods in RevisionReader

gerhardol:
Preparation for #5371 
The end result is pushed to feature/i5371-stash-in-grid
For all stashes (not just the latest)
Not required for 4.0 but should be considered

image

## Proposed changes

- Eliminate static methods in RevisionReader
Prepare RevisionReader for added functionality.
Avoid passing class instance variable with every call.
Parsing revisions may be improved by a milli second in best case.

- Remove RevisionNodeFlags from data grid

Simplify (and a minor optimization) the code when loading revisions.
RevisionNodeFlags has information that is already available in the
GitRevisions, no need to pass that info again.

Set HasNotes after revisions are loaded, that is not required when
displaying a revision.

## Test methodology 

Tests are updated

## Merge strategy

----

### DIFF
--- a/UnitTests/GitCommands.Tests/RevisionReaderTests.cs
+++ b/UnitTests/GitCommands.Tests/RevisionReaderTests.cs
@@ -8,6 +8,7 @@ using ApprovalTests.Reporters;
 using ApprovalTests.Reporters.ContinuousIntegration;
 using FluentAssertions;
 using GitCommands;
+using GitExtUtils;
 using GitUIPluginInterfaces;
 using Newtonsoft.Json;
 using NUnit.Framework;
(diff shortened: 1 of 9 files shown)


gerhardol:
Rebased on #10129 (after multiple changes in master). Squashed the minor fixup.

UnitTests/GitCommands.Tests/RevisionReaderTests.cs
```diff
@@ -115,8 +117,11 @@ public void BuildArguments_check_parameters(RefFilterOptions refFilterOptions, s
         public void TryParseRevisionshould_return_false_if_argument_is_invalid()
         {
             ArraySegment<byte> chunk = null;
+            RevisionReader reader = new(new GitModule(""), _logOutputEncoding, _sixMonths);
```
mstv, line 120: I prefer this over `new(new(""), ...)`. No need to change for testcases.

mstv (approved):
👍

UnitTests/GitCommands.Tests/RevisionReaderTests.cs
gerhardol: I like the compactness and this is tests. But in previous reviews we have often pointed to including the type, so I changed

RussKie (approved):

PR #10135, opened 2022-08-06
Support User local paths for VsCode merge tools

### LINKED ISSUE #10133: GitExtensions is not friendly with vscode as merge tool
### Environment

- Git Extensions 33.33.33
- Build 6a4f9858bfa5d85227f18e0a848dc4259bea8f38
- .NET 6.0.6
- DPI 96dpi (no scaling)

### Issue description

When vscode is in the PATH , you could use:
* ✔`code` or "C:/Users/31594/AppData/Local/Programs/Microsoft VS Code/bin/code" in editor are working (because it resolves to "C:/Users/31594/AppData/Local/Programs/Microsoft VS Code/bin/code" which is a sh script launch by git using sh)
* ✔same values in diff tool are working (because it resolves to "C:/Users/31594/AppData/Local/Programs/Microsoft VS Code/bin/code" which is a sh script launch by git through the `git diff` command)
* ❌**same values in merge tool are NOT working. Because it resolves to "C:/Users/31594/AppData/Local/Programs/Microsoft VS Code/bin/code" which is a sh script that GitExtensions try to launch itself with a `Process.Start` )**

You will have the error:

image

and in the command logs the error:

>System.ComponentModel.Win32Exception (193): An error occurred trying to start process 'C:\Users\31594\AppData\Local\Programs\Microsoft VS Code\bin\code' with working directory 'C:\Taliance\Dev\ArgusTaliance\'. The specified executable is not a valid application for this OS platform.
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at GitCommands.Executable.ProcessWrapper..ctor(String fileName, String prefixArguments, String arguments, String workDir, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding, Boolean useShellExecute, Boolean throwOnErrorExit) in C:\Perso\Dev\github\gitextensions\GitCommands\Git\Executable.cs:line 124

The command run looks like:

```
File name:   C:\Users\31594\AppData\Local\Programs\Microsoft VS Code\bin\code
Arguments:   "code" --wait --merge "C:\Folder\file.cs.REMOTE" "C:\Folder\file.cs.LOCAL" "C:\Folder\file.cs.BASE" "Folder/file.cs"
```

So, the suggested value by GitExtensions `code` is not working and we have 2 options (difficult to understand for a user --it took me long while debugging 😉 )

1. Set the full path toward the exe or the cmd files i.e "C:\Users\31594\AppData\Local\Programs\Microsoft VS Code\Code.exe" or "C:\Users\31594\AppData\Local\Programs\Microsoft VS Code\bin\code.cmd". Very difficult to find 😒
2. Set `code.cmd` in "Path to mergetool" but in this case, GE behave strangely and try to launch as command "Path to mergetool" with arguments "Merge tool command" (so the become misleading and it's no more the command but the arguments) and so if we want to have it woking, we should fill the field with `--wait --merge "$REMOTE" "$LOCAL" "$BASE" "$MERGED"` instead.

I think we will have an issue opened soon (so I created it for us 😁) now that VSCode could be run as a 3way merge tool:

https://code.visualstudio.com/updates/v1_70#_command-line-option-merge

### Steps to reproduce

1. install vscode as per user install and add it in the PATH
2. set `code` as path in merge tool
3. Try to resolve a merge conflict

Don't know.

### Diagnostics

_No response_

gerhardol:
Fixes #10133

## Proposed changes

Add AppData\\Local\\Programs\\/Microsoft\ VS\ Code to install paths for VS Code, the default installation path.
Solution is similar as for SemanticMerge.
One alternative would be to ad "bin" to the paths and change executable to code.cmd

This is not solving the problem reported in #10133, "code" is still not working. "code.cmd" would work though.
However, it handles the user experience adding vscode.

Should be included in 4.0 too.

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitCommands/DiffMergeTools/VsCode.cs
+++ b/GitCommands/DiffMergeTools/VsCode.cs
@@ -1,14 +1,18 @@
+using System;
 using System.Collections.Generic;
+using System.IO;
 
 namespace GitCommands.DiffMergeTools
 {
     internal class VsCode : DiffMergeTool
     {
+        private static readonly string[] Folders = GetFolders();
+
         /// <inheritdoc />
         public override string DiffCommand => "--wait --diff \"$LOCAL\" \"$REMOTE\"";
 
         /// <inheritdoc />
-        public override string ExeFileName => "code.exe";
+        public override string ExeFileName => "Code.exe";
 
         /// <inheritdoc />
         public override string MergeCommand => "--wait \"$MERGED\"";
@@ -17,9 +21,16 @@ namespace GitCommands.DiffMergeTools
         public override string Name => "vscode";
 
         /// <inheritdoc />
-        public override IEnumerable<string> SearchPaths => new[]
+        public override IEnumerable<string> SearchPaths => Folders;
+
+        private static string[] GetFolders()
         {
-            @"Microsoft VS Code\"
-        };
+            string folder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            return new[]
+            {
+                Path.Combine(folder, @"Programs\Microsoft VS Code"),
+                @"Microsoft VS Code\",
+            };
+        }
     }
 }
--- a/GitCommands/DiffMergeTools/VsCode.cs
+++ b/GitCommands/DiffMergeTools/VsCode.cs
@@ -15,7 +15,7 @@ namespace GitCommands.DiffMergeTools
         public override string ExeFileName => "Code.exe";
 
         /// <inheritdoc />
-        public override string MergeCommand => "--wait \"$MERGED\"";
+        public override string MergeCommand => "--wait --merge \"$REMOTE\" \"$LOCAL\" \"$BASE\" \"$MERGED\"";
 
         /// <inheritdoc />
         public override string Name => "vscode";


GitCommands/DiffMergeTools/VsCode.cs
```diff
@@ -1,14 +1,18 @@
...
     internal class VsCode : DiffMergeTool
     {
+        private static readonly string[] Folders = GetFolders();
+
         /// <inheritdoc />
         public override string DiffCommand => "--wait --diff \"$LOCAL\" \"$REMOTE\"";
 
         /// <inheritdoc />
-        public override string ExeFileName => "code.exe";
+        public override string ExeFileName => "Code.exe";
```
RussKie, line 15: Why is this necessary?

RussKie (approved):

GitCommands/DiffMergeTools/VsCode.cs
gerhardol: Not necessary, but this is what a code installs. I was tricked by this when searching for the executable. 

Note: I was not sure if @pmiossec wanted to change the expression or not. It seems to work now at least.

gerhardol:
`--merge` is in 1.170 which is the July release, so the changed options should be OK to merge

PR #10137, opened 2022-08-07
Do not start NBug if mergetool exits with other than 0

gerhardol:
## Proposed changes

Exit code is other than 0 if user aborts a merge.
#9056 is a step in the right direction, this is another situation where other exit codes than 0 are OK though.
Regression in master, to be included in 4.0

Occurs in FormResolve with _Start mergetool_

image

image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -923,7 +923,7 @@ namespace GitCommands
                 fileName.ToPosixPath().QuoteNE()
             };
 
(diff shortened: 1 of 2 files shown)


RussKie (approved):

mstv:
Did it really start the BugReporter? I think it has displayed the git error message and defaulted to ignore it.

gerhardol:
> Did it really start the BugReporter? I think it has displayed the git error message and defaulted to ignore it.

The popup first, that you could ignore.
But this error exit is normally something that the user action causes and should be allowed

PR #9303, opened 2021-06-21
Open Delete Tag dialog from Left Panel with full tag name

mstv:
Fixes #9291

## Proposed changes

- Prepopulate Delete Tag dialog with the `FullName` of the tag

### After

grafik

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 55dec49de1d4b378f27687d36b47b90ab8c6250d
- Git 2.31.1.windows.1
- Microsoft Windows NT 10.0.19042.0
- .NET 5.0.7
- DPI 96dpi (no scaling)

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #5785, opened 2018-11-16
Stash operations directly from revision grid


drewnoakes:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

Right-clicking a stash in the revision grid shows:

![image](https://user-images.githubusercontent.com/350947/48617851-735efc00-e98f-11e8-9fc8-da9cae5aeedf.png)

**What is the expected behavior?**

User should have option of dropping, applying and viewing a stash from here.

Current workaround is to go to the stashes UI, try to find the corresponding stash (this correlation can be hard if you have many stashes) and operate on it from there. It's error prone and inconvenient.

**Environment you encounter the issue:**

- Git Extensions 3.00.00.03-RC2
- 8cbcd2b25223c73fbe9f7dad08e498cf82734bfa  (Dirty)
- Git 2.19.1.windows.1
- Microsoft Windows NT 10.0.17763.0
- .NET Framework 4.7.3190.0
- DPI X:150.00% Y:150.00%

**Did this work in previous version of GitExtensions (which)?**

I don't believe so.

mstv:
GE shows only the latest stash in the Revision Grid, refer to [Gerhard's comment](https://github.com/gitextensions/gitextensions/issues/5371#issuecomment-417574124).

vbjay:
> GE shows only the latest stash in the Revision Grid, refer to Gerhard's comment.

That is a bug.  We should show all stashes like we did before.  

![image](https://user-images.githubusercontent.com/3629489/50563346-1d99a580-0cea-11e9-9686-9ca0e55c32cc.png)

gerhardol:
No change for stashes in 2.x and 3.0: By default only latest stash shown. That stash is shown at all is due to that it is in refs/. This should probably be removed.

All stashes are shown if reflogs are shown in revision grid. You cannot manipulate stashes directly but compare files, reset etc.

vbjay:
Actual it is a change.  Stashes would show up on top of the commit they were based on before.

gerhardol:
>Actually it is a change. Stashes would show up on top of the commit they were based on before.

Not compared to 2.51

No reflogs

![image](https://user-images.githubusercontent.com/6248932/50564658-01682980-0d27-11e9-96bc-bf40dd98e310.png)

reflogs

![image](https://user-images.githubusercontent.com/6248932/50564666-19d84400-0d27-11e9-99e8-7ce1c6a7ab6c.png)

[...]

Issue #5460, opened 2018-09-20
Show stashes tree in sidebar


NN---:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
No tags and stahes in sidebar

- GitExtensions version:
3.0
- GIT version: 
2.9
- OS version: 
Windows
- .NET version: 
4.7

Is it possible to have something similar to SourceTree tags and stashes in sidebar ?
![](https://www.sourcetreeapp.com/dam/jcr:580c367b-c240-453d-aa18-c7ced44324f9/hero-mac-screenshot.png)

pmiossec:
Tags are already there but you should enable it in the "gear" displayed near the search field.

I absolutely agree for stashes.

NN---:
Cannot find it.
Shouldn't it be by default enabled ?

NN---:
I don't see this in GitExtensions 3.0 Beta

pmiossec:
Really strange. It should be there...
You don't have the gear displayed? Or it is empty?
I don't think there is a setting for that....

NN---:
Oh sorry.
It is there, I just didn't notice this one.
Thanks !

francoas:
I would suggest to enable it by default

vbjay:
It should be enabled and remembered. I hate having to hit show tags each
time I load GE.

francoas:
I agree with you: it should be enabled by default (no need for the gear, unless you want the ability to configure also Branches and Remotes nodes)
Now it looks like it is remembered, but it should also remember the "open" or "closed" of the node. 
I also suggest that tags could be "reverse" ordered (just see the gitextensions project tags to understand why)

gerhardol:
For stashes in the sidebar to be meaningful, the stashes need to be displayed in the revision grid too.
This is handled in #5371.

PR #10138, opened 2022-08-07
Show Stash in rev grid

### LINKED ISSUE #5371: Show all stashes in the revision grid

### LINKED ISSUE #9873: Add stash options for context menu
### Feature description

The idea of context menu as I understand it is to have alternative interaction options with given item in most-used first order.

When right-clicking a stash on the tree, the first options should be to:

- pop this stash - allows for easily finding relevant stashes in tree history instead of doing typical error-prone pop or finding the stash through the dedicated dialog
- apply this stash (apply diff without deleting stash)
- delete this stash - allows for stash cleanup as you go instead of finding them in separate dialog and deleting them there.

I'm in the process of migrating from GitKraken and lack of this little UX feature makes using Git Extensions a bit tedious. :(

### Environment

- Git Extensions 3.5.4.12724
- Git 2.35.1.windows.2
- .NET Framework 4.8.4470.0
- DPI 96dpi (no scaling)

gerhardol:
Resolves #5371 (and #5785
Fixes #9873

## Proposed changes

Show stash commits in the revision grid, with menu options to pop and drop stashes.

A stash is implemented in Git with 3 internal commit.
Some more explanations in GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:988
The implementations I have seen do not expose the third commit with untracked files.
The PR includes this "untracked" commit (empty "untracked" are filtered) for the first 10 stashes (see Performance below for details).
This setting is read from the settings but no setter is exposed, mostly because there is no natural place to show it (maybe in View is the best, but other similar settings are in General, that is considered full).

The implementation by default only show the stashes where the parents are shown in the grid.
So for instance if you stash and then rebase then the stash is not shown (but see below).
Unless reflogs are shown, only stashes that are related to commits visible in the grid is shown.
(Another difference with reflog is that stashes are presented in Git order, the "inserted" stashes are always shown related to the main commit).

The label for stash commits is partly for this reason set to the reflog selector like "stash@{0}" (required to pop/drop stashes) (another reason is to not get lost too easy when looking at commits, especially when many stashes are created for the same commit).
(For instance GitKraken shows these commits.)
FormStash can still be used to see these, they could also be added to the side panel (similar to SourceTree) (but the contents would not be shown). See #5460.
(Sublime merge shows both in grid and sidepanel)

GE currently always show the latest stash, this was expected by mstv in the initial version.
The latest stash is therefore always shown if stashes are shown.
This commit will always include the index and untracked commits.
See https://github.com/gitextensions/gitextensions/pull/10138#issuecomment-1236200583 for a discussion about how that could be changed.

There is no new menu operation to create a stash, use stash button or Ctrl-Alt-Up to create a stash.
Create a stash operate on both Index and WorkTree commits, so there is no natural way to connect this to a revision (but _partial_ commit should be possible to create in RevDiff).
FormStash was improved in #9303 too.

### Performance

Loading revisions in 4.0 is heavily optimized, some performance measurements to ensure there is no regression.
(I have tried to make sure the addition has minimal impact on performance).

My GE repo has 19400 commits and 542 stashes, testing on Win10 with Zen2600, 16GB, SSD.
I consider 542 stashes to be an upper limit to what is reasonable to test with.

If stashes are not displayed, basically one null variable is, adding less than 1 ms to process all revisions. So usecases not showing stashes are not affected.
Enabled but no existing stashes has a few more checks, affects loading with about 9 ms.

Times are approximate, some variation (+/- 30 ms), but relative consistent differences.
Time will be affected by number of cores etc
Refreshing refs is done in three partly parallel steps:

1. Getting refs and checkout
Total step 175 ms (git-for-each-ref 125 ms)

2. Getting Stash commit (10 stash checked for untracked)
Total step 175 ms (git-stash-list 95 ms, git-log <untracked-id> 70 ms)
If this time is higher than step 1 the load time is increasing.
Note that the second Git command for untracked is not needed if reflog is shown.

3. Getting revisions
Total git-log time 600 ms but first revision is delivered and could be processed for display after 140 ms (this is when the spinner is replaced with "Loading" in upper right corner) (as refs are slower, revgrid will be updated 35 ms after this)
(if sorting in the grid is enabled, revisions will be shown with at least a 100 ms delay)
Processing time to find if stashes should be added is about 9 ms.
The GUI update is done in about the time that git-log runs.
The "Loading" label is then removed, but some additional CPU may be needed for selecting a new revision (if filtering changed).

Stashes slow down showing revisions in the grid with about 9 ms plus possible additional time due to parallel processing overhead (but I have not seen any faster ref handling with stash disabled).
This is assuming that the 10 "untracked" takes the same time as getting refs, which in my case is 10, which I also believe is an OK value of untracked to show.
0 untracked limits step 2 with 60 ms (but revisions are not shown earlier), 50 untracked adds 30 ms and all 542 adds 400 ms.

Brief tests with Linux repo in wsl limited to 100000 commits shown with one local branch and 10 stashes.
With so few refs, refs are handled in 150 ms and stash-list in 170 and listing untracked will add 100 ms and 50 ms (processing adding to grid) out of total 4.1 s. So hardly noticeable.

## Screenshots 

With reflogs shown, the change is that labels and menu operations are enabled for all commits.

### Before

Latest stash shown

image

### After

If parent commit is not shown, stash is not shown (first is shown though, not updated example)

image

Other commits are shown, "untracked" commit only if it is not empty

image

Menu

image

## Test methodology 

A few tests added, some modified.

## Test environment(s) 

Git 2.37.1
No other version tested, nothing seen in the Git doc about any compatibility.

## Merge strategy

----

### DIFF
--- a/GitCommands/RevisionReader.cs
+++ b/GitCommands/RevisionReader.cs
@@ -42,26 +42,107 @@ namespace GitCommands
         private readonly Encoding _logOutputEncoding;
         private readonly long _sixMonths;
 
-        public RevisionReader(GitModule module)
-            : this(module, module.LogOutputEncoding, new DateTimeOffset(DateTime.Now.ToUniversalTime() - TimeSpan.FromDays(30 * 6)).ToUnixTimeSeconds())
+        // reflog selector to identify stashes
+        private readonly bool _hasReflogSelector;
+
+        private string LogFormat => _hasReflogSelector ? FullFormat.Replace("%B", "%gD%n%B") : FullFormat;
+
+        public RevisionReader(GitModule module, bool hasReflogSelector)
+            : this(module, hasReflogSelector, module.LogOutputEncoding, new DateTimeOffset(DateTime.Now.ToUniversalTime() - TimeSpan.FromDays(30 * 6)).ToUnixTimeSeconds())
         {
         }
 
-        internal RevisionReader(GitModule module, Encoding logOutputEncoding, long sixMonths)
+        private RevisionReader(GitModule module, bool hasReflogSelector, Encoding logOutputEncoding, long sixMonths)
         {
             _module = module;
+            _hasReflogSelector = hasReflogSelector;
             _logOutputEncoding = logOutputEncoding;
             _sixMonths = sixMonths;
         }
 
-        public async Task GetLogAsync(
+        /// <summary>
+        /// Get the git-stash GitRevisions.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation cancellationToken.</param>
+        /// <returns>List with GitRevisions.</returns>
+        public IReadOnlyCollection<GitRevision> GetStashes(CancellationToken cancellationToken)
+        {
+            Debug.Assert(_hasReflogSelector, "_hasReflogSelector must be set to get the reflog selectors (to identify stashes)");
+
+            GitArgumentBuilder arguments = new("stash")
+            {
+                "list",
+                "-z",
+                $"--pretty=format:\"{LogFormat}\""
+            };
+
+            return GetRevisionsFromArguments(arguments, cancellationToken);
+        }
+
+        /// <summary>
+        /// Get the GitRevisions for the listed files, excluding commits without changes.
+        /// </summary>
+        /// <param name="untracked">commit list.</param>
+        /// <param name="cancellationToken">Cancellation cancellationToken.</param>
+        /// <returns>List with GitRevisions.</returns>
+        public IReadOnlyCollection<GitRevision> GetRevisionsFromList(IList<ObjectId> untracked, CancellationToken cancellationToken)
+        {
+            if (untracked.Count == 0)
+            {
+                return Array.Empty<GitRevision>();
+            }
+
+            GitArgumentBuilder arguments = new("log")
+            {
+                "-z",
+                $"--pretty=format:\"{LogFormat}\"",
+                "--dirstat=files",
+                string.Join(" ", untracked),
+                ".",
+            };
+
+            return GetRevisionsFromArguments(arguments, cancellationToken);
+        }
+
+        /// <summary>
+        /// Get the GitRevisions for Git argument.
+        /// </summary>
+        /// <param name="arguments">Git command arguments.</param>
+        /// <param name="cancellationToken">Cancellation cancellationToken.</param>
+        /// <returns>List with GitRevisions.</returns>
+        private IReadOnlyCollection<GitRevision> GetRevisionsFromArguments(GitArgumentBuilder arguments, CancellationToken cancellationToken)
+        {
+            List<GitRevision> stashes = new();
+
+            using (IProcess process = _module.GitCommandRunner.RunDetached(arguments, redirectOutput: true, outputEncoding: GitModule.LosslessEncoding))
+            {
+                byte[] buffer = new byte[4096];
+
+                foreach (ArraySegment<byte> chunk in process.StandardOutput.BaseStream.ReadNullTerminatedChunks(ref buffer))
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    if (TryParseRevision(chunk, out GitRevision? revision))
+                    {
+                        stashes.Add(revision);
+                    }
+                }
+            }
+
+            return stashes;
+        }
+
+        /// <summary>
+        /// Get the git-log revisions for the revision grid.
+        /// </summary>
+        /// <param name="subject">Observer to update the revision grid when the revisions are available.</param>
+        /// <param name="arguments">git-log arguments.</param>
+        /// <param name="cancellationToken">Cancellation cancellationToken.</param>
+        public void GetLog(
             IObserver<GitRevision> subject,
             ArgumentBuilder arguments,
-            CancellationToken token)
+            CancellationToken cancellationToken)
         {
-            await TaskScheduler.Default;
-            token.ThrowIfCancellationRequested();
-
 #if DEBUG
             int revisionCount = 0;
             var sw = Stopwatch.StartNew();
@@ -72,13 +153,13 @@ namespace GitCommands
 #if DEBUG
                 Debug.WriteLine($"git {arguments}");
 #endif
-                token.ThrowIfCancellationRequested();
+                cancellationToken.ThrowIfCancellationRequested();
 
                 var buffer = new byte[4096];
 
                 foreach (var chunk in process.StandardOutput.BaseStream.ReadNullTerminatedChunks(ref buffer))
                 {
-                    token.ThrowIfCancellationRequested();
+                    cancellationToken.ThrowIfCancellationRequested();
 
                     if (TryParseRevision(chunk, out GitRevision? revision))
                     {
@@ -95,7 +176,7 @@ namespace GitCommands
 #endif
             }
 
-            if (!token.IsCancellationRequested)
+            if (!cancellationToken.IsCancellationRequested)
             {
                 subject.OnCompleted();
             }
@@ -114,7 +195,7 @@ namespace GitCommands
             {
                 { maxCount > 0, $"--max-count={maxCount}" },
                 "-z",
-                $"--pretty=format:\"{FullFormat}\"",
+                $"--pretty=format:\"{LogFormat}\"",
                 { AppSettings.RevisionSortOrder == RevisionSortOrder.AuthorDate, "--author-date-order" },
                 { AppSettings.RevisionSortOrder == RevisionSortOrder.Topology, "--topo-order" },
                 { refFilterOptions.HasFlag(RefFilterOptions.Boundary), "--boundary" },
@@ -298,6 +379,7 @@ namespace GitCommands
             var authorEmail = reader.ReadLine();
             var committer = reader.ReadLine();
             var committerEmail = reader.ReadLine();
+            var reflogSelector = _hasReflogSelector ? reader.ReadLine() : null;
 
             bool skipBody = _sixMonths > authorUnixTime;
             (string? subject, string? body, bool hasMultiLineMessage) = reader.PeekSubjectBody(skipBody);
@@ -328,7 +410,8 @@ namespace GitCommands
                 Subject = subject,
                 Body = body,
                 HasMultiLineMessage = hasMultiLineMessage,
-                HasNotes = false
+                HasNotes = false,
+                ReflogSelector = reflogSelector,
             };
 
             return true;
@@ -416,6 +499,11 @@ namespace GitCommands
             internal TestAccessor(RevisionReader revisionReader)
             {
                 _revisionReader = revisionReader;
+            }
+
+            internal static RevisionReader RevisionReader(GitModule module, bool hasReflogSelector, Encoding logOutputEncoding, long sixMonths)
+            {
+                return new RevisionReader(module, hasReflogSelector, logOutputEncoding, sixMonths);
             }
 
             internal bool TryParseRevision(ArraySegment<byte> chunk, [NotNullWhen(returnValue: true)] out GitRevision? revision) =>
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1085,10 +1085,16 @@ namespace GitCommands
             set => SetBool("showReflogReferences", value);
         }
 
-        public static bool ShowLatestStash
+        public static bool ShowStashes
         {
-            get => GetBool("showLatestStash", true);
-            set => SetBool("showLatestStash", value);
+            get => GetBool("showStashes", true);
+            set => SetBool("showStashes", value);
+        }
+
+        // Set manually in settings file
+        public static int MaxStashesWithUntrackedFiles
+        {
+            get => GetInt("maxStashesWithUntrackedFiles", 10);
         }
 
         public static bool ShowSuperprojectTags
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
@@ -29,6 +29,10 @@ namespace GitUI
             this.stopBisectToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.bisectSeparator = new System.Windows.Forms.ToolStripSeparator();
             this.copyToClipboardToolStripMenuItem = new GitUI.UserControls.RevisionGrid.CopyContextMenuItem();
+            this.stashStripSeparator = new System.Windows.Forms.ToolStripSeparator();
+            this.applyStashToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.popStashToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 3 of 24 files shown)


mstv:
I have not had a look at the details yet.
Though the current approach does not display (the most recent) stash if it is based on an invisible commit, e.g. "disappears" after a rebase.
The stash index is shown also only if the reflog is shown.

gerhardol:
Latest stash can always be included (it is explicitly excluded now), but then the latest stash will always look like 
image

But maybe a fixed number of stashes can always be added ("stash@{0} stash@{1} stash@{2} ... stash@{20}") by default.
Then there would be a smaller delay then if stashes have to be queried first (so maybe 20 ms extra instead of 200 ms in my GE clone).

The full info should be seen in the side panel (not implemented).

gerhardol:
Some measurements:
git-log <stash@{} for 542 stashes> adds about 550 ms to the listing of the first commit. Limiting to 20 adds about 20 ms. (So about 1 ms/commit. So if "force include" is used, the list should be limited. This will require a hard to explain configuration. I canparse about 2000 stashes, then the command line get too long.

To avoid adding the "index" commit and empty untracked, processing time added is about 10 ms, similar handling as for inserting untracked commits

mstv:
> Latest stash can always be included (it is explicitly excluded now), but then the latest stash will always look like the following, unless there are some funky code to remove the index and empty untracked commits. image 

That's how it used to be. Some of this info is rarely really needed indeed. Though stashes can be not just one commit. So, alright for me.

> But maybe a fixed number of stashes can always be added ("stash@{0} stash@{1} stash@{2} ... stash@{20}") by default.

Will Git complain if there are fewer stashes?

> Then there would be a smaller delay then if stashes have to be queried first (so maybe 20 ms extra instead of 200 ms in my GE clone).

👍
I think 10 would suffice.

gerhardol:
> > But maybe a fixed number of stashes can always be added ("stash@{0} stash@{1} stash@{2} ... stash@{20}") by default.
> 
> Will Git complain if there are fewer stashes?

No, that is fine.

> > Then there would be a smaller delay then if stashes have to be queried first (so maybe 20 ms extra instead of 200 ms in my GE clone).
> 
> 👍 I think 10 would suffice.

I have a problem explaining this setting, so not keen on it.
As `rev-parse stash@{0} etc` can take long time, the complete stash listing should probably be done before listing in the grid, always delaying commit presentation.
(Except for reflog when the stashes are added anyway)

mstv:
> I have a problem explaining this setting, so not keen on it.

OK, so explicitly request just the most recent stash.

RussKie:
Perhaps worth considering bumping the stash'related menus to the top:
image

gerhardol:
> > I have a problem explaining this setting, so not keen on it.
> 
> OK, so explicitly request just the most recent stash.

Changed to show the latest stash if ShowStashes

If  (currently hidden) and `MinimumStashesAlwaysInGrid` are added in settings, the descriptions could be something like below.
Likely a follow up. Similar with stash in sidepanel.

image

`ShowStashes`
(Checkbox as in _View_ menu)

`MinimumStashesAlwaysInGrid` (default 1)
Stashes in the grid are by default shown in the grid if the parent commit is visible in the revision grid.
(If reflogs are shown, all commits including all stashes are shown.)
This setting will ensure that the stashes are added to the grid.
Note that the _index_ and _untracked_ commits are always included.
Note: Setting this value to a high value may delay the display of commits.

`MaxStashesWithUntrackedFiles` (default 10)
By default only `MinimumStashesAlwaysInGrid` stashes will include the stash commit with untracked files.
This setting will include the untracked commit (unless it is empty) for additional stashes.
Note: Setting this value to a high value may delay the display of commits.

gerhardol:
squashed, rebased (mstv had pulled the latest updates so they could be reviewed...) and updated the description, removed draft.

RussKie (commented):
A quick review, will need more time for a more in depth review

GitCommands/RevisionReader.cs
```diff
@@ -42,18 +42,107 @@ public sealed class RevisionReader
...
+
+                    if (TryParseRevision(chunk, out GitRevision? revision))
+                    {
+                        stashes.Add(revision);
+                    }
+                }
+            }
+
+            return stashes;
+        }
```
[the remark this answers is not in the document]
gerhardol: This is leading revisions, just a different Git command than the git-log currently in the class.
So an instance of RevisionReader is needed regardless.

GitCommands/RevisionReader.cs
```diff
@@ -42,18 +42,107 @@ public sealed class RevisionReader
...
+
+        /// <summary>
+        /// Get the GitRevisions for the listed files,
+        /// excluding commits without changes.
+        /// </summary>
+        /// <param name="untracked">commit list</param>
+        /// <param name="token">Cancellation token</param>
+        /// <returns>List with GitRevisions</returns>
+        public List<GitRevision> GetRevisionsFromList(IList<ObjectId> untracked, CancellationToken token)
```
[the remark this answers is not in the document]
gerhardol: We can settle on cancellationToken, starting with this PR

GitCommands/RevisionReader.cs
```diff
@@ -42,18 +42,107 @@ public sealed class RevisionReader
...
+                    }
+                }
+            }
+
+            return stashes;
+        }
+
+        /// <summary>
+        /// Get the GitRevisions for the listed files,
+        /// excluding commits without changes.
```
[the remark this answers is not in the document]
gerhardol: ok

GitCommands/RevisionReader.cs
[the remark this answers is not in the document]
gerhardol: What is the benefit with this?
I find it easier to read checking the size (and slightly more efficient even if that is not a focus here).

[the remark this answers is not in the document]
gerhardol: Changed the existing code in this module
About 100 more occurrences, excluding tests, translations and externals

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
```diff
@@ -189,6 +193,22 @@ private void InitializeComponent()
...
+            this.popStashToolStripMenuItem.Size = new System.Drawing.Size(264, 24);
+            this.popStashToolStripMenuItem.Text = "Pop &stash...";
+            this.popStashToolStripMenuItem.Click += new System.EventHandler(this.PopStashToolStripMenuItemClick);
+            // 
+            // dropStashToolStripMenuItem
+            // 
+            this.dropStashToolStripMenuItem.Image = global::GitUI.Properties.Images.Stash;
+            this.dropStashToolStripMenuItem.Name = "dropStashToolStripMenuItem";
+            this.dropStashToolStripMenuItem.Size = new System.Drawing.Size(264, 24);
+            this.dropStashToolStripMenuItem.Text = "&Drop stash...";
```
mstv, line 209: These dots imply there was a confirmation dialog. But it is missing. Add the optional confirmation, please. We already have a regarding AppSetting (which I do not turn off).
Dropped stashes are hard to recover. They do not appear in the RefLog.
After `recovering` the `lost object`, the commit is filtered out somehow as long as RefLog references are shown.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -1154,6 +1214,32 @@ void OnRevisionRead(GitRevision revision)
...
+                if (stashesById is not null)
+                {
+                    if (stashesById.TryGetValue(revision.ObjectId, out GitRevision gridStash))
+                    {
+                        revision.ReflogSelector = gridStash.ReflogSelector;
+
+                        // Do not add this again (when the main commit is handled)
+                        stashesById.Remove(revision.ObjectId);
+                    }
+                    else if (stashesByParentId?.Contains(revision.ObjectId) ?? false)
```
mstv, line 1226: Nit: I would prefer
```suggestion
                    else if (stashesByParentId?.Contains(revision.ObjectId) is true)
```

GitCommands/Settings/AppSettings.cs
```diff
@@ -1085,10 +1085,17 @@ public static bool ShowReflogReferences
...
-            set => SetBool("showLatestStash", value);
+            get => GetBool("showStashes", true);
+            set => SetBool("showStashes", value);
+        }
+
+        // No official setting
+        public static int MaxStashesWithUntrackedFiles
+        {
+            get => GetInt("maxStashesWithUntrackedFiles", 10);
+            set => SetInt("maxStashesWithUntrackedFiles", value);
```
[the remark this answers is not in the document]
gerhardol: changing this here too, as well as aligning the comment

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
gerhardol: OK
Making a temporary change to hide the option for now

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: OK

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
mstv: > Making a temporary change to hide the option for now

OK. We can activate it in a follow-up - as well as add `Apply stash`.

mstv:
>> After `recovering` the `lost object`, the commit is filtered out somehow as long as RefLog references are shown.

The restored commit "LOST_FOUND_1" (and the index commit) is not returned by
`git -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B"`
neither with nor without `--reflog`.
How is it found and added if **not** showing RefLog references?
The base commit c9b8c5d is only returned *with* `--reflog`. Weird!
 
image

GitCommands/RevisionReader.cs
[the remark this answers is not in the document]
RussKie: We constantly reading revisions, and this code is on the hot path. So we want to avoid memory churn as much as possible.

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
gerhardol: "stash apply" is basically the same command as git-cherry-pick so it is not really needed. stash apply can be applied to any commit.

But as the menu is only for stash I guess it is not much in the way.

To be clear: I will remove the stash-drop (unless I get around to create a confirmation) before merging this if approved.

GitCommands/RevisionReader.cs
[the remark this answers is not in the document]
gerhardol: I did not see any improvement for allocation in your proposal, just a slower check.

For performance and allocation when loading, the important path is the reading of revisions (in RevisionReader that is optimized with my best effort) and in RevGrid (can likely be a few tuning done). The general improvement is to avoid calling Git when not needed.
I am adding an if and a null allocation in this PR to the rev reading loop, which will add ms.
(Replacing String.Empty with "" and removing precompiled regex is a more obvious optimization...)

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
mstv: > I will remove the stash-drop (unless I get around to create a confirmation) before merging this if approved.

It has already been implemented in FormStash - with optional confirmation.
If you omit this menu item, I will add it myself.

> "stash apply" is basically the same command as git-cherry-pick.

Is this also true for complex stashes, e.g. with untracked files?

gerhardol: I added _Apply stash_ as well as a confirmation for drop stash
(Same keyboard accelerator for stash-apply as for cherry-pick)

> Is this also true for complex stashes, e.g. with untracked files?

The dialog lists the untracked commit but do add the files. (index commit files are included, but they are included in the main commit already). Could be an issue in the dialog, not investigated (do not plan to).

I just did not see the usefulness of appy stash myself, but if you ask for it, other may do that too.
Did not want to add too many bells and whistles before I know if it will be accepted.

GitCommands/RevisionReader.cs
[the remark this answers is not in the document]
gerhardol: I made a simple refactoring and changed this code slightly.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2152,6 +2246,71 @@ private void CherryPickCommitToolStripMenuItemClick(object sender, EventArgs e)
...
+
+        private void PopStashToolStripMenuItemClick(object sender, EventArgs e)
+        {
+            var stashName = LatestSelectedRevision.ReflogSelector;
+            if (string.IsNullOrEmpty(stashName))
+            {
+                return;
+            }
+
+            UICommands.StashPop(this, LatestSelectedRevision.ReflogSelector);
```
mstv, line 2263: nit:
```suggestion
            string stashName = LatestSelectedRevision.ReflogSelector;
            if (string.IsNullOrEmpty(stashName))
            {
                return;
            }

            UICommands.StashPop(this, stashName);
```

mstv (approved):
👍 Thank you
Fixes #9873

gerhardol:
Plan to squash and rebase tomorrow
@RussKie @pmiossec Any objections to merging and including in 4.0?

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
```diff
@@ -120,6 +123,9 @@ private void InitializeComponent()
             this.stopBisectToolStripMenuItem,
             this.bisectSeparator,
             this.copyToClipboardToolStripMenuItem,
+            this.applyStashToolStripMenuItem,
```
RussKie, line 126: Please add a separator after the "Copy to clipboard" menu item.

RussKie (commented):
Looked through on a phone in a browser. I don't have any major objections. Supportive of taking it in to the release.

GitUI/UserControls/RevisionGrid/RevisionGridControl.Designer.cs
gerhardol: Changed this in the latest commit before squashing and updating the commit message.

gerhardol:
Plan to merge this tomorrow

PR #10139, opened 2022-08-08
VS Code build support

gerhardol:
## Proposed changes

Add basic VS Code support to build, test and run GE.
Also recommending plugins.
I personally would like a .code-workspace file too, but current settings can be stored in the folder and there are no folder external dependencies, so this works.

Not much use myself, but the current configuration is broken, this is something to improve on.
VS Code is used more and more, even if a Windows only app as GE is better supported in Visual Studio, this may help occasional users.

Also for 4.0, if users try to build/debug the release.

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+    "recommendations": [
+            "ms-vscode.cpptools",
+            "ms-dotnettools.csharp",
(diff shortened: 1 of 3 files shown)


RussKie (approved):
Wow 🚀

PR #10145, opened 2022-08-10
Open new instance if double click on current submodule

gerhardol:
https://github.com/gitextensions/gitextensions/pull/10118#issuecomment-1211078704

## Proposed changes

Add menu item to open the current submodule (including the
superproject module) instead of reopening the module
(this option was not added to the menu).

Add the menu item to open the new instance too, first in the menu.

### Before

image

image

### After

image

image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitUI/BranchTreePanel/SubmoduleNode.cs
+++ b/GitUI/BranchTreePanel/SubmoduleNode.cs
@@ -49,8 +49,6 @@ namespace GitUI.BranchTreePanel
             ApplyStatus();
         }
 
-        public bool CanOpen => !IsCurrent;
-
         protected override string DisplayText()
         {
             return SubmoduleName + BranchText + Info.Detailed?.AddedAndRemovedText;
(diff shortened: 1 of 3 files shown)


GitUI/BranchTreePanel/SubmoduleNode.cs
```diff
@@ -101,7 +99,14 @@ internal override void OnSelected()
...
         {
-            Open();
+            if (!IsCurrent)
+            {
+                Open();
+            }
+            else
+            {
+                LaunchGitExtensions();
+            }
```
mstv, line 109: I prefer positive logic although the first branch handles the special case then:
```suggestion
            if (IsCurrent)
            {
                LaunchGitExtensions();
            }
            else
            {
                Open();
            }
```

mstv (commented):
LGTM

GitUI/BranchTreePanel/SubmoduleNode.cs
gerhardol: ok

gerhardol: added a comment too
Note that the approved #10146 depends on this PR, can it be approved?

PR #10153, opened 2022-08-17
Do not call DoDragDrop when mouse is up

huangsunyang:
## Proposed changes

- In FileStatusList, DoDragDrop is called even when mouse (left button) is up, which is unnecessary
- Besides, in some windows versions, IME has a [bug](https://support.microsoft.com/en-us/topic/kb4564002-you-might-have-issues-on-windows-10-version-20h2-and-windows-10-version-2004-when-using-some-microsoft-imes-63696506-47d2-9997-0b72-41a68e328692) that DoDragDrop call will hang up. And Git Extensions is always calling DoDragDrop, thus increases the possibilily of this bug

## Test methodology 

- In FileStatusList, shift/ctrl click function remains same
- In FileStatusList, drag drop function remains same

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -1562,7 +1562,7 @@ namespace GitUI
 
         private void FileStatusListView_MouseUp(object sender, MouseEventArgs e)
         {
-            // Stop DRAG when mouse up
+            // Release the drag capture
             if (e.Button == MouseButtons.Left)
             {
                 _dragBoxFromMouseDown = Rectangle.Empty;
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -1560,6 +1560,15 @@ namespace GitUI
             }
         }
 
+        private void FileStatusListView_MouseUp(object sender, MouseEventArgs e)
+        {
+            // Stop DRAG when mouse up
+            if (e.Button == MouseButtons.Left)
+            {
+                _dragBoxFromMouseDown = Rectangle.Empty;
+            }
+        }
+
         private void FileStatusListView_MouseMove(object sender, MouseEventArgs e)
         {
             ListView? listView = sender as ListView;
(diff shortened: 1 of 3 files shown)


gerhardol (commented):
+1
This seem to work for me.
If I understand the reasoning correctly, MouseUp in FileStatusList means that the mouse has not left the control and DnD should be cancelled. No MouseUp if released outside FileStatusList.
How is DnD cancelled when releasing Mouse before this PR? (Is it necessary?)

huangsunyang:
@gerhardol As I know, DnD won't be cancelled on releasing Mouse before this PR... Set a breakpoint at DoDragDrop, after a click (which gives _dragBoxFromMouseDown a value), every mouse move event will call DoDragDrop (even when mouse up).

I think it's quite necessary not to call DnD when mouse is up, though Git Extension is working well for most users. I pay attention to this problem because some of my colleagues confront the ui hang up problem mentioned in the PR. It's not a big problem, but it's not right :(

GitUI/UserControls/FileStatusList.cs
```diff
@@ -1560,6 +1560,15 @@ private void FileStatusListView_MouseDown(object sender, MouseEventArgs e)
             }
         }
 
+        private void FileStatusListView_MouseUp(object sender, MouseEventArgs e)
+        {
+            // Stop DRAG when mouse up
```
RussKie, line 1565:
```suggestion
            // Release the drag capture
```

RussKie (approved):
Looks plausible. Haven't run.

gerhardol:
> @gerhardol As I know, DnD won't be cancelled on releasing Mouse before this PR... Set a breakpoint at DoDragDrop, after a click (which gives _dragBoxFromMouseDown a value), every mouse move event will call DoDragDrop (even when mouse up).

When you release the left mouse button, the mouse icon changes back immediately so it seems like it is cancelled.

> I think it's quite necessary not to call DnD when mouse is up, though Git Extension is working well for most users. I pay attention to this problem because some of my colleagues confront the ui hang up problem mentioned in the PR. It's not a big problem, but it's not right :(

Probably so, but I do not know the behavior well enough to just accept everything.

Is the same needed in GitUI\CommandsDialogs\RevisionFileTreeControl.cs?
I assume no, called in ItemDrag.

huangsunyang:
oh yes drag is cancelled when mouse released, maybe a mouse down detect is doing inside DoDragDrop.

I agree with you about the DoDragDrop in RevisionFileTreeControl.cs

gerhardol:
> oh yes drag is cancelled when mouse released, maybe a mouse down detect is doing inside DoDragDrop.

Then the Windows bug may be to not detect the MouseUp?
Then the description of the issue makes sense and the change should not cause any problems either.

gerhardol (approved):

mstv:
@huangsunyang, could you add your sign off to `contributors.txt`, please?

huangsunyang:
> @huangsunyang, could you add your sign off to `contributors.txt`, please?

(maybe) done, my pleasure

mstv (approved):
Thank you

PR #10159, opened 2022-08-19
Settings: Don't clear setting search when leaving the search box

### LINKED ISSUE #9706: "Type to find" in Settings (Search function) does not remember query string or results
### Feature description

When using the search function in Settings dialog (as in the _'Type to find'_ text box), once I click on any of the results or any other option in the tree, the search query and results are cleared.

**Expected behavior:** I can navigate the Settings dialog and the search results will remain highlighted. The search query will also remain in the text box. An "X" button on the right side of the text box (inside it) should allow me to quickly clear the search query and results.

Additionally, it will be good if the matching options themselves were highlighted in the right side of the Settings dialog, so that I can quickly find them with my eyes.

_PS: Thank you for this project! It's a powerful replacement for `gitk`. I tried searching for any existing issue with this request, but couldn't find one._

### Environment

- GitExtensions version: latest
- GIT version: latest
- OS version: Windows 10 x64
- .NET version: latest

pmiossec:
Fixes #9706

## Proposed changes

- Don't clear setting search when leaving the search box
- Add translation for "Type to find"

### Before

search_setting_before

### After

search_setting_after

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 6d09bcb15c0821a1b19ff5721c8c59f9e684738f
- .NET 6.0.6
- DPI 96dpi (no scaling)

## Merge strategy

----


gerhardol (approved):
Rebase merge?

pmiossec:
> Rebase merge?

Yes, I think it's maybe better (even if squashing could make sense also).
Except if you expect to take it in v4 and so in this case it could be easier if squashed.

GitUI/CommandsDialogs/SettingsDialog/SettingsTreeViewUserControl.cs
```diff
@@ -214,15 +213,18 @@ private void textBoxFind_Enter(object sender, EventArgs e)
...
             {
-                textBoxFind.Text = "";
+                textBoxFind.Text = string.Empty;
             }
         }
 
         private void textBoxFind_Leave(object sender, EventArgs e)
         {
-            SetFindPrompt(true);
+            if (textBoxFind.Text == string.Empty)
```
RussKie, line 224:
```suggestion
            if (string.IsNullOrWhiteSpace(textBoxFind.Text))
```

GitUI/CommandsDialogs/SettingsDialog/SettingsTreeViewUserControl.cs
```diff
@@ -124,7 +123,7 @@ private void textBoxFind_TextChanged(object sender, EventArgs e)
         {
             _nodesFoundByTextBox = new List<TreeNode>();
 
-            if (string.IsNullOrEmpty(textBoxFind.Text) || textBoxFind.Text == FindPrompt)
+            if (string.IsNullOrEmpty(textBoxFind.Text) || textBoxFind.Text == TranslatedStrings.SettingsTypeToFind)
```
RussKie, line 126:
```suggestion
            if (string.IsNullOrWhiteSpace(textBoxFind.Text) || textBoxFind.Text == TranslatedStrings.SettingsTypeToFind)
```

RussKie, line 126: I suppose we could use [`PlaceHolderText`](https://docs.microsoft.com/dotnet/api/system.windows.forms.textbox.placeholdertext) instead of setting the text to `TranslatedStrings.SettingsTypeToFind`.

RussKie (approved):
👍

GitUI/CommandsDialogs/SettingsDialog/SettingsTreeViewUserControl.cs
pmiossec: Updated. I made the change to `PlaceHolderText` (that simplify remove a whole bunch of code) but unfortunately doesn't prevent to use the `TranslatedStrings` because this field is not supported by the translation.

pmiossec:
I have also added a commit with a slight perf improvement.
Can you have a look...

RussKie (approved):

RussKie:
Thank you

mstv (approved):
LGTM

gerhardol:
@pmiossec I suggest you squash or rebase this as you prefer
Also push to release/4.0, to keep the source part identical while still patches are added to both branches

pmiossec:
> @pmiossec I suggest you squash or rebase this as you prefer Also push to release/4.0, to keep the source part identical while still patches are added to both branches

Done.
FYI, I would like to do a merge but did a 'rebase and merge' (because I forgot itbdid not create a merge commit and just do a fast-forward 😕) so I did the merge locally and pushed the fixed history.

PR #10136, opened 2022-08-06
BranchFilter text was added backwards

gerhardol:
Fixes #9872

## Proposed changes

Clearing the dropdown items changed the focus and moved the focus in the textbox to the first position, so next key was added first.

Dropdown was not opened on focus, which made the filter unusable unless you opened the dropdown first (se below).

Cache the function for getting the references with lazy result, to speed up the dropdown.

Still some weird behavior:
* If you focus away/back the first match in the dropdown is selected also if the dropdown is not shown (same in 3.5)
* When moving focus away then back to the control all is selected. In 3.5 allt ext was always selected, you had to start typing from scratch. With 4.0 you can click again, but first position in the box is always selected, you have to move with arrows.

This PR is not perfect but it is similar to the 3.5 behavior and I have no WinForms knowledge and I would like to keep it at this.

## Test methodology 

manual

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10168, opened 2022-08-28
Set branchfilter from sidepanel without checking dropdown

gerhardol:
## Proposed changes

The branchfilters from the sidepanel was not applied unless the
filter dropdown contained any items, but the items were not updated
by default since #10136, only when needed (dropdown clicked or text typed).

The dropdown was updated async FileAndForget, so 'external' usage was not
always successful. This was a reason why the test
'Filters_should_behave_as_expected' had to be disabled in AppVeyor
(the test still occasionally fails though).
(The test was the primary reason why I looked at this, the side panel was detected after.)

## Test methodology 

Manual
Test Filters_should_behave_as_expected works slightly better but still fails though.

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -39,19 +39,6 @@ namespace GitUI.UserControls
             tscboBranchFilter.ComboBox.ResizeDropDownWidth(AppSettings.BranchDropDownMinWidth, AppSettings.BranchDropDownMaxWidth);
         }
 
-        public RefsFilter BranchesFilter
-        {
-            get
-            {
-                // Options are interpreted as the refs the search should be limited too
-                // If neither option is selected all refs will be queried also including stash and notes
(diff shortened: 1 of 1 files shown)


mstv (approved):
LGTM, have not run

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -85,12 +72,8 @@ private void ApplyCustomBranchFilter()
...
             _filterBeingChanged = false;
 
-            string filter = tscboBranchFilter.Items.Count > 0 ? tscboBranchFilter.Text : string.Empty;
-            if (filter == TranslatedStrings.NoResultsFound)
-            {
-                filter = string.Empty;
-            }
-
+            // Apply the textbox contents, no check if the (multiple) options is in tscboBranchFilter.Items (or that the list is generated)
+            string filter = tscboBranchFilter.Text == TranslatedStrings.NoResultsFound ? string.Empty : tscboBranchFilter.Text;
```
gerhardol, line 76: This is the actual change in the PR
No point in checking the contents of the dropdown here

PR #10172, opened 2022-08-30
Decouple filter options

gerhardol:
Follow up to #10129
The filter handling is confusing, some comments in #10129 was due to that. It is also more limiting than needed.

## Proposed changes

*  Decouple simplify-by-decoration, no-merges, boundry as well as sort
 options from first-parent (they can be applied there too).

*  First-parents can be used together with other types of branch filters
(not relevant with reflog). Before it was dominating
reflog and others.

* Set prio order for the various branch-filters from having first-parent as
dominant to:
  Reflog->ShowCurrent->FilteredBranches (if non null filter)->All branches

* Separate ByBranchFilter from being shared CurrentFilter and
FilteredBranches to only be used for BranchFilter

### Before

image

### After

image

## Test methodology 

Tests are updated

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -546,20 +546,10 @@ namespace GitUI.CommandsDialogs
         public override void CancelButtonClick(object sender, EventArgs e)
         {
             // If a filter is applied, clear it
-            if (RevisionGrid.FilterIsApplied(false))
+            if (RevisionGrid.FilterIsApplied())
             {
                 // Clear filter
                 ToolStripFilters.SetRevisionFilter(string.Empty);
-            }
-
-            // If a branch filter is applied by text or using the menus "Show current branch only"
-            else if (RevisionGrid.FilterIsApplied(true) || AppSettings.BranchFilterEnabled)
-            {
-                // Clear branch filter
-                ToolStripFilters.SetBranchFilter(string.Empty);
-
-                // Execute the "Show all branches" menu option
-                RevisionGrid.ShowAllBranches();
             }
         }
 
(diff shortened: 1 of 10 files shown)


RussKie (approved):
LGTM

GitUI/UserControls/RevisionGrid/FormRevisionFilter.Designer.cs
```diff
@@ -470,6 +483,7 @@ private void InitializeComponent()
         private System.Windows.Forms.Label _NO_TRANSLATE_lblBranches;
         private System.Windows.Forms.TextBox BranchFilter;
         private System.Windows.Forms.CheckBox CurrentBranchOnlyCheck;
+        private System.Windows.Forms.Label _NO_TRANSLATE_lblCurrentBranchOnlyCheck;
```
mstv, line 486: How is this new Label used? At least the `Before` screenshot is not update.

gerhardol: Before, `CurrentBranchOnlyCheck` was in column 2 with the text label. I had not removed that label, pushed an update.
Since _Show current branch only_ is dominating `BranchFilter` the checkbox was shifted to column 1 with no space for the text label.
`_NO_TRANSLATE_lblCurrentBranchOnlyCheck` should have been added to column 0, but is too long why it was added to column 2 instead.

Before screenshot updated.

Reflog should be added to FormFilter as well as the branch filter as that option dominates the other filters. This PR cleans up the logic.
image

gerhardol:
(spuashed a correction to the disabled test Filters_should_behave_as_expected)
Plan to merge this tomorrow

mstv (approved):
👍

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1850,7 +1840,7 @@ private void SetGitModule(object sender, GitModuleEventArgs e)
                     // If we're applying custom branch or revision filters - reset them
                     RevisionGrid.ResetAllFilters();
                     ToolStripFilters.ClearQuickFilters();
-                    AppSettings.BranchFilterEnabled = AppSettings.BranchFilterEnabled && AppSettings.ShowCurrentBranchOnly;
+                    AppSettings.BranchFilterEnabled = AppSettings.BranchFilterEnabled;
```
mstv, line 1843: This does nothing any longer. Or have I missed something?
Why have I approved this?
I am removing it with the upcoming View Settings.

RussKie: I guess we both missed it :)
You're right, it's not longer doing anything.

PR #10183, opened 2022-09-06
Blame: fix blame revision older revision when the file has been renamed

pmiossec:
## Proposed changes

by providing the old path to blame and selecting it in the filetree
so that the blame view is not empty anymore
failing to find a not existing file in the filetree

### Before

blame_broken

### After

blame_fixed

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 7447469d3f686c3d28fba1fffbded9f39607688d (Dirty)
- .NET 6.0.8
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -50,6 +50,9 @@ namespace GitUI.Blame
         private static readonly IList<Color> AgeBucketGradientColors = GetAgeBucketGradientColors();
         private IGitRevisionSummaryBuilder _gitRevisionSummaryBuilder;
 
+        // Relative path of the file to blame when blaming a new revision
+        public string? PathToBlame { get; private set; }
+
         public BlameControl()
         {
             InitializeComponent();
@@ -538,7 +541,7 @@ namespace GitUI.Blame
 
             contextMenu.Tag = new GitBlameContext(_fileName, _lineIndex, GetBlameLine(), _blameId);
 
-            if (!TryGetSelectedRevision(out var selectedRevision))
+            if (!TryGetSelectedRevision(out var blameinfo))
             {
                 blameRevisionToolStripMenuItem.Enabled = false;
 
@@ -548,7 +551,7 @@ namespace GitUI.Blame
             }
 
             blameRevisionToolStripMenuItem.Enabled = true;
-            blamePreviousRevisionToolStripMenuItem.Enabled = RevisionHasParent(selectedRevision);
+            blamePreviousRevisionToolStripMenuItem.Enabled = RevisionHasParent(blameinfo.selectedRevision);
 
             // Get parent for the actual revision, the selected revision may have rewritten parents.
             // The menu will be slightly slower in this situation.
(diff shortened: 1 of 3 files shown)


gerhardol:
Looks reasonable. Cannot see if `PathToBlame` is set in all situations.

GitUI/UserControls/BlameControl.cs
```diff
@@ -599,46 +601,47 @@ private void copyCommitHashToClipboardToolStripMenuItem_Click(object sender, Eve
...
             {
                 return;
             }
 
             // Try get actual parent revision, get popup if it does not exist.
             // (The menu should be disabled if previous is not in grid).
-            GitRevision? revision = _revGrid?.GetActualRevision(selectedRevision);
-            BlameRevision(revision?.FirstParentId);
+            GitRevision? revision = _revGrid?.GetActualRevision(blameInfo.selectedRevision);
+            BlameRevision(revision?.FirstParentId, blameInfo.filename);
```
gerhardol, line 637: Can the filename be changed here already?
Not a blocker for the PR, but it warrants a comment.

gerhardol (commented):
What file and commits do you test with?

GitUI/UserControls/BlameControl.cs
gerhardol: I meant if the info.fileName was already changed and be incorrect here.

> I thought that changing the method provided a more understandable contract.

I thought of that too, but will approve just if there are added comments.

> and that let me see that there is a regression in the way we display file path in the blame gutter 😢 -- a problem of UNIX vs Windows path format

Do you want to display with Windows paths?
Presentation is generally in Unix paths so way different here?

pmiossec: > I meant if the info.fileName was already changed and be incorrect here.
> [...] but will approve just if there are added comments

I'm happy to put comments but I don't really get what you are asking me.

If the file has been renamed on the parent commit also (so 2 consecutive renames)?
Yes, we will end up with the same behavior than previously because we don't have the good path. But we can't do better with the information we got.

> Do you want to display with Windows paths? Presentation is generally in Unix paths so way different here?

No, I don't want to change the way the path are displayed. That's just that previously the path was displayed only if it is different than the one of the file you are blaming to have less information to handle as a user.
Like that:
d

But for that you have to compare paths ( https://github.com/gitextensions/gitextensions/blob/master/GitUI/UserControls/BlameControl.cs#L442 ). There is just a behavior change where now, the `filename` received is in Windows format whereas it was in Unix format before to have the comparison working.
I have already the one line fix. Just have to push it...

GitUI/UserControls/BlameControl.cs
```diff
@@ -50,6 +50,8 @@ public sealed partial class BlameControl : GitModuleControl
         private static readonly IList<Color> AgeBucketGradientColors = GetAgeBucketGradientColors();
         private IGitRevisionSummaryBuilder _gitRevisionSummaryBuilder;
 
+        public string? PathToBlame { get; private set; }
```
gerhardol, line 53:
```suggestion
        // Path for the current revision
        public string? PathToBlame { get; private set; }
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -599,46 +601,47 @@ private void copyCommitHashToClipboardToolStripMenuItem_Click(object sender, Eve
...
             // Try get actual parent revision, get popup if it does not exist.
             // (The menu should be disabled if previous is not in grid).
-            GitRevision? revision = _revGrid?.GetActualRevision(selectedRevision);
-            BlameRevision(revision?.FirstParentId);
+            GitRevision? revision = _revGrid?.GetActualRevision(blameInfo.selectedRevision);
+            BlameRevision(revision?.FirstParentId, blameInfo.filename);
         }
 
-        private void BlameRevision(ObjectId? revisionId)
+        private void BlameRevision(ObjectId? revisionId, string filename)
```
gerhardol, line 640: draft...
```suggestion
/// The current Git commit and filename
        private void BlameRevision(ObjectId? revisionId, string filename)
```

gerhardol (commented):
+1

pmiossec:
Comments added.

gerhardol (approved):

mstv (approved):
LGTM, have not run

gerhardol:
@pmiossec merge (I believe this is done), also to release/4.0

pmiossec:
> @pmiossec merge (I believe this is done), also to release/4.0

Thanks. done.

Issue #8006, opened 2020-04-18
"Cannot save settings" when closing multiple instances of GitExtensions


bansan85:
## Current behavior / Steps to reproduce

 * Open GitExtensions 10 times.
 * Close them all at the same time with "Close all windows" menu item from Windows taskbar.
 * Some instances will have "Cannot save settings" error message.

## Expected behavior

No error message.

## Screenshots

![Cannot_save_settings](https://user-images.githubusercontent.com/9248729/79642926-812a3980-81a0-11ea-8942-697597a481b1.png)

## Did this work in previous version of GitExtensions

I don't think so.

## Environment

- Git Extensions 3.3.1.7897
- Build 5a97671645532bcedc443bac7b727f40db47cb5c
- Git 2.25.1.windows.1
- Microsoft Windows NT 10.0.18363.0
- .NET Framework 4.8.4150.0
- DPI 96dpi (no scaling)

## Diagnostics

When Git Extensions closes, it saves his settings in a local file.
When there is multiple instances closed at the same time, one instance opens successfully the file and the others fail.

I suggest to had a small loop to retry after a small random time.
I already made a patch on my computer and this problem disappears. I can make a PR if you think it's a good solution.

It will not change the fact that only one instance of Git Extensions will have his settings saved, the settings of other instances will be override.

gerhardol:
> It will not change the fact that only one instance of Git Extensions will have his settings saved, the settings of other instances will be override.

Maybe the popup should ask if the user wants to wait and overwrite or if prefer that changes are lost?

bansan85:
@gerhardol 
The problem is that I have a project with lots of submodule (like boost.org). Every time I open one submodule, I open a new instance of Git Extensions. After an hour, I may have 20 instances of Git Extensions.
When I close them all together, I may have 15 MessageBox. This is really boring.

If I close Git Extensions instance one by one, only the settings of the last instance will be kept without any warning message. So I think asking a question will be more annoying than useful.

gerhardol:
The answer could be remembered though (like other confirmations in GE), then you can set your behavior once.

bansan85:
> The answer could be remembered though (like other confirmations in GE), then you can set your behavior once.

I don't really understand what you mean.

So you mean that GE should write only changed settings ? Instead of always writing all settings at every exit ?
Or do you mean the popup could have a button like "Ignore" then, with a single validation, all GE instances with this popup will close and not save settings ?

gerhardol:
[...]

PR #10193, opened 2022-09-15
Protect AppSettings saving using global mutex

### LINKED ISSUE #10016: Closing multiple instances at once breaks the configuration
### Environment

- GitExtensions version: master branch
- GIT version: 
- OS version: Windows 10 Professional
- .NET version: 

### Issue description

Closing multiple GitExtensions instances at once breaks the configuration

### Steps to reproduce

1. Open multiple GitExtensions instances.
2. Right click GitExtensions icon in the taskbar.
3. Click "Close all windows" menu item.

Unknown

### Diagnostics

_No response_

mstv:
Fixes #10016

## Proposed changes

- Protect `AppSettings.SaveSettings` using a global mutex common to all GE instances

## Screenshots 

N/A

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 403856d49cf56837f9a746c452dc3ab13b5ecb49
- .NET 6.0.9
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1606,7 +1606,9 @@ namespace GitCommands
             {
                 SettingsContainer.LockedAction(() =>
                 {
-                    _globalMutex ??= new Mutex(initiallyOwned: false, name: SettingsFilePath.ToPosixPath());
+                    // prepend "Global\" in order to be safe in preparation for non-Windows OS, too
+                    _globalMutex ??= new Mutex(initiallyOwned: false, name: @$"Global\Mutex{SettingsFilePath.ToPosixPath()}");
+
                     try
                     {
                         _globalMutex.WaitOne();
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Windows.Forms;
 using GitCommands.Settings;
 using GitExtUtils;
@@ -39,6 +40,8 @@ namespace GitCommands
         public static RepoDistSettings SettingsContainer { get; private set; }
 
         private static readonly SettingsPath DetailedSettingsPath = new AppSettingsPath("Detailed");
+
+        private static Mutex _globalMutex;
 
         public static readonly int BranchDropDownMinWidth = 300;
         public static readonly int BranchDropDownMaxWidth = 600;
@@ -1603,7 +1606,16 @@ namespace GitCommands
             {
                 SettingsContainer.LockedAction(() =>
                 {
-                    SettingsContainer.Save();
+                    _globalMutex ??= new Mutex(initiallyOwned: false, name: SettingsFilePath.ToPosixPath());
+                    try
+                    {
+                        _globalMutex.WaitOne();
+                        SettingsContainer.Save();
+                    }
+                    finally
+                    {
+                        _globalMutex.ReleaseMutex();
+                    }
                 });
 
                 Saved?.Invoke();


GitCommands/Settings/AppSettings.cs
```diff
@@ -1603,7 +1606,16 @@ public static void SaveSettings()
             {
                 SettingsContainer.LockedAction(() =>
                 {
-                    SettingsContainer.Save();
+                    _globalMutex ??= new Mutex(initiallyOwned: false, name: SettingsFilePath.ToPosixPath());
```
gerhardol, line 1609: should this be shared?
https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex.openexisting?view=net-6.0#system-threading-mutex-openexisting(system-string)

mstv: Yes, that's the trick. GE instances must write the settings one after the other. (The last one wins. For this I leave #8006 open.)

gerhardol: Why not use OpenExisting in this PR?
(This PR is probably an improvement anyway.)

RussKie (commented):
Reading the docs, this looks sensible 👍

GitCommands/Settings/AppSettings.cs
mstv: The use of `OpenExisting` needs a lot of additional code for little benefit (protection by access rights).
From https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-6.0#system-threading-mutex-ctor(system-boolean-system-string):
> If a name is provided and a synchronization object of the requested type already exists in the namespace, the existing synchronization object is used.

mstv: A named mutex is always shared between processes - at least in the same session.

> If a synchronization object of a different type already exists in the namespace, a WaitHandleCannotBeOpenedException is thrown.

gerhardol: https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-6.0#system-threading-mutex-ctor(system-boolean-system-string)
```
The name may be prefixed with Global\ or Local\ to specify a namespace. When the Global namespace is specified, the synchronization object may be shared with any processes on the system. When the Local namespace is specified, which is also the default when no namespace is specified, the synchronization object may be shared with processes in the same session. On Windows, a session is a login session, and services typically run in a different non-interactive session. On Unix-like operating systems, each shell has its own session. 
```

So it seems to be working in Windows, append `Global` to prepare for other platforms.

Have you tested multiple closing (by adding a delay or so).

gerhardol (approved):
I would like a confirmation it actually is tested though...

GitCommands/Settings/AppSettings.cs
mstv: OK, I admit I stopped reading after "On Windows, a session is a login session".
Adding `"Global\"`.

> Have you tested multiple closing (by adding a delay or so).

Yes, it takes 30s to close three instances at the same each waiting 10s - both with and without `"Global\"`

pmiossec:
> Fixes #10016

From what I see of the code, this PR is more a fix for  #8006 (error messages because multiple processes fails to acces the same file) than  #10016 (the last process which write the settings overwrite some changes made in the settings in a process that exited earlier).

And for #10016, there is no perfect (or easy) solution. Maybe preventing the instances that has no setting change to write the settings when exiting...

mstv:
> > Fixes #10016
> 
> From what I see of the code, this PR is more a fix for #8006 (error messages because multiple processes fails to acces the same file) than #10016 (the last process which write the settings overwrite some changes made in the settings in a process that exited earlier).
> 
> And for #10016, IMHO, there is no perfect (or easy) solution. Maybe preventing the instances that has no setting change to write the settings when exiting...

Igor proposed a solution in https://github.com/gitextensions/gitextensions/issues/8006#issuecomment-617755250.

I think #10016 has the same root cause: multiple instances are trying to write the settings file.

RussKie:
The settings is a pandora box, isn't it? :)

When I think about this area, I have the following problems in mind:
1. Whenever the Settings dialog is opened it should read the latest settings available, so that when the dialog is closed with [OK] button the settings are up-to-date. 
This operation is the "best we can do", because if there is another instance of the app modifies the settings whilst the Settings dialog is opened in the current instance, the changes from another instance will be overwritten. This is a pure user-error scenario, and I don't believe we have to try to address it.
2. Changes in the Settings dialog must be cancellable, i.e., if the dialog is dismissed (i.e., neither [OK] or [Apply] buttons clicked) then no changes should be written to the file.
I _believe_ this is the case now for all pages except for `ChecklistSettingsPage`:
https://github.com/gitextensions/gitextensions/blob/ad6671df6b114f49759e5dc17a69ffd4c54a8803/GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs#L153
4. Any other settings configurable outside of the Settings dialog (e.g., Show Current Branch) must be persisted immediately.
I _believe_ this is how the app is working now (though I haven't formally verified it).
https://github.com/gitextensions/gitextensions/blob/ad6671df6b114f49759e5dc17a69ffd4c54a8803/GitCommands/Settings/SettingsCache.cs#L119-L133
4. The config needs to be reloaded whenever the settings changed (as follows from \#3).
I _believe_ this is how the app is working now too, as least for some settings - if I apply "Show Current Branch" filter in one instance, and refresh another instance the filter is applied there too.

I don't think we must persist settings on the app exit, and this IMO is the ultimate fix for https://github.com/gitextensions/gitextensions/issues/10016 and other related issues.
https://github.com/gitextensions/gitextensions/blob/ad6671df6b114f49759e5dc17a69ffd4c54a8803/GitUI/CommandsDialogs/FormBrowse.cs#L1435-L1438

mstv:
Writing the settings must be an atomic operation. That's why this PR is necessary and should be taken as is.
Since this PR does not suffice to avoid losing settings, I have proposed to continue the discussion in #8006.

gerhardol:
This is not all settings, see FormSettings Save(), but what may be changed outside FormSettings it seems.
So open Settings could still try modify the settings.

This is an improvement and should be merged and included in 4.0

mstv:
> This is not all settings, see FormSettings Save(), but what may be changed outside FormSettings it seems.

Yes, there are other settings. Though they are neither written on `SessionEnding` nor on `Exit`. So they fall into point 1. of Igor's https://github.com/gitextensions/gitextensions/pull/10193#issuecomment-1250421009.
Yes, readers must be locked out while one instance is updating the file.

> So open Settings could still try modify the settings.

This should be idempotent, i.e. not change on reopen FormSettings right after it was closed.

mstv:
Any concerns yet merging this?

gerhardol:
> Any concerns yet merging this?

No, it should be limiting some issues.

PR #10199, opened 2022-09-18
Improve deletion of local branches

### LINKED ISSUE #10122: Some local branches could not be deleted anymore due to the state of the remote
### Environment

- Git Extensions 33.33.33
- Build f312b6a1206faef9e42cb87cbaf2cf1ea714c223
- Git 2.35.1.windows.2
- .NET 6.0.6
- DPI 96dpi (no scaling)

### Issue description

Following redesign of FormDeleteBranch in #8956, local branches that are merged into HEAD but that have there remote not merged can't be deleted anymore (because the "force delete" checkbox has been removed.

image

### Steps to reproduce

1. Having local branch merged into HEAD
2. Having remote of this branch ahead (because another dev has pushed new commits) or behind (because your merged it locally)
3. try to delete the local branch
4. Git refuse to delete it because it is not in sync with the remote (even if merged into HEAD). See screenshot above

Yes, in v<4.0

### Diagnostics

We can no more force delete (sometimes)

mstv:
Fixes #10122

## Proposed changes

- Always force-delete local branches regardless of the merged state (for unmerged branches after confirmation unless disabled)
- Do not load merged branches if `DontConfirmDeleteUnmergedBranch`

## Screenshots 

N/A

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 5a21c57542fc869811c9804869b11e8039236eae
- .NET 6.0.9
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormDeleteBranch.cs
+++ b/GitUI/CommandsDialogs/FormDeleteBranch.cs
@@ -48,15 +48,22 @@ namespace GitUI.CommandsDialogs
             base.OnRuntimeLoad(e);
 
             Branches.BranchesToSelect = Module.GetRefs(RefsFilter.Heads).ToList();
-            foreach (string branch in Module.GetMergedBranches())
+            if (AppSettings.DontConfirmDeleteUnmergedBranch)
             {
-                if (!branch.StartsWith("* "))
+                _currentBranch = Module.GetSelectedBranch();
+            }
+            else
+            {
+                foreach (string branch in Module.GetMergedBranches())
                 {
-                    _mergedBranches.Add(branch.Trim());
-                }
-                else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
-                {
-                    _currentBranch = branch.Trim('*', ' ');
+                    if (!branch.StartsWith("* "))
+                    {
+                        _mergedBranches.Add(branch.Trim());
+                    }
+                    else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
+                    {
+                        _currentBranch = branch.Trim('*', ' ');
+                    }
                 }
             }
 
@@ -104,7 +111,7 @@ namespace GitUI.CommandsDialogs
                 }
             }
 
-            GitDeleteBranchCmd cmd = new(selectedBranches, force: hasUnmergedBranches);
+            GitDeleteBranchCmd cmd = new(selectedBranches, force: true);
             bool success = UICommands.StartCommandLineProcessDialog(Owner, cmd);
             if (success)
             {
--- a/GitUI/CommandsDialogs/FormDeleteBranch.cs
+++ b/GitUI/CommandsDialogs/FormDeleteBranch.cs
@@ -56,13 +56,13 @@ namespace GitUI.CommandsDialogs
             {
                 foreach (string branch in Module.GetMergedBranches())
                 {
-                    if (!branch.StartsWith("* "))
-                    {
-                        _mergedBranches.Add(branch.Trim());
-                    }
-                    else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
+                    if (branch.StartsWith("* "))
                     {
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -104,7 +111,7 @@ private void Delete_Click(object sender, EventArgs e)
                 }
             }
 
-            GitDeleteBranchCmd cmd = new(selectedBranches, force: hasUnmergedBranches);
+            GitDeleteBranchCmd cmd = new(selectedBranches, force: true);
```
RussKie, line 114: This is the only place where we use `force` argument (outside the tests). Should we completely remove it?

RussKie (approved):
LGTM

It's sad that we don't have integration tests for this dialog.

GitUI/CommandsDialogs/FormDeleteBranch.cs
mstv: I would keep the argument `force` in order to emphasize it is added to the command.

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -48,15 +48,22 @@ protected override void OnRuntimeLoad(EventArgs e)
...
+            else
+            {
+                foreach (string branch in Module.GetMergedBranches())
                 {
-                    _currentBranch = branch.Trim('*', ' ');
+                    if (!branch.StartsWith("* "))
+                    {
+                        _mergedBranches.Add(branch.Trim());
+                    }
+                    else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
```
gerhardol, line 63: Not changed in this PR but still confusing
Also affects comment on line 92
```suggestion
                    else if (!DetachedHeadParser.IsDetachedHead(branch[2..]))
```

GitUI/CommandsDialogs/FormDeleteBranch.cs
```diff
@@ -48,15 +48,22 @@ protected override void OnRuntimeLoad(EventArgs e)
...
             Branches.BranchesToSelect = Module.GetRefs(RefsFilter.Heads).ToList();
-            foreach (string branch in Module.GetMergedBranches())
+            if (AppSettings.DontConfirmDeleteUnmergedBranch)
             {
-                if (!branch.StartsWith("* "))
-                {
-                    _mergedBranches.Add(branch.Trim());
-                }
-                else if (!branch.StartsWith("* ") || !DetachedHeadParser.IsDetachedHead(branch[2..]))
+                _currentBranch = Module.GetSelectedBranch();
```
gerhardol, line 53: Why is this needed?

gerhardol (commented):
+1

GitUI/CommandsDialogs/FormDeleteBranch.cs
mstv: > Not changed in this PR but still confusing

Indeed, this did not make any sense.

> Also affects comment on line 92

👍 moving this, makes the code clearer

GitUI/CommandsDialogs/FormDeleteBranch.cs
mstv: For the check to not delete the current branch.
In the `else` branch (with the original code), it is taken from the output of `GetMergedBranches()`.

gerhardol (approved):

GitUI/CommandsDialogs/FormDeleteBranch.cs
gerhardol, line 53: This comment probably what I missed at the first look, should be cleared too
```suggestion
                // No check for merged branches is needed
                _currentBranch = Module.GetSelectedBranch();
                _mergedBranches.Clear();
```

gerhardol (approved):

RussKie (approved):

mstv:
@RussKie, could you squash the three commits on the master branch, please? I am not allowed to force push.

RussKie:
I can, but I'd rather not do that on the master. We'll live with it now ;)

PR #10203, opened 2022-09-20
FormDeleteRemoteBranch: GetMergedRemoteBranches() in background

mstv:
## Proposed changes

- Factor out `TaskManager` from `ThreadHelper`
  in order to have a `JoinableTaskCollection` and `JoinableTaskFactory` per Form.
  So Forms and their tests can await their own async operations in the future.
- `FormDeleteRemoteBranch`:
  - Use `TaskManager` to run `GetMergedRemoteBranches()` in the background.
  - Await the execution when the data is needed. 

## Screenshots 

N/A

## Test methodology 

- manual
- existing tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build 5688d4848d9e8bcb5c7e35f40e69ce522b4350a5
- .NET 6.0.9
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitExtUtils/GitUI/ThreadHelper.cs
+++ b/GitExtUtils/GitUI/ThreadHelper.cs
@@ -24,7 +24,7 @@ namespace GitUI
 
         public static JoinableTaskFactory JoinableTaskFactory => _taskManager.JoinableTaskFactory;
 
-        public static TaskManager CreateSeparate()
+        public static TaskManager CreateTaskManager()
             => new(_taskManager.JoinableTaskContext);
 
         public static void ThrowIfNotOnUIThread([CallerMemberName] string callerMemberName = "")
--- a/GitExtUtils/GitUI/ThreadHelper.cs
+++ b/GitExtUtils/GitUI/ThreadHelper.cs
@@ -18,7 +18,7 @@ namespace GitUI
 
         public static JoinableTaskContext JoinableTaskContext
         {
-            get => _taskManager.JoinableTaskContext;
+            get => _taskManager?.JoinableTaskContext;
             internal set => _taskManager = value is null ? null : new(value);
         }
 
--- a/GitExtUtils/GitUI/ThreadHelper.cs
+++ b/GitExtUtils/GitUI/ThreadHelper.cs
@@ -6,7 +6,6 @@ using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Windows.Forms;
 using Microsoft.VisualStudio.Threading;
 
 namespace GitUI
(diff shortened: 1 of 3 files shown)


GitExtUtils/GitUI/ThreadHelper.cs
```diff
@@ -6,48 +6,26 @@
...
-                }
-            }
+            get => _taskManager?.JoinableTaskContext;
+            internal set => _taskManager = value is null ? null : new(value);
         }
 
-        public static JoinableTaskFactory JoinableTaskFactory => _joinableTaskFactory;
+        public static JoinableTaskFactory JoinableTaskFactory => _taskManager.JoinableTaskFactory;
+
+        public static TaskManager CreateSeparate()
```
RussKie, line 27:
```suggestion
        public static TaskManager CreateTaskManager()
```

GitExtUtils/GitUI/ThreadHelper.cs
```diff
@@ -6,48 +6,26 @@
...
-                }
-                else
-                {
-                    _joinableTaskContext = value;
-                    _joinableTaskCollection = value.CreateCollection();
-                    _joinableTaskFactory = value.CreateFactory(_joinableTaskCollection);
-                }
-            }
+            get => _taskManager?.JoinableTaskContext;
+            internal set => _taskManager = value is null ? null : new(value);
```
RussKie, line 22:
```suggestion
            internal set => _taskManager = value ?? new(value);
```

RussKie: Since we're in a static context, we need to synchronise to avoid race conditions.

RussKie (commented):
👍

GitExtUtils/GitUI/ThreadHelper.cs
mstv: OK

GitExtUtils/GitUI/ThreadHelper.cs
mstv: I think `value ?? new(value)` would do the opposite of what is needed. We need to keep `null` or else to create the `TaskManager` if we have a context. A `TaskManager` always has a context.

> Since we're in a static context, we need to synchronise to avoid race conditions.

This property must be set and reset in a controlled way - and that's the case: set once on program start, reset synchronously in some tests.
So I think this is one of the rare cases where we do not need bother.

gerhardol (commented):
+1

GitExtUtils/GitUI/ThreadHelper.cs
RussKie: Ah, yes. My mistake.

RussKie (approved):

PR #10212, opened 2022-09-30
Update filter toolbar when filters change

gerhardol:
https://github.com/gitextensions/gitextensions/pull/10205#issuecomment-1262772044

## Proposed changes

BranchFilter for named branches as well as the revision filter for Message/Committer/Author/DiffContent is now updated when the filters have changed.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -152,6 +152,9 @@ namespace GitUI.UserControls
             {
                 // Show filtered branches
                 selectedIndex = 2;
+
+                // Keep value if other filter
+                tscboBranchFilter.Text = e.BranchFilter;
             }
 
             if (e.ShowCurrentBranchOnly)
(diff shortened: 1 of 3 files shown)


GitUI/UserControls/RevisionGrid/FilterChangedEventArgs.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using GitUI.UserControls.RevisionGrid;
+using GitUI.UserControls.RevisionGrid.Graph;
```
mstv, line 3: not used

mstv (commented):
LGTM

Yet another point:

image image 👍

We should update on click of these items. Currently one must hit Enter in the filter textbox.

RussKie (commented):
LGTM

gerhardol:
> We should update on click of these items. Currently one must hit Enter in the filter textbox.

Changed this, it is probably better with some annoying refreshes than the extra enter required.
Skipped when the box is empty though. So if you clear the contents without Enter, chaning checkbox will not refresh (which is what I expect, if you change the contents, you should Enter).

RussKie:
> > We should update on click of these items. Currently one must hit Enter in the filter textbox.
> 
> Changed this, it is probably better with some annoying refreshes than the extra enter required.

Couldn't we send a keypress programmatically?

RussKie:
As you said, multiple checks can lead to multiple refreshes. Where a key
press could be done once after all checks are applied.

mstv (approved):
Thank you

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -329,13 +332,53 @@ private void revisionGridFilter_FilterChanged(object? sender, FilterChangedEvent
...
+                }
+                else
+                {
+                    item.menuItem.Checked = false;
+                }
+            }
+
+            if (!anyChecked)
+            {
+                revFilters[0].menuItem.Checked = true;
```
gerhardol, line 364: Will change this to retain the current setting if no filter is applied.

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -333,37 +333,34 @@ private void revisionGridFilter_FilterChanged(object? sender, FilterChangedEvent
             tsbShowReflog.Checked = e.ShowReflogReferences;
             InitBranchSelectionFilter(e);
 
-            tstxtRevisionFilter.Text = "";
```
mstv: This still is necessary.

mstv: Use case: Untick all checkboxes in the Filter dialog. Then the Filter textbox should be cleared. I.e. this must be done before the `if (any)`.

gerhardol: I missed to save and commit in the first update.
Done after the check.

I want to keep the info in the box, otherwise will text entered but not Enter given always be cleared.

mstv: This can be misleading because the pretty far away image instead of image is the only indication that the filter is not applied anymore.
(just discussing, not requesting a change)

mstv (commented):
+1

GitUI/UserControls/FilterToolBar.cs
gerhardol: It is unchanged from now, if you type in the box the info is only used after Enter

gerhardol:
Plan to merge tomorrow, include in 4.0

gerhardol:
>>>    Couldn't we send a keypress programmatically?

>> What would be the advantage compared to use the Check changed event?

> As you said, multiple checks can lead to multiple refreshes. Where a key press could be done once after all checks are applied.

@RussKie 
But the menu closes on check/uncheck, so keypress (or call to update filterInfo) must be done at every change anyway.

Issue #5117, opened 2018-06-24
Inconsistent "GitExt Browse" icon


mstv:
**Do you want to request a *feature* or report a *bug*?**

inconsistency

**What is the current behavior?**

The icon for "GitExt Browse" in the Explorer context menu is different from the one of the dialog it will open.
![grafik](https://user-images.githubusercontent.com/36601201/41824384-6adc9e08-780f-11e8-9f37-0a6a9d64e253.png)
![grafik](https://user-images.githubusercontent.com/36601201/41824389-80ec4982-780f-11e8-8584-a69409bb6469.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

right click a folder with a repository in the Explorer

**What is the expected behavior?**

I'd prefer the "browse" icon instead of the "folder" symbol in the Explorer context menu.

**Environment you encounter the issue:**

- GitExtensions version: master
- GIT version: 2.17.windows.2
- OS version: Windows 7
- .NET version: N/A

**Did this work in previous version of GitExtensions (which)?**

no

RussKie:
There is one more inconsistency:

![image](https://user-images.githubusercontent.com/4403806/42153298-0d670f8e-7e26-11e8-8162-1dbf051c2be4.png)

mstv:
Sorry, I'm a little bit confused now.
The current master revision seems to use the icon Bin\Logo\git-extensions-logo-final_**mixed**.ico.
I see on Windows 7 with "small icons" for the Windows Task Bar:
* attached to the Task Bar, not running: ![grafik](https://user-images.githubusercontent.com/36601201/42186735-05d781cc-7e4e-11e8-8f20-3d851aeecfc7.png); running: ![grafik](https://user-images.githubusercontent.com/36601201/42186773-261af05e-7e4e-11e8-8a2c-837d1c082334.png)
* commit dialog: ![grafik](https://user-images.githubusercontent.com/36601201/42186900-8a53d18a-7e4e-11e8-8377-586cba8ca0df.png)

(With version 2.49.03 which was the last version I installed using the installer, it was opposite:
[...]

PR #10217, opened 2022-10-03
ShellExtension: rename item in the setting form with same name it have in shell extension menu

pmiossec:
4 years later followup of #5117

Shell menu since #5117:

image

### Before

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build b78e768d55af39e5d79310d8f1b7e772cefcf30a (Dirty)
- .NET 6.0.8
- DPI 96dpi (no scaling)

## Merge strategy

----


mstv (approved):

pmiossec:
Looking at the Shell Extension code, I have a question: 

The shell extension menu is not translated because the strings are hard-coded in the cpp project (and so don't benefit of the translation strings).
So, does that makes sense to have the strings translated in the "Context menu preview" (here in french) whereas in fact the user will never have it in his language?
image

My point of view is that we should prevent the strings to be translated.

gerhardol:
> The shell extension menu is not translated because the strings are hard-coded in the cpp project (and so don't benefit of the translation strings). So, does that makes sense to have the strings translated in the "Context menu preview" (here in french) whereas in fact the user will never have it in his language?

Seem to be better to drop the translation if the shell extension is not translated
(__NO_TRANSLATION prefixes)

RussKie (approved):

RussKie:
> Looking at the Shell Extension code, I have a question:
> 
> The shell extension menu is not translated because the strings are hard-coded in the cpp project (and so don't benefit of the translation strings). So, does that makes sense to have the strings translated in the "Context menu preview" (here in french) whereas in fact the user will never have it in his language? image
> 
> My point of view is that we should prevent the strings to be translated.

Good point. For consistency we can drop the translations.

pmiossec:
> Good point. For consistency we can drop the translations.

Done.

gerhardol (approved):

mstv (approved):

RussKie (approved):

PR #10218, opened 2022-10-03
Worktree form management improvements

pmiossec:
## Proposed changes

- Don't close the worktree form when the creation failed to let the user fix the issue and retry
- Refresh the revision grid even when user decide not opening a new worktree because ha has created another branch or checked one out and we should display the change in a refreshed revision grid.

Note: commits better reviewed separately.

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build b78e768d55af39e5d79310d8f1b7e772cefcf30a (Dirty)
- .NET 6.0.8
- DPI 96dpi (no scaling)

## Merge strategy

- Rebase merge or Merge commit (not squash)

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -3050,6 +3050,10 @@ namespace GitUI.CommandsDialogs
         {
             using FormManageWorktree formManageWorktree = new(UICommands);
             formManageWorktree.ShowDialog(this);
+            if (formManageWorktree.ShouldRefreshRevisionGrid)
+            {
+                RefreshRevisions();
+            }
         }
 
         private void toolStripSplitStash_DropDownOpened(object sender, EventArgs e)
--- a/GitUI/CommandsDialogs/WorktreeDialog/FormCreateWorktree.cs
+++ b/GitUI/CommandsDialogs/WorktreeDialog/FormCreateWorktree.cs
@@ -111,8 +111,7 @@ namespace GitUI.CommandsDialogs.WorktreeDialog
                 }
             };
 
-            UICommands.StartGitCommandProcessDialog(this, args);
-            DialogResult = DialogResult.OK;
+            DialogResult = UICommands.StartGitCommandProcessDialog(this, args) ? DialogResult.OK : DialogResult.None;
         }
 
         private void ValidateWorktreeOptions()
--- a/GitUI/CommandsDialogs/WorktreeDialog/FormManageWorktree.cs
+++ b/GitUI/CommandsDialogs/WorktreeDialog/FormManageWorktree.cs
@@ -16,6 +16,8 @@ namespace GitUI.CommandsDialogs.WorktreeDialog
 
         private List<WorkTree>? _worktrees;
 
+        public bool ShouldRefreshRevisionGrid { get; private set; }
+
         [Obsolete("For VS designer and translation test only. Do not remove.")]
         private FormManageWorktree()
         {
@@ -257,6 +259,7 @@ namespace GitUI.CommandsDialogs.WorktreeDialog
             }
             else
             {
+                ShouldRefreshRevisionGrid = true;
                 Initialize();
             }
         }
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -124,14 +124,14 @@ namespace GitUI
             return success;
         }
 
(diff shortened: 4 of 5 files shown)


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3064,15 +3064,22 @@ private void createWorktreeToolStripMenuItem_Click(object sender, EventArgs e)
...
             {
                 return;
             }
 
-            GitModule newModule = new(formCreateWorktree.WorktreeDirectory);
-            if (newModule.IsValidGitWorkingDir())
+            if (formCreateWorktree.OpenWorktree)
             {
-                SetGitModule(this, new GitModuleEventArgs(newModule));
+                GitModule newModule = new(formCreateWorktree.WorktreeDirectory);
```
gerhardol, line 3074: If this fails, nothing is opened, should be fine to resolve

gerhardol (approved):

GitUI/CommandsDialogs/FormBrowse.cs
mstv: > If this fails, nothing is opened, should be fine to resolve

Do you mean the unchanged code should be wrapped in a `try {} catch{ RefreshRevisions(); throw; }`, too?

mstv (approved):
LGTM

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: No, that there is a risk for failure here, but we can live with that

RussKie:
I have no opinion, I'm yet to discover the usefulness of the worktrees.

pmiossec:
> I have no opinion, I'm yet to discover the usefulness of the worktrees.

The idea is simple : being able to work with different branches of a same repository in the same time. Especially when switching from one to another is costly (i.e. when checkout, restore of packages, build or changing specific configuration takes some time or is painful/boring to do)

vbjay:
Think work tree for our mono supported work and one for main dev in this
project. No need to have to clean between switching.

Think  1 work tree for pull request reviews and one for your main work.

Many uses.

mstv:
You can quickly cherry pick commits from other worktrees without the need to push and pull.
Stashes are shared, too.

pmiossec:
I like to have one worktree for (at least) the last released branch (which become more and more longer to check out as the development branch diverge). That way, I can easily use stack traces of production errors and if needed, try to reproduce an issue in debug with already the config file well configured (with a connection string towards a DB in the good version)

gerhardol:
Worktree is a favorite part of the application, I cannot use apps without this feature.
Some repos I worked with occasionally were growing a lot for some time (some users added binary files not added to LFS, some tool generated .xml files were adding tenths of MB in many updates. Then it was great to just have one clone for all five branches or so was working with for those repos.

The worktree form is also the ugliest form in GE.

pmiossec:
> The worktree form is also the ugliest form in GE.

I have done a redesign in my own branch. I can put a screenshot (tomorrow) and open a PR if you want....

pmiossec:
PR #10224 changed some logic (only one menu item "Manage worktree..." remaining) so I had to adapt this PR.
As a consequence, this PR must now be merged only after #10224 ...

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -3058,6 +3058,10 @@ private void manageWorktreeToolStripMenuItem_Click(object sender, EventArgs e)
         {
             using FormManageWorktree formManageWorktree = new(UICommands);
             formManageWorktree.ShowDialog(this);
+            if (formManageWorktree.ShouldRefreshRevisionGrid)
+            {
+                RefreshRevisions();
+            }
```
RussKie, line 3064: We don't seem to be using the dialog result from `ShowDialog` call. Is there a way to use that instead of introducing a new public property?

pmiossec: No, it seems not possible for me because the user will close with the close button (so with DialogResult.Cancel) but depending on what he did, the state of the flag will be different. So it means we will have to intercept the action and change the dialogResult value. At best it's a big hack and we will loose expressiveness, at worst, I don't know if it's really possible...

gerhardol:
I suggest this is included in 4.0 too. This is the only code change not in 4.0.

PR #10222, opened 2022-10-04
Ignore .gitmodules without path

### LINKED ISSUE #10211: Loading submodules crashes the app
### Environment

- GitExtensions version: 3.5.4.12724
- GIT version: 2.37.3.windows.1
- OS version: Windows 11 build 22000.978
- .NET version: 4.8.4515.0

### Issue description

I cloned a repository that contains submodules. After clone is performed, I can't see any submodule under the "submodules" header in the left side panel.
image

Clicking the yellow submarine button on the top bar shows a dropdown menu with the word "Loading..." only.
image

After a couple of minutes, the app closes, even with zero user interaction.

Sourcetree does show the submodules in the same local copy and does not crash nor close.
image

I uninstalled Git Extensions and git --> deleted any traces in %USERPROFILE%\AppData and Windows registry --> rebooted --> installed git --> rebooted --> installed Git Extensions --> rebooted --> and problem still persists.

### Steps to reproduce

Clone a reponsitory with submodules. Open the repository in Git Extensions. Just wait for Git Extensions to hang and then close.

_No response_

### Diagnostics

_No response_

gerhardol:
Fixes #10211

## Proposed changes

Path was not checked, causing recursive calls and stack overflow

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/UnitTests/GitCommands.Tests/GitModuleTests.cs
+++ b/UnitTests/GitCommands.Tests/GitModuleTests.cs
@@ -731,6 +731,25 @@ namespace GitCommandsTests
             }
         }
 
+        [Test]
+        public void Ignore_empty_gitmodules_path()
+        {
+            using GitModuleTestHelper helper = new();
+            try
+            {
+                helper.CreateFile(helper.Module.WorkingDir, ".gitmodules", @"[submodule ""Externals/NBug""]
+    path = Externals/NBug
+    url = https://github.com/gitextensions/NBug.git
+[submodule ""Externals/Git.hub""]
+    url = https://github.com/gitextensions/Git.hub.gitk");
+                helper.Module.GetSubmodulesLocalPaths().Should().Equal(new string[] { "Externals/NBug" });
+            }
+            finally
+            {
+                helper.Dispose();
+            }
+        }
+
         [Ignore("Flaky test - issue #7681")]
         [Test]
         public void GetSuperprojectCurrentCheckout()
--- a/UnitTests/GitCommands.Tests/GitModuleTests.cs
+++ b/UnitTests/GitCommands.Tests/GitModuleTests.cs
@@ -735,19 +735,12 @@ namespace GitCommandsTests
         public void Ignore_empty_gitmodules_path()
         {
             using GitModuleTestHelper helper = new();
-            try
-            {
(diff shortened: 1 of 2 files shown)


vbjay:
Better idea. See if dupe key and if so merge details with other config values. The original issue was that they had a module with config. Then seperate module entry with extra config. If keys matched instead of different keys, then we should merge settings together. 

https://github.com/gitextensions/gitextensions/issues/10211#issuecomment-1266759824

So if two sections had same key in above comment, we'd override settings in modules file last value of x property wins along with merge any new settings between sections. I understand keys didn't match in his comment but I see where this was going.

RussKie:
Is it possible to add a regression test?

gerhardol:
> Is it possible to add a regression test?

Do not know, this was a late night fix (that was still reasonable).
I doubt that is prepared or easy to do (or meaningful for regression).

> Better idea. See if dupe key and if so merge details with other config values. 

Empty paths must never be allow, it means that the current repo is added again which should never be done.
To check for duplicate keys is another check. The only data used is the path, so that would be discarding or throwing Git exception (causing a popup).
Separate change for anyone interested.

mstv (approved):

gerhardol:
> Is it possible to add a regression test?

Added a test

UnitTests/GitCommands.Tests/GitModuleTests.cs
```diff
@@ -731,6 +731,25 @@ public void GetSubmodulesLocalPaths()
...
+                helper.CreateFile(helper.Module.WorkingDir, ".gitmodules", @"[submodule ""Externals/NBug""]
+    path = Externals/NBug
+    url = https://github.com/gitextensions/NBug.git
+[submodule ""Externals/Git.hub""]
+    url = https://github.com/gitextensions/Git.hub.gitk");
+                helper.Module.GetSubmodulesLocalPaths().Should().Equal(new string[] { "Externals/NBug" });
+            }
+            finally
+            {
+                helper.Dispose();
```
RussKie, line 749: This would get disposed because of the `using` above.

RussKie (approved):

RussKie:
Thank you

PR #10227, opened 2022-10-05
Do not cache current GetRevision data

### LINKED ISSUE #10223: cBranch and cLocalBranch are empty, cHash has wrong value
### Environment

- Git Extensions 3.5.4.12724
- Git 2.37.3.windows.1
- .NET Framework 4.8.4515.0

### Issue description

I use the `cBranch` variable in my GitExt Scripts and it worked fine in version 3.4.3.
After upgrading to 3.5.4 the variable is always empty :(
I tryed `cLocalBranch` then, but it is also empty.
I then tryed `cHash` but it strangely returns a hash of some other commit.

### Steps to reproduce

Create a Script that echos cBranch, cLocalBranch or cHash

3.4.3.9999

### Diagnostics

_No response_

mstv:
Fixes #10223

## Proposed changes

- Do not cache the git output if querying the current checkout, i.e. let `GetRevision` use the cache only if an `ObjectId` is specified

## Screenshots 

N/A

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 22636a7d67522cb59de51fd9e38ec9145eeeeb30
- .NET 6.0.9
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -961,7 +961,7 @@ namespace GitCommands
                 /* Committer EMail*/ "%cE%n" +
                 /* Committer Date */ "%ct%n";
 
-            var format = formatString + (shortFormat ? "%s" : "%B%nNotes:%n%-N");
+            string format = formatString + (shortFormat ? "%s" : "%B%nNotes:%n%-N");
 
             GitArgumentBuilder args = new("log")
             {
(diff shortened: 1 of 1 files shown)


gerhardol (approved):

GitCommands/Git/GitModule.cs
```diff
@@ -970,7 +970,7 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
                 objectId
             };
 
-            var revInfo = _gitExecutable.GetOutput(args, cache: GitCommandCache, outputEncoding: LosslessEncoding);
+            var revInfo = _gitExecutable.GetOutput(args, cache: objectId is null ? null : GitCommandCache, outputEncoding: LosslessEncoding);
```
RussKie, line 973: A comment here would be helpful for any future references.

mstv: comment and variable added

GitCommands/Git/GitModule.cs
```diff
@@ -970,7 +970,10 @@ public GitRevision GetRevision(ObjectId? objectId = null, bool shortFormat = fal
                 objectId
             };
 
-            var revInfo = _gitExecutable.GetOutput(args, cache: GitCommandCache, outputEncoding: LosslessEncoding);
+            // cache output only if revision is specified
+            CommandCache? cache = objectId is null ? null : GitCommandCache;
```
gerhardol, line 974: Related: Cache would not be valid if called with an artificial commit, but the command will then fail anyway, so no check needed here...
Do not suggest a check in the method

gerhardol (approved):

PR #10228, opened 2022-10-05
ShowOnlyCurrentBranch were not showing commits

### LINKED ISSUE #10225: v4 beta "Show current branch only" shows blank page
### Environment

- GitExtensions version: 4.0.0.15291
- GIT version: 2.38.0
- OS version: Windows 10
- .NET version: .NET 6.0.9

### Issue description

From v4 beta, clone gitextensions repo itself, change to Show current branch only, then it shows a blank page.

### Steps to reproduce

View, Show current branch only

Yes.

### Diagnostics

_No response_

gerhardol:
Fixes #10225

## Proposed changes

Git Notes must be excluded for --all only with the previous argument "--not --glob=notes --not".
Changed to the easier to understand "--exclude=refs/notes" that could be used independent of the various branch filters (including reflog).
However, "--exclude=refs/notes" only makes sense with --all, so kept there.

## Test methodology 

Adopted test.
(It is really of the kind to increase test coverage, so quite meaningless.)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -308,44 +308,42 @@ namespace GitUI.UserControls.RevisionGrid
             // Branch filters
             if (ShowReflogReferences)
             {
+                // All commits
                 filter.Add("--reflog");
+            }
+            else if (IsShowCurrentBranchOnlyChecked)
+            {
+                // Default git-log (no option), only current branch (HEAD)
+            }
+            else if (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter))
+            {
+                // Show filtered branches only
+                filter.Add(IsSimpleBranchFilter(BranchFilter) ? BranchFilter : "--branches=" + BranchFilter);
             }
             else
             {
-                // Note that some refs (like notes) requires --all or explicit inclusion (--glob)
-                // (None is evaluated as HEAD)
-                // These options are explicitly excluded when not desired
-                // "other refs" include Gerrit refs like refs/for/ and refs/changes/
+                // refs (like notes) requires --reflog/--all or explicit inclusion (--glob)
+                // (so included for --reflog/--all, not explicitly added for other)
+                // --exclude is ignored for --reflog, only used with --all
+                // Similar but unhandled refs include Gerrit refs like refs/for/ and refs/changes/
                 if (!AppSettings.ShowGitNotes)
                 {
-                    filter.Add("--not --glob=notes --not");
+                    filter.Add($"--exclude={GitRefName.RefsNotesPrefix}");
                 }
 
                 if (!AppSettings.ShowStashes)
                 {
-                    filter.Add("--exclude=refs/stash");
+                    filter.Add($"--exclude={GitRefName.RefsStashPrefix}");
                 }
 
-                if (ShowCurrentBranchOnly)
-                {
-                    // Default git-log, only current branch, no option
-                }
-                else if (!string.IsNullOrWhiteSpace(BranchFilter))
-                {
-                    // Show filtered branches only
-                    filter.Add(IsSimpleBranchFilter(BranchFilter) ? BranchFilter : "--branches=" + BranchFilter);
-                }
-                else
-                {
-                    // All branches
-                    filter.Add("--all");
+                // All refs/
+                filter.Add("--all");
 
-                    // The inclusion of boundary parents to matches is historical
-                    // (why Message etc is handled as a special case)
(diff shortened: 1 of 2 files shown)


gerhardol:
@mstv should we show stash and notes for current only and branches?
Not done in 3.5

RussKie (commented):
I've looked at the changes but I can't currently verify the correctness of the changes.

I think we ought to come up with integration tests which would exercise various filters and assert mber of entries in the revision grid.

gerhardol:
> I've looked at the changes but I can't currently verify the correctness of the changes.

You will need to read https://git-scm.com/docs/git-log and understand the order and dependency of options.
The handling was quite complicated before when the implementation was split in three modules, should be easier now.

> I think we ought to come up with integration tests which would exercise various filters and assert mber of entries in the revision grid.

There are tests, some added in the recent refactoring. What was missing here was that the option was dependent  for this option (most options are unrelated). That test was added (will be added for stash too). The option was also changed to be easier to understand from the man page.

I have not used this myself yet, it is therefore this is a draft. (Was too late when creating.)

gerhardol:
CodeFactor reports an issue due to the added test case generator FilterInfo_NotesStash

mstv:
> Should we show stash and notes for current only and branches?

I think at least `stash@{0}` should always be displayed. Stashes are one's own work. (The `Show stashes` option could be added to the Advanced Filter - only if someone complains.)

I don't use notes (and I am not fond of auxilary commits). Though if one uses them, these commits could be of interest - at least if they are related to the visible branches.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -308,44 +308,42 @@ public ArgumentString GetRevisionFilter()
...
-                {
-                    // Show filtered branches only
-                    filter.Add(IsSimpleBranchFilter(BranchFilter) ? BranchFilter : "--branches=" + BranchFilter);
-                }
-                else
-                {
-                    // All branches
-                    filter.Add("--all");
+                // All refs/
+                filter.Add("--all");
```
mstv, line 340: ❔ Adding this argument in front of the "--exclude"s would be clearer to me. (I have compared this with the `git log` doc yet.)

gerhardol:
> I think at least `stash@{0}` should always be displayed. Stashes are one's own work. (The `Show stashes` option could be added to the Advanced Filter - only if someone complains.)

Added, filter is weird though...

> I don't use notes (and I am not fond of auxilary commits). Though if one uses them, these commits could be of interest - at least if they are related to the visible branches.

These commits could be excluded also for --all, not of interest. The commit they relate are not seen.
image

gerhardol:
Changed the filter to include stash@{0}, added a test.
I was tempted to make this ref independent of the branch filter, but it takes about 1 ms to resolve this ref for me, so it is dependent of current/filtered branches

--
I have 563 stashes in my GE test repo and to resolve all of them 100 times with --all, limiting output to 100 commits and that requires about 1:38 instead of 48 s, which indicates one stash is resolved in 1 ms. As the tests are repeated, a real scenario without caching could require more time. 
So always adding `stash@{0}` will add 1 ms to --reflog and -all, even if not needed. But the tests and the flow would be simpler.

mstv:
The test `GitExtensions.UITests.UserControls.RevisionGrid.RevisionGridControlTests.View_reflects_reset_branch_filter...` seems to fail consistently.

`stash@{0}` makes git think this was the HEAD of the `Current branch (only)`.

I think we must create a (set of) test repo(s) and verify that the expected commits are returned by git.

gerhardol:
> The test `GitExtensions.UITests.UserControls.RevisionGrid.RevisionGridControlTests.View_reflects_reset_branch_filter...` seems to fail consistently.
> 
> `stash@{0}` makes git think this was the HEAD of the `Current branch (only)`.
> 
> I think we must create a (set of) test repo(s) and verify that the expected commits are returned by git.

I pushed an update. If there were no stashes, the query failed. So back to a convoluted logic to get stash@{0} in the grid for current only...

mstv:
When refreshing (or switching between `All branches` and `Filtered branches` with empty filter), the result varies randomly:

image

image

image

image

image

gerhardol:
> When refreshing (or switching between `All branches` and `Filtered branches` with empty filter), 

That is the same...

> the result varies randomly:

The Git arguments is the same in this scenario, so it is a different issue
I cannot see why the revisions are processed in a different way.

gerhardol:
Squashed the changes for "stash@{0}" for current/filtered. That commit could be separated too (I prefer to keep it in this though).

mstv (approved):
Thank you for working this out

gerhardol:
Rebase merge this tomorrow

PR #10232, opened 2022-10-08
Update DiffViewerSettingsPage layout

RussKie:
### Before

image

### After

image

This aligns with other updated settings pages, such as General, Appearance, Advanced, etc.


mstv (commented):
+1

> This aligns with other updated settings pages, such as General, Appearance, Advanced, etc.

and there are others as
- Build Server, Git Config, which should be aligned, too (if we don't want to switch to PropertyGrid)
- Revision Links, Hotkeys, Scripts, which should not be put into frame

RussKie:
You're right - there are other panels to update. I just had 10 minutes spare, so I updated this one. I'll update others whenever I have another 10 minutes spare.


PS. Besides I want another hactoberfest t-shirt and I need few public PRs :)

gerhardol:
Are there other changes than adding the GroupBox?
Hard to review, but seem fine.
This changes translations. Is another push planned for 4.0?

RussKie:
No functional changes, just mechanical moves and tweaks to layout
properties. The translation changed due to the structure change, not due
new strings, so it should get auto translated.

V4 has fundamental regressions which have to be fixed, so there will be new
pushes, yes.

RussKie:
Planning to merge this tomorrow unless objected.

PR #9528, opened 2021-08-22
Enhance the filters toolbar

RussKie:
## Proposed changes

- Add "Show xyz branches", e.g. "Show all", "Show current" and "Show filtered"
- Add "Show reflogs"

Relates to https://github.com/gitextensions/gitextensions/pull/9446

### After
image
image

## Test methodology 

- manual
- unit tests

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10234, opened 2022-10-08
Restore the order of the branch view options

RussKie:
## Proposed changes

I'm struggling with the new order of the branches view options having built a muscle memory for the past 10+ years.

### v3.x
image

### v4 beta

image
image

### Fixed

image
image

## Merge strategy

----

### DIFF
--- a/GitUI/TranslatedStrings.cs
+++ b/GitUI/TranslatedStrings.cs
@@ -125,6 +125,7 @@ the last selected commit.");
         private readonly TranslationString _pathFilter = new("Path filter");
         private readonly TranslationString _showOnlyFirstParent = new("Show only first parent");
         private readonly TranslationString _showReflog = new("Show reflog");
+        private readonly TranslationString _showReflogTooltip = new("Show all reflog references");
         private readonly TranslationString _showCurrentBranchOnly = new("Show current branch only");
         private readonly TranslationString _simplifyByDecoration = new("Simplify by decoration");
 
@@ -260,6 +261,7 @@ the last selected commit.");
         public static string PathFilter = _instance.Value._pathFilter.Text;
         public static string ShowOnlyFirstParent = _instance.Value._showOnlyFirstParent.Text;
         public static string ShowReflog = _instance.Value._showReflog.Text;
+        public static string ShowReflogTooltip = _instance.Value._showReflogTooltip.Text;
         public static string ShowCurrentBranchOnly = _instance.Value._showCurrentBranchOnly.Text;
         public static string SimplifyByDecoration = _instance.Value._simplifyByDecoration.Text;
 
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -3212,7 +3212,7 @@ Do you want to continue?</source>
         <target />
       </trans-unit>
       <trans-unit id="tsbShowReflog.ToolTipText">
-        <source>Show reflog</source>
+        <source>Show all reflog references</source>
         <target />
       </trans-unit>
       <trans-unit id="tsbtnAdvancedFilter.ToolTipText">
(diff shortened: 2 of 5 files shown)


gerhardol:
This was changed (I believe it was modified due to feedback) to have the options in a logical order, most to least.
(Order of Filtered and current could be debated, but multiple branches may be filtered.)
To revert back just due to have it like it was before is not a good enough argument for me.
However, I will not vote against this change.

Good that the 4.0 doc was not updated again.

gerhardol (changes requested):
If this is changed, the default filter should be changed to 0

line 31
```
// Select an option until we get a filter bound.
            SelectShowBranchesFilterOption(selectedIndex: 1);
```

RussKie:
> To revert back just due to have it like it was before is not a good
enough argument for me.

We had this discussion before when the order of context menus changed in
the Commit dialog.

We shouldn't changed orders of menus that would have significant
disruptions to UX/muscle memory.

* Placing reflogs (which not long ago were claimed to be largely unknown to
most users) at the top is wrong, it's not the most used option.
* The current branch option preceded the filtered option for decades, this
is too very disruptive.

gerhardol:
Context commands are not the same as general menus. 
A general menu presents the functionality of the application in some logical way.

Will not block, but also not approve, unless traded for something else.

Related: The toolbar menu should display the shortcuts, as done in the View menu (that could be removed).
As the commands are related to the RevisionGrid, RevisionGrid needs a method to set the shortcuts with the reference to the FilterToolBar with the GetShortcutKeys() method.
(MenuCommands has a _revisionGrid reference and calls GetShortcutKeyDisplayStringFromRevisionGridIfAvailable(), probably not needed).

RussKie:
> A general menu presents the functionality of the application in some logical way.

Correct. The most logical way is how often a given feature is used by the majority of users. 
The vast majority of users would likely use "all branches" most of the time because it's the default option and because it's the git's default. 
Filtering - i.e., viewing the current branch or selected other branches - would likely be the 2nd and 3rd most used options. The order has been like this for at least a decade, so users likely have developed muscle memory. 
As discussed many times here, reflogs is likely the less known feature to many git users, and that was one of the reasons to add to the toolbar in https://github.com/gitextensions/gitextensions/pull/9528 - to make it more visible to users in hope that it'd spark some curiosity. But since it's not likely used by the majority of users, it's inappropriate to have it placed before the other items.



> ..unless traded for something else.

I don't like where this is going. I can start blocking PRs too.


> Related: The toolbar menu should display the shortcuts, as done in the View menu (that could be removed).

I have mixed feelings about this. Adding shortcuts will make the dropdown portion wider than it is now. We have the shortcuts in the menu already.
Also, I don't believe it's a common practice to add shortcuts to the toolstrip dropdowns - I've checked Windows Explorer, Visual Studio and Office for such example, and I couldn't find any.

RussKie:
> If this is changed, the default filter should be changed to 0
> 
> line 31
> 
> ```
> // Select an option until we get a filter bound.
>             SelectShowBranchesFilterOption(selectedIndex: 1);
> ```

Thank you, I missed this.

gerhardol (approved):
I do not like this, but can live with it.

pmiossec:
Taking the time to think a little bit of the change, I think that would be good to add a separator before the "Reflog" options to kind of a "special" option that they should not work with it enabled.

Because from my experience that's sometimes an option that beginners enable and not understanding well what it is doing, kept it enabled and bringing afterwards a lot of confusion. If I remember well, we already had multiple issues opened in the past on GE wrongly behaving and it was in fact the user enabling the feature.

And even perhaps with a more suggestive tooltip.

Something like:
image

PS: To add my point of view on the discussion, I found strange to have the reflog at the top because that's an option that you use not that often and temporarily even if I understand the idea of logical order...

PR #10235, opened 2022-10-08
Remove reflog toolstrip button

RussKie:
## Proposed changes

- Remove the "reflog" toolstrip buttons introduced in #9528 as the reflog option is now available in the toolstrip dropdown.

## Merge strategy

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -3215,10 +3215,6 @@ Do you want to continue?</source>
         <source>&amp;Translate</source>
         <target />
       </trans-unit>
-      <trans-unit id="tsbShowReflog.ToolTipText">
-        <source>Show reflog</source>
-        <target />
-      </trans-unit>
       <trans-unit id="tsbtnAdvancedFilter.ToolTipText">
         <source>Advanced filter</source>
         <target />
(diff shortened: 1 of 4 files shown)


gerhardol (commented):
I like this button, I switch all/reflog a lot and the button is easier.
The shortcut is still there (can it be displayed in the dropdown too?).
Can it be removed from defaults instead?

RussKie:
I remember you were so opposed to having this button 🤪

We placed in the drop down which makes it almost as easy to get to.
However, this causes the drop down change its size quite dramatically, and
if you try to quickly toggle reflogs using the button you won't be able to
that - instead you'll invoke something else... This pretty bad UX.

I tried fixing the ize of the drop down, but didn't look nice.

gerhardol:
> I remember you were so opposed to having this button 🤪

Never happened!
As I remember it though, I wanted to have this button before the branch filters, because they are related.

I have got used to trigger this a lot instead of banning users that squash when they should not.

> We placed in the drop down which makes it almost as easy to get to. However, this causes the drop down change its size quite dramatically, and if you try to quickly toggle reflogs using the button you won't be able to that - instead you'll invoke something else... This pretty bad UX. I tried fixing the ize of the drop down, but didn't look nice.

It is not this button that make a change, the repo/branch filter should be to the right. But that will break muscle memory and work less well for smaller windows.

But I will not block this right now, will try to usethe dropdown.
To approve I will need something to trade with...

RussKie:
> Never happened!
> As I remember it though, I wanted to have this button before the branch filters, because they are related.

My bad, apologies.

gerhardol (approved):
I prefer to keep the button, but will not block removal.

pmiossec:
> I prefer to keep the button, but will not block removal.

I don't really know if it's easily doable but maybe hiding it by default now that we have the possibility to hide buttons of the toolbar is good enough...

image

RussKie:
Closing in favour of #10294

Issue #10213, opened 2022-10-02
Filter grid with Git options


gerhardol:
### Feature description

The arguments to filter the revision grid does not have the full possibilities as is available in Git.
* Hide specific branches
* Show all stashes without reflog
* Certain regexp content searches
Especially https://git-scm.com/docs/git-log#_history_simplification has a lot of possibilities.

A GUI for this would be very complex, still not covering all possibilities.
One way is to allow the user to enter custom options in the revision fulter toolbar box (probably with a dropdown of the last used arguments).
A problem is that if the user is allowed to enter data here, it could cause the Git command to fail, blaming the program.
A possible implementation is to expect that the user add custom arg if the search string starts with a certain string, like "--not" which should be a common filter anyway (--not --not will allow normal filters).

### Environment

- Git Extensions 33.33.33
- Build a07f37895dd708033b88c5d9de2c6adcaff17587 (Dirty)
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.9
- DPI 96dpi (no scaling)

RussKie:
> * Hide specific branches

I think this what a lot of users would want, and I think there have been asks for this feature.
I also think it probably wouldn't be too difficult to implement using [ToolStripControlHost](https://learn.microsoft.com/dotnet/api/system.windows.forms.toolstripcontrolhost) using a custom filter control that would be based on a [CheckedListBox](https://learn.microsoft.com/dotnet/api/system.windows.forms.checkedlistbox) and a few buttons to select/deselect all. 
A little complication could be defining an algorithm to determine which branches to include/exclude depending on how many branches would be selected/deselected, as it would have the direct impact on the git command.

gerhardol:
> > * Hide specific branches
> 
> I think this what a lot of users would want, and I think there have been asks for this feature. I also think it probably wouldn't be too difficult to implement using 

I want to filter out branches using wildcards, no point in excluding something if you have to do it every time there is a change

PR #10239, opened 2022-10-08
Revision filter combobox

gerhardol:
Part of #10213, can be applied independently

## Proposed changes

Remember last 30 used items in the revision filter

Keep the last used revision filter text (but not the type) in settings, so the filters can easily be retained.
I doubt that the revision filter is used much right now, but it still simplifies.

The "memory" is stored in settings when it is changed. There are other discussions about this behavior...

This is possible to include in 4.0 but maybe the full #10213 function should be added

### Before

image

### After

(--exclude is from another usage, the core #10213 functionality)
image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1265,6 +1265,12 @@ namespace GitCommands
             set => SetBool("showfirstparent", value);
         }
 
+        public static string[] RevisionFilterDropdowns
+        {
+            get => GetString("RevisionFilterDropdowns", string.Empty).Split('\n', StringSplitOptions.RemoveEmptyEntries);
+            set => SetString("RevisionFilterDropdowns", string.Join("\n", value ?? Array.Empty<string>()));
+        }
+
         public static bool CommitDialogSelectionFilter
         {
             get => GetBool("commitdialogselectionfilter", false);
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -25,15 +25,9 @@ namespace GitUI.UserControls
             // Select an option until we get a filter bound.
             SelectShowBranchesFilterOption(selectedIndex: 1);
 
-            tstxtRevisionFilter.KeyUp += (s, e) =>
-            {
-                if (e.KeyValue == (char)Keys.Enter)
-                {
-                    ApplyRevisionFilter();
-                }
-            };
-
             tscboBranchFilter.ComboBox.ResizeDropDownWidth(AppSettings.BranchDropDownMinWidth, AppSettings.BranchDropDownMaxWidth);
+            tstxtRevisionFilter.ComboBox.ResizeDropDownWidth(AppSettings.BranchDropDownMinWidth, AppSettings.BranchDropDownMaxWidth);
+            tstxtRevisionFilter.Items.AddRange(AppSettings.RevisionFilterDropdowns);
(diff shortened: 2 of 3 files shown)


mstv (approved):
LGTM, have not run

GitCommands/Settings/AppSettings.cs
```diff
@@ -1272,6 +1272,18 @@ public static bool ShowOnlyFirstParent
...
+        public static string[] RevisionFilterDropdowns
+        {
+            get
+            {
+                return GetString("RevisionFilterDropdowns", string.Empty).Split('\n', StringSplitOptions.RemoveEmptyEntries);
+            }
+            set
+            {
+                SetString("RevisionFilterDropdowns", string.Join("\n", value ?? Array.Empty<string>()));
+            }
```
RussKie, line 1284:
```suggestion
            get => GetString("RevisionFilterDropdowns", string.Empty).Split('\n', StringSplitOptions.RemoveEmptyEntries);
            set => SetString("RevisionFilterDropdowns", string.Join("\n", value ?? Array.Empty<string>()));
```

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -362,6 +356,21 @@ private void revisionGridFilter_FilterChanged(object? sender, FilterChangedEvent
...
+            {
+                if (tstxtRevisionFilter.Items.Contains(filter))
+                {
+                    tstxtRevisionFilter.Items.Remove(filter);
+                }
+
+                tstxtRevisionFilter.Items.Insert(0, filter);
+                tstxtRevisionFilter.Text = filter;
+                AppSettings.RevisionFilterDropdowns = tstxtRevisionFilter.Items.Cast<object>()
+                    .Select(item => item.ToString()).Take(30).ToArray();
```
RussKie, line 371: Please make this at least a local const.

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -362,6 +356,21 @@ private void revisionGridFilter_FilterChanged(object? sender, FilterChangedEvent
...
+            // Add to dropdown and settings, unless already included
+            string filter = tstxtRevisionFilter.Text;
+            if (!string.IsNullOrWhiteSpace(filter) && (tstxtRevisionFilter.Items.Count == 0 || filter != (string)tstxtRevisionFilter.Items[0]))
+            {
+                if (tstxtRevisionFilter.Items.Contains(filter))
+                {
+                    tstxtRevisionFilter.Items.Remove(filter);
+                }
+
+                tstxtRevisionFilter.Items.Insert(0, filter);
```
RussKie, line 368: Should we at least trim the text before comparing/storing?

gerhardol: ok

GitUI/UserControls/FilterToolBar.cs
gerhardol: ok

GitCommands/Settings/AppSettings.cs
gerhardol: ok

GitUI/UserControls/FilterToolBar.cs
gerhardol: The combobox text is trimmed, so all uses are trimmed

RussKie (approved):

PR #10240, opened 2022-10-08
Custom options to git-log

### LINKED ISSUE #10213: Filter grid with Git options

gerhardol:
Fixes #10213

## Proposed changes

If "Message" starts with "--not " or "--exclude=", interpret the string as options to git-log.
This enables the possibility to enter filters that the GUI does not support (like negating conditions).

The options has no checks for Git syntax or if ParentsAreRewritten etc If need to be adjusted, it is up to the user.

For now, this feature will be described in GitDoc only.
See discussion in #10213
Feature simpler to use with #10239 

## Screenshots 

Example hiding filters with wildcard

### Before

image

### After

image

## Test methodology 

Test added

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -18,6 +18,12 @@ namespace GitUI.UserControls.RevisionGrid
         private int _commitsLimit = -1;
 
         /// <summary>
+        /// Prefix to "Message" filters that forces the text to be interpreted as Git options.
+        /// This enables use of options not available in the GUI.
+        /// </summary>
+        private readonly string[] _messageAsGitOptions = new[] { "--not ", "--exclude=" };
+
+        /// <summary>
         ///  Gets whether all properties will unconditionally return the underlying data.
         ///  Otherwise return values will depend on the respective filter, e.g. "get => ByXyz ? Xyz : default".
         /// </summary>
@@ -164,25 +170,25 @@ namespace GitUI.UserControls.RevisionGrid
 
         /// <summary>
         /// Has any filters in addition to revision filters that sets the history.
-        /// Currently, this is only branch filters.
+        /// Currently, this is only branch and date filters.
         /// Note that All/Reflog are not considered as filters.
         /// </summary>
         public bool HasFilter
         {
             get => HasRevisionFilter
+                || ByDateFrom
+                || ByDateTo
                 || ShowCurrentBranchOnly
                 || ShowOnlyFirstParent
                 || !string.IsNullOrWhiteSpace(BranchFilter);
         }
 
         /// <summary>
-        /// Has revision filters that hides parents, not just branches.
+        /// Has revision filters that potentially hides parents, not just branches.
         /// </summary>
         public bool HasRevisionFilter
         {
-            get => ByDateFrom
-                || ByDateTo
-                || ByAuthor
+            get => ByAuthor
                 || ByCommitter
                 || ByMessage
                 || ByDiffContent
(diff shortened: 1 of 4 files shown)


mstv (commented):
Code LGTM, have not run
Test `GitUITests.UserControls.FilterInfoTests.FilterInfo_GetRevisionFilter("author1","committer2","mes...` needs to be adapted: To the changed order?

gerhardol:
> Test GitUITests.UserControls.FilterInfoTests.FilterInfo_GetRevisionFilter("author1","committer2","mes... needs to be adapted: To the changed order?

Yes, that is a "parallel implementation test". 
Squashed the change.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -303,6 +309,77 @@ public ArgumentString GetRevisionFilter()
...
+            }
+
+            if (ByDiffContent && !string.IsNullOrWhiteSpace(DiffContent))
+            {
+                filter.Add($"-G\"{DiffContent}\"");
+            }
+
+            if (ByMessage && !string.IsNullOrWhiteSpace(Message))
+            {
+                if (Message.StartsWithAny(_messageAsGitOptions))
```
RussKie, line 360: `Message` must be trimmed to account for any whitespaces

RussKie (commented):
I'm not going to oppose this change - it very useful, but `GetRevisionFilter()` is starting crying for a refactor.... Cyclomatic complexity is off the scaled :(

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: this time I prefer to not change, this is a special handling....

RussKie:
> `GetRevisionFilter()` is actually a lot simpler now then when it was split in several dependent methods and the options were coupled. This is just changing the order, as well as one if().
> Do you have any proposal to how to refactor?

I'm noting a maintenance risk here, that's all.
I agree, it is certainly much easier to comprehend since we overhauled it and moved most (?) of the related code into FilterInfo. Still, it's a fairly large chain of `if..else` conditions. 
I don't have any immediate suggestions other than possibly break the method into a series of small targeted methods, maybe even with fluent notations. But this would fall completely outside the scope of this change.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
RussKie: I think it's well worth adding several test cases with various content of `Message` to document the intended/expected behaviours.

RussKie:
I find the change useful, but before I sign off, I'd like the noted gaps to be closed.

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
```diff
@@ -1087,6 +1088,33 @@ public void FilterInfo_GitNotes_Stashes(bool showGitNotes, bool showStash, bool
             }
         }
 
+        [TestCase("message1", true)]
+        [TestCase("message1 --not", true)]
+        [TestCase("--not message1", false)]
+        public void FilterInfo_Message(string message, bool expectGrep)
```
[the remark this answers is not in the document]
gerhardol: Will add --exclude and a few whitespace variations (where whitespace is not trimmed, unless a change is requested).
But this is a decoupled setting, not depending on other. Do you want to cross check all?

gerhardol:
> > Do you have any proposal to how to refactor?
> 
> I'm noting a maintenance risk here, that's all. I agree, it is certainly much easier to comprehend since we overhauled it and moved most (?) of the related code into FilterInfo. Still, it's a fairly large chain of `if..else` conditions. I don't have any immediate suggestions other than possibly break the method into a series of small targeted methods, maybe even with fluent notations. But this would fall completely outside the scope of this change.

_revision_ and _branch revision_ filters could maybe be separated, except that this adds some kindof dependency.
I can separate the branch revision handling to separate methods if you prefer and test them separately, without splitting filterInfo in this PR. (See the HasRevision filter changes recently as an example).

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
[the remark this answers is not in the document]
RussKie: This is the test for the new changes, the rest of the settings have just been moved. Or am I missing something?

> (where whitespace is not trimmed, unless a change is requested).

Any specific reason you're opposing the trimming?

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: > " --exclude bla" won't be treated as the filter though

Yes, my intention. Special handling to have special rules.
I can change this though.

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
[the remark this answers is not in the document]
gerhardol: > Yes, my intention. Special handling to have special rules.
I can change this though.

You have asked for this three times now, I take this as a request and will trim.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
RussKie: My concern is that users may find it unintuitive. E.g., a user may copy-paste the "--not" or "--exclude" filter from somewhere, and that string may very well start with a whitespace (space, tab, non-breakable space, etc.).

UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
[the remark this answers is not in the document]
RussKie: I'm just trying to understand the reasoning as I feel like I'm missing the intent.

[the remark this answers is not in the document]
gerhardol: As it is now, this should be documented in the doc, so the discover-ability is low.
It should probably be done in abetter way eventually, even if custom options + combobox is functional, when you know about it. All edges do not have to be polished right now.

gerhardol:
Split up GetRevisionFilter() and GetSummary() in three separate private functions each. Still testing the public functions.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -298,12 +304,124 @@ public ArgumentString GetRevisionFilter()
...
+            if (ByMessage && !string.IsNullOrWhiteSpace(Message))
+            {
+                if (Message.Trim().StartsWithAny(_messageAsGitOptions))
+                {
+                    filter.Add(Message);
+                }
+                else
+                {
+                    filter.Add($"--grep=\"{Message}\"");
+                }
```
RussKie, line 396:
```suggestion
                string message = Message.TrimStart();
                if (message.StartsWithAny(_messageAsGitOptions))
                {
                    filter.Add(message);
                }
                else
                {
                    // The user may want to grep with leading or trailing whitespaces, retain the original message
                    filter.Add($"--grep=\"{Message}\"");
                }
```

RussKie: Should probably apply the same to the similar code below.

RussKie (approved):

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: Trim() applied where info is read from the filter boxes.

PR #10252, opened 2022-10-13
Fix resetting selected lines for renamed files in index

mdonatas:
## Fixes
Fixes an issue where if you reset selected lines in a staged and renamed file then rename is undone too.

## Proposed changes

- Adjust the patch header to remove "reverse rename" part.

### Before

reset-lines-on-renamed-before

### After

reset-lines-on-renamed-after

## Test methodology 

- Manual
- Integration test

## Test environment(s) 

- Git Extensions 33.33.33
- Build 023b2d8438aa0758af2018d0fc9e998418989510 (Dirty)
- Git 2.37.3.windows.1
- .NET 6.0.10
- DPI 192dpi (200% scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Patches/PatchManager.cs
+++ b/GitCommands/Patches/PatchManager.cs
@@ -90,18 +90,37 @@ namespace GitCommands.Patches
 
         private static string CorrectHeaderForRenamedFile(string header)
         {
+            // Expected input:
+            //
+            // diff --git a/original.txt b/original2.txt
+            // similarity index 88%
+            // rename from original.txt
+            // rename to original2.txt
+            // index 0e05069..d4029ea 100644
+            // --- a/original.txt
+            // +++ b/original2.txt
+
+            // Expected output:
+            //
+            // diff --git a/original2.txt b/original2.txt
+            // index 0e05069..d4029ea 100644
+            // --- a/original2.txt
+            // +++ b/original2.txt
+
             string[] headerLines = header.Split(new[] { '\n' });
-            string? oldNameFull = null;
+            string? oldNameWithPrefix = null;
             string? newName = null;
             foreach (string line in headerLines)
             {
                 if (line.StartsWith("+++"))
                 {
+                    // Takes the "original2.txt" part from patch file line: +++ b/original2.txt
                     newName = line[6..];
                 }
                 else if (line.StartsWith("---"))
                 {
-                    oldNameFull = line[4..];
+                    // Takes the "a/original.txt" part from patch file line: --- a/original.txt
+                    oldNameWithPrefix = line[4..];
                 }
             }
 
@@ -114,12 +133,13 @@ namespace GitCommands.Patches
                 {
                     line = $"--- a/{newName}";
                 }
-                else if (line.Contains($" {oldNameFull} "))
+                else if (line.Contains($" {oldNameWithPrefix} "))
                 {
-                    line = line.Replace($" {oldNameFull} ", $" a/{newName} ");
+                    line = line.Replace($" {oldNameWithPrefix} ", $" a/{newName} ");
                 }
                 else if (line.StartsWith("rename from ") || line.StartsWith("rename to ") || line.StartsWith("similarity index "))
                 {
+                    // Note: this logic depends on git not localizing patch file format
                     continue;
                 }
 
--- a/GitCommands/Patches/PatchManager.cs
+++ b/GitCommands/Patches/PatchManager.cs
@@ -28,9 +28,9 @@ namespace GitCommands.Patches
             }
         }
 
-        public static byte[]? GetSelectedLinesAsPatch(string text, int selectionPosition, int selectionLength, bool isIndex, Encoding fileContentEncoding, bool isNewFile)
+        public static byte[]? GetSelectedLinesAsPatch(string text, int selectionPosition, int selectionLength, bool isIndex, Encoding fileContentEncoding, bool reset, bool isNewFile, bool isRenamed)
         {
-            var selectedChunks = GetSelectedChunks(text, selectionPosition, selectionLength, out var header);
+            IReadOnlyList<Chunk>? selectedChunks = GetSelectedChunks(text, selectionPosition, selectionLength, out string header);
 
             if (selectedChunks is null || header is null)
             {
@@ -43,9 +43,15 @@ namespace GitCommands.Patches
                 header = CorrectHeaderForNewFile(header);
             }
 
+            // if file is renamed and selected lines are being reset from index then the patch undoes the rename too
+            if (isIndex && isRenamed && reset)
+            {
+                header = CorrectHeaderForRenamedFile(header);
+            }
+
             string? body = ToIndexPatch(selectedChunks, isIndex, isWholeFile: false);
 
-            if (header is null || body is null)
+            if (body is null)
             {
                 return null;
             }
@@ -77,6 +83,52 @@ namespace GitCommands.Patches
                 {
                     sb.Append(line).Append('\n');
                 }
+            }
+
+            return sb.ToString();
+        }
+
+        private static string CorrectHeaderForRenamedFile(string header)
+        {
+            string[] headerLines = header.Split(new[] { '\n' });
+            string? oldNameFull = null;
+            string? newName = null;
+            foreach (string line in headerLines)
+            {
+                if (line.StartsWith("+++"))
+                {
+                    newName = line[6..];
+                }
+                else if (line.StartsWith("---"))
+                {
+                    oldNameFull = line[4..];
+                }
+            }
+
+            StringBuilder sb = new();
+
+            for (int i = 0; i < headerLines.Length; i++)
+            {
+                string line = headerLines[i];
+                if (line.StartsWith("--- "))
+                {
+                    line = $"--- a/{newName}";
+                }
+                else if (line.Contains($" {oldNameFull} "))
+                {
+                    line = line.Replace($" {oldNameFull} ", $" a/{newName} ");
+                }
+                else if (line.StartsWith("rename from ") || line.StartsWith("rename to ") || line.StartsWith("similarity index "))
+                {
+                    continue;
+                }
+
+                if (i != 0)
+                {
+                    sb.Append('\n');
+                }
+
+                sb.Append(line);
             }
 
             return sb.ToString();
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -434,7 +434,7 @@ namespace GitUI.CommandsDialogs
         {
             if (!_bypassActivatedEventHandler && AppSettings.RefreshArtificialCommitOnApplicationActivated)
             {
-                RescanChanges();
+                RescanChangesAsync();
             }
 
             base.OnApplicationActivated();
@@ -494,7 +494,7 @@ namespace GitUI.CommandsDialogs
         {
             if (!_initialized)
             {
-                Initialize();
+                InitializeAsync();
             }
 
             UpdateAuthorInfo();
@@ -858,7 +858,7 @@ namespace GitUI.CommandsDialogs
                 case Command.EditFile: editFileToolStripMenuItem.PerformClick(); return true;
                 case Command.AddSelectionToCommitMessage: return AddSelectionToCommitMessage();
                 case Command.CreateBranch: createBranchToolStripButton.PerformClick(); return true;
-                case Command.Refresh: RescanChanges(); return true;
+                case Command.Refresh: RescanChangesAsync(); return true;
                 case Command.SelectNext:
                 case Command.SelectNext_AlternativeHotkey1:
                 case Command.SelectNext_AlternativeHotkey2: MoveSelection(1);  return true;
@@ -871,7 +871,7 @@ namespace GitUI.CommandsDialogs
 
         #endregion
 
-        private void ComputeUnstagedFiles(Action<IReadOnlyList<GitItemStatus>> onComputed, bool doAsync)
+        private Task ComputeUnstagedFilesAsync(Action<IReadOnlyList<GitItemStatus>> onComputed, bool doAsync)
         {
             IReadOnlyList<GitItemStatus> GetAllChangedFilesWithSubmodulesStatus()
             {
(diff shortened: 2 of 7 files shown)


mdonatas:
Related discussion that I've opened for integration tests part https://github.com/gitextensions/gitextensions/discussions/10251

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -434,7 +434,7 @@ protected override void OnApplicationActivated()
         {
             if (!_bypassActivatedEventHandler && AppSettings.RefreshArtificialCommitOnApplicationActivated)
             {
-                RescanChanges();
+                RescanChangesAsync();
```
mstv, line 437: The returned `Task` must be run, mustn't it?

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -494,7 +494,7 @@ protected override void OnShown(EventArgs e)
         {
             if (!_initialized)
             {
-                Initialize();
+                InitializeAsync();
```
mstv, line 497: ditto

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -884,26 +884,27 @@ IReadOnlyList<GitItemStatus> GetAllChangedFilesWithSubmodulesStatus()
 
             if (doAsync)
             {
-                ThreadHelper.JoinableTaskFactory.RunAsync(() =>
+                return ThreadHelper.JoinableTaskFactory.RunAsync(() =>
                 {
                     return _unstagedLoader.LoadAsync(GetAllChangedFilesWithSubmodulesStatus, onComputed);
-                });
+                }).Task;
```
mstv, line 890: ❔ just
```
await _unstagedLoader.LoadAsync(GetAllChangedFilesWithSubmodulesStatus, onComputed);
```

GitCommands/Patches/PatchManager.cs
```diff
@@ -82,6 +88,52 @@ private static string CorrectHeaderForNewFile(string header)
...
+            {
+                if (i != 0)
+                {
+                    sb.Append('\n');
+                }
+
+                string line = headerLines[i];
+                if (line.StartsWith("--- "))
+                {
+                    line = $"--- a/{newName}";
```
gerhardol, line 120: This is a risky check. I believe that a/, b/ is forced right now  in GE, but Git could change the output format any time.

GitCommands/Patches/PatchManager.cs
```diff
@@ -82,6 +88,52 @@ private static string CorrectHeaderForNewFile(string header)
...
+                }
+                else if (line.Contains($" {oldNameFull} "))
+                {
+                    line = line.Replace($" {oldNameFull} ", $" a/{newName} ");
+                }
+                else if (line.StartsWith("rename from ") || line.StartsWith("rename to ") || line.StartsWith("similarity index "))
```
gerhardol, line 126: Are any of these strings ever translated in Git?

GitUI/Editor/FileViewer.cs
```diff
@@ -73,6 +73,7 @@ private enum ViewMode
         private Func<Task>? _deferShowFunc;
         private readonly ContinuousScrollEventManager _continuousScrollEventManager;
         private FileStatusItem? _viewItem;
+        private bool _confirmResetLines = true;
```
gerhardol, line 76: This should be marked as test specific, a proposal below
Even better would be to hide this in MessageBox.Show() instead with a test accessor
(assuming that there is no short term way to avoid this)
```suggestion

        // This variable is used by tests to avoid popups
        private bool _unitTestConfirmResetLines = true;
```

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCommitTests.cs
```diff
@@ -244,7 +246,7 @@ public void Should_stage_only_filtered_on_StageAll()
 
                 testform.StageAllToolItem.PerformClick();
 
-                var fileNotMatchedByFilterIsStillUnstaged = testform.UnstagedList.AllItems.Where(i => i.Item.Name == "file2.txt").Any();
+                var fileNotMatchedByFilterIsStillUnstaged = testform.UnstagedList.AllItems.Any(i => i.Item.Name == "file2.txt");
```
gerhardol, line 249: Is this required for this PR or is it fixing any other test failures in FormCommit?
(I occasionally see them locally)

GitUI/CommandsDialogs/FormCommit.cs
mdonatas: It shouldn't be awaited as it's fire-and-forget kind of task. I reverted this commit since I was returning this task because I didn't know about `ThreadHelper.JoinPendingOperations();` and that it does exactly what I need - waits for background tasks to complete (at least that's my assumption).

GitUI/CommandsDialogs/FormCommit.cs
mdonatas: Reverted this as it should only be awaited in tests and there is `ThreadHelper.JoinPendingOperations();` for that.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCommitTests.cs
mdonatas: No, I've changed this out of habit for better readability. If that's unacceptable and shouldn't be here as it's unrelated, then I can revert it.

GitCommands/Patches/PatchManager.cs
mdonatas: I suppose that would be a serious breaking change (and not just for GE)

GitCommands/Patches/PatchManager.cs
mdonatas: Since these are part of a diff format used by git and are called extended headers I think it's safe to assume that that wouldn't happen. At least documentation doesn't mention anything of that sort.
https://git-scm.com/docs/diff-format

GitUI/Editor/FileViewer.cs
mdonatas: `MessageBox.Show()` is not under our control and making a wrapper for it seems a daunting task for me at least :)
I like your suggested rename, thanks!

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCommitTests.cs
gerhardol: It is better, keep

GitUI/Editor/FileViewer.cs
gerhardol: Wrap the call to MessageBox.Show() instead or rather switch to TaskDialog

GitCommands/Patches/PatchManager.cs
gerhardol: This can be changed with diff.mnemonicprefix=true. It was forced to false some time ago so a/b is likely stable. 
But number of spaces etc are not porcelain and can be changed without notice.
It could be different in existing Git versions too.

GitCommands/Patches/PatchManager.cs
gerhardol: It is described as a format, it is not even a 'requested' format, certainly not a porcelain interface.
Adding translations in Git will break too.
It could be OK anyway, if the consequences are low (like current behavior in master).
There need to be at least comments about this.

GitUI/Editor/FileViewer.cs
```diff
@@ -1476,7 +1481,7 @@ public void ResetNoncommittedSelectedLines()
                 return;
             }
 
-            if (MessageBox.Show(this, TranslatedStrings.ResetSelectedLinesConfirmation, TranslatedStrings.ResetChangesCaption,
+            if (_unitTestConfirmResetLines && MessageBox.Show(this, TranslatedStrings.ResetSelectedLinesConfirmation, TranslatedStrings.ResetChangesCaption,
```
RussKie, line 1484: We should change `MessageBox.Show` to `TaskDialog` and then we won't have test-only code outside the `TestAccessor`.

RussKie (commented):
👍

GitUI/Editor/FileViewer.cs
mdonatas: Please have a look at `Use TaskDialog instead of MessageBox.Show for reset selected lines confirmation` commit. I've tried to look for examples of manipulating `TaskDialogPage` in the tests but couldn't. I am not sure if this is a correct approach.

GitUI/Editor/FileViewer.cs
mdonatas: Have a look at `Use TaskDialog instead of MessageBox.Show for reset selected lines confirmation`. Thoughts?

RussKie:
I had a very cursory look from a phone, and I think you're on the
right track.

You can see examples in
https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TaskDialogSamples.cs
and
https://github.com/dotnet/samples/tree/main/windowsforms/TaskDialogDemo/cs.

A task dialog page can be autoclosed,  IIRC, though it's been a while since
I looked at the API.

GitCommands/Patches/PatchManager.cs
mdonatas: Would you approve of using regex here? Something like:
`^---\s+(?<a>.+)/(?<aName>.+)$` and
`^\+\+\+\s+(?<b>.+)/(?<bName>.+)$`

to capture a prefix used and a file name as well as make it more resilient to spacing.

GitCommands/Patches/PatchManager.cs
mdonatas: I'd like to push back if I may. Translation argument sounds almost like saying that translating browsers would mean translating HTTP Headers they send (custom ones, not covered by the standard).
These are headers added by the tools for the tools (at least I would assume so). They are meant to be parsable. Translating them would break older versions thus I can't imaging that ever happening. Although if you still insist, I'll add a comment.

mdonatas:
> A task dialog page can be autoclosed, IIRC

Thanks for the pointers!

https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TaskDialogSamples.cs#L217-L233 uses a timer to call `PerformClick` on a desired button to auto-close while I do it immediately upon creation of the dialog. Having looked at examples I think what I've got is correct.

mdonatas:
AppVeyor is failing because of TaskDialogPage and I've been hitting my head against the problem for quite some time now... @RussKie it seems you are somewhat familiar with the root cause https://github.com/dotnet/wpf/issues/6810 any ideas how to solve this here?

UnitTests/CommonTestUtils/ReferenceRepository.cs
```diff
@@ -8,8 +8,8 @@ namespace CommonTestUtils
 {
     public class ReferenceRepository : IDisposable
     {
-        private GitModuleTestHelper _moduleTestHelper;
-        private string _commitHash;
+        private readonly GitModuleTestHelper _moduleTestHelper;
```
mstv, line 11:
```suggestion
        private readonly GitModuleTestHelper _moduleTestHelper = new();
```

GitCommands/Patches/PatchManager.cs
```diff
@@ -82,6 +88,52 @@ private static string CorrectHeaderForNewFile(string header)
...
+            foreach (string line in headerLines)
+            {
+                if (line.StartsWith("+++"))
+                {
+                    newName = line[6..];
+                }
+                else if (line.StartsWith("---"))
+                {
+                    oldNameFull = line[4..];
+                }
```
mstv, line 105: The expected format of these lines should be described in a comment in order to justify the magic numbers.

mstv (commented):
LGTM, have not run

UnitTests/CommonTestUtils/ReferenceRepository.cs
```diff
@@ -45,7 +45,7 @@ public static void ResetRepo(ref ReferenceRepository refRepo)
 
         public GitModule Module => _moduleTestHelper.Module;
 
-        public string CommitHash => _commitHash;
+        public string? CommitHash => _commitHash;
```
mstv, line 48: ❔
```suggestion
        public string? CommitHash { get; private set; }
```

GitCommands/Patches/PatchManager.cs
mdonatas: I've added comments. Please have another look.

GitUI/Editor/FileViewer.cs
```diff
@@ -197,6 +198,16 @@ public FileViewer()
...
             _fullPathResolver = new FullPathResolver(() => Module.WorkingDir);
             SupportLinePatching = false;
+
+            _NO_TRANSLATE_resetSelectedLinesConfirmationDialog = new()
+            {
+                Text = TranslatedStrings.ResetSelectedLinesConfirmation,
+                Caption = TranslatedStrings.ResetChangesCaption,
+                Icon = TaskDialogIcon.Warning,
+                Buttons = { TaskDialogButton.Yes, TaskDialogButton.No },
+                DefaultButton = TaskDialogButton.Yes,
```
RussKie, line 208: Is this a good idea? I _think_ the default button for destructive operations should be the one that doesn't lead to data loss.

IntegrationTests/UI.IntegrationTests/GlobalSetupOnce.cs
```diff
@@ -0,0 +1,17 @@
+﻿using NUnit.Framework;
+
+[SetUpFixture]
+public class GlobalSetupOnce
```
RussKie, line 4: How does this class get used?

GitUI/Editor/FileViewer.cs
mdonatas: `Yes` is currently the default in the `MessageBox` and because I use it **a lot** in my daily work (select text, `r`, `enter`, _repeat_) I didn't want to change the behavior even though I agree with what you said.

IntegrationTests/UI.IntegrationTests/GlobalSetupOnce.cs
RussKie: Thank you

mstv (approved):
Thank you

Issue #10250, opened 2022-10-12
Right click hidden node raises popup


gerhardol:
### Environment

- Git Extensions 4.0.0.15025
- Build e8588e4c048e94d62054911fc09decb8139931f6
- Git 2.38.0.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.10
- DPI 96dpi (no scaling)

### Issue description

GE 4.0 has feedback when a branch etc in the side panel is not in the grid.
This prevents double clicking on a branch to check it out, but ctrl-select for multi is possible. This may be OK, you can still use the menu.
However, it is annoying that right clicking on the branch etc raises the popup. Fortunately, the menu is not closed so you can select from it.

### Steps to reproduce

* Filter branches, for instance Only current branch.
* Rightclick a hidden branch
![image](https://user-images.githubusercontent.com/6248932/195449221-c9280ede-4dd4-4268-9aed-a3f774680ac7.png)

### Did this work in previous version of GitExtensions?

3.5.4 without the popup

gerhardol:
I expected checking for right mousebutton to be sufficient to suppress this, but `MouseButtons` is always None (also at callers).
Maybe related to the changes in #9723. Any input @h0lg ?

````c#
       protected void SelectRevision()
        {
            TreeViewNode.TreeView?.BeginInvoke(new Action(() =>
            {
                string branch = RelatedBranch is null || !RepoObjectsTree.ModifierKeys.HasFlag(Keys.Alt)
                    ? FullPath : RelatedBranch.Substring(startIndex: GitRefName.RefsRemotesPrefix.Length);
/* ->*/                bool showPopup = !RepoObjectsTree.MouseButtons.HasFlag(MouseButtons.Right);
                UICommands.BrowseGoToRef(branch, showNoRevisionMsg: showPopup, toggleSelection: RepoObjectsTree.ModifierKeys.HasFlag(Keys.Control));
                TreeViewNode.TreeView?.Focus();
            }));
        }
```

gerhardol:
> @gerhardol I wasn't aware of the static `Control.MouseButtons` and didn't handle mouse buttons except for in `RepoObjectsTree.OnNodeClick()`. But wherever I find `Control.MouseButtons` used, it's compared to using `==` instead of `.HasFlag()`. Have you tried that?

I am pretty sure I tried that - no mouse flag was set at all.

vbjay:
[...]

PR #10161, opened 2022-08-20
FormStash: Display index of stashes

pmiossec:
so that the user could more easily track where he is...

(I always found painful to use the stashes ComboBox when you have a lot of stashs and stash index really help me to find the current selected stash once you scroll in the list --because the ComboBox doesn't highlight the current selected item 😢--)

## Proposed changes

- Display index of stashes (like it is done in the git command output)

### Before

Let's play a game: Could you find the currently selected one?

image

### After

And now?

image

## Test methodology 

- Manual
- 
## Test environment(s) 

- Git Extensions 33.33.33
- Build c92862e1c99a55af34ff69b2698742da81bfbba9
- Git 2.35.1.windows.2 (recommended: 2.37.1 or later)
- Microsoft Windows NT 10.0.19042.0
- .NET 6.0.8
- DPI 96dpi (no scaling)

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10258, opened 2022-10-14
SidePanel: Show Stashes

### LINKED ISSUE #5460: Show stashes tree in sidebar

gerhardol:
Fixes #5460 
Follow-up to #10138 (that added stashes to the grid)

## Proposed changes

1. Side panel: Refactor classes and names

Renamed and restructured Node/Tree classes.
Some methods were added on parents but only used in leaves
(especially for submodule vs the "revision grid" branches/tags).
Example for a moved method is aheadBehind for LocalBranchNode.

Renamed and added some classes:
* BaseRevisionTree and BaseBranchNode.cs -> BaseRevisionNode.cs
  for methods pointing to revisions in the grid (i.e. all but submodules).
* BaseRefTree for Refs (Branches/Tags).

Methods were moved to the class where they were used

2. SidePanel:  Visibility for branches and tags was not updated.

If the revision grid had another filter than current branch or
filtered branches, the branches/tags were always set as visible.

The refs are always set as visible at creation and set as
not visible after the grid has loaded.

3.  SidePanel: Show Stashes

Context menu to open/apply/pop/drop as well as stash/stash-staged/manage.

4. RemoteBranches: Show ahead/behind for remotes

Supprt Alt-Click to select the tracking local branches

This is a candidate for 4.0

### Before

image

Revision filter hiding these branches
image

### After

image

image

image

image

Remote had no ahaed/behind info (no alt-click)
image
image

## Test methodology 

Tests are adapted to additions.

## Merge strategy

Rebase merge

----

### DIFF
--- a/GitCommands/Git/AheadBehindDataProvider.cs
+++ b/GitCommands/Git/AheadBehindDataProvider.cs
@@ -1,4 +1,5 @@
-﻿using System.Text;
+﻿using System.Diagnostics;
+using System.Text;
 using System.Text.RegularExpressions;
 using GitExtUtils;
 using GitUIPluginInterfaces;
@@ -8,6 +9,7 @@ namespace GitCommands.Git
     public interface IAheadBehindDataProvider
     {
         IDictionary<string, AheadBehindData>? GetData(string branchName = "");
+        void ResetCache();
     }
 
     public class AheadBehindDataProvider : IAheadBehindDataProvider
@@ -23,10 +25,18 @@ namespace GitCommands.Git
                    (?<remote_p>.*?)::(?<remote_u>.*?)::(?<branch>.*)$",
                 RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.IgnorePatternWhitespace | RegexOptions.ExplicitCapture);
         private readonly string _refFormat = @"%(push:track,nobracket)::%(upstream:track,nobracket)::%(push)::%(upstream)::%(refname:short)";
+        private Lazy<IDictionary<string, AheadBehindData>?> _lazyData;
+        private string _branchName;
 
         public AheadBehindDataProvider(Func<IExecutable> getGitExecutable)
         {
             _getGitExecutable = getGitExecutable;
+        }
+
+        public void ResetCache()
+        {
+            _lazyData = null;
+            _branchName = null;
         }
 
         public IDictionary<string, AheadBehindData>? GetData(string branchName = "")
@@ -36,7 +46,16 @@ namespace GitCommands.Git
                 return null;
             }
 
-            return GetData(null, branchName);
+            // Callers setting branchname has the responisibility to ensure that not all are needed
+            if (string.IsNullOrWhiteSpace(branchName) && !string.IsNullOrWhiteSpace(_branchName))
+            {
+                // Debug.Fail($"Unexpectedly call for all branches after cache filled with specific branch {_branchName}");
+                ResetCache();
+            }
+
+            _lazyData ??= new(() => GetData(null, branchName));
+            _branchName = branchName;
+            return _lazyData.Value;
         }
 
         // This method is required to facilitate unit tests
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -1884,6 +1884,12 @@ namespace GitCommands
             set => SetBool("RepoObjectsTree.ShowTags", value);
         }
 
+        public static bool RepoObjectsTreeShowStashes
+        {
+            get => GetBool("RepoObjectsTree.ShowStashes", true);
+            set => SetBool("RepoObjectsTree.ShowStashes", value);
+        }
+
         public static bool RepoObjectsTreeShowSubmodules
         {
             get => GetBool("RepoObjectsTree.ShowSubmodules", true);
@@ -1912,6 +1918,12 @@ namespace GitCommands
         {
             get => GetInt("RepoObjectsTree.SubmodulesIndex", 3);
             set => SetInt("RepoObjectsTree.SubmodulesIndex", value);
+        }
+
+        public static int RepoObjectsTreeStashesIndex
+        {
+            get => GetInt("RepoObjectsTree.StashesIndex", 4);
+            set => SetInt("RepoObjectsTree.StashesIndex", value);
         }
 
         public static bool BlameDisplayAuthorFirst
--- a/GitUI/BranchTreePanel/BaseBranchLeafNode.cs
+++ b/GitUI/BranchTreePanel/BaseBranchLeafNode.cs
@@ -3,7 +3,7 @@ using GitUIPluginInterfaces;
 
 namespace GitUI.BranchTreePanel
 {
-    internal abstract class BaseBranchLeafNode : BaseBranchNode
+    internal abstract class BaseBranchLeafNode : BaseRevisionNode
     {
         private readonly string _imageKeyMerged;
         private readonly string _imageKeyUnmerged;
@@ -34,9 +34,7 @@ namespace GitUI.BranchTreePanel
             }
         }
 
-        public ObjectId? ObjectId { get; }
-
-        protected override void ApplyStyle()
+        public override void ApplyStyle()
         {
             base.ApplyStyle();
 
--- a/GitUI/BranchTreePanel/BaseBranchLeafNode.cs
+++ b/GitUI/BranchTreePanel/BaseBranchLeafNode.cs
@@ -18,6 +18,10 @@ namespace GitUI.BranchTreePanel
             _imageKeyMerged = imageKeyMerged;
         }
 
+        protected string? AheadBehind { get; set; }
+
+        protected string? RelatedBranch { get; set; }
+
         public bool IsMerged
         {
             get => _isMerged;
@@ -50,6 +54,29 @@ namespace GitUI.BranchTreePanel
             {
                 TreeViewNode.ToolTipText = string.Format(TranslatedStrings.ContainedInCurrentCommit, Name);
             }
+        }
+
+        public void UpdateAheadBehind(string aheadBehindData, string relatedBranch)
+        {
+            AheadBehind = aheadBehindData;
+            RelatedBranch = relatedBranch;
+        }
+
+        protected override string DisplayText()
+        {
+            return string.IsNullOrEmpty(AheadBehind) ? Name : $"{Name} ({AheadBehind})";
+        }
+
+        protected override void SelectRevision()
+        {
+            TreeViewNode.TreeView?.BeginInvoke(() =>
+            {
+                string branch = RelatedBranch is null || !Control.ModifierKeys.HasFlag(Keys.Alt)
+                    ? FullPath
+                    : RelatedBranch;
+                UICommands.BrowseGoToRef(branch, showNoRevisionMsg: true, toggleSelection: Control.ModifierKeys.HasFlag(Keys.Control));
+                TreeViewNode.TreeView?.Focus();
+            });
         }
     }
 }
--- a/GitUI/BranchTreePanel/BaseRefTree.cs
+++ b/GitUI/BranchTreePanel/BaseRefTree.cs
@@ -1,0 +1,91 @@
+﻿using GitUI.UserControls.RevisionGrid;
+using GitUIPluginInterfaces;
+using Microsoft.VisualStudio.Threading;
+
+namespace GitUI.BranchTreePanel
+{
+    internal abstract class BaseRefTree : BaseRevisionTree
+    {
+        // A flag to indicate whether the data is being filtered (e.g. Show Current Branch Only).
+        private protected System.Threading.AsyncLocal<bool> IsFiltering = new();
+
+        // Retains the list of currently loaded refs (branches/tags).
+        // This is needed to apply filtering without reloading the data.
+        // Whether or not force the reload of data is controlled by <see cref="_isFiltering"/> flag.
+        protected IReadOnlyList<IGitRef>? _loadedRefs;
+
+        protected readonly RefsFilter _filter;
+
+        protected BaseRefTree(TreeNode treeNode, IGitUICommandsSource uiCommands, ICheckRefs refsSource, RefsFilter filter)
+            : base(treeNode, uiCommands, refsSource)
+        {
+            _filter = filter;
+        }
+
+        protected override void OnAttached()
+        {
+            IsFiltering.Value = false;
+            base.OnAttached();
+        }
+
+        private async Task<Nodes> LoadNodesAsync(CancellationToken token, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
+        {
+            await TaskScheduler.Default;
+            token.ThrowIfCancellationRequested();
+
+            if (!IsFiltering.Value || _loadedRefs is null)
+            {
+                _loadedRefs = getRefs(_filter);
+                token.ThrowIfCancellationRequested();
+            }
+
+            return FillTree(_loadedRefs, token);
+        }
+
+        protected abstract Nodes FillTree(IReadOnlyList<IGitRef> branches, CancellationToken token);
+
+        /// <summary>
+        /// Requests (from FormBrowse) to refresh the data tree and to apply filtering, if necessary.
+        /// </summary>
+        /// <param name="getRefs">Function to get refs.</param>
+        /// <param name="forceRefresh">Refresh may be required as references may have been changed.</param>
+        /// <param name="isFiltering">
+        ///  <see langword="true"/>, if the data is being filtered; otherwise <see langword="false"/>.
+        /// </param>
+        internal void Refresh(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs, bool forceRefresh, bool isFiltering)
+        {
+            if (!IsAttached)
+            {
+                return;
+            }
+
+            // If we're not currently filtering and no need to filter now -> exit.
+            // Else we need to iterate over the list and rebind the tree - whilst there
+            // could be a situation whether a user just refreshed the grid, there could
+            // also be a situation where the user applied a different filter, or checked
+            // out a different ref (e.g. a branch or commit), and we have a different
+            // set of branches to show/hide.
+            if (!forceRefresh && (!isFiltering && !IsFiltering.Value))
+            {
+                return;
+            }
+
+            IsFiltering.Value = isFiltering;
+            Refresh(getRefs);
+        }
+
+        /// <summary>
+        /// Requests to refresh the data tree and to apply filtering, if necessary.
+        /// </summary>
+        protected internal void Refresh(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
+        {
+            // Break the local cache to ensure the data is requeried to reflect the required sort order.
+            _loadedRefs = null;
+
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await ReloadNodesAsync(LoadNodesAsync, getRefs);
+            });
+        }
+    }
+}
--- a/GitUI/BranchTreePanel/BaseRefTree.cs
+++ b/GitUI/BranchTreePanel/BaseRefTree.cs
@@ -7,24 +7,24 @@ namespace GitUI.BranchTreePanel
     internal abstract class BaseRefTree : BaseRevisionTree
     {
         // A flag to indicate whether the data is being filtered (e.g. Show Current Branch Only).
-        private protected System.Threading.AsyncLocal<bool> IsFiltering = new();
+        private bool _isFiltering = false;
 
         // Retains the list of currently loaded refs (branches/tags).
         // This is needed to apply filtering without reloading the data.
         // Whether or not force the reload of data is controlled by <see cref="_isFiltering"/> flag.
         protected IReadOnlyList<IGitRef>? _loadedRefs;
 
-        protected readonly RefsFilter _filter;
+        protected readonly RefsFilter _refsFilter;
 
         protected BaseRefTree(TreeNode treeNode, IGitUICommandsSource uiCommands, ICheckRefs refsSource, RefsFilter filter)
             : base(treeNode, uiCommands, refsSource)
         {
-            _filter = filter;
+            _refsFilter = filter;
         }
 
         protected override void OnAttached()
         {
-            IsFiltering.Value = false;
+            _loadedRefs = null;
             base.OnAttached();
         }
 
@@ -33,9 +33,9 @@ namespace GitUI.BranchTreePanel
             await TaskScheduler.Default;
             token.ThrowIfCancellationRequested();
 
-            if (!IsFiltering.Value || _loadedRefs is null)
+            if (_loadedRefs is null)
             {
-                _loadedRefs = getRefs(_filter);
+                _loadedRefs = getRefs(_refsFilter);
                 token.ThrowIfCancellationRequested();
             }
 
@@ -65,12 +65,12 @@ namespace GitUI.BranchTreePanel
             // also be a situation where the user applied a different filter, or checked
             // out a different ref (e.g. a branch or commit), and we have a different
             // set of branches to show/hide.
-            if (!forceRefresh && (!isFiltering && !IsFiltering.Value))
+            if (_loadedRefs is not null && !forceRefresh && (!isFiltering && !_isFiltering))
             {
                 return;
             }
 
-            IsFiltering.Value = isFiltering;
+            _isFiltering = isFiltering;
             Refresh(getRefs);
         }
 
@@ -86,6 +86,16 @@ namespace GitUI.BranchTreePanel
             {
                 await ReloadNodesAsync(LoadNodesAsync, getRefs);
             });
+        }
+
+        internal override void UpdateVisibility()
+        {
+            if (!IsAttached || !_isFiltering)
+            {
+                return;
+            }
+
+            base.UpdateVisibility();
         }
     }
 }
--- a/GitUI/BranchTreePanel/LocalBranchNode.cs
+++ b/GitUI/BranchTreePanel/LocalBranchNode.cs
@@ -1,4 +1,5 @@
 ﻿using System.Diagnostics;
+using GitCommands;
 using GitUI.BranchTreePanel.Interfaces;
 using GitUI.Properties;
 using GitUIPluginInterfaces;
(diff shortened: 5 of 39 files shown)


GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -234,22 +237,39 @@ bool IsOverride(MethodInfo m)
...
-        public void Refresh(bool isFiltering, bool forceRefresh, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
+        /// <param name="getRefs">Function to get refs.</param>
+        public void RefreshWhenLoading(bool isFiltering, bool forceRefresh, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
         {
             _branchesTree.Refresh(isFiltering, forceRefresh, getRefs);
             _remotesTree.Refresh(isFiltering, forceRefresh, getRefs);
             _tagTree.Refresh(isFiltering, forceRefresh, getRefs);
+
+            // Clear current stash
+            _stashTree.Refresh(new(() => new List<GitRevision>()));
```
RussKie, line 255: Could this be?
```suggestion
            _stashTree.Refresh(new(() => Array.Empty<GitRevision>()));
```

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -382,6 +403,18 @@ private void CreateSubmodules()
...
+        private void CreateStashes()
+        {
+            TreeNode rootNode = new(TranslatedStrings.Stashes)
+            {
+                Name = TranslatedStrings.Stashes,
+                ImageKey = nameof(Images.Stash),
+                SelectedImageKey = nameof(Images.Stash)
+            };
+
+            _stashTree = new StashTree(rootNode, UICommandsSource, _refsSource);
```
RussKie, line 415: [nit]
```suggestion
            _stashTree = new(rootNode, UICommandsSource, _refsSource);
```

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -887,6 +887,9 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
             Lazy<IReadOnlyList<IGitRef>> getUnfilteredRefs = new(() => (getRefs ?? capturedModule.GetRefs)(RefsFilter.NoFilter));
             _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(getUnfilteredRefs.Value));
 
+            // Get the "main" stash commit, including the reflog selector
+            Lazy<IReadOnlyCollection<GitRevision>> getStashRevs = new(() => (!AppSettings.ShowStashes ? new List<GitRevision>() : new RevisionReader(capturedModule, hasReflogSelector: true).GetStashes(cancellationToken)));
```
RussKie, line 891:
```suggestion
            Lazy<IReadOnlyCollection<GitRevision>> getStashRevs =
                new(() => (!AppSettings.ShowStashes 
                    ? Array.Empty<GitRevision>()
                    : new RevisionReader(capturedModule, hasReflogSelector: true).GetStashes(cancellationToken)));
```

I also suspect that one set of parens is superficial.

GitUI/UserControls/RevisionGrid/RevisionLoadEventArgs.cs
```diff
@@ -4,12 +4,14 @@ namespace GitUI.UserControls.RevisionGrid
 {
     public class RevisionLoadEventArgs : GitUIEventArgs
     {
-        public RevisionLoadEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)
+        public RevisionLoadEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, bool forceRefresh, Lazy<IReadOnlyCollection<GitRevision>> getStashRevs)
```
RussKie, line 7:
```suggestion
        public RevisionLoadEventArgs(IWin32Window? ownerForm, IGitUICommands gitUICommands, Lazy<IReadOnlyList<IGitRef>> getRefs, Lazy<IReadOnlyCollection<GitRevision>> getStashRevs, bool forceRefresh)
```

RussKie (approved):
👍 with few nits and suggestions

RussKie:
> image

I don't think it's improving the UI to have the same icon repeated, it's likely achieving the opposite result. We don't have the icons in the toolbar either for the same reason:
image

We either need dedicated icons, or have no icons at all.

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: When we talked about this feature, we have previously said that the type should be stated if not on left side of '='
These rules should be documented...
I prefer the suggestion, but do not change right now.

GitUI/BranchTreePanel/StashNode.cs
```diff
@@ -0,0 +1,70 @@
...
+using GitUI.Properties;
+using GitUIPluginInterfaces;
+
+namespace GitUI.BranchTreePanel
+{
+    [DebuggerDisplay("(Tag) FullPath = {FullPath}, Hash = {ObjectId}, Visible: {Visible}")]
+    internal sealed class StashNode : BaseBranchNode
+    {
+        public StashNode(Tree tree, in ObjectId? objectId, string reflogSelector, string subject, bool visible)
+            : base(tree, reflogSelector.RemovePrefix("refs/"), visible)
```
gerhardol, line 14: #10161 changed to display this in FormStash
image

The RevGrid show the complete ref (excluding refs/)
image

Should the list present as in FormStash?
```suggestion
            : base(tree, reflogSelector.RemovePrefix("refs/").RemovePrefix("stash"), visible)
```
image

mstv: I have no strong opinion regarding the colon - not really needed but for consistency...

```suggestion
            : base(tree, reflogSelector.RemovePrefix(GitRefName.RefsStashPrefix), visible)
```

GitUI/BranchTreePanel/StashTree.cs
```diff
@@ -0,0 +1,94 @@
...
+        private Nodes FillStashTree(IReadOnlyList<GitRevision> stashes, CancellationToken token)
+        {
+            Nodes nodes = new(this);
+            Dictionary<string, BaseBranchNode> pathToNodes = new();
+
+            foreach (GitRevision stash in stashes)
+            {
+                token.ThrowIfCancellationRequested();
+
+                // Stashes does not supprt filtering, but stashes may not be visible
```
mstv, line 52:
```suggestion
                // Stashes does not support filtering, but stashes may not be visible
```

GitUI/BranchTreePanel/StashTree.cs
```diff
@@ -0,0 +1,94 @@
...
+                bool isVisible = stash.ObjectId is not null && _refsSource.Contains(stash.ObjectId);
+                StashNode node = new(this, stash.ObjectId, stash.ReflogSelector, stash.Subject, isVisible);
+                var parent = node.CreateRootNode(pathToNodes, (tree, parentPath) => new BasePathNode(tree, parentPath));
```
mstv, line 55:
```suggestion
                Node? parent = node.CreateRootNode(pathToNodes, (tree, parentPath) => new BasePathNode(tree, parentPath));
```

mstv (commented):
LGTM, just nits, have not run yet

GitUI/BranchTreePanel/StashNode.cs
gerhardol: Changed this, updated screenshot

mstv (approved):
Thank you, very useful

We should not attempt to select unavailable stash revisions though:

image

gerhardol:
> We should not attempt to select unavailable stash revisions though:

You can still "open stash" after dismissing the popup.
The popup should not be raised when right clicking: #10250
This disables double clicking

vbjay:
Check button and let through for double click of left / primary button.

gerhardol:
Added an incomplete fix for visibility for branches, see updated description.
Normally, the grid is updated just before the branches etc were added. If non-default sorting is used, the sidepanel is updated before the grid and the visibility is incorrect.
For stashes, this was always the case and I added stashes after loading.
The handling was changed for stashes, creating the sidepanel when loading the grid but updating the visibility after the grid was loaded.
Loading earlier is not important but a similar change should be done for branches etc, but there will have to be some more changes.
This second part could be done in a sepate PR too, to get translation changes out.

A related fix:
* Visibility for branches and tags were incorrect if filters other than
branch filers were used.

GitUI/BranchTreePanel/StashTree.cs
```diff
@@ -63,6 +62,28 @@ private Nodes FillStashTree(IReadOnlyList<GitRevision> stashes, CancellationToke
             return nodes;
         }
 
+        internal void UpdateVisibility()
+        {
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
```
mstv, line 67: ❔ `Invoke`

mstv (approved):

gerhardol:
It is not obvious how to restructure the info in a good way (to not add null ObjectId to folders, may need to do that anyway or duplicate some code in tags/local/remote/stash).
There is some renaming I would like to do before merging to reflect usage better, will wait a little until I decide...

> Check button and let through for double click of left / primary button.

I guess this was related to https://github.com/gitextensions/gitextensions/pull/10258#pullrequestreview-1143339265, https://github.com/gitextensions/gitextensions/issues/10250
I tried to debug that but did not find the issue.

GitUI/BranchTreePanel/StashTree.cs
gerhardol: Make this an EventHandler?
What is the benefit in this scenario?

GitUI/BranchTreePanel/RepoObjectsTree.cs
RussKie: We shouldn't use `var` if the type is not obvious from the expression. When newing up the type could be omitted.

This is a non-blocking suggestion.

GitUI/BranchTreePanel/RepoObjectsTree.cs
```diff
@@ -234,22 +237,41 @@ bool IsOverride(MethodInfo m)
...
         /// </summary>
         /// <param name="isFiltering">
         ///  <see langword="true"/>, if the data is being filtered; otherwise <see langword="false"/>.
         /// </param>
-        /// <param name="getRefs">Function to get refs.</param>
         /// <param name="forceRefresh">Refresh may be required as references may have been changed.</param>
-        public void Refresh(bool isFiltering, bool forceRefresh, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs)
+        /// <param name="getRefs">Function to get refs.</param>
+        /// <param name="getStashRevs">Lazy accessor for stash commits.</param>
+        public void RefreshWhenLoading(bool isFiltering, bool forceRefresh, Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs, Lazy<IReadOnlyCollection<GitRevision>> getStashRevs)
```
RussKie, line 248: I find `RefreshWhenLoading` and `RefreshAfterLoaded` confusing. Without reading the xml-docs I couldn't quite figure out what these two do...
I'm having troubles with the meaning of "refreshing while data is loading". `RevisionGrid.RevisionsLoading` event is raised _before_ data is loaded, and `RevisionGrid.RevisionsLoaded` is raised _after_ the data has been loaded and displayed. According to the comment, we have added `RevisionGrid.RevisionsLoading` event specifically to refresh the left panel.

Was there a specific reason why you thought the two methods were necessary? Couldn't we have the single method with an additional flag to indicate whether the data _is being_ or _has been_ loaded?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -887,6 +887,12 @@ public void PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> get
             Lazy<IReadOnlyList<IGitRef>> getUnfilteredRefs = new(() => (getRefs ?? capturedModule.GetRefs)(RefsFilter.NoFilter));
             _ambiguousRefs = new(() => GitRef.GetAmbiguousRefNames(getUnfilteredRefs.Value));
 
+            // Get the "main" stash commit, including the reflog selector
+            Lazy<IReadOnlyCollection<GitRevision>> getStashRevs = new(() =>
+                !AppSettings.ShowStashes
+                ? new List<GitRevision>()
```
RussKie, line 893: Does it have to be `new List<>`?

GitUI/BranchTreePanel/RepoObjectsTree.cs
RussKie, line 248: Looks like in different parts we have different order of the same arguments:
image

gerhardol: Before it was Refresh() in many situations which was more confusing.
The first call is before the revision grid is available. The data to add is there, but not if a revision is visible or not, which is available after loading. RefreshAfterLoaded() do not need any parameters right now, it just updates the visibility.

The previous commit updated stash info after loading only. That strategy could be used for all 'commit' (branches, tags, stashes but not submodules), but then updating of will be delayed.

So I prefer two methods, possibly with updated arguments.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Probably not
This is some us for something done once in a process that may take several seconds, so there is not much point in optimizing.
(I have spent a lot of time optimizing also single ms for 4.0, so performance is not unimportant but focus where it makes a difference.)

mdonatas: In this case it's not even about performance. This simply looks wrong. To something which is **ReadOnly** a collection meant for manipulation is assigned. This raises questions and wastes others time while trying to comprehend the meaning of the code.
`Array.Empty<GitRevision>()` set a good tone and communicates intent clearly.

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: Renaming to `RefreshRevisionsLoading` and `RefreshRevisionsLoaded` so similar to the events.
Keeping the arguments the same, as isFiltering is needed after loading (to skip updating visibility).

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
gerhardol: Updating
(I do not find Array.Empty to look better in any way.)

gerhardol:
Rebased and added three commits:
* review comments
* refactoring without functional changes
* Visibility for branches and tags were incorrect if filters other than branch filers were used.

```
SidePanel: Show Stashes

Context menu to open/apply/pop/drop.

Renamed and restructured Node/Tree classes:
* BaseRevisionTree and BaseBranchNode.cs -> BaseRevisionNode.cs
  for methods pointing to revisions in the grid (i.e. all but submodules).
* BaseRefTree for Refs (Branches/Tags).

Methods were moved to the class where they were used
(like aheadBehind in the LocalBranchNode).

Some renaming to clarify use

* Visibility for branches and tags were incorrect if filters other than
branch filers were used.
```

The Latest refactoring changes some that was added in the first commit and I would like to include it.
I could rebase to make the refactoring first, then 'filter' visibility fixes, then Stash.

There is one omission in features too:

image
image

GitUI/BranchTreePanel/RepoObjectsTree.cs
RussKie: > I will not name who set the original order in the event handlers, the side panel methods or who requested changes previously in this PR. We all have a right to change our minds.

😆

GitUI/BranchTreePanel/StashTree.cs
mstv: I think `Invoke` is the short of `RunAsync(await SwitchToMainThread(); actual call;)`. But I don't mind using JTF for all async operations.

GitUI/BranchTreePanel/LocalBranchNode.cs
```diff
@@ -76,5 +81,27 @@ public bool Rename()
...
+        {
+            return string.IsNullOrEmpty(AheadBehind) ? Name : $"{Name} ({AheadBehind})";
+        }
+
+        protected override void SelectRevision()
+        {
+            TreeViewNode.TreeView?.BeginInvoke(new Action(() =>
+            {
+                string branch = RelatedBranch is null || !RepoObjectsTree.ModifierKeys.HasFlag(Keys.Alt)
+                    ? FullPath : RelatedBranch.Substring(startIndex: GitRefName.RefsRemotesPrefix.Length);
```
mstv: This used to work for remote branches, too...

GitCommands/Settings/AppSettings.cs
```diff
@@ -1914,6 +1920,12 @@ public static int RepoObjectsTreeSubmodulesIndex
             set => SetInt("RepoObjectsTree.SubmodulesIndex", value);
         }
 
+        public static int RepoObjectsTreeStashesIndex
+        {
+            get => GetInt("RepoObjectsTree.StashesIndex", 4);
```
mstv, line 1925: I have moved it to the top.
(The weird auto-reload of settings without UI update confused me once again. If another GE instance denies to work like for `Move up`, hit `F5`.) 
The `Stashes` tree is collapsed by default. So it does not consume much space but is at hand quickly.
(not requesting a change)

mstv (commented):
Clear refactoring 👍

I like feature/rot-stashes-2.

GitCommands/Settings/AppSettings.cs
gerhardol: The index is stored in settings, so it is safer to store new occurrences last.
Do you request a change?

GitUI/BranchTreePanel/LocalBranchNode.cs
gerhardol: Related branch was not set only for Local branch. I was thinking about this myself.
Not tonight, probably separate PR

gerhardol:
Three more commits with one fix, one cleanup in the optimization and addition of context menu on the root stash (screenshot updated).

> Clear refactoring
I like feature/rot-stashes-2.

Squashed and refactored into new commits with refactoring, visibility for branches etc at filtering and stash related separated.
This is how I would like to merge this.

GitCommands/Settings/AppSettings.cs
mstv: >> (not requesting a change)
>
> Do you request a change?

I just wanted to raise this point. Perhaps others vote for a change.

> The index is stored in settings, so it is safer to store new occurrences last.

Yes, else it might become difficult to handle.

GitUI/BranchTreePanel/LocalBranchNode.cs
mstv: OK

gerhardol: I doubt that this has ever worked.
The solution should also eliminate the additional Git command needed for push button vs left panel (as remote/local should share the info anyway). The aheadBehindProvider should be made lazy or the info be provided as the refs. Lazy data reset when the refs are refreshed.

Slightly offtopic, but should remote branches be decorated in some way?
The easiest is to have exactly the same info as locals. The info could be prepended with "local: " or so but that is annoying when you know about it. Tooltip could add details.

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -8,6 +9,7 @@ namespace GitCommands.Git
     public interface IAheadBehindDataProvider
     {
         IDictionary<string, AheadBehindData>? GetData(string branchName = "");
+        void InitLazy();
```
RussKie, line 12:
```suggestion
        void Initialize();
```

Whether it's lazy or not is an implementation detail.

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -23,20 +25,37 @@ public class AheadBehindDataProvider : IAheadBehindDataProvider
...
             if (!AppSettings.ShowAheadBehindData)
             {
                 return null;
             }
 
-            return GetData(null, branchName);
+            // Callers setting branchname has the responisibility to ensure that not all are needed
+            if (string.IsNullOrWhiteSpace(branchName) && !string.IsNullOrWhiteSpace(_branchName))
+            {
+                // Debug.Fail($"Unexpectedly call for all branches after cache filled with specific branch {_branchName}");
```
RussKie, line 52: Is this needed?

GitUI/BranchTreePanel/BaseBranchLeafNode.cs
```diff
@@ -51,5 +55,28 @@ public override void ApplyStyle()
...
+        }
+
+        protected override string DisplayText()
+        {
+            return string.IsNullOrEmpty(AheadBehind) ? Name : $"{Name} ({AheadBehind})";
+        }
+
+        protected override void SelectRevision()
+        {
+            TreeViewNode.TreeView?.BeginInvoke(new Action(() =>
```
RussKie, line 72: It should be possible to write it as:
```suggestion
            TreeViewNode.TreeView?.BeginInvoke(() =>
```

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: ResetCache() or just Reset may be better then - it is resetting the cached result.

GitCommands/Git/AheadBehindDataProvider.cs
gerhardol: The commented out code is to document the behavor. I would have liked to enable the Debug if that had been working in .NET > 4.
For now it auto-resets the cache to get results for the branches.
Should just occur in tests.

GitUI/BranchTreePanel/BaseBranchLeafNode.cs
gerhardol: Will change
This was just moved from BaseBranchNode.cs
I assume this goes into the "refactoring - no changes" commit

gerhardol:
Pushed some more refactoring - had not planned to do that in this PR, but pushing it here to avoid changing the same class twice.
Some duplicate code i inherited functions could be removed.

The actual change for https://github.com/gitextensions/gitextensions/pull/10258#discussion_r999961425
was then quite small and was included here.

feature/rot-stashes-2 is updated

I still want to include this in 4.0

RussKie:
I'd like to get v4 out of the door. We keep dragging on and on with that...

gerhardol:
> I'd like to get v4 out of the door. We keep dragging on and on with that...

Agree in general, but this 'completes' a feature

RussKie:
OK. Let's get it merged then

GitUI/BranchTreePanel/RepoObjectsTree.cs
gerhardol: I renamed the repoObjectsTree method `Refresh` to `RevisionGrid.RevisionsLoading -> RefreshRevisionsLoading` and added `RevisionGrid.RevisionsLoaded -> RefreshRevisionsLoaded`, called from .
The call to RepoObjectTree is now wrapped in `FormBrowse.RefreshLeftPanel`, but keeping the changed names.

gerhardol:
Need to rebased on master,  #10283
I would like to get a review from @mstv too.
The cache handling in part 4 (that saves a Git call ) can be done in other way, I believe this OK

gerhardol:
Squashed to descriptive commits, rebased on master and #10283. Especially #10285 required some redesign.
Some other refining too.

See changes also due to other changes in `feature/rot-stashes-2` that now contain the old PR branch.

gerhardol:
Plan to rebase merge this tomorrow, unless there are review comments

GitUI/BranchTreePanel/StashTree.cs
```diff
@@ -0,0 +1,95 @@
...
+                        loadNodesTask = LoadNodesAsync(token, getStashRevs);
+                    }
+                    finally
+                    {
+                        _updateSemaphore.Release();
+                    }
+
+                    return loadNodesTask;
+                }, null).ConfigureAwait(false);
+            });
```
mstv, line 39: ❔ Do it like in BaseRevisionTree?
```
    <snip>
}).FileAndForget();
```
---

General question: Might `ConfigureAwait()` be the cause why the app sometimes crashes without any popup?

mstv (commented):
Just reviewed on GH

GitCommands/Git/AheadBehindDataProvider.cs
```diff
@@ -23,20 +25,37 @@ public class AheadBehindDataProvider : IAheadBehindDataProvider
...
+                ResetCache();
+            }
+
+            _lazyData ??= new(() => GetData(null, branchName));
+            _branchName = branchName;
+            return _lazyData.Value;
```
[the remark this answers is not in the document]
gerhardol: It is a way of syncing.
Something like `_data ??= GetData(null, branchName);` may cause GetData to be called three times, but using Lazy here blocks this.
Could be used with a lock() or so instead.
Can add a comment.

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        internal override void UpdateVisibility()
+        {
+            if (!IsAttached || !_isFiltering)
+            {
+                return;
+            }
+
+            base.UpdateVisibility();
+        }
```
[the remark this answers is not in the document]
gerhardol: Reloading is always done before UpdateVisibility(), the visibility for refs is always set to true.

GitUI/BranchTreePanel/StashTree.cs
gerhardol: Same for Branches/Tags in BaseRefTree.cs, so not changed.
I guess it could be changed.

> General question: Might ConfigureAwait() be the cause why the app sometimes crashes without any popup?

Maybe Debug.Assert or Debug.Fail that fires if you run in Debug. Hard to debug. Maybe fixed in .net7
https://github.com/dotnet/winforms/issues/3739

pmiossec:
Not a critic but more a remark...

I had to think some longs seconds before understanding why some recent stashes are greyed:
image

before understanding that are the ones based on a deleted commit only reachable when displaying the reflog.

Even if it is right, it is not straightforward and I think we could expect some questions on that after the release 😉 (maybe this comment will be a good answer for those used to search before opening an issue...)

gerhardol:
> I had to think some longs seconds before understanding why some recent stashes are greyed: 

There could be a tooltip to enable reflog to see them in the grid, or doubleclick to open the stash viewer (that has improved navigation in 4.0 too).

I assume you expect all to be seen with _All_ branches, branches not visible when branches are filtered have the same decoration.

Issue #10254, opened 2022-10-14
Fix annoying collapsing of dropdown menus from tool bar buttons


TanukiSharp:
### Feature description

Add an option to have the dropdown menus not to collapse when mouse cursor moves away from the open button.
This behavior is super annoying, so having it behind an option would let the user to choose which behavior to have.

This is related to issue #3832 it used to be fixed, but it is now back. I updated the closed issue but it remained unanswered. The fix done was a kind of a hack, so I will propose a better fix in another PR soon.

The behavior of the `ToolStripDropDownItem` is not a bug in `GitExtensions`, I created a vanilla WinForms application to test this behavior, and this is the default behavior, so technically speaking, it's a feature request.

### Environment

- Git Extensions 4.0.0.15025
- Build e8588e4c048e94d62054911fc09decb8139931f6
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.22621.0
- .NET 6.0.9
- DPI 96dpi (no scaling)

gerhardol:
The menu stays open for some seconds for me.

> Add an option to have the dropdown menus not to collapse when mouse cursor moves away from the open button. This behavior is super annoying, so having it behind an option would let the user to choose which behavior to have.

Try to make this so an option is not needed. GE has too many options already.

TanukiSharp:
On my side they close immediately.
As you can see on the video below, menus close as soon as there is a mouse interaction with another item in the same ToolStrip container, but not with other controls, such as menu bar. This is clearly emphasized from 0:10' to 0:15' in the video.

https://user-images.githubusercontent.com/4018313/195963038-b278ad59-f244-49c4-bd52-ba9bf8c11af8.mp4

The only way I found to disable mouse interaction with other items in the ToolStrip container is to disable them. I couldn't find a way to just prevent mouse interaction like it's possible with WPF via HitTest or other mechanisms. Disabling and re-enabling controls produces a strange visual effect, personally I don't care to trade visual ugliness for comfortable functionality, but not everyone feels the same, hence I was thinking about an option.

I will make another video of my current prototype soon to show you.

TanukiSharp:
Here is the video of what it looks when disabling all controls.
Note that this is per container, so clicking on the pull button does not disable controls in the filters tool bar, and vice versa.

https://user-images.githubusercontent.com/4018313/195964871-5475cafc-479f-419e-8410-fc4e20abf992.mp4

[...]

PR #3833, opened 2017-07-11
Fixed superproject, stash and pull ToolStripSplitButton closing issue

TanukiSharp:
Fixes #3832

Changes proposed in this pull request:
 - Avoid `ToolStripSplitButton` menu to close unexpectedly
 
How did I test this code:
 - Click on `Superproject`, `Stash` of `Pull` button (the small black triangle)
 - Move the mouse 1 pixel out of the button or the left or on the right of the button
 - The context menu that just opened should close
 - The fix prevent the menu to close in such situation

Has been tested on (remove any that don't apply):
 - Windows 10 Pro

Issue #3832, opened 2017-07-10
Annoying closing of menu from tool bar buttons


TanukiSharp:
This is a **bug** report, with a fix proposal.

The current behavior is when I click the small black arrow of tool bar buttons and a contextual menu opens, the menu closes if the mouse leave the control. It happens with all `ToolStripSplitButton` controls. (Superproject, Stash and Pull ones)

![_gitextensions_pr](https://user-images.githubusercontent.com/4018313/28027169-1d6f6cd0-65d3-11e7-81c6-c12d20f15399.png)

I figured out this happens in a certain `Focus` configuration, when mouse overs the buttons and they render an emboss, this bug happens, however if the branch drop down or the filter text box is focused, the tool bar buttons do not render their emboss on mouse over, and in this focus configuration, the menu does not closes when the mouse exit its area.

I implemented a fix based on this workaround.

This issue is of course not related to Git but purely UI and happens even in `2.50.01`. So far I always encountered this issue, which really drives me nuts.

That said, I wonder if I'm the only one pissed off by this issue because I couldn't find any report of it.

I've read the contribution guidelines a bit too late and already implemented a fix, the commit can be reviewed here: https://github.com/TanukiSharp/gitextensions/commit/f4d2e5bd41660493392f227c15a817cb5b3c380f

The way I fixed it is by enforcing the focus to a control that prevent menu to close (branch drop down), only if it is not already focused (and if the filter text box is not focused either). I store the focused control in the button's `Tag` property in order to restore the focus when the `ToolStripSplitButton` closes. This may break something if the `Tag` property can be used for another feature.

This fix looks like a hack, but I couldn't find any other way around.

I implemented a solution based on constraining the mouse position (screen space) in order to ensure the mouse can only leave the control from top, left and bottom but never from the right. This works, but looks even more cheesy. We can talk about it.

Another solution would be to re-implement the `ToolStripSplitButton` based on a regular `Button` with a `ContextMenu` attach to it, but this is quite a lot of work.

RussKie:
> That said, I wonder if I'm the only one pissed off by this issue because I couldn't find any report of it.

I reproduced the behaviour you describe, but I must be more dexterous - I move my mouse down and rarely have the menu closing on me.

> Another solution would be to re-implement the ToolStripSplitButton based on a regular Button with a ContextMenu attach to it, but this is quite a lot of work.

No. We're not reimplementing default Windows controls. 
[...]

PR #10260, opened 2022-10-15
Removed changes of PR #3833, suspend/resume redrawn of ToolStripEx when dropdown menu is opened/closed

TanukiSharp:
This PR does:

- remove changes made in PR #3833, which was shitty, let's be honest
- adds a suspend/resume of redraw of the `ToolTipEx` control when a child element of type `ToolStripDropDownItem` opens/closes a dropdown menus

See related issue #10254 (recent) and #3832 (old).

Previous PR as an attempt to fix this problem is #3833. It used to work but the fix was not satisfying (non-sense and dirty hack) and stopped being effective after a few versions of GE.

Also, see #10259 since this is my first PR from after 23 January 2019.

### DIFF
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -167,7 +167,7 @@ namespace GitUI.UserControls
             tstxtRevisionFilter.ForeColor = toolForeColor;
         }
 
-        private void SelectShowBranchesFilterOption(byte selectedIndex)
+        private void SelectShowBranchesFilterOption(int selectedIndex)
         {
             if (selectedIndex >= tssbtnShowBranches.DropDownItems.Count)
             {
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -167,19 +167,7 @@ namespace GitUI.UserControls
             tstxtRevisionFilter.ForeColor = toolForeColor;
         }
 
-        public void PreventToolStripSplitButtonClosing(ToolStripSplitButton? control)
-        {
-            if (control is null
-                || Parent is not ContainerControl parentContainer)
-            {
-                return;
-            }
-
-            control.Tag = parentContainer.FindFocusedControl();
-            control.DropDownClosed += ToolStripSplitButtonDropDownClosed;
-        }
-
-        private void SelectShowBranchesFilterOption(int selectedIndex)
+        private void SelectShowBranchesFilterOption(byte selectedIndex)
         {
             if (selectedIndex >= tssbtnShowBranches.DropDownItems.Count)
             {
@@ -382,20 +370,6 @@ namespace GitUI.UserControls
             }
         }
 
-        private static void ToolStripSplitButtonDropDownClosed(object sender, EventArgs e)
-        {
-            if (sender is ToolStripSplitButton control)
-            {
-                control.DropDownClosed -= ToolStripSplitButtonDropDownClosed;
-
-                if (control.Tag is Control controlToFocus)
-                {
-                    controlToFocus.Focus();
-                    control.Tag = null;
-                }
-            }
-        }
-
         private void tsbtnAdvancedFilter_ButtonClick(object sender, EventArgs e)
         {
             if (!tsmiResetAllFilters.Enabled)
@@ -406,11 +380,6 @@ namespace GitUI.UserControls
             {
                 tsbtnAdvancedFilter.ShowDropDown();
             }
-        }
-
-        private void tsbtnAdvancedFilter_DropDownOpening(object sender, EventArgs e)
-        {
-            PreventToolStripSplitButtonClosing(sender as ToolStripSplitButton);
         }
 
         private void tstxtRevisionFilter_KeyUp(object sender, KeyEventArgs e)
--- a/GitUI/UserControls/ToolStripEx.cs
+++ b/GitUI/UserControls/ToolStripEx.cs
@@ -1,5 +1,6 @@
 ﻿using System.ComponentModel;
 using System.Reflection;
+using System.Runtime.InteropServices;
 
 namespace GitUI
 {
@@ -16,6 +17,64 @@ namespace GitUI
 
             PropertyInfo propGrip = GetType().GetProperty("Grip", BindingFlags.Instance | BindingFlags.NonPublic);
             _gripButton = propGrip.GetValue(this) as ToolStripButton;
+        }
+
+        protected override void OnItemAdded(ToolStripItemEventArgs e)
+        {
+            base.OnItemAdded(e);
+
+            if (e.Item is ToolStripDropDownItem item)
+            {
+                item.DropDownOpening += SplitButton_DropDownOpening;
+                item.DropDownClosed += SplitButton_DropDownClosed;
+            }
+        }
+
+        protected override void OnItemRemoved(ToolStripItemEventArgs e)
+        {
+            if (e.Item is ToolStripDropDownItem item)
+            {
+                item.DropDownOpening -= SplitButton_DropDownOpening;
+                item.DropDownClosed -= SplitButton_DropDownClosed;
+            }
+        }
+
+        [DllImport("user32.dll")]
+        private static extern int SendMessage(IntPtr hwnd, int wmsg, bool wparam, int lparam);
+
+        private const int WM_SETREDRAW = 11;
+
+        private static void SuspendDrawing(Control control)
+        {
+            if (control != null)
+            {
+                SendMessage(control.Handle, WM_SETREDRAW, false, 0);
+            }
+        }
+
+        private static void ResumeDrawing(Control control)
+        {
+            if (control != null)
+            {
+                SendMessage(control.Handle, WM_SETREDRAW, true, 0);
+                control.Refresh();
+            }
+        }
+
+        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
+        {
+            if (sender is ToolStripDropDownItem item)
+            {
+                SuspendDrawing(item.Owner);
+            }
+        }
+
+        private void SplitButton_DropDownClosed(object? sender, EventArgs e)
+        {
+            if (sender is ToolStripDropDownItem item)
+            {
+                ResumeDrawing(item.Owner);
+            }
         }
 
         /// <summary>
--- a/GitUI/UserControls/ToolStripEx.cs
+++ b/GitUI/UserControls/ToolStripEx.cs
@@ -38,36 +38,22 @@ namespace GitUI
             }
         }
 
-        private static void SuspendDrawing(Control control)
-        {
-            if (control != null)
-            {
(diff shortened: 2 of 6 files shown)


TanukiSharp:
I think that CodeFactor issue at [GitUI\CommandsDialogs\FormBrowse.cs:2952-2960](https://github.com/gitextensions/gitextensions/blob/d1c6120b63ef2aacacc1392900ce8a756433497d/GitUI/CommandsDialogs/FormBrowse.cs#L2952-L2960) is unrelated, but let me know if I can do something.

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -166,18 +166,6 @@ public void InitToolStripStyles(Color toolForeColor, Color toolBackColor)
...
 
-        public void PreventToolStripSplitButtonClosing(ToolStripSplitButton? control)
-        {
-            if (control is null
-                || Parent is not ContainerControl parentContainer)
-            {
-                return;
-            }
-
-            control.Tag = parentContainer.FindFocusedControl();
```
gerhardol, line 177: Should this be removed?
It changes focusing when menu has changed

gerhardol (commented):
+1
Seem to work for me.
One comment on code removed in toolbar.
The core review is for the added code in ToolStripEx.cs, more reviews needed there.

GitUI/UserControls/FilterToolBar.cs
TanukiSharp: Yes, this was part of the dirty hack, in a previous version, giving focus to a focusabled control was preventing the menus from collapsing. Don't ask me why ^^, but it used to work. That's what was really not satisfying. Now it's gone, and it's good.

GitUI/UserControls/ToolStripEx.cs
```diff
@@ -18,6 +19,64 @@ public ToolStripEx()
...
+            {
+                item.DropDownOpening -= SplitButton_DropDownOpening;
+                item.DropDownClosed -= SplitButton_DropDownClosed;
+            }
+        }
+
+        [DllImport("user32.dll")]
+        private static extern int SendMessage(IntPtr hwnd, int wmsg, bool wparam, int lparam);
+
+        private const int WM_SETREDRAW = 11;
```
RussKie, line 45: All interop code resides in https://github.com/gitextensions/gitextensions/tree/master/GitUI/Interops. We already have definitions for `SendMessageW`:
https://github.com/gitextensions/gitextensions/blob/e78a4e1fffd17989b771ed3301aca73d384e0afd/GitUI/Interops/User32/SendMessageW.cs#L8

GitUI/UserControls/ToolStripEx.cs
```diff
@@ -18,6 +19,64 @@ public ToolStripEx()
...
+        }
+
+        private static void ResumeDrawing(Control control)
+        {
+            if (control != null)
+            {
+                SendMessage(control.Handle, WM_SETREDRAW, true, 0);
+                control.Refresh();
+            }
+        }
```
RussKie, line 62: These two methods have single call site each, and should be inlined. There's no need to create more stack frames.

GitUI/UserControls/ToolStripEx.cs
```diff
@@ -18,6 +19,64 @@ public ToolStripEx()
...
+
+        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
+        {
+            if (sender is ToolStripDropDownItem item)
```
RussKie, line 66:
```suggestion
            if (sender is ToolStripDropDownItem item && item.Owner is not null)
```

ditto below

RussKie (changes requested):
Could you please add animations clearly showing the old and the new behaviours?

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: I understand you point of view, but since source code is for humans and not computers, I since it is better to have short and well named methods than saving a stack frame.

Let me know if this is a blocker and I will change as you requested, but I'd like to make sure before.

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: Let's assume we keep `SuspendDrawing()` and `ResumeDrawing()` not inlined, and since the first thing they do is to check for `null`, would this change still be necessary ?

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: I added `WM_SETREDRAW` with XML document taken from MSDN and slightly modified.

I also added `NativeMethods.TRUE` and `NativeMethods.FALSE`, I wonder why they are not already present, let me know if you see a better way.

TanukiSharp:
@RussKie 

> Could you please add animations clearly showing the old and the new behaviours?

Here is a video with the current fix.

https://user-images.githubusercontent.com/4018313/196388956-c0592867-5b28-40e8-9554-27970f8a9f97.mp4

To compare with previous fixes and previous state, have a look at videos posted in #10254

Sorry for the huge black strips, not an expert in video edition :/

GitUI/UserControls/ToolStripEx.cs
RussKie: > but since source code is for humans and not for computers,

There are many different schools of thought and opinions on the subject, but small methods which are only called from one place impose maintenance and mental tax too. (Not to mention potential have perf implications).

For the sake of discussion, suppose, I follow your argument, let's compare the proposed code:
```cs
        private static void SuspendDrawing(Control control)
        {
            if (control != null)
            {
                NativeMethods.SendMessageW(control.Handle, NativeMethods.WM_SETREDRAW, NativeMethods.FALSE, IntPtr.Zero);
            }
        }

        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
        {
            if (sender is ToolStripDropDownItem item)
            {
                SuspendDrawing(item.Owner);
            }
        }
```
...and an alternative:
```cs
        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
        {
            if (sender is ToolStripDropDownItem item && item.Owner is Control control)
            {
                NativeMethods.SendMessageW(control.Handle, NativeMethods.WM_SETREDRAW, NativeMethods.FALSE, IntPtr.Zero);
            }
        }
```

Could you please articulate why the first snippet is better? It's more verbose, it has multiple `if` conditions and, thus, more complex.

gerhardol (approved):
Fine for me with RussKie changesto be decided on (I would like to see some comments at least and wish there were an `inline` in C#, not just advisory, even if I expect that the compiler should be able to inline automatically here).

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: It is more verbose if you take it in its whole, but let's just take the entry points:

```cs
        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
        {
            if (sender is ToolStripDropDownItem item)
            {
                SuspendDrawing(item.Owner);
            }
        }
```

It reads like a book, reading this I know that entering the `if`, it will suspend redraws, and most of the time that's all I need to know. If I choose or need to know how redraws are suspended, I can decide to visit the `SuspendDrawing()` methods. I can also decide to not spend time for that and move on.

With the follwoing:

```cs
        private void SplitButton_DropDownOpening(object? sender, EventArgs e)
        {
            if (sender is ToolStripDropDownItem item && item.Owner is Control control)
            {
                NativeMethods.SendMessageW(control.Handle, NativeMethods.WM_SETREDRAW, NativeMethods.FALSE, IntPtr.Zero);
            }
        }
```

First, the `if` condition is longer and more complex, so it's harder, when overlooking the code, to understand how and when the `if` will enter. Second, the statement in the `if` barely tells what it does by itself, and one will probably have to search MSDN or else to understand what's going on. You could add a comment, but then that will make the code longer and therefore more verbose. Also, comment may go desync the day one decide to send another type of message for some reason but forget to update the comment. You may argue that it's also possible to change the message and forgetting to rename the method, but it's less likely to happen when the function does just one single thing. I believe code beats comments, so I prefer to have a short and well named methods that's self-describing, than obscure code commented somewhere. One more thing, the day someone else needs that feature too, the code is already factorized (partly) and ready to use. With the second snippet, it's easy to move the content of the `if` and forget about also taking the `if` for null check. With the first snippet, moving the method to a dedicated class (could be moved to `NativeMethods`) is all that's to be done as it is already self-sufficient and self-describing.

As for performance, it is also a very important topic for me and I agree with you, but here we are not in a hot path at all, so trying to save a stack frame is not really something that should drive this decision, IMHO.

GitUI/UserControls/ToolStripEx.cs
```diff
@@ -18,6 +18,59 @@ public ToolStripEx()
...
+            if (e.Item is ToolStripDropDownItem item)
+            {
+                item.DropDownOpening -= SplitButton_DropDownOpening;
+                item.DropDownClosed -= SplitButton_DropDownClosed;
+            }
+        }
+
+        private static void SuspendDrawing(Control control)
+        {
+            if (control != null)
```
mstv, line 43:
```suggestion
            if (control is not null)
```
is our coding style.

(If there was more code, rather:
```suggestion
            if (control is null)
            {
                return;
            }
```
)

GitUI/UserControls/ToolStripEx.cs
```diff
@@ -18,6 +18,59 @@ public ToolStripEx()
...
+            {
+                NativeMethods.SendMessageW(control.Handle, NativeMethods.WM_SETREDRAW, NativeMethods.FALSE, IntPtr.Zero);
+            }
+        }
+
+        private static void ResumeDrawing(Control control)
+        {
+            if (control != null)
```
mstv, line 51: ditto

GitUI/UserControls/ToolStripEx.cs
mstv: > Would this change still be necessary?

only if the functions are merged (I tend to keep them separate but YAGNI.)

mstv (commented):
have not run yet

GitUI/UserControls/ToolStripEx.cs
RussKie: Thank you for the explanation.

TanukiSharp:
Based on @RussKie disagreement, I inlined the calls and add the proper `item.Owner` check. Also added comments. I think this should address all the remaining issues, but let me know if there's something else.

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: I inlined the methods.

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: I added this check since methods have been inlined.

GitUI/UserControls/ToolStripEx.cs
TanukiSharp: Methods are gone (inlined) so this code is gone too.

gerhardol (approved):
Include in 4.0

gerhardol:
Merging tomorrow

RussKie:
Thank you

PR #10261, opened 2022-10-15
Git Executable: return stderr also if not throwOnErrorExit is suppressed

gerhardol:
Fixes #10242

## Proposed changes

if throwOnErrorExit  was not set to false for Git commands, stderr was always empty.
This is likely a remains of the first implementation of throwOnErrorExit that raised errors for contents in stderr in addition to non zero error code.
Also git-verify-tag should only parse stderr, not stdout.

## Test methodology 

Manual

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

Issue #10242, opened 2022-10-09
GPG info Regression


vbjay:
### Environment

- Git Extensions 33.33.33
- Build 077046ca17bf43a6fe0fd958b53d2493a780db52
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.9
- DPI 96dpi (no scaling)

### Issue description

Looks like a regression between v3.5.4 and master.  Somehow it is failing to get any output from 
https://github.com/gitextensions/gitextensions/blob/3d4ed902fcd13dba41d66dd1e6b918ba78eac3f5/GitCommands/Git/GitGpgController.cs#L231-L246

---
- Git Extensions 3.5.4.12724
- Build 65f01f39982ccb121c4df7d4b00b506939553958
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET Framework 4.8.9093.0
- DPI 96dpi (no scaling)
![image](https://user-images.githubusercontent.com/3629489/194782238-0df11204-cbb0-44b0-bb9d-a926789dab49.png)

---
- Git Extensions 33.33.33
- Build 077046ca17bf43a6fe0fd958b53d2493a780db52
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.9
- DPI 96dpi (no scaling)

AND

- Git Extensions 33.33.33
- Build 8b80e54f15e8923a85f5801cca033f932ced821f
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.9
- DPI 96dpi (no scaling)

![image](https://user-images.githubusercontent.com/3629489/194782273-7f84e529-31c1-470e-aceb-a3d17a2d77f5.png)

### Steps to reproduce

View a commit with a signed tag and view gpg tab.

### Did this work in previous version of GitExtensions?

- Git Extensions 3.5.4.12724
- Build 65f01f39982ccb121c4df7d4b00b506939553958
- Git 2.37.3.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET Framework 4.8.9093.0
- DPI 96dpi (no scaling)

### Diagnostics

![image](https://user-images.githubusercontent.com/3629489/194782817-e95be451-b704-469c-9c56-ae331d701434.png)

BinHavin0323:
- [ ] []()

vbjay:
[...]

PR #10262, opened 2022-10-15
Cleanup handling of AllOutput for processes

gerhardol:
Initiated from #10242 
Some parts were initially in #10261 , separated to this PR

## Proposed changes

AllOutput is combined use of stderr and stdout and should only be used to display messages for users.

* Change a few commands to explicitly read stderr where required. Some tests had to be adapted (including MockExecutable).

* Clarify usage that AllOutput is used for user popups. Most of these occurrences are related to FormCommit. The structure of these command was changed to return a bool for result and the output as a string, also if result/output were unused.

* Set TODO on a few occurrences that were not investigated.

## Test methodology 

Tests updated

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitGpgController.cs
+++ b/GitCommands/Git/GitGpgController.cs
@@ -232,13 +232,13 @@ namespace GitCommands.Gpg
                 return null;
             }
 
-            var module = GetModule();
+            // '--raw' returns info in stderr
             GitArgumentBuilder args = new("verify-tag")
             {
                 { raw, "--raw" },
                 tagName
             };
-            return module.GitExecutable.GetOutput(args);
+            return GetModule().GitExecutable.Execute(args).StandardError;
         }
 
         private string? EvaluateTagVerifyMessage(IReadOnlyList<IGitRef> usefulTagRefs)
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -1642,14 +1642,21 @@ namespace GitCommands
 
         #region Stage/Unstage
 
-        public string AssumeUnchangedFiles(IReadOnlyList<GitItemStatus> files, bool assumeUnchanged, out bool wereErrors)
+        /// <summary>
+        /// Set/unset assumeUnchanged
+        /// </summary>
+        /// <param name="files">Files to change</param>
+        /// <param name="assumeUnchanged">Set/unset assumeUnchanged</param>
+        /// <param name="alloutput">stdout and stderr output</param>
+        /// <returns>true if no errors occurred</returns>
+        public bool AssumeUnchangedFiles(IReadOnlyList<GitItemStatus> files, bool assumeUnchanged, out string alloutput)
         {
             files = files.Where(file => file.IsAssumeUnchanged != assumeUnchanged).ToList();
 
             if (files.Count == 0)
             {
-                wereErrors = false;
-                return "";
+                alloutput = "";
+                return true;
             }
 
             ExecutionResult execution = _gitExecutable.Execute(
@@ -1668,17 +1675,25 @@ namespace GitCommands
                 SystemEncoding,
                 throwOnErrorExit: false);
 
-            wereErrors = !execution.ExitedSuccessfully;
-            return execution.AllOutput;
+            alloutput = execution.AllOutput;
+            return execution.ExitedSuccessfully;
         }
 
-        public string SkipWorktreeFiles(IReadOnlyList<GitItemStatus> files, bool skipWorktree)
+        /// <summary>
+        /// Set/unset skipworktree
+        /// </summary>
+        /// <param name="files">Files to change</param>
+        /// <param name="skipWorktree">Set/unset skipworktree</param>
+        /// <param name="alloutput">stdout and stderr output</param>
+        /// <returns>true if no errors occurred</returns>
+        public bool SkipWorktreeFiles(IReadOnlyList<GitItemStatus> files, bool skipWorktree, out string alloutput)
         {
             files = files.Where(file => file.IsSkipWorktree != skipWorktree).ToList();
 
             if (files.Count == 0)
             {
-                return "";
+                alloutput = "";
+                return true;
             }
 
             var execution = _gitExecutable.Execute(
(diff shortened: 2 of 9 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -1642,14 +1642,21 @@ public string ApplyPatch(string dir, ArgumentString amCommand)
...
         #region Stage/Unstage
 
-        public string AssumeUnchangedFiles(IReadOnlyList<GitItemStatus> files, bool assumeUnchanged, out bool wereErrors)
+        /// <summary>
+        /// Set/unset assumeUnchanged
+        /// </summary>
+        /// <param name="files">Files to change</param>
+        /// <param name="assumeUnchanged">Set/unset assumeUnchanged</param>
+        /// <param name="alloutput">stdout and stderr output</param>
+        /// <returns>true if no errors occurred</returns>
```
RussKie, line 1651: The docs should be in sentence case and end with a full stop. We should also strive to use xml-doc tags, where appropriate.
```suggestion
        /// <summary>
        /// Set/unset whether given items are assumed unchanged.
        /// </summary>
        /// <param name="files">The files to set the status for.</param>
        /// <param name="assumeUnchanged">The status value.</param>
        /// <param name="allOutput">stdout and stderr output.</param>
        /// <returns><see langword="true"/> if no errors occurred; otherwise, <see langword="false"/>.</returns>
```

Ditto below.

RussKie (approved):
LGTM, though I haven't run this.

RussKie:
I presume we'd want this in v4, right?

GitCommands/Git/GitModule.cs
```diff
@@ -1778,13 +1811,13 @@ public string UnstageFiles(IReadOnlyList<GitItemStatus> files)
 
             if (nonNewFiles.Count != 0)
             {
-                StringBuilder sb = new("reset --");
+                GitArgumentBuilder sb = new("reset") { "--" };
                 foreach (var file in nonNewFiles)
                 {
-                    sb.Append($" \"{file.Name.ToPosixPath()}\"");
+                    sb.Add($"\"{file.Name.ToPosixPath()}\"");
                 }
```
vbjay, line 1819: Quote or QuoteNe extension method?

vbjay (commented):
Not blocking comments but things I noticed.

GitCommands/Git/GitModule.cs
vbjay: Are we sure removing the space still works as expected?

vbjay: Nevermind. I get it. The add adds the space.

GitCommands/Git/GitModule.cs
gerhardol: Updated.
(Not too happy with the xml-doc, the primary intention is to mark that AllOutput is deliberate, not being parsed.)

GitCommands/Git/GitGpgController.cs
```diff
@@ -232,13 +232,13 @@ public string GetCommitVerificationMessage(GitRevision revision)
                 return null;
             }
 
-            var module = GetModule();
+            // '--raw' returns info in stderr
```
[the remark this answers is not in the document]
gerhardol: 'git verify-tag --raw' returns the info in stderr, but the command itself is redundant here.
--raw returning in stderr is just an unexpected behavior that I want to document, commands should normally only use stderr for displaying errors.

I moved the comment to the return that accesses stderr instead.

GitCommands/Git/GitModule.cs
gerhardol: changed to QuoteNE

GitCommands/Git/GitGpgController.cs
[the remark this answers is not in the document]
vbjay: https://github.com/gitextensions/gitextensions/pull/10261#discussion_r996349904

https://github.com/gitextensions/gitextensions/pull/10261#discussion_r996346945

My point is it doesn't matter if it is --raw or not.

gerhardol:
> I presume we'd want this in v4, right?

This makes no functional change, it was a cleanup to ensure that there are no "unexpected" use of stderr as for git-verify-tag.
There are still possible use of stderr in GetOutput, but as `throwOnErrorExit` is not exposed, the usages throws an exception if Git fails.

But I find it safe to include.
We need to stop adding to 4.0 soon and release it...
But yet another beta.

GitCommands/Git/GitGpgController.cs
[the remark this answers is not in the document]
vbjay: https://github.com/git/git/blob/d420dda0576340909c3faff364cfbd1485f70376/gpg-interface.c#L656-L666

See how they output to both streams?  It is not an error if error stream has output. That is what the error code provides.  I have seen a few programs do this where they output a your summary or details 8n standard output and verbose logging in error or a different result set in error. As seen above they have the potential to output a different verify output if the flag is set in standard output. Tte error is the output we expect. I certainly see the issue of they should have done it the other way around.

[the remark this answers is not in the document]
gerhardol: The comment is intended to describe why stderr is parsed instead of stdout.
In general commands should only use stderr to display error messages to the user. I tried to document why stderr was explicitly used with this commit.

Can you make a suggestion on how to describe this behavior?

GitCommands/Git/GitGpgController.cs
```diff
@@ -232,13 +232,14 @@ public string GetCommitVerificationMessage(GitRevision revision)
...
 
-            var module = GetModule();
             GitArgumentBuilder args = new("verify-tag")
             {
                 { raw, "--raw" },
                 tagName
             };
-            return module.GitExecutable.GetOutput(args);
+
+            // '--raw' returns info in stderr
```
vbjay, line 241:
```suggestion
            // 'verify-tag' returns info in stderr
```
https://github.com/gitextensions/gitextensions/pull/10262#discussion_r996409350

vbjay (changes requested):
https://github.com/gitextensions/gitextensions/pull/10262#discussion_r996409350

GitCommands/Git/GitGpgController.cs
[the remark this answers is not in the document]
vbjay: https://github.com/git/git/blob/d420dda0576340909c3faff364cfbd1485f70376/gpg-interface.c#L661
```c++
        if (flags & GPG_VERIFY_VERBOSE && sigc->payload)
```

Typically that is the case but it is not always the case. Looks like it outputs the raw payload to stdoutput if ```GPG_VERIFY_VERBOSE``` is set.

vbjay (approved):
LGTM

vbjay:
I'll run and verify changes later on today for   #10242

vbjay:
https://github.com/gerhardol/gitextensions/pull/5

vbjay:
So testing of this pull and the other together are confirmed that #10242 is resolved if both #10261 and #10262  are merged.  See https://github.com/gerhardol/gitextensions/pull/5 for how I tested.

UnitTests/CommonTestUtils/MockExecutable.cs
```diff
@@ -98,10 +98,10 @@ public IProcess Start(ArgumentString arguments, bool createWindow, bool redirect
 
         private sealed class MockProcess : IProcess
         {
-            public MockProcess([CanBeNull] string output, int? exitCode = 0)
+            public MockProcess([CanBeNull] string output, int? exitCode = 0, [CanBeNull] string error = null)
```
mstv, line 101: ❔
```suggestion
            public MockProcess(string? output, int? exitCode = 0, string? error = null)
```

mstv (approved):
LGTM

UnitTests/CommonTestUtils/MockExecutable.cs
gerhardol: Removed the JetBrains annotation. There are a few other uses remaining, mostly in test files

PR #10282, opened 2022-10-20
Add support for bearer-token auth type in build server adapters

mdonatas:
## Background

Some TeamCity instance don't have username/password auth enabled and only support oAuth. In such cases it's not possible to configure TeamCity build integration as there is no username and password to provide.
image

## Proposed changes

- Change `BuildServerCredentials.UseGuestAccess` from `bool` to a more flexible `BuildServerCredentials.BuildServerCredentialsType` `enum`
- Add `BearerToken` auth option
- Keep backwards compatibility with `UseGuestAccess` setting: try to read its value for initial value of `BuildServerCredentialsType`
- Handle `BearerToken` credentials type in TeamCity build server adapter

### Before

userpass-auth

### After

bearer-auth

## Test methodology 

- Manual testing with TeamCity requiring Bearer token

## Merge strategy

----

### DIFF
--- a/GitUI/BuildServerIntegration/BuildServerWatcher.cs
+++ b/GitUI/BuildServerIntegration/BuildServerWatcher.cs
@@ -131,13 +131,15 @@ namespace GitUI.BuildServerIntegration
         {
             lock (_buildServerCredentialsLock)
             {
-                IBuildServerCredentials? buildServerCredentials = new BuildServerCredentials { UseGuestAccess = true };
-                var foundInConfig = false;
+                IBuildServerCredentials? buildServerCredentials = new BuildServerCredentials { BuildServerCredentialsType = BuildServerCredentialsType.Guest };
+                bool foundInConfig = false;
 
                 const string CredentialsConfigName = "Credentials";
                 const string UseGuestAccessKey = "UseGuestAccess";
+                const string BuildServerCredentialsTypeKey = "BuildServerCredentialsType";
                 const string UsernameKey = "Username";
                 const string PasswordKey = "Password";
+                const string BearerTokenKey = "BearerToken";
                 using (var stream = GetBuildServerOptionsIsolatedStorageStream(buildServerAdapter, FileAccess.Read, FileShare.Read))
                 {
                     if (stream.Position < stream.Length)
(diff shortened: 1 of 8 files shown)


GitUI/BuildServerIntegration/BuildServerWatcher.cs
```diff
@@ -161,10 +163,22 @@ public void CancelBuildStatusFetchOperation()
 
                             if (section is not null)
                             {
-                                buildServerCredentials.UseGuestAccess = section.GetValueAsBool(UseGuestAccessKey,
-                                    true);
+                                string? buildServerCredentialsType = section.GetValue(BuildServerCredentialsTypeKey);
+                                if (!string.IsNullOrWhiteSpace(buildServerCredentialsType))
+                                {
+                                    buildServerCredentials.BuildServerCredentialsType = Enum.Parse<BuildServerCredentialsType>(buildServerCredentialsType);
```
RussKie, line 169: What would happen if `buildServerCredentialsType = 143`? That is, a value is not defined.

Plugins/BuildServerIntegration/TeamCityIntegration/TeamCityAdapter.cs
```diff
@@ -412,18 +412,28 @@ private Task<Stream> GetStreamFromHttpResponseAsync(Task<HttpResponseMessage> ta
...
+                                                && buildServerCredentials.BuildServerCredentialsType == BuildServerCredentialsType.UsernameAndPassword
+                                                && !string.IsNullOrWhiteSpace(buildServerCredentials.Username) && !string.IsNullOrWhiteSpace(buildServerCredentials.Password);
                 if (useBuildServerCredentials)
                 {
                     UpdateHttpClientOptionsCredentialsAuth(buildServerCredentials!);
-                    return GetStreamAsync(restServicePath, cancellationToken);
                 }
-                else
+
+                var useBuildServerBearerToken = buildServerCredentials is not null
```
RussKie, line 422: We shouldn't use `var` for the intrinsic types and when the type can't be easily inferred from the context.

RussKie (commented):
LGTM as reviewed in a browser

GitUI/BuildServerIntegration/BuildServerWatcher.cs
mdonatas: Good point - it would crash :)
image

I'll change it to `TryParse` instead
image

RussKie (approved):
LGTM, but I don't use TC.

PR #10291, opened 2022-10-24
perf: ensure compiled regex are compiled only once

pmiossec:
## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 24e8371e13366c9de155109d4a319ef0c044f9a1 (Dirty)
- .NET 6.0.10
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -48,6 +48,17 @@ namespace GitCommands
         private readonly IExecutable _gitExecutable;
         private readonly IExecutable _gitWindowsExecutable;
 
+        // Parse lines of format:
+        //
+        // 69a7c7a40230346778e7eebed809773a6bc45268 refs/heads/master
+        // 69a7c7a40230346778e7eebed809773a6bc45268 refs/remotes/origin/master
+        // 366dfba1abf6cb98d2934455713f3d190df2ba34 refs/tags/2.51
+        //
+        // Lines may also use \t as a column delimiter, such as output of "ls-remote --heads origin".
+        private static readonly Regex _refRegex = new(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline | RegexOptions.Compiled);
+
+        private static readonly Regex _headerRegex = new(@"^(?<objectid>[0-9a-f]{40}) (?<origlinenum>\d+) (?<finallinenum>\d+)", RegexOptions.Compiled);
+
         /// <summary>
         /// Name of the WSL distro for the GitExecutable, empty string for the app native Windows Git executable.
         /// This can be seen as the Git "instance" identifier.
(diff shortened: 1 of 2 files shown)


gerhardol (commented):
Static methods could be faster, maybe use compiled regex too.
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices#handle-object-instantiation-appropriately

RussKie (approved):

RussKie:
.NET 7 introduces source generation for regex via attribute decorations.

gerhardol:
> .NET 7 introduces source generation for regex via attribute decorations. 

I have considered reviewing all regex use for vNext. Most should use static methods I believe. Maybe wait for  net7 before looking in to that then.

pmiossec:
> .NET 7 introduces source generation for regex via attribute decorations.

Yes looking forward to it but until that, a Compiled RegEx **must** be static otherwise it could degrade the performance instead of improving it...

GitCommands/Git/GitModule.cs
```diff
@@ -3092,17 +3101,7 @@ public IReadOnlyList<string> GetMergedRemoteBranches()
...
         {
-            // Parse lines of format:
-            //
-            // 69a7c7a40230346778e7eebed809773a6bc45268 refs/heads/master
-            // 69a7c7a40230346778e7eebed809773a6bc45268 refs/remotes/origin/master
-            // 366dfba1abf6cb98d2934455713f3d190df2ba34 refs/tags/2.51
-            //
-            // Lines may also use \t as a column delimiter, such as output of "ls-remote --heads origin".
-
-            Regex regex = new(@"^(?<objectid>[0-9a-f]{40})[ \t](?<refname>.+)$", RegexOptions.Multiline | RegexOptions.Compiled);
```
gerhardol, line 3103: Lines that start with sha should use TryParse() and trim whitespace for the rest, likely faster.
Maybe a separate PR.

gerhardol (approved):
There are some more Compiled Regex that are not static, but most can be ignored.
WordDictionary.cs:377 and maybe FileStatusList

Many of these should drop Compiled and use static methods (like Regex.Match(string, regex)
If you are executing in a loop, this is very efficient. I played around with this in the side panel stash PR, even if I did not need to use it. First call was about 1 ms for a quite simple Regex, then less than microseconds.

GitCommands/Git/GitModule.cs
```diff
@@ -3337,6 +3336,8 @@ public GitBlame Blame(string? fileName, string from, Encoding encoding, string?
             }
         }
 
+        private static readonly Regex _headerRegex = new(@"^(?<objectid>[0-9a-f]{40}) (?<origlinenum>\d+) (?<finallinenum>\d+)", RegexOptions.Compiled);
```
RussKie, line 3339: Please move this to the top of the file

mstv: @pmiossec, could you move it before leaving yet?

pmiossec:
done.

mstv (approved):

PR #10293, opened 2022-10-25
FormBrowse: let the user add fetch/pull custom icons in the toolbar

pmiossec:
Improvement over #3923

## Proposed changes

- The user is able to add a direct access button for the fetch/pull actions he wants (because split button is painful to use on high resolution screens)
- By default, the button are not displayed, so nothing changed

### After

Configuration:

image

Result (because I want a fetch button distinct of a pull button --like in all the other git GUIs--):

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build b5ec77dbdf12b00702b5212419a369e6a536bc9b
- Git 2.38.0.windows.1 (recommended: 2.38.1 or later)
- Microsoft Windows NT 10.0.19042.0
- .NET 6.0.10
- DPI 96dpi (no scaling)

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10294, opened 2022-10-25
FormBrowse: Hide Reflog button by default

pmiossec:
Alternative to #10235

### Before

image

### After

By default:

image

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 80f823749a6fda767bd547af21b8f460d4b391e8
- .NET 6.0.10
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
@@ -1,6 +1,7 @@
 ﻿using System.Diagnostics;
 using GitCommands;
 using GitUI.CommandsDialogs.BrowseDialog;
+using GitUI.UserControls;
 using Microsoft;
 using ResourceManager;
 
@@ -153,7 +154,8 @@ namespace GitUI.CommandsDialogs
                     key = toolbarItem.Name;
                 }
 
-                bool visible = LoadVisibilitySetting(key);
+                bool visible = LoadVisibilitySetting(key, IsVisibleByDefault(key));
+
                 toolbarItem.Visible = visible;
 
                 ToolStripMenuItem menuToolbarItem = new(toolbarItem.ToolTipText)
@@ -170,11 +172,11 @@ namespace GitUI.CommandsDialogs
 
                     if (!BelongToAGroup(toolbarItem, out var group))
                     {
-                        SaveVisibilitySetting(toolbarItem.Name, toolbarItem.Visible);
+                        SaveVisibilitySetting(toolbarItem.Name, toolbarItem.Visible, IsVisibleByDefault(toolbarItem.Name));
                     }
                     else
                     {
-                        SaveVisibilitySetting(group, toolbarItem.Visible);
+                        SaveVisibilitySetting(group, toolbarItem.Visible, IsVisibleByDefault(toolbarItem.Name));
                         foreach (ToolStripItem item in senderToolStrip.Items)
                         {
                             if (item.Tag == (object)group)
@@ -195,8 +197,11 @@ namespace GitUI.CommandsDialogs
 
             return;
 
-            static void SaveVisibilitySetting(string key, bool visible) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible ? null : false);
-            static bool LoadVisibilitySetting(string key) => AppSettings.GetBool(toolbarSettingsPrefix + key, true);
+            bool IsVisibleByDefault(string buttonKey) => !buttonKey.Contains(FilterToolBar.ReflogButtonName);
+            static void SaveVisibilitySetting(string key, bool visible, bool defaultValue = true)
+                => AppSettings.SetBool(toolbarSettingsPrefix + key, visible == defaultValue ? null : visible);
+            static bool LoadVisibilitySetting(string key, bool defaultValue = true)
+                => AppSettings.GetBool(toolbarSettingsPrefix + key, defaultValue);
 
             static bool BelongToAGroup(ToolStripItem toolbarItem, out string groupName)
             {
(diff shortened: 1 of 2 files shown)


mstv (approved):
LGTM. It will not affect existing settings files though.

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -195,8 +198,8 @@ private static void CreateToolStripSubMenus(ToolStrip senderToolStrip, ToolStrip
 
             return;
 
-            static void SaveVisibilitySetting(string key, bool visible) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible ? null : false);
-            static bool LoadVisibilitySetting(string key) => AppSettings.GetBool(toolbarSettingsPrefix + key, true);
+            static void SaveVisibilitySetting(string key, bool visible) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible);
```
gerhardol, line 201: Why change this?

mstv: valid question (I didn't notice this change)

pmiossec: There was a choice that have been made, maybe to not "pollute" too much the setting file, to not store the setting value when the setting was 'true' (visible) but if the default value is 'false', this logic doesn't work anymore so you have to store the real value set by the user if it has been changed...

I had the same "problem" in another PR #10293

pmiossec: The other possibility if we want to keep the current behavior is to pass the "default" value and compare with this default.
Do you prefer that?

gerhardol: No strong preference, but I rather keep default as null, then it is easier to change without migrations.
No major difference here I guess. One or two lines of code or save defaults for all elements.

pmiossec: Done. `null` is now for Default value.

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -195,8 +198,8 @@ private static void CreateToolStripSubMenus(ToolStrip senderToolStrip, ToolStrip
 
             return;
 
-            static void SaveVisibilitySetting(string key, bool visible) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible ? null : false);
-            static bool LoadVisibilitySetting(string key) => AppSettings.GetBool(toolbarSettingsPrefix + key, true);
+            static void SaveVisibilitySetting(string key, bool visible, bool defaultValue = true) => AppSettings.SetBool(toolbarSettingsPrefix + key, visible == defaultValue ? null : visible);
+            static bool LoadVisibilitySetting(string key, bool defaultValue = true) => AppSettings.GetBool(toolbarSettingsPrefix + key, defaultValue);
```
mstv, line 202:
```suggestion
            static void SaveVisibilitySetting(string key, bool visible, bool defaultValue = true)
                => AppSettings.SetBool(toolbarSettingsPrefix + key, visible == defaultValue ? null : visible);

            static bool LoadVisibilitySetting(string key, bool defaultValue = true)
                => AppSettings.GetBool(toolbarSettingsPrefix + key, defaultValue);
```

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
```diff
@@ -153,7 +153,10 @@ private static void CreateToolStripSubMenus(ToolStrip senderToolStrip, ToolStrip
                     key = toolbarItem.Name;
                 }
 
-                bool visible = LoadVisibilitySetting(key);
+                bool visible = LoadVisibilitySetting(key, IsVisibleByDefault(key));
+
+                bool IsVisibleByDefault(string buttonKey) => !buttonKey.Contains("tsbShowReflog");
```
mstv, line 158: Although it is short and used here, this nested function should be moved behind the `return`.

mstv (commented):
👍
Second look -> final nitpicks

GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
RussKie, line 158:
```suggestion
                bool IsVisibleByDefault(string buttonKey) => !buttonKey.Contains(nameof(tsbShowReflog));
```

pmiossec: I would have liked (and tried) but it's not possible because that's a private field of another class 😕

pmiossec: I will expose the string as an `internal const`. I think it is better than hard-coded...

pmiossec:
Done for all.

pmiossec:
As it is an alternative to #10235, which is in milestone `v4.0`, I put it in the milestone also for easier tracking...

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -8,6 +8,7 @@ namespace GitUI.UserControls
 {
     internal partial class FilterToolBar : ToolStripEx
     {
+        internal const string ReflogButtonName = nameof(tsbShowReflog);
```
gerhardol, line 11: should this be a readonly instead?
(internal is not exposed outside the assembly so I believe this is fine, public const is not)

mstv (approved):

GitUI/UserControls/FilterToolBar.cs
RussKie: Internal consts are ok.

RussKie (approved):

PR #10295, opened 2022-10-25
range-diff with path filter

gerhardol:
## Proposed changes

Use new feature in Git 2.38

A use case for the feature: Check differences from release/4.0 to master

range-diff is a great way to compare branches, but is too slow when there are many changes.
Displaying the diff requires a long time (I gave up after 6 minutes)
image

There are still few changes in source files.
image

To find where the changes occurs is very hard though.
Also the range-diff is likely to be huge here, just the different commit messages will be hard to filter manually.

Git 2.38 adds the possibility to filter by path.
(It is not possible to filter several paths in the GUI even if paths can be manually entered in the form - a separate feature in RevDiff).

If filtering for  FormManageWorktree.cs, range-diff displays within a second.
FormManageWorktree.cs has appeared in 5 "pairs" of commits, 1:/-: just applied in a different order, 3: is identical except for the commit message, these can relative easily be visually filtered.
4:, 5: are missing in 4.0

image

## Screenshots 

See above.

## Test methodology 

Manual

## Test environment(s) 

- GIT 2.38, 2.25 in wsl

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2427,6 +2427,7 @@ namespace GitCommands
             ObjectId? firstBase,
             ObjectId? secondBase,
             string extraDiffArguments,
+            string? pathFilter,
             CancellationToken cancellationToken)
         {
             // range-diff is not possible for artificial commits, use HEAD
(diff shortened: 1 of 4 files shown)


RussKie (approved):

mstv (approved):
👍

mstv:
Should we take this for v4.0.2?

PR #10297, opened 2022-10-26
Have alternate merge strategies help button point to online manual

### LINKED ISSUE #9259: Help for alternative merge-strategy is closed to fast
## Current behaviour

The very big tool-tip for the alternative merge-strategy is closed after 4 or 5 seconds.

## Expected behaviour
Don't automatic close the tool tip
or increase the auto-close time
or find a better presentation for the many information (like a extra window).

## Steps to reproduce
1. Merge to braches
2. Activate `Alternative options`
2. Activate `Use alternative Merge strategy`
3. Hover over the `Help` link

## Screenshots

bla

No

## Environment

- Git Extensions 3.5.0.11713
- Git 2.31.1.windows.1
- .NET Framework 4.8.4341.0
- DPI 96dpi (no scaling)

lhiginbotham:
Move merge strategies help tooltip to online documentation, open link on click

Fixes #9259

See proposed documentation changes here (and complete the doc PR prior to this PR):
https://github.com/gitextensions/GitExtensionsDoc/pull/144

## Proposed changes

- the help button next to the alternate merge strategies selector in the branch merge advanced options will now open the online documentation when clicked (rather than showing a length tooltip that fades in a relatively short amount of time)

### Before

image

### After

There is no longer a tooltip and clicking the Help link brings us to https://git-extensions-documentation.readthedocs.io/en/main/branches.html#advanced-merge-options

image

## Test methodology 

- tested the link to see if it goes to an appropriate URL, verified it will bring us to the correct URL once the doc page update goes in

## Test environment(s) 

- Windows 10

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormMergeBranch.cs
+++ b/GitUI/CommandsDialogs/FormMergeBranch.cs
@@ -13,7 +13,6 @@ namespace GitUI.CommandsDialogs
     /// <summary>Form to merge a branch into the current branch.</summary>
     public partial class FormMergeBranch : GitModuleForm
     {
-        private readonly TranslationString _strategyTooltipText = new("resolve " + Environment.NewLine + "This can only resolve two heads (i.e. the current branch and another branch you pulled from) using a 3-way merge algorithm." + Environment.NewLine + "It tries to carefully detect criss-cross merge ambiguities and is considered generally safe and fast. " + Environment.NewLine + Environment.NewLine + "recursive " + Environment.NewLine + "This can only resolve two heads using a 3-way merge algorithm. When there is more than one common ancestor that can be " + Environment.NewLine + "used for 3-way merge, it creates a merged tree of the common ancestors and uses that as the reference tree for the 3-way" + Environment.NewLine + "merge. Additionally this can detect and handle merges involving renames. This is the default merge strategy when pulling or " + Environment.NewLine + "merging one branch. " + Environment.NewLine + Environment.NewLine + "octopus " + Environment.NewLine + "This resolves cases with more than two heads, but refuses to do a complex merge that needs manual resolution. It is " + Environment.NewLine + "primarily meant to be used for bundling topic branch heads together. This is the default merge strategy when pulling or " + Environment.NewLine + "merging more than one branch. " + Environment.NewLine + Environment.NewLine + "ours " + Environment.NewLine + "This resolves any number of heads, but the resulting tree of the merge is always that of the current branch head, effectively " + Environment.NewLine + "ignoring all changes from all other branches. It is meant to be used to supersede old development history of side branches." + Environment.NewLine + Environment.NewLine + "subtree " + Environment.NewLine + "This is a modified recursive strategy. When merging trees A and B, if B corresponds to a subtree of A, B is first adjusted to " + Environment.NewLine + "match the tree structure of A, instead of reading the trees at the same level. This adjustment is also done to the common " + Environment.NewLine + "ancestor tree.");
         private readonly TranslationString _formMergeBranchHoverShowImageLabelText = new("Hover to see scenario when fast forward is possible.");
         private readonly string? _defaultBranch;
         private ICommitMessageManager _commitMessageManager;
@@ -144,7 +143,8 @@ namespace GitUI.CommandsDialogs
 
         private void strategyHelp_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
         {
-            strategyToolTip.SetToolTip(strategyHelp, _strategyTooltipText.Text);
+            OsShellUtil.OpenUrlInDefaultBrowser(
+                UserManual.UserManual.UrlFor("branches", "advanced-merge-options"));
         }
 
         private void advanced_CheckedChanged(object sender, EventArgs e)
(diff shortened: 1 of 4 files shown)


lhiginbotham:
FYI @RussKie @TobiasSekan - hopefully this is along the lines of what you were looking for

lhiginbotham:
@RussKie should this be rebased onto origin/release/4.0? Or should both versions (3 and 4) get the fix?  If so, not sure how the docs update will go, never did a doc change before, let alone a version specific change.

RussKie:
> should this be rebased onto origin/release/4.0? Or should both versions (3 and 4) get the fix?

v3 is officiall done. We'll service release/4.0 branch with this after the merge.

GitUI/CommandsDialogs/FormMergeBranch.cs
```diff
@@ -144,7 +143,8 @@ private void NonDefaultMergeStrategy_CheckedChanged(object sender, EventArgs e)
 
         private void strategyHelp_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
         {
-            strategyToolTip.SetToolTip(strategyHelp, _strategyTooltipText.Text);
+            OsShellUtil.OpenUrlInDefaultBrowser(
+                UserManual.UserManual.UrlFor("branches", "advanced-merge-options"));
```
RussKie, line 147: This doesn't look pretty... Please use `using`.

RussKie (commented):
🎉

GitUI/CommandsDialogs/FormMergeBranch.cs
lhiginbotham: @RussKie I tried the following but was met with a compilation error:

image

image


My only guess is that since the namespace name is the same as the class name, the compiler is confused, thinking I am intending the namespace name instead.

Any suggestions here or just leave it alone?

GitUI/CommandsDialogs/FormMergeBranch.cs
```diff
@@ -144,7 +143,8 @@ private void NonDefaultMergeStrategy_CheckedChanged(object sender, EventArgs e)
 
         private void strategyHelp_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
         {
-            strategyToolTip.SetToolTip(strategyHelp, _strategyTooltipText.Text);
```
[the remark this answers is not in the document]
lhiginbotham: @RussKie good catch - removed all references, thanks!

vbjay:
Just my $0.02. I think the help should be offline first and a link to online can be in the offline help. Take a page from git help. Assume not online and should still be able to get the help. 

```bash
git log --help
```

gerhardol:
> Just my $0.02. I think the help should be offline first and a link to online can be in the offline help. Take a page from git help. Assume not online and should still be able to get the help.
> 
> ```shell
> git log --help
> ```

The GE documentation need a serious refresh before something like that may be considered. There has not been many contributions, documentation is not always fun.
(I have spent considerable hours, the doc is still incomplete.)
Git doc is easier as it describes specific commands, where GE has more 'flows' with commands combined.

gerhardol (approved):
Add icon too?
Otherwise fine for me
image

GitUI/CommandsDialogs/FormMergeBranch.cs
RussKie: No worries then.

lhiginbotham:
> Add icon too? Otherwise fine for me 
image

@gerhardol @RussKie 

There is a class `GotoUserManualControl` which seems to be a help icon followed by a help link to a parameterizable help manual URL.  Sadly, I cannot seem to use this here as I get this error when trying to insert it via the designer:
image


And if this is a problem while using the designer, my guess is that if I manually insert it via code (not sure all the properties that would need set if going this route) then the designer probably would still be broken.

We can a) leave it as is, b) insert a plain icon and keep the plain link, or c) I'll need some help from one of you in figuring out how to get the GotoUserManualControl to work with the designer.

What do you think?

mstv (commented):
+0, (i.e. technically OK, no veto)
Luckily, the information can be found at the top of the destination page - in this case...
(Users must live with English explanation only - regardless of that I prefer development tools in English myself, since a company switched to automated translation. And one does not need to guess the English wording when searching for help or reporting issues.)

gerhardol:
> We can a) leave it as is, b) insert a plain icon and keep the plain link, or c) I'll need some help from one of you in figuring out how to get the GotoUserManualControl to work with the designer.

For me it is OK if the designer cannot be used to configure the page, if the designer does not crash.
But for now, I suggest to leave it as is.

RussKie:
> +0, (i.e. technically OK, no veto) Luckily, the information can be found at the top of the destination page - in this case... (Users must live with English explanation only - regardless of that I prefer development tools in English myself, since a company switched to automated translation. And one does not need to guess the English wording when searching for help or reporting issues.)

I thought we had the docs translated

gerhardol:
> I had a look, and apparently, we don't translate the docs. It looks like it's [configurable](https://docs.readthedocs.io/en/stable/localization.html#project-with-multiple-translations), but I must be dumb as I can't understand what's required.

https://readthedocs.org/dashboard/git-extensions-documentation/translations/

There are no subprojects with translations defined, and no attempts to translate (branches in the gitextdoc repo?)
But the doc need to be more comprehensive first, it is not up to date.

gerhardol:
Thanks
Including in vNext only for now. This affects the translations, even if just removing a string, but builds will raise warnings.

PR #10303, opened 2022-10-28
RevDiff: Allow path filter for several files

gerhardol:
Companion to #10295 

## Proposed changes

Allow multi select files in rev diff to filter multiple files. See #10295 for a use case.
It is possible to filter folders in file tree today, but no GUI way to select multiple files.
Multiple files can be added in the advanced filter form though.
The first file is visible in the title
image
All files are visible in the tool tip
image

This exposes the possibility to add several files in rev diff file status list.
I believe this covers the important use case, I have thought about some enhancements and rejected them:

- Possibility to add/remove files to the filter in any form. 
  Adding would be simple, but not so easy to explain in the menu.
- Context menu to remove files in the filter in any form. 
  Would require some data structure changes (files stored as string right now), easier to understand in context menus though.
- Multi-select files in FileTree
  Not enabled now.
- Open a separate window (browse or old file history) with the selected files.
  Possible, but would require some call changes.

### Before

Multiselect disables file filters.

### After

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.InitMenusAndToolbars.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.InitMenusAndToolbars.cs
@@ -103,7 +103,7 @@ namespace GitUI.CommandsDialogs
 
                 if (!string.IsNullOrWhiteSpace(pathFilter))
                 {
-                    SetPathFilter(pathFilter);
+                    SetPathFilter(pathFilter.QuoteNE());
                 }
             }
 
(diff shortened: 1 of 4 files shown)


mstv (approved):
LGTM

mstv:
We should take this into v4.0.2 because it might resolve https://github.com/gitextensions/gitextensions/pull/10438#issuecomment-1352773517.

PR #10307, opened 2022-10-29
Add control with last interactive Git outputs and traces

### LINKED ISSUE #10206: Keep git output
### Feature description

Currently we have 2 options regarding the output printed by potentially long-running git commands, such as `fetch`, `checkout`, etc.:

1. Keep the dialog open after completion of every such command. This may show some useful extra info, or a warning, but it slows you down when you need to do many little things and can get annoying.
2. Dismiss the output right away. This speeds up routine operations, but it deprives the user from information that git developers thought could be useful to show.

There is a "Git command log" window, but it can't show the commands' output, as far as I can tell.

I think a middle ground would be good here: save the output _for later_ and don't require the user to read and acknowledge it before proceeding. The log could be integrated into the existing "Git command log", but maybe it even deserves a tab of its own next to "Console" to make it easy to keep an eye on.

### Environment

- Git Extensions 3.5.4.12724
- Git 2.33.0.windows.2
- .NET Framework 4.8.4515.0
- DPI 120dpi (125% scaling)

mstv:
Fixes #10206

## Proposed changes

- FormBrowse: Add control which is filled with the output from the last 20 Git commands executed in FormStatus
  as well as with trace output
- Show the control as Tab or docked in the lower left corner (GE needs to be restarted in order to bring changed settings into effect)
- Hotkey: `Ctrl+9` for toggling panel visibility and focusing

### Before

image

### After

image

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 63a2b45ca4d9bd313451c6ebec1bd75015a8c735
- Git 2.38.1.windows.1
- .NET 6.0.10
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 3.1.21 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 5.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

## Merge strategy

----


mstv:
@PeterMinin, you could try https://ci.appveyor.com/project/gitextensions/gitextensions/builds/45223566/artifacts.

If it was not such complicated, I would rather place the git output in the lower left corner in order to have it visible always.

image

gerhardol:
Nice!

Changing hotkeys could be sensitive, but probably OK here...
Any commands with more than one output?
What is the effort to add to command log, also for other commands?
Should be added to settings, like console and gpg (and build status). Should be default off.
A context menu to copy would be nice too - ctrl-c works though.

PeterMinin:
Thanks @mstv! This looks useful already. However my idea was to keep output from multiple commands, so that you could return to it after you've done something else. I think that would also mean writing the commands themselves, otherwise it would be hard to navigate.

I took a look at the code to see if it's easy to implement that way. Although `FormStatus` doesn't have access to the command executed, its child `FormProcess` does. I think it's ok to move the hook to `FormProcess`. And perhaps it's even better, because some uses of `FormStatus`, namely `FormStatus.ShowErrorDialog`, aren't for long-running operations. For example, as far as I understand, and error in the staging dialog (which I usually encounter if I try to stage specific lines but have a wrong encoding selected in the viewer) uses that form, which I don't think is useful to keep. What do you think?

Also I guess it would be nice to show some placeholder before there is output to show.

mstv:
I have updated the description.

to be done?
1. internal `AppSetting` for number of displayed results
2. hide by default
3. store splitter position
4. add toggle button to main toolbar or to Left Panel toolbar:
image

gerhardol:
> to be done?
>    internal AppSetting for number of displayed results

yes. Maybe ttal size too?

>    hide by default
 
Probably

*   store splitter position

Related to this PR?

*    add toggle button to main toolbar or to Left Panel toolbar:

I do not want that there, maybe context menus on the tabs (in addition to AppSettings, as for other tabs)

I also want:
* Context menu in the tab, see GPG info, to select/copy text
* Move the tab to rightmost position. Will not change existing shortcuts, also separate use.

No direct comments on the code.

mstv:
> I have updated the description.

... but forgot to push the update button.
There is no tab anymore but a splitter added to the Left Panel.

> see GPG info

Could you elaborate?

gerhardol:
> ... but forgot to push the update button. There is no tab anymore but a splitter added to the Left Panel.

I liked the tab better than this panel
It always occupies space, it is not often I want to view the contents.
With this, I would deactivate by default, rather have the info in Command Log.

> Could you elaborate?

image

Compile error

image

Still no major code comments, I will review (likely approve) when we agree on the display.

mstv:
> I liked the tab better than this panel
It always occupies space, it is not often I want to view the contents.

I don't need it always - but if, switching tabs is a nuisance. It is becoming a killer feature for me. Toggle it using `Ctrl+9`. Tab can be implemented as an option.

> Compile error

My bad. Removed.

RussKie:
I don't like it, sorry. It looks very untidy. I would have thought that we'll have a new tab of some kind....

Also, any new controls should be encapsulated as UserControls (or whatever) and not be coded directly in FormBrowse. We should be decoupling and encapsulating elements as much as possible.

mstv:
I should have marked it as draft / PoC more clearly... Extraction to a dedicated control is planned for after getting a consensus on the UX.

The variant as Tab is available in the first commit. But believe me it is not very usable.

vbjay:
Tried it out

image


Didn't create a new line between commands.  

Maybe we can take a page from visual studio.  This would be a great thing for auto hide from bottom of window.  

image
image

RussKie:
Expandable/movable windows would be great, but it'd either require 3rd party controls (💰💰💰) or custom implementations (⌛⌛⌛).

gerhardol:
I have been using a previous version of this for dome time. No direct comments when browsing this update.
I do not get any output when using this PR though. Hints?
image

Rebase on master to clear the merge conflicts?

mstv:
> Edit: Still fails for me.

It has always worked for me.
It is intended that only output from foreground git calls is shown.
I am running out of ideas. Please do not tell me it was caused by the preview versions of VS or .NET8.

BTW: I do not get notifications on edited comments.

gerhardol:
> > Edit: Still fails for me.
> 
> It has always worked for me. It is intended that only output from foreground git calls is shown. I am running out of ideas. Please do not tell me it was caused by the preview versions of VS or .NET8.

I have tested this from your branch and from command line build, same result. (If I had run it in other situations, I would have noted that when posting.)
Have planned to try debugging, not got started.

> BTW: I do not get notifications on edited comments.

I know, it was not new information.

gerhardol:
>  It is intended that only output from foreground git calls

I got some output:

```
23:24 C:\Program Files\Git\bin\git.exe submodule update --init --recursive
Submodule path 'Externals/ICSharpCode.TextEditor': checked out '4a380941467e67088566dffcc102ff93ce90ce1d'
Submodule path 'Externals/conemu-inside': checked out 'fbcb27b7522603a797f3b2aad1c8dcd3caaecf7c'

23:25 C:\Program Files\Git\bin\git.exe clean -x -d --dry-run
```

Fewer commands than I expected

mstv:
> Fewer commands than I expected

This history control is intended for keeping output the user has seen in FormProcess and - in the future - in exception popups, as well as for displaying error messages which we do not want to bubble up and to abort actions.
It shall not replace the Git Command Log. Otherwise, it would be flooded with less interesting output, e.g. from "git log".

gerhardol:
> This history control is intended for keeping output the user has seen in FormProcess and - in the future - in exception popups, as well as for displaying error messages which we do not want to bubble up and to abort actions. It shall not replace the Git Command Log. Otherwise, it would be flooded with less interesting output, e.g. from "git log".

"Background" commands including those related to a refresh and selecting a new revision should not be included.
But I expected commands related to (other) interactions to be displayed.
A normal user has quite limited knowledge of commands required, just expecting "something".

fetch and push is not included, that could be the example in the feature description
Commands related to to popups like reset another branch could be included too

mstv:
> But I expected commands related to (other) interactions to be displayed.
> Commands related to to popups like reset another branch could be included too.

These are definitely to be dealt with in a follow-up. We need to choose and to mark the relevant commands somehow, as well as to intercept their output.
This PR "just" keeps the output shown in FormStatus and descendants.

mstv:
> > But I expected commands related to (other) interactions to be displayed.
> > Commands related to to popups like reset another branch could be included too.
> 
> These are definitely to be dealt with in a follow-up. We need to choose and to mark the relevant commands somehow, as well as to intercept their output. This PR "just" keeps the output shown in FormStatus and descendants.

... as well as (selected) Trace output.

Whether the output from which other Git commands should be kept by using FormStatus or by using Trace, should be discussed separately in a follow-up.

GitCommands/Settings/AppSettings.cs
```diff
@@ -470,6 +470,8 @@ public static bool ShowResetAllChanges
         public static ISetting<bool> ShowConEmuTab { get; } = Setting.Create(DetailedSettingsPath, nameof(ShowConEmuTab), true);
         public static ISetting<string> ConEmuStyle { get; } = Setting.Create(DetailedSettingsPath, nameof(ConEmuStyle), "<Solarized Light>");
         public static ISetting<string> ConEmuTerminal { get; } = Setting.Create(DetailedSettingsPath, nameof(ConEmuTerminal), "bash");
+        public static ISetting<bool> ProcessHistoryAsTab { get; } = Setting.Create(DetailedSettingsPath, nameof(ProcessHistoryAsTab), true);
```
gerhardol, line 473: Maybe false for now, until there is more info?

mstv: This settings toggles between the Tab and the docked Panel.
If we really would want to hide it, the default of `ProcessHistoryDepth` would be set to `0`.

> until there is more info

This shall never replace the Git Command Log. I want to have the recent interactive git output at hand (with clickable links a.o.), and a trace of those exceptions and hints which shall not be presented using modal popup.
Please elaborate what you think was missing here.

pmiossec:
Unfortunately AFK during a week.
I will keep in mind for next week...

gerhardol:
I do not have any direct comments on the implementation.
Panel<->tab requires a restart, maybe mention with tooltip.
Usage should be mentioned in the tooltip, if panel toggle to see.
When first starting, the panel was covering the file status list and lower part of left panel, but I have installed previous versions of this PR soit may be due to that.

gerhardol (commented):
Add info in Settings and I will approve.

mstv:
> Panel<->tab requires a restart, maybe mention with tooltip.
> Usage should be mentioned in the tooltip, if panel toggle to see.

Good catch! I have found a workaround with better UX than tooltip.

restart hint is here:

image

> When first starting, the panel was covering the file status list and lower part of left panel.

Valid point! Now, the size should be limited by means of `LeftSplitContainer.FixedPanel = FixedPanel.Panel2;`
I wanted to set it, but it got lost somehow.

GitExtUtils/SubscribableTraceListener.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System.Diagnostics;
+using System.Text;
+
+namespace GitExtUtils;
+
+public delegate void TraceHandler(in string message);
+
+/// <summary>
+///   Provides an event for receiving Trace output.
```
RussKie, line 9:
```suggestion
///   Provides an event for receiving the trace output.
```

GitExtUtils/SubscribableTraceListener.cs
```diff
@@ -0,0 +1,82 @@
...
+/// </summary>
+public interface ISubscribableTraceListener
+{
+    event TraceHandler TraceReceived;
+}
+
+public class SubscribableTraceListener : TraceListener, ISubscribableTraceListener
+{
+    private StringBuilder _trace = new();
```
RussKie, line 18:
```suggestion
    private readonly StringBuilder _trace = new();
```

GitExtUtils/SubscribableTraceListener.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (disposing)
+        {
+            Trace.Listeners.Remove(this);
+        }
+
+        base.Dispose(disposing);
+    }
+
+    public event TraceHandler TraceReceived;
```
RussKie, line 35: [nit] move to the top of the class

GitUI/Models/ProcessHistoryModel.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+
+            sb.AppendLine(_endMark);
+
+            return sb.ToString();
+        }
+    }
+
+    public event EventHandler HistoryChanged;
```
RussKie, line 52: [nit] move to the top of the class before the .ctor

GitUI/Models/ProcessHistoryModel.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace GitUI.Models;
+
+public interface IProcessHistoryModel
+{
+    bool Enabled { get; }
+    string History { get; }
+
+    event EventHandler HistoryChanged;
+
```
RussKie, line 12:
```suggestion
    event EventHandler HistoryChanged;

    bool Enabled { get; }
    string History { get; }

```

RussKie (commented):
Looks good, thank you.

Is it possible to add an animation or a video showing the control and the interaction with the control in action?

GitExtUtils/SubscribableTraceListener.cs
mstv: Added `<see cref="" />`

GitExtUtils/SubscribableTraceListener.cs
mstv: This is not a type but a provided event.
I followed https://learn.microsoft.com/en-us/archive/blogs/brada/internal-coding-guidelines#28-file-organization (Is it outdated?)
and took the liberty to put `Dispose` in the ctors section. Should I move `Dispose` to the methods section?

mstv:
> Is it possible to add an animation or a video showing the control and the interaction with the control in action?

added to PR description

gerhardol:
> If explanation should be added here, I could use a ToolTipCheckBox with:
    focus/toggle the history using hotkey
    disable process history by setting depth = 0
    all changes require a restart to come into effect

The description that the hotkey toggles the view of the panel is the primary. 
An explanation of the info is useful too.
Also use InfoClicked to open this part of the settings in the doc

gerhardol:
> > If explanation should be added here, I could use a ToolTipCheckBox with:

I added the settings to a temp page. Feel free to push to this branch too.
https://git-extensions-documentation.readthedocs.io/en/tmp-4.3/settings.html#general-process-history-as-tab-otherwise-as-panel
(just placeholders right now, some diff changes drafts)

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
```diff
@@ -1,10 +1,25 @@
...
+        private const string _processHistoryUrl = "https://git-extensions-documentation.readthedocs.io/settings.html#process-history-as-tab-otherwise-as-panel";
+        private readonly TranslationString _processHistoryTooltip
+            = new("""
+                  The output displayed in the process dialog and the trace output is retained and shown in the Process History.
+                  Focus the Process History or toggle its visibility using the hotkey {0}.
+                  With this set, the Process History is displayed in a tab in the lower pane of the Browse Repository window.
+                  With this unset, the Process History is displayed in a panel docked to the lower left corner of the Browse Repository window.
+                  The panel also uses some space of the file status list of the Diff tab.
+                  The panel height can be adapted using the splitter control of the Left Panel.
+                  The Process History can be disabled by setting the process history depth to 0.
```
gerhardol, line 21: Just suggest shortening, hard to read now
```suggestion
                  The output displayed in the process dialog and the trace output is retained and shown in the Process History.
                  Focus the Process History or toggle its visibility using the hotkey {0}.
                  - With this set, the Process History is displayed in a tab in the lower pane of the Browse Repository window.
                  - With this unset, the Process History is displayed in a panel docked to the lower left corner of the Browse Repository window.
                  The Process History can be disabled by setting the process history depth to 0.
```

gerhardol (approved):

GitExtUtils/SubscribableTraceListener.cs
RussKie: I've largely been using that guidelines too (and pointing others at). The old event syntax (`public event EventHandler FooChanged { add { … } remove { … } }`) makes it look like properties, so in that context it makes sense. Though we haven't used that syntax in a long time (apart from dotnet/winforms codebase, hah). 
In this codebase (though we are not very consistent) many types define event handlers either before or right after .ctors. It generally makes it easier to discover.

But it's a nit.

RussKie:
> ### After
> history_control

I must admit, this looks weird. The output window is spanned across two individual unrelated panels (the left panel and the bottom panel), then when you switch tabs, all of a sudden, the output window only constrained to one... What if the left or the bottom panel are collapsed?

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.Designer.cs
```diff
@@ -108,23 +115,59 @@ private void InitializeComponent()
...
+            _NO_TRANSLATE_ProcessHistoryDepth.Value = new decimal(new int[] { 1000, 0, 0, 0 });
+            // 
+            // lblProcessHistoryDepth
+            // 
+            lblProcessHistoryDepth.AutoSize = true;
+            lblProcessHistoryDepth.Location = new Point(9, 94);
+            lblProcessHistoryDepth.Name = "lblProcessHistoryDepth";
+            lblProcessHistoryDepth.Size = new Size(123, 15);
+            lblProcessHistoryDepth.TabIndex = 9;
+            lblProcessHistoryDepth.Text = "Process history depth:";
```
pmiossec, line 154: I prefer to happen "(0 to disable)" in the label than having the explanation in the tooltip that I find already difficult to read.

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -194,6 +194,7 @@ T   Translate
 
         private readonly TranslationString _buildReportTabCaption = new("Build Report");
         private readonly TranslationString _consoleTabCaption = new("Console");
+        private readonly TranslationString _historyTabCaption = new("History");
```
pmiossec, line 197: I don't like the "History" name of the tab. I found it too misleading because too close to the notion of "git history". I think the (new) users will expect to find something else displayed in a tab named like that.
I prefer something more like "Trace", "Output" or "Processes"...

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.Designer.cs
```diff
@@ -108,23 +115,59 @@ private void InitializeComponent()
...
+            // 
+            chkProcessHistoryAsTab.AutoSize = true;
+            chkProcessHistoryAsTab.AutoSizeMode = AutoSizeMode.GrowAndShrink;
+            chkProcessHistoryAsTab.Checked = false;
+            chkProcessHistoryAsTab.Location = new Point(9, 72);
+            chkProcessHistoryAsTab.Margin = new Padding(4, 3, 4, 3);
+            chkProcessHistoryAsTab.Name = "chkProcessHistoryAsTab";
+            chkProcessHistoryAsTab.Size = new Size(249, 19);
+            chkProcessHistoryAsTab.TabIndex = 8;
+            chkProcessHistoryAsTab.Text = "Process history as tab (otherwise as panel)";
```
pmiossec, line 166: "Show Process history as tab (otherwise as panel)" or "Display Process history as tab (otherwise as panel)" (like all the settings in this setting panel)?

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
```diff
@@ -1,10 +1,23 @@
...
     public partial class FormBrowseRepoSettingsPage : SettingsPageWithHeader
     {
+        private const string _processHistoryUrl = "https://git-extensions-documentation.readthedocs.io/settings.html#process-history-as-tab-otherwise-as-panel";
+        private readonly TranslationString _processHistoryTooltip
+            = new("""
+                  The output displayed in the process dialog and the trace output is retained and shown in the Process History.
+                  Focus the Process History or toggle its visibility using the hotkey {0}.
+                  - With this set, the Process History is displayed in a tab in the lower pane of the Browse Repository window.
+                  - With this unset, the Process History is displayed in a panel docked to the lower left corner of the Browse Repository window.
+                  The Process History can be disabled by setting the process history depth to 0.
```
pmiossec, line 19: This tooltip is a little hard to read, so suggestion to:
- add empty lines to make easier to read
- Put to "focus/hide", which is not "really" linked to the setting explanation at the end
- reword "set" to "check"

I also propose to remove the disable part to put it in the label (as already said in another comment) to make this tooltip a little more easier to read.

```suggestion
                  The output displayed in the process dialog and the trace output is retained and shown in the Process History.

                  - With this checked, the Process History is displayed as a tab in the lower pane of the Browse Repository window.
                  - With this unchecked, the Process History is displayed in a panel docked to the lower left corner of the Browse Repository window.

                  The Process History can be disabled by setting the process history depth to 0.
                  Focus the Process History or toggle its visibility using the hotkey {0}.
```

pmiossec:
Some other comments that others than the ones done in the review.

1. Should this type of tab be displayed *by default* because it is technical (due to the traces also displayed)? It was my case. But maybe because I have the setting to auto-close the git process console form. Is it how it works or will it be displayed for everyone by default?

2. Experience about Hide/Focus with shortcut is surprising...
Maybe updating the tooltip to be clearer with "Focus (when displayed as a tab) the Process History or toggle its visibility (when displayed as a panel) using the hotkey"
**or** when displayed as a panel adopt a 3 states workflow: focus => hidden => shown  or said differently shortcut should maybe bring focus to panel and only next time hide the panel.  

3. Not all lines are displayed with the same police:
image

4. Add it to the menu after "Tools -> Git Command Log"?
5. I find the UX when displayed as panel not perfect (but it could be good enough if it's only dedicated to "power users")
	* Not displayed if left panel is not visible
	* Vertical resize is only available on "left panel" separator (not diff)
history_resize

	* Behavior is different when other tabs than "Diff" are selected. We could expect that the behavior is the same with filetree tab.
image

GitUI/CommandsDialogs/FormBrowse.cs
mstv: I think "Output" as tab label is well-known and fits best. I am going to use "OutputHistory" in class names.

mstv:
> Should this type of tab be displayed by default because it is technical (due to the traces also displayed)? It was my case. But maybe because I have the setting to auto-close the git process console form. Is it how it works or will it be displayed for everyone by default?

I forgot to file a PR with ed73605d32b02f6f4be6acd3f8883040bd5cac37.
I am going to filter out the `RevisionReader` output, too. I kept it for the ongoing optimizations.
The only other message I noticed, is `"Call for all branches after cache filled with specific branch {_branchName}"`.
Perhaps, we should add a hidden setting with an regex for filtering the trace output.

mstv:
> Experience about Hide/Focus with shortcut is surprising...
Maybe updating the tooltip to be clearer with "Focus (when displayed as a tab) the Process History or toggle its visibility (when displayed as a panel) using the hotkey"

👍 I am going to rename the hotkey, too.

> or when displayed as a panel adopt a 3 states workflow: focus => hidden => shown or said differently shortcut should maybe bring focus to panel and only next time hide the panel.

For the rare case you want to _focus_ the output panel, just hit the key twice. Hiding is needed more often, e.g. in case there are many files in the FileStatusList.

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
mstv: done but without
> reword "set" to "check"

because this wording is used in other places, too

---

I am going to rename the classes later.

pmiossec:
> > Should this type of tab be displayed by default because it is technical (due to the traces also displayed)? It was my case. But maybe because I have the setting to auto-close the git process console form. Is it how it works or will it be displayed for everyone by default?
> 
> I forgot to file a PR with [ed73605](https://github.com/gitextensions/gitextensions/commit/ed73605d32b02f6f4be6acd3f8883040bd5cac37).

My question was more about if the default is to have the tab displayed or is it something that the user will enable?
I was questioning myself on the real goal of this feature which has an impact on if the tab should be visible or not.

If I take my personal case, **since I know a little git and I'm confident on it**, I hide the git output and rarely expressed the need to see git outputs. So I was wondering if it will propose more like a debug or understanding tool that user will enable or if it will be always visible (so by default).

If I understand well the settings code, as the default is display as tab, it will be visible by default, right!?!

(As a consequence, some of my remarks, like 4. doesn't stand anymore...)

GitUI/Models/ProcessHistoryModel.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            _processHistory.Clear();
+        }
+
+        HistoryChanged?.Invoke(this, EventArgs.Empty);
+    }
+
+    public void Trace(in string message)
+    {
+        Add(new StringBuilder().Append(DateTime.Now.ToShortTimeString()).Append(' ').AppendLine(message));
```
pmiossec, line 66:
```suggestion
        Add(new StringBuilder(message.Length + 6).Append(DateTime.Now.ToShortTimeString()).Append(' ').AppendLine(message));
```

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
```diff
@@ -1,10 +1,24 @@
...
     {
+        private const string _processHistoryUrl = "https://git-extensions-documentation.readthedocs.io/settings.html#process-history-as-tab-otherwise-as-panel";
+        private readonly TranslationString _outputHistoryTooltip
+            = new("""
+                  The output displayed in the process dialog and some trace output is retained and shown in the output history.
+
+                  - With this set, the output history is displayed in a tab in the lower pane of the Browse Repository window.
+                  - With this unset, the output history is displayed in a panel docked to the lower left corner of the Browse Repository window.
+
+                  Focus the output history or toggle the panel visibility using the hotkey {0}.
```
pmiossec, line 20: Still convinced that could be improved with:
```suggestion
                  Focus (when displayed as a tab) the output history or toggle (when displayed as a panel) the panel visibility using the hotkey {0}.
```

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
```diff
@@ -77,5 +106,10 @@ private void cboTerminal_Enter(object sender, EventArgs e)
         {
             _cboTerminalPreviousIndex = cboTerminal.SelectedIndex;
         }
+
+        private void chkProcessHistoryAsTab_InfoClicked(object sender, EventArgs e)
+        {
+            OsShellUtil.OpenUrlInDefaultBrowser(_processHistoryUrl);
```
pmiossec, line 112: A little surprising this tooltip icon that is also a clickable help link.
I only think about clicking while seeing this code.
I see that we have that elsewhere but here the cursor change to the "hand" is missing...

GitUI/Hotkey/HotkeySettingsManager.cs
```diff
@@ -266,6 +266,7 @@ public IReadOnlyList<HotkeySettings> CreateDefaultSettings()
                     Hk(FormBrowse.Command.StashPop, Keys.Control | Keys.Alt | Keys.Down),
                     Hk(FormBrowse.Command.StashStaged, Keys.Control | Keys.Shift | Keys.Alt | Keys.Up),
                     Hk(FormBrowse.Command.ToggleBetweenArtificialAndHeadCommits, Keys.Control | Keys.OemBackslash),
+                    Hk(FormBrowse.Command.ToggleAndFocusOutputHistory, Keys.Control | Keys.D9),
```
pmiossec, line 269: Link to the behavior, maybe naming it `FormBrowse.Command.Toggle**Or**FocusOutputHistory` makes more sense...

mstv:
> My question was more about if the default is to have the tab displayed or is it something that the user will enable?
I was questioning myself on the real goal of this feature which has an impact on if the tab should be visible or not.

- As integrator, I use it to check what has been pushed to the repo after fetch.
- Git often gives useful hints
- and links, e.g. for opening a PR.

GitUI/CommandsDialogs/SettingsDialog/Pages/FormBrowseRepoSettingsPage.cs
mstv: Let's find a clearer wording (none of my strengths ;) because the panel is focused, too, when it is shown.

pmiossec:
I pushed a commit to remove console output color markers to have a better display looking like that:

image

instead of that:

image

gerhardol:
> I pushed a commit to remove console output color markers to have a better display looking like that:

That parsing is already in GitUI/Editor/Diff/AnsiEscapeUtilities.cs (but not much extra here)
If the control uses ICSharpCode.TextEditor.Document the routine could be used to display the colors too.

mstv:
> If the control uses ICSharpCode.TextEditor.Document the routine could be used to display the colors too.

What needs to be configured in order to get colored git output, @pmiossec?

gerhardol:
> > If the control uses ICSharpCode.TextEditor.Document the routine could be used to display the colors too.
> 
> What needs to be configured in order to get colored git output, @pmiossec?

If it was me you were asking, GitUI/Editor/Diff/AnsiEscapeUtilities.cs will escape the text and convert the escapes to the markers ICSharpText uses to mark text (like from pos 1:3 red)

mstv:
> > > If the control uses ICSharpCode.TextEditor.Document the routine could be used to display the colors too.
> > 
> > 
> > What needs to be configured in order to get colored git output, @pmiossec?
> 
> If it was me you were asking, GitUI/Editor/Diff/AnsiEscapeUtilities.cs will escape the text and convert the escapes to the markers ICSharpText uses to mark text (like from pos 1:3 red)

I will need this part then.
But the major question remains: How to tell git to emit the ANSI escape sequences.

gerhardol:
> But the major question remains: How to tell git to emit the ANSI escape sequences.

By default Git do not emit escape sequences if the output is not a terminal.
Many commands can force color output:
https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---colorltwhengt

So I set this option for git-diff etc for the git coloring, they also must be stored in the command cache.
GE will clean the escape sequences by default.

In the example for git-push the terminal was the original output I assume so colors were seen.

mstv:
> By default Git do not emit escape sequences if the output is not a terminal.

👍 reproducible with ConEmu and forget to force push
I have not seen colored git output for long time because I stopped using ConEmu several months ago (PR will follow) and I use to tick the "force" checkbox because I do not want to wait for the TaskDialog and then wait again.

---

I have changed it to use `AnsiEscapeUtilities.ParseEscape`, i.e. it is prepared for colors. And colors will be useful.
But this PR is big enough and `TextEditorControl` as "editor" needs significant glue code. Let's add colors in a follow-up.

pmiossec (approved):

mstv:
What is really missing yet? Consider please that the variant to display the output control as panel is not the default. Who does not like it can switch back to the Tab variant.

src/app/GitUI/UserControls/OutputHistoryControl.cs
```diff
@@ -0,0 +1,14 @@
+﻿using GitCommands;
+using ResourceManager;
+
+namespace GitUI.UserControls;
+
+public partial class OutputHistoryControl : GitExtensionsControl
```
RussKie, line 6: If there are no specific reasons I'd encourage keeping all new types to be `internal` rather than `public`. Generally, only interfaces that participate in DI should be public.

RussKie (commented):
I think the implementation has gone a long way and it looks much cleaner now. There are few comments but they more on the naming and styling side.

There is one thing, however, that I don't like about this implementation - and it is the positioning of the output window under the left panel across the boundaries of the two splitcontainers
image


IMO this makes our (already uber complex) design/layout and implementation more complex, as well as visually unbalances the layout.

Is there a scenario you wouldn't think it'd work in a tab page?

gerhardol:
> IMO this makes our (already uber complex) design/layout and implementation more complex, as well as visually unbalances the layout.
> 
> Is there a scenario you wouldn't think it'd work in a tab page?

The panel allows you to see the output at the same time as other info.
An alternative is an additional popup.

(This is useful, but has not yet the information I expected, but having an additional tab is fine.)

RussKie:
> The panel allows you to see the output at the same time as other info. An alternative is an additional popup.
> 
> (This is useful, but has not yet the information I expected, but having an additional tab is fine.)

We do have Git CommandLog (F12) which is a long-lived popup dialog. I kind of feel that we're creating competing features located in various places.
Would it make sense for the new log panel to live on the Git CommandLog dialog - just add another tab to it.

This way - we'll need no UI changes to the main form at all (which is a big-big plus IMO), and users will get the new cool feature at the familiar place.

src/app/GitUI/Models/OutputHistoryModel.cs
```diff
@@ -0,0 +1,122 @@
...
+    bool Enabled { get; }
+    string History { get; }
+
+    void Clear();
+    void Trace(in RunProcessInfo runProcess);
+    void Trace(in Exception exception);
+    void Trace(in string message);
+}
+
+public sealed class OutputHistoryModel : IOutputHistoryModel
```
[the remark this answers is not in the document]
mstv: > I'm having troubles reconciling the name with the way it's used, it just doesn't fit my mental model:
> We call it a "model" but we use it as "service" which performs actions. IMO it should be called in a way that denotes actions. E.g., something like IGitOutputLogHistoryProvider.
> image

It is a model. It manages and formats data by means of actions, properties and a change event. It does not retrieve the data, what a provider would do.
`AddService`/`GetRequiredService`: You chose the term "service" for DI. Perhaps it is too specific.

I am renaming the methods as suggested (although I prefer to avoid the repetition of terms - `History` in this case). So the above code will become clearer.

mstv:
> RE: tabs and Git Command Log

The display as tab _is_ the default.
But _I_ definitely want to have it always visible. There uses to be free space in the lower left, and I want to use it to have this information on hand. This strong wish has been clear since the first use: https://github.com/gitextensions/gitextensions/pull/10307#issuecomment-1296310050.
I do not want to juggle with tabs or floating windows in addition to multiple instances of GE. We have just different preferences like about CommitInfo.

Git Command Log is pretty orthogonal to the Output History: diagnosis feature with details about process execution vs. normal user feature with focus on git output.

F.i. I need to watch many repos like this. (Commit hashes might be made clickable later, too.)

image

A PR is just one click away (without the need to configure plugins):

image

mstv:
Finally. it works with a second splitter!

mstv:
Can we finally merge this PR?
It is a nuisance to not have this feature when testing PRs.

RussKie:
Sorry, I can't bring myself to approve with this layout:
image

mstv:
> Sorry, I can't bring myself to approve with this layout:

Then use the tab, which is the default, or turn the entire feature off. I want it, and I am not forcing you to want it.
I am asking whether there are significant concerns about the code itself.

mstv:
I have updated the screenshot in the PR description:

image

gerhardol:
How to progress this PR?

mstv:
> How to progress this PR?

I am going to merge it after v5.1 (which - by the way - is almost ready for release) as I have two approvals and no veto.

gerhardol:
> > How to progress this PR?
> 
> I am going to merge it after v5.1 (which - by the way - is almost ready for release) as I have got two approvals and no veto.

Next should be 5.1, from master not cherry picking to a 5.0.1 (as the milestone).
Then master could migrate to .net9 and possibly add a dark theme again.
For me, this could be added before 5.1

pmiossec:
> For me, this could be added before 5.1

I agréé.

PR #10338, opened 2022-11-05
.NET 6.0.10

gerhardol:
## Proposed changes

Update to latest version for next release.
Also a way to test the prompts to update...

## Test methodology 

Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build a2f795eac8fe7b82b8906ff07581999431bead20
- Git 2.38.1.windows.1
- .NET 6.0.10
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 ﻿{
     "sdk":  {
-                "version":  "6.0.401",
+                "version":  "6.0.402",
                 "rollForward":  "feature"
             }
 }
--- a/scripts/RepoLayout.props
+++ b/scripts/RepoLayout.props
@@ -9,7 +9,7 @@
     <!-- Set version for runtimeconfig.json, used by DotnetRuntimeBootstrapper for minimal version to support
          To be kept in sync with the .NET SDK version in global.json
     -->
-    <RuntimeFrameworkVersion>6.0.9</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>6.0.10</RuntimeFrameworkVersion>
     <SolutionTargetFramework>net6.0-windows</SolutionTargetFramework>
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>


RussKie (approved):

mstv:
Perhaps a coincidence, but since this merge many tests fail with NRE.

vbjay:
You know. They did release 11 pretty quick. I wonder if 10 is a bad version.

mstv:
> Also a way to test the prompts to update...

OK

> Update to latest version for next release.

But we should not take every minor version increment - only if they have fixed something important.

> You know. They did release 11 pretty quick. I wonder if 10 is a bad version.

That's why it might be good to wait a little in order to let others do the beta or gamma testing. We are too busy.

RussKie:
Servicing releases are shipped every month, 10 - October, 11 - November.

scripts/RepoLayout.props
```diff
@@ -9,7 +9,7 @@
     <!-- Set version for runtimeconfig.json, used by DotnetRuntimeBootstrapper for minimal version to support
          To be kept in sync with the .NET SDK version in global.json
     -->
-    <RuntimeFrameworkVersion>6.0.9</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>6.0.10</RuntimeFrameworkVersion>
```
RussKie, line 12: We shouldn't be changing this, as it completely messes the developer experience and requires a specific version installed locally. For example, I now have installed 6.0.403, and this settings forces me to have 6.0.402 installed.
This should stay at `6.0.0`, and only get updated on AppVeyor:
https://github.com/gitextensions/gitextensions/blob/4ee6417f4a7bb00e30e13ccef639f5e247b3aae0/appveyor.yml#L21-L22

global.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
     "sdk":  {
-                "version":  "6.0.401",
+                "version":  "6.0.402",
```
RussKie, line 3: Likewise, I don't think this should be changed either. This implies the lowest acceptable SDK version, as the next line directs to roll forward to the next available feature version, if available.

scripts/RepoLayout.props
gerhardol: Update in all local builds, so we use the expected in test builds?

RussKie: I am tired, and my brain is fried... and I don't understand the question, sorry.

Locally, we build with whatever SDK we have installed - 6.0.400 or 6.0.456. The runtime version depends on the installed SDK - 6.0.8 or 6.0.11...   Forcing the devs to install 6.0.10 is unnecessary.
When we do CI builds though, we will build against the latest available, and this runtime will be required for users to run the app.

gerhardol: I use the Debug builds in my normal usage. Then it would be good if the same runtime is used as a release build would require.
Could be a separate argument too.

vbjay: @gerhardol  if you want to run latest all you need to do is run the .\Scripts\Update-DotnetVersion.ps1 script just as appveyor does.  Then build as normal.  
image

image

image
image

You can even have a build script that is ignored that does this for you.

```ps

$path = Split-Path -Path $PSCommandPath -Parent
$update =Join-Path -Path $path -ChildPath Update-DotnetVersion.ps1

& $update

dotnet build --no-incremental --use-current-runtime

```
Something like 
image

You can play with --use-current-runtime or not.  But  using a git ignore 
image
You can then build your project with just a powershell script run. See scripts\Bisect-Test.ps1 for ideas on building and then running.

vbjay: You could even auto install the sdk and desktop with chocolatey.

```ps
choco upgrade dotnet-6.0-sdk -y
choco upgrade dotnet-6.0-desktopruntime -y
```

gerhardol: 'Can' is not the same as using the same setup as users by default.
I would like me and other pre-release users to have the same environment as users will have.
In a build and in a debugger is a slightly different situation.

RussKie: If a build comes from Appveyor than it'd target the latest available runtime. If a developer builds locally the build can target the installed runtime. There's generally no reason to force a developer to install the latest or specific version of runtime. And, if there's a compelling reason, e.g. a fixed bug or security vulnerability, then we'll bump the minimum required version in global.json.

PR #10339, opened 2022-11-05
SidePanel: AheadBehind null check

gerhardol:
## Proposed changes

aheadBehindData is null when this info is not shown and the grouping of the data fails.
For some reason I only see this in Debug builds, but it should go into master and 4.0.1 as well (if only to not see this when debugging).

Some comments from a late review comment
https://github.com/gitextensions/gitextensions/pull/10258#discussion_r1004968924

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/AheadBehindDataProvider.cs
+++ b/GitCommands/Git/AheadBehindDataProvider.cs
@@ -46,13 +46,14 @@ namespace GitCommands.Git
                 return null;
             }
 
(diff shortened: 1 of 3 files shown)


RussKie:
Is it possible to add a regression test?

gerhardol:
> Is it possible to add a regression test?

This is in the GUI part, not simple what I see.

RussKie (approved):

PR #10389, opened 2022-11-12
Add config to allow file protocol to be used-Tests

### LINKED ISSUE #10387: Adding file ( local folder) submodules broke since git version 2.38.1
### Environment

- Git Extensions 33.33.33
- Build 53714f1c
- Git 2.38.1.windows.1
- .NET 6.0.11
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 3.1.8 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 3.1.31 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 6.0.11 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

Running tests like GetSubmodulesLocalPaths fail since the file protocol is disabled by default as a security feature of git.  

```cmd
git submodule add C:/SomeOtherRepo OtherRepo
```
To use `submodule add` in the example above, run:

```cmd
git -c protocol.file.allow=always submodule add C:/SomeOtherRepo OtherRepo
```

https://git-scm.com/docs/git-config#Documentation/git-config.txt-protocolallow

https://git-scm.com/docs/git#Documentation/git.txt-codeGITPROTOCOLFROMUSERcode

## Cause and Details
- https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#fn-67904-1
- https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586 also affects git for windows

### Steps to reproduce

Run GetSubmodulesLocalPaths test with git version 2.38.1 or later installed. 

Before security of git altered, yes.  

### Diagnostics

_No response_

vbjay:
Fixes #10387 for tests.  Will reopen issue and tie to second pull request.
Closes #10394

## Proposed changes

- Add config to allow file protocol.

## Test methodology 

- Run tests and confirm tests succeed 

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using GitCommands.Config;
 using GitCommands.Utils;
 using GitExtUtils;
 using GitUIPluginInterfaces;
@@ -17,26 +18,36 @@ namespace GitCommands.Git.Commands
             };
         }
 
-        public static ArgumentString SubmoduleUpdateCmd(string? name)
+        public static ArgumentString SubmoduleUpdateCmd(string? name, IEnumerable<GitConfigItem>? configs = null)
         {
-            return SubmoduleUpdateCommand((name ?? "").Trim().QuoteNE());
+            return SubmoduleUpdateCommand((name ?? "").Trim().QuoteNE(), configs);
         }
 
-        public static ArgumentString SubmoduleUpdateCmd(IEnumerable<string> submodules)
+        public static ArgumentString SubmoduleUpdateCmd(IEnumerable<string> submodules, IEnumerable<GitConfigItem>? configs = null)
         {
             string submodulesQuoted = string.Join(" ", submodules.Select(s => s.Trim().QuoteNE()));
-            return SubmoduleUpdateCommand(submodulesQuoted);
+            return SubmoduleUpdateCommand(submodulesQuoted, configs);
         }
 
-        private static ArgumentString SubmoduleUpdateCommand(string name)
+        private static ArgumentString SubmoduleUpdateCommand(string name, IEnumerable<GitConfigItem>? configs)
         {
-            return new GitArgumentBuilder("submodule")
+            GitArgumentBuilder args = new("submodule")
             {
                 "update",
                 "--init",
                 "--recursive",
                 name
             };
+
+            if (configs is not null)
+            {
+                foreach (GitConfigItem cfg in configs)
+                {
+                    args.Add(cfg);
+                }
+            }
+
+            return args;
         }
 
         public static ArgumentString SubmoduleSyncCmd(string? name)
@@ -48,9 +59,9 @@ namespace GitCommands.Git.Commands
             };
         }
 
-        public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath, string branch, bool force)
+        public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath, string branch, bool force, IEnumerable<GitConfigItem> configs = null)
         {
-            return new GitArgumentBuilder("submodule")
+            GitArgumentBuilder argsBuilder = new("submodule")
             {
                 "add",
                 { force, "-f" },
@@ -58,6 +69,26 @@ namespace GitCommands.Git.Commands
                 remotePath.ToPosixPath().Quote(),
                 localPath.ToPosixPath().Quote()
             };
+
+            if (configs is not null)
+            {
+                foreach (GitConfigItem cfg in configs)
+                {
+                    argsBuilder.Add(cfg);
+                }
+            }
+
+            return argsBuilder;
+        }
+
+        /// <summary>
+        /// Gets <see cref="GitConfigItem"/> that sets 'protocol.file.allow' to always.
+        /// </summary>
+        /// <remarks>IEnumerable to allow future concat with other needed configs.</remarks>
+        /// <returns>Config with 'protocol.file.allow set to always</returns>
+        public static IEnumerable<GitConfigItem> GetAllowFileConfig()
+        {
+            yield return new GitConfigItem(SettingKeyString.AllowFileProtocol, "always");
         }
 
         public static ArgumentString GetCurrentChangesCmd(string? fileName, string? oldFileName, bool staged,
--- a/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
+++ b/UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
@@ -2,6 +2,7 @@
 using GitCommands.Git;
 using GitCommands.Git.Commands;
 using GitCommands.Utils;
+using GitExtUtils;
 using GitUIPluginInterfaces;
 using NUnit.Framework;
 using ResourceManager;
@@ -167,29 +168,36 @@ namespace GitCommandsTests.Git.Commands
         [Test]
         public void SubmoduleSyncCmd()
         {
-            Assert.AreEqual("submodule sync \"foo\"", GitCommandHelpers.SubmoduleSyncCmd("foo").Arguments);
-            Assert.AreEqual("submodule sync", GitCommandHelpers.SubmoduleSyncCmd("").Arguments);
-            Assert.AreEqual("submodule sync", GitCommandHelpers.SubmoduleSyncCmd(null).Arguments);
+            string config = "";
+            Assert.AreEqual($"{config}submodule sync \"foo\"", GitCommandHelpers.SubmoduleSyncCmd("foo").Arguments);
+            Assert.AreEqual($"{config}submodule sync", GitCommandHelpers.SubmoduleSyncCmd("").Arguments);
+            Assert.AreEqual($"{config}submodule sync", GitCommandHelpers.SubmoduleSyncCmd(null).Arguments);
         }
 
-        [Test]
-        public void AddSubmoduleCmd()
+        private static IEnumerable<TestCaseData> AddSubmoduleTestCases()
+        {
+            yield return new TestCaseData("", null);
+            yield return new TestCaseData("-c protocol.file.allow=always ", GitCommandHelpers.GetAllowFileConfig());
+        }
+
+        [Test, TestCaseSource(nameof(AddSubmoduleTestCases))]
+        public void AddSubmoduleCmd(string config, IEnumerable<GitConfigItem> configs)
         {
             Assert.AreEqual(
-                "submodule add -b \"branch\" \"remotepath\" \"localpath\"",
-                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", "branch", force: false).Arguments);
+                $"{config}submodule add -b \"branch\" \"remotepath\" \"localpath\"",
+                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", "branch", force: false, configs).Arguments);
 
             Assert.AreEqual(
-                "submodule add \"remotepath\" \"localpath\"",
-                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", branch: null, force: false).Arguments);
+                $"{config}submodule add \"remotepath\" \"localpath\"",
+                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", branch: null, force: false, configs).Arguments);
 
             Assert.AreEqual(
-                "submodule add -f -b \"branch\" \"remotepath\" \"localpath\"",
-                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", "branch", force: true).Arguments);
+                $"{config}submodule add -f -b \"branch\" \"remotepath\" \"localpath\"",
+                GitCommandHelpers.AddSubmoduleCmd("remotepath", "localpath", "branch", force: true, configs).Arguments);
 
             Assert.AreEqual(
-                "submodule add -f -b \"branch\" \"remote/path\" \"local/path\"",
-                GitCommandHelpers.AddSubmoduleCmd("remote\\path", "local\\path", "branch", force: true).Arguments);
+                $"{config}submodule add -f -b \"branch\" \"remote/path\" \"local/path\"",
+                GitCommandHelpers.AddSubmoduleCmd("remote\\path", "local\\path", "branch", force: true, configs).Arguments);
         }
 
         [Test]
--- a/UnitTests/GitCommands.Tests/GitModuleTests.cs
+++ b/UnitTests/GitCommands.Tests/GitModuleTests.cs
@@ -705,13 +705,14 @@ namespace GitCommandsTests
                     var child = moduleTestHelpers[i];
 
                     // Add child as submodule of parent
-                    parent.Module.GitExecutable.Execute(GitCommandHelpers.AddSubmoduleCmd(child.Module.WorkingDir.ToPosixPath(), $"repo{i}", null, true), throwOnErrorExit: false);
-                    parent.Module.GitExecutable.GetOutput(@"commit -am ""Add submodule""");
+                    parent.AddSubmodule(child, $"repo{i}");
                 }
 
                 // Init all modules of root
                 var root = moduleTestHelpers[0];
-                root.Module.GitExecutable.Execute(@"submodule update --init --recursive", throwOnErrorExit: false);
+                IEnumerable<GitConfigItem> cfgs = GitCommandHelpers.GetAllowFileConfig();
+
+                root.Module.GitExecutable.Execute(GitCommandHelpers.SubmoduleUpdateCmd(name: null, cfgs));
 
                 var paths = root.Module.GetSubmodulesLocalPaths(recursive: true);
                 Assert.AreEqual(new string[] { "repo1", "repo1/repo2", "repo1/repo2/repo3" }, paths, $"Modules: {string.Join(" ", paths)}");
--- a/UnitTests/GitCommands.Tests/GitModuleTests.cs
+++ b/UnitTests/GitCommands.Tests/GitModuleTests.cs
@@ -680,7 +680,6 @@ namespace GitCommandsTests
             Assert.AreEqual(status, stagedStatus);
         }
 
-        [Ignore("See https://github.com/gitextensions/gitextensions/issues/10387")]
         [Test]
         public void GetSubmodulesLocalPaths()
         {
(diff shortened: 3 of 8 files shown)


vbjay:
Looking into

  Modules: repo1 repo1/repo2
  Expected is <System.String[3]>, actual is <System.Collections.Generic.List`1[System.String]> with 2 elements
  Values differ at index [2]
  Missing:  < "repo1/repo2/repo3" >

vbjay:
Confirmed fixed both in test and adding a submodule in UI

image

Only thing left is to adjust config to allow the protocol on all submodule commands.  Looks like submodule update needs it too.

vbjay:
Last to check are the SubmoduleStatusProviderTests.

image

GitExtUtils/GitCommandConfiguration.cs
```diff
@@ -27,6 +27,7 @@ static GitCommandConfiguration()
             Default.Add(new GitConfigItem("diff.mnemonicprefix", "false"), "diff");
             Default.Add(new GitConfigItem("diff.ignoreSubmodules", "none"), "diff", "status");
             Default.Add(new GitConfigItem("core.safecrlf", "false"), "diff");
+            Default.Add(new GitConfigItem("protocol.file.allow", "always"), "submodule"); // git version 2.38.1 and later disabled file protocol by default
```
gerhardol, line 30: Should this be a popup for the user to confirm?
(Tests obviously needs this.)

UnitTests/GitCommands.Tests/Git/Commands/GitCommandHelpersTest.cs
```diff
@@ -167,28 +167,30 @@ public void TestUnsetStagedStatus()
...
+            string config = "-c protocol.file.allow=always";
+            Assert.AreEqual($"{config} submodule sync \"foo\"", GitCommandHelpers.SubmoduleSyncCmd("foo").Arguments);
+            Assert.AreEqual($"{config} submodule sync", GitCommandHelpers.SubmoduleSyncCmd("").Arguments);
+            Assert.AreEqual($"{config} submodule sync", GitCommandHelpers.SubmoduleSyncCmd(null).Arguments);
         }
 
         [Test]
         public void AddSubmoduleCmd()
         {
+            string config = "-c protocol.file.allow=always";
```
gerhardol, line 179: config in settings is related to submodule. Same in tests?

vbjay: The point is so if another con fig is needed, you just add to that config and you don't have to repeat the change per Assert.

vbjay: I am working on making tests green now .  Later I'll move to test only config but I wanted to confirm app  behaved and tests succeeded.  Then looking at always vs user and such mentioned in issue.

GitExtUtils/GitCommandConfiguration.cs
RussKie: I agree with @gerhardol, since this may be a security-related config, we shouldn't sneakily enable it. We should ask for an explicit permission from a user. Also, I think, it's a good idea to expose the setting under Git > Config.

RussKie: ...also it's good to check how well the app fares if the setting is set to `never`.

RussKie, line 30: Why not [`user`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-protocolallow)?

vbjay: https://github.com/gitextensions/gitextensions/pull/10389#discussion_r1020801678

vbjay:
Please hold up.  I am doing this methodically.  I just got tests green.  I am now going to play with what the different configs do.  I did make this Pull request WIP for a reason.

vbjay:
What user means:
user - protocol is only able to be used when GIT_PROTOCOL_FROM_USER is either unset or has a value of 1. This policy should be used when you want a protocol to be directly usable by the user but don’t want it used by commands which execute clone/fetch/push commands without user input, e.g. recursive submodule initialization.

found at https://git-scm.com/docs/git-config#Documentation/git-config.txt-protocolallow

So user basically breaks recursive submodule update --init.  

GetSubmodulesLocalPaths test:
image

```git -c protocol.file.allow=user submodule add -f "C:/Users/vbjay/AppData/Local/Temp/ymhgxebg.gsp/repo3/" "repo3"```
image

   at GitCommands.ExecutableExtensions.<ExecuteAsync>d__9.MoveNext() in C:\git\gitextensions\GitCommands\Git\ExecutableExtensions.cs:line 326

File: GitExtUtils/GitCommandConfiguration.cs (hunk 1 above)
Comment (vbjay) [reply GitExtUtils/GitCommandConfiguration.cs]: user is what is happening by default.  

https://git-scm.com/docs/git-config#Documentation/git-config.txt-protocolallow

**Comment by vbjay** [comment]
Ok.  now ready for review.

File: GitExtUtils/GitCommandConfiguration.cs (hunk 1 above)
Comment (vbjay) [reply GitExtUtils/GitCommandConfiguration.cs:30]: Without this change user is the default.
```git submodule add` and ```git submodule update --init --recursive``` fail for file protocol repo paths.  Wjat this does is break --recursive if not always is set.  Always is an opt in to say yes I mean it.  The user is initiating the command.  So it is ok.

Comment (vbjay) [reply GitExtUtils/GitCommandConfiguration.cs:30]: https://www.git-scm.com/docs/git#Documentation/git.txt-codeGITPROTOCOLFROMUSERcode

Comment (vbjay) [reply GitExtUtils/GitCommandConfiguration.cs:30]: So tests need this so I can move to the test config setup and I can work another branch to do a some kind of ui that says something like...

"...." command about to be run.  Do you trust it.  and if yes...add that to the options of that command.  I'll add a needs userTrust feature in a different pull request.  I'll change this pull request for just tests.

Comment (gerhardol) [reply GitExtUtils/GitCommandConfiguration.cs:30]: For sure change in tests.
I can agree that setting by default in commands is good for users, most will not understand the reason anyway and just find the popup annoying.
To follow that "GE is a GUI for Git" with no major deviation from default behavior, there should be a popup.

**Comment by vbjay** [comment]
image
image
image

Even setting default repo config fails the submodule add.  They require -c... 
```git submodule add -f "C:/Users/vbjay/AppData/Local/Temp/piibmzoi.jod/repo3/" "repo3"```
image

**Comment by vbjay** [comment]
This work will be the basis for another pull request that will define ui popup functionality based on needed ui interaction to allow command.

File: GitCommands/Git/Commands/GitCommandHelpers.cs [hunk 3]
```diff
@@ -48,16 +59,38 @@ public static ArgumentString SubmoduleSyncCmd(string? name)
             };
         }
 
-        public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath, string branch, bool force)
+        public static ArgumentString AddSubmoduleCmd(string remotePath, string localPath, string branch, bool force, IEnumerable<GitConfigItem> configs = null)
         {
-            return new GitArgumentBuilder("submodule")
+            ArgumentBuilder cfgArgs = new();
```
gerhardol, line 64: unused

gerhardol (commented):
too tired to review...

GitCommands/Git/Commands/GitCommandHelpers.cs
vbjay: image
image

?

I am planning on the second pull request to call it more.  Please pull the code down and use CodeLens.

vbjay: This pull request is just for making tests green.  Another will be created to handle actual application functionality.

RussKie:
Can you please re-enable all tests that were disabled in https://github.com/gitextensions/gitextensions/pull/10394/commits/fe1d5b2d56d634f5d265520fc7eb14e623e09cb4?

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -57,6 +58,18 @@ public void SetUp()
             _repo2Module = submodules.ElementAt(0);
             _repo3Module = submodules.ElementAt(1);
 
+            GitModule[] actualModdules = new GitModule[]
```
gerhardol, line 61:
```suggestion
            GitModule[] actualModules = new GitModule[]
```

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -17,26 +18,36 @@ public static ArgumentString CherryPickCmd(ObjectId commitId, bool commit, strin
             };
         }
 
-        public static ArgumentString SubmoduleUpdateCmd(string? name)
+        public static ArgumentString SubmoduleUpdateCmd(string? name, IEnumerable<GitConfigItem> configs = null)
```
gerhardol, line 21: ?
```suggestion
        public static ArgumentString SubmoduleUpdateCmd(string? name, IEnumerable<GitConfigItem>? configs = null)
```

gerhardol (commented):
+1
Enable tests as Russkie suggested.
Looks fine, not much else to do I assume.
I would would prefer a string instead of the nullable configs argument in GitCommandHelpers.cs, no strong opinion.

GitCommands/Git/Commands/GitCommandHelpers.cs
vbjay: I am doing the innumerable on purpose. It is so 5fat if other configs are needed in future it's as simple as concat configs. The plan is to pass down these from allow from user in next. Adjusting ab innumerable to allow building needed configs is a lot better than passing strings around.

vbjay:
Approval and merge requested. I fixed up small fixes in review and I am doing the IEnumerable of config items on purpose. That will allow the configs next pull request that builds user permission allowing to run will pass in this config and others as needed.

RussKie:
I'm away. I'll be able to review next week.

GitCommands/Git/Commands/GitCommandHelpers.cs
gerhardol: In this case it was the nullability I commented on.
(I would actually preferred a string here, I see no reason to prepare for multiple settings now. I had to think a few seconds when seeing it and prefer the simple solution. But it was not enough to comment on it in the first place or to request changes.)

UnitTests/GitCommands.Tests/GitModuleTests.cs
```diff
@@ -679,18 +680,18 @@ public void GetStagedStatus(ObjectId firstRevision, ObjectId secondRevision, Obj
...
+            List<GitModuleTestHelper> moduleTestHelpers = new();
             try
             {
                 const int numModules = 4;
 
                 for (int i = 0; i < numModules; ++i)
                 {
-                    moduleTestHelpers.Add(new CommonTestUtils.GitModuleTestHelper($"repo{i}"));
+                    moduleTestHelpers.Add(new GitModuleTestHelper($"repo{i}"));
+                    Debug.WriteLine($"Repo[{i}]:{moduleTestHelpers[i].TemporaryPath}");
```
gerhardol, line 694: Should this printout be kept?
(Setup as for [SubmoduleStatusProviderTests.cs](https://github.com/gitextensions/gitextensions/pull/10389/files#diff-e7dde1041edc7df2802b2896fac5e4004a158dfedc9334f60582133dfefa4c54) is a better fit.)

gerhardol (approved):

UnitTests/GitCommands.Tests/GitModuleTests.cs
vbjay: Yes. It is there to help debugging tests in future. I did it to help track down issues when testing and capturing what folders are involved. I did debug to make sure it was only in debugging.

UnitTests/GitCommands.Tests/Submodules/SubmoduleStatusProviderTests.cs
```diff
@@ -57,6 +58,18 @@ public void SetUp()
...
             _repo3Module = submodules.ElementAt(1);
 
+            GitModule[] actualModules = new GitModule[]
+            {
+                _repo1Module,
+                _repo2Module,
+                _repo3Module
+            };
+
+            for (int i = 0; i < actualModules.Length; i++)
```
mstv, line 68: This belongs together IMO because `actualModules` is used here only.
```suggestion
            };
            for (int i = 0; i < actualModules.Length; i++)
```

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -48,16 +59,36 @@ public static ArgumentString SubmoduleSyncCmd(string? name)
...
+        }
+
+        /// <summary>
+        /// Gets <see cref="GitConfigItem"/> that sets 'protocol.file.allow' to always.
+        /// </summary>
+        /// <remarks>IEnumerable to allow future concat with other needed configs.</remarks>
+        /// <returns>Config with 'protocol.file.allow set to always</returns>
+        public static IEnumerable<GitConfigItem> GetAllowFileConfig()
+        {
+            yield return new GitConfigItem(SettingKeyString.AllowFileProtocol, "always");
```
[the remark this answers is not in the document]
vbjay: Please read all the linked reference material in issue and all my comments in pull request. I did a pretty good job of documenting the issue and what the config does and even using in a config file doesn't solve the problem.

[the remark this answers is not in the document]
vbjay: My understanding is if config is added in version x, then <x ignores config.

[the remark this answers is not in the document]
mstv: What if you specify it explicitly on the command line?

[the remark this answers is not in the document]
vbjay: I have done a ton of testing. If you doubt the work pull it down and rub your own tests to verify. I have already hit it hard and validated the fix.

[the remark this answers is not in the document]
vbjay: This will lay the foundation for a used allows blaa functionality down the road.  This just deals with getting tests green. Run the noted tests before this work on 2.38.1 git version. Take off the error swallowing and see the problem yourself.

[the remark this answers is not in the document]
vbjay: And I mean actually debug a test and not just run a test.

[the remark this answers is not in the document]
vbjay: https://github.com/gitextensions/gitextensions/pull/10389#issuecomment-1312813174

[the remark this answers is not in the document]
vbjay: Any git config that is used that is not looked for is ignored.  You could setup 

```
git config mstv.seceret ssshhhh
````
and git would ignore because it does nothing for git.  Same principle for a config value that is looked for in later versions and not older.  So before this config was added, git is going to say so what.  I don't care.  When it looks for it, command line is just another level of config that is lower in hierarchy of all the config files like --system, --global, --worktree, and --local.  So command line is the lowest and overrides all config file values when passed in.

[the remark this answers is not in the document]
vbjay: https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586

vbjay:
See https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586 for details.  I have tried with never, and user.  Always is required and at least with git for windows even setting config in ```git config protocol.file.allow always``` fails this as shown in the AddSubmodule code found at 

image

I left this code in to make sure that adding submodules still works down the road.

GitCommands/Git/Commands/GitCommandHelpers.cs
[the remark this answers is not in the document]
mstv: My concern was: "What happens if an older git version is installed and such options are added to the above submodule commands?"
I had not noticed that `GetAllowFileConfig()` is used by tests only. It would have been clearer if this function was not be located in application code.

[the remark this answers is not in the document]
mstv: For the test execution we can demand a recent git version, i.e. nothing to do - perhaps except moving the function to a test helper.

mstv (approved):
Thank you very much for making the tests work again and for all the testing.

GitCommands/Git/Commands/GitCommandHelpers.cs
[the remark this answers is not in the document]
vbjay: It will be used down the road in program too.  This is setting up plumbing.  The gui will pass down stuff and that was a plumbing feature.  The how is the config being passed in.  The when it is passed in and whatnot will be  determined by gui flow logic in next pull request.

mstv:
I tend to `rebase and merge` this PR in order to keep the commits (although some are tiny). This merge strategy has the disadvantage that no reference to this PR is added.

vbjay:
I can rebase and force push if you want.

vbjay:
Should be able to just merge now.

mstv:
> > I tend to rebase and merge this PR in order to keep the commits (although some are tiny). This merge strategy has the disadvantage that no reference to this PR is added.

> I can rebase and force push if you want.
> Should be able to just merge now. @mstv

There is no need to manually rebase. This can be done using grafik.

We want to avoid unnecessary merge commits. I want to discuss
- whether we should keep such commits
- how to merge including a reference to the PR

vbjay:
> whether we should keep such commits
I follow the idea of atomic commits.  I try to keep changes as small as possible so that reviews are easier.  If you squash then it gets hairy understanding work like this.  The add config is pretty big but is clear on what it is doing.  So I prefer commits stay.  

>how to merge including a reference to the PR
I can reword last commit and add a reference to pull request if you like.

mstv:
> I can reword last commit and add a reference to pull request if you like.

Yes, please.

We can continue the generic discussion after merging.

vbjay:
Next step is to get the core team to opine on #10426.  This pull request started the plumbing on how to pass in the config when needed.  The discussion is there for how do we want the program to act.  I don't want to go down a rabbit hole and have to rework. Please discuss there and let's get a direction on the discussion and then from discussion I'll wire up the expected functionality in the gui.

mstv:
Thank you

PR #10396, opened 2022-11-13
Fix 'Delete obsolete branches' plugin

### LINKED ISSUE #10347: "Delete obsolete branches" plugin stop working in version 4
"Delete obsolete branches" plugin stop working in version 4

drewnoakes:
Fixes #10347

When adding nullable reference types to this code, an incorrect assertion was added. The nullability of the cancellation object here is quite confusing, as setting IsRefreshing to true causes it to be created. It's also known as not-null if IsRefreshing is true. This commit updates the assertions to be correct, fixing the bug.

### DIFF
--- a/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
+++ b/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
@@ -278,10 +278,9 @@ namespace GitExtensions.Plugins.DeleteUnusedBranches
 
         private async Task RefreshObsoleteBranchesAsync()
         {
-            Validates.NotNull(_refreshCancellation);
-
             if (IsRefreshing)
             {
+                Validates.NotNull(_refreshCancellation);
                 _refreshCancellation.Cancel();
                 IsRefreshing = false;
                 return;
@@ -290,6 +289,8 @@ namespace GitExtensions.Plugins.DeleteUnusedBranches
             Validates.NotNull(_gitUiCommands);
 
             IsRefreshing = true;
+            Validates.NotNull(_refreshCancellation);
+
             var curBranch = _gitUiCommands.GitModule.GetSelectedBranch();
             RefreshContext context = new(
                 _gitCommands,
--- a/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
+++ b/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
@@ -288,6 +288,7 @@ namespace GitExtensions.Plugins.DeleteUnusedBranches
 
             Validates.NotNull(_gitUiCommands);
 
+            // IsRefreshing will set _refreshCancellation
             IsRefreshing = true;
             Validates.NotNull(_refreshCancellation);
 


Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.cs
```diff
@@ -290,6 +289,8 @@ private async Task RefreshObsoleteBranchesAsync()
             Validates.NotNull(_gitUiCommands);
 
             IsRefreshing = true;
```
gerhardol, line 291:
```suggestion
            // IsRefreshing will set _refreshCancellation
            IsRefreshing = true;
```

gerhardol (approved):
Should `IsRefreshing = false` cancel the token?

RussKie:
@drewnoakes can you please respond so we can merge this? Thanks

drewnoakes:
> Should `IsRefreshing = false` cancel the token?

Potentially, though I don't have the bandwidth to investigate at the moment. The change in this PR will restore the pre-NRT behaviour.

gerhardol (approved):

drewnoakes:
@RussKie @gerhardol I don't remember the merge rules for this repo so will let one of you push the merge button. Thanks.

mstv (approved):

gerhardol:
Thanks
Squashed, wrapped the commit message, set milestone to 4.0.1, pushed to release/4.0

RussKie:
Thank you all

carlin-q-scott:
This issue does not seem to have been fixed in 4.0.1. Multiple people, including myself, are still experiencing the behavior described in #10619

pazoozooCH:
Can confirm that this is not really fixed in *4.0.2.16100*. UI hangs after clicking "Delete" with a "Loading..." spinner. When I close the dialog, it seems that 1 branch has been deleted...

What's the process here? Reopen the issue? Or should I create a new one?

vbjay:
1. Make sure if using portable you have extracted to a clean folder and then copied settings over. So not me the different release portable versions. 
1. Create a new issue and make sure you use help->about menu and copy button to paste version info. 
1. Reference this issue in new issue.

pmiossec:
> Can confirm that this is not really fixed in _4.0.2.16100_. UI hangs after clicking "Delete" with a "Loading..." spinner. When I close the dialog, it seems that 1 branch has been deleted...
> 
> What's the process here? Reopen the issue? Or should I create a new one?

@pazoozooCH If fact, once GitExtensions get the list of branches to delete, it launches a command to delete the first one and loop to delete all.
But if there is an exception or if git doesn't return a good exit code, the handling is not perfect and GE display  the "loading' spinner until user close the popup.

So, it would be good if you can have a look at the exit code of the command run (that you could get when 'Git command log' F12 shortcut).
If you could run a similar command on the next branch GE would have deleted and report the output, it would be very helpful. Because I have the feeling that's specific to your configuration and we should find the reason...

pazoozooCH:
> > Can confirm that this is not really fixed in _4.0.2.16100_. UI hangs after clicking "Delete" with a "Loading..." spinner. When I close the dialog, it seems that 1 branch has been deleted...
> > What's the process here? Reopen the issue? Or should I create a new one?
> 
> @pazoozooCH If fact, once GitExtensions get the list of branches to delete, it launches a command to delete the first one and loop to delete all. But if there is an exception or if git doesn't return a good exit code, the handling is not perfect and GE display the "loading' spinner until user close the popup.
> 
> So, it would be good if you can have a look at the exit code of the command run (that you could get when 'Git command log' F12 shortcut). If you could run a similar command on the next branch GE would have deleted and report the output, it would be very helpful. Because I have the feeling that's specific to your configuration and we should find the reason...

Hi @pmioss

Thanks for the feedback. It seems that GitExtensions successfully deletes the first branch, then immediately tries do delete it once more for whatever reason which obviously fails with an Exception:

2023-04-11 GitExtensions Delete Obsolete Branches Log

This results in the first selected branch being deleted and then hanging with the loading spinner...

I could reproduce this easily with a new, empty repo:

1. Create new repo
2. Make three commits, c1, c2, c3 (I first tried with two commits, but obviously there's another bug when a branch is set to the initial commit of a repo which would cause the "Delete obsolete branches" dialog to fail during initial searching - doesn't bother me too much because I usually don't have branches on the initial commit)
3. Set main to c3
4. Create two branches (branch1, branch2) on c2
5. Run "delete obsolete branches"
6. Select both branches (branch1 and branch2)
7. Git Extensions will try to run `git branch -d branch1` twice, the second time failing with an `ExternalOperationException: error: branch 'branch1' not found.

Cheers Martin

pmiossec:
Your investigation remind me that I fixed the issue in #10663 but it has to be released in the v4.1

pazoozooCH:
> Your investigation remind me that I fixed the issue in #10663 but it has to be released in the v4.1

Thanks for the info, looking forward to *v4.1* then 🙏

PR #10400, opened 2022-11-14
Improve recent repository list keyboard navigation

mdonatas:
Fixes _(none)_

## Proposed changes

- Select the first item when recent repository list gets focus
- Give focus to recent repository list when pressing ⬇️ arrow key in a filter search box
- Give focus to filter search box when pressing ⬆️ in a recent repository list when the top item (in any column) is selected

### Before

Notice how when pressing tab while in a filter search box it's not obvious which control gets focus
recent-repo-arrow-no-keys

### After

Full back and forth navigation between the list and the search box with arrow keys
recent-repo-arrow-keys

## Test methodology 

- Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -691,18 +691,15 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
 
         private void listView1_KeyDown(object sender, KeyEventArgs e)
         {
-            if (e.KeyCode == Keys.Up)
+            if (e.KeyCode == Keys.Up && listView1.SelectedItems.Count > 0)
             {
-                if (listView1.SelectedItems.Count > 0)
+                // Compare current item to the very first item to see if it's at the top
+                ListViewItem selectedItem = listView1.SelectedItems[0];
+                if (selectedItem.Bounds.Y == listView1.Items[0].Bounds.Y)
                 {
-                    // Compare current item to the very first item to see if it's at the top
-                    ListViewItem selectedItem = listView1.SelectedItems[0];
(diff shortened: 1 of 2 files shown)


GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
```diff
@@ -677,6 +681,32 @@ private void listView1_MouseLeave(object sender, EventArgs e)
...
+                listView1.Items[0].Selected = true;
+            }
+        }
+
+        private void listView1_KeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.KeyCode == Keys.Up)
+            {
+                if (listView1.SelectedItems.Count > 0)
+                {
```
RussKie, line 697: Could you please merge these two and invert to reduce nesting?

RussKie (commented):
👍

RussKie (approved):

GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
RussKie: > p.s. I know that guard clauses is all the rage right now [youtube.com/shorts/Zmx0Ou5TNJs](https://www.youtube.com/shorts/Zmx0Ou5TNJs) (at least judging from my youtube recommendations) but I think context matters 😊

Now? Haha, more like in the past decade. Since people stopped following advice from a fellow named Uncle Bob (some of which I find rather questionable). There was a time when it was believed that a method must only have a single entry and a single return - .NET Framework code was written under those guidelines. It was rumoured that it yielded a better optimisation. Even if such anecdotes were true, these days it's immaterial, and code maintenability, readability and testability are significantly more important.

RussKie:
Thank you

PR #10402, opened 2022-11-14
Simplify devex

RussKie:
* Allow developers to use any SDK equal or above the specified.
* Do not demand a specific runtime during dev cycle
* Only demand the min runtime for AppVeyor builds

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ environment:
   # Disable the .NET first time experience to skip caching NuGet packages and speed up the build.
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
   # Use latest version of current used .NET version.  6.0 could use the latest of 6.0.  See .\Scripts\Update-DotnetVersion.ps1
-  GE_USE_LATEST_DOTNET: false
+  GE_USE_LATEST_DOTNET: true
 
 init:
   - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 ﻿{
-    "sdk":  {
-                "version":  "6.0.402",
-                "rollForward":  "feature"
-            }
+    "sdk": {
+        "version": "6.0.400",
+        "rollForward": "feature"
+    }
 }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 ﻿{
     "sdk": {
-        "version": "6.0.400",
+        "version": "6.0.401",
         "rollForward": "feature"
     }
 }
--- a/scripts/RepoLayout.props
+++ b/scripts/RepoLayout.props
@@ -9,7 +9,7 @@
     <!-- Set version for runtimeconfig.json, used by DotnetRuntimeBootstrapper for minimal version to support
          To be kept in sync with the .NET SDK version in global.json
     -->
-    <RuntimeFrameworkVersion>6.0.10</RuntimeFrameworkVersion>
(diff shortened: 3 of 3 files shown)


vbjay:
Just an FYI

This will also allow you to bump to later major version of .net by adjusting the 
image

You can locally also run the script adding the version

```ps
.\scripts\UpdateDotnetVersion.ps1 7.0
```
Along with a local script that runs the update script, keeps sdk and desktopruntime up to date and then builds.  you can mimic this locally by saving a script in the scripts folder.  Like below example.  


```ps
choco upgrade dotnet-6.0-sdk -y
choco upgrade dotnet-6.0-desktopruntime -y

$path = Split-Path -Path $PSCommandPath -Parent
$update =Join-Path -Path $path -ChildPath Update-DotnetVersion.ps1

& $update

dotnet build --no-incremental

```

Could play with --use-current-runtime and other values with dotnet build.

global.json
```diff
@@ -1,6 +1,6 @@
...
-    "sdk":  {
-                "version":  "6.0.402",
-                "rollForward":  "feature"
-            }
-}
+    "sdk": {
+        "version": "6.0.400",
+        "rollForward": "feature"
+    }
+}
```
mstv, line 6: Please keep the formatting of the Jay's script.

appveyor.yml
```diff
@@ -18,7 +18,7 @@ environment:
   # Disable the .NET first time experience to skip caching NuGet packages and speed up the build.
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
   # Use latest version of current used .NET version.  6.0 could use the latest of 6.0.  See .\Scripts\Update-DotnetVersion.ps1
-  GE_USE_LATEST_DOTNET: false
+  GE_USE_LATEST_DOTNET: true
```
mstv, line 21: I guess this doesn't match the PR description.

global.json
mstv: IIRC, I was forced to update the SDK although 6.0.402 and 6.0.401 should be compatible to 6.0.400.

mstv (commented):
I appreciate the move to reduce forced updates and hope it works.
Though I would keep the 6.0.9 runtime as minimum as requested by v4.0.

appveyor.yml
vbjay: Yes it does. It eliminates us managing and allows the build server to build with 6.0/latest.

global.json
vbjay: Nope 401 = 6.0.9 runtime. 402 =6.0.10. Yes you can build 6.0.9 with 402 but I would keep runtime and sdk matched. So what he is doing does the work.

appveyor.yml
mstv: Reading it again, yes, the description is not incorrect.
But that it always builds the with the latest, should be clearly stated in the issue description.

vbjay: Yep. I don't like the formatting but if people do use script it will prevent a big dif. I would love to adjust script to make it pretty but all the ways to do that looked hacky and I didn't like the ways. So I chose to just let script be authority of formatting.

global.json
vbjay: Yep. I don't like the formatting but if people do use script it will prevent a big dif. I would love to adjust script to make it pretty but all the ways to do that looked hacky and I didn't like the ways. So I chose to just let script be authority of formatting.

RussKie: Since we shouldn't be changing this file during the dev cycle, and only change it on a build agent, there is no need to keep the weird formatting.

RussKie:
> Though I would keep the 6.0.9 runtime as minimum as requested by v4.0.

Updated.

PR #10421, opened 2022-11-17
Add buttons for un/registration of the Windows Explorer integration

### LINKED ISSUE #7982: Add "Remove Windows Explorer integration" in Settings Menu

mstv:
Fixes #7982
Finishes the abandoned #9701

## Proposed changes

- Factor out `ShellExtensionManager` (cherry-pick from #9701, kudos to @gassanosh)
- Factor out `MessageBoxes.ShowShellExtensionRegistrationError`
- Add `Un/UnregisterButton` to `ShellExtensionSettingsPage`

### Before

grafik

### After

grafik

Exception dialog (in this case, I canceled the privilege elevation)

grafik

## Test methodology 

- manual (both msi and portable build)

## Test environment(s) 

- Git Extensions 33.33.33
- Build 4e92a6ea9e592be86573a44398453b67d933d165
- Git 2.38.1.windows.1
- .NET 6.0.10
- DPI 96dpi (no scaling)

```
```

## Merge strategy

Perhaps we should keep the three commits.

----


GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
```diff
@@ -221,54 +219,13 @@ private void GitBinFound_Click(object sender, EventArgs e)
...
-                catch (System.ComponentModel.Win32Exception)
-                {
-                    // User cancel operation, continue;
-                }
-            }
-            else
+            catch (Exception ex)
             {
-                MessageBox.Show(this, string.Format(_cantRegisterShellExtension.Text, CommonLogic.GitExtensionsShellEx32Name), TranslatedStrings.Error, MessageBoxButtons.OK, MessageBoxIcon.Error);
+                MessageBoxes.ShowShellExtensionRegistrationError(ParentForm, ex);
```
RussKie, line 228: Would could probably use UserExternal exception and remove the custom message box.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -0,0 +1,115 @@
...
+        public static void Unregister() => RunRegSvrForShellExtensionDlls("/u {0}");
+
+        private static void RunRegSvrForShellExtensionDlls(string argumentsPattern)
+        {
+            if (Environment.Is64BitOperatingSystem)
+            {
+                RunRegSvrForSingleDll(CommonLogic.GitExtensionsShellEx64Name, argumentsPattern);
+            }
+
+            RunRegSvrForSingleDll(CommonLogic.GitExtensionsShellEx32Name, argumentsPattern);
```
RussKie, line 60: I'd wrap this in try...catch and throw UseExternal exception. This will allow us not to catch exceptions at callsites.

pmiossec:
Just FYI, I have some commits to be able to remove some items from the contextual menu and with some small settings page changes.

It looks like this:
contextual_menu
image

So you can theoretically end up with the same state with nothing displayed in the menu if you uncheck all.
But your solution to handle this specific case is better.

@mstv I don't think the 2 features are incompatible. What do you think?
Once this one is merged, I could rebase on it and create a PR if it's OK.

mstv:
> Just FYI, I have some commits to be able to remove some items from the contextual menu and with some small settings page changes.

I had seen your branch.

> So you can theoretically end up with the same state with nothing displayed in the menu if you uncheck all. But your solution to handle this specific case is better.
 
> @mstv I don't think the 2 features are incompatible. What do you think? Once this one is merged, I could rebase on it and create a PR if it's OK.

Yes, the combination will give the users full control, especially which portable installation shall be used.

There is one restriction though. The DLL remains locked after unregistering because it is still loaded by an executable (Explorer and / or TotalCommander) until termination.

GitUI/CommandsDialogs/SettingsDialog/Pages/ChecklistSettingsPage.cs
mstv: Done and updated the PR description.

RussKie:
Thank you @mstv. I tried the branch locally, and I'm getting the following error trying to disable the extensions:
image

RussKie:
I also don't think we need two buttons; I think one button should suffice.

mstv:
> I tried the branch locally, and I'm getting the following error trying to disable the extensions:
>
> image

There has been a similar bug report recently. (I must have missed it in the search results.) Could you debug it, please? It works for me on a few (Win10) machines.

---

> I also don't think we need two buttons; I think one button should suffice.

The check for the registration state might be too specific. In order to be able to get rid of leftovers by trying again, I have kept the `Unregister` button enabled always:

unregistered:

unregistered

registered:

grafik

RussKie:
> Could you debug it, please? It works for me on a few (Win10) machines.

This looks like the root cause - lack of quotes:
image

RussKie:
I also wonder if we should pass [`/s` switch](https://learn.microsoft.com/windows-server/administration/windows-commands/regsvr32) to avoid popups.

GitUI/CommandsDialogs/SettingsDialog/ShellExtension/ShellExtensionManager.cs
```diff
@@ -72,7 +72,7 @@ static void RunRegSvrForSingleDll(string dllName, string argumentsPattern)
                         throw new FileNotFoundException(null, dllName);
                     }
 
-                    string arguments = string.Format(argumentsPattern, path);
+                    string arguments = string.Format(argumentsPattern, path.Quote());
```
mstv, line 75: Sorry, this got lost when I squashed the commits.

mstv:
> I also wonder if we should pass [`/s` switch](https://learn.microsoft.com/windows-server/administration/windows-commands/regsvr32) to avoid popups. And show an error message for [anything but exit code 0](https://devblogs.microsoft.com/oldnewthing/20180920-00/?p=99785).

Yes, I am going to add these.

RussKie (approved):

mstv:
I am going to merge this tomorrow.

PR #10431, opened 2022-11-21
Persist Current Branch setting

### LINKED ISSUE #10424: Show current branch only; Not persistent
### Environment

- Git Extensions 4.0.0.15569
- Git 2.38.1.windows.1
- .NET 6.0.11
- DPI 96dpi (no scaling)

### Issue description

View menu
Show current branch only (Ctl-Shift-U) Not persistent.
Always defaults to Show all branches (Ctl-Shift-A) when opening repository.
Worked fin in previous versions.

### Steps to reproduce

Set "Show current branch only" (Ctrl-Shift-U)
Close repository.
Open repository.
All branches being shown.

Yes in all recent previous versions (3....)

### Diagnostics

_No response_

gerhardol:
Fixes #10424

## Proposed changes

Revert change to reset to All when changing repo

Note: This could be replaced with resetting Reflog->All instead, to be consistent
(Filtered branches should always be reset, more or less quering the repo twice needed to find if the filtered branches are available in the new repo.

Tests are not adapted - there are a lot of changes needed.

For 4.0.1 too

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -199,7 +199,7 @@ namespace GitUI.UserControls.RevisionGrid
 
         /// <summary>
         /// Disables all active filters.
-        /// Reflog is not disabled.
+        /// CurrentBranch and Reflog are not disabled.
         /// FullHistory and SimplifyMerges are considered settings and not reset.
         /// </summary>
         public void ResetAllFilters()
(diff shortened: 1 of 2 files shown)


mstv (approved):

RussKie:
+0
I'm not going to actively oppose this, since this restore the original behaviour, and some users used to this. But I think the move to not persist this setting - was correct.

mstv (approved):

gerhardol:
> +0 I'm not going to actively oppose this, since this restore the original behaviour, and some users used to this. But I think the move to not persist this setting - was correct.

Other filters not reset has caused support issues, so that is a good change (even if there are users that may not agree).
For the major "mode" current/reflog I am split, I am not convinced either. 
No other major comments (than the issue), so I am merging this.

PR #10434, opened 2022-11-22
Do not add BOM when commitEncoding is set to upcase UTF-8

orgads:
commitEncoding is case-insensitive, and the predefined UTF8 encoding is using lowercase key utf-8. Convert to lowercase to guarantee that it matches.

## Test methodology 
- `git config i18n.commitEncoding UTF-8`
- Commit using Git Extensions
- `git show` shows a leading space before the commit subject.

## Test environment(s) 

- GIT 2.38.1
- Windows 10

----

### DIFF
--- a/GitCommands/Settings/ConfigFileSettings.cs
+++ b/GitCommands/Settings/ConfigFileSettings.cs
@@ -136,7 +136,12 @@ namespace GitCommands.Settings
 
         private Encoding? GetEncoding(string settingName)
         {
-            string encodingName = GetValue(settingName);
+            // The keys in AppSettings.AvailableEncodings are lower-case, and the actual
+            // configuration is case-insensitive, and can be configured uppercase on the
+            // command line. If configured as UTF-8, then if we don't use the predefined
+            // encoding, it will use the default, which adds BOM.
+            // Convert it to lowercase, to ensure matching.
+            string encodingName = GetValue(settingName).ToLowerInvariant();
 
             if (string.IsNullOrEmpty(encodingName))
             {
--- a/UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
+++ b/UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
@@ -21,7 +21,7 @@ namespace GitCommandsTests
         private ReferenceRepository _referenceRepository;
 
         private readonly string _workingDirGitDir = @"c:\dev\repo\.git";
-        private readonly Encoding _encoding = Encoding.UTF8;
+        private readonly Encoding _encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
 
         private readonly string _amendSaveStatePath;
         private readonly string _commitMessagePath;
@@ -39,6 +39,12 @@ namespace GitCommandsTests
             _commitMessagePath = Path.Combine(_workingDirGitDir, "COMMITMESSAGE");
             _mergeMessagePath = Path.Combine(_workingDirGitDir, "MERGE_MSG");
             _rememberAmendCommitState = AppSettings.RememberAmendCommitState;
+        }
+
+        [OneTimeSetUp]
+        public void Init()
+        {
+            AppSettings.LoadSettings();
         }
 
         [SetUp]
@@ -212,6 +218,24 @@ namespace GitCommandsTests
             manager.WriteCommitMessageToFile(message, CommitMessageType.Normal, false, false);
 
             File.Exists(manager.CommitMessagePath).Should().BeTrue();
+        }
+
+        [TestCase("utf-8")]
+        [TestCase("Utf-8")]
+        [TestCase("UTF-8")]
+        public void WriteCommitMessageToFile_no_bom(string encodingName)
+        {
+            _referenceRepository.Module.EffectiveConfigFile.SetString("i18n.commitEncoding", encodingName);
+            _referenceRepository.Module.CommitEncoding.Preamble.Length.Should().Be(0);
+            CommitMessageManager manager = new(_referenceRepository.Module.WorkingDir, _referenceRepository.Module.CommitEncoding);
+
+            File.Exists(manager.CommitMessagePath).Should().BeFalse();
+
+            string message = "Test message";
+            manager.WriteCommitMessageToFile(message, CommitMessageType.Normal, false, false);
+
+            File.Exists(manager.CommitMessagePath).Should().BeTrue();
+            File.ReadAllBytes(manager.CommitMessagePath).Should().BeEquivalentTo(Encoding.ASCII.GetBytes(message + "\r\n"));
         }
 
         [Test]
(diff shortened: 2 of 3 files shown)


GitCommands/Settings/ConfigFileSettings.cs
```diff
@@ -136,7 +136,7 @@ public Encoding FilesEncoding
 
         private Encoding? GetEncoding(string settingName)
         {
-            string encodingName = GetValue(settingName);
+            string encodingName = GetValue(settingName).ToLower();
```
RussKie, line 139: Could you please elaborate how this is addressing the issue?

orgads: I wrote it in the commit message.

Suppose the setting is UTF-8. Here it is converted to lower case, so `encodingName` is `utf-8`. This value exists in `AppSettings.AvailableEncodings`, and it is mapped to BOM-less UTF-8.

mstv:
Have you tried to configure the encodings which shall be used in GE? I would expect that the encodings will match exactly then.

grafik

grafik

orgads:
Yes, this works, but if the encoding was configured *by command line* as upcase, it doesn't match.

Do you have a reason*not* to have this extra layer of protection?

GitCommands/Settings/ConfigFileSettings.cs
RussKie: Thank you for the clarification. Please add a comment to the code explaining this, so that this change won't get undone by accident.

RussKie, line 139:
```suggestion
            string encodingName = GetValue(settingName).ToLowerInvariant();
```

RussKie (commented):
It would be great to add few tests to verify the issue is fixed and won't regress in the future.
A test would create a test repo, configure `i18n.commitEncoding` with upper/lower/mixed case and verify the encoding is expected.

We have few similar tests, e.g., https://github.com/gitextensions/gitextensions/blob/master/UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs

GitCommands/Settings/ConfigFileSettings.cs
orgads: Done

orgads:
> It would be great to add few tests to verify the issue is fixed and won't regress in the future. A test would create a test repo, configure `i18n.commitEncoding` with upper/lower/mixed case and verify the encoding is expected.
> 
> We have few similar tests, e.g., https://github.com/gitextensions/gitextensions/blob/master/UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs

Done. Not sure if it's the right place for them though :)

GitCommands/Settings/ConfigFileSettings.cs
orgads: Done

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
```diff
@@ -392,6 +392,17 @@ public void ResetCommitMessage()
...
+        [Test, TestCaseSource(nameof(EncodingNames))]
+        public void MergeOrCommitMessage_should_use_correct_encoding_when_name_is_upcase(
+            string encodingName)
+        {
+            AppSettings.LoadSettings();
+            _referenceRepository.Module.EffectiveConfigFile.SetString("i18n.commitEncoding", encodingName);
+            _referenceRepository.Module.CommitEncoding.Preamble.Length.Should().Be(0);
+        }
+
+        private static IEnumerable<string> EncodingNames => new List<string> { "utf-8", "Utf-8", "UTF-8" };
```
gerhardol, line 404: Add to the beginning of the file.

gerhardol (commented):
Maybe better in GitModuleTests.cs

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
orgads: Done

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
```diff
@@ -21,7 +21,7 @@ public class CommitMessageManagerTests
         private ReferenceRepository _referenceRepository;
 
         private readonly string _workingDirGitDir = @"c:\dev\repo\.git";
-        private readonly Encoding _encoding = Encoding.UTF8;
+        private readonly Encoding _encoding = new UTF8Encoding(false);
```
gerhardol, line 24: What is the motivation for this change?

gerhardol (commented):
+1
Someone else to approve too.

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
orgads: To disable BOM also for this test. It's not mandatory, but since I already handled similar cases, I preferred to have no BOM at all.

vbjay (changes requested):
I would recommend adding a test that proves that changing case of config actually results in BOM or not. I'd also verify files are not affected. Add 1 file that has BOM and one that does not and commit. Then delete the working directory files and reset hard. Verify BOM in the one file.

orgads:
> I would recommend adding a test that proves that changing case of config actually results in BOM or not. I'd also verify files are not affected. Add 1 file that has BOM and one that does not and commit. Then delete the working directory files and reset hard. Verify BOM in the one file.

This is the commit message, not a file in the repo.

I added another test for the message content.

vbjay:
> > I would recommend adding a test that proves that changing case of config actually results in BOM or not. I'd also verify files are not affected. Add 1 file that has BOM and one that does not and commit. Then delete the working directory files and reset hard. Verify BOM in the one file.
> 
> This is the commit message, not a file in the repo.
> 
> I added another test for the message content.

Correct. I am suggesting a test that proves files are not affected. A red herring test to spot issues with file content being altered.

orgads:
Why? How is this related to the commit message authoring? How would this test look like?

vbjay:
Create a test repo and create 2 files. 

- file with BOM
- file without BOM

Commit these files under your different cases. Remove files and then checkout each commit files to verify the files still are one BOM and one not. 

The test would confirm that changing the setting doesn't affect the files. I expect it to be fine but I'd like proof that whatever case of i18n.commitEncoding does not affect file content and only alters commit message.  This woukd more detect git changed behaviour but knowing what the setting does and doesn't cover and verifying it hasn't changed on us would be good.

orgads:
Thank you.

orgads:
I'll merge my tests, no reason to have 2 of them.

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
RussKie: I suggest adding a comment to spare others from searching this in docs.

RussKie (approved):
👍

UnitTests/GitCommands.Tests/CommitMessageManagerTests.cs
mstv:
```suggestion
        private readonly Encoding _encoding = new(encoderShouldEmitUTF8Identifier: false);
```
makes it clear.

mstv (commented):
LGTM

orgads:
Done

mstv:
Could you add your sign-off to `Contributors.txt` yet, please?

orgads:
Done

mstv (approved):
Thank you

PR #10438, opened 2022-11-22
Quote file name for the File History form

crypto-rsa:
## Proposed changes

Fixes the display of history for files which include a space in their path.

## Test methodology 

Manual testing - created a file with a space in its path, committed several changes and tried to view its history.

## Test environment(s) 

- Git Extensions 33.33.33
- Build 4afd375249ceede4e08a5aa49379e961409c09a7
- Git 2.38.1.windows.1
- .NET 6.0.11
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1154,6 +1154,16 @@ namespace GitUI
             return true;
         }
 
+        public void ShowFileHistoryDialog(string fileName)
+        {
+            RunFileHistoryCommand(new[]
+            {
+                "",
+                "filehistory",
+                fileName,
+            }, showBlame: false);
+        }
+
         public void StartFileHistoryDialog(IWin32Window? owner, string fileName, GitRevision? revision = null, bool filterByRevision = false, bool showBlame = false)
         {
             string arguments = AppSettings.UseBrowseForFileHistory.Value ? $"browse {PathFilterArg}={fileName.Quote()} -commit={revision?.ObjectId}"
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1154,16 +1154,6 @@ namespace GitUI
             return true;
         }
 
-        public void ShowFileHistoryDialog(string fileName)
-        {
-            RunFileHistoryCommand(new[]
-            {
-                "",
-                "filehistory",
-                fileName,
-            }, showBlame: false);
-        }
-
         public void StartFileHistoryDialog(IWin32Window? owner, string fileName, GitRevision? revision = null, bool filterByRevision = false, bool showBlame = false)
         {
             string arguments = AppSettings.UseBrowseForFileHistory.Value ? $"browse {PathFilterArg}={fileName.Quote()} -commit={revision?.ObjectId}"
@@ -1997,6 +1987,9 @@ namespace GitUI
             internal string NormalizeFileName(string fileName) => _commands.NormalizeFileName(fileName);
 
             internal bool RunCommandBasedOnArgument(string[] args) => _commands.RunCommandBasedOnArgument(args, InitializeArguments(args));
+
+            internal void ShowFileHistoryDialog(string fileName)
+                => _commands.RunFileHistoryCommand(args: new string[] { "", "", fileName }, showBlame: false);
         }
     }
 }
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1746,7 +1746,7 @@ namespace GitUI
                                  () => new FormBrowse(commands: this, new BrowseArguments
                                  {
                                      RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
-                                     PathFilter = fileHistoryFileName,
+                                     PathFilter = fileHistoryFileName.QuoteNE(),
                                      SelectedId = revision?.ObjectId,
                                      IsFileBlameHistory = true
                                  }));
@@ -1754,7 +1754,7 @@ namespace GitUI
             else
             {
                 ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
-                                 () => new FormFileHistory(commands: this, fileHistoryFileName, revision, filterByRevision, showBlame));
+                                 () => new FormFileHistory(commands: this, fileHistoryFileName.QuoteNE(), revision, filterByRevision, showBlame));
             }
 
             return true;
--- a/GitUI/GitUICommands.cs
+++ b/GitUI/GitUICommands.cs
@@ -1727,6 +1727,8 @@ namespace GitUI
                 return false;
             }
 
+            fileHistoryFileName = fileHistoryFileName.QuoteNE();
+
             GitRevision? revision = null;
             if (args.Count > 3)
             {
@@ -1756,7 +1758,7 @@ namespace GitUI
                                  () => new FormBrowse(commands: this, new BrowseArguments
                                  {
                                      RevFilter = filterByRevision ? revision?.ObjectId.ToString() : null,
-                                     PathFilter = fileHistoryFileName.QuoteNE(),
+                                     PathFilter = fileHistoryFileName,
                                      SelectedId = revision?.ObjectId,
                                      IsFileBlameHistory = true
                                  }));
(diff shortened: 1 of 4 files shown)


GitUI/GitUICommands.cs
```diff
@@ -1706,7 +1706,7 @@ private string NormalizeFileName(string fileName)
         /// <returns>false on error.</returns>
         private bool RunFileHistoryCommand(IReadOnlyList<string> args, bool showBlame)
         {
-            string fileHistoryFileName = args[2];
+            string fileHistoryFileName = args[2].QuoteNE();
```
mstv, line 1709: I think this is too early here.
It should be done in `new FormBrowse(commands: this, new BrowseArguments` on line 1749 only.

RussKie:
It would be great to have some kind of automated test for this to avoid breaking this in the future.

crypto-rsa:
> It would be great to have some kind of automated test for this to avoid breaking this in the future.

I agree but I am afraid my knowledge of GE code is not sufficient to implement this. Specifically, I am not sure how to verify that `FormFileHistory` is showing all the correct commits. And it gets even more complicated with `AppSettings.UseBrowseForFileHistory` setting on.

GitUI/GitUICommands.cs
crypto-rsa: Done in the latest revision, though I had to call it also at line 1757 for the case when `FormFileHistory` is used instead of `FormBrowse`.

RussKie:
We have quite a few tests that open dialogs (including `FormBrowse`), as well as tests that verify commits/history in the revision grid. For example,
https://github.com/gitextensions/gitextensions/blob/da718acf06d15b1c9a946deab1af30e0a4c70381/IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormBrowseTests.cs#L93

I think you pretty much can create `FormFileHistoryTests` taking an inspiration (or copying) from `FormBrowseTests`.
For the test, first you'd need to create a test repository and populated with commits. Then open `FormFileHistory` dialog and perform assertions.

crypto-rsa:
I have added some unit tests for `FormFileHistory`. Feel free to suggest changes there, it's a new area for me. I mostly followed the code in `FormBrowseTests`. I can confirm that the three test cases with spaces in them fail without the fix I've provided in this branch and pass with it.

RussKie (approved):

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormFileHistoryTests.cs
```diff
@@ -0,0 +1,67 @@
...
+[Apartment(ApartmentState.STA)]
+public class FormFileHistoryTests
+{
+    [SetUp]
+    public void SetUp()
+    {
+        AppSettings.UseBrowseForFileHistory.Value = false;
+
+        TestComposition? composition = TestComposition.Empty
+            .AddParts(typeof(MockLinkFactory));
```
RussKie, line 22: I think this needs to be this to keep tests happy:
```suggestion
            _composition = TestComposition.Empty
                .AddParts(typeof(MockLinkFactory))
                .AddParts(typeof(MockWindowsJumpListManager))
                .AddParts(typeof(MockRepositoryDescriptionProvider))
                .AddParts(typeof(MockAppTitleGenerator));
```

crypto-rsa: Yeah, that's a part of the test class which I don't understand at all - I've added a minimum code for the tests to run (and satisfy the _you ain't gonna need it_ principle) but perhaps it is _you **are** gonna need it_ after all.

GitUI/GitUICommands.cs
```diff
@@ -1746,15 +1756,15 @@ private bool RunFileHistoryCommand(IReadOnlyList<string> args, bool showBlame)
...
+                                     PathFilter = fileHistoryFileName.QuoteNE(),
                                      SelectedId = revision?.ObjectId,
                                      IsFileBlameHistory = true
                                  }));
             }
             else
             {
                 ShowModelessForm(owner: null, requiresValidWorkingDir: true, preEvent: null, postEvent: null,
-                                 () => new FormFileHistory(commands: this, fileHistoryFileName, revision, filterByRevision, showBlame));
+                                 () => new FormFileHistory(commands: this, fileHistoryFileName.QuoteNE(), revision, filterByRevision, showBlame));
```
mstv, line 1767: Then this can be done before the `if` in one place.

Though IIRC, @gerhardol has added support for filtering for multiple files recently.
Does this still work with the extra quotes?

gerhardol: This still works.
This code is used to open new GE instances. Obviously only one file can be handled with this implementation.
Multiple files in the current instance is not new, it was possible to add files in FormFilter (with or without quotes) before too. What I added was to add several files from RevDiff. This was not straightforward for multiple files and opening new instances, so that was not added.

GitUI/GitUICommands.cs
```diff
@@ -1154,6 +1154,16 @@ public bool StartBrowseDialog(IWin32Window? owner, BrowseArguments? args = null)
             return true;
         }
 
+        public void ShowFileHistoryDialog(string fileName)
```
gerhardol, line 1157: Only used in tests. Can it be moved to TestAccessor?

gerhardol (commented):
+1 
Tests fails
I would rather have removed the old FileHistory form...

RussKie:
```
C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\CommandsDialogs\FormFileHistoryTests.cs(21,13): error CS0103: The name '_composition' does not exist in the current context [C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\UI.IntegrationTests.csproj]
C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\CommandsDialogs\FormFileHistoryTests.cs(27,44): error CS0103: The name 'composition' does not exist in the current context [C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\UI.IntegrationTests.csproj]
C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\CommandsDialogs\FormFileHistoryTests.cs(21,1): error SA1137: Elements should have the same indentation [C:\projects\gitextensions\IntegrationTests\UI.IntegrationTests\UI.IntegrationTests.csproj]
```

crypto-rsa:
Thanks, I shouldn't just blindly accept suggestions from a review 🤦‍♂️

GitUI/GitUICommands.cs
mstv: Since it is needed in all branches, quote the filename in one place, please - right after checking it for `IsNullOrWhiteSpace`.

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormFileHistoryTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+[Apartment(ApartmentState.STA)]
+public class FormFileHistoryTests
+{
+    [SetUp]
+    public void SetUp()
+    {
+        AppSettings.UseBrowseForFileHistory.Value = false;
+
+        var composition = TestComposition.Empty
```
mstv, line 21: Our preferred coding style is:
```suggestion
        TestComposition composition = TestComposition.Empty
```

mstv (commented):
Just three nitpicks yet

mstv (approved):
Thank you

RussKie:
Thank you

alfeg:
Seems like this PR broken File History call from file explorer context menu

image

```
- Git Extensions 4.0.1.15887
- Build f2567dea28bbf0e2cbcd2a95d4c7200e47d0d671
- Git 2.38.1.windows.1
- .NET 6.0.12
- DPI 96dpi (no scaling)
```

mstv:
The pathname has been quoted twice. And the path has not been removed in the title bar.
It is unclear where the initial quotes come from. I cannot reproduce it on Win10 - also not with filenames containing spaces.

As a workaround, you could remove the quotes here, @alfeg:

grafik

alfeg:
Calling history window from the Git Extensions UI is working fine.
For me it's failing in the explorer context menu
image

Heavily used feature by me. Installed v4.0.0 so far to workaround.

mstv:
> from file explorer context menu
> For me it's failing in the explorer context menu

That's what I have tested.

PR #10451, opened 2022-11-26
Set custom difftool list to empty if parsing is cancelled

### LINKED ISSUE #10435: [NBug] Value cannot be null. (Parameter 'source')
### Environment

- Git Extensions 4.0.0.15569
- Git 2.38.1.windows.1
- .NET 6.0.11
- DPI 120dpi (125% scaling)

### Issue description

```
System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at void System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at List<TSource> System.Linq.Enumerable.ToList<TSource>(IEnumerable<TSource> source)
   at async void GitUI.CustomDiffMergeToolProvider.LoadCustomDiffMergeTools(GitModule module, IList<CustomDiffMergeTool> menus, IContainer components, bool isDiff, int delay, CancellationToken cancellationToken)+(?) => { } [0]
```

there is error in settings dialog 
try to change ssh settings

### Steps to reproduce

some action ))

_No response_

### Diagnostics

_No response_

gerhardol:
Fixes #10435

## Proposed changes

If loading custom diff tools is cancelled, the list could be null and a null exception occur.

This has been around since custom difftools list (3.5?) so not urgent for 4.0.

## Test methodology 

Code review

## Merge strategy

----

### DIFF
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -67,16 +67,14 @@ namespace GitCommands
                     _tools = ParseCustomDiffMergeTool(output, defaultTool);
                 }
             }
-            catch
+            finally
             {
                 if (_tools is null)
                 {
                     // Parsing has failed, just provide an empty list, no user notification
                     _tools = Array.Empty<string>();
                 }
-            }
-            finally
-            {
+
                 _mutex.Release();
             }
 
--- a/GitCommands/CustomDiffMergeToolCache.cs
+++ b/GitCommands/CustomDiffMergeToolCache.cs
@@ -67,6 +67,10 @@ namespace GitCommands
                     _tools = ParseCustomDiffMergeTool(output, defaultTool);
                 }
             }
+            catch
+            {
+                // No action.
+            }
             finally
             {
                 if (_tools is null)


mstv (approved):

RussKie (commented):
What'd happen if there was an exception?

gerhardol:
> What'd happen if there was an exception?

just catching that...

RussKie (approved):

GitCommands/CustomDiffMergeToolCache.cs
```diff
@@ -68,15 +68,17 @@ public async Task<IEnumerable<string>> GetToolsAsync(GitModule module, int delay
                 }
             }
             catch
+            {
+                // No action.
```
mstv, line 72: Non-cancellation exceptions are forwarded by the `.FileAndForget()` in `LoadCustomDiffMergeTools()`.
Why should we swallow them?

gerhardol: Keep this as non important, similar to git-status refreshes in the background, to avoid user confusion for non critical issues.

gerhardol: Unless you request changes, I would like to merge this tomorrow.

mstv: I don't. We can ask for the Git Command Log in case of issues.

PR #10453, opened 2022-11-27
Make "blame" setting non-sticky

RussKie:
Resolves https://github.com/gitextensions/gitextensions/issues/10511

"RevisionFileTreeShowBlame" settings was introduced in #9424.

The issue with this setting is that whenever a file history is opened, blame is turned on for all open instances of the app, and affects the File Tree tab experience.

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -849,12 +849,10 @@ namespace GitCommands
             set => SetBool("opensubmodulediffinseparatewindow", value);
         }
 
-        // This is a "view settings" which is configured for the lifetime of the app instance,
-        // and it doesn't get persisted and shared between different instances of the app.
         public static bool RevisionFileTreeShowBlame
         {
-            get;
-            set;
+            get => GetBool("RevisionFileTreeShowBlame", true);
+            set => SetBool("RevisionFileTreeShowBlame", value);
         }
 
         /// <summary>
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -849,12 +849,6 @@ namespace GitCommands
             set => SetBool("opensubmodulediffinseparatewindow", value);
         }
 
-        public static bool RevisionFileTreeShowBlame
-        {
-            get => GetBool("RevisionFileTreeShowBlame", true);
-            set => SetBool("RevisionFileTreeShowBlame", value);
-        }
-
         /// <summary>
         /// Gets or sets whether to show artificial commits in the revision graph.
         /// </summary>
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -849,10 +849,12 @@ namespace GitCommands
             set => SetBool("opensubmodulediffinseparatewindow", value);
         }
 
+        // This is a "view settings" which is configured for the lifetime of the app instance,
+        // and it doesn't get persisted and shared between different instances of the app.
         public static bool RevisionFileTreeShowBlame
         {
-            get => GetBool("RevisionFileTreeShowBlame", true);
-            set => SetBool("RevisionFileTreeShowBlame", value);
+            get;
+            set;
         }
 
         /// <summary>
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -250,12 +250,6 @@ namespace GitUI.CommandsDialogs
             SystemEvents.SessionEnding += (sender, args) => SaveApplicationSettings();
 
             _isFileBlameHistory = args.IsFileBlameHistory;
-            if (_isFileBlameHistory)
-            {
-                // Blame was explicitly requested (otherwise start with last used)
-                AppSettings.RevisionFileTreeShowBlame = true;
-            }
-
             InitializeComponent();
             BackColor = OtherColors.BackgroundColor;
 
@@ -315,7 +309,7 @@ namespace GitUI.CommandsDialogs
             toolStripButtonPush.Initialize(_aheadBehindDataProvider);
             repoObjectsTree.Initialize(_aheadBehindDataProvider, filterRevisionGridBySpaceSeparatedRefs: ToolStripFilters.SetBranchFilter, RevisionGrid, RevisionGrid, RevisionGrid);
             revisionDiff.Bind(RevisionGrid, fileTree, RefreshGitStatusMonitor);
-            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor);
+            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor, _isFileBlameHistory ? true : AppSettings.RevisionFileTreeShowBlame);
             RevisionGrid.ResumeRefreshRevisions();
 
             // Application is init, the repo related operations are triggered in OnLoad()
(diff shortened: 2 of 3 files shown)


mstv (commented):
A comment should be added.

I remember attempts to restore such settings before closing, which are incomplete or do not work as expected.

gerhardol:
This also changes the defaults for the setting, Blame is always off by default now.
If the persistency is removed, AppSettings should be removed completely. The only quirk is that FormBrowse:_isFileBlameHistory must set the value in FileTree.

This PR will hide Blame functionality for users. I would prefer to keep the persistency, but there has been opinions against it (those that are pro has not said anything, as usual negative feedback, that is human nature, me included).

mstv:
We should rather pursue to define a set of "view settings". For these, the user shall be able to set defaults, which will be loaded on startup and perhaps on repo switch.
Changes during runtime will not be persisted neither live nor on close.

RussKie:
> This also changes the defaults for the setting, Blame is always off by default now.
> If the persistency is removed, AppSettings should be removed completely. The only quirk is that FormBrowse:_isFileBlameHistory must set the value in FileTree.

I don't think the settings needs to be removed - it allows persisting the settings for the duration of the session. That is, if I enable in one instance the setting will remain on until I close this instance. And if I open another instance - the setting will be off by default.
This allows auto-enabling the settings when we open a new instance to show file history, and does not negatively impact other running instances.


> This PR will hide Blame functionality for users.

I don't believe it does. The option was always available, in https://github.com/gitextensions/gitextensions/pull/9424 you brought it closer the user - i.e., the blame is available without needing to open another dialog. However, the current implementation has multiple negative impacts, including UX (i.e., once I open the File History dialog all my other instances have blame enabled) and perf (it takes much longer to load blame information when I don't need it because it was auto-enabled for me).

gerhardol:
> I don't think the settings needs to be removed - it allows persisting the settings for the duration of the session.

The mode is set when opening the FileTree instance, no need to use AppSettings to store that.

> > This PR will hide Blame functionality for users.
> 
> I don't believe it does. The option was always available, in #9424 you brought it closer the user - i.e., the blame is available without needing to open another dialog. However, the current implementation has multiple negative impacts, including UX (i.e., once I open the File History dialog all my other instances have blame enabled) and perf (it takes much longer to load blame information when I don't need it because it was auto-enabled for me).

2. Before this PR, Blame was default in FileTree.
I will not vote against changing this default change.

3. Use settings when opening the new FileHistory
I agree that should be changed, AppSettings should not be involved when opening FileHistory (this is much easier to handle than hiding the left panel, also done in this "mode").

I can make these changes if you prefer.

gerhardol:
I pushed a proposal to my repo. Cnges to blame is only persisted at explicit changes in the menu/hotkey.
https://github.com/gerhardol/gitextensions/pull/new/RussKie/_make_blame_non_sticky

If you want to remove the setting completely (never remember the last setting), just remove the appsetting.

RussKie:
> I pushed a proposal to my repo. Cnges to blame is only persisted at explicit changes in the menu/hotkey.
> https://github.com/gerhardol/gitextensions/pull/new/RussKie/_make_blame_non_sticky

Thank you. This is definitely an option, though we'd need to add tests for to ensure we don't accidentally regress the behaviour.


> If you want to remove the setting completely (never remember the last setting), just remove the appsetting.

I'll need to test this scenario.

RussKie:
> If you want to remove the setting completely (never remember the last setting), just remove the appsetting.

It doesn't look simple - even if we remove the setting, we still need to maintain the state of a flag somehow - and that's what appsettings are for :) This settings is different from the rest, as @mstv put it earlier, this is a "view setting", which lives for the lifetime of the app instance but is not persisted between sessions.

gerhardol:
> > If you want to remove the setting completely (never remember the last setting), just remove the appsetting.
> 
> It doesn't look simple - even if we remove the setting, we still need to maintain the state of a flag somehow - and that's what appsettings are for :) This settings is different from the rest, as @mstv put it earlier, this is a "view setting", which lives for the lifetime of the app instance but is not persisted between sessions.

You do not need an extra state, all handled when initializing. 
In my branch, the value was only really read when binding file tree, the other were related to updating the settings..

GitCommands/Settings/AppSettings.cs
```diff
@@ -849,10 +849,12 @@ public static bool OpenSubmoduleDiffInSeparateWindow
...
         }
 
+        // This is a "view settings" which is configured for the lifetime of the app instance,
+        // and it doesn't get persisted and shared between different instances of the app.
         public static bool RevisionFileTreeShowBlame
         {
-            get => GetBool("RevisionFileTreeShowBlame", true);
-            set => SetBool("RevisionFileTreeShowBlame", value);
+            get;
+            set;
```
mstv, line 857: A "view setting" _is_ an AppSetting
- with a user-configurable initial value in the settings file,
- which can be changed in the running instance,
- can be persisted on explicit user request (in the future).

I.e. something like (if the compiler would support it):
```suggestion
        public static bool RevisionFileTreeShowBlame
        {
            get;
            set;
        } = PersistentRevisionFileTreeShowBlame;
        
        public static bool PersistentRevisionFileTreeShowBlame
        {
            get => GetBool("RevisionFileTreeShowBlame", true);
            set => SetBool("RevisionFileTreeShowBlame", value);
        }
```

gerhardol:
I pushed my branch again:
* First commit as before, do not update settings when starting file history or  blaming from RevDiff. Settings is updated when toggling the settings (that should be postponed to when closing the app).
* Second commit removes AppSettings, starting by default with View.

RussKie:
> I pushed my branch again:
> 
> ...
> * Second commit removes AppSettings, starting by default with View.

This works great! Thanks.
I'll take your latest version.

RussKie:
If there're no objections, I'd like to merge this in the next few days and have it serviced in 4.0.1

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
```diff
@@ -134,9 +134,10 @@ public void ExpandToFile(string filePath, int? line, bool requestBlame)
                     return;
                 }
 
-                if (requestBlame && !AppSettings.RevisionFileTreeShowBlame)
+                if (requestBlame)
                 {
-                    blameToolStripMenuItem1.Checked = AppSettings.RevisionFileTreeShowBlame = true;
+                    // Force blame without changing settings
```
gerhardol, line 139: obsolete comment

mstv (commented):
+1 ;-) similar as Gerhard
We can add persistence again when we have managed to implement "view settings".

GitUI/CommandsDialogs/RevisionFileTreeControl.cs
RussKie, line 139:
```suggestion
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -315,7 +309,8 @@ public FormBrowse(GitUICommands commands, BrowseArguments args)
             toolStripButtonPush.Initialize(_aheadBehindDataProvider);
             repoObjectsTree.Initialize(_aheadBehindDataProvider, filterRevisionGridBySpaceSeparatedRefs: ToolStripFilters.SetBranchFilter, RevisionGrid, RevisionGrid, RevisionGrid);
             revisionDiff.Bind(RevisionGrid, fileTree, RefreshGitStatusMonitor);
-            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor);
+
+            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor, _isFileBlameHistory);
```
RussKie, line 313:
```suggestion
            // Show blame by default if not started from command line
            fileTree.Bind(RevisionGrid, RefreshGitStatusMonitor, _isFileBlameHistory);
```

PR #10456, opened 2022-11-27
Reinstate dev build start-up optimisation reverted in 7bd205f23c

RussKie:
Avoid scanning thousands of assemblies while loading plugins.


mstv (approved):
👍 Thank you!

gerhardol:
With this PR only a few of the GE plugins are loaded, for instance "AppVeyorIntegration.dll" is not loaded.
This applies to both debugger builds and commandline builds.
Any other way to change the assembly names than changing the namespace or explicitly setting the AssemblyName?

image

RussKie:
Ah, bugger! I didn't check that all plugins were loaded.

PR #10473, opened 2022-11-29
WIP GPG Improvements Part 1

vbjay:
## Proposed changes

- GPG key id selection from gpg -K output in a dropdown
- Reflecting default actions better for signing tags/commits.  
    - If commit.gpgSign is true then show in commit dialog that a commit will be signed by default and what key is being used. 
    - Do the same for tags but tag.gpgSign config.
- pass no-gpg-sign or config in command line to actually not sign the commit or tag when user does select do not sign commit or relevant tag type.  Fixes flaw where user selected the option but git config defaults the signing.

### Before

 TODO 

### After

image
 

## Test methodology 

- Added tests
-  Debugged and verified functionality of ui and gpg signing info of commits 

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy
- Merge commit. (PR submitter to rebase and squash before merges).
- 

----


RussKie:
@vbjay  a friendly nudge. Is this still something you're working on?

vbjay:
Yes.  I need to rework some forms and add more tests.

vbjay:
image
Verified appveyor generates a key that tests can access and the controls I wrote see the keys.

vbjay:
@gerhardol @RussKie @mstv I'll document changes better in pr later better but please take a look at functionality for a spin.  Let me know if you have feedback.

gerhardol:
First time running this if no key is setup, you get an exception:

GitExtUtils.ExternalOperationException
  HResult=0x80131500
  Message=Cannot start process because a file name has not been provided.
  Source=GitCommands
  StackTrace:
   at GitCommands.Executable.ProcessWrapper..ctor(String fileName, String prefixArguments, String arguments, String workDir, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding, Boolean useShellExecute, Boolean throwOnErrorExit) in C:\dev\gc\gitextensions_4\GitCommands\Git\Executable.cs:line 136
   at GitCommands.Executable.Start(ArgumentString arguments, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding, Boolean useShellExecute, Boolean throwOnErrorExit) in C:\dev\gc\gitextensions_4\GitCommands\Git\Executable.cs:line 48
   at GitCommands.ExecutableExtensions.<GetOutputAsync>d__4.MoveNext() in C:\dev\gc\gitextensions_4\GitCommands\Git\ExecutableExtensions.cs:line 102
   at GitCommands.Gpg.GitGpgController.<GetGpgSecretKeysAsync>d__27.MoveNext() in C:\dev\gc\gitextensions_4\GitCommands\Git\GitGpgController.cs:line 329
   at GitUI.UserControls.GPGKeys.GPGKeysUIController.<GetKeysAsync>d__3.MoveNext() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGKeysUIController.cs:line 22
   at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.<<LoadKeys>b__13_0>d.MoveNext() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 51
   at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.LoadKeys() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 49
   at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.InitControllers() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 44
   at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.OnUICommandsSourceSet(IGitUICommandsSource source) in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 104
   at GitUI.GitModuleControl.set_UICommandsSource(IGitUICommandsSource value) in C:\dev\gc\gitextensions_4\GitUI\GitModuleControl.cs:line 76
   at GitUI.CommandsDialogs.FormCreateTag..ctor(GitUICommands commands, ObjectId objectId) in C:\dev\gc\gitextensions_4\GitUI\CommandsDialogs\FormCreateTag.cs:line 55
   at GitUI.RevisionGridControl.<>c__DisplayClass160_0.<CreateTagToolStripMenuItemClick>b__0() in C:\dev\gc\gitextensions_4\GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:line 1722
   at GitUI.GitUICommands.DoActionOnRepo(IWin32Window owner, Func`1 action, Boolean requiresValidWorkingDir, Boolean changesRepo, EventHandler`1 preEvent, EventHandler`1 postEvent) in C:\dev\gc\gitextensions_4\GitUI\GitUICommands.cs:line 318
   at GitUI.GitUICommands.DoActionOnRepo(Func`1 action) in C:\dev\gc\gitextensions_4\GitUI\GitUICommands.cs:line 338
   at GitUI.RevisionGridControl.CreateTagToolStripMenuItemClick(Object sender, EventArgs e) in C:\dev\gc\gitextensions_4\GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:line 1720
   at System.Windows.Forms.ToolStripItem.FireEventInteractive(EventArgs e, ToolStripItemEventType met)
   at System.Windows.Forms.ToolStripItem.FireEvent(EventArgs e, ToolStripItemEventType met)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, WM msg, IntPtr wparam, IntPtr lparam)

  This exception was originally thrown at this call stack:
    [External Code]
    GitCommands.Executable.ProcessWrapper.ProcessWrapper(string, string, string, string, bool, bool, bool, System.Text.Encoding, bool, bool) in Executable.cs

Inner Exception 1:
InvalidOperationException: Cannot start process because a file name has not been provided.

vbjay:
> First time running this if no key is setup, you get an exception:
> 
> GitExtUtils.ExternalOperationException
>   HResult=0x80131500
>   Message=Cannot start process because a file name has not been provided.
>   Source=GitCommands
>   StackTrace:
>    at GitCommands.Executable.ProcessWrapper..ctor(String fileName, String prefixArguments, String arguments, String workDir, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding, Boolean useShellExecute, Boolean throwOnErrorExit) in C:\dev\gc\gitextensions_4\GitCommands\Git\Executable.cs:line 136
>    at GitCommands.Executable.Start(ArgumentString arguments, Boolean createWindow, Boolean redirectInput, Boolean redirectOutput, Encoding outputEncoding, Boolean useShellExecute, Boolean throwOnErrorExit) in C:\dev\gc\gitextensions_4\GitCommands\Git\Executable.cs:line 48
>    at GitCommands.ExecutableExtensions.<GetOutputAsync>d__4.MoveNext() in C:\dev\gc\gitextensions_4\GitCommands\Git\ExecutableExtensions.cs:line 102
>    at GitCommands.Gpg.GitGpgController.<GetGpgSecretKeysAsync>d__27.MoveNext() in C:\dev\gc\gitextensions_4\GitCommands\Git\GitGpgController.cs:line 329
>    at GitUI.UserControls.GPGKeys.GPGKeysUIController.<GetKeysAsync>d__3.MoveNext() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGKeysUIController.cs:line 22
>    at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.<<LoadKeys>b__13_0>d.MoveNext() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 51
>    at Microsoft.VisualStudio.Threading.JoinableTask.CompleteOnCurrentThread()
>    at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.LoadKeys() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 49
>    at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.InitControllers() in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 44
>    at GitUI.UserControls.GPGKeys.GPGSecretKeysCombobox.OnUICommandsSourceSet(IGitUICommandsSource source) in C:\dev\gc\gitextensions_4\GitUI\UserControls\GPGKeys\GPGSecretKeysCombobox.cs:line 104
>    at GitUI.GitModuleControl.set_UICommandsSource(IGitUICommandsSource value) in C:\dev\gc\gitextensions_4\GitUI\GitModuleControl.cs:line 76
>    at GitUI.CommandsDialogs.FormCreateTag..ctor(GitUICommands commands, ObjectId objectId) in C:\dev\gc\gitextensions_4\GitUI\CommandsDialogs\FormCreateTag.cs:line 55
>    at GitUI.RevisionGridControl.<>c__DisplayClass160_0.<CreateTagToolStripMenuItemClick>b__0() in C:\dev\gc\gitextensions_4\GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:line 1722
>    at GitUI.GitUICommands.DoActionOnRepo(IWin32Window owner, Func`1 action, Boolean requiresValidWorkingDir, Boolean changesRepo, EventHandler`1 preEvent, EventHandler`1 postEvent) in C:\dev\gc\gitextensions_4\GitUI\GitUICommands.cs:line 318
>    at GitUI.GitUICommands.DoActionOnRepo(Func`1 action) in C:\dev\gc\gitextensions_4\GitUI\GitUICommands.cs:line 338
>    at GitUI.RevisionGridControl.CreateTagToolStripMenuItemClick(Object sender, EventArgs e) in C:\dev\gc\gitextensions_4\GitUI\UserControls\RevisionGrid\RevisionGridControl.cs:line 1720
>    at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
>    at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
>    at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
>    at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
>    at System.Windows.Forms.ToolStripItem.FireEventInteractive(EventArgs e, ToolStripItemEventType met)
>    at System.Windows.Forms.ToolStripItem.FireEvent(EventArgs e, ToolStripItemEventType met)
>    at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
>    at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
>    at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
>    at System.Windows.Forms.Control.WndProc(Message& m)
>    at System.Windows.Forms.ScrollableControl.WndProc(Message& m)
>    at System.Windows.Forms.ToolStrip.WndProc(Message& m)
>    at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
>    at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
>    at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
>    at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, WM msg, IntPtr wparam, IntPtr lparam)
> 
>   This exception was originally thrown at this call stack:
>     [External Code]
>     GitCommands.Executable.ProcessWrapper.ProcessWrapper(string, string, string, string, bool, bool, bool, System.Text.Encoding, bool, bool) in Executable.cs
> 
> Inner Exception 1:
> InvalidOperationException: Cannot start process because a file name has not been provided.
> 

Thanks. I should test a clean setup. I'll get that updated.

vbjay:
@gerhardol  Got it right.  

Do the following test

1. Run update-loc
2. Run `.\scripts\Start-CleanMachineTests.ps1` in admin powershell.  It will install the windows feature Windows Sandbox if needed and start a sandbox
3. Wait for sandbox setup to finish
4. Git Extensions will start( may need to minimize and restore sandbox window.  Allows you to click on Language selection button)
5. Try out tag, commit, rebase dialogs, see how they behave.
6. run `git config commit.gpgSign true` and `git config tag.gpgSign true` and try again.  notice behaviour change in dialogs.

I plan on updating readme with sandbox info later.  For now the focus is on gpg functionality and the sandbox was a way to have a clean machine.

vbjay:
```patch
diff --git a/scripts/wsb/Setup-Repo.ps1 b/scripts/wsb/Setup-Repo.ps1
index dc379219e..6af982dd0 100644
--- a/scripts/wsb/Setup-Repo.ps1
+++ b/scripts/wsb/Setup-Repo.ps1
@@ -11,7 +11,7 @@ function CreateCommits()
     [Guid]::NewGuid()|Out-File -FilePath .\initial.txt
     git add .
     # -S signs with default gpg key
-    git commit -S -m "Initial commit"
+    git commit -m "Initial commit"
 }
 git init .\Repo
 cp "$env:USERPROFILE\Desktop\Bin\gitex.cmd" .\Repo
diff --git a/scripts/wsb/setup.ps1 b/scripts/wsb/setup.ps1
index b851a8f37..b98f47cf3 100644
--- a/scripts/wsb/setup.ps1
+++ b/scripts/wsb/setup.ps1
@@ -12,7 +12,7 @@
     Explanation of the function or its result. You can include multiple examples with additional .EXAMPLE lines
 #>
 
-
+Start-Sleep -Seconds 10
 function ShowFileExtensions() {
     # http://superuser.com/questions/666891/script-to-set-hide-file-extensions
     Push-Location
@@ -41,16 +41,16 @@ $paths += 'C:\Users\WDAGUtilityAccount\Desktop\artifacts\Debug\bin\GitExtensions
 Update-SessionEnvironment
 
 #setup a new gpg key
-gpg --batch --generate-key .\wsb\gen-gpg.dat
-#just generate same key info but another key so we can see default vs non default key
-gpg --batch --generate-key .\wsb\gen-gpg.dat
-gpg  -K --with-colons | awk -F: '/^sec:/ { print $5 }' > gpgkey.txt
-$gpgKey = Get-Content gpgkey.txt -Head 1
+# gpg --batch --generate-key .\wsb\gen-gpg.dat
+# #just generate same key info but another key so we can see default vs non default key
+# gpg --batch --generate-key .\wsb\gen-gpg.dat
+# gpg  -K --with-colons | awk -F: '/^sec:/ { print $5 }' > gpgkey.txt
+# $gpgKey = Get-Content gpgkey.txt -Head 1
 
 #configure git
 git config --global user.name 'Tester'
 git config --global user.email 'Test@test.com'
-git config --global user.signingKey "$gpgKey"
+#git config --global user.signingKey "$gpgKey"
 git config --global rebase.autosquash "true"
 git config --global rebase.autostash 'true'
 git config --global rebase.updaterefs 'true'
```

test confirmed no exceptions

bansan85:
First point : when I click on commit button, I have error:

```
GitExtUtils.ExternalOperationException: 'An error occurred trying to start process 'C:\Program Files\Git\bin\gpg.exe' with working directory 'C:\Users\XXXXXX\Documents\git\perso\gitextensions\'. 
```

In my case, gpg.exe is installed on "C:\Program Files\Git\usr\bin\gpg.exe".

If GPG is missing, I think the error should be reported on startup (in the same window when git is obsolete)

image

Maybe the gpg.exe path should be settable in settings window ? Instead of use `git config gpg.program XXX/gpg.exe`.


Second point : when I select the GPG key in the commit form, the menu `Options` closes. It should not. The "Do not sign commit" option doesn't close the menu `Options`.


One last point : in my keyring, I have some key that is expired. Maybe it could be fine to show this information. But not mandatory.


You did a create work. Thanks for this feature :)



```
- Git Extensions 33.33.33
- Build 715b2f510d5e384e7d69b1e996c72eacf6788f7e (Dirty)
- Git 2.40.1.windows.1
- .NET 6.0.16
- DPI 96dpi (no scaling)
- Portable: False
```

vbjay:
1.  I did this on purpose to hopefully find any issues like this. Part 2 will be settings work. I fixed it to point to the Linux bin tools version of gpg if program is not set. Please try latest version and let me know if it works and if not show me the git\paths in ge settings.

1.  Thanks. I think I forgot a part when I reworked it. I'll look at reimplementing that. 

1.  I can definitely add expired detection and display or even filtering. I'll probably go with filtering.

bansan85:
> 1.  I did this on purpose to hopefully find any issues like this. Part 2 will be settings work. I fixed it to point to the Linux bin tools version of gpg if program is not set. Please try latest version and let me know if it works and if not show me the git\paths in ge settings.

I already have the latest version. Commit 715b2f510d5e384e7d69b1e996c72eacf6788f7e is the latest of your branch.

vbjay:
Ok. Show me a screenshot of your git extensions settings git\paths node selected.

bansan85:
I have git for Windows (https://git-scm.com/download/win) installed.

image

image

image

gerhardol:
Linux tools paths must be set to "C:\Program Files\Git\usr\bin\"

GitUI/CommandsDialogs/FormMergeBranch.Designer.cs
```diff
@@ -466,5 +526,9 @@ private void InitializeComponent()
         private System.Windows.Forms.TextBox mergeMessage;
         private System.Windows.Forms.CheckBox addMergeMessage;
         private System.Windows.Forms.Button Ok;
+        private ComboBox comboboxGpgSignType;
+        private Label label1;
```
gerhardol, line 530: Minor: Avoid the defalt name like label1

GitUI/CommandsDialogs/FormRebase.Designer.cs
```diff
@@ -28,679 +28,742 @@ protected override void Dispose(bool disposing)
         /// </summary>
         private void InitializeComponent()
         {
-            this.components = new System.ComponentModel.Container();
```
gerhardol, line 31: Was removal of "this." that Designer normally adds an explicit change?
It makes it harder to review.

gerhardol (commented):
This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).

But no direct comments on how this is implemented or how it looks.
I do not use signing myself. Lack of support for --ff-only in GitHub is a blocker for me.

vbjay:
> I have git for Windows (https://git-scm.com/download/win) installed.
> 
> image
> 
> image
> 
> image
> 

There is your problem. The second setting in paths should be usr\bin folder. That is the Linux bin tools folder not gir\bin

GitUI/CommandsDialogs/FormRebase.Designer.cs
vbjay: Pain in the ..... designer. I'll revert it

vbjay:
I'll still harden it to handle can't find GPG. and update based on feedback.

GitUI/UserControls/GPGKeys/GPGKeysUIController.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public GPGKeysUIController()
+        {
+            // just empty
+        }
+
+        public async Task<IEnumerable<GpgKeyDisplayInfo>> GetKeysAsync()
+        {
+            var keys = await _gpgController.GetGpgSecretKeysAsync();
```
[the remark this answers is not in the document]
vbjay: Sure.  I'll run through those.

bansan85:
> Linux tools paths must be set to "C:\Program Files\Git\usr\bin"

Thanks.

I thought that "Path to linux tools (sh)" means "path to sh.exe". sh.exe is in both "bin" and "usr/bin".

Do you think the label should be updated to avoid this mistake ?

GitCommands/Config/SettingKeyString.cs
```diff
@@ -65,5 +65,31 @@ public static class SettingKeyString
...
         /// </summary>
         public static string AllowFileProtocol = "protocol.file.allow";
+
+        /// <summary>
+        /// commit.gpgSign  Determines if commits are signed by default
+        /// </summary>
+        /// <remarks>
+        /// Min version 2.0
+        /// </remarks>
+        public static string CommitGPGSign = "commit.gpgSign";
```
RussKie, line 75:
```suggestion
        public static string CommitGpgSign = "commit.gpgSign";
```

Ditto everywhere, including type names.

GitCommands/Config/SettingKeyString.cs
```diff
@@ -65,5 +65,31 @@ public static class SettingKeyString
...
+
+        /// <summary>
+        /// user.signingKey  Stores the key ID to use when signing a commit with GPG
+        /// </summary>
+        public static string UserSigningKey = "user.signingKey";
+
+        /// <summary>
+        /// gpg.program  Stores the path to the gpg executable that will run when running gpg commands
+        /// </summary>
+        public static string GPGProgram = "gpg.program";
```
RussKie, line 93: [nit] It'd be great to lexicographically sort all entries in this file.

GitCommands/Git/Commands/GitCommandHelpers.cs
```diff
@@ -555,21 +577,43 @@ public static ArgumentString GetAllChangedFilesCmd(bool excludeIgnoredFiles, Unt
...
+        public static ArgumentString MergeBranchCmd(string branch,
+                                                    bool allowFastForward,
+                                                    bool squash,
+                                                    bool noCommit,
+                                                    string strategy,
+                                                    bool allowUnrelatedHistories,
+                                                    string? mergeCommitFilePath,
+                                                    int? log,
+                                                    bool gpgSign = false,
+                                                    string? gpgKeyId = "",
```
RussKie, line 589: In `RebaseCmd` this argument defaults to `null`. Why is it different here?

GitCommands/Git/Executable.cs
```diff
@@ -27,6 +27,8 @@ public Executable(Func<string> fileNameProvider, string workingDir = "", string
             _prefixArguments = prefixArguments;
         }
 
+        public Func<string> FileNameProvider { get => _fileNameProvider; }
```
RussKie, line 30: Please add xml-doc and descript the purpose of this.

RussKie: Does it need to be `public`?

GitCommands/Git/GPGLine.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace GitCommands.Gpg
+{
+    internal sealed class GPGLine // : IEquatable<GPGLine>
```
RussKie, line 3: Does it need to be a `class`? Can it be a struct? A readonly struct?

Also, remove commented out code.

GitCommands/Git/GPGSecretKeysParser.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public class GPGSecretKeysParser : IGPGSecretKeysParser
+    {
+        /// <inherit/>
+        public IEnumerable<GpgKeyInfo> ParseKeysOutput(string txt, string? defaultKey)
+        {
+            try
+            {
+                var rec = txt.Split('\n').
```
RussKie, line 19: Only use `var` when the type is obvious, here and below that's not the case.

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2945,9 +2946,26 @@ private void LaunchRebase(string command)
...
+            bool commitSign = false;
+            string defaultKey = null;
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                commitSign = await gpg.GetCommitGPGSignAsync();
+                defaultKey = await gpg.GetDefaultKeyAsync();
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            });
+
+            string rebaseCmd = GitCommandHelpers.RebaseCmd(GetActualRevision(LatestSelectedRevision)?.FirstParentId?.ToString(),
```
RussKie, line 2959: I think we better express this as:
```suggestion
            ThreadHelper.JoinableTaskFactory.Run(async () =>
            {
                < await switch to background thread >
                commitSign = await gpg.GetCommitGPGSignAsync();
                defaultKey = await gpg.GetDefaultKeyAsync();
            });

            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
            string rebaseCmd = GitCommandHelpers.RebaseCmd(GetActualRevision(LatestSelectedRevision)?.FirstParentId?.ToString(),
```

IntegrationTests/UI.IntegrationTests/CommandsDialogs/FormCommitTests.cs
```diff
@@ -547,6 +547,44 @@ public void SelectedDiff_remembers_geometry()
...
+            RunFormTest(async form =>
+            {
+                using (CancellationTokenSource cts = new(AsyncTestHelper.UnexpectedTimeout))
+                {
+                    await ThreadHelper.JoinPendingOperationsAsync(cts.Token);
+                }
+
+                var accessor = form.GetTestAccessor();
+                var keys = accessor.toolStripGpgKeyComboBox.CurrentKeys
+                       .Skip(1) // Ignore the No key selected entry
```
RussKie, line 577:
```suggestion
                       .Skip(1) // Ignore the "No key selected" entry
```

GitUI/UserControls/GPGKeys/GpgKeyDisplayInfo.cs
```diff
@@ -0,0 +1,100 @@
+﻿using ResourceManager;
+
+namespace GitUI.UserControls.GPGKeys
+{
+    public class GpgKeyDisplayInfo : IEquatable<GpgKeyDisplayInfo>
+    {
+        internal class Translations : Translate
+        {
+            internal readonly TranslationString _noKeySelected = new("No key selected");
+        }
```
RussKie, line 10: Please use https://github.com/gitextensions/gitextensions/blob/master/GitUI/TranslatedStrings.cs instead

GitUI/UserControls/GPGKeys/GpgKeyDisplayInfo.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public GpgKeyDisplayInfo(string fingerprint, string keyID, string userID, bool isDefault)
```
RussKie, line 12:
```suggestion
        public GpgKeyDisplayInfo(string fingerprint, string keyId, string userId, bool isDefault)
```

GitUI/UserControls/GPGKeys/GpgKeyDisplayInfo.cs
```diff
@@ -0,0 +1,100 @@
+﻿using ResourceManager;
+
+namespace GitUI.UserControls.GPGKeys
+{
+    public class GpgKeyDisplayInfo : IEquatable<GpgKeyDisplayInfo>
```
RussKie, line 5: Can this be `struct` or `readonly struct`?

GitCommands/Git/GitGpgController.cs
```diff
@@ -24,12 +25,30 @@ public enum TagStatus
...
+        /// Gets whether tags are signed by default
+        /// </summary>
+        /// <returns> true for default to signed tags</returns>
+        Task<bool> GetTagGPGSignAsync();
+
+        /// <summary>
+        /// Gets whether commits are signed by default
+        /// </summary>
+        /// <returns> true for default to signed commits</returns>
+        Task<bool> GetCommitGPGSignAsync();
```
RussKie, line 50: I'm not really excited by having multiple new public API.... Have you considered adding only a single method that would read the whole lot of GPG-related settings and return those as a bundle (e.g., a struct)?

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System.ComponentModel;
+using System.Diagnostics;
+using GitCommands.Gpg;
+
+namespace GitUI.UserControls.GPGKeys
+{
+    public partial class GPGSecretKeysCombobox : GitModuleControl
```
RussKie, line 7: Can we avoid inheriting from `GitModuleControl`?

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -2945,9 +2946,26 @@ private void LaunchRebase(string command)
...
+                                                           interactive: true,
+                                                           preserveMerges: false,
+                                                           autosquash: false,
+                                                           autoStash: true,
+                                                           ignoreDate: false,
+                                                           committerDateIsAuthorDate: false,
+                                                           supportRebaseMerges: Module.GitVersion.SupportRebaseMerges,
+                                                           gpgSign: commitSign && !string.IsNullOrEmpty(defaultKey),
+                                                           supportNoGpgSign: Module.GitVersion.SupportNoGPGSign);
```
RussKie, line 2968: This feels very wrong that the revision grid needs to own a GPG controller. This also introduces a non-mockable dependency.

RussKie (changes requested):

GitCommands/Git/GitGpgController.cs
vbjay: The reason is that I avoid calling multiple configs when I only need the piece that I need.  Tjis controller already existed and it drives the  things you do with GPG.  If i do as you suggest then I will have to call all git config options even though I only need x item at the time.

- commit.gpgSign
- tag.gpgSign
- gpg.program
- user.signingKey

No I cannot just use the config settings cache because I must get current config at the time of run.  So this allows me to ask for what I want when I need it.  If I need tag.gpgSign then i ask for it.  If I need the default key... and so on.  Just because this controller only had gpg signature display items, it made sense for this to drive the whole functionality of GPG info.

GitUI/UserControls/GPGKeys/GpgKeyDisplayInfo.cs
vbjay: I'll look into it.  Probably.

GitCommands/Config/SettingKeyString.cs
vbjay: Sure.  I'll do as a post fix after all other review items and completed since it will be a likely merge conflict cause.

GitCommands/Git/Commands/GitCommandHelpers.cs
vbjay: Good catch.  Yes it can match null.

GitCommands/Git/Executable.cs
vbjay: image

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
vbjay: How do you suggest I get the current module and the ui command source and all that?

GitCommands/Git/GPGSecretKeysParser.cs
```diff
@@ -0,0 +1,61 @@
...
+                                let u = users.First(ur => ur.LineNumber > k.LineNumber)
+                                select new { Key = k, Fingerprint = f, User = u }).ToArray();
+                long tmp = 0;
+
+                var output = combined.Select((k, i) => new GpgKeyInfo(
+                    k.Fingerprint.Fields.ElementAt(9).Value,
+                    k.Key.Fields.ElementAt(4).Value,
+                    k.User.Fields.ElementAt(9).Value,
+                    string.IsNullOrEmpty(k.Key.Fields.ElementAt(6).Value) ? null : DateTimeOffset.FromUnixTimeSeconds(long.TryParse(k.Key.Fields.ElementAt(6).Value, out tmp) ? tmp : 0L),
+                    !string.IsNullOrEmpty(defaultKey) && k.Key.Fields.ElementAt(4).Value == defaultKey));
```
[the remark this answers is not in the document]
vbjay: Yep. Once I made sure I didn't have to use any other fields I wanted to replace with a constants like GPGFieldIndexes.Expired or something of that nature for all the indexes.


https://github.com/CSNW/gnupg/blob/master/doc/DETAILS

gerhardol:
Some comments from my initial review before I look at this again:
https://github.com/gitextensions/gitextensions/pull/10473#pullrequestreview-1473500029

> This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).

GitUI/CommandsDialogs/FormRebase.Designer.cs
vbjay: So no.  I will not be trying to fix something that was changed by vs.  https://devblogs.microsoft.com/dotnet/winforms-codegen-update/   Yes I tried to folow editorconfig changes to see if it would fix it.  To no avail.  I am not interested in trying to fix something that is autogenerated.  If you want to figure out the right editorconfig that solves the problem and follows the right way, then great.

vbjay:
> Some comments from my initial review before I look at this again: [#10473 (review)](https://github.com/gitextensions/gitextensions/pull/10473#pullrequestreview-1473500029)
> 
> > This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
> > Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).

I am trying to make it reflect what will happen when committing.  As in if user has commit/tag.GpgSign set then it will automatically say using default key in the right ui.  Same with what key.  For this to be reliable it MUST fetch to the minute config.  Same with the idea of getting what keys exist at the moment.  I am sorry but 108MS is nothing and I don't see the issue.

gerhardol:
> > Some comments from my initial review before I look at this again: [#10473 (review)](https://github.com/gitextensions/gitextensions/pull/10473#pullrequestreview-1473500029)
> > > This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
> > > Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).
> 
> I am trying to make it reflect what will happen when committing. As in if user has commit/tag.GpgSign set then it will automatically say using default key in the right ui. Same with what key. For this to be reliable it MUST fetch to the minute config. Same with the idea of getting what keys exist at the moment. I am sorry but 108MS is nothing and I don't see the issue.

180 ms is certainly noticeable (and this may be a lot more in other setups). Some more features like this and a few seconds are added. If you use the functions, the delay is fine, but irritating otherwise.
At least limit the added delay to one command instead of three commands, probably a setting too.

vbjay:
> > > Some comments from my initial review before I look at this again: [#10473 (review)](https://github.com/gitextensions/gitextensions/pull/10473#pullrequestreview-1473500029)
> > > > This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
> > > > Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).
> > 
> > 
> > I am trying to make it reflect what will happen when committing. As in if user has commit/tag.GpgSign set then it will automatically say using default key in the right ui. Same with what key. For this to be reliable it MUST fetch to the minute config. Same with the idea of getting what keys exist at the moment. I am sorry but 108MS is nothing and I don't see the issue.
> 
> 180 ms is certainly noticeable (and this may be a lot more in other setups). Some more features like this and a few seconds are added. If you use the functions, the delay is fine, but irritating otherwise. At least limit the added delay to one command instead of three commands, probably a setting too.

So basically neuter the feature behind a flag that will never get set like it took forever for us to turn on the show gpg tab before?    KInda irritated with your feedback.  You are basically complaining because I am adding functionality that it take more time to load.  I reject this idea of neutering behind a flag because then it will fail to reflect what will happen and is kinda the whole point of this feature.  It is to not only make it easy to manage the signing in the ui and also provide visibility into what will happen when git is told to create the commit or tag.

1.  I am exposing visibility into what git will do. with the commit/tag.Commit.GpgSign reading and setting the dropdown to show the initial state
2. I am reading the default key to know if one is selected at the time of run of code.  So that if someone went in and changed it in cli or eventually in settings dalog
3. I have to load the keys so that the dropdown has values to show and reflects this key is being used
4. The config of DEFAULT sign the commit/tag is driven by the commit.GpgSign config value

vbjay:
> > > > Some comments from my initial review before I look at this again: [#10473 (review)](https://github.com/gitextensions/gitextensions/pull/10473#pullrequestreview-1473500029)
> > > > > This adds some time to the startup of these forms: Two git-config and gpg.exe (without any keys for me). This adds like 180 ms for me. Probably not much more in big repos, but in slower file systems.
> > > > > Still, gpg signing should maybe be something that is enabled. I do not see an obvious way to do this with the Git configuration. Maybe if there are no keys, do not bother with the git-config (2*55ms). Or read the keys if signing is requested (including if e.g. commit.gpgSign is set).
> > > 
> > > 
> > > I am trying to make it reflect what will happen when committing. As in if user has commit/tag.GpgSign set then it will automatically say using default key in the right ui. Same with what key. For this to be reliable it MUST fetch to the minute config. Same with the idea of getting what keys exist at the moment. I am sorry but 108MS is nothing and I don't see the issue.
> > 
> > 
> > 180 ms is certainly noticeable (and this may be a lot more in other setups). Some more features like this and a few seconds are added. If you use the functions, the delay is fine, but irritating otherwise. At least limit the added delay to one command instead of three commands, probably a setting too.
> 
> So basically neuter the feature behind a flag that will never get set like it took forever for us to turn on the show gpg tab before? KInda irritated with your feedback. You are basically complaining because I am adding functionality that it take more time to load. I reject this idea of neutering behind a flag because then it will fail to reflect what will happen and is kinda the whole point of this feature. It is to not only make it easy to manage the signing in the ui and also provide visibility into what will happen when git is told to create the commit or tag.
> 
> 1. I am exposing visibility into what git will do. with the commit/tag.Commit.GpgSign reading and setting the dropdown to show the initial state
> 2. I am reading the default key to know if one is selected at the time of run of code.  So that if someone went in and changed it in cli or eventually in settings dalog
> 3. I have to load the keys so that the dropdown has values to show and reflects this key is being used
> 4. The config of DEFAULT sign the commit/tag is driven by the commit.GpgSign config value

What I can do is maybe see if I can put some of the setup in onshown so it doesn't affect first show time.

mstv:
> What I can do is maybe see if I can put some of the setup in onshown so it doesn't affect first show time.

Why not async (as for user name & e-mail)?

vbjay:
> > What I can do is maybe see if I can put some of the setup in onshown so it doesn't affect first show time.
> 
> Why not async (as for user name & e-mail)?

Sure.  Just need to figure ot a way to show hey still loading.  

just ran sandbox and told it to generate 1000 keys along with the 3 it already generated and messageboxed the load  time
image
image

image
```patch
From fb01bbf4f981bfca4515b950ec28606e68a9a633 Mon Sep 17 00:00:00 2001
From: Jay Asbury <vbjay.net@gmail.com>
Date: Sat, 5 Aug 2023 12:05:05 -0400
Subject: [PATCH] test

---
 GitCommands/Gpg/GPGLine.cs          | 104 ++++++++++---------
 GitCommands/Gpg/GpgKeyInfo.cs       | 156 ++++++++--------------------
 GitUI/CommandsDialogs/FormCommit.cs |  29 ++++--
 scripts/wsb/setup.ps1               |   8 ++
 4 files changed, 126 insertions(+), 171 deletions(-)
 rewrite GitCommands/Gpg/GpgKeyInfo.cs (78%)

diff --git a/GitCommands/Gpg/GPGLine.cs b/GitCommands/Gpg/GPGLine.cs
index 57528c499..5aa38c3b8 100644
--- a/GitCommands/Gpg/GPGLine.cs
+++ b/GitCommands/Gpg/GPGLine.cs
@@ -1,19 +1,39 @@
 ﻿namespace GitCommands.Gpg
 {
-    internal sealed partial class GPGLine
+    // Convert the following class to a record struct when C# 10 is released
+
+    /// <summary>
+    /// Models the records from running gpg -K --with-colons.  See <see href="https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob_plain;f=doc/DETAILS"/>
+    /// </summary>
+    internal record struct GPGLine()
     {
+        #region Properties
+        public bool Disabled { get; init; } = false;
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
+
+#pragma warning restore SA1313 // Parameter names should begin with lower-case letter
+        public DateTimeOffset? ExpirationDate { get; init; } = null;
+        public readonly record struct Field(int Index, string Value);
+        public IEnumerable<Field> Fields { get; init; } = Enumerable.Empty<Field>();
+        public Capabilities LineCapabilities { get; init; } = Capabilities.None;
+        public int LineNumber { get; init; } = -1;
+        public LineTypes LineType { get; init; } = LineTypes.Other;
+
+        public Validities Validity { get; init; } = Validities.Unknown;
+        #endregion
+
         /// <summary>
         /// Models the records from running gpg -K --with-colons.  See <see href="https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob_plain;f=doc/DETAILS"/>
         /// </summary>
         /// <param name="lineNumber">Line row number</param>
         /// <param name="fields">The : split fields from the line</param>
-        public GPGLine(int lineNumber, IEnumerable<Field> fields)
+        public GPGLine(int lineNumber, IEnumerable<Field> fields) : this()
         {
             LineNumber = lineNumber;
             Fields = fields;
             LineType = DetermineLineType(fields?.FirstOrDefault().Value ?? "");
             Validity = DetermineLineValidity(fields?.Skip(FieldIndexes.Validity)?.FirstOrDefault().Value ?? "");
-            Disabled = fields?.Skip(FieldIndexes.KeyCapabilities)?.FirstOrDefault().Value?.Contains('D') ?? false;
+            Disabled = fields?.Skip(FieldIndexes.KeyCapabilities)?.FirstOrDefault().Value?.ToUpper().Contains('D') ?? false;
             LineCapabilities = DetermineCapabilities(fields?.Skip(FieldIndexes.KeyCapabilities)?.FirstOrDefault().Value ?? "");
             ExpirationDate = DettermineExpiration(fields?.Skip(FieldIndexes.ExpirationDate)?.FirstOrDefault().Value ?? "");
         }
@@ -33,7 +53,7 @@ public sealed class FieldIndexes
 
         #region RecordType
         /*
-*** Field 1 - Type of record
+    *** Field 1 - Type of record
 
     - pub :: Public key
     - crt :: X.509 certificate
@@ -82,7 +102,7 @@ private static LineTypes DetermineLineType(string lineTypeKey)
         /*
         Records marked with an asterisk are described at [[*Special%20field%20formats][*Special fields]].
 
-*** Field 2 - Validity
+    *** Field 2 - Validity
 
     This is a letter describing the computed validity of a key.
     Currently this is a single letter, but be prepared that additional
@@ -126,7 +146,7 @@ private static Validities DetermineLineValidity(string validityKey)
         /*
     If the validity information is given for a UID or UAT record, it
     describes the validity calculated based on this user ID.  If given
-for a key record it describes the validity taken from the best
+    for a key record it describes the validity taken from the best
     rated user ID.
 
 
@@ -134,37 +154,37 @@ private static Validities DetermineLineValidity(string validityKey)
     certificate (i.e. for the trust anchor) and an 'f' for all other
     valid certificates.
 
-In "sig" records, this field may have one of these values as first
+    In "sig" records, this field may have one of these values as first
     character:
 
-- ! :: Signature is good.
-- - :: Signature is bad.
-- ? :: No public key to verify signature or public key is not usable.
-- % :: Other error verifying a signature
+    - ! :: Signature is good.
+    - - :: Signature is bad.
+    - ? :: No public key to verify signature or public key is not usable.
+    - % :: Other error verifying a signature
 
-More values may be added later.The field may also be empty if
-gpg has been invoked in a non-checking mode (--list-sigs) or in a
-fast checking mode.Since 2.2.7 '?' will also be printed by the
-command --list-sigs if the key is not in the local keyring.
+    More values may be added later.The field may also be empty if
+    gpg has been invoked in a non-checking mode (--list-sigs) or in a
+    fast checking mode.Since 2.2.7 '?' will also be printed by the
+    command --list-sigs if the key is not in the local keyring.
     */
         #endregion
         /*
 
-*** Field 3 - Key length
+    *** Field 3 - Key length
 
     The length of key in bits.
 
-*** Field 4 - Public key algorithm
+    *** Field 4 - Public key algorithm
 
     The values here are those from the OpenPGP specs or if they are
     greater than 255 the algorithm ids as used by Libgcrypt.
 
-*** Field 5 - KeyID
+    *** Field 5 - KeyID
 
     This is the 64 bit keyid as specified by OpenPGP and the last 64
     bit of the SHA-1 fingerprint of an X.509 certifciate.
 
-*** Field 6 - Creation date
+    *** Field 6 - Creation date
 
     The creation date of the key is given in UTC.For UID and UAT
     records, this is used for the self-signature date.  Note that the
@@ -175,7 +195,7 @@ valid certificates.
     without using the =--fixed-list - mode = option used a "yyyy-mm-tt"
     format.
 
-* **Field 7 - Expiration date
+    * **Field 7 - Expiration date
 
     Key or UID / UAT expiration date or empty if it does not expire.
          */
@@ -192,21 +212,21 @@ valid certificates.
 
         /*
 
-* **Field 8 - Certificate S / N, UID hash, trust signature info
+    * **Field 8 - Certificate S / N, UID hash, trust signature info
 
     Used for serial number in crt records.For UID and UAT records,
     this is a hash of the user ID contents used to represent that
     exact user ID.For trust signatures, this is the trust depth
     separated by the trust value by a space.
 
-* **Field 9 - Ownertrust
+    * **Field 9 - Ownertrust
 
     This is only used on primary keys.This is a single letter, but
     be prepared that additional information may follow in future
     versions.For trust signatures with a regular expression, this is
     the regular expression value, quoted as in field 10.
 
-* **Field 10 - User - ID
+    * **Field 10 - User - ID
 
     The value is quoted like a C string to avoid control characters
     (the colon is quoted =\x3a =).For a "pub" record this field is
@@ -216,7 +236,7 @@ valid certificates.
     records store the fingerprints here.The fingerprint of a
     revocation key is stored here.
 
-***Field 11 - Signature class
+    ***Field 11 - Signature class
 
     Signature class as per RFC-4880.  This is a 2 digit hexnumber
     followed by either the letter 'x' for an exportable signature or
@@ -230,7 +250,7 @@ valid certificates.
         */
         #region KeyCapabilities
         /*
-*** Field 12 - Key capabilities
+    *** Field 12 - Key capabilities
 
     The defined capabilities are:
 
@@ -265,7 +285,7 @@ private static Capabilities DetermineCapabilities(string capabilities)
         #endregion
         /*
 
-*** Field 13 - Issuer certificate fingerprint or other info
+    *** Field 13 - Issuer certificate fingerprint or other info
 
     Used in FPR records for S/MIME keys to store the fingerprint of
     the issuer certificate.  This is useful to build the certificate
@@ -286,28 +306,28 @@ private static Capabilities DetermineCapabilities(string capabilities)
     if the key is missing but the signature carries an issuer
     fingerprint as meta data.
 
-*** Field 14 - Flag field
+    *** Field 14 - Flag field
 
     Flag field used in the --edit-key menu output
 
-*** Field 15 - S/N of a token
+    *** Field 15 - S/N of a token
 
     Used in sec/ssb to print the serial number of a token (internal
     protect mode 1002) or a '#' if that key is a simple stub(internal
     protect mode 1001).  If the option --with-secret is used and a
     secret key is available for the public key, a '+' indicates this.
 
-*** Field 16 - Hash algorithm
+    *** Field 16 - Hash algorithm
 
     For sig records, this is the used hash algorithm.  For example:
     2 = SHA-1, 8 = SHA-256.
 
-*** Field 17 - Curve name
+    *** Field 17 - Curve name
 
     For pub, sub, sec, ssb, crt, and crs records this field is used
     for the ECC curve name.
 
-*** Field 18 - Compliance flags
+    *** Field 18 - Compliance flags
 
     Space separated list of asserted compliance modes and
     screening result for this key.
@@ -318,7 +338,7 @@ private static Capabilities DetermineCapabilities(string capabilities)
     - 23 :: The key is compliant with compliance mode "de-vs".
     - 6001 :: Screening hit on the ROCA vulnerability.
 
-*** Field 19 - Last update
+    *** Field 19 - Last update
 
     The timestamp of the last update of a key or user ID.  The update
     time of a key is defined a lookup of the key via its unique
@@ -326,31 +346,17 @@ private static Capabilities DetermineCapabilities(string capabilities)
     update time of a user ID is defined by a lookup of the key using a
     trusted mapping from mail address to key.
 
-*** Field 20 - Origin
+    *** Field 20 - Origin
 
     The origin of the key or the user ID.This is an integer
     optionally followed by a space and an URL.  This goes along with
     the previous field.The URL is quoted in C style.
 
-*** Field 21 - Comment
+    *** Field 21 - Comment
 
     This is currently only used in "rev" and "rvs" records to carry
     the the comment field of the recocation reason.The value is
     quoted in C style.
- */
-        #region Properties
-        public bool Disabled { get; private set; }
-#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
-        public readonly record struct Field(int Index, string Value);
-#pragma warning restore SA1313 // Parameter names should begin with lower-case letter
-        public DateTimeOffset? ExpirationDate { get; private set; }
-        public IEnumerable<Field> Fields { get; private set; }
-        public Capabilities LineCapabilities { get; private set; }
-        public int LineNumber { get; private set; }
-        public LineTypes LineType { get; private set; }
-
-        public Validities Validity { get; private set; }
-
-        #endregion
+    */
     }
 }
diff --git a/GitCommands/Gpg/GpgKeyInfo.cs b/GitCommands/Gpg/GpgKeyInfo.cs
dissimilarity index 78%
index b53df81b2..ee0731358 100644
--- a/GitCommands/Gpg/GpgKeyInfo.cs
+++ b/GitCommands/Gpg/GpgKeyInfo.cs
@@ -1,112 +1,44 @@
-﻿namespace GitCommands.Gpg
-{
-    public class GpgKeyInfo : IEquatable<GpgKeyInfo>
-    {
-        private DateTimeOffset? _expires;
-
-        public GpgKeyInfo(
-            string fingerprint,
-            string keyID,
-            string userID,
-            DateTimeOffset? expires,
-            Capabilities capabilities,
-            Validities validities,
-            bool disabled,
-            bool isDefault = false)
-        {
-            Fingerprint = fingerprint;
-            KeyID = keyID;
-            UserID = userID;
-            Expires = expires;
-            Capabilities = capabilities;
-            Validities = validities;
-            IsDefault = isDefault;
-            Disabled = disabled;
-        }
-
-        public override string ToString() => $"{UserID} ({KeyID})";
-
-        public Capabilities Capabilities { get; private set; }
-        public bool Disabled { get; private set; }
-
-        public DateTimeOffset? Expires { get => _expires; private set => _expires = value?.ToLocalTime(); }
-
-        public string Fingerprint { get; private set; }
-
-        public bool IsDefault { get; private set; } = false;
-
-        public string KeyID { get; private set; }
-
-        public string UserID { get; private set; }
-
-        public Validities Validities { get; private set; }
-
-        #region Equality
-        public static bool operator !=(GpgKeyInfo first, GpgKeyInfo second)
-        {
-            return !(first == second);
-        }
-
-        public static bool operator ==(GpgKeyInfo first, GpgKeyInfo second)
-        {
-            if ((object)first == null)
-            {
-                return (object)second == null;
-            }
-
-            return first.Equals(second);
-        }
-
-        public override bool Equals(object obj)
-        {
-            if (obj is GpgKeyInfo gpgKeyInfo)
-            {
-                return Equals(gpgKeyInfo);
-            }
-
-            return base.Equals(obj);
-        }
-
-        public bool Equals(GpgKeyInfo other)
-        {
-            if (ReferenceEquals(null, other))
-            {
-                return false;
-            }
-
-            if (ReferenceEquals(this, other))
-            {
-                return true;
-            }
-
-            return Equals(Fingerprint, other.Fingerprint)
-                   && Equals(KeyID, other.KeyID)
-                   && Equals(UserID, other.UserID);
-        }
-
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                int hashCode = 47;
-                if (Fingerprint != null)
-                {
-                    hashCode = (hashCode * 53) ^ EqualityComparer<string>.Default.GetHashCode(Fingerprint);
-                }
-
-                if (KeyID != null)
-                {
-                    hashCode = (hashCode * 53) ^ EqualityComparer<string>.Default.GetHashCode(KeyID);
-                }
-
-                if (UserID != null)
-                {
-                    hashCode = (hashCode * 53) ^ EqualityComparer<string>.Default.GetHashCode(UserID);
-                }
-
-                return hashCode;
-            }
-        }
-        #endregion
-    }
-}
+﻿namespace GitCommands.Gpg
+{
+    public readonly record struct GpgKeyInfo : IEquatable<GpgKeyInfo>
+    {
+        private readonly DateTimeOffset? _expires;
+
+        public GpgKeyInfo(
+            string fingerprint,
+            string keyID,
+            string userID,
+            DateTimeOffset? expires,
+            Capabilities capabilities,
+            Validities validities,
+            bool disabled,
+            bool isDefault = false) : this()
+        {
+            Fingerprint = fingerprint;
+            KeyID = keyID;
+            UserID = userID;
+            Expires = expires;
+            Capabilities = capabilities;
+            Validities = validities;
+            IsDefault = isDefault;
+            Disabled = disabled;
+        }
+
+        public override string ToString() => $"{UserID} ({KeyID})";
+
+        public Capabilities Capabilities { get; init; }
+        public bool Disabled { get; init; }
+
+        public DateTimeOffset? Expires { get => _expires; init => _expires = value?.ToLocalTime(); }
+
+        public string Fingerprint { get; init; }
+
+        public bool IsDefault { get; init; } = false;
+
+        public string KeyID { get; init; }
+
+        public string UserID { get; init; }
+
+        public Validities Validities { get; init; }
+    }
+}
index ac9882b48..2f1eda7f0 100644
@@ -156,6 +156,8 @@ public sealed partial class FormCommit : GitModuleForm
         private TranslationString _invalidSignCaption = new("Commit Signing Options");
         #endregion
 
+        private Stopwatch _st = new();
+
         private event Action? OnStageAreaLoaded;
 
         private readonly ICommitTemplateManager _commitTemplateManager;
@@ -347,8 +349,6 @@ public FormCommit(GitUICommands commands, CommitKind commitKind = CommitKind.Nor
 
             InitializeComplete();
 
-            toolStripGpgKeyComboBox.UICommandsSource = this;
-
             // By calling this in the constructor, we prevent flickering caused by resizing the
             // form, for example when it is restored to maximised, but first drawn as a smaller window.
             RestorePosition();
@@ -514,14 +514,6 @@ protected override void OnLoad(EventArgs e)
             // The problem is likely caused by 'splitRight.FixedPanel = FixedPanel.Panel2' fact, but other forms
             // have the same setting, and don't appear to suffer from the same bug.
             splitRight.SplitterDistance -= 6;
-
-            // If not changed, by default show "no sign commit"
-            if (gpgSignCommitToolStripComboBox.SelectedIndex == -1)
-            {
-                var gpg = toolStripGpgKeyComboBox.KeysUIController;
-                bool commitSign = gpg.GetCommitGPGSign();
-                gpgSignCommitToolStripComboBox.SelectedIndex = commitSign ? 1 : 0;
-            }
         }
 
         protected override void OnShown(EventArgs e)
@@ -571,6 +563,8 @@ protected override void OnShown(EventArgs e)
             }
 
             base.OnShown(e);
+            Debugger.Log(0, "GpgInfo", $"GPG stuff: {_st.ElapsedMilliseconds}ms{Environment.NewLine}");
+            MessageBox.Show(this, $"GPG stuff: {_st.ElapsedMilliseconds}ms", "GPG", MessageBoxButtons.OK, MessageBoxIcon.Information);
 
             return;
 
@@ -1021,6 +1015,21 @@ private async Task UpdateBranchNameDisplayAsync()
 
         private void Initialize(bool loadUnstaged = true)
         {
+            if (!_initialized)
+            {
+                _st.Start();
+
+                // If not changed, by default show "no sign commit"
+
+                toolStripGpgKeyComboBox.UICommandsSource = this;
+
+                GPGKeysUIController gpg = toolStripGpgKeyComboBox.KeysUIController;
+                bool commitSign = gpg.GetCommitGPGSign();
+                gpgSignCommitToolStripComboBox.SelectedIndex = commitSign ? 1 : 0;
+
+                _st.Stop();
+            }
+
             _initialized = true;
 
             ThreadHelper.JoinableTaskFactory.RunAsync(() => UpdateBranchNameDisplayAsync());
diff --git a/scripts/wsb/setup.ps1 b/scripts/wsb/setup.ps1
index 866286a3e..6fa0c74e4 100644
--- a/scripts/wsb/setup.ps1
+++ b/scripts/wsb/setup.ps1
@@ -55,6 +55,14 @@ $expired = $expired -replace 'Name-Real.*', 'Name-Real: GitExtensions Tester Sho
 $expired | Set-Content -Path $env:USERPROFILE\Desktop\Expired.dat
 gpg --batch --generate-key $env:USERPROFILE\Desktop\Expired.dat 2>&1
 
+$expired = Get-Content .\wsb\gen-gpg.dat 
+for (($i = 0); $i -lt 1000; $i++)
+{
+    $expired = $expired -replace 'Name-Real.*', "Name-Real: GitExtensions Tester $i"
+    $expired | Set-Content -Path $env:USERPROFILE\Desktop\Tons.dat
+    gpg --batch --generate-key $env:USERPROFILE\Desktop\Tons.dat 2>&1
+}
+
 $keyLines = gpg -K --with-colons | awk -F: '/^sec:/ { print $5 }'
 
 Set-Variable -Name 'keys' -Scope 'script' -Value @{
-- 
2.41.0.windows.1

```

So yes variable timing based on count of keys.

- Async... just need to hook into a progress of loading to show hey loading it
- some kind of key cache that can be forced to clear and refresh by user or by some kind of timeout could help also

I wanted to go async with it but had to test it without to deal with the basic functionality. Yes I did not want to run all of it in ui thread.  

Thoughts on timeout of cache?  10 minutes too long?

vbjay:
>...
> * Async... just need to hook into a progress of loading to show hey loading it
> * some kind of key cache that can be forced to clear and refresh by user or by some kind of timeout could help also
> 
> I wanted to go async with it but had to test it without to deal with the basic functionality. Yes I did not want to run all of it in ui thread.
> 
> Thoughts on timeout of cache? 10 minutes too long?

I tried it out with a MemoryCache

image
and then open commit form again quickly 
image
 so I can at least make it so that if the keys were fetched recently, the cache will return the value but if not, it will run gpg again.  I can play with the expiry timing and come up with a good value.

GitCommands/Git/Executable.cs
RussKie: If this is a debug only, should this be wrapped with `#if DEBUG` too?

GitCommands/Gpg/IGitGpgController.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace GitCommands.Git.Gpg
+{
+    public partial interface IGitGpgController
+    {
+        /// <summary>
+        /// Gets whether commits are signed by default
+        /// </summary>
+        /// <returns>true for default to signed commits</returns>
+        bool GetCommitGPGSign();
```
RussKie, line 11: Would this work?
```suggestion
        bool AreCommitSignedByDefault();
```

GitCommands/Gpg/IGitGpgController.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <returns>The key id of the gpg key to sign with.</returns>
+        string GetDefaultKey();
+
+        Task<IEnumerable<GpgKeyInfo>> GetGpgSecretKeysAsync();
+
+        /// <summary>
+        /// Gets whether tags are signed by default
+        /// </summary>
+        /// <returns>true for default to signed tags</returns>
+        bool GetTagGPGSign();
```
RussKie, line 25:
```suggestion
        bool AreTagsSignedByDefault();
```

GitCommands/Gpg/IGitGpgController.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <returns>true for default to signed commits</returns>
+        bool GetCommitGPGSign();
+
+        /// <summary>
+        /// Gets the user's default signing key from git config
+        /// </summary>
+        /// <returns>The key id of the gpg key to sign with.</returns>
+        string GetDefaultKey();
+
+        Task<IEnumerable<GpgKeyInfo>> GetGpgSecretKeysAsync();
```
RussKie, line 19: Need xml-doc.

RussKie, line 19: I had a look at the implementation and I have a feeling these names don't clearly communicate the purpose. `GetDefaultKey` appears to load the configured setting, whilst `GetGpgSecretKeysAsync` is actually loading and parsing keys.

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -1325,6 +1332,16 @@ void DoCommit()
...
 
+                GPGKeysUIController gpg = toolStripGpgKeyComboBox.KeysUIController;
+                if (!gpg.ValidateCommitSign(gpgSignCommitToolStripComboBox.SelectedIndex > 0, toolStripGpgKeyComboBox.KeyID))
+                {
+                    MessageBox.Show(this, TranslatedStrings.InvalidGpgSignOptions, _invalidSignCaption.Text, MessageBoxButtons.OK, MessageBoxIcon.Asterisk);
+                    toolStripMenuItem3.ShowDropDown();
+                    gpgSignCommitToolStripComboBox.Focus();
+
+                    return;
+                }
```
RussKie, line 1343: This looks as though we're leaking implementation details. This should be a validation event on the combobox itself.

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
RussKie: For example:
https://github.com/gitextensions/gitextensions/blob/99ece765b4a76a58b1d3160f914d9c6f0ab5581f/GitUI/CommandsDialogs/FormBrowse.cs#L266-L268
https://github.com/gitextensions/gitextensions/blob/99ece765b4a76a58b1d3160f914d9c6f0ab5581f/GitUI/CommandsDialogs/FormBrowse.cs#L298

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
```diff
@@ -0,0 +1,128 @@
...
+        public bool IsDefaultSelected => _loadedKeys && SelectedIndex >= 0 && (comboBoxKeys.SelectedItem as GpgKeyDisplayInfo).IsDefault;
+        public GPGSecretKeysCombobox() : base()
+        {
+            InitializeComponent();
+            InitializeComplete();
+        }
+
+        private void InitControllers()
+        {
+            _secretKeysParser = new GPGSecretKeysParser();
```
RussKie, line 43: We should avoid newing up concrete implementations as that makes the testing story much more complicated. Absent of a DI, this should be passed in via `Bind()` or `Initialize` mechanisms as implemented in other parts of the codebase.

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System.ComponentModel;
+using System.Diagnostics;
+using GitCommands.Git.Gpg;
+using GitCommands.Gpg;
+
+namespace GitUI.UserControls.GPGKeys
+{
+    public partial class GPGSecretKeysCombobox : GitModuleControl
```
RussKie, line 8: This isn't a combobox control, but a custom control that contains a combobox.
We don't have a consistent naming strategy for custom controls, but based on what we already have I think we should probably name it something like:
```suggestion
    public partial class GpgSecretKeysComboBoxControl : GitModuleControl
```
or
```suggestion
    public partial class GpgSecretKeysSelector : GitModuleControl
```

GitUI/UserControls/GPGKeys/ToolStripGPGSecretKeysComboBox.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace GitUI.UserControls.GPGKeys
+{
+    public class ToolStripGPGSecretKeysComboBox : ToolStripControlHost
```
RussKie, line 3: Ditto

GitCommands/Git/Executable.cs
vbjay: I can make it internal.  But no I can't just exclude it with a #if.  The point of the lambda is to provide a way for the Executable to get the correct program to run from the Gpg.Program git config value which can be set in local repo config, along with the user may have changed the value and had a running instance of GE.  So the lamba provides a way to get the gonfig and figure out the executable to run at the time of needing to run.  

https://github.com/gitextensions/gitextensions/pull/10473/files#diff-862781f8feb70c499c04cf96942cb7fc5b82624e8eec00e5a9db7f8b5421b954R181-R191

GitCommands/Gpg/GitGpgController.cs
```diff
@@ -0,0 +1,66 @@
+﻿using GitCommands.Config;
+using GitCommands.Gpg;
+using GitExtUtils;
+
+namespace GitCommands.Git.Gpg
+{
+    partial class GitGpgController : IGitGpgController
```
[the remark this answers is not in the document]
vbjay: The class is in the Git namespace and is the driver between Git and Gpg functionality.  

image

I split so that the implimentation of the gpg key functionality is defined with  the gpg key providing functionality.  I was thinking of splitting into own class but am 2 ways on it.  Outside the user sees a cohesive this thing drives Git/Gpg interaction and the developers can navigate to the right pieces easily.   
image

GitCommands/Gpg/IGitGpgController.cs
vbjay: Yes GetDefaultKey is how it gets which key to use from git config so that the * can be placed in the listing along with the default being selected if there is one set.  GetSeceretKeys is the go get the list of keys from GPG.  

https://git-scm.com/docs/git-config#Documentation/git-config.txt-usersigningKey and https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work 

I can rename to GetUserSigningKey if you want.  No I did not implement the ssh functionality in this part yet.

>user.signingKey
If [git-tag[1]](https://git-scm.com/docs/git-tag) or [git-commit[1]](https://git-scm.com/docs/git-commit) is not selecting the key you want it to automatically when creating a signed tag or commit, you can override the default selection with this variable. This option is passed unchanged to gpg’s --local-user parameter, so you may specify a key using any method that gpg supports. If gpg.format is set to ssh this can contain the path to either your private ssh key or the public key when ssh-agent is used. Alternatively it can contain a public key prefixed with key:: directly (e.g.: "key::ssh-rsa XXXXXX identifier"). The private key needs to be available via ssh-agent. If not set git will call gpg.ssh.defaultKeyCommand (e.g.: "ssh-add -L") and try to use the first key available. For backward compatibility, a raw key which begins with "ssh-", such as "ssh-rsa XXXXXX identifier", is treated as "key::ssh-rsa XXXXXX identifier", but this form is deprecated; use the key:: form instead.

GitCommands/Gpg/IGitGpgController.cs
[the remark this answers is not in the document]
vbjay: Can do.

GitUI/CommandsDialogs/FormCommit.cs
vbjay: It is a validation of 2 fields.  Index 0 is Do not sign Commit 1 and 2 are sign with default or sign with selected key.  This validation is a you can't commit with these 2 values of these fields.  The cases this covers are

- User has configured the commit.gpgSign value to true but does not have a default key so the key combobox will be no key selected.  Since git config is not fully in our hands this state is possible
    - would result in state of `git commit ... -S` along with the user.signingKey not being set. Git would complain and fail the commit.  
- User has selected sign with specific key and then failed to select a key.
image

invalid selection
User can skip the options menu completely.  So I disagree with the validation event.  Where the validation needs to occur is  when they hit commit.  
The validation logic is the same for tag.gpgSign also.  Along with the validation being coded once in the controller and the ui needing to use it can do what it needs when it detects the state.

GitCommands/Gpg/GitGpgController.cs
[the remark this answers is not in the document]
vbjay: https://github.com/gitextensions/gitextensions/pull/10473/files#r1285221729

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
vbjay: I like selector better.

GitUI/UserControls/GPGKeys/GPGSecretKeysCombobox.cs
vbjay: Ok.  WIll look into.  Yes. DI would have helped a ton in this.

GitUI/UserControls/GPGKeys/GpgKeyDisplayInfo.cs
vbjay: Started work on this.

GitCommands/Git/ExecutableExtensions.cs
```diff
@@ -110,15 +110,15 @@ public static class ExecutableExtensions
             if (input is not null)
             {
 #if DEBUG
-                System.Diagnostics.Debug.WriteLine($"git {arguments} {Encoding.UTF8.GetString(input)}");
+                System.Diagnostics.Debug.WriteLine($"{executable.FileNameProvider()} {arguments} {Encoding.UTF8.GetString(input)}");
```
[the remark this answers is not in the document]
vbjay: The lambda is there to run at time of start.  Kinda defeats the whole reason I did it.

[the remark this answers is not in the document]
vbjay: Sure.

GitCommands/Gpg/GitGpgController.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            string? output = GetModule().GetEffectiveGitSetting(SettingKeyString.TagGPGSign, false).Value;
+
+            bool result = false;
+            if (bool.TryParse(output, out result))
+            {
+                return result;
+            }
+
+            return false;
```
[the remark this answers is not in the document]
vbjay: I get what you mean but github doesn't make your dittos easy to folow what you mean.  Can you at least in future ditto review comments put a link to the message you are dittoing from.   Like this

image

ditto https://github.com/gitextensions/gitextensions/pull/10473/files#r1285179455

RussKie:

PR #10484, opened 2022-12-02
Report git fatal errors as not app errors

### LINKED ISSUE #10483: Improve handling of git 'fatal' errors
### Environment

v4 and the current master

### Issue description

Git "fatal" errors predominantly indicate issues with local git repository, in other words, they indicate issues that are within the realms of a user, and very rarely something attributable to the app itself. For example:
image

The app should handle these errors differently, and whenever one is encountered the error message dialog should not readily allow to open an issue here but rather tell the user to fix the issue. 

Related: https://github.com/gitextensions/gitextensions/issues/9954

RussKie:
Resolves #10483

## Proposed changes

Report git "fatal" errors as not app errors, that is, without links to raise issues on GitHub, because such issues are generally local-repo specific and should be addressable by the user.

Summary changes for Git errors:
* Do not add test "submit report for GE to investigate".
* Change *Report bug* to *View details*

It is still possible to report to GitHub with _View details_.

### Before

image

### After

image

## Merge strategy

----

### DIFF
--- a/GitUI/NBugReports/BugReportInvoker.cs
+++ b/GitUI/NBugReports/BugReportInvoker.cs
@@ -103,7 +103,9 @@ namespace GitUI.NBugReports
                 LogError(exception, isTerminating);
             }
 
-            if (exception is ExternalOperationException externalOperationException
+            ExternalOperationException externalOperationException = exception as ExternalOperationException;
+
+            if (externalOperationException is not null
                 && externalOperationException.InnerException is { }
                 && externalOperationException.InnerException.Message.Contains(_dubiousOwnershipSecurityConfigString))
             {
@@ -132,8 +134,9 @@ namespace GitUI.NBugReports
                                                  or PathTooLongException
                                                  or Win32Exception;
 
-            // Treat all git fatal errors (e.g., those starting with "fatal: ") as user issues
-            if (exception.InnerException?.Message?.StartsWith("fatal: ") ?? false)
+            // Treat all git errors as user issues
+            if (string.Equals(AppSettings.GitCommand, externalOperationException?.Command, StringComparison.InvariantCultureIgnoreCase)
+             || string.Equals(AppSettings.WslGitCommand, externalOperationException?.Command, StringComparison.InvariantCultureIgnoreCase))
             {
                 isUserExternalOperation = true;
             }
--- a/GitUI/NBugReports/BugReportInvoker.cs
+++ b/GitUI/NBugReports/BugReportInvoker.cs
@@ -132,6 +132,12 @@ namespace GitUI.NBugReports
                                                  or PathTooLongException
                                                  or Win32Exception;
 
+            // Treat all git fatal errors (e.g., those starting with "fatal: ") as user issues
+            if (exception.InnerException?.Message?.StartsWith("fatal: ") ?? false)
+            {
+                isUserExternalOperation = true;
+            }
+
             StringBuilder text = GetExceptionInfo(exception);
             string rootError = GetRootError(exception);
 
--- a/GitUI/NBugReports/BugReportInvoker.cs
+++ b/GitUI/NBugReports/BugReportInvoker.cs
@@ -105,9 +105,7 @@ namespace GitUI.NBugReports
 
             ExternalOperationException externalOperationException = exception as ExternalOperationException;
 
-            if (externalOperationException is not null
-                && externalOperationException.InnerException is { }
-                && externalOperationException.InnerException.Message.Contains(_dubiousOwnershipSecurityConfigString))
+            if (externalOperationException?.InnerException?.Message?.Contains(_dubiousOwnershipSecurityConfigString) is true)
             {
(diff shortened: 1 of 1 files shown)


gerhardol:
I agree that the popup is suggesting that this is an error for the user - but there are not so many other design alternatives.
However, it has been a way to find hidden errors in the application and this will remove these reports completely.

What about opting in to NBug in the settings?

mstv:
"If you think this ..." could be turned into a checkbox "I **really** think this ...", which then enables the "Report bug!" button.

BTW: I am not sure whether the exclamation mark encourages users to click that button.

vbjay:
Is it possible to have a collapsible section where buttons are and text of something like 


An error occured but is most likely a local issue with the current state of the repo. This dialog is to provide you the info. Only report an issue if....


They must read that to expand to get the report bug button or any other buton we need to ask them to read before clicking.

mstv:
... or just handle it like a UserExternalOperationException:

grafik

RussKie:
....or that 🤣

RussKie:
Updated

GitUI/NBugReports/BugReportInvoker.cs
```diff
@@ -104,11 +105,19 @@ public static void Report(Exception exception, bool isTerminating)
...
+                    ReportDubiousOwnership(externalOperationException.InnerException);
+                    return;
+                }
+
+                if (innerException.Message.StartsWith(_fatalString))
+                {
+                    ReportGitFatal(externalOperationException, isTerminating);
+                    return;
+                }
             }
```
mstv, line 121: What I have meant is reusing the existing handling, i.e. a few lines below just:
```
bool isUserExternalOperation = exception is UserExternalOperationException
    || exception is ExternalOperationException externalOperationException
        && externalOperationException?.InnerException?.Message?.StartsWith(_fatalString) is true;
```

gerhardol:
I suggest the same handling should be applied to all Git errors, slightly hiding the issue.
This PR hides Git errors for fatal only, which is likely what is the most interesting issues to report.

The reporting from Git errors could be disabled by default (except for _dubious owner_ and other identified situations), enabled in Settings.

RussKie:
> I suggest the same handling should be applied to all Git errors

Are you proposing to test `externalOperationException.Command` for being equal to `AppSettings.GitCommand` or `AppSettings.WslGitCommand`?

RussKie:
> The reporting from Git errors could be disabled by default (except for _dubious owner_ and other identified situations), enabled in Settings.

The current dialog doesn't readily offer to log an exception, though if you chose "View details" the new dialog will have "Log issue" button.
I think that's an acceptable solution; at least for now.

RussKie:
> > I suggest the same handling should be applied to all Git errors
> 
> Are you proposing to test `externalOperationException.Command` for being equal to `AppSettings.GitCommand` or `AppSettings.WslGitCommand`?

Pushed an alternative to that effect.

GitUI/NBugReports/BugReportInvoker.cs
```diff
@@ -103,7 +103,9 @@ public static void Report(Exception exception, bool isTerminating)
                 LogError(exception, isTerminating);
             }
 
-            if (exception is ExternalOperationException externalOperationException
+            ExternalOperationException externalOperationException = exception as ExternalOperationException;
+
+            if (externalOperationException is not null
                 && externalOperationException.InnerException is { }
                 && externalOperationException.InnerException.Message.Contains(_dubiousOwnershipSecurityConfigString))
```
mstv, line 110: just
```suggestion
            if (externalOperationException?.InnerException?.Message?.Contains(_dubiousOwnershipSecurityConfigString) is true)
```

mstv (commented):
👍

GitUI/NBugReports/BugReportInvoker.cs
```diff
@@ -132,6 +134,13 @@ or DirectoryNotFoundException
                                                  or PathTooLongException
                                                  or Win32Exception;
 
+            // Treat all git errors as user issues
+            if (string.Equals(AppSettings.GitCommand, externalOperationException?.Command, StringComparison.InvariantCultureIgnoreCase)
+             || string.Equals(AppSettings.WslGitCommand, externalOperationException?.Command, StringComparison.InvariantCultureIgnoreCase))
+            {
+                isUserExternalOperation = true;
+            }
```
RussKie, line 142: @gitextensions/git-extensions-team are we happy to proceed with this?

RussKie:
Bump

gerhardol (approved):

GitUI/NBugReports/BugReportInvoker.cs
RussKie: I can't say I'm excited by this syntax but I'll give in.

GitUI/NBugReports/BugReportInvoker.cs
RussKie: My bad. Thank you for improving the description.

PR #10525, opened 2022-12-14
FormBrowse: Refresh revisions on main thread

### LINKED ISSUE #10491: View is not refreshed after periodic backgound fetch
### Environment

- Git Extensions 4.0.0.15569
- Git 2.38.1.windows.1
- .NET 6.0.11
- DPI 96dpi (no scaling)

### Issue description

I activated periodic background fetch every 120 seconds and also turned "Refresh view after fetch" on
It seems that the fetch is done in the background, but the view is not refreshed. After pressing F5, the view is updated with data that was fetched in the beckground.

### Steps to reproduce

Setup periodic beckground fetch:
Arguments of git command to rund: fetch --all
Fetch every (seconds): 120
Refresh view after fetch: On
Fetch all submodules: Off

Make sure that there is a change on the remote repository.
Wait for more than 120 seconds (e.g. 240s) to be sure that the background fetch happened.
Press F5, the grid now gets refreshed with the changes that had been fechted but not already displayed.

Worked as expected with 3.5.4.12724

### Diagnostics

Git log

mstv:
Fixes #10491

## Proposed changes

- Since `IGitUICommands.PostRepositoryChanged` can be notified from background tasks, switch to the main thread if necessary
  (Thanks for adding `ThreadHelper.AssertOnUIThread()` to `RevisionConrol.PerformRefreshRevisions()`!)

## Screenshots 

N/A

## Test methodology 

- manually using BackgroundFetchPlugin

## Test environment(s) 

- Git Extensions 33.33.33
- Build e56a5b91e7919a4df530662f11f68baa31f5fa4c
- Git 2.38.1.windows.1
- .NET 6.0.10
- DPI 96dpi (no scaling)

```
```

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -552,7 +552,15 @@ namespace GitUI.CommandsDialogs
         {
             // Note that this called in most FormBrowse context to "be sure"
             // that the repo has not been updated externally.
-            RefreshRevisions(e.GetRefs);
+
+            // It can also be called from background tasks, e.g. from BackgroundFetchPlugin.
+            if (ThreadHelper.JoinableTaskContext.IsOnMainThread)
+            {
+                RefreshRevisions(e.GetRefs);
+                return;
+            }
+
+            Invoke(() => RefreshRevisions(e.GetRefs));
         }
 
         /// <summary>


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -552,7 +552,15 @@ private void UICommands_PostRepositoryChanged(object sender, GitUIEventArgs e)
...
-            RefreshRevisions(e.GetRefs);
+
+            // It can also be called from background tasks, e.g. from BackgroundFetchPlugin.
+            if (ThreadHelper.JoinableTaskContext.IsOnMainThread)
+            {
+                RefreshRevisions(e.GetRefs);
+                return;
+            }
+
+            Invoke(() => RefreshRevisions(e.GetRefs));
```
RussKie, line 563: What do you think about inverting this?
```suggestion
            // It can also be called from background tasks, e.g. from BackgroundFetchPlugin.
            if (!ThreadHelper.JoinableTaskContext.IsOnMainThread)
            {
                Invoke(() => RefreshRevisions(e.GetRefs));
                return;
            }

            RefreshRevisions(e.GetRefs);
```

RussKie (approved):

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: Both are OK with me; positive test versus error exit flow

gerhardol (approved):

GitUI/CommandsDialogs/FormBrowse.cs
mstv: I thought about it and quickly decided to avoid the negation `!` as usual. But in this case, it reads better with the negation.
Done.

PR #10534, opened 2022-12-15
Update version of Ubuntu for fix up check

pmiossec:
Fixes build failing as brownout warning of v18.04 deprecation

## Proposed changes

Switch to last LTS version of Ubunto used by the fixup check

## Test methodology 

- Triggered a PR that should failed and one that should succeed.

## Test environment(s) 

- GitHub PR

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -4,7 +4,7 @@ on: [pull_request]
 
 jobs:
   block-fixup:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v2.0.0


pmiossec:
(for testing)

gerhardol (approved):

README.md
```diff
@@ -11,7 +11,7 @@ Have a question? Come and talk to us: [![Gitter](https://badges.gitter.im/Join%2
 
 <a href="#backers" alt="sponsors on Open Collective"><img src="https://opencollective.com/gitextensions/backers/badge.svg" /></a> <a href="#sponsors" alt="Sponsors on Open Collective"><img src="https://opencollective.com/gitextensions/sponsors/badge.svg" /></a>
 
-### v4 // Next Version ([build instructions](https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions))
+### Version 4.x ([build instructions](https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions))
```
RussKie, line 14: This change isn't related. 
And I don't agree with the removal of "vNext".

RussKie (approved):
👍 modulo the comment

pmiossec:
The README change is done to add a fixup commit and verify that the check
is working well by failing.

I will remove it tomorrow because I'm on my phone now and github web
interface does not have a feature to remove a commit 😕

But I'm interested why you are against removing the 'next' as the v4 is the
'current' version?

RussKie:
> But I'm interested why you are against removing the 'next' as the v4 is the 'current' version?

The current is v4. We haven't discussed or decided what goes in to vNext, and hence it doesn't have a version. E.g., we may target .NET 7, but the instructions for building it may not necessarily change.

PR #10559, opened 2022-12-20
Notify Left Panel of filtering for current branch

### LINKED ISSUE #10558: Impossible to switch between branches if revision filter is set to "current branch"
### Environment

- GitExtensions version: 4.0.1.15887
- GIT version: 2.38.1.windows.1
- OS version:  windows 10 pro
- .NET version: 6.0.12
Screenshot 2022-12-20 151731
Screenshot 2022-12-20 151752

### Issue description

Can't switch to other branch until set filter to "all branches". Previous versions it was available.

### Steps to reproduce

You need two different branches with different commits. 
Switch to the first branch and set filter to "current branch"
Then try to switch to the second branch through the tree on the left side of the screen - you should get the error from the attachments

Yes

### Diagnostics

_No response_

mstv:
Fixes #10558

## Proposed changes

- `RefreshLeftPanel`: Also take `IsShowCurrentBranchOnlyChecked` into account for `isFiltering`

### Before

not filtered, exception on click

### After

image

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 2338c39c1e89ca51149f6ed5e17e7a4c039865f2
- Git 2.39.0.windows.1
- .NET 6.0.12
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

## Merge strategy

----


gerhardol (approved):
Maybe there should be a new method like FilterInfo.HideAnyBranch (with corresponding RevisionGridControl method), but there are two filter levels already...

gerhardol:
Why PR to 4.0 and not master?
Needed in both.

RussKie:
I _personally_ find it easier to fix v4 first, then do a tracking merge
back to the master.

If you're doing dual checking or manual porting you need to remember to
apply the milestone correctly. Otherwise it becomes an overhead (and a real
PIA) during the release process.

RussKie (commented):
Do you think we could add a regression test for this?
Looks good otherwise.

gerhardol:
> Do you think we could add a regression test for this? Looks good otherwise.

There are no such tests for the forms, but could be added to FilterInfo if a new method is added there (and RevisionGrid, FormBrowse) in addition to HasRevisionFilter, HasRevisionFilter. (I suggested HideAnyBranch  above, maybe HasAnyBranchFilter is better - HasRevisionBranchFilter is no good as RevisionBranch is a generic name for the branch filters, see GetBranchFilterSummary).
Is it worth it?

mstv:
> > Do you think we could add a regression test for this? Looks good otherwise.
> 
> There are no such tests for the forms, but could be added to FilterInfo if a new method is added there (and RevisionGrid, FormBrowse) in addition to HasRevisionFilter, HasRevisionFilter. (I suggested HideAnyBranch above, maybe HasAnyBranchFilter is better - HasRevisionBranchFilter is no good as RevisionBranch is a generic name for the branch filters, see GetBranchFilterSummary). Is it worth it?

Work is in progress. Both "move to FilterInfo" and "request for tests" are valid points - aiming in the same direction. There are more conditions to consider.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -181,6 +181,16 @@ public bool HasFilter
...
         }
 
+        /// <summary>
+        /// Has any filters which might hide branches or tags.
+        /// </summary>
+        public bool HasRefsFilter
+        {
+            get => HasFilter
+                || IsShowCurrentBranchOnlyChecked
+                || (ByCommitsLimit && HasCommitsLimit);
```
gerhardol, line 191: Note: The default 100K limit will hide commits in for instance Linux repo, so technically all 'filters' could hide commits. The settings value could be changed too.
I believe this is good enough.

gerhardol (approved):

GitUI/UserControls/RevisionGrid/FilterInfo.cs
mstv: I have tested this: If the number of commits is limited, the Left Panel contains branches and tags which are not visible in the RevisionGrid.
Do you mean `|| AppSettings.MaxRevisionGraphCommits > 0` should be added?

BTW: Bad UX:
Filter_Limit_is_reset

gerhardol: I say that we ignore the case where default limit do not load all refs. 
What is bad with the ui?

mstv: > I say that we ignore the case where default limit do not load all refs.

It is clearer and simpler with `HasCommitsLimit`, which already takes the default limit into account.

---

> What is bad with the ui?

Con, main point: The user input is lost when the Checkbox is unticked.
Pro: The effective value is displayed.

gerhardol: > > I say that we ignore the case where default limit do not load all refs.
> 
> It is clearer and simpler with `HasCommitsLimit`, which already takes the default limit into account.

But then the check is meaningless, as isFiltering will be true except when users have set commits limit to 0, which I would expect is very uncommon.
Then it is better to remove isFiltering for side panel completely.
As this is for 4.0, I suggest to just set the value in RefreshLeftPanel() with a comment and cleanup in master. Maybe rendering is done twice there now (a default rendering applying style must be done, maybe there is a check for visibility too.)

> > What is bad with the ui?
> 
> Con, main point: The user input is lost when the Checkbox is unticked. Pro: The effective value is displayed.

(Not really UI as I would define it but possibly confusing).
This was a decision I took when working with this. Can be discussed.
Welcome to the club of maintainers burnt out by filters, first RussKie, then me, now you too?

RussKie (approved):

GitUI/UserControls/RevisionGrid/FilterInfo.cs
mstv: > But then the check is meaningless.

Perhaps for most of the users. But the check is not very expensive and the result is correct.
Since I had to wonder why `Blame this version` was disabled, I have turned off the default limit.
Filtering always is most safe of course and would reduce the code size a little.
Shall I remove `isFiltering`? There are only a few locations.
(But I will keep the tests added for `HasFilter` and the more consistent test setup.)

gerhardol: > Shall I remove isFiltering? There are only a few locations.

I would like to see that, the side panel code is complicated as it is.

mstv: I have removed `isFiltering` and `FilterInfo.HasFilter`

gerhardol (approved):

RussKie (approved):

PR #10572, opened 2022-12-23
Reduce string allocation in StringExtensions.ShortenTo

lechu445:
Here is comparison of decompiled code of both methods https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE+UkEAbMhAAyIDKAFgPYBOALgKYB2AKjagK512sMAKAqQD8iAM4M6AGkQBLFg0QBbAIYAPADKsA5gyoBKbLiwxc5+QDNEQsgDoAkuIByXADZuA8nQCiygA4MAJ4CknQGRmYWOKbR0SAA7CSkdn6BQQDcxtEAvtkWctahUnbaLHpUiAA8ALwqGmUVkXGY+S2JElJZUbltuIU2alq6+tWI8M1xsS24HWF2FFwARsICpLJDjfoG3S15MH04cyWLK2RrGw0jlQC04waIANSIAER27y+7uPvZCCTEwkotEYrA4AAU6DR/DRxEwACa2USdGTyRT1Yblbb5abRAaIxwudxeXwBYLFcKTaI4mYdYSpUmZQ6IfYtPHzLaVWrojmUizU9pJMJfXo9ApFTbXMYTJn8uLHOgLZardbc647Jk/UXmDoAEheGHmAEFxBR/KoWBdVZjbvccu87J98vsckA=
As you can see instead of string.Concat, DefaultInterpolatedStringHandler directly appends `ReadOnlySpan<char>`, so we don't need to allocate the span on the heap.

PR #10574, opened 2022-12-23
Reduce byte array allocation in JsonSerializer.Serialize

lechu445:


gerhardol (approved):
Could be bundled with #10572 even if I am not convinced this is more efficient (probably so, just not looked into the details).

RussKie (approved):

lechu445:
I do not agree for point 4th of the cla. I don't like that contributors personal data is public. 
Because of that I am closing this PR. Feel free with knowledge that I suggested in this PR.

PR #10585, opened 2022-12-27
Update French.Plugins.xlf

Copernicus13:
Fix some typos / errors / anglicism.

Fixes #

## Proposed changes

- 
- 
- 

## Test methodology 

- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy

----

### DIFF
--- a/GitUI/Translation/French.Plugins.xlf
+++ b/GitUI/Translation/French.Plugins.xlf
@@ -381,7 +381,7 @@ Les branches distantes seront supprimées '{0}'. On ne peut pas annuler.
       </trans-unit>
       <trans-unit id="_deletingUnmergedBranches.Text">
         <source>Deleting unmerged branches will result in dangling commits. Use with caution!</source>
-        <target>Supprimer des branches non fusionnées donnera des commits instables. A utiliser avec prudence !</target>
+        <target>Supprimer des branches non fusionnées donnera des commits instables. À utiliser avec prudence !</target>
         
       </trans-unit>
       <trans-unit id="_loading.Text">
@@ -1061,7 +1061,7 @@ To enable Gerrit support for a repository, a .gitreview file must be created. Th
       </trans-unit>
       <trans-unit id="_installCommitMsgHookMessage.Text">
         <source>Gerrit requires a commit-msg hook to be installed. Do you want to install the commit-msg hook into your repository?</source>
-        <target>Gerrit nécessite qu'un crochet commit-msg soit installé. Voulez vous intaller le crochet commit-msg dans cotre dépôt?</target>
+        <target>Gerrit nécessite qu'un crochet commit-msg soit installé. Voulez vous intaller le crochet commit-msg dans votre dépôt ?</target>
         
       </trans-unit>
       <trans-unit id="_installCommitMsgHookShortText.Text">
@@ -1110,7 +1110,7 @@ To enable Gerrit support for a repository, a .gitreview file must be created. Th
       </trans-unit>
       <trans-unit id="_settingsErrorProjectNotEntered.Text">
         <source>The &quot;project&quot; setting in the .gitreview file is mandatory.</source>
-        <target>The paramètre &quot;projet&quot; dans le fichier .gitreview est obligatoire.</target>
+        <target>Le paramètre &quot;projet&quot; dans le fichier .gitreview est obligatoire.</target>
         
       </trans-unit>
     </body>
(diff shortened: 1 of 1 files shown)


gerhardol:
Thanks for your contribution, but the source for translation is in Transifex: https://www.transifex.com/git-extensions/

pmiossec:
Yes, all the corrections are good but should be done in transifex.
And you should be aware that none of these translations will be included in the following versions of GitExtensions until French translations achieved a good level of translated strings. I think it's 95%.
So if you want to see everything bundled, please help translate the missing translations...

pdelrot-samea:
Bonjour. I will help in french translation. I'm using GitExtensions since one or two years. Good tool. 
BR. Pascal

RussKie:
Awesome

pmiossec:
Merci. Beaucoup.

pdelrot-samea:
Bonjour,
A quick question regarding translation of git specific terms, as the operations names (push, pull, reflog...).

Will you prefer @RussKie @pmiossec :

- french translation, keeping the git commands as they are ("Do a git *pull*" translated in "Faire un git *pull*"). It may be more convenient when referring to many online ressources for the newbies.
- stricter translation, by translating whenever possible the git commands (so "Faire un *tirage* git"). The git book provides translations for most of the operations (https://git-scm.com/book/fr/v2)

I do not have real preference. For the translations made 2 days a go, I was more on the strict translation. Others past contributions kept most of the time the initial git command name.

At least, the whole translation package has to be uniform...

Best regards,
Pascal

pmiossec:
I think you get the point of what is difficult with doing a translation. The "Beaucoup" in my previous message was because I gave up maintaining the french translation to prefer spending the time I have on development and because I had not found the right answer to these questions.

As git is not a straight forward tool where you have to understand a lot of things and read a lot of documentation/help, I think it's better that the user know and see the original names (and in the git book, they display both sometimes).

But in the same time, every times I saw a git tool translated, they tend to adopt the stricter translation strategy which feels cleaner.
But I don't know how the users can improve and learn git by not being in touch with git terminology...

A good example of a strict translation is I find it difficult to find a good and enough separated/clearly defined translation for fetch and pull (and the proposition in the link below doesn't convince me). **And as consequence, I have the feeling that the user is not always able to understand what git command is done**. To be clearer, when you already know git (terminology), your are able to understand the translation but not sure for the other way around.

Stricter translation is better for people really not talking english but I don't think there is a lot of people like that in computer science...

>Others past contributions kept most of the time the initial git command name.

Yes, that was mostly me... But as I said, I have still not found the right answer.

Some times ago, I found a very good glossary of git terms translated in french (I thought it would have helped me a lot when doing the translation) and trying to find it again, I stumble on this one (I don't remember if it's the one I'm looking for):
https://github.com/progit/progit2-fr/blob/master/glossaire-git.asc

PS: I just tried the french translation, thing that I have not done in years, and I really think that "Importer" for pull is really a bad translation. I saw a translation that makes more sense to me "fetch: rapatrier" and "pull: récupérer" (but in glossary "récupérer" is for fetch ☹️)

pdelrot-samea:
I completely agree with you. I translate from time to time (years...) few OSS. Takes time, but necessary too.

I like "rapatrier" for fetching. It's the first time I see it. I wrote "importer" yesterday in one translation, few other OSS use "tirer" which is too "word to word translation" for me.

Let's say :

- fetch : rapatrier
- pull : récupérer
- push : pousser

I will try to keep the initial commands names when possible, a bit difficult without the GUI context at this time for me (I'm mostly an embedded/hardware developper, the tools I have are not the same as for computer software design).

Pascal

pmiossec:
> I translate from time to time (years...) few OSS. Takes time, but necessary too.

🤩

> few other OSS use "tirer" which is too "word to word translation" for me.

Same for me.

> 
> Let's say :
> 
>     * fetch : rapatrier
>     * pull : récupérer
>     * push : pousser

😀


> I will try to keep the initial commands names when possible, a bit difficult without the GUI context at this time for me

I feel you.
You can download the `xlf` file from Transifex and put it in GitExtensions folder to test results.
Also sometimes the context (which is the `id` if the GUI control could help a little):

image

A thing very difficult are also strings with `&` that is used as keyboard shortcuts and where the selected letter is not in the french text. Only seeing the string in the UI  could help you choose the good letter and know wher to put the &...

pmiossec:
Oh! And feel free to discuss about french translations when you have doubts.
Because I remember to feel alone when doing the translation...

PR #10607, opened 2023-01-05
Accelerate builds in Visual Studio

drewnoakes:
## Proposed changes

This PR improves developer productivity while working in the Git Extensions repo by speeding up build times considerably.

It does that with two changes:

1. Enable [Reference Assemblies](https://learn.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies)
2. Enable [Build Acceleration in Visual Studio](https://github.com/dotnet/project-system/blob/main/docs/build-acceleration.md)

## Results

For each of the following tests, a change was made in `GitExtUtils` then the whole solution was built.

- The change did not modify the public API. For example, altering a method body, reformatting code, adding a comment.
- That project was chosen as several other projects depend upon it.

## Before (17.4.2)

- 39 succeeded, 4 up-to-date
- Elapsed times
  - 23.750 seconds
  - 24.710 seconds
  - 23.196 seconds
  - 25.509 seconds

## Enable reference assemblies (17.4.2)

- 29 succeeded, 14 up-to-date
- Elapsed times
  - 18.167 seconds
  - 11.014 seconds
  - 10.148 seconds
  - 10.420 seconds
  - 11.532 seconds

## Enable Build Acceleration (17.5 preview 3)

- 1 succeeded, 41 up-to-date
- Elapsed times
  - 1.628 seconds
  - 1.684 seconds
  - 1.653 seconds
  - 1.906 seconds
  - 1.848 seconds

Overall, this PR cuts build time for the scenario by ~93%.

## Test methodology

- Lots of manual testing

## Test environment(s)

- Visual Studio 2022 version 17.5 preview 3

## Merge strategy

- Merge commit. (PR submitter to rebase and squash before merges).

----

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,7 +22,7 @@
     <!-- TODO once all project migrated to SDK-style, remove this and move properties to Directory.Build.props -->
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 
-    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
 
     <IsPublishable>false</IsPublishable>
   </PropertyGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,6 +25,9 @@
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
 
     <IsPublishable>false</IsPublishable>
+
+    <!-- Opt in to build acceleration in VS (from 17.5 onwards): https://github.com/dotnet/project-system/blob/main/docs/build-acceleration.md -->
+    <AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/Externals/Directory.Build.props
+++ b/Externals/Directory.Build.props
@@ -2,7 +2,7 @@
 <Project>
   <PropertyGroup>
     <UseWindowsForms>true</UseWindowsForms>
-    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
 
     <_SuppressSdkImports>true</_SuppressSdkImports>
     <DeterministicSourcePaths>false</DeterministicSourcePaths>
--- a/Setup/Setup.wixproj
+++ b/Setup/Setup.wixproj
@@ -23,6 +23,8 @@
     <DefineConstants>Version=$(Version);NumericVersion=$(Version);ArtifactsBinPath=$(ArtifactsBinPath);ArtifactsPublishPath=$(ArtifactsPublishPath);PluginManagerSourceDir=$(PluginManagerSourceDir)\GitExtensions.PluginManager\Output</DefineConstants>
     <DefineConstants Condition=" '$(Configuration)' == 'Debug' ">Debug;$(DefineConstants)</DefineConstants>
     <SuppressIces>ICE61;ICE80;ICE91;ICE38</SuppressIces>
+    <!-- Disable build acceleration for WIX projects -->
+    <AccelerateBuildsInVisualStudio>false</AccelerateBuildsInVisualStudio>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="Product.wxs" />


pmiossec:
@drewnoakes I know there is a link toward a documentation but it is very long. Could you sum up a little or do an introduction of what is the main difference between reference assemblies and normal/implementation ones? Or what does contain implementation assemblies that take time to generate and is not needed in our project?

pmiossec:
Or what could probably fail and we should be careful about once testing or switching to it that could break? use of reflection for example?

gerhardol:
Looks great!
I only get 17.5preview2 so this is really early, cannot test yet...
Should this be merged before 17.5 is officially released?

RussKie:
> Could you sum up a little or do an introduction of what is the main difference between reference assemblies and normal/implementation ones?

"Ref assemblies" contain only API contracts, and "implementation assemblies" contain implementations. Because of that, the former are often significantly smaller in size.

As an example, you can see Windows Desktop ref assemblies here: `C:\Program Files\dotnet\packs\Microsoft.WindowsDesktop.App.Ref\7.0.0\ref\net7.0`. If you open any of the assemblies in a decompiler you'll see something similar to:
image

Ref assemblies are used for compilation - all the compiler needs to know such and such types or method exists, and its signatures are correct.

RussKie:
> Should this be merged before 17.5 is officially released?

Whilst it's still technically possible that the feature may be changed before 17.5 goes GA, I think it's totally fine to merge this.

pmiossec:
> Ref assemblies are used for compilation - all the compiler needs to know such and such types or method exists, and its signatures are correct.

It means that in the end in the output folder you will still have the implementation assemblies in the end (and reference assemblies are used only during the building)?

RussKie:
> > Ref assemblies are used for compilation - all the compiler needs to know such and such types or method exists, and its signatures are correct.
> 
> It means that in the end in the output folder you will still have the implementation assemblies in the end (and reference assemblies are used only during the building)?

Yes, I belie so. @drewnoakes can provide more details though as I'm not intimately familiar with this area.

RussKie:
I've run some tests locally.

From command line with MSBuild version 17.3.2+561848881 running `dotnet build` (`git clean -xdf` before each run):
* before: 00:01:01.08
* after: 00:00:35.65

From VS 17.5.0 Preview 2.0 it's a bit of mixed bag:
* before
    - full rebuild from cold: 39.312 seconds
    - subsequent full rebuild: 30.878 seconds
* after: 
    - full rebuild from cold: 52.538 seconds (first try) / 44.471 seconds (second try)
    - subsequent full rebuild: 32.358 seconds

RussKie (approved):

drewnoakes:
@pmiossec

There's no impact on the DLLs that end up in the output assembly or that occur at runtime. Reference assemblies are only used during compilation. Reference assemblies end up in eg: `obj/Debug/ref/MyProject.dll`.

It's hard to summarize in much detail, and the linked doc does a better job that I will here, but the key point is that identifying whether a project needs to build/compile is done (at some level) via timestamps on the file system. With that in mind, if a method body is changed, but not its public API, then the implementation assembly changes on disk but the reference assembly does not. Therefore a referencing project can see that the API surface is unchanged and therefore does not need to recompile itself. Instead, it just needs to copy the referenced output assembly into its output directory, then exit. There's some hand waving here, but that's the gist.

The second part of this PR (Build Acceleration) means that VS itself can perform this copying, which completes in ~milliseconds. Previously we would call MSBuild to do that copying, which would take ~seconds.

@gerhardol

17.5p2 supports Build Acceleration, but it works better with 17.5p3 which will be out soon. The fixes in p3 apply to edge cases that I don't believe exist in the GE repo.

@RussKie

Thanks for running the tests. The delta you're seeing are due to the use of reference assemblies only. Specifically:

- Build Acceleration is VS-only, so there's no impact on the command line.
- Build Acceleration applies only to _incremental_ builds, so performing _re_builds won't have any benefit here. You need to do a normal build to see the gains I posted above.

Could you try your tests again on an incremental build please, making a non-API-change in the `GitExtUtils` project, then building (not rebuilding) the whole solution?

pkunze:
Is there any way to make use of this using VS Build Tools?

drewnoakes:
> Is there any way to make use of this using VS Build Tools?

The VS Build Tools allow you to build Visual Studio projects from a command-line interface.

This PR made two changes, and the answer to your question differs for each change:

1. Enabled `ProduceReferenceAssembly`. This will work with CLI builds (such as those performed by VS Build Tools).
2. Enabled `AccelerateBuildsInVisualStudio`. This will not work with CLI builds. Build Acceleration requires a long-running process (like the IDE) to have a bunch of information about the solution in memory at the moment the request to build is received. CLI builds are one-shot, and not candidates for acceleration.

pkunze:
Thank you for responding. 😊 As far as I understand enabling reference assemblies wouldn't enhance sln build performance (which my question was aiming towards), does it?

pkunze:
Alright. I will check this out. Thanks for clearing that up. 👍😊

Issue #9761, opened 2021-12-04
Console path doesn't change when switching to another project through Start -> Open...


maxulfves:
### Environment

- Git Extensions 3.5.4.12724
- Build 65f01f39982ccb121c4df7d4b00b506939553958
- Git 2.24.0.windows.2 (recommended: 2.30.0 or later)
- Microsoft Windows NT 10.0.19043.0
- .NET Framework 4.8.4420.0
- DPI 96dpi (no scaling)

### Issue description

When switching to another project by opening the start menu and selecting "Open..." and then the project the console stays at the old location. 
![image](https://user-images.githubusercontent.com/11315779/144722245-a9dceeb9-3b6a-4c95-93f5-f74b29a10096.png)

I can't see the use case for it, so I assume it's a bug. 

### Steps to reproduce

1. Open a repository
2. Go to the menu and open a different repository
3. Notice that the console now is still at the old path.

gerhardol:
Cannot reproduce.
Both master and 3.5.4, regardless if start with last repo and how a repo is switched.

TrueRaider:
Same problem on Git Extensions 4.0.0.15569
- Build cba315df5366fca01240d93b6e5a5077fd5e6630
- Git 2.38.1.windows.1
- Microsoft Windows NT 10.0.19044.0
- .NET 6.0.11
- DPI 96dpi (no scaling)

But reproduced on ver 3.5.4 for me. Something in global git settings, maybe...

TrueRaider:
With cmd working fine. Problems with bash
![image](https://user-images.githubusercontent.com/16497965/201335702-474fd42c-5e4e-4635-ab52-5a10307277cd.png)

mstv:
The correct `cd d:\path\to\repo` command is automatically entered to the bash prompt - regardless of how I switch repos. I have to hit `Enter` myself though.
I use the bash from Git for Windows.

TrueRaider:
Yes, but not so long ago it was happening automatically.

pmiossec:
@TrueRaider I'm absolutely not sure it will solve your issue (it's a fix for @mstv issue) but you could test #10613 to see if it fixes your issue. Portable version available in artifacts of build when it will be finished: https://ci.appveyor.com/project/gitextensions/gitextensions/builds/45853432/artifacts

PR #10613, opened 2023-01-05
Conemu: fix switch of repo folder in console tab

pmiossec:
ConEmu handling of new line should have changed
(and the cursor is now passing to the next line and no more run the command entered in the prompt) so now sending the "Enter" key so that the "cd" command is run

Fixes https://github.com/gitextensions/gitextensions/issues/9761#issuecomment-1312139285

For ConEmu macro keys: https://conemu.github.io/en/GuiMacro.html#Keys

### Before

console_switch_failed

### After
console_switch_succeed

## Test methodology 

- Manual with bash and cmd

## Test environment(s) 

- Git Extensions 33.33.33
- Build 88ccf3298246037f10992435997792fab50d7c96
- .NET 6.0.12
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/Shells/ConEmuControlExtensions.cs
+++ b/GitUI/Shells/ConEmuControlExtensions.cs
@@ -16,14 +16,17 @@ namespace GitUI.Shells
             switch (shell.Name)
             {
                 case BashShell.ShellName:
+                    // Use a ConEmu macro to send the sequence for clearing the bash command line
                     terminal.RunningSession.BeginGuiMacro("Keys").WithParam("^A").WithParam("^K").ExecuteSync();
-                    terminal.RunningSession.WriteInputTextAsync(command + Environment.NewLine);
+                    terminal.RunningSession.WriteInputTextAsync(command);
                     break;
 
                 default:
-                    terminal.RunningSession.WriteInputTextAsync($"\x1B{command}{Environment.NewLine}");
+                    terminal.RunningSession.WriteInputTextAsync($"\x1B{command}");
                     break;
             }
+
+            terminal.RunningSession.BeginGuiMacro("Keys").WithParam("Enter").ExecuteSync();
         }
     }
 }


GitUI/Shells/ConEmuControlExtensions.cs
```diff
@@ -16,14 +16,17 @@ public static void ChangeFolder(this ConEmuControl? terminal, IShellDescriptor?
...
                 case BashShell.ShellName:
+                    // Use a ConEmu macro to send the sequence for clearing the bash command line
                     terminal.RunningSession.BeginGuiMacro("Keys").WithParam("^A").WithParam("^K").ExecuteSync();
-                    terminal.RunningSession.WriteInputTextAsync(command + Environment.NewLine);
+                    terminal.RunningSession.WriteInputTextAsync(command);
                     break;
 
                 default:
-                    terminal.RunningSession.WriteInputTextAsync($"\x1B{command}{Environment.NewLine}");
+                    terminal.RunningSession.WriteInputTextAsync($"\x1B{command}");
```
gerhardol, line 25: Which term is which?

mstv (commented):
Looks sensible, have not run yet

GitUI/Shells/ConEmuControlExtensions.cs
pmiossec: 🤔hummm.... I don't get it.

I have remove the new line because it is no more working in ConEmu and replace by sending the <kbd>Enter</kbd> key to run the command...

gerhardol: The first case is BashShell.ShellName, no `\x1B`
Is the macro invoked in both situations, what is the difference?
How to test both paths?

pmiossec: The first case if for bash where we use a macro because it doesn't react like other shells (cmd, powershell, pwsh) that clean the command line when you print the <kbd>Esc</kbd> ( `\x1B` ) char.

>Is the macro invoked in both situations, what is the difference?

Then the macro to send `Enter` is used to press ENter in both cases (now that printing new line doesn't work anymore)

>How to test both paths?

Switching the default shell 
image

RussKie (approved):

mstv (approved):

gerhardol (approved):

PR #10635, opened 2023-01-12
Update README.md

nhance:
Removes dragons because they are not real and have no basis in nature. Have you ever seen a dragon skeleton? There is no way a dragon can support its weight and there is no way the bones for the wings fit into the normal skeletal pattern of actual animals that exist.

## Proposed changes

- Removes dragons.
- They are not real.
- Some people dislike them.

## Test methodology 

- Question: Can you picture a dragon skeleton?
- Answer: No.

## Test environment(s) 

- Earth
- Reality

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ If you want to **update a portable version**, you should delete all the files an
   <tr>
     <td>
       <strong>Current dev stream</strong><br />
-      NB: expect :unicorn: :unicorn: and :dragon: :dragon:
+      NB: expect :unicorn: :unicorn:
     </td>
     <td>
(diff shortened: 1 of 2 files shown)


vbjay (changes requested):
Dragons are real. You just have to believe. Have faith and be a dragon slayer.  Instead of taking away from our menagerie, add an animal.

gerhardol:
What about unicorns?

vbjay:
Frankly I think this is a poor way to become a contributing member. The whole point of it is to point out there are dev steam issues abound. Complaining over using dragons is silly. Instead tackle an open issue and slay away.

RussKie:
I appreciate the effort, however, it carries no value.

Issue #9990, opened 2022-05-12
Compile as x64 app, stop targeting x86.


RussKie:
### Environment

https://github.com/gitextensions/gitextensions/tree/net7.0

### Issue description

I tried [porting the app to .NET 7](https://github.com/gitextensions/gitextensions/tree/net7.0), and whilst the source compiled without any issues, I couldn't run the app:
```
Application: GitExtensions.exe
CoreCLR Version: 7.0.22.18007
.NET Version: 7.0.0-preview.4.22180.7
Description: The process was terminated due to an unhandled exception.
Exception Info: System.BadImageFormatException: Could not load file or assembly 'D:\Development\gitextensions\artifacts\Debug\bin\GitExtensions\net7.0-windows\GitExtensions.dll'. Format of the executable (.exe) or library (.dll) is invalid.
File name: 'D:\Development\gitextensions\artifacts\Debug\bin\GitExtensions\net7.0-windows\GitExtensions.dll'
```

It took some head scratching, until I figured it was related to the app compiled as 32-bit. I tried removing `<Prefer32bit>true</Prefer32bit>`, and it compiled again without any issues. The app even started, opened the dashboard and then a repo. However, the commit dialog would crash the app.
```
Faulting application name: GitExtensions.exe, version: 33.33.33.0, time stamp: 0x624543e5
Faulting module name: ntdll.dll, version: 10.0.22000.527, time stamp: 0x931cda92
Exception code: 0xc0000409
Fault offset: 0x000000000009326f
Faulting process ID: 0xe560
Faulting application start time: 0x01d864e4378c1b9e
Faulting application path: D:\Development\gitextensions\artifacts\Debug\bin\GitExtensions\net7.0-windows\GitExtensions.exe
Faulting module path: C:\WINDOWS\SYSTEM32\ntdll.dll
Report ID: b719ec28-03cc-4e2c-b53a-3a02d40f532c
Faulting package full name: 
Faulting package-relative application ID: 
```

My hypothesis is that we have some 32-bit specific code/incorrect interop definitions that crash the process... 

Relates to https://github.com/gitextensions/gitextensions/discussions/9869.

### Steps to reproduce

- Pull https://github.com/gitextensions/gitextensions/tree/net7.0
- Update all instances of `<Prefer32Bit>` to `false`

gerhardol:
OK for me

RussKie:
Implemented in https://github.com/gitextensions/gitextensions/pull/11625

PR #10649, opened 2023-01-15
Add support to build GE for Windows on Arm64 (WoA)

### LINKED ISSUE #10674: Version 4.0.2 doesn't run on Windows 11 ARM64
### Environment

- GitExtensions version: 4.0.2
- GIT version: 2.38.1.windows.1
- OS version: Version 10.0.22623 Build 22623 **ARM64**
- .NET version: 6 and 7 are both installed

### Issue description

As of version 4.0.2 I can no longer run the gitexe.exe on Windows 11 ARM64. The previous version, 4.0.1 works (although it is noticeably slower than the previous version, and slower than on x64).

Immediately when starting the app I am prompted to install the .Net 6.0 Desktop runtime which is already installed.

### Steps to reproduce

0. Buy a Windows ARM Dev Kit PC
1. Install .NET 6.0 Desktop Runtime v6.0.13 ARM64 version
2. Install GitExtensions version 4.0.2
3. run gitex.exe

RESULT: Prompt appears to install .NET 6.0 Desktop Runtime v6.0.13
EXPECTED: No prompt.

It worked on version 4.0.1

### Diagnostics

_No response_

chirontt:
Resolves #10674 

With this PR, here are the commands to build GE on a Windows on Arm64 (WoA) platform:

- build the portable GE with the command:
```
> dotnet build
```
- build the shell extensions for WoA:
```
> dotnet build .\scripts\native_arm64.proj
```
The above command produces a `GitExtensionsShellEx64.dll` file for WoA, but GE also requires a `GitExtensionsShellEx32.dll` file to exist, so just make a copy of `GitExtensionsShellEx64.dll` to be `GitExtensionsShellEx32.dll` as well:
```
> copy artifacts\Debug\bin\GitExtensionsShellEx\GitExtensionsShellEx64.dll artifacts\Debug\bin\GitExtensionsShellEx\GitExtensionsShellEx32.dll
```
- package the GE distribution:
```
> dotnet publish
```
The last command would fail to produce a `.msi` distribution, due to it using the WiX library which is currently incompatible with the WoA platform, but a portable GE `zip` distribution for WoA will be successfully produced in the folder `artifacts\Debug\publish`

## Proposed changes
Build for Windows on Arm64 platform

## Test methodology 
Mostly manual tests, done on normal daily usage of the app.

Executing the automated tests, with the `dotnet test` command, mostly fail, with error about `testhost.dll` file:
```
Testhost process for source(s) 'C:\dev\dotnet\gitextensions\artifacts\Debug\bin\tests\UnitTests\GitUI.Tests\net6.0-windows\GitUI.Tests.dll' exited with error: Unhandled exception. System.BadImageFormatException: Could not load file or assembly 'C:\dev\dotnet\gitextensions\artifacts\Debug\bin\tests\UnitTests\GitUI.Tests\net6.0-windows\testhost.dll'. An attempt was made to load a program with an incorrect format.
File name: 'C:\dev\dotnet\gitextensions\artifacts\Debug\bin\tests\UnitTests\GitUI.Tests\net6.0-windows\testhost.dll'
. Please check the diagnostic logs for more information.
```
I guess `testhost.dll` is an x86 library, so it fails to load in an Arm64 GE process.

## Test environment(s) 

- [Git for Windows 2.39.0 ARM64](https://github.com/dennisameling/git/releases)
- Windows 11 on Arm64
- No UI scaling

## Merge strategy

----

### DIFF
--- a/GitExtensions/Project.Publish.targets
+++ b/GitExtensions/Project.Publish.targets
@@ -285,7 +285,7 @@
     Creates an MSI.
     ============================================================
     -->
-  <Target Name="CreateMsi" AfterTargets="CreatePortable">
+  <Target Name="CreateMsi" Condition="'$(Arm64Build)' != 'true'" AfterTargets="CreatePortable">
     <PropertyGroup>
       <_AppVeyorSuffix>$(ARTIFACT_BUILD_SUFFIX)</_AppVeyorSuffix>
       <_PublishMsiVersionSuffix>-$(CurrentBuildVersion.ToString())$(_AppVeyorSuffix)</_PublishMsiVersionSuffix>
--- a/scripts/native.proj
+++ b/scripts/native.proj
@@ -39,11 +39,22 @@
 
       <_ProjectArgs32> /p:Configuration=$(Configuration) /p:Platform=Win32</_ProjectArgs32>
       <_ProjectArgs64> /p:Configuration=$(Configuration) /p:Platform=x64</_ProjectArgs64>
+      <_ProjectArgsARM64> /p:Configuration=$(Configuration) /p:Platform=ARM64</_ProjectArgsARM64>
     </PropertyGroup>
 
     <!-- Build GitExtSshAskPass project, x86 -->
     <Exec
+        Condition="'$(Arm64Build)' != 'true'"
         Command="msbuild.exe $(_GitExtSshAskPassPath) $(_ProjectArgs32) /p:OutDir=$(_OutputPath)GitExtSshAskPass\ /bl:$(_ArtifactsLogDir)\GitExtSshAskPass.binlog"
+        WorkingDirectory="$(_MSBuildCurrentPath)"
+        EchoOff="true"
+        ConsoleToMsBuild="true"
+        StandardOutputImportance="High">
+    </Exec>
+    <!-- Build GitExtSshAskPass project, ARM64 -->
+    <Exec
+        Condition="'$(Arm64Build)' == 'true'"
+        Command="msbuild.exe $(_GitExtSshAskPassPath) $(_ProjectArgsARM64) /p:OutDir=$(_OutputPath)GitExtSshAskPass\ /bl:$(_ArtifactsLogDir)\GitExtSshAskPass.binlog"
         WorkingDirectory="$(_MSBuildCurrentPath)"
         EchoOff="true"
         ConsoleToMsBuild="true"
@@ -51,6 +62,7 @@
     </Exec>
     <!-- Build GitExtensionsShellEx project, x86 -->
     <Exec
+        Condition="'$(Arm64Build)' != 'true'"
         Command="msbuild.exe $(_GitExtensionsShellExPath) $(_ProjectArgs32) /p:OutDir=$(_OutputPath)GitExtensionsShellEx\ /bl:$(_ArtifactsLogDir)\GitExtensionsShellEx32.binlog"
         WorkingDirectory="$(_MSBuildCurrentPath)"
         EchoOff="true"
@@ -59,12 +71,28 @@
     </Exec>
     <!-- Build GitExtensionsShellEx project, x64 -->
     <Exec
+        Condition="'$(Arm64Build)' != 'true'"
         Command="msbuild.exe $(_GitExtensionsShellExPath) $(_ProjectArgs64) /p:OutDir=$(_OutputPath)GitExtensionsShellEx\ /bl:$(_ArtifactsLogDir)\GitExtensionsShellEx64.binlog"
         WorkingDirectory="$(_MSBuildCurrentPath)"
         EchoOff="true"
         ConsoleToMsBuild="true"
         StandardOutputImportance="High">
     </Exec>
+    <!-- Build GitExtensionsShellEx project, ARM64 -->
+    <Exec
+        Condition="'$(Arm64Build)' == 'true'"
+        Command="msbuild.exe $(_GitExtensionsShellExPath) $(_ProjectArgsARM64) /p:OutDir=$(_OutputPath)GitExtensionsShellEx\ /bl:$(_ArtifactsLogDir)\GitExtensionsShellEx64.binlog"
+        WorkingDirectory="$(_MSBuildCurrentPath)"
+        EchoOff="true"
+        ConsoleToMsBuild="true"
+        StandardOutputImportance="High">
+    </Exec>
+    <!-- Copy GitExtensionsShellEx64.dll file to GitExtensionsShellEx32.dll, ARM64 -->
+    <Copy
+        Condition="'$(Arm64Build)' == 'true'"
+        SourceFiles="$(_OutputPath)GitExtensionsShellEx\GitExtensionsShellEx64.dll"
+        DestinationFiles="$(_OutputPath)GitExtensionsShellEx\GitExtensionsShellEx32.dll">
+    </Copy>
   </Target>
 
   <!-- Stub out targets required by the Microsoft.NET.Sdk -->
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,12 +22,14 @@
     <!-- TODO once all project migrated to SDK-style, remove this and move properties to Directory.Build.props -->
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 
-    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
-
     <IsPublishable>false</IsPublishable>
+
+    <!-- Set this manually whenever build ARM64-->
+    <Arm64Build>false</Arm64Build>
 
     <!-- Opt in to build acceleration in VS (from 17.5 onwards): https://github.com/dotnet/project-system/blob/main/docs/build-acceleration.md -->
     <AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/GitExtSshAskPass/GitExtSshAskPass.sln
+++ b/GitExtSshAskPass/GitExtSshAskPass.sln
@@ -1,22 +1,28 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio Version 17
+VisualStudioVersion = 17.4.33015.44
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "GitExtSshAskPass", "SshAskPass.vcxproj", "{E8C01071-3B47-4A3F-9168-AA58A4635638}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|ARM64 = Debug|ARM64
 		Debug|Win32 = Debug|Win32
 		Debug|x64 = Debug|x64
+		Release|ARM64 = Release|ARM64
 		Release|Win32 = Release|Win32
 		Release|x64 = Release|x64
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|ARM64.Build.0 = Debug|ARM64
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|Win32.ActiveCfg = Debug|Win32
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|Win32.Build.0 = Debug|Win32
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|x64.ActiveCfg = Debug|x64
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Debug|x64.Build.0 = Debug|x64
+		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Release|ARM64.ActiveCfg = Release|ARM64
+		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Release|ARM64.Build.0 = Release|ARM64
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Release|Win32.ActiveCfg = Release|Win32
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Release|Win32.Build.0 = Release|Win32
 		{E8C01071-3B47-4A3F-9168-AA58A4635638}.Release|x64.ActiveCfg = Release|x64
@@ -24,5 +30,8 @@ Global
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {3AB083AD-E7AD-41BC-8215-41B684CB1A1C}
 	EndGlobalSection
 EndGlobal
--- a/GitExtSshAskPass/SshAskPass.vcxproj
+++ b/GitExtSshAskPass/SshAskPass.vcxproj
@@ -1,6 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|ARM64">
+      <Configuration>Debug</Configuration>
+      <Platform>ARM64</Platform>
+    </ProjectConfiguration>
     <ProjectConfiguration Include="Debug|Win32">
       <Configuration>Debug</Configuration>
       <Platform>Win32</Platform>
@@ -8,6 +12,10 @@
     <ProjectConfiguration Include="Debug|x64">
       <Configuration>Debug</Configuration>
       <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|ARM64">
+      <Configuration>Release</Configuration>
+      <Platform>ARM64</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="Release|Win32">
       <Configuration>Release</Configuration>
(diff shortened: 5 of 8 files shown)


scripts/native_arm64.proj
```diff
@@ -0,0 +1,63 @@
...
+        StandardOutputImportance="High">
+    </Exec>
+    <!-- Build GitExtensionsShellEx project, ARM64 -->
+    <Exec
+        Command="msbuild.exe $(_GitExtensionsShellExPath) $(_ProjectArgsARM64) /p:OutDir=$(_OutputPath)GitExtensionsShellEx\ /bl:$(_ArtifactsLogDir)\GitExtensionsShellEx64.binlog"
+        WorkingDirectory="$(_MSBuildCurrentPath)"
+        EchoOff="true"
+        ConsoleToMsBuild="true"
+        StandardOutputImportance="High">
+    </Exec>
```
RussKie, line 58: Is there a reason we can't do this in native.proj?

RussKie:
Related: https://github.com/gitextensions/gitextensions/discussions/10288

GitExtensions/GitExtensions.csproj
```diff
@@ -10,7 +10,7 @@
     <ApplicationIcon>..\Logo\git-extensions-logo.ico</ApplicationIcon>
     <StartupObject />
 
-    <Prefer32bit>true</Prefer32bit>
+    <!-- Prefer32bit>true</Prefer32bit -->
```
RussKie, line 13: I'm not sure we can do this; this change will require further investigation. See https://github.com/gitextensions/gitextensions/issues/9990

RussKie (commented):
🎉 

The installer will need to be changed as well, see https://github.com/gitextensions/gitextensions/blob/1ee9b7530f9da099bd68ab5bf9652ff025e2e6af/Setup/Product.wxs#L231-L249

Also, ConEmu doesn't work on Arm64, so some code paths will need to be conditioned:
* https://github.com/Maximus5/ConEmu/issues/1488
* https://github.com/Maximus5/ConEmu/issues/2405

scripts/native_arm64.proj
chirontt: As I mentioned before, I can't produce the `.msi` installer for WoA, due to the WiX incompatibility problem, so I haven't got a chance to see what's needed for it. I've tried to recompile the WiX library for WoA, but it's so old that it throws so many errors (it's pre-net6 program, and .NET SDK in WoA only supports net6+) that I just give it up.

ConEmu problem in WoA was reported almost 5 years ago, and no progress since. I guess we need to wait a couple of more years still. So for now, I just disable ConEmu in my GE settings in WoA. Actually I want to even hide the `Console` tab so that I don't accidentally click it and produce a zombie process in the background:

image

Any idea how to hide it from being displayed on screen?

As to the new `native_arm64.proj` file:
> Is there a reason we can't do this in native.proj?

Nope, just combining them together would cause compile problems on either platforms: they would share objs between different platforms (on WoA) which is a no-no, the x64 compiler wouldn't know how to compile the ARM64 platform which necessitates some needed conditionals to go different compile path for each platform, etc., that I reckon more research are needed to make it work. For now, using a separate proj file for Arm64 is the simplest solution I can think of.

GitExtensions/GitExtensions.csproj
chirontt: I have no problem with that setup with GE in both x64 and WoA platforms. I haven't seen the commit problem mentioned in that issue either.

scripts/native_arm64.proj
pmiossec: > Any idea how to hide it from being displayed on screen?

Look at the 'Build report' tab logic. I think we don't display it always. Maybe it's just a matter of not adding it to the tab collection.

gerhardol: > Any idea how to hide it from being displayed on screen?

Make sure these options are always false (and settings disabled) for WoA.

image

image

chirontt: @gerhardol Yes, the following setting will hide the `Console` tab:

image

RussKie:
I wouldn't rely on a setting,  instead I think we should conditionally
exclude the implementation either with `#if ARM64` or `if Platform ==
"arm64"` (pseudo code).

chirontt:
FYI: I've made my unofficial build of GE v4.0.2 for Windows Arm64 [available](https://github.com/chirontt/gitextensions/releases/tag/v4.0.2-win32-aarch64) on my fork.

scripts/native.proj
```diff
@@ -43,7 +43,7 @@
 
     <!-- Build GitExtSshAskPass project, x86 -->
     <Exec
-        Command="msbuild.exe $(_GitExtSshAskPassPath) $(_ProjectArgs32_) /p:OutDir=$(_OutputPath)GitExtSshAskPass\ /bl:$(_ArtifactsLogDir)\GitExtSshAskPass.binlog"
+        Command="msbuild.exe $(_GitExtSshAskPassPath) $(_ProjectArgs32) /p:OutDir=$(_OutputPath)GitExtSshAskPass\ /bl:$(_ArtifactsLogDir)\GitExtSshAskPass.binlog"
```
RussKie, line 46: Thanks for this! I fixed it in 8ce4af03df53456c8822383ec36d07a9832bcf6f.

GitExtensions/GitExtensions.csproj
RussKie: Let's keep this enabled. This can be disabled as `/p:Prefer32bit=false` during the ARM64 builds.

scripts/native.proj
```diff
@@ -43,7 +43,7 @@
```
RussKie, line 43: Suggest renaming this file to `native.amd.proj`

scripts/native_arm64.proj
```diff
@@ -0,0 +1,63 @@
+<Project DefaultTargets="Build">
```
RussKie, line 1: ...and this to `native.arm64.proj`

RussKie (commented):
Looks good 👍 

Please address the outstanding comments and rebase.

What would be the steps to build a portable version on AppVeyor (since we can't produce an msi)?

Also, we'll need to add build instructions for the ARM64 builds. If you can provide a text, we'll have it added to https://github.com/gitextensions/gitextensions/wiki/How-To%3A-build-instructions.

GitExtensions/GitExtensions.csproj
chirontt: OK, although leaving it enabled makes no difference in the resulting GE software for WoA, as far as I can determine, so the `/p:Prefer32bit=false` may not be needed for WoA build.

scripts/native.proj
chirontt: OK, and as a result, the reference to it in `appveyor.yml` is updated too. So this renaming of `native.proj` makes `appveyor.yml` becomes exclusively X64 (and thus should be renamed to `appveyor.amd.yml` to be explicitly clear?)

scripts/native_arm64.proj
chirontt: OK

scripts/native.proj
RussKie: There are few other places where `native.proj` is used: https://github.com/gitextensions/gitextensions/search?q=native.proj

> So this renaming of `native.proj` makes `appveyor.yml` become exclusively X64 (and thus should be renamed to `appveyor.amd.yml` to be explicitly clear?)

I don't follow, sorry. We continue to build x86 as well, hence the suggestion to rename to `native.amd.proj` and not `native.amd64.proj`.

scripts/native_arm64.proj
RussKie: I don't think it's possible to run multiple build pipelines on a free account. Aren't we able to build both of distros in one pipeline?

> Is there a way to change the `dotnet publish` command to produce the portable `zip` distribution only when building for Arm64?

The publishing targets are here: https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Project.Publish.targets#L271. I don't think it should be too difficult to condition `CreateMsi` target based on a MSBuild property.

scripts/native.proj
chirontt: > There are few other places where `native.proj` is used: https://github.com/gitextensions/gitextensions/search?q=native.proj

OK, due to these reasons, I've decided to revert to the original `native.proj` file name, and have merged the `native.arm64.proj` to it so that the `native.proj` will now support both Intel/AMD and ARM64 builds via the usage of an extra property. And that means the original `appveyor.yml` file won't be changed after all, and such it will continue building the x86/x64 products as originally intended.

I don't have any experience with AppVeyor so I can't do much about how to change `appveyor.yml` to support the extra ARM64 build.

With my latest changes, to build the x86/x64 natives would use the same command as originally intended:
```
dotnet build .\scripts\native.proj
```
while to build the ARM64 natives would require setting an extra `Arm64Build` flag:
```
dotnet build .\scripts\native.proj /p:Arm64Build=true
```

mstv: You could duplicate the block at line 73-77:
```
- ps: |
    # build .NET
    $cmd = "dotnet build -c Release /p:Arm64Build=true --verbosity q --nologo /bl:.\artifacts\log\build.binlog $buildArgs"
    Invoke-Expression $cmd
    if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
```
Though it will only come into affect after merging the PR.

scripts/native_arm64.proj
chirontt: > I don't think it's possible to run multiple build pipelines on a free account. Aren't we able to build both of distros in one pipeline?

I'm out of my depth here, so I can't comment any further.

> The publishing targets are here: https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Project.Publish.targets#L271. I don't think it should be too difficult to condition `CreateMsi` target based on a MSBuild property.

Yes, it's trivial to add a condition to the `CreateMsi` target based on a property, but deciding on *which* property to use is tricky. I've tried various existing properties like `ProcessorArchitecture`, `RuntimeIdentifier`, `Platform` by adding extra options such as `-a ARM64` or `-r win-arm64` or `--use-current-runtime` to the `dotnet publish` command, but they have various side-effects in the resulting package. Like, using the `-r win-arm64` or `--use-current-runtime` on WoA would produce a package for Arm64 having similar problems as mentioned in #9990.

Eventually I decided to define my own property `Arm64Build`, so that to produce the x86/x64 distributions the original command would still work as before:
```
dotnet publish
```
while to produce Arm64 package would require setting the `Arm64Build` flag so that it would skip the `CreateMsi` step:
```
dotnet publish /p:Arm64Build=true
```
In summary, here are the complete (unchanged) commands to build GE for x86/x64 platform:
```
dotnet build
dotnet build .\scripts\native.proj
dotnet publish
```
while to build GE for WoA:
```
dotnet build
dotnet build .\scripts\native.proj /p:Arm64Build=true
dotnet publish /p:Arm64Build=true
```
Ideally, the first set of commands should have worked for both situations, without defining the new `Arm64Build` property, but I haven't found any existing MSBuild property up to the task. Any help/suggestion is quite welcome.

scripts/native.proj
```diff
@@ -39,18 +40,30 @@
 
       <_ProjectArgs32> /p:Configuration=$(Configuration) /p:Platform=Win32</_ProjectArgs32>
       <_ProjectArgs64> /p:Configuration=$(Configuration) /p:Platform=x64</_ProjectArgs64>
+      <_ProjectArgsARM64> /p:Configuration=$(Configuration) /p:Platform=ARM64</_ProjectArgsARM64>
     </PropertyGroup>
 
     <!-- Build GitExtSshAskPass project, x86 -->
     <Exec
+        Condition="!$(Arm64Build)"
```
RussKie, line 48: This is error-prone as we don't control the value. We should do this instead:
```suggestion
        Condition=" '$(Arm64Build)' != 'true' "
```

Ditto for all other similar instances.

GitExtensions/Project.Publish.targets
```diff
@@ -268,7 +268,10 @@
     Creates an MSI.
     ============================================================
     -->
-  <Target Name="CreateMsi" AfterTargets="CreatePortable">
+  <PropertyGroup>
+    <Arm64Build>false</Arm64Build>
```
RussKie, line 272: Propose a rename to
```suggestion
    <IsArm64>false</IsArm64>
```

scripts/native.proj
RussKie: That's a good idea. Few things to note:
- since we're build amd and arm we should distingush between the binlogs, hence this one should be named something like build.arm64.binlog;
- we'll need to distinguish the zips as well, i.e., name the zip something like "GitExtensions-portable.arm64.xyz0123.zip".

RussKie (commented):
👍

scripts/native.proj
chirontt: > You could duplicate the block at line 73-77:

No, it wouldn't work. The command:
```
dotnet build .\scripts\native.proj /p:Arm64Build=true
```
is meant to be executed on a WoA machine (or VM); it won't work if running on x64 platform or VM such as the one used by AppVeyor (I presume.) The reason is that the MSVC compiler for x86/x64 doesn't know how to generate Arm64 binaries.

This is why I suggested a new `appveyor.arm64.yml` to produce the Arm64 GE products, while keeping the existing `appveyor.yml` exclusively for x86/x64 GE products. But `appveyor.arm64.yml` would only work _if_ AppVeyor could run on a WoA platform, and this is a big _IF_ that I'm not sure about.

scripts/native.proj
chirontt: I disagree. `Arm64Build` is meant to be a `boolean` flag, with `true` value being the only value having significant effect; any other value is simply `not true` i.e. `false`. I'm just following MSBuild docs here. The fact that `Arm64Build` is implemented as a string is just a shortcoming of MSBuild, so making it explicit like `'$(Arm64Build)' != 'true'` is clunky and unnecessary, in my opinion, as I want to hide its silly string implementation!

GitExtensions/Project.Publish.targets
chirontt: I think a better name for the flag would be:
```
<IsArm64Build>false</IsArm64Build>
```
thus making it more meaningfully a `boolean` property.

scripts/native.proj
RussKie: Makes sense.

scripts/native.proj
RussKie: `dotnet publish /p:Arm64Build=yes` will cause this to fail.
MSBuild doesn't have strong typing, and because of that we should be defensive in our implementations. This is how comparisons are done throughout the .NET SDK.

GitExtensions/Project.Publish.targets
RussKie: Works for me.

scripts/native.proj
chirontt: > `dotnet publish /p:Arm64Build=yes` will cause this to fail.

Yes, and it should fail for Arm64 build. As I mentioned before, `Arm64Build` is a boolean flag with `true` having significant meaning, any other values are `false`.

RussKie: MSBuild properties aren't strongly typed. So they have to be compared in a safe way.

RussKie (changes requested):
As per the earlier comments

RussKie:
@aarnott if you could take this for a spin it'd be great. Also, a review is welcome ;)

AArnott:
@russkie sorry I'm just now getting to this. How do I test it? Are there binaries I can download from somewhere?

RussKie:
Sorry, no binaries, we don't have arm64 infra to build.
If you have an arm64 device you could pull this PR and see if you can build
an arm64-native app.

I think these commands should work:

dotnet build .\scripts\native.proj /p:Arm64Build=true
dotnet publish /p:Arm64Build=true

AArnott:
The first command worked. The second command failed with missing projects.

```
❯ dotnet publish /p:Arm64Build=true
MSBuild version 17.3.2+561848881 for .NET
C:\Program Files\dotnet\sdk\6.0.403\NuGet.targets(369,5): error MSB3202: The project file "C:\Users\andre\source\repos\
gitextensions\Externals\Git.hub\Git.hub\Git.hub.csproj" was not found. [C:\Users\andre\source\repos\gitextensions\GitEx
tensions.sln]
C:\Program Files\dotnet\sdk\6.0.403\NuGet.targets(369,5): error MSB3202: The project file "C:\Users\andre\source\repos\
gitextensions\Externals\conemu-inside\ConEmuWinForms\ConEmuWinForms.csproj" was not found. [C:\Users\andre\source\repos
\gitextensions\GitExtensions.sln]
C:\Program Files\dotnet\sdk\6.0.403\NuGet.targets(369,5): error MSB3202: The project file "C:\Users\andre\source\repos\
gitextensions\Externals\ICSharpCode.TextEditor\Project\ICSharpCode.TextEditor.csproj" was not found. [C:\Users\andre\so
urce\repos\gitextensions\GitExtensions.sln]
```

AArnott:
Thanks. Building was as easy as you say. But the result was an .exe that still ran as x64.

Its headers (from ILSpy) might suggest a reason:

```
// C:\Users\andre\source\repos\gitextensions\artifacts\Debug\publish\GitExtensions\GitExtensions.exe
// DotnetRuntimeBootstrapper.AppHost.Gui, Version=2.3.1.0, Culture=neutral, PublicKeyToken=null
// Global type: <Module>
// Entry point: DotnetRuntimeBootstrapper.AppHost.Gui.Program.Main
// Architecture: AnyCPU (64-bit preferred)
// Runtime: v2.0.50727
// This assembly was compiled using the /deterministic option.
// Hash algorithm: SHA1
```

It's AnyCPU. And it *appears* to target the .NET Framework 2.0 runtime, which is weird because GitExtensions.dll targets .NET 6. So maybe you're playing some weird tricks anyway. But .NET Framework AnyCPU .exes run as x64 even on an arm64 machine. You have to target arm64 specifically in an .exe to get an arm64 process.

RussKie:
Please make sure to clone the submodules.
Also the branch is out of date, we no longer use dotnetbootstrapper.

RussKie:
You're welcome to ping me in teams, if you still have troubles ;)

vbjay:
Fyi. https://git-scm.com/book/en/v2/Git-Tools-Rerere

You can prep with a merge and solve merge conflicts then reset and rebase.

AArnott:
I love `git rerere`, but as each rebased commit can introduce different conflicts, I don't think it is a guarantee that you only have to resolve conflicts once during a rebase. 
And I just don't see the value of rebasing many commits when a merge commit will do, and ensure a more reconstructible history. Part of this is a philosophical difference that probably doesn't need to happen on this github issue. 

I'd love to hear how we can get the arm64 build shipping though.

RussKie:
> This PR has 4 commits, and if there were merge conflicts, rebasing could make me resolve conflicts 4 times, whereas _merging_ only presents me with one set of conflicts.

That's not necessarily true. But in this case, as you said, it doesn't matter.



> You shouldn't need arm64 infra to build arm64 binaries. Ever hear of cross-compiling?
> And how are you going to actually build release binaries for arm64 if your PR/CI can't build arm64 binaries?

I think this is how dotnet/windowsdesktop does it, but I never actually looked into it.
We have no immediate plans to build or release arm64 binaries, I viewed this work as an enabled for folks on arm64 to be able to build it locally. I don't think anyone from the Core team (which is like 3.5 individuals, me being a 0.5) has any spare cycles to work on this initiative. If we get a contributor willing and able to help to make it build, we'll happily take it :)

AArnott:
>  I viewed this work as an enabled for folks on arm64 to be able to build it locally

Well thank you for that!

> If we get a contributor willing and able to help to make it build,

I'll be happy to look into updating your pipeline to include an arm64 build.
Any chance of _this_ PR merging so that I have something to build?

RussKie (approved):

RussKie:
Thank you @chirontt for the original contribution and @AArnott for the verification.

Issue #10662, opened 2023-01-20
Delete obsolete branches plugin - deleting stuck


jaka-logar:
### Environment

- Git Extensions 4.0.2.16100
- Build 25100ec1f7c8da8f798613da74826038af81a50e
- Git 2.39.1.windows.1
- Microsoft Windows NT 10.0.22621.0
- .NET 6.0.13
- DPI 96dpi (no scaling)
- Microsoft.WindowsDesktop.App Versions

```
    Microsoft.WindowsDesktop.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 6.0.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 7.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

Delete obsolete branches is stuck on deleting branches. Loading is spinning all the time and all you can do is to close Delete obsolete branches window.

### Steps to reproduce

- Open Delete obsolete branches plugin
- Select branch
- Click delete and confirm.

### Did this work in previous version of GitExtensions?

Before v4

matthiaslischka:
I was just writing the same issue at this moment 😅

It's stuck here:
![5ea5ed2e-18c2-473f-9f72-12786f3254d4](https://user-images.githubusercontent.com/12046115/213668114-8bc75dec-60bd-4a1d-bdd1-6bbba63c6190.gif)


But it seems to always delete exactly ONE branch - but not all. And even if only one is selected, it gets stuck.

Maybe we can fix this together with #9293. Likely related.

pmiossec:
PR #10663 should fix the issue.

jonnydee:
It seems like this bug made it back into newer versions. I am using

```
- Git Extensions 4.1.0.16698
- Build 1fe52a137acef1fbd2b653dea263b44ad5844177
- Git 2.42.0.windows.1
```

and the progress indicator is (still) indicating progress forever.

RussKie:
Please try 4.2

PR #10671, opened 2023-01-23
Improve delete obsolete branches plugin

pmiossec:
## Proposed changes

- Display an error message when one of the git command failed
- Fix bad branches found when opening the plugin form (search context still not loaded when initial search triggered)
- Refresh revision grid only when some branches has been deleted
- Slight UI improvements (increase a little more form height to display more than 2 branches lines,  better size columns)

### Before

Apparent stucked plugin (like in https://github.com/gitextensions/gitextensions/issues/10662#issuecomment-1398160549 ): 
5ea5ed2e-18c2-473f-9f72-12786f3254d4

### After

The error is displayed to the user:
image

and after the display is refreshed (with branch list refreshed) to give show to the user that the plugin is not stucked:

image

## Test methodology 

- Manually

## Test environment(s) 

- Git Extensions 33.33.33
- Build a0d4a8892a9fd5dac6113f170058101369fa950f (Dirty)
- Git 2.38.1.windows.1
- .NET 6.0.13
- DPI 96dpi (no scaling)

## Merge strategy

Merge commit (as that is multiple not linked small improvements/fixes on the delete obsolete branch plugin)

----

### DIFF
--- a/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.Designer.cs
+++ b/Plugins/DeleteUnusedBranches/DeleteUnusedBranchesForm.Designer.cs
@@ -135,7 +135,7 @@
             // 
             // _NO_TRANSLATE_deleteDataGridViewCheckBoxColumn
             // 
-            this._NO_TRANSLATE_deleteDataGridViewCheckBoxColumn.AutoSizeMode = System.Windows.Forms.DataGridViewAutoSizeColumnMode.None;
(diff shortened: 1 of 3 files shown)


pmiossec:
Side note: I would like to do a normal merge ( with a merge commit creation) for this PR. Is there a reason why the repo setting "Allow merge commits" has been disabled? /cc @RussKie
If I would express my preference, the one I dislike the most is the "rebase and merge" where I think we loose some context like the name of the branch and commits that have been developed together.

RussKie:
You can do rebase-merge that will retain the commits but will also retain
the linear history.

Alternatively, you can temporarily disable the restriction in settings.

mstv:
Alternatively, you could append " (#10671)" to all commit messages. Feature branch names are not of any interest.

PR #10679, opened 2023-01-26
Fix visual glitch in the revision grid

### LINKED ISSUE #10672: Visual glitches in the revision grid when scrolling many lines at a time with mouse wheel
### Environment

- Git Extensions 4.0.2.16100
- Git 2.38.1.windows.1
- .NET 6.0.13
- DPI 134dpi (140% scaling)

```
    Microsoft.WindowsDesktop.App 6.0.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

When scrolling the revision grid by 8 lines at a time with mouse wheel, visual glitches appear:
- Some rows are rendered out of place (shifted up or down), leading to duplicate rows and rows that are cut in half. All columns are affected.
- When selecting these out-of-place rows, they are instantly updated and rendered correctly.
- Tooltips don't seem to be affected.
- The glitch occurs nearly always (while scrolling up or down).

Please see this video in which I first scroll down, then select rows:

https://user-images.githubusercontent.com/39014485/214060645-6efe0b1b-dba9-44a5-ba13-5425d04e94c6.mp4

### Steps to reproduce

1. Configure your mouse wheel to scroll 8 lines at a time (Windows Settings > Devices > Mouse). I can reproduce the issue with any value between 6 and 20, but it becomes rare with bigger values and disappears with very big values (≳ 50).
2. With your mouse wheel scroll up or down in the revision grid.

The issue seems to be introduced by 02d5fdb0f57d5236c219f5460b2d7ea1ea56c0f1. I did not observe it with its parent fbb4000a648c9db317e29a344450a9dd672ac725.

### Diagnostics

_No response_

MaxKoll:
Fixes #10672. Although not knowing C# or the code, I could not resist investigating.

## Proposed changes

- Move the periodic repaint routine to the grid's scroll handler. Message interception via WndProc() seems to happen before the grid even had a chance to handle the message (i.e. move rows up or down). The scroll handler on the other hand seems to be called after the grid handled the scroll event.
- Counting consecutive WM_CTLCOLORSCROLLBAR messages seems to be unreliable in the case of dragging the scroll bar, because even more messages (other than WM_CTLCOLORSCROLLBAR) are being sent to the grid than when scrolling with the mouse wheel. As a result, _consecutiveScrollMessageCnt is reset to 0 before exceeding 5 most of the time, preventing forced repaints. Fix this by using a time-based approach to detect fast scrolling and force periodic repaints, which makes the GUI feel more responsive.
- I replaced
  ```
  var mm = new Message() { HWnd = Handle, Msg = NativeMethods.WM_PAINT };
  base.WndProc(ref mm);
  ```
  with
  ```
  Update();
  ```
  mostly for clarity. `Update()` triggers a repaint only if the grid is invalid. This seems to be the case after scrolling (well, it works), but please correct me if that's not the case.

## Test methodology 

I compared the scrolling behaviour against v3.5.4.12724 and the current master using this repo, and had an eye on visual appearance and performance. I used an app that emulates smooth scrolling to test the intended behaviour of the changed code segment.

## Test environment(s) 

Windows 10.0.19045.0 (134dpi, 140% scaling)
VS2022 Community
Git 2.38.1

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -25,7 +25,7 @@ namespace GitUI.UserControls.RevisionGrid
         private readonly BackgroundUpdater _backgroundUpdater;
         private readonly Stopwatch _lastRepaint = Stopwatch.StartNew();
         private readonly Stopwatch _lastScroll = Stopwatch.StartNew();
-        private readonly Stopwatch _consecutiveScroll = new();
+        private readonly Stopwatch _consecutiveScroll = Stopwatch.StartNew();
         private readonly List<ColumnProvider> _columnProviders = new();
 
         internal RevisionGraph _revisionGraph = new();
@@ -633,13 +633,9 @@ namespace GitUI.UserControls.RevisionGrid
             // is spinning fast (with free-spinning mouse wheels) or while dragging
             // the scroll bar fast. In such cases, force a repaint to make the GUI
             // feel more responsive.
-            if (_lastScroll.ElapsedMilliseconds < 100)
+            if (_lastScroll.ElapsedMilliseconds > 100)
             {
-                _consecutiveScroll.Start();
-            }
-            else
-            {
-                _consecutiveScroll.Reset();
+                _consecutiveScroll.Restart();
             }
 
             if (_consecutiveScroll.ElapsedMilliseconds > 50
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -24,6 +24,8 @@ namespace GitUI.UserControls.RevisionGrid
 
         private readonly BackgroundUpdater _backgroundUpdater;
         private readonly Stopwatch _lastRepaint = Stopwatch.StartNew();
+        private readonly Stopwatch _lastScroll = Stopwatch.StartNew();
+        private readonly Stopwatch _consecutiveScroll = new();
         private readonly List<ColumnProvider> _columnProviders = new();
 
         internal RevisionGraph _revisionGraph = new();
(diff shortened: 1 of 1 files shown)


GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -622,6 +623,35 @@ private void SetRowCountAndSelectRowsIfReady(int rowCount)
...
+            {
+                _consecutiveScroll.Start();
+            }
+            else
+            {
+                _consecutiveScroll.Reset();
+            }
+
+            if (_consecutiveScroll.ElapsedMilliseconds > 50 &&
+                _lastRepaint.ElapsedMilliseconds > 50)
```
mstv, line 646: code style (left reading):
```suggestion
            if (_consecutiveScroll.ElapsedMilliseconds > 50
                && _lastRepaint.ElapsedMilliseconds > 50)
```

mstv (approved):
Thank you

RussKie (commented):
Looks sensible. Haven't tested.

MaxKoll:
This can be simplified slightly by starting `_consecutiveScroll` on initialization like the other timers and then just doing

```
if (_lastScroll.ElapsedMilliseconds > 100)
{
    _consecutiveScroll.Restart();
}
```

instead of:

```
if (_lastScroll.ElapsedMilliseconds < 100)
{
    _consecutiveScroll.Start();
}
else
{
    _consecutiveScroll.Reset();
}
```

May I force push this? (What's best, another commit or replace the existing one?)

mstv:
You may force push. But since we are going to squash merge it, an additional commit is clearer - particularly regarding what has already been reviewed.

mstv (approved):

MaxKoll:
Nice, thanks!

Issue #9418, opened 2021-07-27
Cancel unfinished/hanging `git log`


mstv:
Most likely a git issue:
I have filtered with `Diff contains (SLOW)`. But the git command has not returned for almost 6 hours - I guess due to a child process `astextplain`. `pdf2text.exe` seems to block in conjunction with a pipe to `"C:\Program Files\Git\usr\bin\sed.exe" s/(\^M$)|(^\^L)//`.
After killing `pdf2text.exe`, another one is started for the next PDF.
These PDFs are stored in Git LFS. But the temp files have not been expanded.

This even blocks GE from switching repos.
I agree with @gerhardol: A yet running `git log` command should be canceled/killed when refreshing the RevisionGrid.

![grafik](https://user-images.githubusercontent.com/36601201/127169536-6a08d999-bc25-478f-8c94-2b98aa878326.png)

- Git Extensions 33.33.33
- Build 825d3803734c4ef29c94eb32feb42614c3f4d2f3
- Git 2.30.1.windows.1
- Microsoft Windows NT 10.0.19043.0
- .NET 5.0.5
- DPI 96dpi (no scaling)

_Originally posted by @mstv in https://github.com/gitextensions/gitextensions/issues/9393#issuecomment-887565028_

mstv:
> Diff contains `git-log -G` is not working as expected, I do not get the results I expect and the graph is incorrect. (`--parents` are not included by default, can be forced by adding a path filter like `"**/*" "**/*"`).

Perhaps yet another issue.

> There are pure GE issues too, like the left panel pumps out `git branch --format=%(refname) -a --merged HEAD` while `git-log -G` runs.

Do you think this will cause issues? Which? Due to parallel execution? It makes sense to me to run this command _if_ a revision has been selected.

> In your case I assume that you have changed .gitattributes for *.pdf. 
I assume `-G` generates the diff to find matches, this will be slow if searching all commits.

`pdf2text` _hung_ while processing the temp file which contained only the LFS link.
Git continued with the next PDF only when I killed `pdf2text`.

gerhardol:
RevisionReader has cancellation implementation, need to be added to RevisionGridControl (maybe passing the token to RevisionReader).

This was with `git-log -G`, but also normal git-log may require tenths of seconds.
The filter handling with -G can take this to extremes...

mstv:
might be solved with #9818

gerhardol:
> might be solved with #9818

[...]

PR #10680, opened 2023-01-26
Remove delay from BackgroundUpdater

MaxKoll:
I noticed this while investigating  #10672.

As I understand, the BackgroundUpdater waits for the current _operation() to complete before executing another task. Also, execution happens asynchronously. Therefore I cannot see a reason for the delay before executing the next task.

This makes the UI feel more responsive by removing a noticeable delay between scrolling the revision grid and the subsequent update of the revision graph.

## Proposed changes

Remove the delay from BackgroundUpdater.

## Test methodology 

I compared the scrolling behaviour against v3.5.4.12724 and the current master using this repo, and had an eye on visual appearance and performance.

## Test environment(s) 

Windows 10.0.19045.0 (134dpi, 140% scaling)
VS2022 Community
Git 2.38.1

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.BackgroundUpdater.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.BackgroundUpdater.cs
@@ -8,16 +8,14 @@
         private class BackgroundUpdater
         {
             private readonly Func<Task> _operation;
-            private readonly int _cooldownMilliseconds;
             private readonly object _sync = new();
 
             private volatile bool _executing;
             private volatile bool _rerunRequested;
 
-            public BackgroundUpdater(Func<Task> operation, int cooldownMilliseconds)
+            public BackgroundUpdater(Func<Task> operation)
             {
                 _operation = operation ?? throw new ArgumentNullException(nameof(operation));
-                _cooldownMilliseconds = cooldownMilliseconds;
             }
 
             public void ScheduleExcecution()
@@ -41,11 +39,6 @@
             private async Task WrappedOperationAsync()
             {
                 await _operation();
-
-                if (_rerunRequested)
-                {
-                    await Task.Delay(_cooldownMilliseconds);
-                }
 
                 lock (_sync)
                 {
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -72,7 +72,7 @@ namespace GitUI.UserControls.RevisionGrid
         {
             InitFonts();
 
-            _backgroundUpdater = new BackgroundUpdater(UpdateVisibleRowRangeInternalAsync, BackgroundThreadUpdatePeriod);
+            _backgroundUpdater = new BackgroundUpdater(UpdateVisibleRowRangeInternalAsync);
 
             InitializeComponent();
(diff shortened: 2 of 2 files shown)


mstv:
The BackgroundUpdater executes only one operation: `UpdateVisibleRowRangeInternalAsync`.
This delay (of just 25 milliseconds) has been added on purpose in order to reduce the CPU load by consecutive display updates (while loading the revisions).
The first execution is never delayed.

mstv:
> RussKie requested a review from mstv

My point has not been answered. I am not saying this part of #8809 should not be reconsidered. CC: @wischi-chr

> The `BackgroundUpdater` executes only one operation: `UpdateVisibleRowRangeInternalAsync`.
This delay (of just 25 milliseconds) has been added on purpose in order to reduce the CPU load by consecutive display updates (while loading the revisions).
The first execution is never delayed.

MaxKoll:
Hi, sorry for not getting back to you sooner.

<br>

> Do the graph updates of release 4.1.0 work well enough for you, @MaxKoll?

The glitch that made it unusable for me is gone, which I'm happy about.

I still notice the delay between grid updates (i.e. after scrolling, the grid is stuck for a tiny moment, then updates). Also, when scrolling many rows at a time, there are visual glitches while the grid is in motion, as this screenshot shows (v4.1.0.16698):

<br>

These quirks are not obstructive, though. They just make the grid feel slightly sluggish. Before 02d5fdb it felt snappy and scrolling was instant (disregarding the bug that was fixed by this commit).

<br>

> The BackgroundUpdater executes only one operation: `UpdateVisibleRowRangeInternalAsync`. This delay (of just 25 milliseconds) has been added on purpose in order to reduce the CPU load by consecutive display updates (while loading the revisions). The first execution is never delayed.

- When I removed the delay I noticed the CPU load going up a little, but not in a way I think that mattered.
- As soon as you scroll more than 1 row consecutively, the delay will be applied, I think?
- I thought, as the BackgroundUpdater is running in an async background thread, it is unable to stall the rest of the application, no matter if it is called repeatedly without delay. But as I mentioned before, I don't know .Net or C#, so that's only my intuition.

<br>

Compared to the previous updating method which used CancellationTokens (before 02d5fdb), the BackgroundUpdater method seems kind of naive: When the BackgroundUpdater is triggered, it will work on some specific rows of the grid, no matter if they are still in the view when the update finishes. The previous method seemed more convoluted, but also more fine-tuned, aborting grid updates as soon as possible when they became obsolete. I understand that using exception handling as a means of controlling the flow is generally frowned upon, which might have been one of the incentives to get rid of the previous updating method. But looking into "official" code examples regarding CancellationTokens and related topics, this seems to be the intended way of implementing it.

mstv:
We can reconsider it for v4.2.

Unfortunately, RevisionGrid loading is a construction site yet, f.i. #10957 and #9418... The previous implementation using a manually created thread had its own issues.

RussKie:
@MaxKoll is this something you are still working on?

MaxKoll:
No. And should #11046 get merged, the delay is probably not going to matter in practice (in the context of scrolling the grid). I'm fine with this being closed, if that's what you're asking.

(I still believe there _should_ be a better way than a forced delay between updates, but then again I'm unable to provide it.)

RussKie:
@MaxKoll is you still have plans for this - feel free to re-open. 
Closing as stale.

PR #10688, opened 2023-01-29
FormRebase: edit-todo

gerhardol:
## Proposed changes

Rearrange commit apply order at a rebase.
I often like to change this in bigger rebases, for instance when realizing that a series of commits should be removed.

Note: Keyboard accelerator is in conflict with "Specific range", no other char is available.
I do not think range need an accelrator at all, but could change it to R if preferred.

* Avoid multiple calls to git-ls-files --unmerged
One instead of 4 calls, each requires 40 ms in my small test repo
(Could have been separate.)
Module.InTheMiddleOfRebase() is just file I/O

### After

Adding `Edit todo...`

image

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/Commands/GitCommandHelpers.cs
+++ b/GitCommands/Git/Commands/GitCommandHelpers.cs
@@ -467,6 +467,11 @@ namespace GitCommands.Git.Commands
             return new GitArgumentBuilder("rebase") { "--abort" };
         }
 
+        public static ArgumentString EditTodoRebaseCmd()
+        {
+            return new GitArgumentBuilder("rebase") { "--edit-todo" };
+        }
+
         public static ArgumentString ResolvedCmd()
         {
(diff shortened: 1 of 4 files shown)


mstv (commented):
LGTM

> Note: Keyboard accelerator is in conflict with "Specific range", no other char is available.
I do not think range need an accelerator at all, but could change it to R if preferred.

Yes, please. Buttons with shared accelerator do not work well.

gerhardol:
> > Note: Keyboard accelerator is in conflict with "Specific range", no other char is available.
> > I do not think range need an accelerator at all, but could change it to R if preferred.
> 
> Yes, please. Buttons with shared accelerator do not work well.

Changed the existing to &n (&f and &g is possible too)

mstv (approved):

gerhardol:
merging tomorrow

PR #10743, opened 2023-02-24
Improve robustness of GitStatusMonitor

### LINKED ISSUE #9955: `Number of changed files` disappears
### Environment

- Git Extensions 33.33.33
- Build c3b0d64d8c8b90ead9b1d27571981656571aac2a
- Git 2.35.2.windows.1
- .NET 6.0.3
- DPI 96dpi (no scaling)

### Issue description

The `Number of changed files` disappears from the `Commit` button and from the Artificial Commits occasionally (too often).
The state overlay of the taskbar icon is affected, too.

Basically, the `Commit` button and the Artificial Commits should not be undecorated while the status is updated or is invalidated because a modal dialog is active.
Currently, it looks like there are no changes - very misleading.

### Steps to reproduce

If I knew

yes

### Diagnostics

_No response_

mstv:
Fixes #9955

## Proposed changes

- GitStatusMonitor: Correctly handle cancellation
  - There is no need to stop the monitor on OperationCanceledException.
  - There is no more command running if cancelled.
- GitStatusMonitor: Stop only if retry fails, too
  - E.g. I have seen the git index being locked perhaps due to VS Code getting the status.
- GitStatusMonitor: Call `InvalidateGitWorkingDirectoryStatus()`, too, on enter of `GitStatusMonitorState.Stopped`

## Test methodology 

- manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build cf19d8dc7ec512991faf8fbf8bd01bc1f7459a9c
- Git 2.39.2.windows.1
- .NET 6.0.12
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

## Merge strategy

Please do **not squash merge** this PR

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -46,10 +46,14 @@ namespace GitUI.CommandsDialogs.BrowseDialog
         /// </summary>
         private int _consecutiveErrorCount;
 
+        /// <summary>
+        /// git-status command is running that is not cancelled
+        /// </summary>
+        private bool _commandIsRunning;
+
         private readonly FileSystemWatcher _workTreeWatcher = new();
         private readonly FileSystemWatcher _gitDirWatcher = new();
         private readonly System.Windows.Forms.Timer _timerRefresh;
-        private bool _commandIsRunning;
         private bool _isFirstPostRepoChanged;
         private string? _gitPath;
         private string? _submodulesPath;
@@ -430,15 +434,15 @@ namespace GitUI.CommandsDialogs.BrowseDialog
 
                 // capture a consistent state in the main thread
                 module = Module;
-                cancelToken = _statusSequence.Next();
                 noLocks = !_isFirstPostRepoChanged;
+                cancelToken = _statusSequence.Next();
+                _commandIsRunning = true;
 
                 // Schedule periodic update, even if we don't know that anything changed
                 _nextUpdateTime = commandStartTime
                     + (PathUtil.IsWslPath(_workTreeWatcher.Path) ? PeriodicUpdateIntervalWSL : PeriodicUpdateInterval);
                 _nextEarliestTime = commandStartTime + MinUpdateInterval;
                 _isFirstPostRepoChanged = false;
-                _commandIsRunning = true;
             }
 
             ThreadHelper.JoinableTaskFactory.RunAsync(
@@ -490,21 +494,24 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                         {
                             lock (_statusSequence)
                             {
-                                if (!ModuleHasChanged() && !cancelToken.IsCancellationRequested)
+                                if (!cancelToken.IsCancellationRequested)
                                 {
-                                    // Adjust the min time to next update
-                                    int endTime = Environment.TickCount;
-                                    int commandTime = endTime - commandStartTime;
-                                    int minDelay = Math.Max(MinUpdateInterval, 2 * commandTime);
-                                    _nextEarliestTime = endTime + minDelay;
-                                    if (_nextUpdateTime - commandStartTime < _nextEarliestTime - commandStartTime)
+                                    if (!ModuleHasChanged())
                                     {
-                                        // Postpone the requested update
-                                        _nextUpdateTime = _nextEarliestTime;
+                                        // Adjust the min time to next update
+                                        int endTime = Environment.TickCount;
+                                        int commandTime = endTime - commandStartTime;
+                                        int minDelay = Math.Max(MinUpdateInterval, 2 * commandTime);
+                                        _nextEarliestTime = endTime + minDelay;
+                                        if (_nextUpdateTime - commandStartTime < _nextEarliestTime - commandStartTime)
+                                        {
+                                            // Postpone the requested update
+                                            _nextUpdateTime = _nextEarliestTime;
+                                        }
                                     }
-                                }
 
-                                _commandIsRunning = false;
+                                    _commandIsRunning = false;
+                                }
                             }
                         }
                     })
@@ -550,6 +557,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
             // Start commands, also if running already
             lock (_statusSequence)
             {
+                _commandIsRunning = false;
                 _statusSequence.CancelCurrent();
 
                 int ticks = Environment.TickCount;
--- a/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
@@ -437,6 +437,10 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                                 GitWorkingDirectoryStatusChanged?.Invoke(this, new GitWorkingDirectoryStatusEventArgs(changedFiles));
                             }
                         }
+                        catch (OperationCanceledException)
+                        {
+                            // No action
+                        }
                         catch
                         {
                             try
@@ -455,7 +459,7 @@ namespace GitUI.CommandsDialogs.BrowseDialog
                         {
                             lock (_statusSequence)
                             {
-                                if (_commandIsRunning && !ModuleHasChanged() && !cancelToken.IsCancellationRequested)
+                                if (!ModuleHasChanged() && !cancelToken.IsCancellationRequested)
                                 {
                                     // Adjust the min time to next update
                                     int endTime = Environment.TickCount;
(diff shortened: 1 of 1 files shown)


GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -469,10 +473,7 @@ private void Update()
                                     }
                                 }
 
-                                if (!cancelToken.IsCancellationRequested)
```
gerhardol, line 472: Cancellation is done in ScheduleNextInteractiveTime(), then _commandIsRunning is set to false so a new command can be started.
This could allow starting a second command, not wanted.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -424,35 +431,29 @@ private void Update()
...
                         {
+                            // Try again once
                             try
                             {
-                                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-
-                                // Avoid possible popups on every file changes
-                                CurrentStatus = GitStatusMonitorState.Stopped;
+                                await Task.Delay(InteractiveUpdateDelay);
+                                await UpdateAsync();
```
gerhardol, line 442: Instead (in addition?) to retrying, reenable when switching repos in ScheduleNextInteractiveTime()?

I would probably preferred a failure counter instead of nested updates.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
mstv: Have a look at line 249, please.

> This could allow starting a second command, not wanted.

There cannot be a command running here (it has finished or been cancelled).
What do you mean with "second command"? How could it be triggered?

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
mstv: I had to switch repos too often. In addition, it stopped working without any indication.

A failure counter is worth consideration.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -469,11 +474,33 @@ private void Update()
...
+                            }
+                        }
+
+                        return;
+
+                        async Task UpdateAsync()
+                        {
+                            try
+                            {
+                                GitExtUtils.ArgumentString cmd = GitCommandHelpers.GetAllChangedFilesCmd(true, UntrackedFilesMode.Default, noLocks: noLocks);
```
RussKie, line 487: [nit] please provide the name of the for the `true` arg.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: See line 180, called from FormBrowse.


The Git command runs async, a second invocation can be done in the timer loop.
It occasionally occurred in 3.3 or so when git-status was slow, then git-status running in parallel used all resources.

mstv: Then rather a1d7b2069102fd0d74940c9f279dce09ea75fefa and let the cancellation do the job, i.e. let it reset `_commandIsRunning`.

gerhardol: It may take longer time to start the next refresh (not starting the new until the first is cancelled), but no big difference and the code is easier to understand.

gerhardol (commented):
+1
Leave it to someone else to review the comment I had and to approve. Otherwise fine, have not run.

gerhardol:
> +1
Leave it to someone else to review the comment I had and to approve. Otherwise fine, have not run.

I still have the opinion that the addition on line 440 is the correct.
Removal of _commandIsRunning on line 458/462 may just _delay_ the next update.
IsCancelled check removal on line 472 may cause another parallel Git command
Will approve without these changes, but will not block if someone else approves.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -455,7 +490,7 @@ private void Update()
                         {
                             lock (_statusSequence)
                             {
-                                if (_commandIsRunning && !ModuleHasChanged() && !cancelToken.IsCancellationRequested)
+                                if (!ModuleHasChanged() && !cancelToken.IsCancellationRequested)
```
gerhardol, line 493: What about the following?
Keep line 521; _commandIsRunning = false;

```suggestion
                                if (cancelToken.IsCancellationRequested)
                                {
                                return;
                                }
                                
                                if (!ModuleHasChanged())
```

mstv: I haven't had the time yet to think about your comments in detail. But I have watched it for long time and think to know two things for sure:
* The GitStatusMonitor stops doing anything because of `_commandIsRunning is true` (only resolvable by switching repos).
* If we reach this `finally` block, `GitExecutable.ExecuteAsync()` _has_ finished or been cancelled, i.e. nothing is pending in GE.

What have I missed?

The `git.exe` might continue to run detached in the background. We do not need to take care, don't we?
(I started to send `Ctrl+C` and to terminate the process on timeout - unfinished.)

> Keep line 521; _commandIsRunning = false;

I understand your concern about the delay. I need to take time thinking about how to do it without mixing states.

gerhardol: Suggestion in fa36ff1147b8273b338b46647ef7b3f792d731a9
(_commandIsRunning  is more closely related to the cancellation token).

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
mstv: I have finally got how it was meant. So mainly renaming and some complements.

gerhardol (approved):
Looks fine, basically what I have been using the last days.
Have not tried to provoke errors though.

gerhardol:
The monitor stopped running once for me (in wsl), no real stress test either.
Just that someone else understands the code is a reason to merge.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
```diff
@@ -50,14 +59,23 @@ public sealed class GitStatusMonitor : IDisposable
...
+        /// Earliest time for an scheduled update (interactive requests bypasses this)
+        /// </summary>
         private int _nextEarliestTime;
 
         private GitStatusMonitorState _currentStatus;
 
+        /// <summary>
+        /// git-status command is running that is not cancelled
+        /// </summary>
+        private bool _uncancelledCommandIsRunning;
```
RussKie, line 77: "uncancelled" this looks very odd, ditto for "is running that is not cancelled" - this sounds like an oxymoron...

RussKie (commented):
The naming inhibits my ability to fully grasp the change.

GitUI/CommandsDialogs/BrowseDialog/GitStatusMonitor.cs
gerhardol: > "is running that is not cancelled" - this sounds like an oxymoron...

In a few more words: git-statis is started and the status is nether completed (with success or error) or cancelled explicitly. If false, a new git-status can be started.
I do not have a better description. the variably name can be composed from the words in the description too.

RussKie: Why the original name didn't work?

mstv: The only clear meaning of a `_commandIsRunning` would be:
```
    try
    {
         _commandIsRunning = true;
         RunTheCommand()
    }
    finally
    {
        _commandIsRunning = false;
    }
```
without any conditions.
If it deviates from this scheme, it should be reflected in the identifier - not hidden in the docs.
It should be clear, why this flag must not be reset in case of cancellation. Cancelled commands are detached and ignored. We don't want to wait for their cancellation and start the next command.
`_commandIsRunningAndUncancelled` might be clearer.

mstv: It is not a change. It is a clarification and completes the handling.

RussKie: "Uncancelled" is triggering me :) It's so Unenglish (pun intended).
How about `_commandRunningWithoutCancellation` or `_commandIsRunningAndNotCancelled`?

gerhardol:
Time to merge?

Issue #8439, opened 2020-09-08
"copy new version" and "old version" in diff view keep unwanted text


valeriob:
- Git Extensions 3.3.1.7897
- Build 5a97671645532bcedc443bac7b727f40db47cb5c
- Git 2.27.0.windows.1
- Microsoft Windows NT 10.0.19041.0
- .NET Framework 4.8.4200.0
- DPI 96dpi (no scaling)

Sometime is need to copy some changes from diff between two versions but it not works as it should - see screen - there is two option "copy new version" and "copy old version" both work but do not copy the original file, they also copy all the unwanted chars like : 

diff --git a/src/FileName.cs b/src/FileName.cs
index 6cfe43030..08096e127 100644
--- a/src/FileName.cs
+++ b/src/FileName.cs
@@ -76,7 +76,8 @@ namespace OnRtls.Flotta

mstv:
How to reproduce, @valeriob?

I had a similar issue in the Commit dialog. I marked a few lines with the mouse, then <kbd>Ctrl</kbd>+<kbd>C</kbd>.
Though I had no time to analyze (I cannot share the repo anyway).

valeriob:
Hi,
the problem is in any diff view, just select all text, right click on the selected text and choose "copy new version" or "copy old version", the thing that is copied is not the clean code.

RussKie:
![image](https://user-images.githubusercontent.com/4403806/94263678-c8e79500-ff78-11ea-887b-e5468b1577a7.png)

<details>
<summary>First option from the context menu - Copy</summary>

```
diff --git a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Tags.cs b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Tags.cs
index 9923af5b9..23c46dd08 100644
--- a/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Tags.cs
+++ b/GitUI/BranchTreePanel/RepoObjectsTree.Nodes.Tags.cs
@@ -61,7 +61,11 @@ public bool Merge()
             protected override void ApplyStyle()
             {
                 base.ApplyStyle();
-                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey = nameof(Images.TagHorizontal);
+
+                TreeViewNode.ImageKey = TreeViewNode.SelectedImageKey =
+                    Visible
+                        ? nameof(Images.TagHorizontal)
[...]

PR #10775, opened 2023-02-26
Fix: "copy new version" and "old version" in diff view keep unwanted text

RauulDev:
Fixes https://github.com/gitextensions/gitextensions/issues/8439

## Proposed changes

- Added fix to filter lines before hpos character on FileViewer.cs

## Screenshots

image

## Test methodology

- Manual

## Test environment(s)

- GIT git version 2.33.0.windows.2
- Windows 10 Pro 21H2 19044.2486

## Merge strategy

----

### DIFF
--- a/GitUI/Editor/FileViewer.cs
+++ b/GitUI/Editor/FileViewer.cs
@@ -786,9 +786,12 @@ namespace GitUI.Editor
                     code = " " + code;
                 }
 
-                var lines = code.LazySplit('\n')
-                    .Where(s => s.Length == 0 || s[0] != startChar || (s.Length > 2 && s[1] == s[0] && s[2] == s[0]));
                 var hpos = fileText.IndexOf("\n@@");
+
+                var lines = code.LazySplit('\n')
+                    .Where(s => !s.StartsWith("@@")
+                    && code.IndexOf(s) > hpos
+                    && (s.Length == 0 || s[0] != startChar || (s.Length > 2 && s[1] == s[0] && s[2] == s[0])));
 
                 // if header is selected then don't remove diff extra chars
                 if (hpos <= pos)


mstv:
The AppVeyor CI runs all tests. There is no need to share screenshots of your test runs.
Though since you have changed complex code but no test, this indicates that we do not have tests for this part of code yet.
Could you add testcases, please? In addition, the tests will help to understand your change.

Add "Fixes #8439" to the PR description, please.
Do you have a simple repro? I copy snippets often but have not been affected by the issue.

mstv:
I think the most important request is to remove the `+` and `-` in front of the lines: https://github.com/gitextensions/gitextensions/issues/8439#issuecomment-698728174.
It should work on selections, too. An empty string is copied in the current state of this PR.

We should agree on how to handle the header(s) if multiple sections of a patch are copied: https://github.com/gitextensions/gitextensions/issues/8439#issuecomment-699515380.

gerhardol:
New/old should be the only situation where +/- is discarded, otherwise the patch is copied, possibly with the header.

mstv:
> New/old should be the only situation where +/- is discarded, otherwise the patch is copied, possibly with the header.

🤔 For `Copy (Ctrl+C)`, +/- should be copied only if both of them or headers exist in the selection.
`Copy patch` shall always keep all decorations (in the selection), of course.
(If nothing is selected explicitly, the entire diff shall be used.)

Combined Diff and Range Diff need to be considered (perhaps later).

GitUI/Editor/FileViewer.cs
```diff
@@ -786,10 +786,13 @@ private void CopyNotStartingWith(char startChar)
...
                 }
 
-                var lines = code.LazySplit('\n')
-                    .Where(s => s.Length == 0 || s[0] != startChar || (s.Length > 2 && s[1] == s[0] && s[2] == s[0]));
                 var hpos = fileText.IndexOf("\n@@");
 
+                var lines = code.LazySplit('\n')
+                    .Where(s => !s.StartsWith("@@")
+                    && code.IndexOf(s) > hpos
+                    && (s.Length == 0 || s[0] != startChar || (s.Length > 2 && s[1] == s[0] && s[2] == s[0])));
```
RussKie, line 794: Please indent the lines to clearly indicate which part of the expression these two lines belong to

MatthieuHernandez (approved):

MatthieuHernandez:
Shouldn't we also remove the '-' and '+' chars at the beginning of the line when copying?

pmiossec:
@RauulDev Looking quickly at the code, it should not be too difficult to add some unit tests (the `FileViewerTests.cs` file is already there and using a test accessor https://github.com/gitextensions/gitextensions/blob/master/GitUI/Editor/FileViewer.cs#L1981 so it should be easy to test this private method) and this type of processing does lend itself on testing...

Do you think you could add some tests with meaningful input texts? Because I must admit that I'm a little lost on what we want to achieve...

Maybe updating the first post with an example of what are the outputs before and after the PR for a given input could be also a good idea...

PR #10637, opened 2023-01-14
Render graph lines with less curves

mstv:
Fixes #5782
Inspired by @spdr870's PoC 

## Proposed changes

- Render graph with diagonal lines
- Straighten multi-lane crossings by turning into diagonals
- Unfold one-lane shifts to diagonals
- Join multi-lane crossings
- Both options are active by default - but only as "Experimental" settings.

### Before / After

image image   image image

image image

## Test methodology 

- used for long time in private builds
- adapted existing test cases
- several new test cases 

## Merge strategy

Please do **not squash** merge this PR!

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10778, opened 2023-02-27
Straighten revision graph in more cases

### LINKED ISSUE #5782: Readability revision graph

o-kloster:
Fixes #5782
Improves on #9050 

## Proposed changes

- As of now, segments in the revision graph are straightened only of they contain no intermediate commits (i.e. it's just one segment). This change allows straightening also when there are intermediate commits and/or merges.

### Before

image

### After

image

## Test methodology 

- Manual with GE repo
- Unit tests verify the graph layout

## Test environment(s) 

- Git Extensions 33.33.33
- Build 8d823099f8fc7c1a80efb0bd66775b140923f377
- Git 2.36.0.windows.1 (recommended: 2.38.1 or later)
- .NET 6.0.14
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -501,6 +501,11 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
                             segmentOrAncestor = lookaheadRow.FirstParentOrSelf(segmentOrAncestor);
                         }
+
+                        if (moved)
+                        {
+                            break;
+                        }
                     }
 
                     // if moved, check again whether the lanes of the previous row can be moved, too
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -509,11 +509,15 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
                 bool MovingCausesWidthIncrease(int startIndex, int endIndex, int laneToMove)
                 {
-                    int previousRowWidth = localOrderedRowCache[startIndex - 1].GetLaneCount();
+                    int maxRowWidth = 0;
+                    for (int index = Math.Max(0, startIndex - _straightenLanesLookAhead); index < startIndex + _straightenLanesLookAhead && index < lastIndex; ++index)
+                    {
+                        maxRowWidth = Math.Max(maxRowWidth, localOrderedRowCache[index].GetLaneCount());
+                    }
 
                     for (int index = startIndex; index < endIndex; ++index)
                     {
-                        if (localOrderedRowCache[index].LaneCountAfterMovingLanesRight(laneToMove) > previousRowWidth)
+                        if (localOrderedRowCache[index].LaneCountAfterMovingLanesRight(laneToMove) > maxRowWidth)
                         {
                             return true;
                         }
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -463,26 +463,34 @@ namespace GitUI.UserControls.RevisionGrid.Graph
                     {
                         int previousLane = previousRow.GetLaneIndexForSegment(revisionGraphSegment);
                         int currentLane = currentRow.GetLaneIndexForSegment(revisionGraphSegment);
-                        if (previousLane > currentLane)
-                        {
-                            int straightenedCurrentLane = currentLane + 1;
-                            int lookaheadLane = currentLane;
-                            int nextIndex = currentIndex + 1;
-                            for (int lookaheadIndex = nextIndex; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
-                            {
-                                lookaheadLane = localOrderedRowCache[lookaheadIndex].GetLaneIndexForSegment(revisionGraphSegment);
-                                if ((lookaheadLane == straightenedCurrentLane) || (lookaheadLane > straightenedCurrentLane && previousLane == straightenedCurrentLane))
-                                {
-                                    currentRow.MoveLanesRight(currentLane);
-                                    for (; nextIndex < lookaheadIndex; ++nextIndex)
-                                    {
-                                        localOrderedRowCache[nextIndex].MoveLanesRight(currentLane);
-                                    }
 
-                                    moved = true;
-                                    break;
+                        if (previousLane <= currentLane)
+                        {
+                            continue;
+                        }
+
+                        int straightenedCurrentLane = currentLane + 1;
+                        int lookaheadLane = currentLane;
+                        int nextIndex = currentIndex + 1;
+                        var segmentOrAncestor = localOrderedRowCache[currentIndex].FirstParentOrSelf(revisionGraphSegment);
+
+                        for (int lookaheadIndex = nextIndex; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
+                        {
+                            RevisionGraphRow lookaheadRow = localOrderedRowCache[lookaheadIndex];
+                            lookaheadLane = lookaheadRow.GetLaneIndexForSegment(segmentOrAncestor);
+                            if ((lookaheadLane == straightenedCurrentLane) || (lookaheadLane > straightenedCurrentLane && previousLane == straightenedCurrentLane))
+                            {
+                                currentRow.MoveLanesRight(currentLane);
+                                for (; nextIndex < lookaheadIndex; ++nextIndex)
+                                {
+                                    localOrderedRowCache[nextIndex].MoveLanesRight(currentLane);
                                 }
+
+                                moved = true;
+                                break;
                             }
+
+                            segmentOrAncestor = lookaheadRow.FirstParentOrSelf(segmentOrAncestor);
                         }
                     }
 
--- a/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
+++ b/GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
@@ -471,10 +471,9 @@ namespace GitUI.UserControls.RevisionGrid.Graph
 
                         int straightenedCurrentLane = currentLane + 1;
                         int lookaheadLane = currentLane;
-                        int nextIndex = currentIndex + 1;
                         var segmentOrAncestor = localOrderedRowCache[currentIndex].FirstParentOrSelf(revisionGraphSegment);
 
-                        for (int lookaheadIndex = nextIndex; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
+                        for (int lookaheadIndex = currentIndex + 1; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
                         {
                             RevisionGraphRow lookaheadRow = localOrderedRowCache[lookaheadIndex];
                             lookaheadLane = lookaheadRow.GetLaneIndexForSegment(segmentOrAncestor);
@@ -490,9 +489,9 @@ namespace GitUI.UserControls.RevisionGrid.Graph
                                     break;
                                 }
 
-                                for (int index = currentIndex; index < lookaheadIndex; ++index)
+                                for (int moveIndex = currentIndex; moveIndex < lookaheadIndex; ++moveIndex)
                                 {
-                                    localOrderedRowCache[index].MoveLanesRight(currentLane);
+                                    localOrderedRowCache[moveIndex].MoveLanesRight(currentLane);
                                 }
 
                                 moved = true;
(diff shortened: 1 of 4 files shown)


gerhardol:
@o-kloster @mstv Merge this and #10637 ?

mstv:
> Merge this and #10637 ?

These two PRs are orthogonal and work well together.

---

These changes look promising. There are edge cases though - as most of the time...

The yellow area is unused space without benefit for the curves of the pink segments.
`c20fe54803f3b6875f0b19302e85d5f2b415576c` before / after:

image image

o-kloster:
I've found a few cases where the PR does not behave as desired. I'll be back with fixes.

o-kloster:
Ok, I've fixed the issues that I found:
- Straightening could lead to increasing the width of the graph.
- Straightening could lead to other segments becoming curved or shifting across more lanes than before.

The latter can also happen in the current version of GE, although more rarely. An example is at `e75b40425321694fbf6f91f6898cdec18c4b739c`:

| Today  | With PR |
| ------------- | ------------- |
| image | image  |

mstv:
Apart from cases like the following, many parts of the graph clearly look better.

4.0.2 / This PR above `885eb0c59f0f4ae11b8b510fec962b17df27ff0f`

image  image

The violet segment should be kept straight. There are no nodes in it.

> Straightening could lead to increasing the width of the graph.

This is not a strict rule IMO. I think a few additional lanes should be accepted.
In this example, the graph even has the same width.

RussKie:
> Ok, I've fixed the issues that I found:
> 
> * Straightening could lead to increasing the width of the graph.
> * Straightening could lead to other segments becoming curved or shifting across more lanes than before.
> 
> The latter can also happen in the current version of GE, although more rarely. An example is at `e75b40425321694fbf6f91f6898cdec18c4b739c`:
> 
> Today	With PR
> image	image

Looking at this specific sample, I wonder if we could do this?
image

RussKie:
> 
> 4.0.2 / This PR above `885eb0c59f0f4ae11b8b510fec962b17df27ff0f`
> 
> image image
> 
> The violet segment should be kept straight. There are no nodes in it.
> 
> > Straightening could lead to increasing the width of the graph.
> 
> This is not a strict rule IMO. I think a few additional lanes should be accepted. In this example, the graph even has the same width.

Is is possible to avoid the "bump" on the second lane? It'd allow avoiding all subsequent "bumps"...
image

mstv:
> Is is possible to avoid the "bump" on the second lane? It'd allow avoiding all subsequent "bumps"...
> image

👍 Yes, this would be a valid graph. Though only, because the reason for the "bump" is the node in the left lane.
I.e. there is no risk that the crossing segments get in conflict with a node.
It is crucial that one can distinguish whether the segments just cross each other or whether there is a junction.

The following (not occurring) graph would be unclear:
image

Since crossings and junctions cannot clearly be distinguished, your first suggestion is problematic.
With the diagonals drawn by #10637, it might be doable - as long as there are no nodes or perpendicular segments in the way as in this:
image

Anyway, both proposals affect an even lower layer ("lane creation / selection") than this PR ("lane shifting") or #10637 ("segment drawing").
So they should be addressed in follow-up PRs.

mstv:
> Conflicting files: contributors.txt

Could you file your sign off in a separate PR, please, @o-kloster?

mstv:
grafik

o-kloster:
> 4.0.2 / This PR above `885eb0c59f0f4ae11b8b510fec962b17df27ff0f`
> 
> image image
> 
> The violet segment should be kept straight. There are no nodes in it.

Fixed in latest update. The rule against increasing the width has been relaxed somewhat.

> > Straightening could lead to increasing the width of the graph.
> 
> This is not a strict rule IMO. I think a few additional lanes should be accepted.

While I agree, I'm not sure how to formulate this precisely and fear the logic might become complex. It could quickly turn into an optimization problem of width against straightness.

As for the need to keep the width in check at all, you can see an example if you run the first version of this PR on `6d5defb20f78f52c21882c71fb68029ba1b642e5` and onward:

image

o-kloster:
> > Conflicting files: contributors.txt
> 
> Could you file your sign off in a separate PR, please, @o-kloster?

Done in #10791

o-kloster:
> grafik

Which commit is this?

mstv:
There still is a too attracting condition:

`3644218d8a7954db2b29e4b1af85b564a5edc28c`

image

`ca7e05c421c7c3b80eb97dd2a785be543ceb92bf`

image

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -458,37 +458,98 @@ static void StraightenLanes(int startIndex, int lastIndex, IList<RevisionGraphRo
...
+                                for (int index = currentIndex; index < lookaheadIndex; ++index)
+                                {
+                                    localOrderedRowCache[index].MoveLanesRight(currentLane);
+                                }
+
+                                moved = true;
+                                break;
                             }
+
+                            segmentOrAncestor = lookaheadRow.FirstParentOrSelf(segmentOrAncestor);
```
mstv, line 502:
```suggestion
                            segmentOrAncestor = lookaheadRow.FirstParentOrSelf(segmentOrAncestor);
                        }
                        
                        if (moved)
                        {
                            break;
```
This was missing in the original implementation.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -458,37 +458,98 @@ static void StraightenLanes(int startIndex, int lastIndex, IList<RevisionGraphRo
...
+                            if ((lookaheadLane == straightenedCurrentLane) || (lookaheadLane > straightenedCurrentLane && previousLane == straightenedCurrentLane))
                             {
-                                lookaheadLane = localOrderedRowCache[lookaheadIndex].GetLaneIndexForSegment(revisionGraphSegment);
-                                if ((lookaheadLane == straightenedCurrentLane) || (lookaheadLane > straightenedCurrentLane && previousLane == straightenedCurrentLane))
+                                if (MovingCausesWidthIncrease(currentIndex, lookaheadIndex, currentLane))
                                 {
-                                    currentRow.MoveLanesRight(currentLane);
-                                    for (; nextIndex < lookaheadIndex; ++nextIndex)
-                                    {
-                                        localOrderedRowCache[nextIndex].MoveLanesRight(currentLane);
```
mstv, line 479: This saved one `list` lookup. Though there are a lot of other such lookups. So we can buy it for code clarity.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -458,37 +458,98 @@ static void StraightenLanes(int startIndex, int lastIndex, IList<RevisionGraphRo
...
+                                    break;
+                                }
 
-                                    moved = true;
+                                if (MovingIncreasesASegmentBend(currentRow, previousRow, currentLane))
+                                {
                                     break;
                                 }
+
+                                for (int index = currentIndex; index < lookaheadIndex; ++index)
```
mstv, line 493: nitpick:
```suggestion
                                for (int moveIndex = currentIndex; moveIndex < lookaheadIndex; ++moveIndex)
```

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
```diff
@@ -214,5 +217,57 @@ public void MoveLanesRight(int fromLane)
...
+            }
+
+            return _laneCount + 1;
+        }
+
+        /// <summary>
+        /// If this row is the parent row of <paramref name="segment"/>, returns the segment leading
+        /// to this row's first parent (if any). Otherwise, returns <paramref name="segment"/>.
+        /// </summary>
+        public RevisionGraphSegment FirstParentOrSelf(RevisionGraphSegment segment)
```
mstv, line 240: I guess it should rather be `SelfOrFirstParent` in order to keep the segment straight as far as possible.
Only if it ends, take the first parent a continuation.

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
```diff
@@ -458,37 +458,98 @@ static void StraightenLanes(int startIndex, int lastIndex, IList<RevisionGraphRo
...
+                        {
+                            continue;
+                        }
+
+                        int straightenedCurrentLane = currentLane + 1;
+                        int lookaheadLane = currentLane;
+                        int nextIndex = currentIndex + 1;
+                        var segmentOrAncestor = localOrderedRowCache[currentIndex].FirstParentOrSelf(revisionGraphSegment);
+
+                        for (int lookaheadIndex = nextIndex; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
```
mstv, line 477: in order to finish the elimination of `nextIndex`
```suggestion
                        for (int lookaheadIndex = currentIndex + 1; lookaheadLane == currentLane && lookaheadIndex <= Math.Min(currentIndex + _straightenLanesLookAhead, lastIndex); ++lookaheadIndex)
```

mstv (commented):
unfinished review

mstv:
> This branch cannot be rebased due to conflicts

I think GH is overly cautious - most likely because of `contributors.txt`. I don't see a conflict.
Could you rebase and force-push, please?

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
mstv: The first parent should only be chosen for the first child. If coming from another child, the segment should not be continued.

o-kloster:
> There still is a too attracting condition:

Can you elaborate? I don't see what is the problem here...

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
o-kloster: Done

GitUI/UserControls/RevisionGrid/Graph/RevisionGraph.cs
o-kloster: Done

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
o-kloster: > I guess it should rather be `SelfOrFirstParent` in order to keep the segment straight as far as possible. Only if it ends, take the first parent as continuation.

Are you arguing for just a name change, or for a functionality change? `SelfOrFirstParent` and `FirstParentOrSelf` are essentially the same, since the segment cannot both continue *and* have a parent in this row. But I opted for the latter since it returns self also in the case where the segment ends and has no parent segment.

> The first parent should only be chosen for the first child.

I considered this. But how do you define the 'first' child segment? By order in the `Segments` list? By lane order in the row above (is this the same?)? By distance to the row of the child revision? In each case, I think I can find an example where this would lead to less straightening than desired.

mstv: > `SelfOrFirstParent` and `FirstParentOrSelf` are essentially the same, since the segment cannot both continue *and* have a parent in this row. But I opted for the latter since it returns self also in the case where the segment ends and has no parent segment.

👍 Thank you for the explanation. I agree. 
(From just reading `if (segment.Parent != Revision)`, I do not get the meaning without looking into the docs of all these properties. Let's think about inversion and comments later when we have an agreed implementation.)  

I was just dropping guesses how to avoid the yellow segment being bent one lane to the right unnecessarily early (resulting in the free lane under the mouse cursor).
The cyan branch is branched from and back to the yellow branch - at least as I read the graph.
The cyan branch should not "attract" the yellow branch (although it would save one bend for the cyan branch without adding one to the yellow branch).

> I considered this. But how do you define the 'first' child segment? By order in the Segments list? **By lane order in the row above**?

👍
The yellow branch is the "first", the cyan one is the "second" and should not be continued with the ancestor.

> (is this the same?)

I guess yes. Though this is not guaranteed. It depends on the sorting algorithm. I am experimenting with an (immature) alternative in my private build.
Using `GetLaneIndexForSegment()` will be safe and shouldn't be too expensive.

image  image

mstv:
> The two leftmost segments (yellow and blue) should have gotten the same lane, since neither connects to the revision of the first row, right?

In theory yes. But that's not the point. This is just caused by that I prefer a clearer graph where all branches fan out at their parent: 73e6a839c89a8ef115554a4ca463a9686ccd7faa

> Could you show more of the graph? I don't see how to reproduce this.

I found it in the GE repo, too, at `083eef2e2c5030354bb33cefd14e233c58f72e41` "Merge pull request 5009":

Current | This PR

image   image

RussKie:
💭 I wonder if we should consider allowing a branch to be on the left side as well _if space allows it_, e.g.:
image

RussKie:
> Current | This PR
> 
> image image

I kind of like the left one better - it's "straighter", easier to follow.

RussKie:
So where are we with this?

mstv:
> So where are we with this?

Let's take #10842 first (perhaps after v4.1). Then `StraightenLanes()` is called more consistently.

Open:
* only switch to ancestor from 'first' child segment 
* Bulges like the following need to be straightened again yet:

image

GitUI/UserControls/RevisionGrid/Graph/RevisionGraphRow.cs
mstv: Sorry, the above seems to be not caused by this PR. The yellow and the cyan segments share the same lane since they have the same parent.
Since `StraightenLanes` operates on single segments, it cannot be detected / avoided not easily.
So I am dropping this part of the request.

mstv:
#10842 has been merged. It calls `StraightenLanes()` more consistently. There shouldn't be many conflicts when rebasing.

PR #10793, opened 2023-03-08
FIX: Periodic Background Fetch was not automatically refreshing the UI after a successful fetch

### LINKED ISSUE #9643: Periodic Background Fetch is not doing anything at all
### Environment

- Git Extensions 3.5.4.12724
- Git 2.33.0.windows.2
- .NET Framework 4.8.4400.0
- DPI 96dpi (no scaling)

### Issue description

Configuration

git call stack

Manual git fetch all works

Note: I have minimum two GitExtensions windows open at any given time. 

[Similar issue !5971](https://github.com/gitextensions/gitextensions/issues/5971)

### Steps to reproduce

Enable periodic background fetch and wait the amount of seconds defined as interval. 

Yes, a few versions back. I don't remember exactly which one.

### Diagnostics

_No response_

RemiGaudin:
Fixes #9643

## Proposed changes

- In `BackgroundFetchPlugin.RecreateObservable()`: Retrieve both standard output and standard error from the fetch command because on some configurations the detailed info are written into the error stream.

## Test methodology

- I manually tested that the fix allow now the Periodic Background Fetch to automatically refresh the UI after a successful fetch.

## Test environment(s)

- Git Extensions 4.0.2.16100
- Git 2.39.1.windows.1
- .NET 6.0.14
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
+++ b/Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
@@ -126,7 +126,7 @@ namespace GitExtensions.Plugins.BackgroundFetch
                                       string msg;
                                       try
                                       {
-                                          msg = _currentGitUiCommands.GitModule.GitExecutable.GetOutput(args);
+                                          msg = _currentGitUiCommands.GitModule.GitExecutable.Execute(args).AllOutput;
                                       }
                                       catch
                                       {
--- a/Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
+++ b/Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
@@ -126,7 +126,10 @@ namespace GitExtensions.Plugins.BackgroundFetch
                                       string msg;
                                       try
                                       {
-                                          msg = _currentGitUiCommands.GitModule.GitExecutable.Execute(args).AllOutput;
+                                          // git fetch is writing result details into standard error and not standard output, see:
+                                          // https://github.com/gitextensions/gitextensions/pull/10793
+                                          // https://lore.kernel.org/git/xmqq7cvqrdu6.fsf@gitster.g/
+                                          msg = _currentGitUiCommands.GitModule.GitExecutable.Execute(args).StandardError;
                                       }
                                       catch
                                       {
--- a/contributors.txt
+++ b/contributors.txt
@@ -189,3 +189,4 @@ YYYY/MM/DD, github id, Full name, email
 2023/02/24, RauulDev, Raul Molina, remolina7@gmail.com
 2023/03/03, pedrolamas, Pedro Lamas, pedrolamas@gmail.com
 2023/03/07, o-kloster, Oddvar Kloster, okloster(at)gmail.com
+2023/03/08, RemiGaudin, Rémi Gaudin, remi.gaudin(at)gmail.com


Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
```diff
@@ -126,7 +126,7 @@ private void RecreateObservable()
                                       string msg;
                                       try
                                       {
-                                          msg = _currentGitUiCommands.GitModule.GitExecutable.GetOutput(args);
+                                          msg = _currentGitUiCommands.GitModule.GitExecutable.Execute(args).AllOutput;
```
gerhardol, line 129: Anyone with info why stderr is used in this situation?
Generally, either stdout or stderr should be parsed. 
In this case it may be OK as "it is only a message" that is parsed, not data.

mstv (approved):
Thank you
have not run

Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
vbjay: https://github.com/git/git/commit/96890f4c428ed62cf52c11aba5773d364a8d6234#diff-a8ad3d92e86b1622eb617b0a241f2622911f8e868e3816aebb998a0f3e3bcf95R479

vbjay: https://lore.kernel.org/git/CAH74SFS3hFzYzJ6eb2YenJh4-Pf_VBzVTo8axnEtHDyaJUB-jw@mail.gmail.com/

gerhardol, line 129:
```suggestion
                                          msg = _currentGitUiCommands.GitModule.GitExecutable.Execute(args).StandardError;
```

gerhardol (commented):
+1
Suggest limiting to stderr

Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
vbjay: Agreed. But one step more. We need to run through all use of all output and do the same. Does it need to be all output or std based on what we look for.  Look at the git source code to track what output we look for is written where. It would prevent confusion. Down the road to put good comments in the use of error to mention git writes blaa to error so we read error instead of std and such.

RemiGaudin: @gerhardol I just pushed and tested your suggestion, and I also added some comments.
To me this fix is enough to solve this specific issue. I'm not sure if you want to include the full review of all output use in the scope of this PR or not, because then I'm afraid I don't have enough expertise to do that.

gerhardol (approved):

Plugins/BackgroundFetch/BackgroundFetchPlugin.cs
gerhardol: Good work, hanging on.
Now this is handled not just as a fix but with a known reason and explanation.
This could be included in a 4.0.3 (if there is one), a separate PR there later.

@vbjay Also thanks for hanging on. There were so many opinions that it was hard to find the core facts. (If I tried to get all data right before posting anything, a lot less would be done why asked in the issue as a question.)
There were some investigations before #9488, but a full check would take too much time.
In a community project like GE, the members need to be involved.

RussKie (approved):

RussKie:
I wouldn't want to release 4.0.3, instead I propose we release 4.1.

RussKie:
Thanks @RemiGaudin

PR #10799, opened 2023-03-12
RefreshGrid without revisions

### LINKED ISSUE #10720: [NBug] fatal: bad revision 'HEAD'
### Environment

- Git Extensions 4.0.2.16100
- Git 2.39.1.windows.1
- .NET 6.0.13
- DPI 96dpi (no scaling)

### Issue description

Exit code: 128
Arguments: -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --max-count=100000 --glob=refs/stas[h] HEAD --
Working directory: C:\work\repos\Activities\

```
GitExtUtils.ExternalOperationException: fatal: bad revision 'HEAD'
 ---> System.Exception: fatal: bad revision 'HEAD'
   at void GitUI.RevisionGridControl.PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)+() => { }
   at async Task GitUI.GitUIExtensions.InvokeAsync(Control control, Action action, CancellationToken token)
```

fetch -> abort -> error

### Steps to reproduce

fetch origin
fetch taking too long -> abort
error appears

_No response_

### Diagnostics

_No response_

gerhardol:
Fixes #10720

## Proposed changes

If stashes a are shown, the first stash to show requires explicit handling for the Git default "Show only current branch" to show stashes in the grid.
This is not possible in repos without a current branch.

## Test methodology 

Updated tests

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/FilterInfo.cs
+++ b/GitUI/UserControls/RevisionGrid/FilterInfo.cs
@@ -291,7 +291,7 @@ namespace GitUI.UserControls.RevisionGrid
             return searchParametersChanged;
         }
 
-        public ArgumentString GetRevisionFilter()
+        public ArgumentString GetRevisionFilter(Lazy<string> currentBranch)
         {
             if (IsRaw)
             {
@@ -303,7 +303,7 @@ namespace GitUI.UserControls.RevisionGrid
             // Separate the filters in groups
             GetCommitRevisionFilter(filter);
             GetLimitingRevisionFilter(filter);
-            GetBranchRevisionFilter(filter);
+            GetBranchRevisionFilter(filter, currentBranch);
 
             return filter;
         }
@@ -411,7 +411,7 @@ namespace GitUI.UserControls.RevisionGrid
         /// Branch revision filters, not affecting parent rewriting.
         /// </summary>
         /// <param name="filter">ArgumentBuilder arg</param>
-        private void GetBranchRevisionFilter(ArgumentBuilder filter)
+        private void GetBranchRevisionFilter(ArgumentBuilder filter, Lazy<string> currentBranch)
         {
             if (ShowOnlyFirstParent)
             {
@@ -423,31 +423,27 @@ namespace GitUI.UserControls.RevisionGrid
                 // All commits
                 filter.Add("--reflog");
             }
-            else if (IsShowCurrentBranchOnlyChecked || (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter)))
+            else if (IsShowCurrentBranchOnlyChecked && !string.IsNullOrWhiteSpace(currentBranch.Value))
             {
-                // See handling for --all below, explicitly add most recent stash
-                if (AppSettings.ShowStashes)
-                {
-                    // stash@{0} requires that the repo has commits and --glob=refs/stash is handled as refs/stash/*,
-                    // so this weird pattern must be used: "--glob=refs/stas[h]"
-                    filter.Add($"--glob={GitRefName.RefsStashPrefix.Remove(GitRefName.RefsStashPrefix.Length - 1, 1)}[h]");
-                }
+                // Git default, no option by default (stashes is special).
 
-                if (IsShowCurrentBranchOnlyChecked)
-                {
-                    // Default git-log (no option), only current branch (HEAD)
-                    if (AppSettings.ShowStashes)
-                    {
-                        filter.Add("HEAD");
-                    }
-                }
-                else
-                {
-                    // Show filtered branches only
-                    // Do not add suffix /* to the filter
-                    bool isSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
-                    filter.Add(isSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
-                }
+                AddFirstStashRef();
+
+                // Add as filter (even if Git default is current branch) as the branch (ref) must exist
+                // and the repo must contain commits, otherwise Git will exit with errors.
+                filter.Add($"--branches={GetFilterRefName(currentBranch.Value)}");
+            }
+            else if (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter))
+            {
+                // Show filtered branches only
+
+                AddFirstStashRef();
+
+                // If BranchFilter contains wildcards, "--branches=" must be prepended.
+                // Use the simple branch filter if without wildcards (must be Git reference)
+                // in order to avoid git adding implicit /* to the "--branches=" filter.
+                bool useSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
+                filter.Add(useSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
             }
             else
             {
@@ -474,6 +470,28 @@ namespace GitUI.UserControls.RevisionGrid
                 {
                     filter.Add("--boundary");
                 }
+            }
+
+            return;
+
+            // git-log filters add implicit /* to the filters, unless there are wildcard characters in the ref name
+            // get around this by adding a filter, that matches the branch name
+            string GetFilterRefName(string gitRef)
+            {
+                return $"{gitRef[..^1]}[{gitRef[^1]}]";
+            }
+
+            // Add the most recent stash to the grid
+            void AddFirstStashRef()
+            {
+                if (!AppSettings.ShowStashes)
+                {
+                    return;
+                }
+
+                // stash@{0} requires that the repo has commits and --glob=refs/stash is handled as refs/stash/*,
+                // so this weird pattern must be used: "--glob=refs/stas[h]"
+                filter.Add($"--glob={GetFilterRefName(GitRefName.RefsStashPrefix)}");
             }
         }
 
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -1048,7 +1048,7 @@ namespace GitUI
                     cancellationToken.ThrowIfCancellationRequested();
                     reader.GetLog(
                         observeRevisions,
-                        _filterInfo.GetRevisionFilter(),
+                        _filterInfo.GetRevisionFilter(CurrentBranch),
                         pathFilter,
                         cancellationToken);
                 }).FileAndForget(
--- a/UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
+++ b/UnitTests/GitUI.Tests/UserControls/FilterInfoTests.cs
@@ -941,7 +941,7 @@ namespace GitUITests.UserControls
             };
 
             filterInfo.GetSummary().Should().Be(expectedSummary);
-            filterInfo.GetRevisionFilter().ToString().Should().Be(expectedArgs);
+            filterInfo.GetRevisionFilter(new Lazy<string>(() => "currentBranch")).ToString().Should().Be(expectedArgs);
         }
 
         [TestCase(false, false, "branchFilter")]
@@ -962,7 +962,7 @@ namespace GitUITests.UserControls
                 BranchFilter = branchFilter
             };
 
-            string args = filterInfo.GetRevisionFilter();
+            string args = filterInfo.GetRevisionFilter(new Lazy<string>(() => "currentBranch"));
 
             if (showRefLog)
             {
@@ -999,7 +999,7 @@ namespace GitUITests.UserControls
                 BranchFilter = branchFilter
             };
 
-            string args = filterInfo.GetRevisionFilter();
+            string args = filterInfo.GetRevisionFilter(new Lazy<string>(() => "currentBranch"));
 
             if (expectBranches)
             {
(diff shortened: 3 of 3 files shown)


RussKie (commented):
Do we have a test that mimics the failure scenario?

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -475,6 +497,19 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
                     filter.Add("--boundary");
                 }
             }
+
+            // Add the most recent stash to the grid (see also --all handling)
+            void addFirstStashRef()
```
mstv, line 502: nit:
```suggestion
            void AddFirstStashRef()
```

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -425,8 +425,8 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
             }
             else if (IsShowCurrentBranchOnlyChecked || (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter)))
             {
-                // See handling for --all below, explicitly add most recent stash
-                if (AppSettings.ShowStashes)
+                // See handling for --all below, explicitly add only most recent stash
+                if (AppSettings.ShowStashes && (!IsShowCurrentBranchOnlyChecked || !string.IsNullOrWhiteSpace(currentBranch.Value)))
```
mstv, line 429: Reuse `AddFirstStashRef()`?

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -449,6 +450,27 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
...
+                }
+            }
+            else if (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter))
+            {
+                addFirstStashRef();
+
+                // Show filtered branches only
+                // Do not add suffix /* to the filter
+                bool isSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
+                filter.Add(isSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
```
mstv, line 472: This is too detailed for the Rule Of Three IMO, i.e factor out, please.

mstv (commented):
First minor comments.

What are the STR? I think that a pure fetch cannot destroy HEAD as stated in the issue.

gerhardol:
> What are the STR? I think that a pure fetch cannot destroy HEAD as stated in the issue.

* Bare or empty repo
* Stash is shown
* CurrentBranchOnly

mstv:
Thank you for the STR.
There seems to be yet another special case:

image

gerhardol:
> There seems to be yet another special case:

The branch must exist too. Will change this so `CurrentCheckout` in RevisionGrid is evaluated, so other Lazy functions.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: I left a temp implementation, just for you to have something to comment on - or I forgot to remove it.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: See above...

gerhardol:
> Do we have a test that mimics the failure scenario?

No relevant test.
I will add a unit test for the failure scenario in FilterInfo_GitNotes_Stashes() but that only covers some refactoring of that function, not the grid.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -475,6 +472,19 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
                     filter.Add("--boundary");
                 }
             }
+
+            // Add the most recent stash to the grid (see also --all handling)
+            void AddFirstStashRef()
```
mstv, line 477:
```suggestion

            return;

            // Add the most recent stash to the grid (see also --all handling)
            void AddFirstStashRef()
```

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -423,32 +423,29 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
...
             }
+            else if (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter))
+            {
+                // Show filtered branches only
+                AddFirstStashRef();
+
+                // Simple branch filter without wildcards
+                // Git adds the suffix "/*" implicitly if there is no wildcard
+                bool isSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
+                filter.Add(isSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
```
mstv, line 447: Should we really keep this check for simple filter? Has git changed its behavior? Could be changed here or in a follow-up.

image

mstv (approved):
You've left another nit for me :)

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: https://git-scm.com/docs/git-log#Documentation/git-log.txt---branchesltpatterngt

I assume this is a branch filter.
This PR is only changing the behavior for currentbranchonly.
The same popup, but the branch names should be validated already in the form (toolbar or advanced, normally testing the cached gitrefs first) and give a feedback that the selection is invalid.
That is a separate PR.

mstv: > This PR is only changing the behavior for currentbranchonly.

As stated:

>> Could be changed ... in a follow-up.

My point is: The comment `'Git adds the suffix "/*" implicitly if there is no wildcard'` seems to be wrong or does not apply to how the arguments are added here.

gerhardol: > My point is: The comment 'Git adds the suffix "/*" implicitly if there is no wildcard' seems to be wrong or does not apply to how the arguments are added here.

The wildcard is added with --branches, not as a ref on the command line. 
Clarification?

// Git adds the suffix "/*" implicitly if there is no wildcard for --branches argument

The same 'fix' that is used for refs/stas[h] could be used here too. Then the issue with invalid references for branches could be solved relatively simple. The drawback is that it is not possible to apply any ref (or option) in the branchbox, including multiple branches (multi could be split).
It is still possible to use messages with --init --exclude prefix (_messageAsGitOptions), but that is a special feature...

--branches=<currentbranch> is an alternative for currentbranchonly, no validation of the branch is needed (so solution is more similar to the first version).

mstv: Since we use `--branches=` only if the branch filter already contains wildcards, I do not see the relevance of the comment - although the proposed variant is clearer.
Or does it mean that we want to avoid adding `--branches=` in order to not get the implicit "/*" appended by git?
(I don't have a preference whether branch filters should be an exact match or not. I just want to understand the intention of the code.)

gerhardol: I mean that `feature` could be added as `--branches=featur[e]` similar to refs/stash

mstv: > I mean that `feature` could be added as `--branches=featur[e]` similar to refs/stash

This stops git from returning errors. 👍
But it doesn't work for commit hashes and HEAD, etc. 🤔

gerhardol: There were actually an error when a repo was empty and stashes was shown, so I changed to always use --branches for currentbranchonly.
Actually slightly simpler code.
(I thought I had tested that, I have updated Git since then, I doubt it is that though...)

To be clear, it is the Git error code we must avoid, the behavior in the app is the same. (Just to avoid the popup, no other improvement for the user.)

Filtered branches is a separate issue. It could be possible to use similar command as here, but maybe generic refs like commits should be allowed?

gerhardol: @mstv Still OK with the latest changes?
If so I will merge this and look into filtered branches (probably evaluating when exiting the text box using the ref filter to verify before rev-parse)

mstv (approved):
> If so I will merge this

OK for me (with the updated comment).
I have tried out some cases but most likely not thoroughly.

> and look into filtered branches (probably evaluating when exiting the text box using the ref filter to verify before rev-parse)

Or else use `--branches=GetFilterRefName(BranchFilter)` if without wildcards, too.
`useSimpleBranchFilter` might need to be kept but check for "--" in order to still support manually entered filter arguments.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -423,31 +423,27 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
...
+            else if (IsShowFilteredBranchesChecked && !string.IsNullOrWhiteSpace(BranchFilter))
+            {
+                // Show filtered branches only
+
+                AddFirstStashRef();
+
+                // Simple branch filter without wildcards
+                // Git adds the suffix "/*" if there is no wildcard
+                bool isSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
+                filter.Add(isSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
```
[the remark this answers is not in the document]
RussKie: [nit] Please don't wrap lines too short, 120-150 range is generally something worth striving for
``suggestion
                // If BranchFilter contains wildcards, "--branches=" must be prepended.
                // But use the simple branch filter if without wildcards in order to avoid git adding implicit /* to the "--branches=" filter then.
```

File: GitUI/UserControls/RevisionGrid/FilterInfo.cs [hunk 7]
```diff
@@ -423,31 +423,27 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
...
-                // See handling for --all below, explicitly add most recent stash
-                if (AppSettings.ShowStashes)
-                {
-                    // stash@{0} requires that the repo has commits and --glob=refs/stash is handled as refs/stash/*,
-                    // so this weird pattern must be used: "--glob=refs/stas[h]"
-                    filter.Add($"--glob={GitRefName.RefsStashPrefix.Remove(GitRefName.RefsStashPrefix.Length - 1, 1)}[h]");
-                }
+                // Git default, no option by default (stashes is special).
+                // Add as filter anyway as the branch must exist and the repo must contain commits
+                // otherwise Git will exit with errors.
```
RussKie, line 430: Ditto here

RussKie (approved):

GitUI/UserControls/RevisionGrid/FilterInfo.cs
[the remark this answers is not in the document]
gerhardol: > Please don't wrap lines too short, 120-150 range is generally something worth striving for

Avoid over 80 chars if possible, really exceptions if more than 120 chars. It is then possible to have several editors open side by side. Especially for Visual Studio, not great with multiple monitors.

Reworded slightly. This will likely be changed anyway, if a similar (but different) check is done for filtered branches.
The second line would have been 163 characters

                // If BranchFilter contains wildcards, "--branches=" must be prepended.
                // Use the simple branch filter if without wildcards (must be Git reference)
                // in order to avoid git adding implicit /* to the "--branches=" filter.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -475,6 +471,28 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter)
...
             }
+
+            return;
+
+            // git-log filters add implicit /* to the filters, unless there are wildcard characters in the ref name
+            // get around this by adding a filter, that matches the branch name
+            string GetFilterRefName(string gitRef)
+            {
+                return $"{gitRef.Substring(0, gitRef.Length - 1)}[{gitRef[^1]}]";
+            }
```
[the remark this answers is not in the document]
gerhardol: (I was expecting to get the auto suggestion here, not sure why it did not appear at first)

GitUI/UserControls/RevisionGrid/FilterInfo.cs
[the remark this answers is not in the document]
RussKie: 80?! Developers are generally constrained vertically rather than horizontally.

This is how 80 looks for me:
image

GitUI/UserControls/RevisionGrid/FilterInfo.cs
gerhardol: That line would be 136 chars, too long in my opinion.
Rewrote that too (merged the comment below), which would have resulted in a 176 char comment.

GitUI/UserControls/RevisionGrid/FilterInfo.cs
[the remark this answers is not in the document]
gerhardol: Use several editors if you do not want to use move to other view in npp.
You get much more code on your screen with that.
(I use VSC or npp at work, only with GE I use VS)

VS has some OH, but two instances are OK
image

mstv (approved):

gerhardol:
Squashed with an updated commit message.
The "filtered branches" should be fixed so there are more opportunities to comment on comments.
Merging

PR #10802, opened 2023-03-15
Replace IProcess.WaitForProcessExitAsync with WaitForExitAsync

### LINKED ISSUE #10398: minimize Application doesn't open anymore in latest version
### Environment

- GitExtensions version:
- GIT version: 
- OS version: 
- .NET version: 

### Issue description

Unable to open the application once you minimize  it

### Steps to reproduce

open the application-> minimize it-> go to the taskbar and find the git extension app and click on it

it doesn't open back again

Through the task switcher also, unable to open  (windows+tab   or Alt+tab)

yes

### Diagnostics

_No response_

mstv:
Fixes #10398

## Proposed changes

- Replace `IProcess.WaitForProcessExitAsync` with cancellable variant of `WaitForExitAsync` as proposed by @vjdw

## Screenshots 

N/A

## Test methodology 

- existing tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build 70ebb97a824f4487ca5665630275859e51559dc2
- Git 2.39.2.windows.1
- .NET 6.0.14
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 6.0.14 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 7.0.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

## Merge strategy

----

### DIFF
--- a/Plugins/GitUIPluginInterfaces/IProcess.cs
+++ b/Plugins/GitUIPluginInterfaces/IProcess.cs
@@ -49,13 +49,13 @@ namespace GitUIPluginInterfaces
         /// <summary>
         /// Returns a task that completes when the process exits, or when this object is disposed.
         /// </summary>
-        /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
+        /// <returns>A task that yields the process's exit code, or throws an exception if this object was disposed before the process exited.</returns>
         Task<int> WaitForExitAsync();
 
         /// <summary>
         /// Returns a cancellable task that completes when the process exits, or when this object is disposed.
         /// </summary>
-        /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
+        /// <returns>A task that yields the process's exit code, or throws an exception if this object was disposed before the process exited.</returns>
         Task<int> WaitForExitAsync(CancellationToken token);
 
         /// <summary>
--- a/Plugins/GitUIPluginInterfaces/IProcess.cs
+++ b/Plugins/GitUIPluginInterfaces/IProcess.cs
@@ -53,12 +53,10 @@ namespace GitUIPluginInterfaces
         Task<int> WaitForExitAsync();
 
         /// <summary>
-        /// Instructs the process component to wait for the associated process to exit, or for the cancellationToken to be cancelled (cancells the process).
-        /// Note that the exit task must be awaited too.
+        /// Returns a cancellable task that completes when the process exits, or when this object is disposed.
         /// </summary>
-        /// <param name="token">An optional token to cancel the asynchronous operation.</param>
-        /// <returns>A task that will complete when the process has exited, cancellation has been requested, or an error occurs.</returns>
-        Task WaitForProcessExitAsync(CancellationToken token);
+        /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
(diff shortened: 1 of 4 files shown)


Plugins/GitUIPluginInterfaces/IProcess.cs
```diff
@@ -53,12 +53,10 @@ public interface IProcess : IDisposable
...
 
         /// <summary>
-        /// Instructs the process component to wait for the associated process to exit, or for the cancellationToken to be cancelled (cancells the process).
-        /// Note that the exit task must be awaited too.
+        /// Returns a cancellable task that completes when the process exits, or when this object is disposed.
         /// </summary>
-        /// <param name="token">An optional token to cancel the asynchronous operation.</param>
-        /// <returns>A task that will complete when the process has exited, cancellation has been requested, or an error occurs.</returns>
-        Task WaitForProcessExitAsync(CancellationToken token);
+        /// <returns>A task that yields the process's exit code, or <c>null</c> if this object was disposed before the process exited.</returns>
```
RussKie, line 58: `null`? This would be extremely unexpected. 
Also, the signature is suggesting the result is non-nullable.

RussKie (commented):
👍 modulo the comment

RussKie:
Artifacts for validation: https://ci.appveyor.com/project/gitextensions/gitextensions/builds/46515268/artifacts

Plugins/GitUIPluginInterfaces/IProcess.cs
mstv: I've copied this from `WaitForExitAsync(void)` - without checking. :(

gerhardol (commented):
Seem to work fine for me, regarding minimization and normal use.
From what I see, the APIs are used OK.
Nice!
(RussKie comments remains, null annotation there?)

gerhardol (approved):
Great!

gerhardol:
Note: I have seen this problem also in builds with this fix. However, the explicit Restore is done by window. Not reverting directly.

The workaround https://github.com/gitextensions/gitextensions/issues/10398#issuecomment-1313378235
should probably be renamed, so it does not kick in by default after update.

gerhardol:
After this change the problem occur less often, but often enough to be really irritating.
Only for minimized applications (was maybe that before too.)
A difference is that the context menu item Restore works in the same way as clicking. When the problem occur, I have to click twice on the miniature to restore it.

vjdw:
> After this change the problem occur less often, but often enough to be really irritating. Only for minimized applications (was maybe that before too.) A difference is that the context menu item Restore works in the same way as clicking. When the problem occur, I have to click twice on the miniature to restore it.

I'm finding the same thing intermittently. It's definitely better, I can now click a second time and the window will restore. Before it would never restore.

Maybe it's just a bad idea to start an external process while Windows is waiting on Git Extensions to process a WM_ACTIVATE request?

gerhardol:
> Maybe it's just a bad idea to start an external process while Windows is waiting on Git Extensions to process a WM_ACTIVATE request?

I have a private patch to not run ls-files (to find if the repo status is changed externally) when activating, I still see the problem.
(git-status is not called when the app is minimized.)

RussKie:
I think we need to remove WM_ACTIVATE until we come up with a better solution rather than try to hack.

PR #10827, opened 2023-03-25
CommitInfoTests: Disable GitNotes

gerhardol:
## Proposed changes

If AppSettings.ShowGitNotes is set, -cimmitInfo tests will try to fetch notes which will fail (with popup).
It happened to work mostly, but slight timing changes for #8741 causes the setting to active from earlier tests.

## Test methodology 

Updated tests.

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10828, opened 2023-03-25
Priotize branches and remotes in SidePanel and CommittInfo

### LINKED ISSUE #8741: Improve access to important branches
## Feature description

(Not only) the left panel can be overcrowded with foreign remote branches, which are of little interest for me - especially in long running projects with many contributors. (I know branch naming conventions could help but not fully.)
I want to see an overview of important branches (filtered by regex), e.g. "main", "develop", "release/xx.yy".
Either by tinkering the sort order or rather by showing them again under an additional node. E.g. new nodes "Important" and "All" could be inserted under the "Remotes" node.

This filter could also be applied to "Contained in branches:" of the CommitInfo panel.

Thoughts?

## Environment

- Git Extensions 33.33.33
- Build f96fa18202704df0606d41ff2f5c24e7f7ccc107
- .NET Framework 4.8.4300.0
- DPI 96dpi (no scaling)

gerhardol:
Fixes #8741 
Depends on #10827

Draft, in addition to dependencies I want review on functionality limitation, especially by @mstv The commits are done in such way so the temp solution me and mstv has been using can be reviewed, other can review all commits at once.

## Proposed changes

Prioritize branches and remotes in the sidepanel and commit info.
Regexes (separated by ';') sets the priority, controling the order in the sidepanel.
For CommittInfo, all matching regexes have

Limitations:
- CommittInfo handles all matches with the same priority. It should not make much difference as the branches for a commit is limited. The implementation is regardless "private" for CommittInfo.
- Branches and remotes priorities are not applied anywhere else. Is this desired at all? If so branch order could be made available in the Lazy handling in GitUIEventArgs and some generic extension for remotes.
  * Default order when pushing could be interesting
- No tests, not sure what to add.

### Before

image

image
image

image

### After

image
image

image
image

image

## Merge strategy

- To be decided later.
The maintainer may still request the contributor to squash and rebase, to make sure that merges and commit messages are clarified.

----


GitCommands/Settings/AppSettings.cs
```diff
@@ -1928,6 +1928,18 @@ public static int RepoObjectsTreeStashesIndex
             set => SetInt("RepoObjectsTree.StashesIndex", value);
         }
 
+        public static string PrioritizedBranchNames
+        {
+            get => GetString("PrioritizedBranchNames", "main.*|master.*|release/.*");
```
RussKie, line 1933: Why are `main.*` and `master.*`?

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,56 @@ internal override void UpdateVisibility()
...
+        private IEnumerable<T> OrderByPriority<T>(IEnumerable<T> references, Func<T, string> keySelector, string setting)
+        {
+            // Sort prio branches first (if set) with the compile cache (no need to instanceiate)
+            var regexes = setting.Split(";", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            if (!regexes.Any())
+            {
+                return references;
+            }
+
+            int currPrio = 0;
```
RussKie, line 120: Please don't abbreviate, use full names

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,56 @@ internal override void UpdateVisibility()
...
+            Dictionary<string, int> prios = new();
+            foreach (string regex in regexes.Reverse())
+            {
+                currPrio++;
+                references.ForEach(node =>
```
RussKie, line 125: Suggest using `foreach` instead - it's faster as it has no additional layers of indirection.

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,56 @@ internal override void UpdateVisibility()
...
+                    string key = keySelector(node);
+                    if (Regex.IsMatch(key, $"^({regex})$", RegexOptions.ExplicitCapture))
+                    {
+                        prios[key] = currPrio;
+                    }
+                });
+            }
+
+            // Set prio on all branches, let no match be 0
+            return references.OrderBy(node => prios.TryGetValue(keySelector(node), out int prio) ? -prio : 0);
```
RussKie, line 136: We're re-enumerating the underlying collection; this is not good. `references` has to be a materialised collection like `List<>`.

GitUI/BranchTreePanel/RemoteBranchTree.cs
```diff
@@ -31,7 +31,8 @@ protected override Nodes FillTree(IReadOnlyList<IGitRef> branches, CancellationT
             ConfigFileRemoteSettingsManager remotesManager = new(() => Module);
 
             // Create nodes for enabled remotes with branches
-            foreach (IGitRef branch in branches)
+            // If prio is set, exclude "refs/remotes/<remote>/"
```
RussKie, line 34: Don't abbreviate

GitUI/BranchTreePanel/RemoteBranchTree.cs
```diff
@@ -69,24 +70,23 @@ protected override Nodes FillTree(IReadOnlyList<IGitRef> branches, CancellationT
             }
 
             // Add enabled remote nodes in order
-            enabledRemoteRepoNodes
-                .OrderBy(node => node.FullPath)
+            bool isRemoteRegex = !string.IsNullOrWhiteSpace(AppSettings.PrioritizedRemoteNames);
+            PriotizedRemotes(enabledRemoteRepoNodes)
                 .ForEach(node => nodes.AddNode(node));
```
RussKie, line 75: Use `foreach` instead.

RussKie (commented):
The label looks misaligned:
image

I think a propertygrid could work better here - it would allow future additions with minimal effort (at least as far as UI is concerned), and allow utilising the multiline text editor control - each pattern would be defined on it's own line instead of trying to visually parse the string and see where `;` are.

It'd be really beneficial to have tests for the new functionality.

GitCommands/Settings/AppSettings.cs
gerhardol: Regex for common most important branches.
Like master, master_v3 etc

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: The list is materialized.
The core performance is required for the regex second for the sort anyway.

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: OK, setting to currentPrioritiy, but this name has no importance, i is an alternative.

GitUI/BranchTreePanel/RemoteBranchTree.cs
gerhardol: Removing this comment, not adding info here

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: ok

gerhardol:
Rebased as #10827 is merged, added review comments.

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: With 500 local branches and 3 prio levels, this add a few ms for local branches first times, no noticable time for refresh while the 15 entry static cache in .NET is utilized
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,51 @@ internal override void UpdateVisibility()
 
             base.UpdateVisibility();
         }
+
+        protected IEnumerable<IGitRef> PriotizedBranches(IReadOnlyList<IGitRef> branches)
```
mstv, line 88:
```suggestion
        protected IEnumerable<IGitRef> PrioritizedBranches(IReadOnlyList<IGitRef> branches)
```
plus a few more occurrences including the settings page.

mstv (commented):
The current checkout should always have the highest priority in CommitInfo, shouldn't it?

Branch priorities seem to be applied to the Left Panel only but not to the CommitInfo, e.g. for "release/.*;main|master":

image

> I think a propertygrid could work better here - it would allow future additions with minimal effort (at least as far as UI is concerned), and allow utilising the multiline text editor control - each pattern would be defined on it's own line instead of trying to visually parse the string and see where ; are.

👍

gerhardol:
> The current checkout should always have the highest priority in CommitInfo, shouldn't it?

That would be a special handling.
Currentbranch name would need to be included from RevGrid to Side Panel. (to not delay with some 100 ms.

> Branch priorities seem to be applied to the Left Panel only but not to the CommitInfo, e.g. for "release/.*;main|master":

As displayed in the help (and limitations). I do not think it is worth it to implement priorities in CommitInfo

> image
> 
> > I think a propertygrid could work better here - it would allow future additions with minimal effort (at least as far as UI is concerned), and allow utilising the multiline text editor control - each pattern would be defined on it's own line instead of trying to visually parse the string and see where ; are.

I do not want to do that right now.
Better to change all at once.
I expect the default will be to not use multiple priorities,

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: Cannot spell Prioritize....

gerhardol:
@mstv Can I squash for an official review?

mstv:
> Can I squash for an official review?

Yes, I do not have objections.

GitUI/CommandsDialogs/SettingsDialog/Pages/SortingSettingsPage.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace GitUI.CommandsDialogs.SettingsDialog.Pages
+{
+    public partial class SortingSettingsPage : SettingsPageWithHeader
+    {
+        private readonly TranslationString _revisionSortWarningTooltip = new("Sorting revisions may delay rendering of the revision graph.");
+        private readonly TranslationString _prioBranchNamesTooltip = new("Regex to prioritize branch names in the side panel and commit info.\n" +
+            "The branches matcing the pattern will be shown before the others.\n" +
+            "Separate the priorities with ';'.");
+        private readonly TranslationString _prioRemoteNamesTooltip = new("Regex to prioritize remote names in the side panel.\n" +
+            "The remotes matcing the pattern will be shown before the others.\n" +
```
mstv, line 16:
```suggestion
            "The remotes matcing the pattern will be shown before the others.\n" +
```
```suggestion
            "The remotes matching the pattern will be shown before the others.\n" +
```
Can we assign the tooltips in the designer?

mstv:
>> The current checkout should always have the highest priority in CommitInfo, shouldn't it?

> That would be a special handling.

I was not aware that the behavior is the same as in v4.0.2.
It is a separate issue because `"(no branch)" != "(HEAD detached at ...)"`. The according link to the revision is broken, too.

GitCommands/Settings/AppSettings.cs
RussKie: 🤣 "master_v2_final"

GitUI/BranchTreePanel/BaseRefTree.cs
RussKie: > The list is materialized.

No. `remotes.OrderBy(node => node.FullPath)` is an enumeration which isn't materialised. We enumerate the underlying collection once at line:120 and then re-enumerate it again at line:131.

GitCommands/Settings/AppSettings.cs
gerhardol: The intention is to have regex for integration branches where there is development. GE is not doing active development in other than master.
A different team could have developed in master_mono still cherry picking interesting changes.

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: references is materialised by the caller (it is in the "cache").

For the output: Do you want to return List<T> with ToList() in this function or let the users call ToList()?

GitUI/CommandsDialogs/SettingsDialog/Pages/SortingSettingsPage.cs
gerhardol: PictureBox like RevisionSortOrderHelp does not have ToolTip, they are assigned with ToolTip.SetToolTip() 
So not easily

Fixing the spellings

gerhardol:
> > > The current checkout should always have the highest priority in CommitInfo, shouldn't it?
> 
> > That would be a special handling.
> 
> I was not aware that the behavior is the same as in v4.0.2. It is a separate issue because `"(no branch)" != "(HEAD detached at ...)"`. The according link to the revision is broken, too.

IScriptHostControl could be used (rather than Module.GetSelectedBranch() that may execute a Git command that is not needed).

To sort the current branch first in side panel would require a few more steps, doable. Is this a request?

Squashed with updated commit message, removing draft

mstv:
> IScriptHostControl could be used (rather than Module.GetSelectedBranch() that may execute a Git command that is not needed).

will consider it

> To sort the current branch first in side panel would require a few more steps, doable. Is this a request?

No, not a request. Could be a feature but not here. There are cons though.
I just wondered why it did not work for CommitInfo.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,12 +801,13 @@ protected override void DisposeCustomResources()
...
-                RegexOptions.Compiled | RegexOptions.CultureInvariant);
-            private static readonly Regex _remoteMasterRegex = new($"^{RemoteBranchPrefix}.*/(main|master)[^/]*$",
-                RegexOptions.Compiled | RegexOptions.CultureInvariant);
+            private const string _remoteBranchPrefix = "remotes/";
+            private static readonly Regex _importantRepoRegex = new($"^{_remoteBranchPrefix}({AppSettings.PrioritizedRemoteNames.Trim().Replace(';', '|')})/",
+                RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant);
+            private static readonly Regex _remoteImportantBranchRegex = new($"^{_remoteBranchPrefix}[^/]+/({AppSettings.PrioritizedBranchNames.Trim().Replace(';', '|')})$",
+                RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant);
+            private static readonly Regex _localImportantBranchRegex = new($"^({AppSettings.PrioritizedBranchNames.Trim().Replace(';', '|')})$",
+                RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant);
```
mstv, line 810: Modified settings would require a restart in order to come into effect.

mstv (commented):
just tuning

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,51 @@ internal override void UpdateVisibility()
...
         }
+
+        protected IEnumerable<IGitRef> PrioritizedBranches(IReadOnlyList<IGitRef> branches)
+        {
+            return OrderByPriority(branches, node => node.LocalName, AppSettings.PrioritizedBranchNames);
+        }
+
+        protected IEnumerable<RemoteRepoNode> PrioritizedRemotes(IReadOnlyList<RemoteRepoNode> remotes)
+        {
+            return OrderByPriority(remotes.OrderBy(node => node.FullPath), node => node.FullPath, AppSettings.PrioritizedRemoteNames);
```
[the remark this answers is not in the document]
gerhardol: Branches are sorted with the Git options (last used, alphabetical etc), remotes are not so a default order is needed.

GitUI/CommitInfo/CommitInfo.cs
gerhardol: Add to tooltip?

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,51 @@ internal override void UpdateVisibility()
...
+                    string key = keySelector(node);
+                    if (Regex.IsMatch(key, $"^({regex})$", RegexOptions.ExplicitCapture))
+                    {
+                        prios[key] = currentPrioritiy;
+                    }
+                }
+            }
+
+            // Set prio on all branches, let no match be 0
+            return references.OrderBy(node => prios.TryGetValue(keySelector(node), out int prio) ? -prio : 0);
```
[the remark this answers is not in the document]
mstv: > I believe a higher prio is more important.

I have removed `.Reverse` and the `-prio`.

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +84,51 @@ internal override void UpdateVisibility()
...
+        private IEnumerable<T> OrderByPriority<T>(IEnumerable<T> references, Func<T, string> keySelector, string setting)
+        {
+            // Sort prio branches first (if set) with the compile cache (no need to instanceiate)
+            var regexes = setting.Split(";", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            if (!regexes.Any())
+            {
+                return references;
+            }
+
+            int currentPrioritiy = 0;
```
mstv, line 115:
```suggestion
            int currentPriority = 0;
```
yet another variant :)

GitUI/CommitInfo/CommitInfo.cs
mstv: > Add to tooltip?

🤔 Would be imperfect, because it makes my day when I write a correct regex in the first attempt. On the other hand, no complex regexes are expected here.
Users will wonder why it takes effect for the side panel and not in the commit info.

GitUI/BranchTreePanel/BaseRefTree.cs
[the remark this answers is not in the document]
gerhardol: I see this, but the most important priority is now 0 and least MaxInt
Rename to precedence? queue?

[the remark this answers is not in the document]
gerhardol: order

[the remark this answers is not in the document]
mstv: I am used to non-negative priorities with 0 being the highest in some contexts.
I would not switch to another name (although `precedence` would be good).
You can decrement the priority if you prefer.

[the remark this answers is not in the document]
gerhardol: Also needed to change the handling, checking if prio is set already, no break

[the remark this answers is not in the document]
mstv: > Also needed to change the handling, checking if prio is set already, no break

What I had not noticed when I suggested, was which `foreach` is the inner and the outer loop.
`break` should be possible after swapping the `foreach`s. This avoids the `ContainsKey(key)`, too.

`prios` is `orderByNodeKey` then. (I do not insist in the full name if you do not like it.)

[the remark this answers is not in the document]
gerhardol: Some brief performance measurements with stopwatch in the actual code.
512 local branches (284 remote branches, 6 active remotes, 90 inactive, only brief look at that data)
3 branched normally in branch filter, 19 in max test

0-2 ms (5000 - 15000 ticks) with 3 regex
9-31 ms with 19 regex

With regex in inner loop
Similar for 3 regex
70-210 ms for 19
So the potential worst case is much worse

Already 30 ms is a lot, even if this is done async, even if that is an extreme case, the cache could be used by other and add a similar hit for a realistic number (a couple of priorities, 1 by default).
210 ms is too much.

Setting Regex.CacheSize = 30; decreases the worst case time to 5-15 ms. Similar for "regex inner loop" (maybe typically lower for "regex outer loop" but I have not setup a realistic test case and run it enough times).
Making sure the cache is larger than the no of regex improves a lot, adding that and can change to "regex inner loop".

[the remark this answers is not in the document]
mstv: Thank you for the measurements.
`Regex.IsMatch(key, $"^({regex})$", ...)` looks kind of  complex. What about creating a `List<Regex>` before the nested loops?

gerhardol:
@mstv I pushed a commit to use priorities also in commit info. This will also decrease the size for the cache (with one entry!)
I am not tempted to write tests about that so I do not plan to include it in the PR myself...

GitUI/BranchTreePanel/BaseRefTree.cs
RussKie: > references is materialised by the caller (it is in the "cache").

I can't see that happening.
```cs
        private IEnumerable<T> OrderByPriority<T>(IEnumerable<T> references...)
        {
            // ...

            if (!regexes.Any())
            {
// returns the non-enumerated dataset
                return references;
            }

            // ...

// we're now enumerating the underlying dataset
            foreach (T node in references)
            {
                // ...
            }

// and now we're enumerating the underlying dataset again!
            // Order by the sort match, with no match last as int.Max
            return references.OrderBy(node => orderByNodeKey.GetValueOrDefault(keySelector(node), int.MaxValue));
        }
```

RussKie:
What's puzzling me is that we're sorting at "display" level instead of sorting the data when we get it. Is there something that prevents us from sorting refs once we got those from git?

gerhardol:
> What's puzzling me is that we're sorting at "display" level instead of sorting the data when we get it. Is there something that prevents us from sorting refs once we got those from git?

branches and remotes are slightly different, mostly describing branches

For most refs we try to use the 'cache' so the refs are only retrieved once (each such call can take seconds worst case). The data is sorted by Git according to the user configuration. In most situations, this data do not have to be sorted at all. (Similar to sorting git-log, it will delay the grid display).
But the code in BaseRefTree could be moved to the cache.
For remotes, there is no cache but a helper function could be global.

The handling in CommitInfo is slightly separate, the sorting depends on several condition (as local/remote branches are mixed).
But part of the 'order score' could well be cached.

So sure it could be changed but it will be more complex...

RussKie (approved):

mstv (approved):
Thank you

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,84 @@ protected override void DisposeCustomResources()
...
+                    _orderByBranch[branch] = GetBranchOrder(branch);
+                }
+
+                return;
+
+                // Get the order for each branch.
+                // Add max possible order value to next "level" to sort properly with the order for each regex.
+                int GetBranchOrder(string branch)
+                {
+                    return branch == _currentBranch ? 0
```
mstv, line 830:
```suggestion
                    return (_isDetachedHead ? DetachedHeadParser.IsDetachedHead(branch) : branch == _currentBranch) ? 0
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,84 @@ protected override void DisposeCustomResources()
...
+                            if (Regex.IsMatch(branch, regex, RegexOptions.ExplicitCapture))
+                            {
+                                order = currentOrder;
+                                return true;
+                            }
+
+                            currentOrder++;
+                        }
+
+                        order = int.MaxValue;
```
mstv, line 878:
```suggestion
                        order = currentOrder;
```
could save the accesses to `.Length` in the distinction of cases above. But since this cannot be used in all cases, it reads clearer with `.Length`.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,84 @@ protected override void DisposeCustomResources()
...
+
+                                    // Not prioritized local branch
+                                    ? 0
+                                    : TryGetOrder(branch, remoteRegexes, out remoteOrder)
+
+                                    // Not prioritized branch on prioritized remote (note that remoteOrder was not evaluated previously)
+                                    ? 1 + remoteOrder
+
+                                    // Non match, return int.Max
+                                    : int.MaxValue - 1 - localBranchRegexes.Length - (remoteBranchRegexes.Length * (1 + remoteRegexes.Length)))));
```
mstv, line 860: Calculation to `int.MaxValue` is not necessary. More consistent:
```suggestion
                                    : 1
                                    + (
                                          TryGetOrder(branch, remoteRegexes, out remoteOrder)

                                          // Not prioritized branch on prioritized remote
                                          ? remoteOrder
                                          : remoteRegexes.Length))));
```

GitCommands/Settings/AppSettings.cs
```diff
@@ -1934,6 +1934,18 @@ public static int RepoObjectsTreeStashesIndex
             set => SetInt("RepoObjectsTree.StashesIndex", value);
         }
 
+        public static string PrioritizedBranchNames
+        {
+            get => GetString("PrioritizedBranchNames", "main[^/]*|master[^/]*|release/.*");
```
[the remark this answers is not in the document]
gerhardol: The default is also an example of what can be done. But only one prio for performance reasons.

GitUI/CommitInfo/CommitInfo.cs
gerhardol: using that in some situations

GitUI/CommitInfo/CommitInfo.cs
gerhardol: Used a variant of this, 

Also changing the sort rules slightly to simplify implementation and maybe explanation of the feature. (I will see all 'master' for GitHub remotes that I am not interested higher in the list, but no big deal).
Separate commit, can reverse.

gerhardol:
Some performance measurements: 
For recent commits with a few branches, 0-2 ms are required. This is not changed much by changes, minor.
Current in the PR: Five years ago with 731 branches, about 8 ms.

Currently in master the regex were running more often when sorting, so 50-80 ms was needed.
Before the cache for order compare number was added with "multi prio in commit info", about 50-90 ms was required.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,87 @@ protected override void DisposeCustomResources()
...
+                                    // prioritized branch, order per branch prio then remote (with not prioritized last)
+                                    (remoteBranchOrder * remoteRegexes.Length) + GetOrder(branch, remoteRegexes)
+                                : (remoteBranchOrder * (remoteRegexes.Length + 1))
+                                + (
+                                    IsLocalBranch()
+
+                                    // Not prioritized local branch
+                                    ? 0
+
+                                    // order by priotized remotes with not prioritized last
```
mstv, line 854:
```suggestion
                                    // order by prioritized remotes with not prioritized last
```

UnitTests/GitUI.Tests/UserControls/CommitInfo/BranchComparerTests.cs
```diff
@@ -7,26 +8,35 @@ public class BranchComparerTests
     {
         [SetCulture("en-US")]
         [SetUICulture("en-US")]
-        [Test]
-        public void BranchComparer([Values(null, "", "current", "(no branch)")] string? currentBranch)
+        [TestCase(null)]
+        [TestCase("")]
+        [TestCase("current")]
+        [TestCase("(no branch)")]
+        public void BranchComparer(string? currentBranch)
```
mstv, line 15: (`Values` is useful, particularly if having multiple arguments (creates the cross-product).)

GitUI/BranchTreePanel/BaseRefTree.cs
```diff
@@ -80,5 +82,69 @@ internal override void UpdateVisibility()
...
+            }
+
+            // A lot of regexes will push out entries from the static regex cache, increasing the time a lot (several hundred ms).
+            // (Switching the regex to the outer loop will decrease the effect but the code structure is simpler this way).
+            // The static .NET regex cache must be able to include at least all regexes in the loop,
+            // ideally also all common use in a "refresh" loop, so increase the size.
+            // A few usages in branches vs remote in this method, CommitInfo adds usage for
+            // split length of PrioritizedBranchNames (for remotes) and PrioritizedRemoteNames,
+            // a few usages in submodule status processing etc.
+            const int additionalRegexCacheEntries = 6;
```
mstv, line 123: Why so sparing? Are cache entries such expensive?

mstv (approved):
> Also changing the sort rules slightly to simplify implementation and maybe explanation of the feature.

Sensible

GitUI/BranchTreePanel/BaseRefTree.cs
gerhardol: Probably not expensive, but 6+regex.Length is likely sufficient.
CommitInfo is maybe a better place to check this, or some other centralized init (at startup should be sufficient.). 
I guess the cost is when there additional usages without changing the formula.

Can change it to 10+regex.Length with a note that this should be done at startup.

UnitTests/GitUI.Tests/UserControls/CommitInfo/BranchComparerTests.cs
gerhardol: I was tinkering with some more test cases, this was clearer. I can reverse.

GitUI/CommitInfo/CommitInfo.cs
gerhardol: so tired....

UnitTests/GitUI.Tests/UserControls/CommitInfo/BranchComparerTests.cs
mstv: > I was tinkering with some more test cases, this was clearer. I can reverse.

No need to reverse. That's why I set it in parens. I just wanted to mention it.

gerhardol:
Will push one minor change for CommitInfo that decreases the sort time down to 2 ms, so we can be done with this PR...

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,91 @@ protected override void DisposeCustomResources()
...
+                                TryGetOrder(branch, remoteBranchRegexes, out int remoteBranchOrder)
+
+                                // prioritized branch
+                                ? 0
+
+                                // after non priority local branches
+                                : 1)
+
+                            // group by branch prio then remote (with not prioritized as an extra level)
+                            + (remoteBranchOrder * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes));
```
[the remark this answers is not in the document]
gerhardol: > I fear that (? 0 : 1) + ... + GetOrder(branch, remoteRegexes) overlaps.

Carefully choosen...
The tests should cover that. Can maybe add some some more border names like remote AAA

> Nit: I do not grasp the intention behind the indents and the linebreaks after the opening paren

This is complex, some consideration needed - why this conversation is needed.

If intended, it was harder to see the common adding.
But will rewrite to use if() instead, to have clear indentation rules.
Multiple `order +=` is intentional, to not subindent the ternary statement.
ternary `... ? 0  : ...` could be `if(...){...}` but I want to have the comments (could add them below maybe)
  
```c#
           int GetBranchOrder(string branch)
            {
                int order = 0;
                if (_isDetachedHead ? DetachedHeadParser.IsDetachedHead(branch) : branch == _currentBranch)
                {
                    return order;
                }

                order++;
                if (IsLocalBranch())
                {
                    order += TryGetOrder(branch, localBranchRegexes, out int localBranchOrder)

                        // Matching local branch
                        ? 0
                        
                        // After matching remote branches
                        : prioritizedRemoteBranchesLength();
                    order += localBranchOrder;
 
                    return order;
                }

                order += localBranchRegexes.Length;
                order += TryGetOrder(branch, remoteBranchRegexes, out int remoteBranchOrder)

                    // prioritized branch
                    ? 0

                    // after non priority local branches
                    : 1;

                // group by branch prio then remote (with not prioritized as an extra level)
                order += (remoteBranchOrder * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes);

                return order;
```

gerhardol:
Squashed and rebased without changes.
Plan to merge tomorrow.

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,99 @@ protected override void DisposeCustomResources()
...
+                    // Remote branches
+                    order += localBranchRegexes.Length;
+                    if (!TryGetOrder(branch, remoteBranchRegexes, out int remoteBranchOrder))
+                    {
+                        // after non priority local branches
+                        order += 1;
+                    }
+
+                    // Group by branch prio then remote (with not prioritized remotes as an extra level)
+                    order += (remoteBranchOrder * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes);
```
mstv, line 861:
```suggestion
                    const int branchRemoteVersaLocalBranches = 2;
                    order += (remoteBranchOrder * (remoteRegexes.Length + 1)) + (branchRemoteVersaLocalBranches  * GetOrder(branch, remoteRegexes));
```

GitUI/CommitInfo/CommitInfo.cs
```diff
@@ -801,20 +801,99 @@ protected override void DisposeCustomResources()
...
+
+                        return order;
+                    }
+
+                    // Remote branches
+                    order += localBranchRegexes.Length;
+                    if (!TryGetOrder(branch, remoteBranchRegexes, out int remoteBranchOrder))
+                    {
+                        // after non priority local branches
+                        order += 1;
```
mstv, line 857: nit
```suggestion
                        order++;
```
although I prefer the prefix increment operator `++order` because of runtime costs with e.g. iterators (in C++, which should rather have been named ++C), and it reads "Increment variable" like functions.

mstv (commented):
Much easier to read 👍

GitUI/CommitInfo/CommitInfo.cs
gerhardol: This is intentional, to have the pattern "order +=", it just happen to be one here

...and C++ is a better name than ++C, it is next language that will be better. That is the price for backwards compatibility.

GitUI/CommitInfo/CommitInfo.cs
gerhardol: Why?
Will affect `prioritizedRemoteBranchesLength()` too
It is not like the order has to be without "gaps", but I see no point here.

GitUI/CommitInfo/CommitInfo.cs
gerhardol: Can add a constant for the 1 and do the same after 'current branch' check

GitUI/CommitInfo/CommitInfo.cs
mstv: Yes, OK, now I see that it is the same `+ 1` in `order += 1` and in `remoteBranchOrder * (remoteRegexes.Length + 1)`.
A constant might emphasize this.

```
                    if (!TryGetOrder(branch, remoteBranchRegexes, out int remoteBranchOrder))
                    {
                        // after non priority local branches
                        order += 1;
                    }

                    // Group by branch prio then remote (with not prioritized remotes as an extra level)
                    order += (remoteBranchOrder * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes);
```
is either
`order += 1 + (remoteBranchRegexes.Length * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes)`
or
`order += 0 + (remoteBranchOrder * (remoteRegexes.Length + 1)) + GetOrder(branch, remoteRegexes);`

mstv (approved):

mstv (approved):
👍

mstv (approved):
approving both with and without the last commit regarding "Left Panel" vs. "Side Panel"

PR #10831, opened 2023-03-25
Decouple reflog from other branch filters

gerhardol:
Fixes #10830

## Proposed changes

Decouples --reflog from branch filters as --all
The options for e.g. AllBranches and RefLog can appear at the same time.

See also #10172 that decoupled other options

Some updates to be done in the doc too I assume.

### Before

image
image

image

### After

image

image

## Test methodology 

Tests updated

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10832, opened 2023-03-25
Filtered branches: Check validility of filter toolbar branch filter box

### LINKED ISSUE #10740: [NBug] fatal: bad revision '142480'
### Environment

- Git Extensions 4.0.2.16100
- Git 2.38.1.windows.1
- .NET 6.0.14
- DPI 96dpi (no scaling)

### Issue description

Exit code: 128
Arguments: -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --max-count=60000 --glob=refs/stas[h] 142480 --
Working directory: C:\src\cb5\

```
GitExtUtils.ExternalOperationException: fatal: bad revision '142480'
 ---> System.Exception: fatal: bad revision '142480'
   at void GitUI.RevisionGridControl.PerformRefreshRevisions(Func<RefsFilter, IReadOnlyList<IGitRef>> getRefs, bool forceRefresh)+() => { }
   at async Task GitUI.GitUIExtensions.InvokeAsync(Control control, Action action, CancellationToken token)
```

Upon entering part of the branch name "142480" into filtering field and pressing Enter

### Steps to reproduce

Upon entering part of the branch§'s name "142480" into filtering field and pressing Enter

_No response_

### Diagnostics

_No response_

gerhardol:
Depends on #10831 - only last commit to be reviewed
There is no functional requirement, but it is easier to not have to resolve merge conflicts multiple times.

Fixes #10740

## Proposed changes

- Validate the contents in the filtered branches text box. If a revision is not valid, show a popup and ignore the revision.
- Handle multiple revisions (branches) was not properly none when handling options

There is no check in the Advanced filter form. This is partly due to that GitModule and the cached branches/refs are not available which require some more step, but also to allow use cases that may have been missed.

The data is not cleared in the box, so the user can edit. But no further feedback that the data is ignored.

The text box is still a mess, should be improved...
(since .net5 migration I believe.)

See also #10168, not included in a released version yet

### Before

After refresh

image

### After

Enter in text box

image

## Test methodology 

- Tests added
Note: There is no test for invalid revisions when the popup appear.

## Merge strategy

After dependencies are merged:

----

### DIFF
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -69,6 +69,9 @@ namespace GitUI.UserControls
             {
                 List<string> newFilter = new();
                 IReadOnlyList<IGitRef> refs = _getRefs(RefsFilter.NoFilter);
+
+                // Split at whitespace (char[])null is default) but with split options.
+                // Ignore quouting, Git revisions do not allow spaces.
                 foreach (string branch in filter.Split((char[])null, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries))
                 {
                     bool wildcardBranchFilter = branch.IndexOfAny(new[] { '?', '*', '[' }) >= 0;
--- a/GitUI/UserControls/FilterToolBar.cs
+++ b/GitUI/UserControls/FilterToolBar.cs
@@ -51,7 +51,7 @@ namespace GitUI.UserControls
         /// <summary>
         ///  Applies custom branch filters supplied via the filter textbox.
         /// </summary>
-        private void ApplyCustomBranchFilter()
+        private void ApplyCustomBranchFilter(bool checkBranch = true)
         {
             if (_isApplyingFilter)
             {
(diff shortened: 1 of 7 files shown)


GitUI/UserControls/RevisionGrid/FilterInfo.cs
```diff
@@ -440,10 +447,16 @@ private void GetBranchRevisionFilter(ArgumentBuilder filter, Lazy<string> curren
                 AddFirstStashRef();
 
                 // If BranchFilter contains wildcards, "--branches=" must be prepended.
-                // Use the simple branch filter if without wildcards (must be Git reference)
+                // Use the simple branch filter if without wildcards (must be Git revision)
                 // in order to avoid git adding implicit /* to the "--branches=" filter.
-                bool useSimpleBranchFilter = BranchFilter.IndexOfAny(new[] { '?', '*', '[' }) == -1;
-                filter.Add(useSimpleBranchFilter ? BranchFilter : $"--branches={BranchFilter}");
+                // Also add apparent options
+                foreach (string branch in BranchFilter.Split((char[])null, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries))
```
mstv, line 453: I have noticed the draft state after I wrote the comment:
Is this a one-liner for `.Trim` + `!.IsNullorEmpty`?
If really want to split, then care needs to be taken of arguments containing strings in `"` or `'`...

gerhardol: Draft due to dependencies, not contents.

Split is on whitespace, null is the default. 

Space cannot occur in git revisions but quotes could be cleaned.

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -65,6 +65,46 @@ private void ApplyCustomBranchFilter()
 
             // Apply the textbox contents, no check if the (multiple) options is in tscboBranchFilter.Items (or that the list is generated)
             string filter = tscboBranchFilter.Text == TranslatedStrings.NoResultsFound ? string.Empty : tscboBranchFilter.Text;
+            if (checkBranch && !string.IsNullOrWhiteSpace(filter))
+            {
+                List<string> newFilter = new();
+                IReadOnlyList<IGitRef> refs = _getRefs(RefsFilter.NoFilter);
+                foreach (string branch in filter.Split((char[])null, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries))
```
RussKie, line 72: Can you please help me understand `filter.Split((char[])null, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)`?

RussKie (commented):
Looking at the second commit only - I don't fully understand the mechanics, but I can follow the gist of it. 👍

GitUI/UserControls/FilterToolBar.cs
gerhardol: Like Split() with options how to split.
Null argument is all whitespace

GitUI/UserControls/FilterToolBar.cs
```diff
@@ -69,6 +69,9 @@ private void ApplyCustomBranchFilter(bool checkBranch = true)
             {
                 List<string> newFilter = new();
                 IReadOnlyList<IGitRef> refs = _getRefs(RefsFilter.NoFilter);
+
+                // Split at whitespace (char[])null is default) but with split options.
+                // Ignore quouting, Git revisions do not allow spaces.
```
mstv, line 74:
```suggestion
                // Ignore quoting, Git revisions do not allow spaces.
```

mstv (commented):
+2 LGTM

GitUI/UserControls/FilterToolBar.cs
gerhardol: I cannot spell anymore....

GitUI/UserControls/RevisionGrid/FilterInfo.cs
RussKie: You can no undraft it :)

RussKie (approved):

PR #10836, opened 2023-03-26
Fixup GetAllBranchesWhichContainGivenCommit

### LINKED ISSUE #10819: CommitInfo: "Contained in branches" splits branch name at '+' character
### Environment

- GitExtensions version: 4.0.2

### Issue description

branch names containing "+" character(s) are incorrectly split in the CommitInfo panel at least

### Steps to reproduce

Create branches named "a", "b", merge them to "a+b"
Select "a"; look at "Contained in branches:
actual: "a, a, b"
expected: "a, a+b"

I doubt it

### Diagnostics

_No response_

mstv:
Fixes #10819

## Proposed changes

- correctly split git output, i.e. only at line ends
- remove first two columns only if in expected format "* ", "+ " or "  "

## Screenshots 

n/a

## Test methodology 

- extended existing testcase

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3146,7 +3146,13 @@ namespace GitCommands
             // Remove symlink targets as in "origin/HEAD -> origin/master"
             for (int i = 0; i < result.Length; i++)
             {
-                string item = Regex.Replace(result[i], @"^[*+ ] ", "");
+                string item = result[i];
+
+                // remove prepended branch state "* ", "+ ", "  "
+                if (item.Length >= 2 && item[1] == ' ' && item[0] is (' ' or '*' or '+'))
+                {
+                    item = item[2..];
+                }
 
                 if (getRemote)
                 {
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3141,12 +3141,12 @@ namespace GitCommands
                 return Array.Empty<string>();
             }
 
-            var result = exec.StandardOutput.Split(new[] { '\r', '\n', '*', '+' }, StringSplitOptions.RemoveEmptyEntries);
+            string[] result = exec.StandardOutput.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
 
             // Remove symlink targets as in "origin/HEAD -> origin/master"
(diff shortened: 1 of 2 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -3141,12 +3141,12 @@ public IReadOnlyList<string> GetAllBranchesWhichContainGivenCommit(ObjectId obje
                 return Array.Empty<string>();
             }
 
-            var result = exec.StandardOutput.Split(new[] { '\r', '\n', '*', '+' }, StringSplitOptions.RemoveEmptyEntries);
+            string[] result = exec.StandardOutput.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
```
RussKie, line 3144: It would be good to add a comment explaining what information git may provide and what transformation we do.
IIRC, `*` denotes the current branch, but I can't remember what `+` may be used for...

RussKie (approved):

GitCommands/Git/GitModule.cs
gerhardol: @RussKie  that would be a link to Git documentation
Documentation is poor, does not statate that the line starts with two chars before branch
So something should be added below (this line is fine).

https://git-scm.com/docs/git-branch#_description

+ is used for linked work trees, similar to *

GitCommands/Git/GitModule.cs
```diff
@@ -3141,12 +3141,12 @@ public IReadOnlyList<string> GetAllBranchesWhichContainGivenCommit(ObjectId obje
...
 
             // Remove symlink targets as in "origin/HEAD -> origin/master"
-            for (var i = 0; i < result.Length; i++)
+            for (int i = 0; i < result.Length; i++)
             {
-                var item = result[i].Trim();
+                string item = Regex.Replace(result[i], @"^[*+ ] ", "");
```
gerhardol, line 3149: This should maybe be trimmed with code instead, or we should increase the static regex cache (15 entries)
I do not request changes though

As discussed above, a comment about the format should be added.

gerhardol (approved):

GitCommands/Git/GitModule.cs
gerhardol: All good changes

gerhardol (approved):

PR #10840, opened 2023-03-27
Improve blame performance

pmiossec:
## Proposed changes

 Some small improvements on blame data generation:
    * cache author lines instead of building it every times
(because it happens often that multiple lines belong to same commit)
    * Use TryGetValue to get the commit and avatar image
    * better heuristic for initial line capacity

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 25c13145f0e10caa840621380f8cf1c2441819de (Dirty)
- Git 2.38.1.windows.1
- .NET 6.0.14
- DPI 96dpi (no scaling)

## Merge strategy

Merge commit

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -3378,7 +3378,10 @@ namespace GitCommands
             // is a blank line, and third is an introductory paragraph about the project.
 
             Dictionary<ObjectId, GitBlameCommit> commitByObjectId = new();
-            List<GitBlameLine> lines = new(capacity: 256);
+
+            // Pre-allocate the list with a capacity estimated from the approximate git blame length to describe a file line
+            const int GitBlameLengthPerLineHeuristicValue = 120;
+            List<GitBlameLine> lines = new(capacity: Math.Min(Math.Max(256, output.Length / GitBlameLengthPerLineHeuristicValue), 5000));
 
             bool hasCommitHeader;
             ObjectId? objectId;
@@ -3417,7 +3420,7 @@ namespace GitCommands
                     {
                         // TODO quite a few nullable suppressions here (via ! character) which should be addressed as they hint at a design flaw
 
-                        if (!commitByObjectId.ContainsKey(objectId!))
+                        if (!commitByObjectId.TryGetValue(objectId!, out GitBlameCommit? commitData))
                         {
                             commit = new GitBlameCommit(
                                 objectId!,
@@ -3435,7 +3438,6 @@ namespace GitCommands
                         }
                         else
                         {
-                            var commitData = commitByObjectId[objectId!];
                             if (filename == commitData.FileName)
                             {
                                 commit = commitData;
--- a/GitUI/UserControls/BlameControl.cs
+++ b/GitUI/UserControls/BlameControl.cs
@@ -365,6 +365,7 @@ namespace GitUI.Blame
             string emptyLine = new(' ', lineLength);
             Dictionary<string, Image?> cacheAvatars = new();
             var noAuthorImage = (Image)new Bitmap(Images.User80, avatarSize, avatarSize);
+            Dictionary<ObjectId, string> authorLineCache = new();
             for (var index = 0; index < _blame.Lines.Count; index++)
             {
                 var line = _blame.Lines[index];
@@ -379,9 +380,9 @@ namespace GitUI.Blame
                     {
                         if (authorEmail is not null)
                         {
-                            if (cacheAvatars.ContainsKey(authorEmail))
+                            if (cacheAvatars.TryGetValue(authorEmail, out Image? avatarImage))
                             {
-                                gitBlameDisplays[index].Avatar = cacheAvatars[authorEmail];
+                                gitBlameDisplays[index].Avatar = avatarImage;
                             }
                             else
                             {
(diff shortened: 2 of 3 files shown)


pmiossec:
Future work: During my investigation, I saw that
https://github.com/gitextensions/gitextensions/blob/master/GitUI/Avatars/AvatarDownloader.cs#L74 is **really** slow.

For example, like I said in PR description when no avatar is in cache, blame on `GitModule.cs` takes:
- 6m40s when downloading avatars
- 6s when generating the initials!!!

But a quick test replacing the use of `WebClient` by httpclient (⚠️ without proxy configuration!) reduce the time to around 50s (instead of more than 6 min). I 🤞 that it's not the proxy configuration that introduce the perf cost but there is some investigation to do...

I have not replaced the code because I don't know how to reproduce the same proxy configuration than with using `WebClient`.
If someone knows how to do it....

GitCommands/Git/GitModule.cs
```diff
@@ -1,3 +1,4 @@
+using System.Buffers;
```
mstv, line 1: I guess it can be removed.

GitUI/UserControls/BlameControl.cs
```diff
@@ -379,9 +380,9 @@ private void ProcessBlame(string? filename, GitRevision revision, IReadOnlyList<
                     {
                         if (authorEmail is not null)
                         {
-                            if (cacheAvatars.ContainsKey(authorEmail))
+                            if (cacheAvatars.TryGetValue(authorEmail, out var avatarImage))
```
mstv, line 383:
```suggestion
                            if (cacheAvatars.TryGetValue(authorEmail, out Image? avatarImage))
```

GitUI/UserControls/BlameControl.cs
```diff
@@ -398,11 +399,14 @@ private void ProcessBlame(string? filename, GitRevision revision, IReadOnlyList<
                         }
                     }
 
-                    BuildAuthorLine(line, lineBuilder, dateTimeFormat, filename, AppSettings.BlameShowAuthor, AppSettings.BlameShowAuthorDate, AppSettings.BlameShowOriginalFilePath, AppSettings.BlameDisplayAuthorFirst);
+                    if (!authorLineCache.TryGetValue(line.Commit.ObjectId, out var authorLine))
```
mstv, line 402:
```suggestion
                    if (!authorLineCache.TryGetValue(line.Commit.ObjectId, out string authorLine))
```

mstv (commented):
+2 LGTM

> I have been using this commit for some months

Me, too.

GitCommands/Settings/AppSettings.cs
```diff
@@ -588,7 +588,7 @@ public static bool ShowAuthorAvatarInCommitInfo
 
         public static AvatarProvider AvatarProvider
         {
-            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.Default);
+            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.None);
```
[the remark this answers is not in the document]
gerhardol: I suggest this is separated from this PR.

gerhardol (approved):
(approved but agree with mstv, should be changed)

pmiossec:
Done.

mstv (approved):

GitCommands/Git/GitModule.cs
```diff
@@ -3378,7 +3378,7 @@ internal GitBlame ParseGitBlame(string output, Encoding encoding)
             // is a blank line, and third is an introductory paragraph about the project.
 
             Dictionary<ObjectId, GitBlameCommit> commitByObjectId = new();
-            List<GitBlameLine> lines = new(capacity: 256);
+            List<GitBlameLine> lines = new(capacity: Math.Min(Math.Max(256, output.Length / 120), 1000));
```
RussKie, line 3381: What are these magic numbers?

RussKie:
> * Disable fetching avatar by default (because it really too slow: 6m40s with avatars / 6s without)

Can you please point me to where we're doing this?

> But a quick test replacing the use of `WebClient` by httpclient (⚠️ without proxy configuration!) reduce the time to around 50s (instead of more than 6 min). I 🤞 that it's not the proxy configuration that introduce the perf cost but there is some investigation to do...
> 
> I have not replaced the code because I don't know how to reproduce the same proxy configuration than with using `WebClient`. If someone knows how to do it....

I think as a first step we can provide an implementation without a proxy, and then bolt on a proxy support. Unless, of course, you're behind proxy and need it :)
I don't remember when was the last time I was behind a proxy, so I won't be much help here.

A quick search for how to configure proxy with HttpClient yields these: https://stackoverflow.com/a/58560252/2338036 and https://stackoverflow.com/a/55242207/2338036.

RussKie (commented):
👍

GitCommands/Git/GitModule.cs
pmiossec: It's a heuristic to size the list (to prevent multiple reallocation) depending on the size of the file content to blame with some limits to not allocate a too small list (256 as before) or too big (1000) one.
To 'estimate' the number of lines, I took a number bigger than what nearly all projects will have as mean source code length to guarantee that we will not over-allocate (because we blame nearly 90% of the time source code, right!?!) which is a length of 120 characters...

gerhardol: It would be good to add some hints about this.
I remember looking at this line some minutes when I tried this commit out to understand it.
(Not affecting approval.)

pmiossec: I have added a comment...

gerhardol (approved):

RussKie (approved):

Issue #5492, opened 2018-09-30
Support for .gitconfig conditional includes


markashleybell:
**Do you want to request a *feature* or report a *bug*?**

I imagine this would be a feature due to the amount of work that's probably involved, although arguably the current behaviour is misleading and could be classed as a bug.

**What is the current behavior?**

I use [conditional includes](https://git-scm.com/docs/git-config#_includes) to separate my personal projects from work for my employer. Because of this, my `.gitconfig` doesn't contain a `[user]` section; it's included dynamically based on the folder I'm working in:

### In `.gitconfig`

    [includeIf "gitdir:Work/"]
        path = .gitconfig-work
    [includeIf "gitdir:Personal/"]
        path = .gitconfig-personal

### `.gitconfig-work`

    [user]
        name = Mark Bell
        email = [MY WORK EMAIL]

### `.gitconfig-personal`

    [user]
        name = Mark Bell
        email = [MY PERSONAL EMAIL]

This *works* in GitExtensions (in that any git operations I perform use the correct credentials), but the settings validator displays an error:

![Settings Dialog](https://user-images.githubusercontent.com/483470/46253668-b87bad00-c479-11e8-9fe2-1fca39383d5b.PNG)

And the committer display in the status bar of the commit dialog is always empty:

![Committer](https://user-images.githubusercontent.com/483470/46253818-18c01e00-c47d-11e8-9196-0749623ee011.PNG)

As I mentioned, the correct credentials actually are used when you make a commit etc. (presumably because GE is just using the git cli under the hood), but this is a bit disconcerting when you first encounter it!

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

See above.

**What is the expected behavior?**

The commit dialog committer display should show the correct committer for the current repository folder, taking any conditional includes into account.

For the settings dialog, I'm not even clear what the behaviour _should_ be... perhaps a nested box under the user box? This would also affect the user name and email fields in the **Git Config** settings.

**Environment you encounter the issue:**

- GitExtensions version: 2.51.05
- GIT version: 2.18
- OS version: Windows 10
- .NET version: 4.72

vbjay:
![image](https://user-images.githubusercontent.com/3629489/46256564-65eec080-c47a-11e8-9ad4-09daea9a3479.png)
[...]

PR #10843, opened 2023-03-29
FormCommit: Read committer info using git

mstv:
Fixes https://github.com/gitextensions/gitextensions/issues/5492#issuecomment-1448460964

## Proposed changes

- FormCommit: Read committer info using git,
  use `RunAsync` and
  do not cache the git output in order to update on FormActivate (existing testcase)

## Screenshots 

unchanged

## Test methodology 

- manual
- existing tests

## Merge strategy

----

### DIFF
--- a/GitCommands/Git/GitModule.cs
+++ b/GitCommands/Git/GitModule.cs
@@ -2318,6 +2318,12 @@ namespace GitCommands
             return EffectiveConfigFile.GetValue(setting);
         }
 
+        public string GetEffectiveGitSetting(string setting, bool cache = true)
+        {
+            GitArgumentBuilder args = new("config") { "--includes", "--get", setting };
+            return GitExecutable.GetOutput(args, cache: cache ? GitCommandCache : null).Trim();
+        }
+
         public void UnsetSetting(string setting)
         {
             SetSetting(setting, null);
--- a/GitUI/CommandsDialogs/FormCommit.cs
+++ b/GitUI/CommandsDialogs/FormCommit.cs
@@ -2819,14 +2819,21 @@ namespace GitUI.CommandsDialogs
 
         private void UpdateAuthorInfo()
         {
-            var userName = Module.GetEffectiveSetting(SettingKeyString.UserName);
-            var userEmail = Module.GetEffectiveSetting(SettingKeyString.UserEmail);
+            ThreadHelper.JoinableTaskFactory.RunAsync(
+                async () =>
+                {
+                    await TaskScheduler.Default;
 
-            var committer = $"{_commitCommitterInfo.Text} {userName} <{userEmail}>";
+                    // Do not cache results in order to update the info on FormActivate
+                    string userName = Module.GetEffectiveGitSetting(SettingKeyString.UserName, cache: false);
+                    string userEmail = Module.GetEffectiveGitSetting(SettingKeyString.UserEmail, cache: false);
+                    string committer = $"{_commitCommitterInfo.Text} {userName} <{userEmail}>";
 
-            commitAuthorStatus.Text = string.IsNullOrEmpty(toolAuthor.Text?.Trim())
-                ? committer
-                : $"{committer} {_commitAuthorInfo.Text} {toolAuthor.Text}";
+                    await this.SwitchToMainThreadAsync();
+                    commitAuthorStatus.Text = string.IsNullOrEmpty(toolAuthor.Text?.Trim())
+                        ? committer
+                        : $"{committer} {_commitAuthorInfo.Text} {toolAuthor.Text}";
+                });
         }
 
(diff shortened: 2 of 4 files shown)


GitCommands/Git/GitModule.cs
```diff
@@ -2313,8 +2313,14 @@ public string GetSetting(string setting)
             return LocalConfigFile.GetValue(setting);
         }
 
-        public string GetEffectiveSetting(string setting)
+        public string GetEffectiveSetting(string setting, bool useGit = false, bool useCache = false)
```
gerhardol, line 2316: Use a separate function instead?

gerhardol (commented):
+1
This increases the startup time for the first start of FormCommit with about 150 ms.
Annoying...
The cache should probably be cleared when exiting settings form.

GitCommands/Git/GitModule.cs
RussKie: Yes please. Let's call it `GetEffectiveGitSetting()`. Cache by default until we have a real use case that requires no caching.

RussKie (commented):
Tests simulating the reported scenario would be great to have.

RussKie:
Should the setting be reloaded when a repo is switched?

vbjay:
> Should the setting be reloaded when a repo is switched?

Yes. Config in other repo can be different.

mstv:
> This increases the startup time for the first start of FormCommit with about 150 ms.

I am going to do it async.

> The cache should probably be cleared when exiting settings form.

good point

> > Should the setting be reloaded when a repo is switched?
> 
> Yes. Config in other repo can be different.

I thought we do so already - and was wrong again as in a discussion longer ago. I am adding a non-static cache.

> Tests simulating the reported scenario would be great to have.

I am going to adapt the existing tests as far as necessary.
My interest in this feature is very limited though. I just see the necessity to correctly display the info. Let's leave it up to git.
I hope @Gutza will verify it. Could you setup an integration testcase?

mstv:
> I am adding a non-static cache.

All current usages of the `GitModule.GitCommandCache` seem to be repo-specific as far as I can see.
So, having only a static instance is not correct for these.
Are there objections turning it into a member instance (other than the binding in FormGitCommandLog)?

BTW: Is a cache entry really such expensive that we limit the MRU cache to 50 entries?

GitUI/CommandsDialogs/FormCommit.cs
```diff
@@ -2819,14 +2819,21 @@ private void CommitAndPush_Click(object sender, EventArgs e)
...
+                    // Do not cache results in order to update the info on FormActivate
+                    string userName = Module.GetEffectiveGitSetting(SettingKeyString.UserName, cache: false);
+                    string userEmail = Module.GetEffectiveGitSetting(SettingKeyString.UserEmail, cache: false);
+                    string committer = $"{_commitCommitterInfo.Text} {userName} <{userEmail}>";
 
-            commitAuthorStatus.Text = string.IsNullOrEmpty(toolAuthor.Text?.Trim())
-                ? committer
-                : $"{committer} {_commitAuthorInfo.Text} {toolAuthor.Text}";
+                    await this.SwitchToMainThreadAsync();
+                    commitAuthorStatus.Text = string.IsNullOrEmpty(toolAuthor.Text?.Trim())
```
gerhardol, line 2833: ?
```suggestion
                    commitAuthorStatus.Text = string.IsNullOrWhitespace(toolAuthor.Text))
```

gerhardol (approved):
Have not run, but it looks fine.

mstv:
> Let's call it `GetEffectiveGitSetting()`.

Thank you for directly proposing a name, Igor.

> Cache by default until we have a real use case that requires no caching.

I agree. Caching should be the default for settings.
And we already have a use case: The committer info shall be updated on FormActivate as per existing testcase.

So, since this PR does not depend on caching (`RunAsync` is the key), this PR would be ready if AppVeyor would be able to run Form tests without headstands. 
Any idea how to make `Should_update_committer_info_on_form_activated` pass?

mstv:
Sorry AppVeyor, I have found where to place the missing wait.

mstv:
Does https://ci.appveyor.com/api/buildjobs/whg8axsj2xdhj86v/artifacts/artifacts%2FRelease%2Fpublish%2FGitExtensions-Portable-4.1.0.16396-f9631e23c.zip work for you, @Gutza?

RussKie (approved):

Gutza:
> Does https://ci.appveyor.com/api/buildjobs/whg8axsj2xdhj86v/artifacts/artifacts%2FRelease%2Fpublish%2FGitExtensions-Portable-4.1.0.16396-f9631e23c.zip work for you, @Gutza?

Yes, it works wonderfully, thank you!

Issue #10020, opened 2022-06-01
Using the `git blame` function can be very slow


XMuli:
### Environment

- Git Extensions 3.5.3.12551
- Build 00604a59a673142a3d34a22aa657466a922977f3
- Git 2.35.1.windows.2
- Microsoft Windows NT 10.0.19044.0
- .NET Framework 4.8.4510.0
- DPI 96dpi (no scaling)

### Issue description

**Gitextensions is Very good software, I like it a lot. **

But every time I use the `git blame` function, I get stuck for a long time, and clicking on it will often result in a grayed out UI waiting for a response. If you optimize this feature, you can give up using `TortoiseGit` altogether.

But not as fast as `TortoiseGitBlame.exe`, every time this feature, will be used with this software extra.

### Steps to reproduce

as shown：
![image](https://user-images.githubusercontent.com/33887845/171371640-2999dfc0-0b16-4182-b5f4-f2a74dae7b30.png)

![image](https://user-images.githubusercontent.com/33887845/171371662-da78b048-017c-4005-9c30-3527867a7d45.png)

### Did this work in previous version of GitExtensions?

OK, very slow too.

### Diagnostics

OK, but very slow.

XMuli:
I wonder if it has anything to do with the size of my warehouse?

The same test with a large repository (10G), TortoiseGitBlame.exe is also able to be very fast.

Come on (ง -_-)ง

![20220127154637](https://user-images.githubusercontent.com/33887845/171372718-92d2c51d-b8ff-4b74-9457-20a69e928e09.gif)

XMuli:
I usually look in the explorer's right-click menu to see its modification history (`blame`).

tool log should be this：?

![image](https://user-images.githubusercontent.com/33887845/171596687-78363644-a755-4b0d-83f5-41be6dd82a76.png)


thanks, friend

RussKie:
No.
What's the experience when you run the same command from git cli?

XMuli:
This picture is perhaps what I want to express:

![image](https://user-images.githubusercontent.com/33887845/171600552-bf81e34c-d15c-4870-af39-18f1a70c2e05.png)

XMuli:
> What's the experience when you run the same command from git cli?

The speed is fast. But it is displayed in the terminal (pwsh7.2)

pmiossec:
@XMuli 

I think that what @RussKie is asking you (and also me) is the time to run the blame command which should be a command like this one:

[...]

PR #10846, opened 2023-03-31
Improve avatar handling

pmiossec:
Improvements over #10020

Note: a lot easier to review commit by commit.

## Proposed changes

* perf: increase a lot perf to download avatars (ex on a blame display: from 1m30s to 6s)
* perf: fix serialization that was broken for some avatars
* Always obtain the same Author initials image between 2 GitExtensions session
* return more meaningful author initials
* Better set of colors for Author initials background colors
* perf: make that cache never expire for avatar that never changes
* Perf: slight optimization of GravatarProvider
* Perf: prevent requesting the same avatar multiple times
* saner performance default values

## Screenshots 

For choice of colors:

### Before
| Before  |  After |
|---|---|
| image  | image |

## Test methodology 

- Manual
- Some unit tests

## Test environment(s) 

- Git Extensions 33.33.33
- Build de0dcb3112d960166620a45a5bf6e49f6918ea3b (Dirty)
- Git 2.38.1.windows.1
- .NET 6.0.14
- DPI 96dpi (no scaling)

## Merge strategy

Merge commit

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -576,7 +576,7 @@ namespace GitCommands
 
         public static int AvatarImageCacheDays
         {
-            get => GetInt("authorimagecachedays", 5);
+            get => GetInt("authorimagecachedays", 13);
             set => SetInt("authorimagecachedays", value);
         }
 
@@ -588,8 +588,14 @@ namespace GitCommands
 
         public static AvatarProvider AvatarProvider
         {
-            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.Default);
+            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.None);
             set => SetString("Appearance.AvatarProvider", value.ToString());
+        }
+
+        public static int AvatarCacheSize
+        {
+            get => GetInt("Appearance.AvatarCacheSize", 50);
+            set => SetInt("Appearance.AvatarCacheSize", value);
         }
 
         /// <summary>
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -2066,60 +2066,31 @@ namespace GitCommands
             SettingsContainer.SettingsCache.Import(GetSettingsFromRegistry());
         }
 
-        public static bool? GetBool(string name)
-        {
-            return SettingsContainer.GetBool(name);
-        }
+        #region Save in settings file
 
-        public static bool GetBool(string name, bool defaultValue)
-        {
-            return SettingsContainer.GetBool(name, defaultValue);
-        }
+        // String
+        [return: NotNullIfNotNull("defaultValue")]
+        public static string? GetString(string name, string? defaultValue) => SettingsContainer.GetString(name, defaultValue);
+        public static void SetString(string name, string value) => SettingsContainer.SetString(name, value);
 
-        public static void SetBool(string name, bool? value)
-        {
-            SettingsContainer.SetBool(name, value);
-        }
+        // Bool
+        public static bool? GetBool(string name) => SettingsContainer.GetBool(name);
+        public static bool GetBool(string name, bool defaultValue) => SettingsContainer.GetBool(name, defaultValue);
+        public static void SetBool(string name, bool? value) => SettingsContainer.SetBool(name, value);
 
-        public static void SetInt(string name, int? value)
-        {
-            SettingsContainer.SetInt(name, value);
-        }
+        // Int
+        public static int? GetInt(string name) => SettingsContainer.GetInt(name);
+        public static int GetInt(string name, int defaultValue) => SettingsContainer.GetInt(name, defaultValue);
+        public static void SetInt(string name, int? value) => SettingsContainer.SetInt(name, value);
 
-        public static int? GetInt(string name)
-        {
-            return SettingsContainer.GetInt(name);
-        }
+        // Date
+        public static DateTime? GetDate(string name) => SettingsContainer.GetDate(name);
+        public static DateTime GetDate(string name, DateTime defaultValue) => SettingsContainer.GetDate(name, defaultValue);
+        public static void SetDate(string name, DateTime? value) => SettingsContainer.SetDate(name, value);
 
-        public static DateTime GetDate(string name, DateTime defaultValue)
-        {
-            return SettingsContainer.GetDate(name, defaultValue);
-        }
-
-        public static void SetDate(string name, DateTime? value)
-        {
-            SettingsContainer.SetDate(name, value);
-        }
-
-        public static DateTime? GetDate(string name)
-        {
-            return SettingsContainer.GetDate(name);
-        }
-
-        public static int GetInt(string name, int defaultValue)
-        {
-            return SettingsContainer.GetInt(name, defaultValue);
-        }
-
-        public static void SetFont(string name, Font value)
-        {
-            SettingsContainer.SetFont(name, value);
-        }
-
-        public static Font GetFont(string name, Font defaultValue)
-        {
-            return SettingsContainer.GetFont(name, defaultValue);
-        }
+        // Font
+        public static Font GetFont(string name, Font defaultValue) => SettingsContainer.GetFont(name, defaultValue);
+        public static void SetFont(string name, Font value) => SettingsContainer.SetFont(name, value);
 
         [Obsolete("AppSettings is no longer responsible for colors, ThemeModule is")]
         public static Color GetColor(AppColor name)
@@ -2127,36 +2098,13 @@ namespace GitCommands
             return SettingsContainer.GetColor(name.ToString().ToLowerInvariant() + "color", AppColorDefaults.GetBy(name));
         }
 
-        public static void SetEnum<T>(string name, T value) where T : Enum
-        {
-            SettingsContainer.SetEnum(name, value);
-        }
+        // Enum
+        public static T GetEnum<T>(string name, T defaultValue) where T : struct, Enum => SettingsContainer.GetEnum(name, defaultValue);
+        public static void SetEnum<T>(string name, T value) where T : Enum => SettingsContainer.SetEnum(name, value);
 
-        public static T GetEnum<T>(string name, T defaultValue) where T : struct, Enum
-        {
-            return SettingsContainer.GetEnum(name, defaultValue);
-        }
-
-        public static void SetNullableEnum<T>(string name, T? value) where T : struct, Enum
-        {
-            SettingsContainer.SetNullableEnum(name, value);
-        }
-
-        public static T? GetNullableEnum<T>(string name) where T : struct
-        {
-            return SettingsContainer.GetNullableEnum<T>(name);
-        }
-
-        public static void SetString(string name, string value)
-        {
-            SettingsContainer.SetString(name, value);
-        }
-
-        [return: NotNullIfNotNull("defaultValue")]
-        public static string? GetString(string name, string? defaultValue)
-        {
-            return SettingsContainer.GetString(name, defaultValue);
-        }
+        public static T? GetNullableEnum<T>(string name) where T : struct => SettingsContainer.GetNullableEnum<T>(name);
+        public static void SetNullableEnum<T>(string name, T? value) where T : struct, Enum => SettingsContainer.SetNullableEnum(name, value);
+        #endregion
 
         private static void LoadEncodings()
         {
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -2083,6 +2083,9 @@ namespace GitCommands
         public static int GetInt(string name, int defaultValue) => SettingsContainer.GetInt(name, defaultValue);
         public static void SetInt(string name, int? value) => SettingsContainer.SetInt(name, value);
 
+        // Float
+        public static float GetFloat(string name, float defaultValue) => SettingsContainer.GetFloat(name, defaultValue);
+
         // Date
         public static DateTime? GetDate(string name) => SettingsContainer.GetDate(name);
         public static DateTime GetDate(string name, DateTime defaultValue) => SettingsContainer.GetDate(name, defaultValue);
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -598,6 +598,7 @@ namespace GitCommands
             set => SetInt("Appearance.AvatarCacheSize", value);
         }
 
+        // Currently not configurable in UI
         // Names from here: https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.brushes?view=windowsdesktop-7.0
         // or #AARRGGBB code
         public static string AvatarAuthorInitialsPalette
@@ -605,6 +606,9 @@ namespace GitCommands
             get => GetString("Appearance.AvatarAuthorInitialsPalette", "SlateGray,RoyalBlue,Purple,OrangeRed,Teal,OliveDrab");
             set => SetString("Appearance.AvatarAuthorInitialsPalette", value);
         }
+
+        // Currently not configurable in UI
+        public static float AvatarAuthorInitialsLuminanceThreshold => GetFloat("AvatarAuthorInitialsLuminanceThreshold", 0.5f);
 
         /// <summary>
         /// Loads a setting with GetString and parses it to an enum
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -598,6 +598,14 @@ namespace GitCommands
             set => SetInt("Appearance.AvatarCacheSize", value);
         }
 
+        // Names from here: https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.brushes?view=windowsdesktop-7.0
+        // or #AARRGGBB code
+        public static string AvatarAuthorInitialsPalette
+        {
+            get => GetString("Appearance.AvatarAuthorInitialsPalette", "SlateGray,RoyalBlue,Purple,OrangeRed,Teal,OliveDrab");
+            set => SetString("Appearance.AvatarAuthorInitialsPalette", value);
+        }
+
         /// <summary>
         /// Loads a setting with GetString and parses it to an enum
         /// </summary>
--- a/GitUI/Avatars/AvatarMemoryCache.cs
+++ b/GitUI/Avatars/AvatarMemoryCache.cs
@@ -12,6 +12,7 @@ namespace GitUI.Avatars
     public sealed class AvatarMemoryCache : IAvatarProvider, IAvatarCacheCleaner
     {
         private readonly MruCache<(string email, int imageSize), Image> _cache;
+        private HashSet<(string email, int imageSize)> _requested = new(6);
         private readonly IAvatarProvider _inner;
 
         public AvatarMemoryCache(IAvatarProvider inner, int capacity = 30)
@@ -26,25 +27,74 @@ namespace GitUI.Avatars
         /// <inheritdoc />
         public async Task<Image?> GetAvatarAsync(string email, string? name, int imageSize)
         {
+            (string email, int imageSize) key = (email, imageSize);
             lock (_cache)
             {
-                if (_cache.TryGetValue((email, imageSize), out var cachedImage))
+                if (_cache.TryGetValue(key, out var cachedImage))
                 {
                     return cachedImage;
                 }
             }
 
-            var image = await _inner.GetAvatarAsync(email, name, imageSize);
-
-            if (image is not null)
+            // The revision grid could trigger same request multiple times at the same time
+            // => Wait that the first finish and result is added to the cache.
+            if (IsRequestInProgress(key))
             {
-                lock (_cache)
+                for (int i = 0; i < 10_000; i++)
                 {
-                    _cache.Add((email, imageSize), image);
+                    lock (_cache)
+                    {
+                        if (_cache.TryGetValue(key, out var cachedImage))
+                        {
+                            return cachedImage;
+                        }
+                    }
+
+                    if (!IsRequestInProgress(key))
+                    {
+                        // Early exit when the image is not in the cache and key is no more in the requests list
+                        // => the request has failed!
+                        break;
+                    }
+
+                    await Task.Delay(5); // Approximative time to read from disk to not slow down the grid scrolling
                 }
             }
 
-            return image;
+            try
+            {
+                lock (_requested)
+                {
+                    _requested.Add(key);
+                }
+
+                Image image = await _inner.GetAvatarAsync(email, name, imageSize);
+
+                if (image is not null)
+                {
+                    lock (_cache)
+                    {
+                        _cache.Add(key, image);
+                    }
+                }
+
+                return image;
+            }
+            finally
+            {
+                lock (_requested)
+                {
+                    _requested.Remove(key);
+                }
+            }
+
+            bool IsRequestInProgress((string email, int imageSize) avatarKey)
+            {
+                lock (_requested)
+                {
+                    return _requested.Contains(avatarKey);
+                }
+            }
         }
 
         /// <inheritdoc />
@@ -53,6 +103,11 @@ namespace GitUI.Avatars
             lock (_cache)
             {
                 _cache.Clear();
+            }
+
+            lock (_requested)
+            {
+                _requested.Clear();
             }
 
             CacheCleared?.Invoke(this, EventArgs.Empty);
--- a/GitUI/Avatars/FileSystemAvatarCache.cs
+++ b/GitUI/Avatars/FileSystemAvatarCache.cs
@@ -26,10 +26,10 @@ namespace GitUI.Avatars
         /// <inheritdoc />
         public async Task<Image?> GetAvatarAsync(string email, string? name, int imageSize)
         {
-            var cacheDir = AppSettings.AvatarImageCachePath;
-            var path = Path.Combine(cacheDir, $"{email}.{imageSize}px.png");
+            string cacheDir = AppSettings.AvatarImageCachePath;
+            string path = Path.Combine(cacheDir, $"{email}.{imageSize}px.png");
 
-            var image = ReadImage() ?? await _inner.GetAvatarAsync(email, name, imageSize);
+            Image image = ReadImage() ?? await _inner.GetAvatarAsync(email, name, imageSize);
 
             if (image is not null)
             {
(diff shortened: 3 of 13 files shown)


GitUI/Avatars/GravatarProvider.cs
```diff
@@ -14,6 +14,14 @@ public sealed class GravatarProvider : IAvatarProvider
         private readonly AvatarFallbackType? _fallback;
         private readonly bool _forceFallback;
 
+        #pragma warning disable SYSLIB0021 // 'MD5CryptoServiceProvider' is obsolete
+
+        // Hash specified at http://en.gravatar.com/site/implement/hash/
+        private readonly MD5CryptoServiceProvider _md5 = new();
```
mstv, line 20: The warning message gives a hint about easy-to-use replacement.

gerhardol (commented):
+1
Would prefer to run this some too

GitUI/Avatars/AvatarMemoryCache.cs
```diff
@@ -12,6 +14,7 @@ namespace GitUI.Avatars
     public sealed class AvatarMemoryCache : IAvatarProvider, IAvatarCacheCleaner
     {
         private readonly MruCache<(string email, int imageSize), Image> _cache;
+        private List<(string email, int imageSize)> _requested = new(6);
```
mstv, line 17:
```suggestion
        private HashSet<(string email, int imageSize)> _requested = new(6);
```

mstv (commented):
Looks sensible, have not run
`var` should be replaced with explicit types

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -65,22 +79,42 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
-                    return names[0][0].ToString().ToUpper();
+                    return name[0].ToString().ToUpper();
+                }
+
+                if (char.IsUpper(name[1]))
+                {
+                    return $"{char.ToUpper(name[0])}{name[1]}";
+                }
+
+                var splitNames = name.Split(_emailInitialSeparator);
```
RussKie, line 99: [nit] avoid `var` unless the type is obvious

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -18,12 +18,26 @@ public class InitialsAvatarProvider : IAvatarProvider
...
             var (initials, hashCode) = GetInitialsAndHashCode(email, name);
 
-            var avatarColor = _avatarColors[Math.Abs(hashCode) % _avatarColors.Length];
+            var avatarColor = _avatarColors[hashCode];
             var avatar = DrawText(initials, avatarColor, imageSize);
 
             return Task.FromResult<Image?>(avatar);
         }
 
+        private int GetDeterministicHashCode(string str)
```
RussKie, line 27: It'd be good to have some comments explaining what's going on

GitUI/Avatars/GravatarProvider.cs
pmiossec: Thanks for the head up. I just moved the code without a careful look and was thinking that it was the MD5 hashing algorithm that was obsoleted. In fact, it was just the API. I will push the fix...

GitCommands/Settings/AppSettings.cs
```diff
@@ -588,10 +588,16 @@ public static bool ShowAuthorAvatarInCommitInfo
 
         public static AvatarProvider AvatarProvider
         {
-            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.Default);
+            get => GetEnumViaString("Appearance.AvatarProvider", AvatarProvider.None);
```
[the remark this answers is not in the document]
pmiossec: In fact, the Initials is the default value of `AvatarFallbackType` settings. So yes, initials will end up to be the default avatar image on a fresh install. It has the advantage to be really fast, 100% local, much more recognizable than other fallback and do not expire...

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -18,12 +18,26 @@ public class InitialsAvatarProvider : IAvatarProvider
         {
             var (initials, hashCode) = GetInitialsAndHashCode(email, name);
 
-            var avatarColor = _avatarColors[Math.Abs(hashCode) % _avatarColors.Length];
+            var avatarColor = _avatarColors[hashCode];
```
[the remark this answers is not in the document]
pmiossec: I made this change because it was easier to debug when you have to see if things are working well. But I can revert it...

GitUI/Avatars/FileSystemAvatarCache.cs
```diff
@@ -52,7 +52,19 @@ void WriteImage()
                 }
                 catch
                 {
-                    // do nothing
+                    // Workaround to avoid the "A generic error occurred in GDI+." exception when saving
```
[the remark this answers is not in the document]
pmiossec: I don't really explain the workaround (even if I can guarantee it is working 100% well!).
It's really difficult to have the inner problem that cause the exception. I got it once by luck but I don't remember exactly. If I remember well the resource is locked but copying it like that is working. Don't ask me why...

Even if the answers are not exactly describing our case, I saw multiple times in stackoverflow this workaround, like in this answer ( https://stackoverflow.com/a/59069870 ) and that's right that it works...

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -123,10 +123,11 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
         {
             Color.RoyalBlue,
-            Color.DarkRed,
+            Color.SlateGray,
             Color.Purple,
-            Color.ForestGreen,
-            Color.DarkOrange
+            Color.OrangeRed,
+            Color.Teal,
+            Color.OliveDrab,
```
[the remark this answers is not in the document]
pmiossec: I had a lot of difficulties to find colors that are dark enough (to be able to see initials in white) and different enough to be easily distinguished without ending up with something horrible.

But if you have a lot of time to spend on that, I will be happy to change it.
Maybe it's easy to add a hidden setting for that. I will have a look...

GitCommands/Settings/AppSettings.cs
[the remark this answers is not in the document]
vbjay: Code in c# a way to get composite color so no matter the background color a text color will show. 

To get a composite color that will contrast with any given background color, you can use the following C# code:

scssCopy code

Color GetContrastColor(Color backgroundColor) { // Calculate the luminance of the background color double luminance = (0.2126 * backgroundColor.R + 0.7152 * backgroundColor.G + 0.0722 * backgroundColor.B) / 255; // Use a threshold value to determine whether to use black or white as the foreground color Color foregroundColor = (luminance > 0.5) ? Color.Black : Color.White; return foregroundColor; } 

This method takes a Color object representing the background color and returns a Color object representing the foreground color that will provide the best contrast with the background.

The method works by calculating the luminance of the background color using the standard formula for converting RGB values to grayscale. If the luminance is greater than 0.5, it means the background color is relatively light, so the method returns black as the foreground color. If the luminance is less than or equal to 0.5, the background color is relatively dark, so the method returns white as the foreground color.

You can use this method in your code by calling GetContrastColor() with the background color you want to contrast with. For example:

scssCopy code

Color backgroundColor = Color.FromArgb(255, 128, 0); // orange Color foregroundColor = GetContrastColor(backgroundColor); Console.WriteLine($"Foreground color: {foregroundColor}"); 

This code will output "Foreground color: Color [A=255, R=255, G=255, B=255]" because white is the best foreground color to use with an orange background

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -79,22 +79,42 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
+                if (name.Length == 1)
                 {
                     // ... return that character as uppercase
-                    return names[0][0].ToString().ToUpper();
+                    return name[0].ToString().ToUpper();
+                }
+
+                if (char.IsUpper(name[1]))
+                {
+                    return $"{char.ToUpper(name[0])}{name[1]}";
```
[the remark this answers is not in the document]
pmiossec: done.

GitUI/Avatars/GravatarProvider.cs
pmiossec: done.

GitUI/Avatars/AvatarMemoryCache.cs
pmiossec: done.

GitUI/Avatars/InitialsAvatarProvider.cs
pmiossec: done.

GitUI/Avatars/InitialsAvatarProvider.cs
pmiossec: I wrote something....

GitUI/Avatars/InitialsAvatarProvider.cs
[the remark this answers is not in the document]
gerhardol: I am not good at selecting colors, better at complaining...
The colors are not that different, really just the SlateGray that was added and that happen to be on quite a lot of the example...
Maybe removing SlateGray is an option, keep it for now.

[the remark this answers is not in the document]
gerhardol: > Code in c# a way to get composite color so no matter the background color a text color will show.
>
> To get a composite color that will contrast with any given background color, you can use the following C# code:
>
> scssCopy code
>
> Color GetContrastColor(Color backgroundColor) { // Calculate the luminance of the background color double luminance = (0.2126 * backgroundColor.R + 0.7152 * backgroundColor.G + 0.0722 * backgroundColor.B) / 255; // Use a threshold value to determine whether to use black or white as the foreground color Color foregroundColor = (luminance > 0.5) ? Color.Black : Color.White; return foregroundColor; }
>
> This method takes a Color object representing the background color and returns a Color object representing the foreground color that will provide the best contrast with the background.
>
>The method works by calculating the luminance of the background color using the standard formula for converting RGB values to grayscale. If the luminance is greater than 0.5, it means the background color is relatively light, so the method returns black as the foreground color. If the luminance is less than or equal to 0.5, the background color is relatively dark, so the method returns white as the foreground color.
>
>You can use this method in your code by calling GetContrastColor() with the background color you want to contrast with. For example:
>
> scssCopy code
> 
> Color backgroundColor = Color.FromArgb(255, 128, 0); // orange Color foregroundColor = GetContrastColor(backgroundColor); Console.WriteLine($"Foreground color: {foregroundColor}");
>
>This code will output "Foreground color: Color [A=255, R=255, G=255, B=255]" because white is the best foreground color to use with an orange background

Colors are complex and require tuning. I would have expected that there were more guidelines and libs supporting this.
Still amazed by the work done by @NikolayXHD for the dark theme, so sad to loose it.
This would be even harder with another theme.

GitCommands/Settings/AppSettings.cs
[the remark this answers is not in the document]
gerhardol: git-range-diff was very helpful in this review....

@vbjay Thanks for the response, quoted in the other comment.

gerhardol (approved):

vbjay:
No prob. Why I sent it was exactly for the idea of when theming works. The
dark theme can have a gradient of light color backgrounds to show a good
contrast on top of a dark background. Along with text color being good on
said background. Then flip the idea on light backgrounds.

I used chatgpt and thought it would be cool to just copy the original
convo.

GitUI/Avatars/InitialsAvatarProvider.cs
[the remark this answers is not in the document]
pmiossec: i have pushed a commit so that there is a (hidden) setting for the colors list. Like that every one could remove the Grey 😉

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -65,35 +87,76 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
+                    return Color.Black;
+                }
+            }
+
+            try
+            {
+                var alpha = Convert.ToInt32(c.Substring(1, 2), 16);
+                var red = Convert.ToInt32(c.Substring(3, 2), 16);
+                var green = Convert.ToInt32(c.Substring(5, 2), 16);
+                var blue = Convert.ToInt32(c.Substring(7, 2), 16);
```
mstv, line 152: use `int`?

mstv (approved):
LGTM, have not run

GitUI/Avatars/InitialsAvatarProvider.cs
vbjay: Or use the class already made for this purpose. 

https://learn.microsoft.com/en-us/dotnet/api/system.drawing.colorconverter

vbjay: Or just pass the full number instead of substring 

https://learn.microsoft.com/en-us/dotnet/api/system.drawing.color.fromargb

vbjay: https://www.c-sharpcorner.com/uploadfile/mahesh/the-colorconverter-and-colortranslater-classes-in-gdi/

Basically use the right one based on what you want to store and retrieve.

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -16,27 +16,49 @@ public class InitialsAvatarProvider : IAvatarProvider
...
         }
 
+        /// <summary>
+        /// Calculate the most simpler non-cryptographic deterministic hash (to get the same result every times it is calculated for the same string)
+        /// We just need a not too costly way to convert a string in an integer and calculating a hash is a good way to do it.
+        /// We are avoiding to use `.GetHashCode()` because it returns a different value for each process :(
+        /// Taken from https://stackoverflow.com/a/5155015
+        /// </summary>
+        /// <param name="str">the string to calculate the hash for</param>
+        /// <returns>the hash calculated</returns>
```
RussKie, line 34:
```suggestion
        /// <summary>
        /// Calculate the most simpler non-cryptographic deterministic hash (to get the same result every times it is calculated for the same string)
        /// We just need an inexpensive way to convert a string to an integer, and calculating a hash is a good way to do it.
        /// We are not using <c>GetHashCode()</c> because it returns a different value for each process.
        /// Borrowed from https://stackoverflow.com/a/5155015
        /// </summary>
        /// <param name="str">The string to calculate a hash for.</param>
        /// <returns>The calculated hash.</returns>
```

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -65,35 +87,76 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
             // If only a single valid name-element is found ...
             if (names.Length == 1)
             {
                 // ... and that name-element is only a single character long ...
-                if (names[0].Length == 1)
+                if (name.Length == 1)
                 {
                     // ... return that character as uppercase
-                    return names[0][0].ToString().ToUpper();
+                    return name[0].ToString().ToUpper();
```
RussKie, line 99: Why not this?
```suggestion
                    return char.ToUpper(name[0]);
```

GitUI/Avatars/InitialsAvatarProvider.cs
RussKie: @vbjay do you mean [`ColorTranslator.FromHtml(String)`](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.colortranslator.fromhtml?view=net-7.0#system-drawing-colortranslator-fromhtml(system-string))?

RussKie (commented):
Overall looks good. Few nits. Modulo the question from/suggestion from @vbjay.

GitUI/Avatars/InitialsAvatarProvider.cs
pmiossec: done.

vbjay: @RussKie  yes. or the color converter way... Either way we shouldn't be rolling our own way when .net has ways to translate colors.

vbjay: Once you have the color you can retrieve the parts.
```c#
Color c = Colors.Cyan;

Alpha= c.A;
Red=c.R;
```

GitCommands/Settings/AppSettings.cs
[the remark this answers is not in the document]
vbjay: No prob. Why I sent it was exactly for the idea of when theming works. The
dark theme can have a gradient of light color backgrounds to show a good
contrast on top of a dark background. Along with text color being good on
said background. Then flip the idea on light backgrounds.

I used chatgpt and thought it would be cool to just copy the original
convo.

Github email handler could be better  If replying to a review thread email, should put reply in that thread.

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -65,35 +87,72 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
+            }
+
+            try
+            {
+                return ColorTranslator.FromHtml(c);
+            }
+            catch
+            {
+                return Color.Black;
+            }
```
vbjay, line 154: Don't make it so complicated...

image
image
image


```C#
Color named = Color.Cyan;
Color other = Color.FromArgb(named.ToArgb() + 1);

var translations = new { Named = named, NamedTranslated = ColorTranslator.ToHtml(named), Other = other, OtherTranslated = ColorTranslator.ToHtml(other) };

translations.Dump("TranslatedTo");

var BackToColors = new { Named = ColorTranslator.FromHtml(translations.NamedTranslated), Other = ColorTranslator.FromHtml(translations.OtherTranslated) };
BackToColors.Dump("RoundTrip");

```

vbjay:
```suggestion
             try
            {
                return ColorTranslator.FromHtml(c);
            }
            catch
            {
                return Color.Black;
            }
```

vbjay (changes requested):

GitUI/Avatars/InitialsAvatarProvider.cs
vbjay:
```C#
Dictionary<string, Color> colorDict = new Dictionary<string, Color>();
foreach (var prop in typeof(Color).GetProperties())
{
	if (prop.PropertyType == typeof(Color))
	{
		Color color = (Color)prop.GetValue(null);
		colorDict.Add(prop.Name, color);
	}
}

colorDict.Keys.Select(k => System.Drawing.ColorTranslator.FromHtml(k)).Dump();
```

Did not cause an exception  while 

image

pmiossec: I pushed changes hours ago. It seems that you are still basing your comments on the old version. Have a look...

vbjay: Not needed

Stop using fromname

image
As I have proven in my comments all you need is the colortranslator.

vbjay: https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/ColorConverter.cs,109

https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Advanced/ColorTranslator.cs,228

See how ColorTranslator handles all the names and even the LightGray name difference?

vbjay: .net versions of links 

https://source.dot.net/#System.Drawing.Primitives/System/Drawing/Color.cs,475

Colors used by the abov emethod https://source.dot.net/System.Drawing.Primitives/src/libraries/Common/src/System/Drawing/ColorTable.cs.html#6532ea3b4c32baf5

While Translator is the following

https://source.dot.net/#System.Drawing.Primitives/System/Drawing/ColorTranslator.cs,218

GitUI/Avatars/InitialsAvatarProvider.cs
```diff
@@ -65,35 +87,60 @@ private static (string? name, char[]? separator) NameSelector(string? name, stri
...
         {
-            Color.RoyalBlue,
-            Color.DarkRed,
-            Color.Purple,
-            Color.ForestGreen,
-            Color.DarkOrange
-        };
+            try
+            {
+                return !colorCode.StartsWith("#") ? Color.FromName(colorCode) : ColorTranslator.FromHtml(colorCode);
```
vbjay, line 137:
```suggestion
                return ColorTranslator.FromHtml(colorCode);
```

vbjay: See https://github.com/gitextensions/gitextensions/pull/10846/files#r1155297871 comment chain

GitUI/Avatars/InitialsAvatarProvider.cs
pmiossec: Indeed. Done.

vbjay (approved):

pmiossec:
I think I fixed all the review comments...

vbjay:
Just an improvement on my contrasting colors comment

https://gist.github.com/vbjay/cc56f5e67a9165bd022418167b6af323

image

Open in linqpad and run.  Play with  thresholds and such to tweak.

vbjay:
I'd change the getconteastingcolor function use the dynamic thresholds idea. That allows us to tweak the low or high value if needed.  

This is an extension method and we could make it part of GitGui extensions. That way if something else needs contrasting color it is reusable and consistent. 


https://gist.github.com/vbjay/cc56f5e67a9165bd022418167b6af323

mstv:
(GH supports delta reviews: blue button "Changes since last review". I find this very useful. But GH loses track after too many rebases / force pushes. For the future, please rebase only in order to resolve conflicts or perhaps after a longer inactivity. Thank you.)

pmiossec:
> (GH supports delta reviews: blue button "Changes since last review". I find this very useful. But GH loses track after too many rebases / force pushes. For the future, please rebase only in order to resolve conflicts or perhaps after a longer inactivity. Thank you.)

I didn't know this option that looks great. I will keep for now on the fixup/rebase as much as possible for a last pre-merge step.

pmiossec:
> I'd change the getcontrastingcolor function use the dynamic thresholds idea. That allows us to tweak the low or high value if needed.
> 
> This is an extension method and we could make it part of GitGui extensions. That way if something else needs contrasting color it is reusable and consistent.
> 
> https://gist.github.com/vbjay/cc56f5e67a9165bd022418167b6af323
> 
> Could even adjust extension method to have an overload that accepts the thresholds if we wanted to override for some reason.

@vbjay Reading multiple times the code, I absolutely don't understand threshold part. Either it is very clever and complicated and I don't get it or the implementation is bugged. I tend for the 2nd option without being certain: why changing the threshold value based on the luminance and after comparing them whereas a simple comparison seems enough? 

So I have made something simpler that I think is enough...

Have a look, please.

vbjay:
The idea is if needed you can set <.5 threshold to .2 and >.5 to .8 and it adjusts the points where color flips to white. 


Open script in gist with linkpad.net and play with adjusting threshold values and running.

pmiossec:
> The idea is if needed you can set <.5 threshold to .2 and >.5 to .8 and it adjusts the points where color flips to white.

That's what my code is doing also, I think.
And playing with linqpad does not help understand the code....
I spent int total more than 20min in the 2 loc below without understanding the added value to change the threshold based on the luminance value **AND AFTER** comparing these values instead of changing the threshold value directly which is way more clearer.

Because, for me it's a lot of complexity for something wrong. Let's see with the 0.2 and 0.8 values you are mentioning, so the code:

```
	// Use a dynamic threshold based on the luminance of the background color.  Play with threshold values when <.5 or not to tweak the flip
	double threshold = (luminance < 0.5) ? 0.2 : 0.8;

	// Use the threshold value to determine whether to use black or white as the foreground color
	Color foregroundColor = (luminance > threshold) ? Color.Black : Color.White;
```

The results:

| | luminance < 0.2 |  0.2 < luminance < 0.5 | 0.5 < luminance < 0.8  | luminance > 0.8  |
|-|-----------------|---|---|---|
|    Background |   Very dark | dark  | light | Very light  |
|Threshold (updated)  | 0.2  |  0.2  |  0.8  | 0.8  |
| Foreground |     `false` => White       | `true` => Black  | `false` => White  | `true` => Black  |
| What we want? |    White on Very dark:👍           | Black on dark:👎  |  White on light:👎 | Black on Very light:👍  |

For me the only good behavior is when the 2 values are the same which is equivalent to my (a lot simpler to understand) code...

Or I misunderstood everything and that's not all the values are expected to be set? In this case, can you explain in details?

gerhardol:
Ready to merge?

RussKie:
@vbjay has approved, so I think it is.

vbjay:
Correct.  I was just pushing for a little more control over the threshold and fine tuning if contrast didn't quite work.

pmiossec:
> Correct. I was just pushing for a little more control over the threshold and fine tuning if contrast didn't quite work.

Yes but except if I missed something, your proposition is not working or not adding value, with much more complexity, as I shown in my list comment. I was waiting for your answer on that.

If you are all OK, I could rebase and squash one or 2 commits to be ready to go...

vbjay:
Last attempt. Think of it instead of a single slider you have 2. If low you can independently adjust where the threshold if less than .5 or darker colors. This .5 threshold isn't perfect and allowing 2 thresholds lets us independently adjust the thresholds if light or dark.

vbjay:
The other point. I wanted you to keep it sn extension method. So if and when a color needs to get contrast it is c.GetContrastColor

pmiossec:
> Last attempt. Think of it instead of a single slider you have 2. If low you can independently adjust where the threshold if less than .5 or darker colors. This .5 threshold isn't perfect and allowing 2 thresholds lets us independently adjust the thresholds if light or dark.

I understand the general idea but I would have preferred an explanation with a real example with working threshold values because except if you found an error in the one I gave, we see that the solution with 2 threshold values is not working and add a lot of complexity.

Also, having only one setting is nearly self-explanatory if someone wants to play with it. Good luck for the explanation and the documentation if we include the 2 threshold settings... I spend quite some time investigating it and I still can't get the added value...(for me it's not working!) 

I will be interested by another point of view on the subject. 

> The other point. I wanted you to keep it an extension method. So if and when a color needs to get contrast it is c.GetContrastColor

I will have a look at that...

pmiossec:
> The other point. I wanted you to keep it an extension method. So if and when a color needs to get contrast it is c.GetContrastColor

Done.

I have updated the PR to merge into the new `release/4.1` branch also.
If we decide to not include in the 4.1, I will change it back....

pmiossec:
> @pmiossec merge at will

Thanks. Done.

PR #10847, opened 2023-04-01
Restore from minimized: Configurable avoid async Git

gerhardol:
## Proposed changes

Add an experimental setting to not run git-ls-files when the window is minimized on activation, as async commands when minimized seem to block the app from being restored.

Rename configuration added in #10119 to
deactivate at upgrade as this setting may give a negative experience after #10802 mostly fixes the problem.

--

The "restore from minimize" issue is occurring often for some users in 4.0
There has been occasional reports of similar issues in 2.x and 3.x too, so it is not exactly new though.
#10802 mostly fixes the problem but not completely.
It may be resolved by not running Git commands async while being minimized. I thought I had tested that, but am not sure anymore.
https://github.com/gitextensions/gitextensions/pull/10802#issuecomment-1481845774
further usage points that async Git may be the issue.

This fix adds a setting to not run async Gic, so the following banner may not appear after a restore (but after switching back to normal/maximized state):
image

I hope that we do not need to keep this fix in the production version, that a better fix is found.
The UI refresh check could maybe be done later too (like a check if needed and run after the grid is refreshed).
If not found or improved, this is a kind of workaround.

Anyway, it simplifies narrowing the problem for users.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -2005,9 +2005,14 @@ namespace GitCommands
         }
 
         // Set manually in settings file
-        public static bool WorkaroundRestoreFromMinimize
+        public static bool WorkaroundActivateFromMinimize
         {
-            get => GetBool("WorkaroundRestoreFromMinimize", false);
+            get => GetBool("WorkaroundActivateFromMinimize", false);
+        }
+
+        public static bool GitAsyncWhenMinimized
+        {
+            get => GetBool("GitAsyncWhenMinimized", true);
         }
 
         private static IEnumerable<(string name, string value)> GetSettingsFromRegistry()
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1059,7 +1059,8 @@ namespace GitUI.CommandsDialogs
             notificationBarBisectInProgress.RefreshBisect();
 
             // check if we are in the middle of an action (merge/rebase/etc.)
-            notificationBarGitActionInProgress.RefreshGitAction();
+            notificationBarGitActionInProgress.RefreshGitAction(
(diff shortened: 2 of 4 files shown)


gerhardol:
@vjdw @mlaily Can you check if setting GitAsyncWhenMinimized resolves the restore issue in vNext?

vjdw:
> @vjdw @mlaily Can you check if setting GitAsyncWhenMinimized resolves the restore issue in vNext?

I'm not sure where to get vNext nor where to set GitAsyncWhenMinimized. But I've just updated the master branch to commit `5e79d32ef5d365fb1a23fd59b458ee2f6b82e354` and initial impressions are that the restore issue is fixed.

gerhardol:
> > @vjdw @mlaily Can you check if setting GitAsyncWhenMinimized resolves the restore issue in vNext?
> 
> I'm not sure where to get vNext nor where to set GitAsyncWhenMinimized. But I've just updated the master branch to commit `5e79d32ef5d365fb1a23fd59b458ee2f6b82e354` and initial impressions are that the restore issue is fixed.

If you do not see the issue in master, using the build in this PR (or master if it is merged) will not give anything. A good thing.
You stated that you had the problem occasionally though.

master is vNext, very likely 4.1 that is being considered.

To activate the setting, add the following to C:\Users\<user>\AppData\Roaming\GitExtensions\GitExtensions\GitExtensions.settings

```
  <item>
    <key>
      <string>GitAsyncWhenMinimized</string>
    </key>
    <value>
      <string>false</string>
    </value>
  </item>
```

vjdw:
Yeah, I can't reproduce the issue on master at the moment. FWIW I also tested your branch with and without the new flag and there is no restore issue either way.

mstv (approved):
LGTM, have not run

PR #10849, opened 2023-04-01
IScriptHostControl.CurrentBranch() to decrease overhead

gerhardol:
## Proposed changes

Module.GetSelectedBranch() requires about 1 ms (15000 ticks) also if no Git command is required.
This info is already calculated in the revision grid and can be made available in the IScriptHostControl interface.
(An alternative is to add this to the event like for FilteredGitRefsProvider.)

Adding this in side panel, could be added elsewhere.
In for instance CommitInfo the reference to IScriptHostControl  need to be provided after creation, so a few more steps.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormFormatPatch.Designer.cs
+++ b/GitUI/CommandsDialogs/FormFormatPatch.Designer.cs
@@ -295,7 +295,7 @@ namespace GitUI.CommandsDialogs
             this.CurrentBranch.AutoSize = true;
             this.CurrentBranch.Location = new System.Drawing.Point(75, 0);
             this.CurrentBranch.Margin = new System.Windows.Forms.Padding(4, 0, 4, 0);
-            this.CurrentBranch.Name = "CurrentBranch";
+            this.CurrentBranch.Name = "GetCurrentBranch";
             this.CurrentBranch.Size = new System.Drawing.Size(0, 23);
             this.CurrentBranch.TabIndex = 5;
             // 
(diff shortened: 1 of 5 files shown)


vbjay:
If current branch is changed outside of current ge instance and you run, it will be wrong. Running the git command should be run to make sure of current branch.

gerhardol:
> If current branch is changed outside of current ge instance and you run, it will be wrong. Running the git command should be run to make sure of current branch.

That applies to any change in the grid: commits, current branch etc, so this is even more consistent. You need to refresh if doing external changes.

mstv (approved):

GitUI/Script/IScriptHostControl.cs
```diff
@@ -8,5 +8,6 @@ public interface IScriptHostControl
         GitRevision? GetLatestSelectedRevision();
         IReadOnlyList<GitRevision> GetSelectedRevisions();
         Point GetQuickItemSelectorLocation();
+        string CurrentBranch();
```
RussKie, line 11:
```suggestion
        string GetCurrentBranch();
```

RussKie (commented):
👍

GitUI/CommandsDialogs/FormFormatPatch.Designer.cs
```diff
@@ -295,7 +295,7 @@ private void InitializeComponent()
             this.CurrentBranch.AutoSize = true;
             this.CurrentBranch.Location = new System.Drawing.Point(75, 0);
             this.CurrentBranch.Margin = new System.Windows.Forms.Padding(4, 0, 4, 0);
-            this.CurrentBranch.Name = "CurrentBranch";
+            this.CurrentBranch.Name = "GetCurrentBranch";
```
mstv, line 298: Rather not

gerhardol: fixed...

mstv (approved):

PR #10851, opened 2023-04-02
Store view settings on request only

### LINKED ISSUE #10009: Toolbar button `Show Ref Log` does not work on first click
### Environment

- Git Extensions 33.33.33
- Build 56f797126690cd72178c80ee41ec90437147f6e2
- Git 2.35.1.windows.1
- .NET 6.0.5
- DPI 96dpi (no scaling)

### Issue description

The button state seems to be correctly initialized.
Though when clicked, it does not really hide the Ref Log on first click.

### Steps to reproduce

- activate `Show Ref Log`
- restart GE
- try to hide the Ref Log by clicking the toolbar button

most likely in 3.5.4

### Diagnostics

_No response_

mstv:
Fixes #10009

## Proposed changes

- Store the most important view settings on request only
  (RefLog, CurrentBranchOnly, BranchFilterEnabled, RevisionSortOrder, FileViewer toolbar flags)

## Screenshots 

Before|After
-|-
image|image

## Test methodology 

- manual
- new tests for `RuntimeSetting<T>`

## Please do not squash merge

----

### DIFF
--- a/GitCommands/Settings/BoolRuntimeSetting.cs
+++ b/GitCommands/Settings/BoolRuntimeSetting.cs
@@ -1,0 +1,24 @@
+﻿namespace GitCommands.Settings;
+
+/// <inheritdoc />
+public sealed class BoolRuntimeSetting : RuntimeSetting<bool>
+{
+    /// <summary>
+    ///  Initializes a new instance of the <see cref="BoolRuntimeSetting"/> class with the settings details.
+    /// </summary>
+    /// <param name="settingsSource">The source containing the value.</param>
+    /// <param name="name">The setting name.</param>
+    /// <param name="defaultValue">The default value.</param>
+    public BoolRuntimeSetting(SettingsPath settingsSource, string name, bool defaultValue)
+        : base(Setting.Create(settingsSource, name, defaultValue))
+    {
+    }
+
+    /// <summary>
+    ///  Toggles the setting value.
+    /// </summary>
+    public void Toggle()
+    {
+        Value = !Value;
+    }
+}
--- a/GitCommands/Settings/IRuntimeSetting.cs
+++ b/GitCommands/Settings/IRuntimeSetting.cs
@@ -1,0 +1,23 @@
+﻿namespace GitCommands.Settings;
+
+/// <summary>
+///  Interface for settings which are persisted on explicit request only.
+/// </summary>
+public interface IRuntimeSetting
+{
+    /// <summary>
+    ///  Reloads the value from the settings file.
+    /// </summary>
+    void Reload();
+
+    /// <summary>
+    ///  Resets the value to the hard-coded default value.
+    /// </summary>
+    void ResetToDefault();
+
+    /// <summary>
+    ///  Saves the value to the persistent setting,
+    ///  which is not written to the settings file at once but latest on app exit.
+    /// </summary>
+    void Save();
+}
--- a/GitCommands/Settings/RuntimeSetting.cs
+++ b/GitCommands/Settings/RuntimeSetting.cs
@@ -1,0 +1,94 @@
+﻿namespace GitCommands.Settings;
+
+/// <summary>
+///  Represents a setting which has to be explicitly saved, if changed at runtime.
+/// </summary>
+/// <typeparam name="T">The type of setting.</typeparam>
+public interface IRuntimeSetting<T> : ISetting<T>, IRuntimeSetting
+{
+    /// <summary>
+    ///  Optionally calls <cref>Reload</cref> and returns the current value.
+    /// </summary>
+    public T GetValue(bool reload = false);
+}
+
+/// <summary>
+///  Represents a setting which has to be explicitly saved, if changed at runtime.
+/// </summary>
+/// <typeparam name="T">The type of setting.</typeparam>
+public class RuntimeSetting<T> : IRuntimeSetting<T>
+{
+    private bool _loaded;
+    private readonly ISetting<T> _persistentSetting;
+    private T _value;
+
+    /// <summary>
+    ///  Initializes a new instance of the <see cref="RuntimeSetting{T}"/> class.
+    /// </summary>
+    /// <param name="persistentSetting">
+    ///  The <see cref="ISetting{T}"/> instance used to persist the default value of the setting.
+    /// </param>
+    public RuntimeSetting(ISetting<T> persistentSetting)
+    {
+        _persistentSetting = persistentSetting;
+    }
+
+    public T Default => _persistentSetting.Default;
+
+    public string FullPath => _persistentSetting.FullPath;
+
+    public bool IsUnset => _persistentSetting.IsUnset;
+
+    public string Name => _persistentSetting.Name;
+
+    public SettingsPath SettingsSource => _persistentSetting.SettingsSource;
+
+    public T Value
+    {
+        get => GetValue();
+        set
+        {
+            if (_value.Equals(value))
+            {
+                return;
+            }
+
+            _value = value;
+            if (_loaded)
+            {
+                Updated?.Invoke(this, EventArgs.Empty);
+            }
+        }
+    }
+
+    public event EventHandler? Updated;
+
+    public T GetValue(bool reload = false)
+    {
+        if (reload || !_loaded)
+        {
+            Reload();
+        }
+
+        return _value;
+    }
+
+    public void Reload()
+    {
+        Value = _persistentSetting.Value;
+        _loaded = true;
+    }
+
+    public void ResetToDefault() => Value = Default;
+
+    public void Save()
+    {
+        _persistentSetting.Value = Value;
+    }
+
+    /// <summary>
+    ///  Implicit conversion for direct access to the RuntimeSetting value.
+    /// </summary>
+    /// <param name="setting">The RuntimeSetting whose value is returned as conversion result.</param>
+    public static implicit operator T(RuntimeSetting<T> setting) => setting.Value;
+}
--- a/GitUI/CommandsDialogs/FormRebase.cs
+++ b/GitUI/CommandsDialogs/FormRebase.cs
@@ -398,9 +398,9 @@ namespace GitUI.CommandsDialogs
             finally
             {
                 AppSettings.ShowStashes = previousValueShowStashes;
-                AppSettings.BranchFilterEnabled = previousValueBranchFilterEnabled;
-                AppSettings.ShowCurrentBranchOnly = previousValueShowCurrentBranchOnly;
-                AppSettings.ShowReflogReferences = previousValueShowReflogReferences;
+                AppSettings.BranchFilterEnabled.Value = previousValueBranchFilterEnabled;
+                AppSettings.ShowCurrentBranchOnly.Value = previousValueShowCurrentBranchOnly;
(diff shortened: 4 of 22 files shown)


RussKie:
I'm not really fond of this approach, it kind of looks over complicated. 

> * Store the most important view settings on request only
>   (RefLog, FileViewer toolbar flags)

Are we not persisting the FileViewer toolbar flags already?
I was under impression that the reflog is the only settings we're persisting that, perhaps, we shouldn't be persisting - i.e., a setting that's a repo-specific and not a global setting.
Perhaps, a simpler - both from an implementation and user experience - solution could be to add a new tab, say, "Repository specific settings" and place related settings on it. The organisation of this tag could be done in the similar fashion to the "Confirmation" tag - i.e., select only those settings that are to be persisted globally - that is, if a settings isn't ticked it's state will affect only the current app instance and only the current repo until it is switched.

OT: the settings structure is a mess, and I think we should look at them holistically and rework the hierarchy.


> * Add an example of an experimental view setting
>   option to fan out all children lanes directly at parent


When I suggested experimental features I was thinking of adding a new tab under the Advanced tab:
image

mstv:
> Are we not persisting the FileViewer toolbar flags already?

We do, but in an arguable way. I have just turned these settings into View Settings.
`AppSettings.ShowReflogReferences` is only a facade to `_showReflogReferences.Value` in order to not need to adapt all usages. Should I remove this facade?

You have agreed to my definition of View Settings: https://github.com/gitextensions/gitextensions/issues/10009#issuecomment-1272614922
"I prefer a set of default view settings for the DiffViewer and for the RevisionGrid. (Temporary) changes should not be stored. Rather store them on explicit request."

> I was under impression that the reflog is the only settings we're persisting that, perhaps, we shouldn't be persisting - i.e., a setting that's a repo-specific and not a global setting.

Why should one have repo-specific preferences of View Settings? Supporting different setting levels - this would make it complex.
The point is, that I can change a View Setting in one instance not affecting the view in other instances (even if they show the same repo). 

> Perhaps, a simpler - both from an implementation and user experience - solution could be to add a new tab, say, "Repository specific settings" and place related settings on it. The organisation of this tag could be done in the similar fashion to the "Confirmation" tag - i.e., select only those settings that are to be persisted globally - that is, if a settings isn't ticked it's state will affect only the current app instance and only the current repo until it is switched.

Simply incomplete. We should not patronize users with a fixed set of view settings. They need to be able to define their own defaults.

Please elaborate what can be _simplified_ in the UX of my implementation.

Regarding implementation: If you know how to make C# generics distinguish `T` from `T?`, please let me know! I had a hard time trying to make it work for both `bool` and `enum`.

The settings dialog is not really usable for View Settings. It is too slow. I guess that is why we have a `View` menu. I do not deny to create a settings page, particularly for less frequently changed options.

> When I suggested experimental features I was thinking of adding a new tab under the Advanced tab:

I know but: ditto

> OT: the settings structure is a mess, and I think we should look at them holistically and rework the hierarchy.

`AppSettings` is flat. Hierarchy is created by the settings dialog only.

mstv:
copy of a comment in a - previously depended - PR in order to continue the discussion here (*later*):

> I feel this is taking is further away from the [Options pattern](https://learn.microsoft.com/dotnet/core/extensions/options). I vaguely remember ivangrek did (or was working towards) a POC that converted our settings to IOptons<T>...

The only "challenge" I see is that IOption<TOptions> expects TOptions being a class. Our options are of type bool, int, Enum (and string).
IViewSetting is just an additional interface with the specifics of a View Setting.

image

mstv:
If you mean `ISetting<T>`, `ViewSetting<T?>` would not be very different. But basically, I want to use nullable types only if necessary. Though since `ISettings<T>` is used in order to support different settings sources, the generic use of nullable types seems to be an acceptable tradeoff.

image

I am going to let `ViewSetting<T>` implement both interfaces `ISetting<T>` and `IViewSetting`.

mstv:
This bugfix should be merged before the next release.

GitCommands/Settings/IRuntimeSetting.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace GitCommands.Settings
+{
+    public interface IRuntimeSetting
```
RussKie, line 3: Could you please annotate the type itself?

GitUI/CommandsDialogs/FormRebase.cs
```diff
@@ -379,9 +379,9 @@ private void chkUseFromOnto_CheckedChanged(object sender, EventArgs e)
 
         private void btnChooseFromRevision_Click(object sender, EventArgs e)
         {
-            bool previousValueBranchFilterEnabled = AppSettings.BranchFilterEnabled;
-            bool previousValueShowCurrentBranchOnly = AppSettings.ShowCurrentBranchOnly;
-            bool previousValueShowReflogReferences = AppSettings.ShowReflogReferences;
+            bool previousValueBranchFilterEnabled = AppSettings.BranchFilterEnabled.Value;
```
RussKie, line 382: We need to provide the implicit conversion to/from bool to avoid changes like this one, and will make the changeset much smaller and cleaner.

GitCommands/Settings/RuntimeSetting.cs
```diff
@@ -0,0 +1,87 @@
...
+        private readonly ISetting<T> _persistentSetting;
+
+        private bool _loaded;
+        private T _value;
+
+        /// <inheritdoc />
+        public event EventHandler? Updated;
+
+        /// <inheritdoc />
+        public T Default { get => _persistentSetting.Default; }
```
RussKie, line 14:
```suggestion
        public T Default => _persistentSetting.Default;
```

ditto below

GitCommands/Settings/RuntimeSetting.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private readonly ISetting<T> _persistentSetting;
+
+        private bool _loaded;
+        private T _value;
+
+        /// <inheritdoc />
+        public event EventHandler? Updated;
+
+        /// <inheritdoc />
```
RussKie, line 13: This is unnecessary

GitCommands/Settings/RuntimeSetting.cs
```diff
@@ -0,0 +1,87 @@
+﻿namespace GitCommands.Settings
+{
+    public class RuntimeSetting<T> : ISetting<T>, IRuntimeSetting
```
RussKie, line 3: Could you please annotate the type as well?

GitCommands/Settings/BoolRuntimeSetting.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace GitCommands.Settings
+{
+    public sealed class BoolRuntimeSetting : RuntimeSetting<bool>
+    {
+        public BoolRuntimeSetting(SettingsPath settingsSource, string name, bool defaultValue)
+        : base(Setting.Create(settingsSource, name, defaultValue))
```
RussKie, line 6: [nit] mis-indented
ditto below

GitCommands/Settings/BoolRuntimeSetting.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace GitCommands.Settings
+{
+    public sealed class BoolRuntimeSetting : RuntimeSetting<bool>
```
RussKie, line 3: Could you please annotate all new public types and API?

GitUI/CommandsDialogs/FormRebase.cs
mstv: Done for "to `T`".
I would say it is impossible for "from `T`" because `operator =` cannot be overloaded (in contrast to C++) and I cannot create full-blown `RuntimeSettings<T>` from a bool or enum value.

GitCommands/Settings/IRuntimeSetting.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Reloads the value from the settings file.
+        /// </summary>
+        void Reload();
+
+        /// <summary>
+        /// Resets the value to the hard-coded default value.
+        /// </summary>
+        void ResetToDefault();
```
RussKie, line 16:
```suggestion
        void Reset();
```

GitCommands/Settings/IRuntimeSetting.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public interface IRuntimeSetting
+    {
+        /// <summary>
+        /// Reloads the value from the settings file.
+        /// </summary>
+        void Reload();
+
+        /// <summary>
+        /// Resets the value to the hard-coded default value.
```
RussKie, line 14: hard-coded?

GitCommands/Settings/RuntimeSetting.cs
```diff
@@ -0,0 +1,84 @@
...
+        public RuntimeSetting(ISetting<T> persistentSetting)
+        {
+            _persistentSetting = persistentSetting;
+        }
+
+        /// <summary>
+        /// Optionally calls <cref>Reload</cref> and returns the current value.
+        /// </summary>
+        public T GetValue(bool reload = false)
+        {
```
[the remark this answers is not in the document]
RussKie: After some thinking, I don't think we should provide this method here at all. I don't think the FileViewer settings should be changed to the "runtime". 
However, _IF_ you really think they _need to_, then this convenience method should be a part of the FileViewer implementation.

mstv:
A RuntimeSetting has:
1. a hard-coded initial default value for the property in AppSettings.cs
2. a persisted value, which the user has stored as their preferred value (using the menu item "Save current view settings as default")
3. the current runtime value, which may be remembered or not depending on

image

In addition to
a. "Save current view settings as default",
the user might want to
b. "Reload stored view settings"
c. "Reset view settings to factory default"
Should I add them? Where?

These are the type-agnostic operations on a Runtime Setting -> `IRuntimeSetting`, used for the *generic* implementation of these menu items.
The meaning of the short but ambiguous names for the interface methods `Get`, `Set`, `Reset` would not be clear enough to me.
I have added another typed `IRuntimeSetting<T>` which combines `ISetting<T>` with `IRuntimeSetting`. So it has the property `Value` which provides the `Get` / `Set` operation.

(I think this comment resolves all open review comments.)

GitCommands/Settings/RuntimeSetting.cs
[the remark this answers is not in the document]
mstv: `GetValue` is used for `get` of property `Value`, too.
I have declared it in the new typed interface `IRuntimeSetting<T>`, refer to https://github.com/gitextensions/gitextensions/pull/10851#issuecomment-1762871167.

gerhardol (approved):
I find it hard to understand what settings that are saved as "runtime" settings and why.
But implementation is fine to me.
I can accept it.

mstv:
> I find it hard to understand what settings that are saved as "runtime" settings and why.

"Why?" might be answered by the issue #10009. All chosen settings are affected the same way.
In this PR, I have chosen the most important settings which are a nuisance if changed in other instances unintended and unexpected, particularly later on refresh.
We can adapt more view settings in a follow-up PR.
All settings are saved using the menu item.

RussKie (approved):

Issue #10423, opened 2022-11-18
[NBug] Resetting unstaged part of conflicted file throws error


GrahamTheCoder:
### Environment

- Git Extensions 4.0.0.15569
- Build cba315df5366fca01240d93b6e5a5077fd5e6630
- Git 2.38.1.windows.1
- Microsoft Windows NT 10.0.22000.0
- .NET 6.0.11
- DPI 120dpi (125% scaling)

### Issue description

Exit code: 1
Command: C:\Program Files\Git\bin\git.exe
Arguments: checkout-index --index --force -- "src/BrandVue.FrontEnd/cypress/e2e/menuLinks.cy.ts"
Working directory: C:\Repos\Vue\

```
GitExtUtils.ExternalOperationException: git checkout-index: src/BrandVue.FrontEnd/cypress/e2e/menuLinks.cy.ts is unmerged
 ---> System.Exception: git checkout-index: src/BrandVue.FrontEnd/cypress/e2e/menuLinks.cy.ts is unmerged
   --- End of inner exception stack trace ---
   at GitCommands.ExecutableExtensions.GetOutputAsync(IExecutable executable, ArgumentString arguments, Byte[] input, Encoding outputEncoding, CommandCache cache, Boolean stripAnsiEscapeCodes)
   at Microsoft.VisualStudio.Threading.JoinableTask.CompleteOnCurrentThread()
   at GitCommands.ExecutableExtensions.GetBatchOutput(IExecutable executable, ICollection`1 batchArguments, Byte[] input, Encoding outputEncoding, CommandCache cache, Boolean stripAnsiEscapeCodes)
   at GitCommands.GitModule.ResetFiles(IReadOnlyList`1 files)
   at GitUI.CommandsDialogs.FormCommit.ResetSoftClick(Object sender, EventArgs e)
   at GitUI.CommandsDialogs.FormCommit.ResetSelectedFiles()
   at GitUI.CommandsDialogs.FormCommit.ExecuteCommand(Int32 cmd)
   at ResourceManager.GitExtensionsFormBase.ProcessHotkey(Keys keyData)
   at GitUI.CommandsDialogs.FormCommit.ProcessHotkey(Keys keyData)
   at ResourceManager.GitExtensionsFormBase.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
[...]

PR #10852, opened 2023-04-02
Merge conflicts: Only list files in worktree

gerhardol:
Contributes to #10423 

## Proposed changes

The "change" in index is a status marker only, there are no actual changes in the file.
The two appearances are confusing. It is required for a proper fix for #10423 
as the worktree status must be reset to HEAD, not the empty index.

Similar appearance changes in RevDiff and FormCommit.

### Before

image

image

### After

image

## Test methodology 

Manual

## Merge strategy

----


RussKie (commented):
Do we have tests for this?

gerhardol:
> Do we have tests for this?

No, this is Git behavior so there is not much.
There is maybe something for the parsing of the status characters.

gerhardol:
Added a parsing test, including cases that do not fully work resetting
The cases with added files are not handled
https://git-scm.com/docs/git-status#_short_format

Left for later improvements...
remote_deleted
local_deleted

gerhardol:
For discussion, not this PR.

For the related conflicts where files have been deleted in any of the branches, the two commits provide information. For example an example where a file has been modified in both branches, deleted in the remote:

image

image

The worktree icon need to be changed to show both conflict and deleted in worktree, then the command to reset etc can be adjusted similar to #10854

GitCommands/Git/GetAllChangedFilesOutputParser.cs
```diff
@@ -226,7 +227,12 @@ private static IReadOnlyList<GitItemStatus> GetAllChangedFilesFromString_v2(stri
                         continue;
                     }
 
-                    UpdateItemStatus(x, true, subm, fileName, oldFileName, renamePercent);
+                    // Skip unmerged where both are modified: Only worktree interesting.
+                    if (entryType != 'u' || x != 'U' || y != 'U')
```
mstv, line 231: ditto

GitCommands/Git/GetAllChangedFilesOutputParser.cs
```diff
@@ -100,7 +100,8 @@ internal List<GitItemStatus> GetAllChangedFilesFromString_v1(string getAllChange
                     continue;
                 }
 
-                if (x != '?' && x != '!' && x != ' ')
+                // Skip unmerged where both are modified: Only worktree interesting.
+                if ((x != '?' && x != '!' && x != ' ') || !(x == 'U' && y == 'U'))
```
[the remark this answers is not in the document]
gerhardol: That was a a bigger change, to be consistent
Also modified the checks slightly, still ignoring unexpected data, required to ignore error printouts (this can maybe be changed when error output is no longer parsed, separate change).

GitCommands/Git/GetAllChangedFilesOutputParser.cs
```diff
@@ -175,67 +177,87 @@ private static IReadOnlyList<GitItemStatus> GetAllChangedFilesFromString_v2(stri
...
+                    UpdateItemStatus(entryType, false, "N...", otherFileName, null, null);
+                    continue;
                 }
-                else if (entryType == '1' || entryType == '2' || entryType == 'u')
+
+                const char OrdinaryEntry = '1';
+                const char RenamedEntry = '2';
+                const char UnmergedEntry = 'u';
+
+                if (entryType != OrdinaryEntry && entryType != RenamedEntry && entryType != UnmergedEntry)
```
mstv, line 200:
```suggestion
                if ((entryType != OrdinaryEntry && entryType != RenamedEntry && entryType != UnmergedEntry) || line.Length <= 3)
```

GitCommands/Git/GetAllChangedFilesOutputParser.cs
```diff
@@ -92,19 +92,21 @@ internal List<GitItemStatus> GetAllChangedFilesFromString_v1(string getAllChange
...
+                if (fromDiff && staged == StagedStatus.WorkTree && x == GitItemStatusConverter.UnmergedStatus)
                 {
                     // git-diff has two lines to inform that a file is modified and has a merge conflict
                     continue;
                 }
 
-                if (x != '?' && x != '!' && x != ' ')
+                // Skip unmerged where both are modified: Only worktree interesting.
+                if ((x != GitItemStatusConverter.UntrackedStatus && x != GitItemStatusConverter.IgnoredStatus && x != GitItemStatusConverter.UnmodifiedStatus_v1)
+                    || !(x == GitItemStatusConverter.UnmergedStatus && y == GitItemStatusConverter.UnmergedStatus))
```
mstv, line 105: more similar to the change at line 248
```suggestion
                    || x != GitItemStatusConverter.UnmergedStatus || y != GitItemStatusConverter.UnmergedStatus)
```

GitCommands/Git/GetAllChangedFilesOutputParser.cs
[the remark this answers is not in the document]
mstv: Thank you.

mstv (approved):

gerhardol:
> The worktree icon need to be changed to show both conflict and deleted in worktree, then the command to reset etc can be adjusted similar to https://github.com/gitextensions/gitextensions/pull/10854

GitItemStatus need to keep information about both index/worktree to handle this correctly. One or two icons is maybe separate.

Will merge this, to be included in 4.1

GitCommands/Git/GitItemStatusConverter.cs
```diff
@@ -2,21 +2,33 @@
...
+        public static readonly char CopiedStatus = 'C';
+        public static readonly char DeletedStatus = 'D';
+        public static readonly char ModifiedStatus = 'M';
+        public static readonly char IgnoredStatus = '!';
+        public static readonly char RenamedStatus = 'R';
+        public static readonly char SkippedStatus = 'S';
+        public static readonly char UnmergedStatus = 'U';
+        public static readonly char UnmodifiedStatus_v1 = ' ';
+        public static readonly char UnmodifiedStatus_v2 = '.';
+        public static readonly char UntrackedStatus = '?';
```
RussKie, line 15: I think these can be turned into consts...

RussKie (approved):

GitCommands/Git/GitItemStatusConverter.cs
gerhardol: It will not be a practical problem, the const will not be used in other assemblies. But this is the case with all except the plugin interface.
I set this as `static readonly` to not have to make changes, will be happy to change to `const` with comments...

RussKie:
If these are git-originated identifiers then we can presume these to be
constants, i think.

This is a non-blocking comment anyway.

mstv (approved):

gerhardol:
Two minor changes when looking at the changes, merging in a couple of hours.

mstv (approved):

PR #10857, opened 2023-04-04
Rename to "LeftPanel" in hotkeys

pmiossec:
because that's how it is named everywhere else (and it is misleading)

### Before

image

image

### After

image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 32a99950b8e8ede897e844cc6a10fee321fa8513
- Git 2.40.0.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1881,7 +1881,7 @@ namespace GitUI.CommandsDialogs
             FocusDiff = 5,
             FocusFileTree = 6,
             FocusFilter = 18,
-            ToggleBranchTreePanel = 21,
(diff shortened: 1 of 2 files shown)


RussKie (approved):

gerhardol (approved):
I would prefer renaming everything to SidePanel instead, but LeftPanel is used more.
FocusBranchTree too?
The function could be renamed too.
The code is in GitUI/BranchTreePanel/ too

mstv:
> I would prefer renaming everything to SidePanel instead, but LeftPanel is used more.
FocusBranchTree too?
The function could be renamed too.

valid points

PR #10863, opened 2023-04-06
Filter toolbar: Show reflog first and default

gerhardol:
[formbrowse_toolbar_visibility_](https://github.com/gitextensions/gitextensions/pull/10831#issuecomment-1496690268)
https://github.com/gitextensions/gitextensions/pull/10831#pullrequestreview-1358125079

## Proposed changes

Show reflog button first and by default
first-parents button last, after filter

### Before

Reflog is not visible by default, button activated
image

### After

image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/FormBrowseMenus.cs
@@ -197,7 +197,7 @@ namespace GitUI.CommandsDialogs
 
             return;
 
-            bool IsVisibleByDefault(string buttonKey) => !buttonKey.Contains(FilterToolBar.ReflogButtonName) && !buttonKey.Contains(FormBrowse.FetchPullToolbarShortcutsPrefix);
+            bool IsVisibleByDefault(string buttonKey) => !buttonKey.Contains(FormBrowse.FetchPullToolbarShortcutsPrefix);
             static void SaveVisibilitySetting(string key, bool visible, bool defaultValue = true)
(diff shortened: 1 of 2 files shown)


RussKie:
I'd leave 'first parent' where it currently is or remove it completely. I
fear that moving it closer to the frequently used controls would increase
the rate of accidental activation of that feature, which is likely very
niche.

gerhardol:
> I'd leave 'first parent' where it currently is or remove it completely. I fear that moving it closer to the frequently used controls would increase the rate of accidental activation of that feature, which is likely very niche.

Hide by default?
I believe these should be kept together.

RussKie:
Why together?

I prefer to place items that have high use on the toolbar. The 'first
parent' is on the toolbar squarely because some time ago there was an
influx of users who'd toggle it by accident and then run around like
headless chooks. I don't believe this functionality is very popular and I
would not want to have it amidst frequently used items.

mstv:
I do not have a preference regarding the position of the 'first parent' button because I hide it.
Hiding it by default might be a good compromise.

gerhardol:
> Why together?

They are of a similar type, so grouping is natural to me.

> I prefer to place items that have high use on the toolbar.

So do I, but the controls should be grouped in a logical way.
Putting it after filter box then, to keep the filters together and this button away as much as possible...

> Hiding it by default might be a good compromise.

I hide it too, but it was added to show user that this was activated.
The setting is no longer dominant and reset when switching repos, so no longer that important.

mstv (approved):

RussKie (approved):

PR #10870, opened 2023-04-11
Refactoring: Rename to LeftPanel

pmiossec:
Follow-up of #10857 (See comment https://github.com/gitextensions/gitextensions/pull/10857#issuecomment-1500338666 )

## Proposed changes

- A lot of renaming

## Screenshots 

Nothing changed. Only refactoring....

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build bf1bc0d359f0d7705bcf6c851d0d640403ea02f6 (Dirty)
- Git 2.40.0.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

## Merge strategy

----


GitUI/Properties/Settings.Designer.cs
```diff
@@ -122,12 +122,12 @@ internal sealed partial class Settings : global::System.Configuration.Applicatio
         [global::System.Configuration.UserScopedSettingAttribute()]
         [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
         [global::System.Configuration.DefaultSettingValueAttribute("False")]
-        public bool FormBrowse_BranchTreePanel_Collapsed {
```
RussKie, line 125: Can we rename this one though?

RussKie (commented):
Looks good apart from the change of the Settings question

GitUI/Properties/Settings.Designer.cs
pmiossec: From what I see (and if I understand well because I'm not an expert of the setting feature), this is no more use since more than 6 years...
6 years ago, hotkeys has been migrated.
So I think we can even remove the settings file and the hotkey migration code...

gerhardol (approved):
Added a change myself, so someone else to approve too

RussKie:
The build keeps failing...

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -959,7 +959,7 @@ private void InternalInitialize()
 
                     _formBrowseMenus.InsertRevisionGridMainMenuItems(repositoryToolStripMenuItem);
 
-                    // Request all branches if side panel is shown
+                    // Request all branches if leftpanel is shown
```
RussKie, line 962:
```suggestion
                    // Request all branches if left panel is shown
```

Ditto other places

GitUI/Properties/Settings.Designer.cs
RussKie: I don't think it's related to hotkeys. I think this setting controls whether the left panel is shown or not. If we rename the setting we'll break the users. I'd like to have an assessment of how bad this break will be, if we go along.

pmiossec: Maybe I was not clear: yes, it is not related to hotkey.
Once again, if I'm not wrong trying to understand the code, none of of the settings are used anymore as they are handled by `AppSettings` so we won't break anything.

The only setting still used is the hotkey one ( I see 0 references on the other properties) and it's to migrate the hotkey setting value to `AppSettings`.
It has been done 6 years ago, so I think we could even remove this migration logic and the setting file as I think we don't care if a user loose it's hotkey setting because he will migrate directly from a 6 years old version to a recent one. I keep this cleaning commit for the next version...

So, for me, there will be no breaking by renaming this property just to keep consistency (as the settings are no more used).

pmiossec: And I could also add that, in absolute, breaking the toggling state of the left panel **once**, when installing a new version, is not that important... The user has just to toggle it and the problem is solved...
I understand that we must take care of more important settings but a toggling state... 😉

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: That was intentional, to use leftpanel consistently as the name, not "left panel" "leftpanel" and "left-panel" used randomly.
But it can be changed.

gerhardol:
> I would prefer to remain more or less grammatically correct in comments and docs.

It is a name like CommitInfo, same used in commit messages.
But changed to left panel.

Added a commit for FocusBranchTree too

GitUI/Properties/Settings.Designer.cs
mstv: The splitter states including the visibility of the left panel are stored generically.
So, why not remove this?

GitUI/Translation/French.xlf
mstv, line 1: All translation files but English.xlf should be reverted, shouldn't they?

gerhardol: Can as well be included: The info will be OK after the commit but be lost at any other update for this translation.

RussKie (approved):

GitUI/Properties/Settings.Designer.cs
pmiossec: I pushed a commit to remove only the code related to the setting link to the LeftPanel.
And I kept the remove of everything for `master`/ v4.2. Or do you all think that I should do a PR for the v4.1?

pmiossec:
For me, we are good to go. Do I squash and rebase everything?

gerhardol:
> The build keeps failing...

There is a cost of adding and maintaining tests. Retrigger tests in AppVeyor due to failures requires quite some time time (both manhours and leadtime) as well as motivation.

> For me, we are good to go. Do I squash and rebase everything?

I suggest so

RussKie:
There are far greater costs when tests are absent, wouldn't you agree? 😉

gerhardol:
> For me, we are good to go. Do I squash and rebase everything?

I suggest that
It is easier if master and 4.1 do not differ

> There are far greater costs when tests are absent, wouldn't you agree? 😉

Tests adds considerable OH for maintenance, too keep them running.
Many unit tests tend to just be an additional implementation that need to be scrapped at refactoring anyway.
It depends on what resources you have available for a project, where they are best spent.

So it is a matter to have the right tests...
But it is OK to disagree, you decide - I do not want that job.
But I have scrapped some features and corrections when I have realized that it will not be worth it if tests are requested (both time and motivation (if I find tests meaningless)).

pmiossec:
> It is easier if master and 4.1 do not differ

Sorry but I'm not sure to understand the exact meaning of that.
Do you still think it could be merged in v4.1 (and the release branch later merged in master)? If yes, I merge it right now... 

Or do you prefer to keep it for master?

gerhardol:
Merge it!

PR #10878, opened 2023-04-13
Perf (and bugfix) avatars handling

pmiossec:
## Proposed changes

1. Avatar: prevent always rewrite of avatar file on disk (in FileSystemAvatarCache )

when the avatar has been loaded from file.
Consequences:
 * fix (old time) bug where the avatar is **never** refreshed
if we open the repository enough often (i.e. < avatar cache period)
because the last `LastWriteTime` that we use to check
 if the avatar has expired is always update when the avatar is written
 * better perfs

2. Display Avatar and revision grid (slightly?) more efficiently

* Do all the other drawing things before letting more chance to the Task that fetch the avatar to complete
(reducing the number of redraw) => less generic avatar displayed before displaying the good avatar.
* Invalidate only the avatar region to redraw instead of the whole grid when getting the avatar

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 6d1276457497896d7c4b5e06d6c8af5c9ceee634
- Git 2.40.0.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

```
```

## Merge strategy

I agree that the maintainer Rebase merge this PR (if the commit message is clear).

----

### DIFF
--- a/GitUI/Avatars/FileSystemAvatarCache.cs
+++ b/GitUI/Avatars/FileSystemAvatarCache.cs
@@ -29,7 +29,14 @@ namespace GitUI.Avatars
             string cacheDir = AppSettings.AvatarImageCachePath;
             string path = Path.Combine(cacheDir, $"{email}.{imageSize}px.png");
 
-            Image image = ReadImage() ?? await _inner.GetAvatarAsync(email, name, imageSize);
+            Image image = ReadImage();
+
+            if (image is not null)
+            {
+                return image;
+            }
+
+            image = await _inner.GetAvatarAsync(email, name, imageSize);
 
             if (image is not null)
             {
--- a/GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
@@ -13,6 +13,8 @@ namespace GitUI.UserControls.RevisionGrid.Columns
         private readonly RevisionDataGridView _revisionGridView;
         private readonly IAvatarProvider _avatarProvider;
         private readonly IAvatarCacheCleaner _avatarCacheCleaner;
+        private static readonly int _padding = DpiUtil.Scale(2);
+        private static Bitmap _placeholderImage;
 
         public AvatarColumnProvider(RevisionDataGridView revisionGridView, IAvatarProvider avatarProvider, IAvatarCacheCleaner avatarCacheCleaner)
             : base("Avatar")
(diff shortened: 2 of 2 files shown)


mstv (commented):
+1 LGTM, have not run

gerhardol (commented):
+1

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -46,12 +47,32 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
...
+
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Right - 2, avatarRect.Top, 2, 1);
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Right - 1, avatarRect.Top, 1, 2);
+
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Left, avatarRect.Bottom - 1, 2, 1);
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Left, avatarRect.Bottom - 2, 1, 2);
+
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Right - 2, avatarRect.Bottom - 1, 2, 1);
+            e.Graphics.FillRectangle(style.BackBrush, avatarRect.Right - 1, avatarRect.Bottom - 2, 1, 2);
+
```
RussKie, line 72: What are these magic numbers? Should those be scaled as well?
Also, if I interpret this code correctly, it would be more efficient to use clipped path here and draw it in a single operation.

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -46,12 +47,32 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
 
             Column.Width = e.CellBounds.Height;
 
-            var padding = DpiUtil.Scale(2);
```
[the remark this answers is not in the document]
RussKie: The PMv2 support is still not fully supported by Windows Forms, though it became much better in .NET 7, and likely will be completed in .NET 8.
Once we get there we may need to tweak and adjust some layouts.

GitUI/Avatars/FileSystemAvatarCache.cs
```diff
@@ -29,7 +29,14 @@ public FileSystemAvatarCache(IAvatarProvider inner, IFileSystem? fileSystem = nu
...
 
-            Image image = ReadImage() ?? await _inner.GetAvatarAsync(email, name, imageSize);
+            Image image = ReadImage();
+
+            if (image is not null)
+            {
+                return image;
+            }
+
+            image = await _inner.GetAvatarAsync(email, name, imageSize);
```
RussKie, line 39: Do we need a cancellation token here?

pmiossec: I don't know all the places where this code is called from, but from the one where it is call in this PR (so the avatar column in the grid), no because the call is not awaited (`.FileAndForget()`) and the screen is updated when the task finish blocking nothing...

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
pmiossec: At first I just moved some code without really understand it.
After some investigation, I have updated with comments so that it is now much clearer and also discovered that moving the code break the goal of this code so I move it back and found another optimization (cache the placeholder image)

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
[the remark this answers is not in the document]
pmiossec: I saw no difference caching is or not with 2 screens with different scaling...

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -13,6 +13,8 @@ internal sealed class AvatarColumnProvider : ColumnProvider
         private readonly RevisionDataGridView _revisionGridView;
         private readonly IAvatarProvider _avatarProvider;
         private readonly IAvatarCacheCleaner _avatarCacheCleaner;
+        private static readonly int _padding = DpiUtil.Scale(2);
+        private Bitmap _buffer;
```
mstv, line 17:
```suggestion
        private Bitmap _placeholderImage;
```

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -46,55 +48,70 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
...
+            if (isLoaded)
             {
 #pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
                 image = imageTask.Result;
 #pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
             }
             else
             {
+                // First time, draw resized and cache placholder image
+                if (_buffer == null)
```
mstv, line 72:
```suggestion
                if (_placeholderImage is null)
```

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -46,55 +48,70 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
...
             }
 
-            Rectangle rect = new(
-                e.CellBounds.Left + padding,
-                e.CellBounds.Top + padding,
-                imageSize,
-                imageSize);
-
-            var container = e.Graphics.BeginContainer();
-            e.Graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
```
mstv, line 85: This line should not be removed I guess.

pmiossec: I can reinstate it but for me this line is useless because the image is retrieved from the avatar service already add the right size and so don't need any resize...

mstv: Then it is OK for me. I was not sure whether it got lost when extracting the code for the placeholder.

GitUI/UserControls/RevisionGrid/Columns/AvatarColumnProvider.cs
```diff
@@ -48,51 +48,49 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
...
-#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
-                image = imageTask.Result;
-#pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
-            }
-            else
-            {
-                // First time, draw resized and cache placholder image
-                if (_buffer == null)
+                // First time, draw at the good size the placeholder image and cache it
+                if (_placeholderImage == null)
```
mstv, line 64:
```suggestion
                if (_placeholderImage is null)
```

mstv (approved):
final nit from my side

RussKie (approved):

PR #10879, opened 2023-04-13
Improve rebase onto commit selection

pmiossec:
Improvement over #9686

## Proposed changes

- Hide stashes in revision grid when selecting commit range (now that we display all the stashes in the revision grid)
- Select first parent which is a better default selection. So if the user want to rebase only one commit (which is not so rare), he only has to validate the selection...

### Before

We see stashes and the selection is on the tip of the branch:
image

### After

Stashes are not displayed and the 2nd commit of the branch is selected
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 74ad45025d7877460d65b48ccfdab2fb8b779743
- Git 2.40.0.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

## Merge strategy

Rebase merge.

----

### DIFF
--- a/GitUI/CommandsDialogs/FormRebase.cs
+++ b/GitUI/CommandsDialogs/FormRebase.cs
@@ -386,7 +386,9 @@ namespace GitUI.CommandsDialogs
             try
             {
                 AppSettings.ShowStashes = false;
-                using FormChooseCommit chooseForm = new(UICommands, txtFrom.Text, showCurrentBranchOnly: true);
+                ObjectId firstParent = UICommands.GitModule.RevParse("HEAD~");
+                string preSelectedCommit = !string.IsNullOrWhiteSpace(txtFrom.Text) ? txtFrom.Text : firstParent?.ToString() ?? string.Empty;
+                using FormChooseCommit chooseForm = new(UICommands, preSelectedCommit, showCurrentBranchOnly: true);
                 if (chooseForm.ShowDialog(this) == DialogResult.OK && chooseForm.SelectedRevision is not null)
                 {
                     txtFrom.Text = chooseForm.SelectedRevision.ObjectId.ToShortString();
--- a/GitUI/CommandsDialogs/FormRebase.cs
+++ b/GitUI/CommandsDialogs/FormRebase.cs
@@ -381,9 +381,11 @@ namespace GitUI.CommandsDialogs
             bool previousValueBranchFilterEnabled = AppSettings.BranchFilterEnabled;
             bool previousValueShowCurrentBranchOnly = AppSettings.ShowCurrentBranchOnly;
             bool previousValueShowReflogReferences = AppSettings.ShowReflogReferences;
+            bool previousValueShowStashes = AppSettings.ShowStashes;
 
             try
             {
+                AppSettings.ShowStashes = false;
                 using FormChooseCommit chooseForm = new(UICommands, txtFrom.Text, showCurrentBranchOnly: true);
                 if (chooseForm.ShowDialog(this) == DialogResult.OK && chooseForm.SelectedRevision is not null)
                 {
@@ -392,6 +394,7 @@ namespace GitUI.CommandsDialogs
             }
             finally
             {
+                AppSettings.ShowStashes = previousValueShowStashes;
                 AppSettings.BranchFilterEnabled = previousValueBranchFilterEnabled;
                 AppSettings.ShowCurrentBranchOnly = previousValueShowCurrentBranchOnly;
                 AppSettings.ShowReflogReferences = previousValueShowReflogReferences;


GitUI/CommandsDialogs/FormRebase.cs
```diff
@@ -381,17 +381,22 @@ private void btnChooseFromRevision_Click(object sender, EventArgs e)
             bool previousValueBranchFilterEnabled = AppSettings.BranchFilterEnabled;
             bool previousValueShowCurrentBranchOnly = AppSettings.ShowCurrentBranchOnly;
             bool previousValueShowReflogReferences = AppSettings.ShowReflogReferences;
+            bool previousValueShowStashes = AppSettings.ShowStashes;
 
             try
             {
-                using FormChooseCommit chooseForm = new(UICommands, txtFrom.Text, showCurrentBranchOnly: true);
+                AppSettings.ShowStashes = false;
+                ObjectId firstParent = UICommands.GitModule.RevParse(Currentbranch.Text + "~");
```
mstv, line 389: Does this also work for detached head?

GitUI/CommandsDialogs/FormRebase.cs
```diff
@@ -381,17 +381,22 @@ private void btnChooseFromRevision_Click(object sender, EventArgs e)
...
+                var preSelectedCommit = string.IsNullOrEmpty(txtFrom.Text) ? firstParent?.ToString() ?? string.Empty : txtFrom.Text;
```
mstv, line 390:
```suggestion
                string preSelectedCommit = string.IsNullOrEmpty(txtFrom.Text) ? firstParent?.ToString() ?? string.Empty : txtFrom.Text;
```

GitUI/CommandsDialogs/FormRebase.cs
gerhardol, line 389: Side note: This should be queried from RevGrid to avoid extra Git commands. It is Git access that slows down the app...

gerhardol: yes, failure to parse will give null (no error popup)

gerhardol (commented):
+1

RussKie (approved):
Modulo the other comments

GitUI/CommandsDialogs/FormRebase.cs
```diff
@@ -381,17 +381,22 @@ private void btnChooseFromRevision_Click(object sender, EventArgs e)
...
             bool previousValueShowCurrentBranchOnly = AppSettings.ShowCurrentBranchOnly;
             bool previousValueShowReflogReferences = AppSettings.ShowReflogReferences;
+            bool previousValueShowStashes = AppSettings.ShowStashes;
 
             try
             {
-                using FormChooseCommit chooseForm = new(UICommands, txtFrom.Text, showCurrentBranchOnly: true);
+                AppSettings.ShowStashes = false;
+                ObjectId firstParent = UICommands.GitModule.RevParse("HEAD~");
+                string preSelectedCommit = string.IsNullOrEmpty(txtFrom.Text) ? firstParent?.ToString() ?? string.Empty : txtFrom.Text;
```
RussKie, line 390: `Text` value can't be null, but it can whitespace. So the check should be `IsNullOrWhitespace`. 
Also suggest to invert the condition as it'd read easier.

RussKie (approved):

GitUI/CommandsDialogs/FormRebase.cs
pmiossec: I have a little changed that to improve when in detached head and have the same behavior (even if I really don't see the case of rebasing in detached head)

pmiossec: I disagree in this case.
In this control we don't have access to the revision grid and I had a look, it will add a lot of logic to be able to pass the needed information as the form browse is called in multiple case (rebase, interactive, rebase onto, ....). So more than 4 or 6 different cases to handle.

And for the performance, doing a revparse is very quick (around 40ms) which is much less than requesting the logs to display the revision grid and opening the form. And also the goal of this feature is to make the use more convenient when you want to rebase onto only one commit and in this case, you probably gain 2 or 3s. So the net gain should be good 😉

pmiossec:
> If there isn't a toggle to take off the first parent then I don't think it's a good idea. That will upset a lot of people if they can't select a mid work commit. If a visible toggle is there to allow them to turn it off, then that would be ok.

@vbjay I absolutely don' t understand you point of view.
This feature prevent no one to select the commit he wants. That's just that a better commit is pre-selected (because it makes no sense for the user to select the 1st as base) and so in the best case the user has just to validate. Wort case, he just has to select the commit he wants as base commit.
The only thing that could happen to the user is that he economize 1 click and some seconds...

mstv (approved):

vbjay:
> > If there isn't a toggle to take off the first parent then I don't think it's a good idea. That will upset a lot of people if they can't select a mid work commit. If a visible toggle is there to allow them to turn it off, then that would be ok.
> 
> @vbjay I absolutely don' t understand you point of view.
> This feature prevent no one to select the commit he wants. That's just that a better commit is pre-selected (because it makes no sense for the user to select the 1st as base) and so in the best case the user has just to validate. Wort case, he just has to select the commit he wants as base commit.
> The only thing that could happen to the user is that he economize 1 click and some seconds... 

Select first parent  along with commit grid sounded like he set show first parents.

mstv:
> Select first parent along with commit grid after image sounded like he set show first parents.

Yes, but only for short time. :)

vbjay:
How do you know they didn't want to rebase in-between 2 first parents?  Forcing them to only select between 1st parents without giving them the option to turn it off is my problem with it.

pmiossec:
> How do you know they didn't want to rebase in-between 2 first parents? Forcing them to only select between 1st parents without giving them the option to turn it off is my problem with it.

In 10 years of using GE, I never tried to do that so it should not happen too often.
So in this case, either they use #10885 or they copy/paste the hash from revision grid should be enough....

vbjay:
You ate not the only user of ge. All I'm saying is if you are going to only show first parents in the select commit dialog, then show the first parents togle button on dialog to make it clear and to allow the user to have control. I don't mind first parents. I do have a problem with it being forced.

pmiossec:
> All I'm saying is if you are going to only show first parents in the select commit dialog

If fact I don't know why you start a discussion on a false subject....
The PR is absolutely not about that (it's just hide stash and preselect another commit)
And looking at the code, what you could have easily done, we do not enable the "view first parent" feature.
SO everything is ok....

vbjay:
> > All I'm saying is if you are going to only show first parents in the select commit dialog
> 
> If fact I don't know why you start a discussion on a false subject....
> The PR is absolutely not about that (it's just hide stash and preselect another commit)
> And looking at the code, what you could have easily done, we do not enable the "view first parent" feature.
> So everything is ok....


Awesome. It was the description and the image that made me think the select dialog had been altered to only show first parents.  I could see a ton of people who use that dialog getting very upset over not being able to select a non 1st parent. Having it automatically go down one commit so we don't have to is another thing entirely. The image made me think it was the first thing. So I made sure.

PR #10880, opened 2023-04-13
No async when minimized

gerhardol:
## Proposed changes

Do not run git-ls-files by default when in minimized to not interfere with restoring.
The setting was added in
cccfbc967ce835dfd64452a4c08db478936499b6 (#10847)
but required explicit configuration.

--
I have used this on/off for some days and get quite consistent the problem with the default implementation. 
The drawback with activating the setting is that "is conflicted" is not always detected.

The drawback of making this hack permanent is that the core problem will not be fixed.

## Test methodology 

Manual

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -2042,7 +2042,7 @@ namespace GitCommands
 
         public static bool GitAsyncWhenMinimized
         {
-            get => GetBool("GitAsyncWhenMinimized", true);
+            get => GetBool("GitAsyncWhenMinimized", false);
         }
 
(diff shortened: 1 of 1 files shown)


mstv:
I guess your branch "feature/i9223-async-conflict" is the reason why this bug does not occur to me.

gerhardol:
> I guess your branch "feature/i9223-async-conflict" is the reason why this bug does not occur to me.

Maybe.

The problem mostly fixed in #10802  was not new, it had supposedly occasionally occurred for many years.
Still it did not occur for everyone. I assume it is the same after the change, but not always visible.
This configuration seem to fix it permanently for me.

RussKie (approved):

mstv (commented):
+0

PR #10884, opened 2023-04-14
Improve commit selection for 'Find file' feature

pmiossec:
before launching the search by
* selecting the last selected when multiple commits are selected (otherwise we don't know on which commit we search)
* selecting HEAD if the selected commit is artificial

### Before

We can't launch the feature, start searching and discover that we end up nowhere
original_find_file

### After

If on artificial commit, it switch to HEAD before searching
improve_find_file

When multiselect, only the commit we are searching in stay selected so that we know which one is really used:
multiselect_find_file

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 9855deffe1093125ee7f6b1980fc88c667dc6ab8
- Git 2.40.0.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

```
```

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/FormBrowse.cs
+++ b/GitUI/CommandsDialogs/FormBrowse.cs
@@ -1960,6 +1960,14 @@ namespace GitUI.CommandsDialogs
 
         private void FindFileInSelectedCommit()
         {
+            IReadOnlyList<GitRevision> selectedRevisions = RevisionGrid.GetSelectedRevisions();
+            if (selectedRevisions.Count > 1 || (selectedRevisions.Count == 1 && selectedRevisions[0].IsArtificial))
+            {
+                GitRevision potentialRevision = selectedRevisions[0];
+                ObjectId? targetCommit = potentialRevision.IsArtificial ? RevisionGrid.CurrentCheckout : potentialRevision.ObjectId;
+                RevisionGrid.SetSelectedRevision(targetCommit);
+            }
+
             CommitInfoTabControl.SelectedTab = TreeTabPage;
 
             AppSettings.ShowSplitViewLayout = true;


GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1960,6 +1960,14 @@ private void AddNotes()
 
         private void FindFileInSelectedCommit()
         {
+            var selectedRevisions = RevisionGrid.GetSelectedRevisions();
+            if (selectedRevisions.Count > 1 || (selectedRevisions.Count == 1 && selectedRevisions[0].IsArtificial))
+            {
+                GitRevision potentialRevision = selectedRevisions[0];
+                ObjectId targetCommit = potentialRevision.IsArtificial ? RevisionGrid.CurrentCheckout : potentialRevision.ObjectId;
```
gerhardol, line 1967: Will not work if artificial commit is in the multiselect
Not the most important use case though, but removing the artificial commits from selectedRevisions is not that hard...

gerhardol (commented):
+1

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: I will test again to be sure I didn't broke something when refactoring but I have tested this multi-commits with the 1st artificial and it was working...

gerhardol: First artificial should work, but you could select two artificial commits

pmiossec: For me this case is handled by the ternary in the if clause...

mstv: I agree, Philippe. What have me missed, Gerhard?

The repetition can be avoided (and an empty selection be handled, too):
```
GitRevision? potentialRevision = selectedRevisions.FirstOrDefault();
ObjectId? targetCommit = potentialRevision?.IsArtificial is not false ? RevisionGrid.CurrentCheckout : selectedRevisions.Count == 1 ? null : potentialRevision!.ObjectId;
if (targetCommit is not null)
```

GitUI/CommandsDialogs/FormBrowse.cs
```diff
@@ -1960,6 +1960,14 @@ private void AddNotes()
 
         private void FindFileInSelectedCommit()
         {
+            var selectedRevisions = RevisionGrid.GetSelectedRevisions();
```
mstv, line 1963: nit: in order to emphasize that it is not an enumerator
```suggestion
            IReadOnlyList<GitRevision> selectedRevisions = RevisionGrid.GetSelectedRevisions();
```

GitUI/CommandsDialogs/FormBrowse.cs
gerhardol: Yes this works.
Just null annotation (is OK for SetSelectedRevision())

```suggestion
                ObjectId? targetCommit = potentialRevision.IsArtificial ? RevisionGrid.CurrentCheckout : potentialRevision.ObjectId;
```

gerhardol (approved):

GitUI/CommandsDialogs/FormBrowse.cs
pmiossec: Sorry @mstv I find the LoC too difficult to understand...

RussKie (approved):

mstv (approved):

PR #10887, opened 2023-04-16
FileStatusList: Avoid blocking check for Visual Studio

### LINKED ISSUE #10856: `VisualStudioIntegration.IsVisualStudioRunning` can dead lock the app
### Environment

- Git Extensions 4.0.2.16100
- Git 2.39.2.windows.1
- .NET 6.0.15
- DPI 96dpi (no scaling)

```
    Microsoft.WindowsDesktop.App 8.0.0-preview.2.23128.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

Invoking the context menu that contains "Open in Visual Studio" menu can dead lock the app and the only option is to hard kill the process.
https://github.com/gitextensions/gitextensions/blob/037cb04d3b1168d197643512b5bd2d1b506042e1/GitUI/UserControls/FileStatusList.cs#L1412

I've taken a memdump:
image

### Steps to reproduce

It's quite intermittent but I observed this issue few times over the months. In essence, it occurs whenever I right click on files in FormCommit dialog that I want to stage or reset.

_No response_

### Diagnostics

_No response_

mstv:
Fixes #10856

## Proposed changes

- Do not check for a running Visual Studio on every open of the FileStatusList context menu
- Check for installed MSVS on startup in background
- Open a new instance if no running instance has opened a project with this file instead of error popup

## Screenshot

unchanged

image

removed

image

## Test methodology 

- manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusList.cs
+++ b/GitUI/UserControls/FileStatusList.cs
@@ -102,6 +102,8 @@ namespace GitUI
 
             base.Enter += FileStatusList_Enter;
 
+            VisualStudioIntegration.Init();
+
             return;
 
             ImageList CreateImageList()
@@ -199,14 +201,10 @@ namespace GitUI
                 };
                 item.Click += (_, _) =>
                 {
-                    var itemName = SelectedItemAbsolutePath;
-                    if (itemName != null && !VisualStudioIntegration.TryOpenFile(itemName))
+                    string? itemName = SelectedItemAbsolutePath;
+                    if (itemName is not null)
                     {
-                        MessageBox.Show(
-                            TranslatedStrings.OpenInVisualStudioFailureText,
-                            TranslatedStrings.OpenInVisualStudioFailureCaption,
-                            MessageBoxButtons.OK,
-                            MessageBoxIcon.Warning);
+                        VisualStudioIntegration.OpenFile(itemName);
                     }
                 };
                 return item;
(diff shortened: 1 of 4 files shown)


GitUI/UserControls/FileStatusList.cs
```diff
@@ -1409,8 +1409,10 @@ private void FileStatusListView_ContextMenu_Opening(object sender, CancelEventAr
                 cm.Items.Add(_NO_TRANSLATE_openInVisualStudioMenuItem);
             }
 
-            _NO_TRANSLATE_openInVisualStudioMenuItem.Visible = _openInVisualStudioSeparator.Visible = VisualStudioIntegration.IsVisualStudioRunning;
-            _NO_TRANSLATE_openInVisualStudioMenuItem.Enabled = File.Exists(SelectedItemAbsolutePath);
+            bool fileExists = File.Exists(SelectedItemAbsolutePath);
+            _NO_TRANSLATE_openInVisualStudioMenuItem.Enabled = fileExists;
```
pmiossec, line 1413: A specific reason why you kept the 'Enabled' whereas the item won't be visible?

mstv: Please see the following lines... (I have changed the sequence.)

gerhardol: Remove IsVisualStudioRunning too?

gerhardol (commented):
+1
This kind of removes the motivation for the menu item.
But I have no better fix for VisualStudioIntegration.

mstv:
> This kind of removes the motivation for the menu item.

Do you need an indication in GE whether MSVS is running (perhaps only with a different project)?
Then GE should rather start MSVS instead of the error popup.
The menu item should be hidden if MSVS is not installed.
I will leave these two for a follow-up PR.

mstv:
> Then GE should rather start MSVS instead of the error popup.
The menu item should be hidden if MSVS is not installed.

Done with the help of SO.

> The core usability problem is still that the designer opens by default.

Yes, the Solution Explorer has the same caveat. MSVS should have an internal option whether to open the code or the designer by default.
A quick search did not reveal an alternative to "File.OpenFile".

GitUI/VisualStudioIntegration.cs
```diff
@@ -1,15 +1,52 @@
...
     internal static class VisualStudioIntegration
     {
-        public static bool TryOpenFile(string filePath)
+        static VisualStudioIntegration()
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await TaskScheduler.Default;
+                Executable executable = new($@"{Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}\Microsoft Visual Studio\Installer\vswhere.exe");
```
mstv, line 18: Needs to check yet whether `vswhere.exe` exists.

GitUI/UserControls/FileStatusList.cs
mstv: > Remove IsVisualStudioRunning too?

turned into `IsVisualStudioInstalled`

> A specific reason why you kept the 'Enabled' whereas the item won't be visible?

I am used to set both because I am not sure whether `Visible = false` overrides `Enabled`.
In theory, mnemonics could still be active if the item is enabled but invisible.

RussKie (approved):

PR #10889, opened 2023-04-17
Update English.xlf

1Jesper1:
Fix labels

Fixes #

## Proposed changes

- 
- 
- 

## Test methodology 

- 
- 
- 

## Test environment(s) 

- GIT 
- Windows 

## Merge strategy

----

### DIFF
--- a/GitUI/Translation/English.xlf
+++ b/GitUI/Translation/English.xlf
@@ -2468,7 +2468,7 @@ Do you want to continue?</source>
         <target />
       </trans-unit>
       <trans-unit id="commitInfoRightwardMenuItem.Text">
-        <source>Commit info &amp;right of graph</source>
+        <source>Commit info right of graph</source>
         <target />
       </trans-unit>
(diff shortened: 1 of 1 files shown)


pmiossec:
This file is generated so it is not the right way to fix it as your changes will be overwritten (and the build will not pass so not chance it will be merged)...

gerhardol:
The text is used for menu items where the keyboard accelrator is used as well as for tooltips.
Either the label is duplicated or the label is 'cleaned' before being added as a ToolTip (maybe a hack but the least effort for everyone).

As pmiossec wrote: Must be changed in the .cs files.

pmiossec:
I think we already fixed it somewhere for simple toolbar button but I can't find where anymore. For example for "File Explorer", in v4.1 it's fixed. But not for the split button subitems tooltips....

pmiossec:
So partly done in v4.1 here: #10781

1Jesper1:
@pmiossec Nice, also this one? 
image

PR #10893, opened 2023-04-18
CI: more meaningful icon instead of displaying a circle

pmiossec:
### Before

Unknown and stopped are both grey
image

### After

Status: Failure, Success, Unstable, Stopped, In progress, Unknown
image

## Test methodology 

- Manual

## Test environment(s) 

- Git Extensions 33.33.33
- Build 2bc70dbb9e9d797be7aa8c9a96cf5453786b38cf (Dirty)
- Git 2.40.0.windows.1
- .NET 6.0.16
- DPI 96dpi (no scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
+++ b/GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
@@ -1,5 +1,4 @@
 using System.Diagnostics.CodeAnalysis;
-using System.Drawing.Drawing2D;
 using GitCommands;
 using GitCommands.Settings;
 using GitExtUtils.GitUI;
@@ -79,36 +78,10 @@ namespace GitUI.UserControls.RevisionGrid.Columns
                 return;
             }
 
-            Size size;
+            string text = (AppSettings.ShowBuildStatusIconColumn ? revision.BuildStatus.StatusSymbol : string.Empty)
+                + (AppSettings.ShowBuildStatusTextColumn ? (string)e.FormattedValue : string.Empty);
 
-            if (AppSettings.ShowBuildStatusIconColumn)
-            {
-                size = DpiUtil.Scale(new Size(8, 8));
-
-                Point location = new(
-                    e.CellBounds.Left + (size.Width / 2),
-                    e.CellBounds.Top + ((e.CellBounds.Height - size.Height) / 2));
-
-                GraphicsContainer container = e.Graphics.BeginContainer();
-                e.Graphics.SmoothingMode = SmoothingMode.AntiAlias;
-                using Brush brush = CreateCircleBrush();
-                e.Graphics.FillEllipse(brush, new Rectangle(location, size));
-                e.Graphics.EndContainer(container);
-            }
-            else
-            {
-                size = default;
-            }
-
-            if (AppSettings.ShowBuildStatusTextColumn)
-            {
-                _grid.DrawColumnText(
-                    e,
-                    (string)e.FormattedValue,
-                    style.NormalFont,
-                    GetColor(style.ForeColor),
-                    bounds: e.CellBounds.ReduceLeft(size.Width * 2));
-            }
+            _grid.DrawColumnText(e, text, style.NormalFont, GetColor(style.ForeColor), bounds: e.CellBounds);
 
             Color GetColor(Color foreColor)
             {
@@ -133,42 +106,12 @@ namespace GitUI.UserControls.RevisionGrid.Columns
                         customColor = Color.OrangeRed;
                         break;
                     case BuildInfo.BuildStatus.Stopped:
+                    default:
                         customColor = isSelected ? Color.LightGray : Color.Gray;
                         break;
-
-                    default:
-                        throw new InvalidOperationException("Unsupported build status enum value.");
                 }
 
                 return customColor.AdaptTextColor();
-            }
-
-            Brush CreateCircleBrush()
-            {
-                Color color;
-                switch (revision.BuildStatus.Status)
-                {
-                    case BuildInfo.BuildStatus.Success:
-                        color = Color.LightGreen;
-                        break;
-                    case BuildInfo.BuildStatus.Failure:
-                        color = Color.Red;
-                        break;
-                    case BuildInfo.BuildStatus.InProgress:
-                        color = _lightBlue;
-                        break;
-                    case BuildInfo.BuildStatus.Unstable:
-                        color = Color.DarkOrange;
-                        break;
-                    case BuildInfo.BuildStatus.Stopped:
-                    case BuildInfo.BuildStatus.Unknown:
-                        color = Color.Gray;
-                        break;
-                    default:
-                        throw new InvalidOperationException("Unsupported build status enum value.");
-                }
-
-                return new SolidBrush(color.AdaptBackColor());
             }
         }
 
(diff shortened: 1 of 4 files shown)


RussKie:
-0, sorry.
I'm very much used to the circles - I find the traffic lights sufficiently
meaningful. The icons on the screenshot look somewhat out of place...
But I won't block if others approve.

GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
```diff
@@ -79,36 +78,10 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
                 return;
             }
 
-            Size size;
+            string text = (AppSettings.ShowBuildStatusIconColumn ? revision.BuildStatus.StatusIcon : string.Empty)
```
gerhardol, line 81: I was confused about this change, expected this to be handled as two separate columns, but it only seem to be so in the settings

```suggestion
            // Settings presents this as two columns but the variance is in the text
            string text = (AppSettings.ShowBuildStatusIconColumn ? revision.BuildStatus.StatusIcon : string.Empty)
```

mstv:
The code looks good. Icons are easier to understand. Though I am not sure yet whether I find it too distracting and too uneven.
The tooltip explains the details anyway.

This is how it looks with "Segoe UI, 9":

image

Plugins/GitUIPluginInterfaces/BuildServerIntegration/BuildInfo.cs
```diff
@@ -22,5 +22,27 @@ public enum BuildStatus
...
+                    case BuildStatus.Failure:
+                        return "❌";
+                    case BuildStatus.InProgress:
+                        return "▶️";
+                    case BuildStatus.Stopped:
+                        return "⏹️";
+                    case BuildStatus.Unstable:
+                        return "❗";
+                    default:
+                        return "❓";
```
RussKie, line 43: Where did these pictograms come from? Is this some kind of symbol font?

gerhardol: These are UTF-8 characters, so the font need to handle these Unicode characters.
If not supported the "unknown char" should show up in the "traffic light" color, so not so different from now, I cannot see it as a showstopper.
If there is a way to check if a character has support for a specific Unicode char, maybe it could be replaced with a common char?
(I have no experience in saying where these specific chars are supported.)

mstv (commented):
Using real icons would result in a more reliable and more consistent look.
I fear getting bug reports from users who have chosen fonts without these glyphs.
(And I do not like how the cross is rendered on my PC. Your screenshot looks better. But the colors appear darker than with the circles before.)

gerhardol:
> I fear getting bug reports from users who have chosen fonts without these glyphs.

The question mark icon looks similar to the traffic light so not a major issue. But there may be issues raised.
Maybe special font configuration?
At least in .xml.

RussKie:
I'm still unconvinced this is a good change. 

I'm not aware of any user complaints that the dots are confusing (bar @pmiossec, obviously, since you spent some of you valuable time making this change. I also wish we had a discussion before you spent time working on this).

I'm, however, aware of font-related issues reported here and in dotnet/winforms repo, and this change can become an added source of issues since it's font-dependant.

gerhardol:
Dots are not confusing but this is clearer for me. Colors and shapes. 
The worst case is not bad, question mark has colors still much like the traffic light. 

We can add a setting (hidden?) or a scream test (that RussKie suggested for settings).

vbjay:
@gerhardol @RussKie 'a point is valid. You don't know what dints they have and you will run into the unknown Unicode character because of font. No we should not have a setting to allow your font problem. The right way is to use an image instead of font.

pmiossec:
I made this quickly as I see it as a readability improvement. But no more
discussion is needed to see it does not fit in v4.1

I will re-evaluate. I have one or two idea (like using a default font, not
the user defined one use for the revision grid, adding a setting to display
a symbol or a circle,... )

gerhardol:
Time to resurrect?
I still believe that a colored `?` is not too bad if it would occur (can be an option too).
Rewrite with images could be done too.

RussKie:
I like the current dots

vbjay:
I like the icon idea in consideration of color blind people.

pmiossec:
> I still believe that a colored `?` is not too bad if it would occur (can be an option too).

Same gray than "stopped" state?


> Rewrite with images could be done too.

The advantage of unicode char (or drawing traffic light like before) is that you could easily change the colors when line is selected or not. So I prefer to avoid using images.

>I fear getting bug reports from users who have chosen fonts without these glyphs.
I have pushed a change to always use a default font to not have the problem of selecting a fonts that doesn't support these characters and also have consistent display.

> But the colors appear darker than with the circles before

Before, the build status text and icon were not using the same color (there were 2 methods to get colors from build status). It feels strange to me so I removed one so that both used the same. I think it is better because the colors less catch the eyes and also are consistent with the ones of local and remote branches so we don't end up with a Christmas tree (even if it will be the good period of the year).

mstv (commented):
LGTM

@RussKie, do you wish to have an option?

GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
```diff
@@ -79,36 +79,15 @@ public override void OnCellPainting(DataGridViewCellPaintingEventArgs e, GitRevi
...
-                GraphicsContainer container = e.Graphics.BeginContainer();
-                e.Graphics.SmoothingMode = SmoothingMode.AntiAlias;
-                using Brush brush = CreateCircleBrush();
-                e.Graphics.FillEllipse(brush, new Rectangle(location, size));
-                e.Graphics.EndContainer(container);
-            }
-            else
-            {
-                size = default;
+                _fontWithUnicodeCache = new Font(FontFamily.GenericMonospace, style.NormalFont.Size);
```
RussKie, line 87: Can we pre-allocate or allocate only once?

Plugins/GitUIPluginInterfaces/BuildServerIntegration/BuildInfo.cs
```diff
@@ -22,5 +22,27 @@ public enum BuildStatus
         public bool ShowInBuildReportTab { get; set; } = true;
         public string? Tooltip { get; set; }
         public string? PullRequestUrl { get; set; }
+
+        public string StatusPicto
```
RussKie, line 26: What's a "picto"?
```suggestion
        public string StatusSymbol
```

GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
```diff
@@ -20,6 +19,7 @@ internal sealed class BuildStatusColumnProvider : ColumnProvider
 
         // Increase contrast to selected rows
         private readonly Color _lightBlue = Color.FromArgb(130, 180, 240);
+        private Font _fontWithUnicodeCache = null;
```
RussKie, line 22:
```suggestion
        private Font? _fontWithUnicodeCache = null;
```

RussKie:
> @RussKie, do you wish to have an option?

I'm not excited by this change, but I don't think we need another setting.

GitUI/UserControls/RevisionGrid/Columns/BuildStatusColumnProvider.cs
pmiossec: It is already the case. We allocate only once or when the font size change (which should be not that often)

mstv (approved):

gerhardol (approved):

PR #10908, opened 2023-04-22
Intercept and convert \\wsl.localhost\ to \\wsl$\ paths

### LINKED ISSUE #10025: wsl2 fatal error on windows 11
### Environment

- Git Extensions 4.1.0.13845 (latest dev stream. downloaded earlier today)
- Build 54e6bf95e67d0ea80052d3030c923fad73f4633a (Dirty)
- Git 2.36.1.windows.1
- .NET 6.0.4
- DPI 96dpi (no scaling)

### Issue description

Windows 11 is liable to change `\\wsl$` to `\\wsl.localhost` when navigating from the network dir to the wsl root dir, causing a fatal error when attempting to open a wsl repo when the path is `wsl.localhost`:
image

### Steps to reproduce

`Browse...` to open a local repo
Navigate into the wsl filesystem through the folder structure all the way from `\\wsl$` when attempting to open the repo.

This gif shows when `\\wsl$` changes to `\\wsl.localhost`
wsl-path

opening a repo under the `wsl.localhost` path results in this:
open-repo-faulty-path

Workaround:
typing in a longer path and then browsing from there seems to keep `\\wsl$` (most of the time), which allows GitExtensions to properly load the repo:
wsl-longer-path

I never ran into this when using previous dev streams on Windows 10, but I also have no idea how I originally created the pin I surely used when browsing for the repo, so it's possible this also happens on windows 10 🤷‍♂️. I have not tested this on Windows 11 on previous dev streams.

### Diagnostics

_No response_

### LINKED ISSUE #10815: Use native WSL Git executables also for \\wsl.localhost as for \\wsl$
### Feature description

With Windows 11 the `\\wsl$` path is automatically translated in `\\wsl.localhost` and that path change makes unusable this PR merged in the v4.0: [#9702](https://github.com/gitextensions/gitextensions/pull/9702)
It would be great to add the check also for the new path so that Gitext can use WSL executables in both versions

### Environment

- Git Extensions 4.0.2.16100
- Git 2.39.1.windows.1
- .NET 6.0.15

RussKie:
Resolves #10025
Resolves #10815

## Test methodology 

- manual
- unit tests

## Merge strategy

----

### DIFF
--- a/GitCommands/PathUtil.cs
+++ b/GitCommands/PathUtil.cs
@@ -11,6 +11,7 @@ namespace GitCommands
 
         // Windows build 21354 supports wsl.localhost too, not supported for WSL Git
         private const string WslPrefix = @"\\wsl$\";
+        private const string WslLocalhostPrefix = @"\\wsl.localhost\";
 
         public static readonly char PosixDirectorySeparatorChar = '/';
         public static readonly char NativeDirectorySeparatorChar = Path.DirectorySeparatorChar;
@@ -146,6 +147,23 @@ namespace GitCommands
             }
         }
 
+        /// <summary>
+        ///  Replaces <c>\\wsl.localhost\</c> with <c>\\wsl$\</c>, if found. Else returns the <paramref name="path"/> untouched.
+        /// </summary>
+        /// <param name="path">The path to normalize.</param>
+        /// <returns>The WSL normalized path.</returns>
+        public static string NormalizeWslPath(this string path)
+        {
+            // NOTE: path is expected to be already normalized!
+
+            if (!IsWslLocalhostPrefixPath(path))
+            {
+                return path;
+            }
+
+            return path.Replace(WslLocalhostPrefix, WslPrefix, StringComparison.OrdinalIgnoreCase);
+        }
+
         public static string Resolve(string path, string relativePath = "")
         {
             if (string.IsNullOrWhiteSpace(path))
@@ -198,23 +216,26 @@ namespace GitCommands
         /// Check if the path is any known path for WSL that may require special handling
         /// </summary>
         /// <param name="path">Path to check</param>
-        /// <returns>true if a path is a known WSL path</returns>
-        public static bool IsWslPath(string path)
+        /// <returns><see langword="true"/> if a path is a known WSL path; otherwise, <see langword="false"/>.</returns>
+        public static bool IsWslPath(string? path)
         {
             return !string.IsNullOrWhiteSpace(path)
-                && (IsWslPrefixPath(path)
-                    || path.ToLower().StartsWith(@"\\wsl.localhost\"));
+                && (IsWslPrefixPath(path) || IsWslLocalhostPrefixPath(path));
         }
+
+        /// <summary>
+        ///  Check if the path starts with '\\wsl.localhost\'.
+        /// </summary>
+        /// <param name="path">Path to check</param>
+        /// <returns><see langword="true"/> if the path starts with '\\wsl.localhost\'; otherwise, <see langword="false"/>.</returns>
+        private static bool IsWslLocalhostPrefixPath(string path) => path.StartsWith(WslLocalhostPrefix, StringComparison.OrdinalIgnoreCase);
 
         /// <summary>
         /// Check if the path is has handled specially for WSL
         /// </summary>
         /// <param name="path">Path to check</param>
-        /// <returns>true if the path is a WSL path with internal handling</returns>
-        private static bool IsWslPrefixPath(string path)
-        {
-            return path.ToLower().StartsWith(WslPrefix);
-        }
+        /// <returns><see langword="true"/> if the path is a WSL path with internal handling; otherwise, <see langword="false"/>.</returns>
+        private static bool IsWslPrefixPath(string path) => path.StartsWith(WslPrefix, StringComparison.OrdinalIgnoreCase);
 
         /// <summary>
         /// Get the name of the distribution (like "Ubuntu-20.04") for WSL2 paths.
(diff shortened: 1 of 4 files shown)


GitCommands/PathUtil.cs
```diff
@@ -146,6 +147,23 @@ public static string NormalizePath(this string path)
...
         }
 
+        /// <summary>
+        ///  Repalces <c>\\wsl.localhost\</c> with <c>\\wsl$\</c>, if found. Else returns the <paramref name="path"/> untouched.
+        /// </summary>
+        /// <param name="path">The path the normalise.</param>
+        /// <returns>The WSL normalised path.</returns>
+        public static string NormalizeWslPath(this string path)
+        {
+            // NOTE: path is expected to be already initialised!
```
mstv, line 157:
```suggestion
        /// <summary>
        ///  Replaces <c>\\wsl.localhost\</c> with <c>\\wsl$\</c>, if found. Else returns the <paramref name="path"/> untouched.
        /// </summary>
        /// <param name="path">The path to normalize.</param>
        /// <returns>The WSL normalized path.</returns>
        public static string NormalizeWslPath(this string path)
        {
            // NOTE: Path is expected to be already normalized!
```

vbjay:
Should this be an exe.config thing?  Allow for the matches to be defined and the replacement.

mstv (approved):
just doc typos, have not run

gerhardol (approved):
Have run very briefly. 
I believe this is all that is needed, took a quick look at the implementation.

This will display recent (and current) repos opened as wsl.localhost as wsl$. From a usability view, that is better as the recents are not duplicated. Users may be confused though.
(It is also much simpler to implement it this way.)

I will update the doc.

This feature was something I considered to, as `wslpath` returns wsl.locahost and then you get the extra item.
I decided to not delay 4.1 for this, but have no problem accepting changes from someone else.

RussKie:
> Should this be an exe.config thing? Allow for the matches to be defined and the replacement.

I'm not sure what you mean, sorry.

RussKie:
> Users may be confused though.

Perhaps, but looks like users get confused by "wsl.localhost" more.

GitExtensions/Program.cs
```diff
@@ -34,12 +34,18 @@ private static void Main()
...
             if (ThisAssembly.Git.IsDirty)
             {
-                // In non official builds force to fail for cross-thread operations so we can fix those.
-                Control.CheckForIllegalCrossThreadCalls = true;
+                checkForIllegalCrossThreadCalls = true;
             }
 
+            // In non official builds force to fail for cross-thread operations so we can fix those.
+            Control.CheckForIllegalCrossThreadCalls = checkForIllegalCrossThreadCalls;
+
```
mstv, line 48: unrelated
If intentionally added, then rather shorter:
```
#if !DEBUG
            if (ThisAssembly.Git.IsDirty)
#endif
            {
                // In non official builds force to fail for cross-thread operations so we can fix those.
                Control.CheckForIllegalCrossThreadCalls = true;
            }
```

vbjay:
> > Should this be an exe.config thing? Allow for the matches to be defined and the replacement.
> 
> I'm not sure what you mean, sorry. 

If wsl decides to change aliases or if user can change it to something else.  Thinking if they needs to fix we can at least tell them to change the config file to allow it to work till we make a release.

RussKie:
> > > Should this be an exe.config thing? Allow for the matches to be defined and the replacement.
> > 
> > 
> > I'm not sure what you mean, sorry.
> 
> If wsl decides to change aliases or if user can change it to something else. Thinking if they needs to fix we can at least tell them to change the config file to allow it to work till we make a release.

These are Windows defined, so I don't expect these to change often.

gerhardol:
>>    Users may be confused though.

> Perhaps, but looks like users get confused by "wsl.localhost" more.

There will be issues with users claiming that wsl.localhost is the new name and that it should be used instead of the old name, or that the entered name is used.
I prefer this solution, just wanted to point that out. i prefer this way:
* A solution where the entered name is used will be more complex and may get two entries in the recent list.
* We could translate to wsl.localhost instead, for users with wsl 1.0 or later. That will require a check for older version (either translate differently or use Windows Git). wsl$ is a shorter prefix too.

> These are Windows defined, so I don't expect these to change often.

wsl$ was changed just because it contains illegal characters, requiring special handling. I would not expect this to be changed again.
There are some special handling to wsl handling, like file notification is not working. So wsl is a special handling.

*Nix users (with wine or so) may need a similar "mode, triggered on something else.

andrej-schaefer:
I see a problem with this solution for the following use case:
Use WSL for DEV and runtime, for this purpose code is in the WSL file system
Use wsl.localhost for gitExtensions due to use putty and configure different ssh keys per project / customer with a passphrase

With the new solution it change to wsl and does not take the configured ssh in gitExtensions any more but redirects to default ssh in WSL

Workaround is to downgrade of gitExtensions or move repos to windows
Easy solution to handle multiple ssh files on linux i havent found so far, maybe someone have a solution for this problem
Propose is complex, maybe select in the repository selection or some kind of configuration
Hope this comment is ok, if not, please let me know :)

WSL: WSL 2 with latet ubuntu
Windows 11

gerhardol:
> Git repos mapped to a drive letter will not use the special WSL handling but Windows Git

From https://git-extensions-documentation.readthedocs.io/en/release-4.1/settings.html#wsl-git-notes
(will modify the formatting)

There are too many settings in GE already, I would like to avoid more (there are some hidden settings for WSL, not related to this though).


Originally, I wanted to have wsl.localhost separated as you describe, but Windows insists on wsl.localhost in most conversions so it it inconvenient...

PR #10918, opened 2023-04-26
Do not delay Dashboard start for invalid Git repos

gerhardol:
There has been related issues before, that has been addressed somehow.
I saw this issue after adding some repos to WSL and an USB stick,
Dashboard would require about 1:30 to open.

This could be included in 4.1, but I do not want to delay the release.

## Proposed changes

Add the repos to the lists without any I/O to check if the repo is invalid or not (to set the icon).
Perform the check if a repo has a valid Git dir async, together with branch name retrieval and set the invalid icon then instead.

Rendering of branch name failed if the item was rendered before the branch name was set.
This had a much higher probability to occur after the item creation was changed.

Menu item to remove invalid repos was only set from favorite repos, as _hasInvalidRepos was overwritten.

## Test methodology 

Manual.

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
+++ b/GitUI/CommandsDialogs/BrowseDialog/DashboardControl/UserRepositoriesList.cs
@@ -315,10 +315,6 @@ namespace GitUI.CommandsDialogs.BrowseDialog.DashboardControl
 
             void BindRepositories(IReadOnlyList<RecentRepoInfo> repos, bool isFavourite)
             {
-                var repoValidityArray = repos.AsParallel().Select(r => !_controller.IsValidGitWorkingDir(r.Repo.Path)).ToArray();
-
-                _hasInvalidRepos = repoValidityArray.Any();
-
                 for (var index = 0; index < repos.Count; index++)
                 {
                     RecentRepoInfo recent = repos[index];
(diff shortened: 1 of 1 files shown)


RussKie:
I observed a significant dashboard slowness, and I almost raised an issue for it at one point but I was unable to replicate the issue in debug builds. I've taken few memdumps, and looking at those I think your suggested fix may be right on target:
image

I'm going to take your change for a spin before I provide more feedback.

RussKie (approved):
🚀 
Thank you

gerhardol:
Then I will rebase this to 4.1 tonight.

Some notes I should have added to the description:

There may be ways to improve the detection time as well but this should be the big change. It should primarily affect the slow repos (start wsl?), Showing the status for other. (Wsl branch names are occasionally missing.)

We are probably creating two GitModule in the async code. Some Ms per repo added.

mstv (approved):
Looks sensible, have not run

gerhardol:
> There may be ways to improve the detection time as well but this should be the big change. It should primarily affect the slow repos (start wsl?), Showing the status for other. (Wsl branch names are occasionally missing.)

I do not find an obvious way to check the path for network including wsl any faster.

> We are probably creating two GitModule in the async code. Some Ms per repo added.

No, this is fine, only one module.

PR #10923, opened 2023-04-28
Remove DotnetRuntimeBootstrapper

### LINKED ISSUE #10425: Final gitextensions assets are not bootstrapped
### Environment

This is an issue related to the build process, not the application itself

### Issue description

I was prompted to investigate this after this discussion: https://github.com/Tyrrrz/DotnetRuntimeBootstrapper/discussions/36

I see that GitExtensions uses DotnetRuntimeBootstrapper, but the final build (i.e. `GitExtensions.exe`) downloaded from the latest release **is not** bootstrapped. The exe file is the standard dotnet apphost for Windows applications.

You can verify that quite easily by checking the file details. If the application is bootstrapped, the `OriginalFilename` version info manifest attribute should point to itself, instead of the assembly file (i.e. `.dll`).

Example of the correctly bootstrapped build (built from source with some tweaking):

image

Example of the incorrectly bootstrapped build (taken from latest releases):

image

I was unable to identify the exact issue because running `dotnet publish` fails for me with this:

image

I was, however, able to produce a bootstrapped build using `dotnet build` by setting [`<GenerateBootstrapperOnBuild>true</GenerateBootstrapperOnBuild>`](https://github.com/Tyrrrz/DotnetRuntimeBootstrapper#generate-bootstrapper-on-build) in the csproj.

By scouting the project files, my guess is that the bootstrapper target is either not executed or executed too late. For example, in `GitExtensions/Project.Publish.targets` you have this target:

```xml
<!--
                       CreatePortable

    Creates a portable archive.
    -->
  <!-- Any errors in targets that executed as 'AfterTargets' don't break the build: https://github.com/microsoft/msbuild/issues/3345
       A fix is going out in VS16.6p3, but it is way too long for us to wait.
    -->
  <Target Name="CreatePortable"
          AfterTargets="Publish"
          DependsOnTargets="_RetrieveCurrentBuildVersion;_PublishTranslations;_DownloadPluginManager;_EnsureBundleContent;_CleanupBeforePack;_PatchRuntimeConfigJson">
    <PropertyGroup>
      <_TargetAppConfig>@(AppConfigFileDestination)</_TargetAppConfig>

      <!-- Resolve app.config, so we can set/unset "portable" flag -->
      <_PublishAppConfig>$([System.IO.Path]::GetFileName('$(_TargetAppConfig)'))</_PublishAppConfig>
      <_PublishAppConfigPath>$([System.IO.Path]::Combine('$(PublishDir)', '$(_PublishAppConfig)'))</_PublishAppConfigPath>

      <!-- Resolve the output file -->
      <_PublishPortableVersionSuffix>-$(CurrentBuildVersion.ToString())</_PublishPortableVersionSuffix>
      <_PublishPortableCommitHashSuffix Condition="'$(GitCommit)' != ''">-$(GitCommit)</_PublishPortableCommitHashSuffix>
      <_PublishPortableCommitHashSuffix Condition="'$(env:APPVEYOR_REPO_COMMIT)' != ''">-$(env:APPVEYOR_REPO_COMMIT)</_PublishPortableCommitHashSuffix>
      <_PublishPortableCommitHashSuffix Condition="'$(env:APPVEYOR_PULL_REQUEST_HEAD_COMMIT)' != ''">-$(env:APPVEYOR_PULL_REQUEST_HEAD_COMMIT)</_PublishPortableCommitHashSuffix>
      <_PublishPortableFileName>GitExtensions-Portable$(_PublishPortableVersionSuffix)$(_PublishPortableCommitHashSuffix).zip</_PublishPortableFileName>
      <_PublishPortablePath>$([MSBuild]::NormalizePath('$(ArtifactsPublishDir)', '$(_PublishPortableFileName)'))</_PublishPortablePath>

      <!-- We want to archive the whole publish folder, so get one level up -->
      <_PublishedPath>$([MSBuild]::NormalizeDirectory('$(PublishDir)'))</_PublishedPath>
    </PropertyGroup>

    <!-- Mark the package as "portable" -->
    <XmlPoke XmlInputPath="$(_PublishAppConfigPath)"
            Query="configuration/applicationSettings/GitCommands.Properties.Settings/setting[@name='IsPortable']/value" 
            Value="True"
            ContinueOnError="ErrorAndStop"
          />

    <ZipDirectory
            SourceDirectory="$(_PublishedPath)"
            DestinationFile="$(_PublishPortablePath)"
            Overwrite="true"
            ContinueOnError="ErrorAndStop"
          />

    <!-- Reset the "portable" flag -->
    <XmlPoke XmlInputPath="$(_PublishAppConfigPath)"
            Query="configuration/applicationSettings/GitCommands.Properties.Settings/setting[@name='IsPortable']/value" 
            Value="False"
            ContinueOnError="ErrorAndStop"
          />
  </Target>
```

From my understanding, this is what effectively generates the output assets. However, this target is set to run after the `Publish` target, and it doesn't have the `CreateBootstrapperAfterPublish` target in the list of dependencies. My guess is that the bootstrapper target ends up appearing later in order, which means the bootstrapped apphost never gets picked up by `ZipDirectory`.

This is likely the culprit behind all of these issues:

https://github.com/gitextensions/gitextensions/issues/10337
https://github.com/gitextensions/gitextensions/issues/10319
https://github.com/gitextensions/gitextensions/issues/10317

Note that the "You must install .NET to run this application" popup **does not** come from DotnetRuntimeBootstrapper. It comes from the default .NET apphost. In other words, this popup is produced by a **non-bootstrapped** build:

image

DotnetRuntimeBootstrapper's apphost should look like this instead (taken from my tweaked build):

image

### Steps to reproduce

- `cd GitExtensions`
- `dotnet publish -c Release --verbosity normal`
- See that the `CreateBootstrapperAfterPublish` target never runs
- See that the final build is not bootstrapped

_No response_

### Diagnostics

_No response_

RussKie:
We have no use for it.

Closes #10425.

### DIFF
--- a/GitExtensions/GitExtensions.csproj
+++ b/GitExtensions/GitExtensions.csproj
@@ -21,10 +21,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="DotnetRuntimeBootstrapper">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
     <PackageReference Include="GitInfo">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
--- a/Packages.props
+++ b/Packages.props
@@ -7,7 +7,6 @@
     <PackageReference Update="AppInsights.WindowsDesktop" Version="2.18.1" />
     <PackageReference Update="Ben.Demystifier" Version="0.4.1" />
     <PackageReference Update="ConEmu.Core" Version="22.12.18" />
-    <PackageReference Update="DotnetRuntimeBootstrapper" Version="2.3.1" />
     <PackageReference Update="EnvDTE" Version="17.0.32112.339" />
     <PackageReference Update="ExCSS" Version="4.1.3" />
     <PackageReference Update="GitInfo" Version="2.2.0" />


gerhardol:
The standard check is not working too well.
What is the problem with the bootstrapper, is it removed in Release builds?

gerhardol:
> > What is the problem with the bootstrapper, is it removed in Release builds?
> 
> It's not being used, so it's an overhead. 

Why is it not used?
The message was better than .net default

RussKie:
> Also the issue shows that we can make it work with making the changes to the project. If we adjust it to make that target run on portable. I'd compare msi vs portable build output. I'll bet msi is bootstrapped and portable not. So instead of removing just add the right msbuild target in the mix of dependencies.

All of this is extra work, which isn't going to bring substantial change to what we already have.

PR #10924, opened 2023-04-28
Remove SettingsCache entry when setting empty value

### LINKED ISSUE #10917: Git config setting "core.autocrlf" is not updated after selecting source "Effective" again
### Environment

- Git Extensions v4.0.2 / 33.33.33
- Build be087f852f799476badc2220bd86c64ad1c74c0b
- Git 2.40.0.windows.1
- .NET 6.0.16
- DPI 96dpi (no scaling)

```
```

### Issue description

Git config setting "core.autocrlf" is not updated on switching settings sources since selecting source "Effective" again

grafik

### Steps to reproduce

- configure "core.autocrlf" globally different from "Not set"
- apply
- configure "core.autocrlf" to "Not set" for local repo
- apply
- reopen Settings
- click "Local for current repository"
- click "Effective" -> wrong
- choose other sources -> no update

not checked

### Diagnostics

```
$ git config --show-origin core.autocrlf
file:%HOME%/.gitconfig    false
```

mstv:
Fixes #10917

## Proposed changes

- Remove `SettingsCache` entry when setting empty or `null` value 

## Screenshots 

N/A

## Test methodology 

- added unit test

## Merge strategy

----

### DIFF
--- a/GitCommands/Settings/SettingsCache.cs
+++ b/GitCommands/Settings/SettingsCache.cs
@@ -130,7 +130,14 @@ namespace GitCommands
 
                 SettingsChanged();
 
-                _byNameMap.AddOrUpdate(name, value, (key, oldValue) => value);
+                if (string.IsNullOrEmpty(value))
+                {
+                    _byNameMap.TryRemove(name, out _);
+                }
+                else
+                {
+                    _byNameMap.AddOrUpdate(name, value, (key, oldValue) => value);
+                }
             });
         }
 
--- a/UnitTests/GitCommands.Tests/Settings/ConfigSectionTests.cs
+++ b/UnitTests/GitCommands.Tests/Settings/ConfigSectionTests.cs
@@ -1,0 +1,48 @@
+﻿using FluentAssertions;
+using GitCommands.Config;
+using NUnit.Framework;
+
+namespace GitCommandsTests.Settings
+{
+    [TestFixture]
+    internal sealed class ConfigSectionTests
+    {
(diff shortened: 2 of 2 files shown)


gerhardol:
This blocks the possibility to override global settings in local configs.
I have used this in Build Server, but do not recall the use case right now.

Maybe a good move, but I do not want to include right now...

RussKie:
> but do not recall the use case right now.

I don't know this area well enough, and generally use the default settings. 

It sounds to me we know that there may be situations where retaining null/empty could be useful but we don't know what these situations are.
In this case my general position is - let's do a scream test - i.e., make a change and wait for users to scream. This way we'll learn the use cases and may be able to work on a solution. 
Affected users will be able to use an older version in mean time.

RussKie (approved):

gerhardol:
I rather do the scream test in  master first

mstv:
I am not afraid if _this_ screaming test because I have just aligned the *cache* with the actual change to the settings file. The result in the settings file is as before. But GE is not confused by the wrong cache any longer.

> This is caused by `SetNullableEnum<T>()` saving `null` as an empty string in the `ConfigFileSettingsCache` instead of removing the entry (**as it is correctly done by `ConfigSection.SetValue()`**).

RussKie:
In the master we won't know until we release the next version. I'd rather
take it now and be prepared to fix it quick (or not :)).

gerhardol (commented):
+1

PR #10944, opened 2023-05-04
ShowCurrentBranchOnly: Filter also if no branch checkout

### LINKED ISSUE #10939: "Current branch only" doesn't work for headless checkout
### Environment

- Git Extensions 4.1.0.16666
- Build dd6b150c547f815e8e0e9d74f252d2cf7416e3d5 (Dirty)
- Git 2.40.1.windows.1
- .NET 6.0.16
- DPI 96dpi (no scaling)
- Portable: False

```
    Microsoft.WindowsDesktop.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 8.0.0-preview.2.23128.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

https://user-images.githubusercontent.com/4403806/236095521-3a7b29b7-2acf-4355-99c3-d7a209dba0ed.mp4

Now we're always running:
```
git -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --max-count=100000 --exclude=refs/notes/commits --all --
```

### Steps to reproduce

1. Checkout headless (e.g., use "Checkout this commit only" from the context menu)
2. Try to filter to the current branch only.

Yes, it works in 4.0.x version:
image

```
git -c log.showSignature=false log -z --pretty=format:"%H%T%P%n%at%n%ct%n%aN%n%aE%n%cN%n%cE%n%B" --max-count=100000 --glob=refs/stas[h] HEAD --
```

### Diagnostics

_No response_

gerhardol:
Fixes #10939

I rather not merge this and close #10939 as not planned

## Proposed changes

If no branch checkout, the branch filter was handling ShowCurrentBranchOnly as ALBranches.
This is changed to work similar to before, where ShowCurrentBranchOnly filtered on the current commit instead.
A consequence is that the first stash is not added to the grid if there is no branch checked out.

first stash could be visible in some situations if current commit was added instead (rather than HEAD that may not exist), but there will still be some checks needed and this is fgar too complicated as it is.

## Test methodology 

The 'parallel implementation' tests were updated,

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
@@ -936,6 +936,8 @@ namespace GitUI
                 Lazy<ObjectId?> currentCheckout = new(() =>
                     headRef.Value?.ObjectId ?? capturedModule.GetCurrentCheckout());
 
+                ObjectId? previousCheckout = CurrentCheckout;
+
                 // Evaluate GitRefs and current commit
                 ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
@@ -944,12 +946,11 @@ namespace GitUI
 
                     // If the current checkout (HEAD) is changed, don't get the currently selected rows,
                     // select the new current checkout instead.
-                    if (currentCheckout.Value != CurrentCheckout && currentCheckout.Value is not null)
-                    {
-                        currentlySelectedObjectIds = new List<ObjectId> { currentCheckout.Value };
-                    }
-
                     CurrentCheckout = currentCheckout.Value;
+                    if (CurrentCheckout != previousCheckout && CurrentCheckout is not null)
+                    {
+                        currentlySelectedObjectIds = new List<ObjectId> { CurrentCheckout };
+                    }
 
                     // Exclude the 'stash' ref, it is specially handled when stashes are shown
                     refsByObjectId = (AppSettings.ShowStashes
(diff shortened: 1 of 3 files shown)


RussKie:
> I rather not merge this and close #10939 as not planned

I completely disagree. It breaks a major use case for me, making the new version unusable.

RussKie:
I tested your change locally, and it's fixing https://github.com/gitextensions/gitextensions/issues/10939. Thank you.

I think it's important we add integration tests for these scenarios - e.g., show all, show current (with and without branch), filtered, show stashes in all previous cases, etc., so that we know which scenario work or don't. I'll try to do so over the weekend.

gerhardol:
@mstv What about this change? (First stash are not shown in the grid if no branch is checked out).

mstv:
As I rarely use any kind of branch filtering, I do not have a strong opinion about side effects of this feature.
Anyway: If the display of stashes is enabled, I would not filter them. Amongst others I use stashes in order to move / share changes between worktrees.
Of course, I expect that all stashes are shown if showing "All branches" - regardless of whether a branch is checked out or not.
(Why should I create a local branch if I just want to build it and do not plan to commit to it?)

gerhardol:
I updated the change to use the current checkout instead, covers more cases and slightly simpler code.
(Had to make a second update to handle empty repos.)
With this, I am OK with the change, complexity is as before.

RussKie (approved):

RussKie:
Works well, thank you!


* All branches:
image

* Current branch only (branch):
image

* Current branch only (headless):
image
image

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
```diff
@@ -27,6 +27,7 @@
 using Microsoft.VisualBasic;
 using Microsoft.VisualStudio.Threading;
 using ResourceManager;
+using static System.Windows.Forms.VisualStyles.VisualStyleElement;
```
mstv, line 30: Can be removed as well as `using Microsoft.VisualBasic;`
Intellisense has become kind of hyperactive adding `using` statements.

mstv (approved):
LGTM

GitUI/UserControls/RevisionGrid/RevisionGridControl.cs
RussKie: 🤦‍♂️

gerhardol:
Made the pr and the rework a little too late in the night and missed the using.

The problem with testing is Git behavior: some commands raise errors if there are no commits or stashes and some combinations are hiding wanted info. Tests for that will be hard, i do not know if it is worth it.

Issue #10914, opened 2023-04-25
System git config file is not evaluated


mstv:
### Environment

- Git Extensions v4.0.2 / 33.33.33
- Build be087f852f799476badc2220bd86c64ad1c74c0b
- Git 2.40.0.windows.1
- Microsoft Windows NT 10.0.19045.0
- .NET 6.0.16
- DPI 96dpi (no scaling)
- Microsoft.WindowsDesktop.App Versions

```
    Microsoft.WindowsDesktop.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

GE asserts that the git config "core.autocrlf" was "not set" although it *is* set in the system config file "C:/Program Files/Git/etc/gitconfig".

![grafik](https://user-images.githubusercontent.com/36601201/234284646-0cf4ed67-6310-48df-8eae-e4596426d59f.png)

### Steps to reproduce

install git and GE on a new machine

### Did this work in previous version of GitExtensions?

no

### Diagnostics

```
$ git config --show-origin core.autocrlf
file:C:/Program Files/Git/etc/gitconfig true

$ cat 'C:/Program Files/Git/etc/gitconfig'
[http]
        sslCAInfo = C:/Program Files/Git/mingw64/etc/ssl/certs/ca-bundle.crt
        sslBackend = openssl
[diff "astextplain"]
        textconv = astextplain
[filter "lfs"]
        clean = git-lfs clean -- %f
        smudge = git-lfs smudge -- %f
        process = git-lfs filter-process
        required = true
[credential]
        helper = manager
[core]
        editor = \"C:\\\\Program Files\\\\Notepad++\\\\notepad++.exe\" -multiInst -notabbar -nosession -noPlugin
        autocrlf = true
        fscache = true
        symlinks = false
[pull]
        rebase = false
[credential "https://dev.azure.com"]
        useHttpPath = true
[init]
        defaultBranch = master
```

mstv:
Should we add a (read-only) settings source "System"?

RussKie:
Why read-only though? A user can change the system wide settings at will

gerhardol:
System should be shown

> Why read-only though? A user can change the system wide settings at will
> 

Only as admin

mstv:
Read-only because C:/**Program Files**/Git/etc/gitconfig would require raised privileges.

RussKie:
That's a good point. Though, if a user isn't a local admin then we won't be
[...]

PR #10948, opened 2023-05-07
Fixup path to system git config

mstv:
Relates to #10914

## Proposed changes

- `ConfigFileSettings.CreateSystemWide()`: use `AppSettings.GitCommand` instead of misleading `AppSettings.GitBinDir` 
- Rename `AppSettings.LinuxToolsDir` from `GitBinDir`
- Rename `GitSettingsPage.LinuxToolsDir` from `GitBinPath`

## Screenshots 

N/A

## Test methodology 

- manual

##  Please do not squash merge!

----

### DIFF
--- a/GitCommands/Settings/AppSettings.cs
+++ b/GitCommands/Settings/AppSettings.cs
@@ -696,7 +696,7 @@ namespace GitCommands
             if (!string.IsNullOrEmpty(ssh))
             {
                 // OpenSSH uses empty path, compatibility with path set in 3.4
-                string path = new SshPathLocator().GetSshFromGitDir(GitBinDir);
+                string path = new SshPathLocator().GetSshFromGitDir(LinuxToolsDir);
                 if (path == ssh)
                 {
                     AppSettings.SshPath = "";
@@ -1313,22 +1313,15 @@ namespace GitCommands
             set => SetBool("ShowCommitBodyInRevisionGrid", value);
         }
 
-        /// <summary>Gets or sets the path to the git application executable.</summary>
-        public static string GitBinDir
+        /// <summary>Gets or sets the path to the GNU/Linux tools (bash, ps, sh, ssh, etc.), e.g. "C:\Program Files\Git\usr\bin\"</summary>
+        public static string LinuxToolsDir
         {
-            get => GetString("gitbindir", "");
+            // Migrate the setting value from the from the former "gitbindir"
+            get => GetString(nameof(LinuxToolsDir), defaultValue: null) ?? GetString("gitbindir", "");
             set
             {
-                var temp = value.EnsureTrailingPathSeparator();
-                SetString("gitbindir", temp);
-
-                ////if (string.IsNullOrEmpty(_gitBinDir))
-                ////   return;
-                ////
-                ////var path =
-                ////   Environment.GetEnvironmentVariable("path", EnvironmentVariableTarget.Process) + ";" +
-                ////   _gitBinDir;
-                ////Environment.SetEnvironmentVariable("path", path, EnvironmentVariableTarget.Process);
+                string linuxToolsDir = value.EnsureTrailingPathSeparator();
+                SetString(nameof(LinuxToolsDir), linuxToolsDir);
             }
         }
 
--- a/GitCommands/Settings/ConfigFileSettings.cs
+++ b/GitCommands/Settings/ConfigFileSettings.cs
@@ -56,7 +56,7 @@ namespace GitCommands.Settings
             if (!File.Exists(configPath))
             {
                 // Git 1.xx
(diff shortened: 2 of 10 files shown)


GitCommands/Settings/AppSettings.cs
```diff
@@ -1313,22 +1313,19 @@ public static bool ShowCommitBodyInRevisionGrid
...
-                ////   Environment.GetEnvironmentVariable("path", EnvironmentVariableTarget.Process) + ";" +
-                ////   _gitBinDir;
-                ////Environment.SetEnvironmentVariable("path", path, EnvironmentVariableTarget.Process);
+                string gnuToolsDir = value;
+                if (!string.IsNullOrWhiteSpace(gnuToolsDir))
+                {
+                    gnuToolsDir = gnuToolsDir.EnsureTrailingPathSeparator();
+                }
+
+                SetString(nameof(GnuToolsDir), gnuToolsDir);
```
gerhardol, line 1328: Why not keep the existing name?
This will require users to configure this again

gerhardol (commented):
+1

GitCommands/Settings/AppSettings.cs
```diff
@@ -1313,22 +1313,19 @@ public static bool ShowCommitBodyInRevisionGrid
             set => SetBool("ShowCommitBodyInRevisionGrid", value);
         }
 
-        /// <summary>Gets or sets the path to the git application executable.</summary>
-        public static string GitBinDir
+        /// <summary>Gets or sets the path to the GNU tools.</summary>
+        public static string GnuToolsDir
```
RussKie, line 1317: The original name was meaningful to me, and the new one is not... What's "GNU tool"? Is this something an average Joe developer, who happen to use our app, would understand?

GitCommands/Settings/AppSettings.cs
RussKie: > This will require users to configure this again

I think this takes care of it:
image

Though we should add a comment explaining this takes care of backwards compatibility.

GitCommands/Settings/ConfigFileSettings.cs
```diff
@@ -56,7 +56,7 @@ public static ConfigFileSettings CreateGlobal(ConfigFileSettings? lowerPriority,
             if (!File.Exists(configPath))
             {
                 // Git 1.xx
-                configPath = Path.Combine(AppSettings.GitBinDir, "..", "etc", "gitconfig");
+                configPath = Path.Combine(AppSettings.GitCommand, "..", "..", "etc", "gitconfig");
```
RussKie, line 59: Doesn't `AppSettings.GitCommand` contain the path to git.exe?

mstv: > Doesn't `AppSettings.GitCommand` contain the path to git.exe?

Yes, it does. This setting is correctly named (in contrast to `GitBinDir`). The global settings are located in the Git installation folder.
This is the actual bugfix (done in a separate commit, not to be squashed).

GitCommands/Settings/AppSettings.cs
mstv: "GitBinDir" is totally misleading to me. It does not point to the folder with `git.exe` but to the GNU tool**s** which come bundled with Git.
I am aware of https://en.wikipedia.org/wiki/GNU/Linux_naming_controversy. Since this setting is about the tools and not the OS, "GNU" would be more correct.
I want to have "Tools" in this setting's name. From the point of clarity for the average Joe developer, "LinuxToolsDir" might be the compromise (and is already used in FormSettings).

image

gerhardol: GnuTools is fine for me, but it is maybe better to align with the GUI name - or change the GUI name.

RussKie: Maybe expand the xml-doc to clarify that the purpose (e.g., git/usr/bin)?
And, yes, please update the UI label too.

RussKie (approved):
👍 modulo the open discussion about UI label.

GitCommands/Settings/AppSettings.cs
mstv: Many translated texts and local identifiers used "Linux Tools". So I used this term, and added comments.

RussKie (approved):

gerhardol (approved):

mstv:
> pushed a commit that referenced this pull request

cherry-picked to the release branch

I think there is nothing experimental on the master branch - a few other fixes though. So we should merge master to release/4.0 right before releasing 4.1.1, shouldn't we?

RussKie:
Then why not just release 4.2? :)

PR #10466, opened 2022-11-28
AssemblyName include GitExtensions prefix

gerhardol:
for BuildIntegration, that are handled as plugins.
A few JiraCommitHint files are ignored as the plugin is not activated since .net upgrade.

Fixes regression in #10456

### Before

Only None in list

### After

image

## Test methodology 

Manual in debugger, built standalone package

## Merge strategy

I agree that the maintainer squash merge this PR (if the commit message is clear).

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10953, opened 2023-05-10
Rename build server plugins

RussKie:
v4.1 deviates from v4.0 - the build server plugins didn't have "GitExtensions." prefix in v4.0. That failed the signing project:
    image

At the same time, this is a good time to fix the plugin names and align those with the rest of the plugins, which start with "GitExtensions.Plugins." prefix.

I tested this locally, and upgrade from 4.0.2 to 4.1 seemed to work removing the prefix-less files and replacing those with the new files. Though, an additional validation is required once we have proper signed build artifacts.


gerhardol:
Follow up to #10466

mstv (commented):
+1 looks sensible, have not run

gerhardol:
> Also remove workaround in [Plugins/Directory.Build.props](https://github.com/gitextensions/gitextensions/pull/10466/files#diff-a68d2dfca3b5da08bacf721a04172f90a52cc624dca39e67aadef43de78fc6ac)

Still needed but comment on line 17 should be removed

> Maybe Plugins/GitUIPluginInterfaces/ManagedExtensibility.cs CI fix too

CIBUILD ref to be removed, in 

Cannot build Release in MSVS (command line OK).
Could build Debug if GitCommands were built manually

Build started...
1>------ Skipped Build: Project: GitUIPluginInterfaces ------
1> 
2>------ Skipped Build: Project: GitCommands ------
2> 
3>------ Build started: Project: ResourceManager, Configuration: Release Any CPU ------
3>CSC : error CS0006: Metadata file 'C:\dev\gc\gitextensions_4\artifacts\Release\obj\GitCommands\net6.0-windows\ref\GitCommands.dll' could not be found
3>CSC : error CS0006: Metadata file 'C:\dev\gc\gitextensions_4\artifacts\Release\obj\GitUIPluginInterfaces\net6.0-windows\ref\GitUIPluginInterfaces.dll' could not be found
3>Done building project "ResourceManager.csproj" -- FAILED.

RussKie:
> > Also remove workaround in [Plugins/Directory.Build.props](https://github.com/gitextensions/gitextensions/pull/10466/files#diff-a68d2dfca3b5da08bacf721a04172f90a52cc624dca39e67aadef43de78fc6ac)
> 
> Still needed but comment on line 17 should be removed

I rather think it could stay though should be updated to the full prefix. Also the MEF descoverer should be updated to use the full prefix too. But I'm planning to do this work on the master branch, this isn't important for the release.

PR #10956, opened 2023-05-10
Remove reset and delete from commit working dir menu

gerhardol:
## Proposed changes

Commands are available in the context menu and buttons, where the user is expected to look for them.
The commands were not working correctly for renamed and conflicted files.

- For selected files (delete/rest) the context menu is the natural way to reset
- For Reset all/unstaged, the buttons make the same actions, and the visible/enable is set correctly

A number of similar issues have been reported in 4.0 (after Git errors were no longer ignored).
There are no report for this problem what I see, but some reported issues will build on this (remove a function).

This affects translations (removing strings), but I would like to include it in a 4.1.1 or so.

## Screenshots 

Buttons unchanged
image

### Before

image

### After

image

## Test methodology 

manual

## Merge strategy

----


mstv:
If the buttons are missing, they can be enabled here:

image

PR #10968, opened 2023-05-16
Keep the splitter collapsed state if not stored

### LINKED ISSUE #10959: Blank window area in repo browser
### Environment

- Git Extensions 4.1.0.16698
- Build 1fe52a137acef1fbd2b653dea263b44ad5844177
- Git 2.40.1.windows.1
- .NET 6.0.16
- DPI 96dpi (no scaling)
- Portable: False

```
    Microsoft.WindowsDesktop.App 6.0.14 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 7.0.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
    Microsoft.WindowsDesktop.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Issue description

I upgraded from 4.0 to 4.1 this morning, and now I've got a new blank area on the right side of the repo browser:

Screenshot_1

### Steps to reproduce

Open the repository browser.

Yes, worked in all previous versions.

### Diagnostics

_No response_

mstv:
Fixes #10959

## Proposed changes

- Use the current splitter collapsed state as default value of the setting
- Improve test `SplitterPositionsShouldBeRestored`
  - Enable standalone execution (fixes the main issue of #10986)
  - Make test for `LeftPanelWidth` work
  - Turn `WaitForRevisionsToBeLoaded` into `WaitForRevisionsToBeLoadedAsync` and add `JoinPendingOperationsAsync`
  - Remove unused function `ProcessUntil`
- Disable test `SplitterPositionsShouldBeRestored` for now because of #11017

## Screenshots 

see issues

## Test methodology 

- manual
- run integration testcase locally

## Please do not squash merge

----


gerhardol (approved):
Removed _Panel1Collapsed from settings, no extra panel after starting up.

RussKie:
I had a fresh look at SplitterPersistenceTests, and to be honest I had a wtf moment.... I mean we're testing splitter persistence, yet we aren't using `SplitterManager` anywhere in these tests. Are we doing a sane thing here? Maybe we should reconsider how we testing the splitter persistence/layout?

Splitter positions are managed by `SplitterManager` class, which is "owned" by `FormBrowse`. Shouldn't we (re-)write tests in a way where we'd be interacting with `SplitterManager` directly, thus, controlling values that we persist/restore?

E.g., few tests like these:
1. Verify default positions without any config - i.e., no config, open the form, and check the positions are as expected (some const values).
2. Verify positions are persisted correctly - i.e., open the form, move splitters or change layout, close the form and check `SplitterManager` mock written the right values.
3. Verify positions are loaded correctly - similar to the above, but then re-open the form and check the layout is as expected.

RussKie:
I spent some time tinkering with the tests, and one thing that we may be overlooking but which has a profound effect on the tests' determinism is `WindowPositionManager` - windows' positions are persisted in a completely different file, so whenever a test form is maximised it gets persisted on the test host and it doesn't get reset after the test completes.

mstv:
> Shouldn't we (re-)write tests in a way where we'd be interacting with SplitterManager directly

Since this is an integration test (and because SplitterManager is rather simple), I think there is no need for lower level tests.
Unfortunately, this integration test is necessary, because we manually need to compensate the borders from #8849 and #8557.
I had already forced FormBrowse both to a fixed size and to maximized state for this testcase. It did not help.

The issue is that the form loading behaves differently in tests run by AppVeyer - most likely because the VM can execute "only" two threads simultaneously.

gerhardol (approved):

RussKie:
I'm looking at ways to rewrite the test. I needed Verify changes in.

mstv:
> I'm looking at ways to rewrite the test. I needed Verify changes in.

Nice!

In this case, the failing integration test indicates that there is some kind of race condition anyway. I.e. unit tests do not suffice here. Furthermore, the restore of the FileTree splitter does not work on real PCs often (although the integration testcases never failed locally).

RussKie:
I spent some time going through the code... boy, do I have a headache now :( The whole thing is pretty horrible, we have multiple hard dependencies that are difficult to substitute as well as settings coming from different sources. E.g., the geometry of a dialog is read from WindowsPositions.xml, which is controlled by `WindowPositionManager`. The layout metrics are stored in `ISettingsSource`, which can be anything.... However, some layout settings are stored in `AppSettings`, which is backed by a file.

I provided a way to mock both `WindowPositionManager` and `ISettingsSource`, which make it possible to mock geometry and layout per-test. The appsetings file has to be deleted **before and after** each test to prevent any subsequent tests from loading incorrect configuration, otherwise everything falls apart... All of this still feels convoluted but I think it's still an improvement. Ultimately, I think, all of this burden is due to the lack of the dependency injection.

RussKie:
I re-built this PR few times now, and it's passing... I'm surprised, and I wonder now if I broken something 🤣

GitUI/SplitterManager.cs
```diff
@@ -54,12 +54,15 @@ public SplitterData(SplitContainer splitter, string settingName, int dpi, int? d
...
-            private string Panel2CollapsedSettingsKey => _settingName + "_Panel2Collapsed";
+            internal string SizeSettingsKey => _settingName + "_Size";
+            internal string DpiSettingsKey => _settingName + "_Dpi";
+            internal string DistanceSettingsKey => _settingName + "_Distance";
+            internal string FontSizeSettingsKey => _settingName + "_FontSize";
+            internal string Panel1CollapsedSettingsKey => _settingName + "_Panel1Collapsed";
+            internal string Panel2CollapsedSettingsKey => _settingName + "_Panel2Collapsed";
+
+            [Obsolete("Test only!")]
+            internal SplitContainer Splitter => _splitter;
```
mstv, line 65: not used

IntegrationTests/UI.IntegrationTests/CommandsDialogs/SplitterPersistenceTests.cs
```diff
@@ -48,71 +46,181 @@ public void OneTimeTearDown()
...
+            };
+
+            await RunFormTestAsync(
                 async form =>
                 {
                     FormBrowse.TestAccessor ta = form.GetTestAccessor();
 
                     await WaitForRevisionsToBeLoadedAsync(ta.RevisionGrid);
-
-                    using (new AssertionScope())
```
mstv: This is used in order to not stop verifying on error. Restoring.

mstv (commented):
> I'm surprised, and I wonder now if I broken something

I don't think so. Though it might have become too different from the actual FormBrowse loading.
Of course, much better than disabling the test.

mstv:
> is there anything outstanding here?

just final approvals :)
Then I am going to rebase on and push to master.

RussKie:
I think you have not one but two approvals :) Merge at will.

PR #10960, opened 2023-05-12
Reset from command line

gerhardol:
Resolves #10899 
Resolves #10950 

## Proposed changes

Three parts:
* Refactoring to use common code to reset all changes (just git-reset is not sufficient to reset in all situations).
* Common code to reset file changes.
Note that FormCommit:ResetFilesClick() vs RevisionDiffControl:ResetSelectedItemsWithConfirmation() are not merged. They are slightly different (the first only handles reset to HEAD). They still may require alignment.
* GE command line reset [file] (and revert that really is an alias) raised warnings if new, renamed, conflicted files were selected and did not work at all if 'reset all changes' were used.

Note: This PR (together with #10956) should be included in a 4.1.1, it is one of the most common issues raised.
I thought this was fixed in other PRs, but this is a separate scenario.
(Most issues regarding this have very minimal description.)

Note: This PR does not change the behavior (except command line usage that is aligned to FormCommit). Several of my private test scenarios fail with this PR still. 
I will submit a separate PR to handle these scenarios, not necessarily for 4.1.x

## Test methodology 

Manual

## Merge strategy

- Rebase merge

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10979, opened 2023-05-19
RevDiff: Reset to HEAD for artificial

gerhardol:
## Proposed changes

First Index -> second WorkTree still reset to Index.
All other situations resetting to artificial commits will invisible menu items (and ignored).

Resetting from normal commit to Index does not mean anything
(files are already in Index).

Note: A previous version of this PR used HEAD instead of Index and WorkTreee/Combined reset to HEAD.
This introduced problems when resetting including files only in Index/Worktree, the added/removed status had to be evaluated a second time (possible but too complicated).

Related to #10960 (mentioned, separated to simplify review)
This will allow using common code in a follow up, to fix some reset issues that occur in 4.1 (so this is a candidate for 4.1.x)

Note 2: Resetting to Index is not handled correctly for Index files.
resetting in general is improved in #10980 (some more situations handled, common code). No other changes in this PR.

## Screenshots 

before|after
--|--
worktree_before| same
index_before | image
image | image

## Test methodology 

manual

## Merge strategy

----

### DIFF
--- a/GitUI/CommandsDialogs/RevisionDiffControl.cs
+++ b/GitUI/CommandsDialogs/RevisionDiffControl.cs
@@ -513,9 +513,7 @@ namespace GitUI.CommandsDialogs
                             .Select(item => item.Item.Name)
                             .ToList();
 
-                        // Reset to head for artificial items
-                        ObjectId? resetId = childId.IsArtificial ? _revisionGrid.CurrentCheckout : childId;
-                        Module.CheckoutFiles(itemsToCheckout, resetId, force: false);
+                        Module.CheckoutFiles(itemsToCheckout, GetResetId(childId, _revisionGrid.CurrentCheckout), force: false);
                     }
                 }
                 else
@@ -536,10 +534,13 @@ namespace GitUI.CommandsDialogs
                         Module.RemoveFiles(addedItems.Select(item => item.Item.Name).ToList(), force: false);
                     }
 
+                    // Artificial is to be handled as HEAD, except for Index->WorkTree
+                    bool isIndex = IsParentIndex(selectedItems);
+
                     foreach (ObjectId parentId in selectedItems.FirstIds())
                     {
-                        // Reset to head for artificial items
-                        ObjectId? resetId = parentId.IsArtificial ? _revisionGrid.CurrentCheckout : parentId;
+                        ObjectId? resetId = isIndex ? ObjectId.IndexId : GetResetId(parentId, _revisionGrid.CurrentCheckout);
+
                         List<string> itemsToCheckout = selectedItems
                             .Where(item => !item.Item.IsNew && !(item.Item.IsConflict && parentId == ObjectId.IndexId) && item.FirstRevision?.ObjectId == parentId)
                             .Select(item => RenamedIndexItem(item) ? item.Item.OldName : item.Item.Name)
@@ -560,6 +561,28 @@ namespace GitUI.CommandsDialogs
             {
                 RequestRefresh();
             }
+        }
+
+        /// <summary>
+        /// It is generally not possible to reset to an artificial commit, use HEAD instead.
+        /// </summary>
+        /// <param name="resetId">The selected commit it.</param>
+        /// <param name="headId">The current HEAD id.</param>
+        /// <returns>The id to reset to.</returns>
+        private ObjectId GetResetId(ObjectId? resetId, ObjectId headId)
+        {
+            return (resetId?.IsArtificial ?? true) ? headId : resetId;
+        }
+
+        /// <summary>
+        /// Return if this is a Index->WorkTree selection and reset to parent should be handled as Index
+        /// (artificial commits are normally handled as HEAD).
+        /// </summary>
+        /// <param name="selectedItems">The selected items.</param>
+        /// <returns><see langword="true"/> if this is a Index->WorkTree selection.</returns>
+        private bool IsParentIndex(IEnumerable<FileStatusItem> selectedItems)
+        {
+            return selectedItems.SecondIds().All(i => i == ObjectId.WorkTreeId) && selectedItems.FirstIds().All(i => i == ObjectId.IndexId);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


RussKie (approved):

mstv:
Just my first thoughts. Please convince me.

> Instead of resetting to Index, reset to HEAD when to artificial revisions. Reset to Index does not mean anything.

I think this will break how I use it. This will not be WYSIWYG any longer.
"Reset to Index" means "revert my unstaged - experimental - changes".

> Note: Reverse clicking (artificial before normal commit) will now list the normal commit as "first" and assign the hotkey.

IIUIC, this will be unexpected behavior, too. Sometimes, I want a positive diff, sometimes the reverse.

mstv:
Perhaps it is easier to intercept the git error and to tell the user to unstage the changes before reverting?

gerhardol:
> Just my first thoughts. Please convince me.

> > Instead of resetting to Index, reset to HEAD when to artificial revisions. Reset to Index does not mean anything.
> 
> I think this will break how I use it. This will not be WYSIWYG any longer. "Reset to Index" means "revert my unstaged - experimental - changes".

I do not think it is possible to implement a mostly reliable reset to Index (reset to HEAD is hard enough).
You can still use stage/unstage lines to reset to index, but have to reset conflict state etc manually.

Also, if resetting Index, can WorkTree be changed? It must be, but reset worktree->index indicates it is unchanged.

> > Note: Reverse clicking (artificial before normal commit) will now list the normal commit as "first" and assign the hotkey.
> 
> IIUIC, this will be unexpected behavior, too. Sometimes, I want a positive diff, sometimes the reverse.

Index->HEAD is the only situation, stating Second instead of First. I can fix that, it is just that the code is complex for not much reason.
(Index should be removed regardless.)

mstv:
I am not talking about "resetting index" but about "Reset **to** Index" (which means "revert my unstaged - experimental - changes").
I expect that exactly the displayed diff is reverted.

image

gerhardol:
> I am not talking about "resetting index" but about "Reset **to** Index" (which means "revert my unstaged - experimental - changes"). I expect that exactly the displayed diff is reverted.

The text is correct, it is just if you click Index, then HEAD the First will be "Reset to HEAD" (Reset to Index should be removed here).

I get that you want to keep WorkTree->Index "Reset to Index".

gerhardol:
> > I am not talking about "resetting index" but about "Reset **to** Index" (which means "revert my unstaged - experimental - changes"). I expect that exactly the displayed diff is reverted.
> 
> The text is correct, it is just if you click Index, then HEAD the First will be "Reset to HEAD" (Reset to Index should be removed here).
> 
> I get that you want to keep WorkTree->Index "Reset to Index".

@mstv special: Keep "Reset to Index" when WorkTree is selected (same as click Index, ctrl-click WorkTree commits)? (Suppress errors.)
Maybe with *Reset to HEAD* for this selection?
*Reset to Index* hidden elsewhere. 

Hack implementation for @mstv?
If first is index and second is worktree, reset to Index if worktree. All other artificial is handled as HEAD. If both first/second is HEAD only first is shown in the menu.

mstv:
I don't really understand the above comment.

---

There is a natural "First: A ..." and "Second: B ..." for every one- or two-commit selection (matching the displayed diff). It would be confusing if we tinker it for artificial commits.
Rather intercept the git error and tell the user to unstage the changes before resetting if the handling would become too complex.

gerhardol:
> There is a natural "First: A ..." and "Second: B ..." for every one- or two-commit selection (matching the displayed diff). It would be confusing if we tinker it for artificial commits. Rather intercept the git error and tell the user to unstage the changes before resetting if the handling would become too complex.

WorkTree and Index is not real commits and has special behavior (and handling in both Git and GE).
You cannot reset conflicted files to Index. Current master is already resetting to HEAD (will change that to just ignore the conflicts).
These changes may take a couple of days.

mstv:
If it is just about conflicted files, I am open to any pragmatic solution.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -557,6 +563,28 @@ private void ResetSelectedItemsTo(bool actsAsChild, bool deleteUncommittedAddedI
...
+        /// <param name="resetId">The selected commit it.</param>
+        /// <param name="headId">The current HEAD id.</param>
+        /// <returns>The id to reset to.</returns>
+        private ObjectId GetResetId(ObjectId? resetId, ObjectId headId)
+        {
+            return (resetId?.IsArtificial ?? true) ? headId : resetId;
+        }
+
+        /// <summary>
+        /// Return if this is a Index->WorkTree selection and reset to parent should be handled as Index
```
mstv, line 578:
```suggestion
        /// Return whether this is a Index->WorkTree selection and reset to parent should be handled as Index
```

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -557,6 +563,28 @@ private void ResetSelectedItemsTo(bool actsAsChild, bool deleteUncommittedAddedI
...
+        /// (artificial commits are normally handled as HEAD).
+        /// </summary>
+        /// <param name="selectedItems">The selected items.</param>
+        /// <returns><see langword="true"/> if this is a Index->WorkTree selection.</returns>
```
mstv, line 582: Is there a need for special handling of WorkTree->Index?

mstv (commented):
At WorkDir, "Second: B" should be replaced with "HEAD:" or omitted.
On the other hand, "Reset to HEAD: description" might be a useful menu item regardless of the selected commit.

image

At Index, "Reset file(s) to | Second B: Commit Index" is missing:

image

Basically, the automatic switch to HEAD is misleading. Rather disable the regarding menu item if we do not want to implement it.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -557,6 +563,28 @@ private void ResetSelectedItemsTo(bool actsAsChild, bool deleteUncommittedAddedI
...
 
+        /// <summary>
+        /// It is generally not possible to reset to an artificial commit, use HEAD instead.
+        /// </summary>
+        /// <param name="resetId">The selected commit it.</param>
+        /// <param name="headId">The current HEAD id.</param>
+        /// <returns>The id to reset to.</returns>
+        private ObjectId GetResetId(ObjectId? resetId, ObjectId headId)
+        {
+            return (resetId?.IsArtificial ?? true) ? headId : resetId;
```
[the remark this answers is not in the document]

mstv (commented):
image

"Second B: Commit Index" is missing although it is supported.

image

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: ok

gerhardol:
> Basically, the automatic switch to HEAD is misleading. Rather disable the regarding menu item if we do not want to implement it.

Yes, removed this in a further update. ((#10980 got too complicated.)
Sorry for the second late change.

> "Second B: Commit Index" is missing although it is supported.

The special handling is Index->WorkTree
All other Index is removed, including HEAD->Index.
(The effect of such reset could be that Worktree status is changed when files are reset Index->Index.)

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -557,6 +573,26 @@ private void ResetSelectedItemsTo(bool actsAsChild, bool deleteUncommittedAddedI
...
 
+        /// <summary>
+        /// It is generally not possible to reset to an artificial commit.
+        /// </summary>
+        /// <param name="resetId">The selected commit it.</param>
+        /// <returns><see langword="null"/> if not possible to reset to this revision.</returns>
+        private ObjectId? GetResetId(ObjectId? resetId)
+        {
+            return (resetId?.IsArtificial is false) ? resetId : null;
+        }
```
mstv, line 584:
```suggestion
        private bool CanReset(ObjectId? resetId, ObjectId? secondId)
        {
            return (resetId?.IsArtificial is false) || (resetId == ObjectId.IndexId && secondId == ObjectId.WorkTreeId);
        }
```

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -533,18 +540,27 @@ private void ResetSelectedItemsTo(bool actsAsChild, bool deleteUncommittedAddedI
...
+                    // Artificial is to be handled as HEAD, except for Index->WorkTree
+                    bool isIndex = IsParentIndex(selectedItems);
+
                     foreach (ObjectId parentId in selectedItems.FirstIds())
                     {
+                        ObjectId? resetId = isIndex ? parentId : GetResetId(parentId);
+                        if (resetId is null)
+                        {
+                            continue;
+                        }
```
mstv, line 552:
```suggestion
                    // Artificial is to be handled as HEAD, except for Index->WorkTree
                    ObjectId? secondId = selectedItems.SecondIds().FirstOrDefault();

                    foreach (ObjectId parentId in selectedItems.FirstIds())
                    {
                        if (!CanReset(parentId, secondId))
                        {
                            continue;
                        }
```

mstv (commented):
The function is OK for me.
Since `GetResetId()` returns `null` now, it can return `bool` instead and can be joined with `IsParentIndex()`.

GitUI/CommandsDialogs/RevisionDiffControl.cs
gerhardol: It is possible to multi select, so the check has to be a little more complicated. Something like below.
`IsParentIndex()` will run slightly more often, we can probably ignore that OH though (even if I avoided it above).
Will think about it, not sure I like it.

I want to have a rest-to-head in revdiff too. The original proposal limited to Second for Index->WorkTree would not be so bad, but not in this PR. Need to finish the 4.1stuff, I guess you want reviews on graph changes?

```
private bool CanResetToSecond(ObjectId? resetId)
        {
            return resetId?.IsArtificial is false;
        }
private bool CanResetToFirst(ObjectId? resetId, IEnumerable<FileStatusItem> selectedItems)
        {
            return CanResetToSecond(resetId) || (selectedItems.SecondIds().All(i => i == ObjectId.WorkTreeId) && selectedItems.FirstIds().All(i => i == ObjectId.IndexId));
        }
```

gerhardol: Pushed to tmp/i10899-diff-reset-head-2 for now

mstv: > It is possible to multi select

Yes. (I mixed this with the FileStatusList where all files have the same first & second commit. And below, only the first selection is evaluated.)

> Need to finish the 4.1stuff

Me too. 4.1.1 is for bugfixes only.

gerhardol: I want to add the reset fixes to 4.1.1, this is a prerequisite

gerhardol: Used tmp/i10899-diff-reset-head-2 with suggested changes and doc cleanup

mstv (approved):

PR #10980, opened 2023-05-19
Reset files corrections

gerhardol:
Depends on #10979 #10960 

Follow up to #10853 and #10423 handling some more scenarios

## Proposed changes

Use same code in RevDiff as in FormDiff and command line to reset
individual files.

Reset both index and worktree explicitly before resetting to commits.
(This requires additional call to git-status.)
This handles some more scenarios for renamed reset files and conflicts when both Index and Worktree must be evaluated.
https://git-scm.com/docs/git-status#_short_format

Focus is on avoiding git-errors. There are probably some scenarios with renamed files that could cause Git error codes still, but this should work in most situations.

This should be considered for 4.1.x, there has been a number of issues raised.

## Test methodology 

Some manually setup branches to create some of the scenarios for renamed and conflicts.

## Merge strategy

----


gerhardol:
Blocked by #10993
Some tests added for inverted (not important)
I still want this in 4.1.1
Description edited
May need to update for resetting worktree->head, some more scenarios

mstv:
Is it intended that no command is issued when attempting to reset a conflicted file?

image

gerhardol:
> Is it intended that no command is issued when attempting to reset a conflicted file?

Yes, you cannot reset anything without changing the index.
A reason for #10979

mstv:
> Yes, you cannot reset anything without changing the index.

So rather display the files with conflicts in index, not in workdir? 🤔

Because asking for confirmation and then do nothing is ... unexpected.

gerhardol:
> > Yes, you cannot reset anything without changing the index.
> 
> So rather display the files with conflicts in index, not in workdir? 🤔

The file actually have a status in Index too, so some very special handling has to be done in this case.
I believe it is complicated enough in this PR, but it could probably be handled.

> Because asking for confirmation and then do nothing is ... unexpected.

Despite best effort, there could be other situations that are not handled.
Before 4.0, resetting Unmerged/Conflicted had no effect at all.

The ouput from git-checkout could be shown in a popup, but it requires changes to several levels, I would like to avoid it here.
(Some output is shown in FormCommit.)

mstv (commented):
+1 Looks sensible

An (exception) popup on attempt to reset conflicted / unmerged files would be useful - particularly since those are explicitly mentioned in the PR description.

gerhardol:
> An (exception) popup on attempt to reset conflicted / unmerged files would be useful - particularly since those are explicitly mentioned in the PR description.

I would like to handle that separately, this PR does not change this behavior (other than avoiding NBug, which kind of informed about them).

An integration test for ResetChanges() would be good, but relevant test cases are extremely hard to define, real Git repos should be used.

gerhardol:
@mstv See tmp/i10899-rev-diff-common for a (naive) implementation to support getting the output for batched commands.
Not this PR

gerhardol:
rebased to resolve merge conflict

mstv (approved):

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -1443,20 +1391,21 @@ private bool ResetSelectedFilesWithConfirmation()
...
             return true;
         }
 
-        private void ResetSelectedItemsWithConfirmation(bool resetToSelected)
+        private void ResetSelectedItemsWithConfirmation(bool resetToParent)
         {
             IEnumerable<FileStatusItem> items = DiffFiles.SelectedItems;
 
+            // new could be changed when restting, allow user to add to checkbox
+            bool hasNewFiles = true;
```
mstv, line 1403: This is unlucky. At least for a single file selection, it should be set correctly.

image

gerhardol: You do not know if a file will be considered new after a checkout, like for renamed files.
This enables the checkbox, do not set it by default.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -1443,20 +1391,21 @@ private bool ResetSelectedFilesWithConfirmation()
...
+            ResetSelectedItemsWithConfirmation(resetToParent: true);
             return true;
         }
 
-        private void ResetSelectedItemsWithConfirmation(bool resetToSelected)
+        private void ResetSelectedItemsWithConfirmation(bool resetToParent)
         {
             IEnumerable<FileStatusItem> items = DiffFiles.SelectedItems;
 
+            // new could be changed when restting, allow user to add to checkbox
```
mstv, line 1402: > You do not know if a file will be considered new after a checkout, like for renamed files.
This enables the checkbox, do not set it by default.

Does it cause an error message from git? If not and it only leaves a "new" file, the user can get rid of this "new" file by resetting again. Then the checkbox will be automatically ticked.

```suggestion
            // The "new" state could change when resetting, allow user to tick the checkbox.
```

gerhardol: > Does it cause an error message from git? If not and it only leaves a "new" file, the user can get rid of this "new" file by resetting again. Then the checkbox will be automatically ticked.

The change is just about not disabling the checkbox. It is up to the user to tick the box or to redo the reset with leftover files. (There is no Git error message, the new status is detected by an additional git-status call.)

The handling was incomplete, for instance reset-to-selected were not considered at all.

The reverse approach to disable the checkbox when there are only modified files could work...

Can change the comment.

mstv: > The reverse approach to disable the checkbox when there are only modified files could work...

Yes, please. I am used to the usually disabled checkbox. If it is enabled, I wonder why, and get distracted.

GitUI/CommandsDialogs/RevisionDiffControl.cs
```diff
@@ -1443,20 +1391,22 @@ private bool ResetSelectedFilesWithConfirmation()
...
         }
 
-        private void ResetSelectedItemsWithConfirmation(bool resetToSelected)
+        private void ResetSelectedItemsWithConfirmation(bool resetToParent)
         {
             IEnumerable<FileStatusItem> items = DiffFiles.SelectedItems;
 
+            // The "new" state could change when resetting, allow user to tick the checkbox.
+            // If there are only changed files, it is safe to disable the checkboc (also for restting to selected).
+            bool hasNewFiles = !items.All(item => item.Item.IsChanged);
```
mstv, line 1404: Thank you!

```suggestion
            // The "new" state could change when resetting, allow user to tick the checkbox.
            // If there are only changed files, it is safe to disable the checkbox (also for resetting to selected).
            bool hasNewFiles = !items.All(item => item.Item.IsChanged);
```

mstv (commented):
When "resetting" a really new file from workdir, it gets deleted by git should not be invoked for it:

Exit code: 1
Arguments: checkout -- "test.log"
Working directory: D:\Build\gitextensions3_dev\

```
GitExtUtils.ExternalOperationException: error: pathspec 'test.log' did not match any file(s) known to git
 ---> System.Exception: error: pathspec 'test.log' did not match any file(s) known to git
   at GitCommands.ExecutableExtensions.RunCommandAsync(IExecutable executable, ArgumentString arguments, Byte[] input, Boolean createWindow) in D:\Build\gitextensions3_dev\GitCommands\Git\ExecutableExtensions.cs:line 229
   at Microsoft.VisualStudio.Threading.JoinableTask`1.CompleteOnCurrentThread()
   at Microsoft.VisualStudio.Threading.JoinableTaskFactory.Run[T](Func`1 asyncMethod, JoinableTaskCreationOptions creationOptions)
   at Microsoft.VisualStudio.Threading.JoinableTaskFactory.Run[T](Func`1 asyncMethod)
   at GitCommands.ExecutableExtensions.RunCommand(IExecutable executable, ArgumentString arguments, Byte[] input, Boolean createWindow) in D:\Build\gitextensions3_dev\GitCommands\Git\ExecutableExtensions.cs:line 168
   at GitCommands.ExecutableExtensions.RunBatchCommand(IExecutable executable, ICollection`1 batchArguments, Action`1 action, Byte[] input, Boolean createWindow) in D:\Build\gitextensions3_dev\GitCommands\Git\ExecutableExtensions.cs:line 199
   at GitCommands.GitModule.CheckoutFiles(IReadOnlyList`1 files, ObjectId revision, Boolean force) in D:\Build\gitextensions3_dev\GitCommands\Git\GitModule.cs:line 1544
   at GitCommands.GitModule.ResetChanges(ObjectId resetId, IReadOnlyList`1 selectedItems, Boolean resetAndDelete, IFullPathResolver fullPathResolver, List`1& filesInUse, StringBuilder& output, Action`1 progressAction) in D:\Build\gitextensions3_dev\GitCommands\Git\GitModule.cs:line 1483
   at GitUI.CommandsDialogs.RevisionDiffControl.ResetSelectedItemsTo(Boolean resetToParent, Boolean resetAndDelete) in D:\Build\gitextensions3_dev\GitUI\CommandsDialogs\RevisionDiffControl.cs:line 510
   at GitUI.CommandsDialogs.RevisionDiffControl.ResetSelectedItemsWithConfirmation(Boolean resetToParent) in D:\Build\gitextensions3_dev\GitUI\CommandsDialogs\RevisionDiffControl.cs:line 1419
   at GitUI.CommandsDialogs.RevisionDiffControl.ResetSelectedFilesWithConfirmation() in D:\Build\gitextensions3_dev\GitUI\CommandsDialogs\RevisionDiffControl.cs:line 1394
   at GitUI.CommandsDialogs.RevisionDiffControl.ExecuteCommand(Int32 cmd) in D:\Build\gitextensions3_dev\GitUI\CommandsDialogs\RevisionDiffControl.cs:line 161
   at ResourceManager.GitExtensionsControl.ProcessHotkey(Keys keyData) in D:\Build\gitextensions3_dev\ResourceManager\GitExtensionsControl.cs:line 86
   at GitUI.CommandsDialogs.RevisionDiffControl.ProcessHotkey(Keys keyData) in D:\Build\gitextensions3_dev\GitUI\CommandsDialogs\RevisionDiffControl.cs:line 1326
   at ResourceManager.GitExtensionsControl.ProcessCmdKey(Message& msg, Keys keyData) in D:\Build\gitextensions3_dev\ResourceManager\GitExtensionsControl.cs:line 98
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at System.Windows.Forms.ContainerControl.ProcessCmdKey(Message& msg, Keys keyData)
   at ResourceManager.GitExtensionsControl.ProcessCmdKey(Message& msg, Keys keyData) in D:\Build\gitextensions3_dev\ResourceManager\GitExtensionsControl.cs:line 98
```

gerhardol:
> When "resetting" a really new file from workdir, it gets deleted. But git should not be invoked for it:

Pushed a change
This is very special handling for WorkDir->Index.
I might have indicated that I did not want to implement that due to complexity...

mstv (approved):
Works well and more automatically.
One needs to be aware that resetting a stashed file will also discard unstaged changes - as the confirmation popup clearly states.

GitCommands/Git/GitItemStatus.cs
```diff
@@ -237,6 +237,30 @@ internal void SetSubmoduleStatus(JoinableTask<GitSubmoduleStatus?> status)
             _submoduleStatus = status;
         }
 
+        /// <summary>
+        /// Return an object with the status as if the item was created
+        /// with first and second commit reverse selected.
+        /// </summary>
+        /// <returns>An inverted copy.</returns>
+        public GitItemStatus ReverseSelection()
```
RussKie, line 245: I'm not finding the signature meaningful - it's quite unexpected for a method called "reverse selection" to return "item status". Maybe call something like `InvertStatus()`?

GitCommands/Git/GitModule.cs
```diff
@@ -1370,7 +1370,7 @@ public void Reset(ResetMode mode, string? file = null)
             _gitExecutable.RunCommand(GitCommandHelpers.ResetCmd(mode, commit: null, file));
         }
 
-        public string ResetFiles(IReadOnlyList<string> files)
+        public string CheckoutIndexFiles(IReadOnlyList<string> files)
```
RussKie, line 1373: I see your intent here, but IMO the old name communicated the method's purpose better. Maybe the new name should mention "Force" too? E.g., `ForceCheckoutIndexFiles`....
Xml-doc would be helpful too.

GitCommands/Git/GitItemStatus.cs
gerhardol: Works, changing

GitCommands/Git/GitModule.cs
gerhardol: Added xml-doc, but I would like to keep the name, the behavior is different from git-reset.

GitCommands/Git/GitModule.cs
```diff
@@ -1370,7 +1370,12 @@ public void Reset(ResetMode mode, string? file = null)
             _gitExecutable.RunCommand(GitCommandHelpers.ResetCmd(mode, commit: null, file));
         }
 
-        public string ResetFiles(IReadOnlyList<string> files)
+        /// <summary>
+        /// checkout-index ("reset"), copy files from the index overwrite the working tree.
```
RussKie, line 1374: ❓ 
```suggestion
        /// Executes <c>checkout-index</c> ("reset") command which copies files from the index overwriting the working tree.
```

RussKie (commented):
👍

PR #10994, opened 2023-05-30
support rolling forward to sdk 7

SimonCropp:


mstv (changes requested):

RussKie:
It's totally acceptable as a local tweak, if a contributor doesn't want to install .NET 6 SDK, but I don't think we can roll the change at scale.

PR #11006, opened 2023-05-31
remove some un-used members

SimonCropp:


GitUI/Theming/Renderers/ButtonRenderer.cs
```diff
@@ -280,21 +280,6 @@ public enum CheckBox
                 CBS_EXCLUDEDPRESSED = 19,
                 CBS_EXCLUDEDDISABLED = 20,
             }
-
```
gerhardol, line 283: Theme related, really remove?

BugReporter/Info/AssemblyInfo.cs
```diff
@@ -1,12 +0,0 @@
-﻿// --------------------------------------------------------------------------------------------------------------------
-// <copyright file="AssemblyInfo.cs" company="NBug Project">
-//   Copyright (c) 2011 - 2013 Teoman Soygul. Licensed under MIT license.
-// </copyright>
-// --------------------------------------------------------------------------------------------------------------------
-
-namespace BugReporter.Info
-{
-    public class AssemblyInfo
```
mstv, line 9: Are this and the following two `...Info` classes accessed using reflection in the BugReporter, @RussKie?

RussKie: I don't think so... I did a quick search for these types, and I didn't see any uses. Let's nuke 'em :)

GitUI/Theming/Renderers/ButtonRenderer.cs
RussKie: As an option, if we want to keep these, we can move all these unused theme-related types under `#if SUPPORT_THEME_HOOKS`.

SimonCropp: IMO u should remove them. then mark that commit with a tag. if u ever want to bring them back u can find them via the tag, and then revert the commit

RussKie: We've hidden all theme-related stuff behind `SUPPORT_THEME_HOOKS` pragma. Having all of those under the same pragma will make it easier to bulk-remove or bulk-reenable in the future.

gerhardol: I suggest this PR is changed to only include the first 5 removed files.

The theming related should be adding SUPPORT_THEME_HOOKS.
I still hope theming can be reenabled, then it is easier if the code is not removed.
However if removed, remove all in one dedicated commit, not many. There are probably other theming related methods that will be removed over time.

PR #8843, opened 2021-02-10
dotnet/winforms .editorconfig

RussKie:
Closes #8814 as superseded

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #11007, opened 2023-05-31
add some missing NotNullWhen

SimonCropp:


GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -334,7 +334,7 @@ public void Add(GitRevision revision, bool insertWithMatch = false, int insertRa
                 {
                     foreach (var parentId in parents)
                     {
-                        if (_revisionGraph.TryGetNode(parentId, out RevisionGraphRevision parentRev))
+                        if (_revisionGraph.TryGetNode(parentId, out var parentRev))
```
gerhardol, line 337: Why?

gerhardol: This is being changed from var to the actual type.
I believe this was changed by analyzers with what was possible a couple of years ago, but no generic cleanup.

RussKie:
We kind of adopted the same approach as .NET repo use - i.e., use var only
the type is obvious. Otherwise specify the type explicitly.

SimonCropp:
@RussKie so in this case; is the type obvious?

mstv (approved):
modulo use of `var`

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: Here's the rule I'm referring to: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md#c-coding-style point 10.

We still need to finalise https://github.com/gitextensions/gitextensions/pull/8843 (with additional tweaks, if necessary) and update [our coding style guide](https://github.com/gitextensions/gitextensions/wiki/Guidelines%3A-Coding-Style).

RussKie (approved):

RussKie:
Thank you

PR #11014, opened 2023-06-01
Fixup 3-revision diff with artificial

mstv:
## Proposed changes

- FileStatusDiffCalculator: Do not attempt to detect artificial commit as base commit
- RevisionDiffControl: Fixup silent exceptions from `SetDiffsAsync`

### Before

image

### After

image

## Test methodology 

- manual

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/FileStatusDiffCalculator.cs
+++ b/GitUI/UserControls/FileStatusDiffCalculator.cs
@@ -124,15 +124,16 @@ namespace GitUI
             }
             else
             {
-                // check that the middle commit is base to the first/second
+                // Check whether the middle commit is base to the first/second
                 // Allow commits earlier to the actual base commit
-                ObjectId? testBaseRevId = GetMergeBase(firstRevHead, revisions[1].ObjectId);
-                if (testBaseRevId == revisions[1].ObjectId)
+                GitRevision middleRev = revisions[1];
+                if (!middleRev.IsArtificial)
                 {
-                    testBaseRevId = GetMergeBase(selectedRevHead, revisions[1].ObjectId);
-                    if (testBaseRevId == revisions[1].ObjectId)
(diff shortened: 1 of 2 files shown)


GitUI/UserControls/FileStatusDiffCalculator.cs
```diff
@@ -124,15 +124,16 @@ public IReadOnlyList<FileStatusWithDescription> Reload()
             }
             else
             {
-                // check that the middle commit is base to the first/second
+                // Check whether the middle commit is base to the first/second
                 // Allow commits earlier to the actual base commit
-                ObjectId? testBaseRevId = GetMergeBase(firstRevHead, revisions[1].ObjectId);
-                if (testBaseRevId == revisions[1].ObjectId)
+                GitRevision middleRev = revisions[1];
```
gerhardol, line 129: No need for a middleRev variable

```suggestion
                ObjectId middleId = revisions[1];
```

A better way of writing the check.
One if statement could have been used, this is OK

When is an artificial returned from GitModule.GetMergeBase?

gerhardol (approved):

GitUI/UserControls/FileStatusDiffCalculator.cs
mstv: Ah yes, there is also `ObjectId.IsArtificial`.

> When is an artificial returned from GitModule.GetMergeBase?

Never.
Just select something like "master", "artificial", "release" in this sequence.

mstv: > One `if` statement

I like it.

> Just select something like "master", "artificial", "release" in this sequence.

Then the resulting exception "artificial id shall not be used in git command" was sent to the Nirvana, "Loading data" forever.

gerhardol (approved):

Issue #10991, opened 2023-05-25
Suppress diff with untracked files of stash


mstv:
### Feature description

[The _added_ files in] the third diff group is pointless (all repo files) and takes very long to load:

![grafik](https://github.com/gitextensions/gitextensions/assets/36601201/c1e73910-7a12-4477-adf6-86df21eee86a)  ![grafik](https://github.com/gitextensions/gitextensions/assets/36601201/88b84fc1-9aac-424a-8acf-73fa0f1673bb)

### Environment

- GitExtensions version: 4.1.0

PR #11018, opened 2023-06-01
FileStatusList: Always show diff groups if multiple groups although empty

mstv:
Contributes to #10991

## Proposed changes

- FileStatusList: Always show diff groups if multiple groups although empty 

### Before

image

### After

image

## Test methodology 

- manual

## Merge strategy

----


gerhardol (approved):

RussKie:
Why do we need those if there are no changes?

mstv:
> Why do we need those if there are no changes?

because of "No changes" is not really true for

image

But we do not want to display the third group because it is not really useful and costs much time in larger repos.

RussKie:
👍

PR #10915, opened 2023-04-25
Avoid double graph drawing

mstv:
## Proposed changes

- Do not draw merged parts of the graph multiple times in order to avoid graphical artifacts

### Before

above commit `1a6d9175b626d89bd710ee35a87fe1e2a5d85641`

image

### After

image

## Test methodology 

- manual

## Please Do Not Squash Merge

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #10864, opened 2023-04-07
Do not straighten re-used graph lanes

mstv:
Fixes https://github.com/gitextensions/gitextensions/pull/10778#discussion_r1155193582

## Proposed changes

- Do not straighten re-used graph lanes

### Before | After

image   image

image   image above commit `1a6d9175b626d89bd710ee35a87fe1e2a5d85641`

## Test methodology 

- 

## Please do **not squash** merge

----

:black_nib: I contribute this code under [The Developer Certificate of Origin](../blob/master/contributors.txt).

PR #11036, opened 2023-06-10
Preview of graph with diagonals

mstv:
## Proposed changes

- Draw graph with diagonals to the max
  Thank you, Igor for giving the criterion "irregular angles"!
- Most of the improvements can be turned off in the `View` menu in order to compare with the simple and astonishingly effective initial implementation.

### Before | After | Optional

image image image

image

## Test methodology 

- manual
- add test cases

## Merge strategy

Do not merge this in one!

----


pmiossec:
The optional one seems the better for me... especially because it materialize a lot better the start of the pink branch (that I found out misleading for beginners in the past).

I can't find an issue from the screenshots provided...
👍

gerhardol:
Looks good. I have been using some of your improvements for a while, some more here.
If you believe this is better I am generally in favor.

Is this intended for master as is?
If so, I would like to hide experimental to Settings, at least before merge.

What is the relation to the other graph related PRs? Superseding or should the other be merged first?
How to review?

--

Quick comments:

Any of the new settings that really are controversial?
Is the PR defaults the suggested defaults in next release?
Any of the setting that should be kept in a release?
I suggest the mstv-debug-only is added to the xml only before merging.

What is the point of `DrawGraphLanesPrimary` setting?
Hardly any lines when deactivating...
image

`DrawLanesDifferentEnd` gives artifacts, Similar for `DrawLanesDifferentStart`
image

mstv:
> Is this intended for master as is? 

No, this is a preview of the combination of several PRs plus options to quickly enable / disable the new features for easier comparison.

> What is the relation to the other graph related PRs? Superseding or should the other be merged first? How to review?

#10864 is the base. The next unrelated two depend on this: #10915 and the essence of #10778. These three do not need to be made optional.
Afterwards, we should agree on View Settings.

> Any of the new settings that really are controversial?

No, all combinations are valid. (Only `StraightOneLaneDiagonals` has no effect without `DrawGraphWithDiagonals`.)

> Is the PR defaults the suggested defaults in next release?

We should agree on the defaults based on the full final picture. That's why I created this PR.

> Any of the setting that should be kept in a release?

The option `StraightOneLaneDiagonals` should not be kept - but the feature.
`ReduceGraphCrossings` is very recommended for disabled `MergeGraphLanesHavingCommonParent`.
I like `StraightenGraphDiagonals` very much. I makes sense for the curvy graph, too. Though personal preferences can be different. 

I should have added the prefix `Debug` to `DrawGraphLanesDifferentEnd`, `DrawGraphLanesDifferentStart`, `DrawGraphLanesEntire`, `DrawGraphLanesPrimary`, i.e. to be removed.
The same applies to `StraightenOverCommits` (#10778) and `StraightenOverSecondaryCommits` (#10864).

> I suggest the mstv-debug-only is added to the xml only before merging.

Could you elaborate please?

> What is the point of `DrawGraphLanesPrimary` setting? ...

These four debug options are intended just for the demonstration of how #10864 and #10915 are implemented.

spdr870:
Your proposals look very similar to my earlier experiments (https://github.com/gitextensions/gitextensions/issues/5782). Just for reference:
https://github.com/gitextensions/gitextensions/issues/5782#issuecomment-439721628
https://github.com/gitextensions/gitextensions/issues/5782#issuecomment-439354469

It's a very long discussion, including multiple working examples.

mstv:
> Your proposals look very similar to my earlier experiments (#5782).

Your kudos can be found in the dedicated PR #10637 (to be updated after merging a few other other PRs).

PR #11046, opened 2023-06-14
Improve revision grid scrolling

MaxKoll:
## Proposed changes

- Utilize the OnMouseWheel() override to prevent the raising of redundant Scroll events and skip to the target row instantly while scrolling.

- Calculate the scrolling distance by adding up the mouse wheel Delta values, which leads to smoother scrolling with high-precision devices such as free-spinning mouse wheels or touchpads.

### Before

When rotating the mouse wheel by one unit, multiple Scroll events might be emitted simultaneously in accordance with SystemInformation.MouseWheelScrollLines. This leads to the revision grid being scrolled row by row with intermittent updates, which causes flickering and visual glitches.

### After

This reduces flickering and visual glitches when scrolling the revision grid via mouse wheel, and also improves scrolling with high-precision scrolling devices.

This also considerably reduces the noticeability of the BackgroundUpdater delay as discussed in #10680.

## Test methodology 

- Lots of scrolling with (1) a notched mouse wheel, (2) an app that simulates "smooth scrolling" and (3) a touchpad.
- Inserting Debug.WriteLine(...) at certain places in the code and observing the output window.
- Comparing different revisions of Git Extensions against each other.

## Test environment(s) 

- Git 2.38.1.windows.1
- .NET 6.0.16
- DPI 134dpi (140% scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -890,13 +890,12 @@ namespace GitUI.UserControls.RevisionGrid
                 int totalWheelDelta = (scrollLines * e.Delta) + _mouseWheelDeltaRemainder;
                 int wheelDeltaPerRow = SystemInformation.MouseWheelScrollDelta;
 
-                if (Math.Abs(totalWheelDelta) >= wheelDeltaPerRow)
+                // The total wheel delta is consumed in multiples of wheelDeltaPerRow.
+                // Save the remainder and credit it to the total wheel delta during the next MouseWheel event.
+                _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
+                int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
+                if (rowDelta != 0)
                 {
-                    // The total wheel delta is consumed in multiples of wheelDeltaPerRow.
-                    // Save the remainder and credit it to the total wheel delta during the next MouseWheel event.
-                    _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
-                    int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
-
                     int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, maxRowIndex);
 
                     // This will raise the Scroll event.
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -894,13 +894,15 @@ namespace GitUI.UserControls.RevisionGrid
 
                 // The wheel might be configured to scroll more than one row at once.
                 // Respect this by scaling MouseEventArgs.Delta accordingly.
-                int scrollLines = SystemInformation.MouseWheelScrollLines switch
+                int scrollLines = SystemInformation.MouseWheelScrollLines;
+
+                // Value of -1 indicates the "One screen at a time" mouse option.
+                if (scrollLines == -1)
                 {
-                    // Value of -1 indicates the "One screen at a time" mouse option.
-                    -1 => visibleCompleteRowsCount > 0 ? visibleCompleteRowsCount : 1,
-                    > 0 => SystemInformation.MouseWheelScrollLines,
-                    _ => 1
-                };
+                    scrollLines = visibleCompleteRowsCount;
+                }
+
+                scrollLines = Math.Max(1, scrollLines);
 
                 // Calculate the total wheel delta, which corresponds to the intended scrolling distance, from
                 // MouseEventArgs.Delta, which is usually a multiple of SystemInformation.MouseWheelScrollDelta
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -932,7 +932,7 @@ namespace GitUI.UserControls.RevisionGrid
                     }
                     catch (ArgumentOutOfRangeException)
                     {
-                        // Tried to scroll to nonexistant row.
+                        // Tried to scroll to nonexistent row.
                     }
                 }
             }
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -17,6 +17,7 @@ namespace GitUI.UserControls.RevisionGrid
     public sealed partial class RevisionDataGridView : DataGridView
     {
         private const int BackgroundThreadUpdatePeriod = 25;
+        private const int MouseWheelDeltaTimeout = 1500; // Mouse wheel idle time in milliseconds after which unconsumed wheel delta will be dropped.
         private static readonly AccessibleDataGridViewTextBoxCell _accessibleDataGridViewTextBoxCell = new();
 
         private readonly SolidBrush _alternatingRowBackgroundBrush;
@@ -880,7 +881,7 @@ namespace GitUI.UserControls.RevisionGrid
                 //   for scrolling one row is never reached on the first wheel rotation.
                 // - When using a precision scrolling device, the unconsumed delta will offset the first scroll,
                 //   which makes the user experience a subtle "lag" or "leap" at beginning of a scroll.
-                if (_lastMouseWheel.ElapsedMilliseconds > 1500)
+                if (_lastMouseWheel.ElapsedMilliseconds > MouseWheelDeltaTimeout)
                 {
                     _mouseWheelDeltaRemainder = 0;
                 }
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -873,12 +873,20 @@ namespace GitUI.UserControls.RevisionGrid
             }
             else
             {
-                // Reset unconsumed wheel delta when the mouse wheel is idle.
+                // Reset unconsumed wheel delta when the mouse wheel is idle, because there are at least
+                // two situations in which unconsumed wheel delta causes an issue:
+                // - When switching back to a notched mouse wheel. Whenever the scroll direction is changed,
+                //   unconsumed delta will reduce the absolute value of the total delta so that the threshold
+                //   for scrolling one row is never reached on the first wheel rotation.
+                // - When using a precision scrolling device, the unconsumed delta will offset the first scroll,
+                //   which makes the user experience a subtle "lag" or "leap" at beginning of a scroll.
                 if (_lastMouseWheel.ElapsedMilliseconds > 1500)
                 {
                     _mouseWheelDeltaRemainder = 0;
                 }
 
+                // The wheel might be configured to scroll more than one row at once.
+                // Respect this by scaling MouseEventArgs.Delta accordingly.
                 int scrollLines = SystemInformation.MouseWheelScrollLines switch
                 {
                     // Value of -1 indicates the "One screen at a time" mouse option.
@@ -887,11 +895,15 @@ namespace GitUI.UserControls.RevisionGrid
                     _ => 1
                 };
 
+                // Calculate the total wheel delta, which corresponds to the intended scrolling distance, from
+                // MouseEventArgs.Delta, which is usually a multiple of SystemInformation.MouseWheelScrollDelta
+                // for notched mouse wheels, but can be an arbitrary number in the case of precision scrolling
+                // devices like free-spinning mouse wheels or touchpads.
+                // Consume the total wheel delta in multiples of SystemInformation.MouseWheelScrollDelta, which
+                // is the wheel delta threshold for scrolling one row, and save the remainder for consumption
+                // during the next MouseWheel event.
                 int totalWheelDelta = (scrollLines * e.Delta) + _mouseWheelDeltaRemainder;
                 int wheelDeltaPerRow = SystemInformation.MouseWheelScrollDelta;
-
-                // The total wheel delta is consumed in multiples of wheelDeltaPerRow.
-                // Save the remainder and credit it to the total wheel delta during the next MouseWheel event.
                 _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
                 int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
                 if (rowDelta != 0)
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -921,7 +921,7 @@ namespace GitUI.UserControls.RevisionGrid
 
                     // Drop unconsumed wheel delta when reaching the upper or lower bound of the grid
                     // to prevent the grid being stuck there for a moment.
-                    if (toRowIndex == 0 || toRowIndex >= maxRowIndex - visibleCompleteRowsCount + 1)
+                    if (toRowIndex == 0 || toRowIndex + visibleCompleteRowsCount > maxRowIndex)
                     {
                         _mouseWheelDeltaRemainder = 0;
                     }
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -889,10 +889,13 @@ namespace GitUI.UserControls.RevisionGrid
                     int consumedWheelDelta = _mouseWheelDelta - (_mouseWheelDelta % wheelDeltaPerRow);
                     _mouseWheelDelta -= consumedWheelDelta;
                     int rowDelta = -consumedWheelDelta / wheelDeltaPerRow;
-                    int toIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, _revisionGraph.Count - 1);
+                    if (_revisionGraph.Count > 0)
+                    {
+                        int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, _revisionGraph.Count - 1);
 
-                    // This will raise the Scroll event.
-                    FirstDisplayedScrollingRowIndex = toIndex;
+                        // This will raise the Scroll event.
+                        FirstDisplayedScrollingRowIndex = toRowIndex;
+                    }
                 }
 
                 _lastMouseWheel.Restart();
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -26,6 +26,7 @@ namespace GitUI.UserControls.RevisionGrid
         private readonly Stopwatch _lastRepaint = Stopwatch.StartNew();
         private readonly Stopwatch _lastScroll = Stopwatch.StartNew();
         private readonly Stopwatch _consecutiveScroll = Stopwatch.StartNew();
+        private readonly Stopwatch _lastMouseWheel = Stopwatch.StartNew();
         private readonly List<ColumnProvider> _columnProviders = new();
 
         internal RevisionGraph _revisionGraph = new();
@@ -36,6 +37,7 @@ namespace GitUI.UserControls.RevisionGrid
 
         private int _backgroundScrollTo;
         private int _rowHeight; // Height of elements in the cache. Is equal to the control's row height.
+        private int _mouseWheelDelta;
 
         private VisibleRowRange _visibleRowRange;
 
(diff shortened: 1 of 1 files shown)


GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +867,34 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                {
+                    // Value of -1 indicates the "One screen at a time" mouse option.
+                    -1 => _rowHeight > 0 ? (Height / _rowHeight) : 1,
+                    > 0 => SystemInformation.MouseWheelScrollLines,
+                    _ => 1
+                };
+
+                _mouseWheelDelta += scrollLines * e.Delta;
+
+                // Wheel delta value of 120 corresponds to one row.
```
mstv, line 885:
```suggestion
                const int wheelDeltaPerRow = 120;
```

mstv (commented):
👍

mstv (approved):

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +867,35 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+
+                if (Math.Abs(_mouseWheelDelta) >= wheelDeltaPerRow)
+                {
+                    int consumedWheelDelta = _mouseWheelDelta - (_mouseWheelDelta % wheelDeltaPerRow);
+                    _mouseWheelDelta -= consumedWheelDelta;
+                    int rowDelta = -consumedWheelDelta / wheelDeltaPerRow;
+                    int toIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, _revisionGraph.Count - 1);
+
+                    // This will raise the Scroll event.
+                    FirstDisplayedScrollingRowIndex = toIndex;
```
MaxKoll, line 895: I just noticed that Math.Clamp() will throw if _revisionGraph.Count == 0. Does this ever happen? Maybe for an instant between operations when the grid is cleared and reloaded? I'd like to suggest the following safeguard (the grid is scrollable if there are at least two rows):

```suggestion
                    int maxRowIndex = _revisionGraph.Count - 1;
                    if (maxRowIndex > 0)
                    {
                        int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, maxRowIndex);

                        // This will raise the Scroll event.
                        FirstDisplayedScrollingRowIndex = toRowIndex;
                    }
```

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -889,10 +889,13 @@ protected override void OnMouseWheel(MouseEventArgs e)
                     int consumedWheelDelta = _mouseWheelDelta - (_mouseWheelDelta % wheelDeltaPerRow);
                     _mouseWheelDelta -= consumedWheelDelta;
                     int rowDelta = -consumedWheelDelta / wheelDeltaPerRow;
-                    int toIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, _revisionGraph.Count - 1);
+                    if (_revisionGraph.Count > 0)
+                    {
+                        int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, _revisionGraph.Count - 1);
```
mstv, line 894: `.Count` might be a little expensive. Avoid reading it twice. Could be done first in this function.
```suggestion
                    int maxIndex = _revisionGraph.Count - 1;
                    if (maxIndex < 0)
                    {
                        return;
                    }
                    
                    int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, maxIndex);
```

MaxKoll: Ah, okay, I wasn't aware of that.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +867,38 @@ protected override void OnMouseWheel(MouseEventArgs e)
             }
             else
             {
-                base.OnMouseWheel(e);
+                if (_lastMouseWheel.ElapsedMilliseconds > 1500)
```
gerhardol, line 870: Why 1500, why is this needed at all?
Why is wheelDeltaPerRow  120? (the comment that 120 corresponded to one row may be needed)

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -36,6 +37,7 @@ public sealed partial class RevisionDataGridView : DataGridView
 
         private int _backgroundScrollTo;
         private int _rowHeight; // Height of elements in the cache. Is equal to the control's row height.
+        private int _mouseWheelDelta;
```
gerhardol, line 40: Please add a comment

MaxKoll:
Somehow I find it odd that `RowCount` is apparently unreliable and `_revisionGraph.Count` has to be used instead. Intuitively I would have thought it was the other way around.

gerhardol (commented):
+1

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
MaxKoll: > Why is wheelDeltaPerRow 120?

This seems to be a historic value and the common definition: [https://learn.microsoft.com/.../wm-mousewheel](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel)

I just found out that this value can be accessed via `SystemInformation.MouseWheelScrollDelta` which is just a getter for the internal constant `WHEEL_DELTA = 120`: [https://github.com/dotnet/.../SystemInformation.cs#L884](https://github.com/dotnet/winforms/blob/d892c6c3186f26e763a9f54f80a1379c845e1c4d/src/System.Windows.Forms/src/System/Windows/Forms/SystemInformation.cs#L884)
I will make a change to use `SystemInformation.MouseWheelScrollDelta`. Should I still add comment, and what information should it include?

> Why is the reset timer needed at all?

I stumbled upon this when I tested "smooth scrolling" and changed back to notched scrolling. While a notched wheel produces delta values that are multiples of 120, smooth scrolling produces arbitrary values and therefore leaves behind unconsumed total delta (since it's consumed in multiples of 120). Now assume the total delta is, say, -100 und you rotate the wheel by one notch in the positive direction, then nothing will happen, because the total delta is now at +20, but the delta needed for scrolling one row is 120. Whenever you change direction, the first notch will do nothing. I can imagine myself experiencing this when using a laptop with a mouse attached and switching between touchpad (= smooth scrolling) and mouse randomly.
If you use only a touchpad or a free-spinning mouse, you might also experience a short "lag" or "jump" at the beginning of a scroll for the same reasons.

There might be more intricate ways to reset the total wheel delta (all needing a timer somehow, I guess), but a single timer is presumably good enough, and simple of course.

> Why 1500?

It's a recommendation I found here: https://bugzilla.mozilla.org/show_bug.cgi?id=605648#c90

gerhardol: Thanks for the explanation. The use of the Suggested comments - did I understand?

```
// Mouse wheel idle: Reset unused delta
                if (_lastMouseWheel.ElapsedMilliseconds > 1500)

...
// All scroll movement may not be used, save to next event
_mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
```

Move the following to the beginning of the else (or even function, also for horizontal scroll)
```
int maxRowIndex = _revisionGraph.Count - 1;
                    if (maxRowIndex < 0)
                    {
                        return;
                    }
```

MaxKoll: Yes, correct. I added comments per your suggestion, thanks.

gerhardol (approved):

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +873,37 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                int totalWheelDelta = (scrollLines * e.Delta) + _mouseWheelDeltaRemainder;
+                int wheelDeltaPerRow = SystemInformation.MouseWheelScrollDelta;
+
+                if (Math.Abs(totalWheelDelta) >= wheelDeltaPerRow)
+                {
+                    // The total wheel delta is consumed in multiples of wheelDeltaPerRow.
+                    // Save the remainder and credit it to the total wheel delta during the next MouseWheel event.
+                    _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
+                    int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
+
```
mstv, line 899: ❗
```suggestion
                // The total wheel delta is consumed in multiples of wheelDeltaPerRow.
                // Save the remainder and credit it to the total wheel delta during the next MouseWheel event.
                _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
                int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
                if (rowDelta != 0)
                {
```

MaxKoll: Of course, thank you!

mstv (approved):

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +873,36 @@ protected override void OnMouseWheel(MouseEventArgs e)
             }
             else
             {
-                base.OnMouseWheel(e);
+                // Reset unconsumed wheel delta when the mouse wheel is idle.
+                if (_lastMouseWheel.ElapsedMilliseconds > 1500)
```
RussKie, line 877: I suggest defining a const instead of using the magic number inline.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +873,36 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
+                if (rowDelta != 0)
+                {
+                    int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, maxRowIndex);
+
+                    // This will raise the Scroll event.
+                    FirstDisplayedScrollingRowIndex = toRowIndex;
+                }
+
+                _lastMouseWheel.Restart();
```
RussKie, line 905: So, if I open a repo, interact with it for a little while and then go out for lunch or go home, the stopwatch will continue counting... right? Is this desired? Why not just capture a timestamp at the time we exit the method, and calculate the delta when we enter the method again, and based on this delta decide whether we need to reset `_mouseWheelDeltaRemainder`?

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: > I stumbled upon this when I tested "smooth scrolling" and changed back to notched scrolling. While a notched wheel produces delta values that are multiples of 120, smooth scrolling produces arbitrary values and therefore leaves behind unconsumed total delta (since it's consumed in multiples of 120). Now assume the total delta is, say, -100 und you rotate the wheel by one notch in the positive direction, then nothing will happen, because the total delta is now at +20, but the delta needed for scrolling one row is 120. Whenever you change direction, the first notch will do nothing. I can imagine myself experiencing this when using a laptop with a mouse attached and switching between touchpad (= smooth scrolling) and mouse randomly. If you use only a touchpad or a free-spinning mouse, you might also experience a short "lag" or "jump" at the beginning of a scroll for the same reasons.
> 
> There might be more intricate ways to reset the total wheel delta (all needing a timer somehow, I guess), but a single timer is presumably good enough, and simple of course.

This looks like a very useful context, and I suggest adding it as a comment to the code.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +873,36 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                int scrollLines = SystemInformation.MouseWheelScrollLines switch
+                {
+                    // Value of -1 indicates the "One screen at a time" mouse option.
+                    -1 => _rowHeight > 0 ? (Height / _rowHeight) : 1,
+                    > 0 => SystemInformation.MouseWheelScrollLines,
+                    _ => 1
+                };
+
+                int totalWheelDelta = (scrollLines * e.Delta) + _mouseWheelDeltaRemainder;
+                int wheelDeltaPerRow = SystemInformation.MouseWheelScrollDelta;
```
RussKie, line 891: This would be great to have comments for as well in layman's terms.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
mstv: Why calculate `ElapsedMilliseconds` manually? Is a running `Stopwatch` expensive? I guess it just remembers its start time, and hides the fuzz with the correct handling of the elapsed time.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
MaxKoll: Agreed. Where would I define it?

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
MaxKoll: > So, if I open a repo, interact with it for a little while and then go out for lunch or go home, the stopwatch will continue counting... right?

You make it sound like leaving the stove on :-D. But I share your concern and looked it up:

> Is a running Stopwatch expensive? I guess it just remembers its start time, and hides the fuzz with the correct handling of the elapsed time.

Exactly right. The `Stopwatch` class is very lightweight. It just stores a timestamp and then calculates the difference, see: [https://github.com/dotnet/.../Stopwatch.cs](https://github.com/dotnet/runtime/blob/bff0f8487965a630a8c3fa107e17ec8e02e9b32c/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Stopwatch.cs)

The downside is that `Stopwatch` uses `QueryPerformanceFrequency()` to provide high precision, which isn't needed in this case. I just benchmarked `Stopwatch.ElapsedMilliseconds` against `Environment.TickCount64` (which provides ~15ms resolution) and found the latter to be ~6x faster  on my machine (27µs vs. 4.3µs). As this runs on the UI thread and every bit counts (I think?), I could change it like this:

```
            else
            {
                // Reset unconsumed wheel delta when the mouse wheel is idle.
                long currentTickCount = Environment.TickCount64;
                if (currentTickCount - _lastMouseWheelTickCount > 1500)
                {
                    _mouseWheelDeltaRemainder = 0;
                }

                _lastMouseWheelTickCount = currentTickCount;

                int scrollLines = SystemInformation.MouseWheelScrollLines switch
                {
```

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: At the top of the class should be fine.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
RussKie: Thank you. I confused it with S.W.F.Timer.

> just benchmarked `Stopwatch.ElapsedMilliseconds` against `Environment.TickCount64` (which provides ~15ms resolution) and found the latter to be ~7x faster on my machine (24µs vs. 3.4µs). As this runs on the UI thread and every bit counts (I think?), I could change it like this:

🚀

gerhardol: Hunting µs is something that can make a difference for the revision update loop, this is executed at GUI updates so not making a difference. But keep the change.

The only part that makes sense to optimize is the revision parsing and update loop as well as the rendering in addition to which Git commands that run.. Everything else will not be seen.
(I spent some days on optimizing some regex to sort the branches and tags in the sidepanel to get it down to a few ms instead of just below 100 ms, that was in retrospect not the best spent time, but it was kind of fun.)

MaxKoll: I was not planning on "hunting µs" here (sounds dangerous :sunglasses:), but I try to be open for every concern and suggestion, so I wanted to offer an alternative.

Thanks for the insight!

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -36,6 +37,9 @@ public sealed partial class RevisionDataGridView : DataGridView
 
         private int _backgroundScrollTo;
         private int _rowHeight; // Height of elements in the cache. Is equal to the control's row height.
+        private int _mouseWheelDeltaRemainder; // Corresponds to unconsumed scroll distance while scrolling via mouse wheel, see OnMouseWheel().
+
+        private long _lastMouseWheelTickCount; // Timestamp of the last vertical scroll via mouse wheel.
```
mstv, line 42: insert new fields alphabetically sorted and without empty lines, please

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +875,60 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                int wheelDeltaPerRow = SystemInformation.MouseWheelScrollDelta;
+                _mouseWheelDeltaRemainder = totalWheelDelta % wheelDeltaPerRow;
+                int rowDelta = -(totalWheelDelta - _mouseWheelDeltaRemainder) / wheelDeltaPerRow;
+                if (rowDelta != 0)
+                {
+                    int toRowIndex = Math.Clamp(FirstDisplayedScrollingRowIndex + rowDelta, 0, maxRowIndex);
+
+                    // Drop unconsumed wheel delta when reaching the upper or lower bound of the grid
+                    // to prevent the grid being stuck there for a moment.
+                    if (toRowIndex == 0 || toRowIndex >= maxRowIndex - visibleCompleteRowsCount + 1)
```
mstv, line 924: I find this identical condition clearer:
```suggestion
                    if (toRowIndex == 0 || toRowIndex + visibleCompleteRowsCount > maxRowIndex)
```

mstv (commented):
nits

MaxKoll:
I get `System.ArgumentOutOfRangeException` from `FirstDisplayedScrollingRowIndex = toRowIndex;` again. It happens rarely while rows are being readded after the grid has been cleared, e.g. after closing the _Resolve merge conflicts_ dialogue. Apparently, _revisionGraph.Count is also unreliable in that moment.

Ideas? Should I just suppress this exception?

gerhardol:
> I get `System.ArgumentOutOfRangeException` from `FirstDisplayedScrollingRowIndex = toRowIndex;` again. It happens rarely while rows are being readded after the grid has been cleared, e.g. after closing the _Resolve merge conflicts_ dialogue. Apparently, _revisionGraph.Count is also unreliable in that moment.

#9223

> Ideas? Should I just suppress this exception?

The grid need to be refreshed, but showing NBug will not help here, so why not?

mstv:
I think this is a race condition caused by setting `RowCount` asynchronously.
Using `_revisionGraph.Count` is peeking not yet published information.
Scrolling should be limited to `RowCount` - perhaps `- visibleCompleteRowsCount + 1`.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
MaxKoll: Call me stupid... It's not clear to me, sorry. Without touching existing empty lines? Like this?

```
 ...
 private int _loadedToBeSelectedRevisionsCount = 0;

 private int _backgroundScrollTo;
 private long _lastMouseWheelTickCount;
 private int _mouseWheelDeltaRemainder;
 private int _rowHeight;

 private VisibleRowRange _visibleRowRange;
 ...
```

Or structured by association, like this?

```
 ...
 private int _loadedToBeSelectedRevisionsCount = 0;

 private int _backgroundScrollTo;
 private int _rowHeight;

 private long _lastMouseWheelTickCount;
 private int _mouseWheelDeltaRemainder;

 private VisibleRowRange _visibleRowRange;
 ...
```

mstv: You are not at all. I did not want to request a sorting of all fields. So the first one might be the best compromise.
Structuring is arbitrary too often. "Insert alphabetically" is our simple coding style.

GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
```diff
@@ -865,7 +874,67 @@ protected override void OnMouseWheel(MouseEventArgs e)
...
+                    }
+
+                    try
+                    {
+                        // This will raise the Scroll event.
+                        FirstDisplayedScrollingRowIndex = toRowIndex;
+                    }
+                    catch (ArgumentOutOfRangeException)
+                    {
+                        // Tried to scroll to nonexistant row.
```
mstv, line 935:
```suggestion
                        // Tried to scroll to nonexistent row.
```

mstv (approved):
Thank you

RussKie (approved):
Great comments!
Thank you

MaxKoll:
Thank you all for your help and suggestions!

PR #11048, opened 2023-06-14
Simplify revision grid repainting

MaxKoll:
## Proposed changes

Repaint by invoking Update() at the end of each BackgroundUpdater operation instead of using a quasi-periodic approach.

### Before

Repaints are triggered periodically by the Scroll event and do not coincidence with the asynchronous revision graph updates.

### After

- Repaint after every revision graph update.
- Simpler code.

## Test methodology 

- Observing WM_PAINT messages via WndProc() override.
- Lots of scrolling the revision grid.

## Test environment(s) 

- Git 2.38.1.windows.1
- .NET 6.0.16
- DPI 134dpi (140% scaling)

## Merge strategy

----

### DIFF
--- a/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
+++ b/GitUI/UserControls/RevisionGrid/RevisionDataGridView.cs
@@ -23,9 +23,6 @@ namespace GitUI.UserControls.RevisionGrid
         private readonly SolidBrush _authoredHighlightBrush;
 
         private readonly BackgroundUpdater _backgroundUpdater;
-        private readonly Stopwatch _lastRepaint = Stopwatch.StartNew();
-        private readonly Stopwatch _lastScroll = Stopwatch.StartNew();
-        private readonly Stopwatch _consecutiveScroll = Stopwatch.StartNew();
         private readonly List<ColumnProvider> _columnProviders = new();
 
         internal RevisionGraph _revisionGraph = new();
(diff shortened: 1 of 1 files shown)


mstv:
Occasionally, only one graph lane is shown on refresh after an operation - for long time or until a manual refresh. It might be related to some background git commands which can take very long on the affected (private) repos. I need to observe it longer.

gerhardol (commented):
Seem OK, have not used it

MaxKoll:
> Occasionally, only one graph lane is shown on refresh after an operation - for long time or until a manual refresh. It might be related to some background git commands which can take very long on the affected (private) repos. I need to observe it longer.

I must admit I have not tested it intensely. I'm unsure how this could affect anything other than repaints during scrolling, as the only thing gone is the periodic call to `Update()` in `OnScroll()`. Every other line is just overhead.

I would therefore assume that the added `await this.InvokeAsync(Update);` inside `UpdateVisibleRowRangeInternalAsync()` is causing problems, but how?

RussKie:
The rebase operation is notorious for leaving the graph incomplete and undrawn (e.g., https://github.com/gitextensions/gitextensions/issues/9223). Perhaps, this could provide more testing opportunities.

mstv:
It is different from #9223 where no graph is displayed at all.

mstv:
Thank you for bringing this up. Included in and superseded by #11719.

